PR #556, opened 2014-08-10
Implement tuples as syntax sugar for newtype for record types

andreypopp:
Discuss!

Tuples are represented as

    newtype TupleX a b ... = { _1 :: a, _2 :: b, ... }

This commit implements a desugaring pass which translates TupleLiteral values
and TupleBinder binders into AST corresponding to chosen representation.

It's OK If you are not going to merge this. Anyway, I gained more understanding of PureScript internals.

### DIFF
--- a/prelude/prelude.purs
+++ b/prelude/prelude.purs
@@ -22,6 +22,15 @@ module Prelude
   , not
   , Semigroup, (<>), (++)
   , Unit(..), unit
+  , Tuple2(..)
+  , Tuple3(..)
+  , Tuple4(..)
+  , Tuple5(..)
+  , Tuple6(..)
+  , Tuple7(..)
+  , Tuple8(..)
+  , Tuple9(..)
+  , Tuple10(..)
   ) where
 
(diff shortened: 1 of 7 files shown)


hdgarrood:
Now that #863 has been merged, I guess it's feasible to have something like this approach but with type synonyms rather than newtypes (generated on the fly, in that desugar step?) How would you feel about that?

paf31:
I'm not sure how I feel about it, honestly. I think records are superior to tuples in every regard apart from the lack of support for instances (even #863 won't let us define a `Show` instance for a tuple for example), and I don't think it's a good thing to encourage users to use tuples. But, they are definitely handy sometimes.

paf31:
With the `Data.Tuple.Nested` changes, I'm thinking I should close this.

garyb:
Yeah, I think so. I still do occasionally wish for native tuples, but that's usually after I've been writing a lot of Haskell again ;) I can definitely live with our version.

PR #614, opened 2014-09-26
sketch out what an Int type could look like in the prelude

jdegoes:
More intended for conversation but illustrates what I'd like to see in Prelude. There's a guarantee that an `Int` will always be integral so it's much safer to use for things like array indexing, case matching, etc.

### DIFF
--- a/prelude/prelude.purs
+++ b/prelude/prelude.purs
@@ -98,7 +98,7 @@ module Prelude
     show = showNumberImpl
 
   instance showInt :: Show Int where
-    show (Int a) = show a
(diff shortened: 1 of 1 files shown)


paf31:
I'll close this since it's out of date and needs to be modified for the new `Num` class. I'd also like to wait for the rewrite rules changes.

garyb:
I have a revised version here: https://github.com/garyb/purescript-integers

Not worth using until we have rewrites though, so will wait until that is sorted before moving it over.

PR #674, opened 2014-11-03
Added docstring support to docgen

obmarg:
- Docstrings are now parsed around declarations as `DocStringDeclarations`.
- Docstrings on constructors for data declarations are parsed
- docgen prints out documentation along with type signatures.

I needed to copy a lot of code from `Text.Parsec.Token` so I could override its comment parsing to leave docstrings.  I had tried to copy a small subset, but there's a lot of interdependencies so it ended up easier to copy the whole `makeTokenParser`.  I'm open to any better suggestions.

There's some example output from purescript-drawing & purescript-node-args here: https://gist.github.com/obmarg/838c3770cb4b9a09e5c2

Not sure if this is ready yet, but thought I'd create a PR for discussion.  This is my first purescript contribution (and my first time working with a real haskell project) so feel free to tell me I'm doing it wrong.

I have compiled test-everything & starter-kit with these changes.  test-everything failed on purescript-string-parsers, but I checked and that was the behaviour against master, so don't think that's my fault.


paf31:
Ok, I just merged the lexer and comments changes that I wanted in place before tacking docstrings, so now would be a great time to revisit this if you're still interested in discussing it.

Here's my current line of thinking: we can use markdown for inline documentation, using Haddock-like `|` markers to indicate the start of a docstring line. That markdown would be merged into the output of `psc-docs`. `psc-docs` would probably need to be changed to use some Markdown library instead of just `Writer [String]`.

Pursuit could also be updated to work along the same lines, although that would be a bigger change, since the Pursuit database would have to be updated too.

What do you think?

/cc @garyb

garyb:
:+1:

obmarg:
Sounds good to me.  Great to see you've got comments in the AST now, should make adding proper docstring support a lot easier.

I've actually been meaning to re-visit this PR since you first merged in your new lexer changes, but been a bit short on time the past month or so,  Would like to help with implementing the changes to `psc-docs` if possible,  though I'm not going to be offended if you'd rather get on with it yourself.  I might not have time for the next week or so anyway.

Think this PR can probably be closed either way, the bulk of my code isn't needed any more.

paulyoung:
I'd love it if a [docset](http://kapeli.com/docsets) was generated so I could use it with [Dash](http://kapeli.com/dash).

The instructions under "[7. Any HTML Documentation](http://kapeli.com/docsets#dashDocset)" are probably most relevant, as well as the info at the bottom about contributing.

paf31:
I think that everything in this PR has been gradually incorporated in the past few releases. Let me know if I missed something, but I'll close this for now.

PR #744, opened 2014-11-30
Specify mtl and transformers dependencies

paf31:
I tried moving code over to `ExceptT` instead of `ErrorT`, but it seems to have different semantics for `MonadPlus`, which don't work out easily for us (among other things). I'm going to move that issue out of 0.7 for now.

This change simply specifies the dependency bounds for `mtl` and `transformers` for the various front-end executables.


garyb:
I guess we can delete this branch now?

paf31:
Yes, thanks.

PR #770, opened 2014-12-13
Remove Data.Eq from Prelude

### LINKED ISSUE #769: Remove Data.Eq from Prelude
I'm pretty sure nobody uses this module, which is good because it breaks referential transparency.

garyb:
Resolves #769.

Also make use of multiline strings for all Prelude FFI, I've been meaning to do that for a while.


garyb:
I think this is basically safe to merge now.

paf31:
:ship:

Issue #788, opened 2014-12-27
Type wildcards don't work in record or row extension.


joneshf:
Simple example:

```purescript
module Wild where

  type Foo = { | _}
```

Gives:

```
➜  purs  psc wild.purs
Error at wild.purs line 3, column 3 - line 4, column 1: 
Error in type synonym Foo
Type variable _ is undefined
```

paf31:
Fixed by https://github.com/purescript/purescript/commit/f7685450f21f5e91406ddf0ef7aeea4ffc567ff7

joneshf:
Cool. Any ideas when we might get a new version? There's been a whole bunch of changes.

paf31:
Good point, I'll try to make a release soon.

PR #792, opened 2015-01-02
Added parsing check for wildcard type when parsing row endings.

RossMeikleham:
Partial fix for #788. However the example given in #788:  

```purescript
module Wild where

  type Foo = { | _}
```
Now gives:

➜  purs  psc wild.purs
Warning: "wild.purs" (line 3, column 19):
unexpected type wildcard
expecting :: or operator
ambiguous use of a left associative operator Use --force to continue

due to the noWildcards check in Declarations.hs:

```haskell
parseTypeSynonymDeclaration =
  TypeSynonymDeclaration <$> (P.try (reserved "type") *> indented *> properName)
                         <*> many (indented *> kindedIdent)
                         <*> (indented *> equals *> noWildcards parsePolyType)
```

### DIFF
--- a/examples/passing/TypeWildcardsRecordExtension.purs
+++ b/examples/passing/TypeWildcardsRecordExtension.purs
@@ -1,0 +1,6 @@
+module Main where
+    
+foo :: forall a. {b :: Number | a} -> {b :: Number | _}
+foo f = f
+
+main = Debug.Trace.trace "Done"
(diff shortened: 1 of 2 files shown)


paf31:
:+1: Thanks!

I don't think `type Blah = { | _ }` should be allowed, since wildcards in type synonyms are generally disallowed.

PR #798, opened 2015-01-07
Render comments in generated code

### LINKED ISSUE #518: Preserve comments in generated code
I would like an option to preserve line and block comments in the generated JavaScript.

paf31:
Resolves #518 

This is still a bit of a work in progress, but the code is functional.

Any comments? (no pun intended) @garyb @joneshf

### DIFF
--- a/hierarchy/Main.hs
+++ b/hierarchy/Main.hs
@@ -84,7 +84,7 @@ superClasses :: P.Declaration -> [SuperMap]
 superClasses (P.TypeClassDeclaration sub _ supers@(_:_) _) =
   fmap (\(P.Qualified _ super, _) -> SuperMap (Right (super, sub))) supers
 superClasses (P.TypeClassDeclaration sub _ _ _) = [SuperMap (Left sub)]
-superClasses (P.PositionedDeclaration _ decl) = superClasses decl
+superClasses (P.PositionedDeclaration _ _ decl) = superClasses decl
 superClasses _ = []
 
 inputFile :: Parser FilePath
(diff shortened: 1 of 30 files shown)


paf31:
I think this is ready to review.

If everything looks ok, I'll release this as 0.6.3.

garyb:
Looks good to me, I take it comments are brought through `CoreFn` so they can be preserved for the JS output?

paf31:
@garyb Yep, I added comments to the `Ann` type.

garyb:
Ah sorry, that was badly phrased - I was asking about the motivation there, I had seen that it was done.

paf31:
Oh right, yes. I keep them around to convert them to the `JSComment` constructor in the generated code.

One thing that is probably worth discussing - what format do we want to encourage for our embedded docs? Something like Haddocks seems a bit too heavy. I think Markdown would work quite nicely, and would integrate neatly with things like Pursuit and `psc-docs`. The downside is having to figure out significant whitespace in the context of line and block comments.

garyb:
Markdown would definitely be my preferred format. Maybe the whitespace issue could be mitigated by having some kind of required format for the comment for markdown to be enabled, like
```
-- |
-- | this
-- |
```
or something, which is kinda Haddocksy. It wouldn't have to span multiple lines, just works for any comments starting with `-- |`.

paf31:
Is the idea that the markdown content would begin after the first instance of `| ` on each line, so that `|` is essentially there to align things?

garyb:
Yep, something along those lines is what I had in mind.

paf31:
I think that makes sense.

PR #800, opened 2015-01-07
More completion improvements for PSCI

vkorablin:
Complete type names for `:k`, complete data constructors where identifiers are expected. Also only complete unqualified names when the module that contains them is imported. (Otherwise they're unavailable!)

I had to drill a little deeper into the data structures that represent modules; not all of the resulting stuff is what I'd consider 'pretty', so feedback is very welcome!

### DIFF
--- a/psci/Main.hs
+++ b/psci/Main.hs
@@ -185,19 +185,19 @@ quitMessage = "See ya!"
 
 -- Haskeline completions
 
-data CompletionContext = Command String | FilePath | Module | Identifier | Fixed [String] | Multiple [CompletionContext]
+data CompletionContext = Command String | FilePath String | Module | Identifier | Fixed [String] | Multiple [CompletionContext]
                          deriving (Show)
 
 -- |
 -- Decide what kind of completion we need based on input.
-completionContext :: String -> Maybe CompletionContext
-completionContext cmd@"" = Just $ Multiple [Command cmd, Identifier]
-completionContext cmd@(':' : _ )
+completionContext :: String -> String -> Maybe CompletionContext
+completionContext cmd@"" _ = Just $ Multiple [Command cmd, Identifier]
+completionContext cmd@(':' : _ ) _
   | cmd `elem` C.commands || cmd == ":" = Just $ Command cmd
-completionContext (':' : c : _) = case c of
+completionContext (':' : c : _) word = case c of
   'i' -> Just Module
   'b' -> Just Module
-  'm' -> Just FilePath
+  'm' -> Just $ FilePath word
   'q' -> Nothing
   'r' -> Nothing
   '?' -> Nothing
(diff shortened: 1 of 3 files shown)


paf31:
This looks good, but I'll merge it into my local copy and test first.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -77,6 +81,7 @@ exportedDctors (Module _ decls exps) ident =
   where
   dctors = concatMap getDctors decls
   getDctors (DataDeclaration _ _ _ ctors) = map fst ctors
```
paf31: Probably need to handle data binding groups here too.

psci/Main.hs
```diff
@@ -238,30 +240,60 @@ completion = completeWordWithPrev Nothing " \t\n\r" findCompletions
...
+            getTypeName (P.TypeSynonymDeclaration name _ _) = Just name
+            getTypeName (P.DataDeclaration _ name _ _) = Just name
+            getTypeName (P.PositionedDeclaration _ d) = getTypeName d
+            getTypeName _ = Nothing
+
+  identNames :: P.Module -> [N.Ident]
+  identNames (P.Module _ ds exports) = nub [ ident | ident <- mapMaybe (getDeclName exports) ds ]
+    where getDeclName :: Maybe [P.DeclarationRef] -> P.Declaration -> Maybe P.Ident
+          getDeclName exts decl@(P.ValueDeclaration ident _ _ _)  | P.isExported exts decl = Just ident
+          getDeclName exts decl@(P.ExternDeclaration _ ident _ _) | P.isExported exts decl = Just ident
```
paf31: Probably need to handle binding groups here too.

paf31:
Just tested, and everything looks great, but for some reason `:t f` doesn't autocomplete to `:t flip` for my any more. Any idea why?

vkorablin:
Thanks for the feedback, I'll take a look.

vkorablin:
Regarding breakage: `Language.PureScript.AST.Declarations.isExported` didn't handle a certain case. I'll take care of the binding groups tomorrow.

paf31:
We can fix binding groups separately if you like. I'd like to make a release of the compiler today, and it would be nice to have this included, if the `flip` issue is a simple fix.

vkorablin:
Sure, go ahead. c542b4ff52695a753d95d7a50c46223995bdebd8 has the fix for that. Heads up: including `Language.PureScript.Sugar.BindingGroups` (for `collapseBindingGroups`/`collapseBindingGroupsModule`) from `Language.PureScript.AST.Declarations` results in an import loop.

paf31:
:+1: Awesome, thanks!

I don't think we need to import `collapseBindingGroups`, just unwrap those declarations inside the binding group. We can deal with it separately though.

PR #804, opened 2015-01-09
Parentheses lost when different operators of same precedence in expression

sztupi:
Expressions with different operators of equal JavaScript precedence did not keep the original structure, but were flattened:
"a - (b + c)" compiled as "a - b + c"; 
"a / (b * c)" compliled as "a / b * c"


paf31:
:+1: Thanks!

PR #805, opened 2015-01-10
Add fmap as non-infix alias for <$>

johanneshilden:
As per discussion on IRC chan.

### DIFF
--- a/prelude/prelude.purs
+++ b/prelude/prelude.purs
@@ -8,7 +8,7 @@ module Prelude
   , ($), (#)
   , (:), cons
   , Show, show
-  , Functor, (<$>), (<#>), void
+  , Functor, (<$>), (<#>), void, fmap
   , Apply, (<*>)
   , Applicative, pure, liftA1
   , Bind, (>>=)
@@ -119,6 +119,9 @@ module Prelude
 
   class Functor f where
     (<$>) :: forall a b. (a -> b) -> f a -> f b
+
+  fmap :: forall f a b. (Functor f) => (a -> b) -> f a -> f b
+  fmap = (<$>)
 
   (<#>) :: forall f a b. (Functor f) => f a -> (a -> b) -> f b
   (<#>) fa f = f <$> fa


prelude/prelude.purs
```diff
@@ -120,6 +120,9 @@ module Prelude
   class Functor f where
     (<$>) :: forall a b. (a -> b) -> f a -> f b
 
+  fmap :: forall f a b. (Functor f) => (a -> b) -> f a -> f b
+  fmap = (<$>)
```
michaelficarra: I don't like the name `fmap` for this function. If we have it, it should just be `map`.

paf31: Both good points. `map` would overlap with `Data.Array`, which is not the end of the world and would just necessitate a version bump.

Moving to `fmap` as the class default would break a lot of existing code though.

michaelficarra: I'm okay with both of those. More than okay, actually, I would prefer them.

paf31: Well, we can break lots of things before 1.0, but still, let's put it to a vote. Generally, I'd be in favor of using named things by default.

garyb: :+1: for named and `map`.

It seems like there may be an opportunity coming up for extreme breaking of things with good reason - if we can get rewrite rules to where we want and entirely externalise `Prelude`, then things like `Functor` and `Monad` should be moved into their own modules under `Control` anyway in my opinion. If we go ahead with that, re-naming the operations at the same time seems like the time to do it.

jdegoes:
:-1: I understand the motivation, but this sort of thing as no end in site, as literally all the operators could have named equivalents. Also, you are talking about breaking nearly every library ever written. Which I'm OK with doing for a good reason, but wanting to type `fmap` instead of `(<$>)` is maybe not the best reason, especially considering two changes could eliminate the need for breaking all these libraries.

I vote to close this PR and focus on #806 instead. Alternatively, we could implement support for default functions in type class declarations, which would make changes like this backward compatible (even though, to be clear, I am not in favor of this and view #806 as the superior approach).

paf31:
Closing this, since I think we decided #806 is going to be the way forward.

PR #811, opened 2015-01-13
fixed #807

### LINKED ISSUE #807: negate (-42) compiles to invalid JavaScript
Input:

```purescript
x = negate (-42)
```

Actual output:

```javascript
var x = --42;
```

Desired output:

```javascript
var x = -(-42);
// or...
var x = - -42;
```

philopon:
Hi!

I fixed #807.

### DIFF
--- a/examples/passing/Operators.purs
+++ b/examples/passing/Operators.purs
@@ -67,6 +67,12 @@ module Main where
   test17 :: Number
   test17 = negate (-1)
 
+  test18 :: Number
+  test18 = negate $ negate 1
+
+  test19 :: Number
+  test19 = negate $ negate (-1)
+
   main = do
     let t1 = test1 1 2 (\x y -> x + y)
     let t2 = test2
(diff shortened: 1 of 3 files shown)


michaelficarra:
Well this should fix it, but now all negations will have a space in them.

philopon:
Sorry, I underestimated it.

I'll fix it soon!

paf31:
This might even be an issue in my `pattern-arrows` library. I'll have a look as well.

michaelficarra:
I think the way to solve this is much like the way you've done now, except also disallow negative numbers in `JSNumericLiteral` and require those to be unary negation operations (which is how JavaScript actually works anyway). This will have the added benefit of handling `negate (negate -1)` properly, which this current patch does not do.

philopon:
I agree. so I rewrote. but I think It can fix easily even if JSNumericLiteral has sign.
Moreover, I think that independent negate can optimize easily. `negate (negate a)` to `a`.

Maybe, `removeSignedLiterals` should be renamed.

paf31:
Looks good to me. @michaelficarra ?

michaelficarra:
Very nice. :+1:

Issue #803, opened 2015-01-09
Warnings


paf31:
The type checker monad only supports errors, not warnings. Use `WriterT` to add a warnings capability.

hdgarrood:
Refs #33?

garyb:
I think #33 should raise errors rather than warnings actually.

nicodelpiano:
Hello, is it correct to think that the solution will be something alike?
```
newtype Check a = Check { unCheck :: StateT CheckState (WriterT MultipleWarnings (Either MultipleErrors)) a }
```
where MultipleWarnings depends on creating a new environment for warnings as errors have.
As WriterT is a monad transformer, this attachment to the type checker monad will give warning capability.

paf31:
Yes, but we might want to put `WriterT` inside `ErrorT` so that we still get warnings in the case when an error occurs (I _think_ that's the right way round).

If you're interested in working on this, you could pick one or two very simple warnings to begin with, so that we at least have a place to add new ones as we think of them. Some ideas:

- Name in binder shadows existing name
- Type variable got shadowed
- Unused name in binder
- Unused type variable

garyb:
> (I think that's the right way round)

I'm glad it's not just me :) always takes me a bit of trial and error to work out which way to layer transformers.

paf31:
I think of it this way: `ErrorT` of `m` just replaces the result type `a` inside `m` with `Either e a`, so wherever you would previously return a result, you can now instead return an error. 

Similarly, `WriterT` replaces `a` with `(a, w)`.

So if you put `WriterT` on the outside and fail, you don't get the warnings you accumulated, but if `ErrorT` wraps `WriterT`, you essentially replace `a` in `(a, w)` and get `(Either e a, w)`.

nicodelpiano:
You are right, but `WriterT` needs a monad as a second parameter which will be transformed to add the desired capability.
The signature of `WriterT` is `WriterT w m a`, where `w` is the type of the data we want to accumulate, and `m` is the monad which will be wrapped.
I can't figure out what is the base stack monad (`m`) which we may refer here (I thought it was `ErrorT`)

Also I've been working in a new module to handle warnings. It's basically the same as errors have. I'll make a pull request.

paf31:
@nicodelpiano What I'm suggesting is

```purescript
newtype Check a = Check { unCheck :: StateT CheckState (ErrorT MultipleErrors (Writer MultipleWarnings) a } deriving ...
```

[...]

Issue #745, opened 2014-11-30
Use an explicit type for type errors


paf31:
This is a first step toward moving to `transformers-0.4.*` and `ExceptT`. We need to:

- add a type (not `String`) for errors
- replace calls to `strMsg` with the appropriate constructor everywhere.
- update `monad-unify` to provide an appropriate error type, and handle that error type in `TypeChecker.Monad`.

mjgpy3:
A few questions:
 - Should this new type be an instance of `Control.Monad.Error`?
 - Do you mean to ammend `monad-unify` so that it includes this type, or update the cabal reference?

Sorry if the questions are trivial.

paf31:
I think we have two options:

- Move `monad-unify` back into the PureScript repo. It's going to need some changes anyway, to support some new features, so that probably isn't a bad option.
- Add a generic error type to `monad-unify`, something like `data UnifyError t e = OccursCheckFailed Unknown t | Other e`.

I'm leaning more in the direction of the first option lately, to support things like row constraints and better instance inference.

PR #813, opened 2015-01-14
Added basic warningsWriter

mjgpy3:
See #803. I have no clue if this is right, I thought I'd take a stab at it. Feedback is much appreciated.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -110,6 +110,7 @@ library
                      Language.PureScript.TypeChecker.Unify
                      Language.PureScript.TypeClassDictionaries
                      Language.PureScript.Types
+                     Language.PureScript.Warnings
     exposed: True
     buildable: True
     hs-source-dirs: src
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -1,6 +1,6 @@
 -----------------------------------------------------------------------------
 --
--- Module      :  Language.PureScript.Error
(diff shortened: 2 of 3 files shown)


paf31:
So we definitely want to use `WriterT`, but we will need to integrate it into the existing typechecker monad transformer stack in `Language.PureScript.TypeChecker.Monad`.

Also, I think it makes sense to make the type of warnings different from the type of errors. I would probably like to see both of them become data types, with separate functions to render them for display to the user.

mjgpy3:
Awesome, thanks so much for the feedback. I'll take another stab at it tonight.

paf31:
It might be a good idea to tackle #745 first. That would probably help you to familiarize yourself with how errors work, and the structure of the type checker monad.

A good first step might be to make a list of all of the different types of errors in the application, and a list of possible warnings (e.g. a name got shadowed).

mjgpy3:
Great! I'll have a look at it tonight, thanks!

paf31:
Is this abandoned? It would need quite a lot more work, so I will close it unless there are plans to revive it.

mjgpy3:
Yes, it's abandoned, I haven't had the time, unfortunately.

mjgpy3:
Can you advise on a more entry-level issue? I am very interested in this project and would love to contribute, but feel unable to dive in the deep end.

paf31:
Ok, no worries. Maybe error messages would be a good place to make some improvements? Maybe take a look at the error-messages label on the issue tracker?

PR #826, opened 2015-01-23
Bumped utf8-string version and updated code accordingly.

### LINKED ISSUE #825: utf8-string IO Wrappers deprecated
Since GHC supports UTF-8, System.IO.UTF8 no longer exists. The quick fix would be to put an upper bound on utf8-string. The better fix is to update code so that it doesn't use these wrappers anymore. 

See [here](https://github.com/glguy/utf8-string/commit/31969fdecdab5b0d3732f15937886df2d3d7522e) for more details.

aspidites:
Fixes #825 
* added lower bound of >= 1 to utf8-string
* bumped minor release number
* updated code to not use the IO wrappers

Couldn't find any reference to `purescript-test-everything` or `starter-kit` in the repo (though it's rather late so I could have missed it), but I did reconfigure with tests enabled and ran the one test suite with success.

### DIFF
--- a/hierarchy/Main.hs
+++ b/hierarchy/Main.hs
@@ -25,13 +25,12 @@ import Options.Applicative
 import System.Directory (createDirectoryIfMissing)
 import System.FilePath ((</>))
 import System.Exit (exitFailure, exitSuccess)
-import System.IO (stderr)
+import System.IO (hPutStr, stderr)
 
 import Text.Parsec as Par (ParseError)
 
 import qualified Language.PureScript as P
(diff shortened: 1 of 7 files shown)


paf31:
Looks great. Is there any reason to keep the `utf8-string` dependency though?

aspidites:
When I removed it things weren't compiling, so I thought you might be using it for some other purpose. I can take another look tomorrow after work to see exactly what is using the package.

paf31:
Merged. Let's deal with removing utf8string separately.

PR #827, opened 2015-01-23
Specify version bounds for utf8-string

phadej:
Workaround https://github.com/purescript/purescript/issues/825
http://stackoverflow.com/questions/28110769/system-io-utf8-not-found-installing-purescript/28111390#28111390


phadej:
This is obsolete now

PR #834, opened 2015-02-03
Fix #812

### LINKED ISSUE #812: Failure to type check bug using row type variable
@paf31 asked me via IRC channel to document this bug I encountered.  I used PS 0.6.1 on NixOS when encountering it.  Basic case to reproduce is below.  The bug is still present if `data` is used instead of `newtype`.

```
newtype NewType a = NewType {|a} -- NewType is a wrapper around arbitrary records

myRecord = {a: 0, b: 0, c: 0}    -- make a record
updatedRecord = myRecord {a = 1} -- update it with record update syntax

n1 = NewType updatedRecord                    -- This line type checks
n2 = NewType $ myRecord {a = 1}               -- But this line does not.  It's the same thing!
n3 = NewType $ myRecord {a = 1, b = 0, c = 0} -- This line type checks, however

-- The error for n2 is:
Error at test.purs line 9, column 6: 
Error in declaration n2
Cannot unify (c :: Prim.Number, b :: Prim.Number) with ()

I don't understand why n1 works but n2 doesn't.
I'm trying to get the n2 style to work because the records might be large,
and usually only 1 field needs to be updated at a time.  I can modify
NewType to enable this if necessary.
```

paf31:
@garyb @joneshf Could you please review? I'm running `test-everything` now.

### DIFF
--- a/examples/passing/NewtypeWithRecordUpdate.purs
+++ b/examples/passing/NewtypeWithRecordUpdate.purs
@@ -1,0 +1,15 @@
+-- https://github.com/purescript/purescript/issues/812
+
+module Main where
+
+import Debug.Trace
+
+newtype NewType a = NewType (Object a)
+
+rec1 :: Object (a :: Number, b :: Number, c:: Number)
+rec1 = { a: 0, b: 0, c: 0 } 
+
+rec2 :: NewType (a :: Number, b :: Number, c :: Number)
+rec2 = NewType (rec1 { a = 1 })
+
+main = trace "Done"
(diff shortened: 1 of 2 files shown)


paf31:
`test-everything` seems to be fine.

garyb:
I don't really understand this one, but if it fixes stuff and doesn't break anything :+1: I guess ;)

Why isn't the `go ((p,v):ps') [] u@(TUnknown _)` case needed after adding adding the guarded case?

Issue #814, opened 2015-01-15
Type checker issue with polymorphic constructors in records


paf31:
I've seen this a couple of times now. From @rtrvrtg 's gist:

```
module Metadata.UI where

import Data.Function
import Data.Maybe

import qualified Thermite as T
import qualified Thermite.Html as T
import qualified Thermite.Html.Elements as T
import qualified Thermite.Html.Attributes as A
import qualified Thermite.Action as T
import qualified Thermite.Events as T
import qualified Thermite.Types as T

import Debug.Trace

-- | Depicts all the actions taking place in our app.
data Action = LogIn String | LogOut

-- | Stores app state.
type State = {
    authToken :: Maybe String
}

-- | Initial app state: logged out with no data.
initialState :: State
initialState = { authToken: Nothing }

-- | Renders a page
render :: T.Render State _ Action
render ctx s _ = T.div [ A.className "app-container" ] [ current ]
  where
    current = if isJust s.authToken
                then loggedIn
                else loggedOut
    loggedOut = T.div' [ loginForm ]
    loggedIn = T.p' [ T.text "You're logged in!"
                    , T.button [ T.onClick ctx (\_ -> LogOut) ] [ T.text "Log Out" ] ]
    loginForm = T.form [ A.action "#", A.method "get" ]
                  [ fieldGroup [
                        T.label [] [ T.text "Email Address" ]
                      , T.input [ A._type "text", A.name "email_address", A._id "email-address", A.required "true" ] []
                    ]
                  , fieldGroup [
                        T.label [] [ T.text "Password" ]
                      , T.input [ A._type "password", A.name "password", A._id "password", A.required "true" ] []
                    ]
                  , fieldGroup [
                        T.button [ A._type "submit", T.onClick ctx (\_ -> LogIn "foo") ] [ T.text "Log In" ]
                  ] ]
    fieldGroup children = T.div [ A.className "form-group" ] children

-- | Modifies the global app state based on the action taking place
performAction :: T.PerformAction _ Action (T.Action _ State) 
performAction _ (LogIn t) = T.modifyState (\o -> { authToken: Just t })
performAction _ LogOut    = T.modifyState (\o -> { authToken: Nothing })

spec :: T.Spec _ State _ Action
spec = T.simpleSpec initialState performAction render

main = do
    let component = T.createClass spec
    T.render component {}
```

Problem is as follows:

```
-- Error at src/UI.purs line 55, column 29 - line 57, column 1: 
-- Error in declaration performAction
[...]

PR #835, opened 2015-02-03
Fix #810 and #814: don't infer polymorphic types for data constructors.

### LINKED ISSUE #810: "This might be a bug in the type checker"
I posted an example in the mailing list before, but this issue keeps reappearing for me (Not to say it drives me crazy). So I have come up with a simpler example:

This code:
~~~
module Main where

data ClickIndex a =
      CILabel Number 
    | CIAction (ActionDescription a)
data ActionDescription a = AD a | AB

clicked :: forall a. ClickIndex a -> ActionDescription a
clicked ci = test.ani 
    where
        test = case ci of 
                    CILabel n ->  {ani: AB, num : n}
                    CIAction ad -> {ani: ad, num : 1}
~~~
doesn't compile with error message:
>> Error in declaration clicked
>> Cannot unify a13 with a7.

When I add this type annotation:
~~~
                    CILabel n ->  {ani: (AB :: ActionDescription _), num : n}
~~~
it compiles (which I can't understand). 

Now I use to write ugly code with big functions, and the line number of the error message is not
at the right point, and my brain doesn't tell me at all where I have to add 
such type annotations to make the type checker happy. Hope I made my point clear.

Thanks again for providing the beautiful purescript language to us.

paf31:
@garyb @joneshf Could you please review? `test-everything` passes, and the code in #814 now builds.

### DIFF
--- a/examples/passing/810.purs
+++ b/examples/passing/810.purs
@@ -1,0 +1,11 @@
+module Main where
+
+data Maybe a = Nothing | Just a
+
+test :: forall a. Maybe a -> Maybe a
+test m = o.x
+  where
+    o = case m of Nothing -> { x : Nothing }
+                  Just a  -> { x : Just a }
+
+main = Debug.Trace.trace "Done"
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -52,6 +52,7 @@ import Language.PureScript.Errors
 import Language.PureScript.Kinds
 import Language.PureScript.Names
 import Language.PureScript.Pretty
+import Language.PureScript.Traversals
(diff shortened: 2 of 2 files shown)


garyb:
:+1:

joneshf:
Just to be sure what this fixes. Is this a fix for that bug where you have a `Nothing` say, that isn't at the top level, and the compiler tries to infer `Maybe a` when it should infer `Maybe String` or something?

paf31:
Yes, in `infer`ence mode, the typechecker will now not infer the polymorphic type `Maybe a` for the expression `Nothing`, for example. In `check`ing mode, nothing has changed.

joneshf:
Cool. Thanks!

PR #836, opened 2015-02-03
Fix #816

### LINKED ISSUE #816: Fixity declarations leak into externs even when operator is not exported

paf31:
@garyb @joneshf Could you please review?


garyb:
:+1:

PR #841, opened 2015-02-05
Fix option flags in psc-make.

paulyoung:
This pull request swaps two flags in the `psc-make` options to put them in the correct place.


paf31:
:+1: Thanks!

PR #843, opened 2015-02-06
Name errors

### LINKED ISSUE #740: Overlapping class and data constructor name generate invalid code
```js
var PS = PS || {};
PS.Test = (function () {
    "use strict";

    // here's the data constructor
    function Foo() {

    };
    Foo.value = new Foo();

    // here's the class dictionary constructor
    function Foo(foo) {
        this.foo = foo;
    };

    var foo = function (dict) {
        return dict.foo;
    };
    return {
        Foo: Foo, // oops! we only have one Foo
        foo: foo
    };
})();
```

garyb:
- Fixes #740.
- Fixes a bug where the error message for duplicate data constructor or type definitions wasn't being handled in the right place, and instead was being shown as an import conflict.
- Adds position info to error about duplicate definitions.

### DIFF
--- a/src/Language/PureScript/Sugar/Names.hs
+++ b/src/Language/PureScript/Sugar/Names.hs
@@ -331,7 +331,7 @@ findExports = foldM addModule $ M.singleton (ModuleName [ProperName C.prim]) pri
   addDecl mn env (ExternDataDeclaration tn _) = addType env mn tn []
   addDecl mn env (ValueDeclaration name _ _ _) = addValue env mn name
   addDecl mn env (ExternDeclaration _ name _ _) = addValue env mn name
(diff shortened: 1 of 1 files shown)


paf31:
Looks nice. I wonder if some of the duplicated code could be factored out into a helper function though?

garyb:
I know what you mean, but I couldn't really figure out how in the end, as the checks are a little different in each case:

`addType`:
- Multiple types with the same name (special case, as the check in `addExport` won't work here)
- A type with the same name as a class
- For each data constructor:
  - A data constructor with the same name as any other data constructor in the module
  - A data constructor with the same name as a class

`addTypeclass`
- A class with the same name as a type
- A class with the same name as a data constructor

paf31:
:+1: Makes sense, thanks!

PR #844, opened 2015-02-06
Extended object literals

### LINKED ISSUE #384: Extended Object Literals / Object Sections
I don't like the idea of implementing Scala-like argument placeholders (e.g. `map (_.foo)`) in full generality, but it seems there is a very useful (and much less syntactically ambiguous) specialization of the idea, which can be used to construct objects.

This change would cause the following

```
{ foo: _, bar: _, baz: "baz" }
```

to desugar to

```
\foo bar -> { foo: foo, bar: bar, baz: "baz" }
```

This would be very useful for things like applicative notation:

```
{ foo: _, bar: _, baz: "baz" } <$> readFoo <*> readBar
```

garyb:
This will resolve #384 when done... but it's not a real PR right now. I seem to have broken the linker?

### DIFF
--- a/examples/passing/ObjectWildcards.purs
+++ b/examples/passing/ObjectWildcards.purs
@@ -1,0 +1,14 @@
+module Main where
+
+import Control.Monad.Eff
+import Debug.Trace
+
+mkRecord = { foo: _, bar: _, baz: "baz" }
+
+getValue :: forall e. Eff (| e) Boolean
+getValue = return true
+
+main = do
+  obj <- { value: _ } <$> getValue
+  print obj.value
+  trace (mkRecord 1 "Done!").bar
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -93,6 +93,7 @@ library
                      Language.PureScript.Sugar.CaseDeclarations
                      Language.PureScript.Sugar.DoNotation
                      Language.PureScript.Sugar.Names
+                     Language.PureScript.Sugar.ObjectWildcards
                      Language.PureScript.Sugar.Operators
                      Language.PureScript.Sugar.TypeClasses
                      Language.PureScript.Sugar.TypeDeclarations
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -320,6 +320,11 @@ data Expr
   --
   | ObjectLiteral [(String, Expr)]
   -- |
+  -- An object constructor (object literal with underscores). This will be removed during
+  -- desugaring and expanded into a lambda that returns an object literal.
+  --
+  | ObjectConstructor [(String, Maybe Expr)]
+  -- |
   -- An record property accessor expression
   --
   | Accessor String Expr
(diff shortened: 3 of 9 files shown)


garyb:
It seems to be traversal related. It starts failing when reduced to this:

```
desugarObjectConstructors :: Module -> Module
desugarObjectConstructors (Module mn ds exts) = Module mn (map desugarDecl ds) exts

desugarDecl :: Declaration -> Declaration
desugarDecl d =
  let (f, _, _) = everywhereOnValues id id id
  in f d
```

paf31:
I think you need to add your new module to the Cabal file.

This looks good though, I'm excited to try it out.

garyb:
Duh, of course! It's obviously been too long since I worked on this. :)

paf31:
I just looked at the code. It looks great, but I have a couple of comments:

- I wonder if the syntax could be simplified to something like `{ specified: 1, unspecified }`, meaning `\unspecified -> { specified: 1, unspecified: unspecified }`
- Could you please add an example?

Thanks.

garyb:
Yep, no problem. I've not even tried it yet with the whole linker thing.

garyb:
With the syntax thing, we could definitely do it without the `_`, but I figure that aligns quite nicely with if we do the `_.x` as syntactic sugar for `(\obj -> obj.x)` thing too? Or did you not want to do that anyway?

Also, do we want `obj { x = _ }` too? That might be handy.

paf31:
I like the `_.x` idea, but I think I'd like to find a use case for updates before we implement it.

garyb:
I was thinking for applicative style again, for instance the stuff I was looking at in names:

```
addValue env mn name = updateExportedModule env mn $ \m -> do
  values <- addExport (exportedValues m) name
  return $ m { exportedValues = values }
```
Could be:
```
addValue :: ExportEnvironment -> ModuleName -> Ident -> Either ErrorStack ExportEnvironment
addValue env mn name = updateExportedModule env mn $ \m ->
  m { exportedValues = _ } <$> addExport (exportedValues m) name
```

paf31:
Interesting - we could definitely add it and see how we like it.

garyb:
Updated with stuff working now. This is going to be super nice :)

Can either merge now if you're happy, or hold off until the property/update stuff is done too?

paf31:
Looks great :)

Does `test-everything` work? I'm always wary of parser changes.

Otherwise, :ship:

paf31:
Maybe we can do something about the coverage first though :P

paf31:
Oh, it's a test failure. Never mind...

garyb:
Huh, that's odd! Will figure it out in a sec.

Yep, `test-everything` still works. The only slight change now is `_` is never a valid identifier on its own, previously it would have been treated as a variable so the error would have been `_ is undefined` whereas now it's treated as a parse failure.

garyb:
Whoops, I broke type wildcards and somehow managed not to run the tests... fixing now.

garyb:
Finally...

paf31:
:+1:

PR #845, opened 2015-02-07
More object syntax

garyb:
Getters and updaters, as discussed in the other PR.

### DIFF
--- a/examples/passing/ObjectGetter.purs
+++ b/examples/passing/ObjectGetter.purs
@@ -1,0 +1,9 @@
+module Main where
+
+getX = _.x
+
+point = { x: 1, y: 0 }
+
+main = do
+  Debug.Trace.print $ getX point
+  Debug.Trace.trace $ _.value { value: "Done!" }
--- a/examples/passing/ObjectGetter.purs
+++ b/examples/passing/ObjectGetter.purs
@@ -7,4 +7,5 @@ point = { x: 1, y: 0 }
 main = do
   Debug.Trace.print $ getX point
   Debug.Trace.trace $ (." 123 string Prop Name ") { " 123 string Prop Name ": "OK" }
+  Debug.Trace.trace $ ((.x) >>> (.y)) { x: { y: "Nested" } }
   Debug.Trace.trace $ (.value) { value: "Done!" }
--- a/examples/passing/ObjectGetter.purs
+++ b/examples/passing/ObjectGetter.purs
@@ -1,9 +1,10 @@
 module Main where
 
-getX = _.x
+getX = (.x)
 
 point = { x: 1, y: 0 }
 
 main = do
   Debug.Trace.print $ getX point
-  Debug.Trace.trace $ _.value { value: "Done!" }
+  Debug.Trace.trace $ (." 123 string Prop Name ") { " 123 string Prop Name ": "OK" }
+  Debug.Trace.trace $ (.value) { value: "Done!" }
--- a/examples/passing/ObjectUpdater.purs
+++ b/examples/passing/ObjectUpdater.purs
@@ -1,0 +1,23 @@
+module Main where
+
+import Control.Monad.Eff
+import Debug.Trace
+
+foreign import data Error :: !
+
+foreign import error
+  """
+  function error() {
+    throw new Error("Object update failed");
+  }
+  """ :: forall e. Eff (error :: Error | e) Unit
+
+getValue :: forall e. Eff (| e) Boolean
+getValue = return true
+
+main = do
+  let record = { value: false }
+  record2 <- record { value = _ } <$> getValue
+  if record2.value
+    then trace "Done"
+    else error
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -338,6 +338,11 @@ data Expr
   --
   | ObjectUpdate Expr [(String, Expr)]
   -- |
+  -- Partial record updater. This will be removed during desugaring and
+  -- expanded into a lambda that returns an object update.
+  --
+  | ObjectUpdater Expr [(String, Maybe Expr)]
+  -- |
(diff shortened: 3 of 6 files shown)


garyb:
Checked with `test-everything` too - all seems good.

jdegoes:
These changes will be nice -- help level the playing field wrt Haskell records, as well as provide "lightweight lenses" for disposable records that aren't used often enough to warrant real lenses.

michaelficarra:
I would prefer sections. Then we could use `(.x)` syntax to mirror `(* 2)`.

garyb:
@michaelficarra: That didn't even occur to me, but you're right, it makes much more sense. I'll give it a try.

garyb:
Chained property access was actually a little tricky to parse before, so it only supported `_.x`, but section-style it's easy. The only reservation I have about that is it means the getters differ from the literal constructors and updaters, as you can't do `{ x: { y: _ } }` or `obj { x = { y = _ } }` - that seems like more of a job for proper lenses.

michaelficarra:
You can't compose constructors the same way? Both `{ y: _ } >>> { x: _ }` and `{ x: { y: _ } }` seem fine to me.

garyb:
Yep, that works too, this is all syntactic sugar for normal functions, so `(.x)` is desugared into `(\obj -> obj.x)`, `{ x: _ }` is `(\x -> { x: x })`, `obj { x = _ }` is `(\x -> obj { x = x })`. I just meant being able to compose reduces the pain of not being able to do `(.x.y)` or `{ y: _ } >>> { x: _ }`.

paf31:
Nice :)

Could you please add a couple of cases to the pretty printer so that it doesn't spin if we hit these cases?

Also, I wonder if it would be useful to be able to section the record itself on update:

```purescript
updateFoo o = _ { foo = 1 } <$> o
```

But that could be something for later.

garyb:
Yep no problem, those cases are in now. The other changes in there are just a few things hlint was complaining about.

paf31:
:+1: Thanks very much!

This deserves a version bump - do you think it's good to release now, or did you want to make any other changes?

garyb:
I was going to have a stab at operator sections, it seems like an opportune time as we've just added some section-like syntax. I know that's not actually on the list for 0.7, but hopefully will be relatively easy to do. If I don't get it sorted by tomorrow though, go for it.

paf31:
Ok, no rush.

garyb:
I need to make the change to use `Supply` for this too actually.

paf31:
Good point. I'll wait for that, and then push a release. Thanks very much - 0.6.5 should be a really nice release.

garyb:
No problem!

I'm working on the change now, shouldn't be long.

jdegoes:
The problem with sections for getters is that they don't have a counterpart for setters.

```
(.foo)
???
```

Versus:

```
_.foo r
{ foo = _ }
```

Also via @paf31's suggestion:

```
_ { baz = 1 }
```

So I'd argue that `(.foo)` is a false consistency with operator sections because `.foo` is not really an operator; maybe it helps to think of it that way, maybe not. So for that reason, plus the reason of there being no "operator section" analogue for the other record operations, I think `_.foo` is probably better.

As a bonus, it will be immediately familiar to anyone whose worked with Scala before.

garyb:
Hmm, maybe you're right.

The thing I liked about the sections is it's obvious that they behave in a similar way to standard operators - they're a representation of a partially applied thing, whereas `_.x` looks more like a normal accessor. I suppose `_` generally is taken to be a hole of some kind though, so maybe that's not entirely valid.

Also on your point, there's no section like `(obj.)` - although if we introduce first class labels that could become meaningful.

paf31:
`(.foo)` isn't an operator, but it could become one if we add first class labels, using the optimizer to turn the saturated application `(.) obj "lbl"` into `obj["lbl"]`. Obviously there are some syntactic issues there, but it's a possibility.

I like the symmetry of using `_`. I'm not a fan of the Elm-style `.foo` without parens, due to syntactic ambiguity modulo whitespace, but since we parenthesize, I don't feel strongly either way.

jdegoes:
A section like `(obj.)` would handle polymorphic labels and values, but there would still be no section-like notation for updating or setting. Since these would have to use `_`, it seems more consistent to just say that if you want "sections" for records of any kind, you use the underscore. 

Indeed, the reason Scala has this underscore notation is exactly for products (tuples in Scala's case, but records in Purescript's case), because curried functions don't really need them. It's where you have an n-product and want a function that will stick a value in one of those n slots.

But that said, I don't have a strong preference either way, I just think the analogy with operator sections is (currently) a false one and (a bit) inconsistent with the record update syntax.

garyb:
Sold. :+1: 

Should I change this again now before too many people start making use of `(.prop)` syntax? I guess we'd want to make a new release too.

paf31:
I guess that if we're committing to the changes, then we should do this sooner rather than later and rerelease.

jdegoes:
Either way's fine by me (with the above caveats), I'm just so happy to have this stuff. Lots of code will become much more readable. Thanks @garyb!

michaelficarra:
@paf31: what would the signature of `_ { foo = 1 }` be? `forall a. (| a) -> (foo :: Number | a)`?

garyb:
I guess it would be inferred as 

``` purescript
forall a. { foo :: Number | a } -> { foo :: Number | a }
```

as it would desugar to:

``` purescript
\obj -> obj { foo = 1 }
```

michaelficarra:
So it wouldn't be able to extend a record that doesn't have a `foo` field?

garyb:
I think that's the type that "updater syntax" currently has. It would be nicer if it could extend the record, but I think we'd need some kind of record constraints to make that work, right @paf31?

paf31:
Correct, it's just like regular object updates right now. There's no concept of extension because we don't have lacks-constraints.

PR #846, opened 2015-02-07
Operator sections

### LINKED ISSUE #410: Partial application of operators
So you can do things like `(x >)` instead of having to do `((>) x)`, or `(> a)` instead of `(flip (>) x)`.

garyb:
Resolves #410, successfully ran `test-everything`.

There is one problem, but Haskell gets this wrong too: `(- 1) 1` results in a type error rather than `0`, as the `negate` case is used to parse the `(- 1)` part of the expression, so it's desugared as `(negate 1) 1`.

### DIFF
--- a/examples/passing/OperatorSections.purs
+++ b/examples/passing/OperatorSections.purs
@@ -1,0 +1,21 @@
+module Main where
+
+foreign import eqeqeq
+  """
+  function eqeqeq(x) {
+    return function (y) {
+      if (x == y) return x;
+      throw new Error("Unexpected result: " + x + " /== " + y);
+    };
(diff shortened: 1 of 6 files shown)


garyb:
Actually, I've just realised I should use `Supply` to generate the name in the lambda rather than using `_lhs` too.

garyb:
There we go.

Once this is in, I'm all for the version bump and release. :cocktail:

paf31:
:+1: Brilliant :)

joneshf:
:100:

jdegoes:
:wine_glass:

PR #847, opened 2015-02-08
Supply object wildcards

garyb:
Well that took a bit longer than expected. I hadn't implemented all the traversals the first time so it seemed to be failing randomly after changing to use the `M` traversal... anyway, sorted now, and I've added traversals for the `OperatorSection` stuff too. They'll probably never actually be reached, but I'd like to avoid anything like the problem I just had happening again.


paf31:
:+1: Looks great, thanks!

I have one quick bug that I want to fix, then I'll make a release.

PR #850, opened 2015-02-08
Extended infix operators

paf31:
@garyb Could you please review this one?

### DIFF
--- a/examples/passing/ExtendedInfixOperators.purs
+++ b/examples/passing/ExtendedInfixOperators.purs
@@ -1,0 +1,20 @@
+module Main where
+
+zipWith :: forall a b c. (a -> b -> c) -> [a] -> [b] -> [c]
+zipWith _ [] _  = []
+zipWith _ _  [] = []
+zipWith f (x:xs) (y:ys) = f x y : zipWith f xs ys
+
+test1 = [1, 2, 3] `zipWith (+)` [4, 5, 6]
+
+comparing :: forall a b. (Ord b) => (a -> b) -> a -> a -> Ordering
+comparing f = compare `Data.Function.on` f
+
+sum [] = 0
+sum (x:xs) = x + sum xs
+
+test2 = [1, 2, 3] `comparing sum` [4, 5, 6]
+
+main = do
+  Debug.Trace.print test1
+  Debug.Trace.print test2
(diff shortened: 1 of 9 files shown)


src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -411,7 +419,7 @@ parseValue = withSourceSpan PositionedValue
                   ]
   operators = [ [ P.Prefix (P.try (C.indented *> symbol' "-") >> return UnaryMinus)
                 ]
-              , [ P.Infix (P.try (C.indented *> C.parseIdentInfix P.<?> "operator") >>= \ident ->
+              , [ P.Infix (P.try (C.indented *> parseIdentInfix P.<?> "operator") >>= \ident ->
```
garyb: Perhaps should name this something other than `parseIdentInfix`, and update the hint for parsec?

garyb:
Looks good to me apart from that minor nitpick :)

Will be interesting to see how this pans out. Just because it _can_ be abused doesn't necessarily seem like a reason not to have it either.

:+1:

garyb:
I overhauled the [syntax guide](https://github.com/purescript/purescript/wiki/Language-Guide:-Syntax) this morning by the way (I know we promote the book over that stuff now, but it's still valuable as a quick reference I think) so you might want to update that after releasing this too.

garyb:
Oh, and you might want to tweak the pretty printer for `OperatorSection` now?

paf31:
All good points, thank you.

examples/passing/ExtendedInfixOperators.purs
```diff
@@ -0,0 +1,20 @@
+module Main where
+
+zipWith :: forall a b c. (a -> b -> c) -> [a] -> [b] -> [c]
+zipWith _ [] _  = []
+zipWith _ _  [] = []
+zipWith f (x:xs) (y:ys) = f x y : zipWith f xs ys
+
+test1 = [1, 2, 3] `zipWith (+)` [4, 5, 6]
```
jdegoes: This is really nice and has great readability -- also easier to explain the `` rule as operating over expressions as opposed to just "identifiers". It would be useful to put a nested example in these tests to demonstrate the parser can handle nesting.

paf31: Sure thing. Will be nice to get the coverage up as well.

Issue #824, opened 2015-01-22
Variable named 'undefined' disappears in generated code


philopon:
Hi!

this Reproducing code

```.purs
module Test where

import Control.Monad.Eff

foreign import valid     :: forall a. a

foreign import undefined :: forall a. a
foreign import null      :: forall a. a

testValid :: Eff _ _
testValid = return valid

testUndef :: Eff _ _
testUndef = return undefined

testNull :: Eff _ _
testNull = return null
```

generates

```.js
...
    var testValid = Prelude["return"](Control_Monad_Eff.monadEff)(valid);
    var testUndef = Prelude["return"](Control_Monad_Eff.monadEff)();
    var testNull = Prelude["return"](Control_Monad_Eff.monadEff)($$null);
...
```

I think case of `undefined` and `null` isn't prospective behavior. I think it should generate

```.js
    var testUndef = Prelude["return"](Control_Monad_Eff.monadEff)(undefined);
    var testNull = Prelude["return"](Control_Monad_Eff.monadEff)(null);
```

or report compile time error.

Regards

paf31:
What error would these report? `null` and `undefined` are keywords, so they get escaped.

paf31:
Oh, but `undefined` disappearing definitely _is_ a bug.

garyb:
That's an optimisation that came about when we got rid of applying `{}` in all the typeclass code.

philopon:
> null and undefined are keywords

Yes, I understand it. but, I think

* `foreign import null :: forall a. a` seems like importing javascript null keyword.
* escaped `foreign import null` is useless(I must know mangling rule to use it).

so, I reported `null` case.

paf31:
@garyb But undefined should be escaped here, and then not removed.

garyb:
Ah yep, I'm not saying it's not a bug, just explaining how it's happening.

garyb:
This is no longer an issue, I think we fixed it quite a while back actually - the FFI changes would have done it.

paf31:
The generated code is now

```purescript
exports["null"] = $foreign["null"];
exports["undefined"] = $foreign.undefined;
```

which works when run, but seems a bit ... dodgy. Should we escape `undefined` like `null` here?

garyb:
[...]

PR #855, opened 2015-02-10
Escape built-in JS proper names

garyb:
Resolves purescript/purescript-random#4, purescript/purescript-random#5

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/Common.hs
+++ b/src/Language/PureScript/CodeGen/JS/Common.hs
@@ -135,7 +135,47 @@ nameIsJsReserved name =
               , "volatile"
               , "while"
               , "with"
-              , "yield" ]
+              , "yield" ] || properNameIsJsReserved name
 
 moduleNameToJs :: ModuleName -> String
-moduleNameToJs (ModuleName pns) = intercalate "_" (runProperName `map` pns)
+moduleNameToJs (ModuleName pns) =
+  let name = intercalate "_" (runProperName `map` pns)
+  in if properNameIsJsReserved name then "$$" ++ name else name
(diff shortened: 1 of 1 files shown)


jutaro:
Just stumbled about this problem when using my own random patch. I'm sorry I've not tested it properly, AND I'm happy the problem is resolved more fundamentally, as it is an evil side effect.

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -135,7 +135,47 @@ nameIsJsReserved name =
...
-moduleNameToJs (ModuleName pns) = intercalate "_" (runProperName `map` pns)
+moduleNameToJs (ModuleName pns) =
+  let name = intercalate "_" (runProperName `map` pns)
+  in if properNameIsJsReserved name then "$$" ++ name else name
+
+-- |
+-- Checks whether a proper name is reserved in Javascript.
+--
+properNameIsJsReserved :: String -> Bool
+properNameIsJsReserved name =
```
michaelficarra: "Reserved" already has a meaning in JavaScript. How about `isJsBuiltin`? Also, do we want to catch name conflicts with built-ins such as `parseInt` and `setTimeout`?

garyb: Both good points, I'll update the list to contain those as well, but I'll need to fix #824 first.

paf31:
:+1: Looks good. :ship:

garyb:
Will merge this now then to fix the immediate problem in `random`, and then revise based on @michaelficarra's suggestions when I look at #824.

PR #856, opened 2015-02-10
Fix #815

### LINKED ISSUE #815: @ as an operator no longer works in 0.6.3
I'm writing a library that uses the @ symbol as an operator.  This worked in 0.6.1.  I just upgraded to 0.6.3 and get the following error.  It appeared at export, and when I removed the export of @, it appeared at the declaration, then at the infixl line; just went down the file complaining any time @ was used in an operator context.
```
unexpected @
expecting symbol
```
I can work around it by changing the operator to something else, but I'd like to know if this is a bug in 0.6.3 or desired behavior moving forward in releases.  Thanks.

paf31:
@fluffynukeit Could you please review? Thanks.

### DIFF
--- a/examples/passing/Operators.purs
+++ b/examples/passing/Operators.purs
@@ -73,6 +73,11 @@ module Main where
   test19 :: Number
   test19 = negate $ negate (-1)
 
+  test20 :: Number
+  test20 = 1 @ 2
+    where
+    (@) x y = x + y * y
+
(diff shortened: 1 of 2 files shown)


fluffynukeit:
If test20 passes and the named patterns test still passes, then I think this is good to go.  The test20 is the kind of usage I was trying.

PR #857, opened 2015-02-10
Fix #796. Rebuild modules before each command is executed.

### LINKED ISSUE #796: In-memory modules in psci prevent out-of-date code from being rebuilt
New to Purescript.

I'm following the book, and have an okay workflow. However, when I write a new function in a module that's also loaded in psci and give it an `:r`, then call `FooModule.barFunc` psci will:

* Tab-complete on the name of the function, but;
* Die with:
```
Error at  line 1, column 1 - line 2, column 1: 
Error in module $PSCI
Unknown value 'FooModule.barFunc'
```
Occasionally after much `grunt`'ing and `:r`'ing, it might actually run the new function, but any other new ones I write will be invisible yet visible to psci as described above.

What am I doing wrong?

paf31:
@garyb @fosskers Could you please review? Thanks.

### DIFF
--- a/psci/Main.hs
+++ b/psci/Main.hs
@@ -141,6 +141,16 @@ loadAllModules files = do
     return (Right filename, content)
   return $ P.parseModulesFromFiles (either (const "") id) $ (Left P.RebuildNever, P.prelude) : filesAndContent
 
+-- |
+-- Load all modules, updating the application state
+--
+loadAllImportedModules :: PSCI ()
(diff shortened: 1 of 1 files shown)


garyb:
Looks good to me. Reloading every time isn't too slow I take it?

paf31:
It wasn't too bad in my tests, but yes, it could cause some slowdown.

PR #858, opened 2015-02-10
Reformat psc-docs output slightly, and embed markdown comments.

### LINKED ISSUE #802: Parse markdown in embedded documentation
Doc strings will be indicated in comments by a pipe at the start of each line. The pipe character will be used to disambiguate indentation in mixed-mode comments. Content to the right of the pipe will be markdown, to be inserted into generated documentation in `psc-docs`.

```
-- | Create a copy of the array without its first element.
-- |
-- | Running time: `O(n)`, where `n` is the length of the array.
-- |
-- | This function is partial. Specifically, `tail []` is undefined.
tail :: forall a. [a] -> [a]
```

paf31:
@garyb Would you please review this?

Fixes #802 

The output is also changed slightly to give each entry a (level 4) header, just to delimit the contents of the markdown comments a little bit.

### DIFF
--- a/prelude/README.md
+++ b/prelude/README.md
@@ -4,279 +4,321 @@
 
 ### Types
 
+#### `Ordering`
 
-    data Ordering where
-      LT :: Ordering
-      GT :: Ordering
-      EQ :: Ordering
+    data Ordering
+      = LT 
+      | GT 
+      | EQ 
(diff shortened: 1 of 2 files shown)


garyb:
I like it :+1:

PR #860, opened 2015-02-11
Better let bindings in psci, fix #782

### LINKED ISSUE #782: Commit to types of let-bound names in psci
tl;dr -- interactive evaluation in PSCI can sometimes clobber types of unrelated values

Minimal test case:
```
module Bug where
    newtype Const a b = Const a
    instance functorConst :: Functor (Const a) where
                    (<$>) fn (Const x) = (Const x)

    a = Const 1
    b = (&&) false <$> a
```

Load this file into PSCI:

```
> :m Bug.purs
> :i Bug
Compiling Prelude
Compiling Prelude.Unsafe
Compiling Data.Function
Compiling Data.Eq
Compiling Control.Monad.Eff
Compiling Control.Monad.Eff.Unsafe
Compiling Control.Monad.ST
Compiling Debug.Trace
Compiling Bug
> :t a
forall t21. Bug.Const Prim.Number t21
> :t b
Bug.Const Prim.Number Prim.Boolean
> let b = (&&) false <$> a
> :t a
forall t21. Bug.Const Prim.Number t21
```

So far so good. If, however, we assign `a` and then `b` *interactively*:

```
> let a = Const 1
> :t a
forall t19. Bug.Const Prim.Number t19
> let b = (&&) false <$> a
> :t a
Bug.Const Prim.Number Prim.Boolean <-- !!!
```

Oops! `a`'s type has changed after we evaluated something completely different! And they said Purescript didn't have uncontrolled side effects. :)

This happens only in PSCI; ahead-of-time compilation is ok, as can be seen in the first part of the transcript.

paf31:

_Edit_: note this also allows things like data declarations, type class instances and FFI imports in psci :)

### DIFF
--- a/psci/Commands.hs
+++ b/psci/Commands.hs
@@ -52,7 +52,7 @@ data Command
   -- |
   -- Binds a value to a name
   --
-  | Let (Expr -> Expr)
+  | Let [Declaration]
   -- |
   -- Find the type of an expression
   --
--- a/psci/Main.hs
+++ b/psci/Main.hs
@@ -69,7 +69,7 @@ data PSCiState = PSCiState
   { psciImportedFilenames   :: [FilePath]
   , psciImportedModuleNames :: [P.ModuleName]
   , psciLoadedModules       :: [(Either P.RebuildPolicy FilePath, P.Module)]
-  , psciLetBindings         :: [P.Expr -> P.Expr]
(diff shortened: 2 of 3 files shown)


michaelficarra:
Cool!

garyb:
:+1:

joneshf:
Nice!

PR #863, opened 2015-02-11
Allow rows in instance contexts

### LINKED ISSUE #642: Allow rows in instance contexts
Currently we disallow rows in constraints, but really all we need to do is restrict instance heads, not instance contexts. We are currently disallowing seemingly valid things like 

```
foo :: (MonadState { field :: Type } m) => m Type
```

As long as every instance doesn't mention any non-empty rows, we should be fine.

paf31:
@apsk @garyb Could you please review?

@apsk I worked from your pull request, but I needed to make some changes to support some additional use cases, and also added a couple of comments to explain what's going on and what needs to be improved.

Resolves #642.

### DIFF
--- a/examples/passing/RowPolyInstanceContext.purs
+++ b/examples/passing/RowPolyInstanceContext.purs
@@ -1,0 +1,20 @@
+module Main where
+
+class T s m where
+  state :: (s -> s) -> m Unit
+
+data S s a = S (s -> { new :: s, ret :: a })
+
+instance st :: T s (S s) where
+  state f = S $ \s -> { new: f s, ret: unit }
+
+test1 :: forall r . S { foo :: String | r } Unit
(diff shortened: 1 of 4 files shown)


garyb:
Awesome! :+1: :+1: :+1: 

It all looks great to me too, but I probably don't have a deep enough understanding of exactly what's going on to see any potential problems here anyway.

paf31:
This will probably need more testing, but `test-everything` works for me. I'll make a release soon, and we can see if there are any issues.

pseudonom:
This is separate from allowing anonymous records? Like

```purescript
instance showFoo :: Show ({foo :: String}) where
  show f = "f"
```
?

garyb:
Yeah, that's what @paf31 meant in #863 when he mentioned disallowing rows in instance heads.

Closed rows like you have in that example would perhaps be workable, but maybe even that is problematic. @paf31?

PR #866, opened 2015-02-12
Scoped type variables #347

paf31:
@garyb @joneshf Would you please review this?

High level overview:

- Kind inference updated to peel off top-level `forall`s and return the kinds of their bound variables.
- Skolems can now have their kind inferred, by using the handily stored type variable name they originated from.
- `withScopedTypeVars` brings scoped type variable kinds into scope.
- `skolemizeTypesInValue` updated to skolemize type annotations.

Still TODO:

- [x] Remove `Type` from `ScopedTypeVar` constructor.
- [x] Handle scoped type vars in `let` bindings.

### DIFF
--- a/examples/passing/ScopedTypeVariables.purs
+++ b/examples/passing/ScopedTypeVariables.purs
@@ -6,4 +6,16 @@ test1 f x = g (g x)
   g :: a -> a
   g y = f (f y)
 
+test2 :: forall a. (a -> a) -> a -> a
+test2 = h
+  where
+  h :: forall b. (b -> b) -> b -> b
+  h f x = g (g x)
+    where
+    g :: b -> b
+    g y = f (f y)
+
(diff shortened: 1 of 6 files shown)


paf31:
`test-everything` passes.

joneshf:
Seems alright, but I'm not too well versed in what's supposed to happen.

garyb:
Looks great. Have you tested it much with shadowed type variable names?

paf31:
@garyb Not really, but I can add a test.

garyb:
It's probably fine, just wondered if there might be issues with that for things like the skolem kind inference. I'll have a play myself if I get a chance before you do.

paf31:
Something like this?

garyb:
:shipit: :fireworks:

Issue #859, opened 2015-02-10
Add --no-comments option


paf31:
To squash comments in generated JS.

michaelficarra:
Wouldn't it be a `--comments` option that can be disabled with `--no-comments`?

paf31:
Well, comments are currently on by default, but that could change.

berewt:
What is your preferred option:
- remove them during the generation
- remove them during the pretty printing
- consider comment removal as a JS Optimizer

Although it's not really a code optimization, JS optimizer looks like the better place to implement it.

paf31:
I think I would like to put the changes into the code generator.

berewt:
Ok, I would have to check how to "carry" the no-comments option to the generator then.

paf31:
Just pass the `Options` object along to this function:

https://github.com/purescript/purescript/blob/d683d32aa8ed1cea30354d337bff7d9fac4bb3d6/src/Language/PureScript/CodeGen/JS.hs#L89

PR #869, opened 2015-02-12
no-comments option added

berewt:
Add --no-comments option #859

### DIFF
--- a/psci/Main.hs
+++ b/psci/Main.hs
@@ -28,7 +28,7 @@ import qualified Data.Map as M
 import Control.Applicative
 import Control.Monad
 import Control.Monad.Error (ErrorT(..), MonadError)
-import Control.Monad.Error.Class (MonadError(..))
+import Control.Monad.Reader (MonadReader, ReaderT, runReaderT)
 import Control.Monad.Trans.Class
 import Control.Monad.Trans.Maybe (MaybeT(..), runMaybeT)
 import Control.Monad.Trans.State.Strict
@@ -150,7 +150,7 @@ loadAllImportedModules = do
   modulesOrFirstError <- psciIO $ loadAllModules files
   case modulesOrFirstError of
     Left err -> psciIO $ print err
-    Right modules -> PSCI . lift . modify $ \st -> st { psciLoadedModules = modules } 
+    Right modules -> PSCI . lift . modify $ \st -> st { psciLoadedModules = modules }
 
 -- |
 -- Expands tilde in path.
@@ -314,7 +314,7 @@ completion = completeWordWithPrev Nothing " \t\n\r" findCompletions
 -- | Compilation options.
 --
 options :: P.Options P.Make
-options = P.Options False False False Nothing False False P.MakeOptions
+options = P.Options False False False Nothing False False False P.MakeOptions
 
 -- |
 -- PSCI monad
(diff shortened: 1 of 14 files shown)


paf31:
Looks good, but

- where is `removeComments` defined?
- I'd still prefer to pass the options to the code generator. That way, we can change our minds and make no comments the default, or even support different comment styles later on.

Actually, I've been thinking that @michaelficarra is right about making no comments the default, especially since I want to document the Prelude, and the code size is going to blow up.

berewt:
removeComments was already there (Purescript.CoreFn.Ann): it is used to remove the comments from ne annotation when a JSComment is generated.

At this stage, making no-comments the default can be done quite easily (rename no-comments to keep-comments, invert the condition in moduleToJS). Anyway, i will see how to do it at the codeGen phase.

berewt:
Just to clarify: by codegen, you talk about "Language.PureScript.Pretty.JS.prettyPrintJS" or do I miss the point?

jdegoes:
@paf31 I don't think anyone would deploy a Purescript (or *any* Javascript application) without first minifying. That said, I don't imagine myself looking at the JS code for comments.

Now adding comments for Closure compiler, that's another matter.

garyb:
@berewt by codegen he probably means still in `Language.PureScript.Pretty.JS, but by passing the option through to `nonRecToJS` and skipping the generation of `JSComment` there instead, when appropriate.

berewt:
@garyb Ok, I Functor's instances is an overkill ni this context. I might as well remove them and do the change inside `nonRecToJS` then.
Thanks.

paf31:
I need a `Functor` instance for something to do with rewrite rules, so leave them in, they'll come in handy :)

berewt:
Finally found the time to change it, carrying the options through the code generation was a bit more painful than expected.

garyb:
You could have just passed an extra argument from `moduleToJs` through `bindToJs` to `nonRecToJS`. ;)

I have no strong preference of that vs reader though, I'll let @paf31 weigh in.

paf31:
On the one hand, `ReaderT` seems like overkill, for such a short code path. On the other hand, it doesn't seem enough :) This is exactly what we should be doing with the `Options` object IMO, but we should be doing it everywhere. 

I just don't want to be in a situation where we use `ReaderT` in some places and not in others, so for me it's all or none, but I would prefer all.

berewt:
Using Reader was far more convenient for the refactoring as bindToJs is called by many other functions in the file.
I agree that it would be better if it was used everywhere to handle options though.
I will try to change them into parameter passing then.

berewt:
I can also check if I can generalize ReaderT. The weather forecasts are bad for the weekend anyway.

garyb:
Ah good point, I'd forgotten about the calls to `bindToJs` for `let`, etc.

berewt:
Yes, I used Reader in 7 functions I think and I need to pass options in all of them, that's why I move for a Reader instead of parameter passing.

berewt:
Hear is a version with generalized Reader, we can still go a bit Further:

* using `StateT` for the Environment in `PureScript.hs`
* using `ReaderT` in psc / psci for the options

I also tried to do a Reader version of the `JS.hs` changes but it's cumbersome, really.

src/Language/PureScript.hs
```diff
@@ -152,13 +157,14 @@ traverseEither f (Right y) = Right <$> f y
 -- If timestamps have not changed, the externs file can be used to provide the module's types without
 -- having to typecheck the module again.
 --
-make :: (Functor m, Applicative m, Monad m, MonadMake m) => FilePath -> Options Make -> [(Either RebuildPolicy FilePath, Module)] -> [String] -> m Environment
-make outputDir opts ms prefix = do
+make :: (Functor m, Applicative m, Monad m, MonadMake m) => FilePath -> [(Either RebuildPolicy FilePath, Module)] -> [String] -> ReaderT (Options Make) m Environment
```
paf31: I'd prefer to use a `MonadReader` constraint here, and in other places.

psci/Main.hs
```diff
@@ -402,7 +403,7 @@ handleDeclaration :: P.Expr -> PSCI ()
 handleDeclaration val = do
   st <- PSCI $ lift get
   let m = createTemporaryModule True st val
-  e <- psciIO . runMake $ P.make modulesDir options (psciLoadedModules st ++ [(Left P.RebuildAlways, m)]) []
+  e <- psciIO . runMake $ P.make modulesDir (psciLoadedModules st ++ [(Left P.RebuildAlways, m)]) [] `runReaderT` options
```
paf31: The `ReaderT` transformer can just be moved into the `PSCI` stack.

paf31:
This already looks _much_ nicer in my opinion. 

Please don't take my comments the wrong way - this is tidying up some stuff which I've wanted to do for a while, and this is just a convenient time since we're touching related code. If you don't want to include it in this PR, that's totally fine, but I think this is a good chance to tidy lots of stuff up, if we're doing the `ReaderT` thing anyway. So that said:

- I would like to use `mtl` classes like `MonadReader` instead of `ReaderT` directly. Many functions already work over an arbitrary monad, so it seems a shame to restrict to `ReaderT` on the outside.
- Eventually, I would like to do the same with a `MonadSupply` to remove the fixed `SupplyT` type constructor.

As I say though, I can merge now. Eventually I'd like to move to the `mtl` style in more places though.

berewt:
Don't worry, I totally understand that your comments wasn't negative but positive suggestions. I totally agree that using typeclasses instead of their instances is better too. 
I can go through the changes and move `ReaderT` to `MonadReader` easily I think.

I will do my best to introduce other generic `mtl` changes too.

Thanks for your feedbacks.

berewt:
I think I would need to introduce FlexibleContexts to do so, is it a problem?

paf31:
Nope, any sensible extensions like that are welcome.

berewt:
You can provide feedback on the style but don't push it yet, I will work on the rest of the Reader code today.

paf31:
Wonderful - thank you very much!

I think you might need to do a bit of merging after the recent optimizer changes, but there shouldn't be much to do.

One last thing - can we please change the flag to `--comments` or `-c`, to keep comments, instead of to remove comments? I think no comments makes a good default and matches other compilers like TypeScript.

I'll work on adding a `MonadSupply` class when this is merged.

berewt:
No problem, done.

paf31:
Looks good to me, thanks! @garyb, any comments?

@berewt Could you please squash the commits?

berewt:
Commits squashed!

garyb:
I do have a comment: Thanks! Looks good to me :+1:

berewt:
I amended the last commit to add the -c alternative.
And many thanks to both of you.

paf31:
This is really great, thanks!

PR #871, opened 2015-02-14
Fix #870

### LINKED ISSUE #870: Type classes issue after rows change
Now I remember why we don't allow unknowns to unify in `typeHeadsAreEqual`.

```purescript
> let f x = negate x
  
> :t f
forall t2. t2 -> t2
```

paf31:
I'm just going to merge this since it's pretty bad, and the fix is fairly harmless.


garyb:
:+1:

PR #872, opened 2015-02-15
Revise Num class

### LINKED ISSUE #868: Revise Num class
This could become a huge discussion, but I'll say up front - I don't want to do anything drastic to `Num` all at once.

A very refined type class hierarchy _is_ possible, but I think would be better contained in libraries.

That said, I'm pretty sure `Number` is the only instance of `Num` right now (no `Int` yet, etc.) and I don't think any code uses `Num` in the abstract, so we can change a few things, if we're careful not to break the optimizer.

Here is the current proposal (by @jdegoes):

```purescript
-- addition and multiplication
class Semiring a where 
  (+)   :: a -> a -> a
  zero  :: a
  (*)   :: a -> a -> a
  one   :: a

-- addition, multiplication, and subtraction
class (Semiring a) <= Ring a where
  (-) :: a -> a -> a

negate :: forall a. (Ring a) => a -> a
negate a = zero - a

-- ring with division (possibly non-commutative field)
class (Ring a) <= DivisionRing a where
  (/) :: a -> a -> a

-- definitely commutative field
class (DivisionRing a) <= Num a
```

and in `purescript-monoids`:

```purescript
-- additive monoid
newtype Additive a = Additive a

-- multiplicative monoid
newtype Multiplicative a = Multiplicative a

instance additiveSemiRing :: (SemiRing a) => Semigroup (Additive a) where 
  (<>) (Additive a) (Additive b) = Additive (a + b)

instance additiveMonoid :: (SemiRing a) => Monoid (Additive a) where
  mempty = Additive zero

-- todo: Additive functor, applicative, etc.

instance multiplicativeSemiRing :: (SemiRing a) => Semigroup (Multiplicative a) where 
  (<>) (Multiplicative a) (Multiplicative b) = Multiplicative (a + b)

instance multiplicativeMonoid :: (SemiRing a) => Monoid (Multiplicative a) where
  mempty = Multiplicative one

-- todo: Multiplicative functor, applicative, etc. 
```

garyb:
Resolves #868 

`(%)` is missing at the moment, so this isn't ready to merge yet.

### DIFF
--- a/examples/passing/OneConstructor.purs
+++ b/examples/passing/OneConstructor.purs
@@ -2,6 +2,6 @@ module Main where
 
 data One a = One a
 
-one (One a) = a
+one' (One a) = a
 
 main = Debug.Trace.trace "Done"
--- a/examples/passing/Rank2Data.purs
+++ b/examples/passing/Rank2Data.purs
@@ -12,18 +12,18 @@ module Main where
   runNat = \nat -> case nat of
     Nat f -> f 0 (\n -> n + 1)
 
-  zero = Nat (\zero _ -> zero)
+  zero' = Nat (\zero' _ -> zero')
 
   succ = \n -> case n of
-    Nat f -> Nat (\zero succ -> succ (f zero succ))
+    Nat f -> Nat (\zero' succ -> succ (f zero' succ))
 
   add = \n m -> case n of
     Nat f -> case m of
-      Nat g -> Nat (\zero succ -> g (f zero succ) succ)
+      Nat g -> Nat (\zero' succ -> g (f zero' succ) succ)
 
-  one = succ zero
-  two = succ zero
+  one' = succ zero'
+  two = succ zero'
   four = add two two
   fourNumber = runNat four
 
-  main = Debug.Trace.trace "Done"
+  main = Debug.Trace.trace "Done'"
--- a/examples/passing/ShadowedTCO.purs
+++ b/examples/passing/ShadowedTCO.purs
@@ -2,14 +2,14 @@ module Main where
 
 runNat f = f 0 (\n -> n + 1)
 
-zero z _ = z
+zero' z _ = z
 
-succ f zero succ = succ (f zero succ)
+succ f zero' succ = succ (f zero' succ)
 
-add f g zero succ = g (f zero succ) succ
+add f g zero' succ = g (f zero' succ) succ
 
-one = succ zero
-two = succ one
+one' = succ zero'
+two = succ one'
 four = add two two
 fourNumber = runNat four
 
--- a/prelude/README.md
+++ b/prelude/README.md
@@ -59,6 +59,13 @@
     class (Semigroupoid a) <= Category a where
       id :: forall t. a t t
 
+#### `DivisionRing`
+
+Ring where every nonzero element has a multiplicative inverse (possibly
+a non-commutative field) so that ```a `mod` b = zero```
+
+    class (ModuloRing a) <= DivisionRing a where
+
 #### `Eq`
 
     class Eq a where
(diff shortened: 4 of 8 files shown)


jdegoes:
Ah, forgot about modulo.

I'd be interested in feedback on this slight tweak to the hierarchy:

```purescript
-- definition: a / b * b + (a % b) = a
class (Ring a) <= ModuloRing a where
  (%) :: a -> a -> a
  (/) :: a -> a -> a

-- definition: a % b = zero
class (ModuloRing a) <= DivisionRing a
```

The idea here is that some types support modulo ring but not division ring: for example, `Int`, and `Polynomial`. Whereas, because division rings are capable of producing true multiplicative inverses for every value, the quantity `a / b * b` is definitionally equal to `a`, so `a % b` is `zero` for all `b`.

paf31:
Looks great. The tests don't really test the inlining stuff, so I assume that's all working. Also, can you please rerun `updateDocs.sh`.

Other than that, :ship: :+1:

garyb:
How does this look?

Inlining seems to be ok, tested with:
``` purescript
module Test where

import Debug.Trace

main = do
  print (-(1 + 2 / 3 * 4))
  print (1 + 2 - 3)
  print (1 + 2 / 3)
  print (one + 2)
  print (2 * zero)
  print (negate (2 * zero))
  print (-1)
  print (negate 1)
  print (1 * 2)
  print (4 % 3)
  print (4 `mod` 3)
```

Compiles to:
``` js
var PS = PS || {};
PS.Test = (function () {
    "use strict";
    var Prelude = PS.Prelude;
    var Debug_Trace = PS.Debug_Trace;
    var Control_Monad_Eff = PS.Control_Monad_Eff;
    var main = function __do() {
        Debug_Trace.print(Prelude.showNumber)(-(1 + (2 / 3) * 4))();
        Debug_Trace.print(Prelude.showNumber)((1 + 2) - 3)();
        Debug_Trace.print(Prelude.showNumber)(1 + 2 / 3)();
        Debug_Trace.print(Prelude.showNumber)(1 + 2)();
        Debug_Trace.print(Prelude.showNumber)(2 * 0)();
        Debug_Trace.print(Prelude.showNumber)(-(2 * 0))();
        Debug_Trace.print(Prelude.showNumber)(-1)();
        Debug_Trace.print(Prelude.showNumber)(-1)();
        Debug_Trace.print(Prelude.showNumber)(1 * 2)();
        Debug_Trace.print(Prelude.showNumber)(4 % 3)();
        return Debug_Trace.print(Prelude.showNumber)(Prelude.mod(Prelude.moduloRingNumber)(4)(3))();

    };
    return {
        main: main
    };
})();
```

I could add a case to automatically inline ```4 `mod` 3``` as `0`, but is it worth it?

garyb:
Ah yes the docs, will do!

paf31:
Looks great. I don't think inlining `mod` is necessary really.

jdegoes:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

garyb:
I'll let the tests run anyway and then will merge. :raised_hands:

paf31:
Also, did you run `test-everything`? I don't see any potential for breakage, but probably worth running anyway.

garyb:
Yep, it's still happy.

One thing I noticed while working on this, I used `pursuit` to check for definitions of `one` and `zero` - they show up in `Data.Monoid`, however that's not the "standard" monoid, it's from @joneshf's `purescript-algebra` - is it possible to add a URL for where things come from in `pursuit`?

PR #874, opened 2015-02-16
Relax `ModuloRing` to `ModuloSemiring`

### LINKED ISSUE #873: Loosen requirement: ModuloRing -> ModuloSemiring
It occurs to me that `ModuloRing` should really be turned into `ModuloSemiring`, so that it can be implemented for `Positive`, etc. Indeed, its law only requires a `Semiring` -- not a full `Ring`.

This change would also necessitate listing `Ring` as a superclass of `DivisionRing`, since right now, it gets that relationship through `ModuloRing`.

Apologies for not recognizing this earlier.

garyb:
Resolves #873.


paf31:
:+1:

PR #875, opened 2015-02-16
First cut of documenting some prelude functions.

fluffynukeit:
Documented some prelude functions, usually with accompanying short example.  My only concern is some of the examples use `Maybe`, which is not in the prelude, but which I think is a really easy to grok example of `Functor` and friends.

### DIFF
--- a/prelude/prelude.purs
+++ b/prelude/prelude.purs
@@ -34,10 +34,14 @@ module Prelude
   otherwise = true
 
   -- | Flips the order of the arguments to a function of two arguments. 
+  -- |
+  -- | `flip (/) 6 3 == 3 / 6`
   flip :: forall a b c. (a -> b -> c) -> b -> a -> c
   flip f b a = f a b
 
   -- | Returns its first argument and ignores its second. 
+  -- |
+  -- | `const 7 "whatever" == 7`
   const :: forall a b. a -> b -> a
   const a _ = a
 
@@ -59,9 +63,15 @@ module Prelude
   class Semigroupoid a where
     (<<<) :: forall b c d. a c d -> a b c -> a b d
 
+  -- | Function composition in right to left sense.
+  -- |
+  -- | `(f <<< g) x == f (g x)`
   instance semigroupoidArr :: Semigroupoid (->) where
     (<<<) f g x = f (g x)
 
+  -- | Function composition in left to right sense.
+  -- |
+  -- | `(f >>> g) x == g (f x)`
   (>>>) :: forall a b c d. (Semigroupoid a) => a b c -> a c d -> a b d
   (>>>) f g = g <<< f
 
@@ -74,14 +84,23 @@ module Prelude
   infixr 0 $
   infixl 0 #
 
+  -- | Applies the first argument to the second, making a backward pipe.
+  -- |
+  -- | `reverse $ "str" ++ "ing" == "gnirts"`
   ($) :: forall a b. (a -> b) -> a -> b
   ($) f x = f x
 
+  -- | Applies the second argument to the first, making a forwards pipe.
+  -- |
+  -- | `"str" ++ "ing" # reverse == "gnirts"`
   (#) :: forall a b. a -> (a -> b) -> b
   (#) x f = f x
 
   infixr 6 :
 
+  -- | Prepends an element to a list.
+  -- |
+  -- | `"try" : ["purescript", "today"] == ["try", "purescript", "today"]`
   (:) :: forall a. a -> [a] -> [a]
   (:) = cons
 
@@ -94,6 +113,9 @@ module Prelude
     }
     """ :: forall a. a -> [a] -> [a]
 
+  -- | Converts data to a String.
+  -- |
+  -- | `show 1 == "1"`
   class Show a where
     show :: a -> String
 
@@ -104,12 +126,15 @@ module Prelude
     }
     """ :: String -> String
 
+  -- | `show (Unit {}) == "Unit {}"`
   instance showUnit :: Show Unit where
     show (Unit {}) = "Unit {}"
 
+  -- | `show "abc" == "abc"`
   instance showString :: Show String where
     show = showStringImpl
 
+  -- | `show true == "true"`
   instance showBoolean :: Show Boolean where
     show true = "true"
     show false = "false"
@@ -121,6 +146,7 @@ module Prelude
     }
     """ :: Number -> String
 
+  -- | `show 1 == "1"`
   instance showNumber :: Show Number where
     show = showNumberImpl
 
@@ -137,34 +163,56 @@ module Prelude
     }
     """ :: forall a. (a -> String) -> [a] -> String
 
+  -- | `show [1, 2, 3] == "[1, 2, 3]"`
   instance showArray :: (Show a) => Show [a] where
     show = showArrayImpl show
 
   infixl 4 <$>
   infixl 1 <#>
 
+  -- | Apply a function to the data contained in a Functor type, returning a
+  -- | new wrapped result.
+  -- | 
+  -- | `not <$> Just true == Just false`
   class Functor f where
     (<$>) :: forall a b. (a -> b) -> f a -> f b
 
+  -- | Reversed version of (<$>).
+  -- |
+  -- | `Just true <#> not == Just false`
   (<#>) :: forall f a b. (Functor f) => f a -> (a -> b) -> f b
   (<#>) fa f = f <$> fa
 
+  -- | Replace the contents of a Functor with Unit.
+  -- |
+  -- | `void (Just false) == Just unit`
   void :: forall f a. (Functor f) => f a -> f Unit
   void fa = const unit <$> fa
 
   infixl 4 <*>
 
+  -- | Pass arguments into a Functor-wrapped function, returning a new wrapped
+  -- | result.
+  -- |
+  -- | `Just (+) <*> Just 5 <*> Nothing == Nothing`
   class (Functor f) <= Apply f where
     (<*>) :: forall a b. f (a -> b) -> f a -> f b
 
   class (Apply f) <= Applicative f where
     pure :: forall a. a -> f a
 
+  -- | Apply a function to the data contained in an Applicative type, returning
+  -- | a new Applicative result.
+  -- |
+  -- | `liftA1 not (Just true) == Just false`
   liftA1 :: forall f a b. (Applicative f) => (a -> b) -> f a -> f b
   liftA1 f a = pure f <*> a
 
   infixl 1 >>=
 
+  -- | Compose monadic actions together, passing the result of one as the 
+  -- | input to the next.
+  -- |
   class (Apply m) <= Bind m where
     (>>=) :: forall a b. m a -> (a -> m b) -> m b
 
@@ -173,6 +221,8 @@ module Prelude
   return :: forall m a. (Monad m) => a -> m a
   return = pure
 
+  -- | Apply a function to the data in a Monad type, returning a new Monad 
+  -- | result.
   liftM1 :: forall m a b. (Monad m) => (a -> b) -> m a -> m b
   liftM1 f a = do
     a' <- a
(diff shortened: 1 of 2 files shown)


prelude/prelude.purs
```diff
@@ -74,14 +84,23 @@ module Prelude
...
+  -- |
+  -- | `"str" ++ "ing" # reverse == "gnirts"`
   (#) :: forall a b. a -> (a -> b) -> b
   (#) x f = f x
 
   infixr 6 :
 
+  -- | Prepends an element to a list.
+  -- |
+  -- | `"try" : ["purescript", "today"] == ["try", "purescript", "today"]`
```
paf31: I think it's worth adding something to the effect of "Running time is `O(n)`" here.

prelude/prelude.purs
```diff
@@ -913,6 +1056,8 @@ module Control.Monad.Eff where
 
   type Pure a = forall e. Eff e a
 
+  -- | Run an effectful computation with no side effects, either because there 
+  -- | are none or because all possible side effects are handled.
```
paf31: I don't think this is a case of "either". The key is that the argument to `runPure` can be run in any context, because it doesn't require any fixed effects.

paf31:
This is great, thank you very much. I can't merge because of conflicts but even so, I would like to wait until the `--comments` PR is merged, so this doesn't result in an explosion of generated code.

I added some notes as well. I think it would be nice to add some comments related to class laws, and to talk about classes rather than class-members.

fluffynukeit:
No problem.  I'm happy to change wording to whatever, and to that end some guiding examples would be helpful.  I can fix up the PR when you think you're ready to take it.

prelude/prelude.purs
```diff
@@ -34,10 +34,14 @@ module Prelude
   otherwise = true
 
   -- | Flips the order of the arguments to a function of two arguments. 
+  -- |
+  -- | `flip (/) 6 3 == 3 / 6`
```
michaelficarra: It would render better if you instead indented these by four spaces.

garyb: Or perhaps:

``` purescript
flip (/) 6 3 == 3 / 6
```

prelude/prelude.purs
michaelficarra: In the future, this may be generalised so that it also applies to other data structures, with different complexity. Do we really want to add that kind of info to the documentation?

prelude/prelude.purs
```diff
@@ -104,12 +126,15 @@ module Prelude
     }
     """ :: String -> String
 
+  -- | `show (Unit {}) == "Unit {}"`
   instance showUnit :: Show Unit where
     show (Unit {}) = "Unit {}"
 
+  -- | `show "abc" == "abc"`
```
michaelficarra: You forgot the delimiters.

prelude/prelude.purs
```diff
@@ -137,34 +163,56 @@ module Prelude
...
   infixl 1 <#>
 
+  -- | Apply a function to the data contained in a Functor type, returning a
+  -- | new wrapped result.
+  -- | 
+  -- | `not <$> Just true == Just false`
   class Functor f where
     (<$>) :: forall a b. (a -> b) -> f a -> f b
 
+  -- | Reversed version of (<$>).
```
michaelficarra: Flipped.

prelude/prelude.purs
```diff
@@ -281,6 +333,7 @@ module Prelude
   infix 4 ==
   infix 4 /=
 
+  -- | Determine whether two data of the same type are equal.
```
michaelficarra: Two values or two inhabitants.

prelude/prelude.purs
```diff
@@ -352,11 +422,17 @@ module Prelude
     show GT = "GT"
     show EQ = "EQ"
 
+  -- | Compare the ranking of two data of the same type.
```
michaelficarra: Is this documentation for `compare` or `Ord`? If the former, I again prefer "values" or "inhabitants". If the latter, I would say "define a total ordering for a type".

prelude/prelude.purs
```diff
@@ -371,13 +450,21 @@ module Prelude
 
   infixl 4 <=
 
+  -- | Returns true if the left argument is "greater than or equal to" the 
+  -- | right argument.
```
michaelficarra: You have `<=` and `>=` documentation swapped.

prelude/prelude.purs
```diff
@@ -401,21 +488,41 @@ module Prelude
...
+  -- | Compare two Booleans.  True is greater than false.
+  -- |
+  -- | `compare false true == LT`
   instance ordBoolean :: Ord Boolean where
     compare false false = EQ
     compare false true  = LT
     compare true  true  = EQ
     compare true  false = GT
 
+  -- | Compare the value of two Numbers.
```
michaelficarra: Just "compare two numbers".

prelude/prelude.purs
```diff
@@ -401,21 +488,41 @@ module Prelude
...
+  -- |
+  -- | `compare 1 100 == LT`
   instance ordNumber :: Ord Number where
     compare = unsafeCompare
 
+  -- | Compare Strings ranked alphabetically (by character value).
```
michaelficarra: Also lexicographically.

prelude/prelude.purs
```diff
@@ -401,21 +488,41 @@ module Prelude
...
+  -- | 
+  -- | `compare "alphabetically" "bob" == LT`
   instance ordString :: Ord String where
     compare = unsafeCompare
 
+  -- | Compare the contents of two Arrays of the same type.  The first element
```
michaelficarra: Again, you can just mention lexicographic ordering.

prelude/prelude.purs
```diff
@@ -547,6 +660,7 @@ module Prelude
 
   infixr 5 <>
 
+  -- | Concatenation-like operation.
```
michaelficarra: I don't think this adds anything. If anything, it would confuse those unfamiliar with it.

prelude/prelude.purs
```diff
@@ -570,16 +687,27 @@ module Prelude
...
+  -- | function.
+  -- |
+  -- | `on (*) length "argument1" "arg2" == 9 * 4`
   on :: forall a b c. (b -> b -> c) -> (a -> b) -> a -> a -> c
   on f g x y = g x `f` g y
 
+  -- | Function of 0 arguments using Javascript calling conventions.
   foreign import data Fn0 :: * -> *
+  -- | Function of 1 arguments using Javascript calling conventions.
   foreign import data Fn1 :: * -> * -> *
```
michaelficarra: 1 argument.

prelude/prelude.purs
```diff
@@ -608,6 +739,7 @@ module Data.Function where
     }
     """ :: forall a b. (a -> b) -> Fn1 a b
 
+  -- | Convert a purescript function of 2 argument to JS calling convention.
```
michaelficarra: 2 arguments.

prelude/prelude.purs
fluffynukeit: What I found when typing it up is that using the four-spaces method made it really easy to accidentally combine the examples with the function type signature in the rendered MD.  The backticks were a way to handle that in a visually unambiguous way.  I prefer @garyb suggestion as a good alternative to try.

michaelficarra: @garyb: You can indent it by four spaces:

    ```purescript
    flip (/) 6 3 == 3 / 6
    ```

fluffynukeit:
Hey guys, I think this issue is ripe for [bikeshedding](http://en.wikipedia.org/wiki/Parkinson%27s_law_of_triviality).  Since Phil wants to hold off on bringing in this PR for a bit, let's use this space to discuss the desired format and/or style for the docs to come up with some "doc guidelines" that we think would work (either for core libraries or all purescript libraries, whatever scope we want to adopt).  Then I can go back and make the suggested fixes in accordance with the guidelines, most of which I agree with.  For better or worse, the prelude doc will probably influence the doc style for PS libraries everywhere, whether we have a de jur set of guidelines or not.

Possible points of interest

1.  How should examples be formatted?  I used single backticks in the PR.  @garyb suggested instead using backtick fences with explicit purescript syntax highlighting, which sounds ok to me.  I personally prefer usage of backticks to spaces because they are easy to identify visually.  However, they are not standard markdown, and psc-docs is already using spaces for the function signatures.  I wouldn't mind migrating everything to backticks, including the MD generated by psc-docs, because GFM is pretty ubiquitous for coders.  It might even be helpful if psc-docs did this automatically for commented examples, freeing the author from writing "``` purescript" over and over.  It could use a `EX:` tag or something.
2.  What style should examples take?  I obvious like the style of example invocation == result or some other logic statement that evaluates to true because it's short, informative, and allows every function to get an example without the size of the docs ballooning too much.
3.  To what audience level are we writing?  I like the occasional "useful for" notes.  But, at the same time, the prelude has some class instances that I've never used and have no experience in what they would be useful for (I skipped over some classes for this reason).  I find it a little weird that these rarely used instances are included in the prelude with lot of basic operation kind of stuff.  That comment's more about prelude content than audience level for doc, but I think the two are related.
4.  What differences would we see for classes and instances?  @paf31 mentioned a couple ideas above.  I have no opinions on this.

michaelficarra:
0. I'm for four-spaces. It stands out visually in the comment, has a different purpose than the single-backtick inline code examples, and we can compile it to use triple-backticks with explicit purescript highlighting as you suggest.
0. That seems fine.
0. That's tricky. I would write it for the most beginner audience I could without sacrificing correctness. And of course, as in Wikipedia, we should not explain anything that can be looked up elsewhere in the documentation.
0. I think @paf31 was saying that he would like the comment to be about what the class itself represents and the laws that its members obey, not a description of each member's purpose or a particular instance.

garyb:
I have no strong preferences about any of this stuff, as long as we come up with something consistent :). Was just throwing out the code fences as an potential alternative.

Re: 3. and uncommon instances, they are probably there because we really like to avoid orphan instances (and might even disallow them entirely at some point), so they kind of have to live in the prelude because they involve `Prim` types and `Prelude` is where the classes are defined.

natefaubion:
One thing I think we need to be careful about is the conflation of Lists and Arrays. This has been a source of a lot of confusion, and I really wish we hadn't pretended that Arrays are like Lists at all. This is something that has even leaked into the APIs of core libraries (e.g. `Data.Map.toList` and `Data.Map.fromList`). So when documenting things that work with Arrays, can we make sure to not call them Lists?

garyb:
Excellent point!

prelude/prelude.purs
hdgarrood: How about "an associative operation"? If my understanding of what a semigroup is is correct, that's everything we can possibly say about `<>`.

paf31: :+1:

hdgarrood: Re laws, I think this typeclass only has associativity: `(a <> b) <> c == a <> (b <> c)`

paf31: Yes, and I think it's good to spell it out in code like that in each class.

prelude/prelude.purs
```diff
@@ -570,16 +687,27 @@ module Prelude
...
+  -- | `[1, 2, 3] ++ [4, 5] == [1, 2, 3, 4, 5]`
   (++) :: forall s. (Semigroup s) => s -> s -> s
   (++) = (<>)
 
 module Data.Function where
 
+  -- | Apply a function to two arguments, then combine the results with a second
+  -- | function.
+  -- |
+  -- | `on (*) length "argument1" "arg2" == 9 * 4`
```
hdgarrood: I always thought this function was intended to be used infix and point-free - maybe something like this might be a bit more natural?
```
let f = compare `on` length
in f "argument1" "arg2"
```

garyb: That's the way I've generally used it also.

paf31:
Closing as duplicate.

PR #877, opened 2015-02-17
Add named fields to data constructor representation

garyb:
This is part of the work towards the revised typeclass representation, but the generated code should be identical to the current compiler output at the moment. I'd like to merge this now as it's relevant to `CoreImp` as it changes the constructor representation from just having arity.

The thing that still needs doing for the typeclass representation is to change the class constructor functions generated in `CoreFn` desugaring into data constructors instead, and `TypeClassDictionaryAccessor` needs desugaring into a pattern match rather than an object accessor.


src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -180,26 +180,23 @@ valueToJs _ (Constructor (_, _, _, Just IsNewtype) _ (ProperName ctor) _) =
...
-          body = [ JSAssignment (JSAccessor arg (JSVar "this")) (JSVar arg) | arg <- args ]
-      in JSFunction (Just ctorName) args (JSBlock body)
-    go :: String -> Int -> Int -> [JS] -> JS
-    go pn _ 0 values = JSUnary JSNew $ JSApp (JSVar pn) (reverse values)
-    go pn index n values =
-      JSFunction Nothing ["value" ++ show index]
-        (JSBlock [JSReturn (go pn (index + 1) (n - 1) (JSVar ("value" ++ show index) : values))])
-
+    makeConstructor :: String -> [Ident] -> JS
+    makeConstructor ctorName fs =
```
paf31: `fields` should be in scope in the body of these functions. Doesn't really matter, but you could trim things down by omitting `fs`.

src/Language/PureScript/CoreFn/Desugar.hs
```diff
@@ -134,61 +135,64 @@ mkTypeClassConstructor ss com supers members =
 -- |
 -- Desugars expressions from AST to CoreFn representation.
 --
-exprToCoreFn :: Environment -> Maybe SourceSpan -> [Comment] -> Maybe Type -> A.Expr -> Expr Ann
-exprToCoreFn _ ss com ty (A.NumericLiteral v) =
+exprToCoreFn :: ModuleName -> Environment -> Maybe SourceSpan -> [Comment] -> Maybe Type -> A.Expr -> Expr Ann
```
paf31: Again, not essential, but I see a lot of these types of things in the codebase generally. I think we should try to close over constant things like this using `where` clauses to tidy things up.

paf31:
Looks good. What `fields` get used for a type class dictionary constructor now? The names of the members?

src/Language/PureScript/CoreFn/Desugar.hs
garyb: Yeah, that makes sense.

garyb:
Type class dictionary constructors don't make use of the new stuff yet, they're still functions that produce records and there's a case in codegen where the metadata is examined to produce the JS prototypical class constructor instead. But yep, the intention is to replace all that fragile stuff with a data constructor where fields are named after the members and superclasses.

The generated code should still be identical after the change over is complete, but the implementation will be simpler and better contained.

paf31:
Otherwise, looks good to me, if everything passes. :ship:

paf31:
Thanks!

garyb:
I'll keep an eye out for those unnecessary argument situations in the future, as you're right, there's a lot of that stuff we can probably use `where`s to avoid. :)

Issue #879, opened 2015-02-17
Support declarations split over multiple commands in psci


paf31:
The order of declarations are reversed in the generated module. I have to say:

```text
> let f n = f (n - 1) * n
> let f 0 = 1
> f 10
3628800
```

As far as I can tell, this should be as simple as reversing the arguments to `++` in `updateLets`.

PR #880, opened 2015-02-17
Support declarations split over multiple commands in psci #879

paf31:
@garyb This is pretty straightforward I think, but can you please review?


garyb:
Looks good to me :+1:

The ability to split declarations over multiple `let`s in psci is pretty cool, I saw you accidentally discover that in the meetup video too :)

PR #884, opened 2015-02-18
A simple attempt at operator aliases, fix #806

### LINKED ISSUE #806: Aliases for operators
Suggested by @jdegoes

We could write e.g. (suggested syntax, open to change):

```purescript
infixl 4 <$> as fmap
```

and then use `fmap` instead of `(<$>)` in code. Also `<$>` would be codegen'd as `fmap`, instead of having to generate a name.

Should this be mandatory even?

paf31:
This doesn't implement the spec that I wrote out in #806, but it allows us to give aliases for top-level fixity declarations, which is a start. They are neither required, nor supported in `where` clauses.

Fully implementing the spec laid out in the issue could be quite difficult, so I'm going to try to do it in stages.

I also tidied up the `Operators` module a bit.

@garyb @joneshf Thoughts?

### DIFF
--- a/examples/passing/OperatorAlias.purs
+++ b/examples/passing/OperatorAlias.purs
@@ -1,0 +1,17 @@
+module Main where
+
+data Maybe a = Nothing | Just a
+
+infixl 4 ? as orElse
+
+(?) :: forall a. Maybe a -> a -> a
+(?) Nothing a = a
+(?) (Just a) _ = a
+
+test1 x = x ? 0
+
+test2 x = x ? 0
+  where
+  (?) _ y = y
+
+main = Debug.Trace.trace "Done"
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -180,9 +180,9 @@ data Declaration
   --
   | ExternInstanceDeclaration Ident [Constraint] (Qualified ProperName) [Type]
   -- |
-  -- A fixity declaration (fixity data, operator name)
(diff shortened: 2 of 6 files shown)


joneshf:
I apologize, I haven't really been following the conversation. I'll try and go through it tomorrow or so.

garyb:
Is this the way round that we're doing it then? Operators have names as aliases, rather than named functions can have operator aliases defined?

jdegoes:
Yeah, I'd suggest the syntax be: `infixl 4 orElse as ?`, because it gives a fixity and operator alias to the already-existing function `orElse`.

michaelficarra:
Agreed. That makes the fixity declaration optional.

paf31:
I just want something with a clear upgrade path. Breaking every library with an operator in it doesn't seem like a good option to me, so I'd like to do things in stages if possible. I completely agree that having no operators in the final core AST is the best situation we can get to, it's just a question of how to get there. 

This was just intended as a patch, so that people can start adding aliases now, and maybe in 0.7.0 we can make them required. Ideally, we would emit a warning if an alias was not present.

That said, if we decide operators-as-aliases-for-names is better, then that's a different set of changes entirely.

I'm happy to close this, it was mostly just for my experimentation.

jdegoes:
Since the parser is backtracking, why can't a syntax like `infixl 4 orElse as ?` be backward-compatible?

I do think allowing a version for people to upgrade all their aliases would be really nice.

paf31:
No, it's definitely possible. I'm just saying that whatever way we pick, I'd like to have a version (or set of versions) to transition people over, with warnings.

PR #885, opened 2015-02-18
Added semigroupOrdering instance

pseudonom:
It's pretty useful for sorting by multiple criteria.


joneshf:
Is the failure showing that type classes must be defined before instances can use them?

Has this always been the case and we've just never run into it until now?

garyb:
Wow, yes, that certainly seems to be the case!

paf31:
:+1:

garyb:
This is now merge-able, as of #903.

paf31:
Thanks!

PR #892, opened 2015-02-19
Switch to ExceptT and transformers/mtl 0.4.*

hdgarrood:
I looked through the code for every use of `guard`, and I don't think we need to do anything about any of them:

* `Parser.Lexer` - it's the `Parsec` monad
* `Parser.Common` - it's the `Parsec` monad again
* `TypeChecker.Entailment` - it's the `Check` monad, which uses an `Either ErrorStack`
* `Sugar.Operators` - it's the `Parsec` monad again
* `psci:Main` - it's the IO monad
* `psc-make:Main` it's the IO monad

This compiles and all the tests pass on my machine. Is that everything, then?


paf31:
Well, that's embarrassing. Are there no warnings emitted any more? I think I would like to also remove the `Error` constraint on things like `lookupVariable`, but we could deal with that separately.

hdgarrood:
Ok wait a minute, it appears I don't know how to use `grep` properly. I'll check again and get back to you.

paf31:
FYI: If you turn deprecation warnings back on in the Cabal file, you should see everywhere we use `Error` and `ErrorT`.

hdgarrood:
It was bash. Should have guessed, probably.

I'd only missed one `guard` in `Parser.Common`, I've just added that to the issue description. It was the `Parsec` monad.

Yep - no warnings when I `cabal build` now. I've just realised that it still imports `Control.Monad.Error` everywhere - I'll change those to `Control.Monad.Except`

The following functions have `Error` constraints:

* `TypeChecker.Monad`:
  * `lookupVariable`
  * `lookupTypeVariable`
  * `{get,check}Visibility`
* `TypeChecker.Synonyms`
  * `PreplaceAll,expand,expandAll}TypeSynonyms`

All of these occur in `Language.PureScript.TypeChecker` or submodules, and so I'm pretty sure that means that they all end up only using `e ~ ErrorStack`. I see three potential solutions:

* Lose generality by adding `e ~ ErrorStack` in the constraints
* Switch the instance `Error e` for a new argument of type `String -> e` (preserves generality, but feels a bit yucky)
* Do #745 first and then re-evaluate options

paf31:
Did you remove the deprecation warning flag from `purescript.cabal`? If so, that's great that there are no warnings! I thought `Error` gave some, but we can see..

My vote would be to pass a function `String -> e` if we really need the generality (I don't think we do), or to specialize to `ErrorStack`. #745 will be somewhat orthogonal as the error ADT will appear inside the definition of `ErrorStack` replacing the string which is currently there.

hdgarrood:
No, all the build targets in the cabal file still have `fno-warn-warnings-deprecations`. I haven't tried removing those. I'll try getting rid of the `Error` constraints first and then see what happens if I do.

hdgarrood:
`ghc-options: -Wall -fno-warn-warnings-deprecations` means all warnings except deprecations, right?

paf31:
Correct, it's only there because of the `ExceptT` issue.

hdgarrood:
There's a couple of `strMsg` calls, and therefore `Error` constraints, in `monad-unify`; if you're planning in on including it in this repo can we do that now? I think it would make it easier to remove uses of `Error`.

paf31:
I'm happy to move `monad-unify` back inside `purescript`. Would you like to move it over as part of this PR?

hdgarrood:
Ok, will do.

hdgarrood:
See #896

PR #896, opened 2015-02-20
Switch to ExceptT and transformers/mtl 0.4.*

hdgarrood:
Continues from #892, this is the same but with a cleaner history.

### DIFF
--- a/psc-make/Main.hs
+++ b/psc-make/Main.hs
@@ -17,7 +17,7 @@
 module Main where
 
 import Control.Applicative
-import Control.Monad.Error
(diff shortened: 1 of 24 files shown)


paf31:
@hdgarrood This looks good. The one remaining issue, as we discussed on IRC, is the invalid `MonadPlus` instance.

I compiled this, but removing the `Monoid ErrorStack` instance caused some issues. I think the fix is simple - we just need to move `mappend` into a function and convert all uses of `<>` to the new function - but there are quite a few places where it is used. I can work through it tomorrow.

PR #898, opened 2015-02-21
Extract MonadSupply class and mtl-ify the Sugar modules

paf31:
@joneshf @garyb This is just a bit of tidying I've wanted to do for a bit. Could you please review? Thanks.


garyb:
Looks nice :) :+1:

joneshf:
There needs to exist a code review tool where if the type signature changes but the implementations stays the same it hides that line. Sort of like `git diff -w`.

Of course, it'd only work well for certain languages.

PR #899, opened 2015-02-21
Psci node flags

### LINKED ISSUE #793: Pass node flags to psci
Since psci starts up a node process, there should be a way to delegate flags to the node process by invoking psci. I can see 2 ways of doing this:

1. Any unknown flags can just go to the node executable.
2. Any flags after a `--` go to the node executable.

I personally vote for option 2, as it is a more deliberate action from the user.

MichaelXavier:
This resolves #793


paf31:
Looks good to me, thanks! Any chance you could rebase the commits into one, please?

PR #903, opened 2015-02-22
Fix #886

### LINKED ISSUE #886: Class must be defined before instance in module
As discovered in #885
``` purescript
module Test where

  instance fooNumber :: Foo Number where
    foo x = x
    
  class Foo a where
    foo :: a -> a
```
Results in:
```
Error at test.purs line 3, column 3 - line 6, column 3:
Error in type class instance Test.Foo Prim.Number:
Type class Test.Foo is undefined
```
This is happening during typeclass desugaring.

paf31:
This just sorts declarations before desugaring in `TypeClasses.hs`.

@garyb @joneshf I think this one is fairly self-explanatory, but could you please review?

### DIFF
--- a/examples/passing/InstanceBeforeClass.purs
+++ b/examples/passing/InstanceBeforeClass.purs
@@ -1,0 +1,9 @@
+module Main where
+
+instance fooNumber :: Foo Number where
+  foo = 0
+
+class Foo a where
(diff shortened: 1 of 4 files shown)


garyb:
:shipit:

PR #904, opened 2015-02-22
Fix #902.

### LINKED ISSUE #902: Comments appear in generated code even without --comments
At the top of each module, there is a (sometimes empty) comment. It shouldn't be there unless the `--comments` flag is provided, and an empty comment shouldn't be generated either way.

/cc @joneshf

joneshf:
Let me know if I missed anything.


garyb:
:+1:

paf31:
Thanks!

PR #906, opened 2015-02-23
Use a data type for errors, and link to wiki in error messages

paf31:
@garyb @joneshf Could you please review?


src/Language/PureScript/Errors.hs
```diff
@@ -12,108 +12,378 @@
...
-  fromString = strMsg
+-- |
+-- Pretty print an ErrorMessage
+--
+prettyPrintErrorMessage :: ErrorMessage -> String
+prettyPrintErrorMessage e =
+  go e ++ "\n\nSee " ++ wikiUri ++ " for more information, or to contribute content related to this error."
+  where
+  wikiUri :: String
+  wikiUri = "http://wiki.purescript.org/Error-Code-" ++ errorCode e
```
michaelficarra: Is there any way we can use TLS? What's wrong with `https://github.com/purescript/purescript/wiki/Error-Code-`?

paf31: Sounds good.

paf31: Actually, the `wiki.purescript.org` link only redirects over HTTP. I'll need to take a look at the domain settings to see if I can do anything about that. Otherwise, I'll link to `github.com`.

paf31: Done.

paf31:
Ok, I think this is good to go. As a bonus, the coverage report now tells us exactly which error types aren't covered by a failing test. Maybe we could use the new error type to make stronger assertions about our failing tests actually.

garyb:
:shipit:

Issue #852, opened 2015-02-08
Document all Prelude functions


paf31:
The generated markdown docs now contain code comments, so we should start documenting all public functions in the core libraries. The Prelude would be a good place to start.

This involves a bit of work, but would be a fantastic contribution for any beginners.

michaelficarra:
Weren't we trying to reduce the Prelude as much as possible? If I was to take this on, I would want to know what the plans were for the Prelude first.

paf31:
Well, anything removed from Prelude would move into a core library anyway, and I would like to document those too eventually. The "Prelude" module itself is unlikely to go anywhere though.

cschneid:
Can you point to some example documentation? How do I format the docs, provide examples, links, etc.

paf31:
The plan is to parse comments as markdown if they are indented past a pipe:

```purescript
-- | Map a function over an array
-- |
-- | For example: `map (+ 1) [1, 2, 3]`
map :: forall a b. (a -> b) -> [a] -> [b]
```

PR #910, opened 2015-02-23
Added some documentation of typeclasses

japesinator:
This commit commit adds some documentation for Semigroupoid, Category, Functor, Apply, Applicative, Bind, and Monad, partially addressing #852


prelude/prelude.purs
```diff
@@ -61,6 +61,14 @@ module Prelude
   infixr 9 >>>
   infixr 9 <<<
 
+  -- | Semigroupoids are Categories, less `id`, which is to say they consist of objects and
```
paf31: You should be able to link to `Category` and get links in the generated Markdown like this:

    A `Semigroupoid` is similar to a [`Category`](#Category) but lacks `id`...

prelude/prelude.purs
```diff
@@ -61,6 +61,14 @@ module Prelude
...
+  -- | morphisms between them which can be composed.
+  -- |
+  -- | Semigroupoids should obey the following rule:
+  -- |
+  -- | Association
```
paf31: "Associativity"?

prelude/prelude.purs
```diff
@@ -70,6 +78,20 @@ module Prelude
   (>>>) :: forall a b c d. (Semigroupoid a) => a b c -> a c d -> a b d
   (>>>) f g = g <<< f
 
+  -- | Categories consist of objects and composable morphisms between them, and as such are
+  -- | semigroupoids, but unlike semigroupoids must have an identity element.
```
paf31: Again, we can link to `Semigroupoid`, and format it as code here.

prelude/prelude.purs
```diff
@@ -70,6 +78,20 @@ module Prelude
...
+  -- |
+  -- | Categories should obey the following rules.
+  -- |
+  -- | Association (as semigroupoids)
```
paf31: I would probably leave the `Semigroupoid` laws out since they are implied by the superclass relationship. It's not so bad here, but for something like `Monad`, listing all the superclass laws could get time-consuming, and this is a good time to establish a default.

prelude/prelude.purs
```diff
@@ -159,9 +192,34 @@ module Prelude
 
   infixl 4 <*>
 
+  -- | Applys are intuitively Applicatives less `pure`, and more formally a strong lax semi-
```
paf31: This could possibly put some beginners off, while being technically correct (I _think_ :smile:)

The way I explain this in the book is that `Apply` allows us to lift functions of `n > 1` arguments over a the type constructor `f`, i.e. to take functions `a -> b -> ...` to functions `f a -> f b -> f c`. `Functor` and `Applicative` then give the `n=0` and `n=1` cases respectively.

Not sure what's best, this is just my 2¢.

prelude/prelude.purs
```diff
@@ -170,9 +228,30 @@ module Prelude
...
+  -- |
   class (Apply m) <= Bind m where
     (>>=) :: forall a b. m a -> (a -> m b) -> m b
 
+  -- | Monad is a class which can be intuitively thought of as an abstract datatype of actions or
+  -- | more formally though of as a monoid in the category of endofunctors.
+  -- |
+  -- | A Monad should obey the following rules.
+  -- |
+  -- | Associative (as Bind)
```
paf31: I wonder if it's worth writing these out in `do` notation as well. I always found them easier to understand that way.

paf31:
This looks really great, thank you for taking the time to put this together. Sorry for the picky comments, I just think that now is a good time to establish some standard commenting practices, like links and formatting.

prelude/prelude.purs
paf31: Another thought is that we can always link to the online book in comments for more details if the user wants them.

garyb:
There's also the stuff I started in `Maybe`, https://github.com/purescript/purescript-maybe that might be a little useful? It also goes into `Apply` a bit and a bunch of other stuff that is probably unnecessary.

japesinator:
I fixed most of the things you pointed out, but at the moment Apply is still unchanged, as I wasn't sure the best way to do that, and I have left the Monad laws without do notation, as I was unsure how to best format that. If anyone has suggestions as to how to implement either of these, I'm happy to add them

paf31:
Sounds good. Just one more thing: could you please run `prelude/updateDocs.sh` and regenerate the Markdown and make sure the links work?

Thanks.

paf31:
Merged, thank you!

We can always continue the `Monad` debate and update this again later if need be.

PR #913, opened 2015-02-24
Docs for some basic functions

japesinator:
This PR adds documentation and/or examples for `flip`, `const`, `(#)`, `($)`, and `(:)`

### DIFF
--- a/prelude/README.md
+++ b/prelude/README.md
@@ -13,8 +13,8 @@ otherwise :: Boolean
 An alias for `true`, which can be useful in guard clauses:
 
 ```purescript
-    max x y | x >= y = x
-            | otherwise = y
+max x y | x >= y = x
+        | otherwise = y
 ```
 
 
@@ -27,7 +27,7 @@ flip :: forall a b c. (a -> b -> c) -> b -> a -> c
 Flips the order of the arguments to a function of two arguments.
 
 ```purescript
-    flip const 1 2 = const 2 1 = 2
+flip const 1 2 = const 2 1 = 2
 ```
 
 
@@ -40,7 +40,7 @@ const :: forall a b. a -> b -> a
 Returns its first argument and ignores its second.
 
 ```purescript
-    const 1 "hello" = 1
+const 1 "hello" = 1
 ```
 
 
@@ -54,7 +54,7 @@ This function returns its first argument, and can be used to assert type equalit
 This can be useful when types are otherwise ambiguous.
 
 ```purescript
-    main = print $ [] `asTypeOf` [0]
+main = print $ [] `asTypeOf` [0]
 ```
 
 If instead, we had written `main = print []`, the type of the argument `[]` would have
@@ -125,12 +125,12 @@ instance categoryArr :: Category Prim.Function
 Applies a function to its argument
 
 ```purescript
-    (+) 1 ((*) 2 (const 1 2)) = (+) 1 $ (*) 2 $ const 1 2
+length $ groupBy productCategory $ filter isInStock products =
+length (groupby productcategory (filter isinstock (products)))
 ```
 
-`($)` is different from [`(#)`](#-2) because it is right-infix instead of left. Right-infix
-operators parse like (a $ (b $ (c $ (d ...)))) whereas left-infix operators parse like
-`(((a # b) # c) # d ...)`.
+`($)` is different from [`(#)`](#-2) because it is right-infix instead of left, so
+`a $ b $ c $ d x` = `a (b (c (d x)))`
 
 
 #### `(#)`
@@ -142,12 +142,12 @@ operators parse like (a $ (b $ (c $ (d ...)))) whereas left-infix operators pars
 Applies a function to its argument
 
 ```purescript
-    (((3 + 1) * 2) + 4) = 3 # ((+) 1) # ((*) 2) # ((+) 4)
+products # groupby productcategory # filter isinstock # length =
+length (groupby productcategory (filter isinstock (products)))
 ```
 
-`(#)` is different from [`($)`](#-1) because it is left-infix instead of right. Right-infix
-operators parse like (a $ (b $ (c $ (d ...)))) whereas left-infix operators parse like
-`(((a # b) # c) # d ...)`.
+`(#)` is different from [`($)`](#-1) because it is left-infix instead of right, so
+`x # a # b # c # d` = `(((x a) b) c) d`
 
 
 #### `(:)`
@@ -159,7 +159,7 @@ operators parse like (a $ (b $ (c $ (d ...)))) whereas left-infix operators pars
 Attaches an element to the front of a list.
 
 ```purescript
-    1 : [2, 3, 4] = [1, 2, 3, 4]
+1 : [2, 3, 4] = [1, 2, 3, 4]
 ```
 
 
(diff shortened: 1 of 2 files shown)


prelude/README.md
```diff
@@ -25,6 +25,11 @@ flip :: forall a b c. (a -> b -> c) -> b -> a -> c
 
 Flips the order of the arguments to a function of two arguments.
 
+E.g.
```
michaelficarra: All of these e.g. are unnecessary.

prelude/README.md
```diff
@@ -33,6 +38,11 @@ const :: forall a b. a -> b -> a
 
 Returns its first argument and ignores its second.
 
+E.g.
+
+    const 1 2 = 1
```
michaelficarra: Maybe the example should use values of two different types.

prelude/README.md
```diff
@@ -111,6 +121,16 @@ instance categoryArr :: Category Prim.Function
 ($) :: forall a b. (a -> b) -> a -> b
 ```
 
+Applies a function to its argument
+
+E.g.
+
+    const $ 1 2 = const 1 2 = 1
+
+`($)` is different from [`(#)`](#-2) because it is right-infix instance instead of left.
```
michaelficarra: Looks like "instance" was accidental and should be removed.

garyb:
Can we use fenced code sections:

    ``` purescript 
    -- code
    ```

for the examples please?

prelude/README.md
```diff
@@ -111,6 +121,16 @@ instance categoryArr :: Category Prim.Function
 ($) :: forall a b. (a -> b) -> a -> b
 ```
 
+Applies a function to its argument
+
+E.g.
+
+    const $ 1 2 = const 1 2 = 1
```
michaelficarra: This example should use a more realistic use case, since here it would be better not to use this function at all.

prelude/README.md
```diff
@@ -118,6 +138,16 @@ instance categoryArr :: Category Prim.Function
 (#) :: forall a b. a -> (a -> b) -> b
 ```
 
+Applies a function to its argument
+
+E.g.
+
+    3 # ((+) 1) = 3 + 1 = 4
```
michaelficarra: Same here. This example adds nothing.

michaelficarra:
@garyb can't we just have the generator add those for us? All code should be PureScript, no?

garyb:
In an ideal world yeah, unfortunately the markdown is not processed at all currently, it's just dropped straight into the generated file. Phil couldn't find a suitable markdown parser that would also do pretty printing.

michaelficarra:
So do we go back through and delete all the backticks once one is found/created?

garyb:
Sorry, I'm not sure I follow what you're referring to with that?

japesinator:
Alright, I think I took care of everything you pointed out. Thanks for the feedback!

paf31:
Well, there's no guarantee that all the code sections will contain PureScript. JavaScript and plain text are two very possible alternatives, for example.

In `psc-pages`, I use the `cheapskate` library, and I can do traversals of the markdown AST, so I suggest we format things how we want them to appear on GitHub for now.

paf31:
Also, the indentation is incorrect inside the code fences.

prelude/README.md
```diff
@@ -111,6 +122,16 @@ instance categoryArr :: Category Prim.Function
 ($) :: forall a b. (a -> b) -> a -> b
 ```
 
+Applies a function to its argument
+
+```purescript
+    (+) 1 ((*) 2 (const 1 2)) = (+) 1 $ (*) 2 $ const 1 2
```
paf31: Something using composition might make a nice example. This seems a bit dense for a beginner due to all the sections. What about something list-based like:

```purescript
length $ groupBy productCategory $ filter isInStock products
```

prelude/README.md
```diff
@@ -111,6 +122,16 @@ instance categoryArr :: Category Prim.Function
...
+```
+
+`($)` is different from [`(#)`](#-2) because it is right-infix instead of left. Right-infix
```
paf31: I don't know if showing the parse would help a beginner, and they can get this info from the wiki anyway. I think the key is that you can rewrite the nested application `f (g (h x))` as `f $ g $ h x` or as `f <<< g <<< h $ x` or as `x # h >>> g >>> f`.

paf31:
I'm a little late on this one, but please see my comments. Looks nice generally though, and thank you for taking the time to work on this.

prelude/README.md
```diff
@@ -118,6 +139,16 @@ instance categoryArr :: Category Prim.Function
 (#) :: forall a b. a -> (a -> b) -> b
 ```
 
+Applies a function to its argument
+
+```purescript
+    (((3 + 1) * 2) + 4) = 3 # ((+) 1) # ((*) 2) # ((+) 4)
```
michaelficarra: Like above, we should not use sections in these examples, and they should be more understandable. Would you want to read documentation like this?

prelude/README.md
japesinator: I was trying to make sure the examples ran with only the Prelude. Is it kosher to do otherwise?

paf31: I think it's fine to take examples from the core libraries, or just to make up functions from a standard domain like product/customer/category or something.

prelude/README.md
```diff
@@ -111,6 +122,16 @@ instance categoryArr :: Category Prim.Function
 ($) :: forall a b. (a -> b) -> a -> b
 ```
 
+Applies a function to its argument
+
+```purescript
+length $ groupBy productCategory $ filter isInStock products =
+length (groupby productcategory (filter isinstock (products)))
```
michaelficarra: `groupBy`, `productCategory`, `isInStock`

prelude/README.md
```diff
@@ -118,6 +139,16 @@ instance categoryArr :: Category Prim.Function
 (#) :: forall a b. a -> (a -> b) -> b
 ```
 
+Applies a function to its argument
+
+```purescript
+products # groupby productcategory # filter isinstock # length =
+length (groupby productcategory (filter isinstock (products)))
```
michaelficarra: `groupBy`, `productCategory`, `isInStock`

prelude/README.md
paf31: Also, sorry to be pedantic, but this isn't valid PureScript. Can we please separate this into two code sections, and say something like "which is equivalent to:" in between them?

japesinator: Alright, that should be done now

prelude/prelude.purs
```diff
@@ -98,14 +109,42 @@ module Prelude
   infixr 0 $
   infixl 0 #
 
+  -- | Applies a function to its argument
+  -- |
+  -- | ```purescript
```
paf31: Like this:

    ```purescript
    length $ groupBy productCategory $ filter isInStock products
    ```
   
    is equivalent to
    
    ```purescript
    length (groupBy productCategory (filter isInStock (products)))
    ```

Also, you need to regenerate the docs.

Thanks.

paf31:
Thanks! And sorry for all the back-and-forth, but I think it's good to establish some conventions now.

michaelficarra:
Should've squashed first. Seven commits per feature makes things like `git bisect` infeasible.

PR #923, opened 2015-02-28
Restructure code with where clause

garyb:
Purely cosmetic really


paf31:
Very nice :)

garyb:
Good call on this adopting this style, it's much nicer to deal with!

PR #932, opened 2015-03-02
Prelude documentation

paf31:
Work on #852.

/cc @garyb


paf31:
I'm not sure how coverage increased :)

garyb:
Magic!

Looks great to me. Only comment is we sort of have a mixture of using `forall` and not in the laws - I'd been omitting it so far, as I think it's a little more pleasant to read without.

paf31:
Yep, makes sense.

paf31:
@garyb How's this?

garyb:
:+1:

michaelficarra:
We use "its arguments flipped" and "its arguments reversed" interchangeably. We should pick one and stick with it.

PR #933, opened 2015-03-02
Document the remaining Prelude modules, fix #852

### LINKED ISSUE #852: Document all Prelude functions

paf31:
@garyb @hdgarrood Could you please review?

### DIFF
--- a/prelude/README.md
+++ b/prelude/README.md
@@ -878,6 +878,14 @@ instance semigroupArr :: (Semigroup s') => Semigroup (s -> s')
 on :: forall a b c. (b -> b -> c) -> (a -> b) -> a -> a -> c
 ```
 
+The `on` function is used to change the domain of a binary operator.
+
+For example, given a function `height :: Person -> Number`, we can create a function which compares two people by height:
+
+```purescript
+taller :: (>) `on` height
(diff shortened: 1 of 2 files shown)


garyb:
Looks good!

Could perhaps include the ```compare `on` bar``` style example for `on` as well? Or do you think people will get that usage from the example you gave?

hdgarrood:
Looks good to me too. I feel like we only need one example of usage of `on`. I like this one - using `(>)` rather than `compare` means that people coming from languages that use `int` for comparisons instead of an actual `Ordering` type have fewer new concepts to get their heads around. I don't feel particularly strongly either way though.

garyb:
That makes sense.

paf31:
I agree that we only need one, but I'm going to change it to use `compare`, since I think it's more idiomatic.

paf31:
How does this look? I picked a slightly more interesting function for the RHS too.

hdgarrood:
:+1:

PR #936, opened 2015-03-03
Fix functor law in prelude documentation

hdgarrood:
I just realised the arguments to `<$>` were the wrong way around while trying to write a QuickCheck property to verify well-behaved-ness of a Functor instance.

### DIFF
--- a/prelude/README.md
+++ b/prelude/README.md
@@ -243,7 +243,7 @@ types use the type constructor `f` to represent some computational context.
 `Functor` instances should satisfy the following laws:
 
 - Identity: `(<$>) id = id`
-- Composition: `(<$>) (f <<< g) = (<$> f) <<< (<$> g)`
+- Composition: `(<$>) (f <<< g) = (f <$>) <<< (g <$>)`
 
 
 #### `(<#>)`
--- a/prelude/prelude.purs
+++ b/prelude/prelude.purs
@@ -229,7 +229,7 @@ module Prelude
   -- | `Functor` instances should satisfy the following laws:
   -- |
   -- | - Identity: `(<$>) id = id`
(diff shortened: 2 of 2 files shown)


garyb:
Oops, I think that was me.

Also, have you seen this? https://github.com/purescript/core-tests/blob/master/src/Test/Classes.purs Might already have what you need.

garyb:
I wrote those a long time ago so not sure if they still work actually...

hdgarrood:
I've found it's very easy to make mistakes like these with the infix `<$>`, possibly a reason to consider `fmap`?

garyb:
I think that'll happen as part of #806/#884, would probably just go with `map` too.

hdgarrood:
Cool. Also with Test.Classes, does `core-tests` get run frequently? Because if so, presumably you'd hear about it if Test.Classes stopped working?

garyb:
I'm not sure actually, I can't remember the last time I run it. Generally while working on the compiler I've just been running that test suite and then checking `test-everything` still compiles - `core-tests` was more supposed to be tests for the core libraries than it was for the compiler, however I don't think it's totally up to date now.

It came about as we wanted to avoid circular dependencies in `bower`, but as soon as you include `quickcheck` that becomes tricky as it uses a bunch of the core libraries in its implementation. I'd still much prefer the tests lived with their corresponding libraries, perhaps in a kind of sub-project of the libraries themselves, but never quite got around to figuring it out.

hdgarrood:
Ahh, ok then.

I just copied that file into my project, and it seems to work fine. In fact it just caught a non-compliant Applicative instance in my code!

garyb:
Good stuff! I should look at breaking those out into their own library really, and include all the steps between `Functor` and `Monad`, etc too.

Issue #352, opened 2014-04-05
CommonJS module generation in psc-make


garyb:
At least investigate whether it's workable for PureScript. I'm pretty sure it is, and would potentially be a much better default than the current export method. It would mean having to using Browserify or something to package up the result for client-side js though.

It would also help with `psc-make` as the individual generated js files aren't really usable in Node at the moment, due to how `_ps` is resolved.

bstro:
How is one supposed to use this functionality within an environment like Browserify? There are no relative paths, so it seems the environment must be made aware of the Prelude… its unclear to me how to make this happen. I have psc-make half-way working, but its building a directory full of CommonJS modules that don't know how to find each other.

joneshf:
The way I've found to make this work is to use the env variable `NODE_PATH`. `node` will search for modules on these paths. Some documentation is here: http://nodejs.org/api/modules.html#modules_loading_from_the_global_folders

So if you are just using the defaults you'll get modules in `output`, something like `output/Prelude/index.js` or so. If you set `process.env.NODE_PATH = path.resolve('output')` it will look for modules in `output` and resolve things properly.

Since we have modules as folders: http://nodejs.org/api/modules.html#modules_folders_as_modules it suffices to `require('Prelude')` and things will load properly.

N.B. modifying `process.env` is a hack. I can't remember where I read it, but apparently you're not supposed to go willy nilly messing with `process.env`. Given the situation, I just do it as there isn't much other option. Though, maybe someone has node module that does it properly.

bstro:
That worked! Seems fine for now…hope someone figures out how to do this properly.

garyb:
One option is to move/rename the `output` as `node_modules`. @paf31 took that approach here: https://github.com/purescript-contrib/grunt-init-purescript/blob/master/root/Gruntfile.js

PR #938, opened 2015-03-04
Make module imports relative.

paulyoung:
This pull request addresses the issue described in https://github.com/purescript/purescript/issues/352#issuecomment-68677435.

This means that modules can be `require`d without setting `process.env.NODE_PATH` or renaming the output directory to `node_modules`.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -80,7 +80,7 @@ moduleToJs (Module coms mn imps exps foreigns decls) = do
   importToJs mn' = do
     additional <- asks optionsAdditional
     let moduleBody = case additional of
-          MakeOptions -> JSApp (JSVar "require") [JSStringLiteral (runModuleName mn')]
+          MakeOptions -> JSApp (JSVar "require") [JSStringLiteral ("../" ++ runModuleName mn')]
           CompileOptions ns _ _ -> JSAccessor (moduleNameToJs mn') (JSVar ns)
     return $ JSVariableIntroduction (moduleNameToJs mn') (Just moduleBody)
 


garyb:
Seems like a good solution to me, I'm not sure why we didn't do this in the first place! :+1:

paf31:
I'm not sure I see the benefit of this, sorry. To be honest, I don't fully understand the issue yet :)

The way I understand it, CommonJS defines what the module loader should do, which is why I like that we do the minimum possible work when generating imports.

Maybe the relative path could be a command line option, but I'm not sure I like making `../` the default. Maybe I'm missing something.

paulyoung:
No need to apologize! I thought it was better to make the change and submit a pull request rather than open an issue for this one.

I think the benefit is that people can use `psc-make` and then `require` any of the generated modules (including the entry point) without running into `Cannot find module` errors.

It took me a while to find the linked issue and make things work otherwise and both workarounds aren't ideal, it seems.

garyb:
Hmm, I see your point @paf31, but yeah I think it would be nice to have as a command line option. Avoiding the copy etc. stage during compliation would be handy, as it seems to have tripped up a few people.

paulyoung:
Might want to consider making the [ES6 module format](https://babeljs.io/docs/learn-es6/#modules) an option too.

paf31:
I like the ES6 module idea, maybe as a separate option for now.

I guess what I don't understand is: why not just set `-o output/node_modules` or something?

paulyoung:
My concern was that a package in `node_modules/` might conflict with my output. It hadn't occurred to me to use a subdirectory.

paf31:
Could you try it and see if it solves your problem? It might not, but if it does, we could just change the default output directory to that.

paulyoung:
That works, although obviously I now need to do:
```js
var Main = require("./output/node_modules/Main");
```
instead of:
```js
var Main = require("Main");
```
which seems reasonable to me.

paulyoung:
I'm happy to submit a pull request which changes the default output directory, if that's what people want.

paf31:
I'm trying to think if it will mess up any tools, but honestly I'm not sure. Maybe `pulp`, which would be bad. It's probably worth checking.

paulyoung:
Pulp has a hard-coded build path instead of deferring to `psc-make`:

https://github.com/bodil/pulp/blob/e73b9c29ac7e02e6f487037d8c2c6898915f70a7/index.js#L163

I just confirmed locally that even with the change to `psc-make`, `pulp build` will still use `output/`

paulyoung:
To clarify, I'm not pushing for this either way.

Manually setting `output/node_modules/` works for me, but would also be glad to contribute a change.

michaelficarra:
It feels like a bit of a hack to abuse the `node_modules` directory in the node module resolution algorithm. I think relative imports are better. :+1:

paf31:
How is it a hack? Doesn't CommonJS define the resolution algorithm? My concern is that we end up tying the import code to the particular way in which we lay out things on disk, which means we might not be able to play well with other modules later.

michaelficarra:
No, node specifies their module resolution algorithm here: https://nodejs.org/docs/latest/api/modules.html#modules_all_together

I'm sure other module resolution algorithms for CommonJS systems would implement relative modules, but would be very unlikely to include `node_modules` directories in their algorithm.

garyb:
What about this one? I'm still on the side of slightly preferring relative imports to relying on `node_modules` to resolve...

paf31:
I don't really fully understand the benefit of this. If the goal is just to be able to run the `Main` module without renaming, then I don't think it's worth it.

The current approach isn't a hack at all - we create a valid set of CommonJS modules, each in their own directory. How they get used, and how they reference each other depends on the CommonJS implementation and its module resolution algorithm. The approach is simple (a module name is a `String`) and works well with NodeJS and also with Browserify, but I don't know about any others.

On the other hand, using `../` forces us to lay files out on disk in a particular arrangement. What if we don't even want all the modules to be available locally on disk. I can imagine an algorithm which looks certain modules up in other ways, from a web server, or whatever. That's why I don't like the filename approach.

garyb:
The other consideration is that by doing it this way we're trampling on the module name space root - without using `./` or `../` it means the PS-generated modules can only ever be placed in the root, with the relative URLs they could be placed under a `PS` namespace or something like that if so desired.

paf31:
To offer an alternative, I would merge a PR which made this a command line option.

garyb:
That would work for me, it'd be the `psc-make` equivalent of `--browser-namespace` for `psc`.

paf31:
True.

A better example of a different algorithm would be a situation in which you wanted one set of modules to be placed in `node_modules`, and the rest in `node_modules/my-module/node_modules`, for whatever reason (maybe different teams are responsible for different modules). Point is, the rename is just one way of making things work with Node, but there are other ways, and we should be free to lay modules out on disk as we choose.

paf31:
One more thought: I don't think it's the job of `psc-make` to package modules perfectly for any one target. That's better done by a tool like `pulp` IMO.

I think `psc-make` should do the minimal amount of work to create a valid set of modules, which it currently does. The fact that we have to do manual work right now is just a result of the lack of better tools, but that I don't think we should build such tools into `psc-make`.

kRITZCREEK:
The way I currently use my PureScript modules is with @ethul's brilliant https://github.com/ethul/purs-loader and [webpack](http://webpack.github.io/). I can then import functions from that module with whatever import syntax I prefer.
For example:

- ES6 destructuring

https://github.com/ethul/purs-loader

- "Usual" ES6 import

https://github.com/kRITZCREEK/squirrelzon/blob/purs-engine/src/js/stores/ShoppingCartStore.js#L3

So I think if people want to improve their imports they should use a packaging tool that does that. The compilers/make-tool's job is rather to enable packaging tools to use the output.

starper:
I've found one example where it could be very useful. Let's say you have some js library MyLib downloaded using npm and you wrap it using FFI like this:

```purescript
module MyLib where
foreign import foo """
  function foo (a) {
    return require('mylib').foo(a);
  } """ :: a -> a
```

Then you compile it using `psc-make` (after this it can be bundled using webpack, for example). If your output folder is `node_modules` then the original mylib will be replaced with compiled one, but if output folder is something like `libs\node_modules` (`libs` may also contain `index.js` file which requires Main module) then compiled MyLib module would require itself. Relative imports will solve this problem.

PR #943, opened 2015-03-06
Fix #942

### LINKED ISSUE #942: Imports not added in PSCI
```text
> :i Data.Profunctor.Choice 
> ((+ 1) ||| (+ 2)) $ Data.Either.Left 10
```

gives an error at runtime for `Data.Either.Left` not found. It's probably something to do with the codegen for qualified imports.

paf31:
@garyb How's this? I think this is all the cases we need, since types get handled in `Names`.


garyb:
Looks good, but I guess there's still something wrong given the tests are failing.

paf31:
Hmm it works in `psc` but not in `psc-make`.

paf31:
No wait, it works in `psc-make` too. Just not in the test suite. How odd.

paf31:
Oh right, yes, only the Main module gets code gen'd during tests. I'll have to move this to `manual/`.

garyb:
Ah yep, :+1: then.

PR #944, opened 2015-03-06
Fix #922

### LINKED ISSUE #922: Instance Resolution Problem

```purescript
module Foo where

import Data.Either
import Debug.Trace

foo = return 1

foo' :: Either Unit Number
foo' = foo

main = do
  case foo of
    (Right a) -> print a
```

It generates the error:
```
```

paf31:
@garyb How does this look?

### DIFF
--- a/examples/passing/922.purs
+++ b/examples/passing/922.purs
@@ -1,0 +1,14 @@
+module Main where
+
+class Foo a where
+  foo :: Number -> a
+
+instance fooNumber :: Foo Number where
+  foo = id
+
+bar :: forall a. (Foo a) => a
+bar = foo 1
+
+main =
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

paf31:
I just want to spend a bit of time testing this with `test-everything`, then I'll merge.

paf31:
Ok, this breaks `purescript-free` in a confusing way, so I'm going to leave this one for now.

garyb:
This isn't a viable fix then?

paf31:
Possibly, and I'll leave the code here for now, but I need more time to figure out what's really going on. The typechecker needs some real attention, and better documentation too.

PR #948, opened 2015-03-07
fix docs for (#)

hdgarrood:
We could alternatively write `d (c (b (a x)))` but I think this is a bit
clearer.


paf31:
Oops, thanks :)

PR #949, opened 2015-03-07
Update markdown docs for (#)

hdgarrood:
Whoops, forgot to do this before the previous commit


paf31:
While we're at it, we might as well make `$` consistent.

Another idea: you could always show both equalities:

`x # a # b # c # d` = `(((x # a) # b) # c) # d` = `d (c (b (a x)))`

hdgarrood:
I agree, I like that better.

paf31:
:+1: Thanks!

PR #951, opened 2015-03-08
Fix typechecking of constrained type values

garyb:
As mentioned at the end of #928, see new test for an example - previously `sequenceArray''` and `sequenceArray'''` would not pass typechecking.


paf31:
Looks good, thanks for taking a look at this.

The CI build should run the core libs, but it might be worth running `test-everything` as well to try this on things like `lens`, if you didn't already.

:ship:

joneshf:
Cool! That probably fixes that tuple issue in lens. I'll check when the next release is made.

Issue #935, opened 2015-03-03
Use whole words for psci commands


hdgarrood:
Because want both `:show` and `:set`, and potentially both `:import` and `:info`.

We already have tab completion, so this shouldn't be too onerous for users. Alternatively we could also allow commands to expand as long as they are unambiguous, eg `:i` -> error, but `:im` -> `:import`.

This would also make the output of `:?` a little clearer.

hdgarrood:
Fixed by #955

PR #955, opened 2015-03-11
PSCi commands overhaul

mrhania:
I just started playing with PureScript and I was immediately annoyed by the way PSCi handles commands (I prefer to type full forms instead of abbreviations). Turned out that there was already an issue #935 addressing this thing so I decided to fix that.

Now it should work like in GHCi: each command has unabbreviated form like `:type` or `:import` and as long as it is unambiguous you are allowed to use any prefix of this full form so `:t`, `:ty`, `:i`, `:im` work just fine. Since so far each command begins with different letter, this change is backwards-compatible (well, maybe except some weird things like `:t5` that were accepted before and now they are not) so you may not even notice. This also works well with auto-completion so hitting tab key with `:ty` will expand to `:type`.

In the code I refer to `:` commands as *directives* (whereas *command* stands for any user input). I am not very happy with that choice but I had no better idea, sorry :( Having this distinction makes code a bit more modular I think and keeps all directive strings in one place so adding new directive doesn't require you to dig through all the files.

I couldn't find any official style guidelines so I tried to mirror what I saw when I was getting familiar with the code base but obviously I might have smuggled some of my personal habits - feel free to point them out.

PS. My first pull request ever, yay!


garyb:
I like it. I guess there may be problems later if we did have commands with overlapping prefixes, but I'm sure that can be worked around too.

paf31:
Looks great. Maybe @hdgarrood has some comments? Otherwise, let's merge this. @garyb would you please do the honours, I'm out for the rest of the day?

hdgarrood:
:+1: looks good to me!

Issue #753, opened 2014-12-09
better imports in psci


hdgarrood:
Would be nice to be able to do eg
```
> :i Data.Array hiding ((..))
> :i qualified Data.Map as M
> :i Text.Smolder.HTML (head, body)
```
This is particularly handy when two modules have symbol clashes.

hdgarrood:
Fixed by #956

Issue #817, opened 2015-01-18
Importing the same module with and without qualification causes issues


garyb:
Depending on the order of the imports, the imported values are only available with or without qualification.

paf31:
Is there an example of this?

michaelficarra:
```purs
module Main where

import Debug.Trace
import qualified Debug.Trace as D

main = trace "Hello, World!"
```

paf31:
Thanks!

michaelficarra:
Note that this example works if you reverse the order of the imports.

alexchandel:
It seems the first import shadows all others:
```purs
> import Signal.Channel (send)
> import Signal.Channel (send, channel)
> :t channel
Error in module $PSCI:
Error at  line 1, column 1 - line 1, column 2:
  Unknown value channel
See https://github.com/purescript/purescript/wiki/Error-Code-UnknownValue for more information, or to contribute content related to this error.

>
```

Intuitively I would expect the union of all imports to be available.

PR #956, opened 2015-03-11
First stab at better imports in psci

hdgarrood:
refs #753 
Issues:

* [ ] What happens if the same module is imported multiple times? Do we want to include an :unimport command?
* [ ] This probably needs some kind of refactoring (in particular, it feels a bit wrong having `parseImportDeclaration'` in the export list)
* [ ] Completion needs some work
* [x] Fails silently in some situations, eg `:import Prelude.Unsafe hiding (doesNotExist)`
* [ ] `:browse Q` after doing `import qualified M as Q` should work
* [x] Disallow `let import ...` and any other declarations that shouldn't really follow 'let'
* [x] Allow bare declarations where sensible to make up for `let import ...` etc being disallowed

But importing qualified modules, as well as importing modules with explicit import or hiding lists seems to work :) Some feedback at this stage would be nice.

### DIFF
--- a/psci/Commands.hs
+++ b/psci/Commands.hs
@@ -66,4 +66,13 @@ data Command
   --
   | Show String
 
+-- | All of the data that is contained by an ImportDeclaration in the AST.
+-- That is:
+--
+-- * A module name, the name of the module which is being imported
+-- * An ImportDeclarationType which specifies whether there is an explicit
+--   import list, a hiding list, or neither.
+-- * If the module is imported qualified, its qualified name in the importing
+--   module. Otherwise, Nothing.
+--
 type ImportedModule = (ModuleName, ImportDeclarationType, Maybe ModuleName)
--- a/psci/Commands.hs
+++ b/psci/Commands.hs
@@ -50,9 +50,9 @@ data Command
   --
   | Reset
   -- |
-  -- Binds a value to a name
+  -- Add some declarations to the current evaluation context.
   --
-  | Let [Declaration]
+  | Decls [Declaration]
   -- |
   -- Find the type of an expression
   --
--- a/psci/Commands.hs
+++ b/psci/Commands.hs
@@ -32,7 +32,7 @@ data Command
   -- |
   -- Import a module from a loaded file
   --
-  | Import ModuleName
+  | Import ImportedModule
   -- |
   -- Browse a module
   --
@@ -65,3 +65,5 @@ data Command
   -- Show command
   --
   | Show String
+
+type ImportedModule = (ModuleName, ImportDeclarationType, Maybe ModuleName)
--- a/psci/Parser.hs
+++ b/psci/Parser.hs
@@ -60,7 +60,7 @@ parseCommand cmdString =
       Just D.Help -> return C.Help
       Just D.Quit -> return C.Quit
       Just D.Reset -> return C.Reset
-      Just D.Import -> C.Import <$> parseRest P.moduleName arg
+      Just D.Import -> C.Import <$> parseRest P.parseImportDeclarationTail arg
       Just D.Browse -> C.Browse <$> parseRest P.moduleName arg
       Just D.Load -> return $ C.LoadFile (trim arg)
       Just D.Show -> return $ C.Show (trim arg)
--- a/psci/Parser.hs
+++ b/psci/Parser.hs
@@ -13,8 +13,8 @@
 --
 -----------------------------------------------------------------------------
 
-module Parser (
-    parseCommand
+module Parser
+  ( parseCommand
   ) where
 
 import Prelude hiding (lex)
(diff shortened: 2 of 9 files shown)


paf31:
Looks nice. 

Maybe we should just start using `parseImportDeclaration`, and match `ghci`?

paf31:
Also, note that you can already say this:

```
let import qualified Foo as Bar
```

which is a bit odd.

hdgarrood:
Matching ghci - yeah, I can't think of any good reason to use `:import` instead.

Re `parseImportDeclaration`, I'm a little uneasy about relying on getting an `ImportDeclaration` back when the types say we could get any kind of `Declaration` though.

I guess we should disallow `import let ...`?

hdgarrood:
That most recent commit appears to have fixed the problem where nonexistent values in explicit import/hiding lists don't produce errors... somehow :)

psci/Parser.hs
```diff
@@ -86,3 +85,6 @@ parseCommand cmdString =
     where
     manyDecls :: P.TokenParser [P.Declaration]
     manyDecls = C.mark (many1 (C.same *> P.parseDeclaration))
```
hdgarrood: Regarding the `let import ...` issue, I think we can fix that by using `parseLocalDeclaration` here instead of `parseDeclaration`.

If we were going to do that, though, I think we should allow bare data, type synonym, type class, and type instance declarations (ie, without a 'let') at the same time, so that we aren't removing useful features that people might be using.

hdgarrood: And I think we should potentially allow foreign import declarations, too. This currently works:

```
> let foreign import foo "var foo = 1 + 2" :: Number
> foo
3
```

hdgarrood:
Also, not directly related to this PR: we currently have `data ImportDeclarationType = Unqualified | Qualified [DeclarationRef] | Hiding [DeclarationRef]`. I feel like the `Unqualified` and `Qualified` constructors would be better named `Implicit` and `Explicit` respectively? Whether a module is qualified in an `ImportDeclaration` is specified by the `Maybe ModuleName` argument, not the `ImportDeclarationType` one.

garyb:
> I feel like the `Unqualified` and `Qualified` constructors would be better named `Implicit` and `Explicit` respectively?

Yeah, that definitely makes sense.

hdgarrood:
I'd be happy to make that change, then. Do you think it would be ok to expand the scope of this PR a little to include it?

garyb:
If you make that change separately I'll just merge it immediately, but this looks good too so if you add it to this PR I'm sure it will get merged either way. :)

hdgarrood:
Regarding an `:unimport` command, I don't think we need one, given that we have `:reset`.

joneshf:
Being able to remove a module at a time is quite beneficial.

psci/Commands.hs
```diff
@@ -65,3 +65,5 @@ data Command
   -- Show command
   --
   | Show String
+
+type ImportedModule = (ModuleName, ImportDeclarationType, Maybe ModuleName)
```
paf31: Could you please add a comment to describe the three parts?

hdgarrood: Will do.

hdgarrood:
@joneshf ok, I agree. On second thoughts, though, I would rather tackle that separately, and probably after #817 has been addressed.

hdgarrood:
This is a big enough change that I'm not really comfortable without tests. I was thinking that a test could be an input and expected output; `[(String, String)]`, perhaps. Then we just write a bunch of them and work out how to get haskeline to run them. Does that sound sensible?

hdgarrood:
Scope creep! Anyway this works now:
```
$ cabal run psci -- -m
<snip>
> data Foo = Bar | Baz

> class Showy a where
    showy :: a -> String

> instance showyFoo :: Showy Foo where
    showy Bar = "Bar"
    showy Baz = "Baz"

> showy Bar

"Bar"
```

paf31:
The code looks good, but I'm not sure I follow your last code sample...

hdgarrood:
The point is just that more types of declarations (eg, data, type class, instance) now work in psci. The above example wouldn't work in the current psci - you get 'expecting let or expression'. At the same time, I've disallowed `let import M as Q`, `let data Foo = Bar | Baz`.

paf31:
I think it looks great. I'll stare at it a bit more tomorrow and try to get it merged.

hdgarrood:
Cool. :) In that case I will wait, and then address the other incomplete items in the description separately.

paf31:
@garyb Shall I merge this? You know most about imports, but it looks good to me.

garyb:
Looks good to me too, the import changes are all `psci` related really aside from the `Qualified` -> `Explicit` rename stuff which is :+1: to me.

paf31:
Hmm. Completion for imports seems to be slightly off now. If I complete `import Data.`, I get a bunch of function names, not module names.

hdgarrood:
I think that issue has existed since 0.6.8 at least, I can reproduce it there. But yeah, I will take a look at this shortly.

paf31:
Oh ok, never mind then :)

PR #958, opened 2015-03-13
Fix JS operator precedence

### LINKED ISSUE #957: 1 < 0 == not (1 < 0)
The above expression evaluates to true, somewhat surprisingly, as @Javran pointed out recently on IRC.

A module demonstrating the bug:

```purescript
module Main where

import Debug.Trace

main = do
  trace (show (not (1 < 0)))
```

And it compiles to:

```javascript
PS.Main = (function () {
    "use strict";
    var Debug_Trace = PS.Debug_Trace;
    var Prelude = PS.Prelude;
    var main = Debug_Trace.trace(Prelude.show(Prelude.showBoolean)(!1 < 0));
    return {
        main: main
    };
})();
```

Looks like it needs extra parens around `1 < 0`?

garyb:
Resolves #957


garyb:
This is right according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

hdgarrood:
Are we able to make assertions about program output in the test suite? This seems like a good candidate for this kind of test.

michaelficarra:
I'd rather not. We should only care about the behaviour.

garyb:
Kind of, we can throw errors within the tests. In fact there is already a precedence test which does this, but it doesn't cover this case: https://github.com/purescript/purescript/blob/145c4c5b6bab8273a9176542fad0db04ca16d045/examples/passing/OperatorAssociativity.purs

hdgarrood:
Ahh, good, that's reassuring.

What's the difference between caring about a program's output and its behaviour? Can an automated test even make that distinction?

michaelficarra:
@hdgarrood I thought you meant making assertions about the output program text, not its result.

hdgarrood:
ohhh, I see. Never mind then!

michaelficarra:
@garyb It doesn't look like you've grouped them into precedence classes, but instead created a total ordering.

garyb:
Yeah, you're right. Making changes like this makes me a little nervous given the lack of comprehensive tests for operators, so I wanted to be conservative about it - I wasn't sure if @paf31 hadn't grouped them for a reason.

I probably should just do it properly and add more tests :)

paf31:
I think I added tests for any previous breakages here, but I'm sure this used to be the other way round, and got changed for some reason. Anyway, if the tests pass, then I'm happy. We should add some cases for things in the same precedence group probably.

garyb:
@michaelficarra does this look ok to you now?

michaelficarra:
LGTM. I'd want to see more tests with `new`/`new` and `new`/function-call combinations, though. Here's some inspiration: https://github.com/shapesecurity/shift-codegen-js/blob/881f3ec8a60fc6e96fc35414c960f8973415915b/test/simple.js#L280-L298

garyb:
Yeah, I'd like to do those too, but we don't really have a way of generating that kind of code from `.purs` input - `new` is only used inside the `create` function for data constructors and for class instances so we'd need to set up some additional non-purescript-code-driven tests. 

Alternatively, perhaps we can eliminate our JS-specific code entirely in favour of using `language-javascript` or something when the `CoreImp` stuff is in. We'd get source maps as well then.

PR #962, opened 2015-03-15
Rebindable do notation

### LINKED ISSUE #742: Do notation for semigroups
This is a bit of an odd idea, but I'll put it here for discussion purposes.

We could add an analogue of do notation for semigroups, which gave the nice notational advantages of do notation (with no support for `<-`).

For example, we could build a DOM library which looked like Blaze HTML but without the need to make Markup into a monad:

(Not sure what is the best syntax here...)

```haskell
render $
  html do
    head do
      title $ text "PureScript"
      link' [rel := "stylesheet", href := "style.css"]
    body do
      h1 $ text "PureScript"
      div' [id_ := "intro"] $ text "PureScript is a strongly-typed language which compiles to JavaScript"
```

Originally, I was thinking that we would use a separate keyword, but maybe we can just overload `do`.

Maybe we could even use the fact that every `Monad` is a `Semigroup`, and only require a `Monad` constraint in the case of a `<-`, but that seems to put a strange restriction on the `Monad` type class.

paf31:
@garyb @joneshf @jdegoes Could you please review? I thought this would be pretty neat.

Resolves #742.

### DIFF
--- a/examples/passing/RebindableSyntax.purs
+++ b/examples/passing/RebindableSyntax.purs
@@ -1,0 +1,12 @@
+module Main where
+
+example :: String
+example = do
+  "Do"
+  " notation"
+  " for"
+  " monoids"
+  where
+  (>>=) x f = x <> f unit
+
+main = Debug.Trace.trace example
--- a/examples/passing/RebindableSyntax.purs
+++ b/examples/passing/RebindableSyntax.purs
@@ -1,12 +1,37 @@
 module Main where
 
-example :: String
-example = do
+example1 :: String
+example1 = do
   "Do"
   " notation"
   " for"
-  " monoids"
+  " Semigroup"
   where
   (>>=) x f = x <> f unit
 
-main = Debug.Trace.trace example
+(*>) :: forall f a b. (Apply f) => f a -> f b -> f b
+(*>) fa fb = const id <$> fa <*> fb
+
+newtype Const a b = Const a
+
+runConst :: forall a b. Const a b -> a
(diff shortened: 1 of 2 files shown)


paf31:
This currently breaks `StrMap` (see [here](https://travis-ci.org/purescript/purescript/builds/54497740) - also the CI build doesn't break when core-tests does :-1:), since the `Prelude` is imported qualified, but it can be easily fixed by importing `(>>=)` separately, or by shadowing it.

jdegoes:
Sweet. Would make smoulder, affjax, etc., DSLs faster & more type safe. Looks good to me.

paf31:
This is working as of the last `maps` release. Shall I merge?

garyb:
How did it break `maps`?

paf31:
`maps` used to import `Prelude` qualified, which meant `>>=` wasn't in scope, but now it is since I removed the qualified import in my docs commit.

garyb:
So this will break `do` syntax in any module that currently imports `Prelude` qualified or hides `(>>=)`?

I don't think that's unreasonable, I'd just like to know in advance what it may kick up.

paf31:
Yes, `(>>=)` needs to be in scope, unqualified. You can always fix it by either adding an unqualified import for it, or using `hiding` or just rebinding it in a `where` block.

garyb:
:+1:

garyb:
By :+1: I mean :shipit:

paf31:
:)

PR #964, opened 2015-03-15
New module for warnings capability.

nicodelpiano:
This PR comes from #803 . 
It's not finished, there's a lot to do, but I guess it's a good start (It's almost the same as `Errors.hs`).

### DIFF
--- a/src/Language/PureScript/Warnings.hs
+++ b/src/Language/PureScript/Warnings.hs
@@ -42,7 +42,7 @@ import qualified Text.PrettyPrint.Boxes as Box
 data WarningMessage
  = UnusedTypeVariable ProperName
  | ShadowedTypeVariable ProperName
- | PositionedWarning
+ | PositionedWarning SourceSpan WarningMessage
  deriving (Show)
 
 -- |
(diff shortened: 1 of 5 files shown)


src/Language/PureScript/Warnings.hs
```diff
@@ -0,0 +1,70 @@
...
+import Language.PureScript.Kinds
+import Language.PureScript.TypeClassDictionaries
+
+import qualified Text.PrettyPrint.Boxes as Box
+
+-- |
+-- A type of warning messages
+--
+data WarningMessage
+ = UnusedTypeVariable Type
```
paf31: Probably can just make these `ProperName` instead of `Type`.

src/Language/PureScript/Warnings.hs
```diff
@@ -0,0 +1,70 @@
...
+ | ShadowedTypeVariable Type
+ deriving (Show)
```
paf31: You'll want to add `PositionedWarning`.

paf31:
Looks like a good start, thanks! If you have any questions, let me know.

src/Language/PureScript/Warnings.hs
```diff
@@ -0,0 +1,72 @@
...
+
+import qualified Text.PrettyPrint.Boxes as Box
+
+-- |
+-- A type of warning messages
+--
+data WarningMessage
+ = UnusedTypeVariable ProperName
+ | ShadowedTypeVariable ProperName
+ | PositionedWarning
```
paf31: See here:
https://github.com/purescript/purescript/blob/a60f34a100bf0981f4a0278b2fca4a1958bf2c8a/src/Language/PureScript/Errors.hs#L116

nicodelpiano:
There are some functions that are duplicated (like `indent`, for example), because I think they were used only in the errors module.
Should we create a new module, or putting them in another place?

Issue #421, opened 2014-05-22
Superclass constraints are ignored in instances


garyb:
``` haskell
-- With this class definition:

class (Monad m) <= MonadError e m where
  throwError :: forall a. e -> m a
  catchError :: forall a. m a -> (e -> m a) -> m a
  
-- This instance doesn't work:
  
instance monadErrorMaybeT :: (MonadError e m) => MonadError e (MaybeT m) where
  throwError e = lift (throwError e)
  catchError = liftCatchMaybe catchError
  
-- You have to add another `Monad m` constraint:

instance monadErrorMaybeT :: (Monad m, MonadError e m) => MonadError e (MaybeT m) where
  throwError e = lift (throwError e)
  catchError = liftCatchMaybe catchError
```

PR #965, opened 2015-03-15
WIP on #421

paf31:
This isn't ready yet, but it's close. It breaks on `RWS.Class` in `transformers` because of an infinite search. The problem is in the `typeHeadsAreEqual` function.

I think we have two options:

1. Don't use `TypeVar`, and introduce a new constructor in `Type` for "unsolved variables".
2. (I like this one because it moves us closer to something like fundeps) - allow unification of `TypeVar`s on the left during solving, but mark it as a "partial match". If all types in the instance head match, we require that at least one of them be non-partial, otherwise we fail with an "Underspecified type" error.

### DIFF
--- a/examples/failing/438.purs
+++ b/examples/failing/438.purs
@@ -1,0 +1,9 @@
+module Main where
+    
+data Fix f = In (f (Fix f))
+
+instance eqFix :: (Eq (f (Fix f))) => Eq (Fix f) where
+  (==) (In f) (In g) = f == g
+  (/=) a b = not (a == b)
+  
+example = In [] == In []
--- a/examples/passing/421.purs
+++ b/examples/passing/421.purs
@@ -1,0 +1,12 @@
+module Main where
+
+class (Monad m) <= MonadBlah b m where
+  blah :: b -> m Unit
+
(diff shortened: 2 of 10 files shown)


paf31:
I've implemented 2) in the `421` branch.

On the plus side, this now works, so if we're willing to change `RWS`, we should be good:

```purescript
instance monadRWS :: (MonadReader r m, MonadWriter w m, MonadState s m) => MonadRWS r w s m
```

paf31:
Also, with some shuffling of constraints (generally in the right direction, I would say), I can make `MonadRWS` work as it exists now.

I'll test this some more before merging though, even if we decide to go this route.

paf31:
Ok, the situation is now that everything in core builds, but `RWS` is slow. However, the reason it is slow is that we are doing a lot of redundant searching-and-pruning-and-bailing-out, which can be reduced by tidying up the `MonadRWS` hierarchy a bit.

garyb:
Do you think we should be exposing `maxSearchDepth` as an option? I assume the idea is it's set to a "reasonable" value by default, and people can increase it in special cases, but do you think the average user is going to want to know or have to understand something like that?

Nice work though!

PR #967, opened 2015-03-16
psci - refactoring, add tests, fix completion (partially)

hdgarrood:
This is pretty big, sorry.

However, on the plus side:
* There are no longer any data constructor name clashes, which means significantly less needless qualification everywhere
* There is now an ADT for things that can be displayed via `:show`, which simplifies the code a bit (you don't have to check for invalid strings now)
* better error messages when you try to `:show` something that isn't showable
* There is now a small test suite for completions
* Completion for `:browse`, which broke somewhere between 0.6.8 and now (that might have been me, sorry!), now works again.
* Completions are now forbidden from modifying the `PSCiState`, by using a `ReaderT` instead of a `StateT`
* Fix some naming confusion for 'command' vs 'directive'; `Directive.commands` was a misnomer, for example.

`psci-tests` is currently an executable in the cabal file because I want to be able to fire up a REPL with `cabal repl psci-tests` and play with things. Obviously this isn't ideal; would you like me to change it into a test suite of its own, or include it in the main test suite? Whatever option we choose, I think there should be a way of only running the psci tests. Having to run the whole compiler test suite every time when working on psci would be irritating.

Completion is still broken in a few places. For example, the tests currently fail because `import` completions still give you identifiers, not modules. I intend to address that - I just wanted to check in at this stage for a review, before I end up with a diff that's just far too large to take in in one go.


hdgarrood:
Also I'd be happy to rework these commits so that the history is nicer, and so that it's easy to review this by looking at each commit individually, since this is not really the case at the moment. In particular there are some changes in this branch that I later realised are silly and which I reverted in a later commit.

paf31:
I think this all looks good. I don't mind big changes like this as long as stuff doesn't break and the structure is sensible, which both seem to be true here.

I think it would be nice to turn `psci-tests` into a test suite if only because it would increase the coverage. Maybe you could keep your `cabal repl` feature by moving code into a library as necessary?

hdgarrood:
Cool. :)

I've just realised that if I make it into a test suite I can just do `cabal repl test:psci-tests`, so that seems like the best option.

hdgarrood:
Turns out you don't even need to add the `test:` prefix, `cabal repl psci-tests` just works.

hdgarrood:
The failing tests are new ones in the psci-tests suite that test `import` completions. I can fix them on this branch, or in a new PR if you'd rather merge this now. Up to you.

hdgarrood:
Actually I wonder if we'd be better off doing:

```
hpc-coveralls --exclude-dir=tests tests
hpc-coveralls --exclude-dir=src psci-tests
```

The psci test suite doesn't really test the code in `src/` and so it seems a bit misleading for functions like `isExported` and so on to be marked as covered.

hdgarrood:
So I think the last thing to do here is to fix the coverage. Currently it's only displaying coverage for the psci tests. I think we can sort this out by adding a TEST_SUITE environment variable to the `env` matrix, and get travis to run two builds - one with the normal test suite, and one with the psci test suite.

hdgarrood:
Yes! We now get two separate coverage reports; one for each test suite. I think we should also ignore stuff in dist/build/autogen; I'll just do that and then I think this is ready to merge.

paf31:
Nice! 

Could you please rebase this before I merge? Does it need anything else or is it good to go?

I wonder if we could increase `psci` coverage by simulating the console using a free monad or something like that, and running simulated user sessions.

hdgarrood:
I rebased it onto `master` after 0.6.9.3 was released so this should be mergeable now. Or would you like me to rebase in order to clean up the history?

There is one last thing to do before I think this is good to go: wait and see what the result of the most recent travis CI build and the coverage is, and check they work ok.

Definitely agree re simulating psci sessions. I don't like that the current test harness tries to reconstruct completions based on what I guessed haskeline would do.

paf31:
Sorry, I was referring to the git history.

hdgarrood:
What specifically about the history should I change?

paf31:
Oh, I was just hoping the commits could be collapsed. There seems to be quite a bit of backtracking and it would be nice to have 1 commit in the log rather than 39 :) If you'd rather break it into 2 or 3 commits, that's cool too, but one should be simple with `rebase -i`.

hdgarrood:
:+1: will do

paf31:
@hdgarrood Thanks!
@garyb ok to merge?

paf31:
This looks great to me, by the way. I'll spend a bit more time digging into it over the next few days.

hdgarrood:
:) thanks!

Also if you wonder about the `String` arguments for `CtxDirective` and `CtxFilePath` - I don't think they're necessary, and I'd like to try to remove them.

garyb:
:+1:

PR #969, opened 2015-03-16
Don't export refEq and refIneq

### LINKED ISSUE #966: Should refEq be in Prelude.Unsafe?
Generally, I would expect that `a == b` implies `f a == f b` for any a, b, f, as long as you're not using the FFI.

But if we have:
```purescript
data Thing = Thing Number
instance eqThing :: Eq Thing where
  (==) (Thing x) (Thing y) = x == y
  (/=) x y = not (x == y)

a = Thing 0
b = Thing 0
f x = if x `refEq` a then true else false
```

then `a == b` but `f a /= f b`, and information about how `a` and `b` are laid out in memory is leaking into the program... I'm not sure if this is really a problem. It feels a bit like a subversion of the type system though. Just a thought.

garyb:
Resolves #966.

`test-everything` still passes (as one would hope!)

### DIFF
--- a/prelude/prelude.purs
+++ b/prelude/prelude.purs
@@ -20,7 +20,7 @@ module Prelude
   , negate
   , DivisionRing
   , Num
-  , Eq, (==), (/=), refEq, refIneq
+  , Eq, (==), (/=)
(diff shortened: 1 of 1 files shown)


paf31:
:+1:

PR #971, opened 2015-03-17
Core imperative representation

### LINKED ISSUE #711: Refactor imperative core AST

garyb:
Resolves #711, I finally got around to making this work on the train this morning.

There are still things to do, but I thought I'd get it on the radar now it can run the test suite.
- [x] `CoreImp` modules should still contain declarations rather than `Statement`s - more suitable for a wider range of targets, and everything at that level already is a `var` or `function` anyway
- [ ] If I remember rightly a lot of annotations are being thrown away during `CoreFn -> CoreImp` at the moment
- [x] Mega cheating going on using the `Op` constructor rather than `Ident` to get `this` through to the JS without it being turned into `$$this`
- [x] CoreImp needs traversals
- [ ] Update as much of `Optimizer` as possible to work on `CoreImp`
- [ ] Move comments into metadata annotations
- [ ] Use annotations to indicate zero-argument functions if/where possible rather than using `__unused`

Probably some other stuff too that I'm not thinking of right now.


paf31:
Nice! I'll have a look through this later today.

puffnfresh:
:fireworks:

andyarvanitis:
Really looking forward to rewriting the C++11 backend on this baseline. Do you guys feel comfortable enough with what's here to recommend me working off of it? I do of course anticipate changes, but I'd also like to be able to provide feedback (for C++'s needs) at the right time.

garyb:
I think so, but then I would say that as it's my work so far ;)

We've not really discussed where it's at yet, but I'm mostly happy with the `CoreImp` AST as it is here (except for potentially adding those cases for dealing with array length checking/index reading/slicing).

I'd really like to know how it works out for your case and if there are changes we need to make. At the moment it was mostly built with JS/Lua/JVM capabilities in mind, but only because that's what I'm most familiar with.

I'll rebase when I get a minute tonight so it's not lagging behind `master`.

andyarvanitis:
Great, thanks. I'll at least start having a closer look, then. I'm currently reviewing (and possibly rethinking) how typeclasses are represented in C++, so this might in turn have an influence on that.

garyb:
Typeclasses are desugared into regular data constructors now, so hopefully you won't need a special representation for them at all. That currently happens in the `CoreFn -> CoreImp` desugaring step, but we hope to move it way up the chain before typechecking even (#928 - the typechecker has some issues that prevent it currently).

garyb:
@andyarvanitis Ok, this is in a more up-to-date state now.

andyarvanitis:
Thanks!

garyb:
Maybe we should shift this to 0.9 too.

paf31:
This is the PR, right? The issue is in 0.9 already :smile:

garyb:
Doh :smile: I just looked at the milestone and this popped up.

Issue #931, opened 2015-03-02
Parser or other problem in externs.purs


jutaro:
My code compiles, but when I use .psci I get the error: 
"unexpected LBrace
expecting no indentation or end of input"

The offending code in the externs.purs file looks this way:
foreign import shaders :: Graphics.WebGL.Shaders {  | ConeCanvas.Frontend.DrawTypes3d.MyBindings}

The corresponding source code looks like:
shaders :: GL.Shaders (Object MyBindings)
shaders = GL.Shaders ...

which uses the following type (abbreviated):
type MyBindings = (aVertexPosition :: GL.Attribute GL.Vec3, ...)

PR #973, opened 2015-03-17
Restrict record type pretty printer

mrhania:
This is my attempt at solving #931. Simplifying provided example mentioned in the issue:

    type Foo = (foo :: Number)

    bar :: Object Foo
    bar = { foo: 0 }

yields `foreign import bar :: {  | Foo }` in generated `externs.purs` file which is not accepted by the compiler. This happens because pretty printer has a rule to turn all `Object row` expressions to neatly printed records. However, pretty printer is not aware of aliases, thus yields this incorrect statement.

So I figured out three solutions:

1. Restrict pretty printer to output record syntax only with *pretty rows* (most straightforward, implemented in this PR).
2. Make pretty printer expand type aliases (which is bad because it would output ugly code in some cases and would require massive pretty printer refactoring).
3. Just support `foo :: { | Foo }` syntax. I don't know if this is by design but I see no charm in making it valid syntax. Actually, I was confused as why PureScript allows only type variable after the `|` and not row in any kind of form (e.g. `foo :: { x :: Number | (y :: Number, z : Number) }`).

So, for now pretty printing type of bar from the provided example would yield `Object Foo`. If there are no objections I can also work on the third option - adding syntax support should be quite easy, I am not sure what adjustments would be required in type checker though.

### DIFF
--- a/examples/passing/Objects.purs
+++ b/examples/passing/Objects.purs
@@ -15,6 +15,11 @@ module Main where
   typed :: { foo :: Number }
   typed = { foo: 0 }
 
+  type Foo = (foo :: Number)
+
+  typed2 :: Object Foo
+  typed2 = { foo: 0 }
+
   test2 = \x -> x."!@#"
 
   test3 = typed."foo"
--- a/src/Language/PureScript/Pretty/Types.hs
+++ b/src/Language/PureScript/Pretty/Types.hs
@@ -93,6 +93,9 @@ insertPlaceholders = everywhereOnTypesTopDown convertForAlls . everywhereOnTypes
     go idents other = PrettyPrintForAll idents other
   convertForAlls other = other
 
+  -- Object type application should be sugared into record syntax. However,
+  -- this may produce illegal code when object is applied to non-trivial row
+  -- (e.g. type alias). This function checks whether row is of sugarable form.
   prettyRow :: Type -> Bool
   prettyRow (RCons _ _ r) = prettyRow r
   prettyRow REmpty = True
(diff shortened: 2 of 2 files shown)


paf31:
I like it, thanks! Does this fix #931? Could you please add a code comment explaining why this is necessary?

paf31:
Oh, and maybe a test?

mrhania:
> Does this fix #931?
Yep.

> Could you please add a code comment explaining why this is necessary?
Sure, I wonder why I haven't.

> Oh, and maybe a test?
Actually, I wanted to. The problem is that this does not cause the compilation itself to fail - its just that it generates incorrect code. If you have an idea how such test could look like, please let me now - I will gladly add it.

paf31:
> The problem is that this does not cause the compilation itself to fail

That's true, but it's still nice to cover the new code anyway.

mrhania:
Done. Maybe in the future you can add some extra pass to the test suite where generated `externs.purs` files are also checked?

paf31:
Sounds good. If you want to just parse the externs file at the end, that should be pretty simple.

paf31:
I just thought - what about:

```purescript
(a :: A | B)
```

which is still invalid? This PR definitely improves the situation, but can we come up with a plan to fix this too?

paf31:
I'd be in favor of just allowing type constructors on the right of `|`.

mrhania:
Totally agree. As I said above in 3rd proposed solution I was actually confused why PureScript does not support this in the first place. I can try to work on that but that would probably require me to dive into type checker which is quite terrifying for me right now.

mrhania:
Alternative solution was proposed and accepted in #978 so this is no longer relevant.

PR #976, opened 2015-03-18
Add missing module 'Directive' to other-modules

hdgarrood:
I tried to install purescript 0.6.9 into my global package database, but got this:

```
psci/Parser.hs:23:18: Could not find module ‘Directive’
```

I think this is because it was missing from `other-modules` in the .cabal file. This commit hopefully will fix that.


paf31:
:+1:

PR #978, opened 2015-03-18
Flexible row syntax

### LINKED ISSUE #931: Parser or other problem in externs.purs

mrhania:
As discussed in #973, I made row syntax more flexible. That is, now constructs such as `(foo :: Number | (bar :: Number))` and `(foo :: Number | Bar)` (where `Bar` is some row) are supported. In particular, this causes `Object Foo` and `{ | Foo }` to be the same thing and thus fixes #931 also making #973 obsolete.

Turned out that only tiny changes to the parser were required and everything type checks just fine, although it would be nice if someone more familiar with type checker could confirm that it is not just coincidence. Generally, this improvement should be backwards compatible.

### DIFF
--- a/examples/failing/RowConstructors1.purs
+++ b/examples/failing/RowConstructors1.purs
@@ -1,0 +1,6 @@
+module Main where
+
+data Foo = Bar
+type Baz = { | Foo }
+
+main = Debug.Trace.trace "Done"
--- a/examples/failing/RowConstructors2.purs
+++ b/examples/failing/RowConstructors2.purs
@@ -1,0 +1,6 @@
+module Main where
+
+type Foo r = (x :: Number | r)
+type Bar = { | Foo }
+
+main = Debug.Trace.trace "Done"
--- a/examples/failing/RowConstructors3.purs
+++ b/examples/failing/RowConstructors3.purs
@@ -1,0 +1,6 @@
+module Main where
+
+type Foo = { x :: Number }
+type Bar = { | Foo }
+
+main = Debug.Trace.trace "Done"
--- a/examples/passing/RowConstructors.purs
+++ b/examples/passing/RowConstructors.purs
@@ -22,4 +22,19 @@ bar' = id' bar
 baz :: Baz
 baz = { x: 0, y: 0, z: 0, w: 0 }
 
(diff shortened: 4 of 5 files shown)


garyb:
Looks good to me. @paf31 will have to weigh in on the typechecker implications, but I doubt there is much of a problem here - type synonyms are expanded during checking, so it was mainly the parser that prevented this from working before. I assume it fails in a way you'd expect if you try to use an ADT on the rhs of `|`?

mrhania:
> I assume it fails in a way you'd expect if you try to use an ADT on the rhs of `|`?

Yep, it fails during type checking: RHS of `|` has to be of kind `# *`.

garyb:
:thumbsup:

paf31:
Nice :) You might need type applications too. We could probably exclude `REmpty` and `RCons` though because they would never be pretty printed, but that's not a big deal.

mrhania:
> You might need type applications too.

Hmm, now I wonder: wouldn't it be best to just parse any type on the RHS of `|` and let the type checker take care of possible errors? I checked and looks like it works. Can you think of any pathological scenario where applying this strategy could led to some unexpected behavior in the type checker?

> We could probably exclude REmpty and RCons though because they would never be pretty printed, but that's not a big deal.

I am not sure what do you mean by that, could you elaborate a bit?

paf31:
I'm fine with that. If the tests pass, I'll merge :)

I just meant that the pretty printer will never generate something like `{ a :: A | (b :: B | r) }`, it will just print, `{ a :: A, b :: B | r }`, so we don't need explicit support, but just using `parseType` is simple, so let's do that.

paf31:
Thanks!

PR #985, opened 2015-03-19
Provide more information in `exportedDctors`

hdgarrood:
Currently, `exportedDctors` returns only a `[ProperName]` with the names
of the data constructors. This is no good if you want to know about
the arguments of each constructor. This commit fixes the issue by
changing the return type to `[(ProperName, [Type])]`.

On a related note, I wonder if the documentation for `isExported` and
`isDctorExported` should encourage you to use `exportedDecls` and
`exportedDctors`, as these are 'higher level' versions of what is really
the same function, which won't require you to pattern match (and
therefore hopefully be safer and less likely to break because of
updates).

### DIFF
--- a/psci/Completion.hs
+++ b/psci/Completion.hs
@@ -218,7 +218,7 @@ dctorNames m = nubOnFst $ concatMap dctors dnames
   onlyDataDecls = (filter P.isDataDecl (P.exportedDeclarations m))
 
   dctors :: (N.ProperName, P.Declaration) -> [(N.ProperName, P.Declaration)]
-  dctors (name, decl) = map (\n -> (n, decl)) (P.exportedDctors m name)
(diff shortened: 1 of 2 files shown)


hdgarrood:
There are two separate commits listed here because somehow, in my first attempt, I ended up basing the branch off of a commit before the merge of my recent psci PR.

hdgarrood:
Also this is a breaking change. I don't think it will cause much pain, though; it might just be pursuit and possibly psc-pages that use this function currently.

paf31:
Looks good to me. @garyb Does this affect anything you use?

garyb:
Nope, all good by me. I think this is only used for externs and stuff - the `Environment` is used everywhere else.

garyb:
Oh actually, I guess not even that, as otherwise this wouldn't build given the changes only update its use in `psci`. :)

paf31:
:+1:

PR #986, opened 2015-03-19
Better psci help

hdgarrood:
Now that `:i` has been replaced with `import`, and someone asked in IRC about where `:i` had gone, I thought it might be a good idea to state it explicitly in the `:help` output. This commit also contains a couple of other small tweaks.

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -138,22 +138,10 @@ helpMessage = "The following commands are available:\n\n    " ++
         , desc
         ]
 
-  extraHelp = unlines
-    [ "Once a file has been loaded, values and types in the module(s) that it"
-    , "contains are available fully qualified:"
-    , ""
-    , "    > :type Data.Function.on"
-    , "    forall a b c. (b -> b -> c) -> (a -> b) -> a -> a -> c"
-    , ""
-    , "Alternatively, you can import a loaded module in order to bring types and"
-    , "values into the current scope - then, you can use them unqualified."
-    , ""
-    , "Modules can be imported normally, as in PureScript code, eg:"
-    , ""
-    , "    > import Control.Monad"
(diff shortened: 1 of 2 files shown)


psci/PSCi.hs
```diff
@@ -124,17 +124,36 @@ expandTilde p = return p
...
+    let cmd = ':' : D.stringFor dir
+    in intercalate " "
+        [ cmd
+        , replicate (11 - length cmd) ' '
+        , arg
+        , replicate (11 - length arg) ' '
+        , desc
+        ]
+
+  extraHelp = unlines
```
paf31: Does this show up every time? Not necessarily a bad thing, I'm just curious to see how it looks. I wonder if this level of detail is best reserved for a "user's guide" on the wiki or something.

hdgarrood: With this code, yes, every time you do `:?` or `:help` (and only when you do that).

Linking to the wiki also sounds good. Up to you.

hdgarrood: https://gist.github.com/hdgarrood/c136089e9bc358a3276c

paf31: I guess my preference would be to stick the extra content on the wiki with a link in `psci` honestly. I like that approach because it means we don't have to do a release to update textual content in the compiler, like we do for error messages now.

hdgarrood: Oh yes, good point. Ok then, give me a sec

hdgarrood: https://github.com/purescript/purescript/wiki/PSCi :+1: / :-1: ?

paf31: :+1: :+1: :+1:

hdgarrood: :smile:

paf31:
:+1: Thanks!

Issue #981, opened 2015-03-18
Incorrect error code link for occurs check failures


hdgarrood:
```
> let x = \_ -> [x]
Error in module $PSCI:
Error in value declaration x:
Error at  line 1, column 5 - line 1, column 17:
  Infinite type detected:
    u1 -> [u0]
See https://github.com/purescript/purescript/wiki/Error-Code-ErrorUnifyingTypes for more information, or to contribute content related to this error.
```

Should this be linking to a separate page about type errors arising from infinite types?

paf31:
Yes definitely.

paf31:
The error is here if any newcomer would like to tackle this:

https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Errors.hs#L180

PR #993, opened 2015-03-22
Introduce more detailled pattern matching for ErrorUnifyingTypes

sleepomeno:
Solving issue #981


paf31:
This looks fine, but I think the better solution would be to change the simplify function.

PR #998, opened 2015-03-24
Remove duplicates from 'overlapping instances' error report

balajirrao:


paf31:
Thanks! Do you have an example of where this change is required?

balajirrao:
No, sorry I don't have an example. I hit this bug when I was trying to make psci give a type to (\x -> x ++ x). It partially succeeded but failed with an overlapping instances error (with duplicates).

paf31:
:+1: Thanks!

balajirrao:
Thanks for merging!

PR #999, opened 2015-03-25
Tidy up MonadError code a bit more

paf31:
@garyb Could you please review? Thanks..

@nicodelpiano You might find this useful for your warnings changes. I intend to chip away at this over the next few days until we can just add `MonadWriter`.

### DIFF
--- a/psc-make/Main.hs
+++ b/psc-make/Main.hs
@@ -52,27 +52,29 @@ readInput InputOptions{..} = do
   content <- forM ioInputFiles $ \inFile -> (Right inFile, ) <$> readFile inFile
   return (if ioNoPrelude then content else (Left P.RebuildNever, P.prelude) : content)
 
-newtype Make a = Make { unMake :: ReaderT (P.Options P.Make) (ExceptT String IO) a }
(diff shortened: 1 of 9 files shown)


paf31:
@nicodelpiano Actually, it should be very simple to add warnings in some places now: desugaring and codegen specifically. The type checker will be possible, but harder. If you're interested, I suggest the following:

1. Change `Check` to `CheckT m`, and update all the code which uses it (maybe very time consuming, but at least you'll become familiar with lots of the typechecker code). You will find that you'll need `MonadError` constraints on some `m`s.
1. Simplify `runCheck`. Remove `interpretMultipleErrors`, it won't be needed any more.
1. Maybe even introduce a `MonadCheck` class, so that the type checker functions become polymorphic, just requiring some `MonadCheck`. 
1. Start adding `MonadWriter` constraints in the typechecker where warnings are needed. Bubble the constraints up to `compile`.

nicodelpiano:
I had started tackling the first two points, so I guess them would be finished in a few days. Thanks!

garyb:
Looks good!

PR #1001, opened 2015-03-25
Reform BoolLike

### LINKED ISSUE #907: Revise `BoolLike`
#900

garyb:
Resolves #907.

### DIFF
--- a/prelude/README.md
+++ b/prelude/README.md
@@ -119,16 +119,16 @@ instance categoryArr :: Category Prim.Function
 ($) :: forall a b. (a -> b) -> a -> b
 ```
 
-Applies a function to its argument
+Applies a function to its argument.
 
 ```purescript
-length $ groupBy productCategory $ filter isInStock products
+length $ groupBy productCategory $ filter isInStock $ products
 ```
 
 is equivalent to
 
 ```purescript
-length (groupBy productCategory (filter isInStock (products)))
+length (groupBy productCategory (filter isInStock products))
 ```
 
 `($)` is different from [`(#)`](#-2) because it is right-infix instead of left, so
@@ -141,16 +141,16 @@ length (groupBy productCategory (filter isInStock (products)))
 (#) :: forall a b. a -> (a -> b) -> b
 ```
 
-Applies a function to its argument
+Applies an argument to a function.
 
 ```purescript
-products # groupBy productCategory # filter isInStock # length
+products # filter isInStock # groupBy productCategory # length
 ```
 
 is equivalent to
 
 ```purescript
-length (groupBy productCategory (filter isInStock (products)))
+length (groupBy productCategory (filter isInStock products))
 ```
 
 `(#)` is different from [`($)`](#-1) because it is left-infix instead of right, so
@@ -190,7 +190,7 @@ class Show a where
 
 The `Show` type class represents those types which can be converted into a human-readable `String` representation.
 
-While not required, it is recommended that for any expression `x`, the string `show x` be executable PureScript code 
+While not required, it is recommended that for any expression `x`, the string `show x` be executable PureScript code
 which evaluates to the same value as the expression `x`.
 
 #### `showUnit`
@@ -264,7 +264,7 @@ types use the type constructor `f` to represent some computational context.
 void :: forall f a. (Functor f) => f a -> f Unit
 ```
 
-The `void` function is used to ignore the type wrapped by a [`Functor`](#functor), replacing it with `Unit` and 
+The `void` function is used to ignore the type wrapped by a [`Functor`](#functor), replacing it with `Unit` and
 keeping only the type information provided by the type constructor itself.
 
 `void` is often useful when using `do` notation to change the return type of a monadic computation:
@@ -311,8 +311,8 @@ class (Apply f) <= Applicative f where
 The `Applicative` type class extends the [`Apply`](#apply) type class with a `pure` function, which can be used to
 create values of type `f a` from values of type `a`.
 
-Where [`Apply`](#apply) provides the ability to lift functions of two or more arguments to functions whose arguments are wrapped using `f`, 
-and [`Functor`](#functor) provides the ability to lift functions of one argument, `pure` can be seen as the function which lifts functions of 
+Where [`Apply`](#apply) provides the ability to lift functions of two or more arguments to functions whose arguments are wrapped using `f`,
+and [`Functor`](#functor) provides the ability to lift functions of one argument, `pure` can be seen as the function which lifts functions of
 _zero_ arguments. That is, `Applicative` functors support a lifting operation for any number of function arguments.
 
 `Applicative` instances should satisfy the following laws:
@@ -362,7 +362,7 @@ where the function argument of `f` is given the name `y`.
 
 - Associativity: `(x >>= f) >>= g = x >>= (\k => f k >>= g)`
 
-Or, expressed using `do` notation: 
+Or, expressed using `do` notation:
 
 - Associativity: `do { z <- do { y <- x ; f y } ; g z } = do { k <- x ; do { y <- f k ; g y } }`
 
@@ -388,7 +388,7 @@ represent type constructors which support sequential composition, and also lifti
 - Left Identity: `pure x >>= f = f x`
 - Right Identity: `x >>= pure = x`
 
-Or, expressed using `do` notation: 
+Or, expressed using `do` notation:
 
 - Left Identity: `do { y <- pure x ; f y } = f x`
 - Right Identity: `do { y <- x ; pure y } = x`
@@ -612,21 +612,7 @@ The `Eq` type class represents types which support decidable equality.
 - Transitivity: if `x == y` and `y == z` then `x == z`
 - Negation: `x /= y = not (x == y)`
 
-`(/=)` may be implemented in terms of `(==)`, but it might give a performance improvement to implement it separately.  
-
-#### `refEq`
-
-``` purescript
-refEq :: forall a. a -> a -> Boolean
-```
-
-
-#### `refIneq`
-
-``` purescript
-refIneq :: forall a. a -> a -> Boolean
-```
-
+`(/=)` may be implemented in terms of `(==)`, but it might give a performance improvement to implement it separately.
 
 #### `eqUnit`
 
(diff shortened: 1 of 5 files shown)


garyb:
Disregard that, I gave `ComplementedLattice` the "relatively complemented" law instead.

paf31:
Looks good to me. A couple of things:

- `BoolLike` can be renamed if we like. `Boolean`?
- Is it worth introducing `Ord <= Bounded` and `BoundedLattice`?
- It might be good to use `meet` and `join`, with aliases for `(&&)` and `(||)`?
- Does the optimizer need changes?

garyb:
1. I did wonder about renaming, `BooleanAlgebra` was my first choice, but there are one or two things that broke in `test-everything` then. I think it may have been `lens` related again though.

2. So we'd have `Ord <= Lattice`, `Ord <= Bounded`, `Lattice, Bounded <= BoundedLattice`? That's fine by me, I mashed `Lattice` and `BoundedLattice` together to cut down on the number of classes required, making the (perhaps erroneous) assumption that most lattices we'd have would be finite. Would `Bounded` have any laws of its own?

3. Good point with the operators. That'll save pain when we do the aliasing stuff. :+1:

4. Yep, it does, I meant to mention that in the opening too.

paf31:
> Would Bounded have any laws of its own?

```purescript
bottom <= x <= top
```

is the only thing I can think of.

garyb:
Updated with `Bounded` etc.

I also put the complemented and relatively complemented laws back in. I'd appreciate an opinion from someone who actually knows what they're talking about in regards to that! Can reword the docs to avoid the quantification symbols, but I just wanted to make it clear what I thought they are, whether they're right, whether we *should* make `ComplementedLattice` relatively complemented, etc

I have a slight concern about using `join` after thinking about this, as we also have `Control.Monad.join` and I think there may be a `String` or `Array` `join` too.

paf31:
Good point. `sup` and `lub`?

I'll read a bit about complemented lattices.

garyb:
Wouldn't it be `sup` and `inf` or `lub` and `glb`?

Can't say I'm crazy about any of those names either ;)

paf31:
Yes, sorry :)

garyb:
Regarding the laws, I think it is right the way I have it now:

- It is not the case that every bounded lattice must be relatively complemented
- A distributive lattice is only complemented iff it is relatively complemented

But stating the latter law may be redundant, as I assume it's not actually possible to define a lattice that satisfies both **distributivity** and **complemented** unless it is relatively complemented.

garyb:
I think I'd prefer to rename the `Monad` `join` that this one actually, this seems the more generic operation, so perhaps the other could become `joinM` - having _operations_ called `inf` and `sup` seems a little confusing.

I've updated `Optimizer` to inline `Boolean` operations again, and added more content to the docs now.

prelude/README.md
```diff
@@ -150,7 +150,7 @@ products # groupBy productCategory # filter isInStock # length
 is equivalent to
 
 ```purescript
-length (groupBy productCategory (filter isInStock (products)))
+length (filter isInStock (groupBy productCategory (products)))
```
paf31: Actually, I think the other one was wrong :) The idea was that `isInStock` applies to a product.

prelude/README.md
```diff
@@ -782,49 +768,169 @@ instance ordArray :: (Ord a) => Ord [a]
...
 ```
 
-The `Bits` type class identifies types which support bitwise operations.
+The `Bounded` type class represents types that are finite partially
+ordered sets.
 
-#### `bitsNumber`
+Instances should satisfy the following law in addition to the `Ord` laws:
+
+- Ordering: `bottom <= top`
```
paf31: It's stronger: `forall x. bottom <= x <= top`

prelude/README.md
```diff
@@ -782,49 +768,169 @@ instance ordArray :: (Ord a) => Ord [a]
...
+```
+
+The `ComplementedLattice` type class represents types that are lattices
+where every member is also complemented.
+
+Instances should satisfy the following law in addition to the
+`BoundedLattice` laws:
+
+- Complemented:
+  - For every `a` in interval `[c, d]` there should be a `b` such that `a || b = top`
```
paf31: What are `c` and `d` here?

prelude/README.md
```diff
@@ -869,6 +975,28 @@ instance semigroupArr :: (Semigroup s') => Semigroup (s -> s')
 
 `(++)` is an alias for `(<>)`.
 
+#### `Bits`
+
+``` purescript
+class Bits b where
```
paf31: Just this one left to fix I guess :)

paf31:
Looks good. Just see my comments, and also we still need to decide on `join` vs `sup` vs whatever. I don't feel too strongly against `sup` and `inf`, since I guess people using them abstractly will probably be familiar with the terms. Actually, I quite like `sup`/`inf`, because the name suggests "bigger"/"smaller", whereas `join`/`meet` don't IMO.

prelude/README.md
garyb: Oh, I didn't actually do this, it must have been a change made in `prelude.purs` at some point and then the docs hadn't been regenerated...

prelude/README.md
garyb: `[c, d]` is just "any interval in the poset", I guess since `c` and `d` aren't used here (unlike the relatively complemented law) we could just say it that way instead?

paf31: Can we just say `not a || a = top` and `not a && a = bottom`?

paf31: What I'm saying is that we don't require _some_ element to have that property, it's `not a` in particular that has it. I guess that's implied though. Maybe I'm missing the point of `[c, d]`, but if you pick `c = d = a = bottom`, then it's not true, right? Do we also need `not bottom = top` and vice versa?

garyb: I got this wrong actually, the interval stuff is only relevant to relatively complemented lattices - the laws according to wikipedia at least are that given an `a` there is some `b` where `a || b = bottom` and `a && b == top`.

If we can use `¬a` instead of `b` then great, is that definitely right?

paf31: I think so, yes.

garyb:
Alright, there we go.

paf31:
Looks good to me. @jdegoes @joneshf Any comments?

paf31:
The coverage decrease must be the new optimizer rules. Any chance you could add a quick test for those please?

garyb:
Eh?

paf31:
Yeah, something's not right there :)

garyb:
I couldn't quite figure out from Coveralls what it though was being missed either - I can see which files it says coverage has decreased in (all of which look suspect!) but not the exact lines.

prelude/prelude.purs
```diff
@@ -702,6 +706,167 @@ module Prelude
...
+  instance boundedBoolean :: Bounded Boolean where
+    top = true
+    bottom = false
+
+  -- | The `Lattice` type class represents types that are partially ordered
+  -- | sets with a supremum (`sup` or `||`) and infimum (`inf` or `&&`).
+  -- |
+  -- | Instances should satisfy the following laws in addition to the `Ord`
+  -- | laws:
+  -- |
```
joneshf: It might make sense to write the laws with `sup` and `inf`.

prelude/prelude.purs
```diff
@@ -702,6 +706,167 @@ module Prelude
...
+
+  -- | The `BoundedLattice` type class represents types that are finite
+  -- | lattices.
+  -- |
+  -- | Instances should satisfy the following law in addition to the `Lattice`
+  -- | and `Bounded` laws:
+  -- |
+  -- | - Identity:
+  -- |   - `a || bottom = a`
+  -- |   - `a && top = a`
```
joneshf: Should Annihiliation be here?

```
a || top = top
a && bottom = bottom
```

jdegoes:
At a high-level, looks really good to me. Gonna clean up the "boolean" hierarchy quite a lot! :+1:

prelude/prelude.purs
```diff
@@ -702,6 +706,167 @@ module Prelude
...
+
+  -- | The `ComplementedLattice` type class represents types that are lattices
+  -- | where every member is also complemented.
+  -- |
+  -- | Instances should satisfy the following law in addition to the
+  -- | `BoundedLattice` laws:
+  -- |
+  -- | - Complemented:
+  -- |   - `not a || a == top`
+  -- |   - `not a && a == bottom`
```
joneshf: Do we need to mention double negation?
```
not >>> not = id
```

joneshf: Also, what about DeMorgan's?

```
not a || not b == not (a && b)
not a && not b == not (a || b)
```

prelude/prelude.purs
```diff
@@ -702,6 +706,167 @@ module Prelude
...
+  -- | - Distributivity: `x && (y || z) = (x && y) || (x && z)`
+  class (Lattice a) <= DistributiveLattice a
+
+  instance distributiveLatticeBoolean :: DistributiveLattice Boolean
+
+  -- | The `BooleanAlgebra` type class represents types that are Boolean
+  -- | algebras, also known as Boolean lattices.
+  -- |
+  -- | Instances should satisfy the `ComplementedLattice` and
+  -- | `DistributiveLattice` laws.
```
joneshf: Actually, you probably need distributativity and negation to prove DeMorgan's. Though it may be derivable free from the other laws?

prelude/prelude.purs
joneshf: Do we need to say something about `top` and `bottom`?

```
not top == bottom
not bottom == top
```

joneshf:
Looks good.

It doesn't have to be this PR, but I think we might be able to add some instances for `Number` and `Unit`.

`-Infinity`/`Infinity` for `Bounded`, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Infinity.

And `Unit` should satisfy all of them I believe.

prelude/prelude.purs
garyb: Yeah, I wasn't sure which way to go with that, but I figured it would be easier to read with the operators. That's also why I listed their equivalents in the comment above, even though they're not part of the class definition itself.

prelude/prelude.purs
garyb: `not >>> not = id` makes sense, but is that implied by the fact it's uniquely complemented anyway? I guess it does no harm to list it. I guess the description should mention that it's uniquely complemented also, even though from the type signature doesn't admit anything else.

garyb:
Good call, I intend to add a bunch of `Unit` instances after this is merged actually.

prelude/prelude.purs
garyb: It has to be complemented _and_ distributive for DeMorgan's doesn't it? So would only apply to `BooleanAlgebra` in this hierarchy, and would be implied by the other laws anyway (I think).

prelude/prelude.purs
garyb: Yep, that makes sense.

garyb:
Ok, updated with a few extra laws, and squashed the commits.

garyb:
Should we just go ahead and remove `Bits` while we're at it, and move it to `purescript-integers`?

jdegoes:
:+1:

paf31:
I was thinking that too. It'll break the optimizations, but that's probably not the end of the world.

Fresheyeball:
Sooo awesome. :+1:

PR #1002, opened 2015-03-25
Fix #365

### LINKED ISSUE #365: Cannot fully eta reduce in instances
I noticed this while working on a collections library.

For example: https://github.com/joneshf/purescript-collections/blob/master/src/Data/Collection.purs#L92 has to have at least one argument. If not, it will generate something like

```javascript
var foldableArray = function (_) {
    return {
        "__superclasses": {}, 
        size: fold(foldableArray({}))(_ps.Prelude["const"](_ps.Prelude["+"](_ps.Prelude.numNumber({}))(1)))(0)
    };
};
```

which just recurses down itself when it attempts to construct the object, since `size` doesn't close over the instance it is defining.

paf31:

### DIFF
--- a/examples/failing/365.purs
+++ b/examples/failing/365.purs
@@ -1,0 +1,11 @@
+module Main where
+
+class C a where
+  f :: a -> a
+  g :: a -> a
+
+instance cS :: C String where
+  f s = s
+  g = f
+
+main = g "Done"
--- a/examples/failing/TypeSynonyms5.purs
+++ b/examples/failing/TypeSynonyms5.purs
@@ -1,0 +1,3 @@
+module Main where
+
+type T = T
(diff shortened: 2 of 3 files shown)


paf31:
`test-everything` passes (as much as before)

garyb:
What does this do, disallow the eta-reduced implementations?

paf31:
Yep exactly. The issue was in strongly connected components of size 1. At some point, they were treated differently, but now they're handled using the defined/undefined code in the typechecker.

garyb:
It's unfortunate we can't eta convert these automatically, but seems like a reasonable fix to me. :+1:

paf31:
I'm not sure. I generally think minimal munging of user code by the compiler is a good thing. You can make a case here because the user doesn't see how the code gets desugared, but I think predictability of the code gen is more important, honestly.

PR #1006, opened 2015-03-27
Remove some JS dependencies related to foreign declarations from core

andyarvanitis:
Changes due to some dependencies on JS in core that I encountered while beginning to rewrite C++ support. The use of `String` in `ExternDeclaration` might be too loose, but I'm not sure how much it matters if the FFI will change soon.

### DIFF
--- a/src/Language/PureScript.hs
+++ b/src/Language/PureScript.hs
@@ -51,6 +51,7 @@ import System.FilePath ((</>))
 import Language.PureScript.AST as P
 import Language.PureScript.Comments as P
 import Language.PureScript.CodeGen as P
+import Language.PureScript.CodeGen.JS as C
 import Language.PureScript.DeadCodeElimination as P
 import Language.PureScript.Environment as P
 import Language.PureScript.Errors as P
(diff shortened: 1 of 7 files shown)


paf31:
@garyb I'll leave this one to you to review, since you know the code the best. Maybe newtyping the String would be good though? Given this construct is likely going away in 0.8, it might not matter too much.

garyb:
Looks basically fine to me, I'd like to move a few things around perhaps to defer the mapping of extern type to `CoreImp` rather than `CoreFn`, but I'll merge this as it is and then have a play. Thanks!

andyarvanitis:
Cool, thanks!

garyb:
Ok, done my tweaks - I introduced a `ForeignCode` newtype even if it is just a temporary thing, and it maintains that type throughout the `Core` steps now, I just remap it to JS in the JS codgen. Does that still work for what you need? I figure that should still work as we can just insert another step in the compilation process if it needs some special handling before going into codegen.

andyarvanitis:
I'm pretty sure that will be fine -- I'll make sure when I get back to my desk in a little while.

andyarvanitis:
:+1:

PR #1010, opened 2015-03-29
Initial work on warnings #803

paf31:

For some reason I can't figure out, it thinks there are shadowed names in Prelude.

/cc @nicodelpiano

### DIFF
--- a/src/Language/PureScript/Renamer.hs
+++ b/src/Language/PureScript/Renamer.hs
@@ -14,6 +14,8 @@
 --
 -----------------------------------------------------------------------------
 
+{-# LANGUAGE FlexibleContexts, ScopedTypeVariables #-}
+
 module Language.PureScript.Renamer (renameInModules) where
 
 import Control.Applicative
@@ -74,13 +76,12 @@ updateScope :: Ident -> Rename Ident
 updateScope i@(Ident name) | name == C.__unused = return i
 updateScope name = do
   scope <- get
-  let name' = case name `S.member` rsUsedNames scope of
-                True ->
-                  let
-                    newNames = [ Ident (runIdent name ++ "_" ++ show (i :: Int)) | i <- [1..] ]
-                    Just newName = find (`S.notMember` rsUsedNames scope) newNames
-                  in newName
-                False -> name
+  name' <- case name `S.member` rsUsedNames scope of
+             True -> do
+               let newNames = [ Ident (runIdent name ++ "_" ++ show (i :: Int)) | i <- [1..] ]
+                   Just newName = find (`S.notMember` rsUsedNames scope) newNames
+               return newName
+             False -> return name
   modify $ \s -> s { rsBoundNames = M.insert name name' (rsBoundNames s)
                    , rsUsedNames  = S.insert name' (rsUsedNames s)
                    }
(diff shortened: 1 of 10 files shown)


paf31:
@nicodelpiano This is a model of what I think we should try to do in the typechecker: get `Check` and `UnifyT` to the point where we can recast things in terms of `CheckT m`: see `RenameT`.

garyb:
Looks good! Do you think we should have a separate type for warnings?

paf31:
I thought about that, but so much of the code would be common (position, module and declaration info, `Monoid` instance, etc.) that I thought it would be best to just share it. It is a little weird that we have one big sum for errors and warnings, but I guess the neat thing is that it's really easy to turn warnings into errors later on if we want to :)

paf31:
Oh, and do you have any idea why the renamer would give three warnings for the Prelude?

```text
Multiple warnings:
  Error in module Prelude:
  Error in value declaration semigroupOrdering:
  Name 'y' was shadowed.

  Error in module Prelude:
  Error in value declaration const:
  Name 'a' was shadowed.

  Error in module Prelude:
  Error in value declaration asTypeOf:
  Name 'x' was shadowed.
```

garyb:
Perhaps after the case changes you updated yesterday to reuse non-generated names where possible?

paf31:
Possibly, I'll spend some more time on this later hopefully and see if that was the cause.

paf31:
@garyb I think this is good to go now. Could you please have a look and see what you think? It found about 10 shadowing warnings in the core libraries which is quite nice.

src/Language/PureScript/Renamer.hs
```diff
@@ -175,7 +177,7 @@ renameInLiteral _ l = return l
 -- Renames within case alternatives.
 --
 renameInCaseAlternative :: CaseAlternative Ann -> Rename (CaseAlternative Ann)
-renameInCaseAlternative (CaseAlternative bs v) =
+renameInCaseAlternative (CaseAlternative bs v) = newScope $
```
paf31: The changes in this module are just leftovers which I can revert for the most part, but I think this needs to be here, yes?

garyb: Yep, that makes total sense. I assume this is where the spurious overlap warnings were coming from?

paf31: No I thought so, but you were right - it was the case desugaring. There are legitimate cases where we cause overlaps, but I figure it's ok, since we know where they happen and can ensure it's safe.

paf31:
@garyb Looks like the core-tests build on master is failing due to `complement` being missing and `purescript-arrays` requiring it, FYI. I think we need to add a dependency on `purescript-integers` before we release `psc`.

paf31:
Let's fix it this weekend if you have some time?

garyb:
Yes, absolutely, I meant to ask about that actually - how would you feel about using `Int` for all the `Number`s in `arrays`?

paf31:
Sounds great :+1:

paf31:
Actually, given that we're now inlining non-Prelude stuff, I'd be fine with moving out `purescript-prelude` too. What do you think?

garyb:
Sounds good to me, but maybe still worth waiting for 0.8? That'll break a lot more stuff than even these changes... I'll catch up with you tomorrow about it if you're around.

paf31:
Makes sense.

garyb:
I haven't forgotten about this, but updating arrays means a mass-dependency-update, so I'm working on the tool for that first.

PR #1011, opened 2015-03-29
Avoid 'WARNING - If this if/for/while really shouldn't have a body, use ...

jacereda:
...{}' when using the Closure Compiler.

### DIFF
--- a/prelude/prelude.purs
+++ b/prelude/prelude.purs
@@ -1304,7 +1304,7 @@ module Control.Monad.Eff
     """
     function untilE(f) {
       return function() {
-        while (!f());
+        while (!f()) {};
         return {};
       };
     }


michaelficarra:
That is a dumb warning. Empty blocks are more empty than empty statements...

Do you have to use closure compiler?

garyb:
> That is a dumb warning.
Haha!

I guess it's more of a lint thing really, strange that closure compiler warns on it.

jacereda:
But it’s easy to forget the semicolon and have something like:

while (foo())
statementIntendedToBeOutOfTheLoop();

Adding an explicit {} is quite common in C land as well, lint tools warn about empty blocks without explicit braces.

I don’t have an immediate need for it, but it should be possible to use the closure compiler and, if possible, without warnings.

garyb:
Oh sure, if this was code you were writing by hand I could understand why it's a warning. I just didn't think linting was closure's responsibility.

jacereda:
It does indeed perform some linting:

https://github.com/google/closure-compiler/blob/8ce967b4c75a6d5f9eef8a9947c8c8b572e1719c/src/com/google/javascript/jscomp/CheckSuspiciousCode.java

https://github.com/google/closure-compiler/tree/8ce967b4c75a6d5f9eef8a9947c8c8b572e1719c/src/com/google/javascript/jscomp/lint

paf31:
Shall I close this then?

garyb:
I don't feel strongly about this either way.

PR #1014, opened 2015-03-30
Fix #1012

### LINKED ISSUE #1012: Regression since 0.6.9.3, NameNotInScope
I'm using master and getting the following when compiling strongcheck:

```
Compiling Test.StrongCheck
Error in module Test.StrongCheck:
Error in value declaration statCheckPure:
Error at ~/slamdata/bower_components/purescript-strongcheck/src/Test/StrongCheck.purs line 119, column 23 - line 121, column 15:
  measure' may not be defined in the current scope
See https://github.com/purescript/purescript/wiki/Error-Code-NameNotInScope for more information, or to contribute content related to this error.
```

https://github.com/purescript-contrib/purescript-strongcheck/blob/master/src/Test/StrongCheck.purs#L119

The function definitely seems defined in the few lines above. PureScript 0.6.9.3 seems to agree and compiles the code just fine.

paf31:

### DIFF
--- a/examples/passing/Let2.purs
+++ b/examples/passing/Let2.purs
@@ -1,0 +1,15 @@
+module Main where
+
+test = 
+  let f :: Number -> Boolean
+      f 0 = false
+      f n = g (n - 1)
+
+      g :: Number -> Boolean
+      g 0 = true
+      g n = f (n - 1)
+
+      x = f 1
+  in not x
+
+main = Debug.Trace.print test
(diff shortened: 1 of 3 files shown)


garyb:
Ah interesting, I thought from your reply to #1012 you meant it was a bug in `strongcheck` before - I see the problem now though. :ship:

garyb:
Coveralls is still being crazy I see.

paf31:
Yeah it might be something to do with the new `psci-tests`.

PR #1016, opened 2015-03-30
Organise prelude, add Unit instances

garyb:
Tried to make the arrangement of instances a bit more consistent and added Unit instances for all kind `*` classes.


paf31:
:+1: This was long overdue, I always end up hunting for things in the Prelude source... :)

PR #1017, opened 2015-03-30
Remove `Bits`

garyb:


paf31:
So there are a couple of libraries which use `Bits`, and one which defines an instance: `purescript-int64` by @freebroccolo, based on a GitHub search. I guess we could either make everything monomorphic, or keep the class elsewhere.

garyb:
Argh. I had made it monomorphic in `Data.Int.Bits`, I'll restore the class.

paf31:
No, no, hold on. I think that's probably fine.

paf31:
I think the question has to be - is there any useful set of laws we can give to `Bits`, or any way in which we can break it up into sensible parts? I think the answer is "no" because of all the corner cases.

freebroccolo:
FWIW, I'm not really maintaining `purescript-int64` at the moment so it's okay with me if you need to remove `Bits`. Thanks for mentioning it though.

paf31:
@freebroccolo Sounds good, thanks. We spoke on IRC about keeping a class, but decided that to do anything useful, we'd need a much better version of the current class, so I think we're going to make things monomorphic for `Int` for now, and maybe add classes later if we figure out what they should be.

paf31:
@garyb so I think this is good to go, right?

garyb:
:+1:

jutaro:
Where has the class gone, has it vanished forever? :rage:

garyb:
The class has, yes, see @paf31's explanation above. The operations are available here however: https://github.com/purescript/purescript-integers/blob/master/src/Data/Int/Bits.purs

It's currently just sitting in `master` though, it will be released as a version once psc v0.6.10 is as well.

paf31:
@jutaro has expressed some unhappiness about this change on IRC, and I think we should probably discuss the alternative before we make a release, since this will be the first time we have removed a JS operator from Prelude, essentially removing an optimization. 

Here are two options:

- "Progress" - `shl` and `shr` use the type `Int` and become more type safe, but bitwise code becomes slower overall until rewrites are added.
- "Preservation" - `Bits` is removed, its functions get put somewhere in `Prelude`, maybe in a new `Data.Bits` module. We plan to extract the code to a library once rewrites are in place. Existing code should continue to work, and will still be optimized, at the expense of type safety.

Any preferences? /cc @garyb @joneshf @jdegoes

garyb:
Another suggestion is to add cases for `purescript-integers` to the optimiser as a stop gap until rewrites arrive, despite it not being in prelude. As @paf31 pointed out, it's still in core at least.

This would probably be my preference for now, as I've already started using `Int` in my code anyway.

paf31:
I'm happy to do that for now as well. As an added bonus, we can add rewrites for the other operators in `integers` too.

garyb:
Ok, cool, I'll work on that now, should be a fairly quick fix.

PR #1018, opened 2015-03-31
Further update prelude docs

garyb:
- Expanded out the laws for `Semiring`, etc. into the same form we use elsewhere.
- Dropped the `do` syntax alternatives for laws where they were used, as they weren't valid PureScript (used `;` separators)
- Rearranged stuff further to hopefully improve the flow


garyb:
It also got me thinking about the orphan instances for `Array` again - we're pretty inconsistent about what's provided in `Prelude`. We have instances for `Eq`, `Ord`, `Show`, yet not for any of the `Control` stuff. Should we fix that?

paf31:
Looks good.

I think the `do` notation is quite useful actually. I find the laws for `Monad` much easier to read that way, and I don't think it's hard to make the connection to valid `do` notation. I would like to support braces and semicolons anyway.

paf31:
For `Array`, I think I'd rather do things in the other direction, long term, even moving `Array` out of `Prim` and into `purescript-arrays`.

garyb:
Yeah, I did wonder about that. I guess that's a “Meltdown” change, as I assume it ties in with the removal or change in pattern matching and `[]` syntax that was/is being discussed.

garyb:
It would be nice to have some way of dealing with kind of flexible syntax for that kind of thing actually - it would also help with the int and char literal problem. `OverloadedStrings` in GHC isn't wonderful, but it's damn useful.

paf31:
> some way of dealing with kind of flexible syntax

Templates?

```purescript
myCharLiteral = [char|x]
myIntLiteral = [int|100]
```

Verbose, but very flexible.

garyb:
Yeah, I meant perhaps something less flexible so it would also be less verbose than just calling a `char` or `int` function. ;) Like the `IsString` stuff.

paf31:
Well one advantage of templates is that they can verify validity at compile time.

`[char|actually a string]` would give a compile time error.

garyb:
Ah, that's a good point.

garyb:
Tests failing as I misplaced `(%)` by accident it would seem. Should we remove it anyway, and promote `purescript-integers` for that usage?

PR #1020, opened 2015-04-01
Fix #968

### LINKED ISSUE #968: tco in transformers
This is not work. 
```purescript 
data TemplateEl = Placeholder String | Key String 
type Template = [TemplateEl]


parseEl :: Template -> P.ParserT String (State (M.StrMap String)) Unit
parseEl template =
  sequence_ $ reverse $ parseEl' template []
  where parseEl' t parsers = 
          case t of
            [] -> parsers
            (Placeholder str):ts -> 
              parseEl' ts ((pure unit <* P.string str):parsers)
            (Key str):ts ->
              -- It says t is [] here
              parseEl' ts ((tempVar >>= \v ->
                             lift $ modify (M.insert str v)):parsers)

parser = parseEl $ [Key "foo", Key "bar"]

main = void $ do 
  fprint $ (runState (runParserT "foo/bar/baz" parser) empty)
``` 
But this is works 
```purescript 
import qualified Array.Unsafe as U

parseEl :: Template -> P.ParserT String (State (M.StrMap String)) Unit
parseEl template =
  sequence_ $ reverse $ parseEl' template []
  where parseEl' t parsers =
          if length t == 0 then
            parsers
          else case U.head t of 
            Key str ->
              let add :: P.ParserT String (State (M.StrMap String)) Unit
                  add = P.try (tempVar >>= \v -> lift $ modify (M.insert str v))
              in parseEl' (U.tail t) (add:parsers)
            Placeholder str ->
              let add :: P.ParserT String (State (M.StrMap String)) Unit
                  add = P.string str *> return unit
              in 
               parseEl' (U.tail t) (add:parsers)
```
psc version - 0.6.8

Is it a bug, or I'm doing something wrong?

paf31:
@garyb @cryogenian Could you please review?

This reminds me actually - I think I'd like to just remove TCO completely, and force people to use things like `map`, `fold` and `tailRec` which are far more predictable. Thoughts?

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
+++ b/src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
@@ -15,6 +15,8 @@
 
 module Language.PureScript.CodeGen.JS.Optimizer.TCO (tco) where
 
+import Data.Monoid
+
 import Language.PureScript.Options
 import Language.PureScript.CodeGen.JS.AST
 
@@ -30,10 +32,13 @@ tco' = everywhereOnJS convert
   where
   tcoLabel :: String
   tcoLabel = "tco"
+
   tcoVar :: String -> String
   tcoVar arg = "__tco_" ++ arg
+
   copyVar :: String -> String
   copyVar arg = "__copy_" ++ arg
+
   convert :: JS -> JS
   convert js@(JSVariableIntroduction name (Just fn@JSFunction {})) =
     let
@@ -46,6 +51,7 @@ tco' = everywhereOnJS convert
               JSVariableIntroduction name (Just (replace (toLoop name allArgs body')))
         | otherwise -> js
   convert js = js
+
   collectAllFunctionArgs :: [[String]] -> (JS -> JS) -> JS -> ([[String]], JS, JS -> JS)
   collectAllFunctionArgs allArgs f (JSFunction ident args (JSBlock (body@(JSReturn _):_))) =
     collectAllFunctionArgs (args : allArgs) (\b -> f (JSFunction ident (map copyVar args) (JSBlock [b]))) body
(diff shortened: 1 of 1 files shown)


paf31:
Actually, scratch this, it stops far too many things from being TCO'd. Let's discuss the no-more-TCO option, please.

andyarvanitis:
From a backend impl perspective, the TCO logic is currently a bit of a pain since it's down in the lang-specific parts (at least for now). However, it is handy for end-users getting some out-of-the-box performance/non-crashing.

paf31:
My concern is that it's currently completely unpredictable when TCO will kick in.

Maybe we could turn the TCO optimizer into a warning - any self-recursive function could generate a message like "maybe you should consider using `purescript-tailrec`".. I think beginners should be encouraged to use higher-level primitives anyway.

andyarvanitis:
Right. A warning sounds like a good and interesting idea.

andyarvanitis:
Also, the fact that PureScript, unlike something like GHC, is very close to the semantics of the backend language, seems to point in that direction philosophically.

paf31:
Yes, one thing we talk about quite a lot is "predictability of codegen", or at least I do :), and I really think TCO flies in the face of that.

I should have a patch for this in a few mins though, so we can get it working again at least.

paf31:
Ok, this _should_ be good now. Again, I think it's fine as a bandaid, but maybe this is something we can reconsider for "Meltdown".

cryogenian:
:+1: for warnings and pointing to `fold` `tailRec` etc.

garyb:
> From a backend impl perspective, the TCO logic is currently a bit of a pain since it's down in the lang-specific parts (at least for now).

I had started working on moving that to apply to `CoreImp` along with all the other things currently in the JS optimiser, but it's not quite there yet.

> My concern is that it's currently completely unpredictable when TCO will kick in.

Is that really true? Like, are there any cases when a recursive call in tail position is not TCO'd? People not understanding when exactly that is the case is one thing, but I'd not characterise it as being completely unpredictable because of that.

I'm not sure how I feel about dropping TCO. It definitely was an attractive feature to me, and something that even the likes of [Babel](https://babeljs.io) provides now. Writing recursive functions is something very natural to functional programming, even if it can often be avoided with other combinators.

ES6 is supposed to have proper tail calls (I think... certainly TCO at a minimum), which is a whole other can of worms.

paf31:
TCO is currently difficult to predict for

- Case statements
- `Eff`

The reason is generally that TCO can be triggered by other optimizations, and it matters what order they happen in. Sometimes, TCO will be blocked by the inlining of `bindEff`, or just because of the code generated for a `case`.

Maybe imperative core will help the situation. Does impcore have explicit cases or are they already desugared in that representation?

TCO is probably my least favourite module of the PS compiler code :) Honestly, I'd be happy to see it go in 0.8, encouraging people to use `tailrec` instead (which gives us a better way to support new backends IMO, support for TCO in more monads, 100% predictability and the ability to do static analysis), but I can see a case for keeping it.

paf31:
Also, how does this PR look? :) I'll merge it as a stopgap solution if it looks ok.

PR #1025, opened 2015-04-03
Inline native int and bitwise operators

garyb:


andyarvanitis:
Have you guys ever discussed having typeclass for types that would inline these (and maybe other) operators, instead of doing it by instances?

paf31:
This looks good to me.

@andyarvanitis I'm not sure I follow. Do you mean a typeclass in Haskell, or in PureScript?

andyarvanitis:
Sorry, I meant in PureScript, since right now we check for `semiringNumber`, `semiringInt`, etc. in the inliner. My comment was strictly off-the-cuff, since I haven't thought through the implementation.

garyb:
Hopefully we'll be ditching this explicit operator inlining stuff anyway in favour of rewrite rules, I guess that's kind of similar to what you're saying? I assume the idea is to make it PS-directed rather than explicit in the Haskell anyway?

andyarvanitis:
Something allowing creators of new types to have the operators inlined, without adding new syntax to the language. Maybe like a "magic" empty typeclass (instead of "magic" instances):
```PureScript
class InlinedOperators a where
...
instance inlinedOpsNumber :: InlinedOperators Number where
instance inlinedOpsFoo :: InlinedOperators Foo where
...
```
Anyway, like I said, I'd have to think about where you would have the relevant info in the complier to do it, without major changes.

Also, I confess I haven't looked into rewrite rules, so maybe it supersedes this general concept, which is hacky anyway.

paf31:
That's a nice idea. I generally approve of using the type system to indicate things like this.

The issue is that the "inliner" isn't really an inliner right now. It just inlines a handful of functions, and does it by explicitly rewriting the AST in each case :(

But this could be possible.. and very elegant. GHC has a magic "inline" function if I remember correctly. It's a little like that, but actually typed.

PR #1029, opened 2015-04-05
Inline integer divisions

garyb:
`x / y` becomes `(x / y) | 0`


paf31:
:+1:

michaelficarra:
`| 0` doesn't just floor. It also converts to int32. This may break for large numbers.

garyb:
I don't think that's a problem as long as we document it clearly. We've also moved bitwise operations to be monomorphic on `Int`, so it probably makes sense that it's restricted to `Int32`.

garyb:
Thanks for the reminder though! I had totally forgotten about that.

reuben364:
Note that `(x / y) | 0` is not the same as floor division, as numbers round towards zero rather then in the negative direction.

reuben364:
But together with the behavior of (%) it forms a valid ModuloSemiring, I think.

PR #1032, opened 2015-04-05
Ensure all integer ops are in the Int32 range

garyb:
Also inlines `mod` now I've checked that `(%)` does actually give us the right behaviour for ints.


paf31:
Seems good to me. @michaelficarra, any comments?

garyb:
Made the same change in `purescript-integers` here: purescript/purescript-integers@e0d71323854185bb48cb6827d079cbee8aea6851 but will hold off releasing it as 0.1.0 until we hear from @michaelficarra.

michaelficarra:
Both LGTM.

garyb:
Thanks!

Issue #979, opened 2015-03-18
`psc-make` runs out of memory rather than returning proper error


pseudonom:
With this code:

```purescript
module Foo where

import qualified Data.Map as M
import Data.Tuple

norm :: forall a. (Ord a) => [Tuple a Number] -> [Tuple a Number]
norm = M.toList <<< M.fromListWith (+)

newtype Dist a = Dist [Tuple a Number]

instance semiringDist :: (Semiring a, Ord a) => Semiring (Dist a) where
  (+) x y = Dist []
  (*) x y = Dist []
  one = Dist []
  zero = Dist []
instance moduloSemiringDist :: (ModuloSemiring a, Ord a) =>
         ModuloSemiring a where
  (/) x y = Dist []
  mod x y = Dist []
```

the `moduloSemiring` instance is erroneously for `a` instead of `Dist a`. 

Rather than point out this error, `psc-make` grinds away for awhile and eventually throws `psc-make: out of memory (requested 1048576 bytes)`. 

Strangely, each element in the above code seems necessary for the problem to present.

paf31:
This should be fixed by my changes which terminate the search after a finite number of steps. The problem is that the superclass search for `moduloSemiringDist` will just spin.

pseudonom:
Is that fix in 0.6.9? Because I'm getting the problem in 0.6.9.

paf31:
Nope, not yet.

PR #1033, opened 2015-04-06
Fix #438 and #979

### LINKED ISSUE #438: Self referencing super class constraint stalls compiler
This little compiler makes the PureScript compiler hang:

````haskell
module Main where
data Fix f = In (f (Fix f))
instance eqFix :: (Eq (f (Fix f))) => Eq (Fix f) where
  (==) (In f) (In g) = false -- f == g
  (/=) a b = true -- not (a == b)
main = In [] == In []
```

Probably due to the self reference in the super class constraint for `Fix`. It would be useful to be able to express constraints like this.

Note that in Haskell you would need the `UndecidableInstances` extension to be able to write this, but it would work.

paf31:
@garyb Could you please review? Thanks.

### DIFF
--- a/examples/failing/438.purs
+++ b/examples/failing/438.purs
@@ -1,0 +1,9 @@
+module Main where
+      
+data Fix f = In (f (Fix f))
+
+instance eqFix :: (Eq (f (Fix f))) => Eq (Fix f) where
+  (==) (In f) (In g) = f == g
+  (/=) a b = not (a == b)
+        
+example = In [] == In []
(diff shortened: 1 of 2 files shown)


paf31:
I tried using `LogicT` here as well, which I thought might give better performance, but it didn't help enough to justify the additional dependency IMO.

garyb:
Looks good to me. Do you think we need a different error message if we bail out vs actually exhausting all possibilities? I guess a depth of 1000 is a pretty safe assumption to work to though, so maybe that's unnecessary.

paf31:
I originally had it that way and it's probably better. I'll need to wedge an `ExceptT` inside `StateT []` somewhere, which is fine. 

I need to test this some more when `test-everything` builds again.

Issue #765, opened 2014-12-12
Ability to define binary operators infixed


Fresheyeball:
Requesting that both of the following syntax be allowed:

```
(<$>) f (Foo a) = Foo (f a)
f <$> Foo a = Foo (f a)
```

hdgarrood:
Can this be closed, now that we've decided on operator alises?

paf31:
Yes, I think so.

PR #1035, opened 2015-04-07
[#765] - allow binary operators to be defined infix.

stevejb71:
I couldn't find starter-kit (as mentioned in the guidelines for contributing) so I didn't run it.

The error message if an infix operator is defined with not exactly two parameters is not very nice. Otherwise it seems to work. 
It's my first attempt at touching compiler code, and I'm no Haskell expert, so I'd be interested to know what's wrong with this PR, and I'd hope learn to contribute.

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -91,25 +91,38 @@ parseTypeSynonymDeclaration =
                          <*> (indented *> equals *> noWildcards parsePolyType)
 
 parseValueDeclaration :: TokenParser Declaration
-parseValueDeclaration = do
+parseValueDeclaration = P.try parseValueDeclarationWithIdentFirst <|> parseValueDeclarationWithInfixOperator
+
+parseValueDeclarationWithIdentFirst :: TokenParser Declaration
+parseValueDeclarationWithIdentFirst = do
   name <- parseIdent
   binders <- P.many parseBinderNoParens
-  value <- Left <$> (C.indented *>
-                       P.many1 ((,) <$> parseGuard
-                                    <*> (indented *> equals *> parseValueWithWhereClause)
-                               ))
-       <|> Right <$> (indented *> equals *> parseValueWithWhereClause)
+  value <- parseValueDeclarationValue
   return $ ValueDeclaration name Value binders value
-  where
-  parseValueWithWhereClause :: TokenParser Expr
-  parseValueWithWhereClause = do
-    value <- parseValue
-    whereClause <- P.optionMaybe $ do
-      C.indented
-      reserved "where"
-      C.indented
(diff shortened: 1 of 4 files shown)


stevejb71:
Ahh - this doesn't compile (needs brackets around Foo)

data Foo = Foo Number
f <$$$> Foo a = Foo (f a)

garyb:
I'd perhaps organise things a little differently:

``` haskell
parseValueDeclaration :: TokenParser Declaration
parseValueDeclaration = do
  (name, binder) <- P.try identFirst <|> infixOperator
  ValueDeclaration name Value binders <$> value
  where

  identFirst :: TokenParser (Ident, [Binder])
  identFirst = (,) <$> parseIdent <*> P.many parseBinderNoParens

  infixOperator :: TokenParser (Ident, [Binder])
  infixOperator = do
    lhs <- parseBinderNoParens
    binaryOp <- parseOp
    rhs <- parseBinderNoParens
    return (binaryOp, [lhs, rhs])

  value :: TokenParser (Either [(Guard, Expr)] Expr)
  value =
    Left <$> (C.indented *>
                P.many1 ((,) <$> parseGuard
                             <*> (indented *> equals *> parseValueWithWhereClause)
                        )) <|>
    Right <$> (indented *> equals *> parseValueWithWhereClause)
    where parseValueWithWhereClause = do
            value <- parseValue
            whereClause <- P.optionMaybe $ do
              C.indented
              reserved "where"
              C.indented
              C.mark $ P.many1 (C.same *> parseLocalDeclaration)
            return $ maybe value (`Let` value) whereClause
```

(That's totally untested but you get the gist)

For `f <$$$> Foo a = Foo (f a)` you'll probably need a special case of `parseBinderNoParens`.

garyb:
Maybe just `parseBinder` would work actually.

stevejb71:
Thanks for the feedback!!
The above doesn't quite work:
```identFirst``` parses ```a <$> b = 0``` as ```(a,[])```
Moving ```infixOperator``` to be first in the try means ```test -1 = 0``` is seen as a definition of ```-```
So I think the parsing of the value needs to go inside the try. Would that be a problem?
```parseBinder``` does indeed work for ```infixOperator```.

File: src/Language/PureScript/Parser/Declarations.hs [hunk 1]
```diff
@@ -91,25 +91,38 @@ parseTypeSynonymDeclaration =
                          <*> (indented *> equals *> noWildcards parsePolyType)
 
 parseValueDeclaration :: TokenParser Declaration
-parseValueDeclaration = do
+parseValueDeclaration = P.try parseValueDeclarationWithIdentFirst <|> parseValueDeclarationWithInfixOperator
```
paf31: We need to be careful that this doesn't affect the quality of the error messages too much...

puffnfresh:
I could really use this feature!

PR #1037, opened 2015-04-07
Better Show instance for SourcePos/SourceSpan

hdgarrood:
Use the derived Show instance and define separate functions for
displaying SourcePos/SourceSpan values.


paf31:
:+1:

PR #1039, opened 2015-04-10
Add mod name to constraint dict names and pass type information to IsTagOf

andyarvanitis:
Needed by the refactored C++11 backend. This obviously isn't the ideal mechanism to get this information to the backend, but it's very simple, doesn't affect anything else, and works for now. Even providing the module name in the proper place (within the outgoing qualified name) causes some problems in PureScript/JS. I'd definitely like to revisit this when time permits.

### DIFF
--- a/src/Language/PureScript/CoreFn/Desugar.hs
+++ b/src/Language/PureScript/CoreFn/Desugar.hs
@@ -149,8 +149,8 @@ moduleToCoreFn env (A.Module coms mn decls (Just exps)) =
   binderToCoreFn ss com (A.VarBinder name) =
     VarBinder (ss, com, Nothing, Nothing) name
   binderToCoreFn ss com (A.ConstructorBinder dctor@(Qualified mn' _) bs) =
-    let (_, tctor, _, _) = lookupConstructor env dctor
-    in ConstructorBinder (ss, com, Nothing, Just $ getConstructorMeta dctor) (Qualified mn' tctor) dctor (map (binderToCoreFn ss []) bs)
+    let (_, tctor, ty, _) = lookupConstructor env dctor
+    in ConstructorBinder (ss, com, Just ty, Just $ getConstructorMeta dctor) (Qualified mn' tctor) dctor (map (binderToCoreFn ss []) bs)
   binderToCoreFn ss com (A.ObjectBinder bs) =
     LiteralBinder (ss, com, Nothing, Nothing) (ObjectLiteral $ map (second (binderToCoreFn ss [])) bs)
   binderToCoreFn ss com (A.ArrayBinder bs) =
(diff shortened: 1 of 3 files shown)


garyb:
Can you not get this info from a `Meta` annotation? I should probably make it available that way if not, as it seems like a useful thing for backends to have access to in general.

andyarvanitis:
Right that would be the proper place to get it, but currently it's just `Nothing` for any type of dictionary.

garyb:
Hi Andy.

Sorry, I'd kinda forgotten about this. I've put my work on hold for the `CoreImp` stuff as it won't be going in to the 0.7 release now. Once that's sorted I'll get back to it and fix up the `Meta` annotation. Do you want me to merge this into my branch anyway, or can you just keep this change in your branch for now?

andyarvanitis:
Hey Gary,
No problem at all. That's ok, I can just keep this in my branch, since it's just temporary anyway it `Meta` support will be added.
I'm curious about `core-imp` not going into 0.7 though -- is it just a matter of not having time, or is there something recently up for discussion?

garyb:
Yeah it's down to time basically - we've made some biggish changes to the Prelude, and want to get them out there, but as they're going to break a lot of things it seems sensible to release them as part of a major-ish release. Part of that is moving Prelude out into its own library so this won't be a problem in the future.

Pushing `core-imp` back to 0.8 isn't so bad - we need to overhaul the FFI to truly support multiple back-ends, and 0.8 is where we intend to make a lot of major changes like that.

andyarvanitis:
Makes sense. Ok, I'll continue to work off of `core-imp` for now. I doubt I'll attempt to merge recent `master` into it, but if I do and it's not a mess to merge, I'll let you know in case you want to pull it in.

paf31:
I think the versioning scheme is probably going to change. We previously used 0.* to mean "we did a bunch of work, things are stable, and now we want to announce it", but I think we're going to move to something more like semver. So I wouldn't read too much into the 0.7 release stuff, honestly.

As for the change, it seems fine, but how does it get rendered? Won't all the dots have to be escaped?

andyarvanitis:
Regarding the escaped dots, yeah, they get escaped downstream, like the `__superclass_` names (which also carry their module name). So with this change you will end up with e.g. `__dict_Prelude$dotShow_17` in the generated code. I can try to make time to prototype/implement changes to `Meta` do to this properly. It would clean up some other dubious parts in my stuff too.

andyarvanitis:
Well, I've added a couple more changes, so maybe pulling these in would be best after all (if they look ok and you don't mind). There might be more after that, but I'm guessing they'll be small like these.

andyarvanitis:
Superseded by #1093

Issue #747, opened 2014-12-02
Instances for private classes and types still appear in psc-docs output


garyb:
Probably need to make a check while elaborating exports in name desugaring.

paf31:
The environment gets passed to `psc-docs`. Maybe we could treat this like we treat data constructors?

garyb:
Yeah, it looks like we'll have to do something with the environment anyway - instances aren't included in the elaborated exports list after all, so they must be coming through the environment after typechecking or something.

paf31:
Sorry, I don't know what I was referring to there. The environment isn't involved in `psc-docs`, just the exports list (not elaborated). In `psc-pages`, the `isExported` check seemed to work fine. What's an example of this issue?

hdgarrood:
Here's one: https://github.com/hdgarrood/purescript-sequences/blob/master/docs/Data.Sequence.md

`Size` and `Elem` are defined in `Data.Sequence` and not exported, but all of their instances appear in the docs.

garyb:
Thanks, that'll make a handy test case.

PR #1041, opened 2015-04-13
Integrate some of the work done in psc-pages

hdgarrood:
Some context:

* We want to be able to render documentation in at least 3 different forms: Markdown (`psc-docs`), HTML (`psc-pages`), and as an input file for Hoogle (not really "documentation", but the process is very similar). The code I recently contributed to psc-pages splits the rendering into two stages, which makes it easy to render documentation in different forms in this way. In fact psc-pages can currently produce both HTML and ~~Markdown~~ Hoogle.
* We want to fix bugs like #747, and also only have to do it once.
* We eventually want users to be able to run a tool locally which will generate HTML documentation and Hoogle input files, and then publish the results on the web.

It seems like the best way of accomplishing these 3 goals is to start looking at moving chunks of `psc-pages` into the library part of the `purescript` package. This PR is a step in this direction.

One small change I've made in this code that's not in the current `psc-pages` code, or even in the code in the open PR I have on `psc-pages`, is that the `RenderedDeclaration` etc types now have comments as `String` rather than `Html` - the code is the same, it just omits the final step transforming the comments into an HTML element. This allows us to output comments as Markdown, like `psc-docs` currently does, or as HTML.

Also, I decided to base off of `0.7-dev`, since #747 is marked for 0.7.0, but I can change this easily if you'd like.

### DIFF
--- a/src/Language/PureScript/Docs/AsMarkdown.hs
+++ b/src/Language/PureScript/Docs/AsMarkdown.hs
@@ -1,0 +1,100 @@
+{-# LANGUAGE RecordWildCards #-}
+
+module Language.PureScript.Docs.AsMarkdown (
+  renderModulesAsMarkdown
+) where
+
+import Control.Monad.Writer
+import Data.Foldable (for_)
+import Data.List (partition)
+
+import qualified Language.PureScript as P
+
+import Language.PureScript.Docs.Types
+import Language.PureScript.Docs.RenderedCode
+import qualified Language.PureScript.Docs.Render as Render
+
+-- |
+-- Take a list of modules and render them all in order, returning a single
+-- Markdown-formatted String.
+--
+renderModulesAsMarkdown :: [P.Module] -> String
+renderModulesAsMarkdown =
+  runDocs . modulesAsMarkdown . map Render.renderModule
+
+modulesAsMarkdown :: [RenderedModule] -> Docs
+modulesAsMarkdown ms = do
+  headerLevel 1 "Module Documentation"
+  spacer
+  mapM_ moduleAsMarkdown ms
+
+moduleAsMarkdown :: RenderedModule -> Docs
+moduleAsMarkdown RenderedModule{..} = do
+  headerLevel 2 $ "Module " ++ rmName
+  spacer
+  for_ rmComments tell'
+  mapM_ declAsMarkdown rmDeclarations
+  spacer
+
+declAsMarkdown :: RenderedDeclaration -> Docs
+declAsMarkdown RenderedDeclaration{..} = do
+  headerLevel 4 (ticks rdTitle)
+  spacer
+
+  let (instances, children) = partition ((==) ChildInstance . rcdType) rdChildren
+  fencedBlock $ do
+    tell' (codeToString rdCode)
+    zipWithM_ (\f c -> tell' (childToString f c)) (True : repeat False) children
+  spacer
+
+  unless (null instances) $ do
+    headerLevel 5 "Instances"
+    fencedBlock $ do
+      mapM_ (tell' . childToString False) instances
+    spacer
+
+  for_ rdComments tell'
+
+codeToString :: RenderedCode -> String
+codeToString = outputWith elemAsMarkdown
+  where
+  elemAsMarkdown (Syntax x)  = x
+  elemAsMarkdown (Ident x)   = x
+  elemAsMarkdown (Ctor x _)  = x
+  elemAsMarkdown (Kind x)    = x
+  elemAsMarkdown (Keyword x) = x
+  elemAsMarkdown Space       = " "
+
+childToString :: Bool -> RenderedChildDeclaration -> String
+childToString isFirst RenderedChildDeclaration{..} =
+  case rcdType of
+    ChildDataConstructor ->
+      let c = if isFirst then "=" else "|"
+      in  "  " ++ c ++ " " ++ codeToString rcdCode
+    ChildTypeClassMember ->
+      "  " ++ codeToString rcdCode
+    ChildInstance ->
+      codeToString rcdCode
+
+type Docs = Writer [String] ()
+
+runDocs :: Docs -> String
+runDocs = unlines . execWriter
+
+tell' :: String -> Docs
+tell' = tell . (:[])
+
+spacer :: Docs
+spacer = tell' ""
+
+headerLevel :: Int -> String -> Docs
+headerLevel level hdr = tell' (replicate level '#' ++ ' ' : hdr)
+
+fencedBlock :: Docs -> Docs
+fencedBlock inner = do
+  tell' "``` purescript"
+  inner
+  tell' "```"
+
+ticks :: String -> String
+ticks = ("`" ++) . (++ "`")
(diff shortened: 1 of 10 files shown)


hdgarrood:
Oh, also, to clarify - this code gets us part of the way towards fixing #747 but not all of the way there. It still needs to be wired up to `psc-docs`, and even then it doesn't handle every case. For example, if you have a module which:
* defines and exports a type class
* defines and does not export a data type
* makes the data type an instance of the type class,

then the data type will still incorrectly appear in the list of instances for the type class. I think the solution would involve putting more information into the `IntermediateDeclaration` type in order to be able to check that none of the data types and type classes involved in an instance declaration are both defined in the current module, and also not exported.

hdgarrood:
Oh, actually, maybe #747 should be fixed inside the compiler? This code already calls `P.exportedDeclarations` to get a filtered declaration list, so arguably instances containing private classes or types shouldn't be in that list at all?

Come to think of it, on a similar note, perhaps `P.exportedDeclarations` should additionally look deeper through the AST, and filter out things like unexported data constructors and type class members as well? As far as I can tell, the `exportedDeclarations` function is only used inside psci's completion, psc-docs, and psc-pages, so I don't think it would break anything to change that. This change would be really nice from the point of view of tools like psci and psc-pages, since we wouldn't have to pass a `[P.DeclarationRef]` around all the time and make a bunch of separate calls in different places to avoid accidentally including private things.

paf31:
Looks great, but I think I'd rather put the code under the `psc-docs` source directory.

Also, I'm not sure what the best plan of attack is w.r.t. `0.7-dev` right now. Hopefully we can get it all merged soon, and then I could include this in the 0.7 release.

paf31:
Oh also, do you have plans to use this in `psc-docs`? I assume so, and that then `psc-pages` will just be subsumed by `psc-docs`?

hdgarrood:
I had assumed that this code would also become part of the library section of the `purescript` package, which is why I included it in `src`, but come to think of it, I'm not sure it really needs to be used anywhere other than `psc-docs` - so I'll move it, as you say.

Yes, I was planning to use this in `psc-docs`, and then that should subsume `psc-pages`.

hdgarrood:
This now seems to work, with the code from `psc-pages` being used to render markdown output. Example: https://github.com/hdgarrood/purescript-sequences/blob/master/docs/Data.FingerTree.md

The main difference I've noticed is that the generated code now has some redundant brackets in type signatures, where it didn't have them before. (The HTML output in psc-pages has this too; I hadn't noticed!) I suppose we might want to use `pattern-arrows` to sort them out, in `psc-docs/RenderedCode.hs`?

Also, as it is currently, instances' docs (if any exist) are no longer displayed, based on the discussion we had in paf31/psc-pages#15, but if you only wanted that behaviour in `psc-pages` I can revert it easily.

hdgarrood:
I've now tested this with `purescript-sequences` and `purescript-random` and I think it's good to go.

paf31:
Cool, I'll review this today. Could you please rebase it though?

src/Language/PureScript/Docs/AsMarkdown.hs
```diff
@@ -0,0 +1,93 @@
+{-# LANGUAGE RecordWildCards #-}
+
+module Language.PureScript.Docs.AsMarkdown where
```
paf31: Can the exports be refined at all?

src/Language/PureScript/Docs/AsMarkdown.hs
```diff
@@ -0,0 +1,93 @@
...
+import Data.Foldable (for_)
+import Data.List (partition)
+
+import qualified Language.PureScript as P
+
+import Language.PureScript.Docs.Types
+import Language.PureScript.Docs.RenderedCode
+import qualified Language.PureScript.Docs.Render as Render
+
+renderModules :: [P.Module] -> String
```
paf31: Could you please add a Haddock comment to exported things too?

src/Language/PureScript/Docs/Types.hs
```diff
@@ -0,0 +1,49 @@
+
+module Language.PureScript.Docs.Types
```
paf31: This module seems to be defined twice, in `psc-docs` and here.

hdgarrood:
Sure, no problem. Should I squash all the commits?

src/Language/PureScript/Docs/AsMarkdown.hs
hdgarrood: Oh of course, sorry - will do.

paf31:
Yes please. I guess you would rebase from `0.7-dev` right?

src/Language/PureScript/Docs/Types.hs
hdgarrood: Whoops - I meant to remove the one in `psc-docs`. I've done that now.

src/Language/PureScript/Docs/AsMarkdown.hs
paf31: No worries. I'm just trying to check off a few things like this with new modules.

hdgarrood:
Yeah, or I can rebase from `master` if that's easier

paf31:
No, `0.7-dev` is good. I'll merge it there.

puffnfresh:
I didn't know about psc-pages before. Great idea!

The Hoogle compatible stuff could be very useful :+1:

paf31:
@puffnfresh Yep, the plan is to integrate Pursuit with Hoogle for @hdgarrood's GSOC project :)

hdgarrood:
Rebased and squashed, and I've also added explicit export lists and Haddock docs where appropriate. I think this is good to go.

src/Language/PureScript/Docs/RenderedCode/Render.hs
```diff
@@ -0,0 +1,172 @@
+
+-- | Functions for producing RenderedCode values from PureScript Type values.
+
+module Language.PureScript.Docs.RenderedCode.Render (
```
paf31: Maybe later we can try to unify this with the existing `prettyPrintType` function. We could maybe write a "plain text renderer".

paf31:
Looks good. I'll merge this later when I do some work on 0.7.

src/Language/PureScript/Docs/RenderedCode/Render.hs
hdgarrood: I thought about doing this. Composing `renderType` with `codeToString` from the `AsMarkdown` module gets you most of the way there, there are just a couple of things that are not handed (eg `Skolem`).

paf31: Just a word of warning - `pattern-arrows` will spin if a case isn't handled, effectively adding parens forever, so you'll want to make sure your base cases are exhaustive :)

hdgarrood:
Arrghh, I messed up the rebase and somehow not all the changes I made to `psc-docs` were picked up. I'll redo them.

hdgarrood:
I did some more work building on this, and now (unfortunately) I think it's now in a position where if this PR is merged, my other work won't apply cleanly. So I've submitted it in a separate PR instead (#1055).

PR #1044, opened 2015-04-14
Constructor names in pattern match errors

puffnfresh:
We now get a more useful error message when pattern matching fails.

Given a partially matching function like:

    example :: Ordering -> Unit
    example LT = unit

Supplying `GT` now crashes with:

    Error: Failed pattern match: GT

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -256,7 +256,7 @@ moduleToJs (Module coms mn imps exps foreigns decls) = do
     jss <- forM binders $ \(CaseAlternative bs result) -> do
       ret <- guardsToJs result
       go valNames ret bs
-    return $ JSApp (JSFunction Nothing [] (JSBlock (assignments ++ concat jss ++ [JSThrow $ JSUnary JSNew $ JSApp (JSVar "Error") [JSStringLiteral "Failed pattern match"]])))
+    return $ JSApp (JSFunction Nothing [] (JSBlock (assignments ++ concat jss ++ [JSThrow $ failedPatternError valNames])))
                    []
     where
       go :: [String] -> [JS] -> [Binder Ann] -> m [JS]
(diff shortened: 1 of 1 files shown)


puffnfresh:
As of ES6, `.constructor.name` is specified and was pretty standard amongst implementations before ES6 but < IE9 will print:

    Error: Failed to pattern match: undefined

puffnfresh:
We could generate code like so:

```js
GT.name = "GT";
```

If IE 6, 7 & 8 support is necessary for this error message.

garyb:
I think we're standardising on ES5 so <IE9 is out anyway, without a shim. Since this is a error you'd hope would be picked up during development I'd not be too concerned about the `undefined` in there personally.

puffnfresh:
@garyb that's right, I don't think we should be too concerned since this should never be code hit in production.

But @hdgarrood just pointed out that IE10 and IE11 don't even support the property:

https://kangax.github.io/compat-table/es6/#function_name_property_function_expressions

paulyoung:
In my experience, using `constructor.name` is generally unreliable for the reasons mentioned in https://github.com/arsnebula/classx/pull/2.

I've mostly been bitten by it in minified code or in attempts to set it manually that didn't work.

paf31:
I agree with @garyb. It's an error anyway, and we can't possibly make things worse that they are now, but this has big potential benefits if we _can_ display the constructor name. Let me just figure out where to merge this.

paf31:
Just thinking about this a little more. What do you think about using the `debugger` statement in the event of a pattern match failure? If we assume partiality failures shouldn't happen (we don't want people to catch the exception, etc.) then I think that would give us the best way to move those issues during development.

michaelficarra:
:+1: debugger statement

garyb:
I like it :+1:

hdgarrood:
:+1: debugger is my favourite suggestion so far too

puffnfresh:
One of the problems with debugger is that it just gets ignored by default under node.js - you won't know you've hit broken code without specifically running under `node debug`.

paulyoung:
Same thing in the browser, `debugger` only works if dev tools are open.

garyb:
Could keep the `throw` right after `debugger` then?

puffnfresh:
What's the value for that? Just set break on exceptions and you have the same behaviour.

garyb:
Good point.

michaelficarra:
I forgot about break on uncaught exceptions. Does the node debugger do that too, or is that just available in dev tools?

paf31:
Good point, I didn't think about that.

But it brings up the question - if we can break on exceptions, is that enough in itself? Can we just break and inspect the unmatched values?

paf31:
By the way, I'll almost certainly merge this before we release 0.7. I just want to make sure we pick the best approach.

puffnfresh:
I don't always have control over whether the `debug` flag is used for the `node` executable. I think this error is useful in that case, even if it's not as useful as debugging.

paf31:
Another thing which would be nice is the line number of the pattern match case. (Inspired by https://github.com/slamdata/purescript-halogen/issues/87)

puffnfresh:
@paf31 huge :+1:

garyb:
:100:

Although I guess we'd have that already if we had source maps ;)

paf31:
@garyb Well in the case above, the exception was being rethrown, so I think the JS line number was being lost, but I'm not sure.

garyb:
That error was caught by `Aff` judging by the wording of it, but yeah, point taken.

PR #1045, opened 2015-04-15
Make `exportedDeclarations` filter everything

### LINKED ISSUE #747: Instances for private classes and types still appear in psc-docs output

hdgarrood:
fixes #747 

I've added a test case to `examples/manual/passing` for testing psc-docs.

Also, the new functions in Language.PureScript.AST.Declarations don't necessarily need to be exported; I put them at the top level so that there's fewer bound variables to worry about, and so that they can be tested in ghci. I would be happy to move them into a separate module to avoid them being exported if you think that's a good idea.


hdgarrood:
An alternative option would be to:

* Only export `exportedDeclarations`; all the other functions for testing exports are redundant now, as far as I can tell.
* Move `exportedDeclarations` and all of the functions it requires into a separate module; `Language.PureScript.AST.Exported` perhaps.

hdgarrood:
Actually, I'm going to close this and resubmit to `0.7-dev` after #1041 is merged, because changing the API like that will necessitate changes in psc-docs which would cause merge conflicts.

PR #1048, opened 2015-04-15
Relax lower bound of transformers to 0.3.0.0

hdgarrood:
These commits do 3 things:

* Replace all occurrences of `Control.Monad.Except` with one or more of `Control.Monad`, `Control.Monad.Error.Class (MonadError(..))`, and `Control.Monad.Trans.Except`, since `Control.Monad.Except` seems to just re-export all of these. This is necessary in the case where we are using `transformers==0.3.*`, because `transformers-compat` doesn't have `Control.Monad.Except` (not sure why). I think this is worth doing anyway; I don't like how `Control.Monad.Except` re-exports so much stuff.
* Add an explicit dependency on `transformers-compat`. purescript was previously bringing it in as a transitive dependency, so the only effect that this should have is allowing the importing of its modules where necessary.
* Relax `transformers` lower bound to 0.3.\*; now that we have `transformers-compat`, purescript can compile with `transformers==0.3.*`. I tested this by doing:

```
cabal sandbox delete
cabal sandbox init
cabal install --only-dependencies --enable-tests --constraint='transformers ==0.3.*'
cabal build
cabal test
```

The effect of this is that you can now easily install `yesod-bin` together with `purescript` in a sandbox. Also, it is still possible to get  `transformers==0.4.*` when resolving constraints - these changes appear to have no effect on the versions selected in those cases which currently can be resolved.

### DIFF
--- a/psc-make/Main.hs
+++ b/psc-make/Main.hs
@@ -17,7 +17,9 @@
 module Main where
 
 import Control.Applicative
-import Control.Monad.Except
+import Control.Monad
+import Control.Monad.Error.Class (MonadError(..))
+import Control.Monad.Trans.Except
 import Control.Monad.Reader
 
 import Data.Version (showVersion)
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -17,7 +17,7 @@
 module Main where
 
 import Control.Applicative
-import Control.Monad.Except
+import Control.Monad
 import Control.Monad.Reader
 
 import Data.Maybe (fromMaybe)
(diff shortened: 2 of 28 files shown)


hdgarrood:
Sorry, it looks like I wasn't quite right; by default, `transformers==0.3.0.0` is chosen by the solver. But if you specify `constraint: transformers >= 0.4` then it still manages to find a solution.

hdgarrood:
On further investigation it seems that `yesod-bin` only needs to have executables on your PATH, not be in the same sandbox as purescript. If you don't want to depend on `transformers-compat` unless it's definitely necessary, perhaps I could revert the changes to pursuit.cabal, but leave the rest? That way it will be easy to start using `transformers-compat` if we ever need to.

paf31:
No, I think this is good. I'll merge. Thanks!

PR #980, opened 2015-03-18
Fix psc-docs' other-modules in purescript.cabal

hdgarrood:
It seems as though installing purescript 0.6.9.2 from hackage is not currently possible; it just failed on my machine. I think this will fix it.

PR #1050, opened 2015-04-15
Check source distributions during Travis builds

hdgarrood:
This commit adds a step in travis-script.sh to generate a source
distribution and then check that the source distribution can be built.
Hopefully this will stop things like #980 from happening again.

the idea comes from here:
https://github.com/hvr/multi-ghc-travis/blob/master/.travis.yml#L98-L103

### DIFF
--- a/travis-script.sh
+++ b/travis-script.sh
@@ -9,3 +9,9 @@ fi
 
 cabal build
 cabal test
+
+# Check that a source distribution can be successfully generated, and that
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Excellent!

hdgarrood:
If you want, I could try removing a module from the Cabal file and then we could see if this fails.

paf31:
Let's try it in a branch after I merge.

PR #1055, opened 2015-04-17
Steps toward pursuit 2

### LINKED ISSUE #747: Instances for private classes and types still appear in psc-docs output

hdgarrood:
Everything that was in #1041 (which has been squashed into one commit), plus:

* Fix #747 
* An initial version of the tool that people will hopefully end up using to upload their packages to Pursuit. The eventual plan is for it to generate a bunch of JSON and then upload it. Since the server doesn't exist yet, it currently just dumps the JSON to standard output.

Sorry this is so huge. The commits are logically separated this time, at least.


paf31:
Cool. I'll review today. Don't worry about the upload step. I think that's probably best left to a tool like `gulp-purescript` or `pulp` anyway.

hdgarrood:
Thanks :)

hdgarrood:
Thinking about this a bit more, I'm happy with what I did here in the first two commits, but not so much the third one. In particular I don't think the "depsModules" key is necessary, because that information (i.e. which package a particular module is in) is already contained in the "bookmarks". I'll think about this a bit more and try to come up with a better way of representing it.

hdgarrood:
Just had a thought: this program should probably fail if you have code in your bower_components which you rely on and which is not declared as a dependency in your bower.json. It doesn't currently do that but I don't think it would be too hard to implement it.

PR #1056, opened 2015-04-18
Optimise (f <<< g $ x) as (f (g x))

puffnfresh:
When writing PureScript code like:

    not <<< not $ id true

We see a few function calls:

    Prelude["<<<"](Prelude.semigroupoidArr)
        (Prelude.not(Prelude.boolLikeBoolean))
        (Prelude.not(Prelude.boolLikeBoolean))
        (Prelude.id(Prelude.categoryArr)(
            true));

We now see something very straight forward:

    !!Prelude.id(Prelude.categoryArr)(true);

The implementation of this optimisation contains a lot of similarities
to the common binary operator inlining. We should extract some
functions out in another commit.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/Optimizer.hs
+++ b/src/Language/PureScript/CodeGen/JS/Optimizer.hs
@@ -77,7 +77,8 @@ optimize' js = do
     , inlineOperator (C.prelude, (C.$)) $ \f x -> JSApp f [x]
     , inlineOperator (C.prelude, (C.#)) $ \x f -> JSApp f [x]
     , inlineOperator (C.preludeUnsafe, C.unsafeIndex) $ flip JSIndexer
-    , inlineCommonOperators ]) js
+    , inlineCommonOperators
+    , inlineAppliedArrComposition ]) js
 
 untilFixedPoint :: (Eq a) => (a -> a) -> a -> a
 untilFixedPoint f = go
--- a/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
+++ b/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
@@ -18,6 +18,7 @@ module Language.PureScript.CodeGen.JS.Optimizer.Inliner (
   inlineValues,
   inlineOperator,
   inlineCommonOperators,
+  inlineAppliedArrComposition,
   etaConvert,
   unThunk,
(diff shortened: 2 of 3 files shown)


puffnfresh:
Hacked this up pretty quick and it's getting late. Haven't even run tests yet, just putting my work somewhere to track early.

src/Language/PureScript/CodeGen/JS/Optimizer.hs
```diff
@@ -77,7 +77,8 @@ optimize' js = do
     , inlineOperator (C.prelude, (C.$)) $ \f x -> JSApp f [x]
     , inlineOperator (C.prelude, (C.#)) $ \x f -> JSApp f [x]
     , inlineOperator (C.preludeUnsafe, C.unsafeIndex) $ flip JSIndexer
-    , inlineCommonOperators ]) js
+    , inlineCommonOperators
+    , inlineAppliedArrComposition ]) js
```
puffnfresh: Would this be the right place to put this optimisation? Is there a better order?

src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
```diff
@@ -31,6 +32,10 @@ import Language.PureScript.Names
 import Language.PureScript.CodeGen.JS.Optimizer.Common
 import qualified Language.PureScript.Constants as C
 
+-- TODO: Potential bug:
```
paf31: I didn't realize `0..toFixed` was even valid syntax ... interesting :)

paf31:
Looks great. I was thinking about this today too funnily enough. I should be able to merge this tomorrow.

src/Language/PureScript/CodeGen/JS/Optimizer.hs
paf31: Finding the best order is tricky in general. I've done it by trial and error in the past, but usually I have issues with things like `runST` and do notation. I don't think there's any issue here though.

garyb:
Do we inline `f $ x` at the moment? That case is worth covering too I'd think. Both of these things I'd intended to add in a `CoreFn` optimizer step, but we don't have one of those yet.

puffnfresh:
@garyb yeah, we do $ already. This optimisation even relies upon that step.

PR #1064, opened 2015-04-20
Use Console module in PSCi

### LINKED ISSUE #1062: psci now depends on Console

paf31:
@garyb 

Fixes #1062.

Do we want to include a PSCi support module instead? The benefit is that we could replace the `Show` class with an `Exec` class like we had in the `ps-in-ps` repo, allowing users to execute `Eff` actions.

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -214,8 +214,8 @@ createTemporaryModule :: Bool -> PSCiState -> P.Expr -> P.Module
 createTemporaryModule exec PSCiState{psciImportedModules = imports, psciLetBindings = lets} val =
   let
     moduleName = P.ModuleName [P.ProperName "$PSCI"]
-    traceModule = P.ModuleName [P.ProperName "Debug", P.ProperName "Trace"]
-    trace = P.Var (P.Qualified (Just traceModule) (P.Ident "print"))
+    consoleModule = P.ModuleName [P.ProperName "Console"]
+    trace = P.Var (P.Qualified (Just consoleModule) (P.Ident "print"))
     mainValue = P.App trace (P.Var (P.Qualified Nothing (P.Ident "it")))
     itDecl = P.ValueDeclaration (P.Ident "it") P.Value [] $ Right val
     mainDecl = P.ValueDeclaration (P.Ident "main") P.Value [] $ Right mainValue
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -59,6 +59,35 @@ import Parser (parseCommand)
 import Completion (completion)
 import Types
 
+-- | The name of the PSCI support module
+supportModuleName :: P.ModuleName
+supportModuleName = P.ModuleName [P.ProperName "$PSCI", P.ProperName "Support"]
+
+-- | Support module, contains code to evaluate terms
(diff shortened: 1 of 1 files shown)


garyb:
Being able to execute `Eff` in `psci` would be very useful, so maybe that is the way to go.

paf31:

psci/PSCi.hs
```diff
@@ -202,7 +231,7 @@ instance P.MonadMake Make where
   writeTextFile path text = makeIO (const (P.CannotWriteFile path)) $ do
     mkdirp path
     writeFile path text
-  progress s = unless (s == "Compiling $PSCI") $ liftIO . putStrLn $ s
+  progress s = unless (take 15 s == "Compiling $PSCI") $ liftIO . putStrLn $ s
```
hdgarrood: could this be written using `isPrefixOf`?

psci/PSCi.hs
```diff
@@ -59,6 +59,35 @@ import Parser (parseCommand)
...
+    , "import Control.Monad.Eff.Unsafe"
+    , ""
+    , "class Eval a where"
+    , "  eval :: a -> Eff (console :: CONSOLE) Unit"
+    , ""
+    , "instance evalShow :: (Show a) => Eval a where"
+    , "  eval = print"
+    , ""
+    , "instance evalEff :: (Eval a) => Eval (Eff eff a) where"
+    , "  eval x = unsafeInterleaveEff x >>= eval"
```
hdgarrood: :+1: :+1: :+1:

hdgarrood:
Does this also introduce a dependency on `purescript-eff`?

paf31:
Yes, but `console` depends on `eff` anyway.

psci/PSCi.hs
paf31: Done, thanks!

hdgarrood:
Ah, of course. LGTM!

garyb:
:shipit:

psci/PSCi.hs
```diff
@@ -59,6 +59,35 @@ import Parser (parseCommand)
...
+
+-- | Support module, contains code to evaluate terms
+supportModule :: P.Module
+supportModule = 
+  case P.parseModulesFromFiles id [("", code)] of
+    Right [(_, P.Module cs _ ds exps)] -> P.Module cs supportModuleName ds exps
+    _ -> error "Support module could not be parsed"
+  where
+  code :: String
+  code = unlines
```
michaelficarra: Gross. Use `"""` for multiline strings.

michaelficarra: Whoops, Haskell, not PureScript. Never mind.

garyb: Haskell doesn't have `"""` multiline strings :wink:

It does have the `\` stuff, but that's not really any better, and strips linebreaks.

paf31: :) Yeah, triple quotes would be nice. The other reason I did it this way was that I thought it might be nice to have a flag to remove the `Eff` dependency, in which case I'd end up gluing lines together one day, but now we have a pretty heavy `Eff` dependency because of the type of `eval`.

Issue #1061, opened 2015-04-20
Move Int and Char into Prim, and add Int and Char literals


paf31:
Should we do this as part of the 0.7 changes, or after?

/cc @garyb

garyb:
I guess for the 0.7 release would be ideal.

Would we allow casting integer literals to `Number`?
``` purescript
four :: Number
four = 4
```
 Or will `Number`s strictly require the `x.0` ending?

garyb:
:scream:

This will made importing `Data.Int (Int())` redundant, will have to go back and take it out everywhere, as `Int` will already be in scope.

paf31:
I was thinking `1 :: Int` but `1.0 :: Number` with no implicit casting.

Yes, it would be quite a bit more work sadly, which is why I wanted to figure out when we wanted to do it.

andyarvanitis:
Any plans for `Char` as well, including `''` literals?

garyb:
@paf31 Yeah, that makes sense I think, at least that way it's never ambiguous.

It's unfortunate, but I'd rather not have to go through all the libraries yet again at a later date, so may as well get it done now.

jdegoes:
> It's unfortunate, but I'd rather not have to go through all the libraries yet again at a later date, so may as well get it done now.

Agreed. Also changes this drastic will have a negative affect on the PureScript community, so it's best to Break All the Things now. With each point release comes increasing pressure to preserve backward compatibility.

jdegoes:
> Any plans for Char as well, including 'c' literals?

Please oh please! :+1:

garyb:
:+1: for Char literals also, I don't think they will interfere with prime'd identifiers?

paf31:
I can get behind `Char` literals too.

paf31:
Anyone care to take a look at this? It would be a good intro to the parser and code gen and bits of the type checker.

If not, I'll take care of it early next week.

andyarvanitis:
Just FYI, I have them (`Char` / literals) in pure11, although they just end up as `Int`s.

michaelficarra:
[...]

PR #1065, opened 2015-04-20
Int literals #1061

paf31:
Work in progress, tests currently broken.

/cc @garyb

### DIFF
--- a/examples/passing/Arrays.purs
+++ b/examples/passing/Arrays.purs
@@ -2,23 +2,23 @@ module Main where
 
 import Prelude.Unsafe (unsafeIndex)
 
-test1 arr = arr `unsafeIndex` 0 + arr `unsafeIndex` 1 + 1
+test1 arr = arr `unsafeIndex` 0.0 + arr `unsafeIndex` 1.0 + 1.0
 
 test2 = \arr -> case arr of
   [x, y] -> x + y
   [x] -> x
-  [] -> 0
+  [] -> 0.0
   (x : y : _) -> x + y
 
 data Tree = One Number | Some [Tree]
 
 test3 = \tree sum -> case tree of
   One n -> n
-  Some (n1 : n2 : rest) -> test3 n1 sum * 10 + test3 n2 sum * 5 + sum rest
+  Some (n1 : n2 : rest) -> test3 n1 sum * 10.0 + test3 n2 sum * 5.0 + sum rest
 
 test4 = \arr -> case arr of
-  [] -> 0
-  [_] -> 0
+  [] -> 0.0
+  [_] -> 0.0
   x : y : xs -> x * y + test4 xs
 
 main = Debug.Trace.trace "Done"
--- a/examples/passing/AutoPrelude.purs
+++ b/examples/passing/AutoPrelude.purs
@@ -2,7 +2,7 @@ module Main where
 
 import Debug.Trace
 
-f x = x * 10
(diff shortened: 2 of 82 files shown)


garyb:
So far so :+1:

paf31:
The choice `1 :: Int` and `1.0 :: Number` is nice from a type checker POV, but it's certainly not normal, and I wonder if it's going to lead to confusing errors. Things like `1 does not have type Number` crop up, which might confuse a beginner. We can always document this stuff, but I think a lot of people will expect `1 :: Number` to work. But then, I don't really want to go the overloaded literals route either.

garyb:
That's why I was asking about the behaviour of it yesterday... I think I still do prefer this stricter option though. We can always spend some time ensuring the error is helpful though, adding something like `did you mean to use the Number 1.0 rather than the Int 1?`

paf31:
Improving errors sounds like the way to go.

garyb:
Or if we were going to allow coercion, at least make it a warning.

paf31:
Coercion without overloaded literals is going to be a nightmare from a typechecking perspective.

andyarvanitis:
Would allowing implicit integral literals to `Number` conversion (but not the other way around) be ok?

garyb:
I guess this whole thing comes down to: What's more inconvenient? Using `fromNumber` a lot, or having to add `.0` to floats sometimes. Either one is going to be a slight hurdle for a beginner, and add `.0` is the less gross option to me at least. :smile:

natefaubion:
Could we at least allow '1.'. I think this is fairly common in languages other than JS, eg Python:

```
>>> type(3)
<type 'int'>
>>> type(3.)
<type 'float'>
```

paf31:
@natefaubion Sure, if we can get the lexer to recognize it unambiguously.

paf31:
@andyarvanitis I'd rather avoid any sort of subtyping rules, since we risk losing principal types.

joneshf:
I'm not sure if it's been discussed before, but what about something like what haskell does with the `IsString` class:
http://hackage.haskell.org/package/base-4.8.0.0/docs/Data-String.html#t:IsString except for numeric literals (akin to `fromInteger` from `Num` or `fromRational` from `Fractional` except not muddled with other functions)?

Though, I think it'd fall prey to the same lack of constraint inference that we have now, so maybe that's not an immediate thing.

paf31:
> Though, I think it'd fall prey to the same lack of constraint inference that we have now

That's the only reason I want to avoid this for now, but we could move in that direction.

paf31:
@garyb @andyarvanitis @michaelficarra Could you please review?

paf31:
CI will probably fail due to core-tests.

garyb:
Can we not use `Int` or `Char` as a type in the tests then?

paf31:
We can, but I need to update the test prelude to include `Semiring` etc. for `Int` first. This was just a quick pass to get stuff working. I added a new `IntAndChar` test though.

garyb:
Yeah, I figured that about the `Prelude`, I just wondered as I didn't see the `Int` or `Char` type used in the test, although the literals are.

:+1:

andyarvanitis:
Looks good to me. It gives backends control over how to translate the `CharLiteral` itself (rather than, say, making it a `StringLiteral` upstream), so I'm happy!  :+1:

PR #1067, opened 2015-04-21
Steps toward pursuit 2

### LINKED ISSUE #747: Instances for private classes and types still appear in psc-docs output

hdgarrood:
Fixes #747, continues from #1041, #1055. Much of this code comes from psc-pages.

The idea with the `AsHtml` module is that it is general enough that it could potentially be used in different environments with different stylesheets and page hierarchies - the rest of the code required to display documentation as HTML (eg including style sheets, templating) should end up in the pursuit server, I think.

I renamed `psc-upload-package` to `psc-publish` since it doesn't upload anything.

I commented on #1055 that I thought this tool should check for dependencies which aren't declared in bower.json, but I can't see an easy way of handling transitive dependencies. That is, if package A exists in bower_dependencies but is not listed in bower.json, we would need to tell the difference between A being pulled in as a transitive dependency of one of the other packages (which is OK), and A not being a transitive dependency of any other package, but having just been `bower install`-ed without being added to `bower.json` (which is an error on the part of the maintainer, and could cause issues for users of that package). Given that this tool only exists to publish package information, and given that it would be quite complex, I no longer think this check should be in scope for `psc-publish`.


hdgarrood:
The build is failing because I was using `cabal sandbox add-source` for `aeson-better-errors`, and I forgot to publish the changes I made, so this code depends on an unreleased version of `aeson-better-errors`. Now Hackage is not letting me publish a new version. :(

psc-publish/ErrorsWarnings.hs
```diff
@@ -0,0 +1,195 @@
...
+  let names = map getPackageName warns in
+  case length warns of
+    0 -> nullBox
+    1 -> pluraliseMsg "package" "was" "this" "this" names
+    _ -> pluraliseMsg "packages" "were" "any of these" "these" names
+
+  where
+  getPackageName (ResolutionNotVersion n) = runPackageName n
+
+  pluraliseMsg packages were anyOfThese these names =
```
paf31: Nice :)

paf31:
Looks good. Could you please do me a favour and summarise what requests `wreq` is used for?

psc-publish/Main.hs
```diff
@@ -0,0 +1,230 @@
...
+-- linking to documentation for any types from that package.
+--
+-- The rationale for this is: people will prefer to use a released version
+-- where possible. If they are not using a released version, then this is
+-- probably for a reason. However, docs are only ever available for released
+-- versions. Therefore there will probably be no version of the docs which is
+-- appropriate to link to, and we should omit links.
+getResolvedDependencies :: PrepareM [(PackageName, Version)]
+getResolvedDependencies = do
+  pathsBS <- fromString <$> readProcess' "bower" ["list", "--paths", "--json"] ""
```
hdgarrood: It's just occurred to me that this won't work if any purescript package author decides to use the `main` key in bower.json. This code assumes that `bower list --paths --json` prints a JSON object mapping package names to directory paths, which is not true if the `main` key is specified.

For example, `purescript-timers#0.0.8` specifies `main` in its bower.json, and in the output of `bower list --paths --json`, you get:

```
  "purescript-timers": "bower_components/purescript-timers/src/Control/Reactive/Timer.purs",
```

Instead, I think this code should just glob `bower_components/*` and then attempt to parse the second component as the package name.

hdgarrood:
At the moment, `wreq` is only used to ask Bower about the package, in order to get its git url (which we then extract a github username / repository name from). See https://github.com/purescript/purescript/pull/1067/files#diff-8668135e8cf73d8f5109bc17a79e0826R119

I was originally planning on using `wreq` to upload the generated JSON blob to pursuit as well, but if this ends up happening in a separate program then obviously we won't need to.

hdgarrood:
Come to think of it, given the HTTP client needs of this program are quite a lot simpler than I originally imagined, it may well be worth switching to `HTTP`, which is in the Haskell Platform, and has a much smaller list of dependencies.

paf31:
We don't necessarily have to use HTTP, that's fine. I'm just wondering, what can we get from the service that we can't get from the file? We could always require custom fields for the file. I'd prefer to have everything in one place honestly. That way, if Bower ever disappears, we can still keep the file format.

hdgarrood:
Bower does have the `repository` key, which we could use instead: https://github.com/bower/bower.json-spec#repository

So, we could check that the `type` is `git`, and check that `url` is a GitHub URL, and then extract GitHub user/repo from the URL in the same way as this code currently does. I would hope that `Bower` has some kind of check in place to ensure that the `repository` key is the same as what is on the Bower registry (ie, where the code comes from when you `bower install pkg` but I am not sure about this.

The downsides are:

* it's a bit of extra hassle for PureScript package maintainers to have to put this in; I'm not aware of any that do have it currently. None of the core packages do.
* it's not the 'real' source of that information - the Bower registry is.

See also https://github.com/paf31/psc-pages/pull/18, @bodil's comment in particular.

hdgarrood:
We also aren't really getting everything from one place at the moment - a large chunk comes from the `.purs` files in the current directory, the version comes from `git tag --points-at HEAD`, the resolved dependencies (i.e. specific versions of dependencies which we use for links in published HTML documentation) come from digging around in `bower_components`...

paf31:
Well, given that we're going to create a cached snapshot of the docs at a particular version, the URL could become out of date at any point later anyway. At least with the `bower.json` approach, we know that it was the right URL for the version of the code at the time the docs were created.

Rather than parsing the `repository` field, we could require an optional, custom field, containing the GitHub user and repo names - just a thought. If those were missing, we could just omit source links.

hdgarrood:
Given that specifying a repository ought to be a very small amount of effort on the part of the maintainer however we choose to do it, and also given that we're happy assuming that PureScript packages are hosted on GitHub, I think I'd rather require this information be present, to ensure that we have source links in Pursuit documentation. I find them very useful for Haskell documentation on Hackage - and when they don't appear, I think it's quite irritating.

I think if we want to know that we've got the right URL for the code at the time the docs were created, we're better off using the URL from the Bower registry, since this is the one that gets used when you `bower install` things.

Also, if we do want to get this information from the Bower registry, we could switch to `http-conduit`, which provides an equally simple interface for what we want, and has a tiny dependency list in comparison.

bodil:
I can have Pulp enforce the presence of this information to some extent as well. Absent the ability to reject published packages lacking it, that's probably the closest we'll get to a guarantee.

psc-publish/Main.hs
hdgarrood: On second thoughts, we can get all the information we need from `bower list --json`.

Using `bower list --json` also allows us to easily identify bower dependencies which haven't been installed, because Bower includes `"missing": true` in the output where appropriate. So we can fail with a nice error if the user forgot to `bower install`.

hdgarrood:
Sorry, the presence of which information?

bodil:
Specifying a repository, in particular.

hdgarrood:
Ok then - let's take it from `bower.json`. I'd rather use the Bower-specified `repository` key - as far as I can tell, the only benefit to adding a custom key would be that it's marginally easier to put the information in, and I don't see that as a good enough reason to deviate from the standard.

hdgarrood:
I've fixed the issue stemming from `bower list --paths` not always returning directories, and the repository now comes from the repository field of bower.json. I think this is good to go. Perhaps a rebase would be a good idea, but I'll wait for you to look over the new commits.

hdgarrood:
Oh, one last thing I forgot, but would like to do - make all of the errors pluralise properly. The errors that refer to a list of things should use the singular if the list only has one thing in it.

hdgarrood:
Ok, I'm really done now. :)

hdgarrood:
Oh, also, I suppose we should warn people that `psc-docs` (and probably also Pursuit, when we get there) will not display instance documentation. If any instances are worth documenting, I think we could suggest that documentation should go on the type itself.

paf31:
Getting ready to merge this. Code looks great (@garyb, any comments?), but when I build, I pull in a _lot_ of dependencies, including `wai`. Any idea where they're coming from?

hdgarrood:
Not sure - I'll take a look.

paf31:
Looks like `cheapskate` depends on `wai` for something :(

paf31:
Looks like it's used by the binary. How disappointing.

hdgarrood:
Ahhh, ok. I did waver about whether the `AsHtml` module should go into the `purescript` package. If we moved it somewhere else we would no longer need `lucid`, `blaze-html`, `cheapskate`, or `data-default`. Also it might be useful to be able to update the HTML generation code without having to produce a new compiler release. Up to you.

paf31:
So I compiled and ran this with `purescript-prelude`, and the results are nice, but should we be concerned at all about the verbosity of the JSON files (all the `space`s, etc.)? I'm not sure, but I thought it might be worth bringing up.

hdgarrood:
I'm not really bothered about the size of the JSON files - `purescript-prelude` is 66K, gzipping takes it down to 7K, and I think it's reasonable to guess that the vast majority of PureScript packages will be, say, no more than 10x that.

With respect to how human-readable it is, it's admittedly not great. Running it through a JSON formatting tool does make it a lot easier, though. Also, I think ease of encoding/decoding is much more important, and I can't immediately see a good way of improving the verbosity without adding a fair bit of complexity to the encoding/decoding steps.

paf31:
Makes sense :+1:

garyb:
@paf31: Sorry, I've not really been keeping track of the progress with this - if you're fine with it, I am!

hdgarrood:
I've just realised I've introduced a bug in `psc-docs` - the code in `Language.PureScript.AST.Exported` assumes that the `desugarImports` step has been performed, but `psc-docs` doesn't currently run it. The effect of this is that `psc-docs` incorrectly thinks that all instance declarations are 'private' and doesn't display them.

The obvious fix is to require `desugarImports` to be called first, and this requires providing the same source file list as would be necessary to compile the file normally. This is unfortunate, because `psc-docs` is currently very simple to use: simply running `psc-docs file.purs` works at the moment.

Alternatively, I could investigate modifying `exportedDeclarations` so that it works before or after calling `desugarImports`.

paf31:
That is a bit of a shame, but honestly, I think most people use a tool to generate docs anyway.

The one case which would break is where we want to generate lots of small docs files, one per module. Maybe we need something analogous to `--codegen` from `psc` - to only generate docs for the given modules.

I think desugaring imports etc. _is_ the right way to go though. If we ever want to do things like add links in the generated Markdown, we need to do that.

hdgarrood:
Ok, sounds good. I thought the same re `--codegen`. How about specifying it like `moduleName:destFile`, so for example:

```
$ psc-docs src/**/*.purs bower_components/*/src/**/*.purs \
   --docgen Data.FingerTree:docs/Data.FingerTree.md \
            Data.Sequence:docs/Data.Sequence.md
```

paf31:
Sounds good, but is that sort of thing easily supported with `optparse-applicative`? Can we give good errors etc?

hdgarrood:
I'm not sure, I'll have a go and see what happens.

hdgarrood:
`psc-docs` seems to work, and the error messages are decent, I think. Perhaps the wording could be improved for this one - any ideas?

```
$ psc-docs Foo.purs --docgen Foo --docgen Foo:Foo.md
psc-docs: error in --docgen option:
  Can't mix module names and module name/file path pairs in the same invocation.
```

Here's a draft changelog entry for these changes:

* A bug where instances of nonexported types were appearing in `psc-docs` output has been fixed. (#747)
* `psc-docs` now groups all instances directly under the relevant type and/or type class.
* The comand line interface for `psc-docs` has changed. See `psc-docs --help` for details.

As far as I can tell, all we need to do now is:

* [ ] update grunt-purescript
* [ ] update gulp-purescript
* [ ] update pulp
* [ ] decide what to do about the dependencies issue / where the HTML generation code goes. I think we should put all HTML generation code inside the Pursuit server for now. We can always move it to a standalone library if we need to reuse that code elsewhere.

paf31:
I think putting the HTML code into the Pursuit server sounds fine. :+1: 

Also, before I forget, whatever new dependencies we add should be noted in the LICENSE file, since we're redistributing them in binary form.

hdgarrood:
LICENSE file should be done now (see https://github.com/purescript/purescript/commit/719dde6d64bf946021948097396722ab5c076542).

psc-publish/Utils.hs
```diff
@@ -0,0 +1,26 @@
+
+module Utils where
+
+import Data.List
+import Data.Maybe
+import System.Directory
+import qualified System.FilePath.Glob as Glob
+
+headMay :: [a] -> Maybe a
```
paf31: Is it worth using the `safe` package for this?

paf31:
Reviewed, looks great. Sorry this took so long.

Could you please squash commits, and I'll merge this in.

psc-publish/Utils.hs
hdgarrood: Yes, I think so. I'll add it

hdgarrood:
Is that ok?

paf31:
:+1: x 100, thanks!

hdgarrood:
Hooray :)

PR #1070, opened 2015-04-21
Inline semigroupoidArr composition

### LINKED ISSUE #1066: Optimisation: (f <<< g) ~ function(_0) { return f(g(_0)); }
Special cased on `semigroupoidArr`. The goal is to remove intermediately created function objects.

Should work for more than 2 compositions:

```purescript
f <<< g <<< h
```

```js
function(_0) {
  return f(g(h(_0)));
}
```

Requires expanding `optimize` in `Language.PureScript.CodeGen.JS.Optimizer` to allow fresh variable creation. We have to make sure `untilFixedPoint` still works but this optimisation should work fine since it should be idempotent, not creating fresh variables on repeat rewrites.

We may also have to be careful around the recent https://github.com/purescript/purescript/pull/1056 optimisation.

puffnfresh:
An example of the changes:

```purescript
g :: Boolean
g = not <<< not $ true

f :: Boolean -> Boolean
f = not <<< not <<< not <<< not

f' x = (\y -> f y) x
```

```js
var g = !!true;

var f = function (_44) {
    return !!!!_44;
};

var f$prime = function (x) {
    return f(x);
};
```

I did the following:

1. Allowed `everywhereOnJSTopDown` to work in a `Monad`, default is `Identity`.
2. Made `untiFixedPoint` work on a `Monad`.
3. Added `inlineAppliedVars` optimisation, turning `(\x -> f x) y` into `f y`.
4. Added `inlineArrComposition` optimisation, turning `a >>> b` into `(\x -> a (b x))`.

What are the drawbacks?

Should fix #1066.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/AST.hs
+++ b/src/Language/PureScript/CodeGen/JS/AST.hs
@@ -17,6 +17,8 @@
 
 module Language.PureScript.CodeGen.JS.AST where
 
+import Control.Monad ((>=>))
+import Control.Monad.Identity
 import Data.Data
 
 import Language.PureScript.Comments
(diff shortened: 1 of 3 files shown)


garyb:
Looks great to me, and good call on the `Identity` monad, I think we have some AST traversals we could have done that with.

paf31:
Looks good, thanks!

I'm always wary of inlining operations, since they have broken code in odd ways in the past. In this case, I thought the existing inlining code would have taken care of it, so I'd like to figure out why not:

https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs#L48

That code is very strict in terms of what it accepts, but only because so much code has broken in the past, and it's been changed a few times.

What is the set of transformations we want to enable, and can we figure out why the existing optimization isn't firing?

The fresh name code and `<<<` optimization look great. Should we optimize `>>>` too?

puffnfresh:
@paf31 I removed `inlineAppliedVars`. At some point it seemed like it was making `(<<<)` work but now it doesn't seem necessary. I imagine the `(<<<)` optimisation just wasn't working properly at the time.

paf31:
Looks good. I'll just wait for the CI build to finish.

PR #1073, opened 2015-04-22

nicodelpiano:

### DIFF
--- a/travis-after-script.sh
+++ b/travis-after-script.sh
@@ -1,10 +1,10 @@
 set -e
 
-# pushd core-tests/
+pushd core-tests/
 
-# ./test-everything.sh
+./test-everything.sh
 
-# popd
+popd
 
 if [ -z $( git describe --tags --exact-match 2>/dev/null ) ]
 then


paf31:
This looks great! Let's just add the `censor` call as we discussed, to substitute the correct type, and maybe add a test to `examples/manual` and then I'll merge this.

nicodelpiano:
Is there a way to get the substitution table without re-running the computation after `liftUnify`?          Maybe we should wedge `censor` elsewhere...

paf31:
You might have to modify `liftUnify`, I'm not sure. Just try things out and see what works, but we shouldn't run the computation twice.

nicodelpiano:
```haskell
liftUnifyType :: UnifyT Type Check a -> Check (a, Substitution Type)
liftUnifyType unify = do
  st <- get
  let ru = runUnify (defaultUnifyState { unifyNextVar = checkNextVar st }) unify
  (a, ust) <- ru
  modify $ \st' -> st' { checkNextVar = unifyNextVar ust }
  let uust = unifyCurrentSubstitution ust
  censor (onErrorMessages $ replace uust) $ ru
  return (a, uust)
    where
      replace sub (WildcardInferredType ty) = WildcardInferredType (sub $? ty)
      replace _ x = x
```
This gives the following:
```
  The wildcard type definition has the inferred type TUnknown 1883
  See https://github.com/purescript/purescript/wiki/Error-Code-WildcardInferredType for more information, or to contribute content related to this error.

  The wildcard type definition has the inferred type TypeConstructor Prim.String
  See https://github.com/purescript/purescript/wiki/Error-Code-WildcardInferredType for more information, or to contribute content related to this error.

  The wildcard type definition has the inferred type TUnknown 1885
  See https://github.com/purescript/purescript/wiki/Error-Code-WildcardInferredType for more information, or to contribute content related to this error.

  The wildcard type definition has the inferred type TypeConstructor Prim.Number
  See https://github.com/purescript/purescript/wiki/Error-Code-WildcardInferredType for more information, or to contribute content related to this error.
```
As you can see, it duplicates the warnings because of the calling of `ru` inside `censor`, but it throws the correct inferred types.

Any idea of how we can avoid running again `runUnify`? (I've tried using `pass` instead of `censor` but it didn't work)

paf31:
Looks great! I'll have a think about the censor issue and get back to you. For now, here are some notes:

- You should try to add position info to your warnings if possible, using `PositionedError`. You will need to modify `replace` to handle this case.
- Use `prettyPrintType` instead of `show` on types.
- At some point, you'll want to merge master and rebase your commits.

Thanks! Looking forward to merging this in.

paf31:
For now, maybe just use `censor . listen` to keep the warnings in the result of the first call without logging them. Then, instead of running `ru` again, just use `tell`.

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -200,16 +201,17 @@ modifyEnv f = modify (\s -> s { checkEnv = f (checkEnv s) })
...
 runCheck = runCheck' initEnvironment
 
 -- |
 -- Run a computation in the Check monad, failing with an error, or succeeding with a return value and the final @Environment@.
 --
-runCheck' :: (MonadError MultipleErrors m) => Environment -> Check a -> m (a, Environment)
-runCheck' env c = interpretMultipleErrors $ do
-  (a, s) <- flip runStateT (CheckState env 0 0 Nothing) $ unCheck c
-  return (a, checkEnv s)
+runCheck' :: (MonadError MultipleErrors m, MonadWriter MultipleErrors m) => Environment -> Check a -> m (a, Environment, MultipleErrors)
```
paf31: Why does this need to return `MultipleErrors`? Also, we shouldn't run the `Writer` computation twice.

paf31: Also, I don't think you use the `MonadWriter` constraint. I think you want to use `runWriter` to get the log back, and then pass it to `tell`.

Later, we can hopefully tidy things up by just getting rid of this function altogether, and just working with `MonadState Environment`.

nicodelpiano: Yes, you are right. I'm not using the constraint and also running `Writer` twice. I fixed both issues.
I returned `MultipleErrors` to use `tell` outside inside `compile'`, as we can't use `tell` in `runCheck`.
Do you think that is a bad decission?
Thanks!

paf31: Why can't you use `tell`? You have `MonadWriter`.

To be clear, the way it is now is a hack. What I should have done is to have the typechecker run in any `Monad` with the right `mtl` constraints, like everything else, and get rid of `Check` entirely, but it's a big change. I hope to look at it after this gets merged.

nicodelpiano:
```haskell
runCheck' :: (MonadError MultipleErrors m, MonadWriter MultipleErrors m) => Environment -> Check a -> m (a, Environment)
runCheck' env c = interpretMultipleErrors $ do
  let (runc, warns) = runWriter . runExceptT . flip runStateT (CheckState env 0 0 Nothing) $ unCheck c
  (a, s) <- runc
  tell warns
  return (a, checkEnv s)
```
This throws:
```
Could not deduce (MonadWriter
                        MultipleErrors (Either MultipleErrors))
      arising from a use of ‘tell’
```

paf31: Here's the goal: we want to unwrap all the constructors of the monad transformers making up `Check` and get some data structure. It's something involving `Either`, `(,)`, `Environment`, `MultipleErrors` and `CheckState`.

Then we want to pass that data structure to `interpretMultipleErrors` to interpret it in some arbitrary `Monad` with the right `mtl` constraints. 

Try breaking the problem down into those two parts. Write `runCheck' env = interpretMultipleErrors . unwrapCheck env`, and try to implement `unwrapCheck` in a `where` clause. Then, you need to modify `interpretMultipleErrors` to handle warnings. Right now it only interprets errors. I would also rename it, to something like `interpretErrorsAndWarnings`.

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -200,16 +201,19 @@ modifyEnv f = modify (\s -> s { checkEnv = f (checkEnv s) })
...
 -- Run a computation in the Check monad, failing with an error, or succeeding with a return value and the final @Environment@.
 --
-runCheck' :: (MonadError MultipleErrors m) => Environment -> Check a -> m (a, Environment)
-runCheck' env c = interpretMultipleErrors $ do
-  (a, s) <- flip runStateT (CheckState env 0 0 Nothing) $ unCheck c
-  return (a, checkEnv s)
+runCheck' :: (MonadError MultipleErrors m, MonadWriter MultipleErrors m) => Environment -> Check a -> m (a, Environment)
+runCheck' env = interpretMultipleErrorsAndWarnings . unwrapCheck env
+  where
+  unwrapCheck e c = do
```
paf31: Could you please add a type signature? I think you have the tuple and `Either` the wrong way round. We want warnings even in the case of errors. i.e. `(Either MultipleErrors a, MultipleErrors)`

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -242,3 +250,17 @@ liftUnify unify = do
   (a, ust) <- runUnify (defaultUnifyState { unifyNextVar = checkNextVar st }) unify
   modify $ \st' -> st' { checkNextVar = unifyNextVar ust }
   return (a, unifyCurrentSubstitution ust)
+
+-- |
+-- Run a computation in the substitution monad, generating a return value, the final substitution and updating warnings values.
+--
+liftUnifyWarnings :: (Partial t) => (Substitution t -> ErrorMessage -> ErrorMessage) -> UnifyT t Check a -> Check (a, Substitution t)
```
paf31: `liftUnify = liftUnifyWarnings (const id)`

paf31:
This is looking really nice. Please just see my two suggestions. Also, if you could please rebase your commits into a single commit, I'll get this merged in!

PR #1079, opened 2015-04-23
Add source position info to match errors

puffnfresh:


puffnfresh:
An example:

```purescript
module Main where

import Control.Monad.Eff

partial :: Ordering -> Number 
partial GT = 1

main :: Eff () Unit
main = do
  let x = partial LT
  return unit
```

Crashes at runtime:

```
        throw new Error("Failed pattern match at /tmp/partial.purs line 5, col
              ^
Error: Failed pattern match at /tmp/partial.purs line 5, column 1 - line 6, column 1: LT
    at PS.Main.partial ([stdin]:142:15)
    at [stdin]:145:17
    at [stdin]:147:7
    at [stdin]:152:3
```

garyb:
:+1: :+1: :+1: 

Hopefully I brought through the source pos metadata everywhere it should have been done.

paf31:
Very nice!

PR #1082, opened 2015-04-26
Errors

paf31:
@garyb @balajirrao Could you please review? Thanks.

I added a bit of code for tidying kind error messages, but haven't finished it yet. I think that can be a separate PR.


garyb:
:+1:

PR #1083, opened 2015-04-26
Remove array type notation and cons binders

paf31:
@garyb Could you please review this one? It's mostly deletions and modified tests, but you might like to look at the way the new errors work. I just hacked them in using `error` for now, since they will be temporary.


garyb:
The `error`s are a bit icky, but I guess this is as good a way as any to do it. It would be nice to include position info though.

This is actually going to make `CoreImp` easier too, as one of the problems there was to abstract the array behaviours of cons patterns.

paf31:
Yeah, `error` is ugly, but using `fail` won't work since Parsec will just try the next alternative. I'll see if I can add position info.

Issue #1023, opened 2015-04-03
Suggest (<<<) when it appears (.) is being used for composition


hdgarrood:
For example, in psci:

```
> let f = g . g where g = (+1)
```

Gives the error:
```
Error checking that type
  u3 -> u4
subsumes type
  { g :: u5 | u6 }
Error checking that type
  { g :: u5 | u6 }
subsumes type
  u3 -> u4
Error at  line 1, column 9 - line 1, column 15:
  Cannot unify type
    Prim.Object
  with type
    Prim.Function u3
See https://github.com/purescript/purescript/wiki/Error-Code-ErrorUnifyingTypes for more information, or to contribute content related to this error.
```

This seems to trip people up a lot, and it's not immediately obvious what's wrong. It seems like it should be fairly easy to detect cases where people have tried to use `(.)` as function composition in the compiler. In those cases, perhaps we could add a suggestion like this to the end of the type error?

```
Did you mean to use function composition, (<<<)?
```

It might also be nice to give suggested fixes in other situations too - perhaps a general mechanism for suggesting fixes for type errors could be a nice thing to have?

hdgarrood:
Fixed by #1096

PR #1085, opened 2015-04-28
Add error suggestion per #1023

mjgpy3:
See #1023, not sure if that specific case spans across other nearby errors (e.g. `KindsDoNotUnify`, `ConstrainedTypeUnified`).

Any feedback is much appreciated.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -93,6 +93,7 @@ data ErrorMessage
   | EscapedSkolem (Maybe Expr)
   | UnspecifiedSkolemScope
   | TypesDoNotUnify Type Type
+  | ExpectedFunCompMisue Type Type
   | KindsDoNotUnify Kind Kind
   | ConstrainedTypeUnified Type Type
   | OverlappingInstances (Qualified ProperName) [Type] [DictionaryValue]
@@ -131,7 +132,6 @@ data ErrorMessage
   | ErrorInValueDeclaration Ident ErrorMessage
   | ErrorInForeignImport Ident ErrorMessage
   | PositionedError SourceSpan ErrorMessage
-  | ErrorWithSuggestion String ErrorMessage
   deriving (Show)
   
 instance UnificationError Type ErrorMessage where
@@ -188,6 +188,7 @@ errorCode (PartiallyAppliedSynonym _)   = "PartiallyAppliedSynonym"
 errorCode (EscapedSkolem _)             = "EscapedSkolem"
 errorCode UnspecifiedSkolemScope        = "UnspecifiedSkolemScope"
 errorCode (TypesDoNotUnify _ _)         = "TypesDoNotUnify"
+errorCode (ExpectedFunCompMisue _ _)    = "ExpectedFunCompMisue"
 errorCode (KindsDoNotUnify _ _)         = "KindsDoNotUnify"
 errorCode (ConstrainedTypeUnified _ _)  = "ConstrainedTypeUnified"
 errorCode (OverlappingInstances _ _ _)  = "OverlappingInstances"
@@ -227,7 +228,6 @@ errorCode (ErrorInTypeSynonym _ e)      = errorCode e
 errorCode (ErrorInValueDeclaration _ e) = errorCode e
 errorCode (ErrorInForeignImport _ e)    = errorCode e
 errorCode (PositionedError _ e)         = errorCode e
-errorCode (ErrorWithSuggestion _ e)     = errorCode e
   
 -- |
 -- A stack trace for an error
@@ -395,6 +395,12 @@ prettyPrintSingleError full e = prettyPrintErrorMessage <$> onTypesInErrorMessag
                                                , line "with type"
                                                , indent $ line $ prettyPrintType t2
                                                ]
+    go (ExpectedFunCompMisue t1 t2)    = paras [ line "Cannot unify type"
+                                               , indent $ line $ prettyPrintType t1
+                                               , line "with type"
+                                               , indent $ line $ prettyPrintType t2
+                                               , suggest "function composition (<<<)"
+                                               ]
     go (KindsDoNotUnify k1 k2)         = paras [ line "Cannot unify kind"
                                                , indent $ line $ prettyPrintKind k1
                                                , line "with kind"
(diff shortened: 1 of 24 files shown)


paf31:
I think this is good, but will apply in too many cases. We don't want _every_ unification error to give this message, just those from misuse of the `.` syntax.

mjgpy3:
Gotcha, I'll keep looking into it

mjgpy3:
@paf31 I limited the cases slightly, still not entirely sure if it's right. The issue spells out a case where it thinks it's looking at a `Prim.Object` and a `Prim.Function`, is that the only case? Can you provide any advice on how to make this more specific?

mjgpy3:
Looks like tests took too long...

garyb:
I think Travis is having issues at the moment generally, but I've restarted the build.

paf31:
I think we probably do want to restrict this to `Prim.Object` and `Prim.Function`. Also, I liked the previous approach of handling this in `Errors`, since eventually I'd like all "clean up" passes to go there. The typechecker shouldn't really be generating English text directly any more.

mjgpy3:
Gotcha, thanks! Can you give me an insight into how to restrict to those? I started looking at `TypeConstructor` but I ended up with very ugly pattern match that reached deeply into `ProperName`.

paf31:
I think you can just test `== tyObject` and `== tyFunction`, after looking under `TypeConstructor`.

mjgpy3:
Latest push should be close

paf31:
Sorry to drag this out, but I don't really want to add even more constructors to `ErrorMessage` to deal with special cases. This sort of thing belongs in the error printer. As I hinted above: "I liked the previous approach of handling this in Errors, since eventually I'd like all "clean up" passes to go there." Matching particular classes of errors is already tricky.

mjgpy3:
Alright, will update, thanks

mjgpy3:
Sorry about all the back and forth on this, I hope the current push is reasonable :)

paf31:
No problem at all. I just want to make sure we get this right since I'd like to add more things like this. Thanks for the help!

mjgpy3:
@paf31 this isn't done, just curious if it matches what you're thinking -- if it looks okay, I'll handle the other wrapped-error types

paf31:
I think it's definitely on the right track :+1:

mjgpy3:
Is there a more concise way to unwrap those? I saw `simplifyErrorMessage` but I don't think it does exactly what I want.

paf31:
I don't think so, but it might be worth extracting an `unwrapErrorMessage` function or something like that, to pull off the outer data constructors.

mjgpy3:
Alright, will do. Thanks again!

mjgpy3:
Any feedback is greatly appreciated.

src/Language/PureScript/Errors.hs
```diff
@@ -261,6 +248,31 @@ data LabelType = TypeLabel | SkolemLabel String deriving (Show, Eq, Ord)
 -- | A map from rigid type variable name/unknown variable pairs to new variables.
 type UnknownMap = M.Map (LabelType, Unknown) Unknown
 
+-- |
+-- Extract nested error messages from wrapper errors
+--
+unwrapErrorMessage :: ErrorMessage -> Maybe ErrorMessage
+unwrapErrorMessage em = case em of
+  (ErrorCheckingKind _ err)       -> Just err
```
paf31: This isn't going to work for certain errors. You need to recursively call `unwrapErrorMessage` on the wrapped error until you get down to an unwrapped error. On the plus side, this means you can get rid of the `Maybe`.

mjgpy3:
@paf31: more like this? Do you know a way to ensure exhaustive matches in `errorCode` (or error in counter cases)? I was going to fail when the wrapped and unwrapped were equal, but I notice `ErrorMessage` isn't an instance of `Eq`. Is there a particular reason for that?

paf31:
Looks great. For the exhaustivity issue, I would use a `case` at the top level, and then add a wildcard match with an `error` call inside. A better approach, allowing us to make use of GHC's exhaustivity checks, would be to extract the simple errors into another data type (`SimpleErrorMessage` maybe?)

mjgpy3:
I'm getting a little confused, would it be enough to make `unwrapErrorMessage` exhaustive across `ErrorMessage` and have no wildcard matcher?

mjgpy3:
I hope you don't mind all this back and forth. To be honest, I am fully enjoying the discussion.

mjgpy3:
With the `SimpleErrorMessage` what type would it be, and how would it help the compiler detect non-exhaustive matching?

paf31:
No problem at all. So `SimpleErrorMessage` would have all of the constructors which do not wrap another error, and `ErrorMessage` would have all the others, and an extra constructor for wrapping a `SimpleErrorMessage`. This would have the downside that all calls to `throwError` might have to be modified slightly. Then `unwrapErrorMessage` would have type `ErrorMessage -> SimpleErrorMessage`, and it would be exhaustive, checked by GHC.

mjgpy3:
Excellent, thanks for the explanation. I see what I can do :+1:

mjgpy3:
Hey @paf31, let me know your thoughts.

I wanted to make a function summarizing:
```
throwError . errorMessage $ SimpleErrorWrapper
```

but I couldn't get the type signature to work and it's getting late, so maybe another day :)

paf31:
I think what you want is `throwError . errorMessage . SimpleErrorWrapper $ ...` but I would actually move the `SimpleErrorWrapper` constructor inside `errorMessage` since every call site seems to use it now.

Thank you for working on this - it's a really valuable change, since now we have the framework to give suggestions generally, which is something we really needed.

mjgpy3:
Awesome, thank you!

mjgpy3:
That's straing

mjgpy3:
Oops

mjgpy3:
@paf31, I pulled that constructor in. Let me know what you think

paf31:
Looks great. Could you please do the following and then I'll get this merged into master:

- Split `errorMessage` into `errorMessage` (what it is now) and `singleError = MultipleErrors . pure` to replace the many instances of `MultipleErrors . (: [])`.
- Make sure the tests still pass
- Revert the files which didn't need to change (the uses of `$`), just to keep the history clean.
- Merge master up, and squash your commits into a single commit.

Also, could you please show me the output of, say `:type const 1 . const 2` in `psci`?

Thanks!

PR #1093, opened 2015-05-02
core-imp updated with latest master

andyarvanitis:
Also includes my changes to `core-imp` from the other pull request, so this would supersede it. Everything seems fine after merging and fixing conflicts, but this could use a thorough review.


paf31:
:+1: Thanks!

paf31:
I'll let @garyb review and merge since this is his area.

andyarvanitis:
Sure!

PR #1095, opened 2015-05-02
Move JS codegen for compile into psc

garyb:
@paf31 how does this approach look for `psc`?

I really like the idea of using `require` for the `psc-make` foreigns, if we can figure out the details for that.

- [x] Load foreign files in `psci`
- [x] Don't use `error` for foreign-related errors in `psc`
- [x] Don't use `error` for foreign-related errors in `psc-make`
- [ ] Don't use `error` for foreign-related errors in `psci`
- [x] Error on duplicate foreign modules (multiple JS files with the same `// module X.Y.Z` comment)
- [x] Output deprecation message when encountering `foreign import` with literal
- [ ] Restore DCE for `psc`
- [x] Make tests work again


garyb:
A proposal for how `psc-make` might work with require (I mostly rambled this on IRC but not sure if you saw it):

1. We still use the method of a second command line option to accept FFI files
2. We still read these files for a `// module X.Y.Z` comment
3. During make, copy these js files to `output/X.Y.Z/foreign.js`
4. Then we just need to include `var _ffi = require("./foreign.js");` (or whatever for the var) in the `output/X.Y.Z/input.js` file.

paf31:
:+1: Looks nice, and I like the FFI suggestion.

I guess we can tidy up the `Options` structure if we manage to do a similar thing with `psc-make` too.

So what's the plan for `psc-make`? I guess if we want to use `require` for module imports, it seems like we're heading down the path of using `psc-make` as the main JS backend. That's good, but we would need to figure out a way to do dead code elimination afterwards. I like the idea of adding simple delimiting comments to the Common JS module, and also indicating dependencies. That way, `psc-make` could generate those comments too, and `psc-bundle` (or whatever we want to call it) could process those comments and do DCE. The comments might get a bit large though.

garyb:
Like an idiot I spent half the afternoon using `language-javascript` to try and extract the module comment out of the source, which worked, but was a ton of code as I made a traversal type thing for the AST, etc. and then as soon as I was done realised that just parsing the lines in the file would be a lot easier. Oh well.

I'm going out for a bit, but will be back later to get this working at last.

How does `$foreign` sound for the JS name for the `require`'d/embedded object containing the foreign definitions?

garyb:
- [ ] Load foreign files in `psci`
- [ ] Don't use `error` for foreign-related errors in `psc`
- [ ] Don't use `error` for foreign-related errors in `psc-make`
- [ ] Don't use `error` for foreign-related errors in `psci`
- [ ] Output deprecation message when encountering `foreign import` with literal
- [ ] Restore DCE for `psc`

garyb:
- [ ] Make `tests` work again

src/Language/PureScript/Parser/JS.hs
```diff
@@ -0,0 +1,48 @@
...
+      Just name -> return (name, file)
+      Nothing -> throwError (errorMessage $ ErrorParsingFFIModule path)
+  return $ M.fromList foreigns
+
+findModuleName :: [String] -> Maybe ModuleName
+findModuleName = msum . map parseComment
+  where
+  parseComment :: String -> Maybe ModuleName
+  parseComment s = either (const Nothing) Just $
+    lex "" s >>= runTokenParser "" (symbol' "//" *> reserved "module" *> moduleName <* PS.eof)
```
paf31: Using `TokenParser` here seems a little odd, since we're not really dealing with PS tokens. Maybe just using `lines`, and looking for `//` at the start is enough.

Also, maybe we want to define some type of custom comment, since it's not out of the question that a normal JS comment would begin with `// module`.

Finally, if we use `lines`, we can define a custom format for other metadata, like dependencies for DCE.

garyb: Yeah, I guess that's true. It is a convenient way of extracting a proper `ModuleName` from a string though.

This does run over all the lines in the file, so if there are multiple comments that look like `// module` it will only take the first one that actually parses as having a `ModuleName` following it.

paf31:
@garyb I fixed up `psci` and the test suites. All that's left now is hopefully DCE so let's discuss that when you have time. I think we should make a release candidate though, so that we can start updating libraries again, and so that @hdgarrood and @nicodelpiano have a reasonable place to send PRs for GSOC.

Fingers crossed that the CI build works ...

paf31:
Well, CI is passing apart from the `core-tests` step, which we expect to break right now.

garyb:
I'm happy to merge this then, if you are?

paf31:
Yep!

PR #1100, opened 2015-05-05

nicodelpiano:
Also, I added some tests to manual.

Sorry for the second PR, but I had some problems rebasing :(.


paf31:
Looks good, thanks!

I can't merge this though, it looks like you need to merge with `master`.

One more thing we need to do is to include position information by using `censor` at every `PositionedValue` or `PositionedDeclaration`. Including module information would also be nice. Maybe we can create a variant of `censor` which works for both errors and warnings. We can deal with these things later though.

PR #1101, opened 2015-05-05

nicodelpiano:
Also, I added some tests to manual.

Hope this is OK now :smile:

@paf31
In all, things to do:
* Add module and position information
* There's an issue when calling recursive functions (that have wildcard types). It seems like it duplicates the wildcard warnings. Perhaps for several calls to `replaceTypeWildcards`? For example, when running this test: `examples/passing/TypeWildcards.purs`

### DIFF
--- a/examples/manual/passing/ShadowedName.purs
+++ b/examples/manual/passing/ShadowedName.purs
@@ -1,0 +1,10 @@
+module Main where
+
+import Prelude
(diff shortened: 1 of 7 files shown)


paf31:
I think you need to merge master into your branch and resolve the conflicts, or the CI build won't run and I won't be able to merge.

PR #1105, opened 2015-05-06
Correct ConflictingImport error, add suggestion. Resolves #1088.

### LINKED ISSUE #1088: Conflicting import error seems to use the wrong module name
If we have:

```purescript
module ConflictingImport where

id :: Number
id = 1
```

Then this will fail with:

```
>> Error in module ConflictingImport:
>> Declaration id conflicts with import ConflictingImport
```

I think a better error message could be:

```
Cannot declare a value `id` when another declaration with that name was
imported from Prelude.
Possible fix: hide `id` in the Prelude import:
   import Prelude hiding (id)
```

mjgpy3:
Here's how the output for the error in #1088 looks now:
```
Error:
Error in module ConflictingImport:
Cannot declare `id` since another declaration of that name was imported from `Prelude`
Possible fix: hide `id` when importing `Prelude`:
  import Prelude hiding (id)
See https://github.com/purescript/purescript/wiki/Error-Code-ConflictingImport for more information, or to contribute content related to this error.
```

Should I also suggest qualified imports? I believe I saw an MR about those being added.


paf31:
:+1: Thank you, this has been bugging me for ages!

mjgpy3:
@paf31, glad to help! Let me know if it needs anything.

PR #1108, opened 2015-05-07
Add versionTag to psc-publish output

hdgarrood:
This allows us to get the README via the Github API on the Pursuit server.


paf31:
:+1:

Issue #990, opened 2015-03-21
Purescript default operator precedence


alexchandel:
So there are a lot of default operators in PureScript, some at the value level, some at the type level.
```purescript
->
<<<
>>>
$ # :
<$> <#>
<*>
>>=
+ * / -
== /=
< > <= >=
& | ^
&& ||
<> ++
`some_function` `mod` `shr` `shl`
= <-
```
From the documentation it's not immediately obvious what the priorities of the operators are. For example, I would expect something like this for the familiar binary operators:
```rust
* / `mod`
+ -
`shr` `shl`
&
^
|
== \= < > <= >=
&&
||
=
```
but that has two issues: it assigns multiple priorities to certain, special backtick functions (which makes no sense), and it leaves out half the operators (`-> <<< >>> $ # : <$> <#> <*> >>= <> ++ <-`). And in the mix is also the precedence of plain-old function invocation: `const "a" 1`.

What exactly are the operator precedences? And the associativities, when it's significant?

alexchandel:
Oh and **completely undocumented** is the modulus operator `%`. Where does it fit in?

alexchandel:
As well several syntactic tokens `@ . ... : , ::` and `\x ->` that can appear in expressions.

garyb:
That's a good point, this stuff isn't output in the documentation at the moment, although at least it's not too hard to look it up in the source.

Backticked functions all share the same precedence and fixity at the moment, which I think works out as `infixl 9`. I couldn't find that documented anywhere so we should probably add it to the wiki, but need to check that I got that right first!

`->` and `<-` aren't really operators, they're syntax. `:` is usually an operator, except when it appears in patterns, at which point it is syntax. That's something we've not fully figured out yet, native array patterns are possibly going to be removed due to their poor performance when treated like a cons list.

The `(->)` type is a special case and is the only operator that can appear at the type level (currently) so fixity is irrelevant there for now.

No operator can have a higher precedence than function application.

`(%)` should probably also go away when `Int` from `purescript-integers` is as efficient to use as `Number`. That hopefully won't be too long from now, when we add rewrite rules. It's a leftover from before we reformed the `Num` hierarchy, it's now monomorphic on `Number` and does whatever `%` does in JS, whereas the laws for `DivisionRing` mean that `mod` for `Number` always returns 0.

brainrape:
+1 for including the fixity declaration in the doc

alexchandel:
[...]

PR #1116, opened 2015-05-12
Collect fixity information in docs

hdgarrood:


garyb:
Nice!

hdgarrood:
I was only thinking about pursuit when I did this, but I guess it would be a good idea it to psc-docs too.

hdgarrood:
Fixes #990

src/Language/PureScript/Docs/AsMarkdown.hs
```diff
@@ -62,6 +64,21 @@ codeToString = outputWith elemAsMarkdown
   elemAsMarkdown (Keyword x) = x
   elemAsMarkdown Space       = " "
 
+fixityAsMarkdown :: P.Fixity -> Docs
+fixityAsMarkdown (P.Fixity associativity precedence) = do
+  tell' $ concat [ "associativity: **"
```
paf31: Why not just use the existing notation (`infixl 4`), like in Haddocks?

paf31:
Looks nice, thanks!

hdgarrood:
I suppose we should also give operators the default fixity information if there is no declaration? Is that `infixl 9`, like Haskell?

src/Language/PureScript/Docs/AsMarkdown.hs
hdgarrood: I think using English words makes it a bit easier to understand. Given that this concept exists outside of PureScript and Haskell, I would guess that more people understand the term "left-associative" than "infixl", for example. And if you do understand "infixl", you probably also understand "left-associative".

If you were referring to putting it inline with the code like Haddock does, I can't see a good way of doing that in Markdown that preserves some kind of visual separation between the type signature and the fixity declaration, unfortunately.

paf31: Ok, sounds good.

hdgarrood: Come to think of it, "associativity: left-associative" is a bit redundant. I think it would be a bit nicer to just say "left-associative" (assuming we stick with the English style).

hdgarrood:
It looks like, if you don't provide a fixity declaration for an operator, it becomes left-associative with a precedence lower than 0. This appears impossible to do with an explicit declaration, since it's a parse error to provide a negative number.

See: https://gist.github.com/hdgarrood/5ee5ccf6310cade96fd3

paf31:
@hdgarrood Could you please see how much work it would be to rebase this against latest master, and I'll merge it in for the 0.7 release? Thanks.

hdgarrood:
Sure, I'll be able to do that on Tuesday. I don't think it will be too hard to rebase.

hdgarrood:
Rebased, I think this is good to go. cc @paf31

paf31:
:fireworks: Thanks!

PR #1117, opened 2015-05-13
Allow psc-docs to write multiple modules to a file

### LINKED ISSUE #1113: Support writing multiple modules to the same file in psc-docs
So you can do something like this:
``` 
psc-docs ... --docgen X:docs/X.md --docgen X.Y:docs/X.md --docgen X.Z:docs/X.md
```
Ideally it should preserve the order the --docgen options are listed in also, or maybe just sort hierarchically and alphabetically, so `X` is above `X.Y` in the output file.

garyb:
Resolves #1113

### DIFF
--- a/psc-docs/Main.hs
+++ b/psc-docs/Main.hs
@@ -16,11 +16,13 @@
 module Main where
 
 import Control.Applicative
-import Control.Monad.Writer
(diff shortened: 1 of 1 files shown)


paf31:
:+1:

Issue #1125, opened 2015-05-24
Numeric literals with leading zeros


hdgarrood:
This is a little surprising:

```
>> 010 == 10
true
```

In most other languages that I'm aware of, including JavaScript, a leading 0 indicates an octal literal, so `010 == 8`. I think we should either disallow leading zeros, or interpret them as octal.

I prefer the first option; if I ever wanted an octal literal, I think I'd rather do something like `readInt 8 "777" ` (from `purescript-globals`), or simply write the decimal version in the code and add a comment with the octal version.

michaelficarra:
Agreed, disallow leading zeros.

garyb:
Another :+1: for disallow leading zeros.

michaelficarra:
Let's add the `easy` tag to this one.

paf31:
:+1:

garyb:
Closed by #1127

PR #1127, opened 2015-05-24
Numeric literals with leading zeros #1125

kRITZCREEK:
This feels a little hacky. If you have suggestions for a nicer way I'd love to improve my Parsec-Fu. 
I'll add a failing test aswell.

### DIFF
--- a/examples/failing/LeadingZeros1.purs
+++ b/examples/failing/LeadingZeros1.purs
@@ -1,0 +1,3 @@
+module Main where
+
+x = 01
--- a/examples/failing/LeadingZeros2.purs
+++ b/examples/failing/LeadingZeros2.purs
@@ -1,0 +1,3 @@
+module Main where
+
+x = 00.1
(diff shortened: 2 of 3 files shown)


garyb:
Seems reasonable to me, might want a better message than "intended" though? :wink:

kRITZCREEK:
That error message never gets shown. I only need the parser to fail without consuming input so that choice tries the next one.

garyb:
Ah, interesting. I assumed that at first, but wanted to check - it had no description in the documentation and reading the implementation suggests it is used, I guess not for the resulting error message. Would `parserZero/mzero/empty` work instead?

kRITZCREEK:
Of course you're right! That's what I'm looking for.

kRITZCREEK:
A parse failure does now error as follows:

``` 
> 02
(line 1, column 3):
unexpected '2'
expecting no leading zero in number literal
```

I think that should do it.

garyb:
Thanks!

Issue #1008, opened 2015-03-28
Bug in code gen


paf31:
I think this was added relatively recently (maybe the case decl. changes):

```purescript
instance semigroupMaybe :: (Semigroup a) => Semigroup (Maybe a) where
  (<>) Nothing a = a
  (<>) a Nothing = a
  (<>) (Just a) (Just b) = Just (a <> b)
```

becomes

```javascript
    var semigroupMaybe = function (__dict_Semigroup_19) {
        return new Prelude.Semigroup(function (a) {
            return function (a_1) {
                if (a_1 instanceof Nothing) {
                    return a_1;
                };
                if (a_1 instanceof Nothing) {
                    return a_1;
                };
                if (a_1 instanceof Just && a_1 instanceof Just) {
                    return new Just(Prelude["<>"](__dict_Semigroup_19)(a_1.value0)(a_1.value0));
                };
                throw new Error("Failed pattern match");
            };
        });
    };
```

PR #1132, opened 2015-05-28
Fix preservation of binding names in functions (introduced in 703dd66)

andyarvanitis:
I happened to notice that this got broken at some point, so this is a quick fix for it.


paf31:
What case does this fix?

michaelficarra:
Please add tests so we don't have another regression.

paf31:
Sorry, to be clear, I rewrote this at one point since it was causing some errors. I decided to err on the side of caution, and it's still a little brittle. A test would definitely be appreciated, although I'm pretty sure I added some at the time.

andyarvanitis:
Yeah, sorry, it just occurred to me that a test is needed for it. The case I found is just the fibonacci:
```
  fib :: Int -> Int
  fib 0 = 0
  fib 1 = 1
  fib n = fib (n - 2) + fib (n - 1)
```
Do you recall what kind of errors it created?

paf31:
Basically it would just try to bind the same name twice. If you `git blame` that file, you should see my notes and hopefully a test somewhere.

andyarvanitis:
Ok, I'll go over it more thoroughly tomorrow then. Sorry it introduced errors in the first place; I thought I had worked through and tested it carefully enough, but clearly not!

paf31:
No problem. Looks like the tests are passing at least. I'll have a look tomorrow too.

andyarvanitis:
So testing for breakage (e.g. the same name appearing twice) is easy enough (and Phil has indeed added a test for it), but checking for a correct pretty name is a bit tricky, at least without inline access to `eval`*. My js is pretty weak, though, and I'm not finding alternatives on SO, etc.. Any ideas?

\* You can't alias `eval`.

andyarvanitis:
Never mind, looks like it's trivial to get the function itself as a string in js, so making a test should be easy enough.

andyarvanitis:
Is there anything else you guys would like to see added to this before it's ready to merge?

paf31:
Sorry, right now I'm trying to get the external FFI stuff merged. I'll get onto this ASAP.

Your test currently passes on master though. Is there a test where this passes but master fails? I don't quite understand the change.

andyarvanitis:
No rush at all, I was just tying up some loose ends.

Hmm, it failed (as it should have) when I ran it on a branch without this fix -- let me have a look at it again.

paf31:
I just ran the code through Try PureScript and it rendered the argument as "n".

andyarvanitis:
Weird, I just cloned a fresh master and hand-tested using `pcc`, and got the "ugly" names. Running the automated test now.

andyarvanitis:
Ok, failed for me (fresh clone). I'll see if I can think of what might be going on.

paf31:
Maybe it got broken between those two versions?

paf31:
Try PureScript is on 0.6.9.3.

andyarvanitis:
Yeah, I was just dealing with master/0.7.0. I'm not surprised about TryPS/0.6 working.

andyarvanitis:
At least we know the test works!

paf31:
The `Just Nothing` special case is no longer used. If this does work, we should squash down the two layers of `Maybe` to a single layer.

andyarvanitis:
Issue #1008 covers the work where the "pretty" case got affected -- milestone 0.7.0.

andyarvanitis:
Agreed; it actually started out that way, so easy enough to restore it (but of course keeping the added uniqueness check).

andyarvanitis:
I still need to update the test (break out FFI parts) and move it.

andyarvanitis:
Updated test. Please let me know if adding it to the test prelude this way is ok.

tests/prelude/Prelude.purs
```diff
@@ -556,6 +558,8 @@ module Prelude.Unsafe where
 
   foreign import unsafeIndex :: forall a. Array a -> Number -> a
 
+  foreign import reflectParameterName :: (Int -> String) -> String
```
paf31: Can we make this `(a -> b) -> String`? Seems like it might be useful for other tests one day.

tests/prelude/Prelude.purs
```diff
@@ -548,6 +548,8 @@ module Assert where
 
   foreign import assertPartial :: forall e a. (Unit -> a) -> Eff (assert :: Assert | e) Unit
 
+  foreign import assertEqual :: forall eff a. a -> a -> Eff eff Unit
```
paf31: Why not write this in PureScript using `Eq`?

tests/prelude/Prelude.purs
andyarvanitis: Yeah, good idea, I just broke it out from the specific test quickly.

tests/prelude/Prelude.purs
andyarvanitis: Ah true, I borrowed it from another test and didn't think of it.

andyarvanitis:
The build/test failure seems unrelated to the changes. Has anyone seen the following before?:
```
### Error in:   completionTests:0
/home/travis/build/purescript/purescript/tests/prelude/prelude.purs: openFile: does not exist (No such file or directory)
### Error in:   completionTests:1
/home/travis/build/purescript/purescript/tests/prelude/prelude.purs: openFile: does not exist (No such file or directory)
### Error in:   completionTests:2
...
```

paf31:
Well I changed `prelude.purs` to `Prelude.purs`. That might be related.

andyarvanitis:
Yeah, probably. The running it locally on my mac (case-insensitive).

andyarvanitis:
Fixing the `Prelude.pus` name change problem in another (forthcoming) pull request.

andyarvanitis:
No rush, but in case it wasn't clear, this is ready to go as of the merge from master in b4b17ff.

paf31:
I'll try to get this in before 0.7, I'm just trying to convince myself that the extra case isn't needed.

PR #1134, opened 2015-05-31
core-imp + latest master + external-ffi-files

andyarvanitis:
As mentioned briefly on IRC, this is what I'm using for the latest pure11/14 work. This pull request is mostly for review purposes, although it's working well for me so far.


garyb:
Can you open this as a PR to `core-imp` instead please? Also it looks like it won't merge now, I think that's after I made the change to remove the implicit Prelude.

andyarvanitis:
Sure -- I'll open a new PR, since I can't find a way to switch merge targets.

Issue #647, opened 2014-10-24
typeclass deriving for Eq, Ord, Show


jfilip:
as per Haskell

gbaz:
This is fully developed in my branch of 0.6.9.5: https://github.com/gbaz/purescript/tree/deriving-generic

The code needs to be accompanied by a library module Generics.hs as given in this gist: https://gist.github.com/gbaz/a2c867b6c07731a403eb

I'd like to get this reviewed and merged -- not sure what branch to request a pull into, etc.

It would also be nice to develop more examples besides the show and the toFrom -- to/from json, eq, ord, etc. I'm fairly confident that the API is expressive enough for all the typical generics examples, but working a few more out would be nice.

paf31:
See Generic Deriving in 0.7.3.

PR #1138, opened 2015-06-03
Generic Deriving

gbaz:
Addresses issue #647 among other features.

Works hand in hand with userland library code as in https://gist.github.com/gbaz/a2c867b6c07731a403eb


gbaz:
(Also note the silliness in how I had to write the Proxy type in the gist. This may be fixed in 7, but in 6.9.5 doing the obvious phantom type thing failed)

joneshf:
This is awesome!

gbaz:
The Gist is now mainly updated to work with purescript 0.7. Either the typechecker got better or the explicit ascription trick for Proxy worked or both, as now I can do things the sane way.

Since it was a very quick translation, the code for genericShow is now especially ugly -- but it was always included simply as a demonstration and not a proper implementation anyway.

gbaz:
The library component to go with this now lives here, btw: https://github.com/paf31/purescript-generics/pull/3

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -0,0 +1,150 @@
...
+pattern ObjectType rec = TypeApp (TypeConstructor (Qualified (Just (ModuleName [ProperName "Prim"])) (ProperName "Object"))) rec
+
+lam :: String -> Expr -> Expr
+lam s = Abs (Left (Ident s))
+
+lamCase :: String -> [CaseAlternative] -> Expr
+lamCase s = lam s . Case [mkVar s]
+
+liftApplicative :: Expr -> [Expr] -> Expr
+liftApplicative = foldl' (\x e -> App (App (mkPrelVar "ap") x) e)
```
paf31: Maybe use `apply` here to avoid an indirection via `Monad` in the generated code?

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -229,8 +229,29 @@ typeClassMemberToDictionaryAccessor _ _ _ _ = error "Invalid declaration in type
...
 
+fillDerivingDecls :: (MonadError MultipleErrors m, Applicative m) => Qualified ProperName -> [Type] -> [Declaration] -> [Declaration] -> Desugar m [Declaration]
+fillDerivingDecls className tys tyDecls decls = do
+  let missingDecls = mapMaybe declName tyDecls \\ mapMaybe declName decls
+  let mkMissingDecl cls fun =
+          case tys of
+            [TypeConstructor t] | unQualify className == ProperName cls && Ident fun `elem` missingDecls -> return [ValueDeclaration (Ident fun) Public [] (Right $ TypeClassInstanceMemberFunction (Ident fun) className t)]
+            _ -> return []
+  extraDecls <- concat <$> mapM (uncurry mkMissingDecl)
+                  [("Generic","toSpine")
```
paf31: Typically, function names out of the Prelude go into the `Constants` module. Maybe these should go there too.

paf31:
Sorry this PR has been left unmerged for a while. Here are my thoughts on it:

- If it's merged, I'd like to fix the couple of minor issues I mentioned above.
- I'm nervous about modifying the typechecker to add new features at this point. I wouldn't be so nervous if the typechecker were better, so maybe this will no longer be an issue after it's updated (whenever that happens). Ideally though, I'd find a solution which doesn't involve the new member placeholder.
- I feel like this would be a good example of an external tool, and doesn't necessarily need to be added to the compiler. Support for specific libraries probably doesn't belong in the compiler if we have a nice way to run external tools as part of the build. Template programing would be ideal, but maybe there is something simpler which can do the job? Even something as simple as a step in a Makefile which runs the tool and modifies the source files. Maybe see `purescript-derive-lenses` as an example of that approach (although, I admit it's slightly different because you can't create another module here, without creating orphan instances)
- Is there a reasonable variant of this and `purescript-generics` which doesn't use type classes? Something like

```purescript
newtype Generic a = Generic 
  { toSpine :: a -> Spine 
  , toSignature :: Proxy a -> GenericSignature
  , fromSpine :: GenericSpine -> Maybe a
  }

showGeneric :: forall a. Generic a -> a -> String
```

This way, wrapping up `Generic` in a type class would be simple, writing instances would be trivial, and we could move this PR into an external tool without worrying about generating orphans.

gbaz:
I'd be happy to make the changes requested if we can agree that then this could be merged :-)

On the concern that this modifies the typechecker... I understand this. However, bear in mind that the only change to the typechecker proper is a "one liner" as follows:

+check' e@TypeClassInstanceMemberFunction{} ty = flip check' ty =<< elaborateInstance e

This should be really robust to virtually any refactor. If there was an "elaboration" phase where we had more information available but prior to the checking phase, then the checker wouldn't need to be changed at all, imho.

I've been thinking about this for a few ways now, and given a desire to _not_ have orphans, I don't see how this can be a standalone tool :cry: 

In particular, we don't just want to do "instance Generic Foo" but then also "instance Show Foo where show = genericShow" etc.

So being able to derive generic in the same file as the definition isn't just about putting it to work elsewhere, but then piggybacking on that to derive a lot of _other_ instances. And since those _other_ instances need to be "real" instances not handed a manual dictionary -- i.e. so that the Show instance can be used by something that shows pairs of things, etc, then there isn't much of a way around this...

I think the power/weight ratio remains good here, since the idea is that by deriving Generic, then "once and forall" a whole bunch of userland things become possible at once, and the net effect is to take a whole bunch of other potential feature requests and allow them to be undertaken in userland instead.

So if you'll agree to merge once I make the requested changes, I'll proceed?

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -229,8 +229,29 @@ typeClassMemberToDictionaryAccessor _ _ _ _ = error "Invalid declaration in type
 unit :: Type
 unit = TypeApp tyObject REmpty
 
+fillDerivingDecls :: (MonadError MultipleErrors m, Applicative m) => Qualified ProperName -> [Type] -> [Declaration] -> [Declaration] -> Desugar m [Declaration]
+fillDerivingDecls className tys tyDecls decls = do
+  let missingDecls = mapMaybe declName tyDecls \\ mapMaybe declName decls
+  let mkMissingDecl cls fun =
+          case tys of
+            [TypeConstructor t] | unQualify className == ProperName cls && Ident fun `elem` missingDecls -> return [ValueDeclaration (Ident fun) Public [] (Right $ TypeClassInstanceMemberFunction (Ident fun) className t)]
```
paf31: I think we'd want to match the whole qualified name, since having something called `Generic` in a different not-generics module shouldn't apply, right?

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -0,0 +1,150 @@
...
+import Control.Monad.Error.Class (MonadError(..))
+
+import Language.PureScript.AST
+import Language.PureScript.Environment
+import Language.PureScript.Errors
+import Language.PureScript.Names
+import Language.PureScript.TypeChecker.Monad
+import Language.PureScript.Types
+
+elaborateInstance :: (Functor m, MonadState CheckState m, MonadError MultipleErrors m) => Expr -> m Expr
```
paf31: Could you please add a short comment to this, since it's exported?

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -0,0 +1,150 @@
...
+elaborateInstance (TypeClassInstanceMemberFunction funName className t) =
+    fromMaybe noInstanceError $ lookup (unQualify className, funName) derivingList
```
paf31: Same comment as above about qualified names.

gbaz:
I take it these additional comments indicate a willingness to pull?

The qualified names choice was deliberate, but is open to discussion.

The idea of pulling whatever "Generic" is in scope is to let users swap in whatever Generic class they want, if they don't like the version in that particular package -- as long as instances can be given in the same way, you could write something that worked over an alternate sort of skeleton datatype, etc.

I wanted to be as general-purpose and agnostic as possible.

If you think that this will be more confusing than useful, then I can of course change it to specify.

paf31:
I think you're right about the power-to-weight ratio being well worth it. I put a couple of other minor comments on the code above.

Also, I guess we'd need to update the `purescript-generics` library again for 0.7?

I don't think the qualified name thing is about "what's in scope" so much as, what if someone doesn't know this feature exists and defines a `Generic` class, then they'll get errors because we tried to define members which didn't exist? We can always just switch out the implementation of `Data.Generic.Generic` by replacing the package entirely, but maybe that's a versioning nightmare waiting to happen. Another ugly solution is to add a command line argument for the Generics module you want to use, with a default of `Data.Generic`.

And yes regarding willingness to pull :smile: 

Could you please add a link to the MIT license to the top of your new file as well? We're probably going to go through and do this for other files too. I know it says `LICENSE: MIT` right now, but since the copyright ascription is different from the other files, it seems to make sense to be clear about how this file in particular is licensed.

gbaz:
Ok. So I'll go ahead and try to clean all the above. And I'll go ahead and add qualified names for now to point to "Data.Generic.Generic". We can always revisit and generalize later.

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -0,0 +1,150 @@
...
+  let prodConstructor = App (Constructor $ findName "SProd" ctors)
+      recordConstructor = App (Constructor $ findName "SRecord" ctors)
+      genericModule = case findName "Generic" tcs of (Qualified mn _) -> mn
+
+      mkCtorClause (ctorName, (_,_,typ,idents)) = CaseAlternative [ConstructorBinder ctorName (map VarBinder idents)] (Right caseResult)
+          where caseResult = App (prodConstructor (StringLiteral . runProperName . unQualify $ ctorName)) . ArrayLiteral $ zipWith toSpineFun (map (Var . Qualified Nothing) idents) (argTypes typ)
+
+      toSpineFun :: Expr -> Type -> Expr
+      toSpineFun i (ObjectType rec) =
+          lam "q" . recordConstructor . ArrayLiteral .
```
paf31: Might this `q` shadow some other name? Why not use a `NullBinder`?

gbaz: didn't knwo about nullbinders is why :smile:. That said, the shadowing won't hurt afaik because there shouldn't be any names inside the thing we're building... But I'll go put in null binders where appropriate.

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -0,0 +1,150 @@
...
+              map (\(str,typ) -> ObjectLiteral [("recLabel", StringLiteral str), ("recValue", toSpineFun (Accessor str i) typ)])
+              $ decomposeRec rec
+      toSpineFun i _ = lam "q" $ App (mkVarMn genericModule "toSpine") i
+
+  return . lamCase "x" . map mkCtorClause . constructorsForType t $ ctors
```
paf31: `x` is used, so we can't use `NullBinder` to avoid shadowing warnings, but maybe we can use a name which is valid JS but not valid PS, so nobody can use the same identifier for a top-level name. Dollar symbols are the usual hack here.

src/Language/PureScript/TypeChecker/Deriving.hs
paf31: true but you might shadow something at the top level.

paf31: Not that it matters, but it will generate a warning.

gbaz:
Ok, changed things vis a vis constants and null binders. Still have yet to force only using "Data.Generics" as the modulename, but otherwise things seem to still be working.

gbaz:
I think I've made all requested changes and tested. To integrate with the main testsuite it would make sense to have the new generic module on bower since that's where the testsuite seems to pull the modules from, so that can be a second pass.

paf31:
I'll merge your `purescript-generic` changes soon so that we can add a test to the test suite here.

paf31:
Also, thanks! :+1:

paf31:
@gbaz I'm getting ready to merge this, thanks. From above:

> Could you please add a link to the MIT license to the top of your new file as well? We're probably going to go through and do this for other files too. I know it says LICENSE: MIT right now, but since the copyright ascription is different from the other files, it seems to make sense to be clear about how this file in particular is licensed.

gbaz:
I haven't added something to use proxy from the proxy library yet. To be honest I'd just like to tie off this pull request as is and do that as a separate patch next :-)

paf31:
Yep, sounds good. Thanks :+1:

paf31:
I'll merge this and then work on adding tests later today.

paf31:
Hmm, CI build is failing on 7.6.3 because of `PatternSynonyms`. I think we'd prefer to keep 7.6.3 compatibility if we can. Could these patterns be inlined, or could we use `PatternGuards` or something else instead if they're too unweildy?

gbaz:
done

paf31:
Let's see what CI says this time...

PR #1149, opened 2015-06-07
Fix #1147, do not inline fromNumber

paf31:
Resolves #1147. @garyb Does this look ok?

Thanks.


garyb:
#1147 is already a PR :smile:

paf31:
Oh, sorry :D

PR #1155, opened 2015-06-09
Fix #1152, support globs in file paths

### LINKED ISSUE #1152: Alternative way of providing arguments to psc
It appears Windows has a limit on the length of command line arguments. I thought it was an issue with the gulp plugin, but it's not just that: purescript-contrib/gulp-purescript#36.

One option is to allow a file input to parse the commands from, or perhaps we could accept glob paths in the executables themselves - it would certainly make for a friendlier interface for people not using pulp, gulp, etc.

This is going to become a real problem for me soon, as `slamdata` is very close to exceeding that limit :wink:

paf31:


garyb:
This is _awesome_. :tada:

I guess we should do the same for the source inputs for `psc-docs`, source and foreign inputs for `psci` also?

paf31:
Yeah, `psci` is a little trickier, but not much.

paf31:
@garyb I think this should work for `psci` and `psc-docs`.

PR #1157, opened 2015-06-10
Fix Docs Render compilation with GHC 7.10.1

puffnfresh:
GHC 7.10.1 infers this type for `isOp`:

    (Eq a, Data.String.IsString [a]) => [a] -> Bool

Just needed [a] to be fixed to String.


paf31:
Thanks!

PR #1160, opened 2015-06-11
Improve performance of type-class dict lookup

puffnfresh:
We previously filtered ALL type-class dictionaries
every time we had an import.

We previously stored dictionaries in a Map keyed
by identifier and module name. We never actually
used that key.

We now store a list of dictionaries by module name
and look them up by that name when we have an
import.

Here's some numbers from Criterion:

    benchmarking unpatched
    time                 142.9 s    (133.9 s .. 149.8 s)
                         1.000 R²   (0.998 R² .. 1.000 R²)
    mean                 145.8 s    (144.0 s .. 146.9 s)
    std dev              1.699 s    (0.0 s .. 1.938 s)
    variance introduced by outliers: 19% (moderately inflated)

    benchmarking patched
    time                 56.14 s    (41.02 s .. 68.60 s)
                         0.992 R²   (0.969 R² .. 1.000 R²)
    mean                 58.99 s    (56.64 s .. 60.73 s)
    std dev              2.642 s    (0.0 s .. 3.008 s)
    variance introduced by outliers: 19% (moderately inflated)

### DIFF
--- a/src/Language/PureScript/CodeGen/Externs.hs
+++ b/src/Language/PureScript/CodeGen/Externs.hs
@@ -107,7 +107,7 @@ moduleToPs (Module _ moduleName ds (Just exts)) env = intercalate "\n" . execWri
 
     exportToPs (TypeInstanceRef ident) = do
       let TypeClassDictionaryInScope { tcdClassName = className, tcdInstanceTypes = tys, tcdDependencies = deps} =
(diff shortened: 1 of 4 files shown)


jdegoes:
:fireworks: Nice improvement!

paf31:
:100: Thanks!

PR #1162, opened 2015-06-12
Optimise replaceAllTypeSynonyms

puffnfresh:
Now does a single Type tree traversal and replacement using a Map,
instead of a traversal per element in the Map.

Criterion numbers:

    benchmarking unpatched
    time                 33.37 s    (31.62 s .. 37.20 s)
                         0.998 R²   (0.997 R² .. 1.000 R²)
    mean                 33.30 s    (32.74 s .. 33.67 s)
    std dev              556.4 ms   (0.0 s .. 642.5 ms)
    variance introduced by outliers: 19% (moderately inflated)

    benchmarking patched
    time                 19.63 s    (13.87 s .. 24.89 s)
                         0.990 R²   (0.961 R² .. 1.000 R²)
    mean                 19.84 s    (18.75 s .. 20.65 s)
    std dev              1.241 s    (0.0 s .. 1.408 s)
    variance introduced by outliers: 19% (moderately inflated)

### DIFF
--- a/src/Language/PureScript/TypeChecker/Synonyms.hs
+++ b/src/Language/PureScript/TypeChecker/Synonyms.hs
@@ -41,28 +41,22 @@ import Language.PureScript.Types
 -- |
 -- Build a type substitution for a type synonym
 --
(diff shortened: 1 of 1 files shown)


paf31:
Very nice!

PR #1164, opened 2015-06-13
Allow a prefix to be specified for require() paths

garyb:
Resolves #938 (at last)!

At least the option is here now, but by default we stick to non-relative imports.


garyb:
@paf31 bump!

paf31:
:+1: Looks good!

PR #1167, opened 2015-06-16
Fix 1166, Use bind instead of >>= in do desugaring

### LINKED ISSUE #1166: Use bind instead of >>= in do desugaring
No reason to use `>>=` any more, and this should give slightly nicer errors.

paf31:
Fix #1166
@garyb Could you please review this? I think it should lead to slightly nicer errors and generated code.

### DIFF
--- a/examples/passing/Do.purs
+++ b/examples/passing/Do.purs
@@ -16,8 +16,8 @@ instance applicativeMaybe :: Applicative Maybe where
   pure = Just
 
 instance bindMaybe :: Bind Maybe where
-  (>>=) Nothing _ = Nothing
-  (>>=) (Just a) f = f a
+  bind Nothing _ = Nothing
+  bind (Just a) f = f a
 
 instance monadMaybe :: Prelude.Monad Maybe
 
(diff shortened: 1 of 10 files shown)


paf31:
Hmm, I'm not sure. Do you think so?

garyb:
Just as I figure the optimization is more "magic `Eff` monad" than "magic `do`", but I don't feel strongly about it.

mjgpy3:
How does using `bind` improve errors and generated code?

paf31:
Well, `bind` is a bit nicer to look at than `>>=`, and "bind was not implemented" reads better than ">>= was not implemented" IMO.

mjgpy3:
Gotcha! Thanks for the answer :) I was unsure if it was more complicated than that

paf31:
@garyb I'd rather change `>>=` to `bind` separately later on. Does this look ok otherwise?

garyb:
That makes sense.

Are there any failing tests that need updating too? We should probably do a check to see if they're all failing in the way they're supposed to actually, after all the recent changes.

Looks good to me though. :+1:

paf31:
Yeah, it would be good to verify that things fail in the right way, which should be easier now that we have an actual error type to match on. I'll write something up for that.

PR #1168, opened 2015-06-16
Include more rendered code in child declarations

hdgarrood:
The Hoogle output needs the following additional bits of rendered code
for child declarations:

* for data constructors: their name and type. For example: `Just ::
  forall a. a -> Maybe a`
* for type class members: their 'real' type, not in the context of
  the type class. For example: `add :: forall a. (Semiring a) => a -> a
  -> a` rather than just `a -> a -> a`

This commit modifies the RenderedChildDeclaration data type to
accommodate these extra bits of rendered code, as well as updating the
documentation renderer to produce these additional values.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -42,7 +42,7 @@ library
                    aeson >= 0.8 && < 0.9,
                    vector -any,
                    bower-json >= 0.7,
-                   aeson-better-errors >= 0.7,
+                   aeson-better-errors >= 0.8,
                    bytestring -any,
                    text -any,
                    split -any
--- a/src/Language/PureScript/Docs/Render.hs
+++ b/src/Language/PureScript/Docs/Render.hs
@@ -150,7 +150,6 @@ renderDeclaration (P.DataDeclaration dtype name args ctors) title =
   typeApp  = foldl P.TypeApp (P.TypeConstructor (P.Qualified Nothing name)) (map toTypeVar args)
   code = keyword (show dtype) <> sp <> renderType typeApp
   children = map renderCtor ctors
-  -- TODO: Comments for data constructors?
   renderCtor (ctor', tys) =
     RenderedChildDeclaration (show ctor') Nothing Nothing (ChildDataConstructor ctorSignature ctorCode)
     where
@@ -200,7 +199,6 @@ renderDeclaration (P.TypeClassDeclaration name args implies ds) title = do
 
   children = map renderClassMember ds
 
-  -- TODO: Comments for type class members?
   renderClassMember (P.PositionedDeclaration _ _ d) = renderClassMember d
   renderClassMember (P.TypeDeclaration ident' ty) =
     RenderedChildDeclaration (show ident') Nothing Nothing (ChildTypeClassMember contextualType actualType)
--- a/src/Language/PureScript/Docs/Render.hs
+++ b/src/Language/PureScript/Docs/Render.hs
@@ -152,9 +152,14 @@ renderDeclaration (P.DataDeclaration dtype name args ctors) title =
   children = map renderCtor ctors
   -- TODO: Comments for data constructors?
   renderCtor (ctor', tys) =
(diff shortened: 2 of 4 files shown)


purescript.cabal
```diff
@@ -42,7 +42,7 @@ library
                    aeson >= 0.8 && < 0.9,
                    vector -any,
                    bower-json >= 0.7,
-                   aeson-better-errors >= 0.7,
+                   aeson-better-errors >= 0.8,
```
hdgarrood: This version bump is for the `throwCustomError` function, which I realised should be added to aeson-better-errors while I was writing this code, so I just added it and released a new version

src/Language/PureScript/Docs/Render.hs
```diff
@@ -152,9 +152,14 @@ renderDeclaration (P.DataDeclaration dtype name args ctors) title =
   children = map renderCtor ctors
   -- TODO: Comments for data constructors?
   renderCtor (ctor', tys) =
-          let typeApp' = foldl P.TypeApp (P.TypeConstructor (P.Qualified Nothing ctor')) tys
-              childCode = renderType typeApp'
-          in  RenderedChildDeclaration (show ctor') Nothing childCode Nothing ChildDataConstructor
+    RenderedChildDeclaration (show ctor') Nothing Nothing (ChildDataConstructor ctorSignature ctorCode)
```
hdgarrood: I've been referring to how a data constructor appears in the data declaration as a "signature", eg `Just a`, `Right b`, `Tuple a b`. The overlap with the more common meaning of "signature" in this context is undesirable, but I've been unable to come up with a better alternative. Any ideas?

paf31: I think "signature" is as good as anything else I can think of :)

src/Language/PureScript/Docs/Render.hs
```diff
@@ -196,16 +201,16 @@ renderDeclaration (P.TypeClassDeclaration name args implies ds) title = do
 
   children = map renderClassMember ds
 
-  -- TODO: Comments for type class members
+  -- TODO: Comments for type class members?
```
paf31: Is this still TODO?

src/Language/PureScript/Docs/Render.hs
```diff
@@ -152,9 +152,14 @@ renderDeclaration (P.DataDeclaration dtype name args ctors) title =
...
   renderCtor (ctor', tys) =
-          let typeApp' = foldl P.TypeApp (P.TypeConstructor (P.Qualified Nothing ctor')) tys
-              childCode = renderType typeApp'
-          in  RenderedChildDeclaration (show ctor') Nothing childCode Nothing ChildDataConstructor
+    RenderedChildDeclaration (show ctor') Nothing Nothing (ChildDataConstructor ctorSignature ctorCode)
+    where
+    ctor'' = P.TypeConstructor (P.Qualified Nothing ctor')
+    typeApp' = foldl P.TypeApp ctor'' tys
+    ctorSignature = renderType typeApp'
+    -- TODO: add forall for all free type variables
```
paf31: still todo?

src/Language/PureScript/Docs/Render.hs
hdgarrood: The question mark should have been there originally; it means 'decide if we even want these, and then act accordingly'. I'm leaning towards no, but don't feel strongly.

src/Language/PureScript/Docs/Render.hs
hdgarrood: not any more, the code two lines below this comment does this. whoops! will fix.

src/Language/PureScript/Docs/Render.hs
paf31: I think I'd agree with no.

hdgarrood: Ok - I've removed this. I also removed a similar one about data constructors having their own documentation.

PR #1172, opened 2015-06-16
Add psc-publish warning for unacceptable versions

hdgarrood:
A problem arose with a package that depended on purescript-maybe, and
was using the version 0.3.0-rc.1. Haskell's Data.Version refuses to
parse this as a Version, which means purescript-maybe was missing from
the resolvedDependencies.

Now, if the resolved Bower version for dependencies is not parseable
into a Version, a warning will be emitted, although it will still be
missing from resolvedDependencies.

In the longer term, we should look at using a Version data type that
supports semver properly, which would probably eliminate this issue
completely.

### DIFF
--- a/psc-publish/ErrorsWarnings.hs
+++ b/psc-publish/ErrorsWarnings.hs
@@ -1,10 +1,14 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module ErrorsWarnings where
 
+import Control.Applicative ((<$>))
 import Data.Aeson.BetterErrors
 import Data.Version
 import Data.Maybe
+import Data.Monoid
+import Data.Foldable (foldMap)
 import Data.List (intersperse)
 import Data.List.NonEmpty (NonEmpty(..))
 import qualified Data.List.NonEmpty as NonEmpty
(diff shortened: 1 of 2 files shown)


psc-publish/ErrorsWarnings.hs
```diff
@@ -247,32 +252,39 @@ displayOtherError e = case e of
...
-    ]
-  where
-  collectWarnings patterns warns =
-    let boxes = mapMaybe (collectWarnings' warns) patterns
-        result = vcat
-                    [ para "Warnings:"
-                    , indented (vcat (intersperse spacer boxes))
-                    ]
-    in if null boxes then nullBox else result
+data CollectedWarnings = CollectedWarnings
```
paf31: Why not just `[PackageWarning]` out of curiosity?

hdgarrood: I wanted to collect all the warnings of each type up together, so that I can then create one `Box` for each type of warning, in order to make the messages easier to understand. This code makes sure that, for example, all of the `UndeclaredDependency` warnings are collected up and displayed together, giving a message like "the following are installed but not declared: [list]". It works similarly for the other two `PackageWarning` constructors.

PR #1183, opened 2015-06-19
Preserve exports list in externs files

garyb:
Well, this seems just fine.

@andyarvanitis Extending this to include a case for a module re-export `DeclarationRef` should be all that's needed on top of whatever you have to give externs all the info that's needed too.


paf31:
:+1: if it works :) I'm not too familiar with this area of the code any more, so I'll let you decide.

garyb:
I'll not merge this for now - Andy can pull it into his code and see if it gives us what we need.

PR #1184, opened 2015-06-20
Support for entire-module (re-)exports

andyarvanitis:
This doesn't have @garyb 's changes from PR #1183 yet, but I might not be able to work on it again until Sunday, so I'm sharing it here in case you guys need/want to have a look before then. In addition to making sure externs work properly, I'm sure it needs other work/cleanup too.

### DIFF
--- a/src/Language/PureScript/CodeGen/Externs.hs
+++ b/src/Language/PureScript/CodeGen/Externs.hs
@@ -38,7 +38,8 @@ import Language.PureScript.Types
 moduleToPs :: Module -> Environment -> String
 moduleToPs (Module _ _ _ Nothing) _ = error "Module exports were not elaborated in moduleToPs"
 moduleToPs (Module _ moduleName ds (Just exts)) env = intercalate "\n" . execWriter $ do
-  tell [ "module " ++ runModuleName moduleName ++ " (" ++ listExports exts ++ ") where"]
+  let exps = listExports exts
+  tell ["module " ++ runModuleName moduleName ++ (if null exps then "" else " (" ++ exps ++ ")") ++ " where"]
   mapM_ declToPs ds
   mapM_ exportToPs exts
   where
@@ -52,6 +53,7 @@ moduleToPs (Module _ moduleName ds (Just exts)) env = intercalate "\n" . execWri
     listExport (TypeRef name (Just dctors)) = Just $ show name ++ "(" ++ intercalate ", " (map show dctors) ++ ")"
     listExport (ValueRef name) = Just $ show name
     listExport (TypeClassRef name) = Just $ show name
+    listExport (ModuleRef name) = Just $ "module " ++ show name
     listExport _ = Nothing
 
     declToPs :: Declaration -> Writer [String] ()
@@ -123,6 +125,8 @@ moduleToPs (Module _ moduleName ds (Just exts)) env = intercalate "\n" . execWri
                               [] -> ""
                               cs -> "(" ++ intercalate ", " (map (\(pn, tys') -> show pn ++ " " ++ unwords (map prettyPrintTypeAtom tys')) cs) ++ ") => "
       tell ["foreign import instance " ++ show ident ++ " :: " ++ constraintsText ++ show className ++ " " ++ unwords (map prettyPrintTypeAtom tys)]
+
+    exportToPs (ModuleRef _) = return ()
 
     toTypeVar :: (String, Maybe Kind) -> Type
     toTypeVar (s, Nothing) = TypeVar s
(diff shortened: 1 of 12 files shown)


garyb:
Nice! Looks pretty good to me. The externs thing indeed doesn't work currently, but I'm pretty sure my changes in the other branch will fix that, also a module can't export itself - convenient if you want to export all of the members in a module without listing them explicitly, along with another module:

``` purescript
module A (module B, module A) where
-- ...tons of definitions that all want exporting...
```
I'll probably have a look at fixing both those things today, it'd be awesome to get this in.

andyarvanitis:
Great! I'll be on the road most of today, so feel free to change whatever you like. 



    _____________________________

michaelficarra:
I don't know, another way to import/export without explicitly listing names? I foresee headaches caused by this. Let's say I'm using your module `A` and want to go to the definition of `A.x`. I open up `A` but see that it is just re-exporting `B` and `C` wholesale. Now I need to look through each of those (and the modules they re-export, etc.) for `x`. @puffnfresh would tell me that I just need better tooling, but this is how people interact with programs right now, and I don't want to add to the language another feature that I have to tell people to avoid if they want their programs to be as accessible/maintainable as possible.

andyarvanitis:
@garyb I just added self exporting (with test case). Let me know what you think. Did you get a chance to merge and try out the externs change?

@michaelficarra Those are valid points -- we're following Haskell's example here, which is not always the "right" thing to do. But my feeling is that the convenience/readability afforded here outweighs the cost of having to trace a symbol's origin. Others with more Haskell experience than me would be better at saying how bad one is vs the other in practice, though.

andyarvanitis:
@garyb I think it's ready to go. Please give it a thorough review, if you don't mind, as I might have missed something critical w.r.t. extern files. Also, if you spot anything inefficient, or even just messy, please don't hesitate to let me know (or even change it at will, if you need to get this out before I'm back).

garyb:
Awesome, thanks. I'll take a look at it tonight.

@michaelficarra your comment definitely gave me pause, but I think the convenience outweighs the cost in this case. Most modules in SlamData have 30+ imports, it's extremely tedious managing them. Again, better tooling could be the answer there, but at least with the coming updates to pursuit hopefully the situation you describe above will be much improved.

puffnfresh:
Maybe PureScript plugins for [srclib](https://srclib.org/) and [SourceGraph](https://sourcegraph.com/) would help.

garyb:
:+1:

src/Language/PureScript/CodeGen/Externs.hs
```diff
@@ -38,11 +38,24 @@ import Language.PureScript.Types
 moduleToPs :: Module -> Environment -> String
 moduleToPs (Module _ _ _ Nothing) _ = error "Module exports were not elaborated in moduleToPs"
 moduleToPs (Module _ moduleName ds (Just exts)) env = intercalate "\n" . execWriter $ do
-  tell [ "module " ++ runModuleName moduleName ++ " where"]
+  let exps = listExports exts
+  tell ["module " ++ runModuleName moduleName ++ (if null exps then "" else " (" ++ exps ++ ")") ++ " where"]
```
garyb: I think this is my fault actually, but can we change this to `if null exps then " ()" ...`?

garyb:
Love it :+1: 

Could you just tweak that export thing and squash your commits? Thanks! (And sorry I only just got around to checking this out.)

src/Language/PureScript/CodeGen/Externs.hs
andyarvanitis: No it was me, and I meant to point it out and ask you about it, but completely forgot. Your version was:
```
tell [ "module " ++ runModuleName moduleName ++ " (" ++ listExports exts ++ ") where"]
```
Which resulted in the empty parens. However, that caused a parsing error (of the generated extern), so I thought it wasn't intended. So I guess it needs to be fixed on the parsing side?

garyb: Ah! In that case we can probably ignore it, if it's not possible to produce a module with `()` exports manually I don't think we'll ever make a module with `null exps` except in situations where the module is empty anyway.

I was thinking it might be handy for a module that only exports instances, but that shouldn't be a thing anyway, as it means the instances are all orphans if no types or classes are exported along with them.

andyarvanitis: Sounds good, I'll leave it as is then.

andyarvanitis:
Cool thanks, and no problem with the timing. I'll rebase in a bit.

andyarvanitis:
@garyb Rebased and ready to go, unless you spot anything else. I'll also add notes to the wiki (specifying it's a 0.7 feature) as soon as you merge it.

garyb:
Thank you!

PR #1187, opened 2015-06-21
Resolves #1169. Throw constructor arity error when case omits params

### LINKED ISSUE #1169: Produce better error when a case does not match all of a data constructor's arguments
Currently this results in a type error, but an explicit error message for not matching all of the arguments would be much clearer.

mjgpy3:
First shot.

Let me know if it looks okay :)

### DIFF
--- a/examples/failing/CaseDoesNotMatchAllConstructorArgs.purs
+++ b/examples/failing/CaseDoesNotMatchAllConstructorArgs.purs
@@ -1,0 +1,14 @@
+module Main where
+
+import Prelude
+
+data Person = Person String Int
+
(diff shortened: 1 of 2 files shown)


paf31:
Let me think about this one. I want to see if we can extract out a function so that we can avoid the duplication of `throwIncorrectArity`

mjgpy3:
Alright, thanks for looking!

mjgpy3:
@paf31, had any time to think about this?

paf31:
Sorry I'll have a look today.

paf31:
:+1: Thanks!

PR #1193, opened 2015-06-23
Small improvements to the pretty printer

wuzzeb:
- A do block had the first statement on the same line as the do block itself
  which caused later lines to not be indented properly.  Instead, add a newline
  after the do so that all statements in the do block are indented properly.

- The guards in a case expression were not indented at all, appearing on the
  very first character of the line.  Instead, use withIndent and prettyPrintMany
  to properly indent each guard of a case expression.


paf31:
:+1: Thanks!

PR #1194, opened 2015-06-24
Update core-imp to latest master (again)

andyarvanitis:
This required a fair bit of cleanup in the process of merging, so a review is definitely in order. I'm going to update pure14 with it next, which might uncover some issues.

Also, tests are running and passing.


garyb:
@andyarvanitis I'll close this if you're ok with that? When I revive CoreImp I'll probably just take the AST and start over with the other changes, it'll probably be easier than trying to bring it up to date again.

andyarvanitis:
@garyb Yeah, definitely!

PR #1195, opened 2015-06-25
Minor fixes to pretty printer

wuzzeb:
Here are a couple more improvements to the pretty printer.  There is one bug that I found but did not fix in this pull, since it is a more complicated bug.

If you have an expression like `App something (ObjectLiteral ...)` which is common for example when something is a data constructor, the indentation is not correct.  The bug is that in the `app` function in Pretty/Values.hs, it just prints the argument starting from zero indentation level.  My first guess is that changing app in Values.hs to the following...
```
app :: Pattern PrinterState Expr (String, Expr)
app = mkPattern' match
  where
  match (App val arg) = (,val) <$> prettyPrintValue' arg
  match _ = mzero
```

The above does prettyprint `App something (ObjectLiteral ...)` correctly, but I am not sure if it breaks anything else.

### DIFF
--- a/src/Language/PureScript/Pretty/Values.hs
+++ b/src/Language/PureScript/Pretty/Values.hs
@@ -124,7 +124,7 @@ prettyPrintDoNotationElement (DoNotationBind binder val) =
     ]
 prettyPrintDoNotationElement (DoNotationLet ds) =
   concat <$> sequence
-    [ return "let "
+    [ return "let\n"
     , withIndent $ prettyPrintMany prettyPrintDeclaration ds
     ]
 prettyPrintDoNotationElement (PositionedDoNotationElement _ _ el) = prettyPrintDoNotationElement el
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/Pretty/Values.hs
```diff
@@ -124,7 +124,7 @@ prettyPrintDoNotationElement (DoNotationBind binder val) =
     ]
 prettyPrintDoNotationElement (DoNotationLet ds) =
   concat <$> sequence
-    [ return "let "
+    [ return "let\n"
```
michaelficarra: So all `let`s are going to be followed by a newline? I think this style should be conditional on the length of `ds`.

vendethiel: :+1: for a single-line `let` if ds's length is 1

wuzzeb: So I don't see a good solution to pretty-printing the let without the newline.  Do you have an idea on how to do it?  The code as it is broken since the remainder of the printing of the let assumes a newline has been printed and an indentation has been added.

wuzzeb:
Ok, I added another commit.  After thinking about it, pretty printing the let with a single entry can just be a separate implementation.  I also fixed the pretty printing of arrays and included my change to the `app` function I mentioned above, since in more testing I didn't find anything that it broke.

wuzzeb:
So looking closer, it seems you have already made changes to the pretty printer, so this can probably be closed.  I'll run the new pretty printer through some of my tests and see if it works, and if not I will re-open a pull request.  But at the moment I am working on the yesod-purescript stuff :)

PR #1197, opened 2015-06-26
add support for building with stack

etrepum:
This `stack.yaml` file makes it quicker and easier to build and install purescript using [stack](https://github.com/commercialhaskell/stack). Fewer steps than using cabal, less redundant building of dependencies. Uses the latest GHC 7.8 lts stackage resolver.

Usage:

```
stack install # or just `stack build` to not also install it to ~/.local/bin/
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ cabal.sandbox.config
 .psci_modules/
 bower_components/
 tmp/
+.stack-work/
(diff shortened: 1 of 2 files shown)


paf31:
:+1: Thanks! Glad this works.

paf31:
It's a little bit of a shame that Stack can't read the versions for the extra dependencies from the cabal file, so we don't have to duplicate that data.

PR #1198, opened 2015-06-26
docs: Separate conversion and rendering stages

hdgarrood:
* Modify the data types in Language.PureScript.Docs.Types so that they
  no longer contain any RenderedCode values, but rather values of the
  types Language.PureScript.Type or Language.PureScript.Kind, in order
  to preserve more structure
* Rename Language.PureScript.Docs.Render to
  Language.PureScript.Docs.Convert, and remove all the code related to
  producing RenderedCode values
* Add a new module Language.PureScript.Docs.Render with code
  rendering functions (extracted from what was previously
  Language.PureScript.Docs.Render), to be used in markdown/html docs

These changes mean that information about the real types and kinds are stored in the JSON-serialized packages that pursuit works with, meaning that pursuit now has access to this information. This enables us to do the transformations necessary for Hoogle input files, and gives us more flexibility with regard to how we work around differences in PureScript's and Haskell's type systems (with respect to, eg, rows).

Also, as a bonus, the old renderDeclaration function (now split into the functions convertDeclaration, renderDeclaration, and renderChildDeclaration) is significantly simpler (resulting from concerns being better separated).

### DIFF
--- a/src/Language/PureScript/Docs/Convert.hs
+++ b/src/Language/PureScript/Docs/Convert.hs
@@ -1,0 +1,228 @@
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ViewPatterns #-}
+
+-- | Functions for converting PureScript ASTs into values of the data types
+-- from Language.PureScript.Docs.
+
+module Language.PureScript.Docs.Convert
+  ( convertModule
+  , collectBookmarks
+  ) where
+
+import Control.Monad
+import Control.Category ((>>>))
+import Data.Either
+import Data.Maybe (mapMaybe, isNothing)
+import Data.List (nub, isPrefixOf, isSuffixOf)
+
+import qualified Language.PureScript as P
+
+import Language.PureScript.Docs.Types
+
+-- |
+-- Convert a single Module.
+--
+convertModule :: P.Module -> Module
+convertModule m@(P.Module coms moduleName  _ _) =
+  Module (show moduleName) comments (declarations m)
+  where
+  comments = convertComments coms
+  declarations =
+    P.exportedDeclarations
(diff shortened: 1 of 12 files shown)


paf31:
Sounds good, I'll review this this afternoon. :+1:

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -0,0 +1,228 @@
...
+  = AugmentChild ChildDeclaration
+  | AugmentFixity P.Fixity
+
+-- | Augment top-level declarations; the second pass. See the comments under
+-- the type synonym IntermediateDeclaration for more information.
+augmentDeclarations :: [IntermediateDeclaration] -> [Declaration]
+augmentDeclarations (partitionEithers -> (augments, toplevels)) =
+  foldl go toplevels augments
+  where
+  go ds (parentTitles, a) =
```
paf31: Maybe it's worth using a `Map` here? Later perhaps.

paf31:
Looks good. Presumably this doesn't change the output of `psc-docs`?

I'll merge this when the CI build finishes. :+1:

hdgarrood:
I just had a look and there is a small change to the output of psc-docs, and it's a result of a little bug I've introduced here: https://github.com/purescript/purescript/pull/1198/files#diff-d65b568ab4d89646257197a39e8fc696R48. It's checking if there are any child declarations, rather than what I intended, which was to check if there are any members in the type class (this is a stronger condition).

Rather than fixing it, though, I'm tempted to drop the `where`s entirely; I don't think they add anything in markdown or HTML docs, where the structure is already clear from the markup.

hdgarrood:
To clarify: that bug is that type classes with no members now have a `where` keyword on the end, whereas previously, type classes would only have a `where` if they had one or more members.

paf31:
That's fine with me. Is this good to merge then?

hdgarrood:
At the moment, yes. However, I would additionally like to drop the `where` keywords from type classes - if that's ok with you, I'll make another commit to do that.

paf31:
:+1: Sounds good.

PR #1199, opened 2015-06-27
docs: Fix display of 'where' keyword after typeclasses

hdgarrood:
A 'where' should only be displayed in documentation for a type class declaration if the type class has at least one member. This commit fixes a bug introduced in #1198 where the 'where' keyword was sometimes displayed even if a type class has no members.

In #1198 I was advocating getting rid of 'where' entirely, but it turns out that doesn't work well for the output of psc-docs. Instead, I can make that change inside Pursuit rather than here.


paf31:
:+1:

PR #1200, opened 2015-06-27
Remove psc, DCE module, simplify options.

paf31:
@garyb @hdgarrood @andyarvanitis Could you please review this?

In case you're not aware, the plan is to remove `psc` with [`psc-bundle`](https://github.com/paf31/psc-bundle), which will turn the output of `psc-make` into a browser bundle.


paf31:
Still to do: get tests working again.

puffnfresh:
Good plan :+1:

michaelficarra:
Will psc-bundle be distributed with a PureScript install? If so, you can just rename it psc. If not, we'll have the same problem that we have with the compiler: I need to specify a version to list in our dependencies, which are managed through npm and package.json or bower and bower.json.

andyarvanitis:
Looks ok to me. I've been mainly working from the `core-imp` branch, and these changes look fairly compatible with the changes there.

garyb:
@michaelficarra `psc-bundle` doesn't actually run the compiler, so renaming it `psc` would be a bit misleading - we'd talked about renaming `psc-make` to `psc` actually. I think you're right though, distributing it as part of the install seems like a good idea.

@paf31 :+1:

psc-make/Main.hs
```diff
@@ -53,6 +53,9 @@ data InputOptions = InputOptions
 compile :: PSCMakeOptions -> IO ()
 compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = do
   input <- concat <$> mapM glob inputGlob
+  when (null input) $ do
+    putStrLn "psc-make: No input files."
```
hdgarrood: What do you think about sending this to standard error?

hdgarrood:
Looks good to me :+1:

psc-make/Main.hs
paf31: Sounds good.

paf31:
@michaelficarra I'm fine moving it into the compiler repo, maybe after it's been tested a bit more. 

As for versioning, I'd also be interested to see if we could standardize the format we generate, even loosely, so that we just target that standard, and the bundler tool could be used with other front-ends.

paf31:
Ok, the tests are back. Should I merge this then?

garyb:
Are you going to change that "No input files" to use `stderr` first? Either way :shipit:

paf31:
Yes, good point, thanks.

PR #1202, opened 2015-06-28
Remove backticks from error messages.

mjgpy3:
Minor consistency thing.

I noticed some of the errors surround identifiers and other tokens in backticks (this was partially my doing).

I removed backticks, however, if they're more preferable I will generate an inverse "backtick all the code" PR.

Thoughts?

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -438,9 +438,9 @@ prettyPrintSingleError full e = prettyPrintErrorMessage <$> onTypesInErrorMessag
     goSimple (UnknownDataConstructor dc tc) =
       line $ "Unknown data constructor " ++ show dc ++ foldMap ((" for type constructor " ++) . show) tc
     goSimple (ConflictingImport nm mn) =
-      line $ "Cannot declare `" ++ nm ++ "` since another declaration of that name was imported from `" ++ show mn ++ "`"
+      line $ "Cannot declare " ++ show nm ++ " since another declaration of that name was imported from " ++ show mn
     goSimple (ConflictingImports nm m1 m2) =
-      line $ "Conflicting imports for " ++ nm ++ " from modules " ++ show m1 ++ " and " ++ show m2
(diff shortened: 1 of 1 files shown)


michaelficarra:
Without an indicator, error messages could be confusing. By the way, the maintainers are going to yell at you for not opening an issue first.

paf31:
If it's just a string, why not use `show`?

mjgpy3:
Could you explain further?

paf31:
Why not

```haskell
"Cannot declare " ++ show nm ++ ... " since another ..."
```

garyb:
The point being show will render a string `foo` as `"foo"`, so you get consistent quoting "for free".

mjgpy3:
Oh cool, I get it -- I forgot about that. Thanks for the suggestion, I'll get to that

mjgpy3:
Latest push uses `show`. As far as consistency is concerned I still think a lot of types are non-string. Shall I quote them? Or perhaps create a new typeclass a la `ShowCode`?

paf31:
I'm not sure I understand. You only used `show` for `String`s right? I don't want to introduce an extra `Show` class, I just don't want the `\`` characters.

paf31:
If I understand correctly, then no, they shouldn't be quoted. Their pretty-printers already handle quotes if necessary.

mjgpy3:
Gotcha. Current push looks good, I'm sorry for my confusion.

Here's what this push is looking like:
> Error:
> Error in module Foobar:
> Error at /home/michael/dev/contrib/purescript/../Foobar.purs line 3, column 1 - line 5, column 1:
>   Cannot declare id since another declaration of that name was imported from Prelude
> Possible fix: hide id when importing Prelude:
>   import Prelude hiding (id)
> See https://github.com/purescript/purescript/wiki/Error-Code-ConflictingImport for more information, or to contribute content related to this error.

paf31:
Thanks!

mjgpy3:
Thank you!

PR #1207, opened 2015-06-29
Add module names for bundle failures

khanage:
I needed this when I was using the tool - might help others as well.

I'm happy to take suggestions on a better way to thread this through, but this works for now.


paf31:
Could we use `ErrorInFile`?

hdgarrood:
I guess this fixes #1233

paf31:
Fixed by #1307.

PR #1208, opened 2015-06-29
Support exports using square brackets in FFI modules under psc-bundle

paf31:

This came up when testing `purescript-parsing` with Pulp. Exports of the form `exports["primed'"] = ...` weren't being identified as module exports.


paf31:
In other news, I'm really starting to like `PatternGuards` :+1:

garyb:
:+1:

PR #1218, opened 2015-07-01
Exhaustivity Checker

nicodelpiano:
It covers exhaustivity for data constructors, objects, guards and boolean literals.
Basically, we analyse the clauses of a definition one by one from top to bottom, where in each step we have the cases already missing (uncovered), and we generate the new set of missing cases.
Starting with `uncovered = { _ }` (nothing covered), we partition that set with the new uncovered cases.

Example:
```haskell
module Main where

data Nat = Zero | Succ Nat

f Zero = Zero
```
This returns
```
Pattern Match(es) are not exhaustive
The definition of "" has the following uncovered cases:
  Main.Succ _
```
Records:
```haskell
g {bar : Zero} = Zero
g {foo: Zero, bar : _} = Zero
```
we obtain:
```
Pattern Match(es) are not exhaustive
The definition of "" has the following uncovered cases:
  {
  bar: Main.Succ _
  foo: Main.Succ _}
```
For guards, we simply say that are not exhaustive those clauses who hasn't an `otherwise` guard.

TODO:
- Remove redundant and repeated cases (this is already done, but it makes the main algorithm's complexity nearly `O(n^2)`) 
- Add function names to warnings and improve the pretty-printing
- Add tests and documentation

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -19,7 +19,7 @@
 module Language.PureScript.Errors where
 
 import Data.Either (lefts, rights)
-import Data.List (intercalate)
+import Data.List (intercalate, transpose)
 import Data.Function (on)
 import Data.Monoid
 import Data.Foldable (fold, foldMap)
@@ -120,6 +120,7 @@ data SimpleErrorMessage
   | TransitiveExportError DeclarationRef [DeclarationRef]
   | ShadowedName Ident
   | WildcardInferredType Type
+  | NotExhaustivePattern [[Binder]]
   | ClassOperator ProperName Ident
   deriving (Show)
 
@@ -229,6 +230,7 @@ errorCode em = case unwrapErrorMessage em of
   (TransitiveExportError _ _)   -> "TransitiveExportError"
   (ShadowedName _)              -> "ShadowedName"
   (WildcardInferredType _)      -> "WildcardInferredType"
+  (NotExhaustivePattern _)      -> "NotExhaustivePattern"
   (ClassOperator _ _)           -> "ClassOperator"
 
 -- |
@@ -559,6 +561,11 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage <$> onTypesInError
             ]
     goSimple (WildcardInferredType ty) =
       line $ "The wildcard type definition has the inferred type " ++ prettyPrintType ty
+    goSimple (NotExhaustivePattern bs) =
+      paras $ [ line "Pattern could not be determined to cover all cases."
+              , line $ "The definition has the following uncovered cases:\n"
+              , indent $ Box.hsep 1 Box.left (map (paras . map (line . prettyPrintBinderAtom)) (transpose bs))
+              ]
     go (NotYetDefined names err) =
       paras [ line $ "The following are not yet defined here: " ++ intercalate ", " (map show names) ++ ":"
             , indent $ go err
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -1,0 +1,275 @@
+-----------------------------------------------------------------------------
+--
+-- Module      :  Language.PureScript.Exhaustive
+-- Copyright   :  (c) 2013-14 Phil Freeman, (c) 2014 Gary Burgess, and other contributors
+-- License     :  MIT
+--
+-- Maintainer  :  Phil Freeman <paf31@cantab.net>
+-- Stability   :  experimental
+-- Portability :
+--
+-- |
+-- | Module for exhaustivity checking over pattern matching definitions
+-- | The algorithm analyses the clauses of a definition one by one from top
+-- | to bottom, where in each step it has the cases already missing (uncovered),
+-- | and it generates the new set of missing cases.
+-- 
+-----------------------------------------------------------------------------
+
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Language.PureScript.Linter.Exhaustive 
+  ( checkExhaustive
+  , checkExhaustiveModule
+  ) where
+
+import qualified Data.Map as M
+import Data.Maybe (fromMaybe)
+import Data.List (foldl', sortBy, nub)
+import Data.Function (on)
+
+import Control.Monad (unless)
+import Control.Applicative
+import Control.Monad.Writer.Class
+
+import Language.PureScript.AST.Binders
+import Language.PureScript.AST.Declarations
+import Language.PureScript.Environment
+import Language.PureScript.Names as P
+import Language.PureScript.Kinds
+import Language.PureScript.Types as P
+import Language.PureScript.Errors
+
+import Language.PureScript.AST.Traversals (everywhereOnValuesTopDownM)
+
+-- |
+-- Qualifies a propername from a given qualified propername and a default module name
+--
+qualifyName :: a -> ModuleName -> Qualified a -> Qualified a
+qualifyName n defmn qn = Qualified (Just mn) n
+  where
+  (mn, _) = qualify defmn qn
+
+-- |
+-- Given an environment and a datatype or newtype name,
+-- this function returns the associated data constructors if it is the case of a datatype
+-- where: - ProperName is the name of the constructor (for example, "Nothing" in Maybe)
+--        - [Type] is the list of arguments, if it has (for example, "Just" has [TypeVar "a"])
+--
+getConstructors :: Environment -> ModuleName -> (Qualified ProperName) -> [(ProperName, [Type])]
+getConstructors env defmn n = extractConstructors lnte
+  where
+  qpn :: Qualified ProperName
+  qpn = getConsDataName n
+
+  getConsDataName :: (Qualified ProperName) -> (Qualified ProperName)
+  getConsDataName con = qualifyName nm defmn con
+    where
+    nm = case getConsInfo con of
+           Nothing -> error $ "ProperName " ++ show con ++ " not in the scope of the current environment in getConsDataName."
+           Just (_, pm, _, _) -> pm
+
+  getConsInfo :: (Qualified ProperName) -> Maybe (DataDeclType, ProperName, Type, [Ident])
+  getConsInfo con = M.lookup con dce
+    where
+    dce :: M.Map (Qualified ProperName) (DataDeclType, ProperName, Type, [Ident])
+    dce = dataConstructors env
+
+  lnte :: Maybe (Kind, TypeKind)
+  lnte = M.lookup qpn (types env)
+
+  extractConstructors :: Maybe (Kind, TypeKind) -> [(ProperName, [Type])]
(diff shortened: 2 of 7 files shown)


hdgarrood:
:heart_eyes: :heart_eyes: :heart_eyes:

src/Language/PureScript.hs
```diff
@@ -148,7 +148,8 @@ make MakeActions{..} ms = do
     go env' ms'
   go env ((True, m@(Module coms moduleName' _ exps)) : ms') = do
     lift $ progress $ "Compiling " ++ runModuleName moduleName'
-    (Module _ _ elaborated _, env') <- lift . runCheck' env $ typeCheckModule Nothing m
+    (modu@(Module _ _ elaborated _), env') <- lift . runCheck' env $ typeCheckModule Nothing m
```
paf31: Maybe `checked` instead of `modu`?

src/Language/PureScript/Pretty/Values.hs
```diff
@@ -194,30 +194,33 @@ prettyPrintBinderAtom :: Pattern PrinterState Binder String
...
-  match (StringBinder str) = return $ show str
-  match (CharBinder c) = return $ show c
-  match (NumberBinder num) = return $ either show show num
-  match (BooleanBinder True) = return "true"
-  match (BooleanBinder False) = return "false"
-  match (VarBinder ident) = return $ show ident
-  match (ConstructorBinder ctor args) = concat <$> sequence
+  match = match' 0
+
+  match' :: Int -> Binder -> StateT PrinterState Maybe String
```
paf31: Is this extra `Int` still necessary?

paf31:
Looks nice. I'll do a full review later today or tomorrow. :smiley:

codygman:
:+1: :)

garyb:
Really looking forward to having this!

src/Language/PureScript/Errors.hs
```diff
@@ -556,6 +558,16 @@ prettyPrintSingleError full e = prettyPrintErrorMessage <$> onTypesInErrorMessag
             ]
     goSimple (WildcardInferredType ty) =
       line $ "The wildcard type definition has the inferred type " ++ prettyPrintType ty
+    goSimple (NotExhaustivePattern bs) =
+      paras $ [ line "Pattern Match(es) are not exhaustive"
```
paf31: Purely cosmetic: maybe `Pattern match is not exhaustive`? I tend to think of the whole `case` as a singular "pattern match" (although I might be the only one). In any case, I think it makes the error slightly easier to read.

Also, I think a line break might be nice between the second line and the list of cases.

paf31: Actually, what about `Patterns could not be determined to cover all cases`? That way, it's slightly clearer that they _might_ be exhaustive, but that we couldn't prove it.

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -0,0 +1,228 @@
...
+-- the first argument is the list of uncovered binders
+-- the second argument is a clause of a pattern matching def
+--
+missingCasesMultiple :: Environment -> [Binder] -> [Binder] -> [[Binder]]
+missingCasesMultiple env = go
+  where
+  go [] _ = []
+  go (x:xs) (y:ys)
+    | null miss = map (x :) (go xs ys)
+    | otherwise = map (: xs) miss ++ map (y :) (go xs ys)
```
paf31: Let's document this line, since I was intially confused why `(y :)` was adequate :)

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -0,0 +1,228 @@
+-----------------------------------------------------------------------------
+--
+-- Module      :  Language.PureScript.Exhaustive
+-- Copyright   :  (c) 2013-14 Phil Freeman, (c) 2014 Gary Burgess, and other contributors
+-- License     :  MIT
+--
+-- Maintainer  :  Phil Freeman <paf31@cantab.net>
+-- Stability   :  experimental
+-- Portability :
+--
```
paf31: Could you please add a line or two for the Haddock docs?

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -0,0 +1,228 @@
...
+-- |
+-- Replicates a wildcard binder
+--
+initialize :: Int -> [Binder]
+initialize l = replicate l NullBinder
+
+-- |
+--
+--
+completeMissing :: Ord a =>
```
paf31: Maybe this could be called `genericMerge` or something like that? Also, could you please comment it?

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -0,0 +1,228 @@
...
+--     | otherwise = 1
+-- Is exhaustive, whereas `f x | x < 0` is not
+--
+-- The function below say whether or not a guard has an otherwise expression
+--
+isExhaustiveGuard :: Either [(Guard, Expr)] Expr -> Bool
+isExhaustiveGuard (Left gs) = not . null $ filter (\(g, _) -> isOtherwise g) gs
+  where
+  isOtherwise :: Expr -> Bool
+  isOtherwise (TypedValue _ (BooleanLiteral True) _) = True
```
paf31: Does `otherwise` get inlined to `true`? If not, I think we need to look for the value `otherwise` explicitly. If it does, let's document it.

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -0,0 +1,228 @@
...
+  step :: [[Binder]] -> CaseAlternative -> [[Binder]]
+  step uncovered ca = concatMap (missingAlternative env ca) uncovered
+
+  initial :: [Binder]
+  initial = initialize numArgs
+    where
+    numArgs = length . caseAlternativeBinders . head $ cas 
+
+  makeResult :: [[Binder]] -> m ()
+  makeResult bss | null bss = return ()
```
paf31: You could use `unless` here.

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -0,0 +1,228 @@
...
+    | null miss = map (x :) (go xs ys)
+    | otherwise = map (: xs) miss ++ map (y :) (go xs ys)
+    where
+    miss = missingCasesSingle env x y
+  go _ _ = error "Argument lengths did not match in missingCasesMultiple."
+
+-- |
+-- Is the first list of binders covered by the second?
+--
+isExhaustive :: Environment -> [Binder] -> [Binder] -> Bool
```
paf31: I think these two functions can be removed.

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -0,0 +1,228 @@
...
+  | s > s' = (f s' Nothing (Just b')) : completeMissing f bsl bs'
+  | otherwise = (f s (Just b) (Just b')) : completeMissing f bs bs'
+
+-- |
+-- Find the uncovered set between two binders:
+-- the first binder is the case we are trying to cover
+-- the second one is the matching binder
+--
+missingCasesSingle :: Environment -> Binder -> Binder -> [Binder]
+missingCasesSingle _ _ NullBinder = []
```
paf31: Don't forget `NamedBinder` :)

paf31:
@garyb Everything look ok to you? @nicodelpiano has a couple more minor commits (mostly around how warnings are generated) and then I was planning to merge this.

garyb:
Yep, looks great to me :+1:

paf31:
@nicodelpiano Looks like you added all the changes we discussed. Is this good to go?

Thanks! :+1:

nicodelpiano:
@paf31 Yes Phil! I made all the changes. Good to go :+1: :)

paf31:
From IRC:

- [ ] Check out the Travis log. There are apparently issues with type class instances.
- [ ] I found this case reports the wrong binders, but reversing the lines works, strangely:

    ```purescript
    x true true = false
    x false false = true
    ```

nicodelpiano:
@paf31 
- About `x` and `y` thing:
```haskell
f true true = false
f false false = true
```

Using `x` just works because of luck:

```haskell
module Main where
f true true = true
```
generates: `{ false _, _ false }`

If we add the case:

```haskell
f false false = false
```

Then, checking `false _` against `true true`, leaves us with the new uncovered case: `false true`.
Then, checking `_ false` against `true true`, leaves us with `true false`.
So we are ok: `{false true, true false}`

What happens if we have three constructors:
```haskell
data B = True | False | Idk
f True True = True
f False False = False
```
It throws:
```haskell
Main.False Main.True
Main.False Main.Idk
Main.Idk _
Main.Idk Main.True
Main.Idk Main.Idk
Main.True Main.False
Main.Idk Main.False
Main.True Main.Idk
Main.Idk Main.Idk
_ Main.Idk
```
a lot of redundant cases! 

Using `y` happens something similar:

```haskell
Main.False Main.True
Main.False Main.Idk
Main.Idk _
Main.False Main.True
Main.False Main.Idk
Main.True Main.False
Main.True Main.Idk
Main.False Main.Idk
```

The big difference is this, let's consider the following program:

```haskell
module Main where
data N = Z | S N
f Z Z = Z
```
uncovered cases | Using `x` | Using `y`
------------- | ------------ | -------------
1 | `(S _) _` | `(S _) _`
2 | `_ (S _)` | `Z (S _)`

Using `y` we don't generate overlapping cases, since we can't instantiate the `2` (`1`) uncovered case to be one which overlaps the `1` (`2`), whereas with `x`, we can instantiate `2` to be `(S Z) (S Z)`, and this is already covered by `1`.
I just tried to follow what `GHC` does.

Clearly, this shouldn't be merged as it doesn't work properly.

What I suggested in my first version to solve this is to filtering in each step (we need to check if we are generating correct uncovered cases), but I guess the complexity blows a bit.

I'll make those modifications just to see what happens with the tests.

Thanks for the reviews!

nicodelpiano:
Log: https://s3.amazonaws.com/archive.travis-ci.org/jobs/70315916/log.txt

paf31:
I'm not worried so much about the errors containing overlapping cases - we can solve that by case reduction later if we want to. What's more important is that every case shown is actually uncovered. Looks like the problem was a different bug, so I'm fine keeping `y`, but we probably should clearly document why `y` is ok. 

Also, by adding the extra `isExhaustive` check at each step, I worry that the complexity of your approach will blow up. What is the check actually checking? If it's filtering out only those cases which cover new things, then I would hope that could be fused into the `uncovered` function itself.

nicodelpiano:
@paf31 Yes, I agree, I guess redundancy checking will help us to do this more efficiently. I also think the filtering stuff is necessary to maintain the invariant (without redundancy), since the filtering stuff does what you said: it checks if the new uncovered cases generated by the mapping are actually the right cases.

paf31:
Merged, thanks! :100:

garyb:
:tada:

nicodelpiano:
:+1: thanks! :smiley:

PR #1221, opened 2015-07-02
Fix transitive dependencies not being included

hdgarrood:
There was a bug where resolvedDependencies was only including top level
dependencies and not transitive ones. This commit fixes it.

'top level dependencies' here means those which appear at the top level
of the 'dependencies' key of `bower list --json`, as a result of either
having being declared in bower.json or having been installed manually
using `bower install`.

### DIFF
--- a/psc-publish/Main.hs
+++ b/psc-publish/Main.hs
@@ -8,10 +8,11 @@ import Prelude hiding (userError)
 import Data.Maybe
 import Data.Char (isSpace)
 import Data.String (fromString)
-import Data.List (stripPrefix, isSuffixOf, (\\))
+import Data.List (stripPrefix, isSuffixOf, (\\), nubBy)
 import Data.List.Split (splitOn)
 import Data.List.NonEmpty (NonEmpty(..))
 import Data.Version
+import Data.Function (on)
 import Safe (headMay)
 
 import qualified Data.ByteString.Lazy.Char8 as BL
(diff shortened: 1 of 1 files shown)


hdgarrood:
This only concerns psc-publish, and fixes missing links in Pursuit docs, like Maybe not being linked [here](http://new-pursuit.purescript.org/packages/purescript-arrays/0.4.0/docs/Data.Array#d:head)

psc-publish/Main.hs
```diff
@@ -196,18 +197,35 @@ getResolvedDependencies :: [PackageName] -> PrepareM [(PackageName, Version)]
...
+--   `bower list --json --offline`
+asResolvedDependencies :: Parse BowerError [(PackageName, DependencyStatus)]
+asResolvedDependencies = nubBy ((==) `on` fst) <$> go
+  where
+  go =
+    fmap (fromMaybe []) $
+      keyMay "dependencies" $
+        (++) <$> eachInObjectWithKey (parsePackageName . T.unpack)
+                                     asDependencyStatus
+             <*> (concatMap snd <$> eachInObject asResolvedDependencies)
```
paf31: If there is a cyclic dependency (unlikely but possible, I think), won't this `<<loop>>`?

hdgarrood: This is parsing the output of 'bower list --json --offline' so I think that would have to produce an infinite amount of output for this to cause an infinite loop. In such a case we probably wouldn't even get as far as this line; I think readProcess would never return. I will check this though.

hdgarrood: I just checked this - if you create a bower package with cyclic dependencies and then run `bower list --json --offline`,  it prints a dependency tree where packages are not listed where they would appear as children of themselves. So if A depends on B depends on A, the dependency tree will just look like `A -> B`. So I am pretty sure this is fine.

paf31:
:+1: Thanks!

PR #1226, opened 2015-07-03
Simplify pretty printer for binders

paf31:
This should make the errors in #1218 a bit nicer.

@garyb @nicodelpiano Could you please review?

### DIFF
--- a/src/Language/PureScript/Pretty/Values.hs
+++ b/src/Language/PureScript/Pretty/Values.hs
@@ -95,7 +95,7 @@ prettyPrintDeclaration _ = error "Invalid argument to prettyPrintDeclaration"
 prettyPrintCaseAlternative :: CaseAlternative -> StateT PrinterState Maybe String
 prettyPrintCaseAlternative (CaseAlternative binders result) =
   concat <$> sequence
-    [ intercalate ", " <$> forM binders prettyPrintBinder'
+    [ return (unwords (map prettyPrintBinderAtom binders))
     , prettyPrintResult result
     ]
   where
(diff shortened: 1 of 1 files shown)


garyb:
Looks good from trying to read the code, but I probably can't tell without seeing the differences in output :smile:

paf31:
Sure:

```
> let f s = (prettyPrintBinder <$>) <$> runTokenParser "" parseBinder <$> P.lex "" s

> f "A b c"
Right (Right "A b c")

> f "A b (C d e)"
Right (Right "A b (C d e)")

> f "A (B c) (D e)"
Right (Right "A (B c) (D e)")

> f "A [ B c, D e ]"
Right (Right "A [ B c, D e ]")
```

There is probably a useful QuickCheck test here.

paf31:
I'd actually be in favour of replacing more simple uses of `pattern-arrows` with simple functions like this. The one place where `pattern-arrows` is more useful IMO is when you have a dynamic set of precedence rules (e.g. if we pretty printed inline operators, which we don't). As it is, almost every one of our pretty printers could be simplified by not using it.

nicodelpiano:
I tested it, and it looks good!

```
module Main where
data Tree a = Empty | Leaf a | Branch (Tree a) (Tree a)
f (Leaf 0) Empty = Empty
```

```haskell
Main.Empty
(Main.Leaf _)
(Main.Branch _ _) _
(Main.Leaf 0) (Main.Leaf _)
(Main.Leaf 0) (Main.Branch _ _)
```

That's what I expected to see. Now the general pretty print call should be `prettyPrintBinderAtom`, right? I suspect we should export it from the module.

paf31:
Yes, you need `prettyPrintBinderAtom`. Maybe when you rebase you can export it?

Issue #1121, opened 2015-05-22
Please document supported OS/ghc versions clearly


mcclure:
Hello. Last night I spent some time unsuccessfully attempting to install Purescript. My machine is running fairly out of date software: Mac OS X 10.6.8 on 64-bit Intel, and I have a copy of GHC 7.6.3 (w/ Cabal 1.16.0.2) installed since I don't even remember when. Running the OS X binaries from your github "release page" failed (running any one simply resulted in the message "illegal instruction") and cloning git and building master (by running `cabal install` in directory) failed with a cryptic message ( https://pbs.twimg.com/media/CFlh4K_VEAA8coU.png:large ).

I understand you cannot necessarily support old operating system versions or old GHC versions; my point is not to ask that be changed, and I am not asking for help installing (a friend with a newer GHC helped and I now have purescript running). However, I think your messaging about which software versions are supported can and should be improved. Doing so would have saved me a bunch of time. It seems labeling "supported" operating system version for binaries is especially important on Linux and OS X (where Apple routinely introduces backward-incompatible changes to the dev tools, although I do not know how sensitive Haskell programs are to this).

Here are some places a brief message about supported versions could have been, but aren't:

- The "release" page on Github says nothing https://github.com/purescript/purescript/releases/tag/v0.6.9.5 (this is just a commit message, but since it is linked from "latest release page on GitHub" on your website it would be reasonable to expect it to contain this information)
- The README in macos.tar.gz says nothing
- Nor the README in linux64.tar.gz
- The README.md in git does not mention it either in master (which effectively is your "front page" for anyone who happens to stumble onto your github project instead of going to purescript.org first) or the v0.6.9.5 source release, incidentally it is not in the release notes either
- The build could have detected early that it was in an unsupported GHC version, but didn't (note: I *did* build master, maybe a build from v0.6.9.5 or directly from cabal is better-behaved)
- http://www.purescript.org/download/ does not mention this information-- looking now, oops, I *do* see that the "Cabal" install section says it wants "Haskell Platform 2014 or GHC 7.8"!, although this is maybe a vague statement (does it refer to building the source or installing via cabal?).

[...]

PR #1219, opened 2015-07-02
Use sndM instead of traverse, fix #1211

paf31:
@garyb @hdgarrood Look ok?

PR #1230, opened 2015-07-05
Require base >= 4.7

hdgarrood:
GHC 7.6 and earlier is not supported. Any attempt to build purescript
with GHC 7.6 or earlier will now fail at the `cabal configure` step,
instead of giving a compile error which is impossible to understand for
people not familiar with the compiler internals.

Refs #1121

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -26,7 +26,7 @@ source-repository head
     location: https://github.com/purescript/purescript.git
 
 library
-    build-depends: base >=4.6 && <5,
+    build-depends: base >=4.7 && <5,
                    containers -any,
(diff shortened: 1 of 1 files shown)


hdgarrood:
Oh wait, I saw the Travis build on https://github.com/purescript/purescript/pull/1219 fail and assumed that GHC 7.6 wasn't supported for some other reason. I just checked, though, and I was able to build purescript from master with GHC 7.6. We can add GHC 7.6 to the Travis build but we just need to make sure we have a sufficiently recent version of cabal-install (or disable coverage for 7.6).

hdgarrood:
See #1231 instead.

PR #1231, opened 2015-07-05
Run Travis build on GHC 7.6

hdgarrood:
But, only run coverage on GHC 7.8, so as not to confuse Coveralls


paf31:
Nice. Hopefully the Travis build doesn't time out :)

PR #1236, opened 2015-07-06
Refactor parseAndDesugar

hdgarrood:
In preparation for writing a slightly different version, which does not
do importPrim. This is in order to be able to produce JSON files using
psc-publish for purescript-prim; we need to be able to tell psc-publish
not to add a Prim import, so that we can avoid RedefinedModule errors.


hdgarrood:
Oh, wait, I guess skipping `map importPrim` won't do it. Hmm. I think this is still worth doing, though, since the parseAndDesugar function is _very_ hard to follow (as this has just reminded me).

hdgarrood:
I tested by running with purescript-quickcheck before and after, and the results were identical.

paf31:
:+1:

PR #1237, opened 2015-07-06
Handle 'bower link' properly in psc-publish

hdgarrood:
Previously, if you added a bower dependency using 'bower link',
psc-publish would fail with an internal error about a missing key
"_resolution" while trying to ascertain the DependencyStatus for a
package.

Now, that case is handled gracefully. Any such packages are listed in
the existing warning for packages which have been specified by something
other than a version tag (eg, a branch or commit). The warning message
wording and relevant data constructors have been amended slightly to
account for this.


paf31:
:+1: Looks nice, thanks!

PR #1238, opened 2015-07-07
Add the compiler version to psc-publish output

hdgarrood:
For pursuit: to determine whether a given JSON-serialized package was
produced by a sufficiently recent compiler, so that it can fail
gracefully if necessary.

We also modify the parsers to accept an extra argument which is the
minimum allowable version. The FromJSON instance sets the minimum
allowable version to 0.0.0.0 (that is, it never fails with a
CompilerTooOld error).

### DIFF
--- a/psc-publish/Main.hs
+++ b/psc-publish/Main.hs
@@ -38,6 +38,7 @@ import Web.Bower.PackageMeta (PackageMeta(..), BowerError(..), PackageName,
                               runPackageName, parsePackageName, Repository(..))
 import qualified Web.Bower.PackageMeta as Bower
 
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

paf31:
:+1: Thanks!

AppShipIt:
Y'all are killing it with awesome PRs this weekend 👯

garyb:
Oops, think I meant to link to #1283 actually :wink:

Thanks though! Everything has turned out to be easier than I expected so far, aside from type operators. The typed holes thing has been something I've wanted more and more lately, along with the quick chat about it at the meetup yesterday reminding me to get it done.

Issue #1229, opened 2015-07-03
Move some make and bundle code into the library


wuzzeb:
So I'm working on updating [yesod-purescript](https://github.com/mpietrzak/yesod-purescript) to version 0.7, and it would help if some of the code that exists in `psc` and `psc-bundle` were moved into the library.  If you don't want to do this, we can change yesod-purescript to execute the `psc` and `psc-bundle` binaries instead of using the library directly.  The only downside to this is then yesod-purescript needs options to be able to find these executables, so needs options for the path to the cabal or stack sandbox (since yesod-purescript depends on the purescript package, it will be installed into the sandbox).  Also, reporting errors is harder since we need to examine the output of the binaries, instead of more controlled error reporting we get from using the purescript library directly.

What I would suggest is the following:

* move the contents of `psc/Make.hs` and the make stuff from `src/Language/PureScript.hs` into a new module at `src/Language/PureScript/Make.hs`.  This might even be a beneficial cleanup since it moves the make stuff out of the root of the library hierarchy at Language.PureScript.  We don't need anything from `psc/Main.hs`.

* move `withDeps`, `compile`, and `codeGen` from `psc-bundle/Main.hs` to a new `src/Language/PureScript/Bundle.hs` module and create a wrapper function `bundle` which calls these three functions similar to how `app` does now.  I would only export this new `bundle` function, not the individual steps in the bundle.

What do you think?

paf31:
I'm in favor of moving some code over, since we can probably share code between `psci`, `psc` and the test suite. I'll also need it when I update the Try PureScript site.

wuzzeb:
So to help the discussion, I made the following commit which moves the make code so you can see what it would look like.  I haven't done any testing but it does compile, and I didn't change any of the code, just moved it.

  https://github.com/wuzzeb/purescript/commit/9ef45690ab1ca46c51da9a492092329dc06e4f2c

wuzzeb:
This is now fixed with pull request #1245 being merged.

PR #1245, opened 2015-07-09
Move code from executables into the library

wuzzeb:
This should implement #1229.  There is very little new code here, it is primarily moving code and updating a few of the comments.  To recap, it moves code from psc/Make.hs and psc-bundle/Main.hs into the main library.


paf31:
Looks good to me. This will make it much easier to convert Try PureScript to 0.7 as well :)

Is it possible to extract some of the similar `Make` code from `psci` and `tests` now?

wuzzeb:
For psci/Make.hs, it looks like it was originally copied but there are a few differences.

* src/Language/PureScript/Make.hs version of getInputTimestamp uses the maximum time between the code and the foreign file.  psci/Make.hs just uses the code.  I suspect this is a bug in psci, since it doesn't check the timestamp of foreign files.

* src/Language/PureScript/Make.hs version of writeFile displays a message to the screen for the
  file that it wrote, psci/Make.hs does not.  Not sure how critical this is, could probably be removed to make the two versions the same. 

*  psci/Make.hs omits some progress messages for modules starting with $PSCI.   We could still do this by changing out the progress function inside psci/Make.hs.  That is, use the MakeActions from src/Language/PureScript/Make.hs but replace just the progress function.

For the make stuff in tests, it looks like it is different make actions so no sharing needed.

paf31:
If it's possible to extract out the common stuff, and use a record update inside `psci`, I think that would be a big win. If it's not possible, or just a pain, we can figure it out later.

wuzzeb:
Ok, I added a commit which makes psci use the common make code.

paf31:
That was quick :) Thanks!

puffnfresh:
Will hopefully clean up truffled-purescript a little :+1:

paf31:
Looks good to me. @garyb any thoughts on this?

garyb:
All fine by me!

PR #1248, opened 2015-07-10
Pretty print array literals without newlines

wuzzeb:
This is a follow-up to pull #1195.  I ran the most recent pretty-printer through some tests (https://bitbucket.org/wuzzeb/purescript-aeson-interop, although it isn't 100% working but that isn't pretty-printing problem).  There was just one issue with array literals appearing in expressions, after that the output from the pretty printer could be fed back into the compiler.


paf31:
Looks good, thanks. I actually never really intended that the pretty printer would be used to generate code, and it should probably be improved quite a bit. I only ever planned to use it for error messages.

`purescript-aeson-interop` looks exciting :) It is working?

PR #1256, opened 2015-07-13
Fix psc-publish tests

hdgarrood:
For some reason, an explicit "module Main where" is required for other modules to be able to import things from Main.

This is a little program I wrote to test the package serialization of `psc-publish`; it downloads the current master version of `purescript-prelude`, creates a fake tag, runs `psc-publish`, and does a roundtrip to and from JSON.

Currently it's not integrated into the actual test suite, and so it only gets run manually, by me, when I change something to do with serialization. I would be happy to make this into a real test that is listed in the Cabal file (and so gets run by Travis etc).


paf31:
:+1:

PR #1262, opened 2015-07-14
Fix the haddock for the Language.PureScript.Bundle module

wuzzeb:
The bundle module has two haddock comments for the module itself, which resulted in a build error:

https://hackage.haskell.org/package/purescript-0.7.1.0/reports/1

This fixes it by changing the module to only have one comment.  When I moved the code, I didn't see the haddock comment that was already there.  I did test with `stack haddock` and the haddocks build now after this commit.


paf31:
Thanks!

Issue #684, opened 2014-11-07
Add Purescript to Stackage


NightRa:
It would be really good if someone added Purescript to Stackage.
https://github.com/fpco/stackage#get-your-package-included

phadej:
I added purescript to stackage https://github.com/fpco/stackage/pull/730 it will ping me for now.

PR #1263, opened 2015-07-14
Stackage

phadej:
Update travis scripts. Update to use [container infrastructure](http://docs.travis-ci.com/user/workers/container-based-infrastructure/). Builds seems to start faster, but also to take longer (maybe it's the time of the day though).

This is motivated by https://github.com/purescript/purescript/issues/684 issue. I don't see any problems of adding purescript to stackage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ install:
   - cabal --version
   - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
   - travis_retry cabal update
+  - ./travis-stackage.sh
   - cabal install --only-dependencies --enable-tests --enable-benchmarks
 script:
   - ./travis-script.sh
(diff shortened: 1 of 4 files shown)


mjgpy3:
:+1:

hdgarrood:
I'm just trying to understand what's going on here - I thought all that was necessary for getting on Stackage was putting your package name into their build-constraints.yaml file, and then they contact you if there are any problems? What is accomplished by checking that it builds with Stackage in the CI build here?

phadej:
@hdgarrood This refactors travis builds a bit, e.g. verify that `purescript` can be built against newest `lts` and `nightly` stackage snapshots. Maybe those aren't truly necessary, but with them we can be more sure when accepting other pull requests. For example nightly has upper version constraints (so not using the latest version on some package, even when for `purescript` alone build plan could use them).

hdgarrood:
Ah, right, yeah that makes sense. Thanks for explaining.

paf31:
Is there anything left to do on this, now that #1270 is merged?

hdgarrood:
We could still add a couple more entries into our `env` matrix in order to verify `purescript` can be built with Stackage's `lts` and `nightly` snapshots; we don't do that currently.

Also, as Stackage is built with 7.8.4, perhaps it makes sense to switch 7.8.3 -> 7.8.4 for at least the Stackage builds, maybe also the other 7.8 builds as well? I can't see any particular reason to stick with 7.8.3.

phadej:
I'll make a new PR

PR #1266, opened 2015-07-15
fixes #1265: escape all but printable ASCII (except space) in strings

### LINKED ISSUE #1265: escape all non-printable or non-ASCII characters in strings
I ran into this error.

```
$ cat test.purs
module Main where
string = "\x1B"
```
```
$ psc test.purs
Compiling Main
Writing output/Main/index.js
Writing output/Main/externs.purs
```
```
$ psc-bundle 'output/**/*.js'
The module could not be parsed.
```
```
$ cat output/Main/index.js
// Generated by psc version 0.7.1.0
"use strict";
var string = ";
module.exports = {
    string: string
};
```

The error from psc-bundle occurs because a literal `\x1B` character is printed in the string. This is legal JavaScript, but the library we're using to parse it doesn't seem to handle it properly. I'm sure there is also poor support for unicode in this and other JavaScript tooling. We should use escape sequences for all but the simplest set of characters: printable ASCII.

michaelficarra:
Fixes #1265. Untested, since this shouldn't really be observable. Let me know what you think is the best strategy for testing this.

### DIFF
--- a/src/Language/PureScript/Pretty/JS.hs
+++ b/src/Language/PureScript/Pretty/JS.hs
@@ -163,6 +163,7 @@ string s = '"' : concatMap encodeChar s ++ "\""
   encodeChar '\\' = "\\\\"
   encodeChar c | fromEnum c > 0xFFF = "\\u" ++ showHex (fromEnum c) ""
   encodeChar c | fromEnum c > 0xFF = "\\u0" ++ showHex (fromEnum c) ""
+  encodeChar c | fromEnum c > 0x7E || fromEnum c < 0x20 = "\\x" ++ showHex (fromEnum c) ""
   encodeChar c = [c]
 
 conditional :: Pattern PrinterState JS ((JS, JS), JS)


paf31:
Looks good to me. Can we not test by just asserting that the failing example from the issue compiles and runs?

michaelficarra:
That doesn't test this, though. That could be made to work by fixing language-JavaScript. We can't test the actual representation used in the compiled output.

hdgarrood:
I suppose ideally, we could test by by compiling some stuff with psc, then parse it with language-javascript, and then use a function `everythingOnJSNodesTopDownM`, using a writer monad and a function which just logs all the string literals which we encounter. Then we could go through this list of strings and check every char code was in the ASCII range.

Unfortunately `everythingOnJSNodesTopDownM` does not exist, so we'd have to write it by hand.

paf31:
@michaelficarra Sorry this one got left behind a bit. Could you please ELI5 why this is needed, because I'm not sure I see it. Will `language-javascript` parse the output in this format?

michaelficarra:
@paf31 Our code generator, as of `master`, will output string literals containing some non-printable characters. This is allowed, and should work just fine. The problem is, we're relying on a library (language-ecmascript) that has a bug: it cannot parse strings containing these "odd" characters. Instead, we must represent those same characters using an escape sequence. Instead of just labelling this a bug with language-ecmascript and moving on, I am proposing we implement the escaping workaround because I foresee similar issues arising with other poorly written JavaScript tooling.

paf31:
:+1: Thanks!

michaelficarra:
I'd still like to get this tested some day, but it requires considerable effort to do something like @hdgarrood suggested.

hdgarrood:
Turns out it's not that hard! - https://github.com/erikd/language-javascript/pull/41

PR #1270, opened 2015-07-15
Use the new Travis infrastructure

hdgarrood:
Hopefully this should make builds faster.


hdgarrood:
If this works, I'm going to try using [directory caching](http://docs.travis-ci.com/user/caching/) for cabal dependencies which I'm hoping will speed things up significantly.

hdgarrood:
This is starting to get a bit bigger. Now, in addition to using the new infrastructure, these commits also:

* Cache two separate sandboxes: one for running `cabal build && cabal test` in, and one for testing that a source distribution can be installed. This should speed builds up.
* Fix a bug in the scripts which meant that coverage reports never got generated.
* Modify the cabal file so that ghc optimizations `-O2` can be turned off by setting a `development` flag (this will probably be useful for local development too, come to think of it)
* Set the `development` flag unless the current commit is tagged (i.e. it's a release build), disabling -O2. This should further speed things up.
* Separate the build and test-install stages so that Travis tells us how long each individual stage takes.

hdgarrood:
I think this is good now. Unfortunately I didn't see #1263 before I did this. May I ask that if you want to merge both, could you merge this one first? I'll happily fix merge conflicts in #1263 if so. This is just because fixing merge conflicts here if #1263 were merged first would be quite a lot more effort, I think.

hdgarrood:
For some reason, the `cabal build` step is slower on the new infrastructure. In the job for the above commit, it took 1182 seconds (= 19.7 minutes) just to run `cabal build` [see here](https://travis-ci.org/purescript/purescript/jobs/71511086#L295). In contrast, a recent master build managed to run the whole script `travis-script.sh` in 455 seconds (= 7.6 minutes) [see here](https://travis-ci.org/purescript/purescript/jobs/71234540#L442), which includes cabal build, cabal test, and test installing the source distribution (ie compile again)!

This slowdown seems to cancel out with the gains from caching the sandbox, leaving us with more or less the same build time. However, Travis will probably deprecate the old infrastructure eventually, and it seems likely that they'll improve whatever is causing the long build time on the new infrastructure, so I think these changes are still worth making.

paf31:
Thanks!

Issue #1281, opened 2015-07-17
Exhaustivity hangs compiler


nicodelpiano:
When we have a lot of cases to print as warnings, the compiler hangs.

Example:

```haskell
module Main where
 
data Alpha = A | B | C | D | E | F | G | H | I | J | K | L | M | N | O | P | Q | R | S | T | U | V | W | X | Y | Z

f A A A A A A = A
f A B Z X _ _ = A
f B B _ _ _ _ = B
f Z _ Z _ _ Z = Z
```

nicodelpiano:
[Exhaustive Line 244](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Linter/Exhaustive.hs#L244)
Here, I changed `bss` to `take 5 bss`, and the program finished well.

paf31:
Great, could you please make that into a PR?

garyb:
I don't think that's a fix though, is it? I guess unless we just cut it off at a maximum of 5 or whatever warnings...

paf31:
@garyb Why not? `take 5` should be cutting off the number of binders inside a warning. The number of warnings should be one, I think.

garyb:
Oops, I should probably have read the code first.

nicodelpiano:
I've chosen five because it's what haskell takes as a maximum number of cases, is that ok?

paf31:
I think 5 is fine. As I mentioned on IRC, we should be careful to avoid massive binder lists during `checkExhaustive`, but that can be written up as a separate issue.

nicodelpiano:
:+1:

nicodelpiano:
@paf31 I guess this can be closed, I created another issue for avoiding the unnecessary non-exhaustive cases: https://github.com/purescript/purescript/issues/1288

PR #1282, opened 2015-07-17
Print five exhaustivity checking warnings for each definition

nicodelpiano:
This is because the compiler hangs printing the whole set of warnings, as mentioned in https://github.com/purescript/purescript/issues/1281

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -120,7 +120,7 @@ data SimpleErrorMessage
   | TransitiveExportError DeclarationRef [DeclarationRef]
   | ShadowedName Ident
   | WildcardInferredType Type
-  | NotExhaustivePattern [[Binder]]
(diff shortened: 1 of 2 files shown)


paf31:
Maybe we could indicate if there are more than 5 with an ellipsis (...)?

nicodelpiano:
@paf31 Yes, that'd be great

paf31:
You'll have to be careful not to evaluate the whole list though. Using `length` will do so, I think, so you'd need a function `take5 :: [a] -> ([a], Bool)` which matched the first 5 head elements explicitly (lazily).

garyb:
Or maybe `drop 5` and check if the remainder is `null`?

paf31:
Yeah, that works. Might want to do it in one pass though. I think there is a function called `break` or something like that.

garyb:
Ah yep, `splitAt 5` then perhaps? `span` and `break` need a predicate.

PR #1294, opened 2015-07-21
Resolves #883. Look in home directory (after pwd) for '.psci'

### LINKED ISSUE #883: Look for a .psci file in the user's home directory
The file in the current directory takes precedence if it exists.

mjgpy3:
Resolves #883. Tested on Linux, I don't have the hardware to test elsewhere.

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -91,12 +91,15 @@ supportModule =
 
 -- File helpers
 
+onFirstFileMatching :: Monad m => (b -> m (Maybe a)) -> [b] -> m (Maybe a)
+onFirstFileMatching f pathVariants = runMaybeT . msum $ map (MaybeT . f) pathVariants
+
 -- |
 -- Locates the node executable.
 -- Checks for either @nodejs@ or @node@.
 --
 findNodeProcess :: IO (Maybe String)
-findNodeProcess = runMaybeT . msum $ map (MaybeT . findExecutable) names
+findNodeProcess = onFirstFileMatching findExecutable names
   where names = ["nodejs", "node"]
 
 -- |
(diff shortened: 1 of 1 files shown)


psci/PSCi.hs
```diff
@@ -486,18 +486,29 @@ whenFileExists filePath f = do
...
-  exists <- doesFileExist configFile
-  if exists
-  then do
-    ls <- lines <$> readFile configFile
+  localConfigPath <- configFileAt getCurrentDirectory
+  localExists <- doesFileExist localConfigPath
+
+  homeConfigPath <- configFileAt getHomeDirectory
+  homeExists <- doesFileExist homeConfigPath
+  case (localExists, homeExists) of
```
paf31: This looks like it might be a job for `ExceptT` and `<|>`?

mjgpy3: Thanks, I'll look into those! I've never used those before, so I'm really grateful for the suggestion :+1:

paf31: Actually, `MaybeT` would probably do the job, since we don't need an error message.

mjgpy3: Okay, thanks again!

mjgpy3: Do we want to load both ($PWD/.psci and $HOME/.psci) if they exist? I assumed not in the above implementation.

mjgpy3:
@paf31 latest push uses `MaybeT`. Awesome suggestion, worked like a charm! If you can spot any further optimizations, let me know. Tested again on my Linux machine.

paf31:
Looks nice! The only thing I would change is to move `readCommands` inside the `where` block, since that's the only place it's used, right?

mjgpy3:
Good catch! Latest push fixes that (I was trying to minimize the git diff but visibility, or lack thereof, is more important)

paf31:
Looks good, thanks! Will merge once Travis is done.

mjgpy3:
Thanks for the review @paf31. I am learning a lot from these PRs :+1:

paf31:
:zap:

PR #1298, opened 2015-07-23
Resolves #1276. Stop showing "Writing" and "Compiling" messages in psci

### LINKED ISSUE #1276: PSCi has started printing "Writing..." and "Compiling..." messages again

mjgpy3:
Not sure if this is the best approach (it was definitely one of the easiest that comes to mind).

Unfortunately bower's acting up for me so I'm not able to test at this very moment using `psci` (I can in half an hour or so).

It worked for `psc` using the new flag.

Thoughts?

### DIFF
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -25,6 +25,7 @@ module Language.PureScript.Make
   -- * Make API
     RebuildPolicy(..)
   , MakeActions(..)
+  , OutputFormatters(..)
   , SupplyVar()
   , Externs()
   , make
@@ -47,7 +48,7 @@ import Control.Monad.Supply.Class (fresh)
 
 import Data.Function (on)
 import Data.List (sortBy, groupBy)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, isJust, fromJust)
 import Data.Time.Clock
 import Data.Traversable (traverse)
 import Data.Version (showVersion)
@@ -107,7 +108,24 @@ data MakeActions m = MakeActions {
   -- Respond to a progress update.
   --
   , progress :: String -> m ()
+  -- |
+  -- Formatters for output messages
+  --
+  , formatters :: OutputFormatters
   }
+
+data OutputFormatters = OutputFormatters {
+  -- |
+  -- Accepts the name of a module and formats the log message. If `Nothing`
+  -- doesn't output a message
+  --
+    formatCompilingMessage :: Maybe (String -> String)
+  -- |
+  -- Accepts the name of a file path and formats the log message. If `Nothing`
+  -- doesn't output a message
+  --
+  , formatWritingMessage :: Maybe (String -> String)
+}
 
 -- |
 -- Generated code for an externs file.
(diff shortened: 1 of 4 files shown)


mjgpy3:
Just tested in `psci`; compilation messages are not shown, :+1:

paf31:
This code looks fine, but I don't think this is the approach we want:

- I want to avoid options in the library which only apply to one executable
- We already use the `-v` or `--verbose` flags elsewhere.
- We can use a record update to simply override `writeTextFile` for PSCi.

mjgpy3:
@paf31 thanks for the feedback, that makes sense.

mjgpy3:
> We can use a record update to simply override writeTextFile for PSCi.

Can you clarify this some?

mjgpy3:
Are you saying attach a definition of `writeTextFile` to the `Make` `newtype` and just call out to that?

paf31:
I'm saying the `MakeActions` for PSCi can override the `writeTextFile` member so that it doesn't print that message.

mjgpy3:
@paf31, I see. Thank you!

mjgpy3:
@paf31, latest push is in the spirit of your suggestion. Unfortunately overriding the `writeTextFile` wasn't enough since it only accounted for the "Reading" message and was not near the "Compiling" message. Let me know what you think of the new pattern. Thanks for the help!

paf31:
I hate to drag this out, but I want to find the right abstraction here, and I'm not sure this is it. I also don't want to gut `MakeActions` just yet. Let me have a think about this and I'll get back to you.

mjgpy3:
No worries, let me know if you come up with something better

src/Language/PureScript/Make.hs
```diff
@@ -160,7 +178,7 @@ make MakeActions{..} ms = do
     (_, env') <- lift . runCheck' env $ typeCheckModule Nothing m
     go env' ms'
   go env ((True, m@(Module coms moduleName' _ exps)) : ms') = do
-    lift $ progress $ "Compiling " ++ runModuleName moduleName'
+    when (isJust formatter) $ lift $ progress $ fromJust formatter $ runModuleName moduleName'
```
hdgarrood: You can avoid `when (isJust ...) $ fromJust ...` by using `Data.Foldable.for_` instead, with the `Foldable Maybe` instance:

`when (isJust x) $ f $ fromJust x == for_ x f`

This might be easier to read if you define `whenJust` as a specialization of  `for_`, so `whenJust :: (Applicative f) => Maybe a -> (a -> f b) -> f ()`

hdgarrood:
Maybe rather than `OutputFormatters`, a set of callbacks might work better?
```
data MakeCallbacks = MakeCallbacks
  { afterWrite :: FilePath -> IO ()
  , afterCompile :: P.ModuleName -> IO ()
  }
```

paf31:
I'd like to rethink the `MakeActions` API generally.

mjgpy3:
Obviously no rush, @paf31 if you have any neat ideas for reworking `MakeActions` I'd be happy to take a shot at implementing them.

mjgpy3:
@paf31 what if I had a `reportProgress` flag passed to `buildMakeActions` that modifies the `progress` member?

paf31:
Why not just override the progress member directly with a record update?

mjgpy3:
Cool idea, I'll give it a shot!

paf31:
I'll close this, since I patched it in another PR so that we can release 0.7.2. Sorry for the back and forth on this.

PR #1300, opened 2015-07-23
Add --version and --help options to psc-publish

garyb:
Maybe there's room for improvement on the help message, I know it does nothing without a `bower.json`, is there anything else worth mentioning?

### DIFF
--- a/psc-publish/Main.hs
+++ b/psc-publish/Main.hs
@@ -31,6 +31,8 @@ import Control.Monad.Trans.Except
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Writer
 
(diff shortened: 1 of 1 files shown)


hdgarrood:
We could list all the things it expects to be in place (eg, "repository" field in bower.json, etc), but if we do that, people might try to make sure their package is set up correctly before running it. This is probably a lot of wasted effort; I'm sure it would be much easier to just run it and see if it complains. So perhaps it's better to avoid documenting everything.

garyb:
That's fine by me. If people go to the upload page it has a link to a guide on how to use `psc-publish` anyway.

paf31:
:+1:

PR #1304, opened 2015-07-24
Move psc-publish code into the library

hdgarrood:
This doesn't change the behaviour of `psc-publish` at all. There's no obvious need for this now, since we're not planning for the Pursuit server to do any data generation. But I had already made these changes and they seem like the right thing to do anyway.


paf31:
:+1: Seems like a good idea, thanks!

PR #1305, opened 2015-07-25
Include test files for nix package fix

### LINKED ISSUE #1302: nix fails to build from cabal package
https://github.com/NixOS/nixpkgs/issues/8965

If we take a look at the [tar](http://hackage.haskell.org/package/purescript-0.7.1.0/src/) we can see that the `tests/` folder is missing. This [breaks purescript package on nix](https://hydra.nixos.org/build/23422483/nixlog/1#line-553). I suggest adding

```
extra-source-files: examples/passing/*.purs
                  , examples/failing/*.purs
                  , tests/prelude/*.purs
                  , tests/prelude/js/*.js
```
at https://github.com/purescript/purescript/blob/b34ec66d3f360097bd112ae80da1aef853d78114/purescript.cabal#L21

and republishing `version: 0.7.1.1` to hackage. Thanks

garyb:
Should resolve #1302

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -20,6 +20,8 @@ tested-with: GHC==7.8
 
 extra-source-files: examples/passing/*.purs
                   , examples/failing/*.purs
+                  , tests/prelude/*.purs
+                  , tests/prelude/js/*.js
 
 source-repository head
     type: git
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Thanks!

PR #1307, opened 2015-07-25
Improved error reporting in psc-bundle

hdgarrood:
This is along the same lines as #1207 but it uses `ErrorInFile`, although I renamed to `ErrorInModule`, as we don't have the file information readily to hand. I also included the error from language-javascript (which is just a `String`) if the parsing fails. Here's an example output (the error arises from non-ascii characters in string literals, #1265):

```
$ pulp build --include benchmarks --main Benchmark.Main -O --to main.js
* Building project in /home/harry/documents/code/purescript-sequences
* Build successful.
* Bundling Javascript...
Error in module Benchotron.UI.Console (Regular):

  The module could not be parsed:
  lexical error @ line 67 and column 176
```

(by the way, that `--include` option for pulp doesn't exist yet, I'm developing it at the moment)

At the moment, it prints the `ModuleType` (`Regular` or `Foreign`) using the `Show` instance. Perhaps there's a better way of displaying this information?


paf31:
:+1: Thanks!

PR #1308, opened 2015-07-26
Add --dry-run option to psc-publish

hdgarrood:


psc-publish/Main.hs
```diff
@@ -21,8 +26,13 @@ main = execParser opts >> publish
   version :: Parser (a -> a)
   version = abortOption (InfoMsg (showVersion Paths.version)) $ long "version" <> help "Show the version number" <> hidden
 
-publish :: IO ()
-publish = do
-  pkg <- preparePackage
-  BL.putStrLn (A.encode pkg)
-
+publish :: Bool -> IO ()
```
paf31: This all looks good, but do you think it's worth introducing an Options type now, so that we don't have to break the API again later?

hdgarrood: Yeah - I did introduce PublishOptions: https://github.com/purescript/purescript/pull/1308/files#diff-072e63067f4580fc7a99055d7d5655d0R55

I don't really consider this function part of the API, I think the real API is in `Language.PureScript.Publish`. If we do want to expand the command line API later, then this function's type will have to change, but considering that this module can't be imported anywhere*, I don't think that will be a problem.

*at least, I am pretty sure that this is the case.

PR #1316, opened 2015-07-29
Update test suite to use real supporting libraries

hdgarrood:
* Use real versions of supporting libraries
* Use bower to make it easy to update supporting libraries

The reason I did this is that I noticed that some of the tests had started producing warnings, because of type classes with operator members, and I think this might have been interfering with their results. (Investigating now).


hdgarrood:
The bower.json for the set of supporting libraries is currently pointing to a specific commit in my fork. Once https://github.com/purescript/purescript-assert/pull/1 is merged we can point at the normal version.

hdgarrood:
The CI failure on the "fix psci tests" commit turns out to be a real bug, which was that type class members didn't get completed in psci. I've just fixed this.

paf31:
Looks good but did you mean to check in the `flattened` directory?

hdgarrood:
I did, I was thinking it would be nice to be able to run the tests without needing bower. I don't feel very strongly either way though. Would you rather the tests shelled out to bower at the beginning to get this code?

paf31:
Hmm, I'm not sure. It'd be nice to be able to accept PRs from Haskellers who don't have Bower, I suppose. I'm not sure it's so much more convenient that it's worth all the duplication of code though TBH.

hdgarrood:
I don't think PRs are a problem if we don't check it in - if the submitter hasn't set their environment up, it won't matter, because we'll hear from Travis soon after they create the PR. I'll change that, then.

natefaubion:
Does it make sense to pull from master for these repos? I'm thinking if one makes a Prelude change, and it needs to be tested before a release.

hdgarrood:
What sort of change are you thinking of? I think the test suites for individual libraries are better equipped to test changes in those libraries, and the tests here should really just be for the compiler.

natefaubion:
I was thinking about the bug wrt to inlining boolean operators. Changes were made to the published Prelude that affected the compiler's optimizations. If we change something like that again, would we want to coordinate and do compiler/library releases at the same time?

paf31:
:+1: Thanks very much!

paf31:
Something isn't quite right with the tests now, but I'm not sure if it's related to this. Several tests break on my machine, but `cabal test` is reporting success.

hdgarrood:
Perhaps because not all of the instances have been updated? The old Prelude was taken from before all of the operator members had been changed, so for example `Eq` had a member `(==)`. I was planning on fixing all these in #1323.

paf31:
The code compiles fine with `psc` though.

paf31:
Also, why would `cabal test` pass?

paf31:
Ah, had to delete `.test_modules`. Sorry, never mind...

Issue #1181, opened 2015-06-19
Verify 'failing' tests fail in the correct way


paf31:
We should use the `ErrorMessage` type to make sure that each failing test is generating the correct general category of error.

puffnfresh:
This was something I was worried about while making my optimisation changes :+1:

mjgpy3:
Maybe it would be useful to include descriptions of the example failure message. I've started running through them and am having difficulty understanding if a given failure is as originally intended, or due to some new change.

mjgpy3:
Or maybe rename the files to be more descriptive, e.g. `DuplicateDeclarationsInLet.purs`

LiamGoodacre:
As part of `tests/Main.hs` would we explicitly write out the test cases loading specific examples and checking the outcome?

Could probably do it incrementally too - where we track tested cases and those that are missed are still checked for compilation failure as they are now?

---

Could I try to do this once we've decided what to do?  As my first contribution to the project :smile:

paf31:
One approach might be to modify the tests temporarily to print out the filename and failure code, then turn that into the mapping we need by hand.

As for where the mapping goes, it could live in code, or just in a text file inside the examples/failing directory.

PR #1323, opened 2015-08-02
Check the right kind of error is thrown in tests

hdgarrood:
Add a basic mechanism for declaring what kind of errors a test should
throw, and add a few of these declarations to failing examples.

If you think this looks good I'll go ahead and add these declarations to all of the failing examples.

Refs (possibly fixes?) #1181


paf31:
Looks fantastic, thank you! Once we have all tests marked up, perhaps it makes sense to use the lack of an annotation to mean "should pass"?

michaelficarra:
@hdgarrood Can you show some examples of multiple errors or compilation errors?

hdgarrood:
> Once we have all tests marked up, perhaps it makes sense to use the lack of an annotation to mean "should pass"?

I'm not sure, we do already have a way of indicating this based on what directory the file is in; that is, `passing` vs `failing`. I do think it would be a good idea to make sure that every failing test does fail with a specific error though; perhaps a good way to achieve this would be to require that every test inside the `failing` directory has at least one shouldFailWith declaration?

> Can you show some examples of multiple errors or compilation errors?

I've put 3 examples in already, I'm going to start going through and adding more now. Come to think of it I'm not sure if there are any that should produce more than 1 error, we'll see in a moment.

examples/failing/MultipleErrors.purs
```diff
@@ -1,11 +1,13 @@
+-- @shouldFailWith ExprDoesNotHaveType
```
hdgarrood: @michaelficarra Here's an example with a test raising multiple errors.

hdgarrood:
Quite a few of these no longer testing what they are supposed to test since 0.7! I think this is all good now, though.

hdgarrood:
Also, some next steps from here:
* add tests for suggestions (I think we should have a real data type for suggestions too, not just strings).
* add tests for warnings
* work out how to run all the tests in the manual directory automatically. Perhaps this mechanism is enough, but we can presumably also embed specific examples inside the Haskell source if we need a more elaborate check.

paf31:
Brilliant! I've wanted this for so long and never got round to implementing it. Thanks!

PR #1324, opened 2015-08-02
Fix optimizer for booleanAlgebraBoolean

### LINKED ISSUE #1312: Optimizer rules broken after BooleanAlgebra change
https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs#L159

garyb:
Resolves #1312. Tested locally for now:

``` purescript
main = do
  print $ not $ true && false || false || true
  print $ (true `conj` not false) `disj` true
```

``` javascript
var main = function __do() {
    Control_Monad_Eff_Console.print(Prelude.showBoolean)(!(true && false || (false || true)))();
    return Control_Monad_Eff_Console.print(Prelude.showBoolean)(true && !false || true)();
};
```


paf31:
:+1: Thanks!

PR #1285, opened 2015-07-19
Improve error message formatting for tool interop

epost:
Improve editor interop by printing compiler errors in a format like that of gcc, ghc, javac, etc.

The new format allows editors like emacs to highlight errors and warnings correctly, and in such a way that the highlighting depends on the error level. Additionally, (emacs) features like 'go to error line' and 'go to next error' now work (more) correctly. See also #1098.

Changes:

- The headings ``Error:`` and ``Warning:`` at the top of the error list are interpreted by emacs as proper errors, which causes problem with highlighting and with jumping to the error line, because it doesn't exist. Fix: replace ``Error:`` by ``Error found:`` and so on. The introduction of a space character fixes the problem.

- ``PositionErrors`` are now formatted like ``/path/to/File.purs:3:5:warning:Some descriptive message``, following gcc. This allows emacs to parse the source position and error loevel correctly.

- The ending positions of the span are no longer printed. They are still available in the JSON format.

Before:

    Error:
    Error in module Foo:
    Error at /Users/erik/dev/ex/examples-purescript/purescript-0.7-project/error-msg-formats/bla.purs line 3, column 1 - line 4, column 1:
      Orphan type declaration for f

After:

    Error found:
    Error in module Foo:
    /Users/erik/dev/ex/examples-purescript/purescript-0.7-project/error-msg-formats/bla.purs:3:1:Error:
      Orphan type declaration for f

PR #1327, opened 2015-08-02
Make error list heading emacs-friendly

epost:
Reworded the text here to introduce a space, so that emacs no longer incorrectly interprets the lines with `Error:` and `Warning:` as proper errors. This fixes  error highlighting and 'jump to next error' and similar functionality in the case of a single warning/error.

These changes were also a part of https://github.com/purescript/purescript/pull/1285.


paf31:
:+1: Thanks!

PR #1340, opened 2015-08-07
Fix #1260

### LINKED ISSUE #1260: <<loop>> in psc
This error can be reproduced by reversing the arguments `x` and `next` here:

https://github.com/purescript-contrib/purescript-channels/blob/096a1027a24de931d2296200ed580855f8109c70/src/Channels/Combinators.purs#L13

It's not a pretty printer error, as it usually is, because I replaced all of the pretty printers with `show` to test this and it still happens. My money is on the type class instance lookup function, since it's the only other place I can think of which involves a possibly non-terminating loop.

paf31:

Not sure if we want to make a patch release or not, this bug is pretty nasty.

### DIFF
--- a/examples/failing/InfiniteKind.purs
+++ b/examples/failing/InfiniteKind.purs
@@ -1,0 +1,5 @@
+-- @shouldFailWith InfiniteKind 
+
+module Main where
+
+data F a = F (a a)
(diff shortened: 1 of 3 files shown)


paf31:
Perhaps this one is sufficiently trivial to fall through the review process actually :) I'll merge.

garyb:
Ooh snap, that makes sense.

paf31:
`git bisect` is pretty handy every now and then :smile:

PR #1342, opened 2015-08-08
Tune GHC parameters

phadej:
`-j2` actually decreases parallellism, which on shared travis containers cause bad performance. See https://ghc.haskell.org/trac/ghc/ticket/9221

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,14 @@ install:
   - echo "$(ghc --version) [$(ghc --print-project-git-commit-id 2> /dev/null || echo '?')]"
   - travis_retry cabal update
   - cabal sandbox init
+  # Run with limited concurrency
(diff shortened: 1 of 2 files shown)


paf31:
I'm not sure I follow about `-j2`. Are you saying you are decreasing parallelism on purpose, to increase the speed of Travis builds? 

Could you please explain the `-A32m` option too?

Maybe these could benefit from a code comment as well, to explain both of these things.

paf31:
Also, thanks for working on this :smile:

phadej:
Added comments. Build times seems to be largely improved (down to ~7 min, from over 20 for >=GHC7.8 builds).

paf31:
This does look much better, thanks. :+1:

hdgarrood:
:100: amazing!

PR #1346, opened 2015-08-08
Enhance core tests script

erdeszt:
Update core-tests/test-everything.sh with a few small things. I hope this will improve the experience for others as well, it proved helpful to me.

* Only install bower if it's not already installed.
* Accept the newer version in case of conflicts in bower install.
* Add force recompile (-c) and force reinstall (-i) options. It's useful if you want to run it multiple times.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ cabal.sandbox.config
 .psci_modules/
 .test_modules/
 bower_components/
+node_modules
 tmp/
 .stack-work/
 tests/support/flattened/
--- a/core-tests/test-everything.sh
+++ b/core-tests/test-everything.sh
@@ -18,11 +18,10 @@ if [ "$force_reinstall" = "true" ] && [ -d "bower_components" ]; then
   rm -r bower_components
 fi
 
-if ! type bower ; then
-  npm install -g bower
-fi
+npm install bower
 
-bower i purescript-prelude \
+node_modules/.bin/bower i \
+        purescript-prelude \
         purescript-eff \
         purescript-st \
         purescript-integers \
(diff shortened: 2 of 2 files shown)


hdgarrood:
Perhaps we should consider installing bower without the `-g` flag, and then invoking it as `node_modules/.bin/bower`?

erdeszt:
@hdgarrood Fine for me. Drop a comment if you want me to update the PR.

michaelficarra:
> Perhaps we should consider installing bower without the `-g` flag

Yes.

erdeszt:
I've updated the PR with the local bower install.

hdgarrood:
Also, I'm a little hesitant about simply accepting the newer version when conflicts arise. This script is mainly meant to be run as part of the CI build, as far as I'm aware, and if a situation arises where it's not possible to install all of these libraries side-by-side, I think we probably do want it to fail. But I'm not sure about this, perhaps one of the real maintainers can clarify :)

paf31:
Yes @hdgarrood is right, it should fail in that case.

erdeszt:
@hdgarrood @paf31 I'm fine with that option as well, but as far as I know bower doesn't support this behavior and I'm not familiar with Travis/Bash enough to set something up in the travis/after.sh. If you guys want I can remove that option from the PR and we can address this in a separate branch when we have the right solution.

paf31:
Any updates on this?

erdeszt:
@paf31 I can remove the commit that enables the "accept newer" policy in bower and the rest should be fine. Is that ok?

paf31:
Yes I think so, thanks!

erdeszt:
Removed the aforementioned commit.

paf31:
:+1: Thanks!

PR #1348, opened 2015-08-09
Tidy up MakeActions API, close #1264

### LINKED ISSUE #1264: Review MakeActions API
It's not clear to me how it works. Maybe we can revisit it when we implement parallel builds?

paf31:
@garyb Does this look ok?

I want to look at parallel builds in 0.8 for performance reasons and tidying this up a bit first seemed useful.

### DIFF
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -24,6 +24,7 @@ import Control.Monad
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Writer
 
+import Data.Maybe (fromMaybe)
 import Data.Version (showVersion)
 import qualified Data.Map as M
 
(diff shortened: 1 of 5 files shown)


garyb:
I guess the problem is now the other backends can't use `MakeActions`, that's a big part of why I did things the way they were previously - to keep `make` generally useful, not just for `psc`.

paf31:
Oh, I thought we were just trying to abstract away `IO`. That makes sense.

paf31:
I'll keep the branch around while I experiment with parallel builds though.

garyb:
I'm sure it can probably still be improved, but yeah, sorry I should have mentioned that was the motivation behind the previous version.

PR #1349, opened 2015-08-09
Fix #1328

### LINKED ISSUE #1328: Linter runs every time on unmodified modules
Should this be the case? We don't get warnings about partial matches every time the compiler runs, only shadowed name warnings - but also, this is probably not helping incremental performance... not that it will have any impact on the overall memory usage explosion unfortunately.

paf31:
/cc @garyb @hdgarrood 

How's this?


paf31:
Actually, there is a slight issue with this. Investigating...

PR #1350, opened 2015-08-09
Typecheck main, fix #1333

### LINKED ISSUE #1333: Warnings for missing top-level type declarations

paf31:
@garyb What do you think? Anything with the name `main` is now typechecked against the type `Eff _ _`.

Maybe we could add an option `--no-main-check` in case this is too strict?


garyb:
_Anything_ with the name `main`? I think that could well be a bit too strict... a tricky problem now given that the main argument is from `psc-bundle` I guess. It seems like we should do the check there somehow, but that's not really possible either. Can we glean some clue from the JS as to whether it might be `Eff` instead? Not reliably I suppose. Argh.

PR #1351, opened 2015-08-09
Don't lint modules if they won't be rebuilt

paf31:
/cc @garyb @hdgarrood Slightly simpler solution than in my other PR.

### DIFF
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -49,6 +49,7 @@ import Data.Function (on)
 import Data.List (sortBy, groupBy)
 import Data.Maybe (fromMaybe)
 import Data.Time.Clock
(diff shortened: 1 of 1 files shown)


paf31:
Looks like builds are failing because of a `core-tests` version mixup. @garyb Do we need to bump some core library dependencies or is this an issue with the `bower.json` file?

garyb:
We need to bump some core libraries for `strings` I think... _maybe_ something else too. I'll see what I can do.

garyb:
Fix looks nice though.

garyb:
Actually QC was the culprit, I was holding off on a release until we fixed the stack safety issues but I think that's still being looked at. Restarted Travis checks now so hopefully will be green once they're done.

PR #1354, opened 2015-08-09
MakeActions changes

### LINKED ISSUE #1264: Review MakeActions API

paf31:
@garyb This is just some minor stuff, and adding a comment, now that I understand the role of `MakeActions` better.

Resolves #1264.


garyb:
:+1: Using `SupplyT` is much more sensible.

You mentioned you were working on parallel builds and that `MakeActions` ties in with that, does the current design hamper parallel builds? If so I guess we can't keep it as is and will have to specialise for different backends. If that's the case then I'm all for it, just at the time I updated this it seemed like sharing it would be viable.

paf31:
No, it should be fine. I was planning to make slight changes like this before parallel builds, but got carried away.

paf31:
@garyb I just pushed another change to reduce some code duplication in the test suite. Hopefully it's ok too.

garyb:
For sure! Much better :+1:

PR #1358, opened 2015-08-11
Fix #1246

### LINKED ISSUE #1246: psci: ability to reload FFI definitions without losing bindings from the session
This would be useful when testing things inside psci.

paf31:
/cc @garyb @hdgarrood

### DIFF
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -84,7 +84,7 @@ readInput InputOptions{..} = forM ioInputFiles $ \inFile -> (Right inFile, ) <$>
 parseInputs :: (Functor m, Applicative m, MonadError P.MultipleErrors m, MonadWriter P.MultipleErrors m)
             => [(Either P.RebuildPolicy FilePath, String)]
             -> [(FilePath, P.ForeignJS)]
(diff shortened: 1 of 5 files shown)


paf31:
The issue seems to be that the content of the foreign module is captured inside the `MakeActions` structure.

garyb:
Oops! It's much simpler now too, not sure why I didn't do it this way in the first place. :+1:

garyb:
Huh, looks like the core library stuff is working now, it's one of the stackage builds that is failing.

paf31:
@phadej Do you have any idea why this might happen? Maybe a build cache issue?

paf31:
Looks like the sandbox is being cached between different builds, perhaps? Maybe we have to revert to no caching.

hdgarrood:
Yeah, I think this is because of the sandbox being cached between builds (which was intended). However, oddly, it's the Stackage LTS 2.22 build that's failing; surely, if the Stackage snapshot version is fixed, none of the dependencies' versions should change? I think we should start by clearing the Travis cache and rerunning the build. I think this can be done inside Travis' web UI.

Perhaps it also makes sense to add `--force-reinstalls` to the `cabal install` command?

paf31:
My guess is that it's using the same cache between all of the different builds?

PR #1363, opened 2015-08-12
Test Data.Generic in core-tests instead

paf31:


paf31:
@garyb I'll merge this then release 0.7.3. Sound good?

garyb:
Sounds great! What was that other PR that said it made the tests too slow? How did it differ from this?

paf31:
This just ensures that the new module builds, it doesn't run it, which requires reading a whole bunch of dependency extern files in every test. For this, compiling is probably enough.

Issue #1268, opened 2015-07-15
Easier FFI inclusion for `psc` command-line


mjgpy3:
Hello all,

I tend to compile my PureScript code using `psc` directly a good amount. Is there value in making JS FFI files included more easily? I can think of a couple approaches off the top of my head, the one that sounds most promising is a new flag, something like:
```
psc --auto-ffi-ext=js ...
```
Here it would simply look for files near the `*.purs` files, with the same names (except `js` in place of `purs`), and automatically include them as if they were given to `--ffi`.

Other approaches I can think of are:
 1. Simply assuming the file with the same name (but different extension) is an FFI
 2. Allowing the user to type the JS file name, without the `--ffi` flag, inspect the extension, and conclude it's an FFI

Thoughts?

michaelficarra:
I like other option 2.

paf31:
Option 2 sounds nice.

garyb:
We currently don't even check if the PureScript source files have any particular extension, so we'll have to do that too :wink:

michaelficarra:
That's okay, I think `.purs` is official enough.

mjgpy3:
Ah sorry, that button order gets me sometimes. So other option 2 it is?

garyb:
I only wonder what effect it would have on #1269

mjgpy3:
@garyb, what potential effect(s) are you considering?

garyb:
Uhh, don't mind me. I misread what option 2 was proposing.

PR #1364, opened 2015-08-12
Resolves #1176. Warn when psc is given globs that match no files.

### LINKED ISSUE #1176: Feature request: `psc-make` warn about non-existent files
Recently I ran a command akin to:
```
~/.cabal/bin/psc-make purescript-prelude/src/Prelude.purs --ffi purescript-prelude/src/Prelude.js  IDontExist.purs Foobar.purs
```

And got the output:
> Compiling Foobar
Writing output/Foobar/index.js
Writing output/Foobar/externs.purs

It would have been nice if `psc-make` told me that `IDontExist.purs` didn't exist, since it was what I was actually interested in building/testing. Thoughts?

mjgpy3:
@paf31, I hope you don't mind that I worked on this even though you are assigned -- it seemed like a fun one. Let me know your thoughts.

### DIFF
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -52,12 +52,12 @@ data InputOptions = InputOptions
 
 compile :: PSCMakeOptions -> IO ()
 compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = do
-  input <- concat <$> mapM glob inputGlob
+  input <- globWarningOnMisses warnFileTypeNotFound inputGlob
   when (null input) $ do
     hPutStrLn stderr "psc: No input files."
     exitFailure
   moduleFiles <- readInput (InputOptions input)
-  inputForeign <- concat <$> mapM glob inputForeignGlob
+  inputForeign <- globWarningOnMisses warnFileTypeNotFound inputForeignGlob
   foreignFiles <- forM inputForeign (\inFile -> (inFile,) <$> readFile inFile)
   case runWriterT (parseInputs moduleFiles foreignFiles) of
     Left errs -> do
@@ -77,6 +77,18 @@ compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = d
           when (P.nonEmpty warnings') $
             hPutStrLn stderr (P.prettyPrintMultipleWarnings (P.optionsVerboseErrors opts) warnings')
           exitSuccess
+
+warnFileTypeNotFound :: String -> IO ()
+warnFileTypeNotFound = hPutStrLn stderr . ((++) "psc: No files found using pattern: ")
+
+globWarningOnMisses :: (String -> IO ()) -> [FilePath] -> IO [FilePath]
+globWarningOnMisses warn = concatMapM globWithWarning
+  where
+  globWithWarning pattern = do
+    paths <- glob pattern
+    when (null paths) $ warn pattern
+    return paths
+  concatMapM f = liftM concat . mapM f
 
 readInput :: InputOptions -> IO [(Either P.RebuildPolicy FilePath, String)]
 readInput InputOptions{..} = forM ioInputFiles $ \inFile -> (Right inFile, ) <$> readFile inFile


mjgpy3:
Examples:

```
$ ./dist/build/psc/psc IDont*.purs
psc: No input files found using: "IDont*.purs"
psc: No input files.
```

```
$ ./dist/build/psc/psc IDont*.purs IExist.purs 
psc: No input files found using: "IDont*.purs"
Compiling IExist
```

```
$ ./dist/build/psc/psc IExist.purs  --ffi IDontExist.ffi
psc: No foreign imports found using: "IDontExist.ffi"
Compiling IExist
```

```
$ ./dist/build/psc/psc IE*purs
Compiling IExist
```

psc/Main.hs
```diff
@@ -78,6 +78,20 @@ compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = d
             hPutStrLn stderr (P.prettyPrintMultipleWarnings (P.optionsVerboseErrors opts) warnings')
           exitSuccess
 
+globWarningOnMisses :: String -> [FilePath] -> IO [FilePath]
```
paf31: This looks like a job for [`concatMapM`](http://haddock.stackage.org/lts-2.22/extra-1.1/Control-Monad-Extra.html#v:concatMapM). We could use the `extra` library, or just define it as a utility function here.

psc/Main.hs
```diff
@@ -78,6 +78,20 @@ compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = d
...
+globWarningOnMisses fileType = go ([], [])
+  where
+  go (hits, misses) [] = do
+    unless (null misses) $ do
+      let formattedMisses = tail $ init $ show misses
+      hPutStrLn stderr $ "psc: No " ++ fileType ++ " found using: " ++ formattedMisses
```
paf31: Instead of piecing strings together, maybe take a function with throws the warning?

psc/Main.hs
mjgpy3: Thanks @paf31, looking now!

psc/Main.hs
```diff
@@ -52,12 +52,12 @@ data InputOptions = InputOptions
...
 compile :: PSCMakeOptions -> IO ()
 compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = do
-  input <- concat <$> mapM glob inputGlob
+  input <- globWarningOnMisses "input files" inputGlob
   when (null input) $ do
     hPutStrLn stderr "psc: No input files."
     exitFailure
   moduleFiles <- readInput (InputOptions input)
-  inputForeign <- concat <$> mapM glob inputForeignGlob
+  inputForeign <- globWarningOnMisses "foreign imports" inputForeignGlob
```
paf31: Perhaps this is a good time to deal with #1268 as well, then we don't need two separate calls to `globWarningOnMisses`.

We'd have to try to be backwards compatible though.

paf31:
Not at all, I'm grateful that you're working on it :)

psc/Main.hs
mjgpy3: Would it be alright if I resolved this first then worked on #1268 in a separate PR?

paf31: Yeah, of course.

psc/Main.hs
```diff
@@ -78,6 +78,20 @@ compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = d
             hPutStrLn stderr (P.prettyPrintMultipleWarnings (P.optionsVerboseErrors opts) warnings')
           exitSuccess
 
+globWarningOnMisses :: String -> [FilePath] -> IO [FilePath]
+globWarningOnMisses fileType = go ([], [])
+  where
+  go (hits, misses) [] = do
+    unless (null misses) $ do
+      let formattedMisses = tail $ init $ show misses
```
paf31: Why `tail . init` out of curiosity?

paf31: Oh I see, it's a string. Maybe use `intercalate ", "` instead.

mjgpy3: I initially used that, but it was nastier than showing and chopping the ends off. The `concatMapM` pattern leads me to printing a warning for each glob pattern anyways, so this will be much cleaner

paf31: If we're going to use a warning for each glob anyway, I'd just use `for_`.

paf31: Also, I disagree about chopping being less nasty :) It's a small example, but it's a bad practice in general because you're forcing evaluation of the whole string.

mjgpy3: Good point! Thanks for the advice.

mjgpy3:
@paf31 the suggestions cleaned things up a good bit, thanks! Let me know your thoughts.

mjgpy3:
Here's an example:
```
$ ./dist/build/psc/psc We Do Not Exist IExist*purs --ffi NoExisto.js
psc: No input files found using pattern: We
psc: No input files found using pattern: Do
psc: No input files found using pattern: Not
psc: No input files found using pattern: Exist
psc: No foreign imports found using pattern: NoExisto.js
Compiling IExist
```

psc/Main.hs
```diff
@@ -52,12 +52,12 @@ data InputOptions = InputOptions
...
 compile :: PSCMakeOptions -> IO ()
 compile (PSCMakeOptions inputGlob inputForeignGlob outputDir opts usePrefix) = do
-  input <- concat <$> mapM glob inputGlob
+  input <- globWarningOnMisses (warnFileTypeNotFound "input files") inputGlob
   when (null input) $ do
     hPutStrLn stderr "psc: No input files."
     exitFailure
   moduleFiles <- readInput (InputOptions input)
-  inputForeign <- concat <$> mapM glob inputForeignGlob
+  inputForeign <- globWarningOnMisses (warnFileTypeNotFound "foreign imports") inputForeignGlob
```
paf31: Nitpick: `foreign import` usually means a declaration. This might be confusing. I would just have both messages say "No files found using pattern:" to keep it simple. It will be obvious which type from the extension, and in any case, we're going to get rid of `--ffi` eventually anyway.

mjgpy3:
@paf31, sounds good, latest push fixes that

paf31:
:+1: Thanks!

PR #1366, opened 2015-08-12
Bump upper bounds on aeson

lukerandall:
Aeson 0.9 seems a fairly innocuous update. Moreover, the lack of support for aeson 0.9 is preventing stackage from upgrading (fpco/stackage#572).

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -44,7 +44,7 @@ library
                    pattern-arrows >= 0.0.2 && < 0.1,
                    time -any,
                    boxes >= 0.1.4 && < 0.2.0,
-                   aeson >= 0.8 && < 0.9,
+                   aeson >= 0.8 && < 0.10,
                    vector -any,
                    bower-json >= 0.7,
                    aeson-better-errors >= 0.8,


garyb:
Thanks!

phadej:
@garyb btw, could you or someone, make a new revision on hackage: `0.7.2.1` with bumped version? Or will it be against your practicies?

garyb:
That should be no problem at all, @paf31 has the keys to hackage so he'll have to do it though. :)

@paf31 I guess we might want to cherry-pick this commit for a 0.7.2.1 release, as we don't want the generics stuff included in that - if we can make this just a hackage release/fix we won't have versioning problems with the `npm` release, etc. for something that isn't going to affect the binaries anyway.

phadej:
you don't need even to upload new package. I'll find a howto, sec.

phadej:
There is a convenient tool: https://github.com/hvr/hackage-cli

paf31:
Why not just release 0.7.3 with generics now? It's tested.

garyb:
Oh, ok, sure :)

phadej:
0.7.3 is even better! :+1:

paf31:
0.7.2.1 is on Hackage with this patch. Thanks!

lukerandall:
Thanks everyone for the quick turnaround on this!

PR #1373, opened 2015-08-13
Standalone Deriving

### LINKED ISSUE #1371: Derive instances, not instance members?
We could add new syntax to derive a whole `Generic` instance, instead of deriving any missing instance members. It seems like asking for trouble to let users derive some but not all of an instance's members.

Hopefully we could come up with syntax which would generalize to `newtype` deriving unambiguously:

```purescript
deriving instance genericFoo :: Generic Foo

deriving newtype instance showBar :: Show Bar
```

perhaps?

paf31:
@garyb @gbaz @jdegoes Could you please review this?

Resolves #1371

This uses Haskell-like standalone deriving syntax, forcing users to implement all methods for an instance, or derive the whole thing:

```purescript
deriving instance genericFoo :: Generic Foo
```

It also moves deriving into the `Sugar` phase, desugaring derived instances into regular instances. I pull the data declaration info out of the current module, which means

a) it has to be a data declaration
b) it has to be in the current module (no orphans, hardly a big deal)

Thanks.


garyb:
This is really picky, but I still think it should be something other than `deriving` - it makes sense when attached to a declaration in Haskell, but not so much at the start of a standalone declaration. The reasoning for my other options:

- `derive` as in an instruction to the compiler to do so
- `derived` as a declaration that this thing exists through deriving

Another suggestion would be `auto instance` or something like that, but that might be sligtly less good for the eventual newtype case.

garyb:
:+1: for the PR though :smile:

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -154,7 +154,7 @@ data Declaration
   -- A type instance declaration (name, dependencies, class name, instance types, member
   -- declarations)
   --
-  | TypeInstanceDeclaration Ident [Constraint] (Qualified ProperName) [Type] [Declaration]
+  | TypeInstanceDeclaration Ident [Constraint] (Qualified ProperName) [Type] (Maybe [Declaration])
```
jdegoes: Another option here would be a separate term in the sum which encodes derived instances, with the advantages that (a) you don't force all the code to handle the `Nothing` case, and (b) it's arguably another constructor may be clearer to developers than an optional `Declaration`.

jdegoes:
Looks good. I would also echo @garyb's suggestion to name this something like `derive` or `auto`, on grounds that the current syntax is ungrammatical ('data foo deriving x' makes sense, 'deriving x' less so).

paf31:
Yeah agreed on `deriving`. I was copying Haskell's `StandaloneDeriving` extension, but I suspect they just wanted to save adding a new keyword.

src/Language/PureScript/AST/Declarations.hs
paf31: I think I'd prefer to reuse it, but to use a sum type inside in place of `Maybe`. That way, we can extend it to `newtype` deriving easily later, and it's still clear.

gbaz:
Looks good to me. Moving to the earlier phase as pure sugar is clearly the right thing, and adding syntax is the right thing -- I just wanted to leave it to the "owners" to figure out the right syntax :-)

Also agreed that as long as a keyword is added `derive instance GenericFoo` is nicer.

paf31:
@gbaz Does the last commit look ok? Seemed like an easy restriction to lift, but I'm not sure if there's any reason not to add it.

paf31:
Ok, this is ready for review again. `deriving` -> `derive` and getting rid of `Maybe` are both done.

jdegoes:
:+1:

gbaz:
@paf31 lgtm.

PR #1376, opened 2015-08-13
Disallow orphan instances

### LINKED ISSUE #1247: Disallow orphan instances
We keep talking about this but don't seem to have an issue for it. I'm quite an enthusiastic proponent of disallowing them and not providing an escape hatch either.

Worth discussing too perhaps, what counts as an orphan in the presence of multi-parameter typeclasses? I guess for an instance to not be an orphan it must satisfy either:

1. The instance is declared in the same module as its class.
2. At least one of the types appearing in the instance head is declared in the current module.

garyb:
Resolves #1247

### DIFF
--- a/examples/failing/OrphanInstance.purs
+++ b/examples/failing/OrphanInstance.purs
@@ -1,0 +1,12 @@
+-- @shouldFailWith OrphanInstance
+module Class where
+
+  class C a where
+    op :: a -> a
+
+module Test where
+
+  import Class
+
+  instance cBoolean :: C Boolean where
+    op a = a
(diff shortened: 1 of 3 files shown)


puffnfresh:
:fireworks:  :+1:

tonymorris:
:+1:

Typo in comment: `OrhpanInstance`

garyb:
I thought you'd like this :smile:

garyb:
@tonymorris Ah thanks!

paf31:
:100:

Hopefully we can use this restriction to speed up instance search a bit too.

garyb:
Initial attempt at the error message page: https://github.com/purescript/purescript/wiki/Error-Code-OrphanInstance

It's hard to write about without being confusing!

PR #1378, opened 2015-08-13
Added stack config for building with 7.10

zudov:
Now when lts-3.0 came out it makes sense to provide config for building against it.

There are two configs, for building with 7.8 or with 7.10. `stack.yaml` symbolic
link points to 7.10 config.


phadej:
I thought about this, maybe we should have `stack-lts-2`, `stack-lts-3` and `stack-nightly` ?

zudov:
@phadej It makes sense.

One issue with `stack-nightly`, since stack doesn't allow to point to "latest nightly release" (would make builds inconsistent), we would need to regularly update the `resolver` field. I think it would be OK, if we change nightly resolver on every release.

paf31:
:+1: Thanks!

PR #1379, opened 2015-08-13
Avoid massive binder list in `checkExhaustive`.

nicodelpiano:
- Fixed comments.
- Changed `foldl'` to `foldr`.
- Removed `sequenceA` and used `sequence` instead.
- If a case is redundant don't add the uncovered cases.

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -9,10 +9,10 @@
 -- Portability :
 --
 -- |
--- | Module for exhaustivity checking over pattern matching definitions
--- | The algorithm analyses the clauses of a definition one by one from top
--- | to bottom, where in each step it has the cases already missing (uncovered),
--- | and it generates the new set of missing cases.
+-- Module for exhaustivity checking over pattern matching definitions.
+-- The algorithm analyses the clauses of a definition one by one from top
+-- to bottom, where in each step it has the cases already missing (uncovered),
+-- and it generates the new set of missing cases.
 -- 
 -----------------------------------------------------------------------------
 
@@ -26,7 +26,7 @@ module Language.PureScript.Linter.Exhaustive
 
 import qualified Data.Map as M
 import Data.Maybe (fromMaybe)
-import Data.List (foldl', sortBy, nub)
+import Data.List (sortBy, nub)
 import Data.Function (on)
 
 import Control.Monad (unless)
@@ -43,6 +43,12 @@ import Language.PureScript.Types as P
 import Language.PureScript.Errors
 
 import Language.PureScript.AST.Traversals (everywhereOnValuesTopDownM)
+
+-- |
+-- The maximum number of non-exhaustive cases.
+--
+maxCases :: Int
+maxCases = 1000
 
 -- |
 -- Qualifies a propername from a given qualified propername and a default module name
@@ -210,7 +216,7 @@ isExhaustiveGuard (Left gs) = not . null $ filter (\(g, _) -> isOtherwise g) gs
 isExhaustiveGuard (Right _) = True 
 
 -- |
--- Returns the uncovered set of case alternatives
+-- Returns the uncovered set of case alternatives.
 --
 missingCases :: Environment -> ModuleName -> [Binder] -> CaseAlternative -> ([[Binder]], Maybe Bool)
 missingCases env mn uncovered ca = missingCasesMultiple env mn uncovered (caseAlternativeBinders ca)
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -223,22 +229,20 @@ missingAlternative env mn ca uncovered
...
 -- |
--- Main exhaustivity checking function
+-- Main exhaustivity checking function.
 -- Starting with the set `uncovered = { _ }` (nothing covered, one `_` for each function argument),
 -- it partitions that set with the new uncovered cases, until it consumes the whole set of clauses.
 -- Then, returns the uncovered set of case alternatives.
 -- 
 checkExhaustive :: forall m. (MonadWriter MultipleErrors m) => Environment -> ModuleName -> [CaseAlternative] -> m ()
-checkExhaustive env mn cas = makeResult . first nub $ foldl' step ([initial], (pure True, [])) cas
+checkExhaustive env mn cas = makeResult . first (nub . take maxCases) $ foldr step ([initial], (pure True, [])) cas
```
paf31: We need to bail out early if there are too many cases, and at each step. We need to build an example which causes the current implementation to blow up. Do you have one?

paf31: Won't `foldr` process the cases in the wrong order?

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -223,22 +229,20 @@ missingAlternative env mn ca uncovered
...
   where
-  step :: ([[Binder]], (Maybe Bool, [[Binder]])) -> CaseAlternative -> ([[Binder]], (Maybe Bool, [[Binder]]))
-  step (uncovered, (nec, redundant)) ca =
+  step :: CaseAlternative -> ([[Binder]], (Maybe Bool, [[Binder]])) -> ([[Binder]], (Maybe Bool, [[Binder]]))
+  step ca (uncovered, (nec, redundant)) =
     let (missed, pr) = unzip (map (missingAlternative env mn ca) uncovered)
-        cond = or <$> sequenceA pr
-    in (concat missed, (liftA2 (&&) cond nec,
+        cond = or <$> sequence pr
+    in (if fromMaybe True cond then concat missed else uncovered, (liftA2 (&&) cond nec,
```
paf31: Can you explain this change please?

nicodelpiano: Yes, actually, it's something I came up when writing the library. When we have that `cond` is `Just False`, we know that the case doesn't cover anything. So, it doesn't generate new uncovered cases, i.e., we maintain the previous `uncovered` list.

paf31: Wouldn't they be the same then?

nicodelpiano: No, it seems like it generates another uncovered case that can be removed:

```purescript
data Y = Y | N
f Y Y = Y
f Y Y = Y
```
```purescript
Main.N _
Main.N Main.N   -- this is the unnecessary case generated
_ Main.N
```
I think we can't do this check this inside `missingCasesSingle`.

src/Language/PureScript/Linter/Exhaustive.hs
paf31: @nicodelpiano ping.

nicodelpiano: Ups, sorry about that. Yes, you're right. I've changed it to `foldl'`.
I also have an example that blows this implementation up, I just write it here:

```purescript
module Main where
data ABC = A | B | C | D | E | F | G | H | I | J | K | L | M | N | O | P | Q | R | S | T | U | V | W | X | Y | Z | AA | BB | CC | DD | EE | FF | GG | HH | II | JJ | KK | LL | MM | NN | OO | PP | QQ | RR | SS | TT | UU | VV | WW | XX | YY | ZZ

f AA _ B _ C A = A
f B _ BB _ C X = A
f _ AA X X Z ZZ = A
```

I'm working on this right now. I'll let you know how it goes.

paf31: Since we've hit the pencils down date for GSOC, don't worry about spending too much time on this. I'd rather spend the remaining time implementing tests for your `ecl` library and getting it published, as well as a wrap-up blog post.

nicodelpiano: Sounds good! :+1:

PR #1320, opened 2015-08-01
psc-publish: Try "bower.cmd" if "bower" is not found

hdgarrood:
Fixes #1317. I haven't tested this, though, as I don't have a Windows machine to hand.

PR #1382, opened 2015-08-14
Tests could be run in random order

phadej:


paf31:
What does this gain us? I'm not sure I follow what's going on here.

phadej:
I tried whether stackage could build tests (i.e. tests pass when package fetched from hackage). Seems they do, except for some reason it run tests in `tests-psci` `tests` order, so the first one failed.

.travis.yml
```diff
@@ -1,14 +1,31 @@
...
-  - GHCVER=7.6.3
-  - GHCVER=7.10.1
-  - GHCVER=7.8.4 STACKAGE=lts-2.22
-  # 7.10.2 Pending https://github.com/travis-ci/apt-package-whitelist/pull/686
-  - GHCVER=7.10.1 STACKAGE=lts-3.0
-  - GHCVER=7.10.1 STACKAGE=nightly-2015-08-12
+matrix:
+  include:
+    - env: GHCVER=7.8.4 COVERAGE_SUITE=tests
+      compiler: ": #GHC 7.8.4 - tests"
```
phadej: compiler hack creates distinct caches for each entry.

phadej:
Ok, now I'm done. Seems to work, and caching is more robust.

I'll make yet another PR to separate test running from `sdist` package in separate job.

Sorry for a lot of travis tweaking and PR noise, but I just [nerdsnipped](https://xkcd.com/356/) myself into the CI stuff, and purescript is complex enough project to be interesting to tweak.

garyb:
This seems to break on Windows due to `tests.exe: npm: callProcess: does not exist` - we had a similar issue in `psc-publish`, I think on `mingw32` you'll need to look for `npm.cmd` for it to work. (See #1320)

phadej:
I guess this doesn't change windows behaviour, i.e. it doesn't work in master either. I unfortunately don't have windows machine available to me right now to test on.

garyb:
Yeah, it doesn't work in master for a different reason currently (it fails in a different way I mean), that's why I tried this out in hopes of having it work. :smile:

If you can add the #1320 style executable finding I'd be more than happy to test it again.

phadej:
Yeah, tests already search for `node` and `nodejs`.

garyb:
Those are fine as it is `node.exe` on Windows (which works with the normal method of finding executables). It's specifically `npm` and things installed by `npm` that need to have the `.cmd` extension - the failure was due to some of the `psci` tests, so it might be there's a different problem entirely.

tests/common/TestsSetup.hs
```diff
@@ -0,0 +1,42 @@
...
+import Control.Monad
+
+import Control.Applicative
+
+import Control.Monad.Trans.Maybe
+
+import System.Process
+import System.Directory
+
+findNodeProcess :: IO (Maybe String)
```
phadej: This could be generalised and moved somewhere into shared code, @garyb opinions? seems quite the same as in `psc-publish`

phadej:
I meant something like
```hs
findFirstExecutable :: [String] -> IO (Maybe String)
findFirstExecutble = ...

findNode = findFirstExecutable ["node", "nodejs"]
findBower = findFirstExecutable ["bower", "bower.cmd"]
```

IMHO it's not bad to look for `bower.cmd` on unix systems too.

garyb:
That would make sense. I don't feel strongly about whether we should always search for `bower.cmd` as a fallback or not, but @michaelficarra was against it.

tests/common/TestsSetup.hs
```diff
@@ -0,0 +1,42 @@
+-----------------------------------------------------------------------------
+--
+-- Module      :  Main
+-- Copyright   :  (c) Oleg Grenrus 2015
```
paf31: For code which is licensed with a different copyright ascription, we've started requiring the license to be explicitly listed or linked here, to make it clear. Could you add a link to the MIT license if that's the one you're using?

Thanks.

garyb: For example: https://github.com/purescript/purescript/blob/0ec9d41cfbaf5fde354ad4b3bc12bfa117d8df0b/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs#L4-L5

phadej:
@garyb, Do I need to do a windows fix in this branch, or can this be merged? I'll be a bit more busy this week unfortunately.

garyb:
I'm happy for it to go in as-is since it's not newly breaking anything anyway. I can fix the Windows specific stuff afterwards. :)

tests/common/TestsSetup.hs
paf31: Actually, I just noticed this isn't new code, it's just been extracted verbatim. Since this code isn't written by me (or @garyb), I'm not at liberty to transfer copyright of it. If it's going to be relicensed back to the project under different copyright assumptions, it needs to be done under the terms of the original license, which means including that here. I think I'm going to put something in `CONTRIBUTING.md` about how copyright notices should be added, but I think at a minimum, the code must be new. IANAL though, etc. I hate to be pedantic about these things, but as I say, this isn't my code.

paf31: I've written up the new approach, which is hopefully as simple as possible, here:
https://github.com/purescript/purescript/issues/1394

phadej: done

phadej:
ping @paf31 ?

paf31:
I need to understand the code changes still, but first, slightly less pleasant business:

Copyright notices: New files which are entirely original can have copyright notices added for their author, but the file you added is largely copied verbatim from elsewhere. If it were my code, I probably wouldn't care, but none of that code was written by me, so the copyright belongs with the original authors and is licensed to the project users. Claiming copyright over the whole file would be incorrect. So there are two options: a) leave the copyright notice as it was, with copyright assigned to your code implicitly by `CONTRIBUTORS.md`, or b) relicense the existing code under the terms of the MIT license (or a compatible license) using the existing code, as a _derived work_.

Sorry to make a deal about it. Perhaps the original authors don't care about this honestly, but it's the first time I've had to think about things like this, and I feel it's important to be crystal clear about code ownership in a large(ish) code base.

phadej:
The file functions are copied  from are `Copyright   :  (c) Phil Freeman 2013`, so that file could have copyright by you.

For me (as a conributor) would be easier to have `Copyright : (c) PureScript compiler project contributors`, after all now we have `CONTRIBUTORS.md` and vcs history to track who did what (and when).

paf31:
Yes, I plan to remove those headers once all the contributors have their entries in `CONTRIBUTORS.md`. You could leave it off entirely if you like. The `LICENSE` file would cover it.

phadej:
Copyright notice removed

.travis.yml
```diff
@@ -1,14 +1,32 @@
...
+      # ^ HACK before https://github.com/travis-ci/travis-ci/issues/4393 is resolved
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.8.4,happy-1.19.5,alex-3.1.4], sources: [hvr-ghc]}}
+    - env: GHCVER=7.8.4 COVERAGE_SUITE=tests
```
paf31: Should be `psci-tests`?

phadej: fixed

.travis.yml
```diff
@@ -23,13 +41,20 @@ install:
   # cabal and ghc tries to use them all. Which is bad idea on a shared box.
   # See also: https://ghc.haskell.org/trac/ghc/ticket/9221
   - sed -i 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config
-  - mkdir -p sandboxes/$GHCVER/${STACKAGE:-none}
-  - cabal sandbox init --sandbox sandboxes/$GHCVER/${STACKAGE:-none}
-  # Download stackage cabal.config
-  # Not sure whether filtering out purescript is necessary
+  # Cache sandboxes in ~/cabal-sandboxes
+  # Move right sandbox to .cabal-sandbox if exists
+  - if [ -d ~/cabal-sandboxes/$GHCVER-${STACKAGE:-none} ]; then
```
paf31: Won't this share the sandbox between the main build and the `sdist` test? @hdgarrood Is that ok?

phadej: @paf31 no, the `compiler` trick in build matrix does the job. In https://github.com/travis-ci/travis-ci/issues/4393 is an issue, and this hackish - workaround mentioned.

paf31: Which sandbox does `test-install.sh` use then? It says `cabal sandbox init --sandbox ../.cabal-sandbox` below.

phadej: The same, but the sandbox is snapshotted away (to the cache location) before we install `purescript` into it. Thus we don't need to install dependencies twice, ever.

.travis.yml
```diff
@@ -23,13 +41,20 @@ install:
...
+    fi
+  - mkdir -p .cabal-sandbox
+  - cabal sandbox init --sandbox .cabal-sandbox
+  # Download stackage cabal.config, not sure whether filtering is necessary
   - if [ -n "$STACKAGE" ]; then curl http://www.stackage.org/$STACKAGE/cabal.config | grep -v purescript > cabal.config; fi
   - cabal install --only-dependencies --enable-tests
   - cabal install hpc-coveralls
+  # Snapshot state of the sandbox now, so we don't need to make new one for test install
+  - rm -rf ~/cabal-sandboxes/$GHCVER-${STACKAGE:-none}
+  - cp -r .cabal-sandbox ~/cabal-sandboxes/$GHCVER-${STACKAGE:-none}
```
phadej: @paf31 here we move sandbox contents to the cached location

.travis.yml
paf31: `test-install` is there to verify that everything that needs to be exported is listed in the `.cabal` file, so I think this is fine. :+1:

travis/test-install.sh
```diff
@@ -1,13 +1,18 @@
...
-  mkdir -p sandboxes/$GHCVER/${STACKAGE:-none}
-  cabal sandbox init --sandbox sandboxes/$GHCVER/${STACKAGE:-none}
-  cabal install -j2 --ghc-options='+RTS -A32m -RTS' "$SRC_TGZ"
+  if [ "$RUNSDISTTESTS" = "YES" ]; then
+    mkdir test-install
+    cd test-install
+    tar --strip-components=1 -xzf $SRC_TGZ
+    cabal sandbox init --sandbox ../.cabal-sandbox
+    cabal install --enable-tests --force-reinstalls
+    cabal test
```
paf31: Do we need `cabal test` here?

travis/test-install.sh
```diff
@@ -1,13 +1,18 @@
...
+# the generated source distribution can be installed and tested
 cabal sdist
 if SRC_TGZ="$PWD/dist/$(cabal info . | awk '{print $2;exit}').tar.gz"
 then
-  mkdir -p ../install-test
-  cd ../install-test
-  mkdir -p sandboxes/$GHCVER/${STACKAGE:-none}
-  cabal sandbox init --sandbox sandboxes/$GHCVER/${STACKAGE:-none}
-  cabal install -j2 --ghc-options='+RTS -A32m -RTS' "$SRC_TGZ"
+  if [ "$RUNSDISTTESTS" = "YES" ]; then
```
paf31: What's the rationale for testing again here?

phadej: So we can run tests successfully even when we fetch purescript from hackage, e.g. stackage test runner does that for us nightly (or would do soon)

phadej: And as previously mentioned: it makes one job longer, but I'll separate this later: have normal job, but which runs from unzipped sdist tarball, not from git repository directly.

garyb:
What's happening with this? I'd like to get it in so I can update and merge #1402 :)

Although since this already needs rebasing, perhaps I should merge that anyway?

phadej:
I'll rebase

phadej:
Rebased, waiting only for non-cache lts-3.1 build to complete (cache works!)

paf31:
Yes, I'm going to merge this before I release 0.7.4.

paf31:
Thank you very much!

Issue #1380, opened 2015-08-13
derive Generic for phantom types


zudov:
Compiling the following code
```purescript
data Void
derive instance genericVoid :: GenericVoid
```
Throws
```
psc: Prelude.head: empty list
```

The derived instance should probably look like
```purescript
toSignature _ = SigProd []
fromSpine _ = Nothing
toSpine = absurd
```

zudov:
Defining Void as
```purescript
newtype Void = Void Void
```
doesn't cause that problem (obviously because the constructor list isn't empty anymore :) )

paf31:
@nicodelpiano Would you mind looking into this? I think it's in the exhaustivity checker.

paf31:
You can get the length of the list of arguments from the `Case` statement and pass it as a function argument.

nicodelpiano:
@paf31 Sure! I'm working on this.

PR #1384, opened 2015-08-14
Pass the length of arguments in as a function argument.

nicodelpiano:
This solves the issue #1380.

Now the error (`head empty list`) doesn't appear:

```purescript
module Main where

import Prelude
import Data.Generic

data Void
derive instance genericVoid :: Generic Void
```

and this example works fine.

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -228,8 +228,8 @@ missingAlternative env mn ca uncovered
 -- it partitions that set with the new uncovered cases, until it consumes the whole set of clauses.
 -- Then, returns the uncovered set of case alternatives.
 -- 
-checkExhaustive :: forall m. (MonadWriter MultipleErrors m) => Environment -> ModuleName -> [CaseAlternative] -> m ()
-checkExhaustive env mn cas = makeResult . first nub $ foldl' step ([initial], (pure True, [])) cas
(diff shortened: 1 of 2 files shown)


paf31:
Could you please add that example in `core-tests/tests/generic-deriving`?

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -274,7 +272,9 @@ checkExhaustiveDecls env mn ds =
   in mapM_ f' ds
   where
   checkExpr :: Expr -> m Expr
-  checkExpr c@(Case _ cas)  = checkExhaustive env mn cas >> return c
+  checkExpr c@(Case _ cas)  = checkExhaustive env mn na cas >> return c
+    where
+    na = length . caseAlternativeBinders . head $ cas
```
paf31: I don't see how this solves the issue. You're still calling `head` on `cas`.

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -242,8 +242,6 @@ checkExhaustive env mn cas = makeResult . first nub $ foldl' step ([initial], (p
 
   initial :: [Binder]
   initial = initialize numArgs
```
paf31: Might as well get rid of this now, and `numArgs` is in scope in `initialize` anyway.

paf31: If you can just remove this, and the function argument from `initialize`, I'll merge this. Thanks!

PR #1388, opened 2015-08-17
Include position info in RedefinedModule errors

### LINKED ISSUE #1024: Better "module defined multiple times" error
from IRC - for the "module X has been defined multiple times" error, it would be nice to show every filename that defines a module X. Currently, if you get that error, you have to use grep or something and find it yourself, which is not catastrophic but also not the nicest user experience.

garyb:
Resolves #1024.

I'm going to update the copyright thing for consistency with the cabal file/MIT license link whenever I touch a file from now on, we'll get there eventually :wink:

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -1,8 +1,8 @@
 -----------------------------------------------------------------------------
 --
 -- Module      :  Language.PureScript.Parser.Declarations
--- Copyright   :  (c) Phil Freeman 2013
--- License     :  MIT
+-- Copyright   :  (c) 2013-15 Phil Freeman, (c) 2014-15 Gary Burgess
+-- License     :  MIT (http://opensource.org/licenses/MIT)
 --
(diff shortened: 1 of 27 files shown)


src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -277,19 +277,24 @@ parseLocalDeclaration = positioned (P.choice
 parseModule :: TokenParser Module
 parseModule = do
   comments <- C.readComments
+  start <- P.getPosition
```
paf31: Don't we have a combinator for this somewhere?

paf31:
:+1: Looks great!

src/Language/PureScript/Parser/Declarations.hs
garyb: We do, but it doesn't quite work for this case, it's for `PositionedX` style constructors. I considered doing something like:

``` haskell
withSourceSpan Module $ do
  ...
  return ($ name decls exports)
```

But that doesn't work either without changing `withSourceSpan`.

PR #1389, opened 2015-08-17
Merge multiple imports of the same module

### LINKED ISSUE #817: Importing the same module with and without qualification causes issues

### LINKED ISSUE #1112: Change error for when module is imported twice
I recently did this:

```purescript
import Data.Date (Now(), now)
import Data.Maybe (maybe, Maybe(..))
import Data.Array (filter, modifyAt, (!!))
import Data.Function (on)
import Data.Date (Date(), toEpochMilliseconds)
```

And kept getting a strange error:

```
  Unknown type Now
```

Took me a long time to figure out that I had `Data.Date` imports twice, and I only figured it out as quickly as I did because I've run into it before. It would be more useful to error:

```
  Data.Date module has more than one unqualified import statement.
```

garyb:
Resolves #817, resolves #1112.

#1112 suggests changing the error, but this removes the need for that as having multiple imports for a module is resolved properly now. It could perhaps do with a warning however, I'll open separate issues for that.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -454,7 +454,7 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage <$> onTypesInError
     goSimple (ConflictingImport nm mn) =
       line $ "Cannot declare " ++ show nm ++ " since another declaration of that name was imported from " ++ show mn
     goSimple (ConflictingImports nm m1 m2) =
-      line $ "Conflicting imports for " ++ show nm ++ " from modules " ++ show m1 ++ " and " ++ show m2
+      line $ "Conflicting imports for " ++ nm ++ " from modules " ++ show m1 ++ " and " ++ show m2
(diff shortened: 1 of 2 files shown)


hdgarrood:
I'm not sure if we need a warning for multiple imports. I often want to do something like:

```
import Data.Map (Map)
import qualified Data.Map as M
```

It might also be nice to be able to accomplish this with a single import statement, come to think of it.

garyb:
Worry not! #1390 

Being able to do that is one of the reasons I really wanted this fixed.

hdgarrood:
Hooray :)

garyb:
Argh, looks like there was a duplicate import in `transformers` which now isn't handled correctly ("Conflicting imports for "strMsg" from modules Control.Monad.Error and Control.Monad.Error").

paf31:
:+1:

PR #1393, opened 2015-08-17
Improve import/export errors for unknown values

### LINKED ISSUE #1232: Wording change for unknown explicit import
For example:

```
module Main where

import Prelude (thisDoesNotExist)
```
currently gives "Unknown value Prelude.thisDoesNotExist". I think "Prelude does not export thisDoesNotExist" would be better, in order to help differentiate it from, eg, errors arising from values not being in scope in expressions.

garyb:
Resolves #1232


paf31:
:+1: Great!

Issue #1394, opened 2015-08-17
Clarify licensing terms from all contributors


paf31:
To all existing contributors,

Thank you very much for contributing to the PureScript compiler code.

There have been some cases recently where we have had to think quite a bit about the way code in the PureScript compiler codebase is licensed by its various contributors. Now seems like a good time to clarify things.

I have never wanted to impose a Contributor License Agreement to force transfer of copyright. As I see it, each piece of code is licensed to the project owners by its contributor under the terms of the MIT license (unless specified otherwise), and the MIT license is enough. Instead of a CLA, I'd like to ask you to take five minutes to just clarify the situation regarding any code you may have contributed in the past (if you have not contributed code, please ignore this message):
- I've created a file called [`CONTRIBUTORS.md`](https://github.com/purescript/purescript/blob/master/CONTRIBUTORS.md) in the root of this repository.
- Please add a comment below which includes your name and clarifies how your code is (and optionally, will in future) be licensed to the project. I will add your comment to the `CONTRIBUTORS.md` file (this saves having too many merge conflicts and email notifications).
- The `CONTRIBUTORS.md` file will be append-only, so you can always change your licensing terms for future contributions by _appending_ new or modified terms to the file.
- Here are some templates which you might like to use:
  - "My existing contributions and all future contributions until further notice are Copyright <your name>, and are licensed to the owners and users of the PureScript compiler project under the terms of the [MIT license](http://opensource.org/licenses/MIT)."
  - "I assign copyright of my contributions to date (mm/dd/yyyy) to the owners of the PureScript compiler project."

Hopefully this is as unimposing and lightweight as possible. It accomplishes the following things:
- Individual developers keep ownership of their contributions.
- We have a framework for accepting new contributions under different sorts of license going forward.

Going forward, we'll require updates to this file for new contributors.

paf31:
Hmm apparently you can only mention so many people in a comment.

/cc @pseudonom @puffnfresh @robdaemon @RossMeikleham @sebastiaanvisser @sinelaw @sztupi @taku0 @timmytofu @trofi @utkarshkukreti @vendethiel @vkorablin @zudov

michaelficarra:
Doing this through PRs like this is going to create a *lot* of notifications (65 contributors * 101 watchers). @paf31 Maybe contributors can give their permission here to have you add them directly using one of the proposed templates?

paf31:
Ah right, good idea, I didn't think about that. Your approach would also solve the merge problem.

[...]

PR #1395, opened 2015-08-17
Pass the length of arguments in as a function argument.

nicodelpiano:
This solves the issue #1380.

Now the error (`head empty list`) doesn't appear:

```purescript
module Main where

import Prelude
import Data.Generic

data Void
derive instance genericVoid :: Generic Void
```

and this example works fine.

### DIFF
--- a/core-tests/tests/generic-deriving/Main.purs
+++ b/core-tests/tests/generic-deriving/Main.purs
@@ -4,11 +4,16 @@ import Prelude
 
 import Data.Generic
 
+data Void
+
+derive instance genericVoid :: Generic Void
+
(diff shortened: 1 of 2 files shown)


paf31:
:+1: Thanks!

nicodelpiano:
:smile:

paf31:
Might as well put the new policy into effect now :smile: 

Could you please add your licensing terms here? https://github.com/purescript/purescript/issues/1394

nicodelpiano:
Done :smile: :+1:

paf31:
Thanks!

PR #1396, opened 2015-08-17
Update CONTRIBUTORS.md

robdaemon:
Adding licensing terms per request.

For the record, this was entirely unimposing and lightweight :)


paf31:
Thanks so much! Could you just fill in the `\<your name\>` template please? :smile:

paf31:
Oh dear, I didn't think an append-only file would be such a pain for merge commits. I'll figure these out :)

robdaemon:
Oops, I missed that "your name" thing - fixed. If you want to change the mechanism here though, that's fine too. Maybe contributors emailing you their response and they can be applied without the merge commits?

paf31:
Yes, I've changed the issue to ask people to just post the markdown in a comment, and I'll copy it manually. I'll copy yours over now.

Thanks!

PR #1398, opened 2015-08-17
Warn on misleading Type(..) imports

### LINKED ISSUE #1391: Warn on importing X(..) when X has no exported data constructors
Applies to importing `type`s with `(..)` also.

garyb:
Resolves #1391. 

@paf31 I can't figure out why the warning doesn't have position info though. It's probably something silly, can you take a look?

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -134,6 +134,7 @@ data SimpleErrorMessage
   | NotExhaustivePattern [[Binder]] Bool
   | OverlappingPattern [[Binder]] Bool
   | ClassOperator ProperName Ident
+  | MisleadingEmptyTypeImport ModuleName ProperName
   deriving (Show)
 
 -- |
@@ -256,6 +257,7 @@ errorCode em = case unwrapErrorMessage em of
   NotExhaustivePattern{} -> "NotExhaustivePattern"
   OverlappingPattern{} -> "OverlappingPattern"
   ClassOperator{} -> "ClassOperator"
+  MisleadingEmptyTypeImport{} -> "MisleadingEmptyTypeImport"
 
 -- |
 -- A stack trace for an error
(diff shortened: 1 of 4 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -845,11 +849,14 @@ rethrow f = flip catchError $ \e -> throwError (f e)
...
-  withPosition (PositionedError _ err) = withPosition err
-  withPosition err = PositionedError pos err
+rethrowWithPosition pos = rethrow (onErrorMessages (withPosition pos))
+
+warnWithPosition :: (MonadWriter MultipleErrors m) => SourceSpan -> m a -> m a
+warnWithPosition pos = censor (onErrorMessages (withPosition pos))
+
+withPosition :: SourceSpan -> ErrorMessage -> ErrorMessage
+withPosition pos (PositionedError _ err) = withPosition pos err
+withPosition pos err = PositionedError pos err
```
garyb: Also, is this right? It's what we had before, but if we're going top down surely it would be better to keep the original position, as it will be more specific.

paf31:
Is it definitely hitting the `PositionedDeclarationRef` case?

Looks good otherwise :+1:

src/Language/PureScript/Errors.hs
paf31: Yeah, I think we should always get the deepest position information we can.

garyb: I'll change it to do that then and see what effect it has, we can always revert it before releasing if it is worse.

garyb:
I guess that would be the sensible explanation for why it's not... perhaps they're getting `map`ped out somewhere. I'll look at it again tomorrow, just thought I'd have a quick go at this one before signing off for the night.

garyb:
:sob:

I'll get the tests working on Windows before attempting to "fix" this any further.

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -529,6 +529,7 @@ resolveImport currentModule importModule exps imps impQual =
   isHidden hidden ref@(TypeRef _ _) =
     let
       checkTypeRef _ True _ = True
+      checkTypeRef r acc (PositionedDeclarationRef _ _ h) = checkTypeRef r acc h
```
garyb: @paf31 this not existing was the cause of my woes, previously `checkedRefs` below was deleting `PositionedDeclarationRef` from the AST which meant this case wasn't needed...

PR #1402, opened 2015-08-18
Fixes running tests on Windows

### LINKED ISSUE #1401: Test suite failure on Windows
This is proving rather tricky. I added a fix so it would use `npm.cmd` as appropriate, but that cmd file relies on a magic variable called `%~dp0` which is supposed to be the directory that the current cmd file resides within, however when it's launched from Haskell, it's not - it's getting the support folder path instead.

That kinda sounds like a bug in `callProcess` to me, so I'm not sure where to go now...

garyb:
Fixes #1401... but maybe should wait for #1382 first?

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -23,6 +23,7 @@ extra-source-files: examples/passing/*.purs
                   , tests/support/setup.js
                   , tests/support/package.json
                   , tests/support/bower.json
+                  , tests/support/setup-win.cmd
                   , psci/tests/data/Sample.purs
 
 source-repository head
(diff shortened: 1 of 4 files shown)


paf31:
:+1:

phadej:
The `tests/support/setup-win.cmd` needs to be in `extra-source-files` (the travis build won't catch windows problems :( ).

garyb:
Good call!

paf31:
@garyb Do you want to merge this before 0.7.4?

garyb:
I'll probably update this before my next PR so I can actually test it, so yeah I guess so!

garyb:
This should be good now, I'll merge once travis completes.

paf31:
:+1:

PR #1405, opened 2015-08-20
Bumped lts-3 resolver

zudov:


paf31:
How does this work? Isn't LTS supposed to not change very often?

zudov:
@paf31 Minor LTS releases only include patch and minor library updates (as by semver). Needed so that libraries could receive important (backported) updates, e.g. security vulnerabilities.

paf31:
:+1: Thanks!

Issue #1244, opened 2015-07-09
Error when explicitly importing value re-exported from module


joneshf:
I wasn't sure if this was known already or not, but I think I ran into a bug:

```purescript
module Foo where

  foo :: Int
  foo = 3

module Bar (module Foo) where

  import Foo

module Baz where

  import Bar (foo) -- this is the problem

  baz :: Int
  baz = foo
```

```
➜  rexport  psc src/Foo.purs 
Error:
Error in module Baz:
Error at /home/joneshf/programming/purs/rexport/src/Foo.purs line 12, column 3 - line 14, column 3:
  Unknown value Bar.foo
See https://github.com/purescript/purescript/wiki/Error-Code-UnknownValue for more information, or to contribute content related to this error.
```

If the problem line is changed to `import Bar`, then things compile fine.

garyb:
I think that's a new one, thanks for the report :disappointed:

themoritz:
Same holds when `hiding` values.

andyarvanitis:
I can have a look at this -- I'm pretty sure I know the fix.

svisser:
I just got this error. Also note that the page at the indicated URL does not exist:

https://github.com/purescript/purescript/wiki/Error-Code-UnknownValue

PR #1406, opened 2015-08-20
Fix #421, match type instance heads eagerly

### LINKED ISSUE #421: Superclass constraints are ignored in instances

paf31:
/cc @garyb 

Still a WIP, but probably near enough to review.


paf31:
Looks like `quickcheck` breaks due to a false positive in the overlapping instance check.

garyb:
Ah, well, aside from it not fully working it looks really nice :wink:

paf31:
It's much simpler now :smile:

garyb:
How are the false positives arising?

paf31:
Not sure yet, but I'll look into it.

paf31:
Hah, it's failing because `monadStateStateT` and `monadStateStateT1` overlap now, i.e. because it should :smile:

paf31:
It just means we need to patch `transformers` after this is released.

garyb:
What on earth is that? :wink:

Apparently needed by `ParserT`, but I assume that will work after this too?

paf31:
I haven't tested `parsers` yet, but I will do now.

paf31:
`parsers` compiles fine with the patch on `transformers`.

paf31:
- [ ] Merge and release https://github.com/purescript/purescript-transformers/pull/49, verify `core-tests`
- [ ] Pick simplest instance when superclasses are involved.

garyb:
Looks great! +:100: :shipit:

garyb:
Although, there are non-transformers test failures in there too, is that because the tests are wrong now?

paf31:
Oops, you're right. Looks like it's matching the `suArray` instance too eagerly in `Superclasses2.purs`. I'll look into it.

paf31:
Hmm, GHC gives

```
/tmp/1.hs:21:13:
    Could not deduce (a ~ [a])
    from the context (Cl a)
      bound by the type signature for test :: Cl a => a -> a
      at /tmp/1.hs:20:9-24
      ‘a’ is a rigid type variable bound by
          the type signature for test :: Cl a => a -> a at /tmp/1.hs:20:9
    Relevant bindings include
      x :: a (bound at /tmp/1.hs:21:6)
      test :: a -> a (bound at /tmp/1.hs:21:1)
    In the first argument of ‘su’, namely ‘[cl x x]’
    In the expression: su [cl x x]
    In an equation for ‘test’: test x = su [cl x x]
```

So maybe we should let that test fail now.

garyb:
Should we turn that into a failing test then, or do you think it's something that could be addressed? I guess if the new behaviour agrees with GHC then perhaps it is right :wink:

paf31:
I think it should probably be a failing test, yeah. I can't get to it today, but hopefully can make this mergeable tomorrow.

garyb:
Oops, merging #1410 means this now conflicts.

garyb:
I'm more than happy for this to go in whenever you're ready, by the way.

paf31:
`core-tests` is breaking, but should work again once `transformers` is merged and released. I need to also release 0.7.4 to make that `transformers` release build, so I just need to coordinate a few things. If you think it's all good to go, I'll merge and release everything this afternoon.

Could you please run your tool and find reverse dependencies for `transformers`?

garyb:
I'm currently working on a fix for #1244, I'd really like to get that in for 0.7.4, hopefully it shouldn't take too much longer.

I'll run off that list once I'm done :)

Also suffering from internet problems today so not actually present in IRC, posting this from my phone...

paf31:
I can figure out the dependent modules, that's ok. I can make a `transformers` release now, but make it pre-release?

paf31:
I get:

- [ ] `free`
- [ ] `parallel`
- [ ] `quickcheck`
- [ ] `parsing`

garyb:
- [ ] purescript-quickcheck
- [ ] purescript-maps
- [ ] purescript-catenable-lists
- [ ] purescript-free
- [ ] purescript-strongcheck
- [ ] purescript-sets
- [ ] purescript-graphs

paf31:
`maps`, really?

paf31:
Oh `devDependencies` though.

garyb:
Yeah, via `quickcheck` I assume

paf31:
No need for a version bump on those then, but we can update the dependency version. It's not going to break any code to pull in `maps`, since `transformers` won't be pulled in too, right?

Issue #1297, opened 2015-07-23
psc consumes enormous amount of RAM


cryogenian:
During update of slamdata I've noticed that builds take lot more memory. 0.6.9.7 version used to consume about 250Mb in `psc` mode and 150Mb in `psc-make` mode. Now `psc` consumes more then 4Gb :disappointed:

jdegoes:
With this issue, we cannot get our automated builds to pass, because Travis CI instances do not have a spare 4GB of memory. I'm surprised it's not affecting other OSS projects yet.

@puffnfresh How can we bribe you to take a look at this? :wink:

puffnfresh:
It's tricky because the kernel will kill the process and so the GHC runtime won't have a chance to give any profiling information. I do have some rough info from compiling a small chunk of SlamData, though:

We allocated 24GB, around 1/3 under parseInputs, 1/5 under rebuildIfNecessary and 1/5 under typeCheckModule. Although I can't really get a fair comparison, this looks pretty similar to what we used to do.

The only obvious difference I've seen that I can see is that 15% of allocations come from `compile` itself (i.e. it allocates that much by itself, not the things under it) - a pretty big bump from 0%

puffnfresh:
![master-heap-profile](https://cloud.githubusercontent.com/assets/37715/8873613/992c6070-324b-11e5-9827-a8163bac41b9.png)

jdegoes:
@puffnfresh Wow, thanks, this is great.

`makeBindingGroupVisible` seems to have runaway behavior (as well as OTHER, unfortunately, collection of lots of other things). But overall no single culprit for explaining how the SlamData build went from 250MB to 4 GB. Could still be, e.g., a dependency bump in the compiler, used at a fairly low level and pervasively (e.g. mtl or something). But unless it's something like that, it's going to be a long hard road back to 250 MB.

puffnfresh:
And this is the heap, by type of value:

![master-heap-type-profile](https://cloud.githubusercontent.com/assets/37715/8891894/2f4ec966-337e-11e5-81c3-c39284f8a87f.png)

bitc:
I've added to purescript an alternative compile command, that compiles only a single `.purs` file. This way it can be called from an external build system. I have integrated it with [Shake](http://shakebuild.com/)

This way I get fast incremental builds, and support for parallel builds.

[...]

PR #1410, opened 2015-08-22
Use strict writers to avoid leaking memory

garyb:
Part of #1297, there may be a round 2 yet.

An unscientific test of compiling `SlamData` from scratch and eyeballing the process monitor with these changes applied reduces the peak memory usage from ~12.5gb to ~2.5gb. Also compilation time is reduced from ~68s to ~37s.

:boom::boom::boom:

### DIFF
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -22,7 +22,7 @@ module Main where
 import Control.Applicative
 import Control.Monad
 import Control.Monad.Error.Class (MonadError(..))
-import Control.Monad.Writer
+import Control.Monad.Writer.Strict
 
 import Data.Version (showVersion)
 import qualified Data.Map as M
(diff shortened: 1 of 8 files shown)


garyb:
Alright, culprit 2 was 6d3a777baeb10040b642787a465af4e37ec67294, so this probably is as good as it gets as it makes that `Writer` strict too.

We're still using 10x as much as memory as the pre-warnings version though, I'm sure there's room for some other improvements.

paf31:
:+1: Amazing.

jdegoes:
:+1: :+1: :fireworks: :boom: :fireworks: :+1: :+1:

jdegoes:
10x down, 10x more to go. :wink:

garyb:
I hope so! :wink: I think SlamData's Travis should start working again at least, even with just this.

Issue #1385, opened 2015-08-15
Use -Werror, and turn off unnecessary import warnings


paf31:
Do we want to do this to allow us to identify PRs which introduce warnings?

hdgarrood:
This sounds good to me. Although this might introduce problems where dependencies produce warnings. Perhaps we need a way of saying "-Werror, but only for stuff in `src/`".

Also, why turn off unnecessary import warnings? Better to fix those, surely?

paf31:
Ah, because we'd need `CPP` to make import warnings go away in 7.10 I think?

hdgarrood:
Ohhh, sorry, I thought you were talking about purescript core library code. Then :+1:

erdeszt:
@paf31 What's the problem with using ```CPP```? I can do the PR unless you have something against it.

paf31:
I just always thought it was a bit hacky, but maybe it's necessary. Yeah, we could do that. If you're willing to make the PR, that'd be great, thanks!

paf31:
Duplicate.

PR #1411, opened 2015-08-23
Fix warnings and enable -Werror

erdeszt:
Related issue: https://github.com/purescript/purescript/issues/1385

I'm not sure if the last commit is necessary, but most of the places used the one pragma per line convention so I thought I unify it.

### DIFF
--- a/psci/Completion.hs
+++ b/psci/Completion.hs
@@ -1,13 +1,19 @@
+{-# LANGUAGE CPP #-}
+
 module Completion where
 
 import Data.Maybe (mapMaybe)
 import Data.List (nub, nubBy, sortBy, isPrefixOf, stripPrefix)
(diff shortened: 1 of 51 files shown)


paf31:
Looks good to me, thanks! @garyb Any comments before I merge?

garyb:
Nope! Aside from being glad we'll never have to do this in PureScript :smile:

hdgarrood:
I don't think specifying `-Werror` in the cabal file is a good idea, as if any warnings are produced when someone tries to `cabal install purescript`, it will fail. I think Hackage will reject packages with `-Werror` now, actually.

I think changing these back and putting `-Werror` into the Travis script would accomplish the same goal.

paf31:
Ok sounds good.

PR #1412, opened 2015-08-23
Fix failing test

erdeszt:
I've just noticed that my changes broke the travis build(which is a bit strange because it was green while it was a PR). This PR (hopefully) fixes the build.


paf31:
:+1: Thanks!

PR #1418, opened 2015-08-24
Better re-export handling

### LINKED ISSUE #291: Allow modules to re-export imported values
So, for example, `Control.Monad.Writer` can also export the functions from `Control.Monad.Writer.Class`.

``` haskell
module Control.Monad.Writer
  ( Writer()
  , runWriter
  , writer
  , listen
  , pass
  , tell
  , listens
  , censor
  ) where

import Prelude
import Control.Monad.Identity
import Control.Monad.Writer.Trans
import Control.Monad.Writer.Class
import Data.Monoid
import Data.Tuple

type Writer w a = WriterT w Identity a

runWriter :: forall w a. Writer w a -> Tuple a w
runWriter = runIdentity <<< runWriterT
```

### LINKED ISSUE #1244: Error when explicitly importing value re-exported from module

garyb:
Resolves #1244. Resolves #291 in a slightly different way too (if `A` explicitly imports values from `B` and then re-exports `B`, only the values it imported will be re-exported).


garyb:
All green at last!

paf31:
:+1: Thanks!

_Edit_: Actually, after I `cabal install`ed this, `transformers` test suite no longer builds - it complains about things not being in scope, but they should be because of module reexports. Not sure what's going on, but I think we should figure it out before releasing the compiler.

garyb:
They run okay for me now, did you make those changes for v0.7.0-rc.2 after posting this?

The changes you made look right to me - you can't export a module anymore unless you import it, the reasoning for this being only the specific values imported into the current module are re-exported when you export a module.

paf31:
The tests run ok? I get `Unknown value callCC` since `callCC` is imported via a module reexport in the tests. This is with what's on `master`.

garyb:
Yep. Tried cleaning the output folder?

garyb:
Yup, definitely works for me - I just installed the `master` version of `psc` on my Mac, and with a fresh checkout of `transformers` the tests run fine.

paf31:
Yes, I just removed `output` and I get the same error. I'll reinstall `psc` to be sure.

paf31:
Yes, I don't see module reexports in the externs files. Is that an issue?

paf31:
Seems like it would be an issue if we want to support separate compilation without looking at original source files.

garyb:
#1425 should sort that out.

PR #1420, opened 2015-08-24
Move -Werror to travis config

erdeszt:
As suggested in #1411 moving the -Werror to the travis config from the cabal file.
Not sure though that this is the right way of doing it, I've never used travis before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ install:
   - cp -r .cabal-sandbox ~/cabal-sandboxes/$GHCVER-${STACKAGE:-none}
 script:
   - ./travis/configure.sh
-  - cabal build
+  - cabal build --ghc-options="-Werror"
   - cabal test
   - ./travis/test-install.sh
 after_script:
--- a/psci/tests/Main.hs
+++ b/psci/tests/Main.hs
@@ -1,11 +1,14 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE CPP #-}
 
 module Main where
 
 import Control.Monad.Trans.State.Strict (runStateT)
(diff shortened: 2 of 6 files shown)


paf31:
:+1: Thanks. Let's see what Travis does this time.

garyb:
This'll conflict with my changes I think, as I removed `-Werror` also. I guess what we really want is to only run `-Werror` on the newer GHCs though, right? As otherwise this will suffer from the same problem I had.

erdeszt:
@garyb I've fixed the only non 7.10 related warning in #1411 so there should be no warnings on earlier versions either(and the travis build for that branch was green for the old versions too, let's see how this one turns out).

hdgarrood:
Why only run `-Werror` on the newer GHCs? I think I'd rather have it so that there are no warnings on any of the GHCs that we test with.

Also, sorry I didn't think to mention this earlier, but perhaps [base-compat](http://hackage.haskell.org/package/base-compat) might be a bit nicer than CPP all over the place? I think it would also make it easier to deal with any changes in future `base` releases, if any situations similar to the Foldable/Traversable Proposal arise in the future.

paf31:
Both of those sound like good ideas to me.

garyb:
> Why only run `-Werror` on the newer GHCs? I think I'd rather have it so that there are no warnings on any of the GHCs that we test with.

I've already discovered how painful it is to try and track down warnings by waiting 5-10 minutes for Travis to fail so an import can be tweaked... several times :) It wouldn't be so bad if you got all the errors at once.

paf31:
Wouldn't `base-compat` help with that though?

garyb:
Looks like it, yep!

paf31:
This needs to be merged with `master` now, sorry.

paf31:
Thanks!

Issue #1375, opened 2015-08-13
psc: No files found using pattern: src/**/*.js


garyb:
After #1176 this gets printed when using `pulp` and your project has no FFI files in its main source tree. Since I would hope that's most projects, should we do something about it?

I guess it is useful to know that the glob doesn't match if you think it should, but maybe we can hide it for the case of FFI files as if they're missing the compiler will raise a different error anyway. Although, that error sometimes happens because of the missing or badly formatted comment, so having this extra error in might be helpful there... argh, I don't know. Thoughts?

mjgpy3:
>  but maybe we can hide it for the case of FFI files

I think that when I do #1268, I'm going to turn FFI warnings off, especially if inferring the `.js` becomes the standard way of including FFIs.

mjgpy3:
Wait, nevermind... I've got to think about this some more.

paf31:
I think we can close this now since the `src/**/*` wildcard works instead.

mjgpy3:
:+1:

PR #1423, opened 2015-08-25
Resolves #1268. JavaScript (FFI) files can be specified without `--ffi`.

### LINKED ISSUE #1268: Easier FFI inclusion for `psc` command-line

mjgpy3:
The `--ffi` flag still works, however.

I didn't really address #1375 (everything still warns on glob misses) as I'm not really sure how to best handle the concerns therein.

### DIFF
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -24,6 +24,7 @@ import Control.Monad
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Writer.Strict
 
+import Data.List (isSuffixOf, partition)
 import Data.Version (showVersion)
 import qualified Data.Map as M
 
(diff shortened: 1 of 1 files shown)


paf31:
I'll leave this one until after today's 0.7.4 release, but it looks good.

mjgpy3:
Cool, thanks!

paf31:
Thanks!

PR #1428, opened 2015-08-26
Fix function composition inlining

### LINKED ISSUE #1427: Function composition inlining is not triggering
Minimal example:

```purescript
module Example where

import Prelude

f :: forall a b c. (b -> c) -> (a -> b) -> (a -> c)
f g h = g <<< h
```

```javascript
var Prelude = require("Prelude");
var f = function (g) {
    return function (h) {
        return Prelude["<<<"](Prelude.semigroupoidFn)(g)(h);
    };
};
```

garyb:
Resolves #1427


phadej:
Should we have a test for this?

garyb:
I don't think we have a good way of testing any of the inlining operations currently. We'd need a "proper" test suite rather than running compiled code.

paf31:
:+1:

puffnfresh:
@garyb I had a feeling that was the problem. Thanks!

PR #1436, opened 2015-08-29
foreign -> foreign_

zudov:
`foreign` is a keyword and it's better not to use it as a name to avoid parser errors.

Interestingly it was causing an error only when in ghci (7.10.2).


paf31:
Thanks!

PR #1437, opened 2015-08-29
Do not allow empty case statement

### LINKED ISSUE #1435: Case parse bug
```purescript
error err = case err of \_ -> 1
```

parses but probably shouldn't.

zudov:
Fixes #1435

### DIFF
--- a/examples/failing/EmptyCase.purs
+++ b/examples/failing/EmptyCase.purs
@@ -1,0 +1,4 @@
+-- @shouldFailWith ErrorParsingModule
+module Main where
+
+error err = case err of \_ -> 1
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -366,7 +366,7 @@ parseConstructor = Constructor <$> C.parseQualified C.properName
 
 parseCase :: TokenParser Expr
 parseCase = Case <$> P.between (P.try (reserved "case")) (C.indented *> reserved "of") (return <$> parseValue)
-                 <*> (C.indented *> C.mark (P.many (C.same *> C.mark parseCaseAlternative)))
+                 <*> (C.indented *> C.mark (P.many1 (C.same *> C.mark parseCaseAlternative)))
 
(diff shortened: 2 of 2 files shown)


garyb:
:+1:

phadej:
What about `Void`?

garyb:
You could still write `case v of _ -> ...`. I don't think there's any good explanation of what a case with no patterns represents.

zudov:
@phadej If I understand correctly, doing
```purescript
foo :: Void -> _
foo void = case void of _ -> ..
```
is useful only in a lazy setting, as a way to force the void and make the hack which was used to create it reveal itself (e.g. throw `error: undefined` if Void was created with `undefined :: Void`).

In strict setting `absurd` should be as good. Correct me if I'm wrong.

Although, this only applies to the actual `Void` for which we have absurd.

zudov:
By the way, it now makes sense to use NonEmpty for case alternatives, I'll take a look a bit later how well-suited that is.

phadej:
@zudov check `-XEmptyCase` https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/syntax-extns.html

```hs
case onlyVoid of _ -> ..
```
has *redundant* case expression

I cannot come with real world example where writing like that would be useful, but I can imagine some type level trickery (not only `Void` is non-inhabitant type)

paf31:
:+1:

zudov:
@phadej Yeah, I had `-XEmptyCase` in mind, AFAIK the canonical example are things like
```purescript
instance Show Void where
   show void = case void of {}
instance Eq Void where
   (==) void1 void2 = (case void1 of {}) `seq` (case void2 of {})
```
This would have advantage over `absurd` as it would evaluate `void`.

zudov:
Given [that](https://github.com/purescript/purescript/issues/1435#issuecomment-136021828), it seems like it would be better to live `Case` AST definition as it is (do not change to `Case Expression (NonEmpty CaseAlternative)`.

I guess we can merge it now.

phadej:
:+1:

PR #1450, opened 2015-09-08
Properly encode UTF-8 String into Bytestring

### LINKED ISSUE #1449: psc-publish doesn't allow my name in bower.json
Cause I'm german and it contains the Umlaut ü.

Error in JSON in the output of `bower list --json --offline`:              
  The input could not be parsed as JSON aeson said: Failed reading: Cannot 
  decode byte '\xfc': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8
  stream

zudov:
Fixes #1449

### DIFF
--- a/src/Language/PureScript/Publish.hs
+++ b/src/Language/PureScript/Publish.hs
@@ -23,7 +23,6 @@ import Prelude hiding (userError)
 
 import Data.Maybe
 import Data.Char (isSpace)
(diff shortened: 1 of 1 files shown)


zudov:
I've grepped around a bit trying to find similar bugs, but it doesn't seem like there are any.

paf31:
:+1: Thanks!

PR #1454, opened 2015-09-09
Make some errors simpler. Fix #1387.

### LINKED ISSUE #1387: Odd exhaustivity warnings with arrays
Compiling this:
```
module The where

the :: forall a. Array a -> a
the [x] = x
```

gives this:
```
Warning in value declaration the:
Warning at /home/harry/documents/code/ps-scratch/src/Main.purs line 3, column 1 - line 4, column 1:
    Pattern could not be determined to cover all cases.
    The definition has the following uncovered cases:

    _
```

which I would interpret to mean that the pattern covers no cases at all?

Incidentally, what's the recommended way of writing this function `the` in a way that won't produce warnings? I know it's unsafe and have taken that into account where I've used it.

paf31:
/cc @hdgarrood @garyb


garyb:
Looks good :+1:

hdgarrood:
When describing a case, does redundant = unreachable?

paf31:
Yes I think so. I'll change that.

PR #1455, opened 2015-09-09
Implement @hdgarrood's suggestion for psc-publish

### LINKED ISSUE #1374: psc-publish: Make the --dry-run flag more obvious
My ambition here is to make the situation described by @Fresheyeball [here](https://github.com/Fresheyeball/purescript-context/pull/4#issuecomment-130521502) less painful.

One thing we could do is print a note about the `--dry-run` flag if the problem is simply that no tagged version is checked out. For example:

```
$ psc-publish
published.                                                                   
Details:                                                                     
  psc-publish requires a tagged version to be checked out in order to build  
  documentation, and no suitable tag was found. Please check out a previously
  tagged version, or tag a new version.                                      
                                                                             
  Note: tagged versions must be in one of the following forms:               
    * v{MAJOR}.{MINOR}.{PATCH} (example: "v1.6.2")                           
    * {MAJOR}.{MINOR}.{PATCH} (example: "1.6.2")

  If the version you are publishing is not yet tagged, you might want to use
  the --dry-run flag instead, which removes this requirement. Run
  psc-publish --help for more details.
```

The current message is the same, except that it doesn't have the last paragraph.

paf31:
Fixes #1374. /cc @hdgarrood Does this look ok?

### DIFF
--- a/src/Language/PureScript/Publish/ErrorsWarnings.hs
+++ b/src/Language/PureScript/Publish/ErrorsWarnings.hs
@@ -162,6 +162,12 @@ displayUserError e = case e of
         , para "Note: tagged versions must be in one of the following forms:"
         , indented (para "* v{MAJOR}.{MINOR}.{PATCH} (example: \"v1.6.2\")")
         , indented (para "* {MAJOR}.{MINOR}.{PATCH} (example: \"1.6.2\")")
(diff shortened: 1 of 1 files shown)


hdgarrood:
:+1:

PR #1457, opened 2015-09-11
Support constraints without forall

paf31:
/cc @garyb

### DIFF
--- a/examples/passing/NakedConstraint.purs
+++ b/examples/passing/NakedConstraint.purs
@@ -1,0 +1,12 @@
+module Main where
+
+import Control.Monad.Eff.Console
+
+class Partial where
+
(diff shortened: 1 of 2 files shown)


examples/passing/NakedConstraint.purs
```diff
@@ -0,0 +1,12 @@
+module Main where
+
+import Control.Monad.Eff.Console
+
+class Partial where
+
+data List a = Nil | Cons a (List a)
+
+head :: (Partial) => List Int -> Int
```
phadej: are parentheses still required here?

paf31: Right now, yes.

garyb:
:+1: 

Will be nice to be able to try out `Partial` at last!

PR #1458, opened 2015-09-11
Fix #1042, better messages for ExpectedType errors.

### LINKED ISSUE #1042: Better errors for kind mismatch in typed expressions
For example:
```purescript
module Test where

data List a
  = Nil
  | Cons a (List a)

x = Nil :: List
```

Gives a type error:
```
Error in module Test:
Error in value declaration x:
Error at test.purs line 7, column 5 - line 7, column 12:
  Expected type of kind *, was * -> *
See https://github.com/purescript/purescript/wiki/Error-Code-ExpectedType for more information, or to contribute content related to this error.
```

We might improve this error by adding something like the following:

> In a type-annotated expression `x :: t`, the type `t` must have kind `*`. The error arises from the type  `List` having the kind `* -> *` instead.

We might even be able to go further for types of kind `* -> *`:

> Perhaps you meant `forall a. List a`?

paf31:
/cc @garyb @hdgarrood

### DIFF
--- a/examples/failing/KindStar.purs
+++ b/examples/failing/KindStar.purs
@@ -1,0 +1,8 @@
+-- @shouldFailWith ExpectedType
+
+module X where
+
+data List a = Nil | Cons a (List a)
+
+test :: List
+test = Nil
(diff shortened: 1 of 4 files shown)


puffnfresh:
:+1:

hdgarrood:
:+1:

garyb:
:+1:

PR #1461, opened 2015-09-11
Only use Show for debugging.

paf31:
/cc @garyb @hdgarrood

I've wanted to change this for a while. We were using `Show` to print strings for humans. If we want to switch to `Text`, or use `Show` for things like externs file, this isn't good, and it's also bad for new developers. Now the only place where we use `Show` is in Parsec to print tokens, since Parsec forces it on us.


hdgarrood:
:100: this bothered me too.

paf31:
Build is failing because of core-tests AFAICT.

garyb:
+:100: also!

PR #1462, opened 2015-09-12
Fix #1310

### LINKED ISSUE #1310: Inject seems to do something funny with row unification
This typechecks without the `CONSOLE` effect:

``` purescript
foreign import data Oops :: !

what :: forall eff. Eff (oops :: Oops | eff) Unit
what = inj (Control.Monad.Eff.Console.log "wow")
```

If you try and drop the row and just use `eff` the error is:
```
No instance found for Data.Inject.Inject (Control.Monad.Eff.Eff (console :: Control.Monad.Eff.Console.CONSOLE | _13)) (Control.Monad.Eff.Eff eff8)
```

So perhaps there's something happening with an unknown unifying with the extra effects perhaps? I don't know if that's even possible. I guess it's a problem with instances not types directly though.

paf31:
@garyb This matches the `unifyRows` function now. Could you please review this?

### DIFF
--- a/examples/failing/1310.purs
+++ b/examples/failing/1310.purs
@@ -1,0 +1,18 @@
+-- @shouldFailWith NoInstanceFound
+
+module Issue1310 where
+
+import Prelude
+import Control.Monad.Eff
+import Control.Monad.Eff.Console
+
+class Inject f g where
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

PR #1464, opened 2015-09-13
Typevar warnings

### LINKED ISSUE #889: Warning for unused type variables
From purescript-validation:

```ps
isValid :: forall err result r. V err result -> Boolean
isValid (Valid _) = true
isValid _ = false
```

The `r` in the type signature appears to be unused. Should `psc` warn about this? `ghci` appears to do so:

```
λ: :set -XRankNTypes
λ: :set -Wall
λ: let x = 3 :: forall a. Int
<interactive>:9:21: Warning:
    Unused quantified type variable ‘a’
    In the type ‘forall a. Int’
    In an expression type signature
```

### LINKED ISSUE #1463: Warning for shadowed type variables

garyb:
Resolves #1463, resolves #889.

### DIFF
--- a/src/Language/PureScript/Linter.hs
+++ b/src/Language/PureScript/Linter.hs
@@ -19,7 +19,7 @@
 
 module Language.PureScript.Linter (lint, module L) where
 
-import Data.List (mapAccumL, nub)
+import Data.List (mapAccumL, nub, (\\))
 import Data.Maybe (mapMaybe)
 import Data.Monoid
 
@@ -60,7 +60,7 @@ lint (Module _ _ mn ds _) = censor (onErrorMessages (ErrorInModule mn)) $ mapM_ 
 
         f' :: Declaration -> MultipleErrors
         f' (PositionedDeclaration pos _ dec) = onErrorMessages (PositionedError pos) (f' dec)
-        f' dec = f dec <> checkShadowedTypeVars dec
+        f' dec = f dec <> checkTypeVars dec
 
     in tell (f' d)
     where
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/Linter.hs
```diff
@@ -75,17 +78,42 @@ lint (Module _ _ mn ds _) = censor (onErrorMessages (ErrorInModule mn)) $ mapM_
...
+
+  checkTypeVars :: Type -> MultipleErrors
+  checkTypeVars ty = everythingWithContextOnTypes S.empty mempty mappend step ty <> findUnused ty
+    where
+    step :: S.Set String -> Type -> (S.Set String, MultipleErrors)
+    step s (ForAll tv _ _) = bindVar s tv
+    step s _ = (s, mempty)
+    bindVar :: S.Set String -> String -> (S.Set String, MultipleErrors)
+    bindVar = bind ShadowedTypeVar
+    findUnused :: Type -> MultipleErrors
```
paf31: Is it possible to fuse this traversal with the one above?

garyb: Ah, yep, it should be actually - good thinking.

garyb: Oh, actually, I'm not sure it is - we don't get the accumulated context back when using `everythingWithContextOnTypes`, unless you had something else in mind?

paf31:
:+1:

garyb:
I did come up with a way of only traversing once, but it made the code much harder to understand so I'll merge it as is for now. This code will probably need refactoring to support #1465 anyway.

PR #1466, opened 2015-09-13
Fix #922

### LINKED ISSUE #922: Instance Resolution Problem

paf31:
Another stab at #922. I'd like to be able to test with `core-tests` before merging this since it's in a pretty critical part of the type checker.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -282,10 +282,12 @@ infer' v@(Constructor c) = do
     Just (_, _, ty, _) -> do (v', ty') <- sndM (introduceSkolemScope <=< replaceAllTypeSynonyms) <=< instantiatePolyTypeWithUnknowns v $ ty
                              return $ TypedValue True v' ty'
 infer' (Case vals binders) = do
-  ts <- mapM infer vals
+  (vals', ts) <- fmap unzip $ forM vals $ \val -> do
+    TypedValue _ val' ty <- infer val
+    instantiatePolyTypeWithUnknowns val' ty
   ret <- fresh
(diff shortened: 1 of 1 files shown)


paf31:
Ok, `core-tests` seems to pass. Not sure why I backed out the previous commit for this, maybe something else was failing to build. Could you try `slamdata` when you get chance please @garyb?

garyb:
Cool! I'll have to back port it to 0.7.2 to test, but will definitely check it out tomorrow.

paf31:
>  I'll have to back port it to 0.7.2 to test

Maybe a good use case for https://github.com/ThomasCrvsr/psvm-js ?

garyb:
:shipit: :+1: - tested on slamdata now.

paf31:
Thanks!

garyb:
`purescript-freet` doesn't compile after this :cry:

garyb:
The error:

```
Error found:
Error in module Control.Monad.Free.Trans:
Error in binding group bimapFreeT:
Error checking that type
  f0 _0 -> g0 _0
subsumes type
  _1 a0 -> _2 a0
Error checking that type
  _1 a0
subsumes type
  f0 _0
Error at D:\Home\GitHub\purescript-freet\src\Control\Monad\Free\Trans.purs line 98, column 88 - line 98, column 110:
  Cannot unify type
    a1
  with type
    a0
See https://github.com/purescript/purescript/wiki/Error-Code-TypesDoNotUnify for more information, or to contribute content related to this error.
```

PR #1467, opened 2015-09-13
Fix #862

### LINKED ISSUE #862: Operator sections with identical types type-check differently

paf31:
@garyb While you're testing SlamData, would you mind giving this one a try too?

I'm sort of surprised things worked as well as they did for as long as they did without this final clause :smile:

Anyway, `core-tests` passes on my machine and I'm going to run the test suite now.

Thanks.

### DIFF
--- a/examples/failing/Arrays.purs
+++ b/examples/failing/Arrays.purs
@@ -1,4 +1,4 @@
--- @shouldFailWith ExprDoesNotHaveType
+-- @shouldFailWith TypesDoNotUnify
 module Main where
 
 import Prelude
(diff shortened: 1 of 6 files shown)


garyb:
Yikes! :smile:

I'm hopeful that one or both of these fixes might enable some progress with #928.

garyb:
:+1: For this one too.

PR #1471, opened 2015-09-14
Restrict the number of binders when checking exhaustivity, fix #1469,…

### LINKED ISSUE #1288: Avoid massive binder lists during checkExhaustive
As Phil clarified:
we need to cap the total number of binders so that we don't run out of memory.

### LINKED ISSUE #1469: Larger pattern matches on 2 arguments gets very slow
The following code takes several seconds to compile on my machine as is, and doesn't finish compiling even after several minutes if I uncomment all the `bar2` lines.

It seems restricted to pattern matching multiple parameters. Compiling several pattern match cases with a single parameter doesn't seem to be slow. I'm guessing there's something very nonlinear happening when adding parameters to the pattern match?

I discovered this when experimenting w/ using ADTs to define nodes in a graph and a function to define edges.

```purescript
module Main where

data Foo =
  A |
  B |
  C |
  D |
  E |
  F |
  G |
  H |
  I |
  J |
  K |
  L |
  M |
  N |
  O |
  P |
  Q |
  R |
  S |
  T |
  U |
  V |
  W |
  X |
  Y |
  Z

-- This one builds fine and fast
bar1 A = "A"
bar1 B = "B"
bar1 C = "C"
bar1 D = "D"
bar1 E = "E"
bar1 F = "F"
bar1 G = "G"
bar1 H = "H"
bar1 I = "I"
bar1 J = "J"
bar1 K = "K"
bar1 L = "L"
bar1 M = "M"
bar1 N = "N"
bar1 O = "O"
bar1 P = "P"
bar1 Q = "Q"
bar1 R = "R"
bar1 S = "S"
bar1 T = "T"
bar1 U = "U"
bar1 V = "V"
bar1 W = "W"
bar1 X = "X"
bar1 Y = "Y"
bar1 Z = "Z"

-- This one takes forever to build w/ A-Z cases
-- On my machine It starts taking several seconds when you uncomment up to ~L
bar2 A A = "A"
bar2 B B = "B"
bar2 C C = "C"
bar2 D D = "D"
bar2 E E = "E"
bar2 F F = "F"
bar2 G G = "G"
bar2 H H = "H"
bar2 I I = "I"
bar2 J J = "J"
bar2 K K = "K"
bar2 L L = "L"
-- bar2 M M = "M"
-- bar2 N N = "N"
-- bar2 O O = "O"
-- bar2 P P = "P"
-- bar2 Q Q = "Q"
-- bar2 R R = "R"
-- bar2 S S = "S"
-- bar2 T T = "T"
-- bar2 U U = "U"
-- bar2 V V = "V"
-- bar2 W W = "W"
-- bar2 X X = "X"
-- bar2 Y Y = "Y"
-- bar2 Z Z = "Z"
bar2 _ _ = ""

```

paf31:
… fix #1288.

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -223,6 +223,8 @@ missingAlternative env mn ca uncovered
   where
   mcases = missingCases env mn uncovered ca
 
+
+
 -- |
 -- Main exhaustivity checking function
 -- Starting with the set `uncovered = { _ }` (nothing covered, one `_` for each function argument),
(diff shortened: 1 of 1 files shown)


nicodelpiano:
:+1: !
I'll add an example like the one in #1469 if it's needed.

garyb:
Does this mean exhaustivity checking "gives up" at a certain point?

paf31:
@garyb Yes it does. I'll probably fix this so that a warning is issued if the limit is hit and the function was determined to be exhaustive.

PR #1472, opened 2015-09-14
Vars should always synthesize monotypes

paf31:
/cc @garyb

### DIFF
--- a/examples/passing/1439.purs
+++ b/examples/passing/1439.purs
@@ -1,0 +1,7 @@
+module Main where
+
+import Prelude
+
+ints = [{f: (+)}, {f: (-)}] <#> \x -> x.f 1 2
+
+main = Control.Monad.Eff.Console.print ints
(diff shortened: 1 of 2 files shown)


paf31:
Hmm, this breaks `liftCoyonedaT`.

garyb:
It'll probably break Halogen and some other stuff too then :(

paf31:
Yeah, I think this means there's something wrong somewhere else, but `Var` should definitely not be synthesizing polymorphic types.

garyb:
What's the `liftCoyonedaT` error?

paf31:
They're all `EscapedSkolem` errors. Maybe this means I should just to implement the new type checker. I'll probably put this on hold until I get a couple of other things finished then.

jdegoes:
> Maybe this means I should just to implement the new type checker.

:+1: :heart_eyes:

PR #1474, opened 2015-09-16
documents alex and happy dependency

### LINKED ISSUE #1446: Document happy and alex as dependencies
I don't fully understand how dependencies are installed by cabal but when installing purescript both from npm and the Arch AUR, the pre-build tests were failing and the install would hang for a while. Once the build complete, it the binary would fail to execute. 

After running the tests myself, I saw that language-javascript hadn't been installed. When trying to install language-javascript, some if its dependencies also hadn't been installed.

I have installed all the dependencies and run the build manually. It is now working so the issue is resolved HOWEVER: should the dependencies have installed on their own?

kRITZCREEK:
Fixes #1446


michaelficarra:
I'd rather fix this through technical means. Until language-javascript lists them as dependencies, we should.

garyb:
We talked about doing that, but don't think it works - they won't necessarily be built and installed before `language-javascript`.

zudov:
I'd rather recommend to install using stack

kRITZCREEK:
:+1: for suggesting stack. I just used it to build the repo without any hassle.

zudov:
@kRITZCREEK :+1:

Perhaps we should change section heading to "Installing from source" and add additional  section "Compiling from source" which would contain something like
```
git clone "https://github.com/purescript/purescript"
cd purescript
stack build
```

A [link to instructions for getting stack](https://github.com/commercialhaskell/stack/wiki/Downloads) also wouldn't hurt.

kRITZCREEK:
I messed this PR up... I'll create a new one :D

Issue #1360, opened 2015-08-11
Capitalized key names in records?


ethul:
Should it be permitted to use capitalized key names in records?

For example:
```purescript
type Test = { "Foo" :: Int }
```
Produces the following externs:
```purescript
type Test = { Foo :: Prim.Int }
```

And the compiler fails with:

```bash
unexpected "Foo"
expecting type
```

garyb:
Yeah, I think this should be allowed - any valid identifier or quote-escaped string should be acceptable if we're following the same rules as javascript for what's allowable in a record... I guess whatever we do here applies to rows as a whole, but again I don't see any issues with that, there are no namespace clashes that can result from allowing captialised row labels.

Looks like we're not pretty-printing the externs file here correctly, but @paf31 was talking about overhauling the externs format soon to just used something based on Show/Read s-exprs, so that will probably fix this.

ethul:
Sounds good. Thanks for taking a look.

rhurmala:
Probably related:
```purescript
type Record = { "label with whitespace" :: String }
```
produces externs
```purescript
type Record = { label with whitespace :: Prim.String }
```
which also fails to compile
```
  Unable to parse module:
    unexpected "with"
    expecting type
```

PR #1480, opened 2015-09-19
Externs file

paf31:
/cc @garyb

This saves externs files using `read` and `show` instead of using `.purs` files. The goal is to fix various encoding bugs, and also to allow editors to access type and kind information. 

This changes the following core functionality:

- Externs information is loaded directly into the `Environment`, instead of type checking it. This should hopefully speed things up, and allow for parallel builds, and more interesting integrations with things like Try PureScript.
- Class lookup now only searches in the modules allowed by the orphan instance restriction. It turned out it was just easier to solve this here than to do it later. This should speed things up, and reduces the memory overhead of duplicating instances.

Still to be done:

- [x] Work on module re-exports (@garyb, I'm not really sure how the new code works here, and I'm almost certain I've broken something) 
- [x] Make the new type class solver to respect imports.
- [x] Check extern file versions.
- [x] Choose a better file format.
- [x] Fix QuickCheck build issue.


garyb:
I'm about to go to bed so I'll look through this tomorrow, but regarding re-exports: that information cannot be constructed from the environment. I think we have to base it on the current behaviour - take the export `DeclarationRef`s parsed from the module, elaborate the list to include instances[1] and elaborate any non-specific exports (when there's no export list that means making `DeclarationRef`s for everything in the module, when there is an export list the only change made is to ensure all data constructors are explicitly listed if they were exported as `X(..)`).

[1] Instance exports maybe irrelevant with your other changes now? I'm pretty sure we can do away with the concept of explicitly exported instances anyway, if that currently isn't the case... it may have even been possible for a while. The reason they exist at all is because of something I did back in the early days - it was a simple way to propagate instances transitively.

paf31:
@garyb We should probably discuss options when we're both online. I'd really like to understand the imports better, and working on this, I definitely had some questions.

My main goal for 0.8 is to understand everything in the compiler again :smile:

paf31:
Instance visibility is only an issue because of how we handle compilation order. Ordinarily, we should be ok, because modules get compiled in order, so the instances in scope certainly include the ones we want. The only issue is when there are too many in scope. We can fix this by doing a depth first traversal of the module dependency graph, instead of doing a topsort and compiling in that order.

Since I'm breaking so many things with this change anyway, I'll see if I can change this too.

paf31:
Ok, this is ready for review.

/cc @kRITZCREEK @garyb 

This should make parallel builds (and real incremental builds) a bit simpler now, which I'm quite excited about :smile: /cc @bitc Perhaps now it will be easier to merge in your `psc-compile` changes?

garyb:
@paf31 I know we agreed this was okay anyway, but to try and clarify my explanation from last night about why I don't think we need to worry about tracking imports to resolve instances now:

1. With the no-orphans rule the instance must either be defined with the class, or with the type
2. Instances are always imported and re-exported from every module they encounter, so that they transitively reach all modules below the defining module in the dependency hierarchy
3. If both the class and the type appear in a module, it means that the instance *must* be defined somewhere above that module in the dependency hierarchy because of 1, and therefore the instance definitely reaches the current module because of 2

So basically: with no-orphans there's no observable difference between the previous behaviour and instances being global, which really is how instances are supposed to be seen anyway, right? :smile:

Ok, now I'll look at the changes...

garyb:
Looks like there are a few tweaks needed for GHC < 7.10.

garyb:
Hmm:

```
Error found:
Error in module Control.Coroutine:
Error in value declaration producer:
Error at purescript-coroutines/src/Control/Coroutine.purs line 83, column 1 - line 84, column 1:
  No instance found for Prelude.Bind (Control.Coroutine.Producer<o436> m434)
See https://github.com/purescript/purescript/wiki/Error-Code-NoInstanceFound for more information, or to contribute content related to this error.
```

paf31:
@kRITZCREEK How's this?

@garyb Seems a little tricky to reproduce. I saw it earlier, now I can't make it fail again ...

paf31:
I'll have a go at the synonym changes as part of this PR actually.

kRITZCREEK:
:+1: that helps a lot. Thanks!

garyb:
Weird, I was getting that error consistently yesterday trying to compile Halogen, but it seems to be fine now. :+1: I guess!

paf31:
Hmm, I'll try to figure out why it was breaking, because I saw the error message too.

garyb:
I'm using this branch as my compiler now, so will let you know if I find a way to reproduce again.

bitc:
@paf31 I'm currently busy with other things unrelated to purescript. Feel free to use what you can from my fork https://github.com/bitc/purescript/tree/psc-compile and I'm also happy to help if you have any specific questions

paf31:
@bitc Sure, just wanted to let you know I'm finally catching up on this .. :)

paf31:
@garyb I'm able to reproduce the issue only with the code prior to `7cf48f0250c4a072822029a3ce19eb76c4242268`.

paf31:
@garyb I'll merge then?

garyb:
Ah okay, cool! Perhaps I checked out the 72ce900 commit earlier and then didn't update when I tried it the other day.

Do you want to combine the synonyms stuff first? Or going to revisit that after it's in? I'm happy with merging though, for sure. This closes about 5 open issues doesn't it?

paf31:
Yeah I'll see if I can sort the synonyms stuff out first, and if not, I'll merge this tonight.

garyb:
Sounds good. Can you run the synonyms stuff on `master` of Halogen as a test when you get further with it? That's quite synonymy and has dependencies on a few things outside of core that are probably worth checking. Perhaps give `optic` a whirl too.

paf31:
Fixes #1360

PR #1483, opened 2015-09-21
Revert "Fix #922"

### LINKED ISSUE #922: Instance Resolution Problem

paf31:
Reverts purescript/purescript#1466

I'll just reimplement this properly.


garyb:
:+1: :cry:

PR #1485, opened 2015-09-24
Expand synonyms during desugaring

paf31:
/cc @garyb 

Probably still a few kinks to work out here.


garyb:
Yeah, looks like it - various tests are failing. It's nice though, tidies up quite a bit (as expected!).

paf31:
@garyb I made some changes to the tests to get things working. Some tests were just incorrect, but the `Superclasses3.purs` test was interesting. It started passing, because after desugaring, the `b` type argument became unimportant, so it didn't matter that it wasn't bound on the right hand side. I think this is probably ok.

I'll do a bit of performance testing/tuning, but does everything look ok otherwise?

paf31:
Irritatingly, it's slower compiling `coroutines` by about 50%, but I'm not sure why.

paf31:
Ok, those tests are passing on my machine. Something odd is going on..

PR #1487, opened 2015-09-25
Remove dodgy check rule

paf31:
I don't think we need this any more. Waiting on tests...


garyb:
Does this depend on #1490 being merged first?

paf31:
I think there will be conflicts either way but they should hopefully be easily resolved.

garyb:
Actually I was wondering if removing this rule comes as a consequence of changes made in #1490, or whether removing it is still valid even before then. If we can get rid of it now, can you explain why we don't need it?

paf31:
It's independent. It used to be necessary for some obscure reason, but now that we change into inference mode whenever checking fails, we don't need it.

garyb:
Right, that makes sense. Thanks for explaining! :+1:

Issue #1013, opened 2015-03-30
Control-D in psci should quit


enolan:
This is standard behavior in pretty much every repl. I'm using 0.6.9.3 from Hackage.

puffnfresh:
:+1:

paf31:
Control-D is used to signal end-of-statement in multi-line mode. Control-C can be used to kill the process.

sharkdp:
:+1: I wanted to suggest this yesterday and just saw the bug report. I am also very much used to Control-D from all kinds of programs and usually think of it as the "safer" way to quit a program: end of input as compared to immediate (C)ancel. There's probably no difference for most REPLs, but anyway.

paf31:
This behavior is inherited from `haskeline`. Changing it would require a patch there I think.

joneshf:
There's a specific command to quit, namely `:quit`.

That said, it'd be nice to interact well with other unix programs. Least surprise and all that stuff. One thing to be aware of is the exit code the process sends. <kbd>CTRL</kbd>+<kbd>C</kbd> will give a non-zero exit code. Whereas <kbd>CTRL</kbd>+<kbd>D</kbd> should give a zero exit code.

Also, idris uses haskeline, and they catch <kbd>CTRL</kbd>+<kbd>D</kbd> and handle it properly. Maybe some insight can be gleaned from there.

joneshf:
Also, doesn't windows use <kbd>CTRL</kbd>+<kbd>Z</kbd> as EOF?

cdepillabout:
I would expect Ctrl-D to quit, where as Ctrl-C will just ignore the current command and go down to an empty prompt, like it does in bash.  Currently in psci, the functionality of Ctrl-D and Ctrl-C are opposite of bash.  It's confusing for a long-time shell user.

zudov:
Same in `ghci`, Ctrl-C causes to cancel the current expression, Ctrl-D quits.

`ghci` also uses `haskelline`

paf31:
Well if someone can figure out the right code to poach from `ghci`, I'll take a PR.

zudov:
This looks useful https://hackage.haskell.org/package/haskeline-0.7.2.1/docs/System-Console-Haskeline.html#v:withInterrupt , at least it should allow not to quit on Ctrl-C.

I'll try to figure it out, after I wake up (if no one would fix it or start working on it it by that time)

zudov:
IMO, in the repl it makes sense that:

[...]

PR #1489, opened 2015-09-25
Ghci-like handling of Ctrl-C, Ctrl-D in psci

zudov:
- When Ctrl-C (UserInterrupt) is pressed, cancel the current input
- When Ctrl-D (End of Input) is pressed:
  - if the input is empty interpret as "there would be no more commands" and quit
  - if the input isn't empty interpret as "the end of the current command"
    and execute the current command (whenether it's single-line or multi-line mode)

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -437,10 +437,10 @@ handleKindOf typ = do
 -- Parses the input and returns either a Metacommand, or an error as a string.
 --
 getCommand :: Bool -> InputT (StateT PSCiState IO) (Either String (Maybe Command))
-getCommand singleLineMode = do
-  firstLine <- getInputLine "> "
+getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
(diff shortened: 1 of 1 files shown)


paf31:
That simple, eh? :astonished: 

Thanks, will merge once CI finishes! :+1:

zudov:
Perhaps we also could catch Ctrl-C *while* the command is running and prevent psci from terminating, but I am not sure what is a better way to do that.
We need to somehow catch AsyncException in the `PSCi` monad, ExceptT? MonadError?, wrapping unwrapping the IO?

paf31:
Sounds like a job for `lifted-base` although I'm not an expert here.

zudov:
@paf31 lifted-base looks nice, I'll try it now

zudov:
@paf31 Heh, it turned out to be much simpler. Should be ready to merge after CI passes.

paf31:
:+1:

paf31:
Fixes #1013

paf31:
:+1: Thanks!

PR #1490, opened 2015-09-26
Expand type synonyms eagerly

paf31:
/cc @garyb 

Tested on Halogen master.

I think this makes things quite a bit simpler, which I like. The only possible issue is error messages, but I think in practice, it's not so much of a concern. In many cases, you actually want to expand synonyms in errors, for example when dealing with `lens` for example. And in most cases, getting to the root of a type issue often requires looking under synonyms anyway.

Thoughts?

### DIFF
--- a/src/Language/PureScript/Docs/Convert.hs
+++ b/src/Language/PureScript/Docs/Convert.hs
@@ -166,7 +166,6 @@ convertDeclaration (P.TypeInstanceDeclaration _ constraints className tys _) tit
   unQual x = let (P.Qualified _ y) = x in P.runProperName y
 
   extractProperNames (P.TypeConstructor n) = [unQual n]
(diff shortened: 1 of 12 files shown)


garyb:
Still failing in all compilers :cry:

paf31:
huh.

paf31:
Ah, it's just `-Werror` saving me from being an idiot again. Should hopefully work this time.

What do you think of the approach though?

garyb:
Oops, I thought this was an updated version of the desugaring approach. I'm getting ready for tomorrow's trip to Boulder at the moment, but I'll try and review properly in a bit.

garyb:
I missed you asking about this last night - seems like a good approach to me, and well, it works! :smile:

Issue #1087, opened 2015-04-28
Parallel builds


paf31:
I think `monad-par` would be a nice approach.

puffnfresh:
Slow builds are killing me right now :+1:

paf31:
Related: https://github.com/bitc/purescript/tree/psc-compile/psc-compile

This is probably a better approach: simplify the `psc` build when modules didn't change, and get parallelism by using multiple processes and make/Shake.

`psc` could even support a `--dump-make-file` option.

PR #1491, opened 2015-09-27
Parallel Builds

paf31:
Working towards parallel builds. This breaks right now because of something to do with imports. It seems to be related to modules with selective imports.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -31,6 +31,7 @@ PureScript uses the following Haskell library packages. Their license files foll
   array
   attoparsec
   base
+  binary
   blaze-builder
   bower-json
   boxes
@@ -45,9 +46,10 @@ PureScript uses the following Haskell library packages. Their license files foll
   haskeline
   integer-gmp
   language-javascript
+  lifted-base
+  monad-control
   mtl
   nats
-  old-locale
   optparse-applicative
   parsec
   pattern-arrows
@@ -59,12 +61,14 @@ PureScript uses the following Haskell library packages. Their license files foll
   scientific
   semigroups
   split
+  stm
   syb
   template-haskell
   terminfo
   text
   time
   transformers
+  transformers-base
   transformers-compat
   unix
   unordered-containers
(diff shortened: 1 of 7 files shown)


paf31:
@garyb This reduces the recompile time for Halogen master from 2.4s to 1.6s on my machine. I suspect we could improve it quite a bit more if we add

- parallel compilation
- caching externs

puffnfresh:
Parallel compilation didn't achieve much for compilation of a project when I got a version working - it took around 0.1s off of a 6s project:

https://github.com/puffnfresh/purescript/commit/6c443ecb5a271f5954aadf6614572077b7a1ae17

The way I did it was pretty gross so I decided it wasn't worth it.

I'd be interested to see your results!

paf31:
@puffnfresh Interesting, were you typechecking every module, even externs? The main change here is to avoid typechecking externs, and just assuming they are correct. Since typechecking is the slowest part, I'm hoping there will be some degree of parallelism available there, but maybe not...

paf31:
So with the last commit, Halogen master builds in 8.6 seconds (with a <= 1 second artificial delay until I properly implement waiting for completion) vs 10 seconds on my machine with 0.7.4.1. Now, some of that might be due to externs file changes, I'm not sure, but it seems worth continuing.

Also, could someone please check if I'm using the runtime system options correctly?

garyb:
RTS options look good to me, or at least that's what I did for the few programs I've written that are parallel.

paf31:
Halogen master is down to 5.5 with `-N4` :smile:

paf31:
One issue with this approach is that I get `psc: thread blocked indefinitely in an MVar operation` whenever compilation fails, so it'll be necessary to catch errors and fill any empty MVars accordingly.

garyb:
Do you need to specify `-N4`? I thought `-N` with no number threads it as much as the machine supports.

paf31:
`-N4` just seemed to be the sweet spot for my machine, but I don't think it's necessary.

paf31:
I hope to work on this a bit more today.

paf31:
@garyb @puffnfresh Could you please review this? The current Halogen build takes just over 5s on my machine, with 4 threads, compared to 10s before. The rebuild takes 1s, almost all of which is taken computing the module graph, so I intend to look into that next, and hopefully we can get a really zippy rebuild time.

paf31:
Resolves #1087

puffnfresh:
This seems very similar to what I did a while ago, unsure why I got poor results, so good job! :+1:

paf31:
Confirmed: the current slowness on a rebuild is due to parsing.

PR #1498, opened 2015-10-01
fixed #1496.

### LINKED ISSUE #1496: psc can not compile source files contains multibyte sequence charactor on Windows
psc can't compile this file on Windows System.
```haskell
module Main where

main = do
  log "Hello λ→ world!"

-- '→' is multibyte sequence \u2192.
```

Error message:
```
Main.purs: hGetContents: invalid argument (invalid byte sequence)
```
Because System.IO.readFile convert byte sequence from system's default encoding. And unlike Linux or mac, default encoding on Windows is not UTF-8, but CPxxx.
So psc should not call readFile, use openFile and hSetEncoding to force encoding set to UTF-8.

senju:
I fix #1496.


paf31:
:+1:  Looks good, thanks.

Could you please also add an entry to `CONTRIBUTORS.md` since this is your first PR?

senju:
Thank you for the quick response.

garyb:
:+1:

PR #1501, opened 2015-10-03
Combined feature branch

paf31:
This contains feaures from the `synonyms2` and `incremental` branches.

@garyb If everything passes in CI, I'm going to spend some time using this build this week, and then if we're happy with the code, I'll just merge it back to `master` in one hit. Would you be able to put it through its paces too, please?


paf31:
Heh, this actually found a bug in my code, which for some reason wasn't showing up before.

garyb:
Anecdotally seems quite a bit faster compiling Halogen :)

I'll stick with it this week too, no problems so far though.

paf31:
Resolves #1087.

paf31:
Same with this one, any other comments for review, or shall we test on `master` for a bit longer before 0.7.5?

PR #1504, opened 2015-10-04
fixes #1502: make `qualified` keyword optional in qualified imports

### LINKED ISSUE #1502: make "qualified" keyword optional
Is there a reason why the `qualified` keyword in qualified imports is required? I don't see any reason why we can't write

```purs
import qualified A.B.C as C
```

as

```purs
import A.B.C as C
```

michaelficarra:
Fixes #1502.

### DIFF
--- a/examples/passing/OptionalQualified.purs
+++ b/examples/passing/OptionalQualified.purs
@@ -1,0 +1,13 @@
+module Main where
+
+-- qualified import with the "qualified" keyword
+import qualified Prelude as P
+
+-- qualified import without the "qualified" keyword
+import Control.Monad.Eff.Console as Console
+
+bind = P.bind
+
+main = do
+  message <- P.return "success!"
+  Console.log message
--- a/examples/passing/QualifiedQualifiedImports.purs
+++ b/examples/passing/QualifiedQualifiedImports.purs
@@ -1,0 +1,6 @@
+module Main where
+
+-- qualified import with qualified imported names
+import qualified Control.Monad.Eff.Console (log) as Console
+
+main = Console.log "Success!"
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -151,24 +151,25 @@ parseImportDeclaration' = do
   where
   stdImport = do
     moduleName' <- moduleName
-    stdImportHiding moduleName' <|> stdImportQualifying moduleName'
+    suffixHiding moduleName' <|> suffixQualifyingList moduleName'
     where
-    stdImportHiding mn = do
+    suffixHiding mn = do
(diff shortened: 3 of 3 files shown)


examples/failing/QualifiedQualifiedImports.purs
```diff
@@ -0,0 +1,9 @@
+-- @shouldFailWith ErrorParsingModule
+module Main where
+
+import Prelude
+
+-- qualified import with qualified imported names
+import qualified Control.Monad.Eff.Console (log) as Console
```
garyb: I don't think this should fail - this style of import allows you to selectively re-export members from another module that you also want to import as qualified.

So re-exporting `Console` here would only re-export `Console.log`.

michaelficarra: It was an untested feature, so I figured it was accidentally allowing this form since I didn't consider this use case. I'll move this to the passing tests and make it work again.

garyb: Thanks! And good thinking, I didn't realise we weren't covering this case.

michaelficarra:
Alright, I've made that a passing test now. Please take another look, @garyb.

garyb:
Looks good :+1:

garyb:
@paf31 any comments?

paf31:
Sorry, been a bit swamped this week.

:+1: Looks good.

Two observations:

1. Do we need to update the wiki?
2. While we're considering what should be valid combinations, why don't we allow things like `import Foo (foo, bar) as X hiding (baz)`. I know this seems silly, but there is one potential use case: if you have overlaps in two modules, but want to import them qualified into the same phantom module. I don't feel strongly about this, I just found it interesting that the code now makes `Hiding`, `Explicit` mutually exclusive, given that we can now have two modules imported into the same namespace.

garyb:
1. Yeah, I think that needs doing anyway as the current permutations of import behaviours aren't covered anyway.
2. This one made my head hurt :wink:. But actually the import you illustrated selectively imports `foo, bar` into `X`, it doesn't import `foo, bar` as unqualified and then the rest of the module as `X`. You can only import unqualified or qualified. There is a case for hiding qualified: the same thing you might want to do with an unqualified import, `import Foo as X hiding (baz)`. I guess the more natural formulation of this would be `import Foo hiding (baz) as X`.

I'm happy for this to go in as is though and we can figure out 2 separately, if you are?

paf31:
Sorry, that's what I meant :smile:

michaelficarra:
> I'm happy for this to go in as is though and we can figure out 2 separately, if you are

I think that would be best. This conversation could get pretty heated and I'd rather not have it block this simple change.

garyb:
:+1:

Issue #975, opened 2015-03-18
psci: browse does not include data constructors


hdgarrood:
Arguably `:browse` in psci should include data constructors.

hdgarrood:
It also does not include type class members.

hdgarrood:
And I guess it should include types, type classes, and instances, too? Maybe some of these things would be better done by `:info`, when we get around to it.

PR #1515, opened 2015-10-10
including data constructors, data and type declarations in psci :browse command #975

soupi:
[before](http://lpaste.net/142718) -> [after](http://lpaste.net/142693)

this is related to issue #975


paf31:
Thanks, I'll try to review this over the weekend.

For now, could you please add your licensing terms to [this file](https://github.com/purescript/purescript/blob/master/CONTRIBUTORS.md) as part of this PR? Thanks.

psci/PSCi.hs
```diff
@@ -380,17 +381,50 @@ printModuleSignatures :: P.ModuleName -> P.Environment -> PSCI ()
...
-        findType envNames m@(_, mIdent) = (mIdent, M.lookup m envNames)
-        showType :: (P.Ident, Maybe (P.Type, P.NameKind, P.NameVisibility)) -> String
-        showType (mIdent, Just (mType, _, _)) = show mIdent ++ " :: " ++ P.prettyPrintType mType
-        showType _ = error "The impossible happened in printModuleSignatures."
+                  _ -> do
+                    (outputStr   . unlines . sort . mapMaybe (showType typeSynonymsEnv . findType typesEnv)) moduleTypes
+                    (outputStr   . unlines . map showDataConstructor . sortBy compareDatatypes . map (findDataConstructor dataConstructorsEnv)) moduleDataConstructors
+                    (outputStrLn . unlines . sort . map (showNameType . findNameType namesEnv)) moduleNamesIdent
+
+  where findNameType :: M.Map (P.ModuleName, P.Ident) (P.Type, P.NameKind, P.NameVisibility) -> (P.ModuleName, P.Ident) -> (P.Ident, Maybe (P.Type, P.NameKind, P.NameVisibility))
```
paf31: Could you please add a space between declarations here? I think it will make it slightly easier to read. Also, to be consistent with other code, please align the declarations with the `where`.

psci/PSCi.hs
```diff
@@ -380,17 +381,50 @@ printModuleSignatures :: P.ModuleName -> P.Environment -> PSCI ()
 printModuleSignatures moduleName env =
   PSCI $ let namesEnv = P.names env
              moduleNamesIdent = (filter ((== moduleName) . fst) . M.keys) namesEnv
+             typesEnv = P.types env
```
paf31: Why not use `RecordWildcards` to bring these into scope all at once?

psci/PSCi.hs
```diff
@@ -380,17 +381,50 @@ printModuleSignatures :: P.ModuleName -> P.Environment -> PSCI ()
...
+             typeSynonymsEnv = P.typeSynonyms env
+             moduleTypes = (filter (\(P.Qualified maybeName _) -> maybeName == Just moduleName) . M.keys) typesEnv
+             dataConstructorsEnv = P.dataConstructors env
+             moduleDataConstructors = (filter (\(P.Qualified maybeName _) -> maybeName == Just moduleName) . M.keys) dataConstructorsEnv
              in case moduleNamesIdent of
```
paf31: This check is too restrictive now.

paf31: Why not factor this into a function which produces a list, and then a function which renders it? Then you can write this check using `null`.

psci/PSCi.hs
```diff
@@ -380,17 +381,50 @@ printModuleSignatures :: P.ModuleName -> P.Environment -> PSCI ()
...
-        showType :: (P.Ident, Maybe (P.Type, P.NameKind, P.NameVisibility)) -> String
-        showType (mIdent, Just (mType, _, _)) = show mIdent ++ " :: " ++ P.prettyPrintType mType
-        showType _ = error "The impossible happened in printModuleSignatures."
+                  _ -> do
+                    (outputStr   . unlines . sort . mapMaybe (showType typeSynonymsEnv . findType typesEnv)) moduleTypes
+                    (outputStr   . unlines . map showDataConstructor . sortBy compareDatatypes . map (findDataConstructor dataConstructorsEnv)) moduleDataConstructors
+                    (outputStrLn . unlines . sort . map (showNameType . findNameType namesEnv)) moduleNamesIdent
+
+  where findNameType :: M.Map (P.ModuleName, P.Ident) (P.Type, P.NameKind, P.NameVisibility) -> (P.ModuleName, P.Ident) -> (P.Ident, Maybe (P.Type, P.NameKind, P.NameVisibility))
+        findNameType envNames m@(_, mIdent) = (mIdent, M.lookup m envNames)
```
paf31: Consider using `M.mapWithKey` instead of mapping and looking up each key.

psci/PSCi.hs
```diff
@@ -380,17 +381,50 @@ printModuleSignatures :: P.ModuleName -> P.Environment -> PSCI ()
...
+        showNameType :: (P.Ident, Maybe (P.Type, P.NameKind, P.NameVisibility)) -> String
+        showNameType (mIdent, Just (mType, _, _)) = P.showIdent mIdent ++ " :: " ++ P.prettyPrintType mType
+        showNameType _ = error "The impossible happened in printModuleSignatures."
+        findDataConstructor :: M.Map (P.Qualified P.ProperName) (P.DataDeclType, P.ProperName, P.Type, [P.Ident]) -> P.Qualified P.ProperName -> (P.Qualified P.ProperName, Maybe (P.DataDeclType, P.ProperName, P.Type, [P.Ident]))
+        findDataConstructor envDataCons name = (name, M.lookup name envDataCons)
+        compareDatatypes :: (P.Qualified P.ProperName, Maybe (P.DataDeclType, P.ProperName, P.Type, [P.Ident])) -> (P.Qualified P.ProperName, Maybe (P.DataDeclType, P.ProperName, P.Type, [P.Ident])) -> Ordering
+        compareDatatypes (_, Just (_, dtName1, _, _)) (_, Just (_, dtName2, _, _)) = dtName1 `compare` dtName2
```
paf31: Consider using `comparing`.

psci/PSCi.hs
```diff
@@ -380,17 +381,50 @@ printModuleSignatures :: P.ModuleName -> P.Environment -> PSCI ()
...
+            (Just (_, P.TypeSynonym), Just (_, dtType)) ->
+                Just ("type " ++ P.runProperName name ++ " = " ++ P.prettyPrintType dtType)
+
+            (Just (_, P.DataType _ pt), _) ->
+                Just ("data " ++ P.runProperName name ++ if null pt then "" else " = " ++ printCons pt)
+
+            _ ->
+              Nothing
+
+          where printCons pt = intercalate " | " (map (\(cons,idents) -> unwords (P.runProperName cons : map prettyPrintType idents)) pt)
```
paf31: There might be some duplication with `psc-docs` here. Perhaps this can be made into common code?

paf31:
Looks good, thanks. I added some comments.

Also, as I mentioned on IRC, `prettyPrintType` will change in 0.7.5 to render types over multiple lines, so we should consider using the `boxes` library here too. It might be good to rebase your branch from `box-errors`, since that branch is ready, and will probably be merged before this one.

soupi:
tried to make some changes and add `boxes` support. here is the new "[after](http://lpaste.net/4596355708189933568)"

tfausak:
:+1:

paf31:
@soupi Could you please rebase this? I have a few more comments but I'd like to get it merged before 0.7.6 if possible. Thanks!

soupi:
@paf31 pulled from master :)

psci/PSCi.hs
```diff
@@ -376,20 +378,66 @@ handleTypeOf val = do
...
+        showDataConstructor _ = P.internalError "The impossible happened in printModuleSignatures."
+
+        findType :: M.Map (P.Qualified P.ProperName) (P.Kind, P.TypeKind) -> P.Qualified P.ProperName -> (P.Qualified P.ProperName, Maybe (P.Kind, P.TypeKind))
+        findType envTypes name = (name, M.lookup name envTypes)
+
+        showType :: M.Map (P.Qualified P.ProperName) ([(String, Maybe P.Kind)], P.Type) -> (P.Qualified P.ProperName, Maybe (P.Kind, P.TypeKind)) -> Maybe String
+        showType typeSynonymsEnv (n@(P.Qualified _ name), typ) =
+          case (typ, M.lookup n typeSynonymsEnv) of
+            (Just (_, P.TypeSynonym), Just (typevars, dtType)) ->
+                Just (Box.render $ Box.text ("type " ++ P.runProperName name ++ concatMap ((' ':) . fst) typevars) Box.// Box.moveRight 2 (Box.text "=" Box.<+> P.typeAsBox dtType))
```
paf31: Consider using `unwords` instead of `concatMap` here?

psci/PSCi.hs
```diff
@@ -376,20 +378,66 @@ handleTypeOf val = do
...
+            _ ->
+              Nothing
+
+          where printCons pt =
+                    Box.vcat Box.left $
+                    map (Box.moveRight 2) $
+                    mapFirstRest (Box.text "=" Box.<+>) (Box.text "|" Box.<+>) $
+                    map (\(cons,idents) -> (Box.text (P.runProperName cons) Box.<> Box.hcat Box.left (map prettyPrintType idents))) pt
+
+                prettyPrintType t@(P.TypeApp _ _) = Box.text " (" Box.<> P.typeAsBox t Box.<> Box.text ")"
```
paf31: `typeAtomAsBox`?

psci/PSCi.hs
```diff
@@ -376,20 +378,66 @@ handleTypeOf val = do
...
+
+            (Just (_, P.DataType typevars pt), _) ->
+                Just $ Box.render (Box.text ("data " ++ P.runProperName name ++ concatMap ((' ':) . fst) typevars) Box.// printCons pt)
+
+            _ ->
+              Nothing
+
+          where printCons pt =
+                    Box.vcat Box.left $
+                    map (Box.moveRight 2) $
```
paf31: Maybe just `moveRight` the result of `vcat` and avoid an extra `map`?

psci/PSCi.hs
```diff
@@ -376,20 +378,66 @@ handleTypeOf val = do
...
-        showType :: (P.Ident, Maybe (P.Type, P.NameKind, P.NameVisibility)) -> String
-        showType (mIdent, Just (mType, _, _)) = show mIdent ++ " :: " ++ P.prettyPrintType mType
-        showType _ = P.internalError "The impossible happened in printModuleSignatures."
+printModuleSignatures moduleName (P.Environment {..}) =
+  PSCI $ let moduleNamesIdent = (filter ((== moduleName) . fst) . M.keys) names
+             moduleTypes = (filter (\(P.Qualified maybeName _) -> maybeName == Just moduleName) . M.keys) types
+             moduleDataConstructors = (filter (\(P.Qualified maybeName _) -> maybeName == Just moduleName) . M.keys) dataConstructors
+
+         in do
+           printModule's "types"             (sort . mapMaybe (showType typeSynonyms . findType types)) moduleTypes
```
paf31: I think it would be nice if we could make only one call to `outputStrLn` here, and combine all of the text into one `Box` first, instead of calling `outputStrLn` once for each entry.

psci/PSCi.hs
```diff
@@ -376,20 +378,66 @@ handleTypeOf val = do
...
+           printModule's "data constructors" (map showDataConstructor . sortBy compareDatatypes . map (findDataConstructor dataConstructors)) moduleDataConstructors
+           printModule's "functions"         (sort . map (showNameType . findNameType names)) moduleNamesIdent
+           outputStrLn   ""
+
+
+  where printModule's what _  [] = outputStrLn $ "This module '" ++ P.runModuleName moduleName ++ "' does not export " ++ what ++ ".\n"
```
paf31: I'm not sure we need to say this. Why not just print the entries if they're there?

paf31:
Could you please paste the latest output for something like `Prelude`?

psci/PSCi.hs
soupi: wasn't sure exactly how to use unwords here, did you mean something like `(' ':) . unwords . map fst` ?

soupi:
[This](http://lpaste.net/145159) is the latest output for `Prelude` and my own [`Main`](http://lpaste.net/167099361963016192).

soupi:
[Here](http://lpaste.net/145178) is the new output on `Prelude` and `Main`. Showing type class definitions and distinguishing between `data` and `newtype`.

paf31:
This looks great, but just three small points:

- There is no break between classes and types, but there is a break between items in each of those groups, which looks a little odd. I'd rather both have breaks, or just have breaks between groups.
- You might like to trim whitespace from the ends of lines, like I do with the errors, so that you don't get weird wrapping errors with wide output.
- This appears in the output when I `:browse Prelude`:

```purescript
data Ordering                                                         
  = LT                                                                
  | GT                                                                
  | EQ               
                                                  
EQ :: Ordering                                               
GT :: Ordering                                              
LT :: Ordering  
```

I don't think we need both representations of the constructors.

paf31:
Closing as duplicate.

Issue #1357, opened 2015-08-11
Type errors should track terms they originated from


paf31:

hdgarrood:
Does this subsume #909?

Issue #908, opened 2015-02-23
Errors which occur inside a type class instance are reported at the definition of the instance


paf31:
Reported by @jdegoes in #703

PR #1516, opened 2015-10-11
Hints

paf31:
@garyb Does this look ok? I think it simplifies things nicely and gives a nicer set of error messages.

Basically the new simplification rules is: take at most one hint from each hint category.

Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ node_modules
 tmp/
 .stack-work/
 tests/support/flattened/
+output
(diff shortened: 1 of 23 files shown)


paf31:
I'd say this resolves #1357.

paf31:
This also simplifies the approach to position info slightly. Position errors always get consed to the front of the hint list, which means the most specific position data is usually at the end of the list. By reversing before picking unique hints by category, we automatically get the most specific position info.

paf31:
I don't know if this was fixed by this change or something else, but #908 is definitely working on this branch.

Resolves #908

paf31:
Resolves #909

garyb:
Sounds awesome, but think this will take me a while to digest, so it'll probably be tomorrow before I get time to read it properly.

paf31:
@garyb Unless you see anything major, I'll tidy a couple of things up and merge into `box-errors`, then everything can be reviewed together. Sound good?

garyb:
Sure thing. Code looks good to me too, I haven't tried it out yet but the idea behind it sounds solid.

PR #1518, opened 2015-10-11
Fix #1261

### LINKED ISSUE #1261: Fix name on Let exhaustivity checking warnings
```purescript
module Main where
f 0 = let g 0 = 0 in g
```
Should return `g` once, instead of two `f`s.

paf31:

In the end, it was easier to just write the traversal out by hand.


src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -273,25 +271,36 @@ checkExhaustive env mn numArgs cas = makeResult . first nub $ foldl' step ([init
...
+  onExpr (UnaryMinus e) = onExpr e
+  onExpr (ArrayLiteral es) = mapM_ onExpr es
+  onExpr (ObjectLiteral es) = mapM_ (onExpr . snd) es
+  onExpr (TypeClassDictionaryConstructorApp _ e) = onExpr e
+  onExpr (Accessor _ e) = onExpr e
+  onExpr (ObjectUpdate o es) = onExpr o >> mapM_ (onExpr . snd) es
+  onExpr (Abs _ e) = onExpr e
+  onExpr (App e1 e2) = onExpr e1 >> onExpr e2
+  onExpr (IfThenElse e1 e2 e3) = onExpr e1 >> onExpr e2 >> onExpr e3
+  onExpr (Case es cas) = checkExhaustive env mn (length es) cas >> mapM_ onCaseAlternative cas
```
paf31: Oops, looks like I need to recur on `es`.

garyb:
I'm not sure I understand this fix, but then I've not seen the error the issue produces either, so... :)

Issue #1315, opened 2015-07-29
Suggestions for row unification errors


hdgarrood:
Two examples of modules which have type errors, and the error messages that they currently produce:
* Extra record field
```purescript
module ExtraRecordField where
import Prelude ((<>))

full :: { first :: String, last :: String } -> String
full p = p.first <> " " <> p.last
oops = full { first: "Jane", last: "Smith", age: 29 }
```
```
Row last :: Prim.String, first :: Prim.String lacks required property "age"
```
* Missing record field
```purescript
module MissingRecordField where
import Prelude ((>))

john = { first: "John", last: "Smith" }
isOver50 p = p.age > 50.0
result = isOver50 john
```
```
  Cannot unify type
    ()
  with type
    (age :: Prim.Number | _1)
```

I think these could be improved a little. In the first example, I think it makes more sense to say that the object `{ first: "Jane", last: "Smith", age: 29 }` has an *extra* field "age", rather than the row `{ first :: String, last :: String }` missing the required field "age". In the second, I think a `PropertyIsMissing` error would be more appropriate than the current "Cannot unify type" error.

paf31:
:+1:

PR #1520, opened 2015-10-11
Remove inferProperty judgment

paf31:
@garyb Just this last one for today then I'm calling it a day. I spotted this while starting to work on #1315, and I think it's better to remove the special case and just use `check`.

Look ok?


garyb:
Makes sense, seems a lot simpler to do it this way - I assume it doesn't break anything obvious? :wink:

paf31:
Not that I can see. It potentially has slightly different behavior in some rank-N cases, but nothing major I think.

PR #1522, opened 2015-10-13
changes the recommended build tool to stack

### LINKED ISSUE #1446: Document happy and alex as dependencies

kRITZCREEK:
fixes #1446, because stack takes care of installing alex and happy.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -24,14 +24,24 @@ requirements.
 ## Compiling from source
 
 GHC 7.6.1 or newer is required to compile from source. The easiest way is to
-use `cabal-install`:
+use stack:
 
 ```
-$ cabal update && cabal install purescript
+$ stack install purescript
 ```
+
+This will then copy the compiler and utilities into `~/.local/bin`.
+
+
+If you don't have stack installed yet there are install instructions
+[here](https://github.com/commercialhaskell/stack/blob/master/doc/install_and_upgrade.md).
+
+If you don't have ghc installed yet, stack will prompt you to run `stack setup`
+which will install ghc for you.
 
 The PureScript compiler has been known to run on OS X 10.6 when built with GHC
 7.6.
+
 
 ## The "curses" library
 


paf31:
Looks good to me, thanks!

paf31:
Any comments on this before I merge it?

garyb:
:+1: I think the claim that `stack` is easier than `cabal` given the alex/happy situation is definitely true, so makes sense to me.

hdgarrood:
There is one problem in recommending stack over cabal, which is that `stack install purescript` currently gives you 0.7.4.1, and it's not clear (to me at least) how to install 0.7.6.1 using stack.

hdgarrood:
```
$ stack install purescript-0.7.6.1
Run from outside a project, using implicit global project config
Using resolver: lts-3.10 from implicit global project's config file: /home/harry/.stack/global-project/stack.yaml
Error parsing targets: Specified target version 0.7.6.1 for package purescript does not match snapshot version 0.7.4.1
```

kRITZCREEK:
You can get the newest purescript version by changing your global resolver from lts-3.10 to the newest lts-3.19. You can change that inside `/home/harry/.stack/global-project/stack.yaml`.

The main benefit stack gives us that installing purescript from source gets a lot easier. I for example need to switch in between different purescript versions for psc-ide development and it's a lot easier with stack.
My usual workflow for stack with purescript:

```sh
git clone git@github.com:purescript/purescript.git
cd purescript
git checkout v0.7.6.1 # or whatever version/commit I need
stack install
```

hdgarrood:
I'm not arguing that we should change it back necessarily, I just think we can't assume understanding of stack resolvers, so we need to expand INSTALL.md a little. Can you specify the resolver to use on the command line? I think that would be better than asking people to modify their global config. Also, is there a shortcut for the latest lts? Ideally I think we would change the command to something like

stack install --resolver latest-lts purescript

Basically, it should always install the same version as cabal update && cabal install purescript would, I think.

kRITZCREEK:
From the [official documentation](http://docs.haskellstack.org/en/stable/GUIDE.html#resolvers-and-changing-your-compiler-version):

> When passed on the command line, you also get some additional “short-cut” versions of resolvers: 
> --resolver nightly will use the newest Nightly resolver available,
> --resolver lts will use the newest LTS, 
> and --resolver lts-2 will use the newest LTS in the 2.X series.

So I gues we suggest the `--resolver lts` option?

hdgarrood:
Normally I would have looked myself but I'm out ;) thanks!

PR #1525, opened 2015-10-14
Don't require that all types are exported

garyb:
This never entirely worked properly when synyonms and the like got involved, and now we generate a decent externs format we don't need the check to prevent errors anymore either.

Also the check isn't necessary, as maybe you only want to reveal a type via synonym and then use prisms to access it or the like, so imposing that All Types Must Be Exported is a fairly arbitrary rule now that their absence won't break things.


paf31:
Another option I'd be fine with is to have synonyms be always implicitly exported.

If you agree that exports/imports should represent what's happening at the "linker" level (Common JS in our case), then this seems like the right way.

That said, this PR looks good to me in any case :+1:

garyb:
^ Hahaha, a bit late there Coveralls

PR #1527, opened 2015-10-14
Fix #1524, render kinds as single line strings.

### LINKED ISSUE #1524: Row rendering issue
@paf31 I'm seeing things like this in the generated docs when using the current master:

``` purescript
data ExistsR (f :: # * -> *
)
```

``` purescript
newtype IProp (r :: # *
) i
  = IProp (Prop i)
```

We can probably do better than that :smile:. I assume errors for these types would appear similiarly in the output too?

paf31:
@garyb This should fix the formatting issue in the documentation.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -447,7 +447,7 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
             ]
     renderSimpleErrorMessage (InfiniteKind ki) =
       paras [ line "An infinite kind was inferred for a type: "
-            , indent $ kindAsBox ki
+            , indent $ line $ prettyPrintKind ki
             ]
     renderSimpleErrorMessage (MultipleFixities name) =
       line $ "Multiple fixity declarations for " ++ showIdent name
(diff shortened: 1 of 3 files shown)


garyb:
I'm not so sure this is the right fix - isn't it more a case of not boxifying single elements? As you get things like this too:

```
Error checking that value
  (($) const) ((div [ initializer
                    ]
               )
               []
              )
```

paf31:
I think the issue is that putting kinds in a box results in a newline at the end.

Also, that layout is by design :) What do you think is wrong with it? (Apart from the additional parens, which I agree can be improved)

garyb:
I think that particular example split over rows is harder to read than a one liner:

```
(($) const) ((div [ initializer ]) [])
```

I can see the advantage if the array literal had multiple values or if the expression as a whole reached a certain level of complexity, but I guess it's pretty hard to encode something like that!

garyb:
I guess maybe... preserve single line for one-element array literals, one-argument functions, one-value rows would be a start?

I think the most ideal thing would be to use the actual source with its original layout, but that's not retrievable after desugaring so probably a bit tricky to do without some serious work.

paf31:
Possibly, but you can end up with arbitrarily long strings that way, made up of lots of single-element arrays (locally small, globally large), which is why I decided to only put empty arrays one one line.

This PR should fix the issue with generating docs, which is limited to kinds anyway, and we can always discuss this array issue separately.

paf31:
Any more thoughts on this @garyb ?

garyb:
Oh, sorry - :+1:

paf31:
Thanks!

PR #1528, opened 2015-10-15
Kind check constraint contexts, fix #1071

### LINKED ISSUE #1071: Multi-parameter type class constraints can be missing type arguments?
This seems like it shouldn't compile, but it does:

```purescript
class Foo a b where
  foo :: a -> b

-- `Foo a` shouldn't be a valid constraint, since it still needs another type argument
bar :: forall a. (Foo a) => a -> a
bar = id
```

paf31:
Also fixes an off-by-one-space rendering issue.

### DIFF
--- a/examples/failing/1071.purs
+++ b/examples/failing/1071.purs
@@ -1,0 +1,8 @@
+-- @shouldFailWith KindsDoNotUnify
+module Main where
+
+class Foo a b where
+  foo :: a -> b
+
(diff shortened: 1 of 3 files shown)


garyb:
!

Surprised I've not been bitten by this one actually :+1:

paf31:
It's only an issue when the class has no instances, since they would be checked then in any case, so it's not so surprising that you didn't run into it.

Issue #1318, opened 2015-07-30
Invalid externs.purs generated containing Unknown Type


bitc:
The following module:

``` purescript
module BugTest (B(..)) where

data A = A

data B = B1 | B2 A
```

generates the following `externs.purs`:

``` purescript
module BugTest (B(B1, B2)) where
import Prim ()
data B = B1  | B2 BugTest.A
```

It does not compile and gives the error: `Unknown type BugTest.A`

This is code that you probably should not write, but still valid (also in Haskell where there are no warnings even).

I think the fix is to put in the `externs.purs` also the line `data A = A`

garyb:
Thanks for the report. This should actually fail for another reason - it should complain about `A` not being exported despite being used in data constructor `B2`.

hdgarrood:
Not sure if this is the same bug, but the following will fail at runtime with "undefined is not a function":

```purescript
module Main where

import Prelude
import Control.Monad.Eff.Console
import Dep

main = print ok
```
```purescript
module Dep (Ok(), ok) where

import Prelude

data Secret a = Secret a

type Ok = Secret Int

instance showSecret :: (Show a) => Show (Secret a) where
  show (Secret x) = "Secret " <> show x

ok :: Ok
ok = Secret 0
```

paf31:
This is fixed by #1480, but only because the externs file is just assumed to be correct now. I guess the issue is in `Names`?

paf31:
The issue reported by @hdgarrood is still present, however. I guess both of these can be summarized as "there should be a transitive export error thrown", right?

paf31:
@garyb Is this fixed now?

garyb:
Yep

```
Error in module Dep:
An export for Ok requires the following to also be exported:
Secret
```

Issue #1179, opened 2015-06-18
psc-make generates invalid externs file in presence of un-exported synonyms


qxjit:
When a module exports a type synonym but fails to export synonyms used in its definition, psc-make generates an externs file with references to the un-exported synonyms. On subsequent psc-make runs the un-exported synonym is reported as an unknown type since it is not found in the externs file.

Presumably the first build should fail because of the un-exported synonym in the first place?

garyb:
Yep, if it fails on subsequent builds it should fail the first time too. 

Just checking I understand correctly, it's when you have something like this:

``` purescript
module Test (X()) where

type X = Y
type Y = String

```

?

qxjit:
You're too fast! Here's a minimal example: https://github.com/flipstone/purescript_issue_1179

garyb:
Ah perfect, thanks!

paf31:
Possible duplicate of #1318?

paf31:
Can we close this now?

garyb:
Yep!

Issue #959, opened 2015-03-13
Check all required types are exported


garyb:
It seems there is still a problem with this, discovered when I was hacking down QC:

``` purescript
module Test.QuickCheck.Gen
  ( Gen()
  , uniform
  , perturbGen
  , repeatable

  , GenState()
  , GenOut()
  , runGen
  , evalGen

  -- it should be an error not to export these, as they're used in `GenState`
  -- but it's not an error until trying to recompile as they're missing from the externs
  -- , Size()
  -- , LCG()

  , stateful
  ) where
```

paf31:
Possibly the same as #1179, #1318?

garyb:
Not quite - this didn't generate invalid externs, it just worked even though in theory there's a check to disallow it.

I'm thinking maybe we should remove the check now though, or relax it, as there are actually times where not exporting a type that is used in a type synonym is actually perfectly okay.

garyb:
Irrelevant as of #1525

garyb:
This was resolved in #1529. Still want to revisit the idea of being able to use "abstract type synonyms", but at least this particular check works as it should now.

PR #1529, opened 2015-10-15
Revert "Don't require that all types are exported"

garyb:
Reverts purescript/purescript#1525

### DIFF
--- a/examples/failing/TransitiveDctorExport.purs
+++ b/examples/failing/TransitiveDctorExport.purs
@@ -1,0 +1,5 @@
+-- @shouldFailWith TransitiveExportError
+module Main (Y(..)) where
+
+type X = Int
+data Y = Y X
--- a/examples/failing/TransitiveSynonymExport.purs
+++ b/examples/failing/TransitiveSynonymExport.purs
@@ -1,0 +1,5 @@
+-- @shouldFailWith TransitiveExportError
+module Main (Y()) where
+
+type X = Int
+type Y = X
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -280,20 +280,35 @@ typeCheckModule (Module ss coms mn decls (Just exps)) = warnAndRethrow (addHint 
   where
 
   checkMemberExport :: (Type -> [DeclarationRef]) -> DeclarationRef -> Check ()
+  checkMemberExport extract dr@(TypeRef name dctors) = do
+    env <- getEnv
+    case M.lookup (Qualified (Just mn) name) (typeSynonyms env) of
+      Nothing -> return ()
+      Just (_, ty) -> checkExport dr extract ty
+    case dctors of
+      Nothing -> return ()
+      Just dctors' -> forM_ dctors' $ \dctor ->
+        case M.lookup (Qualified (Just mn) dctor) (dataConstructors env) of
+          Nothing -> return ()
+          Just (_, _, ty, _) -> checkExport dr extract ty
+    return ()
   checkMemberExport extract dr@(ValueRef name) = do
     ty <- lookupVariable mn (Qualified (Just mn) name)
-    case filter (not . exported) (extract ty) of
-      [] -> return ()
-      hidden -> throwError . errorMessage $ TransitiveExportError dr hidden
-      where
-      exported e = any (exports e) exps
-      exports (TypeRef pn1 _) (TypeRef pn2 _) = pn1 == pn2
-      exports (ValueRef id1) (ValueRef id2) = id1 == id2
-      exports (TypeClassRef pn1) (TypeClassRef pn2) = pn1 == pn2
-      exports (PositionedDeclarationRef _ _ r1) r2 = exports r1 r2
-      exports r1 (PositionedDeclarationRef _ _ r2) = exports r1 r2
-      exports _ _ = False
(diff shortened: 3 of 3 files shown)


garyb:
I'll append the fix for nested synonyms to this branch before merging.

paf31:
Looks good, thanks!

garyb:
Ok I see where the problems come from with this now - we're only looking at the export list and not the actual declarations, hence not following nested synonyms - I'll see about fixing it tomorrow.

paf31:
I think only considering the exports is probably the right thing to do, since the externs should represent the public API of the module (I know this is still up for debate to some extent). Let's discuss a bit more before committing to anything for 0.7.5?

garyb:
Considering the exports _list_ isn't enough though, that's why cases like this:

``` purescript
module Test (X()) where

type X = Y
type Y = Int
```

are not picked up by the current warning.

paf31:
But the purpose of the externs files should be to represent the public API that a module provides, and since `Y` is a part of that right now (I know it might not be in a later version, depending on what we decide about synonyms, but for now...) it should be in externs. So I think that either that exports list should give an error, or (better) automatically desugared in `Names` to include `Y`.

garyb:
That's what I'm talking about - the current check is not good enough to throw the `TransitiveExportError` in the case I mentioned above. I'm not trying to change the behaviour to support a new use case here, this is fixing the existing check so that it actually works properly.

paf31:
Ah I see. You're referring to the check in `Names` then?

I just want to make sure we're not planning on requiring any information not in externs, since obviously we can't rely on that being around in the separate compilation case any more.

garyb:
Yep - I'm just saying that since we're keeping this check for now until we discuss abstraction with synonyms etc. it's probably a good time to fix the various problems with it :smile:

e.g. #1318 and #1179  (although these don't have the problem with invalid externs files now, but they still are wrong as the transitive export check should have rejected them in the first place) and #959

paf31:
:+1: Sounds good to me.

paf31:
@garyb What's the status of this? Are you still planning to update it before 0.7.5, or do you want to merge this now and push any fixes into a patch release?

garyb:
Yep, I'd still like to get it fixed for 0.7.5. Will do so today hopefully.

garyb:
I'm not too sure what to do about this now. The check works even less well than ever because of the change to desugaring:

``` purescript
module Fail1 (fail) where

type X = Int

fail :: X -> X
fail a = a

```
becomes:

``` purescript
module Fail1 (fail) where

type X = Int

fail :: Int -> Int
fail a = a
```

By the time we make this check, so it is no longer able to complain about `X` not being exported.

paf31:
Where are you seeing that desugaring?

garyb:
That's what I see when I trace `fail`'s type from `Environment` when compiling `Fail1`.

I've made improvements anyway - at least it propery requires nested synonyms to be exported and synonyms referenced by data constructors now. Both cases were previously were unchecked.

garyb:
Also, here's the externs file for `Fail1`:

``` json
{
   "efExports":[
      {
         "ValueRef":{
            "Ident":"fail"
         }
      }
   ],
   "efImports":[
      {
         "eiImportedAs":null,
         "eiModule":[
            {
               "runProperName":"Prim"
            }
         ],
         "eiImportType":{
            "Implicit":[

            ]
         }
      }
   ],
   "efVersion":"0.7.4.1",
   "efDeclarations":[
      {
         "EDValue":{
            "edValueType":{
               "tag":"TypeApp",
               "contents":[
                  {
                     "tag":"TypeApp",
                     "contents":[
                        {
                           "tag":"TypeConstructor",
                           "contents":"Prim.Function"
                        },
                        {
                           "tag":"TypeConstructor",
                           "contents":"Prim.Int"
                        }
                     ]
                  },
                  {
                     "tag":"TypeConstructor",
                     "contents":"Prim.Int"
                  }
               ]
            },
            "edValueName":{
               "Ident":"fail"
            }
         }
      }
   ],
   "efFixities":[

   ],
   "efModuleName":[
      {
         "runProperName":"Fail1"
      }
   ]
}
```

paf31:
Ah I always assumed this would happen in `Names`, hence pre-desugaring. Looks good though.

garyb:
Oh right. Yeah, we can't do it there as it the check needs to look at members that have inferred types too. Good point though, maybe I should check twice?

garyb:
Interesting that synonyms aren't desugared in the arguments of data constructors though - maybe they should be too?

paf31:
I don't think we should consider this "desugaring", it's part of type elaboration which happens on types of terms, so it doesn't happen during kind checking, hence not on data constructors.

This is why I'm a little wary of using the `Environment` for things other than type checking.

garyb:
Yeah, desugaring isn't really what I meant - I guess my question was more "if we have to do this for values to get synonym things to work out properly, why not constructors"?

garyb:
I'll merge this once the tests are passed though, I'm happy with it for 0.7.5 as we're going to be looking at the rules for synonyms again anyway. Anything else we need for 0.7.5? I'm looking forward to it being out there :)

paf31:
I don't think so, if you're happy with this, then I think let's release 0.7.5.

garyb:
About time Travis!

:shipit: :ship: :shipit:

PR #1495, opened 2015-09-29
Typed Binders

5outh:
Adds typed binders (#1227)

PR #1530, opened 2015-10-15
Typed binders

5outh:
Creating a new pull request for typed binders, because the other one had Travis issues. See #1495 for more discussion


paf31:
Thanks! Let's see how Travis manages this time. We should also test pretty-printing with the new types pretty-printer, with something like a record type which will span multiple lines.

5outh:
Looks like travis failed again with the same issue. I need to pull in the most recent changes to master as well

5outh:
@paf31 I was able to replicate this problem locally by updating master!

paf31:
What's the issue?

5outh:
@paf31 I hadn't merged with `master`; When I did merge, the import `prettyPrintBinder` wasn't being imported. I just had to add the import and everything worked. 

Also, it looks like you've changed some of the pretty printing to output `Box`es. I think I could make this change for the binders, but do you think this should go in this PR or a separate PR?

paf31:
Well, pretty printing a binder with the current code will result in incorrect layout unfortunately.

The reason I didn't use `boxes` for binders was that it wasn't clear to me how to nicely pretty-print a case statement whose input spanned multiple lines, but I think we probably have to change to `boxes` now anyway.

I would say, moving to `boxes` would be nice to have, but it's not a big deal if you'd rather we tackle it in a separate PR. I could help out with that if you like.

paf31:
A compromise might be to just not print the type annotation, like we do with `TypedValue`. Actually, that's probably the neatest option for now.

Oh, and one other thing: I don't see a new case in the exhaustivity checker - can you verify that a typed binder gets checked correctly. You might need to just add a pass-through case there.

Thanks!

5outh:
Okay, I think I handled both of those cases. I removed the type annotation from the pretty print case and added the pass-through case for `TypedBinder`.

paf31:
Oops, just spotted three more, sorry:

- `isVarBinder`
- `fromVarBinder`
- `binderRequiresMonotype`

but I think that's it.

Could you please add those cases, and then I'll merge. Thanks!

5outh:
@paf31 Thanks for catching that! `binderRequiresMonotype` is `True` for a typed binder I think, since we're not performing the subsumption check in `inferBinder` (if I recall correctly). Totally missed the other two, though.

paf31:
So let's work through this:

`binderRequiresMonotype` determines whether or not to instantiate the type arguments of the type of the argument to a binder before typechecking it.

So since you use `parseType` and not `parsePolyType`, polymorphic types are not allowed syntactically, and you're right, we should always instantiate.

We should probably add a TODO in a good place so that we don't forget to change this when we add support for polymorphic types.

5outh:
Sounds good, I'll add the TODO. Thanks for all your help!

paf31:
Looks great, thanks! I'll wait for Travis, then merge.

paf31:
Thanks!

kRITZCREEK:
:+1:

PR #1531, opened 2015-10-16
Include stack files in the distribution

zudov:
https://twitter.com/ndm_haskell/status/654764513221185537

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -29,6 +29,10 @@ extra-source-files: examples/passing/*.purs
                   , stack-lts-2.yaml
                   , stack-lts-3.yaml
                   , stack-nightly.yaml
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Looks good, thanks!

zudov:
How about we also include README?

paf31:
The markdown file? I don't see any reason not to include it.

Does the LICENSE file already get included? That and CONTRIBUTORS.md might be worth adding..

zudov:
`LICENSE` is included since it's mentioned in `license:` field.

I'll add the rest.

zudov:
Should be ready to merge now if no additions are wanted.

paf31:
:+1: Thanks!

PR #1532, opened 2015-10-16
Don't deploy from coverage build

### LINKED ISSUE #1451: *.tix files from binaries
Seems like whenever I run `psc`, or `psci`, it leaves something like `psc.tix` file relative to the where I ran the command. Seems similar to the issue here https://github.com/purescript/purescript/issues/697

paf31:
Resolves #1451 

/cc @garyb


garyb:
:+1: 

Do we even use the coverage stuff anymore? I've not seen a peep out of Coveralls for a long time now...

paf31:
I think it still gets run, but doesn't post comments any more for some reason.

PR #1533, opened 2015-10-16
Fix #881, improve error for out-of-order declarations in instance

### LINKED ISSUE #881: Improve error message for class member implementations that are not contiguous
This:
``` purescript
instance eqMaybe :: (Eq a) => Eq (Maybe a) where
  (==) Nothing   Nothing   = true
  (==) (Just a1) (Just a2) = a1 == a2
  (/=) a b = not (a == b)
  (==) _ _ = false -- <------------- oops, out of sequence
```
Raises this error:
```
>> Compiling Data.Maybe
>> Error at src/Data/Maybe.purs line 65, column 1 - line 71, column 1:
>> Error in declaration eqMaybe
>> Duplicate property names
```

Which is caught in `ensureNoDuplicateProperties` during type checking instead of somewhere more sensible.

paf31:
/cc @garyb

### DIFF
--- a/examples/failing/881.purs
+++ b/examples/failing/881.purs
@@ -1,0 +1,13 @@
+-- @shouldFailWith DuplicateValueDeclaration
+module Main where
+
+data X = X | Y
+
+class Foo a where
+  foo :: a -> a
+  bar :: a
+
+instance fooX :: Foo X where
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

PR #1540, opened 2015-10-18
Fix #1535

### LINKED ISSUE #1535: Warnings in master not showing
Or at least partiality warnings are not raised, and nor are extraneous type variable warnings.

paf31:
/cc @garyb 

Seems `fork` in `lifted-base` doesn't keep `Writer` state around, which totally makes sense.


garyb:
Phew! :smile: :+1: :+1: :+1: :+1: :+1:

Issue #1509, opened 2015-10-05
Warn when using "open" imports


garyb:
Include a copy-paste-able fix in the warning message that provides a version of the import with the used members selectively listed.

I had been thinking we could change the syntax for open imports and require `import SomeModule (..)`, but upon reflection that's kinda arbitrary breakage - the warning that states its own fix should be sufficient.

hdgarrood:
Sorry if I'm a bit late to this discussion, but I'm not convinced that open imports should always be discouraged. I remember reading that the GHC team think open imports are fine, and often preferable, if you're importing a module in the same package.

garyb:
Yeah, importing within the same package is a good case for allowing them actually. I guess we could allow open imports of a module name with the same prefix as the current module, but that's not quite the same thing - it would at least work for cases where we're re-exporting a bunch of inner modules though.

Thoughts @paf31?

paf31:
I really like the idea of warning on unqualified whole module imports, but only if we can suggest the minimal imports.

I think qualified imports are enough of a fix for package-internal whole imports, no?

garyb:
Yeah, I think so. I had in mind something like the warning would suggest a list below a certain number of imports, and qualification if something is heavily used. This might be a super annoying warning though :smile:

hdgarrood:
The other thing is that modules like `Data.Maybe`, `Data.Tuple`, `Data.Either`, and of course `Prelude` are pretty much always imported open and I think it might be annoying to have either to import those qualified or maintain an explicit import list.

If not being able to work out where an identifier comes from is a problem I'd rather we address that through tooling like `psc-ide`, with a "display which module this identifier comes from" feature or something.

hdgarrood:
Has opting in or out of individual warnings been discussed anywhere, incidentally?

michaelficarra:
@hdgarrood There are other use cases that benefit from explicit imports, such as browsing code on GitHub.

hdgarrood:
That's true. If we eventually copy Hackage and have highlighted source code on Pursuit with hyperlinked identifiers, do you think that would solve that issue?

Even so, there are a fairly large set of extremely commonly used types and functions that I don't think need explicit imports.

[...]

PR #1545, opened 2015-10-21
Give warnings on unused imports (implicit and explicit)

nwolverson:
WIP presented for comments/direction. Still to do:

- [x] Exclude `Prim`!
- [x] `bind` - anything else from desugaring?
- [x] Qualified imports excluded (should be easy)
- [x] Re-exports
- [ ] Tests - is there a way to test warnings are generated?

Originally was more self-contained but working out usages is a little interesting.


src/Language/PureScript/Sugar/Names.hs
```diff
@@ -137,17 +139,22 @@ elaborateImports imps (Module ss coms mn decls exps) = Module ss coms mn decls'
...
+renameInModule env imports m@(Module ss coms mn decls exps) = do
+  (decls', used) <- flip runStateT M.empty $ parU decls go
+  findUnusedImports m imports used
+  return $ Module ss coms mn decls' exps
   where
+
   (go, _, _, _, _) = everywhereWithContextOnValuesM (Nothing, []) updateDecl updateValue updateBinder updateCase defS
 
-  updateDecl :: (Maybe SourceSpan, [Ident]) -> Declaration -> m ((Maybe SourceSpan, [Ident]), Declaration)
+  updateDecl :: (Maybe SourceSpan, [Ident]) -> Declaration -> StateT UsedImports m ((Maybe SourceSpan, [Ident]), Declaration)
```
garyb: Instead of using `StateT` explicitly everywhere here, why not add a `MonadState UsedImports m` constraint to `renameInModule`? You won't have to `lift` the state operations then either.

nwolverson: Thanks, I knew I was missing something, so much simpler.

nwolverson:
Should address #1392 now. Should be good setup for #1509

src/Language/PureScript/Errors.hs
```diff
@@ -687,6 +691,11 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
       paras [ line "An exhaustivity check was abandoned due to too many possible cases."
             , line "You may want to decomposing your data types into smaller types."
             ]
+    renderSimpleErrorMessage (UnusedImport name) =
+      line $ "Unused import of module " ++ runModuleName name
+
+    renderSimpleErrorMessage (UnusedExplicitImport name names) =
+      line $ "Unused identifiers in import of module " ++ runModuleName name ++ ": " ++ intercalate ", " names
```
paf31: I would prefer to put the names in an indented block with each name on its own line, to make use of the space.

garyb:
I'll review again too when I get a minute.

src/Language/PureScript/Errors.hs
paf31: Also, I've been trying to reword some of the errors so they read more like full sentences rather than fragments lately. So maybe something like "The import of <module name> contains the following redundant references:"

paf31:
I think @garyb would be the best person to review this, but I'll just say: could you please add any license terms to `CONTRIBUTORS.md`? Thanks.

src/Language/PureScript/Errors.hs
```diff
@@ -687,6 +691,12 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
       paras [ line "An exhaustivity check was abandoned due to too many possible cases."
             , line "You may want to decomposing your data types into smaller types."
             ]
+    renderSimpleErrorMessage (UnusedImport name) =
+      line $ "The import of module " ++ runModuleName name ++ " is redundant"
+
+    renderSimpleErrorMessage (UnusedExplicitImport name names) =
+      paras [ line $ "The import of module " ++ runModuleName name ++ " contains the following redundant references:"
```
garyb: Perhaps we could phrase this as "contains the following unused references" instead?

garyb:
Aside from that comment on the wording of the warning - looks good to me!

nwolverson:
Made wording change, rebased & squashed.

garyb:
Thanks! It'll be great to have this in, and this implementation is pretty nice, I had briefly tried something when I was making other changes to the module code but it was much uglier.

Any further comments @paf31? I guess this will block or be blocked by #1556 depending on which we merge first. (I'll re-review that one later today, along with the other pending PRs we have - heading out for a bit now0.

garyb:
Sorry, looks like you'll need to rebase again :smile:

nwolverson:
No worries, rebased.

paf31:
Looks good to me! Can you please just add a comment on any new exported functions and types?

paf31:
@garyb I assume this will go into 0.7.6?

paf31:
Is this ready to merge then?

garyb:
I think so :+1:

garyb:
This needs rebasing yet again, sorry!

Also there is a slight problem: if you import a `Type(..)` and make use of its constructors but never use `Type` itself you get a warning that the import is unnecessary.

nwolverson:
Rebased, now handle the data constructor usage.

garyb:
Thanks! :+1: 

I've discovered a few other enhancements we could include but I'll open an issue for those.

nwolverson:
Brilliant! So happy to get my first compiler change in :)

PR #1546, opened 2015-10-21
Language.PureScript.Externs: fix haddock markup

trofi:
Haddock is unhappy about trailing '-- ^' comments
being used as leading.

Signed-off-by: Sergei Trofimovich <siarheit@google.com>


paf31:
:+1: Thanks! Does this need a new Hackage release for Stackage then?

PR #1552, opened 2015-10-23
Fix #1543

### LINKED ISSUE #1543: Printing of constrained types is incorrect
E.g.

```purescript
reify :: forall a r. a -> (Reifies a) => r -> r
```

should be

```purescript
reify :: forall a r. a -> ((Reifies a) => r) -> r
```

paf31:
/cc @garyb


garyb:
I'm less confident reasoning about this one just by reading the code, but the change seems sensible, and I assume you've checked the output regression is fixed, so :+1:!

paf31:
> I assume you've checked the output regression is fixed

yep, just can't really test it effectively, but it works properly in PSCi now.

PR #1554, opened 2015-10-23
Try to fix #1548

### LINKED ISSUE #1548: Thread block on recompile
```
psc: make: dependency should already have been built.
psc: thread blocked indefinitely in an MVar operation
```
Message after change in already compiled module. Removing `psc` output and recompile works. Error is raised non-determenistically.

paf31:
@cryogenian Could you please test this?


paf31:
This is the only real blocker to 0.7.5.1 going out, so if someone could test this (who could reproduce the original issue), I'd really appreciate it.

cryogenian:
Sure, will test

paf31:
Thanks @cryogenian !

cryogenian:
Fixed :+1:

paf31:
Excellent, thanks very much!

PR #1555, opened 2015-10-23
Fix #1534

### LINKED ISSUE #1534: Improvement to untyped top level declaration warning?
Instead of prompting to use `_` to infer the type can we not just print the inferred type straight away? Throwing that extra step in the way seems a little user unfriendly.

paf31:
/cc @garyb 

New warning (from `TypedBinders.purs`):

```text
Warning 2 of 3:

  Warning in value declaration runState:
  Warning in module Main:
  No type declaration was provided for the top-level declaration of runState.
  It is good practice to provide type declarations as a form of documentation.
  The inferred type of runState was:

    forall t29 t30. t30 -> State t30 t29 -> Tuple t30 t29

Warning 3 of 3:

  Warning in value declaration main:
  Warning in module Main:
  No type declaration was provided for the top-level declaration of main.
  It is good practice to provide type declarations as a form of documentation.
  The inferred type of main was:

    forall t112. Eff ( console :: CONSOLE
                     | t112
                     )
                 Unit
```

Much nicer, I think.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -142,7 +142,7 @@ data SimpleErrorMessage
   | ShadowedTypeVar String
   | UnusedTypeVar String
   | WildcardInferredType Type
-  | MissingTypeDeclaration Ident
+  | MissingTypeDeclaration Ident Type
   | NotExhaustivePattern [[Binder]] Bool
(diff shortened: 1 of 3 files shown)


garyb:
:heart_eyes: :+1:

PR #1556, opened 2015-10-24
Friendlier errors

paf31:
/cc @garyb 

This changes the wording of quite a few errors, either to make them easier to understand, or to turn them into full sentences rather than fragments.

I'm not sure about some of them, but I think it's an improvement overall.

I also

- Use `error` in place of `throwError` for errors which are really compiler bugs
- Moved the `suggestions` inside the branches of `prettyPrintErrorMessage`
- Got rid of the object accessor/function suggestion, since it rarely worked, and I have an idea for a better solution.

Thoughts?

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -42,7 +42,6 @@ import Control.Monad.Trans.State.Lazy
 import Control.Arrow(first)
 
 import Language.PureScript.AST
-import Language.PureScript.Environment (isObject, isFunction)
 import Language.PureScript.Pretty
 import Language.PureScript.Types
 import Language.PureScript.Names
@@ -62,13 +61,11 @@ data SimpleErrorMessage
   | MissingFFIModule ModuleName
   | MultipleFFIModules ModuleName [FilePath]
   | UnnecessaryFFIModule ModuleName FilePath
-  | InvalidExternsFile FilePath
   | CannotGetFileInfo FilePath
   | CannotReadFile FilePath
   | CannotWriteFile FilePath
   | InfiniteType Type
   | InfiniteKind Kind
-  | CannotReorderOperators
   | MultipleFixities Ident
   | OrphanTypeDeclaration Ident
   | OrphanFixityDeclaration String
@@ -112,7 +109,6 @@ data SimpleErrorMessage
   | UndefinedTypeVariable ProperName
   | PartiallyAppliedSynonym (Qualified ProperName)
   | EscapedSkolem (Maybe Expr)
-  | UnspecifiedSkolemScope
   | TypesDoNotUnify Type Type
   | KindsDoNotUnify Kind Kind
   | ConstrainedTypeUnified Type Type
@@ -126,10 +122,9 @@ data SimpleErrorMessage
   | ArgListLengthsDiffer Ident
   | OverlappingArgNames (Maybe Ident)
   | MissingClassMember Ident
-  | ExtraneousClassMember Ident
+  | ExtraneousClassMember Ident (Qualified ProperName)
   | ExpectedType Type Kind
   | IncorrectConstructorArity (Qualified ProperName)
-  | SubsumptionCheckFailed
   | ExprDoesNotHaveType Expr Type
   | PropertyIsMissing String Type
   | CannotApplyFunction Type Expr
@@ -202,13 +197,11 @@ errorCode em = case unwrapErrorMessage em of
   MissingFFIModule{} -> "MissingFFIModule"
   MultipleFFIModules{} -> "MultipleFFIModules"
   UnnecessaryFFIModule{} -> "UnnecessaryFFIModule"
-  InvalidExternsFile{} -> "InvalidExternsFile"
   CannotGetFileInfo{} -> "CannotGetFileInfo"
   CannotReadFile{} -> "CannotReadFile"
   CannotWriteFile{} -> "CannotWriteFile"
   InfiniteType{} -> "InfiniteType"
   InfiniteKind{} -> "InfiniteKind"
-  CannotReorderOperators -> "CannotReorderOperators"
   MultipleFixities{} -> "MultipleFixities"
   OrphanTypeDeclaration{} -> "OrphanTypeDeclaration"
   OrphanFixityDeclaration{} -> "OrphanFixityDeclaration"
@@ -252,7 +245,6 @@ errorCode em = case unwrapErrorMessage em of
   UndefinedTypeVariable{} -> "UndefinedTypeVariable"
   PartiallyAppliedSynonym{} -> "PartiallyAppliedSynonym"
   EscapedSkolem{} -> "EscapedSkolem"
-  UnspecifiedSkolemScope -> "UnspecifiedSkolemScope"
   TypesDoNotUnify{} -> "TypesDoNotUnify"
   KindsDoNotUnify{} -> "KindsDoNotUnify"
   ConstrainedTypeUnified{} -> "ConstrainedTypeUnified"
@@ -269,7 +261,6 @@ errorCode em = case unwrapErrorMessage em of
   ExtraneousClassMember{} -> "ExtraneousClassMember"
   ExpectedType{} -> "ExpectedType"
   IncorrectConstructorArity{} -> "IncorrectConstructorArity"
-  SubsumptionCheckFailed -> "SubsumptionCheckFailed"
   ExprDoesNotHaveType{} -> "ExprDoesNotHaveType"
   PropertyIsMissing{} -> "PropertyIsMissing"
   CannotApplyFunction{} -> "CannotApplyFunction"
@@ -385,9 +376,9 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
   prettyPrintErrorMessage (ErrorMessage hints simple) =
     paras $
       map renderHint hints ++
-      renderSimpleErrorMessage simple :
-      suggestions simple ++
-      [line $ "See " ++ wikiUri ++ " for more information, or to contribute content related to this " ++ levelText ++ "."]
+      [ renderSimpleErrorMessage simple
+      , line $ "See " ++ wikiUri ++ " for more information, or to contribute content related to this " ++ levelText ++ "."
+      ]
     where
     wikiUri :: String
     wikiUri = "https://github.com/purescript/purescript/wiki/Error-Code-" ++ errorCode e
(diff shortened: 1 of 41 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -410,37 +401,37 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
+            , indent $ line "// module A.B.C"
             ]
     renderSimpleErrorMessage (ErrorParsingModule err) =
       paras [ line "Unable to parse module: "
             , prettyPrintParseError err
             ]
     renderSimpleErrorMessage (MissingFFIModule mn) =
-      line $ "Missing FFI implementations for module " ++ runModuleName mn
+      paras [ line $ "The foreign module implementation for module " ++ runModuleName mn ++ " is missing."
+            , line $ "Provide your foreign module implementation using the --ffi command line option, and ensure that your module contains the following module header:"
```
garyb: A minor issue - errors are generic (shared across any potential backend) but with these messages the FFI errors become specific to the PureScript backend now.

src/Language/PureScript/Errors.hs
```diff
@@ -450,17 +441,17 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
-      line $ "Orphan fixity declaration for " ++ show op
+      line $ "The fixity/precedence declaration for " ++ show op ++ " should appear in the same module as its definition."
     renderSimpleErrorMessage (RedefinedModule name filenames) =
-      paras [ line ("Module " ++ runModuleName name ++ " has been defined multiple times:")
+      paras [ line ("The module " ++ runModuleName name ++ " has been defined multiple times:")
             , indent . paras $ map (line . displaySourceSpan) filenames
             ]
     renderSimpleErrorMessage (RedefinedIdent name) =
-      line $ "Name " ++ showIdent name ++ " has been defined multiple times"
+      line $ "The name " ++ showIdent name ++ " has been defined multiple times"
```
garyb: "The value" perhaps?

src/Language/PureScript/Errors.hs
```diff
@@ -474,25 +465,30 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
     renderSimpleErrorMessage (UnknownDataConstructor dc tc) =
       line $ "Unknown data constructor " ++ showQualified runProperName dc ++ foldMap ((" for type constructor " ++) . showQualified runProperName) tc
     renderSimpleErrorMessage (UnknownImportType mn name) =
-      line $ "Module " ++ runModuleName mn ++ " does not export type " ++ runProperName name
+      line $ "The type " ++ runProperName name ++ " is not exported by the module " ++ runModuleName mn
```
garyb: How about phrasing this one like the `UnknownExportModule` error. Something along the line of

```
Cannot import the type X from module Y
It either does not exist or the module does not export the type
```

In an ideal world also followed by:

```
Did you mean:
  X'
  X''
  X'''
```

:wink: But that's for a later date I think.

src/Language/PureScript/Errors.hs
```diff
@@ -474,25 +465,30 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
     renderSimpleErrorMessage (UnknownExportValue name) =
       line $ "Cannot export unknown value " ++ showIdent name
     renderSimpleErrorMessage (UnknownExportModule name) =
-      line $ "Cannot export unknown module " ++ runModuleName name ++ ", it either does not exist or has not been imported by the current module"
+      paras [ line $ "Cannot export unknown module " ++ runModuleName name
+            , line "It either does not exist or has not been imported by the current module."
+            ]
     renderSimpleErrorMessage (UnknownImportDataConstructor mn tcon dcon) =
-      line $ "Module " ++ runModuleName mn ++ " does not export data constructor " ++ runProperName dcon ++ " for type " ++ runProperName tcon
+      line $ "The data constructor " ++ runProperName dcon ++ " (for type " ++ runProperName tcon ++ ") is not exported by the module " ++ runModuleName mn
```
garyb: Maybe it's just me but I think this was easier to parse before.

src/Language/PureScript/Errors.hs
```diff
@@ -500,33 +496,41 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
-      line $ "Module " ++ runModuleName mn ++ " has been defined multiple times."
+      line $ "The module " ++ runModuleName mn ++ " has been defined multiple times."
     renderSimpleErrorMessage (DuplicateClassExport nm) =
       line $ "Duplicate export declaration for type class " ++ runProperName nm
     renderSimpleErrorMessage (DuplicateValueExport nm) =
       line $ "Duplicate export declaration for value " ++ showIdent nm
     renderSimpleErrorMessage (CycleInDeclaration nm) =
-      line $ "Cycle in declaration of " ++ showIdent nm
+      paras [ line $ "The declaration of " ++ showIdent nm ++ " contains a cycle."
+            , line $ "The value of " ++ showIdent nm ++ " may not be defined here, so this reference is not allowed."
```
garyb: I'm not sure this second line helps - it's just a reiteration of the error, I think just:

```
The declaration of x contains a cycle.
You might be able to evaluate this expression lazily, using something like Control.Lazy.fix
```

Would be better. Or if you think further elaboration is needed state something about the declaration being infinitely recursive / non-terminating.

src/Language/PureScript/Errors.hs
```diff
@@ -500,33 +496,41 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
+    renderSimpleErrorMessage (CycleInTypeSynonym name) =
+      paras [ line $ case name of
+                       Just pn -> "A cycle appears in the definition of the type synonym " ++ runProperName pn
+                       Nothing -> "A cycle appears in a set of type synonym definitions."
+            , line "Cycles are disallowed because they can lead to loops in the type checker."
+            , line "Consider using a 'newtype' instead."
+            ]
     renderSimpleErrorMessage (NameIsUndefined ident) =
-      line $ showIdent ident ++ " is undefined"
+      line $ "The name " ++ showIdent ident ++ " is undefined."
```
garyb: "The value" again? I think everywhere we use "the name" when referring to values is probably better?

src/Language/PureScript/Errors.hs
```diff
@@ -500,33 +496,41 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
     renderSimpleErrorMessage (NameNotInScope ident) =
-      line $ showIdent ident ++ " may not be defined in the current scope"
+      line $ "The name " ++ showIdent ident ++ " may not be defined in the current scope."
```
garyb: The phrasing of this is and was ambiguous: it's sounds like `ident` is not allowed to be defined in the current scope. Also there's no "may" about it - the value is definitely not defined in the current scope :smile:

src/Language/PureScript/Errors.hs
```diff
@@ -535,25 +539,25 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
                                           , indent $ prettyPrintValue expr
                                           ]) binding
     renderSimpleErrorMessage (TypesDoNotUnify t1 t2)
-      = paras [ line "Cannot unify type"
+      = paras [ line "The type"
               , indent $ typeAsBox t1
-              , line "with type"
+              , line "cannot be made equal to the type"
```
garyb: I'm really not sure about that wording ("equal")! I guess it's not wrong, as that's what the process of applying the substitution returned by unification is, but it sounds weird to me...

The biggest problem people seem to have with this error message is they don't know which way the types go: is `t1` the type we have and `t2` the type it should be? Or vice versa. I think Haskell does a pretty good job of making it clear:

```
Couldn't match expected type X with actual type Y
```

src/Language/PureScript/Errors.hs
```diff
@@ -535,25 +539,25 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
-      paras [ line "Cannot unify constrained type"
+      paras [ line "The constrained type"
             , indent $ typeAsBox t1
-            , line "with type"
+            , line "cannot be made equal to the kind"
             , indent $ typeAsBox t2
             ]
     renderSimpleErrorMessage (OverlappingInstances nm ts (d : ds)) =
-      paras [ line "Overlapping instances found for"
+      paras [ line "Overlapping type class instances found for"
```
garyb: Do we need to add a one line explanation of why this is bad perhaps? Or is that one for the wiki.

src/Language/PureScript/Errors.hs
```diff
@@ -562,43 +566,43 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
             , indent $ Box.hsep 1 Box.left [ line (showQualified runProperName nm)
                                            , Box.vcat Box.left (map typeAtomAsBox ts)
                                            ]
             ]
     renderSimpleErrorMessage (CannotFindDerivingType nm) =
-      line $ "Cannot derive instance, because the type declaration for " ++ runProperName nm ++ " could not be found."
+      line $ "Cannot derive a type class instance, because the type declaration for " ++ runProperName nm ++ " could not be found."
     renderSimpleErrorMessage (DuplicateLabel l expr) =
-      paras $ [ line $ "Duplicate label " ++ show l ++ " in row." ]
+      paras $ [ line $ "The duplicate label " ++ show l ++ " appears in a row type." ]
```
garyb: `The label x appears more than once in a row type`?

src/Language/PureScript/Errors.hs
```diff
@@ -608,29 +612,28 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
             , indent $ prettyPrintValue expr
             , line "does not have type"
             , indent $ typeAsBox ty
             ]
     renderSimpleErrorMessage (PropertyIsMissing prop row) =
-      paras [ line "Row"
+      paras [ line "The row type"
             , indent $ prettyPrintRowWith '(' ')' row
-            , line $ "lacks required property " ++ show prop
+            , line $ "lacks the required property " ++ show prop
```
garyb: Should we call it a label again here rather than a property?

src/Language/PureScript/Errors.hs
```diff
@@ -608,29 +612,28 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
             ]
     renderSimpleErrorMessage (CannotApplyFunction fn arg) =
-      paras [ line "Cannot apply function of type"
+      paras [ line "A function of type"
             , indent $ typeAsBox fn
-            , line "to argument"
+            , line "can not be applied to the argument"
```
garyb: "...to an argument of type..."?

src/Language/PureScript/Errors.hs
```diff
@@ -639,29 +642,31 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
             , line "Consider moving the instance, if possible, or using a newtype wrapper."
             ]
     renderSimpleErrorMessage InvalidNewtype =
-      line "Newtypes must define a single constructor with a single argument"
+      line "Newtypes must define a single constructor with a single argument."
```
garyb: Do we need to include the name of the newtype here, or is that obvious in context?

src/Language/PureScript/Errors.hs
```diff
@@ -639,29 +642,31 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
...
     renderSimpleErrorMessage (InvalidInstanceHead ty) =
-      paras [ line "Invalid type in class instance head:"
+      paras [ line "An invalid type appears in a type class instance head:"
             , indent $ typeAsBox ty
             ]
```
garyb: Can we explain this further? Like actually enumerate the rules for what is considered to be an acceptable type in an instance head. Also, this error makes it sound like the type itself is somehow invalid rather than the instance head...

"Type class instance head is invalid due to the use/appearance/... of type x`?

src/Language/PureScript/Errors.hs
```diff
@@ -705,9 +710,9 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage . positionHintsFir
       paras [ lineWithLevel $ "in module " ++ runModuleName mn ++ ":"
             ]
     renderHint (ErrorInSubsumption t1 t2) =
-      paras [ lineWithLevel "checking that type "
+      paras [ lineWithLevel "checking that the type"
             , indent $ typeAsBox t1
-            , line "subsumes type"
+            , line "is more general than the type"
```
garyb: "at least as general"? Or is it strictly more general?

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -50,7 +49,7 @@ subsumes' val ty1 (ForAll ident ty2 sco) =
       sko <- newSkolemConstant
       let sk = skolemize ident sko sco' ty2
       subsumes val ty1 sk
-    Nothing -> throwError . errorMessage $ UnspecifiedSkolemScope
+    Nothing -> error "subsumes: unspecified skolem scope"
```
garyb: Makes sense to move these to `error`, but perhaps we should make these messages more friendly on the off chance they do appear too. Perhaps we could make a `fatalError` function that wraps errors:

```
An internal error ocurred during compilation: "subsumes: unspecified skolem scope"
```

Something along those lines anyway? So it's not so much that error itself needs to be friendly, but just there's some kind of context that is readable by non-compiler-developers.

garyb:
I like it!

The only other general comment I'd have: I think sprinkling "the" everywhere makes the errors harder to read than before, some of the phrasing gets really awkward ("inferring the type of the expression") and in most cases it just adds more stuff that you need to mentally ignore when scan-reading the errors - which is almost always the case for how they will be read. Succinct errors with slightly more wordy hints/explanation is the way to go in my opinion.

paf31:
I think you're right about these. Thanks for the pointers, I'll change those and run it by you again before merging.

Any thoughts on removing the `<<<` suggestion?

garyb:
:+1: for removing it - it's definitely better than it used to be, but still crops up in weird/misleading situations occasionally. Perhaps open an issue for the potential new version of it?

src/Language/PureScript/Errors.hs
paf31: It's the value though, not the type.

paf31:
@garyb  How does this look?

src/Language/PureScript/Errors.hs
garyb: Oops :)

garyb:
Looks excellent! :+1: Hopefully this will help with the comprehension of errors a great deal, and we can always refine further if some particular error is problematic still.

paf31:
Thanks for the review!

PR #1562, opened 2015-10-25
Fix #1557

### LINKED ISSUE #1557: Add error hint for property accessor checks
Something like "Error checking property accessor <expression>: ..." so that we can add a hint about `<<<` not being `.` like in Haskell, to replace the old suggestion, which would fire in incorrect cases.

paf31:
/cc @garyb 

Merging into `master` instead.


garyb:
:+1:

paf31:
Thanks!

PR #1565, opened 2015-10-25
Display hints after main error, resolve #1563

### LINKED ISSUE #1563: Format errors as Markdown?
I've been thinking about ways to format error messages which would break up the different parts of an error nicely. First of all, I added horizontal rules between parts, like so:

```text
Error found:
Error in module M1:
--------------------------------------------------------------------------------
Error in value declaration compose:
--------------------------------------------------------------------------------
Error at test.purs line 4, column 15 - line 4, column 19:
--------------------------------------------------------------------------------
Error checking type of property accessor

  g."f"

--------------------------------------------------------------------------------
Error checking that expression

  g

has type

  { f :: _0 -> _1
  | _2
  }

--------------------------------------------------------------------------------
Error while trying to match type

  Object

with type

  Function _3

--------------------------------------------------------------------------------
Could not match expected type

  Object

with actual type

  Function _3

--------------------------------------------------------------------------------
```

The main issue with this is that is doesn't deal nicely with multiple errors.

So another option is to format errors as Markdown, using bulleted lists to break up the parts, and indenting values and types by 4 chars, not 2, so that they are valid code sections. Editors could choose to render the markdown in a pretty format, but even on the terminal it should be easier to read. And it generalizes nicely to multiple errors.

As a plus, we can also copy and paste the errors into a GitHub issue.

Any thoughts?

## Sample (as text)

```text
- Error found:
- Error in module M1:
- Error in value declaration compose:
- Error at test.purs line 4, column 15 - line 4, column 19:
- Error checking type of property accessor

        g."f"

- Error checking that expression

        g

  has type

      { f :: _0 -> _1
      | _2
      }

- Error while trying to match type

        Object

  with type

      Function _3

- Could not match expected type

        Object

  with actual type

      Function _3
```

## Sample (formatted)

- Error found:
- Error in module M1:
- Error in value declaration compose:
- Error at test.purs line 4, column 15 - line 4, column 19:
- Error checking type of property accessor

        g."f"

- Error checking that expression

        g

  has type

      { f :: _0 -> _1
      | _2
      }

- Error while trying to match type

        Object

  with type

      Function _3

- Could not match expected type

        Object

  with actual type

      Function _3

paf31:
/cc @garyb 

Here is an example with verbose errors turned on now:

```text
Error 2 of 2:

  in module MultipleErrors
  at MultipleErrors.purs line 12, column 1 - line 13, column 1
  at MultipleErrors.purs line 12, column 9 - line 13, column 1

    Could not match expected type

      String

    with actual type

      Int


  while trying to match type String
    with type Int
  while checking that type String
    is at least as general as type Int
  while checking that expression "Test"
    has type Int
  while checking that expression case _1 of
                                   0 -> "Test"
                                   n -> foo (((-) n) 1)
    has type Int
  while checking that expression \_1 ->
                                   case _1 of
                                     0 -> "Test"
                                     n -> foo (((-) n) 1)
    has type Int -> Int
  in binding group foo, bar

```

### DIFF
--- a/examples/failing/MutRec.purs
+++ b/examples/failing/MutRec.purs
@@ -1,4 +1,5 @@
 -- @shouldFailWith CycleInDeclaration
+-- @shouldFailWith CycleInDeclaration
 module MutRec where
 
 import Prelude
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -56,8 +56,7 @@ import Text.Parsec.Error (Message(..))
 
 -- | A type of error messages
 data SimpleErrorMessage
-  = ErrorParsingExterns P.ParseError
-  | ErrorParsingFFIModule FilePath
+  = ErrorParsingFFIModule FilePath
   | ErrorParsingModule P.ParseError
   | MissingFFIModule ModuleName
   | MultipleFFIModules ModuleName [FilePath]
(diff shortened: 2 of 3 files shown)


garyb:
Cool :+1:

PR #1566, opened 2015-10-25
Fix #1169

### LINKED ISSUE #1169: Produce better error when a case does not match all of a data constructor's arguments

paf31:

Thanks.


garyb:
Should we add test(s) for this again too, just in case? Could use the [two here](https://github.com/purescript/purescript/issues/1169#issuecomment-148418504).

paf31:
Sounds good.

PR #1567, opened 2015-10-25
Fix #1315

### LINKED ISSUE #1315: Suggestions for row unification errors

paf31:

Thanks.

### DIFF
--- a/examples/failing/ExtraRecordField.purs
+++ b/examples/failing/ExtraRecordField.purs
@@ -1,5 +1,4 @@
--- @shouldFailWith PropertyIsMissing
--- TODO: Make this fail with a new error ExtraProperty instead.
+-- @shouldFailWith AdditionalProperty
 module ExtraRecordField where
 
 import Prelude ((<>))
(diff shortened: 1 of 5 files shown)


paf31:
Ok, this is ready for review.

hdgarrood:
I don't understand the stuff inside the type checker, but everything else :+1:

garyb:
Looks good to me - interesting place to catch the error too, I wasn't really sure how you were going to do this. I wonder about missing "row" out of the error message now, but I think it's probably fine, as labels only reside in rows, so :+1:

paf31:
Thanks for the reviews.

PR #1568, opened 2015-10-26
Fix #1534 again

### LINKED ISSUE #1534: Improvement to untyped top level declaration warning?

paf31:
Adding the types broke the "only at the top-level" change.

Let's definitely fix this before 0.7.5.1.


garyb:
Oops! :smile:

Looks good to me.

PR #1571, opened 2015-10-26
Fix #1570

### LINKED ISSUE #1570: Scoped type variables and type annotations issue
I think I found a bug:
```purescript
foo :: forall a. a -> a
foo = \(a :: a) -> a
```
`foo` throws "Error unifying type `a0` with type `a`".

paf31:

Thanks.

### DIFF
--- a/examples/failing/1570.purs
+++ b/examples/failing/1570.purs
@@ -1,0 +1,6 @@
+-- @shouldFailWith ExpectedType
+module M where
+
+data F a = F a
+
+test = \(x :: F) -> x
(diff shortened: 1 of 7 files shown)


paf31:
@garyb Actually, contrary to what I said before, I think this should probably go into 0.7.5.1 as well, since it's a patch for a 0.7.5 feature. But this should be the last one.

garyb:
Looks good :+1: and yep, I think it should be for 0.7.5.1.

paf31:
Thanks for the review. Shall we do some testing today and release this evening or tomorrow?

PR #1581, opened 2015-10-28
Fix #1072

### LINKED ISSUE #1072: Type class with an empty signature followed by a top-level term declaration confuses parser
Minimal example:

```purescript
class Foo x where

bar :: String
bar = "hello"
```

This fails to compile with
```
Error in module Main:
Error at src/Main.purs line 9, column 1 - line 26, column 1:
  Duplicate export declaration for value bar
See https://github.com/purescript/purescript/wiki/Error-Code-DuplicateValueExport for more information, or to contribute content related to this error.
```

The intention of the code is that `Foo` is a type class with no signature, and `bar` is unrelated to it, but based on that error message the compiler seems to be interpreting `bar` as both a top-level definition and a member of `Foo`'s signature?

Inserting an instance between them causes it to parse correctly:

```purescript
class Foo x where

instance numberFoo :: Foo Number

bar :: String
bar = "hello"
```

paf31:
@garyb Does this look ok?

### DIFF
--- a/examples/failing/EmptyClass.purs
+++ b/examples/failing/EmptyClass.purs
@@ -1,0 +1,7 @@
+-- @shouldFailWith ErrorParsingModule
+module Main where
+
+class Foo x where
+
+bar :: String
+bar = "hello"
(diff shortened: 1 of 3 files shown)


garyb:
Yep :+1:

PR #1583, opened 2015-10-29
Fix #1577, fix #1130

### LINKED ISSUE #1130: Better error message for type errors from non-Boolean guards
For example:

```
> data Foo = Bar String | Baz String
> case Bar "hi" of Bar x | Baz x -> x
Error in module $PSCI:
Error in value declaration it:
Error checking that type
  $PSCI.Foo
subsumes type
  Prim.Boolean
Error at  line 1, column 26 - line 1, column 32:
  Cannot unify type
    $PSCI.Foo
  with type
    Prim.Boolean
See https://github.com/purescript/purescript/wiki/Error-Code-ErrorUnifyingTypes for more information, or to contribute content related to this error.
```

This would be much clearer if we added something like:

> Hint: A pattern guard (that is, the part following the | character) must have type Boolean.

### LINKED ISSUE #1577: Actual vs Expected type in errors are often flipped
```purescript
module ErrorTest where

foo :: Int -> String
foo x = case x of
  1 -> "one"
  2 -> "two"
  3 -> 4
```
```
Error found:
in module ErrorTest
at ... line 7, column 8 - line 7, column 8

  Could not match expected type

    Int

  with actual type

    String


  with type String
while checking that expression 4
  has type String
in value declaration foo
```

It points to an Int, calls it a String, and says the expected type is Int.

paf31:
@garyb @natefaubion Could you please review? I'd rather just removed "expected" and "actual" since it's tricky to track which types actually correspond to real values in the program, and we should strive to give the best information in the checking judgment anyway, if possible.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -156,6 +156,7 @@ data ErrorMessageHint
   | ErrorCheckingAccessor Expr String
   | ErrorCheckingType Expr Type
   | ErrorCheckingKind Type
+  | ErrorCheckingGuard
   | ErrorInferringType Expr
   | ErrorInApplication Expr Type Expr
   | ErrorInDataConstructor ProperName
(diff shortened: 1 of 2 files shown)


garyb:
:+1: I guess, it's a shame we can't figure out "which way is up" as this is definitely one the errors I've seen people wrestle with most often, not knowing which is the type they have vs the expected type. Not much we can do though I suppose.

paf31:
Well, we could probably pass a bit of info along, something like

```purescript
data UnifyHint = LeftExpected | RightExpected | DontKnow
```

That might be better.

paf31:
But in any case, I'd rather just patch this now, and put a real fix into 0.9, because the unification code isn't set up for changes like this.

PR #1584, opened 2015-10-29
Fix #1582

### LINKED ISSUE #1582: Investigate overzealous exhaustivity bail-out
As discussed on IRC yesterday.

An example case: https://github.com/purescript-contrib/purescript-dom/blob/72ff57e6528cf896aa211f8e911fa55d97e79e05/src/DOM/Node/NodeType.purs#L21-L34 there are only 144 possible permutations but it bails out on this.

paf31:
/cc @garyb


garyb:
Awesome! :smile: :+1: 

Worthy of a 0.7.5.3 perhaps? (With the other error tweaks, etc.)

paf31:
Yep, we can do that. Let me see if there is anything else that can be squeezed into a patch.

paf31:
0.7.5.3 is tagged and on Hackage.

PR #1586, opened 2015-10-29
Fix #1465

### LINKED ISSUE #1465: Warning for shadowed type variables in a scoped type variables situation

paf31:
/cc @garyb

### DIFF
--- a/src/Language/PureScript/TypeChecker/Monad.hs
+++ b/src/Language/PureScript/TypeChecker/Monad.hs
@@ -24,6 +24,7 @@
 module Language.PureScript.TypeChecker.Monad where
 
 import Data.Maybe
+import Data.Foldable (for_)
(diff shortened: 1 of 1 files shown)


paf31:
We might want to consider doing all shadowing checks like this?

paf31:
Let's not put this into a patch. It also seems to suffer from the multiple warnings issue.

paf31:
Actually, the multiple warnings are correct. `Data.Map` has _a lot_ of shadowed scoped type variables.

garyb:
Yeah, let's not put this one in the patch - I think more than `Data.Map` will explode with warnings with this?

paf31:
Shall I merge this one now?

garyb:
Sure :+1:

PR #1587, opened 2015-10-29
Fix #1578

### LINKED ISSUE #1578: (..) import warning printed when reading externs
e.g. try building `psc-gui`.

paf31:
Maybe not the best fix, but it works.


garyb:
Yeah, that'll do :wink:

Let's keep the issue open though, I'll fix up the elaboration properly too.

Issue #1597, opened 2015-11-01
Improvements to unused import warnings


garyb:
- [x] Warn when importing data constructors for a type but only the type is used
- [x] Don't warn if the imported module is being re-exported (basically assume everything imported from that module is being used: it is!).
- [x] Warn for unqualified/non-selective imports (e.g. an unnecessary `import Prelude`)

/cc @nwolverson If I get some time I might look at these too, but if you fancy taking a further look that'd be great.

nwolverson:
The third case should be covered already, assuming you mean unqualified implicit import that isn't used as opposed to #1509. `Prelude` may be special (ignored) since I assumed `bind` may appear due to `do` desugaring, maybe there's a way to improve that?

garyb:
Ah okay, if it's just `Prelude` that that happens with I guess it makes sense, due to `bind` as you say.

nwolverson:
OK I was wrong, I was looking for `bind` in every module so nothing was warned. New PR fixes that, but will never warn on `Prelude` (due to `bind`) yet potentially warn on some other module if `bind` is being provided elsewhere.

PR #1600, opened 2015-11-01
Unused imports 2

nwolverson:
Improvements #1597. Fix warning for unqualified implicit imports, suppress for re-exported modules.

Doesn't yet work for modules re-exported by qualified import alias.

### DIFF
--- a/src/Language/PureScript/Linter/Imports.hs
+++ b/src/Language/PureScript/Linter/Imports.hs
@@ -39,7 +39,7 @@ findUnusedImports (Module _ _ _ mdecls _) env usedImps = do
   imps <- findImports mdecls
   forM_ (M.toAscList imps) $ \(mni, decls) -> unless (mni `elem` autoIncludes) $
     forM_ decls $ \(ss, declType, qualifierName) -> censor (onErrorMessages $ addModuleLocError ss) $
-      let usedNames = mapMaybe (matchName (typeForDCtor mni) qualifierName) $ sugarNames ++ M.findWithDefault [] mni usedImps in
+      let usedNames = mapMaybe (matchName (typeForDCtor mni) qualifierName) $ sugarNames mni ++ M.findWithDefault [] mni usedImps in
       case declType of
         Implicit -> when (null usedNames) $ tell $ errorMessage $ UnusedImport mni
         Explicit declrefs -> do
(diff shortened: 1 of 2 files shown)


garyb:
Good stuff :+1: 

I have one last suggestion: currently if we import multiple references from a module, none of which are used, we still print the "unused references" error and list everything - perhaps we could just go straight to the "redundant import" error instead?

garyb:
There seems to be a slight regression too: importing `Type(..)` doesn't warn if the data constructors are unused.

garyb:
Oops, sorry, that's not a regression, it was one of my suggestions again :smile:

nwolverson:
Can take a look at those remaining 2 suggestions, probably a couple of days until I get time.

garyb:
Cool, thanks!

src/Language/PureScript/Linter/Imports.hs
```diff
@@ -36,24 +36,34 @@ type UsedImports = M.Map ModuleName [Name]
...
+          Explicit declrefs -> do
+            let idents = mapMaybe runDeclRef declrefs
+            let diff = idents \\ usedNames
+            unless (null diff) $ tell $ errorMessage $ UnusedExplicitImport mni diff
+          _ -> return ()
   where
-  sugarNames :: [ Name ]
-  sugarNames = [ IdentName $ Qualified Nothing (Ident C.bind) ]
+  sugarNames :: ModuleName -> [ Name ]
+  sugarNames (ModuleName [ProperName n]) | n == C.prelude = [ IdentName $ Qualified Nothing (Ident C.bind) ]
```
paf31: Does `negate` need to be here?

garyb: I don't think it works like that yet, does it? Wasn't there an issue opened about making it do so recently?

paf31: Yes, but I wasn't fully sure what this function was for :smile:

nwolverson: That function identifies names that may be introduced at a later desugaring stage, in the case of `negate` it seems that happens earlier in the pipeline.

nwolverson:
Think those later two points are now covered, couldn't come to a great wording, bit messy maybe.

garyb:
Nice :)

How about these for the wording?

```
The import of type X includes data constructors but only the type is used
```

```
The import of type X includes the following unused data constructors:
```

I don't know :smile:

nwolverson:
That wording is at least better, thanks.

src/Language/PureScript/Linter/Imports.hs
garyb: I think `negate` works the same way as `bind` now: previously it didn't have that "whatever's in scope" magic, but it was updated to work that way recently.

garyb:
I haven't really given it much thought yet, but is there anything that prevents us from warning about unused class imports too? I guess we need to traverse type contexts, instances, and instance contexts for the references, but is there anything else that makes it difficult?

src/Language/PureScript/Linter/Imports.hs
nwolverson: Sure, the difference between `negate` and `bind` is that this analysis happens after `negate` is desugared but before `bind` is desugared, by this point `negate` is correctly resolved to the appropriate fully qualified identifier (warning seems to work both ways as I just tested).

To  push the analysis later and do that properly would need to keep around the mapping of used names/their resolved fully qualified & original versions.

garyb:
I'm happy for this to go in now. How about you @paf31?

paf31:
@garyb I'll defer to you on this, but everything looks sensible at first glance. :+1:

garyb:
Cool. I've been using this branch for a while too, it's nice :smile:

Issue #1429, opened 2015-08-26
Infinite type not being detected


zudov:
If I try to compile this:
```purescript
import Prelude
import Unsafe.Coerce
import Control.Monad.Eff
import DOM

type Storage eff = { clear :: Eff eff Unit }

getLocalStorage :: forall eff. Eff (dom :: DOM | eff) (Storage eff)
getLocalStorage = unsafeCoerce "undefined"

test = do
  storage <- getLocalStorage
  storage.clear
```
It would fail with following error message:
```purescript
Error in value declaration test:
Error checking that type
  WebStorage.Storage<_0>
subsumes type
  { clear :: Control.Monad.Eff.Eff (dom :: DOM.DOM | _0) _1 | _2 }
Error checking that type
  { clear :: Control.Monad.Eff.Eff _0 Prelude.Unit }
subsumes type
  { clear :: Control.Monad.Eff.Eff (dom :: DOM.DOM | _0) _1 | _2 }
Error checking that type
  Control.Monad.Eff.Eff _0 Prelude.Unit
subsumes type
  Control.Monad.Eff.Eff (dom :: DOM.DOM | _0) _1
Error at /home/zudov/prog/purescript/purescript-playground/raw-bindings/src/WebStorage.purs line 19, column 3 - line 19
, column 11:
  Infinite type detected:
    (dom :: DOM.DOM | _0)
```

Which can be fixed by changing the type to
```purescript
getLocalStorage :: forall eff eff2. Eff (dom :: DOM | eff) (Storage eff2)
```

However, if I change the first code sniplet to
```purescript
import Prelude
import Unsafe.Coerce
import Control.Monad.Eff
import DOM

foreign import data WEBSTORAGE :: !

type Storage eff = { clear :: Eff (webstorage :: WEBSTORAGE | eff) Unit }

getLocalStorage :: forall eff. Eff (dom :: DOM | eff) (Storage eff)
getLocalStorage = unsafeCoerce "undefined"

test = do
  storage <- getLocalStorage
  storage.clear
```

The compiler hangs and hogs all the memory.

And again changing the type declaration to
```purescript
getLocalStorage :: forall eff eff2. Eff (dom :: DOM | eff) (Storage eff2)
```
solves the problem.

PR #1601, opened 2015-11-01
New approach to unification

paf31:
This is step one in the move toward a new typechecker approach.

- Break up `Control.Monad.Unify`
- Use `mtl` style type signatures
- Walk the substitution _before_ unification instead of the weird old approach where substitution updates could trigger other unifications.


paf31:
I think this might fix #1429 but I need to test more.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -48,177 +46,213 @@ import Language.PureScript.Names
...
+  sub <- gets checkSubstitution
+  go (substituteKind sub k1) (substituteKind sub k2)
+  where
+  go (KUnknown u1) (KUnknown u2) | u1 == u2 = return ()
+  go (KUnknown u) k = solveKind u k
+  go k (KUnknown u) = solveKind u k
+  go Star Star = return ()
+  go Bang Bang = return ()
+  go (Row k1') (Row k2') = go k1' k2'
+  go (FunKind k1' k2') (FunKind k3 k4) = do
```
jdegoes: This is a perfect example of where shadowing is a good thing, and can prevent you from accidentally referencing stale parameters; i.e. you can bind `k1` and `k2` to make sure you cannot reference them from inside the definition of the function. Albeit in this case, I manually verified the code so it looks good to me. :+1:

paf31: Yeah this is less than ideal, I'll tidy it up, thanks!

jdegoes:
While reviewing the code, I also couldn't help notice the (usual) duplication between kind and type inference (`KUnknown Int` / `TUnknown Int`, `unifyKinds` / `unifyTypes`, etc.), and found [a proposal to unify kind and type inference in UHC](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=0CB0QFjAAahUKEwiv_7fklvLIAhULOj4KHeCNAag&url=http%3A%2F%2Fwww.cs.uu.nl%2Fdocs%2Fvakken%2Fts%2FPresentations%2FRemkoJeroen.ppt&usg=AFQjCNHsFL-ebz5b_prI_GWIUktbtcBjQQ&bvm=bv.106379543,d.cWw). Too big a change for this PR but I thought I'd at least link it for future reference.

paf31:
Yes, this PR actually is a move backwards in that regard, sort of. The previous code used to share a constraint solving module, but this pulls that apart. The reasons are that a) it makes the code much simpler, b) the old approach required orphan instances and c) I plan for the two implementations to differ more in the future, as I add new types of constraints and change the approach in the type checker portion to rely less on unification.

jdegoes:
Yeah, if they're going to be diverging, makes sense. :+1:

paf31:
This should build now.

paf31:
This fixes #1429 technically, in that it doesn't cause the compiler to spin any more, but it also allows an invalid program through (duplicate rows). I'm personally fine with that, since the duplicate rows check needs attention anyway, and might even be removed soon.

garyb:
I haven't reviewed this yet, but just thought I'd report I've had no problems with it either, after a day of heavy usage.

garyb:
:shipit:

paf31:
Thanks!

PR #1606, opened 2015-11-04
Add --require-path options to psc-bundle

### LINKED ISSUE #1605: psc-bundle doesn't work when code has been built with `--require-path`
I like to build with `--require-path=../` to avoid `node_modules` hackery, and `psc-bundle` [expects all require statements to be unprefixed](https://github.com/purescript/purescript/blob/dc54faf480c3e5dd64da69ef6b7e21d31bf17016/src/Language/PureScript/Bundle.hs#L145-L146). Should `psc-bundle` support `--require-path` as well?

natefaubion:
Fixes #1605

### DIFF
--- a/psc-bundle/Main.hs
+++ b/psc-bundle/Main.hs
@@ -48,6 +48,7 @@ data Options = Options
   , optionsEntryPoints :: [String]
   , optionsMainModule  :: Maybe String
   , optionsNamespace   :: String
+  , optionsRequirePath :: Maybe FilePath
   } deriving Show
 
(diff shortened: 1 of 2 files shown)


paf31:
:+1: Looks good. I can't test right now, but I can merge this evening if you think it works how one would expect.

natefaubion:
I've tested it on a small hello world app with and without the require-path, it works identically as before in the latter case, and using require-path yields identical results in the former case.

paf31:
Ack, sorry, could you please rebase this?

natefaubion:
Done.

paf31:
:+1: Thanks!

PR #1608, opened 2015-11-04
Add GHC 7.10.3 to build matrix

phadej:


paf31:
Thanks, but does it make sense for this to just replace 7.10.1 instead? Do we gain anything by building against multiple 7.10.* versions (given `base` shouldn't change and we probably want people to be using the latest patch in any release cycle)?

phadej:
We could drop 7.10.1 when 7.10.3. is officially released.

paf31:
Gotcha thanks.

Issue #1604, opened 2015-11-03
Use base-compat to reduce CPP boilerplate


phadej:
I could do a PR, if this is OK.

ping @paf31

paf31:
I'm not really familiar with `base-compat`. Sounds good to me, but are there any trade-offs?

phadej:
Additional dependency :)

paf31:
Ah that's fine.

PR #1607, opened 2015-11-04
Use base-compat in tests

phadej:
First step for https://github.com/purescript/purescript/issues/1604

This touches only tests, to act as an example. I could convert the rest either in single big PR or few smaller.

PR #1610, opened 2015-11-05
Base compat

phadej:
Follow-up to https://github.com/purescript/purescript/pull/1607, resolves https://github.com/purescript/purescript/issues/1604

Actually there are no CPP conditionals in the code base after this patch.


paf31:
Fantastic, thanks!

PR #1613, opened 2015-11-07
Fix #1175

### LINKED ISSUE #1175: Type annotations not checked inside where clause inside instance
It seems like both a where clause and an instance are required.

Simple example:

```purescript
module X where

class Foo where
  foo :: String

instance f :: Foo where
  foo = "a"
  where
    bar :: String
    bar = 1
```

paf31:
@garyb Could you please review this? Thanks.

### DIFF
--- a/examples/failing/1175.purs
+++ b/examples/failing/1175.purs
@@ -1,0 +1,11 @@
+-- @shouldFailWith TypesDoNotUnify
+module X where
+
+class Foo where
+  foo :: String
+
+instance f :: Foo where
+  foo = "a"
+    where
+    bar :: String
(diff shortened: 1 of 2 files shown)


garyb:
Looks good :+1: 

A side note - something I ran into recently is [it's _really_ hard to figure out what PRs are for when they're all titled "Fix #xxxx"](https://github.com/purescript/purescript/pulls?q=is%3Apr+is%3Aclosed), perhaps we should start including a hint as to what the referenced problem actually is too?

paf31:
Yes, you're right. It's just laziness on my part :smile:

PR #1614, opened 2015-11-07
Fix #1335, track scoped type variables when skolemizing

### LINKED ISSUE #1335: Scoped type variables issue with shadowed type variables
This declaration: https://github.com/slamdata/purescript-halogen/blob/f0022c672447c3501e8ef4746adbc617775c39a3/src/Halogen/Component.purs#L261-L262 does not typecheck when indented into the `where` block above it:

```
No instance found for Prelude.Functor (Data.Functor.Coproduct.Coproduct f524 (Data.Functor.Coproduct.Coproduct g522 h878))
```

Perhaps it's just related to instance resolution, but whatever it is there's some kind of scope-related problem.

paf31:

### DIFF
--- a/examples/passing/1335.purs
+++ b/examples/passing/1335.purs
@@ -1,0 +1,12 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+
+x :: forall a. a -> String
+x a = y "Done"
+  where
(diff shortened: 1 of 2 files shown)


garyb:
Ah, I see the problem now! I wasn't too sure how it was arising.

:+1:

PR #1615, opened 2015-11-07
Fix #1591, use the 'negate' in scope

### LINKED ISSUE #1591: Syntactic sugar for negative numbers requires Prelude
Most of the syntactic sugar uses whichever functions are in scope. That means I can use `do` notation with my own `bind` function, for instance. But the sugar for negative numbers like `-0` apparently requires the Prelude.

``` purescript
module Example where
x = -0
```

```
Error found:
in module Example
at /.../Example.purs line 2, column 5 - line 2, column 6

  Unknown module Prelude


See https://github.com/purescript/purescript/wiki/Error-Code-UnknownModule for more information,
```

It seems like it should use whichever `negate` function is in scope.

paf31:

### DIFF
--- a/src/Language/PureScript/Sugar/Operators.hs
+++ b/src/Language/PureScript/Sugar/Operators.hs
@@ -66,7 +66,7 @@ removeSignedLiterals (Module ss coms mn ds exts) = Module ss coms mn (map f' ds)
   where
   (f', _, _) = everywhereOnValues id go id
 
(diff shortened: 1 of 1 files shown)


garyb:
:+1: I assume we get an error about `negate` being undefined like with `bind` now, if there's no Prelude or anything?

paf31:
Yep that's right.

PR #1616, opened 2015-11-08
Fix #1590, limit depth of pretty-printed expressions

### LINKED ISSUE #1590: Possible to get absurdly long error message when types don't unify
This happened to me when my 127-line `main` function had an explicit type annotation with the wrong type: https://gist.github.com/tfausak/b57727bfbc4987f8e1bc

It took a while just to print out the message. It pretty-printed the entire desugared `main` function into the error message. 

Since that example is out of control, here's a simpler example the exhibits the same problem:

``` purescript
import Prelude (Unit(), bind, pure, unit)
import Control.Monad.Eff.Console (print)

main :: Eff () Unit
main = do
  x <- pure unit
  print x
```

```
Error found:
in module Test.Main
at /Users/tfausak/Documents/GitHub/purescript-ax/test/Main.purs line 8, column 8 - line 10, column 9

  Could not match expected type

    ( console :: CONSOLE
    | _0
    )

  with actual type

    ()


while trying to match type ( console :: CONSOLE
                           | _0
                           )
  with type ()
while checking that expression (bind (pure unit)) (\_0 ->
                                                     case _0 of
                                                       x -> print x
                                                  )
  has type Eff () Unit
in value declaration main

```

In general, I think that printing the expression is helpful. However it seems like there should be some limit to the size of the output.

paf31:
New error message for `examples/failing/Eff.purs`:

```text
Error found:
in module Main
at examples/failing/Eff.purs line 9, column 16 - line 13, column 23

  Could not match type

    ( console :: CONSOLE
    | _2
    )

  with type

    r0


while trying to match type ( console :: CONSOLE
                           , st :: ST _1
                           | _2
                           )
  with type ( st :: ST h0
            | r0
            )
while checking that expression (bind (newSTRef 0)) (\_0 ->
                                                      case _0 of
                                                        ref -> ...
                                                   )
  has type Eff ( st :: ST h0
               | r0
               )
           _0
in value declaration test
```

`--verbose-errors` prints full expressions, but by default it is limited to depth 3.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -513,7 +513,7 @@ prettyPrintSingleError full level e = prettyPrintErrorMessage <$> onTypesInError
     renderSimpleErrorMessage (EscapedSkolem binding) =
       paras $ [ line "A type variable has escaped its scope." ]
                      <> foldMap (\expr -> [ line "Relevant expression: "
-                                          , indent $ prettyPrintValue expr
+                                          , indent $ prettyPrintValue valueDepth expr
                                           ]) binding
     renderSimpleErrorMessage (TypesDoNotUnify t1 t2)
       = paras [ line "Could not match type"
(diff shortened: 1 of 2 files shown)


garyb:
Seems like a good idea to me.

paf31:
Thanks!

PR #1617, opened 2015-11-08
Missing space in an error message

zudov:
Before:

```
Overlapping names in function/binder in declarationsomeFunction
```

After:
```
Overlapping names in function/binder in declaration someFunction
```


paf31:
Thanks!

PR #1618, opened 2015-11-08
Record constructor field puns, fix #921

### LINKED ISSUE #921: Object property punning
When creating or matching objects, I'd really love to be able to write `{ x }` as shorthand for `{ x: x }`:

```purescript
data Person = Person { name :: String, age :: Number }

instance showPerson :: Show Person where
  show (Person { name, age }) = name <> ", who is " <> show age <> " years old"

mkPerson name age = Person { name, age }
```

balajirrao:


src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -334,9 +334,15 @@ parseObjectLiteral :: TokenParser Expr
 parseObjectLiteral = ObjectConstructor <$> braces (commaSep parseIdentifierAndValue)
 
 parseIdentifierAndValue :: TokenParser (String, Maybe Expr)
-parseIdentifierAndValue = (,) <$> (C.indented *> (lname <|> stringLiteral) <* C.indented <* colon)
-                              <*> (C.indented *> val)
+parseIdentifierAndValue =
+  do
+    name <- ident
+    b <- P.option (Just $ Var $ (Qualified Nothing) $ (Ident name)) rest
```
paf31: You have a couple of unnecessary `$`s here.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -334,9 +334,15 @@ parseObjectLiteral :: TokenParser Expr
...
+    return (name, b)
+  <|> (,) <$> ident <*> rest
```
paf31: One branch should be `stringLiteral` (with a required value) and the other should be an `lname` (without) since `Var` needs to take an `lname` not an arbitrary string.

paf31:
Looks good, but please see my comments. The issue here is that `{ "NotAnIdentifier" }` is a valid term and shouldn't be.

balajirrao:
Thanks for the comments! Updated the PR.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -334,9 +334,14 @@ parseObjectLiteral :: TokenParser Expr
 parseObjectLiteral = ObjectConstructor <$> braces (commaSep parseIdentifierAndValue)
 
 parseIdentifierAndValue :: TokenParser (String, Maybe Expr)
-parseIdentifierAndValue = (,) <$> (C.indented *> (lname <|> stringLiteral) <* C.indented <* colon)
-                              <*> (C.indented *> val)
+parseIdentifierAndValue =
+  do
+    name <- lname
+    b <- P.option (Just $ Var $ (Qualified Nothing) (Ident name)) rest
```
paf31: The parens around `Qualified Nothing` are unnecessary here.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -334,9 +334,14 @@ parseObjectLiteral :: TokenParser Expr
...
+    return (name, b)
+  <|> (,) <$> (C.indented *> (lname <|> stringLiteral)) <*> rest
```
paf31: No need for `lname` here. Do you see why?

balajirrao: Ah! The first branch handles the cases of `lname` with and without a value!

Updating the PR. Thanks!

paf31:
:+1: Great, thanks!

balajirrao:
Thanks!

PR #1622, opened 2015-11-11
support escapes for characters outside of BMP (code points > 0xFFFF)

michaelficarra:


paf31:
I'll take your word for it :smile: :+1:

PR #1624, opened 2015-11-12
Fix #1596, don't show type checker warnings in the event of an error

### LINKED ISSUE #1596: Wildcard warning with no position or useful information
Compiler version: `0.7.5.3` (62192bd6b5ecd2d7dd12512fbd0a2f6fbca0be68). No source file or position given, and the inferred type printed isn't useful either. My file had errors as well as this warning.

    Warning found:

        _4

    See https://github.com/purescript/purescript/wiki/Error-Code-WildcardInferredType for more information,

paf31:
I'm not particularly happy with this fix, but it's as good as it's going to get, without making our own `mtl`-like classes. The reason for the lack of position information is related to the interaction between `MonadWriter`, `censor` and `ExceptT`.

This change just omits warnings from the typechecker whenever an error occurs. Those warnings are never useful anyway, since they always contain unknown types due to the lack of a substitution.

/cc @garyb

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -944,7 +944,7 @@ renderBox = unlines . map trimEnd . lines . Box.render
 rethrow :: (MonadError e m) => (e -> e) -> m a -> m a
 rethrow f = flip catchError $ \e -> throwError (f e)
 
-reifyErrors :: (MonadError e m) => m a -> m (Either e a)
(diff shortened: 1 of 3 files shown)


paf31:
@garyb Any thoughts on this?

garyb:
I think it's a great idea. Wading through warnings when there's a fatal error currently is something I hadn't realised was irritating me until I noticed this solution (but it was) :smile:. :+1: for sure.

paf31:
Thanks!

PR #1626, opened 2015-11-13
Fix missing spaces in psc-publish error message

hdgarrood:
This is what the `psc-publish` error message in question looks like in 0.7.5.3:

```
  If the version you are publishing is not yet tagged, you might want to usethe
  --dry-run flag instead, which removes this requirement. Runpsc-publish --help
  for more details.
```

(Note missing spaces between "use" and "the", and "Run" and "psc-publish"). This commit adds these spaces in.


paf31:
Thanks!

Issue #1560, opened 2015-10-25
Minor problems with psc-publish error reporting


hdgarrood:
A couple of minor things I noticed from https://github.com/hdgarrood/bower-json/issues/1:

* The `BowerJSONNameMissing` constructor should not exist, that kind of validation is all already done within `bower-json`.
* The error message for the `CouldntParseBowerJSON` is a bit misleading, as it implies the problem is a JSON syntax error. However, take the case where your `bower.json` is valid JSON but not a valid bower package manifest; say, because it doesn't have a "name" key. In this case the error message is not appropriate.
* ~~Somehow~~ we are using the `Show` instance for a `ParseError BowerError` in the error message. We should use the existing functions inside `aeson-better-errors` and `bower-json` which can display the same information in a human-readable way. The culprit is https://github.com/purescript/purescript/blob/0e6bfaed6fc05dbd82834fa9aedf153175a0511f/src/Language/PureScript/Publish/ErrorsWarnings.hs#L141

PR #1627, opened 2015-11-14
Psc publish error reporting

hdgarrood:
Fix all of the problems noted in #1560.

Example error message now:

```
published.
Details:
  There was a problem with your bower.json file:
    At the path: ["name"]
    Invalid package name: The following characters are not permitted in package
    names: $

  Please ensure that your bower.json file is valid.
```

as opposed to:

```
published.                                                                       
Details:                                                                         
  The bower.json file could not be parsed as JSON:                               
    aeson reported: BadSchema [ObjectKey "name"] (CustomError (InvalidPackageName
    (InvalidChars "$")))                                                         
                                                                                 
  Please ensure that your bower.json file is valid JSON. 
```

### DIFF
--- a/src/Language/PureScript/Publish.hs
+++ b/src/Language/PureScript/Publish.hs
@@ -122,7 +122,7 @@ preparePackage' opts = do
   requireCleanWorkingTree
 
   pkgMeta <- liftIO (Bower.decodeFile "bower.json")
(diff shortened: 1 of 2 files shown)


paf31:
:+1: Thanks!

PR #1628, opened 2015-11-14
Reduce noise in instance declarations

hdgarrood:
`instance showFoo :: Show Foo` is now rendered as `Show Foo`, since the
extra information is mostly just noise.

I did this mainly with pursuit in mind (to fix purescript/pursuit#124), although it has the same effect on `psc-docs`.

The only situation I can think of where the instance name would be important is if you were calling into PureScript code from JS, but if you were doing that, it's surely easier to write a little PS wrapper exporting monomorphic versions of the functions you want, in most cases at least. In any case I think instance names are not important enough for 99% of use cases to be featured in the docs like this.


garyb:
I'm fine with this :+1:

paf31:
:+1: Thanks!

PR #1633, opened 2015-11-15
Remove unnecessary findQualModules from CoreFn

### LINKED ISSUE #1408: Check whether we really do need findQualModules in CoreFn desguaring
It shouldn't be necessary due to https://github.com/purescript/purescript/blob/7e8e8ca83bbc36dcdbf35c2be1eb25eba7b37bc8/src/Language/PureScript/Sugar/Names.hs#L251-L266

garyb:
This already covered by `elaborateImports` in name desugaring.

Resolves #1408


paf31:
:+1: Sounds good to me.

garyb:
Obviously there's something not quite right here, I made a few test cases that I thought would break with this change that did not, so I'll dig into the failing test and figure out what's happening.

garyb:
Verdict: It's not unnecessary :smile:

PR #1634, opened 2015-11-15
Fix #1632, instantiate type variables in anyProxy calls in generic instances

### LINKED ISSUE #1632: Issue with type arguments in derived generics
```purescript
data D a = D a

derive instance genericD :: (Generic b) => Generic (D b)
```

gives

```text
Type variable a is undefined.

while checking the kind of Proxy a
while checking the kind of Proxy a
while checking that expression anyProxy
  has type Proxy _0
while applying a function toSignature
  of type (Generic _0) => Proxy _0 -> GenericSignature
  to argument anyProxy
in value declaration genericD
```

paf31:
@garyb @sclv Could you please review this? This also allows generic instances of the form

```purescript
deriving instance genericX :: Generic (X String)
```

i.e. flexible instances.

### DIFF
--- a/core-tests/tests/generic-deriving/Main.purs
+++ b/core-tests/tests/generic-deriving/Main.purs
@@ -17,7 +17,11 @@ data A a
   | D { a :: a }
   | E Void
 
-derive instance genericA :: (Generic a) => Generic (A a)
+derive instance genericA :: (Generic b) => Generic (A b)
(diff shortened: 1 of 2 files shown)


garyb:
Seems reasonable :+1:

paf31:
Thanks!

Issue #1306, opened 2015-07-25
Should psc-publish refuse to run if the working tree is dirty?


hdgarrood:
that is, if any files have been added, deleted, or modified since the last commit. I think it should at least warn (currently it doesn't do anything).

paf31:
:+1: Definitely.

michaelficarra:
Let's warn and/or ask for confirmation. Not everyone uses git.

hdgarrood:
Having thought about this a bit more, my vote is to error if we are inside a git repository.

If we want to accommodate for people who don't use git, we need to come up with an alternative solution for getting the version. Currently, we do it by shelling out to git, to get a tag pointing at the currently checked out commit, and then we attempt to parse that.

hdgarrood:
Also, we'd need to come up with an alternative for all the GitHub things Pursuit relies on: authentication, source links, readmes, and links to the project (which currently always point to its github page).

I don't mean to argue that we shouldn't do this, but it is a fairly big undertaking, so I'd like to wait until there's at least some demand for it.

tfausak:
Sorry to revive an old thread, but this behavior is driving me nuts. I run `psc-publish --dry-run` as part of my tests, and the fact that it *fails* (instead of warning) if I have a dirty working tree is extremely frustrating.

hdgarrood:
yeah, good call, :+1: for making psc-publish warn instead of failing when the working tree is dirty.

paf31:
:+1:

tfausak:
In fact, I would prefer it to always be a warning. I ran into this issue again while trying to automatically upload documentation to Pursuit from Travis CI. Since Travis checks out a branch, it's in a detached head state. psc-publish treats that as dirty, even if it's not. See the build for details: https://travis-ci.org/tfausak/purescript-neon/builds/89132864#L1345-L1353

```
$ git status
# HEAD detached at 63df46f
nothing to commit, working directory clean
$ npm run psc-publish | tail -n 1 > purescript-neon-${TRAVIS_TAG}.json
There is a problem with your package, which meant that it could not be         
published.                                                                     
Details:                                                                       
  Your git working tree is dirty. Please commit, discard, or stash your changes
  first.                                                                       
```

michaelficarra:
Well then the method psc-publish uses for a dirty check is wrong. The messages from git and psc-publish should not conflict.

tfausak:
[...]

PR #1641, opened 2015-11-17
psc-publish: only warn on dirty working tree on dry runs

hdgarrood:
As opposed to erroring. See discussion on #1306


paf31:
:+1: Thanks, I've wanted this for a while.

Issue #1313, opened 2015-07-28
Could psc-bundle check for missing FFI functions?


hdgarrood:
If I write `foreign import spawn :: [...]` inside `Module.purs`, and then I forget to write `exports.spawn = [...]` inside `Module.js`, it would be nice for one out of `psc` or `psc-bundle` to warn me about this.

hdgarrood:
Thinking about this a little more. I might have a go at this.

I feel like this should be done by `psc`, actually; one of the selling points of PureScript ought to be the fact that errors are caught at compile time rather than run time, right? In particular, if we don't check this inside `psc`, it may well end up as a runtime error. People won't use `psc-bundle` in every case.

zudov:
I just thought about how would it work, and thought that it would be much easier to implement if we declared the FFI exports using [`export` statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import).

hdgarrood:
Yeah, I think this is one reason ES6 modules are on the roadmap. This has already been implemented in `psc-bundle` to some extent, though &mdash; using code that's already in `psc-bundle`, it's pretty easy to take an FFI module and get a list of identifiers it exports.

zudov:
> Using code that's already in psc-bundle, it's pretty easy to take an FFI module and get a list of identifiers it exports.

Then we should do it, and probably provide some pragma which would make the psc shut up (you would use it if for some reason you want to put something into `exports` in a non-straightforward way)

zudov:
On a second thought `exports.foo = null` would work equally well to disable "Undefined FFI export" for warning for identifier `foo`.

hdgarrood:
Yes, although there might be other cases where an escape hatch might be useful. For example, suppose you wanted to write something like this:

```javascript
for (let name of something) {
  exports[name] = somethingElse(name);
}
```

zudov:
I see, putting `exports.foo = null` in FFI whenever you want to use some of those names isn't fun.

zudov:
On the other hand you probably should instead do

```javascript
exports.things = {};
for (let name of something) {
  exports.things[name] = somethingElse(name);
}
```

```purescript
foreign import things :: { fooThingy :: Foo, barThingy :: Bar }
```

kRITZCREEK:
ES6 modules lose the ability to conditionally import modules which gives you sometimes needed flexibility.

[...]

PR #1642, opened 2015-11-17
Check that FFI imports match with implementations

hdgarrood:
A first pass at #1313. Not ready for merging just yet.

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -229,53 +229,21 @@ toModule requirePath mids mid top
   err = throwError . ErrorInModule mid
 
   toModuleElement :: JSStatement -> m ModuleElement
-  -- var ModuleName = require("file");
   toModuleElement stmt
-    | JSVariable _ jsInit _ <- stmt
-    , [JSVarInitExpression var varInit] <- commaList jsInit
-    , JSIdentifier _ importName <- var
-    , JSVarInit _ jsInitEx <- varInit
-    , JSMemberExpression req _ argsE _ <- jsInitEx
-    , JSIdentifier _ "require" <- req
-    , [ Just importPath ] <- map fromStringLiteral (commaList argsE)
-    , importPath' <- checkImportPath requirePath importPath mid mids
-    = pure (Require stmt importName importPath')
-  -- var foo = expr;
+    | Just (importName, importPath) <- matchRequire requirePath mids mid stmt
+    = pure (Require stmt importName importPath)
   toModuleElement stmt
-    | JSVariable _ jsInit _ <- stmt
-    , [JSVarInitExpression var varInit] <- commaList jsInit
-    , JSIdentifier _ name <- var
-    , JSVarInit _ decl <- varInit
-    = pure (Member stmt False name decl [])
-  -- exports.foo = expr; exports["foo"] = expr;
-  toModuleElement stmt
-    | JSAssignStatement e (JSAssign _) decl _ <- stmt
-    , Just name <- accessor e
-    = pure (Member stmt True name decl [])
-    where
-    accessor :: JSExpression -> Maybe String
-    accessor (JSMemberDot exports _ nm)
-      | JSIdentifier _ "exports" <- exports
-      , JSIdentifier _ name <- nm
-      = Just name
-    accessor (JSMemberSquare exports _ nm _)
-      | JSIdentifier _ "exports" <- exports
-      , Just name <- fromStringLiteral nm
-      = Just name
-    accessor _ = Nothing
+    | Just (exported, name, decl) <- matchMember stmt
+    = pure (Member stmt exported name decl [])
   -- module.exports = { ... }
   toModuleElement stmt
-    | JSAssignStatement e (JSAssign _) decl _ <- stmt
-    , JSMemberDot module' _ exports <- e
-    , JSIdentifier _ "module" <- module'
-    , JSIdentifier _ "exports" <- exports
-    , JSObjectLiteral _ props _ <- decl
+    | Just props <- matchExportsAssignment stmt
     = (ExportsList <$> traverse toExport (trailingCommaList props))
     where
       toExport :: JSObjectProperty -> m (ExportType, String, JSExpression, [Key])
       toExport (JSPropertyNameandValue name _ [val]) =
         (,,val,[]) <$> exportType val
-                   <*> extractLabel name
+                   <*> extractLabel' name
       toExport _ = err UnsupportedExport
 
       exportType :: JSExpression -> m ExportType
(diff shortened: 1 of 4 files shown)


src/Language/PureScript/Bundle.hs
```diff
@@ -229,46 +230,20 @@ toModule requirePath mids mid top
...
-    accessor (JSMemberDot [ exports ] _ nm)
-      | JSIdentifier "exports" <- node exports
-      , JSIdentifier name <- node nm
-      = Just name
-    accessor (JSMemberSquare [ exports ] _ nm _)
-      | JSIdentifier "exports" <- node exports
-      , JSExpression [e] <- node nm
-      , JSStringLiteral _ name <- node e
-      = Just name
-    accessor _ = Nothing
```
hdgarrood: This is me extracting code out of `toModule` and into `matchMember`, `matchExport`, etc. so that it could be used in both `toModule` and `getExportedIdentifiers`.

src/Language/PureScript/Make.hs
```diff
@@ -384,3 +389,34 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
 
   progress :: ProgressMessage -> Make ()
   progress = liftIO . putStrLn . renderProgressMessage
+
+-- | For a given CoreFn module and corresponding FFI module file path,
+-- check that there is exactly one FFI declaration for each
+checkForeignDecls :: CF.Module CF.Ann -> FilePath -> SupplyT Make ()
+checkForeignDecls m path = do
```
hdgarrood: I know this is suuuper gross with `internalError` all over the place. I just wanted to see if this approach made sense first, before worrying about the errors.

src/Language/PureScript/Bundle.hs
```diff
@@ -279,16 +254,89 @@ toModule requirePath mids mid top
...
+  | JSSourceElementsTop ns <- node top = concat <$> traverse go ns
+  | otherwise = err InvalidTopLevel
+  where
+  err = throwError . ErrorInModule (ModuleIdentifier mname Foreign)
+
+  go :: JSNode -> m [String]
+  go n
+    | Just props <- matchExport n
+    = traverse toIdent props
+    | Just (True, name, _) <- matchMember n
```
hdgarrood: Was I right in thinking that the `Bool` argument to `Member` indicates whether the member is exported? What even is a `Member`? (Would be nice to add that to the comments in this module, perhaps).

src/Language/PureScript/Make.hs
```diff
@@ -384,3 +389,34 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
...
+  jsStr <- lift $ makeIO (internalError . show) (readFile path)
+  js <- either throwUnableToParse pure $ JS.parse jsStr path
+
+  foreignIdents <- fmap S.fromList . lift . Make . lift $
+    catchE (getExportedIdentifiers (runModuleName mname) js)
+           (internalError . show)
+  let importedIdents = S.fromList $ map (runIdent . fst) (CF.moduleForeign m)
+
+  let extraneousFFI = foreignIdents  S.\\ importedIdents
+  let missingFFI    = importedIdents S.\\ foreignIdents
```
hdgarrood: It may well be possible to do this more neatly, without traversing both sets twice. I'm not aware of any function to achieve this, though.

src/Language/PureScript/Bundle.hs
hdgarrood: Ok, looking at this again, I think I've worked out the answers: a Member is just a declaration inside an FFI module: either `exports.foo = bar` or `var foo = bar` at the top level, and the `Bool` argument does indeed seem to indicate whether or not it is exported.

hdgarrood:
For the sake of clarify: I would be glad to take this further, but before I do, I would like to get confirmation that the approach I've taken here looks vaguely sensible.

src/Language/PureScript/Make.hs
```diff
@@ -397,3 +402,34 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
 
   progress :: ProgressMessage -> Make ()
   progress = liftIO . putStrLn . renderProgressMessage
+
+-- | For a given CoreFn module and corresponding FFI module file path,
+-- check that there is exactly one FFI declaration for each
+checkForeignDecls :: CF.Module CF.Ann -> FilePath -> SupplyT Make ()
```
garyb: Could perhaps leave `Module` polymorphic here, I don't think knowing about the annotations is necessary in here, is it?

garyb:
Seems like a sensible approach to me, and I'd love to have this stuff checked. I'm not all that familiar with the bundle process so probably still worth waiting for @paf31 to chime in though :)

src/Language/PureScript/Make.hs
kRITZCREEK: I know that it's called the [Symmetric difference](https://en.wikipedia.org/wiki/Symmetric_difference), but I don't think you get any faster than this without losing the ability to differentiate between `extraneousFFI ` and `missingFFI`.

hdgarrood: Ah, no, I mean I was hoping that a function exists `Set a -> Set a -> (Set a, Set a)` such that I could do `let (extraneousFFI, missingFFI) = foreignIdents ??? importedIdents` or something.

nwolverson:
Oh didn't realise this was active again. Now going to conflict horribly on the bundle side with #1989

hdgarrood:
Yeah that occurred to me too. No worries, I'll update it when I get around to it.

hdgarrood:
By which I mean - please don't worry about conflicts this PR, I think #1989 ought to take precedence and get merged first.

paf31:
:+1: The general approach looks good to me. Maybe we should create an issue to track this.

hdgarrood:
Up to you, as long as it makes it into the release notes :) I'll create an issue if you want.

I pushed some more commits fixing the conflicts from the language-javascript upgrade, and removing all the calls to `internalError`, so this is done and ready for review from my point of view now.

hdgarrood:
Wait, we do have an issue for this - #1313

hdgarrood:
This is turning out to be useful already! https://github.com/purescript-node/purescript-node-buffer/commit/3e1f0ab719845969ae82f73713374cd15342d51d :)

src/Language/PureScript/Bundle.hs
```diff
@@ -287,14 +255,105 @@ toModule requirePath mids mid top
...
+matchMember stmt
+  -- var foo = expr;
+  | JSVariable _ jsInit _ <- stmt
+  , [JSVarInitExpression var varInit] <- commaList jsInit
+  , JSIdentifier _ name <- var
+  , JSVarInit _ decl <- varInit
+  = Just (False, name, decl)
+  | JSAssignStatement e (JSAssign _) decl _ <- stmt
+  , Just name <- accessor e
+  -- exports.foo = expr; exports["foo"] = expr;
```
nwolverson: comment in a slightly confusing place, makes more sense a couple lines up

nwolverson:
Read through this and didn't try it out, but I like the look of it.

src/Language/PureScript/Bundle.hs
hdgarrood: :+1: thanks, fixed

paf31:
Thanks!

PR #1644, opened 2015-11-18
Make `let` optional in psci

hdgarrood:
Only tested by hand. I tried to get some sort of thing wired up so that we could give a list of commands and an expected output but it turns out that's actually quite difficult. I am beginning to wonder whether the best approach is to actually run psci in a separate process, and to use something like https://github.com/stroan/haskell-libexpect/.

### DIFF
--- a/psci/Parser.hs
+++ b/psci/Parser.hs
@@ -98,10 +98,12 @@ psciExpression = Expression <$> P.parseValue
 -- we actually want the normal @let@.
 --
 psciLet :: P.TokenParser Command
-psciLet = Decls <$> (P.reserved "let" *> P.indented *> manyDecls)
+psciLet = Decls <$> (optionalLet *> manyDecls)
   where
   manyDecls :: P.TokenParser [P.Declaration]
   manyDecls = mark (many1 (same *> P.parseLocalDeclaration))
+
+  optionalLet = optional (P.reserved "let" *> P.indented)
 
 -- | Imports must be handled separately from other declarations, so that
 -- :show import works, for example.


hdgarrood:
Example:
```
:? shows help
> import Prelude
> x = 6
> x
6

> y = x + 7
> :t y
Int

> y
13
```

paf31:
My concern is that this will make parser error messages much worse, since there is no initial parse to identify a `let` bindings. Is this intended for editor integration?

hdgarrood:
Not really; I was reminded of this by someone in IRC because they wanted to use it for editor integration, but I personally think this is a good idea because it confuses Haskell beginners all the time and I wasn't aware of any technical reason not to. Can you give an example of what kind of input would result in a worse parser error message?

AppShipIt:
I agree it's the way I'd expect it to work intuitively. Did GHCi just implement it ([#7253](https://ghc.haskell.org/trac/ghc/ticket/7253))?

AppShipIt:
@hdgarrood @paf31 Dropped? GHCi just followed suit with iHaskell.

paf31:
Not dropped, just on hold, I suppose. I'd like to know that this won't affect parse errors too much, that's my main concern here.

rvion:
in case it's not dead, +1
I think "let" is not usefull, and only adding some complexity to the repl.

🍭  Also, not having to write `let` would increase ide integrations who provide a
"Send Line to PSCi buffer" or  "Send Selection to PSCi buffer" command, like https://github.com/nwolverson/atom-ide-purescript

Wouldn't this PR be a good case to write a few PSCIi integration tests ?

paf31:
I'd like to come to some decision on this one. I'm not completely against it, but I'd like to investigate the impact this will have on error messages. I think a possible alternative is to support a "declaration mode" which works like `:paste`.

Since this is in a conflicting state, and I'm tidying up old PRs, I'm going to close this for now, but we can continue to discuss it here.

hdgarrood:
Fair enough - if I ever get around to looking at this again I will investigate the impact on error messages. Are you thinking of parse errors specifically? I opened this originally not for the reason @rvion mentioned, but because I think the current way it works can be a little annoying and confusing (back when I used to spend time on /r/haskell, people who had been confused by it seemed to appear fairly often).

paf31:
I agree it can be irritating. The place I run into it is when pasting examples from the book. But I find bad parse errors even more irritating, and we've done quite a bit to reduce backtracking and improve them lately, so it'd be a shame to have worse errors only in PSCi.

But it might turn out it's not a big deal in practice, so I'd just like to do some testing to find out before we merge this.

PR #1648, opened 2015-11-21
Revise reserved & built in names list

garyb:
@michaelficarra could you take a look at this please?

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/Common.hs
+++ b/src/Language/PureScript/CodeGen/JS/Common.hs
@@ -9,6 +9,11 @@ import Data.List (intercalate)
 import Language.PureScript.Crash
 import Language.PureScript.Names
 
+moduleNameToJs :: ModuleName -> String
+moduleNameToJs (ModuleName pns) =
+  let name = intercalate "_" (runProperName `map` pns)
+  in if nameIsJsBuiltIn name then "$$" ++ name else name
+
 -- |
 -- Convert an Ident into a valid Javascript identifier:
 --
@@ -19,8 +24,9 @@ import Language.PureScript.Names
 --  * Symbols are prefixed with '$' followed by a symbol name or their ordinal value.
 --
 identToJs :: Ident -> String
-identToJs (Ident name) | nameIsJsReserved name = "$$" ++ name
-identToJs (Ident name) = concatMap identCharToString name
+identToJs (Ident name)
+  | nameIsJsReserved name || nameIsJsBuiltIn name = "$$" ++ name
+  | otherwise = concatMap identCharToString name
 identToJs (Op op) = concatMap identCharToString op
 identToJs (GenIdent _ _) = internalError "GenIdent in identToJs"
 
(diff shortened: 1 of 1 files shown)


garyb:
Bump! /cc @michaelficarra

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,160 @@ identCharToString c = '$' : show (ord c)
...
+  concat
+    [ jsReserved
+    , jsFutureReserved
+    , jsFutureReservedStrict
+    , jsOldReserved
+    , jsLiterals
+    ]
+
+jsReserved :: [String]
+jsReserved =
```
michaelficarra: Can you rename this list to `jsKeywords`?

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,160 @@ identCharToString c = '$' : show (ord c)
...
+  , "export"
+  , "extends"
+  , "finally"
+  , "for"
+  , "function"
+  , "if"
+  , "import"
+  , "in"
+  , "instanceof"
+  , "let"
```
michaelficarra: Including `let` and `yield` in here is a bit inappropriate. I'd prefer to see `let`, `yield`, and `static` in their own list of `jsSometimesReserved`.

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,160 @@ identCharToString c = '$' : show (ord c)
...
+  ]
+
+jsFutureReserved :: [String]
+jsFutureReserved =
+  [ "enum"
+  , "await"
+  ]
+
+jsFutureReservedStrict :: [String]
+jsFutureReservedStrict =
```
michaelficarra: Can you sort this list and remove `static`?

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,160 @@ identCharToString c = '$' : show (ord c)
...
+  , "package"
+  , "protected"
+  , "static"
+  , "interface"
+  , "private"
+  , "public"
+  ]
+
+jsOldReserved :: [String]
+jsOldReserved =
```
michaelficarra: This list is missing `throws`.

src/Language/PureScript/CodeGen/JS/Common.hs
garyb: I found `static` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Future_reserved_keywords), is that incorrect then?

michaelficarra: See the NOTE in [11.6.2.1](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-keywords).

garyb: Oh sorry, I see now that you wanted to move it to `jsSometimesReserved`, I thought you were suggesting removing it entirely.

garyb:
@michaelficarra thanks for the feedback! I've updated the PR with your suggestions.

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,166 @@ identCharToString c = '$' : show (ord c)
...
+    , "Symbol"
+    , "SyntaxError"
+    , "TypeError"
+    , "Uint16Array"
+    , "Uint32Array"
+    , "Uint8Array"
+    , "Uint8ClampedArray"
+    , "undefined"
+    , "unescape"
+    , "uneval"
```
michaelficarra: Where is this one coming from?

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,166 @@ identCharToString c = '$' : show (ord c)
...
+    , "parseFloat"
+    , "parseInt"
+    , "Promise"
+    , "Proxy"
+    , "RangeError"
+    , "ReferenceError"
+    , "Reflect"
+    , "RegExp"
+    , "Set"
+    , "SIMD"
```
michaelficarra: Not official yet, but I'm okay with it.

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,166 @@ identCharToString c = '$' : show (ord c)
...
+jsSometimesReserved :: [String]
+jsSometimesReserved =
+  [ "let"
+  , "static"
+  , "yield"
+  ]
+
+jsFutureReserved :: [String]
+jsFutureReserved =
+  [ "await"
```
michaelficarra: Ugh, this is my fault for not noticing earlier. I would also move `await` to `jsSometimesReserved`, as it's only reserved in module code.

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,166 @@ identCharToString c = '$' : show (ord c)
...
+    , "encodeURI"
+    , "encodeURIComponent"
+    , "Error"
+    , "escape"
+    , "eval"
+    , "EvalError"
+    , "Float32Array"
+    , "Float64Array"
+    , "Function"
+    , "Generator"
```
michaelficarra: `Generator` and `GeneratorFunction` definitely shouldn't be in this list.

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -75,107 +81,166 @@ identCharToString c = '$' : show (ord c)
...
+    , "GeneratorFunction"
+    , "Infinity"
+    , "Int16Array"
+    , "Int32Array"
+    , "Int8Array"
+    , "InternalError"
```
michaelficarra: This doesn't belong here either.

michaelficarra:
LGTM otherwise.

src/Language/PureScript/CodeGen/JS/Common.hs
garyb: It's non-standard but always present in Firefox.

src/Language/PureScript/CodeGen/JS/Common.hs
garyb: Another FF global value...

src/Language/PureScript/CodeGen/JS/Common.hs
michaelficarra: Do we care about those? There's also plenty of non-standard globals in other browsers, such as `chrome` in Chrome and `opera` in Opera.

garyb: Yeah, probably not; we'll never cover every possible non-standard value in every browser. I'll take them out.

garyb:
Updated again. This should be good to merge now @paf31.

michaelficarra:
:+1:

paf31:
:+1: Sounds good.

PR #1649, opened 2015-11-21
Fix codegen error with instance for re-exported class

### LINKED ISSUE #1595: Class not defined when defining instance
If you import a class from a module that re-exports that class, you cannot define instances of it. 

``` purescript
module A (module A.B) where
import A.B

module A.B where
class X a

module Z where
import A
data Y
instance z :: X Y
```

``` javascript
// Generated by psc version 0.7.5.3
"use strict";
var A = require("A");
var z = A_B.X;
module.exports = {
    z: z
};
```

On the line where `z` is defined, `A_B` is not in scope. I end up getting a reference error when I run it:

```
ReferenceError: A_B is not defined
```

garyb:
Resolves #1595


paf31:
:+1: Looks good to me.

PR #1650, opened 2015-11-21
Fix missing data constructors in re-exports

### LINKED ISSUE #1594: Data constructors get lost when re-exported
It looks like data constructors do not get re-exported if there is more than one level of indirection.

``` purescript
module A (module B) where
import B

module B (module C) where
import C

module C where
data X = Y

module Z where
import A (X(Y))
```

```
Error found:
at /.../Z.purs line 2, column 11 - line 2, column 15

  Module A does not export data constructor Y for type X


See https://github.com/purescript/purescript/wiki/Error-Code-UnknownImportDataConstructor for more information,
```

I can get that to compile by importing `X(Y)` from `B` (or `C`) instead of `A`.

garyb:
Resolves #1594 

I'm not 100% sure on this one, I want to run it against Halogen and SlamData and the like first, but on the train currently so can't right now :)


garyb:
Ok, it works. I realise this is almost impossible to review... but basically the problem is the filtering of data constructors didn't take into account re-exporting deeper than one step removed.

Switching the filter here means instead of checking the data constructor qualification against the module they originally reside in, it instead only checks one step back in the chain.

paf31:
I'll take your word for it :smile:

PR #1651, opened 2015-11-21
Allow import hiding with qualified imports

### LINKED ISSUE #1507: Allow qualified import with hiding
``` purescript
import Foo hiding (baz) as X
```

I think putting the `hiding` before the `as` is clearer than having `hiding` follow it, considering the effect this should have. Forming it this way also makes it clear why you can't combine selective imports with `hiding`, as it makes `hiding` a modifer of the selection.

garyb:
Resolves #1507

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -242,7 +242,7 @@ createTemporaryModuleForImports PSCiState{psciImportedModules = imports} =
     P.Module (P.internalModuleSourceSpan "<internal>") [] moduleName (importDecl `map` imports) Nothing
 
 importDecl :: ImportedModule -> P.Declaration
-importDecl (mn, declType, asQ) = P.ImportDeclaration mn declType asQ
+importDecl (mn, declType, asQ) = P.ImportDeclaration mn declType asQ False
 
 indexFile :: FilePath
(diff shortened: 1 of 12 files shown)


paf31:
Looks good but don't we need to handle `hiding` in the `qualImport` branch too? Or are we just going to deprecate the `qualified` keyword?

garyb:
Yeah I thought we were going to do that. I was trying to think of a good way to raise the warning - tagging the `ImportDefinition` with some flag and then warning during desugaring was the first thing that sprang to mind, but is there some better way we can capture warnings in the parser without resorting to that or switching parsing to use a transformer stack?

paf31:
I'm not above using `unsafePerformIO` if we're going to remove it in 0.8 anyway :smile:

garyb:
:scream: :laughing: 

Ok, I'll append that change to this!

paf31:
Although having the parser run in `IO` probably isn't too difficult either. I don't mind either way.

garyb:
Actually, I'll probably do the warn-in-desugar thing, as at least then the warnings will appear consistently with the rest.

garyb:
Updated with the warning now, how's this?

paf31:
:+1:

PR #1655, opened 2015-11-22
Warn about unused class imports

### LINKED ISSUE #1392: Warn about unused imports

### LINKED ISSUE #1597: Improvements to unused import warnings

garyb:
Resolves #1392, resolves #1597 (@nwolverson deserves the credit for these, this is just the last bit I think)

Also fixes a bug in the current warnings: when `import X` only imports classes from `X`, it thinks `X` is a redundant import.

### DIFF
--- a/src/Language/PureScript/Linter/Imports.hs
+++ b/src/Language/PureScript/Linter/Imports.hs
@@ -25,7 +25,11 @@ import Language.PureScript.Sugar.Names.Imports
 import qualified Language.PureScript.Constants as C
 
 -- | Imported name used in some type or expression.
-data Name = IdentName (Qualified Ident) | IsProperName (Qualified ProperName) | DctorName (Qualified ProperName)
+data Name
+  = IdentName (Qualified Ident)
+  | TypeName (Qualified ProperName)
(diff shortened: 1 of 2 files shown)


nwolverson:
I think `matchClass` could just be another case in `matchName` and don't need the `++` ? 
If I read this right there's really 1 key line that you've added in `runDeclRef` and the rest is cleanup (removing that horrible data constructor name!)

garyb:
Yep, you're absolutely right. In fact, all these changes could be undone aside from the tweak to `runDeclRef` and it would fix the problem. I guess the differentiation for `IsProperName` perhaps makes things slightly easier to understand though.

nwolverson:
Completely agree that it's clearer. Also seems to work for me.

paf31:
:+1:

PR #1656, opened 2015-11-22
Add information about skolems to type errors

### LINKED ISSUE #1111: Explain the origin of type variables in type errors, where possible
For example, ghci with `-XScopedTypeVariables` says:

```
λ: let x :: forall i. i -> String; x s = s

<interactive>:3:39:
    Couldn't match expected type ‘String’ with actual type ‘i’
      ‘i’ is a rigid type variable bound by
          the type signature for x :: i -> String at <interactive>:3:17
   [...]
```

The `i is a rigid type variable bound by the type signature...` stuff could potentially make some errors from psc much easier to understand, I think.

paf31:

Fixes #1111


garyb:
Looks nice, this should be really helpful in some situations.

paf31:
Thanks!

PR #1657, opened 2015-11-22
Warn on duplicate imports

### LINKED ISSUE #1390: Warn on duplicate imports
Conditions for this:

- [x] Two whole-module imports of the same module
- [x] A whole-module import and a partial-import of the same module
- [x] Overlapping items in two partial-imports of the same module (suggest combining the list)
- [x] Two qualified imports of the same module

garyb:
Resolves #1390

### DIFF
--- a/src/Language/PureScript/Sugar/Names/Imports.hs
+++ b/src/Language/PureScript/Sugar/Names/Imports.hs
@@ -24,9 +24,9 @@ module Language.PureScript.Sugar.Names.Imports
 import Prelude ()
 import Prelude.Compat
 
-import Data.List (find)
-import Data.Maybe (fromMaybe, isNothing, fromJust)
-import Data.Foldable (traverse_)
+import Data.List (find, delete)
+import Data.Maybe (fromMaybe, isJust, isNothing, fromJust)
+import Data.Foldable (traverse_, for_)
 
 import Control.Arrow (first)
 import Control.Monad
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/Sugar/Names/Imports.hs
```diff
@@ -70,9 +70,36 @@ findImports = foldM (go Nothing) M.empty
...
+              maybe id warnWithPosition pos $
+                tell $ errorMessage $ DuplicateSelectiveImport mn
+
+    let scope = M.insert currentModule [(Nothing, Implicit, Nothing)] imports
     foldM (resolveModuleImport currentModule env) nullImports (M.toList scope)
 
+  where
+  selfCartesianSubset :: [a] -> [(a, a)]
+  selfCartesianSubset (x : xs) = [(x, y) | y <- xs] ++ selfCartesianSubset xs
+  selfCartesianSubset [] = []
```
garyb: Surely this must be a thing that exists elsewhere?

garyb:
When it comes to qualified imports it only warns if there is an exact duplicate match. There is at least one valid reasons for why a module might be imported qualified more than once (importing into pseudo-module for exports as well as an import that is used within in the module somewhere).

src/Language/PureScript/Sugar/Names/Imports.hs
paf31: `replicateM 2` is pretty close, but not the same since it returns `[[a]]`.

garyb:
Oops, will fix whatever is wrong first :smile:

garyb:
Ok, that should do it.

paf31:
:+1: Looks nice, thanks!

PR #1658, opened 2015-11-22
Fix two cases where errors were missing context

### LINKED ISSUE #1637: It's unclear where some error messages occur now
For example, an conflicting import error gives no context as to where the conflict occurs.

```
Error found:

  Cannot declare "_accessType", since another declaration of that name was imported from module Notebook.Component.State
  Consider hiding "_accessType" when importing Notebook.Component.State:

    import Notebook.Component.State hiding (_accessType)



See https://github.com/purescript/purescript/wiki/Error-Code-ConflictingImport for more information,
```

garyb:
Resolves #1637. There may be other cases, but the two mentioned in there are fixed at least.


paf31:
:+1: Looks good!

PR #1659, opened 2015-11-23
Fix #1185, fix #1369, add everythingWithScope traversal to correct some scoping issues.

### LINKED ISSUE #1185: Unexpected "may not be defined" Error
I wrote the following code:
```
module Main where

data Person = Person String Int

getName p = case p of
  (Person name 20) -> name
  _ -> "Unknown"

name = getName (Person "Jimmy" 20)
```
To my surprise, it failed with:
```
Compiling Main
Error found:
Error in module Main:
Error in binding group getName, name:
Error at /home/michael/dev/me/purescript/examples/passing/InternalBindingNameShouldExistAtTopLevel.purs line 9, column 8 - line 9, column 34:
  getName may not be defined in the current scope
See https://github.com/purescript/purescript/wiki/Error-Code-NameNotInScope for more information, or to contribute content related to this error.
```
If I change `name =` to something like `jimmyName =` everything works

Thoughts?

### LINKED ISSUE #1369: Shadowing warning doesn't apply to let bindings
This code does not throw "Duplicate value declaration" (or any other) error:
```purescript
foo :: forall e. Eff (console :: CONSOLE | e) Unit
foo = do
  a <- pure 5
  log $ "a is " ++ (show a)
  a <- pure 10
  log $ "a is " ++ (show a)
  let a = 5
  log $ "a is " ++ (show a)
  let a = 10
  log $ "a is " ++ (show a)
```
And if you run it you'll see:

    a is 5
    a is 10
    a is 5
    a is 10

paf31:
This isn't quite right, but it's pretty close.

### DIFF
--- a/examples/passing/1185.purs
+++ b/examples/passing/1185.purs
@@ -1,0 +1,13 @@
+module Main where
+
+data Person = Person String Boolean
+
+getName :: Person -> String
+getName p = case p of
+              Person name true -> name
+              _ -> "Unknown"
+
+name :: String
+name = getName (Person "John Smith" true)
+
+main = Control.Monad.Eff.Console.log "Done"
--- a/src/Language/PureScript/AST/Traversals.hs
+++ b/src/Language/PureScript/AST/Traversals.hs
@@ -401,8 +401,12 @@ everythingWithScope ::
   (S.Set Ident -> Binder -> r) ->
   (S.Set Ident -> CaseAlternative -> r) ->
   (S.Set Ident -> DoNotationElement -> r) ->
-  S.Set Ident -> Declaration -> r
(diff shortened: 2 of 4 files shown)


paf31:
Ok, this is good to go, I think.

paf31:
Works with Halogen, and correctly identifies a few new shadowed names in `Data.Array` and `Control.Monad.Rec.Class`.

garyb:
Looks good to me. Shame we needed a new traversal for it, but it makes sense :+1:

paf31:
Yeah, it's a bit of unfortunate, but none of the existing traversals really did what I needed.

PR #1660, opened 2015-11-23
Check int values are within range in codegen

### LINKED ISSUE #1094: Should Int literals be `|0`d?
It's currently possible to construct `Int` values outside the 32-bit range using literals:

```
> let y = 99999999999999999 :: Int
> y
100000000000000000

> y + 1
1569325056
```

Should `Int` literals come out as `x|0` to avoid this? Alternatively, should it be an error to write an `Int` literal outside the 32-bit range?

garyb:
Resolves #1094

### DIFF
--- a/examples/failing/IntOutOfRange.purs
+++ b/examples/failing/IntOutOfRange.purs
@@ -1,0 +1,6 @@
+-- @shouldFailWith IntOutOfRange
+
+module Main where
+
+n :: Int
+n = 35028715023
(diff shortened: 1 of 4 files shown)


paf31:
Looks good, but what do we want to do about supporting multiple backends with this? Presumably this code will be different, so it's ok?

garyb:
I made an accomodation for this by adding a backend name and min/max values in the error message so we can tweak it as necessary. The check is in the JS codegen specifically so it's not going to interfere with other backends as it stands.

paf31:
:+1: Sounds good to me.

paf31:
Only two years too late @coveralls 😄

PR #1663, opened 2015-11-23
Fix up shadowed module names in JS codegen

### LINKED ISSUE #867: duplicated names of imported module and adt
suppose I have 

    import Foo
    data Bar = Foo Number | Baz Number 

then it will be compiled to

    var Foo = require("Foo");
    var Foo = (function() {
      function Foo(value0) {this.value0 = value0}; 
      Foo.create = function(value0) {return new Foo(value0)}; 
      return Foo; 
    })();

I think this is bug

garyb:
Resolves #867.

### DIFF
--- a/examples/passing/ShadowedModuleName.purs
+++ b/examples/passing/ShadowedModuleName.purs
@@ -1,0 +1,15 @@
+module Test where
+
+  data Z = Z String
+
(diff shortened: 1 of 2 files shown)


paf31:
:+1: Looks good.

PR #1665, opened 2015-11-24
Fix #1664, check kind before expanding wildcards.

### LINKED ISSUE #1664: Invalid argument to infer
I get the following error message from psc
```
psc: An internal error ocurred during compilation: "Invalid argument to infer"
psc: thread blocked indefinitely in an MVar operation
```

when trying to compile this code:
``` purescript
module Main where

import Prelude
import Control.Monad.Eff

data Identity a = Identity a

newtype IdentityEff e a = IdentityEff (Eff e (Identity a))

test :: forall e a. IdentityEff e a -> IdentityEff e Unit
test action = Identity $ do
  (Identity x :: Identity _ a) <- action
  return $ Identity unit
```

Some comments:
- This is obviously not the code I wanted to write :smile:. It's just the minimal example I could find which reproduces the error.
- The problem is the type annotation `Identity _ a` in the second-to-last line.
- Without this type annotation, the code does *not* compile (I actually can't figure out why) and gives a proper compiler error:
```
Could not match type

    Identity _1

  with type

    a0
```

paf31:


garyb:
:+1:

PR #1666, opened 2015-11-24
Fix #1662, display extra type info in instance errors

### LINKED ISSUE #1662: More hints for error with missing signature of top-level declaration which contains a type class constraint
Can we provide an additional hint when "No type class instance was found for" error is suspected to be connected to a restriction described here https://github.com/purescript/purescript/wiki/Differences-from-Haskell#type-classes-and-type-inference ? Just an additional link in message when no typeclass instance was found for top-level declaration without type signature. Right now it is superconfusing error for beginners like me.

paf31:
The error now looks like

```text

  Main.EQ _0
          (S Z)
          T

where _0 is an unknown type
```

and in general, type info immediately follows the error summary, and comes before error hints.

This also fixes some issues in type variable renaming.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -370,6 +370,15 @@ onTypesInErrorMessageM f (ErrorMessage hints simple) = ErrorMessage <$> traverse
   gSimple (ExprDoesNotHaveType e t) = ExprDoesNotHaveType e <$> f t
   gSimple (CannotApplyFunction t e) = CannotApplyFunction <$> f t <*> pure e
   gSimple (InvalidInstanceHead t) = InvalidInstanceHead <$> f t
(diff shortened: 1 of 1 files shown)


garyb:
Do you think it's worth mentioning that adding a type annotation may help when there are unknowns involved?

This is great though. The refactoring you did to add the hinting mechanism is really paying off!

paf31:
I was thinking it might be best to leave that for the wiki page, since it might change soon once fundeps are added. Also, in some cases, the reason _why_ the unknown is not inferrable is quite subtle, so I'd want to explain it in some detail.

paf31:
@garyb How about this?

garyb:
Nice :+1: 

Does adding the additional hint there not suffer from the same drawback as you mentioned above? I guess not, as in this case the unknown cannot ever work without some kind of extra type info.

paf31:
I don't know if it's an issue, but I think we can add extra info to the wiki now, to explain the new error message.

PR #1668, opened 2015-11-24
Remove unnecessary elaborateImports from Names

garyb:
This isn't required anymore as we lookup instances globally rather than referring to the module imports, and then we re-do this in `CoreFn` to ensure the appropriate imports are present for codegen.


paf31:
:beer: Sounds good!

PR #1671, opened 2015-11-25
Fix #1523, add --json-errors flag.

### LINKED ISSUE #1523: Support JSON errors

paf31:
@kRITZCREEK @nwolverson @garyb How does this look?

Thanks.

### DIFF
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -27,6 +27,9 @@ import Control.Monad.Writer.Strict
 import Data.List (isSuffixOf, partition)
 import Data.Version (showVersion)
 import qualified Data.Map as M
+import qualified Data.Aeson as A
+import qualified Data.ByteString.Lazy as B
+import qualified Data.ByteString.UTF8 as BU8
 
 import Options.Applicative as Opts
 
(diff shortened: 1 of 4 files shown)


kRITZCREEK:
Looks good from skimming the changes. I can't test it before this weekend though. One thing that seems strange is to have the types for errors inside the Main module. Other than that super happy to get JSON errors ;) :+1:

psc/Main.hs
```diff
@@ -46,39 +51,90 @@ data PSCMakeOptions = PSCMakeOptions
...
+data JSONResult = JSONResult
+  { warnings :: [JSONError]
+  , errors :: [JSONError]
+  }
+
+$(A.deriveJSON A.defaultOptions ''ErrorPosition)
+$(A.deriveJSON A.defaultOptions ''JSONError)
+$(A.deriveJSON A.defaultOptions ''JSONResult)
+
+-- | Argumnets: verbose, use JSON, warnings, errors
```
garyb: Typo :smile:

src/Language/PureScript/Errors.hs
```diff
@@ -184,6 +185,23 @@ data HintCategory
...
+-- | Get the source span for an error
+errorSpan :: ErrorMessage -> SourceSpan
+errorSpan = fromMaybe (error "errorSpan: no position information") . findHint matchSpan
+  where
+  matchSpan (PositionedError ss) = Just ss
+  matchSpan _ = Nothing
+
+-- | Get the module name for an error
+errorModule :: ErrorMessage -> ModuleName
+errorModule = fromMaybe (error "errorModule: no position information") . findHint matchModule
```
garyb: Is it safe to do this? I know we _should_ have module info included with every warning and error, but I'm not 100% confident that they all do right now.

garyb:
Looks good, modulo comments :+1:

nwolverson:
Good start...

The message still includes the lines "in module Blah" "at file.purs line...." which is now redundant. Also the _See wiki_ line which I don't think is required when the error code is provided.

src/Language/PureScript/Errors.hs
paf31: Probably not. We should really change the error message type to enforce these few things, but for now, I'll just change this to use `Maybe`.

paf31:
@kRITZCREEK Yeah, I'll probably pull that out into its own `JSON` module.

@nwolverson I'll try to remove those.

Thanks everyone for the feedback on this.

FrigoEU:
Hey, I'm super excited about this so I'm working with this branch now and made a syntastic (for vim) plugin for it. I noticed that there's some things not working yet. Should I log them here, or is it known that it's not finished yet? Maybe take a stab at it myself?

paf31:
I just pushed a couple of fixes, but if you see anything else, yes, please log it.

FrigoEU:
Just a few things I found:

1. "The type declaration for x should be followed by its definition": The line/col numbers in the JSON error point to the first statement after the module definition, while the line/col numbers in the text error point to the correct lines.
2. Parsing errors don't seem to return json errors yet.
3. As mentioned above, the file, line and column information in the message property is less useful in JSON errors.
4. Type unification errors don't have the correct position info in the json position property, while they do have the correct one in the error text.

paf31:
@FrigoEU Thanks! I'm looking forward to syntastic support. The error locations should be fixed now.

paf31:
I think this is ready for review again. I just need to fix the issue reported by @nwolverson.

Another idea was to use the JSON response to report which modules were rebuilt. I'm not sure if we should deal with that here or not.

FrigoEU:
The issues I reported above seem to all be fixed.

paf31:
@garyb Shall I merge this one then?

garyb:
Yeah, definitely! :+1:

Issue #547, opened 2014-08-06
Use Text instead of String


paf31:
Some unicode characters are not being handled properly apparently.

paf31:
Moving to 0.8 since I think this can be classified as "performance".

aspidites:
I'll be taking a look at this soon (this week) unless someone else is interested in it.

faineance:
I'd like to have a try at this.

paf31:
That would be great!

zjhmale:
@paf31 i just tried all these unicode [here](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Parser/Lexer.hs#L191-200) and all works fine. so, is this issue be fixed already?

hdgarrood:
Not yet; even though I think most (hopefully all? 😉 ) of the unicode-related issues have been fixed, there's still the issue that the compiler extensively uses the `String` type internally, which is not a very good choice for representing textual data.

zjhmale:
@hdgarrood can you give some cases related to this issue? thx :)

hdgarrood:
Sorry, I don't understand the question. What do you mean by "case"?

zjhmale:
@hdgarrood sorry 😂 , i mean you said that 

>there's still the issue that the compiler extensively uses the String type internally, which is not a very good choice for representing textual data.

so can you give some code snippets that will cause the unicode issue? then maybe i can find the entry point to fix it. i just tried some but all works good. 😂

kRITZCREEK:
@zjhmale Haskell's `String` type handles unicode allright, the problem is that `[Char]` is a very inefficient representation for textual data. Haskell's `text` package provides the `Text` datatype and associated operations which use stream fusion to greatly improve performance.

zjhmale:
@kRITZCREEK got it, thx!

kRITZCREEK:
I'd like to tackle this one :) *BUT* I want two concessions:

1. OverloadedStrings
2. Protolude/our own custom Prelude which banishes String as much as possible

I know there was pushback around 1. because of inference problems and 2. is quite a big step no doubt, but the boilerplate will make this a drag otherwise.

This will also cause conflicts with pretty much every other PR, are we up for the extra work?^^

garyb:
[...]

PR #1674, opened 2015-11-25
fixes purescript/#1673

mgmeier:
Among other solutions

* using a different data type (e.g ```Data.Text```)
* using strict I/O (package strict-io)

this fix is the least invasive.

It's tested on my system and works. It forces ```readFile``` to advance till reaching EOF (or an exception) and change the corresponding file handle's state, so that it can be GC'd.


paf31:
Neat, thanks! :+1: 

Perhaps it makes sense to put this in its own function, but could you please document it with a line comment in either case?

hdgarrood:
As an aside, I think switching to `Text` is probably a good idea, and I think there's an issue open for it somewhere anyway? #547? Not sure if that's related.

mgmeier:
comment added.
IMHO it's preferable to leave ```length js``` in the same closure as ```js <- readFile ...```, so we prevent leaking of lazily read content in any case. This also guarantees the file handle being marked as "semi-closed" by the Haskell runtime before the next iteration, so it doesn't interfere with Mac OS 's crappy resource management.

mgmeier:
and yeah, +1 for switching to ```Text``` ;)

paf31:
Could you please include yourself in https://github.com/purescript/purescript/blob/master/CONTRIBUTORS.md?

mgmeier:
(y)

garyb:
Thanks! Sorry about the delay in merging, Phil is busy at the moment and I hadn't been keeping track of this.

mgmeier:
no problem. looking forward to the next release :+1:

PR #1678, opened 2015-11-28
Remove unhelpful isExplicitQualModule check

### LINKED ISSUE #1499: Qualified pseudo-modules don't merge with actual modules
This is kinda weird, but I think the current behaviour is incorrect:

```
module A where
  a = 1

module X where
  x = 1

module Test where

  import qualified X as A

  thing = A.a -- fails
```

This example makes it look pretty odd, but when `a` is `DOM` for instance, this can happen easily. e.g. slamdata/purescript-browserfeatures#10

garyb:
Resolves #1499

The check disallows the case mentioned in the issue but doesn't gain us anything, as we go on to check if the referenced module exists regardless of whether it was a pseudo module or not.


paf31:
:+1:

PR #1682, opened 2015-11-29
Show identifiers correctly in ctags

nwolverson:
Was using show and for identifiers we would see in the tags file

    Ident "foo"

Change to show

    foo


paf31:
Thanks!

PR #1684, opened 2015-11-30
Introduce class keyword for refs, allow type refs without ()

### LINKED ISSUE #1508: Allow importing of types without the use of () if no constructors are required
Currently the behaviour of an import for a proper name is dictated by syntax, but that's basically due to laziness on my part. Also, importing `X()` is a bit pointless, especially in the case of synonyms where `()` can only ever be empty.

garyb:
Resolves #1508

Or will do when I fix whatever stupid mistake I've made that is breaking the tests.

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -321,12 +321,12 @@ handleShowImportedModules = do
   refsList refs = "(" ++ commaList (map showRef refs) ++ ")"
 
   showRef :: P.DeclarationRef -> String
-  showRef (P.TypeRef pn dctors) = show pn ++ "(" ++ maybe ".." (commaList . map N.runProperName) dctors ++ ")"
-  showRef (P.ValueRef ident) = show ident
-  showRef (P.TypeClassRef pn) = show pn
-  showRef (P.ProperRef pn) = show pn
-  showRef (P.TypeInstanceRef ident) = show ident
-  showRef (P.ModuleRef name) = "module " ++ show name
+  showRef (P.TypeRef pn dctors) = N.runProperName pn ++ "(" ++ maybe ".." (commaList . map N.runProperName) dctors ++ ")"
+  showRef (P.ValueRef ident) = N.runIdent ident
+  showRef (P.TypeClassRef pn) = N.runProperName pn
+  showRef (P.ProperRef pn) = N.runProperName pn
+  showRef (P.TypeInstanceRef ident) = N.runIdent ident
+  showRef (P.ModuleRef name) = "module " ++ N.runModuleName name
   showRef (P.PositionedDeclarationRef _ _ ref) = showRef ref
 
   commaList :: [String] -> String
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -1,23 +1,11 @@
------------------------------------------------------------------------------
---
--- Module      :  PSCi
--- Copyright   :  (c) 2013-15 Phil Freeman, (c) 2014-15 Gary Burgess
--- License     :  MIT (http://opensource.org/licenses/MIT)
---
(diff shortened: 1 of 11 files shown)


garyb:
There we go. Pesky `Positioned` constructors.

michaelficarra:
You've not added any new tests.

garyb:
True. We can't test for warnings yet, so I can't add tests for that, but I'll add a test for the `class` syntax and type without `()`.

garyb:
Added tests, thanks for the reminder @michaelficarra.

michaelficarra:
How about in explicit export list?

garyb:
Good thinking, added those too.

psci/PSCi.hs
```diff
@@ -336,6 +324,7 @@ handleShowImportedModules = do
   showRef (P.TypeRef pn dctors) = show pn ++ "(" ++ maybe ".." (commaList . map N.runProperName) dctors ++ ")"
   showRef (P.ValueRef ident) = show ident
```
paf31: Should these really be using `Show`?

src/Language/PureScript/Errors.hs
```diff
@@ -781,6 +771,14 @@ prettyPrintSingleError full level e = do
       paras [ line $ "The import of type " ++ runModuleName name ++ " as " ++ runModuleName qualName ++ " uses the deprecated 'import qualified' syntax."
             , line $ "This syntax form will be removed in PureScript 0.9." ]
 
+    renderSimpleErrorMessage (DeprecatedClassImport mn name) =
+      paras [ line $ "The import of class " ++ runProperName name ++ " from " ++ runModuleName mn ++ " uses the deprecated syntax that omits the class keyword."
```
paf31: Does it make sense to show the new syntax in full for the import in question?

psci/PSCi.hs
garyb: I don't know, but I suspect not!

src/Language/PureScript/Errors.hs
garyb: We'd have to reconstruct the whole import/export declaration to be able to do that in a useful manner I think - what can we show them apart from `class SomeProperName`?

garyb:
Rebased and fixed that `show` usage in `psci`.

src/Language/PureScript/Errors.hs
michaelficarra: It's possible it would be helpful to show that.

garyb: Something like this then?
```
Class import from ModuleName uses deprecated syntax that omits the 'class' keyword:

   ClassName

Should instead use the form:

  class ClassName
```

Likewise for exports:

```
Class export uses deprecated syntax that omits the 'class' keyword:

   ClassName

Should instead use the form:

  class ClassName
```

garyb: I'll go ahead and make this change, it's certainly no worse than my original error message.

garyb:
Updated with the new warnings, also updated the deprecated use of `qualified` warning to follow the same style.

paf31:
:+1: Looks great, thanks!

michaelficarra:
Someone's going to have to update the [language guide](https://github.com/purescript/purescript/wiki/Language-Guide#importing-modules).

garyb:
Yeah, there will need to be a whole bunch of updating soon.

- deprecated `qualified` keyword
- deprecated imports without `class`
- type imports with no constructors no longer require `()`
- `import X` warns without qualification, an explicit list, or `(..)`
- operator aliases and correspondingly deprecated naming any function as an operator

I think we're going to 0.8 next most likely, as we're almost done with the issues in that milestone now.

PR #1685, opened 2015-12-02
Warn on unspecified imports

### LINKED ISSUE #1509: Warn when using "open" imports

garyb:
Resolves #1509

This also combines the changes from #1684, so we should merge that first - I didn't want to base this on `master` due to the amount of conflicts that would need resolving after.

### DIFF
--- a/psci/Completion.hs
+++ b/psci/Completion.hs
@@ -171,7 +171,7 @@ getAllQualifications sho m (declName, decl) = do
   qualificationsUsing (_, importType, asQ') =
     let q = qualifyWith asQ'
     in case importType of
-          P.Implicit      -> [q]
+          P.Implicit _    -> [q]
           P.Explicit refs -> [q | referencedBy refs]
           P.Hiding refs   -> [q | not $ referencedBy refs]
 
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -315,10 +315,11 @@ handleShowImportedModules = do
       Just mn' -> "qualified " ++ N.runModuleName mn ++ " as " ++ N.runModuleName mn'
       Nothing  -> N.runModuleName mn ++ " " ++ showDeclType declType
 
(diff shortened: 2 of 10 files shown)


garyb:
Rebased now #1684 has been merged.

paf31:
Looks nice. What's the motivation for this though? If I understand correctly, `import X` is now equivalent to `import X(..)`, but with a warning, is that right?

paf31:
FWIW I'm still very much in favor of giving a warning with the explicit imports listed out.

garyb:
Yeah, the `(..)` case is for those situations where it would be unbelievably annoying to maintain without an "import everything", such as when you have modules that only exist to re-export other modules (`Data.Lens`, for one).

The idea being if you want to develop without having to list all the stuff, you use the normal `import X`, and then once you're done you can just copy the import list out of the warning message. We never suggest using `import X (..)` in the message, only suggest the explicit import list, as obviously it should be used sparingly otherwise it's no better than not having the warning at all.

paf31:
I see. Could you please paste an example warning?

garyb:
```
Warning 956 of 1007:

  in module Halogen.HTML.Events
  at purescript-halogen\src\Halogen\HTML\Events.purs line 9, column 1 - line 11, column 1

    Module Prelude has unspecified imports, consider using the explicit form:

      import Prelude (Unit, bind, pure, ($))


```

paf31:
Neat :+1:

paf31:
You've been really busy :smile:

garyb:
Most of these changes were pretty easy to do actually! :)

PR #1688, opened 2015-12-05
Delay errors about scope conflicts until they are unavoidable

### LINKED ISSUE #1497: Conflicting imports
I can run this code in Haskell
```
module Foo where
test = 1

module Bar where
test = 2

module Main where
import Foo
import Bar
main = print "test"
```

But PureScript compiler return an error

```Conflicting imports for test from modules Bar and Foo```

Is this a bug?

garyb:
Resolves #1497

### DIFF
--- a/examples/failing/ConflictingExports.purs
+++ b/examples/failing/ConflictingExports.purs
@@ -1,0 +1,16 @@
+-- @shouldFailWith ScopeConflict
+module A where
+
+  thing :: Int
+  thing = 1
+
+module B where
+
+  thing :: Int
+  thing = 2
+
+-- Fails here because re-exporting forces any scope conflicts to be resolved
+module Main (module A, module B) where
+
+  import A
+  import B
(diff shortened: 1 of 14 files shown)


garyb:
Also rebased now #1684 has been merged, but will probably have to do so again after #1685 :smile:

zudov:
It still gives a warning, right?

garyb:
No, it errors when things are brought into conflict by making an ambiguous reference rather than when two potentially conflicting references just happen to be in scope but are not used. Re-exporting also forces errors about conflicts.

zudov:
Yeah, but probably it would make sense to produce a warning in the cases when conflicting names are imported, and of course error only when there is an actual conflict?

garyb:
Ah. I didn't really think it warranted a warning, as it's unambiguously wrong or not. Warnings are mostly for cases where it looks like there may be a mistake being made. Also, with #1685 we'll be warning about indiscriminate importing anyway, which is the case that this most commonly arises in.

paf31:
:+1: Looks nice, and I agree this is the better way for the errors to work.

PR #1689, opened 2015-12-05
Fix psc-publish test

passy:
psc-publish wasn't updated to the signature changes in
`preparePackage` which now expects PublishOptions as argument.

(Did this as a preparation for a 24 pull requests submission. /cc @codebar @ladieswhocode)


paf31:
:+1: Thanks!

PR #1691, opened 2015-12-06
Introduce operator aliases

### LINKED ISSUE #806: Aliases for operators

garyb:
Resolves #806

This probably isn't the best way of handling it in the long term, it would be nice to desugar during codegen rather than upfront as we'll probably get better error messages that way, but while we're supporting both old and new syntax this seemed a reasonable approach for now.

### DIFF
--- a/src/Language/PureScript/Docs/Convert.hs
+++ b/src/Language/PureScript/Docs/Convert.hs
@@ -11,6 +11,9 @@ module Language.PureScript.Docs.Convert
   , collectBookmarks
   ) where
 
+import Prelude ()
+import Prelude.Compat
+
 import Control.Monad
 import Control.Category ((>>>))
 import Data.Either
@@ -68,7 +71,7 @@ type IntermediateDeclaration
 -- with their associativity and precedence.
 data DeclarationAugment
   = AugmentChild ChildDeclaration
-  | AugmentFixity P.Fixity
+  | AugmentFixity P.Fixity (Maybe P.Ident)
 
 -- | Augment top-level declarations; the second pass. See the comments under
 -- the type synonym IntermediateDeclaration for more information.
@@ -86,8 +89,8 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
     case a of
       AugmentChild child ->
         d { declChildren = declChildren d ++ [child] }
-      AugmentFixity fixity ->
-        d { declFixity = Just fixity }
+      AugmentFixity fixity alias ->
+        d { declFixity = Just (fixity, P.runIdent <$> alias) }
 
 -- | Add the default operator fixity for operators which do not have associated
 -- fixity declarations.
@@ -97,7 +100,7 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
 addDefaultFixity :: Declaration -> Declaration
 addDefaultFixity decl@Declaration{..}
   | isOp declTitle && isNothing declFixity =
-        decl { declFixity = Just defaultFixity }
+        decl { declFixity = Just (defaultFixity, Nothing) }
   | otherwise =
         decl
   where
@@ -113,7 +116,7 @@ getDeclarationTitle (P.ExternDataDeclaration name _)         = Just (P.runProper
 getDeclarationTitle (P.TypeSynonymDeclaration name _ _)      = Just (P.runProperName name)
 getDeclarationTitle (P.TypeClassDeclaration name _ _ _)      = Just (P.runProperName name)
 getDeclarationTitle (P.TypeInstanceDeclaration name _ _ _ _) = Just (P.showIdent name)
-getDeclarationTitle (P.FixityDeclaration _ name)             = Just ("(" ++ name ++ ")")
+getDeclarationTitle (P.FixityDeclaration _ name _)           = Just ("(" ++ name ++ ")")
 getDeclarationTitle (P.PositionedDeclaration _ _ d)          = getDeclarationTitle d
 getDeclarationTitle _                                        = Nothing
 
@@ -170,8 +173,8 @@ convertDeclaration (P.TypeInstanceDeclaration _ constraints className tys _) tit
 
   childDecl = ChildDeclaration title Nothing Nothing (ChildInstance constraints classApp)
   classApp = foldl P.TypeApp (P.TypeConstructor className) tys
-convertDeclaration (P.FixityDeclaration fixity _) title =
-  Just (Left ([title], AugmentFixity fixity))
+convertDeclaration (P.FixityDeclaration fixity _ alias) title =
+  Just (Left ([title], AugmentFixity fixity alias))
 convertDeclaration (P.PositionedDeclaration srcSpan com d') title =
   fmap (addComments . addSourceSpan) (convertDeclaration d' title)
   where
--- a/src/Language/PureScript/Docs/Types.hs
+++ b/src/Language/PureScript/Docs/Types.hs
@@ -84,7 +84,7 @@ data Declaration = Declaration
   , declComments   :: Maybe String
   , declSourceSpan :: Maybe P.SourceSpan
   , declChildren   :: [ChildDeclaration]
-  , declFixity     :: Maybe P.Fixity
+  , declFixity     :: Maybe (P.Fixity, Maybe String)
   , declInfo       :: DeclarationInfo
   }
   deriving (Show, Eq, Ord)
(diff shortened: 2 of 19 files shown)


garyb:
I've done some testing with this too - made a local version of the prelude that has all its operators defined as aliases and then compiled a Halogen example that still works succesfully. Also checked the output is correct in the presence of re-exports of aliases.

Once this and my other outstanding PRs are in I'd like to revisit purescript/purescript-prelude#41, update the compiler for any changed identifier locations, and then release a prelude 1.0.0 to go along with the psc 0.8 release, if that sounds good?

src/Language/PureScript/Docs/Types.hs
```diff
@@ -307,9 +307,12 @@ asDeclaration =
...
               <*> key "info" asDeclarationInfo
 
-asFixity :: Parse PackageError P.Fixity
-asFixity = P.Fixity <$> key "associativity" asAssociativity
-                    <*> key "precedence" asIntegral
+asFixity :: Parse PackageError (P.Fixity, Maybe String)
+asFixity = do
+  fixity <- P.Fixity <$> key "associativity" asAssociativity
+                     <*> key "precedence" asIntegral
+  alias <- key "alias" (perhaps asString)
```
paf31: So this needs an update to Pursuit, correct? Or is this a backwards-compatible change? Does `perhaps` handle undefined fields?

Maybe we need to bump the min compiler version there. /cc @hdgarrood

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -124,8 +124,9 @@ parseFixityDeclaration :: TokenParser Declaration
 parseFixityDeclaration = do
   fixity <- parseFixity
   indented
+  alias <- P.optionMaybe $ parseIdent <* reserved "as"
```
paf31: `parseIdent` -> `lname`? We don't want operators here, right?

src/Language/PureScript/Docs/Types.hs
garyb: I'm not sure, so yeah, one for @hdgarrood.

src/Language/PureScript/Sugar/Operators.hs
```diff
@@ -1,22 +1,3 @@
...
--- Module      :  Language.PureScript.Sugar.Operators
--- Copyright   :  (c) 2013-15 Phil Freeman, (c) 2014-15 Gary Burgess
--- License     :  MIT (http://opensource.org/licenses/MIT)
---
--- Maintainer  :  Phil Freeman <paf31@cantab.net>
--- Stability   :  experimental
--- Portability :
---
--- |
--- This module implements the desugaring pass which reapplies binary operators based
```
paf31: Worth keeping the comments?

src/Language/PureScript/Sugar/Operators.hs
```diff
@@ -57,9 +40,24 @@ import qualified Language.PureScript.Constants as C
...
+
+  makeLookupEntry :: (Qualified Ident, SourceSpan, Fixity, Maybe Ident) -> Maybe (Qualified Ident, Qualified Ident)
+  makeLookupEntry (qname@(Qualified qual _), _, _, Just alias) = Just (qname, Qualified qual alias)
+  makeLookupEntry _ = Nothing
+
+  renameAliasedOperators :: M.Map (Qualified Ident) (Qualified Ident) -> Module -> Module
+  renameAliasedOperators aliased (Module ss coms mn ds exts) = Module ss coms mn (map f' ds) exts
+    where
+    (f', _, _) = everywhereOnValues id go id
+    go (Var name) = Var $ fromMaybe name (name `M.lookup` aliased)
```
paf31: If the value is not in the map, should that be an `internalError`?

paf31:
I like it! I tried this before but didn't think of the renaming-after-reordering approach for some reason, which threw me off. 

Agreed about handling this in codegen, but this is great for now.

Prelude changes sound good to me, but why 1.0? Don't we want to wait until other things are ready and bump everything together?

src/Language/PureScript/Sugar/Operators.hs
garyb: No, because this matches all `Var`s, the map only has the aliased operators in it, so if the lookup fails it just means it's not an operator, or one that was defined the old way.

src/Language/PureScript/Sugar/Operators.hs
garyb: Oops, yep. I did so with the other ones I took the header out of but missed it here.

src/Language/PureScript/Sugar/Operators.hs
paf31: Oh right. But if we matched `Op` only, then in 0.9 it could be an error?

src/Language/PureScript/Parser/Declarations.hs
garyb: Ah, good point!

src/Language/PureScript/Sugar/Operators.hs
garyb: Yeah exactly :) There are quite a few details of this that can be simplified in 0.9 I think.

garyb:
Updated - `identifier` is the one we wanted rather than `lname` I think, as that covers `_`-prefixed idents too. Added the comments back to the operators sugar module.

It would seem that using `perhaps` in the aeson stuff will probably be backwards compatible [given the description here](https://hackage.haskell.org/package/aeson-better-errors-0.9.0/docs/src/Data-Aeson-BetterErrors-Internal.html#perhaps), but Harry should definitely know given it's his library :smile:

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -97,7 +100,7 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
 addDefaultFixity :: Declaration -> Declaration
 addDefaultFixity decl@Declaration{..}
   | isOp declTitle && isNothing declFixity =
-        decl { declFixity = Just defaultFixity }
+        decl { declFixity = Just (defaultFixity, Nothing) }
```
hdgarrood: Just checking: once we require operators to be defined as aliases for other existing values, this `addDefaultFixity` function will no longer be necessary, right?

src/Language/PureScript/Docs/Types.hs
hdgarrood: Apparently not, `perhaps` is kind of like purescript's `Data.Foreign.Null.readNull`: https://hackage.haskell.org/package/aeson-better-errors-0.9.0/docs/Data-Aeson-BetterErrors.html#v:perhaps
But there is also `keyMay`, which would give you `Nothing` if the property wasn't there, so as far as I can tell, if we used that, we would probably be able to keep accepting uploads from people on 0.7.x. I'm not really sure about what to do here; it's hard to predict whether that would be worth doing.

By the way, does this line up with the `ToJSON` instance? You can run the test in `psc-publish/test` to do a quick to-JSON-and-back test. (I should make that into an actual test.) I think this might be using Aeson's `ToJSON (,)` instance at the moment, which might not work with this parser.

src/Language/PureScript/Docs/Convert.hs
garyb: Yeah, that's right.

garyb:
Switched to use `keyMay` instead of `perhaps`. I couldn't get the `psc-publish` tests to work though, so I'm not sure if the mentioned JSON instance issue is a problem or not.

hdgarrood:
Ok - I'll try to fix those tests up and integrate them into the normal test suite.

hdgarrood:
@garyb can you cherry-pick https://github.com/hdgarrood/purescript/commit/4ab687c69de1b6ea3695c31ded47bb1663cc7bb5?

garyb:
@hdgarrood done! (A while back actually, Travis is pretty slow currently). Weirdly failing in just one of the test setups...

hdgarrood:
Ah, looks like that new module `TestPscPublish` needs to be added to the `other-modules` section in the cabal file. Although I have no idea why it only failed once...

phadej:
@hdgarrood because `RUNSDISTTESTS=YES` is only in one travis job. There are almost no sense to run it with all compiler versions, would just make builds slower.

hdgarrood:
Ah, of course - thanks.

garyb:
That did it, thanks. I think this is ready to go if you're happy then @paf31?

paf31:
:+1: Looks great!

garyb:
:confetti_ball:

Issue #121, opened 2014-01-20
Source Maps


paf31:

darrrk:
:+1:

LiamGoodacre:
This would be really nice to have :]

dkoontz:
As an anecdotal addition to this discussion, when I floated PS as a potential successor to vanilla JS amongst my team, lack of sourcemaps was raised as one of the primary concerns. I know runtime exceptions shouldn't happen in PS, but being able to set a breakpoint and step through code is viewed as an indispensable feature to have in a language. I'm not saying it actually is a necessity just that I see it perceived that way, especially from people transitioning from a JS mindset. So from a marketing/adoption standpoint sourcemaps might have a lot of value even though they may not have as much day-to-day value.

garyb:
Agreed, that's pretty much what I was trying to say [here](https://github.com/purescript/purescript/pull/1693#issuecomment-162364507) :smile:

codygman:
I think having source maps would be amazing for purescript adoption.

AppShipIt:
@codygman 👍 Another oft ignored use case is inspecting data at runtime. Could its infrastructure also help the compiler expose types of all expressions to tools by tracking source locations?

TehShrike:
When a compile-to-JS language doesn't have source maps, that signals to me "we expect you to have to debug our transpiler output, not the original language".

When a compile-to-JS language has first-class source maps, it signals to me "we expect the transpiler to always work and you can just pretend you're writing a different language."

I personally prefer my language abstractions to be non-leaky, so I only use languages with full sourcemap support.

You may find my signal-reading inaccurate, but telling me so is unlikely to change how I read those signals :-x

To me, first-class source map support (viewed from Chrome) includes
1. thrown errors have stack traces with all original file names and line numbers, and clicking on the file name/number opens it in the source tab
2. breakpoints and tracing work without any jumping around, as if the original language was natively supported by Chrome

Tastes vary, I do see plenty of people willing to use compiled JS without good sourcemaps.

But be careful about considering sourcemaps as having "not  as much day-to-day value" - once you're used to accurate line numbers that open the right file when you click on them, and full breakpoint/step support, you get value from them every day.

garyb:
Well, the thing is, you shouldn't be seeing runtime errors from PureScript code anyway... that's one of the reasons why the language exists.

There's a PR with progress on this now anyway: #1693

AppShipIt:
@garyb ^ See tooling question?

[...]

PR #1693, opened 2015-12-06
Add Source Map support.

nwolverson:
Per #121. Seems to work decently for me now testing via https://github.com/lydell/source-map-visualize 

   * Needs cleaned up
   * Need to put behind a command line flag
   * Not sure if there are performance implications, could be bad. Seemed super slow to build my project just now but need to compare against master
   * Could have broken other things by introducing new nodes in the AST, could this prevent some optimisation from firing?
   * For best effect `psc-bundle` would need to support, not sure how hard that is

Did manage to combine these source maps for a larger app with webpack (custom webpack config), I'm getting a runtime error which may or may not be related to my changes.


nwolverson:
Worse than I thought, this is not going to work:

real	0m11.382s
user	0m27.428s
sys	0m4.606s

real	3m55.869s
user	10m27.173s
sys	2m52.620s

paf31:
Very interesting. Do you find the source maps useful for debugging? Maybe it's possible to fix the speed issue.

nwolverson:
Well in general I find source maps useful. e.g. I've used with coffescript, TypeScript, babel, even though you can get away fine without it. Particularly when you have 1 combined output file. I think having clean/straightforward output as per the PureScript design should also mean that source maps work better anyway.

Not had time yet to really experience this implementation since the point it actually worked in a usable way.

I presume some (lesser) performance hit is acceptable when the option is actually enabled but I suspect most of this is encountered at the JS pretty-print stage.

garyb:
I've actually not yet encountered a situation where source maps would be helpful in PureScript - the difference with all the languages you list is they have unchecked runtime exceptions :wink:

I think I may have been the one who suggested source map support originally, so I'm not at all against it, but just thought I'd mention that although they're nice to have I think their utility is perhaps not what I expected - probably their most useful aspect is a selling point for JS devs who know little about PS, that think source maps are required for transpiled JS to be practical.

garyb:
I assume this is the embedded sourcemap style? Perhaps we could get better performance out of generating the separate map file kind instead?

nwolverson:
I actually use dev tools a lot to understand program flow and logic errors when there is no run time error, rather than printf debugging (which eg I did a lot of in preparing this PR as I didn't invest the time to see what better tools are available).

I was wrong my guess, seems like the perf problem is in actual source map generation, adding a flag and turning off the actual generation just gave 13.7s `time`. The maps are not the embedded style. I guess this means the issue is likely in the source-maps library, it really shouldn't be that expensive.

nwolverson:
Found the major issue in `sourcemaps`, with that change:

real	0m13.572s
user	0m34.961s
sys	0m6.025s

purescript.cabal
```diff
@@ -70,7 +70,8 @@ library
                    process >= 1.2.0 && < 1.4,
                    safe >= 0.3.9 && < 0.4,
                    semigroups >= 0.16.2 && < 0.19,
-                   parallel >= 3.2 && < 3.3
+                   parallel >= 3.2 && < 3.3,
+                   sourcemap >= 0.1.3
```
paf31: You'll need to rerun the `license-generator` executable.

paf31:
Not bad at all :smile:

What was the issue?

paf31:
Does this change affect the speed when source maps are not generated?

nwolverson:
Issue in sourcemaps, fixed in latest version `0.1.5`: https://github.com/chrisdone/sourcemap/commit/00e776b7f84e9f750427be8052625ba4edf91c23

Still some performance impact outside of that, including without sourcemaps being generated, was looking at some profiling but have been laid up with a cold.

Can't for the life of me get cabal-dependency-licenses to run.

nwolverson:
Original (median of 3 runs with above project, which does generate about 500 errors)

real	0m11.173s
user	0m27.750s
sys	0m4.486s

With changes, source-maps not enabled (generally seeing 5-10% hit, and I can't seem to improve it):
real	0m12.161s
user	0m29.225s
sys	0m4.968s

With sourcemaps:
real	0m15.844s
user	0m39.553s
sys	0m7.191s

Turns out adding an intermediate node in the JS AST for position information broke both optimisation and   correct parenthesisation, so should maybe have bitten the bullet and added to each constructor. If it's worth continuing with this.

paf31:
A thought on performance: if we had two separate code generators, one for source maps and one for JS, we could choose to run one or both. But obviously we need information from the JS generator to construct the source maps. 

But what if we wrote one polymorphic code generator:

```haskell
class (Monoid gen) <= Emit gen where
  emit :: String -> gen

codegen :: (Emit gen) => JS -> gen
```

(approximately) then give instances of `Emit` for `String` (JS) and some counting data structure (source maps). With some inlining and specialization, that might get close to the performance of separate generators.

Just a thought, but maybe one for another PR. It would definitely be nice to have zero-cost when not using source maps though.

nwolverson:
That does look nicer than trying to build a data structure that works conditionally for both cases as I've done. I wasn't clear where any remaining performance issue for the non-map case was, the only things that I noticed in profiling seemed like they should be an issue already (bunch of string concatenation), I thought maybe there was some effect from additional AST nodes that were inserted pre-codegen.

(I lack Haskell experience to have any good feel for performance issues, style etc...)

nwolverson:
So I redid this to use the `Emit` class suggested, this seems to all get specialised and inlined OK. Also removed the extra `JSPositioned` constructor and folded in the sourcespan to each constructor, still some cruft due to that but can at least be sure the optimiser doesn't get broken. Seeing pretty good performance with source maps off, I disabled all error messages (because they were the majority of the build time!) and ran on the same largish project, couldn't really distinguish performance vs master over the variance between runs.

Still needs work like ensuring a good set of positions are outputted (but some of that is maybe due to bad sourcespans coming from earlier in the compilation process).

src/Language/PureScript/CodeGen/JS/AST.hs
```diff
@@ -126,120 +130,190 @@ data JS
   -- |
   -- A numeric literal
   --
-  = JSNumericLiteral (Either Integer Double)
+  = JSNumericLiteral (Maybe SourceSpan) (Either Integer Double)
```
paf31: Maybe it makes sense to add a type argument here?

paf31:
Looks great so far! I'll take another look tomorrow. What sort of issues are you seeing with position info? Is it possible to drop this into Chrome and use it in the debugger?

src/Language/PureScript/CodeGen/JS/AST.hs
nwolverson: You mean `JS a = JSNumericLiteral a (Either Integer Double)` ? I guess that was my original thought, the type does appear everywhere so maybe an alias `type JS = JS1 (Maybe SourceSpan)` ?

nwolverson:
If you can run directly against CommonJS modules you can use this as-is (I'm doing that for example with atom plugin, would at least be the same in an Electron app). But I set up a demo using webpack which will combine purescript source maps with its own: https://github.com/nwolverson/purescript-sourcemap-test 
screen shot 2016-01-31 at 11 16 19

The issue I'm seeing with positions is in eg `a + b + c`, for e.g. `b` I see a span for the entire expression, not investigated further why.

paf31:
Looks great! Any idea why the CI build is breaking?

nwolverson:
Needed to fix versions of `sourcemaps` used since the version on stackage didn't satisfy the constraint - guess we could just drop the constraint, if you build with an old version and use the flag it will take all day  - but I don't know how official binaries are built.

And can't derive `Typeable` on older ghc versions - not sure where this constraint came from as I seemed to need to add a bunch extra when I rebased onto master. Presume it's something to do with the phantom type. If you know the original change maybe it's possible to remove all these - maybe it was removed on master and I re-added it when I rebased?

garyb:
I removed the `Data`/`Typeable` deriving when we started using `DataKinds` to distinguish between the various meanings of `ProperName` (that's what the older GHCs struggle with, the interaction of `DataKinds` and `DeriveTypeable`).

nwolverson:
Thank @garyb, I guess dropping syb meant those were not required.

paf31:
@nwolverson So would you say this is ready to merge? If so, we should release it with 0.8.1.

nwolverson:
Yes - with the proviso that I'm sure the accuracy of the maps could be improved and no doubt other tweaks, at a later date

src/Language/PureScript/Pretty/Common.hs
```diff
@@ -26,9 +32,87 @@ import Text.PrettyPrint.Boxes
...
+
+-- |
+-- Generalize intercalate slightly for monoids
+--
+intercalate :: Monoid m => m -> [m] -> m
+intercalate x xs = mconcat (intersperse x xs)
+
+class (Monoid gen) => Emit gen where
+  emit :: String -> gen
+  addMapping :: (Maybe SourceSpan) -> gen
```
paf31: Does it make sense to simplify this to `addMapping :: SourceSpan -> gen` and to send `Nothing` to `mempty` in a helper function instead?

paf31:
@garyb I'm going to merge this soon for inclusion in 0.8.1, unless you have any other comments?

Thanks for all the work on this @nwolverson!

AppShipIt:
Wow this is going to be the biggest dot release ever including operator sections :shipit:

src/Language/PureScript/Pretty/Common.hs
nwolverson: Yep that seems cleaner.

nwolverson:
OK, made that 1 change from last comment and rebased.

nwolverson:
Not sure why those builds timed out other than 1 which blew up on memory. How about now...

nwolverson:
As discussed on IRC this is hanging locally since recent rebase; it seems to be hanging on printing errors. I don't see the same on master but it's odd, I have no errors changes.

paf31:
My guess is that this is due to a missing pattern match in the `Pretty` modules.

nwolverson:
So I added a position to atoms to improve source mapping, this seems to have caused infinite recursion between `prettyPrintValueAtom` and `prettyPrintValue`, I'm seeing `Parens` and `BinaryNoParens` ctors - the extra `PositionedValue` may have prevented some desugaring? Removing this for now.

paf31:
@nwolverson Having that extra constructor in there should be fine. I think you can just add the extra pass-through case to `prettyPrintValue(Atom)`

paf31:
Thank you very much!

codygman:
:+1:

PR #1699, opened 2015-12-11
Add native Partial constraint, relax tuple syntax for constraints

### LINKED ISSUE #1669: Parens for constraints
The following is valid Haskell but not PureScript:
```haskell
let f = ("!" ++) <<< show :: forall a. Show a => a -> String
```
PureScript requires parentheses around the constraints when there's only one, unlike Haskell.

Currently, the error you get is not very helpful:
```
(line 1, column 22):
unexpected ::
expecting ., {, `, qualifier or symbol
```

I propose we either improve this error message, or make the parens optional in the case where there is only one constraint - or maybe always optional? is that possible?

### LINKED ISSUE #1694: Take some steps towards Partial support
@paf31 since we're breaking stuff again (in a gentle, thousands-of-warnings kind of way :wink:) perhaps we should try and get `Partial` in too, as I think it shouldn't be too hard.

Proposed steps for 0.8:

- Introduce `Partial` class to `Prim`
- Suppress lint warnings for partial pattern matches for values with a `Partial` constraint
- Update the lint warning to state that partial-patterns-without-`Partial` will become errors in 0.9

It'll be a bit of a faff as I think we lint before typechecking currently, but I'm happy to have a go at it.

garyb:
Resolves #1694, resolves #1669.

- Hides non-exhaustive pattern warning when applied
- Hides exhaustivity bail-out warning when applied
- Adds a warning for when the `Partial` constraint it used but the function is exhaustive.
- Updated the error to state `Partial` constraints for non-exhaustive patterns will be required from 0.9
- Constraints no longer need bracketing, even when comma-separated… maybe we don’t want that? Since we don’t have `ConstraintKind`s or actual tuple syntax it seemed reasonable to do so.

### DIFF
--- a/examples/passing/EmptyTypeClass.purs
+++ b/examples/passing/EmptyTypeClass.purs
@@ -2,11 +2,11 @@ module Main where
 
 import Prelude
 
-class Partial
+class PartialP
 
-head :: forall a. (Partial) => Array a -> a
+head :: forall a. (PartialP) => Array a -> a
 head [x] = x
 
-instance allowPartials :: Partial
+instance allowPartials :: PartialP
 
 main = Control.Monad.Eff.Console.log $ head ["Done"]
--- a/examples/passing/NakedConstraint.purs
+++ b/examples/passing/NakedConstraint.purs
@@ -2,8 +2,6 @@ module Main where
 
 import Control.Monad.Eff.Console
 
-class Partial 
-
 data List a = Nil | Cons a (List a)
 
 head :: (Partial) => List Int -> Int
--- a/examples/passing/UntupledConstraints.purs
+++ b/examples/passing/UntupledConstraints.purs
@@ -1,0 +1,14 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console
+
+data List a = Nil | Cons a (List a)
+
+head :: Partial => List Int -> Int
+head (Cons x _) = x
+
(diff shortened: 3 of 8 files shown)


paf31:
I'm going to try to catch up on PRs today.

garyb:
Saweet! Back home now then?

Once the current batch are in I have a few things I'd like to do that I've been holding off on, as one of them at least will cause a merge nightmare (I'd like to introduce `newtype`s to distinguish `ProperNames` for types from data constructors from classes). I'd also like to add a feature to the test suite to check warnings are behaving as they should - neither of these are "features" though, so don't need to be completed before 0.8.

paf31:
Yep, got back yesterday.

I like those two ideas!

paf31:
Can we break this into two PRs? I have more to say about one than about the other.

For `Partial`: I think this is going to be handled better by the typechecker. As one example, the user might add their own subclass: `class (Partial, Unsafe) <= PartialUnsafe`, and then expect the constraint to be discharged whenever `PartialUnsafe` is in scope. As another, `Partial` might be in scope because of a call to something like `unsafePartial`, and this can't be detected syntactically. Only the typechecker really knows if the constraint can be satisfied.

For constraint parens: I like the idea for a single constraint, but I think I'd prefer to keep the parens for more than one. I don't know why, I think it just scans better from left to right that way.

PR #1703, opened 2015-12-12
Drop paren requirement for constraints

### LINKED ISSUE #1669: Parens for constraints

garyb:
Resolves #1669.

I've left parens optional again here to explain why I think it might be nice: one of the slightly annoying things about the tuple-style syntax is it makes it awkward to break constraints over multiple lines. If parens are optional it makes it a little more natural to do something like this:

``` purescript
someFunc
  :: forall m a
   . Monad m
   , Semigroup (m a)
  => m a
  -> m a
  -> m Unit
```

Plus, you still can use them if you prefer to do so, this is a backwards compatible change.

### DIFF
--- a/examples/passing/UntupledConstraints.purs
+++ b/examples/passing/UntupledConstraints.purs
@@ -1,0 +1,17 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console
+
+class Show a <= Nonsense a where
+  method :: a -> a
+
+data Box a = Box a
+
+instance showBox :: Show a => Show (Box a) where
+  show (Box a) = "Box " <> show a
+
+strangeThing :: forall m. Semigroup (m Unit) => m Unit -> m Unit -> m Unit
+strangeThing x y = x <> y
+
+main = log "Done"
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -33,6 +33,7 @@ import Language.PureScript.Environment
 import Language.PureScript.Errors
 import Language.PureScript.Kinds
 import Language.PureScript.Names
+import Language.PureScript.Types
 import Language.PureScript.Parser.Common
 import Language.PureScript.Parser.Kinds
 import Language.PureScript.Parser.Lexer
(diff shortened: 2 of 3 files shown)


garyb:
Although, if I'd thought about it for 5 seconds longer I would have seen that parens is no big deal either:

``` purescript
someFunc'
  :: forall m
   . ( Monad m
     , Semigroup (m a)
     )
  -> m a
  -> m a
  => m Unit
```

If you think dropping parens would be detrimental I'm fine either way, my main motivation for making this change was to make use of the `Partial` constraint more convenient, it just so happened that relaxing the parens requirement for multiple constraints wasn't ambiguous so I figured I'd just go with it.

garyb:
Actually, I'll extend this to other places that currently require it too: superclass and instance constraints. Maybe it'll be ambiguous in those situations anyway.

garyb:
Ok, well it's possible, but I think I agree it's less pleasant to read.

paf31:
As a paren-free version, there is also `Show a => Eq a => ...` which should work already, but which is slightly different at the value level.

I don't know, I just feel irrationally uneasy about commas without parens, for some reason :smile:

Code looks fine though :+1:

garyb:
Alright, parens are back for multiple constraints :wink:. Can you take another glance?

paf31:
Anyone else have comments on the syntax? I don't want my irrational fear of commas to be the only deciding factor here :smile: 

I thought a bit about why I don't like multiple constraints without parens, and the best reason I came up with was this: once we support something like `ConstraintKinds` (not too far off being possible, actually), we'd be able to make `=>` a type operator just like `->`, and make things much more simple. Then, the tuple punning which GHC uses makes less sense, and I'd rather just have multiple constraints passed just like regular function arguments, as I described above.

`Show a => Eq a => ...` would be a regular nested function, and so would `forall c. c a =>`, etc.

garyb:
Sounds good, but what would we do in class/instance contexts? `class X a <= Y a <= Z a` doesn't really read quite right, does it? As there's no `X <= Y` implication, it is something more like `X, Y <= Z`.

paf31:
Ah right, good point. I'm not sure then actually.

garyb:
I guess this is fine for now anyway, we preserve all existing syntax and add the least controversial version of the relaxed syntax: single constraints. Shall I merge?

paf31:
Yeah, sounds good. I'm happy to discuss other options some more too.

garyb:
:+1:

PR #1704, opened 2015-12-12
Add native Partial constraint

### LINKED ISSUE #1694: Take some steps towards Partial support

garyb:
Resolves #1694.

- Hides non-exhaustive pattern warning when applied
- Hides exhaustivity bail-out warning when applied
- Adds a warning for when the `Partial` constraint it used but the function is exhaustive.
- Updated the error to state `Partial` constraints for non-exhaustive patterns will be required from 0.9

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -19,10 +19,7 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
-module Language.PureScript.Linter.Exhaustive
-  ( checkExhaustive
-  , checkExhaustiveModule
-  ) where
+module Language.PureScript.Linter.Exhaustive (checkExhaustiveModule) where
 
 import Prelude ()
 import Prelude.Compat
@@ -48,7 +45,7 @@ import Language.PureScript.Errors
 
 -- | There are two modes of failure for the redudancy check:
 --
--- 1. Exhaustivity was incomeplete due to too many cases, so we couldn't determine redundancy.
+-- 1. Exhaustivity was incomplete due to too many cases, so we couldn't determine redundancy.
 -- 2. We didn't attempt to determine redundancy for a binder, e.g. an integer binder.
 --
 -- We want to warn the user in the first case.
@@ -239,8 +236,8 @@ missingAlternative env mn ca uncovered
 -- it partitions that set with the new uncovered cases, until it consumes the whole set of clauses.
 -- Then, returns the uncovered set of case alternatives.
 --
-checkExhaustive :: forall m. (MonadWriter MultipleErrors m) => Environment -> ModuleName -> Int -> [CaseAlternative] -> m ()
-checkExhaustive env mn numArgs cas = makeResult . first nub $ foldl' step ([initialize numArgs], (pure True, [])) cas
+checkExhaustive :: forall m. (MonadWriter MultipleErrors m) => Bool -> Environment -> ModuleName -> Int -> [CaseAlternative] -> m ()
+checkExhaustive hasConstraint env mn numArgs cas = makeResult . first nub $ foldl' step ([initialize numArgs], (pure True, [])) cas
   where
   step :: ([[Binder]], (Either RedudancyError Bool, [[Binder]])) -> CaseAlternative -> ([[Binder]], (Either RedudancyError Bool, [[Binder]]))
   step (uncovered, (nec, redundant)) ca =
@@ -258,13 +255,13 @@ checkExhaustive env mn numArgs cas = makeResult . first nub $ foldl' step ([init
 
   makeResult :: ([[Binder]], (Either RedudancyError Bool, [[Binder]])) -> m ()
   makeResult (bss, (rr, bss')) =
-    do unless (null bss) tellExhaustive
+    do unless (hasConstraint || null bss) tellNonExhaustive
        unless (null bss') tellRedundant
        case rr of
-         Left Incomplete -> tellIncomplete
+         Left Incomplete -> unless hasConstraint tellIncomplete
          _ -> return ()
     where
-    tellExhaustive = tell . errorMessage . uncurry NotExhaustivePattern . second null . splitAt 5 $ bss
+    tellNonExhaustive = tell . errorMessage . uncurry NotExhaustivePattern . second null . splitAt 5 $ bss
     tellRedundant = tell . errorMessage . uncurry OverlappingPattern . second null . splitAt 5 $ bss'
     tellIncomplete = tell . errorMessage $ IncompleteExhaustivityCheck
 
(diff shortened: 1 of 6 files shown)


garyb:
Oh right, I forgot about the implied constraint thing, so this still needs work.

paf31:
My idea for this was to have the linter insert a function call to a built-in function `__partial :: forall a. Partial => a -> a`, wrapping the case statement. This would force the type checker to look for the `Partial` dictionary during typechecking, or error out.

Then we'd need the type error to be modified so that we don't lose the nice message.

That's why I'd prefer to wait until the constraint machinery is a bit more developed, or for something like programmable type errors, since we could stick the error message string in the constraint itself.

garyb:
Exhaustivity checking currently happens after typechecking (yeah, I know I said otherwise earlier). I'm not sure if that's actually necessary though.

I was thinking it would be good to make this a change that isn't going to immediately break everything, like the other stuff we've been doing with deprecation errors, etc - I don't see how that's possible without an approach like this.

paf31:
Perhaps we just wait and deprecate this in 0.9 and break it in 0.10? I'm not sure it's urgent, is it?

garyb:
It's not urgent, no, but it would be nice to be able to fix the warnings ahead of time in the manner that will eventually be required (which is what this was attempting to allow - I'm not saying this is the final way to implement it, like with some of the other things we're currently supporting until 0.9).

I've been thinking about the case you mentioned where `Partial` is a superclass, aside from the example you listed, it could also be applied to classes with type variables - what does it mean then? If you have something like `class Partial <= Something a where`, does that mean that all of the class members become `(Partial, Something a) => ...`?

paf31:
No, those would only get a `Something a` dictionary, and the typechecker would look up the `Partial` superclass dictionary as needed.

garyb:
Yeah, that's kinda weird. Maybe we shouldn't actually allow `Partial` to appear in superclass constraints then? That's be one way to make this approach never be incorrect :smile:

paf31:
I think it just forces the user to propagate the `Partial` constraint in all instance declarations. I can't think of any use cases immediately, but it doesn't seem all that weird to me.

garyb:
Isn't it saying "a `Partial` instance exists for these members when dealing with type `a`"? We know there is no such thing, as it's impossible to provide a `Partial` instance - `runPartial` is the only way to eliminate that constraint.

paf31:
Well, you wouldn't necessarily write instances for subclasses at the top level, just like with `Partial`. Subclasses would just represent additional assumptions about functions, like "this is partial, and also has property X"...

e.g. `class Partial <= NonTerminating` or `class Partial <= NonExhaustive`, to refine the particular flavor of partiality.

garyb:
Oh sure, it makes sense in those cases, I was saying it's weird with non-nullary classes.

paf31:
Ah right. So in that case, I'd think of the type arguments as phantom types, documenting something about the assumption. I can't think of anything off the top of my head, but maybe something like "partial due to non termination for values greater than `N`" where `N` is a type level `Int` or something.

```haskell
class Partial <= NonTerminatingFor n
```

garyb:
Ok, so I updated the predicate to check superclasses for the presence of `Partial` constraints too, although only in the case where nullary classes are involved.

Tested with this:

``` purescript
module Main where

data Test = X | Y | Z

class (Partial) <= Unsafe

class (Unsafe) <= Super

foo :: (Super) => Test -> Boolean
foo X = true
```

I know it's still imperfect, but I think this gets us most of the way there with few downsides: we can write code as if the `Partial` check is already required and have the warnings suppressed, and then when we implement the check proper it won't break anything that is warning-free, but may allow some additional cases that this isn't clever enough to deal with (the `Partial` super for non-nullary classes situation).

The `runPartial` case you mentioned in the other issue does work as we are using type information, not syntax to detect the constraint:

```
module Main where

data Test = X | Y | Z

class (Partial) <= Unsafe

class (Unsafe) <= Super

foo :: Test -> Boolean
foo t = runPartial (case t of X -> true)

runPartial :: forall a. ((Partial) => a) -> a
runPartial = Unsafe.Coerce.unsafeCoerce
```

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -258,13 +255,14 @@ checkExhaustive env mn numArgs cas = makeResult . first nub $ foldl' step ([init
...
+    do unless (hasConstraint || null bss) tellNonExhaustive
        unless (null bss') tellRedundant
        case rr of
-         Left Incomplete -> tellIncomplete
-         _ -> return ()
+         Left Incomplete -> unless hasConstraint tellIncomplete
+         _ -> when (hasConstraint && null bss) tellExhaustive
     where
-    tellExhaustive = tell . errorMessage . uncurry NotExhaustivePattern . second null . splitAt 5 $ bss
+    tellExhaustive = tell . errorMessage $ NotPartial
```
paf31: What about

```haskell
loop :: forall a b. Partial => a -> b
loop a = loop a
```

? I'd call that `Partial`. I don't think the class should be restricted to one particular type of use.

Or what if the user wants to propagate a `Partial` constraint to an otherwise total caller?

garyb: Yeah good point, drop this error/case then?

paf31:
I think this looks good, but I'd like to come up with a plan for implementing this in the typechecker in `0.8.x`. Some thoughts:

- Instead of warning directly, annotate the AST with a call to `(\a -> a) :: forall a. Partial => a -> a`.
- That will throw a type error if there is no `Partial` in scope.
- We need a way to turn that type error into a good exhaustivity error, so perhaps we need to add a new syntactic form for the typed identity function above, so that we can store the error there. After more thought, I'm not sure the programmable type errors stuff would work.

garyb:
Sounds like a good approach to me. I don't think we need to worry about `Partial` not being in scope, as we'd insert the constraint with a fully-qualified reference to `Prim.Partial` anyway.

Do you think we would need to do that to handle the error? Could we not inspect the class involved when raising the missing instance error and handle it differently there when it's `Prim.Partial`? Or don't we have enough information there to give a good error.

paf31:
The issue is that we need a way to say what binders were not matched. So we need a place to put a `[[Binder]]` inside the term.

garyb:
Ah yeah, good point. We'd have to integrate exhaustivity checking into the typechecker then I guess, or move it to before type checking (I'm not sure if there's a reason why it's currently after) and annotate the AST.

I've taken out the "this isn't actually `Partial`" error. Was there anything else in here that needed addressing, or is this okay to go now?

paf31:
No, I think it's good. But first:

- I seem to remember there being some subtle reason for exhaustivity checking happening where it does. Perhaps it would be good to investigate that before merging this, since I'd really like to have a good understanding of the path this will take in `0.8.x`.
- I'm personally very sold on using `Partial` for representing partiality in types, but I remember there being some resistance on the original thread. Obviously, we can always change our minds later, but was there any good reason given to not commit to this approach?

paf31:
> We'd have to integrate exhaustivity checking into the typechecker

I don't know if that's necessary. The simplest approach is probably to add a new type of expression:

```haskell
data Expr = 
  = ...
  | PartialMarker [[Binder]] Expr
  | ...
```

which can be used to trigger the correct error using `rethrow`. It will need some work to be done on the constraint solving side first though.

garyb:
Oh, really? I didn't remember seeing any objections. Do you remember where that was?

paf31:
I can't find it now, maybe it was on IRC. I remember something being said about orphans, and maybe a discussion about whether partiality in this sense is an effect or not. Anyway, I consider this approach to be gradual enough that people can complain before it becomes required, so let's merge :ship:

garyb:
Looking at the code again, it doesn't look like it's strictly necessary to run exhaustivity checking after type checking.

The `Environment` is used to look up data constructors, so we'll need some other way of rebuilding that (or maybe can use the `Env` from name desugaring), and there are some pattern matches on `TypedValues` that will need updating (the type is unchecked, we're just checking for a `Prelude.otherwise` or `true` literal guard).

PR #1707, opened 2015-12-12
Fix #1700, remove featureRemoved calls in parser (for 0.8)

### LINKED ISSUE #1700: Remove deprecation warnings and TODOs for 0.8

paf31:
This removes the deprecation errors for features broken in `0.7.*`.

### DIFF
--- a/src/Language/PureScript/Docs/Convert.hs
+++ b/src/Language/PureScript/Docs/Convert.hs
@@ -93,7 +93,7 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
 -- fixity declarations.
 --
 -- TODO: This may no longer be necessary after issue 806 is resolved, hopefully
--- in 0.8.
+-- in 0.9.
 addDefaultFixity :: Declaration -> Declaration
 addDefaultFixity decl@Declaration{..}
   | isOp declTitle && isNothing declFixity =
--- a/src/Language/PureScript/Parser/Common.hs
+++ b/src/Language/PureScript/Parser/Common.hs
@@ -27,11 +27,6 @@ import Language.PureScript.Names
 
 import qualified Text.Parsec as P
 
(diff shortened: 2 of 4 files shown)


garyb:
:+1:

hdgarrood:
Could we consider leaving in some form of these checks? I think it's probably right that the error should not say "this feature is from an old version of the compiler" any more, but I think it's plausible that Haskellers will use the Haskell list syntax by accident, for example.

paf31:
I agree, but can we deal with that in documentation instead? There are lots of basic syntactic features of Haskell which we don't support, but we shouldn't parse them all (tuples, nameless instances, infix function declarations etc.). If we did, we might as well just use `haskell-src-exts`.

hdgarrood:
Ok, fair enough. I don't really know how awkward it is inside the compiler, perhaps I ought to try implementing it first ;)

Personally, I would like it if the compiler to provide nicer errors for some or all of these things. Although maybe having Haskell-style special cases is the wrong approach. What about, when a parse error occurs, we could try to guess which syntactic construct the user was really after, and then give a small example of a valid version of that construct in the error? eg, "It looks like you wanted to define an instance; here's what they look like:"?

paf31:
> What about, when a parse error occurs, we could try to guess which syntactic construct the user was really after, and then give a small example of a valid version of that construct in the error?

I think that makes a lot of sense, at least for top-level constructs where things tend to be introduced by an eagerly-matched keyword (`data`, `instance`, `class`, `infix` etc.)

paf31:
This is rebased, and good to merge. I'd probably prefer to defer the parser changes to 0.9, since there are a bunch of other parser improvements which can be made in the same general category.

paf31:
I'll merge then, yes?

garyb:
Sure thing :+1:

paf31:
Ok, thanks.

PR #1708, opened 2015-12-12
Fix incorrect unused class warning

garyb:
Oops, not sure what I was thinking here.


paf31:
:+1: Makes sense to me.

PR #1710, opened 2015-12-13
Renamer updates

### LINKED ISSUE #1697: Fresh name collision from `do` desugaring
This came from purescript-contrib/purescript-profunctor-lenses#17. It's awesome.

``` purescript
module Test where

import Prelude

x :: forall m. (Monad m) => m Unit
x = do
  _ <- pure unit
  pure unit

y :: forall m. (Monad m) => m Unit
y = do
  _ <- pure unit
  pure unit

wtf :: forall m. (Monad m) => m Unit
wtf = do
  _ <- pure unit
  let tmp = _2 1
  pure unit
```

There's no `_2` in scope, but:

```
  A function of type

    Unit

    1

while applying a function _2
  of type Unit
  to argument 1
while inferring the type of _2 1
in value declaration wtf
```

Also, you can't take out either of the definitions above `wtf`, and you can't take the binds out of them either, otherwise this doesn't happen. Also there needs to be a bind above the `_2` usage in `wtf`, otherwise you get the expected:

```
Unknown value _2
```

The `_2` is a fresh name we generate:

```
var wtf = function (__dict_Monad_2) {
    return Prelude.bind(__dict_Monad_2["__superclass_Prelude.Bind_1"]())(Prelude.pure(__dict_Monad_2["__superclass_Prelude.Applicative_0"]())(Prelude.unit))(function (_2) {
        return Prelude.pure(__dict_Monad_2["__superclass_Prelude.Applicative_0"]())(Prelude.unit);
    });
};
```

garyb:
Resolves #1697

I also took the opportunity to tweak two other aspects of codegen:

- instead of generating `x$prime` for `x'` it will now generate a name like `x1` (or `x2` if `x1` is in scope, or `x3` if `x1` and `x2` are in scope, etc.)
- the generated class dictionary names are improved, so instead of `__dict_Functor_0` it will generate `dictFunctor` now, and then start numbering if there are multiple in scope

I understand if you object to the `x'` change, as it's slightly rewriting the input, but I think it's not a very difficult rule to understand, and generates much nicer names IMO.

### DIFF
--- a/src/Language/PureScript/Renamer.hs
+++ b/src/Language/PureScript/Renamer.hs
@@ -1,22 +1,9 @@
------------------------------------------------------------------------------
------------------------------------------------------------------------------
---
--- Module      :  Language.PureScript.Renamer
--- Copyright   :  (c) 2013-14 Phil Freeman, (c) 2014 Gary Burgess, and other contributors
--- License     :  MIT
---
--- Maintainer  :  Phil Freeman <paf31@cantab.net>
--- Stability   :  experimental
--- Portability :
---
--- |
--- Renaming pass that prevents shadowing of local identifiers.
---
------------------------------------------------------------------------------
-
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+-- |
+-- Renaming pass that prevents shadowing of local identifiers.
+--
 module Language.PureScript.Renamer (renameInModules) where
 
 import Prelude ()
@@ -25,6 +12,7 @@ import Prelude.Compat
 import Control.Monad.State
 
 import Data.List (find)
+import Data.Maybe (fromJust, fromMaybe)
 
 import qualified Data.Map as M
 import qualified Data.Set as S
@@ -76,19 +64,31 @@ newScope x = do
 -- unique name is generated and stored.
 --
 updateScope :: Ident -> Rename Ident
-updateScope i@(Ident name) | name == C.__unused = return i
-updateScope name = do
-  scope <- get
-  name' <- if name `S.member` rsUsedNames scope
-             then do
-               let newNames = [ Ident (runIdent name ++ "_" ++ show (i :: Int)) | i <- [1..] ]
-                   Just newName = find (`S.notMember` rsUsedNames scope) newNames
-               return newName
-             else return name
-  modify $ \s -> s { rsBoundNames = M.insert name name' (rsBoundNames s)
-                   , rsUsedNames  = S.insert name' (rsUsedNames s)
-                   }
-  return name'
+updateScope ident =
+  case ident of
+    Ident name
+      | name == C.__unused -> return ident
+      | last name == '\'' -> go ident $ Ident $ init name ++ "ʹ" -- '\x02b9'
+    GenIdent name _ -> go ident $ Ident (fromMaybe "v" name)
+    _ -> go ident ident
+  where
+  go :: Ident -> Ident -> Rename Ident
+  go keyName baseName = do
+    scope <- get
+    let usedNames = rsUsedNames scope
+        name' =
+          if baseName `S.member` usedNames
+          then getNewName usedNames baseName
+          else baseName
+    modify $ \s -> s { rsBoundNames = M.insert keyName name' (rsBoundNames s)
+                     , rsUsedNames  = S.insert name' (rsUsedNames s)
+                     }
+    return name'
+  getNewName :: S.Set Ident -> Ident -> Ident
+  getNewName usedNames name =
+    fromJust $ find
+      (`S.notMember` usedNames)
+      [ Ident (runIdent name ++ show (i :: Int)) | i <- [1..] ]
 
 -- |
 -- Finds the new name to use for an ident.
--- a/examples/passing/1697.purs
+++ b/examples/passing/1697.purs
@@ -1,0 +1,24 @@
+module Main where
+
+import Prelude
+
+_2 :: forall a. a -> a
+_2 a = a
+
+x :: forall m. (Monad m) => m Unit
+x = do
+  _ <- pure unit
+  pure unit
+
+y :: forall m. (Monad m) => m Unit
+y = do
+  _ <- pure unit
+  pure unit
+
+wtf :: forall m. (Monad m) => m Unit
+wtf = do
+  _ <- pure unit
+  let tmp = _2 1
+  pure unit
+
+main = Control.Monad.Eff.Console.log "Done"
--- a/src/Control/Monad/Supply/Class.hs
+++ b/src/Control/Monad/Supply/Class.hs
@@ -1,20 +1,8 @@
------------------------------------------------------------------------------
---
--- Module      :  Control.Monad.Supply.Class
--- Copyright   :  (c) PureScript 2015
--- License     :  MIT
---
--- Maintainer  :  Phil Freeman <paf31@cantab.net>
--- Stability   :  experimental
--- Portability :
---
+{-# LANGUAGE MultiParamTypeClasses #-}
+
 -- |
 -- A class for monads supporting a supply of fresh names
 --
------------------------------------------------------------------------------
-
-{-# LANGUAGE MultiParamTypeClasses #-}
-
 module Control.Monad.Supply.Class where
 
 import Control.Monad.Supply
@@ -22,15 +10,15 @@ import Control.Monad.State
 
 class (Monad m) => MonadSupply m where
   fresh :: m Integer
-  
+
 instance (Monad m) => MonadSupply (SupplyT m) where
   fresh = SupplyT $ do
     n <- get
     put (n + 1)
     return n
-  
+
 instance (MonadSupply m) => MonadSupply (StateT s m) where
   fresh = lift fresh
 
 freshName :: (MonadSupply m) => m String
-freshName = liftM (('_' :) . show) fresh
+freshName = liftM (('$' :) . show) fresh
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -1,23 +1,11 @@
------------------------------------------------------------------------------
---
--- Module      :  Language.PureScript.CodeGen.JS
--- Copyright   :  (c) Phil Freeman 2013
--- License     :  MIT
---
--- Maintainer  :  Phil Freeman <paf31@cantab.net>
--- Stability   :  experimental
--- Portability :
---
--- |
--- This module generates code in the simplified Javascript intermediate representation from Purescript code
---
------------------------------------------------------------------------------
-
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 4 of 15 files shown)


paf31:
I was thinking about the prime thing the other day actually. I'd be in favor of just getting rid of primes altogether actually. JS doesn't allow them, and now we have the new operator rules, there's not much reason for not using JS identifiers IMO.

paf31:
This looks great to me.

Let's just chat a bit about the prime thing before merging, since now would be the time to break it.

garyb:
Well, I'd argue very strongly against that personally. I understand the "better javascript" angle, but I'd also like PS to be a language that I actually like, and I much prefer priming to numbering variables - you can use them as semantically different things, `a1` and `a2` are different things entirely although perhaps in the same category of thing, where `a` and `a'` are representing different "versions" of a.

Also there are still lots of other things we'd have to deal with, given all the reserved and built-in names.

I forgot to mention, this prime-renaming rule only applies to local variables. If you name something `foo'` in a module it will still be exported as such.

paf31:
What about `a` -> `a_`? I consider that to be like priming, although I'd be wary of attaching too much meaning to a naming scheme anyway.

paf31:
Actually, come to think of it, why not just rename ticks to underscores? It's deterministic, and unlikely to clash. And on the off chance it does, we can give the user a nice error message.

paf31:
Or for that matter, what about `'` becoming just a `$`. It _can't_ clash, at least not if we always generate `$`s at the front of an identifier.

garyb:
I was trying to improve the readability of the JS as well, which is why eliminated the underscores (and numbers where possible) from dictionary names. Both `a___` and `a$$$` look pretty bad compared to `a3` IMO, but yeah, I think that would work too if you want to do it?

paf31:
I definitely see a lot of value in numbering identifiers and I agree underscores and dollars look bad in generated code.

I'd just like to see `a$prime` removed from generated code, ideally. I'm not sure any of my suggestions above are good solutions. Choosing to use JS identifiers is like the path of least resistance, but maybe not ideal if people really like using primes. I personally don't find myself wanting them that much.

How would you name the equivalent of `a'` in idiomatic JavaScript?

src/Language/PureScript/Renamer.hs
```diff
@@ -76,19 +64,33 @@ newScope x = do
...
-                   }
-  return name'
+updateScope ident =
+  case ident of
+    Ident name
+      | name == C.__unused -> return ident
+      | last name == '\'' -> do
+          scope <- get
+          go ident $ getNewName (rsUsedNames scope) (Ident $ init name)
+    GenIdent name _ -> go ident $ Ident (fromMaybe "v" name)
```
paf31: Any particular reason for using `v`?

Didn't you want to try the lettering scheme instead here? `a, b, c, ..., aa, ab, ...` etc.?

garyb:
I would probably be updating a `var`, so wouldn't "need" them, as I don't tend to write code SSA-style in JS. Numbering would be my second choice, hence my go-to here.

src/Language/PureScript/Renamer.hs
garyb: I would prefer that, yeah, I couldn't figure out an elegant way to generate identifers like that though. If you have one I'll definitely use that instead!

garyb:
Damn, tests failed as I removed some constraints. I thought `base-compat` was supposed to allow that?

src/Language/PureScript/Renamer.hs
paf31: Something like 

```haskell
toIdent :: Int -> ShowS
toIdent = showIntAtBase 26 (chr . (ord 'a' +))
```

paf31:
Yeah, I ran into that issue the other day too. I'm not sure, I think it might be a `transformers` thing?

paf31:
If we just use JS identifiers, we could use a [unicode prime](http://www.fileformat.info/info/unicode/char/2032/index.htm) instead.

paf31:
Sorry, [this one](http://www.fileformat.info/info/unicode/char/02b9/index.htm) works.

paf31:
Any thoughts on unicode?

garyb:
I'm slightly hesitant, as I have low expectations of what JS engines will accept, but `aʹ` is way nicer than the alternatives we've discussed so far, so I like it (also it _should_ work, as it is in the standard). It does mean we're definitely producing ES5 JS if we go with it though, so we could update a few other things too, however this is un-shimmable (although I guess it could be post-processed to remove them).

I'd suggest we rewrite `'` to `ʹ` however, and not expect our users to have to reach for a unicode reference every time they want to prime a name :wink:

tfausak:
>  I'd be in favor of just getting rid of primes altogether actually.

I'd be in favor of that too. I think that primes are too hard to notice compared to numbers or underscores. And they quickly get out of control if you have more than one.

5outh:
:+1: on removing primes from me as well; even in Haskell code nowadays I use `_` over `'` because it's easier to spot the difference with more than one.

paf31:
So this is obviously a bit of a point of contention, but the only reason I want to figure out what we're doing with it, is that we're at a good point in time to break things. I think if we can aim to use JS identifiers in 0.9 (no mangling at all), that would be a good thing. And if we break this in 0.8 now, we only have to finish off operator aliases to make that be the case.

garyb:
Not mangling is something we don't have the option of doing, unless we want to be subject to the arbitrary banning of various other words due to existing keywords and globally available values in JS. I _really_ don't want that.

jdegoes:
I must have prime. But I'm fine with unicode prime. Would be cool if ordinary prime were rewritten to unicode prime to escape it.

paf31:
> Not mangling is something we don't have the option of doing

So what I'm proposing is:

- Identifiers in code will be JS identifiers (with the possible exception of the `'` character, which can be turned into a unicode prime)
- `$` is not a valid character in a PS identifier, but everything else is fine.
- We use the `$xyz...` "namespace" for generated names
- Reserved keywords which are fine as PS identifiers simply get a `$` prepended to them.

garyb:
Okidoke, I'm totally cool with that - I think the only thing we need to make that happen is start requiring aliases, so they can disappear entirely during codegen? (and a decision on primes)

paf31:
Yes I think so. So here are some options for primes:

- Get rid completely (unpopular, but an option)
- Turn them into `_`s (ES3)
- Turn them into unicode primes (ES5)

Any others?

garyb:
Oh wait! I assume this keeps the current starts-with-lowercase-or-underscore rule for PS identifiers too?

paf31:
Ah right, yes.

garyb:
The ES5 unicode primes solution would be my preference - we've been talking about ES5ifying some of our other codegen too, like for traversing the keys of records when copying, etc.

paf31:
Let's discuss "get rid completely" though. Given that converting primes to unicode primes can only be described as a hack, then if the issue is the difficulty of typing a unicode prime, isn't the solution really to set up the editor in such a way as to make the conversion automatic? It would seem that unicode operators will make editor support necessary eventually anyway.

michaelficarra:
So I'm guessing we won't ever be allowing unicode primes in PS identifiers then to avoid the collision?

paf31:
@michaelficarra I'm interpreting this as "they'll be equivalent".

garyb:
Yeah, that's the way I'd see it too - `a'` and `aʹ` would be treated as the same identifier.

paf31:
I mean, any editor worth it's salt should be able to bind `CTRL+'` to a unicode prime.

garyb:
You could say the same about removing `\` and requiring λ` instead...

garyb:
Speaking of which, it'd be nice if that worked :smile:

paf31:
Well yes, but there's not really a good reason for that IMO. Closer compatibility with JS identifiers and cleaner codegen is pretty worthwhile I think.

tfausak:
> `a'` and `aʹ` would be treated as the same identifier

I don't like that idea at all. A few reasons:

1. If you want to find where a value is used, now you have to search for two values (`a'` and `aʹ`) instead of only one.

2. I think this sets a bad precedent. Which other characters or strings will be considered equivalent? For instance, can I write `a''` as `a″`? 

3. I typically only write ASCII in my source files. Obviously this is just a preference, but it makes everything easier because you don't really have to worry about character encodings. 

4. I hate aliases. Pick one way to do something and do it that way.

garyb:
Prime as a symbol seems common enough that that automatic conversion seems worthwhile to me.

My only other argument is things kinda freak out when you use it at the moment when messages are printed:

```
> whoopsʹ = true

No type declaration was provided for the top-level declaration of whoopspsc.exe: <stderr>: hPutChar: invalid argument (invalid character)
```

garyb:
@tfausak no, `a''` would be `aʹʹ`. The conversion would just be a 1:1 character remapping.

paf31:
I'm really on the fence. I don't use primes much, but I see their appeal. On the other hand, anything to do with automatic replacement of characters seems unpleasant.

garyb:
There are more primed identifiers used in the compiler than there were lines in the initial commit now... I'm sure they can't all be me :smile:

paf31:
Ok :smile: I'll say this - I don't use them _at the top level_ much, and for intermediate variables, there are better choices _when compiling to JS_ IMO.

Given we warn on all shadowed names, there needs to be some way of saying "like variable X, but not variable X" in `where` clauses etc., so in that sense, I see why primes are useful. I just don't think they're worth a hack like character replacement in code gen, so there must be a better approach.

garyb:
Well, there's always a `P` suffix too, but the use of that in types has already confused quite a few people.

paf31:
Oh, _that's_ what `P` stands for! :laughing:

garyb:
Haha, oh dear :smile:

That's probably only obvious if you compare the types in Haskell lens and PureScript lens. In fact, maybe nobody else at all gets it other than Hardy and I, as I think I've probably had to explain the P to anyone I've ever talked to about it.

I guess my concern with the unicode approach is: it will almost certainly be used at the top level at some point, and when it happens it will confuse people that trying to call or `foo'` or `foo’` or `foo‘` or `foo՚`... results in an unknown value. Maybe that isn't our problem, but... it seems less than ideal also. If prime wasn't common, I wouldn't be concerned. 

Also if we sort this rule out it would be really nice to enable its use in types too to avoid the `P` thing :wink:

garyb:
The fact there are many things it can be mistaken for is part of the problem too: that's a can of worms unicode opens for almost any character, but given that it's a common symbol it would perhaps be unexpected that when it appears in code it's not the most easy-to-hand variant of it.

natefaubion:
I vote :-1: on unicode prime because I think its confusing and is barely indistinguishable from a lot of other quote-like characters. I think the current approach is fine.

paf31:
So unicode prime is already possible, since any unicode letter is fine. I'd just like to decide what to do with regular primes, since they currently get codegen'd as `$prime`, which I don't think is very nice for either generated code or the FFI.

paf31:
There shouldn't be any issue with adding primes to the letters allowed in `ProperNames`, I think.

paf31:
In any case, I think we should merge this, and open a "Discussion" issue, and discuss options while we're testing the 0.8 release candidate.

garyb:
Ok, cool

PR #1711, opened 2015-12-13
Move error code pages over from wiki

paf31:
Perhaps 0.8 is a good time to handle this too?

If so, I'll change the links in the error messages as well.

### DIFF
--- a/docs/errors/Error-Code-AdditionalProperty.md
+++ b/docs/errors/Error-Code-AdditionalProperty.md
@@ -1,0 +1,1 @@
+Please consider sending a pull request to document this error.
--- a/docs/errors/Error-Code-ArgListLengthsDiffer.md
+++ b/docs/errors/Error-Code-ArgListLengthsDiffer.md
@@ -1,0 +1,1 @@
+Please consider sending a pull request to document this error.
--- a/docs/errors/Error-Code-CannotApplyFunction.md
+++ b/docs/errors/Error-Code-CannotApplyFunction.md
@@ -1,0 +1,1 @@
+Please consider sending a pull request to document this error.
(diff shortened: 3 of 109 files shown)


paf31:
I've also added the PSCi guide and Language Guide files, since they seem to be the most like actual tutorial material, as opposed to generally useful links etc.

garyb:
Maybe these should go under `/docs/`?

paf31:
The errors, or the guide? I'm not sure if errors should go there, just because they make it harder to find anything non-error related. Unless they go in `docs/errors` or something.

garyb:
Yes, sorry, that's what I meant, and yeah the guide could either be under `docs/` or `docs/guide/`

paf31:
Done

garyb:
:+1: 

Shall we update the error page links to `https://github.com/purescript/purescript/tree/v$VERSION/docs/errors/SomeErrorCode`? Is there a convenient way to propagate the current compiler version into the error codes? Is `Paths_purescript` specific to the executable or can we use it anywhere in the library?

natefaubion:
If we are versioning the docs with the compiler, it would nice to bundle them with the executable as well so you can get more info at the command line.

paf31:
That should be simple enough if we just include everything under `docs/`.

hdgarrood:
Just FYI, I just updated https://github.com/purescript/purescript/wiki/Error-Code-NoInstanceFound, I hope that's not too irritating :)

hdgarrood:
Also, I'd say we should just link to the file on github, and not include version numbers. The documentation is always improving, and we want to take advantage of that without having to wait for a new compiler release. I also think we've gotten to a stage where not very much of the error documentation needs to change often. Also, most people seem to update to the latest version of the compiler anyway.

hdgarrood:
Oh, wait, silly me, I've just realised I misunderstood @garyb's suggestion. We wouldn't have to wait for new compiler releases if we did it that way, of course. Never mind!

paf31:
I think I'll probably reopen this once we decide on the right approach.

PR #1715, opened 2015-12-14
Reduce backtracking in parser

paf31:
@garyb 

This removes some unnecessary uses of `try` in the types and kinds parsers. Hopefully this will give better parser error messages.

Now to try the really tricky cases: decls, binders and expressions :smile:


paf31:
This is ready to review. Halogen builds and the tests should pass.

paf31:
Rebased

garyb:
Nice :+1:

paf31:
Thanks. Let's test this quite a bit more before 0.8.

PR #1725, opened 2015-12-18
Ensure no duplicates appear in the PSCiState

hdgarrood:
This is a potential fix for an issue that can be reproduced as follows:

```
$ pulp init
$ pulp psci # note: this adds `:load src/Main.purs` to .psci
> :load src/Main.purs 
> 0
Error found:

  Module Main has been defined multiple times.
```

The way it works is:

- Use a Set for imported filenames, to ensure there are no duplicates
- Use a Map for loaded modules, keyed by module name, to ensure that
  there are no duplicates
- Provide new accessors which preserve the old types

Maybe if a duplicate module is loaded, silently forgetting the old module isn't the right approach, though. Especially if the filenames were different. Perhaps, instead, we should just ignore the second `:load` command if the filenames were the same, and show the error if the modules were different?

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -258,10 +258,10 @@ make st@PSCiState{..} ms = P.make actions' (map snd loadedModules ++ ms)
   allModules = map (first Right) loadedModules ++ map (Left P.RebuildAlways,) ms
 
 -- |
--- Takes a value declaration and evaluates it with the current state.
+-- Takes a value expression and evaluates it with the current state.
 --
-handleDeclaration :: P.Expr -> PSCI ()
-handleDeclaration val = do
+handleExpression :: P.Expr -> PSCI ()
+handleExpression val = do
   st <- PSCI $ lift get
   let m = createTemporaryModule True st val
   let nodeArgs = psciNodeFlags st ++ [indexFile]
@@ -530,7 +530,7 @@ getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
 -- Performs an action for each meta-command given, and also for expressions.
 --
 handleCommand :: Command -> PSCI ()
-handleCommand (Expression val) = handleDeclaration val
+handleCommand (Expression val) = handleExpression val
 handleCommand ShowHelp = PSCI $ outputStrLn helpMessage
 handleCommand (Import im) = handleImport im
 handleCommand (Decls l) = handleDecls l
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -129,7 +129,7 @@ loadAllImportedModules = do
   modulesOrFirstError <- psciIO $ loadAllModules files
   case modulesOrFirstError of
     Left errs -> printErrors errs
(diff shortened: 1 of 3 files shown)


hdgarrood:
Yeah, I rubber-ducked myself in the comment box just now. I'm going to change the Map to be keyed by filenames.

hdgarrood:
This turned into a slightly larger refactoring. See the most recent commit comments. I'm quite pleased with how it's turned out. I still need to test though.

psci/PSCi.hs
```diff
@@ -248,12 +248,14 @@ makeIO f io = do
...
-make :: PSCiState -> [(Either P.RebuildPolicy FilePath, P.Module)] -> P.Make P.Environment
-make PSCiState{..} ms = P.make actions' (map snd (psciLoadedModules ++ ms))
+make :: PSCiState -> [P.Module] -> P.Make P.Environment
+make st@PSCiState{..} ms = P.make actions' (map snd loadedModules ++ ms)
   where
-  filePathMap = M.fromList $ (first P.getModuleName . swap) `map` (psciLoadedModules ++ ms)
+  filePathMap = M.fromList $ (first P.getModuleName . swap) `map` allModules
   actions = P.buildMakeActions modulesDir filePathMap psciForeignFiles False
   actions' = actions { P.progress = const (return ()) }
+  loadedModules = psciLoadedModules st
```
paf31: Just `psciLoadedModules` with the wildcard no?

hdgarrood: Not any more, I changed the data type. The field is now `_psciLoadedModules`.

paf31:
> Perhaps, instead, we should just ignore the second :load command if the filenames were the same, and show the error if the modules were different?

Yes, I like that idea. The code looks good though, and tidies things up nicely.

We probably should rename `Main` while we're at it, to some slightly less likely name, like `PSCi.Main`, perhaps?

hdgarrood:
I think we've gotten away with the module being called `Main` because we specify in the cabal file to only look inside this directory when it's trying to build `psci`. What sort of problem are you imagining?

psci/PSCi.hs
paf31: Ah, I see now.

paf31:
No, I mean renaming the generated `.purs` module name.

hdgarrood:
Isn't that currently `$PSCI`? See `createTemporaryModuleName` in `psci/PSCi.hs`.

In the same vein, though, it might be nice for its `main` function to be mangled a bit, as currently, you can't have any declaration called `main` in scope. I remember having done `import Main as Main` in order to run `Main.main` in psci before.

paf31:
Ah I see the issue now.

And yes, we should rename `main`, I think.

hdgarrood:
Ok, cool. I'll send a separate PR in a bit to rename `main`.

paf31:
So, would you say this is ready to merge then?

hdgarrood:
Yep, I've just tested this locally and I'm pretty sure it works.

PR #1737, opened 2015-12-19
Remove special handling of primes from renamer

### LINKED ISSUE #1717: Handling of prime symbols in identifiers
Continuation from the discussion in #1710.

The behaviour that was merged renames `'` to `ʹ` (unicode prime symbol), and will append a number if an identifier with the unicode-prime-name is already in scope.

Options brought up so far:

1. Remove primes entirely, but support the unicode version.
2. Treat `'` and `ʹ` as identical, and rename `'` to `ʹ` during codegen. This is fine for ES5.
3. Replace prime with `$` or `_` in codegen, for ES3 compatibility.

Option 2 (or "option zero", the currently-merged behaviour) is my preference, as I think prime is a common enough symbol that some special handling is worthwhile.

I think Option 1 is problematic because prime is a common symbol, and there are many variants on things that look like it or apostrophes in unicode, so not taking the most obvious candidate will give a less than wonderful user experience if primes are used in top-level definitions, as without the knowledge of exactly which unicode symbol was used, it will be non-obvious which one was used.

garyb:
Resolves #1717

### DIFF
--- a/src/Language/PureScript/Renamer.hs
+++ b/src/Language/PureScript/Renamer.hs
@@ -66,9 +66,7 @@ newScope x = do
 updateScope :: Ident -> Rename Ident
 updateScope ident =
   case ident of
-    Ident name
-      | name == C.__unused -> return ident
-      | last name == '\'' -> go ident $ Ident $ init name ++ "ʹ" -- '\x02b9'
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Thanks!

PR #1738, opened 2015-12-19
Allow use of bottom integer

### LINKED ISSUE #1731: Can't use bottom integer values
```
> let x = -2147483648
Error found:
in module $PSCI
at  line 1, column 5 - line 1, column 10

  Integer value 2147483648 is out of range for the JavaScript backend.
  Acceptable values fall within the range -2147483648 to 2147483647 (inclusive).
```

Maybe make a special case for `negate 2147483648`?

garyb:
Resolves #1731

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -10,7 +10,6 @@ module Language.PureScript.CodeGen.JS
   ( module AST
   , module Common
   , moduleToJs
-  , mainCall
   ) where
 
 import Prelude ()
@@ -19,10 +18,11 @@ import Prelude.Compat
 import Data.List ((\\), delete, intersect)
 import Data.Maybe (isNothing, fromMaybe)
 import qualified Data.Map as M
-import qualified Data.Traversable as T (traverse)
+import qualified Data.Foldable as F
+import qualified Data.Traversable as T
 
 import Control.Arrow ((&&&))
-import Control.Monad (replicateM, forM)
+import Control.Monad (replicateM, forM, void)
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Reader (MonadReader, asks)
 import Control.Monad.Supply.Class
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -414,5 +415,19 @@ moduleToJs (Module coms mn imps exps foreigns decls) foreign_ =
       js <- binderToJs elVar done'' binder
       return (JSVariableIntroduction elVar (Just (JSIndexer (JSNumericLiteral (Left index)) (JSVar varName))) : js)
 
+  -- Check that all integers fall within the valid int range for JavaScript.
+  checkIntegers :: JS -> m ()
+  checkIntegers = void . everywhereOnJSTopDownM go
+    where
+    go :: JS -> m JS
+    go (JSUnary Negate (JSNumericLiteral (Left i))) = return $ JSNumericLiteral (Left (-i))
```
paf31: Is this line useful since `go` isn't recursively called and the function is `void`ed?

paf31: In fact, why not use `everythingOnJS`? Does that exist?

src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -414,5 +415,19 @@ moduleToJs (Module coms mn imps exps foreigns decls) foreign_ =
...
+    go js@(JSNumericLiteral (Left i)) =
+      let minInt = -2147483648
+          maxInt = 2147483647
+      in if i < minInt || i > maxInt
+         then throwError . errorMessage $ IntOutOfRange i "JavaScript" minInt maxInt
+         else return js
+    go other = return other
+
 mainCall :: ModuleName -> String -> JS
```
paf31: Is this unused now?

src/Language/PureScript/CodeGen/JS.hs
garyb: This is the only monadic traversal to hand for `JS`, and yeah this case is useful as otherwise the case below errors. Without moving the negation into the value we end up with `JSUnary Negate (JSNumericLiteral (Left 2147483648))`, which is larger than `maxInt`.

src/Language/PureScript/CodeGen/JS.hs
garyb: Looks like it might be, but it's exported so wasn't picked up as unused.

src/Language/PureScript/CodeGen/JS.hs
paf31: Oh I see now. It was a little confusing to me since the result is usused, except for by the traversal of the subnodes, since the traversal is top down. I would maybe add a comment here just to explain that. Other than that, PR looks good to me :+1:

garyb:
I've added an explanation to that strange traversal usage and removed `mainCall` now.

paf31:
:+1: Thanks!

PR #1746, opened 2015-12-20
Fix #1742, revert import (..) change

### LINKED ISSUE #1742: Remove `import X (..)` syntax again
After a discussion on IRC. There is an annoyance factor to this, but that can probably be resolved with tooling, and the warnings generated are only a copy-and-paste away from fixing too.

The motivation is so that all imported values must be fully enumerated. When this is done conflicts cannot arise due to change in upstream modules.

paf31:
@garyb Sorry if I'm stepping on your toes here, this one seemed simple enough.


garyb:
No problem :smile:

PR #1747, opened 2015-12-21
Remove further remnants of `import M (..)`

garyb:
Sorry, I forgot this boolean could be dropped too when looking at your PR @paf31


paf31:
I figured we'd want to distinguish between generated imports and user-typed imports?

garyb:
We don't really need to: we never generate import declarations, we just build an `Env` with the imports inserted (`Prim`, and the current module). The boolean is only read from a declaration in the linter, so the imports we insert this way are never linted.

paf31:
Ah ok. :+1: Sounds good then.

PR #1748, opened 2015-12-21
Don't warn about implicit qualified imports

### LINKED ISSUE #1721: Incorrect warnings for qualified imports
Not sure if you knew this already, but with 0.8.0-rc1:

```purescript
module Main where

import Prelude (..)
import Control.Monad.Eff.Console as Console

main :: Eff _ Unit
main = Console.log "hi"
```

produces:
```
Module Control.Monad.Eff.Console has unspecified imports, consider using the explicit form:

```

I was expecting no warning to be generated, because it's a qualified import.

Aside, it's a bit unfortunate that `psc --version` produces `0.8.0.0`, but the release page says `0.8.0-RC1`. I'm fairly sure the latter is a valid version according to Cabal anyway?

garyb:
Resolves #1721

### DIFF
--- a/src/Language/PureScript/Linter/Imports.hs
+++ b/src/Language/PureScript/Linter/Imports.hs
@@ -7,7 +7,7 @@ import Prelude ()
 import Prelude.Compat
 
 import qualified Data.Map as M
-import Data.Maybe (mapMaybe)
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Looks good, thanks!

Issue #1719, opened 2015-12-16
Type System Wishlist


paf31:
This issue is here for me to track things I want to finish in the typechecker in `0.8.x`.

- Constraint inference (#202)
- Get rid of `meet` judgment
- Constraint kinds (#1764)
- Functional dependencies (#1580)
- Remove duplicate rows restriction (#2311)
- Ability to pass around error "stack" (#1433)
- Tidy up skolem escape check (#2310)
- Restrict unification to monotypes (#2312)
- Document the type system properly (https://github.com/purescript/roadmap/issues/43)
- Row constraints (#469)
- Polymorphic labels (#823)
- Instance chains
- Kind polymorphism (#1763)

natefaubion:
Is let generalization out of the question? What about a pony?

paf31:
:smile: I can certainly look at let generalization. Let's put it under stretch goals, because I'm not sure how it plays with Rank N. Perhaps I could deal with it as part of constraint inference, since we'll want to generalize over constraints too.

jdegoes:
This looks really awesome! Constraint inference and fundeps are big wins that will greatly increase the expressive power of PureScript and clean up a lot of code.

Related, I'm not sure if this would have any impact on the type system, but something like [instance chains](https://ghc.haskell.org/trac/ghc/ticket/9334) would solve a lot of boilerplate pains in day-to-day PureScript coding.

paf31:
Good point, I've added instance chains to the stretch goals list. Chances are, most of those, if not all, will end up being bumped to some later release, but if I have time, I'll try to work on some of them. Basically, we've had a bug fix release and I'd like to spend 0.9 working only on neat features in the typechecker :smile:

garyb:
- kind polymorphism?

paf31:
I'll put that on the stretch goals list too, but I'm not really sure how to implement it yet. It's not clear to me if kind information has to live at both the term and type levels if we have polytyped values with polykinded types.

paf31:
Basically, the first list is the stuff that actually makes it inconvenient to use PureScript today (the `MonadState` issue, constraint generalization etc.)

Everything else is "nice to have" to a certain extent.

garyb:
Yeah, I meant it as a nice to have :) It would just be great for those cases like `Exists` and `Proxy` where we need to define kinded variations.

PipocaQuemada:
What do you mean by "Remove duplicate rows restriction"?

[...]

PR #1749, opened 2015-12-21
Remove meet judgment, #1719

paf31:
For 0.9.

Tested with Halogen.


paf31:
@garyb Would you mind reviewing this one? I'd like to put it into 0.8.3 if possible, since we'll have other type checker changes and we can test them all at once.

garyb:
Looks good, and the branch name almost suits my vegetarian preferences too. :wink:

Do you remember why this was done with `meet` rather than unification in the first place?

paf31:
It was a fix for a weird corner case related to subsumption vs. unification, but I think the actual issue was caused elsewhere. In any case, I think we should be using unification here. We can always use a type annotation to force checking mode in any particularly weird cases where this is an issue.

garyb:
Sounds good to me.

Issue #1724, opened 2015-12-18
psc-publish does not include operators


tfausak:
Compare this documentation: http://pursuit.purescript.org/packages/purescript-neon/0.1.1/docs/Neon.Operator

To this source: https://github.com/tfausak/purescript-neon/blob/v0.1.1/src/Neon/Operator.purs#L19-L36

paf31:
These are operator aliases, correct?

tfausak:
Yes. Sorry I forgot to specify.

hdgarrood:
This is now fixed. Check the link :)

tfausak:
:smile:

PR #1750, opened 2015-12-21
Allow operator aliases for imported values

### LINKED ISSUE #1722: Cannot define operators for imported values
With a module that defines a function like this:

``` purescript
module A where
f :: forall a b. a -> b -> a
f x _ = x
infix 5 f as @
```

You cannot define an operator version of `f` in another package. 

``` purescript
module B where
import A (f)
infix 5 f as @@
-- Warning: The import of module A is redundant
-- Error: Value f is undefined.
```

``` purescript
module C where
import A as A
infix 5 A.f as @@@
-- Error: Unable to parse module: unexpected qualifier expecting identifier or symbol
```

This is with PureScript v0.8.0-rc.1

garyb:
Resolves #1722

### DIFF
--- a/src/Language/PureScript/Docs/Convert.hs
+++ b/src/Language/PureScript/Docs/Convert.hs
@@ -71,7 +71,7 @@ type IntermediateDeclaration
 -- with their associativity and precedence.
 data DeclarationAugment
   = AugmentChild ChildDeclaration
-  | AugmentFixity P.Fixity (Maybe P.Ident)
+  | AugmentFixity P.Fixity
 
 -- | Augment top-level declarations; the second pass. See the comments under
 -- the type synonym IntermediateDeclaration for more information.
@@ -89,8 +89,8 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
     case a of
       AugmentChild child ->
         d { declChildren = declChildren d ++ [child] }
-      AugmentFixity fixity alias ->
-        d { declFixity = Just (fixity, P.runIdent <$> alias) }
+      AugmentFixity fixity ->
+        d { declFixity = Just fixity }
 
 -- | Add the default operator fixity for operators which do not have associated
 -- fixity declarations.
@@ -100,7 +100,7 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
 addDefaultFixity :: Declaration -> Declaration
 addDefaultFixity decl@Declaration{..}
   | isOp declTitle && isNothing declFixity =
-        decl { declFixity = Just (defaultFixity, Nothing) }
+        decl { declFixity = Just defaultFixity }
   | otherwise =
         decl
   where
@@ -173,8 +173,10 @@ convertDeclaration (P.TypeInstanceDeclaration _ constraints className tys _) tit
 
   childDecl = ChildDeclaration title Nothing Nothing (ChildInstance constraints classApp)
   classApp = foldl P.TypeApp (P.TypeConstructor className) tys
-convertDeclaration (P.FixityDeclaration fixity _ alias) title =
-  Just (Left ([title], AugmentFixity fixity alias))
+convertDeclaration (P.FixityDeclaration fixity _ Nothing) title =
+  Just (Left ([title], AugmentFixity fixity))
+convertDeclaration (P.FixityDeclaration fixity _ (Just alias)) title =
+  Just $ Right $ (mkDeclaration title (AliasDeclaration alias fixity)) { declFixity = Just fixity }
 convertDeclaration (P.PositionedDeclaration srcSpan com d') title =
   fmap (addComments . addSourceSpan) (convertDeclaration d' title)
   where
--- a/src/Language/PureScript/Docs/Types.hs
+++ b/src/Language/PureScript/Docs/Types.hs
@@ -84,7 +84,7 @@ data Declaration = Declaration
   , declComments   :: Maybe String
   , declSourceSpan :: Maybe P.SourceSpan
   , declChildren   :: [ChildDeclaration]
-  , declFixity     :: Maybe (P.Fixity, Maybe String)
+  , declFixity     :: Maybe P.Fixity -- TODO: remove in 0.9
   , declInfo       :: DeclarationInfo
   }
   deriving (Show, Eq, Ord)
@@ -126,6 +126,12 @@ data DeclarationInfo
   -- members are represented as child declarations.
   --
   | TypeClassDeclaration [(String, Maybe P.Kind)] [P.Constraint]
+
+  -- |
+  -- An operator alias declaration, with the member the alias is for and the
+  -- operator's fixity.
+  --
+  | AliasDeclaration (P.Qualified P.Ident) P.Fixity
   deriving (Show, Eq, Ord)
 
 declInfoToString :: DeclarationInfo -> String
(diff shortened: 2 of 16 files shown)


src/Language/PureScript/Externs.hs
```diff
@@ -73,7 +73,7 @@ data ExternsFixity = ExternsFixity
   -- | The operator symbol
   , efOperator :: String
   -- | The value the operator is an alias for
-  , efAlias :: Maybe Ident
+  , efAlias :: Maybe (Maybe ModuleName, Ident)
```
garyb: I'm not really sure why I had to do this. I got an error from the `deriveJSON` macro when trying to use `efAlias :: Maybe (Qualified Ident)`.

paf31:
:+1: Looks good, but what happens with the export check now?

garyb:
It's still there (there's a `failing` test for it), it only applies when the aliased thing is in the same module as the alias now: https://github.com/purescript/purescript/pull/1750/files#diff-38e9503380302b1c46351f6e11a234e1R437

paf31:
Ah ok. Do we verify that a qualified alias is actually in scope at the definition site, then?

garyb:
~~No, but the identifier used in the alias is substituted into...~~

I need to check something actually. It works out when the alias is declared for an unknown value in the current module and then used in the module also (you get the error from the substituted value usage, rather than the fixity declaration), but I'm not sure what happens if the module just declares the alias and doesn't use it.

hdgarrood:
Looking at #1724. If we're going ahead with this, then I think we need to modify the behaviour of `Language.PureScript.Docs.Convert` to produce a whole standalone top-level declaration whenever it encounters a new-style fixity declaration. That is, `convertDeclaration`, when given a new-style fixity declaration, needs to start returning a `Right` value.

In the same vein, what about using a separate field in `Language.PureScript.Docs.Types` for aliases? I don't think the `declFixity` field can suitably expand to cover both meanings, because they're too different. How about:

```haskell
data Declaration = Declaration
  { declTitle      :: String
  , declComments   :: Maybe String
  , declSourceSpan :: Maybe P.SourceSpan
  , declChildren   :: [ChildDeclaration]
  -- | Old-style fixity declaration. TODO: Remove in 0.9
  , declFixity     :: Maybe P.Fixity
  -- | New-style alias declaration; operator, fixity.
  , declAlias      :: Maybe (String, P.Fixity)
  , declInfo       :: DeclarationInfo
  }
  deriving (Show, Eq, Ord)
```

hdgarrood:
Oh wait, that doesn't quite work. Maybe I'll just wait for this stuff to settle down before addressing the `psc-publish` stuff.

garyb:
Sounds good. I didn't really know what I was supposed to be doing in `Language.PureScript.Docs` so it's a minor miracle it still works at all :wink:

garyb:
Oh... should I do it anyway? Or is there a better alternative?

paf31:
I think right now it just treats new style declarations as old style declarations. Maybe we can write the JSON in a compatible way, using `undefined` or no field at all in the missing case, and Pursuit will continue to be compatible with the old way.

hdgarrood:
Sorry - what I think we should do, assuming we go ahead with making it possible to define aliases for values in other modules, is:

- leave the declFixity field alone and remove it entirely in 0.9
- make new style fixity declarations into standalone declarations in the docs

If we do this, then i think (but am not 100% sure) that we need a new declAlias field which is Just for aliases, and contains a qualified ident for the value it is an alias of, and a fixity. So I think the only problem with the above is use of String.

hdgarrood:
How about: go for it now if you want / if it doesn't seem to hard. Otherwise, I'll wait for this to be merged and have a crack myself.

hdgarrood:
I guess it might be a good idea to have more people other than me familiar with that part of the code, come to think of it. Are there any confusing bits I could help explain?

garyb:
There wasn't much too confusing about the code, I just wasn't sure what we wanted to actually do with aliases, and wasn't sure how what I changed in here would effect pursuit.

garyb:
Ok, the case I was worried about is fine. I fixed some issues with when using a qualified reference but not importing the module.

I also reworked the docs stuff - it seemed to make sense to add an actual declaration for the aliases instead of having a separate category for them. I'm not too sure what we'd want the docs to look like for them though, so it just spits out the fixity statement again.

I found out why the `deriveJSON` macro was failing too, the `Qualified` instances had a `a ~ ProperName` constraint. I've replace this custom instance with a `deriveJSON`, as I need a `Qualified Ident` instance... but obviously this isn't great as now the JSON format will _definitely_ not be backwards compatible, so I'm not really sure what should be done.

Input would be appreciated @hdgarrood, or if it's easier and @paf31is fine with the other changes we could merge this and you can fix things up as necessary.

paf31:
:+1: Looks good to me, but maybe @hdgarrood should be the one to do a final review.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -126,6 +126,12 @@ data DeclarationInfo
   -- members are represented as child declarations.
   --
   | TypeClassDeclaration [(String, Maybe P.Kind)] [P.Constraint]
+
+  -- |
+  -- An operator alias declaration, with the member the alias is for and the
+  -- operator's fixity.
+  --
+  | AliasDeclaration (P.Qualified P.Ident) P.Fixity
```
hdgarrood: :+1: this is much better than my original suggestion

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -173,8 +173,8 @@ convertDeclaration (P.TypeInstanceDeclaration _ constraints className tys _) tit
 
   childDecl = ChildDeclaration title Nothing Nothing (ChildInstance constraints classApp)
   classApp = foldl P.TypeApp (P.TypeConstructor className) tys
-convertDeclaration (P.FixityDeclaration fixity _ alias) title =
-  Just (Left ([title], AugmentFixity fixity alias))
+convertDeclaration (P.FixityDeclaration fixity _ Nothing) title =
+  Just (Left ([title], AugmentFixity fixity))
```
hdgarrood: I think we need a new case here, matching `(P.FixityDeclaration fixity op (Just name))`, returning `Just (Right (...))`. At the moment, as far as I can tell, operator alias declarations won't show up in the produced docs, because they aren't being matched here.

garyb: Ah yep, that sounds about right, good spot.

garyb:
Ok, that's done now, and tested the markdown output:

---

#### `($)`

``` purescript
infixr 0 Prelude.Internal.apply as $
```

_right-associative / precedence 0_

---

The qualification is dropped if the operator is an alias for something in the current module:

---

#### `(>>>)`

``` purescript
infixr 9 compose as >>>
```

_right-associative / precedence 9_

---

Comments on aliases are still brought through too, which is nice:

---

#### `($)`

``` purescript
infixr 0 apply as $
```

_right-associative / precedence 0_

This is the bomb.

#### `(#)`

``` purescript
infixl 1 applyFlipped as #
```

_left-associative / precedence 1_

This is the bmob.

hdgarrood:
:+1: LGTM

paf31:
:+1:

PR #1752, opened 2015-12-21
allow git@github.com in extractGithub. Fixes #1549

### LINKED ISSUE #1549: Support git@...:... repository URLs
The [extractGithub](https://github.com/purescript/purescript/blob/3794bda/src/Language/PureScript/Publish.hs#L195) function requires URLs to start with `git://github.com/`. The clone URLs provided by GitHub start with `git@github.com:`, so they are rejected.

codedmart:
Didn't see anywhere that the url is tested. Please let me know if I missed anything. Just starting to look around the codebase.

### DIFF
--- a/src/Language/PureScript/Publish.hs
+++ b/src/Language/PureScript/Publish.hs
@@ -41,6 +41,7 @@ import qualified Data.Text.Lazy.Encoding as TL
 
 import Control.Category ((>>>))
 import Control.Arrow ((***))
+import Control.Applicative ((<|>))
 import Control.Exception (catch, try)
 import Control.Monad.Trans.Maybe (MaybeT(..), runMaybeT)
 import Control.Monad.Trans.Except
@@ -197,7 +198,7 @@ getBowerInfo = either (userError . BadRepositoryField) return . tryExtract
 
 extractGithub :: String -> Maybe (D.GithubUser, D.GithubRepo)
 extractGithub =
-  stripPrefix "git://github.com/"
+  matchUrl
    >>> fmap (splitOn "/")
    >=> takeTwo
    >>> fmap (D.GithubUser *** (D.GithubRepo . dropDotGit))
@@ -211,6 +212,10 @@ extractGithub =
   dropDotGit str
     | ".git" `isSuffixOf` str = take (length str - 4) str
     | otherwise = str
+
+  matchUrl :: String -> Maybe String
+  matchUrl str =
+    stripPrefix "git@github.com:" str <|> stripPrefix "git://github.com/" str
 
 readProcess' :: String -> [String] -> String -> PrepareM String
 readProcess' prog args stdin = do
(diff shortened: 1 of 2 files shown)


phadej:
I'd almost use `regex-applicative` or `parsec` here. The ad-hoc parsing is getting out-of-hands.

paf31:
Looks good, but I agree with @phadej that a parsing library might be a better fit now.

Could you please update `CONTRIBUTORS.md` too?

codedmart:
@paf31 OK updated `CONTRIBUTORS.md` should I make the change to a parsing library now? Or should we leave this and add another issue for that. Are there other areas we could use the parser throughout the codebase? Thanks!

src/Language/PureScript/Publish.hs
```diff
@@ -212,6 +212,11 @@ extractGithub =
     | ".git" `isSuffixOf` str = take (length str - 4) str
     | otherwise = str
 
+  matchUrl :: String -> Maybe String
+  matchUrl str
+    | "git@github.com:" `isPrefixOf` str = stripPrefix "git@github.com:" str
+    | otherwise = stripPrefix "git://github.com/" str
+
```
hdgarrood: Would `stripPrefix "git@github.com:" str <|> stripPrefix "git://github.com/" str` work here?

hdgarrood: Although I guess it doesn't matter if you're about to switch to parsec or something anyway. I am fairly sure this is the only part of the codebase that deals with these kinds of URLs, too.

paf31:
I'll leave this up to @hdgarrood to decide if it's worth switching to a parser library or not.

hdgarrood:
I'm generally :-1: for regexes in real code (as opposed to, say, a shell pipeline command that you might run as a one-off). Yes, this is a little ad-hoc, but so are regexes, and this kind of code is much easier to read or modify.

`parsec` seems reasonable, but I'm not sure it's worth it here, especially since the chance of what we need to parse here becoming any more complex seems rather slim.

paf31:
I think `regex-applicative` makes things a little less ad-hoc, but you still get the speed benefits. Not that speed is a particular concern here.

codedmart:
@hdgarrood What would you like me to do? Fix this up with something like `regex-applicative` or leave it as is or make the other change you suggested `stripPrefix "git@github.com:" str <|> stripPrefix "git://github.com/" str`?

hdgarrood:
Now that I've had a proper look at regex-applicative it doesn't look like too bad an option. When I said :-1: I was imagining a single regex that would extract the GitHub username and password through capturing groups.

Anyway I still think sticking with the current approach is ok for now, but I could be persuaded otherwise. If we do stick with the current approach then i do think <|> would be better. Using isPrefixOf as well as stripPrefix in the same line is a code smell, I think.

codedmart:
Fair enough I will make some changes. Thanks!

paf31:
I'm fine with keeping the current approach, but I agree that `<|>` would make the code easier to read.

codedmart:
Sorry guys took a break for a few days. I will get these changes done this Monday. Thanks!

paf31:
No rush :smile:

codedmart:
Changes made. Let me know if there is anything else.

paf31:
Looks good to me, thanks! Could you please squash your commits?

codedmart:
Done

paf31:
@hdgarrood Is this ok now?

hdgarrood:
Looks good to me :+1:

paf31:
Thanks!

tfausak:
:+1: Thanks!

PR #1759, opened 2015-12-26
Fix warning about values missing from virtual modules

### LINKED ISSUE #1733: Incorrect error message for unexported values
This code:
```purescript
module Main where

import Thingy as Thing

main = Thing.doesntExist "hi"

module Thingy where

foo :: Int
foo = 1
```
gives this error:
```
Error found:
in module Main
at /home/harry/documents/code/ps-scratch/src/Main.purs line 10, column 8 - line 12, column 1

  Unknown module Thing
```

I was expecting "Unknown value Thing.doesntExist" or something along those lines.

However, the right error is produced when the module is imported qualified using the same virtual name as its real name.

garyb:
Resolves #1733

### DIFF
--- a/examples/failing/1733.purs
+++ b/examples/failing/1733.purs
@@ -1,0 +1,13 @@
+-- @shouldFailWith UnknownValue
+
+module Main where
+
+import Thingy as Thing
+
+main = Thing.doesntExist "hi"
+
+module Thingy where
+
+foo :: Int
+foo = 1
+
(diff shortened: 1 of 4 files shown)


garyb:
I might come back to this later, as I think there probably is still a less confusing representation of the import names `Env` that will help clear the code up generally (perhaps), where the env is indexed by module rather than having `Qualified x` names for the keys of the map. This new field could be dropped again then too.

paf31:
:+1: Looks good to me.

Issue #1677, opened 2015-11-27
Improve documentation for re-exports


hdgarrood:
See https://github.com/tfausak/purescript-batteries, for instance.

One potential design: suppose we have code like this:

```purescript
module Main (foo, bar, module A) where
import Other as A
foo = 1
bar = 2

module Other where
baz = 3
```

Then we could generate docs like this:

```
## Module Main
#### `foo`
foo :: Int
#### `bar`
bar :: Int
### Re-exported from Other
#### `baz`
baz :: Int
```

That is, put a separate section at the end for re-exported values, noting which module each came from.

This design doesn't work brilliantly if you want to do something like having a `Foo.Internal` module in addition to a `Foo` module, where `Foo` re-exports selectively from `Foo.Internal`, though. Perhaps we could copy haddock's mechanism of comment directives inside module export lists to give more control if people want it? Also, perhaps we want some mechanism to tell Pursuit that a module is internal and should not be displayed in the documentation?

Anyway I think this is probably good enough to start with and see how well it works. It's certainly better than the current behaviour, I think.

tfausak:
:+1: 

The current situation for re-exports leaves a lot to be desired. [The generated documentation for Batteries](http://pursuit.purescript.org/packages/purescript-batteries/0.3.0/docs/Batteries) was so useless I had to add my own hard-coded documentation.

hdgarrood:
@garyb in #1415:
> Although actually, with the changes I'm about to open a PR for it makes partial re-exporting possible too, so maybe we need to be more specific than just saying which modules are.

Can you explain what the current situation is? I thought partial re-exporting was only possible by importing part of a module into a virtual module, and then re-exporting that, like this:

```
module Foo (module Virtual) where

import Prelude (Unit, unit) as Virtual
```

Is that correct?

garyb:
Not quite, if you re-export a module after selectively importing from it only the imported values are re-exported, so:

``` purescript
module Foo (module Prelude) where

import Prelude (Unit, unit)
```

would export `(Unit, unit)`

hdgarrood:
Ah, cool, thanks. Is there a function inside the library somewhere which will allow me to pass in a module, and get all the re-exported values from it and which module they came from? Something like `P.Module -> [(P.ModuleName, [P.Declaration])]`?

garyb:
[...]

PR #1760, opened 2015-12-27
Remove ParseDesugarError data type

hdgarrood:
There's no reason to distinguish the individual cases, I think, so this
data type just adds noise for no benefit.

I came across this because I am currently looking into #1677, and getting this in first would simplify that, too, particularly in `psc-docs`.


paf31:
:+1: Looks good!

PR #1761, opened 2015-12-27
uses the latest resolver when stack installing

kRITZCREEK:
This is so that users get an "up to date" version of purescript when their global stack config is set to an old resolver.


paf31:
:+1: Thanks!

PR #1765, opened 2015-12-27
WIP, kind polymorphism

### LINKED ISSUE #1763: Kind polymorphism

paf31:
This will need some more testing, but should be ready enough to merge by the time 0.8 is closed.

Resolves #1763.

### DIFF
--- a/examples/passing/PolyKinds.purs
+++ b/examples/passing/PolyKinds.purs
@@ -1,0 +1,59 @@
+module Main where
+
+import Prelude (Unit)
+
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (CONSOLE, log)
+
+data Proxy a = Proxy
+
+test1 :: Proxy Array
+test1 = Proxy
+
+test2 :: Proxy Object
+test2 = Proxy
+
+test3 :: Proxy Int
+test3 = Proxy
+
+liftProxy :: forall a. Proxy a -> Proxy (Proxy a)
+liftProxy _ = Proxy
+
+unsafeCoerce :: forall a b. a -> b
+unsafeCoerce a = unsafeCoerce a
+
+data Exists (f :: k -> *)
+
+runExists :: forall f r. (forall a. f a -> r) -> Exists f -> r
+runExists = unsafeCoerce
+
+mkExists :: forall f a. f a -> Exists f
+mkExists = unsafeCoerce
+
+existsObject :: forall a. a -> Exists Object
+existsObject _ = mkExists { polykinded: true }
+
+existsArray :: forall a. a -> Exists Array
+existsArray _ = mkExists [0]
+
+test4 :: forall a. a -> Int
+test4 a = runExists (\_ -> 0) (existsArray a)
+
+-- | Mutually recursive polykinded types
+data Foo a = Foo (Bar a)
+
+data Bar a = Bar (Foo a)
+
+test5 :: forall a b. b -> Foo a
+test5 b = Foo bar
+  where
+  bar :: Bar a
+  bar = Bar (test5 b)
+
+type OnEmpty f = f ()
+
+test6 :: OnEmpty Object
+test6 = {}
+
+main :: forall eff. Eff (console :: CONSOLE | eff) Unit
+main = log "Done"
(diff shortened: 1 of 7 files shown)


paf31:
Here is one bug:

```text
$ psci
> foreign import data F :: (# k -> k1) -> k1
> :kind F Object
k0
```

The result of the query should be `*`.

paf31:
Ok, that bug is fixed, but here's one more:

```purescript
foreign import data Any :: k

foreign import any :: Any
```

gives

```text
In a type-annotated expression x :: t, the type t must have kind *.
The error arises from the type
  Any
having the kind
  k14
instead.
```

since we check kind _equality_ with `Star` in type annotations, which is too strict, we just need unification.

paf31:
Ok, all of the above are now fixed. This is ready for review.

hdgarrood:
Looks good to me, although I'm not too familiar with the type checker.

garyb:
Hell yeah! Code looks great, it's not really any more complicated than what was there before too, which is nice.

I kinda dread to ask this, but do you think kind variables should be explicitly quantified also? They don't have the same scope or rank issues as type variables of course, but I just thought I'd mention it, given the recent discussions about `forall` being useful for comprehension too sometimes.

paf31:
> do you think kind variables should be explicitly quantified also?

I thought about this quite a bit, and realized something else: explicit quantification makes the meaning of `TypeVar`/`KindVar` a little clearer, since it makes it really obvious where instantiation needs to happen. I didn't implement it for now since we're not going to get "rank N kinds", although thinking about it, we might reasonably still have scoped kind variables. I'm open to the idea, certainly.

jdegoes:
> I kinda dread to ask this, but do you think kind variables should be explicitly quantified also?

:+1:

paf31:
Thinking about this some more, there isn't any case where scoped type variables would make sense, so the only reason to add `forall` in kinds would be to aid code comprehension (or to make things symmetrical with types, but I think that might give the wrong impression that `forall` is valid anywhere, like in types).

If we do decide to add `forall`, I suggest we keep the inference the same as it is now, and implement scope checking in the  parser. The only place where `forall` should be allowed is in `foreign data` declarations, since it doesn't make sense to add a `forall` to something like `data Proxy (a :: k) = Proxy`.

paf31:
Hmm, it seems we do need to track more kind information at the value level after all. If we instantiate a value with a type in which a kind variable has itself been instantiated, we don't later verify that more specific kind when the value is used.

paf31:
I should say: it's not clear to me where that kind information should go - whether we should pass the kind arguments as explicit arguments to type constructors and type formers:

```purescript
data Proxy k (a :: k) = Proxy

-- Not clear where to bind `k` here though
foreign import data Proxy :: k -> *

-- Applications of the type constructor mention the kind explicitly
-- The forall syntax here seems a bit dodgy
useProxy :: forall k a. Proxy k a -> Proxy (# k) a
```

or try to infer things and store the kinds inside the types somehow.

jdegoes:
```purescript
foreign import data Proxy k :: k -> *
```

paf31:
What about this syntax?

```purescript
-- Kind arguments in data use a keyword
data Proxy (kind k) (a :: k) = Proxy

-- No need for a keyword here, since we're declaring a kind anyway.
-- But we do make the restriction that all foralls must go on the outside.
foreign import data Proxy :: forall k. k -> *

-- Kinds are qualified in `forall` but with a special keyword
useProxy :: forall (kind k) a. Proxy k a -> Proxy (# k) a
```

garyb:
I think passing/defining the kinds explicitly is a little strange actually. I'm also not sure I understand what something like the `useProxy` example means, as surely there'd be no `a` that is both `k` and `# k`?

jdegoes:
First off, I don't really like any `data` syntax, because it overloads the meaning of `=`, which usually means definition (= substitution semantics). I prefer GADT syntax:

```haskell
data Maybe a where
  Nothing :: Maybe a
  Just    :: a -> Maybe a
```

This might, possibly, suggest an alternative syntax for introducing kind arguments.

Second, it is a bit strange to pass/define kinds explicitly. I have some thoughts on why this is strange but it will take some time to write them up.

Issue #1681, opened 2015-11-29
psc-docs is using Show instance for errors


hdgarrood:
For example, on current slamdata master (I guess they're using an older compiler?)

```
$ pulp docs
* Generating documentation in /home/harry/documents/code/slamdata
psc-docs: error: unknown module "ModuleName [ProperName {runProperName = "View"},ProperName {runProperName = "File"},ProperName {runProperName = "Modal"},ProperName {runProperName = "Mount"}]"
* ERROR: Subcommand terminated with error code 1
```

garyb:
> I guess they're using an older compiler?

Yeah, master is stuck on psc 0.7.2. Main development is in the `new-halogen` branch currently.

paf31:
So is this a bug in master?

hdgarrood:
Sorry, yes, my point was that it would probably be better for `psc-docs` to not use the `Show ModuleName` instance when reporting errors.

hdgarrood:
I can't reproduce this now. Maybe it's been fixed, somehow.

I think the problem must have been somewhere inside `Language.PureScript` (i.e. the library) rather than inside the code specific `psc-docs`, given that the "unknown module" part is displayed in a human-friendly way, it's just the actual module name that's using `Show`.

If I try to run `pulp docs` without first running `bower install`, in order to purposefully trigger an unknown module error, I do now get a properly formatted error:

```
Error found:
at /home/harry/documents/code/slamdata/src/View/Modal/Common.purs line 21, column 1 - line 23, column 1

  Unknown module Control.Plus


See https://github.com/purescript/purescript/wiki/Error-Code-UnknownModule for more information,
or to contribute content related to this error.
```

So I'm going to close this.

hdgarrood:
Silly me, the error occurs when you pass something to --docgen that doesn't exist.

Better repro steps:

```
$ pulp init
$ psc-docs 'bower_components/*/src/**/*.purs' 'src/**/*.purs' --docgen LOL
psc-docs: error: unknown module "ModuleName [ProperName {runProperName = "LOL"}]"
```

The culprit is here: https://github.com/purescript/purescript/blob/896bbcdb91c43b3054297e8c3a28592f00de07d3/psc-docs/Main.hs#L99-L107

PR #1770, opened 2015-12-29
Reexport docs

hdgarrood:
Addresses #1677.

Changes:
* `Env` is now exported by `Language.PureScript`
* Added a variant of `desugarImports` which also returns the `Env`.
* Generalise `parseAndDesugar` to any `MonadIO m, MonadError P.MultipleErrors m`
* Make `parseAndDesugar` also return the `Env`.
* Remove the useless callback parameter to `parseAndDesugar` (what was I thinking??)
* Add a re-exports field to the `Module` type in `Language.PureScript.Docs`
* Print re-exported values in `psc-docs`.

For now, I just want to see if this passes CI (for some reason, tests aren't working on my machine). Before merging, this still needs:
* [x] History cleanup
* [x] Tests

### DIFF
--- a/src/Language/PureScript/Docs/Convert.hs
+++ b/src/Language/PureScript/Docs/Convert.hs
@@ -2,230 +2,86 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE FlexibleContexts #-}
 
 -- | Functions for converting PureScript ASTs into values of the data types
 -- from Language.PureScript.Docs.
 
 module Language.PureScript.Docs.Convert
-  ( convertModule
+  ( convertModules
+  , convertModulesInPackage
   , collectBookmarks
   ) where
 
 import Prelude ()
 import Prelude.Compat
 
-import Control.Monad
+import Control.Monad.Error.Class (MonadError)
+import Control.Arrow ((&&&))
 import Control.Category ((>>>))
-import Data.Either
-import Data.Maybe (mapMaybe, isNothing)
-import Data.List (nub, isPrefixOf, isSuffixOf)
+import qualified Data.Map as Map
 
 import qualified Language.PureScript as P
 
 import Language.PureScript.Docs.Types
+import Language.PureScript.Docs.Convert.Single (convertSingleModule, collectBookmarks)
+import Language.PureScript.Docs.Convert.ReExports (updateReExports)
 
 -- |
--- Convert a single Module.
+-- Like convertModules, except that it takes a list of modules, together with
+-- their dependency status, and discards dependency modules in the resulting
+-- documentation.
 --
-convertModule :: P.Module -> Module
-convertModule m@(P.Module _ coms moduleName  _ _) =
-  Module (P.runModuleName moduleName) comments (declarations m)
+convertModulesInPackage ::
+  (Functor m, MonadError P.MultipleErrors m) =>
+  P.Env ->
+  [InPackage P.Module] ->
+  m [Module]
+convertModulesInPackage env modules =
+  go modules
   where
-  comments = convertComments coms
-  declarations =
-    P.exportedDeclarations
-    >>> mapMaybe (\d -> getDeclarationTitle d >>= convertDeclaration d)
-    >>> augmentDeclarations
-    >>> map addDefaultFixity
+  localNames =
+    map (P.runModuleName . P.getModuleName) (takeLocals modules)
+  go =
+    map ignorePackage
+     >>> convertModules env
+     >>> fmap (filter ((`elem` localNames) . modName))
 
--- | The data type for an intermediate stage which we go through during
--- converting.
+-- |
+-- Convert a group of modules to the intermediate format, designed for
+-- producing documentation from. It is also necessary to pass an Env containing
+-- imports/exports information about the list of modules, which is needed for
+-- documenting re-exports.
 --
--- In the first pass, we take all top level declarations in the module, and
--- collect other information which will later be used to augment the top level
--- declarations. These two situation correspond to the Right and Left
--- constructors, respectively.
+-- Preconditions:
 --
--- In the second pass, we go over all of the Left values and augment the
--- relevant declarations, leaving only the augmented Right values.
+--     * If any module in the list re-exports documentation from other
+--     modules, those modules must also be included in the list.
+--     * The modules passed must have had names desugared and re-exports
+--     elaborated first.
 --
--- Note that in the Left case, we provide a [String] as well as augment
--- information. The [String] value should be a list of titles of declarations
--- that the augmentation should apply to. For example, for a type instance
--- declaration, that would be any types or type classes mentioned in the
--- instance. For a fixity declaration, it would be just the relevant operator's
--- name.
-type IntermediateDeclaration
-  = Either ([String], DeclarationAugment) Declaration
-
--- | Some data which will be used to augment a Declaration in the
--- output.
+-- If either of these are not satisfied, an internal error will be thrown. To
+-- avoid this, it is recommended to use
+-- Language.PureScript.Docs.ParseAndDesugar to construct the inputs to this
+-- function.
 --
--- The AugmentChild constructor allows us to move all children under their
--- respective parents. It is only necessary for type instance declarations,
--- since they appear at the top level in the AST, and since they might need to
--- appear as children in two places (for example, if a data type defined in a
--- module is an instance of a type class also defined in that module).
+convertModules ::
+  (Functor m, MonadError P.MultipleErrors m) =>
+  P.Env ->
+  [P.Module] ->
+  m [Module]
+convertModules env =
+  P.sortModules >>> fmap (convertSorted env . fst)
+
+-- |
+-- Convert a sorted list of modules.
 --
--- The AugmentFixity constructor allows us to augment operator definitions
--- with their associativity and precedence.
-data DeclarationAugment
-  = AugmentChild ChildDeclaration
-  | AugmentFixity P.Fixity
-
--- | Augment top-level declarations; the second pass. See the comments under
--- the type synonym IntermediateDeclaration for more information.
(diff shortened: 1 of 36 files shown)


hdgarrood:
This is what `Neon` from `purescript-neon` looks like with these changes: https://gist.github.com/hdgarrood/b55b7dd7e9e1ca43b7b2

/cc @tfausak

hdgarrood:
Oh also, fixes #1681

tfausak:
:+1: Looks great!

hdgarrood:
Ok, this seems to work. I think it's ready for review. I cleaned the history up so that you can probably review them each individually.

hdgarrood:
Actually, this is not quite there yet.

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -38,6 +102,95 @@ convertModule m@(P.Module _ coms moduleName  _ _) =
...
+expandReExports ::
+  (Functor m, Applicative m,
+  MonadState (Map P.ModuleName (P.Module, Module)) m) =>
+  P.Env ->
+  P.Module ->
+  Module ->
+  m Module
+expandReExports env ast@(P.Module _ _ mn' _ mexports) mdl = do
+  let exports = fromMaybe (err "exports were not elaborated") mexports
+  let reExportNames = mapMaybe takeReExport exports
```
hdgarrood: One problem with this code is that it can't handle re-exported virtual modules (see also: `examples/docs/Virtual.purs`: https://github.com/hdgarrood/purescript/blob/reexport-docs/examples/docs/src/Virtual.purs, which, with this code, would cause an internal compiler error).

I want to get a list of all the (real) modules from which values are being re-exported, but I don't want the modules they were originally exported from; I want the modules they were imported from the current module's point of view, that is, I don't want it to follow the whole path of re-exports of re-exports.

This seems to work but of course it chokes on virtual modules. I'm not particularly familiar with this part of the compiler, so I'd appreciate some guidance.

garyb: The information you're after isn't really accessible, as the renaming never takes the intermediate step, it always has multiply-re-exported values pointing at the original export locations.

What's actually the problem with that? You'll have to manually follow the chain of the links through the docs back to the original module if you want to get the definition's comments anyway, so this just takes those steps out of it...

Instead of grouping the re-exported values under a heading for each module, why don't we just lump them all under "Re-exported definitions" in general, and then show the original module info per re-exported member instead? That'd cut down on the number of headers at least, if that's one of the concerns.

hdgarrood: I'm not worried about the number of headings; I arrived at this design because I think it's the best way of presenting this information.

Suppose we have a ByteString type defined in Data.ByteString.Internal, re-exported by Data.ByteString, and then re-exported again by some module A. In that case, I think we would want to write in A's docs that ByteString is re-exported from Data.ByteString. More generally, I think the documentation should present things the same way as they are in the source, because I think that's more likely to make sense to readers of the docs.

The reason for the call to sortModules in this code is to make sure dependencies have already been rendered; this way, we can include the whole declaration for re-exports, including comments, so you wouldn't have to follow the chain back to the source.

garyb: Okidoke. I'm not sure what to suggest for now; the information you need is never constructed. I guess we'll have to change the way the imports/exports are handled to accumulate it somehow, either by reworking the data structures generally or maybe we can tack something in there that allows us to capture the info just for docs.

hdgarrood: Ah ok, maybe this is going to be a little harder then. Save this for after 0.8 perhaps?

garyb: Perhaps, yeah. I'll have a quick look at the "tack on extra info" approach before giving up totally though, as I did that recently to fix the bug where it was warning about some module not existing when it's the value that was actually missing. I added a set of virtual modules to the `Env`, so perhaps we can change that into a `Map` and track the imported values there instead or something.

garyb: [This](https://github.com/purescript/purescript/blob/bc5f300eed5fa3dfa786ff911262731ec22dd78b/src/Language/PureScript/Sugar/Names/Env.hs#L65) is what I'm referring to if you want to take a look yourself?

hdgarrood: Ok, that does sound promising - I'll take a look :)

hdgarrood:
@garyb Your suggestion worked wonderfully! I fixed all the failing tests and cleaned up the history again. I think this is good to go now.

garyb:
Nice :smile:

Although I'm not too sure about the fixed part, seems the build is failing still.

hdgarrood:
Looks like it's the `psc-publish` tests which are failing. They pass on my machine, so I think they can be disabled for now. They aren't running in master at all at the moment anyway, because of a call to `exitSuccess` before we even reach those tests: https://github.com/purescript/purescript/blob/eb144ec09de12192c573befbf7ecfd1484ff1476/tests/Main.hs#L204

I can look at them again after this is merged.

hdgarrood:
Ah, no, the docs tests are actually faililng. `cabal test` seems to be swallowing stderr, too >:( https://travis-ci.org/purescript/purescript/jobs/99953632#L3171

hdgarrood:
Oh, no, it's just printing stdout and stderr out of order: https://travis-ci.org/purescript/purescript/jobs/99953632#L3090-L3097

How do we feel about running `dist/build/tests/tests` and `dist/build/psci-tests/psci-tests` explicitly, instead of `cabal test`?

hdgarrood:
Fixed now, I think. On second thoughts, I'd rather just leave the CI scripts alone.

garyb:
Looks like it's still failing unfortunately, for the build with disttests enabled. The log is so massive I'm having a hard time seeing what's going wrong though... any ideas?

hdgarrood:
I think I just worked it out - the new examples, used in the new tests, weren't listed in `extra-source-files` in purescript.cabal, meaning the tests couldn't run out of an sdist. I just amended the commit and force pushed so hopefully it will work now.

src/Language/PureScript/Sugar/Names/Env.hs
```diff
@@ -60,16 +60,16 @@ data Imports = Imports
   , importedModules :: S.Set ModuleName
   -- |
   -- The names of "virtual" modules that come into existence when "import as"
-  -- is used.
+  -- is used, mapped to the real module being imported.
   --
-  , importedVirtualModules :: S.Set ModuleName
+  , importedVirtualModules :: M.Map ModuleName ModuleName
```
garyb: I'm not sure this is quite right actually, shouldn't it be `M.Map ModuleName [ModuleName]`? As you can have things like:

``` purescript
import A as X
import B as X
```

That makes the problem of figuring out where re-exports came from a little awkward again though, so maybe we actually need even more information in the RHS value (something like `Imports` again with qualified `X.foo` names pointing to the imported `A.foo`/`B.foo` names)

hdgarrood: Oh, yes, of course. Although maybe a `[ModuleName]` would be sufficient? I think we can look in the other fields of the `Imports` value for that information?

hdgarrood: Actually, all this code in `Docs.Convert` is worryingly similar to what's already in `Sugar.Names`, I think. How would you feel about instead adding a little extra information to the `Exports` data type, mentioning the first "real" module that declarations come from? Then, in `Docs.Convert`, we could just walk over the `Exports` value. We then wouldn't have to look at any `DeclarationRef` values, or `Imports` values, or anything.

garyb: That's totally fine by me, it shouldn't complicate the names stuff much, right?

We're already on the second revision of the names stuff and I don't think it's any clearer what is going on than my first attempt at it (although it works much better this time) - I'd like to try yet again to make it a little easier to comprehend and extract this kind of info out of, but that's not something I want to get into right now :wink:

hdgarrood:
Half a note to myself, half for the benefit of anyone reading this later:

I just had a closer look and I'm pretty sure all the information I need is already in the Env. The Imports value has the module things were imported from on the RHS; in the case of re-exports, it lists the module things were re-exported from. So I can iterate over the Exports, looking in the Imports for each exported thing to find which module I should display it as having come from. I'll also update the comments in Sugar.Names.Env to reflect this. :D

hdgarrood:
Ok now that I've looked a bit more, we don't quite have enough information as it is now. Take this example:

```purescript
module Docs (module Clash1) where

import Clash1 as Clash1
import Clash2 as Clash2
```

And suppose both `Clash1` and `Clash2` re-export values called `x` from two different other modules.

Then, the `Exports` value for `Docs` tells me that a value called `x` is exported, and that `x` is originally defined in one of those two other modules, but the `Imports` values only lists them as having come from `Clash1` or `Clash2`. So I can't work out which one is exported without also looking at the exports list. I'd rather add this information to the `Env`, I think, as the code in `Sugar.Names.Env` is probably best equipped to handle collecting this information.

hdgarrood:
Wait, we do! I was wrong, the `Imports` values lists both the module things are originally defined in as well as the module they were imported from (via the `Qualified`)! Just give me another day or two.

hdgarrood:
Ok, I think this is good to go now.

garyb:
I'll review this sometime this evening. It'll be great to have it solved at last!

garyb:
Looks good to me :+1:

It's crazy how complicated selective re-exports make things unfortunately (I spent quite a few hours working stuff out related to them for the warnings today), but they are really rather useful.

garyb:
@paf31 shall I merge this or do you want to review too / have any comments?

paf31:
@garyb If you're happy with everything, go ahead and merge. Thanks @hdgarrood!

garyb:
Yeah, thanks @hdgarrood!

hdgarrood:
:fireworks: you're welcome. :)

PR #1771, opened 2015-12-29
Ensure tests don't exit early

hdgarrood:
The test suite was calling exitSuccess after the compiler tests were
complete, which meant that the psc-publish tests were never running.
Oops.


hdgarrood:
Actually, I think it's easier to address this inside #1770.

PR #1773, opened 2015-12-30
Distinguish between the different ProperNames

garyb:
Figured we might want to get this is sooner rather than later with TC changes in 0.9... although it's already a bit late, it'll cause conflicts with the already open PRs I imagine.

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -1,24 +1,12 @@
------------------------------------------------------------------------------
---
--- Module      :  Language.PureScript.Exhaustive
--- Copyright   :  (c) 2013-15 Phil Freeman, (c) 2014-15 Gary Burgess
--- License     :  MIT (http://opensource.org/licenses/MIT)
---
--- Maintainer  :  Phil Freeman <paf31@cantab.net>
--- Stability   :  experimental
--- Portability :
---
--- |
--- | Module for exhaustivity checking over pattern matching definitions
--- | The algorithm analyses the clauses of a definition one by one from top
--- | to bottom, where in each step it has the cases already missing (uncovered),
--- | and it generates the new set of missing cases.
---
------------------------------------------------------------------------------
-
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
+-- |
+-- Module for exhaustivity checking over pattern matching definitions
+-- The algorithm analyses the clauses of a definition one by one from top
+-- to bottom, where in each step it has the cases already missing (uncovered),
+-- and it generates the new set of missing cases.
+--
 module Language.PureScript.Linter.Exhaustive (checkExhaustiveModule) where
 
 import Prelude ()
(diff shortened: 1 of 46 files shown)


garyb:
I chose to go with phantom types over newtypes as it meant less changes to the code. If we'd rather newtype instead that's fine too, I can do that instead.

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -54,7 +42,11 @@ data RedudancyError = Incomplete | Unknown
 -- |
 -- Qualifies a propername from a given qualified propername and a default module name
 --
-qualifyName :: a -> ModuleName -> Qualified a -> Qualified a
+qualifyName
+  :: (ProperName a)
+  -> ModuleName
+  -> Qualified (ProperName b)
```
paf31: should these tags be different?

garyb: Apparently, yeah - sometimes a constructor name is used to qualify a type name or vice versa.

src/Language/PureScript/Names.hs
```diff
@@ -51,12 +51,21 @@ freshIdent' = liftM (GenIdent Nothing) fresh
 -- |
 -- Proper names, i.e. capitalized names for e.g. module names, type//data constructors.
 --
-newtype ProperName = ProperName { runProperName :: String } deriving (Show, Read, Eq, Ord, Data, Typeable)
+newtype ProperName a = ProperName { runProperName :: String }
+  deriving (Show, Read, Eq, Ord, Data, Typeable)
+
+data TypeName = TypeName deriving (Show, Read, Eq, Ord, Data, Typeable)
```
paf31: Does it make sense to define a closed data kind here?

garyb: Possibly, yeah, but I think it might interfere with `Data`, `Typeable`, and aeson's instances, as even defining these with no constructor (`EmptyDataDecls`) caused problems for them.

garyb:
Ok, that wasn't too bad after all - it's now using a data kinded phantom type. I added the extension to the cabal file as it's used so pervasively now, it that ok?

garyb:
Looks like we can't do that if we want to support GHC <7.10 :cry:

paf31:
Seems odd, `DataKinds` has been around for a while.

garyb:
It's the deriving for Data/Typeable that fails.

garyb:
What should we do about this then? Is there actually much value to providing `Data` / `Typeable` instances now, given how traversals work now, and that we're probably going to multiplate in the future.

paf31:
I'm not sure if we use data-typeable for anything else. Maybe get rid and see what happens? :smile:

garyb:
Ok, dropped those. The removal of the derived instances means propagating the change pretty far so I just stripped them all out in the end, as if the ASTs etc. don't have them it doesn't seem like there's much value in general in having them. And we provide traversals anyway. We haven't needed the instances since we stopped using SYB, but I think at the time you suggested keeping them for people using PS as a library.

garyb:
Is this okay to merge then?

paf31:
Yes I think so.

PR #1778, opened 2015-12-31
Remove errorlink in JSON to separate field

nwolverson:
I wanted to do this anyway but @natefaubion just mentioned.


paf31:
:+1: Thanks!

PR #1780, opened 2015-12-31
Fix #1779 spurious UnusedDctorImport

### LINKED ISSUE #1779: Spurious UnusedDctorImport when not importing any dctors
I have say
```
main :: Eff _ _
```
or which I think is the same:
```
```
But I get:
```
The import of type Eff includes data constructors but only the type is used
```

Since https://github.com/purescript/purescript/commit/0c31b54a71b5cb47471914eb2737aae3e7e68f3e @garyb

nwolverson:
Don't warn on explicitly importing no data constructors when none are used.


garyb:
Thanks!

PR #1782, opened 2016-01-01
Allow https://github.com URLs for psc-publish

passy:
Follow-up to #1549, #1752.

The HTTPS git access is preferable in a lot of network scenarios (e.g.
public WiFi) where other ports are restricted and should work just fine
as only read access is needed.

`https` wasn't mentioned in the original bug report, so it's possible that
I'm just missing something and that was deliberate. Please let me know. :)

/cc @codedmart @hdgarrood

[Review on Reviewable](https://reviewable.io/reviews/purescript/purescript/1782)


codedmart:
@hdgarrood maybe `regex-applicative` wasn't a bad idea?

hdgarrood:
I still think this is fine - I think `msum` is well suited here, actually.

codedmart:
Fair enough :)

paf31:
@passy Could you please update CONTRIBUTORS.md too? Thanks!

passy:
@paf31 Absolutely. Sorry I missed that before.

paf31:
Great, thanks!

Issue #1741, opened 2015-12-19
Warning suggestion on (..) imports for a type's data constructors


garyb:

natefaubion:
I know this has already landed, but I don't think `(..)` for data constructors is a dangerous thing. It's bad for module imports because it means adding stuff necessarily constitutes a breaking change because you don't know how it will clash with downstream modules. But I'd argue that adding constructors should be a breaking change anyway since it affects things like exhaustive checking. You can't transparently add/change a constructor without breaking downstream users.

paf31:
That's an interesting point, especially once we enforce the `Partial` constraints. Seems we want to distinguish between importing for matching and importing for construction, which obviously isn't going to work out well. Yet another case for exporting smart constructors/prisms, I suppose :disappointed:

garyb:
That's an excellent point... should I revert this then? I certainyly don't want to make warnings any more obnoxious than they need to be.

natefaubion:
I know I certainly find this warning somewhat tips the scales for me. Listing every constructor is obnoxious, especially for any type with more than 3 constructors. It's mostly import noise, since you already have the types to tell you where it came from. If we insist on listing everything, or qualifying everything, then I'd like to see something like [how Swift does it (with the leading dot)](https://developer.apple.com/library/mac/documentation/Swift/Conceptual/Swift_Programming_Language/Enumerations.html#//apple_ref/doc/uid/TP40014097-CH12-ID148), just to cut out some noise.

garyb:
I vote for getting rid of it since there's no "safety" issue like there is with open module imports. As you say, adding a constructor to an ADT is definitely a breaking change.

PR #1784, opened 2016-01-02
Remove warning about implicitly importing data constructors

garyb:
The warning is obnoxious and does not help with the safety issue the
same way implicit module imports does, as per later discussion in #1741


paf31:
:+1: Sounds good.

PR #1788, opened 2016-01-03
Better unused import warnings

### LINKED ISSUE #1774: Unspecified imports warning shown instead of unused import
If I have an unused implicit import of `Data.Array`:
```purescript
import Data.Array
```
previously I would be told it is a redundant import, now I am told to use the explicit import:
```purescript
import Data.Array ()
```

If I change to use this I see no warnings (this is probably unchanged); if I import a random identifier I get a redundant import warning as before. I think both forms above should show the unused import warning.

garyb:
Resolves #1774


paf31:
:+1:

PR #1798, opened 2016-01-06
Resolve #1787. Better error when let is at end of file, and missing `in`

### LINKED ISSUE #1787: Include expected indentation level in parser errors
This code:
```purescript
module Main where

main = do
  f $ \x ->
    let y = z
        z = 0

g :: Int
g = 3
```
gives this error:
```
  Unable to parse module:
  expecting no indentation or indentation
```

I guess the parser doesn't believe in the law of excluded middle? ;)

Seriously though, I wonder if we can do anything to improve this?

_Edit_ (@paf31): I think a decent fix here is to include the expected indentation level and say something like

```
  Unable to parse module:
  expecting indentation at column 9 or indentation at column 5
```

mjgpy3:
New error looks like:

```
at /home/michael/dev/contrib/purescript/examples/failing/1787.purs line 9, column 1 - line 9, column 1

  Unable to parse module:
  expecting no indentation or properly indented in keyword

```

### DIFF
--- a/examples/failing/1787.purs
+++ b/examples/failing/1787.purs
@@ -1,0 +1,10 @@
+-- @shouldFailWith ErrorParsingModule
+module Main where
+
+main = do
+  f $ \x ->
+    let y = z
+        z = 0
+
+g :: Int
+g = 3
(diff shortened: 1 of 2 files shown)


paf31:
This works, but I think there may be a more general solution. Also, we need to think about how to handle other cases like this, where a keyword follows an indent. We should probably build the solution into `indented` itself.

paf31:
#1787 has been resolved now by including indentation information in the error messages, which should handle more cases like this.

PR #1808, opened 2016-01-09
Don't apply module collision renaming to current module

### LINKED ISSUE #1789: purescript-flare has a runtime error when using 0.8.0 or master
This seems related to the fix for data types with the same name as the enclosing module: <https://github.com/sharkdp/purescript-flare/blob/6c473b8d9c1a9dcfb3628f4f838e5db2caeeef05/src/Flare.purs#L64>

If you try to build Flare, it will succeed but the compiled code will refer to the non-existent `Flare_1` module.

garyb:
Resolves #1789


paf31:
:+1: Looks good, thanks!

PR #1811, opened 2016-01-09
Fix #1796, don't print to stderr when using json errors

### LINKED ISSUE #1796: --json-errors writes things to stderr that aren't JSON
For example, warnings like
```
psc: No files found using pattern: test/*.js
```

paf31:


garyb:
:+1:

PR #1812, opened 2016-01-09
Fix #1802, fix #1797, remove redundant hints and whitespace in errors

### LINKED ISSUE #1797: JSON error messages should omit indentation
Right now, error messages have indentation and extraneous whitespace from the default error printer. It would be nice to have it sans indentation so I don't have to strip it.

### LINKED ISSUE #1802: Type unification errors contain redundant information
An error like:
```
Error found:
in module Test.Main
at /Users/nathan/Sites/purescript/purescript-psa/test/Main.purs line 6, column 3 - line 15, column 1

  A function of type

    String

    foo (foo (foo (foo (...))))


while applying a function foo
  of type String
  to argument foo (foo (foo (foo (...))))
while inferring the type of foo (foo (foo (foo (...))))
in value declaration main
```

Contains redundant information: "String can not be applied to argument while applying String to argument". The first item in the "trace" is always just a repeat of the error above.

paf31:


garyb:
:+1:

PR #1814, opened 2016-01-10
Quotes psc-docs examples.

damncabbage:
Currently:
```
Examples:
  print documentation for Data.List to stdout:
    psc-docs src/**/*.purs bower_components/*/src/**/*.purs \
      --docgen Data.List
```

With this change:
```
Examples:
  print documentation for Data.List to stdout:
    psc-docs 'src/**/*.purs' 'bower_components/*/src/**/*.purs' \
      --docgen Data.List
```

This was triggered by a good half-hour of digging around in Pulp trying to figure out why the examples from the `psc-docs` help weren't working. I eventually figured out the shell expansion was the difference between `pulp docs` working and my manual attempts not.

For example:
```
$ psc-docs 'src/**/*.purs' 'bower_components/purescript-*/src/**/*.purs' | head -n 3
## Module Main

#### `globals`
$ psc-docs src/**/*.purs bower_components/purescript-*/src/**/*.purs | head -n 3
Error found:
in module Node.FS
at /Users/rhoward/code/purescript/pulp/bower_components/purescript-node-fs/src/Node/FS.purs line 15, column 1 - line 18, column 1

  Unknown module Prelude


See https://github.com/purescript/purescript/wiki/Error-Code-UnknownModule for more information,
```

I don't know _why_ the latter fails, only that it does right now.


damncabbage:
A side-note (asked in IRC): is the `stack.yaml` for this repo to have a null for `extra-deps`, instead of `[]`? My copy of here Stack had a whinge until I made the edit myself.

garyb:
It's odd, I've seen someone else say they need quotes for the globs to work too (in fact they don't work if I use single quotes, but do with double quotes), but I definitely do not on Windows or OSX. What platform/shell are you using?

damncabbage:
```
$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14)
```
(OS X 10.10.5)

natefaubion:
It requires quotes, because otherwise bash will try to expand the globs. Which just means you'll probably end up passing in an empty list of files.

damncabbage:
The key seems to be the glob-expansion itself. Presume a directory structure that looks like:
```
abc/
  |-- file.purs
  '-- def/
        '-- file2.purs
```

Bash 3:
```
$ ls abc/**/*.purs
abc/def/file2.purs
```

Bash 4 and zsh have proper `**`-expansion (my assumption; I have access to neither from here):
```
$ ls abc/**/*.purs
abc/def/file.purs
abc/def/file2.purs
```

So it works by accident in some environments because some shells have proper expansion and some don't. When quoted `System.FilePath.Glob` will always perform the `**` expansion correctly.

garyb:
Do double quotes work for you then? As single quoting the globs prevents them from working at all on Windows at least, although double quoting does still work.

damncabbage:
Oh damn, I forgot about Windows again. Yes, double-quotes work; I've updated the branch to reflect this.

hdgarrood:
LGTM :+1:

paf31:
Thanks!

PR #1777, opened 2015-12-31
Expose error suggestions in JSON

nwolverson:
So this works for me and I think the concept is sound but scope for discussion and refinement of format in particular.

I'm starting from a point where I see the compiler telling me in a warning how I should update my import statement, but I want my editor to do it for me:
quick-fix-explicit-import-l
(https://github.com/nwolverson/atom-ide-purescript/pull/38)

Discussion on IRC showed another way of working, "I want the compiler to output a diff that I could apply". It was then suggested that this might be the result of another frontend tool making use of the compiler output.

So thoughts on this version:
  * For the cases I implemented a source span to be replaced + replacement text was enough. Not sure if there is a better representation, presumably you could construct a diff from this if you wanted
  * I can see wanting a separate source span for the suggestion more specific than for the error itself
  * I could see in some cases you might want a list of edits to make

PR #1817, opened 2016-01-10
Enhancements to import linting

### LINKED ISSUE #1740: Import suggestions don't take into account imports-via-re-export
e.g. if you `import X`, and `import Y (foo)`, and `foo` is actually a re-export from `X`, the import suggestion list for `X` will contain `foo` again.

### LINKED ISSUE #1743: Warn when importing-qualified into a virtual module without an explicit list
For similar reasons to #1742 - this will ensure name collisions can't be introduced by changes in upstream modules.

### LINKED ISSUE #1804: Warn on unused qualified import
Right now if you don't use a qualified import, no warnings are generated. It should give a redundant import warning.

garyb:
Resolves #1804, resolves #1743, resolves #1740

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -146,6 +146,10 @@ isImplicit :: ImportDeclarationType -> Bool
 isImplicit Implicit = True
 isImplicit _ = False
 
+isExplicit :: ImportDeclarationType -> Bool
+isExplicit (Explicit _) = True
+isExplicit _ = False
+
 -- |
 -- The data type of declarations
 --
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -138,7 +138,9 @@ data SimpleErrorMessage
   | DuplicateExportRef String
   | IntOutOfRange Integer String Integer Integer
   | RedundantEmptyHidingImport ModuleName
+  | ImplicitQualifiedImport ModuleName ModuleName [DeclarationRef]
   | ImplicitImport ModuleName [DeclarationRef]
+  | HidingImport ModuleName [DeclarationRef]
   | CaseBinderLengthDiffers Int [Binder]
   deriving (Show)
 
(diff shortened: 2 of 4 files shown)


garyb:
Oh, also, I realised we should warn on use of `import hiding`, as this is another form of implicit importing, so it does that too now.

nwolverson:
I was just writing a comment on #1804, I don't see immediately if it's still valid as the code has changed drastically so here it is as-was:

Just noticed that the `ImplicitImport` warning and by extension my enhanced warning in #1777 (which I should have made as a separate change) handles qualification wrongly too - if I import the same module qualified and unqualified the error lists even things that are used only qualified. `getClassName`/`getIdentName`/`getTypeName` should insist on no qualification (or rather the qualifier being `qualifierName` if it's pulled up as I did.)

garyb:
I think that's fixed now, if I understand the problem. Could you perhaps whip up an example snippet for something that would cause this problem to be apparent and I'll double check?

garyb:
Oh, I think I see, right, I'll look at that :wink:

This?

``` purescript
module Main where

import Prelude (Unit)
import Control.Monad.Eff.Console as X

main :: forall e. Eff (console :: X.CONSOLE | e) Unit
main = do
  log "Hello sailor!"

```

```
  Module Control.Monad.Eff.Console has unspecified imports, consider using the explicit form:

```

(only need to include `log` there)

paf31:
LGTM :+1: 

On a related note, does it make sense to warn on qualified uses of a name in a module without that module being explicitly imported (since we might deprecate that to support parsing module headers for incremental compilation)?

nwolverson:
Exactly. (Where the intention is to make explicit the unqualified one later)

garyb:
@nwolverson Fixed :ok_hand: - just needed to thread through the `qualifierName` into `getClassName` etc.

I'll merge once Travis okays it again.

nwolverson:
Yep that's basically what I meant, had a quick play and don't see any issues.

PR #1826, opened 2016-01-13
Relax rules for docs comments, fixes #1820

### LINKED ISSUE #1820: Relax rules for documentation comments
Currently, all of the documentation lines before a declaration must begin with a pipe character for those comments to be considered documentation (which determines whether they show up in `psc-docs` and pursuit).

This means that lots of documentation gets accidentally lost. For instance, `slice` from `arrays`:
* Source: https://github.com/purescript/purescript-arrays/blob/v0.4.4/src/Data/Array.purs#L416-L421
* Pursuit: http://pursuit.purescript.org/packages/purescript-arrays/0.4.4/docs/Data.Array#v:slice

The documentation disappears because of the preceding "Subarrays" comment.

I propose that instead of requiring every line starts with a pipe, we simply take only those that start with a pipe. I doubt that this would lead to us including anything that people didn't intend to include - I don't consider this change very risky.

See also the relevant source: https://github.com/purescript/purescript/blob/a0b92c797fdb283a1a8356249517df3e4a9093b4/src/Language/PureScript/Docs/Convert.hs#L201-L218

hdgarrood:
Comparing the output of `pulp docs` inside `arrays` before and after this change: https://gist.github.com/hdgarrood/56c437163745b250b48d

### DIFF
--- a/src/Language/PureScript/Docs/Convert/Single.hs
+++ b/src/Language/PureScript/Docs/Convert/Single.hs
@@ -11,6 +11,7 @@ module Language.PureScript.Docs.Convert.Single
 
 import Prelude ()
 import Prelude.Compat
+import Data.Maybe (catMaybes)
 
 import Control.Monad
 import Control.Category ((>>>))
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -200,21 +201,22 @@ convertDeclaration _ _ = Nothing
...
   toLines (P.BlockComment s) = lines s
 
-  hasPipe s = case dropWhile (== ' ') s of { ('|':_) -> True; _ -> False }
-
-  stripPipes = dropPipe . dropWhile (== ' ')
-
-  dropPipe ('|':' ':s) = s
-  dropPipe ('|':s) = s
-  dropPipe s = s
+  stripPipe s' =
```
garyb: Wouldn't this be more like "`takePiped`" perhaps?

garyb:
Looks good to me apart from pedantic naming thing :+1: (and I don't feel that strongly about it)

src/Language/PureScript/Docs/Convert/Single.hs
hdgarrood: When I chose this name, I was thinking along the lines of stripPrefix :: String -> Maybe String, but sure, I can change it if you think takePiped is more appropriate.

paf31:
:+1: LGTM

garyb:
:shipit: then, the name makes sense knowing the motivation :smile:

PR #1827, opened 2016-01-14
Rename `foreign` argument to fix compiling issue

anttih:
This fixes a compilation issue with ghc 7.10.2 when running `stack ghci`:

```
/Users/antti/code/purescript/purescript/psci/Types.hs:54:29:
    parse error on input ‘foreign’
```

Not exactly sure why it doesn't compile in ghci but does with `stack build`.

### DIFF
--- a/psci/Types.hs
+++ b/psci/Types.hs
@@ -51,11 +51,11 @@ mkPSCiState :: [ImportedModule]
             -> [P.Declaration]
             -> [String]
             -> PSCiState
-mkPSCiState imported loaded foreign lets nodeFlags =
+mkPSCiState imported loaded foreigns lets nodeFlags =
   (initialPSCiState
     |> each imported updateImportedModules
     |> updateModules loaded)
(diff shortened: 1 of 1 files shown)


kRITZCREEK:
I had this same problem... There was one more thing.... I think it needed a type annotation somewhere. I'll take a look.

hdgarrood:
I've just seen the same thing that @kRITZCREEK had, except with `cabal repl` (but not `cabal build`) using GHC 7.10.3. I've been using 7.10.3 for a while so I have no idea how I haven't noticed this before.

paf31:
Thanks!

hdgarrood:
Do either of you have a global `.ghci` file, in `$HOME/.ghci`? I just realised that in my case, a global `.ghci` file was the cause of the `IsString` problem that @kRITZCREEK mentioned in his comment. My `$HOME/.ghci` was enabling the `OverloadedStrings` extension. Removing that extension fixed it.

PR #1830, opened 2016-01-16
Fix #1825, generate fresh binder names unless all names in case are equal

### LINKED ISSUE #1825: Variable in scope from another binding
```purescript
module Main where

data W
    = X 
    | Y 
    | Z 

bad X a = a 
bad Y _ = a 
bad Z a = a 

main = print (bad Y 0)
```

This program compiles and prints `0`.  I'd expect it to fail compilation because `a` is not in scope in the second case.  If I rename the `a` binding in either the case above or below psc spits out `Unknown value a` for the `Y` case.

I'm using Purescript 0.7.6.1

paf31:
@andyarvanitis I hope you don't mind that I used the code from your (unfortunately neglected - sorry!) PR here? On the plus side, this clears up what I wasn't certain about (the difference between the two approaches).

### DIFF
--- a/examples/failing/1825.purs
+++ b/examples/failing/1825.purs
@@ -1,0 +1,9 @@
+-- @shouldFailWith UnknownValue
+
+module Main where
+
+data W = X | Y | Z
+
+bad X a = a 
+bad Y _ = a 
+bad Z a = a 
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

andyarvanitis:
@paf31 Not at all! Nice to see it in.

PR #1840, opened 2016-01-20
Fix generic deriving bug with >1 type argument

hdgarrood:
When trying to derive an instance for a data type which has more than one type argument, the arguments get mixed up. This appears to be because `unwrapTypeConstructor` reverses the order of the type arguments due to the way it's implemented, so I addressed it by re-reversing the type arguments at the end.

The tests in https://github.com/paf31/purescript-foreign-generic/pull/10 pass with this modification.

For example, the following code will trigger this bug:
```purescript
data MyTuple a b = MyTuple a b
derive instance genericMyTuple :: (Generic a, Generic b) => Generic (MyTuple a b)
```

It manifests in the `toSignature` implementation; the resulting `GenericSignature` has the type arguments the wrong way around. For example, `toSignature (Proxy :: Proxy (MyTuple Int String))` would have `[ \_ -> SigString, \_ -> SigInt ]` for the `sigValues`.

If we wanted to test this, we would need to compile code like the above, make assertions about the resulting `GenericSignature` in a PureScript source file, and then run that under `node`. Is that worth the effort? Maybe we can just use the tests in `purescript-foreign-generic` (once that PR is merged)? Also, I noticed `core-tests/tests/generic-deriving/Main.purs` but it doesn't seem to be referenced in any test files.

### DIFF
--- a/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
+++ b/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
@@ -17,6 +17,7 @@ import Data.List (foldl', find, sortBy)
 import Data.Maybe (fromMaybe)
 import Data.Ord (comparing)
 
+import Control.Arrow (second)
 import Control.Monad (replicateM)
 import Control.Monad.Supply.Class (MonadSupply)
 import Control.Monad.Error.Class (MonadError(..))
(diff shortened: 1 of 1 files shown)


paf31:
Great, thanks!

PR #1841, opened 2016-01-20
Fix #1839, Renaming for nested constructor binders

### LINKED ISSUE #1839: Bug when a module has the same name as constructor
```
module Main where
import Prelude
import Foo

data X = Foo Foo.Bar

eval (Foo (Foo.Bar x)) = x

main = print $ eval $ Foo None

module Foo where
data Bar = Bar Int | None

```

```
        if ($2 instanceof Foo.Bar) {
                             ^

TypeError: Expecting a function in instanceof check, but got undefined
```

sharkdp:
This fix applies the module collision renaming recursively to nested constructor binders.


garyb:
Hmm, the traversal is already supposed to do this I think, so perhaps it's that that needs updating instead. Thoughts @paf31?

paf31:
It definitely should work already. The traversal must be wrong.

sharkdp:
Okay, I gave this another try. Let me know if I'm causing more trouble instead of help :smile:.

garyb:
I can't believe that case has been missing for so long :scream:, I assume that change still fixes the related issue?

sharkdp:
Yes, it still fixes the issue. Unfortunately, I didn't know how to add a test for potential runtime errors like this.

If my research is correct, this function (`CoreFn.Traversals.everywhereOnValues`) is only called from the module renaming function, which explains why it didn't pop up anywhere else(?).

garyb:
If a test throws an exception it's considered to fail, so you could use the example from the original issue I think. I'm happy to merge this as-is though, unless you really want to add a test? :wink:

sharkdp:
I could add a test if you want, but I'm also fine with you merging it directly, of course.

garyb:
:+1:

garyb:
And thanks!

PR #1844, opened 2016-01-21
Add missing case to usedModules

### LINKED ISSUE #1837: UnknownModule instead of UnknownType
Foo.purs
```
module Foo where
```
Bar.purs
```
module Bar where
type X = Foo.X
```

```
Error found:
in module Bar
  Unknown module Foo
```

garyb:
This is to resolve #1837 but I'd kinda think there are a bunch of other errors that could occur due to this not having been handled too.

### DIFF
--- a/src/Language/PureScript/ModuleDependencies.hs
+++ b/src/Language/PureScript/ModuleDependencies.hs
@@ -23,7 +23,7 @@ import Control.Monad.Error.Class (MonadError(..))
 
 import Data.Graph
 import Data.List (nub)
-import Data.Maybe (fromMaybe, mapMaybe)
+import Data.Maybe (fromMaybe)
 
 import Language.PureScript.Crash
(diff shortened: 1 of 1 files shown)


paf31:
Do we need to handle types in instance declarations here too?

:+1: Looks good.

garyb:
I'm not sure actually... I guess we probably need to handle qualified stuff in any declaration actually. Shall I add those cases? Perhaps we should be using `everythingOnTypes` too rather than just unwrapping the first ctor we get in `forTypes`?.

garyb:
I went ahead and did it anyway: at least now (assuming our traversals are right) there is definitely not going to be any missed cases involving types, and I checked, instances are the only declaration that use a `Qualified` anything, so this should cover it all now.

paf31:
:+1: Looks good!

PR #1845, opened 2016-01-22
Support new location of `otherwise` in exhaustivity checker

garyb:


paf31:
What's the motivation for moving `otherwise`?

garyb:
I thought it would have been weird for `Prelude` to only have one member defined :smile:. `Data.BooleanAlgebra` seemed a reasonable place for it, since we don't have an actual module for `Boolean`.

I've been updating the core libraries so they don't import `Prelude` at all, but instead import from the "home" modules of the classes/types. It may make it marginally easier for people to track down references when browsing the library code, and it also means that if we do split things up further in the future when we have rewrites, etc. then potentially nothing will have to depend on `purescript-prelude` directly, and it truly will be more like a `base` kind of thing.

paf31:
Ah, I see. I'd argue `Data.BooleanAlgebra` isn't a good place actually, just because it shouldn't be overloaded and therefore isn't related to `BooleanAlgebra`. GHC has `Data.Bool`, what about `Data.Boolean`?

Actually, do we need updates for other things like `bind`?

garyb:
Sure, `Data.Boolean` makes more sense.

I did the updates for things like `bind` in that other PR earlier. `Prelude.bind`, `Prelude.(>>=)` and `Control.Monad.bind` are all accepted now, etc. for every inline-able operation.

garyb:
Updated here and in the PR for `Prelude` now.

paf31:
:+1: Looks good. Sorry to nitpick.

garyb:
No problem, I agree it's better this way too :smile:

Issue #766, opened 2014-12-12
Implicit forall


Fresheyeball:
requesting that both syntaxes be supported (at least for top level declarations)

```
foo :: forall a e. a -> e
foo :: a -> e
```

paf31:
This is another one which is discussed frequently enough that I think it merits immortality on the Discussion list.

Please cast your :+1: s and :+1: s below.

:-1: from me. My reason is that in the presence of rank N types and scoped type variables, I think it becomes very important to name type variables. Even when you're in the rank 1 case, it doesn't hurt, doesn't take long, and becomes a form of documentation.

puffnfresh:
:+1:

Because convenience and consistency.

phadej:
I'd support unicode `∀ a e. a -> e` or even `∀ a e. a → e`, then it will be as short as value-level abstraction ( `\x y -> x :: y`).

jdegoes:
:-1: Due to ambiguity, `forall` will always be required; you can "infer" what the user wants in only common cases. Therefore, for consistency, I think it should always be required.

But, a big :+1: for unicode ∀. :smile:

joneshf:
Why is `forall` only required in function type signatures and foreign type signatures?
It isn't required in data type definitions.
It isn't required in newtype definitions.
It isn't required in type aliases definitions.
It isn't required in typeclass definitions.
It isn't required in instance definitions.

How does haskell get away with not requiring it?

If you use ghc and turn on `RankNTypes` can't you still write stuff like:

```haskell
id :: a -> a
```

> Even when you're in the rank 1 case, it doesn't hurt, doesn't take long, and becomes a form of documentation.

Similar things have been said about java :wink:

garyb:
It's `ScopedTypeVariables` more than `RankNTypes` that makes the requirement more solid - scoped type vars only apply in Haskell if you use the `forall`.

I'm abstaining from voting on this one because I really don't mind either way. I'll probably keep writing `forall` out of habit, as that happens when I write Haskell now too :laughing:

natefaubion:
> It isn't required in data type definitions.
> It isn't required in newtype definitions.
> It isn't required in type aliases definitions.

I think because these aren't type signatures. They are more analogous to function declarations for types. They are declared by being on the lhs of the `=`. Classes/Instances are inconsistent, though.

joneshf:
> I'm abstaining from voting on this one because I really don't mind either way.

[...]

PR #1847, opened 2016-01-23
Extended parser to accept unicode forall and arrows

DavidLindbom:
Saw it was mentioned in #766. This commit should now accept `f ∷ ∀a m. Monad m ⇒ a → m a` but also `⇐` and `←` in class definition resp. do notation.

### DIFF
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -185,10 +185,15 @@ parsePositionedToken = P.try $ do
 parseToken :: P.Parsec String u Token
 parseToken = P.choice
   [ P.try $ P.string "<-" *> P.notFollowedBy symbolChar *> pure LArrow
+  , P.try $ P.string "←"  *> P.notFollowedBy symbolChar *> pure LArrow
   , P.try $ P.string "<=" *> P.notFollowedBy symbolChar *> pure LFatArrow
+  , P.try $ P.string "⇐"  *> P.notFollowedBy symbolChar *> pure LFatArrow
   , P.try $ P.string "->" *> P.notFollowedBy symbolChar *> pure RArrow
+  , P.try $ P.string "→"  *> P.notFollowedBy symbolChar *> pure RArrow
   , P.try $ P.string "=>" *> P.notFollowedBy symbolChar *> pure RFatArrow
+  , P.try $ P.string "⇒"  *> P.notFollowedBy symbolChar *> pure RFatArrow
   , P.try $ P.string "::" *> P.notFollowedBy symbolChar *> pure DoubleColon
+  , P.try $ P.string "∷"  *> P.notFollowedBy symbolChar *> pure DoubleColon
   , P.try $ P.char '('    *> pure LParen
   , P.try $ P.char ')'    *> pure RParen
   , P.try $ P.char '{'    *> pure LBrace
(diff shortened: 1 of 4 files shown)


src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -185,10 +185,15 @@ parsePositionedToken = P.try $ do
 parseToken :: P.Parsec String u Token
 parseToken = P.choice
   [ P.try $ P.string "<-" *> P.notFollowedBy symbolChar *> pure LArrow
+  , P.try $ P.string "←"  *> P.notFollowedBy symbolChar *> pure LArrow
```
michaelficarra: The `notFollowedBy` is unnecessary for these, no?

paf31: Is `←` a unicode symbol character? If so, it's still necessary.

paf31:
Cool :smile: Could you please update `CONTRIBUTORS.md` too?

src/Language/PureScript/Parser/Lexer.hs
michaelficarra: Oh so you're saying that that may be used as a binding name?

paf31: Yes, since we added unicode operators. @DavidLindbom could you please just verify that something like `←→` can be used as an operator (ideally in a test)?

DavidLindbom: Yes they should work, just updated my test.

garyb:
:+1: for this :smile:

zudov:
Very cool, can't wait for this to get merged. But what about unicode lambda?

garyb:
> But what about unicode lambda?

Oh yeah, that'd be nice to have too.

DavidLindbom:
I will take a look at it later. I tried a similar quick fix but `λ` isn't a symbol but a identifier so I have to do some more work.

paf31:
I don't think we need a new lexeme. We can just use `reserved` to parse a lambda, surely?

DavidLindbom:
No `reserved` doesn't work because `λa` is a valid identifier. Lexing `λa → a` will result in `Right ["λa",->,"a"] instead of `Right [\,"a",->,"a"]`.

paf31:
Oh, that's a shame. I'm not sure what the best option is, in that case. 

Perhaps we merge without `λ`, as you had it before, in 0.8, and revisit `λ` in 0.8.*. That would be my vote.

Another option is to use a space between the `λ` and the binder, but that's not going to look very good.

In either case, though, I think we can solve this without adding `\` as a lexeme. Maybe a better solution is for `char 'λ' *> notFollowedBy letter` to be a lexeme.

garyb:
Related discussion for the same thing in GHC: https://ghc.haskell.org/trac/ghc/ticket/1102

DavidLindbom:
@paf31 yes I agree that it would be best to revert to the previous commit and discuss later how to solve the lambda problem

paf31:
Could you please rebase, and I'll get this merged. Thanks.

paf31:
I'm sorry, by rebase I meant "and squash commits". Could you please do that so we can keep the commit log as simple as possible?

paf31:
Looks great, thanks! I'll wait for Travis and then get this merged in.

AppShipIt:
@DavidLindbom Would you be interested in showing this off in 5-10 minutes at the next online meetup? Are you on Twitter? Otherwise I can email.

AppShipIt:
> Maybe a better solution is for `char 'λ' *> notFollowedBy letter` to be a lexeme.

I had the same idea. Were there any downsides?

PR #1851, opened 2016-01-25
Add failing test case for TCO bug

hdgarrood:
The following code causes a stack overflow where n >= 1 on `master`, but works as intended on 0.7.6.1. I had a quick look at the generated JS but wasn't able to quite pin down what was happening.

PR #1852, opened 2016-01-25
Workaround for TCO/composition inlining interaction

garyb:
Resolves #1851

### DIFF
--- a/examples/passing/TCO.purs
+++ b/examples/passing/TCO.purs
@@ -1,0 +1,20 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (print)
+
+main = do
+  let f x = x + 1
+  let v = 0
+  print (applyN 0 f v)
+  print (applyN 1 f v)
+  print (applyN 2 f v)
+  print (applyN 3 f v)
+  print (applyN 4 f v)
+
+applyN :: forall a. Int -> (a -> a) -> a -> a
+applyN = go id
+  where
+  go f n _ | n <= 0 = f
+  go f n g = go (f >>> g) (n - 1) g
+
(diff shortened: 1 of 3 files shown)


garyb:
It's arguable as to whether this is a fix or a workaround, but I'd lean more to the side of fix: running a composed chain of functions can cause a stack overflow just recursing directly would, so although this means TCO won't trigger for the example case now, it's probably no worse than it triggering and then building a function that will stack overflow anyway.

paf31:
Looks ok, but we should test this like crazy before merging :smile:

garyb:
Alright, this seemed like the best way to handle it in the end: I was originally going to split out the "magicDo loops" and put those with TCO at the end, but I was not too confident that was going to work correctly.

The optimizations in `tidyUp` only remove code, so applying them again after we introduce `JSWhile` should not be an issue, as it's the introduction of additional functions where things go wrong. I think we still need to run these, as the post-TCO code can be pretty ugly as it introduces even more variables that are "unnecessary" that this `tidyUp` pass removes.

I unrelatedly renamed `inlineValues` to `inlineCommonValues` to make it clear it's a function that inlines specific values, unlike `inlineVariables` which applies to any variables that match its rules.

paf31:
:+1: Looks great, thanks!

paf31:
Does this affect performance at all?

garyb:
The performance of the compiler, or of the generated code? Probably both to a certain extent. :smile:

The compiler as there's basically another pass on the JS codegen now, but codegen has never been our problem for performance anyway.

By running `tidyUp` post-TCO/`whileE` it should apply most of the relevant optimisations to the JS still though (eliminating unnecessary IIFEs, eta conversions, etc), so I don't think there will be any serious impact. It may be that TCO doesn't trigger in some situations now where it used to, but those cases would not benefit anyway (see [earlier comment](https://github.com/purescript/purescript/pull/1852#issuecomment-174536905)).

paf31:
:+1: Looks good to me. One more question though - do we need to run `tidyUp

garyb:
What was that about `tidyUp`? It got cut off in the email and I see you've removed it from here, but if there's anything at all to discuss let's talk... I really worry about making changes in here :smile:

paf31:
Oh, no problem. I was wondering if it was possible to only call it in one of the loops, but I can see it is enabling other optimizations.

Issue #1853, opened 2016-01-25
When duplicate patterns exist an unhelpful warning is given


dkoontz:
Given the following code

```purescript
module Main where

data Options = Foo | Bar | Baz

brokenPatternMatch :: Options -> Boolean
brokenPatternMatch Foo = true
brokenPatternMatch Foo = true
brokenPatternMatch Bar = false
brokenPatternMatch Baz = false
```

produces the warning `A case expression contains unreachable cases: Baz` which is probably technically true but not very helpful. "duplicate pattern Foo" would seem better.

paf31:
No, that'd be a bug. `Baz` is not unreachable. The error should be that `Foo` is unreachable.

cc @nicodelpiano, any thoughts on this?

nicodelpiano:
Yes, it seems to be the redundancy checker. Let me check that out.

PR #1854, opened 2016-01-26
Solves #1853

nicodelpiano:
The problem was the condition `cond` we were using: I compared it with an older version of the redundancy checker that worked well and changed that piece of code.

Now it's working fine:

```purescript
module Main where

data Options = Foo | Bar | Baz

brokenPatternMatch :: Options -> Boolean
brokenPatternMatch Foo = true
brokenPatternMatch Foo = true
brokenPatternMatch Bar = false
brokenPatternMatch Baz = false
```

Throws:

```
  in module Main
  at /home/delpi/Desktop/Development/purescript/psctests/test.purs line 5, column 1 - line 6, column 1

    A case expression contains unreachable cases:

    Foo

  in value declaration brokenPatternMatch
```

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -31,13 +31,13 @@ import Language.PureScript.Kinds
 import Language.PureScript.Types as P
 import Language.PureScript.Errors
 
--- | There are two modes of failure for the redudancy check:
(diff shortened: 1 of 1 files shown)


paf31:
Thanks for looking into this!

nicodelpiano:
No problem Phil :+1:, if there's another issue just let me know!

garyb:
Nice work :+1:

Issue #1834, opened 2016-01-17
psc-docs: internal error on values without type signatures


hdgarrood:
(this is a different problem to #1831)

`psc-docs` and `psc-publish` will currently give an internal error if you run them on a module with any exported values which do not have type signatures. For example:

```purescript
module Foo where
foo = 0
```
using `psc-docs` from current master:
```
$ PATH=../purescript/dist/build/psc-docs:$PATH pulp docs
* Generating documentation in /home/harry/code/ps-scratch
psc-docs: An internal error ocurred during compilation: "Docs.Convert.ReExports: lookupValueDeclaration: unexpected result:\nother: []\nident: Ident \"foo\"\ndecls: []"
Please report this at https://github.com/purescript/purescript/issues
* ERROR: Subcommand terminated with exit code 1
```

Previously, it would have just ignored them, so `foo` wouldn't appear in the output. The way I see it, the easiest way forward for now would be to just produce a better error in this case. Perhaps later, we could start using inferred types for exported declarations which are missing documentation; see also https://github.com/purescript/pursuit/issues/21.

On an unrelated note, can we stop calling `show` on the argument to `Language.PureScript.Crash.internalError`? It's already a `String` so I think it would make errors easier to read if we left them as-is.

sharkdp:
I just experienced this error when running `psc-docs` with purescript `master` on an old codebase (it was running fine previously). It took me 15min to figure out what was going on - that is, until I found this bug report. The error message is very irritating.

I know its best-practice to have explicit top-level type declarations, but I guess this bug will be experienced by a lot of people if it doesn't get fixed until 0.8. It basically renders `psc-docs` unusable until you have fixed all your type declarations (or exports) - which could be considered okay, but not with this error message.

Please excuse the non-constructive post.

hdgarrood:
I understand your frustration. On the other hand, there are two points I think should be made:

* whenever you elect to use some software project at master instead of a released version, all bets are off, effectively. If the maintainers thought it was ready to go, it would have been released.
* the error does state pretty clearly that it's an internal error. Not understanding what that means is perhaps reasonable, though. What "internal error" basically means is that, at the time the code was written, the scenario that led to this error being thrown did not occur, or was considered impossible, to whoever was writing this code (in this case, me). Given that, at the time I wrote the code, I did think this error would never be thrown, I actually think it's very appropriate. Of course, with hindsight, it's rather bad, but that's beside the point.

[...]

PR #1859, opened 2016-01-31
Infer psc docs types

hdgarrood:
WIP fix for #1834 

I'm not quite sure why this doesn't work. The error I'm getting is "Unknown type constructor Prelude.Unit", from `examples/docs/src/UTF8.purs`.

`examples/docs` is set up like a normal PureScript package, and notably, `Prelude` is in `bower_components`. So I was wondering if that might have something to do with it. However, it's failing during type checking, before we remove dependencies modules from the output, so maybe that's just a red herring. I was also wondering if I haven't set up the `CheckEnv`, or specifically, the `Environment`, correctly.

### DIFF
--- a/src/Language/PureScript/Docs/Convert/Single.hs
+++ b/src/Language/PureScript/Docs/Convert/Single.hs
@@ -139,6 +139,8 @@ convertDeclaration (P.ValueDeclaration _ _ _ (Right (P.TypedValue _ _ ty))) titl
   basicDeclaration title (ValueDeclaration ty)
 convertDeclaration (P.ValueDeclaration _ _ _ _) title =
   P.internalError ("Should have been desugared: " ++ title)
+convertDeclaration (P.ExternDeclaration _ ty) title =
+  basicDeclaration title (ValueDeclaration ty)
 convertDeclaration (P.DataDeclaration dtype _ args ctors) title =
   Just (Right (mkDeclaration title info) { declChildren = children })
   where
--- a/src/Language/PureScript/Docs/Convert/Single.hs
+++ b/src/Language/PureScript/Docs/Convert/Single.hs
@@ -109,7 +109,7 @@ addDefaultFixity decl@Declaration{..}
   defaultFixity = P.Fixity P.Infixl (-1)
 
 getDeclarationTitle :: P.Declaration -> Maybe String
-getDeclarationTitle (P.TypeDeclaration name _)               = Just (P.showIdent name)
+getDeclarationTitle (P.ValueDeclaration name _ _ _)          = Just (P.showIdent name)
 getDeclarationTitle (P.ExternDeclaration name _)             = Just (P.showIdent name)
 getDeclarationTitle (P.DataDeclaration _ name _ _)           = Just (P.runProperName name)
 getDeclarationTitle (P.ExternDataDeclaration name _)         = Just (P.runProperName name)
@@ -135,10 +135,10 @@ basicDeclaration :: String -> DeclarationInfo -> Maybe IntermediateDeclaration
 basicDeclaration title info = Just $ Right $ mkDeclaration title info
 
 convertDeclaration :: P.Declaration -> String -> Maybe IntermediateDeclaration
-convertDeclaration (P.TypeDeclaration _ ty) title =
+convertDeclaration (P.ValueDeclaration _ _ _ (Right (P.TypedValue _ _ ty))) title =
   basicDeclaration title (ValueDeclaration ty)
-convertDeclaration (P.ExternDeclaration _ ty) title =
-  basicDeclaration title (ValueDeclaration ty)
+convertDeclaration (P.ValueDeclaration _ _ _ _) title =
+  P.internalError ("Should have been desugared: " ++ title)
 convertDeclaration (P.DataDeclaration dtype _ args ctors) title =
   Just (Right (mkDeclaration title info) { declChildren = children })
   where
--- a/src/Language/PureScript/Sugar.hs
+++ b/src/Language/PureScript/Sugar.hs
@@ -14,8 +14,9 @@
 -----------------------------------------------------------------------------
 
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TupleSections #-}
 
-module Language.PureScript.Sugar (desugar, module S) where
+module Language.PureScript.Sugar (desugar, desugarWithEnv, module S) where
 
 import Prelude ()
 import Prelude.Compat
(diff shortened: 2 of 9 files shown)


hdgarrood:
Just to clarify: I'm opening this now because I'm stuck. Any suggestions?

hdgarrood:
It turns out, at some point between sorting the modules and the desugared modules coming out of `desugarWithEnv`, they were being reversed, so that they were in a reverse topological order, which of course broke type checking. I have no idea how, so I've just reversed them back again for now. Currently investigating.

At least the tests pass now, though. :)

hdgarrood:
I found out where the modules are being reversed: https://github.com/purescript/purescript/blob/2fffbcb748793d98c059ab69e0b144273368c238/src/Language/PureScript/Sugar/Names.hs#L57

No idea how this isn't breaking things elsewhere, or if we should be fixing that there...?

hdgarrood:
Oh also, I wonder if `ParseAndDesugar` has become a bit of a misnomer, as it's doing type-checking as well now. Really it's for transforming modules into a state where they can be used by psc-docs and psc-publish. I've been unable to come up with a better name so far.

src/Language/PureScript/Sugar.hs
```diff
@@ -65,14 +66,25 @@ import Language.PureScript.Sugar.TypeDeclarations as S
...
-    >=> createBindingGroupsModule
+    >=> desugarImportsWithEnv externs
+    >=> withSecond
+        (   rebracket externs
+        >=> traverse deriveInstances
+        >=> desugarTypeClasses externs
+        >=> createBindingGroupsModule
+        )
+  where
+  withSecond f (x, y) = (x,) <$> f y
```
paf31: `withSecond` is `fmap . fmap` isn't it?

paf31:
Looks good, but I'd like to figure out the `reverse` issue. I think that it works in the regular compiler because the list of modules is always length-one now (after the parallel build changes). I think the better place to fix it, if possible, would be in `Names.hs`.

garyb:
:+1: for reversing after the `foldM` in names.

src/Language/PureScript/Sugar.hs
hdgarrood: Apparently not:

```
λ: let withSecond = (fmap . fmap) :: Functor f => (a -> f b) -> (z, a) -> f (z, b)

<interactive>:12:26:
    Couldn't match type ‘a1’ with ‘(z1, a1)’
      ‘a1’ is a rigid type variable bound by
           an expression type signature:
             Functor f1 => (a1 -> f1 b1) -> (z1, a1) -> f1 (z1, b1)
           at <interactive>:12:18
    Expected type: (a1 -> f1 b1) -> a1 -> (z1, b1)
      Actual type: (a1 -> b1) -> (z1, a1) -> (z1, b1)
    In the second argument of ‘(.)’, namely ‘fmap’
    In the expression:
        (fmap . fmap) :: Functor f => (a -> f b) -> (z, a) -> f (z, b)

<interactive>:12:26:
    Couldn't match type ‘b1’ with ‘f1 b1’
      ‘b1’ is a rigid type variable bound by
           an expression type signature:
             Functor f1 => (a1 -> f1 b1) -> (z1, a1) -> f1 (z1, b1)
           at <interactive>:12:18
    Expected type: (a1 -> f1 b1) -> a1 -> (z1, b1)
      Actual type: (a1 -> b1) -> (z1, a1) -> (z1, b1)
    In the second argument of ‘(.)’, namely ‘fmap’
    In the expression:
        (fmap . fmap) :: Functor f => (a -> f b) -> (z, a) -> f (z, b)
```

I'm not a fan of using the tuple instances in cases like these anyway, I think `\f (x, y) -> (x,) <$> f y` is always going to be clearer than anything using instances like `Functor ((,) a)`.

hdgarrood:
Yeah, I think @paf31 is right. Looking at `Make` I can see why it doesn't make a difference to `psc` whether or not we reverse it. It's also easy to see why the list gets reversed if you look at the definition of `updateEnv`, so I think we can consider this issue figured out. :)

I changed it to reverse inside `Sugar.Names`, rebased, and pushed, so I think this is good to go now.

src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -135,10 +135,10 @@ basicDeclaration :: String -> DeclarationInfo -> Maybe IntermediateDeclaration
 basicDeclaration title info = Just $ Right $ mkDeclaration title info
 
 convertDeclaration :: P.Declaration -> String -> Maybe IntermediateDeclaration
-convertDeclaration (P.TypeDeclaration _ ty) title =
-  basicDeclaration title (ValueDeclaration ty)
-convertDeclaration (P.ExternDeclaration _ ty) title =
```
garyb: Does the inferred type come out as exactly same as the annotated cases? (I'm thinking in case of type variable numbering, etc.)

hdgarrood: I had assumed that explicit type declarations, where given, would be left alone by the compiler, but I will check this. But I think I've found another bug in the meantime...

hdgarrood: Yes, it does seem to. For example. `Prelude.unsafeCompare` comes out as `forall a. a -> a -> Ordering`.

garyb: Great, I thought it probably would, just thought it was worth checking as in error messages we end up with renamed type variables.

hdgarrood:
I've just found a couple more bugs. Give me a day or two.

user471:
Would it fix a problem when functions doesn't show up in TAGS file if they doesn't have type declaration?

hdgarrood:
I don't think so, tags files are handled differently.

hdgarrood:
I think what is happening is that some of the desugaring steps (in particular, type class desugaring) mean that what we're getting is too desugared. For instance, type class members have started appearing at the top level in the output.

What I'm going to try now is to revert `ParseAndDesugar` to what it was before, although perhaps add in the `desugarTypeDeclarationsModule` so that it's easy to tell top level value declarations which have explicit type signatures from those that don't, and then make a note of all the value declarations which are missing type signatures, and *at the end*, typecheck everything and add the missing type signatures then. So the pipeline would be like this:

* Partially desugar (as before)
* Convert each module individually (`Convert.Single`), taking explicit type signatures where they exist
* Fully desugar everything
* Typecheck everything
* Add any missing type signatures using the result of type checking
* Add any re-exported declarations

garyb:
Sounds like a good approach. I forgot about the sugar passes before typechecking, so yeah I can see how that output on its own wouldn't be too useful; without the prior information it'd be impossible to reconstruct some of the declarations.

hdgarrood:
Cool :) using the above approach, do you think the best way to extract types after type checking would be to take them from the `Environment`? That is, if I typecheck all the modules together like I did here, will the environment from the final `CheckState` have everything in it?

garyb:
Yep!

hdgarrood:
Awesome, it seems to work :D I'll clean this up a bit and push a new branch.

Issue #1833, opened 2016-01-17
Allow operator aliases to be declared for data constructors


garyb:

user471:
Would operator aliases work with pattern matching?

cryogenian:
What about type constructors? e.g. `type a \/ b = Either a b`

garyb:
@user471 maybe, I hadn't considered that. We'd need to do some work on the parser, but it might well be possible.

@cryogenian I'm in favour of that too, but it's a separate issue. I'd like it to work much like it does for value operators, so you can still set associativity and precedence.

This issue is just about tweaking the parser to accept propernames as well as idents.

paf31:
What's the use case here? Shouldn't we be encouraging smart constructors?

jdegoes:
Smart constructors don't provide a solution to pattern matching, e.g.:

```purescript
foo :: forall a b c d. Tuple4 a b c d -> d
foo (a /\ b /\ c /\ d) = d
```

paf31:
I think we should be very careful not to introduce too many opportunities for abusing operators.

jdegoes:
@paf31 I'd argue that a discrepancy between construction and deconstruction leads to code that is inherently difficult to read. Compare:

```purescript
let foo = a /\ b /\ c /\ d
in case foo of
  (Tuple (Tuple (Tuple a b) c) d) -> d
```

with:

```purescript
let foo = a /\ b /\ c /\ d
in case foo of
  a /\ b /\ c /\ d -> d
```

If you allow operators, then operators will be used for constructors. If operators are used for constructors, then it greatly improves readability to allow them for deconstructors.

In addition, I'd suggest that PureScript (the language) should be fairly un-opinionated about whether operators are good or bad, and let that decision live in user-land communities so that every (niche) community can make their own choice about that.

paf31:
> I'd suggest that PureScript (the language) should be fairly un-opinionated about whether operators are good or bad

Sounds reasonable. Let's just be careful to avoid overuse in core libraries then.

jdegoes:
> Sounds reasonable. Let's just be careful to avoid overuse in core libraries then.

:+1:

garyb:
[...]

PR #1861, opened 2016-02-03
#1846, start work on '_ - 1' syntax for sections

### LINKED ISSUE #1846: Allow using (- 1) for operator section?
I would like to be able to use `(- 1)` as an operator section that subtracts one. Currently it is parsed as `negate 1`. I still want to be able to negate things with a prefix `-` operator. I think both could be allowed by using whitespace. If there is a space, use the `-` binary operator. If there's not a space, use `negate`. 

``` purescript
(-1)    -- negate 1
(- 1) x -- x - 1 -- (`sub` 1)
```

Obviously this behaves differently than all other operators. But I would prefer it to using `sub` as a makeshift section.

paf31:
Resolves #1846.


paf31:
This is probably ready to review, even though I still need to deprecate the old syntax.

Here are some ideas for other anonymous functions:

- `case` (aka _lambda case_): `runFoo = case _ of Foo x -> x`
- `if..then..else`
- Array literals (just like we can section object literals)

paf31:
Looks like the Stackage CI builds are failing. Does anyone have any idea why?

LiamGoodacre:
@paf31 the `STACKAGE` variable is wrong in some of the build scripts.  https://travis-ci.org/purescript/purescript/jobs/106913297#L161

~~~
$ export STACKAGE=lts=3.6
~~~

It should be `lts-3.6`.

This causes the `curl` to fail. :smile_cat:

LiamGoodacre:
I don't know why the nightly one is failing though as this exists: https://www.stackage.org/nightly-2015-09-29/cabal.config

garyb:
- :+1: for lambda case, as we'd discussed elsewhere.
- What about `if b then _ else x`, `if _ then _ else _` etc.? I've more often wanted to "curry" the other parts than the boolean, but I'm not sure how parseable that is.
- The array literal one makes me nervous for some reason actually, but I don't know why, it seems like a logical counterpart to record placeholders.

Code looks good to me. Do we need tests for things like `1 + _ * x` too? (I assume that's an error without parens, before becoming a type error). And what about:

``` purescript
_ `fn` x
```
?

jdegoes:
```
case (aka lambda case): runFoo = case _ of Foo x -> x
if..then..else: (if _ then 0 else 1) :: Boolean -> Int
Array literals (just like we can section object literals): [1, _] :: Int -> Array Int.
```

:+1: 

This change is good because it's consistent with record "sections" and is much more friendly: `(1 + _)` versus `(+1)`, e.g. One of the examples of PureScript improving on Haskell.

hdgarrood:
Re Stackage, see #1864

LiamGoodacre:
@garyb I'd say parentheses should be required and it shouldn't be supported for arbitrary expressions, just sections and literals: `1 + _ * x` seems a bit much.  

Like-wise for lambda `case` and `if`.

~~~purescript
-- required parentheses
foo = (if _ then A else B)
bar = (case _ of X -> Y)
~~~

If you allow `(if a then _ else _)` I feel it'd be inconsistent to not also have

~~~purescript
(case a of
   X -> _
   Y -> _)

-- meaning
\x y -> case a of
          X -> x
          Y -> y
~~~

@paf31 I do like the array literal case `[1, _]`

paf31:
Should we require parens on object updates and literals eventually?

paf31:
Ok, ready to review again. Let's do lambda case and other things in another PR.

paf31:
Notice there is less backtracking in the parser now, due to less ambiguity when seeing an underscore. This should help with the currently-terrible parse errors inside object updates.

src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -24,30 +24,51 @@ desugarObjectConstructors (Module ss coms mn ds exts) = Module ss coms mn <$> ma
...
+    , isAnonymousArgument u = return $ OperatorSection op (Left val)
+    | b' <- stripPositionInfo b
+    , BinaryNoParens op u val <- b'
+    , isAnonymousArgument u = return $ OperatorSection op (Right val)
+  desugarExpr (ObjectLiteral ps) = wrapLambda ObjectLiteral ps
+  desugarExpr (ObjectUpdate u ps) | isAnonymousArgument u = do
     obj <- freshIdent'
     Abs (Left obj) <$> wrapLambda (ObjectUpdate (Var (Qualified Nothing obj))) ps
-  desugarExpr (ObjectGetter prop) = do
+  desugarExpr (ObjectUpdate obj ps) = wrapLambda (ObjectUpdate obj) ps
```
garyb: Use an `otherwise` guard here and combine with the previous pattern?

garyb:
Looks good to me. Nice that we could get rid of a bunch of constructors from the AST too. :+1:

jdegoes:
> Should we require parens on object updates and literals eventually?

Probably. Then the "underscore rule" can be stated simply: underscores inside a parenthetical expression are mechanically translated into curried lambda expressions.

paf31:
@garyb Shall I merge this then?

When do you want to push out 0.8.1? We have a good few PRs already.

Actually, I wanted to discuss the idea of biweekly or monthly releases, where we try to commit to something minimal, like one PR each, and whatever bug fixes can go into those releases too. What do you think?

garyb:
:shipit: 

As for 0.8.1... it's not even been a week yet :laughing:. I'd really like to get #1833 in first, I have started working on it.

paf31:
Oh, I know, I was thinking we could aim for a release next weekend?

tfausak:
:+1: This is awesome!

AppShipIt:
@LiamGoodacre or @tfausak Would either of you be interested in showing off this feature in 5-10 minutes at the next online meetup since y'all proposed/motivated it? Sorry I couldn't raise Liam on Twitter.

tfausak:
I'd be down for that. Send me an email or tweet to work out the details.

AppShipIt:
I'll DM you before the weekend!

PR #1863, opened 2016-02-04
Infer types in psc-{docs,publish}

### LINKED ISSUE #1834: psc-docs: internal error on values without type signatures

hdgarrood:
Fixes #1834

Instead of taking types from the partially-desugared AST, we now use
placeholder types for all value declarations in the first pass,
then we typecheck all the modules, and insert the real types for all
value declarations later on, using the Environment obtained during
typechecking.

This involved:

* Changing the module name field in the Docs.Types.Module type to a real
  P.ModuleName; to be frank, it really ought to have been that type to
  begin with.

* Refactoring ParseAndDesugar so that it no longer does any desugaring.
  With this new pipeline, it is easier to pass modules that are not
  desugared at all into Docs.Convert. Also renamed ParseAndDesugar
  appropriately.

* Re-reversing the modules in Sugar.Names; desugarImports was previously
  (unintentionally) reversing the list of modules. This new code depends
  on those modules remaining in a topologically sorted order, which is
  why this change was necessary.

### DIFF
--- a/examples/docs/src/ExplicitTypeSignatures.purs
+++ b/examples/docs/src/ExplicitTypeSignatures.purs
@@ -1,0 +1,16 @@
+
+module ExplicitTypeSignatures where
+
+-- This should use the explicit type signature so that the type variable name
+-- is preserved.
+explicit :: forall something. something -> something
+explicit x
+  | true = x
+  | false = x
+
+-- This should use the inferred type.
+anInt :: _
+anInt = 0
+
+-- This should infer a type.
+aNumber = 1.0
--- a/psc-docs/Main.hs
+++ b/psc-docs/Main.hs
@@ -69,19 +69,19 @@ docgen (PSCDocsOptions fmt inputGlob output) = do
     Etags -> dumpTags input dumpEtags
     Ctags -> dumpTags input dumpCtags
     Markdown -> do
-      ms <- runExceptT (D.parseAndDesugar input []
-                           >>= ((\(ms, _, env) -> D.convertModulesInPackage env ms)))
+      ms <- runExceptT (D.parseAndBookmark input []
+                           >>= (fst >>> D.convertModulesInPackage))
                >>= successOrExit
 
       case output of
         EverythingToStdOut ->
           putStrLn (D.runDocs (D.modulesAsMarkdown ms))
         ToStdOut names -> do
-          let (ms', missing) = takeByName ms (map P.runModuleName names)
+          let (ms', missing) = takeByName ms names
           guardMissing missing
           putStrLn (D.runDocs (D.modulesAsMarkdown ms'))
         ToFiles names -> do
-          let (ms', missing) = takeByName' ms (map (first P.runModuleName) names)
+          let (ms', missing) = takeByName' ms names
           guardMissing missing
 
           let ms'' = groupBy ((==) `on` fst) . sortBy (compare `on` fst) $ map swap ms'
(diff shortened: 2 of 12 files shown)


hdgarrood:
Also:

* Even though the `Module` data type changed, the JSON format should remain the same, because I used `P.moduleNameFromString` and `P.runModuleName` in the parser and `ToJSON` instance respectively.

* I was a bit worried that this would be difficult and awkward, but it turned out to be super easy, because of how well organised and documented the compiler is. So thanks for that! :fireworks:

hdgarrood:
Ah - I've just realised that doing it this way expands type synonyms, which is not what we want. Give me a minute, and I'll make it prefer existing type signatures over compiler-generated ones.

garyb:
Tying in with that, I was going to suggest, perhaps we don't even need to do the typecheck stage if every member has an explicit type? It might give a nice performance boost for that case then.

hdgarrood:
Ah yeah, that makes sense.

Can you think of a good "sentinel" `P.Type` value to use inside the `DeclarationInfo` for value declarations to mean "this didn't have an explicit type signature"? Maybe `TUnknown (-1)`?

The alternative is to do something like parameterize `DeclarationInfo` like this:

```
data DeclarationInfo f
  = ValueDeclaration (f P.Type)
```

and then put `Maybe ~ f` on the first pass, and then later use `Identity ~ f`, but that seems a bit awkward.

hdgarrood:
Oh - maybe we could use `TypeWildcard`, which already means "put the type in for me". It would probably improve the output whenever people actually use wildcards as well.

garyb:
Nice, I like that. I was going to say I prefer the parameterised version, as that better expresses what's going on when looking at the types, but `TypeWildcard` is a perfect sentinel for this already.

paf31:
What about a wildcard?

hdgarrood:
I think the build is failing because stackage is giving us 301 redirects to HTTPS. I guess this is new?

paf31:
@garyb @hdgarrood What's the status of this? Should it go into 0.8.1?

hdgarrood:
I've used it a few times, I'm happy that it works (or at least, that it's a huge improvement over 0.8.0). :+1: :+1: :+1: for putting it into 0.8.1.

paf31:
Thank you both!

paf31:
I'm getting "The import of ‘Data.Monoid’ is redundant" and a failing build after merging this. Not fully sure if it's related to this change though.

kRITZCREEK:
Just pushed a change removing that import to the psc-ide-integration branch. So if that still is a problem it'll get resolved with the psc-ide PR

hdgarrood:
Seems like it was, looking at the changes in test/TestDocs.hs - I think the import was only for `(<>)`, and I guess Prelude.Compat exports `(<>)` now. On the other hand, that doesn't explain why it was passing before and failing now.

PR #1864, opened 2016-02-04
Use stackage over HTTPS

hdgarrood:
Stackage has started redirecting permanently to HTTPS, which is breaking
the CI build, so this should fix it.


paf31:
:+1: Thanks!

PR #1870, opened 2016-02-06
Eq and Ord deriving

### LINKED ISSUE #1874: Eq and Ord deriving
Sorry to create this now, I originally thought there was an issue for it, but I'd like to put it into 0.8.1 since I think it's very useful for users and simple enough to implement.

paf31:
Resolves #1874.

### DIFF
--- a/examples/passing/Deriving.purs
+++ b/examples/passing/Deriving.purs
@@ -9,6 +9,10 @@ derive instance eqX :: Eq X
 
 derive instance ordX :: Ord X
 
+newtype Z = Z { left :: X, right :: X }
+
+derive instance eqZ :: Eq Z
+
 main = do
   assert $ X 0 == X 0
   assert $ X 0 /= X 1
@@ -17,3 +21,5 @@ main = do
   assert $ X 0 < X 1
   assert $ X 0 < Y "Foo"
   assert $ Y "Bar" < Y "Baz"
+  assert $ z == z where
+    z = Z { left: X 0, right: Y "Foo" }
--- a/examples/passing/Deriving.purs
+++ b/examples/passing/Deriving.purs
@@ -1,0 +1,19 @@
+module Main where
+
+import Prelude
+import Test.Assert
+
+data X = X Int | Y String
+
+derive instance eqX :: Eq X
+
+derive instance ordX :: Ord X
+
+main = do
+  assert $ X 0 == X 0
+  assert $ X 0 /= X 1
+  assert $ Y "Foo" == Y "Foo"
+  assert $ Y "Foo" /= Y "Bar"
+  assert $ X 0 < X 1
+  assert $ X 0 < Y "Foo"
+  assert $ Y "Bar" < Y "Baz"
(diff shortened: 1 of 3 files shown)


jdegoes:
In an ideal world, there would be no coupling to the Prelude library and the library itself would implement (through something like template PureScript) the auto-derivation. Although that requires a ton of machinery that does not yet exist, and plenty of people want Eq/Ord now, so :+1:

paf31:
True. Another option is to bake things like `Eq` and `Ord` into `Prim` and put the code gen into the constraint solver. But that's kinda gross.

paf31:
Ok, this is ready to review.

hdgarrood:
Presumably it still works when you want to derive an instance which has a field that itself derives the instance? Perhaps the tests should cover that too? Not sure.

Also, does this work for records, like `newtype X = X { foo :: Int, bar :: String }`? If so, I guess the test should include a record too?

paf31:
Good points @hdgarrood, thanks. It should work in both cases. I'll modify the tests.

paf31:
Actually, the problem with `Ord` on records is that the labels are unordered, so lexicographical ordering doesn't make sense. So perhaps we should disallow those cases?

paf31:
I guess the point is that we're just picking some ordering for records, not that there is a canonical one, but  since most `Ord` instances will just be used to put things into maps and sets, I think that's okay.

Good to review again.

hdgarrood:
Re Ord and records: yeah, that's what I was about to say. :+1:

jdegoes:
> Actually, the problem with Ord on records is that the labels are unordered, so lexicographical ordering doesn't make sense. So perhaps we should disallow those cases?

Well, on sum types, the order is the declared "text" order of the terms in the sum. It'd be nice if the order for fields were the declared "text" order.

paf31:
@jdegoes Yes, it is.

jdegoes:
:+1:

paf31:
@jdegoes @hdgarrood Another option is to use the ordering given by taking labels in alphabetical order. Thoughts on this before this goes into 0.8.1?

hdgarrood:
I think I'd expect the ordering to be given by the order labels were written, although if I really cared about it, I'd probably just write the Ord instance out explicitly. So, no particularly strong feelings either way.

jdegoes:
> Another option is to use the ordering given by taking labels in alphabetical order. 

Very strongly opposed to this on the following grounds:

1. It's inconsistent with derived `Ord` for sum types, in which the ordering is the declared order. Observation: users know this and will arrange their constructors to obtain a desired ordering (e.g. `data DoW = Sunday | Monday | Tuesday ...`).
2. It's inconsistent with derived `Ord` for the "other" product type (`data`), in which the ordering is the declared order. Observation: users know this and will arrange their constructors to obtain a desired ordering (e.g. `data OrderDetails = OrderDetails OrderId ...`.
3. If a user is looking at a derived `Ord` for a record, what will they *predict* the ordering will be? They could randomly guess "alphabetic ascending", "alphabetic descending", or some such, but given both (1) and (2), and given the user has *already declared* an ordering of the fields (declaration order), I believe very strongly most users will predict declared order. If we want to totally surprise users and make their predictions wrong, then we need to have a very good reason for doing so.

PR #1871, opened 2016-02-06
fix stack.yaml syntax error

kRITZCREEK:
Without this change stack fails with:
```
Could not parse '/home/creek/Documents/purescript/stack.yaml':
AesonException "Error in $.extra-deps: failed to parse field 'extra-deps': failed to parse field extra-deps: expected [a], encountered Null"
See http://docs.haskellstack.org/en/stable/yaml_configuration.html.
```

### DIFF
--- a/stack-lts-3.yaml
+++ b/stack-lts-3.yaml
@@ -1,5 +1,5 @@
 flags: {}
 packages:
 - '.'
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Thanks!

PR #1873, opened 2016-02-06
Tidy up the tests

hdgarrood:
Also enable and fix warnings in the tests, and rearrange a bit.

I can't remember exactly why the psc-publish tests were disabled, but we'll see what happens when we re-enable them now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ bower_components/
 node_modules
 tmp/
 .stack-work/
-tests/support/flattened/
 output
 examples/docs/docs/
 core-tests/full-core-docs.md
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,9 @@ language: c
 sudo: false
 matrix:
   include:
-    - env: GHCVER=7.8.4 COVERAGE_SUITE=tests
+    - env: GHCVER=7.8.4 COVERAGE=true
       compiler: ": #GHC 7.8.4 - tests"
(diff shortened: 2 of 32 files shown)


hdgarrood:
Ah - what's happening is that the test support code, which we usually fetch using bower, isn't there for the dist tests. I think we should just check it all in and only rerun it locally when that support code needs updating. That would allow us to remove purescript from the allow-failures thing on Stackage too.

hdgarrood:
Unfortunately the diff is almost impossible to read now so I'll sum up what's happened:

* Enable warnings on the test suites, and fix all the warnings
* Move the psci tests into the "tests" suite, and remove the "psci-tests" suite (having two separate test suites is just a bit weird and confusing, I think)
* Check in all the support code instead of fetching it every time, so that the tests don't need npm or bower to run (so, eg, Stackage can run them)
* Re-enable the psc-publish tests, which work now that the support code is checked in.
* Rename all the psci modules like `Types` -> `PSCi.Types` so that it's clearer what they're for, especially in the tests.
* Tidy up the tests a bit so that the structure is a bit clearer - specifically, `tests/Main.hs` no longer has any tests directly inside it. Instead, it imports tests from other modules, like `TestPscPublish`, `TestCompiler`, etc.
* Remove the need to import `TestPscPublish` from `TestDocs` - it was only because of `pushd`, which has been moved into `TestUtils`.

paf31:
Thanks @hdgarrood! This looks great.

Issue #1672, opened 2015-11-25
Further reduce import boileplate using `NoImplicitPrelude`


zudov:
@phadej

After https://github.com/purescript/purescript/pull/1610 we have `import Prelude ()` in quite many modules. Is there something that prevents us from using `NoImplicitPrelude`. I think we could even do it package-wide from `purescript.cabal`. I doubt purescript developers would be scared of `import Prelude` on top of every file :p

*P.S: I should have ask that in original PR, but didn't notice it back then since I was too lazy to scroll down the diff a little bit

hdgarrood:
While we're at it, it's probably worth adding a bunch of other pragmas to `purescript.cabal`, to reduce noise from adding/removing them in diffs. For example, I think the following (at least) is reasonable:

* `FlexibleContexts`
* `ScopedTypeVariables`
* `RecordWildCards`
* `RankNTypes`
* `TupleSections`
* `DeriveFunctor`
* `DeriveDataTypeable`
* `OverloadedStrings`
* `GADTs`

zudov:
@hdgarrood I would disagree on adding `OverloadedStrings` there since it makes the type interference a bit harder and you sometimes have to put explicit type definitions for string literals.

:+1: to the rest

paf31:
Yes, I agree with @zudov.

hdgarrood:
Ok, that does make sense.

zudov:
Yesterday @phadej said that he has nothing against `NoImplicitPrelude`.

I think I could go around doing that. I'll then add the following instances to cabal file:

```
NoImplicitPrelude
FlexibleContexts
ScopedTypeVariables
RecordWildCards
RankNTypes
TupleSections
DeriveFunctor
DeriveDataTypeable
GADTs
```

garyb:
I've already added `DataKinds` globally in #1773 too (maybe, depending on whether we keep the approach in there).

nwolverson:
`DeriveDataTypeable` no longer used.

PR #1876, opened 2016-02-08
Adds psc-ide to the project

kRITZCREEK:
## Todos:
- [x] Move tests into the testsuite
- [x] Tidy up dependencies and figure out version ranges
- [x] Figure out where to put the protocol documentation/readme
- [x] GHC < 7.10 compatibility

## Why did I move the library code over aswell?
When I moved just the executables cabal/stack would complain about the circular dependency between psc-ide <=> purescript.

## Pursuit Client
These dependencies result out of the Pursuit querying module, I could drop it for now as it really doesn't offer a lot of features and revise it later using a more lowlevel interface like conduit-http or pipes-http. The Pursuit client is also responsible for the orphan instance warning. I'm not quite sure how to get around that warning without making a way too big "Types" module.
```
wreq,
lens,
lens-aeson,
http-client
```

Edit:
I switched to using `pipes-http` in this PR

## Renaming psc-ide to psc-ide-client
I think this is a more consistent name.

### DIFF
--- a/psc-ide-server/Main.hs
+++ b/psc-ide-server/Main.hs
@@ -1,0 +1,135 @@
+{-# LANGUAGE ConstraintKinds       #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE PackageImports        #-}
+{-# LANGUAGE TemplateHaskell       #-}
+module Main where
+
+import           Prelude                           ()
+import           Prelude.Compat
+
+import           Control.Concurrent                (forkFinally)
+import           Control.Concurrent.STM
+import           Control.Exception                 (bracketOnError)
+import           Control.Monad
+import           "monad-logger" Control.Monad.Logger
+import           Control.Monad.Reader
+import           Control.Monad.Trans.Except
+import qualified Data.Text                         as T
+import qualified Data.Text.IO                      as T
+import           Data.Version                      (showVersion)
+import           Language.PureScript.Ide
+import           Language.PureScript.Ide.CodecJSON
+import           Language.PureScript.Ide.Error
+import           Language.PureScript.Ide.Types
+import           Language.PureScript.Ide.Watcher
+import           Network                           hiding (socketPort)
+import           Network.BSD                       (getProtocolNumber)
+import           Network.Socket                    hiding (PortNumber, Type,
+                                                    accept, sClose)
+import           Options.Applicative
+import           System.Directory
+import           System.FilePath
+import           System.IO
+
+import qualified Paths_purescript                  as Paths
+
+-- "Borrowed" from the Idris Compiler
+-- Copied from upstream impl of listenOn
+-- bound to localhost interface instead of iNADDR_ANY
+listenOnLocalhost :: PortID -> IO Socket
+listenOnLocalhost (PortNumber port) = do
+  proto <- getProtocolNumber "tcp"
+  localhost <- inet_addr "127.0.0.1"
+  bracketOnError
+    (socket AF_INET Stream proto)
+    sClose
+    (\sock -> do
+      setSocketOption sock ReuseAddr 1
+      bindSocket sock (SockAddrInet port localhost)
+      listen sock maxListenQueue
+      pure sock)
+listenOnLocalhost _ = error "Wrong Porttype"
+
+data Options = Options
+  { optionsDirectory  :: Maybe FilePath
+  , optionsOutputPath :: FilePath
+  , optionsPort       :: PortID
+  , optionsDebug      :: Bool
+  }
+
+main :: IO ()
+main = do
+  Options dir outputPath port debug  <- execParser opts
+  maybe (pure ()) setCurrentDirectory dir
+  serverState <- newTVarIO emptyPscIdeState
+  cwd <- getCurrentDirectory
+  _ <- forkFinally (watcher serverState (cwd </> outputPath)) print
+  let conf =
+        Configuration
+        {
+          confDebug = debug
+        , confOutputPath = outputPath
+        }
+  let env =
+        PscIdeEnvironment
+        {
+          envStateVar = serverState
+        , envConfiguration = conf
+        }
+  startServer port env
+  where
+    parser =
+      Options <$>
+        optional (strOption (long "directory" <> short 'd')) <*>
+        strOption (long "output-directory" <> value "output/") <*>
+        (PortNumber . fromIntegral <$>
+         option auto (long "port" <> short 'p' <> value (4242 :: Integer))) <*>
+        switch (long "debug")
+    opts = info (version <*> parser) mempty
+    version = abortOption
+      (InfoMsg (showVersion Paths.version))
+      (long "version" <> help "Show the version number")
+
+startServer :: PortID -> PscIdeEnvironment -> IO ()
+startServer port env = withSocketsDo $ do
+  sock <- listenOnLocalhost port
+  runLogger (runReaderT (forever (loop sock)) env)
+  where
+    runLogger = runStdoutLoggingT . filterLogger (\_ _ -> confDebug (envConfiguration env))
+
+    loop :: (PscIde m, MonadLogger m) => Socket -> m ()
+    loop sock = do
+      (cmd,h) <- acceptCommand sock
+      case decodeT cmd of
+        Just cmd' -> do
+          result <- runExceptT (handleCommand cmd')
+          $(logDebug) ("Answer was: " <> T.pack (show result))
+          liftIO (hFlush stdout)
+          case result of
+            -- What function can I use to clean this up?
+            Right r  -> liftIO $ T.hPutStrLn h (encodeT r)
+            Left err -> liftIO $ T.hPutStrLn h (encodeT err)
+        Nothing -> do
+          $(logDebug) ("Parsing the command failed. Command: " <> cmd)
+          liftIO $ do
+            T.hPutStrLn h (encodeT (GeneralError "Error parsing Command."))
+            hFlush stdout
+      liftIO (hClose h)
+
+
+acceptCommand :: (Applicative m, MonadIO m, MonadLogger m)
+                 => Socket -> m (T.Text, Handle)
+acceptCommand sock = do
+  h <- acceptConnection
+  $(logDebug) "Accepted a connection"
+  cmd <- liftIO (T.hGetLine h)
+  $(logDebug) cmd
+  pure (cmd, h)
+  where
+   acceptConnection = liftIO $ do
+     (h,_,_) <- accept sock
+     hSetEncoding h utf8
+     hSetBuffering h LineBuffering
+     pure h
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -14,7 +14,8 @@ Homepage: http://www.purescript.org/
 author: Phil Freeman <paf31@cantab.net>,
         Gary Burgess <gary.burgess@gmail.com>,
         Hardy Jones <jones3.hardy@gmail.com>,
-        Harry Garrood <harry@garrood.me>
+        Harry Garrood <harry@garrood.me>,
+        Christoph Hegemann <christoph.hegemann1337@gmail.com>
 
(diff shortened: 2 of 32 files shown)


hdgarrood:
That type signature in Language.PureScript.Parser.Lexer shouldn't be necessary, as OverloadedStrings shouldn't be enabled, neither in that module, nor in purescript.cabal. I think you must be enabling it with something else. A global `.ghci` file, perhaps?

kRITZCREEK:
@hdgarrood I did enable it in this commit. psc-ide uses Text extensibly. I could either annotate all of the psc-ide source files or we could make the compiler OverloadedStrings "safe". I think it is one of the more common extensions.

hdgarrood:
Ah, I see. I just thought I'd mention it because in https://github.com/purescript/purescript/issues/1672 people weren't very keen on OverloadedStrings. Is that change in the lexer the only one that the compiler needs to build with OverloadedStrings on, then?

kRITZCREEK:
@hdgarrood "works on my machine" :smile:

Regarding the travis failure: I'm using `Control.Monad.Except` which is only available in the latest `mtl`. It is however deprecating `Control.Monad.Error`. What's the policy here? Should I bump the mtl dependency or stick with `Control.Monad.Error` for now? Looking around the codebase suggests we are using `Control.Monad.Error` for now

hdgarrood:
There are no imports of `Control.Monad.Error` in the compiler - they would cause the build to fail because of the deprecation warnings. If you want `MonadError`, that's in `Control.Monad.Error.Class`. For `ExceptT`, there's the one in `transformers-compat`. I think it's `Control.Monad.Trans.Except`.

kRITZCREEK:
So all builds except for the lts-2 build work now. It fails because it can't satisfy the fsnotify dependency but I don't know how to get around this... I also had to disable the orphan warning for the Pursuit client...

src/Language/PureScript/Ide/Externs.hs
```diff
@@ -0,0 +1,90 @@
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE RecordWildCards       #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# OPTIONS_GHC -fno-warn-unused-do-bind #-}
```
hdgarrood: Is this necessary?

src/Language/PureScript/Ide/Pursuit.hs
```diff
@@ -0,0 +1,110 @@
...
+import           Data.Text                     (Text)
+import qualified Data.Text                     as T
+import           Language.PureScript.Ide.Types
+import           Network.HTTP.Client           (HttpException (StatusCodeException))
+import           Network.Wreq
+
+import           Text.Parsec
+import           Text.Parsec.Text
+
+instance FromJSON PursuitResponse where
```
hdgarrood: Why not define this instance in the same module as PursuitResponse?

hdgarrood:
We'll need HTTPS for the pursuit stuff, and Haskell doesn't come with any SSL library, so this might mean that people need to install OpenSSL separately to get this to work? might be worth checking, especially for Windows. This might be a job better suited for purescript on node even, since that does come with HTTPS.

src/Language/PureScript/Ide/Pursuit.hs
hdgarrood: Sorry, should have read the issue description more thoroughly :)

IMO a large types module is significantly better than an orphan instance.

zudov:
@hdgarrood Is there anything wrong with `http-client` + `http-client-tls`?

hdgarrood:
I don't know, but IIRC they use OpenSSL and expect to get it via a shared library? I just think it's worth checking on Windows.

paf31:
`tls` is a pure Haskell implementation, no need for OpenSSL, AFAIK.

zudov:
At least according to [`tls` description](https://hackage.haskell.org/package/tls):

> Native Haskell TLS and SSL protocol implementation for server and client.

I haven't used it on Windows myself, but I would expect it to work.

hdgarrood:
Ah ok, I wasn't remembering correctly. ignore me then :)

src/Language/PureScript/Ide/Externs.hs
kRITZCREEK: Absolutely not. Thanks for finding it!

kRITZCREEK:
At this point I feel kind of stuck. I have no idea why Travis fails with

```
tests/Main.hs:28:18:
    Could not find module ‘TestPscIde’
    Use -v to see a list of the files searched for.
Failed to install purescript-0.8.0.0
cabal: Error: some packages failed to install:
purescript-0.8.0.0 failed during the building phase. The exception was:
ExitFailure 1
```
in the 7.10.2 build, when the file gets found just fine in the other runs.

For the 7.8.4 lts build I'd need to either manipulate the downloaded `cabal.config` file to change the version of fsnotify or work around it with CPP but I have never done that before.

For the 8.0.1 build I suspect I need to pin down some dependencies but again I have no idea of how to find the right versions.

hdgarrood:
Just checking: is it because you need to add TestPscIde to other-modules somewhere?

kRITZCREEK:
@hdgarrood awesome! That seems to fix the 7.10.2 build. The 0.8.1 build apparently fixed itself, so now only the 7.8.4 `fsnotify` problem remains. The concrete issue is, that fsnotify switched from `system-filepath` to the built-in `filepath` from 0.1.* to 0.2.*.

hdgarrood:
Drop support for lts 2 then? They're up to 6 now IIRC.

zudov:
@hdgarrood Would that mean that we don't support 7.8 anymore?

hdgarrood:
No, the GHC 7.8.4 build works fine, so I don't think that's necessary. I think we would just stop putting subsequent versions of PureScript into later versions of Stackage lts-2. Come to think of it, if 0.7.x was in lts-2, will lts-2 ever upgrade to 0.8.x anyway?

hdgarrood:
Oh, I see - purescript isn't in lts-2, but even so, we have that build in order to check that purescript works when its dependencies are coming from an lts-2 snapshot.

They're up to lts-5 now (not 6, that was wrong, but I think 6 is coming soon with aeson 0.11), so I think it should be fine to just remove the lts-2 build completely. We aren't worrying about backwards compatibility / supporting older versions for anything else yet.

kRITZCREEK:
I guess this is up to @paf31 to decide. I'd of course like not to have to fiddle with CPP. You can even build the project against lts2 if you use the stack-lts2.yaml file.

purescript.cabal
```diff
@@ -79,7 +79,16 @@ library
                    process >= 1.2.0 && < 1.5,
                    safe >= 0.3.9 && < 0.4,
                    semigroups >= 0.16.2 && < 0.19,
-                   parallel >= 3.2 && < 3.3
+                   parallel >= 3.2 && < 3.3,
+                   stm,
+                   regex-tdfa,
+                   edit-distance,
```
paf31: Could you please rerun `license-generator` and make sure everything is compatible with MIT?

paf31:
I'll take a proper look over this tomorrow. 

Don't forget to add yourself to the Copyright section in the `.cabal` file, if you like, or in individual files.

psc-ide-client/client/Main.hs
```diff
@@ -0,0 +1,56 @@
+{-# LANGUAGE OverloadedStrings #-}
```
paf31: This file looks like a duplicate?

psc-ide-server/Main.hs
```diff
@@ -0,0 +1,161 @@
...
+      setSocketOption sock ReuseAddr 1
+      bindSocket sock (SockAddrInet port localhost)
+      listen sock maxListenQueue
+      pure sock)
+listenOnLocalhost _ = error "Wrong Porttype"
+
+data Options = Options
+  { optionsDirectory  :: Maybe FilePath
+  , optionsOutputPath :: FilePath
+  , optionsPort       :: Int
```
paf31: Does it make sense to use the `PortNumber` type here?

purescript.cabal
paf31: Also, is it possible to constrain the bounds here for those old timers like me who haven't adopted `stack` yet? :wink:

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -523,4 +523,4 @@ reservedTypeNames = [ "forall", "where" ]
 -- The characters allowed for use in operators
 --
 isSymbolChar :: Char -> Bool
-isSymbolChar c = (c `elem` ":!#$%&*+./<=>?@\\^|-~") || (not (isAscii c) && isSymbol c)
+isSymbolChar c = (c `elem` (":!#$%&*+./<=>?@\\^|-~" :: String)) || (not (isAscii c) && isSymbol c)
```
paf31: Is this because of `OverloadedStrings`? I'd rather avoid having it turned on globally, for this reason.

paf31:
I haven't reviewed this in full, because I think the code has already proved to work well. I reviewed things generally, and it all looks good to me so far. I made a couple of minor notes.

@kRITZCREEK What do you want to do about PRs which have been merged on your other repo since this was opened? Can you port them over?

Also, what do you want to do about things like issues? I could create a label, and we could move them over here if you like (seems like the most sensible option to me).

Thanks very much for working on this!

kRITZCREEK:
I think I adressed all your notes.

I already ported the PRs against the original repository into this PR.

I think moving the issues under something like a `psc-ide` tag sounds like the most sensible thing.

It's a little easier to see the most recent changes before I rebase, so I'll wait until you can check that your suggestions have been properly adressed. Give me the word and I'll smash it all together :smiling_imp: 

The (from my point of view) last remaining issue is that this breaks compatibility with lts-2.

hdgarrood:
:+1: from me. I'm not a fan of partial functions in data declarations [like this](https://github.com/purescript/purescript/pull/1876/files#diff-86e52ea0bf901b253b094f131a1bfdc0R19), and maybe it would be nice to change that at some point in the future; anyway, as @paf31 said, the code clearly works.

FrigoEU:
Great change :+1:

paf31:
@hdgarrood I assume you're referring to things like this (your link didn't work for me):

```haskell
data Command
    = Load { loadModules      :: [ModuleIdent]
           , loadDependencies :: [ModuleIdent]}
    | Type { typeSearch  :: DeclIdent
           , typeFilters :: [Filter]}
```

I'm not a big fan of that, either, but perhaps we can address it later, if need be. The Aeson instances are all hand-written in any case.

kRITZCREEK:
I thought about wrapping up the "records" inside newtypes. Would that fix the partiality problem?

paf31:
@kRITZCREEK Yes, at the expense of more indirections. This is one of the benefits of extensible records, of course :wink:

hdgarrood:
Yep, that's what I meant (the link is broken for me too now, not sure what happened there).

kRITZCREEK:
@hdgarrood force-pushing happened :D I'll rebase now.

paf31:
So, are we dropping support for LTS 2 then? Seems fine to me.

cc @phadej

paf31:
:fireworks: :+1: 

Now we need to figure out a way to move issues over. I think there are some automated tools.

phadej:
Sorry for not noticing this. Dropping `lts-2` support is ok, if time passes by. But one have to update `stack-lts-2.yaml` by either bumping `fsnotify` version there, or renaming it to `stack-ghc-7.8.yaml` and using `ghc-7.8.4` resolver with manual `extra-deps`.

I could also add `stack` based builds to travis so we don't forget this next time (and add matrix entries / `stack.yaml` for `lts-5`).

kRITZCREEK:
@phadej I did add the `fsnotify-0.2.0` `extra-deps`. So if you are using stack you can still build against lts-2. It's just that with the fixed `cabal.config` file it won't build.

phadej:
@kRITZCREEK afaik you forgot to add
```
- hfsevents-0.1.6
- pipes-http-1.0.2
```

`fsnotify` depends on `hfsevents`, and `pipes-http` was forgotten to add previously, I guess.

PR #1884, opened 2016-02-13
refactor PSCi code

izgzhen:
I simply factored the old `PSCi.hs` into several modules (including `PSCi.IO`, `PSCi.Message`, `PSCi.Option`, `PSCi.Module`, `PSCi.Printer`). No real modification is made.

The reason is that, the old module is relatively hard to read and maintain (over 600 lines of code). And by splitting out different sub-functionalities, the code might be reused somewhere else in the future.

Besides, I added `Paths_purescript` to `other-modules` fields in .cabal file, so some warnings can be eliminated.

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -7,16 +7,14 @@
 -- |
 -- PureScript Compiler Interactive.
 --
-module PSCi where
+module PSCi (runPSCi) where
 
 import Prelude ()
 import Prelude.Compat
 
 import Data.Foldable (traverse_)
-import Data.Maybe (mapMaybe)
-import Data.List (intersperse, intercalate, nub, sort, find)
+import Data.List (intercalate, nub, sort, find)
 import Data.Tuple (swap)
-import Data.Version (showVersion)
 import qualified Data.Map as M
 
 import Control.Arrow (first)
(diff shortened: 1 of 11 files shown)


psci/PSCi/IO.hs
```diff
@@ -14,8 +14,42 @@
 
 module PSCi.IO where
 
-import System.Directory (createDirectoryIfMissing)
-import System.FilePath (takeDirectory)
+import System.Directory
```
hdgarrood: I think I'd rather stick to explicit import lists, especially if the code had them to start with. Not only do they help avoid name collisions, they also make it easier to see where a function is coming from.

hdgarrood:
Did you see the build failures on GHC <7.10? I think they're because functions like `<$>` have only been in the Prelude since GHC 7.10, so the answer is probably to start using `base-compat`. Most compiler modules start with this:

```haskell
import Prelude ()
import Prelude.Compat
```

so we could do the same here.

izgzhen:
@hdgarrood Yes, I am fixing it ... Thank you very much :D

psci/PSCi.hs
```diff
@@ -531,55 +343,7 @@ handleKindOf typ = do
...
-  loadAllImportedModules
-handleCommand (TypeOf val) = handleTypeOf val
-handleCommand (KindOf typ) = handleKindOf typ
-handleCommand (BrowseModule moduleName) = handleBrowse moduleName
-handleCommand (ShowInfo QueryLoaded) = handleShowLoadedModules
-handleCommand (ShowInfo QueryImport) = handleShowImportedModules
-handleCommand QuitPSCi = P.internalError "`handleCommand QuitPSCi` was called. This is a bug."
+-- Misc
 
 whenFileExists :: FilePath -> (FilePath -> PSCI ()) -> PSCI ()
```
hdgarrood: This could be generalised to `MonadIO`, right? If so, I guess it should go into `PSCi.IO` with the other stuff?

izgzhen: Do you mean `handleXXX`? It seems like the logic in the handlers involves nearly every components (`inputT`, `StateT` and `IO`) in the transformer stack, so it might not be easy to factor out.

The `PSCi.IO` is basically inheriting the idea that all pure `IO` helper functions should be put here in the previous version (although there was only a lonely `mkdirp` then).

hdgarrood: No, I mean `whenFileExists`. I agree that `handleXXX` should stay here.

hdgarrood: Ah, `PSCI` has no `MonadIO` instance. @paf31 How do you feel about using a derived `MonadIO` instance for `PSCI`? I'm pretty sure it would end up being the same as `psciIO`.

hdgarrood: Oh wait, no, we don't need a `MonadIO PSCI` instance to be able to do that, I don't think.

hdgarrood:
> Besides, I added Paths_purescript to other-modules fields in .cabal file, so some warnings can be eliminated.

I'm not sure that this is the right thing to do. Which warnings are eliminated? Did you see any documentation that recommends doing this?

hdgarrood:
Oh wait, never mind, this is recommended by the Cabal user guide: https://www.haskell.org/cabal/users-guide/developing-packages.html#accessing-data-files-from-package-code :+1:

izgzhen:
Fixed above:

* The `Prelude` import
* Explicit import in `PSCi.IO`
* Move around the `whenFileExists`

psci/PSCi/IO.hs
```diff
@@ -14,8 +14,55 @@
...
+
+-- |
+-- Expands tilde in path.
+--
+expandTilde :: FilePath -> IO FilePath
+expandTilde ('~':p:rest) | isPathSeparator p = (</> rest) <$> getHomeDirectory
+expandTilde p = return p
+
+
+whenFileExists :: MonadIO m => FilePath -> (FilePath -> InputT m ()) -> InputT m ()
```
hdgarrood: `InputT` itself has a `MonadIO` instance, so I don't think this function needs to refer to `InputT` at all, does it?

izgzhen: This is due to the `outputStrLn :: MonadIO m => String -> InputT m ()` in the `else` branch

hdgarrood: Oh, of course, sorry. Hm. I wonder if it was better off specialised to `PSCI` then, and in the `PSCi` module... I'll defer to others' judgement here.

paf31:
@izgzhen Thanks for this!

PR #1890, opened 2016-02-21
Operator aliases for data constructors

garyb:
Still needs work :wink:

### DIFF
--- a/src/Language/PureScript/AST/Binders.hs
+++ b/src/Language/PureScript/AST/Binders.hs
@@ -41,6 +41,24 @@ data Binder
   --
   | ConstructorBinder (Qualified (ProperName 'ConstructorName)) [Binder]
   -- |
+  -- A operator alias binder. During the rebracketing phase of desugaring,
+  -- this data constructor will be removed.
+  --
+  | OpBinder (Qualified Ident)
+  -- |
+  -- Binary operator application. During the rebracketing phase of desugaring,
+  -- this data constructor will be removed.
+  --
+  | BinaryNoParensBinder Binder Binder Binder
+  -- |
+  -- Explicit parentheses. During the rebracketing phase of desugaring, this
+  -- data constructor will be removed.
+  --
+  -- Note: although it seems this constructor is not used, it _is_ useful,
+  -- since it prevents certain traversals from matching.
+  --
+  | ParensInBinder Binder
+  -- |
   -- A binder which matches a record and binds its properties
   --
   | ObjectBinder [(String, Binder)]
@@ -70,6 +88,8 @@ binderNames = go []
   where
   go ns (VarBinder name) = name : ns
   go ns (ConstructorBinder _ bs) = foldl go ns bs
+  go ns (BinaryNoParensBinder b1 b2 b3) = foldl go ns [b1, b2, b3]
+  go ns (ParensInBinder b) = go ns b
   go ns (ObjectBinder bs) = foldl go ns (map snd bs)
   go ns (ArrayBinder bs) = foldl go ns bs
   go ns (NamedBinder name b) = go (name : ns) b
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -113,6 +113,7 @@ data SimpleErrorMessage
   | InvalidNewtype (ProperName 'TypeName)
   | InvalidInstanceHead Type
   | TransitiveExportError DeclarationRef [DeclarationRef]
+  | TransitiveDctorExportError DeclarationRef (ProperName 'ConstructorName)
   | ShadowedName Ident
   | ShadowedTypeVar String
   | UnusedTypeVar String
@@ -145,6 +146,7 @@ data SimpleErrorMessage
   | HidingImport ModuleName [DeclarationRef]
   | CaseBinderLengthDiffers Int [Binder]
   | IncorrectAnonymousArgument
+  | InvalidOperatorInBinder Ident Ident
   deriving (Show)
 
 -- | Error message hints, providing more detailed information about failure.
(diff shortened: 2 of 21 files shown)


src/Language/PureScript/AST/Binders.hs
```diff
@@ -53,6 +53,16 @@ data Binder
...
+  -- Binary operator application. During the rebracketing phase of desugaring, this data constructor
+  -- will be removed.
+  --
+  | BinaryNoParensBinder (Qualified Ident) Binder Binder
+  -- |
+  -- Explicit parentheses. During the rebracketing phase of desugaring, this data constructor
+  -- will be removed.
+  --
+  | ParensInBinder Binder
+  -- |
```
garyb: I haven't really made use of this yet, as the parsing is not correct, but I assume we'll need something like this to enable fixity overriding as with the normal operators?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -281,9 +283,17 @@ typeCheckAll moduleName _ ds = traverse go ds <* traverse_ checkFixities ds
...
     ty <- lookupVariable moduleName alias
     addValue moduleName (Op name) ty Public
+  checkFixities (FixityDeclaration _ name (Just (Right alias))) = do
+    env <- getEnv
+    case M.lookup alias (dataConstructors env) of
+      Nothing -> throwError . errorMessage $ UnknownDataConstructor alias Nothing
+      Just (_, _, ty, _) -> do
+        let v = Constructor alias
+        (_, ty') <- sndM (introduceSkolemScope <=< replaceAllTypeSynonyms) <=< instantiatePolyTypeWithUnknowns v $ ty
+        addValue moduleName (Op name) ty' Public
```
garyb: This is basically just lifted from `infer`, do I actually need to do all this here, or would it be safe to just `addValue` the `ty`?

garyb:
Ok, I think this is good to review now. Do you know what's going on with the errors Travis keeps giving? Something about aeson? It's been building locally without errors or warnings for me for the last few pushes, yet keeps failing on Travis.

src/Language/PureScript/Errors.hs
```diff
@@ -789,6 +793,10 @@ prettyPrintSingleError full level showWiki e = flip evalState defaultUnknownMap
       paras [ line $ "An export for " ++ prettyPrintExport x ++ " requires the following to also be exported: "
             , indent $ paras $ map (line . prettyPrintExport) ys
             ]
+    renderSimpleErrorMessage (TransitiveDctorExportError x ctor) =
+      paras [ line $ "An export for " ++ prettyPrintExport x ++ " requires the following data constructor to also be exported: "
+            , indent $ line $ runProperName ctor
+            ]
```
garyb: I did this with a new error message as we can't really generate a good error message for the export using a `DeclarationRef` and the `TransitiveDctorExportError`, as it would mean doing quite a bit of work to figure out if the type for the data constructor is already exported, and also what data constructors are exported for that type. Also `TransitiveDctorExportError` doesn't print data constructors currently, for some other complicated reasons (data constructors for the types in here are `error` sometimes).

src/Language/PureScript/TypeChecker.hs
```diff
@@ -383,6 +388,11 @@ typeCheckModule (Module ss coms mn decls (Just exps)) = warnAndRethrow (addHint
     exports (PositionedDeclarationRef _ _ r1) r2 = exports r1 r2
     exports r1 (PositionedDeclarationRef _ _ r2) = exports r1 r2
     exports _ _ = False
+    -- We avoid Eq for `nub`bing as the dctor part of `TypeRef` evaluates to
+    -- `error` for the values generated here (we don't need them anyway)
+    nubEq (TypeRef pn1 _) (TypeRef pn2 _) = pn1 == pn2
+    nubEq r1 r2 = r1 == r2
+
```
garyb: The addition of `nub`bing here isn't specifically related to the other changes, just a little improvement to prevent a type from being printed multiple times in the error message in some situations.

garyb:
Ok, so there was a mistake in that last commit that was causing a failure despite what I said :wink:

The test suite is still failing locally for me, but I have no idea why, as I can't interpret the output anymore. Having scanned through it I see nothing appears to be failing in the psc tests, but it still comes out with `Test suite tests: FAIL`, so something in docs or publish? I don't know what the docs test output means (like, are the things it prints failures or just what it is doing?), and the publish test output is empty.

garyb:
Oh, it fails in `master` too...

hdgarrood:
That might be my fault. Some of the tests print to stderr, and some print to stdout, so the actual error might be a few lines back (I am hoping to fix this soon). Otherwise, I can a look at this tomorrow if you'd like.

garyb:
I was about to ask if you had any ideas actually, as the tests started failing once `infer-docs-types-2` was merged :wink:: https://travis-ci.org/purescript/purescript/builds/110685540

garyb:
Ok, well it seems Travis is happy now even though it isn't locally. Weird!

src/Language/PureScript/Sugar/Operators.hs
```diff
@@ -20,51 +20,83 @@ module Language.PureScript.Sugar.Operators (
...
+import Data.Maybe (mapMaybe)
 import qualified Data.Map as M
 
-import qualified Text.Parsec as P
-import qualified Text.Parsec.Pos as P
-import qualified Text.Parsec.Expr as P
-
 import qualified Language.PureScript.Constants as C
 
+type AliasName = Either (Qualified Ident) (Qualified (ProperName 'ConstructorName))
```
paf31: Could you please add some short comments here?

src/Language/PureScript/TypeChecker.hs
paf31: Looks like you removed this, which I think is probably fine, and correct.

paf31:
:+1: Looks great to me!

garyb:
Rebased & I added some comments for `AliasName`, etc.

PR #1899, opened 2016-02-25
Update travis 2016-02-25

phadej:
Adding `stack` build would be the follow-up PR as it needs a bit more of though


paf31:
Thanks @phadej!

PR #1903, opened 2016-02-28
Allow one open import without warning

### LINKED ISSUE #1869: Permit a single open import without a warning
Or: "What happened to `(..)`?"

Let's say I have a module `B` that should reexport everything from `A`, like so:

``` purescript
-- A.purs
module A where data X
-- B.purs
module B (module A) where import A
```

Compiling that with PureScript 0.8.0, I get the following warning:

```
Warning found:
in module B
at /private/tmp/example/src/B.purs line 1, column 27 - line 1, column 34

  Module A has unspecified imports, consider using the explicit form:

    import A (X)



```

I used to be able to explicitly get open imports by saying `import A (..)`, but that's a syntax error now.

``` purs
-- B.purs
module B (module A) where import A (..)
```

```
Error found:
at /private/tmp/example/src/B.purs line 1, column 37 - line 1, column 37

  Unable to parse module:
  unexpected ..
  expecting identifier, (, proper name, "class", "module" or )


See https://github.com/purescript/purescript/wiki/Error-Code-ErrorParsingModule for more information,
```

What's the right way to do this?

garyb:
Resolves #1869


examples/failing/ConflictingExports2.purs
```diff
@@ -1,13 +0,0 @@
...
-  thing :: Int
-  thing = 1
-
--- Fails here because re-exporting forces any scope conflicts to be resolved
-module Main (thing, module A) where
-
-  import A
-
-  thing :: Int
-  thing = 2
```
garyb: This is now a warning rather than error, so the test no longer fails (`thing` is resolved to `Main.thing` due to it being a local declaration, and therefore always preferred over a conflicting value from an open import).

src/Language/PureScript/Errors.hs
```diff
@@ -604,6 +606,10 @@ prettyPrintSingleError full level showWiki e = flip evalState defaultUnknownMap
       paras [ line $ "Conflicting definitions are in scope for " ++ nm ++ " from the following modules:"
             , indent $ paras $ map (line . runModuleName) ms
             ]
+    renderSimpleErrorMessage (ScopeShadowing nm ms) =
+      paras [ line $ "Ignored shadowed definitions are in scope for " ++ nm ++ " from the following modules:"
```
paf31: Does it make sense to say "from the following _open module imports_:" or something along those lines, and then to print the list with the word `import`?

src/Language/PureScript/Sugar/Names/Env.hs
```diff
@@ -41,19 +68,19 @@ data Imports = Imports
   -- |
   -- Local names for types within a module mapped to to their qualified names
   --
-    importedTypes :: M.Map (Qualified (ProperName 'TypeName)) [(Qualified (ProperName 'TypeName), ModuleName)]
+    importedTypes :: M.Map (Qualified (ProperName 'TypeName)) [ImportRecord (ProperName 'TypeName)]
```
paf31: Much nicer :smile:

src/Language/PureScript/Sugar/Names/Imports.hs
```diff
@@ -162,11 +163,13 @@ resolveModuleImport
    . (Applicative m, MonadError MultipleErrors m, MonadWriter MultipleErrors m)
   => Env
   -> Imports
-  -> (ModuleName, [(Maybe SourceSpan, ImportDeclarationType, Maybe ModuleName)])
+  -> (ModuleName, [(Maybe SourceSpan, Maybe ImportDeclarationType, Maybe ModuleName)])
```
paf31: What does the `Maybe` represent here?

src/Language/PureScript/Errors.hs
garyb: Yep, that'd be much better :+1:

paf31:
Did you want to include this in 0.8.1?

src/Language/PureScript/Sugar/Names/Imports.hs
garyb: When `Nothing` it's the current ("self") module temporarily injected into the environment to make local names resolvable.

Later on the `ImportDeclarationType` gets mapped to `ImportProvenance`, and `Nothing -> Local`. I didn't want to add a value to `ImportDeclarationType` that only exists for a moment during name desugaring.

garyb:
That would be good. Almost done with this error message, I've tweaked it a bit further:

```
Compiling A
Compiling Main
Warning found:

  Shadowed definitions are in scope for value thing from the following open imports:

    import A

  These will be ignored and the local declaration will be used.


See https://github.com/purescript/purescript/wiki/Error-Code-ScopeShadowing for more information,
```
```
Compiling B
Compiling A
Compiling Main
Warning found:

  Shadowed definitions are in scope for value thing from the following open imports:

    import B

  These will be ignored and the declaration from A will be used.


See https://github.com/purescript/purescript/wiki/Error-Code-ScopeShadowing for more information,
```

garyb:
Ok, those changes are in. I also turned `ConflictingExports2` into a passing test (`ResolvableScopeConflict3`) rather than just deleting it.

paf31:
:+1: :ship: Looks good!

I'll test and release 0.8.1 after this gets merged if that works?

garyb:
Definitely :tada: 

I'm off for the night, so I'll update the npm release / add Windows binaries in the morning.

garyb:
Actually, I'm going to hang around and make sure it passes ;)

PR #1908, opened 2016-02-29
Fix pretty printer spinning

### LINKED ISSUE #1905: 0.8.1 hangs compiling SlamData
I assume that's the case anyway, I cancelled the build after 10 minutes of near 100% CPU usage.

I'm not sure what the hold up is, as `output` has 779 subdirectories at that point, which is the same as when 0.7.6 succesfully completes.

garyb:
Resolves #1905 

The code is much uglier, but since this keeps exploding and isn't that easy to track down perhaps it's better this way than relying on a catch-all case.


garyb:
#1905 was caused by warning about old style operator sections by the way, as the `BinaryNoParens` case wasn't handled.

paf31:
:+1: Looks good!

PR #1909, opened 2016-02-29
Fix true not being treated as an infallible guard

### LINKED ISSUE #1906: true is not recognized as a catch-all guard value in 0.8.1
It was in 0.8.0. A bit weird, as the code is still there: https://github.com/purescript/purescript/blob/37323734bd07faadd54e70030a63186c24756873/src/Language/PureScript/Linter/Exhaustive.hs#L204

garyb:
Resolves #1906

I have some worries that maybe there are other things not behaving properly now, if we don't have a step-through case for ever `PositionedValue`, but not sure how we can catch them all at this point. Nothing else is obviously broken anyway, I think...

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -201,9 +201,11 @@ isExhaustiveGuard :: Either [(Guard, Expr)] Expr -> Bool
 isExhaustiveGuard (Left gs) = not . null $ filter (\(g, _) -> isOtherwise g) gs
   where
   isOtherwise :: Expr -> Bool
-  isOtherwise (TypedValue _ (BooleanLiteral True) _) = True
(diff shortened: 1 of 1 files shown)


paf31:
:+1: Looks good.

paf31:
I'll tag v0.8.1.1 once these are merged, yes?

garyb:
Yeah I guess so, but still release it as v0.8.1 on `npm`?

hdgarrood:
Can we just call it 0.8.2 please? I'd really rather tags were consistent across different registries. Most registries support something like hackage's deprecated versions, in the event something is broken.

paf31:
I'm fine with 0.8.2.

AppShipIt:
Shouldn't it have been tagged release candidate? If no issues are reported just bump the hard coded version number and release. Since @paf31 is interested in regular releases why not make every other week a release candidate (branch per Git Flow) and the intermittent weeks the official release (merge back and tag on the <q>stable</q> branch) of those candidates. Breaking a patch release doesn't seem like a necessary evil especially approaching greater adoption.

Issue #1862, opened 2016-02-03
psc suggesting to import everything from Prelude


cdepillabout:
When compiling the following file in the linked commit from the purescript-email-validate package, psc suggests that I import everything from Prelude, even though I am not using everything from Prelude.

https://github.com/cdepillabout/purescript-email-validate/blob/4b47b9c05fbaa9596d4139a7c375eff6c89a0b3c/src/Text/Email/Validate.purs#L11

Here is the output of `pulp build`:

```
$ pulp build
* Building project in /home/ec2-user/git/purescript-email-validate
psc: No files found using pattern: src/**/*.js
Warning found:
in module Text.Email.Validate
at /home/ec2-user/git/purescript-email-validate/src/Text/Email/Validate.purs line 11, column 1 - line 13, column 1

  Module Prelude has unspecified imports, consider using the explicit form:

    import Prelude (class Applicative, class Apply, class Bind, class BooleanAlgebra, class Bounded, class BoundedOrd, class Category, class DivisionRing, class Eq, class Functor, class ModuloSemiring, class Monad, class Num, class Ord, class Ring, class Semigroup, class Semigroupoid, class Semiring, class Show, Unit, Ordering(EQ, GT, LT), add, ap, append, apply, asTypeOf, bind, bottom, compare, compose, conj, const, disj, div, eq, flip, id, liftA1, liftM1, map, mod, mul, negate, not, one, otherwise, pure, return, show, sub, top, unit, unsafeCompare, void, zero, (#), ($), (&&), (*), (+), (++), (-), (/), (/=), (<), (<#>), (<$>), (<*>), (<<<), (<=), (<>), (==), (>), (>=), (>>=), (>>>), (||))



See https://github.com/purescript/purescript/wiki/Error-Code-ImplicitImport for more information,
or to contribute content related to this warning.

* Build successful.
```

It should only be suggesting something like this:

```purescript
import Prelude ((==), (<<<), ($), show, map)
```

garyb:
Woah, thanks for the report. I've definitely not seen this before, so the example will be useful to test against too.

telser:
Hey @garyb I've hit this bug using 0.8.1-rc.1 that was published via npm while building slamdata/purescript-jtable master branch. Is this still on track to be fixed? Anything I can do to help?

garyb:
0.8.1 has some more serious problems (#1905) so I imagine we'll be releasing another version soonish. I've not looked into this at all yet though, so don't really know what would be causing it.

[...]

PR #1910, opened 2016-02-29
Fix detection of single open import

garyb:
This is a fix for a secondary issue that arose in the discussion of #1862: importing a module qualified without an explicit import list would be treated as if it was an open import.

Also `hiding` imports weren't being treated as open (oops).

It doesn't deal with #1862 entirely though, it's still possible to get a warning that thinks everything needs re-exporting from `Prelude`.


paf31:
:+1: LGTM

paf31:
So, 0.8.2 then, or 0.8.1.1?

garyb:
0.8.2 is fine by me too :+1:

PR #1913, opened 2016-03-01
Fix indentation bug #1881

### LINKED ISSUE #1881: Empty function's body is parsed without error
```
foo =
bar :: Int
bar = 1
```

DavidLindbom:
This should fix #1881 as `indented` behavior is a bit unclear.

### DIFF
--- a/examples/failing/1881.purs
+++ b/examples/failing/1881.purs
@@ -1,0 +1,6 @@
+-- @shouldFailWith ErrorParsingModule
+module Main where
+
+foo =
+bar :: Int
+bar = 3
--- a/examples/passing/1881.purs
+++ b/examples/passing/1881.purs
@@ -1,0 +1,17 @@
+module Main where
+
+foo = 
+ 1
+
+bar 
+ = 2
+
+baz
+ =
+ 3
+
+qux
+  =
+ 3
+
+main = Control.Monad.Eff.Console.log "Done"
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -93,6 +93,7 @@ parseValueDeclaration = do
   where
   parseValueWithWhereClause :: TokenParser Expr
   parseValueWithWhereClause = do
+    C.indented
     value <- parseValue
     whereClause <- P.optionMaybe $ do
       C.indented


paf31:
:+1: Looks good, thanks!

paf31:
Any thoughts on why CI is breaking?

garyb:
Looking at the log, it's because there seem to be a ton of warnings like `Redundant constraint: Functor m` now. I guess that's new in `ghc-8.0.0.20160229`.

hdgarrood:
Unfortunately `Monad` only became a subclass of `Applicative` in base-4.8.0.0 (that is, GHC 7.10), so we'd have to drop support for GHC 7.6 and 7.8 if we wanted to remove those constraints. Or put CPP everywhere (yuck). Just using `base-compat` isn't enough, as it re-exports class definitions from `base`.

I guess the options are, modify the cabal file to ignore those warnings, or remove them and stop supporting GHC < 7.10. I'd vote to drop support for older GHCs, as I've found these warnings can be very useful.

paf31:
As long as we're using Stack to build (which I am now, for example, for the Mac binaries), then I'm fine with dropping < 7.10. Does the Linux64 binary build use Stack?

paf31:
I guess we can't build with GHC8 using Stack/Stackage right now.

I'd really like to have CI working again, so maybe disabling the GHC8 build makes the most sense for now.

paf31:
Thanks!

paf31:
This actually breaks `purescript-aff` since the `Par` module has some incorrect indentation on line 50. We'll have to remember to fix that when we release this.

AppShipIt:
> This actually breaks `purescript-aff`

Milestone `0.9`?

paf31:
Possibly, yes.

DavidLindbom:
Just to make sure, is it an error in the Par module or in the compiler. If it's in the module I can make a PR now to fix it.

garyb:
Hmm, we try not to release breaking changes in the 0.x.y releases, maybe we'd be better off reverting this until 0.9? I'm not sure.

paf31:
I think `aff` is common enough that we should probably hold this off until 0.9. While it's not technically a breaking change, it is in the sense that we don't have a specification for indentation written down anywhere, so "it is what it does" right now :smile:

paf31:
@DavidLindbom I've reverted this, but if you'd like to reopen it, we can merge it in for 0.9.

Thanks!

DavidLindbom:
I can open it again but I just want to make sure first if this should be the correct behavior? More exact should that line in `purescript-aff`be disallowed or should I try to rework the patch?

paf31:
I think `aff` needs to be changed either way, but it would probably be a major bump for both compiler and the package itself.

garyb:
It wasn't immediately obvious to me what's wrong with the indentation, but I think I see it now:

``` purescript
 instance altPar :: Alt (Par e) where
  alt (Par a1) (Par a2) =
    let maybeKill va ve err =
      do e <- takeVar ve
         if e == 1 then killVar va err else return unit
         putVar ve (e + 1)
    in Par do
      va <- makeVar     -- the `a` value
      ve <- makeVar' 0  -- the error count (starts at 0)
      c1 <- forkAff $ attempt a1 >>= either (maybeKill va ve) (putVar va)
      c2 <- forkAff $ attempt a2 >>= either (maybeKill va ve) (putVar va)
      (takeVar va) `cancelWith` (c1 <> c2)
```

Should it be

``` purescript
    let
      maybeKill va ve err =
        do e <- takeVar ve
           if e == 1 then killVar va err else return unit
           putVar ve (e + 1)
```

instead? I happened to notice we have a `do` with indentation that way in the tests today, so I assume it's `let` related instead?

AppShipIt:
Oh I missed it too that's exactly right! There was a recent issue for that.

paf31:
The first lexeme after `let` defines a new indentation level, and similarly for `do`. The issue is just that  the `do` is to the left of the `let` (or the `maybeKill` is to the right of the `let`, depending on how you look at it).

PR #1914, opened 2016-03-02
Infer types with class constraints

paf31:
For 0.8.3

- [x] Only generalize types for declarations without type annotations
- [x] Check all unknowns in unsolved constraints will be generalized
- [x] Basic context simplification
- [x] Tests

### DIFF
--- a/examples/passing/ContextSimplification.purs
+++ b/examples/passing/ContextSimplification.purs
@@ -1,0 +1,9 @@
+module Main where
+
+import Prelude
+
+-- Here, we should simplify the context so that only one Eq
+-- constraint is added.
+usesEqTwice x = if x == x then x == x else false
+
+main = Control.Monad.Eff.Console.log "Done"
--- a/examples/passing/ContextSimplification.purs
+++ b/examples/passing/ContextSimplification.purs
@@ -1,9 +1,13 @@
 module Main where
 
 import Prelude
+import Control.Monad.Eff.Console
 
--- Here, we should simplify the context so that only one Eq
+shout = log <<< (<> "!") <<< show
+
+-- Here, we should simplify the context so that only one Show
 -- constraint is added.
-usesEqTwice x = if x == x then x == x else false
+usesShowTwice true = shout
+usesShowTwice false = print
 
-main = Control.Monad.Eff.Console.log "Done"
+main = usesShowTwice true "Done"
--- a/examples/failing/ConstraintInference.purs
+++ b/examples/failing/ConstraintInference.purs
@@ -1,0 +1,10 @@
+-- @shouldFailWith NoInstanceFound
+
+module Main where
+
+import Prelude
+
+spin :: forall a b. a -> b
+spin x = spin x
+
+test = show <<< spin
--- a/examples/passing/ConstraintInference.purs
+++ b/examples/passing/ConstraintInference.purs
@@ -1,0 +1,7 @@
+module Main where
+
+import Prelude
+
+shout = Control.Monad.Eff.Console.log <<< (<> "!") <<< show
+
+main = shout "Done"
--- a/examples/passing/MonadState.purs
+++ b/examples/passing/MonadState.purs
@@ -1,12 +1,16 @@
 module Main where
 
 import Prelude
+import Control.Monad.Eff.Console
 
 data Tuple a b = Tuple a b
 
-class MonadState s m where
+instance showTuple :: (Show a, Show b) => Show (Tuple a b) where
+  show (Tuple a b) = "(" <> show a <> ", " <> show b <> ")"
+
+class Monad m <= MonadState s m where
   get :: m s
-  put :: s -> m {}
+  put :: s -> m Unit
 
 data State s a = State (s -> Tuple s a)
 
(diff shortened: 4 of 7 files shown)


garyb:
:dancers:

kRITZCREEK:
This gon' be guuud :ice_cream:

jdegoes:
:tada:

paf31:
What are people's thoughts on overloaded literals now?

garyb:
What did you have in mind, numbers/integers I guess, anything else?

I'm torn. They're definitely rather handy sometimes, but they can make for some pretty confusing errors.

paf31:
Yeah, I'm not sure either. If we did, I'd like to do everything: strings, numbers, integers, booleans and maybe even arrays.

jdegoes:
> What are people's thoughts on overloaded literals now?

:+1: It's the PureScript way, which is nearly 100% rebindable syntax out of the box (literals are, of course, a form of syntax).

Good point on the confusing errors, of course.

paf31:
The other downside is that there is no good place for the overloading class/function to go. Literals are very useful in the no-prelude use case.

garyb:
> It's the PureScript way, which is nearly 100% rebindable syntax out of the box (literals are, of course, a form of syntax).

That's true. I was rather enthusiastic about this originally, as having to use `.0` for numbers seemed like a kludge, but I've gotten used to it.

@paf31 Maybe they're `Prim` classes with instances for the `Prim` types?

paf31:
I think the way to go would be to introduce new things which look like type classes but whose dictionaries are just functions. That way, we can have a special case in the constraint solver and we don't need lookup functions in the generated code for `Prim`. If/when constraint kinds lands, this sort of thing would be implementable as a foreign constraint: `foreign import data IsString :: * -> Constraint` etc.

paf31:
The issue is that those instances have to go somewhere and wherever they go, they generate code, so they can't go in `Prim`, and they'd be orphans anywhere else.

paf31:
We can always bake the `Prim` instances into the solver though.

garyb:
Hmm, good points. Is this something that needs figuring out now, or just something that occurred to you while working on constraint inference?

paf31:
No, if we did it, I'd probably put it into 0.9, since it's technically breaking for some cases where you'd need a type signature.

rgrempel:
On the general question of overloaded literals, I find overloaded array literals to be particularly interesting. Essentially, it would seem nice to have an easy literal syntax for the different array-ish types, e.g. `Prim.Array`, `Data.List`, `Data.Sequence`.

Though, I suppose what I do now for `Data.List` is usually something like:

```purescript
("some" : "things" : "in" : "a" : "list" : Nil)
```

... which isn't so much worse than:

```purescript
["some", "things", "in", "a", "list"]
```

I am also intrigued by the idea of overloading integer literals ... just as another data point, I have a 53-bit integer type that might benefit from that, if I'm understanding it correctly (i.e. that one could write integer-like literals for `Int53` somehow):

https://github.com/rgrempel/purescript-elm/blob/master/src/Elm/Int53.purs

paf31:
@rgrempel Yes, exactly, we'd have a class with something like your `fromInt` which you'd write an instance for.

hdgarrood:
The way GHC does OverloadedLists, as far as I can tell, is to desugar `[a,b,c]` to `fromList [a,b,c]`, and that's as far as the language goes. Then, there's an `IsList` class in `base` providing `fromList` with the sensible instances. In practice, I think almost everyone uses `IsList`, but there's nothing stopping you from using your own.

If you don't want to use the Prelude, then you can just define `fromString`, `fromArray`, and so on as `id` (or anything else). So if we were going to do this, I think it would be ok to put the class in Prelude. In fact I'd encourage it; the less code there is in the compiler, the better, I think.

However, I'm not very keen on overloading everything. Numbers and ints I'd definitely not want to be overloaded, as 99.9% of the time, I really do want an `Int` or a `Number`. For anything fancy like a `BigInt`, `Rational`, or `Complex`, it's really not very much effort to use a function. I also like that I can immediately tell which type out of `Int` or `Number` I'm getting without needing to look at the context.

I think most of the String use cases are better addressed with some form of quasiquoting. If you're using a type `a`, which has an `IsString` instance but which isn't a `String`, then in most cases, there isn't a total function `String -> a`. I think you really want a `fromString :: String -> Either error a` which then fails at compile time if the result is `Left`. For instance: URLs, path pieces, package names, file paths like purescript-pathy, etc. IMO, Haskell only needs OverloadedStrings because of the uselessness of `String`, which is a problem that we've managed to side-step.

If we overloaded arrays, then we'd probably have to overload array patterns too, right? If we didn't, I think it would be a bit asymmetrical. If we did, I guess we'd have to have the `IsArray` class take a `* -> *` parameter, and in addition to `fromArray :: Array a -> array a` have a `toArray :: array a -> Array a`, so that

```purescript
f [x,y,z] = x + y + z
```

becomes

```purescript
f (toArray -> [x,y,z]) = x + y + z
```.

So I think we'd need view patterns too. I guess it's not a big deal if there were some period where we have overloaded literals but not patterns, though.

Also, this would break the expectation that pattern matching is O(1). `toArray` is probably quite costly for some larger types, and if we had no way to check the length before converting to Array, that could be quite costly.

**Comment by hdgarrood** [comment]
Also, I view the current system of rebindable syntax as an implementation detail that we leverage in order to have the compiler know as little as possible about things like `Prelude.Monad`, but not something that should really be used. If I were doing a code review where someone had used rebindable syntax to use a do block as a semigroup, or something like that, then I'd ask they change it.

**Comment by paf31** [comment]
I think you're half right about overloaded do notation, but there are some useful cases IMO, like indexed monads.

I agree on the other points. I'm generally not sure if the overloading is worth the error message cost and codegen/optimizer overhead.

**Comment by rgrempel** [comment]
If an overloaded array literal means that we'd actually construct a `Prim.Array` at run-time and the compiler would just magically insert something like `fromArray`, then I'd be much, much less excited about it, since I don't really mind typing `fromArray` if that's what I want. And, if there's going to be an intermediate `Prim.Array` at run-time, I think it's better to be explicit about it.

What I had been sort-of hoping for (without really thinking about whether it was reasonable), was an overloaded syntax that would actually construct the final type without involving `Prim.Array` at all. So, in a way, it would be `cons` that is overloaded (though I may be expressing that in the wrong way -- perhaps you see what I mean).

**Comment by paf31** [comment]
Update: this is working now, I "just" need to figure out context simplification, which I hope to tackle this weekend.

**Comment by paf31** [comment]
This is ready for review and testing.

File: examples/passing/ContextSimplification.purs [hunk 1]
```diff
@@ -0,0 +1,13 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console
+
+shout = log <<< (<> "!") <<< show
+
+-- Here, we should simplify the context so that only one Show
+-- constraint is added.
```
hdgarrood: Does this example test that the context is simplified - would it cause a type error if two `Show` constraints were added?

garyb:
I had to rebase to get my literals build to pass, as without that they still test with the old `.travis.yml`. I'll review this soon by the way, sorry about the delay!

paf31:
The latest Travis build succeeded, but didn't show up here for some reason.

examples/passing/ContextSimplification.purs
paf31: No, it only tests that _if_ simplification happens, then the generated code works. This is less trivial than it sounds since simplification involves moving type class dictionaries around in the generated code.

hdgarrood: Ah, makes sense. Thanks!

PR #1917, opened 2016-03-03
Use generic Literal in the AST

garyb:
This is something I noticed when working on the pretty printing the other day - I figured the `CoreFn` handling of this was a bit nicer, and actually since `Literal` was already parametric it means it was already suited to be used in either place.

### DIFF
--- a/psci/PSCi.hs
+++ b/psci/PSCi.hs
@@ -214,7 +214,7 @@ handleDecls :: [P.Declaration] -> PSCI ()
 handleDecls ds = do
   st <- PSCI $ lift get
   let st' = updateLets ds st
-  let m = createTemporaryModule False st' (P.ObjectLiteral [])
+  let m = createTemporaryModule False st' (P.Literal (P.ObjectLiteral []))
   e <- psciIO . runMake $ make st' [m]
   case e of
     Left err -> PSCI $ printErrors err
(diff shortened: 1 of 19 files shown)


paf31:
:+1: Nice!

AppShipIt:
Your productivity is unreal 😳

garyb:
@AppShipIt Ah I don't know about that, I'm not as productive as I'd like :wink: - it's pretty easy to do this kind of thing when you're familiar with most of the compiler.

@paf31 what do we want to do about these GHC 8 warnings?

AppShipIt:
> what do we want to do about these GHC 8 warnings?

@hdgarrood gave his feel in https://github.com/purescript/purescript/pull/1913#issuecomment-191229068 too. What are potential downsides of dropping `< 7.10`?

hdgarrood:
The only downside I can think of is that not as many people will be able to compile `psc` etc from source. I hope that we can get tooling to a stage where the majority of people don't feel the need to do this, though.

garyb:
I'd be surprised if many people build from source as it is.

hdgarrood:
I use `stack` to manage mine at the moment, and I think a few others do too. It's marginally easier than downloading a binary bundle, extracting, and copying onto my PATH. `stack` does take care of installing the right GHC for you anyway, though. I don't really like the npm installer because the wrapper script it uses has caused problems with psci.

Issue #1838, opened 2016-01-20
Better errors for unrecognised command line options


hdgarrood:
At the moment, if you pass a command line option that `psc` (or any of the other tools) do not recognise, it won't tell you which option wasn't recognised. This can be particularly irritating when you're using tools that invoke `psc`, because it's not always clear what the actual `psc` invokation was. Can `psc` and the other compiler command line tools say which options were unrecognised, if any? 

Currently, it looks like this:
```
$ psc --lol
Usage: psc [FILE] [-f|--ffi ARG] [-o|--output ARG] [--no-tco] [--no-magic-do]
           [--no-opts] [-v|--verbose-errors] [-c|--comments]
           [-r|--require-path ARG] [-p|--no-prefix]
```

Ideally, I think it would look like this:
```
$ psc --lol
Usage: psc [FILE] [-f|--ffi ARG] [-o|--output ARG] [--no-tco] [--no-magic-do]
           [--no-opts] [-v|--verbose-errors] [-c|--comments]
           [-r|--require-path ARG] [-p|--no-prefix]
psc: Unrecognised flag: --lol
```

One example of how this can cause pain is https://github.com/bodil/pulp/pull/163#issuecomment-172675944

garyb:
:+1:

stevejb71:
I was looking for an easy issue to work on. This one seems to work already in the from-source build:

psc --lol
Invalid option `--lol'

Usage: psc [FILE] [-f|--ffi ARG] [-o|--output ARG] [--no-tco] [--no-magic-do]
           [--no-opts] [-v|--verbose-errors] [-c|--comments]
           [-r|--require-path ARG] [--source-maps] [-p|--no-prefix]
           [--json-errors]

hdgarrood:
@stevejb71 Ah, nice! Same for me on `master`. Seems this is implemented in `optparse-applicative` as of 0.12.1.0: http://hackage.haskell.org/package/optparse-applicative-0.12.1.0/changelog.

Perhaps we should require optparse-applicative >= 0.12.1.0 in purescript.cabal?

paf31:
A version bump sounds good.

PR #1918, opened 2016-03-03
Upgrade optparse-applicative to >= 0.12.1 for all executables.

stevejb71:
The upgraded library fixes issue #1838 - better errors for unrecognised command line options.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -228,7 +228,7 @@ library
 executable psc
     build-depends: base >=4 && <5, base-compat >=0.6.0,
                    containers -any, directory -any, filepath -any,
-                   mtl -any, optparse-applicative >= 0.10.0, parsec -any, purescript -any,
+                   mtl -any, optparse-applicative >= 0.12.1, parsec -any, purescript -any,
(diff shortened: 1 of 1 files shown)


paf31:
Thanks! I'll merge this once I figure out the CI issues.

paf31:
Looks like stack isn't happy on LTS 3.22. Perhaps we should just provide an upper bound instead.

hdgarrood:
I think this is because lts 3.22 has optparse-applicative 0.11.0.2, so we'll have to drop support for lts-3 if we want to ensure we get optparse-applicative >= 0.12.

I don't think this is a huge deal, though. lts-3 also has purescript 0.7, which isn't supported either, so if people are using purescript with stack and lts-3, they will probably have added purescript to extra-deps already. If we go ahead with this, they'll just have to add optparse-applicative to extra-deps too (I think).

What would adding an upper bound achieve?

hdgarrood:
I think this just needs rebasing now.

stevejb71:
Rebase done.

paf31:
Thanks!

Issue #1030, opened 2015-04-05
Can't override bindings in psci


stevejb71:
In PSCI, I'd prefer the below to result in x being equal to 4:
> let x = 3
> let x = 4

but instead...

> Error in module $PSCI:
> Error at  line 1, column 5 - line 1, column 9:
>   Duplicate value declaration for x
> See https://github.com/purescript/purescript/wiki/Error-Code-DuplicateValueDeclaration for more information, or to contribute content related to this error.

sgronblo:
I agree with this. By virtue of GHCI also seeming to allow it.

rndnoise:
This is tricky when other things depend on the old value. For instance, in GHCi:

```haskell
λ: let x = 10
λ: let y = x + 20
λ: let x = "message"
λ: y
30
λ: x
"message"
```

If PSCi simply replaced the old let binding with a new one, then redefining `x` will cause `y` to fail type checking. Even if the type of `x` wasn't changed, changing its value would (perhaps unexpectedly) alter what `x` refers to in `y` after it was defined. The same issue would arise if there was an `:undef`, since other things might depend on the thing being removed.

If those caveats don't need to be addressed, I'm happy to take a shot at implementing this. I can't think of a nice way to emulate GHCi.

hdgarrood:
Have you considered re-using the existing machinery for the browser repl backend? I think that might allow us to provide the ghci behaviour (which I think is preferable).

PR #1920, opened 2016-03-04
Unset bindings in psci

stevejb71:
This is intended to address #1030 

Seems to work and is simple, so hopefully not too hard to review. 

This is different to ghci, where the last result would be 15:

```
> import Prelude 
> let x = 10
> let f a = x + a
> f 5
15

> let x = 20
> f 5
25
```

### DIFF
--- a/psci/PSCi/Types.hs
+++ b/psci/PSCi/Types.hs
@@ -91,11 +91,11 @@ psciLoadedModules = collect . Map.toList . _psciLoadedModules
 type ImportedModule = (P.ModuleName, P.ImportDeclarationType, Maybe P.ModuleName)
 
 psciImportedModuleNames :: PSCiState -> [P.ModuleName]
-psciImportedModuleNames (PSCiState{psciImportedModules = is}) =
+psciImportedModuleNames PSCiState{psciImportedModules = is} =
   map (\(mn, _, _) -> mn) is
 
 allImportsOf :: P.Module -> PSCiState -> [ImportedModule]
-allImportsOf m (PSCiState{psciImportedModules = is}) =
+allImportsOf m PSCiState{psciImportedModules = is} =
   filter isImportOfThis is
   where
   name = P.getModuleName m
--- a/psci/PSCi/Types.hs
+++ b/psci/PSCi/Types.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE DataKinds #-}
+
 -----------------------------------------------------------------------------
 --
 -- Module      :  Types
@@ -22,7 +24,8 @@ import Control.Arrow (second)
 import Data.Map (Map)
 import qualified Data.Map as Map
 import qualified Language.PureScript as P
-import Data.Maybe (isNothing)
+import qualified Data.List as L
+import qualified Data.Bifunctor as B
 
 data PSCiOptions = PSCiOptions
   { psciMultiLineMode     :: Bool
@@ -41,26 +44,29 @@ data PSCiState = PSCiState
   { psciImportedModules     :: [ImportedModule]
   , _psciLoadedModules      :: Map FilePath [P.Module]
   , psciForeignFiles        :: Map P.ModuleName FilePath
-  , psciLetBindings         :: [P.Declaration]
+  , psciValueDeclarations   :: Map P.Ident P.Declaration
+  , psciDeclarations        :: [P.Declaration]
   , psciNodeFlags           :: [String]
   }
 
 initialPSCiState :: PSCiState
 initialPSCiState =
-  PSCiState [] Map.empty Map.empty [] []
+  PSCiState [] Map.empty Map.empty Map.empty [] []
 
 mkPSCiState :: [ImportedModule]
             -> [(FilePath, P.Module)]
             -> Map P.ModuleName FilePath
+            -> Map P.Ident P.Declaration
             -> [P.Declaration]
             -> [String]
             -> PSCiState
-mkPSCiState imported loaded foreigns lets nodeFlags =
+mkPSCiState imported loaded foreigns valueBindings decls nodeFlags =
   (initialPSCiState
     |> each imported updateImportedModules
     |> updateModules loaded)
     { psciForeignFiles = foreigns
-    , psciLetBindings = lets
+    , psciValueDeclarations = valueBindings
+    , psciDeclarations = decls
     , psciNodeFlags = nodeFlags
     }
   where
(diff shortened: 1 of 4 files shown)


psci/PSCi/Types.hs
```diff
@@ -121,12 +121,17 @@ updateModules modules st =
...
 -- |
 -- Updates the state to have more let bindings.
+-- Replaces ValueDeclarations if they have the same identifier.
 --
 updateLets :: [P.Declaration] -> PSCiState -> PSCiState
-updateLets ds st = st { psciLetBindings = psciLetBindings st ++ ds }
-
+updateLets ds st = st { psciLetBindings = keptBindings ++ ds }
+  where keptBindings = filter shouldKeep (psciLetBindings st)
+        shouldKeep decl = valueDeclarationIdent decl `notElem` map valueDeclarationIdent ds
```
paf31: Why not use a `Map`?

stevejb71: Not sure what to key against. If it were just value declarations I'd use ```Ident```, but ```psciLetBindings```can include data declarations or types as well (it looks misnamed to me).

paf31:
I'd consider using `Either` or creating your own sum type.

stevejb71:
Here's a go at using a Map to join the declarations into ```PSCiState```
Producing keys for all of the ```P.Declaration``` is a little messy and I don't think all of them are used in PSCi (e.g. ```BindingGroupDeclaration```).
So I've tried just putting ValueDeclarations into a Map and having a separate field in ```PSCiState``` for non-```ValueDeclaration```s.

My original commit seems simpler though - what is the benefit of using a Map?

LiamGoodacre:
I'm not happy with your example/suggested behaviour:

~~~purescript
> import Prelude 
> let x = 10
> let f a = x + a
> f 5
15

> let x = 20
> f 5
25
~~~

I feel that `f` should still give `15` - it was defined in terms of `x_0` not `x_1`.  Consider the following potentially confusing result:

~~~purescript
> let a = 10
> let b = a * 10
> let f x = [x, a, b]
> f 0
[0, 10, 100]

> let a = 20
> f 0
[0, 20, 100]
~~~

Here by replacing `a` you've changed `f` but not `b`, even though they were both defined in terms of a previous `a`.

If you want to take advantage of this behaviour you will have to be careful never to be point-free.  If you mix them up then you'll get strange results and have to re-`let` everything anyway.

What are your thoughts on this? @stevejb71 @paf31

AppShipIt:
> What are your thoughts on this?

I agree this would be unpleasantly surprising. I expect new bindings to shadow new uses not pervious ones.

paf31:
I think maybe a warning is a good enough way to handle this situation? Changing the meaning of the inputs to PSCi from a list of declarations in a module to something which tracks dependencies is going to be quite a lot more work.

psci/PSCi/Types.hs
```diff
@@ -118,15 +125,17 @@ updateModules modules st =
...
 --
-updateLets :: [P.Declaration] -> PSCiState -> PSCiState
-updateLets ds st = st { psciLetBindings = psciLetBindings st ++ ds }
-
+updateDeclarationBindings :: [P.Declaration] -> PSCiState -> PSCiState
+updateDeclarationBindings ds st = st { psciValueDeclarations = valueDeclarations `Map.union` psciValueDeclarations st, psciDeclarations = psciDeclarations st ++ otherDeclarations }
+  where (valueDeclarations, otherDeclarations) = B.first (Map.fromList . map (\d -> (identFor d, d))) . L.partition P.isValueDecl $ ds
+        identFor (P.ValueDeclaration i _ _ _) = i
+        identFor (P.PositionedDeclaration _ _ x) = identFor x
+        identFor _ = error "this should not have happened"
```
paf31: Could you provide a better error and use `internalError` here please?

paf31:
One other downside:

```
> import Prelude 
> let x = 10
> let f a = x + a
> f 5
15

> let x = "explode"
```

will now give an error.

hdgarrood:
Especially given this most recent example. I'm tempted to say we should at least investigate how hard it would be to avoid this. Maybe tracking dependencies isn't necessary, maybe we can add numeric suffixes to reused identifiers behind the scenes or something?

AppShipIt:
When I watched the iHaskell talk he described GHCi as if each instruction introduced a new `do` block. Would that be an effective solution? New names would shadow old ones but old ones would continue to be closed over.

paf31:
@AppShipIt I thought about that, but it's not clear how or when to put things in a `let`.

Suppose you define `a` and then `b` and then `c`. `b` and `c` use `a`. Then you redefine `a`. Should the existing `a` be pushed into `let` clauses in `b` and `c` and the new `a` be defined at the top level? Until something is redefined it needs to be available for use at the top level.

An alternative could be to always push dependencies into `let` expressions, duplicating them as many times as is needed. It might result in code explosion though. If we did it the naive way, you'd end up with things like:

```purescript
let a = ...
     b = let a = ...
           in f a
     c = let a = ...
               b = let a = ...
                       in f a
           in g a b
in h a b c
```

psci/PSCi/Types.hs
```diff
@@ -118,15 +125,17 @@ updateModules modules st =
   st { _psciLoadedModules = Map.union (go modules) (_psciLoadedModules st) }
   where
   go = Map.fromListWith (++) . map (second (:[]))
-
```
AppShipIt: I think this deleted line was accidental.

psci/PSCi/Types.hs
AppShipIt: https://github.com/purescript/purescript/blob/279527f200ca5d6c1fe890e09aa86fe9a847c78b/src/Language/PureScript/Sugar/BindingGroups.hs#L160

AppShipIt:
IIUC @stevejb71 changed `let` bindings to a map that overwrites duplicate identifiers? My suggestion seems <q>non-trivial</q> by comparison. I feel the current behavior is preferable until addressed architecturally.

hdgarrood:
> Should the existing a be pushed into let clauses in b and c and the new a be defined at the top level?

That seems reasonable to me, yeah. Is the problem with this approach just that it would be complex?

Alternatively, going for the naive route, maybe we can put up with the code explosion for now, and when we get around to it, implement common subexpression elimination in the compiler?

paf31:
Closing due to inactivity, but we can reopen if the issues above can be resolved.

PR #1925, opened 2016-03-06
Fix #1916

### LINKED ISSUE #1916: Operator sections with unary operators
The following expression:

~~~purescript
(x `op` _)
~~~

desugars to

~~~purescript
op x
~~~

instead of

~~~purescript
(\y -> op x y)
~~~

This permits expressions such as

~~~purescript
(true `not` _)
-- => false
~~~

which I feel is misleading, given that `not` (when special cased to `Boolean`) does not accept a second argument.  In this example I would expect it to desugar to

~~~purescript
(\y -> not true y)
~~~

resulting in the type error:

~~~
﻿﻿﻿  A function of type

    Boolean

    y
~~~

bagl:


LiamGoodacre:
Awesome :+1:

garyb:
I'm not sure this is the ideal fix actually, as it imposes a (minor) performance hit when using operator sections.

paf31:
Perhaps we can use a type annotation of some sort. Or maybe the optimizer will just reduce this for us?

On the other hand, this approach will probably be _faster_ for operators which are inlined by the compiler :smile: 

Also, please update `CONTRIBUTORS.md`.

paf31:
@garyb What do you think? I suspect the additional inlining for primitives might be worth the slight slowdown for bigger operators.

paf31:
> we can eliminate it again later in an optimisation pass even if we're not doing so right now.

[I think we might already be doing this](https://github.com/purescript/purescript/blob/279527f200ca5d6c1fe890e09aa86fe9a847c78b/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs#L40).

paf31:
@garyb I just checked, and you're right, this isn't optimized right now.

I do still think it is the "right way" to fix the issue though, until we can optimize the extra code away. The alternative would be to add a type annotation, but I don't see a way to do that without using wildcards, which would give warnings at compile time, unless we add some sort of support for internal wildcard types.

garyb:
Yeah, I'm cool with doing it this way for now, but perhaps we should open an issue to track the need for the optimisation once we merge it.

paf31:
Thanks!

PR #1932, opened 2016-03-07
catches EOF exceptions thrown in acceptCommand

kRITZCREEK:
This stops the server from crashing if the client terminates the
connection before sending any data. The emacs plugin somehow does this
from time to time and I consider it reasonable to not have the server
die because a client disconnects anyway.

### DIFF
--- a/psc-ide-server/Main.hs
+++ b/psc-ide-server/Main.hs
@@ -11,8 +11,9 @@ import           Prelude.Compat
 
 import           Control.Concurrent                (forkFinally)
 import           Control.Concurrent.STM
-import           Control.Exception                 (bracketOnError)
+import           Control.Exception                 (bracketOnError, catchJust)
 import           Control.Monad
+import           Control.Monad.Error.Class
 import           "monad-logger" Control.Monad.Logger
 import           Control.Monad.Reader
 import           Control.Monad.Trans.Except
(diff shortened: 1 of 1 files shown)


kRITZCREEK:
Closing to trigger rebuild

psc-ide-server/Main.hs
```diff
@@ -101,32 +103,43 @@ startServer port env = withSocketsDo $ do
...
-            Left err -> liftIO $ T.hPutStrLn h (encodeT err)
-        Nothing -> do
-          $(logDebug) ("Parsing the command failed. Command: " <> cmd)
-          liftIO $ do
-            T.hPutStrLn h (encodeT (GeneralError "Error parsing Command."))
-            hFlush stdout
-      liftIO (hClose h)
+      accepted <- runExceptT $ acceptCommand sock
+      case accepted of
+        Left err -> $(logDebug) err
```
paf31: In this case, if the debug flag is not set, won't this error just be ignored? Is that a problem?

kRITZCREEK: That's basically the intention. If this error occurs it means the client has disconnected before he sent a command and so there is no point in trying to send back an error. Instead we just close the socket and wait for a new connection.

paf31:
Thanks!

PR #1933, opened 2016-03-08
Remove GHC 8 build for now

paf31:
Until someone can make the build work with GHC 8, I want the CI build to work, so I'm removing this sub-build.


paf31:
I'll restart all of the open builds.

PR #1938, opened 2016-03-08
Drop support for GHC 7.8 and older

### LINKED ISSUE #1919: Remove support for LTS 3.* and GHC < 7.10

hdgarrood:
Resolves #1919

* Bump minimum `base` version bound in purescript.cabal so that GHC 7.8
  and older get errors at dependency resolution time instead of compile
  time
* Update "tested-with" section in purescript.cabal
* Update minimum supported version to GHC 7.10.1 in INSTALL.md.
* Use --resolver=nightly for the suggested stack install command (the
  latest lts release is often out of date and unsupported)
* Remove the travis builds which use GHC < 7.10.

We will also need to update http://www.purescript.org/download/. I think
that's everything.


hdgarrood:
I can update the download page on the website too but I'd rather wait until after the Hakyll stuff is resolved.

.travis.yml
```diff
@@ -2,19 +2,9 @@ language: c
...
 matrix:
   include:
-    - env: GHCVER=7.8.4 CABALVER=1.22 COVERAGE=true
-      compiler: ": #GHC 7.8.4 - tests"
-      # ^ HACK before https://github.com/travis-ci/travis-ci/issues/4393 is resolved
-      addons: {apt: {packages: [cabal-install-1.22,ghc-7.8.4,happy-1.19.5,alex-3.1.4], sources: [hvr-ghc]}}
-    - env: GHCVER=7.6.3 CABALVER=1.22
-      compiler: ": #GHC 7.6.3"
-      addons: {apt: {packages: [cabal-install-1.22,ghc-7.6.3,happy-1.19.5,alex-3.1.4], sources: [hvr-ghc]}}
     - env: GHCVER=7.10.1 CABALVER=1.22 DEPLOY=yes
```
paf31: Is there any reason why we deploy the 7.10.1 version? Should we switch to the most recent version?

paf31:
Looks good to me. @phadej would you mind taking a look too?

Another possibility is to switch to only supporting Stack/Stackage based builds. I'd be fine with that, and it would certainly simplify things.

.travis.yml
hdgarrood: :+1:

hdgarrood:
I nearly suggested dropping the cabal-install builds for the same reason too. I'll try that.

kRITZCREEK:
grrr... :D All my backporting work on psc-ide...

:+1:

hdgarrood:
Actually I'm inclined to leave this as this is instead of switching `cabal-install` for `stack`, at least for now, because trying to switch from `cabal-install` to `stack` seems nontrivial. In particular, we'd need to figure out how to test the source distributions, to make sure that no modules or files were forgotten in `purescript.cabal`.

paf31:
Makes sense.

phadej:
LGTM.

Dropping `cabal` support (either overall or on CI) is a bad idea, because `purescript` is used as a library too. Otherwise it's arguable, whether it should be on Hackage at all (except that it have to be there to be on Stackage, which is useful for QA-reasons).

But, I'd like to add `stack` based CI builds (or at least one) *in addition* to cabal builds, because `stack` is useful in development. And might make sense to make bindists with `stack`, as then they are much more reproducible. Yet the release process is up to you to decide :)

paf31:
Thanks!

PR #1942, opened 2016-03-10
lots of formatting.

kRITZCREEK:
The "real" change in this commit is getting rid of the partial accessor
functions that @hdgarrood  pointed out in PureScript.Ide.Types.


paf31:
:+1: Looks good. Will merge after CI is done.

PR #1943, opened 2016-03-10
[WIP] removes superfluous constraints.

kRITZCREEK:
Since GHC 7.10 Monad is a superclass of Applicative.

Should I also check if we need Prelude.Compat anymore?

### DIFF
--- a/psc-bundle/Main.hs
+++ b/psc-bundle/Main.hs
@@ -51,7 +51,7 @@ data Options = Options
   } deriving Show
 
 -- | Given a filename, assuming it is in the correct place on disk, infer a ModuleIdentifier.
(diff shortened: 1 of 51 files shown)


paf31:
Wonderful. Does this allow us to add GHC 8 back to the CI build?

kRITZCREEK:
It'll probably still fail. I literally just grepped for lines with `Applicative/Functor m` in them :D I haven't tried to set up GHC 8 locally, which would probably be better for finding the rest of the superfluous constraints.

paf31:
Well, if you're feeling enthusiastic, the last CI build report from GHC 8 has a complete list of redundant constraints :smile: 

But I'll get to this soon in any case, so no worries.

paf31:
Let's deal with GHC8 separately. Merging, thanks!

PR #1949, opened 2016-03-13
Fix #1889, improve performance by avoiding whitespace operations on large strings

### LINKED ISSUE #1889: Performance of 0.8 compared to 0.7.6

paf31:
Before, we were combining warnings into one huge string, and then carving out the excess whitespace. Now we carve and then combine, which halves the Halogen build time (was 16s, now 8s) on my machine.

It's not as fast as #1948, which gets down to about 6s, but we keep all of the warning output.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -1181,31 +1181,32 @@ prettyPrintRef (PositionedDeclarationRef _ _ ref) = prettyPrintExport ref
 -- Pretty print multiple errors
 --
 prettyPrintMultipleErrors :: Bool -> MultipleErrors -> String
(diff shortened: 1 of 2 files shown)


garyb:
Nice. Regardless of whether we do something like #1948 this seems worthwhile then :+1:

PR #1950, opened 2016-03-13
Try Trusty dist for Travis

garyb:
Since this does great things for the build time of PS libraries, I wonder if it will help with the compiler build too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: c
-sudo: false
+dist: trusty
+sudo: required
 matrix:
   include:
     - env: GHCVER=7.10.3 CABALVER=1.22 STACKAGE=lts-5.4 RUNSDISTTESTS=YES


paf31:
:pray:

paf31:
Maybe we should enable parallel builds?

garyb:
Oh yeah, definitely worth trying that too. Do I just need to take the `sed` line out?

garyb:
Nice, looks like it's a decent bit faster with sequential builds even (20ish mins vs 50ish).

paf31:
Are you comparing even after Harry's change to remove old GHCJS versions?

garyb:
I believe so, see the log here: https://travis-ci.org/purescript/purescript/builds

The build for the whitespace changes is what I was comparing it to.

paf31:
Looks good, the individual builds are down to about 7 minutes :smile:

PR #1955, opened 2016-03-13
Expose hiding import suggestion in JSON

nwolverson:
As per the other import suggestions, support editor 'quick-fix'.


paf31:
:+1: Thanks!

nwolverson:
Should have caught that warning, I had some strange build issues at first. Should be fixed.

paf31:
Looks good!

Issue #1714, opened 2015-12-14
psc-publish: warn if you forgot a LICENSE?


hdgarrood:
`psc-publish` could warn or maybe even error if:

* `bower.json` has no LICENSE specified
* LICENSE file does not exist

Maybe we could also start rejecting packages without a license in pursuit too.

paf31:
:+1:

hdgarrood:
Just realised I didn't make this clear before - my vote is for it to be an error.

hdgarrood:
Thinking about this a bit more - I think we should have a whitelist of open source licences, and reject Pursuit uploads which are not licensed with an open source licence.

michaelficarra:
@hdgarrood That's going to be a touchy subject. Do you define open source as OSI-approved, FSF-approved, Copyfree-approved, etc?

hdgarrood:
Definitely nothing anywhere near as strong as FSF approved. I just mean something that we can be sure we won't have any issues displaying documentation for and redistributing (if/when pursuit becomes a registry). Also, it's quite easy to forget to license your code even if you meant to, that's the kind of problem I'm really thinking about.

michaelficarra:
In that case, I think the existence of the LICENSE file is good enough.

hdgarrood:
Yeah - on second thoughts, I agree.

paf31:
One problem is that we don't require a package to be uploaded by the author. The `LICENSE` file might be empty, or say "You are not allowed to redistribute the documentation", at which point we'd pretty much have to delete a package from Pursuit if someone uploaded it :smile:

PR #1962, opened 2016-03-17
Bump dependency on bower-json

hdgarrood:
bower-json-0.8.0 fixes parsing of license, main, and moduleType fields.
In particular, we need the license field for #1714.


paf31:
Thanks!

PR #1963, opened 2016-03-17
Update for names in new prelude

### LINKED ISSUE #1953: Update deriving to work with new Prelude module names

garyb:
Resolves #1953 

This has changes for deriving, but also further updates to the inliner to reflect the new (as of yet unmerged) boolean and number class hierarchies.


hdgarrood:
s/euclidian/euclidean/ ?

garyb:
Uh, whoops.

AppShipIt:
> new (as of yet unmerged) boolean and number class hierarchies.

@garyb Where my I read about this?

hdgarrood:
@AppShipIt https://github.com/purescript/purescript-prelude/issues/61

garyb:
@AppShipIt 

- `HeytingAlgebra <= BooleanAlgebra`: https://github.com/purescript/purescript-prelude/issues/62
- `EuclideanDomain`, etc.: https://github.com/purescript/purescript-prelude/issues/61#issuecomment-196014850

AppShipIt:
Thanks both I don't know why I didn't consider checking PRs there!!

paf31:
Looks good, thanks! We also need to update `$PSCI.Support` for changes in `-console`.

garyb:
Ah good point. I haven't checked what the error in the build is here yet either, will do both things later on today.

garyb:
Ok, these changes are done now. The tests now use the new `Prelude` also, so the deriving tests work, and I had to update some stuff that has changed (no `return`, `(++)`, `print` -> `logShow`, etc).

The `psci` tests fail, but I'm not sure if it's the tests that are the problem, or if it is actually `psci`:

```
### Failure in: completionTests:18
expected: [":type unit"]
 but got: []
### Failure in: completionTests:19
expected: [":type EQ"]
 but got: []
### Failure in: completionTests:20
expected: [":kind Control.Monad.Eff.Pure"]
 but got: []
### Failure in: completionTests:21
expected: [":kind Ordering"]
 but got: []
### Failure in: completionTests:31
expected: ["unit"]
 but got: []
### Failure in: completionTests:33
expected: ["GT"]
 but got: []
Cases: 37  Tried: 37  Errors: 0  Failures: 6
```

Some of these failures seem like they always should have been failures, since they're trying to autocomplete things that have not been imported (`:type unit`, `:type EQ`)? But then some of the others seem like they definitely should work (`:kind Control.Monad.Eff.Pure`) so I don't really know. Maybe it's a mixture of test and `psci` problems?

hdgarrood:
I thought those tests should have imported Prelude open already? So I'd expect them all to work.

garyb:
Maybe I'm misreading, but it looks like the `Prelude` import:

https://github.com/purescript/purescript/blob/7fe6f0696455a622ee13d508bd9a93f67c3f8adc/tests/TestPsci.hs#L95

Is after the `:type` tests that assume it has been imported:

https://github.com/purescript/purescript/blob/7fe6f0696455a622ee13d508bd9a93f67c3f8adc/tests/TestPsci.hs#L79

garyb:
Ah, unless that's just the completions and the import is already done.

In that case, `psci` doesn't understand re-exports?

paf31:
So if you do `import Prelude` in PSCi now, you don't see anything? :worried:

garyb:
I hadn't checked actually. It would appear that `psci` does indeed not understand re-exports - `> :t unit` works, but it won't autocomplete it.

garyb:
`:browse Prelude` is empty too, I assume it should print "everything"?

paf31:
Should we release 0.8.3 without this? If we wait, we'll have to also wait for a decent subset of the existing libraries to be updated too.

garyb:
What do you mean? We can't really update the libraries until this is in, as neither inlining nor deriving will work for the new prelude, and the old prelude cannot be made warning-free for 0.8 without breaking changes.

garyb:
Oh, sorry, I just parsed the sentence the wrong way I think.

I'm not sure how hard this is to fix, I assume it's just reading the exports from the module directly rather than using the desugared scope - but we did some work to make that useful elsewhere, as `psc-publish` uses it too now, so maybe the fix won't be too tricky?

paf31:
I'm just thinking there is enough to make a 0.8.3 release now, and we could make a 0.8.4 release whenever the core libraries come out of RC mode.

garyb:
Hmm, we have a bit of an issue here actually... Deriving (generic or ord/eq) will no longer work with the pre-1.0 libraries if we release this now. I guess that means we'll have to release this as 0.9. 

Perhaps we could do that soon(ish) and have a truncated 0.8 cycle? It'd be good to get at least 0.8.4 out with the non-broken `psc-bundle`, etc. before 0.9, but maybe that should be the last one.

paf31:
That makes sense. I'd like to see if I can fix up let generalization before 0.8.4, since it's related to a potentially rather bad soundness bug in 0.8.3. Also, if we're releasing 0.9.0, we should get busy breaking all the stuff we said we were going to break :smile:

AppShipIt:
Agreed on 0.9.0 if it's okay because mine are breaking too.

PR #1965, opened 2016-03-20
escapes regexy chars when using the flexMatcher

### LINKED ISSUE #1960: [psc-ide] Escape Regex Chars when flexmatching
psc-ide-server will crash if you pass a search string that contains "regexy" characters. Escape user input before turning it into the flexmatch regex.

kRITZCREEK:
This stops the server from crashing on queries like "<*"

fixes #1960

### DIFF
--- a/src/Language/PureScript/Ide/Matcher.hs
+++ b/src/Language/PureScript/Ide/Matcher.hs
@@ -86,15 +86,25 @@ flexRate pattern c@(Completion (_,ident,_)) = do
 -- the matchas a (start, length) tuple if there's a match.
 -- If match fails then it would be (-1,0)
 flexScore :: Text -> DeclIdent -> Maybe Double
-flexScore "" _ = Nothing
 flexScore pat str =
-  case TE.encodeUtf8 str =~ TE.encodeUtf8 pat' :: (Int, Int) of
-    (-1,0) -> Nothing
-    (start,len) -> Just $ calcScore start (start + len)
-  where
-    Just (first,pattern) = T.uncons pat
-    -- This just interleaves the search string with .*
-    -- abcd -> a.*b.*c.*d
-    pat' = first `T.cons` T.concatMap (T.snoc ".*") pattern
-    calcScore start end =
-      100.0 / fromIntegral ((1 + start) * (end - start + 1))
+  case T.uncons pat of
+    Nothing -> Nothing
+    Just (first, pattern) ->
+      case TE.encodeUtf8 str =~ TE.encodeUtf8 pat' :: (Int, Int) of
+        (-1,0) -> Nothing
+        (start,len) -> Just $ calcScore start (start + len)
+      where
+        escapedPattern :: [Text]
+        escapedPattern = map escape (T.unpack pattern)
+
+        -- escape prepends a backslash to "regexy" characters to prevent the
+        -- matcher from crashing when trying to build the regex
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Ide/Matcher.hs
```diff
@@ -86,15 +86,25 @@ flexRate pattern c@(Completion (_,ident,_)) = do
...
+        (-1,0) -> Nothing
+        (start,len) -> Just $ calcScore start (start + len)
+      where
+        escapedPattern :: [Text]
+        escapedPattern = map escape (T.unpack pattern)
+
+        -- escape wraps "regexy" characters in brackets to prevent the matcher
+        -- from crashing when trying to build the regex
+        escape :: Char -> Text
+        escape c = if c `elem` ("*.?$|\\" :: String)
```
michaelficarra: `^`, `(`, `)`, `{`, `}`, `[`, and `]` are all also meaningful regex characters.

src/Language/PureScript/Ide/Matcher.hs
```diff
@@ -86,15 +86,25 @@ flexRate pattern c@(Completion (_,ident,_)) = do
...
+                   then "[" <> T.singleton c <> "]"
```
michaelficarra: You need to backslash escape these, since `^`, `]`, and `\` are meaningful inside a character class.

AppShipIt: Yep `[\^$.|?*+(){}`) ([reference](http://www.regular-expressions.info/refcharacters.html)).

kRITZCREEK: Does this also count for regex-tdfa? I don't think it implements the "real" perl regex spec.

AppShipIt: I'll look into it but IME using several implementations (mostly PCRE flavors) you can't escape too much.

kRITZCREEK:
@michaelficarra did I get all of them? If yes i'll squash this.

michaelficarra:
@kRITZCREEK Looks like it.

kRITZCREEK:
Thanks! If we find more we can still add them. This is definitely an improvement over the status quo.

src/Language/PureScript/Ide/Matcher.hs
```diff
@@ -86,15 +86,25 @@ flexRate pattern c@(Completion (_,ident,_)) = do
...
+        (-1,0) -> Nothing
+        (start,len) -> Just $ calcScore start (start + len)
+      where
+        escapedPattern :: [Text]
+        escapedPattern = map escape (T.unpack pattern)
+
+        -- escape prepends a backslash to "regexy" characters to prevent the
+        -- matcher from crashing when trying to build the regex
+        escape :: Char -> Text
+        escape c = if c `elem` ("[\\^$.|?*+(){}" :: String)
```
michaelficarra: What's with the explicit type signature here? Is that necessary?

garyb: `OverloadedStrings` is enabled in here I think, so probably.

kRITZCREEK: ^ Yep, that's why

michaelficarra:
LGTM.

paf31:
Thanks!

PR #1967, opened 2016-03-21
[psc-ide] - Import helper commands

kRITZCREEK:
## New set of commands under: `import`
For now all of the import related commands work with a file on the filesystem.

You can specify it with the `file` parameter.

If you supply the optional `outfile` parameter, the output will be written to that file, and an info message will be returned from the client.

If you don't supply `outfile`, the server responds with a list of strings which, when inserted into a file linewise create the module with the applied changes.

#### Subcommand `addImplicitImport`

This command just adds an unqualified import for the given modulename.

Arguments:
- `moduleName :: String`

Commandexample:
https://gist.github.com/kRITZCREEK/25787ed4f753da58a918#file-addimplicitimport-json

#### Subcommand `addImport`

This command takes an identifier and searches the currently loaded modules for it. If it finds no matches it responds with an Error. If it finds exactly one match it adds the import and returns. If it finds more than one match it responds with a list of the found matches as completions like the complete command.
You can also supply a list of filters like the ones for completion. This way you can narrow down the search to a certain module and resolve the case in which more then one match was found.

Arguments:
- `moduleName :: String`
- `filters :: [Filter]`

Commandexample:
https://gist.github.com/kRITZCREEK/25787ed4f753da58a918#file-addimport-json

## Also in this PR: Integration Tests for psc-ide
Because the import command will modify files on the filesystem we need a way to test these files.

The new module `Language.PureScript.Ide.Integration` inside the `tests` folder provides a simple API to manage a server process, compile an example project inside `tests/support/psc-ide` and query the server.


kRITZCREEK:
The changes inside this PR are made up of two different topics:

1. A haskell client to spawn a psc-ide-server and send commands to it. This then gets used to write integration tests. I'm pretty happy with how this turned out except for the repetitive addtions to travis's `PATH` inside .travis.yml. Maybe @phadej could suggest a better way to proceed here?

2. The import functionality:

It's currently possible to tell psc-ide to complete identifiers that are "values" in the sense that they can be imported "just" by name.

`import Prelude (bind)`

This breaks down when it comes to importing typeclasses and dataconstructors. 

`import Prelude (class Show)`
`import Data.Maybe (Maybe(Just, Nothing))`

The information necessary to figure out what kind of identifier we are looking at is not present inside the current psc-ide intermediate format. Because I want to keep the psc-ide internal declaration format separate from the externs file format, I'll have to work on changing the `ExternDecl` datatype.

I suggest we merge the somewhat crude functionality inside this PR and I work on changing the internal format inside a new PR, as that will touch most of the code inside psc-ide.

kRITZCREEK:
No idea how the last commit broke the build but it happens during the `./travis/test-install.sh`  script. I'll take a look tomorrow.

kRITZCREEK:
While I was at it, I added module headers to all of the psc-ide files and added some documentation here or there.

Travis seems to be working fine after I added the .purs sources in the support folder to the cabal file.
 
I'd say this is ready for review.

kRITZCREEK:
Just thought I'd leave this here ;)

FrigoEU:
That's very cool. In the gif above, do you have emacs set up to immediately reload files when they change on the file system? Since as I understand these new commands impact files on the filesystem directly instead of the buffer you're currently working on.

kRITZCREEK:
@FrigoEU I've added the command to reload the file from disk to the "[success callback](https://github.com/kRITZCREEK/psc-ide-emacs/commit/c4ea8aeed99c58d580eb5fcc5217ea850fff327d)" (Actually it's all synchronous)

src/Language/PureScript/Ide/CaseSplit.hs
```diff
@@ -1,3 +1,17 @@
+-----------------------------------------------------------------------------
+--
+-- Module      : Language.PureScript.Ide.Imports
```
paf31: Wrong module name here.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -0,0 +1,193 @@
...
+parseImportsFromFile :: (MonadIO m, MonadError PscIdeError m) =>
+                        FilePath -> m ([Text], [Import], [Text])
+parseImportsFromFile fp = do
+  file <- liftIO (TIO.readFile fp)
+  pure (sliceImportSection (T.lines file))
+
+sliceImportSection :: [Text] -> ([Text], [Import], [Text])
+sliceImportSection ls =
+  let
+    preImportSection = takeWhile (not . hasImportPrefix) ls
```
paf31: I would consider using `span` here.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -0,0 +1,193 @@
...
+    importSection =
+      takeWhile continuesImport $
+        dropWhile (not . hasImportPrefix) ls
+    postImportSection =
+      dropWhile continuesImport $
+        dropWhile (not . hasImportPrefix) ls
+    hasImportPrefix = T.isPrefixOf "import"
+    continuesImport x = hasImportPrefix x || T.isPrefixOf " " x || x == ""
```
paf31: What is the `isPrefixOf " "` option for here?

How would you handle imports preceded by whitespace? Or punctuated with comments?

Why not just parse the whole module here?

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -0,0 +1,193 @@
...
+    step acc t = if T.isPrefixOf " " t
+                 then init acc ++ [last acc <> t]
+                 else acc ++ [t]
+  in
+    mapMaybe parseImport concatMultilineImports
+
+parseImport :: Text -> Maybe Import
+parseImport t =
+  case P.lex "<psc-ide>" (T.unpack t)
+       >>= P.runTokenParser "<psc-ide>" P.parseImportDeclaration' of
```
paf31: Since you're parsing anyway, I think it might be best to just parse the whole module. It might be slightly slower, but shouldn't be too much slower, and it'll be less fragile IMO.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -0,0 +1,193 @@
...
+addImplicitImport fp mn = do
+  (pre, imports, post) <- parseImportsFromFile fp
+  let newImportSection = addImplicitImport' imports mn
+  pure $ pre
+    ++ newImportSection
+    ++ post
+
+addImplicitImport' :: [Import] -> P.ModuleName -> [Text]
+addImplicitImport' imports mn =
+  List.sort (map prettyPrintImport' (imports ++ [Import mn P.Implicit Nothing])) ++ [""]
```
paf31: Sorting on strings might not be exactly what you want here, especially if `qualified` imports are still used.

src/Language/PureScript/Ide/Imports.hs
kRITZCREEK: The idea is that you might add an import in the middle of an edit. But if the entire file needs to be in a syntantically valid state that won't work.

src/Language/PureScript/Ide/Reexports.hs
```diff
@@ -1,6 +1,22 @@
+-----------------------------------------------------------------------------
+--
+-- Module      : Language.PureScript.Ide.Reexports
+-- Description : Resolves reexports for psc-ide
+-- Copyright   : Christoph Hegemann 2016
+--               Brian Sermons 2016
```
paf31: It would be nice to add Brian Sermons to `CONTRIBUTORS.md` if possible.

paf31:
This looks generally very good to me, thanks! Please see my comments above.

I think it would be nice to use the same replacements code which @nwolverson wrote.

The nice thing about that, other than that we'd have a single approach, is that it'd be easier to import something without having to save your changes in the editor first.

src/Language/PureScript/Ide/Reexports.hs
kRITZCREEK: We can ping @bsermons and ask him to open a PR against my  kRITZCREEK:import-helper-commands branch

nwolverson:
I don't really follow your point on the replacements code and how that relates to saving changes. If you mean the error suggestion, that requires saving changes, having a compilable file and running psc.

What you do get in that case is (basically) a diff, while here you are getting full file output, and I do see some advantage to the former.

So I just integrated this (mostly) for atom and I think details might help. What I do is save a temp file copy of the current buffer (you probably have changes), run this command on the tmpfile, and then set the new buffer contents to the result. It turns out atom provides a way to update only the diff of old/new buffer contents, which preserves undo behaviour, cursor position and inline text annotations etc. On this point I think it might be more of a struggle to be as seamless in other editors (for me thinking of vscode), and *having explicit diff would be better*, but then I'm sure there are other tools which apply fixes to the entire file and are presumably integrated into editors just fine. Maybe worth looking into the standard way emacs/vim integrate with such tools.

If there's not a rush to merge this for 0.8.3, I should have an initial version of this up this weekend, probably today, it might be useful to see what this looks like.

src/Language/PureScript/Ide/Imports.hs
kRITZCREEK: Not saying this is necessarily a good thing, but since we are pretty printing the parsed imports there won't be no `qualified` keywords left.

kRITZCREEK:
Allright. So I've got typeclass imports to work. Dataconstructors and types always import their corresponding type with open dataconstructors for now.
This is something that I intend to tackle later on but for now this means we don't generate invalid code anymore.

```purescript
-- addImport "Just"
import Data.Maybe(Maybe(..))

-- addImport "Maybe"
import Data.Maybe(Maybe(..))

-- addImport "Unit"
import Prelude (Unit(..))
```

When working on this I realised there is an issue with the separated namespace of types and values which can lead to ambiguity. I discussed the situation on irc with @nwolverson and @damncabbage and we decided to require more information from the user in this rather rare case:

```
3:39 PM <kritzcreek> the separate namespace of types and values means (newtype A = A String) produces two matches. That case is "handleble" by giving the dctor priority and importing as A(..)
3:40 PM <kritzcreek> but if you do this: data A; data B = A Int
3:40 PM <kritzcreek> I cannot decide on the string A alone which of the two you wanted
3:41 PM <damncabbage> Yeah, that's definitely "parse to figure out if type signature or value expression" land.
3:42 PM <kritzcreek> damncabbage: Yeah that would be one way. I could also prompt for which of the two you wanted
3:45 PM <nwolverson> isn't the point that you just asked for 'A', outside the context of which you want and where you intend to put it? so don't you just require more information to be specified?
3:46 PM <kritzcreek> nwolverson: Yes. I need to know if we are in type or value land
3:47 PM <damncabbage> foo = show A   vs  foo x = show (x :: A), for a dumb example
3:47 PM <damncabbage> or foo :: A -> ...
3:48 PM <nwolverson> seems to me if it rarely comes up, you can have an optional specifier, and if it is in fact ambiguous treat it the same as ambiguous identifiers are already, ie choose
3:49 PM <damncabbage> Yeah, prompting is the stick-a-pin-in-it-and-move-on solution. :D
3:49 PM <kritzcreek> nwolverson: If we go down the "choose" route I'd like to unify it with the ambigours identifier approach and add a new Filter that can filter on declarationtype
3:50 PM <nwolverson> kritzcreek: sounds good
3:50 PM <kritzcreek> nwolverson: great! I'll write our decision process up on the issue to see if Phil's got another idea
```

kRITZCREEK:
voila'. That was the last major inconvenience I found. Now I'll finally stop bloating this PR anymore except for bugfixes or suggestions^^.

I can't find anything that doesn't work as expected in my manual testing anymore. Finally "really" ready to review.

FrigoEU:
Was looking into implementing this in psc-ide-vim, but now I'm wondering which modules I should "load". Currently, I'm just loading the module we're in via the "dependencies" keyword. But now I'd make like to go more broad and just load all modules available in the project so I get autocompletion and automatic import statements. Does that sound right?

kRITZCREEK:
@FrigoEU That's why this PR adds a version of the `load` command that loads everything it finds in the output directory. Just call load without a params object.

nwolverson:
@FrigoEU I had the same thought process, I think that would be the best experience

@kRITZCREEK Awesome, must use this.

FrigoEU:
@kRITZCREEK Awesome, didn't notice that. Thanks!

FrigoEU:
Hey, 

I implemented a basic version of this in vim, but got an error. I made this dummy module:

```
module Main where

import Prelude
import Control.Monad.Eff

main = do
  log "Hello sailor!"
```

And got this as output when running addImport on "log":

```
module Main where

import Control.Monad.Eff
import Prelude

main = do
  log "Hello sailor!"
```

Probably just a fluke since other import calls do work fine, but just wanted to let you know.

kRITZCREEK:
@FrigoEU Yeah that case... I haven't decided what I want yet. Does it not add an import at all because the import is already open? I guess replacing the open import with a closed one would lead to breakage...

FrigoEU:
Yup, keeping the open one seems the best approach. Works!

Just one more thing, I suppose it's quite touch the keep the original ordering of modules? I suppose you're doing alphabetical ordering? I think once I start using these automatic imports I won't be bothered too much, but still wanted to ask.

nwolverson:
I like the alphabetic ordering apart from I wish prelude still came at the top! On a serious note it changes the behaviour of open import warnings.

kRITZCREEK:
@nwolverson good point. So I'd presumable need to find the first open import and keep that one at the top?

@FrigoEU keeping the original ordering isn't hard at all. But keeping the blank lines people use to group their imports... these are hard to retain when parsing them. After using the automated import thing for abuot a day I didn't worry about imports anymore so that nagging feeling of *my imports are not pretty* will hopefully go away ^^

kRITZCREEK:
Okay so I don't apply any sorting to the implicit imports now and move them to the top of the imports. That's probably a good thing anyway since these are the only imports that you can not "grep" for.

kRITZCREEK:
restarting travis build

kRITZCREEK:
````
bower ECONFLICT     Unable to find suitable version for purescript-proxy


Done. Your build exited with 0.
```

That doesn't seem right...

kRITZCREEK:
rebased on top of master. Can I do anything to make reviewing/merging this easier?

paf31:
@kRITZCREEK Could you please rebase? Thanks!

paf31:
Oh right, I can rebase for you.

kRITZCREEK:
Thank you! :Party:

PR #1969, opened 2016-03-22
[psc-ide] - [CaseSplit] - respond with an error instead of crashing

kRITZCREEK:
If the user supplies a type that cannot be parsed

### DIFF
--- a/src/Language/PureScript/Ide/CaseSplit.hs
+++ b/src/Language/PureScript/Ide/CaseSplit.hs
@@ -54,7 +54,8 @@ noAnnotations = WildcardAnnotations False
 caseSplit :: (PscIde m, MonadLogger m, MonadError PscIdeError m) =>
              Text -> m [Constructor]
 caseSplit q = do
-  (tc, args) <- splitTypeConstructor (parseType' (T.unpack q))
+  type' <- parseType' (T.unpack q)
+  (tc, args) <- splitTypeConstructor type'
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Ide/CaseSplit.hs
```diff
@@ -119,11 +120,17 @@ addClause s wca =
...
-                     runTokenParser "" (parseType <* P.eof) ts
-               in t
+parseType' :: (MonadError PscIdeError m) =>
+              String -> m Type
+parseType' s = let
+                 parseResult = do
+                   ts <- lex "" s
+                   runTokenParser "" (parseType <* P.eof) ts
+               in
+                 case parseResult of
```
paf31: `case lex "" s >>= runTokenParser "" (parseType <* P.eof) of ...` perhaps?

Also, consider using a filename instead of `""` since the filename shows up in errors now.

kRITZCREEK: This is a type supplied by the user so there is no file involved. The oneliner seems simple enough.

paf31: Sure, I usually use a placeholder in those cases. It's just to help debugging, so I'd normally put something like `<psc-ide>` in there.

Doesn't really matter though.

kRITZCREEK:
I'll squash this if the changes are fine :)

paf31:
Yes, thanks!

kRITZCREEK:
Thank you!

PR #1970, opened 2016-03-23
Better error message for psc-publish tests

kRITZCREEK:
* Uses Either instead of exitFailure inside preparePackage

* Exposes a new unsafePreparePackage which is basically the old
  preparePackage.

* Prints a suggestion to update the git submodules in case of an error
  inside prepare Package

/cc @hdgarrood

### DIFF
--- a/tests/TestDocs.hs
+++ b/tests/TestDocs.hs
@@ -18,6 +18,7 @@ import System.Exit
 import qualified Language.PureScript as P
 import qualified Language.PureScript.Docs as Docs
 import qualified Language.PureScript.Publish as Publish
+import qualified Language.PureScript.Publish.ErrorsWarnings as Publish
 
 import TestUtils
 
@@ -29,16 +30,18 @@ publishOpts = Publish.defaultPublishOptions
   where testVersion = ("v999.0.0", Version [999,0,0] [])
 
 main :: IO ()
-main = do
-  pushd "examples/docs" $ do
-    Docs.Package{..} <- Publish.preparePackage publishOpts
-    forM_ testCases $ \(P.moduleNameFromString -> mn, pragmas) ->
-      let mdl = takeJust ("module not found in docs: " ++ P.runModuleName mn)
-                         (find ((==) mn . Docs.modName) pkgModules)
-      in forM_ pragmas (flip runAssertionIO mdl)
+main = pushd "examples/docs" $ do
+  res <- Publish.preparePackage publishOpts
+  case res of
+    Left e -> Publish.printErrorToStdout e >> exitFailure
+    Right Docs.Package{..} ->
+      forM_ testCases $ \(P.moduleNameFromString -> mn, pragmas) ->
+        let mdl = takeJust ("module not found in docs: " ++ P.runModuleName mn)
+                          (find ((==) mn . Docs.modName) pkgModules)
+        in forM_ pragmas (flip runAssertionIO mdl)
 
 takeJust :: String -> Maybe a -> a
-takeJust msg = maybe (error msg) id
+takeJust msg = fromMaybe (error msg)
 
 data Assertion
   -- | Assert that a particular declaration is documented with the given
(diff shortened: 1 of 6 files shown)


hdgarrood:
:+1: :+1: :+1: but can we use the 'safe' version inside TestDocs too?

kRITZCREEK:
Pushed an update

tests/TestPscPublish.hs
```diff
@@ -73,3 +65,15 @@ testPackage dir = pushd dir $ do
         putStrLn ("psc-publish tests failed on " ++ dir ++ ":")
         putStrLn (show other)
         exitFailure
+
+-- | Fail with a helpful error message when preparePackage fails
+preparePackageError :: Publish.PackageError -> IO ()
+preparePackageError e = do
+  Publish.printError e
```
hdgarrood: Is it possible to inspect this error to make a better guess at whether it's appropriate to print the submodules message?

tests/TestDocs.hs
```diff
@@ -29,13 +30,15 @@ publishOpts = Publish.defaultPublishOptions
...
-  pushd "examples/docs" $ do
-    Docs.Package{..} <- Publish.unsafePreparePackage publishOpts
-    forM_ testCases $ \(P.moduleNameFromString -> mn, pragmas) ->
-      let mdl = takeJust ("module not found in docs: " ++ P.runModuleName mn)
-                         (find ((==) mn . Docs.modName) pkgModules)
-      in forM_ pragmas (flip runAssertionIO mdl)
+main = pushd "examples/docs" $ do
+  res <- Publish.preparePackage publishOpts
+  case res of
+    Left e -> preparePackageError e
```
hdgarrood: I don't think it would be appropriate to print a message about git submodules here. Everything under examples/docs is just a normal file, a part of this repo, isn't it?

tests/TestPscPublish.hs
kRITZCREEK: I had that planned and then forgot about it. Currently it's `UserError (BowerJsonNotFound)` that gets thrown. Is it fragile to check for that?

tests/TestDocs.hs
kRITZCREEK: Oh allright. I didn't realise that they work on different projects.

kRITZCREEK: So what would we do here in case of a failure? Just printing the error message and failing is the same as using the unsafe version

hdgarrood: Oh, of course, you're right. Sorry. I forgot that most of the tests work by just calling `exitFailure` rather than using some testing framework. I think we want to do the same as the unsafe version except print to standard output instead of standard error (because the rest of the tests print to stdout). You might have seen a travis build where the actual error is muddled up with some of the compiler test suite output because they're printing to different streams.

hdgarrood: Sorry, I led you astray a little bit here. Does what I've said make sense?

kRITZCREEK: Sure. We want the testfailures to generally print to stdout then I guess?

hdgarrood: Yeah, I don't think it matters, as long as we choose one and stick to it. I think most of the compiler test suite uses stdout right now.

Maybe later we can get everything to use the framework that the psc-ide uses, then we wouldn't have to worry about issues like this. I'd be in favour of that, at least.

hdgarrood:
:+1: LGTM

kRITZCREEK:
another push

paf31:
Thanks!

Could you please squash? After Travis is done, I'll merge this.

kRITZCREEK:
Cool! squashed

hdgarrood:
Btw, I think it probably is a bit fragile to check for the bower.json not found error specifically, as I think it's then dependent on the order of checks inside `Language.PureScript.Publish`. I'd rather attempt to shell out to git to ask whether the submodule is initialized, I think. But we can (should) do that later.

PR #1972, opened 2016-03-25
Document behaviour of prettyPrintRef

kRITZCREEK:
make prettyPrintRef unwrap a PositionedDeclarationRef and recurse, instead
of pretty printing the ref as an Export.

While I was at it I applied some hlint suggestions

@nwolverson git blame tells me you wrote that function. Was that intended?

EDIT:

Given this code:

```purescript
import Data.Maybe (Maybe(Just))
import Data.Maybe (Maybe(Just))
```

we get the warning: `Duplicate import of Data.Maybe (Maybe)`

After this PR we get: `Duplicate import of Data.Maybe (Maybe(Just))`

There is one more Warning (`RedundantUnqualifiedImport`) where this is happening and those are the only uses I found by grepping the codebase.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -637,7 +637,7 @@ prettyPrintSingleError full level showWiki e = flip evalState defaultUnknownMap 
     renderSimpleErrorMessage (CycleInDeclaration nm) =
       line $ "The value of " ++ showIdent nm ++ " is undefined here, so this reference is not allowed."
     renderSimpleErrorMessage (CycleInModules mns) =
-      paras [ line $ "There is a cycle in module dependencies in these modules: "
+      paras [ line "There is a cycle in module dependencies in these modules: "
             , indent $ paras (map (line . runModuleName) mns)
             ]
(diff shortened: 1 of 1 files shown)


nwolverson:
I willhave moved this, would be @garyb who added duplicate import warnings

garyb:
It is intentional, there some unprintable (`internalError`) data constructors that can appear in the refs. I forget the exact reason why now, but it's something along the lines of us not having the information to be able to give the exact ref as we don't have the data constructor information at the point the warning/error is raised.

garyb:
Here's one example: https://github.com/purescript/purescript/blob/406695a8bb51d9e0e691d9242196fbe3c542f787/src/Language/PureScript/TypeChecker.hs#L405, there may be others.

It's probably fixable too, but would take a bit of figuring out...

kRITZCREEK:
argh... bottom strikes again.

I'll change it to a comment explaining the situation then, and leave the hlint edits in?

kRITZCREEK:
I just realized that all the cleanup I propose here is already in #1964 . I'll close this.

Issue #1712, opened 2015-12-14
Fast recompilation by keeping externs files in memory


paf31:
I would like to support a new compilation mode which only recompiles a single module (the one which is active in your editor) and keeps its dependency externs files in memory. This is the approach taken in Try PureScript and it is very fast.

@user471 also suggests a command line option to skip code generation for a further speed up in some workflows.

paf31:
For reference, this is the approach I'd like to use

https://github.com/paf31/purescript-playground/blob/9fe9f44efd339fe05dc8b9c68b305898973bad95/src/index.js#L70-L83

garyb:
I was thinking about this after we talked about it the other week - shouldn't the compiler already do this? If we have a set of modules somewhere separate from the temporary input for tryps, then I would have thought they will only build once anyway - there's nothing the temporary input can do to modify these standard source files, so psc shouldn't recompile them as the timestamps will be in sync, and there will be no way to insert a dependency "above" them that would cause them to be recompiled.

paf31:
Yes, but the input files and output files have to exist. I'd like to avoid parsing the input files at all, and just assume the output files will have been compiled already. The externs would be added to the dependencies of every module.

garyb:
We still parse purs inputs even if the timestamp matches the compilation products then? Why do we use the timestamp mechanism at all then? :smile:

paf31:
We need to know what order to build modules in, so we need to parse all source files to figure that out.

garyb:
Ah of course, as we talked about storing that stuff in the externs files too. I guess doing that would also solve this problem then?

paf31:
We can't do that since the externs wouldn't update if a module's imports did. The solution seems to be to parse only as far as the module imports, if speed is the concern. But you still have to parse every input file.

In this case, I'm assuming the externs files represent modules which are compiled and won't change.

paf31:
I started working on this but ran into issues with module imports. Will continue to work on this, because I think it could be very good for incremental compilation, since we don't need to reparse anything that won't ever be changed (e.g. `bower_components` in most cases).

paf31:
This is implemented in Try PureScript now, and it's very fast. I think it would be good to provide this as an alternative compilation mode for editors.

kRITZCREEK:
:+1:

user471:
:+1:

paf31:
We should also do this in PSCi.

paf31:
[...]

PR #1975, opened 2016-03-26
#1712, fast recompilation in psc-ide-server

paf31:
@kRITZCREEK  and I are still working on this, but I'll put the code here for early review anyway.

Still to do:

- [x] Get all transitive dependencies and topologically sort them in the "load all" command
- [ ] Pass warnings back to the editor
- [ ] Handle foreign modules
- [x] Move code into its own module
- [x] Make output file into a command parameter

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -212,6 +212,7 @@ library
                      Language.PureScript.Ide.SourceFile
                      Language.PureScript.Ide.Watcher
                      Language.PureScript.Ide.Reexports
+                     Language.PureScript.Ide.Rebuild
 
                      Control.Monad.Logger
                      Control.Monad.Supply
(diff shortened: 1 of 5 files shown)


kRITZCREEK:
I'll look into the last two points now :)

kRITZCREEK:
The CI failure seems to be about dependency conflicts.

hdgarrood:
I think that CI failure can safely be ignored:

> transformers-compat-0.5.1.4 failed while downloading the package. The exception was: user error (Failed to download http://hackage.haskell.org/packages/archive/transformers-compat/0.5.1.4/transformers-compat-0.5.1.4.tar.gz: ErrorMisc "Error HTTP code: 503")

paf31:
I'll restart Travis now.

kRITZCREEK:
I think it makes sense to work on this after #1967 is finished. Then we can write tests that instruct the server and not have to drop into the editor/fiddle with ghci too much.

kRITZCREEK:
Did some more work on this in #1998

PR #1977, opened 2016-03-27
Fix #1978, disallow generalization over constraints for mutually-recursive functions

### LINKED ISSUE #1978: Soundness issue with constraint generalization
Constraint generalization doesn't play nicely with recursion in untyped binding groups:

```purescript
go f n _ | n <= 0 = f
go f n g = go (f >>> g) (n - 1) g
```

This program fails at runtime since the dictionary isn't inserted on the recursive call.

paf31:
This should be ready to review finally. This doesn't do let generalization yet, but it does the following things:

- Disallow generalization over constraints for mutually-recursive functions, since I can't think of a simple way to generate code for mutually recursive functions where we need to introduce new constraints, without running the type checker twice.
- Allow generalization over constraints for self-recursive functions. These get code gen'd as a let binding with a monomorphic helper function. This is more in line with the typing judgments.
- Allow generalization over constraints for non-recursive functions. This is the same as before.

### DIFF
--- a/examples/failing/Generalization1.purs
+++ b/examples/failing/Generalization1.purs
@@ -1,0 +1,11 @@
+-- @shouldFailWith CannotGeneralizeRecursiveFunction
+module Main where
+
+import Prelude
+
+foo 0 x _ = x
+foo n x y = x <> bar (n - 1) x y
+
+bar 0 x _ = x
+bar n x y = y <> foo (n - 1) x y
+
--- a/examples/passing/Generalization1.purs
+++ b/examples/passing/Generalization1.purs
@@ -1,0 +1,10 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (print)
+
+main = do
+  print (sum 1.0 2.0)
+  print (sum 1 2)
+
+sum x y = x + y
--- a/examples/passing/Generalization2.purs
+++ b/examples/passing/Generalization2.purs
@@ -1,0 +1,11 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (print)
+
+main = do
+  print (test 3 [1, 2])
+
+test n m | n <= 1 = m
+         | otherwise = test (n - 1) (m <> m)
+
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -120,6 +120,7 @@ data SimpleErrorMessage
   | UnusedTypeVar String
   | WildcardInferredType Type
   | MissingTypeDeclaration Ident Type
+  | CannotGeneralizeRecursiveFunction Ident Type
   | NotExhaustivePattern [[Binder]] Bool
   | OverlappingPattern [[Binder]] Bool
   | IncompleteExhaustivityCheck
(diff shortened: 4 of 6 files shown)


hdgarrood:
Are we sure we want to start warning for any declaration missing a type signature as opposed to just top-level ones? That seems a bit drastic.

paf31:
I would like to. I think we should treat let bound things the same way as top level declarations. Why not? main is essentially one big let binding anyway.

hdgarrood:
I would guess that lots of (maybe even most?) let-bound declarations' types are already clear enough from their encompassing top-level declaration's type signature (assuming it exists), that adding another type signature would do little to make the program easier to read, and may just end up being noise. Also, everyone is already used to writing code without them; perhaps we should instead ask whether they add enough that it's worth the disruption?

Maybe a good way of evaluating whether this is helpful or simply annoying would be to try to conform to it for a while before implementing it. What about, say, in the compiler or core libraries?

paf31:
FWIW, I already try to do this, and it bothers me when I realize that I forgot. I also tend to use let and where quite a lot.

kRITZCREEK:
I'm worried this will promote long lispy expressions, because of the inconvenience when introducing a binding. The idea is that where bindings are the place to introduce names for subexpressions?

garyb:
I can't say I'm super enthusiastic about the idea of warning about untyped `let` either, it seems like we're saying "although we have type inference, you shouldn't really use it".

I tend to type `where` declarations, but don't think I ever have done in a `let`, for the reasons Harry mentioned.

paf31:
I tend to use let bindings in a fairly non-standard way, I'll admit. I do think the warnings are still useful, because we might infer an overly polymorphic type, which could have actual performance implications. It falls into the category of irritating warnings for some use cases though, so I'll leave it out for now. We can add it back in if we see fit.

hdgarrood:
Would it be workable to infer the least polymorphic type possible in all bindings except top level ones?

paf31:
Well, that's what we do now. It's possible, but it means you have to put type signatures on things like lenses which you define in `where` clauses.

hdgarrood:
Oh, sorry, I mean, the least polymorphic type possible which still passes type checking. I suppose that could be quite a lot harder, though?

paf31:
Can you give an example of the difference?

hdgarrood:
Ok, given this:

```purescript
norm :: Number -> Number -> Number -> Number -> Number
norm x1 y1 x2 y2 = Math.sqrt (go x1 y1 + go x2 y2)
  where
  go a b = let c = a - b in c * c
```

the most general type of `go` is `Ring a => a -> a -> a`, but could the compiler notice that this can be specialised to `Number -> Number -> Number`?

Suppose we then change `norm` to this:

```purescript
norm' :: Number -> Number -> Number -> Number -> Int -> Number
norm' x1 y1 x2 y2 z = Math.sqrt (go x1 y1 + go x2 y2 + Int.toNumber (go 0 z))
  where
  go a b = let c = a - b in c * c
```

Then, the specialisation isn't possible, so `go` has to have the more general type with the `Ring` constraint. So I'm imagining that both of the above would compile successfully.

paf31:
Oh, I got it, so if a type variable gets instantiated at more than one distinct type, generalize, otherwise keep the ungeneralized type. That seems ... possible but tricky, and an interesting optimization.

paf31:
I've fixed the issue for self-recursive functions, and I'm working on a general solution for non-trivial binding groups.

paf31:
This is ready for review now.

Resolves #1978.

paf31:
Actually, there are still some bugs here.

paf31:
There seems to be at least two problems still, one in the renamer, and one somewhere else.

I'm giving up on this for tonight, but I don't want to release 0.8.4 until I can fix #1978 or remove constraint generalization.

PR #1984, opened 2016-03-30
Fix for rendered constrained types needing parens.

LiamGoodacre:
See https://github.com/purescript/pursuit/issues/176


hdgarrood:
:+1: LGTM. Thank you! Is it possible to add a test that fails with the old code?

garyb:
Uhh actually, it's not required now. See how they're not used for the `Ord` constraints here: https://github.com/purescript/purescript-maps/blob/master/src/Data/Map.purs

They're only required when there are multiple constraints.

hdgarrood:
@garyb the problem is that the type of `unsafePartial` comes out as `forall a. Partial => a -> a` with the old code. I think this code will only add parens where they are necessary.

garyb:
Ahh ok, a kinda precedence issue in the parens then, I see. I was just commenting that this seemed to include the optional parens change so wasn't sure what it was fixing. To make it super obvious for the future record this should be typed as `forall a. (Partial => a) -> a`.

LiamGoodacre:
Is it okay for me to export `codeToString` from `Language.PureScript.Docs.AsMarkdown` to use it in a test?

Also, is this how you envisioned it being tested?  Anything I should do differently?

hdgarrood:
:+1: from me. Exporting `codeToString` is fine, I think. The tests look great!

LiamGoodacre:
Also would you want this all in one commit?

hdgarrood:
The way it is now is fine, I think. I guess that's up to @paf31 and/or @garyb though.

LiamGoodacre:
For documentation purposes; running the new tests without the fix fails the test:
~~~purescript
Failed: TypeSynonymMismatch
  (ModuleName [ProperName {runProperName = "ConstrainedArgument"}])
  "WithoutArgs"
  "forall a. (Partial => a) -> a" -- expected
  "forall a. (Partial) => a -> a" -- actual
~~~

paf31:
Looks great! If you could please squash commits, I'll get this merged in.

paf31:
Thanks!

Issue #1986, opened 2016-03-30
Bundling fails in PureScript 0.8.3 on unicode characters in strings


garyb:
One example case: `"…"` is output as `"\u2026"`, but then when the module is bundled we're getting:

```
  The module could not be parsed:
  lexical error @ line 36 and column 54
```

Maybe something to do with `language-javascript`?

nwolverson:
https://github.com/erikd/language-javascript/issues/48 ?

nwolverson:
Actually should be fixed on latest 0.5 release https://github.com/erikd/language-javascript/commit/67c46ae0124c781a2b8d43e8c1f3adb3ec406903

nwolverson:
Again with latest prelude:
```
Error in module Prelude (Foreign):

  The module could not be parsed:
  lexical error @ line 217 and column 14
```
(this case is on `\b`, a valid JS escape). Tested with https://github.com/purescript/purescript/pull/1989

teh:
Unfortunately I forgot the \a escape because it wasn't documented on the es5 standard. If someone is blocking on this just add `a` and submit PR otherwise I'll do it this weekend

hdgarrood:
I'm fairly sure the `\a` escape is only a thing in Haskell, not JavaScript. I think language-javascript was right to reject that.

teh:
I think it's coming from here: https://github.com/purescript/purescript-prelude/blob/ce827333e813579b569a545235ffeb002c89d6b6/src/Data/Show.js#L18

teh:
https://github.com/purescript/purescript-prelude/issues/65 looks like it's already fixed? I think we can close this issue then with my fixes in language-javascript

nwolverson:
@hdgarrood Sort of right, *any* non escape character (ish) should be valid (http://www.ecma-international.org/ecma-262/5.1/#sec-7.8.4), the value of `\a` is `a`, similarly for other unrecognised escapes, so I think language-javascript is technically being too strict there.

@teh you're pointing at a fixed line, and I think the issue should be closed only when `psc-bundle` is actually using the fixed language-javascript? Either by bumping the version bounds or via #1989.

paf31:
Fixed by @nwolverson's PR, correct?

garyb:
It should be, yep. I'll test it now.

hdgarrood:
Using `psc-bundle` with both unicode string literals and the string escapes in Prelude are working for me on master now, yep.

garyb:
[...]

PR #1989, opened 2016-03-30
Update to use language-javascript 0.6.x

### LINKED ISSUE #1966: support language-javascript-0.6
Related: https://github.com/fpco/stackage/issues/1292

From `language-javascript` readme:

>There are currently two versions:
>
>* 0.5 series : Is a continuation of the 0.5.X.Y series, from the [master]
>(https://github.com/erikd/language-javascript/tree/master) branch of this
>github repository.
>
>* 0.6 series : This has a vastly different and improved AST which makes if far
>more difficult to build an non-sensical Javascript AST. This code is in the
>[new-ast](https://github.com/erikd/language-javascript/tree/new-ast) branch of
>this github repository.

So the change isn't trivial, but probably still quite straightforward. I hope.

nwolverson:
Fix #1966.

Not sure what the release path is (plus obviously this isn't going to build with our stackage/cabal builds as it is just now).

Must say this was more painful than anticipated, particularly trying to give similar output, whitespace in particular. To be fair the new AST is nicer, at least with syb to hand.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,8 @@ install:
     fi
   - mkdir -p .cabal-sandbox
   - cabal sandbox init --sandbox .cabal-sandbox
-  # Download stackage cabal.config, not sure whether filtering is necessary
-  - if [ -n "$STACKAGE" ]; then curl https://www.stackage.org/$STACKAGE/cabal.config | egrep -v 'purescript|sourcemap|bower-json' > cabal.config; fi
+  # Download stackage cabal.config. Filter this package and 'extra deps'
+  - if [ -n "$STACKAGE" ]; then curl https://www.stackage.org/$STACKAGE/cabal.config | egrep -v 'purescript|language-javascript|bower-json' > cabal.config; fi
   - cabal install --only-dependencies --enable-tests
   - cabal install hpc-coveralls
   # Snapshot state of the sandbox now, so we don't need to make new one for test install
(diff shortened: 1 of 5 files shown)


phadej:
`language-javascript`? :)

nwolverson:
Durr

hdgarrood:
We get around problems like this one (where we want different versions of packages to those that are in the snapshot) by filtering the `cabal.config`: https://github.com/purescript/purescript/blob/master/.travis.yml#L39

If you add `language-javascript` to that `grep -v` command, I think that should do it.

hdgarrood:
Also if you are going to modify that line, maybe remove the comment on the preceding line. It definitely is necessary ;)

nwolverson:
Yeah I did update that before. Also can now remove `sourcemap` as we're on later lts.

paf31:
So this is good to merge, yes? Thanks @nwolverson !

nwolverson:
Yes GTG. Definitely good to get a sanity check of the output, I've been comparing diffs for a couple of projects and should only differ in a little whitespace and couple of semicolons.

aspidites:
@nwolverson thanks for this. Until it's merged, I went ahead and compiled purescript using this branch and things work as expected. (I was running into #1986).

garyb:
Looks good to me, I'm not too concerned about imperfect formatting as what this is fixing is way more important!

@paf31 we should probably get a release out ASAP with this in, as the released `psc-bundle` 0.8.3 is basically unusable.

paf31:
Thanks!

PR #1993, opened 2016-04-02
Fix #1991, instantiate polymorphic types before unification

### LINKED ISSUE #1991: type inference regression in 0.8.3
The following compiles fine in 0.8.2, but fails in 0.8.3.

```
module Main.Test where

import Prelude
import Data.Set as S
import Data.Foldable (foldMap)

regression :: S.Set Int
regression =
  let as = [1,2,3,4,5]
      as' =
        foldMap
          (\x -> if 1 < x && x < 4 then S.singleton x else S.empty)
          as
  in
  as'
```

The compile error is.

```
  Could not match type

    a0

  with type

    Int


while trying to match type Set a0
  with type Set Int
while inferring the type of if (((&&) ((...) x)) (((<) x) 4))
                              then (singleton x)
                              else empty
in value declaration regression
```

using ```S.empty :: S.Set Int``` makes the code compile again.

Not sure if there was a design decision between 0.8.2 and 0.8.3 to help the compiler.

paf31:
This is a symptom of a bigger problem, which is that unification should only really ever happen on monotypes. For now, this should fix the immediate issue. I'm tracking the bigger issue elsewhere.

Tested with Halogen and Thermite.

### DIFF
--- a/examples/passing/1991.purs
+++ b/examples/passing/1991.purs
@@ -1,0 +1,20 @@
+module Main where
+
+import Prelude
+
+singleton :: forall a. a -> Array a
+singleton x = [x]
+
+empty :: forall a. Array a
+empty = []
+
+foldMap :: forall a m. (Semigroup m) => (a -> m) -> Array a -> m
+foldMap f [a, b, c, d, e] = f a <> f b <> f c <> f d <> f e
+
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

PR #1995, opened 2016-04-05
Disallow constraint generalization for recursive functions.

### LINKED ISSUE #1978: Soundness issue with constraint generalization

paf31:
Fixes #1978

I'll work on let generalization later, once 0.8.4 is out.

### DIFF
--- a/examples/failing/Generalization1.purs
+++ b/examples/failing/Generalization1.purs
@@ -1,0 +1,11 @@
+-- @shouldFailWith CannotGeneralizeRecursiveFunction
+module Main where
+
+import Prelude
+
+foo 0 x _ = x
+foo n x y = x <> bar (n - 1) x y
+
+bar 0 x _ = x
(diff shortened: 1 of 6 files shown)


paf31:
If someone could review this, I'll release 0.8.4.

garyb:
Looks good to me :+1: 

Is this something we can recover later potentially, or do you think it's better off just requiring a type signature?

paf31:
We should be able to relax this to "non-recursive functions, or self-recursive functions" with some minor changes to the renamer.

Mutually-recursive functions in general will either require a cunning code gen trick, or for us to run the typechecker twice. I don't particularly like either of those options, so I'd rather just exclude that corner case for now.

PR #1998, opened 2016-04-05
Sort Externfiles before creating the initial Environment

kRITZCREEK:
Seems to be working... Didn't try it on a big project yet


kRITZCREEK:
This is kind of hard to test without the "loadAll" command from #1967.

paf31:
Looks good, thanks!

PR #1999, opened 2016-04-05
Add library-only flag to the cabal manifest

adinapoli:
Hey @paf31 !

I had recently the need for building a tool on top of the PureScript AST/CodeGen/Printer, but I didn't want to install gazillion of extra packages and  produces lots of extra executables just to depend of `purescript`, the library.

Therefore, I have created this PR which ship an extra `library-only` flag to the manifest, so that I can install only the library if I want:

```
stack install --fast --flag purescript:library-only
```

Last but not least, this patch prevents the installation of new binaries in my `.local/bin` directory, which could override the current version I'm using for work (< 0.8).

Do you see this as something useful you could merge into master?

Thanks a lot!

Alfredo

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,6 +5,7 @@ This file lists the contributors to the PureScript compiler project, and the ter
 ### Individuals
 
 - [@5outh](https://github.com/5outh) (Benjamin Kovach) - My existing contributions and all future contributions until further notice are Copyright Benjamin Kovach, and are licensed to the owners and users of the PureScript compiler project under the terms of the MIT license.
+- [@adinapoli](https://github.com/adinapoli) (Alfredo Di Napoli) My existing contributions and all future contributions until further notice are Copyright Alfredo Di Napoli, and are licensed to the owners and users of the PureScript compiler project under the terms of the [MIT license](http://opensource.org/licenses/MIT).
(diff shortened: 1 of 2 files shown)


paf31:
Looks fine, I'm just a bit surprised that `stack` doesn't support this out of the box. Could you please update `contributors.md` too?

adinapoli:
I will as soon as I get hold of a real computer. Thanks Pil!

paf31:
I wonder how hard it would be to make a PR on stack itself to add this. I know I would use it all the time. `stack install` does too much work right now, it seems.

adinapoli:
> I wonder how hard it would be to make a PR on stack itself to add this.

I guess can certainly be done, although having the flag here means also the cabal-install folks can benefit from this. For example, at work, there are certainly individuals which haven't switched to stack yet, believe it or not :wink: 

I added myself and my company IRIS Connect to the list of contributors. Thanks Mr. @paf31 !

adinapoli:
Hey @paf31 , I didn't push through this branch as it wasn't clear with conversation with you if you were happy with the current shape of this PR or how you wanted to improved it, if not!

Let me know and I will solve the conflicts and re-submit! (Although I think it would be quicker for me to start fresh from `HEAD` instead of bothering with painful conflict solving 😉  )

paf31:
One issue here is that we now have many more dependencies in the library itself, so not building the executables won't help as much now. A lot of the PSCi code has moved into the library, for example.

adinapoli:
Ah, I see what you mean. I guess that makes this PR moot then. The whole point of it was to avoid extra deps, which now seems is not possible anymore 😁  (or, better, not in the same order of magnitude as before!)

PR #2001, opened 2016-04-07
Language/PureScript/Ide/Types.hs: fix haddock markup

trofi:
Failed as:

    src/Language/PureScript/Ide/Types.hs:63:7: parse error on input ‘(’

Signed-off-by: Sergei Trofimovich <siarheit@google.com>

### DIFF
--- a/src/Language/PureScript/Ide/Types.hs
+++ b/src/Language/PureScript/Ide/Types.hs
@@ -59,9 +59,9 @@ data ExternDecl
         [DeclIdent] -- The exported identifiers
     -- | A data/newtype declaration
     | DataConstructor
-      DeclIdent -- ^ The type name
+      DeclIdent -- The type name
       (P.ProperName 'P.TypeName)
-      P.Type      -- ^ The "type"
+      P.Type      -- The "type"
     -- | An exported module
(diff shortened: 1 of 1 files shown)


hdgarrood:
Thanks, but won't this mean that these comments will no longer appear in the haddock documentation - can this be done differently so that these comments are preserved?

paf31:
Move them into their own record?

hdgarrood:
I think defining a record type here would introduce partial functions for the accessors, though.

paf31:
I meant a separate record.

hdgarrood:
Ahh, right, I see. Yeah that would be nice. I've just realised that none of the other constructors for this data type have haddock markup on individual fields either, just normal comments (which won't be picked up by haddock), and given that this error stops us from generating Haddock docs completely, maybe we should just go with this approach for now and worry about documenting all the fields later.

Given that haddock can fail even on valid Haskell source files (this is new to me!) I think we should add `stack haddock` to at least one of our CI builds, too.

kRITZCREEK:
Argh. :D Looks like we should add a haddock build step... Just normal comments should be enough for now.

PR #2002, opened 2016-04-08
Windows CI

### LINKED ISSUE #1952: Include exact versions of used libraries in binary bundles?
This could turn out to be very useful for debugging.

hdgarrood:
* Add appveyor.yml for windows CI, with automatic release binary bundle
  publishing
* Rework bundle/build-stack script to work on all OSes
* Use SHA1 instead of MD5 for the Windows checksum.
* Refactor the build-stack script a little
* Remove all the other bundle build scripts
* Include dependencies in binary bundles (resolves #1952)

I haven't managed to get `strip` to work on AppVeyor, but this script
has produced an identically sized bundle to the 0.8.4 one (21.3MB) so
maybe that's not worth worrying about.

The encrypted github token here is tied to my github account, but it appears I have the permission to edit releases now so that should work fine.


bundle/build-stack.sh
```diff
@@ -10,39 +11,51 @@ then
...
+pushd bundle/build > /dev/null
+tar -zcvf ../${OS}.tar.gz purescript
 popd > /dev/null
 
 # Calculate the SHA hash
-shasum $OS.tar.gz > $OS.sha
+if [ "$OS" = "win64" ]
+then
+  # msys/mingw does not include shasum. :(
+  SHASUM="openssl dgst -sha1"
```
garyb: Ah cool, I wondered how you'd "fixed" that.

paf31:
Looks great! Is Appveyor able to build Mac binaries too?

For the secret key, can it be put into an environment variable like in Travis? This way seems like a security issue.

hdgarrood:
No, Appveyor is just Windows. Travis can build Mac binaries, though.

The github token here is encrypted, so this ought to be safe (assuming Appveyor doesn't get hacked or anything) but yeah, we can do it like Travis too.

hdgarrood:
I'm going to merge this for now to stop it from making all the other CI builds fail; I think we can always revert this later if it turns out not to work well.

PR #2004, opened 2016-04-09
Remove RedundantUnqualifiedImport

### LINKED ISSUE #1823: Invalid redundant import warning
Consider the following module:

``` purescript
module Example (module Example, module X) where

import Prelude (Unit()) as X
import Prelude as Prelude

foo :: X.Unit
foo = Prelude.unit
```

It exports `Unit()` from the prelude and also exports `foo` as an alias for `unit`. Compiling it with PureScript 0.8.0-rc.1 gives the following warning:

```
Warning 1 of 2:

  in module Example
  at /private/tmp/ps/src/Example.purs line 3, column 1 - line 4, column 1

    Import of Prelude (Unit) is redundant due to a whole-module import



  See https://github.com/purescript/purescript/wiki/Error-Code-RedundantUnqualifiedImport for more information,
```

I think that is incorrect. I am explicitly importing `Prelude` twice under different aliases in order to selectively export some values and refer to others.

garyb:
Resolves #1823. Also fixes DuplicateSelectiveImport.

There was a mistake where both these warnings were only being applied to qualified imports, but the intention was they should be applied to qualified imports.

DuplicateSelectiveImport now catches both cases, so RedundantUnqualifiedImport is not necessary. DuplicateSelectiveImport  will also work correctly for both qualified and unqualified imports:

``` purescript
import Prelude
import Prelude (unit)
```

``` purescript
import Prelude (Unit)
import Prelude (unit)
```

``` purescript
import Prelude (Unit) as P
import Prelude (unit) as P
```

All raise the warning, whereas:

``` purescript
import Prelude (Unit) as P
import Prelude
```

``` purescript
import Prelude (Unit) as P
import Prelude (unit) as Q
```

etc. do not now.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -139,7 +139,6 @@ data SimpleErrorMessage
   | DeprecatedQualifiedSyntax ModuleName ModuleName
   | DeprecatedClassImport ModuleName (ProperName 'ClassName)
   | DeprecatedClassExport (ProperName 'ClassName)
-  | RedundantUnqualifiedImport ModuleName ImportDeclarationType
   | DuplicateSelectiveImport ModuleName
   | DuplicateImport ModuleName ImportDeclarationType (Maybe ModuleName)
   | DuplicateImportRef String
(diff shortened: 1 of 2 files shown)


paf31:
👍 Looks good!

PR #2007, opened 2016-04-09
Fix re-exports to include intended exports only

### LINKED ISSUE #1872: Reexport wrongly includes aliased modules
When reexporting a module, other imports of the same module are included even if they are aliased. This seems like a bug to me.

``` purescript
-- A.purs
module A where
data X = X
data Y = Y

-- B.purs
module B (module A) where
import A (X(X))
import A (Y(Y)) as C

--
import B as B
B.X -- available
B.Y -- also available
```

Curiously I do get a warning about the qualified import.

```
Warning found:
in module B
at /private/tmp/example/src/B.purs line 3, column 1 - line 3, column 17

  The import of module A is redundant


See https://github.com/purescript/purescript/wiki/Error-Code-UnusedImport for more information,
```

I can work around this by importing the module qualified as itself.

``` purescript
-- B.purs
module B (module A) where
import A (X(X)) as A
import A (Y(Y)) as C

--
import B as B
B.X -- available
B.Y -- no longer available
```

garyb:
Resolves #1872

When importing a module as both qualified an unqualified, re-exporting the unqualified version would also re-export any values the qualified import referenced.

### DIFF
--- a/src/Language/PureScript/Sugar/Names/Exports.hs
+++ b/src/Language/PureScript/Sugar/Names/Exports.hs
@@ -119,9 +119,10 @@ resolveExports env mn imps exps refs =
   extract useQual name render = fmap (map (importName . head . snd)) . go . M.toList
     where
     go = filterM $ \(name', options) -> do
-      let isMatch = if useQual then eqQual name name' else any (eqQual name . importName) options
+      let isMatch = if useQual then isQualifiedWith name name' else any (checkUnqual name') options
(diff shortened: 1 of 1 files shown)


hdgarrood:
You probably worked it out already, but the appveyor failures can be safely ignored here - it's only failing because the commits which added appveyor.yml weren't in this branch yet.

garyb:
I hadn't even noticed they'd failed actually! I assumed everything was alright as I got the passing email from Travis.

paf31:
👍

PR #2013, opened 2016-04-10
#1712, fast recompilation in psc-ide-server

kRITZCREEK:
@kRITZCREEK and I are still working on this, but I'll put the code here for early review anyway.

Still to do:

 - [x] Get all transitive dependencies and topologically sort them in the "load all" command
 - [x] Pass warnings back to the editor
 - [x] Handle foreign modules
 - [x] Move code into its own module
 - [x] Make output file into a command parameter


paf31:
I'll try to merge these changes into the other PR later today.

kRITZCREEK:
This PR is basically the other one rebased on top of master and then my latest 3 commits added. I don't think there is anything missing.

paf31:
Oh right, it's just that I can't commit to your branch 😄 But that's fine, I can send you a patch in a separate PR or something.

paf31:
@kRITZCREEK  Does this work for you in Emacs after my changes?

paf31:
Here's the failing test case:

```text
Language.PureScript.Ide.Rebuild, Rebuilding single modules, fails to rebuild an incorrect module without dependencies and returns the errors
       uncaught exception: IOException of type OtherError (readCreateProcess: psc-ide-client (exit 1): failed)
```

Maybe this is just an issue with the test runner, since it only fails in one of three runs.

kRITZCREEK:
hmm maybe I was a bit too hasty... :D I get a bunch of unknown Type errors for Prim Types. Do we need to somehow add those to the environment before rebuilding?

The CI Error seems to be a timing thing, the tests run fine on my local machine.

kRITZCREEK:
Okay for a minimal case:

If I create a single file Main.purs

```purescript
module Main where
y = 4
```

It rebuilds fine and tells me to annotate y with `Int`

```purescript
module Main where
y :: Int
y = 4
```

Fails with Unkown Type Int though

src/Language/PureScript/Ide/Rebuild.hs
```diff
@@ -52,28 +43,12 @@ rebuildFile path outpath = do
 
   outputDirectory <- confOutputPath . envConfiguration <$> ask
 
-  let initEnv = foldl' (flip P.applyExternsFileToEnvironment) P.initEnvironment externs
+  let ma = P.buildMakeActions outputDirectory M.empty M.empty False
 
-  (resultMay, warnings) <- liftIO . Logger.runLogger' . runExceptT . flip runReaderT P.defaultOptions $ do
-    ((P.Module ss coms moduleName elaborated exps, env), nextVar) <- P.runSupplyT 0 $ do
-      [desugared] <- P.desugar externs [ P.addDefaultImport (P.ModuleName [P.ProperName "Prim"]) m ]
```
kRITZCREEK: This line added the `Prim` import before

kRITZCREEK:
That fixed the `Prim` thing. Next up would be to construct the Map from Module -> Foreign. As a best effort I think it's fine to assume the Foreign definitions are in a file right next to the module with the extension swapped from `.purs` to `.js`?

paf31:
Without parsing all of the foreign modules, that's the simplest approach, yeah.

kRITZCREEK:
Oh dear, I think we missed a major problem. If we just take all the externs and sort them we still keep modules that can depend on the module we are trying to compile.
Just sorting all the Externs isn't going to cut it. Instead we'll need to parse the imports in the file to rebuild and start gathering its transitive dependencies.
The other possibility would be to search for modules that depend on the module we are trying to build and transitively prune this and any module that depends on them.

garyb:
I previously suggested we start storing the dependency info in the externs, seems like it would be helpful here. 

Phil reckons that isn't good enough though and we need to make some changes so we can parse module headers to determine the dependency graph though. That would involve quite a bit more work at the moment, as we'd also have to disallow multiple modules per file, and that means figuring out a new way of making the test suite work as something like 50% of them have multiple modules :wink:

kRITZCREEK:
@garyb the dependency information is stored in the ExternsFile format: https://github.com/purescript/purescript/blob/83896db8fcc752b609654c91da84cd84ad788af5/src/Language/PureScript/Externs.hs#L48

When building the Externs the compiler even gathers the "inline dependencies" (ala Debug.Trace.spy).

garyb:
Oh right, nevermind, I don't read good apparently :wink:

paf31:
Well, we know what externs we have loaded in memory, and we know what imports are in the latest file. If the first doesn't contain all of the second, can we just refresh the externs before rebuilding?

I would definitely like to see changes eventually so that we only need to parse module headers to get the module graph, but that's not a priority. Maybe something for 1.0? My point with the externs there was just that you can't use the externs alone to figure out the module graph, because they don't change if the imports change in the source files.

garyb:
> My point with the externs there was just that you can't use the externs alone to figure out the module graph, because they don't change if the imports change in the source files.

Sure, but anything "above" a changed module in the tree can be loaded from externs, you only need to re-parse the modules that have changed. I was envisioning more of a patch operation on the externs-loaded tree.

kRITZCREEK:
It's not about dependencies of the current file missing, but rather that we could have a module that depends on the module we are trying to build. That will throw a Module Not Found error when we try to apply it to the Environment.

Example:

```purescript
module A where
import B
```

```purescript
module B where

id x = x
```

=> We can rebuild module A because B will be aplied to the Environment.

When we try to build B we apply A to the Environment but because we are building B we don't apply its ExternsFile to the Environment... The problem is my naive approach of just applying ALL the loaded ExternsFiles to the Environment before rebuilding^^

paf31:
Well, I think a decent case can be made for it, but I'd still much rather do things the simplest way, and the "usual" way. Explaining the compilation pipeline at the top level shouldn't be too hard to do, but I think it could be if we have to explain this sort of thing.

Plus, adding in a quick `Debug.Trace.spy` or whatever is becoming _much_ simpler now with the editor support which adds the import for you. (And we've come full circle...)

paf31:
@kRITZCREEK But aren't we working under the assumption that the user is editing a single file here? If they switch files, we need to reload dependencies _for the new file_ anyway.

garyb:
@kRITZCREEK yeah, I realised why my point was irrelevant after your reply, I went back and read more what you were saying. The rest of this has just been a tangent trying to elaborate my thoughts on the externs dependencies thing that Phil and I have much discussed in the past without ever entirely agreeing :smile:

kRITZCREEK:
The commit I just pushed work on my minimal examples but fails when I try it against `Halogen`. I get a module not found Control.Alt in that case. Is there some edge case I'm missing in my very naive gatherTransitiveDependencies?

paf31:
Shouldn't there be a recursive call to `go` somewhere?

paf31:
Also, `Data.Graph` has a function to find transitive closures, FWIW.

jdegoes:
Super cool, btw. :tada:

paf31:
Wait, `sortModules` already does this, the data is in the second component of the tuple.

kRITZCREEK:
:facepalm: Thanks 💃 
So I tried to make it fly with the result from sortModules but I couldn't make it work. Maybe I'll take another look tomorrow.

It does work now but it's super slow when building `Halogen`...

paf31:
I can take a look with you tomorrow if you like. I have some ideas for how we could speed this up.

AppShipIt:
A few of us will be on Zoom if y'all would like to use it.

kRITZCREEK:
That would be great :) Let me know when it suits you. I've basically got time whenever today

AppShipIt:
I'll be home all day working on and off. DM me on Twitter whenever is convenient!

kRITZCREEK:
I had another look and I missed the `inOrderOf` call inside when using the result from the graph sort. This  fells crazy fast now O.o Halogen checks instantaneous

garyb:
All of us at SlamData are pretty excited about this, since our current incremental compile is 20+ seconds :wink:

kRITZCREEK:
Do you want me to try and rebuild the toplevel module of SlamData?^^ I want to write a few benchmarks to track regressions anyway :)

garyb:
Yeah, that'd be great! You should be able to do so without any particular effort, as apart from using gulp to produce two separate bundles it is still buildable with a standard pulp setup (there are kinda two entry point modules, but neither is `Main`).

kRITZCREEK:
I did something wrong O.o Rebuild.hs is a binary file to git O.o

kRITZCREEK:
I just thought of another cool thing this enables.

Right now psc-ide can only complete exported identifiers. But if we'd strip away the export list from the module we are rebuilding we would get an ExternsFile that contained all of the module-local definitions.

As long as the editor keeps track of what file we are editing right now, we could hold on to that ExternsFile to complete module-local identifiers without writing something "wrong" to disc. But that's a thing for another pull request.

psc-ide-server/Main.hs
```diff
@@ -125,7 +125,7 @@ startServer port env = withSocketsDo $ do
           case decodeT cmd of
             Just cmd' -> do
               result <- runExceptT (handleCommand cmd')
-              $(logDebug) ("Answer was: " <> T.pack (show result))
+              -- $(logDebug) ("Answer was: " <> T.pack (show result))
```
paf31: Is this meant to be removed?

kRITZCREEK:
I should've deleted that line instead of commented. That's part of the I want to make the logging better/ well formatted. Just outputting the raw json doesn't actually help with anything since you get that back anyway.

paf31:
💯

kRITZCREEK:
🎉

FrigoEU:
Just want to say that I think this is amazing. I integrated it into psc-ide-vim, and it's just insanely fast. Seriously a joy to work with. Combined with @garyb's typed holes, 0.8.5.0 has been amazing for the dev experience. Thanks!

robinp:
A quick question: is it possible (or would be possible without much effort) to have separate compilation with the psc tool? In 0.8.5 I see it outputs an `externs.json` and an `index.js` file for each module. 

However, in case module B depends on A, I see no way to recompile only module B and take the already compiled externs/js for module A. Would be nice to expose this functionality, so third-party build tools could build incrementally.

paf31:
The compiler should do this automatically if module A has build products which are more recent than its source files.

Issue #2014, opened 2016-04-13
Qualified constructors are not valid identifiers for operator aliases


LiamGoodacre:
The following code:
~~~purescript
import Data.Tuple as T
infix 4 T.Tuple as &
~~~

Fails with this error:
~~~
  Unable to parse module:
  unexpected "Tuple"
  expecting qualifier or identifier
~~~

However this does work:
~~~purescript
import Data.Tuple (Tuple(Tuple))
infix 4 Tuple as &
~~~

PR #2015, opened 2016-04-13
Qualified constructors are not valid identifiers for operator aliases

LiamGoodacre:
For https://github.com/purescript/purescript/issues/2014

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -126,8 +126,8 @@ parseFixityDeclaration = do
   name <- symbol
   return $ FixityDeclaration fixity name alias
   where
-  aliased = (Left <$> parseQualified (Ident <$> identifier))
-        <|> (Right <$> parseQualified (ProperName <$> uname))
+  aliased = P.try (Left <$> parseQualified (Ident <$> identifier))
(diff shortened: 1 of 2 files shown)


garyb:
That was quick :wink:

Looks good to me.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -126,7 +126,7 @@ parseFixityDeclaration = do
   name <- symbol
   return $ FixityDeclaration fixity name alias
   where
-  aliased = (Left <$> parseQualified (Ident <$> identifier))
+  aliased = P.try (Left <$> parseQualified (Ident <$> identifier))
         <|> (Right <$> parseQualified (ProperName <$> uname))
```
LiamGoodacre: Also `ProperName <$> uname` is the same as `properName` from `Common` - should I switch it to that?

garyb: Yeah, I guess so! I probably just copied the `Ident` case when adding constructors. We intentionally don't use the `Common` one there as `parseIdent` will accept `Op` or `Ident`.

LiamGoodacre: Yeah I noticed the difference to `parseIdent` :]

PR #2020, opened 2016-04-16
Wildcards for case and if

### LINKED ISSUE #1558: LambdaCase
e.g.
``` purescript
fn = \case 
  X -> true
  Y -> false
```

Perhaps an "if lambda" would be neat too, although I'm not too sure how to arrange that so the phrasing works nicely.

``` purescript
fn = \if X else Y
```

Although I guess LambdaCase lets you do it anyway:

``` purescript
fn = \case
  true -> X
  false -> Y
```

garyb:
Resolves #1558

### DIFF
--- a/examples/passing/CaseInputWildcard.purs
+++ b/examples/passing/CaseInputWildcard.purs
@@ -1,0 +1,18 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff (Eff)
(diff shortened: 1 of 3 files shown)


AppShipIt:
You're a gentleman and a scholar 🎩

paf31:
👍 Looks great!

PR #2021, opened 2016-04-16
Don't warn for a single hiding import

### LINKED ISSUE #2017: Don't warn on a single `hiding` import
This should be allowed, just as one fully open import is (but not a `hiding` + full open, etc. just that `hiding` is a special case of open and is currently not treated the same way).

garyb:
Resolves #2017

@kRITZCREEK I updated `Ide.Imports` too, can you take a look?


paf31:
👍

kRITZCREEK:
:+1:

garyb:
Thanks!

PR #2022, opened 2016-04-16
Fix cycles being detected incorrectly with qualified imports

### LINKED ISSUE #2018: `CycleInModules` thrown for `import Some.Module.Name as Main`
There seems to be an issue with qualified imports of the form `import Some.Module.Name as Main`. Take e.g. this example:

Main.purs:
```purescript
module Main where

import MyProject.App ( runApp )

main = runApp
```
MyProject/App.purs:
```purescript
module MyProject.App
  ( runApp )
where

import MyProject.Components.Main as Main

runApp = Main.runComponent
```
MyProject/Components/Main.purs:
```purescript
module MyProject.Components.Main
  ( runComponent )
where

import Control.Monad.Eff.Console ( log )

runComponent = log "PureScript is great!"
```
This example will give an error

```
  There is a cycle in module dependencies in these modules:

    MyProject.App
    Main

```
If you replace `Main` with something else in `MyProject/App.purs`, everything works fine.

garyb:
Resolves #2018

### DIFF
--- a/examples/passing/2018.purs
+++ b/examples/passing/2018.purs
@@ -1,0 +1,25 @@
+module B where
+
+  data Foo = X | Y
+
+module A where
+
+  import B as Main
+
+  -- Prior to the 2018 fix this would be detected as a cycle between A and Main.
+  foo ∷ Main.Foo → Main.Foo
+  foo x = x
+
+module Main where
+
+import Prelude
+import A (foo)
+import B (Foo(..))
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (CONSOLE, log)
+
+main :: forall e. Eff (console :: CONSOLE | e) Unit
+main = do
+  let tmp = foo X
+  log "Done"
+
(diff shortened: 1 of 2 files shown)


paf31:
👍 If I understand correctly, `B` still comes before `A` in the ordering because of the import declaration case? If so, maybe that's worth a comment, since it took me a few minutes to see how it worked. Looks good though!

garyb:
@paf31 Sorry, which is the part that's unclear? In the example/test?

paf31:
Just how `B` comes before `A` in the ordering. If I understand correctly, it's because of the line 

```purescript
forDecls (ImportDeclaration mn _ _ _) = [mn]
```

garyb:
Ah ok, I see. I'll add some more explanation while I'm rebasing now then :+1:

garyb:
How's this? It's hard to write clear explanations when you're talking about this module and some other hypothetical other module along with "pseudo" modules that may or may not exist with the same name as the some other module :smile:

paf31:
👍 Looks great, thanks!

garyb:
The build failure on this was a AppVeyor glitch, it succeeded everywhere else so I'll merge it as is.

PR #2023, opened 2016-04-16
Typed holes

### LINKED ISSUE #1283: Typed holes
Like how GHC does it. For example, if I write this:
```
foo :: List Int -> List Int
foo = map _
```
then I should get an error along the lines of:
```
Found typed hole with type Int -> Int:
  foo = map _
```
Perhaps distinguishing them from the record sugar like `{ foo: _ }` will be a problem though? If so, we could maybe require typed holes to be named, like `_foo`.

garyb:
Will resolve #1283.

This works up to the point where it only warns instead of errors when a hole is encountered (I want the error to happen after we're done with all the modules in typechecking, so poking around in `Make` to do so right now), but I figured we may want to talk about syntax.

Currently any identifier that is prefixed with `?` is considered a hole. This means we're in whitespace-aware territory, and I had to make it happen in the lexer. I don't feel too bad about that at all, because this is going to be awesome :wink:, but thought I'd open it for discussion anyway. The identifier part is used in the error message, so you get a message like:

```
Type hole 'ummm' has the inferred type:
 ...
```

Our choices seem to be either do it this way, or we'll have to introduce a new reserved symbol or name to make it happen.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Monad.hs
+++ b/src/Language/PureScript/TypeChecker/Monad.hs
@@ -225,17 +225,18 @@ runCheck = runCheck' initEnvironment
 -- | Run a computation in the typechecking monad, failing with an error, or succeeding with a return value and the final @Environment@.
 runCheck' :: (Functor m) => Environment -> StateT CheckState m a -> m (a, Environment)
 runCheck' env check = second checkEnv <$> runStateT check (emptyCheckState env)
+
 -- | Make an assertion, failing with an error message
 guardWith :: (MonadError e m) => e -> Bool -> m ()
 guardWith _ True = return ()
 guardWith e False = throwError e
 
 -- | Run a computation in the substitution monad, generating a return value and the final substitution.
-liftUnify ::
-  (MonadState CheckState m, MonadWriter MultipleErrors m, MonadError MultipleErrors m) =>
-  m a ->
-  m (a, Substitution)
-liftUnify ma = do
+withEmptySubstitution
+  :: (MonadState CheckState m)
+  => m a
+  -> m (a, Substitution)
+withEmptySubstitution ma = do
   orig <- get
   modify $ \st -> st { checkSubstitution = emptySubstitution }
   a <- ma
--- a/src/Language/PureScript/TypeChecker/Monad.hs
+++ b/src/Language/PureScript/TypeChecker/Monad.hs
@@ -18,7 +18,7 @@ import qualified Data.Map as M
 import Control.Arrow (second)
 import Control.Monad.State
 import Control.Monad.Error.Class (MonadError(..))
-import Control.Monad.Writer.Class (MonadWriter(..), listen, censor)
+import Control.Monad.Writer.Class (MonadWriter(..))
 
 import Language.PureScript.Environment
 import Language.PureScript.Errors
@@ -235,19 +235,11 @@ liftUnify ::
   (MonadState CheckState m, MonadWriter MultipleErrors m, MonadError MultipleErrors m) =>
   m a ->
   m (a, Substitution)
-liftUnify = liftUnifyWarnings (const id)
-
--- | Run a computation in the substitution monad, generating a return value, the final substitution and updating warnings values.
-liftUnifyWarnings ::
-  (MonadState CheckState m, MonadWriter MultipleErrors m, MonadError MultipleErrors m) =>
-  (Substitution -> ErrorMessage -> ErrorMessage) ->
-  m a ->
-  m (a, Substitution)
-liftUnifyWarnings replace ma = do
+liftUnify ma = do
   orig <- get
   modify $ \st -> st { checkSubstitution = emptySubstitution }
-  (a, w) <- reflectErrors . censor (const mempty) . reifyErrors . listen $ ma
+  a <- ma
   subst <- gets checkSubstitution
-  tell . onErrorMessages (replace subst) $ w
   modify $ \st -> st { checkSubstitution = checkSubstitution orig }
   return (a, subst)
+
--- a/src/Language/PureScript/TypeChecker/Monad.hs
+++ b/src/Language/PureScript/TypeChecker/Monad.hs
@@ -18,7 +18,7 @@ import qualified Data.Map as M
 import Control.Arrow (second)
 import Control.Monad.State
 import Control.Monad.Error.Class (MonadError(..))
-import Control.Monad.Writer.Class (MonadWriter(..))
+import Control.Monad.Writer.Class (MonadWriter(..), listen, censor)
 
 import Language.PureScript.Environment
 import Language.PureScript.Errors
@@ -232,15 +232,23 @@ guardWith _ True = return ()
 guardWith e False = throwError e
 
 -- | Run a computation in the substitution monad, generating a return value and the final substitution.
-withEmptySubstitution
-  :: (MonadState CheckState m)
-  => m a
-  -> m (a, Substitution)
-withEmptySubstitution ma = do
+liftUnify ::
+  (MonadState CheckState m, MonadWriter MultipleErrors m, MonadError MultipleErrors m) =>
+  m a ->
+  m (a, Substitution)
+liftUnify = liftUnifyWarnings (const id)
+
+-- | Run a computation in the substitution monad, generating a return value, the final substitution and updating warnings values.
+liftUnifyWarnings ::
+  (MonadState CheckState m, MonadWriter MultipleErrors m, MonadError MultipleErrors m) =>
+  (Substitution -> ErrorMessage -> ErrorMessage) ->
+  m a ->
+  m (a, Substitution)
+liftUnifyWarnings replace ma = do
   orig <- get
   modify $ \st -> st { checkSubstitution = emptySubstitution }
-  a <- ma
+  (a, w) <- reflectErrors . censor (const mempty) . reifyErrors . listen $ ma
   subst <- gets checkSubstitution
+  tell . onErrorMessages (replace subst) $ w
   modify $ \st -> st { checkSubstitution = checkSubstitution orig }
   return (a, subst)
-
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -74,7 +74,7 @@ typesOf ::
   [(Ident, Expr)] ->
   m [(Ident, (Expr, Type))]
 typesOf bindingGroupType moduleName vals = do
-  ((tys, subst), warnings) <- reflectErrors . censor (const mempty) . reifyErrors . listen $ withEmptySubstitution $ do
+  ((tys, warnings), subst) <- withEmptySubstitution . reflectErrors . censor (const mempty) . reifyErrors . listen $ do
     (untyped, typed, dict, untypedDict) <- typeDictionaryForBindingGroup moduleName vals
     ds1 <- parU typed $ \e -> checkTypedBindingGroupElement moduleName e dict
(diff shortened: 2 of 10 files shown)


AppShipIt:
> Currently any identifier that is prefixed with ? is considered a hole.

Inspired by Idris?

> This means we're in whitespace-aware territory,

Because `?` isn't a valid identifier token? Could that be changed in the parser?

garyb:
> Inspired by Idris?

Inspiration once-removed :smile: `?todo` was mentioned in the issue (#1238).

> Because `?` isn't a valid identifier token? Could that be changed in the parser?

Not unless we want to lose the ability to have operators involving `?`, of which there are a few commonly used already (`.?` in argonaut, `?~` in lens). This seemed the lesser of two evils.

paf31:
Nice :) I think the lexer is the right place to add the new syntax, for sure.

paf31:
Next step - hook it up to Pursuit's type search, and have the editor integration fill in type holes for us 😄

garyb:
Haha. `psc-ide` will put us all out of jobs if we keep going down that route. PureScript projects will write themselves :laughing:

AppShipIt:
> Not unless we want to lose the ability to have operators involving ?

I plan to look myself (unless it's a waste of time) but could it not switch based on characters that follow?

garyb:
> I plan to look myself (unless it's a waste of time) but could it not switch based on characters that follow?

Yeah, that's the change I've made - so it's not specifically whitespace aware so much as "everything that is not an identifier"-aware. But if you write `? foo` it will be parsed differently than `?foo` was the case I meant.

paf31:
I think the trick could be to warn in the typechecker and throw in the code generator?

AppShipIt:
> But if you write `? foo` it will be parsed differently than `?foo` was the case I meant.

Oh wow. That's potentially problematic in the case of a ternary-like operator I suppose 😬

garyb:
@paf31 I thought that too, but it codegens immediately after typechecking. Each module is done end-to-end pretty much. I'm thinking I'll `unless` everything after type checking to prevent codegen/externs for the module, and then have `buildModule` return the hole warnings, and then finally concat & throw out the warnings as needed.

paf31:
Why problematic? There's a difference between whitespace parsing in the lexer, and whitespace ambiguity in the parser. The latter leads to bad error messages due to backtracking.

paf31:
@garyb But we want it to be per-module, surely?

paf31:
Oh also, we should discuss these latest merges in the context of 0.8.5.

garyb:
@paf31 I figured you'd want it project-wide, that way you know everything would compile, aside from the fact there are holes that need filling in, even if you're working across multiple modules, etc.

paf31:
Possibly, but that just seems like asking for trouble with separate compilation.

AppShipIt:
> Why problematic?

It just reminds me an awful lot of the issue with `(-)` operator sections in that `(- 1)` meant something different than `(-1)`. Now `(? x)` would mean something different than `(?x)` although more corner case.

garyb:
@paf31 You could be right, I'll give this a try and poke at it anyway. I've had it working at the per-module level already, so can go back to it easily enough.

@AppShipIt that won't be valid soon, you'd have to write (_ ? x) anyway.

paf31:
It's similar, but not the same. The ambiguity before was in the parser, this is handled in the lexer. An alternative solution to the operator sections issue would have been to modify the lexer. This way, whitespace might change the lexer output, but the parse is still unambiguous for a given lexer output.

My concern over ambiguity in the parser is related to error messages and backtracking. That's not an issue here.

paf31:
@garyb For the single module case, I think you just need to add a case to the `check` function, and update the function which applies the substitution to warnings.

AppShipIt:
> My concern over ambiguity in the parser

I'm just thinking about user experience and potential surprises.

> you'd have to write (_ ? x) anyway.

Then `x ? y` would still mean something different than `x ?y`.

AppShipIt:
Just realized `x ? y` : `x ?y` :: `x - 1` : `x -1` i.e. not any more whitespace dependent than subtraction currently 😅

garyb:
@paf31 Yeah, I'm gonna go with the module-only option, it's getting pretty hairy to make it do the other thing!

src/Language/PureScript/Make.hs
```diff
@@ -221,15 +221,16 @@ make MakeActions{..} ms = do
                               _ -> True
 
         let rebuild = do
-              (exts, warnings) <- listen $ do
+              (exts, warnings) <- listen $ censor suppressHoleWarnings $ do
```
garyb: @paf31 attempting to censor the warnings here (to knock out the hole appearing as both warning and error) has no effect. In fact, using `mempty` instead of `warnings` down at `markComplete` seems to have no effect either. Do you know where it is that they're being printed from? And is there any point to `listen`ing to the warnings here if they're as good as unused?

paf31: I would just add a line [here](https://github.com/purescript/purescript/blob/dc58903b8a19af7cb44b63ead3295e6133b740f5/src/Language/PureScript/TypeChecker/Types.hs#L120) and then throw during codegen if you see the `Hole` constructor.

garyb:
I found a slightly different way around it: no need to supress the warning this time as I don't raise a warning in the first place.

I did the throwing from `typeCheckModule` rather than codegen still because currently `CoreFn` codegen is not operating in a `MonadError`, or any `Monad` for that matter.

paf31:
This looks fine, but I'd really like to reuse the machinery we have for wildcards, where we warn or throw, and then change the wrapped type when the substitution is known.

garyb:
I wondered about that, I tried constructing a few cases to see if I could get a bad type back, but it seemed that this was enough.

garyb:
How about this version then?

I had to pull stuff out of `liftUnifyWarnings` as that only allowed us to map over the errors, but unless we do want to monadify `CoreFn.Desugar` then extracting the warning and turning it into an error here needs a bit more access.

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -235,19 +235,11 @@ liftUnify ::
   (MonadState CheckState m, MonadWriter MultipleErrors m, MonadError MultipleErrors m) =>
```
paf31: `MonadWriter` is redundant here I think, now.

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -235,19 +235,11 @@ liftUnify ::
...
-liftUnify = liftUnifyWarnings (const id)
-
--- | Run a computation in the substitution monad, generating a return value, the final substitution and updating warnings values.
-liftUnifyWarnings ::
-  (MonadState CheckState m, MonadWriter MultipleErrors m, MonadError MultipleErrors m) =>
-  (Substitution -> ErrorMessage -> ErrorMessage) ->
-  m a ->
-  m (a, Substitution)
-liftUnifyWarnings replace ma = do
+liftUnify ma = do
```
paf31: Maybe let's rename this now. How about `withEmptySubstitution`?

paf31:
I think this looks really close, I just would like to discuss any possible alternatives to the current warning/error approach. Originally, I thought we'd warn on any holes, and then throw one error if any holes were present, but I suppose this is better from a UX point of view.

Another possibility is to just throw every time you see a `Hole`, and use `parU` to try to throw multiple errors, i.e. max one per binding group.

garyb:
Yeah, I figured if we promote all of the hole warnings to errors it makes them more obvious when the compiler errors out - if you have other warnings in there you don't have to go hunting through them to find the messages relating to holes.

Given that originally I wanted to compile "everything" before showing the hole errors, I'd definitely be apprehensive about restricting it way down to the binding group level :wink:. Also, wouldn't that prevent the type from being well inferred in some cases, the same way that not applying the substitution did? What's the motivation behind this idea, to make the code a bit simpler? I thought it was already pretty trivial, apart from having to substitute "manually".

I'm fine with changing it if there is a better approach though. It's always an adventure when I try to do anything in the typechecker, it seemed like this was a pretty straightforward way to make it happen.

garyb:
I've made the other suggested changes up to now anyway.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -74,7 +74,7 @@ typesOf ::
   [(Ident, Expr)] ->
   m [(Ident, (Expr, Type))]
 typesOf bindingGroupType moduleName vals = do
-  ((tys, subst), warnings) <- reflectErrors . censor (const mempty) . reifyErrors . listen $ liftUnify $ do
+  ((tys, subst), warnings) <- reflectErrors . censor (const mempty) . reifyErrors . listen $ withEmptySubstitution $ do
```
paf31: Does `withEmptySubstitution` need to go on the outside here? I just remember this code was pretty tricky to get right originally, so I want to make sure the three effects still get interleaved in the right way.

garyb: I'm not sure how to tell whether it's behaving correctly or not actually. What is the purpose of reify and then reflecting? The main thing I was concerned about here was "replacing" the warnings with `censor` and then `tell` just after.

paf31: It's something to do with types in error messages, but I forget the exact details. Somehow, the substitution was going missing in the presence of an error, so I had to do some manual interleaving of effects. It might be possible to do a better job by breaking up the typechecker state into different parts.

garyb:
Ok, that got flipped turned upside down now, so we have `withEmptySubstitution` on the outside. It definitely makes sense that it should be that way looking at where the reify etc. stuff was before.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -74,13 +74,17 @@ typesOf ::
...
 typesOf bindingGroupType moduleName vals = do
-  tys <- fmap tidyUp . liftUnifyWarnings replace $ do
+  ((tys, warnings), subst) <- withEmptySubstitution . reflectErrors . censor (const mempty) . reifyErrors . listen $ do
     (untyped, typed, dict, untypedDict) <- typeDictionaryForBindingGroup moduleName vals
     ds1 <- parU typed $ \e -> checkTypedBindingGroupElement moduleName e dict
     ds2 <- forM untyped $ \e -> typeForBindingGroupElement e dict untypedDict
     return (map (\x -> (False, x)) ds1 ++ map (\x -> (True, x)) ds2)
 
-  forM tys $ \(shouldGeneralize, (ident, (val, ty))) -> do
+  let (holeErrors, warnings') = partition isHoleError $ map (replace subst) $ runMultipleErrors warnings
```
paf31: I think I'd be fine with this way of treating warnings if we could just pull it out into a combinator. Something like 

```haskell
escalate 
  :: (MonadWriter [MultipleErrors] m, MonadError [MultipleErrors] m)
  => (ErrorMessage -> Bool)
  -> m a
  -> m a
```

garyb:
I think maybe we're there this time? :wink:

It definitely seems a fair bit cleaner this way. We're back to `liftUnifyWarnings` as the error promotion needs to `listen`, so it needs to be part of that chain of things handling the `do`.

paf31:
👍 🚢 Thanks for putting up with all of my comments, looks great!

garyb:
I think it was worth it, this version came out much better. Thanks!

bergmark:
I added a list item for typed holes on https://github.com/purescript/purescript/wiki/Differences-from-Haskell

PR #2024, opened 2016-04-16
Remove legacy operator sections

paf31:
For 0.9


paf31:
Rebased. I've also removed a `try` in the parser which I don't think is necessary any more.

PR #2027, opened 2016-04-17
Fix prettyprinter for types

garyb:
@paf31 this'll do it. I found out why I was having so many difficulties at last - docs has its own type pretty printer!

I'm not sure rebracketing during docs is the way to go (which is what I've done here) as it means the types may become uglier in the docs than they are in the actual code, but the pretty-printing is a bit funky without it, as quite a few extra sets of parens can get inserted. If we can figure out how to avoid that, rebracketing will be unnecessary.

Maybe the docs pretty printer for types could be made much dumber actually, if it's running where we still have `ParensInType` then I guess it doesn't need to insert its own parens to ensure rendering is correct?

### DIFF
--- a/src/Language/PureScript/Docs/Convert/ReExports.hs
+++ b/src/Language/PureScript/Docs/Convert/ReExports.hs
@@ -111,13 +111,14 @@ collectDeclarations imports exports = do
   valsAndMembers <- collect lookupValueDeclaration     impVals  expVals
   typeClasses    <- collect lookupTypeClassDeclaration impTCs   expTCs
   types          <- collect lookupTypeDeclaration      impTypes expTypes
+  typeOps        <- collect lookupTypeOpDeclaration    impTypeOps expTypeOps
 
   (vals, classes) <- handleTypeClassMembers valsAndMembers typeClasses
 
   let filteredTypes = filterDataConstructors expCtors types
   let filteredClasses = filterTypeClassMembers (map fst expVals) classes
 
-  pure (Map.toList (Map.unionsWith (<>) [filteredTypes, filteredClasses, vals]))
+  pure (Map.toList (Map.unionsWith (<>) [filteredTypes, typeOps, filteredClasses, vals]))
 
   where
   collect lookup' imps exps = do
@@ -129,6 +130,9 @@ collectDeclarations imports exports = do
 
   expTypes = map (first fst) (P.exportedTypes exports)
   impTypes = concat (Map.elems (P.importedTypes imports))
+
+  expTypeOps = P.exportedTypeOps exports
+  impTypeOps = concat (Map.elems (P.importedTypeOps imports))
 
   expCtors = concatMap (snd . fst) (P.exportedTypes exports)
 
@@ -180,7 +184,7 @@ lookupValueDeclaration importedFrom ident = do
   let
     rs =
       filter (\d -> declTitle d == P.showIdent ident
-                    && (isValue d || isAlias d)) decls
+                    && (isValue d || isValueAlias d)) decls
     errOther other =
       internalErrorInModule
         ("lookupValueDeclaration: unexpected result:\n" ++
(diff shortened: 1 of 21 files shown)


hdgarrood:
> docs has its own type pretty printer!

Yes, sorry, this is my fault IIRC. At some point, I'm hoping to combine them so that we just have one.

garyb:
Thinking about it, making the prettyprinter dumber won't work, as we run the typechecker to infer missing type annotations.

garyb:
@hdgarrood no problem, it totally makes sense that we need a different version for rendering the docs, I just hadn't realised it was there so was confused when my changes seemed entirely ineffectual :smile:

If we do unify them I guess the docs-based one is probably already more along the right lines, in that it doesn't produce text.

garyb:
Just pushed some more changes - think I have this figured out now, I'm actually using the pattern table to insert the parens, which prevents the multiple unnecessary nesting. The test types render as they should, and from experimenting it seems everything else does too:

--- 

## Module Test

#### `Natural`

``` purescript
type Natural f g = forall a. f a -> g a
```

#### `(~>)`

``` purescript
infixr 6 type Natural as ~>
```

_right-associative / precedence 6_

#### `test`

``` purescript
test :: forall f g. f ~> g
```

src/Language/PureScript/Docs/RenderedCode/Types.hs
```diff
@@ -159,7 +161,10 @@ syntax :: String -> RenderedCode
 syntax x = RC [Syntax x]
 
 ident :: String -> RenderedCode
-ident x = RC [Ident x]
+ident x = RC [Ident x ThisModule]
+
+ident' :: String -> ContainingModule -> RenderedCode
+ident' x m = RC [Ident x m]
```
garyb: @hdgarrood I didn't find the `ContainingModule` really being used anywhere in the code we have here, but I assume pursuit uses it to generate links between modules in the documentation? 

`Ident` didn't previously need this (we'd only show idents for the current module, but now an external ident can appear in the guise of a type operator). So I guess this will also be a breaking change for Pursuit.

hdgarrood: Yes, Pursuit does use `ContainingModule` to generate links between modules in the documentation. Changes to the `purescript` library are usually no problem for Pursuit, it's just breaking changes to the JSON format that you have to be aware of (which it seems this is). But we already have a breaking change to the JSON format lined up for the next release anyway, right?

hdgarrood:
How does `f ~> g ~> h` come out? Also, is it possible to add some tests for type operator bracketing?

src/Language/PureScript/Docs/RenderedCode/Types.hs
garyb: Yeah exactly. I just thought I'd mention this one too so it was on your radar, as it's probably a wider change than adjusting the sum type for aliases was.

src/Language/PureScript/Docs/Render.hs
```diff
@@ -70,12 +70,12 @@ renderDeclarationWithOptions opts Declaration{..} =
   renderType' = renderTypeWithOptions opts
   renderAlias (P.Qualified mn alias)
     | mn == currentModule opts =
-        P.foldFixityAlias P.runIdent P.runProperName P.runProperName alias
+        P.foldFixityAlias P.runIdent P.runProperName (("type " ++) . P.runProperName) alias
```
hdgarrood: Does this affect rendering of normal operator aliases?

hdgarrood: Oh, just seen the definition of `foldFixityAlias`. Never mind :)

garyb:
It renders as `f ~> g ~> h`. Bracketing generally works out fine: `f ~> (g ~> h) ~> i` renders correctly too.

I'll add some tests though, sure :smile:

src/Language/PureScript/Docs/Render.hs
garyb: Yeah there used to be an `either` here before I introduced a real sum type, having a fold seemed slightly more elegant than explicit pattern matching in this case. I just noticed when working on the other stuff that it was mis-rendering type operator aliases so fixed that while I was at it.

garyb:
Ok, there were some other problems in the docs with type operators which have now been fixed (re-exports related). Also, there were a bunch of places where we weren't using `readUTF8File`, which meant that on Windows at least the docs and publish executables failed when encountering source files with unicode operators and the like.

src/Language/PureScript/Docs/Convert/ReExports.hs
```diff
@@ -25,6 +25,8 @@ import qualified Language.PureScript as P
 
 import Language.PureScript.Docs.Types
 
+import Debug.Trace
```
hdgarrood: Were you intending to remove this?

garyb: Oops.

hdgarrood:
Looks great :+1:

PR #2028, opened 2016-04-17
Require types to be exported if an type operator alias is

garyb:
Another thing that should have been done in the original PR. Pretty sure this should be the last of it though, once the other prettyprint PR is merged too.


paf31:
👍 Looks good!

PR #2029, opened 2016-04-17
Relax licensing rules

hdgarrood:
See individual commits for a description of the changes.

### DIFF
--- a/src/Language/PureScript/Publish/ErrorsWarnings.hs
+++ b/src/Language/PureScript/Publish/ErrorsWarnings.hs
@@ -202,6 +202,12 @@ displayUserError e = case e of
           [ "Note that distributing code without a license means that nobody "
           , "will (legally) be able to use it."
           ])
+      , spacer
+      , para (concat
+          [ "It is also recommended to add a LICENSE file to the repository, "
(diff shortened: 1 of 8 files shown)


src/Language/PureScript/Publish/ErrorsWarnings.hs
```diff
@@ -182,11 +182,27 @@ displayUserError e = case e of
...
     displayRepositoryError err
   NoLicenseSpecified ->
-    para (concat
-      ["No license specified in bower.json. Please add one. ",
-       "Distributing code without a license means that nobody ",
-       "will be able to (legally) use it."
-      ])
+    vcat $
+      [ para (concat
+          [ "No license is specified in bower.json. Please add one, using the "
```
paf31: I would just add that we _recommend_ adding a LICENSE file, even if it isn't necessary.

paf31:
👍 Looks good, once CI passes

src/Language/PureScript/Publish/ErrorsWarnings.hs
hdgarrood: :+1:

PR #2033, opened 2016-04-18
Use readUTF8 file in IDE rebuild

### LINKED ISSUE #2031: IDE server not reading files as UTF8?
This occurs on Windows.

Yesterday I had to update a bunch of places to use `readUTF8File` rather than `readFile` for `psc-publish`, `psc-docs`, etc, but the IDE executables use `readFile` from `Data.Text`. I was under the impression that was already going to read them as UTF8, but perhaps not, as I get parse errors for any file using unicode symbols for things like `→` or `∷`.

garyb:
Resolves #2031


garyb:
@kRITZCREEK I think it will be treated as text again after this is merged, even though it's still not showing the diff. There was a `NUL` char in the comment on line 83.

The delete/restore commits didn't actually do anything, but you can see the diff works for them individually inspecting them, so I'll squash this on merge and hopefully it all work out alright!

kRITZCREEK:
Awesome thank you!

Am 18.04.2016 2:53 nachm. schrieb "Gary Burgess" <notifications@github.com>:
>
> @kRITZCREEK I think it will be treated as text again after this is
merged, even though it's still not showing the diff. There was a NUL char
in the comment on line 83.
>
> The delete/restore commits didn't actually do anything, but you can see
the diff works for them individually inspecting them, so I'll squash this
on merge and hopefully it all work out alright!

src/Language/PureScript/Ide/Rebuild.hs
```diff
@@ -0,0 +1,85 @@
...
+    Right (sorted, graph) -> do
+      let deps = fromJust (lookup (P.getModuleName m) graph)
+      pure $ mapMaybe getExtern (deps `inOrderOf` map P.getModuleName sorted)
+  where
+    mkShallowModule P.ExternsFile{..} =
+      P.Module undefined [] efModuleName (map mkImport efImports) Nothing
+    mkImport (P.ExternsImport mn it iq) =
+      P.ImportDeclaration mn it iq False
+    getExtern mn = M.lookup mn ex
+    -- Sort a list so its elements appear in the same order as in another list.
```
hdgarrood: Not that this should delay merging, but is this right? I don't think this will do what the comment says it will do; I'm fairly sure `filter` doesn't change the order of elements in a list.

garyb:
Hmm, still getting some unicode issues even with this. The file parses correctly but it has trouble with unicode operators:

```
[1/1 UnknownValue] src\SlamData\Notebook\Card\Save\Component.purs:67:47

  67              , HE.onValueInput $ HE.input \s → right ∘ UpdatePathString s
                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^

  Unknown value (∘)
```

I'll have to look into it more later.

src/Language/PureScript/Ide/Rebuild.hs
kRITZCREEK: I didn't put too much thought in it and took it from:
https://github.com/purescript/purescript/blob/8bab5090ce3559d89c10c34e240356a5b371a489/src/Language/PureScript/Make.hs#L223-L225

It does feel like a hack to me so I didn't want to pull it out into an API function. It actually doesn't work without so it must change *something* :D

garyb:
Hmm, so there's some kind of encoding issue still. For example., the operator `×` should be `"\215"` but is coming through as `"\9500\249"`. I'm not sure exactly where this occurs yet, but it's why the the second case is failing.

garyb:
With some additional help from @kRITZCREEK everything is now working as it should on Windows with this patch. :dancer:

paf31:
👍

Issue #2039, opened 2016-04-19
Bad results from the expression pretty printer


LiamGoodacre:
The following appears in the error message for the expression: `(let o = {a: 0} in o).b`

``` purescript
(let           ."b"
   o = { a: ...
       }
 in o
)
```

Firstly the `"b"` is quoted.  Secondly it's in the wrong place :smiley:

natefaubion:
Its not in the wrong place if you think of the expressions as being side by side :laughing:

AppShipIt:
Ever the optimist 😂

garyb:
Which is exactly what I think is happening :wink: `<>` in boxes joins things top & horizontally I think.

LiamGoodacre:
Woah, quoting is somewhat correct? I didn't know this was valid: `_."a m a z i n g"` !

garyb:
Yeah, it's the PS equivalent to `obj["foo"]` in JS, only it's restricted to literal strings, not expressions. It's so you can have record fields with keyword names and the like. The record syntax is `{ "s i m i l a r": true }`.

LiamGoodacre:
Noticed some other oddities with the quoting - PR incoming :smiley:

paf31:
We have a combinator for this sort of layout, I think.

nwolverson:
Side by side rendering is really confusing.

https://github.com/nwolverson/atom-ide-purescript/issues/112 (search 'psci')

PR #2043, opened 2016-04-19
Fix some pretty printing issues

LiamGoodacre:
Partially related to #2039

### Old pretty printing behaviour:

**Some keys aren't quoted when they should be:**

~~~purescript
> :t _."a b"
forall t2 t3.
  { a b :: t2                 -- key contains space, should be quoted
  | t3
  }
  -> t2
~~~

**Accessors are always quoted, even when not necessary:**

~~~purescript
> {}.a
Error found:
in module $PSCI
...roperty accessor {}."a"    -- no need for this one to be quoted
~~~

### New pretty printing behaviour:

~~~purescript
> {}.a

-- in record
{ a :: _0
| _1
}

-- as accessor
{}.a
~~~

~~~purescript
> {}."a b"

-- in record
{ "a b" :: _0
| _1
}

-- as accessor
{}."a b"
~~~

### DIFF
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -69,12 +69,13 @@ module Language.PureScript.Parser.Lexer
   , reservedPsNames
   , reservedTypeNames
   , isSymbolChar
+  , isUnquotedKey
   )
   where
 
 import Prelude hiding (lex)
 
-import Data.Char (isSpace, isAscii, isSymbol)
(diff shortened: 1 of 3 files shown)


LiamGoodacre:
Hmm, I guess the test for whether or not to quote should be if it is required when you type it.  Which means this needs to be slightly different.  For example the key `}` doesn't get quoted.

LiamGoodacre:
The integration builds are mostly failing due to an import warning - I'll fix it tonight.

garyb:
:+1: aside from that!

paf31:
👍  Thanks!

PR #2044, opened 2016-04-20
Fix #2042, use JSON for parser errors in rebuild

### LINKED ISSUE #2042: [psc-ide] Rebuild parse error should be JSON
https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Ide/Rebuild.hs#L40

Regular compiler parse errors are output as JSON, I feel this should be similar, avoid having to parse the location (imperfect as it may be) out of a string

paf31:


kRITZCREEK:
👍

PR #2045, opened 2016-04-20
Error on missing Partial constraint

paf31:
This needs some more work, namely to fix up the test suite, but it's pretty close.

This will warn on any non-exhaustive function, but now also error if a Partial dictionary is not in scope.


paf31:
We probably shouldn't show the warning if the `Partial` constraint is there.

paf31:
Another option is to implement something like programmable type errors.

Issue #2026, opened 2016-04-17
Some issues with the released binaries


jplatte:
I'm the maintainer of the [purescript-bin arch linux package][aur]. I just re-ran namcap (a package inspection tool) on the latest build to confirm the errors it got for the last release – I mentioned those on IRC but they seem to not have been seen by the right people.

I left out some messages because they seem to be bugs in namcap, but there are quite some that seem to come from issues with the binaries:

```
purescript-bin E: Insecure RPATH '/home/travis/build/purescript/purescript/.cabal-sandbox/lib/x86_64-linux-ghc-7.10.3/purescript-0.8.4.0-EVFfWBi1t9N9UoZBf84XL6' in file ('usr/bin/psc')
purescript-bin E: Insecure RPATH '/home/travis/build/purescript/purescript/.cabal-sandbox/lib/x86_64-linux-ghc-7.10.3/purescript-0.8.4.0-EVFfWBi1t9N9UoZBf84XL6' in file ('usr/bin/psc-bundle')
purescript-bin E: Insecure RPATH '/home/travis/build/purescript/purescript/.cabal-sandbox/lib/x86_64-linux-ghc-7.10.3/purescript-0.8.4.0-EVFfWBi1t9N9UoZBf84XL6' in file ('usr/bin/psc-docs')
purescript-bin E: Insecure RPATH '/home/travis/build/purescript/purescript/.cabal-sandbox/lib/x86_64-linux-ghc-7.10.3/purescript-0.8.4.0-EVFfWBi1t9N9UoZBf84XL6' in file ('usr/bin/psci')
purescript-bin E: Insecure RPATH '/home/travis/build/purescript/purescript/.cabal-sandbox/lib/x86_64-linux-ghc-7.10.3/Glob-0.7.5-J8eYhnaKSeW8dehUANn0F1' in file ('usr/bin/psc-ide-client')
purescript-bin E: Insecure RPATH '/home/travis/build/purescript/purescript/.cabal-sandbox/lib/x86_64-linux-ghc-7.10.3/purescript-0.8.4.0-EVFfWBi1t9N9UoZBf84XL6' in file ('usr/bin/psc-ide-server')
purescript-bin E: Insecure RPATH '/home/travis/build/purescript/purescript/.cabal-sandbox/lib/x86_64-linux-ghc-7.10.3/purescript-0.8.4.0-EVFfWBi1t9N9UoZBf84XL6' in file ('usr/bin/psc-publish')
[...]

PR #2046, opened 2016-04-20
Add --as-needed linker option to stop linking against unused shared objects

jplatte:
See #2026

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -244,6 +244,7 @@ executable psc
     hs-source-dirs: psc
     other-modules: Paths_purescript
     ghc-options: -Wall -O2 -fno-warn-unused-do-bind -threaded -rtsopts "-with-rtsopts=-N"
+    ld-options: -Wl,--as-needed
 
 executable psci
     build-depends: base >=4 && <5, containers -any, directory -any, filepath -any,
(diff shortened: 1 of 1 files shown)


hdgarrood:
:+1: from me (assuming CI passes).

paf31:
Thanks!

nwolverson:
This doesn't build for me now 
```
Linking .stack-work/dist/x86_64-osx/Cabal-1.22.5.0/build/psc/psc ...
ld: unknown option: --as-needed
```

paf31:
Ok, if we can't figure out why, I'll revert this.

jplatte:
@nwolverson The important thing is which linker you are using. I assume your ld is not the GNU linker?

hdgarrood:
Same here, OSX:

```
$ ld -v
@(#)PROGRAM:ld  PROJECT:ld64-264.3.101
configured to support archs: armv6 armv7 armv7s arm64 i386 x86_64 x86_64h armv6m armv7k armv7m armv7em (tvOS)
LTO support using: LLVM version 7.3.0
```

nwolverson:
```
@(#)PROGRAM:ld  PROJECT:ld64-241.8
configured to support archs: armv6 armv7 armv7s arm64 i386 x86_64 x86_64h armv6m armv7m armv7em
LTO support using: LLVM version 3.5svn
```

hdgarrood:
Come to think of it, this seems to me like something that should be addressed in Cabal, really.

jplatte:
Yeah, it's probably better to revert my commit for now.

Before we try to make this the default in cabal or anything like that though, we should try to find out why it's not the default in the GNU linker. Hopefully it's just some sort of backwards compatibility and not a thing where the linker produces broken executables.

Issue #1702, opened 2015-12-11
Minimal source spans


natefaubion:
Right now, source span tracking is greedy, and it will list the span of a term as all the way up to where the next term begins. It would be nice to have the minimal span (ie. removing white space from the span).

paf31:
:+1:

natefaubion:
It seems like this is mainly due to https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Parser/Lexer.hs#L214

Would it make sense to eat the whitespace in `parsePositionedToken` instead of in `parseToken`?

paf31:
Yes, that's very possible. Worth a try anyway.

natefaubion:
Unfortunately its not that easy :|

The token parser just gets the start position, then in the AST parser it uses the start position, and then whatever position the parser has advanced to as the end position. Then maybe you could use the end position of the _most recently consumed_ token rather than the start position of the next token. I don't know if Parsec supports something like that.

PR #2052, opened 2016-04-24
Tighten source-spans to token end position

nwolverson:
An attempt at #1702. Don't seem to get much help from parsec here, this is bit of a fudge - and (at least?) the Parsec parsers used for string & number literals still consume whitespace (`lexeme` itself eats whitespace). I don't see any way around that other than not using `LanguageDef`.

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -52,7 +52,11 @@ withSourceSpan f p = do
   comments <- C.readComments
   x <- p
   end <- P.getPosition
-  let sp = SourceSpan (P.sourceName start) (toSourcePos start) (toSourcePos end)
+  input <- P.getInput
+  let end' = case input of
+        pt:_ -> ptPrevEndPos pt
+        _ -> Nothing
+  let sp = SourceSpan (P.sourceName start) (toSourcePos start) (toSourcePos $ fromMaybe end end')
   return $ f sp comments x
 
 kindedIdent :: TokenParser (String, Maybe Kind)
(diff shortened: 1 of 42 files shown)


hdgarrood:
How easy is it to add tests for this?

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -153,6 +153,8 @@ prettyPrintToken (HoleLit name)    = "?" ++ name
 
 data PositionedToken = PositionedToken
   { ptSourcePos :: P.SourcePos
+  , ptEndPos :: P.SourcePos
+  , ptPrevEndPos :: Maybe P.SourcePos
```
paf31: Could you please add some Haddock comments to these fields?

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -281,6 +285,13 @@ parseModulesFromFiles toFilePath input = do
   inParallel :: [Either P.ParseError (k, [Module])] -> [Either P.ParseError (k, [Module])]
   inParallel = withStrategy (parList rseq)
 
+  updatePositions :: [PositionedToken] -> [PositionedToken]
```
paf31: Why not put this function in the lexer module, or even fuse it into the lexer function?

nwolverson:
On tests, I think specific parser tests could be done invoking `parseModules` and checking the results, maybe without external files. Not sure what's appropriate here, that's a bit different from existing tests? Input required.

paf31:
👍 LGTM, thanks!

nwolverson:
Squashed and rebased to 0.9. Would you like me to open a new PR against 0.9 or will you merge manually?

paf31:
Yes please.

nwolverson:
https://github.com/purescript/purescript/pull/2077

PR #2053, opened 2016-04-24
Type suggestions

nwolverson:
2 new suggestions here for editor support etc.
- `MissingTypeDeclaration` - adding a separate source range in the JSON suggestion field to support this, I think that's likely something that could be used elsewhere. Idea being the replacement range may not coincide with the warning/error range
- `TypeWildCard` - Main thing here is this needed a source position for the wildcard itself rather than the whole thing. I tried a general `PositionedType` constructor as per `PositionedValue` but ended up scattering junk all over the typechecker without managing to make it work. I imagine types may want detailed error positions in the future obsoleting some of this, but I can't work out the details. This looks good with psa now.

### DIFF
--- a/examples/failing/OverlappingReExport.purs
+++ b/examples/failing/OverlappingReExport.purs
@@ -7,4 +7,4 @@ module B where
 
 module C (module A, module M2) where
   import A
-  import qualified B as M2
+  import B as M2
--- a/examples/failing/SuggestComposition.purs
+++ b/examples/failing/SuggestComposition.purs
@@ -4,4 +4,4 @@ module SuggestComposition where
 
 import Prelude
 
-f = g . g where g = (+1)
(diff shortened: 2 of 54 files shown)


paf31:
Could you please rebase this and your other PR against the 0.9 branch? I'm going to use `master` for emergency 0.8 fixes until 0.9 is released. Thanks!

nwolverson:
Rebased on 0.9. As per other PR let me know if you want to open a fresh PR.

LiamGoodacre:
@nwolverson I believe we want new PRs on the 0.9 branch - see #2067

paf31:
Yes, I think you need to open a fresh PR. Thanks!

nwolverson:
https://github.com/purescript/purescript/pull/2076

PR #2054, opened 2016-04-25
Work in progress: parse module header

paf31:
WIP. I'm still experimenting here, but the main thing remaining is to make sure we're as lazy as possible in the list of fully parsed modules. Right now, laziness is being killed by the use of `Either`.


paf31:
Compared to `master`, this improves the rebuild time of Thermite on my machine from 0.8s to about 0.6s, and the rebuild-with-no-changes time from 0.6s to 0.4s.

I think there is more room for optimization in the lexer, especially if we switch to `Text`, and maybe away from `parsec`.

garyb:
I'd love to try this on SlamData, as it suffers from a long rebuild even with _no_ changes, nevermind an incremental build, but unfortunately I imagine the rules will be broken somewhere, either in the project itself or one of the many dependencies.

paf31:
Oh it definitely breaks a whole bunch of core libraries.

garyb:
I'm working on some module name related things that may help with this (I'll disallow fully qualified references without `import`, for instance - I'm not touching the import syntax right now, just the behaviour), we may have some conflicts as I guess I might be duplicating work, but if you're not in progress with this right now it might be worth it.

One thing I'm doing is making it so we can actually have multiple modules per test without them being in separate files, which is something this will need too if we want to keep all our current tests working. I'm thinking something like this: every test still has at least one purs file, but if it requires additional modules we can provide a subdirectory with the same name, and any modules within there will be included in the path when that test module is run.

garyb:
I just rebased this on 0.9 and fixed up the tests to avoid using `Control.Monad.Eff.Console.log`. Maybe the PR needs flipping over to 0.9 too? Also there is one test that fails: an empty class declaration parses that should not.

rvion:
has this been abandoned ?

garyb:
I haven't spoken to Phil about it, but I'd assume not. This branch is seriously out of date and parts of it have been merged already, so it was probably a bit of a mess to deal with, and will be easy enough to just re-apply the functionality changes to where `master` is at.

Issue #1249, opened 2015-07-10
Optimise "pass-through" cases


garyb:
The code generated for something like:
``` purescript
f (Prop e) = Prop e
```
Should be equivalent to:
``` purescript
f p = p
```
Even if it wouldn't have type checked had it been written that way (think of a case like this in a `Functor` implementation, for example).

rightfold:
Started working on this: https://github.com/rightfold/purescript/commit/8d827f140630ff146c8be2ce85aad8f23b6e4fd1 :+1:

garyb:
Yeah, it could definitely be applied there. We could do it in `CoreFn` too though, as the trick here is the type coercion, and we typecheck before producing the `CoreFn` AST.

rightfold:
I'll take a look at that. Are optimisations already being performed on the `CoreFn` data structure? If so, in which modules does this take place? Or does this happen directly during the phase that generates `CoreFn`? The only optimisation modules I can find are JS-specific.

rightfold:
Ok, I rewrote it to work on CoreFn instead of JS: https://github.com/rightfold/purescript/commit/0fb74c4a88e42517181d8dfbf074aa0567ec09f1

garyb:
As you probably saw, we don't have an optimiser pass for `CoreFn` yet, but it was always the intention that we'd introduce something, as that's the reason `CoreFn` exists at all (that and the rewrites that we don't have yet). Your branch there looks pretty good to me, that's exactly what I would have done to introduce `CoreFn` optimisations. I've not looked at the actual optimisation too closely yet, if you open a PR we can review it properly.

Thanks for kicking this off!

rightfold:
No problem. I enjoy it. :)

PR #2057, opened 2016-04-26
Optimize pass-through cases

rightfold:
See #1249.

The following points are still left to do but may be fixed in a later version. They are not required for correct functioning of this optimisation; they just catch more cases:
- Nested patterns (i.e. `case ... of C (D x) -> C (D x)`).
- Multiple expressions being matched against (i.e. `case a, b, c of ...`).
- [ ] Case alternatives involving guards.
- [ ] Find fixed point in CoreFn optimization driver.
- [x] Nested expressions (e.g. `case ... of C x -> f (C x)`).

I doubt the first two cases come up more often than rarely, so I'm leaving them for what they are right now (hence no checkboxes).

The following points may need attention:
- [x] Use `Test.Assert` in the unit test? Not sure whether that's what that module is in tests/support for.
- [x] Suspicious use of `disqualify` and `runProperName`. I am not fully aware of the way names work in psc, so I'll have to figure this out first.

### DIFF
--- a/src/Language/PureScript/CoreFn/Traversals.hs
+++ b/src/Language/PureScript/CoreFn/Traversals.hs
@@ -4,9 +4,7 @@
 --
 module Language.PureScript.CoreFn.Traversals where
 
-import Control.Arrow (second, (***), (+++))
 import Data.Functor.Identity (runIdentity)
-import Language.PureScript.AST.Literals
 import Language.PureScript.AST.Literals
 import Language.PureScript.CoreFn.Binders
 import Language.PureScript.CoreFn.Expr


paf31:
Could you please update `CONTRIBUTORS.md`?

src/Language/PureScript/CoreFn/Optimizer.hs
```diff
@@ -0,0 +1,19 @@
...
+module Language.PureScript.CoreFn.Optimizer (optimize) where
+
+import Control.Monad.Reader (asks, MonadReader)
+import Language.PureScript.CoreFn.Ann (Ann)
+import Language.PureScript.CoreFn.Module (Module)
+import Language.PureScript.CoreFn.Optimizer.PassThroughCases
+import Language.PureScript.Options (Options, optionsNoOptimizations)
+
+-- |
+-- Apply a series of optimizer passes to CoreFn
```
paf31: Could you please list the optimizations here, even if it's just the one for now?

src/Language/PureScript/CoreFn/Optimizer/PassThroughCases.hs
```diff
@@ -0,0 +1,81 @@
+-- |
+-- Optimizer step for simplifying "pass-through cases", for example:
```
paf31: Could you please explain here why this optimization is not possible elsewhere (types) and why it is safe?

src/Language/PureScript/CoreFn/Optimizer/PassThroughCases.hs
```diff
@@ -0,0 +1,81 @@
...
+  onExpr :: Expr Ann -> Expr Ann
+  onExpr (Case ss ts cs) = Case ss ts (map optimize cs)
+  onExpr e = e
+
+  optimize :: CaseAlternative Ann -> CaseAlternative Ann
+  optimize (CaseAlternative [bndr@(ConstructorBinder bndrAnn _ ctor prms)] (Right body))
+    | isReconstruction ctor prms body =
+        let (_, comments, type_, _) = bndrAnn
+            varAnn = (Nothing, comments, type_, Nothing)
+         in CaseAlternative [NamedBinder bndrAnn v bndr] (Right $ Var varAnn qv)
```
paf31: `qv` might shadow something here. Could you use `MonadSupply`?

src/Language/PureScript/CoreFn/Optimizer/PassThroughCases.hs
```diff
@@ -0,0 +1,81 @@
...
+  optimize a = a
+
+  isReconstruction :: Qualified (ProperName 'ConstructorName)
+                   -> [Binder Ann]
+                   -> Expr Ann
+                   -> Bool
+  isReconstruction ctor prms body =
+    case dissectConstruction body of
+      Just (ctor', args) ->
+        Ident (runProperName $ disqualify ctor) == disqualify ctor'
```
paf31: I think you need to compare module names here too.

src/Language/PureScript/CoreFn/Optimizer/PassThroughCases.hs
```diff
@@ -0,0 +1,81 @@
...
+  qv :: Qualified Ident
+  qv = Qualified Nothing v
+
+
+-- |
+-- Return the constructor and arguments from what looks like a data constructor
+-- call.
+--
+dissectConstruction :: Expr Ann -> Maybe (Qualified Ident, [Expr Ann])
+dissectConstruction e = (,) <$> findConstructor e <*> findArguments e
```
paf31: I think you can do this in one pass. In fact, I think it exists elsewhere, but I forget the name.

tests/support/flattened/utility.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+// module Main
+
+exports.same = function(a) {
```
paf31: I think this should probably go in `Test.Assert`.

paf31:
Thanks very much for working on this!

src/Language/PureScript/CoreFn/Optimizer/PassThroughCases.hs
rightfold: @paf31 `unApp` works on the JavaScript AST. I see how it's tail-recursive and single-pass though, so I'll use a similar algorithm for `dissectConstruction`. Thanks.

src/Language/PureScript/CoreFn/Optimizer/PassThroughCases.hs
rightfold: Gah, `everywhereOnValues` isn't monadic; this might take some work.

rightfold: Ok, implemented this.

tests/support/flattened/utility.js
rightfold: It's quite a low-level function that I don't think is useful enough to distribute with `Test.Assert`. It's also quite unreliable as more optimisations get added later.

src/Language/PureScript/CoreFn/Optimizer/PassThroughCases.hs
rightfold: Good catch.

tests/support/flattened/utility.js
garyb: Agreed. It's testing reference equality, so not a property we'd usually want to care about. There is `refEq` in the prelude for the `Prim` types, but it's not exported.

rightfold:
Ok, I've fixed those points @paf31.

src/Language/PureScript/Make.hs
```diff
@@ -158,16 +158,17 @@ rebuildModule MakeActions{..} externs m@(Module _ _ moduleName _ _) = do
...
   checkExhaustiveModule env' checked
   regrouped <- createBindingGroups moduleName . collapseBindingGroups $ elaborated
   let mod' = Module ss coms moduleName regrouped exps
       corefn = CF.moduleToCoreFn env' mod'
-      [renamed] = renameInModules [corefn]
+  (optCoreFn, nextVar2) <- runSupplyT nextVar1 $ CF.optimize corefn
+  let [renamed] = renameInModules [optCoreFn]
       exts = moduleToExternsFile mod' env'
-  evalSupplyT nextVar . codegen renamed env' . BU8.toString . B.toStrict . encode $ exts
+  evalSupplyT nextVar2 . codegen renamed env' . BU8.toString . B.toStrict . encode $ exts
```
garyb: Very minor thing, but we tend to prime new versions of a value, whereas numbering might indicate semantically similar values that are used in different places.

rightfold: OK!

src/Language/PureScript/CoreFn/Traversals.hs
```diff
@@ -13,34 +14,59 @@ everywhereOnValues :: (Bind a -> Bind a) ->
...
   h' b = h b
 
-  handleCaseAlternative ca =
-    ca { caseAlternativeBinders = map h' (caseAlternativeBinders ca)
-       , caseAlternativeResult = (map (g' *** g') +++ g') (caseAlternativeResult ca)
-       }
+  handleCaseAlternative ca = do
+    caseAlternativeBinders' <- mapM h' (caseAlternativeBinders ca)
+    caseAlternativeResult' <- case caseAlternativeResult ca of
+      Left guardeds -> Left <$> mapM (\(grd, e) -> (,) <$> g' grd <*> g' e) guardeds
```
garyb: You could use `pairM` from `Language.PureScript.Traversals` here I think, like we do in the `AST` traversals:

https://github.com/purescript/purescript/blob/5aaf615decfe20a84be051e118260ded4743f2b0/src/Language/PureScript/AST/Traversals.hs#L218-L222

rightfold:
As for this point:

> - Nested expressions (e.g. `case ... of C x -> f (C x)`).

This has an issue with shadowing. While easy to overcome, many optimisations will have to deal with it. Is it an idea to perform a global alpha renaming step across all of non-global bindings in CoreFn before doing any optimisations, such that each binding has a unique name?

paf31:
Actually, if this pass runs after the `Renamer` pass, we might be okay. I didn't think of that.

rightfold:
Does that pass already exist? If it does then I can extend this optimization to recurse into case alternative bodies.

paf31:
Yes, have a look at `Renamer`. I just don't remember the order they run in.

rightfold:
👍🏻

rightfold:
Apparently renamer runs on the AST, not CoreFn, so it obviously runs before this optimization.

garyb:
Could you rebase this please? I'd love to have this in!

rightfold:
@garyb My test case needs an FFI function. Where must I put it? In the past it was in tests/support/flattened but that directory is gone now

garyb:
Just add a .js file with the same name as the .purs test file.

paf31:
What's the status of this? It seems like it was pretty close to complete.

paf31:
Ping

rightfold:
I totally forgot about this, I think it doesn't merge anymore so it needs some changes to avoid merge conflicts, I'll take a look at it tomorrow.

Sent from my iPhone.

paf31:
Closing as potentially abandoned for now, but please do reopen if that's not the case. Thanks!

Issue #1620, opened 2015-11-10
Add test-compile command for unit testing


kRITZCREEK:
For unit tests you usually want to test all functions in a module, not just the ones exported.

The usual advice in Haskell is to write your functions inside an `Internal` module, test that, and hope users don't import it. This isn't great as you now have to jump back and forth between different files that are concerned with the same piece of logic.

My Proposal: Add a test-compile command in which export lists are ignored for a specified set of modules so that you can import and test them freely in your unit tests.

paf31:
This is also useful in PSCi, I think. See `:load` in GHCi.

hdgarrood:
What about if you put everything in the Internal module which contains all the code and exports everything, and then created another "external" module which only imports the Internal module and selectively re-exports the bits that should be public?

paf31:
Closing, because this seems like something better implemented in tooling, e.g. Pulp.

PR #2058, opened 2016-04-27
[psc-ide] Cache last rebuild

kRITZCREEK:
This is supposed to allow us to access private values of a module for completions inside that module.

If we recently rebuilt the module in question we can cache a version that opens up the exports for the module.

This is still very much a work in progress but I just put it into a PR to gather early feedback or ideas. This would also be a way to attack the issue outlined in #1620


kRITZCREEK:
Works fine for my few minimal testcases. Too bad hackage is down right now... If we were using stack for the builds we'd be fine.... just saying :D

kRITZCREEK:
Not really sure why the CI failed but it might be related to the hackage outage we had a few days ago. The one thing I'd really like to figure out for this PR is when to invalidate the rebuild cache. I'd like to minimize the amount of false positives we give.

kRITZCREEK:
retriggering the build

kRITZCREEK:
I'd like to get some input on this. 

I'm not sure about the API. Basically after every successful rebuild, we have the chance to recompile the module with open exports and store it inside the server, so it can be used for autocompletion and type lookup. This will also add an optional module field to the completion and type commands.

But we could also write the externs file, js file, and foreign file that we get this way to some point on the filesystem, and enable people to write unit tests for functions that are not exported by requiring these files.

While writing this I realised that this is not specific to psc-ide either and we could do the same with a normal compiler run. We can however only do it for one module at a time, because opening up/removing export lists will remove reexports and might cause nameclashes in modules that depend on the opened module, but looking at the speed of rebuilds this should be a tolerable overhead.

kRITZCREEK:
Okay after talking about this with @hdgarrood and @nwolverson I agree that writing anything to disc that potentially allows accessing private module members is way out of scope of this PR so I'd say this is a topic for another day.

I think this is ready for review. If you can think of a way to make reviewing changes like this easier let me know. I can offer to implement the functionality in my psc-ide-emacs fork but I don't think anyone but me is using emacs?

kRITZCREEK:
see #2083

Issue #2056, opened 2016-04-25
Quote types in type errors?


hdgarrood:
If you refer to the `Error` type without importing it first, then you get this:

```
Unknown type Error
```

which is easy to misread as "a type error occurred which was unknown to the compiler" rather than "your code refers to an unknown type called 'Error'". Perhaps we can make this distinction a little clearer? `Unknown type: Error` or `Unknown type 'Error'` perhaps?

hdgarrood:
Same for `UnknownValue` too.

soupi:
I littered the code parts of the errors with backticks. [commit here](https://github.com/soupi/purescript/commit/92e79592c3729375477af7addc366542c93a8d79).
You can try to run `stack test` and see if the errors are to your liking :)

AppShipIt:
Naively it seems like their ought to be a common formatting function?

soupi:
yeah, i'll make the changes when i have time today. :)

paf31:
This is fixed now that we color types in errors.

PR #2060, opened 2016-04-28
Marking code with backticks in error messages

soupi:
Implementing the request at #2056, is this what we might want?

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -522,11 +522,11 @@ prettyPrintSingleError full level showWiki e = flip evalState defaultUnknownMap 
       skolemInfo :: (String, Int, Maybe SourceSpan) -> Box.Box
       skolemInfo (name, s, ss) =
         paras $
-          line (name ++ show s ++ " is a rigid type variable")
+          line ("`" ++ name ++ show s ++ "` is a rigid type variable")
           : foldMap (return . line . ("  bound at " ++) . displayStartEndPos) ss
 
       unknownInfo :: Int -> Box.Box
-      unknownInfo u = line $ "_" ++ show u ++ " is an unknown type"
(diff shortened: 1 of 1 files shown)


paf31:
👍

hdgarrood:
👍 from me too.

hdgarrood:
Actually I wonder if we could make it a little more type-safe by having `markIdent :: Ident -> String` and `markProperName :: ProperName a -> String`, so that you can't accidentally apply either of the `mark` functions to something that isn't code?

hdgarrood:
That way, you only need to call one function, too; `markIdent ident` as opposed to `markCode (runIdent ident)`.

hdgarrood:
Sort of related, should the following two constructors have `Ident` arguments instead of `String`? https://github.com/purescript/purescript/blob/5aaf615decfe20a84be051e118260ded4743f2b0/src/Language/PureScript/Errors.hs#L128-L129

AppShipIt:
What @hdgarrood said (I just didn't have as clear an idea how to fix it) 👍

hdgarrood:
Also, I don't know what the right answer is here, but it might be worth considering putting the backticks directly into `typeAsBox` instead of applying this new `withBoxBackTicks` function everywhere, so that we don't have to worry about accidentally forgetting them whenever we modify functions in this module.

garyb:
Re: the above suggestion - I'm not sure every name shown in errors comes through with an `Ident` or `ProperName` type. I think `Sugar.Names` does some stuff with strings to avoid having to specify many different error messages that differ by one word.

hdgarrood:
Ok, then this LGTM.

paf31:
Could you please reopen this as a PR on the 0.9 branch?

paf31:
Thanks!

PR #2067, opened 2016-04-30
Merge 0.9 into master

paf31:
This branch will be used to collect changes for 0.9, just in case we need to make a 0.8.6 in the meantime.

Alternatively, we can choose to merge some open PRs into `master` instead, and if we get enough before 0.9, make a 0.8.6 release.

@garyb Is your library changes PR ready to merge? I wasn't sure, but I can do the merge if it is (you might want to squash commits first though).


garyb:
Perhaps we should close this for now to prevent the Travis build queue filling up with changes when things get merged?

paf31:
Yeah, good point 😄

PR #2068, opened 2016-04-30
Make as-patterns bind less tightly

### LINKED ISSUE #2049: Pattern matching on aliases doesn't work well with as-patterns
```purescript
data List a = Cons a (List a) | Nil

infixr 6 Cons as :

f :: List { x :: Int, y :: Int } -> Int
f ( r@{ x } : _) = x + r.y

  Could not match type
  
    Object
  
  with type
  
    List
  
  while trying to match type { y :: Int
                             | _0
                             }
    with type List
                { x :: Int
                , y :: Int
                }
  while checking that expression r
    has type { y :: Int
             | _0
             }
  while checking type of property accessor r.y
  in value declaration f
  
  where _0 is an unknown type
```

whereas changing the definition of f to `f ( (r@{ x }) : _) = x + r.y` works fine.

paf31:
Fixes #2049

### DIFF
--- a/examples/passing/2049.purs
+++ b/examples/passing/2049.purs
@@ -1,0 +1,13 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+
+data List a = Cons a (List a) | Nil
+
+infixr 6 Cons as :
+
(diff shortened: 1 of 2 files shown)


garyb:
Looks good :+1:

PR #2071, opened 2016-04-30
Always reload Externs

kRITZCREEK:
* The current usage of psc-ide involves loading all modules anyway, so
  we don't worry about loading previously unloaded modules.

* Don't attempt to reload deleted files

This could lead to surprising behaviour when you're managing the loaded modules manually. Since all the editor plugins have switched to just loading all the modules upfront anyway I don't think it's too big of a problem.

This change will cause psc-ide to reload all externs files that get added or changed inside the output folder, thus picking up newly added sourcefiles (through bower or by creating a new module) when they get compiled for the first time.

### DIFF
--- a/src/Language/PureScript/Ide/Watcher.hs
+++ b/src/Language/PureScript/Ide/Watcher.hs
@@ -12,44 +12,37 @@
 -- File watcher for externs files
 -----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


paf31:
👍 LGTM

paf31:
Thanks!

PR #2072, opened 2016-05-01
Require Partial instance on partial functions

paf31:
This needs a bit more work. Even though it works, the error message is much less helpful than the warning message, and we shouldn't show the warning if the `Partial` constraint can be satisfied.


paf31:
Ok, I think I have this working now.

paf31:
I think the PR build is failing due to stack caching the results of other PRs.

paf31:
@hdgarrood @garyb Could you please review this when you get a few minutes? Not necessarily the code (but that'd be good) but the general approach with errors and warnings.

Thanks!

examples/passing/PartialFunction.purs
```diff
@@ -2,9 +2,11 @@ module Main where
...
+import Control.Monad.Eff
+import Control.Monad.Eff.Console
 
-fn :: Number -> Number
+fn :: Partial => Number -> Number
 fn 0.0 = 0.0
 fn 1.0 = 2.0
 
-main = assertThrows $ \_ -> fn 2.0
+main = log "Done"
```
hdgarrood: What do you think about the following?

```purescript
do
  assertThrows \_ -> fn 2.0
  log "Done"
```

src/Language/PureScript/Types.hs
```diff
@@ -109,12 +107,30 @@ data Type
   | ParensInType Type
   deriving (Show, Read, Eq, Ord)
 
--- |
--- A typeclass constraint
---
-type Constraint = (Qualified (ProperName 'ClassName), [Type])
+-- | Additional data relevant to type class constraints
+data ConstraintData
+  = PartialConstraintData [[String]] Bool -- ^ a constraint generated by the exhaustivity checker
```
hdgarrood: Do you think you could perhaps make the comments around this data declaration a little clearer? What do the arguments mean, for instance?

src/Language/PureScript/Types.hs
```diff
@@ -109,12 +107,30 @@ data Type
...
+
+-- | A typeclass constraint
+data Constraint = Constraint
+  { constraintClass :: Qualified (ProperName 'ClassName)
+  -- ^ constraint class name
+  , constraintArgs  :: [Type]
+  -- ^ type arguments
+  , constraintData  :: Maybe ConstraintData
+  -- ^ additional data relevant to this constraint
+  } deriving (Show, Read, Eq, Ord)
```
hdgarrood: I like this, this seems like a nice improvement over a tuple. 👍

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -179,7 +179,9 @@ parseTypeClassDeclaration = do
   return $ TypeClassDeclaration className idents implies members
 
 parseConstraint :: TokenParser Constraint
```
hdgarrood: Should we be using `fromAesonParser` here?

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -246,69 +258,83 @@ checkExhaustive hasConstraint env mn numArgs cas = makeResult . first nub $ fold
...
+                                 []
+                                 (Right (TypedValue True (Abs (Left (Ident "x"))
+                                                              (Var (Qualified Nothing (Ident "x"))))
+                                                         ty))
+
+      ty :: Type
+      ty = ForAll "a" (ConstrainedType [ Constraint (Qualified (Just (ModuleName [ProperName C.prim]))
+                                                               (ProperName "Partial"))
+                                                    []
+                                                    (Just (PartialConstraintData (map (map prettyPrintBinderAtom) bss) complete))
```
hdgarrood: Why pretty print the binders here (as opposed to, say, than keeping them as `Binder` values until they need to be printed in the Errors module)?

src/Language/PureScript/Errors.hs
```diff
@@ -746,7 +744,19 @@ prettyPrintSingleError full level showWiki e = flip evalState defaultUnknownMap
             , line "They may be disallowed completely in a future version of the compiler."
             ]
     renderSimpleErrorMessage OverlappingInstances{} = internalError "OverlappingInstances: empty instance list"
-    renderSimpleErrorMessage (NoInstanceFound nm ts) =
+    renderSimpleErrorMessage (NoInstanceFound (Constraint (Qualified (Just (ModuleName [ProperName "Prim"]))
+                                                                     (ProperName "Partial"))
```
hdgarrood: Is there a declaration in `Language.PureScript.Constants` for the value `Qualified (Just ModuleName [ProperName "Prim"]) (ProperName "Partial")`? Might it make sense to add one, if not?

examples/passing/EmptyTypeClass.purs
```diff
@@ -1,12 +1,11 @@
...
-
-head :: forall a. (PartialP) => Array a -> a
+head :: forall a. Partial => Array a -> a
 head [x] = x
 
-instance allowPartials :: PartialP
-
-main = Control.Monad.Eff.Console.log $ head ["Done"]
+main :: Eff _ _
+main = log "Done"
```
hdgarrood: Why remove the call to `head` here?

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -179,7 +179,9 @@ parseTypeClassDeclaration = do
...
-parseConstraint = (,) <$> parseQualified properName <*> P.many (noWildcards parseTypeAtom)
+parseConstraint = Constraint <$> parseQualified properName
```
hdgarrood: Should we be using `fromAesonParser` here?

hdgarrood:
> If the user omits the type signature at the top level, they get a warning with the inferred type, which will include the generalized Partial constraint.

Do you think it would be worth adding a test to `TestDocs` to ensure that inferred Partial constraints appear properly in documentation when top-level type signatures are omitted?

This looks great to me 👍 I also like how it makes it easy to add different kinds of additional constraint information later.

src/Language/PureScript/Types.hs
hdgarrood: (Now that I've reviewed this properly I know what they mean, but I think it might still be worth adding more information to the comments here.)

src/Language/PureScript/Parser/Declarations.hs
hdgarrood: I'm a bit worried that this didn't cause the tests to fail, actually, unless this instance just happens to match exactly up with the derived `ToJSON` instance in `Language.PureScript.Types`.

src/Language/PureScript/Linter/Exhaustive.hs
garyb: It introduces a circular dependency between the `Binders` and `Types` modules. We talked about it a bit, and since this is basically a temporary solution until we have constraint kinds and programmable error messages, using a string didn't seem so bad.

hdgarrood: Ah, that makes sense.

paf31:
@hdgarrood Thanks for all the great comments!

examples/passing/PartialFunction.purs
paf31: The issue is that I don't have a way to call `unsafePartial`, so I have to annotate `fn` with a `Partial` constraint, at which point `Eff` has the wrong type. So I'm just skipping this for now.

src/Language/PureScript/Linter/Exhaustive.hs
paf31: I'll add a note.

src/Language/PureScript/Errors.hs
paf31: 👍

examples/passing/EmptyTypeClass.purs
paf31: Same reason as above. I don't have `unsafePartial` in scope. Maybe we should add it to the list of support modules?

src/Language/PureScript/Parser/Declarations.hs
paf31: Are you looking at the right parser module here?

src/Language/PureScript/Errors.hs
paf31: I used a pattern synonym to address this. Generally, I think we should be using them all over the place to tidy up code, especially in the optimizer and sugar passes.

paf31:
Looks like tests are now failing because `uncaught exception: ErrorCall (Prelude.last: empty list)`. I'll have a look tomorrow, but it looks like the check for `Done` is failing for some reason. I still can't run tests on my machine, so it's a bit slow to diagnose.

Speaking of tests, we should try to get `core-tests` working again before 0.9.

src/Language/PureScript/Parser/Declarations.hs
hdgarrood: lol, I thought this was Docs.Types. Yeah, never mind then :)

src/Language/PureScript/Errors.hs
hdgarrood: :+1: that seems like a great application of pattern synonyms, yeah

examples/passing/EmptyTypeClass.purs
hdgarrood: Possibly, I think you're probably better equipped than I am to decide whether it's worth doing.

garyb:
I'd vote for adding `unsafePartial` to the support code. Either that, or providing it in the tests as needed since it should be possible to provide FFI code in there now.

garyb:
Re: tests - Just wipe out the `.test_modules` and `support` folder, then revive `support` with git. Or if it still doesn't work, make a fresh clone of the repo. That worked for Christoph in the end.

hdgarrood:
I think that exception must be coming from `assertCompiles` in TestCompiler, which began would be caused if a test doesn't produce any output, so maybe we should reverse then pattern match instead of calling `last` in TestCompiler?

garyb:
Yeah, I figured it probably would be there too. A case for `null lines` would also work. I didn't add the case originally as I figured a failure's a failure, but it probably would be less confusing if handled properly, in case the `last` exception was arising elsewhere.

paf31:
Tests are now running, and I think this is probably good to review/merge if people are happy with the approach.

hdgarrood:
👍

garyb:
:+1:

PR #2069, opened 2016-04-30
New Prelude names in 0.9

garyb:
This may need some looking at again now, I'm not sure yet.

PR #2073, opened 2016-05-01
0.9 Remove deprecated class import/export syntax

LiamGoodacre:
We should wait until the Prelude updates (#2069) are in so we don't temporarily break the publish tests.

### DIFF
--- a/examples/docs/src/TypeClassWithoutMembers.purs
+++ b/examples/docs/src/TypeClassWithoutMembers.purs
@@ -8,4 +8,4 @@ module Intermediate
   ( module SomeTypeClass )
   where
 
-import SomeTypeClass (SomeClass)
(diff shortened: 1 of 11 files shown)


paf31:
What is the issue with the test suite?

LiamGoodacre:
@paf31 The existing Prelude has exports using the deprecated syntax. `Semigroupoid` for example, giving the error `Cannot export unknown type Semigroupoid`.

paf31:
If you merge in @garyb's latest changes, does that change anything?

LiamGoodacre:
@paf31 Tests all pass locally now that I've rebased :]

paf31:
Brilliant, thanks!

garyb:
Thanks!

PR #2079, opened 2016-05-01
marking code in error messages with backticks

soupi:
as discussed on #2056 and #2060 :)

### DIFF
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -41,10 +41,10 @@ data PSCMakeOptions = PSCMakeOptions
 printWarningsAndErrors :: Bool -> Bool -> P.MultipleErrors -> Either P.MultipleErrors a -> IO ()
 printWarningsAndErrors verbose False warnings errors = do
   when (P.nonEmpty warnings) $
-    hPutStrLn stderr (P.prettyPrintMultipleWarnings True verbose warnings)
+    hPutStrLn stderr (P.prettyPrintMultipleWarnings (P.defaultPPEOptions { P.ppeFull = verbose }) warnings)
   case errors of
     Left errs -> do
-      hPutStrLn stderr (P.prettyPrintMultipleErrors True verbose errs)
+      hPutStrLn stderr (P.prettyPrintMultipleErrors (P.defaultPPEOptions { P.ppeFull = verbose }) errs)
       exitFailure
     Right _ -> return ()
 printWarningsAndErrors verbose True warnings errors = do
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -41,10 +41,10 @@ data PSCMakeOptions = PSCMakeOptions
 printWarningsAndErrors :: Bool -> Bool -> P.MultipleErrors -> Either P.MultipleErrors a -> IO ()
 printWarningsAndErrors verbose False warnings errors = do
   when (P.nonEmpty warnings) $
-    hPutStrLn stderr (P.prettyPrintMultipleWarnings verbose warnings)
+    hPutStrLn stderr (P.prettyPrintMultipleWarnings True verbose warnings)
   case errors of
     Left errs -> do
-      hPutStrLn stderr (P.prettyPrintMultipleErrors verbose errs)
+      hPutStrLn stderr (P.prettyPrintMultipleErrors True verbose errs)
       exitFailure
     Right _ -> return ()
 printWarningsAndErrors verbose True warnings errors = do
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -11,6 +11,7 @@ import           Control.Monad
 import           Control.Monad.Writer.Strict
 
 import qualified Data.Aeson as A
+import           Data.Bool (bool)
 import qualified Data.ByteString.Lazy as B
 import qualified Data.ByteString.UTF8 as BU8
 import qualified Data.Map as M
@@ -24,6 +25,7 @@ import           Options.Applicative as Opts
 
 import qualified Paths_purescript as Paths
 
+import qualified System.Console.ANSI as ANSI
 import           System.Exit (exitSuccess, exitFailure)
 import           System.FilePath.Glob (glob)
 import           System.IO (hSetEncoding, hPutStrLn, stdout, stderr, utf8)
@@ -40,11 +42,13 @@ data PSCMakeOptions = PSCMakeOptions
 -- | Argumnets: verbose, use JSON, warnings, errors
 printWarningsAndErrors :: Bool -> Bool -> P.MultipleErrors -> Either P.MultipleErrors a -> IO ()
 printWarningsAndErrors verbose False warnings errors = do
+  cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stderr
+  let ppeOpts = P.defaultPPEOptions { P.ppeCodeColor = cc, P.ppeFull = verbose }
   when (P.nonEmpty warnings) $
-    hPutStrLn stderr (P.prettyPrintMultipleWarnings (P.defaultPPEOptions { P.ppeFull = verbose }) warnings)
+    hPutStrLn stderr (P.prettyPrintMultipleWarnings ppeOpts warnings)
   case errors of
     Left errs -> do
-      hPutStrLn stderr (P.prettyPrintMultipleErrors (P.defaultPPEOptions { P.ppeFull = verbose }) errs)
+      hPutStrLn stderr (P.prettyPrintMultipleErrors ppeOpts errs)
       exitFailure
     Right _ -> return ()
 printWarningsAndErrors verbose True warnings errors = do
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -133,7 +133,8 @@ library
                    transformers-compat >= 0.3.0,
                    unordered-containers -any,
                    utf8-string >= 1 && < 2,
-                   vector -any
+                   vector -any,
+                   ansi-terminal >= 0.6.2 && < 0.7
 
     exposed-modules: Language.PureScript
                      Language.PureScript.AST
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -304,6 +304,7 @@ executable psc
     build-depends: base >=4 && <5,
                    purescript -any,
                    aeson >= 0.8 && < 0.12,
+                   ansi-terminal >= 0.6.2 && < 0.7,
                    base-compat >=0.6.0,
                    bytestring -any,
                    containers -any,
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -94,6 +94,7 @@ library
     build-depends: base >=4.8 && <5,
                    aeson >= 0.8 && < 0.12,
                    aeson-better-errors >= 0.8,
+                   ansi-terminal >= 0.6.2 && < 0.7,
                    base-compat >=0.6.0,
                    bower-json >= 0.8,
                    boxes >= 0.1.4 && < 0.2.0,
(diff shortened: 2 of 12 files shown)


paf31:
This looks good to me, but I feel like regular quote characters (single or double) would be more suitable than backticks, and would match some of our other errors.

soupi:
Alright. I'll make the change tonight.

jplatte:
Are you going with double quotes or single quotes? I think single quotes could be a problem since they can also be part of an identifier.

garyb:
Good point. Maybe backticks is best actually, as we can also have quoted labels for records and accessors.

soupi:
Yes, I initially chose backticks because `'` or `"` can be used in other contexts :)

jplatte:
Hmmm, can't we have backticks in the quoted code too, when functions as operators? I didn't have the code snippet in the output when I used the deprecated syntax for operator sections to try to get a good example, but I think there will always be a few cases where we will have wrong quotes. I think colored output might be the best solution, but we don't do that anywhere else yet, right?

paf31:
Another solution is to put code in an indented block, but that could lead to long errors.

garyb:
@soupi Sorry, but this will need rebasing now due to other changes that went in. I may have made your job easier now though, as most types of name go through one function in the error printing.

garyb:
Thanks for rebasing, there's one thing that will need a little tweak though - we probably want to move uses of `markCode` inside `printName`, as now it will print things like ``Unknown `value x` `` or ``Unknown `type operator (~>)` ``.

soupi:
Ah, all right. I'll make the necessary changes :)

paf31:
Does this work if something spans multiple lines?

soupi:
it does something like:
```
Error found:
in module `Slides`
at /Users/suppi/code/purescript/purescript-slides/src/Slides.purs line 56, column 27 - line 59, column 3

  Could not match type

    Slides

  with type

    { pos :: Int
    , slides :: Array Slide
    }


while trying to match type `HalogenFP _0 Slides`
  with type `HalogenFP EventSource    `
               { pos :: Int
               , slides :: Array Slide
               }
while checking that expression `(bind (modify (\$1 ->  `
                                                 ...
                                              )
                                      )
                                )
                                (\__unused ->
                                   pure next
                                )
  has type `Free                        `
              (HalogenFP EventSource
                 { pos :: Int
                 , slides :: Array Slide
                 }
                 Move
                 g0
              )
              a1
in value declaration `ui`

where `g0` is a rigid type variable
        bound at line 44, column 1 - line 65, column 1
      `a1` is a rigid type variable
        bound at line 56, column 3 - line 59, column 3
      `_0` is an unknown type

```

natefaubion:
> ```
`Free                        `
```

hconcat 😆

**Comment by paf31** [comment]
We have a combinator to put the final tick in the right place somewhere in `Pretty.Common`.

**Comment by soupi** [comment]
Boxes should now look like this:

```
tests/TestCompiler.hs:197:
  90) Failing examples 'SuggestComposition.purs' should fail with 'TypesDoNotUnify'
       Error found:
       in module `SuggestComposition`
       at /Users/suppi/code/haskell/purescript/examples/failing/SuggestComposition.purs line 7, column 5 - line 7, column 10

         Could not match type

           Record

         with type

           Function Int


       while trying to match type `{ g :: _0
                                   | _1
                                   }        `
         with type `_2 -> _3`
       while checking that expression `g`
         has type `{ g :: _0
                   | _1
                   }        `
       while checking type of property accessor `g.g`
       in value declaration `f`

       where `_2` is an unknown type
             `_3` is an unknown type
             `_0` is an unknown type
             `_1` is an unknown type

```

**Comment by paf31** [comment]
I don't know about this. The appearance really shouldn't matter, but the backticks just look .. odd. I'd rather go with coloring the types in the output, but then we'd need to modify the `boxes` library to support things spanning multiple lines. Is it really so bad to just leave things unquoted?

**Comment by soupi** [comment]
I'm not sure I'm satisfied with how it looks either and I think colored types will look better as well. Being able to immediately focus on the types is nice. Perhaps we should close this for now and open an issue about colored types instead? Or maybe ask in the #2056

**Comment by AppShipIt** [comment]
I think it's #1517.

**Comment by hdgarrood** [comment]
I agree with @paf31, yeah. Looking back at the original issue, as far as I can tell, the only ones that have a risk of causing confusion are "unknown type X" and "unknown value X" (imagine eg X is Error, but I guess there probably are others). What about quoting just those?

**Comment by soupi** [comment]
I can do that. should I do that?

**Comment by garyb** [comment]
In that case, it's probably just `printName` that needs to use `markCode`. Or any time a name appears in the middle of a sentence perhaps, so there are a bunch of other cases that are fine too.

**Comment by paf31** [comment]
Yes please 😀

**Comment by soupi** [comment]
I made a change so that backticks will only show when there's code that does not span on multiple lines. for example:

```
     Error found:
       in module `Main`
       at /Users/suppi/code/haskell/purescript/examples/failing/TypedBinders2.purs line 7, column 3 - line 7, column 14

         Could not match type

           Unit

         with type

           String


       while checking that expression case $0 of
                                        s -> log "Done"
         has type Eff
                    ( console :: CONSOLE
                    | _3
                    )
                    _2
       while applying a function `(bind (#dict Bind _1)) (log "Foo")`
         of type `(_0 -> _1 _2) -> _1 _2`
         to argument \$0 ->
                       case $0 of
                         s -> log "Done"
       in value declaration `main`

       where `_1` is an unknown type
             `_2` is an unknown type
             `_0` is an unknown type
             `_3` is an unknown type

```

I can still remove more cases, but I wanted to check with you first to see how you feel about this.

**Comment by garyb** [comment]
That approach looks good to me!

**Comment by paf31** [comment]
Well, if things are on one line, we should definitely be able to use colors instead of ticks. Actually, then we could use different colors for identifiers, operators and `ProperName`s.

Also, I would expect `Unit` and `String` to be ticked in your example.

Looks good though.

**Comment by AppShipIt** [comment]
> Looks good though, I'd just prefer coloring to ticks.

Could you could switch based on the terminal mode? #1517 sounds like a bit more to iron out?

**Comment by paf31** [comment]
In general, I think so, but maybe not so easily in pure code.

**Comment by natefaubion** [comment]
I think you'd have to anyway with `--json-errors`, unless you intend to push all the ANSI escapes through 😱. It would be cool if you were going to highlight stuff, to push the ranges through in `--json-errors` so a pretty printer could highlight them if desired.

**Comment by paf31** [comment]
We always call the printer from somewhere in IO, so we could just provide a `Bool` as an argument. An options record would probably be better though.

**Comment by soupi** [comment]
how about putting ansi colors in the regular `markCode` code and when there's a flag to remove them, for example with `--json-errors`, `filter` them out?

**Comment by paf31** [comment]
We could, but I think it would be a bit cleaner to just not put them there in the first place.

**Comment by soupi** [comment]
screen shot 2016-05-27 at 3 18 38 pm

I've only added the yellow color around code. the rest are from the haskell testing framework I think :)

Note that I've also taken the liberty to choose whether some places will report errors with color or not, but I'm not very familiar with the compiler so I might have made some mistakes there as well.

File: purescript.cabal [hunk 1]
```diff
@@ -133,7 +133,8 @@ library
                    transformers-compat >= 0.3.0,
                    unordered-containers -any,
                    utf8-string >= 1 && < 2,
-                   vector -any
+                   vector -any,
+                   ansi-terminal >= 0.6.2 && < 0.7
```
paf31: Could you please resort these?

src/Language/PureScript/Errors.hs
```diff
@@ -457,11 +460,37 @@ showSuggestion suggestion = case errorSuggestion suggestion of
...
+      , Box.vcat Box.top $ replicate (Box.rows b) $ Box.text ansiColorReset
+      ]
+
 -- |
 -- Pretty print a single error, simplifying if necessary
 --
-prettyPrintSingleError :: Bool -> Level -> Bool -> ErrorMessage -> Box.Box
-prettyPrintSingleError full level showWiki e = flip evalState defaultUnknownMap $ do
+prettyPrintSingleError :: Bool -> Bool -> Level -> Bool -> ErrorMessage -> Box.Box
+prettyPrintSingleError (colorCode &&& colorCodeBox -> (markCode, markCodeBox)) full level showWiki e = flip evalState defaultUnknownMap $ do
```
paf31: Is this `ViewPatterns`?

paf31: I think it would be nice to factor out an options record here.

src/Language/PureScript/Errors.hs
```diff
@@ -457,11 +460,37 @@ showSuggestion suggestion = case errorSuggestion suggestion of
...
+colorCode shouldColorCode code
+  | shouldColorCode = concat [ansiColor codeColor, code, ansiColorReset]
+  | otherwise = code
+
+colorCodeBox :: Bool -> Box.Box -> Box.Box
+colorCodeBox shouldColorCode b
+  | not shouldColorCode = b
+  | shouldColorCode && Box.rows b == 1 = Box.text (ansiColor codeColor) Box.<> b `endWith` Box.text ansiColorReset
+  | otherwise = Box.hcat Box.left
+      [ Box.vcat Box.top $ replicate (Box.rows b) $ Box.text $ ansiColor codeColor
```
paf31: Interesting approach :)

psc/Main.hs
```diff
@@ -41,10 +41,10 @@ data PSCMakeOptions = PSCMakeOptions
 printWarningsAndErrors :: Bool -> Bool -> P.MultipleErrors -> Either P.MultipleErrors a -> IO ()
 printWarningsAndErrors verbose False warnings errors = do
   when (P.nonEmpty warnings) $
-    hPutStrLn stderr (P.prettyPrintMultipleWarnings verbose warnings)
+    hPutStrLn stderr (P.prettyPrintMultipleWarnings True verbose warnings)
```
paf31: Do we need to check if the output is a terminal?

src/Language/PureScript/Errors.hs
soupi: Yes for `ViewPatterns` (one of my favorite extensions). I'll try to work on refactroing an options record. :)

paf31: I've always used `PatternGuards` in the past, I tend to find this way a bit harder to read, but I'm fine with it this way.

AppShipIt:
Purely aesthetic: make sure to test in terminal with a light theme. I recall yellow working at work though.

paf31:
Hmm, good point. Perhaps an alternative is something like bold text, or underlined, or swapping the background and foreground colors. Or just picking a different color, if necessary.

paf31:
Worst case, we can provide a `--monochrome` flag.

soupi:
Perhaps the `Vivid` color intensity is a bit too much for a light theme, maybe `Dull` will look better.
We can also support flags to change the color, or `--monochrome`, or choose a different color. I'm open to suggestions :)

psc/Main.hs
```diff
@@ -40,11 +42,12 @@ data PSCMakeOptions = PSCMakeOptions
 -- | Argumnets: verbose, use JSON, warnings, errors
 printWarningsAndErrors :: Bool -> Bool -> P.MultipleErrors -> Either P.MultipleErrors a -> IO ()
 printWarningsAndErrors verbose False warnings errors = do
+  cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stderr
   when (P.nonEmpty warnings) $
-    hPutStrLn stderr (P.prettyPrintMultipleWarnings (P.defaultPPEOptions { P.ppeFull = verbose }) warnings)
+    hPutStrLn stderr (P.prettyPrintMultipleWarnings (P.defaultPPEOptions { P.ppeCodeColor = cc, P.ppeFull = verbose }) warnings)
```
paf31: Let's pull the options out into a `let` binding here.

paf31:
Thanks!

AppShipIt:
What about just dimming the color?

paf31:
It should be dull yellow now.

PR #2081, opened 2016-05-01
Update the support modules list in the tests

garyb:


paf31:
👍 Thanks!

paf31:
Is there something I need to do to get tests running? They all seem to be failing for me.

garyb:
Hmm, no. That kinda was the point of switching back to bower, that it should "just work"... Travis is managing okay too. Perhaps clear out the stuff you have in `support` and try again?

hdgarrood:
Should we be sharing this list across both of these modules?

garyb:
That would make sense... I'll revise it.

garyb:
The list is shared now.

hdgarrood:
:+1:

PR #2082, opened 2016-05-02
Improve license errors in psc-publish

hdgarrood:
Currently, you get the 'no license specified' error even if you do
specify a license but it's not a valid SPDX license expression. There
is quite a lot of code in the wild which specifies a license but does
not use the SPDX format, and in these cases the error is not very good.

This PR does three things:
* adds a new error type to differentiate between these two cases
* removes the (now redundant) LicenseNotFound constructor
* Adds `Apache-2.0` to the SPDX license suggestions, as there seem to be quite a few packages that have `Apache 2.0` in their `license` field (i.e. with a space rather than a hyphen).


paf31:
👍 LGTM, thanks!

PR #2083, opened 2016-05-02
Cache last rebuild

kRITZCREEK:
Rebased on top of the 0.9 branch.

I think this is ready for review. If you can think of a way to make reviewing changes like this easier let me know. I can offer to implement the functionality in my psc-ide-emacs fork but I don't think anyone but me is using emacs?

### DIFF
--- a/src/Language/PureScript/Ide/Rebuild.hs
+++ b/src/Language/PureScript/Ide/Rebuild.hs
@@ -1,43 +1,66 @@
 {-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE PackageImports        #-}
+{-# LANGUAGE RecordWildCards       #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TupleSections         #-}
 
-module Language.PureScript.Ide.Rebuild where
-
-import           Prelude.Compat
-
-import           Language.PureScript.Ide.Error
-import           Language.PureScript.Ide.State
-import           Language.PureScript.Ide.Types
+module Language.PureScript.Ide.Rebuild
+  ( rebuildFile
+  ) where
 
 import           Control.Monad.Error.Class
 import           Control.Monad.IO.Class
 import           "monad-logger" Control.Monad.Logger
-import           Control.Monad.Reader.Class
+import           Control.Monad.Reader
 import           Control.Monad.Trans.Except
 import qualified Data.Map.Lazy                   as M
 import           Data.Maybe                      (fromJust, mapMaybe)
+import           Data.Monoid                     ((<>))
 import qualified Data.Set                        as S
 import qualified Language.PureScript             as P
 import           Language.PureScript.Errors.JSON
-import           System.IO.UTF8 (readUTF8File)
+import           Language.PureScript.Ide.Error
+import           Language.PureScript.Ide.State
+import           Language.PureScript.Ide.Types
+import           Language.PureScript.Ide.Util
+import           Prelude.Compat
+import           System.IO.UTF8                  (readUTF8File)
 
+-- | Given a filepath performs the following steps:
+--
+-- * Reads and parses a PureScript module from the filepath.
+--
+-- * Builds a dependency graph for the parsed module from the already loaded
+-- ExternsFiles.
+--
+-- * Attempts to find an FFI definition file for the module by looking
+-- for a file with the same filepath except for a .js extension.
+--
+-- * Adds a default import for Prim to the parsed module and passes all the
+-- created artifacts to @rebuildModule@.
+--
+-- * If the rebuilding succeeds, returns a @RebuildSuccess@ with the generated
+-- warnings, and if rebuilding fails, returns a @RebuildError@ with the
+-- generated errors.
 rebuildFile
   :: (PscIde m, MonadLogger m, MonadError PscIdeError m)
   => FilePath
+  -> Bool
   -> m Success
-rebuildFile path = do
+rebuildFile path cacheSuccess = do
 
   input <- liftIO $ readUTF8File path
 
-  m <- case map snd <$> P.parseModulesFromFiles id [(path, input)] of
-         Left parseError ->
-           throwError . RebuildError . toJSONErrors False P.Error $ parseError
-         Right [m] -> pure m
-         Right _ -> throwError . GeneralError $ "Please define exactly one module."
+  m <- P.addDefaultImport (P.ModuleName [P.ProperName "Prim"]) <$>
+    case map snd <$> P.parseModulesFromFiles id [(path, input)] of
+      Left parseError ->
+        throwError . RebuildError . toJSONErrors False P.Error $ parseError
+      Right [m] -> pure m
+      Right _ -> throwError . GeneralError $ "Please define exactly one module."
 
   -- Externs files must be sorted ahead of time, so that they get applied
   -- correctly to the 'Environment'.
-  externs <- sortExterns m . M.delete (P.getModuleName m) =<< getExternFiles
+  externs <- sortExterns m =<< getExternFiles
 
   outputDirectory <- confOutputPath . envConfiguration <$> ask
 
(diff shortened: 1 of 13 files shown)


hdgarrood:
I'm actually giving spacemacs another try at the moment.

LiamGoodacre:
I use spacemacs too.

kRITZCREEK:
@hdgarrood @LiamGoodacre well if you want to give it a try you can `stack install` the binaries from this branch and use my [private layer](https://github.com/kRITZCREEK/dotfiles/tree/master/.spacemacs.d/private/kc-purescript).

I'll document it :soon:, but for now if you disable the purescript layer and instead use the kc-purescript layer you can open a file in a purescript project and do:

`, m s Return` - Start the server
`, m l` - Load all modules

If you now try to get completions in a module you should only be able to complete exported identifiers.

`, m b` - rebuild

If rebuilding succeeds and you try to get completions you should now also get suggestions for private identifiers.

paf31:
Do we know why the build fails?

kRITZCREEK:
Sorry, but I have no idea O.o

paf31:
Looks like Appveyor is having trouble with Node.

kRITZCREEK:
@garyb Could you maybe run the tests for this branch on windows? I don't see why this wouldn't work on windows

garyb:
I can do sure, I think this is just an AppVeyor glitch though. @hdgarrood can you restart it? Also, is there any way you can give us permission to do stuff with setup/restarting on there too?

hdgarrood:
I've restarted it. I have no idea how to give you access, though. :(

garyb:
Hmm, it timed out again, weird. Got a lot further that time though.

@hdgarrood No problem about the AppVeyor thing, you'd think it would just work anyway, as it should be able to tell from our organisation membership the way Travis does.

AppShipIt:
Maybe http://help.appveyor.com/discussions/questions/1459-mirroring-github-organization-permissions? Y'all are doing awesome code but I figured I could at least google in between configuring WebPack 😄

garyb:
That probably deserved a "lmgtfy", I was just too lazy to check :laughing:. Thanks though! I need to do the same for SlamData soon.

hdgarrood:
Thanks @AppShipIt, I think that should do the trick. Anyone in the owners team of the purescript org should be able to restart builds now, I think.

AppShipIt:
Much obliged jealous of y'all's productivity today 👏

kRITZCREEK:
Sorry but I have no idea why this fails on appveyor. I can squash and have it rebuild again but it's just guessing from my side.

garyb:
I'll try it locally, see if there's any more info.

garyb:
Nope, no more useful info, but it does get up to:

```
Language.PureScript.Ide.Imports.Integration
  Adding imports
```

and then hangs indefinitely... well, for at least 5 minutes before I killed it anyway, and more like 60mins on AppVeyor by the looks of it, since it's timing out there. :cry:

kRITZCREEK:
Argh... Thank you for the information! I'll see if I can add in a few timeouts and see where it hangs.

garyb:
The only other thing I can say is it doesn't appear to be using any CPU time and the memory usage is static when it gets stuck too.

kRITZCREEK:
Setting up a Windows environment atm. I have no idea what is going on :D

garyb:
Doh :see_no_evil:

kRITZCREEK:
Sank enough time into this for today... this is really frustrating :D It looks like something is deadlocking, but I think that really shouldn't happen when I'm only using `STM`.

tests/support/pscide/src/RebuildSpecWithHiddenIdent.purs
```diff
@@ -1,4 +1,6 @@
 module RebuildSpecWithHiddenIdent (exported) where
 
 hidden x _ = x
+
+exported :: forall a. a -> a
```
kRITZCREEK: @paf31 @garyb This is the change that fixes the hanging. Without it `psc` gets stuck. I can't reproduce it on the commandline though.... it only happens in `compile` in `tests/Language.PureScript.Ide.RebuildSpec`.

kRITZCREEK:
Squashed and rebased on the newest 0.9 branch

kRITZCREEK:
The CI failure:

```
Executable named ghc not found on path: ["/Users/travis/.stack/programs/x86_64-osx/ghc-7.10.3/bin","/Users/travis/.rvm/gems/ruby-2.0.0-p481/bin","/Users/travis/.rvm/gems/ruby-2.0.0-p481@global/bin","/Users/travis/.rvm/rubies/ruby-2.0.0-p481/bin","/Users/travis/.rvm/bin","/Users/travis/bin","/Users/travis/.local/bin","/usr/local/bin","/usr/bin","/bin","/usr/sbin","/sbin","/opt/X11/bin"]
```

That shouldn't be related to this PR.

hdgarrood:
Maybe stack didn't manage to download GHC properly? I've restarted that build. If that doesn't work I'll blow away the cache, maybe `~/.stack` got into an inconsistent state somehow.

hdgarrood:
I just deleted the cache for this PR, so if it still breaks, you could rebase and push again and it should build from scratch, which will hopefully fix this.

paf31:
Sorry, this has conflicts now, but they should be relatively minor.

kRITZCREEK:
The conflicts are very minor. They are not even because of the psci PR

paf31:
Code looks good, but let me make sure I understand:

- The editor plugin is responsible for calling `rebuildFile` with `cacheSuccess` set to `True`.
- If the flag is set, we build _twice_, once with exports and once without, and cache the result of the second run.

How does the editor plugin determine when to set the flag? Could we decide on their behalf?

kRITZCREEK:
You got all the steps right.

The reason for why I put the flag in, is that rebuilding is not free in terms of performance. Rebuilding twice will almost double the time for a rebuild and I think that will make the UX considerably worse on big projects like slamdata.

We could decide to always cache a successful rebuild but then I'd like to change the signature of rebuildModule like I said in the psci PR, so that we only have to build up the Environment once.

Another option that I thought of is that instead of caching only the last rebuild, we could cache the last rebuild for any module we rebuild. This will increase the memory footprint of psc-ide slightly, but simplify the code. What do you think?

paf31:
Perhaps it is easier to just do externs generation twice then?

kRITZCREEK:
I kind of talked about this with @hdgarrood and @nwolverson and they didn't like the idea that this would now allow users to access private members of their modules. I was kind of asking for this in #1620.

paf31:
I tend to agree, but we don't have to _save_ the modified externs.

kRITZCREEK:
rebased again and adopted Rebuild.hs to utilize #2126

kRITZCREEK:
I don't think changing the Make API to accomodate for this very narrow usecase is a good idea. I still think it's a good idea to give tools the ability to turn off the cache, because tools that only report errors (pscid, webpack-loader) and don't offer any completions shouldn't have to wait twice as long.

I image that Atom or Emacs would turn it on by default and allow the user to disable it with an option.

Once we start working emulating ghci's `:load` for psci we can still look to unify the two approaches but I think it's best to defer that decision until we know what we need for psci. It's not that much code after all.

paf31:
What if the secondary build were sparked on another thread in the background? That way, we can cache the result, but not hold up the UI.

kRITZCREEK:
I don't have a working strategy for "being careful" yet.. Spinning up worker threads makes testing a lot harder and I don't want to take that complexity until we have a really good reason.

Thinking about this a bit more, I am fine with just removing the flag and always rebuilding twice, since a lot of the "rebuild" time is actually spent sorting the externs and we don't have to do that twice.

kRITZCREEK:
Rebased and merged again :)

kRITZCREEK:
Is this PR going in with 0.9? It's not a breaking change so if it's too much to review until then we could easily put it into 0.9.1

paf31:
I think it should, but I won't be able to review it for a couple of days.

src/Language/PureScript/Ide/Rebuild.hs
```diff
@@ -51,12 +50,11 @@ rebuildFile path cacheSuccess = do
...
-    case map snd <$> P.parseModulesFromFiles id [(path, input)] of
-      Left parseError ->
-        throwError . RebuildError . toJSONErrors False P.Error $ parseError
-      Right [m] -> pure m
-      Right _ -> throwError . GeneralError $ "Please define exactly one module."
+  m <- case map snd <$> P.parseModulesFromFiles id [(path, input)] of
+    Left parseError ->
+      throwError . RebuildError . toJSONErrors False P.Error $ parseError
+    Right [m] -> pure m
+    Right _ -> throwError . GeneralError $ "Please define exactly one module."
```
paf31: I think this case should be an `internalError` now, since we _only_ allow a single module per file.

paf31:
Looks good to me!

paf31:
Thanks @kRITZCREEK !

src/Language/PureScript/Ide/Rebuild.hs
kRITZCREEK: Seems like you're looking at an outdated diff. I did that in this PR

kRITZCREEK:
Cool thanks!

src/Language/PureScript/Ide/Rebuild.hs
paf31: Oh, weird. Thanks, looks good.

PR #2085, opened 2016-05-02
Use HSpec for the compiler tests

garyb:


garyb:
Apart from it being easier to figure out what and where problems are when they occur, I'm also setting this up so we can add some cases that are a bit more involved than just running the compiler. (Checking prelude inlining works correctly, export-elaboration, etc.)

hdgarrood:
👍 👍 👍 

Presumably `silently` is to avoid all the "Compiling" messages every time? Do you think it would make it a little easier to read if you imported it qualified?

garyb:
Yeah, the `Compiling ...` messages and also the errors for the failing cases. I don't mind either way about qualifying, `System.IO.Silently` only exports `silence`, `capture` and some variants.

kRITZCREEK:
Yes please!

Where is the `context` function coming from? Would `describe` from hspec be nicer?

garyb:
`context` is from HSpec too, it's the same as `describe` but seemed like a better fit for the description of what the tests were doing here; all the following `it`s are in the context of things that should pass or fail.

hdgarrood:
Also, perhaps something to consider later, I think it might be worth making assertions about the output of some of the passing examples, eg. that they print "done" or something.

garyb:
Yeah, there's a few more things I want to do with the tests, like checking warnings as well. Asserting a `Done` final output is probably worth adding to the list, I think most of the tests currently do that. We'll probably have to pipe the node output to a temp file and read that to check.

I figured just getting this in first would be a worthwhile step though, I've learned how painful large and/or long running PRs can be while working on SlamData :wink:

hdgarrood:
👍 don't we have the stdout from node available already from the result of `readProcessWithExitCode`? But yeah, agreed re long-running PRs. :)

garyb:
Oh yeah, of course! In fact I deleted the line that printed the node output to stdout.

paf31:
👍 Looks great, thanks!

garyb:
Just adding the "Done" detection now as it's pretty straightforward - in fact, it's done, I just need to update some tests.

garyb:
This is good for another look now. I've also added a check that if anything is printed to `stderr` the test is considered to be failing.

hdgarrood:
LGTM

PR #2087, opened 2016-05-03
Multi-file test setup

garyb:
This allows tests to consist of multiple files, by pulling in anything in a directory with the same name as a test file. Also FFI code can be provided in the tests now.

I've split every multi-module test into using separate files for each of its modules too, which will help with some of the upcoming changes, and also proves the thing actually works.

The `manual` tests are now included in the automatically run tests too. I think they could have been moved over a while ago actually, but anyway, they're done and passing now as part of the "automatic" tests.


paf31:
👍 Looks good to me, thanks! It's good to be able to run the manual tests again.

garyb:
The tests failed because of an assumption I made about how the paths would be sorted within each group of test files. I've made a change to that to ensure it picks the right one as the entry point now, so if the tests pass this time I'll go ahead and merge.

Issue #1441, opened 2015-08-30
PSCi should support alternatives to Prelude/Console


paf31:
PSCi currently relies on `purescript-console`, but for teaching purposes, I would like to be able to use `purescript-preface` with PSCi, for example.

I don't know how to support this, but I think it would be good to be able to support different sets of core libraries.

Perhaps we can do something like we do with rebindable do notation, and rely on some name being in scope. Maybe we need to also make the `$PSCI.Support` module available for import.

garyb:
Perhaps something else we could add `:command` to configure it, that way it could be included in a `.psci` file too?

tfausak:
:+1: This would be useful for [purescript-neon](https://github.com/tfausak/purescript-neon).

paf31:
This could be as simple as moving the `PSCI.Support` module into its own versioned package, so that writers of custom preludes can ship an instance for the `Eval` class.

hdgarrood:
Hm, I think it might be a problem if we want to give something both `Show` and `Eval` instances. Perhaps we can cross that bridge when we get to it, though.

paf31:
Yes, it's an overlapping instance problem generally, but maybe we can fix that with instance chains one day.

paf31:
I'll leave this open since `purescript-psci-support` still needs to be updated and moved to core.

paf31:
Closing, since the support module is now available [here](https://github.com/purescript/purescript-psci-support).

PR #2088, opened 2016-05-03
PSCi speed improvements and design changes

paf31:
This is an attempt to speed up evaluation in PSCi by simplifying things somewhat, and using the same approach as `psc-ide-server`.

Specifically, we will perform one big `make` when PSCi loads, and keep the externs in memory. When evaluating an expression, type, kind, import, etc., we will simply `rebuild`, which is _much_ faster.

This PR isn't complete, but in my tests, this makes PSCi much faster.

We can do better still by having Node run in the background, instead of starting a fresh process each time, but that's a job for another PR.


kRITZCREEK:
Looks like we duplicated some work :D One question though. Without watching the output folder won't the loaded externs get stale quickly?

paf31:
Yes, there's definitely some overlap, and we should try to combine the two. For this, I was just trying to make PSCi itself faster, not reload files automatically. In fact, I'd like to change that behavior in PSCi to match GHCi.

paf31:
@hdgarrood Would it be a pain to change pulp to pass globs to PSCi instead of creating a `.psci` file?

hdgarrood:
Not really. I'd be happy if it did change to do that, too, because I think that would be a much less error-prone interface.

paf31:
Okay, I'm thinking of simplifying PSCi a bit to get faster builds, and one change will potentially be to ditch `.psci` files.

paf31:
Okay, this branch is ready to review. The speed seems much better to me now, and I think this puts us in a better place to implement things like `:load`. Here is what has changed. Some are simplifications, some are optimizations:

- The `.psci` file is no longer used. Modules are loaded on the command line.
- We compile all dependencies once on startup and save the externs. `:reload` causes a recompile. Otherwise, we reuse the externs we have in memory.
- Evaluation and `:type` now use `rebuild` instead of `make`, which is much faster.
- General tidying and better use of `mtl` here and there.

kRITZCREEK:
I'm getting this message when building/testing the project now:

```
There were multiple candidates for the Cabal entry "Main.hs" 
(/home/creek/Documents/purescript/psci/Main.hs),
 picking /home/creek/Documents/purescript/tests/Main.hs
```

This is because of the changes you made to the structure of `psci/` which now includes a `Main.hs` just like `tests/`.

I see 3 possibilities to clean this up:

1. revert to the `psci/main/Main.hs` structure - *meh*
2. Move the PSCi code into the library so we don't have to pull in files from `psci/` into the tests - *meh*
3. Rename `tests/Main.hs` to `tests/Test.hs` - *YAY*

psci/PSCi/Completion.hs
```diff
@@ -138,12 +139,12 @@ getModuleNames :: CompletionM [String]
 getModuleNames = moduleNames <$> getLoadedModules
 
 mapLoadedModulesAndQualify :: (a -> String) -> (P.Module -> [(a, P.Declaration)]) -> CompletionM [String]
-mapLoadedModulesAndQualify sho f = do
-  ms <- getLoadedModules
-  let argPairs = do m <- ms
-                    fm <- f m
-                    return (m, fm)
-  concat <$> traverse (uncurry (getAllQualifications sho)) argPairs
+mapLoadedModulesAndQualify sho f = undefined --do
```
kRITZCREEK: Do we want to check this in?

psci/Main.hs
```diff
@@ -0,0 +1,349 @@
...
+-- | Read a file in the 'P.Make' monad, handling errors.
+readFileMake :: FilePath -> P.Make String
+readFileMake path = P.makeIO (const (P.ErrorMessage [] $ P.CannotReadFile path)) (readUTF8File path)
+
+-- | Rebuild a module, using the cached externs data for dependencies.
+rebuild :: PSCiState -> P.Module -> P.Make (P.ExternsFile, P.Environment)
+rebuild PSCiState{..} m = do
+    let loadedExterns = map snd psciLoadedExterns
+    externs <- P.rebuildModule buildActions loadedExterns m
+    return (externs, foldl' (flip P.applyExternsFileToEnvironment) P.initEnvironment (loadedExterns ++ [externs]))
```
kRITZCREEK: If `P.rebuildModule` returned an `(ExternsFile, Environment)` we wouldn't have to duplicate all the work of folding the dependency graph into an `Environment`. I have the same issue when caching rebuilds as I need to change the Makeoptions but rebuild against the same set of dependencies.

psci/Main.hs
```diff
@@ -0,0 +1,349 @@
...
+--
+handleTypeOf :: P.Expr -> PSCI ()
+handleTypeOf val = do
+  st <- get
+  let m = createTemporaryModule False st val
+  e <- liftIO . runMake $ rebuild st m
+  case e of
+    Left errs -> PSCI $ printErrors errs
+    Right (_, env') ->
+      case M.lookup (P.ModuleName [P.ProperName "$PSCI"], P.Ident "it") (P.names env') of
```
kRITZCREEK: Not that it makes to big of a difference, but we should also be able to look up the value and it's type by looking at the produced `ExternsFile`. If this would also work for `handleKindOf`, `rebuild` wouldn't need to return an `Environment`

psci/Main.hs
paf31: I like the second option. I'd like to avoid passing around environments though, since those are really an internal structure for the typechecker. Externs should capture everything in the public API for a module, so if we accidentally end up with more stuff in the `Environment` after typechecking, it's fine, because we'll reconstruct it from the externs anyway.

paf31: I think it's important to nail down the types of these functions though, since they are the main entry points for other integrations. Things like externs should allow us to change features without changing the public API.

kRITZCREEK: I'd also prefer the second option since it seems more "compositional". 

Agreed the whole Make module seems to be the entry into "PS as a library". Maybe we want to add a few codeexamples into the docstrings to show how to use these.

paf31: Generally, I'd like to reduce the visible surface area of PS as a library, and only expose high-level APIs like "compile this", "build docs for this", etc. where possible.

kRITZCREEK: It was easier to follow that approach when I didn't understand anything about the compiler... :D the more I get what it is doing the more I am tempted to couple psc-ide...

paf31:
This PR now also includes some performance improvements to the test suite - tests with a single module are run using `rebuildModule`, and we cache the externs, just like in PSCi and psc-ide, making them run quite a bit faster.

paf31:
Is there anything else to work on here?

paf31:
I'll move out the `PSCI.*` modules to `Language.PureScript.Interactive.*` as part of this PR as well.

paf31:
I'm going to spend a bit more time on this and try to extract a nicer API based on `free` which will allow us to use `Language.PureScript.Interactive` in more contexts (web browser, etc.)

Marked as `needs more work` for now.

kRITZCREEK:
Can we maybe extract the faster tests into their own PR? That sounds like something useful to have regardless of a new Interactive API. Also +1 for a free API because that will force me to finally learn about `Free` :D

paf31:
> Can we maybe extract the faster tests into their own PR?

Yes, I think so. I'll have a look at that tomorrow, hopefully.

paf31:
This is ready to review again. I'll continue to think about the `free` approach, but this can be reviewed and hopefully merged independently of that.

This exposes quite a lot of commonality between the `Ide` modules and the `Interactive` modules, which I suspect @kRITZCREEK is already aware of, but for the sake of posterity:

- Both use the command pattern
- Both use externs as state in some sense
- Both have a concept of word completion
- Both use incremental rebuild

Perhaps we could merge them over time. Also, both have features which would be useful in the other context:

- Getting the type of an arbitrary expression would be great in an editor.
- Browsing a module's exports might be nice in an editor popup, especially if we could show documentation too.
- Pursuit search in PSCi would be great.

purescript.cabal
```diff
@@ -414,7 +415,7 @@ test-suite tests
                    vector -any, utf8-string -any, silently -any
     ghc-options: -Wall
     type: exitcode-stdio-1.0
-    main-is: Main.hs
+    main-is: Test.hs
```
kRITZCREEK: If we don't pull Sources from `psci` anymore this can be reverted

purescript.cabal
```diff
@@ -430,9 +431,5 @@ test-suite tests
                    Language.PureScript.Ide.RebuildSpec
                    Language.PureScript.Ide.ReexportsSpec
                    Language.PureScript.IdeSpec
-                   PSCi.Completion
-                   PSCi.Directive
-                   PSCi.Module
-                   PSCi.Types
     buildable: True
     hs-source-dirs: tests psci
```
kRITZCREEK: Can we now get rid of the psci directories for the tests?

psci/Main.hs
```diff
@@ -0,0 +1,140 @@
...
+getOpt = Opts.execParser opts
+    where
+      opts        = Opts.info (version <*> Opts.helper <*> psciOptions) infoModList
+      infoModList = Opts.fullDesc <> headerInfo <> footerInfo
+      headerInfo  = Opts.header   "psci - Interactive mode for PureScript"
+      footerInfo  = Opts.footer $ "psci " ++ showVersion Paths.version
+
+-- | Read a file in the 'P.Make' monad, handling errors.
+readFileMake :: FilePath -> P.Make String
+readFileMake path = P.makeIO (const (P.ErrorMessage [] $ P.CannotReadFile path)) (readUTF8File path)
```
hdgarrood: Perhaps we should export `readTextFile` from `Language.PureScript.Make` instead of redefining it here?

psci/Main.hs
```diff
@@ -0,0 +1,140 @@
...
+    Just "" -> return (Right Nothing)
+    Just s | singleLineMode || head s == ':' -> return .fmap Just $ parseCommand s
+    Just s -> fmap Just . parseCommand <$> go [s]
+  where
+    go :: [String] -> InputT (StateT PSCiState IO) String
+    go ls = maybe (return . unlines $ reverse ls) (go . (:ls)) =<< getInputLine "  "
+
+-- | Checks if the Console module is defined
+consoleIsDefined :: [P.ExternsFile] -> Bool
+consoleIsDefined = any ((== P.ModuleName (map P.ProperName [ "Control", "Monad", "Eff", "Console" ])) . P.efModuleName)
```
hdgarrood: Would it make sense to have a pattern synonym for `P.ModuleName (map P.ProperName [ "Control", "Monad", "Eff", "Console" ])`?

kRITZCREEK: Could use `moduleNameFromString "Control.Monad.Eff.Console"`

hdgarrood:
👍 LGTM. Given that this will break `pulp psci`, I think we should make sure `pulp` can handle the new `psci` interface before announcing the 0.9 release. I will probably be able to do that fairly soon after this is merged.

paf31:
This is ready for review again. I've also tried to fix #1441 by moving the support module out into its own package. See https://github.com/paf31/purescript-psci-support. If this sounds like a decent approach, I'll tidy up the package and move it to core.

hdgarrood:
All looks good to me. Failing immediately is definitely an improvement 👍

paf31:
Thanks for the review!

Issue #2012, opened 2016-04-10
Planned 0.9 Breaking Changes


paf31:
- Alias changes
  - [x] Required fixity declarations for operators
  - [x] Remove non-aliased operator declarations (require `as`) (operator names cannot be introduced by any means other than alias - no function arguments with operator names anymore)
- [x] Properly implement `Partial` constraints
- [x] Remove `qualified` keyword (#1997)
- [x] Remove legacy operator sections
- [x] `class` imports and exports
- [x] Indentation fix (#1968) 
- [x] Update for names in new prelude (#1963)
- [x] Remove `--require-path` (#1996)
- [x] Rename `Object` to `Record` (#1768)
- [x] Require modules to be imported for names to resolve (#1511)
- [x] Only allow one module per file
- [x] Find foreign modules by name rather than comment-search (#2065)
- [x] Updated behaviour of imports so only exactly things that are imported are brought into scope (part of #1901)
- [ ] <s>Syntax updates for imports? (the more controversial part of #1901 still)</s>

garyb:
Remove non-aliased operator declarations

paf31:
Isn't that the same as "Required fixity declarations for operators"?

garyb:
I guess so, there are two parts to it though: requiring the `as` part of an `infix` declaration, but also removing operator parsing from declared module/class members. I'm pretty sure `Ident` can lose its `Op` constructor entirely when we do that, in fact.

adinapoli:
Sorry so much for hijacking the thread guys (feel free to delete this comment), but is there any blogpost/manuscript/testament explaining the upcoming breaking changes and why are they necessary? Of course nobody expect for a pre 1.0 project to be stable, but at the same time those described here seems (at least without any other extra context) quite massive changes affecting virtually all the projects out there. As lots of companies (including us!) are writing PureScript for a living, I would like to try understanding how bad this is gonna impact us, and how we can react on time to not sink :grin: 

Thank you for you hard work folks!

hdgarrood:
[...]

Issue #1768, opened 2015-12-28
Rename Prim.Object to Prim.Record


natefaubion:
I think `Object` is a confusing name because its not what we use anywhere in our literature, and its ultimately the JS nomenclature leaking through. It's obviously a breaking change, but it would be nice to say "the record syntax { ... } is just sugar for `Record ( ... )`".

SimonRichardson:
:+1: from me, this has always been confusing.

michaelficarra:
:+1: I like it.

paf31:
I think this makes sense, but we should do it as part of a breaking release. Move to 0.9?

garyb:
:+1:

LiamGoodacre:
:+1:

zudov:
I am not sure if that would actually work, but how about we introduce it in a non-breaking way:

```haskell
foreign import data Record :: # * -> *
type Object = Record
```

And then in 0.9 we can remove `type Object = Record`

For almost any other type this would be a breaking change for following reasons:

1. Type synonyms can't appear in instance heads.
2. IIRC, there are some differences between imports of type synonyms and imports of data types.

However in this case neither of those two problems apply, (1) doesn't apply because you are not allowed to have an `Object` in an instance head, (2) doesn't apply because `Prim` is implicitly imported.

However the error messages would change since type synonyms get fully expanded in there, but I don't think this can be considered as a breaking change.

I won't be surprised if I missed something, so please correct me.

LiamGoodacre:
@zudov you can have `Object` in an instance head:

~~~purescript
class Foo t
instance fooObject :: Foo (Object e)
~~~

garyb:
Maybe we should do this for 0.9, since we're breaking stuff?

kRITZCREEK:
ping I think this would be nice for 0.9

paf31:
Agreed, I've moved this over.

Issue #1511, opened 2015-10-06
Require modules to be imported even if their members are being referenced fully qualified


garyb:
(From a discussion on IRC).

The motivation is for enabling better tooling, and with some additional changes (require imports at the top, disallow multiple modules per file) potential faster recompilation times.

One thing that would need figuring out: how would it actually work, since we're dropping the `qualified` modifier, so it wouldn't be `import qualified X.Y` as it is in Haskell. Change `import X.Y` to not implicitly import anything and require `import X.Y (..)` as mentioned in #1509? Require `import X.Y as X.Y`?

The three of us who talked about it (@paf31, @natefaubion) don't feel particularly strongly about it, but the pros do seem to slightly outweigh the cons, so, discuss:

paf31:
I've confirmed that the rebuild slowness in the `incremental` branch is due to having to parse every module fully before building.

paf31:
I'd very much like to see this added to the "break before 0.8" list, if it can give a significant performance increase for incremental builds.

A simple way to verify it would be to truncate the module at the last line beginning with `import`, then parse as usual, figure out module dependencies, and reparse the whole module if we determine we need to rebuild it. Of course, for a final version, this won't be good enough, but it would be fine to test the idea out.

garyb:
Fine by me. Wishful thinking, but I'd like us to maybe not break anything in 0.9 (aside from things we mark as deprecated - if we do the suggested `class` imports, for instance, and dropping the `qualified` modifier) if we can manage it - so hopefully get operator aliases in 0.8 too... 

Did you have any thoughts on the "how to qualify as itself" question I mentioned in the issue?

paf31:
I had a quick go at this in the `1511` branch. The time to recompile Halogen is now 0.9s on my machine. The main obstacle seems to be lexing, since we need to lex all the way to the end of the file before we can start parsing. Ideally, we need some way of telling when the imports have finished in the lexer, which doesn't seem simple.

Ironically, if we didn't split the lexer out, this would be faster :)

Another approach might be to parallelize the reading of module headers.

AppShipIt:
Just read this because @paf31 referenced it in mine. @garyb You asked for this not to happen right? In particular to facilitate debugging. Perhaps there's a solution to both simultaneously?

garyb:
[...]

PR #2090, opened 2016-05-04
Separate the `Op` and `Ident` namespaces, add phantom type for `Op` for values/types

garyb:
And remove a few remnants of pre-0.9 operator code.

Still very much a WIP, this got a bit more involved than I expected.


paf31:
I had a go at this, but it was a bit beyond my understanding. Thanks!

src/Language/PureScript/Parser/Common.hs
```diff
@@ -44,10 +44,16 @@ parseQualified parser = part []
   qual path = if null path then Nothing else Just $ ModuleName path
 
 -- |
--- Parse an identifier or parenthesized operator
+-- Parse an identifier.
 --
 parseIdent :: TokenParser Ident
-parseIdent = (Ident <$> identifier) <|> (Op <$> parens symbol)
+parseIdent = Ident <$> identifier
```
paf31: 👍 Much nicer.

garyb:
Yeah, hopefully when this PR is done the result should be much more understandable! Since I delved into the operators lately I mostly have it loaded into my brain still.

paf31:
Also note there are some comments about 0.9 in the parser module. See `parseVarOrNamedBinder` and `parseAbs` for places where we can now reduce backtracking.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -188,9 +194,15 @@ data Declaration
...
+  -- A fixity declaration
+  -- (fixity data, operator name, value the operator is an alias for)
   --
-  | FixityDeclaration Fixity String (Maybe (Qualified FixityAlias))
+  | FixityDeclaration Fixity (OpName 'ValueOpName) (Qualified (Either Ident (ProperName 'ConstructorName)))
+  -- |
+  -- A fixity declaration for a type operator
+  -- (fixity data, operator name, type the operator is an alias for)
+  --
+  | TypeFixityDeclaration Fixity (OpName 'TypeOpName) (Qualified (ProperName 'TypeName))
```
garyb: I'm not sure about this approach yet (separating the type fixities entirely). It might result in a ton of duplicated code, but on the other hand it will eliminate a bunch of cases where we have a multiplication of possibilities due to type/value split.

I tried some stuff out with GADTs, but it quickly spiralled into madness.

paf31:
This will finish the last items in #2012, correct?

hdgarrood:
What about #1511, was that implemented already?

garyb:
#1511 will need doing too, that shouldn't be too difficult to sort out though. There's also whether we're going to do anything about the syntax for imports to go along with that. Also removing multiple modules per file, which is a one-line change again.

hdgarrood:
I guess it depends whether #2012 is intended to be a complete list of breaking changes. If it is, I think it's missing #1768 too, right?

garyb:
Good point! I think it is supposed to be a comprehensive list.

garyb:
This is now ready for review. It cleans up the remainder of the 0.9 stuff relating to operators, and I reduced the number of different types of errors we need by making wider use of the `Name` sum type, rather than it just being for the linter. Sorry the diff is so horrendous. :wink:

There's a problem in the `psc-ide` tests on OSX though that I don't really know what the deal is:

```
Failures:

  tests/Language/PureScript/Ide/Imports/IntegrationSpec.hs:46: 

  1) Language.PureScript.Ide.Imports.Integration, Adding imports, adds an implicit import

       uncaught exception: IOException of type OtherError (readCreateProcess: psc-ide-client (exit 1): failed)
```

Any ideas @kRITZCREEK?

garyb:
This will interfere with #2079 actually, but I think it might help: the printing of names should be much more consistent now, as I think they go through the same function. Aside from type variables and row labels.

hdgarrood:
From the [travis log](https://travis-ci.org/purescript/purescript/jobs/129143311#L798):
```
psc-ide-server: bind: resource busy (Address already in use)
```

garyb:
Oh, oops, I didn't see that up there, just the failure output. That's weird.

kRITZCREEK:
Omw home right now, I'll check soon but it might be related to the psc-ide network PR. Most likely a server instance didn't die fast enough.

hdgarrood:
I guess it's because it was from the output of a subprocess (`psc-ide-server`)? Yeah, I was about to say, perhaps the `setup` and `teardown` functions should use `waitForProcess` to ensure the next test doesn't start prematurely: http://hackage.haskell.org/package/process-1.4.2.0/docs/System-Process.html#v:waitForProcess

garyb:
No joy. Will have to try the `waitForProcess` approach. Strange that this doesn't break elsewhere though.

garyb:
There we go. Seems quitting the server before the tests started was causing the problem for some reason.

paf31:
:ship it!

PR #2094, opened 2016-05-06
NoImplicitPrelude, enable some global extensions

### LINKED ISSUE #1672: Further reduce import boileplate using `NoImplicitPrelude`

garyb:
Resolves #1672 

The only extension here I'd consider controversial is `PatternSynonyms`, as it reserves the name `pattern`, but I think we're going to be using them more in the future, so probably worth it? The rest are pretty widely used already.

I didn't turn GADTs on, which was suggested in the issue, because doing so resulted in compilation errors.

### DIFF
--- a/psc-ide-server/Main.hs
+++ b/psc-ide-server/Main.hs
@@ -12,7 +12,6 @@
 -- The server accepting commands for psc-ide
 -----------------------------------------------------------------------------
 
-{-# LANGUAGE ConstraintKinds       #-}
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings     #-}
--- a/psci/PSCi/Directive.hs
+++ b/psci/PSCi/Directive.hs
@@ -1,18 +1,6 @@
------------------------------------------------------------------------------
---
--- Module      :  Directive
--- Copyright   :
--- License     :  MIT
---
--- Maintainer  :
--- Stability   :  experimental
--- Portability :
---
 -- |
 -- Directives for PSCI.
 --
------------------------------------------------------------------------------
-
 module PSCi.Directive where
 
 import Prelude ()
@@ -116,4 +104,3 @@ help =
   , (Show,    "import",   "Show all imported modules")
   , (Show,    "loaded",   "Show all loaded modules")
   ]
-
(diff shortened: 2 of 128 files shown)


hdgarrood:
:+1: for PatternSynonyms. LGTM

paf31:
The only one I'd be wary of is `GeneralizedNewtypeDeriving` since it has some odd interactions with other extensions.

garyb:
Ok, I'll take it out. I think that was one of the lesser used.

paf31:
Looks good otherwise!

While we're changing every module though, maybe this is a good time to minimize and tidy up the imports, if there is a simple way to do it. I know GHC used to have a flag which would do it for you. I'll look for it ...

garyb:
I tidied up the imports a little while I was going through - made them alphabetical, etc. but I'm pretty sure we don't have any unnecessary as that would result in a warning, wouldn't it? Or did you have something else in mind?

paf31:
I was thinking we could make them all explicit, having GHC do the work. Also, this is minor, but I'd like to space out the non-qualified imports the way @kRITZCREEK does.

garyb:
Really? I've become quite anti whitespace alignment over time, it's pretty annoying to maintain due to having to re-align everything if you introduce a slightly longer name, makes for worse diffs, etc.

paf31:
I'm not too bothered about that one, but I would like imports ideally to be sorted by module name, and explicit in what they import.

But we can do that in another PR...

kRITZCREEK:
I'm using stylish-haskell which does all the formatting for imports for me...

paf31:
Perhaps we should just agree on a style and have stylish-haskell enforce it.

garyb:
Well they're alphabetical aside from `qualified` now, at least :smile:.

kRITZCREEK:
I think for a complete style we'd need to add `hindent` which can feel a little opinionated from time to time but removes the burden of having to worry about formatting at all.

garyb:
Updated to remove GND. What are the interactions by the way? I could only find something about it in combination with `DeriveAnyClass`.

paf31:
That's the one I was thinking of.

PR #2098, opened 2016-05-08
Fix #2092, precedence of type application

### LINKED ISSUE #2092: Parsing of type operators isn't quite right
They bind too tightly, 

``` purescript
StateF s ~> StateF s'
```

is not parsed as:

``` purescript
(StateF s) ~> (StateF s')
```

but instead something like:

``` purescript
StateF (s ~> StateF) s'
```

paf31:
This doesn't change the precedence of `->`, but it adds a test to make sure it binds less tightly than other type operators.


garyb:
:+1:

Issue #2034, opened 2016-04-18
Travis does not use stack for deployment builds


hdgarrood:
This is bad, because the build-stack script assumes that you are using stack. This will mean that the deployment will fail, because cabal and stack put executables in different places, but also if it somehow succeeds, it will probably put an incorrect list of dependencies into dependencies.txt in the bundle. I think we will probably have to create binary bundles manually until this is fixed. 

@phadej following from your comment: https://github.com/purescript/purescript/pull/1938#issuecomment-194507533 is it a problem if we switch to `stack` in CI builds, and stop using `cabal-install` entirely? They're both using `Cabal` under the hood, and I don't think I've ever seen an instance where something could be compiled with `stack` but not with `cabal-install`. I guess you could have a case where `cabal` ends up doing a massive search for dependencies that takes ages, but given that we know which Stackage snapshots to use, it ought to be very easy to work out what constraints to add to get it to work, right?

I would like to stop using `cabal-install` if we can because only using stack (like we do in appveyor currently) makes everything *so* much simpler.

/cc @jplatte

hdgarrood:
See also #1974.

paf31:
👍

paf31:
Can this be merged with #1974?

hdgarrood:
I guess, yeah.

paf31:
Let's make this a priority for 0.9, since the release build is breaking right now (`bundle.sh not found`).

PR #2099, opened 2016-05-08
Switch to stack for CI, resolves #1974

### LINKED ISSUE #1974: Use Stack/LTS for CI builds
As long as we have reproducible builds and a simple way to get new contributors up and running, I don't see a need to build against multiple GHC versions any more, since Stack will download the correct version of GHC anyway.

Why not just standardize on a single LTS version (which can change over time)?

hdgarrood:
* Remove the stack-nightly.yaml, since it always needs updating.
  Instead, we pass --resolver=nightly in one of the build jobs to ensure
  we always have the latest nightly. As an added bonus we no longer have
  to set the STACK_YAML environment variable on Windows, because
  stack.yaml is no longer a symlink.
* Remove use of cabal-install, which drastically simplifies the CI
  scripts :)
* Use the 'compiler' key in the build matrix as a hack to ensure each
  build job gets a separate cache.
* Add OSX tests.
* Add haddock tests, on a separate build job (for speed).
* Use separate build jobs for sdist tests (for speed).
* Fix binary bundle uploading on Travis
* Add binary bundle uploading for OSX.

Coverage did not work when I was building on Travis with my fork, but I think that was because of an API key environment variable missing, or something like that. So hopefully it should work here.


hdgarrood:
Resolves #2034.

hdgarrood:
In case you were wondering, I've canceled the Travis CI build for the second commit. I'm going to restart it when the first one finishes so that it can use the cache.

paf31:
Great work, thanks very much! 

Just a note: this also seems to remove the `core-tests` run. I'm not sure if that was even working on `0.9` anyway due to the breaking changes, but it'll be worth trying to reinstate it at some point. We'll have to remember to run it manually before releasing 0.9.

hdgarrood:
👍 , opened #2101

paf31:
So this is ready to merge?

hdgarrood:
I think so.

paf31:
Thanks very much!

PR #2102, opened 2016-05-08
Rename Prim.Object to Prim.Record, fix #1768

### LINKED ISSUE #1768: Rename Prim.Object to Prim.Record

paf31:


src/Language/PureScript/Environment.hs
```diff
@@ -196,16 +196,16 @@ tyArray :: Type
...
--- Type constructor for objects
+-- Type constructor for records
 --
-tyObject :: Type
-tyObject = primTy "Object"
+tyRecord :: Type
+tyRecord = primTy "Record"
 
 -- |
 -- Check whether a type is an object
```
kRITZCREEK: a record

paf31: Fixed, thanks.

garyb:
:ship:

PR #2110, opened 2016-05-11
Speed up tests by using rebuildModule

paf31:
This is extracted from my PSCi pull request. It makes the tests run significantly faster by using `rebuildModule`. It also includes some minor refactoring which was necessary to support that.


kRITZCREEK:
Looks good to me :+1:

garyb:
This is now conflicting after merging the Op/Ident thing, but :+1:!

paf31:
Standard Travis failure on MacOS.

PR #2112, opened 2016-05-11
[psc-ide] adds a reset command

kRITZCREEK:
Let's see how this performs. It's actually so really strange...

PR #2114, opened 2016-05-12
Small fixes

kRITZCREEK:
Just some things I found while browsing the code.


garyb:
> Coverage increased (+0.001%)

It's still an increase! :laughing: 

:+1:

garyb:
OSX bug is back? :confused: 

I restarted the build as it only failed on one of the osx images.

kRITZCREEK:
see #2112 . I'm hoping that will fix the OSX errors for good

paf31:
Looks good, but if we're supporting base 4.7 now, let's remove `traverseEither` from `Language.PureScript.Make` too.

kRITZCREEK:
removed `traverseEither` and fixed conflicts

PR #2116, opened 2016-05-12
Fix float parsing

hdgarrood:
The diff for ExportedInstanceDeclaration has all changed because I converted it to unix line endings, but the only meaningful change is the module name (from ExportedInstanceDeclarations to Main).


examples/passing/ExportedInstanceDeclarations.purs
```diff
@@ -1,44 +1,45 @@
...
--- Case 2: type class
-instance nonExportedNonexportedType :: NonexportedClass (Const Int a) where
-  notExported = Const 0
-
--- Case 3: instance types
-instance constFoo :: Foo (Const NonexportedType b) where
-  foo = Const NonexportedType
-
-main = log "Done"
+-- Tests that instances for non-exported classes / types do not appear in the
```
AppShipIt: Did line endings change the diff is funny? I was just looking to learn 😉

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -254,7 +254,7 @@ parseToken = P.choice
 
   parseNumber :: P.Parsec String u (Either Integer Double)
   parseNumber = (consumeLeadingZero >> P.parserZero) <|>
-                  (Right <$> P.try (PT.float tokenParser) <|>
+                  (Right <$> P.try parseDouble <|>
```
AppShipIt: Would it be worth commenting the ticket number here in case anyone stumbles on it and wonders why Parsec wasn't used?

examples/passing/ExportedInstanceDeclarations.purs
garyb: From Harry's comment:

> The diff for ExportedInstanceDeclaration has all changed because I converted it to unix line endings, but the only meaningful change is the module name (from ExportedInstanceDeclarations to Main).

:stuck_out_tongue:

hdgarrood: When I opened this file, Vim told me it had DOS line endings, which I assume means `\r\n`, so I typed in `:set ff=unix` which I think converts them all. I'm not really sure what's going on with the diff either.

garyb: It's almost certainly my fault - just had a similar problem with Halogen. I must have mangled my git settings somehow.

AppShipIt: I meant to check the comments sorry!

src/Language/PureScript/Parser/Lexer.hs
garyb: I think the comment on `parseDouble` itself is probably enough to explain it.

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -263,6 +263,54 @@ parseToken = P.choice
     consumeLeadingZero = P.lookAhead (P.char '0' >>
       (P.notFollowedBy P.digit P.<?> "no leading zero in number literal"))
 
+-- This implementation is based on the Haskell 98 report. We avoid Parsec's
```
AppShipIt: Ignore earlier comment I should've read further 😅

src/Language/PureScript/Parser/Lexer.hs
AppShipIt: Missed it first time it's quite thorough 👍

hdgarrood:
Just FYI I also opened https://github.com/aslatter/parsec/pull/54, although I wouldn't rely on that being merged quickly. When/if that's merged, we could revert this code and add a lower bound on the first fixed version to make sure this doesn't reappear.

examples/passing/ExportedInstanceDeclarations.purs
jplatte: To prevent commits with CRLF in the future, we could add a `.gitattributes` file with just `* text=auto` in it (tells git to automatically convert file endings if needed): https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings

paf31:
👍 Looks great. Does the current LICENSE file cover the stuff from the report?

hdgarrood:
Ooh, yes, the current LICENSE mentions the report, but probably isn't sufficient. I'll check the report's terms.

hdgarrood:
Here are the terms:

> Copyright (c) Simon Peyton Jones.
The authors intend this Report to belong to the entire Haskell community, and so we grant permission to copy and distribute it for any purpose, provided that it is reproduced in its entirety, including this Notice. Modified versions of this Report may also be copied and distributed for any purpose, provided that the modified version is clearly presented as such, and that it does not claim to be a definition of the language Haskell 98.

Taken from https://www.haskell.org/onlinereport/.

I guess we could add that to the LICENSE together with a short paragraph explaining that some code in the compiler is derived from the report?

hdgarrood:
My fix got merged into Parsec so I'll put this on hold for now. If a new Parsec release comes out fairly soon I'll send a new PR adjusting the dependency bounds instead.

hdgarrood:
A new version of parsec has been released with the fix, so we probably won't need this after all.

Issue #2065, opened 2016-04-29
psc-ide and psc consistency finding FFI


AppShipIt:
I'm trying to set up our mixed project to keep `.purs` and `.js` files in the same directory tree. @ethul's PureScript WebPack plugin [requires](https://www.npmjs.com/package/purescript-webpack-plugin#ffi-string-array) a glob of FFI files as input. @spicydonuts suggested naming FFI `*.purs.js` for the WebPack plugin. But I remembered @kRITZCREEK mentioning in https://github.com/purescript/purescript/pull/2013#issuecomment-210560323 that `psc-ide` expects those to be named like `[ps-name].js` (which sounded like a reasonable assumption) for fast recompilation. @hdgarrood also [voiced](https://twitter.com/hdgarrood/status/725955514249568257) that `psc-ide` finding FFI differently than `psc` violates principle of least surprise. Just reporting a use case we're working around. I don't have a suggestion how it should work differently.

kRITZCREEK:
One thing to add, is that you want your FFI files to be TypeScript/ES2015 transpiled before fed into the compiler by the webpack plugin.

hdgarrood:
I think 'surprise' isn't quite the right word for what I'm worried about. My aim, generally, is to minimise unnecessary variations in how projects are set up. The more variation there is, the less likely any two random build tools are likely to work nicely (or even at all) together. This same idea, and particularly because we were getting reports of build tools breaking for lots of users, is what led me to advocate removing the `--require-path` option.

Given the following observations:
* Almost all PureScript code in the wild uses one module per file,
* Almost all PureScript code in the wild uses a predictable scheme for locating the path of FFI files (i.e. `moduleFileName.replace('.purs', '.js')`,
* `psc-ide-server` assumes that your code is arranged as above,
* this is different from how `psc` works (multiple modules per file, scan FFI modules for module comments to match them up),

I worry that we're setting ourselves up for unnecessary pain; at the moment, we have two separate ways of indicating to compiler tools how normal and FFI modules match up. Not only does this mean we're asking users to do extra unnecessary work to get things to work, I would also bet that this is going to come back and bite us sooner or later.

To get the two in sync, we will have to either make `psc` stricter in what it accepts, or `psc-ide-server` looser. I vote the former; not only does it seem very feasible, but also it's in line with the language's philosophy (of adding power by removing unnecessary flexibility, eg. forcing `Eff _ a` to be different from `a`).

[...]

PR #2118, opened 2016-05-13
Find foreign modules by name rather than comment-search

paf31:
This changes `psc` but not `psci` (I'll wait for my other PR to be merged first). It changes the command line interface of `psc`, removing the `--ffi` option, and infers the foreign modules based on filename.


AppShipIt:
Fixes #2065.

paf31:
This is ready for review, although we'll probably want to commit to one-module-per-file if we decide to merge it.

src/Language/PureScript/Make.hs
```diff
@@ -297,6 +289,24 @@ makeIO f io = do
...
+  :: forall m
+   . MonadIO m
+  => M.Map ModuleName (Either RebuildPolicy FilePath)
+  -> m (M.Map ModuleName FilePath)
+inferForeignModules = fmap (M.mapMaybe id) . traverse inferForeignModule
+  where
+    inferForeignModule :: Either RebuildPolicy FilePath -> m (Maybe FilePath)
+    inferForeignModule (Left _) = return Nothing
+    inferForeignModule (Right path) = do
+      let jsFile = replaceExtension path "js"
```
hdgarrood: I thought the plan was `path ++ ".js"`? So, eg, you might have `src/Data/Show.purs.js`.

hdgarrood:
Looks good to me (once we've sorted out the issue of `.js` vs `.purs.js`).

garyb:
Yeah, I think that was the proposal. I think I prefer just `.js` personally, but I don't feel that strongly about it, although I think changing to `.purs.js` will break psc-ide.

hdgarrood:
Oh yeah, I think we should change the psc-ide rebuild code to use the same function to infer foreign module paths; whatever we decide on, they should use the same mechanism, right?

I don't feel strongly either, but the argument was that `.purs.js` makes it easier to mix regular JS with purescript. It's unfortunate that it will break pretty much everything, but I guess we're doing that already anyway.

AppShipIt:
> `.purs.js` makes it easier to mix regular JS with purescript.

Yeah so build tools can decipher which to feed to psc, so psc-ide can be consistent with psc (it currently uses a different criteria to identify FFI files), but also keeps us from having to parse a relatively arbitrary header convention which could make it trickier e.g. for loaders that prepend a copyright header.

garyb:
@AppShipIt the modification here also avoids the need for the header parsing, the difference being it's currently looking for `module.js` in the same directory as the module, not `module.purs.js`.

hdgarrood:
As of 0.9, this PR makes it so that you only have to supply the purs files to `psc`, and the compiler will pick up the rest automatically based on filename. Come to think of it, I guess there's no reason to change the convention at all; build tools no longer need to worry about JS files now. I've just pushed some commits for 0.9 compatibility for Pulp into a branch and the code there will work regardless of which option we choose.

AppShipIt:
Huh I don't immediately see any downsides to that. @ethul Your loader still has to implicitly require those JS files. Would that too difficult with `Module.js` vs `Module.purs.js` CC @spicydonuts.

hdgarrood:
Can you explain what "implicitly require" means in this context?

AppShipIt:
WebPack builds a dependency graph by following imports regardless which module system they use. WebPack would know the `*.purs` needs processed but not to include the FFI in any other processing including bundling. Since there's no explicit require @ethul's loader should add it through the <q>loader context</q> API.

paf31:
Since we don't pass `.js` files to `psc` any more, I would prefer to keep the current approach and reduce the amount of code which needs to be changed to migrate to 0.9.

AppShipIt:
Still should probably allow @ethul to weigh in since you original motivation was interop? I think it should work I'm just not positive.

paf31:
Of course.

ethul:
Thanks for the heads up.

@AppShipIt I don't think the choice of naming convention would pose an issue for the webpack loader. It could discover the FFI dependencies using the new convention and implicitly add the FFI as a webpack dependency, as you mention.

Also, if @alexmingoia's ethul/purs-loader#47 is merged, then it may be a non issue since the FFI dependency would be picked up as a dependency directly from the require statement in the compiled purescript file.

@alexmingoia, I don't know if you want to add anything on this?

AppShipIt:
In that case I don't see any reason to advocate `*.purs.js`.

paf31:
So, good to merge then?

hdgarrood:
Yeah, I think this is good to merge; if it does later turn out we want .purs.js for some reason we haven't thought of here, we can always change that later.

Also cc @spicydonuts who I think suggested .purs.js in the first place.

hdgarrood:
Oh wait, can we also change psc-ide to use the same function so that these are less likely to accidentally diverge again?

paf31:
Done.

spicydonuts:
Thanks for the cc, no objections here. This removes the problem without breaking the current convention, rather than just patching it with new convention 😄

src/Language/PureScript/Ide/Rebuild.hs
```diff
@@ -37,23 +35,25 @@ rebuildFile path = do
...
+  foreigns <- P.inferForeignModules filePathMap
 
-  let ma = P.buildMakeActions outputDirectory
-                              (M.singleton (P.getModuleName m) (Left P.RebuildAlways))
-                              (if foreignExists
-                                 then M.singleton (P.getModuleName m) foreignModule
-                                 else M.empty)
-                              False
+  -- Silence progress update messages during the build
+  let actions = (P.buildMakeActions outputDirectory filePathMap foreigns False)
```
kRITZCREEK: These changes will conflict with my cache-rebuild branch since I already made similar changes to the action building

paf31: I'm happy to do the merge.

kRITZCREEK: Actually don't worry about it, I'll do it after this is merged.

paf31:
I'm going to merge this, since CI is actually finished apart from one job which I had to stop due to a Stack issue.

PR #2119, opened 2016-05-13
Check that passing tests have a Main module

hdgarrood:
Thought that this should probably come separately. This PR adds an assertion that all the passing tests have a Main module. Previously, if you called the main module something else, it would compile the code, but then run the code in a different Main module left over from the previous test, so any failures at runtime would be missed.

### DIFF
--- a/examples/passing/ExportedInstanceDeclarations.purs
+++ b/examples/passing/ExportedInstanceDeclarations.purs
@@ -1,9 +1,10 @@
 -- Tests that instances for non-exported classes / types do not appear in the
 -- result of `exportedDeclarations`.
-module ExportedInstanceDeclarations
+module Main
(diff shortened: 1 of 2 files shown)


garyb:
Thanks for adding this :+1:

hdgarrood:
CI failures are just OSX again, so I think we can safely ignore those for now.

garyb:
That's annoying, the tests should be using the reset command now, which we were hoping would solve that issue.

hdgarrood:
I think now it's just that the server doesn't start quickly enough, or the tests start before the server is ready. @kRITZCREEK said in IRC a moment ago that he was looking at it.

paf31:
👍 Thanks!

PR #2121, opened 2016-05-13
poll to check if the server is running

kRITZCREEK:
throw an exception after 2.5 sec

Let's see if this fixes the Travis errors on OSX.

EDIT: How long should we wait?

### DIFF
--- a/tests/Language/PureScript/Ide/Integration.hs
+++ b/tests/Language/PureScript/Ide/Integration.hs
@@ -31,6 +31,7 @@ module Language.PureScript.Ide.Integration
        , addImplicitImport
        , loadModule
        , loadModuleWithDeps
+       , getCwd
        , getFlexCompletions
        , getType
        , rebuildModule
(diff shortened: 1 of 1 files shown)


garyb:
Looks like that did it!

hdgarrood:
I guess 2.5 seconds is probably enough, we can always increase the number of tries later if it turns out we need to?

kRITZCREEK:
I checked the ouput on Travis and it seemed like the psc-ide tests waited a while before going off so it seems to work :tada:

garyb:
:shipit: assuming AppVeyor passes too :wink:

kRITZCREEK:
there is an old AppVeyor instance running and it doesn't seem to terminate.... grml looks like I'll have to try and quit the server both via its's handle and the quit command...

kRITZCREEK:
It's fine to call quitServer (send a command to the server) at that point, because if we get there we can be sure that the server is already running.

kRITZCREEK:
consistently worked on OSX and now also terminates on windows

hdgarrood:
🎉 🎉 🎉

paf31:
💯 Thanks!

PR #2129, opened 2016-05-15
Use Map rather than list in Exports

### LINKED ISSUE #2124: Possible compile time optimization in Sugar/Names/Env.hs?
Hi everyone,

I have a fairly large purescript project. As the project grows, my compile times worsen. Currently, it takes a little over 3 minutes to perform a fresh build. I am using psc-0.8.5.0.

```
$ find output -type f | wc -l
    1178
$ find output -type f | xargs cat | wc -l
   77524
```

Here's an example fresh build: removing the output directory, on an 8 core macbook pro:
```
* Build successful.
real    3m4.977s
user    12m59.221s
sys     0m56.923s
```

I recently compiled psc with profiling enabled so that I could obtain a .prof dump (pulp build +RTS -p -RTS):
- https://gist.github.com/adarqui/b83ecd2e94db1e3e89a07f4dcbdfe55e

As you can see from that dump, most of the time is spent in the addExport function:

```
==                        Language.PureScript.Names            46.6    0.0
addExport                 Language.PureScript.Sugar.Names.Env  44.9    5.0
```

Is it possible that module could be optimized? Maybe using a Map instead of lists? I have no idea.

Hope that dump actually helps & makes sense.

Thanks!

-- Andrew

btw, how I built psc using profiling:
```
stack build --executable-profiling --ghc-options="-fprof-auto -rtsopts"
```

### LINKED ISSUE #2142: Remote connection to PSCi (REPL)
Hi,

Didn't found anything - is it possible to evaluate code in remote clients (like browser/mobile devices,etc) using repl? 

According to this [PSCi.hs#L194-L208](https://github.com/purescript/purescript/blob/master/psci/PSCi.hs#L194-L208) psci runs Node.JS process env evaluates code there. Other tools/languages like ClojureScript/ghcjs using a (web)socket in order to build a bridge between different clients and a repl process, can we do the same here?

I would be more than happy to contribute as it's an only missing feature for me in PureScript, although I'm very new to Haskell and PureScript, so I would appreciate some help.

Thanks

garyb:
Resolves #2124 (maybe). It avoids the lists anyway, and improves the code in most places where the list was used previously.

### DIFF
--- a/src/Language/PureScript/Docs/Convert/ReExports.hs
+++ b/src/Language/PureScript/Docs/Convert/ReExports.hs
@@ -4,7 +4,7 @@ module Language.PureScript.Docs.Convert.ReExports
 
 import Prelude.Compat
 
-import Control.Arrow ((&&&), first, second)
+import Control.Arrow ((&&&), second)
 import Control.Monad
 import Control.Monad.Reader.Class (MonadReader, ask)
 import Control.Monad.State.Class (MonadState, gets, modify)
@@ -96,7 +96,7 @@ getReExports env mn =
 --      * Filters type class declarations to ensure that only re-exported type
 --      class members are listed.
 --
-collectDeclarations ::
+collectDeclarations :: forall m.
   (MonadState (Map P.ModuleName Module) m, MonadReader P.ModuleName m) =>
   P.Imports ->
   P.Exports ->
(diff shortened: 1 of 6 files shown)


hdgarrood:
👍 agreed, seems like an improvement to me even if we don't consider performance.

hdgarrood:
I think it's #2124 that this (hopefully) fixes, incidentally.

garyb:
Oops, right you are!

adarqui:
Hi garyb,

I think i'm going to add this to a 0.8.5 branch and test it with my code. If it works, i'll report back if there are (or aren't) any compile speed improvements.

Thanks!

paf31:
Looks good! How does the profiler output look now?

Note (not sure if you're aware or not, but I'll record it here anyway) that Stack will make a profiling build with

```
stack build --executable-profiling --library-profiling --ghc-options="-fprof-auto -rtsopts"
```

If you run `psc` with `+RTS -p`, you'll get a profiling report, and with `-hc` will give a heap profile.

Creating a separate Stack working directory with `--work-dir` can also be useful, since you'll end up doing a lot of rebuilding with and without profiling options otherwise.

garyb:
I haven't got anything significant I can try it on yet, since this is the 0.9 branch.

paf31:
Ah right, of course. Even building something like `purescript-strings` should give you a reasonable profile to look at though.

Issue #1901, opened 2016-02-28
Qualify imports by default


texastoland:
Continued from #1869:

## Changes

1. `import X (..)` ⇠ `import X` (to make open imports explicitly opt-in via previous PR)
2. `import Control.Monad.X` ⇠ `import Control.Monad.X as X` (to make qualified imports easy via @tfausak)
3. <s>`import X hiding (f)` only exposes `A` and not `X.A` or `X.f` (via @gary)</s> https://github.com/purescript/purescript/pull/2131#issuecomment-219313295
4. <s>`import X` required to expose `X.A` fully qualified (via @paf31)</s> #1511

## Grammar

```ts
;// except for  `".." |` this is the same grammar as today
<identifiers>  = ".." | <identifier> { "," <identifier> }
<specifiers>   = [ "hiding" ] "(" <identifiers> ")"
<qualifier>    = "as" <qualified name>
;// <import>   = "import" <module name> [ <specifiers> ] [ <qualifier> ]
<import long>  = "import" <module name>   <specifiers>   [ <qualifier> ]
<import short> = "import" <module name>                  [ <qualifier> ]
<import>       = <import long> | <import short>
```

In the notation above:

- `<>` reference previous rules
- `[]` represent zero or one (`?`in regex)
- `{}` represent zero or many (`*` in regex)
- `|` represents one or the other (same as regex)
- `;` introduces a comment

The rule could technically have been written as `<import>` but the short form is nuanced sugar for the long one:

- Long form introduces a set of `<identifiers>` (`hiding` negates the set or `..` subsumes it) into the `<qualified name>` (or else unqualified local scope).
- Short form assumes the set to be `..`. Unlike long form the `<qualified name>` defaults to the trailing part of the module name (e.g. `X` in `Control.Monad.X`). Any `<import short>` can be rewritten as `<import long>`:
  - `import Control.Monad.X` ⇠ `import Control.Monad.X (..) as X`
  - `import Control.Monad.X as Y` ⇠ `import Control.Monad.X (..) as Y`

## Examples

Import | Exposing | Module | Identifiers
---|---|---|---
`import Control.Monad.X` | `X.A`, `X.f` | qualified | all
`import X (A)` | `A` | unqualified | inclusive
`import X (..)` | `A`, `f` | unqualified | all
`import X hiding (f)` | `A` | qualified | exclusive
`import X as Y` | `Y.A` `Y.f` | qualified | all
[...]

PR #2131, opened 2016-05-15
Name resolving

### LINKED ISSUE #1511: Require modules to be imported even if their members are being referenced fully qualified

### LINKED ISSUE #2148: Unspecified import suggestions do not include operators
This is probably because operators have been desguared at this point.

garyb:
This is still WIP. Trickier than it first seemed, since it seems we introduce qualified names in a few places during desugaring, so when we do so we need to update the imports also... or something.

Resolves #1511, fixes #2148.

### DIFF
--- a/src/Language/PureScript/Sugar/Names.hs
+++ b/src/Language/PureScript/Sugar/Names.hs
@@ -16,7 +16,7 @@ import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State.Lazy
 import Control.Monad.Writer (MonadWriter(..), censor)
 
-import Data.List (find, nub)
+import Data.List (nub)
 import Data.Maybe (fromMaybe, mapMaybe)
 import qualified Data.Map as M
 import qualified Data.Set as S
@@ -65,9 +65,9 @@ desugarImportsWithEnv externs modules = do
   externsEnv env ExternsFile{..} = do
     let members = Exports{..}
         ss = internalModuleSourceSpan "<Externs>"
-        env' = M.insert efModuleName (ss, nullImports, members) env
+        env' = M.insert efModuleName (ss, primImports, members) env
         fromEFImport (ExternsImport mn mt qmn) = (mn, [(Nothing, Just mt, qmn)])
-    imps <- foldM (resolveModuleImport env') nullImports (map fromEFImport efImports)
+    imps <- foldM (resolveModuleImport env') primImports (map fromEFImport efImports)
     exps <- resolveExports env' ss efModuleName imps members efExports
     return $ M.insert efModuleName (ss, imps, exps) env
     where
@@ -101,7 +101,7 @@ desugarImportsWithEnv externs modules = do
       Just m' -> throwError . errorMessage $ RedefinedModule mn [envModuleSourceSpan m', ss]
       Nothing -> do
         members <- findExportable m
-        let env' = M.insert mn (ss, nullImports, members) env
+        let env' = M.insert mn (ss, primImports, members) env
         (m', imps) <- resolveImports env' m
         exps <- maybe (return members) (resolveExports env' ss mn imps members) refs
         return (m' : ms, M.insert mn (ss, imps, exps) env)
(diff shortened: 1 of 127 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -582,6 +584,10 @@ prettyPrintSingleError full level showWiki e = flip evalState defaultUnknownMap
       line $ "Cannot export unknown " ++ printName (Qualified Nothing name)
     renderSimpleErrorMessage (UnknownExportDataConstructor tcon dcon) =
       line $ "Cannot export data constructor " ++ runProperName dcon ++ " for type " ++ runProperName tcon ++ ", as it has not been declared."
+    renderSimpleErrorMessage (UnimportedModule mn) =
+      paras [ line $ "Unknown module " ++ runModuleName mn
+            , line "A module with this name does exist but has not been imported into the current file."
```
paf31: Should this be "but has been imported"?

paf31:
Is this #1511?

garyb:
Yep, although it also includes part of the proposal for #1901, as the requirements overlap: if you `import M (x)` you'll only get `x` in scope, not `M.x` as well as `x`.

I'm having second thoughts about that though, as it's partially what is causing the current failure... plus it will have to behave a bit differently for module re-exports to work as they do now, since if you `import M (x)` you can still export `module M`, even though `M` "doesn't exist" when resolving names in the module.

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -219,10 +221,9 @@ renameInModule env imports (Module ss coms mn decls exps) =
     (,) s <$> (ConstructorBinder <$> updateDataConstructorName name pos <*> pure b)
   updateBinder s@(pos, _) (OpBinder op) =
     (,) s <$> (OpBinder <$> updateValueOpName op pos)
-  updateBinder s (TypedBinder t b) = do
-    (s'@ (span', _), b') <- updateBinder s b
```
garyb: This was unnecessary, the traversal already steps into the binder (duh).

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -244,14 +245,18 @@ renameInModule env imports (Module ss coms mn decls exps) =
     updateType :: Type -> m Type
     updateType (TypeOp name) = TypeOp <$> updateTypeOpName name pos
     updateType (TypeConstructor name) = TypeConstructor <$> updateTypeName name pos
-    updateType (ConstrainedType cs t) = ConstrainedType <$> updateConstraints pos cs <*> pure t
+    updateType (ConstrainedType cs t) = ConstrainedType <$> traverse updateInConstraint cs <*> pure t
     updateType t = return t
+    updateInConstraint :: Constraint -> m Constraint
+    updateInConstraint (Constraint name ts info) =
+      Constraint <$> updateClassName name pos <*> pure ts <*> pure info
```
garyb: Like that above case, need a different version of `updateConstraints` here that does not recurse on `updateTypesEverywhere`, as the traversal already does that.

These two traversal hiccups became apparent as the naming pass is even less idempotent than it was before. I'm pretty sure a case could be constructed that would cause both of these to fail.

garyb:
Think I've got this figured out now. It's working the way where only exactly what you import can be referenced when it comes to values, types, etc. but you can still `export SomeModule` where `SomeModule` was not imported qualified. So `import M (x)` only introduces `x`, not `M.x`, as mentioned above.

The traversal cases I commented on are the thing that mainly broke this before (aside from an update needed to `purescript-partial`)... yeah, it took me this long to track them down!

garyb:
The warnings causing the compile to fail should be fixed now.

garyb:
...forgot to update `extra-source-files`. 15th time lucky?

garyb:
Rebased for the recent merges.

garyb:
Updated with a fix for #2148. It may also have been causing 2148. :wink:

paf31:
Looks great!

PR #2132, opened 2016-05-15
Fix #2070, typed hole errors now include environment information

### LINKED ISSUE #2070: Typed hole errors should include environment information
This is really helpful for top-down development in `ghc`.

paf31:
E.g. in PSCi:

```text
> :t \e -> either ?a (const e) (Left 1)
Error found:
in module $PSCI
at  line 1, column 8 - line 1, column 35

  Hole 'a' has the inferred type

    Int -> t0

  in the following context:

    e :: t0
    it :: t0 -> t0


in value declaration it

where t0 is an unknown type
```

I would also like to look at ordering the `Environment`, possibly using [`insert-ordered-containers`](https://hackage.haskell.org/package/insert-ordered-containers), so that the most useful info appears at the top.


garyb:
This looks great :+1: 

Thinking about using `t` for the variables again though: couldn't we (in theory, with some bad luck) end up with `t1` referring to a type variable named `t` in the same error message as an unknown 1?

paf31:
Yes, we could. The logic is that "where t0 is an unknown type" should help to disambiguate the two.

garyb:
You wouldn't know which `t1` was the unknown though, if there were two `t1`s in scope.

paf31:
That's a good point. Maybe we should just fix this in the renaming (`TypeMap`) pass in `Errors` then?

paf31:
@garyb I've changed this to avoid overlaps during renaming in `Errors.hs`.

paf31:
```text
> :t foldl ?step 0
Error found:
in module $PSCI
at  line 1, column 1 - line 1, column 11

  Hole 'step' has the inferred type

    Int -> t0 -> Int


in value declaration it

where t0 is an unknown type
```

Now I'll finally get the argument to `foldl` right 👍

paf31:
@garyb Does this look okay now? If so, I'll merge once CI finishes, and make a release candidate. Thanks.

garyb:
Looks great :+1: 

Should we include the same info for type wildcards too?

paf31:
We _should_, but that's a bigger change which involves passing that info around during type checking, so I was going to leave it for later.

garyb:
Ah ok, I didn't realise it would be any different than this.

paf31:
Well, skolems already carried this data, but unknowns didn't. Now we've separated the old `monad-unify` stuff out, that should be easier.

paf31:
Ack, meant to `--squash`. Oh well, all of the commits build, so it's not the end of the world.

Alright, RC time?

Issue #1764, opened 2015-12-27
Constraint kinds


paf31:

LiamGoodacre:
:+1: Yes please :smile:

paf31:
Pushing to post-1.0. There are some open questions here still, specifically around code generation and the way imports/exports work.

PR #2133, opened 2016-05-16
Constraint Kinds

paf31:
Still in progress. In particular:
- [ ] Need to figure out namespacing issues. Type classes need to be in scope as types of kind `Constraint`.
- [x] Parsing needs some work, since the LHS of `_ => _` is now a type, so parens make the parse ambiguous.


paf31:
@garyb I might need a bit of help with imports on this one. The basic idea is that every type class needs to introduce a type with the right kind. I have that working in `Sugar/TypeClasses` but I'm not sure how to add it to `Sugar/Names`. If you could point me in the right direction, I'd appreciate it.

paf31:
@garyb As another example, this type checks/kind checks just fine, but fails with a `TransitiveExportError`:

```purescript
class Lifted (c :: * -> ?) (f :: * -> *) where
  lifted :: forall a. c a => Dict (c (f a))
```

So I suspect we would need a bit of a change to the way we handle type class imports/exports across the board.

texastoland:
Would this enable me to generalize these 2 files (and associated tests)?!
https://github.com/texastoland/purescript-localstorage/blob/refactor/src/DOM/WebStorage/Generic.purs
https://github.com/texastoland/purescript-localstorage/blob/refactor/src/DOM/WebStorage/JSON.purs

LiamGoodacre:
Resolves #1764

PR #2135, opened 2016-05-16
Use (..) for data constructors in suggestion

garyb:
This is something I find rather annoying about the current warnings, particularly if the type only has one constructor, or if it has many constructors. I always go though and manually edit the suggestion to use `X(..)` rather than explicit lists, and since there is no conflict problem here (adding a data constructor would be a breaking change) there's no real requirement for listing them explicitly.

### DIFF
--- a/src/Language/PureScript/Linter/Imports.hs
+++ b/src/Language/PureScript/Linter/Imports.hs
@@ -213,7 +213,15 @@ lintImportDecl env mni qualifierName names declType allowImplicit =
   checkImplicit warning =
     if null allRefs
     then unused
-    else warn (warning mni allRefs)
+    else warn (warning mni (map simplifyTypeRef allRefs))
+    where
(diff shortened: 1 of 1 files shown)


hdgarrood:
Seems like a good idea; I don't have strong feelings either way. I think the only occasion I've wanted to explicitly list a data constructor, as opposed to importing all of them, is when I do `import Node.Encoding (Encoding(UTF8))`.

garyb:
Yeah, I do the same there. This is just when it's making suggestions for open imports, so it won't suggest you change explicit lists to `(..)` or anything like that.

hdgarrood:
Cool, in that case (and having thought about it a little more) I'm definitely in favour.

natefaubion:
Yes, please. This is the only reason I don't automatically apply import suggestions.

paf31:
LGTM, thanks!

PR #2137, opened 2016-05-17
Add tests for warnings & fix some warnings

garyb:
This isn't a comprehensive test suite for the warnings, but I think it's the majority of them. The main one that I know for sure needs covering is the `ShadowedName` stuff.

I removed the `RedundantEmptyHidingImport` check, as it wasn't working, and you'd have to work really hard for that case to arise now anyway, thanks to the unused import warnings. The only way it can happen is if you import a module, `hiding` everything, and are re-exporting that module from the current module.

The fixes I speak of are things that I inadvertently broke during #2090 - had an `unless'` function that was actually `when'`, ran the import linter on the wrong version of a module, etc.

### DIFF
--- a/examples/warning/DuplicateExportRef.purs
+++ b/examples/warning/DuplicateExportRef.purs
@@ -1,0 +1,30 @@
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+module Main
+  ( X(X, X), X
+  , fn, fn
+  , (!), (!)
+  , class Y, class Y
+  , Natural, type (~>), type (~>)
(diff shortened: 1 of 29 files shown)


garyb:
Fixed the `extra-source-files` again.

examples/warning-should/ShadowedTypeVarClass.purs
```diff
@@ -0,0 +1,5 @@
+-- @shouldWarnWith ShadowedTypeVar
```
hdgarrood: Is this just in a separate directory so that it doesn't break the build?

garyb: Yeah, this is something I discovered that I figure should warn, but we currently don't catch. I made the warning detection like the errors: it expect an exact set of warnings, so yeah, it would break the tests currently.

Perhaps I should just open an issue instead and delete this.

hdgarrood: Opening an issue instead sounds sensible to me, yeah.

garyb: Opened as #2140 instead.

paf31:
LGTM, thanks!

PR #2143, opened 2016-05-19
More portable shebang line in test-everything.sh

bmjames:
Because `/bin/bash` doesn't exist on my NixOS box.


paf31:
👍 Looks good, thanks!

PR #2144, opened 2016-05-19
Fix issue 2128

### LINKED ISSUE #2128: Poor error messaging for syntax errors inside type class and instance declarations
I am using `psc` version 0.8.5.0. It seems that whenever I make a syntax error inside of a type class or instance declaration, the error message is always the same:

    Unable to parse module:
    unexpected "where"
    expecting no indentation or end of input

and the location of the `where` keyword is given, not the location of my syntax error, which may be several lines below.

Here is a minimal example:

    module Main where

    class Foo a where
      foo :: a -> !!!

Compiling with `psc` produces:

    C:\Users\Ben\purescript\hello-world>psc src\Main.purs
    Error found:
    at C:\Users\Ben\purescript\hello-world\src\Main.purs line 3, column 13 - line 3, column 13

      Unable to parse module:
      unexpected "where"
      expecting no indentation or end of input

This is not helpful for finding the real syntax error, which was on line 4.

bmjames:
This is intended to fix #2128.

That issue relates to the error messaging from the parser; the two failing cases I have added need to be compiled manually to demonstrate that the error message is accurate:

    [nix-shell:~/purescript]$ psc examples/failing/2128-instance.purs
    Error found:
    at /home/ben/purescript/examples/failing/2128-instance.purs line 8, column 9 - line 8, column 9

      Unable to parse module:
      unexpected !!!
      expecting expression

and

    [nix-shell:~/purescript]$ psc examples/failing/2128-class.purs
    Error found:
    at /home/ben/purescript/examples/failing/2128-class.purs line 5, column 15 - line 5, column 15

      Unable to parse module:
      unexpected !!!
      expecting qualifier, proper name, (, {, _, "forall", "\8704" or identifier

If you think the testing should be improved to guard automatically against a regression, please could you give me some guidance on how to approach that?

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -183,7 +183,7 @@ parseTypeClassDeclaration = do
   className <- indented *> properName
   idents <- P.many (indented *> kindedIdent)
   members <- P.option [] $ do
-    P.try $ indented *> reserved "where"
+    indented *> reserved "where"
     indented *> mark (P.many (same *> positioned parseTypeDeclaration))
   return $ TypeClassDeclaration className idents implies members
 
@@ -207,7 +207,7 @@ parseTypeInstanceDeclaration :: TokenParser Declaration
 parseTypeInstanceDeclaration = do
   instanceDecl <- parseInstanceDeclaration
   members <- P.option [] $ do
-    P.try $ indented *> reserved "where"
+    indented *> reserved "where"
     mark (P.many (same *> positioned parseValueDeclaration))
   return $ instanceDecl (ExplicitInstance members)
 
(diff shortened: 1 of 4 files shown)


src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -182,8 +182,8 @@ parseTypeClassDeclaration = do
     return implies
   className <- indented *> properName
   idents <- P.many (indented *> kindedIdent)
-  members <- P.option [] . P.try $ do
-    indented *> reserved "where"
+  members <- P.option [] $ do
+    P.try $ indented *> reserved "where"
```
paf31: Do we need this `try`? `where` should either consume everything or nothing, there should be no reason to backtrack.

paf31:
Thanks for this. Could you please update `CONTRIBUTORS.md` too?

src/Language/PureScript/Parser/Declarations.hs
bmjames: You're right, `try` is superfluous here.

paf31:
Thanks!

kRITZCREEK:
Cool!

paf31:
I just realized this is on `master`. We'll have to remember to merge and cut 0.9.1 from `master` then.

PR #2152, opened 2016-05-22
Fix #2149, remove support for = in ObjectBinder

### LINKED ISSUE #2149: Remove `=` syntax for record pattern matching?
This was brought up in IRC today. It's so long since `:` didn't work that I forgot it was even possible to use `=`. It would probably be better if we just went with one of the two.

paf31:


garyb:
:+1:

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -504,9 +504,10 @@ parseNullBinder = underscore *> return NullBinder
 
 parseIdentifierAndBinder :: TokenParser (String, Binder)
 parseIdentifierAndBinder =
-    do name <- lname <|> stringLiteral
+    do name <- lname
        b <- P.option (VarBinder (Ident name)) rest
        return (name, b)
+    <|> (,) <$> stringLiteral <*> rest
```
garyb: What's happening here now?

paf31: This is the same as before. I thought it could be changed, but I changed it back.

garyb: What's actually happening here in the first place? :smile: I don't quite understand what the `stringLiteral` is doing here.

paf31: You can have `{ foo: bar }` or `{ foo }` or `{ "!!!": baz }, but not `{ "!!!" }`, since that wouldn't be a valid `Ident`.

garyb: Ohh, right, I got confused... I'm not sure what I was thinking, but I didn't notice that `parseIdentifierAndBinder` is only for record pairs.

PR #2157, opened 2016-05-23
Combine multiple export refs for types

### LINKED ISSUE #2138: Shadowed type exports should be merged so all listed data constructors are exported
Repro case:
```purescript
module Problem (A, A(..)) where
data A = B | C
```

```purescript
import Problem (A(B,C))
```

Compiling the second file generates `Module Problem does not export data constructor B for type A`

I suggest giving an error for such shadowing or at least warning if such export might be useful in some case (which I can't think of).

garyb:
Resolves #2138.

Also fixed the error message so it refers to the type constructor and data constructor the right way around.

### DIFF
--- a/examples/passing/2138.purs
+++ b/examples/passing/2138.purs
@@ -1,0 +1,7 @@
+module Main where
+
+import Control.Monad.Eff.Console (log)
+
+import Lib (A(B,C))
+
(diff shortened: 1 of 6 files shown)


paf31:
👍 LGTM!

PR #2160, opened 2016-05-24
Return a completion for operators

### LINKED ISSUE #2159: [psc-ide] Fails to complete or give type information for operators
Crashes with "The unexpected happened" because of a failed pattern match in `Language.PureScript.Ide.Util`.

kRITZCREEK:
Because we can't get at the type/kind for now we just print the
referenced function/type constructor

fixes #2159

### DIFF
--- a/src/Language/PureScript/Ide/Externs.hs
+++ b/src/Language/PureScript/Ide/Externs.hs
@@ -103,11 +103,19 @@ convertDecl P.EDInstance{} = Nothing
 
 convertOperator :: P.ExternsFixity -> ExternDecl
 convertOperator P.ExternsFixity{..} =
-  ValueOperator efOperator (T.pack (P.showQualified (either P.runIdent P.runProperName) efAlias))
+  ValueOperator
+    efOperator
+    (T.pack (P.showQualified (either P.runIdent P.runProperName) efAlias))
+    efPrecedence
+    efAssociativity
 
 convertTypeOperator :: P.ExternsTypeFixity -> ExternDecl
 convertTypeOperator P.ExternsTypeFixity{..} =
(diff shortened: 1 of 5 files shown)


paf31:
Could we show the full fixity info in the tooltip?

kRITZCREEK:
First off: We could show the fixity if you feel that would be useful and I'll tackle that right away.

Operators in Externsfiles only contain a `Qualified something` where `Qualified` can be a totally different module. 
Right now we mostly match 1 ExternsDeclaration of the Externsfile into one ExternsDecl of the psc-ide internal format. [Here] (https://github.com/purescript/purescript/blob/957270644dca12c1090963348cd8c4fb940aa0a8/src/Language/PureScript/Ide/Externs.hs#L98)

"Mostly", because afterwards there is some cleanup that relates 1 ExternsDeclaration to all the converted Declarations of 1 Externsfile.  [Here]
(https://github.com/purescript/purescript/blob/957270644dca12c1090963348cd8c4fb940aa0a8/src/Language/PureScript/Ide/Externs.hs#L77)

In order to get at the type of an Operator however we might have to look up a different module, that potentially isn't even loaded yet. This could be solved by looking up the type at query time, but that's where we need low latency and preferably only one pass (Reexports violate this right now and it causes lots of complexity).

So what I'm planning to do (I'll open another issue and explain more), is splitting the loading into multiple phases. Where in the first phase we load ExternsFiles and Sourcefiles into memory and then start denormalizing and chopping it into a shape/ast that gives us simple to implement and fast queries.

kRITZCREEK:
type now gets returned as: `infix 2 Main.bind as >>=`

kRITZCREEK:
The CI failure is not about this PR. One of the OSX instances is failing to download GHC

paf31:
Looks great, thanks!

PR #2162, opened 2016-05-25
Fix core-tests

paf31:


hdgarrood:
👍 although is there any way we can keep the `psc-docs` test in? The re-export handling code has caused a few internal errors in the past, and while I think they're all fixed now, it would definitely make me happier to know that at least the core libraries don't cause any.

paf31:
How's this?

hdgarrood:
Looks great! 👍

PR #2163, opened 2016-05-25
Typecheck main in PSCi

paf31:
I noticed `Partial` functions would run incorrectly in PSCi since we didn't typecheck `main`. Now that we generalize constraints, we need to add a signature.


hdgarrood:
Will this throw an error unless you add in an `unsafePartial`? I think it might be nice to just evaluate expressions with Partial constraints automatically, without needing an `unsafePartial`. Not sure how difficult that is from an implementation perspective.

paf31:
It does, yes. I figured that was probably the best default behavior, since you can always use `unsafePartial` in the REPL.

paf31:
@hdgarrood So what do you think then?

hdgarrood:
My gut says we should just let people evaluate partial expressions without requiring unsafePartial, but I think this is fine for now; we can try it and see what happens. It's not a big deal.

paf31:
Alright, sounds good. We can create an issue to track it, but I'll merge this for now, so that we can get a release out.

PR #2166, opened 2016-05-27
Primed types

### LINKED ISSUE #1569: Allow prime symbols in types
I don't think that it's very hard to implement (maybe I'm wrong), but I think it can be useful sometimes. Here is an example:

```purescript
module Foo_Test where

import Test.QuickCheck
import Test.QuickCheck.Arbitrary
import Foo (Foo(..))

newtype Foo' = Foo' Foo

instance arbitraryFoo' :: Arbitrary Foo' where ...
```

Underscore could be useful, for example, for some tool that automatically finds all test modules (or something else) and runs them and prime symbol could be useful when you have some wrapper, but you don't want to create a meaningful name for it.

garyb:
Resolves #1569


paf31:
`master`?

garyb:
WTF

PR #2167, opened 2016-05-27
Allow use of prime symbol in type names.

### LINKED ISSUE #1569: Allow prime symbols in types

garyb:
Resolves #1569

### DIFF
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -41,8 +41,8 @@ module Language.PureScript.Parser.Lexer
   , commaSep1
   , lname
   , qualifier
+  , tyname
   , uname
-  , uname'
   , mname
   , reserved
   , symbol
@@ -213,8 +213,9 @@ parseToken = P.choice
   , P.try $ P.char '_'    *> P.notFollowedBy identLetter *> pure Underscore
   , HoleLit <$> P.try (P.char '?' *> P.many1 identLetter)
   , LName         <$> parseLName
-  , do uName <- parseUName
-       (guard (validModuleName uName) >> Qualifier uName <$ P.char '.') <|> pure (UName uName)
+  , parseUName >>= \uName ->
+      (guard (validModuleName uName) >> Qualifier uName <$ P.char '.')
+        <|> pure (UName uName)
   , Symbol        <$> parseSymbol
   , CharLiteral   <$> parseCharLiteral
   , StringLiteral <$> parseStringLiteral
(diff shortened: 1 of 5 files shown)


garyb:
I opened the last PR from the GitHub client, either forgot to switch the branch or something odd happened :smile:

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -213,8 +216,10 @@ parseToken = P.choice
   , P.try $ P.char '_'    *> P.notFollowedBy identLetter *> pure Underscore
   , HoleLit <$> P.try (P.char '?' *> P.many1 identLetter)
   , LName         <$> parseLName
-  , do uName <- parseUName
-       (guard (validModuleName uName) >> Qualifier uName <$ P.char '.') <|> pure (UName uName)
+  , parseUName >>= \uName ->
+      (guard (validModuleName uName) >> Qualifier uName <$ P.char '.')
+        <|> (guard (validUName uName) >> pure (UName uName))
```
paf31: Does it make sense to move these checks into `mname`, `uname`, etc.

garyb: That sounds good to me, I certainly found this a bit strange. Shall I do that then?

garyb:
I need to update the parsing of `DeclarationRef` to use `typeName` parsing too. I'll push that with any other revisions in a moment.

src/Language/PureScript/Parser/Lexer.hs
garyb: Actually, can we? The `P.char '.'` thing in there for `Qualifier` probably obstructs that. Should work for `uname` / `tyname` though.

garyb:
`DeclarationRef` handling and lexer tweak now in.

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -440,12 +444,6 @@ mname = token go P.<?> "module name"
   go (UName s) | validModuleName s = Just s
   go _ = Nothing
 
-uname' :: String -> TokenParser ()
-uname' s = token go P.<?> show s
-  where
-  go (UName s') | s == s' = Just ()
-  go _ = Nothing
```
garyb: Deleted this as it was unused, and I wasn't sure what the right implementation would be anymore: should the `validUName` predicate apply? etc.

paf31:
🚢 thanks!

src/Language/PureScript/Parser/Lexer.hs
paf31: If it's unused, just get rid, I think.

garyb:
Looks like the nightly build is broken now? Not by this I imagine, there's some message about the compiler.

paf31:
Yes, the Mac build has been broken for a few days. I've been ignoring it, but maybe we should turn that build off.

PR #2169, opened 2016-05-28
Merge 0.9 branch into master

paf31:
I think we're pretty close to making a 0.9.1 release, so we can probably merge this back now.


garyb:
Sounds good. I'm working on some of the outstanding 0.9.1 bits now.

paf31:
I'll merge this when CI finishes then. I suppose the last few bits can be PR'd onto `master`.

We should probably come up with a plan to address the remaining issues on the release checklist, but none of them should block a release, AFAICT.

I'll be working on updating my libraries and the book over the next few weekends. One of the low-level libraries which needs some non-trivial updates is `purescript-dom`, so we should probably chat about that at some point.

paf31:
@garyb I've turned off the nightly Stackage build for now.

I think this is good to merge and tag now, yes?

garyb:
Assuming it passes :+1: :smile:

garyb:
The build failure is the same thing I was getting on some PRs lately, the error is similar to that of the now-removed nightly build.

hdgarrood:
Just FYI, in case it wasn't clear, the `allow_failures: true` line in .travis.yml means that the final build status does not depend on that specific job, so it should just be the OSX build causing the failing build status there. There may be a good argument for disabling it to speed the build up, though.

paf31:
@hdgarrood Let's leave the nightly build out for now, if that's okay? We can have a stab at supporting GHC 8 after 0.9.1 is out, I think.

PR #2170, opened 2016-05-28
Fix issues with multiple data/type decls

### LINKED ISSUE #1945: Compiler allows multiple data declarations
The compiler allows duplicate data declarations within a module. Not caused me any particular problems, but it might be confusing.

```
module Main where

data X = X Int
data X = X String
```

garyb:
Resolves #1945


garyb:
Making use of the name sum to eliminate a whole bunch of error variations too :)

paf31:
Looks good to me!

PR #2174, opened 2016-05-30
fixes #2172: allow escape sequences in JS FFI exported using strings

### LINKED ISSUE #2172: Bug in FFI identifier validation
For the following FFI module:

    exports['replace\''] = function(x) { ... };

I get the following error:

> The identifier `replace\'` is not valid in PureScript.
> Note that exported identifiers in FFI modules must be valid PureScript identifiers.

To fix it, I have to write:

    exports["replace'"] = function(x) { ... };

This is silly.

michaelficarra:
Fixes #2172.

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -19,6 +19,7 @@ import Prelude.Compat
 import Control.Monad
 import Control.Monad.Error.Class
 
+import Data.Char (chr, digitToInt)
 import Data.Generics (everything, everywhere, mkQ, mkT)
 import Data.Graph
 import Data.List (nub, stripPrefix)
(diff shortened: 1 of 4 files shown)


michaelficarra:
Some of the Travis builds passed and some failed 😕. Not sure what to do.

hdgarrood:
Is examples/passing/2172.js being included in the source distribution? If not, the purescript.cabal file may need amending. It looks like the OSX builds may be failing because of a corrupted cache or a bug in the build script anyway though, which I can investigate this weekend if nobody else gets there first.

michaelficarra:
Thanks, @hdgarrood. I've amended the `purescript.cabal` file to include `examples/passing/*.js`.

src/Language/PureScript/Bundle.hs
```diff
@@ -186,11 +187,32 @@ withDeps (Module modulePath es) = Module modulePath (map expandDeps es)
...
 -- String literals include the quote chars
 fromStringLiteral :: JSExpression -> Maybe String
-fromStringLiteral (JSStringLiteral _ str) = Just $ trimStringQuotes str
+fromStringLiteral (JSStringLiteral _ str) = Just $ strValue str
 fromStringLiteral _ = Nothing
 
-trimStringQuotes :: String -> String
-trimStringQuotes str = reverse $ drop 1 $ reverse $ drop 1 $ str
+strValue :: String -> String
+strValue str = go $ reverse $ drop 1 $ reverse $ drop 1 str
```
paf31: So `language-javascript` gives the raw escaped string, straight out of the JS source?

src/Language/PureScript/Bundle.hs
```diff
@@ -186,11 +187,32 @@ withDeps (Module modulePath es) = Module modulePath (map expandDeps es)
 
 -- String literals include the quote chars
 fromStringLiteral :: JSExpression -> Maybe String
-fromStringLiteral (JSStringLiteral _ str) = Just $ trimStringQuotes str
+fromStringLiteral (JSStringLiteral _ str) = Just $ strValue str
 fromStringLiteral _ = Nothing
 
-trimStringQuotes :: String -> String
-trimStringQuotes str = reverse $ drop 1 $ reverse $ drop 1 $ str
+strValue :: String -> String
```
paf31: This isn't blocking, but we should try to avoid the two `reverse` applications here, by looking for the final character in `go`.

paf31:
Thanks for this! I've cleared out the build cache in Travis, so we'll see what happens.

src/Language/PureScript/Bundle.hs
michaelficarra: Yeah, unfortunately.

src/Language/PureScript/Bundle.hs
michaelficarra: Will do.

paf31:
CI passes with the cache cleared out.

Is this ready to merge then? If so, I'll merge this and 0.9 into master, and then we can release 0.9.1 tomorrow.

michaelficarra:
@paf31 Should be good to go once CI finishes.

paf31:
Failing due to `-Werror` and 

```
In an equation for `go': Patterns not matched: []
```

michaelficarra:
@paf31 Fixed (even though it's unreachable 😓).

paf31:
Thanks!

PR #2191, opened 2016-06-10
Elaborate re-exports

garyb:
I forgot I worked on this until @kRITZCREEK mentioned something in IRC a minute ago.

Basically, in its current state, this doesn't do anything visible. But it should enable us to remove or improve a ton of code that currently has to figure out things with re-exports and the like - so docs, publish, ide, and psci should all be able to benefit from this change I think.

Opening this to see if it looks like it will help @kRITZCREEK out now, as he's looking at fixing up some name related code in ide.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -80,19 +80,25 @@ data DeclarationRef
   --
   | ModuleRef ModuleName
   -- |
+  -- A value re-exported from another module. These will be inserted during
+  -- elaboration in name desugaring.
+  --
+  | ReExportRef ModuleName DeclarationRef
+  -- |
   -- A declaration reference with source position information
   --
   | PositionedDeclarationRef SourceSpan [Comment] DeclarationRef
   deriving (Show, Read)
 
 instance Eq DeclarationRef where
-  (TypeRef name dctors)  == (TypeRef name' dctors') = name == name' && dctors == dctors'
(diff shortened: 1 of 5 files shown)


garyb:
@kRITZCREEK IRC isn't working for me right now, I got disconnected just as I tried to answer your question!

I'm pretty sure I made it so the `ModuleName` in the `ReExportRef` refers to the module that the re-exported thing originally came from, so no having to step back recursively to figure it out.

Now I'm thinking of it though, that actually isn't quite what we want for the documentation stuff, as we only want one layer of re-exporting to be visible there...

It might be safer to construct a test case and see which behaviour we have right now :smile: I should have submitted this when I actually did it, as I would have remembered more clearly then.

kRITZCREEK:
:D I think it makes more sense to have the ModuleName only defer 1 Step at a time as then you could have a `ReExportRef ModuleName (ReExportRef ModuleName DeclarationRef)` and wouldn't lose any information. Not sure if that would break "modularity" though.

garyb:
Yeah, something like that would work! Alternatively we could just have two `ModuleName`s on the re-export - the 1st step and the original defining module, as that's all we'd ever need really.

paf31:
👍 Looks good, sorry for the slow review, I'm catching up now.

garyb:
No problem! I would have pestered more if this was more than groundwork for further changes - been pretty busy lately myself too :smile:

PR #2185, opened 2016-06-05
[psc-ide] multi phase load

kRITZCREEK:
Adresses #2164 

Remarks and todos in no particular order

- Removes the `dependencies` field from the load command, since it doesn't play nicely with reexports and the new multiphase loading

- [x] Document the changes to the `load` format

- Still uses the old psc-ide internal module format to resolve reexports. I'd like to reuse the functionality inside the compiler library but I think it'll need some refactoring for that to work out. @garyb maybe we could chat a bit about that?

- [x] Add docstrings for the new functions inside `State`

In the current state this PR doesn't add any features. Because it's already kind of a mess to review I'll add new features in follow up PRs and only use this one to lay the proper foundation.

PR #2192, opened 2016-06-11
[psc-ide] protolude

kRITZCREEK:
On top of #2185.

Switches to the Protolude (for psc-ide only), an alternative Prelude that reduces import boilerplate and "enforces" the use of Text over String.


kRITZCREEK:
If you like how this looks I try to slowly start moving more parts of the compiler... That would create lots of merge conflicts in all the other PR's though.

psc-ide-server/Main.hs
```diff
@@ -43,19 +36,19 @@ import           Network                           hiding (socketPort, accept)
...
+import           Options.Applicative               hiding ((<>))
 import           System.Directory
 import           System.FilePath
-import           System.IO
+import           System.IO                         hiding (putStrLn, print)
 import           System.IO.Error                   (isEOFError)
 
 import qualified Paths_purescript                  as Paths
 
 -- "Borrowed" from the Idris Compiler
```
paf31: Just a note: we should definitely try to give proper attribution in the license file if this is indeed taken from the Idris compiler. It might be simpler to just use something like http://hackage.haskell.org/package/network-simple-0.4.0.5/docs/Network-Simple-TCP.html though.

paf31:
Protolude changes look good. I'll get onto the other part of this PR soon.

kRITZCREEK:
Thank you!

PR #2193, opened 2016-06-11
[psc-ide] Parse Modules

kRITZCREEK:
On top of #2192 

Things left to do:

- [x] Document new command line options
- [x] Save the Filepath into the state
- [x] ~~Reparse Modules when the filewatcher triggers~~

Watching Sourcefiles is a thing for another PR


src/Language/PureScript/Ide.hs
```diff
@@ -133,71 +137,67 @@ addClause
...
+  -- Finally we kick off the worker with @async@ and return the number of
+  -- successfully parsed modules.
   env <- ask
-  let runLogger = runStdoutLoggingT . filterLogger (\_ _ -> confDebug (ideConfiguration env))
-  _ <- liftIO $ async (runLogger (runReaderT populateStage2 env))
-  pure (TextResult "All modules loaded.")
+  let runLogger =
+        runStdoutLoggingT
+        . filterLogger (\_ _ -> confDebug (ideConfiguration env))
+  _ <- liftIO (async (runLogger (runReaderT populateStage2 env)))
```
paf31: Can you please document why this result is not needed?

paf31:
👍 Looks great!

kRITZCREEK:
I've now benchmarked this on halogen using the measuring approach described [here](http://chrisdone.com/posts/measuring-duration-in-haskell).

For `purescript-halogen` that is:
`[Debug] Finished populating Stage2 in 1.54 µs`

The worker runs inside a single STM transaction so I don't think any cancellation is needed. Using this approach for measuring adds both the `clock` aswell as the `formatting` dependency. Do you think that's worth it or should I rather stick with the `time` approach for now and look at `criterion` for benchmarking?

src/Language/PureScript/Ide/State.hs
```diff
@@ -194,11 +196,11 @@ populateStage2 :: (Ide m, MonadLogger m) => m ()
 populateStage2 = do
   st <- ideStateVar <$> ask
   duration <- liftIO $ do
-    start <- getCurrentTime
+    start <- getTime Monotonic
     atomically (populateStage2STM st)
```
paf31: Doesn't this function just fork a thread? In which case, you'd be measuring the wrong thing here, I think.

paf31: Sorry, ignore me, I read this wrong. 1.5 microseconds is blazingly fast though, are you sure that's right?

paf31:
I'm not worried about any dependencies which are a) in Stackage, and b) permissively licensed. But please do rerun the license generator.

src/Language/PureScript/Ide/State.hs
kRITZCREEK: I've little to no experience in haskell benchmarking. It might totally be that lazyness is playing games with me here. I'd expect things inside STM to be evaluated after `atomically`, but that's just me guessing.

paf31: Well, you're definitely going to have to make sure whatever is getting evaluated actually is evaluated fully before calling `getTime` a second time. You could do that in `populateStage2STM` probably.

A simpler approach might be to benchmark on the command line, using something like the `bench` tool.

kRITZCREEK:
Both packages are licensed under BSD-3.

src/Language/PureScript/Ide/State.hs
kRITZCREEK: I have to go afk for a while now. I'll set something up later, thanks for all the comments!

kRITZCREEK: I'd love to use tools like bench, but the whole point of restructuring the loading process was to move computation into workers and respond to clients before all the work is actually done.

kRITZCREEK:
The remaining Travis failures are some issue with the OS X runners. I had to remove the dependency on formatting, since it didn't build on windows. Good to go from my side

paf31:
Thanks!

We can figure out the Travis issues later, when we release 0.9.2.

PR #2199, opened 2016-06-19
Evaluate PSCi expressions in the browser

### LINKED ISSUE #2142: Remote connection to PSCi (REPL)

paf31:
This is _very_ rough right now, but it works, and [here's a demo](https://www.youtube.com/watch?v=SzoI1nD6cQE).

This assumes some client JavaScript code which understands the websocket protocol being used. I put [this](https://github.com/paf31/psci-experiment) together as a quick experiment based on the Try PureScript code.

- [x] Handle the websocket close event
- [x] Limit to a single connected client?
- [x] Handle CTRL+C correctly
- [x] Add a `--browser` command line option
- [x] Serve static resources using Warp
- [x] Avoid sending compiled code over the websocket. The client can load it from the static web server.
- [x] Bundle all JS resources on startup
- [x] Re-bundle on `:reload`

Fixes #2142.

cc @artemyarulin I see you've been working on this too. We should combine efforts, if you're interested?


paf31:
@artemyarulin If you're still interested in working on this, I refactored my code to allow for the two backends (see the `nodeBackend` and `browserBackend` functions), so it should be possible to port your CLI code to build the `Backend` from the command line arguments (using either `--browser <port>` or `--node-opts`).

If you don't have time, that's totally fine obviously, but I wanted to let you know.

artemyarulin:
Thanks, really nice, I keep eye on everything related to PSCi. Unfortunately I don't have time to work on it now, but I'll try to play with it and report if anything interesting pop up.

artemyarulin:
purescript-repl-react-native
Just quickly played with remote REPL and it works with React Native just fine, I'm very happy 👍

paf31:
Alright, that's much cooler than my demo. 😄  I'm glad it worked for you, thank you for testing it out.

garyb:
That's so cool!

psci/Main.hs
```diff
@@ -92,6 +113,205 @@ getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
...
+  , "      , '(function(module) {'"
+  , "      , replaced"
+  , "      , '})(module);'"
+  , "      , 'return module.exports[\"$main\"] && module.exports[\"$main\"]();'"
+  , "      ].join('\\n');"
+  , "  new Function(wrapped)();"
+  , "  return buffer.join('\\n');"
+  , "};"
+  , "window.onload = function() {"
+  , "  var socket = new WebSocket('ws://0.0.0.0:9160');"
```
paf31: Must remember to fix this port.

psci/Main.hs
```diff
@@ -92,6 +113,205 @@ getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
...
+  , "        var result = evaluate(response);"
+  , "        socket.send(result);"
+  , "      } catch (ex) {"
+  , "        socket.send(ex.stack);"
+  , "      }"
+  , "    }, function(err) {"
+  , "      socket.send('Error sending JavaScript');"
+  , "    });"
+  , "  };"
+  , "  socket.onopen = function () {"
```
paf31: Handle `onclose` etc.

psci/Main.hs
```diff
@@ -92,6 +113,205 @@ getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
...
+  , "  };"
+  , "};"
+  ]
+
+indexPage :: IsString string => string
+indexPage = fromString . unlines $
+  [ "<!DOCTYPE html>"
+  , "<html>"
+  , "  <head>"
+  , "    <script src='js/bundle.js'></script>"
```
paf31: `<title>`

artemyarulin:
Can we put JS/HTML file into separate file? Now editing it is quiet difficult. And would be cool to have auto-reconnect logic

paf31:
Yes, I think that's a good idea, although the plan was that you wouldn't need to edit the JS/HTML, and would instead edit some foreign module in your project.

paf31:
I should be able to finish this off tomorrow, but it's pretty much ready for review now.

artemyarulin:
I'm happy to test, just ping me when it's needed. Curious though - why limit client to only one connection? Only for simplicity of the first version?

paf31:
I thought about allowing multiple connections, but the issue is that I need the browser to return its result for printing, and with multiple connected browsers, that might get confusing. We can look into it later though, it might be interesting for testing out things like webworkers.

paf31:
Ok, this is ready for review now. I've tested it quite a bit, and it works quite nicely, but any additional testing is definitely appreciated.

@artemyarulin I added support for multiple clients too, although I only show the result from the first browser to respond.

artemyarulin:
Oh, wow, awesome! I'll test it during today/tomorrow

paf31:
Could someone please review the code here, so that we can try to release it in 0.9.2?

paf31:
Clearing the Travis cache seems to have fixed the issue, but we might need to keep a separate cache (or periodically clear the existing one) on MacOS.

garyb:
I'm probably not going to have any time to review until later in the week, sorry.

paf31:
Ok, no worries, thanks.

artemyarulin:
I'm not qualified enough to check the code, but from the actual functionality - everything works just fine.

psci/Main.hs
```diff
@@ -92,6 +129,267 @@ getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
...
+bundle :: IO (Either Bundle.ErrorMessage String)
+bundle = runExceptT $ do
+  inputFiles <- liftIO (glob (".psci_modules" </> "node_modules" </> "*" </> "*.js"))
+  input <- for inputFiles $ \filename -> do
+    js <- liftIO (readUTF8File filename)
+    mid <- Bundle.guessModuleIdentifier filename
+    length js `seq` return (mid, js)
+  Bundle.bundle input [] Nothing "PSCI"
+
+-- TODO: use JMacro here?
```
hdgarrood: I haven't used JMacro, so I can't speak to its applicability here, but I have used [file-embed](http://hackage.haskell.org/package/file-embed) successfully for this sort of thing before.

psci/Main.hs
```diff
@@ -92,6 +129,267 @@ getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
...
+  , "  request.addEventListener('load', function() {"
+  , "    callback(request.responseText);"
+  , "  });"
+  , "  request.addEventListener('error', onError);"
+  , "  request.open('GET', uri);"
+  , "  request.send();"
+  , "};"
+  , "var evaluate = function evaluate(js) {"
+  , "  var buffer = [];"
+  , "  console.log = function(s) {"
```
hdgarrood: Presumably we should only need to overwrite `console.log` once, not every time `evaluate` is called?

hdgarrood: Oh, sorry, of course you need to capture the new `buffer` reference.

hdgarrood: Might it make sense to restore `console.log` to its original value after evaluating, so that if people try to do stuff inside the browser window directly, it does what they would expect?

psci/Main.hs
```diff
@@ -92,6 +129,267 @@ getCommand singleLineMode = handleInterrupt (return (Right Nothing)) $ do
...
+
+        onException :: Maybe Wai.Request -> SomeException -> IO ()
+        onException req ex
+          | Just (_ :: WS.ConnectionException) <- fromException ex
+          = return () -- ignore websocket disconnects
+          | otherwise = Warp.defaultOnException req ex
+
+        staticServer :: Wai.Application
+        staticServer req respond
+          | [] <- Wai.pathInfo req =
```
hdgarrood: Bit of a nitpick, but why not `case Wai.pathInfo req of {...}`?

hdgarrood:
Looks good! :+1: I just watched the video, this looks really exciting!

paf31:
@hdgarrood Thanks for the review, I'll polish this up some time this weekend.

paf31:
@hdgarrood How's this?

kRITZCREEK:
Code looks good to me! Would be nice if we could figure out how to test this.

hdgarrood:
Looks good. Just checking: are these new files going to be included in source distributions?

paf31:
@hdgarrood I put them under `extra-source-files`, so I think so?

hdgarrood:
Yeah, cool, thats what I was imagining. I guess I missed that change in the cabal file the first time I looked.

paf31:
So shall I merge this then? Maybe we can make a 0.9.2 release soon, with this, @kRITZCREEK's psc-ide changes, and hopefully fixes for some of the minor issues in the 0.9.2 milestone?

kRITZCREEK:
I think what we have makes for a pretty nice `0.9.2`

Issue #1186, opened 2015-06-20
Make superclass code gen less ugly


paf31:
After the operator renaming, the only really ugly thing (to my eyes) in the current generated code is the code for a superclass instance:

``` javascript
var Category = function (__superclass_Prelude$dotSemigroupoid_0, id) {
    this["__superclass_Prelude.Semigroupoid_0"] = __superclass_Prelude$dotSemigroupoid_0;
    this.id = id;
};
```

How can we tidy this up and keep the same performance?

One simple way might be to just change the naming. The indexing is enough to guarantee uniqueness of argument names, so we don't need the module name, and `__superclass_` is a bit cumbersome too. Since class names are `ProperName`s, and members are `Ident`s, their names can't overlap, so we don't even need the `__`:

``` javascript
var Category = function (Semigroupoid_0, id) {
    this.Semigroupoid_0 = Semigroupoid_0;
    this.id = id;
};
```

If the class name is unique in the set, which it tends to be in a large number of cases, we could dispense with the index too:

``` javascript
var Category = function (Semigroupoid, id) {
    this.Semigroupoid = Semigroupoid;
    this.id = id;
};
```

An alternative could be to put all superclass instances in an array, but then we lose the names:

``` javascript
var Category = function (__superclasses, id) {
    this.__superclasses = __superclasses;
    this.id = id;
};
```

garyb:
Making superclasses an array would impose a performance penalty over what we currently have, so tweaking the naming is probably the way to go.

garyb:
:+1: though, the names are hideous :smile:

We just have to make sure it's easy to reconstruct the names where needed, although #928 will probably help with this if we can fix the typechecker to support it.

paf31:
@garyb This is fixed now, right?

garyb:
Not yet, it'll be part of #928 when I return to that. The dictionary names passed around are better now though.

Issue #950, opened 2015-03-08
Eta expansion is required for higher rank types with constraints


garyb:
This might already be known:

``` purescript
data Sequence t
  = Sequence (forall m a. (Monad m) => t (m a) -> m (t a))

-- This is fine
mkSequence :: forall t. (forall m a. (Monad m) => t (m a) -> m (t a)) -> Sequence t
mkSequence f = Sequence f

-- This is not:
mkSequence' :: forall t. (forall m a. (Monad m) => t (m a) -> m (t a)) -> Sequence t
mkSequence' = Sequence
```

```
Error in module Main:
Error in value declaration mkSequence':
Error checking that type
  forall t. (forall m a. (Prelude.Monad m) => t (m a) -> m (t a)) -> Main.Sequence t
subsumes type
  (forall m a. (Prelude.Monad m) => t1498 (m a) -> m (t1498 a)) -> Main.Sequence t1498
Error checking that type
  (forall m a. (Prelude.Monad m) => u1502 (m a) -> m (u1502 a)) -> Main.Sequence u1502
subsumes type
  (forall m a. (Prelude.Monad m) => t1498 (m a) -> m (t1498 a)) -> Main.Sequence t1498
Error checking that type
  forall m a. (Prelude.Monad m) => t1498 (m a) -> m (t1498 a)
subsumes type
  forall m a. (Prelude.Monad m) => u1502 (m a) -> m (u1502 a)
Error checking that type
  forall m. (Prelude.Monad m) => t1498 (m u1503) -> m (t1498 u1503)
subsumes type
  forall m a. (Prelude.Monad m) => u1502 (m a) -> m (u1502 a)
Error checking that type
  (Prelude.Monad u1504) => t1498 (u1504 u1503) -> u1504 (t1498 u1503)
subsumes type
  forall m a. (Prelude.Monad m) => u1502 (m a) -> m (u1502 a)
Error checking that type
  (Prelude.Monad u1504) => t1498 (u1504 u1503) -> u1504 (t1498 u1503)
subsumes type
  forall m. (Prelude.Monad m) => u1502 (m a1505) -> m (u1502 a1505)
Error checking that type
  (Prelude.Monad u1504) => t1498 (u1504 u1503) -> u1504 (t1498 u1503)
subsumes type
  (Prelude.Monad m1506) => u1502 (m1506 a1505) -> m1506 (u1502 a1505)
Error at examples\passing\SequenceDesugared.purs line 12, column 1 - line 12, column 15:
[...]

PR #2205, opened 2016-06-22
Typeclass codegen

garyb:
This is a revival of #928 I worked on on the plane the other day, still needs some work - there's some kind of scope issue making names unresolvable in the generated code. Perhaps related to the "you only get what you import" change in 0.9.1.


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -153,14 +153,17 @@ isExplicit :: ImportDeclarationType -> Bool
 isExplicit (Explicit _) = True
 isExplicit _ = False
 
+type DataConstructor
```
paf31: Can this be a record please?

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -282,21 +293,30 @@ typeInstanceDictionaryDeclaration name mn deps className tys decls =
...
       -- Create values for the type instance members
-      members <- zip (map typeClassMemberName decls) <$> traverse (memberToValue memberTypes) decls
+      members <- mapM (memberToValue memberTypes) decls
+
+      -- The member names in the order defined in the class declaration
+      let memberOrder = typeClassMemberNames tyDecls
+
+      -- The members of the instance resorted to match the order in the class
+      -- declaration
+      let sortedMembers =
```
paf31: Don't we have some `inOrderOf` function somewhere for this sort of thing?

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -317,14 +337,16 @@ typeInstanceDictionaryDeclaration name mn deps className tys decls =
...
+typeClassMemberNames :: [Declaration] -> [Ident]
+typeClassMemberNames = map go
+  where
+  go (TypeDeclaration ident _) = ident
+  go (ValueDeclaration ident _ _ _) = ident
+  go (PositionedDeclaration _ _ d) = go d
+  go d = error $ "Invalid declaration in type class definition: " ++ show d
 
-superClassDictionaryNames :: [Constraint] -> [String]
+superClassDictionaryNames :: [Constraint] -> [Ident]
```
paf31: These are only used for the accessors now, yes? I'd probably just get rid, honestly.

paf31:
Looking good so far!

src/Language/PureScript/Sugar/TypeClasses.hs
garyb: Yeah, this will probably change before I mark the PR as done anyway, as I was going to address #1186 as part of it.

src/Language/PureScript/Sugar/TypeClasses.hs
garyb: Ah maybe, I'll check!

src/Language/PureScript/ModuleDependencies.hs
```diff
@@ -71,8 +71,10 @@ usedModules ams d =
     [mn]
   forDecls (FixityDeclaration fd)
     | Just mn <- extractQualFixity fd, mn `notElem` ams = [mn]
-  forDecls (TypeInstanceDeclaration _ _ (Qualified (Just mn) _) _ _)
-    | mn `notElem` ams = [mn]
+  forDecls (TypeClassDeclaration _ _ cs _)
+    = filter (`notElem` ams) (constraintModules cs)
+  forDecls (TypeInstanceDeclaration _ cs (Qualified (Just mn) _) _ _)
+    = filter (`notElem` ams) (mn : constraintModules cs)
```
garyb: @paf31 am I right in thinking we already should have been considering constraints when determining module dependencies?

This wasn't obviously broken before (only so far as I never encountered an issue with it), but maybe it was possible to construct a breaking case.

garyb: Hey wait, do we need this at all?! Now we require explicit imports, could we just rely on that instead?

kRITZCREEK: it's just a warning so I don't think we can "rely" on it

garyb: You mean extraneous imports are a warning? I don't think that's a problem - I was talking about the fact that names are unresolvable when they're imported now, so we might have some extra dependencies in the graph if we only consider imports, but the advantage would be we can delete almost all of this code here if we do just rely on imports instead.

kRITZCREEK: Oh I'm sorry. I just walked in without context... Sounds great! :balloon:

garyb:
There are still some typechecker issues with this approach. :cry:

This instance is problematic:

``` purescript
instance heytingAlgebraFunction :: HeytingAlgebra b => HeytingAlgebra (a -> b) where
  ff _ = ff
  tt _ = tt
  implies f g a = f a `implies` g a
  conj f g a = f a && g a
  disj f g a = f a || g a
  not f a = not (f a)
```

With this error:

```

    (HeytingAlgebra a2) => a2

  with type

    b1


while trying to match type (HeytingAlgebra a2) => a2
  with type b1
while checking that expression ((((...) (...)) (\f ->
                                                  ...
                                               )
                                )
                                (\f ->
                                   \g ->
                                     ...
                                )
                               )
                               (\f ->
                                  \a ->
                                    not (...)
                               )
  has type HeytingAlgebra (a0 -> b1)
in value declaration heytingAlgebraFunction

where b1 is a rigid type variable
      a0 is a rigid type variable
      a2 is a rigid type variable

See https://github.com/purescript/purescript/wiki/Error-Code-ConstrainedTypeUnified for more information,
```

Although if `tt` and `ff` are defined with lambdas, it works!

``` purescript
instance heytingAlgebraFunction :: HeytingAlgebra b => HeytingAlgebra (a -> b) where
  ff = \_ -> ff
  tt = \_ -> tt
  implies f g a = f a `implies` g a
  conj f g a = f a && g a
  disj f g a = f a || g a
  not f a = not (f a)
```

I'm slightly baffled as I thought they'd desugar to the same thing.

src/Language/PureScript/ModuleDependencies.hs
paf31: I think you're right that most of this can go away now. I'd give it a try and test on some reasonably large project to make sure.

paf31:
I'm confused too, they should definitely desugar to the same code. I'll go through this again and see if I can spot any issues which might cause that.

paf31:
This seems similar to #950. In the existing code, the check is basically on a record literal, which I would expect to be more likely to give this error. Perhaps I need to look at the check for `Constructor` again.

garyb:
I've found out what the difference between

```
ff _ = false
```
and
```
ff = \_ -> ff
```
is at least.

The first one desugars to:
```
ff = \?0 -> ff
```
(hah)

And the second one desugars to:

```
ff = \?0 ->
  case ?0 of
    _ -> ff
```

That doesn't obviously explain the change in typechecking behaviour to me, but perhaps you have some idea where I should be looking @paf31?

garyb:
Ah! The inferred type for the `ff _ = ff` version is:

```
forall t1. t1 -> (forall a. (HeytingAlgebra a) => a)
```

But for `\ff _ -> ff` it's:

```
forall t3 t4. (HeytingAlgebra t4) => t3 -> t4
```

garyb:
It seems like we should be able to fix this by applying `moveQuantifiersToFront` somewhere, but I can't find the magic place that makes it work.

garyb:
@paf31 did you ever get a chance to have a look or think about this? 

Fixing the fact that `\x -> ...` typechecking fails where `\a -> case a of x -> ...` works seems like it would potentially fix other bugs too, in regards to constrained type errors at least.

Issue #2164, opened 2016-05-25
[psc-ide] Plans to rework the loading process


kRITZCREEK:
# Reworking psc-ide's loading process/state management

I'm planning to do a major overhaul of how psc-ide loads assets into memory.

## Current situation

We currently only look at the `externs.json` files inside the specified output folder. At the heart of the loading process lies the function `loadModule`:

1. Decode ExternsFile
2. Convert each ExternsDeclaration into a `psc-ide` internal Externsdecl
3. Add the decoded ExternsFile into the server state, indexed by its ModuleName
4. Add the converted Extersdecls into the server state, indexed by their ModuleName

### Benefits of this approach

- Data we save in the internal state is always indexed by a ModuleName. 
- No data "leaks" into other modules and we can easily update stale data for a changed ExternsFile, by calling `loadModule` again.
- Reloading single modules is fast
- Reloading modules is just generally fast because aeson is amazing, and we only need to parse JSON
- We don't have to worry about the order in which we load modules, because there are no inter module dependencies inside the loading process.

### Problems that arise

Every cross-module dependency has to be resolved at query-time rather than loading time. This is currently the case for resolving Reexports, which get build up and resolved for every query from the loaded state. 

With the recent changes to Operators and TypeOperators we have a similiar situation. Because Operators only contain a reference like ´Control.Monad.bind` we cannot return type information for them unless we look up the identifier at query time.

## Planned Features

The features that I am planning to implement in the near future of `psc-ide` are:

- [x] go-to-definition
- [ ] docstrings for identifiers
- [ ] show-usage
- [ ] show instances for typeclasses

Because all of these need access to the parsed AST, `psc-ide` will start accepting globs for sourcefiles and also hold the parsed AST in memory.

- The additional parsing will make loading slower
- having to match AST nodes with declarations in the externs will need additional computation time
- 3 & 4 need cross module access

In order to enable these new features and move the handling of Reexports and Operators into load-time I propose the following changes.

### Stage 1: ExternsFiles and parsed AST's

These correspond directly to entities on disc, and need to be updated via file-watchers

### Stage 2: AST information

This data only needs to be reloaded whenever the sourcefile for a module changes on disc. Here we can still directly map a resource on disc to the loaded data and don't need to track any dependencies for save partial reloads.

- Docstrings for declarations
- Sourcespans for definitions

### Stage 3: Declarations per module

This data needs to be recomputed whenever a sourcefile or externsfile change.

- Reexports are already resolved here
[...]

Issue #1136, opened 2015-06-01
"Info" functionality for psci


tel:
Right now when interrogating names in psci there are two options, `:type` and `:kind`. GHCi also offers `:info` which is highly valuable for a few reasons:

* It can interrogate types and constructors
* It can return more information when interrogating a class method
* It returns the definition site when interrogating any name

Some examples:

```
Prelude> :i Just
data Maybe a = ... | Just a 	-- Defined in ‘Data.Maybe’
```

```
Prelude> :i Maybe
data Maybe a = Nothing | Just a 	-- Defined in ‘Data.Maybe’
instance Eq a => Eq (Maybe a) -- Defined in ‘Data.Maybe’
instance Monad Maybe -- Defined in ‘Data.Maybe’
instance Functor Maybe -- Defined in ‘Data.Maybe’
instance Ord a => Ord (Maybe a) -- Defined in ‘Data.Maybe’
instance Read a => Read (Maybe a) -- Defined in ‘GHC.Read’
instance Show a => Show (Maybe a) -- Defined in ‘GHC.Show’
```

```
Prelude> :i return
class Monad (m :: * -> *) where
  ...
  return :: a -> m a
  ...
  	-- Defined in ‘GHC.Base’
```

```
Prelude> :i foldr
foldr :: (a -> b -> b) -> b -> [a] -> b 	-- Defined in ‘GHC.Base’
```

These features are highly valuable for exploring unknown code.

paf31:
:+1:

paf31:
Dupe of #1053, but this probably one has more information.

tel:
I saw that one, but I'm not sure it's a dupe. Or rather, I'm not sure if there's a lot of replication between `ghc-mod`'s info API and `:info` in the interpreter.

paf31:
It's a duplicate of #529 then :P, but again, this card has more detail, so I'm keeping this one.

tel:
Ha, and, yep ;P

leohahn:
I am interested in looking into this feature. I am a beginner, should I just fork this repository here?

jFransham:
[...]

Issue #1053, opened 2015-04-16
IDE support for info


rikvdkleij:
Add `info` command to psci. Similar like `info` in `ghc-mod`.

`info <file> <identifier>`

This command would make it easier to get nice IDE support for Purescript. 

Most important information would be definition and module where identifier is defined. Nice to have is for example list of instances in case identifier is type class.

hdgarrood:
Is this related to #529? Can you post example output?

If this is for IDE support, I'm not sure that `psci` is the right place to put it. Maybe a separate command line tool would be better.

rikvdkleij:
@hdgarrood hdgarrood
> Is this related to #529? Can you post example output?

Yeah, it's the same. I did not see that issue.

> If this is for IDE support, I'm not sure that psci is the right place to put it. Maybe a separate command line tool would be better.

Well, `info` in both would be great.  You are right, for IDE support a separate tool (like ghc-modi) would be even nicer because output would be easier to parse (also for checking code, getting type information, browse module etc.).

rikvdkleij:
@hdgarrood 
IDE support is also related to https://github.com/purescript/purescript/issues/555

So maybe create separate issue for IDE support which includes all commands?

hdgarrood:
Thanks for clarifying :) I'm not sure it's worth making a separate issue, maybe just retitle this one?

rikvdkleij:
Retitle to "IDE support for `info`"?

paf31:
Does `psc-ide` provide this functionality?

kRITZCREEK:
Not yet :( But we could move this issue to the psc-ide repo and track it from here.

hdgarrood:
Just for the sake of clarity: is it correct that what we're aiming for here is a way of asking "tell me everything you know about this identifier (which might refer to a value, type, or type class)"?

We all seem to agree that this would be useful both inside `psci` and for IDE support. Presumably, then, we want to start off by implementing the version for `psci`, since this would need to be inside the compiler? (In this case, the original title for this issue would perhaps be more appropriate?)

paf31:
#1136

PR #2210, opened 2016-06-27
[psc-ide] Goto definition

kRITZCREEK:
Works surprisingly well in my first tests. I still need to add a few integration tests, and I think the API could use some polish.

### DIFF
--- a/src/Language/PureScript/Ide/SourceFile.hs
+++ b/src/Language/PureScript/Ide/SourceFile.hs
@@ -69,11 +69,13 @@ getImportsForFile fp = do
 -- | Given a surrounding Sourcespan and a Declaration from the PS AST, extracts
 -- definition sites inside that Declaration.
 extractSpans
-  :: P.SourceSpan -- ^ The surrounding span
-  -> P.Declaration -- ^ The declaration to extract spans from
-  -> [(Either Text Text, P.SourceSpan)] -- ^ A @Right@ corresponds to a type
-                                        -- level declaration, and a @Left@ to a
-                                        -- value level one
+  :: P.SourceSpan
+  -- ^ The surrounding span
+  -> P.Declaration
+  -- ^ The declaration to extract spans from
+  -> [(Either Text Text, P.SourceSpan)]
+  -- ^ A @Right@ corresponds to a type level declaration, and a @Left@ to a
+  -- value level one
 extractSpans ss d = case d of
   P.PositionedDeclaration ss' _ d' ->
     extractSpans ss' d'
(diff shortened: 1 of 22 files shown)


kRITZCREEK:
Allright I'm happy now :) One thing that concerns me, is that I can't make the defaults assumption of `src/**/*.purs` and `bower_components...` for the sourcefiles, since we are in the compiler. This makes starting the server a bit cumbersome.

kRITZCREEK:
Allright :) Ready for review from my side. The Windows CI is fixed and OS X are failing as always...

src/Language/PureScript/Ide/SourceFile.hs
```diff
@@ -68,31 +66,39 @@ getImportsForFile fp = do
         unwrapImportType (P.Hiding decls)   = P.Hiding (map unwrapPositionedRef decls)
         unwrapImportType P.Implicit         = P.Implicit
 
-getDeclPosition :: P.Module -> Text -> Maybe P.SourceSpan
-getDeclPosition m ident = getFirst (foldMap (match ident) decls)
+-- | Given a surrounding Sourcespan and a Declaration from the PS AST, extracts
+-- definition sites inside that Declaration.
+extractSpans
+  :: P.SourceSpan -- ^ The surrounding span
```
paf31: You might like to move the comments to the next line. The `^` is meant to point upwards to the code above, I believe.

src/Language/PureScript/Ide/SourceFile.hs
```diff
@@ -68,31 +66,39 @@ getImportsForFile fp = do
...
+                                        -- value level one
+extractSpans ss d = case d of
+  P.PositionedDeclaration ss' _ d' ->
+    extractSpans ss' d'
+  P.ValueDeclaration i _ _ _ ->
+    [(Left (runIdentT i), ss)]
+  P.TypeSynonymDeclaration name _ _ ->
+    [(Right (runProperNameT name), ss)]
+  P.TypeClassDeclaration name _ _ members ->
+    [(Right (runProperNameT name), ss)] ++ concatMap (extractSpans' ss) members
```
paf31: You can use `:` instead of `++` for some of these.

paf31:
Looks good! So, this adds position info for type queries, but no new commands for general name queries yet, correct?

kRITZCREEK:
Yep. I'm planning to turn the `type` command into something like ghci's `:info` to adress #1136 and #1053.

paf31:
Could you please create an issue for this, so we have a place for comments etc.?

paf31:
I'd like to know why Travis keeps failing on MacOS. Maybe we should just disable those builds until we can figure out what's going on.

kRITZCREEK:
I've listed my current backlog for `psc-ide` in #2164 under *Planned Features*. I'll turn that into a checklist to show the current progress.

kRITZCREEK:
Thanks for taking a look :) It would be nice if there was another way than me creating GIFs to showcase the new functionalities... I should either learn to hack the atom plugin so that you can try things out, or you could install a version of emacs, just to try out how the new stuff works.

I thought about making a little online playground with ace editor backed by a `psc-ide` instance to be able to play and test new features but I'll need to think about it a bit more, especially how that's supposed to work on the per-file level :)

kRITZCREEK:
Let me know if you want me to squash.

kRITZCREEK:
rebased and cleaned up in #2215

PR #2213, opened 2016-07-01
Update Data.Function constant for prelude 1.0

felixSchl:
This fixes the compiler not inlining mkFn, since `Data.Function` has
moved to `Data.Function.Uncurried`


garyb:
I think there needs to be a bit more updated than this - we need both `Data.Function` and `Data.Function.Uncurried` constants, since changing it like this will break the inlining for `($)` and `(#)`.

felixSchl:
@garyb you're right. I gave it another go if you don't mind checking it. I had to update the inlining code to inline `mkFnN` in the `Data.Function.Uncurried` module

garyb:
Looks good to me! :+1: 

Can you add yourself to the `CONTRIBUTORS.md` as well please, since this is your first compiler contribution? (& congrats!)

felixSchl:
Sure thing, done :)

paf31:
Thanks!

PR #2215, opened 2016-07-03
New reexports

kRITZCREEK:
after #2210.

This cleans up some of the cruft left from #2193. In particular we can finally get rid of the `ModuleOld` data type, by using the machinery introduced with  #2191 to resolve reexports.


paf31:
👍 Looks good!

kRITZCREEK:
Strangely enough the OS X tests are working for this PR. Looks like once you've got a corrupted OS X vm for a PR it gets reused?

Issue #1866, opened 2016-02-05
Include more information about why constraints are missing


hdgarrood:
Given this code:
```
> let f x y = x + x * y
Error found:
in module $PSCI
at  line 1, column 5 - line 1, column 21
  No type class instance was found for
    Prelude.Semiring _0
  The instance head contains unknown type variables. Consider adding a type annotation.
in value declaration f
where _0 is an unknown type
```

the error message does the job well enough, I think.

However, for functions with large bodies, it can be difficult to track down why the constraint is required - which is a problem if you weren't intending your function to have that constraint at all. GHC includes a little extra information for missing instance errors, something like this:

```
    No instance for (Num a1) arising from a use of ‘+’
    In the expression: x + x * y
```

Could psc do this too?

user471:
Is it just bad error position? I added it to https://github.com/purescript/purescript/issues/1801#issuecomment-180215949

garyb:
No, the idea here is to explain which value within an expression causes a constraint to be required, so if you use `get` for the `State` monad it will specifically mention the expression involving `get`, not just print an error for the whole thing.

user471:
>the idea here is to explain which value within an expression causes a constraint to be required

If there is a value that causes an error, it could be explain by error position. With correct position it would seem pretty obvious. 
![error](https://cloud.githubusercontent.com/assets/5988253/12845753/2d08e0ae-cc19-11e5-8d23-58d0f051c621.png)

hdgarrood:
There's no harm in a little redundancy here, I think, and in fact I think we should be doing everything we can to make errors easier to understand (as long as it doesn't make them too long). Also, not all errors are reported through editors with fancy tooling like that; take CI build logs, for example.

user471:
>Also, not all errors are reported through editors with fancy tooling like that; take CI build logs, for example.

Then I think there should be a common solution for all errors not just for specific one. 
Instead of error location compiler could show the code and point the error position in it, similar to Elm
http://elm-lang.org/assets/blog/error-messages/0.15.1/naming.png

So the error would be
```
Error found:
in module $PSCI

let f x y = x + x * y
              ^
  No type class instance was found for
    Prelude.Semiring _0
  The instance head contains unknown type variables. Consider adding a type annotation.
in value declaration f
where _0 is an unknown type
```

hdgarrood:
[...]

Issue #1433, opened 2015-08-28
Improve error location information inside type class instance lookup


hdgarrood:
I'm currently converting the tests in purescript-sequences to use newtype wrappers to avoid orphan instances. For example, [this file](https://github.com/hdgarrood/purescript-sequences/blob/83e81f02154a70b8c94bcb33f0336c1a9a2aeacd/test/Data/Sequence/NonEmpty.purs#L24-L147) contains some of the offending tests. In these files, all the tests are defined using one big `do` block.

In my working copy, I now have the necessary newtype wrappers, and they all have Arbitrary instances. I thought I had converted all the tests over to use these wrappers, but it seems I must have missed at least one, because I'm getting "No instance found for Arbitrary (NonEmpty.Seq Number)" errors.

Unfortunately the source span is huge: "Error at /home/harry/documents/code/purescript-sequences/test/Data/Sequence/NonEmpty.purs line 20, column 1 - line 145, column 57:" - being told the error is somewhere between line 20 and line 145 is not very helpful.

As far as I can tell, this applies to errors arising from missing instances only.

garyb:
I think #1357 would makes this possible when solved.

paf31:
Is this still happening?

hdgarrood:
Yes, using the compiler from master. See purescript-sequences at the commit 25233d98529

https://github.com/hdgarrood/purescript-sequences/blob/25233d985292a75b51d008763c16360efbdb06e8/test/Data/Sequence/Ordered.purs

The error is:

```
Error found:
Error in value declaration orderedSequenceTests:
Error at /home/harry/documents/code/purescript-sequences/test/Data/Sequence/Ordered.purs line 27, column 1 - line 116, column 70:
Error in module Tests.Data.Sequence.Ordered:
No instance found for

  Test.QuickCheck.Arbitrary.Arbitrary (OrdSeq Number)
```

paf31:
Ah ok, this isn't related to do blocks, it's just an issue with errors during type class instance lookup.

paf31:
This is actually quite difficult because of the way we propagate position information using `MonadError`. We might want to consider just passing the list of hints as a function argument.

paf31:
Even though this is related to errors, I'd like to move it to 0.9, and handle it with the new type checker changes.

PR #2230, opened 2016-07-16
Improved error messages in the constraint solver

paf31:
Behold:

```text

    Issue1310.Inject (Eff
                        ( console :: CONSOLE
                        | t0
                        )
                     )
                     (Eff
                        ( oops :: Oops
                        | eff1
                        )
                     )

while checking that expression inj (log "Oops")
  has type forall eff.
             Eff
               ( oops :: Oops
               | eff
               )
               Unit
while applying a function inj
  of type forall a g f. (Inject f g) => f a -> g a
  to argument log "Oops"
in value declaration main

where eff1 is a rigid type variable
        bound at line 18, column 1 - line 18, column 23
      t0 is an unknown type
```

This probably needs rebasing, I wrote it on the plane.


paf31:
Perhaps we should give up on trying to separate the various AST modules, and just create one big `Types` module. Here, it was pretty much unavoidable, and I had to merge the errors types into the AST module. Merging in `Binders` would let us fix the `String` hack I came up with for the `Partial` error messages too.

paf31:
Fixes #1866 and #1433

garyb:
Merging in binders would mean we need to merge `CoreFn` (and probably `CoreImp` when it's done) in with the input AST also, which I'm not so sure would be great? I don't know.

This is great though :+1: :tada:

PR #2240, opened 2016-07-21
[WIP][psc-ide] Force computations in worker threads

kRITZCREEK:
Still figuring this out... Is there a more effective way for profiling this other than inserting trace statements thoughout the code and checking what gets evaluated?

It seems to already have an impact though, since the worker threads now take more time to finish on the codebase for pscid.

Before:
```
[Debug] Finished populating Stage2 in TimeSpec {sec = 0, nsec = 1181} 
[Debug] Finished populating Stage3 in TimeSpec {sec = 0, nsec = 1222} 
```

After:
```
[Debug] Finished populating Stage2 in TimeSpec {sec = 0, nsec = 569836} 
[Debug] Finished populating Stage3 in TimeSpec {sec = 0, nsec = 793256} 
```

@nwolverson You noticed major slowdowns while working on atom-ide-purescript, does this help?

### DIFF
--- a/src/Language/PureScript/Ide/State.hs
+++ b/src/Language/PureScript/Ide/State.hs
@@ -16,7 +16,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PackageImports    #-}
 {-# LANGUAGE TemplateHaskell   #-}
-{-# LANGUAGE BangPatterns      #-}
 
 module Language.PureScript.Ide.State
   ( getLoadedModulenames
--- a/src/Language/PureScript/Ide/State.hs
+++ b/src/Language/PureScript/Ide/State.hs
@@ -12,9 +12,11 @@
 -- Functions to access psc-ide's state
 -----------------------------------------------------------------------------
 
-{-# LANGUAGE OverloadedStrings     #-}
-{-# LANGUAGE PackageImports        #-}
-{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE BangPatterns      #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PackageImports    #-}
+{-# LANGUAGE TemplateHaskell   #-}
+{-# LANGUAGE BangPatterns      #-}
 
 module Language.PureScript.Ide.State
   ( getLoadedModulenames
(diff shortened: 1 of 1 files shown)


nwolverson:
Will take a look. I changed things up so that just time-to-first-completion from that initial 1-time load matters (still slower than without passing globs but I feel acceptable).

nwolverson:
Right now I see no difference in terms of time it takes `load` to reply - about 6-7 seconds for me right now on `atom-ide-purescript` itself. There is a difference for sure on the times logged:

```
[Debug] Finished populating Stage2 in TimeSpec {sec = 0, nsec = 130639} @(pures_EYGw6hyUrvWGJ6gCMSHJNm:Language.PureScript.Ide.State src/Language/PureScript/Ide/State.hs:178:5)
[Debug] Finished populating Stage3 in TimeSpec {sec = 0, nsec = 7493} @(pures_EYGw6hyUrvWGJ6gCMSHJNm:Language.PureScript.Ide.State src/Language/PureScript/Ide/State.hs:199:5)
```

```
[Debug] Finished populating Stage2 in TimeSpec {sec = 0, nsec = 3649102} @(pures_KTc4sWdhKzA3pjGUCII4oA:Language.PureScript.Ide.State src/Language/PureScript/Ide/State.hs:180:5)
[Debug] Finished populating Stage3 in TimeSpec {sec = 0, nsec = 1811317} @(pures_KTc4sWdhKzA3pjGUCII4oA:Language.PureScript.Ide.State src/Language/PureScript/Ide/State.hs:202:5)
```

kRITZCREEK:
Ahh, so you're actually seeing the delay when calling the `load` command and not on the first command *after* load.

I think that this is caused by parsing all the source files synchronously. The idea here was to be able to give a sensible response to the editors. We could make parsing the source and externs files asynchronous as well, but then we'll definitely need a better system to report errors or warnings from these asynchronous processes.

nwolverson:
So it's not ideal but I don't think making things more complicated is worthwhile if this is a 1-time cost when there are plenty more productive things to be doing...

src/Language/PureScript/Ide/State.hs
```diff
@@ -181,7 +182,8 @@ populateStage2 = do
 populateStage2STM :: TVar IdeState -> STM ()
 populateStage2STM ref = do
   modules <- s1Modules <$> getStage1STM ref
-  let spans = map (\((P.Module ss _ _ decls _), _) -> M.fromList (concatMap (extractSpans ss) decls)) modules
+  let !spans = map (\((P.Module ss _ _ decls _), _) ->
```
paf31: What is the type of `spans` here? Forcing it to WHNF probably isn't doing much extra work if it's a list.

paf31:
@kRITZCREEK What's the status of this one? Do you want to try to include it in 0.9.3, or leave it for later?

kRITZCREEK:
I want to leave it for later. I still need to figure out what to force and how I can properly benchmark this.

kRITZCREEK:
I'll close this PR. I need to come back to this at some point but this is just needlessly occupying the PR list.

PR #2247, opened 2016-07-24
Better context information for typed hole errors

### LINKED ISSUE #2165: Show most relevant names in typed hole errors
The context which is shown right now is sometimes helpful, but other times useless, especially when lots of things are in scope. We should compare types to pick the most relevant things to show.

paf31:
Fixes #2165 by only showing "local" things in context, not names of global values/functions. We can add something using the subsumption check later, hopefully.

Also a bit of minor tidying.


paf31:
Could someone please review this (@kritzcreek maybe, if you have time?), and I'll try to include it in a 0.9.3 release along with the other error message fix.

Thanks.

paf31:
This now includes context info in type wildcard warnings too.

garyb:
Looks good to me, some nice simplifications in `TypeChecker/Types` too now the `ModuleName` lookup can be skipped in a bunch of cases :+1:

src/Language/PureScript/Interactive.hs
```diff
@@ -234,7 +234,7 @@ handleTypeOf val = do
   case e of
     Left errs -> printErrors errs
     Right (_, env') ->
-      case M.lookup (P.ModuleName [P.ProperName "$PSCI"], P.Ident "it") (P.names env') of
+      case M.lookup (P.mkQualified (P.Ident "it") (P.ModuleName [P.ProperName "$PSCI"])) (P.names env') of
```
kRITZCREEK: Just something I noticed: Probably a good candidate to extract into a function? Something like `lookupLastResult`?

paf31: Yes, good point. I'd like to defer this to another clean-up PR though.

PR #2253, opened 2016-07-28
Decode externs with correct encoding

natefaubion:
Externs failed to decode properly when they contained unicode, causing
files to be rebuilt too often.

Reference:
http://stackoverflow.com/questions/27669418/aeson-does-not-decode-strings-with-unicode-characters


paf31:
Looks good, thanks!

PR #2258, opened 2016-07-30
Fix #2252, instantiate types in array literals before unification

### LINKED ISSUE #2252: Impredicative type inferred when using polymorphic values but not with constructors
``` purescript
data T a = T

-- inferred as `Array t0`
xs :: _
xs = [T, T, T]

t :: forall a. T a
t = T

-- inferred as `Array (forall a. T a)`
xs' :: _
xs' = [t, t, t]
```

paf31:
cc @garyb 

We just need to make sure we instantiate types before unifying. There are some more places like this, and I'm going to try to fix them in another PR.

### DIFF
--- a/examples/passing/2252.purs
+++ b/examples/passing/2252.purs
@@ -1,18 +1,15 @@
 module Main where
-
+	
 import Control.Monad.Eff.Console (log)
 
 data T a = T
-
--- inferred as `Array t0`
-xs :: _
-xs = [T, T, T]
+	 
+ti :: T Int
+ti = T
 
 t :: forall a. T a
 t = T
 
--- inferred as `Array (forall a. T a)`
-xs' :: _
-xs' = [t, t, t]
+xs = [ti, t, t]
 
 main = log "Done"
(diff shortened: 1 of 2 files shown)


paf31:
@garyb Mind having a look at this one, and then I'll release 0.9.3 today? Thanks.

garyb:
Oh right, yes - I was going to propose a different test, one that matches how I discovered the bug in the first place. I'm about to eat, but will do so when I get back.

paf31:
Cool, thanks.

garyb:
This is what I had in mind:

``` purescript
module Main where

data T a = T

ti :: T Int
ti = T

t :: forall a. T a
t = T

xs = [ti, t, t]

main = log "Done"
```

Now you get a 

```
  Could not match type

    a0

  with type

    Int
```

But `[ti, T, T]` would work, and so should the fix I assume :smile:

paf31:
Much better, thanks. Updated.

garyb:
:+1: :shipit:

paf31:
Thanks!

PR #2260, opened 2016-07-31
Do not solve for unknowns in rows during instance solving

### LINKED ISSUE #2236: Compiler infers wrong type
I have created sample repo that showcases this problem. Basically I can't figure out what is right type annotation for this function https://github.com/akegalj/pux-types/blob/master/src/Main.purs#L23 :
```purescript
data Action
    = Button
    | Nop

type State = Int

-- NOTE: here I tried to do some action with effect and I picked `throw`
-- update :: forall eff. Action -> State -> EffModel State Action (err' :: CHANNEL | eff)
update Button state = onlyEffects state [ liftEff $ throw "bla" *> pure Nop ]
update Nop state = noEffects state

view :: State -> Html Action
view count =
  div
    []
    [ button [ onClick (const Button) ] [ text "Button" ]
    ]

main :: forall e. Eff (err :: EXCEPTION, channel :: CHANNEL | e) Unit
main = do
  app <- start
    { initialState: 0
    , update: update
    , view: view
    , inputs: []
    }

renderToDOM "#app" app.html
```

when I compile without type annotation compiler infers this type:
```
      forall t10 t27.                               
        Action                                      
        -> t27                                      
           -> { state :: t27                        
              , effects :: Array                    
                             (Aff                   
                                ( channel :: CHANNEL
                                | t10               
                                )                   
                                Action              
                             )                      
              }         
```
and when I put this suggested type I am getting:
```
                                                            
    Control.Monad.Eff.Class.MonadEff ( err :: EXCEPTION     
                                     | t0                   
                                     )                      
                                     (Aff                   
                                        ( channel :: CHANNEL
                                        | eff1              
                                        )                   
                                     )    
```
So for the start it looks that it infers the wrong type. As it looks to me that this suggestion is valid because we use `throw` to make an effect (which is using `EXCEPTION` effects) - so if replaced with `CHANNEL` with `EXCEPTION` I am getting:
```
in module Main
at /home/akegalj/projects/pux-types/src/Main.purs line 24, column 1 - line 24, column 89

    Control.Monad.Eff.Class.MonadEff ( err :: EXCEPTION     
                                     | t0                   
                                     )                      
                                     (Aff                   
                                        ( channel :: CHANNEL
                                        , err' :: EXCEPTION 
                                        | eff1              
                                        )                   
                                     )     
```
There is a whole discussion about this issue starting at http://ircbrowse.net/browse/purescript?id=367336&timestamp=1468883448#t1468883448 and with the conclusion that it is probably a bug. It looks like it is related to the problem described here as well https://github.com/afcondon/puxws/commit/e45e864ad21c51e16c8f3ee92f4ba6f80053abdf .

Maybe someone knows a workaround for this? In this example a workaround is just to drop type annotation for `update` and compiler infers it himself but I have one project where this trick doesn't work so I have to annotate it properly.

paf31:
Fixes #2236

This updates the constraint solver to make rows work like regular types. For some reason, rows were left out when things were fixed to disallow unknowns in possible solutions.

#2236 demonstrates how this can lead to differences between checking and inference mode when unknowns are allowed.

The downside is that this breaks quite a lot of stuff, especially `MonadEff`. Let's discuss the pros and cons of merging this into 0.9.3.

### DIFF
--- a/examples/passing/RowPolyInstanceContext.purs
+++ b/examples/passing/RowPolyInstanceContext.purs
@@ -9,13 +9,13 @@ class T s m where
 data S s a = S (s -> { new :: s, ret :: a })
 
 instance st :: T s (S s) where
-  state f = S $ \s -> { new: f s, ret: unit }
+  state f = S \s -> { new: f s, ret: unit }
 
-test1 :: forall r . S { foo :: String | r } Unit
-test1 = state $ \o -> o { foo = o.foo <> "!" }
+test1 :: forall r. S { foo :: String | r } Unit
+test1 = state \o -> o { foo = o.foo <> "!" } :: { foo :: String | r }
 
-test2 :: forall m r . (T { foo :: String | r } m) => m Unit
-test2 = state $ \o -> o { foo = o.foo <> "!" }
+test2 :: forall m r. T { foo :: String | r } m => m Unit
+test2 = state \o -> o { foo = o.foo <> "!" } :: { foo :: String | r }
 
 main = do
   let t1 = test1
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -202,8 +202,7 @@ typeHeadsAreEqual m r1@RCons{} r2@RCons{} =
   where
   go :: [(String, Type)] -> Type -> [(String, Type)] -> Type -> Maybe [(String, Type)]
   go [] REmpty            [] REmpty            = Just []
-  go [] (TUnknown _)      _  _                 = Just []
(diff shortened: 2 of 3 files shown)


garyb:
I agree this is the right fix, by the way, just it's going to make life painful for a while.

One fairly big con though: I thought we were trying to avoid breaking things between "major" releases? It may well be a fix, but it's going to break a lot more than it fixes in this case.

paf31:
I'm happy to leave this until 0.10 if that's easier.

garyb:
Any chance we'll have fundeps then too? :stuck_out_tongue_winking_eye:

paf31:
Yeah, let's aim to do this at the same time as fundeps then.

garyb:
I think leaving till 0.10 is probably a good idea regardless though, as it's more semver-y.

PR #2261, opened 2016-08-05
gitignore .psc-ide-port globally

rvion:


hdgarrood:
Can you explain what this file does and what tools use it please?

kRITZCREEK:
The atom plugin and pscid use it keep book about a started psc-ide-server and its port.

hdgarrood:
Ah, cool, thanks.

rightfold:
I think this should not be in the repository, as it is specific to a tool some people use, instead of something generated by the build tools. You don't put `*.swp` in every gitignore file either.

Rather, it should be in the global ignores file, which you can set with `git config --global core.excludesfile`.

texastoland:
The difference is every IDE ostensibly uses that file (VS Code too) versus being editor specific (Eclipse in the case of your example I believe). Depending on global configuration generally is a bad idea.

rvion:
@rightfold Since most tools will use psc-ide-server, I still think it makes sense for the file to be included in some gitignore. (either this one or the global root .gitignore). this being said, I have no strong opinion about that and understand your concern

rvion:
from my experience, I like to include sensible defaults in gitignores even when they're tool related (eg. https://github.com/github/gitignore/blob/master/Haskell.gitignore)

paluh:
But there is also local file:  .git/info/exclude which everyone can use, to ignore ide/editor/debugger specific files.

rvion:
@paluh I mean, even when they are tool-specific-to-your-language-that-mostly-everyone-will-use related

as @texastoland said
> The difference is every IDE (VS Code too) ostensibly uses that file (see bodil/pulp#223) versus being editor specific (Eclipse in the case of your example I think). Depending on global configuration is typically a bad idea.

I really don't care much, feel free to dismiss this PR, I won't commit any .psc-ide-port anyway, and debating about this is not very productive ^^

paluh:
@rvion Ok, maybe it's just my preference, to put more specific ignores outside project's .gitignore and maybe this puritanism is not really practical ;-) I've just wanted to point out that there is local per repository place for them. Nothing more ;-)

paf31:
I agree that it probably is a better solution to put this in some local config, but I don't see any harm in merging this either.

PR #2263, opened 2016-08-05
export desugarDecl from Sugar.ObjectWildcards

rvion:
Could we make desugarDecl visible ?

Without desugarDecl available, it’s much harder to write
desugaring-related tests without building whole modules.

context: 

As I was discovering the codebase, I wanted to write this

```purescript
let decl = ValueDeclaration (Ident "test") Public [NullBinder] (Right (Accessor "bar" (Accessor "ok" AnonymousArgument)))
runMake defaultOptions $ evalSupplyT 1 $ desugarDecl decl
```

to see this:

```purescript
(Right (ValueDeclaration (Ident "test") Public [NullBinder] (Right (Accessor "bar" (Abs (Left (GenIdent Nothing 1)) (Accessor "ok" (Var (Qualified Nothing (GenIdent Nothing 1)))))))),MultipleErrors {runMultipleErrors = []})
```


src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -21,15 +22,14 @@ desugarObjectConstructors
...
   -> m Module
 desugarObjectConstructors (Module ss coms mn ds exts) = Module ss coms mn <$> mapM desugarDecl ds <*> pure exts
-  where
 
-  desugarDecl :: Declaration -> m Declaration
-  desugarDecl (PositionedDeclaration pos com d) = rethrowWithPosition pos $ PositionedDeclaration pos com <$> desugarDecl d
-  desugarDecl other = f other
-    where
-    (f, _, _) = everywhereOnValuesTopDownM return desugarExpr return
+desugarDecl :: (MonadSupply m, MonadError MultipleErrors m) => Declaration -> m Declaration
```
paf31: Can we call this `desugarObjectConstructorsInDeclaration` or something like that?

paf31: Actually, this code works with more than just objects now, so maybe these functions (and the module) should all be renamed to just `...Wildcards`.

paf31:
👍 I'm fine with exporting this, thanks.

garyb:
We could probably add a traversal for declarations-in-module, as I imagine there's a few desugaring passes that have something much like `desugarModule` here.

rvion:
I forced push to clean the diff and fix the build with ghc8, but travis seems to have timedout in one of the targets
image

rvion:
closing as it is part of https://github.com/purescript/purescript/pull/2264
I'll open an other PR later to address the module renaming suggested by paf31 (in the - _now outdated_ - diff)

Issue #1807, opened 2016-01-09
Nested record update


user471:
Is it possible to support this?

```
foo = {x: {y: 1}}
bar = foo {x.y = 2}
```

would be equal to 

```
bar = foo {x = foo.x {y = 2}}
```

hdgarrood:
I like the look of this. :)

paf31:
Neat :)

LiamGoodacre:
:+1: nice.  For reference: [Idris has this too](http://docs.idris-lang.org/en/latest/tutorial/typesfuns.html#nested-record-update)

kRITZCREEK:
One thing that is a little sad is that this discourages the use of "newtype" because it can't benefit from all the sugar.

paf31:
I like this idea, but on the other hand, lenses handle this more compositionally and more generally.

user471:
> lenses handle this more compositionally and more generally.

But, you need an additional hundreds lines of code for them. So, in some cases it's easier to use update syntax

user471:
Also, I think it should work for 
```
x = _.foo.bar
```
Right now it doesn't really make much sense for trying to get field `bar` from function `_.foo`

rvion:
> bar = foo {x.y = 2}

yes please !!

rvion:
big +1 for me. 👍 

I understand lenses bring some more power, but they also brings dependencies, and complexity.
If the language change is small, the following 3 features would be great to have IMO 
(2 first are @user471, last one is a small proposal addition)

```purescript
x = {a:{b:{c:2}}

get = _.a.b.c
set = _{a.b.c=2}
upd = _{a.b.c->(_+2)}
```

paf31:
Since we have anonymous accessor chains (#2264), I think it would be nice to have nested updates in 1.0 too.

rvion:
> Since we have anonymous accessor chains (#2264), I think it would be nice to have nested updates in 1.0 too.

@paf31 and nested types annotations ? 

``` purescript
f :: { b.c.d :: Int | stuff } -> Int 
f = _.b.c.d
```

@paf31 Also, regarding the update syntax mentioned in this [comment above](https://github.com/purescript/purescript/issues/1807#issuecomment-232920461), I now really think it it is needed to deal properly with union types;

```
x = {a: {b: c: {Left "ok"}}}

y :: {a.b.c :: Either String Int | x} -> {a.b.c :: Int | x}
y = x {
  a.b.c -> case _ of 
[...]

PR #2264, opened 2016-08-06
Allow anonymous accessor chains (_.a.b)

rvion:
this PR extend purescript desugarer so we can write

```purescript
a = _.b.c.d     -- a = \obj -> obj.b.c.d
```

this idea was discussed in #1807 and validated from what I understand

--------------

⚠️  With this implementation, since I check recursively children of accessors during desugaring TD traversal until I reach either an AnonymousArgument or something neither Accessor nor Anonymous, in cases like

`Accessor "foo" (Accessor "bar" (Accessor "baz" (Var ...))))`

will yield (n-1) * (n-2) / 2 checks during the complete traversal so it is O(n^2) (with a very small constant, and very small large cases)

🍏  this being said, it should be ok in this case because there is few a.b.c.d.e.f.g.h.(...) constructions in regular code, and because the constant is very small

------------------

I also have a prototype of nested record update, but before spending too much time on it, 
 @paf31 @garyb  can you tell me how stupid this idea is:

should we consider the (__b.c.d__) part in (a.__b.c.d__)  as a single AST node ?
thous would have several benefit

 - no overhead to support `(_.a.b.d)`
 - record update will benefit the code  ( `obj {a.b.d = 2}` )
 - `_ {a.b.d = 2}` would have type `forall a. { b.c.d :: Int |a} -> { b.c.d :: Int |a}` ☀️  

if this AST node is eliminated during desugaring, 
 - no need to extend rules to prevent `obj {a.b= x, a.c= y}`
 - no need to add extra code to typechecker, inference, etc. of the compiler.

if it's not suggar only, it would also mean 

 - less nodes, so faster compiler, codegen, etc.
 - less recursions needed to infer types, constraints, etc

but if would also requires

 - to rewrite some stuff
 - to deal with a little bit more complex model (or not ?)
 - to write some code to merge and flatten those types in various places, etc.


rvion:
(this PR is ready for review)

paf31:
Nice idea! Looks good to me, thanks, and I think it's the right approach to treat this as syntactic sugar.

rvion:
I'll create an other branch from scratch to experiment with adding this new AST node, so this PR can be merged in the meantime if you want.

------------ 

👥 If someone wants to give it a try first before I do, please, tell me. I'll be happy to let someone more familiar with the compiler do it.

-----------

Here are how I intend to do it:

 1. I plan to change to introduce a new AST node : `AttributeChain [String]` (name to be discussed)
 2. and to update
      - `Accessor String Expr` to `Accessor AttributeChain Expr`
      - `ObjectUpdate Expr [(String, Expr)]` to `ObjectUpdate Expr [(AttributeChain, Expr)]`
      - The place where row types ast nodes are (__ can somebody points it to me along a few notes related to record ?__)

3. then to update the desugarrer 
      - for values
      - and for type annotations

src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -45,9 +45,9 @@ desugarObjectConstructors (Module ss coms mn ds exts) = Module ss coms mn <$> ma
     obj <- freshIdent'
     Abs (Left obj) <$> wrapLambda (ObjectUpdate (argToExpr obj)) ps
   desugarExpr (ObjectUpdate obj ps) = wrapLambda (ObjectUpdate obj) ps
-  desugarExpr (Accessor prop u) | isAnonymousArgument u = do
+  desugarExpr (Accessor prop u) | isAnonymousAccessorChain u = do
```
paf31: Instead of doing this in two passes, why not collect the property names on the way down?

```haskell
peelAccessorChain :: Expr -> Maybe [String]
```

at which point you can use a fold in place of `buildAccessor`, and pattern guard here:

```haskell
desugarExpr (Accessor prop u)
  | Just props <- peelAccessorChain u = ...
```

rvion: indeed, good idea :)

rvion:
@paf31 do you have any feedback about my above message ?

paf31:
I think this approach is fine. I wouldn't introduce new AST nodes, that will introduce the risk of complicating various analyses I think. Obviously feel free to experiment with the idea, but I think this looks good to merge as is, aside from my one comment about avoiding two passes over the AST.

src/Language/PureScript/Sugar/ObjectWildcards.hs
paf31: @rvion I'm going through PRs for the 0.10 release at the moment. Do you think you'll be able to finish this off for 0.10 or would you prefer to leave it until 0.10.1?

rvion: I'll try to do it today, thanks for the reminder

rvion:
@paf31 done

paf31:
Thanks very much!

Issue #1896, opened 2016-02-23
Improve error messages when rows do not match


user471:
```
foo :: {foo :: Int} -> Int
foo x = x.x
```

```
 Could not match type

    ()

  with type

    ( x :: Int
    | _1
    )


while checking that type { foo :: Int
                         }
  is at least as general as type { x :: Int
                                 | _0
                                 }
while checking that expression x
  has type { x :: Int
           | _0
           }
while checking type of property accessor x."x"
in value declaration foo

where _0 is an unknown type
      _1 is an unknown type
```

I think it should be

```
Could not match type

    {foo :: Int}

 with type

    ( x :: Int
    | _1
    )

```

sengaya:
The output changed already with #2271. Is this issue still relevant and if so what is the expected error message?

paf31:
I'd like to diff the fields in two rows which don't match, and show the differences in a more readable way.

paf31:
The error is now:

```
  Type of expression lacks required label "x".

while checking that type { "foo" :: Int
                         }
  is at least as general as type { "x" :: Int
                                 | t0
                                 }
while checking that expression x
  has type { "x" :: Int
           | t0
           }
while checking type of property accessor x."x"
in value declaration foo

where t0 is an unknown type
```

but there are probably other cases where this is still an issue.

paf31:
The new row unification function makes these errors much better now, even when the better errors like the one above don't fire. Closing.

PR #2271, opened 2016-08-08
Improved error messages for record subsumption

FrigoEU:
Trying to get to know the typechecker better so I looked for the easiest bug I could find: #1896 . It's my first PR to the compiler, so if I missed anything please let me know.

One question I do have is about tests: Since the bug depends on sorting of keys in the records, I had to go through a few iterations to get it right and to wrap my head around the full problem. I have/had testcases that were failing for all those iterations. I'm not sure if I should be adding these. Feels a bit stupid to add 5+ files with always a tiny variation of a case.


src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -68,18 +68,18 @@ subsumes' val (TypeApp f1 r1) (TypeApp f2 r2) | f1 == tyRecord && f2 == tyRecord
   go ((p1, ty1) : ts1) ((p2, ty2) : ts2) r1' r2'
     | p1 == p2 = do _ <- subsumes Nothing ty1 ty2
                     go ts1 ts2 r1' r2'
+    | p1 /= p2 && r1' == REmpty && p2 `notElem` map fst ts1 =
```
paf31: We should be able to use sortedness here to avoid the `O(n^2)` complexity introduced by calling `notElem` at each element.

FrigoEU:
Ok, second try :). I pulled it out of the iteration and just check missing properties once, at the start. It should be much more efficient. 

I took the minusBy from Data.List.Ordered. I wasn't sure if I should just copy it, or if I should add it to the dependency, but for such a small function I didn't want to bother.

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -72,16 +79,32 @@ subsumes' val (TypeApp f1 r1) (TypeApp f2 r2) | f1 == tyRecord && f2 == tyRecord
...
-                       _ -> unifyTypes r1' (RCons p2 ty2 rest)
+                     unifyTypes r1' (RCons p2 ty2 rest)
                      go ((p1, ty1) : ts1) ts2 rest r2'
+
 subsumes' val ty1 ty2@(TypeApp obj _) | obj == tyRecord = subsumes val ty2 ty1
 subsumes' val ty1 ty2 = do
   unifyTypes ty1 ty2
   return val
+
+firstMissingProp :: forall a b. Ord a => [(a, b)] -> [(a, b)] -> Maybe (a, b)
```
paf31: Could you move these two functions into the `where` clause of `subsumes'` please?

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -60,6 +63,10 @@ subsumes' val (TypeApp f1 r1) (TypeApp f2 r2) | f1 == tyRecord && f2 == tyRecord
     (ts2, r2') = rowToList r2
     ts1' = sortBy (comparing fst) ts1
     ts2' = sortBy (comparing fst) ts2
+  when (r1' == REmpty) 
+    (maybe (return ()) (throwError . errorMessage . PropertyIsMissing . fst) (firstMissingProp ts2' ts1'))
```
paf31: Nitpick: `maybe (return ()) f = for_ f`

paf31:
Could you please paste an example of the error now?

As for `Data.List.Ordered`, let's pull it from the library actually. It has minimal dependencies, and that way, we can add it to the license file automatically.

FrigoEU:
I rebased this onto the current master. Added your suggestions.

PureScript code: 

````
foo :: {foo :: Int} -> Int
foo x = x.x
````

Error message before:
````
Could not match type

    ()

  with type

    ( x :: Int
    | t1
    )
````

Error message after:
````
Type of expression lacks required label x.
````

Because it's dependant on ordering, this code does give the correct error on the current master:
````
foo :: {foo :: Int} -> Int
foo x = x.a

Type of expression lacks required label a.
````

File: purescript.cabal [hunk 4]
```diff
@@ -109,6 +109,7 @@ library
                    bytestring -any,
                    containers -any,
                    clock -any,
+                   data-ordlist >= 0.4.7.0,
```
paf31: Please could you rerun the `license-generator` script?

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -60,6 +65,10 @@ subsumes' val (TypeApp f1 r1) (TypeApp f2 r2) | f1 == tyRecord && f2 == tyRecord
     (ts2, r2') = rowToList r2
     ts1' = sortBy (comparing fst) ts1
     ts2' = sortBy (comparing fst) ts2
+  when (r1' == REmpty)
```
paf31: It took me a little while to see why this works, but I think I understand now. Could you please add a comment along the lines of:

```
-- For { ts1 | r1 } to subsume { ts2 | r2 } when r1 is empty, every property in ts2 must appear in ts1. 
-- If not, then the candidate expression is missing a required property.
-- Conversely, when r2 is empty, even property in ts1 must appear in ts2, or else the expression has
-- an additional property which is not allowed. 
```

Thanks!

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -72,16 +81,14 @@ subsumes' val (TypeApp f1 r1) (TypeApp f2 r2) | f1 == tyRecord && f2 == tyRecord
...
-                     _ -> unifyTypes r2' (RCons p1 ty1 rest)
+                   unifyTypes r2' (RCons p1 ty1 rest)
                    go ts1 ((p2, ty2) : ts2) r1' rest
     | otherwise = do rest <- freshType
-                     case r1' of
-                       REmpty -> throwError . errorMessage $ PropertyIsMissing p2
-                       _ -> unifyTypes r1' (RCons p2 ty2 rest)
+                     unifyTypes r1' (RCons p2 ty2 rest)
                      go ((p1, ty1) : ts1) ts2 rest r2'
+  firstMissingProp t2 t1 = fst <$> (uncons $ minusBy' (comparing fst) t2 t1)
```
paf31: Could you please add a comment here too? Sorry to be so picky, but I'm trying to improve documentation in the typechecker as I go these days.

FrigoEU:
No problem at all. For my next PR's, I'll try to already provide some more comments.

I regenerated the License file and added about 800 lines so not sure what happened there. If I look through it though, I did found the new ones, either as direct dependencies (eg: wai) or transitive (eg: blaze) so I suppose this is how it's supposed to be. I just ran the command as mentioned in generate-license.hs:

````
stack list-dependencies | cut -f 1 -d ' ' | stack exec runhaskell license-generator/generate.hs > LICENSE
````

paf31:
Thanks very much!

PR #2274, opened 2016-08-11
[psc-ide] Return JSON errors for cycles in module dependencies

kRITZCREEK:
We want to return the proper JSON formatted error in the case of a cycle in the rebuild dependencies.

fixes kRITZCREEK/pscid#20


paf31:
Thanks!

Issue #876, opened 2015-02-17
--dump-imperative-core


paf31:
This would write the imperative core to a JSON file, just like `--dump-corefn` dumps the functional core.

garyb:
Might be useful for editor support also, as the annotations could help with completions, etc.

paf31:
Moving to 0.9, I think this will be really useful when we start focussing properly on alternative backends.

paf31:
Duplicate of #711

metaleap:
#3094

PR #2275, opened 2016-08-14
Add --dump-corefn command line option

rightfold:
WORK IN PROGRESS. I'd like some feedback on these changes.

A few things are still left to do:

 - [x] Convert `Ann` to JSON.
 - [x] Convert `Type` to JSON.
 - [x] Add version number to JSON dump.
 - [ ] Tests probably (super tedious).
 - [ ] Documentation.

JSON is dumped to `corefn.json` in the module's output directory (i.e. alongside `index.js`).

Addresses #876.


garyb:
Would it be easier to use the macros to generate Aeson instances? I don't know if that's bad or generates unhelpful JSON or whatever as I only have peripheral contact with it, but just wondered if it would make things simpler.

rightfold:
Do you mean the Template Haskell facilities? The current JSON output isn't particularly _nice_ either, but it's very systematic too, which means it's predictable if you're familiar with the ADTs.

garyb:
Yeah exactly. We already use it in a few places, like `Types`: https://github.com/purescript/purescript/blob/9b34d87d4428c2f76613ae128ec9363334cbdf63/src/Language/PureScript/Types.hs#L102-L104 and probably anything else that ends up in `externs.json`.

rightfold:
That's probably easier then, yes xD

rightfold:
Alright, did that. Much better.

hdgarrood:
I'm just a little apprehensive about using derived instances here, unless the expectation is that people will only be consuming this JSON in other Haskell programs which are using the compiler as a library, i.e. not using the json itself as an interface. For example, with derived instances, is it likely that we might want to change the internal ADT and not be able to do so without breaking third party programs?

garyb:
That's a good point, I hadn't really considered that aspect of it.

rightfold:
That's a good point. At least I gained more experience with Aeson 😺

rightfold:
Alright, reversed that, I'll now work on `Type` and `Ann`.

rightfold:
All the conversions are now implemented.

CONTRIBUTORS.md
```diff
@@ -57,6 +57,7 @@ This file lists the contributors to the PureScript compiler project, and the ter
...
 - [@pseudonom](https://github.com/pseudonom) (Eric Easley) My existing contributions and all future contributions until further notice are Copyright Eric Easley, and are licensed to the owners and users of the PureScript compiler project under the terms of the [MIT license](http://opensource.org/licenses/MIT).
 - [@puffnfresh](https://github.com/puffnfresh) (Brian McKenna) All contributions I made during June 2015 were during employment at [SlamData, Inc.](#companies) who owns the copyright. I assign copyright of all my personal contributions before June 2015 to the owners of the PureScript compiler.
+- [@rightfold](https://github.com/rightfold) (rightfold) My existing contributions and all future contributions until further notice are Copyright rightfold, and are licensed to the owners and users of the PureScript compiler project under the terms of the [BSD 3-Clause license](https://opensource.org/licenses/BSD-3-Clause).
```
paf31: I think we need to figure out how this will work, since I don't think BSD3 code can be redistributed under MIT, due to the additional clause about no advertising. I'd be fine to switch to BSD3 personally, which would be fine since MIT code can definitely be redistributed under BSD3. But we should discuss it.

rightfold: I'm fine with MIT if BSD causes too many issues

rightfold:
@paf31 I switched to MIT. The relicensing of the old commits isn't an issue because I plan to squash everything before merging anyway (commit history is a bit of a mess)

src/Language/PureScript/Make.hs
```diff
@@ -369,6 +372,14 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
       for_ (mn `M.lookup` foreigns) (readTextFile >=> writeTextFile foreignFile)
       writeTextFile externsFile exts
     lift $ when sourceMaps $ genSourceMap dir mapFile (length prefix) mappings
+    dumpCoreFn <- lift $ asks optionsDumpCoreFn
+    when dumpCoreFn $ do
+      let coreFnFile = outputDir </> filePath </> "corefn.json"
+      let jsonPayload = CFJ.moduleToJSON CFJ.annToJSON m
+      let json = Aeson.object [ ("version", Aeson.toJSON $ showVersion Paths.version)
+                              , ("payload", jsonPayload)
```
paf31: Is this json inside json now?

rightfold: `Aeson.toJSON` returns `Value`, not `String`

paf31:
@rightfold Could you please paste an example of the generated JSON here for something small like `map (_ * 2) [1, 2, 3]`? We should get this merged for the next release.

psc/Main.hs
```diff
@@ -143,6 +143,11 @@ sourceMaps = switch $
      long "source-maps"
   <> help "Generate source maps"
 
+dumpCoreFn :: Parser Bool
+dumpCoreFn = switch $
+     long "dump-corefn"
+  <> help "Dump CoreFn"
```
paf31: Please replace "Dump  CoreFn" here with something like "Dump the (functional) core representation of the compiled code at output/*/corefn.json"

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -0,0 +1,180 @@
+module Language.PureScript.CoreFn.ToJSON where
```
paf31: Please add a short module comment here. Also, consider making the exports explicit (`moduleToJSON` and `annToJSON`, I think?)

src/Language/PureScript/Make.hs
paf31: I see, thanks.

rightfold:
@paf31 http://lpaste.net/2664468466282528768

paf31:
@rightfold Here are my thoughts:

- Module names could be an array of strings: `[ "Prelude", "Control.Monad.Eff" ]` and we probably don't need to track individual value imports right now
- I would leave out source spans for now
- I would also omit type information until we find we need it
- Many types only have a single constructor (e.g. `ProperName`), so we don't need to represent those as arrays.
- I'm not sure if we need to represent numeric literals as `Either Int Double` in the generated JSON.
- Object literals and updates should probably be represented as JSON objects.

As I mentioned on IRC, it'll be easy to add features if we need them, but hard to remove them. I'd prefer that the JSON be compact and semi-readable for now. Otherwise, files are going to get _really_ big really quickly. This short module already generated 2188 lines of JSON, for example.

Thanks!

rightfold:
@paf31 JSON doesn't distinguish between floats and ints so you need the discriminator. `Int` and `Number` is probably more descriptive than `Left` and `Right` though.

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -0,0 +1,189 @@
...
+import Data.Text (pack)
+
+import Language.PureScript.AST.Literals
+import Language.PureScript.Comments
+import Language.PureScript.CoreFn
+import Language.PureScript.Names
+import Language.PureScript.Types
+
+literalToJSON :: (a -> Value) -> Literal a -> Value
+literalToJSON _ (NumericLiteral (Left n)) = toJSON ("NumericLiteral", "Int", n)
```
paf31: These could just be called `IntLiteral` and `NumberLiteral`. The fact that they're represented using the same constructor in the compiler is just a historical detail, really.

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -0,0 +1,189 @@
...
+literalToJSON _ (NumericLiteral (Right n)) = toJSON ("NumericLiteral", "Number", n)
+literalToJSON _ (StringLiteral s) = toJSON ("StringLiteral", s)
+literalToJSON _ (CharLiteral c) = toJSON ("CharLiteral", c)
+literalToJSON _ (BooleanLiteral b) = toJSON ("BooleanLiteral", b)
+literalToJSON t (ArrayLiteral xs) = toJSON ("ArrayLiteral", map t xs)
+literalToJSON t (ObjectLiteral xs) = toJSON ("ObjectLiteral", map (fmap t) xs)
```
paf31: Can this use a JSON object rather than nested arrays?

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -0,0 +1,189 @@
...
+
+annToJSON :: Ann -> Value
```
paf31: I would leave all of this out for the first version.

paf31:
@garyb @andyarvanitis Would you like to review the output format here at all, before this gets merged?

rightfold:
Hmm. It looks like the distinction between top-level let and letrec is gone?

paf31:
Yes, but we could add it back. I'd probably like to change the syntax a bit though.

paf31:
It'd be nice if non-recursive things appeared like they do now, and we added a `Rec` constructor to wrap recursive bindings IMO.

rightfold:
Or:

```json
"decls": {
  "let": [{…}, {…}],
  "letrec": [
    [{…}, {…}],
    [{…}, {…}]
  ]
}
```

paf31:
Do we need to distinguish self-recursive declarations from non-recursive ones? If not, I'd rather do something like:

```
"decls": [
  [ { "foo": ["Abs", "f", ... ] } ]
  [ { "bar": ["Abs", "a", ["Var", "baz"]] }
  , { "baz": ["Abs", "b", ["Var", "bar"]] }
  ]
}
```

where the top level is a list of binding groups, each of which is a list of function declarations.

andyarvanitis:
@paf31 Sorry, I was on an overseas flight (with no wifi). 

I'm still jet lagged right now, but it looks good. I also just looked at the sample posted by @rightfold earlier which included type information -- which is good, since the C++ backend is back to using type information (for optimizations now).

One question, though: are all imported values available? The C++ backend also uses those for some optimations (basically grabbing things from `Environment` currently).

garyb:
> Do we need to distinguish self-recursive declarations from non-recursive ones? If not, I'd rather do something like:

That looks good to me. It works for data binding groups too.

I don't think it should be the responsibility of the AST format to indicate when a function is self-recursive, since that's not something we even track in the AST, so in doing that we're kinda making the JSON representation a different thing. It should be easy enough to figure out by anything that does need that distinction too.

paf31:
> since that's not something we even track in the AST

I thought we did track self-recursive functions in the core AST, as a `Rec` constructor with a singleton list of bindings.

rightfold:
Core has Let/LetRec distinction for both top-level declarations and let-expressions

garyb:
> I thought we did track self-recursive functions in the core AST, as a Rec constructor with a singleton list of bindings.

Oh right. I should probably know that since I contributed `CoreFn`. :laughing:

paf31:
I think the question should be: do we need this for the first version, bearing in mind that it might be tricky to implement it in a non-breaking way if we decide to use it later? For that reason, I think we should add it, but I'm happy to spend some time figuring out the syntax.

paf31:
@andyarvanitis Actually, there's no type annotations in this version. The reason is that I would like to move towards something a bit more like GHC core, where only binders get type annotations. With that approach, the type checker would elaborate directly to functional core, rather than decorating the initial AST.

> One question, though: are all imported values available?

The core JSON will include a list of imports, but you would have to pull the definitions of the imports from the appropriate core JSON/externs file. Does that work?

paf31:
@garyb @andyarvanitis Binding group information has now been added back to the JSON output.

andyarvanitis:
Ok, thanks, I've gone back through the comments and see now what's there for the first version. I probably won't switch the C++ backend over to it just yet, then, but it's looks good for some other things I've been thinking about doing.

PR #2279, opened 2016-08-17
Functional Dependencies

paf31:
Given

```purescript
class Inject f g

instance injectRefl :: Inject f f
```

The current behavior is that choosing `injectRefl` requires the two type arguments to be equal on the nose. With the new approach, it is as if we defined the instance as

```purescript
instance injectRefl :: f ~ g => Inject f g
```

That is, any `f` and `g` match, and then the subgoal forces them to unify.

Of course, this is a potential problem for things like `purescript-inject`, depending on the order in which the instances get defined, but this solves some other problems without having to implement full functional dependencies.

For example, this now typechecks, where it wouldn't have before:

```purescript
test = inject [1, 2, 3]
```

and forces `test` to have type `Array Int` (better examples would be things like `MonadEff`, but I haven't tested those yet).


paf31:
This is now ready for review. Here are a few things which now typecheck which wouldn't have before:

```purescript
-- Checked, not inferred
test1 :: forall a. Show a => Reader a String
test1 = show <$> ask

-- Type inferred as MonadReader Int m => m Int
test2 = add 1 <$> ask
```

In `test1`, the `MonadReader r (Reader r)` instance is picked, causing `r` to unify with the `a` from the type signature. This allows us to solve the constraint given by `Show`.

`test2` now works because I also relaxed the rules for generalization of constraints in this PR: we now generalize if either of the following are true:

- We are solving a nullary type class constraint (special case, to support `Partial`, but I'm happy to take this out)
- _Any_ of the types in the unsolved instance head are unknowns or skolems. _Note_: this used to be "all", not "any". In `test2`, `Int` is not generalizable, but `m` is, so we're good. Note, we have to have at least one thing be generalizable, otherwise we could try to generalize over fully applied things like `Semiring String`.

Now, this isn't full functional dependencies, but it's a good first step. Notably, this still won't check (since we can't choose the `MonadReader` instance for the usual reason):

```purescript
test1 :: forall a. (MonadReader a m, Show a) => m String
test1 = show <$> ask
```

This should, however, solve some pain points for users right now. We can build functional dependencies on top of this, and I have a plan for how to implement them if we decide we need more inference.

paf31:
This also seems to fix #2236.

rvion:
juste tested it, no problem per see, and it alleviates some pain I had 👍 

(small typo in your last example, you did not existentially quantified `m`: `forall a.` should be `forall a m.`)

> Now, this isn't full functional dependencies, but it's a good first step. Notably, this still won't check (since we can't choose the MonadReader instance for the usual reason):

@paf31 shouldn't what you've done + rank-n types be enough to make it work, though ?
For instance, those work well too for me on your branch:

```purescript
test3 :: forall a m. (MonadReader a m, Show a) => m String
test3 = (show :: a -> String) <$> ask

test4 :: forall a m. (MonadReader a m, Show a) => m String
test4 = show <$> (ask :: m a)
```

Isn't there just some unification missing in the very last step ?

paf31:
@rvion Thanks for testing

> Isn't there just some unification missing in the very last step ?

There is more to it - we need to add some syntax at the class level to indicate dependencies between type arguments in classes. The issue with `test3` is that there isn't enough information in the instance we're given right now - we can't commit to choosing that instance just because the `m` matches. Compare that with `instance MonadReader r (Reader r)`, where we can use the fact that the variable `r` appears twice.

So the solution I have planned is as follows:

- Add functional dependency info to class definitions
- When choosing an instance, use functional dependencies to decide whether or not we have enough type matches to commit (i.e. every type either matches the instance head, or is implied by some functional dependency)
- After choosing an instance, force functional dependencies to hold using unification.

Once this is done, we might revisit how type arguments which appear multiple times in instance heads affect the choice of instance.

Note, I've also added a simple test case for something which would have failed to compile before, but which works now - a class for streams which is parameterized over both the stream type and the element type.

kRITZCREEK:
I think that this is an improvement and a valuable step towards functional dependencies, but we should make the caveats clear:

```purescript
module Main where

import Unsafe.Coerce (unsafeCoerce)

class MyClass f g where
  member :: f -> g -> String

instance myClassA :: MyClass Int String where
  member = unsafeCoerce

instance myClassB :: MyClass f f where
  member = unsafeCoerce

main = member 0 "Hello"
```

Compiles just fine, but if we change `myClassA` to `myClassC`...

```purescript
instance myClassC :: MyClass Int String where
```

...we fail with *Could not match type Int with type String*. This happens because the two instances are now considered overlapping and their ordering decides to which instance we commit. So in a sense this PR gives semantic meaning to the instance names, which they didn't have before. Unless this is temporary and functional dependencies solve this fully, I think it would be helpful if we could diagnose these cases and help out in the error message. Maybe we could add the available dictionaries to the error thrown when unifying class constraints and backtrack during error handling, so that we can alert the user about his overlapping instances?

paf31:
Thanks for clarifying this @kRITZCREEK.

> So in a sense this PR gives semantic meaning to the instance names, which they didn't have before.

They have the same meaning as before, which is that instances are tried in module-then-name order (athough I prefer to think of the behavior in the overlapping case as undefined behavior, hence the warning). What's changed is the matching criteria, and subsequently the definition of overlapping. Note, this is only an issue for instances which were potentially overlapping anyway.

paf31:
This is now ready for a full review.

@LiamGoodacre Thanks for your work on testing this, your code came in very useful. For reference, the compiler can now not only check the type of example 4, but also infer it 😄  Would you mind if I added your code to the test suite?

Everyone, please do continue to try to break this.

@kRITZCREEK I've taken out the strange treatment of multiple type variables, and we're now doing the same thing as before, where we try to unify before committing to an instance.

paf31:
Actually, there is still one minor issue here, related to how we decide whether to commit to an instance. We should not always allow a unification variable to unify in `unifiesWith`. That depends on whether or not the unification variable appeared in a type argument which was solved by a functional dependency.

I'll look into this tomorrow, but I'm fairly confident this can be fixed quickly.

LiamGoodacre:
Please do use it as a test case 😃 
I'll try to come up with some more complicated scenarios over the weekend.

paf31:
Okay, _now_ this is done. I've added your tests @LiamGoodacre, thank you.

paf31:
With the [new Generics example](https://github.com/purescript/purescript/blob/e1ecc2b822a41c892b533efe558530ee845fd012/examples/passing/GHCGenerics.purs), I'm now confident this works (with the possible issue of non-termination as noted by @LiamGoodacre), so if someone could please review the code, I'll try to merge this in for the next release. If someone would like to dig into the details of the checking algorithm, that would be great, but probably not necessary. A review of the style and general approach will probably be enough, since we'll probably end up ironing out a few bugs in the algorithm after this gets released anyway.

garyb:
I started reviewing this a bit today and should be done soon / maybe tomorrow.

Once we have this in, maybe we should call the next release 0.10? This is a pretty huge feature, and will let us fix some issues or improve things in several of the core libraries, along with us having a couple of other misc breaking changes throughout the core libraries that could do with releasing too.

Regarding that too, since we have fundeps now, how would you feel about a `mono-traversable` style `foldable-traversable`? (Well, maybe not like the current one which is all type familied up, but I mean in the sense of having the classes over kind `*` types with a fundep directing the "member" type, so we could have a `Foldable String Char` kinda thing for a `Foldable f a | f -> a`. Maybe there are some problems with that that I'm not aware of? But it would be nice to not have both "mono" and "non mono" versions of the fold/traverse operations.

paf31:
Thanks for reviewing! 👍 to both 0.10 and a `monotraversable` type thing in core.

Maybe we should consider what else we want to include in 0.10 then?

jdegoes:
This is sort of a minor pet peeve, but don't we generally use `=>` for implication instead of `->`? (I know Haskell FD uses `->`, but they also get the implication arrow wrong in some cases.)

garyb:
I think I have found a bug unfortunately. It's bug-ish at least:

``` purescript
class Newtype n o | n -> o where
  wrap :: o -> n
  unwrap :: n -> o

ala
  :: forall f n o n' o'
   . (Functor f, Newtype n o, Newtype n' o')
  => (o -> n)
  -> ((o' -> n') -> f n)
  -> f o
ala _ f = map unwrap (f wrap)

test = ala Multiplicative foldMap [1,2,3,4]
```

The inferred type for `test` is `forall t9. (Newtype (Multiplicative t9) t9) => Int`, and it appears that constraint cannot be resolved - you can't type `test :: Int` explicitly, or pretty much do anything with it really.

Maybe it's not a bug? But I don't really understand what's happening in that case, since things have progressed enough that we know the value should be `Int`. I see in `Control.Lens.Wrapped` in Haskell there is some stuff with a `Rewrapping` class to help with inference so I'll try applying that here and see if it's related.

garyb:
^ Actually I can't do that, it needs `~`... but I'm pretty sure this is a bug, as if I drop the array from `test` we get an inferred type that does not compile when applied:

``` purescript
test = ala Multiplicative foldMap
```

Inferred type:
```
    forall t10 t15 t16.
      ( Newtype (Multiplicative t15) t15
      , Foldable t16
      , Semiring t15
      ) => t16 t10 -> t10
```

And when applied:
```
  
    Main.Newtype (Multiplicative t100)
                 t100
  
  while checking that expression (ala Multiplicative) foldMap
    has type forall t10 t15 t16.
               ( Newtype (Multiplicative t15) t15
               , Foldable t16
               , Semiring t15
               ) => t16 t10 -> t10
  while applying a function ala
    of type forall f n o n' o'.
              ( Functor f
              , Newtype n o
              , Newtype n' o'
              ) => (o -> n) -> ((o' -> n') -> f n) -> f o
    to argument Multiplicative
  in value declaration test
  
  where t100 is a rigid type variable
          bound at line 32, column 1 - line 32, column 27
```

garyb:
Hahahaha. Ok, I'm an idiot. The above is because I didn't actually make an `Newtype Multiplicative` instance...

garyb:
Although there is still a problem with the above:

``` purescript
test = ala Multiplicative foldMap [1,2,3,4]
```

Results in:

```
                                    
    Main.Newtype (Multiplicative t0)
                 Int                
                                    
```

Would we need something like

``` purescript
instance newtypeMultiplicative :: (a ~ b) => Newtype (Multiplicative a) b where
```

For that to work? I would perhaps naively assumed that `Newtype (Multiplicative a) a` would have been good enough in the presence of the fundep.

garyb:
Code for what I was working on is here: https://github.com/garyb/purescript-newtype

It'd be nice to have something like that in core so we can provide instances for the `newtype`s provided in libraries like `Monoid`, etc. Also maybe I could add `Newtype` instance deriving if it seems like this is a worthwhile class?

paf31:
Yes, I think this is a good example of where we need the `~` constraints. I can try to implement them after this is merged, if we think it'll be important.

Note though, that if you had a dependency in the other direction, this should be solvable. Of course, that's not how newtypes work, so we shouldn't add that dependency in practice.

Thanks for testing this out, that library was on my list too 😄

garyb:
I'm pretty much :+1: on this - it's awesome! :smile: :fireworks:

I do think type equality constraints would be super useful to have, as I think the monotraversable things of kind `* -> *` might suffer from a similar issue as the `Multiplicative` example above (I'll try it out now to see actually).

I just have comments about two other things:

1. Can we do something to make the user-error case of mine above throw an "instance not found error" rather than inferring a type like `forall t9. (Newtype (Multiplicative t9) t9) => Int`?
2. There's an overlapping instances warning for `PSCI.Support.Eval a0` now - is that fixable somehow? And why is there an overlap now, where it wasn't before? I thought the thing that would have caused that was reverted again (although I must admit I wasn't 100% paying attention to the discussion at the time).

paf31:
1. Yes, I changed the rules for when an instance is generalizable. We should decide how flexible we want the generalization to be. I'm fine with making it quite strict, but that might confuse some users.
2. There was a bug in the overlap check, where I was only checking if _all_ instance pairs overlapped, not _any_ pair. So, that instance pair should always have been overlapping, but we never reported it.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -80,139 +120,278 @@ entails shouldGeneralize moduleName context = solve
...
+                  return (s, t)
+
+            unique :: [(a, TypeClassDictionaryInScope)] -> m (EntailsResult a)
+            unique [] | deferErrors = return Deferred
+                      -- We need a special case for nullary type classes, since we want
+                      -- to generalize over Partial constraints.
+                      | shouldGeneralize && (null tys' || any canBeGeneralized tys') = return (Unsolved con')
+                      | otherwise = throwError . errorMessage $ NoInstanceFound con'
+            unique [(a, dict)] = return $ Solved a dict
+            unique tcds | pairwiseAny overlapping (map snd tcds) = do
```
paf31: @garyb Here is the change to fix the overlap check.

garyb:
Re: 1. I guess I was thinking of a rule like "if the type variables involved in the constraint don't appear in the type, throw an error" - do we have enough info to do that? I guess it would have to be pushed quite late in the inference.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -80,139 +120,278 @@ entails shouldGeneralize moduleName context = solve
...
+
+                withFreshType s = do
+                  t <- freshType
+                  return (s, t)
+
+            unique :: [(a, TypeClassDictionaryInScope)] -> m (EntailsResult a)
+            unique [] | deferErrors = return Deferred
+                      -- We need a special case for nullary type classes, since we want
+                      -- to generalize over Partial constraints.
+                      | shouldGeneralize && (null tys' || any canBeGeneralized tys') = return (Unsolved con')
```
paf31: @garyb Here is the generalization check. Note the change to use `any` instead of `all`.

paf31:
Oh actually, we _do_ have a check for that. I'll look into why it's not firing.

mckeankylej:
I am very excited for this. I have this whole module that doesn't complie with a comment on the top of it saying:
```haskell
Please gods bring us fun deps
```
Thanks so much for all your work!

src/Language/PureScript/TypeChecker/Entailment.hs
paf31: The downside of using `all` is that we'd like to have `put 1` type check with the inferred type `MonadState Int m => m Unit`, but not all of `Int` and `m` are generalizable.

The downside of using `any` is that we need a special case for nullary type classes if we want to generalize over `Partial` things (maybe we don't...?) and also that it leads to some complex inferred types in some cases.

paf31:
@garyb It seems you've found a subtle bug related to when we can generalize constraints. I'm going to look into it some more and try to merge sometime over the weekend.

Thanks!

paf31:
@garyb That issue is fixed now. It turned out that reusing solutions to generalized constraints was causing problems with the new deferred constraints mechanism.

I've added a reduced version of your example as a test as well.

Please continue to try to break this... if nothing else comes up, I think we should aim to release it this weekend.

rgrempel:
I tried this in an empty directory, while using psc from the `fundeps` branch.

```
pulp init
pulp run
```

Here is the output:

```
Warning found:
in module PSCI.Support
at /Users/rgrempel/projects/purescript/fundeps/bower_components/purescript-psci-support/src/PSCI/Support.purs line 23, column 1 - line 26, column 10

    PSCI.Support.Eval a0

    $dictEval2 (chosen)
    PSCI.Support.evalShow

  Overlapping type class instances can lead to different behavior based on the order of module imports, and for that reason are not recommended.
  They may be disallowed completely in a future version of the compiler.

in value declaration evalEff

See https://github.com/purescript/purescript/wiki/Error-Code-OverlappingInstances for more information,


* Build successful.
Hello sailor!
```

So, it works, but there is an overlapping instance warning that didn't appear in 0.9.3.

Then, I tried this:

```
bower install purescript-test-unit
```

And I edited `Test.Main` to look like this:

```
module Test.Main where

import Control.Monad.Aff.AVar (AVAR)
import Control.Monad.Aff.Console (log)
import Control.Monad.Eff.Console (CONSOLE)
import Prelude (Unit)
import Test.Unit.Console (TESTOUTPUT)
import Test.Unit.Main (runTest)
import Test.Unit (test)

main :: Eff
    ( console :: CONSOLE
    , testOutput :: TESTOUTPUT
    , avar :: AVAR
    ) Unit

main =
    runTest do
        test "blah" do
            log "blah"
```

Under 0.9.3, `pulp test` then works as expected.

With the `fundeps` branch, I get this:

```
* Building project in /Users/rgrempel/projects/purescript/fundeps
[1/2 OverlappingInstances] bower_components/purescript-psci-support/src/PSCI/Support.purs:23:1

      v
  23  instance evalEff :: Eval a => Eval (Eff eff a) where
  24    eval x = do
  25      a <- unsafeInterleaveEff x
  26      eval a
             ^

    PSCI.Support.Eval a0

    $dictEval2 (chosen)
    PSCI.Support.evalShow

  Overlapping type class instances can lead to different behavior based on the order of module imports, and for that reason are not recommended.
  They may be disallowed completely in a future version of the compiler.

  in value declaration evalEff

[2/2 OverlappingInstances] bower_components/purescript-transformers/src/Control/Monad/State/Trans.purs:66:1

      v
  66  instance altStateT :: (Monad m, Alt m) => Alt (StateT s m) where
  67    alt (StateT x) (StateT y) = StateT \s -> x s <|> y s
                                                           ^

    Data.Functor.Functor m0

    $dictAlt40 (chosen)
    $dictMonad39

  Overlapping type class instances can lead to different behavior based on the order of module imports, and for that reason are not recommended.
  They may be disallowed completely in a future version of the compiler.

  in value declaration altStateT

  where m0 is a rigid type variable

[1/3 NoInstanceFound] bower_components/purescript-aff/src/Control/Monad/Aff/Console.purs:36:1

  36  warnShow = liftEff <<< C.warnShow
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    Control.Monad.Eff.Class.MonadEff ( console :: CONSOLE
                                     | t0
                                     )
                                     (Aff
                                        ( console :: CONSOLE
                                        | eff1
                                        )
                                     )

  while checking that expression (compose liftEff) warnShow
    has type forall a eff.
               (Show a) => a
                           -> Aff
                                ( console :: CONSOLE
                                | eff
                                )
                                Unit
  while applying a function compose
    of type forall b c d a. (Semigroupoid a) => a c d -> a b c -> a b d
    to argument liftEff
  in value declaration warnShow

  where eff1 is a rigid type variable
          bound at line 36, column 1 - line 36, column 34
        t0 is an unknown type

[2/3 NoInstanceFound] bower_components/purescript-quickcheck/src/Test/QuickCheck/Gen.purs:191:1

       v
  191  perturbGen n gen = do
  192    modify \s -> s { newSeed = lcgPerturb (toNumber (float32ToInt32 n)) s.newSeed }
                                                                                       ^

    Control.Monad.State.Class.MonadState { newSeed :: Seed
                                         | t0
                                         }
                                         (StateT
                                            { newSeed :: Seed
                                            , size :: Int
                                            }
                                            Identity
                                         )

  while checking that expression \n ->
                                   \gen ->
                                     (bind (...)) (\__unused ->
                                                     ...
                                                  )
    has type forall a.
               Number
               -> StateT
                    { newSeed :: Seed
                    , size :: Int
                    }
                    Identity
                    a
                  -> StateT
                       { newSeed :: Seed
                       , size :: Int
                       }
                       Identity
                       a
  while applying a function bind
    of type forall a b m. (Bind m) => m a -> (a -> m b) -> m b
    to argument modify (\s ->
                          s { newSeed = (...) (...)
                            }
                       )
  in value declaration perturbGen

  where t0 is an unknown type

[3/3 NoInstanceFound] bower_components/purescript-test-unit/src/Test/Unit/Output/Fancy.purs:27:1

      v
  27  printLive t = walkSuite runSuiteItem t
  28    where
  29      runSuiteItem path (TestGroup (Group label content) rest) = do
      ...
  53            print " because "
  54            printFail $ message err
  55            print "\n"
                         ^

    Control.Monad.Eff.Class.MonadEff ( testOutput :: TESTOUTPUT
                                     | t0
                                     )
                                     (Aff
                                        ( avar :: AVAR
                                        , testOutput :: TESTOUTPUT
                                        | e1
                                        )
                                     )

  while checking that expression \t ->
                                   let
                                     runSuiteItem = \path ->
                                                      ...
                                   in (walkSuite runSuiteItem) t
    has type forall e.
               Free
                 (TestF
                    ( testOutput :: TESTOUTPUT
                    , avar :: AVAR
                    | e
                    )
                 )
                 Unit
               -> Aff
                    ( testOutput :: TESTOUTPUT
                    , avar :: AVAR
                    | e
                    )
                    (List
                       (Tuple (List String)
                          (Aff
                             ( testOutput :: TESTOUTPUT
                             , avar :: AVAR
                             | e
                             )
                             Unit
                          )
                       )
                    )
  while applying a function liftEff
    of type forall a m eff. (MonadEff eff m) => Eff eff a -> m a
    to argument (bind ((apply print) (indent' path))) (\__unused ->
                                                         (bind (...)) (\__unused ->
                                                                         ...
                                                                      )
                                                      )
  in value declaration printLive

  where e1 is a rigid type variable
          bound at line 27, column 1 - line 57, column 1
        t0 is an unknown type
```

So, just thought I'd share, to see if others can reproduce ... of course, there may be something in my build process that isn't right.

garyb:
There are a couple of things there - the psci support error I mentioned above and Phil knows about.

The `MonadEff` inference failing is because it should never have actually worked in the first place without fundeps - see #2260 for the original PR relating to that issue. It applies to classes involving rows generally, so probably the same goes for the `MonadState` error here.

The overlapping `Functor` I'm not so sure about, but playing with `transformers` is next on my list :smile:

garyb:
Sorry to say I have a bunch of issues to report. :sob:

#### 1. Fundep information doesn't propagate to subclasses:

``` purescript
class Super a b | a -> b where
  go :: a -> b

instance superBool :: Super Boolean Int where
  go _ = 0

class Super a b <= Sub a b where
  goo :: a -> b

instance subBool :: Sub Boolean Int where
  goo _ = 0

test1 = go true -- inferred as `Int`

test2 = goo true -- inferred as `forall t1. (Sub Boolean t1) => t1`
```

#### 2. Wildcard type inference behaves differently than inferred type:

``` purescript
class Super a b | a -> b where
  go :: a -> b

instance superBool :: Super Boolean Int where
  go _ = 0

class Super a b <= Sub a b where
  goo :: a -> b

instance subBool :: Sub Boolean Int where
  goo _ = 0

test1 :: _ -- wildcard inferred as `Int`
test1 = go true

test2 :: _ -- NoInstanceFound: No type class instance was found for Main.Sub Boolean t0
test2 = goo true
```

I can see that wildcard inference might not always work out the same way, but when the whole type is a wildcard I was surprised by this.

#### 3. Overlapping instances arise in situations they didn't used to, and are not correct (as it's raised as an error when there are multiple routes to the same instance). 

This is still a bit convoluted, but is as simple as I could make it (boiled down from a case in transformers, as noted by @rgrempel):

``` purescript
newtype Box a = Box a

class Super a
class Super a <= Sub a

class X a
class X a <= Y a

instance x :: Super a => X (Box a)
instance y :: (Super a, Sub a) => Y (Box a)
```

And the overlap:

```

    Main.Super a0

    $dictSub1 (chosen)
    $dictSuper0
```

In my instance `y` I'm listing `Sub` and then `Super` directly, but it also applies if there are more steps down the chain in either cases (like `SubSub` and `SubSubSub` - the point being there is a shared `Super`). The superclass constraint on `Y` is required to trigger this, as well as all of the instance constraints.

#### 4. I think the `any` / `all` thing might mean we're still going to have problematic behaviour for inference involving rows:

``` purescript
test = do
  foo <- gets _.foo
  pure foo
```

Gives:

```

    Control.Monad.State.Class.MonadState { foo :: t0
                                         | t1
                                         }
                                         t2
```

I'm not sure how much of an issue this will be since it's all good in the presence of type signatures (including situations that probably previously wouldn't have checked without additional annotation).

It does result in some unfortunate things though, like this failing to infer:

``` purescript
test :: String -> Eff _ Unit
test x = liftEff (log x)
```

Because the constraint comes out as `Control.Monad.Eff.Class.MonadEff (console :: CONSOLE | t1) (Eff t0)`. I guess that probably wouldn't have worked before  though.

What this means is we probably can't move the common `Eff`-functions (logging, refs, etc) to be `MonadEff`-constrainted by default (rather than `Eff`-specialised) as it means any code using them won't be able to take advantage of constraint inference since it will just error instead. But I dunno, I can live with that, just a bit of a shame. Maybe this is something that can be improved later.

#### 5. Using wildcards in constraints can infer the wrong (too general) type:

``` purescript
test :: forall m. MonadEff _ m => String -> m Unit
test x = liftEff (log x)
```

The inferred type for the wildcard is `t0`. I'm not sure what should happen here really, whether this should even work at all, but whatever it's doing now isn't right either. I would have hoped it would give `(console :: CONSOLE | t0)`.

paf31:
Thanks Gary, I'll have a look at these today:

1. Fundep information doesn't propagate to subclasses: this definitely seems like a bug.
2. Wildcard type inference behaves differently than inferred type: a type annotation with only a wildcard isn't the same as no annotation, since wildcards only solve monotypes. So I'm going to say "not a bug" for this one, even though it seems counter-intuitive.
3. Overlapping instances arise in situations they didn't used to, and are not correct: something is definitely not quite right with the overlap check, and I think it's been broken for MPTCs since forever, basically. I'll take a look.
4. I think the any / all thing might mean we're still going to have problematic behaviour for inference involving rows: this one is actually caused by the compiler inferring an ambiguous type in a constraint (the tail of the row doesn't appear in the type to the right of the `=>`). I think we can solve this by allowing ambiguous types if they can be solved by some functional dependency.
5. Using wildcards in constraints can infer the wrong (too general) type: I never considered using wildcards to infer types in constraints like this before. I agree it'd be nice though, so I'll take a look and see if it's simple. If not, let's defer this one.

garyb:
All sounds good to me!

Re: 4 that sounds like a nice solution. I wasn't sure what the options were with it really, since it's not a problem with a Haskell analogue, given that it's only related to row types (I think).

With 5 the thing is more that the wildcard-inferred type is wrong - if you plug it in you'll get a type error, and I think generally we prefer to avoid that? :wink: Assuming the ideal version isn't a quick fix, would it be possible to make it into some kind of error instead? Perhaps we can just error on wildcards in constraints generally, as I don't think wildcard inference ever gives us meaningful info?

paf31:
> would it be possible to make it into some kind of error instead

Yes, I think that's best.

paf31:
@garyb It turned out (5) wasn't too tricky after all. Would you mind reviewing the last couple of commits?

I need to think a bit about how to deal with the superclass issue, although I'm tempted to defer it to a later PR, since it doesn't seem to lead to any unsoundness.

garyb:
5 fix is awesome!

I have two more things relating to 4:

This thing:

``` purescript
test :: String -> Eff _ Unit
test x = liftEff (log x)
```

Still doesn't infer (missing instance due to an unknown again), which seems a little surprising since the version with the constraint in the wildcard instead does.

And the other thing is fairly minor, but for this:
``` purescript
test = do
  foo <- gets _.foo
  pure foo
```
The inferred type has a bunch of redundant constraints:
```
   forall t16 t17 t23.
      ( Bind t16
      , MonadState { foo :: t17 t16
                   | t23
                   }
      , Applicative t16
      ) => t16 t17
```
`MonadState` has a `Monad` superclass, so the only constraint needed here is the `MonadState` one.

paf31:
@garyb I've added back the constraint simplifier, and fixed part of the redundant constraint problem, but it's not completely solved. The inferred types for your examples are now

```
   forall t16 t17 t23.
      ( Bind t16
      , MonadState { foo :: t17 t16
                   | t23
                   }
      ) => t16 t17
```

(improved, not fixed) and

```
   forall t1.
      ( Bind t1
      , Applicative t1
      ) => t1 Int
```

(correct)

Now the issue is that if we solve a superclass constraint early, we won't go back and rewrite its dictionaries if a better subclass constraint gets solved along later.

I tried to reproduce the earlier bug with `ala Multiplicative foldMap [1, 2, 3]` but couldn't, so I'll assume that's fixed.

I'm working on the `MonadEff` fix.

paf31:
Alright, the `MonadEff` thing is working too now.

@jdegoes Sorry for not responding. I'd prefer to keep `->` actually - the reason is that I want to think of dependencies as type functions. In fact, I might try to add sugar to give the functions names eventually:

```purescript
class MonadState s m | State :: m -> s where 
  ...
```

so that we can say things like

```purescript
gets :: forall m s. (State m -> s) -> m s
```

jdegoes:
@paf31 That makes sense, and function names for them would be great.

paf31:
Ok, if nobody has any other comments on the code, I'm going to merge this. I've tested it quite a bit at this point.

garyb:
Sounds good to me, I've not had a chance to try it with the latest change, but I was pretty happy with it before that too. :shipit:

PR #2280, opened 2016-08-17
Added typelevel string functions

FrigoEU:
Probably will not be merged as-is, but I've been working on it for a couple days and would like to show what I have and see if it's the right direction.

I added the TypeString and TypeConcat functions to Prim. TypeString is supposed to just convert a type to a symbol, so it can be used in TypeConcat with a type-level-string-literal to make more descriptive error messages (see the example for a case). I didn't add a concat operator to Prim, but you can easily do it yourself (as shown in the example).

Main changes are in the Pretty Printing file, where I'm "evaluating" the two functions. I just made hard coded implementations for them and reused the insertPlaceHolders function to recurse down through all the types and insert a specific placeholder for TypeConcat, which I'm later "evaluating" in the matchType function. I decided to not use a placeholder for TypeString since I basically just want to show the type and that's what will happen if I strip the call to TypeString.

I added something that I think is not particularly good in Parser/Declarations.hs. The case I was having problems with was this:

````
-- Doesn't get parsed correctly, returns a "Unknown type Show" exception
instance dontShowMyType :: Fail (TypeConcat (TypeString a) "bla") => Show (MyType a) where
  show _ = ""

-- Gets correctly parsed, notice the double parens
instance dontShowMyType :: Fail ((TypeConcat (TypeString a) "bla")) => Show (MyType a) where
  show _ = ""
````

With the change I did it works, but I'm having a hard time assessing the impact of my change, so some guidance here would be very much appreciated.

### DIFF
--- a/examples/failing/ProgrammableTypeErrorsTypeString.purs
+++ b/examples/failing/ProgrammableTypeErrorsTypeString.purs
@@ -1,0 +1,18 @@
+-- @shouldFailWith NoInstanceFound
+
+module Main where
+
+import Prelude
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (log)
+
+newtype MyType a = MyType a
+
+instance cannotShowFunctions :: Fail ("Don't want to show" <> TypeString (MyType a) <> "because.") => Show (MyType a) where
+  show _ = "unreachable"
+
+infixl 6 type TypeConcat as <>
+
+main :: Eff _ _
+main = do
+  log $ show (MyType 2)
--- a/examples/failing/ProgrammableTypeErrorsTypeString.purs
+++ b/examples/failing/ProgrammableTypeErrorsTypeString.purs
@@ -8,7 +8,7 @@ import Control.Monad.Eff.Console (log)
 
 newtype MyType a = MyType a
 
-instance cannotShowFunctions :: Fail ("Don't want to show" <> TypeString (MyType a) <> "because.") => Show (MyType a) where
+instance cannotShowFunctions :: Fail ("Don't want to show " <> TypeString (MyType a) <> " because.") => Show (MyType a) where
   show _ = "unreachable"
 
 infixl 6 type TypeConcat as <>
--- a/src/Language/PureScript/Environment.hs
+++ b/src/Language/PureScript/Environment.hs
@@ -242,7 +242,7 @@ primTypes =
     , (primName "Partial",  (Star, ExternData))
     , (primName "Fail",     (FunKind Symbol Star, ExternData))
     , (primName "TypeString", (FunKind Star Symbol, ExternData))
-    , (primName "TypeConcat", (FunKind Symbol Symbol, ExternData))
+    , (primName "TypeConcat", (FunKind Symbol (FunKind Symbol Symbol), ExternData))
     ]
 
 -- |
--- a/src/Language/PureScript/Environment.hs
+++ b/src/Language/PureScript/Environment.hs
@@ -241,6 +241,8 @@ primTypes =
     , (primName "Boolean",  (Star, ExternData))
     , (primName "Partial",  (Star, ExternData))
     , (primName "Fail",     (FunKind Symbol Star, ExternData))
+    , (primName "TypeString", (FunKind Star Symbol, ExternData))
+    , (primName "TypeConcat", (FunKind Symbol Symbol, ExternData))
     ]
 
 -- |
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -33,6 +33,7 @@ import Language.PureScript.Types
 import Language.PureScript.Pretty.Common (endWith)
 import qualified Language.PureScript.Bundle as Bundle
 import qualified Language.PureScript.Constants as C
+import Language.PureScript.Pretty.Common (before)
 
 import qualified System.Console.ANSI as ANSI
 
@@ -597,7 +598,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
     renderSimpleErrorMessage OverlappingInstances{} = internalError "OverlappingInstances: empty instance list"
     renderSimpleErrorMessage (NoInstanceFound (Constraint C.Fail [ ty ] _)) | Just box <- toTypelevelString ty =
       paras [ line "A custom type error occurred while solving type class constraints:"
-            , indent $ box
+            , indent box
             ]
     renderSimpleErrorMessage (NoInstanceFound (Constraint C.Partial
                                                           _
@@ -1209,7 +1210,7 @@ toTypelevelString (TypeLevelString s) = Just $ Box.text s
 toTypelevelString (TypeApp (TypeConstructor f) x)
   | f == primName "TypeString" = Just $ typeAsBox x
 toTypelevelString (TypeApp (TypeApp (TypeConstructor f) x) ret)
-  | f == primName "TypeConcat" = (Box.<>) <$> (toTypelevelString x) <*> (toTypelevelString ret)
+  | f == primName "TypeConcat" = before <$> (toTypelevelString x) <*> (toTypelevelString ret)
 toTypelevelString _ = Nothing
 
 -- |
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -19,7 +19,7 @@ import Data.Either (lefts, rights)
 import Data.Foldable (fold)
 import Data.Functor.Identity (Identity(..))
 import Data.List (intercalate, transpose, nub, nubBy, sortBy, partition)
-import Data.Maybe (maybeToList, fromMaybe, mapMaybe)
+import Data.Maybe (maybeToList, fromMaybe, mapMaybe, isJust)
 import Data.Ord (comparing)
 import Data.Text (strip, pack, unpack)
 import qualified Data.Map as M
@@ -596,9 +596,9 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
             , line "They may be disallowed completely in a future version of the compiler."
             ]
     renderSimpleErrorMessage OverlappingInstances{} = internalError "OverlappingInstances: empty instance list"
-    renderSimpleErrorMessage (NoInstanceFound (Constraint C.Fail [ ty ] _)) =
+    renderSimpleErrorMessage (NoInstanceFound (Constraint C.Fail [ ty ] _)) | isJust (toTypelevelString ty) =
       paras [ line "A custom type error occurred while solving type class constraints:"
-            , markCodeBox $ indent $ typeAsBox $ evalTypelevelFunctions ty
+            , indent $ line $ maybe "" id (toTypelevelString ty)
             ]
     renderSimpleErrorMessage (NoInstanceFound (Constraint C.Partial
                                                           _
@@ -1205,13 +1205,18 @@ renderBox = unlines
   dropWhileEnd p = reverse . dropWhile p . reverse
   whiteSpace = all isSpace
 
+toTypelevelString :: Type -> Maybe String
+toTypelevelString t = case evalTypelevelFunctions t of
+  (TypeLevelString str) -> Just str
+  _ -> Nothing
+
 evalTypelevelFunctions :: Type -> Type
 evalTypelevelFunctions = everywhereOnTypes go
   where
   go (TypeApp (TypeConstructor f) x)
     | f == primName "TypeString" = TypeLevelString $ unpack $ strip $ pack $ prettyPrintTypeAtom x
   go (TypeApp (TypeApp (TypeConstructor f) (TypeLevelString x)) (TypeLevelString ret))
-    | f == primName "TypeConcat" = TypeLevelString (x ++ " " ++ ret)
+    | f == primName "TypeConcat" = TypeLevelString (x ++ ret)
   go t = t
 
 -- |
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: src/Language/PureScript/Environment.hs [hunk 1]
```diff
@@ -241,6 +241,8 @@ primTypes =
     , (primName "Boolean",  (Star, ExternData))
     , (primName "Partial",  (Star, ExternData))
     , (primName "Fail",     (FunKind Symbol Star, ExternData))
+    , (primName "TypeString", (FunKind Star Symbol, ExternData))
+    , (primName "TypeConcat", (FunKind Symbol Symbol, ExternData))
```
paf31: This doesn't seem right, shouldn't it be `Symbol -> Symbol -> Symbol`?

src/Language/PureScript/Pretty/Types.hs
```diff
@@ -32,7 +32,7 @@ typeLiterals = mkPattern match
   where
   match TypeWildcard{} = Just $ text "_"
   match (TypeVar var) = Just $ text var
-  match (TypeLevelString s) = Just . text $ show s
+  match (TypeLevelString s) = Just $ text s
```
paf31: Unfortunately this will cause problems when type level string is rendered outside of a `Fail`.

src/Language/PureScript/Pretty/Types.hs
```diff
@@ -99,6 +99,10 @@ kinded = mkPattern match
 insertPlaceholders :: Type -> Type
 insertPlaceholders = everywhereOnTypesTopDown convertForAlls . everywhereOnTypes convert
   where
+  -- Replacing applications of type level functions with placeholders
+  -- that we'll interpret later.
+  convert (TypeApp (TypeApp (TypeConstructor f) x) ret) | f == primName "TypeConcat" = PrettyPrintTypeApplication "TypeConcat" x ret
```
paf31: I'd rather specialize to `PrettyPrintTypeConcat` or something, instead of putting a string in the type like this.

paf31:
Thanks for working on this!

I'm not sure about mixing evaluation into the pretty printer like this, maybe we should think of alternatives. Originally I had planned to implement the evaluator in the `Errors` module actually.

src/Language/PureScript/Pretty/Types.hs
FrigoEU: Right, I see that now. Will change.

src/Language/PureScript/Environment.hs
FrigoEU: That is weird indeed, total typo, should be (Funkind Symbol (Funkind Symbol Symbol)). I'll walk through the kind checker to see why it's not picking this up.

src/Language/PureScript/Pretty/Types.hs
FrigoEU: I was thinking that as well, will do.

src/Language/PureScript/Environment.hs
FrigoEU: OK, I looked up why the kind checker doesn't flag this. It's because the kind checking of the instance happens when we're in typeCheckAll and we reach the TypeClassDictionaryAppConstructor. However, we don't get there because we first check the main function, and hit the "NoInstanceFound" check in typesOf. When you remove the main function, you do get the kinds error, as expected.

The ordering of declarations is still a blind spot for me at the moment, so not sure yet how to fix it.

FrigoEU:
Next try! I took the evaluation out of the pretty printing and just made an evalTypelevelFunctions function inside Errors. It's quite a bit simpler I think. TypeString becomes an evaluation from a type to a TypeLevelString, TypeConcat concatenates two TypeLevelStrings into one. The first remark then also becomes irrelevant.

I didn't add the evalTypelevelFunctions everywhere, just on the specific Fail error message. I thought it was safer for the time being since we know we really only need it there for the moment. 

Not sure if I can do this unpack $ strip $ pack better, but it doesn't really bother me that much.

src/Language/PureScript/Errors.hs
```diff
@@ -1195,6 +1197,15 @@ renderBox = unlines
   dropWhileEnd p = reverse . dropWhile p . reverse
   whiteSpace = all isSpace
 
+evalTypelevelFunctions :: Type -> Type
+evalTypelevelFunctions = everywhereOnTypes go
+  where
+  go (TypeApp (TypeConstructor f) x)
+    | f == primName "TypeString" = TypeLevelString $ unpack $ strip $ pack $ prettyPrintTypeAtom x
+  go (TypeApp (TypeApp (TypeConstructor f) (TypeLevelString x)) (TypeLevelString ret))
+    | f == primName "TypeConcat" = TypeLevelString (x ++ " " ++ ret)
```
paf31: I don't think we should add a space here.

examples/failing/ProgrammableTypeErrorsTypeString.purs
```diff
@@ -0,0 +1,18 @@
...
+
+module Main where
+
+import Prelude
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (log)
+
+newtype MyType a = MyType a
+
+instance cannotShowFunctions :: Fail ("Don't want to show" <> TypeString (MyType a) <> "because.") => Show (MyType a) where
```
paf31: You know what might be neat later - type level string interpolation:

```purescript
instance cannotShowFunctions :: Fail "Don't want to show {{MyType a}} because." => Show (MyType a)
```

src/Language/PureScript/Errors.hs
```diff
@@ -1195,6 +1197,15 @@ renderBox = unlines
   dropWhileEnd p = reverse . dropWhile p . reverse
   whiteSpace = all isSpace
 
+evalTypelevelFunctions :: Type -> Type
```
paf31: There's something not quite right about this. Evaluating something of kind `Symbol` should result in a pretty printed string, but everything else should just show the regular `NoInstanceFound` error. Right now, if you have a concatenated string inside another type constructor, you'll get a mixture of the two.

I think you have to have a function of type `Type -> Maybe String` which tries to evaluate a string, and fails if it gets something else.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -190,7 +190,8 @@ parseTypeClassDeclaration = do
 
 parseConstraint :: TokenParser Constraint
 parseConstraint = Constraint <$> parseQualified properName
-                             <*> P.many (noWildcards parseTypeAtom)
+                             <*> (P.many (noWildcards (parens parsePolyType)
```
paf31: Why did this need to change again? This doesn't look quite right.

examples/failing/ProgrammableTypeErrorsTypeString.purs
FrigoEU: That would be cool indeed :). But that would need changes to the parser and the AST I think. Maybe we keep that for a seperate PR?

src/Language/PureScript/Errors.hs
FrigoEU: Right, I interpreted it more as a general function like "Just evaluate whatever you can on the type level!". I think we can still keep this general function, but then in the actual error message handling I'll check if after evaluation, I end up with something of kind Symbol and then display it, otherwise skip to the next case.

src/Language/PureScript/Parser/Declarations.hs
FrigoEU: It was for this case:
````
-- Doesn't get parsed correctly, returns a "Unknown type Show" exception
instance dontShowMyType :: Fail (TypeConcat (TypeString a) "bla") => Show (MyType a) where
  show _ = ""

-- Gets correctly parsed, notice the double parens
instance dontShowMyType :: Fail ((TypeConcat (TypeString a) "bla")) => Show (MyType a) where
  show _ = ""
````
I mentioned it also in the first message on this PR, I'm not sure if it actually is doing anything wrong, but I agree that it does look ugly.

File: src/Language/PureScript/Errors.hs (hunk 4 above)
Comment (FrigoEU) [inline src/Language/PureScript/Errors.hs]: Right, my bad.

File: src/Language/PureScript/Errors.hs (hunk 6 above)
Comment (paf31) [inline src/Language/PureScript/Errors.hs]: You definitely shouldn't need to kind check anything. I think this is fine, but we should add a check to make sure we're showing a `TypeLevelString`.

File: src/Language/PureScript/Parser/Declarations.hs (hunk 7 above)
Comment (paf31) [inline src/Language/PureScript/Parser/Declarations.hs]: There's definitely an issue, I just think this isn't the right fix. The issue seems to be that it's parsed as 

```
Fail ((TypeConcat (TypeString a) "bla") => Show (MyType a))
```

but constrained types shouldn't be allowed in instance heads.

File: examples/failing/ProgrammableTypeErrorsTypeString.purs (hunk 5 above)
Comment (paf31) [inline examples/failing/ProgrammableTypeErrorsTypeString.purs]: Yes, definitely a separate PR.

File: src/Language/PureScript/Errors.hs [hunk 8]
```diff
@@ -594,9 +594,9 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
             , line "They may be disallowed completely in a future version of the compiler."
             ]
     renderSimpleErrorMessage OverlappingInstances{} = internalError "OverlappingInstances: empty instance list"
-    renderSimpleErrorMessage (NoInstanceFound (Constraint C.Fail [ ty ] _)) =
+    renderSimpleErrorMessage (NoInstanceFound (Constraint C.Fail [ ty ] _)) | isJust (toTypelevelString ty) =
```
paf31: You can use a pattern guard here, to avoid evaluating twice.

src/Language/PureScript/Errors.hs
```diff
@@ -1197,13 +1197,18 @@ renderBox = unlines
   dropWhileEnd p = reverse . dropWhile p . reverse
   whiteSpace = all isSpace
 
+toTypelevelString :: Type -> Maybe String
+toTypelevelString t = case evalTypelevelFunctions t of
+  (TypeLevelString str) -> Just str
```
paf31: These parens aren't needed.

src/Language/PureScript/Errors.hs
```diff
@@ -1197,13 +1197,18 @@ renderBox = unlines
...
+  _ -> Nothing
+
 evalTypelevelFunctions :: Type -> Type
 evalTypelevelFunctions = everywhereOnTypes go
   where
   go (TypeApp (TypeConstructor f) x)
     | f == primName "TypeString" = TypeLevelString $ unpack $ strip $ pack $ prettyPrintTypeAtom x
```
paf31: Instead of packing and unpacking, I'd prefer an interpretation function which returns `Maybe Box`. That way, we can handle newlines properly, as well as multi-line types.

FrigoEU:
At first I tried to migrate TypeLevelString String to TypeLevelString Box, but Box doesn't have an Ord or an Eq instance, so I decided against that.

I changed toTypeLevelString to just be a simple recursion. It's small and simple, but it's just for TypeLevelString's. I think that's OK though. We'll see later if we ever want to expand this to different types.

src/Language/PureScript/Errors.hs
```diff
@@ -1195,6 +1196,14 @@ renderBox = unlines
   dropWhileEnd p = reverse . dropWhile p . reverse
   whiteSpace = all isSpace
 
+toTypelevelString :: Type -> Maybe Box.Box
+toTypelevelString (TypeLevelString s) = Just $ Box.text s
+toTypelevelString (TypeApp (TypeConstructor f) x)
+  | f == primName "TypeString" = Just $ typeAsBox x
+toTypelevelString (TypeApp (TypeApp (TypeConstructor f) x) ret)
+  | f == primName "TypeConcat" = (Box.<>) <$> (toTypelevelString x) <*> (toTypelevelString ret)
```
paf31: Instead of `<>`, we can use `before`, which will correctly handle boxes with multiple lines, like some types.

src/Language/PureScript/Errors.hs
```diff
@@ -592,9 +593,9 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
             , line "They may be disallowed completely in a future version of the compiler."
             ]
     renderSimpleErrorMessage OverlappingInstances{} = internalError "OverlappingInstances: empty instance list"
-    renderSimpleErrorMessage (NoInstanceFound (Constraint C.Fail [ TypeLevelString message ] _)) =
+    renderSimpleErrorMessage (NoInstanceFound (Constraint C.Fail [ ty ] _)) | Just box <- toTypelevelString ty =
       paras [ line "A custom type error occurred while solving type class constraints:"
-            , indent . paras . map line . lines $ message
+            , indent $ box
```
paf31: `$` is redundant here (from HLint)

FrigoEU:
Thanks for all the help with my beginner errors :).

paf31:
Thanks for working on this!

I just need to stare at the parser changes a bit, then I'll get this merged in hopefully.

src/Language/PureScript/Parser/Types.hs
```diff
@@ -126,3 +127,11 @@ parseRowEnding = P.option REmpty $ indented *> pipe *> indented *> parseType
 
 parseRow :: TokenParser Type
 parseRow = (curry rowFromList <$> commaSep (parseNameAndType parsePolyType) <*> parseRowEnding) P.<?> "row"
+
+parseConstraintArg :: TokenParser Type
+parseConstraintArg = indented *> P.choice [ P.try parseFunction
```
paf31: I think we need `parseRow` here too.

But I'm wondering if we should just take `parseConstrainedType` out of `parseTypeAtom`, and try to parse `=>` like a type operator. Let me think about this a bit.

paf31:
@FrigoEU Would you mind updating this now that the parser change is done? Thanks!

FrigoEU:
Ok, done! I rebased my branch back on master, since I thought it would be better for commit continuity. Hope that's ok. I just went back to parseTypeAtom for parsing constraints in this last commit, the rest didn't change.

paf31:
Thanks very much! Sorry for all the nitpicking, but I'm glad we got it figured out.

FrigoEU:
Thanks for all the help! Can't wait to get some more work done.

PR #2281, opened 2016-08-17
Fix most HLint warnings

ianbollinger:
This leaves in place 10 warnings, mostly for reducing duplication.

### DIFF
--- a/src/Language/PureScript/Publish/ErrorsWarnings.hs
+++ b/src/Language/PureScript/Publish/ErrorsWarnings.hs
@@ -182,10 +182,10 @@ displayUserError e = case e of
       , spacer
       ] ++ spdxExamples ++
       [ spacer
-      , para (concat
-          [ "Note that distributing code without a license means that nobody "
-          , "will (legally) be able to use it."
-          ])
+      , para (
+          "Note that distributing code without a license means that nobody "
+          ++ "will (legally) be able to use it."
+          )
       , spacer
       , para (concat
           [ "It is also recommended to add a LICENSE file to the repository, "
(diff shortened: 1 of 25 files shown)


src/Language/PureScript/Publish/ErrorsWarnings.hs
```diff
@@ -182,10 +182,10 @@ displayUserError e = case e of
       , spacer
       ] ++ spdxExamples ++
       [ spacer
-      , para (concat
-          [ "Note that distributing code without a license means that nobody "
-          , "will (legally) be able to use it."
-          ])
+      , para (
```
hdgarrood: Does HLint not warn for the other instances of `concat` + a list literal? Changing this makes it inconsistent within the file. I don't really mind which approach we take but I do want it to be consistent within this file.

ianbollinger: It only complains if it's given a list literal with two elements.

paf31:
@hdgarrood Good to merge then?

hdgarrood:
I'd still rather keep it consistent within that file that I commented on. Other than that, it looks great.

ianbollinger:
Er, it was already inconsistent within that file. (Using either ++ or concat.) But I can revert the edits if you like.

hdgarrood:
How is it inconsistent currently?

ianbollinger:
[Here](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Publish/ErrorsWarnings.hs#L101), [here](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Publish/ErrorsWarnings.hs#L126), [here](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Publish/ErrorsWarnings.hs#L232), and [here](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Publish/ErrorsWarnings.hs#L268).

hdgarrood:
Ah, wasn't aware of those. Might as well merge as-is then.

paf31:
Thanks!

PR #2282, opened 2016-08-18
Import suggestions for UnusedDctorImport's

FrigoEU:
This one has been annoying me for some time, especially because psc-ide's import functionality imports both the type constructor and data constructor automatically. This is the correct behavior a lot of the time, but just as often I want to import only the type constructor and then I need to go and manually correct the warning, very annoying! I'm getting more familiar with the compiler so I thought I'd take a stab at this.

I added the same suggestion for UnusedDctorImport and UnusedDctorExplicitImport as is already used for UnusedExplicitImport. I added the moduleName, possible qualification for the module and the used declaration references to both data constructors, and then use them to make the import suggestion. Maybe it's worth it to pull this moduleName/qual/declrefs into a type synonym or something, but I wasn't sure if that would have any real value at this point.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -108,8 +108,8 @@ data SimpleErrorMessage
   | ImportHidingModule ModuleName
   | UnusedImport ModuleName
   | UnusedExplicitImport ModuleName [String] (Maybe ModuleName) [DeclarationRef]
-  | UnusedDctorImport (ProperName 'TypeName)
-  | UnusedDctorExplicitImport (ProperName 'TypeName) [ProperName 'ConstructorName]
+  | UnusedDctorImport ModuleName (ProperName 'TypeName) (Maybe ModuleName) [DeclarationRef]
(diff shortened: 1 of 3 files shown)


rvion:
>  but just as often I want to import only the type constructor and then I need to go and manually correct the warning

(This, or the opposite) This is one of the very minor annoyance I'd be happy to see fixed too :)

FrigoEU:
Not sure if it's clear, but this PR is about this case: 
````
module Main where
import Data.Tuple (Tuple(Tuple))
data MyType a = MyType (Tuple a a)
````
This will get a warning of the form "The import of type Tuple includes data constructors but only the type is used". This PR adds a suggestion (that can automatically be applied by any of the psc-ide editor plugins) to change to this:
````
module Main where
import Data.Tuple (Tuple)
data MyType a = MyType (Tuple a a)
````

paf31:
Is it possible to add this to the suggestions in the JSON output, so tools like `pscid` can perform the replacement automatically?

FrigoEU:
I just tested it with psc-ide-vim, psc-ide-emacs and pscid again and they all seem to pick up these suggestions and can apply them automatically. Could you elaborate on which case and which tool is not able to automatically apply these?

paf31:
Sorry, I misunderstood something. Looks great, thanks!

PR #2286, opened 2016-08-24
[psc-bundle] Verify entry points exist

### LINKED ISSUE #1755: psc-bundle: error if the main module doesn't exist
It would be nice if `psc-bundle` would error if the provided `--main` module does not exist.

For example:
```
$ psc-bundle 'output/*/*.js' --main ThisDoesntExist
```
currently succeeds, but puts a call to `PS["ThisDoesntExist"].main()` at the end, which will fail at runtime. Instead, I think psc-bundle should fail and not produce any output.

kRITZCREEK:
Throws an error when an entry point could not be found in the input
modules.

Fixes #1755 

I've checked for all the entry points instead of the passed `--main` module here. I'm happy to change that, but it seemed like the more sensible choice to me.


hdgarrood:
Does this address the case where `--main` is specified but not `--module`?

kRITZCREEK:
Could you elaborate? What command should I try? I'm not all that familiar with `psc-bundle`

paf31:
I think @hdgarrood is saying that `mainModule` also needs to be in the list of parsed modules. So something like `psc-bundle *.purs --main This.Does.Not.Exist`.

hdgarrood:
Right, exactly. Also I'm a bit surprised it's necessary to check entry points, come to think of it. What happens with 0.9.3 if you specify a nonexistent entry point with ‘--module`?

(Also 'entry point' doesn't quite feel right; is that what we usually refer to modules specified with the `--module` flag as?)

kRITZCREEK:
> Does this address the case where --main is specified but not --module?

`psc-bundle output/**/*.js --main Main` doesn't perform any DCE, but it bundles all the found modules and adds a `PS["Main"].main();` to the end of the file. For a non-existing module like `Mai` it adds `PS["Mai"].main();` to the end of the file, so that should probably also be checked.

> Right, exactly. Also I'm a bit surprised it's necessary to check entry points, come to think of it. What happens with 0.9.3 if you specify a nonexistent entry point with ‘--module`?

This originally came up because someone on IRC copied an example file to `Main.purs` but didn't change the module header so it still said `Test.Main`. When he then ran `pulp build --to example.js`, he got an empty bundle that called out to the non-existent Main module. I'm not sure what flags pulp passes on to psc-bundle, but it throws the right error after applying this patch.

>(Also 'entry point' doesn't quite feel right; is that what we usually refer to modules specified with the --module flag as?)

At least in the README for psc-bundle we do https://github.com/purescript/purescript/tree/master/psc-bundle#usage.

hdgarrood:
Ah great, thanks for answering my questions! In that case this looks great, +1: to merge. The wording issue with 'entry point' isn't a big deal and can be addressed separately anyway (if it needs addressing at all).

paf31:
Thanks!

PR #2291, opened 2016-08-28
Functional Dependencies

paf31:
This adds actual functional dependencies to the work in my other PR.

We should review and test this carefully. I've got some simple tests working here, but I'd like to put it through its paces quite a bit before releasing it.

TODO:

- [x] Add an error message hint along the lines of "while checking functional dependency a -> b for selected instance ..."
- [x] Add some failing tests
- [x] Go back to the old approach where we check that type vars unify before committing to an instance

### DIFF
--- a/examples/failing/OverlappingVars.purs
+++ b/examples/failing/OverlappingVars.purs
@@ -1,4 +1,4 @@
--- @shouldFailWith TypesDoNotUnify
+-- @shouldFailWith NoInstanceFound
 module Main where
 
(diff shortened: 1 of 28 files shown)


garyb:
I'm super excited about this, haven't commented or tried anything out yet as I'm in a busy patch right now! 

I'm on holiday next week so might get some time to catch up with PRs and stuff, but then again the if the weather is good I might not :wink:

PR #2294, opened 2016-08-30
Safer installation from source in INSTALL.md

hdgarrood:
This commit ensures that the stackage snapshot used when building from
source following the instructions in INSTALL.md will be the same as the
one used in development and in binary bundles.

This is safer (because you'll end up with a set of dependencies which
have been tested together) and it also means you'll always get the
latest version according to Hackage. In comparison, the Stackage nightly
is often behind because of waiting for purescript or other libraries to
make changes to allow all packages in the nightly to build together.

See also discussion in #2278.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,9 +11,8 @@ If you would like to contribute, please consider the issues in the current miles
 Please follow the following guidelines:
 
 - Add at least a test to `examples/passing/` and possibly to `examples/failing`.
-- Build the binaries and libs with `cabal build`
-- Install the binaries and libs with `cabal install`.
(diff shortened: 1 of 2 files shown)


hdgarrood:
Note also that this brings `INSTALL.md` in line with what's already on the website.

kRITZCREEK:
Not that it needs to happen in this PR, but the CONTRIBUTING.md file is also out of date (it still lists `cabal --enable-tests ...`).

hdgarrood:
Oh thanks for the heads up, might as well take care of that now too.

hdgarrood:
Oops, accidentally pushed this most recent to commit to master (but reverted again).

paf31:
Thanks for this!

PR #2295, opened 2016-08-31
display error hint. Fixes #2250

### LINKED ISSUE #2250: Messages for unexpected record update syntax
I factored out a record update and have been bitten by leftover `=`s a few times. Here are difficult to sort errors I encountered:

```hs
r = g {foo = "bar"}
{-
Could not match type
    Record
  with type
    Function
      { foo :: t2
      | t3
      }
-}
g {foo} = foo
```

```hs
r' = {foo = "bar"}
{-
  Unable to parse module:
  unexpected =
  expecting :, , or }
-}
```

```hs
f baz = g
  { foo = "bar"
  , baz -- punny
  }
{-
  Unable to parse module:
  unexpected {
  expecting no indentation or end of input
-}
```

Both former messages are fairly good. An extra step to point out record literal versus update syntax would've made it obvious. If possible it'd be nice to have a special case message for these like <q>Did you mean `({ field: value })` instead of `({ field = value })`?</q>

kRITZCREEK:
Got here through https://www.codetriage.com/. I'll try to help reduce the number of issues a bit.

```purescript
r' = {foo = "bar"}
```

Now produces the following error:

```
Error found:
at /home/creek/Documents/purs/temp/src/Main.purs line 7, column 11 - line 7, column 11

  Unable to parse module:
  unexpected =
  expecting :, , or }

  You are using record update syntax {foo = ...} inside a record literal.
  Did you mean to write {foo : ...} instead?


See https://github.com/purescript/purescript/wiki/Error-Code-ErrorParsingModule for more information,
```

> A remaining issue is that `r' = {foo ? "bar"}` now produces the following expecting error:

> ```
  Unable to parse module:
  unexpected ?
  expecting :, =, , or }
```

> The `=` sign shouldn't appear in there, but maybe someone knows how I can avoid that from appearing? I'm already using the `lookAhead` combinator but that doesn't seem to cut it.

Fixed that by using `<?>`

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -336,8 +336,8 @@ parseIdentifierAndValue =
     return (name, b)
   <|> (do name <- C.indented *> stringLiteral; b <- rest name; return (name, b))
   where
-  rest name = C.indented *> (colon <|> P.lookAhead equals *> fail (updateSyntaxError name)) *> C.indented *> parseValue
-  updateSyntaxError name = "\nYou are using record update syntax {" ++ name ++ " = ...} inside a record literal.\nDid you mean to write {" ++ name ++ " : ...} instead?"
+  rest name = C.indented *> (colon <|> P.try (P.lookAhead equals) *> fail (updateSyntaxError name)) *> C.indented *> parseValue
+  updateSyntaxError name = "\n  You are using record update syntax {" ++ name ++ " = ...} inside a record literal.\n  Did you mean to write {" ++ name ++ " : ...} instead?"
 
 parseAbs :: TokenParser Expr
 parseAbs = do
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -332,11 +332,12 @@ parseIdentifierAndValue :: TokenParser (String, Expr)
 parseIdentifierAndValue =
   do
     name <- C.indented *> lname
-    b <- P.option (Var $ Qualified Nothing (Ident name)) rest
+    b <- P.option (Var $ Qualified Nothing (Ident name)) (rest name)
     return (name, b)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by texastoland** [comment]
Where's the 😍 button 🙌

File: src/Language/PureScript/Parser/Declarations.hs [hunk 1]
```diff
@@ -332,11 +332,13 @@ parseIdentifierAndValue :: TokenParser (String, Expr)
...
     name <- C.indented *> lname
-    b <- P.option (Var $ Qualified Nothing (Ident name)) rest
+    b <- P.option (Var $ Qualified Nothing (Ident name)) (rest name)
     return (name, b)
-  <|> (,) <$> (C.indented *> stringLiteral) <*> rest
+  <|> (do name <- C.indented *> stringLiteral; b <- rest name; return (name, b))
   where
-  rest = C.indented *> colon *> C.indented *> parseValue
+  rest name = C.indented *> separator name *> C.indented *> parseValue
+  separator name = colon <|> P.try (P.lookAhead equals) *> fail (updateSyntaxError name) P.<?> ":"
```
paf31: I would typically address this by either:

- Making the two constructs use the same syntax (i.e. updates and literals would share the opening `{` and label) and handling errors later, or
- Just using `<?>` to provide a better error for `:`. As I see it, the existing error is fine, but could be improved by giving `:` a name.

Anyway, let me think about this a bit. The reason I left this in Discussion was that I wanted to think about how to handle the remaining ambiguities in the parser. This is a fine fix, but I think there might be a different fix which will handle ambiguities better in general.

kRITZCREEK:
Closing as the original issue was rejected.

garyb:
@kRITZCREEK I don't think we realised/remembered there was a PR for this, maybe it is still worth having for now? (Ping @paf31.)

kRITZCREEK:
I actually think it's the right call to not merge this. It's a very ad-hoc solution and would better be solved by making the grammer unambiguous around record update/literal syntax.

PR #2296, opened 2016-09-02
remove a pursuit workaround

kRITZCREEK:
* make imports explicit
* clean up imports

See https://github.com/purescript/pursuit/issues/153

### DIFF
--- a/src/Language/PureScript/Ide/Pursuit.hs
+++ b/src/Language/PureScript/Ide/Pursuit.hs
@@ -12,33 +12,31 @@
 -- Pursuit client for psc-ide
 -----------------------------------------------------------------------------
 
-{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 module Language.PureScript.Ide.Pursuit
   ( searchPursuitForDeclarations
   , findPackagesForModuleIdent
   ) where
 
-import           Protolude                     hiding (fromStrict)
+import           Protolude
 
 import qualified Control.Exception             as E
-import           Data.Aeson
-import           Data.ByteString.Lazy          (fromStrict)
-import           Data.String
-import qualified Data.Text                     as T
+import           Data.Aeson                    (Array, Result (Success), decode,
+                                                fromJSON)
 import           Language.PureScript.Ide.Types
 import           Network.HTTP.Types.Header     (hAccept)
-import           Pipes.HTTP
+import           Pipes.HTTP                    (HttpException, newManager,
+                                                parseRequest, queryString,
+                                                requestHeaders, responseBody,
+                                                tlsManagerSettings, withHTTP)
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Ide/Pursuit.hs
```diff
@@ -12,33 +12,31 @@
...
+                                                requestHeaders, responseBody,
+                                                tlsManagerSettings, withHTTP)
 import qualified Pipes.Prelude                 as P
 
--- We need to remove trailing dots because Pursuit will return a 400 otherwise
--- TODO: remove this when the issue is fixed at Pursuit
 queryPursuit :: Text -> IO ByteString
 queryPursuit q = do
-    let qClean = T.dropWhileEnd (== '.') q
     req' <- parseRequest "http://pursuit.purescript.org/search"
```
paf31: Can we make this HTTPS now too?

paf31:
👍  thanks

src/Language/PureScript/Ide/Pursuit.hs
kRITZCREEK: I'll try if it works

kRITZCREEK:
Well actually Pursuit support seems to be totally broken atm...

````
> http GET "https://pursuit.purescript.org/search?query=map" -j    
HTTP/1.1 302 Found
Cache-Control: no-cache, must-revalidate
Connection: keep-alive
Content-Length: 0
Content-Type: text/plain; charset=utf-8
Date: Sat, 03 Sep 2016 20:40:30 GMT
Expires: Thu, 01 Jan 1970 05:05:05 GMT
Location: https://pursuit.purescript.org/
Server: nginx/1.4.6 (Ubuntu)
Set-Cookie: _SESSION=yT9qUj+Htu49rbQyxcjV7JGJQ9+H1sC1yFT7+4nKgXToLCCDvvRRaSsoUlz7LHgHycdopLgcS4n8aQWPkimMPdpuDUly8z4tqKXhN/0luC5/NG/tHma8p+BkjMl/33gMuejikpGL3c8=; Path=/; Expires=Sat, 03-Sep-2016 22:40:30 GMT; HttpOnly
Vary: Accept, Accept-Language
X-Cache: Miss
```

It doesn't seem to return any `JSON` when queried with the `Accept: application/json` header.

paf31:
@kRITZCREEK So should this be merged or not? Are you saying Pursuit isn't serving JSON responses at all right now?

kRITZCREEK:
@paf31 yes, I'm saying that. I don't think this should be merged as is. I think we either

- leave psc-ide pursuit support as it is and fix it once pursuit returns json again or
- remove pursuit support from `psc-ide`

I'd favor the latter as the http libraries seem to have a huge dependency footprint and pursuit support never was much better than a simple webscraper anyway.

PR #2299, opened 2016-09-05
Fix weird parser issue identified by @FrigoEU

paf31:
@FrigoEU Could you please review this one? It should unblock your PR too.

Thanks!


FrigoEU:
This looks good to me. I also tested against the problem cases I encountered and they're all working fine now when just using parseTypeAtom, so thumbs up from me!

paf31:
Thanks!

PR #2300, opened 2016-09-07
Fix inlining for apply operators

### LINKED ISSUE #2298: `apply` and `applyFlipped` are no longer inlined
I'm pretty sure `$` and `#` both used to be inlined to a standard function call when saturated, but this is no longer the case.

garyb:
Resolves #2298

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
+++ b/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
@@ -4,7 +4,6 @@
 module Language.PureScript.CodeGen.JS.Optimizer.Inliner
   ( inlineVariables
   , inlineCommonValues
-  , inlineOperator
   , inlineCommonOperators
   , inlineFnComposition
   , etaConvert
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
```diff
@@ -107,6 +107,7 @@ inlineOperator (m, op) f = everywhereOnJS convert
   convert (JSApp _ (JSApp _ op' [x]) [y]) | isOp op' = f x y
   convert other = other
   isOp (JSIndexer _ (JSStringLiteral _ op') (JSVar _ m')) = m == m' && op == op'
```
paf31: Do we still need the `JSIndexer` case?

paf31:
👍 LGTM, thanks!

garyb:
You're right, we don't need the other case, and in fact there are a bunch of other things we don't need now too, as they were for the pre-0.9 release before we made breaking changes. I'll update.

garyb:
Looks like they had already mostly been sorted out actually. I moved these into what seemed a more appropriate place though, and removed the redundant cases (both in terms of inlining-via-indexer and inlining the operators rather than the named functions).

paf31:
👍 Thanks!

PR #2303, opened 2016-09-11
[psc-ide] Parse type annotations from source files

### LINKED ISSUE #2302: psc-ide should use type aliases
Using the purescript-by-example example:

```
type Entry =
  { firstName :: String
  , lastName :: String
  , address :: Address
  }

entryIdentity :: Entry -> Entry
entryIdentity entry = entry

identity entry = entryIdentity entry
```

requesting the type of `identity` gives:
```
identity :: { firstName :: String , lastName :: String , address :: { street :: String , city :: String , state :: String } } -> { firstName :: String , lastName :: String , address :: { street :: String , city :: String , state :: String } }
```
rather than `Entry -> Entry`

kRITZCREEK:
Then use these parsed type annotations to give back the non-expanded
types for functions that contain type synonyms in their annotation.

Fixes #2302 

I didn't add any tests for this yet but it's kind of trivial so I'm willing to accept that for now. The extra information about type annotations is only available on the `type` command and not the `complete` command yet, just like with definitions sites. This can definitely be changed but I'll leave that for another PR.

### DIFF
--- a/src/Language/PureScript/Ide/Externs.hs
+++ b/src/Language/PureScript/Ide/Externs.hs
@@ -17,9 +17,9 @@
 {-# LANGUAGE FlexibleContexts  #-}
 
 module Language.PureScript.Ide.Externs
-  ( readExternFile,
-    convertExterns,
-    annotateLocations
+  ( readExternFile
+  , convertExterns
+  , annotateModule
   ) where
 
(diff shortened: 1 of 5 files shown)


paf31:
Looks good to me, thanks!

Issue #514, opened 2014-07-27
Newtype Deriving


paf31:
Since the representation of newtypes is the same as the underlying type, we should be able to "derive" instances by just using the existing instance directly.

Suggested syntax:

```
newtype Trace a = Trace (Eff (trace :: Trace) a) deriving (functorEff, applyEff, applicativeEff, bindEff, monadEff)
```

Note this would allow us to apply optimizations like "magic do" to monads derived from `Eff`.

davidpelaez:
Just to confirm (since I'm a total purescript newbie), this would mean that I have to manually write the instances, e.g:

```haskell
newtype Context = Context String
instance showContext :: Show Context where
  show (Context x) = x
```

Would this be the correct approach in the meantime?

garyb:
Yep, that's right, although the "deriving equivalent" would be:
``` haskell
newtype Context = Context String
instance showContext :: Show Context where
  show (Context x) = show x
```

davidpelaez:
Thanks for the correction @garyb +1

hdgarrood:
Why not use the same syntax that Haskell does?

```
newtype Trace a = Trace (Eff (trace :: Trace) a) deriving (Functor, Apply, Applicative, Bind, Monad)
```

Fresheyeball:
:100: :100: :100:

hdgarrood:
Now that orphan instances are disallowed, this would be very useful if you are trying to use something like `purescript-quickcheck-laws`. Currently, if you had a type `X` and you wanted to check that it obeyed type class laws, you'd probably have to use a newtype wrapper (since libraries won't want to depend on quickcheck in most cases) and you'd have to write out every instance you want to check, which is a lot of boilerplate.

gbaz:
Now that we have a new syntax for deriving, can we use that for newtype deriving too?

i.e. if we say `derive instance Foo a` where Foo is anything but generic, it tries to newtype derive instead?

I'm open to tackling the issue proper if we agree on syntax first :-)

paf31:
Yes I think so. Maybe

```purescript
derive newtype instance fooBar :: Foo Bar
``` 

?

We could use `unsafeCoerce` under the hood I think, or the equivalent in the typechecker without bringing in the `unsafe-coerce` library.

garyb:
Yep, I don't think you need to, the implementation is just "do nothing" so I doubt we need to rely on that library for a dodgy typed identity function :wink:

[...]

PR #2304, opened 2016-09-11
Newtype Deriving

paf31:
This adds `newtype instance` syntax, which implements instances for newtypes by simply casting the corresponding dictionary for the type under the newtype.

### DIFF
--- a/src/Language/PureScript/AST/Traversals.hs
+++ b/src/Language/PureScript/AST/Traversals.hs
@@ -583,6 +583,6 @@ accumTypes f = everythingOnValues mappend forDecls forValues (const mempty) (con
   forDecls _ = mempty
 
   forValues (TypeClassDictionary c _ _) = mconcat (map f (constraintArgs c))
-  forValues (SuperClassDictionary _ tys) = mconcat (map f tys)
+  forValues (DeferredDictionary _ tys) = mconcat (map f tys)
   forValues (TypedValue _ _ ty) = f ty
   forValues _ = mempty
--- a/examples/failing/NewtypeInstance.purs
+++ b/examples/failing/NewtypeInstance.purs
@@ -1,0 +1,8 @@
+-- @shouldFailWith InvalidNewtypeInstance
+module Main where
+
+import Prelude
+
+data X = X
+
+derive newtype instance showX :: Show X
--- a/examples/failing/NewtypeInstance2.purs
+++ b/examples/failing/NewtypeInstance2.purs
@@ -1,0 +1,8 @@
+-- @shouldFailWith InvalidNewtypeInstance
+module Main where
+
+import Prelude
+
+data X a = X a a
+
+derive newtype instance showX :: Show a => Show (X a)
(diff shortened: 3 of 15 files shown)


paf31:
Fixed #514

garyb:
I'll review this properly shortly, but I have a minor syntax comment: I think it's a little weird to start the declaration with `newtype`. I don't think we overload "keywords" quite in that way anywhere else so it sticks out to me a bit.

`derive newtype instance` perhaps? (I assume `derive instance` is reserved for cases where an instance is wanted for the type directly, like `Generic`?)

garyb:
Actually, that distinction probably shouldn't be allowed/necessary, should it? The newtype version of a `Generic` instance wouldn't behave correctly. In that case maybe the `derive instance` syntax can just work?

paf31:
I think there needs to be some distinction, since for example, derived `Eq` and `Ord` are not the same as newtype-derived `Eq` and `Ord`.

garyb:
Well, I was thinking about that too. Their implementation differs, but are they not the same? I would assume the derived instance would be along the lines of:

``` purescript
newtype Foo = Foo String

instance eqFoo :: Eq Foo where
  eq (Foo a) (Foo b) = eq a b
```

Which when you remove the newtype unwrapping is identical?

paf31:
Oh right, that's true. I was thinking the derived instances would compare the deep structure of the two values, but that's not right.

paf31:
@garyb Updated, how's this? I'd prefer to keep the syntax for derived newtype instances separate for now, since we don't know if all future derivable instances will be the same as their newtype-derived counterparts.

src/Language/PureScript/AST/Traversals.hs
```diff
@@ -583,6 +583,6 @@ accumTypes f = everythingOnValues mappend forDecls forValues (const mempty) (con
   forDecls _ = mempty
 
   forValues (TypeClassDictionary c _ _) = mconcat (map f (constraintArgs c))
-  forValues (SuperClassDictionary _ tys) = mconcat (map f tys)
+  forValues (DeferredDictionary _ tys) = mconcat (map f tys)
```
kRITZCREEK: Isn't this just `foldMap` btw?

garyb:
> I'd prefer to keep the syntax for derived newtype instances separate for now

Fair enough. Should we explicitly disallow it for `Generic` though?

paf31:
> Should we explicitly disallow it for Generic though?

I'm not sure. The instances for `Generic` are technically invalid, but I can see a case for them. Consider something like deriving a JSON parser for example. You want the parser to work on the newtype too, but want the newtype for the distinction at the type level. This seems somehow "less wrong" than a `Generic` instance where the representations are not isomorphic, like an instance for `Date` which coerces to an integer.

I'd rather not make a special case for any particular class actually.

garyb:
My worry was that `fromSpine <<< toSpine` would fail for a newtype-derived `Generic`, but actually, it wouldn't would it? I think I was just being overly concerned.

garyb:
:shipit: :fireworks:

PR #2305, opened 2016-09-11
[psc-ide] Unify Completion Commands

kRITZCREEK:
Follow up on #2303 and #2302.

This PR merges the two different response types for the `type` and `completion` command. We now provide all the AST information for both of these.

The format for the returned completions has thus changed from:

```json
{ "module" : "Text"
, "identifier" : "Text"
, "type" : "Text"
}
```

to:
```json
{ "module" : "Text"
, "identifier" : "Text"
, "type" : "Text"
, "expandedType" : "Text"
, "definedAt" : "Maybe SourceSpan"
, "documentation" : "Maybe Text"
}
```

The `type` field now contains the annotated type, if we found it by parsing the sourcefiles. If we didn't find any annotation it still contains the type with fully expanded type synonyms. This should provide a better default for the editors and if they still want to show the expanded types that is now available via the `expandedType` field. 

The `documentation` field is always Nothing for now, but I intend to fill out that information soon, so I thought I'll change the API just once instead of twice.

Because all the old fields remain we shouldn't break any editor plugins, give better type information for the end users and allow for more features on the editor side through the new fields.

TODO:

- [x] Document the new API
- [x] Write a few tests
- [x] Squash commits

### DIFF
--- a/psc-ide-server/PROTOCOL.md
+++ b/psc-ide-server/PROTOCOL.md
@@ -51,27 +51,7 @@ definition position, if it can be found in the passed source files.
 ```
 
 **Result:**
-The possible types are returned in the same format as completions + eventual position information
-```json
-[
-  {
-  "module": "Data.Array",
-  "identifier": "filter",
-  "type": "forall a. (a -> Boolean) -> Array a -> Array a"
-  },
-  {
-  "module": "Data.Array",
-  "identifier": "filter",
-  "type": "forall a. (a -> Boolean) -> Array a -> Array a",
-  "definedAt":
-    {
-    "name": "/path/to/file",
-    "start": [1, 3],
-    "end": [3, 1]
-    }
-  }
-]
-```
+The possible types are returned in the same format as completions
 
 ### Complete
 The `complete` command looks up possible completions/corrections.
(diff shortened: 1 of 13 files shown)


kRITZCREEK:
Ready for review. Once it's good to go I'll squash things.

psc-ide-server/PROTOCOL.md
```diff
@@ -104,12 +84,23 @@ The `complete` command looks up possible completions/corrections.
 
 The following format is returned as the Result:
 
+Bpth the `definedAt` aswell as the `documentation` field might be `null` if they
```
paf31: Typo ("Bpth"))

tests/Language/PureScript/Ide/SourceFile/IntegrationSpec.hs
```diff
@@ -31,6 +31,11 @@ spec = beforeAll_ setup $
...
 testCase :: Text -> (Int, Int) -> IO ()
 testCase s (x, y) = do
-  P.SourceSpan f (P.SourcePos l c) _ : _ <- Integration.getInfo s
+  P.SourceSpan f (P.SourcePos l c) _ <- getLocation s
   toS f `shouldSatisfy` T.isSuffixOf "SourceFileSpec.purs"
   (l, c) `shouldBe` (x, y)
+
+getLocation :: Text -> IO P.SourceSpan
+getLocation s = do
+  (_, _, _, Just location) : _ <- Integration.getType s
```
paf31: Can this match fail?

paf31:
Looks great!

tests/Language/PureScript/Ide/SourceFile/IntegrationSpec.hs
kRITZCREEK: Sure, but in that case I want the test to fail anyway. The pattern match failure should point me at this line and I'll know what's up.

psc-ide-server/PROTOCOL.md
kRITZCREEK: Thanks!

kRITZCREEK:
Thanks! Squashed.

Issue #2222, opened 2016-07-10
[psc-ide] Near future roadmap


kRITZCREEK:
Features/ that I'm working on/that are planned in the near future for psc-ide:

- [ ] docstrings for identifiers
- [ ] show-usages
- [ ] show instances for typeclasses
- [x] resolve types/kinds for operators

If you've got suggestions or critique feel free to comment on this issue.

rvion:
Would it make sense for psc / psc-ide to implement a similar mechanism as -fdefer-type-errors in haskell for non parsing-related issues ? It would basically tells the compiler to keep typechecking expressions in the presence of errors, if those expressions are not depending of problematic expressions.

As of now, we can't have those 2 errors in the same time:

A|B
|----|----|
| ![image](https://cloud.githubusercontent.com/assets/2150990/16741010/c5a27b32-47a1-11e6-80c1-9375c998298b.png) |  ![image](https://cloud.githubusercontent.com/assets/2150990/16741025/d68aed76-47a1-11e6-8f92-7163055b9de6.png) |

with such a mechanism, it would be possible for psc-ide to report both.

paf31:
@rvion That would need changes in the compiler itself, not just `psc-ide`. I think it would be better tracked in its own issue.

texastoland:
#1868

rvion:
i have one more suggestion, because I really miss one powerfull and __producti__ feature from my Haskell setup. (both ghc-mod and intero suport it)

-------------

When selecting some code, I wish I could send my selection (startLine, startColumn, endLine, endColumn) to psc-ide, and get back the  __span__ (startLine, startColumn, endLine, endColumn) __and the type of the closest enclosing expression__ in real time

![image](https://cloud.githubusercontent.com/assets/2150990/15269839/0efffc1e-1a0c-11e6-810b-133e56baf617.png)
here, I select `" "`, and ghc-mod detect that the smallest enclosing expression is `concat3 "aa" "bb"`that have type `String -> String`. (In light grey is my selection, in dark blue-ish: the `smallest enclosing expression`)

👍  knowing the smallest enclosing expression is needed so ide plugins can add refactoring features such as `refactor expression as a function`

[...]

PR #2307, opened 2016-09-16
[psc-ide] resolve types/kinds for operators

kRITZCREEK:
One point of #2222.

Things left to do:

- [x] Speed up the pass in `getAllModules`
- [x] Write tests
- [x] discuss how we want to preserve the fixity/alias information in the protocol

Where operators used the type field to pass their fixity information to the editors that field is now used for types/kinds again.

Before:
```purescript
Control.Apply.$> :: 
  infixl 4 Data.Functor.voidLeft as $>
```

Now:
````purescript
Control.Apply.$> :: 
  forall f a b. (Functor f) => f a -> b -> f b
```

The server still has that information but because of the generic completion format there is no proper way to communicate it to the clients. Maybe we should think about a sum type (with a "tag" field in the json) to return completions. I like the simplicity of the current approach and would want to preserve generic fields for `module` `type` and `identifier` on every completion to make the integration with a new editor or other tools as simple as possible. Maybe we could add a `details` field that has information specific to the type (Value, Type, Constructor, ...) of completion.

Thoughts?
/cc @nwolverson @FrigoEU @rvion @bsermons

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -521,5 +521,6 @@ test-suite tests
                    Language.PureScript.Ide.ReexportsSpec
                    Language.PureScript.Ide.SourceFile.IntegrationSpec
                    Language.PureScript.Ide.SourceFileSpec
+                   Language.PureScript.Ide.StateSpec
     buildable: True
     hs-source-dirs: tests
--- a/src/Language/PureScript/Ide/Externs.hs
+++ b/src/Language/PureScript/Ide/Externs.hs
@@ -87,17 +87,19 @@ convertOperator :: P.ExternsFixity -> IdeDeclaration
 convertOperator P.ExternsFixity{..} =
   IdeValueOperator
     efOperator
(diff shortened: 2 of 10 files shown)


rvion:
> Maybe we could add a details field that has information specific to the type (Value, Type, Constructor, ...) of completion.

👍

paf31:
👍 Looks great. Is this ready to merge?

nwolverson:
No particular thoughts, but IMO key point of supporting backwards compatability in the IDE protocol is allowing handling multiple versions without specific behaviour according to version number, i.e. adding new optional fields is good.

kRITZCREEK:
Squashed and ready to merge.

I'll do a little writeup of the completion-type specific information we can access within psc-ide in another issue/psc-ide's roadmap issue.

paf31:
Thanks!

PR #2313, opened 2016-09-18
Relax the one-way unification check for repeated variables

paf31:
@garyb @kRITZCREEK 

This is the one last thing I'd like to decide before merging the fundeps branch. I think it simplifies the rules quite a bit, but at the expense of some possibly unexpected inferred types (compared to Haskell) in a few more complicated cases.

Hopefully the test changes summarize the difference, but the new rule is:

> If a type variable appears more than once in an instance head, we will commit to that instance if all occurences of that type variable will be instantiated to things which can be unified, *including unknowns*.

This rule allows things like this to work, where they wouldn't have before (in the `fundeps` branch):

```haskell
> import Prelude
> import Control.Monad.Error.Class 
> import Control.Monad.State.Class
> import Control.Monad.State
> import Control.Monad.Except

> :t get >>= throwError
forall t1 t2 t6.    
  ( Bind t1         
  , MonadState t6 t1
  , MonadError t6 t1
  ) => t1 t2        

> :t runExceptT (get >>= throwError)
forall t1 t2 t3.        
  ( Monad t2            
  , MonadState t3 t2    
  ) => t2 (Either t3 t1)

> :t runState (runExceptT (get >>= throwError))
forall t5 t7. t7 -> Tuple (Either t7 t5) t7

> :t runState (runExceptT (get >>= throwError)) 0 -- This one used to fail for a confusing reason
forall t3. Tuple (Either Int t3) Int
```

However, it means that we now might pick an instance where you might not expect us to:

```haskell
> class EQ a b
> instance eq :: EQ a a
> let test = (\x y -> x) :: forall a b. EQ a b => a -> b -> a

> :t test 'a'
Char -> Char
```

Here, Haskell would infer `EQ Char a => a -> Char`, although I'm not really sure why it infers this but allows some of the things listed above.

My rationale for allowing this is that any case where this would lead to unexpected inferred types seems to require possibly overlapping instances anyway (like `EQ` above). We warn on overlapping instances, so if you use them, you should expect odd things to happen.

I've put this in a separate branch since I'd like to discuss it in isolation before we commit to it.

It's reassuring, having gone through the literature, that the Haskell folks had to think about these issues quite hard some twenty years ago. For reference, a lot of these sorts of things are documented in "Type classes: an exploration of the design space".

### DIFF
--- a/examples/failing/1310.purs
+++ b/examples/failing/1310.purs
@@ -1,4 +1,4 @@
--- @shouldFailWith NoInstanceFound
+-- @shouldFailWith TypesDoNotUnify
 
 module Issue1310 where
 
--- a/examples/passing/UnifyInTypeInstanceLookup.purs
+++ b/examples/passing/UnifyInTypeInstanceLookup.purs
@@ -2,9 +2,6 @@ module Main where
 
 import Control.Monad.Eff.Console (log)
 
-data Z = Z
-data S n = S n
-
 data T
 data F
 
@@ -15,11 +12,11 @@ instance eqF :: EQ x y F
 test :: forall a b. (EQ a b T) => a -> b -> a
 test a _ = a
 
-spin :: forall a b. a -> b
-spin a = spin a
-
--- Expected type: 
--- forall t. (EQ t (S Z) T) => t
-test1 = test (spin 1) (S Z)
+-- Note the expected type: 
+--   Char -> Char 
+-- Not 
+--   forall t. (EQ t Char T) => t -> t
+-- as in Haskell
+test1 a = test a 'x'
 
 main = log "Done"
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -280,9 +280,7 @@ matches deps TypeClassDictionaryInScope{..} tys = do
     -- Now, use any functional dependencies to infer any remaining types
     guard $ covers matched
     -- Verify that any repeated type variables are unifiable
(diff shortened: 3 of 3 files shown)


kRITZCREEK:
I can't really judge what this means for the applications of functional dependencies as I have not ever used them myself so far. What I can tell is that this is simpler (more predictable?) behaviour and the examples/tests look like it's doing the right thing. One thing that might be nice is if we could add an error hint if we run into unification errors after choosing an instance that shows the instances we chose.

garyb:
Does this have implications for `Inject` again? I'm not too worried about it if so, I guess, since at the moment I consider it unusable anyway. When/if we get something like instance chains I don't think this would be a problem anyway, since the `injectRefl` instance would be at the end of the chain.

garyb:
(I'll experiment with this and review it tomorrow, off to bed now)

paf31:
Yes, the `Inject a a` instance will now match any pair of unifiable types. This is not quite the same as the version in the original PR I made, where that instance would always be chosen. Now, the two types must be unifiable before we commit to that instance, but they can also contain unification variables.

garyb:
Ah, ok! So if you tried `x = inj 1` it would probably be typed as `x :: Int` then, but also wouldn't obstruct the likes of `x = inj 1 :: Either String Int`? That sounds perfectly reasonable to me.

paf31:
Right. So there's something not quite right going on with respect to principal typing, but I'm not too worried since these cases should raise overlapping instance warnings anyway.

paf31:
I'm now convinced that the other branch can do better than this approach. Closing ...

PR #2317, opened 2016-09-20
[psc-ide] Return qualifier from explicit/hiding imports

nwolverson:


src/Language/PureScript/Ide/Types.hs
```diff
@@ -149,16 +149,16 @@ instance ToJSON ModuleImport where
     object $ [ "module" .= mn
              , "importType" .= ("implicit" :: Text)
              ] ++ fmap (\x -> "qualifier" .= x) (maybeToList qualifier)
-  toJSON (ModuleImport mn (P.Explicit refs) _) =
-    object [ "module" .= mn
+  toJSON (ModuleImport mn (P.Explicit refs) qualifier) =
+    object $ [ "module" .= mn
            , "importType" .= ("explicit" :: Text)
```
kRITZCREEK: Keep the alignment of `[` and `,` please

src/Language/PureScript/Ide/Types.hs
```diff
@@ -149,16 +149,16 @@ instance ToJSON ModuleImport where
...
            , "identifiers" .= (identifierFromDeclarationRef <$> refs)
-           ]
-  toJSON (ModuleImport mn (P.Hiding refs) _) =
-    object [ "module" .= mn
+           ] ++ fmap (\x -> "qualifier" .= x) (maybeToList qualifier)
```
kRITZCREEK: The `Protolude` is imported here so just `map` is actually the `Functor` version.

src/Language/PureScript/Ide/Types.hs
```diff
@@ -149,16 +149,16 @@ instance ToJSON ModuleImport where
...
+  toJSON (ModuleImport mn (P.Hiding refs) qualifier) =
+    object $ [ "module" .= mn
            , "importType" .= ("hiding" :: Text)
            , "identifiers" .= (identifierFromDeclarationRef <$> refs)
-           ]
+           ] ++ fmap (\x -> "qualifier" .= x) (maybeToList qualifier)
```
kRITZCREEK: Same two comments as above

kRITZCREEK:
Oh yeah and it would be nice if you could add the two fields to `psc-ide-server/PROTOCOL.md`, I'm trying to keep it up-to-date.

src/Language/PureScript/Ide/Types.hs
nwolverson: Hmm, this was ambiguous due to the non-qualified import of `Data.Map.Lazy`. Had no idea Haskell had both `import qualified ... as ...` and `import ... as ...` forms...

kRITZCREEK: Haskell's namespacing is way too complex...

PR #2318, opened 2016-09-20
[psc-ide] Lens experiments

kRITZCREEK:
I wanted to see if using Lenses and Prisms for IDE declaration types improve the code and I'm not really sure about the result, but I put in some work and I think it's an improvement if only slight.

It definitely makes playing around in the repl more convenient and I'm still very much a Lens beginner so I might have misused or missed lens combinators aswell.

Just thought I'd put it out there and check if someone is feeling generous and takes a look :)

Things that definitely would still need to be done:

- [x] proper bounds for lens dependency
- [x] run the import sections through stylish-haskell

### DIFF
--- a/src/Language/PureScript/Ide/Types.hs
+++ b/src/Language/PureScript/Ide/Types.hs
@@ -12,41 +12,92 @@
 -- Type definitions for psc-ide
 -----------------------------------------------------------------------------
 
-{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE DeriveFoldable    #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
 
 module Language.PureScript.Ide.Types where
 
 import           Protolude
 
 import           Control.Concurrent.STM
+import           Control.Lens.TH
 import           Data.Aeson
-import qualified Data.Map.Lazy                        as M
-import qualified Language.PureScript.Errors.JSON      as P
-import qualified Language.PureScript as P
+import qualified Data.Map.Lazy                       as M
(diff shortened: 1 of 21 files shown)


src/Language/PureScript/Ide/Types.hs
```diff
@@ -31,17 +33,64 @@ import           Text.Parsec.Text
...
+makeLenses ''IdeValue
+makeLenses ''IdeType
+makeLenses ''IdeSynonym
+makeLenses ''IdeDataConstructor
+makeLenses ''IdeValueOperator
+makeLenses ''IdeTypeOperator
+
+data IdeDeclarationAnn = IdeDeclarationAnn
+  { _idaAnnotation :: Annotation
+  , _idaDeclaration :: IdeDeclaration
```
LiamGoodacre: Did you mean `ida` or `ide`?

kRITZCREEK: **I**de**D**eclaration**A**nn... It's a terrible name in the first place but it would be way too long for an accessor. I tried to make it shorter...

LiamGoodacre: Ah I see, just looked like a typo because they're so similar 😄 - especially because of `_idaDeclaration :: IdeDeclaration`.

kRITZCREEK: Oh yeah now I see how that looks confusing^^ I'd love to drop all these prefixes and use the new OverloadedRecord thingy but I've heard it produces horrible error messages and we're not all GHC 8 yet.

kRITZCREEK:
@paf31 gave his OK so I'll go ahead and squash-merge this.

PR #2319, opened 2016-09-21
Fix function application judgment

paf31:
I came across this while testing my `fundeps` branch. This would fail previously:

```
$ pulp psci
> import Control.Monad.Reader.Class
> ask 0
```

Now it correctly returns `0`.

### DIFF
--- a/examples/failing/DoNotSuggestComposition2.purs
+++ b/examples/failing/DoNotSuggestComposition2.purs
@@ -1,4 +1,4 @@
--- @shouldFailWith CannotApplyFunction
+-- @shouldFailWith TypesDoNotUnify
 -- TODO: Check that this does not produce a "function composition is (<<<)"
 -- suggestion.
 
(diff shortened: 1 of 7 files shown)


paf31:
@kRITZCREEK Could you please review this one?

kRITZCREEK:
Looks good to me! I'm curious why we only tried to unify with unknowns before? And it would be really helpful if the function had a few haddocks that described what the different arguments are. :+1:

paf31:
> I'm curious why we only tried to unify with unknowns before?

Basically just an error on my part, I think.

> if the function had a few haddocks 

Sounds good, will do!

paf31:
@kRITZCREEK Updated, and I've also moved things around a little bit so that it's hopefully easier to follow.

kRITZCREEK:
Thanks for the documentation and refactoring! Was a lot easier to understand now.

PR #2324, opened 2016-09-22
[psc-ide] Update pursuit JSON parsing

nwolverson:
Per https://github.com/purescript/pursuit/pull/246 - this was not working
since JSON format changed. Now no need to parse out type, and can provide
additional summary text.


paf31:
👍 LGTM, thanks!

PR #2326, opened 2016-09-23
Add deriving for Data.Newtype

### LINKED ISSUE #2314: Add `Newtype` class deriving
@paf31 any objections to this? I can do the work of making the instance derivable, etc.

garyb:
Resolves #2314


src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -404,6 +421,47 @@ deriveOrd mn ds tyConNm = do
...
+  checkIsWildcard unwrappedTy
+  go =<< findTypeDecl tyConNm ds
+  where
+
+  go :: Declaration -> m ([Declaration], Type)
+  go (DataDeclaration Data name _ _) =
+    throwError . errorMessage $ CannotDeriveNewtypeForData name
+  go (DataDeclaration Newtype name _ dctors) = do
+    checkNewtype name dctors
+    let (ctorName, [ty]) = head dctors
```
paf31: Why not match this as a singleton above?

tests/support/bower.json
```diff
@@ -5,8 +5,9 @@
     "purescript-console": "1.0.0-rc.1",
     "purescript-eff": "1.0.0-rc.1",
     "purescript-functions": "1.0.0-rc.1",
-    "purescript-prelude": "1.0.0-rc.3",
+    "purescript-prelude": "1.1.0",
     "purescript-st": "1.0.0-rc.1",
-    "purescript-partial": "1.1.2"
+    "purescript-partial": "1.1.2",
+    "purescript-newtype": "0.1.0"
```
paf31: I assume this will be bumped to 1.0 eventually?

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -404,6 +421,47 @@ deriveOrd mn ds tyConNm = do
...
+deriveNewtype mn ds tyConNm unwrappedTy = do
+  checkIsWildcard unwrappedTy
+  go =<< findTypeDecl tyConNm ds
+  where
+
+  go :: Declaration -> m ([Declaration], Type)
+  go (DataDeclaration Data name _ _) =
+    throwError . errorMessage $ CannotDeriveNewtypeForData name
+  go (DataDeclaration Newtype name _ dctors) = do
+    checkNewtype name dctors
```
paf31: This will be caught at the newtype definition, won't it?

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
garyb: Could do, but if you have an invalid `newtype` you'll get an error about an un-derivable class before you get the error that the `newtype` is invalid (since the check is made during type checking), which seems... not very useful.

paf31:
👍 Looks good, thanks!

I keep wondering about implementing GHC-style Generics as well, but that's probably too much work for 0.10. Hopefully we can add it in a minor release as a non-breaking change.

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
paf31: Ah ok. We can revisit this after I move deriving into the solver then.

tests/support/bower.json
garyb: Yeah, I just made a release so we could add the test thing - we'll need to make a 2.0 prelude anyway, so when that is released the dependency will need bumping in `newtype`, at which point it'll be worth releasing as 1.0... phew :wink:

paf31: Why do we need Prelude 2.0?

garyb:
Tests are passing again now - looks like one of them was depending on a misbehaving `Show` instance for `Number` that was fixed some time after the `Prelude` dependency we were using previously.

paf31:
Looks good!

PR #2327, opened 2016-09-24
Fix the duplicate/redefined module error

### LINKED ISSUE #1828: DuplicateModuleName error does not specify what file Modules are defined in
I split up my file into multiple files and missed a `module Parser where` somewhere so I get

     Module Parser has been defined multiple times.

It would be good if this said what files the modules are defined in.

I had a quick look and `RedefinedModule` does seem to specify filenames but `DuplicateModuleName` does not. I don't know why my example triggers the latter (using `pulp build`) and I can't see a way to get at filenames where `DuplicateModule` is called.

Happy to have another look at fixing this with some guidance.

garyb:
Resolves #1828

### DIFF
--- a/src/Language/PureScript/Sugar/Names.hs
+++ b/src/Language/PureScript/Sugar/Names.hs
@@ -99,15 +99,12 @@ desugarImportsWithEnv externs modules = do
     exportedRefs f = M.fromList $ (, efModuleName) <$> mapMaybe f efExports
 
   updateEnv :: ([Module], Env) -> Module -> m ([Module], Env)
-  updateEnv (ms, env) m@(Module ss _ mn _ refs) =
-    case mn `M.lookup` env of
-      Just m' -> throwError . errorMessage $ RedefinedModule mn [envModuleSourceSpan m', ss]
-      Nothing -> do
-        members <- findExportable m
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/Sugar/Names.hs
```diff
@@ -99,15 +99,12 @@ desugarImportsWithEnv externs modules = do
     exportedRefs f = M.fromList $ (, efModuleName) <$> mapMaybe f efExports
 
   updateEnv :: ([Module], Env) -> Module -> m ([Module], Env)
-  updateEnv (ms, env) m@(Module ss _ mn _ refs) =
-    case mn `M.lookup` env of
-      Just m' -> throwError . errorMessage $ RedefinedModule mn [envModuleSourceSpan m', ss]
```
garyb: No need to check if a module already exists in the name environment anymore since it will be caught during `Make` and the error raised there.

garyb:
Any ideas on why that test would fail on some platforms/configs? 😕

garyb:
Hmm, weird, still failing.

garyb:
Ok, we're good now, it was due to missing files in the cabal file. The coverage build is failing due to a dependency issue though... the same is true of `master`.

paf31:
👍

PR #2328, opened 2016-09-24
Fix handling of duplicate module imports in JS codegen

### LINKED ISSUE #2245: Duplicate names of generated module imports
When importing the same module more than once I would expect one of two things to happen in the generated JavaScript:

1. They each get different names
2. They collapse into one import

What actually happens is that they all declare the same variable name:

---

One import of Prelude:

```purescript
import Prelude as P0
```
```javascript
// generated js
var Prelude = require("../Prelude");
```

Two imports of Prelude:

```purescript
import Prelude as P0
import Prelude as P1
```
```javascript
// generated js
var Prelude_1 = require("../Prelude");
var Prelude_1 = require("../Prelude");
```

Three imports of Prelude:

```purescript
import Prelude as P0
import Prelude as P1
import Prelude as P2
```
```javascript
// generated js
var Prelude_2 = require("../Prelude");
var Prelude_2 = require("../Prelude");
var Prelude_2 = require("../Prelude");
```

etc...

garyb:
Resolves #2245

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -15,7 +15,7 @@ import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Reader (MonadReader, asks)
 import Control.Monad.Supply.Class
 
-import Data.List ((\\), delete, intersect)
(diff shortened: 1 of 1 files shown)


paf31:
👍 LGTM

PR #2329, opened 2016-09-24
Fix usage detection for operators

### LINKED ISSUE #2269: No unused warning is shown for operators

### LINKED ISSUE #2270: Incorrect message about redundant import when operators are used and values are not
The fix for #2269 might handle this too, but it's kinda a separate thing.

I had something like this:

``` purescript
import Data.Leibniz (type (~), coerce, symm)
```

and was using `(~)` but not `coerce` or `symm`, and the warning told me the entire import was redundant.

garyb:
Resolves #2269, resolves #2270.

Operators were being ignored entirely in the "unused names" calculation, so the linked issues aren't exactly surprising! It's something that slipped through the cracks when the `Ident` and `Op` namespaces were separated.

### DIFF
--- a/examples/warning/UnusedExplicitImportTypeOp.purs
+++ b/examples/warning/UnusedExplicitImportTypeOp.purs
@@ -1,0 +1,9 @@
+-- @shouldWarnWith UnusedExplicitImport
+module Main where
+
+import Prelude (Unit, unit, pure)
+import Control.Monad.Eff (Eff)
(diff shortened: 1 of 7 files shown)


garyb:
Weird, maybe the test failures I'm getting on some of these are to do with the caches or something?

paf31:
👍 LGTM. Maybe clear the Travis caches and restart the build?

PR #2334, opened 2016-09-29
Fix #2332, allow newtype deriving when newtype is not fully applied

### LINKED ISSUE #2332: `newtype instance` deriving failing for `Functor`
As reported here: http://ircbrowse.net/browse/purescript?id=404681&timestamp=1474868573#t1474868573

I assume this applies to deriving of any non-`*` kinded things.

paf31:


garyb:
🎉

PR #2335, opened 2016-09-29
Fix #2331, substutute types in newtype class deriving

### LINKED ISSUE #2331: `Newtype` deriving with typevars needs work
Unfortunately my suspicion with this was right:

``` purescript
newtype First a = First a
derive instance newtypeFirst :: Newtype (First b) _
```

````
  Could not match type

    a0

  with type

    b1
```

paf31:
Sorry that the indentation makes this hard to read. The only real change is in the line with

```haskell
replaceAllTypeVars subst ty
```

### DIFF
--- a/examples/passing/DeriveNewtype.purs
+++ b/examples/passing/DeriveNewtype.purs
@@ -14,4 +14,14 @@ t = wrap "hello"
 a :: String
 a = unwrap t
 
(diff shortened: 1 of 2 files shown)


garyb:
Thanks! I had intended to fix this at the weekend. :smile:

PR #2337, opened 2016-09-29
Add psc-package

paf31:
You can use`psc-package` like this:

``` sh
> psc-package init
> # ... will install prelude
> # Now edit psc-package.json to add a new dependency, or use install:
> psc-package install transformers
> psc-package build
```
- [x] Support multiple package sets (these can correspond to branches or tags in the package set repo)
- [x] Make the package set repo URL configurable
- [x] Add a command which can dump out the JSON which can go into a pull request on the package set repo
- [x] Better error mesages
- [x] Add a command to list out the source folders of all active dependencies, so that we can easily pass them to `psci` or run tests, etc.

Bugs: 
- [x] Updating should delete removed repos
- [x] Don't add package to JSON file in `install` if it doesn't install correctly

Later:
- Globally cache cloned package repos
- Support a list of additional dependencies or overrides in the form of Git repos and SHA hashes
- Support automatically updating to minor/patch releases without having to update the package set.
- Make `psc-package init` interactive?


paf31:
Maybe we should come up with a list of features to turn this into a minimally viable package manager. Here are some I would like:

- [ ] Support a list of additional dependencies or overrides in the form of Git repos and SHA hashes
- [ ] Support multiple package sets (these can correspond to branches or tags in the package set repo)
- [ ] Make the package set repo URL configurable
- [ ] Some sort of `upload` command, or at least something which can dump out the JSON which can go into a pull request on the package set repo.

paf31:
Here is a little demo:

https://asciinema.org/a/bcun93ys42gacwqul5jcmmwkq

paf31:
I've made a little bit of progress here. The source repo and branch are now configurable in the project JSON file, and different package sets can be used side-by-side since they are managed under different parts of the working directory.

paf31:
I think this is working well enough that it can be reviewed and we can have people start to test it out.

There are still features which we'll need to add, but it should be possible to get up and running with a new project pretty quickly this way.

The important thing to remember here is that this is just another building block, like `psc`, and it will need tooling support eventually if we decide to commit to it, probably by changing tools like Pulp. So I've added very basic commands to list out relevant bits of metadata, like the source directories for the active packages, or the complete list of transitive dependencies.

If we do commit to this approach, then there will be quite a bit of work to do afterwards to get the library developer workflow (pulp publish etc.) up-to-date. Ideally, publishing would involve updating the package set repo by PR, and documentation for the package set would be regenerated by the CI job. But we'll get there eventually...

paf31:
I wonder if we should serve the package set JSON file over HTTP instead of Git. That would allow us to move to a simpler package publishing model later, which could probably integrate more easily with `pulp publish`. Of course, we could still keep a copy of the package set JSON in Git.

Any thoughts?

paf31:
psc-package now uses HTTP(S) to download the package set JSON file, which should allow us to support other APIs later if necessary.

This is ready for review again.

paf31:
I've reverted to using Git for pulling the package set repo, after @anttih experienced some issues. We can try HTTP again later if we need it.

I've been using this successfully for a few weeks now, with no major issues. Has anyone else had a chance to try it?

psc-package/Main.hs
```diff
@@ -0,0 +1,259 @@
...
+import qualified Data.Map as Map
+import           Data.Maybe (mapMaybe)
+import qualified Data.Set as Set
+import           Data.Text (pack)
+import qualified Data.Text.Lazy as TL
+import qualified Data.Text.Lazy.Builder as TB
+import           Data.Text.Encoding (encodeUtf8)
+import           Data.Traversable (for)
+import           Data.Version (showVersion)
+import qualified Filesystem.Path.CurrentOS as Path
```
hdgarrood: Ah, ok, seems like turtle is sticking with `system-filepath` until the abstract filepath proposal gets implemented: https://github.com/Gabriel439/Haskell-Turtle-Library/issues/54. Ok, in that case, I guess we should leave this as-is for now.

psc-package/Main.hs
```diff
@@ -0,0 +1,259 @@
...
+  } deriving (Show, Generic, Aeson.FromJSON, Aeson.ToJSON)
+
+pathToTextUnsafe :: Turtle.FilePath -> Text
+pathToTextUnsafe = either (error "Path.toText failed") id . Path.toText
+
+defaultPackage :: Text -> PackageConfig
+defaultPackage pkgName =
+  PackageConfig { name    = pkgName
+                , depends = [ "prelude" ]
+                , set     = "psc-" <> pack (showVersion Paths.version)
```
hdgarrood: As far as I can tell, this is the only part of `psc-package` that currently uses the `purescript` library. Distributing `psc-package` like this as a part of the compiler suite does seem like a good way of getting package sets to match up with the installed compiler without needing people to write it in, and has the additional advantage that it wouldn't make the installation instructions any longer, but have we considered whether it would make sense to develop `psc-package` as a separate tool i.e. outside this repository?

psc-package/Main.hs
```diff
@@ -0,0 +1,259 @@
...
+import qualified System.IO as IO
+import           Turtle hiding (fold)
+
+packageFile :: Path.FilePath
+packageFile = "psc-package.json"
+
+data PackageConfig = PackageConfig
+  { name    :: Text
+  , depends :: [Text]
+  , set     :: Text
```
hdgarrood: Are package sets defined in the `package-sets` repository?

paf31: Yes, and I'm hoping that `package-sets` will be able to use `psc-package` for its sanity checks soon after this gets merged, and we can just add the validation commands and any other useful tools for curators directly into `psc-package`.

psc-package/Main.hs
paf31: The reasons I'd like to keep them together:

- I'd like the default install to have some sensible packaging solution without having to use NPM or another package manager to install additional tools.
- Eventually I'd like to think about compiler integrations like checking minor version bumps are really non-breaking.
- I'd like to get to the point where we have one executable before 1.0, and then this would be run as `psc package <command>`.

psc-package/Main.hs
```diff
@@ -0,0 +1,259 @@
...
+readPackageFile :: IO PackageConfig
+readPackageFile = do
+  exists <- testfile packageFile
+  unless exists $ do
+    echo "psc-package.json does not exist"
+    exit (ExitFailure 1)
+  mpkg <- Aeson.decodeStrict . encodeUtf8 <$> readTextFile packageFile
+  case mpkg of
+    Nothing -> do
+      echo "Unable to parse psc-package.json"
```
paf31: I think I like the idea of using YAML too. Although I would like it if developers didn't have to edit by hand for most tasks anyway.

psc-package/Main.hs
hdgarrood: Agreed, that all sounds sensible.

hdgarrood:
Oh also I forgot to add, the code all looks good to me :)

paf31:
Thanks for the review!

psc-package/Main.hs
paf31: I do like the idea of keeping `psc-package` minimal though, and scriptable by tools like Pulp, which is why I included the `sources` and `dependencies` commands (they seems like the minimal set of things needed to make `psc-package` usable by other tools).

bodil:
Please, no YAML, it's a mess. TOML would be OK.

LiamGoodacre:
From TOML readme: https://github.com/toml-lang/toml/blame/master/README.md#L11

```
Be warned, this spec is still changing a lot. Until it's marked as 1.0, you
should assume that it is unstable and act accordingly.
```

bodil:
I'm not going to advocate for TOML, beyond saying that if Cargo is fine with it, I am too. I'm also perfectly fine with staying with JSON. Just, whatever we do, please, no YAML.

hdgarrood:
Yeah, the fact that Cargo uses it was what made me not worry about that instability notice in the readme. Also, TOML v0.4.0 was released almost 2 years ago, and is still the most recent version.

paf31:
Let's just go with JSON for now then.

PR #2338, opened 2016-09-29
Strict handling of UTF8 files (fixes fd issue)

chrisdone:
I found that on OS X when trying to build the basic purescript-react example I got this error:
```
psc: /Users/chris/Projects/psc/purescript-dom/src//DOM/HTML/Location.js: openFile: resource exhausted (Too many open files)
```

This was fixed in https://github.com/purescript/purescript/pull/1674 for psc-bundle, but actually my error comes from `psc` itself, because it uses `readUTF8File` in various places and the fix was placed only in psc-bundle. Instead, I've fixed the problem at the source by using the already-present utf8-string package to simply decode the strict `ByteString` into a `String`. This is also what `hindent` does.

This fixes the issue on my OS X.


paf31:
Looks good, thanks! Could you please update `CONTRIBUTORS.md` and I'll get this merged for 0.10.1?

garyb:
> I'll get this merged for 0.10.1

Isn't it just going to be 0.10? We've not released a non-RC yet...

paf31:
Last time we used 0.9.1 since people had been using the 0.9 release candidate for a while, but I don't mind.

chrisdone:
Done!

paf31:
I'm not sure why GitHub thinks there is a conflict here.

chrisdone:
I rebased. =)

PR #2352, opened 2016-10-02
Type directed search

kRITZCREEK:
This is what I've been working on for a while now and the subject of my thesis. It's an extension to typed holes that uses the typechecker to find definitions that subsume the type inferred for the typed hole.

To put it in another way it searches possible substitutions for the typed hole. Eg:

```purescript
> import Prelude
> let y = 1
> 1 + ?x
Error found:
in module $PSCI
at  line 1, column 1 - line 1, column 5

  Hole 'x' has the inferred type

    Int

    $PSCI.it             :: Int
    $PSCI.y              :: Int
    Data.Bounded.bottom  :: forall a. (Bounded a) => a
    Data.Bounded.top     :: forall a. (Bounded a) => a
    Data.Semiring.one    :: forall a. (Semiring a) => a
    Data.Semiring.zero   :: forall a. (Semiring a) => a

```
A trickier example:

```purescript
> import Data.Maybe
> import Data.Traversable
> ?x :: (Int -> Maybe String) -> Array Int -> Maybe (Array String)
Error found:
in module $PSCI
at  line 1, column 1 - line 1, column 64

  Hole 'x' has the inferred type

    (Int -> Maybe String) -> Array Int -> Maybe (Array String)

    $PSCI.it                          :: (Int -> Maybe String) -> Array Int -> Maybe (Array String)
    Data.Monoid.mempty                :: forall m. (Monoid m) => m
    Data.Traversable.traverse         :: forall a b m t.
                                           ( Traversable t
                                           , Applicative m
                                           ) => (a -> m b) -> t a -> m (t b)
    Data.Traversable.traverseDefault  :: forall t a b m.
                                           ( Traversable t
                                           , Applicative m
                                           ) => (a -> m b) -> t a -> m (t b)
    Unsafe.Coerce.unsafeCoerce        :: forall a b. a -> b

```

This is nowhere close to being polished enough for merging but I'd like to ask you to give it a spin and see what you think. Maybe you could report back if you find some strange artifacts or how this could be made more useful.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -29,7 +29,7 @@ import qualified Text.Parsec as P
 -- | A map of locally-bound names in scope.
 type Context = [(Ident, Type)]
 
-newtype TypeSearch = TypeSearch (Type -> [Qualified Ident])
+newtype TypeSearch = TypeSearch (Type -> M.Map (Qualified Ident) (Type, NameKind, NameVisibility))
 
 instance Show TypeSearch where
   show _ = "TypeSearch"
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -29,10 +29,8 @@ import qualified Text.Parsec as P
 -- | A map of locally-bound names in scope.
 type Context = [(Ident, Type)]
 
-newtype TypeSearch = TypeSearch (Type -> M.Map (Qualified Ident) (Type, NameKind, NameVisibility))
-
-instance Show TypeSearch where
-  show _ = "TypeSearch"
+data TypeSearch = TSBefore Environment | TSAfter [(Qualified Ident, Type)]
+  deriving Show
 
 -- | A type of error messages
 data SimpleErrorMessage
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -29,6 +29,11 @@ import qualified Text.Parsec as P
 -- | A map of locally-bound names in scope.
 type Context = [(Ident, Type)]
 
+newtype TypeSearch = TypeSearch (Type -> [Qualified Ident])
+
+instance Show TypeSearch where
+  show _ = "TypeSearch"
+
 -- | A type of error messages
 data SimpleErrorMessage
   = ErrorParsingFFIModule FilePath (Maybe Bundle.ErrorMessage)
@@ -99,7 +104,7 @@ data SimpleErrorMessage
   | ShadowedTypeVar String
   | UnusedTypeVar String
   | WildcardInferredType Type Context
-  | HoleInferredType String Type Context
+  | HoleInferredType String Type Context TypeSearch
   | MissingTypeDeclaration Ident Type
   | OverlappingPattern [[Binder]] Bool
   | IncompleteExhaustivityCheck
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -29,7 +29,13 @@ import qualified Text.Parsec as P
 -- | A map of locally-bound names in scope.
 type Context = [(Ident, Type)]
 
-data TypeSearch = TSBefore Environment | TSAfter [(Qualified Ident, Type)]
+-- | Holds the data necessary to do type directed search for typed holes
+data TypeSearch
+  = TSBefore Environment
+  -- ^ An Environment captured for later consumption by type directed search
+  | TSAfter [(Qualified Ident, Type)]
+  -- ^ Results of applying type directed search to the previously captured
+  -- Environment
   deriving Show
 
 -- | A type of error messages
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -737,21 +737,28 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
       paras $ [ line "Wildcard type definition has the inferred type "
               , markCodeBox $ indent $ typeAsBox ty
               ] ++ renderContext ctx
-    renderSimpleErrorMessage (HoleInferredType _ _ _ (TSBefore _)) =
-      internalError "oh noez"
-    renderSimpleErrorMessage (HoleInferredType name ty ctx typeSearch) =
-      case typeSearch of
-        TSBefore _ -> internalError "type search should've been run at this point"
-        TSAfter idents ->
-          paras $ [ line $ "Hole '" ++ markCode name ++ "' has the inferred type "
-                  , markCodeBox $ indent $ typeAsBox ty
-                  ]
-          ++ [line "You could substitute the hole with one of these values:"]
-          ++ map (\(i, t) -> (Box.text (showQualified runIdent i))
-                             Box.<>
-                             BoxHelpers.indented (Box.text ":: "
-                                                  Box.<> typeAsBox t)) idents
-          ++ renderContext ctx
+    renderSimpleErrorMessage (HoleInferredType name ty ctx (TSBefore _)) =
+      paras $ [ line $ "Hole '" ++ markCode name ++ "' has the inferred type "
+              , markCodeBox (indent (typeAsBox ty))
+              ] ++ renderContext ctx
+    renderSimpleErrorMessage (HoleInferredType name ty ctx (TSAfter idents)) =
+      paras $ [ line $ "Hole '" ++ markCode name ++ "' has the inferred type "
+              , markCodeBox (indent (typeAsBox ty))
+              ]
+              ++ let formatTS (names, types) =
+                        let
+                          idBoxes = Box.text . showQualified runIdent <$> names
+                          tyBoxes = (\t -> BoxHelpers.indented (Box.text ":: " Box.<> typeAsBox t)) <$> types
+                          longestId = maximum (map Box.cols idBoxes)
+                        in
+                          Box.vcat Box.top $
+                            zipWith (Box.<>)
+                            (Box.alignHoriz Box.left longestId <$> idBoxes)
+                            tyBoxes
+                  in [ line "You could substitute the hole with one of these values:"
+                     , markCodeBox (indent (formatTS (unzip idents)))
+                     ]
+              ++ renderContext ctx
     renderSimpleErrorMessage (MissingTypeDeclaration ident ty) =
       paras [ line $ "No type declaration was provided for the top-level declaration of " ++ markCode (showIdent ident) ++ "."
             , line "It is good practice to provide type declarations as a form of documentation."
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -266,7 +266,7 @@ onTypesInErrorMessageM f (ErrorMessage hints simple) = ErrorMessage <$> traverse
   gSimple (ExpectedType ty k) = ExpectedType <$> f ty <*> pure k
   gSimple (OrphanInstance nm cl ts) = OrphanInstance nm cl <$> traverse f ts
   gSimple (WildcardInferredType ty ctx) = WildcardInferredType <$> f ty <*> traverse (sndM f) ctx
-  gSimple (HoleInferredType name ty ctx) = HoleInferredType name <$> f ty <*> traverse (sndM f) ctx
+  gSimple (HoleInferredType name ty ctx env) = HoleInferredType name <$> f ty <*> traverse (sndM f) ctx  <*> pure env
   gSimple (MissingTypeDeclaration nm ty) = MissingTypeDeclaration nm <$> f ty
   gSimple (CannotGeneralizeRecursiveFunction nm ty) = CannotGeneralizeRecursiveFunction nm <$> f ty
   gSimple other = pure other
@@ -733,10 +733,12 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
       paras $ [ line "Wildcard type definition has the inferred type "
               , markCodeBox $ indent $ typeAsBox ty
               ] ++ renderContext ctx
-    renderSimpleErrorMessage (HoleInferredType name ty ctx) =
+    renderSimpleErrorMessage (HoleInferredType name ty ctx (TypeSearch typeSearch)) =
       paras $ [ line $ "Hole '" ++ markCode name ++ "' has the inferred type "
               , markCodeBox $ indent $ typeAsBox ty
-              ] ++ renderContext ctx
+              ]
+      ++ map (line . showQualified runIdent) (typeSearch ty)
+      ++ renderContext ctx
     renderSimpleErrorMessage (MissingTypeDeclaration ident ty) =
       paras [ line $ "No type declaration was provided for the top-level declaration of " ++ markCode (showIdent ident) ++ "."
             , line "It is good practice to provide type declarations as a form of documentation."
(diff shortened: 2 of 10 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -29,6 +29,9 @@ import qualified Text.Parsec as P
 -- | A map of locally-bound names in scope.
 type Context = [(Ident, Type)]
 
+data TypeSearch = TSBefore Environment | TSAfter [(Qualified Ident, Type)]
```
paf31: Could you please break this over a couple of lines and add some Haddock comments?

src/Language/PureScript/Errors.hs
```diff
@@ -733,10 +737,21 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
...
+      internalError "oh noez"
+    renderSimpleErrorMessage (HoleInferredType name ty ctx typeSearch) =
+      case typeSearch of
+        TSBefore _ -> internalError "type search should've been run at this point"
+        TSAfter idents ->
+          paras $ [ line $ "Hole '" ++ markCode name ++ "' has the inferred type "
+                  , markCodeBox $ indent $ typeAsBox ty
+                  ]
+          ++ [line "You could substitute the hole with one of these values:"]
+          ++ map (\(i, t) -> (Box.text (showQualified runIdent i))
```
paf31: Please indent these lines with `indent` and use `markCodeBox` too.

It would be great if the `::`s could be aligned too. A simple way would be to `unzip` the list of tuples, `vcat` each side and combine the two columns with `<>`.

src/Language/PureScript/Infernal.hs
```diff
@@ -0,0 +1,133 @@
...
+  mkEnv <$> traverse readExternFile modules
+  where
+    readExternFile :: MonadIO m => FilePath -> m P.ExternsFile
+    readExternFile fp = do
+      parseResult <- liftIO (decodeStrict <$> BS.readFile fp)
+      case parseResult of
+        Nothing -> internalError "parsing an extern failed"
+        Just externs -> pure externs
+
+-- Potentially needed for psc-ide integration
```
paf31: Before this is merged, please take out the psc-ide specific code. We can hopefully move it under `Language.PureScript.Ide` later. Actually, while we're talking modules, this one should probably go under `Language.PureScript.Typechecker`

src/Language/PureScript/Infernal.hs
```diff
@@ -0,0 +1,133 @@
...
+  -> P.Type
+  -- ^ The user supplied type
+  -> Either P.MultipleErrors ((P.Expr, [(P.Ident, P.Constraint)]), P.Environment)
+filtering env x t = xrunSubsume env $ do
+  let initializeSkolems = Skolem.introduceSkolemScope <=< P.replaceAllTypeSynonyms <=< P.replaceTypeWildcards
+
+  x' <- initializeSkolems x
+  t' <- initializeSkolems t
+
+  let dummyExpression = P.Var (P.Qualified Nothing (P.Ident "x"))
```
paf31: Please add a comment here explaining why `dummyExpression` is necessary.

src/Language/PureScript/Infernal.hs
```diff
@@ -0,0 +1,133 @@
...
+  elab <- runExceptT $ subsumes (Just dummyExpression) t' x'
+  subst <- gets TC.checkSubstitution
+  case elab of
+    Left _ -> throwError undefined
```
paf31: If we don't need the errors, let's use `Maybe`/`MaybeT`.

src/Language/PureScript/Infernal.hs
```diff
@@ -0,0 +1,133 @@
...
+    Right (Just expP) -> do
+      let expPP = overTypes (P.substituteType subst) expP
+      Entailment.replaceTypeClassDictionaries True expPP
+    Right Nothing -> throwError undefined
+
+typeSearch
```
paf31: Is this version used?

paf31:
This is looking great so far! I'm excited to get this merged in. I think we should aim to get this merged in a minor release shortly after 0.10, and to add the `psc-ide` changes later. What do you think?

src/Language/PureScript/Infernal.hs
```diff
@@ -0,0 +1,133 @@
...
+  t' <- initializeSkolems t
+
+  let dummyExpression = P.Var (P.Qualified Nothing (P.Ident "x"))
+  elab <- runExceptT $ subsumes (Just dummyExpression) t' x'
+  subst <- gets TC.checkSubstitution
+  case elab of
+    Left _ -> throwError undefined
+    Right (Just expP) -> do
+      let expPP = overTypes (P.substituteType subst) expP
+      Entailment.replaceTypeClassDictionaries True expPP
```
kRITZCREEK: True -> False

src/Language/PureScript/Infernal.hs
kRITZCREEK: I've tried this for about 30min now but can't figure it out. Dealing with transformers is still a little tricky for me. Maybe we could take a look at it together?

src/Language/PureScript/Errors.hs
```diff
@@ -733,10 +737,28 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
...
+      paras $ [ line $ "Hole '" ++ markCode name ++ "' has the inferred type "
+              , markCodeBox (indent (typeAsBox ty))
+              ]
+              ++ let formatTS (names, types) =
+                        let
+                          idBoxes = Box.text . showQualified runIdent <$> names
+                          tyBoxes = (\t -> BoxHelpers.indented (Box.text ":: " Box.<> typeAsBox t)) <$> types
+                          longestId = maximum (map Box.cols idBoxes)
+                        in
+                          Box.vcat Box.top $
```
paf31: If you `vcat` each column first, then use `<>`, you won't need to track the width of the left column.

paf31: Oh, of course. Could you please update the PR description to show how this looks now?

kRITZCREEK: Will do when I get home.

kRITZCREEK: Done.

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -0,0 +1,80 @@
...
+typeSearch
+  :: P.Environment
+  -> P.Type
+  -> Map (P.Qualified P.Ident) P.Type
+typeSearch env type' =
+  Map.mapMaybe (\(x, _, _) -> if isRight (filtering env type' x)
+                              then Just x
+                              else Nothing) (P.names env)
+
+overTypes :: (P.Type -> P.Type) -> P.Expr -> P.Expr
```
kRITZCREEK: I did actually copy it from TypeChecker.Types. `everywhereOnValues` is from `Language.PureScript.AST.Traversals` which in turn depends on `Types` and most of the `AST` modules, so I'm not really sure where to put this...

paf31:
The new layout looks great! What's left to be done here?

src/Language/PureScript/TypeChecker/TypeSearch.hs
paf31: Can it be moved into `Language.PureScript.AST.Traversals`?

paf31:
We might need to apply a limit to the number of results we show. In some cases it can basically return everything in context 😄

paf31:
I learned a neat trick by experimenting with this. I used PSCi to try to fill in `?x` here:

```purescript
> traverse ?x [1, 2, 3]
```

and it suggested a lot of functions. The function instance for `Applicative` lets you do some neat things like

```purescript
> traverse take [1, 2, 3] ['x', 'y', 'z']
[['x'], ['x', 'y'], ['x', 'y', 'z']]
```

So I think there is a lot of potential here for exploratory programming.

I did notice some interesting results though, which we should take a look at. Here are some examples:

- The code doesn't check skolem escape, so sometimes it will guess things like `pureST`, but substituting them leads to an escaped skolem. This is not really blocking though.
- `foldMap ?x [1,2, 3]` suggests a bunch of things like `negate` which don't actually work. I'm not sure why, but we should definitely try to figure this one out.

paf31:
Another nice little data point, based on a recent talk by Conor McBride, where the typechecker fills in the only valid implementation by checking that `1 < 2` at compile time:

```
> data Z
> data S n
>
> class LEQ n m
> instance zeroLEQ :: LEQ Z n
> instance succLEQ :: LEQ n m => LEQ (S n) (S m)
>
> data Compared n m = Compared
> let le = Compared :: forall n m. LEQ n m => Compared n m
> let ge = Compared :: forall n m. LEQ m n => Compared n m
>
> ?test :: Compared (S Z) (S (S Z))

Error found:
in module $PSCI
at  line 1, column 1 - line 1, column 30

  Hole 'test' has the inferred type
                         
    Compared (S Z) (S (S Z))
                         
    $PSCI.it  :: Compared (S Z) (S (S Z))            
    $PSCI.le  :: forall n m. (LEQ n m) => Compared n m
                                                  
in value declaration it
```

kRITZCREEK:
with the latest commit, which tries to run the subsumption check with the current substitution the following only permits the `range` function:

`x = traverse ?x [1,2]`

```
  Hole 'x' has the inferred type

    Int -> t0 t1

    Data.Array.range  :: Int -> Int -> Array Int
```

`drop` and `take` are equally valid here though. The code before the last commit generates:

```

  Hole 'x' has the inferred type

    Int -> t0 t1

    Control.Monad.Eff.forE         :: forall e. Int -> Int -> (Int -> Eff e Unit) -> Eff e Unit
    Data.Array.alterAt             :: forall a. Int -> (a -> Maybe a) -> Array a -> Maybe (Array a)
    Data.Array.cons                :: forall a. a -> Array a -> Array a
    Data.Array.delete              :: forall a. (Eq a) => a -> Array a -> Array a
    Data.Array.deleteAt            :: forall a. Int -> Array a -> Maybe (Array a)
    Data.Array.drop                :: forall a. Int -> Array a -> Array a
    Data.Array.elemIndex           :: forall a. (Eq a) => a -> Array a -> Maybe Int
    Data.Array.elemLastIndex       :: forall a. (Eq a) => a -> Array a -> Maybe Int
    Data.Array.insert              :: forall a. (Ord a) => a -> Array a -> Array a
    Data.Array.insertAt            :: forall a. Int -> a -> Array a -> Maybe (Array a)
    Data.Array.modifyAt            :: forall a. Int -> (a -> a) -> Array a -> Maybe (Array a)
    Data.Array.range               :: Int -> Int -> Array Int
    Data.Array.singleton           :: forall a. a -> Array a
    Data.Array.slice               :: forall a. Int -> Int -> Array a -> Array a
    Data.Array.take                :: forall a. Int -> Array a -> Array a
    Data.Array.updateAt            :: forall a. Int -> a -> Array a -> Maybe (Array a)
    Data.Eq.eq                     :: forall a. (Eq a) => a -> a -> Boolean
    Data.Eq.notEq                  :: forall a. (Eq a) => a -> a -> Boolean
    Data.EuclideanRing.div         :: forall a. (EuclideanRing a) => a -> a -> a
    Data.EuclideanRing.mod         :: forall a. (EuclideanRing a) => a -> a -> a
    Data.Function.applyFlipped     :: forall a b. a -> (a -> b) -> b
    Data.Function.const            :: forall a b. a -> b -> a
    Data.Maybe.fromMaybe           :: forall a. a -> Maybe a -> a
    Data.Maybe.maybe               :: forall a b. b -> (a -> b) -> Maybe a -> b
    Data.Ord.between               :: forall a. (Ord a) => a -> a -> a -> Boolean
    Data.Ord.clamp                 :: forall a. (Ord a) => a -> a -> a -> a
    Data.Ord.compare               :: forall a. (Ord a) => a -> a -> Ordering
    Data.Ord.greaterThan           :: forall a. (Ord a) => a -> a -> Boolean
    Data.Ord.greaterThanOrEq       :: forall a. (Ord a) => a -> a -> Boolean
    Data.Ord.lessThan              :: forall a. (Ord a) => a -> a -> Boolean
    Data.Ord.lessThanOrEq          :: forall a. (Ord a) => a -> a -> Boolean
    Data.Ord.max                   :: forall a. (Ord a) => a -> a -> a
    Data.Ord.min                   :: forall a. (Ord a) => a -> a -> a
    Data.Ord.Unsafe.unsafeCompare  :: forall a. a -> a -> Ordering
    Data.Ring.sub                  :: forall a. (Ring a) => a -> a -> a
    Data.Semiring.add              :: forall a. (Semiring a) => a -> a -> a
    Data.Semiring.mul              :: forall a. (Semiring a) => a -> a -> a
    Main.x                         :: t0 (Array t1)

```

paf31:
#2366 fixes the issue I mentioned above actually: 

> foldMap ?x [1,2, 3] suggests a bunch of things like negate which don't actually work.

I think this PR is probably good to merge now, after some small amount of tidying (like the `MaybeT` thing mentioned above).

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -116,11 +118,18 @@ typesOf bindingGroupType moduleName vals = withFreshSubstitution $ do
...
     -- instance resolution (by functional dependencies).
     finalSubst <- gets checkSubstitution
-    escalateWarningWhen isHoleError . tell . onErrorMessages (replaceTypes finalSubst) $ w
+    escalateWarningWhen isHoleError . tell . onErrorMessages (replaceTypes False finalSubst) $ w
 
     return inferred
   where
-    replaceTypes subst = onTypesInErrorMessage (substituteType subst)
+    replaceTypes shouldGeneralize subst =
+      (if not shouldGeneralize then runTypeSearch else id)
```
paf31: Could you please pull this out into a value in the `where` clause and use a guard instead of `if`, for readability?

kRITZCREEK:
Would this be a potential topic to add to the Haskell Communities and Activities Report? https://mail.haskell.org/pipermail/haskell-cafe/2016-October/125204.html

paf31:
Yes good idea. I think `psc-ide` could be its own entry, or we could fold it into the existing PureScript entry. What do you prefer?

kRITZCREEK:
I just learned that there are three principled ways to make an empty array with the Prelude :)

```
    > ?x :: Array Void
    ...
    Control.Plus.empty    :: forall a f. (Plus f) => f a
    Data.Monoid.mempty    :: forall m. (Monoid m) => m
    Data.Unfoldable.none  :: forall f a. (Unfoldable f) => f a
```

We only show the "You could substitute the hole... " part now when we found any suggestions and the maximum number of suggestions is limited to 15.

src/Language/PureScript/Errors.hs
```diff
@@ -733,10 +737,31 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
...
               ] ++ renderContext ctx
-    renderSimpleErrorMessage (HoleInferredType name ty ctx) =
-      paras $ [ line $ "Hole '" ++ markCode name ++ "' has the inferred type "
-              , markCodeBox $ indent $ typeAsBox ty
-              ] ++ renderContext ctx
+    renderSimpleErrorMessage (HoleInferredType name ty ctx ts) =
+      let
+        maxTSResults = 15
+        tsResult = case ts of
+          (TSAfter idents) | not (null idents) ->
```
paf31: An alternative here is to put the guard in the list comprehension:

```haskell
[ line "You could substitute the hole with one of these values:"
, markCodeBox (indent (formatTS (unzip (take maxTSResults idents))))
| not (null idents)
]
```

paf31:
Looks great!

src/Language/PureScript/Errors.hs
kRITZCREEK: Would you actually prefer that? I think it's nicer to put the guard next to the pattern match and I can't get rid of the `_` case either way.

paf31: > I can't get rid of the _ case either way.

You can - if the guard doesn't hold, it'll give the empty list.

kRITZCREEK: The wildcard also catches the `TSBefore` case

Issue #2347, opened 2016-10-01
Update LICENSE file to include new dependencies


paf31:
Another small thing which needs to be done before 0.10

PR #2354, opened 2016-10-02
Use BSD 3-clause license, increment version

### LINKED ISSUE #2341: Switch to BSD3 license
So I don't forget to do this before 0.10.

We use a lot of BSD3 code from Hackage, and we shouldn't redistribute it as binary under MIT, since MIT doesn't include the no-advertising clause.

paf31:
cc @garyb @joneshf @kRITZCREEK @hdgarrood 

We should have been using something compatible with BSD 3 all along, since we release binaries which link in Haskell code released under BSD 3, and BSD 3 licensed code cannot be relicensed under MIT.

Since all contributions are licensed under explicit terms in `CONTRIBUTORS.md` (and it seems MIT code _can_ be relicensed under BSD 3), this should be fine. It will now be easier if individual contributors wan to switch over to BSD3.

Fixes #2341, #2347

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,24 +1,16 @@
-The MIT License (MIT)
-
-Copyright (c) 2013-15 Phil Freeman, (c) 2014-2015 Gary Burgess, and other
+Copyright (c) 2013-16 Phil Freeman, (c) 2014-2016 Gary Burgess, and other
 contributors
+All rights reserved.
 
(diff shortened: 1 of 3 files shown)


rightfold:
I would like to change my contributions license from "existing contributions and all future contributions" to "pre-#2354 contributions", and future contributions to BSD-3-Clause, is that possible?

paf31:
Yes, absolutely, please send a PR once this is merged.

PR #2355, opened 2016-10-02
Use aeson 0.11.* for now, for Pursuit JSON compatibility

### LINKED ISSUE #2346: Make Pursuit JSON format compatible with 0.9.*
Long term, we should probably write the instances by hand (see #2186), but for 0.10, we should try to make the format compatible with 0.9 if possible. This incompatibility might have been caused by using the new version of `aeson`, since the error points to a missing `contents` field in the JSON for something of type `Maybe`.

paf31:
Fixes #2346 temporarily. We should migrate our JSON format to something more robust in the longer term.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -102,7 +102,7 @@ source-repository head
 
 library
     build-depends: base >=4.8 && <5,
-                   aeson >= 0.8 && < 1.1,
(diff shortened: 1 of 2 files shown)


garyb:
:+1:

PR #2356, opened 2016-10-03
Derive Data.Generic.Rep.Generic

### LINKED ISSUE #2348: Derive Data.Generic.Rep.Generic

paf31:
Fixes #2348

This needs a little more work, since it doesn't handle record types yet, but everything else seems to be working.

Here is a short [demo](https://asciinema.org/a/c6ud16qlbob5gtjs5sp79e6y3).


paf31:
I've added basic support for records (that is, data types with a single argument which is a record - we can expand this later if necessary).

Tests should start passing once purescript/purescript-generics-rep#2 is merged.

paf31:
Tests are now passing, apart from the one which was failing already anyway.

tests/support/bower.json
```diff
@@ -8,6 +8,7 @@
     "purescript-prelude": "1.1.0",
     "purescript-st": "1.0.0-rc.1",
     "purescript-partial": "1.1.2",
-    "purescript-newtype": "0.1.0"
+    "purescript-newtype": "0.1.0",
+    "purescript-generics-rep": "2.0.0"
```
garyb: Should we update the rest of the versions of these while we're at it? Will probably need to tweak some tests, but may as well get it done, since the dependencies conflict with this (I think?).

paf31: I think we definitely should soon, but I'd like to leave it for a separate PR after this one. I think there might be a lot more changes needed than we suspect 😄

garyb: Okidoke. How does these even work now without `resolutions` specified?!

paf31: I think v2.0 of `generics-rep` used the pre-`0.10` libraries.

garyb:
👍 then, but I guess we should bump things so we can test against the latest `generics-rep` before releasing 0.10.2 :smile:

paf31:
Yes, agreed. I'll write up an issue.

PR #2364, opened 2016-10-06
Refactor subsumes function

paf31:
@kRITZCREEK This should remove the need for the `dummyExpression` in your PR. Could you please review this?

I got the idea from the "Practical type inference for arbitrary-rank types" paper, but this really just codifies some assumptions I was making in the code, and asserting using `error`. Now that information is tracked in the type system using a type level boolean which says whether we're currently inserting type class dictionaries or not.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Subsumption.hs
+++ b/src/Language/PureScript/TypeChecker/Subsumption.hs
@@ -1,6 +1,8 @@
--- |
--- Subsumption checking
---
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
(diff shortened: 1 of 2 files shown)


kRITZCREEK:
This PR generally improves the `subsumes` codes readability and I'm all for merging it!

I don't think it allows me to get rid of `dummyExpression` though, since I still need to pass an expression to `replaceTypeClassDictionaries`. I rebased my `type-directed-search` branch onto this PR and got up to this point: https://github.com/kRITZCREEK/purescript/commit/2cafd38a15a729c3c74f8abec510384af1f45d08

PR #2366, opened 2016-10-06
Treat type annotations on top-level expressions as if they were type declarations

### LINKED ISSUE #2362: Compiler infers wrong constraints for inline annotations
In `psci`:

```purescript
import Prelude
f = (\x -> show (x + x)) :: forall a. a -> String
```
compiles just fine and infers the following type for `f`:
```
(Show a0, Semiring a0) => (forall a. a -> String)
```

However with a standalone type declaration like so:
```purescript
import Prelude
f  :: forall a. a -> String
f = (\x -> show (x + x))
```

`psc` correctly fails with:

```

    Data.Show.Show a0
```

paf31:
Treat type annotations on top-level expressions as if they were type declarations.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -27,12 +27,14 @@ module Language.PureScript.TypeChecker.Types
 
 import Prelude.Compat
 
+import Control.Arrow (second)
 import Control.Monad
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State.Class (MonadState(..), gets)
 import Control.Monad.Supply.Class (MonadSupply)
 import Control.Monad.Writer.Class (MonadWriter(..))
 
+import Data.Bifunctor (bimap)
 import Data.Either (lefts, rights)
 import Data.List (transpose, nub, (\\), partition, delete)
 import Data.Maybe (fromMaybe)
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -199,6 +201,10 @@ typeForBindingGroupElement (ident, val) dict untypedDict = do
 -- | Check if a value contains a type annotation
 isTyped :: (Ident, Expr) -> Either (Ident, Expr) (Ident, (Expr, Type, Bool))
 isTyped (name, TypedValue checkType value ty) = Right (name, (value, ty, checkType))
+isTyped (name, PositionedValue pos c value) =
+  bimap (second (PositionedValue pos c))
+        (second (\(e, t, b) -> (PositionedValue pos c e, t, b)))
+        (isTyped (name, value))
```
kRITZCREEK: Does this mean `PositionedValue` can only appear at the top level? Or does isTyped only get called on values that are at the top level?

kRITZCREEK:
Looks good to me! Just the one question to make sure I understood things right.

src/Language/PureScript/TypeChecker/Types.hs
paf31: At the top level, there are `TypedValue`s under `PositionedValue`, but those should be the only two constructors we need to remove to find a type annotation.

PR #2369, opened 2016-10-07
Pretty-print suggested types differently

### LINKED ISSUE #1793: Types generated by wildcards and typed holes should work when substituted
note sure how to name or describe this problem, but there is a concise way to reproduce it:

http://stackoverflow.com/questions/34593686/purescript-type-signature-fails-to-compile-code-works-fine-without-suggested-s

don't think these are related, but they are the closest (open) matches i found: #1110, #1579.

paf31:
Fixes #1793

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -731,7 +731,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
             ]
     renderSimpleErrorMessage (WildcardInferredType ty ctx) =
       paras $ [ line "Wildcard type definition has the inferred type "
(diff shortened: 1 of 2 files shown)


kRITZCREEK:
` where a0 is a rigid type variable` still refers to the old name.

paf31:
Thanks. Now only the suggestion is changed, the type shown to the user still uses the numbered skolems and named unification variables.

paf31:
@kRITZCREEK Do you think this looks ok now?

kRITZCREEK:
Yes :+1:

Am 11.10.2016 4:40 vorm. schrieb "Phil Freeman" <notifications@github.com>:

PR #2375, opened 2016-10-08
Fix #2370, allow rows in instance contexts

### LINKED ISSUE #2370: Rows should be allowed in instance contexts
Not in instance heads, but in contexts, like they are allowed in function contexts. This would allow us to derive things like `Newtype` instances for newtypes around records:

```purescript
instance newtypeRecordNewtype ::
  TypeEquals inner { message :: String }
    => Newtype RecordNewtype inner where
  wrap = RecordNewtype
  unwrap (RecordNewtype rec) = rec
```

paf31:
If this looks okay, then I'll open an issue to derive `Newtype` instances like this whenever a record appears under a newtype.

### DIFF
--- a/examples/passing/RowsInInstanceContext.purs
+++ b/examples/passing/RowsInInstanceContext.purs
@@ -1,0 +1,25 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff (Eff)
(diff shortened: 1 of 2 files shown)


garyb:
Yes please! 👍

PR #2381, opened 2016-10-10
Small cleanup to Language.PureScript.Interactive.IO

phiggins:
Hello,

I was poking around in the code while looking into some of the beginner issues and noticed some small problems:

* `expandTilde` is not referenced anywhere else. It looks like the last caller was removed in 9e562663210e0bb798cf183a25fb84a142a7f17b.
* A few other methods were being exported implicitly and were not referenced outside of this module.

I am very new to purescript and haskell so I'm sorry if these sorts of contributions are not welcome.


paf31:
Travis says

```
/home/travis/build/purescript/purescript/src/Language/PureScript/Interactive/IO.hs:8:1: Warning:
    The import of ‘isPathSeparator’
    from module ‘System.FilePath’ is redundant
```

Could you please fix that and then we can merge this in?

phiggins:
@paf31 Oops, I didn't notice the Travis output. It looks like I needed to `stack clean` before `stack build` would report the failure locally. Good to know!

I fixed that error in https://github.com/purescript/purescript/pull/2381/commits/8f3f2380a7e591c32277f1b10474c363b1ec6f40.

paf31:
Thanks!

PR #2384, opened 2016-10-10
Allow symbols in data contructors

### LINKED ISSUE #2208: Allow prime symbols in data constructors
It's great to have primes in type names, but can we have them in data constructors too? It feels a little weird to write something like this:
```purescript
module Bar where

import Foo (Foo)

newtype Foo' = FooP Foo

### PR DESCRIPTION
**Pull request by brandonhamilton**
I would like to attempt to contribute a solution to resolve #2208 

As I am new to the Purescript compiler, I am not sure if there is anything that I have overlooked here or if there is perhaps a better way to achieve this.

In essence, the changes I have made are:

- Modify the Lexer to accept the same range of tokens for data constructors as those for type constructors.
- Modify the Codegen to pass data constructor strings through the logic for generating appropriate javascript names from identifiers.
- Added an example to use as a test case.

### DIFF
--- a/examples/passing/DctorName.purs
+++ b/examples/passing/DctorName.purs
@@ -1,11 +1,24 @@
 module Main where
 
+import Prelude
 import Control.Monad.Eff.Console (log)
 
 newtype Bar' = Bar' Int
 
 data Foo' = Foo' Bar'
 
-data Baz'' = Baz''
+data Baz'' = Baz'' | Baz'
+
+f :: Foo' -> Boolean
+f a = case a of Foo' b -> true
+
+g :: Baz'' -> Int
+g Baz'' = 0
+g Baz' = 1
+ 
+h :: Bar' -> Int
+h (Bar' x)
+ | x <= 10 = x * 2 
+ | otherwise = 10
 
 main = log "Done"
--- a/examples/passing/DctorName.purs
+++ b/examples/passing/DctorName.purs
@@ -9,16 +9,25 @@ data Foo' = Foo' Bar'
 
 data Baz'' = Baz'' | Baz'
 
-f :: Foo' -> Boolean
-f a = case a of Foo' b -> true
+f ∷ Foo' → Boolean
+f a = case a of Foo' b → true
 
-g :: Baz'' -> Int
+f' ∷ Boolean
+f' = f $ Foo' $ Bar' 0
+
+g ∷ Baz'' → Int
 g Baz'' = 0
 g Baz' = 1
- 
-h :: Bar' -> Int
+
+g' ∷ Int
+g' = g Baz''
+
+h ∷ Bar' → Int
 h (Bar' x)
  | x <= 10 = x * 2 
  | otherwise = 10
+
+h' ∷ Int
+h' = h $ Bar' 4
 
 main = log "Done"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: examples/passing/DctorName.purs [hunk 1]
```diff
@@ -0,0 +1,11 @@
...
+
+import Control.Monad.Eff.Console (log)
+
+newtype Bar' = Bar' Int
+
+data Foo' = Foo' Bar'
+
+data Baz'' = Baz''
+
+main = log "Done"
```
paf31: Could you please add an example where you `case` on one of these new constructors? Thanks.

src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -244,23 +244,23 @@ moduleToJs (Module coms mn imps exps foreigns decls) foreign_ =
     ret <- valueToJs val
     return $ JSApp Nothing (JSFunction Nothing Nothing [] (JSBlock Nothing (ds' ++ [JSReturn Nothing ret]))) []
   valueToJs' (Constructor (_, _, _, Just IsNewtype) _ (ProperName ctor) _) =
-    return $ JSVariableIntroduction Nothing ctor (Just $
+    return $ JSVariableIntroduction Nothing (identToJs . Ident $ ctor) (Just $
```
paf31: It's probably worth pulling this out into its own small function.

paf31: `properNameToJs`?

paf31:
Looks great so far, thanks!

examples/passing/DctorName.purs
brandonhamilton: Thanks, this has caught the case where the parser for the constructor binding in case statements did not deal with the symbols. I have updated the parser and example.

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -16,6 +16,11 @@ moduleNameToJs (ModuleName pns) =
   let name = intercalate "_" (runProperName `map` pns)
   in if nameIsJsBuiltIn name then "$$" ++ name else name
 
+properToJs :: String -> String
+properToJs name
+  | nameIsJsReserved name || nameIsJsBuiltIn name = "$$" ++ name
+  | otherwise = concatMap identCharToString name
```
LiamGoodacre: Might be worth refactoring `identToJs` to use this in the `(Ident name)` case - as the implementation is the same.

brandonhamilton: Thanks !

examples/passing/DctorName.purs
```diff
@@ -0,0 +1,24 @@
...
+data Baz'' = Baz'' | Baz'
+
+f :: Foo' -> Boolean
+f a = case a of Foo' b -> true
+
+g :: Baz'' -> Int
+g Baz'' = 0
+g Baz' = 1
+ 
+h :: Bar' -> Int
```
paf31: Could you please make sure these cases actually get run? You could do this by adding additional top-level declarations of type `Int` for example, where you apply each function.

brandonhamilton: Thanks for the feedback. Would this do ?

    f' ∷ Boolean
    f' = f $ Foo' $ Bar' 0

    g' ∷ Int
    g' = g Baz''

    h' ∷ Int
    h' = h $ Bar' 4

paf31:
Thanks!

Issue #2383, opened 2016-10-10
Shadow warning issue with do


rightfold:
```purescript
do
 x <- let x = 1 in pure x
 -- …
```

This warns about `x` being shadowed. However, it clearly isn't

LiamGoodacre:
I've looked into this - PR incoming.
Basically the `everythingWithScope` traversal is using the wrong identifier `Set` for the bind expression.

PR #2385, opened 2016-10-11
Fix scope traversal for do-notation bind.

LiamGoodacre:
Added a warning test which lists no `@shouldWarnWith` pragmas to check that the shadowing warning isn't raised.  Is this an acceptable use of the test system?

Unfortunately I can't verify that the test passes as I haven't been able to build at all due to: https://github.com/commercialhaskell/stack/issues/2577

### DIFF
--- a/examples/warning/2383.purs
+++ b/examples/warning/2383.purs
@@ -1,0 +1,12 @@
+-- | This specifically shouldn't warn about `x` being shadowed in `main`
+-- | See https://github.com/purescript/purescript/issues/2383
+module Main where
+
+import Prelude
+
+import Control.Monad.Eff (Eff)
+
+main :: Eff () Unit
+main = do
+  x <- let x = pure unit in x
+  pure unit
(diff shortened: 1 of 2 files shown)


LiamGoodacre:
Oh, this is for https://github.com/purescript/purescript/issues/2383

rvion:
did you put examples/warning/2383.purs in the wrong folder ?

https://travis-ci.org/purescript/purescript/jobs/166656389#L836

```
Warning examples
  '2383.purs' should compile with warning(s) '' FAILED [1]
```

LiamGoodacre:
It depends, the test wants to check that a specific warning does not get raised - therefore, it feels to me like it should be a warning test.  However it seems we don't currently have a way to express this?  I was expecting it to check that no warnings were raised if I didn't list any `@shouldWarnWith` pragmas, but that doesn't appear to be the case.

paf31:
The fix looks good, thanks. If you could figure out the test issue, that'd be great, but we might have to change the test runner.

LiamGoodacre:
Okay the test is passing now.  The reason the previous test run failed is that my warning example was written badly 😸 (type error, oops).

paf31:
Thanks!

PR #2386, opened 2016-10-12
Fix issue with typed holes in inference mode

paf31:
@kRITZCREEK Can you please review this? It fixes the `foldMap` issue I ran into earlier.

### DIFF
--- a/src/Language/PureScript/TypeChecker/TypeSearch.hs
+++ b/src/Language/PureScript/TypeChecker/TypeSearch.hs
@@ -19,6 +19,7 @@ import           Language.PureScript.TypeChecker.Unify       as P
 import           Control.Monad.Supply                        as P
 import           Language.PureScript.AST                     as P
 import           Language.PureScript.Environment             as P
+import           Language.PureScript.Errors             as P
 import           Language.PureScript.Names                   as P
 import           Language.PureScript.TypeChecker.Skolems     as Skolem
 import           Language.PureScript.TypeChecker.Synonyms    as P
@@ -26,29 +27,32 @@ import           Language.PureScript.Types                   as P
 
 checkInEnvironment
   :: Environment
-  -> Integer
-  -> StateT TC.CheckState (SupplyT (WriterT b (Except e))) a
+  -> TC.CheckState
+  -> StateT TC.CheckState (SupplyT (WriterT b (Except P.MultipleErrors))) a
   -> Maybe (a, Environment)
-checkInEnvironment env nextVar =
+checkInEnvironment env st =
   either (const Nothing) Just
   . runExcept
   . evalWriterT
-  . P.evalSupplyT nextVar
-  . TC.runCheck' env
+  . P.evalSupplyT 0
+  . TC.runCheck' (st { TC.checkEnv = env })
 
 evalWriterT :: Monad m => WriterT b m r -> m r
 evalWriterT m = liftM fst (runWriterT m)
 
 checkSubsume
-  :: P.Environment
+  :: Maybe [(P.Ident, Entailment.InstanceContext, P.Constraint)]
+  -- ^ Additional constraints we need to satisfy
+  -> P.Environment
   -- ^ The Environment which contains the relevant definitions and typeclasses
-  -> Integer
+  -> TC.CheckState
+  -- ^ The typechecker state
   -> P.Type
   -- ^ The user supplied type
   -> P.Type
   -- ^ The type supplied by the environment
-  -> Maybe ((P.Expr, [(P.Ident, P.Constraint)]), P.Environment)
-checkSubsume env nextVar userT envT = checkInEnvironment env nextVar $ do
+  -> Maybe ((P.Expr, [(P.Ident, Entailment.InstanceContext, P.Constraint)]), P.Environment)
+checkSubsume cons env st userT envT = checkInEnvironment env st $ do
   let initializeSkolems =
         Skolem.introduceSkolemScope
         <=< P.replaceAllTypeSynonyms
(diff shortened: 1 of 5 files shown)


kRITZCREEK:
So I've tried this and it does not show any suggestions for

```
> traverse ?x [1, 2, 3]
> foldMap ?x [1, 2, 3]
```

It does show correct ones for:

```
> foldMap ?x [1, 2, 3] :: String
> traverse ?x [1, 2, 3] :: Array Int -> Array (Array Int)
```

but I think it's a little sad it doesn't find anything for `traverse ?x [1, 2, 3]` anymore, since that's how you found about about the nice trick with `take` in the first place.

paf31:
@kRITZCREEK Correct, it doesn't show anything in inference mode any more, which includes PSCi, but I think we can fix that. We can hold off this PR until we find the right fix if you prefer.

kRITZCREEK:
I'd like to do that then. I've come to really enjoy exploring with typed holes in psci 😸

paf31:
@kRITZCREEK Updated with the changes we discussed, but still no fix for the `Additive` issue.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -113,46 +110,47 @@ typesOf bindingGroupType moduleName vals = withFreshSubstitution $ do
...
       checkDuplicateLabels val'
-      return (ident, (foldr (Abs . Left . fst) val' unsolved, generalized))
+      return ((ident, (foldr (Abs . Left . (\(x, _, _) -> x)) val' unsolved, generalized)), unsolved)
 
     -- Show warnings here, since types in wildcards might have been solved during
     -- instance resolution (by functional dependencies).
     finalSubst <- gets checkSubstitution
+    checkState <- get
+    forM_ tys $ \(shouldGeneralize, ((_, (_, _)), w)) -> do
+      escalateWarningWhen isHoleError . tell . onErrorMessages (replaceTypes (guard shouldGeneralize $> concatMap snd inferred) finalSubst checkState) $ w
```
kRITZCREEK: Why use `concatMap` over `foldMap`? Just curious, since I'd almost forgotten about that function :D

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -19,36 +19,40 @@ import           Language.PureScript.TypeChecker.Unify       as P
...
+  -> TC.CheckState
+  -- ^ The typechecker state
   -> P.Type
   -- ^ The user supplied type
   -> P.Type
   -- ^ The type supplied by the environment
-  -> Maybe ((P.Expr, [(P.Ident, P.Constraint)]), P.Environment)
-checkSubsume env nextVar userT envT = checkInEnvironment env nextVar $ do
+  -> Maybe ((P.Expr, [(P.Ident, Entailment.InstanceContext, P.Constraint)]), P.Environment)
+checkSubsume cons env st userT envT = checkInEnvironment env st $ do
```
kRITZCREEK: `cons` is a bit of a loaded term, I'd prefer `constraints` here.

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -62,12 +66,21 @@ checkSubsume env nextVar userT envT = checkInEnvironment env nextVar $ do
   elab <- subsumes envT' userT'
   subst <- gets TC.checkSubstitution
   let expP = P.overTypes (P.substituteType subst) (elab dummyExpression)
-  Entailment.replaceTypeClassDictionaries False expP
+
+  -- Now check that any unsolved constraints have not become impossible
+  (traverse_ . traverse_) (\(_, context, constraint) -> do
+    let constraint' = P.mapConstraintArgs (map (P.substituteType subst)) constraint
+    flip evalStateT Map.empty . evalWriterT $
+      Entailment.entails True False (P.TypeClassDictionary constraint' context [])) cons
```
kRITZCREEK: Maybe add a comment here, or haddocks on the `entails` definition for the two Booleans? The first one is `shouldGeneralize` and the second one is `defer`.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -113,46 +110,47 @@ typesOf bindingGroupType moduleName vals = withFreshSubstitution $ do
...
-    replaceTypes shouldRunTypeSearch subst nextVar =
-      (if shouldRunTypeSearch then runTypeSearch else id)
-      . onTypesInErrorMessage (substituteType subst)
+    replaceTypes
+      :: Maybe [(Ident, InstanceContext, Constraint)]
+      -> Substitution
+      -> CheckState
+      -> ErrorMessage
+      -> ErrorMessage
+    replaceTypes cons subst st =
```
kRITZCREEK: Maybe we should pull this out of the where clause and give it a few haddocks? It doesn't depend on any of the arguments of `typesOf` and is complicated enough to need its own signature.

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -19,36 +19,40 @@ import           Language.PureScript.TypeChecker.Unify       as P
 import           Control.Monad.Supply                        as P
 import           Language.PureScript.AST                     as P
 import           Language.PureScript.Environment             as P
+import           Language.PureScript.Errors             as P
```
kRITZCREEK: OCD kicking in...

paf31:
@kRITZCREEK Updated, thanks for the review! How does this look?

kRITZCREEK:
Great, LGTM!

PR #2389, opened 2016-10-15
Fix inliner for Data.Array.unsafeIndex

### LINKED ISSUE #2388: Invalid generated code in Data.Array.Partial
```purescript
var head = function (dictPartial) {
    return function (xs) {
        return dictPartial[xs](0);
    };
};
```

This must be relatively recent.

brandonhamilton:
Fixes #2388

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
+++ b/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
@@ -160,7 +160,7 @@ inlineCommonOperators = applyAll $
 
   , inlineNonClassFunction (isModFn (C.dataFunction, C.apply)) $ \f x -> JSApp Nothing f [x]
   , inlineNonClassFunction (isModFn (C.dataFunction, C.applyFlipped)) $ \x f -> JSApp Nothing f [x]
-  , inlineNonClassFunction (isDictModFn (C.partial, C.dataArray, C.unsafeIndex)) $ flip (JSIndexer Nothing)
+  , inlineNonClassFunction (isModFnWithDict (C.dataArray, C.unsafeIndex)) $ flip (JSIndexer Nothing)
   ] ++
   [ fn | i <- [0..10], fn <- [ mkFn i, runFn i ] ]
   where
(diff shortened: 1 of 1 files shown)


paf31:
Thanks! Just one nitpick about naming, looks great otherwise.

src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
```diff
@@ -233,6 +224,21 @@ inlineCommonOperators = applyAll $
...
     go _ _   _ = Nothing
 
+  inlineNonClassFunction :: (JS -> Bool) -> (JS -> JS -> JS) -> JS -> JS
+  inlineNonClassFunction p f = everywhereOnJS convert
+    where
+    convert :: JS -> JS
+    convert (JSApp _ (JSApp _ op' [x]) [y]) | p op' = f x y
+    convert other = other
+
+  isMF :: (String, String) -> JS -> Bool
```
brandonhamilton: I was going for MF (Module and Function) and DMF (Dictionary, Module and Function), but realize that using acronyms was not very readable. I have pushed changes that better describe these functions. Thanks :)

src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
```diff
@@ -233,6 +224,21 @@ inlineCommonOperators = applyAll $
...
+    convert :: JS -> JS
+    convert (JSApp _ (JSApp _ op' [x]) [y]) | p op' = f x y
+    convert other = other
+
+  isModFn :: (String, String) -> JS -> Bool
+  isModFn (m, op) (JSAccessor _ op' (JSVar _ m')) = m == m' && op == op'
+  isModFn _ _ = False
+
+  isDictModFn :: (String, String, String) -> JS -> Bool
+  isDictModFn (d, m, op) (JSApp _ (JSAccessor _ op' (JSVar _ m')) [(JSVar _ d')]) = m == m' && op == op' && d' == "dict" ++ d
```
paf31: I would ignore the dictionary name here entirely, actually. After type-checking, it's not like we can apply this function to something which is not a `Partial` dictionary anyway, so I would just leave it as a wildcard in case the code generation changes later.

paf31:
👍 Thanks!

PR #2391, opened 2016-10-16
Support record updates on records with field named hasOwnProperty

rightfold:
Previously this would fail.


src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -291,7 +291,7 @@ moduleToJs (Module coms mn imps exps foreigns decls) foreign_ =
       evaluate = JSVariableIntroduction Nothing evaluatedObj (Just obj)
       objAssign = JSVariableIntroduction Nothing newObj (Just $ JSObjectLiteral Nothing [])
       copy = JSForIn Nothing key jsEvaluatedObj $ JSBlock Nothing [JSIfElse Nothing cond assign Nothing]
-      cond = JSApp Nothing (JSAccessor Nothing "hasOwnProperty" jsEvaluatedObj) [jsKey]
+      cond = JSApp Nothing (JSAccessor Nothing "call" (JSAccessor Nothing "hasOwnProperty" (JSObjectLiteral Nothing []))) [jsEvaluatedObj, jsKey]
```
paf31: Does this have any performance implications?

rightfold: May allocate `{}` every time, but maybe V8 optimizes that away.

garyb: It most certainly does not optimize {}, or didn't not too long ago - we got a major performance benefit from avoiding those (they were quite pervasive though). Using `Object.prototype.hasOwnProperty` would be the better option I think.

rightfold: But that would break when a variable named `Object` is in scope.

natefaubion: You could add Object to the list of reserved JS identifiers?

garyb: `(0).hasOwnProperty` is kinda weird but I prefer it to the cache option for sure. Making `Object` reserved is probably an excellent idea if we don't already do that though, as I imagine it could muck up a thing or two.

natefaubion:
(0) will implicitly box the number. I'd be very interested to see perf comparisons for stuff like this. It's obviously very sensitive.

rightfold:
It will not do that in strict mode

natefaubion:
Are you sure that's the case here? `hasOwnProperty` is part of the Object prototype, so it has to get an Object from somewhere. My personal vote goes to using `Object.keys`, as its universally faster (and safer) than `hasOwnProperty`, but I realize we unofficially support ES3 (though I think going ES5 and telling users to use polyfills is perfectly reasonable).

/cc @michaelficarra

rightfold:
It will short circuit to `Number.prototype` without boxing, and `this` will also be unboxed in the callee, when in strict mode.

michaelficarra:
:+1: I use `{}.hasOwnProperty.call(o, property)` all the time and recommend that others do so as well. I would get a microbenchmark together to show that it's not a perf concern, but http://jsperf.com is down right now.

paf31:
@michaelficarra So this looks good to merge as-is? Thanks for the review.

michaelficarra:
Yep, LGTM. If we're really concerned about perf, we could set up a benchmark using benchmark.js directly (looks like JSPerf is permanently down), but I don't think that's necessary.

paf31:
Thanks!

PR #2393, opened 2016-10-20
Handle TypeLevelString when checking orphans

### LINKED ISSUE #2378: Error "Invalid type in instance in checkOrphanInstance"
I get this error when instances include type-level strings.

joneshf:
Fixes #2378

### DIFF
--- a/examples/failing/2378.purs
+++ b/examples/failing/2378.purs
@@ -1,0 +1,6 @@
+-- @shouldFailWith OrphanInstance
+module Main where
+
+import Lib
+
+instance fooX :: Foo "x"
--- a/examples/failing/2378/Lib.purs
+++ b/examples/failing/2378/Lib.purs
@@ -1,0 +1,3 @@
+module Lib (class Foo) where
+
+class Foo (a :: Symbol)
--- a/examples/passing/2378.purs
+++ b/examples/passing/2378.purs
@@ -1,0 +1,9 @@
+module Main where
+
+import Control.Monad.Eff.Console (log)
+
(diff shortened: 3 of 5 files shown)


joneshf:
I'm not sure why those two builds failed on travis. I ran both `BUILD_TYPE=sdist COVERAGE=true stack test` and `BUILD_TYPE=sdist stack test`, and the tests pass in both cases. Is there some other state I'm missing here?

paf31:
Hmm, odd. Thanks for the fix, it looks good.

CI says

```
1) Failing examples '2378.purs' should fail with 'OrphanInstance'
       Expected these errors: ["OrphanInstance"], but got these: ["UnknownName"]
```

but I'm not sure why.

kRITZCREEK:
it's because `examples/failing/2378/Lib.purs` is not matched by the wildcards in `.cabal`

joneshf:
Strange. Do you know why it worked for the other builds? In anycase, I'll add the glob and push up.

Would it make sense to have `examples/failing/**/*.purs` instead?

garyb:
Annoyingly globs like that don't work for `extra-source-files` - it only allows filename wildcards.

kRITZCREEK:
@joneshf the dist builds package the compiler like Cabal would before uploading it to hackage. This only grabs the files specified in the cabal file. We then unpack the distribution into a fresh directory and run the build in there. This then misses the files that were present before from cloning the repo and fails.

joneshf:
Ah, makes sense!

paf31:
Thanks!

PR #2396, opened 2016-10-22
Fix #2379, add error message for unknown classes

### LINKED ISSUE #2379: Error "entails: type class not found in environment"
I get this error message when a module uses a type class in an instance declaration's context, but does not export it, and then I try to use that instance in another module.

paf31:
This seemed preferable to throwing on the declaration of the instance itself, since we might want classes and instances which are internal to a module.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -601,7 +601,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
             ]
     renderSimpleErrorMessage OverlappingInstances{} = internalError "OverlappingInstances: empty instance list"
     renderSimpleErrorMessage (UnknownClass nm) =
-      paras [ line "No instance instance was found for class"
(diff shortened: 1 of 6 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -599,6 +600,11 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
             , line "They may be disallowed completely in a future version of the compiler."
             ]
     renderSimpleErrorMessage OverlappingInstances{} = internalError "OverlappingInstances: empty instance list"
+    renderSimpleErrorMessage (UnknownClass nm) =
+      paras [ line "No instance instance was found for class"
```
garyb: "instance instance"

garyb:
You could use the `UnknownName` error for this, but the message would be less specific about how it arose then, so maybe this is the better way.

Looks like the cabal file needs updating to include the new test `Lib` file. Looks good to me apart from that!

garyb:
👍

PR #2401, opened 2016-10-24
Include module comments in functional core dump

paulyoung:
I noticed that these were missing from the JSON output when using the `--dump-corefn` flag.

### DIFF
--- a/src/Language/PureScript/CoreFn/ToJSON.hs
+++ b/src/Language/PureScript/CoreFn/ToJSON.hs
@@ -13,6 +13,7 @@ import Data.Version (Version, showVersion)
 import Data.Text (pack)
 
 import Language.PureScript.AST.Literals
+import Language.PureScript.Comments
 import Language.PureScript.CoreFn
 import Language.PureScript.Names
 
@@ -24,6 +25,9 @@ literalToJSON _ (CharLiteral c) = toJSON ("CharLiteral", c)
 literalToJSON _ (BooleanLiteral b) = toJSON ("BooleanLiteral", b)
 literalToJSON t (ArrayLiteral xs) = toJSON ("ArrayLiteral", map t xs)
 literalToJSON t (ObjectLiteral xs) = toJSON ("ObjectLiteral", recordToJSON t xs)
+
+commentToJSON :: Comment -> Value
+commentToJSON = toJSON
 
 identToJSON :: Ident -> Value
 identToJSON = toJSON . runIdent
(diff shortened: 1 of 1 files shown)


paulyoung:
Here's a sample:

```purescript
-- |
-- Consume the core functional representation in JSON format
--
module CoreFn.FromJSON
  ( moduleFromJSON
  ) where

import CoreFn.Module (Module)
import Data.Either (Either)
import Data.Foreign (ForeignError)
import Data.Foreign.Class (readJSON)

moduleFromJSON :: String -> Either ForeignError Module
moduleFromJSON = readJSON
```

```json
{"CoreFn.FromJSON":{"imports":["Prim","CoreFn.Module","Data.Either","Data.Foreign","Data.Foreign.Class"],"builtWith":"0.10.1","exports":["moduleFromJSON"],"decls":[{"moduleFromJSON":["App",["Var","Data.Foreign.Class.readJSON"],["Var","CoreFn.Module.isForeignModule"]]}],"comments":[{"LineComment":" |"},{"LineComment":" Consume the core functional representation in JSON format"},{"LineComment":""}],"foreign":[]}}
```

```json
{
  "CoreFn.FromJSON":{
    "imports":[
      "Prim",
      "CoreFn.Module",
      "Data.Either",
      "Data.Foreign",
      "Data.Foreign.Class"
    ],
    "builtWith":"0.10.1",
    "exports":[
      "moduleFromJSON"
    ],
    "decls":[
      {
        "moduleFromJSON":[
          "App",
          [
            "Var",
            "Data.Foreign.Class.readJSON"
          ],
          [
            "Var",
            "CoreFn.Module.isForeignModule"
          ]
        ]
      }
    ],
    "comments":[
      {
        "LineComment":" |"
      },
      {
        "LineComment":" Consume the core functional representation in JSON format"
      },
      {
        "LineComment":""
      }
    ],
    "foreign":[

    ]
  }
}
```

paf31:
This looks good, thanks, but I'm wondering if we need comments in the core output at all. If we add them for modules, then why not for declarations? The possible issue is that lots of comments could slow down parsing and make it more difficult to read the JSON. 

Also, I would probably not differentiate between line and block comments, but instead just bundle them up into one string or array of strings.

paulyoung:
I'm still getting familiar with the functional core and assumed that other comments may have been accounted for elsewhere.

I could close this and submit another pull request to remove `moduleComments` from `Module` if that's the preference.

paf31:
> remove moduleComments from Module if that's the preference

From the core AST? No, I think we probably should keep it there.

I'm not saying this is a bad idea, I just wanted to understand the motivation before we commit to it (it's harder to remove things later than to add them now, etc.)

paulyoung:
I thought if it was in `Module` then the intention must have been for it to be included in the JSON output.

I can see now that there's not a 1-1 relationship, so happy to close this. The changes were so easy to introduce that I just created a PR rather than an issue in this case.

PR #2406, opened 2016-10-27
[WIP][psc-ide] Show operators in completions for their alias

### LINKED ISSUE #2373: [psc-ide] Show operators for functions
When searching for `bind` it might be nice to also return `>>=` in the completions. This would be an easy way to discover operators and to insert unicode operators. Operators should always be last in the completion list to reduce the surprise factor.

Thoughts? @nwolverson @FrigoEU @rvion

kRITZCREEK:
This still needs tests for the type operator case and I want to make sure operator completions appear last when we are searching for identifiers.

Fixes #2373

### DIFF
--- a/src/Language/PureScript/Ide/Filter.hs
+++ b/src/Language/PureScript/Ide/Filter.hs
@@ -25,6 +25,7 @@ module Language.PureScript.Ide.Filter
 
 import           Protolude                     hiding (isPrefixOf)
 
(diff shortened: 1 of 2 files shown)


kRITZCREEK:
While this partially works, I don't like the approach and the emacs plugin has been fighting me on the sorting. I'll need to think of a different angle for this.

PR #2408, opened 2016-10-28
Change NullBinder serialization to simplify parsing

jacereda:


jacereda:
With my limited `purescript-foreign` experience I was unable to find a way to parse the current Binder structure. The code is at:

 https://github.com/jacereda/purescript-corefn-parser/blob/master/src/Data/CoreFn/Parser.purs

hdgarrood:
This code should parse the current structure, so I don't think we need to change the encoding. I've used the `Alt Either` instance to avoid nested `case` expressions.

```
instance isForeignBinder :: IsForeign Binder where
  read v = readOtherBinder <|> readNullBinder
  where
  readOtherBinder = do
    nb <- v ! 0 >>= read
    case nb of
      "LiteralBinder" -> LiteralBinder <$> (v ! 1 >>= read)
      "VarBinder" -> VarBinder <$> (v ! 1 >>= read)
      "ConstructorBinder" -> ConstructorBinder <$> (v ! 1 >>= read) <*> (v ! 2 >>= read) <*> (v ! 3 >>= read)
      "NamedBinder" -> NamedBinder <$> (v ! 1 >>= read) <*> (v ! 2 >>= read)
      l -> fail $ JSONError $ "binder:" <> l
  readNullBinder =
    str <- read v
    case str of
      "NullBinder" -> pure NullBinder
      l -> fail $ JSONError $ "binder: " <> l
```

jacereda:
Oh, got it. In any case, wouldn't it be better to just use my proposal to avoid the overhead of backtracking?

hdgarrood:
Perhaps, but it's not immediately clear to me that there would be enough of a performance difference to justify it.

paf31:
Closing for the reason above, but we can discuss this some more if there is a compelling reason to break the format.

Issue #2374, opened 2016-10-08
Add error message for ambiguous type variables in inferred contexts


paf31:
Right now we reuse `NoInstanceFound` for this, but it would be better if we indicated that we found an instance but didn't use it because it involved ambiguous type variables.

joneshf:
Can this be closed? Seems like https://github.com/purescript/purescript/pull/2410 and https://github.com/purescript/purescript/pull/2428 addressed this enhancement.

paf31:
Yes, thanks.

PR #2410, opened 2016-10-29
2374: Add error message for ambiguous type variables in inferred contexts

bbqbaron:
This adds [ambiguous inferred type errors](https://github.com/purescript/purescript/issues/2374) for 1.0. How does the error message look? It's my first time mucking about in the compiler; did I infer correctly when to throw the error?

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -638,8 +638,10 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
         where
         go TUnknown{} = True
         go _ = False
-    renderSimpleErrorMessage (AmbiguousTypeVariables t (Constraint nm ts _)) =
-      paras [ line $ "The inferred type " ++ (markCode (prettyPrintTypeAtom t)) ++ " has type variables which are not mentioned in the body of the type. Consider adding a type annotation."
+    renderSimpleErrorMessage (AmbiguousTypeVariables t _) =
+      paras [ line "The inferred type"
+            , indent $ line $ markCode $ prettyPrintType t
+            , line "has type variables which are not mentioned in the body of the type. Consider adding a type annotation."
             ]
     renderSimpleErrorMessage (PossiblyInfiniteInstance nm ts) =
       paras [ line "Type class instance for"
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -118,6 +118,7 @@ errorCode em = case unwrapErrorMessage em of
   ConstrainedTypeUnified{} -> "ConstrainedTypeUnified"
   OverlappingInstances{} -> "OverlappingInstances"
   NoInstanceFound{} -> "NoInstanceFound"
+  AmbiguousTypeVariables{} -> "AmbiguousTypeVariables"
   UnknownClass{} -> "UnknownClass"
   PossiblyInfiniteInstance{} -> "PossiblyInfiniteInstance"
   CannotDerive{} -> "CannotDerive"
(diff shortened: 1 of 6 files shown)


src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -111,7 +111,7 @@ typesOf bindingGroupType moduleName vals = withFreshSubstitution $ do
           let solved = foldMap (S.fromList . fdDetermined) typeClassDependencies
           let constraintTypeVars = nub . foldMap (unknownsInType . fst) . filter ((`notElem` solved) . snd) $ zip (constraintArgs con) [0..]
           when (any (`notElem` unsolvedTypeVars) constraintTypeVars) $ do
-            throwError . onErrorMessages (replaceTypes currentSubst) . errorMessage $ NoInstanceFound con
+            throwError . onErrorMessages (replaceTypes currentSubst) . errorMessage $ AmbiguousTypeVariables ident generalized con
```
bbqbaron: Whoops; this code doesn't confirm that there are multiple instances of the constraint typeclass in existence. It seems like the strictly most correct response would be:

No instances -> NoInstanceFound
One instance -> Try to solve the type signature with that instance's type argument
Two instances -> AmbiguousTypeVariables

I'm not sure I have the expertise for that just yet, though.

paf31:
What you have looks correct to me. Thanks!

src/Language/PureScript/TypeChecker/Types.hs
paf31: this will be handled elsewhere, so you don't need to worry about it here. An ambiguous type means something else, where a type variable appears in a constraint, but not in the body of the type, so it can't be inferred.

src/Language/PureScript/Errors.hs
```diff
@@ -636,6 +638,15 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
         where
         go TUnknown{} = True
         go _ = False
+    renderSimpleErrorMessage (AmbiguousTypeVariables ident t (Constraint nm ts _)) =
+      paras [ line $ markCode (showIdent ident) ++ " doesn't have a type declaration, so it was inferred to be of type: "
```
paf31: I would omit this line, since it will just duplicate the information from the value declaration level (` ... in declaration x ...`) anyway. In fact, I'd remove the `Ident` from the `SimpleErrorMessage` constructor.

paf31:
Looks great so far, thanks! Please see my one comment, and update `CONTRIBUTORS.md`.

src/Language/PureScript/Errors.hs
```diff
@@ -636,6 +638,15 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
         where
         go TUnknown{} = True
         go _ = False
+    renderSimpleErrorMessage (AmbiguousTypeVariables t (Constraint nm ts _)) =
+      paras [ line "Inferred type"
+            , markCodeBox $ typeAtomAsBox t
+            , line " has constraints, but we don't know what type values they require. Please add an explicit type declaration for: "
```
paf31: Sorry to be picky, but can this be just "The inferred type ... has type variables which are not mentioned in the body of the type. Consider adding a type annotation"? Thanks.

bbqbaron: No problem; picky is good. Incoming.

src/Language/PureScript/Errors.hs
```diff
@@ -636,6 +638,9 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
         where
         go TUnknown{} = True
         go _ = False
+    renderSimpleErrorMessage (AmbiguousTypeVariables t (Constraint nm ts _)) =
+      paras [ line $ "The inferred type " ++ (markCode (prettyPrintTypeAtom t)) ++ " has type variables which are not mentioned in the body of the type. Consider adding a type annotation."
```
paf31: Sorry, one more 😄 :

Please put the type on its own line, `indent` it, and use `prettyPrintType` not `prettyPrintTypeAtom`. This will make it more consistent with other errors which display full types. The problem right now is that you risk it spilling over multiple lines.

Thanks!

src/Language/PureScript/Errors.hs
```diff
@@ -636,6 +638,11 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
         where
         go TUnknown{} = True
         go _ = False
+    renderSimpleErrorMessage (AmbiguousTypeVariables t _) =
+      paras [ line "The inferred type"
+            , indent $ line $ markCode $ prettyPrintType t
```
paf31: I just realized this should use `typeAsBox`, since this way, the indentation gets messed up if the type takes up multiple lines.

I can put together a PR for this later, unless you want to take a look @bbqbaron ?

bbqbaron: Happy to; I'll set a reminder for tomorrow.

bbqbaron: [Here we are](https://github.com/purescript/purescript/pull/2428).

Issue #2411, opened 2016-10-30
Missing ShadowedName cases


garyb:
``` purescript
test :: forall m. Monad m => Int -> m Int
test x =
  let x = 0
  in pure x
```

``` purescript
test :: forall m. Monad m => Int -> m Int
test x = do
  x <- pure x
  pure x
```

I did check these with what's in `master` this time :wink:

PR #2412, opened 2016-10-30
Traversal should pick up bindings in all value declarations.

LiamGoodacre:
For #2411

I still can't compile due to macOS Sierra 😦 so I'm relying on the build servers to tell me things are okay.

We seem to have lots of problems with these traversal functions.
I might have a play sometime to see if we can structure it in such a way that we can't make a mistake... No idea what that would look like though 😺.

### DIFF
--- a/examples/warning/2411.purs
+++ b/examples/warning/2411.purs
@@ -1,0 +1,15 @@
+-- @shouldWarnWith ShadowedName
+module Main where
+
+import Prelude
+
+import Control.Monad.Eff (Eff)
+
+test :: forall m. Monad m => Int -> m Unit
+test x =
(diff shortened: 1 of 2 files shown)


garyb:
Offtopic, but what's the trouble with Sierra? I updated and haven't had any trouble compiling (with Stack at least).

LiamGoodacre:
I've been getting very strange issues.  Currently trying to nuke my `~/.stack` and `purescript/.stack-work` and see if I can do everything from scratch.  Issues I've seen include `malformed mach-o` and `ghc: panic! (the 'impossible' happened)`.

paf31:
Thanks!

Issue #2323, opened 2016-09-22
Sort when adding an explicit import with psc-ide


kRITZCREEK:
Right now we just add the new identifier to the beginning of the import list. This creates unnecessary noise in git diffs.

Preferably we'd not only sort alphabetically but also sort categorically so that datatype imports or class imports are grouped together.

garyb:
We have sorting like that in the suggestions, so should use the same ordering here I guess :)

kRITZCREEK:
That's a good hint. I'll try to find it

garyb:
It's probably just whatever comes out thanks to the `DeclarationRef` `Ord` instance.

nwolverson:
I think the change for this is now merged?

kRITZCREEK:
Thanks!

PR #2413, opened 2016-10-31
2323: Sort IDE-generated explicit imports

bbqbaron:
[Issue 2323](https://github.com/purescript/purescript/issues/2323) makes the IDE sort generated explicit imports instead of just prepending.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -260,6 +260,33 @@ instance Eq DeclarationRef where
   r == (PositionedDeclarationRef _ _ r') = r == r'
   _ == _ = False
 
+-- enable sorting lists of explicitly imported refs
+instance Ord DeclarationRef where
+  compare = compDecRef
+
+compDecRef :: DeclarationRef -> DeclarationRef -> Ordering
+compDecRef (TypeRef name _) (TypeRef name' _) = compare name name'
+compDecRef (TypeOpRef name) (TypeOpRef name') = compare name name'
+compDecRef (ValueRef ident) (ValueRef ident') = compare ident ident'
+compDecRef (ValueOpRef name) (ValueOpRef name') = compare name name'
+compDecRef (TypeClassRef name) (TypeClassRef name') = compare name name'
+compDecRef (TypeInstanceRef ident) (TypeInstanceRef ident') = compare ident ident'
+compDecRef (ModuleRef name) (ModuleRef name') = compare name name'
+compDecRef (ReExportRef name _) (ReExportRef name' _) = compare name name'
+compDecRef (PositionedDeclarationRef _ _ ref) (PositionedDeclarationRef _ _ ref') = compare ref ref'
+compDecRef ref ref' = compare (orderOf ref) (orderOf ref')
+    where
+      orderOf :: DeclarationRef -> Int
+      orderOf (TypeRef _ _) = 7
+      orderOf (TypeOpRef _) = 3
+      orderOf (ValueRef _) = 8
+      orderOf (ValueOpRef _) = 4
+      orderOf (TypeClassRef _) = 5
+      orderOf (TypeInstanceRef _) = 6
+      orderOf (ModuleRef _) = 2
+      orderOf (ReExportRef _ _) = 1
+      orderOf (PositionedDeclarationRef _ _ _) = 0
+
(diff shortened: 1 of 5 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -260,6 +260,33 @@ instance Eq DeclarationRef where
...
+compDecRef :: DeclarationRef -> DeclarationRef -> Ordering
+compDecRef (TypeRef name _) (TypeRef name' _) = compare name name'
+compDecRef (TypeOpRef name) (TypeOpRef name') = compare name name'
+compDecRef (ValueRef ident) (ValueRef ident') = compare ident ident'
+compDecRef (ValueOpRef name) (ValueOpRef name') = compare name name'
+compDecRef (TypeClassRef name) (TypeClassRef name') = compare name name'
+compDecRef (TypeInstanceRef ident) (TypeInstanceRef ident') = compare ident ident'
+compDecRef (ModuleRef name) (ModuleRef name') = compare name name'
+compDecRef (ReExportRef name _) (ReExportRef name' _) = compare name name'
+compDecRef (PositionedDeclarationRef _ _ ref) (PositionedDeclarationRef _ _ ref') = compare ref ref'
```
kRITZCREEK: I think we just want to drop the position info for the compare instance, so I'd prefer two cases here:

```haskell
compDecRef (PositionedDeclarationRef _ _ ref) ref' = compare ref ref'
compDecRef ref (PositionedDeclarationRef _ _ ref') = compare ref ref'
```

This way we get the expected result when only one of the two declarations carries position info.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -216,7 +216,7 @@ addExplicitImport' decl moduleName imports =
     -- TypeDeclaration "Maybe" + Data.Maybe (maybe) -> Data.Maybe(Maybe, maybe)
     insertDeclIntoImport :: IdeDeclaration -> Import -> Import
     insertDeclIntoImport decl' (Import mn (P.Explicit refs) Nothing) =
-      Import mn (P.Explicit (insertDeclIntoRefs decl' refs)) Nothing
+      Import mn (P.Explicit (sort $ insertDeclIntoRefs decl' refs)) Nothing
```
kRITZCREEK: Please use braces over `$` for consistency here.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -230,7 +230,7 @@ addExplicitImport' decl moduleName imports =
 
     insertDtor dtor (P.TypeRef tn' dtors) =
       case dtors of
-        Just dtors' -> P.TypeRef tn' (Just (ordNub (dtor : dtors')))
+        Just dtors' -> P.TypeRef tn' (Just (sort $ ordNub (dtor : dtors')))
```
kRITZCREEK: Please use braces over `$` for consistency here.

tests/Language/PureScript/Ide/ImportsSpec.hs
```diff
@@ -3,6 +3,7 @@
 module Language.PureScript.Ide.ImportsSpec where
 
 import           Protolude
+import           Data.List                       (nub)
```
kRITZCREEK: You can use `ordNub` which is exported by `Protolude`.

src/Language/PureScript/AST/Declarations.hs
kRITZCREEK: Actually on second thought this breaks consistency with regards to the `Eq` instance so maybe it's better to not make this an `Ord` instance and just use it as a function with `sortWith`

nwolverson:
My impression was that consensus was more for `Blah(..)` rather than explicit data constructors, and I *think* the warning suggestion was originally the other way and changed.

garyb:
^ Yeah, we've been moving towards `(..)` in most cases, as there's no benefit to being specific with constructors the way there is with members.

kRITZCREEK:
Allright so we want to use the same sorting for suggestions, which we use for psc-ide's import formatting. Also we don't want to explicitly list data constructors anymore, but instead just use `..` whenever there is at least one data constructor.

bbqbaron:
Updates are in! I've left comments tagging changes. I think the only thing I know I don't know is the ordering for declref types (drawn from the existing linter code).

kRITZCREEK:
Does the right thing in my testing and the code looks great! Thanks!

paf31:
Great work, thanks!

Issue #2209, opened 2016-06-24
psc-ide: filewatcher issues on MacOS X


adamw:
This doesn't happen always, but the larger the project & the source file, the more often this happens. When running `psc-ide` and e.g. adding a definition in one file, this isn't then visible when editing another file. 

Here's the original issue as reported in `psc-ide-emacs`: https://github.com/epost/psc-ide-emacs/issues/51#issuecomment-228432565

One suggestion by @kRITZCREEK was:

"If the problems only pop up on OSX it might be a possibility to use the bindings for [kqueue](https://hackage.haskell.org/package/kqueue) on OSX and stick with fsnotify on the other systems."

paf31:
@kRITZCREEK What would you like to do with this issue?

kRITZCREEK:
Well I can't reliably reproduce it and I really have no idea how to tackle this either. One way to circumvent this issue would be to implement a custom polling solution and offer it as an alternative to the filewatcher. This isssue has been pretty stale though, so it might be time spent on a very rare problem.

paf31:
What do we currently use for watching files?

kRITZCREEK:
https://hackage.haskell.org/package/fsnotify

natefaubion:
FWIW I hit this issue every day :) working on slamdata (which is around 1000 modules) I have to reload the server every time I switch files because literally nothing propagates correctly. I manually invoke the checker when I save, so I don't really know architecture-wise why it needs a watcher. Should the server know dependent modules based on its in memory representation of the DAG?

natefaubion:
What I mean to say, my editor knows which file it's being invoked on. Maybe there's a way I can  manually tell pscide to update via a command. A file watcher shouldn't even be necessary then. I realize it's needed for other use cases.

kRITZCREEK:
The file watcher watches the `output` directory and reloads `externs.json` files when they change.

> literally nothing propagates correctly

That sounds terrible 🙀 We should definitely look into it then. One thing you could do for now instead of restarting the server is to send a `reset` command followed by a `load` command.

So the problematic workflow you're having is:

1. Everything compiles fine
2. Edit a source file and add a new exported definition `foo`
3. Save and invoker the rebuild command
4. It succeeds
5. You change to a different file in your editor and invoke completions
6. `foo` doesn't show up

If that sounds about right I'll download the 500mb of slamdata onto my macbook 💃

natefaubion:
[...]

PR #2414, opened 2016-11-01
[psc-ide] Polling option for psc-ide-server

kRITZCREEK:
This is investigative work for #2209. It forces the use of polling for the filewatcher.

@garyb or @natefaubion could you give this branch a try? Right now this branch ignores the `--force-polling` flags and always polls, but if it works out I'll add the option and we can add a toggle to the editor plugins.

### DIFF
--- a/psc-ide-server/Main.hs
+++ b/psc-ide-server/Main.hs
@@ -41,6 +41,7 @@ import           Network.Socket                    hiding (PortNumber, Type,
 import           Options.Applicative               (ParseError (..))
 import qualified Options.Applicative               as Opts
 import           System.Directory
+import           System.Info                       as SysInfo
 import           System.FilePath
 import           System.IO                         hiding (putStrLn, print)
 import           System.IO.Error                   (isEOFError)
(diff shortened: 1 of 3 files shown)


garyb:
Thanks - I'll give this a try soon, it'll probably be once we're done updating SlamData for 0.10, since it'll be easier to replicate there :wink:

kRITZCREEK:
Oh right... I could make a branch and backport it to 0.8.5, unless the slamdata update is coming soon^^

garyb:
We're not that out of date! Currently on 0.9.3 - but yep, the update is underway: https://github.com/slamdata/slamdata/issues/1185

garyb:
Giving this a try, finally!

garyb:
It's not perfect - I have a case here where I've changed the definition of a type in one file and it's not being picked up in another - but it's _way_ better (it's actually usable again :smile:). I have a feeling the problem I'm describing there happens some other way though, as I've had it happen on other projects where `psc-ide-server` isn't suffering from this particular problem.

kRITZCREEK:
That sounds fantastic! In that case I'll try the polling option a bit while playing around in pscid and if it doesn't degrade my experience I'll set is as the default and otherwise make it an option.

garyb:
I probably should have mentioned in the above, that was on Windows at least. I'm unable to test on OSX since I can't compile it currently (#2426).

kRITZCREEK:
After more testing with the current master and this branch @natefaubion and @garyb report that the file watcher works fine as-is on OSX and needs the polling option on Windows, so I've made polling the default on Windows and added a flag `--polling` that toggles the polling behaviour to the `psc-ide-server` executable.

I don't have a proper Windows Haskell development environment atm, so it would be nice if someone could check that the flag is actually reversed on Windows. I added logging of the parsed options when the server is started with `--debug`.

kRITZCREEK:
I gave this a spin on windows and made a minor modification, but it seems to detect the OS properly now. Ready to go if CI passes.

paf31:
👍 Great!

PR #2417, opened 2016-11-03
Speeds up parsing by reading files as Text

kRITZCREEK:
This doesn't change anything inside the parsed AST, it just changes the initial file read to use Text, so we don't need to load all the source files into memory as `String`.

The problem is I have no idea how to properly benchmark this, but consecutive runs through unix's `time` show a 5% improvement on `atom-ide-purescript`.


hdgarrood:
:+1: seems sensible enough to me.

paf31:
Brilliant, thanks!

Issue #1801, opened 2016-01-07
Incorrect error/warning position ranges


natefaubion:
I'm working on a frontend for  that prints original source ranges, so it makes it very easy to spot which errors have incorrect position reporting. This issue is to catalog all those we find so we don't have to make separate issues for them.

natefaubion:
**ShadowedName**

If you do something like
```purescript
foo = \baz ->
  case baz of
    Just baz -> 42
    Nothing  -> 12
```

It returns the position of the entire top level expression, which can require some hunting. It should point to where the shadowed name was bound.

Also, interestingly, if I change it to:
```purescript
foo' baz = case baz of
  Just baz -> 42
  Nothing  -> 12
```

No warning is generated.

natefaubion:
**WildcardInferredType**

The position ranges over the entire type declaration. It should point to a specific wildcard, otherwise its confusing if you have multiple wildcards.

natefaubion:
**UnknownValue**

Given:
```purescript
module Test.Foo where

foo n = n + 1
```

It reports the error as
```
Error found:
in module Test.Foo
at /Users/nathan/Sites/purescript/purescript-psa/test/Foo.purs line 3, column 9 - line 3, column 13

  Unknown value (+)
```

The correct start range would be line 3, column 11.

natefaubion:
**ErrorParsingModule**

Given:
```purescript
module Test.Foo where

improt Prelude
import Bar
```

It reports the error as
```
Error found:
at /Users/nathan/Sites/purescript/purescript-psa/test/Foo.purs line 4, column 1 - line 4, column 1

  Unable to parse module:
  unexpected "import"
  expecting binder or indentation
```

Which is an empty span.

user471:
**NoInstanceFound**
```
module Test where
import Prelude

type Foo = {foo::Int}
foo :: Foo -> Foo -> Boolean
foo x y = x == y
```
```
Error found:
in module Test
at c:\prog\ps\test2\src\Test.purs line 5, column 1 - line 6, column 1

  No type class instance was found for

    Prelude.Eq { foo :: Int
               }


in value declaration foo


See https://github.com/purescript/purescript/wiki/Error-Code-NoInstanceFound for more information,
or to contribute content related to this error.

```

natefaubion:
**InfiniteType**
```purescript
module InfiniteType where
import Prelude (map)

foo = "wat"
  where
  woo = "woo"
  wat x = map (\k -> wat k x)
```
[...]

PR #2427, opened 2016-11-11
Add value source positions

nwolverson:
Addresses a few errors with overgenerous source spans listed in https://github.com/purescript/purescript/issues/1801 - specifically `UnknownName` for data constructors, identifiers and operators.

(I previously tried to add this for value atoms for source maps, at which point it broke something, though tests at least seem to pass now).


paf31:
Looks good to me, thanks!

nwolverson:
So this might be better if operators actually work as well as giving improved source spans! Tests should hopefully now pass

paf31:
Thanks!

Issue #2400, opened 2016-10-23
Test suite should use latest versions of core libraries


paf31:

paf31:
It would be really great if we could switch to `psc-package` for pulling test dependencies at the same time actually.

paf31:
@LiamGoodacre I think you fixed this with your most recent PR, yes?

LiamGoodacre:
Yes, fixed in https://github.com/purescript/purescript/pull/2511

Although, do you want to convert this issue into: use psc-package for test deps?

paf31:
Sounds good, but I'll open another issue.

PR #2429, opened 2016-11-15
Solving IsSymbol instances

LiamGoodacre:
Opening this PR for discussion: there are some TODOs and other things.

Currently many of the `example` tests fail since I updated `tests/support/bower.json` to specify the current latest versions of libraries.  They fail with `The foreign module implementation for module Unsafe.Coerce is missing.`.  I added `purescript-unsafe-coerce` as a dependency, but it didn't help.  So I'm not sure why that fails.

However my `SolvingIsSymbol` test does pass 👍

### DIFF
--- a/examples/passing/SolvingIsSymbol.purs
+++ b/examples/passing/SolvingIsSymbol.purs
@@ -1,0 +1,10 @@
+module Main where
+
+import Prelude
+import Data.Symbol
+import Control.Monad.Eff
+import Control.Monad.Eff.Console
+
+main = do
+  let lit = reflectSymbol (SProxy :: SProxy "literal")
+  when (lit == "literal") (log "Done")
--- a/examples/passing/SolvingIsSymbol.purs
+++ b/examples/passing/SolvingIsSymbol.purs
@@ -1,10 +1,13 @@
 module Main where
 
 import Prelude
-import Data.Symbol
 import Control.Monad.Eff
 import Control.Monad.Eff.Console
 
+-- Here we import as alias of reflectSymbol without importing Data.Symbol. However,
+-- Data.Symbol should be implicitly imported as we have an instance of IsSymbol solved.
+import SolvingIsSymbol.Lib (literalSymbol, libReflectSymbol)
+
 main = do
-  let lit = reflectSymbol (SProxy :: SProxy "literal")
+  let lit = libReflectSymbol literalSymbol
   when (lit == "literal") (log "Done")
--- a/src/Language/PureScript/CoreFn/Desugar.hs
+++ b/src/Language/PureScript/CoreFn/Desugar.hs
@@ -212,6 +212,10 @@ findQualModules decls =
   fqValues :: A.Expr -> [ModuleName]
   fqValues (A.Var q) = getQual' q
   fqValues (A.Constructor q) = getQual' q
+  -- IsSymbol instances for literal symbols are automatically solved and the type
+  -- class dictionaries are built inline instead of having a named instance defined
+  -- and imported.  We therefore need to import the IsSymbol constructor from
+  -- Data.Symbol if it hasn't already been imported.
   fqValues (A.TypeClassDictionaryConstructorApp C.IsSymbol _) = getQual' C.IsSymbol
   fqValues _ = []
 
--- a/src/Language/PureScript/CoreFn/Desugar.hs
+++ b/src/Language/PureScript/CoreFn/Desugar.hs
@@ -13,6 +13,7 @@ import Language.PureScript.AST.Literals
 import Language.PureScript.AST.SourcePos
 import Language.PureScript.AST.Traversals
 import Language.PureScript.Comments
+import qualified Language.PureScript.Constants as C
 import Language.PureScript.CoreFn.Ann
 import Language.PureScript.CoreFn.Binders
 import Language.PureScript.CoreFn.Expr
@@ -110,7 +111,9 @@ moduleToCoreFn env (A.Module _ coms mn decls (Just exps)) =
     exprToCoreFn ss com (Just ty) v
   exprToCoreFn ss com ty (A.Let ds v) =
     Let (ss, com, ty, Nothing) (concatMap (declToCoreFn ss []) ds) (exprToCoreFn ss [] Nothing v)
-  exprToCoreFn ss com _  (A.TypeClassDictionaryConstructorApp name (A.TypedValue _ (A.Literal (A.ObjectLiteral vs)) _)) =
+  exprToCoreFn ss com ty (A.TypeClassDictionaryConstructorApp name (A.TypedValue _ lit@(A.Literal (A.ObjectLiteral _)) _)) =
+    exprToCoreFn ss com ty (A.TypeClassDictionaryConstructorApp name lit)
+  exprToCoreFn ss com _ (A.TypeClassDictionaryConstructorApp name (A.Literal (A.ObjectLiteral vs))) =
     let args = map (exprToCoreFn ss [] Nothing . snd) $ sortBy (compare `on` fst) vs
         ctor = Var (ss, [], Nothing, Just IsTypeClassConstructor) (fmap properToIdent name)
     in foldl (App (ss, com, Nothing, Nothing)) ctor args
@@ -209,6 +212,7 @@ findQualModules decls =
   fqValues :: A.Expr -> [ModuleName]
   fqValues (A.Var q) = getQual' q
   fqValues (A.Constructor q) = getQual' q
+  fqValues (A.TypeClassDictionaryConstructorApp C.IsSymbol _) = getQual' C.IsSymbol
   fqValues _ = []
 
   fqBinders :: A.Binder -> [ModuleName]
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -19,7 +19,7 @@ import Control.Monad.State
 import Control.Monad.Supply.Class (MonadSupply(..))
 import Control.Monad.Writer
 
-import Data.Foldable (for_)
+import Data.Foldable (for_, fold, toList)
 import Data.Function (on)
 import Data.List (minimumBy, nub)
 import Data.Maybe (fromMaybe, maybeToList, mapMaybe)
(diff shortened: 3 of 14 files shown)


src/Language/PureScript/TypeClassDictionaries.hs
```diff
@@ -8,10 +8,10 @@ import Language.PureScript.Types
 -- |
 -- Data representing a type class dictionary which is in scope
 --
-data TypeClassDictionaryInScope
+data TypeClassDictionaryInScope n
   = TypeClassDictionaryInScope {
     -- | The identifier with which the dictionary can be accessed at runtime
-      tcdName :: Qualified Ident
+      tcdName :: n
```
paf31: Could we just make this into an `Expr`?

paf31:
Looks great, thanks!

Yes, you'll have some issues with library versions if you update them all to v2.0 like that. There is an issue to track the update that needs to happen (#2400) but it's probably quite a bit of work. Maybe you can avoid it for now?

src/Language/PureScript/TypeClassDictionaries.hs
LiamGoodacre: I think I tried something like that previously (think it was `Either Expr (Qualified Ident)`) but it caused a module dependency cycle.  I'll have another look later today.

LiamGoodacre:
In that case I should probably use `purescript-symbols` at version `1.0.1` and drop the `TypeConcat` example.

examples/passing/SolvingIsSymbol.purs
```diff
@@ -0,0 +1,11 @@
+module Main where
+
+import Prelude
+import Data.Symbol
+import Control.Monad.Eff
+import Control.Monad.Eff.Console
+
+main = do
+  log (reflectSymbol (SProxy :: SProxy "literal"))
```
rvion: How about checking the result too ?
```
let 
    a = (reflectSymbol (SProxy :: SProxy "literal"))
    b = (reflectSymbol (SProxy :: SProxy (TypeConcat "Hello" "World")))
when (a == "literal" && b == "HelloWorld") (log "Done")
```

LiamGoodacre: Sounds good, thanks

LiamGoodacre: (also I didn't realise outputting `Done` was interpretted as 'test passing')

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -103,6 +106,18 @@ data SolverOptions = SolverOptions
...
+-- | Build a type class instance dictionary for the `IsSymbol` class for a symbol
+isSymbolDictionary :: String -> TypeClassDict
+isSymbolDictionary sym =
+  TypeClassDictionaryInScope (Left dict) [] C.IsSymbol [TypeLevelString sym] Nothing
+  where
+    inst = ObjectLiteral
+      [ ("reflectSymbol", Abs (Left (Ident C.__unused)) (Literal (StringLiteral sym)))
+      ]
+    classTy = TypeConstructor (fmap coerceProperName C.IsSymbol)
+    expr = TypedValue False (Literal inst) (TypeApp classTy (TypeLevelString sym))
```
LiamGoodacre: I wasn't sure about `TypedValue` - have I done this correctly?  I didn't work out what the boolean argument was for and I didn't see it mentioned in the doc-comment for the constructor.

LiamGoodacre:
@paf31: making `TypeClassDictionaryInScope` store the `Expr` is going to require some reorganising of module dependencies.

The following is a trimmed down version of our modules and their dependencies:
```
TypeClassDictionaries

Environment
  TypeClassDictionaries

AST.Declarations
  TypeClassDictionaries
  Environment

AST
  AST.Declarations

Externs
  AST
  Environment
  TypeClassDictionaries

TypeChecker.Monad
  Environment
  TypeClassDictionaries

TypeChecker.Entailment
  AST
  Environment
  TypeChecker.Monad
  TypeClassDictionaries

TypeChecker.Types
  TypeChecker.Entailment

TypeChecker
  AST
  Environment
  TypeChecker.Monad
  TypeChecker.Types
  TypeClassDictionaries
```

We need `TypeClassDictionaries` to depend on `Expr` which is defined in `AST.Declarations`. This causes a cycle between `AST.Declarations`, `Environment`, and `TypeClassDictionaries`.

LiamGoodacre:
An alternative might be to remove the name/expr from TypeClassDictionaryInScope and just pair the dict up with the Expr where needed.  I'll have a think about it.

LiamGoodacre:
Although that is actually the same solution as what I've already done with adding the type parameter.

paf31:
@LiamGoodacre I think we should stick with the original approach actually, using a sum type, now that I understand the problem properly (sorry!)

But let's rename some things to make them more understandable:

- `tcdName` should be named something like `tcdValue` since it's more general now.
- `NamedTypeClassDictionaryInScope` is a bit wordy. How about just `NamedDict` or something?

src/Language/PureScript/TypeClassDictionaries.hs
```diff
@@ -21,4 +21,7 @@ data TypeClassDictionaryInScope
     -- | Type class dependencies which must be satisfied to construct this dictionary
     , tcdDependencies :: Maybe [Constraint]
     }
-    deriving (Show)
+    deriving (Show, Functor)
```
paf31: Might as well derive the usual `Functor`, `Foldable` and `Traversable` here.

paf31:
A third option is a sum type like

```haskell
data Evidence
  = NamedDictionary Ident
  | IsSymbolDictionary String
```

right?

src/Language/PureScript/TypeChecker/Entailment.hs
paf31: Do we need `TypedValue` here? I think it can be removed actually.

paf31:
I can confirm that this makes my [`generics-rep` PR](https://github.com/purescript/purescript-generics-rep/pull/5) build successfully.

I did run into one weird issue though. If I don't import `Data.Symbol` (and `IsSymbol` specifically) in my test suite, then I get the following error:

```
psc: An internal error ocurred during compilation: "Missing value in mnLookup"
```

src/Language/PureScript/TypeChecker/Entailment.hs
LiamGoodacre: It appeared to be required by `exprToCoreFn` in `src/Language/PureScript/CoreFn/Desugar.hs` here:
```haskell
exprToCoreFn ss com _  (A.TypeClassDictionaryConstructorApp name (A.TypedValue _ (A.Literal (A.ObjectLiteral vs)) _)) =
```

Otherwise it hits the error case:
```haskell
exprToCoreFn _ _ _ e =
    error $ "Unexpected value in exprToCoreFn mn: " ++ show e
```

paf31: Maybe you could add a case for 

```
exprToCoreFn ss com _  (A.TypeClassDictionaryConstructorApp name (A.Literal (A.ObjectLiteral vs))) =
```

or use a pattern guard here then?

src/Language/PureScript/TypeClassDictionaries.hs
paf31: Makes sense, thanks!

paf31:
I think there is relatively little needed to push this over the finish line, right? Just

- Switching back to a sum type instead of using `Expr` directly.
- Dealing with the `TypedValue` issue.
- Maybe some renamings here and there (`tcdName` in particular)

I think once this is merged, we should release 0.10.3, since it will enable some nice library features which depend on `generics-rep` deriving (among others).

LiamGoodacre:
Hey thanks for the review.  I've been away without internet for a few days so I haven't had a chance to work on this.  I should be able to get this done today/tomorrow though 👍

LiamGoodacre:
I've been looking into the `Missing value in mnLookup` issue from the `renameModules` part of `moduleToJs`.  This occurs because the dictionaries are created inline and use the type class constructor (which hasn't necessarily been imported).  The easiest way to fix this is to not use the constructor and just have the object literal - is that okay to do?  This also means that I don't need to change `exprToCoreFn`.

garyb:
I would say no, not really. The dictionaries are not going to be records when the typechecker gets fixed for #928.

LiamGoodacre:
@garyb yeah that makes sense, thanks.  So we'll need to implicitly import the type class constructor during code gen - what would be the preferred way of doing that?

garyb:
Hmm, that's a tricky one actually... I'm not too sure! Probably as part of the `CoreFn` desugaring process, if we can figure it out there? There used to be some extra import elaboration done in there, but it was removed after it got figured out propery during name desugaring instead. Pretty sure we won't be able to do it that early on in this case though, right?

src/Language/PureScript/CoreFn/Desugar.hs
```diff
@@ -209,6 +212,7 @@ findQualModules decls =
   fqValues :: A.Expr -> [ModuleName]
   fqValues (A.Var q) = getQual' q
   fqValues (A.Constructor q) = getQual' q
+  fqValues (A.TypeClassDictionaryConstructorApp C.IsSymbol _) = getQual' C.IsSymbol
```
LiamGoodacre: This lets us implicitly add an import of `Data.Symbol` if the `IsSymbol` class is used.  The import also gets deduped later if it's already there.
Is this a good place to add this?

LiamGoodacre: Perhaps it's not great to explicitly mention `IsSymbol` here: it means we'd have to update multiple places if we're to do a similar instance solving in the future.  But it's possibly odd to do it for all constructor applications?

paf31: Could you please add a comment here explaining why this line is necessary?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -37,11 +37,13 @@ import Language.PureScript.TypeClassDictionaries
 import Language.PureScript.Types
 import qualified Language.PureScript.Constants as C
 
+type TypeClassDict =
+  TypeClassDictionaryInScope (Either Expr (Qualified Ident))
```
paf31: I think I'd rather use a sum type here, with a constructor for named instances, and one for symbol dictionaries. Then instead of using `either id` in `mkDictionary`, you would pattern match, and include the instance generation code there. `isSymbolDictionary` would just apply the appropriate constructor.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -103,6 +105,15 @@ data SolverOptions = SolverOptions
   -- ^ Should the solver be allowed to defer errors by skipping constraints?
   }
 
+-- | Build a type class instance dictionary for the `IsSymbol` class for a symbol
+isSymbolDictionary :: String -> TypeClassDict
```
paf31: I originally read this as a question: "is this a symbol dictionary?" 😄 

Maybe something like `makeIsSymbolDictionary` would be better?

LiamGoodacre:
Currently updating the test example so that it depends on the implicit importing.

LiamGoodacre:
Okay I'm confused: the test is passing on some build jobs and failing on others 😕.

paf31:
I think you need to update the cabal file.

paf31:
Nice work!

PR #2430, opened 2016-11-15
Add 'available' command to psc-package

andyarvanitis:
I found myself wanting this command when working on a package set. If you'd rather not have it, then feel free to just decline the PR. 

Output looks like this:
```
assert (v1.0.0p11)
bifunctors (v2.0.0)
console (v1.0.0p11)
control (v2.0.0)
prelude (v2.1.0p11)
st (v1.0.0p11)
tuples (v2.0.0)
```

### DIFF
--- a/psc-package/Main.hs
+++ b/psc-package/Main.hs
@@ -188,6 +188,15 @@ listDependencies = do
   trans <- getTransitiveDeps db depends
   traverse_ (echo . fst) trans
 
+listPackages :: IO ()
+listPackages = do
+  pkg <- readPackageFile
+  db <- readPackageSet pkg
+  traverse_ echo (fmt <$> Map.assocs db)
+  where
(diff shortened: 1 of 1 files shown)


paf31:
So this list everything in the package set, not just the things which are installed (which would be `psc-package dependencies`, minus version info)?

I'm fine with adding this, but maybe with a different name?

andyarvanitis:
@paf31  Yeah, exactly, a list of everything available in the set. Sorry, somehow my initial PR comment got truncated while I was typing it in.

I agree that the command name is vague. Any preferences? Maybe `available` or even just `all` to make it consistent with `dependencies`?

paf31:
`available` makes sense to me. Or even something like `list-all` or `list-set`.

paf31:
Thanks!

PR #2426, opened 2016-11-11
Bump stack lts version

LiamGoodacre:
Changing this let me build on macOS Sierra - which I have been unable to do since I upgraded.

PR #2432, opened 2016-11-16
Exhaustive pattern match on JSAnnot.

tmcgilchrist:
Warnings about exhaustive pattern matches on `JSAnnotSpace`
That constructor was added as short hand for a JSAnnot containing a single whitespace.


paf31:
Looks good, thanks. Can you please update `contributors.md` too?

tmcgilchrist:
Looks like the stack lts versions aren't up to date enough :-(

The extra constructor was added in `0.6.9` https://www.stackage.org/haddock/lts-7.9/language-javascript-0.6.0.9/Language-JavaScript-Parser-AST.html

paf31:
Can this be fixed by adding something to `stack.yaml`?

hdgarrood:
AFAICT combining this with #2426 is probably what we want.

tmcgilchrist:
I believe so. Spent some time updating the two stack yaml files to more
recent versions with great success.

Is the goal to support both ghc 7.10 and 8.0?

paf31:
I think we do compile with 7.10 and 8.0 right now (see the `.travis.yml` config file for the complete list).

tmcgilchrist:
@hdgarrood It looks like in that ticket it's bumping `stack.yaml` to use lts-7.9 and GHC 8, which is perhaps not correct if you want to keep both 7.10 and 8.0 working. My last change on this branch seems to be closer to the right answer if you want to keep both versions.

hdgarrood:
My view is that ideally, purescript should build without warnings when using the snapshot specified in `stack.yaml` (and, by extension, the GHC version specified there too), and we should occasionally update that snapshot; I don't think we should worry too much about supporting different configurations / dependency versions / GHC versions simultaneously. I'm also not too worried about what that compiler version is. So I have nothing against bumping the LTS version in `stack.yaml`, even if that means we stop building with 7.10. But the changes in this branch look good to me too. I'll leave it up to others to decide how to proceed.

paf31:
I'm fine with breaking compatibility with 7.10 as long as we build with whatever is in `stack.yaml`. I think this can be merged, but please update `contributors.md` first. Thanks!

tmcgilchrist:
@paf31 done.

paf31:
Thanks!

PR #2436, opened 2016-11-18
Fix exhaustive warnings

charleso:
Not quite sure about the logic. FWIW I'm personally not a fan of fallback pattern matching, it doesn't pick up new constructors.

Would it be possible to add `-Werror` somewhere that's not the cabal file to pick these up at PR/build time?


hdgarrood:
We do add -Werror in the CI build but we generally don't try to support building the compiler with versions of things different to what stack.yaml specifies; this looks like #2432 so I think you must have a newer version of language-javascript?

charleso:
@hdgarrood Ahh, double apologies. Same thing as Tim.

hdgarrood:
No problem, thanks for the PR anyway :)

PR #2441, opened 2016-11-19
Warn for shadowed type class variables

### LINKED ISSUE #2140: No ShadowedTypeVar is raised for member type vars shadowing class vars
``` purescript
-- @shouldWarnWith ShadowedTypeVar
module Main where

class Test a where
  f :: (forall a. a -> a) -> a -> a
```

joneshf:
Should fix #2140 

The error seems less than ideal:

```
psc: No input files.
➜  purescript git:(2140-fix) stack exec psc -- examples/warning/2140.purs                   
Compiling Main
Warning found:
in module Main
at /home/joneshf/programming/purescript/examples/warning/2140.purs line 4, column 1 - line 5, column 35

  Type variable a was shadowed.

in type class declaration for Test

See https://github.com/purescript/purescript/wiki/Error-Code-ShadowedTypeVar for more information,
```

Should we add which type definition is shadowing? If so, how do we nest that properly? Or is this a fine message?

### DIFF
--- a/src/Language/PureScript/Linter.hs
+++ b/src/Language/PureScript/Linter.hs
@@ -44,10 +44,14 @@ lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDecl
 
     f :: Declaration -> MultipleErrors
     f (PositionedDeclaration pos _ dec) = addHint (PositionedError pos) (f dec)
-    f dec@(ValueDeclaration name _ _ _) = addHint (ErrorInValueDeclaration name) (warningsInDecl moduleNames dec <> checkTypeVarsInDecl dec)
-    f (TypeDeclaration name ty) = addHint (ErrorInTypeDeclaration name) (checkTypeVars ty)
-    f (TypeClassDeclaration name args _ _ decs) = addHint (ErrorInTypeClassDeclaration name) (foldMap (checkTypeVarsInDecl' (S.fromList $ fst <$> args)) decs)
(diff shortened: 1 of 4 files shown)


src/Language/PureScript/Linter.hs
```diff
@@ -46,6 +46,7 @@ lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDecl
     f (PositionedDeclaration pos _ dec) = addHint (PositionedError pos) (f dec)
     f dec@(ValueDeclaration name _ _ _) = addHint (ErrorInValueDeclaration name) (warningsInDecl moduleNames dec <> checkTypeVarsInDecl dec)
     f (TypeDeclaration name ty) = addHint (ErrorInTypeDeclaration name) (checkTypeVars ty)
+    f (TypeClassDeclaration name args _ _ decs) = addHint (ErrorInTypeClassDeclaration name) (foldMap (checkTypeVarsInDecl' (S.fromList $ fst <$> args)) decs)
```
paf31: Does it make sense to call `f` recursively on the elements of `decs` here? Then you'd get a `ErrorInTypeDeclaration` wrapper applied too.

joneshf: Probably, though I'm not following how we'd do that. :smile:

paf31: Oh, I guess we'd need to modify `f` to add an accumulator parameter then.

Either way, we should probably add an `ErrorInTypeDeclaration` somehow.

paf31:
Looks great, thanks!

PR #2442, opened 2016-11-23
[WIP] :info functionality for psci

### LINKED ISSUE #1136: "Info" functionality for psci

jFransham:
Closes #1136.

Currently the code in this branch will print type synonyms, data and newtype definitions (the former as GADTs, the latter as "classic" type definitions), summarise data definitions if you ask for info on a constructor, and print the type of functions. It will also comment each with the module it is defined in. The existing code is not production-ready as I've rewritten some pretty-printing code where some functions exist already as a helper function for the module exports printer that could be adapted and reused. Plus, my programming style is pretty pointfree-heavy which might not be considered readable (I try to structure it to be more readable than the pointful alternative but if you're not familiar with the combinators it might be rough). If we ignore this, though, I've run up against three main issues:

1. I can't figure out how to resolve names in a free-standing way (i.e. If I have a name like `map`, how do I resolve that to `Data.Functor.map`). This seems to only exist as part of the compilation code that simply mutates a module, there is no pure function to do this and the logic is complicated enough that I am worried about attempting to extract it. `:t` and `:k` get around this by just sticking the argument into a syntactically-valid place in a module, compiling it, and extracting the information after the fact. Since the argument to `:i` could be a constructor, a type, or a function, this is not really workable. I can probably get quite far by trying various syntactic positions in turn until one of them compiles, but that seems like overkill if all I want is to resolve the name.
2. I can't figure out how to resolve the classes a type implements, given an `Environment`. There's a type called `TypeClassDictionaryInScope` which seems to be related to instances, and I think I can use the `tcdInstanceTypes` field with `any (unifiesWith myType)` but I don't know what the `unifiesWith` function corresponds to. `Type` has a derived `Eq` instance so, for example, `forall a b c. a -> b -> c` and `forall a b c. c -> b -> a` would not be considered equal using `(==)` despite obviously being the same type.
3. Related to 2: I don't know how to tell if a given function is from a typeclass or not, apart from to iterate through every typeclass in the module it was imported from and check all its members for matching names. What if a module reexports a typeclass member but not the typeclass? What if the typeclass is private but not the function? Is that even allowed? Do I have to traverse the import tree to find the original definition? Would that be done for me if I solved the name resolution issue in 1?

### DIFF
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -287,16 +287,20 @@ data InfoType =
     (P.ProperName 'N.TypeName)
     [(String, Maybe P.Kind)]
 
+data InfoConstructor =
+  InfoConstructor
+    P.Type
+    -- ^ Type of constructor (as displayed in GADT form)
+    (Maybe [P.Type])
+    -- ^ Arguments to constructor (as displayed in classic form)
+
 data InfoTarget
   = IValue
       P.Type
   | IConstructor
       InfoType
       P.Type
-  | IType
-      InfoType
-      [(String, P.Type)]
-      [(Maybe P.ModuleName, P.ProperName 'N.ClassName)]
+  | IType InfoType [(String, InfoConstructor)]
   | ITypeClass
       P.TypeClassData
       [(Maybe P.ModuleName, P.ProperName 'N.TypeName)]
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -17,9 +17,12 @@ import           Prelude ()
 import           Prelude.Compat
 
 import           Data.List (intercalate, nub, sort, find, foldl')
-import           Data.Maybe (mapMaybe)
+import           Data.Maybe (mapMaybe, fromMaybe, listToMaybe)
 import qualified Data.Map as M
+import qualified Data.Text as T
+import           Data.Foldable (fold)
 
+import           Control.Applicative ((<|>))
 import           Control.Monad.IO.Class (MonadIO, liftIO)
 import           Control.Monad.State.Class
 import           Control.Monad.Reader.Class
@@ -100,6 +103,7 @@ handleCommand c _ (Expression val)          = handleExpression c val
 handleCommand _ _ (Import im)               = handleImport im
 handleCommand _ _ (Decls l)                 = handleDecls l
 handleCommand _ _ (TypeOf val)              = handleTypeOf val
+handleCommand _ _ (InfoFor ident)           = handleInfoFor ident
 handleCommand _ _ (KindOf typ)              = handleKindOf typ
 handleCommand _ _ (BrowseModule moduleName) = handleBrowse moduleName
 handleCommand _ _ (ShowInfo QueryLoaded)    = handleShowLoadedModules
@@ -138,7 +142,7 @@ handleExpression evaluate val = do
   case e of
     Left errs -> printErrors errs
     Right _ -> do
-      js <- liftIO $ readFile (modulesDir </> "$PSCI" </> "index.js")
+      js <- liftIO $ readFile (modulesDir </> tempModuleNameRaw </> "index.js")
       evaluate js
 
 -- |
@@ -234,7 +238,13 @@ handleTypeOf val = do
   case e of
     Left errs -> printErrors errs
     Right (_, env') ->
-      case M.lookup (P.mkQualified (P.Ident "it") (P.ModuleName [P.ProperName "$PSCI"])) (P.names env') of
+      case
+        M.lookup
+          (P.mkQualified
+            (P.Ident "it")
+            tempModuleName)
+          (P.names env')
+      of
         Just (ty, _, _) -> liftIO . putStrLn . P.prettyPrintType $ ty
         Nothing -> liftIO $ putStrLn "Could not find type"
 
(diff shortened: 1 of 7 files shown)


jFransham:
In the ticket that this is based on there is mention of merging `psci` and `psc-ide`, I think that's a good idea but I don't know the codebase well enough to know if that could be considered a blocker to integrating this feature. If so, then I'll have to hand off my work to someone else, that's definitely not a first-time-contributor's job to do.

kRITZCREEK:
Whether or not we merge psc-ide and psci doesn't make this code obsolete. If we want the `:info` command to work for things defined in the repl we'll need to look that information up from the `Environment` like psci does as there is no ExternsFile that psc-ide could load up.

```
> data A = A | B
> :i A
```

This is also a good example for your point 1. There is no way to determine whether `A` refers to a type, type class or data constructor here, because we don't have the usual disambiguation between types and values through the parser. `psc-ide` does the resolution by requiring the caller to specify the imported modules for the file they are working on which usually works "good enough".

src/Language/PureScript/Interactive.hs
```diff
@@ -263,6 +274,257 @@ handleKindOf typ = do
...
+      . firstJust
+      . map (\f -> f env' ident')
+      $ [ getPrintableFromNames
+        , getPrintableFromTypeclasses
+        , getPrintableFromTypes
+        , getPrintableFromConstructors
+        ]
+
+firstJust :: [Maybe a] -> Maybe a
+firstJust = foldr (<|>) Nothing
```
kRITZCREEK: I think this is just `asum`

src/Language/PureScript/Interactive.hs
```diff
@@ -263,6 +274,257 @@ handleKindOf typ = do
...
+getPrintableFromTypes :: GetPrintable
+getPrintableFromTypes env qul@(P.Qualified _ unQul) =
+  M.lookup (P.ProperName <$> qul) (P.types env) >>= go
+  where
+    ctorInfo
+      :: P.ProperName 'N.ConstructorName
+      -> [P.Type]
+      -> Maybe (P.DataDeclType, (String, InfoConstructor))
+    ctorInfo nm argTypes =
+      go' <$> M.lookup (const nm <$> qul) (P.dataConstructors env)
```
kRITZCREEK: `nm <$ qul` is a little nicer as `const nm <$> qul`

src/Language/PureScript/Interactive.hs
```diff
@@ -263,6 +274,257 @@ handleKindOf typ = do
...
+getPrintableFromConstructors env qul =
+  M.lookup (P.ProperName <$> qul) (P.dataConstructors env) >>= go
+  where
+    go
+      :: (P.DataDeclType, N.ProperName 'N.TypeName, P.Type, [N.Ident])
+      -> Maybe InfoTarget
+    go (ddt, tn, t, is) =
+      uncurry IConstructor . second (InfoConstructor t)
+      <$>
+        tyInfo (N.disqualify qul) ddt (const tn <$> qul)
```
kRITZCREEK: `<$` like above

jFransham: I wasn't familiar with `<$`, I'll change the code to use it

paf31:
> I can't figure out how to resolve names in a free-standing way

You can either compile an `Environment` and look up the names there by using the current import list, or you can use the `ExternsFile` structures directly. I would recommend the first approach. As you say, you can make the type checker do the work if you create a temporary module and compile it. It's not exactly elegant, but it works. 

> I can't figure out how to resolve the classes a type implements

Short of an brute-force search, I can't think of a way to do this either. Maybe leave this out of the first version.

> I don't know how to tell if a given function is from a typeclass or not

We could add metadata to the `Environment` to support this, which will probably be more efficient than searching through all classes.

What would you say is necessary now to get this into a state where we could merge it and call it a first version. It doesn't have to be complete; it would be great to have anything at all in PSCi for `:info` really, and this already looks like it would be quite useful.

jFransham:
So, I think that the two things that are definitely needed before initial release are

- [ ] Check for partial functions/testing code (I wrote this a month ago so I can't remember what's in there)
- [ ] Style fixes
- [ ] Resolve names - I'd say this is definitely not ready for production if it can't do this

Going forward the things that need to be changed are

- [ ] Merge the display functions from this and the display functions found in the module browse view, maybe even extracting this to a typeclass or similar (no reason to keep it messy if it's a simple refactor)
- [ ] Add metadata for resolving if a function is from a typeclass
- [ ] Add metadata for resolving the typeclasses a type implements maybe? I can implement it with brute-force search first and see if that's fast enough. Since this is always going to be used in an interactive context there's not much of a speed requirement.

paf31:
This seems to have been abandoned (maybe?), so I'll close it, but please reopen it if you plan to continue. Thanks!

PR #2446, opened 2016-11-25
Update CONTRIBUTING.md -- bower needs to be in path for tests

osa1:


paf31:
Isn't this handled [here](https://github.com/purescript/purescript/blob/master/tests/support/package.json#L4)?

LiamGoodacre:
Should we remove the sentence ```You will need `npm` and `node` on your PATH to run the tests.``` altogether?

hdgarrood:
That line in the `package.json` file is there so that `TestUtils.updateSupportCode` works even if `bower` is not on your PATH. I've just checked, though, and the tests do still require `bower` to be in your PATH because `psc-publish` requires `bower` to be in your PATH (and the tests exercise that part of `psc-publish`). So we should either decide that we're going to use this locally installed `bower` for the tests, and modify the PATH to include that folder before we start running `psc-publish` tests, or merge this (and probably get rid of that line in that `package.json` file, and modify `TestUtils.updateSupportCode` to use the global one too). I don't have a strong preference either way.

hdgarrood:
(You can check the above by running `PATH="/usr/bin" stack test` - the 'updating support code' part works, but the psc-publish part fails, with the 'bower is not installed' message.)

paf31:
Thanks!

Issue #2431, opened 2016-11-15
KindsDoNotUnify can be entirely useless


garyb:
I know we have an issue for tracking the missing positional info here, but look at the state of this:
```
Error 2 of 2:

  in module SlamData.Workspace.Card.Eval.Monad

    Could not match kind

      *

    with kind

      * -> *


  in data binding group

  See https://github.com/purescript/purescript/wiki/Error-Code-KindsDoNotUnify for more information,
  or to contribute content related to this error.
```

We could at least mention what's involved in the data binding group!

rightfold:
This would help a lot when migrating to the new Halogen. 😂

LiamGoodacre:
Could you comment with a reproducible test case?  Each one I've tried has given more info than this.

LiamGoodacre:
Found an example that works:

```purescript
data Foo = MakeFoo Bar
data Bar = MakeBar (Foo Int)
```

Gives the error:

```
  Could not match kind

    * -> *

  with kind

    *

  in data binding group
```

PR #2447, opened 2016-11-25
Update error message of `ErrorInDataBindingGroup` to include participating identifiers

LiamGoodacre:
Should help with #2431 

Given the following code:
```purescript
data Foo = MakeFoo Bar
data Bar = MakeBar (Foo Int)
```

The error message is now:
```
  Could not match kind

    * -> *

  with kind

    *

  in data binding group Bar, Foo
```

### DIFF
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -16,6 +16,7 @@ import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State.Class (MonadState(..), modify)
 import Control.Monad.Supply.Class (MonadSupply)
 import Control.Monad.Writer.Class (MonadWriter(..))
+import Control.Lens ((^..), _1, _2)
 
 import Data.Foldable (for_, traverse_)
 import Data.List (nub, nubBy, (\\), sort, group)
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/TypeChecker.hs
```diff
@@ -205,9 +206,10 @@ typeCheckAll moduleName _ = traverse go
       addDataType moduleName dtype name args' dctors ctorKind
     return $ DataDeclaration dtype name args dctors
   go (d@(DataBindingGroupDeclaration tys)) = do
-    warnAndRethrow (addHint ErrorInDataBindingGroup) $ do
-      let syns = mapMaybe toTypeSynonym tys
-      let dataDecls = mapMaybe toDataDecl tys
+    let syns = mapMaybe toTypeSynonym tys
+    let dataDecls = mapMaybe toDataDecl tys
```
paf31: Nitpick: can we remove the unnecessary `let`s here?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -205,9 +206,10 @@ typeCheckAll moduleName _ = traverse go
...
+    let bindingGroupNames = (syns^..traverse._1) ++ (dataDecls^..traverse._2)
```
paf31: I think you could use `both` here. Or just `map fst` and `map snd` to keep things simple, perhaps?

Also, does this need a `nub`?

paf31:
👍  Great, thanks!

Closes #2431

src/Language/PureScript/TypeChecker.hs
LiamGoodacre: I don't think I can use `map snd` as `dataDecls` elements are triples.
I can switch to `map (\(_, n, _) -> n) dataDecls` if you prefer?  There is similar code on the next few lines.
Using `nub` sounds good.

LiamGoodacre: sorry 4-tuples not triples

paf31: Ah, gotcha. Then this is fine as-is.

PR #2450, opened 2016-11-26
Print functional dependencies in psc-docs output, fix #2439

### LINKED ISSUE #2439: Docs on Pursuit don't show fundeps
I think this is an issue with psc-docs, hence I report it on this repo.

The declaration:

```purescript
class (Semiring d, Ord d) <= Metric a d | a -> d where
```

results in the documentation:

```purescript
class (Semiring d, Ord d) <= Metric a d where
```

Note the absence of the fundep.

soupi:
Related to #2439

### DIFF
--- a/src/Language/PureScript/Docs/Types.hs
+++ b/src/Language/PureScript/Docs/Types.hs
@@ -413,7 +413,7 @@ asDeclarationInfo = do
     "typeClass" ->
       TypeClassDeclaration <$> key "arguments" asTypeArguments
                            <*> key "superclasses" (eachInArray asConstraint)
-                           <*> keyOrDefault [] "fundeps" asFunDeps
+                           <*> keyOrDefault "fundeps" [] asFunDeps
     "alias" ->
       AliasDeclaration <$> key "fixity" asFixity
                        <*> key "alias" asFixityAlias
--- a/src/Language/PureScript/Docs/Types.hs
+++ b/src/Language/PureScript/Docs/Types.hs
@@ -122,10 +122,10 @@ data DeclarationInfo
   | TypeSynonymDeclaration [(String, Maybe P.Kind)] P.Type
 
   -- |
-  -- A type class, with its type arguments and its superclasses. Instances and
-  -- members are represented as child declarations.
+  -- A type class, with its type arguments, its superclasses and functional
+  -- dependencies. Instances and members are represented as child declarations.
   --
-  | TypeClassDeclaration [(String, Maybe P.Kind)] [P.Constraint]
+  | TypeClassDeclaration [(String, Maybe P.Kind)] [P.Constraint] [([String], [String])]
 
   -- |
   -- An operator alias declaration, with the member the alias is for and the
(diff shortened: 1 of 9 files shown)


psc-docs/Ctags.hs
```diff
@@ -1,5 +1,6 @@
 module Ctags (dumpCtags) where
 
+import Prelude
```
LiamGoodacre: Why have these been added?

soupi: Right. forgot to mention. This was added because intero gave a bunch of errors for not being able to find the functions from `Prelude` so I thought explicit is better than implicit and this might help others who try to work on this with intero in the future.

hdgarrood:
Looks great! Could you please add a test to the docs tests that includes a functional dependency, so that eg. we are testing that the serialization works for fundeps?

Also, @paf31 / @garyb, is there some other place in the compiler where it might make sense to put the conversion from the integer index representation for functional dependencies to the type variable string representation, i.e. what this PR has in `Docs.Convert.Single`? I think it's fine as-is but I just thought it might be worth considering.

soupi:
@hdgarrood sure. I'll work on it.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -413,6 +413,7 @@ asDeclarationInfo = do
     "typeClass" ->
       TypeClassDeclaration <$> key "arguments" asTypeArguments
                            <*> key "superclasses" (eachInArray asConstraint)
+                           <*> key "fundeps" asFunDeps
```
paf31: This is a breaking change for the Pursuit format, right? Can this field be optional in the JSON?

hdgarrood: Good point. I think we could use `keyOrDefault "fundeps" [] asFunDeps` here, and then JSON produced by older compilers could still be submitted to Pursuit - any docs produced by older compilers would then be missing functional dependencies, but otherwise would still work.

soupi: will do.

psc-docs/Ctags.hs
paf31: I'd rather keep these changes out of the repo if we can. This should be fixed in intero if it's really a bug.

soupi: This bug happens with `ghcid` by the way, so i'd suggest adding them these imports in a few more places just so both `intero` and `ghcid` will work. I agree that this is ugly and I can remove if you wish but the tradeoff of having them worth it in my opinion. Your call :)

paf31: Are these files special, or would we need them in all files?

soupi: All files that use `Prelude` AFAICT...

paf31: So that's every module bar a few in `psc-ide` right now. I think that's too much honestly. Can't these small patches be maintained locally instead?

soupi: It looks like changing:
```
	modified:   psc-docs/Ctags.hs
	modified:   psc-docs/Etags.hs
	modified:   psc-docs/Main.hs
	modified:   psc-docs/Tags.hs
	modified:   tests/TestPscIde.hs
	modified:   tests/TestPscPublish.hs
```

Makes ghcid work.

I can probably just do that locally. Should I remove the `import Prelude` statements?

paf31:
Thanks!

Issue #2376, opened 2016-10-08
Allow rows in instance heads, under functional dependencies


paf31:
`Newtype` instances can't be derived for newtypes around records, but we can derive instances which use `TypeEquals` in the context (see `RowsInInstanceContexts.purs` in `examples`).

paf31:
I think there is a better alternative here actually. We should allow rows in instance heads if they appear in a type argument which can be solved by a functional dependency. That way, we still avoid having different instances for different record fields, at least so long as some other type argument differs.

LiamGoodacre:
I'm having a go at this (pretty much done I think).

PR #2451, opened 2016-11-26
Add support for rows in instance head under fundeps

LiamGoodacre:
For #2376

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -726,7 +726,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
     renderSimpleErrorMessage (InvalidInstanceHead ty) =
       paras [ line "Type class instance head is invalid due to use of type"
             , markCodeBox $ indent $ typeAsBox ty
-            , line "All types appearing in instance declarations must be of the form T a_1 .. a_n, where each type a_i is of the same form."
+            , line "All types appearing in instance declarations must be of the form T a_1 .. a_n, where each type a_i is of the same form, unless the type is fully determined by other type class arguments via functional dependencies."
             ]
     renderSimpleErrorMessage (TransitiveExportError x ys) =
       paras [ line $ "An export for " <> markCode (prettyPrintExport x) <> " requires the following to also be exported: "
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -22,10 +22,10 @@ import Data.Foldable (for_, traverse_)
 import Data.List (nub, nubBy, (\\), sort, group)
 import Data.Maybe
 import qualified Data.Map as M
+import qualified Data.Set as S
 import Data.Monoid ((<>))
 import qualified Data.Text as T
 import Data.Text (Text)
-import Data.Graph as G
 
 import Language.PureScript.AST
 import Language.PureScript.Crash
@@ -125,12 +125,7 @@ addTypeClass moduleName pn args implies dependencies ds =
     modify $ \st -> st { checkEnv = (checkEnv st) { typeClasses = M.insert (Qualified (Just moduleName) pn) newClass (typeClasses . checkEnv $ st) } }
   where
     newClass :: TypeClassData
-    newClass =
-      TypeClassData { typeClassArguments    = args
-                    , typeClassMembers      = map toPair ds
-                    , typeClassSuperclasses = implies
-                    , typeClassDependencies = dependencies
-                    }
+    newClass = makeTypeClassData args (map toPair ds) implies dependencies
 
     toPair (TypeDeclaration ident ty) = (ident, ty)
     toPair (PositionedDeclaration _ _ d) = toPair d
@@ -158,33 +153,15 @@ checkDuplicateTypeArguments args = for_ firstDup $ \dup ->
 checkTypeClassInstance
   :: (MonadState CheckState m, MonadError MultipleErrors m)
   => TypeClassData
-  -> Int -- type argument index
+  -> Int -- ^ index of type class argument
   -> Type
   -> m ()
 checkTypeClassInstance cls i = check where
-
-  -- we could precompute which vars are determined and store in TypeClassData?
-  -- there's no point in working this out for every instance
-
-  -- here we build a graph of which variables contribute to determining other variables
-  --
-  -- if we have a dependency such as `a b -> c`, then we have the following contributions:
-  -- `a -> c`, `b -> c`, note that this doesn't mean that, for example, `a` implies `c`,
-  -- it just means that `c` is determined by at least `a`
-  (depGraph, _, fromKey) = G.graphFromEdges ((\(n, v) -> (n, n, nub v)) <$> M.toList contributingDeps)
-    where
-      contributingDeps = M.fromListWith (++) $ do
-        fd <- typeClassDependencies cls
-        src <- fdDeterminers fd
-        (src, fdDetermined fd) : map (, []) (fdDetermined fd)
-
-  -- do there exist any vars that contribute to `i` that `i` doesn't contribute to
-  isFunDepDetermined = case fromKey i of
-    Nothing -> False -- not mentioned in fundeps
-    Just v -> let contributesToVar = G.reachable (G.transposeG depGraph) v
-                  varContributesTo = G.reachable depGraph v
-              in any (\r -> not (r `elem` varContributesTo)) contributesToVar
-
+  -- If the argument is determined via fundeps then we are less restrictive in
+  -- what type is allowed. This is because the type cannot be used to influence
+  -- which instance is selected. Currently the only weakened restriction is that
+  -- row types are allowed in determined type class arguments.
+  isFunDepDetermined = S.member i (typeClassDeterminedArguments cls)
   check = \case
     TypeVar _ -> return ()
     TypeLevelString _ -> return ()
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -22,7 +22,7 @@ import Data.Foldable (for_, traverse_)
 import Data.List (nub, nubBy, (\\), sort, group)
 import Data.Maybe
 import qualified Data.Map as M
-import Data.Monoid ((<>))
+import Data.Monoid ((<>), Any(..))
 import qualified Data.Text as T
 import Data.Text (Text)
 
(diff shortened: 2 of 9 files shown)


src/Language/PureScript/TypeChecker.hs
```diff
@@ -280,12 +293,16 @@ typeCheckAll moduleName _ = traverse go
...
   go (d@(TypeInstanceDeclaration dictName deps className tys body)) = rethrow (addHint (ErrorInInstance className tys)) $ do
-    traverse_ (checkTypeClassInstance moduleName) tys
-    checkOrphanInstance dictName className tys
-    _ <- traverseTypeInstanceBody checkInstanceMembers body
-    let dict = TypeClassDictionaryInScope (Qualified (Just moduleName) dictName) [] className tys (Just deps)
-    addTypeClassDictionaries (Just moduleName) . M.singleton className $ M.singleton (tcdValue dict) dict
-    return d
+    env <- getEnv
+    case M.lookup className (typeClasses env) of
+      Nothing -> internalError "typeCheckAll: Encountered unknown type class in instance declaration"
```
LiamGoodacre: Am I correct in the assumption that by this point the referenced type class should already have been added to the environment?

src/Language/PureScript/Errors.hs
```diff
@@ -724,7 +724,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showWiki) e = flip evalS
     renderSimpleErrorMessage (InvalidInstanceHead ty) =
       paras [ line "Type class instance head is invalid due to use of type"
             , markCodeBox $ indent $ typeAsBox ty
-            , line "All types appearing in instance declarations must be of the form T a_1 .. a_n, where each type a_i is of the same form."
+            , line "All types appearing in instance declarations must be of the form T a_1 .. a_n, where each type a_i is of the same form. Unless the type is fully determined by other type class arguments via functional dependencies."
```
paf31: Wording nitpick: I'd use a comma or parens here rather than a period. Although I wonder if we should just link to the wiki for further details here.

src/Language/PureScript/TypeChecker.hs
paf31: Yes, I believe so.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -152,18 +153,30 @@ checkDuplicateTypeArguments args = for_ firstDup $ \dup ->
...
-  return ()
-checkTypeClassInstance m (TypeApp t1 t2) = checkTypeClassInstance m t1 >> checkTypeClassInstance m t2
-checkTypeClassInstance _ ty = throwError . errorMessage $ InvalidInstanceHead ty
+checkTypeClassInstance cls i = check where
+  -- check if the argument is determined by a single fundep group
+  isDeterminedInGroup :: FunctionalDependency -> Bool
+  isDeterminedInGroup fd = not (i `elem` fdDeterminers fd) && (i `elem` fdDetermined fd)
+
+  -- is this argument fully determined via fundeps
+  isFunDepDetermined = Just (All True) == foldMap (Just . All . isDeterminedInGroup) (typeClassDependencies cls)
```
paf31: I don't think this needs to be determined by all fundeps, just by some type variable which is _not_ itself determined. This means that variable must be solved by the solver, which means that we know the variable we care about will be determined.

Eventually, we might want to allow the variable to be determined transitively (you have an example where this would be useful), but we don't need to deal with that yet.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -152,18 +153,30 @@ checkDuplicateTypeArguments args = for_ firstDup $ \dup ->
...
+  check ty = case ty of
+    TypeVar _ -> return ()
+    TypeLevelString _ -> return ()
+    TypeConstructor ctor -> do
+      env <- getEnv
+      when (ctor `M.member` typeSynonyms env) . throwError . errorMessage $ TypeSynonymInstance
+      return ()
+    TypeApp t1 t2 -> check t1 >> check t2
+    REmpty | isFunDepDetermined -> return ()
+    RCons _ hd tl | isFunDepDetermined -> check hd >> check tl
```
paf31: You'll keep performing the check here for each `RCons`. Maybe pull this out into a `checkRow` function?

src/Language/PureScript/TypeChecker.hs
LiamGoodacre: Yeah I messed this up.  Hmm, shouldn't it be: if the type is detemined in at least one fundep and never a determiner?  For example in `class C a b c | a -> b, b -> c` we should allow rows in `c` right?  However this sounds like it goes against what you said.

LiamGoodacre: Or is that the transitive case you're talking about?  Although isn't *"just by some type variable which is not itself determined"* inherently transitive?

LiamGoodacre: No wait, ignore that last bit. I'm confusing myself 😄

src/Language/PureScript/TypeChecker.hs
LiamGoodacre: Well, it will check `if True` or `if False` each time, but only actually compute it for the first time.  I could check it up front instead though, something like calling `check` in:

```haskell
goCheck f = \case
  TypeVar _ -> return ()
  TypeLevelString _ -> return ()
  TypeConstructor ctor -> do
    env <- getEnv
    when (ctor `M.member` typeSynonyms env) . throwError . errorMessage $ TypeSynonymInstance
    return ()
  TypeApp t1 t2 -> check t1 >> check t2
  ty -> f ty

checkDetermined = \case
  REmpty -> return ()
  RCons _ hd tl -> check hd >> checkDetermined tl
  ty -> invalidInstanceHead ty

check = goCheck (if isFunDepDetermined
                 then checkDetermined
                 else invalidInstanceHead)
```

paf31: Oh right, good point.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -155,18 +155,28 @@ checkDuplicateTypeArguments args = for_ firstDup $ \dup ->
...
-  env <- getEnv
-  when (ctor `M.member` typeSynonyms env) . throwError . errorMessage $ TypeSynonymInstance
-  return ()
-checkTypeClassInstance m (TypeApp t1 t2) = checkTypeClassInstance m t1 >> checkTypeClassInstance m t2
-checkTypeClassInstance _ ty = throwError . errorMessage $ InvalidInstanceHead ty
+checkTypeClassInstance cls i = check where
+  -- is this argument fully determined via fundeps
+  isFunDepDetermined = (Any False, Any True) == foldMap determining (typeClassDependencies cls)
+    where determining fd = (Any (i `elem` fdDeterminers fd),
+                            Any (i `elem` fdDetermined fd))
```
LiamGoodacre: Not sure if this is exactly what we want, but it makes more sense than what I had before.

LiamGoodacre: Okay I think I understand what you were saying before, I was originally reading it differently.

Here are a few examples cases though and what I think the outcome is:

| fundeps | fully determined |
|-|-|
| `a -> b` | `b` |
| `a -> b, b -> a` | |
| `a -> b, b -> a, c -> a` | `a, b` |
| `a -> b, b -> c` | `b, c` |
| `a -> b, b -> a, a -> c` | `c` |

paf31: Well let's write out explicitly what we want.

We want to prevent the user from casing on types in rows (i.e. labels in rows shouldn't determine behavior).

So we can't have two _different_ instances with different rows in some type argument and every other type argument the same. 

One way to achieve that with a single functional dependency (e.g. `MonadState`, where `m -> s`) is to say that we have a functional relationship on type arguments (`m` determines `s`), so changing the result of the function forces the argument to change (if `s` differs, `m` must differ too). If the input types are the same but the outputs differ, that will lead to overlap.

In general, there must be some type argument which is forced to differ between the two instances. We can find the set of such type arguments by following functional dependencies backwards, but I think to do this properly we need a full (static) overlap check, which is why I suggest limiting things to a single functional dependency for now.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -155,18 +155,28 @@ checkDuplicateTypeArguments args = for_ firstDup $ \dup ->
...
-checkTypeClassInstance _ (TypeConstructor ctor) = do
-  env <- getEnv
-  when (ctor `M.member` typeSynonyms env) . throwError . errorMessage $ TypeSynonymInstance
-  return ()
-checkTypeClassInstance m (TypeApp t1 t2) = checkTypeClassInstance m t1 >> checkTypeClassInstance m t2
-checkTypeClassInstance _ ty = throwError . errorMessage $ InvalidInstanceHead ty
+checkTypeClassInstance cls i = check where
+  -- is this argument fully determined via fundeps
+  isFunDepDetermined = (Any False, Any True) == foldMap determining (typeClassDependencies cls)
+    where determining fd = (Any (i `elem` fdDeterminers fd),
```
paf31: Is ```i `elem` fdDeterminers fd``` enough if `fdDeterminers` has more than one thing in it?

LiamGoodacre: I'm not sure that makes sense 😕, wouldn't that mean `x -> y, y -> z` and `x -> y z` are different?  Using that algorithm:

Walkthrough of checking `z` in `x -> y z`
```
determined by: x ("determiners for that fundep")
x isn't determined by anything
therefore, c is determined
```

Walkthrough of checking `z` in `x -> y, y -> z`:
```
determined by: y ("determiners for that fundep", from: y -> z)
y is determined by: x ("determined by some other fundep")
therefore, z is not determined ???
```

I have come up with what I think is a correct general solution: I'll push a new commit in a minute for you to look at.  It uses a graph to compute contributing dependencies between variables.  We will want to move computing the determined args out of the instance check, but it's there for now whilst we discuss the algorithm.

paf31: You're right, with the approximation I wrote down, they wouldn't behave the same. In terms of a graph, I would say the correct general version is:

    There is some node (let's call it an initial node, for want of a better name, but it'll be the one which the solver cases on to make its decision) which is not determined by any fundep, and a path from that node to the one we're interested in.

In the case of fundeps with multiple inputs, it's slightly trickier:

    Working backwards from the target, every path ends in an initial node.

Either way, this means that every pair of instances with different types in the target node will be forced to differ in some initial node, or else will be overlapping.

LiamGoodacre:
Okay I've moved where we compute the determined arguments and stored the result in `TypeClassData`.  How's this look?

paf31:
@LiamGoodacre Looks great, but I'd like to go over the `isFunDepDetermined` a bit more to make sure it's doing the right thing. Maybe we can chat about it on IRC.

LiamGoodacre:
I tried to write up how I'm thinking about the algorithm in a more descriptive way.  Hope this helps get my idea across :)

Cycles in a fundep constitute arguments of the same determinacy.  That is, say `x` is determined, and there is a cycle between `x` and `y`, then `y` must also be determined.

From the point of view of calculating determinacy, we can consider cycles of arguments together.  So consider each cyclic group as a node in a DAG.  Where an edge exists between nodes A and B if an argument in A determines and argument in B.  Implicitly each argument depends on itself.  Therefore each argument must be in exactly one group.  Then the collection of initial nodes of each disconnected path is the set of determiners of everything else.  Everything else being determined.

Relating to the implementation: for an argument X, if there exists a node (in a cyclic group) that determines X, which X does not determine (isn't in the same cyclic group, or implied by the cyclic group), then X is determined.  If there doesn't exist such an argument, then the cyclic group in which X resides is at the root of a path - as such it is a determiner.

paf31:
It sounds like we might be able to phrase that in terms of the strongly-connected components? But again, things are possibly complicated by edges like `a b -> c` with multiple types in the source, since we don't have a graph any more, but a hypergraph (I think).

Actually, now that I think about it in terms of SCC, I think your algorithm must be right after all. Could you just please add a comment explaining what we discussed?

paf31:
Thanks very much!

Issue #2438, opened 2016-11-18
Unicode encoding for record keys fails for surrogate pairs


kRITZCREEK:
This piece of code type checks:

```haskell
x :: String
x = {"𝌆": "hi"}."𝌆"
```

but generates invalid JS like so:

```javascript
var x = {"\119558": "hi"}["\ud834\udf06"];
```

The issue here is that the key needs to be encoded as a surrogate pair just like the accessor is.

Just something I stumbled upon while trying to work through the String -> Text unicode issues.

michaelficarra:
I would fix this right now, but I think it's best to wait until we change the internal representation in #2434. Then the codegen is simple and obvious (and this will just work).

andyarvanitis:
Even though this ticket is closed, I'm guessing you guys know already that this is not working correctly in 0.10.3? Sorry, a bit hard to tell from all the threads in issues and PRs, so I thought I'd report it, just in case.

For this character and others requiring surrogate pairs, you always get just the replacement code pair `\ufffd\ufffd`. As others have pointed out, invalid code units in `Text` get replaced during most operations. BTW, test `2843.purs` passes due to matching `\ufffd\ufffd` on both key and accessor.

paf31:
Thanks @andyarvanitis, I wasn't aware. Is anyone still working on this? cc @kRITZCREEK @michaelficarra 

Thanks.

michaelficarra:
Yep. As soon as we get Text replaced with our own string representation, this will work. *Speaking of which*, my [branch](https://github.com/michaelficarra/purescript/tree/string-newtypes) is just about ready to go except it's rendering `module."exports"` and `$foreign."warn"(Data_Show."show"(dictShow)(a));` and I can't for the life of me figure out why. If someone can help, I can get that PR submitted right away.

paf31:
Maybe put up a PR and we can try to spot it there?

michaelficarra:
@paf31 Done. See https://github.com/purescript/purescript/pull/2488. :confounded:

PR #2418, opened 2016-11-05
Use Text over String

kRITZCREEK:
Fixing #547.

I think I overdid it with the changes inside `Make` and there still are some spots with lots of converting.

To get rid of the conversions inside the pretty printer I'm planning on porting a `Text` version of `boxes`, and the conversions inside `Docs` might need some help from @hdgarrood.

My plans for this PR:

1. Release v0.10.2
2. Fix the outstanding issues around `Make`
3. Merge this
4. Try it on lots of existing PS code and see what breaks...

Because this PR touches the entire codebase it will create lots of merge conflicts ff it stays open too long. So I'd suggest we reject it outright or work on getting it merged asap.

The commit history is a bit messy, but I think taking it file-wise is still besser than just squashing all the commits.

EDIT: on top of #2417, so it should be merged after that one

PR #2454, opened 2016-11-28
fixes #2434 and #2438: PureScript chars must be UTF-16 code units

### LINKED ISSUE #2434: Treatment of surrogate pairs in string literals
After looking at #2418 I realised that we allow invalid surrogate pairs in string literals. For example:

```purescript
module Main where

import Prelude

main :: forall e. Eff (console :: CONSOLE | e) Unit
main = do
  log "\xDFFF\xD800"
```

Produces the following:

```
��
```

That is, two copies of U+FFFD, the unicode replacement character. At the moment, valid surrogate pairs come out how you would expect them to (although I expect backends other than JS will handle this less gracefully).

I think we should almost certainly throw an error in the parser if a string literal contains an invalid surrogate pair, and also consider throwing an error in the parser if a string literal contains any escape sequence for a character in the range reserved for surrogate pairs i.e. U+D800 to U+DFFF, i.e. we should also consider throwing errors for *valid* surrogate pairs.

Ideally I think string literals in PureScript source code would be represented in the compiler as the precise values they refer to, e.g. if you wrote `"\x1F4A1"` in a PureScript source file, the compiler would represent that internally as that actual code point (using some encoding which I don't think is relevant to this discussion). Then whichever backend is being used would be responsible for writing that value out as a string literal which would be interpreted correctly by whatever language implementation will end up reading it (i.e. usually JS but potentially many others).

The reason I would vote for erroring even on valid surrogate pairs is that allowing them suggests that strings are always going to be encoded as UTF-16 at runtime, which is true for JS, but probably not for other backends. It also makes the implementation of the lexer simpler. I also think it makes more sense to put any awkward code for handling surrogate pairs in the JS code generator (as opposed to the lexer), as dealing with this kind of encoding issue in generated code will be specific to the backend you're targeting.

Relevant background info from [Wikipedia](https://en.wikipedia.org/wiki/UTF-16):

> ### U+D800 to U+DFFF
> The Unicode standard permanently reserves these code point values for UTF-16 encoding of the high and low surrogates, and they will never be assigned a character, so there should be no reason to encode them. The official Unicode standard says that no UTF forms, including UTF-16, can encode these code points.
>
> However UCS-2, UTF-8, and UTF-32 can encode these code points in trivial and obvious ways, and large amounts of software does so even though the standard states that such arrangements should be treated as encoding errors. It is possible to unambiguously encode them in UTF-16 by using a code unit equal to the code point, as long as no sequence of two code units can be interpreted as a legal surrogate pair (that is, as long as a high surrogate is never followed by a low surrogate). The majority of UTF-16 encoder and decoder implementations translate between encodings as though this were the case[citation needed] and Windows allows such sequences in filenames.

/cc @kRITZCREEK @michaelficarra @andyarvanitis

michaelficarra:
Fixes #2434 and #2438.

### DIFF
--- a/examples/failing/2434.purs
+++ b/examples/failing/2434.purs
@@ -1,0 +1,5 @@
+-- @shouldFailWith ErrorParsingModule
+module Main where
+
+x :: Char
+x = '\x10000'
--- a/examples/passing/2438.purs
+++ b/examples/passing/2438.purs
@@ -1,0 +1,8 @@
+module Main where
+
+import Control.Monad.Eff.Console (log)
+
+done :: String
+done = {"𝌆": "Done"}."𝌆"
+
+main = log done
--- a/examples/passing/StringEscapes.purs
+++ b/examples/passing/StringEscapes.purs
@@ -1,7 +1,7 @@
 module Main where
 
-import Prelude ((==), (<>), bind)
-import Test.Assert (assert)
+import Prelude ((==), (/=), (<>), bind)
+import Test.Assert (assert, assert')
 import Control.Monad.Eff.Console (log)
 
 singleCharacter = "\0\b\t\n\v\f\r\"\\" == "\x0\x8\x9\xA\xB\xC\xD\x22\x5C"
@@ -12,12 +12,15 @@ highSurrogate = "\xD834"
 lowSurrogate = "\xDF06"
 loneSurrogates = (highSurrogate <> lowSurrogate) == "\x1D306"
 outOfOrderSurrogates = (lowSurrogate <> highSurrogate) == "\xDF06\xD834"
+replacement = "\xFFFD"
+notReplacing = replacement /= highSurrogate
 
 main = do
-  assert singleCharacter
-  assert hex
-  assert decimal
-  assert surrogatePair
-  assert loneSurrogates
-  assert outOfOrderSurrogates
+  assert' "single-character escape sequences" singleCharacter
+  assert' "hex escape sequences" hex
+  assert' "decimal escape sequences" decimal
+  assert' "astral code points are represented as a UTF-16 surrogate pair" surrogatePair
+  assert' "lone surrogates may be combined into a surrogate pair" loneSurrogates
+  assert' "lone surrogates may be combined out of order to remain lone surrogates" outOfOrderSurrogates
+  assert' "lone surrogates are not replaced with the Unicode replacement character U+FFFD" notReplacing
   log "Done"
--- a/examples/passing/StringEscapes.purs
+++ b/examples/passing/StringEscapes.purs
@@ -22,5 +22,5 @@ main = do
   assert' "astral code points are represented as a UTF-16 surrogate pair" surrogatePair
   assert' "lone surrogates may be combined into a surrogate pair" loneSurrogates
   assert' "lone surrogates may be combined out of order to remain lone surrogates" outOfOrderSurrogates
-  assert' "lone surrogates are not replaced with the Unicode replacement character U+FFFD" notReplacing
+  -- assert' "lone surrogates are not replaced with the Unicode replacement character U+FFFD" notReplacing
   log "Done"
--- a/examples/passing/StringEscapes.purs
+++ b/examples/passing/StringEscapes.purs
@@ -1,6 +1,6 @@
 module Main where
 
-import Prelude ((==), bind)
+import Prelude ((==), (<>), bind)
 import Test.Assert (assert)
 import Control.Monad.Eff.Console (log)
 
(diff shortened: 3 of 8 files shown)


michaelficarra:
*Bonus:* I thought I would have to make some changes to `purescript-strings`, but it looks like the `Data.Char` functions [already wouldn't have worked](https://github.com/purescript/purescript-strings/blob/94d6526ca7c692cc6917225f68a7bcabeebf9555/src/Data/Char.js#L7-L9) for code points above `U+FFFF` and the `Data.String` functions were already treating strings as sequences of UTF-16 code units. 😎

hdgarrood:
I think we might also want to test that we aren't falling victim to `text`'s replacement of invalid scalar values; I think adding the following examples/passing/StringEscapes.purs to ought to do that:

```purescript
replacement = '\0xFFFD'
notReplacing = replacement /= highSurrogate

main = do
  [...]
  assert notReplacing
```

Unless I'm missing something and this is already covered. See the "Acceptable data" paragraph in [the Data.Text documentation](https://hackage.haskell.org/package/text-1.2.2.1/docs/Data-Text.html) for what I mean by 'replacement of invalid scalar values'.

I realise the suggestion I'm about to make would make this change significantly bigger and more awkward, but have we considered introducing a separate type for PureScript strings (I think @kRITZCREEK suggested this to me initially)? Possibly a newtype over `[Int]` or maybe `Vector Int`, and maybe also with smart constructors to ensure that every element of the list is between 0 and 0xFFFF?

hdgarrood:
It might additionally be nice to use [`assert'`](https://pursuit.purescript.org/packages/purescript-assert/2.0.0/docs/Test.Assert#v:assert%27) in the tests, i.e. the version that allows you to specify an error message, so that it's clearer which case has been broken if we later break one of these cases.

michaelficarra:
CI is failing due to what appears to be an unrelated issue introduced by #2418. I'll look into it.

@hdgarrood I like your suggestions. I'll use `assert'` and add that test, but I'll wait to hear from others before switching internal string representation from `Data.Text` to `Vector Word16`.

kRITZCREEK:
This is how far I got: 

https://github.com/kRITZCREEK/purescript/commit/b1ba4727f8b581239f4b4d3e7d6b481d8bb7305a

It's a bit of a messy change so I'd prefer if we could fix the regressions introduced by the `Text` PR with the current PR and then figure out a good `PSString` solution together. `Vector Word16` sounds about right but I think we want to give a "stringy" API on top and hide it behind smart constructors.

paf31:
Any idea why `StringEscapes.purs` is failing here?

hdgarrood:
Yeah, this the problem I anticipated earlier in the thread which led me to suggest adding that extra test. `Text` is not going to work for representing string literals in the compiler, because of `Data.Text.pack`'s replacement of invalid UTF-16 sequences. Since we're going to have to change the type for string literals to something else everywhere it's used anyway, I think we should go straight ahead with having a dedicated `PSString` type, like in @kRITZCREEK's gist. But may I suggest the following instead:

```haskell
module Data.PSString (PSString, fromString, unwrap) where

import Data.Vector.Unboxed as V

-- |
-- Strings in PureScript are sequences of UTF-16 code units, which do not
-- necessarily represent UTF-16 encoded text. For example, it is permissible
-- for a string to contain *lone surrogates,* i.e. characters in the range
-- 0xD800 - 0xDFFF which do not appear as a part of a surrogate pair.
newtype PSString = PSString (Vector Word16)
  deriving (Eq, Ord, Show)

fromString :: String -> PSString
fromString =
  foldMap (encodeSurrogates . fromEnum)
  >>> map (fromIntegral :: Int -> Word16)
  >>> V.fromList
  >>> PSString
  where
  encodeSurrogates ch
    | ch > 0xFFFF = [highSurrogate, lowSurrogate]
                    where highSurrogate = [...]
    | otherwise = [ch]

unwrap :: PSString -> [Word16] -- this is easier to deal with than Vector
unwrap (PSString str) = V.toList str
```

Then, the rest happens in the (backend-specific) code generator. For JS I think it would be:

```haskell
string :: PSString -> Text
string str = "\"" <> go str <> "\""
  where
  go = PSString.unwrap >>> foldMap encodeCodeUnit
```

and `encodeCodeUnit :: Word16 -> Text` is like the `encodeChar` in `Language.PureScript.Pretty.JS` except with the `fromEnum c > 0xFFFF` case removed.

paf31:
What's the status of this now? Is it simple enough to fix that we could include it in 0.10.3?

michaelficarra:
I am going to implement @hdgarrood's suggestion in the next few days, which should fix the remaining failing test case. It's pretty simple.

michaelficarra:
@paf31 I've disabled the test that exercised `Text`'s replacement of lone surrogates with the Unicode replacement character. This PR should now pass CI and still be a strict improvement on what's currently in `master`. I'll follow up this PR with another which adds the `PSString` type, but for now this will allow you to go ahead and make a release.

paf31:
Great, thanks!

michaelficarra:
Alright, CI passed. :relieved:

paf31:
Thanks! Could you please provide a summary for the release notes if you get a minute?

michaelficarra:
@paf31 

* Fixed #2434: moved the expansion of astral code points to UTF-16 surrogate pairs from the JS code generator to the parser
* Fixed #2438: allow astral code points in record literal keys

paf31:
Thanks!

PR #2458, opened 2016-12-03
Add overrides section to psc-package config

### LINKED ISSUE #2433: Add extra-deps section to psc-package.json
The idea is that the `update` command should pull any tags listed under `extra-deps`, and include any dependencies of those packages when retrieving things from the package set.

I don't think we should follow `extra-deps` recursively, at least not in the first version. I'm fine with requiring users to list any transitive `extra-deps` instead. Thoughts?

paf31:
Fixes #2433, allowing users to override versions or add entirely new packages.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -15,7 +15,10 @@ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND 
 PureScript uses the following Haskell library packages. Their license files follow.
 
   Glob
+  QuickCheck
   SHA
+  StateVar
+  adjunctions
   aeson
   aeson-better-errors
   aeson-pretty
@@ -31,7 +34,9 @@ PureScript uses the following Haskell library packages. Their license files foll
   auto-update
   base
   base-compat
+  base-orphans
   base64-bytestring
+  bifunctors
   binary
   blaze-builder
   blaze-html
@@ -46,16 +51,19 @@ PureScript uses the following Haskell library packages. Their license files foll
   cereal
   clock
   cmdargs
+  comonad
   conduit
   conduit-extra
   connection
   containers
+  contravariant
   cookie
   cryptonite
   data-default-class
   data-ordlist
   deepseq
   directory
+  distributive
   dlist
   easy-file
   edit-distance
@@ -65,12 +73,15 @@ PureScript uses the following Haskell library packages. Their license files foll
   fast-logger
   file-embed
   filepath
+  foldl
+  free
   fsnotify
   ghc-prim
   hashable
   haskeline
   hex
   hfsevents
+  hostname
   hourglass
   http-client
   http-client-tls
@@ -79,8 +90,12 @@ PureScript uses the following Haskell library packages. Their license files foll
   http2
   integer-gmp
   iproute
+  kan-extensions
+  keys
   language-javascript
+  lens
   lifted-base
+  managed
   memory
   mime-types
   mmorph
@@ -88,10 +103,12 @@ PureScript uses the following Haskell library packages. Their license files foll
   monad-logger
   monad-loops
   mtl
+  mwc-random
   network
   network-uri
   old-locale
   old-time
+  optional-args
   optparse-applicative
   parallel
   parsec
@@ -99,16 +116,21 @@ PureScript uses the following Haskell library packages. Their license files foll
   pem
   pipes
   pipes-http
+  pointed
+  prelude-extras
   primitive
   process
+  profunctors
   protolude
   psqueues
   random
+  reflection
   regex-base
   regex-tdfa
   resourcet
   safe
   scientific
+  semigroupoids
   semigroups
   simple-sendfile
   socks
@@ -120,16 +142,21 @@ PureScript uses the following Haskell library packages. Their license files foll
   streaming-commons
   stringsearch
   syb
+  system-fileio
+  system-filepath
   tagged
   template-haskell
+  temporary
   terminfo
   text
+  tf-random
+  these
   time
-  time-locale-compat
   tls
   transformers
   transformers-base
   transformers-compat
+  turtle
   unix
   unix-compat
   unix-time
@@ -137,6 +164,7 @@ PureScript uses the following Haskell library packages. Their license files foll
   utf8-string
   vault
   vector
+  vector-instances
   void
   wai
   wai-app-static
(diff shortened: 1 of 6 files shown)


paf31:
I'm not sure what's going on with Travis here. It seems to be pulling in an older version of the code.

LiamGoodacre:
`uninstall` is still referring to `depends` instead of `packageConfigDependencies`?

LiamGoodacre:
appveyor is really not happy :S `'stack' is not recognized as an internal or external command`

paf31:
Ok, this should be ready to review now.

paf31:
If someone could review this please, I'll make a 0.10.3 release including this and @LiamGoodacre's work on fundeps.

kRITZCREEK:
Just learned about `These` and `Align` :) Interesting! Code looks good to me, I'm just curious where the `Text` import is coming from when you're only importing `pack`. Does Aeson reexport `Text`?

hdgarrood:
Code looks good, but I'm still a bit nervous about doing this without checking version ranges, as I described in #2433. I think it might be possible to get bower to check for us whether an install plan violates any version constraints after running `psc-package install` with some overrides; that could be a good temporary solution while we're still using bower.

Is there a long-term goal for `psc-package`? Are we intending for it to be a porcelain thing that is rarely used directly, kind of like `psc` for compiling now? Will there eventually be an official higher-level tool that wraps it?

hdgarrood:
Oh also I think we should hold off on a new release until #2454 is merged.

paf31:
> I think it might be possible to get bower to check for us whether an install plan violates any version constraints after running psc-package install with some overrides

But I'm trying to avoid using Bower here 😄 

I've put some suggestions in a comment in #2433.

> Is there a long-term goal for psc-package? Are we intending for it to be a porcelain thing that is rarely used directly, kind of like psc for compiling now? Will there eventually be an official higher-level tool that wraps it?

Well, I don't want it to turn into a build tool. I have no intention of replacing Pulp at all. So ideally, yes, something would need to wrap `psc-package` for more advanced use cases. I think of it as an interface to the `.psc-package` directory, via Git.

> Oh also I think we should hold off on a new release until #2454 is merged.

Is someone working on getting it to build? I don't want to hold off a 0.10.3 release indefinitely, but if it's going to be fixed soonish, then we can wait.

paf31:
Perhaps we need to find a time to talk on IRC actually. It seems like we might have different goals, and I know others have opinions too, so we should work out a plan.

andyarvanitis:
I haven't done anything "fancy" with it yet, but I have switched to using psc-package entirely for my C++ work (I'm not really a bower user), so I'd be happy to join any discussion.

Also, I'm still using `make` for my build tool, in case anyone is wondering.

michaelficarra:
@paf31 

> Is someone working on getting it to build?

I've been working on it in my spare time. Changing the representation of strings from Text to our own PSString has proven to be a ton of work.

paf31:
@michaelficarra Thanks for the update. Do you think it makes sense to revert the `Text` changes for now and to release 0.10.3 then?

Actually, could someone please summarize the regression that happened, because I don't fully understand it yet.

paf31:
I've thought about the `overrides` issue a bit more, and I'd like to add another little note.

Originally, I didn't want to add `overrides` at all. I thought that users would use the default package sets, and simply make their own package set if the packages they needed were missing or incompatible. That would satisfy the "beginners getting started" requirement, and I still like the idea.

But perhaps that's too much of a pure approach and it maybe it's not going to be practical if people feel like constructing package sets is too much effort. So maybe the solution is to make _that_ process simpler by building tools to help modify and verify package sets.

hdgarrood:
The regression is due to switching from `String` to `Text` for representing string literals in the AST. The parser is the same as before except with a call to `Data.Text.pack` so that we get a `Text` rather than a `String`. However, `Data.Text.pack` replaces characters in the range U+D800 to U+DFFF with the replacement character (U+FFFD), whether they form valid surrogate pairs or not. So any purescript string literals that people write as surrogate pairs (eg `lightbulb = "\xd83d\xdca1"`)  will come out differently in the compiled JS output (e.g. this `lightbulb` would just come out as two U+FFFD characters). People can work around this by writing it out normally e.g. `lightbulb = "\x1f4a1"` or simply `lightbulb = "💡"`.

Additionally, in some weird cases, you might want to have a `String` that contains a sequence of code units that isn't valid UTF16-encoded text. JavaScript allows this, so PureScript should too. Until we fix this issue, though, the only way to get such a string into a PureScript program will be to use the FFI.

paf31:
What about using `String` for string literals and keeping `Text` for internal use.

michaelficarra:
@paf31 We concluded in #2434 that it would be best for the internal representation not to represent strings as sequences of code points but instead sequences of UTF-16 code units.

If you'd like to make a release, you can disable [this test](https://github.com/michaelficarra/purescript/blob/0ccb13f6687c33327b24207062e1de0c02da575d/examples/passing/StringEscapes.purs#L25) and merge #2454. Then I'll follow up with the change of the underlying data structures for strings (which I've been working on locally and believe is mostly finished).

hdgarrood:
> What about using String for string literals and keeping Text for internal use?

Now that the `Text` PR is in, switching back to `String` for string literals is probably the same amount of effort as using a dedicated type, so we might as well do it properly and use a dedicated type.

kRITZCREEK:
It's not just string literals, but also everything related to record accessors, which is why the change touches so many locations. I'm happy to do the busy work for you @michaelficarra if you write safe from/to PSString functions.

michaelficarra:
@kRITZCREEK Here's my current diff on top of #2454: https://gist.github.com/michaelficarra/8ad559526c1602bd02dde021105a5da1

If you want to take it from here, let me know in the next ~10 hours, otherwise I'll continue working on it this evening.

paf31:
So I wonder if we need `overrides` at all. They will be useful for quick changes, but the other option is to make it easier for users to modify and verify the package set locally, and then to hopefully contribute those patches upstream.

I would actually prefer to not support `overrides`, but I'm also interested in providing a good experience for users.

Maybe the solution is to leave them out for now, and work on getting more users to start using `psc-package`. 

Thoughts?

hdgarrood:
I think it would be hard to justify asking users to start using `psc-package` while the only way to depend on a library is to have it in a package set, personally.

paf31:
> I think it would be hard to justify asking users to start using psc-package while the only way to depend on a library is to have it in a package set, personally.

Yes, that's probably true. Although, without `overrides`, users can still modify the local package set (the one in `.psc-package/set`. In fact, that's possibly better in the sense that it would encourage users to send those changes back upstream (especially once #2459 gets merged).

hdgarrood:
I think a necessary feature is the ability to depend on any library, whether in a package set or not, in such a way that other users can easily run an install command to get the same set of dependencies. Would this be possible if people were modifying the local package set?

Ideally I think this feature would also extend to libraries that haven't yet been published, or libraries at a particular commit (even if there is no tag pointing to that commit). That is, the equivalent of writing something like this in bower.json:

```
  "dependencies": {
    "purescript-whatever": "https://github.com/someuser/purescript-whatever#4612ddffe13046ac23526"
  }
```
```

paf31:
So that's closer to what's implemented in this PR, except unlike Bower, there is no compatibility checking right now. And there can't be, really, because most libraries in the package set don't even have `psc-package.json` files in their repositories (some do).

With the local package set approach, it would look more like:

- I need a package which isn't in the package set
- I'll modify the package set I'm using locally
- Use `psc-package verify-set` (#2459) to make sure I didn't break anything
- If it works, fork the package set and make a PR
- Add the package I want to my own `psc-package.json` and point `psc-package` at the fork
-`psc-package` update should work

So yes, it's cumbersome, but you can still have the property that people can build your library while you're waiting for the PR to be merged, if you want (although hopefully, we'll merge those quickly if the verification step works)

hdgarrood:
What do you do if you want to depend on 3 or 4 (or more) packages that aren't in a particular package set? I can't imagine how to make that workable without a tool that can do dependency solving.

Another thing that's just occurred to me is that I am not sure that a workflow where people send PRs to package sets is going to be workable. Specifically, I think we can expect a situation to arise where an official package set receives two separate PRs, and both of them apply fine on their own, except they're incompatible together (and we might not realise this until after merging). I am pretty sure I suggested this somewhere else already but I think the way Stackage deals with this is likely to work better - i.e., people don't submit libraries at specific versions, they just submit their library, and then the curating script tries to pick the latest versions of everything and pings maintainers whenever their libraries are stopping that from happening. And if we're going to change the curation script in this way (which I do think we should), then I think we should do it sooner rather than later, because quite a fundamental change which I would guess is likely to affect how other tools use package sets.

paf31:
> What do you do if you want to depend on 3 or 4 (or more) packages that aren't in a particular package set? I can't imagine how to make that workable without a tool that can do dependency solving.

Well, I imagine you'd do it by adding one package at a time to your local package set, in dependency order, just like I built the original package set. As long as the `verify-set` step keeps passing, you can carry on.

Why do you think this needs dependency solving?

> Specifically, I think we can expect a situation to arise where an official package set receives two separate PRs, and both of them apply fine on their own, except they're incompatible together (and we might not realise this until after merging)

The only way I can think this would happen is if they shared a module name. This is the sort of thing we'd ideally add to the `verify-set` command.

> And if we're going to change the curation script in this way (which I do think we should)

We don't have a curation script right now. That's another thing which needs to be done, but I was thinking of something slightly different from what I think you're suggesting. I was thinking the script would automatically upgrade all packages via minor and patch version bumps, which could then be run through CI. It would also make a note of any major version bumps and then a human could make a PR to apply those if appropriate.

hdgarrood:
I guess it doesn't 'need' dependency solving in a strict sense, but the way I see it, we would essentially be offloading the work of dependency solving on to the person who is trying to get their project to build in the process you describe. So I guess I just view having a dependency solver to help with this as a necessary feature before we can really start recommending psc-package instead of bower (whether the solver is built in to psc-package itself or part of some other tool).

> The only way I can think this would happen is if they shared a module name.

What about if A and B both depend on C, but A needs version 1.x of C and B needs 2.x of C, and one PR adds A and C 1.x to the set, and the other PR adds B and C 2.x to the set?

paf31:
> What about if A and B both depend on C, but A needs version 1.x of C and B needs 2.x of C, and one PR adds A and C 1.x to the set, and the other PR adds B and C 2.x to the set?

I was assuming each PR would only add one package, but sure, that's a possible issue if we allow more than one.

paf31:
Closing until we can decide if this is the best approach or not.

PR #2466, opened 2016-12-06
Fix windows CI

hdgarrood:
The stack windows binary archive seems to have changed from a .zip to a
.tar.gz, which is confusing 7zip - it seems 7zip uses the file name to
guess what kind of archive it is.


paf31:
Great, thanks!

PR #2468, opened 2016-12-07
Use writeUTF8File in psc-bundle

hdgarrood:
I am hoping that this change will fix an error I saw when running psc-bundle for Pulp in the windows CI:

```
psc-bundle: pulp.js: commitBuffer: invalid argument (invalid character)
```

Unfortunately I can't test this effectively since I'm not on windows. We might also need to do something about the `putStrLn` on the next line but I'm not sure what that might be.

I believe the above error was triggered by non-ascii characters in comments in the `purescript-now` package, e.g.: https://github.com/purescript-contrib/purescript-now/blob/dcdd05d630ee4ca8973665999c05ac627712fb19/src/Control/Monad/Eff/Now.purs#L24


paf31:
👍 LGTM, thanks!

PR #2473, opened 2016-12-11
psc-package: display full path in 'packages.json does not exist' error messsage

andyarvanitis:
This is a very minor change to the user experience, but helpful for a check in my Makefile (and maybe some other scripts in the future). Not critical, though, so let me know if you'd rather not pull in the change.

With this change, doing a `psc-package sources` without having done an `update` first displays
```
.psc-package/psc-0.10.2/.set/packages.json does not exist
```

instead of
```
packages.json does not exist
```


paf31:
Makes sense, thanks!

andyarvanitis:
Thank you!

Issue #2455, opened 2016-11-30
Moving the wiki to a separate repository


hdgarrood:
I think quite a few people have suggested that we replace the wiki with a git repository where changes get reviewed. This should help us keep it well-structured and correct, and also would mean that people are more likely to trust the information contained within.

To this end, I have cloned the github wiki, written a little script to convert wiki links into regular relative Markdown links, and pushed the result to https://github.com/hdgarrood/documentation.

If everyone agrees, I suggest we proceed as follows:
* Transfer the above repository to the "purescript" organisation
* Add a link to the README in this repository, pointing to this new documentation repository
* Modify each page on the current wiki, replacing the content with a note pointing to the new location (I can automate this fairly easily),
* Restrict edits on the wiki to collaborators only (so that well-meaning people don't contribute changes to the wiki that later get forgotten about and lost when we eventually disable it).

Once this has been done, I can start cleaning it up and structuring it better. I think there's a bit of duplicated or outdated information in the wiki at the moment.

Eventually we can remove the wiki completely but I don't think we should do that for a while because of how the compiler includes links to it in error messages.

I mentioned this in IRC earlier today and some people mentioned GitBook and/or ReadTheDocs. Once the above is done, we can think about presenting this information nicely using tools/services like these; I don't want to bite off too much in one go.

paf31:
Previous discussion is at #1771 (closed, but it makes sense to reopen this, I would still like to move pages into a repo)

paf31:
The plan sounds good to me, thanks for writing this up.

paf31:
One thought though: does it make sense to have a separate repo for the error messages? Or at least a subdirectory? Those pages tend to make it hard to scan the list.

garyb:
👍 for errors subdirectory. Now it's a repo generally we can impose a bit more hierarchy I think!

Also 👍 for this generally, as discussed on IRC :smile:

hdgarrood:
Yep, I was planning to create an errors subdirectory. :)

michaelficarra:
Why a separate repo? If it was maintained under a `docs` directory in this repo, it would always be in sync with the implementation and the same git tags could be used for versioning them.

hdgarrood:
I think because there's quite a lot of information in there that isn't really to do with the compiler. Also, I'm not sure if we do want to use the same tags; often, documentation corresponding to a particular version is modified after that version is released.

paf31:
[...]

PR #2476, opened 2016-12-11
Update links to wiki

LiamGoodacre:
The compiler still displays links to all of these:
* https://github.com/purescript/purescript/wiki/Error-Code-*
* https://github.com/purescript/purescript/wiki/psci
* http://wiki.purescript.org/PSCi

However we are moving to use the [purescript/documentation](https://github.com/purescript/documentation) repo.

Also do you think it makes sense to tag releases on the documentation repo and have the compiler point at specific versions, instead of always pointing at master?

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -289,8 +289,8 @@ onTypesInErrorMessageM f (ErrorMessage hints simple) = ErrorMessage <$> traverse
   gTypeSearch (TSBefore env) = pure (TSBefore env)
   gTypeSearch (TSAfter result) = TSAfter <$> traverse (traverse f) result
 
-wikiUri :: ErrorMessage -> Text
-wikiUri e = "https://github.com/purescript/purescript/wiki/Error-Code-" <> errorCode e
+errorDocUri :: ErrorMessage -> Text
+errorDocUri e = "https://github.com/purescript/documentation/blob/master/errors/" <> errorCode e <> ".md"
 
 -- TODO Other possible suggestions:
 -- WildcardInferredType - source span not small enough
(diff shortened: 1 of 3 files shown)


hdgarrood:
:+1: Except the sentence that mentions the PureScript wiki probably wants updating too, as it's no longer a wiki.

I think we should probably keep pointing at master for the time being, since we haven't yet started supporting older versions of the compiler, and since documentation relevant to a particular release is often updated after that release is made.

LiamGoodacre:
I've updated some more references to the wiki.

Yes that makes sense, thanks for the feedback :)

paf31:
I think I would prefer it if we put the error message docs in the compiler repo, and require that anyone making error messages PRs actually update the docs at the same time.

What do we think?

@LiamGoodacre Thanks for working on this!

hdgarrood:
I don't feel strongly either way but I think it might be nice to keep all the docs (i.e. what's currently in the documentation repo) together. If we are putting error message docs in the compiler repo, might it also make sense to put everything else that's currently in the documentation repo back in the compiler repo too?

LiamGoodacre:
Adding everything into the main repo means we can start to document all new features as we implement them, not just errors.

hdgarrood:
See also [my comment](https://github.com/purescript/purescript/issues/2455#issuecomment-264167368) on this topic from earlier, in #2455.

paf31:
I don't think everything belongs in the compiler repo. Take the wiki page about meetups for example.

I think there are three possible places:

- *Compiler repo*, for docs which are relevant to some versioned part of the compiler (executables, errors, etc.)
- *Wiki*, for the few things which are specific to some external group, like online meetups or conferences
- *Documentation repo*, for everything else (this basically means everything which is a guide or article in the docs repo right now)

I don't feel strongly about this break-down at all, but I think it's one way we could do it where we minimize the amount of fuss for non-compiler related things, and keep compiler-related things in sync with the compiler.

hdgarrood:
I actually have exams in the next couple of weeks so I won't be able to devote much time to purescript things between now and Christmas. Please feel free to rearrange documentation however seems reasonable without waiting for my input in the meantime.

Just a couple of things that I'd like to add:

* I'd rather not have a wiki at all, just because it's so easy for misleading or duplicated information to appear on new wiki pages, and then stay there for a while before anyone who would know how to correct it stumbles upon it. Perhaps information regarding separate meetup groups or conferences should go elsewhere, like say, their own repositories, and we can link to those repositories where appropriate?
* If we are going to split the documentation repo up as you've suggested, perhaps it would make more sense to just put the rest on the website, purescript.org, and not have an independent documentation repo? As far as I can tell, the website does currently contain everything that you've put in the "documentation repo" category.

paf31:
Let's stick with the docs repo for now, and we can always move them into the compiler repo later, maybe once they've been tidied up a bit.

Thanks @LiamGoodacre!

Issue #919, opened 2015-02-25
User-defined Kinds


paf31:
## Simple Version

The simplest version of this allows the user to write

```
kind Nat
```

and then things like

```
foreign import data Zero :: Nat
foreign import data Succ :: Nat -> Nat
```
## Possible Extensions
- Allow sorts other than `BOX`: `kind List :: BOX -> BOX`
- Full-blown `DataKinds`.
- Singleton types including numbers (a little hairy until we have `Int`) and strings, and type-level computation for those (addition, etc.)

michaelficarra:
DataKinds with rows will be so powerful! I was longing for this a few weeks ago.

natefaubion:
If you get DataKinds, I think you'd pretty quickly want either GADTs or type families. How far can you get with DataKinds alone?

paf31:
@michaelficarra Neat :)

We could also do some neat stuff like defining an actual kind for memory regions when using `ST`:

```purescript
runST :: forall a. (forall r. ST (r :: Region) a) -> a
```

michaelficarra:
Exactly. This would be so great.

LiamGoodacre:
Here's some info about data type promotion in Haskell [8.0-latest#datatype-promotion](https://downloads.haskell.org/~ghc/8.0-latest/docs/html/users_guide/glasgow_exts.html#datatype-promotion)

Things we should discuss: (basically, should we do it in the same way)

### Syntax of promoted constructors
Haskell uses an optional single-quote prefix. For example: `'True`.  Note [this section](https://downloads.haskell.org/~ghc/8.0-latest/docs/html/users_guide/glasgow_exts.html#distinguishing-between-types-and-constructors):
> ...our experience has shown that the quote mark helps to make code more readable and less error-prone...

Perhaps we should always require the prefix.

### Restrictions on promotion
Only promote data with kinds of the form `* -> ... -> * -> *`

### Auto promotion
Should all data types be promoted (where possible) or should it require enabling per data type.

rvion:
IMO, the `'` character is not the best one:
  - overloaded meaning with `'c'` char notation
  - need to fight against automatic `'` paring in ide
  - many syntax coloring tools will be buggy

if a mandatory prefix is choosen, I'll be happer it it is an other one

jdegoes:
How about just:

```scala
kind Foo = Bar | Baz
```

& the like? If you want a type & value constructors of the same name, you can just define them separately via `data`. Seems more orthogonal that way.

[...]

PR #2484, opened 2016-12-16
Add Ordering kind with LT, EQ, GT types in Prim

LiamGoodacre:


LiamGoodacre:
Originally I was thinking:
1. Add TypeOrdering kind
2. Add CompareSymbol class to purescript-symbols
3. Solve CompareSymbol

However perhaps we should just do 1 and 3 together?  Otherwise we'd need to have a release in the middle, which seems unnecessary.

LiamGoodacre:
Maybe wait for https://github.com/purescript/purescript/issues/919 instead.

paf31:
Another option here is to use kind `*` for now, and to implement symbol comparison anyway. We can always make a breaking change in the `symbols` library when custom kinds are ready.

LiamGoodacre:
Closed in favour of https://github.com/purescript/purescript-typelevel-prelude/pull/1

PR #2486, opened 2016-12-17
Add support for user defined kinds

LiamGoodacre:
Initial work towards https://github.com/purescript/purescript/issues/919

Kind declarations:
```purescript
foreign import kind Nat
```

Import/export kinds:
```purescript
module F (kind N) where
import B (kind N)
```

Use of kinds in various places (see example/passing/ForeignKind/Lib.purs):
```purescript
foreign import data Zero :: Nat
foreign import data Succ :: Nat -> Nat
...
```

Should I also move `Symbol` to use the new `KindConstructor`?
There are probably a lot more tests I should add, but it's a start.
Any issues, things I've missed, etc?
Thanks 😄


src/Language/PureScript/Ide/Externs.hs
```diff
@@ -95,6 +95,7 @@ convertDecl P.EDValue{..} = Just $ IdeDeclValue $
   IdeValue edValueName edValueType
 convertDecl P.EDClass{..} = Just (IdeDeclTypeClass edClassName)
 convertDecl P.EDInstance{} = Nothing
+convertDecl P.EDKind{} = Nothing
```
LiamGoodacre: I wasn't sure what this was.  I'll investigate further to see what actually needs doing here.  I assume I'll need to add a new `IdeDeclaration` constructor.

rightfold:
> Should I also move `Symbol` to use the new `KindConstructor`?
I think putting it in `Prim` is less surprising when someone wants to shadow it, although this is a very specific use-case. It may also simplify some of the `Kind` case analysis in the compiler.

> Any issues, things I've missed, etc?
Kind aliases? E.g. `kind Type = *`.

src/Language/PureScript/Sugar/Names/Exports.hs
```diff
@@ -19,7 +19,7 @@ import Language.PureScript.AST
 import Language.PureScript.Crash
 import Language.PureScript.Errors
 import Language.PureScript.Names
-import Language.PureScript.Sugar.Names.Env
```
LiamGoodacre: All the lines that have changed but don't look different in this file are line-ending changes.

LiamGoodacre:
If I move `Symbol` to use `KindConstructor`, should I also add `Ordering` while I'm there and cancel https://github.com/purescript/purescript/pull/2484

src/Language/PureScript/Ide/Externs.hs
paf31: @kRITZCREEK Can you comment?

paf31:
I think `Symbol` needs to be in `Prim` since there is built-in syntax which needs it, but `Ordering` can defined in a library (and if it can, I think it should be). Maybe something like `purescript-typelevel-prelude`?

src/Language/PureScript/Kinds.hs
```diff
@@ -20,6 +22,8 @@ data Kind
   | FunKind Kind Kind
   -- | Type-level strings
   | Symbol
+  -- | A kind constructor
+  | KindConstructor (Qualified (ProperName 'KindName))
```
paf31: Could we just call this `NamedKind` or `OtherKind`? Constructor seems to imply that we can have kinds which take other kinds as arguments, which isn't true.

paf31:
Looks great, thanks!

paf31:
One more thing you'll probably need to update is `usedModules` in `ModuleDependencies.hs`.

LiamGoodacre:
@paf31 for updating `usedModules` it looks like a need to update `everythingOnTypes` to support traversing kind annotations, is that correct?

src/Language/PureScript/Kinds.hs
LiamGoodacre: Thanks, I've gone with `NamedKind`.

paf31:
Yes, that sounds right.

LiamGoodacre:
Changing `everythingOnTypes` looks really annoying :P it is used everywhere - is it acceptable to instead write an `accumKinds` like we have `accumTypes`?

LiamGoodacre:
Actually, changing `everythingOnTypes` wouldn't be enough anyway, it wouldn't pick up kinds in foreign data declarations - I think.

LiamGoodacre:
Okay so I added `accumKinds` because kinds don't just appear in types, they appear in type class declarations, foreign data declarations, etc.

paf31:
Actually, having thought about this more, I'm not sure why `usedModules` does what it does now actually. @garyb shouldn't module sort-order be determined entirely by imports now?

(in which case there would be less work to do here, so you might want to hold off until this is clarified)

garyb:
> shouldn't module sort-order be determined entirely by imports now?

Yep, I mentioned this a while ago in another PR!

paf31:
Sorry :) So does that mean we can just get rid of `usedModules` apart from the `ImportDeclaration` bit then?

LiamGoodacre:
Even if I don't change `usedModules`, the `accumKinds` function still looks useful to have?

src/Language/PureScript/Ide/Externs.hs
LiamGoodacre: Yes they have their own namespace.  What kind of changes are required?

kRITZCREEK: Oh sry I read that as though they didn't :D the changes required are in `extractSpans` in `Ide.SourceFile` where I used an `Either` to disambiguate names on the type and value level

rightfold:
Are changes to psc-docs also necessary? E.g. https://github.com/purescript/purescript/blob/master/psc-docs/Tags.hs

LiamGoodacre:
I think we need to add stuff to `Interactive/Completion` and `Sugar/BindingGroups` too.

LiamGoodacre:
Unfortunately I can't build psci and psc-ide-server locally, so I'll have to depend on CI for that.

paf31:
> Even if I don't change usedModules, the accumKinds function still looks useful to have?

Yes, I think so.

src/Language/PureScript/Ide/Types.hs
```diff
@@ -293,3 +295,10 @@ instance ToJSON PursuitResponse where
       , "package" .= package
       , "text"    .= text
       ]
+
+-- TODO: better names?
+data IdeDeclName
+  = IdeNameValue Text
+  | IdeNameType Text
+  | IdeNameKind Text
+  deriving (Show, Eq, Ord)
```
LiamGoodacre: @kRITZCREEK is this okay for replacing the use of `Either Text Text` that you mentioned?

LiamGoodacre:
Actually I don't think anything needs adding to `Interactive/Completion`, right?

src/Language/PureScript/Ide/Types.hs
kRITZCREEK: Maybe replace `Name` with `NS` or `Namespace` but other than that looks great!

src/Language/PureScript/Sugar/BindingGroups.hs
```diff
@@ -73,6 +73,7 @@ createBindingGroups moduleName = mapM f <=< handleDecls
         valueVerts = map (\d -> (d, declIdent d, usedIdents moduleName d `intersect` allIdents)) values
     bindingGroupDecls <- parU (stronglyConnComp valueVerts) (toBindingGroup moduleName)
     return $ filter isImportDecl ds ++
+             filter isExternKindDecl ds ++
```
LiamGoodacre: I'm slightly confused about this.  Without adding in this filter, does that mean that the kind declarations disappear when creating binding groups?  Also is this all that is needed here?

LiamGoodacre:
Just been converting `Symbol` to use `NamedKind` and it highlighted some stuff I'd missed :)

LiamGoodacre:
Updated the `ForeignKind` test with a kinded type annotation and I see the following error:

```
  1) Passing examples 'ForeignKind.purs' should compile and run without error
       Error found:
       in module ForeignKinds.Lib
       at /Users/liamgoodacre/git/purescript/examples/passing/ForeignKind/Lib.purs line 25, column 1 - line 25, column 22

         Could not match kind

           Nat

         with kind

           Nat


       while checking the kind of Kinded Zero
       in type synonym N0
```

Something is wrong :P

paf31:
Did you update the `unify` function for kinds?

LiamGoodacre:
@paf31 yes I believe `unifyKinds` is fine, it looks like the problem is that the kind in a `KindedType` hasn't been fully qualified.  E.g:
```
Qualified Nothing (ProperName "Nat")
Qualified (Just (ModuleName [ProperName "ForeignKinds", ProperName "Lib"])) (ProperName "Nat")
```

LiamGoodacre:
I managed to build psci locally.  Here's an example usage:

```purescript
> foreign import kind K
> foreign import data D :: K
> :k D
$PSCI.K
```

We probably don't want to show the `$PSCI`, right?  I'll have a look into this at some point today.

paf31:
Ideally not, but I don't think we should make a special case for the `$PSCI` module either. Really, this should take into account which modules are imported, but we have a separate issue for that already, so maybe we can handle it then.

src/Language/PureScript/Sugar/BindingGroups.hs
paf31: Correct

src/Language/PureScript/Kinds.hs
```diff
@@ -18,8 +20,8 @@ data Kind
   | Row Kind
   -- | Function kinds
   | FunKind Kind Kind
-  -- | Type-level strings
-  | Symbol
+  -- | A named kind
+  | NamedKind (Qualified (ProperName 'KindName))
```
paf31: I wonder if `Star` and `Bang` should be moved into `Prim` too? Eventually, I would like `Bang` to be moved into `purescript-eff`, but in order to do that, we need to introduce a name for it, other than `!` (perhaps `Effect`?). So how about we move it into `Prim` now, allow both `!` and `Effect` in the parser, and then move it out to `purescript-eff` in 0.11?

LiamGoodacre: That sounds great, I should hopefully be able to get this done tonight/tomorrow.

LiamGoodacre: In updating this I've found that I'd missed updating `primImports` in `Sugar/Names/Env.hs`.

LiamGoodacre:
I removed the `ShadowPrimKind` test as it now fails due to https://github.com/purescript/purescript/issues/2499 since I added `primKinds` to `primImports` in `Sugar/Names/Env`.

paf31:
@LiamGoodacre Looks great, but could you please see if it's possible/simple to add documentation for the kinds in `Prim` now that @hdgarrood's PR #2494 has been merged?

Thanks!

hdgarrood:
That will require changes to Language.PureScript.Docs.Types to add a new DeclarationInfo constructor for user defined kinds. As far as I can tell, I think that would be fairly straightforward.

LiamGoodacre:
@hdgarrood I had already added a new `ExternKindDeclaration` constructor to `DeclarationInfo`, so I just needed to extend what you added in your PR.

How does the "Add docs for kinds in Prim" commit look?  I haven't written that much for each one.

LiamGoodacre:
I renamed `lookupPrimKind` to `lookupPrimTypeKind` as I thought it was slightly confusing as it is really "lookup this type to find its kind" - and originally I was going to add a `lookupPrimKind` for kinds - but I can change it back if you want.

hdgarrood:
Oops :) What you've got there all looks great to me, I agree that the way that you've renamed `lookupPrimKind` is an improvement. :+1:

paf31:
@LiamGoodacre Could you please rebase this?

paf31:
Thanks very much!

PR #2489, opened 2016-12-20
Error hints in psc-ide

kRITZCREEK:
This makes psc-ide add hints to specific errors.

1. `ModuleNotFound`: If the module was parsed as a source file but doesn't show up in the output/ directory it's likely the project needs to be fully rebuilt.

2. `MissingFFIModule`: Point out where to create the FFI file

3. `HoleInferredType`: JSON describing the result of type directed search for the typed hole 

I then pretty print these hints and tack them onto the json errors at a `"psc-ide-hint"` field, which I consider a horrible hack 🤕 I'd prefer to pass the json errors on unmodified and instead add more structure to the error reporting. Just thought I'd put the code out there so we could brainstorm about a few more errors that `psc-ide` could add hints to.

### DIFF
--- a/src/Language/PureScript/Ide/Error.hs
+++ b/src/Language/PureScript/Ide/Error.hs
@@ -29,7 +29,7 @@ data PscIdeError
     | ModuleNotFound ModuleIdent
     | ModuleFileNotFound ModuleIdent
     | ParseError P.ParseError Text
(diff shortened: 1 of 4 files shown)


kRITZCREEK:
Nothing apart from the `ModuleNotFound` is specific to `psc-ide` so it might be better to move these suggestions into the compiler errors. I'll revisit this at another time.

Issue #2490, opened 2016-12-20
Kinded types can cause instance solving to fail


LiamGoodacre:
I found this whilst working on https://github.com/purescript/purescript/pull/2486

The following works as expected:
```purescript
eg = show (unit :: Unit)
```

If we annotate `Unit` with its kind, it fails:
```purescript
eg = show (unit :: (Unit :: *))
```

We get:
```purescript
No type class instance was found for

    Data.Show.Show (Unit :: *)


while inferring the type of show unit
while applying a function show
  of type forall a. (Show a) => a -> String
  to argument unit
in value declaration eg
```

LiamGoodacre:
Looks like `KindedType` is missing from various functions in `TypeChecker.Entailment`; e.g: `typeHeadsAreEqual`, `typesAreEqual`.

paf31:
Oops, thanks :)

LiamGoodacre:
Okay I've worked out a solution to this, I'll submit a PR shortly.

PR #2491, opened 2016-12-20
Unwrap KindedType when instance solving

LiamGoodacre:
For #2490

### DIFF
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -146,6 +146,7 @@ entails SolverOptions{..} constraint context hints =
     ctorModules (TypeConstructor (Qualified (Just mn) _)) = Just mn
     ctorModules (TypeConstructor (Qualified Nothing _)) = internalError "ctorModules: unqualified type name"
     ctorModules (TypeApp ty _) = ctorModules ty
+    ctorModules (KindedType ty _) = ctorModules ty
     ctorModules _ = Nothing
 
     findDicts :: InstanceContext -> Qualified (ProperName 'ClassName) -> Maybe ModuleName -> [TypeClassDict]
@@ -266,6 +267,7 @@ entails SolverOptions{..} constraint context hints =
 
             canBeGeneralized :: Type -> Bool
             canBeGeneralized TUnknown{} = True
+            canBeGeneralized (KindedType t _) = canBeGeneralized t
             canBeGeneralized _ = False
 
             -- |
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -146,6 +146,7 @@ entails SolverOptions{..} constraint context hints =
     ctorModules (TypeConstructor (Qualified (Just mn) _)) = Just mn
     ctorModules (TypeConstructor (Qualified Nothing _)) = internalError "ctorModules: unqualified type name"
     ctorModules (TypeApp ty _) = ctorModules ty
+    ctorModules (KindedType ty _) = ctorModules ty
```
LiamGoodacre: This was preventing us from loading any instances at all :P

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -266,6 +267,7 @@ entails SolverOptions{..} constraint context hints =
 
             canBeGeneralized :: Type -> Bool
             canBeGeneralized TUnknown{} = True
+            canBeGeneralized (KindedType t _) = canBeGeneralized t
```
LiamGoodacre: I wasn't sure about this one or how to test it.  Any thoughts?

paf31: Maybe something like `test x = show (x :: _ :: *)`.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -403,6 +411,8 @@ matches deps TypeClassDictionaryInScope{..} tys = do
           in all (uncurry typesAreEqual) int && go sd1 r1' sd2 r2'
         where
           go :: [(Text, Type)] -> Type -> [(Text, Type)] -> Type -> Bool
+          go [] (KindedType t1 _) [] t2                = go [] t1 [] t2
```
paf31: I think this should be

```purescript
go l (KindedType t1 _) r t2                = go l t1 r t2
```

just in case someone kind-annotates a row.

paf31:
Thanks!

src/Language/PureScript/TypeChecker/Entailment.hs
LiamGoodacre: I tried disabling it and added this to the test file:

```purescript
main = do
  when (show (unit :: Unit :: *) == "unit") (log "Done")
  let show' x = show (x :: _ :: *)
  when (show' unit == "unit") (log "Done")
```

But the test still passed.  Just about to try:

```purescript
main = do
  when (show (unit :: Unit :: *) == "unit") (log "Done")
  when (show (unit :: _ :: *) == "unit") (log "Done")
```

LiamGoodacre: Oh I see why you wrote it like that, we need no instances available.

LiamGoodacre: Okay it does work, but only if I make it a separate declaration and not in a let. :smile:

paf31:
Thanks!

PR #2498, opened 2016-12-23
Add Prim docs to the library, resolves #2494

### LINKED ISSUE #2494: Include documentation for Prim types in the library
I'm trying to address a Pursuit issue where links to primitive types are broken: https://github.com/purescript/pursuit/issues/218

Currently we have a hacky solution for Prim types where I have published a pseudo-package 'purescript-prim' which mirrors the Prim module and provides documentation. However this has turned out to be too easy to forget to update. If we could include this information in the purescript library instead, Pursuit could just use that and we wouldn't need to worry about forgetting to update it. If the Prim module were to ever change again, Pursuit could then pick up any change just by updating its dependency on the purescript library and nothing else would need to happen.

I am only imagining a very small change to Language.PureScript.Environment, the rest would happen in Pursuit. If this sounds good to people I'll send a PR.

hdgarrood:


tests/TestPrimDocs.hs
```diff
@@ -0,0 +1,27 @@
...
+import qualified Data.Map as Map
+import qualified Data.Text as T
+import qualified Language.PureScript as P
+import qualified Language.PureScript.Docs as D
+import qualified Language.PureScript.Docs.AsMarkdown as D
+
+main :: IO ()
+main = do
+  putStrLn "Test that there are no bottoms hiding in primDocsModule"
+  seq (T.pack (D.runDocs (D.modulesAsMarkdown [D.primDocsModule]))) (return ())
```
paf31: You could use `Control.Exception.evaluate` here.

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -0,0 +1,205 @@
...
+  , "[the Custom Type Errors guide](https://github.com/paf31/24-days-of-purescript-2016/blob/master/21.markdown)."
+  ]
+
+typeString :: Declaration
+typeString = primType "TypeString" $ unlines
+  [ "The TypeString type constructor renders any concrete type into a Symbol"
+  , "in a custom type error."
+  , ""
+  , "For more information, see"
+  , "[the Custom Type Errors guide](https://github.com/paf31/24-days-of-purescript-2016/blob/master/21.markdown)."
```
paf31: I wonder if we should just create a placeholder page in the docs repo for now instead of linking to "24 Days" which is more likely to become out of date.

paf31:
👍 Great, thanks!

hdgarrood:
I think @garyb's point about Partial not always throwing is a good one, I'll try to get around to addressing that soon.

PR #2502, opened 2016-12-24
Further conversions to Text in Docs modules

hdgarrood:


hdgarrood:
The majority of this was fairly obvious and mindless, I just changed the data declarations and then fixed all the compile errors. The only bits weren't were small bits in `Language.PureScript.Publish` where it calls other executables, and also bits of the Docs stuff which depend on other libraries that still use String, such as bower-json, so there are a still a few occurrences of `T.pack` and `T.unpack` dotted about. I plan to come back to this later but I think this is already a big improvement and didn't want to try to do too much in one go.

PR #2503, opened 2016-12-26
Add package information to re-exports

hdgarrood:
This gets us most of the way towards fixing the following issue:

  https://github.com/purescript/pursuit/issues/209

This change preserves backwards-compatibility in the parser by
pretending that all re-exports come from local modules (i.e. in the same
package) in JSON produced by older compilers. This is obviously not
ideal but it's most likely going to be temporary and it's better than
any alternatives I can think of.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -511,6 +511,7 @@ test-suite tests
                    aeson -any,
                    aeson-better-errors -any,
                    base-compat -any,
+                   bower-json -any,
                    boxes -any,
                    bytestring -any,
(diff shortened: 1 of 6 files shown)


hdgarrood:
Yeah, it should be possible to do that. Right now `psc-publish` uses bower.json for the name of the package you're publishing, and it gets the dependencies' package names from their file paths, so the `bower_components` path is hardcoded right now. That should be easy to change to the psc-package paths if/when we switch.

PR #2507, opened 2016-12-28
Add --node-path option to PSCi

### LINKED ISSUE #1407: Add PSCi option to customize path to Node executable
psci.exe: C:\cygwin\bin\node.exe: readCreateProcessWithExitCode: permission denied (Permission denied)

Potentially because the file it asks node to read from hasn't been closed from psci?

paf31:
Fixes #1407.

### DIFF
--- a/psci/Main.hs
+++ b/psci/Main.hs
@@ -308,7 +308,7 @@ nodeBackend nodePath nodeArgs = Backend setup eval reload shutdown
     eval :: () -> String -> IO ()
     eval _ _ = do
       writeFile indexFile "require('$PSCI')['$main']();"
-      process <- (nodePath <|>) <$> findNodeProcess
+      process <- maybe findNodeProcess (pure . pure) nodePath
       result <- traverse (\node -> readProcessWithExitCode node (nodeArgs ++ [indexFile]) "") process
       case result of
         Just (ExitSuccess, out, _)   -> putStrLn out
--- a/psci/Main.hs
+++ b/psci/Main.hs
@@ -72,10 +72,16 @@ inputFile = Opts.strArgument $
      Opts.metavar "FILE"
   <> Opts.help "Optional .purs files to load on start"
 
-nodeFlagsFlag :: Opts.Parser [String]
-nodeFlagsFlag = Opts.option parser $
+nodePathOption :: Opts.Parser (Maybe FilePath)
+nodePathOption = Opts.optional . Opts.strOption $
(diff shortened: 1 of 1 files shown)


hdgarrood:
Discussion in #1407 suggested using an environment variable instead - I don't see any reason to prefer one over the other but perhaps worth mentioning?

psci/Main.hs
```diff
@@ -293,21 +299,21 @@ browserBackend serverPort = Backend setup evaluate reload shutdown
...
+nodeBackend nodePath nodeArgs = Backend setup eval reload shutdown
   where
     setup :: IO ()
     setup = return ()
 
     eval :: () -> String -> IO ()
     eval _ _ = do
       writeFile indexFile "require('$PSCI')['$main']();"
-      process <- findNodeProcess
+      process <- (nodePath <|>) <$> findNodeProcess
```
hdgarrood: Bit of a nit, but I think I'd prefer `process <- maybe findNodeProcess pure nodePath` here, so that `findNodeProcess` doesn't get run in the case where `nodePath` is provided. That way we can be sure that `findNodeProcess` isn't going to do anything dodgy like write to the console or throw an error in a case where we don't even need it because a `nodePath` has been provided, without having to look at its definition.

paf31: Yes, good point, thanks!

paf31:
I thought about adding the environment variable, but I figure it's easily scriptable, and I couldn't remember any reason to prefer the environment variable approach.

psci/Main.hs
paf31: Updated, thanks.

rvion:
> I don't see any reason to prefer one over the other but I thought it might be worth mentioning?

I think CLI option is better because 
 - you can use it both with an env var and a CLI while env is only env
 - it's self referential, no hidden behaviour

chexxor:
Hope this isn't bikeshedding, as I think it's a reasonable concern. Just wanted to note here that "--node-path" is a kinda confusing flag name, because NODE_PATH is also a thing. NODE_PATH relates to paths to node libs whereas "--node-path" here relates to the path to the node executable. "--node" would be a better name for this flag, IMO.

``` bash
NODE_PATH=output:node_modules psci --node=~/bin/node
## Is more clear than
NODE_PATH=output:node_modules psci --node-path=~/bin/node
```

hdgarrood:
@chexxor Good point, agreed!

PR #2509, opened 2016-12-28
More robust license generator script

hdgarrood:
Changes:

* Ensure that all responses are 200 OK responses, i.e. avoids errors
  such as a 404 error page being treated as the actual license file.
* Use the license-file field of the package's Cabal file -- most
  packages simply use LICENSE as their license-file, but not all do,
  and this is safer than the previous approach of maintaining a list
  of packages that deviate from the convention of using LICENSE.
* Exits with a nonzero code if at least one dependency returned a
  non-200 response for its license file.
* Uses the license files from the actual versions of dependencies which
  we are using instead of the very latest one.
* Use http-client instead of curl (mainly for the response code checking
  but this has the nice side effect that the script should work
  basically anywhere now).

Packages who list their 'license' field as PublicDomain (at the time of
writing, this is just monad-loops) will not be treated as failures if a
license-file is not found.

I also reran the script as our LICENSE file is a bit out of date now.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -16,6 +16,8 @@ PureScript uses the following Haskell library packages. Their license files foll
 
   Glob
   SHA
+  StateVar
+  adjunctions
   aeson
   aeson-better-errors
   aeson-pretty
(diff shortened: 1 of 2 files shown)


hdgarrood:
Once this is merged it might be worth amending the CI script so that the license-generator is run as part of the CI script - if the result of running the script doesn't match what's in the LICENSE file, we could make that cause a CI failure.

paf31:
Sounds like a good idea.

PR #2510, opened 2016-12-28
Add command to apply minor package updates

paf31:
This command checks for new tags on all repos and can also apply minor updates automatically. I've used it to update the package set and also find all those packages which are currently not on the latest major version.

### DIFF
--- a/psc-package/Main.hs
+++ b/psc-package/Main.hs
@@ -7,6 +7,7 @@
 
 module Main where
 
+import qualified Control.Foldl as Foldl
 import qualified Data.Aeson as Aeson
 import           Data.Aeson.Encode.Pretty
 import           Data.Foldable (fold, for_, traverse_)
@@ -15,9 +16,11 @@ import qualified Data.Map as Map
 import           Data.Maybe (fromMaybe, mapMaybe)
 import qualified Data.Set as Set
 import           Data.Text (pack)
+import qualified Data.Text as T
+import           Data.Text.Encoding (encodeUtf8)
 import qualified Data.Text.Lazy as TL
 import qualified Data.Text.Lazy.Builder as TB
-import           Data.Text.Encoding (encodeUtf8)
+import qualified Data.Text.Read as TR
 import           Data.Traversable (for)
 import           Data.Version (showVersion)
 import qualified Filesystem.Path.CurrentOS as Path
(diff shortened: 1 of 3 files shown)


psc-package/Main.hs
```diff
@@ -233,6 +254,73 @@ exec exeName = do
...
+            Just tag ->
+              case parseVersion tag of
+                Just parts -> pure parts
+                _ -> Nothing
+            _ -> Nothing
+        _ -> Nothing
+
+    parseVersion :: Text -> Maybe [Int]
+    parseVersion ref =
+      case T.stripPrefix "v" ref of
```
hdgarrood: Alternatively, we could change psc-publish to require the leading 'v'; I think most people do include it.

paf31: Thanks!
Updated to require the `v` prefix in `psc-publish`. How does this look now?

src/Language/PureScript/Publish/ErrorsWarnings.hs
```diff
@@ -148,9 +148,8 @@ displayUserError e = case e of
             , "version."
             ])
         , spacer
-        , para "Note: tagged versions must be in one of the following forms:"
+        , para "Note: tagged versions must be in the following form:"
```
hdgarrood: Would it make sense to remove the bullet point now that there's only one form listed?

psc-package/Main.hs
hdgarrood: Looks good. I think it will be great improvement to have a one-to-one correspondence between versions and expected forms of tags (in fact I probably would have done it that way in the first place if I'd thought about it a bit more).

I think we should also remove the `pkgVersionTag` from the definition of `Package` in `Docs.Types`, but continue adding the field in the `ToJSON` instance for now for backwards compatibility. We can add a TODO comment to remove it the next time we break the JSON format.

src/Language/PureScript/Publish/ErrorsWarnings.hs
paf31: Yes, of course, thanks :)

psc-package/Main.hs
paf31: Sounds good, but should that be part of this PR?

hdgarrood: Good point; no, I think we could do that separately afterwards.

src/Language/PureScript/Publish/ErrorsWarnings.hs
paf31: Updated, how's this?

PR #2511, opened 2016-12-28
Solving CompareSymbol and AppendSymbol

LiamGoodacre:
Opened for discussion.

I haven't been able to figure out why, but currently the tests fail to build the support libraries: http://lpaste.net/1293011848764850176

Any help working that out would be appreciated.

EDIT: building tests works - I needed to clear `.test_modules`.  However, updating test support is causing a lot of tests to now fail.


src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -138,7 +142,18 @@ entails SolverOptions{..} constraint context hints =
     solve constraint
   where
     forClassName :: InstanceContext -> Qualified (ProperName 'ClassName) -> [Type] -> [TypeClassDict]
-    forClassName _ C.IsSymbol [TypeLevelString sym] = [TypeClassDictionaryInScope (IsSymbolInstance sym) [] C.IsSymbol [TypeLevelString sym] Nothing]
+    forClassName _ C.IsSymbol [TypeLevelString sym] =
+      [TypeClassDictionaryInScope (IsSymbolInstance sym) [] C.IsSymbol [TypeLevelString sym] Nothing]
+    forClassName _ C.CompareSymbol [arg0@(TypeLevelString lhs), arg1@(TypeLevelString rhs), _] =
```
paf31: You're going to need to do some unification on the result type here, I think.

LiamGoodacre: Unification happens later using `matches` when computing `instances` in `solve:go`, right?

LiamGoodacre: and when I say later, I mean immediately after:
```haskell
            let instances =
                  [ (substs, tcd)
                  | tcd <- forClassName (combineContexts context inferred) className' tys''
                    -- Make sure the type unifies with the type in the type instance definition
                  , substs <- maybeToList (matches typeClassDependencies tcd tys'')
                  ]
```

paf31: Sorry, yes, you're right.

tests/TestCompiler.hs
```diff
@@ -197,27 +197,29 @@ runTest = P.runMake P.defaultOptions
...
   -> ([P.Module] -> IO ())
   -> IO (Either P.MultipleErrors [P.ExternsFile], P.MultipleErrors)
-compile supportExterns inputFiles check = silence $ runTest $ do
+compile supportExterns supportForeigns inputFiles check = silence $ runTest $ do
   fs <- liftIO $ readInput inputFiles
   ms <- P.parseModulesFromFiles id fs
   foreigns <- inferForeignModules ms
   liftIO (check (map snd ms))
-  let actions = makeActions foreigns
+  let actions = makeActions (foreigns `M.union` supportForeigns)
```
LiamGoodacre: So here is where we `makeActions`.  This makes a `codegen` function in `Make.hs` which checks for an FFI file using the `foreigns` argument passed to `makeActions`.  But we were only passing in the foreign files for the modules of the test input files (I think).  So I've passed through the FFI support files too.  The tests started passing once I did this.

paf31:
Thanks!

Issue #1875, opened 2016-02-07
Derive Functor instances


paf31:
See https://mail.haskell.org/pipermail/haskell-prime/2007-March/002137.html.

LiamGoodacre:
:+1: :+1:

phadej:
I'd try to solve this in `TemplateHaskell` / reflection way. I.e. that one could write PureScript to generate PureScript code. So the instance deriving code could live in the user land.

jdegoes:
> I'd try to solve this in TemplateHaskell / reflection way. I.e. that one could write PureScript to generate PureScript code. So the instance deriving code could live in the user land.

If that becomes possible at some point, then we can also do deriving for contravariant functors, invariant functors, and profunctors.

paf31:
Agreed, that would be ideal. I will propose Template PureScript as a GSOC project again this year.

jdegoes:
@paf31 Good idea! Are you tracking possible projects yet? I have a few ideas.

paf31:
[Yes](https://github.com/purescript/gsoc)

Let's please discuss projects on IRC though. I want to come up with a short, focused list of ideas, especially since we'll probably get at most two slots.

damncabbage:
Please pardon the thread necromancy. 😅

So currently deriving for Eq, Ord and Generic are currently checking for specific classes exposed by modules named in a particular way, for example as per [Eq from `Language.PureScript.Sugar.TypeClasses.Deriving`](https://github.com/purescript/purescript/blob/144dd6f020167d2a21095e315afc5e60b3c2bc2f/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs#L47):
```haskell
dataEq :: ModuleName
dataEq = ModuleName [ ProperName "Data", ProperName "Eq" ]

-- ...
 | className == Qualified (Just dataEq) (ProperName "Eq")
-- ...
```

If a `derive instance …` for Functor (and maybe Traversable and Foldable) were to be added, would you do similarly (checking for `class Functor` from `Data.Functor` specifically), or would you want a different approach taken?

paf31:
I think it would use the same approach.

damncabbage:
Got it; thanks. 😄

PR #2515, opened 2016-12-30
Add naive functor deriving

LiamGoodacre:
This is very basic support for deriving `Functor` in special cases: It assumes that all type constructors of kind `* -> *` applied to a type, which includes the index, implement `Functor`.  You can see examples in `example/passing/DerivingFunctor.purs`:

It doesn't look at any other argument position, thus the following is not yet supported:

```purescript
data X a = X (Either a Int)
```

Eventually I see this analysing type argument variance to work out which class (`Bifunctor`, `Profunctor`, `Contravariant`, etc) it should use when transforming.  Once we have that, we could extend it to deriving those other kinds of functor too.  Also it could probably be generalised to some kind of structural deriving thing that we can use to implement deriving `Foldable`, `Bifoldable`, `Traversable`, etc.

/cc @damncabbage for discussion :)

### DIFF
--- a/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
+++ b/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
@@ -6,7 +6,7 @@ module Language.PureScript.Sugar.TypeClasses.Deriving (deriveInstances) where
 import Prelude.Compat
 
 import Control.Arrow (second)
-import Control.Monad (replicateM)
+import Control.Monad (replicateM, zipWithM)
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.Supply.Class (MonadSupply)
 
@@ -740,8 +740,8 @@ deriveFunctor mn ds tyConNm = do
     mkCtorClause :: Text -> Ident -> (ProperName 'ConstructorName, [Type]) -> m CaseAlternative
     mkCtorClause iTyName f (ctorName, ctorTys) = do
       idents <- replicateM (length ctorTys) (freshIdent "v")
-      let args = zipWith transformArg idents ctorTys
-          ctor = Constructor (Qualified (Just mn) ctorName)
+      args <- zipWithM transformArg idents ctorTys
+      let ctor = Constructor (Qualified (Just mn) ctorName)
           rebuilt = foldl App ctor args
           caseBinder = ConstructorBinder (Qualified (Just mn) ctorName) (VarBinder <$> idents)
       return $ CaseAlternative [caseBinder] (Right rebuilt)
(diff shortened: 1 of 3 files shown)


paf31:
I would actually prefer that we _don't_ track constructor variance for the purposes of deriving, and instead recommend users use newtypes like `Data.Bifunctor.Join` to derive more interesting functors.

I'll review this later, thanks!

Fixes #1875

LiamGoodacre:
Should we add explicit support for records?  E.g:

```purescript
newtype F t = F { foo :: String, bar :: t }
derive instance functorF :: Functor F
```
It feels a little odd to do, but given that we have record syntax and it's a common way to write newtypes, perhaps we should make an exception?

Maybe later as an extension if it becomes annoying that we don't do it.

paf31:
Yes, I think we probably will need support for records.

paf31:
Bear in mind, deriving functor is a cheap way to get something like a semantic-editor-combinator for a single special field, so it'll be super useful to be able to do that for records.

paf31:
Quick example:

```purescript
data Expr a 
  = Lit { value :: Int
        , annotatoion :: a
        }
  | Add { left :: Expr annot
        , right :: Expr annot
        , annotatoion :: a
        }
  | ...

derive instance functorExpr :: Functor Expr 
```

LiamGoodacre:
I've got records working locally, but I'm not happy with how I've written it: so I'm redoing it 😄

LiamGoodacre:
Currently if you try to derive functor for something we don't support you'll see something like:

```purescript
data X a = X (Either a Int)
derive instance functorX :: Functor X
```

```
[1/1 TypesDoNotUnify] src/Main.purs:15:1

  15  derive instance functorX :: Functor X
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  Could not match type

    a2

  with type

    b0

  while trying to match type X t1
    with type X b0
  while checking that expression X $v11
    has type X b0
  in value declaration functorX

  where b0 is a rigid type variable
        a2 is a rigid type variable
        t1 is an unknown type
```

This is pretty rubbish.

However, for a kind error:

```purescript
data F f = F (f Int)
derive instance functorF :: Functor F
```

```
[1/1 KindsDoNotUnify] src/Main.purs:14:1

  14  derive instance functorF :: Functor F
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  Could not match kind

    Prim.Type

  with kind

    Prim.Type -> Prim.Type

  while checking the kind of Functor F
  in value declaration functorF
```

This one isn't too bad.

I'll have a look into making these nicer.

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -702,3 +714,68 @@ decomposeRec :: Type -> [(Text, Type)]
...
+              buildRecord updates = do arg <- freshIdent "o"
+                                       let argVar = mkVar arg
+                                           mkAssignment (l, x) = (l, App x (Accessor l argVar))
+                                       return (lam arg (ObjectUpdate argVar (mkAssignment <$> updates)))
+
+          -- under a `* -> *`, just assume functor for now
+          goType (TypeApp _ t) = fmap (App mapVar) <$> goType t
+
+          -- otherwise do nothing - will fail type checking if type does actually contain index
+          goType _ = return Nothing
```
paf31: What happens here with something like `data F a = F String`?

paf31:
Great stuff, thanks!

PR #2516, opened 2016-12-31
Derive instances when data types use type synonyms

### LINKED ISSUE #1443: Fail to derive Generic with record argument referenced through type synonym
``` purescript
type MyArray   = Array String
type MyString  = String
type MyNumber  = Number
type MyInt     = Int
type MyBoolean = Boolean
type MyProduct = Maybe String
type MyRecord  = { x :: Number, y :: Number }

data Foo = Foo MyArray   -- fine
               MyString  -- fine
               MyNumber  -- fine
               MyInt     -- fine
               MyBoolean -- fine
               MyProduct -- fine
               MyRecord  -- FAIL
derive instance fooGeneric :: Generic Foo
```

fails with

``` shell
No instance found for Data.Generic.Generic MyModule.MyRecord<>
```

### LINKED ISSUE #2416: Allow `Newtype` deriving for newtypes that use type synonyms

### LINKED ISSUE #2481: Allow `derive newtype instance` for newtypes that use type synonyms
My minimal reproduction is
```
type MyString = String
newtype WrappedString = WrappedString MyString
derive newtype instance semigroupWrappedString :: Semigroup WrappedString
```
which produces the error
```

      Data.Semigroup.Semigroup MyString


  while checking that expression #dict Semigroup
    has type { append :: WrappedString -> WrappedString -> WrappedString
             }
  in value declaration semigroupWrappedString
```
even though 
```
newtype WrappedString = WrappedString String
derive newtype instance semigroupWrappedString :: Semigroup WrappedString
```
compiles without error.

(Originally posted incorrectly to #2416, but this is a different bug.)

paf31:
Fixes #2481 
Fixes #2416
Fixes #1443

I would have liked to fix this by moving deriving into the type checker, but then we'd have to rearrange a whole bunch of stuff, so instead this just uses the externs to build a map of type synonym data, and then reuses the existing `replaceAllTypeSynonyms'` function to eliminate any synonyms in the instance head.

### DIFF
--- a/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
+++ b/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
@@ -3,43 +3,71 @@
 --
 module Language.PureScript.Sugar.TypeClasses.Deriving (deriveInstances) where
 
-import Prelude.Compat
+import           Prelude.Compat
 
-import Control.Arrow (second)
-import Control.Monad (replicateM)
-import Control.Monad.Error.Class (MonadError(..))
-import Control.Monad.Supply.Class (MonadSupply)
-
-import Data.List (foldl', find, sortBy, unzip5)
-import Data.Maybe (fromMaybe)
-import Data.Ord (comparing)
-import Data.Text (Text)
-
-import Language.PureScript.AST
-import Language.PureScript.Crash
-import Language.PureScript.Environment
-import Language.PureScript.Errors
-import Language.PureScript.Names
-import Language.PureScript.Types
-import Language.PureScript.TypeChecker (checkNewtype)
+import           Control.Arrow (second)
+import           Control.Monad (replicateM)
+import           Control.Monad.Error.Class (MonadError(..))
+import           Control.Monad.Supply.Class (MonadSupply)
+import           Data.List (foldl', find, sortBy, unzip5)
+import qualified Data.Map as M
+import           Data.Maybe (fromMaybe, mapMaybe)
+import           Data.Ord (comparing)
+import           Data.Text (Text)
+import           Language.PureScript.AST
+import           Language.PureScript.Crash
+import           Language.PureScript.Environment
+import           Language.PureScript.Errors
+import           Language.PureScript.Externs
+import           Language.PureScript.Kinds
+import           Language.PureScript.Names
+import           Language.PureScript.Types
+import           Language.PureScript.TypeChecker (checkNewtype)
+import           Language.PureScript.TypeChecker.Synonyms (replaceAllTypeSynonyms')
 import qualified Language.PureScript.Constants as C
+
+type SynonymMap = M.Map (Qualified (ProperName 'TypeName)) ([(Text, Maybe Kind)], Type)
 
 -- | Elaborates deriving instance declarations by code generation.
 deriveInstances
-  :: (MonadError MultipleErrors m, MonadSupply m)
-  => Module
+  :: forall m
+   . (MonadError MultipleErrors m, MonadSupply m)
+  => [ExternsFile]
+  -> Module
   -> m Module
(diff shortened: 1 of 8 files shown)


src/Language/PureScript/TypeChecker/Synonyms.hs
```diff
@@ -5,43 +5,47 @@
...
+import           Language.PureScript.Environment
+import           Language.PureScript.Errors
+import           Language.PureScript.Kinds
+import           Language.PureScript.Names
+import           Language.PureScript.TypeChecker.Monad
+import           Language.PureScript.Types
+
+-- | Replace fully applied type synonyms
+replaceAllTypeSynonyms'
+  :: M.Map (Qualified (ProperName 'TypeName)) ([(Text, Maybe Kind)], Type)
```
garyb: Could move the `SynonymMap` type into this module, and then use it here?

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -3,43 +3,71 @@
...
+          Just (Qualified (Just mn) name, (args, ty))
+        fromLocalDecl (PositionedDeclaration _ _ d) = fromLocalDecl d
+        fromLocalDecl _ = Nothing
+
+replaceAllTypeSynonymsM
+  :: MonadError MultipleErrors m
+  => SynonymMap
+  -> Type
+  -> m Type
+replaceAllTypeSynonymsM syns = either throwError pure . replaceAllTypeSynonyms' syns
```
garyb: Maybe this should go in `Language.PureScript.TypeChecker.Synonyms` too?

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -3,43 +3,71 @@
...
 -- elaborates that into an instance declaration via code generation.
 deriveInstance
   :: (MonadError MultipleErrors m, MonadSupply m)
   => ModuleName
+  -> SynonymMap
   -> [Declaration]
   -> Declaration
   -> m Declaration
-deriveInstance mn ds (TypeInstanceDeclaration nm deps className tys@[ty] DerivedInstance)
+deriveInstance mn _ ds (TypeInstanceDeclaration nm deps className tys@[ty] DerivedInstance)
```
LiamGoodacre: I think these need to do synonym substitution too.  They inspect data constructor arguments; for example, `deriveOrd` calls `objectType` to check for a record.

LiamGoodacre: The following fail on 0.10.3:
```purescript
type L = {}
data X = X L
derive instance eqX :: Eq X

type M = {}
data Y = Y {foo :: M}
derive instance eqY :: Eq Y

type N = {}
data Z = Z N
derive instance eqZ :: Eq Z
```

paf31: Yes, I realized this last night after I finished working on this 😄 

I'll try to finish this up later.

paf31:
This is ready for review again, those issues should now be fixed.

LiamGoodacre:
Looks good to me 👍

paf31:
Ok, I'll merge this since @LiamGoodacre has signed off and the tests are passing.

esad:
This still doesn't work for record aliases. I can't get the @garyb example to compile with v0.10.5, it fails with `No type class instance was found for Data.Generic.Generic { "baz" :: { "foo" :: String } }`.

paf31:
That's a different issue though. It means that we don't derive `Generic` for nested records (yet).

esad:
@paf31 This minimal, non-nested example fails too:

```
type Baz = { baz :: String }
newtype T = T Baz

derive instance genericT :: Generic T
```

(`No type class instance was found for Data.Generic.Generic { "baz" :: String }`)

paf31:
Then that is a bug, thanks for the report 😄

PR #2521, opened 2017-01-01
Remove Stackage badges

kRITZCREEK:
Now that we no longer are in stackage I think we should remove these. LTS 2 and 3 were horribly outdated anyway.


hdgarrood:
Can't find a way to approve on mobile but :+1:

Issue #2394, opened 2016-10-20
Functional dependencies can make instances unusable


garyb:
While trying to get `-parallel` into shape I tried to move `ParCont` out of the `Class` module, as it doesn't look like it has any reason to be there - it could depend on the `Class` module. However, when doing so, the tests start failing:

```
Control.Parallel.Class.Parallel t0
                                    (ContT Unit
                                       (Eff
                                          ( http :: HTTP
                                          , console :: CONSOLE
                                          | eff1
                                          )
                                       )
                                    )
```

I found this very surprising, and I'm pretty sure it's not right - instances are supposed to be global, so it shouldn't matter if the `Parallel (ParCont m) (ContT Unit m)` instance is defined alongside the `Parallel` class or not.

paf31:
I'll try to have a look at this later today.

paf31:
What's going on here is that we only check modules which are mentioned in the instance head. So in the example you gave, we will check the modules which define `Parallel`, `ContT`, `Eff`, `Unit`, `HTTP`, and `CONSOLE`. But the instance for `ParCont` is not defined in any of those!

So I think that we need to modify our orphan instance rule so that instances like the one above are considered orphaned if they are not defined in one of those modules.

garyb:
Since there's a bidirectional fundep here that still doesn't seem right to me - I would have thought we could use that to resolve the instance.

Assuming there's some problem with that I'm not seeing, how would we even determine this is an orphan? By which I mean, I'm not sure what the rule would be, since instance head definition doesn't suggest orphan in any way, it only seems to arise during inference from one of its members.

paf31:
How does the bidirectional fundep help though? The problem is that we don't look in the right module, because we have no hints from the type as to where it might be.

The orphan rule has to take into account any minimal covering sets for the instance head. In this case, there are two sets, the one including only the first argument, and the one including only the second. We have to take the intersection of the modules inferred by those minimal sets. The only module used by the first would be the module which `ParCont` is defined in. The modules for the second set are the ones which define `ContT`, `Eff`, `Unit`, `HTTP`, and `CONSOLE`. The intersection is empty, so the only solutions are to put the instance with the class or to move `ParCont` into one of the modules listed above.

natefaubion:
> How does the bidirectional fundep help though?

I think intuitively you could say that in this case you should use the union of modules inferred by those sets, since the variables determine each other.

paf31:
[...]

PR #2522, opened 2017-01-01
Update orphan instance check to use covering sets

LiamGoodacre:
For https://github.com/purescript/purescript/issues/2394
Let me know if you think of any more tests, want me to comment more with explanations, etc.

I think the error message for `OrphanInstance` also wants updating - I'll have a look at that shortly.

### DIFF
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -18,7 +18,7 @@ import Control.Monad.Supply.Class (MonadSupply)
 import Control.Monad.Writer.Class (MonadWriter(..))
 import Control.Lens ((^..), _1, _2)
 
-import Data.Foldable (for_, traverse_)
+import Data.Foldable (for_, traverse_, toList)
 import Data.List (nub, nubBy, (\\), sort, group)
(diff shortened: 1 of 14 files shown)


src/Language/PureScript/TypeChecker.hs
```diff
@@ -324,19 +324,31 @@ typeCheckAll moduleName _ = traverse go
...
+
+    modulesByTypeIndex :: M.Map Int (Maybe ModuleName)
+    modulesByTypeIndex = M.fromList (zip [0 ..] (typeModule <$> tys'))
+
+    lookupModule :: Int -> S.Set ModuleName
+    lookupModule idx = case M.lookup idx modulesByTypeIndex of
+      Just ms -> S.fromList (toList ms)
+      Nothing -> internalError "Unknown type index in checkOrphanInstance"
+
+    nonOrphanModules :: S.Set ModuleName
```
paf31: Could you please add a short comment explaining why we take the intersection here?

LiamGoodacre:
(I already added a comment; perhaps I should have made a new commit instead of amending, so it was more obvious there was a change.)

PR #2524, opened 2017-01-01
'kind' is no longer a reserved word

paf31:
This removes `kind` from the reserved word list, so that we can continue to use the word `kind` as an `Ident` (e.g. in Halogen and `purescript-dom`).

Before 0.11, we should decide whether we want to make `kind` a reserved word or not.

After this is merged, I'll make a 0.10.4 release. I will probably mark the user-defined kinds feature as experimental in the release notes for now.

cc @LiamGoodacre

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -110,7 +110,7 @@ parseValueDeclaration = do
 
 parseExternDeclaration :: TokenParser Declaration
 parseExternDeclaration = reserved "foreign" *> indented *> reserved "import" *> indented *> parseExternAlt where
-  parseExternAlt = parseExternData <|> parseExternKind <|> parseExternTerm
+  parseExternAlt = parseExternData <|> P.try parseExternKind <|> parseExternTerm
 
   parseExternData = ExternDataDeclaration <$> (reserved "data" *> indented *> typeName)
                                           <*> (indented *> doubleColon *> parseKind)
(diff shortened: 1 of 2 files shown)


hdgarrood:
Could you perhaps explain a little more of the context of this? What are the advantages and disadvantages of each approach?

garyb:
> Could you perhaps explain a little more of the context of this? What are the advantages and disadvantages of each approach?

Reserving `kind` is a breaking change, meaning we shouldn't release things as they are as 0.10.4, as there are several libraries in the wild that would suffer for this.

As for advantages to reserving, I'm not sure?

paf31:
Yes, what @garyb said. Here is the discussion from IRC:

```text
4:54 PM <paf31> we could just say it's a breaking change, but it's potentially quite a bad one for a minor compiler release
4:54 PM we could revert it, but I don't want to
4:55 PM and we could change the syntax, but I can't think of anything better

...

5:03 PM <natefaubion> paf31: is the issue just `foreign import kind Foo` and `foreign import kind :: …`?
5:04 PM <LiamGoodacre> and import/export: `import F (kind Foo)`
5:04 PM <paf31> Phillip Freeman kind is a keyword now
5:04 PM so you can't define a function called kind, for example
5:04 PM which ps-dom does, and so does react
5:04 PM <natefaubion> those are all unambiguous though? I know it’s not _great_, but you can make it compatible and issue warnings?
5:05 PM <paf31> true
5:05 PM LiamGoodacre: what do you think?
5:07 PM L<LiamGoodacre> I think natefaubion's idea is a fair solution, I can't think of anything nicer really
```

Issue #1253, opened 2015-07-11
Improve error message when variable name "bind" is used in a do block.


jutaro:
This gives a typechecker error, because desugaring uses the name "bind". This is a tiny little inconvenience, but I will not be the last one who wasted the time on this strange typechecker problem, since bind is (I guess) a relativ frequently used name.

michaelficarra:
Additionally, you have to import `bind` to use do blocks in 0.7. I do not believe this was the case in 0.6, since I would always use explicit imports with Prelude and never needed to do that.

nwolverson:
Error when not importing `bind` and writing a `do` block is a little confusing. The error when the `bind` of purescript-d3 was in scope instead was a lot confusing.

michaelficarra:
@nwolverson See my edit above, #962, and #742. This is actually a PureScript 0.7+ feature.

nwolverson:
Sure, I think this bug should either be renamed or closed in favour of a new one re the error message. Errors that mention identifiers not in your program are unfortunate to say the least, and this one will probably bite both those migrating and newcomers.

(I guess the redefining `bind` accidentally one is both harder to do anything about and less common than not importing it at all)

paf31:
I'm not sure what we can do about this if we want to have rebindable syntax. Whatever name we pick could be used by some function somewhere.

michaelficarra:
Yes, we should just have a better error for this occurrence.

paf31:
It's worth noting that the error here is much better now. For example,

```purescript
foo = do
  bind <- [1,2,3]
  x <- [4, 5, 6]
  pure x
```

gives

```text
  Could not match type
       
    Int
       
  with type
                   
    Array Int -> t0
                   

while applying a function bind
  of type Int
  to argument [ 4
              , 5
              , 6
              ] 
```

The key is to read the last line:

```text
while applying a function bind of type Int to argument [ 4, 5, 6 ]
```

which I think makes it much clearer what's going on.

PR #2527, opened 2017-01-02
Adds specific error message when failing to import bind

FrigoEU:
I saw #1253 as being assigned to the 1.0 release and thought it would be an easy fix. Turns the error:

```
Unknown value bind
```
into
```
Unknown value bind. You're probably using do-notation, which the compiler replaces with calls to the bind function. Please import bind from module Prelude
```

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -498,6 +498,8 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs) e = flip evalS
       line $ "The type declaration for " <> markCode (showIdent nm) <> " should be followed by its definition."
     renderSimpleErrorMessage (RedefinedIdent name) =
       line $ "The value " <> markCode (showIdent name) <> " has been defined multiple times"
+    renderSimpleErrorMessage (UnknownName name@(Qualified Nothing (IdentName (Ident "bind")))) =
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -498,6 +498,8 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs) e = flip evalS
       line $ "The type declaration for " <> markCode (showIdent nm) <> " should be followed by its definition."
     renderSimpleErrorMessage (RedefinedIdent name) =
       line $ "The value " <> markCode (showIdent name) <> " has been defined multiple times"
+    renderSimpleErrorMessage (UnknownName name@(Qualified Nothing (IdentName (Ident "bind")))) =
+      line $ "Unknown " <> printName name <> ". You're probably using do-notation, which the compiler replaces with calls to the bind function. Please import " <> markCode "bind" <> " from module " <> markCode "Prelude"
```
paf31: Nitpick: Could you please use `markCode` for the other occurrence of `bind` too?

paf31:
This looks good, but I don't think it fixes #1253 actually. That is about the case where you do something like

```purescript
foo = do
  bind <- bar
  ...
```

since then `bind` will get shadowed.

FrigoEU:
Indeed it doesn't. I'm not immediately sure how I'd solve that. This just
adds a better error message if you didn't import bind yet. I think this is
a simple fix for a very common (but small) problem.

PR #2532, opened 2017-01-02
[psc-ide] use unicode forall and arrows in completions

kRITZCREEK:
Using the unicode syntax for the reported types looks nice and has the added benefit of reducing the popup/message size.

Doing a simple string replacement is not the most robust solution, but I'm fairly certain it's good enough for our purposes. If a typevariable ending in `forall` gets altered it's easy enough to figure out what has happened for the user.


paf31:
Replacing strings seems a bit sketchy. Why not pass options to `prettyPrintType` instead?

kRITZCREEK:
You're right, done.

kRITZCREEK:
Hmm the windows failure needs a bit of encoding massaging...

paf31:
Thanks! Looks good once the encoding issue is sorted.

kRITZCREEK:
The tests pass again... I had to take explicit control over the psc-ide process for the integration tests. If this looks fine I'll squash and merge.

Issue #2186, opened 2016-06-06
Reduce frequency of breaking changes to pursuit json format


hdgarrood:
I think we are having to break the pursuit json format too often, and I would like to try modifying the format in such a way that we can reduce how often we need to break it. I can think of two options:

1. Keep an approximately equivalent approach, but reduce the number of internal compiler data types which appear inside the JSON blob that is sent, such as `Type`, `Constraint`, and `Fixity`. On a conceptual level, it seems to me that the internal representations of these in the compiler ought to be able to change without needing to regenerate everything.
2. Change the whole approach and just copy what Hackage does; namely, generate the HTML client-side and host it on Pursuit.

2) is easier, but makes it significantly easier for people to put nasty stuff on the website. content-security-policy headers should help there (aside: they seem to have disappeared), but it's still unsettling. I want to try 1) first.

garyb:
I think the churn should slow down a bit now anyway (famous last words) as we're done with the operator changes, etc. I guess there will be some changes for constraint kinds, but I'm not sure there's much else planned for 0.10 that should necessitate it?

Anything we can do that makes it less painful to update sounds good to me, although something along the lines of 1 would be my preference also.

paf31:
I don't think we should generate HTML, but rather some data structure which looks like the subset of HTML we need to use. That way, we can restyle later without having to regenerate everything.

garyb:
Markdown? Haha.

paf31:
Another upside of generating HTML or some other markup is that users can generate it for offline reading.

Markdown would work, except for that we need some way to represent syntax highlighting.

jplatte:
Markdown has code blocks. Or, if we don't want every code block to be highlighted as PureScript (which would require some custom code anyway instead of just a Markdown -> HTML converter), any converter that supports GitHub markdown would work (https://help.github.com/articles/creating-and-highlighting-code-blocks/).

paf31:
Code blocks might not be enough, since what constitutes valid code might change over time.

hdgarrood:
Markdown doesn't seem a great fit to me, given that I want some way of indicating links between modules or packages that isn't tied to the url structure that the pursuit web app uses. The current code does have this, and I'd rather keep it that way. Also, we need to be able to produce Hoogle input files from the same data.

hdgarrood:
[...]

PR #2533, opened 2017-01-02
Restore backwards compatibility for parsing Kinds

### LINKED ISSUE #2530: Breaking change to pursuit JSON format
Trying to use the current Pursuit package database with the new compiler gives you errors such as:

> When parsing Language.PureScript.Kinds.Kind expected an Object with a tag field where the value is one of [KUnknown, Row, FunKind, NamedKind], but got Bang.

In retrospect it makes sense why this would be a breaking change but it's easy to see how we missed it initially. I think the easiest way to address this might be to add a special case for kinds which were moved to Prim into the parsers in `Language.PureScript.Docs` and release it as 0.10.5; does that sound sensible?

*edit:* to clarify, it should be possible to restore backwards compatibility with this approach so that JSON generated with e.g. 0.10.3 is still usable.

hdgarrood:
Fixes #2530

This turned out to be more invasive than I thought; I can't see a less invasive way to achieve it, sadly. It's the fact that the type `Kind` is recursive that makes it more difficult - because from outside, we can only easily adjust the top-level parsing, and not the parsing of `Kind` values which are nestled inside other `Kind` values.

I've tested this locally and this parser was able to successfully parse JSON produced by both 0.10.3 and 0.10.4. There is a test for this in the psc-publish test suite already too, although it doesn't test backwards compatibility at the moment (which is why I've tested for backwards compatibility manually).

I think moving away from derived JSON instances is worth considering as something to aim for in the long term anyway, though, see #2186.


src/Language/PureScript/Kinds.hs
```diff
@@ -20,7 +23,66 @@ data Kind
...
+
+-- This is equivalent to the derived Aeson FromJSON instance, except that it
+-- also handles JSON generated by compilers up to 0.10.3 and maps them to the
+-- new representations (i.e. NamedKinds which are defined in the Prim module).
+kindFromJSON :: Parse Text Kind
+kindFromJSON = do
+  t <- key "tag" asText
+  case t of
+    "KUnknown" -> do
+      KUnknown <$> key "contents" (nth 0 asIntegral)
```
LiamGoodacre: Unneeded `do` :)

hdgarrood: Fixed, thanks :)

PR #2535, opened 2017-01-04
Replace type wildcards earlier

### LINKED ISSUE #2534: Type checker bug involving wildcards
Not sure what's going on here, but it results in a runtime error. It requires the annotation on `sizeOf`.

```purescript
import Prelude

import Control.Comonad
import Control.Comonad.Cofree

data ExprF a
  = Lit Int
  | Add a a
  
derive instance functorExprF :: Functor ExprF

type Expr = Cofree ExprF

sized :: Expr Unit -> Expr Int
sized = extend (sizeOf <<< tail) where
  sizeOf :: ExprF _ -> Int
  sizeOf (Lit _) = 1
  sizeOf (Add x y) = sizeOf x + sizeOf y
```

Obviously the definition of `sizeOf` is wrong, but it passes the type checker.

paf31:
Fixes #2534

This looks like a bigger change than it is. The key is to move the call to `replaceTypeWildcards` from `checkTypedBindingGroupElement` into `typeDictionaryForBindingGroup`. This function is called earlier and returns the map from function names in a binding group to their types. This forces the necessary unification to happen on the expanded type instead of the original type, which includes the wildcard.

I also did a bit of tidying to make the formatting in this module more consistent.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -239,11 +239,11 @@ typeForBindingGroupElement (ident, val) dict untypedDict = do
   return (ident, (TypedValue True val' ty, ty))
 
 -- | Check the kind of a type, failing if it is not of kind *.
-checkTypeKind ::
-  (MonadError MultipleErrors m) =>
-  Type ->
-  Kind ->
-  m ()
+checkTypeKind
+  :: MonadError MultipleErrors m
+  => Type
+  -> Kind
+  -> m ()
 checkTypeKind ty kind = guardWith (errorMessage (ExpectedType ty kind)) $ kind == kindType
 
 -- | Remove any ForAlls and ConstrainedType constructors in a type by introducing new unknowns
@@ -251,11 +251,11 @@ checkTypeKind ty kind = guardWith (errorMessage (ExpectedType ty kind)) $ kind =
 --
 -- This is necessary during type checking to avoid unifying a polymorphic type with a
 -- unification variable.
-instantiatePolyTypeWithUnknowns ::
-  (MonadState CheckState m, MonadError MultipleErrors m) =>
-  Expr ->
-  Type ->
-  m (Expr, Type)
+instantiatePolyTypeWithUnknowns
+  :: (MonadState CheckState m, MonadError MultipleErrors m)
+  => Expr
+  -> Type
+  -> m (Expr, Type)
 instantiatePolyTypeWithUnknowns val (ForAll ident ty _) = do
   ty' <- replaceVarWithUnknown ident ty
   instantiatePolyTypeWithUnknowns val ty'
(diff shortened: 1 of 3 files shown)


paf31:
Not sure what the problem is with Prelude, but I'll have a look tomorrow.

paf31:
It turned out the remaining issue was fairly simple. If the test suite passes, this should be ready to review.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -170,49 +170,58 @@ type TypeData = M.Map (Qualified Ident) (Type, NameKind, NameVisibility)
...
+-- | This function breaks a binding group down into two sets of declarations:
+-- those which contain type annotations, and those which don't.
+-- This function also generates fresh unification variables for the types of
+-- declarations without type annotations, returned in the 'UntypedData' structure.
 typeDictionaryForBindingGroup
-  :: (MonadState CheckState m)
+  :: (MonadState CheckState m, MonadError MultipleErrors m, MonadWriter MultipleErrors m)
   => Maybe ModuleName
   -> [(Ident, Expr)]
   -> m ([(Ident, Expr)], [(Ident, (Expr, Type, Bool))], TypeData, UntypedData)
```
kRITZCREEK: Great docstring for the function, but it would be nice if we either create a data type for the return type or add a haddock describing what the different parts of the return type mean.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -170,49 +170,58 @@ type TypeData = M.Map (Qualified Ident) (Type, NameKind, NameVisibility)
...
-    typedDict = map (\(ident, (_, ty, _)) -> (ident, ty)) typed
-
-  -- Create fresh unification variables for the types of untyped declarations
-  untypedNames <- replicateM (length untyped) freshType
-
-  let
+    -- Filter the typed and untyped declarations and make a map of names to typed declarations.
+    -- Replace type wildcards here so that the resulting dictionary of types contains the
+    -- fully expanded types.
+    let (untyped, typed) = partitionEithers (map isTyped vals)
```
kRITZCREEK: `typed` sounds a little vague. Isn't this actually `typeAnnotated`?

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -170,49 +170,58 @@ type TypeData = M.Map (Qualified Ident) (Type, NameKind, NameVisibility)
...
+    let untypedDict = zip (map fst untyped) untypedNames
+    -- Create the dictionary of all name/type pairs, which will be added to the
+    -- environment during type checking
+    let dict = M.fromList (map (\(ident, ty) ->
+                 (Qualified moduleName ident,
+                   (ty, Private, Undefined))) $ typedDict ++ untypedDict)
+    return (untyped, typed', dict, untypedDict)
+  where
+    -- | Check if a value contains a type annotation
+    isTyped :: (Ident, Expr) -> Either (Ident, Expr) (Ident, (Expr, Type, Bool))
```
kRITZCREEK: `containsAnnotation` and the comment would no longer be needed.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -170,49 +170,58 @@ type TypeData = M.Map (Qualified Ident) (Type, NameKind, NameVisibility)
...
+      ty' <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< replaceTypeWildcards $ ty
+      return ((ident, ty'), (ident, (expr, ty', checkType)))
+    -- Create fresh unification variables for the types of untyped declarations
+    untypedNames <- replicateM (length untyped) freshType
     -- Make a map of names to the unification variables of untyped declarations
-    untypedDict = zip (map fst untyped) untypedNames
-    -- Create the dictionary of all name/type pairs, which will be added to the environment during type checking
-    dict = M.fromList (map (\(ident, ty) -> (Qualified moduleName ident, (ty, Private, Undefined))) $ typedDict ++ untypedDict)
-  return (untyped, typed, dict, untypedDict)
+    let untypedDict = zip (map fst untyped) untypedNames
```
kRITZCREEK: Depending on how many elements we get here it might make sense to do this in one pass with `zipWith`

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -170,49 +170,58 @@ type TypeData = M.Map (Qualified Ident) (Type, NameKind, NameVisibility)
...
-  -- Create fresh unification variables for the types of untyped declarations
-  untypedNames <- replicateM (length untyped) freshType
-
-  let
+    -- Filter the typed and untyped declarations and make a map of names to typed declarations.
+    -- Replace type wildcards here so that the resulting dictionary of types contains the
+    -- fully expanded types.
+    let (untyped, typed) = partitionEithers (map isTyped vals)
+    (typedDict, typed') <- fmap unzip . for typed $ \(ident, (expr, ty, checkType)) -> do
+      ty' <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< replaceTypeWildcards $ ty
```
kRITZCREEK: It would be nice if we could track whether this already happened with a phantom type. During the thesis it appeared like a magic incantation that "does things" when you get a certain error.

kRITZCREEK:
I'm a little confused by the test failures, since `TUnknown` is a `Type` and not an `Expr` which `infer` takes as its argument...

paf31:
@kRITZCREEK Thanks for the review! How does this look?

I agree with you about the phantom type thing. There are a few things like that which I would like to fix, but I'm going to leave it for later in this case.

kRITZCREEK:
Much better :+1: Thanks :)

PR #2536, opened 2017-01-04
Restore JSON backwards compatibility for PSStrings

hdgarrood:
This allows the parser to deal with JSON produced by earlier versions of
the compiler (which were generated as JSON strings) as well as the
current generator (which produces arrays of integers).

I've actually tested it this time with the entire pursuit DB and I think after this we will have restored backwards compatibility completely.

### DIFF
--- a/src/Language/PureScript/CoreFn/ToJSON.hs
+++ b/src/Language/PureScript/CoreFn/ToJSON.hs
@@ -9,6 +9,7 @@ module Language.PureScript.CoreFn.ToJSON
 
 import Prelude.Compat
 
+import Data.Maybe (fromMaybe)
 import Data.Aeson
 import Data.Version (Version, showVersion)
 import Data.Text (Text)
@@ -17,7 +18,7 @@ import qualified Data.Text as T
 import Language.PureScript.AST.Literals
 import Language.PureScript.CoreFn
 import Language.PureScript.Names
-import Language.PureScript.PSString (PSString)
+import Language.PureScript.PSString (PSString, decodeString)
 
 literalToJSON :: (a -> Value) -> Literal a -> Value
 literalToJSON _ (NumericLiteral (Left n)) = toJSON ("IntLiteral", n)
@@ -52,8 +53,17 @@ bindToJSON :: Bind a -> Value
 bindToJSON (NonRec _ n e) = object [ runIdent n .= exprToJSON e ]
 bindToJSON (Rec bs) = object $ map (\((_, n), e) -> runIdent n .= exprToJSON e) bs
 
+-- If all of the labels in the record can safely be converted to JSON strings,
+-- we generate a JSON object. Otherwise the labels must be represented as
+-- arrays of integers in the JSON, and in this case we generate the record as
+-- an array of pairs.
 recordToJSON :: (a -> Value) -> [(PSString, a)] -> Value
-recordToJSON f = toJSON . map (\(key, a) -> (toJSON key, f a))
+recordToJSON f rec = fromMaybe (asArrayOfPairs rec) (asObject rec)
+  where
+  asObject = fmap object . traverse (uncurry maybePair)
+  maybePair label a = fmap (\l -> l .= f a) (decodeString label)
+
+  asArrayOfPairs = toJSON . map (\(label, a) -> (toJSON label, f a))
 
 exprToJSON :: Expr a -> Value
 exprToJSON (Var _ i)              = toJSON ( "Var"
--- a/src/Language/PureScript/PSString.hs
+++ b/src/Language/PureScript/PSString.hs
@@ -41,6 +41,11 @@ import qualified Data.Aeson.Types as A
 -- The Show instance for PSString produces a string literal which would
 -- represent the same data were it inserted into a PureScript source file.
 --
+-- Because JSON parsers vary wildly in terms of how they deal with lone
+-- surrogates in JSON strings, the ToJSON instance for PSString produces JSON
+-- strings where that would be safe (i.e. when there are no lone surrogates),
+-- and arrays of UTF-16 code units (integers) otherwise.
+--
 newtype PSString = PSString { toUTF16CodeUnits :: [Word16] }
   deriving (Eq, Ord, Monoid)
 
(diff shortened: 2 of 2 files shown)


hdgarrood:
Oh also - can we please do a 0.10.5 release after this?

paf31:
Looks good, but which JSON formats does this new representation affect? Externs? What about `core.json`?

Does it make sense to use JSON strings if the value falls into that subset when encoding?

hdgarrood:
All of them: corefn, externs, and psc-publish, since all of them contain `Type` values, and the `TypeLevelString` constructor contains a `PSString`. #2488 changed all of these formats in a breaking way, i.e. other parsers will have to be updated to support the new format (sadly, it wasn't possible to do this in a non-breaking way).

I briefly considered using JSON strings if a `PSString` can be safely represented as one, but I decided not to in the end since I think it would make it easier to accidentally write a parser which expects all values to be JSON strings. I also think it will make creating parsers simpler since they only have to deal with one representation rather than two, and also I think it will be easier to write alternative backends that do the right thing if they have access to the array of integers representation.

hdgarrood:
That reminds me: purescript-corefn is one of the things that will have to be updated because of #2488 (/cc @paulyoung, see the commit message in https://github.com/purescript/purescript/pull/2488/commits/45478c23ccebf7ca56e7f83ace4efb4e48a7fd25 for details of what has changed).

Come to think of it, should this change mean that we should treat this as a breaking change and bump the version to 0.11 in the next release?

paf31:
The only one I would really prefer not to break is corefn.

paf31:
And if we decide to make a 0.11 release, then I would like to deal with more of the things under the <kbd>breaking</kbd> tag at the same time.

paf31:
What would be the problem with just turning the code points into a string literal in the same way as we do during codegen for literals, for example?

hdgarrood:
Because different JSON parsers deal with strings like that in completely different ways. For example, given the JSON string `["\ud800"]`:

* Parsing with JavaScript's `JSON.parse` gives you `["\ud800"]`
* Attempting to parse with aeson gives you `Nothing`
* Attempting to parse with Go gives you ["\ufffd"] (the unicode replacement character)
* Attempting to parse with ruby's stdlib JSON parser gives you `[nil]`

paf31:
Presumably though there is still one correct way to decode such string, defined by some spec? I don't think we should make the default case, which is going to be _way_ more common, more difficult for all backend authors because implementations differ on some corner case.

hdgarrood:
No :( there are multiple conflicting JSON specs, and none is clear about what should happen in this case.

paf31:
Well, how are we judging the correctness of our own implementation then? Compatibility with `JSON.parse`?

hdgarrood:
What do you mean, our own implementation? I'm just talking about the process of going from `ByteString -> Maybe Data.Aeson.Value`, so we don't implement that ourselves, we have aeson do that for us.

hdgarrood:
Ohh, I see. I'd be quite hesitant to require JSON parsers to be compatible with JavaScript's to be able to parse the corefn JSON. We'd have to either fork aeson or persuade the maintainers to merge a breaking change, to start with, and additionally anyone looking to consume the corefn JSON in any language other than JS would probably be forced to use a nonstandard parser too.

hdgarrood:
Generating JSON strings where that's safe and arrays of integers where it isn't also works, but it's still a breaking change (albeit one that any corefn consumers can probably get away with ignoring for a while).

paf31:
Well yes, but a) all of this is only relevant to a tiny, unlikely fraction of the encodable strings (right?), and b) I think the correct path here would be to encourage maintainers to fix any broken implementations, should such a need arise. Two things would need to happen for this to be a problem: someone would need to write a backend in a language other than PureScript or JavaScript (I think all current corefn backends are written in one or the other), and some user would need to write PureScript code which used these unusual `PSString`s somewhere too. If that were to happen, I'd be happy placing the blame on the JSON parser in the other language.

And above all else, the `corefn.json` format is supposed to be relatively readable, and this pretty much removes that property for any code using records.

hdgarrood:
Right, that makes sense. Yes, all of this is only relevant to a tiny fraction of the possible `PSString` values. In that case would you like me to push another commit to this PR which changes the `ToJSON PSString` instance so that it generates a JSON string where this is safe, and an array of integers otherwise?

How do you think we should generate record literals in the corefn JSON when some of the labels are invalid UTF-16? Right now, as of #2488, all record literals are generated as arrays of pairs (two-element arrays). Perhaps I could use the same approach: generate record literals as before except where we have any invalid UTF-16 keys, in which case use the arrays of pairs format?

hdgarrood:
Oh wait, sorry, no, you're suggesting we use JSON strings for everything and generate them the same way as we do in JS output? I think that would be quite difficult because Haskell programs need to be able to parse this JSON, e.g. Pursuit and `psc-ide`. I think @andyarvanitis is using Haskell rather than PureScript for the majority of his work on pure11 too.

paf31:
If you could, I'd appreciate it. That way, it's technically still a breaking change, but not really for all intents and purposes, and I'd be happy calling this 0.10.5.

Wouldn't the format for records be the same as what we had before, but with this new mixed string representation?

> you're suggesting we use JSON strings for everything and generate them the same way as we do in JS output?

No, only for the "safe" subset, arrays of integers for anything else is fine with me.

hdgarrood:
Ok great, will do. With records, we used to generate JSON objects, but I don't think we can do that for invalid UTF-16 labels.

hdgarrood:
It's a shame that "JSON parser" refers to both `ByteString -> Maybe Value` and `Value -> Maybe SomeSpecificType`...

paf31:
Ah ok, then your idea for record literals or lists of pairs sounds fine for now. We can always come back to it before 1.0 if it proves to be problematic.

hdgarrood:
Great. I'll just test this a bit more before merging.

paulyoung:
@hdgarrood thanks for the heads up.

paulyoung:
Sorry to ask, but was there ever a breaking change released?

michaelficarra:
@paulyoung Yeah, the 0.10.4 `FromJSON` implementation would not accept a string.

paulyoung:
@michaelficarra thanks, I didn't see it in the 0.10.4 release notes.

paulyoung:
It sounds like compatibility was restored in 0.10.5 if I'm understanding the release notes correctly, so I don't think I need to make any updates after all.

paf31:
Yes, that's right. It was an oversight which was corrected in 0.10.5.

PR #2539, opened 2017-01-05
Update pretty printer for Kinds

hdgarrood:
* Use aliases for Type and Effect
* Use Symbol rather than Prim.Symbol
* A couple more conversions to Text


paf31:
Actually, I was thinking of moving away from the single character names soon. Could we keep `Type` and `Effect`? I agree that they should not show `Prim` though.

hdgarrood:
Sure, I'll push another commit in a little bit.

src/Language/PureScript/Pretty/Kinds.hs
```diff
@@ -19,17 +19,16 @@ import Language.PureScript.Crash
...
 typeLiterals = mkPattern match
   where
-  match (KUnknown u) = Just $ T.cons 'u' (T.pack (show u))
-  match k@(NamedKind name)
-    | k == kindType   = Just "*"
-    | k == kindEffect = Just "!"
-    | k == kindSymbol = Just "Symbol"
-    | otherwise       = Just $ showQualified runProperName name
+  match (KUnknown u) =
+    Just $ T.cons 'u' (T.pack (show u))
```
paf31: Maybe we should use `k` here actually, since we now use `t` for types?

hdgarrood: :+1:

PR #2544, opened 2017-01-06
Add git tag time to psc-publish JSON

hdgarrood:
This is done in a backwards-compatible way by making the time in the
Package type a Maybe UTCTime. Later we can change this to just a
UTCTime.

Prerequisite for addressing https://github.com/purescript/pursuit/issues/216


hdgarrood:
Oops, CI is failing because I merged this a bit hastily. Hopefully it will pass on `master`.

PR #2546, opened 2017-01-07
Don't expand synonyms until after kind checking

### LINKED ISSUE #2542: False positives with type variable shadow warning in 0.10.5
This function now raises a warning about a shadowed `a` (the `updateSelect` part, the rest is highlighted to illustrate there's no other `a` in scope):

https://github.com/slamdata/slamdata/blob/5988934c04e9b52902d74b9e1e7b685bc9d3664f/src/SlamData/Workspace/Card/BuildChart/Funnel/Component.purs#L234-L250

paf31:
cc @garyb @LiamGoodacre

### DIFF
--- a/examples/passing/2542.purs
+++ b/examples/passing/2542.purs
@@ -1,0 +1,12 @@
+module Main where
+
+import Control.Monad.Eff.Console
+
+type T = forall a. Array a
+
+foo :: T
+foo = bar where
+  bar :: forall a. Array a
+  bar = []
+
+main = log "Done"
--- a/examples/warning/2542.purs
+++ b/examples/warning/2542.purs
@@ -1,3 +1,4 @@
+-- @shouldWarnWith MissingTypeDeclaration
 module Main where
 
 import Control.Monad.Eff.Console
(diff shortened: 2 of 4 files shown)


examples/passing/2542.purs
```diff
@@ -0,0 +1,12 @@
+module Main where
+
+import Control.Monad.Eff.Console
+
+type T = forall a. Array a
+
+foo :: T
+foo = bar where
+  bar :: forall a. Array a
+  bar = []
```
LiamGoodacre: What do you think about this being a warning test instead of a passing test?  This would have passed before the change - as our passing tests ignore warnings (if I recall correctly).

paf31: Good idea, thanks!

examples/warning/2542.purs
```diff
@@ -0,0 +1,13 @@
...
+import Control.Monad.Eff.Console
+
+type T = forall a. Array a
+
+foo :: T
+foo = bar where
+  bar :: forall a. Array a
+  bar = []
+
+main = log "Done"
```
garyb: Maybe we should add a comment in here to explain what is actually being tested?

LiamGoodacre: Something like this?

> This should not raise a `ShadowedTypeVar` warning as the type `a` introduced in `T` shouldn't scope over the type `a` introduced in `bar`.

PR #2547, opened 2017-01-07
Fail early when 'bind' is brought into scope inside 'do'

### LINKED ISSUE #1253: Improve error message when variable name "bind" is used in a do block.

paf31:
Fixes #1253

This fails in the do notation desugaring pass instead.

The error message span isn't great right now, but I think we can fix that when we fix the more general problem.

### DIFF
--- a/examples/failing/BindInDo-2.purs
+++ b/examples/failing/BindInDo-2.purs
@@ -1,0 +1,9 @@
+-- @shouldFailWith CannotUseBindWithDo
+module Main where
+
+import Prelude
+
+foo = do
+  let bind = 42
+  x <- [4, 5, 6]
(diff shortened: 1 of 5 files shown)


LiamGoodacre:
Should we also check for this in a `DoNotationLet`?

paf31:
Oh right, of course. Thanks!

paf31:
Fixed, thanks!

LiamGoodacre:
Looks good 👍

PR #2549, opened 2017-01-08
Fix object key quoting

hdgarrood:
At the moment, names which require quoting are not quoted and names which do not require quoting are quoted. For example, `{ "💡" :: Int }` will come out as `{ 💡 :: Int }` in error messages, even though the parser will reject the latter, but `{ foo :: Int }` comes out as `{ "foo" :: Int }`.


hdgarrood:
(I accidentally broke this in #2488).

PR #2550, opened 2017-01-08
Store more information in RenderedCode / refactoring

hdgarrood:
* Changes the RenderedCodeElement type to preserve more information for
  creating links in RenderedCode values. In particular, the namespace of
  a symbol is now stored.
* Refactor Docs.RenderedCode.Types so that the functions for creating
  RenderedCode values are easier to use and more type-safe.
* Include information about the module that kinds are contained in in
  the RenderedCode renderer.
* Split Docs.RenderedCode.Render into two modules:
  Docs.RenderedCode.RenderType and Docs.RenderedCode.RenderKind.

The JSON format has changed but in a backwards-compatible way. So once
this is released and Pursuit is updated with these changes, old
compilers will not be able to upload to Pursuit, but Pursuit will still
understand JSON produced by old compilers.

This gets us most of the way towards fixing https://github.com/purescript/pursuit/issues/271 and https://github.com/purescript/pursuit/issues/119.

### DIFF
--- a/src/Language/PureScript/Docs/RenderedCode/RenderType.hs
+++ b/src/Language/PureScript/Docs/RenderedCode/RenderType.hs
@@ -1,13 +1,13 @@
 -- | Functions for producing RenderedCode values from PureScript Type values.
 
-module Language.PureScript.Docs.RenderedCode.Render
+module Language.PureScript.Docs.RenderedCode.RenderType
   ( renderType
   , renderTypeAtom
   , renderRow
-  , renderKind
   , RenderTypeOptions(..)
   , defaultRenderTypeOptions
   , renderTypeWithOptions
+  , renderTypeAtomWithOptions
   ) where
 
 import Prelude.Compat
@@ -20,15 +20,16 @@ import Control.Arrow ((<+>))
 import Control.PatternArrows as PA
 
 import Language.PureScript.Crash
-import Language.PureScript.Docs.RenderedCode.Types
-import Language.PureScript.Docs.Utils.MonoidExtras
 import Language.PureScript.Environment
 import Language.PureScript.Kinds
 import Language.PureScript.Names
-import Language.PureScript.Pretty.Kinds
 import Language.PureScript.Pretty.Types
 import Language.PureScript.Types
 import Language.PureScript.Label (Label)
+
+import Language.PureScript.Docs.RenderedCode.Types
+import Language.PureScript.Docs.Utils.MonoidExtras
+import Language.PureScript.Docs.RenderedCode.RenderKind (renderKind)
 
 typeLiterals :: Pattern () Type RenderedCode
 typeLiterals = mkPattern match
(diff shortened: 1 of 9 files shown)


garyb:
> The JSON format has changed but in a backwards-compatible way. So once this is released and Pursuit is updated with these changes, old compilers will not be able to upload to Pursuit, but Pursuit will still understand JSON produced by old compilers.

I'm not sure I understand this description of "backwards compatible" :wink: - is it that the uploads will start failing, but any existing "old" content will still work then?

src/Language/PureScript/Docs/RenderedCode/RenderType.hs
```diff
@@ -154,12 +154,6 @@ insertPlaceholders :: RenderTypeOptions -> Type -> Type
 insertPlaceholders opts =
   everywhereOnTypesTopDown convertForAlls . everywhereOnTypes (convert opts)
 
-dePrim :: Type -> Type
-dePrim ty@(TypeConstructor (Qualified _ name))
-  | ty == tyBoolean || ty == tyNumber || ty == tyString =
-    TypeConstructor $ Qualified Nothing name
-dePrim other = other
```
garyb: Do we not want to remove `Prim.` qualification anymore? Or is this handled somewhere else that I didn't notice now?

hdgarrood:
Yeah. Basically we don't have to try to regenerate the entire pursuit database, which is nice because that's a pain. I guess it's not fully backwards compatible but it is backwards compatible in the way that's most important from our point of view ;)

garyb:
Got it 👍

src/Language/PureScript/Docs/RenderedCode/RenderType.hs
hdgarrood: It is handled inside Pursuit, yeah. Really this function should never have been there at all, it was an oversight to include it. We want to keep this information in because we use it to generate links; see for example https://pursuit.purescript.org/packages/purescript-react/2.0.0/docs/React#t:KeyboardEvent - some of those are broken and I am fairly sure it is because of this function.

garyb: Ah ok, that totally makes sense! It was probably in here originally for `psc-docs`.

hdgarrood: Oh just to clarify: `psc-docs` output shouldn't change as a result of this PR either. Also here's the part of Pursuit where this information is used, in case you're interested: https://github.com/purescript/pursuit/blob/681559794a31e3de18452124a718ab09c64aaec6/src/Model/DocsAsHtml.hs#L201-L216

hdgarrood: I think I wrote this code originally, and this module was mostly copied from `Language.PureScript.Pretty.Types` as it was at that time - of course this was the behaviour we wanted in `prettyPrintType`, but not what we wanted here, although I failed to realise that at the time.

hdgarrood:
Oh wait a minute, I'm talking nonsense. Old compilers will still be able to upload to Pursuit!

hdgarrood:
I force-pushed so that I could fix the erroneous commit message. Also no rush, but it would be nice to get this reviewed, as afterwards I can look at generating HTML docs in `psc-docs`, e.g. for generating docs from package sets.

paf31:
@garyb Did you already review this?

paf31:
Travis is passing now.

hdgarrood:
Bump - is there anything I can do to make reviewing this easier?

hdgarrood:
Thanks!

garyb:
Sorry about the slow review :)

hdgarrood:
No worries, I appreciate that it was quite a big PR :)

PR #2551, opened 2017-01-08
Use Map.foldlWithKey' instead of foldl

hdgarrood:
This change should improve memory usage. Thanks to @kRITZCREEK for pointing this out!


hdgarrood:
@LiamGoodacre suggests using `(fst <$> m, snd <$> m)` instead, which might be better still?

kRITZCREEK:
`fmap fst &&& fmap snd` ? :D

Issue #2130, opened 2016-05-15
Try to get rid of Language.PureScript.AST.Exported


hdgarrood:
Just a refactoring suggestion: I would like to get rid of this module, or at least try to, as it overlaps fairly significantly with `.Sugar.Names.Env`, and is only used in a couple of places in the Docs modules and in psci. The Docs stuff has access to real `Env` values already so that should be easy to change, I think.

garyb:
I think that should be possible, I'm going to elaborate the re-exports during `Names`, so all the information should be readable from then on. That should help with the psci autocompletion too.

hdgarrood:
I think this should probably be closed, actually. It's difficult to imagine how to get rid of this module given that we need information from the AST to make sure that docs appear in the right order; see #3003 and #2525.

paf31:
Ok, we can always reopen it, thanks.

PR #2553, opened 2017-01-08
Remove Docs.Bookmarks

hdgarrood:
We don't actually need the full Bookmarks to be able to generate links
between packages properly, and in fact all that Pursuit uses the
pkgBookmarks field for now is to work out which package it needs to link
to, given a module name. That is, all it needs is a Map ModuleName
PackageName. So we can simplify and speed up the JSON generation by
replacing the pkgBookmarks with such a Map.

This gets us one step closer to being able to get rid of
Language.PureScript.AST.Exported, refs #2130.


src/Language/PureScript/Docs/Convert.hs
```diff
@@ -3,28 +3,27 @@
...
 module Language.PureScript.Docs.Convert
   ( convertModules
+  , convertModulesWithEnv
   , convertModulesInPackage
-  , collectBookmarks
+  , convertModulesInPackageWithEnv
   ) where
 
-import Prelude.Compat
+import Prelude (id, String)
```
kRITZCREEK: Just in case you didn't know `Protolude` exports `id` as `identity` to free up the name `id`.

hdgarrood: Ah right, thanks!

hdgarrood:
CI failure is that intermittent issue with parsing HPC files again. I'm going to try clearing that job's cache.

hdgarrood:
There's no cache for this PR appearing in the travis settings. :slightly_frowning_face: I guess we can look at it later. Maybe we should disable coverage for now, actually, since apparently the last time it worked was 6 months ago: https://coveralls.io/github/purescript/purescript

hdgarrood:
Cool :) I think I'll wait a little bit and see if this CI issue sorts itself out before sending a PR disabling coverage.

PR #2554, opened 2017-01-08
remove JSAccessor; replace with JSIndexer

### LINKED ISSUE #2513: remove JSAccessor, replacing its usages with JSIndexer
We can then render instances of `JSIndexer` with a `StringLiteral` as static member access when the ECMAScript grammar allows for it. See discussion at https://github.com/purescript/purescript/pull/2488#discussion_r94237545. I'll send a PR once I'm done with #2488.

michaelficarra:
Fixes #2513. Originally, I had planned to codegen a JSIndexer with a JSStringLiteral using static member access. But I figured it's okay to just codegen everything using computed member access and have a minifier do the replacement. Let me know if you'd rather do the codegen simplification in the compiler.


michaelficarra:
I've pushed a second commit to codegen static member accesses. Now you can either cherry-pick the first commit if you want to go with the strategy in my PR message or just merge the whole PR if you want to generate static member accesses.

src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -69,7 +69,7 @@ moduleToJs (Module coms mn imps exps foreigns decls) foreign_ =
     let standardExps = exps \\ foreignExps
     let exps' = JSObjectLiteral Nothing $ map (mkString . runIdent &&& JSVar Nothing . identToJs) standardExps
                                ++ map (mkString . runIdent &&& foreignIdent) foreignExps
-    return $ moduleBody ++ [JSAssignment Nothing (JSAccessor Nothing "exports" (JSVar Nothing "module")) exps']
+    return $ moduleBody ++ [JSAssignment Nothing (JSIndexer Nothing (JSStringLiteral Nothing "exports") (JSVar Nothing "module")) exps']
```
hdgarrood: Would it make sense to use `accessorString` here (and in the other places in this module) to make the code a bit more concise?

src/Language/PureScript/Pretty/JS.hs
```diff
@@ -162,8 +162,10 @@ conditional = mkPattern match
 accessor :: (Emit gen) => Pattern PrinterState JS (gen, JS)
 accessor = mkPattern match
   where
-  -- WARN: if `prop` does not match the `IdentifierName` grammar, this will generate invalid code; see #2513
-  match (JSAccessor _ prop val) = Just (emit prop, val)
+  match (JSIndexer _ (JSStringLiteral _ prop) val) =
```
hdgarrood: I think it might be nice to provide a pattern synonym called `JSAccessor` which more or less does this - then I think that would be useful in e.g. the inliner.

hdgarrood:
I think this is looks good; I've made a couple of minor suggestions but I don't think either of them are at all vital. Maybe get a second opinion before you go off and do what I've suggested, though, since I haven't done a huge amount of work on this part of the compiler before.

src/Language/PureScript/CodeGen/JS.hs
michaelficarra: Yep, I'll do that.

src/Language/PureScript/Pretty/JS.hs
michaelficarra: I don't like the idea of pattern synonyms. I think the additional indirection will be confusing. I think this is the only place it'd be used anyway.

paf31:
The code looks fine to me, but what's the benefit of replacing `JSAccessor`? One could argue that this is now less strongly-typed than before, since we need to inspect a string to decide how to act on what used to be two constructors.

I couldn't figure out from the issue if this fixed a bug or if it was just cosmetic.

michaelficarra:
@paf31 The point was actually to eliminate the possibility of constructing a `JSAccessor` with a `Text` value that is not syntactically allowed as a JS static member access. You could say one solution to that problem would be smart constructors, but I don't think it even makes sense to have a mental model that distinguishes the two different syntaxes for the same concept. Instead, the mental model should just be "access a property with this name". Whether that's allowed by the JavaScript syntax shouldn't matter and shouldn't have to enter your thoughts.

> we need to inspect a string to decide how to act on what used to be two constructors

I've simply moved that test to the pretty printer from a helper in the codegen that switched over which constructor to use. Same thing. And as I suggested above, we can just omit that test and always generate computed member accesses, leaving the optimisation up to minifiers or other tooling.

paf31:
Ok, that makes sense to me, thanks for clarifying.

paf31:
Could you please resolve the conflict, and I'll get this merged in. It looks like a fairly minor conflict, you might be able to fix it in the web UI actually.

paf31:
@michaelficarra I just tried this and noticed that the code generator is now emitting the `M["x"]` form pretty much everywhere. Do you know why that might happen?

paf31:
@michaelficarra I'll create an issue.

PR #2556, opened 2017-01-10
[psc-ide] Complete record accessors

kRITZCREEK:
Adds a new completion command to psc-ide which accepts a file path and a cursor position to figure out possible record accessors at that position. 

A demo gif as posted on twitter:

psc-ide-racc

Right now detecting where the typed hole should go is still very basic. The code just deletes the dot and wraps everything that's not a space character in front of it like so:

```purescript
record.

--------

(?typedHole record)
```

It then just takes the first argument to the inferred function type and if it is a record returns a list of completions for the field names. This completion command is fundamentally different from our other completions, as it treats the server less like a database and instead in a more "RPC"ish way, but there's no real way around it and it's commonplace in other language servers so it's natural that it came up for us as well.

I'd like to make figuring out where the typed hole needs to go more robust, but maybe this works for a start? We could special case parenthesized expressions to demonstrate that this works through actual type inference and not by parsing some type annotation, but I think it's already pretty fun to play with in its current state.

EDIT:

For the emacs'ers out there you can pull from this branch to try it out:
https://github.com/epost/psc-ide-emacs/pull/85

### DIFF
--- a/src/Language/PureScript/Ide/Completion.hs
+++ b/src/Language/PureScript/Ide/Completion.hs
@@ -1,17 +1,24 @@
+{-# LANGUAGE PackageImports #-}
 module Language.PureScript.Ide.Completion
        ( getCompletions
        , getExactMatches
+       , completeInFile
        ) where
 
 import           Protolude
 
+import Data.Char
+import "monad-logger" Control.Monad.Logger
+import qualified Data.Text as T
 import           Language.PureScript.Ide.Filter
 import           Language.PureScript.Ide.Matcher
 import           Language.PureScript.Ide.Types
+import           Language.PureScript.Ide.Rebuild
+import           Language.PureScript.Ide.Error
+import           Language.PureScript.Ide.Util
 import qualified Language.PureScript as P
 
 type Module = (P.ModuleName, [IdeDeclarationAnn])
-
 -- | Applies the CompletionFilters and the Matcher to the given Modules
 --   and sorts the found Completions according to the Matching Score
 getCompletions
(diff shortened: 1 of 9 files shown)


src/Language/PureScript/Ide/Completion.hs
```diff
@@ -28,3 +38,65 @@ completionsFromModules = foldMap completionFromModule
...
+    insertHole t =
+      let
+        (before, line:after) = splitAt (row - 1) (T.lines t)
+        (b, a) = T.splitAt (col - 1) line
+        (start, ident) = breakEnd (not . isSpace) b
+        withHole = case ident of
+          -- in this special case we are looking at syntax sugar for a record
+          -- accessor function
+          "_" ->
+            start <> "?magicUnicornHole" <> T.tail a
```
paf31: `magicUnicornHole`? 😄

paf31:
Presumably this will only work in certain cases, like if the file parses and type checks successfully already, yes?

Still, very cool 😄 

Maybe it could benefit from a comment or two though ...

src/Language/PureScript/Ide/Completion.hs
kRITZCREEK: yeah a little too clowny :D I had a beer or two ;)

kRITZCREEK:
Yes, this depends on the file parsing/typechecking once the replacement logic has inserted the typed hole. I'm currently working on making that replacement logic more robust and adding a few tests.

kRITZCREEK:
Would you take another look? This isn't perfect, but I think it's a good first increment.

PR #2559, opened 2017-01-13
Fix parsing for named binders

paf31:
I spotted this today. You can see that the example I added actually parses incorrectly, if you run it in Try PureScript.

The only change here is in `parseVarOrNamedBinder`, which now uses `parseBinderNoParens`, but I'm trying to go through and tidy things up a little bit as I make fixes.

### DIFF
--- a/examples/passing/ParseNamedBinder.purs
+++ b/examples/passing/ParseNamedBinder.purs
@@ -1,0 +1,9 @@
+module Main where
+
+import Control.Monad.Eff.Console (log)
+
+data X = X
+
+f a@X X = a
+
+main = log "Done"
--- a/src/Language/PureScript/Parser/Common.hs
+++ b/src/Language/PureScript/Parser/Common.hs
@@ -1,52 +1,39 @@
--- |
--- Constants and utility functions to be used when parsing
---
+-- | Useful common functions for building parsers
 module Language.PureScript.Parser.Common where
 
-import Prelude.Compat
+import           Prelude.Compat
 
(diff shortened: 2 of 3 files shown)


paf31:
Thanks @kRITZCREEK!

natefaubion:
Following up from #2580, this change looks like it broke some code in SlamData. I'd argue that the code in SlamData is correct given how we bracket operator aliases in patterns. The offending pattern is something like:

```purescript
case a of
  [ foo@Foo { bar } ] -> ...
```

Which is unambiguous, but the changes require this to be bracketed as:

```purescript
case a of
  [ foo@(Foo { bar }) ] -> ...
```

paf31:
I don't think that code is obviously correct. There are two contexts in which we parse binders:

- Regular binders, like those which appear at the top level of a `case` expression, and
- Reduced binders, like those which appear as constructor arguments.

Whether we require parentheses depends on which context you're in.

I would rather not have two different rules for `@` for the two contexts, and I much prefer requiring parens in your example than requiring them here:

```purescript
case th of
  This a -> ...
  These a@(Foo) Bar -> ...
```

Notice the parens are required around `Foo` without the change in this PR.

GHCi does the same thing, for what it's worth:

```
Prelude> data A = A Int Int
Prelude> case A 1 2 of a@A 1 2 -> 1
<interactive>:3:15: Parse error in pattern: a@A

Prelude> case A 1 2 of a@(A 1 2) -> 1
1
```

PR #2560, opened 2017-01-13
Fix inlining for negateInt

### LINKED ISSUE #2136: Inlining of negate can produce out of bounds Int values
For example, with prelude v1.0.0-rc.3:

```
> let x = negate (bottom :: Int)
> x
2147483648

> x > top :: Int
true
```

I think this is because `negate` is being inlined without a `| 0` for `Int` values. For example:

``` purescript
module Main where

import Prelude

main = negate (bottom :: Int) > top
```

produces

``` javascript
var main = -Data_Bounded.bottom(Data_Bounded.boundedInt) > Data_Bounded.top(Data_Bounded.boundedInt);
```

Adding a bitwise-or with 0 would mean that we would get `negate bottom == bottom :: Int` instead, which is a bit weird, but at least means that `Int` is actually a 32-bit integer.

paf31:
Fixes #2136

### DIFF
--- a/examples/passing/2136.purs
+++ b/examples/passing/2136.purs
@@ -1,0 +1,9 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+
+main = 
+  if (negate (bottom :: Int) > top)
+    then log "Fail"
+    else log "Done"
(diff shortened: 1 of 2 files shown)


andyarvanitis:
FYI, LGTM (just added it to C++ inliner)

paf31:
Thank you both!

Issue #2562, opened 2017-01-13
Internal error: "Unknown type index in checkOrphanInstance"


joshuahhh:
I ran into an internal error ("Unknown type index in checkOrphanInstance") during compilation, and was able to strip the offending code down to:

```purescript
import Prelude
import Control.Monad.Writer (class MonadTell, WriterT)

newtype Wrapper m a = Wrapper (
  WriterT Unit m a
)

derive newtype instance monadTellChangeableT ∷ Monad m => MonadTell (Wrapper m)
```

This code is invalid because `MonadTell` takes two type parameters. If I replace `MonadTell (Wrapper m)` with `MonadTell Unit (Wrapper m)`, the code compiles successfully.

Thanks. 😃

paf31:
Thanks for the report!

@LiamGoodacre You might be able to guess what's going on here, but I'll take a look this weekend if it's not obvious.

LiamGoodacre:
In the orphan instance check there is an assumption that the class has all the arguments it expects.  I can look into a fix for this later this evening.

PR #2563, opened 2017-01-14
Fix 'Unknown type index' on mismatch between class and instance argument counts

LiamGoodacre:
Error earlier on argument count mismatch (fixes https://github.com/purescript/purescript/issues/2562)
Removed unused error: DeprecatedRequirePath
Fix error message for ExpectedType: print Type instead of *


paf31:
👍 looks great, thanks!

PR #2566, opened 2017-01-14
Support nested record update

LiamGoodacre:
This adds sugar for nested record updates:

```purescript
x { foo.bar = 3, foo.baz = 2 }
```

Also allows wildcards:

```purescript
_ { foo.bar = _ }
```

### DIFF
--- a/src/Language/PureScript/Sugar/ObjectWildcards.hs
+++ b/src/Language/PureScript/Sugar/ObjectWildcards.hs
@@ -30,8 +30,8 @@ import           Language.PureScript.PSString (PSString)
 -- We represent the updates as the `PathTree`:
 --
 --  M.fromList [ ("foo", Leaf 3)
---             , ("bar", Branch M.fromList [ ("baz", Leaf 1)
---                                         , ("qux", Leaf 2) ]) ]
+--             , ("bar", Branch (M.fromList [ ("baz", Leaf 1)
+--                                          , ("qux", Leaf 2) ]) ])
 --
 -- Which we then convert to an expression representing the following:
 --
@@ -39,6 +39,9 @@ import           Language.PureScript.PSString (PSString)
 --   in x' { foo = 0
 --         , bar = x'.bar { baz = 1
 --                        , qux = 2 } }
+--
+-- The `let` here is required to prevent re-evaluating the object expression `x`.
+-- However we don't generate this when using an anonymous argument for the object.
 --
 type PathTree = Map PSString PathNode
 data PathNode = Leaf Expr | Branch PathTree
--- a/src/Language/PureScript/Sugar/ObjectWildcards.hs
+++ b/src/Language/PureScript/Sugar/ObjectWildcards.hs
@@ -20,15 +20,35 @@ import           Language.PureScript.Label (Label(..))
 import           Language.PureScript.Names
 import           Language.PureScript.PSString (PSString)
 
+-- `PathNode` and `PathTree` are used as an intermediate form when desugaring a nested object update.
+-- For an update such as:
+--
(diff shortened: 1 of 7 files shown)


LiamGoodacre:
For https://github.com/purescript/purescript/issues/1807

src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -24,6 +27,9 @@ desugarObjectConstructors
   -> m Module
 desugarObjectConstructors (Module ss coms mn ds exts) = Module ss coms mn <$> mapM desugarDecl ds <*> pure exts
 
+data PathNode = Leaf Expr | Branch PathTree
```
paf31: I know this isn't exported, but could you add a short comment explaining what it's for anyway? Thanks.

src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -3,19 +3,45 @@ module Language.PureScript.Sugar.ObjectWildcards
...
+-- For an update such as:
+--
+--  x { foo = 0
+--    , bar.baz = 1
+--    , bar.qux = 2 }
+--
+-- We represent the updates as the `PathTree`:
+--
+--  M.fromList [ ("foo", Leaf 3)
+--             , ("bar", Branch M.fromList [ ("baz", Leaf 1)
```
paf31: Nitpick: You need parens around the `M.fromList [ ... ]` for this to be valid 😄

src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -3,19 +3,45 @@ module Language.PureScript.Sugar.ObjectWildcards
...
+--                                         , ("qux", Leaf 2) ]) ]
+--
+-- Which we then convert to an expression representing the following:
+--
+--   let x' = x
```
paf31: Why do you need to generate the `let` here?

LiamGoodacre: I've updated the comment with an explanation.

src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -3,19 +3,48 @@ module Language.PureScript.Sugar.ObjectWildcards
...
+--   in x' { foo = 0
+--         , bar = x'.bar { baz = 1
+--                        , qux = 2 } }
+--
+-- The `let` here is required to prevent re-evaluating the object expression `x`.
+-- However we don't generate this when using an anonymous argument for the object.
```
paf31: I guess what I'm asking is - wasn't this an issue for object updates previously then?

LiamGoodacre: With regular object updates, in our AST, the object expression is only mentioned once.  In the conversion to JavaScript, the expression is evaluated and its result is cached in a variable when building up the new object - but this is a JavaScript implementation detail.

In this case we would be duplicating the expression over multiple unrelated object updates - each of which would evaluate it separately.

```purescript
(f x) { foo.bar = 2 }

-- without a let, this would evaluate `f x` multiple times
(f x) { foo = (f x).foo { bar = 2 } }

-- with the let, we don't
let y = f x in y { foo = y.foo { bar = 2 } }
```

paf31: Oh, I see, thanks for the clarification.

paf31:
Fixes #1807.

paf31:
Thanks!

PR #2570, opened 2017-01-15
Require dependencies to exist during sorting phase

### LINKED ISSUE #1921: Deleted modules still get found
This is somewhat related to #1912 : 
- Create a new module Foo
- Create a new module Bar
- Import Bar from Foo
- Compile everything
- Remove module Bar
- Compile everything <-- no errors

paf31:
Fixes #1921

I did some tidying up again as I went through here. Sorry for all the commit noise.

The main changes are:

- Adding the `ModuleNotFound` error, and throwing it during module sorting.
- Simplifying the module sorting logic, since we only need to look at imports now, I think.

It turns out the problem was that if a dependency were deleted, it was never getting passed into `Make` in the first place, so the "should rebuild" test was always returning `False`, since `Make` thought the other module had no dependencies.

Since `sortModules` is the first thing to happen, it makes sense to me that we should check for missing imports there.

### DIFF
--- a/src/Language/PureScript/ModuleDependencies.hs
+++ b/src/Language/PureScript/ModuleDependencies.hs
@@ -4,13 +4,10 @@ module Language.PureScript.ModuleDependencies
   , ModuleGraph
   ) where
 
-import           Prelude.Compat
+import           Protolude
 
-import           Control.Monad (forM_, when)
-import           Control.Monad.Error.Class (MonadError(..))
 import           Data.Graph
-import           Data.List (nub)
-import           Data.Maybe (fromMaybe)
+import qualified Data.Set as S
 import           Language.PureScript.AST
 import qualified Language.PureScript.Constants as C
 import           Language.PureScript.Crash
@@ -29,7 +26,7 @@ sortModules
   => [Module]
   -> m ([Module], ModuleGraph)
 sortModules ms = do
-    let mns = map getModuleName ms
+    let mns = S.fromList $ map getModuleName ms
     verts <- mapM (toGraphNode mns) ms
     ms' <- mapM toModule $ stronglyConnComp verts
     let (graph, fromVertex, toVertex) = graphFromEdges verts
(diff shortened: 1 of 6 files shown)


paf31:
It seems this messes up fast rebuild somehow, which I can verify using `pscid`. Thankfully the test suite caught it 😄 

I'll look more tomorrow.

src/Language/PureScript/ModuleDependencies.hs
```diff
@@ -1,99 +1,65 @@
...
+                         return (mn, filter (/= mn) (map toKey deps))
+    return (ms', moduleGraph)
   where
-  goModule :: Module -> (Module, ModuleName, [ModuleName])
-  goModule m@(Module _ _ _ ds _) =
-    let ams = concatMap extractQualAs ds
-    in (m, getModuleName m, nub (concatMap (usedModules ams) ds))
+    toGraphNode :: [ModuleName] -> Module -> m (Module, ModuleName, [ModuleName])
+    toGraphNode mns m@(Module _ _ mn ds _) = do
+      let deps = nub (concatMap usedModules ds)
```
kRITZCREEK: This could be very expensive for lists with 30 modulenames that start with: "Slamdata.Editor". ordNub is just better.

src/Language/PureScript/ModuleDependencies.hs
```diff
@@ -1,99 +1,65 @@
...
---
-usedModules :: [ModuleName] -> Declaration -> [ModuleName]
-usedModules ams d =
-  let (f, _, _, _, _) = everythingOnValues (++) forDecls forValues (const []) (const []) (const [])
-      (g, _, _, _, _) = accumTypes (everythingOnTypes (++) forTypes)
-  in nub (f d ++ g d)
-  where
+-- | Calculate a list of used modules based on explicit imports and qualified names.
+usedModules :: Declaration -> [ModuleName]
+usedModules d = nub (f d) where
```
kRITZCREEK: another `nub` on module names

paf31:
@kRITZCREEK I've addressed the test failure and verified it using `pscid` locally. I also updated the `Integration.hs` module so that we can see error messages when a test fails.

I agree that we probably shouldn't be using `nub`, but I'd like to leave that for later, for two reasons:

- It's not high in the profile right now anyway
- I'm planning to work on performance issues in general soon.

kRITZCREEK:
> It's not high in the profile right now anyway

Equality on `Name`'s is in all the profiles though? Isn't that an indication of `nub` creating performance problems?

src/Language/PureScript/ModuleDependencies.hs
garyb: We could just use a `Set` here anyway perhaps? I think there are a few places we use lists where sets would better express the intention and probably perform better also.

rvion:
There is quite a few nub as of now: https://github.com/purescript/purescript/search?p=3&q=nub&utf8=%E2%9C%93

paf31:
Updated. Better?

hdgarrood:
Did you use stylish-haskell on the files you modified during this PR? Would it make sense to run that (or whichever tool it was) on all the Haskell source files in this repo and add a note to CONTRIBUTING.md that any changes to source files should run the same tool on any changed source files too?

paf31:
No, I've been formatting things by hand. I haven't found a template which will generate things how I want to see them, and I haven't had time to create one, but maybe I should ..

PR #2572, opened 2017-01-15
Inline Partial.Unsafe.unsafePartial

### LINKED ISSUE #2288: Tail call elimination does not work with unsafePartial
psc version 0.9.3

First version is TC optimized 

``` purescript
length1 :: forall a. Array a -> Int
length1 =
  unsafePartial length1' 0
  where
  length1' :: forall b. (Partial) => Int -> Array b -> Int
  length1' acc arr =
    if null arr
    then acc
    else length1' (acc + 1) (tail arr)
```

``` javascript
var length1 = (function () {
      var length1$prime = function (__copy_dictPartial) {
          return function (__copy_acc) {
              return function (__copy_arr) {
                  var dictPartial = __copy_dictPartial;
                  var acc = __copy_acc;
                  var arr = __copy_arr;
                  tco: while (true) {
                      var $42 = Data_Array["null"](arr);
                      if ($42) {
                          return acc;
                      };
                      if (!$42) {
                          var __tco_dictPartial = dictPartial;
                          var __tco_acc = acc + 1 | 0;
                          var __tco_arr = Data_Array_Partial.tail(dictPartial)(arr);
                          dictPartial = __tco_dictPartial;
                          acc = __tco_acc;
                          arr = __tco_arr;
                          continue tco;
                      };
                      throw new Error("Failed pattern match at Main line 39, column 5 - line 41, column 39: " + [ $42.constructor.name ]);
                  };
              };
          };
      };
      return Partial_Unsafe.unsafePartial(function (dictPartial) {
          return length1$prime(dictPartial);
      })(0);
  })();
```

Second version is not TC optimised :

``` purescript
length2  :: forall a. Array a -> Int
length2 =
  length2' 0
  where
  length2' :: forall b. Int -> Array b -> Int
  length2' acc arr =
    if null arr
    then acc
    else length2' (acc + 1) (unsafePartial tail arr)
```

``` javascript
  var length2 = (function () {
      var length2$prime = function (acc) {
          return function (arr) {
              var $41 = Data_Array["null"](arr);
              if ($41) {
                  return acc;
              };
              if (!$41) {
                  return length2$prime(acc + 1 | 0)(Partial_Unsafe.unsafePartial(function (dictPartial) {
                      return Data_Array_Partial.tail(dictPartial);
                  })(arr));
              };
              throw new Error("Failed pattern match at Main line 29, column 5 - line 31, column 53: " + [ $41.constructor.name ]);
          };
      };
      return length2$prime(0);
  })();
```

paf31:
Fixes #2288, since this now unblocks the TCO optimization in that case.

### DIFF
--- a/examples/passing/2288.purs
+++ b/examples/passing/2288.purs
@@ -1,0 +1,19 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff
+import Control.Monad.Eff.Console
+import Data.Array
+import Data.Array.Partial as P
+import Partial.Unsafe
+
+length :: forall a. Array a -> Int
+length = go 0 where
+  go acc arr =
+    if null arr
+    then acc
+    else go (acc + 1) (unsafePartial P.tail arr)
+
+main = do
+  logShow (length [1, 2, 3])
+  log "Done"
--- a/examples/passing/2288.purs
+++ b/examples/passing/2288.purs
@@ -15,5 +15,5 @@ length = go 0 where
     else go (acc + 1) (unsafePartial P.tail arr)
 
 main = do
-  logShow (length [1, 2, 3])
+  logShow (length (1 .. 10000))
   log "Done"
--- a/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
+++ b/src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
@@ -67,6 +67,8 @@ evaluateIifes = everywhereOnJS convert
   where
   convert :: JS -> JS
   convert (JSApp _ (JSFunction _ Nothing [] (JSBlock _ [JSReturn _ ret])) []) = ret
+  convert (JSApp _ (JSFunction _ Nothing idents (JSBlock _ [JSReturn ss ret])) [])
+    | not (any (`isReassigned` ret) idents) = replaceIdents (map (, JSVar ss C.undefined) idents) ret
   convert js = js
 
 inlineVariables :: JS -> JS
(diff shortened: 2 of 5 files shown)


examples/passing/2288.purs
```diff
@@ -0,0 +1,19 @@
...
+
+length :: forall a. Array a -> Int
+length = go 0 where
+  go acc arr =
+    if null arr
+    then acc
+    else go (acc + 1) (unsafePartial P.tail arr)
+
+main = do
+  logShow (length [1, 2, 3])
```
LiamGoodacre: Does it make sense to have either this test or another test loop with input large enough that it would have blow the stack if we didn't optimize?

paf31: Yes, will do, thanks.

paf31: Done

paf31:
Ok, this is ready for review again. I've moved some code around so that the optimization for applying a function to `undefined` gets applied more generally (and safely).

src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
```diff
@@ -66,6 +67,8 @@ evaluateIifes = everywhereOnJS convert
   where
   convert :: JS -> JS
   convert (JSApp _ (JSFunction _ Nothing [] (JSBlock _ [JSReturn _ ret])) []) = ret
+  convert (JSApp _ (JSFunction _ Nothing idents (JSBlock _ [JSReturn ss ret])) [])
```
hdgarrood: Ah right, so this takes `(function(x, y, z) { return x + y + z })()` to `undefined + undefined + undefined`?

michaelficarra:
You probably shouldn't refer to the variable `undefined` to get an undefined value. It can be shadowed and, in some environments, written to. Instead, use `void 0` to generate an undefined value.

src/Language/PureScript/CodeGen/JS/Optimizer/Inliner.hs
paf31: Yes, correct

paf31:
`undefined` will get removed by the optimize, but we should probably use `void 0` anyway, or better still, add something to the imperative core later to represent an undefined function argument.

PR #2580, opened 2017-01-16
Change nested record update syntax and fix bug with depth > 2

LiamGoodacre:


src/Language/PureScript/Sugar/ObjectWildcards.hs
```diff
@@ -137,7 +137,7 @@ desugarDecl other = fn other
         goLayer path (key, Branch branch) =
           let path' = path ++ [key]
               updates = goLayer path' <$> M.toList branch
-              accessor = foldr Accessor val path'
+              accessor = foldl (flip Accessor) val path'
```
LiamGoodacre: This is the bug fix.  Apparently I never tested with depth > 2.  It would generate the accessor in reverse.

kRITZCREEK: it's just a little thing, but `foldl'` is preferrable to `foldl` in almost all cases

examples/passing/NestedRecordUpdate.purs
```diff
@@ -3,21 +3,22 @@ module Main where
...
 expected :: T
-expected = { foo: 10, bar: { baz: 20, qux: 30 } }
+expected = { foo: 10, bar: { baz: 20, qux: { lhs: 30, rhs: 40 } } }
 
 check l r =
   l.foo == r.foo &&
   l.bar.baz == r.bar.baz &&
-  l.bar.qux == r.bar.qux
+  l.bar.qux.lhs == r.bar.qux.lhs
+  l.bar.qux.rhs == r.bar.qux.rhs
```
LiamGoodacre: oops missing `&&`

LiamGoodacre:
For https://github.com/purescript/purescript/issues/1807

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -399,13 +400,19 @@ parseInfixExpr
 parseHole :: TokenParser Expr
 parseHole = Hole <$> holeLit
 
-parsePropertyUpdate :: TokenParser (NonEmpty PSString, Expr)
+parsePropertyUpdate :: TokenParser [(NonEmpty PSString, Expr)]
```
paf31: Why not just parse the tree structure directly here? It seems like now you're turning it back and forth into a list of paths.

paf31:
I like the new syntax, thanks!

LiamGoodacre:
I'm pretty much rewriting it from scratch now. 😛 
`wrapLambdaM` is so much nicer with a traversable functor of expressions.

LiamGoodacre:
Now that I'm building the tree earlier, I can't use Map - I need information about the order when processing the wildcards (they correspond to lambdas from right-to-left).

LiamGoodacre:
The parse error is bad when you have duplicate keys:

```
[1/1 ErrorParsingModule] src/Main.purs:8:14

  8  m = {foo: 0} {foo = 1, foo = 2}


  Unable to parse module:
  unexpected {
  expecting indentation at column 1 or end of input
```

In the parser I wrote an error saying `"Duplicate key in record update: " ++ show key`.  But I'm not sure how to make it show that instead?

LiamGoodacre:
I might have worked out how to get a better error message.

LiamGoodacre:
Maybe not.  It only works for the first level.

LiamGoodacre:
The `parseUpdaterBody` is behind a `try` at https://github.com/purescript/purescript/pull/2580/files#diff-476f52dcaa59ca2e52c4e867fda6a60fR441
I've tried moving the `try` inside `parseUpdaterBodyFields` before we build the tree, and it works for the first level, but not nested.

paf31:
Ah ok. So solving this won't be trivial, but it's possible. We'd have to combine records and record updates into a single parser, I think.

kRITZCREEK:
I just tried to compile `slamdata` with the current master for benchmarking purposes and the parser fails on:

```purescript
    collapse ∷ Card.Model → Array Card.Model → Maybe Card.Model
    collapse parent child = case parent, child of
      CM.Draftboard { layout }, [ cm@CM.Draftboard { layout: subLayout } ] → do
        cursor ← Pane.getCursorFor (Just deckId) layout
        layout' ← Pane.modifyAt (const subLayout) cursor layout
        pure $ CM.Draftboard { layout: layout' }
```

with: 

```
[1/1 ErrorParsingModule] src/SlamData/Workspace/Eval/Persistence.purs:385:52

  385        CM.Draftboard { layout }, [ cm@CM.Draftboard { layout: subLayout } ] → do
                                                          
  Unable to parse module:
  unexpected {
  expecting ::, operator, , or ]
```

I guess it's related to the recent accessor chain parser changes? If it doesn't belong here lmk and I'll opene a separate issue.

LiamGoodacre:
@kRITZCREEK would you also be able to try this with the latest commit in this pull request?

kRITZCREEK:
@LiamGoodacre Still get the same error with this branch.

LiamGoodacre:
@kRITZCREEK actually I think what you're seeing is related to https://github.com/purescript/purescript/pull/2559
This bit is causing the failure: `cm@CM.Draftboard { layout: subLayout }`

kRITZCREEK:
Thanks for the help `cm@(CM.Draftboard { layout: subLayout })` fixes the compiler error! Might want to add that to a migration guide for the next release.

natefaubion:
That seems like a bug, doesn't it? Parens aren't required like that for operator aliases.

LiamGoodacre:
@kRITZCREEK @natefaubion Could we move this conversation to either https://github.com/purescript/purescript/pull/2559 or a new issue, thanks :)

PR #2583, opened 2017-01-19
generate JS static member accesses whenever possible

### LINKED ISSUE #2581: JSIndexer overused in codegen
I just tried the latest compiler (after #2554 specifically) and noticed that the code generator is now emitting the `M["x"]` form for accessors pretty much everywhere.

Here is an example I was just working on:

```js
var length = (function () {
    var go = function (acc) {
        return function (arr) {
            var $1 = Data_Array["null"](arr);
            if ($1) {
                return acc;
            };
            if (!$1) {
                return go(acc + 1 | 0)((function (dictPartial) {
                    return Data_Array_Partial["tail"](dictPartial);
                })()(arr));
            };
            throw new Error("Failed pattern match at Main line 13, column 5 - line 15, column 49: " + [ $1["constructor"]["name"] ]);
        };
    };
    return go(0);
})();
```

michaelficarra:
Fixes #2581. The changes on lines 162 and 167 are unimportant. Turns out the `OperatorTable` prefers later entries over earlier ones. I just had to switch the order of `indexer` and `accessor` to make it work how we intended.

No tests since it's not a semantic change. Let me know if we should add some, and if so, what they should look like.

### DIFF
--- a/src/Language/PureScript/Pretty/JS.hs
+++ b/src/Language/PureScript/Pretty/JS.hs
@@ -159,12 +159,12 @@ conditional = mkPattern match
   match (JSConditional ss cond th el) = Just ((ss, th, el), cond)
   match _ = Nothing
 
-accessor :: (Emit gen) => Pattern PrinterState JS (gen, JS)
+accessor :: Pattern PrinterState JS (Text, JS)
 accessor = mkPattern match
   where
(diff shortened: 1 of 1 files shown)


paf31:
Great, thanks!

PR #2585, opened 2017-01-19
Update `psc-package` to use turtle 1.3

taktoa:
## Notes

I'm guessing you'll want to wait until you are using a version of Stackage LTS with `turtle-1.3.1` before you merge this (unless I introduce some CPP; more on that later), but I'm making this PR now because I know this work will get duplicated if I don't (I'll forget to make the PR later).

There are only [three other packages][packdeps] that are blocking a `turtle` version bump for Stackage LTS, and I'll probably make PRs for those packages too.

I haven't run `stack test` on this source tree, but I did make sure that the output of `echoT` matches that of the old code, e.g.: by running:
```zsh
# Get a blank directory, which will trigger the "psc-package.json does not exist" error.
TEMP="$(mktemp -d --tmpdir)"; cd "$TEMP"
# Execute the version compiled without my changes
OLD="$(psc-package update |& cat)"
# Execute the version with my changes
NEW="$(~/purescript/dist/build/psc-package/psc-package update |& cat)"
if [[ "$OLD" = "$NEW" ]]; then echo "SUCCESS"; else echo "FAILURE"; fi
cd "$OLDPWD"; rmdir "$TEMP" # Should be empty
```
It seems like `psc-package` isn't particularly well-tested anyway, and this is a fairly minor change, so I think this is probably a sufficient amount of testing.

The version bound **could** be loosened to `turtle -any`, if you're willing to use `CPP` in `psc-package/Main.hs`. Since `cd purescript; git grep CPP; git grep '^#' '*.hs'` outputs nothing, I didn't add the code. Note that if I did add the CPP code, this could be merged immediately.

I renamed all the uses of `echo` to `echoT` because otherwise we would be shadowing the new `echo`, which seemed like it would be more confusing to read.

[packdeps]: http://packdeps.haskellers.com/reverse/turtle

## Change summary

* `purescript.cabal`
  * `turtle` constraint is now `== 1.3.*`
  * `optparse-applicative` constraint is now `>= 0.13.0`
* `stack.yaml` and `stack-ghc-8.0.yaml`
  * Added `turtle-1.3.1` and `optparse-applicative-0.13.0.0` to `extra-deps`.
* `psc-package/Main.hs`
  * Added `echoT`, which mimics the old (1.2.8) behavior and type of `Turtle.echo`.
  * Replaced all uses of `echo` with `echoT`.
  * `New.listRemoteTags = Turtle.lineToText <$> Old.listRemoteTags`
* `CONTRIBUTORS.md`
  * Added @taktoa (Remy Goldschmidt)


hdgarrood:
Just in case it wasn't clear, purescript is no longer in Stackage so it shouldn't be holding anything back.

I don't think there's any CPP in purescript right now and I am very keen to keep it that way. I think the options are to modify the stack.yaml files to add turtle as an extra-dep so that this can be built with Stack and merge this, or wait until turtle 1.3 is in the snapshot we're using.

taktoa:
I've opened PRs on the other three repos that break with `turtle-1.3.1`:
* https://github.com/pjrt/Dot-Linker/pull/1
* https://github.com/Gabriel439/bench/pull/11
* https://github.com/sgraf812/cloben/pull/3

paf31:
Looks fine, but can we add `turtle` to extra-deps as @hdgarrood suggested, please?

taktoa:
I added myself to `CONTRIBUTORS.md`.
To get `turtle-1.3.1` to correctly resolve, I also needed to add `optparse-applicative-0.13.0.0` to `extra-deps` as well, and I'm not sure if this will break the build. I guess I should probably also tighten the `optparse-applicative` version bound in `purescript.cabal`.

taktoa:
* Added `extra-deps` to `stack-ghc-8.0.yaml` as well.
* Updated `optparse-applicative` constraint to `>= 0.13.0`, except in places where `-any` was specified.

taktoa:
All the builds seem to succeed now, except for the Mac OS build, which seems to run out of time. The only other concern I would have is that the `optparse-applicative` version bump could have changed the semantics in a way that would not be caught by tests (are there test cases for the option parsers?).

paf31:
Thanks!

PR #2586, opened 2017-01-20
TypeLevelString/TypeConcat should not be quoted

### LINKED ISSUE #2568: Custom error messages printing has changed
The constraint `Fail (TypeConcat "A" "B")` used to print `AB` but now prints `"A""B"`.
I believe this has been introduced since the 0.10.5 release - but I haven't investigated any further.

michaelficarra:
Fixes #2568. No tests, but I manually confirmed by compiling `examples/warning/CustomWarning.purs`. Before this change, the warning would be quoted. After, it is not. If you want me to add a test, let me know how it should be done.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -32,6 +32,7 @@ import           Language.PureScript.Label (Label(..))
 import           Language.PureScript.Names
 import           Language.PureScript.Pretty
 import           Language.PureScript.Pretty.Common (before, endWith)
+import           Language.PureScript.PSString (decodeString)
 import           Language.PureScript.Traversals
 import           Language.PureScript.Types
 import qualified Language.PureScript.Publish.BoxesHelpers as BoxHelpers
@@ -1264,7 +1265,7 @@ renderBox = unlines
   whiteSpace = all isSpace
 
 toTypelevelString :: Type -> Maybe Box.Box
-toTypelevelString (TypeLevelString s) = Just $ Box.text $ T.unpack $ prettyPrintString s
+toTypelevelString (TypeLevelString s) = (Box.text . T.unpack) <$> decodeString s
 toTypelevelString (TypeApp (TypeConstructor f) x)
   | f == primName "TypeString" = Just $ typeAsBox x
 toTypelevelString (TypeApp (TypeApp (TypeConstructor f) x) ret)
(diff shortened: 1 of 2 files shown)


hdgarrood:
It would of course be nice to test this, but I don't think we have a good way of testing error messages with the current test suite, so I'm inclined to leave it without tests for now.

hdgarrood:
What happens if you have a Warn constraint with a type level string which contains lone surrogates?

michaelficarra:
@hdgarrood It will just print the type like any other.

Changing line 4 of `examples/warning/CustomWarning.purs ` to

```hs
foo :: forall t. Warn (TypeConcat "Custom warning \xD800 " (TypeString t)) => t -> t
```

results in

> A custom warning occurred while solving type class constraints:
> 
> &nbsp;&nbsp;&nbsp;&nbsp;TypeConcat "Custom warning \55296 " (TypeString Int)

If you'd like, we could print it instead using Unicode replacement characters or omitting the lone surrogates. Additionally, TypeConcat on a lone high surrogate and lone low surrogate will not result in a surrogate pair, but that doesn't bother me so much. We could also fix that if you'd really like.

michaelficarra:
@hdgarrood I went ahead and pushed an additional commit to add support for combining lone surrogates into surrogate pairs and using the Unicode replacement character when attempting to print lone surrogates.

This type:

```hs
foo :: forall t. Warn (TypeConcat (TypeConcat "Custom warning \xD834" "\xDF06 ") (TypeString t)) => t -> t
```

now prints

> A custom warning occurred while solving type class constraints:
> 
> &nbsp;&nbsp;Custom warning 𝌆 Int

Please review again.

hdgarrood:
Now that we have a use for it, I think I'd rather export a new function from the module which defines PSString rather that implementing it in terms of `decodeStringEither` here. Also if memory serves correctly, isn't the replacement character U+FFFD, not U+FFEF?

michaelficarra:
@hdgarrood I can do that. Also, yes, that was a mistake which I've already corrected.

michaelficarra:
@hdgarrood See https://github.com/purescript/purescript/pull/2586/commits/5dda947f5d69882c8c43658c4a6640d8c4897568.

paf31:
👍 LGTM, thanks!

hdgarrood:
Looks great, would you mind resolving the conflicts and then I think we can merge this?

michaelficarra:
Rebased.

michaelficarra:
Hmm, I'm not sure why Appveyor has failed or why Travis is seemingly stuck forever without starting any builds. I'll try just amending and force-pushing to restart everything.

hdgarrood:
That's weird. Do the tests pass locally for you?

michaelficarra:
@hdgarrood Yes.

michaelficarra:
@hdgarrood Yes. On both of my development machines (Fedora and macOS).

hdgarrood:
In that case I'm tempted to just go ahead and merge this, what do people think?

paf31:
Sounds good.

hdgarrood:
Thanks @michaelficarra!

Issue #1123, opened 2015-05-22
Pattern Guards


andyarvanitis:
I've been finding myself using them more and more: https://wiki.haskell.org/Pattern_guard

paf31:
Can be implemented as sugar. If someone wants to take this on, feel free. Would also be interested in "with-patterns".

paf31:
I've moved this back to Ideas, but I'm in favor.

andyarvanitis:
Great! Funny, I was just thinking about this the other day. I'd like to find some time to implement it if no one else does it first.

PR #2588, opened 2017-01-20
Pattern guards

alexbiehl:
This pull requests aims to implement PatternGuards for Purescript as known from the Haskell language. 

A quick motivation (stolen from [here](http://research.microsoft.com/en-us/um/people/simonpj/Haskell/guards.html)) Currently `clunky` needs to be written like this: 

```haskell
  clunky env var1 var2 | ok1 && ok2 = val1 + val2
                       | otherwise  = var1 + var2
                          where
                            m1 = lookup env var1
                            m2 = lookup env var2
                            ok1 = maybeToBool m1
                            ok2 = maybeToBool m2
                            val1 = expectJust m1
                            val2 = expectJust m2 
```

where `lookup :: FinteMap -> Int -> Maybe Int`. 

Pattern guards let you write this: 
```haskell
  clunky env var1 var2
   | Just val1 <- lookup env var1
   , Just val2 <- lookup env var2
   = val1 + val2
   | otherwise = var1 + var2
```

The desugaring currently happens in `Language.PureScript.Sugar.CaseDeclarations`. The desugaring itself is tricky. You must not introduce unreachable cases or unused variables as the linter goes crazy. Also pattern guards and guards with multiple clauses need additional let-bindings to cover all alternatives  if we wouldn't want to duplicate them several times. 

Please have a look, ask question and start discussion.


paf31:
I haven't read the code fully, but a) thanks for taking this on!, and b) I would have thought it'd be simplest to desugar all guards into nested `case` expressions early on, even before the exhaustivity check. Have you considered that option? Is there something preventing it?

andyarvanitis:
Yay! Requested in issue #1123

src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -354,10 +354,24 @@ moduleToJs (Module coms mn imps exps foreigns decls) foreign_ =
...
+          -- it stops generating code for the rest of the guards.
+          genCondVal False cond val
+              | condIsTrue cond = do
+                  js <- JSReturn Nothing <$> valueToJs val
+                  return (js, True)
+              | otherwise = do
+                  cond' <- valueToJs cond
+                  val'  <- valueToJs val
+                  return $ (JSIfElse Nothing cond' (JSBlock Nothing [JSReturn Nothing val']) Nothing, False)
+          genCondVal True _ _ = return $ (JSRaw Nothing "", True)
```
paf31: Instead of using a `mapAccumL` here and returning `JSRaw` Nothing when done, perhaps use a short-circuiting fold?

src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -202,17 +202,20 @@ missingCasesMultiple env mn = go
...
-  isOtherwise :: Expr -> Bool
-  isOtherwise (Literal (BooleanLiteral True)) = True
-  isOtherwise (Var (Qualified (Just (ModuleName [ProperName "Prelude"])) (Ident "otherwise"))) = True
-  isOtherwise (Var (Qualified (Just (ModuleName [ProperName "Data", ProperName "Boolean"])) (Ident "otherwise"))) = True
-  isOtherwise (TypedValue _ e _) = isOtherwise e
-  isOtherwise (PositionedValue _ _ e) = isOtherwise e
-  isOtherwise _ = False
-isExhaustiveGuard (Right _) = True
+    checkPatGuard :: Binder -> Bool
+    checkPatGuard NullBinder = True
```
paf31: A `VarBinder` would be fine here too, presumably?

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -548,7 +548,11 @@ flattenDecls = concatMap flattenOne
 -- |
 -- A guard is just a boolean-valued expression that appears alongside a set of binders
 --
-type Guard = Expr
+data Guard = ConditionGuard Expr
+           | PatternGuard Binder Expr
+           deriving (Show)
+
+type GuardedExpr = ([Guard], Expr)
```
paf31: Can you please make this into a record type?

src/Language/PureScript/Sugar/CaseDeclarations.hs
```diff
@@ -39,6 +38,186 @@ desugarCasesModule (Module ss coms name ds exps) =
...
+    dsAlt :: (Monad m, MonadSupply m) => [CaseAlternative] -> m [CaseAlternative]
+    dsAlt [] = pure []
+
+    -- the trivial case: no guards
+    dsAlt (a@(CaseAlternative _ [([], _)]) : as) =
+      (a :) <$> dsAlt as
+
+    -- we make sure to clean up all single conditional clauses
+    -- as we would otherwise generate the out-of-line code for
+    -- them which is unecessary.
```
paf31: Typo

src/Language/PureScript/Sugar/CaseDeclarations.hs
```diff
@@ -39,6 +38,186 @@ desugarCasesModule (Module ss coms name ds exps) =
...
+    dsAlt (CaseAlternative ab ge : as)
+      | not (null cond_guards) =
+          (CaseAlternative ab cond_guards :)
+            <$> dsGrdAlt ab rest as
+      | otherwise = dsGrdAlt ab ge as
+      where
+        (cond_guards, rest) = span isSingleCondGuard ge
+
+        isSingleCondGuard ([ConditionGuard _], _) = True
```
paf31: Do we really need to make a special case for single-condition pattern guards?

src/Language/PureScript/Sugar/CaseDeclarations.hs
```diff
@@ -39,6 +38,186 @@ desugarCasesModule (Module ss coms name ds exps) =
...
+        isSingleCondGuard _ = False
+
+    dsGrdAlt :: (Monad m, MonadSupply m)
```
paf31: As a general point, could you please use longer names here? It's hard to infer what these are supposed to be doing without reading the code.

src/Language/PureScript/Sugar/CaseDeclarations.hs
```diff
@@ -39,6 +38,186 @@ desugarCasesModule (Module ss coms name ds exps) =
...
+          alt_fail' | all isVarBinder bs = []
+                    | otherwise = alt_fail
+
+        in Case scrut
+            (CaseAlternative bs [([], dsGrd gs e alt_fail)]
+              : alt_fail')
+
+      return [ CaseAlternative [NullBinder] [([], rhs)]]
+      where
+        isVarBinder :: Binder -> Bool
```
paf31: Maybe `isIrrefutable`?

src/Language/PureScript/Sugar/CaseDeclarations.hs
alexbiehl: Good point. I thought about this too and since we can express conditional guards in CoreFn natively I decided to keep. If we decide not to special case it I think support for guards can be removed in CoreFn.

alexbiehl:
When compiling 

```haskell
module T where

fromJust :: forall a. Partial => Maybe a -> a
fromJust (Just x) = x
```

```
Error found:
in module Main
at /Users/alexbiehl/git/purescript/pat.ps line 57, column 1 - line 57, column 46

  A type variable has escaped its scope.
  Relevant expression:
            
    __unused
            

in the expression \$dictPartial4 -> 
                    \$2 ->          
                      $2            
in value declaration fromJust
``` 

The case desugarer is not in fault here (I checked)  because there are no guards. I think I messed up `Language.PureScript.AST.Traversals` but I can't find the culprit here.

alexbiehl:
A flag for dumping desugared source would be nice. I tried to make one myself but it seems to require major refactoring since `rebuildModule` doesn't have access to `outputDir` and `options`

alexbiehl:
For example this simple module

```
module X where 
data Maybe a = Just a | Nothing
...
```

```
-- ImportDeclaration (ModuleName [ProperName {runProperName = "Prim"}]) Implicit Nothing                                                                                        
-- DataDeclaration Data (ProperName {runProperName = "Maybe"}) [("a",Nothing)] [(ProperName {runProperName = "Just"},[TypeVar "a"]),(ProperName {runProperName = "Nothing"},[])]
...
```

alexbiehl:
I have straightened this out a bit. Just so you have an imagination on how `clunky` is being desugared: 

```haskell
clunky = \var1 ->                                                                                                                                                                
           \var2 ->                                                                                                                                                              
             let                                                                                                                                                                 
               $2 = \$17 ->                                                                                                                                                      
                      case var1 var2 of                                                                                                                                          
                        var1 var2 -> (add var1) var2                                                                                                                             
             in case var1 var2 of                                                                                                                                                
                  var1 var2 -> case (lookup var1) of                                                                                                                             
                                 (Just val1) -> case (lookup var2) of                                                                                                            
                                                  (Just val2) -> (add val1) val2                                                                                                 
                                                  _ -> $2 true                                                                                                                   
                                 _ -> $2 true                                                                                                                                    

```

@paf31 Note the `$2 true`: I use it as kind of `goto` branching to the remaining case expression. Passing `true` is quite a hack but I haven't found a builtin type like `unit` which doesn't require some Prelude module to be defined. GHC for example has `void#` which is a zero-width function value for this purpuse. It is stripped off during code generation.

src/Language/PureScript/CoreFn/Desugar.hs
```diff
@@ -132,9 +132,14 @@ moduleToCoreFn env (A.Module _ coms mn decls (Just exps)) =
...
-    go :: Either [(A.Guard, A.Expr)] A.Expr -> Either [(Guard Ann, Expr Ann)] (Expr Ann)
-    go (Left ges) = Left $ map (exprToCoreFn ss [] Nothing *** exprToCoreFn ss [] Nothing) ges
-    go (Right e) = Right (exprToCoreFn ss [] Nothing e)
+    go :: [A.GuardedExpr] -> Either [(Guard Ann, Expr Ann)] (Expr Ann)
+    go [A.MkUnguarded e]
+      = Right (exprToCoreFn ss [] Nothing e)
+    go gs
+      = Left [ (exprToCoreFn ss [] Nothing cond, exprToCoreFn ss [] Nothing e)
+             | A.GuardedExpr g e <- gs
+             , let [A.ConditionGuard cond] = g -- TODO: check invariant
```
paf31: What does this comment refer to?

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -544,18 +544,32 @@ checkBinders nvals ret (CaseAlternative binders result : bs) = do
...
+  -> Type
+  -> m GuardedExpr
+checkGuardedRhs (GuardedExpr [] rhs) ret = do
+  rhs' <- TypedValue True <$> check rhs ret <*> pure ret
+  return $ GuardedExpr [] rhs'
+checkGuardedRhs (GuardedExpr (ConditionGuard cond : guards) rhs) ret = do
+  cond' <- withErrorMessageHint ErrorCheckingGuard $ check cond tyBoolean
+  GuardedExpr guards' rhs' <- checkGuardedRhs (GuardedExpr guards rhs) ret
+  return $ GuardedExpr (ConditionGuard cond' : guards') rhs'
+checkGuardedRhs (GuardedExpr (PatternGuard binder expr : guards) rhs) ret = do
```
paf31: Won't these have been removed by this point?

src/Language/PureScript/Sugar/CaseDeclarations.hs
```diff
@@ -39,6 +37,196 @@ desugarCasesModule (Module ss coms name ds exps) =
...
+          (CaseAlternative ab cond_guards :)
+            <$> desugarGuardedAlternative ab rest as
+      | otherwise = desugarGuardedAlternative ab ge as
+      where
+        (cond_guards, rest) = span isSingleCondGuard ge
+
+        isSingleCondGuard (GuardedExpr [ConditionGuard _] _) = True
+        isSingleCondGuard _ = False
+
+    desugarGuardedAlternative :: (MonadSupply m)
```
paf31: Why not use the same `m` as above here, using `ScopedTypeVariables`?

src/Language/PureScript/TypeChecker/Types.hs
alexbiehl: Good catch! I guess we don't need changes the type checker at all!

src/Language/PureScript/Sugar/CaseDeclarations.hs
alexbiehl: Right, oversight.

src/Language/PureScript/CoreFn/Desugar.hs
alexbiehl: I think I meant to give some meaningful error when pattern guard is encountered.

src/Language/PureScript/Sugar/CaseDeclarations.hs
```diff
@@ -213,13 +233,21 @@ desugarCase (Case scrut alternatives) =
     scrut_nullbinder :: [Binder]
     scrut_nullbinder = replicate (length scrut) NullBinder
 
+    -- Case expressions with a single alternative whichs has
```
vendethiel: typo: "whichs"

paf31:
@alexbiehl I think this looks good to merge soon, but could you please add a few examples under `examples/passing`, and if appropriate, under `examples/failing` as well? This way, we can exercise the pattern guards code while running the test suite.

BebeSparkelSparkel:
Has anyone considered view patterns? It seems like the implementation would be similar.

PR #2589, opened 2017-01-21
fix tests

kRITZCREEK:
Seems like a new release of purescript-arrays added a new module


hdgarrood:
I've managed to get the cache set up before by just commenting out the actual build script, so that it installs all the dependencies and then exits with a success. The build cache only gets stored if the build succeeds, AFAICT.

PR #2590, opened 2017-01-21
Abstraction for dumping intermediate representations

alexbiehl:
While implementing https://github.com/purescript/purescript/pull/2588 I often need to view desugared source. My solutions until now were rather ad-hoc and I needed to be careful not to commit that trace messages. I hope I can convince you this is useful.

### DIFF
--- a/src/Language/PureScript/Pretty/Values.hs
+++ b/src/Language/PureScript/Pretty/Values.hs
@@ -5,6 +5,7 @@ module Language.PureScript.Pretty.Values
   ( prettyPrintValue
   , prettyPrintBinder
   , prettyPrintBinderAtom
+  , prettyPrintDeclaration
   ) where
 
 import Prelude.Compat
@@ -134,7 +135,7 @@ prettyPrintDeclaration d (BindingGroupDeclaration ds) =
   where
   toDecl (nm, t, e) = ValueDeclaration nm t [] (Right e)
 prettyPrintDeclaration d (PositionedDeclaration _ _ decl) = prettyPrintDeclaration d decl
-prettyPrintDeclaration _ _ = internalError "Invalid argument to prettyPrintDeclaration"
+prettyPrintDeclaration _ x = text ("-- " ++ show x)
 
 prettyPrintCaseAlternative :: Int -> CaseAlternative -> Box
 prettyPrintCaseAlternative d _ | d < 0 = ellipsis
--- a/psc/Main.hs
+++ b/psc/Main.hs
@@ -144,6 +144,16 @@ sourceMaps = switch $
      long "source-maps"
   <> help "Generate source maps"
 
+dumpParsed :: Parser Bool
+dumpParsed = switch $
+     long "dump-parsed"
+  <> help "Dump the abstract syntax tree of the compiled code at output/*/parsed.dump-parsed"
+
+dumpDesugared :: Parser Bool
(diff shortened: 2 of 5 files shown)


src/Language/PureScript/Pretty/Values.hs
```diff
@@ -134,7 +135,7 @@ prettyPrintDeclaration d (BindingGroupDeclaration ds) =
   where
   toDecl (nm, t, e) = ValueDeclaration nm t [] (Right e)
 prettyPrintDeclaration d (PositionedDeclaration _ _ decl) = prettyPrintDeclaration d decl
-prettyPrintDeclaration _ _ = internalError "Invalid argument to prettyPrintDeclaration"
+prettyPrintDeclaration _ x = text ("-- " ++ show x)
```
hdgarrood: Could you explain the purpose of this change please?

alexbiehl: I like the desugared ast to be pretty printed. But `prettyPrintDeclaration` only implements half of the declarations. So if it encounters an unimplemented declaration it just switches to `show` here.

paf31:
> ideally I think we would use JSON instead

I agree regarding `Show` instances, but why not dump the desugared output as actual PS source code? It is a `Module` after all. For the parsed AST, do we even need that option? What's the use case for it? We can `show` a module easily in GHCi after all.

Thanks for working on this @alexbiehl, it seems like it'll be really useful!

alexbiehl:
@paf31 yes, that what I did for the desugared output already. To remove the `show` output we need to complete the pretty printer.

paf31:
Can you show some example output please, so that I can understand what sort of declarations are missing?

alexbiehl:
For example this simple module

```
module X where 
data Maybe a = Just a | Nothing
...
```

```
-- ImportDeclaration (ModuleName [ProperName {runProperName = "Prim"}]) Implicit Nothing                                                                                        
-- DataDeclaration Data (ProperName {runProperName = "Maybe"}) [("a",Nothing)] [(ProperName {runProperName = "Just"},[TypeVar "a"]),(ProperName {runProperName = "Nothing"},[])]
...
```

paf31:
I wonder if we should wait until `psc-format` is ready, at which point we could print out all desugared declarations. @hdgarrood You were working on that recently, right?

hdgarrood:
I was, but I heard you were working on a new AST which ensured that there were source spans at every level, so I decided to put my work on psc-format on hold temporarily.

paf31:
@hdgarrood Actually, I stopped working on that. I might get back to it soon, but it turned out to be a much larger project than I anticipated.

alexbiehl:
So, how do we proceed here? Making `prettyPrintDeclaration` correctly print any type of declaration?

paf31:
I think I would prefer to wait for `purs format` to be a thing, and then to use the same pretty printer.

hdgarrood:
Re `purs format`: now that I'm back at uni my work on that has stalled a bit. I've written a short progress report at https://github.com/purescript/roadmap/issues/23#issuecomment-270561451. I'm very happy to help if anyone wants to try to pick this up again.

paf31:
Ok, let's close this then until `purs format` is a bit further along. The PR will still exist in the history in case anyone wants to pick it up then. Thanks @alexbiehl.

Issue #2520, opened 2017-01-01
psc-package megathread


paf31:
- Do people find `psc-package` to be useful compared to Bower?
- Is `psc-package` the Bower replacement we want?
- Can we continue to support Bower for alternative workflows?
- What is needed to increase adoption of `psc-package` before 1.0?

Pauan:
* Yes.

* Yes.

* I'd rather drop Bower support and focus on making `psc-package` really good, especially because Bower is unmaintained.

  I don't mind people using Bower (it's their project after all), I just don't think it should be officially supported.

* Pulp support, and evangelism: we need a blog post letting everybody know, an announcement on IRC, an announcement on Gitter, etc.

  The book also needs to be updated to use `psc-package`

----

Something that I noticed (which is also a problem with Bower) is that it's possible for a project to use a dependency even if it's not listed in `psc-package.json`

As an example, let's suppose that your project uses a module which is defined in the `foo` package, but `foo` isn't listed in `psc-package.json`. Instead, `psc-package.json` has a dependency on `bar`, and `bar` has a dependency on `foo`

That means that if `bar` is changed so that it no longer has a dependency on `foo`, then your project will break. I think your project should only have access to packages which are explicitly listed in `psc-package.json` (and dependencies should only have access to their explicitly listed sub-dependencies)

Pauan:
Also, the various IDE integrations (like [`atom-ide-purescript`](https://github.com/nwolverson/atom-ide-purescript)) need to be updated to work with `psc-package`. Hopefully it's as simple as just changing the default file glob.

paulyoung:
I don't think any of the following is news to anyone, but to perhaps offer a different perspective –

Having been working on alternative backends I think it makes more sense to use `psc-package` which ships with the compiler rather than asking users to install `bower` which is completely foreign to the target ecosystem.

I think there has been some discussion around how to depend on a package (or perhaps a version of a package) that doesn't exist in a package set. I'm not sure if that got resolved but it seems important for development purposes for `psc-package` to have an alternative to `bower link`.

andyarvanitis:
I've mentioned it elsewhere, but for the C++ backend/target, I'm currently using `psc-package` exclusively for package management. I have it integrated with my standard GNU Make `Makefile`.

hdgarrood:
[...]

PR #2591, opened 2017-01-21
Add modules for rendering HTML documentation

hdgarrood:
Refs #2520. This code is more or less copied from Pursuit, although I
managed to drop the `hxt` dependency by instead using Cheapskate's
provided functions for walking a rendered Markdown document.

This is just a starting point towards generating HTML documentation from
a given package set. Next, I plan to make psc-docs capable of producing
HTML documentation as well as Markdown.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -117,9 +117,11 @@ library
                    aeson-better-errors >= 0.8,
                    ansi-terminal >= 0.6.2 && < 0.7,
                    base-compat >=0.6.0,
+                   blaze-html >= 0.8.1 && < 0.9,
                    bower-json >= 1.0.0.1 && < 1.1,
                    boxes >= 0.1.4 && < 0.2.0,
                    bytestring -any,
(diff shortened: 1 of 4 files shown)


purescript.cabal
```diff
@@ -117,9 +117,11 @@ library
                    aeson-better-errors >= 0.8,
                    ansi-terminal >= 0.6.2 && < 0.7,
                    base-compat >=0.6.0,
+                   blaze-html >= 0.8.1 && < 0.9,
                    bower-json >= 1.0.0.1 && < 1.1,
                    boxes >= 0.1.4 && < 0.2.0,
                    bytestring -any,
+                   cheapskate >= 0.1 && < 0.2,
```
paf31: Do we need to update the license file for this?

paf31:
👍 Looks good, thanks!

purescript.cabal
hdgarrood: Ah, yes we do

PR #2592, opened 2017-01-22
[psc-ide] Collect type class instances

### LINKED ISSUE #2350: [psc-ide] Collect instances for a typeclass

kRITZCREEK:
The main addition of this PR is to collect all instances from the ExternsFiles and add them to their respective type class declaration inside Stage3. While I was at it I cleaned up the different resolving steps for Stage3 into a more uniform pipeline of functions.

There is currently no way to access the gathered information through psc-ide's commands, but I'd like to think about how to expose the information without bastardizing the stringly typed completion format we have right now.

Fixes #2350


src/Language/PureScript/Ide/Util.hs
```diff
@@ -129,3 +131,6 @@ prettyTypeT =
   . T.lines
   . T.pack
   . P.prettyPrintTypeWithUnicode
+
+lensSatisfies :: forall a s. Getting (First a) s a -> (a -> Bool) -> s -> Bool
```
paf31: I would probably use `anyOf` or `allOf` for this.

PR #2593, opened 2017-01-22
[psc-ide] Fix #2504

### LINKED ISSUE #2504: psc-ide-server crashes once project gets a little bigger
We are using spacemacs with purescript layer. We have are in the process of building a React Native Mobile app with PureScript.

It used to work fine earlier, nowadays it crashes after sometime. When we start it uses 100% cpu for a few seconds and then it crashes once we make changes to the code in emacs. We started adding more files and then it started crashing.

May be its a memory leak or something?

kRITZCREEK:
When commands started processing and the client then disconnected during the process the server crashed when trying to write to a vanished Handle. This catches the error, logs and ignores it.

@sudhirvkumar could you try building this branch and check if this stops the crashes you're experiencing inside Emacs?

### DIFF
--- a/psc-ide-server/Main.hs
+++ b/psc-ide-server/Main.hs
@@ -29,6 +29,7 @@ import           "monad-logger" Control.Monad.Logger
 import qualified Data.Text.IO                      as T
 import qualified Data.ByteString.Lazy.Char8        as BS8
 import           Data.Version                      (showVersion)
+import           GHC.IO.Exception                  (IOErrorType(..), IOException(..))
 import           Language.PureScript.Ide
 import           Language.PureScript.Ide.Command
 import           Language.PureScript.Ide.Util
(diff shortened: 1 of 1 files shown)


sudhirvkumar:
@kRITZCREEK Thanks a lot. Will try to build and check.

sudhirvkumar:
@kRITZCREEK sorry about not getting back to you earlier. I am yet to find some free time to build purescript and check. It would be super useful. If there was a nightly build available to downlod. Would make testing things super easy.

kRITZCREEK:
@sudhirvkumar no worries. I've recently had a bit of time to play around with PS and it seems like this fix worked in combination with a recent fix on psc-ide-emacs. You might want to make sure you're running the last version of the emacs plugin when testing this.

sudhirvkumar:
@kRITZCREEK Thanks will check it out with the latest version of emacs plugin.

sudhirvkumar:
@kRITZCREEK I am thinking of waiting for 0.10.6 which is due this week ;) @paf31 has mentioned in another thread about 0.10.6 release this week.

sudhirvkumar:
@kRITZCREEK I tested the binaries and looks like the emacs issue is solved! Thanks for all the effort!

Issue #2514, opened 2016-12-30
Comment style


spicydonuts:
It seems a bit silly in a typed, compiled language like PureScript, but in the dynamic world of JavaScript comments have become a common way to provide hints to code analysis tools.  A project I'm working on is using Istanbul for code coverage reports and there's a lot of compiler-generated code I'd like to skip.  For example:

```purescript
-- istanbul ignore next
derive instance genericGUID :: Generic GUID
```

That comment doesn't work because Istanbul is expecting a comment in exactly this form:

```javascript
/* istanbul ignore next */
```

The "next" statement or expression can be on the following line or immediately to the right of the comment after a space:

```javascript
/* istanbul ignore next */
foo
```

```javascript
/* istanbul ignore next */ foo
```

Currently a PS comment always gets transformed to a block comment in JS:

```purescript
-- some insightful comment
someCode ...
```

```javascript
/**
 *  some insightful comment
 */
someCompiledCode ...
```

This works most of the time, but there are some JS analysis tools which use comments as hints or extra metadata which are more picky about the format.

Below are some potential ways comment format could be controlled.

Single line, until end of line:

```purescript
-- comment
```

```javascript
// comment
```

They would not automatically be transformed to blocks:
```purescript
-- comment
-- another comment
```

```javascript
// comment
// another comment
```

Single line, block (`--|` would be nice, but it would change the behavior of existing code):

```purescript
--| comment |--
```

```javascript
/* comment */
```

Multi line, block:

```purescript
--| comment
--| more comments
--| these need to line up
--| to become a single block comment
```

```javascript
/**
 *  comment
 *  more comments
 *  these need to line up
 *  to become a single block comment
 */
```

Again, I know from a PS perspective the details of comment formatting seem a bit silly, but one of the main draws of PS is how close it is to its main compile target of JS.  You can include a little PS in almost any JS project.  Allowing this control when it's needed makes existing JS-analysis tools usable in PS projects.

hdgarrood:
I won't take a stance on this issue but if we were going to do this I'd say it would be more appropriate to translate PureScript line comments into JS line comments, and PS block comments into JS block comments:

```purescript
-- this is a line comment

{- this is a block comment
  it can span multiple lines -}
```
==>
```javascript
// this is a line comment

/* this is a block comment
  it can span multiple lines */
```

I think it would be best not to overload the `--|` syntax as currently that just means a line comment which should be included in generated documentation such as Pursuit.

paf31:
[...]

PR #2596, opened 2017-01-23
comment-styles: style comment types differently

### LINKED ISSUE #2154: Optimize tail calls modulo cons
This Boolean-returning function is not tail-recursion optimized:

```
    go fx fy | fx == w   = go 0 (fy + 1)
             | fy == h   = true
             | otherwise = something fx fy && go (fx + 1) fy
```

However, since `&&` is lazy, this should be tail-recursion optimizable, right? Because `x && f y` is equivalent to `if not x then false else f y`, where `f y` is clearly in tail position.

matthewleon:
closes #2154

In JS output, style single line and block comments differently.

As this is my first PR on this project, I naturally expect it to require some revision.

### DIFF
--- a/src/Language/PureScript/Pretty/JS.hs
+++ b/src/Language/PureScript/Pretty/JS.hs
@@ -122,22 +122,30 @@ literals = mkPattern' match'
     [ return $ emit $ lbl <> ": "
     , prettyPrintJS' js
     ]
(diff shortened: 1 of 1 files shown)


spicydonuts:
Did you mean to reference #2514?

matthewleon:
@spicydonuts Yes. I'll redo this PR. Thanks.

PR #2597, opened 2017-01-23
comment-styles: style comment types differently

### LINKED ISSUE #2514: Comment style

matthewleon:
closes #2514

In JS output, style single line and block comments differently.

This is my first PR on this project, so I expect it to require some revision.


paf31:
Could you please also update `CONTRIBUTORS.md`?

I think this looks fine, as long as the tests pass 😄

matthewleon:
@paf31 done. thank you.

paf31:
Thanks!

PR #2602, opened 2017-01-27
[psc-ide] Also detect location information for operators

kRITZCREEK:
Bug found by @kwohlfahrt

With this change go-to-definition also works for type and value operators and jumps to their respective fixity declaration.


kRITZCREEK:
Yes I added them during debugging, but I figured derived Show instances are always good to have so I put them in.

kRITZCREEK:
Actually I just realized that I don't need the supporting PureScript file for tests anymore after the refactoring of the test suite so I removed that as well and squashed again.

PR #2603, opened 2017-01-27
[psc-ide] fix 2537

### LINKED ISSUE #2537: psc-ide shouldn't crash when rebuilding a non-existent file
See https://github.com/nwolverson/atom-ide-purescript/issues/140

kRITZCREEK:
Use a safer readFile that checks the files existence before reading it and utilizes psc-ide's error reporting instead of crashing.

Also renames `PscIdeError` to `IdeError` to be more consistent with naming

Edit: fixes #2537

### DIFF
--- a/src/Language/PureScript/Ide.hs
+++ b/src/Language/PureScript/Ide.hs
@@ -43,7 +43,7 @@ import           System.FilePath.Glob (glob)
 
 -- | Accepts a Commmand and runs it against psc-ide's State. This is the main
 -- entry point for the server.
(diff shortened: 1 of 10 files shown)


hdgarrood:
I think this implementation has a race condition - the file could disappear in between checking its existence and attempting to read it. I would probably just go ahead and attempt to read it and then use catch from Control.Exception to handle the case where it does not exist.

kRITZCREEK:
Good thought. Does this look better?

PR #2605, opened 2017-01-29
Revert binder syntax

paf31:
This reverts (the key part of) #2559, since it's a breaking change and I shouldn't really have merged it to begin with. I think we should add it back before 0.11.

### DIFF
--- a/examples/passing/ParseNamedBinder.purs
+++ b/examples/passing/ParseNamedBinder.purs
@@ -1,9 +1,0 @@
-module Main where
-
-import Control.Monad.Eff.Console (log)
-
-data X = X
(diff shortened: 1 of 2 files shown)


paf31:
If someone could please review this, I'll make a 0.10.6 release this week.

hdgarrood:
Do you think there will be an 0.10.7 or do you want to start thinking about 0.11.0 after 0.10.6?

paf31:
I think we have enough issues tagged with `breaking` that we could reasonably make a 0.11 release next. In any case, I'd like to make a milestone for it to figure out what we want to break.

Issue #2538, opened 2017-01-05
Add a HasCallStack constraint to internalError?


hdgarrood:
This would require dropping support for GHC 7.x but it could be quite useful.

paf31:
Sounds useful, I think dropping 7.x support would be okay, maybe at 0.11 though?

hdgarrood:
Yep, sounds good to me.

PR #2608, opened 2017-01-29
Add HasCallStack to internalError

alexbiehl:
This might be useful in case of a crash.


hdgarrood:
The line endings on the diff look a bit funny - could you check to see if this has a mixture of Linux and windows style line endings?

hdgarrood:
This fixes #2538. I generally like to avoid CPP but this seems like a good reason to use it. :+1: once line endings are sorted.

alexbiehl:
Updated the patch. This even works for GHC-7.x. They won't print a callstack though.

paf31:
Looks good, thanks! Please can you update `contributors.md` here as well.

paf31:
Thanks!

PR #2611, opened 2017-01-31
Add stack-ghc-8.0.2.yaml

noraesae:
In macOS Sierra, there has been an issue concerning `malformed mach-o`. The issue's detail is well-documented in the following thread.

- https://github.com/commercialhaskell/stack/issues/2577

The problem is solved in GHC 8.0.2, and it is released as Stackage Nigthly for the time being. It will be [available in LTS in 2 weeks](http://lwm.github.io/stackage-8.0.2/).

This PR added `stack-ghc-8.0.2.yaml` to enable macOS Sierra build for the time being. I open this PR to help others to at least build on their local env. About merge, we may do followings.

1. We can merge this, and update `stack-ghc-8.0.2.yaml` again when the LTS is released.
1. We can wait until the LTS is released, update PR then, and merge.

Also, I just added `stack-ghc-8.0.2.yaml` in this PR, but not sure if we should keep `stack-ghc-8.0.yaml`. I think we can just replace the old.


hdgarrood:
Looks good to me - I think it probably makes the most sense to replace the current stack-ghc-8.0.yaml.

noraesae:
Thanks for your comment. I will update the PR in a while.

hdgarrood:
Actually come to think of it, what are the errors you get if you don't hide <|>?

noraesae:
`Ambiguous occurrence ‘<|>’` occurs. It's because both `Protolude` and `Data.Aeson.BetterErrors` are exporting `<|>`.

Now I noticed that it may be better to just use `aeson-better-errors-0.9.0.1`?

hdgarrood:
Where is the occurrence?

noraesae:
Here: https://github.com/noraesae/purescript/blob/b132aa3dbd9215bde167f951ea73cd955518dfae/src/Language/PureScript/Docs/Types.hs#L408

Sorry that I am in another env and building again. I'd better pasting the error message itself when it's done.

hdgarrood:
Ah, right, thanks. Don't worry, I just wanted to check that we do want Alternative's `<|>` and not aeson-better-errors' there, and this does seem to be the case. I think it makes sense to stick with aeson-better-errors 0.9.1.0 because that's the one that's in Stackage LTS 7 and Nightly.

noraesae:
Thanks. Just replaced stack-ghc-8.0.yaml with stack-ghc-8.0.2.yaml.

hdgarrood:
I'd suggest calling the new GHC 8 stack.yaml `stack-ghc-8.0.yaml` like before: then we won't need to update things like CI scripts, and if people have set environment variables for working on purescript, those will continue to work.

noraesae:
Sorry for the ambiguity. It is exactly how I updated. The original `stack-ghc-8.0.yaml` is updated to use the new GHC.

hdgarrood:
Ah oops, I must have not been paying attention when I looked at the diff. Thanks! :smile:

noraesae:
No worries. Thanks for your review!

garyb:
It's going to be so great to be able to work on the compiler with my laptop again... thanks!

hdgarrood:
I've just realised one more thing: we should update the lower bound on `aeson-better-errors` in `purescript.cabal` to be 0.9.1.0, so that `cabal` won't try to choose it any more, because `import Mod hiding (x)` will cause a compile error if `Mod` does not actually export x.

In fact I guess it might be better still to switch that open import for an explicit import; that way we don't need to tighten the dependency bounds at all.

noraesae:
For example, [src/Language/PureScript/Kinds.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Kinds.hs) imports `Data.Aeson.BetterErrors` with explicit import. Would it be better to make other imports explicit? `Data.Aeson.BetterErrors` is currently imported in the following modules.

- src/Language/PureScript/Docs/RenderedCode/Types.hs
- src/Language/PureScript/Docs/Types.hs
- src/Language/PureScript/Kinds.hs
- src/Language/PureScript/Publish/ErrorsWarnings.hs
- src/Language/PureScript/Publish.hs

If it seems okay, I will upload an additional commit.

hdgarrood:
Yeah, I think it would be good to update all of those which don't already use an import of the form `import Data.Aeson.BetterErrors (x, y, z, ...)` so that they do. Thanks!

hdgarrood:
Although if one of those modules already uses qualified imports there's no need to change that; as you probably already know, it's just `import Data.Aeson.BetterErrors` and `import Data.Aeson.BetterErrors hiding (...)` that can cause this kind of problem.

noraesae:
Just uploaded the additional commit. None of them was imported as qualified, so just made them explicit.

noraesae:
No content change, name and email in commits were wrong, so I updated them.

paf31:
Thanks!

PR #2617, opened 2017-02-01
Return operators in psc-ide imports list

nwolverson:
@kRITZCREEK

### DIFF
--- a/src/Language/PureScript/Ide/Types.hs
+++ b/src/Language/PureScript/Ide/Types.hs
@@ -230,6 +230,7 @@ identifierFromDeclarationRef (P.ValueRef ident) = P.runIdent ident
 identifierFromDeclarationRef (P.TypeClassRef name) = P.runProperName name
 identifierFromDeclarationRef (P.KindRef name) = P.runProperName name
 identifierFromDeclarationRef (P.ValueOpRef op) = P.showOp op
+identifierFromDeclarationRef (P.TypeOpRef op) = P.showOp op
 identifierFromDeclarationRef _ = ""
 
 data Success =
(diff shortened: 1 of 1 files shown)


kRITZCREEK:
Is there a test that should've failed because of this? Just not entirely sure what this fixes.

nwolverson:
Sure, I can add a test. Currently it returns "" for operators in imports, and this fixes the fact that I don't get any operators when I ask psc-ide for imports :)

kRITZCREEK:
Oh I see it's about the `list import` command thing, I had almost forgotten about that :D Well you probably want to add type level operators then as well? If you don't need them right now or you want to add them later I'm happy to merge this.

nwolverson:
I can add those too. This is all a bit odd to me, I'd expect firstly not to translate to empty string but to omit any ignored entries from the list, and then also I was surprised this isn't already done elsewhere in the compiler (but the only place I saw does have a little different syntax, so makes sense).

kRITZCREEK:
I've never used that piece of the API in the emacs plugin so I think it just survived from the very start but hasn't gotten any real attention from me. I'm happy to make it a target for 0.12/1.0 to do the import parsing in a robust fashion.

nwolverson:
Honestly there's a higher-level requirement that I'm satisfying by using the imports API, these are used to filter explicit identifiers and modules for various queries to the current file (so e.g. I get the type only of the identifier explicitly in scope), and this is the main piece of state I keep in the plugin - the better solution would probably be to defer to psc-ide to filter the query results by "things in scope of file X"

nwolverson:
You're right, type level operators also apply the same way. Added.

kRITZCREEK:
Okay I'd like to do this in a more robust way soon, but this is a net worth plus over the current situation so I'll just merge it.

PR #2619, opened 2017-02-03
Generate data constructors without IIFEs

### LINKED ISSUE #2206: Avoid IIFE definitions in generated JavaScript
Tagging this with discussion, because I'm not sure if doing so is problematic.

An example of the offending code:

``` javascript
var Just = (function () {
    function Just(value0) {
        this.value0 = value0;
    };
    Just.create = function (value0) {
        return new Just(value0);
    };
    return Just;
})();

var Nothing = (function () {
    function Nothing() {
    };
    Nothing.value = new Nothing();
    return Nothing;
})();
```

The IIFE exists so that the definition of `Just` and `Nothing` do not involve statements at the module level (the `create` / `value` property assignments). In theory this makes them _easier_ to erase, since the whole value can be eliminated, but apparently Uglify considers this to be side-effectful. I assume it takes that stance with any IIFE definition, since here only values in the internal scope of the IIFE are referenced.

I actually don't remember exactly why I did introduce the IIFE... it may purely be to avoid the module-level statements, or perhaps codegen can only produce single AST values for declarations. The latter wouldn't be too much of a problem to change, it's just a bit of refactoring.

/cc @AppShipIt @spicydonuts

hdgarrood:
Fixes #2206.

I've manually tested that comments are preserved in the way we want them to be. I haven't yet gotten around to testing how this affects `psc-bundle`, that still needs doing.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -138,8 +138,8 @@ moduleToJs (Module coms mn imps exps foreigns decls) foreign_ =
   -- Generate code in the simplified Javascript intermediate representation for a declaration
   --
   bindToJs :: Bind Ann -> m [JS]
-  bindToJs (NonRec ann ident val) = return <$> nonRecToJS ann ident val
-  bindToJs (Rec vals) = forM vals (uncurry . uncurry $ nonRecToJS)
+  bindToJs (NonRec ann ident val) = nonRecToJS ann ident val
+  bindToJs (Rec vals) = concat <$> forM vals (uncurry . uncurry $ nonRecToJS)
 
   -- |
   -- Generate code in the simplified Javascript intermediate representation for a single non-recursive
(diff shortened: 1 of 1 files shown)


paf31:
Could you please paste some sample output, to make it easier to read the new code? Thanks.

hdgarrood:
Sure, will do the next time I'm at a pc, (which won't be until Monday).

paf31:
@hdgarrood No worries, I can build this branch later today.

paf31:
Here is the output of `Data.Maybe`:

```purescript
var Nothing = function Nothing() {

};
Nothing.value = new Nothing();
function Just(value0) {
    this.value0 = value0;
};
Just.create = function (value0) {
    return new Just(value0);
};
```

This looks good to me, but I'll mark it as Needs More Work, since you mentioned you need to test with `psc-bundle`.

Thanks!

hdgarrood:
Ok, I've tested with `psc-bundle`, and it does work. I tested that data constructors are completely removed if they are unused, and also that the statements setting the `value` or `create` property on data constructors are preserved in the bundle. As far as I can tell that should be sufficient.

hdgarrood:
:tada: thanks!

Issue #2618, opened 2017-02-03
Remove command line arguments which should trigger rebuild or are unnecessary


hdgarrood:
Some of `psc`'s command line arguments affect what is written to the `output` directory; for example, `--comments`, `--dump-corefn`. If you run `pulp build` followed by `pulp build -- --comments`, the second time, nothing will happen as `psc` seems to consider everything in the `output` directory up-to-date. I was expecting see everything get rebuilt with comments.

Would it make sense to store the command-line arguments inside `externs.json` and check that those are the same when deciding whether to rebuild?

paf31:
Just a note that `--dump-corefn` should not require a full rebuild, since it only results in creating strictly more output.

I would actually also be fine with removing some of these options before 1.0. There are existing tools for stripping comments, for example, so why not just always include comments? And things like `--no-tco` really don't need to be there any more.

hdgarrood:
I think it would be great to remove `--comments` and `--no-tco` :)

paf31:
I changed the title and moved this to 0.11.

natefaubion:
Should we remove `no-magic-do` and `no-opts` as well? I'd imagine there are programs that would break without magic-do (due to stack around binds and ST behavior). I don't see any reason to have these flags except maybe if you were debugging their implementations.

paf31:
Yes, I'd like to remove all of those before 1.0.

hdgarrood:
There's still the issue that building with `--dump-corefn` doesn't actually write the corefn.json if that module has already been built, though, unless we want to always dump corefn?

paf31:
Ah right. I would be fine with always dumping out the core, but it might affect compilation speed, so maybe we should just fix the timestamp check to include the `corefn.json` file.

PR #2623, opened 2017-02-05
Rebuild modules if necessary when using --dump-corefn

paf31:
See #2618

This rebuilds a module if `--dump-corefn` is set but `corefn.json` does not exist.

### DIFF
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -326,10 +326,15 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
 
   getOutputTimestamp :: ModuleName -> Make (Maybe UTCTime)
   getOutputTimestamp mn = do
+    dumpCoreFn <- asks optionsDumpCoreFn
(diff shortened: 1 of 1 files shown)


paf31:
Perhaps. I've been trying to emulate `make` in the sense that we create files which don't exist depending on the timestamp information, but we could potentially delete files too if need be. I'm generally wary of deleting files on the user's behalf though, and they can always clean the output directory if need be.

hdgarrood:
I think it's much better than leaving stale files around, personally. I think if any issue is solved by deleting the `output` directory, assuming that the user hasn't been poking around in there manually, that should be considered an indication of a bug in `psc`. Relatedly, I don't think it's good UX that as things are now, you occasionally have to run `rm -r output`.

paf31:
That bug should've been fixed a while ago. There should be no need to have to wipe out the output directory anymore in order to get a good build. If that is necessary then it's a bug.

hdgarrood:
Ok, thanks. I do still see it occasionally. I'll try to put together repro steps.

Issue #2445, opened 2016-11-24
Crash when annotating a constructor binder with a quantified type


LiamGoodacre:
The following code:

```purescript
eg = \(Proxy :: (forall a. Proxy a)) -> Proxy
```

Causes this:

```
psc: An internal error ocurred during compilation: "unifyTypes: unspecified skolem scope"
Please report this at https://github.com/purescript/purescript/issues
CallStack (from HasCallStack):
  error, called at src/Language/PureScript/Crash.hs:8:3 in purescript-0.10.2-Hp2rHRU6NraBMzRyfJIcIX:Language.PureScript.Crash
psc: thread blocked indefinitely in an MVar operation
```

Although it's not necessarily a sensible thing to write, it probably shouldn't crash 😄

joneshf:
What is the expected behavior?

hdgarrood:
I think we would expect the "a type variable has escaped its scope" error in this case, right?

LiamGoodacre:
No I think this should be valid.  At least it's valid in GHCi.

hdgarrood:
Ah ok, so we should get `eg :: (forall a. Proxy a) -> (forall a. Proxy a)`?

LiamGoodacre:
Yes. Well, `forall b. (forall a. Proxy a) -> Proxy b`, but same thing 😄

PR #2624, opened 2017-02-05
Disallow polymorphic types in binders, prevent internalError

paf31:
@LiamGoodacre Could you please review? This "fixes" #2445, although I'm deferring a proper fix until after I dust off the skolem escape check code and fix a few issues there.

_Edit_: I think we should open a separate issue to deal with polymorphic types in typed binders more generally.


LiamGoodacre:
This looks fine.  I've been playing with it for a little while.

I also noticed that everything works properly when using aliases:
```purescript
type Empty f = forall t. f t
eg (a :: Empty Array) = a
```

paf31:
Thanks Liam

PR #2367, opened 2016-10-06
Instantiate polymorphic types in record literals

paf31:
Fixes #1110

Issue #2312, opened 2016-09-17
Restrict unification to monotypes


paf31:
Extracted from #1719

paf31:
This is another thing I'd like to break in 0.11, but we'd need to have a workaround for the skolem escape issue in Halogen. cc @garyb @natefaubion

PR #2625, opened 2017-02-05
Instantiate types in record literals as necessary

### LINKED ISSUE #1110: Inferred types of record fields should not be polymorphic
``` purescript
module Bug where

-- from purescript-maps
foreign import data StrMap :: * -> *
foreign import empty "" :: forall a. StrMap a
foreign import singleton "" :: forall a. String -> a -> StrMap a

-- from my code
type R = {f :: StrMap Number}

funA :: forall m . (Monad m) => m R
--funA = return {f:singleton "" 0} -- GOOD
funA = return {f:empty} -- BUG with this message:
```

``` text
Error in module Bug:
Error in value declaration funA:
Error checking that type
  u1499 u1498
subsumes type
  m1497 { f :: Bug.StrMap Prim.Number }
Error at bug2.purs line 17, column 1 - line 17, column 23:
  Cannot unify type
    a1501
  with type
    Prim.Number
See https://github.com/purescript/purescript/wiki/Error-Code-ErrorUnifyingTypes for more information, or to contribute content related to this error.
```

paf31:
Fixes #1110 and #2312. Replaces #2367.

I'm not really very happy with this fix, but I've been stuck thinking about this for a long time now, and it's been difficult to fix this and keep Halogen compiling as-is. So this is a compromise, and I've added a note in a comment to explain the fix.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -241,7 +241,7 @@ checkTypedBindingGroupElement mn (ident, (val, ty, checkType)) dict = do
   ty' <- introduceSkolemScope <=< replaceAllTypeSynonyms $ ty
   -- Check the type with the new names in scope
   val' <- if checkType
-            then withScopedTypeVars mn args $ bindNames dict $ TypedValue True <$> check val ty' <*> pure ty'
+            then withScopedTypeVars mn args $ bindNames dict $ check val ty'
             else return (TypedValue False val ty')
   return (ident, (val', ty'))
 
@@ -297,7 +297,8 @@ infer val = withErrorMessageHint (ErrorInferringType val) $ infer' val
 
 -- | Infer a type for a value
 infer'
-  :: (MonadSupply m, MonadState CheckState m, MonadError MultipleErrors m, MonadWriter MultipleErrors m)
+  :: forall m
+   . (MonadSupply m, MonadState CheckState m, MonadError MultipleErrors m, MonadWriter MultipleErrors m)
   => Expr
   -> m Expr
 infer' v@(Literal (NumericLiteral (Left _))) = return $ TypedValue True v tyInt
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -315,10 +316,25 @@ infer' (Literal (ArrayLiteral vals)) = do
   return $ TypedValue True (Literal (ArrayLiteral ts')) (TypeApp tyArray els)
 infer' (Literal (ObjectLiteral ps)) = do
   ensureNoDuplicateProperties ps
-  ts <- traverse (infer . snd) ps
-  let fields = zipWith (\name (TypedValue _ _ t) -> (Label name, t)) (map fst ps) ts
-      ty = TypeApp tyRecord $ rowFromList (fields, REmpty)
-  return $ TypedValue True (Literal (ObjectLiteral (zip (map fst ps) ts))) ty
+  -- We make a special case for Vars in record labels, since there are the
```
michaelficarra: I think you meant "these".

paf31: Thanks!

paf31:
@LiamGoodacre Would you mind reviewing this one, please?

LiamGoodacre:
Will do, I'll have a look through this tonight :)

LiamGoodacre:
What would you expect to happen in this code:
```purescript
type Z t = forall x. t x -> (forall a. t a) -> t x
class C t where c :: Z t
instance cA :: C Array where c x _ = x
test2 :: forall m. Monad m => m { ccc :: Z Array }
test2 = pure { ccc: c }
```

paf31:
I'm not sure, but I could try it. Why do you ask about that snippet specifically?

LiamGoodacre:
Sorry, I forgot to add an explanation.  I tried added the above code to your test file.  The idea was to also test a polymorphic var with a constraint and also a quantified argument.  I probably shouldn't have lumped them together, but anyway: the test failed with an escaped skolem.  I would have expected it to succeed, so I was wondering if you thought the same.

paf31:
Constrained functions in record literals are still a bit of a pain. Does it work if you annotate the property?

LiamGoodacre:
With an annotation it still fails in the same way:
```purescript
test2 :: forall m. Monad m => m { ccc :: Z Array }
test2 = pure { ccc: (c :: Z Array) }
```

```
  1) Passing examples '1110.purs' should compile and run without error
       Error found:
       in module Main
       at /Users/liamgoodacre/git/purescript/examples/passing/1110.purs line 25, column 1 - line 25, column 51

         A type variable has escaped its scope.

       in the expression { ccc: c cA
                         }
       in value declaration test2
```

paf31:
Oh, then that's a bug. Thanks.

paf31:
I think it's easily fixed, thankfully. I'll add a test for this.

paf31:
@LiamGoodacre How's this?

paf31:
The whole package set continues to compile with this change, so it seems safe enough.

LiamGoodacre:
This looks good :) - although I'm not an expert on this part of the code.

Issue #764, opened 2014-12-11
Pattern matching in let and where clauses


jutaro:
This is what I miss most currently as bloody beginner (beside more precise parser and type checker error messages). Don't know if it is in the issue list so far but can't find it. e.g:

```
let (X a) = X b
in a
```

currently is not working, right? As far as I see the same applies to all other patterns.
I can write something like this instead, but it can make the code less clear (not in this example):

```
case X b of
  X a -> a
```

nrolland:
wouldn't a version where the ordering has to be linearly specified, no topo sort, be useful ?  

I think in many of the cases I use it with only one case or parallel destructors anyway. 
having to introduce extraneous binders is strange after being too spoiled. I guess it makes the generated code closer to input code but in most case i think that's moot

user471:
:+1:

AndrasKovacs:
:+1:

libscott:
+1

jdegoes:
I'd definitely use this.

felixSchl:
So would I, this would be great to have. :+1:

vstastny:
I would also vote for having this.

zhangchiqing:
Any update on this feature? 

I just ran into this:
```purescript
  inst <- liftEff now
  let (Milliseconds n) = unInstant inst
```
psc 0.10.1 will throw this error
```
  Unable to parse module:
  unexpected (
  expecting local declaration
```

paf31:
No update, but it's on the Approved list, so if someone wants to work on it, we can give advice on the design/implementation.

noraesae:
I would like to work on this. @paf31, can you give me hints on how to implement this? Maybe with desugaring into case expression as mentioned above?

paf31:
@noraesae Great! This will definitely be implemented as a desugaring pass, but the details need to be worked out:

- What changes we need in the parser, and how we can minimize backtracking
- How this fits into the desugaring process, taking into account the need to look at binding groups

noraesae:
Thanks for your comment. This is my first contribution on PureScript and can take a while to be accustomed to its code base.

As you suggested, I will start from looking into the parser, and see what kind of change is needed. I will update the process as frequently as possible, as I may not be sure if I am doing it right :smile:

noraesae:
Working on the parser. I have created an example too.

- Add an example: https://github.com/noraesae/purescript/commit/cdec7991d96d91e40ef7e257a98048c9a9014553
[...]

PR #2627, opened 2017-02-06
Desugar patterns in let-in expressions

noraesae:
The original issue: #764 

Let me know if there are something I have done wrong. I will catch up and update ASAP 😄


examples/passing/LetPattern.purs
```diff
@@ -0,0 +1,128 @@
...
+
+import Prelude
+import Partial.Unsafe (unsafePartial)
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (CONSOLE, log)
+import Test.Assert (ASSERT, assert')
+
+patternSimple :: Boolean
+patternSimple =
+  let (x) = 25252
```
paf31: Are the parens significant here?

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -564,6 +571,15 @@ parseBinderNoParens = P.choice
                       , ParensInBinder <$> parens parseBinder
                       ] P.<?> "binder"
 
+parseBinderLocalDecl :: TokenParser Binder
```
paf31: Why not use the existing `parseBinderAtom` parser for binders?

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -231,6 +237,7 @@ parseLocalDeclaration :: TokenParser Declaration
 parseLocalDeclaration = positioned (P.choice
                    [ parseTypeDeclaration
                    , parseValueDeclaration
+                   , parseBoundValueDeclaration
```
paf31: I think it would be better to always try to parse many binders here, and then turn the single `VarBinder` case into a `ValueDeclaration`. That way we can avoid backtracking in the parser.

src/Language/PureScript/Sugar/LetPattern.hs
```diff
@@ -0,0 +1,48 @@
...
+desugarLetPattern decl =
+  let (f, _, _) = everywhereOnValuesM return replace return
+  in f decl
+  where
+  replace :: Expr -> m Expr
+  replace (Let ds e) = go ds [] e
+  replace other = return other
+
+  go :: [Declaration] -- ^ Declarations to desugar
+        -> [Declaration] -- ^ Declarations already desugared, in reverse order
```
paf31: Nitpick: can you please line up the `->` and `::`s and move the comments underneath?

src/Language/PureScript/Sugar/LetPattern.hs
```diff
@@ -0,0 +1,48 @@
...
+import Control.Monad.Supply.Class
+
+import Language.PureScript.AST
+import Language.PureScript.Errors
+
+-- |
+-- Replace every @BoundValueDeclaration@ in @Let@ expressions with @Case@
+-- expressions.
+--
+desugarLetPatternModule :: forall m. (MonadSupply m, MonadError MultipleErrors m) => Module -> m Module
```
paf31: You don't seem to be using these constraints.

examples/passing/LetPattern.purs
```diff
@@ -0,0 +1,128 @@
...
+  let [a, b] = [1, 2]
+  in
+   a == 1 && b == 2
+
+patternDoArray :: forall e. Eff e Boolean
+patternDoArray = unsafePartial do
+  let [a, b] = [1, 2]
+  pure $ a == 1 && b == 2
+
+patternMultiple :: Boolean
```
paf31: Do the declarations have to be in dependency order? I'm guessing so, since this pass happens before the binding groups pass.

paf31:
Thanks!

examples/passing/LetPattern.purs
noraesae: Not significant at all. I just thought it's worth checking if it can match `(expr)` well. Would I remove this?

paf31: I think it's fine, it was just a bit confusing to see it. I assumed it was significant somehow.

src/Language/PureScript/Parser/Declarations.hs
noraesae: It's because `parseBinderAtom` includes literal binds and null binder as well. I thought it's better not allowing them in parsing and use separate sets of binders.

src/Language/PureScript/Parser/Declarations.hs
noraesae: I am not sure I understand well. Can you explain more?

src/Language/PureScript/Sugar/LetPattern.hs
noraesae: Oops, will fix this!

src/Language/PureScript/Sugar/LetPattern.hs
noraesae: Will remove `MonadSupply`. `MonadError` may be needed with `rethrowWithPosition`?

examples/passing/LetPattern.purs
noraesae: I have just tested this and as you said they have to be in dependency order. Should they work in non-dependency order? In the case, what could be a fix for it?

noraesae:
Thanks for the comments. I applied reviews and leave some comments.

src/Language/PureScript/Parser/Declarations.hs
paf31: Why disallow them?

src/Language/PureScript/Parser/Declarations.hs
paf31: Well, `parseValueDeclaration` and `parseBoundValueDeclaration` overlap, so the parser has to backtrack and we can't guarantee good error messages. But we could make them share a parser, and process the result into the right sort of declaration. If we parse using `many1 parseBinderAtom`, then lists where the first element is a `VarBinder` represent `ValueDeclaration`s and singleton lists represent `BoundValueDeclaration`s. Everything else is an error, but now we don't need to backtrack.

examples/passing/LetPattern.purs
paf31: Well, we would need to run the binding groups pass before the `let` pass. It would be quite a bit more work. There are also issues related to code gen, as described on the original issue, so maybe we should leave that for later.

Any thoughts, anyone else?

src/Language/PureScript/Sugar/LetPattern.hs
paf31: No, you still don't need `MonadError`. `rethrowWithPosition` only applies when you might throw something, which I don't think you do.

src/Language/PureScript/Parser/Declarations.hs
noraesae:
```purescript
let
  _ = foo
  123 = bar
in
  ...
```

The code above doesn't look proper to me, because it doesn't add any value declaration. Would we just allow them too?

paf31: Would you consider this to be valid?

```purescript
case foo of
  _ -> ...
```

I admit it's unusual, but it might be useful (using `purescript-debug` for example), and I don't think it's worth fragmenting the parser for.

src/Language/PureScript/Parser/Declarations.hs
noraesae: Thanks for your explanation, now I understand. Will work on it.

1. Remove `parseBoundValueDeclaration`
2. Edit `parseValueDeclaration` like following:  
    ```haskell
    binders <- many1 parseBinderAtom
    case binders of
      VarBinder Ident : binders -> ... -- return ValueDeclaration
      [binder] -> ... -- return BoundValueDeclaration
      _ -> ... -- error
    ```

By the way, how should I throw error from parse functions? Thinking about `throwError`, but cannot find a case it's used with `TokenParser a`.

examples/passing/LetPattern.purs
noraesae: Agree with leaving it for later.

src/Language/PureScript/Parser/Declarations.hs
paf31: You can use Parsec to throw an error. Maybe use `unexpected`.

src/Language/PureScript/Parser/Declarations.hs
noraesae: Right. I will change it to use `parseBinderAtom`. Thanks!

src/Language/PureScript/Parser/Declarations.hs
noraesae: maybe we need to use `parseBinderNoParens`? `parseBinderAtom` will mis-parse the case where there are constructors other than nullary one.

```hs
f X a b = ...
```

noraesae:
Modified the parser as discussed.

src/Language/PureScript/Parser/Declarations.hs
paf31: Yes, you're right.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -390,7 +390,7 @@ data Declaration
   --
   | ValueDeclaration Ident NameKind [Binder] (Either [(Guard, Expr)] Expr)
   -- |
-  -- A binded value declaration (binder, optional guard, value)
+  -- A bound value declaration (binder, optional guard, value)
```
paf31: Could you please expand this comment a bit? I'm not sure what it means for a declaration to be "bound", but maybe we can just say it's a declaration paired with a binder in a `let` clause.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -63,33 +63,30 @@ parseTypeSynonymDeclaration =
...
+  case binders of
+    VarBinder name : binders' -> do
+      value <- Left <$> (indented *>
+                          P.many1 ((,) <$> parseGuard
+                                        <*> (indented *> equals *> parseValueWithWhereClause)
+                                  ))
+          <|> Right <$> (indented *> equals *> parseValueWithWhereClause)
+      return $ ValueDeclaration name Public binders' value
+    [binder] -> BoundValueDeclaration binder <$> (indented *> equals *> parseValueWithWhereClause)
+    _ -> P.unexpected "unexpected binders in value declaration"
```
paf31: Parsec will print out "unexpected unexpected ..." here. Can you try it and see what the error looks like?

paf31:
Thanks! Could you please also add some failing test cases  for the various failure modes?

- Different types of parse errors
- Declarations which are out of order, or mutual dependencies
- etc.

src/Language/PureScript/Parser/Declarations.hs
noraesae: You are right.

```
  Unable to parse module:
  unexpected unexpected binders in value declaration
  expecting binder
```

src/Language/PureScript/Parser/Declarations.hs
noraesae: Oh, sorry I mistook. The first pattern should be able to be n-ary, because the following pattern is also correct.

```purescript
let X a = X 10
in
```

noraesae:
Modified according to the comments, added some failing examples.

src/Language/PureScript/Sugar/LetPattern.hs
```diff
@@ -0,0 +1,50 @@
...
+  go [] [] e = return e
+  go (PositionedDeclaration pos com (BoundValueDeclaration binder boundE) : ds) ds' e =
+    rethrowWithPosition pos $ go [] ds' . PositionedValue pos com =<< desugared
+    where
+    desugared :: m Expr
+    desugared = do
+      resultE <- go ds [] e
+      return $ Case [boundE] [CaseAlternative [binder] (Right resultE)]
+  go (d:ds) ds' e = go ds (d:ds') e
+  go [] ds' e = return $ Let (reverse ds') e
```
paf31: Why use `reverse` here instead of just consing the new element onto the front of the result in the recursive case?

src/Language/PureScript/Sugar/LetPattern.hs
```diff
@@ -0,0 +1,50 @@
...
+-- Replace every @BoundValueDeclaration@ in @Let@ expressions with @Case@
+-- expressions.
+--
+desugarLetPatternModule :: forall m. (MonadError MultipleErrors m) => Module -> m Module
+desugarLetPatternModule (Module ss coms mn ds exts) = Module ss coms mn <$> parU ds desugarLetPattern <*> pure exts
+
+-- |
+-- Desugar a single let expression
+--
+desugarLetPattern :: forall m. (MonadError MultipleErrors m) => Declaration -> m Declaration
```
paf31: `MonadError` is still not needed.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -65,14 +65,17 @@ parseTypeSynonymDeclaration =
...
 parseValueDeclaration = do
-  name <- parseIdent
-  binders <- P.many parseBinderNoParens
-  value <- Left <$> (indented *>
-                       P.many1 ((,) <$> parseGuard
-                                    <*> (indented *> equals *> parseValueWithWhereClause)
-                               ))
-       <|> Right <$> (indented *> equals *> parseValueWithWhereClause)
-  return $ ValueDeclaration name Public binders value
+  binders <- (:) <$> parseBinderAtom <*> (P.many parseBinderNoParens)
```
paf31: Should this be `parseBinder`? I'm not sure.

paf31:
👍 Thanks for the great work on this!

src/Language/PureScript/Parser/Declarations.hs
noraesae: What's the difference between `parseBinder` and `parseBinderAtom`? I have just tried with `parseBinder` and it breaks many modules from being parsed.

src/Language/PureScript/Sugar/LetPattern.hs
paf31: I know why you need reverse if you write this as a left fold, but I think it's simpler as a right fold.

src/Language/PureScript/Sugar/LetPattern.hs
paf31: `rethrowWithPosition` needs the constraint, but you don't need `rethrowWithPosition`, since nothing will be thrown 😄

src/Language/PureScript/Parser/Declarations.hs
paf31: `parseBinder` allows operators and type annotations in binders. I'm curious - what broke? It certainly would be nice to support something like `let x : xs = list in ...`

src/Language/PureScript/Sugar/LetPattern.hs
noraesae: Oh, I see. `Monad m =>` may be enough then?

paf31: You don't need a monad. You can use a regular function.

src/Language/PureScript/Parser/Declarations.hs
noraesae: When using `parseBinder` instead of `paresBinder`, modules in the test directory seem like failing to build. I copied the error in the following gist:

https://gist.github.com/noraesae/f9b2b0e1dfc8514d468b09a14bc149a7

Do you have any idea?

noraesae:
I have removed the `MonadError` constraint, and made `go` to right fold.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -65,14 +65,17 @@ parseTypeSynonymDeclaration =
...
+  case binders of
+    VarBinder name : binders' -> do
+      value <- Left <$> (indented *>
+                          P.many1 ((,) <$> parseGuard
+                                        <*> (indented *> equals *> parseValueWithWhereClause)
+                                  ))
+          <|> Right <$> (indented *> equals *> parseValueWithWhereClause)
+      return $ ValueDeclaration name Public binders' value
+    [binder] -> BoundValueDeclaration binder <$> (indented *> equals *> parseValueWithWhereClause)
```
paf31: What happens now if a `BoundValueDeclaration` is parsed at the top level?

noraesae: It silently parses without error. How about separate `parseValueDeclaration` and `parseLocalValueDeclaration`? For the top level, we can just parse ident at the beginning of the decl, and for the local one, we can parse `parseBinder` as you commented above. I will check if it's possible.

src/Language/PureScript/Parser/Declarations.hs
noraesae: It's because of not handling `PositionedBinder`. I will fix this.

src/Language/PureScript/Parser/Declarations.hs
paf31: Could you please add that as a failing test case too?

noraesae: Sure. Thanks!

noraesae:
I updated the PR to use `parseBinder` instead of `parseBinderAtom` and handle the failing top-level case.

paf31:
Great! Could you please add a passing example using the `(:)` constructor? Then I'll get this merged in for 0.11.

Thanks again!

noraesae:
I added test suites to test infix operator pattern, resolved conflict by merging master, and fix build caused by the latest update on master.

noraesae:
Thanks for your comments and directions too 😃

Issue #2336, opened 2016-09-29
Single consolidated executable


paf31:
This is something I'd like to do before 1.0:

``` text
> psc build 'src/**/*.purs'
> psc bundle 'output/*/*.js'
> psc docs
> psc publish
> psc repl
> psc ide
```

etc.

It should also reduce the size of the binary distribution a bit.

kRITZCREEK:
I was hoping we could move the executables into their own package. With stack it seems easy to keep them in the same repo and manage them together. The benefit I'm hoping for would be to reduce build times when working on PS "the library".

paf31:
> The benefit I'm hoping for would be to reduce build times when working on PS "the library".

I think reducing down to one executable would also help there though.

rvion:
> I was hoping we could move the executables into their own package.

This is a bit less pleasant to work with than having only one executable in the same package IMO
I personally really prefer having one large executable, with a clean and good module hierarchy so build times are acceptable

kRITZCREEK:
> I think reducing down to one executable would also help there though.

And I'm totally :+1: there. Would also be a nice opportunity to combine psc-ide-server and client into `psc ide` and `psc ide --client`.

hdgarrood:
Should we do this in 0.11?

paf31:
I think that would be good. I'll make a milestone so we can make a list of things to concentrate on.

We should come up with a migration plan though. We can't have all existing blog posts etc. immediately being invalidated 😄 It could be as simple as providing small wrapper executables for now.

Issue #2517, opened 2016-12-31
Remove dependency on Bower from the compiler


paf31:
The compile depends on Bower in the following places:

- PSCi
- Test suite
- `psc-publish`

We should be able to sever the tie to Bower by providing generic support for other package managers in the form of command line arguments.

paf31:
@hdgarrood I started looking at this, but it was a lot more work than I expected it would be 😄 

To begin with, can we come up with a list of fields we would need to provide to `psc-publish`? So far I have these:

- Name
- Description
- License
- Author(s)
- Repository URL
- Dependencies with version ranges

Are there any others that you can think of?

hdgarrood:
Yes, I think that would cover it. I was imagining that we might provide this data as JSON on stdin and `psc-publish` could ignore any extraneous fields - that way it should be easy to just pipe bower.json into `psc-publish` / `purs publish` for now and also allow easy switching later.

rvion:
> it should be easy to just pipe bower.json into psc-publish

I would prefer a regular invocation style (windows compat, clearer in scripts, etc.)

hdgarrood:
Windows does support specifying standard input to be read from a file on the command line, though. I'd prefer to use stdin because then if any preprocessing is needed there's no need to write an intermediate file (which could dirty your git working tree). The kind of preprocessing I'm imagining is that e.g. some other package manager might have a manifest file that's not quite compatible with the structure `purs publish` is expecting.

rvion:
> Windows does support specifying standard input to be read from a file on the command line, though

> then if any preprocessing is needed there's no need to write an intermediate file

if any preprocessing is needed, I think windows would need an intermediate file anyway. 
I'm not a windows shell expert though.

http://stackoverflow.com/questions/14574170/how-do-i-use-a-pipe-to-redirect-the-output-of-one-command-to-the-input-of-anothe

-----------

What I wanted to say is that regardless if it is possible to pipe bower.json as stdin to psc-publish, I would prefer to also be able to write

if there is a standard package(s?) definition file:

```
psc publish
```

if there is not:

```
psc publish --from-config=bower.json
```

or :

```
psc publish
 --name foo
 --description="lib to print 'foo' strings"
 --license=bsd3
 --authors=rvion
 --repository=git@github.com:rvion/foo.git
 --deps="..."
```

If you want other reasons why I prefer to avoid piping: 

  - it is harder to document (via available commands and options)
[...]

PR #2632, opened 2017-02-07
Remove psc-package, combine remaining executables

### LINKED ISSUE #2604: Move psc-package to another repo, support package management more generically
While we can continue to work on psc-package, I think we should not aim to support it as part of 1.0. Instead, we should try to remove any dependencies on any particular package managers (Bower or psc-package) from the compiler, and instead provide a generic set of hooks and command line arguments to support different package managers.

paf31:
Fixes #2604, #2618 and #2336.

I created a separate repo for `psc-package` at https://github.com/purescript/psc-package/.


hdgarrood:
I think this only partially addresses #2604 - we also need to remove hardcoded bower paths from psc-publish and maybe other compiler tools. But perhaps it would be better to make a separate issue for that.

paf31:
@hdgarrood There is #2517 to track the work to remove Bower. I'll merge all of these together now. Thanks.

paf31:
To do:

- [ ] Remove reference to psc-docs
- [ ] Fix comment typo
- [ ] Update bundle script

paf31:
Do we want to continue to call the single executable `psc`? We might want to rename it and have `psc` be a wrapper script which calls the `compile` command.

hdgarrood:
`purs` sounds good to me, especially since a lot of what it does isn't really compiling. Also just to check, are we intending to remove these wrapper scripts at the next major release?

app/Command/IDE.hs
```diff
@@ -19,7 +19,7 @@
 {-# LANGUAGE TemplateHaskell       #-}
 {-# LANGUAGE NoImplicitPrelude     #-}
 
-module Main where
+module Command.IDE (command) where
```
kRITZCREEK: Would prefer `Ide` like in the rest of the codebase, just for consistencies sake.

paf31: Sounds good

hdgarrood:
You've checked the wrapper scripts item in your checklist but I can't find any wrapper scripts in this PR's diff, am I missing something?

paf31:
Sorry, forgot to push my changes.

hdgarrood:
Looks great :) perhaps we should consider wrapper scripts for Windows too? I think we want e.g. `psc.cmd` with just a single line, `purs compile %*` &mdash; apparently the Windows analogue for `$@` is `%*`.

hdgarrood:
If anyone has a Windows environment around to check that that works, that would be handy.

paf31:
Shall I merge this then? I guess if we want to make a 0.10.7 then we can branch from the tag, yes?

hdgarrood:
Yep, sounds good to me.

paf31:
Thanks everyone.

PR #2633, opened 2017-02-07
Single consolidated executable

### LINKED ISSUE #2336: Single consolidated executable

paf31:
Fixes #2336. This also removes `psc-ide-client`, which had been discussed previously.


kRITZCREEK:
Looking good! I've pr'd #2634 against this branch to fix the test failures. We said we wanted to provide wrappers around the consolidated executable to smoothen the breakage or do we just go full burning bridges with 0.11?

src/Language/PureScript/Options.hs
```diff
@@ -1,42 +1,20 @@
...
+  { optionsVerboseErrors :: Bool
+  -- ^ Verbose error message
   , optionsNoComments :: Bool
-    -- |
-    -- Generate soure maps
+  -- ^ Remove the comments from the generated js
   , optionsSourceMaps :: Bool
-    -- |
-    -- Dump CoreFn
+  -- ^ Generate soure maps
```
hdgarrood: Typo?

hdgarrood:
Perhaps we could provide shell scripts named after the old executables for backwards compatibility in the binary bundles, just for the 0.11 series? They'd only need to be a few lines I think.

For true backwards compat we would also need to have `psc` work like the new `psc build`, I guess that could be triggered if the first argument doesn't match any of the sub-commands.

app/Command/Docs.hs
```diff
@@ -218,20 +215,15 @@ buildOptions (fmt, input, mapping) =
...
-  version = abortOption (InfoMsg (showVersion Paths.version)) $ long "version" <> help "Show the version number" <> hidden
+command :: Opts.Parser (IO ())
+command = (buildOptions >=> docgen) <$> (version <*> Opts.helper <*> pscDocsOptions) where
+  version :: Opts.Parser (a -> a)
+  version = Opts.abortOption (Opts.InfoMsg (showVersion Paths.version)) $ Opts.long "version" <> Opts.help "Show the version number" <> Opts.hidden
+
+infoModList :: Opts.InfoMod a
+infoModList = Opts.fullDesc <> footerInfo where
+  footerInfo = Opts.footerDoc $ Just $ PP.vcat
+                 [ examples, PP.empty, PP.text ("psc-docs " ++ showVersion Paths.version) ]
```
hdgarrood: Oh wait, should we remove this reference to `psc-docs`?

app/Command/Compile.hs
```diff
@@ -89,94 +84,66 @@ globWarningOnMisses warn = concatMapM globWithWarning
...
-  where
-  opts        = info (version <*> helper <*> pscMakeOptions) infoModList
-  infoModList = fullDesc <> headerInfo <> footerInfo
-  headerInfo  = header   "psc - Compiles PureScript to Javascript"
-  footerInfo  = footer $ "psc " ++ showVersion Paths.version
-
-  version :: Parser (a -> a)
-  version = abortOption (InfoMsg (showVersion Paths.version)) $ long "version" <> help "Show the version number" <> hidden
+command :: Opts.Parser (IO ())
+command = compile <$> (version <*> Opts.helper <*> pscMakeOptions) where
```
hdgarrood: I'm not sure the sub commands need to be able to respond to `--version`. Having `psc --version` work is enough IMO, we don't also need to support `psc build --version` etc.

src/Language/PureScript/Options.hs
paf31: Yes, thanks.

app/Command/Compile.hs
paf31: True :)

app/Command/Docs.hs
paf31: Yes, looks like it.

paf31:
I'll fix the remaining bits on the other branch, after merging this. I also need to update the bundle script.

kRITZCREEK:
I just checked and it seems like `psc-ide-vim` still relies on `psc-ide-client`. Maybe it would be best to keep it around behind `purs ide --client`.

paf31:
@kRITZCREEK Sounds good, we can add it back.

Issue #2631, opened 2017-02-06
Should error when a type class member doesn't reference at least a covering set of arguments


LiamGoodacre:
Consider the following class, for example:

```purescript
class C a b | a -> b where
  m :: a
  n :: b
```

Given the functional dependency here, `m` is perfectly fine because a type for `a` must be known for us to select an instance, but `n` couldn't possibly be used.

LiamGoodacre:
Actually, I think this make sense to check on all constraints not just type class members? ...maybe

It's also not as trivial as I first thought, an argument could only appear in a type via a constraint - so whether or not the member could be used may depend on the constraints functional dependencies.

But perhaps just checking the free variables in each member's type is good enough for now.

paf31:
> Actually, I think this make sense to check on all constraints not just type class members?

Yes, but I think classes would be a good place to start.

PR #2637, opened 2017-02-08
Add basic usability check for type class members

LiamGoodacre:
For https://github.com/purescript/purescript/issues/2631

### DIFF
--- a/examples/failing/UnusableTypeClassMethodSynonym.purs
+++ b/examples/failing/UnusableTypeClassMethodSynonym.purs
@@ -1,0 +1,9 @@
+-- @shouldFailWith UnusableDeclaration
+module Main where
+
+type M x = forall a. a
+
+class C a where
+  -- after synonym expansion, the type doesn't actually contain an `a`
+  c :: M a
+
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -171,6 +171,7 @@ errorCode em = case unwrapErrorMessage em of
   CannotUseBindWithDo{} -> "CannotUseBindWithDo"
   ClassInstanceArityMismatch{} -> "ClassInstanceArityMismatch"
   UserDefinedWarning{} -> "UserDefinedWarning"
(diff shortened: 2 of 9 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -886,6 +887,11 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs) e = flip evalS
             , indent msg
             ]
 
+    renderSimpleErrorMessage (UnusableDeclaration ident) =
+      paras [ line $ "The declaration " <> markCode (showIdent ident) <> " is considered to be unusable."
```
paf31: Nitpick: Is "considered to be" necessary?

examples/failing/UnusableTypeClassMethodSynonym.purs
```diff
@@ -0,0 +1,9 @@
+-- @shouldFailWith UnusableDeclaration
+module Main where
+
+type M x = forall a. a
+
+class C a where
+  -- after synonym expansion, the type doesn't actually contain an `a`
+  c :: M a
```
paf31: There is a related problem: a type variable can be shadowed:

```purescript
c :: forall a. a -> a
```

has the same problem.

paf31:
Thanks for working on this! Let's figure out a plan for dealing with shadowing before merging though. I think there might be an existing traversal we can use.

examples/failing/UnusableTypeClassMethodSynonym.purs
LiamGoodacre: Have I misunderstood what `freeTypeVariables` does?  I was expecting to not get back any explicitly bound variables, such as those in introduced in a forall.

LiamGoodacre: In any case, I'll add a test case with a bound variable with the same name as a class argument.

paf31: Oh, you're right, that should do the job.

PR #2641, opened 2017-02-08
Fix builtin links

hdgarrood:
Unfortunately, links to Prim declarations are still broken on Pursuit; I guess this bug sneaked in while I was transferring this code to this repo. I've fixed it again (and verified that it fixes the issue in Pursuit) and also added a test.

### DIFF
--- a/src/Language/PureScript/Docs/Types.hs
+++ b/src/Language/PureScript/Docs/Types.hs
@@ -397,8 +397,9 @@ data LinkLocation
   | BuiltinModule P.ModuleName
   deriving (Show, Eq, Ord)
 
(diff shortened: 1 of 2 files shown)


hdgarrood:
Argh, one of these days I'm going to remember to wait for CI to finish first.

hdgarrood:
Thanks! I was planning to point directly to this commit in `master` from Pursuit's stack.yaml so that we don't need to release a 0.10.7, does that sound okay to you?

garyb:
We probably need a .7 soon anyway: #2642

hdgarrood:
Argh, that's my fault too, sorry!

Issue #2310, opened 2016-09-17
Improve skolem escape check


paf31:
Extracted from #1719

PR #2643, opened 2017-02-09
Improved skolem escape check

### LINKED ISSUE #2567: Disappearing constraints
I've noticed that if you specify a constraint but don't use it, then it never gets checked:
```purescript
-- no error
class M s
foo :: forall t. t -> Int
foo = ((\_ -> 0) :: M t => t -> Int)
```

```purescript
-- no error
foo :: Int
foo = (0 :: Fail "Bananas" => Int)
```

```purescript
-- fails
foo :: Int
foo = (0 :: Fail "Bananas" => Int) + 1
```

paf31:
Fixes #2567 and #2310.

Here is an example of the new error message:

```text
Error found:
in module Main
at examples/failing/SkolemEscape2.purs line 8, column 10 - line 10, column 8

  The type variable h, bound at

    examples/failing/SkolemEscape2.purs line 9, column 14 - line 9, column 26

  has escaped its scope, appearing in the type
                                              
    (STRef h0 Int -> Eff t5 Int) -> Eff t5 Int
                                              
in the expression (bind bindEff) (runST (newSTRef 0))
in value declaration test
```


hdgarrood:
Nice!

paf31:
Does this look ok to merge?

garyb:
I think so... I didn't see anything wrong in the code, but my review was pretty superficial. The new errors look awesome! Does the package set / Halogen, etc. still compile?

paf31:
I tested Halogen but not the whole package set. Let me see ...

garyb:
If Halogen passes it's probably fine, it's probably one of the harder tests due to the great deal of existentials :smile:. I guess maybe lensy things would be the only other thing potentially tripped up by it?

paf31:
Well Thermite is okay too, which uses lenses quite a bit. But I'll test the package set anyway.

paf31:
@garyb Everything builds, except for `Data.StrMap`, but that's because it actually contains some escaped skolems 😄 

The error is this:

```
Error found:
in module Data.StrMap
at /private/tmp/x/.psc-package/psc-0.10.6/maps/v2.1.1/src/Data/StrMap.purs line 169, column 14 - line 169, column 42

  The type variable h, bound at

    /private/tmp/x/.psc-package/psc-0.10.6/maps/v2.1.1/src/Data/StrMap.purs line 169, column 21 - line 169, column 42

  has escaped its scope, appearing in the type
                             
    forall h e.              
      STStrMap h a24         
      -> Eff                 
           ( st :: ST h      
           | e               
           )                 
           (STStrMap h26 a24)
                             

in the expression \s ->           
                    ((poke s) k) v
in value declaration insert
```

and you can fix it by inserting a few `void`s in `insert` and similar functions. Since this won't go out until 0.11, I think that's probably okay since we can make a new release of `maps`. Right?

garyb:
Yeah, I think we have a bunch of breaking changes in the core libs queued up, so makes sense to do that with 0.11 :)

garyb:
Oh, if it doesn't need a breaking change, but just a fix before this goes out - even better.

PR #2645, opened 2017-02-09
Compress contributors list

paf31:
I thought it would be good to compress the list a bit, to make it easier to see the actual contributor names in the list.

I used a regular expression to extract contributors which used the standard text, so that I could find the ones which did not. I added those to a separate list at the end.


LiamGoodacre:
Most of the `MIT License` references are links, but a handful of them aren't.

paf31:
@LiamGoodacre Yes, I noticed, but I wanted to maintain the content of the original list without modifications.

Issue #2635, opened 2017-02-08
Travis CI cache doesn't work with new stack.yml


noraesae:
I think the cache doesn't work now in Travis Ci and the build takes much longer (~40mins) than before (~6mins).

I think it's because of #2611. I will look into this.

noraesae:
Maybe it's not caching `extra-deps`? It seems like packages listed in the `extra-deps` are rebuilt everytime.

noraesae:
Also, to clear the previous cache and rebuild one with the new stack yaml, we need to change (or remove) the `compiler` field, I guess.

@hdgarrood, may you have done the work in the [extend-travis](https://github.com/purescript/purescript/tree/extend-travis) branch?

hdgarrood:
I'm looking at Travis CI and caching at the moment, yeah. I'm currently trying to find a way of addressing the issue that builds often time out when switching to new snapshots or compiler versions. Once I've sorted that out I'll look at this.

hdgarrood:
I think I've found a way to re-enable OSX builds: https://gist.github.com/hdgarrood/29fa8ebcd5b5acd679cf429fd0a93e3e

I'm fairly sure that `extra-deps` get stored in `.stack-work` local to the project directory, but I don't think we should include that directory in the build cache unless it's really necessary, as I think there might be some risk of accidentally persisting state that we don't want to be persisted in CI jobs that way.

noraesae:
Yes. I also think `extra-deps` are not really a big deal in reducing the build time, #2636 will already removed most of them though. The biggest problem was that the new stack env had not been properly cached, because it was using the same `compiler` name and thus the same cache(of the previous resolver). I think it's fixed by removing `compiler` fields, as you have done in the extend-travis branch.

noraesae:
#2636 is now merged. Travis CI build also takes about 10 mins in the extend-travis branch (with macOS build enabled too!), so I think this issue can be closed after the branch is merged into master.

PR #2647, opened 2017-02-09
Address travis timeout issues

hdgarrood:
These commits amend the CI scripts so that if a build does time out, the next one can continue where the previous left off. For more info see https://gist.github.com/hdgarrood/29fa8ebcd5b5acd679cf429fd0a93e3e

This enables us to enable OSX builds again (which I have done).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,11 @@ cache:
   directories:
   - $HOME/.local/bin
   - $HOME/.stack
+  # Maximum amount of time in seconds spent attempting to upload a new cache
+  # before aborting. Since our cache can get rather large, increasing this
+  # value helps avoid situations where caches fail to be stored. The default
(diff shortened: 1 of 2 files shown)


hdgarrood:
Fixes #2635

paf31:
👍 Looks great!

hdgarrood:
Thanks! I think the best way forward now is for me to do the following:

* delete all the Travis caches,
* merge this and wait for CI to finish
* if it fails (which it probably will because there's no cache), push one more empty commit to master to get it to build again

Then, the caches should be set up, and also it should speed up future CI builds as there won't be as much useless data from older versions of the compiler in the caches which has to be downloaded and re-uploaded for every job. If that sounds good I'll go ahead and do that.

paf31:
Sounds good

PR #2649, opened 2017-02-10
Capitalise *script into *Script

noraesae:
This PR changes Javascript and Purescript into JavaScript and PureScript, which are official names as I understand.

I also would like 'Javascript' in header descriptions in the following pages to be changed. (In [the PureScript homepage](http://www.purescript.org/), it's correctly written in JavaScript)

- https://github.com/purescript
- https://github.com/purescript/purescript


paf31:
Thanks!

Issue #2573, opened 2017-01-15
Suggest pulp when running psci without arguments


LiamGoodacre:
There's been talk about this in IRC many times and I couldn't find an issue for it.

Currently if you run `psci` without arguments it'll show:
```
PSCi requires the psci-support package to be installed.
You can install it using Bower as follows:

  bower i purescript-psci-support --save-dev

Or using psc-package:

  psc-package install psci-support

For help getting started, visit https://github.com/purescript/documentation/blob/master/PSCi.md
```

It does mention `pulp psci` on the linked page, but I think we should display it here too.

Something along the lines of:

```
> psci
No input given to PSCi; try running `pulp psci` instead.
For help getting started, visit https://github.com/purescript/documentation/blob/master/PSCi.md
```

hdgarrood:
:+1: :+1: :+1:

nwolverson:
Should this not at least indicate the proper usage of psci (i.e. `psci [input glob]`) as well as suggesting pulp?

noraesae:
How about defaulting `[input glob]` part to `psci 'src/**/*.purs' 'bower_components/purescript-*/src/**/*.purs'`, which may be the most common use case? I sometimes just want to check something very quickly wich `psci`, like how I usually do with `ghci`.

The usage will be like below:

### with neither pulp nor psci-support

```
> psci 
With Pulp, try running `pulp psci` instead.

Without Pulp, PSCi requires psci-support to be installed.
You can install it using Bower as follows:

  bower i purescript-psci-support --save-dev

Or using psc-package:

  psc-package install psci-support

For help getting started, visit https://github.com/purescript/documentation/blob/master/guides/PSCi.md
```

### with psci-support

```
> psci
```

... just defaults to ...

```
> psci 'src/**/*.purs' 'bower_components/purescript-*/src/**/*.purs'
```

noraesae:
Also, the documentation URL should be modified to https://github.com/purescript/documentation/blob/master/guides/PSCi.md.

hdgarrood:
It's a good suggestion, but I'd rather not do that, because we're trying to remove any hardcoded assumptions of particular package managers from the compiler; see #2604.

noraesae:
I can understand. Then this issue may be obsolete?

hdgarrood:
No, I think suggesting Pulp is still a good idea (Pulp isn't a package manager).

noraesae:
Oh, right. I will work on this in a while.

[...]

PR #2650, opened 2017-02-10
Update CLI error and helper messages

noraesae:
This is concerning recent changes in executables(#2632, #2633), and also #2573.

### DIFF
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -55,7 +55,9 @@ compile :: PSCMakeOptions -> IO ()
 compile PSCMakeOptions{..} = do
   input <- globWarningOnMisses (unless pscmJSONErrors . warnFileTypeNotFound) pscmInput
   when (null input && not pscmJSONErrors) $ do
-    hPutStrLn stderr "psc: No input files."
+    hPutStrLn stderr $ unlines [ "purs compile: No input files."
+                               , "Usage: For basic information, try the `--help' option."
+                               ]
     exitFailure
   moduleFiles <- readInput input
   (makeErrors, makeWarnings) <- runMake pscmOpts $ do
@@ -68,7 +70,7 @@ compile PSCMakeOptions{..} = do
   exitSuccess
 
 warnFileTypeNotFound :: String -> IO ()
-warnFileTypeNotFound = hPutStrLn stderr . ("psc: No files found using pattern: " ++)
+warnFileTypeNotFound = hPutStrLn stderr . ("purs compile: No files found using pattern: " ++)
 
 globWarningOnMisses :: (String -> IO ()) -> [FilePath] -> IO [FilePath]
 globWarningOnMisses warn = concatMapM globWithWarning
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -22,7 +22,7 @@ import qualified Options.Applicative as Opts
 import qualified System.Console.ANSI as ANSI
 import           System.Exit (exitSuccess, exitFailure)
 import           System.FilePath.Glob (glob)
-import           System.IO (hPutStrLn, stderr)
+import           System.IO (hPutStr, hPutStrLn, stderr)
 import           System.IO.UTF8 (readUTF8FileT)
 
 data PSCMakeOptions = PSCMakeOptions
@@ -55,9 +55,9 @@ compile :: PSCMakeOptions -> IO ()
 compile PSCMakeOptions{..} = do
   input <- globWarningOnMisses (unless pscmJSONErrors . warnFileTypeNotFound) pscmInput
   when (null input && not pscmJSONErrors) $ do
-    hPutStrLn stderr $ unlines [ "purs compile: No input files."
-                               , "Usage: For basic information, try the `--help' option."
-                               ]
+    hPutStr stderr $ unlines [ "purs compile: No input files."
+                             , "Usage: For basic information, try the `--help' option."
+                             ]
     exitFailure
   moduleFiles <- readInput input
   (makeErrors, makeWarnings) <- runMake pscmOpts $ do
--- a/src/Language/PureScript/Interactive/Message.hs
+++ b/src/Language/PureScript/Interactive/Message.hs
@@ -41,9 +41,16 @@ prologueMessage = unlines
   , "Type :? for help"
   ]
 
+noInputMessage :: String
+noInputMessage = unlines
+  [ "purs repl: No input files; try running `pulp psci` instead."
+  , "For help getting started, visit " ++ guideURL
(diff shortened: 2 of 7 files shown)


src/Language/PureScript/Interactive/Message.hs
```diff
@@ -28,27 +32,20 @@ helpMessage = "The following commands are available:\n\n    " ++
...
-  , "You can install it using Bower as follows:"
-  , ""
-  , "  bower i purescript-psci-support --save-dev"
-  , ""
-  , "Or using psc-package:"
-  , ""
-  , "  psc-package install psci-support"
-  , ""
-  , "For help getting started, visit https://github.com/purescript/documentation/blob/master/PSCi.md"
+  [ "purs repl: No input files, or no psci-support loaded; try running `pulp psci` instead."
```
hdgarrood: It's straightforward enough to distinguish these two cases that I think we should be doing so. I'd rather leave the `supportModuleMessage` more or less as it is and define a new `noInputMessage` similar to what you have here, but which is only displayed when no input files are provided.

The place to do that seems to be here: https://github.com/noraesae/purescript/blob/005a84f031e975224a31564309296134bbe1a4ad/app/Command/REPL.hs#L313-L315

noraesae: Sounds appropriate. Will upload a follow-up commit to fix it.

noraesae:
I updated the PR with a patch to distinguish `purs repl` failure cases, as @hdgarrood suggested.

src/Language/PureScript/Interactive/Message.hs
```diff
@@ -28,27 +32,27 @@ helpMessage = "The following commands are available:\n\n    " ++
...
   extraHelp =
     "Further information is available on the PureScript documentation repository:\n" ++
-    " --> https://github.com/purescript/documentation/blob/master/PSCi.md"
+    " --> " ++ guideURL
 
 -- | The welcome prologue.
 prologueMessage :: String
 prologueMessage = unlines
-  [ "PSCi, version " ++ showVersion Paths.version
+  [ "PureScript REPL, version " ++ showVersion Paths.version
```
paf31: I think it's fine to continue to call the app "PSCi", but when referring to the command, it should say `purs repl`.

src/Language/PureScript/Interactive/Message.hs
```diff
@@ -28,27 +32,27 @@ helpMessage = "The following commands are available:\n\n    " ++
...
   , "Type :? for help"
   ]
 
+noInputMessage :: String
+noInputMessage = unlines
+  [ "purs repl: No input files; try running `pulp psci` instead."
```
paf31: Great!

src/Language/PureScript/Interactive/Message.hs
noraesae: Will revert this part :)

src/Language/PureScript/Interactive/Message.hs
paf31: @hdgarrood Looks good now?

src/Language/PureScript/Interactive/Message.hs
```diff
@@ -37,18 +41,18 @@ prologueMessage = unlines
...
-  , "You can install it using Bower as follows:"
-  , ""
-  , "  bower i purescript-psci-support --save-dev"
-  , ""
-  , "Or using psc-package:"
-  , ""
-  , "  psc-package install psci-support"
-  , ""
-  , "For help getting started, visit https://github.com/purescript/documentation/blob/master/PSCi.md"
+  [ "purs repl: psci-support is not loaded; try running `pulp psci` instead."
```
hdgarrood: Now that the "no input files" case is separately handled, I think the current message is actually more appropriate. I'm specifically thinking about a case where someone *has* invoked `purs repl` via `pulp psci` but `psci-support` is not installed.

paf31: Ah good point.

noraesae: I think removing the `try running ...` part for `supportModuleMessage` could be enough, because for both `pulp psci` and `purs repl` cases, what's needed is to install `psci-support`.

hdgarrood: Right, although I don't think "psci-support is not loaded" is informative enough that people new to PureScript will be able to understand what they need to do; this is why I prefer the current message.

hdgarrood: To clarify, I realise we are trying to remove hardcoded assumptions of particular package managers in the compiler, but for me that doesn't need to extend as far as changing this message; it really is more directed at e.g. `psc-publish` which currently assumes that you are using Bower as your package manager and simply does not work otherwise. The aim of this is to ease a transition away from Bower, at least that's how I see it. So I don't think we really gain anything by removing this information from this error message.

app/Command/Compile.hs
```diff
@@ -55,7 +55,9 @@ compile :: PSCMakeOptions -> IO ()
 compile PSCMakeOptions{..} = do
   input <- globWarningOnMisses (unless pscmJSONErrors . warnFileTypeNotFound) pscmInput
   when (null input && not pscmJSONErrors) $ do
-    hPutStrLn stderr "psc: No input files."
+    hPutStrLn stderr $ unlines [ "purs compile: No input files."
```
hdgarrood: I think this will insert an extra line break; if we're using `unlines`, then I think the `hPutStrLn` should change to `hPutStr`

noraesae: You are right. I will fix.

src/Language/PureScript/Interactive/Message.hs
noraesae: I understand and agree. How about:

```
purs repl: PSCi requires the psci-support package to be installed.
You can install it using psc-package or Bower as follows:

  psc-package install psci-support
  bower i purescript-psci-support --save-dev

For help getting started, visit https://github.com/purescript/documentation/blob/master/guides/PSCi.md
```

paf31: Let's just remove the reference to `psc-package` while we're at it. It's not going to be a supported option any more.

noraesae: Then, like below?

```
purs repl: PSCi requires the psci-support package to be installed.
You can install it using Bower as follows:

  bower i purescript-psci-support --save-dev

For help getting started, visit https://github.com/purescript/documentation/blob/master/guides/PSCi.md
```

paf31: 👍

noraesae:
Removed extra newlines and updated the error messages to be more informative.

PR #2652, opened 2017-02-11
[WIP] Annotations for ASTs

garyb:
This started off as just adding an annotation for `Type`, but since it was also going to affect everywhere that `Expr` / `Declaration` is used I figured I may as well add one for those too, and get it all out of the way at once, since it's super time consuming.

I did consider sharing the same annotation type for types and exprs, so there would be only one type var for `Expr`, etc. but I can imagine cases where it might be handy to have differing annotation values.

Still got a way to go I think, it's a bit hard to say as fixing one set of errors brings up a fresh batch.

### DIFF
--- a/src/Language/PureScript/AST/Binders.hs
+++ b/src/Language/PureScript/AST/Binders.hs
@@ -14,33 +14,33 @@ import Language.PureScript.Types
 -- |
 -- Data type for binders
 --
-data Binder
+data Binder a b
   -- |
   -- Wildcard binder
   --
-  = NullBinder
+  = NullBinder b
   -- |
   -- A binder which matches a literal
   --
-  | LiteralBinder (Literal Binder)
+  | LiteralBinder (Literal (Binder a b)) b
   -- |
   -- A binder which binds an identifier
   --
-  | VarBinder Ident
+  | VarBinder Ident b
   -- |
   -- A binder which matches a data constructor
   --
-  | ConstructorBinder (Qualified (ProperName 'ConstructorName)) [Binder]
+  | ConstructorBinder (Qualified (ProperName 'ConstructorName)) [Binder a b] b
   -- |
   -- A operator alias binder. During the rebracketing phase of desugaring,
   -- this data constructor will be removed.
   --
-  | OpBinder (Qualified (OpName 'ValueOpName))
+  | OpBinder (Qualified (OpName 'ValueOpName)) b
   -- |
   -- Binary operator application. During the rebracketing phase of desugaring,
   -- this data constructor will be removed.
   --
-  | BinaryNoParensBinder Binder Binder Binder
+  | BinaryNoParensBinder (Binder a b) (Binder a b) (Binder a b) b
   -- |
   -- Explicit parentheses. During the rebracketing phase of desugaring, this
   -- data constructor will be removed.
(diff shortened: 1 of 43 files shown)


hdgarrood:
Might it be good to explain in a comment somewhere why some types have two different type parameters for annotations, but others have just one, and what each parameter means? I guess it might make sense to wait until this is a bit closer to being done to do that though.

garyb:
Yeah, makes sense, I'll do that!

src/Language/PureScript/AST/Binders.hs
```diff
@@ -14,78 +14,90 @@ import Language.PureScript.Types
 -- |
 -- Data type for binders
 --
-data Binder
+data Binder a b
```
paf31: May I suggest using the names `type` and `binder` for the type variables?

garyb: You may, but `type` is an illegal name, and `b` is for value annotations in general 😜 

Point taken though... I'll need to find something to help me do the refactoring. It took me ~5 hours to get this far even!

src/Language/PureScript/Types.hs
```diff
@@ -110,63 +224,65 @@ $(A.deriveJSON A.defaultOptions ''ConstraintData)
...
-                               in ForAll v' (go (v' : bs) m t') sco
-                             | otherwise = ForAll v (go (v : bs) m t) sco
+  go :: [Text] -> [(Text, Type a)] -> Type a -> Type a
+  go _  m (TypeVar v a) = fromMaybe (TypeVar v a) (v `lookup` m)
+  go bs m (TypeApp t1 t2 a) = TypeApp (go bs m t1) (go bs m t2) a
+  go bs m f@(ForAll v t sco a)
+    | v `elem` keys = go bs (filter ((/= v) . fst) m) f
+    | v `elem` usedVars =
+        let v' = genName v (keys ++ bs ++ usedVars)
+            t' = go bs [(v, TypeVar v' a)] t
```
paf31: I got stuck here when I tried this (I didn't get nearly as far). The annotation on `TypeVar v' a` seems wrong, since if the annotations include kinds (for example), then this would have the wrong kind. I think we'd need to take the annotation from the type variable we're replacing.

Probably not a big deal for now though.

src/Language/PureScript/AST/Binders.hs
```diff
@@ -12,80 +12,97 @@ import Language.PureScript.Comments
...
   --
-  | PositionedBinder SourceSpan [Comment] Binder
+  | PositionedBinder SourceSpan [Comment] (Binder a b) b
   -- |
   -- A binder with a type annotation
   --
-  | TypedBinder Type Binder
-  deriving (Show, Eq)
+  | TypedBinder (Type a) (Binder a b) b
+  deriving (Show, Functor)
```
paf31: It's probably useful to derive `(Bi)Traversable` for these too.

paf31:
🎉 Amazing, thanks for starting this!

src/Language/PureScript/AST/Binders.hs
paf31: I don't think this should hold you up, we can always change it later.

src/Language/PureScript/AST/Binders.hs
garyb: For sure, I just need to look up whatever TH magic is needed for it, since I don't think GHC can do it natively unfortunately.

paf31: Maybe you can do it without TH if you derive `Traversable X` and `Traversable (Flip X)` 😄

kRITZCREEK:
Just because I just came across it, there is no constructor for a `PositionedCaseAlternative` yet. A full case alternative is neither a value nor a type, but it's still a syntactic entity we might want to have position information for.

Would this need an additional type parameter `c` for purely syntactic constructs in the design of this PR?

garyb:
@paf31 Ok, I've started a new branch on this stuff, but just wanted to discuss something for a moment, to check you're ok with it.

Instead of introducing multiple annotation types, I've thought of something that will be less crazy for the actual AST types, and gives us a much greater amount of flexibility in the annotations we can use. 

When I started on this today I added annotations to kinds too, since they currently have poor errors, and position info would surely help. That led me to `Expr ka ta va`, which is already pretty nasty, and then I thought... well, we probably want to annotate `DeclarationRef`s too, as they have a `Positioned` constructor, but not with `va` because they're not really things at the value level, etc. but that way lies madness, as we'll have as many type variables as we have varieties in annotation. After a bunch of messing around, I came up with this kind of thing:

``` haskell
data Type a
  = SomeType (a 'TypeAnn)
  -- | ...

data Expr a
  = SomeExpr (a 'ExprAnn)
  | TypedExpr (a 'ExprAnn) (Type a) (Expr a)
  -- | ...

data Declaration a
  = SomeDecl (a 'DeclAnn) (Expr a)
  -- | ...
```

Unfortunately, it doesn't work if we want to derive instances. I fought with `StandaloneDeriving` and some other stuff for a while to try and convince it to work without resorting to `UndecidableInstances` without much success. The idea seemed good though: parameterise the annotations so we can use one type to represent them but then have some way of indexing varieties of annotation, so we don't have to resort to partiality in whatever form when extracting the annotations.

After various attempts I finally ended up with this:

``` haskell
data AnnType
  = KindAnn
  | TypeAnn
  | ExprAnn
  | DeclAnn
  | RefAnn
  -- etc., could also have `BinderAnn`, `CaseAlternativeAnn` or as many others as might be useful

data Ann a (i :: k) where
  Ann :: Annotation a i x => a i -> Ann a i

class (Ord x, Show x) => Annotation a (i :: k) x | a i -> x where
  getAnn :: a i -> x

instance Eq (Ann a i) where
  (Ann a1) == (Ann a2) = f a1 == f a2

instance Ord (Ann a i) where
  compare (Ann a1) (Ann a2) = compare (f a1) (f a2)

instance Show (Ann a i) where
  show (Ann a) = show (f a)

data Type a
  = SomeType (Ann a 'TypeAnn)
  -- | ...
  deriving (Eq, Ord, Show)

data Expr a
  = SomeExpr (Ann a 'ExprAnn)
  | TypedExpr (Ann a 'ExprAnn) (Type a)
  -- | ...
  deriving (Eq, Ord, Show)

data Declaration a
  = SomeDecl (Ann a 'DeclAnn) (Expr a)
  -- | ...
  deriving (Eq, Ord, Show)
```

So essentially we have a GADT wrapping up a dictionary for the `Annotation` class that lets us do the projection safely based on the annotation type and index, and can preserve deriving for the stuff that we want. Having the `Eq, Ord, Show` instances defined for `Ann` is what allows us to recover deriving.

For now we'll probably only have one annotation:

```
newtype PosAnnotation i = PosAnn SourceSpan

instance Annotation PosAnnotation a SourceSpan where
  getAnn (PosAnn b) = b
```

This depends on `FlexibleInstances` as we have the `a` in there, but it could be avoided by just duplicating the instance with all the promoted `AnnType` constructors substituted into `a`.

This scheme will extend nicely and allow us to specialise the annotations as much as we might ever want, by adding new `AnnType` indices, and then just updating the instances for `Annotation` accordingly.

Are you ok with me going ahead with doing it this way? And/or can you see any ways of simplifying it? I went through a very tortuous process to get to here, so might be missing something obvious due to going through all that.

garyb:
From the Slack discussion, here's an option that would keep `Generic` deriving, etc:

``` haskell
-- Ann' would be hidden, only the type and `Ann` pattern synonyms exported:

data Ann a (i :: k)
  = Ann'
      (a i -> a i -> Bool)
      (a i -> a i -> Ordering)
      (a i -> String)
      (a i)
  deriving Generic

pattern Ann :: (Ord (a i), Show (a i)) => a i -> Ann a i
pattern Ann x <- Ann' _ _ _ x
  where
  Ann x = Ann' (==) compare show x

instance Eq (Ann a i) where
  (Ann' eq _ _ a1) == (Ann' _ _ _ a2) = eq a1 a2

instance Ord (Ann a i) where
  compare (Ann' _ comp _ a1) (Ann' _ _ _ a2) = comp a1 a2

instance Show (Ann a i) where
  show (Ann' _ _ sh a) = sh a
```

It's a little gross, but with the construction enforced by pattern / smart constructor it means there will be no funny business when using Eq, Ord, etc.

paf31:
I thought about this a bit more, and I think there is quite a lot of value in keeping the main AST as simple as possible. That is the data type which the most newcomers interact with, and it was one of the original goals of the project to only use the simple parts of Haskell (and I think one of the reasons the project has been successful). 

So here's another option to consider. We could have one type argument which is common, and which would contain things like source spans and comments, and then add additional type arguments on an as-needed basis, for things like type synonyms, if we decide that they are important enough to require special support. The common type argument can be the last one, which means that `Functor`, `Traversable` etc. will all be available to solve the most common problems.

In any case, I think we should make a list of things we might want to use arguments for in future. If the list is relatively short, and many things are common across all AST types, I think we should consider something simple and incremental. I'll start:

- Source spans and comments (common)
- Type synonym information (on types, before 1.0 hopefully)
- Inferred types and data from the elaborator (on terms, before 1.0 hopefully)
- Information relevant to polymorphic kinds (on types, after 1.0)

garyb:
Another case we'd want to get in sooner rather than later: the pre-desugared names of things.

garyb:
I was talking to @kRITZCREEK about this yesterday, and thought of another thing - we could just add SourceSpan/comment info to all the constructors without sticking it in an annotation field. That doesn't solve the other problems, but it will help a great deal with our ability to have decent error positions, and has some other benefits since we can more reasonably invent/copy sourcespan info when desugaring code etc. in a way that might be less sensible for generic annotations. 

Saves introducing a type variable everywhere for a case that will never change too.

PR #2653, opened 2017-02-11
Use Discard class to catch values which are implicitly discarded

### LINKED ISSUE #1803: Add warning for absence of `_ <-` when throwing away non-Unit values in a `do`
It can help prevent some rather annoying bugs, like if you do something like this:

``` purescript
do
  maybe (pure unit) ($ (Left act)) <$> readRef emitRef
  pure next
```

When you meant to do this:

``` purescript
do
  maybe (pure unit) ($ (Left act)) =<< readRef emitRef
  pure next
```

(The former case has a return like `m (m Unit)` rather than `m Unit`, but because I didn't notice that the action was not sequenced and therefore nothing happened)

paf31:
Fixes #1803

PR for `prelude` incoming.

Example error:

```
Error 1 of 2:

  in module Control.Monad.Rec.Class
  at /private/tmp/x/.psc-package/psc-0.10.6/tailrec/v2.0.1/src/Control/Monad/Rec/Class.purs line 114, column 1 - line 130, column 42

    A result of type
                
      Step a0 b1
                
    was discarded in a do notation block.
    You can use _ <- ... to explicitly discard the result.
    Alternatively, add a Discard constraint to the type of the discarded value.

  while checking that expression \f ->               
                                   \a ->             
                                     let             
                                       fromDone = ...
                                       f' = ...      
                                     in runST (...)  
    has type forall a b eff. (a -> Eff eff (Step a b)) -> a -> Eff eff b
  while applying a function runST
    of type forall a r.        
              (forall h.       
                 Eff           
                   ( st :: ST h
                   | r         
                   )           
                   a           
              )                
              -> Eff r a       
    to argument (bind (f' a)) (\$3 ->      
                                 case $3 of
                                   e -> ...
                              )            
  in value declaration tailRecEff

  where b1 is a rigid type variable
          bound at line 114, column 1 - line 130, column 42
        a0 is a rigid type variable
          bound at line 114, column 1 - line 130, column 42
```


src/Language/PureScript/Errors.hs
```diff
@@ -612,13 +612,20 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs) e = flip evalS
...
                   (map (paras . map (line . markCode)) (transpose bs))
                   : [line "..." | not b]
             , line "Alternatively, add a Partial constraint to the type of the enclosing value."
             ]
+    renderSimpleErrorMessage (NoInstanceFound (Constraint C.Discard [ty] _)) =
+      paras [ line "A result of type"
+            , markCodeBox $ indent $ typeAsBox ty
+            , line "was implicitly discarded in a do notation block."
+            , line ("You can use " <> markCode "_ <- ..." <> " to explicitly discard the result.")
+            , line "Alternatively, add a Discard constraint to the type of the discarded value."
```
garyb: Perhaps we shouldn't say this? It seems like an odd thing to advertise in the types since it's an implementation detail.

garyb:
I wonder if we should add something to inline `discardUnit` applications too?

src/Language/PureScript/Errors.hs
paf31: I was thinking maybe we'd need to add a `Discard` constraint if the type were polymorphic, but actually that seems very unlikely.

paf31:
Inlining `discardUnit` does sound useful.

rightfold:
Should MagicDo be updated for this?

paf31:
I don't think so, we just need a new inliner rule in the regular inliner.

paf31:
Ok, this should be be ready for review again.

@rightfold Sorry, I see what you mean. Yes, we probably should inline `discard` like we inline `bind`.

garyb:
I meant to comment earlier, everything looks good to me now, just the tests are failing. I guess the `bower.json` will fix that :smile:

paf31:
Yes, and once I make releases, I'll turn those back into `^` ranges.

PR #2654, opened 2017-02-11
Emit _ instead of false case for if-then-else

rightfold:
This makes for better code generation.


paf31:
Thanks!

> This makes for better code generation.

I believe you, but can you give an example? 😄

rightfold:
```purescript
x a = if a then 1 else 2
```

before:
```javascript
var x = function (a) {
    if (a) {
        return 1;
    };
    if (!a) {
        return 2;
    };
    throw new Error("Failed pattern match at Main line 5, column 7 - line 5, column 24: " + [ a.constructor.name ]);
};
```

after:
```javascript
var x = function (a) {
    if (a) {
        return 1;
    };
    return 2;
};
```

paf31:
I thought we used `? ... : ...` for `if then else`. Did that change?

paf31:
Looks good, thanks!

rightfold:
`?:` operator might have been pre-CoreFn?

paf31:
Ah right.

PR #2656, opened 2017-02-11
Support polymorphic types in typed binders

### LINKED ISSUE #2626: Support polymorphic types in typed binders
See also #2445

paf31:
Fixes #2626

Thanks to @LiamGoodacre for realizing that this was pretty much working already 😄

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -531,25 +531,23 @@ parseIdentifierAndBinder =
 -- | Parse a binder
 parseBinder :: TokenParser Binder
 parseBinder =
-  withSourceSpan
-    PositionedBinder
-    ( P.buildExpressionParser operators
-    . buildPostfixParser postfixTable
-    $ parseBinderAtom
-    )
+    withSourceSpan
+      PositionedBinder
+      ( P.buildExpressionParser operators
+      . buildPostfixParser postfixTable
+      $ parseBinderAtom
+      )
   where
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -545,7 +545,7 @@ parseBinder =
     ]
 
   -- TODO: parsePolyType when adding support for polymorphic types
```
kRITZCREEK: can this TODO be removed now? ;)

paf31: Ah, it wasn't as simple as I thought ...

paf31:
I tried this, which I expected to fail:

```purescript
> class Default a where def :: a

> let f = \(x :: (forall a. Default a => a) -> Int) -> x def

> :type f
((forall a. (Default a) => a) -> Int) -> Int

> g \_ -> 0
0
```

I'd like to understand why this case succeeds before merging, since I would expect this to fail because we try to unify constrained types somewhere.

rightfold:
Does this support e.g.

    data T a = A | B a

    f :: T Int -> T String
    f x@(A _ :: forall a. T a) = x
    f (B n) = B (show n)

LiamGoodacre:
@rightfold no that's not supported

paf31:
I've convinced myself using `trace` that this is unifying the right things. Unifying constrained types might result in errors in some cases, but for simple things like this, it should be fine.

Issue #2657, opened 2017-02-12
Type annotations in CoreFn?


pkamenarsky:
Right now CoreFn is untyped (as far as I can see).

I'd like to experiment with a Haskell backend for Purescript, but it seems that without type annotations type inference in the generated Haskell code might suffer (though I'm not sure about this).

What would it take to add type annotations in CoreFn?

garyb:
Are you talking about the dumped JSON, or `CoreFn` itself? Type annotations are carried through the `a` annotation parameter within the compiler, via the [`Ann` type](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CoreFn/Ann.hs) , but I guess maybe aren't in the JSON currently.

rightfold:
Last time I checked they were often absent.

paf31:
> I'd like to experiment with a Haskell backend for Purescript

You might like to look at https://github.com/tfausak/thran.

rintcius:
Not sure if it would be beyond what to expect of `CoreFn` but I would be interested to have a structure that wouldn't loose any information compared to the source code (e.g. it would be possible to exactly reconstruct the sources from the corefn JSON)

DanielRrr:
Phil! There's some moments which I can't understand. What's exactly needed for modifying the compiler for getting type info? I need to use initial AST without desugaring and rewrite CoreFn.hs for that or I need to do it someway differently?

paf31:
The CoreFn AST has the information, it's just not in the JSON. There is an open PR to add it, which might get merged at some point, but for now, you'd need to use the compiler as a library, or fork the compiler.

PR #2658, opened 2017-02-12
Typed CoreFn WIP

pkamenarsky:
First attempt at dumping type annotations from CoreFn.

Addresses #2657.


paf31:
One possible problem is that we can end up with internal details like skolem constants leaking into the JSON representation. The core can probably be typed, but we don't have any documentation for what the types are, or how they correspond to the types the user wrote. 

Thank you for looking at this and making the PR, but I would like to think about that problem first, which is why I put the issue under "Ideas".

pkamenarsky:
Yeah, I thought so (i.e. WIP), I was just curious how a naive dump would look like. Feel free to close if you feel the PR is premature.

paf31:
Thank you for the PR, but I think #3049 is more up to date at this point, so I'm going to close this. Apologies for letting this bitrot, especially since we might end up going a similar route.

Issue #2609, opened 2017-01-31
Compiler cannot compile exported data constructor that has a quote in its name


chrisdotcode:
```purescript
module Test (Foo(Bar, Bar')) where

data Foo = Bar | Bar'
```
fails to compile with:
```bash
  Unable to parse module:
  unexpected "Bar'"
  expecting proper name
```
But:

```purescript
module Test (Foo(..)) where

data Foo = Bar | Bar'
```
compiles successfully.

chrisdotcode:
This seems (potentially) easy enough that I might like to attempt to resolve it. If somebody could point me in the right direction, that would be great :)

LiamGoodacre:
@chrisdotcode :)
For this I think you will need to change `parseDeclarationRef` in `src/Language/PureScript/Parser/Declarations.hs` to use the same parser for constructors that `parseDataDeclaration` does.
This should fix it for both exports and imports.

Or perhaps `properName` should parse single-quotes? @paf31 

We should also add a test case for this in `examples/passing/`.  There are some multi-file tests you could copy for inspiration, but you will also need to list the folder you add to `purescript.cabal` with all the others.

Thanks!

chrisdotcode:
Speaking of that, I should mention that quotes aren't allowed in foreign import names (e.g.: `foreign import insert'`), either. If it's using the same code, and we update it, that would be A Bad Thing(TM) because we know JavaScript function names  can't have quotes in them.

Just something to be aware of.

rvion:
could we orgnanize a poll about primes ?
I have the impression that most people are against those nowdays

chrisdotcode:
(I personally find them highly convenient and use them frequently for slight variations of functions; I wasn't necessarily aware of any stigma, either).

hdgarrood:
@rvion that's not my impression, it's been suggested a couple of times before and people haven't been keen. I don't have the issue numbers to hand but there have been at least two discussions on this issue tracker.

hdgarrood:
Primes are definitely allowed in foreign imports, Data.Array imports an FFI function called `uncons'` (but doesn't export it).

chrisdotcode:
My mistake. Corrected above.

PR #2660, opened 2017-02-13
Use data ctor name parser for data ctors in imports/exports

LiamGoodacre:
For #2609

### DIFF
--- a/examples/passing/2609.purs
+++ b/examples/passing/2609.purs
@@ -1,12 +1,12 @@
 module Main where
 
 import Prelude
-import Eg (Foo'(Bar'))
+import Eg (Foo'(Bar'), (:->))
 import Control.Monad.Eff (Eff)
 import Control.Monad.Eff.Console (CONSOLE, log)
 
 bar' :: Foo'
-bar' = Bar'
+bar' = 4 :-> 5
 
 main :: forall e. Eff (console :: CONSOLE | e) Unit
-main = case bar' of Bar' -> log "Done"
+main = case bar' of Bar' l r -> log "Done"
--- a/examples/passing/2609.purs
+++ b/examples/passing/2609.purs
@@ -1,0 +1,12 @@
+module Main where
+
+import Prelude
+import Eg (Foo'(Bar'))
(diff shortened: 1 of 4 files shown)


hdgarrood:
Oh wait hold on, it seems a bit odd to me that  `properName` won't successfully parse things with primes, even though primes are allowed in proper names (aren't they)? Is `properName` used elsewhere?

LiamGoodacre:
For declarations, `properName` appears to be used:
* in class names (declaration, import, export, constraints, instances)
* fixity declarations

garyb:
It's also used for the parts of module names, class names, etc. But actually, `ModuleName` is probably the only one we want to exclude primes from, so we could just make that a special case instead.

garyb:
Ah, or maybe not, if that ^ is all it's used for.

LiamGoodacre:
The value fixity declarations are for data constructors, right?  So I should probably change that too.

garyb:
Yep! 👍

hdgarrood:
Why does it not make sense to allow primes in `properName` and then special-case for module names?

garyb:
That's what I was suggesting above. :smile:

hdgarrood:
Oh, right, cool. Your message following that one made me think you had realised there was a problem with that approach.

garyb:
Ah, I was pointing to Liam's message that came in just a few seconds before; he didn't mention module names in the places it was used so I was saying that might invalidate my suggestion.

paf31:
Is this good to merge then?

hdgarrood:
I think this is okay as is but ideally I think we would do a bit more refactoring of some of the code in the lexer to better reflect that we're allowing primes in more places. For example I think the `validUName` function probably ought to change a bit.

paf31:
Well feel free to merge when you think it's ready, obviously.

hdgarrood:
@LiamGoodacre if it's not clear what I mean let's just go ahead and merge this, we can do the refactoring I'm imagining later.

paf31:
Let's merge this and do the tidying you describe in a separate PR then. Thanks @LiamGoodacre!

PR #2662, opened 2017-02-14
More Position info and Traversals

kRITZCREEK:
This is a byproduct of me playing around with the AST in: https://github.com/kRITZCREEK/morgana

I just wanted to confirm with you that we could make these kind of changes without breaking anything.

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -79,10 +79,10 @@ parseValueWithWhereClause = do
 parseValueWithIdentAndBinders :: Ident -> [Binder] -> TokenParser Declaration
 parseValueWithIdentAndBinders ident bs = do
   value <- indented *> (
-    (\v -> [MkUnguarded v]) <$> (equals *> parseValueWithWhereClause) <|>
+    (\v -> [MkUnguarded v]) <$> (equals *> withSourceSpan PositionedValue parseValueWithWhereClause) <|>
       P.many1 (GuardedExpr <$> parseGuard
                            <*> (indented *> equals
-                                         *> parseValueWithWhereClause))
+                                         *> withSourceSpan PositionedValue parseValueWithWhereClause))
     )
   return $ ValueDeclaration ident Public bs value
 
@@ -97,7 +97,7 @@ parseLocalValueDeclaration = join $ go <$> parseBinder <*> (P.many parseBinderNo
   where
   go :: Binder -> [Binder] -> TokenParser Declaration
   go (VarBinder ident) bs = parseValueWithIdentAndBinders ident bs
-  go (PositionedBinder _ _ (VarBinder ident)) bs = parseValueWithIdentAndBinders ident bs
+  go (PositionedBinder _ _ b) bs = go b bs
   go binder [] = BoundValueDeclaration binder <$> (indented *> equals *> parseValueWithWhereClause)
   go _ _ = P.unexpected $ "patterns in local value declaration"
 
@@ -348,7 +348,7 @@ parseIdentifierAndValue =
 parseAbs :: TokenParser Expr
 parseAbs = do
   symbol' "\\"
-  args <- P.many1 (indented *> (Abs <$> (Left <$> parseIdent <|> Right <$> parseBinderNoParens)))
+  args <- P.many1 (indented *> (Abs <$> parseBinderNoParens))
   indented *> rarrow
   value <- parseValue
   return $ toFunction args value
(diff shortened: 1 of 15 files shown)


src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -531,25 +533,27 @@ parseIdentifierAndBinder =
...
     PositionedBinder
     ( P.buildExpressionParser operators
     . buildPostfixParser postfixTable
     $ parseBinderAtom
     )
   where
-  operators =
-    [ [ P.Infix (P.try (indented *> parseOpBinder P.<?> "binder operator") >>= \op ->
-          return (BinaryNoParensBinder op)) P.AssocRight
+    flattenPositionedBinder (PositionedBinder p c (PositionedBinder p' _ b)) | p == p' = PositionedBinder p c b
```
paf31: It seems like this shouldn't be necessary.

paf31:
This seems fine to me, but we would need to go through all of the traversals and recursive functions on these types to make sure they don't need to be updated. I suspect a few would. We might be better off waiting for @garyb's work to add annotations, and then we can move position info into the annotations.

The traversal changes look necessary though, so good catch on that :)

kRITZCREEK:
Okay, I'll remove the changes to the parser soon then.

We're still missing a couple of traversal cases for `BoundValueDeclaration`, but I can't figure all of them out. Maybe we could do them together? I'd like to understand what's going on in some of these Traversals :D

paf31:
Sure, ping me on IRC and we'll find a time.

kRITZCREEK:
The last commit changed the AST representation of `Abs` to always use a Binder instead of `Either Ident Binder`. The `Left Ident` is now represented as a `VarBinder`.

I'm sure at this point this PR has become a mess to merge after #2652 but if this seems good we could maybe try to incorporate this change into that PR. I checked and `pscid` still compiles fine with these changes.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -636,11 +637,12 @@ check' (Literal (ArrayLiteral vals)) t@(TypeApp a ty) = do
   unifyTypes a tyArray
   array <- Literal . ArrayLiteral <$> forM vals (`check` ty)
   return $ TypedValue True array t
-check' (Abs (Left arg) ret) ty@(TypeApp (TypeApp t argTy) retTy) = do
-  unifyTypes t tyFunction
-  ret' <- withBindingGroupVisible $ bindLocalVariables [(arg, argTy, Defined)] $ check ret retTy
-  return $ TypedValue True (Abs (Left arg) ret') ty
-check' (Abs (Right _) _) _ = internalError "Binder was not desugared"
+check' (Abs binder ret) ty@(TypeApp (TypeApp t argTy) retTy)
+  | (VarBinder arg) <- binder = do
```
paf31: Redundant parens here

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -348,12 +348,13 @@ infer' (Accessor prop val) = withErrorMessageHint (ErrorCheckingAccessor val pro
...
   typed <- check val (TypeApp tyRecord (RCons (Label prop) field rest))
   return $ TypedValue True (Accessor prop typed) field
-infer' (Abs (Left arg) ret) = do
-  ty <- freshType
-  withBindingGroupVisible $ bindLocalVariables [(arg, ty, Defined)] $ do
-    body@(TypedValue _ _ bodyTy) <- infer' ret
-    return $ TypedValue True (Abs (Left arg) body) $ function ty bodyTy
-infer' (Abs (Right _) _) = internalError "Binder was not desugared"
+infer' (Abs binder ret)
+  | (VarBinder arg) <- binder = do
```
paf31: Same

src/Language/PureScript/Parser/Declarations.hs
paf31: The PR seems good to me, but I'd really like to be able to remove this. Seems the parser should be able to guarantee we only add one `PositionedBinder`.

kRITZCREEK:
I've applied your suggestions and removed the changes to the parser.

paf31:
@kRITZCREEK Looks good, but I didn't want you to necessarily remove all of that code, just to hopefully find a way to remove the duplicate annotations. We can come back to this though.

kRITZCREEK:
I still have all the code so it's no big deal. I'm hoping @garyb's PR will subsume all of it anyway.

paf31:
Yes, that sounds like a good idea. Many libraries don't build right now because of the breaking changes in 0.11, but most core libraries do.

kRITZCREEK:
I've readded the position information for binder parsers and got around the special check by making a desugaring step in the parser recurse on positioned binders properly. This compiles both pscid and slamdata fine except for discard errors in slamdata, but these are to be expected.

paf31:
👍 Thanks, looks good to me!

kRITZCREEK:
Cool I'll tidy the commit history up and merge then, @garyb is that OK with you? This will probably cause lots of conflicts with your AST PR.

garyb:
Yeah, no problem. I might start that PR over and try and hammer it out in one day, since it's getting kinda stale now :wink:. Will see how bad the conflicts are first though.

PR #2665, opened 2017-02-15
PSCi command changes

### LINKED ISSUE #2630: let ... in doesn't work in psci
For example:

```
> let f _ = 0 in f 0
(line 1, column 13):
unexpected "in"
expecting ., {, _, number, char literal, string literal, "true", "false", [, "\\", qualifier, data constructor name, identifier, "case", "if", "do", "let", (, hole literal, ::, operator, "where", indentation at column 5 or end of input
```

I was hoping to just get `0` printed just as if I'd typed `0` at the prompt.

noraesae:
Resolve #2630.

This PR includes the followings:

1. Enable `let ... in` in PSCi
2. Add support for value declaration without `let` in PSCi
3. Remove special `let` from PSCi

I personally think 3 makes sense, but surely it can be reverted as discussed in the original issue.

### DIFF
--- a/src/Language/PureScript/Interactive/Parser.hs
+++ b/src/Language/PureScript/Interactive/Parser.hs
@@ -35,9 +35,8 @@ psciCommand = choice (map try parsers)
   where
   parsers =
     [ psciImport
-    , psciOtherDeclaration
+    , psciDeclaration
     , psciExpression
-    , psciLet
     ]
 
 trim :: String -> String
@@ -75,18 +74,6 @@ parseDirective cmd =
 psciExpression :: P.TokenParser Command
 psciExpression = Expression <$> P.parseValue
 
--- |
--- PSCI version of @let@.
--- This is essentially let from do-notation.
--- However, since we don't support the @Eff@ monad,
--- we actually want the normal @let@.
---
-psciLet :: P.TokenParser Command
-psciLet = Decls <$> (P.reserved "let" *> P.indented *> manyDecls)
-  where
-  manyDecls :: P.TokenParser [P.Declaration]
-  manyDecls = mark (many1 (same *> P.parseLocalDeclaration))
-
 -- | Imports must be handled separately from other declarations, so that
 -- :show import works, for example.
 psciImport :: P.TokenParser Command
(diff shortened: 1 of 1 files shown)


hdgarrood:
I think the `Decls` constructor being able to hold multiple declarations is important so that we can do things like this:

```
> import Prelude 
> :paste
… let x 0 = 0
…     x 1 = 2
…     x other = other * 2
… 
> x 0
0

> x 1
2

> x 35
70
```

Does that work with this code?

noraesae:
Oh, didn't notice about it. Will fix.

noraesae:
Now the case is covered by `psciDeclaration`.

```
> import Prelude
> :paste
… x 0 = 0
… x 1 = 2
… x other = other * 2
…
> x 0
0

> x 1
2

> x 35
70

```

paf31:
Thanks for working on this!

This looks good generally, but allowing multiple declarations simultaneously outside of a `let` just seems .. odd somehow.

noraesae:
I agree that it's not how it worked, but to me it seems natural because it's how it works in normal `.purs` files. FYI, it also works similarly in GHCi, except that GHCi also provides the special `let`.

```
λ> :{
Prelude| x 0 = 0
Prelude| x 1 = 2
Prelude| x other = other * 2
Prelude|
Prelude| :}

λ> x 10
20

λ> x 0
0

```

hdgarrood:
Haha, I just realised how bad an example this is; my `x` is just `(_ * 2)`!

paf31:
@noraesae How are the parser error messages now?

noraesae:
Maybe when using the PSCi `let`? It's like below:

```
> let a = 10
(line 1, column 9):
unexpected end of input
expecting ., {, ::, operator, "where", indentation at column 5 or "in"
>
```

I think the error message is somewhat unpleasant. Will check if I can make it better.

paf31:
Actually, I think it's quite good 😄 

What about a bad declaration?

noraesae:
#### Pattern matching

```
> X a = X 10
(line 1, column 5):
unexpected =
expecting ., {, _, number, char literal, string literal, "true", "false", [, "\\", qualifier, data constructor name, identifier, "case", "if", "do", "let", (, hole literal, ::, operator or end of input
> [x, y] = [1, 2]
(line 1, column 8):
unexpected =
expecting ., {, _, number, char literal, string literal, "true", "false", [, "\\", qualifier, data constructor name, identifier, "case", "if", "do", "let", (, hole literal, ::, operator or end of input
```

Actually I'm thinking about adding a pattern matching support for PSCi after this PR.

#### Duplicated declaration

```
> a = 20
> a
20

> a = 30
Error found:
in module $PSCI

  Multiple value declarations exist for a.


See https://github.com/purescript/documentation/blob/master/errors/DuplicateValueDeclaration.md for more information,
```

#### Circular reference

```
> c = c + c
Error found:

  The value of c is undefined here, so this reference is not allowed.


See https://github.com/purescript/documentation/blob/master/errors/CycleInDeclaration.md for more information,
```

Although recursion surely works.

```
> f x = if x == 0 then 1 else x * (f $ x - 1)
> f 10
3628800
```

If you have further bad decls to check, please let me know.

noraesae:
Thanks!

Issue #2663, opened 2017-02-15
ReferenceError: Warn is not defined


leighman:
I have a function with a `Warn "Prefer X"` constraint.

I run `pulp build` and I correctly get a warning from Main.

When I run `pulp run` I get `ReferenceError: Warn is not defined` and a js stacktrace.

If I add `var Warn` to output/Main/index.js it runs correctly.

Using 0.10.7

hdgarrood:
I can reproduce this too, with the following program:
```purescript
module Main where

import Prelude
import Control.Monad.Eff (Eff)
import Control.Monad.Eff.Console (log)

msg :: Warn "oops" => String
msg = "hi"

main = log msg
```

Here's `output/Main/index.js`:

```javascript
"use strict";
var Prelude = require("../Prelude");
var Control_Monad_Eff = require("../Control.Monad.Eff");
var Control_Monad_Eff_Console = require("../Control.Monad.Eff.Console");
var msg = function (dictWarn) {
    return "hi";
};
var main = Control_Monad_Eff_Console.log(msg(Warn));
module.exports = {
    main: main, 
    msg: msg
};
```

paf31:
The dictionary generated is

```purescript
TypeClassDictionaryConstructorApp C.Warn (Literal (ObjectLiteral []))
```

So maybe this has never worked in the way we're expecting?

LiamGoodacre:
I think I know what the problem is here.  The dictionary is fine, it's just that we aren't importing the type class constructor when the dictionary is used - we have done this with IsSymbol, so it's not difficult.  I'll make a PR shortly.

LiamGoodacre:
Or maybe not.  Hmm, I'll have a think.

LiamGoodacre:
Actually I think I'm wrong on that - although we should be doing that for the other classes we solve for, not just IsSymbol - (so that's also a bug).  I think this one is different because Warn is defined in Prim - and therefore doesn't have a real definition.

paf31:
@LiamGoodacre I think the solution is to change the evidence to an empty record literal instead of a `TypeClassDictionaryConstructorApp`. There is nothing to say that the evidence has to be in that form.

LiamGoodacre:
@paf31 yes, I've already done so and everything seems to be working :)

pitzqa:
Hello,
I faced Same issue "**E/configParser - ReferenceError: WARN is not defined**" while running my protractor.conf.js file in protractor cli.

protractor.conf.js file,

[...]

PR #2666, opened 2017-02-16
Fix Warn type class dicts and ensure solved type classes are imported

LiamGoodacre:
For #2663

### DIFF
--- a/examples/passing/2663.purs
+++ b/examples/passing/2663.purs
@@ -1,0 +1,9 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+
+foo :: forall t. Warn "Example" => t -> t
+foo x = x
+
+main = when (foo 42 == 42) $ log "Done"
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/CoreFn/Desugar.hs
```diff
@@ -221,11 +221,11 @@ findQualModules decls =
...
-  -- IsSymbol instances for literal symbols are automatically solved and the type
-  -- class dictionaries are built inline instead of having a named instance defined
-  -- and imported.  We therefore need to import the IsSymbol constructor from
-  -- Data.Symbol if it hasn't already been imported.
-  fqValues (A.TypeClassDictionaryConstructorApp C.IsSymbol _) = getQual' C.IsSymbol
+  -- Some instances are automatically solved (e.g IsSymbol, AppendSymbol, etc)
+  -- and the type class dictionaries are built inline instead of having a
+  -- named instance defined and imported.  We therefore need to import the type
+  -- class constructor from if it hasn't already been imported.
+  fqValues (A.TypeClassDictionaryConstructorApp typeClassName _) = getQual' typeClassName
```
paf31: Is this necessary for the fix to work?

LiamGoodacre: For Warn this won't change anything.  But for the other type classes that we solve for we should have been doing this.  But I can undo this for now and when I get more time tomorrow I'll make sure that there is an issue and make a new PR.

PR #2667, opened 2017-02-16
Better error message for removed PSCi let decl

noraesae:
Regarding #2665.

PSCi will suggest declaration without `let` for the obsolete `let` usage.

```
> a = 10
> let b = 10
(line 1, column 9):
unexpected end of input
expecting ., {, ::, operator, "where", indentation at column 5 or "in"
For declarations, try without 'let'
```

How to tell the obsolete case is somewhat rough, but I guess it works well enough.

### DIFF
--- a/src/Language/PureScript/Interactive/Parser.hs
+++ b/src/Language/PureScript/Interactive/Parser.hs
@@ -123,4 +123,4 @@ psciDeprecatedLet = do
   P.indented
   _ <- mark (many1 (same *> P.parseLocalDeclaration))
   notFollowedBy $ P.reserved "in"
-  fail "for declaration in PSCi, try without \"let\""
+  fail "Declarations in PSCi no longer require \"let\", as of version 0.11.0"
--- a/src/Language/PureScript/Interactive/Parser.hs
+++ b/src/Language/PureScript/Interactive/Parser.hs
@@ -122,4 +122,5 @@ psciDeprecatedLet = do
   P.reserved "let"
   P.indented
   _ <- mark (many1 (same *> P.parseLocalDeclaration))
+  notFollowedBy $ P.reserved "in"
   fail "for declaration in PSCi, try without \"let\""
--- a/src/Language/PureScript/Interactive/Parser.hs
+++ b/src/Language/PureScript/Interactive/Parser.hs
@@ -9,7 +9,6 @@ import           Prelude.Compat hiding (lex)
 
 import           Data.Bifunctor (first)
 import           Data.Char (isSpace)
-import           Data.List (isPrefixOf, isInfixOf)
 import           Data.List (intercalate)
 import qualified Data.Text as T
 import           Text.Parsec hiding ((<|>))
@@ -28,16 +27,9 @@ parseCommand cmdString =
     _ -> parseRest psciCommand cmdString
 
 parseRest :: P.TokenParser a -> String -> Either String a
-parseRest p s = first (handleError s) $ do
+parseRest p s = first show $ do
   ts <- P.lex "" (T.pack s)
   P.runTokenParser "" (p <* eof) ts
-  where
-  handleError :: String -> ParseError -> String
-  handleError str e
-    -- If 'let' seems like being used for a declaration, suggest usage without 'let'
-    | "let " `isPrefixOf` trimStart str && "\"in\"" `isInfixOf` show e =
-      show e ++ "\nFor declarations, try without 'let'"
-    | otherwise = show e
 
 psciCommand :: P.TokenParser Command
 psciCommand = choice (map try parsers)
@@ -46,6 +38,7 @@ psciCommand = choice (map try parsers)
     [ psciImport
     , psciDeclaration
     , psciExpression
+    , psciDeprecatedLet
     ]
 
 trim :: String -> String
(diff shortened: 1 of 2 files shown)


hdgarrood:
The approach seems good, but perhaps it would be more reliable to decide whether to display this extra message by inspecting the list of tokens returned by the lexer?

noraesae:
I've actually tried to inspect the tokens, but data constructors for the tokens are not exported and cannot be used to pattern-match or compare with others. Is there a better way to check the tokens?

hdgarrood:
Maybe we should consider exporting the token ADT?

noraesae:
I'm not sure if they are intentionally hidden. Unless, I'll export the tokens and inspect them.

By the way, how should we determine the deprecated use case with the tokens? We can start by checking if the first token is `let`, but checking if the error message ends with `"in"` doesn't seem so reliable.

paf31:
I think we should try to handle this in the parser using `<?>`.

noraesae:
I've changed to handle the problem in `parseLet`.

Now the error message is like below.

```
> let a = 10
(line 1, column 9):
unexpected end of input
expecting ., {, ::, operator, "where", indentation at column 5 or "in" (for declaration in PSCi, try without "let")
```

However, this makes the error comes for PSC too.

```purescript
module Temp where

a :: Int
a = let x = 10, x
```

```
❯ purs compile temp.purs
Error found:
at /Users/jp21046/Works/personal/contrib/purescript/temp.purs line 4, column 15 - line 4, column 15

  Unable to parse module:
  unexpected ,
  expecting ., {, _, number, char literal, string literal, "true", "false", [, "\\", qualifier, data constructor name,identifier, "case", "if", "do", "let", (, hole literal, ::, operator, "where", indentation at column 9 or "in" (for declaration in PSCi, try without "let")


See https://github.com/purescript/documentation/blob/master/errors/ErrorParsingModule.md for more information,
```

Nevertheless, I think the `expecting` part doesn't look so awkward.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -380,7 +380,7 @@ parseLet = do
   indented
   ds <- mark $ P.many1 (same *> parseLocalDeclaration)
   indented
-  reserved "in"
+  reserved "in" P.<?> "\"in\" (for declaration in PSCi, try without \"let\")"
```
paf31: I was hoping this could go in the PSCi parser actually. It seems like a shame to have to include it in the regular compiler and elsewhere.

noraesae:
For an alternative way to do this, I restored the original parser but with the error message at the end. I think we can keep the error message for a while and remove it later, after 1.0 is released for example.

src/Language/PureScript/Interactive/Parser.hs
```diff
@@ -112,3 +114,12 @@ parseReplQuery' str =
...
                       intercalate ", " replQueryStrings ++ ".")
     Just query -> Right query
+
+-- | To show error message when 'let' is used for declaration in PSCI,
+-- which is deprecated.
+psciDeprecatedLet :: P.TokenParser Command
+psciDeprecatedLet = do
+  P.reserved "let"
+  P.indented
+  _ <- mark (many1 (same *> P.parseLocalDeclaration))
```
paf31: You could just use `_ <- parseLocalDeclaration` here. One declaration is enough to fail.

src/Language/PureScript/Interactive/Parser.hs
```diff
@@ -112,3 +114,12 @@ parseReplQuery' str =
...
+  fail "for declaration in PSCi, try without \"let\""
```
paf31: How does this show up?

paf31:
Thanks! It seems much better this way.

noraesae:
I added a `Show` instance for `Command` for better inspecting with REPL.

About parseLocalDeclaration, I had actually tried it before I uploaded the previous patch but in the case the error doesn't show up with multiple decls in `:paste`.

```
> :paste
… let a = 10
…     b = 20
…
(line 2, column 9):
unexpected end of input
expecting indentation at column 5 or "in"
>
```

I've further investigated on this but still not sure why this happens. Instead, I added `notFollowedBy` to check to specify the declaration case. Just using the `many1` with this check seems better to me. Please leave a comment if any.

src/Language/PureScript/Interactive/Parser.hs
```diff
@@ -112,3 +114,13 @@ parseReplQuery' str =
...
+
+-- | To show error message when 'let' is used for declaration in PSCI,
+-- which is deprecated.
+psciDeprecatedLet :: P.TokenParser Command
+psciDeprecatedLet = do
+  P.reserved "let"
+  P.indented
+  _ <- mark (many1 (same *> P.parseLocalDeclaration))
+  notFollowedBy $ P.reserved "in"
+  fail "for declaration in PSCi, try without \"let\""
```
paf31: How about "Declarations in PSCi no longer require "let", as of version 0.11.0"?

noraesae:
Updated the message. Coming up with an English text is always a thing for me. 😅 

```
> let b = 10
(line 1, column 9):
unexpected end of input
expecting ., {, ::, operator, "where", indentation at column 5 or "in"
Declarations in PSCi no longer require "let", as of version 0.11.0
> :paste
… let f 0 = 100
…     f x = x
…
(line 2, column 11):
unexpected end of input
expecting indentation at column 5 or "in"
Declarations in PSCi no longer require "let", as of version 0.11.0
>
```

paf31:
Thanks!

Issue #2610, opened 2017-01-31
psc-publish: Unable to parse the date for a git tag


owickstrom:
I have a tag `v0.1.2` in https://github.com/owickstrom/hyper which causes `psc-publish` to fail. I'd normally run it using pulp publish, but it seems to be an issue with psc-publish itself.

```bash
$ psc-publish --version
0.10.5
$ git checkout v0.1.2
...
HEAD is now at c065744... v0.1.2
$ psc-publish
Internal error: this is probably a bug. Please report it:
  https://github.com/purescript/purescript/issues/new

Details:
Unable to parse the date for a git tag: v0.1.2
```

Looking at the tag:

```bash
$ git show v0.1.2
tag v0.1.2
Tagger: Oskar Wickström <oskar.wickstrom@gmail.com>
Date:   Tue Jan 31 07:59:50 2017 +0100

v0.1.2

commit c065744f9b5c895b06566caf13393294ab8fe882
Author: Oskar Wickström <oskar.wickstrom@gmail.com>
Date:   Tue Jan 31 07:59:50 2017 +0100

    v0.1.2
```

The command run at https://github.com/purescript/purescript/blob/62a30ad868c764b36e93a44544325e2eb9c49034/src/Language/PureScript/Publish.hs#L199 seems to be the one that fails. When I run it manually, I get multiline output, which might be the problem (the parsing code seems to expect the first line to be the UTC date):

```bash
$ git show v0.1.2 --no-patch --format='%aI'
tag v0.1.2
Tagger: Oskar Wickström <oskar.wickstrom@gmail.com>

v0.1.2
2017-01-31T07:59:50+01:00
```

OK, if I try `--dry-run` instead:

```bash
$ psc-publish  --dry-run
fatal: ambiguous argument '0.0.0': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
An error occurred, and your package could not be published.
Details:
  While running `git show 0.0.0 --no-patch --format=%aI`:
    readCreateProcess: git "show" "0.0.0" "--no-patch" "--format=%aI" (exit 128):
    failed
```

It tries to show `0.0.0`, which does not exist.

hdgarrood:
Does `psc-publish` normally work on your machine? Are you able to identify anything unique about this tag? Also what version of git are you using?

owickstrom:
[...]

PR #2669, opened 2017-02-16
Include git commit information in non-release builds

hdgarrood:
I've wanted to do this for a while but only just learned about the `gitrev` Haskell package, which makes this much easier.

---

This should help avoid confusion like in #2610. The output of `purs
--version` is unchanged for release builds but includes extra
information like the git tag for non-release builds.

By default, `stack build` will produce a non-release build, which means
that `purs --version` will include information like this:

$ purs --version
0.10.6 [development build; commit: b8136e312fbda48990a576e0802e0c0c7d6cda5f]

To make a release build, set the "release" cabal flag. With stack:

$ stack build --flag purescript:RELEASE

The Travis CI and AppVeyor scripts have been modified to do this where
appropriate.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -111,6 +111,11 @@ source-repository head
     type: git
     location: https://github.com/purescript/purescript.git
 
+flag release
+  description: Mark this build as a release build: prevents inclusion of extra
+               info e.g. commit SHA in --version output)
+  default: False
+
 library
     build-depends: base >=4.8 && <5,
                    aeson >= 0.8 && < 1.1,
@@ -390,6 +395,11 @@ executable purs
                    Command.Publish
                    Command.REPL
     ghc-options: -Wall -O2 -fno-warn-unused-do-bind -threaded -rtsopts "-with-rtsopts=-N"
+
+    if flag(release)
+      cpp-options: -DRELEASE
+    else
+      build-depends: gitrev >= 1.2.0 && <1.3
 
 test-suite tests
     build-depends: base >=4 && <5,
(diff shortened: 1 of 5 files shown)


hdgarrood:
Travis is failing because one of the OSX jobs is failing to download the cache. This might be a bug in Travis. I will try to follow up soon.

hdgarrood:
Huh, ok, travis cache seemed to work the second time. Now appveyor is failing and I'm not sure why. I think powershell might be considering a command to have failed regardless of its exit code if it prints anything to stderr. If it is that, that's really disappointing because MS should know better...

purescript.cabal
```diff
@@ -391,6 +396,11 @@ executable purs
                    Command.REPL
     ghc-options: -Wall -O2 -fno-warn-unused-do-bind -threaded -rtsopts "-with-rtsopts=-N"
 
+    if flag(release)
+      cpp-options: -DRELEASE
+    else
+      build-depends: gitrev >= 1.2.0 && <1.3
```
paf31: Presumably we don't need to update the license because we don't ship with this library. Does `license-generator` handle that properly?

hdgarrood: Good question, probably not. I'll check.

hdgarrood:
Thanks for the review :)

To do list for me:

* [ ] fix windows CI
* [ ] ensure license-generator handles this appropriately

paf31:
@hdgarrood Ready to merge now?

hdgarrood:
I was planning to look at what's going on with the CI first, but I haven't had a chance yet. I'm hoping to do that tomorrow

PR #2673, opened 2017-02-18
publish: Allow for older git versions

mcoffin:
Not adding any tests per same reasoning as #2613.

Currently, git versions < 2.2 will fail with `psc-publish` because they don't have the `iso-strict` date format available. Since we're just parsing in to unix time anyway, we can easily have git just output the unix timestamp in epoch seconds and parse that instead, ensuring compatibility with older git versions (most notably the one currently used by travis-ci).

EDIT: Some discussion also happened in #2610


paf31:
Thanks, but please also update CONTRIBUTORS.md.

mcoffin:
@paf31 Done. My bad. That's what I get for only skimming the contributing guidelines.

PR #2674, opened 2017-02-18
Update LICENSE

hdgarrood:
Now that we no longer depend on `turtle`.


rightfold:
>  Copyright (c) 2013-16 Phil Freeman, (c) 2014-2016 Gary Burgess, and other

Might as well update these year numbers?

hdgarrood:
Ahh, travis seems to be intermittently failing to download build caches for OSX builds. We might have to disable them again...

Issue #2437, opened 2016-11-18
Require module names to only be unique per package


hdgarrood:
Related to https://github.com/purescript/purescript/issues/2415.  /cc @ivan-kleshnin

At the moment, module names are globally unique. This presents two problems:

1. If you depend on two separate packages which both define a module with the same name, you are stuck. The only recourse is to nicely ask the authors of one or the other to change a module name, or fork one and change the module name yourself.
2. If you're reading someone else's code and you see code from packages you're not familiar with, it's not clear which packages certain modules come from.
3. We have this slightly weird hierarchy, inherited from Haskell, that sometimes makes it difficult to guess where a module goes and is often a bit verbose. ("Do I want `Data.Lazy` or `Control.Lazy`?" "Why is it `Data.Functor` but `Control.Monad`?" "I am a little bored of typing out `Control.Monad.Eff.Console`.")

Before continuing I would like to draw attention to other initiatives that do partially address both of these:

1. Package sets address the uniqueness problem by requiring that all packages in a set build together, thereby ensuring that within a package set, module names are unique. However I don't think that we should expect all library code will end up in a package set; putting libraries into a package set is a commitment that not everyone will want or be able to make.
2. Editor integration helps track down where names or modules come from. However editor integration isn't available everywhere - for example, when reviewing code on web apps like GitHub, when reading blog posts on the web which include PureScript code, or when it just isn't installed on the machine you're using (for whatever reason - maybe it isn't your main dev machine).
3. It should be possible to input module names in the search box in Pursuit, and the search results should link to all the packages which define modules of that name. But again, sometimes I'm writing code and I don't have internet access, and if there are lots of imports in a module which I don't recognise, this is a lot of hassle.

The only way I can see to fully address these issues is to change the language so that modules must only be unique within a package, and change import declarations so that they explicitly say which package a module is coming from. I agree that we ideally want to keep the compiler as unaware of our approach to packaging as possible - this enables us to work on experiments like `psc-package` without compiler changes, which is great. So I propose that, as far as the compiler is concerned, a "package" is just a name (which must be unique) and a collection of modules. Import declarations could look like this:

```purescript
import "prelude" Prelude
import "lists" List as List
import "foldable-traversable" Traversable (traverse)
import "console" Console (log, logShow)
import "transformers" Writer.Trans as WriterT
import "transformers" Reader.Class (asks)
```

[...]

PR #2676, opened 2017-02-19
Remove dependence on Bower

paf31:
WIP:

- [ ] Make `description` field not required
- [ ] Fix `repository` field so that it's still compatible with `bower.json`
- [ ] Finish implementing `ToJSON` for `PackageMeta`
- Remove the few remaining references to Bower:
  - [ ] In the tests?
  - [ ] In the globs in `Publish.Utils` (so we need to figure out a way for `purs publish` to find out the source files for each dependency on disk, so that we can figure out the modules)
- [ ] Figure out what needs to change in Pulp to support Bower there


app/Command/Publish.hs
```diff
@@ -13,6 +13,11 @@ import           Language.PureScript.Publish.ErrorsWarnings
 import           Options.Applicative (Parser)
 import qualified Options.Applicative as Opts
 
+manifestPath :: Parser FilePath
```
hdgarrood: Did you consider having it read from stdin instead of using a file path? I'm not actually sure which approach I prefer now.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -38,6 +37,22 @@ import Language.PureScript.Docs.RenderedCode as ReExports
 --------------------
 -- Types
 
+newtype PackageName = PackageName { getPackageName :: Text }
```
hdgarrood: So I think I'd rather define this type in its own module and not export the constructor.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -38,6 +37,22 @@ import Language.PureScript.Docs.RenderedCode as ReExports
...
+newtype VersionRange = VersionRange { getVersionRange :: Text }
+  deriving (Show, Eq, Ord)
+
+data PackageMeta = PackageMeta
+  { packageMetaName            :: PackageName
+  , packageMetaDescription     :: Text
+  , packageMetaLicense         :: [Text]
+  , packageMetaAuthors         :: [Text]
+  , packageMetaRepository      :: Text
+  , packageMetaDependencies    :: [(PackageName, Version)]
```
hdgarrood: You could probably just copy the bits we need over from bower-json actually.

src/Language/PureScript/Docs/Types.hs
hdgarrood: Oh also I think we want an NFData instance for Pursuit.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -48,7 +63,6 @@ data Package a = Package
   , pkgTagTime              :: Maybe UTCTime
   , pkgModules              :: [Module]
   , pkgModuleMap            :: Map P.ModuleName PackageName
-  , pkgResolvedDependencies :: [(PackageName, Version)]
```
paf31: How would we get the specific versions without running the `bower` executable though?

paf31: I agree we should probably try to keep compatibility with `bower.json` though. In fact, after working through this, I'm thinking we should continue using your `bower-json` library.

app/Command/Publish.hs
paf31: I didn't really think much about it yet, and we can definitely change it. I don't feel strongly either way.

paf31:
There are some open questions still:

- How do we get specific versions of installed libraries without depending on Bower (the executable)? Should the caller provide them?
- How do we find source files for dependencies so that we can construct links? Perhaps we could assume they are under `<some_subdirectory>/**/purescript-<name>/src/**/*.purs` or something like that. That would at least be compatible with Bower and `psc-publish`.

hdgarrood:
I think the caller will have to provide specific versions of installed libraries if we want to be package manager agnostic.

Your suggestion for locating dependencies together with their names makes sense to me too - although I think in the long term, for the sake of uniformity, I'd prefer to use normal globs like `purs build` does and deduce package names from module names (of course this can't happen unless we start allowing package names in module names).

paf31:
> I think in the long term, for the sake of uniformity, I'd prefer to use normal globs like purs build does and deduce package names from module names

Sounds good, although I think we should try to decide on an API that we want to commit to before 1.0.

hdgarrood:
How would you feel about doing #2437 first so that we can then do this with normal globs straight away? That way there wouldn't be as many breaking changes to the psc-publish API

paf31:
@hdgarrood I'm reading through the issues again, to remind myself of everything, but I think in the short term, I'd like to understand at a high level what problem we're trying to solve, and then to decide whether it's worth whatever breakages would be necessary.

Absent a real module system, which is definitely not a 1.0 feature in my book, I'm leaning towards saying the current system is good enough, but if we can improve things incrementally with non-breaking changes, then that's different.

hdgarrood:
I think the design we were converging on in #2437 would be an incremental non-breaking improvement (although libraries would not be able to take advantage of the new feature without releasing a fairly significant breaking change).

paf31:
@hdgarrood Can you please summarize the solution as you understand it?

hdgarrood:
The solution of #2437?

paf31:
Yes, is there a brief summary? There are multiple suggestions there, and I'm not sure exactly which one you're referring to.

hdgarrood:
Sure - I'll write one and post it on that issue.

paf31:
Here is what I would prefer: every package gets an assigned namespace which it owns by convention. So `foldable-traversable` might be defined under `FoldableTraversable.*` or something. Tools like `purs publish` could warn or error if the modules don't use that namespace. That solves the discoverability problem, and pushes the uniqueness problem to whatever registry we use (Bower, etc.)

I like this option because it builds on what we have now, it's opt-in by module developers, and the improved tooling it enables will put pressure on module developers to break their libraries to use the convention we adopt.

hdgarrood:
Ok then - do we want to try to come up with an injective function from package names to namespaces, so that we can deduce the namespace from the package name, or would we store all these namespace assignments somewhere?

paf31:
I think if we want to solve the discoverability problem properly then it needs to be a function that the user can evaluate mentally, so it can't depend on some centralized mapping.

hdgarrood:
Ok cool, agreed. So I guess the obvious mapping is `split "-" >>> map capitalize >>> joinWith ""` where `capitalize` is defined so that `capitalize "foldable" == "Foldable"`? The only potential issue I see is that Bower packages are allowed to contain dots; I guess we could warn or error in `purs publish` if someone is trying to do something with a package that contains a dot.

I guess we'd want to start broadcasting that we plan to do this and start making those changes in libraries now, so that people have time to adjust before we release a version of `purs publish` that requires you to follow the convention?

paf31:
Yes, let's try to figure out the details a bit more and write it up. Obviously this will require breaking a LOT of library code, but now is the time to do it of course.

hdgarrood:
To clarify: `purs publish` will need to be able to determine which packages a particular module comes from, and I think the options are to deduce package names from a) file paths or b) module names. I'd rather do b) because then the command line API is simpler: you just pass file globs like you do for the other `purs` subcommands. a) requires coming up with some file path templating scheme and also means that the API would be somewhat different to other `purs` subcommands, as you pointed out earlier in the thread.

hdgarrood:
Cool, I'll try to do that this evening.

paf31:
I suppose the other option would be to require the caller to pass a mapping from packages to globs on the command line. That wouldn't be so bad, but it wouldn't solve the other related issues (discoverability, enforcing uniqueness globally). Whether those are worth the breakage, I don't know, but we can discuss it.

The other thing we need to solve here is how to get `purs publish` to figure out the exact versions of each library which are being used.

hdgarrood:
I think it would be fine to take the code which figures out exact versions being used via Bower's CLI and move that code into pulp, and then have the caller pass exact versions in. I think most people use `pulp publish` rather than `psc-publish` directly now so we could do that without breaking most people's workflows (although I can't be certain of this).

paf31:
Well, maybe it wouldn't be so bad to have Pulp supply the glob mapping as well for now, in that case. That would simplify  `purs publish`, allow us to remove the Bower dependency, keep the workflow the same for most users, and give us more time to figure out the convention we want for package naming.

hdgarrood:
As in, have Pulp supply a mapping of package names to file globs? Something like 

```json
{
  "foldable-traversable": "bower_components/purescript-foldable-traversable/src/**/*.purs",
  ...
}
```

paf31:
I'll recreate this soon.

michaelficarra:
@paf31 Was there ever any follow-up to this?

hdgarrood:
Yes, it made it into v0.11.0. See [the release notes](https://github.com/purescript/purescript/releases/tag/v0.11.0).

PR #2677, opened 2017-02-19
simplify a hairy use of M.insert

matthewleon:
Makes it a bit clearer what the code in question is doing.

### DIFF
--- a/src/Language/PureScript/Sugar/Names.hs
+++ b/src/Language/PureScript/Sugar/Names.hs
@@ -350,11 +350,8 @@ renameInModule imports (Module ss coms mn decls exps) =
       -- in scope, we throw an error.
       (Just options, _) -> do
         (mnNew, mnOrig) <- checkImportConflicts mn toName options
-        modify $ \result ->
-          M.insert
-            mnNew
-            (maybe [fmap toName qname] (fmap toName qname :) (mnNew `M.lookup` result))
-            result
+        modify $ \usedImports ->
+          M.insertWith (++) mnNew [fmap toName qname] usedImports
         return $ Qualified (Just mnOrig) name
 
       -- If the name wasn't found in our imports but was qualified then we need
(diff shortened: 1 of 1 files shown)


paf31:
Thanks, but why use `++`? Can't this use cons?

matthewleon:
Uses cons now.

paf31:
Sorry, I forgot that `insertWith` uses a function of type `a -> a -> a` and I thought you could write `insertWith (:)`. I think your original version was better.

matthewleon:
Done.

By the way, please let me know if this kind of PR is generally acceptable/encouraged or if it's something of a distraction. I came across this code while working on something else, and it seemed worthwhile to make it clearer... But such things are of course subjective.

hdgarrood:
I think this is an improvement, and in general I'm in favour of PRs like this; we can always accept or reject them case by case anyway. Also I wasn't aware of `insertWith` so it's nice to be made aware of it.

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -350,11 +350,8 @@ renameInModule imports (Module ss coms mn decls exps) =
       -- in scope, we throw an error.
       (Just options, _) -> do
         (mnNew, mnOrig) <- checkImportConflicts mn toName options
-        modify $ \result ->
-          M.insert
-            mnNew
-            (maybe [fmap toName qname] (fmap toName qname :) (mnNew `M.lookup` result))
-            result
+        modify $ \usedImports ->
+          M.insertWith (++) mnNew [fmap toName qname] usedImports
```
garyb: Much better 👍 but could be further simplified to:

``` haskell
modify $ M.insertWith (++) mnNew [fmap toName qname]
```

rvion: I think it is slightly more readbale this way for newcomers to the codebase like me :)
if you don't mind the redundance, I would prefer to avoid pointfree in this particular case where the variable name helps understanding what's going on

paf31: I see where you're coming from, but I think this is a relatively straightforward use of point-free style.

PR #2679, opened 2017-02-20
replace nub with ordNub

matthewleon:
Replace pervasive uses of `nub` with `ordNub` from Protolude, which has better asymptotics. Where necessary, derive some new `Ord` instances to help.

Would be very curious to see some benchmarks with this change. I don't have anything at hand to test (sorry).

### DIFF
--- a/app/Command/Docs.hs
+++ b/app/Command/Docs.hs
@@ -2,6 +2,8 @@
 
 module Command.Docs (command, infoModList) where
 
+import           Protolude (ordNub)
+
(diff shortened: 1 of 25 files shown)


matthewleon:
Anecdotally, there are some reports of speed increase even with relatively small data: https://mail.haskell.org/pipermail/haskell-cafe/2015-January/117567.html

http://neilmitchell.blogspot.com/2015/02/nub-considered-harmful.html

paf31:
👍 Thanks!

PR #2681, opened 2017-02-20
Add psc-ide-client script

paf31:
This also modifies the existing scripts to use quotes. This seems to be necessary in order to make things like `scripts/psci 'bower_components/purescript-*/src/**/*.purs'` work properly.

### DIFF
--- a/scripts/psc
+++ b/scripts/psc
@@ -1,2 +1,2 @@
 #!/bin/sh
-purs compile $@
+purs compile "$@"
--- a/scripts/psc-bundle
+++ b/scripts/psc-bundle
@@ -1,2 +1,2 @@
 #!/bin/sh
-purs bundle $@
+purs bundle "$@"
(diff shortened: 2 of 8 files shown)


hdgarrood:
One thing I've just realised is that Pulp runs `psc --version` to check that you have a sufficiently recent version of the compiler. It might be worth changing the `psc` wrapper to run `purs --version` if it receives a `--version` argument. See: https://github.com/bodil/pulp/blob/master/src/Pulp/Validate.purs

hdgarrood:
Alternatively I can try to deal with this within Pulp, and try to get a version of Pulp released that can cope with the `purs` executable now / before 0.11 is released.

paf31:
It seems to work actually:

```
$ scripts/psc --version
0.10.6
```

paf31:
`purs compile --version` works for me actually.

hdgarrood:
Oh cool, it works for me too, and it must be using the one defined in `app/Main.hs` because it includes the info about it being a development build. Great :)

PR #2687, opened 2017-02-22
Remove dependency on the Bower executable

paf31:
@hdgarrood I think this simplifies things a bit.

We continue to use the Bower JSON formats for `bower.json` (called the "manifest") and the output of `bower list --json` (called the "resolutions" file). Both are passed in as command line arguments so we don't rely on the Bower executable directly. Maybe `psc-package` could create these files somehow later.

- [x] Extract dependency paths from the output of `bower list --json`
- [x] Get the tests working

For #2517

### DIFF
--- a/src/Language/PureScript/Publish.hs
+++ b/src/Language/PureScript/Publish.hs
@@ -26,7 +26,7 @@ import Control.Arrow ((***))
 import Control.Category ((>>>))
 import Control.Monad.Writer.Strict (MonadWriter, WriterT, runWriterT, tell)
 
-import Data.Aeson.BetterErrors (Parse, parse, keyMay, eachInObjectWithKey, eachInObject, key, keyOrDefault, asBool, asText)
+import Data.Aeson.BetterErrors (Parse, parse, keyMay, eachInObjectWithKey, eachInObject, key, keyOrDefault, asBool, asString, asText)
 import qualified Data.ByteString.Lazy as BL
 import Data.Char (isSpace)
 import Data.String (String, lines)
@@ -128,20 +128,24 @@ preparePackage' manifestFile resolutionsFile opts = do
   (pkgVersionTag, pkgVersion) <- publishGetVersion opts
   pkgTagTime                  <- Just <$> publishGetTagTime opts pkgVersionTag
   pkgGithub                   <- getManifestRepositoryInfo pkgMeta
-  (pkgModules, pkgModuleMap)  <- getModules
 
   let declaredDeps = map fst (bowerDependencies pkgMeta ++
                               bowerDevDependencies pkgMeta)
-  pkgResolvedDependencies     <- getResolvedDependencies resolutionsFile declaredDeps
+  resolvedDeps                <- getResolvedDependencies resolutionsFile declaredDeps
+
+  (pkgModules, pkgModuleMap)  <- getModules (map (second fst) resolvedDeps)
 
   let pkgUploader = D.NotYetKnown
   let pkgCompilerVersion = P.version
+  let pkgResolvedDependencies = map (second snd) resolvedDeps
 
   return D.Package{..}
 
-getModules :: PrepareM ([D.Module], Map P.ModuleName PackageName)
-getModules = do
-  (inputFiles, depsFiles) <- liftIO getInputAndDepsFiles
+getModules
+  :: [(PackageName, FilePath)]
+  -> PrepareM ([D.Module], Map P.ModuleName PackageName)
+getModules paths = do
+  (inputFiles, depsFiles) <- liftIO (getInputAndDepsFiles paths)
   (modules', moduleMap) <- parseFilesInPackages inputFiles depsFiles
 
   case runExcept (D.convertModulesInPackage modules' moduleMap) of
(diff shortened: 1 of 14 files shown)


hdgarrood:
Looks good so far; what are you planning to do with the TODO?

hdgarrood:
Oops, you already answered that

paf31:
@hdgarrood I've updated the code to extract the dependency source paths from the output of `bower list`.

I'll work on fixing up the tests now.

paf31:
@hdgarrood This should be good to review now - I've updated the tests to generate the resolutions files using `bower` and to pass them to the publish command by filename.

src/Language/PureScript/Publish/Utils.hs
```diff
@@ -16,14 +16,20 @@ import qualified System.FilePath.Glob as Glob
 globRelative :: Glob.Pattern -> IO [FilePath]
 globRelative pat = do
   currentDir <- getCurrentDirectory
-  filesAbsolute <- Glob.globDir1 pat currentDir
+  glob pat currentDir
+
+-- | Glob relative to the specified directory, and produce relative pathnames.
```
hdgarrood: This seems a bit odd; these two functions have the same docstring.

hdgarrood: Oh my god, I'm an idiot.

hdgarrood: However it does seem a bit odd that `glob` is using `getCurrentDirectory`; shouldn't it be using `target` instead of the result of `getCurrentDirectory`?

tests/TestUtils.hs
```diff
@@ -165,3 +165,8 @@ pushd dir act = do
   result <- try act :: IO (Either IOException a)
   setCurrentDirectory original
   either throwIO return result
+
+generateResolutionsFile :: FilePath -> IO ()
+generateResolutionsFile fp =
+  readProcess "bower" ["list", "--offline", "--json"] "" >>=
```
hdgarrood: Should we be worrying about lazy IO here?

src/Language/PureScript/Publish.hs
```diff
@@ -300,111 +297,93 @@ getResolvedDependencies declaredDeps = do
...
-      withVersions <- catMaybes <$> traverse tryExtractVersion' installed
-      filterM (liftIO . isPureScript . bowerDir . fst) withVersions
+      traverse_ (warn . MissingPath) missingPath
+      catMaybes <$> traverse tryExtractVersion' installed
 
   where
-  partitionDeps = foldr go ([], [], [])
-  go (pkgName, d) (ms, os, is) =
+  partitionDeps = foldr go ([], [], [], [])
+  go (pkgName, (Nothing, _)) (ms, os, is, mp) =
```
hdgarrood: I really don't like this code (which I take full responsibility for), btw. It's way too easy to miss things out accidentally. If you can think of a better way of doing it I'm all ears.

hdgarrood:
Looks good, I've made a few minor comments. One last thing that just occurred to me is that I think it might make sense to try to get `psc-package` to produce JSON in this format so that `purs publish` will accept it, and it might make sense to try to do that before merging this, just in case it turns out to be awkward?

src/Language/PureScript/Publish.hs
paf31: It feels like there is a `Monoid` in here somewhere :)

tests/TestUtils.hs
paf31: I don't think so - I would expect `readProcess` to produce data incrementally, and `writeFile` should consume its input string entirely, so the resource should be cleaned up when the input string is fully consumed.

src/Language/PureScript/Publish/Utils.hs
paf31: Well, I want to find files in `target`, but then test for a prefix which begins with the current directory. Maybe I should add that as a comment?

paf31:
This is odd, the test suite runs file for me locally.

src/Language/PureScript/Publish/Utils.hs
hdgarrood: So the result will be a list of path names relative to the current directory? I don't understand why you might want that, why does it work like that? It's not really what I would expect given the comment it has at the moment.

tests/TestUtils.hs
hdgarrood: Ok, cool, just checking :smile:

src/Language/PureScript/Publish/Utils.hs
paf31: I've updated the comment. I started with the original code, and this seemed to work fine.

paf31:
Seems like the tests are failing for different reasons on Windows and Linux. I don't have time to fix this today, but I might tomorrow. Maybe the simplest thing is to just check in the resolutions files.

paf31:
@hdgarrood This time, I've checked in the resolutions files for the various tests, so hopefully Travis and Appveyor should be happy now.

paf31:
As for the `psc-package` idea, I'm almost certain there won't be any issues there. Have a look at the reduced JSON I checked in for example. Generating that should be fairly trivial.

paf31:
Weird, Travis seems to be working now, except for the coverage build, where it can't find the `resolutions.json` file.

paf31:
@hdgarrood Aside from the weird test failure, does this look okay to you now?

paf31:
Ok, I think I have the test issue figured out now.

paf31:
Thank you both!

PR #2696, opened 2017-02-27
Allow duplicate labels

### LINKED ISSUE #2311: Remove duplicate rows restriction
Extracted from #1719.

paf31:
I've added some tests for the sorts of things I'd like to allow, but it's not working yet. Next, I'm going to work through the type checker and make sure that we always keep types for the same label in order.

Fixes #2311


andyarvanitis:
Will this rely on JavaScript's duplicate object properties behavior?

chexxor:
This is pure type-checker stuff, I believe. PS type rows don't hit the target language, do they?

paf31:
@andyarvanitis No, the meaning of a record would only use the first type with that label. Since we don't support deletion of fields, I think that should be safe.

andyarvanitis:
Great, thanks

paf31:
This is now ready for review. There isn't actually all that much here - most of this is refactoring to extract some common code into the new `alignRowsWith` function. The important thing is that, unlike the previous approach, `alignRowsWith` preserves the order of the (possibly multiple) types for a given label.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -392,24 +392,19 @@ matches deps TypeClassDictionaryInScope{..} tys = do
...
+
+        go :: ([(Label, Type)], Type) -> ([(Label, Type)], Type) -> (Bool, Matching [Type])
+        go (l,  KindedType t1 _)  (r,  t2)                            = go (l, t1) (r, t2)
+        go (l,  t1)               (r,  KindedType t2 _)               = go (l, t1) (r, t2)
+        go ([], REmpty)           ([], REmpty)                        = (True, M.empty)
+        go ([], TUnknown u1)      ([], TUnknown u2)      | u1 == u2   = (True, M.empty)
+        go ([], TypeVar v1)       ([], TypeVar v2)       | v1 == v2   = (True, M.empty)
+        go ([], Skolem _ sk1 _ _) ([], Skolem _ sk2 _ _) | sk1 == sk2 = (True, M.empty)
+        go (sd, r)                ([], TypeVar v)                     = (True, M.singleton v [rowFromList (sd, r)])
+        go _ _                                                       = (False, M.empty)
```
kRITZCREEK: This made me twitch

src/Language/PureScript/Types.hs
```diff
@@ -107,24 +110,22 @@ $(A.deriveJSON A.defaultOptions ''Type)
...
+  go (RCons name ty row) =
+    first ((name, ty) :) (rowToList row)
+  go r = ([], r)
 
--- |
--- Convert a list of labels and types to a row
---
+-- | Convert a row to a list of pairs of labels and types, sorted by the labels.
+rowToSortedList :: Type -> ([(Label, Type)], Type)
+rowToSortedList = first (sortBy (comparing fst)) . rowToList
```
kRITZCREEK: Is this safe if the ordering of entries with the same label matters? I guess only if sortBy is stable?

paf31: Correct, it needs to be stable

src/Language/PureScript/TypeChecker/Entailment.hs
paf31: Ugh yes, fixed :)

paf31:
Thanks @kRITZCREEK

PR #2697, opened 2017-02-28
Errors for open rows in derived instances

### LINKED ISSUE #2616: Lack of warning/error when using deriving with polymorphic records
Let's say I have some code like this:

```
newtype Foo = Foo { foo :: Int }

derive instance eqFoo :: Eq Foo
derive instance ordFoo :: Ord Foo

(Foo { foo: 1 }) == (Foo { foo: 2 })
(Foo { foo: 1 }) > (Foo { foo: 2 })
```

This works perfectly fine, exactly what I expected.

But now let's say I have some code like this:

```
newtype Foo r = Foo { | r }

derive instance eqFoo :: Eq (Foo r)
derive instance ordFoo :: Ord (Foo r)

(Foo { foo: 1 }) == (Foo { foo: 2 })
(Foo { foo: 1 }) > (Foo { foo: 2 })
```

This does not behave at all like I expected: `==` always returns `true`, and all of the `Ord` operations always return `EQ`:

```
var eqFoo = new Data_Eq.Eq(function (x) {
    return function (y) {
        return true;
    };
});

var ordFoo = new Data_Ord.Ord(function () {
    return eqFoo;
}, function (x) {
    return function (y) {
        return Data_Ordering.EQ.value;
    };
});
```

Unfortunately, PureScript does nothing to prevent this: there isn't any errors, and there isn't even a warning.

I think PureScript should warn/error when deriving instances that don't work in an expected way (like with polymorphic records).

paf31:
Fixes #2616

Note, open rows are allowed in `Functor` instances, but not elsewhere. While the `Eq` and `Ord` instances are technically valid, I don't think we want to derive instances like that, right?

### DIFF
--- a/examples/failing/2616.purs
+++ b/examples/failing/2616.purs
@@ -1,0 +1,9 @@
+-- @shouldFailWith NoInstanceFound
+module Main where
+	
+import Prelude
+
+newtype Foo r = Foo { | r }
+
+derive instance eqFoo :: Eq (Foo r)
+derive instance ordFoo :: Ord (Foo r)
--- a/examples/passing/2616.purs
+++ b/examples/passing/2616.purs
@@ -1,0 +1,13 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+
+newtype F r a = F { x :: a | r }
+
+unF :: forall r a. F r a -> { x :: a | r }
+unF (F x) = x
+
+derive instance functorF :: Functor (F r)
+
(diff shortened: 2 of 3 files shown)


garyb:
I think open rows are also valid for `Newtype`, if we're making the exception for `Functor` also. Code LGTM though. A shame the error will be `NoInstanceFound` though, as that's not the most obvious way of explaining the problem :wink:

paf31:
@garyb `Newtype` should still work.

I think there is a separate issue to deal with the error messages, which is why I left it like this for now.

PR #2706, opened 2017-03-05
Fix Generic deriving with synonyms

### LINKED ISSUE #2695: Record synonyms fail to derive with Generic
I didn't notice an issue about this (maybe it is expected?), but while most synonyms will work with deriving, record synonyms still don't.

http://try.purescript.org/?gist=ea6794e4608f13c46a77148ff3b36dcf

If you inline the record synonym in the newtype, or use something other than a record, it will work.

paf31:
Fixes #2695

### DIFF
--- a/examples/passing/2695.purs
+++ b/examples/passing/2695.purs
@@ -1,0 +1,13 @@
+module Main where
+
+import Prelude
+import Data.Generic
+import Control.Monad.Eff.Console (log)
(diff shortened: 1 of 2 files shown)


LiamGoodacre:
This looks good :+1:

paf31:
Thanks @LiamGoodacre!

PR #2710, opened 2017-03-05
Find record accessors in Type Directed Search

kRITZCREEK:
This PR consists of two parts:

## 1. Infer record accessors for types that subsume `{| r} -> a`

This means that for `?hole {x: 1, y: "hello"}` TDS infers `_.x` and `_.y`. This also works for situations like `{x: 1, y: "hello"} # ?hole`.

These results are not shown in the textual error reporting, because I felt they were to noisy, but they are exposed through the new

## 2. JSON error output for Type Directed Search results

Here I considered using the existing suggestion API to report the results, but decided it lacks two crucial capabilities:
- Reporting multiple suggestions for a single error
- A structured format, since TDS results consist of module name + identifier + type

This is why I decided to use purs ide's existing Completion format and monkeypatch it onto the JSON format for `HoleInferred` errors. Right now this only works for errors reported by purs ide, but this can easily be changed by lifting the `Completion` datatype into a module above `Language.PureScript.Errors`. I wanted to gather feedback before doing that though.

As a proof of concept I've written a branch of psc-ide-emacs that can fill in typed holes by using the result reported through rebuilding.

This PR makes the too fragile #2556 obsolete.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -39,8 +39,11 @@ data TypeSearch
   = TSBefore Environment
   -- ^ An Environment captured for later consumption by type directed search
   | TSAfter
-    { tsAfterIdentifiers :: [(Qualified Ident, Type)]
+    { tsAfterIdentifiers :: [(Qualified Text, Type)]
+    -- ^ The identifiers that fully satisfy the subsumption check
     , tsAfterRecordFields :: Maybe [(Label, Type)]
+    -- ^ Record fields that are available on the first argument to the typed
+    -- hole
     }
   -- ^ Results of applying type directed search to the previously captured
   -- Environment
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -38,10 +38,20 @@ type Context = [(Ident, Type)]
 data TypeSearch
   = TSBefore Environment
   -- ^ An Environment captured for later consumption by type directed search
-  | TSAfter [(Qualified Ident, Type)]
+  | TSAfter
+    { tsAfterIdentifiers :: [(Qualified Ident, Type)]
+    , tsAfterRecordFields :: Maybe [(Label, Type)]
+    }
   -- ^ Results of applying type directed search to the previously captured
   -- Environment
   deriving Show
+
+onTypeSearchTypes :: (Type -> Type) -> TypeSearch -> TypeSearch
+onTypeSearchTypes f = runIdentity . onTypeSearchTypesM (Identity . f)
+
+onTypeSearchTypesM :: (Applicative m) => (Type -> m Type) -> TypeSearch -> m TypeSearch
+onTypeSearchTypesM f (TSAfter i r) = TSAfter <$> (traverse (traverse f) i) <*> (traverse (traverse (traverse f)) r)
+onTypeSearchTypesM _ (TSBefore env) = pure (TSBefore env)
 
 -- | A type of error messages
 data SimpleErrorMessage
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -50,7 +50,7 @@ onTypeSearchTypes :: (Type -> Type) -> TypeSearch -> TypeSearch
 onTypeSearchTypes f = runIdentity . onTypeSearchTypesM (Identity . f)
 
 onTypeSearchTypesM :: (Applicative m) => (Type -> m Type) -> TypeSearch -> m TypeSearch
-onTypeSearchTypesM f (TSAfter i r) = TSAfter <$> (traverse (traverse f) i) <*> (traverse (traverse (traverse f)) r)
+onTypeSearchTypesM f (TSAfter i r) = TSAfter <$> traverse (traverse f) i <*> traverse (traverse (traverse f)) r
 onTypeSearchTypesM _ (TSBefore env) = pure (TSBefore env)
 
 -- | A type of error messages
--- a/src/Language/PureScript/TypeChecker/TypeSearch.hs
+++ b/src/Language/PureScript/TypeChecker/TypeSearch.hs
@@ -4,7 +4,7 @@ module Language.PureScript.TypeChecker.TypeSearch
 
 import           Protolude
 
-import           Control.Monad.Writer
+import           Control.Monad.Writer (WriterT, runWriterT)
 import qualified Data.Map                                    as Map
 import qualified Language.PureScript.TypeChecker.Entailment  as Entailment
 
--- a/src/Language/PureScript/TypeChecker/TypeSearch.hs
+++ b/src/Language/PureScript/TypeChecker/TypeSearch.hs
@@ -83,7 +83,7 @@ accessorSearch
   -> P.Environment
   -> TC.CheckState
   -> P.Type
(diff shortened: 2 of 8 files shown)


nwolverson:
@kRITZCREEK definitely think there's mileage in general structured error/warning results, whether there's a generalised solution or something more tailored as-needed

paf31:
Perhaps we just need to add a new field to the JSON, which will contain a short readable name for each replacement?

kRITZCREEK:
We'd still need to allow for multiple suggestions on a single error and editors would need to parse the module name out of the readable name to add an import.

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -76,6 +77,30 @@ checkSubsume unsolved env st userT envT = checkInEnvironment env st $ do
   -- Finally, check any constraints which were found during elaboration
   Entailment.replaceTypeClassDictionaries (isJust unsolved) expP
 
+accessorSearch
```
paf31: Looks good, but is it possible to pull out some common logic here? Also, why don't we need the instance check?

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -85,6 +110,9 @@ typeSearch
   -- ^ The typechecker state
   -> P.Type
   -- ^ The type we are looking for
-  -> Map (P.Qualified P.Ident) P.Type
+  -> ([(P.Qualified P.Ident, P.Type)], Maybe [(Label, Type)])
```
paf31: Why not return a single list of `(Expr, Type)` pairs here?

paf31: If we do return two things, maybe it could be in a record (i.e. move the record type from `Types.hs` into this module`)?

kRITZCREEK: I think there is a misunderstanding here:

`?x {x: 1, y: "hello"} :: Int` will find both `_.x` as well as `_.y`, so I don't think we should mix these results with the others, because they don't make the code type check necessarily. It might be a nice idea to actually add another check and add *just*  `_.x` to the first list.

src/Language/PureScript/TypeChecker/TypeSearch.hs
kRITZCREEK: I'd like to wait for a third occurrence before drying it up. For the instance check I don't think it comes up when applying the subsumption rule to `{| u1} -> u2`. 

It would definitely be necessary if we do the thing I outlined in the comment below and attempt to unify the accessor with the expected return type for the typed hole.

kRITZCREEK:
I just thought of an argument for using the suggestions API over the completions one. If we augment TDS to find things like `flip map` we could make suggestions that actually altered the order in which arguments are applied to the typed hole. eg:

`?map [1,2,3] (_ + 1)` would create a suggestion like: `map (_+ 1) [1,2,3]`

src/Language/PureScript/TypeChecker/TypeSearch.hs
paf31: Is there a good reason to include `_.y` there? I think we could exclude it by unifying the type of the label with the result type.

kRITZCREEK: That's why I chose to not include these results in the error message, but they are very useful in an editor setting:

`?x {x: {y: 1}} :: Int` would still show `_.x` which would then lead to `_.y`

kRITZCREEK:
How does this look? We now add all the record accessors that match the return type to the usual displayed suggestions.

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -114,5 +117,7 @@ typeSearch
 typeSearch unsolved env st type' =
   let
     resultMap = Map.mapMaybe (\(x, _, _) -> checkSubsume unsolved env st type' x $> x) (P.names env)
+    (allLabels, solvedLabels) = accessorSearch unsolved env st type'
+    solvedLabels' = first (P.Qualified Nothing . P.Ident . ("_." <>) . P.prettyPrintLabel) <$> solvedLabels
```
paf31: Let's just return `Text` instead of `Ident` here, if they're not all really `Ident`s?

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -38,11 +38,21 @@ type Context = [(Ident, Type)]
 data TypeSearch
   = TSBefore Environment
   -- ^ An Environment captured for later consumption by type directed search
-  | TSAfter [(Qualified Ident, Type)]
+  | TSAfter
+    { tsAfterIdentifiers :: [(Qualified Ident, Type)]
```
paf31: Could you please add a short `-- ^` comment to one or both of these fields?

src/Language/PureScript/TypeChecker/TypeSearch.hs
```diff
@@ -76,6 +78,32 @@ checkSubsume unsolved env st userT envT = checkInEnvironment env st $ do
...
+  rowType <- freshType
+  resultType <- freshType
+  let recordFunction = TypeApp (TypeApp tyFunction (TypeApp tyRecord rowType)) resultType
+  _ <- subsumes recordFunction userT'
+  subst <- gets TC.checkSubstitution
+  let solvedRow = fst (rowToList (substituteType subst rowType))
+  tcS <- get
+  pure (solvedRow, filter (\x -> checkAccessor tcS (substituteType subst resultType) x) solvedRow)
+  where
+    checkAccessor tcs x (_, type') = isJust (checkSubsume unsolved env tcs x type')
```
paf31: Why not use `subsumes` directly above?

paf31:
Looks good, thanks, just a few minor comments.

src/Language/PureScript/TypeChecker/TypeSearch.hs
kRITZCREEK: Sorry, not sure I understand what you are saying? Do you want me to inline checkAccessor?

`checkSubsume` is the type directed search checker which performs the Entailment checks on top of subsumes.

paf31: I see, thanks. It just seems odd to leave and reenter the state monad like this.

kRITZCREEK: I didn't find any combinator in TypeChecker.Monad that would allow me to do the checks independent of one another.

paf31: I'm just suggesting we could create a variant of `checkSubsumes` which works in `StateT ...`, which essentially amounts to extracting the call to `checkInEnvironment`.

kRITZCREEK: Oh I see. I'll try to tackle that later then

kRITZCREEK: I thought about this some more... And running the subsumption checks for the labels within the same `StateT` computation would mean they influence one another, wouldn't it? In that case I prefer to keep them apart to make sure each is checked for fitness individually.

paf31: Of, of course, that's right.

PR #2712, opened 2017-03-05
Add InvalidDerivedInstance error

### LINKED ISSUE #2595: Provide an additional reason with the 'cannot derive instance' error
Currently the following attempts to derive instances all produce a compiler error "cannot derive a type class instance for X" and nothing else:

```purescript
-- Eq has just one type argument
derive instance eqIntInt :: Eq Int Int

-- missing the wildcard type argument
derive instance genericFoo :: Data.Generic.Rep.Generic Foo

-- cannot derive this class
class Foo
derive instance foo :: Foo
```

I think it would be nice to provide a little more detail in this error message. Possible reasons the compiler might have failed to derive an instance that I can think of:

* It's not aware of the class
* The instance was given the wrong number of type arguments (in particular, I think `derive instance genericFoo :: Generic Foo` missing the second wildcard might be a common mistake)

It's also possible that we might later add more reasons to this list as the compiler is able to derive more instances.

paf31:
Fixes #2595

### DIFF
--- a/examples/failing/InvalidDerivedInstance.purs
+++ b/examples/failing/InvalidDerivedInstance.purs
@@ -1,0 +1,8 @@
+-- @shouldFailWith InvalidDerivedInstance
+module Main where
+
+import Prelude
+
+data X = X
+
+derive instance eqX :: Eq X X
--- a/examples/failing/InvalidDerivedInstance2.purs
+++ b/examples/failing/InvalidDerivedInstance2.purs
@@ -1,0 +1,6 @@
+-- @shouldFailWith ExpectedTypeConstructor
+module Main where
+
+import Prelude
+
+derive instance eqRecord :: Eq {}
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -93,7 +93,8 @@ data SimpleErrorMessage
   | UnknownClass (Qualified (ProperName 'ClassName))
   | PossiblyInfiniteInstance (Qualified (ProperName 'ClassName)) [Type]
   | CannotDerive (Qualified (ProperName 'ClassName)) [Type]
-  | InvalidDerivedInstance (Qualified (ProperName 'ClassName)) [Type]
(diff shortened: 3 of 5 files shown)


hdgarrood:
Looks good, although ideally I think this would go a little further and introduce another new type of error, to be used when the class is not known to the compiler, of the form "Don't know how to derive the type class Foo". Also I think it should be feasible to hardcode the expected number of arguments for each type class that has special compiler support for instance deriving, so that we can give better error messages (i.e. "Eq expects 1 type argument but has been given 2" rather than "the form of the derived instance is incorrect, make sure it has the correct number of type arguments").

:+1: to merge now if doing that is a lot of extra effort, but I think I would rather keep #2595 open until these things have also been addressed.

paf31:
> ideally I think this would go a little further and introduce another new type of error, to be used when the class is not known to the compiler

That is `CannotDeriveInstance`, right?

👍 for adding the argument count to the error.

hdgarrood:
There's `CannotDerive`, but it doesn't say why.

paf31:
Right, but the only reason we use that error message is because the class is not one for which we support deriving. So I think "Cannot derive `Foo ...`" is a fine error in that case, no?

hdgarrood:
It's fine, but I think it would be better to say why (especially if this change means that the only reason you'd get a `CannotDerive` is indeed that the type class is not one for which deriving is supported; that's not the case right now AFAIK).

hdgarrood:
Agh, sorry I'm talking past you a bit; I've been travelling most of today and I'm quite tired.

paf31:
No, that sounds good. I'll improve the message on my next pass at this.

paf31:
@hdgarrood How's this?

garyb:
Well I like it :smile:

"Type argument" or "Type parameter" though? I guess they're much the same.

paf31:
I tend to say "type argument", but if we use "parameter" elsewhere then we should be consistent.

paf31:
Ok, I'll merge this then, and we can fix up any remaining issues in a separate PR if need be.

hdgarrood:
Looks good to me too :)

Issue #469, opened 2014-07-08
Row Constraints


paf31:
A new type of constraint which can express row equalities between a sum and another row, a la Ermine:

For example:

```
union :: forall r1 r2 r3. (r1 | r2 ~ r3) => { | r1 } -> { | r2 } -> { | r3 }
```

I would be interested to see to what extent something like this could help with expressing interesting constraints on effect sets in `Eff`.

LiamGoodacre:
What (other) operations would we want?
* RowUnion - disjoint union: labels must be distinct across rows
* RowUnionL - non-disjoint union, left biased on overlapping labels
* RowUnionR - non-disjoint union, right biased on overlapping labels
* RowCons - add a field to a row that lacks it
* RowLacks - compute a row without a label
* RowDiff - compute a row without all the labels of another

We'd have to get kind variables in before this.  If we then implement the classes in `Prim` as, for example:
```purescript
class RowUnion (l :: # k) (r :: # k) (o :: # k) | l r -> o
class RowUnionL (l :: # k) (r :: # k) (o :: # k) | l r -> o
class RowUnionR (l :: # k) (r :: # k) (o :: # k) | l r -> o
class RowCons (s :: Symbol) (t :: k) (r :: # k) (o :: # k) | s r -> t o
class RowLacks (s :: Symbol) (r :: # k) (o :: # k) | s r -> o
class RowDiff (l :: # k) (r :: # k) (o :: # k) | l r -> o
```
We wouldn't need to depend on type families/named fundeps, and the type equality constraint.

paf31:
In Ermine, there is just the one constraint, for disjoint partitions, as I understand it. I think you can encode many of these in terms of that one.

doolse:
As an introduction for contributing to the compiler, I implemented a `*` fixed `RowUnionR`:

```purescript 
class TypeRowUnionR (l :: # *) (r :: # *) (o :: # *) | l r -> o
```
which allows for a library (lets say purescript-records) to implement `merge` (or union?):
```purescript
-- | Merge two records together with the second record overriding fields from the first
foreign import merge :: forall a b c. TypeRowUnionR a b c => {|a} -> {|b} -> {|c}
```

I can see a benefit in releasing this before the full row constraints implementation is available (unless it's right around the corner!) as it'll allow for easily creating type safe purescript wrappers around existing Javascript libraries which are based on large amounts of optional properties (React components / React Native etc..) 

Want me to polish up and make a pull request?

[...]

PR #2715, opened 2017-03-07
Record Union support

doolse:
I've created a stub purescript-records repo now and made the tests use that instead


paf31:
Fixes #469.

paf31:
This looks great!

Could you please also update `CONTRIBUTORS.md`?

I was thinking if it might be possible to have any two rows infer the third, especially now that we allow duplicate labels. I think that could be really useful.

LiamGoodacre:
@paf31 do you think it makes sense to have the accompanying module in the [typelevel-prelude](https://github.com/purescript/purescript-typelevel-prelude/)?

LiamGoodacre:
Also this is a `RowUnion`, which isn't specific to records; so I don't think it should live in a record specific library.

paf31:
That's true. I suppose it probably should go into `typelevel-prelude`, yes.

examples/passing/RecordUnionDupe.purs
```diff
@@ -0,0 +1,12 @@
...
+import Prelude
+import Data.Record (merge)
+import Control.Monad.Eff.Console (log, CONSOLE)
+import Control.Monad.Eff (Eff)
+import Unsafe.Coerce (unsafeCoerce)
+
+main :: Eff (console::CONSOLE) Unit
+main = do
+  let result = (merge {left:1} {left:"Done"}).left
+  log $ unsafeCoerce result
```
paf31: What does this return? I'd expect it to return the integer `1`, not `"Done"`.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -174,6 +169,36 @@ entails SolverOptions{..} constraint context hints =
     forClassName ctx cn@(Qualified (Just mn) _) tys = concatMap (findDicts ctx cn) (ordNub (Nothing : Just mn : map Just (mapMaybe ctorModules tys)))
     forClassName _ _ _ = internalError "forClassName: expected qualified class name"
 
+    -- Right biased union of two row types
+    unionRows :: Type -> Type -> Type -> Maybe (Type, Maybe [Constraint])
+    unionRows l r o =
+        let (ls, lEnd) = rowToList l
+            (rs, rEnd) = rowToList r
+            (os, oEnd) = rowToList o
+            lm = M.fromList ls
```
paf31: This seems like it will remove duplicates.

paf31: There is a function you can use to merge rows, called `alignRowsWith`.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -174,6 +169,36 @@ entails SolverOptions{..} constraint context hints =
...
+            rm = M.fromList rs
+            lrUnion = M.toList $ M.union rm lm
+            unioned = rowFromList (lrUnion, oEnd)
+            constraintDiff a b m = Just $ [ Constraint C.RowUnion [a, b, rowFromList (M.toList $ M.difference (M.fromList os) m, oEnd)] Nothing ]
+        in case (isClosedRow lEnd, isClosedRow rEnd) of
```
paf31: I would have expected that we could just union the known labels and then add a `RowUnion` constraint on the tails, regardless of what the tails are. We could provide an instance for `RowUnion () () ()` with the class.

Also bear in mind  that the tail could be a skolem constant, which is not being handled here. If you do it this way, you won't need any special cases.

doolse: @paf31 Yeah I didn't think what I was doing was quite correct, I'm still very much a noob as far as compiling haskell like functional languages, so it might not be correct (I didn't know if checking equality of the tail ending with == is enough for example..)

I will have another crack doing it the way you've described.. So the union is a simple matter of appending the known labels (do I need to remove exact duplicates?) The good thing is that I have a decent set of test cases now.

examples/passing/RecordUnionDupe.purs
doolse: It was returning the left side originally when the RowUnion was left biased however I think it makes more sense to be able to partially apply `merge` with the "defaultValues" and the other argument be the values which overwrite them, what do you think?

src/Language/PureScript/TypeChecker/Entailment.hs
paf31: > do I need to remove exact duplicates?

No, we should keep duplicates in the order that we saw them, left-to-right.

examples/passing/RecordUnionDupe.purs
paf31: Well if the type of `merge` is `RowUnion l r o => Record l -> Record r -> Record o`, then we really only have only possible implementation, at least if `RowUnion` means a union which allows duplicates. The meaning of a duplicate label in row `x` in `Record x` is "take the leftmost type", so the merge on records has to be left-biased.

doolse: When I say right biased i mean:

```purescript
RowUnion (same::L1, same::L2) (same::R1, same::R2) -> (same::R1,same::R2,same::L1,same::L2)
```
This is ok still isn't it? 

However, I just realised to get the desired partial application behaviour `merge` could just be changed to:
`merge :: RowUnion b a c => Record a -> Record b -> Record c`

So it's probably best to do it left biased as it's more intuitive to have: 

```purescript
RowUnion (same::L1, same::L2) (same::R1, same::R2) -> (same::L1,same::L2,same::R1,same::R2)
```

paf31: @doolse Row unions which happen by unification are always left-biased, so I think it makes sense to do the same thing with the `RowUnion` constraint. As you say, you would have to change the signature of `merge` to get the correct behavior otherwise.

doolse:
The latest commit fails on RecordUnionClosed.purs because when it tries to union the 
default onClick label with the inline onClick which hasn't been inferred yet, you get the following constraint which fails:

```purescript
RowUnion ( onClick :: Unit -> Eff ( console :: CONSOLE | eff0 ) Unit )
         ( onClick :: t1   -> Eff ( console :: CONSOLE | t2   ) Unit )
         ( onClick :: Unit -> Eff ( console :: CONSOLE | eff0 ) Unit )
```
To which `unionRows` spits out the expected instance:

```purescript
RowUnion ( onClick :: Unit -> Eff ( console :: CONSOLE | eff0 ) Unit )
         ( onClick :: t1   -> Eff ( console :: CONSOLE | t2   ) Unit )
         ( onClick :: Unit -> Eff ( console :: CONSOLE | eff0 ) Unit
         , onClick :: t1   -> Eff ( console :: CONSOLE | t2   ) Unit )
```
Which ends up failing unification:

Could not match type

```purescript
( onClick :: Unit -> Eff ( console :: CONSOLE | eff0 ) Unit
, onClick :: t1   -> Eff ( console :: CONSOLE | t2   ) Unit )
```
with type

```purescript
( onClick :: Unit -> Eff ( console :: CONSOLE | eff0 ) Unit
, onClick :: Unit -> Eff ( console :: CONSOLE | eff0 ) Unit )
```

@paf31 This should unify though shouldn't it?

paf31:
@doolse I'll try to review this over the weekend, thanks.

paf31:
@doolse I had a look at this, and came up with this implementation:

```purescript
-- | Left biased union of two row types
unionRows :: Type -> Type -> Type -> Maybe (Type, Type, Type, Maybe [Constraint])
unionRows l r _ =
    guard madeProgress $> (l, r, rowFromList (merged, tail3), cons)
  where
    (s1, tail1) = rowToSortedList l
    (s2, tail2) = rowToSortedList r

    tailVar = TypeVar "r"

    (tail3, cons) =
      case (tail1, tail2) of
        (REmpty, REmpty) -> (REmpty, Nothing)
        _ -> (tailVar, Just [ Constraint C.Union [tail1, tail2, tailVar] Nothing ])

    merged = merge s1 s2
    madeProgress = not (null merged)

    merge xs [] = xs
    merge [] xs = xs
    merge lhs@((l1, t1) : r1) rhs@((l2, t2) : r2)
      | l1 < l2 = (l1, t1) : merge r1 rhs
      | l2 < l1 = (l2, t2) : merge lhs r2
      | otherwise = (l1, t1) : (l2, t2) : merge r1 r2
```

What do you think?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -164,6 +170,31 @@ entails SolverOptions{..} constraint context hints =
...
     forClassName _ _ _ = internalError "forClassName: expected qualified class name"
 
+    -- Left biased union of two row types
+    unionRows :: Type -> Type -> Type -> Maybe (Type, Type, Type, Maybe [Constraint])
+    unionRows l r o = if solverDeferErrors || not (hasKnown l || hasKnown r) then Nothing
+      else
+        let (ls, lEnd) = rowToList l
+            (rs, rEnd) = rowToList r
+            (_, oEnd) = rowToList o
+            lrUnion = ordNub (ls <> rs)
```
paf31: The `ordNub` will remove duplicates here, which we don't want.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -316,6 +347,7 @@ entails SolverOptions{..} constraint context hints =
             -- Make a dictionary from subgoal dictionaries by applying the correct function
             mkDictionary :: Evidence -> Maybe [Expr] -> m Expr
             mkDictionary (NamedInstance n) args = return $ foldl App (Var n) (fold args)
+            mkDictionary RowUnionInstance _ = return $ Literal (ObjectLiteral [])
```
paf31: We need a placeholder for the subgoal here. We can't just ignore the arguments list, since it might not be empty. If it's non-empty, the new dictionary needs to appear here because we might have deferred the constraint.

src/Language/PureScript/TypeChecker/Entailment.hs
doolse: Thanks! I actually tried the just unify tails but ran into the same error that this currently gets which is that you can't unify:
`(a::Int, a::Int, b::Int, b::Int)` with `(a::Int, b::Int)`
All the use cases rely on the fact that this will unify, am I wrong in assuming this?

paf31: No, those shouldn't unify. A row really is a `Map Symbol [Type]` at the type level now, so those two are different.

paf31:
I have a local branch which takes a similar approach to yours. If you like, I can push it to a branch, and we can compare?

doolse:
Yeah please push your branch :+1:

paf31:
@doolse Here is what I came up with. It seems to work in the few cases I've tried https://github.com/purescript/purescript/compare/phil/row-union

doolse:
This is the killer use case for me:

```purescript
foreign import merge
  :: forall r1 r2 r3
   . Union r2 r1 r3
  => Record r1
  -> Record r2
  -> Record r3

type Mandatory r = (x :: Int | r)
type Optional = (y :: Int, z :: Int)
type Properties = Record (Mandatory Optional)

withDefaults :: forall o. Record (Mandatory o) -> Properties
withDefaults p = merge {x:1} p
```

(Note the swapping of r2 and r1 in the `merge` constraint as it's the second parameter's values which will override)

So this use case doesn't work currently with your branch and I can't see how it'll ever work without at some point removing duplicate labels?

paf31:
Why do you say it won't work?

doolse:
This is the instance which my example tries to solve.. 

```
Prim.Union ( x :: Int
             | o0      
             )         
             ( x :: Int
             )         
             ( x :: Int
             , y :: Int
             , z :: Int
             )
```
Which fails with:

```
Could not match type
              
    ( x :: Int
    , x :: Int
    | o0      
    )         
              
  with type
              
    ( y :: Int
    , z :: Int
    , x :: Int
    , x :: Int
    )         
```

doolse:
Sorry that example was slightly wrong, should be:

```purescript
withDefaults = merge {x:1,y:1,z:1}
```

Which gives the following error:
```
  Could not match type
              
    ( x :: Int
    , x :: Int
    , y :: Int
    , z :: Int
    | o0      
    )         
              
  with type
              
    ( x :: Int
    , x :: Int
    , y :: Int
    , z :: Int
    )         
              

  Prim.Union ( x :: Int
             | o0      
             )         
             ( x :: Int
             , y :: Int
             , z :: Int
             )         
             ( x :: Int
             , y :: Int
             , z :: Int
             )         
```

paf31:
I came up with

```purescript
type Mandatory r = (x :: Int | r)
type Optional r = (x :: Int, y :: Int, z :: Int | r)

withDefaults
  :: forall r
   . Record (Mandatory r)
  -> Record (Optional r)
withDefaults p = merge p { y: 1, z: 1 }
```

which sort of works, but not quite. Actually, the types are wrong, since 

```purescript
(withDefaults { x: 1, y: true }).y
```

has type `Int`, not `Boolean`, but the value is a boolean.

doolse:
I can see this use case working if we had another class which only provided instances for the first occurrence of a label, so essentially only had instances for the runtime representation of a record:

```purescript
class NormalRecord (r :: # *) (o :: # *) | r -> o 

(x::Int, x::String) (x::Int)
```

What do you think?

paf31:
This highlights a bug in my implementation actually. The compiler will happily solve

```purescript
Union (x :: Int | r) (y :: Int) (x :: Int, y :: Int | r)
```

for any row `r`. This is not correct, since `r` might be `y :: String`, giving

```purescript
Union (x :: Int, y :: String) (y :: Int) (x :: Int, y :: Int, y :: String)
```

but the order of `y` is wrong, the union should be 

```purescript
(x :: Int, y :: Int, y :: String)
```

I think I know the solution, I just need some more time to work through it than I have right now.

paf31:
@doolse Please have another look at my branch with the latest changes:

https://github.com/purescript/purescript/compare/phil/row-union

I think I've figured out how to deal with the case you described, without the bug I showed above, and without needing a `Normalize` constraint.

As an added bonus, we can now express certain lacks constraints, since

```purescript
forall a. Union r (l :: a) (l :: a | r)
```

says that `r` lacks label `l` (since if `l` appears in `r`, then it appears on the left of `l :: a` in the input, and on the right in the output, but in order to commute them, `a` must unify with whatever type `l` has in `r`, and if that is a skolem, it can't)

Issue #2714, opened 2017-03-06
Improved `if` codegen (and pattern matching in general) breaks TCO


garyb:
I'm pretty sure this is down to #2654 anyway... if you use an `if` in a function that should be tail recursive TCO no longer triggers. I guess it's not entirely down to #2654, as it implies any kind of `case` will break TCO... but, example:

``` purescript
module Main where

import Prelude
import Data.Array.Partial (head, tail)
import Partial.Unsafe (unsafePartial)

count :: forall a. (a -> Boolean) -> Array a -> Int
count p = count' 0
  where
  count' acc [] = acc
  count' acc xs =
    let h = unsafePartial head xs
    in count' (acc + if p h then 1 else 0) (unsafePartial tail xs)
```

generates:

``` javascript
var count = function (p) {
    var count$prime = function (acc) {
        return function (v) {
            if (v.length === 0) {
                return acc;
            };
            var h = Data_Array_Partial.head()(v);
            return count$prime(acc + (function () {
                var $5 = p(h);
                if ($5) {
                    return 1;
                };
                if (!$5) {
                    return 0;
                };
                throw new Error("Failed pattern match at Main line 13, column 22 - line 13, column 42: " + [ $5.constructor.name ]);
            })() | 0)(Data_Array_Partial.tail()(v));
        };
    };
    return count$prime(0);
};
```

but:

``` purescript
count :: forall a. (a -> Boolean) -> Array a -> Int
count p = count' 0
  where
  count' acc [] = acc
  count' acc xs =
    let h = unsafePartial head xs
    in count' (acc + if' (p h) 1 0) (unsafePartial tail xs)
  if' = if _ then _ else _
```

generates TCO'd code as it should:

``` javascript
var count = function (p) {
    var if$prime = function (v) {
        return function (v1) {
            return function (v2) {
                if (v) {
                    return v1;
                };
                if (!v) {
                    return v2;
                };
                throw new Error("Failed pattern match at Main line 14, column 9 - line 14, column 26: " + [ v.constructor.name ]);
            };
        };
    };
    var count$prime = function (__copy_acc) {
        return function (__copy_v) {
            var acc = __copy_acc;
            var v = __copy_v;
            tco: while (true) {
                if (v.length === 0) {
                    return acc;
                };
                var h = Data_Array_Partial.head()(v);
                var __tco_acc = acc + if$prime(p(h))(1)(0) | 0;
                var __tco_v = Data_Array_Partial.tail()(v);
[...]

PR #2719, opened 2017-03-08
TCO Fixes

### LINKED ISSUE #2689: var can capture mutable value in TCO loop
psc version: 0.10.7

In the `sumTCObug` function below, `n` is reset to zero in TCO code, even though `f'` binds `n`. The result is that `f'` adds `a` to zero. If `f'` is inlined, i.e. `go (\a -> n + a) 0` on the last line, then TCO is not triggered and the correct result is returned. If `f'` is `(+) n` then the correct result is also returned, since no JavaScript function closing over `n` is generated.

    sumTCObug = go id
        where 
      go f n = 
        if n == 0 then
          f
        else
          let
            f' a = n + a
          in
            go f' 0

    sumTCObug 7 3 -- Returns 3 instead of 10.

In the generated JavaScript code below, lines 13 and 16 make up the problematic combination.

    var sumTCObug = (function () {
        var go = function (__copy_f) {
            return function (__copy_n) {
                var f = __copy_f;
                var n = __copy_n;
                tco: while (true) {
                    var $0 = n === 0;
                    if ($0) {
                        return f;
                    };
                    if (!$0) {
                        var f$prime = function (a) {
                            return n + a | 0;
                        };
                        f = f$prime;
                        n = 0;
                        continue tco;
                    };
                    throw new Error("Failed pattern match at Main line 27, column 5 - line 35, column 1: " + [ $0.constructor.name ]);
                };
            };
        };
        return go(Control_Category.id(Control_Category.categoryFn));
    })();

paf31:
This moves values which might capture variables into a separate function which introduces new bindings for the same names. This makes TCO more generally applicable and fixes #2689 and #2714.

Here is the generated code for the example: https://gist.github.com/paf31/57cd7d3cce310cf2e9dd347cd7caccf5

@michaelficarra Would you mind having a look at the generated code as well, please?

### DIFF
--- a/src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
+++ b/src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
@@ -6,6 +6,7 @@ import Prelude.Compat
 import Data.Text (Text)
 import Data.Monoid ((<>))
 import Language.PureScript.CodeGen.JS.AST
+import Language.PureScript.AST.SourcePos (SourceSpan)
 
 -- | Eliminate tail calls
 tco :: JS -> JS
@@ -65,8 +66,6 @@ tco = everywhereOnJS convert where
       = countSelfReferences js1 == 0 && allInTailPosition body && all allInTailPosition el
     allInTailPosition (JSBlock _ body)
       = all allInTailPosition body
-    allInTailPosition (JSLabel _ _ body)
-      = allInTailPosition body
     allInTailPosition _
       = False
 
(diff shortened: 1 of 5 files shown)


src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
```diff
@@ -43,65 +48,76 @@ tco = everywhereOnJS convert where
...
-                  ++ [ JSContinue ss tcoLabel ]
+    loopify (JSReturn ss ret)
+      | isSelfCall ident ret =
+        let
+          allArgumentValues = concat $ collectArgs [] ret
+        in
+          JSBlock ss $
+            zipWith (\val arg ->
+              JSAssignment ss (JSVar ss (tcoVar arg)) val) allArgumentValues allArgs
+              ++ [ JSReturn ss (JSVar ss "undefined") ]
```
michaelficarra: You should be able to just delete this line.

paf31: Well since we're replacing an existing `return`, and that block might fall through otherwise, I think we need to leave this in.

paf31:
Thanks for the review!

src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
```diff
@@ -43,65 +48,76 @@ tco = everywhereOnJS convert where
...
+      JSBlock rootSS $
         map (\arg -> JSVariableIntroduction rootSS arg (Just (JSVar rootSS (copyVar arg)))) allArgs ++
-        [ JSLabel rootSS tcoLabel $ JSWhile rootSS (JSBooleanLiteral rootSS True) (JSBlock rootSS [ everywhereOnJS loopify js ]) ]
+        [ JSVariableIntroduction rootSS tcoDone (Just (JSBooleanLiteral rootSS False))
+        , JSVariableIntroduction rootSS tcoResult Nothing
+        ] ++
+        map (\arg ->
+          JSVariableIntroduction rootSS (tcoVar arg) Nothing) allArgs ++
+        [ JSFunction rootSS (Just tcoLoop) allArgs (JSBlock rootSS [loopify js])
+        , JSLabel rootSS tcoLabel $
```
michaelficarra: This label is no longer used.

src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
michaelficarra: Does the PureScript compiler ever generate code with early returns like that? I figured this would only ever occur at the natural end of execution anyway. But it doesn't hurt I guess. I mostly just didn't like the reference to `undefined`. `return void 0` or `return` or simply leaving it out would all be preferable.

src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
paf31: Ah good point, thanks.

src/Language/PureScript/CodeGen/JS/Optimizer/TCO.hs
paf31: Yes, I think I'll change this to a plain `return`. I don't think we generate early returns during code gen, but other optimizations can generate them.

paf31:
@michaelficarra How's this?

Once this is merged, I think I'll rename the `JS.*` modules to `CoreImp` and move `Optimizer.*` out of `CodeGen.JS`.

michaelficarra:
Not really sure why we needed to add `JSReturnNoResult`. Couldn't we have just made `JSReturn` take a `Maybe JS`?

paf31:
It was just much faster to do it this way, and I figured since we'll be doing refactoring to extract the core imperative AST soon anyway, that we can combine them at that point.

paf31:
Thanks!

PR #2720, opened 2017-03-08
Fix everythingWithScope traversal bug, fix #2718

### LINKED ISSUE #2718: A case where a shadowed variable name is not detected
``` purescript
setKey ∷ EJ.EJson → EJ.EJson → EJ.EJson → EJ.EJson
setKey k (EJ.EJson x) v = case EJ.head v of
  EJ.Map fields →
    EJ.EJson <<< roll <<< EJ.Map $ map
      (\(kv@(Tuple k v)) → if k == k then Tuple k x else kv) fields
  _ → v
```

`k` is shadowed as it's introduced in `setKey` as well as the lambda/`Tuple k v` pattern.

paf31:
Another small traversal bug.

### DIFF
--- a/examples/warning/ShadowedNameParens.purs
+++ b/examples/warning/ShadowedNameParens.purs
@@ -1,0 +1,5 @@
+-- @shouldWarnWith ShadowedName
+module Main where
+
+f :: Int -> Int -> Int
+f n = \(n) -> 1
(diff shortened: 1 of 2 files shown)


LiamGoodacre:
We seem to have lots of problems with these traversals :(
I might have a think at some point about how we can do better.

paf31:
@LiamGoodacre Thanks for the review! The problem is that this particular traversal is not structural. Perhaps if we used something like `bound` we could do better.

garyb:
Thanks for looking at this! Sorry the example wasn't too minimal, I intended to take a look myself when I got some time again.

paf31:
Sure - it took me a while to find the minimal example actually, because I wasn't thinking about the parens :)

Issue #2713, opened 2017-03-06
:reload for PSCi


noraesae:
Previous issues: #2088, #2125

Currently we have `:reset`, which clears all the imported modules. I wonder if there was a reason that it was implemented as it is, not `:reload` in which GHCi works?

I may like `:reload`, with `:module` for the clearing as they work in GHCi.

paf31:
We could just modify `:reset` instead if we could support an `:unimport` command.

noraesae:
I'm thinking of implementing this by re-importing modules in `:show import` after resetting. Is there any better method for reload?

paf31:
Sounds about right. I think we can just not clear the imports on reset.

soupi:
I think for me `reset` says "start from scratch as if you just started `psci`". May I suggest that if you plan on changing this behaviour to change the name as well? `reload` sounds good. Ideally I would prefer having both :)

paf31:
The only problem with `:reload` is that `:r` will stop working.

noraesae:
I was thinking  of changing `:reset` to `:reload`. In this way, `:r` will be a different thing. The original `:reset` can be replaced with `:unimport`, which also support a form of `:unimport A.Module`. What does it look like?

chexxor:
@soupi, What do you think about removing `:reset`, and instead advise users to simply do `:quit` and re-enter PSCi?

garyb:
How about renaming `:reset` to `:clear`, and adding `:reload`?

soupi:
@chexxor, Not having to quit gives a better user experience.  I don't think it will be very difficult to support both so IMO we should do it.

chexxor:
I hope you're successful, @noraesae!

noraesae:
I followed @garyb's suggestion as it looks like the best one.

evgenykochetkov:
Thank you so much for this!
I am just a beginner working through "PureScript by Example" book, and this change will make playing with examples much easier!

PR #2721, opened 2017-03-08
New PSCi command: reload

noraesae:
Regarding #2713 

This PR update the list of commands in PSCi. The current `:reset` command is renamed to `:clear`, and a new command `:reload` is added. `:reload` is expected to work the same as `:reload` in GHCi.

### DIFF
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -13,7 +13,7 @@ import           Prelude ()
 import           Prelude.Compat
 import           Protolude (ordNub)
 
-import           Data.List (sort, find, foldl')
+import           Data.List (sort, find, foldl', deleteBy)
 import           Data.Maybe (mapMaybe)
 import qualified Data.Map as M
 import           Data.Monoid ((<>))
@@ -217,12 +217,15 @@ handleImport
   => ImportedModule
   -> m ()
 handleImport im = do
-   st <- gets (updateImportedModules (im :))
+   st <- gets (updateImportedModules (renew im))
    let m = createTemporaryModuleForImports st
    e <- liftIO . runMake $ rebuild (map snd (psciLoadedExterns st)) m
    case e of
      Left errs -> printErrors errs
      Right _  -> put st
+   where
+   renew :: ImportedModule -> [ImportedModule] -> [ImportedModule]
+   renew m ms = m : deleteBy (\ (name, _, _) (name', _, _) -> name == name') m ms
 
 -- | Takes a value and prints its type
 handleTypeOf
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -95,7 +95,7 @@ handleCommand
   -> Command
   -> m ()
 handleCommand _ _ ShowHelp                  = liftIO $ putStrLn helpMessage
-handleCommand _ r ResetState                = handleResetState r
+handleCommand _ r ClearState                = handleClearState r
 handleCommand c _ (Expression val)          = handleExpression c val
 handleCommand _ _ (Import im)               = handleImport im
 handleCommand _ _ (Decls l)                 = handleDecls l
@@ -106,12 +106,12 @@ handleCommand _ _ (ShowInfo QueryLoaded)    = handleShowLoadedModules
 handleCommand _ _ (ShowInfo QueryImport)    = handleShowImportedModules
 handleCommand _ _ _                         = P.internalError "handleCommand: unexpected command"
 
--- | Reset the application state
-handleResetState
+-- | Clear the application state
+handleClearState
   :: (MonadReader PSCiConfig m, MonadState PSCiState m, MonadIO m)
   => m ()
   -> m ()
-handleResetState reload = do
+handleClearState reload = do
   modify $ updateImportedModules (const [])
          . updateLets (const [])
   files <- asks psciLoadedFiles
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -13,7 +13,7 @@ import           Prelude ()
 import           Prelude.Compat
 import           Protolude (ordNub)
 
-import           Data.List (sort, find, foldl', delete)
+import           Data.List (sort, find, foldl')
 import           Data.Maybe (mapMaybe)
 import qualified Data.Map as M
 import           Data.Monoid ((<>))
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/Interactive.hs
```diff
@@ -217,12 +226,15 @@ handleImport
...
 handleImport im = do
-   st <- gets (updateImportedModules (im :))
+   st <- gets (updateImportedModules (renew im))
    let m = createTemporaryModuleForImports st
    e <- liftIO . runMake $ rebuild (map snd (psciLoadedExterns st)) m
    case e of
      Left errs -> printErrors errs
      Right _  -> put st
+   where
+   renew :: ImportedModule -> [ImportedModule] -> [ImportedModule]
```
paf31: What is this function for?

paf31:
Looks great! Could you please add a test to `tests/TestPsci.hs` as well?

src/Language/PureScript/Interactive.hs
noraesae: It's to prevent import duplication. Previously, when a module was imported twice, it was listed in the imported modules twice. Maybe the name renew is somewhat confusing?

noraesae:
Will add a test.

src/Language/PureScript/Interactive.hs
paf31: > Previously, when a module was imported twice, it was listed in the imported modules twice.

Is that necessary? Maybe that's what we need to fix?

noraesae: Yes, that's what I fixed with the patch cd13ee4234203e9faed27c488ca55b7edc8e9e49 included in this PR.

Before:

```
> import Data.Functor
> import Data.Functor
> import Data.Functor
> import Data.Functor
> :show import
import Data.Functor
import Data.Functor
import Data.Functor
import Data.Functor

>
```

After:

```
> import Data.Functor
> import Data.Functor
> import Data.Functor
> import Data.Functor
> :show import
import Data.Functor

>
```

paf31: I think we want to allow multiple imports with the same name though. Otherwise you can't do something like import something qualified and with selective imports at the same time.

noraesae: Oh, right.

Then maybe we can only renew when it's really a duplication. How do you think about it?

noraesae: I've updated the PR. A comment is below.

noraesae:
Updated the renew to not only check the name, but the `ImportedModule` tuple itself. By this change, multiple imports with different forms work as expected:

```
> import Prelude
> import Prelude ((+))
> import Prelude
> import Prelude
> import Prelude as P
> import Prelude
> import Prelude ((+))
> import Prelude
> :show import
import Prelude
import Prelude ((+))
import Prelude as P

>
```

Also, I've tried to add directive tests. To make it actually work like actual PSCi, I've added a build steps, but `stack test` seems like failing.

```
Cases: 30  Tried: 11  Errors: 11  Failures: 0Compiling Data.Maybe
MultipleErrors {runMultipleErrors = [ErrorMessage [ErrorInModule (ModuleName [ProperName {runProperName = "Data"},ProperName {runProperName = "Maybe"}]),PositionedError (SourceSpan {spanName = "tests/support/bower_components/purescript-maybe/src/Data/Maybe.purs", spanStart = SourcePos {sourcePosLine = 11, sourcePosColumn = 1}, spanEnd = SourcePos {sourcePosLine = 11, sourcePosColumn = 27}}),PositionedError (SourceSpan {spanName = "tests/support/bower_components/purescript-maybe/src/Data/Maybe.purs", spanStart = SourcePos {sourcePosLine = 11, sourcePosColumn = 17}, spanEnd = SourcePos {sourcePosLine = 11, sourcePosColumn = 26}})] (UnknownImport (ModuleName [ProperName {runProperName = "Data"},ProperName {runProperName = "Eq"}]) (TyClassName (ProperName {runProperName = "Eq1"})))]}
```

I guess it's because of recent addition of `Eq1` into `Prelude`. I've tried to fix this by modifying `bower.json` in the test support dir, but no success. Any suggestion?

src/Language/PureScript/Interactive.hs
paf31: I don't understand why we need `renew` at all.

noraesae: Does it mean that exactly same imports can be allowed and be stacked in the import list? I'm afraid I do not yet understand the reason to keep the duplications.

paf31: If the user adds duplicate imports, that's fine. When the user runs the `:r` command, we should keep all of those imports, but remove the bindings.

noraesae: I understand users can add duplicate imports and PSCi works without any problem. However, according to my understanding, duplicate imports mean nothing, but they are all listed in `:show import`. I still think we do not need to keep the duplicates.

It's actually not about `:r`, but `import X` in PSCi. May I revert it from this PR and open a discussion?

paf31: > However, according to my understanding, duplicate imports mean nothing

No, they are useful:

```
> import Prelude ((+))
> import Prelude as P
```

This means that I want to use `+` globally but qualify everything else. If the user typed that, we should preserve both.

noraesae: Yes, but my update was removing only the identical ones.

```
> import Prelude ((+))
> import Prelude as P
> :show import
import Prelude ((+))
import Prelude as P

> import Prelude
> import Prelude
> import Prelude ((+))
> import Prelude as P
> :show import
import Prelude
import Prelude ((+))
import Prelude as P
```

paf31: I think if we were going to do that, we should consolidate any imports under the same namespace. So for example

```
import Prelude ((+))
import Prelude (bind)
import Prelude as P ((*))
import Prelude as P (show)
```

should give

```
> :show import
import Prelude ((+), (*))
import Prelude as P (bind, show)
```

Without that, I'm not sure it's worth it to be honest.

noraesae:
There have been some errors about compiling `Eq1` and `Ord1` but they are fixed by `bower cache clean`.

src/Language/PureScript/Interactive.hs
noraesae: Your suggestion looks better. I may upload an issue for the consolidation, and revert the `renew` from PR. Thanks again for your review 👍

noraesae:
Reverted duplication removal for PSCi imports.

PR #2723, opened 2017-03-10
Imperative core, initial refactoring

paf31:
This is not what I think the final imperative core should look like, but it does the initial refactoring to move the modules around and rename various things. I would still like to (at least)

- Separate expressions and statements
- Add an annotation type argument

but I would prefer to do these things after this PR is merged, so that it doesn't sit and get out of date while we decide what the AST should look like.

Thoughts?


michaelficarra:
Should a generic imperative core really have constructs like `typeof` and `instanceof`? Those concepts are very specific to JS, no? Does the compiler even generate them?

garyb:
I agree with @michaelficarra, and have a bunch of suggestions of my own, but if the idea is to do this incrementally then I'm fine with merging it as we have it just now. I just have two immediate points:

- I'd suggest we call the type `CoreImp` rather than `AST`, to match `CoreFn`.
- I think we might actually need a JS AST as well though, as I'm not sure `CoreImp` will be trivially printable as JS.

paf31:
Sounds good. I'll try to work on those later.

Yes, I agree, ideally the core language shouldn't use JS constructs like `typeof` directly, but hopefully we can refactor things from here into a more sensible AST with a well-defined semantics.

paf31:
@garyb It's called `Expr` in `CoreFn`. I was thinking we'd break this up into expressions and statements, which is why I chose `AST` as a generic  name for now.

I think we should wait to add a separate JS AST until we know we need it. Allocating a whole new AST might incur a performance penalty, however slight. So, since the two structures are identical right now, I'd like to see if we can do this incrementally instead.

paf31:
I've removed `typeof`, but we need `instanceof` for tag checks on ADTs.

paf31:
I've made some progress on breaking the AST into expressions and statements. I tried something a bit different this time - instead of using a separate ADT for each syntactic domain, I just modified the existing AST type by making it into a GADT. It works quite nicely and simplifies the traversals quite a bit.

michaelficarra:
@paf31 Wait, we're still using `instanceof` for ADT tag checks? I thought we were moving away from that. Something more like `value.__tag === Constructor.prototype.__tag` has to be faster.

paf31:
@michaelficarra Yes we are, but regardless of whether we use `instanceof` for JavaScript, we might still need _something_ in the AST to say "extract the tag from this value". We could just always rely on object accessors, but then we lose the ability to use faster representations in other backends.

michaelficarra:
:+1: I love it.

PR #2724, opened 2017-03-10
[psc-ide] parse imports with hanging right paren

matthewleon:
Often, when I am doing a lot of explicit imports, I use a multi-line import syntax that looks something like this:

    import X (
      A(..), doThingWithA,
      B(..), doThingWithB
    )

This is valid PS syntax, but it makes psc-ide choke when it tries to add an explicit import through the `addImport` command.

This PR fixes the issue.

It has been tested in two ways.

1. From `ghci`, behavior on `master`:

```
stack ghci --ghci-options Language.PureScript.Ide.Imports
*Language.PureScript.Ide.Imports> sliceImportSection ["module Main where", "", "import Prelude (a,",  "  b", ")", "import Meow", "", "a = b"]
Left "(line 4, column 3):\nunexpected end of input\nexpecting , or )"
```

Behavior with this change:

```
*Language.PureScript.Ide.Imports> sliceImportSection ["module Main where", "", "import Prelude (a,",  "  b", ")", "import Meow", "", "a = b"]
Right (ModuleName [ProperName {runProperName = "Main"}],["module Main where",""],[Import (ModuleName [ProperName {runProperName = "Prelude"}]) (Explicit [ValueRef (Ident "a"),ValueRef (Ident "b")]) Nothing,Import (ModuleName [ProperName {runProperName = "Meow"}]) Implicit Nothing],["","a = b"])
```

2. Ran the newly compiled `purs` command's psc-ide mode, connected to it through the `vim` plugin, and played with imports. They now work with a hanging right paren.

### DIFF
--- a/src/Language/PureScript/Ide/Imports.hs
+++ b/src/Language/PureScript/Ide/Imports.hs
@@ -136,7 +136,8 @@ step (ModuleHeader mi) (ix, l)
   | T.isPrefixOf "import " l = ImportSection ix ix
   | otherwise = ModuleHeader mi
 step (ImportSection start lastImportLine) (ix, l)
-  | any (`T.isPrefixOf` l) ["import", " "] = ImportSection start ix
+  | any (`T.isPrefixOf` l) ["import", " "]
+    || l == ")"                            = ImportSection start ix
   | T.isPrefixOf "--" l || l == ""         = ImportSection start lastImportLine
   | otherwise                              = Res start lastImportLine
 step (Res start end) _ = Res start end


matthewleon:
(Hopefully the post above serves not only as a workaround, but as a significant hint as to what stack can do to deal with the issue, as well.)

kRITZCREEK:
Thank you for taking the time and creating this PR. I'd however prefer to not accumulate more hacks on top the hack that I used for parsing in purs ide. Instead I rewired the import parsing in purs ide to use the real PureScript parser in #2725. I'd like to ask you to give that branch a try and see if it fixes your problem. Sorry for just stealing your problem and running with it, but if been thinking about making this change for a while already.

`purs ide` will still flatten out your imports after adding an explicit one, so I'm not really sure why you'd want to go through the trouble of nicely formatting them...

matthewleon:
Yes, #2725 is a much better way of addressing such issues, I was thinking along the same lines while looking at the code.

And yes, your final comment is right as well. I suppose the behavior bothered me in some way, and I was able to see in my mind's eye a bright future in which code style is maintained during these automated changes. But that is probably not really worth the effort.

PR #2725, opened 2017-03-10
[purs ide] improve import parsing

kRITZCREEK:
Still needs a few comments, but this replaces the custom parser for adding imports with the real import parser.

### DIFF
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -175,7 +175,7 @@ parseTokens :: Lexer u [PositionedToken]
 parseTokens = whitespace *> P.many parsePositionedToken <* P.skipMany parseComment <* P.eof
 
 -- | Lexes the given file, and on encountering a parse error, returns the
--- progress made up to that point, instead of returning an error.
+-- progress made up to that point, instead of returning an error
 lexLenient :: FilePath -> Text -> Either P.ParseError [PositionedToken]
 lexLenient f s = updatePositions <$> P.parse parseTokensLenient f s
 
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -174,6 +174,8 @@ updatePositions (x:xs) = x : zipWith update (x:xs) xs
 parseTokens :: Lexer u [PositionedToken]
 parseTokens = whitespace *> P.many parsePositionedToken <* P.skipMany parseComment <* P.eof
 
+-- | Lexes the given file, and on encountering a parse error, returns the
+-- progress made up to that point, instead of returning an error.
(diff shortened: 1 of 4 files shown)


paf31:
Can this reuse the existing parser code?

kRITZCREEK:
It basically does, except for me needing to extract specific position information irrelevant to the existing parser code. The `parseModuleHeader` I wrote could basically be used to speed up the building of the dependency graph, but it's not the "real" solution since it still completely lexes the file in question before starting to parse.

paf31:
But there is still some code which can be shared, right? You could move most of `parseModuleHeader` into the parser and pull the common code out of the actual module parser.

kRITZCREEK:
Like this?

paf31:
Wonderful, thanks.

kRITZCREEK:
Okay I added a few comments and tests. I'll try this out on a few test files as soon as I get the scripts set up properly... The tests are fairly comprehensive, so I'd still say this is good to go.

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -173,6 +174,12 @@ updatePositions (x:xs) = x : zipWith update (x:xs) xs
 parseTokens :: Lexer u [PositionedToken]
 parseTokens = whitespace *> P.many parsePositionedToken <* P.skipMany parseComment <* P.eof
 
+lexLenient :: FilePath -> Text -> Either P.ParseError [PositionedToken]
+lexLenient f s = updatePositions <$> P.parse parseTokensLenient f s
+
+parseTokensLenient :: Lexer u [PositionedToken]
```
paf31: Could you please add a comment here?

paf31: I assume this is here to handle cases where something wouldn't fully parse normally? If so, could you also please add a test for that?

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -174,6 +174,8 @@ updatePositions (x:xs) = x : zipWith update (x:xs) xs
 parseTokens :: Lexer u [PositionedToken]
 parseTokens = whitespace *> P.many parsePositionedToken <* P.skipMany parseComment <* P.eof
 
+-- | Lexes the given file, and on encountering a parse error, returns the
```
paf31: Thanks!

src/Language/PureScript/Parser/Lexer.hs
kRITZCREEK: Where should I put unit tests for compiler code? I can add them to purs ide's suite, which kind of steals the usual namespace for hspec test suites.

paf31: I just meant that you could maybe add a copy of an existing `psc-ide` test which has a syntax error after the imports somewhere. Does that work?

paf31:
Thanks @kRITZCREEK, looks good!

PR #2727, opened 2017-03-11
Extremely small fix for the PSCi script

mrkgnao:
I'm not sure the merge request should've been included, but my Git-fu isn't exactly the best -- is this a candidate for "cherry-picking", whatever that is?


hdgarrood:
Should this have quotes at all? Does the quoted form work with multiple arguments?

michaelficarra:
Yeah, we want the quotes. Shell quoting is a bit unintuitive.

PR #2730, opened 2017-03-12
PS 0.11 breaking changes

### LINKED ISSUE #2686: Make => into a type operator
Here's another breaking change which I'd like to consider for inclusion before 1.0.

This is motivated by adding constraint kinds. Instead of having the current syntax `(X, Y, Z) => t` for multiple constraints, `=>` would only apply to single constraints. We would write things like

```purescript
traverse
  :: forall f t a b
   . Applicative f
  => Traversable t
  => (a -> f b)
  -> t a
  -> f (t b)
```

There are a few benefits:

- `=>` becomes a regular type operator with kind `Constraint -> Type -> Type`. We could even pass it as an argument to a type function.
- Constraints could be represented using `Type`
- The parser becomes simpler and parentheses no longer require backtracking.
- It makes the implementation of constraint kinds simpler, because one of the most difficult parts to implement last time was the parser, because of the ambiguity above.

And some disadvantages:

- Dictionary application becomes curried, which entails a performance hit. Inlining should deal with this, either in the compiler later, or in Rollup by the time we get to 1.0.
- It would break a lot of code. We could address this by deprecating the current syntax in 0.11 and removing it before 1.0.

Thoughts?

### LINKED ISSUE #2701: Deprecate kind symbols and move Effect into eff
Deprecate `!` and `*` in favor of the named equivalents and move `Effect` into `eff` as a foreign kind.

garyb:
Resolves #2701 / Resolves #2686 

I just made this branch so I could do the library updates for the compiler tests - but I guess we could use it to make the changes too. Just need to consider the changes a little more carefully as I just hacked them out.

I think ideally we should print errors for the cases we've changed in the parser, like we did in the past for removed / changed syntax stuff.

### DIFF
--- a/src/Language/PureScript/Parser/Types.hs
+++ b/src/Language/PureScript/Parser/Types.hs
@@ -70,7 +70,7 @@ parseTypeAtom = indented *> P.choice
 
 parseConstrainedType :: TokenParser Type
 parseConstrainedType = do
-  constraints <- P.try (return <$> parseConstraint) <|> parens (commaSep1 parseConstraint)
+  constraints <- return <$> parseConstraint
   _ <- rfatArrow
   indented
   ty <- parseType
--- a/examples/docs/src/ConstrainedArgument.purs
+++ b/examples/docs/src/ConstrainedArgument.purs
@@ -4,6 +4,5 @@ class Foo t
 
 type WithoutArgs      = forall a. (Partial => a) -> a
 type WithArgs         = forall a. (Foo a => a) -> a
-type MultiWithoutArgs = forall a. ((Partial, Partial) => a) -> a
-type MultiWithArgs    = forall a b. ((Foo a, Foo b) => a) -> a
-
+type MultiWithoutArgs = forall a. (Partial => Partial => a) -> a
+type MultiWithArgs    = forall a b. (Foo a => Foo b => a) -> a
--- a/examples/failing/1071.purs
+++ b/examples/failing/1071.purs
@@ -4,5 +4,5 @@ module Main where
 class Foo a b where
   foo :: a -> b
 
-bar :: forall a. (Foo a) => a -> a
+bar :: forall a. Foo a => a -> a
 bar a = a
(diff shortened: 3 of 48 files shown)


garyb:
I guess we should rename `#` to `Row` too.

garyb:
For any watchers who are interested... after a discussion on IRC, `Row` is being kept as `#` for now. 

The `Type` kind is of sort `S`, as is any foreign kind that is introduced, whereas `Row` would be `S -> S`, meaning it can't be defined the normal way. Giving it the name `Row` would suggest it's less magical than it is, so the symbol is being kept (for now at least).

src/Language/PureScript/Parser/Types.hs
```diff
@@ -70,7 +70,7 @@ parseTypeAtom = indented *> P.choice
 
 parseConstrainedType :: TokenParser Type
 parseConstrainedType = do
-  constraints <- P.try (return <$> parseConstraint) <|> parens (commaSep1 parseConstraint)
+  constraints <- return <$> parseConstraint
```
paf31: I guess we should change `ConstrainedType` to only include a single `Constraint` while we're at it. This should not block anything in this PR though.

garyb: Yeah, I thought about that too. I'll clean this PR up in a little while and check I've not done anything silly.

One thing I'll relax is that currently parsing even a single constraint in parens will fail, but that seems unnecessary - you can wrap a type in parens, so why not constraints.

paf31:
Thanks for working on these!

I'd also like to try to reduce backtracking in the parser by making `=>` into part of the operator table, but we can do that after 0.11.

I'm happy to merge this and to see what people think of the new syntax using the release candidate. Any thoughts on this from anyone else before merging?

src/Language/PureScript/Parser/Types.hs
paf31: Right, and once we make it into a type operator, the thing on the left of the `=>` will also be a type, which will simplify things quite a bit.

paf31:
If there are no additional comments then I think we should merge this and take comments during the RC period.

PR #2733, opened 2017-03-12
Add "--format html" to psc-docs

hdgarrood:
Following on from #2520. This commit adds a --format html option to
psc-docs which produces HTML documentation in a given directory.

All HTML pages are self-contained (in particular, the CSS is duplicated
in every page; it's not very large though, so I don't think this is a
problem), and all links are relative, so it should be easy to just stick
these HTML files pretty much anywhere and have them 'just work'.

The CSS is copied from Pursuit's CSS, with a minor change made via
inline styles where we want to differ slightly from what Pursuit does in
the header.

We also generate an 'index.html' file with a list of all the modules in
the documentation bundle.

---

This is a WIP, because there's just one thing left to do: sort out what the CLI should look like. Ideally it would be similar to the Markdown one (which I have wanted to do something about for a long time, because the `--docgen` option's API feels like a bit of a hack to me). The issue stems from the fact that (IMO) you always want to generate HTML for everything so that links work, whereas with Markdown you might not necessarily want to do that.

Speaking of which, do we want to continue to support producing Markdown in `purs docs` after this?

### DIFF
--- a/app/Command/Docs.hs
+++ b/app/Command/Docs.hs
@@ -1,6 +1,4 @@
 {-# LANGUAGE TupleSections #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE RecordWildCards #-}
 
 module Command.Docs (command, infoModList) where
 
@@ -260,4 +258,3 @@ examples =
     , "      --docgen Data.List:docs/Data.List.md \\"
     , "      --docgen Data.List.Lazy:docs/Data.List.Lazy.md"
     ]
-
--- a/app/Command/Docs.hs
+++ b/app/Command/Docs.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 
 module Command.Docs (command, infoModList) where
 
@@ -6,6 +8,7 @@ import           Protolude (ordNub)
 
 import           Command.Docs.Etags
 import           Command.Docs.Ctags
+import           Command.Docs.Html
 import           Control.Applicative
 import           Control.Arrow (first, second)
 import           Control.Category ((>>>))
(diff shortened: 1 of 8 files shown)


hdgarrood:
I've uploaded example output at http://harry.garrood.me/purs-html-docs-example, which I created from running `purs docs` on `purescript-node-irc`.

paf31:
This looks great! Thanks for taking the time to work on it.

I've was thinking that we could possibly get rid of Markdown output now that relatively few repos use it, and GitHub supports serving HTML from `docs/` anyway. That said, maybe it's not that much maintenance work to keep both.

This will definitely be very useful for building docs for offline use, but I think it would be nice to support building docs for a single repo as well, which means we might want to have the ability to only include certain pages on the home page (similar to how we only generate certain Markdown pages). I'd be happy to have the default be to generate every page though, which would probably simplify the command line a bit. What do you think?

hdgarrood:
If we were building docs for a single repo, would we omit links for types which come from dependencies?

hdgarrood:
As in, display the type in the HTML, but have it just not be a link?

paf31:
I think we should generate the files and the links, just not show those modules on the main page.

hdgarrood:
Cool - do you think it would make sense to do that by adding a CLI option for modules that shouldn't be listed, which takes a glob as an argument, so you might use it like this?

```
purs docs --format html "src/**/*.purs" --unlisted "bower_components/purescript-*/src/**/*.purs"
```

paf31:
I don't mind. I imagined it would take a whitelist of module names actually, but whatever is simplest.

hdgarrood:
Ah, that hadn't occurred to me. A whitelist of module names would work too, and that would be marginally easier from an implementation perspective inside `purs`, but I think I'd rather not because it could be quite awkward for tools like pulp, or even just people, to generate a whitelist of module names based on modules in `src/`. I think the most common situation is that people will want to include/exclude modules based on what directory they're in..? We could also add a module whitelist option later if people want it.

paf31:
Good point. Either way works for me - I'll be using it from Pulp anyway :)

hdgarrood:
Do you have a particular use case in mind? At the moment, within any particular module, all modules are listed in the column on the right. I'm struggling to imagine what should go in that column in the case where we're not listing all modules on the index page; the obvious answer is the same list as the one on the index page, but I think that might be a bit odd if you've already clicked through a link into a module from a dependency which isn't part of the main set..?

paf31:
The two cases I'm thinking about are

- Producing formatted docs for offline reading, and
- Producing docs for inclusion in a repo

> but I think that might be a bit odd if you've already clicked through a link into a module from a dependency which isn't part of the main set..?

I agree, but another option is to get rid of the list and just provide a link back to the index.

We could even separate the home page and the index page.

I should add that this shouldn't block you from merging this.

hdgarrood:
Ok cool; I'd like to revisit this later too, but for now, I think I will just add an option for the output directory and leave it at that.

PR #2735, opened 2017-03-12
Add back .psci file support

### LINKED ISSUE #2734: Read commands from .psci file when starting session
In the past, PSCi would check for a ".psci" file in the project directory or `~` directory, then execute the commands in their when starting a PSCi session. This was removed several months ago (I can't find the issue/PR), but it would be nice to have that back. One motivation for it is to have `pulp init` create a ".psci" file having "import Prelude". With this, newcomers can do `pulp psci` followed by `1+1` and see a success case, rather than `Unknown operator (+)` error.

paf31:
Fixes #2734

It looks like this:

```
$ echo "import Prelude" > .psci
$ psc-package repl
PSCi, version 0.11.0
Type :? for help

import Prelude
> 1 + 1
2
```

cc @chexxor

### DIFF
--- a/app/Command/REPL.hs
+++ b/app/Command/REPL.hs
@@ -363,7 +363,7 @@ command = loop <$> options
                           case parseCommand l of
                             Left err -> liftIO (putStrLn err >> exitFailure)
                             Right cmd@Import{} -> handleCommand' state cmd
-                            Right _ -> liftIO (putStrLn "Only import declarations are supported in the .psci file" >> exitFailure)
+                            Right _ -> liftIO (putStrLn "The .psci file only supports import declarations")
 
                     handleCommandWithInterrupts
                       :: state
--- a/app/Command/REPL.hs
+++ b/app/Command/REPL.hs
@@ -30,6 +30,7 @@ import           Control.Monad.Trans.Except (ExceptT(..), runExceptT)
 import           Control.Monad.Trans.State.Strict (StateT, evalStateT)
 import           Control.Monad.Trans.Reader (ReaderT, runReaderT)
 import           Data.FileEmbed (embedStringFile)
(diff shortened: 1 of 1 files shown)


hdgarrood:
Does this allow you to put basically anything inside .psci? Not that I feel particularly strongly that we should be restricting it, but it might be worth discussing.

paf31:
Yes it does right now. I think it might be better to limit it to imports only. There is also a security risk associated with running arbitrary code on behalf of the user.

paf31:
Fixed.

app/Command/REPL.hs
```diff
@@ -347,14 +352,28 @@ command = loop <$> options
...
+                      configFile <- (</> ".psci") <$> liftIO getCurrentDirectory
+                      exists <- liftIO $ doesFileExist configFile
+                      when exists $ do
+                        ls <- lines <$> liftIO (readUTF8File configFile)
+                        for_ ls $ \l -> do
+                          liftIO (putStrLn l)
+                          case parseCommand l of
+                            Left err -> liftIO (putStrLn err >> exitFailure)
+                            Right cmd@Import{} -> handleCommand' state cmd
+                            Right _ -> liftIO (putStrLn "Only import declarations are supported in the .psci file" >> exitFailure)
```
hdgarrood: Might be nicer to print a warning, ignore declarations which aren't imports, and continue?

paf31: Sounds good

hdgarrood:
:+1:

kRITZCREEK:
`.psci` seems inconsistent after `purs repl` call it `.purs-repl`?

paf31:
Yeah, that makes sense. It is called PSCi elsewhere though.

Issue #2732, opened 2017-03-12
Errors about `discard` could be improved


garyb:
The error about `discard` not being defined is a bit weird, it mentions `do` notation and asks you to import `bind` from `Prelude` to resolve it.

Also, this is a side thing, but I think maybe `bind_` would have been a better name, to go with the other discardifying things we have like `for_`, `traverse_`, `sequence_`, etc. When I started getting errors about `discard` I found it confusing as I knew I was using `void $ ...` or `_ <- ...`, so thought it should still be `bind`-y... the name `discard` seemed to suggest ignoring something entirely, not running its effects and discarding the result (yeah, I know I even used the word "discard" in that description :smile:, but still...).

paf31:
Agreed, these are not ideal. The position information is also not as helpful as it could be. The errors are reported at the function level.

paf31:
Closing this, and we can take care of renaming `discard` separately if need be.

PR #2736, opened 2017-03-12
Improve errors for bind/discard 

paf31:
For #2732

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -494,7 +494,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs) e = flip evalS
     renderSimpleErrorMessage (RedefinedIdent name) =
       line $ "The value " <> markCode (showIdent name) <> " has been defined multiple times"
     renderSimpleErrorMessage (UnknownName name@(Qualified Nothing (IdentName (Ident i)))) | i `elem` [ C.bind, C.discard ] =
-      line $ "Unknown " <> printName name <> ". You're probably using do-notation, which the compiler replaces with calls to the " <> markCode "bind" <> " function. Please import " <> markCode "bind" <> " from module " <> markCode "Prelude"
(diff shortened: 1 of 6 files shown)


garyb:
Any thoughts on switching to `bind_` rather than `discard`?

paf31:
I guess it would need a (technically) breaking change in Prelude, but I don't mind, we can.

garyb:
Oh! Now you mention it, in the `ps-0.11` branch I made for the `Prelude`, I made another breaking change anyway - by removing the `EuclideanRing` and `Field` instances for `unit`. I doubt they're used much, if anywhere, but I would have completely forgotten about that if you hadn't said.

I'm fine with fixing everything again for that. Since everything from `-eff` down is "broken" anyway it's not much of a difference to update for `-prelude` too.

paf31:
I was assuming we'd need to make some major releases anyway, so sure, let's change it.

paf31:
This now includes fixes to the type checker (error hints were backwards when solving instances - this should improve other things too) and the parser to generally improve error quality in `do` blocks.

src/Language/PureScript/Errors.hs
```diff
@@ -1120,7 +1120,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs) e = flip evalS
 
   -- | Simplify an error message
   simplifyErrorMessage :: ErrorMessage -> ErrorMessage
-  simplifyErrorMessage (ErrorMessage hints simple) = ErrorMessage (simplifyHints hints) simple
+  simplifyErrorMessage (ErrorMessage hints simple) = ErrorMessage (traceShow (filter ((PositionHint == ) . hintCategory) hints) $ simplifyHints hints) simple
```
garyb: `traceShow`? :smile:

paf31: Oops :)

paf31:
@garyb If this looks ok, I'd like to merge it and then fix the naming issue separately. Sound good?

PR #2746, opened 2017-03-15
ConstrainedType only takes a single Constraint

paf31:
Just a bit of tidying


paf31:
Thank you both for reviewing!

Issue #2664, opened 2017-02-15
Test for PSCi evaluation.


noraesae:
It would be nice to have test structure for PSCi evaluation.

I'm not sure which data type to use for defining the tests, but maybe something like below.

```haskell
[ ("let f _ = 10 in f 0", shouldReturn 10)
, ("[a, b] = [1, 2]",     shouldFail) -- hope to see though!
]
```

It's from #2630.

paf31:
Do you mean in the test suite?

noraesae:
It could be in the test suite. I imagined something similar to `completionTestData`.

noraesae:
I think now it became better to tackle this issue, because #2721 allow PSCi test env to actually run commands in test suites. I will do some experiments of ExprCommand evaluations.

PR #2747, opened 2017-03-15
Implement evaluation test for PSCi

noraesae:
*Regarding #2664*

We can write evaluation tests like below:

```haskell
do
  run "import Prelude"
  run "a = 10"
  "a + 100" @?=> "110"
```

It will evaluate both left and right commands and compare the result.

The operator `@?=>` is from `@?=` of HUnit. I thought the operator looked straightforward, but please let me know if there is one more appropriate.

### DIFF
--- a/tests/TestPsci/CommandTest.hs
+++ b/tests/TestPsci/CommandTest.hs
@@ -15,21 +15,21 @@ commandTests = TestLabel "commandTests" $ TestList $ map (TestCase . runTestPSCi
       run "import Data.Functor"
       run "import Control.Monad"
       before <- psciImportedModules <$> get
-      length before @?== 3
+      length before `equalsTo` 3
       run ":clear"
       after <- psciImportedModules <$> get
-      length after @?== 0
+      length after `equalsTo` 0
   , do
       run "import Prelude"
       run "import Data.Functor"
       run "import Control.Monad"
       before <- psciImportedModules <$> get
-      length before @?== 3
+      length before `equalsTo` 3
       run ":reload"
       after <- psciImportedModules <$> get
-      length after @?== 3
+      length after `equalsTo` 3
   , do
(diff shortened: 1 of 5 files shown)


hdgarrood:
Can you add some more comments to explain what these new functions do? I'm struggling a bit to see what is going on where. Also, do you think it's necessary to treat both arguments of `@?=>` as repl commands to be evaluated, or would it be a little simpler to interpret the "expected" argument as just a string i.e. what the "actual" argument should end up evaluating to? Finally I'm not very keen on this operator, as I think there's a risk of people thinking it comes from HUnit. I think I'd prefer a normal function, perhaps ``"a + 100" `evaluatesTo` "110"`` or something.

noraesae:
I will add some comments soon.

About the expected part and the operator, I think that just `assert` may be enough. What do you think?

```
do
  run "import Prelude"
  run "a = 10"
  assert "a + 100 == 110"
```

With this, we can just check if the result becomes `"true\n"`, and also it enables assertions for data without a Show instance.

hdgarrood:
I think it makes more sense to assert that some particular string has been printed. Unless I'm missing something I think it's possible to implement that `assert` in terms of the current `@?=>`, so I don't see how it could be more flexible.

noraesae:
Yes, I realized that with the assert we cannot make use of HUnit output. I will just rename the op to 'evaluatesTo' and make it only evaluate the left part as you suggested :)

noraesae:
Updated the PR.

Fix test, rename ops, don't evaluate the `expected` part, and comments and renaming.

noraesae:
Hmm, coudn't stop Travis CI mac build from failing for timeout.

paf31:
@hdgarrood Does this look okay now?

paf31:
Thanks!

PR #2750, opened 2017-03-16
Normalised row primitive

doolse:
This adds a Prim class for computing a "normalised" row (please suggest a better name).
It will only find an instance if all the types for a label are the same, so:

```
(a :: Int, a :: Int, a :: Int, b :: Int) -> (a::Int, b::Int)
(a :: Int, a :: String, b :: Int) -> No instance
```

This primitive combined with the duplicate label `Union` allows the use cases I was concerned about to work properly.

One thing which I couldn't work out though, was that the `Constraint`s were being ignored if the instances were found during the "defered" pass. I thought this was fixed by having a proper implementation for `EmptyInstance` in `mkDictionary`? Perhaps it's still not right..

### DIFF
--- a/examples/failing/RecordUnionExtra.purs
+++ b/examples/failing/RecordUnionExtra.purs
@@ -1,0 +1,13 @@
+-- @shouldFailWith TypesDoNotUnify
+module Main where
+
+import Prelude
+import Data.Record (merge)
+
+type Mand r = (title :: String, flag :: Boolean | r)
+type Opt = (optional :: Int)
+
+withDefaults :: forall o. Normalised (Mand (optional :: Int|o)) (Mand Opt) => {|Mand o} -> {|Mand Opt}
+withDefaults = merge {optional:1}
+
+main = withDefaults {title:"Title", flag: true, others:""}
--- a/examples/passing/RecordUnionClosed.purs
+++ b/examples/passing/RecordUnionClosed.purs
@@ -1,0 +1,21 @@
+module Main where
+
+import Prelude
+import Data.Record (merge)
+import Control.Monad.Eff.Console (log, CONSOLE)
+import Control.Monad.Eff (Eff)
+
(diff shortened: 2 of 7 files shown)


paf31:
Thanks for putting this together, but I'd like to make sure that we choose a set of minimal primitives for working with rows in the type system, so I'd like to think about this some more. Could you please write out what an example generated instance looks like? How do you deal with possible overlaps between the known labels and the tail?

doolse:
I had another think about this and realized that I can still encode all my uses cases if we had a primitive which simply picked the first (left most) type for the label, which makes reasoning about it even easier:

Closed rows are trivial:
```purescript
Normalised (a :: T1, a :: T2, b :: T2, b :: T1) (a :: T1, b :: T2)
```

Open rows aren't much harder because the known labels always win out:
```purescript
   Union (a :: T1, b :: T2) o u 
=> Normalised u n 
=> Normalised (a :: T1, a :: T2, b :: T2, b :: T1 | o) n
```

So for example where `o` has types for both a b as well `(a :: T3, b :: T3, c :: T3)`:
```purescript
   Union (a :: T1, b :: T2) (a :: T3, b :: T3, c :: T3) (a :: T1, a :: T3, b :: T2, b :: T3, c :: T3) 
=> Normalised (a :: T1, a :: T3, b :: T2, b :: T3, c :: T3) (a :: T1, b :: T2, c :: T3) 
=> Normalised (a :: T1, a :: T2, b :: T2, b :: T1 | (a :: T3, b :: T3, c :: T3)) (a :: T1, b :: T2, c :: T3)
```

I don't think I'm missing anything here?

I can achieve the same goal of making sure that the types of both sides of the merge are the same by ensuring that the union of left + right and right + left are normalised to the same thing:

```purescript
   Union l r u1
=> Union r l u2
=> Normalised u1 n
=> Normalised u2 n
```

PR #2751, opened 2017-03-16
[purs ide] Track where a Reexport was resolved from

kRITZCREEK:
This is in preparation to group reexports of the same value in the completion
results


kRITZCREEK:
Thanks for reviewing, missed this one.

PR #2760, opened 2017-03-19
Add useful errors for removed kind syntax

garyb:


src/Language/PureScript/Parser/Common.hs
```diff
@@ -143,3 +143,8 @@ withSourceSpan f p = do
         _ -> Nothing
   let sp = SourceSpan (P.sourceName start) (toSourcePos start) (toSourcePos $ fromMaybe end end')
   return $ f sp comments x
+
+featureWasRemoved :: String -> P.Parsec s ParseState a
+featureWasRemoved err = do
+  pos <- P.getPosition
+  error $ "It looks like you are trying to use a feature from a previous version of the compiler:\n" ++ err ++ "\nat " ++ show pos
```
paf31: `parserFail` instead of `error`, perhaps? That should include the source position info for you as well.

garyb: This function was copied exactly from the last time we deprecated syntax 😆, but sure.

paf31: Oops :) Yeah, I mean it doesn't really matter that much, but we can hopefully fail with a real error message this way.

paf31:
Apparently `RebuildSpecWithDeps` is failing in the pscide test suite though.

garyb:
I'll try the `parserFail` first anyway, that does seem better than using `error`.

> Apparently `RebuildSpecWithDeps` is failing in the pscide test suite though.

I'm not sure what's going on there, it passed in AppVeyor and all of the Travis build configs but one.

garyb:
Alright, just switched to using `parserFail` directly now - the errors are explanatory enough without the common header bit I think. Let's see what happens with the tests this time :wink:

src/Language/PureScript/Parser/Kinds.hs
```diff
@@ -12,14 +12,25 @@ import Language.PureScript.Parser.Lexer
 import qualified Text.Parsec as P
 import qualified Text.Parsec.Expr as P
 
+parseStar :: TokenParser Kind
+parseStar = symbol' "*" *>
+  P.parserFail "The `*` symbol is no longer used for type kinds.\n  The new equivalent is the named kind `Type`."
```
garyb: Would "type kinds" be better as "the kind of types" perhaps? (Same for effects below)

paf31: Yeah, I do prefer that actually.

PR #2761, opened 2017-03-19
Row union constraints

### LINKED ISSUE #469: Row Constraints

paf31:
This is based on the idea in #2715 and resolves #469, but takes a slightly different approach to the solution in #2715 and #2750.

It adds a single new constraint to `Prim` called `Union`, which acts like a left-biased union on rows. It is generated by instances of the form

```purescript
instance unionClosed :: Union (l_1 :: a_1, ..., l_n :: a_n) r (l_1 :: a_1, ..., l_n :: a_n | r)
```

and

```purescript
instance unionOpen :: Union r1 r2 r3 => Union (l_1 :: a_1, ..., l_n :: a_n | r1) r2 (l_1 :: a_1, ..., l_n :: a_n | r3)
```

Notice that the order of the types for each label is preserved, and we correctly handle duplicates. 

This means that you need to do some work when working with possibly-open rows, but it has the happy consequence that we can express certain types of lacks constraints. For example, the constraint

```purescript
Union r (l :: a) (l :: a | r)
```

means that if `l` appears in `r` at all, then it can only appear associated with the type `a`, since we need to be able to append `l :: a` on the right and have that be equivalent to prepending `l :: a` on the left.

cc @doolse Can you please have a look and tell me if this solves the problem you had? Thanks.

### DIFF
--- a/src/Language/PureScript/Environment.hs
+++ b/src/Language/PureScript/Environment.hs
@@ -358,6 +358,7 @@ primTypes =
     , (primName "Int",        (kindType, ExternData))
     , (primName "Boolean",    (kindType, ExternData))
     , (primName "Partial",    (kindType, ExternData))
+    , (primName "Union",      (FunKind (Row kindType) (FunKind (Row kindType) (FunKind (Row kindType) kindType)), ExternData))
     , (primName "Fail",       (FunKind kindSymbol kindType, ExternData))
     , (primName "Warn",       (FunKind kindSymbol kindType, ExternData))
     , (primName "TypeString", (FunKind kindType kindSymbol, ExternData))
@@ -376,6 +377,12 @@ primClasses =
     [ (primName "Partial", (makeTypeClassData [] [] [] []))
     , (primName "Fail",    (makeTypeClassData [("message", Just kindSymbol)] [] [] []))
     , (primName "Warn",    (makeTypeClassData [("message", Just kindSymbol)] [] [] []))
+    , (primName "Union",   (makeTypeClassData
+                             [ ("l", Just (Row kindType))
+                             , ("r", Just (Row kindType))
+                             , ("u", Just (Row kindType))
+                             ] [] []
+                             [ FunctionalDependency [0, 1] [2] ]))
     ]
 
 -- |
--- a/src/Language/PureScript/Environment.hs
+++ b/src/Language/PureScript/Environment.hs
@@ -382,7 +382,10 @@ primClasses =
                              , ("r", Just (Row kindType))
                              , ("u", Just (Row kindType))
                              ] [] []
-                             [ FunctionalDependency [0, 1] [2] ]))
+                             [ FunctionalDependency [0, 1] [2]
+                             , FunctionalDependency [1, 2] [0]
+                             , FunctionalDependency [2, 0] [1]
+                             ]))
     ]
 
 -- |
--- a/src/Language/PureScript/Environment.hs
+++ b/src/Language/PureScript/Environment.hs
@@ -152,9 +152,9 @@ makeTypeClassData args m s deps = TypeClassData args m s deps determinedArgs cov
 data NameVisibility
   = Undefined
   -- ^ The name is defined in the current binding group, but is not visible
-  | Defined deriving (Show, Eq)
+  | Defined
   -- ^ The name is defined in the another binding group, or has been made visible by a function binder
-
+  deriving (Show, Eq)
 -- | A flag for whether a name is for an private or public value - only public values will be
 -- included in a generated externs file.
 data NameKind
--- a/src/Language/PureScript/Environment.hs
+++ b/src/Language/PureScript/Environment.hs
@@ -72,14 +72,11 @@ data FunctionalDependency = FunctionalDependency
   -- ^ the determined type arguments
   } deriving Show
 
--- |
--- The initial environment with no values and only the default javascript types defined
---
+-- | The initial environment with no values and only the default javascript types defined
 initEnvironment :: Environment
 initEnvironment = Environment M.empty primTypes M.empty M.empty M.empty primClasses primKinds
 
--- |
--- A constructor for TypeClassData that computes which type class arguments are fully determined
+-- | A constructor for TypeClassData that computes which type class arguments are fully determined
 -- and argument covering sets.
 -- Fully determined means that this argument cannot be used when selecting a type class instance.
 -- A covering set is a minimal collection of arguments that can be used to find an instance and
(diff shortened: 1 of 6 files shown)


doolse:
The only problem I can see with this is that you can't end up with a closed union unless both of your row's are already closed. So currently `withDefaults`

```purescript
type Mandatory r = (x :: Int | r)
type Optional r = (x :: Int, y :: Int, z :: Int | r)

withDefaults
  :: forall r
   . Union r (y :: Int, z :: Int) (y :: Int, z :: Int | r)
  => Record (Mandatory r)
  -> Record (Optional r)
withDefaults p = merge p { y: 1, z: 1 }
```

Is going to allow you call it with any arbitrary extra labels. The killer use case for me is to be able to close the row so you can't accidentally add non-existent fields. 

```purescript
withDefaults
  :: forall r
   . ???
  => Record (Mandatory r)
  -> Record (Optional ())
withDefaults p = merge p { y: 1, z: 1 }
```

I tried in vain to come up with constraints to make this work but couldn't wrap my head around how it could be possible without some way of normalising the row?

paf31:
@doolse What about this?

```purescript
withDefaults
  :: forall r s t
   . Union r (y :: Int, z :: Int) (y :: Int, z :: Int | s)
  => Union s t (y :: Int, z :: Int) -- s is a subrow of y, z
  => Record (Mandatory r)
  -> Record (Optional s)
withDefaults p = merge p { y: 1, z: 1 }
```

doolse:
That doesn't compile because `t` isn't mentioned in the body of the type

paf31:
Sorry, I have changes on my branch locally. I'll push them in a bit, but the example does work :)

paf31:
@doolse I've pushed those changes, and simplified things as well.

doolse:
Yep this works perfectly now :+1: 

The only problem i'm having now is making higher level type classes on top of `Union`, for example I can define:

```purescript
class RowSubset (r :: # Type) (or :: # Type -> # Type)

instance rowSubset :: (Union r (or ()) (or s), Union s t (or ())) => RowSubset r or
```

but I can't work out how I can add a constraint with it because I can't see a way to provide a `# Type -> # Type` type as you can't partially apply type synonyms or row type literals?

paf31:
I'd considered something like `(x :: a, y :: b | _)` recently, meaning the type constructor of kind `# Type -> # Type`. For now, I think you're reaching the limits of what the type system can express :)

What is `RowSubset` meant to represent, in words?

paf31:
@kRITZCREEK or @LiamGoodacre Could I get a second review if one of you has time please?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -334,6 +345,28 @@ entails SolverOptions{..} constraint context hints =
...
           App (Accessor (mkString (superclassName className index)) dict) valUndefined
 
+    -- | Left biased union of two row types
+    unionRows :: Type -> Type -> Type -> Maybe (Type, Type, Type, Maybe [Constraint])
+    unionRows l r _ =
+        guard canMakeProgress $> (l, r, rowFromList out, cons)
+      where
+        (fixed, rest) = rowToList l
+
+        rowVar = TypeVar "r"
```
LiamGoodacre: Could `r` conflict anywhere? or is this safe?

paf31: I believe this is safe. This is like the user introducing a `TypeVar` in an instance declaration. By the time we get to the solver, everything else will be skolemized or instantiated.

doolse:
Ignore my previous problem, what I was really trying to 
represent was: 

```purescript 
class RowSubset (r :: # Type) (s :: # Type)
```
In plain english, that r is a subset of s. Which I eventually solved with:
```purescript
instance rowSubset :: (Union r s st, Union s t st, Union t u s) => RowSubset r s
```

I've put what I came up with so far into `purescript-records` - https://github.com/doolse/purescript-records/tree/ps-0.11

@paf31 @LiamGoodacre @kRITZCREEK perhaps someone could have a look and see if it makes sense? 

In particular I wonder if it's possible to not need the `OptionalDefaults m` phantom type which I'm using because I couldn't think of a way to extract the mandatory part from just the `mr` record alone.

paf31:
I think if we added the other fundeps, we could just define

```purescript
instance rowSubset :: Union r t s => RowSubset r s
```

(`r` is a subset if there is some other row which unions to give `s`)

This instance should be allowed if we add the fundep `r s -> t`, which is reasonable given the instances and meaning for `Union`.

paf31:
@doolse I've added `Subrow` to the example.

I'll merge this soon if there are no objections.

src/Language/PureScript/Environment.hs
```diff
@@ -376,6 +377,15 @@ primClasses =
     [ (primName "Partial", (makeTypeClassData [] [] [] []))
     , (primName "Fail",    (makeTypeClassData [("message", Just kindSymbol)] [] [] []))
     , (primName "Warn",    (makeTypeClassData [("message", Just kindSymbol)] [] [] []))
+    , (primName "Union",   (makeTypeClassData
+                             [ ("l", Just (Row kindType))
+                             , ("r", Just (Row kindType))
+                             , ("u", Just (Row kindType))
+                             ] [] []
+                             [ FunctionalDependency [0, 1] [2]
```
kRITZCREEK: Maybe add a comment here along the lines of: `l -> r, r -> u, u -> l`

doolse:
:+1: I can now express the `mergeDefaults` functionality without the need for the phantom type. I'm excited!

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -315,6 +322,10 @@ entails SolverOptions{..} constraint context hints =
             -- Make a dictionary from subgoal dictionaries by applying the correct function
             mkDictionary :: Evidence -> Maybe [Expr] -> m Expr
             mkDictionary (NamedInstance n) args = return $ foldl App (Var n) (fold args)
+            mkDictionary UnionInstance (Just [e]) =
+              -- We need the subgoal dictionary to appear in the term somewhere
+              return $ App (Abs (VarBinder (Ident C.__unused)) (Literal (ObjectLiteral []))) e
+            mkDictionary UnionInstance _ = return $ Literal (ObjectLiteral [])
```
garyb: I think we should be using `C.undefined` values for all of these; like `C.__unused` it gets removed during codegen... we introduced it to remove the `ObjectLiteral []` applications we used to have everywhere, as they add up to be a significant performance penalty.

paf31: How's this?

src/Language/PureScript/Environment.hs
paf31: Done, thanks!

src/Language/PureScript/TypeChecker/Entailment.hs
garyb: Looks good 👍

paf31:
Ok I'll merge this once Travis finishes then, and we can include it in 0.11.

PR #2762, opened 2017-03-20
Add hint for constraint syntax change

garyb:
I meant to include this in the other PR, but forgot.

### DIFF
--- a/src/Language/PureScript/Parser/Types.hs
+++ b/src/Language/PureScript/Parser/Types.hs
@@ -70,7 +70,7 @@ parseTypeAtom = indented *> P.choice
 
 parseConstrainedType :: TokenParser Type
 parseConstrainedType = do
-  constraint <- parens parseConstraint <|> parseConstraint
+  constraint <- parseConstraint <|> parens (parseConstraint <* oldComma)
   _ <- rfatArrow
   indented
   ty <- parseType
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Parser/Types.hs
```diff
@@ -81,9 +81,11 @@ parseConstrainedType = do
     indented
     ty <- P.many parseTypeAtom
     return (Constraint className ty Nothing)
+  oldComma = comma *>
+    P.parserFail "Class constraints in type annotations can no longer be grouped in parentheses.\n  Each constraint should now be separated by `=>`, for example:\n    `(Applicative f, Semigroup a) => a -> f a -> f a`\n  would now be written as:\n    `Applicative f => Seimgroup a => a -> f a -> f a`."
 
 parseAnyType :: TokenParser Type
-parseAnyType = P.buildExpressionParser operators (buildPostfixParser postfixTable (P.try parseConstrainedType <|> parseTypeAtom)) P.<?> "type"
+parseAnyType = P.buildExpressionParser operators (buildPostfixParser postfixTable (P.try parseConstrainedType <|> P.try parseTypeAtom <|> parseConstrainedType)) P.<?> "type"
```
garyb: This is gross, but aside from using `error` or something, I'm not sure how else to get the new message to appear.

kRITZCREEK: Mayb you could use the `lookAhead` trick I used [here](https://github.com/purescript/purescript/blob/23fa5dba801e420a0b91a4e46ce15524328e9c30/src/Language/PureScript/Parser/Lexer.hs#L277-L286)? Not sure if that makes things any nicer though.

PR #2766, opened 2017-03-23
Use git show to get tag date

### LINKED ISSUE #2753: Can't publish anymore
It worked before I updated to the latest psc.

```
$ pulp publish
Internal error: this is probably a bug. Please report it:
  https://github.com/purescript/purescript/issues/new    
                                                         
Details:                                                 
Unable to parse the date for a git tag: v0.0.21          
                                                         
```

I looked at the source and found that it invokes this command:
```
$ git tag -l v0.0.21 --format='%(taggerdate:unix)'
error: unknown option `format=%(taggerdate:unix)'
```

paf31:
Fixes #2753 for me (I was having the same problem while testing `purs publish`)

### DIFF
--- a/src/Language/PureScript/Publish.hs
+++ b/src/Language/PureScript/Publish.hs
@@ -196,7 +196,7 @@ getVersionFromGitTag = do
 -- | Given a git tag, get the time it was created.
 getTagTime :: Text -> PrepareM UTCTime
 getTagTime tag = do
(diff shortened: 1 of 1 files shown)


hdgarrood:
Looks good, although I think it would also be good to find out what the earliest version of git which supports this is, and then check that we have at least that version just before running this command.

paf31:
Will reopen after rebasing

PR #2778, opened 2017-03-26
Use GHC 8 with stack by default

### LINKED ISSUE #2777: Suggestion: use stack-ghc-8.0.yaml as default
I suggest using `stack-ghc-8.0.yaml` as default, and rename the current `stack.yaml` as `stack-ghc-7.10.yaml`. The personal reason is that GHC 7.10 no longer works in the newest macOS.

Would there be a reason to keep GHC 7.10 as a default?

garyb:
Resolves #2777

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-8.0
+resolver: lts-8.5
 packages:
 - '.'
 extra-deps:
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,7 +1,5 @@
-resolver: lts-6.25
+resolver: lts-8.0
 packages:
 - '.'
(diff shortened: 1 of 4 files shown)


noraesae:
Great! I think configs like `circle.yml` should change too.

stack.yaml
```diff
@@ -1,7 +1,5 @@
-resolver: lts-6.25
+resolver: lts-8.0
```
kRITZCREEK: The newest snapshot is 8.5 at this point, might as well use the newest versions? https://www.stackage.org/lts-8.5

paf31:
👍 LGTM when the libraries can be fixed too, thanks!

PR #2784, opened 2017-03-26
Polymorphic Labels

paf31:
This defines a new class called `Lookup` in `Prim`, which is a three-way relation between labels, rows and types. It gives us a basic form of polymorphic labels which is enough to define

- a generic record property lens (see the example)
- polymorphic variants (as a library)


examples/passing/PolyLabels.purs
```diff
@@ -0,0 +1,65 @@
...
+get
+  :: forall r l a
+   . IsSymbol l
+  => Lookup l r a
+  => SProxy l
+  -> Record r
+  -> a
+get l = unsafeGet (reflectSymbol l)
+
+set
```
paf31: Actually, this isn't quite right, since `r1` and `r2` are unrelated apart from that they must both contain `l`. We might have to stick to monomorphic updates for now.

paf31: I wonder if we need a 4-way relation which acts like `uncons` instead of `head`.

LiamGoodacre: If we could construct a singleton row from a symbol and a type I think we could do this.  Something like this?
```purescript
class Singleton (l :: Symbol) (t :: Type) (r :: # Type) | l t -> r

set :: forall l a b u r1 r2.
  IsSymbol l =>       -- l = "x"
  Singleton l b u =>  -- l = "x", b = X, u = (x :: X)
  Union u r1 r2 =>    -- u = (x :: X), r1 = (x :: Y, z :: Z), r2 = (x :: X, x :: Y, z :: Z)
  Lookup l r1 a =>    -- l = "x", r1 = (x :: Y, z :: Z), a = Y
  SProxy l -> b -> Record r1 -> Record r2
```

paf31: @LiamGoodacre Yes, but I think I prefer to modify `Lookup` so that it also returns the tail (the row with the chosen label removed). That way, inference works both ways - we can view it as subtracting the label from one argument, or adding it to the other (both functional dependencies are valid).

LiamGoodacre: Yeah that's quite nice, and it also covers `Singleton` with: `Singleton l t r = Lookup l t r ()`.
Although the fun deps are slightly different; would it also make sense for `Lookup` to have `r1 r2 -> l a`?

paf31:
I've simplified things quite a bit by moving to a `RowCons` relation with 4 arguments:

- a label
- a type
- an input row
- the output row, which is the result of consing that label and type onto the front of the input

This literally makes the implementation of `forClassName` into a one-liner 😄 

Feedback/reviews appreciated!

PR #2785, opened 2017-03-26
Use latest 0.11-compatible libraries for tests

paf31:
Thanks @garyb and @noraesae for doing the bulk of the work here.


garyb:
Restarted everything now generics has a release ➰

PR #2786, opened 2017-03-27
Wrap decl title in span for better double-click selection

rightfold:
Without this, double-clicking the decl title will also select the source link, which makes copying the name too difficult.


hdgarrood:
Do you know why CI is failing?

rightfold:
@hdgarrood Something with library versions I think.

paf31:
Thank you both!

Issue #2787, opened 2017-03-27
`between` be all insane


garyb:
I don't... it, I... 🙊 

``` purescript
between :: forall a. Ord a => a -> a -> a -> Boolean
between low hi x
  | x < low = false
  | x > hi = false
  | true = true
```

``` javascript
var between = function (dictOrd) {
    return function (low) {
        return function (hi) {
            return function (x) {
                return true;
            };
        };
    };
};
```

paf31:
I'm guessing this is due to some dodgy optimization which is looking at `if (true)` in the last case.

paf31:
This worked before 0.11?

garyb:
A knock-on of #2654 perhaps?

paf31:
My money is on either that or pattern guards, since I know there was some new logic for fallthrough cases there too.

garyb:
It was picked up by the tests in `purescript-orders`, so yeah it must have worked before... should be in the prelude tests now though actually, so I'll move it there.

paf31:
I can confirm this is not an optimizer issue.

PR #2789, opened 2017-03-28
#2787 Fix special case in guard codegen

paf31:
`foldr`/`foldl` mixup. I optimize the `true` case away elsewhere now, which should be simpler.

I think we should perhaps make a 0.11.1 tag soon after this gets merged.

### DIFF
--- a/examples/passing/2787.purs
+++ b/examples/passing/2787.purs
@@ -1,0 +1,8 @@
+module Main where
+
+import Prelude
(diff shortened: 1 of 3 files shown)


garyb:
Agreed that we should make a 0.11.1 ASAP too, for this and the TCO fix.

paf31:
Merging, since the builds are just timing out.

PR #2790, opened 2017-03-28
List instance info under correct sections

### LINKED ISSUE #2780: Documentation issue: type class instances are listed under kinds that have the same name as the class
I have a kind and a class both named `SocketType`. All the instances of the class are also listed under the kind in the documentation. Example: https://pursuit.purescript.org/packages/purescript-zmq/0.0.1/docs/Network.ZMQ#k:SocketType

paf31:
Fixes #2780


paf31:
Again, just timing out.

PR #2791, opened 2017-03-28
PSCi eval test suite

noraesae:
This PR enables writing PSCi eval tests as test suites.

We can write a PSCi test suite like below:

```purescript
import Prelude
import Data.Array

let fac n = foldl mul 1 (1..n) in fac 10 -- 3628800

fac n = foldl mul 1 (1..n)
fac 10 -- 3628800
```

A line with a following comment (`--`) will be evaluated and the result will be compared with the comment. Multiple `--`s in a line will break the test though, I think just splitting by `--` and comparing both sides is quite enough for test suites.

The result output is like below.

Success:
```
###############################################################################
# psci test suite
###############################################################################

Cases: 31  Tried: 30  Errors: 0  Failures: 0
PSCi eval test suites
  BasicEval.purs
  BasicEval2.purs
Cases: 31  Tried: 31  Errors: 0  Failures: 0
```

Failure:
```
###############################################################################
# psci test suite
###############################################################################

Cases: 31  Tried: 30  Errors: 0  Failures: 0
PSCi eval test suites
  BasicEval.purs
  BasicEval2.purs
### Failure in: 2:evalTests
tests/TestPsci/TestEnv.hs:75
expected: "3628801\n"
 but got: "3628800\n"
Cases: 31  Tried: 31  Errors: 0  Failures: 1

Completed 2 action(s).
Test suite failure for package purescript-0.11.0
    tests:  exited with: ExitFailure 1
Logs printed to console
```

### DIFF
--- a/examples/psci/BasicEval.purs
+++ b/examples/psci/BasicEval.purs
@@ -1,7 +1,10 @@
 import Prelude
 import Data.Array
 
-let fac n = foldl mul 1 (1..n) in fac 10 -- 3628800
+-- @shouldEvaluateTo 3628800
+let fac n = foldl mul 1 (1..n) in fac 10
 
 fac n = foldl mul 1 (1..n)
-fac 10 -- 3628800
+
+-- @shouldEvaluateTo 3628800
+fac 10
--- a/examples/psci/BasicEval.purs
+++ b/examples/psci/BasicEval.purs
@@ -1,0 +1,7 @@
+import Prelude
+import Data.Array
+
+let fac n = foldl mul 1 (1..n) in fac 10 -- 3628800
+
+fac n = foldl mul 1 (1..n)
+fac 10 -- 3628800
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -439,7 +439,8 @@ test-suite tests
                    transformers -any,
                    transformers-compat -any,
                    utf8-string -any,
-                   vector -any
+                   vector -any,
+                   split -any
     ghc-options: -Wall
     type: exitcode-stdio-1.0
     main-is: Main.hs
@@ -451,6 +452,7 @@ test-suite tests
                    TestPsci
                    TestPsci.CommandTest
                    TestPsci.CompletionTest
+                   TestPsci.EvalTest
                    TestPsci.TestEnv
                    TestPscIde
                    PscIdeSpec
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -433,14 +433,14 @@ test-suite tests
                    process -any,
                    protolude >= 0.1.6,
                    silently -any,
+                   split -any,
                    stm -any,
                    text -any,
                    time -any,
                    transformers -any,
                    transformers-compat -any,
                    utf8-string -any,
-                   vector -any,
-                   split -any
+                   vector -any
     ghc-options: -Wall
     type: exitcode-stdio-1.0
     main-is: Main.hs
--- a/tests/TestPsci/EvalTest.hs
+++ b/tests/TestPsci/EvalTest.hs
@@ -5,6 +5,7 @@ import Prelude.Compat
 
 import           Control.Monad (forM_)
 import           Data.Char (isSpace)
+import           Data.List (dropWhileEnd)
 import           Data.List.Split (splitOn)
 import           System.Directory (getCurrentDirectory)
 import           System.IO.UTF8 (readUTF8File)
@@ -35,8 +36,7 @@ evalTest f = do
       _ -> run line
 
 trim :: String -> String
-trim = f . f
-  where f = reverse . dropWhile isSpace
+trim = dropWhile isSpace . dropWhileEnd isSpace
 
 trimAndFilter :: [String] -> [String]
 trimAndFilter = filter (not . null) . map trim
--- a/tests/TestPsci/EvalTest.hs
+++ b/tests/TestPsci/EvalTest.hs
@@ -1,0 +1,42 @@
+module TestPsci.EvalTest where
+
+import Prelude ()
+import Prelude.Compat
+
+import           Control.Monad (forM_)
+import           Data.Char (isSpace)
+import           Data.List.Split (splitOn)
+import           System.Directory (getCurrentDirectory)
+import           System.IO.UTF8 (readUTF8File)
+import           System.FilePath ((</>), takeFileName)
+import qualified System.FilePath.Glob as Glob
+import           Test.HUnit
+import           TestPsci.TestEnv
+
+evalTests :: Test
+evalTests = TestLabel "evalTests" . TestCase $ do
+  putStrLn "\nPSCi eval test suites"
+  testFiles <- evalTestFiles
+  forM_ testFiles evalTest
+
+evalTestFiles :: IO [FilePath]
+evalTestFiles = do
+  cwd <- getCurrentDirectory
+  let psciExamples = cwd </> "examples" </> "psci"
+  Glob.globDir1 (Glob.compile "**/*.purs") psciExamples
+
+evalTest :: FilePath -> IO ()
+evalTest f = do
+  putStrLn $ "  " ++ takeFileName f
+  ls <- trimAndFilter . lines <$> readUTF8File f
+  execTestPSCi $ forM_ ls $ \ line ->
+    case trimAndFilter $ splitOn " -- " line of
+      [expr, expected] -> expr `evaluatesTo` expected
+      _ -> run line
+
+trim :: String -> String
+trim = f . f
+  where f = reverse . dropWhile isSpace
+
+trimAndFilter :: [String] -> [String]
+trimAndFilter = filter (not . null) . map trim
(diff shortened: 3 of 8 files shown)


tests/TestPsci/EvalTest.hs
```diff
@@ -0,0 +1,42 @@
...
+  putStrLn $ "  " ++ takeFileName f
+  ls <- trimAndFilter . lines <$> readUTF8File f
+  execTestPSCi $ forM_ ls $ \ line ->
+    case trimAndFilter $ splitOn " -- " line of
+      [expr, expected] -> expr `evaluatesTo` expected
+      _ -> run line
+
+trim :: String -> String
+trim = f . f
+  where f = reverse . dropWhile isSpace
```
kRITZCREEK: There is `dropWhileEnd` in Data.List, that would make this a little easier to read: `dropWhile isSpace . dropWhileEnd isSpace`

noraesae: Didn't know that! Will update.

paf31:
What about using something like the `-- @shouldFailWith` and `-- @shouldWarnWith` headers that we use for the failing and warning tests?

Looks good otherwise, thanks!

purescript.cabal
```diff
@@ -439,7 +439,8 @@ test-suite tests
                    transformers -any,
                    transformers-compat -any,
                    utf8-string -any,
-                   vector -any
+                   vector -any,
+                   split -any
```
paf31: Can you please resort these?

noraesae:
Hmm, yes, I looked over multiline input. We can put `>` in front of each line.

```purescript
> import Prelude
> import Data.Array

> let fac n = foldl mul 1 (1..n) in fac 10 -- 3628800

> fac :: Int -> Int
fac n = foldl mul 1 (1..n)
> fac 10 -- 3628800
```

But it may not look so nice, and is not how PSCi works (`:paste`). Any idea?

paf31:
Since we're only testing evaluation right now, I think this is fine, but later we might want to test things like `:browse`, at which point we'll need to support multiline output.

Looks good to me, thanks!

examples/psci/BasicEval.purs
```diff
@@ -0,0 +1,7 @@
+import Prelude
+import Data.Array
+
+let fac n = foldl mul 1 (1..n) in fac 10 -- 3628800
```
paf31: Is this comment used to test anything?

examples/psci/BasicEval.purs
```diff
@@ -0,0 +1,7 @@
...
+
+fac n = foldl mul 1 (1..n)
+fac 10 -- 3628800
```
paf31: How about this?

```purescript
-- @shouldEvaluateTo 3628800
fac 10
```

noraesae: Separated line comments may be safer to note the eval tests! I will update this PR.

examples/psci/BasicEval.purs
noraesae: It's to test the evaluated result of the let-in expr, `let fac n = foldl mul 1 (1..n) in fac 10`.

noraesae:
I update the PR to use `@shouldEvaluateTo` comment to note eval tests.

If invalid `@` comment is used, an error message like below will show.

```
PSCi eval test suites
  BasicEval.purs
invalid comment: -- @someInvalidComment 3628800
### Error in:   2:evalTests
ExitFailure 1
```

noraesae:
I've just come up with an idea for multiline input.

```purescript
import Prelude
import Data.Array

-- @paste
fac :: Int -> Int
fac n = foldl mul 1 (1..n)
-- @paste

-- @shouldEvaluateTo 3628800
fac 10
```

How do you think of it?

noraesae:
I added multiline support as I suggested. Please feel free to leave comments on it.

tests/TestPsci/EvalTest.hs
```diff
@@ -0,0 +1,67 @@
...
+                 | None
+                 deriving (Show)
+
+evalCommentPrefix :: String
+evalCommentPrefix = "-- @"
+
+parseEvalLine :: String -> EvalLine
+parseEvalLine "" = Empty
+parseEvalLine line
+  | evalCommentPrefix `isPrefixOf` line =
```
paf31: You can use `splitPrefix` here.

tests/TestPsci/EvalTest.hs
```diff
@@ -0,0 +1,67 @@
...
+  testFiles <- evalTestFiles
+  forM_ testFiles evalTest
+
+evalTestFiles :: IO [FilePath]
+evalTestFiles = do
+  cwd <- getCurrentDirectory
+  let psciExamples = cwd </> "examples" </> "psci"
+  Glob.globDir1 (Glob.compile "**/*.purs") psciExamples
+
+data EvalLine = Line String
```
paf31: Can we use `Text` throughout please?

tests/TestPsci/EvalTest.hs
```diff
@@ -0,0 +1,67 @@
...
+  | evalCommentPrefix `isPrefixOf` line =
+    case splitOn " " $ drop (length evalCommentPrefix) line of
+      "shouldEvaluateTo" : args -> Comment (ShouldEvaluateTo $ intercalate " " args)
+      "paste" : [] -> Comment (Paste [])
+      _ -> Invalid line
+  | otherwise = Line line
+
+evalTest :: FilePath -> IO ()
+evalTest f = do
+  putStrLn $ "  " ++ takeFileName f
```
paf31: Debugging code?

paf31:
Looks good, and I think this is a great start which we can iterate on. Thanks! I just have a few other comments which aren't related to the syntax.

noraesae:
Summing up your comments, I may consider rewrite this in hspec, which is also used for PSC compiler test suites. Thanks!

hdgarrood:
I've wanted to go through the test suite and convert them to use some kind of library like that for quite a while actually. Some of the tests already use hspec so I think that would be the obvious choice.

hdgarrood:
Oops, that's what you just said.

tests/TestPsci/EvalTest.hs
noraesae: I'm not aware of `splitPrefix`, is it a util function? I cannot find one in hoogle.

tests/TestPsci/EvalTest.hs
noraesae: I've once tried to use `Text`, but many functions used in the test is just using `String`, and I though if we use `Text`, there will be lots of `T.pack` and `T.unpack` everywhere.

For example, in PSCi tests and PSCi itself uses `readProcessWithExitCode` to run JS code in Node.js, and it just returns `String`. Also, `parseCommand` of PSCi parser is using `String` as its input. The lines will be evaluated and compared with results often, so I think `String` will work here.

tests/TestPsci/EvalTest.hs
noraesae: It was a title of each test case. Not we don't need this because Hspec's `specify` is used instead.

tests/TestPsci/EvalTest.hs
```diff
@@ -0,0 +1,67 @@
...
+  execTestPSCi $ foldM_ handleLine None evalLines
+
+handleLine :: EvalContext -> EvalLine -> TestPSCi EvalContext
+handleLine context Empty = pure context
+handleLine None (Line stmt) = run stmt >> pure None
+handleLine None (Comment context) = pure context
+handleLine (ShouldEvaluateTo expected) (Line expr) = expr `evaluatesTo` expected >> pure None
+handleLine (Paste ls) (Line l) = pure . Paste $ ls ++ [l]
+handleLine (Paste ls) (Comment (Paste _)) = run (intercalate "\n" ls) >> pure None
+handleLine _ line = liftIO $ putStrLn ("unexpected: " ++ show line) >> exitFailure
```
noraesae: Now Hspec is used and it will not exit instantly.

noraesae:
I changed it to use Hspec.

tests/TestPsci/EvalTest.hs
paf31: Sorry, [stripPrefix](https://hackage.haskell.org/package/base-4.9.1.0/docs/Data-List.html#v:stripPrefix).

noraesae: Oh, didn't know that. Thanks!

paf31:
@kRITZCREEK Good to merge?

kRITZCREEK:
```
[ 2 of 14] Compiling Version          ( app\Version.hs, .stack-work\dist\ca59d0ab\build\purs\purs-tmp\Version.o )
Segmentation fault/access violation in generated code
```

Doesn't sound too good... I think `hspec` isn't supposed to be called multiple times in the same testsuite. We can merge this, and I'll look into tasty to give us a tool that supports multiple test suites.

paf31:
Let's refactor it to only call hspec once then.

kRITZCREEK:
I think the way I set up the purs ide test suite at the moment makes that unnecessarily hard. It just appeared to me that we've been running multiple `hspec`'s all this time, since both the compiler tests as well as the ide tests call it. 

Let's merge this for now, I played with tasty on another project of mine and it seems like a good fit for what we're doing here. Once this is in, I'll figure things out and make a PR.

paf31:
Thank you both!

PR #2799, opened 2017-03-30
Use XDG Base Directory Specification for psci_history

legrostdg:
Use $HOME/.purescript/psci_history if it exists, else it uses $XDG_DATA_DIR/purescript/psci_history. Works on Windows, too.

https://www.stackage.org/haddock/lts-8.6/directory-1.3.0.0/System-Directory.html#v:getAppUserDataDirectory


legrostdg:
"The job exceeded the maximum time limit for jobs, and has been terminated."

paf31:
Thanks, but can you explain the benefit of this? Also, can you please update contributors.md?

legrostdg:
On linux, it avoids polluting the user's home directory. On windows, it uses a directory that does not look like a dirty adaptation of a linux specific `.` hidden dir. More and more applications uses this spec.

paf31:
Thanks for clarifying that. I'm fine with merging this, but we should make sure there are no other dotfiles that we create, for which we need to do the same thing.

paf31:
It appears this is the only occurrence actually.

PR #2798, opened 2017-03-30
[purs ide] Collect more information for classes and synonyms

kRITZCREEK:
When serializing to the Externs format, the compiler decomposes and duplicates
some information about type classes and synonyms. This PR reconstructs that
information and stores it in their respective IdeDeclaration.

I still want to try this on a few codebases, before merging it.

PR #2800, opened 2017-03-31
[WIP][purs ide] Completion formats

kRITZCREEK:
After #2798 

Still a work in progress, but I decided to share what I have so we can talk about the design.

Currently the pipeline for completions (and type information) looks as follows:

`Declarations |> Filters |> Matcher |> Completions`

I propose to add a new Formatting step:

`Declarations |> Filters |> Matcher |> _Formatter_ |> Completions`

For now I want to use this to allow configuring whether and how Reexports should be grouped, but in the future I'd like to implement Formatters that are not necessarily tailored at providing completions, but rather the full breadth of information available inside purs ide's state. Some Ideas:

- A GraphQL like specification of the to be returned schema
- Max number of results
- Type format: (Unicode, Single/Multiline, a structured format?)
- Append possible followup commands, like say the correct add import command

/cc @nwolverson @chexxor @FrigoEU

### DIFF
--- a/src/Language/PureScript/Ide/Completion/Formatter.hs
+++ b/src/Language/PureScript/Ide/Completion/Formatter.hs
@@ -1,0 +1,135 @@
+module Language.PureScript.Ide.Completion.Formatter
+  ( CompletionFormatter(..)
+  , ReexportStrategy(..)
+  , runFormatter
+  , defaultFormatter
+  ) where
+
+import Protolude hiding ((&), from, to)
+import Control.Lens hiding (op)
+import Data.Aeson
+import qualified Data.Map as Map
+import qualified Data.Text as T
+import Language.PureScript.Ide.Types
+import Language.PureScript.Ide.Util
+import qualified Language.PureScript as P
+
+data ReexportStrategy
+  = NoFlatten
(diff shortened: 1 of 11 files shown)


src/Language/PureScript/Ide/Completion/Formatter.hs
```diff
@@ -0,0 +1,134 @@
...
+      Nothing -> Just (Match (origin, d), [moduleName])
+      Just x -> Just (second (moduleName :) x)
+    insertDeclaration moduleName d old = case old of
+      Nothing -> Just (Match (moduleName, d), [])
+      Just x -> Just x
+
+
+getNS :: IdeDeclaration -> Text -> IdeDeclNamespace
+getNS d = case d of
+  IdeDeclValue _ -> IdeNSValue
```
nwolverson: Feel like I'm reading ObjectiveC

kRITZCREEK: Sorry for doing that to you... I'll fully spell it out :D

kRITZCREEK:
I don't like this approach, I'll try to do something more lightweight first.

Issue #2795, opened 2017-03-30
Pattern guards give false exhaustivity errors


paf31:
Example:

```purescript
module Main where

import Prelude

data X = X Int | Y

x :: X -> Int
x = case _ of
      Y -> 0
      X n | 1 <- n -> 1
          | otherwise -> 2
```

 which gives:

```purescript
Error found:
in module Main
at src/Main.purs line 9, column 5 - line 12, column 26

  A case expression could not be determined to cover all inputs.
  The following additional cases are required to cover all inputs:

    Y

  Alternatively, add a Partial constraint to the type of the enclosing value.

while applying a function __unused
  of type Partial => t0 -> t0
  to argument case $0 of              
                (X n) | otherwise -> 2
while inferring the type of __unused (case $0 of              
                                        (X n) | otherwise -> 2
                                     )                        
in value declaration x

where t0 is an unknown type
```

It looks like the desugaring is at fault here, since `X n` is the only alternative left after things are expanded.

@alexbiehl Any ideas here?

alexbiehl:
`x` is transformed into

```haskell
x :: X -> Int 
x y = case y of 
    Y -> 0 
    _ -> let 
           k _ = case y of 
                 X n | otherwise -> 2
          in case y of 
            X n -> case n of -- guard 1 <- n
                1 -> 1
                _ -> k true 

```

We scrutinize `y` multiple times here but only bind to `X n`, missing `Y` in those cases.

alexbiehl:
Lets see if I can come up with something better.

alexbiehl:
I think the transformation is right, the exhaustivity checker just doesn't know that in subsequent matches missing cases can be ruled out. 

We have different alternatives I guess: 

- Remember what constructors we already matched and generate "unreached" alternatives. I think we lack a notion of "unreached" alternative and I am not sure we want such thing. We can't just use wildcard binders alone as that would result false-exhaustive cases which is even worse.
- Just don't desugar cases before exhaustivity checking. Desugar them in the ps2core pass instead. I had the part for type checking and exhaustivity around when developing the original feature so this should be no big deal.

btw. Are there any written down notes on invariants of case expressions in surface language?

paf31:
Thanks for looking at this.

If it's possible to move the transformation after linting and/or type-checking, I think that would be the way to go.

matthewleon:
I am still having this problem on 0.11.2.

Code:

```
head :: forall a. Stream a -> a
[...]

PR #2801, opened 2017-03-31
Desugar pattern guards *after* type checking

alexbiehl:
@paf31 

This is a fix for https://github.com/purescript/purescript/issues/2795

It's not working properly currently:
```
purs.exe: An internal error occurred during compilation: "Binders should have been desugared"
  error, called at src\Language\PureScript\Crash.hs:24:3 in purescript-0.11.1-inplace:Language.PureScript.Crash
  internalError, called at src\Language\PureScript\Sugar\BindingGroups.hs:112:33 in purescript-0.11.1-inplace:Language.PureScript.Sugar.BindingGroups
purs.exe: thread blocked indefinitely in an MVar operation
```

I am using `everywhereOnValuesM` traversal to traverse the list of `Declaration`s. It seems to miss the expressions in there. Please take a look at `desugarCaseGuards` for a hint. Any pointer on how to fix this?

### DIFF
--- a/src/Language/PureScript/Sugar/CaseDeclarations.hs
+++ b/src/Language/PureScript/Sugar/CaseDeclarations.hs
@@ -5,6 +5,7 @@
 module Language.PureScript.Sugar.CaseDeclarations
   ( desugarCases
   , desugarCasesModule
+  , desugarCaseGuards
   ) where
 
 import Prelude.Compat
@@ -37,14 +38,23 @@ desugarCasesModule (Module ss coms name ds exps) =
       <$> (desugarCases <=< desugarAbs <=< validateCases $ ds)
       <*> pure exps
 
+desugarCaseGuards
+  :: forall m. (MonadSupply m, MonadError MultipleErrors m)
+  => [Declaration]
+  -> m [Declaration]
+desugarCaseGuards declarations = parU declarations f
+  where
+    (f, _, _) = everywhereOnValuesM return desugarGuardedExprs return
+
 -- |
 -- Desugar case with pattern guards and pattern clauses to a
 -- series of nested case expressions.
 --
-desugarCase :: forall m. (MonadSupply m)
-            => Expr
-            -> m Expr
-desugarCase (Case scrut alternatives)
+desugarGuardedExprs
+  :: forall m. (MonadSupply m)
+  => Expr
+  -> m Expr
+desugarGuardedExprs (Case scrut alternatives)
   | any (not . isTrivialExpr) scrut = do
     -- in case the scrutinee is non trivial (e.g. not a Var or Literal)
     -- we may evaluate the scrutinee more than once when a guard occurrs.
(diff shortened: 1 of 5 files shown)


alexbiehl:
(Ah I discovered binding groups are a part of `Purescript.Sugar`, so this explains it)

alexbiehl:
I have discovered another issue. For a test I wrote: 

```
f x
  | 1 <- x = x
  | true = 1234
```

This gets correctly translated to: 

```
f x = let 
          v _ = 1234
        in case x of 
               1 -> x
               _ -> v true
``` 

Now the codegen:

``` 
var g = function (x) {
    var v = function(x0) { return 1234; }

    if (x === 1) {
        return x;
    };
    return v(true);
};
```

It rightfully spots the very small function `v´ (btw. this seems to be new, when I wrote the pattern guards patch it didn't do such things) and inlines it: 

``` 
var g = function (x) {
    if (x === 1) {
        return x;
    };
    return 1234(true);
};
```

It still applies `true` to `1234` which is of course not correct!

paf31:
@alexbiehl are you able to verify that the issue is in the optimizer? You could turn the optimizer off and see if the generated code is correct.

alexbiehl:
Using
```haskell
-- | Apply a series of optimizer passes to simplified JavaScript code
optimize :: MonadSupply m => AST -> m AST
optimize js = pure js
```

I get 
```js
"use strict";
var f = function (x) {
    return (function () {
        var v = (function () {
            var $1 = x;
            var x1 = $1;
            if (true) {
                return 1234;
            };
            throw new Error("Failed pattern match at X line 3, column 1 - line 5, column 12: " + [ $1.constructor.name ]);
        })();
        return (function () {
            var $2 = x;
            var x1 = $2;
            return (function () {
                var $3 = x1;
                if ($3 === 1) {
                    return x1;
                };
                return v(true);
                throw new Error("Failed pattern match at X line 3, column 1 - line 5, column 12: " + [ $3.constructor.name ]);
            })();
            throw new Error("Failed pattern match at X line 3, column 1 - line 5, column 12: " + [ $2.constructor.name ]);
        })();
    })();
};
```

This looks correct. Variable`v` is only bound once. This makes it susceptible for inlining and then maybe `etaConvert` does the rest?

alexbiehl:
Ah! No, we can see `v` is being invoked immediately. It doesn't reduce to a function, it reduces to `1234` which of course is replaced by `v`s occurrence. Somewhere a bug is lurking...

paf31:
Is it possible to run your pass between the exhaustivity and type checking passes? Would that help?

src/Language/PureScript/Sugar/CaseDeclarations.hs
```diff
@@ -216,8 +228,8 @@ desugarCase (Case scrut alternatives) =
           alt_fail = [CaseAlternative [NullBinder] [MkUnguarded goto_rem_case]]
 
         pure $ Let [
-            ValueDeclaration rem_case_id Private [NullBinder]
-              [MkUnguarded desugared]
+          ValueDeclaration rem_case_id Private []
+            [MkUnguarded (Abs (VarBinder unused_binder) desugared)]
```
alexbiehl: This was the culprit. After desugaring we need to generate code which obeys the invariants of desugared code..

alexbiehl:
@paf31 Please take a look; I think this is ready for review now. Test suite succeeds locally.

PR #2802, opened 2017-03-31
Use hpack to generate the .cabal file

kRITZCREEK:
I recently stumbled upon https://github.com/sol/hpack and thought it did solve some of the annoyances with the `.cabal` format. Also the integration with stack is neat, since it automatically generates the `.cabal` file without any extra options.

If we were to use this, I think we'd want to keep the generated `.cabal` file checked in, so someone could still attempt to build the compiler with `cabal` without having to install `hpack` beforehand.

I think this is an improvement, but I'm totally fine with just closing this if there is a good reason to do so.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -1,0 +1,155 @@
+name: purescript
+version: '0.11.1'
+synopsis: PureScript Programming Language Compiler
+description: A small strongly, statically typed programming language with expressive
+  types, inspired by Haskell and compiling to JavaScript.
+category: Language
+author: >
+  Phil Freeman <paf31@cantab.net>,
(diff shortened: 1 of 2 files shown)


paf31:
I like it!

rvion:
I have been using it for months: hpack works very well. 

:memo: You can generate package manifests automatically with hpack-convert https://hackage.haskell.org/package/hpack-convert

kRITZCREEK:
@rvion I did that, but it errored on a few of the multiline fields we had in the compiler. With some tweaking I got it to work like it does in the PR now :)

Issue #2722, opened 2017-03-09
Consolidate import list in PSCi


noraesae:
From #2721.

```
import Prelude ((+))
import Prelude (bind)
import Prelude as P ((*))
import Prelude as P (show)
```

should give

```
> :show import
import Prelude ((+), (*))
import Prelude as P (bind, show)
```

noraesae:
Also I guess,

```
import Prelude ((+))
import Prelude (bind)
import Prelude as P ((*))
import Prelude as P (show)
import Prelude
import Prelude
```

should give

```
> :show import
import Prelude
import Prelude as P (bind, show)
```

rndnoise:
This might need to be designed with #982 in mind, since they could interact with one another.

LiamGoodacre:
@utatti in those examples, shouldn't `(*)` and `bind` be swapped in the expected output?

utatti:
@LiamGoodacre You're right. I've updated the example.

PR #2808, opened 2017-04-04
Accept print function as handleCommand argument

noraesae:
Currently I'm working on #2722 and found that output of `:show import` is printed into console, and thus there is no way to test the output.

This PR makes PSCi `handleCommand` to accept a print function as its argument, and enables checking output of PSCi commands and adding test suites for them in the future.

### DIFF
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -90,22 +90,23 @@ make ms = do
 -- | Performs a PSCi command
 handleCommand
   :: (MonadReader PSCiConfig m, MonadState PSCiState m, MonadIO m)
-  => (String -> m ())
-  -> m ()
+  => (String -> m ()) -- ^ evaluate JS
+  -> m () -- ^ reload
+  -> (String -> m ()) -- ^ print into console
   -> Command
   -> m ()
-handleCommand _ _ ShowHelp                  = liftIO $ putStrLn helpMessage
-handleCommand _ r ReloadState               = handleReloadState r
-handleCommand _ r ClearState                = handleClearState r
-handleCommand c _ (Expression val)          = handleExpression c val
-handleCommand _ _ (Import im)               = handleImport im
-handleCommand _ _ (Decls l)                 = handleDecls l
-handleCommand _ _ (TypeOf val)              = handleTypeOf val
-handleCommand _ _ (KindOf typ)              = handleKindOf typ
-handleCommand _ _ (BrowseModule moduleName) = handleBrowse moduleName
-handleCommand _ _ (ShowInfo QueryLoaded)    = handleShowLoadedModules
-handleCommand _ _ (ShowInfo QueryImport)    = handleShowImportedModules
-handleCommand _ _ _                         = P.internalError "handleCommand: unexpected command"
+handleCommand _ _ p ShowHelp                  = p helpMessage
+handleCommand _ r _ ReloadState               = handleReloadState r
+handleCommand _ r _ ClearState                = handleClearState r
+handleCommand e _ _ (Expression val)          = handleExpression e val
+handleCommand _ _ _ (Import im)               = handleImport im
+handleCommand _ _ _ (Decls l)                 = handleDecls l
+handleCommand _ _ p (TypeOf val)              = handleTypeOf p val
+handleCommand _ _ p (KindOf typ)              = handleKindOf p typ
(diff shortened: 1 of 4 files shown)


src/Language/PureScript/Interactive.hs
```diff
@@ -90,22 +90,23 @@ make ms = do
 -- | Performs a PSCi command
 handleCommand
   :: (MonadReader PSCiConfig m, MonadState PSCiState m, MonadIO m)
-  => (String -> m ())
-  -> m ()
+  => (String -> m ()) -- ^ evaluate JS
```
paf31: I wonder if we should put these into an MTL-style type class. The pile of wildcard binders below is getting a little silly :)

paf31:
👍 Seems good generally, just one comment.

src/Language/PureScript/Interactive.hs
noraesae: I was also thinking about it. We can implement something like `PSCiT`.

```haskell
type PSCiEvalJS = String -> IO ()
type PSCiReload = IO ()
type PSCiPrint = String -> IO ()
type PSCiFunctions = (PSCiEvalJS, PSCiReload, PSCiPrint)

newtype PSCiT m a = { runPSCiT :: PSCiFunctions -> m a }

class MonadPSCi m where
  psciEvalJS ::  String -> m ()
  psciReload :: m ()
  psciPrint :: String -> m ()

instance MonadIO m => MonadPSCi (PSCiT m) where
  psciEvalJS js = runPSCiT >>> \ (e, _, _) -> liftIO $ e js
  psciReload = runPSCiT >>> \ (_, r, _) -> liftIO r
  psciPrint str = runPSCiT >>> \ (_, _, p) -> liftIO $ p str
```

Does it look good?

noraesae:
Added `PSCiT` and `MonadPSCi` as I suggested.

paf31:
Sorry for the delay. I would keep `MonadPSCi` but get rid of `PSCiT` and just provide instances for the existing monads, personally.

noraesae:
You mean, for example, `IO`?

noraesae:
I think we cannot just declare `MonadPSCi` instance for existing monads, e.g. `MonadPSCi IO` or `MonadIO m => MonadPSCi m`, as PSCi functions are defined dynamically.

```haskell
Backend setup eval reload (shutdown :: state -> IO ()) ->
  case e of
    Left errs -> putStrLn (P.prettyPrintMultipleErrors P.defaultPPEOptions errs) >> exitFailure
    Right (modules, externs, env) -> do
      ...
      let ...

          handleCommand' :: state -> Command -> StateT PSCiState (ReaderT PSCiConfig IO) ()
          handleCommand' state cmd =
            let fns = PSCiFns (eval state) (reload state) putStrLn
            in runPSCiT (handleCommand cmd) fns
```

There may be need for keeping the functions inside command handling, and thus I guessed we need a new data type `PSCiT`. If I misunderstood something, please let me know.

noraesae:
The PSCi functions are provided locally and dynamically, so I guess we cannot just do `instance Monad m => MonadPSCi (MonadIO m)`. I will thinking of better way not to define `PSCiT` as a new data type.

noraesae:
I made it very short like below:

```haskell
-- | A list of functions used in PSCi
data PSCiFns = PSCiFns { _evalJS :: String -> IO ()
                       , _reload :: IO ()
                       , _print :: String -> IO ()
                       }

-- | A monadic type class for PSCiT
class MonadPSCi m where
  psciEvalJS :: String -> m ()
  psciReload :: m ()
  psciPrint :: String -> m ()

instance (MonadReader PSCiFns m, MonadIO m) => MonadPSCi m where
  psciEvalJS js = ask >>= \ (PSCiFns e _ _) -> liftIO $ e js
  psciReload = ask >>= \ (PSCiFns _ r _) -> liftIO r
  psciPrint str = ask >>= \ (PSCiFns _ _ p) -> liftIO $ p str
```

But it's not possible because the existing PSCi mtl stack already uses `MonadReader PSCiConfig`, and thus has functional dependency of MonadReader, `m -> PSCiConfig`. We can consider putting these functions into PSCiConfig, but I'm not sure you agree.

paf31:
I'll think about this. For now, I think I'd actually rather just stick with the original version, sorry :)

noraesae:
Alright. Will revert the latter then :)

PR #2814, opened 2017-04-07
[purs ide] Use the user-specified log level for file-watcher logs

kRITZCREEK:
Also makes for a more consistent log output.

### DIFF
--- a/app/Command/Ide.hs
+++ b/app/Command/Ide.hs
@@ -119,7 +119,7 @@ command = Opts.helper <*> subcommands where
       putText "psc-ide needs you to compile your project (for example by running pulp build)"
 
     unless noWatch $
-      void (forkFinally (watcher polling ideState fullOutputPath) print)
+      void (forkFinally (watcher polling logLevel ideState fullOutputPath) print)
     let conf = Configuration {confLogLevel = logLevel, confOutputPath = outputPath, confGlobs = globs}
         env = IdeEnvironment {ideStateVar = ideState, ideConfiguration = conf}
     startServer port env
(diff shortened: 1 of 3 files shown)


kRITZCREEK:
While `stack` automatically regenerates the cabal file on every command, it wasn't rerun in e9c0d47c0097c1074be412f9cc95c2fd3c006481, @alexbiehl I guess you're using `cabal`? How can we make sure that the cabal file and the `package.yaml` do not go out of sync?

Maybe we should add a travis script that explicitly runs `hpack` and checks whether the working tree is dirty afterwards?

alexbiehl:
I am not sure I understand. That commit didn't touch the cabal file at all. Why would I need to regenerate the cabal file then?

alexbiehl:
(But yes, I am using cabal to build purescript)

kRITZCREEK:
The test files you added were missing from the cabal file. I added them in https://github.com/purescript/purescript/pull/2814/commits/18e70c0fbc292ce1a8f5e03df11e4f8109bf8707 in this PR. This happened, because `hpack` expands the [Globs](https://github.com/purescript/purescript/blob/5980159aa5d61800830245b5f982f70bb4157e7b/package.yaml#L18-L21) when it generates the cabal file. I guess this is necessary, because `.cabal` files don't support full Glob syntax, in particular the directory wildcard `**`.

alexbiehl:
I see. I will remember that the next time I submit a PR. Thank you!

kRITZCREEK:
Thanks for reviewing @paf31!

Issue #2772, opened 2017-03-25
purs docs --html gets stuck somewhere on QuickCheck


paf31:
It seems like `purs docs` gets stuck when I run

```
purs docs --format html 'src/**/*.purs' 'bower_components/purescript-*/src/**/*.purs'
```

@hdgarrood Do you have any time to look at this?

hdgarrood:
I'll try to look at this soon but I might struggle to find time; I have a few big pieces of coursework due in the next couple of weeks. After that I'll have more time though.

hdgarrood:
Started looking at this. I think it's getting stuck on `Data.List.ZipList`.

hdgarrood:
I think it would be good to do another minor release and update Pursuit with this fix, as currently asking Pursuit to render anything that has a type-level string in it (e.g. `Data.List.ZipList`) seems to trigger the same issue.

paf31:
Sounds good, I'll make a release later today. Thanks!

PR #2817, opened 2017-04-08
Add a bunch of NFData instances

hdgarrood:
I also added derived Generic instances in order to be able to define
the NFData instances without having to write the necessary code by hand;
I expect I'll do it incorrectly if I try to do it by hand.

I am mainly doing this because I want to use it to help diagnose bugs
like #2772 but I also think it might come in handy in real code at some
point too; e.g. if we ever want to store these types in Pursuit's
database.

### DIFF
--- a/src/Language/PureScript/AST/Operators.hs
+++ b/src/Language/PureScript/AST/Operators.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveGeneric #-}
 -- |
 -- Operators fixity and associativity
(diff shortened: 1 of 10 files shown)


hdgarrood:
Thanks! I opted to do it this way because apparently `DeriveAnyClass` doesn't work great together with `GeneralizedNewtypeDeriving` (which we use in one or two places IIRC).

paf31:
Oh good point.

Issue #2448, opened 2016-11-25
psc-docs can consume lots of RAM


LiamGoodacre:
Just tried running `pulp docs` on https://github.com/LiamGoodacre/purescript-type-map.  After a few minutes `psc-docs` was trying to use over 6G of RAM.  My poor 4G macbook pro was not particularly happy with it.

hdgarrood:
I think the problem is coming from inside `Language.PureScript.Docs.Convert`, possibly `Language.PureScript.Docs.ReExports`.

PR #2824, opened 2017-04-10
Fix #2772, match type level strings in docs renderer

### LINKED ISSUE #2772: purs docs --html gets stuck somewhere on QuickCheck

hdgarrood:
This might be the same problem as in #2448 but I haven't been able to verify.

### DIFF
--- a/examples/docs/src/TypeLevelString.purs
+++ b/examples/docs/src/TypeLevelString.purs
@@ -1,0 +1,7 @@
+module TypeLevelString where
+
+data Foo
+
+class Bar a
+
+instance fooBar :: Fail "oops" => Bar Foo
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -1,4 +1,4 @@
--- This file has been generated from package.yaml by hpack version 0.15.0.
+-- This file has been generated from package.yaml by hpack version 0.17.0.
 --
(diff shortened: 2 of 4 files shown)


paf31:
Good catch - these printer bugs are always a pain to identify. Thanks!

hdgarrood:
Great. I think I'll just add a test and check whether it is really #2448 before merging.

hdgarrood:
The instance that was causing this was `zipListIsNotBind`, which is a custom type error for attempts to use any `Bind ZipList` instance. Do you think we should remove any instance that mentions `Prim.Fail` in its constraints for the purposes of docs?

paf31:
I don't think so, I think the few use cases will be few and far between.

hdgarrood:
Ok then. Having that information in the docs will likely be useful anyway, come to think of it.

hdgarrood:
Yes! This does fix #2448 as well.

hdgarrood:
AppVeyor CI failure looks spurious so I think I'll ignore it, especially since the branch build succeeded.

PR #2825, opened 2017-04-11
Bump aeson lower bound to 1.0

hdgarrood:
We do actually rely on aeson being at least 1.0 for our JSON formats.


hdgarrood:
Gah, travis has timed out because it failed to download the cache for some reason, and one of the appveyor builds has started segfaulting while building the `purs` executable. Neither of these are related to this change, and one appveyor and one travis build has succeeded, so I think I'll go ahead and merge this for now.

PR #2826, opened 2017-04-12
Qualify types as necessary in hints in warnings/errors

### LINKED ISSUE #1647: Qualify types as necessary in hints in warnings/errors
For example, I have a module which has a toplevel function which returns a `Text.Parsing.Parsers.Pos.Position` which isn't annotated with a type. The hint that appears with the warning says the inferred type was `forall t11. t11 -> Position`, however, if I write that verbatim, I get a type error "Unknown type Position".

Could we make it so that it if the type isn't already in scope, it gets qualified? So it could print the inferred type as `Text.Parsing.Parsers.Pos.Position` in this case.

paf31:
Fixes #1647

The new test case gives this error now:

```
Error found:
in module RequalifyErrors
at examples/failing/RequalifyErrors.purs line 7, column 6 - line 7, column 10

  Could not match type
        
    M2.X
        
  with type
        
    M1.X
```

It might be nice to test that at some point.

I was originally planning to build a `Map` data structure, but I figured we can do that later if the types become big enough that it becomes a performance issue. Since this only gets evaluated if there is an error, I can't see it being a big problem.

### DIFF
--- a/examples/failing/RequalifyErrors.purs
+++ b/examples/failing/RequalifyErrors.purs
@@ -1,0 +1,8 @@
+-- @shouldFailWith TypesDoNotUnify
+module RequalifyErrors where
+
+import RequalifyErrors.M1 as M1
+import RequalifyErrors.M2 as M2
+
+test M1.X = true
+test M2.X = false
--- a/examples/failing/RequalifyErrors/M1.purs
+++ b/examples/failing/RequalifyErrors/M1.purs
@@ -1,0 +1,3 @@
+module RequalifyErrors.M1 where
+
+data X = X
--- a/examples/failing/RequalifyErrors/M2.purs
+++ b/examples/failing/RequalifyErrors/M2.purs
@@ -1,0 +1,3 @@
+module RequalifyErrors.M2 where
+
+data X = X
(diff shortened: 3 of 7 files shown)


paf31:
I noticed this breaks `TypeString` and `TypeConcat`, since we remove the `Prim` qualifier before matching the module name.

paf31:
Actually, this is quite an awkward problem, and it also breaks the special cases for `Partial`, `Fail`, `bind` and `discard` for the same reason. If we want to rewrite the errors in `Make.hs`, then we would need to keep around both the qualified and "requalified" module names somehow.

garyb:
Yeah, I was thinking this would require the annotations work to be done - I was surprised that it looked so simple to achieve, but that's why I guess :smile:

I'll probably have some time this weekend to restore/redo the annotations stuff. I'm going to try and sit down and just get it all done in one go, as it's unmanageable if drags on much more than that, keeping it up to date with other changes.

paf31:
I'll close this for now, but we can always reuse the code once the annotations are there. Seems like they will be useful for more and more things :)

PR #2828, opened 2017-04-12
Allow user to propagate Warn constraints

### LINKED ISSUE #2813: `Warn` constraints warning on definitions
I wanted to add `Warn` constraints to the functions in `Debug.Trace` so that if they're left in the code the compiler will pick them up.

This works, but the warnings arise for each of the definitions in the module, as well as the places they're used - I can kinda see why it does that, but is it necessary?

I've worked around it by doing this for now:

```
class DebugWarning
instance warn :: Warn "Debug.Trace usage" => DebugWarning
```

and then using the `DebugWarning` constraint on the functions instead.

paf31:
Fixes #2813 

_Edit_: This actually doesn't _quite_ work, since in this case:

```purescript
foo :: Warn "foo" => Int -> Int
foo x = x

bar :: Warn "bar" => Int
bar = foo 42
```

we don't generate a dictionary for `Warn "foo"`, since `Warn "bar"` is available. But then `"foo"` doesn't match `"bar"` and you get a `NoInstanceFound` error, which is a bit weird. I don't think it's worth resorting to backtracking though.

_Edit 2_: I've fixed the issue described above now. See the `CustomWarning3` example.

### DIFF
--- a/examples/warning/CustomWarning2.purs
+++ b/examples/warning/CustomWarning2.purs
@@ -1,0 +1,11 @@
+-- @shouldWarnWith UserDefinedWarning
+module Main where
+
+foo :: Warn "foo" => Int -> Int
+foo x = x
+
+bar :: Warn "foo" => Int
+bar = foo 42
+
+baz :: Int
+baz = bar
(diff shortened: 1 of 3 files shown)


paf31:
This is ready for review again.

garyb:
Oops, wrong option 😄

PR #2831, opened 2017-04-12
TCO simplification

### LINKED ISSUE #2822: TCO introduces memory leak
It looks like the JavaScript code generated by TCO retains the `__copy_var` function argument unnecessarily, creating a memory leak.

Some test code demonstrating this is available here: https://github.com/matthewleon/purescript-lists/tree/profile-length. `npm run qa` will create a `test.js` file in the `qa` directory, which you can then serve to verify the memory issues I'll discuss below.

The test is quite simple: in a single line of code, print the `length` of a sufficiently long lazy List (1000000 in the case of this test).

Expected behavior: `length` is a left fold of `( _ + 1)`. Since we aren't retaining a reference to the list anywhere in our PS code, as the list gets progressively forced and the length progressively calculated, GC should progressively clean up the forced left side of the list.

What actually happens: Alas, no. GC cannot kick in until the entire list is forced, by which point, with a sufficiently long list of sufficiently fat members, we will have devoured an enormous amount of RAM.

Why this happens: TCO compiles `foldl` into the JS below:

```javascript
    var go = function (__copy_b) {
        return function (__copy_xs) {
            var b = __copy_b;
            var xs = __copy_xs;
            var __tco_done = false;
            var __tco_result;
            var __tco_b;
            var __tco_xs;
            function __tco_loop(b, xs) {
                var v = step(xs);
                if (v instanceof Nil) {
                    __tco_done = true;
                    return b;
                };
                if (v instanceof Cons) {
                    __tco_b = op(b)(v.value0);
                    __tco_xs = v.value1;
                    return;
                };
                throw new Error("Failed pattern match at Data.List.Lazy.Types line 117, column 7 - line 119, column 40: " + [ v.constructor.name ]);
            };
            while (!__tco_done) {
                __tco_result = __tco_loop(b, xs);
                b = __tco_b;
                xs = __tco_xs;
            };
            return __tco_result;
        };
    };
    return go;
``` 

Line 2 retains a reference to `__copy_xs`, which is then copied into `xs` and never touched again. This reference results in retaining the entire linked list that we are gradually forcing, preventing GC from kicking in.

Solutions:

After line 4, you can add a line `__copy_xs = undefined`, which fixes the issue and allows GC to function properly.

Likewise, simple eliminating the copy of `__copy_xs` to `xs` and directly referencing `__copy_xs` fixes the issue.

As a side note, there is probably something similar happening above with `__copy_b` and `b`, but the potentially leak is far less damaging in that case.

matthewleon:
fixes #2822 

Will post before/after code below.

### DIFF
--- a/src/Language/PureScript/CoreImp/Optimizer/TCO.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer/TCO.hs
@@ -82,18 +82,13 @@ tco = everywhere convert where
   toLoop :: Text -> [Text] -> AST -> AST
   toLoop ident allArgs js =
       Block rootSS $
-        map (\arg -> VariableIntroduction rootSS arg (Just (Var rootSS (copyVar arg)))) allArgs ++
+        map (\arg -> VariableIntroduction rootSS (tcoVar arg) (Just (Var rootSS (copyVar arg)))) allArgs ++
         [ VariableIntroduction rootSS tcoDone (Just (BooleanLiteral rootSS False))
         , VariableIntroduction rootSS tcoResult Nothing
-        ] ++
-        map (\arg ->
-          VariableIntroduction rootSS (tcoVar arg) Nothing) allArgs ++
-        [ Function rootSS (Just tcoLoop) allArgs (Block rootSS [loopify js])
+        , Function rootSS (Just tcoLoop) allArgs (Block rootSS [loopify js])
         , While rootSS (Unary rootSS Not (Var rootSS tcoDone))
             (Block rootSS
-              (Assignment rootSS (Var rootSS tcoResult) (App rootSS (Var rootSS tcoLoop) (map (Var rootSS) allArgs))
-              : map (\arg ->
-                  Assignment rootSS (Var rootSS arg) (Var rootSS (tcoVar arg))) allArgs))
+              [(Assignment rootSS (Var rootSS tcoResult) (App rootSS (Var rootSS tcoLoop) (map (Var rootSS . tcoVar) allArgs)))])
         , Return rootSS (Var rootSS tcoResult)
         ]
     where
--- a/src/Language/PureScript/CoreImp/Optimizer/TCO.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer/TCO.hs
@@ -7,6 +7,7 @@ import Data.Text (Text)
 import Data.Monoid ((<>))
 import Language.PureScript.CoreImp.AST
 import Language.PureScript.AST.SourcePos (SourceSpan)
+import Safe (headDef, tailSafe)
 
 -- | Eliminate tail calls
 tco :: AST -> AST
@@ -29,10 +30,11 @@ tco = everywhere convert where
   convert :: AST -> AST
   convert (VariableIntroduction ss name (Just fn@Function {}))
       | isTailRecursive name body'
-      = VariableIntroduction ss name (Just (replace (toLoop name allArgs body')))
+      = VariableIntroduction ss name (Just (replace (toLoop name outerArgs innerArgs body')))
     where
+      innerArgs = headDef [] argss
+      outerArgs = concat . reverse $ tailSafe argss
       (argss, body', replace) = collectAllFunctionArgs [] id fn
-      allArgs = concat $ reverse argss
   convert js = js
 
   collectAllFunctionArgs :: [[Text]] -> (AST -> AST) -> AST -> ([[Text]], AST, AST -> AST)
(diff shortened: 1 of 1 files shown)


matthewleon:
before

```javascript
    var go = function (__copy_b) {
        return function (__copy_xs) {
            var b = __copy_b;
            var xs = __copy_xs;
            var __tco_done = false;
            var __tco_result;
            var __tco_b;
            var __tco_xs;
            function __tco_loop(b, xs) {
                var v = step(xs);
                if (v instanceof Nil) {
                    __tco_done = true;
                    return b;
                };
                if (v instanceof Cons) {
                    __tco_b = op(b)(v.value0);
                    __tco_xs = v.value1;
                    return;
                };
                throw new Error("Failed pattern match at Data.List.Lazy.Types line 117, column 7 - line 119, column 40: " + [ v.constructor.name ]);
            };
            while (!__tco_done) {
                __tco_result = __tco_loop(b, xs);
                b = __tco_b;
                xs = __tco_xs;
            };
            return __tco_result;
        };
    };
```

after:

```
    var go = function (__copy_b) {
        return function (__copy_xs) {
            var __tco_b = __copy_b;
            var __tco_done = false;
            var __tco_result;
            function __tco_loop(b, xs) {
                var v = step(xs);
                if (v instanceof Nil) {
                    __tco_done = true;
                    return b;
                };
                if (v instanceof Cons) {
                    __tco_b = op(b)(v.value0);
                    __copy_xs = v.value1;
                    return;
                };
                throw new Error("Failed pattern match at Data.List.Lazy.Types line 117, column 7 - line 119, column 40: " + [ v.constructor.name ]);
            };
            while (!__tco_done) {
                __tco_result = __tco_loop(__tco_b, __copy_xs);
            };
            return __tco_result;
        };
    };
```

matthewleon:
In the comments of #2822 @paf31 provided the following tricky case:

`rec x f = rec (x + 1) (\y -> f y + x)`

Before:
```javascript
var rec = function (__copy_x) {
    return function (__copy_f) {
        var x = __copy_x;
        var f = __copy_f;
        var __tco_result;
        var __tco_x;
        var __tco_f;
        function __tco_loop(x, f) {
            __tco_x = x + 1 | 0;
            __tco_f = function (y) {
                return f(y) + x | 0;
            };
            return;
        };
        while (!false) {
            __tco_result = __tco_loop(x, f);
            x = __tco_x;
            f = __tco_f;
        };
        return __tco_result;
    };
};
```

After:
```javascript
var rec = function (__copy_x) {
    return function (__copy_f) {
        var __tco_x = __copy_x;
        var __tco_result;
        function __tco_loop(x, f) {
            __tco_x = x + 1 | 0;
            __copy_f = function (y) {
                return f(y) + x | 0;
            };
            return;
        };
        while (!false) {
            __tco_result = __tco_loop(__tco_x, __copy_f);
        };
        return __tco_result;
    };
};
```

paf31:
Looks good at a glance, but could you please add a short comment explaining what the change is and why this is safe? Thanks!

matthewleon:
Not sure how short I can make this, sorry...

There are two changes here, each with its own commit.

Because of JS's scoping rules, when doing TCO, we need to capture variables in two places. 1) Upon invocation, to protect closed-over variables from being overwritten by invocations of partially-applied copies of the function. We do this by performing `__tco_arg = __copy_arg` when the curried function is fully applied. 2) Upon entry into the `while` loop, because scoping is function-level rather than block-level, and the while loop might need to capture variables as well. We do this by invoking a function a function, `__tco_loop(args...)`, that is invoked on each pass through the `while` loop.

In the first commit. https://github.com/purescript/purescript/pull/2831/commits/9481c6aee58c895b055b068e7c7321f86e90b71a, we simply eliminate a redundant, third case of copying that was done upon invocation. The `__copy_args` were also being copied into `args` variables, to no discernible benefit. My personal theory here is that this was probably being done at a time when protection 2), above, was not in place. Then protection 2) was probably implemented, and the implementer forgot that this meant you could get rid of the now-redundant copying above. This is to some degree borne-out by the fact that the variable names within the `__tco_loop` function were shadowing the copied `args` (see the examples above). The code in question goes back to 2014 and has been through a lot of shifting through different files, so it's hard to figure it out for sure.

In the second commit,  https://github.com/purescript/purescript/pull/2831/commits/36d2e9687d1817f59b8686bf5b8493b2460a61f4, we eliminate the copying from 1) for parameters to the final function in the currying chain. These get closed over in 2), so it should be safe to overwrite them. This is the commit that fixes the issue discussed in #2822

PR #2834, opened 2017-04-14
switch core-tests to psc-package. fixes #2830

### LINKED ISSUE #2830: test-everything.sh fails
The script in `core-tests` appears to still be using 0.10 libs.

matthewleon:


core-tests/psc-package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "core-tests",
+  "set": "psc-0.11.1",
+  "source": "https://github.com/purescript/package-sets.git",
+  "depends": [
```
paf31: Could you please add the newer core packages here too. I notice `typelevel-prelude` is missing. Thanks!

core-tests/psc-package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "core-tests",
+  "set": "psc-0.11.1",
```
paf31: I think we can use 0.11.3.

paf31:
Thanks! Maybe we can add this to the test suite later too. We'd need to install `psc-package` though.

core-tests/psc-package.json
matthewleon: is there a convenient list of these somewhere?

paf31: Just those under the `purescript` GitHub organization.

matthewleon:
ready for review again... well, pending a successful build.

paf31:
I'm surprised Travis is working. I thought we would have to download psc-package.

core-tests/test-everything.sh
```diff
@@ -13,21 +13,20 @@ while getopts 'ci' flag; do
...
-yes 1 | bower i
+psc-package update
 
 if [ "$force_recompile" = "true" ] && [ -d "output" ]; then
   echo "Recompiling..."
   rm -r output
 fi
 
-stack exec purs compile 'tests/**/*.purs' 'bower_components/purescript-*/src/**/*.purs'
+stack exec purs compile 'tests/**/*.purs' '.psc-package/psc-0.11.1/*/*/src/**/*.purs'
```
paf31: This could use `psc-package build`.

core-tests/test-everything.sh
```diff
@@ -13,21 +13,20 @@ while getopts 'ci' flag; do
...
 
-stack exec purs docs 'bower_components/purescript-*/src/**/*.purs' > core-docs.md
+stack exec purs docs '.psc-package/psc-0.11.1/*/*/src/**/*.purs' > core-docs.md
```
paf31: You can use `psc-package sources` here.

matthewleon: I would really like to get that to work, but my bash-fu is weak, and I've tried every permutation of `${ $( $[ $@ xargamafoobar` possible to try to get it to work, and it always fails to find the Prelude.

matthewleon: Ah, you mean just in the `docs` line! Got it.

core-tests/test-everything.sh
matthewleon: Is there a way to ensure that `psc-package` uses the version of `purs` that we just compiled?

matthewleon:
Travis does not run this script. See https://github.com/purescript/purescript/issues/2837

core-tests/test-everything.sh
hdgarrood: I think `stack exec` should do that.

matthewleon:
Also, it seems that upgrading to the 0.11.3 package set and adding those other packages makes it so `purs compile` claims it can't find the Prelude. Joy.

core-tests/test-everything.sh
matthewleon: ah good point, will give it a shot now

matthewleon:
Should be good to go again.

matthewleon:
Okay, it's failing to find tons of modules now. There's probably something obvious I'm missing.

core-tests/test-everything.sh
```diff
@@ -13,21 +13,20 @@ while getopts 'ci' flag; do
...
-yes 1 | bower i
+psc-package update
 
 if [ "$force_recompile" = "true" ] && [ -d "output" ]; then
   echo "Recompiling..."
   rm -r output
 fi
 
-stack exec purs compile 'tests/**/*.purs' 'bower_components/purescript-*/src/**/*.purs'
+stack exec psc-package build
```
paf31: I'm confused - why does this one use `stack exec` but the others don't?

matthewleon: At the risk of seeming really dumb, I have to admit I'm not sure what the "one" in your sentence is referring to. That line uses `stack exec` in order to ensure that it runs the recently-compiled version of `purs`, no?

core-tests/test-everything.sh
matthewleon: Sadly, it seems that that doesn't work. Which is a separate issue I'll create.

matthewleon:
Will probably try to solve #2842 and then submit a new PR.

PR #2836, opened 2017-04-14
remove newline from printed custom type errors

### LINKED ISSUE #2827: Newlines appear in custom type errors
See `examples/failing/ProgrammableTypeErrorsTypeString.purs` for an example.

matthewleon:
fixes #2827

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -21,7 +21,6 @@ import           Data.Functor.Identity (Identity(..))
 import           Data.List (transpose, nubBy, sortBy, partition, dropWhileEnd)
 import           Data.Maybe (maybeToList, fromMaybe, mapMaybe)
 import           Data.Ord (comparing)
-import           Data.String (fromString)
 import qualified Data.Map as M
 import qualified Data.Text as T
 import           Data.Text (Text)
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -1318,7 +1319,7 @@ toTypelevelString t = (Box.text . decodeStringWithReplacement) <$> toTypelevelSt
   toTypelevelString' :: Type -> Maybe PSString
   toTypelevelString' (TypeLevelString s) = Just s
   toTypelevelString' (TypeApp (TypeConstructor f) x)
-    | f == primName "TypeString" = Just $ fromString $ prettyPrintType x
+    | f == primName "TypeString" = fromString <$> (initMay . Box.render $ typeAsBox x)
```
paf31: Using `Box.render` might mess up indentation, and I'm not sure `initMay` is always what we want here.

matthewleon: `prettyPrintType` was already calling `Box.render`. But yes, it's a hack. I'll think about it more.

matthewleon: All the pretty printing operations rely on the box model. Are you suggesting coming up with an alternative pretty printer for types?

paf31: Then that should be using `typeAsBox` actually.

matthewleon: A bit confused by that comment. What should be using `typeAsBox`?

paf31: Instead of `fromString . prettyPrintType` on the line above.

paf31: In fact, the whole thing should return a `Box`.

matthewleon: Sounds good to me. Will update it asap.

matthewleon:
This should be much cleaner.

PR #2846, opened 2017-04-15
Add proxies

### LINKED ISSUE #2839: Proxy syntax
Discussed some of this on Slack with Phil and Nathan. This is a way to make proxies easier to use without going full polykinds.

New type syntax for proxies. `Proxy` will no longed be a user-defined type, but rather a language construct.

    <type> ::= 'Proxy' <type>
             | …

![kindinf]

New expression syntax for proxy literals. This syntax comes from the recent [GHC syntax for type arguments][ghc], however we prefer proxies so that the order of nested universal quantifiers stays insignificant.

    <expr> ::= '@' <type>
             | …

![typeinf]

Code generation of such expressions may just be `{}` or `null` or something. The values can't be inspected so it doesn't matter what their JS representation is.

[ghc]: https://ghc.haskell.org/trac/ghc/wiki/TypeApplication
[kindinf]: http://mathurl.com/n7set37.png
[typeinf]: http://mathurl.com/mkbnlyd.png

rightfold:
Fixes #2839.

---
This change is [Reviewable](https://reviewable.io/reviews/purescript/purescript/2846)

### DIFF
--- a/examples/passing/KindedType.purs
+++ b/examples/passing/KindedType.purs
@@ -15,10 +15,10 @@ f s = s
 
 test2 = f "test"
 
-data FProxy (f :: Type -> Type) = FProxy
+data Proxy (f :: Type -> Type) = Proxy
 
-test3 :: FProxy Array
-test3 = FProxy
+test3 :: Proxy Array
+test3 = Proxy
 
 type Test (f :: Type -> Type) = f String
 
--- a/examples/passing/KindedType.purs
+++ b/examples/passing/KindedType.purs
@@ -15,10 +15,10 @@ f s = s
 
 test2 = f "test"
 
-data Proxy (f :: Type -> Type) = Proxy
+data FProxy (f :: Type -> Type) = FProxy
 
-test3 :: Proxy Array
-test3 = Proxy
+test3 :: FProxy Array
+test3 = FProxy
 
 type Test (f :: Type -> Type) = f String
 
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -575,11 +575,10 @@ reservedPsNames = [ "data"
                   , "false"
                   , "in"
                   , "where"
-                  , "proxy"
                   ]
 
 reservedTypeNames :: [Text]
-reservedTypeNames = [ "forall", "where" ]
+reservedTypeNames = [ "forall", "where", "proxy" ]
 
 -- |
 -- The characters allowed for use in operators
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -578,7 +578,7 @@ reservedPsNames = [ "data"
                   ]
 
 reservedTypeNames :: [Text]
-reservedTypeNames = [ "forall", "where", "proxy" ]
+reservedTypeNames = [ "forall", "where" ]
 
 -- |
 -- The characters allowed for use in operators
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -575,6 +575,7 @@ reservedPsNames = [ "data"
                   , "false"
                   , "in"
                   , "where"
+                  , "proxy"
                   ]
 
 reservedTypeNames :: [Text]
--- a/src/Language/PureScript/Pretty/Types.hs
+++ b/src/Language/PureScript/Pretty/Types.hs
@@ -135,7 +135,7 @@ matchTypeAtom tro@TypeRenderOptions{troSuggesting = suggesting} =
         | otherwise = Just $ text $ T.unpack name ++ show s
       match REmpty = Just $ text "()"
       match row@RCons{} = Just $ prettyPrintRowWith tro '(' ')' row
-      match (ProxyType t) = Just $ text "proxy " <> typeAtomAsBox t
+      match (ProxyType t) = Just $ text "@" <> typeAtomAsBox t
       match (BinaryNoParensType op l r) =
         Just $ typeAsBox l <> text " " <> typeAsBox op <> text " " <> typeAsBox r
       match (TypeOp op) = Just $ text $ T.unpack $ showQualified runOpName op
--- a/src/Language/PureScript/Pretty/Types.hs
+++ b/src/Language/PureScript/Pretty/Types.hs
@@ -37,9 +37,6 @@ import Text.PrettyPrint.Boxes hiding ((<+>))
 
 -- TODO(Christoph): get rid of T.unpack s
 
-proxyAsBox :: Box -> Box
-proxyAsBox ty = text "proxy " <> ty
-
 constraintsAsBox :: TypeRenderOptions -> Constraint -> Box -> Box
 constraintsAsBox tro con ty =
     constraintAsBox con `before` (" " <> text doubleRightArrow <> " " <> ty)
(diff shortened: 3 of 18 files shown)


src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -388,6 +388,10 @@ infer' (IfThenElse cond th el) = do
 infer' (Let ds val) = do
   (ds', val'@(TypedValue _ _ valTy)) <- inferLetBinding [] ds val infer
   return $ TypedValue True (Let ds' val') valTy
+infer' (Proxy ty) = do
```
rightfold: Not sure if this branch is correct. Also, I think this could be the source of the infinite loop.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -388,6 +388,10 @@ infer' (IfThenElse cond th el) = do
...
+  _ <- kindOfWithScopedVars ty
```
paf31: I think we can use `kindOf` here instead, possibly.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -388,6 +388,10 @@ infer' (IfThenElse cond th el) = do
...
+  ty' <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< replaceTypeWildcards $ ty
+  return $ TypedValue False (Proxy ty') (ProxyType ty')
```
paf31: `TypedValue True` here?

paf31:
Thanks! 

A few comments:

- I think you need a new clause in `unifyTypes`.
- I would prefer to not overload the type constructor notation, since it's not a regular type constructor, and to use something like a keyword instead, such as `proxy T`.

rightfold:
- Ok
- I agree

src/Language/PureScript/TypeChecker/Types.hs
paf31: This doesn't matter too much, but I use this boolean to indicate "this should be checked", and then "this _was_ checked", so it'd be good to be consistent is all.

examples/passing/KindedType.purs
```diff
@@ -15,10 +15,10 @@ f s = s
 
 test2 = f "test"
 
-data Proxy (f :: Type -> Type) = Proxy
+data FProxy (f :: Type -> Type) = FProxy
 
-test3 :: Proxy Array
-test3 = Proxy
+test3 :: FProxy Array
+test3 = FProxy
```
LiamGoodacre: If we've switched from `Proxy` to `proxy` then we don't need this change?

rightfold: Indeed. :)

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -575,6 +575,7 @@ reservedPsNames = [ "data"
                   , "false"
                   , "in"
                   , "where"
+                  , "proxy"
```
paf31: Can this go under `reservedTypeNames`?

paf31:
A couple more thoughts:

- Don't forget to update the types pretty printer in `Docs`/`Publish`, otherwise Pursuit will crash on any modules containing `proxy` types.
- Perhaps `RowCons` can use `proxy` to simulate poly kinds now, like @natefaubion has done with `FProxy` in his polymorphic variants implementation.

natefaubion:
> Perhaps RowCons can use proxy to simulate poly kinds now, like @natefaubion has done with FProxy in his polymorphic variants implementation.

That was one question I had, whether I could write something like this:

```purescript
prj
  ∷ ∀ sym f a r1 r2
  . RowCons sym (proxy f) r1 r2
  ⇒ IsSymbol sym
  ⇒ proxy sym
  → VariantF r2 a
  → Maybe (f a)
prj p = on p Just (const Nothing)
```

Note the use of `proxy` in `RowCons`.

paf31:
> That was one question I had, whether I could write something like this:

I think so, because it's not like `RowCons` uses any value with that type, it's a purely type-level function. So there doesn't need to be an actual polykinded type anywhere.

garyb:
What was the reason for not using the symbol at the type level too?

paf31:
Sorry, which symbol again? :)

garyb:
So it's `@"foo"` at the value level, but `proxy "foo"` at the type level currently?

garyb:
Bump for the `@` vs `proxy` question, I know at least @kRITZCREEK, @natefaubion, and me don't love that it's different between types and values. :smile:

hdgarrood:
I'm also a bit confused about that, I think it would be nice to use the same syntax for the type and the value if possible.

paf31:
I'd be fine with `@` in types.

rightfold:
I'm fineish with `@a` being the type syntax for "proxy of `a`". However, usually, the syntax of a value and the syntax of the type of that value do not coincide.

paf31:
Travis says

```
  tests/TestCompiler.hs:218: 
  1) Passing examples 'DuplicateProperties.purs' should compile and run without error
       Error found:
       in module Main
       at /home/travis/build/purescript/purescript/examples/passing/DuplicateProperties.purs line 6, column 1 - line 6, column 55
       
         Type variable proxy is undefined.
       
       while checking the kind of proxy     
                                    ( x :: a
                                    | r     
                                    )       
                                  -> proxy r
       in value declaration subtractX
       
  tests/TestCompiler.hs:218: 
  2) Passing examples 'FunctionalDependencies.purs' should compile and run without error
       Error found:
       in module Main
       at /home/travis/build/purescript/purescript/examples/passing/FunctionalDependencies.purs line 14, column 1 - line 14, column 75
       
         Type variable proxy is undefined.
       
       while checking the kind of Append a b c => proxy a -> proxy b -> proxy c
       in value declaration appendProxy
       
```

paf31:
> with this change, this PR will no longer be a breaking change.

I think we should include this in 0.11.5 then, once the tests pass of course.

natefaubion:
Isn't it still a breaking change because @ can't be used as an operator? FWIW I think it was always dubious due to @ in patterns.

paf31:
Yes, actually, that's true. This currently works, for example:

```purescript
data Test a b = Test a b

infix 4 type Test as @

x :: forall a. a @ a -> a
x (Test a _) = a
```

paf31:
Now I guess the question is, would we rather reserve `@` as a type operator, or `proxy` as a type argument name? I would actually lean towards reserving `proxy`, since we can always pick some other name, but `@` seems like it might be handy as a type operator for some things.

garyb:
Using `proxy` is "more breaking" than `@` given there were numerous examples of its existing usage up there ^, but I don't feel strongly about the syntax for it either way... although I think maybe the fact that a single `@` is actually allowed as an operator currently is a bit weird, given that it doesn't work in patterns:

``` purescript
data Test a b = Test a b

infix 4 type Test as @
infix 4 Test as @

x :: forall a. a @ a -> a
x (a @ _) = a -- Type error, Test a0 a0 /~ a0
```

garyb:
Actually, given that, I think I do prefer `@`. It would be weirder that there's a single reserved lname in the types, vs a magic prefix type operator that is already somewhat-reserved at the value level.

paf31:
Alright then, `@` it is. Thanks!

garyb:
I guess the above parse does mean you can't use a proxy literal in a pattern, but then you don't really need to since they're singly-inhabited.

rightfold:
@garyb the list of uses of `proxy` above are values, not types. `proxy` would only be a keyword in types.

src/Language/PureScript/Pretty/Types.hs
```diff
@@ -135,6 +135,7 @@ matchTypeAtom tro@TypeRenderOptions{troSuggesting = suggesting} =
         | otherwise = Just $ text $ T.unpack name ++ show s
       match REmpty = Just $ text "()"
       match row@RCons{} = Just $ prettyPrintRowWith tro '(' ')' row
+      match (ProxyType t) = Just $ text "proxy " <> typeAtomAsBox t
```
NasalMusician: `"@"`?

paf31:
🎆 Thanks!

PR #2848, opened 2017-04-15
Switch test suite to use tasty

kRITZCREEK:
*This is a suggestion, feel free to shoot it down*

This switches the test suite to use [tasty](http://documentup.com/feuerbach/tasty) as its test runner.

The primary benefit this has right now, is that we can run individual tests by running eg: `stack test --test-arguments="-p ide"`. We can also switch off things like colored output with environment variables, which is nice in a CI setting.

I've seen some strange artifacts with Powershell and the compiler test suite: 
` getConsoleCursorInfo: invalid argument (Invalid handle.)`
I can avoid the error with setting `--color never` but it's not really a solution. (I decided I'll check what happens on AppVeyor before researching further)

Still missing docs and publish tests, since these use a custom test runner mechanism and I couldn't figure out how to make them use hspec in the limited time I invested.


hdgarrood:
Looks good to me! I'll happily continue this for publish and docs when I get a moment.

paf31:
👍 I like it

kRITZCREEK:
I don't know what's happening there, but as long as the test suite can't be run properly on Windows this is not ready to go in...

paf31:
Appveyor doesn't seem to complain?

If you merge with `master` then Travis should start working too.

kRITZCREEK:
`tasty` detects that it's not running inside a tty and turns off colors by default in CI environments. The problem is that if someone with a windows machine clones the repository and runs `stack test` his tests will fail with the current state of this PR.

kRITZCREEK:
Okay, so after a long time of letting this sit... I decided to look into it again, and it turns out that the problems on Windows can be fixed by not using `silently` on the compiler output in the compiler tests. Instead I set the `progress` function on the `MakeActions` to not output anything. Was there anything else the `silently` was supposed to silence?

I've added a bit of documentation on how to run the individual tests to `CONTRIBUTING.md` but there is probably a place forsome more good example invocations.

kRITZCREEK:
Thanks :)

parsonsmatt:
Eh, can we reconsider this?

Test output is a lot worse with this. Specifically, failing test messages are entirely muted, so we don't get anything more than a red `FAIL`. There's no way to hide the successful tests (`--hide-successes` only hides the green `OK` thing, which doesn't reduce noise at all). `tasty` isn't reproducing the failing test cases, so it's necessary to scroll up 100s of lines to see what tests failed. It seems to be running slower now, as well.

kRITZCREEK:
From one the recent CI runs: `All 767 tests passed (196.29s)`

@garyb and me also see the failing test output just fine (that's how we found the missing error spans)

I'm totally fine with switching back to pure hspec if someone does the work of pulling the different test suites together under a single invocation of `hspec` so we can use the pattern syntax and hide successes etc.

parsonsmatt:
Ah, I'm seeing the failure messages now. Still, this seems to be worse than `hspec` to me, so I'm not entirely sure what motivated the change to `tasty`. If it's just pulling the test suites into a single `hspec` invocation then I can put that together.

kRITZCREEK:
Since this is merged I've been using the ability to only run the `ide` testsuite (which was possible before, but the invocation was hideous and I always forgot it), and more importantly being able to run specific tests from the passing or failing examples like `stack test --ta="-p **1110.purs*"`.

I've previously searched through the docs for hspec, but it seems like the cmdline pattern syntax only allows to restrict the tests to be run to source files not individual tests.

Also I'm not sure in what sense the test output is a lot worse with this, since it print the test names just as hspec.

PR #2851, opened 2017-04-17
display help text by default

matthewleon:
When `purs` is invoked with no command line arguments, this change shows the help text.


matthewleon:
For whatever reason, only one of the Travis builds fails. Hm.

matthewleon:
Ah, it fails on all the PRs. Joy.

PR #2852, opened 2017-04-17
Turn off coveralls upload for now

paf31:
Seems `stack-hpc-coveralls` doesn't like the new `http-client` API, or something like that.


paf31:
I'm going to merge this, and try to get a 0.11.4 release out today.

hdgarrood:
:+1:

Issue #2853, opened 2017-04-18
.cabal file has invalid url to issue tracker


bergmark:
https://hackage.haskell.org/package/purescript-0.11.4 points to purescript.git/issues. I'm assuming it has something to do with hpack, is `.git` in the package.yaml for a reason?

seanwestfall:
I was trying to use hpack for this PR https://github.com/purescript/purescript/pull/2849 but can't get it to update the cabal file. Is this related?

hdgarrood:
@seanwestfall I don't think so, #2854 fixes this.

kRITZCREEK:
fixed in #2854

PR #2854, opened 2017-04-19
Update package.yaml

sol:
Currently the bug tracker URL is broken on Hackage.  This change fixes this.


hdgarrood:
Thanks! Would you mind checking in the updated cabal file? We keep it checked in for the sake of people who prefer to use cabal-install.

sol:
done

hdgarrood:
Fixes #2853

PR #2857, opened 2017-04-19
Bumps lower bound for directory

### LINKED ISSUE #2856: lower bound on `directory` inaccurate
The .cabal file says `>= 1.2`, whereas it should rather say `directory >= 1.2.3`. This was already reported 
 in #2816 but despite that, 0.11.3 and 0.11.4 got released still with the inaccurate lower bound 

/cc @bergmark

kRITZCREEK:
fixes #2856

The PR for the .cabal file in #2816 coincided with the switch to `hpack` and so it seems like the change was overwritten unnoticed.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -46,7 +46,7 @@ dependencies:
   - containers
   - data-ordlist >=0.4.7.0
   - deepseq
-  - directory >=1.2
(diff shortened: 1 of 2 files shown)


paf31:
Thanks!

PR #2859, opened 2017-04-22
Fix kind signature for RowCons typeClassData

tslawler:
I haven't looked too deeply into the compiler, but decided to take a peek at the new RowCons and Union stuff -- noticed this seemed a bit off.


paf31:
Thanks!

@hdgarrood I think we use `primTypes` only during type checking actually, since we don't define instances for these classes by hand.

paf31:
@tslawler Next time you make a PR, could you please add yourself to `CONTRIBUTORS.md` as well, thanks!

PR #2863, opened 2017-04-23
Desugar let properly when generating docs

### LINKED ISSUE #2861: Purs publish fails when using let-patterns
This function will compile fine with `purs compile`
```purescript
case_ ∷ ∀ a. Variant () → a
case_ r = unsafeCrashWith error
  where
  Tuple tag _ = unsafeCoerce r
  error = "Data.Variant: pattern match failure [" <> tag <> "]"
```

But when calling `publish`:
```
published.
Details:
  Compile error:
    Error found:
    in module Data.Variant
    at ./src/Data/Variant.purs line 88, column 54 - line 88, column 57

      Unknown value tag


```

paf31:
Fixes #2861

### DIFF
--- a/examples/docs/src/Desugar.purs
+++ b/examples/docs/src/Desugar.purs
@@ -1,0 +1,8 @@
+module Desugar where
+
+data X a b = X a b
+
+test :: forall a b. X (a -> b) a -> b
+test x =
+  let X a b = x
+  in a b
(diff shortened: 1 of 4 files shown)


hdgarrood:
Maybe we should just run docs on all of the examples we use in the normal test suite too, and just check it doesn't crash? That should probably come in a separate PR though.

paf31:
Sounds like a good idea to me. Thanks for the review!

PR #2864, opened 2017-04-24
Disallow invalid newtype-derived instances

### LINKED ISSUE #2339: Improve error messages for types of same name
If there are two types of the same name that are imported. For ex:

```purescript
import Foo (Bar(Norf)) as Foo
import Baz (Bar(Norf)) as Baz

qux :: Foo.Bar -> String
qux _ = "blah"

norf = qux Baz.Norf
```

It will throw a type error saying something like it could not match type `Bar` with type `Bar`, but it would be more useful for it to say that it could not match type `Baz.Bar` with type `Foo.Bar`.

### LINKED ISSUE #2399: Derived newtype instances can give wrong superclass through entailment
If you manually declare an `Apply` instance, and then `derive newtype` an `Applicative` instance, you can get the wrong instance through entailment. Because newtype deriving is just a dictionary coercion, it will give the "original" `Apply` if you ask for an `Applicative` rather than the declared instance.

paf31:
Fixes #2399


LiamGoodacre:
:) I think you mean #2399

paf31:
I realized this won't work for something like `MonadWriter`, with a `Monoid` superclass constraint. So I think the condition must be that _if_ a superclass constraint mentions the newtype, all type arguments must match, and we should only check such constraints for derived instances.

paf31:
This is ready to review again. I've made this into a warning, so it's technically non-breaking, and we only consider those cases where the superclass also mentions the newtype in the last type argument, which covers most practical cases like the MTL classes. This is still only a guess though, and it's not a complete check.

paf31:
Any thoughts on this?

hdgarrood:
Could you explain why `MissingNewtypeSuperclassInstance` is a warning and not an error?

hdgarrood:
Oh, so it's non-breaking. :)

hdgarrood:
I'm not sure I fully understand why the check fails to be complete; do you have an example which is invalid but the check doesn't currently flag as a problem?

paf31:
Well right now, we only check the superclass if the newtype appears in the last argument.

paf31:
@hdgarrood Here's an example which doesn't warn, but which probably should:

```purescript
module Main where

class Su a b
class Su a b <= Cl b a

newtype S = S String

instance suString :: Su String String
instance clString :: Cl String String

derive newtype instance clS :: Cl String S
```

If we change the definition of `Cl` to 

```purescript
class Su a b <= Cl b a
```

then it warns as expected.

Perhaps we should disallow derived newtype instances if the superclass constraints don't "persist the last type argument"? Of course, that would be breaking, so it'd have to wait until 1.0.

paf31:
Having thought about it a bit more, I suggest the following:

1. For each superclass constraint:
1. If the superclass constraint mentions the last type argument of the subclass (i.e. the newtype argument), then check where that argument appears:
  - If it occurs only in the last argument of the superclass constraint, unmodified, then check whether the superclass was derived for the same newtype. If not, warn with `MissingNewtypeSuperclassInstance`.
  - If it occurs in any other configuration, warn with some new `UnverifiableNewtypeSuperclassConstraint` warning.

Both of these should become errors before 1.0.

@garyb @hdgarrood @LiamGoodacre Thoughts?

hdgarrood:
That sounds reasonable to me, although it's not entirely clear to me what is meant by "unmodified" in

> If it occurs only in the last argument of the superclass constraint, unmodified

paf31:
@hdgarrood I mean that the type variable which represents the newtype (i.e. the last argument), let's call it `a`, only appears as `a` in the last argument of the superclass as well. i.e. you couldn't have something like

```purescript
class Su a
class Su (Array a) <= Cls a
```

or

```purescript
class Su a b
class Su b a <= Cls a b
```

and derive `Cls` for a `newtype`.

hdgarrood:
Ahh, yes, I see. Thank you!

paf31:
I've implemented my proposal now, but it has the unfortunate problem that some things we actually want to be verifiable are not verifiable any more. Namely, `MonadTell w m` as a superclass of `MonadWriter w m` causes a problem.

I'm tempted to say we can allow extra arguments like these in superclasses if they are all determined by the newtype argument via functional dependencies on the superclass. That way, it wouldn't have been possible to derive a different superclass instance than the one wrapped up in the derived instance for the subclass.

paf31:
This is ready for review again. I've added the fundep check mentioned above.

@LiamGoodacre Maybe you'd be interested in this? Would you mind having a look over the new code, please?

paf31:
If someone could just please scan the code for glaring mistakes, I'll get this merged in.

Also, I'm thinking of making a 1.0 or 0.12.0 branch soon, since we have a lot of open PRs which we should have available for testing but which won't go into a 0.11.x release.

paf31:
Thanks @kRITZCREEK!

PR #2867, opened 2017-04-24
Apply relative paths during pretty printing of errors

### LINKED ISSUE #2855: [purs ide] Sourcespans are relative instead of absolute since 0.11
This breaks all the editors go-to-definition. It's most likely just a matter of prepending the current working directory of the server when extracting the source spans from the ASTs.

The error reporting is also affected by this, but that's because the JSON format of the compiler errors changed, so we can't really fix it in purs ide.

EDIT: Introduced in #2707

Question @garyb @paf31 :
If this was done solely for prettier error messages, wouldn't it be better to trim the paths in the pretty printer rather than during parsing?

kRITZCREEK:
Fixes #2855


kRITZCREEK:
Seems to work fine in my manual testing.

kRITZCREEK:
```
bower ECONFLICT     Unable to find suitable version for purescript-eff
tests: callProcess: /home/travis/.nvm/versions/node/v4.1.2/bin/node "node_modules/.bin/bower" "--allow-root" "install" "--config.interactive=false" (exit 1): failed
``` 😢

hdgarrood:
Oh dear, I get the same thing if I delete my `tests/support/bower_components` directory in my local copy. I guess this is because bower doesn't do any backtracking. I think what's happening is that the latest release of `type-equality`, namely 2.1.0, added a dependency on `eff ^3.1.0`, but the tests bower.json file specifies 3.0.0 exactly. We lock down the immediate dependencies but not transitive ones, and at the moment we only depend on `type-equality` transitively via `typelevel-prelude`, so since this new version of `type-equality` was published bower has started selecting it, causing the clash.

I think the fix should be to update `eff` to 3.1.0 in `tests/support/bower.json`.

hdgarrood:
Oh and we should probably consider adding an explicit dependency on `typelevel-prelude` to help avoid this situation in the future.

kRITZCREEK:
I just bumped the dependency in `bower.json` and had to make lots of changes to the psci tests to make them pass again, because of various module additions.

Could we make these tests less flaky by writing checked in test-modules or locking down all the transitive dependencies please? /cc @noraesae 

Other than that, this should pass CI again :)

hdgarrood:
I'm in favour of either of those options. I think I'd prefer locking down all transitive deps.

hdgarrood:
I also considered inferring the list of support modules before the tests run rather than hardcoding it at one point. I don't remember why i didn't do that, maybe I just didn't get around to it. I'd like to find a way to not need that hardcoded list eventually.

paf31:
👍 Ship it!

PR #2873, opened 2017-05-04
Allow aeson 1.1

felixonmars:
It seems to work fine here.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -649,7 +649,7 @@ flag release
 
 library
     build-depends:
-          aeson >=1.0 && <1.1
+          aeson >=1.0 && <1.2
         , aeson-better-errors >=0.8
         , ansi-terminal >=0.6.2 && <0.7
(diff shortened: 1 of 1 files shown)


hdgarrood:
We've been bitten in the past by making changes like this which seemed to work without really testing properly. We really need to check the aeson release notes before doing this, e.g. to make sure that no instances we are using have been changed.

paf31:
Now might be a good time to add those JSON unit tests we discussed recently.

hdgarrood:
If we were using derived instances and checking a hash based on the data type definition, and Aeson's TH instance derivation changed, a bug could still slip through that way though surely?

kRITZCREEK:
After reading through the aeson's changelog I didn't see anything breaking for us. Every new addition defaults to backwards compatible behaviour.

paf31:
Thanks @kRITZCREEK

PR #2876, opened 2017-05-07
[purs ide] Decodes source files as UTF8 when parsing out the imports

kRITZCREEK:
The previous change to this used the default Text's readFile, which decodes as UTF16 and thus messed up unicode symbols when adding imports. Encodings & cross platform issues, fun times!

### DIFF
--- a/src/Language/PureScript/Ide/Util.hs
+++ b/src/Language/PureScript/Ide/Util.hs
@@ -38,9 +38,8 @@ import           Protolude                           hiding (decodeUtf8,
 import           Control.Lens                        hiding ((&), op)
 import           Data.Aeson
 import qualified Data.Text                           as T
-import qualified Data.Text.IO                        as TIO
 import qualified Data.Text.Lazy                      as TL
-import           Data.Text.Lazy.Encoding             (decodeUtf8, encodeUtf8)
+import           Data.Text.Lazy.Encoding             as TLE
 import qualified Language.PureScript                 as P
 import           Language.PureScript.Ide.Error       (prettyPrintTypeSingleLine, IdeError(..))
 import           Language.PureScript.Ide.Logging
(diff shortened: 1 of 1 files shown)


hdgarrood:
:+1: although IIRC the encoding that `Data.Text.IO.readFile` assumes depends on environment variables, so isn't necessarily UTF-16.

kRITZCREEK:
Turns out this doesn't actually do what it says on the label... I'll continue to investigate. I'm very close to manually search replacing `\r\n` with `\n`...

src/Language/PureScript/Ide/Util.hs
```diff
@@ -152,4 +154,9 @@ ideReadFile = ideReadFile' readUTF8FileT
 -- rather than @TextMode@ line endings get mangled on Windows otherwise. This is
 -- irrelevant for parsing, because the Lexer strips these additional @\r@'s.
 ideReadTextFile :: (MonadIO m, MonadError IdeError m) => FilePath -> m Text
-ideReadTextFile = ideReadFile' TIO.readFile
+ideReadTextFile =
+  ideReadFile' (map TE.decodeUtf8 . readFileNonBinary)
+  where
+    readFileNonBinary fp = withFile fp ReadMode $ \h -> do
+      hSetBinaryMode h False
```
hdgarrood: Should this be `hSetBinaryMode h True`? The comment leads me to think that it should be.

kRITZCREEK: Unfortunately that doesn't work. It seems hGetContents just ignores Binary/TextMode.

kRITZCREEK:
@hdgarrood this is not what you initially approved, would you mind taking another look?

hdgarrood:
I trust your judgement on the implementation but is the comment now out of date?

kRITZCREEK:
Adjusted the comment, thanks!

Issue #2792, opened 2017-03-28
[purs ide] Add a new namespace filter


kRITZCREEK:
This is a new feature for `purs ide` which is suitable as a first contribution. In this issue I've written down _why_ we'd want to add a namespace filter and _how_ you could go about implementing it. A good point to get started is to read the `CONTRIBUTING.md` file in the root of this repository. If you want to pick this as your first contribution to `purs ide` I'd be happy to answer any questions or help you if you run into problems. Just ask here in the issue or on IRC at #purescript where I'm usually available during the European day.

## Motivation

The idea is that we'd like to be able to specify to which namespaces `purs ide` should limit the results of a query. Here are some future features this would enable in editors:

- Split the _type of_  command into _type of_ or _kind of_, to let the user decide which of the two he wanted, in cases where type and constructor name overlap. With the new filter in place, _type of_  would only allow declarations in the **value namespace** and the _kind of_  command would limit its query results to the **type namespace**.
- An editor could determine that the user is editing a specific syntax construct and limit the completions to the correct namespace. For example, if we determined (eg. with a regex) that the user is editing an instance head we know that we only need to allow completions from the **type or kind namespace**.
- Allowing the editors to specify that they only want to import a type constructor, if it was determined that the to-be-imported identifier was at the type level. Right now, for example for newtypes which expose their constructor, it's impossible to _only_ import the type.

## Implementation

`purs ide` provides completions and type information through a set of commands. Through a JSON protocol editors can query the server. These queries are built up from a list of _filters_ and an optional _matcher_. As the issue title already says, implementing this feature requires that we add a new _filter_. To do so we'll need to take two main steps:

1. Implementing the filtering logic
2. Parsing the new filter as JSON

### Implementing the filtering logic

We'll need to implement a new function `namespaceFilter :: NonEmpty IdeNamespace -> Filter`, where the `NonEmpty` type is from `Data.List.NonEmpty`. It should be located in the `Language.PureScript.Ide.Filter` module, where all the other `purs ide` filters live. Inside this filter we'll need to only keep the declarations that live in the correct namespace.

For example, if we were filtering with `IdeNSValue :| [IdeNSType]` as the argument to `namespaceFilter` we'd need to remove all the `IdeKindDecl` declarations from the filtered modules. One thing to note here, is that the filters are also supposed to remove modules that contain no more declarations after the filter has been applied.

[...]

PR #2878, opened 2017-05-10
[purs ide] extract namespace adt

kRITZCREEK:
This is just a minor refactoring in preparation for #2792.

### DIFF
--- a/src/Language/PureScript/Ide/Filter.hs
+++ b/src/Language/PureScript/Ide/Filter.hs
@@ -26,11 +26,12 @@ import           Protolude                     hiding (isPrefixOf)
 
 import           Data.Aeson
 import           Data.Text                     (isPrefixOf)
-import qualified Language.PureScript           as P
 import           Language.PureScript.Ide.Types
 import           Language.PureScript.Ide.Util
+import qualified Language.PureScript           as P
 
-newtype Filter = Filter (Endo [Module]) deriving(Monoid)
+newtype Filter = Filter (Endo [Module])
+  deriving (Monoid)
 
 type Module = (P.ModuleName, [IdeDeclarationAnn])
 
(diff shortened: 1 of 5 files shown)


paf31:
👍 LGTM

hdgarrood:
Would it make sense to unify this with the equivalent type in Language.PureScript.Docs?

kRITZCREEK:
@hdgarrood but where would the type live? I guess in one of the `AST` modules?

hdgarrood:
I think Names might be more appropriate? Is the Ide code not able to depend on Docs at the moment though?

kRITZCREEK:
I think I'll leave this PR as is for now. I think it totally makes sense to have this Enum somewhere in the compiler, but it's probably not only Docs and Ide that would use it.

kRITZCREEK:
```
[ 2 of 14] Compiling Version          ( app\Version.hs, .stack-work\dist\ca59d0ab\build\purs\purs-tmp\Version.o )
Segmentation fault/access violation in generated code
```

Doesn't sound good, but seems to be related to getting the git revision on the AppVeyor instance? Definitely isn't caused by this PR.

hdgarrood:
Oh dear :/ Might be worth trying to repro with a local Windows VM, and perhaps seeing what happens with GHC HEAD too? I found this patch which was recently merged https://ghc.haskell.org/trac/ghc/ticket/13108#no1 which slightly improves this error message we're seeing.

kRITZCREEK:
The tests works just fine on my local Windows machine. Must be something AppVeyor specific.

hdgarrood:
I suppose there's a possibility it's due to us running `stack` slightly differently locally and on AppVeyor? Eg, does it make a difference if you run with `--jobs=1` like CI does?

kRITZCREEK:
Still doesn't break anything on my machine. I'll just merge this PR since it doesn't cause the failures and I'll open a separate issue for the CI failures?

hdgarrood:
Sounds good to me.

PR #2882, opened 2017-05-11
[purs ide] Simplify state type

kRITZCREEK:
This merges Stage2 into Stage3 and renames the state types into `IdeFileState` and `IdeVolatileState`. While these names are definitely better than their predecessors I'm open to suggestions for better names....

The reasoning for the names is as follows:

- `IdeFileState` contains all the state that directly maps to an entity in the file system

- `IdeVolatileState` contains all the state we derive from IdeFileState and needs to be fully reloaded whenever the file system changes.

I also added some documentation while I was at it.

To give a bit of context, I'm currently working on documenting the architecture and some of the design decision made inside `purs ide` to make it easier to navigate and contribute to. So expect more refactoring/documentation/cleanup PRs.

### DIFF
--- a/app/Command/Ide.hs
+++ b/app/Command/Ide.hs
@@ -120,7 +120,7 @@ command = Opts.helper <*> subcommands where
 
     unless noWatch $
       void (forkFinally (watcher polling logLevel ideState fullOutputPath) print)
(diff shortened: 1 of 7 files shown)


kRITZCREEK:
Ping. Could I get someone to take a look? It's primarily a refactoring.

kRITZCREEK:
@hdgarrood Thanks for reviewing!

PR #2884, opened 2017-05-11
Converts \r\n into \n after reading files as ByteStrings

### LINKED ISSUE #2881: PSCI test failures on Windows
Not really sure what's causing this, but when running the testsuite on a fresh Windows install I get these failures:

```
evalTests
(line 1, column 1):
unexpected end of input
expecting "import", declaration, expression or "let"
  Multiline.purs FAILED [1]
(line 1, column 1):
unexpected end of input
expecting "import", declaration, expression or "let"
  BasicEval.purs FAILED [2]

Failures:

  tests\TestPsci\EvalTest.hs:55:
  1) evalTests Multiline.purs
       uncaught exception: ExitCode (ExitFailure 1)

  tests\TestPsci\EvalTest.hs:55:
  2) evalTests BasicEval.purs
       uncaught exception: ExitCode (ExitFailure 1)
```

They work fine on my other Windows machine, so I'm really not sure what's going on but maybe @noraesae or @hdgarrood have an idea?

One difference is that this machine is running node 7.10, so that might be related.

I tried it with both `cmd.exe` and powershell and got the same result

kRITZCREEK:
Fixes #2881

### DIFF
--- a/src/System/IO/UTF8.hs
+++ b/src/System/IO/UTF8.hs
@@ -3,13 +3,20 @@ module System.IO.UTF8 where
 import Prelude.Compat
 
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Lazy as BSL
+import qualified Data.ByteString.Search as BSS
 import qualified Data.ByteString.UTF8 as UTF8
 import           Data.Text (Text)
 import qualified Data.Text.Encoding as TE
 
+-- | Unfortunately ByteString's readFile does not convert line endings on
+-- Windows, so we have to do it ourselves
+fixCRLF :: BS.ByteString -> BS.ByteString
+fixCRLF = BSL.toStrict . BSS.replace "\r\n" ("\n" :: BS.ByteString)
+
 readUTF8FileT :: FilePath -> IO Text
 readUTF8FileT inFile =
-  fmap TE.decodeUtf8 (BS.readFile inFile)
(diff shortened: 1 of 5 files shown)


paf31:
I assume this has been tested in pscide quite a bit, so 👍 

But we probably should test this quite a bit too.

src/System/IO/UTF8.hs
```diff
@@ -3,21 +3,28 @@ module System.IO.UTF8 where
...
+import qualified Data.ByteString.Lazy as BSL
+import qualified Data.ByteString.Search as BSS
 import qualified Data.ByteString.UTF8 as UTF8
 import           Data.Text (Text)
 import qualified Data.Text.Encoding as TE
 
+-- | Unfortunately ByteString's readFile does not convert line endings on
+-- Windows, so we have to do it ourselves
+fixCRLF :: BS.ByteString -> BS.ByteString
+fixCRLF = BSL.toStrict . BSS.replace "\r\n" ("\n" :: BS.ByteString)
```
paf31: Why not use `T.replace`?

kRITZCREEK: So that I can use it in `readUTF8File` as well, without going through `Text` first.

kRITZCREEK:
what the heck? This time Appveyor passed?

kRITZCREEK:
I'm running this change through a `psc-package verify-set` right now, but I don't see how it would cause any problems.

Issue #2435, opened 2016-11-17
Desugar do-notation into Applicative operations (ApplicativeDo)


zudov:
[The paper](http://research.microsoft.com/en-us/um/people/simonpj/papers/list-comp/applicativedo.pdf) describes many details and all that is easily mapped to purescript.

Essentially the process of desugaring is about splitting the do-notation statements into subsequences (parallel blocks) in such a way that the blocks don't share any binders. Those parallel blocks are then combined with `map`, `apply` and `join`.

That allows to make use of all the applicative benefits while still being able to use do-notation sugar.

It comes at cost of making do-notation desugaring much harder for understanding, though you could still use the simple `bind` desugaring as a mental model until you need to know more. 

This would cause some breaking changes too:

- the imports of `map` and `apply` would have to be added in relevant places;
- the rebindable syntax trick would get more involved. Previously you could just define the `bind`, but now you'd have to also define the `map` and `apply`;
- when `apply`'s semantic differs from `ap`'s the meaning of a program might get changed.

paf31:
I would rather add an `ado` keyword or something like that. The reason is that switching to `Applicative` can change the evaluation order. And if the `Applicative` instance is dodgy, it can change behavior too.

zudov:
> I would rather add an `ado` keyword or something like that.

I agree, with all the possible dodgyness, it would be nicer to have a special syntax form. This would avoid all the compatibility problems and make a release much-much safer.

> The reason is that switching to `Applicative` can change the evaluation order. 

I wouldn't have thought that this would be a problem unless with Haxl-like (or just any dodgy) type whose `Applicative` is parallelized, and `Monad` is sequential. Maybe you have some example to illustrate?
Of course, dodgy `Applicative` instances do exist and we don't want to potentially involve them in normal do-blocks.

The rules used in the paper never explicitly reorder the statements (for getting better parallel-block splitting). I understand that this kind of reordering is not the same as effect/evaluation order that you are talking about (I guess).

> Reordering the statements is only valid in a commutative
monad, where the order of effects is not observable. The Haxl
monad is not commutative, because it supports effects in the form
of exceptions, so reordering statements can change which excep-
tions are reported. In our design, we therefore never reorder com-
putations. We leave for future work the possiblity of allowing re-
ordering for commutative monads.

It might be quite fun to also have intentional reorderings under a commutative monad constraint of some sort...

paf31:
[...]

PR #2889, opened 2017-05-13
Add applicative do notation

rightfold:
Usually we discuss this stuff before PRing but I was bored so here it is anyway. If I've done it totally wrong I'd be glad to hear that, if not that's great! 😅 

Bindings are only in scope in the `pure`'d result and in `let` statements.


hdgarrood:
Previously: https://github.com/purescript/purescript/issues/2435. Seems to have been closed for lack of a proper proposal; having a proof of concept changes that, presumably..?

rightfold:
@hdgarrood I think so, yes. This PR implements exactly what was discussed in that issue. I can write up a proposal that matches it and then we can see how it works out immediately.

rightfold:
I think this documents the behavior of this PR quite well, with rationale for each deviation from GHC (which is quite a lot lol):

# Applicative do-notation

The benefits of applicative do-notation are documented elsewhere. This proposal is about the specific implementation in PureScript. This proposal uses `(<*>)` several times, for ease of reading. The actual implementation does not use these; instead it uses `apply`.

## Differences from GHC

GHC implements applicative do-notation. This proposal adds applicative do-notation to PureScript, with two important differences:

1. The syntax is opt-in rather than applied to every do-block.
2. The implemenation described in this proposal addresses only those computations that use only applicative combinators. To mix in computations that depend on each other's results, one needs to resort to monadic do-notation (which may be nested inside applicative do-notation, and vice versa). A future proposal may extend this implementation to allow the use of `bind`, although that may interfere with the opt-in syntax.

## Syntax

The syntax of applicative do-notation differs from that of monadic do-notation in two aspects.

### Opt-in applicative desugaring

Due to strict evaluation, it is not preferred to transform any `do` expression into an applicative computation whenever the types would allow this. Efficiency and termination could suffer due to the lack of thunking. Consider:

```purescript
f 0 = pure 0
f n = do
  x <- a
  y <- f (n - 1)
  pure $ x + y
```

If this were translated to applicative combinators, `f n = pure (+) <*> a <*> f (n - 1)`, it would not work as expected because the second argument to `(<*>)` is immediately evaluated.

Therefore this proposal adds a new keyword `ado` that complements `do`. Monadic do-notation will remain entirely unchanged.

### No special interpretation of `pure`

It is not desired to interpret application of the `pure` function specially. This would incur problems with higher-order combinators such as `$`:

```purescript
ado
  x <- a
  y <- b
  pure $ x y -- application of ($), not of pure
```

Instead, we use the `in` keyword, as suggested by @ElvishJerricco. This is mandatory and must be and must be only the last statement in the applicative do-block. It can be compared with the `yield` keyword in Scala.

```purescript
ado
  x <- a
  y <- b
  in x y
```

## Desugaring procedure

An applicative do-block is desugared into a series of `apply` applications and one `pure` application. An alternative implementation could use `f <$> ...` instead of `pure f <*> ...`. However, the latter is preferred because `pure` is already necessary when there is only an in statement in the applicative do-block, and so would require fewer names to be in scope in general (`pure` + `apply` vs `pure` + `apply` + `map`).

This desugaring procedure is systematic, according to the following rules:

1. For each bind and value statement, the result expression is wrapped, in order, by a lambda abstraction that receives the result of the statement.
   1. For each bind statement, the parameter of this lambda abstraction is the binder of the bind statement.
   2. For each value statement, the parameter of this lambda abstraction is the null binder.
2. `pure` is applied to the outer lambda abstraction around the result expression.
3. `foldl (\a b -> App (App apply a) b) result binds-and-values`.

Some examples:

```purescript
ado
  in h x
-- desugars into
pure (h x)

------------------------------

ado
  x <- g
  in f x
-- desugars into
pure (\x -> f x) <*> g

------------------------------

ado
  x <- g
  y <- h
  in f x y
-- desugars into
pure (\x y -> f x y) <*> g <*> h

------------------------------

ado
  x <- g
  h
  in f x
-- desugars into
pure (\x _ -> f x) <*> g <*> h

------------------------------

-- and so on
```

## Open issues

- How should we desugar let statements?
- Does the use of `in` cause ambiguity in the presence of let expressions?
- Should we omit `pure` in the case of applicative do-blocks with multiple statements, and use `map` instead of the first `apply`?

paf31:
Thanks for the PR and the detailed explanation!

I have a few comments:

- It seems like `let` can only be desugared if the bound names are only used in the result, not the intermediate computations.
- Can you please add an example where the computations on the RHS of `<-` have different types? That way we can make sure they're being applied in the right order.
- The use of `pure`/`apply` is probably nicer in terms of codegen, but unfortunately means we can't use `ado` with just an `Apply` (we always need a full `Applicative`)

And one random thought:

- It might be nice to have a variant of `ado` (and `do` for that matter) which supports recursion using `Lazy`.

rightfold:
@paf31 Thank you for the comments. I will now comment on them.

 - `let x = y` could be desugared to `x <- pure y`, or could be inlined into the generated lambdas.
 - Yes, I will. The order is already checked by `main` (through `Ref`) but having different types does not harm.
 - That's right, it is not difficult to use `(<$>)` in place of `pure`. It may also be more efficient.

rightfold:
There should be a test case for duplicate bindings, and this should be prohibited.
```
ado
  x <- pure 1
  x <- pure 2
  in x
```

rightfold:
I updated the description above to match the new implementation.

rightfold:
Now it prefers `(<$>)`.

paf31:
I figured out a way to get something like ApplicativeDo using just rebindable syntax:

http://try.purescript.org/?gist=91c8c58419ec2683d46b8d83eb88614c

So we should consider whether it's worth new code in the compiler if we can get 90% of the feature as a library.

To compare the two, this approach requires you to rebind `bind` and `discard` (or use `_ <-`) and the syntax is slightly less nice.

paf31:
@rightfold Could you please rebase this one?

rightfold:
@paf31 On it!

paf31:
Thanks!

PR #2898, opened 2017-05-18
[purs ide] Adds a command to add qualified imports

kRITZCREEK:


kRITZCREEK:
1. Inside a file, the user executes an addImport command while his cursor rests on `List.filter`
2. The editor uses the list imports command to check whether the current file already has an import with that qualifier. If it doesn't:
3. The editor runs a completion command that searches for exact matches on the identifier `filter`.
4a. We only get results from a single module -> insert qualified import for that module as `List`
4b. We get results from multiple modules, the editor asks the user to select a module, just like with the imports for unqualified identifiers

Issue #2874, opened 2017-05-04
Internal error unspecified skolem scope


LiamGoodacre:
Using `forall` in either an instance's or class's constraint such as:

```purescript
class T a b | a -> b
instance tT :: T Int (forall a. a) => T Int String

ddd :: Int
ddd = 0 :: forall t. T Int t => Int
```

Or

```purescript
class X a b | a -> b
class X a (forall t. t) <= Y a b | a -> b
instance tX :: X Int String
instance tY :: Y Int Boolean

ggg :: Int
ggg = 0 :: forall t. Y Int t => Int
```

Causes the following crash:

```
purs: An internal error occurred during compilation: "unifyTypes: unspecified skolem scope"
Please report this at https://github.com/purescript/purescript/issues
CallStack (from HasCallStack):
  error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-0.11.4-10BKLVNBlGN4rJNVF4PBMN:Language.PureScript.Crash
  internalError, called at src/Language/PureScript/TypeChecker/Unify.hs:105:28 in purescript-0.11.4-10BKLVNBlGN4rJNVF4PBMN:Language.PureScript.TypeChecker.Unify
purs: thread blocked indefinitely in an MVar operation
```

paf31:
Hmm, we should just disallow these syntactically.

Cmdv:
I'd been keen to take this on but might need a little hand holding!!
Could I be pointed to which file I should look into? 😄

Cmdv:
someone else can try pick this up if they like, I just got too lost 😞

PR #2901, opened 2017-05-20
Update `build` command to run test

sectore:
of example test files.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ To build and run a specific test in `examples/passing/` or `examples/failing/`, 
 
 ``` bash
 # Build
-stack exec purs compile 'tests/support/bower_components/purescript-*/src/**/*.purs' examples/blah/Blah.purs
+stack exec purs -- compile 'tests/support/bower_components/purescript-*/src/**/*.purs' examples/blah/Blah.purs
 
 # Run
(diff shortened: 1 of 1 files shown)


hdgarrood:
Have you tested this? As far as I am aware it will still require the `--` so that `stack` knows those extra arguments are for `purs` and not for `stack`.

sectore:
You are right, but it seems that in this case `stack` runs `purs compile` without `--`, too. I've run this command for manually tests of #2874.

For example with `--`

```
stack exec purs -- compile 'tests/support/bower_components/purescript-*/src/**/*.purs' examples/failing/2874-forall.purs
Error found:
at examples/failing/2874-forall.purs line 5, column 36 - line 5, column 36

  Unable to parse module:
  unexpected forall
```

Without `--`
```
stack exec purs compile 'tests/support/bower_components/purescript-*/src/**/*.purs' examples/failing/2874-forall.purs
Error found:
at examples/failing/2874-forall.purs line 5, column 36 - line 5, column 36

  Unable to parse module:
  unexpected forall
```

Anyway, I'll update this PR to use `--` as usual. Thanks for your feedback @hdgarrood !

hdgarrood:
That's good to know too - maybe `stack` has changed to be a bit more smart about command arguments since I last checked.

paf31:
Thanks!

PR #2905, opened 2017-05-23
[purs ide] Work around laziness when measuring command perf

kRITZCREEK:
Before this completion and type commands would always be measured as 0.01ms. Laziness is a tricky beast when you want to measure things ^^

### DIFF
--- a/app/Command/Ide.hs
+++ b/app/Command/Ide.hs
@@ -170,9 +170,9 @@ startServer port env = withSocketsDo $ do
                       <> displayTimeSpec duration
               logPerf message $ do
                 result <- runExceptT (handleCommand cmd')
-                case result of
-                  Right r  -> liftIO $ catchGoneHandle (BSL8.hPutStrLn h (Aeson.encode r))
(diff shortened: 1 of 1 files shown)


app/Command/Ide.hs
```diff
@@ -164,15 +164,16 @@ startServer port env = withSocketsDo $ do
...
-              -- $(logDebug) ("Answer was: " <> T.pack (show result))
+                    "Command "
+                      <> commandName cmd'
+                      <> " took "
+                      <> displayTimeSpec duration
+              logPerf message $ do
+                result <- runExceptT (handleCommand cmd')
+                case result of
+                  Right r  -> liftIO $ catchGoneHandle (BSL8.hPutStrLn h (Aeson.encode r))
+                  Left err -> liftIO $ catchGoneHandle (BSL8.hPutStrLn h (Aeson.encode err))
```
garyb: Could reduce duplication here while you're at it?

``` haskell
liftIO $ catchGoneHandle $ BSL8.hPutStrLn h $ case result of ...
```

kRITZCREEK: Needs to be explicitly annotated with `forall` so I didn't do it so far, since it ends up being more code.

PR #2907, opened 2017-05-24
[purs ide] Groups reexports in completions

### LINKED ISSUE #2752: Group completions for re-exports of the same value
With the new lenient import parser we could also offer a filepath parameter on the completion command. We could then filter by the imported modules and prefer a reexported version from a module that has already been imported.

### LINKED ISSUE #2908: [purs ide] remove duplicates from completion
The completion results often contain the same function re-exported from different modules.  For example if I complete `ge` I might get:
```
get :: forall m s. MonadState s m => m s (Control.Monad.RWS)
get :: forall m s. MonadState s m => m s (Control.Monad.State)
get :: forall m s. MonadState s m => m s (Control.Monad.State.Class)
get :: forall m s. MonadState s m => m s (Control.Monad.State.Trans)
```

And this is just one single `get` defined in `Control.Monad.State.Class`.  It would be nice to have a filter that would pick
* the module where it is defined or
* the one with the shortest path - some libraries export all their functions from top level module (`Data.Lenses` is a good example), though in this case this strategy would not find unique module, but that's ok.

kRITZCREEK:
Adds an option to control how reexport should be grouped in the completions.

TODO:

- [x] Actually describe what this does behaviour wise
- [ ] Figure out if this is the desired (default) behaviour
- [x] tests
- [x] Return the list of reexporting modules in the completion format

Fixes #2752, Fixes #2908

### DIFF
--- a/psc-ide/PROTOCOL.md
+++ b/psc-ide/PROTOCOL.md
@@ -108,7 +108,8 @@ couldn't be extracted from a source file.
     "start": [1, 3],
     "end": [3, 1]
     },
-  "documentation": "A filtering function"
+  "documentation": "A filtering function",
+  "exportedFrom": ["Data.Array"]
   }
 ]
 ```
@@ -615,11 +616,12 @@ Completion options allow to configure the number of returned completion results.
 If specified limits the number of completion results, otherwise return all
 results.
 
-- groupReexports :: Maybe (Array String)
+- groupReexports :: Maybe Boolean (defaults to False)
 
-If specified groups reexports under their original module and prefers the
-specified set of modules.
-
+If set to True, groups all reexports of an identifier under the module it
+originated from (the original export is also treated as a "reexport"). These
+reexports then populate the `exportedFrom` field in their completion results and
+the `module` field contains the originating module.
 
 ### Error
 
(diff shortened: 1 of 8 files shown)


coot:
Adding where the identifier comes from to the output would allow to filter results (or group them) at the editor level.

kRITZCREEK:
This is ready for review/discussion about API and whether this should become the default behaviour.

I kept the documentation somewhat brief, but I'll happily explain in more detail if something is unclear.

/cc @nwolverson @coot @FrigoEU

psc-ide/PROTOCOL.md
```diff
@@ -82,6 +82,7 @@ The `complete` command looks up possible completions/corrections.
     "currentModule": "Main",
     "options": {
       "maxResults": 50
+      "groupReexports": ["Prelude", "Data.Array", "Data.List"]
```
nwolverson: Is this outdated, having been switched to a boolean?

kRITZCREEK: Good catch :+1:

kRITZCREEK:
I'll merge this in for the upcoming release, since it's non-breaking. If I decide I don't like the API I'll change it again in the next breaking release. Please give feedback on whether this is useful if you happen to use it in your editor plugins.

PR #2910, opened 2017-05-25
Filter out 'unsafeCoerce' from type hole suggestions

justinwoo:
A bit naive implementation of what I wanted, hopefully this lets us have a discussion that's written down if we want this or not.

Phil was saying that we should have something similar to how Pursuit ranks entries by relevance when displaying them, but I guess that would involve code being pulled in from it to here?

I at least don't think all unsafe____ methods should be filtered out though, since many times you actually do want to use `unsafeStringify` or something. It's just that unsafeCoerce is just basically never the thing you want if you're using a type hole to get suggestions.

example screenshot:
jiefjfbne

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -21,6 +21,7 @@ import           Data.Functor.Identity (Identity(..))
 import           Data.List (transpose, nubBy, sort, partition, dropWhileEnd)
 import           Data.Maybe (maybeToList, fromMaybe, mapMaybe)
 import qualified Data.Map as M
+import qualified Data.Set as S
 import qualified Data.Text as T
 import           Data.Text (Text)
 import           Language.PureScript.AST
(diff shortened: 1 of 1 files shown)


hdgarrood:
How about instead filtering out anything whose type is `forall a b. a -> b`?

justinwoo:
Yeah, I guess that'd be more correct, but I don't see any results other than UnsafeCoerce with such a signature on Pursuit though...

hdgarrood:
True, but someone might have defined their own version of it (not knowing it exists), or someone might have defined a different function with that name, or we might want to change its name.

In general I'd be a little reluctant to make assumptions about purescript code in general based on what Pursuit says - in my experience application code looks quite different to library code, and of course only the latter is represented on Pursuit.

kRITZCREEK:
@justinwoo Thank you very much for looking into this ❤️ I thought a bit about this, and maybe one solution would be to check that the function does not return an unconstrained type variable.

This way we still get to keep things like `throwError` since its return `m a` is constrained with `MonadError` on the `m`.

justinwoo:
Updated: I did the easiest thing I could think of compared the set of constrained variables to the set of the right-most operand of Function/a unit...

paf31:
So `unsafeCoerce` is never a _wrong_ suggestion, but it is a _bad_ one, because there is almost always a better one. So I think we need to define "better" more precisely, and then sort the results, instead of looking for special cases. I would suggest we start with the logic in Pursuit's type search.

justinwoo:
In the meanwhile, would it be useful enough to just change this to render a new box of these "less desirable" suggestions below the "regular" suggestions?

paf31:
> would it be useful enough to just change this to render a new box of these "less desirable" suggestions below the "regular" suggestions

If we can define "less desirable" in a better way than "is `unsafeCoerce`" 😄

justinwoo:
Updated with a better commit message this this. The approach for "less desirable" described in the commit: I'm taking any function with the rightmost side containing no constrained variables and putting it in the penalty box 😛

src/Language/PureScript/Errors.hs
```diff
@@ -800,14 +801,40 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
       in
         paras $ [ line $ "Hole '" <> markCode name <> "' has the inferred type "
                 , markCodeBox (indent (typeAsBox ty))
                 ] ++ tsResult ++ renderContext ctx
+          where
+            resultVariablesInConstraints (_, t) =
+              not . S.null $ S.intersection constrained results
+              where
+                constrained = S.fromList (getConstrainedVariables t)
+                results = S.fromList (usedTypeVariables . last $ splitFunctionType t)
```
paf31: Please fuse the call to `last` into `splitFunctionType` here.

src/Language/PureScript/Errors.hs
```diff
@@ -800,14 +801,40 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
+            getConstrainedVariables :: Type -> [Text]
+            getConstrainedVariables t = go t
+              where
+                go (ForAll _ t' _) = go t'
+                go (ConstrainedType (Constraint{constraintArgs=args}) t') =
+                  (concat $ usedTypeVariables <$> args) <> go t'
+                go (TypeApp lhs rhs) = go lhs <> go rhs
```
paf31: Is this case necessary?

src/Language/PureScript/Errors.hs
```diff
@@ -800,14 +801,40 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
+               , makeCodeBox ys
                ]
           _ -> []
       in
         paras $ [ line $ "Hole '" <> markCode name <> "' has the inferred type "
                 , markCodeBox (indent (typeAsBox ty))
                 ] ++ tsResult ++ renderContext ctx
+          where
+            resultVariablesInConstraints (_, t) =
+              not . S.null $ S.intersection constrained results
```
paf31: This test seems wrong to me. If there are no constraints, then the intersection is empty, so every function with no constraints goes in the useless bucket. I think the test should be something like _every type variable in the output is either constrained or mentioned in some input_.

paf31: Oops, I just realized I was looking at outdated code 😄

src/Language/PureScript/Errors.hs
```diff
@@ -800,14 +801,51 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
                ]
           _ -> []
       in
         paras $ [ line $ "Hole '" <> markCode name <> "' has the inferred type "
                 , markCodeBox (indent (typeAsBox ty))
                 ] ++ tsResult ++ renderContext ctx
+          where
+            conditions x =
+              resultVariablesInConstraints x || rightSideAppearsElsewhere x
+            rightSideAppearsElsewhere (_, t) =
```
paf31: We should avoid two traversals here.

src/Language/PureScript/Errors.hs
paf31: Ideally, we should collect all three sets in one pass: constrained variables, variables appearing on the left, and variables appearing on the right.

justinwoo:
Updated to have the splitting and cvar extraction out at the top and then pass on the results for the others.

paf31:
I think this is another case where we need an issue to discuss options.

paf31:
I'm going to close this one for now, but I'm happy for us to discuss approaches in an issue and maybe reopen this later.

I still have a concern that this is not the right filter. `unsafeFromForeign` would be filtered out, for example, which doesn't seem to be in the same category of bad as `unsafeCoerce`.

I would even consider reading a `.do-not-suggest` dotfile 😄

PR #2912, opened 2017-05-27
Allow type signatures in instances

### LINKED ISSUE #2902: Add InstanceSigs
It would be nice to be able to put type signatures in type class instances.  This is similar to Haskell's [`InstanceSigs`](https://downloads.haskell.org/~ghc/8.0.2/docs/html/users_guide/glasgow_exts.html#instance-signatures-type-signatures-in-instance-declarations) extension.

For instance, the [`Data.Maybe`](https://github.com/purescript/purescript-maybe/blob/master/src/Data/Maybe.purs) module has the following code:

```purescript
data Maybe a = Nothing | Just a

instance functorMaybe :: Functor Maybe where
  map fn (Just x) = Just (fn x)
  map _  _        = Nothing
```

It would be nice to be able to write the following:

```purescript
instance functorMaybe :: Functor Maybe where
  map :: forall a b. (a -> b) -> Maybe a -> Maybe b
  map fn (Just x) = Just (fn x)
  map _  _        = Nothing
```

Notice how `map` has been given an explicit signature.

## Pros and Cons

**Cons**:

-   Having a type signature in a type class instance is redundant, because the signature is already given in the type class definition itself.  Reducing duplication from code is generally a good thing.

**Pros**:

-   The additional types act as documentation.  This is especially helpful in code like the `Maybe` example above, where the instance is defined in a separate module from the type class.  A user reading through [`Data.Maybe`](https://github.com/purescript/purescript-maybe/blob/master/src/Data/Maybe.purs) would have an easier time understanding the implementation of the instances if they had signatures.

    When I'm writing code, sometimes I want the code to be as easy to understand as possible, even if that makes it slightly longer or more redundant.
-   Instance signatures allow you to bring scoped type variables into scope.  For instance, the following code becomes possible:

    ```purescript
    data Foo a = Foo a a

    class Bar b where
      bar :: forall x. x -> b -> Tuple b (Array x)

    instance barFoo :: Bar (Foo a) where
      bar :: forall x. x -> Foo a -> Tuple (Foo a) (Array x)
      bar x foo = Tuple foo xs
        where
          xs :: Array x
          xs = [x, x, x]
    ```

    Notice how the type variable `x` is being used in the definition of `xs` in the `where` clause.

## Additional References

The following pages have some more discussion about `InstanceSigs` in Haskell:

-   [Stack Overflow post](http://stackoverflow.com/questions/8367426/why-cant-one-put-type-signatures-in-instance-declarations-in-haskell) that inspired the `InstanceSigs` extension in Haskell.
-   [GHC Ticket](https://ghc.haskell.org/trac/ghc/ticket/5676) for `InstanceSigs`.
-   [GHC User's Guide](https://downloads.haskell.org/~ghc/8.0.2/docs/html/users_guide/glasgow_exts.html#instance-signatures-type-signatures-in-instance-declarations) on `InstanceSigs`.

## Unresolved Questions

Should instance signatures be allowed to be more general than the definition in the type class?

For example, should the following code be allowed?

```purescript
data Foo a = Foo a a

class Eq x where
  eq :: x -> x -> Boolean

instance eqFoo :: Eq (Foo a) where
  eq :: forall b. b -> b -> Boolean
  eq x y = true
```

## Implementation Plan

I'll implement it in the compiler.  I think adding the ability to have instance signatures will be a (relatively?) simple change, and a good project for a first-time contributor.

The main areas in the compiler that will be affected are the parser and the type-checker.

This implementation should be split into two PRs:

-   The first PR will implement basic instance signatures functionality, as described above.
-   The second PR will add the ability for instance signatures to be more general than the definition in the type class, if it is decided that this is a good idea.

## Note

Keep in mind that whether or not the compiler should support instance signatures is a slightly different argument from whether or not most code should use instance signatures.  For example, I don't necessarily think that modules like [`Data.Maybe`](https://github.com/purescript/purescript-maybe/blob/master/src/Data/Maybe.purs) should use instance signatures. But, I do want to use them in my own code.

cdepillabout:
This is a WIP PR for implenting InstanceSigs in PureScript.  

Closes #2902.

Commit 6768e35 changes the parser to allow type signatures inside of type class instances.

Commit c56c615 adds some example files.

This is mostly working except for one thing.   Instance members are allowed to have signatures that are more general than the signature declared in the type class.

The PureScript file is able to be compiled successfully:

```purescript
module Main where

class Eq a where
  eq :: a -> a -> Boolean

instance eqNumber :: Eq Number where
  eq :: forall x y. x -> y -> Boolean
  eq _ _ = true
```

In #2902, @paf31 made the following [comment](https://github.com/purescript/purescript/issues/2902#issuecomment-303142729):

>> 
>> Should instance signatures be allowed to be more general than the definition in the type class?
> 
> No, I don't think so. Instance members already have an implicit type signature, so I think this should just optionally replace that.

I was trying to figure out where to make a change to disallow more general instance signatures.  However, I couldn't figure out where it would make the most sense.

Here are the three potential places I found, ordered by code flow (so, for instance, (1) happens first in the compilation process, etc):

1.  [Language.PureScript.Sugar.TypeDeclarations.desugarTypeDeclarationsModule](https://github.com/purescript/purescript/blob/0a6de6ec9431df0b01d00e9dd10aaade0aee043a/src/Language/PureScript/Sugar/TypeDeclarations.hs#L57): This function changes `TypeDeclaration`s to `ValueDeclaration`s.  

    This feels like a good place to do the check, but  we don't have access to the environment with all the available type classes.
2.  [Language.PureScript.Sugar.TypeClasses.typeInstanceDictionaryDeclaration](https://github.com/purescript/purescript/blob/0a6de6ec9431df0b01d00e9dd10aaade0aee043a/src/Language/PureScript/Sugar/TypeClasses.hs#L274): This function changes a `TypeInstanceDeclaration` into a `ValueDeclaration` wrapped around a `TypeClassDictionaryConstructorApp`.

    This also feels like it might be a good place to do the check, but we've sort-of lost the type information in (1).  I imagine it would be possible to try to inspect the `ValueDeclaration`s created in (1) and figure out which ones are wrapping around a `GuardedExpr [] (TypedValue True val ty)`.  But this seems like it would be slightly hacky?
3.  Somewhere in the type checker.  Not knowing anything about the compiler, originally I thought this sort of check would happen in the type checker.  However, I wasn't sure if there would be a good place for it here.  Just like with (2), we've sort-of lost the type definition in (1), so the solution would be somewhat hacky.

    Maybe a function like [Language.PureScript.TypeChecker.Types.checkTypedBindingGroupElement](https://github.com/purescript/purescript/blob/0a6de6ec9431df0b01d00e9dd10aaade0aee043a/src/Language/PureScript/TypeChecker/Types.hs#L225) would be a good place to start?

I can think of two possible ways to continue from here.

1.  Merge this PR (after changing the one test from "failing" to "passing").  It's not completely horrible to accept instance members that are too general.  It's how Haskell handles it.
2.  Maybe someone could take a look at this PR and give me some ideas on where this sort of check should go.  A high-level overview of how it might be implemented would also be helpful :-)

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -216,8 +216,14 @@ parseTypeInstanceDeclaration = do
   instanceDecl <- parseInstanceDeclaration
   members <- P.option [] $ do
     indented *> reserved "where"
-    mark (P.many (same *> positioned parseValueDeclaration))
+    mark (P.many (same *> positioned declsInInstance))
   return $ instanceDecl (ExplicitInstance members)
+  where
+    declsInInstance :: TokenParser Declaration
+    declsInInstance = P.choice
+      [ parseTypeDeclaration
+      , parseValueDeclaration
+      ] P.<?> "type declaration or value declaration in instance"
 
 parseDerivingInstanceDeclaration :: TokenParser Declaration
 parseDerivingInstanceDeclaration = do
--- a/examples/failing/InstanceSigsBodyIncorrect.purs
+++ b/examples/failing/InstanceSigsBodyIncorrect.purs
@@ -1,0 +1,10 @@
+-- @shouldFailWith TypesDoNotUnify
+
+module Main where
+
+class Foo a where
+  foo :: a
+
+instance fooNumber :: Foo Number where
+  foo :: Number
+  foo = true
--- a/examples/failing/InstanceSigsDifferentTypes.purs
+++ b/examples/failing/InstanceSigsDifferentTypes.purs
@@ -1,0 +1,10 @@
+-- @shouldFailWith TypesDoNotUnify
+
+module Main where
+
+class Foo a where
+  foo :: a
+
+instance fooNumber :: Foo Number where
+  foo :: Int
+  foo = 0.0
--- a/examples/failing/InstanceSigsIncorrectType.purs
+++ b/examples/failing/InstanceSigsIncorrectType.purs
@@ -1,0 +1,10 @@
+-- @shouldFailWith TypesDoNotUnify
+
+module Main where
+
+class Foo a where
+  foo :: a
+
+instance fooNumber :: Foo Number where
+  foo :: Boolean
+  foo = true
--- a/examples/failing/InstanceSigsOrphanTypeDeclaration.purs
+++ b/examples/failing/InstanceSigsOrphanTypeDeclaration.purs
@@ -1,0 +1,10 @@
+-- @shouldFailWith OrphanTypeDeclaration
+
+module Main where
+
+class Foo a where
+  foo :: a
+
+instance fooNumber :: Foo Number where
+  bar :: Int
+  foo = 0.0
--- a/examples/failing/InstanceSigsTypeTooGeneral.purs
+++ b/examples/failing/InstanceSigsTypeTooGeneral.purs
@@ -1,0 +1,10 @@
+-- @shouldFailWith TypesDoNotUnify
+
+module Main where
+
+class Eq a where
+  eq :: a -> a -> Boolean
+
+instance eqNumber :: Eq Number where
+  eq :: forall x y. x -> y -> Boolean
+  eq _ _ = true
--- a/examples/passing/InstanceSigs.purs
+++ b/examples/passing/InstanceSigs.purs
@@ -1,0 +1,12 @@
+module Main where
+
+import Control.Monad.Eff.Console (log)
+
+class Foo a where
+  foo :: a
+
+instance fooNumber :: Foo Number where
+  foo :: Number
+  foo = 0.0
+
+main = log "Done"
--- a/examples/passing/InstanceSigsGeneral.purs
+++ b/examples/passing/InstanceSigsGeneral.purs
@@ -1,6 +1,6 @@
--- @shouldFailWith TypesDoNotUnify
-
 module Main where
+
+import Control.Monad.Eff.Console (log)
 
 class Eq a where
   eq :: a -> a -> Boolean
@@ -8,3 +8,5 @@ class Eq a where
 instance eqNumber :: Eq Number where
   eq :: forall x y. x -> y -> Boolean
   eq _ _ = true
+
+main = log "Done"
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -185,9 +185,10 @@ extra-source-files:
       examples/failing/InfiniteType.purs
       examples/failing/InstanceExport.purs
       examples/failing/InstanceExport/InstanceExport.purs
+      examples/failing/InstanceSigsBodyIncorrect.purs
       examples/failing/InstanceSigsDifferentTypes.purs
+      examples/failing/InstanceSigsIncorrectType.purs
       examples/failing/InstanceSigsOrphanTypeDeclaration.purs
-      examples/failing/InstanceSigsTypeTooGeneral.purs
       examples/failing/IntOutOfRange.purs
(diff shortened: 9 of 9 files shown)


paf31:
After more thought, I'm fine with allowing more general type signatures.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -216,8 +216,14 @@ parseTypeInstanceDeclaration = do
   instanceDecl <- parseInstanceDeclaration
   members <- P.option [] $ do
     indented *> reserved "where"
-    mark (P.many (same *> positioned parseValueDeclaration))
+    mark (P.many (same *> positioned declsInInstance))
   return $ instanceDecl (ExplicitInstance members)
+  where
+    declsInInstance :: TokenParser Declaration
+    declsInInstance = P.choice
+      [ parseTypeDeclaration
```
paf31: Wow, that's all? I'm slightly surprised this works, although it makes sense now that I think about it 😄

paf31:
Could you please add a test where there is a type signature with the wrong type?

src/Language/PureScript/Parser/Declarations.hs
cdepillabout: I was surprised as well!  Much simpler than I thought it would be :-)

cdepillabout:
@paf31 I've pushed a new commit ae41a97 that changes the "failing" example for signatures that are too general to be "passing", since we want to allow it.

I've also added "failing" examples dealing with incorrect type signatures.  Here they are:

`examples/failing/InstanceSigsIncorrectType.purs`: `foo` has a type signature that is incorrect, but it _is correct_ for the given body of the function.

```purescript
-- @shouldFailWith TypesDoNotUnify

module Main where

class Foo a where
  foo :: a

instance fooNumber :: Foo Number where
  foo :: Boolean
  foo = true
```

`examples/failing/InstanceSigsBodyIncorrect.purs`: The type signature on `foo` is correct, but the body of `foo` is not of the correct type.

```purescript
-- @shouldFailWith TypesDoNotUnify

module Main where

class Foo a where
  foo :: a

instance fooNumber :: Foo Number where
  foo :: Number
  foo = true
```

`examples/failing/InstanceSigsDifferentTypes.purs`: The type signature for `foo` has the incorrect type, but the body of `foo` has the type expected.

```purescript
-- @shouldFailWith TypesDoNotUnify

module Main where

class Foo a where
  foo :: a

instance fooNumber :: Foo Number where
  foo :: Int
  foo = 0.0
```

cdepillabout:
One last though:  The [Extensions](https://github.com/purescript/purescript/wiki/Differences-from-Haskell#extensions) section on the "Differences from Haskell" wiki page should be updated to include `InstanceSigs`.  

I imagine the timing for this should be when the next version of the compiler is released.

hdgarrood:
Maybe it's time to fork an 0.12 branch in the documentation repo for changes like this.

garyb:
Do we need to wait for 0.12 for this? It doesn't seem obviously breaking to me, but I'm probably missing something.

hdgarrood:
Oh yes, of course. Ignore me then!

paf31:
Yes, I'll merge this before 0.11.5.

I wasn't planning to have a 0.12 release actually. Is there anything in the 1.0 milestone which we think should be separated out?

hdgarrood:
I'm not aware of anything like that, it just didn't occur to me that the next major release would be 1.0.

paf31:
Thanks!

PR #2915, opened 2017-05-29
fix psci tests

kRITZCREEK:
What's the plan for making these less flaky again? Maybe worth writing up an issue.

PR #2916, opened 2017-05-29
Parse support modules from actual test support purs

noraesae:
This may be an alternative solution for #2915.

`unsafePerformIO ` is used to make it quickly work without other source changes and memoize the IO result at the same time. I guess it may be okay given it's test code, but please let me know if there is a better way.

### DIFF
--- a/tests/TestCompiler.hs
+++ b/tests/TestCompiler.hs
@@ -61,26 +61,23 @@ main = hspec spec
 spec :: Spec
 spec = do
 
-  (supportExterns, supportForeigns, passingTestCases, warningTestCases, failingTestCases) <- runIO $ do
+  (supportModules, supportExterns, supportForeigns, passingTestCases, warningTestCases, failingTestCases) <- runIO $ do
     cwd <- getCurrentDirectory
     let passing = cwd </> "examples" </> "passing"
     let warning = cwd </> "examples" </> "warning"
     let failing = cwd </> "examples" </> "failing"
-    let supportDir = cwd </> "tests" </> "support" </> "bower_components"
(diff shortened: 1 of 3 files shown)


noraesae:
[Text] seems like a better fit. I will check.

kRITZCREEK:
You won't need to unsafePerformIO if you use hspec's `runIO` in `completionTests`:

```haskell
completionTests :: Spec
completionTests = do
  supportModules <- runIO getSupportModules
  context "completionTests" $ do
    mapM_ assertCompletedOk (completionTestData supportModules)
```

noraesae:
Yes, I've once tried it, but without memoization compiler tests take too much time. I've tried `io-memoize's [`once`](https://www.stackage.org/haddock/lts-8.15/io-memoize-1.1.1.0/System-IO-Memoize.html#v:once) too, but without success. What is the best practice to memoize IO result?

kRITZCREEK:
`completionTests` is only run once, right? It shouldn't be reevaluated for every test in completionTestData

noraesae:
Yes, but we use it elsewhere too.

https://github.com/purescript/purescript/blob/master/tests/TestCompiler.hs#L182

kRITZCREEK:
I see. Well I'm sure we could hoist the call IO call and pass the results down, but since it's "just" test code I'm happy either way.

noraesae:
Yes, actually, I think it's not so difficult to make it an IO action. I will look into it tomorrow.

noraesae:
I've removed `unsafePerformIO` by refactoring `TestCompiler.hs` and implementing IO actions for support modules.

paf31:
Thanks!

PR #2918, opened 2017-05-30
Encode PSCI's server content as utf-8 string

### LINKED ISSUE #2914: ℏ (planck constant) encoding problem in browser psci
When I run  `purs repl 'src/**/*.purs' 'bower_components/purescript-*/src/**/*.purs' -p 8000` for some reason [planck's constant](http://www.fileformat.info/info/unicode/char/210f/index.htm) doesn't get encoded correctly in the bundle.js file.

```purescript
import Prelude

ℏ :: String
ℏ = "planck"

main :: forall e. Eff (console :: CONSOLE | e) Unit
main = do
  log ℏ
```

image

However, if I build the bundle with `pulp build --to test.js`, the generated file works fine.  Other lowercase unicode characters work fine.

dgendill:
I think this should fix #2914.  I'm not sure exactly how we could automate testing for this though.  Let me know if you think a test is necessary and if so how we should go about it.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -36,6 +36,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@davidchambers](https://github.com/davidchambers) | David Chambers | [MIT license](http://opensource.org/licenses/MIT) |
 | [@DavidLindbom](https://github.com/DavidLindbom) | David Lindbom | [MIT license](http://opensource.org/licenses/MIT) |
 | [@dckc](https://github.com/dckc) | Dan Connolly | [MIT license](http://opensource.org/licenses/MIT) |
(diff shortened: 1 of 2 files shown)


hdgarrood:
We should be sending the text encoding in the content type header in every request imo, not just the HTML ones. The syntax is the same as far as I am aware: "Content-Type: application/javascript; charset: utf-8". Also is there any way we could avoid the `String` type entirely?

hdgarrood:
I imagine this will be more effort to test properly than it is worth, though, so I don't mind not having tests.

dgendill:
From what I understand, javascript content types like application/javascript default to UTF-8, so that's why I left them out on those requests.

We might be able to avoid String and just use ByteString, but it seems like there has to be some function that says "interpret this as UTF-8."  I couldn't fine a `ByteString -> ByteString` function that did that.

PR #2928, opened 2017-06-06
Add empty case expressions

### LINKED ISSUE #2922: Allow case expressions without brnaches
Can be used for empty data decls such as `Void`. Proposed syntax is to leave off the `of` keyword.

    data Void

    absurd :: forall a. Void -> a
    absurd = case _

Type inference may suffer. What would the inferred type of `absurd` be if no explicit type was given?

Issue about exhaustiveness checker exists elsewhere.

rightfold:
Fixes #2922.

WIP:

 - [ ] Exhaustivity checker now accepts all empty case expressions. Check that at least one of the arguments is of an empty type.

### DIFF
--- a/src/Language/PureScript/Linter/Exhaustive.hs
+++ b/src/Language/PureScript/Linter/Exhaustive.hs
@@ -235,21 +235,33 @@ missingAlternative env mn ca uncovered
 
 -- |
 -- Main exhaustivity checking function
--- Starting with the set `uncovered = { _ }` (nothing covered, one `_` for each function argument),
--- it partitions that set with the new uncovered cases, until it consumes the whole set of clauses.
--- Then, returns the uncovered set of case alternatives.
+-- If there are no cases, check that at least one argument is of an empty type.
+-- Otherwise, starting with the set `uncovered = { _ }` (nothing covered, one
+-- `_` for each function argument), it partitions that set with the new
+-- uncovered cases, until it consumes the whole set of clauses. Then, returns
+-- the uncovered set of case alternatives.
 --
 checkExhaustive
   :: forall m
    . (MonadWriter MultipleErrors m, MonadSupply m)
    => Environment
    -> ModuleName
-   -> Int
+   -> [Expr]
    -> [CaseAlternative]
    -> Expr
    -> m Expr
-checkExhaustive env mn numArgs cas expr = makeResult . first ordNub $ foldl' step ([initialize numArgs], (pure True, [])) cas
+checkExhaustive env mn args cas expr =
+  makeResult . first ordNub $
+    if null cas
+      then checkArgsOfEmptyType
+      else foldl' step ([initialize numArgs], (pure True, [])) cas
   where
+  numArgs :: Int
+  numArgs = length args
+
+  checkArgsOfEmptyType :: ([[Binder]], (Either RedundancyError Bool, [[Binder]]))
+  checkArgsOfEmptyType = ([], (Right (any (const True) args), []))
+
   step :: ([[Binder]], (Either RedundancyError Bool, [[Binder]])) -> CaseAlternative -> ([[Binder]], (Either RedundancyError Bool, [[Binder]]))
   step (uncovered, (nec, redundant)) ca =
     let (missed, pr) = unzip (map (missingAlternative env mn ca) uncovered)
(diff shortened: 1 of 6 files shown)


src/Language/PureScript/Linter/Exhaustive.hs
```diff
@@ -235,21 +235,33 @@ missingAlternative env mn ca uncovered
...
    -> ModuleName
-   -> Int
+   -> [Expr]
    -> [CaseAlternative]
    -> Expr
    -> m Expr
-checkExhaustive env mn numArgs cas expr = makeResult . first ordNub $ foldl' step ([initialize numArgs], (pure True, [])) cas
+checkExhaustive env mn args cas expr =
+  makeResult . first ordNub $
+    if null cas
```
paf31: I feel like an empty case should not be treated differently. For example, what about this?

```
x :: Void -> a
x a = x a
x b = x b
```

That's redundant and exhaustive, but we will still say it's not exhaustive.

I think the trick has to be to consider the final set of binders after the `foldl`, and to realize that a wildcard binder does not necessarily represent unhandled values, since a data type might be empty.

paf31: Also, what about

```
y :: forall a. Void -> Void
y a = a
```

That should be redundant now. Ideally, I think we'd realize that and report it, recommending `case _` instead.

paf31:
@nicodelpiano might be interested in this too.

src/Language/PureScript/Linter/Exhaustive.hs
rightfold: I am not familiar with the exhaustivity checker, I'll have to spend some more time on it.

I don't think we want to signal that as redundant, because then you have no way of propagating the `Void` proof to something else anymore.

paf31: But it _is_ redundant 😄 why do you need to propagate anything when `case _` is a full implementation?

rightfold: Yes, I'm dumb. Edited.

nicodelpiano:
Sure, I'll have a look at this!

rightfold:
@nicodelpiano Did you take a look at this? I don't understand how the exhaustiveness checker works.

nicodelpiano:
Hey @rightfold, sorry for my delayed response. Yes, I had a look while back ago, and what we could do is to lookup the environment and verify whether or not they're empty types.

For example, the environment types entry for
```purescript
data Nat = Zero | Succ Nat
```
is
```
Qualified (Just (ModuleName [ProperName {runProperName = \"Main\"}])) (ProperName {runProperName = \"Nat\"}),(NamedKind (Qualified (Just (ModuleName [ProperName {runProperName = \"Prim\"}])) (ProperName {runProperName = \"Type\"})),DataType [] [(ProperName {runProperName = \"Zero\"},[]),(ProperName {runProperName = \"Succ\"},[TypeConstructor (Qualified (Just (ModuleName [ProperName {runProperName = \"Main\"}])) (ProperName {runProperName = \"Nat\"}))])]))
```
In contrast, for
```purescript
data Void
```
is
```
Qualified (Just (ModuleName [ProperName {runProperName = \"Main\"}])) (ProperName {runProperName = \"Void\"}),(NamedKind (Qualified (Just (ModuleName [ProperName {runProperName = \"Prim\"}])) (ProperName {runProperName = \"Type\"})), DataType [] []))
```

Basically, the exhaustivity checker analyses every `Case` expression (see `checkExhaustiveExpr` here [1]) and calculates the missing cases from the clauses of the pattern matching definition from top to bottom. Getting a constructor of some type (say, `Zero` in the `Nat` type defined above), it retrieves `Succ` looking up the environment (see `getConstructors` here [2]).

I think we can say that an empty type is actually empty checking if the argument highlighted in bold below is an empty list
```
Qualified (Just (ModuleName [ProperName {runP    roperName = \"Main\"}])) (ProperName {runProperName = \"Void\"}),(NamedKind (Qualified (Just (ModuleName [ProperName {runProperName = \"Prim\"}])) (ProperName {run    ProperName = \"Type\"})),DataType [] **[]**))
```

An interesting example is this one
```purescript
y :: forall a. Void -> Void
y a = a
```
which isn't transformed to a `Case` expression and thus isn't checked for exhaustiveness (see `checkExhaustiveExpr` in [1]).

I reckon we could fix this, by checking the environment for empty types like I said before, and then capturing some expressions which aren't of the form `Case`. Maybe @paf31 could give us some hints and correct me if I'm wrong.

[1]: https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Linter/Exhaustive.hs#L347
[2]: https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Linter/Exhaustive.hs#L67

paf31:
Closing for now, since we need to figure out the exhaustivity issue, and that needs a spec. As usual, happy to reopen after discussion. Thanks!

Issue #2315, opened 2016-09-18
Instance chains


garyb:
This is the only thing left from the type system wishlist that doesn't have it's own issue.
- The paper: http://homepages.inf.ed.ac.uk/jmorri14/pubs/morris-icfp2010-instances.pdf
- Some interesting discussions on the GHC Trac: https://ghc.haskell.org/trac/ghc/ticket/9334

garyb:
I was thinking named instances might help us out when it comes to instance chains. In the original paper I think they have to be defined in a closed setting as the chain is syntactically specified together, but we could perhaps do something with the names so that they are extensible, so the chain (or maybe, graph I guess) can be extended by other libraries. Although I guess we'd need to be careful in what we allow there to prevent inference from behaving differently in the presence of different combinations of libraries.

Having just had a quick read of the GHC discussion, there's an interesting point too that may be worth considering: perhaps a restricted form of this can be allowed when overlapping instances are declared together in the same module, by essentially making the order of definitions in that module the chain - and allow the overlaps without warning, as the behaviour can be considered well specified (assuming we make the type checker actually use the ordering). If overlapping instances occur across module boundaries then the warning reappears. This does have the problem of only allowing closed setups for things like `Inject`, but I wonder if it would be easier / a half way fix to get us some amount of the benefit of instance chains without making things too complex.

Just some thoughts I had knocking around anyway.

LiamGoodacre:
I am currently prototyping the instance groups part of this.  Not doing the backtracking/guards bit.
I'll open a PR soon for discussion.

toraritte:
(The original link above for the paper is dead, and its new home is at [J Garrett Morris' page](https://jgbm.github.io): https://jgbm.github.io/pubs/morris-icfp2010-instances.pdf . His PhD dissertation, [Type Classes and Instance Chains](https://jgbm.github.io/pubs/morris-dissertation.pdf), is also related.)

PR #2929, opened 2017-06-06
Instance chain groups

LiamGoodacre:
For https://github.com/purescript/purescript/issues/2315


paf31:
I like the simplicity of this approach a lot, but there is a potential issue, as we discussed on Slack:

The solver works in stages, since there are two types of constraints:

- Type class constraints with functional dependencies
- Equality constraints which are generated by functional dependencies

Equality constraints cause types to become more solved by unification, which lets us try to solve more type class constraints, which generates more equalities, and so on.

The problem is that if we have a chain containing more than one instance, we cannot decide to use the second instance in the chain without knowing with certainty that we can never match the first. The problem is that we don't necessarily know this until we have enough information about the types involved to rule out the first instance, and that might require other functional dependencies to be solved first.

For example, suppose we have

```purescript
class C a

instance x :: C String where
  ...
else instance y :: C a where
  ...
```

and suppose we need to solve `C t0` where `t0` is unknown. With the current approach, we would try `x` and it would fail, but `y` would succeed, since it matches everything. There is no overlap and we choose `y`. But some fundep might later be solved which forces `t0 ~ String`, and then we should have chosen `x`!

So before we can consider `y` we need to know that the types in the instance head are _apart_ from the ones in `x`, not just that we won't choose `x` this time round.

I think we can do this by modifying the `typeHeadsAreEqual` function to differentiate the two cases.

LiamGoodacre:
@paf31 There are still some TODOs and refactoring that could be done.  But is this the kind of thing you were thinking?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -430,35 +453,37 @@ matches deps TypeClassDictionaryInScope{..} tys = do
...
-        go _ _                                                        = (False, M.empty)
-    typeHeadsAreEqual _ _ = (False, M.empty)
-
-    both :: (Bool, Matching [Type]) -> (Bool, Matching [Type]) -> (Bool, Matching [Type])
-    both (b1, m1) (b2, m2) = (b1 && b2, M.unionWith (++) m1 m2)
+        go ([], REmpty)           ([], REmpty)                        = (Match (), M.empty)
+        go ([], TUnknown u1)      ([], TUnknown u2)      | u1 == u2   = (Match (), M.empty)
+        go ([], TypeVar v1)       ([], TypeVar v2)       | v1 == v2   = (Match (), M.empty)
+        go ([], Skolem _ sk1 _ _) ([], Skolem _ sk2 _ _) | sk1 == sk2 = (Match (), M.empty)
+        go (sd, r)                ([], TypeVar v)                     = (Match (), M.singleton v [rowFromList (sd, r)])
```
paf31: Do you need an `Unknown` case here too?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -430,35 +453,37 @@ matches deps TypeClassDictionaryInScope{..} tys = do
...
+        go _ _                                                        = (Apart, M.empty)
+    typeHeadsAreEqual (TypeVar _) _ = (Unknown, M.empty)
```
paf31: Does this case get hit?

src/Language/PureScript/Parser/Lexer.hs
```diff
@@ -562,6 +562,7 @@ reservedPsNames = [ "data"
                   , "infix"
                   , "class"
                   , "instance"
+                  , "instances"
```
paf31: Also `else`?

paf31:
Looks great!

Is this breaking? I guess, technically yes, since it reserves new keywords. In that case, maybe we should also disallow overlapping instances, and release this in 0.12.0.

src/Language/PureScript/Parser/Lexer.hs
LiamGoodacre: Oh I can revert this change.  I updated the syntax to use interleaved `else` - which are already in this list.

LiamGoodacre:
The only breaking part here would be that the order of overlapping instances might be different (if we don't disallow them).  But we don't advise people use overlapping instances anyway?  There are no new keywords.

paf31:
Ok, great, then we can merge this into 0.11.x and disallow overlaps in 0.12.0 as a breaking change!

Very excited to merge this 😄

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -298,7 +319,8 @@ entails SolverOptions{..} constraint context hints =
               | pairwiseAny overlapping (map snd tcds) = do
                   tell . errorMessage $ OverlappingInstances className' tyArgs (tcds >>= (toList . namedInstanceIdentifier . tcdValue . snd))
                   return $ uncurry Solved (head tcds)
-              | otherwise = return $ uncurry Solved (minimumBy (compare `on` length . tcdPath . snd) tcds)
+              | otherwise =
+                return $ uncurry Solved (minimumBy (compare `on` (\d -> (tcdChain d, tcdIndex d, length (tcdPath d))) . snd) tcds)
```
LiamGoodacre: Now that we are considering instances in order, I shouldn't need to do this any more.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -315,7 +337,7 @@ entails SolverOptions{..} constraint context hints =
             overlapping _ TypeClassDictionaryInScope{ tcdPath = _ : _ } = False
             overlapping TypeClassDictionaryInScope{ tcdDependencies = Nothing } _ = False
             overlapping _ TypeClassDictionaryInScope{ tcdDependencies = Nothing } = False
-            overlapping tcd1 tcd2 = tcdValue tcd1 /= tcdValue tcd2
+            overlapping tcd1 tcd2 = tcdValue tcd1 /= tcdValue tcd2 && (null (tcdChain tcd1) || tcdChain tcd1 /= tcdChain tcd2)
```
LiamGoodacre: Shouldn't need this either.

src/Language/PureScript/TypeChecker/Entailment.hs
LiamGoodacre: I've been trying for a little while, but I'm not sure how to get this situation to happen.  Any ideas?

paf31: Actually, I don't think either is required. But either both are, or neither is.

paf31:
@LiamGoodacre Is this ready for a final review then?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -207,12 +220,20 @@ entails SolverOptions{..} constraint context hints =
...
-                  [ (substs, tcd)
-                  | tcd <- forClassName (combineContexts context inferred) className' tys''
-                    -- Make sure the type unifies with the type in the type instance definition
-                  , substs <- maybeToList (matches typeClassDependencies tcd tys'')
-                  ]
+            let instances = do
+                  -- TODO: efficiency
+                  chain <- groupBy ((==) `on` tcdChain) $ forClassName (combineContexts context inferred) className' tys''
+                  -- process instances in a chain in index order
+                  let found = for (sortBy (compare `on` tcdIndex) chain) $ \tcd ->
```
LiamGoodacre: Hmm, I might need to special-case when `tcdChain = []`.  The processing of any non-chain instances shouldn't interfere with each other.  In this case we wouldn't get the overlapping instances warning.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -207,12 +220,20 @@ entails SolverOptions{..} constraint context hints =
...
               Just tcd -> pure tcd
-            let instances =
-                  [ (substs, tcd)
-                  | tcd <- forClassName (combineContexts context inferred) className' tys''
-                    -- Make sure the type unifies with the type in the type instance definition
-                  , substs <- maybeToList (matches typeClassDependencies tcd tys'')
-                  ]
+            let instances = do
+                  -- TODO: efficiency
+                  chain <- groupBy ((==) `on` tcdChain) $ forClassName (combineContexts context inferred) className' tys''
```
LiamGoodacre: The efficiency note here was to consider grouping using a Map instead.

garyb:
This only supports a single chain of instances per class then? It seems if you try to define multiple there's no error aside from the fact overlaps start occurring again. Maybe it should have an explicit error instead? (Or ideally support multiple chains 😜)

LiamGoodacre:
@garyb This is supposed to support multiple chains.  But, in the same way that you shouldn't define overlapping instances, you shouldn't define overlapping chains.

Although, I did just try to use overlapping chains and it failed to solve.

```purescript
class Learn a b | a -> b
instance learnInst :: Learn a String

class IsString t o | t -> o

instance isStringString :: IsString String True
else instance isStringElse :: IsString a False

instance isStringString' :: IsString String True
else instance isStringElse' :: IsString a False

learnIsString :: forall a t o.
  IsString t o =>
  Learn a t =>
  Proxy o
learnIsString = Proxy

isStringEg0 :: Proxy True
isStringEg0 = learnIsString
```

With the output:

```purescript
[1/1 OverlappingInstances]

    Foo.IsString t0 True

    Foo.isStringElse (chosen)
    Foo.isStringElse'

  Overlapping type class instances can lead to different behavior based on the order of module imports, and for that reason are not recommended.
  They may be disallowed completely in a future version of the compiler.

  where t0 is an unknown type

[1/1 TypesDoNotUnify] src/Foo.purs:71:15

  71  isStringEg0 = learnIsString

  Could not match type

    False

  with type

    True

    Foo.IsString t0 True
```

If I don't have the duplicate chain, then it passes.

```purescript
--instance isStringString' :: IsString String True
--else instance isStringElse' :: IsString a False
```

So something is wrong 😄

garyb:
Hmm I think I see what's happening with the case I tried then, I'd need to be cleverer with the definitions (or maybe it's impossible)... I tried writing an `Inject` instance for second type isomorphic to `Coproduct`, and including any of the instances for `Coproduct2` results in an overlap in a test case where I was trying to inject just into `Coproduct`.

LiamGoodacre:
Turns out that `groupBy` doesn't do what I thought it did 😄

LiamGoodacre:
Ummmmm, travis is complaining about code that doesn't exist? @paf31

```
/home/travis/build/purescript/purescript/src/Language/PureScript/AST/Declarations.hs:534:11: error:
    • The constructor ‘TypeInstanceDeclaration’ should have 8 arguments, but has been given 6
    • In the pattern: TypeInstanceDeclaration _ n _ _ _ _
      In an equation for ‘declName’:
          declName (TypeInstanceDeclaration _ n _ _ _ _) = Just (IdentName n)
```

LiamGoodacre:
Or am I completely crazy?

LiamGoodacre:
Oh, did it rebase my commits onto master and then try to build? 😕

LiamGoodacre:
Okay, I've rebased and fixed the issue that Travis was complaining about.

hdgarrood:
> Oh, did it rebase my commits onto master and then try to build?

I think CI will usually merge your branch into master (locally, ie without modifying actual master in the repo), and attempt the build that. I think that is probably what it should be doing, too.

paf31:
@LiamGoodacre Could you please rebase this?

LiamGoodacre:
Ah rebased at the wrong time, need to do it again 😄

LiamGoodacre:
This should be good to go now.

PR #2931, opened 2017-06-10
[purs ide] improve reexport bundling

kRITZCREEK:
- Fix encoding of exportedFrom field
  It used the derived json encoding for module names before, which is not what we
  want for the editors

- run reexportGrouping before applying maxResults

/cc @nwolverson


nwolverson:
Works for me, the limit definitely applies post-grouping 
screen shot 2017-06-10 at 10 44 11

PR #2932, opened 2017-06-10
Warn for type constructor aliases w/ suggestions

justinwoo:
This is a PR for the slightly crazy idea I had to ban simple* type aliases. I chose to just make a warning for now with a suggestion that writes out a newtype form.

\* In this case just meaning "is a type constructor on its own"

In screenshots:

screen shot 2017-06-10 at 2 56 49 pm
screen shot 2017-06-10 at 2 56 55 pm
screen shot 2017-06-10 at 2 57 03 pm

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -84,6 +84,7 @@ errorCode em = case unwrapErrorMessage em of
   CannotGetFileInfo{} -> "CannotGetFileInfo"
   CannotReadFile{} -> "CannotReadFile"
   CannotWriteFile{} -> "CannotWriteFile"
+  TypeConstructorAlias{} -> "TypeConstructorAlias"
   InfiniteType{} -> "InfiniteType"
   InfiniteKind{} -> "InfiniteKind"
   MultipleValueOpFixities{} -> "MultipleValueOpFixities"
@@ -304,6 +305,7 @@ errorSuggestion err =
       HidingImport mn refs -> suggest $ importSuggestion mn refs Nothing
       MissingTypeDeclaration ident ty -> suggest $ showIdent ident <> " :: " <> T.pack (prettyPrintSuggestedType ty)
       WildcardInferredType ty _ -> suggest $ T.pack (prettyPrintSuggestedType ty)
+      TypeConstructorAlias name args ty -> suggest $ T.pack (prettyPrintNewtypeForType name args ty)
       _ -> Nothing
   where
     emptySuggestion = Just $ ErrorSuggestion ""
@@ -481,6 +483,8 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
       line "The last statement in a 'do' block must be an expression, but this block ends with a let binding."
     renderSimpleErrorMessage OverlappingNamesInLet =
       line "The same name was used more than once in a let binding."
+    renderSimpleErrorMessage (TypeConstructorAlias _ _ _) =
+      line "This type synonym aliases a type constructor. Consider using a newtype or using the original type instead."
     renderSimpleErrorMessage (InfiniteType ty) =
       paras [ line "An infinite type was inferred for an expression: "
             , markCodeBox $ indent $ typeAsBox ty
--- a/src/Language/PureScript/Pretty/Types.hs
+++ b/src/Language/PureScript/Pretty/Types.hs
@@ -7,6 +7,7 @@ module Language.PureScript.Pretty.Types
   , prettyPrintType
   , prettyPrintTypeWithUnicode
   , prettyPrintSuggestedType
+  , prettyPrintNewtypeForType
   , typeAtomAsBox
   , prettyPrintTypeAtom
   , prettyPrintRow
@@ -217,6 +218,17 @@ prettyPrintTypeWithUnicode = prettyPrintType' unicodeOptions
 -- | Generate a pretty-printed string representing a suggested 'Type'
 prettyPrintSuggestedType :: Type -> String
 prettyPrintSuggestedType = prettyPrintType' suggestingOptions
+
+-- | Generate a pretty-printed string newtype
+prettyPrintNewtypeForType :: ProperName 'TypeName -> [(Text, Maybe Kind)] -> Type -> String
+prettyPrintNewtypeForType (ProperName name) args ty =
+  "newtype " ++ name' ++ args' ++ " = " ++ name' ++ " " ++ ctr
+  where
+    name' = T.unpack name
+    args' = (" " ++) . T.unpack . fst =<< args
+    ctr = case ty of
+      TypeConstructor (Qualified _ (ProperName tn)) -> T.unpack tn
+      _ -> error "Type constructor was expected here."
 
 prettyPrintType' :: TypeRenderOptions -> Type -> String
 prettyPrintType' tro = render . typeAsBoxImpl tro
--- a/examples/warning/TypeConstructorAlias.purs
+++ b/examples/warning/TypeConstructorAlias.purs
@@ -1,0 +1,4 @@
+-- @shouldWarnWith TypeConstructorAlias
+module Main where
+
+type TypeConstructorAlias = String
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -616,6 +616,7 @@ extra-source-files:
       examples/warning/ShadowedBinderPatternGuard.purs
       examples/warning/ShadowedNameParens.purs
       examples/warning/ShadowedTypeVar.purs
+      examples/warning/TypeConstructorAlias.purs
       examples/warning/UnnecessaryFFIModule.js
       examples/warning/UnnecessaryFFIModule.purs
       examples/warning/UnusedDctorExplicitImport.purs
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -70,6 +70,7 @@ data SimpleErrorMessage
   | CannotGetFileInfo FilePath
   | CannotReadFile FilePath
   | CannotWriteFile FilePath
+  | TypeConstructorAlias (ProperName 'TypeName) [(Text, Maybe Kind)] Type
   | InfiniteType Type
   | InfiniteKind Kind
   | MultipleValueOpFixities (OpName 'ValueOpName)
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -163,6 +163,16 @@ addTypeClassDictionaries mn entries =
   modify $ \st -> st { checkEnv = (checkEnv st) { typeClassDictionaries = insertState st } }
   where insertState st = M.insertWith (M.unionWith M.union) mn entries (typeClassDictionaries . checkEnv $ st)
 
+checkTypeConstructorAlias
+  :: (MonadState CheckState m, MonadWriter MultipleErrors m)
+  => ProperName 'TypeName
+  -> [(Text, Maybe Kind)]
+  -> Type
+  -> m ()
+checkTypeConstructorAlias name args ty@(TypeConstructor _) = do
+  tell . errorMessage $ TypeConstructorAlias name args ty
+checkTypeConstructorAlias _ _ _ = return ()
+
 checkDuplicateTypeArguments
   :: (MonadState CheckState m, MonadError MultipleErrors m)
   => [Text]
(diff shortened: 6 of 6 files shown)


rightfold:
Could you explain what "simple" means?

kRITZCREEK:
Does this also warn in situations where we're using the synonym arguments in multiple positions, or are applying parts of the aliased type constructor?

For example in Halogen:

```purescript
type ParentHTML f g p m = HTML (ComponentSlot HTML g m p (f Unit)) (f Unit)
```

hdgarrood:
I think `type Pair a = Tuple a a` should be allowed without a warning, and as far as I can tell it would produce a warning with this code. What do we think about only warning when the type arguments list is empty?

rightfold:
Check should be applied after eta reduction.

paf31:
I'm not sure I agree that aliases for type constructors are always a bad thing. What if I just want to use a type synonym to make a name shorter? Or what if I just want to use a new naming convention without changing semantics?

```purescript
-- Maybe if I work with lots of Scala developers ...
type Option = Maybe
```

@hdgarrood I don't think your example would be an issue here since `Tuple a a` is a type application, not just a type constructor.

justinwoo:
Aren't all type constructor aliases confusing? I think any case of `Option : Maybe` or `Result : Either` or whatnot would just end up being confusing since the constructor names don't match either, and you'd very quickly(?) run into cases where you'd need to start using the concrete constructors (at least, for easy to understand code)

I hope to not warn on type applications though, even if it's as simple as `Maybe String`, since I have many cases where I use type aliases for those like `type FilesRoute = Route Unused FilesData`

hdgarrood:
Ah I see, thanks.

Having thought about this a little more, and similar to what @paf31 said, I am a little uneasy about putting this into the compiler as a warning. I agree that this kind of type alias is almost always a bad idea, but also I'm not sure I'm comfortable having the compiler generate a warning about it in every case.

paf31:
I don't think we should try to prevent users from renaming things.

jdegoes:
I agree with @justinwoo. A type alias is not a rename because it does not create appropriate constructors and deconstructors. In practice, type aliases like this are used in places where a `newtype` would be superior, because of the added type safety, yet there is reluctance to introduce a `newtype` because of the attendant ceremony. Warning on silly type aliases like this and focusing on reducing that ceremony (which is already pretty minimal at this point) seems like it can only improve the quality of the code.

garyb:
I'm also not a fan of `type`-renaming things when it's purely a rename. For all the above reasons, plus they make error messages worse, since the error will still talk about `Maybe` even if the `Option` synonym was used in the code.

paf31:
Every type alias is just renaming, so by this logic, they're all bad, right? I would say lenses give a nice example where aliases are doing something worthwhile, but even then you have all of the same error message issues.

I agree it's almost never a good practice to use aliases like this, but then what is a good use case for an alias? Can we characterize those? If not, I'd rather not add a warning for something so specific.

jdegoes:
> Every type alias is just renaming, so by this logic, they're all bad, right? I would say lenses give a nice example where aliases are doing something worthwhile, but even then you have all of the same error message issues.

In my opinion, a type alias is only useful for the sole case in which you are extracting type duplication from type declarations. i.e. it's the moral equivalent of using a let binding for an expression to avoid repeating the expression in multiple places. In this case the name is not useful, per se, it's almost irrelevant, the point of using the alias is to remove the duplication to make maintenance easier.

All other uses of type aliases are design smells, and warning against them would be preferable to me in any code base I work on or oversee.

I think you can detect this with relatively simple logic: _a type alias that does not apply type parameters to some higher-kinded type constructor is always bad_.

garyb:
> I would actually seriously consider just removing type aliases completely 😄

I find that idea somewhat horrifying 😆. There are cases where having aliases improves readability of types hugely, reducing duplication or avoiding the need to write out entire complicated constructions.

I think the suggestion here is just `type X = Y` cases, when there is no type application at all. `type File = String`, `type Length = Number`, etc.

paf31:
I don't think it's possible to characterize bad aliases, but I agree that `Alias = TypeConstructor` is almost always bad, and probably worth a warning.

However, we should probably handle things like `type MyList a = List a` as well.

garyb:
> However, we should probably handle things like `type MyList a = List a` as well.

Definitely! I have a branch with a warning for "over applied" aliases that I did months ago, I'll spruce it up and get it open as a WIP at least.

src/Language/PureScript/Pretty/Types.hs
```diff
@@ -218,6 +219,19 @@ prettyPrintTypeWithUnicode = prettyPrintType' unicodeOptions
...
+prettyPrintNewtypeForType (ProperName name) args ty =
+  "newtype " ++ name' ++ args' ++ " = " ++ name' ++ " " ++ ctr
+  where
+    name' = T.unpack name
+    args' = case args of
+      [] -> ""
+      xs -> " " ++ (intercalate " " $ T.unpack . fst <$> xs)
+    ctr = case ty of
+      TypeConstructor (Qualified _ (ProperName tn)) -> T.unpack tn
+      _ -> ""
```
paf31: Can we use `error` here?

src/Language/PureScript/Pretty/Types.hs
```diff
@@ -218,6 +219,19 @@ prettyPrintTypeWithUnicode = prettyPrintType' unicodeOptions
...
 prettyPrintSuggestedType = prettyPrintType' suggestingOptions
 
+-- | Generate a pretty-printed string newtype
+prettyPrintNewtypeForType :: ProperName 'TypeName -> [(Text, Maybe Kind)] -> Type -> String
+prettyPrintNewtypeForType (ProperName name) args ty =
+  "newtype " ++ name' ++ args' ++ " = " ++ name' ++ " " ++ ctr
+  where
+    name' = T.unpack name
+    args' = case args of
+      [] -> ""
```
paf31: Same here

src/Language/PureScript/Errors.hs
```diff
@@ -84,6 +84,7 @@ errorCode em = case unwrapErrorMessage em of
   CannotGetFileInfo{} -> "CannotGetFileInfo"
   CannotReadFile{} -> "CannotReadFile"
   CannotWriteFile{} -> "CannotWriteFile"
+  TypeConstructorAlias{} -> "TypeConstructorAlias"
```
paf31: After this is merged, can you please add docs to the docs repo for this error too?

src/Language/PureScript/Errors.hs
```diff
@@ -481,6 +483,8 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
       line "The last statement in a 'do' block must be an expression, but this block ends with a let binding."
     renderSimpleErrorMessage OverlappingNamesInLet =
       line "The same name was used more than once in a let binding."
+    renderSimpleErrorMessage (TypeConstructorAlias _ _ _) =
+      line "The type synonym provided was an alias for a Type Constructor. These kinds of types can lead to confusing programs."
```
paf31: Can we say this?

```
This type synonym aliases the type constructor XYZ. Consider using a newtype instead.
```

"confusing programs" is a bit vague, and we can give examples in the docs repo instead.

hdgarrood: Perhaps "consider using a newtype or using the original type instead"?

paf31: 👍

joneshf:
Throwing my hat in. I'd be really interested to see how removal of type aliases played out. I feel like life would be better overall. It would push us to make using newtypes even easier to work with; they're already pretty easy to use with newtype deriving and `Data.Newtype`. But also, it would eliminate a bad design choice that we only suggest to  you shouldn't do. Sort of like orphan instances.

parsonsmatt:
I strongly disagree with removing type aliases entirely. They're an extremely useful tool in Haskell development. Removing them would make development with eg Servant or Lens totally untenable. It would require us to write `StateT s Identity a` everywhere instead of `State s a` (etc). 

I would be annoyed to see the original post's warnings from the compiler, mostly because it seems more like an appropriate thing for a linter tool to have as an option. I use type aliases like these somewhat frequently in development when I expect the type of some related group of functions to change, or if I'm not certain of it. For example:

```haskell
type Filter = Foo -> Bar -> Either [String] Quux

hasThing :: Filter
hasWhatever :: Filter
hasSomeOther :: Filter
-- repeat about 40 times
```

then a change to the signature (eg, `Either [String] Quux -> AccValidation [Error] Quux`) becomes much easier without having to resort to `sed` trickery.

vendethiel:
I think the warning is just to alias types like Number or String

paf31:
> I strongly disagree with removing type aliases entirely.

Don't worry, I don't think anyone is seriously considering it 😄 

> because it seems more like an appropriate thing for a linter tool to have as an option

Now that you bring it up, having this as a linter option makes a lot of sense. The question is, should all linter options be baked into the compiler as warnings? Without the ability to filter them on the editor side, I think probably not.

justinwoo:
@nwolverson's VSCode/Atom editor extensions let you exclude specific warnings at-will, and I've used this to not warn on inferred wildcard and function signatures for my tutorial videos to not have too much noise, e.g. https://twitter.com/jusrin00/status/850375735864950784

Probably someone should make some easy way for IDE server clients to get this list of warnings?...

natefaubion:
I'm not in favor of this as a warning. Which programs are confusing? If we preserved aliases after desugaring, would they still be confusing? What bugs would be prevented by following this warning?

natefaubion:
I feel like this strongly implies we should also warn for aliases at the value level. Maybe someone wants to preserve backwards compatibility under a rename?

hdgarrood:
> Maybe someone wants to preserve backwards compatibility under a rename?

Very good point. This is the strongest argument against this warning I've seen so far, I think.

natefaubion:
If we preserved synonyms after desugaring, which is a longstanding issue, then the errors could probably be far less confusing, and I think this warning would be moot:
`Could not match Int with Pattern (ie. String)` or something.

jdegoes:
> Maybe someone wants to preserve backwards compatibility under a rename?

That won't work because you won't get constructors / deconstructors just from a type alias. In any case, you're arguing to avoid making something a warning to allow teams to increase technical debt. I'd say that's an excellent argument for making it a warning. 😉

src/Language/PureScript/Errors.hs
justinwoo: Yes, definitely needs some docs

src/Language/PureScript/Pretty/Types.hs
justinwoo: Ah, yeah, definitely.

src/Language/PureScript/Pretty/Types.hs
justinwoo: In this section, I just want to print out an empty string if I don't have any type arguments to my type synonym. Should I rewrite this section in some way?

hdgarrood:
The whole point of the technical debt metaphor is that it's sometimes reasonable to take out a loan as long as you pay it off later.

paf31:
Aren't all type operator aliases just as bad as these sorts of aliases?

garyb:
I don't think so - operators don't attempt to convey meaning, the way `type File = String` suggests.

garyb:
Maybe this is controversial enough that we don't do it though. Seems like there's not a consensus here, and it is the kind of thing a separate linter could deal with. It's not so much a "it's likely you're making a mistake"-level lint warning as it is a "that's a questionable choice".

paf31:
I agree that `type File = String` and `type Beside = Coproduct` are different, but distinguishing them isn't really something the compiler can do, since it would need to know the user's _intent_.

justinwoo:
I'd be okay with shelving this for now, but I'd like to really try polling more users somehow about this. Type aliasing constructors is a very common way I've shot my feet off in the past (and sometimes in the present/future), and it seem like something where 90%+ of the cases would benefit from this being a warning. Maybe needs a "maybe later" tag or something 😛

paf31:
@justinwoo Could you please create an issue?

jdegoes:
> type Filter = Foo -> Bar -> Either [String] Quux

This is takes 0 type parameters and applies 6, so would be valid under my proposed rule for bad type aliases. Applying more than you require is a sign you're doing work, i.e. reducing duplication. Applying less than or the same as what you require is a sign you're doing no work, i.e. it's purely "semantic" change with no actual benefit to reducing duplication.

paf31:
Closing as this needs an issue to discuss, but let's reopen this if we come to a decision. Thanks!

PR #2933, opened 2017-06-10
Add source span annotations to DeclarationRef

garyb:
There's a couple of things I need to finish up in here, but it's mostly there.

This isn't general purpose annotations, but I think given previous discussions we may as well have add source position info to the ASTs, since we always want that, and don't want to have to deal with partiality in extracting it from annotations an whatnot. Also means we won't have to deal with the precarious `PositionedN` constructors anymore.

I chose `DeclarationRef` to start with as it's "easy"... turned out, not so much. But at least now this sets up a bunch of stuff for the way forward. Ideally in time we won't have the old `errorMessage` function at all, and it will only be possible to throw with a `SourceSpan`.

One thing I don't love is how I dealt with the annotations in parsing though, any ideas there would be welcome.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -191,7 +191,7 @@ errorMessage err = MultipleErrors [ErrorMessage [] err]
 
 -- | Create an error set from a single simple error message and source annotation
 errorMessage' :: SourceSpan -> SimpleErrorMessage -> MultipleErrors
-errorMessage' ss err = MultipleErrors [ErrorMessage [] err]
+errorMessage' ss err = MultipleErrors [ErrorMessage [PositionedError ss] err]
 
 -- | Create an error set from a single error message
 singleError :: ErrorMessage -> MultipleErrors
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -189,6 +189,10 @@ nonEmpty = not . null . runMultipleErrors
 errorMessage :: SimpleErrorMessage -> MultipleErrors
 errorMessage err = MultipleErrors [ErrorMessage [] err]
 
+-- | Create an error set from a single simple error message and source annotation
+errorMessage' :: SourceSpan -> SimpleErrorMessage -> MultipleErrors
+errorMessage' ss err = MultipleErrors [ErrorMessage [] err]
+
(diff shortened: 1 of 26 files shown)


src/Language/PureScript/Linter/Imports.hs
```diff
@@ -69,22 +69,20 @@ lintImports (Module ss _ mn mdecls (Just mexports)) env usedImps = do
...
-        maybe id warnWithPosition ss' $ do
-          let names = ordNub $ M.findWithDefault [] mni usedImps'
-              usedRefs = findUsedRefs env mni (Just mnq) names
-          unless (null usedRefs) $
-            tell $ errorMessage $ ImplicitQualifiedImport mni mnq usedRefs
+      for_ implicits $ \(ss', _, mni) -> do
+        let names = ordNub $ M.findWithDefault [] mni usedImps'
+            usedRefs = findUsedRefs env mni (Just mnq) names
+        unless (null usedRefs) .
+          tell . errorMessage' ss $ ImplicitQualifiedImport mni mnq usedRefs
```
paf31: Should this be `ss'`?

src/Language/PureScript/Errors.hs
```diff
@@ -189,6 +189,10 @@ nonEmpty = not . null . runMultipleErrors
 errorMessage :: SimpleErrorMessage -> MultipleErrors
 errorMessage err = MultipleErrors [ErrorMessage [] err]
 
+-- | Create an error set from a single simple error message and source annotation
+errorMessage' :: SourceSpan -> SimpleErrorMessage -> MultipleErrors
+errorMessage' ss err = MultipleErrors [ErrorMessage [] err]
```
paf31: You don't use `ss` here - is that deliberate?

paf31:
👍 Looks great so far!

src/Language/PureScript/Errors.hs
garyb: Yeah, that's one of the bits I hadn't finished up :) I left it bound but unused so the warning would occur.

src/Language/PureScript/Linter/Imports.hs
garyb: Yes, good catch!

garyb:
This is ready for "final" review now I think.

Issue #2924, opened 2017-06-04
[purs ide] Add a declaration type Filter


kRITZCREEK:
Add a filter which allows to filter for the type of declaration.

## Usecases

I'll list a couple of example use cases that would use this feature.

### Add type class instance

An editor could provide an `add type class instance` function to the user which inserted a snippet like:

`instance ?t?c :: ?c ?t where`

which would then prompt for a completion for the class name, filtered to only contain type class names. It would then use that completion to fill in the `?c` slots. Next it could check whether that type class is defined in the current module. If it isn't:

Ask the user for a type completion, with types defined in the current module (importantly we don't want type synonyms here) and use the result to fill in the `?t` slot.

If the type class is defined in the current module, just prompt for all the types defined in the current project and add an import on selection (filter the import addition down to the type declaration right away, so we don't end up with data constructors or kinds with the same name).

## Add deriving clauses
Very similar to `add type class`

## Adding more specific filters

Filtering options that are specific to some kind of declaration could live within the declaration type filters. For example, a filter which only keeps types of a certain kind, or a filter which only keeps values that mention a certain type in their signature, etc.

# Implementation

To not have to break the format as soon as we start adding more declaration specific filters I propose to not just use a non-empty array of strings for the declaration types, but to wrap them in objects right away. Something like:

```json
{
  "filter": "declaration",
  "params": [
    { 
      "declarationtype": "synonym"
    },
    {
      "declarationtype": "value"
    },
    {
      "declarationtype": "typeclass"
    }
  ]
}
```

As this filter is strictly more general than the namespace filter, I propose we implement the namespace filter in terms of the declaration filter.

This is a good follow-up to the namespace filter so I'd reserve this one for @stefanholzmueller or @sectore, if you want it.

stefanholzmueller:
Thanks for mentioning me, but I do not have enough spare time, at least in the next two weeks. So I'm fine if someone else picks this up. :)

sectore:
@kRITZCREEK Nice, I will take over this part. Thx!

kRITZCREEK:
Fixed in #2934

PR #2934, opened 2017-06-11
[purs ide] Add declaration type filter (#2924)

sectore:


psc-ide/PROTOCOL.md
```diff
@@ -560,6 +560,43 @@ Valid namespaces are `value`, `type` and `kind`.
...
+      "declarationtype": "value"
+    },
+    {
+      "declarationtype": "type"
+    },
+    {
+      "declarationtype": "synonym"
+    },
+    {
+      "declarationtype": "constructor"
```
kRITZCREEK: I'd prefer dataconstructor here, since we also have type constructors, which could make this slighty confusing.

src/Language/PureScript/Ide/Filter.hs
```diff
@@ -77,6 +79,15 @@ declarationFilter :: (IdeDeclaration -> Text -> Bool) -> Text -> [Module] -> [Mo
...
   filterModuleDecls (flip predicate search)
 
+-- | Only keeps Identifiers in the given type declarations
+declarationTypeFilter :: [D.IdeDeclaration] -> Filter
+declarationTypeFilter [] = mkFilter identity
+declarationTypeFilter decls =
+    mkFilter $ filterModuleDecls filterDecls
+    where
+      filterDecls :: IdeDeclaration -> Bool
+      filterDecls decl = D.typeDeclarationForDeclaration decl `elem` decls
```
kRITZCREEK: I think it would be nicer if we could move the filtering logic into `Language.PureScript.Ide.Filter.Declaration`. I think it could expose a function of type `[D.IdeDeclaration]-> Map ModuleName IdeDeclarationAnn -> Map ModuleName IdeDeclarationAnn` which we could then simply wrap up with mkFilter in here.

This way we can keep the Declarationfilter specific logic that we plan to add in the new added module.

src/Language/PureScript/Ide/Filter.hs
```diff
@@ -103,10 +114,13 @@ instance FromJSON Filter where
         return $ prefixFilter search
       "modules" -> do
         params <- o .: "params"
-        modules <- map P.moduleNameFromString <$> params .: "modules"
+        modules <- fmap P.moduleNameFromString <$> params .: "modules"
```
kRITZCREEK: not necessary, since `Protolude` exports the `Functor` map as `map`

src/Language/PureScript/Ide/Filter/Declaration.hs
```diff
@@ -0,0 +1,57 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE DeriveFoldable  #-}
```
kRITZCREEK: Do we really need these extensions? Now that the lens deriving is gone I think we can get rid of these.

src/Language/PureScript/Ide/Filter/Declaration.hs
```diff
@@ -0,0 +1,57 @@
...
+  | Synonym
+  | DataConstructor
+  | TypeClass
+  | ValueOperator
+  | TypeOperator
+  | Kind
+  deriving (Show, Eq, Ord)
+
+instance FromJSON DeclarationType where
+  parseJSON (String s) = case s of
```
kRITZCREEK: This can be written with aeson's `withText` function to give a better parse error.

`parseJson = withText "Declaration type tag" $ \t -> case t of`

src/Language/PureScript/Ide/Filter.hs
sectore: @kRITZCREEK Unfortunately I can't see what you mean here. Because `mkFilter` takes a function `([Module] -> [Module])`, but in your case it would be `Map ModuleName `???

Also `filterModuleDecls` as a part of `declarationTypeFilter` is already in `Language.PureScript.Ide.Filter` and it is used by other functions. Why should it moved (or maybe duplicate) into `Language.PureScript.Ide.Filter.Declaration`?

Could you give more information on this, pls?

P.S. Everything else is clear and has been already fixed with latest commit 7c680bd ;)

kRITZCREEK: Oh yes sorry, the type would be `[Module] -> [Module]`. The idea for creating the new module was so that we could split out the logic around the new declaration filters, because I intend to continue adding to them in the new future. If think it's totally fine to move `filterModuleDecls` into `Ide.Filter.Declaration` and import it from `Ide.Filter`.

sectore: @kRITZCREEK Moving `filterModuleDecls` into `Ide.Filter.Declaration` will cause a cycle dependency between `Ide.Filter` and `Ide.Filter.Declaration`. Because `Module` is defined in `Ide.Filter` and has to be imported into `Ide.Filter.Declaration` and `filterModuleDecls` has to be imported into `Ide.Filter`. To solve this cycle dependency we have to move less or more types etc. into other places (e.g. into a new module `Ide.Filter.Types` or similar), which "breaks" current module structure.

I would suggest that you move all these functions / types into other places as you like, but by creating another (new) issue for this. Just to avoid that current module structure will "destroyed" by my attempt (and not to waste to much time on this, of course). Thx!

sectore: @kRITZCREEK Could you re-start CI builds of travis, please? It seems that there are some `timed out` issues, e.g. https://travis-ci.org/purescript/purescript/jobs/242937235#L667 and here https://travis-ci.org/purescript/purescript/jobs/242937236#L809 

Also I haven't seen the error `Segmentation fault/access violation in generated code` before by running AppVendor: https://ci.appveyor.com/project/hdgarrood/purescript-lfgff/build/2447#L772  Any idea?

Thx :)

kRITZCREEK: Don't worry about the CI failures... :( I've checked your branch out locally and all the tests pass so I'm happy. Thanks for working on this, I'll get this merged then :)

PR #2935, opened 2017-06-12
Add source span annotations to Declaration

garyb:
Approx 1,000,000 things left to re-examine still. I got bored dealing with `Parser` hence the `undefined`s for now :stuck_out_tongue:

### DIFF
--- a/tests/Language/PureScript/Ide/SourceFileSpec.hs
+++ b/tests/Language/PureScript/Ide/SourceFileSpec.hs
@@ -49,7 +49,6 @@ member1 = P.TypeDeclaration ann2 (P.Ident "member1") P.REmpty
 spec :: Spec
 spec = do
   describe "Extracting Spans" $ do
-    --TODO-gb: is this worth testing now? the spans part not so much, but returning the appropriate IdeNSValue/IdeNSType is useful perhaps?
     it "extracts a span for a value declaration" $
       extractSpans value1 `shouldBe` [(IdeNamespaced IdeNSValue "value1", span1)]
     it "extracts a span for a type synonym declaration" $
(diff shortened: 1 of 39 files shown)


garyb:
I decided to go from the top this time, with the `Declaration`s, as it should be possible to eliminate the non-`SourceSpan`-accepting `errorMessage` right now. We'll always be guaranteed to have at least a vaguely local source span available if `Declaration`s have it.

tests/Language/PureScript/Ide/SourceFileSpec.hs
```diff
@@ -16,55 +16,62 @@ span0 = P.SourceSpan "ModuleLevel" (P.SourcePos 0 0) (P.SourcePos 1 1)
...
-member1 = P.TypeDeclaration (P.Ident "member1") P.REmpty
+foreign1 = P.ExternDeclaration ann1 (P.Ident "foreign1") P.REmpty
+foreign2 = P.ExternDataDeclaration ann1 (P.ProperName "Foreign2") P.kindType
+foreign3 = P.ExternKindDeclaration ann1 (P.ProperName "Foreign3")
+member1 = P.TypeDeclaration ann2 (P.Ident "member1") P.REmpty
 
 spec :: Spec
 spec = do
   describe "Extracting Spans" $ do
+    --TODO-gb: is this worth testing now? the spans part not so much, but returning the appropriate IdeNSValue/IdeNSType is useful perhaps?
```
garyb: What say you @kRITZCREEK?

kRITZCREEK: These tests have caught the stupidest mistakes before, so I'll say they're worth it. At the very least they make sure all the pattern matches are exhaustive.

garyb:
This is ready for review now.

I've not fully eliminated the old `errorMessage` yet - the remaning cases could be fixed now, but would involve some work that will just be undone again in the next PR, which will be for annotating expressions/binders.

I think there may be two cases now that are potentially unpositioned, as I removed a `rethrow`, but those will be fixed again anyway once `errorMessage` is done.

paf31:
👍 Fantastic, thanks!

PR #2942, opened 2017-06-16
[purs ide] Repopulates the rebuild cache when populating volatile state

kRITZCREEK:
Fixes a bug introduced when refactoring the IdeState type. Before this PR the reload after every successful rebuild would wipe the rebuild cache, and we wouldn't see any module local completions.


kRITZCREEK:
Thanks!

Issue #2941, opened 2017-06-14
Instance signatures don't bring type variables into scope


matthewleon:
```purescript
class Test t where
  test :: forall a. t -> a

instance testUnit :: Test Unit where
  test :: forall a. Unit -> a
  test = go
    where
      go :: Unit -> a
      go = unsafeCoerce
```

Error:

```
  Type variable a is undefined.

while checking the kind of Unit -> a0
while checking that expression let                
                                 go = unsafeCoerce
                               in go              
  has type Unit -> a0
in value declaration testUnit

where a0 is a rigid type variable
        bound at line 10, column 3 - line 10, column 20
```

garyb:
I wondered about this myself, as if it worked it would make instance signatures way more useful than "just" being there for documentation!

matthewleon:
Looks like `introduceSkolemScope` should be getting called, but isn't.

matthewleon:
Looks like the problem is more general and applies to records as well, though not sure if we consider it a bug there. It comes in once the typeclass is desugared:

```purescript
rec :: {attr :: forall a. a -> Int}
rec = {attr: let go :: a -> Int
                 go = \_ -> 0 
             in go
      }
```

This will fail for the same reason. http://try.purescript.org/?session=4e9f8283-f493-b286-5c42-0515f39cb415

matthewleon:
To clarify, I guess this isn't really a _problem_ with records themselves, per se. It's just that the desugaring mechanism turns the TC instances into records, which don't have a mechanism for specifying that the type variables from the property types should be brought into scope.

This could be resolved by desugaring instances with signatures into `let`s. But that results in wrapping the JS output for that property in a redundant anonymous function invocation. There should be a better solution.

hdgarrood:
I think this example

```purescript
rec :: {attr :: forall a. a -> Int}
rec = {attr: let go :: a -> Int
                 go = \_ -> 0 
             in go
      }
````

should fail; I think that type should be treated differently from `rec :: forall a. { attr :: a -> Int }`. Consider the following type declaration:

```purescript
rec :: {attr :: forall a. a -> Int, attr2 :: forall a. a -> Int }
```

If we automatically hoisted the `forall a.` to the start of the type, what could we do in cases like this one? I don't think there's a good answer.

paf31:
I don't think we should do this by adding any special cases for polymorphic types in records. Maybe we need to rethink the way the type class desugaring works.

LiamGoodacre:
[...]

PR #2943, opened 2017-06-16
Instance signatures scoped type variables

matthewleon:
Addresses https://github.com/purescript/purescript/issues/2941

Typeclass instances are desugared into records. When an instance is provided with a type signature, that creates a `TypedValue` property. This pull request detects such creatures, and brings the corresponding type variables into scope.

### DIFF
--- a/examples/passing/InstanceSigsScopedTypeVariables.purs
+++ b/examples/passing/InstanceSigsScopedTypeVariables.purs
@@ -1,0 +1,23 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+
+class Foo t where
+  foo :: forall a. a -> t
+
+instance fooUnit :: Foo Unit where
+  foo :: forall a. a -> Unit
+  foo = go
+    where
+      go :: a -> Unit
+      go = \_ -> unit
+
+instance fooNumber :: Foo Int where
+  foo :: forall b. b -> Int
+  foo = go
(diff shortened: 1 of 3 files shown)


paf31:
What about doing this with a `let`?

matthewleon:
> What about doing this with a let?

That adds an IIFE to the JS output. It seems unfortunate for there to be a penalty, however tiny, for adding typing, no? Or is there a simple way to have the type checker simulate a `let` binding without actually transforming the imperative core output? I'd be happy to try if you don't see any potential hitches.

paf31:
Can you give an example of the generated code please?

matthewleon:
> Can you give an example of the generated code please?

I checked the JS output of the code posted by @LiamGoodacre (https://github.com/purescript/purescript/issues/2941#issuecomment-309122783) and realized that it looks like I'm wrong about the extra IIFE. I think desugaring explicit instances to something like what he posted (and then recursively desugaring that, of course) seems like a reasonable way to go.

I'd be very happy to leave this one to you guys from here on out, though. I don't feel like I have the necessary background or expertise to advocate one solution over the other, and my attempts at implementation end up being muddled and hackish. I enjoyed trying my hand, though :)

paf31:
I'm closing this one since I don't want to implement this as a special case on records. Also, I think we can fix it a different way, by using `let`.

PR #2944, opened 2017-06-16
Add Match class

rightfold:
Provides the `Prim.Match` type class, constraints of which are only satisfied when a symbol matches a regex in another symbol. This is useful for the pattern where you have an abstract newtype with a smart constructor, but want to allow the construction of literals without `unsafePartial fromJust`.

For example:

```purescript
module Data.Positive
  ( Positive
  , fromInt
  , lit
  ) where

import Data.Int as Int
import Data.Maybe (fromJust)
import Data.Symbol (class IsSymbol, reflectSymbol)
import Partial.Unsafe (unsafePartial)
import Prelude

newtype Positive = Positive Int

fromInt :: Int -> Maybe Positive
fromInt n | n < 1 = Nothing
          | otherwise = Just (Positive n)

lit :: forall s. IsSymbol s => Match "^[1-9][0-9]{0,8}$" s => @s -> Positive
lit = unsafePartial fromJust <<< fromInt <=< Int.fromString <<< reflectSymbol
```

Compare:

```purescript
import Data.Maybe (fromJust)
import Data.Positive (fromInt, lit)
import Partial.Unsafe (unsafePartial)
import Prelude

example1 :: Positive
-- example1 = unsafePartial fromJust $ fromInt 0 -- ugly and unsafe
example1 = unsafePartial fromJust $ fromInt 1 -- ugly but ok

example2 :: Positive
-- example2 = lit @"0" -- type error
example2 = lit @"1" -- 😍
```

### DIFF
--- a/examples/failing/Match.purs
+++ b/examples/failing/Match.purs
@@ -1,3 +1,4 @@
+-- @shouldFailWith NoInstanceFound
 module Main where
 
 import Prelude
@@ -12,6 +13,5 @@ lit = Lowercase <<< reflectSymbol
 
 main :: Eff (console :: CONSOLE) Unit
 main = do
-  _ <- pure <<< lit $ SProxy :: SProxy ""
-  _ <- pure <<< lit $ SProxy :: SProxy "foo bar baz"
+  _ <- pure <<< lit $ SProxy :: SProxy "foo BAR baz"
   log "Done"
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -205,6 +205,7 @@ extra-source-files:
       examples/failing/LetPatterns2.purs
       examples/failing/LetPatterns3.purs
       examples/failing/LetPatterns4.purs
+      examples/failing/Match.purs
       examples/failing/MissingClassExport.purs
       examples/failing/MissingClassMemberExport.purs
       examples/failing/MissingRecordField.purs
--- a/src/Language/PureScript/Constants.hs
+++ b/src/Language/PureScript/Constants.hs
@@ -387,6 +387,9 @@ pattern Union = Qualified (Just Prim) (ProperName "Union")
 pattern RowCons :: Qualified (ProperName 'ClassName)
 pattern RowCons = Qualified (Just Prim) (ProperName "RowCons")
 
+pattern Match :: Qualified (ProperName 'ClassName)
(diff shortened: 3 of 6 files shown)


paf31:
I quite like the idea of using this for some small metaprogramming tasks, at least as an experiment. But I think that's only possible if we extend this with an output parameter for the unconsumed suffix. Is that possible?

LiamGoodacre:
What do you think about having `Match pattern str groups` where `groups` is a typelevel symbol list (or row) of matching groups? 😄 

```purescript
Match "^(.)" "abc" ("a" : Nil)
```

paf31:
Hmm, how would groups work with rows? I guess we'd need some sort of named groups? Or we could exploit the fact that rows allow for duplicate labels.

paf31:
Another aspect to think about here is whether we really want/need full POSIX regex support. Maybe we can get by with our own well-defined subset, which would include literals, alternation, repetition and named capturing groups.

LiamGoodacre:
Also, should this live in `Prim` or in typelevel-prelude under `Type.Data.Symbol`?

garyb:
I really don't like this, personally. If we wanted to do something to enable enforceable newtype invariants at compile time, surely there is a less hacky solution.

paf31:
I'm not particularly interested in this for the newtype thing. I'm much more interested in using it as a form of meta programming.

rightfold:
PCRE supports named groups

hdgarrood:
Could you perhaps give an example of the kind of metaprogramming you have in mind?

I'm with @garyb re: using this for literals for values of user-defined types. Very few parsers can be adequately expressed in terms of regular expressions, in my experience.

paf31:
@hdgarrood Well, I could write a type-safe printf library, for example, by using this to lex a type-level format string at compile time, and then using a functional dependency to construct the type of the formatting function with the correct number of arguments.

paf31:
I think we should open an issue to discuss options here.

paf31:
I would quite like to implement a `SymbolAppend` class which could be run in both directions, either to append two symbols, or to remove a prefix or suffix from one. That should be enough to implement quite a lot of interesting things.

paf31:
Closing as this needs discussion, but we can reopen later if needed. Thanks!

PR #2945, opened 2017-06-16
Solve RowToList

LiamGoodacre:
Will add a test once https://github.com/purescript/purescript-typelevel-prelude/pull/11 is merged.

### DIFF
--- a/src/Language/PureScript/Constants.hs
+++ b/src/Language/PureScript/Constants.hs
@@ -359,6 +359,20 @@ orderingEQ = Qualified (Just typeDataOrdering) (ProperName "EQ")
 orderingGT :: Qualified (ProperName 'TypeName)
 orderingGT = Qualified (Just typeDataOrdering) (ProperName "GT")
 
+-- Type.Row
+
+pattern TypeRow :: ModuleName
+pattern TypeRow = ModuleName [ProperName "Type", ProperName "Row"]
+
+pattern RowToList :: Qualified (ProperName 'ClassName)
+pattern RowToList = Qualified (Just TypeRow) (ProperName "RowToList")
+
+pattern RowListNil :: Qualified (ProperName 'TypeName)
+pattern RowListNil = Qualified (Just TypeRow) (ProperName "Nil")
+
+pattern RowListCons :: Qualified (ProperName 'TypeName)
+pattern RowListCons = Qualified (Just TypeRow) (ProperName "Cons")
+
 -- Main module
 
 main :: forall a. (IsString a) => a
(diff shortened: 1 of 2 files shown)


paf31:
Looks good! What do you think about including the types along with the labels?

LiamGoodacre:
Yeah, I'm not opposed.  The two are equivalent in that we can use rowcons to look up the types.  It'd probably be more convenient to have both though.  In that case, what do you think about the names `RowEntries`/`EntryList`?

paf31:
I was thinking `RowToList` originally. I'm hoping that we can build up a library of these sorts of classes: folds, maps, etc. once we have constraint kinds so that we can take a MPTC as an argument:

```purescript
class RowMap f xs ys | f xs -> ys

instance auto1 :: RowMap f () ()
instance auto2 :: (f a b, RowMap f r1 r2) => RowMap f (l :: a | r1) (l :: b | r2)
```

LiamGoodacre:
Oh instead of two classes we can make have the fundeps in both directions `RowToList r l | r -> l, l -> r`.

paf31:
With the new bidirectional fundep, what would the instances look like if you wrote them out? Does the coverage condition hold in both directions?

paf31:
What I'm getting at is that a fundep doesn't just depend on the relation you're defining, but also how you define it, right? Like, numeric addition as a 3-way relation has the property that if you know two, you know the third, but in practice we define it by recursion on one or the other argument, so the coverage condition doesn't actually hold for one of the fundeps, I think.

LiamGoodacre:
Ah yeah, because RowList doesn't have the ordering invariant it means multiple RowLists can compute the same rows.  So the bidirectional fundep doesn't make sense.  Thanks!
I'll switch back to two classes when I get some time.

paf31:
I wasn't thinking about ordering so much, but that's true too, unless you require sorted lists. I was thinking more that you have to define the mapping by inspecting one argument or or other.

LiamGoodacre:
Rebased on master.  If I add a test now, it will either depend on my typelevel-prelude clone, or it will fail until we decide where to put the class.

paf31:
Thanks, looks great!

paf31:
A little test I did in the REPL, for reference:

```
> data LProxy (list :: RowList) = LProxy
> :p
… f :: forall proxy row list. RowToList row list => proxy row -> LProxy list
… f _ = LProxy

> :t f { x: 1, y : 2}
LProxy (Cons "x" Int (Cons "y" Int Nil))

> -- testing open rows
> :t \o -> o.x (f o)
forall t4 t5 t8.            
  RowToList                 
    ( x :: LProxy t5 -> t8  
    | t4                    
    )                       
    t5                      
   => { x :: LProxy t5 -> t8
      | t4                  
      }                     
      -> t8
```

👍 Looks really great!

PR #2946, opened 2017-06-17
[purs ide] Adds declarations for Prim

### LINKED ISSUE #2925: [purs ide] Add declarations for the Prim module
Missing completions for Array or String make people think purs ide isn't working. Add hardcoded `IdeDeclarations` for these and automatically add them when a `load all` command is triggered.

kRITZCREEK:
On top of #2942, fixes #2925 

I'm not entirely happy with this yet, because I can't distinguish between has been loaded via `load all` or  `load selectively` when repopulating the cache, so `Prim` always ends up in the cache right now. Also having to special case `Prim` things in the import module is a little unfortunate but I guess I won't get around it.

For these completions to show up in the editor (if they filter by imported modules) the editors would always need to add a `Prim` to their moduleFilters right now. Again I'm not sure if I should special case `Prim` for the completion command and make a separate `query` command without any magic.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -785,6 +785,7 @@ library
           Language.PureScript.Ide.Imports
           Language.PureScript.Ide.Logging
           Language.PureScript.Ide.Matcher
+          Language.PureScript.Ide.Prim
           Language.PureScript.Ide.Pursuit
           Language.PureScript.Ide.Rebuild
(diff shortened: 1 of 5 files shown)


kRITZCREEK:
I've decided it's better to inject the Prim declarations for the completion and type command instead of actually adding the Prim module to the state. I'm happy with this now and it's ready to review.

PR #2951, opened 2017-06-25
Allow things to be hidden from Prim

### LINKED ISSUE #2499: Cannot hide things from Prim
The following code attempts to hide `Int` from `Prim` and define a new local `Int`:
```purescript
module Main where
import Prim hiding (Int)
type Int = String
m :: Int
m = "hi"
```

This fails with the following error:
```
Conflicting definitions are in scope for type Int from the following modules:

    Main
    Prim
```

garyb:
Resolves #2499

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -29,6 +29,7 @@ import Language.PureScript.TypeClassDictionaries
 import Language.PureScript.Comments
 import Language.PureScript.Environment
 import qualified Language.PureScript.Bundle as Bundle
+import qualified Language.PureScript.Constants as C
 
 import qualified Text.Parsec as P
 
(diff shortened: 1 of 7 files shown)


purescript.cabal
```diff
@@ -962,7 +964,7 @@ executable purs
 
 test-suite tests
     build-depends:
-          aeson >=1.0 && <1.2
+          aeson >=1.0 && <1.1
```
garyb: Hmm, looks like that PR that went in changed this instead of the package.yaml. Maybe this should be removed from and gitignored in the repo now, if we're relying on hpack?

garyb:
For the record, when digging into this I discovered the environment was being initialized for every module as if it had these imports

```
import Prim
import Prim as Prim
```

regardless of what the user was doing. Now it allows these imports to be rewritten if they are specifically included. It's not great that it needs importing with and without qualification, but I have an idea about how to fix that as part of the fix for #1647 too. It's currenty necessary to have it qualified as generated names (like when {} is desugared into `Prim.Record`) still go through the name desugaring process, so the usual rules about the available scope apply. I'm thinking some generated names should/can be handled differently. This is a knock-on from the change a while back that only allows explicitly imported things to be resolved, even if fully qualified.

purescript.cabal
paf31: I think we need both if we want Cabal users to be able to build things.

hdgarrood: Yes, this has happened too many times now; I think we do need to either gitignore the cabal file or stop using hpack. I'm not particularly attached to one option over the other but if it was solely up to me I'd probably go for stopping using hpack.

paf31:
Can you explain a little more why we need both imports, maybe by giving an example? I'm not seeing it.

purescript.cabal
paf31: I'd be fine with not using hpack.

garyb:
``` purescript
type Test = {}
```
is desugared into

``` purescript
type Test = Prim.Record ()
```

With no `import Prim as Prim` it complains that `Prim.Record` is unknown.

paf31:
I assume this is to do with the order in which the various desugaring steps happen? Couldn't we just desugar it to `Record` (unqualified) then?

garyb:
> I assume this is to do with the order in which the various desugaring steps happen?

Well, kinda. Desugaring isn't the right term I guess, since it comes directly out of the parser as `Prim.Record`, so the other representation is lost before we get to the AST even.

> Couldn't we just desugar it to `Record` (unqualified) then?

We could, but that only works if someone hasn't done `import Prim hiding (Record)`, and/or defined `type Blah = Record`. I suppose it's true someone could still do `import Prim (String) as Prim` and still break it, but... whatever - like I said, I'll have another fix for that aspect soon 😛. This change is only an improvement on what we have, it's not adding something that didn't used to occur, it's just now it's done in one place rather than two.

PR #2953, opened 2017-06-25
Raise upper bound on aeson in package.yaml

garyb:
To match the cabal file. Assuming we're not dropping hpack... 😅


paf31:
👍 Thanks!

PR #2954, opened 2017-06-25
trim whitespace in typelevel string extraction

justinwoo:
Small cleanup PR for extra spaces in custom error constraints.

asdasdajsdlkzjxczxc

Left: stripped spaces, Right: original

For me, this is definitely what I want since I write all of my Fail constraints with `"""`, leading to having a newline and spaces before and after my message.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -1333,7 +1333,7 @@ renderBox = unlines
 
 toTypelevelString :: Type -> Maybe Box.Box
 toTypelevelString (TypeLevelString s) =
-  Just . Box.text $ decodeStringWithReplacement s
+  Just . line . T.strip $ decodeStringWithReplacement s
 toTypelevelString (TypeApp (TypeConstructor f) x)
   | f == primName "TypeString" = Just (typeAsBox x)
 toTypelevelString (TypeApp (TypeApp (TypeConstructor f) x) ret)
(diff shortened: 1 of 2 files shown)


paf31:
I'm not sure this is a good default. It works for things like this, but what if I want spaces for some reason? When using `TypeConcat` I can see the current behavior being useful.

Maybe you could put your error message in a type synonym if the lines are getting too long?

src/Language/PureScript/Errors.hs
```diff
@@ -1333,7 +1333,7 @@ renderBox = unlines
 
 toTypelevelString :: Type -> Maybe Box.Box
 toTypelevelString (TypeLevelString s) =
-  Just . Box.text $ decodeStringWithReplacement s
+  Just . line . T.strip $ decodeStringWithReplacement s
```
paf31: This isn't going to handle errors which span multiple lines very well. If we're going to strip spaces at all, we should strip the same number from each line.

hdgarrood:
I agree with @paf31 on this one. In your case, triple-quoted strings don't necessarily have to start with a newline and spaces, surely? could you write something like this instead?

```purescript
instance blah :: Fail """first line of the error message
  second line of the error message, indented""" => MyClass MyType
```

paf31:
I'm going to close this, but please open an issue and we can discuss options, and maybe reopen this. Thanks!

PR #2955, opened 2017-06-26
Allow `import M (C)` for importing class C

garyb:
edit: I'll add a few tests before calling this done.

### DIFF
--- a/src/Language/PureScript/Linter/Imports.hs
+++ b/src/Language/PureScript/Linter/Imports.hs
@@ -13,7 +13,7 @@ import Control.Monad.Writer.Class
 import Data.Function (on)
 import Data.Foldable (for_)
 import Data.List (find, intersect, groupBy, sortBy, (\\))
-import Data.Maybe (mapMaybe)
+import Data.Maybe (mapMaybe, maybeToList)
 import Data.Monoid (Sum(..))
 import Data.Traversable (forM)
 import qualified Data.Text as T
@@ -230,7 +230,7 @@ lintImportDecl env mni qualifierName names ss declType allowImplicit =
   checkExplicit declrefs = do
     let idents = ordNub (mapMaybe runDeclRef declrefs)
         dctors = mapMaybe (getDctorName <=< disqualifyFor qualifierName) names
-        usedNames = mapMaybe (matchName (typeForDCtor mni) <=< disqualifyFor qualifierName) names
+        usedNames = matchName (typeForDCtor mni) =<< mapMaybe (disqualifyFor qualifierName) names
         diff = idents \\ usedNames
 
     didWarn <- case (length diff, length idents) of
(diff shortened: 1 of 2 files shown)


kRITZCREEK:
I think you'll also need to update `prettyPrintImport`: https://github.com/purescript/purescript/blob/f1128ecb6f01bb1809adc9787df2958d9fefc096/src/Language/PureScript/Errors.hs#L1225-L1226

garyb:
Yeah, there are a couple of issues actually. You can't explicitly export without using the `class` syntax too, for example.

paf31:
I have an idea for an alternative to constraint kinds that I'd like to discuss before merging this. I need to prototype it first, but it wouldn't necessitate joining the two namespaces.

hdgarrood:
Aren't the namespaces joined already?

```purescript
module Main where
data C
class C
```
"Error: Declaration for type class C conflicts with an existing type of the same name."

paf31:
Hmm. I guess that's because of the generated type synonym for the class, but they could reasonably be made separate if we wanted to.

paf31:
I'm hoping I'll be able to clean up that type synonym hack at some point anyway.

garyb:
They're inconsistently joined - they are supposed to be fully, but there are cases that are not caught.

The ref-using-`class` is just a thing used to syntax-direct the lookup, it's not really necessary if we make the behaviour actually consistent :wink:

paf31:
So could we make them completely separate namespaces then?

kRITZCREEK:
That's going to be bad for editors afaict. It's going to be tricky to tell classes from types for go-to-definition, show kind, etc. 

And I think beginners will have the same problem.

paf31:
But classes don't appear in the same places as types, syntactically speaking.

paf31:
So what's the plan here then? Are the two namespaces going to be made separate, or joined together?

Note that constraint kinds don't need to influence the answer, if we go with the constraint proxy approach.

garyb:
I don't really have a preference as to whether they're joined or not. :)

There are some odd rules around namespace collisions already, as you can't name a data constructor and class the same thing within a module, since the generated code would overlap unless we auto-rename one or other of the definitions.

paf31:
I've been thinking about this in the context of constraint kinds, and I think it might be nice to keep them separate, but fix up those weird corner cases where they overlap, if possible.

The reason is that I think the errors will be nicer if the namespaces are separate. And constraint kinds/proxies/whatever will be a relatively little-used feature anyway.

kRITZCREEK:
I'm not saying it's ambiguous, I'm saying it's hard to tell apart.

```haskell
myFn
  ∷ ∀ f m
  . Functor f
  ⇒ Applicative f
  ⇒ Parralel f m
  → Monad m
  → Int

myFn
  ∷ ∀ f m
  . Functor f
  ⇒ Applicative f
  ⇒ Parralel f m
  ⇒ Monad m
  → Int
```

Also I think it should be a valid question to ask psci for the kind of `Monad` even though it's a class. That will lead to collisions if we separate the namespaces.

garyb:
> I've been thinking about this in the context of constraint kinds, and I think it might be nice to keep them separate, but fix up those weird corner cases where they overlap, if possible.
> 
> The reason is that I think the errors will be nicer if the namespaces are separate. And constraint kinds/proxies/whatever will be a relatively little-used feature anyway.

Sounds good to me 👍 

I'll close this in that case?

paf31:
> Also I think it should be a valid question to ask psci for the kind of Monad even though it's a class. 

Well, with the constraint proxies approach, you would be able to say `:kind class Monad`.

> I'll close this in that case?

Let's leave it for now, so other people can weigh in. I would like to wait until we have a prototype constraint proxies branch, so that we at least know that's an option.

hdgarrood:
@kritzcreek bit tangential but regarding that example: this is why I prefer putting the arrows on the right:

```purescript
myFn :: forall m a.
  Functor m =>
  Applicative m =>
  Parallel m =>
  m a ->
  m a
```

I don't have any strong feelings re whether classes are in a separate namespace. If it would improve errors I'm inclined to say yes to separating them.

Issue #2872, opened 2017-05-01
Documentation of type does not list instances if instances were defined using operator alias for type


rightfold:
The instance is defined as such:

```instance eqQuotient :: (Eq a, Canonical a e) => Eq (a / e) where```

Instead of such:

```instance eqQuotient :: (Eq a, Canonical a e) => Eq (Quotient a e) where```

Because of this, the documentation does not list the instances under the `Quotient` type. Example: https://pursuit.purescript.org/packages/purescript-quotient/0.0.3/docs/Type.Quotient#t:Quotient

paf31:
Sounds like we need to run the operator desugaring as part of the docs generation process.

seanwestfall:
I'll take this one on as a new contributor, to be a new contributor.

hdgarrood:
I've removed the new contributor label, as this turns out to be not that simple after all. Also I'm looking at it :)

seanwestfall:
@hdgarrood You're right, it's not that easy. If you want, I can run you through what I've found so far. I don't even think the fix exists in this repo, but is more along the lines of being somewhere in here https://github.com/purescript-contrib/pulp with the `docs` command -- but even when you get that far, you'll find that what actually gets on to the pursuit page is something else, that's built into json somewhere I don't know, that maybe only @paf31 knows...

hdgarrood:
It's fixed in this repo already: #2991 :smile:

hdgarrood:
The majority of the HTML generation code is now in this repo, by the way, to support `purs docs --format html`; there's just a small amount of additional stuff inside Pursuit to take care of generating the right links between different packages and so on.

PR #2960, opened 2017-06-29
Desugar Operators during Doc Generation

mankyKitty:
#2872 

Type operators are now desugared during the documentation generation
phase. This leads to some pretty gnarly type signatures (see the
changes in the tests), but does also clear up TypeClass declarations
to use the unaliased type constructor and not the operator.

Some of my ImportSpec tests are failing for an IOException that I have not
yet been able to track down.

Any any all feedback appreciated. :)

### DIFF
--- a/src/Language/PureScript/Sugar/Operators.hs
+++ b/src/Language/PureScript/Sugar/Operators.hs
@@ -7,6 +7,7 @@
 --
 module Language.PureScript.Sugar.Operators
   ( desugarSignedLiterals
+    , desugarTypeClassTypeOpsForModule
   , rebracket
   , checkFixityExports
   ) where
@@ -57,6 +58,9 @@ type FixityRecord op alias = (Qualified op, SourceSpan, Fixity, Qualified alias)
 type ValueFixityRecord = FixityRecord (OpName 'ValueOpName) (Either Ident (ProperName 'ConstructorName))
 type TypeFixityRecord = FixityRecord (OpName 'TypeOpName) (ProperName 'TypeName)
 
+makeLookupEntry :: FixityRecord op alias -> (Qualified op, Qualified alias)
+makeLookupEntry (qname, _, _, alias) = (qname, alias)
+
 -- |
 -- Remove explicit parentheses and reorder binary operator applications.
 --
@@ -72,7 +76,7 @@ rebracket externs modules = do
   let (valueFixities, typeFixities) =
         partitionEithers
           $ concatMap externsFixities externs
-          ++ concatMap collectFixities modules
+          ++ concatMap collectFixitiesFromModule modules
 
   ensureNoDuplicates' MultipleValueOpFixities valueFixities
   ensureNoDuplicates' MultipleTypeOpFixities typeFixities
@@ -100,9 +104,6 @@ rebracket externs modules = do
     :: [FixityRecord op alias]
     -> [[(Qualified op, Associativity)]]
   customOperatorTable' = customOperatorTable . map (\(i, _, f, _) -> (i, f))
-
-  makeLookupEntry :: FixityRecord op alias -> (Qualified op, Qualified alias)
-  makeLookupEntry (qname, _, _, alias) = (qname, alias)
 
   renameAliasedOperators
     :: M.Map (Qualified (OpName 'ValueOpName)) (Qualified (Either Ident (ProperName 'ConstructorName)))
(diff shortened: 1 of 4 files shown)


kRITZCREEK:
Which of them in particular? Could you paste the output here, because Traviss succeeds?

hdgarrood:
Can you show me what the html docs look like now for that module whose tests you had to change? I think it would really be better to avoid changing how that renders. Also could you please add a test which checks that #2872 is addressed?

mankyKitty:
@kRITZCREEK Here is a snippet of the errors I've received. It'll be a pebkac issue RE my setup, just haven't found it yet.

Failures:  tests/Language/PureScript/Ide/ImportsSpec.hs:264: 
1) Language.PureScript.Ide.Imports, importing from a loaded IdeState, adds an explicit unqualified import
uncaught exception: IOException of type UserError (user error (Pattern match failure in do expression at tests/Language/PureScript/Ide/Imp8))

@hdgarrood I understood that changing how the docs render was, in part, the purpose of the request? But perhaps not as much as I've done... This should show what the docs look like now http://jsbin.com/pocivizuyi . Let me know if that link doesn't work.

I'll start on the test, sorry about that.

hdgarrood:
No, no need to apologise at all, thanks for picking this up!

I think it's probably better to use the underlying type name in instances, like in the Quotient example given originally, since instances will appear under the type rather than the operator alias. However in a value declaration I think it is preferable to keep as close to what's written in the source file as possible, ie in this context to use type operators where the code does. I appreciate that it might not be easy to see how to do this, though. Would it be possible to perform the type operator desugaring step inside Docs.Convert.Single so that we can make sure it only gets applied to instance declarations, or alternatively modify the pass that collects instances for a given type so that it can handle this?

mankyKitty:
This is my first contribution to PureScript the compiler so any of those things are possible! I just don't know how to do it yet. I'll have another go. 👍 

However I'm off for a holiday in the next couple of days so if I can't get it sorted I might just have to close my PR as I won't be in front of a computer for about a month. :(

mankyKitty:
I've added "selective type instance declaration with type operator desugaring" to the partial desugaring that happens during Doc generation and it's producing the more finer grained desugaring that I think we're after.

I'll try to make a test for it and then push it up.

src/Language/PureScript/Sugar/Operators.hs
```diff
@@ -249,9 +250,38 @@ externsFixities ExternsFile{..} =
       , Fixity assoc prec
       , name
       )
+desugarTypeClassTypeOpsForModule
```
hdgarrood: Would it make more sense to call this `desugarInstanceTypeOpsForModule`? Also all the other functions in this module seem to use the full word "operators" rather than abbreviating it, so I think it would be best to stick to that here too.

hdgarrood: Also, it's not clear to me that we gain much from separating this into two functions — have you considered just using one function? Presumably we wouldn't need to change `collectFixities` if we only used one function?

hdgarrood: Finally could you please add a comment explaining why this exists?

hdgarrood:
Thanks for starting on this! I'm going to close this for now and try to get it to a point where it's mergeable myself.

mankyKitty:
That's cool, sorry I went radio silent, out of the country on holidays. I'm happy to pick this back up when I get back in early August, understand if you'd rather just knock it over. To answer your questions though... 

I didn't think to use just Tuple and Show, I was just trying to be a bit descriptive in case people had issues tracing the types through. In hindsight, creating weird new types probably wouldn't help. 

The change to the testing code was because as far as I could establish, that particular test always passed regardless of the actual result. I have a new test that checks that child instance docs appear correctly. But this test was passing when it should have failed. Making this change caused the test to fail correctly. Other tests using this code didn't need to be updated as they met expectations. 

As for the whitespacing and naming, that's a fat fingering on my part, and I hate naming things so I have no particular attachment, happy to change. :p

For the extra function, the two entry points that run this code both receive a different Module type. I didn't have quite enough info to create the alternate type, and it seemed silly to when collectFixities only uses two attributes so I'd be doing more work for nothing. So it seemed simpler to me to have one function only taking what it needs and another that was a drop in replacement for the existing one. 

Also the module type in question isn't a record type and I was sure about the etiquette for adding Prisms etc to existing data types. :) 

Soz for brain dump on a closed PR, but I wanted to let you know what had happened. :)

hdgarrood:
Yeah no worries at all - you did say you were leaving as well which is why I decided to pick this up. :smile: 

After looking a bit I realised that this second implementation wouldn't quite work either as it wasn't taking account of operator precedence, so e.g. if you had an instance like `instance c :: C (Int × Int ⊕ String)` where `×` has higher precedence than `⊕`, it should be treated as `(⊕) ((×) Int Int) String`, but it was coming out in the docs as if it were `(×) ((⊕) Int String) Int`. What I ended up doing was providing a new version of the `rebracket` function which lets you choose which declarations it will desugar; see #2991 if you're interested.

PR #2964, opened 2017-07-03
Allow operators shadowed by variables of the same name as the alias

### LINKED ISSUE #2803: Can't use an operator shadowed by a variable of the same name as the aliasee
```purescript
f :: Int -> Int -> Int
f = (+)

infixl 6 f as %

g :: Int -> Int -> Int
g a b = let f = (-) in a % b
```

Should be functionally equivalent to:

```purescript
f :: Int -> Int -> Int
f = (+)

infixl 6 f as %

g :: Int -> Int -> Int
g = f
```

Instead, I get the error:

<pre>  Value f is undefined.

while inferring the type of f
in value declaration g</pre>

LightAndLight:
Fixes #2803 
Fix transcript: https://gist.github.com/LightAndLight/67168114153cb3519c50185f3149b56c

The fix feels a bit heavy-handed, I didn't see any other way around it. I'll sleep on it.

Open to suggestions.

### DIFF
--- a/examples/passing/2803.purs
+++ b/examples/passing/2803.purs
@@ -1,0 +1,17 @@
+module Main where
+
+import Control.Monad.Eff.Console (log)
+import Prelude ((+), (-), (==))
+
+f :: Int -> Int -> Int
+f = (+)
+
+infixl 6 f as %
+
+g :: Int -> Int -> Int
+g a b = let f = (-) in a % b
+
+main =
+  if g 10 5 == 15
+    then log "Done"
+    else log "Failed"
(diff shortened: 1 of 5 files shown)


paf31:
@garyb Do you think this is good to merge?

garyb:
Intriguing bug! The fix seems pretty reasonable to me. Perhaps another alternative that would work is to use `Either Ident Ident` instead where one side is local defs and the other is top-level. That would remove some possibilities, since the only qualification we're interested in is that of the current module anyway?

kRITZCREEK:
Could we circumvent this problem by resolving names before desugaring operators?

kRITZCREEK:
Fixed in #3320

Issue #2947, opened 2017-06-17
Instances with just one method do not require that the method be indented


rightfold:
That is, the following compiles:

    data Unit = Unit

    instance eqUnit :: Eq Unit where
    eq _ _ = true

quesebifurcan:
New contributor here -- I'd be happy to give this one a try.

kpadmasola:
@paf31 , New contributor here -- I'd be happy to give this one a try. What more is needed here?

paf31:
@kpadmasola There was a previous PR for this, which I need to include in the changes for 0.12.0.

https://github.com/purescript/purescript/pull/2965

Thank you though, and if you're interested in finding something else to work on, maybe we can help on Slack.

PR #2965, opened 2017-07-03
Require single-method instance declarations to be indented

quesebifurcan:
Resolves https://github.com/purescript/purescript/issues/2947

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -82,6 +82,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@phiggins](https://github.com/phiggins) | Pete Higgins | [MIT license](http://opensource.org/licenses/MIT) |
 | [@philopon](https://github.com/philopon) | Hirotomo Moriwaki | [MIT license](http://opensource.org/licenses/MIT) |
 | [@pseudonom](https://github.com/pseudonom) | Eric Easley | [MIT license](http://opensource.org/licenses/MIT) |
+| [@quesebifurcan](https://github.com/quesebifurcan) | Fredrik Wallberg | [MIT license](http://opensource.org/licenses/MIT) |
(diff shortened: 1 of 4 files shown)


quesebifurcan:
Thanks for the swift review. Looks like `purescript.cabal` was added to `.gitignore` recently: https://github.com/purescript/purescript/commit/947c950a58a9949015bdc63f41246578e9696118

LiamGoodacre:
Oh interesting, thanks!

hdgarrood:
I don't think this previously would have required a cabal file change anyway, it used to be only when you added a directory that you would need to change it, since the globbing in cabal files doesn't allow any kind of recursive includes like `**/*.purs`. Now that we generate the cabal file using hpack it should be even more rare that adding test files needs a corresponding change in package.yaml. Anyway yeah this looks good to me :+1:

paf31:
Thanks! Please update CONTRIBUTORS.md as well and I'll get this merged in.

paf31:
Thanks!

paf31:
Unfortunately, this is a breaking change, since the compiler no longer allows code like this, which occurs quite frequently:

```purescript
instance foo :: Foo String where

instance bar :: Bar String where
  member = ...
```

Note how `foo` has no members but still uses `where`.

I'll revert this, but we can try to include it in the next breaking release or find a way to make it non-breaking.

paf31:
@quesebifurcan Would you like to reopen this now that we're putting 0.12.0 together? Thanks!

quesebifurcan:
@paf31 Sure, I'll see if I can come up with something (preferably non-breaking) over the weekend. If not, I'm happy to let someone else take over.

paf31:
Oh it's fine to break now, we're at a major release. You could just reopen the PR if you like.

PR #2967, opened 2017-07-04
Fixed type shadowing error

### LINKED ISSUE #2197: Type shadowing error
The following mechanism used to compile and work before but does not compile with v0.9.1:

```
module Shadow where

import Prim as P

type Number = P.Number
z :: Number
z = 0.0
```

Error:

```
Error found:
in module Shadow
at ..../Shadow.purs line 6, column 1 - line 6, column 12

  Conflicting definitions are in scope for type Number from the following modules:

    Prim
    Shadow
```

LightAndLight:
Fixes #2197 

`Prim` was being imported unqualified to every single module. I changed `addDefaultImport` to only add an unqualified import if the module-to-import has not already been imported qualified. I think this is in line with the behaviour that the person who raised the issue expected.


LightAndLight:
Note for later:

 This hasn't accounted for a case like the following:

```
module Main where

import Prim (Number)
...
```

I think in this case *only* number should be imported from `Prim`. Meaning: a default module should only be imported unqualified if it is never imported, qualified or unqualified.

garyb:
Can you add a test to see what happens when you add `type Blah = {}` with a qualified import of `Prim`? Pretty sure it won't compile, we can't make this change until I do something else with the names that I have as a WIP. See #2951.

LightAndLight:
I will get back to you in a few hours, but if `{}` is represented as `Prim.Record` then it should compile fine. The behaviour of `addDefaultImport` for qualified imports hasn't changed.

paf31:
@garyb Could you please review this again?

garyb:
Sorry about the delay on this, I'll definitely take a look this weekend. I'm sure it is fine, I just want to better understand for myself what's going on too :smile:

PR #2969, opened 2017-07-06
Instance signatures don't bring type variables into scope

### LINKED ISSUE #2941: Instance signatures don't bring type variables into scope

paf31:
Fixes #2941

This uses a `let`, as discussed before.

### DIFF
--- a/examples/passing/ScopedTypesInInstance.purs
+++ b/examples/passing/ScopedTypesInInstance.purs
@@ -1,0 +1,15 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+
+class Test t where
+  test :: forall a. t -> a -> a
+
+instance testUnit :: Test Unit where
+  test :: forall a. Unit -> a -> a
+  test = go where
+    go :: Unit -> a -> a
+    go _ a = a
+
+main = log "Done"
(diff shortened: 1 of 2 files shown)


matthewleon:
@paf31 one quick thought that came to mind: as I recall, the source file you're modifying has an extensive comment section that gives examples of the desugaring transformations. A quick look at this PR shows that there were no modifications there. Do these comments require a change?

paf31:
@matthewleon Fixed, thanks!

I'm going to quickly test this against the current package set, and then I'll merge it.

paf31:
Hmm, this seems to have a problem with `profunctor-lenses`:

```
Error found:
in module Data.Lens.Internal.Wander
at .psc-package/psc-0.11.5/profunctor-lenses/v3.2.0/src/Data/Lens/Internal/Wander.purs line 25, column 1 - line 25, column 56

    Applicative t5 => (a0 -> t5 b1) -> s2 -> t5 t3
                                                  
  with type
                                                      
    Applicative f06 => (t7 -> f06 t8) -> t9 -> f06 t10
```

garyb:
```
```
Everyone's favourite error! 😆 Well, maybe second place after skolem escape.

paf31:
This is actually proving to be quite tricky. I'll leave it out for 0.11.6.

paf31:
Closing since this is incomplete.

Issue #2972, opened 2017-07-06
Type alias strings don't seem to work when used as typeclass constraints


justinwoo:
I want to make a type alias string, but it doesn't seem to work when I want to use it as a typeclass constraint. see:

```purs
module Main where

import Prelude

type E = "what"

class TestThing f where
  method :: f -> f

instance ttInt :: Fail E => TestThing Int where --  No type class instance was found for Prim.Fail E
-- instance ttInt :: Fail "inline works" => TestThing Int where -- this works
  method a =  a

foo :: Int
foo = method 0 -- fails with the above error
```

LiamGoodacre:
I think I have a fix for this.  Just testing it.

PR #2973, opened 2017-07-06
Replace synonyms in instance constraints

LiamGoodacre:
For #2972


src/Language/PureScript/TypeChecker.hs
```diff
@@ -323,7 +323,8 @@ typeCheckAll moduleName _ = traverse go
           sequence_ (zipWith (checkTypeClassInstance typeClass) [0..] tys)
           checkOrphanInstance dictName className typeClass tys
           _ <- traverseTypeInstanceBody checkInstanceMembers body
-          let dict = TypeClassDictionaryInScope (Qualified (Just moduleName) dictName) [] className tys (Just deps)
+          deps' <- traverse (overConstraintArgs $ traverse replaceAllTypeSynonyms) deps
```
paf31: I like that these compose like lenses

```purescript
(traverse . overConstraintArgs . traverse) replaceAllTypeSynonyms deps
```

😃

PR #2976, opened 2017-07-07
[purs ide] Collect data constructors

### LINKED ISSUE #2797: [purs ide] Collect data constructors for types
Right now we collect data constructors and the type they construct, but we discard the mapping in the opposite direction `Type -> [DataConstructor]`. This information is already present in the ExternsFile format, it just needs to be added as a field to `IdeType` and populated when converting Externs.

kRITZCREEK:
The information still needs to be made queryable, but I have an idea for how to use the recently added DeclarationFilters to do so.

Fixes #2797


kRITZCREEK:
Thanks!

PR #2981, opened 2017-07-09
Include comments in child declarations in HTML docs

hdgarrood:
That is, if doc-comments are provided in a source file for any of the
following:

- type class members
- instance declarations
- data constructors

they will now be included in the HTML.

Fixes https://github.com/purescript/pursuit/issues/159.

### DIFF
--- a/app/static/pursuit.css
+++ b/app/static/pursuit.css
@@ -413,6 +413,10 @@ ol li {
 .decl__body .syntax {
   color: #0B71B4;
 }
+.decl__child_comments {
+  margin-top: 1rem;
+  margin-bottom: 1rem;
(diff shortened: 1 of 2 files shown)


hdgarrood:
Oh of course, yes: http://harry.garrood.me/purs-html-docs-example/Data.Either.html

I was previously thinking of making these collapsible (see the Pursuit issue), but since library authors can choose whether to use these or not, I don't think there's too much risk of docs becoming excessively verbose; I think we can and should trust authors to use this appropriately.

paf31:
Looks really good!

hdgarrood:
Thanks :)

PR #2983, opened 2017-07-09
Improve instance ordering in HTML docs

hdgarrood:
Instances now appear in the same order as in the source file, which is
in general likely to be the most sensible option.

Fixes https://github.com/purescript/pursuit/issues/125


hdgarrood:
Oops, this isn't quite right. I'll send another PR shortly.

PR #2985, opened 2017-07-10
Revert "Require single-method instance declarations to be indented"

paf31:
Reverts purescript/purescript#2965


paf31:
Tested locally

hdgarrood:
Did you mean to revert all of these?

paf31:
It's just how GitHub's UI does the revert unfortunately. If you see, they're all empty commits actually, and after squashing, it's just one with the commit I need removed.

hdgarrood:
Ahh, I see. Thanks for clarifying!

PR #2989, opened 2017-07-10
Fix fragment links to type constructors in HTML

### LINKED ISSUE #2988: Fragment links to type constructors in HTML are broken
Everything is being treated as a type operator, e.g. if you declare `foo :: Int`, the HTML docs will render a link for `Int` in that declaration as `[...]/Prim.html#t:type (Int)` rather than `Prim.html#t:Int`.

This must be a result of #2979. I'm looking at this now.

hdgarrood:
Fixes #2988


hdgarrood:
Thanks for reviewing :)

Also, the fact that this happened suggests we should consider running tests on the html that `purs docs` produces, maybe... but it might be hard to do well so I'm not super inclined to work on that right now.

PR #2997, opened 2017-07-14
Fix source links for re-exports

felixSchl:
Fixes purescript/pursuit#304

Refer to purescript/pursuit#319


hdgarrood:
Thanks for looking at this! I'm not quite sure about what approach is the best though, I need to have a think.

hdgarrood:
Ok, this looks good - if you could just fix the warning which is causing CI to fail then I think this is good to merge. Thanks!

hdgarrood:
Does anyone else want to look at this or should I just go ahead and merge?

paf31:
@hdgarrood I haven't had a chance to review yet, but please merge if you're happy (in general).

hdgarrood:
Ok :)

@felixschl thanks very much!

PR #2998, opened 2017-07-15
Provide right git url in purs publish error message

coot:
cc @hdgarrood

### DIFF
--- a/src/Language/PureScript/Publish.hs
+++ b/src/Language/PureScript/Publish.hs
@@ -215,7 +215,7 @@ getManifestRepositoryInfo pkgMeta =
 
   where
     format :: (D.GithubUser, D.GithubRepo) -> Text
-    format (user, repo) = "https://github.com/" <> (D.runGithubUser user) <> "/" <> (D.runGithubRepo repo) <> ".git"
+    format (user, repo) = "https://github.com/" <> D.runGithubUser user <> "/" <> D.runGithubRepo repo <> ".git"
 
 checkLicense :: PackageMeta -> PrepareM ()
 checkLicense pkgMeta =
--- a/src/Language/PureScript/Publish.hs
+++ b/src/Language/PureScript/Publish.hs
@@ -202,15 +202,16 @@ getTagTime tag = do
     _ -> internalError (CouldntParseGitTagDate tag)
 
 getManifestRepositoryInfo :: PackageMeta -> PrepareM (D.GithubUser, D.GithubRepo)
-getManifestRepositoryInfo = either (userError . BadRepositoryField) return . tryExtract
-  where
-  tryExtract pkgMeta =
-    case bowerRepository pkgMeta of
-      Nothing -> Left RepositoryFieldMissing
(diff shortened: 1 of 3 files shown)


hdgarrood:
In many cases the result of `git config remote.origin.url` is going to be an SSH url, and I don't think we should suggest that, because then I think people can only clone it if they have a public key which GitHub (or whatever host) recognises. Also, many networks block SSH and the Git protocol, but HTTPS works pretty much everywhere. So I think we should only ever suggest HTTPS urls. We could automatically rewrite GitHub SSH urls to GitHub HTTPS urls to avoid this, though.

Also, the main purpose of that hint is so that people don't have to go and look up how to put this information into bower.json, so I think that hint should be displayed every time, even if we do have to use a default in some cases.

How about this:

- if it's an HTTPS url, suggest it as-is
- if it's an SSH or Git url and hosted on GitHub, rewrite it to an HTTPS url and suggest that
- if it's any other kind of url, ignore it and display the hint with a default url: `https://github.com/myuser/myrepo.git` - it should then be more obvious than it is currently that you should change it.

coot:
I'd just capitalize: `https://github.com/USER/REPO.git` so it's easy to spot.

hdgarrood:
Good idea :+1:

coot:
@hdgarrood it's ready for review.

src/Language/PureScript/Publish/ErrorsWarnings.hs
```diff
@@ -234,19 +234,18 @@ spdxExamples =
...
 displayRepositoryError :: RepositoryFieldError -> Box
 displayRepositoryError err = case err of
-  RepositoryFieldMissing ->
+  RepositoryFieldMissing giturl ->
     vcat
       [ para (concat
          [ "The 'repository' field is not present in your package manifest file. "
          , "Without this information, Pursuit would not be able to generate "
-         , "source links in your package's documentation. Please add one - like "
-         , "this, for example:"
```
hdgarrood: I think I prefer the original text here, actually. We don't want to be too prescriptive: people might have a reason for having `remote.origin.url` differ from what's inside bower.json.

src/Language/PureScript/Publish.hs
```diff
@@ -202,15 +202,20 @@ getTagTime tag = do
...
   where
-  tryExtract pkgMeta =
-    case bowerRepository pkgMeta of
-      Nothing -> Left RepositoryFieldMissing
-      Just Repository{..} -> do
-        unless (repositoryType == "git")
-          (Left (BadRepositoryType repositoryType))
-        maybe (Left NotOnGithub) Right (extractGithub repositoryUrl)
+    format :: (D.GithubUser, D.GithubRepo) -> Text
+    format (user, repo) = "https://github.com/" <> (D.runGithubUser user) <> "/" <> (D.runGithubRepo repo) <> ".git"
```
hdgarrood: Sorry to nitpick but could you remove the redundant parens here, please?

PR #3002, opened 2017-07-24
Prevent browser from treating declTitle and linkToSource as one word

Rufflewind:
The lack of space between the identifier and the "Source" link means that if anyone tries to double-click-select the identifier, they will end up selecting the "Source" too. (Curiously, Haddock once had the [same problem](https://github.com/haskell/haddock/issues/224).)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -85,6 +85,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@rightfold](https://github.com/rightfold) | rightfold | [MIT license](https://opensource.org/licenses/MIT) |
 | [@robdaemon](https://github.com/robdaemon) | Robert Roland | [MIT license](http://opensource.org/licenses/MIT) |
 | [@RossMeikleham](https://github.com/RossMeikleham) | Ross Meikleham | [MIT license](http://opensource.org/licenses/MIT) |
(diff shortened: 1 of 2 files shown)


hdgarrood:
Thanks for the PR :) can you let me know which browser you saw this in, and was it on Pursuit or with locally generated html docs?

Rufflewind:
It was observed on Chromium Linux in both Pursuit and locally generated docs.  Adding the space fixed the problem locally.

michaelficarra:
I was able to reproduce on Chrome/Linux. :+1: This fix should work. Also, Haddock fixed it by disallowing selection of the "Source" link: https://github.com/haskell/haddock/pull/655. I don't think that's a good way to solve this issue.

hdgarrood:
Ok great, thanks, looks good! Would you mind updating CONTRIBUTORS.md?

Rufflewind:
Done.  Thanks for the quick review!

ad-si:
Can we please reopen this ticket?
This is still a problem in Chrome and derivatives like Brave! 🙈

rhendric:
The code added in this PR hasn't changed. Seems like a bug in Blink's double-click behavior: https://issues.chromium.org/issues/40711666

If you have an idea for a fix (that doesn't entirely disallow selection of the Source link), you can submit a new PR.

ad-si:
Wrapping the declTitle in a div seems to be enough to fix the problem.

PR #3005, opened 2017-07-29
Fix a loop in the kind checker

### LINKED ISSUE #2995: Compiler spins forever on data Tree m a = Tree a (m (Tree a))
I meant to write `data Tree m a = Tree a (m (Tree m a))` but I wrote `data Tree m a = Tree a (m (Tree a))`.

This seems to make the compiler spin forever.

paf31:
Fixes #2995


hdgarrood:
Would it make sense to add a test so that this doesn't happen again?

paf31:
Yes, definitely.

PR #3006, opened 2017-07-30
[purs ide] Editor mode

kRITZCREEK:
Adds a new flag to the server invocation: `--editor-mode`

When the editor-mode flag is specified at startup the server will not start a file watcher process anymore. Instead it only reloads after successful rebuild commands. This is a lot less fragile than relying on the file system APIs, but will mean that a manual load needs to be triggered after builds that didn't go through purs ide.

Looking forward to get Feedback :) After using psc-ide more intensively the last months it seems like the file watcher is the most fragile part of the system so I'm looking for ways to circumvent that problem.

### DIFF
--- a/app/Command/Ide.hs
+++ b/app/Command/Ide.hs
@@ -69,6 +69,7 @@ data ServerOptions = ServerOptions
   , _serverNoWatch    :: Bool
   , _serverPolling    :: Bool
   , _serverLoglevel   :: IdeLogLevel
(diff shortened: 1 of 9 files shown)


paf31:
Is this different from `noWatch` then?

kRITZCREEK:
Yeah, since it loads the parsed file and generated ExternsFile after rebuilding into the cache.

Issue #2911, opened 2017-05-26
Generic.Rep doesn't derive inner records


pkamenarsky:
Is this expected behaviour?

     data A = A { b :: { c :: Array String } }

The `Generic.Rep` representation is:

    (Constructor "A"
       (Rec
          (Field "b"
             { c :: Array String
             }
          )
       )
    )

I would have expected the inner record to result in `Rec (Field "c" ...` and so on.

garyb:
I guess this would also fix `genericShow` working for a nested record?

paf31:
I think so, yes.

coot:
I started working on this issue [here](https://github.com/coot/purescript/tree/generic-rep-for-records).

What I have is this:
```
data A = A { a :: { b :: String, c :: String } }

derive instance genericRepA :: Generic A _
```

The generic rep for A has type:
```
Constructor "A" (Rec (Field "a" (Rec (Product (Field "b" String) (Field "c" String)))))
```

But `genericShow a` fails with
```
> genericShow a
Error found:
in module $PSCI
at  line 1, column 1 - line 1, column 13

  No type class instance was found for
                                                                        
    Data.Show.Show (Rec (Product (Field "b" String) (Field "c" String)))
                                                                        

while applying a function genericShow
  of type Generic t0 t1 => GenericShow t1 => t0 -> String
  to argument a
while inferring the type of genericShow a
in value declaration it

where t1 is an unknown type
      t0 is an unknown type

See https://github.com/purescript/documentation/blob/master/errors/NoInstanceFound.md for more information,
or to contribute content related to this error.
```

I need to look into `Generic.Rep`, maybe it needs some changes.

paf31:
I think we need an instance for `GenericShowFields (Rec fields)`.

coot:
Adding:
```
instance genericShowFieldsRec :: GenericShowFields a => GenericShowFields (Rec a) where
  genericShowFields (Rec a) = ["{ " <> intercalate ", " (genericShowFields a) <> " }"]
```
is not enough, there is still missing a link from a `Rec a` that has an instance of `GenericShowFields` to `Field name (Rec a)`.  There is [genericShowFieldsField](https://github.com/purescript/purescript-generics-rep/blob/master/src/Data/Generic/Rep/Show.purs#L62) but it has a `Show a` constraint.  This is needed to show non nested fields.  I thought that adding a similar instance with a `GenericShow a` contstraint instead of `Show a` would make it work but it does not.

```
[...]

PR #3007, opened 2017-07-30
Remove old generics and generics-rep record support

paf31:
This removes

- the old `generic` deriving mechanism
- deriving for records in `generics-rep`

It needs some changes to `generics-rep`, and we'd probably want to deprecate `generics` too.

This is breaking, so it would go into 1.0 or 0.12.0.

Finally, this means we need to figure out how to derive `Eq`, `Show`, `Ord`, etc. for records using `RowToList`, which means moving some things around in the core libraries.


hdgarrood:
I was expecting that we wouldn't derive `Eq`, `Show`, `Ord` etc, but rather just define them as normal instances; is that not the case? What are the unresolved questions around instances such as `Show (Record a)`?

hdgarrood:
Also, this seems fairly significant, so I think it might be sensible to call the first release containing this v0.12.0, just in case we later notice some problem with it which we want to fix, but fixing it requires another breaking change.

paf31:
Yes, sorry I meant we should write instances for those.

And yes, I think we do need a 0.12.0 release separate from 1.0.0 at this point.

hdgarrood:
Is there anything stopping us from writing those instances now, and releasing updated core libraries with those instances, so that people can use them with the 0.11.x compiler?

paf31:
Yes, we'd need to move `RowToList` somewhere where the new instances in `Prelude` could use it.

paf31:
Fixes #2911.

paf31:
@hdgarrood Done, thanks! I'll merge once CI passes.

joneshf:
Would it be possible to add a `Fail` instance to generics/generics-rep describing this particular change and the resolution?

paf31:
How would it work? The compiler won't derive instances for records any more, so the issue is not that we don't want an instance to exist, but that we want other instances to exist (e.g. `Show (Record r)`) now.

joneshf:
I see. What sort of message do you get when you try to derive now?

paf31:
> No instance for `YourClass { ... }`

fsoikin:
What's the official way of migrating old `Data.Generic`-based code to `Data.Generic.Rep`? Is there, perhaps, a stopgap `instance New.Generic a r => Old.Generic a` implemented somewhere?

paf31:
You could try using [this code](https://github.com/purescript/purescript-generics-rep/pull/7). If it doesn't get merged, I'll probably close the PR and make a `generics-compat` repo.

fsoikin:
Got it, thanks @paf31.

One question, out of curiosity though: [a comment](https://github.com/purescript/purescript-generics-rep/pull/7/files#diff-60912c1fe64dd20d56036eab89061210R105) on that PR states that "_For standard data structures, you can simply write `derive instance deepFoo :: Deep Foo` in the module they are declared, and the instance methods will be filled in for you._" How does that work exactly? PureScript can't do `DeriveAnyClass`, can it?

paf31:
I may have copied that incorrectly from the original docs 😄

fsoikin:
Ok, got it.

PR #3011, opened 2017-07-30
Remove scripts

### LINKED ISSUE #2993: Deprecate & remove old wrapper scripts
This was brought up by https://github.com/purescript/purescript/pull/2952#issuecomment-314399620

I don't think the scripts `psc`, `psci` etc are providing any value any more, so I'd like to remove them. It's possible that some people are using them though, so I think we should wait until the next breaking release to remove them.

In the meantime I think we should start including the scripts in the Hackage source distribution, since people (eg, Homebrew) will be using that now instead of getting the source from GitHub, as the latter doesn't include a cabal file any more.

paf31:
Breaking, for 0.12.0, fixes #2993

### DIFF
--- a/bundle/build.sh
+++ b/bundle/build.sh
@@ -35,7 +35,6 @@ fi
 cp "$FULL_BIN" bundle/build/purescript
 
 # Copy extra files to the staging directory
-cp scripts/*             bundle/build/purescript/
 cp bundle/README         bundle/build/purescript/
 cp LICENSE               bundle/build/purescript/
 cp INSTALL.md            bundle/build/purescript/
--- a/scripts/psc
+++ b/scripts/psc
@@ -1,2 +1,0 @@
-#!/bin/sh
-purs compile "$@"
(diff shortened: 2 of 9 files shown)


garyb:
> Does this mean we aren't going to have any more releases in the 0.11.x series? Not that that's a problem, I'm just wondering.

I'm hoping to get source span annotations in for the `Expr` part of the AST (I'm working on it now, although it won't be finished tonight), as we're missing position info from some errors since I added annotations elsewhere, and then release that as maybe the last 0.11.x. Phil suggested these PRs could either be merged into a branch or hang around until that's done.

hdgarrood:
Ok, cool: in that case can we add the scripts to `extra-source-files` for the next 0.11.x?

hdgarrood:
Actually, come to think of it, it might be confusing to add them and immediately remove them again, and Homebrew already have a workaround for 0.11.6 which ought to continue to work for 0.11.7.

Issue #3012, opened 2017-07-31
Hierarchy creates improper files


joneshf:
The `hierarchy` command is making some improper files:

```dot
➜  purescript-prelude git:(master) purs hierarchy src/Control/Monad.purs 
digraph Control_Monad {
  ProperName {runProperName = "Applicative"} -> ProperName {runProperName = "Monad"};
  ProperName {runProperName = "Bind"} -> ProperName {runProperName = "Monad"};
}
```

It produces this image if run through `dot`:

![control_monad](https://user-images.githubusercontent.com/1356417/28758128-ae396df0-7547-11e7-9c4c-e2a027cb98f7.png)

The problem stems from not unwrapping the `ProperName`.

PR impending.

hdgarrood:
Oh, wow, this must have been broken for years - it was a really long time ago that we switched to derived Show instances! Thanks for the report!

hdgarrood:
Fixed by #3013 .

Issue #3010, opened 2017-07-30
Appveyor issues


paf31:
Just consolidating these under a single issue

- [x] #2774 AppVeyor deployment script fails
- [ ] #2880 AppVeyor CI fails with Access Violation
- [x] #2990 Caching is broken in AppVeyor CI again

hdgarrood:
I looked at GHC trac and saw that a patch for improving Access Violation reporting made it into GHC 8.2.1: https://ghc.haskell.org/trac/ghc/ticket/13108. I'll try updating to GHC 8.2.1 in a branch and see if that sheds any light.

hdgarrood:
[This build](https://ci.appveyor.com/project/hdgarrood/purescript-lfgff/build/2596), which I triggered by pushing to a feature branch, seems to have cleared the cache for the `master` branch, which seems to have fixed the caching issue — all recent builds are completing in 10 minutes or so. I guess AppVeyor must not keep separate caches for separate branches.

hdgarrood:
#2774 should be fixed by https://github.com/purescript/purescript/commit/ee44ff01d217429e210aa143452d52fc137dfc33

hdgarrood:
AppVeyor seems to be working now: https://ci.appveyor.com/project/hdgarrood/purescript-lfgff/history

So I'm going to close this and just hope that it was something which has been fixed since this was opened (either by GHC or by AppVeyor).

garyb:
The deployment stuff still doesn't work, I manually uploaded the windows binaries for the last two releases. I didn't look into why or anything though.

hdgarrood:
Oh ok - I can look into this, thanks for the heads up.

PR #3013, opened 2017-07-31
Don't use `Show` instance for pretty printing

joneshf:
This is a fix for https://github.com/purescript/purescript/issues/3012

The output looks like this now:

```dot
➜  purescript git:(hierarchy-fix) ✗ stack exec purs -- hierarchy ../purescript-prelude/src/Control/Monad.purs
digraph Control_Monad {
  Applicative -> Monad;
  Bind -> Monad;
}
```

And produces this image if run through `dot`.

control_monad

---

This feels like it should be verified/checked in some way so it is less likely to break in the future. Should it be added to the test suite somehow? I'm not familiar enough with it to know if that's possible or a huge amount of work.

### DIFF
--- a/tests/TestHierarchy.hs
+++ b/tests/TestHierarchy.hs
@@ -62,4 +62,4 @@ main = hspec $ do
 
         let graph = typeClassGraph mainModule
 
-        graph `shouldBe` Just (GraphName "Main", Graph "digraph Main {\n  A;\n  A -> B;\n}")
+        graph `shouldBe` Just (Graph (GraphName "Main") (Digraph "digraph Main {\n  A;\n  A -> B;\n}"))
--- a/tests/TestHierarchy.hs
+++ b/tests/TestHierarchy.hs
@@ -1,0 +1,26 @@
+module TestHierarchy where
+
+import System.Process (readCreateProcess, shell)
+
+import Test.Hspec (Spec, describe, hspec, it, shouldBe)
+
+import TestUtils (pushd)
+
+main :: IO ()
+main = pushd "tests/support/hierarchy" (hspec spec)
+
+spec :: Spec
+spec = do
+  describe "hierarchy" $ do
+    it "generates usable graphviz graphs" $ do
+      let expected = unlines
+            [ "digraph Main {"
+            , "  A;"
+            , "  A -> B;"
+            , "  C;"
+            , "}"
+            ]
+
+      actual <- readCreateProcess (shell "purs hierarchy src/Main.purs") ""
+
+      actual `shouldBe` expected
--- a/tests/TestHierarchy.hs
+++ b/tests/TestHierarchy.hs
@@ -1,26 +1,38 @@
+{-# LANGUAGE OverloadedStrings #-}
 module TestHierarchy where
 
-import System.Process (readCreateProcess, shell)
-
-import Test.Hspec (Spec, describe, hspec, it, shouldBe)
-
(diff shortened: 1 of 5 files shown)


hdgarrood:
It's probably not too hard to add a little bit to the test suite to check the output matches a hardcoded expected output for a given module; that might be worth doing.

joneshf:
Cool. I'll look into it.

joneshf:
It's unclear to me why one of the five travis builds failed. Is that a common thing, or did I do something wrong?

kRITZCREEK:
You need to run hpack/run a build with stack so that the `.cabal` file gets updated with your newest file additions. Possibly you need to add something to these globs: https://github.com/purescript/purescript/blob/db99634aca48f3b57a781767fbd5d344a539f18d/package.yaml#L18-L33 if your new additions aren't covered by them.

hdgarrood:
> You need to run hpack/run a build with stack so that the .cabal file gets updated with your newest file additions.

I don't think this is right - the cabal file is no longer in git and all the CI scripts use stack, so there's no way CI can fail because of a stale cabal file.

> Possibly you need to add something to these globs: https://github.com/purescript/purescript/blob/db99634aca48f3b57a781767fbd5d344a539f18d/package.yaml#L18-L33 if your new additions aren't covered by them.

This is correct, I think.

The reason only one job failed is that that job is the one that checks the package (and tests) still work when built from a source distribution, i.e. what gets uploaded to Hackage. The extra-source-files list controls what goes into a source distribution, and currently it looks like the new test files you added aren't on that list.

tests/TestHierarchy.hs
```diff
@@ -0,0 +1,26 @@
...
+    it "generates usable graphviz graphs" $ do
+      let expected = unlines
+            [ "digraph Main {"
+            , "  A;"
+            , "  A -> B;"
+            , "  C;"
+            , "}"
+            ]
+
+      actual <- readCreateProcess (shell "purs hierarchy src/Main.purs") ""
```
hdgarrood: I think I'd rather not involve a shell in this test, actually. How about refactoring the `App.Command.Hierarchy` module to extract a function `[Module] -> Text`, putting that function into the library portion of the purescript package, and just exercising that function in these tests?

joneshf: Sounds good.

joneshf: Well wait, it sort of feels like we should have more than one test. Would you be against having at least one test that ensures everything is wired up properly for the command people would use?

I don't think hierarchy is really used much, but it'd be nice if breakages didn't take years to find out.

hdgarrood: Having spent far too long messing about with CI scripts, my general approach is the fewer shells, the better. I think they're a nightmare. I reckon it's very likely that a test that relies on them will break because of a different value of the PATH variable in some environment somewhere, or something like that.

All the other tests in this repo work the same way: we don't test the CLI for the other commands, just the library part. But after moving as much as is reasonable into the library part, I think that's fine.

joneshf: No worries.

joneshf:
Let me know if there's anything else you'd like to see here. I'm happy to squash everything to one commit if you'd prefer, or squash through GitHub's UI.

hdgarrood:
That AppVeyor failure is not to do with your code (see #3010) so I'll merge this. Thanks!

PR #3014, opened 2017-07-31
Remove some references to old executables

hdgarrood:
This removes all references to the old executables except for those in
the IDE code, as I wasn't sure what to do about those.


paf31:
👍 Thanks!

PR #3015, opened 2017-08-02
Use order given in export list in generated docs

### LINKED ISSUE #3003: Docs should use the same order as the export list
For example, given the following module:

```purescript
module Main
  ( foo
  , bar
  ) where

bar = 0
foo = 0
```

produces:

    ## Module Main

    #### `bar`

    ``` purescript
    bar :: Int
    ```

    #### `foo`

    ``` purescript
    foo :: Int
    ```

If there is no export list, the docs should use the order that declaration appear in the source file.

See also #2525 and #2242.

hdgarrood:
If a module has an explicit export list, the declarations in the
generated docs now appear in the same order as in the export list.
Otherwise, they appear in the same order as they did in the source file.
Fixes #3003

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -528,11 +528,15 @@ declName (ValueDeclaration _ n _ _ _) = Just (IdentName n)
 declName (ExternDeclaration _ n _) = Just (IdentName n)
 declName (ExternDataDeclaration _ n _) = Just (TyName n)
 declName (ExternKindDeclaration _ n) = Just (KiName n)
-declName (ValueFixityDeclaration _ _ _ n) = Just (ValOpName n)
-declName (TypeFixityDeclaration _ _ _ n) = Just (TyOpName n)
+declName (FixityDeclaration _ (Left (ValueFixity _ _ n))) = Just (ValOpName n)
+declName (FixityDeclaration _ (Right (TypeFixity _ _ n))) = Just (TyOpName n)
 declName (TypeClassDeclaration _ n _ _ _ _) = Just (TyClassName n)
 declName (TypeInstanceDeclaration _ n _ _ _ _) = Just (IdentName n)
-declName _ = Nothing
+declName ImportDeclaration{} = Nothing
+declName BindingGroupDeclaration{} = Nothing
+declName DataBindingGroupDeclaration{} = Nothing
+declName BoundValueDeclaration{} = Nothing
+declName TypeDeclaration{} = Nothing
 
 -- |
 -- Test if a declaration is a value declaration
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -342,6 +342,17 @@ declRefSourceSpan (ModuleRef ss _) = ss
 declRefSourceSpan (KindRef ss _) = ss
 declRefSourceSpan (ReExportRef ss _ _) = ss
 
+declRefName :: DeclarationRef -> Name
+declRefName (TypeRef _ n _) = TyName n
+declRefName (TypeOpRef _ n) = TyOpName n
+declRefName (ValueRef _ n) = IdentName n
+declRefName (ValueOpRef _ n) = ValOpName n
+declRefName (TypeClassRef _ n) = TyClassName n
+declRefName (TypeInstanceRef _ n) = IdentName n
+declRefName (ModuleRef _ n) = ModName n
+declRefName (KindRef _ n) = KiName n
+declRefName (ReExportRef _ _ ref) = declRefName ref
+
 getTypeRef :: DeclarationRef -> Maybe (ProperName 'TypeName, Maybe [ProperName 'ConstructorName])
 getTypeRef (TypeRef _ name dctors) = Just (name, dctors)
 getTypeRef _ = Nothing
@@ -509,6 +520,19 @@ declSourceAnn (TypeInstanceDeclaration sa _ _ _ _ _) = sa
 
 declSourceSpan :: Declaration -> SourceSpan
 declSourceSpan = fst . declSourceAnn
+
+declName :: Declaration -> Maybe Name
+declName (DataDeclaration _ _ n _ _) = Just (TyName n)
+declName (TypeSynonymDeclaration _ n _ _) = Just (TyName n)
+declName (ValueDeclaration _ n _ _ _) = Just (IdentName n)
+declName (ExternDeclaration _ n _) = Just (IdentName n)
+declName (ExternDataDeclaration _ n _) = Just (TyName n)
+declName (ExternKindDeclaration _ n) = Just (KiName n)
+declName (ValueFixityDeclaration _ _ _ n) = Just (ValOpName n)
+declName (TypeFixityDeclaration _ _ _ n) = Just (TyOpName n)
+declName (TypeClassDeclaration _ n _ _ _ _) = Just (TyClassName n)
+declName (TypeInstanceDeclaration _ n _ _ _ _) = Just (IdentName n)
+declName _ = Nothing
 
 -- |
 -- Test if a declaration is a value declaration
--- a/src/Language/PureScript/AST/Exported.hs
+++ b/src/Language/PureScript/AST/Exported.hs
@@ -4,10 +4,12 @@ module Language.PureScript.AST.Exported
   ) where
 
 import Prelude.Compat
+import Protolude (sortBy, on)
 
 import Control.Category ((>>>))
 
 import Data.Maybe (mapMaybe)
+import qualified Data.Map as M
 
 import Language.PureScript.AST.Declarations
 import Language.PureScript.Types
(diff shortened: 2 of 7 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -510,6 +521,19 @@ declSourceAnn (TypeInstanceDeclaration sa _ _ _ _ _) = sa
...
+declName (TypeSynonymDeclaration _ n _ _) = Just (TyName n)
+declName (ValueDeclaration _ n _ _ _) = Just (IdentName n)
+declName (ExternDeclaration _ n _) = Just (IdentName n)
+declName (ExternDataDeclaration _ n _) = Just (TyName n)
+declName (ExternKindDeclaration _ n) = Just (KiName n)
+declName (ValueFixityDeclaration _ _ _ n) = Just (ValOpName n)
+declName (TypeFixityDeclaration _ _ _ n) = Just (TyOpName n)
+declName (TypeClassDeclaration _ n _ _ _ _) = Just (TyClassName n)
+declName (TypeInstanceDeclaration _ n _ _ _ _) = Just (IdentName n)
+declName _ = Nothing
```
kRITZCREEK: Depending on how many more there are I'd suggest to write the cases out. Wildcards like this one have caused the most annoying bugs...

src/Language/PureScript/AST/Exported.hs
```diff
@@ -144,3 +141,16 @@ isDctorExported ident (Just exps) ctor = test `any` exps
   test (TypeRef _ ident' Nothing) = ident == ident'
   test (TypeRef _ ident' (Just ctors)) = ident == ident' && ctor `elem` ctors
   test _ = False
+
```
kRITZCREEK: Could we do these reorderings inside `Docs` instead? The ExternsFile format should still contain all the DeclarationRefs.

- This way we don't slow down the existing compilation process
- The implicit requirement that these declarations have to be sorted are similar to the topo sort ordering requirement on ExternsFiles, they are not document/enforced (in the types) and cause weird errors.

hdgarrood: I don't see how we would, that information is gone by then. Also this code isn't used in the normal compilation process; just by docs and the repl (and the repl should be using Environment instead anyway, this is mistake I made ages ago)

src/Language/PureScript/AST/Declarations.hs
hdgarrood: :+1:. There aren't many more cases.

src/Language/PureScript/AST/Exported.hs
hdgarrood: To clarify: Docs currently uses the AST, not externs. It might make sense to use externs instead but if we are going to change that I'd rather do it separately.

hdgarrood: Also if we do accidentally break the declaration ref ordering later, a) there's now a test so it'll be caught, and b) the only way that would manifest would be that some of the docs appear in the wrong order.

hdgarrood: I suppose we could make a note of the original export lists before desugaring so that we don't have to make any changes to Sugar.Names, but I think that would be a bit awkward. Also I think a good reason to have it there is that other places can benefit, e.g. I think :browse in the repl currently uses an ordering which doesn't resemble either the decl ordering or the export list ordering, which is not ideal.

kRITZCREEK: > I suppose we could make a note of the original export lists before desugaring so that we don't have to make any changes to Sugar.Names

I think this is the way to go. This would also allow doing things like sectioning by putting doc comments in between the exports like haddock allows. `psci` could just run the same ordering code as docs, which would also allow the sections to appear within the repl (I think that would be nice).

hdgarrood: I'm not sure, this seems like premature optimisation to me. We would just as easily be able to handle doc comments in between the exports inside Sugar.Names by preserving the source spans there, right? And that would likely lead to better errors if we did it there.

kRITZCREEK: Oh, I just saw your comment about this code being `docs & repl` specific anyway. In that case I'm not opposed to doing it this way, but I still think having an actual document structure derived from the export section would be a nice way to go about extracting documentation.

hdgarrood: I only realised after I wrote that that you were probably referring to my changes in Sugar.Names, which of course is used by the main compiler :sweat_smile:

hdgarrood: Just to be safe I tested compiling `purescript-css` (because the `CSS` module is probably the largest one I know of in terms of the number of declarations it exports), and there was no observable difference between the time it took on this branch vs the current master.

hdgarrood: > I still think having an actual document structure derived from the export section would be a nice way to go about extracting documentation.

The problem is, we need the `desugarImports` part to have run before we can start getting useful information out of it, because of things like modules without export lists, or export lists which use `Foo(..)` to export all the data constructors, or so on. Of course it would be possible to handle those things too, but at that point we're basically reimplementing `Sugar.Names`.

kRITZCREEK: I don't think premature optimization applies on the critical path.

As for having to reimplement `Sugar.Names`, you're basically just mapping over the declaration refs part of the modules, can't you compose that with the existing desugarings or if that's impossible make it configurable? I don't think we should be doing more work in the compile process because it's more convenient for documentation generation.

hdgarrood: I tested to see if there was any change in compile times for modules with large numbers of exports, and the result was that any change is dwarfed by random fluctuation; this is why I call it premature.

kRITZCREEK: I'm not saying there is any accidental quadratic behaviour in that code. It's never going to be "noticeable" on its own when you're running the full compiler. But it's doing more work and creating more garbage that we don't need to do except for when generating documentation.

This is also a case for separation of concerns. Ide will want to run the desugaring pipeline decoupled from parser and type checker to do name resolution (It doesn't care for the ordering of exports), and when the user is directly interacting with it milliseconds do degrade the user experience.

hdgarrood: Results for compiling https://github.com/hdgarrood/solitaire from scratch on current master vs this branch, n=10:

```
master:
mean: 15.907499999999999s
std dev: 0.3209788933870889s
reordered:
mean: 15.8984s
std dev: 0.29505531684753594s
```

I chose that repo because it depends on halogen, css, and dom, all of which contain some fairly large modules with lots of exports and re-exports.

Note that the std dev is much larger than the difference between the means, so there's no evidence to suggest this change has had any performance effect.

Full results at https://gist.github.com/hdgarrood/963eb75bc566074300e438996eeda11e

hdgarrood: In case it's not clear, my stance is that the cost of adding extra paths outweighs the performance benefit  of not doing this work in the main compiler path, in terms of maintainability.

kRITZCREEK: If you really want to benchmark this you want to set the function up in isolation and benchmark it with criterion. Of course you won't see it in the noise of _16 seconds_ of running the compiler. But performance is not some magical thing that we can only assess through measurement. It's obvious that the new version is doing more work than the old one.

kRITZCREEK: Anyway, I don't want to continue this. You seem to feel very strongly about not duplicating code, but I don't agree that changing the desugaring pass to do additional work and complexity that is only used in Docs is worth that.

hdgarrood: Yes, I don't dispute that it's doing more work, I'm just saying that it's negligible. I don't think benchmarking the function in isolation would give us any meaningful information.

hdgarrood: It's not duplicated code I'm worried about, it's about unnecessary API surface. The larger an API is, the harder it is for people to read the code, the more possibilities there are to consider when you're using it, and just in general the harder it is to work with. I do genuinely believe that if we added an option or an alternate variant to any function in the compiler like this because not all the work it was doing was necessary in every situation where it gets called, if we applied this principle consistently throughout the code base, we'd very quickly end up with a big mess.

hdgarrood:
@paf31 / @garyb what do you think about reordering the declaration refs in `desugarImports` every time vs having a separate version of `desugarImports` for docs? I'd prefer it this way, obviously, but I'm not dead set against the alternative.

paf31:
I can't remember if there is a reason why we reorder the imports, but if the tests pass, I think it sounds fine. Might want to test on the package set too, just to be sure.

hdgarrood:
Ok, will do. Looking at the code I would guess that the reason the export list comes out of `desugarImports` in a different order to how it went in is just that it was more convenient to write the code that way.

hdgarrood:
If I run `stack exec bash` from inside my `purescript` working directory to make sure the right `purs` is at the front of my PATH, then do `psc-package init && psc-package verify-set` in a new, empty directory, everything passes. Is that what you had in mind?

paf31:
Yes, that's it, thanks!

hdgarrood:
Great, so do you think this is ok to merge?

paf31:
I think so 👍

hdgarrood:
Fab, thanks both :)

PR #3017, opened 2017-08-03
Embed CSS for HTML docs

### LINKED ISSUE #3016: Canonical location for CSS for HTML docs
Currently we have `app/static/pursuit.css` in this repo, as well as `static/css/pursuit.css` in the Pursuit repo, and `static/css/pursuit.less` (which generates it). Since these are starting to diverge I think we need to have a single place for this CSS. How about:

* Moving the LESS source to this repo
* Checking in the result of compiling it to CSS (we've kind of done this already)
* Creating a module `Language.PureScript.Docs.CSS`, which just exports the compiled CSS as `docsCSS :: Text`, via `file-embed`, so that Pursuit can use that.

hdgarrood:
Fixes #3016 

At the moment, we have a part of the library reaching into `app/static` for embedded files, which is perhaps a bit wrong; should I create a `src/static` directory for files which are embedded in the library?

### DIFF
--- a/app/Command/Docs/Html.hs
+++ b/app/Command/Docs/Html.hs
@@ -15,7 +15,6 @@ import           Data.List (sort)
 import           Data.Text (Text)
 import           Data.Text.Lazy (toStrict)
 import qualified Data.Text as T
-import           Data.FileEmbed (embedStringFile)
 import qualified Language.PureScript as P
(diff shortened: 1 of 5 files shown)


hdgarrood:
Thanks for the review!

PR #3025, opened 2017-08-05
Run AppendSymbol in reverse

paf31:
Should be merged with https://github.com/purescript/purescript-typelevel-prelude/pull/19

We can run `AppendSymbol` in reverse, stripping a prefix or suffix from a type level string.

This lets us write simple parsers at the type level, which might lead to some interesting applications.

@LiamGoodacre Could you please review this?

### DIFF
--- a/examples/passing/AppendInReverse.purs
+++ b/examples/passing/AppendInReverse.purs
@@ -1,0 +1,24 @@
+module Main where
+
+import Prelude
+import Type.Data.Symbol (class AppendSymbol, SProxy(..))
+import Control.Monad.Eff.Console (log)
+
+class Balanced (sym :: Symbol)
+
+instance balanced1 :: Balanced ""
+instance balanced2
+  :: ( AppendSymbol "(" sym1 sym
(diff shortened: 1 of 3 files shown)


paf31:
Thanks!

paf31:
Oops, I just realized we need a breaking change to `typelevel-prelude` for this.

What do people think? Should I revert this, or shall we just merge in the things for 0.12.0 and make that release now?

hdgarrood:
I'm happy either way. Were there any other breaking changes we want to do soonish?

paf31:
Constraint kinds was the only one, and I think I have a plan to implement something which is non-breaking instead.

paf31:
Note to self: I also need to fix the tag for typelevel-prelude for the tests.

garyb:
I guess we can release 0.11.6 from a branch then, and maybe not merge much else into master until that is done? 😛 

I'm hoping to have expr source annotations done today, but it's hard to tell since the errors come in an endless stream rather than being able to tell all at once.

PR #3026, opened 2017-08-06
Add inlining for runEffFnN/mkEffFnN

nwolverson:
Reusing the same inlining code from runFnN/mkFnN


src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
```diff
@@ -190,37 +192,54 @@ inlineCommonOperators = everywhereTopDown $ applyAll $
...
     collectArgs :: Int -> [Text] -> AST -> Maybe ([Text], [AST])
     collectArgs 1 acc (Function _ Nothing [oneArg] (Block _ js)) | length acc == n - 1 = Just (reverse (oneArg : acc), js)
     collectArgs m acc (Function _ Nothing [oneArg] (Block _ [Return _ ret])) = collectArgs (m - 1) (oneArg : acc) ret
     collectArgs _ _   _ = Nothing
 
-  isNFn :: Text -> Int -> AST -> Bool
-  isNFn prefix n (Var _ name) = name == (prefix <> T.pack (show n))
-  isNFn prefix n (Indexer _ (StringLiteral _ name) (Var _ dataFunctionUncurried)) | dataFunctionUncurried == C.dataFunctionUncurried =
+  isNFn :: Text -> Text -> Int -> AST -> Bool
+  isNFn _ prefix n (Var _ name) = name == (prefix <> T.pack (show n))
```
nwolverson: I think this line (preserved from the existing code) means that all `runFnX` defined in some arbitrary local module are inlined by this same scheme (verified this with an example) - shouldn't it only apply to the target module as per the next clause?

natefaubion:
Hah! I worked this up last night, and my version was exactly the same, modulo function names. One thing that I think really needs to be solved though, is that the compiler will insert something like this in magic do blocks (if you use case/if):
```js
main = function __do () {
  foo = bar();
  (function () {
    var $1 = ...;
    if ($1 ...) {
      return someEff;
    }
    return otherEff;
  })()();
  return unit;
}
```

This is a problem because with EffFn inlining, you lose a significant optimization opportunity:

```js
(function () {
  if (...) {
    return function () {
      return someEffFn(1, 2, 3);
    }
  }
  return function () {
    return someEffFn(4, 5, 6);
  }
})()()
```

If we floated that call into the leaves (I think `applyReturns` in `MagicDo.hs`), then the IIFE code should take care of it.

natefaubion:
FWIW, I'm working converting halogen-vdom to EffFn, so we can compare output and performance.

nwolverson:
Oops.

So I verified that the simple binding in magic-do just becomes a straight call:

```javascript
function __do() {
...
  $foreign.log2("Hello", "World");
...
}
```
But yeah this is just the case for the straightforward discard or binding cases of the magic-do optimisation.

natefaubion:
I added this case to `undo` in `MagicDo.purs`, which eliminates a lot of redundant applications. Otherwise, if you have any cases that use EffFn, you'll probably undo the optimization due to redundant applications.

```haskell
  undo (App _ (App s1 (Function s2 Nothing [] (Block ss body)) []) []) =
    App s1 (Function s2 Nothing [] (Block ss (applyReturns `fmap` body))) []
```

nwolverson:
Yup, I can see that working in my example. Feel free to PR any other changes to this branch if that would be easiest?

natefaubion:
Another change I made was to do `untilFixedPoint (return . tidyUp . tco . magicDo) js'`, which removes a lot of leftover things like `pure` applications. I don't know what the implications overall for this are.

natefaubion:
Just using `untilFixedPoint (return . tidyUp) . tco =<< untilFixedPoint (return . magicDo) js'` works well. Applying tco to a fixed point seems a bit dodgy.

src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
paf31: It certainly seems like it. Is it possible to break the current code?

src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
```diff
@@ -73,6 +73,8 @@ magicDo = inlineST . everywhere undo . everywhereTopDown convert
   -- Remove __do function applications which remain after desugaring
   undo :: AST -> AST
   undo (Return _ (App _ (Function _ (Just ident) [] body) [])) | ident == fnName = body
+  undo (App _ (App s1 (Function s2 Nothing [] (Block ss body)) []) []) =
```
paf31: I'm not sure why this case is needed, but it makes the code harder to read in my opinion. `applyReturns` is only used in `convert` right now, and `undo` runs after the full traversal.

Can you give an example of why this is necessary? Is it possible to accomplish the same thing by modifying the `convert` function?

src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
nwolverson: Yes, I defined in my `Main` module
```purescript
runFn3 f a b c = f a b c
```
and application of this was inlined, leading to runtime failure. I think this line can just be removed, the optimisation still seems to apply otherwise - I just don't understand what case this was meant to address.

paf31: Ok great, could you please add a test for that then, to make sure it doesn't break any more? Thanks.

paf31: Removing the line sounds fine to me if it works.

src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
nwolverson: Defer to @natefaubion here. But the example code is in https://github.com/purescript/purescript/pull/3026#issuecomment-320511849 - with this change an extra thunk is removed.

natefaubion: I made a PR to this branch that rearranges the optimizations, and removes undo in favor of running convert to a fixed point. The problem was that some cases that covert inlines only arise after undo is run.

nwolverson: Oh I didn't get a notification, sorry

src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
nwolverson: Hmm, I didn't realise the compiled test output actually gets run, for some reason I thought there were just compiles/warns/errors tests.

paf31:
Given the new `untilFixedPoint` call, we should try to measure any performance impact here before this gets merged.

Also, since this code is quite fragile, we should try to build the package set, and actually run some package test suites to make sure nothing gets broken.

natefaubion:
Maybe the fixed point isn't necessary? It should stabilize after another run, So maybe you'd just need to split up convert into two functions, and inline the repetition.

nwolverson:
Removed the case that was firing on local definitions rather than the particular module name, now there is a test failure for `AppendInReverse.purs` which I hope is unrelated

paf31:
If you merge master, the failure should go away.

paf31:
We can try this out extensively once the release candidate is done, just to make sure there isn't a significant performance regression.

Thanks!

PR #3028, opened 2017-08-06
Use Hspec in TestDocs

hdgarrood:
This doesn't have any immediate effects apart from making the output of
the docs tests slightly nicer. It does, however, get us one closer to
switching over to a nicer test runner such as tasty's (see #2848 for
more details).


hdgarrood:
The Travis failure is unrelated (due to it not being able to restore the cache).

PR #3031, opened 2017-08-07
Allow blaze-html 0.9

felixonmars:
None of the changelog items were used by us, and the build is successful.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -37,7 +37,7 @@ dependencies:
   - ansi-terminal >=0.6.2 && <0.7
   - base >=4.8 && <5
   - base-compat >=0.6.0
-  - blaze-html >=0.8.1 && <0.9
+  - blaze-html >=0.8.1 && <0.10
   - bower-json >=1.0.0.1 && <1.1
   - boxes >=0.1.4 && <0.2.0
   - bytestring


felixonmars:
I am packaging purescript for Arch, and we want to keep the packages up to date if not broken :)

https://www.archlinux.org/packages/community/x86_64/purescript/

hdgarrood:
What happened when Arch's blaze-html package was updated to 0.9 but purescript still had an upper bound preventing use of it? Does it become impossible to install Arch's purescript package at that point? Or is it that any system using purescript is now unable to update blaze-html?

felixonmars:
Arch won't update blaze-html to 0.9 if it's indeed not compatible, and will wait for a fix from purescript.

hdgarrood:
Ok, thanks - so does that mean the script that packages purescript for Arch is using --allow-newer to ignore this upper bound for now?

felixonmars:
It's using a simple sed for now - so it doesn't affect other bounds.

https://git.archlinux.org/svntogit/community.git/commit/trunk?h=packages/purescript&id=61a644393ec6534c104800d3e32cc05a38914a4b

hdgarrood:
Ah right, I see. Thanks!

paf31:
Thanks!

PR #3032, opened 2017-08-08
Fix compatibility with protolude 0.2

felixonmars:
protolude 0.2 exports <&> which will conflict the one imported from Control.Lens. This patch fixes it.

```
[128 of 147] Compiling Language.PureScript.Ide.Completion ( src/Language/PureScript/Ide/Completion.hs, dist/build/Language/PureScript/Ide/Completion.o )

src/Language/PureScript/Ide/Completion.hs:41:3: error:
    Ambiguous occurrence ‘<&>’
    It could refer to either ‘Protolude.<&>’,
                             imported from ‘Protolude’ at src/Language/PureScript/Ide/Completion.hs:12:1-26
                          or ‘Control.Lens.<&>’,
                             imported from ‘Control.Lens’ at src/Language/PureScript/Ide/Completion.hs:14:1-46
                             (and originally defined in ‘Control.Lens.Lens’)

src/Language/PureScript/Ide/Completion.hs:53:3: error:
    Ambiguous occurrence ‘<&>’
    It could refer to either ‘Protolude.<&>’,
                             imported from ‘Protolude’ at src/Language/PureScript/Ide/Completion.hs:12:1-26
                          or ‘Control.Lens.<&>’,
                             imported from ‘Control.Lens’ at src/Language/PureScript/Ide/Completion.hs:14:1-46
                             (and originally defined in ‘Control.Lens.Lens’)

src/Language/PureScript/Ide/Completion.hs:54:3: error:
    Ambiguous occurrence ‘<&>’
    It could refer to either ‘Protolude.<&>’,
                             imported from ‘Protolude’ at src/Language/PureScript/Ide/Completion.hs:12:1-26
                          or ‘Control.Lens.<&>’,
                             imported from ‘Control.Lens’ at src/Language/PureScript/Ide/Completion.hs:14:1-46
                             (and originally defined in ‘Control.Lens.Lens’)
```

### DIFF
--- a/src/Language/PureScript/Ide/Completion.hs
+++ b/src/Language/PureScript/Ide/Completion.hs
@@ -9,7 +9,7 @@ module Language.PureScript.Ide.Completion
        , applyCompletionOptions
        ) where
 
-import           Protolude
+import           Protolude hiding ((<&>))
 
 import           Control.Lens hiding ((&), op)
 import           Data.Aeson


felixonmars:
Hmm, looks like it's easier to require protolude >0.2 to avoid the warning. Do you think it's acceptable?

kRITZCREEK:
I will handle it this evening.

hdgarrood:
I think this is at risk of becoming too much of a time sink, actually. I don't think we should aim to support the latest versions of everything all the time - if we did want to do that we wouldn't have removed PureScript from Stackage. Is there any way you can package PureScript for Arch in such a way that it uses the versions of dependencies based on the Stackage snapshot we're on?

felixonmars:
I would wonder if it makes a difference if I actually help with trivial changes like this one. For bigger changes I would assume other projects need some time to migrate, too. If you don't want to merge a PR too quickly, it is fine.

Supporting multiple versions of some libraries at the same time would not be easy for distributions like Arch :(

hdgarrood:
It's not just that we can't guarantee we'll get around to looking at PRs quickly, it's that I don't really want to spend any more time than is absolutely necessary thinking about dependencies and version ranges. In general I would expect that supporting building purescript with larger ranges of dependencies, and especially across breaking changes of things which we use in the compiler, would be some (not a lot, but some) extra effort for not really anything in return. One of the reasons I really like Stack is that we can essentially forget about dependencies entirely unless we have a specific reason to want to move to a newer compiler or newer version of some dependency.

Is there not any way you can package PureScript for Arch in such a way that the Haskell libraries it depends on are brought in via some mechanism other than Arch's package manager?

hdgarrood:
We could also consider statically linking libtinfo; that way the Linux binaries should hopefully just work on Arch too.

kRITZCREEK:
@hdgarrood it's not just libtinfo. Arch links all haskell libraries dynamically. Not that I think that's a great idea, but it is how it is. I think we should update all our current libraries to get to GHC 8.2 as quickly as possible, but I agree that there's not a lot of value in keeping up to date _all the time_ for the compiler.

hdgarrood:
Surely if we statically linked everything when building the release binaries, the Linux binary would work on Arch though, regardless of whether Arch's package manager uses dynamic linking for Haskell libraries by default?

hdgarrood:
I think it might be safer to wait a little bit before upgrading to GHC 8.2, it hasn't been out very long.

felixonmars:
I don't really get it here - if you are going to upgrade to a newer GHC and/or stack snapshot some day anyway, it doesn't change what needs to be changed and if some of the patches are ready, it will save some time instead.

hdgarrood:
And it's not safe to assume that we always will update eventually - for example, I'd like to drop the Protolude dependency.

felixonmars:
I see. I don't think I can do anything to help, so maybe it would be better maintained in the AUR with a stack-based solution.

danr:
Workaround when installing with `cabal-install`:

    cabal install purescript protolude-0.1.10

Issue #3023, opened 2017-08-04
Some thoughts around our AST design


kRITZCREEK:
I've spent some time looking into how other compilers handle their ASTs to see if there is something we could learn or improve our current design. I'm listing my findings here and make a few suggestions to hopefully get some feedback/discussion from you folks :)

## Current shortcomings

I'll start out motivating this disussion by talking about some of the shortcomings/problems I ran into with our current AST types.

- A lot of functions that operate on our AST are partial, because only subsets of the available constructors can appear in certain stages of compilation.

- We don't have "record" labels for our AST datatypes, because they would be partial accessors, since our datatypes are huge sums. This means one needs a lot of lookups in the documentation when pattern matching, and all the pattern matches have to be updated when new fields are added. (The recent SourceSpan addition PRs from Gary are a good example of the problems this causes)

- There's no way for our AST nodes to refer to another node that is a parent or higher up in the hierarchy. This is fine when doing the kind of desugaring top-down traversals we do in the compiler, but it doesn't work as well if I find a node in the hierarchy (for example by matching on a source position) and now try to collect data starting from that node

- Our AST lacks the ability to accurately describe the source it was parsed from (because some desugarings already happen during parsing). This means our AST is not well suited to writing a formatter or other tooling that tries to operate on the AST and translate the results back into source code.

## Extract our individual variants into their own datatypes

I did a similar transformation for IDE's data types and it makes it possible to have record accessors and helper functions that operate on these types with a proper signature.

Eg:
```haskell
data Declaration 
  = TypeDeclaration SSpan Ident Type 
  | ...

pprintTypeDeclaration :: (SSpan, Ident, Type) -> Text

-- vs:

data Declaration 
  = DTypeDeclaration TypeDeclaration 
  | ...
data TypeDeclaration = TypeDeclaration
  { tdSpan :: SSPan
  , tdIdentifier :: Ident
  , tdType :: Type
  }

pprintTypeDeclaration :: TypeDeclaration -> Text
```

While this involves a fair bit of boilerplate it makes it easier to pass around proof that a certain pattern match has already been done. It also makes it easier to build sums for subsets of all the different Declarations, and still allows me to use the utility functions defined for individual Declarations. In IDE's code the boilerplate pain is mitigated by using Prisms, which would increase the barrier of entry for new contributors. Maybe something could be done with patterns instead.

## Introduce a node _index_

[...]

PR #3034, opened 2017-08-11
Introduces a type for type declarations

kRITZCREEK:
This is a first foray into one of the changes suggested in #3023. I just picked TypeDeclarations, because they seemed like a simple enough data type. We'll have to decide if it's worth doing this for all the constructors and if we can improve the ergonomics around it with patterns. While doing this change I already spotted some opportunities for decreasing partiality around type class declarations.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -407,6 +407,19 @@ isExplicit :: ImportDeclarationType -> Bool
 isExplicit (Explicit _) = True
 isExplicit _ = False
 
+data TypeDeclarationData = TypeDeclarationData
+  { tydeclSourceAnn :: !SourceAnn
+  , tydeclIdent :: !Ident
+  , tydeclType :: !Type
+  } deriving (Show, Eq)
+
+onTypeDeclaration :: (TypeDeclarationData -> TypeDeclarationData) -> Declaration -> Declaration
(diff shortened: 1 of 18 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -407,6 +407,22 @@ isExplicit :: ImportDeclarationType -> Bool
 isExplicit (Explicit _) = True
 isExplicit _ = False
 
+data TypeDeclarationData = TypeDeclarationData
```
paf31: Could you please add some line comments?

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -407,6 +407,22 @@ isExplicit :: ImportDeclarationType -> Bool
...
+  { tydeclSourceAnn :: !SourceAnn
+  , tydeclIdent :: !Ident
+  , tydeclType :: !Type
+  } deriving (Show, Eq)
+
+onTypeDeclaration :: (TypeDeclarationData -> TypeDeclarationData) -> Declaration -> Declaration
```
paf31: I've typically called these `overX` in the past, in line with `lens`.

kRITZCREEK:
I added a top-level comment. I think adding comments to the fields would be very repetitive in this instance, thanks for the review!

PR #3035, opened 2017-08-13
[WIP] Source annotations for `Expr` and `Binder`

garyb:
I think there's probably still quite a lot to do here actually... pushing it as-is in case anyone else wants to help :wink: since I haven't much progress with it over the last week.


garyb:
Yeah, I don't know how much, it doesn't compile yet :)

kRITZCREEK:
I tried to continue work on this, but I'm stuck. With `Skolem` always requiring a SourceSpan now, I can't conjure something up, since we don't have positions on types yet.

This is inside `unifyTypes`.

I think we need to keep the SourceSpan for Skolem's optional until we have positions on all the types.

garyb:
Ok, makes sense. I wasn't sure when I made that change whether it would be possible, for the reasons you mention. That's not terrible though - the error itself can have a position for sure (since it'll arise from an expression/decl), it's just we won't always know it for the originating scope.

kRITZCREEK:
Unfortunately this didn't work out, we need to find a more incremental approach.

Issue #3036, opened 2017-08-13
[purs ide] Externs parse failure needs a better error


kRITZCREEK:
This is where the error is coming from: https://github.com/purescript/purescript/blob/7a59b02221df1f906efa4a911b8ad83c86d2fdc6/src/Language/PureScript/Ide/Externs.hs#L34-L54

I added a case to show the version mismatch, but that only works if the outdated externsfile still parses with the new version, which is a silly assumption to make. We should ignore the remaining structure and read the version field with Aeson which should allow us to give a version mismatch error even if the structure of the ExternsFile format has changed.

joneshf:
Should we attempt to parse just the version in the `Nothing` case? Or is there some other path to success here?

kRITZCREEK:
Yeah I think that's the path forward, the version should always be available as a top level field: https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Externs.hs#L41. And even if that fails we should talk about a potential compiler version mismatch.

kRITZCREEK:
Fixed in #3037

PR #3037, opened 2017-08-13
Improves ExternsFile parse errors

jkachmar:
Took a quick pass at this, behavior's pretty much unchanged except that it parses the `efVersion` field directly so we can get better errors even if the entire `externs.json` doesn't successfully parse.

Tests pass, and it looks good from poking at it in the REPL.

Addresses #3036

### DIFF
--- a/src/Language/PureScript/Ide/Externs.hs
+++ b/src/Language/PureScript/Ide/Externs.hs
@@ -24,6 +24,7 @@ import           Protolude hiding (to, from, (&))
 import           Control.Lens
 import           "monad-logger" Control.Monad.Logger
 import           Data.Aeson (decodeStrict)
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/Ide/Externs.hs
```diff
@@ -36,16 +37,19 @@ readExternFile
   => FilePath
   -> m P.ExternsFile
 readExternFile fp = do
-   parseResult <- liftIO (decodeStrict <$> BS.readFile fp)
-   case parseResult of
+   externsFile <- liftIO (BS.readFile fp)
+   case (decodeStrict externsFile) of
```
kRITZCREEK: Could you get rid of these parens?

PR #3038, opened 2017-08-14
Add `:complete` directive to psci

actionshrimp:
Some wrappers around psci (such as [emacs-psci](https://github.com/ardumont/emacs-psci)) are [unable to hook into the tab-complete functionality](https://stackoverflow.com/questions/20396698/get-tab-completion-in-custom-comint-mode) provided by psci. There is some basic tab completion in emacs-psci but it currently seems to work only on filenames as far as I'm able to work out.

This exposes the completions produced usually by hitting tab to the normal eval/print mechanism via a `:complete` directive so they can be accessed by these kinds of tools (inspired by https://github.com/commercialhaskell/intero/blob/master/elisp/intero.el#L819).

I have a feeling there might be a plan afoot to have psc-ide take over more of these responsibilities in the future? Hopefully this will be able to serve as a stop-gap as the functionality is already tucked away inside psci 😃

### DIFF
--- a/src/Language/PureScript/Interactive/Directive.hs
+++ b/src/Language/PureScript/Interactive/Directive.hs
@@ -23,15 +23,16 @@ directives = map fst directiveStrings
 --
 directiveStrings :: [(Directive, [String])]
 directiveStrings =
-    [ (Help   , ["?", "help"])
-    , (Quit   , ["quit"])
-    , (Reload , ["reload"])
-    , (Clear  , ["clear"])
-    , (Browse , ["browse"])
-    , (Type   , ["type"])
-    , (Kind   , ["kind"])
-    , (Show   , ["show"])
-    , (Paste  , ["paste"])
+    [ (Help      , ["?", "help"])
+    , (Quit      , ["quit"])
+    , (Reload    , ["reload"])
+    , (Clear     , ["clear"])
+    , (Browse    , ["browse"])
+    , (Type      , ["type"])
+    , (Kind      , ["kind"])
+    , (Show      , ["show"])
+    , (Paste     , ["paste"])
+    , (Complete  , ["complete"])
     ]
 
 -- |
@@ -93,14 +94,16 @@ hasArgument _ = True
 --
 help :: [(Directive, String, String)]
 help =
-  [ (Help,    "",         "Show this help menu")
-  , (Quit,    "",         "Quit PSCi")
-  , (Reload,  "",         "Reload all imported modules while discarding bindings")
-  , (Clear,   "",         "Discard all imported modules and declared bindings")
-  , (Browse,  "<module>", "See all functions in <module>")
-  , (Type,    "<expr>",   "Show the type of <expr>")
-  , (Kind,    "<type>",   "Show the kind of <type>")
-  , (Show,    "import",   "Show all imported modules")
-  , (Show,    "loaded",   "Show all loaded modules")
-  , (Paste,   "paste",    "Enter multiple lines, terminated by ^D")
+  [ (Help,     "",          "Show this help menu")
+  , (Quit,     "",          "Quit PSCi")
+  , (Reload,   "",          "Reload all imported modules while discarding bindings")
+  , (Clear,    "",          "Discard all imported modules and declared bindings")
+  , (Browse,   "<module>",  "See all functions in <module>")
+  , (Type,     "<expr>",    "Show the type of <expr>")
+  , (Kind,     "<type>",    "Show the kind of <type>")
+  , (Show,     "import",    "Show all imported modules")
+  , (Show,     "loaded",    "Show all loaded modules")
+  , (Paste,    "paste",     "Enter multiple lines, terminated by ^D")
+  , (Complete, "<(expr?)>", "Show completion output as if pressing tab")
   ]
+
--- a/src/Language/PureScript/Interactive/Directive.hs
+++ b/src/Language/PureScript/Interactive/Directive.hs
@@ -104,6 +104,6 @@ help =
   , (Show,     "import",    "Show all imported modules")
   , (Show,     "loaded",    "Show all loaded modules")
   , (Paste,    "paste",     "Enter multiple lines, terminated by ^D")
-  , (Complete, "<(expr?)>", "Show completion output as if pressing tab")
+  , (Complete, "<prefix>",  "Show completions for <prefix> as if pressing tab")
   ]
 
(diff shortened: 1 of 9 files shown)


hdgarrood:
I think generally it's better to open issues to discuss this sort of thing first. For instance does purs ide not already provide an api for completions?

kRITZCREEK:
Well ide is another (long running) process you have to manage, and it doesn't know the things the repl knows about its current environment (what modules are imported under what alias).

src/Language/PureScript/Interactive/Parser.hs
```diff
@@ -61,16 +61,16 @@ parseDirective cmd =
...
+    Help     -> return ShowHelp
+    Quit     -> return QuitPSCi
+    Reload   -> return ReloadState
+    Clear    -> return ClearState
+    Paste    -> return PasteLines
+    Browse   -> BrowseModule <$> parseRest P.moduleName arg
+    Show     -> ShowInfo <$> parseReplQuery' (trim arg)
+    Type     -> TypeOf <$> parseRest P.parseValue arg
+    Kind     -> KindOf <$> parseRest P.parseType arg
+    Complete -> CompleteStr <$> (Right arg)
```
kRITZCREEK: Parental control 😃

hdgarrood: Haha, I should learn to stop trying to review PRs on my phone.

actionshrimp:
De-paren'd (hopefully I haven't snuck any more in - that's lispy habits leaking through 😉), and I added a command test similar to the ones that were there for a couple of other commands. Required exposing the printed output in the TestPsci TestEnv, which I hope makes sense.

tests/TestPsci/TestEnv.hs
```diff
@@ -57,25 +57,38 @@ jsEval = liftIO $ do
...
+    eval = do
+      actual <- jsEval
+      actual `equalsTo` (expected ++ "\n")
+    evalPrinted _ = return ()
+
+prints :: String -> [String] -> TestPSCi ()
+prints command expected = runAndEval command eval evalPrinted
+  where
+    eval = return ()
+    evalPrinted s = s `equalsTo` (unlines (map (" " ++) expected))
```
hdgarrood: Why prepend a space here?

tests/TestPsci/TestEnv.hs
```diff
@@ -57,25 +57,38 @@ jsEval = liftIO $ do
     Nothing                      -> putStrLn "Couldn't find node.js" >> exitFailure
 
 -- | Run a PSCi command and evaluate the output with 'eval'.
-runAndEval :: String -> TestPSCi () -> TestPSCi ()
-runAndEval comm eval =
+runAndEval :: String -> TestPSCi () -> (String -> TestPSCi ()) -> TestPSCi ()
```
hdgarrood: Can you update the documentation here so that it's clear what the new argument does please?

actionshrimp: Will do.

tests/TestPsci/TestEnv.hs
actionshrimp: The auto-completion output comes out this way, so the prepending is for what's asserted against - I'll add a comment to that effect here as well.

hdgarrood: Is it necessary that the auto-completion output comes out like that?

actionshrimp: Good point - I thought it was standard for all those kinds of commands but having just double checked it turns out others don't do it! I dont remember adding it myself but I'll have a dig through and try and tidy that up

actionshrimp:
Those changes are all in now, thanks for taking the time to have a look. The space had snuck in due to lack of trimming the string passed to the Completion command during parsing - I'd assumed it was some kind of output formatting, good catch!

src/Language/PureScript/Interactive/Parser.hs
```diff
@@ -70,7 +70,7 @@ parseDirective cmd =
     Show     -> ShowInfo <$> parseReplQuery' (trim arg)
     Type     -> TypeOf <$> parseRest P.parseValue arg
     Kind     -> KindOf <$> parseRest P.parseType arg
-    Complete -> CompleteStr <$> Right arg
+    Complete -> CompleteStr <$> Right (trim arg)
```
hdgarrood: Rather than putting the `trim` here, I think it would be better to change the declaration of `arg` so that the space is already trimmed. How about `(dstr, arg) = second trim . break isSpace cmd`? The reason for this is that I don't think either extra spaces in between the directive and the argument or trailing spaces should be able to have any effect on the result, and that's for any directive, not just `:complete`.

Also, any expression of the form `f <$> Right x` is just `Right (f x)` so I think the latter is better style, as with the former you're kind of just making the reader jump through hoops. In this specific case I think `return (CompleteStr arg)` makes the most sense.

tests/TestPsci/TestEnv.hs
```diff
@@ -56,26 +56,40 @@ jsEval = liftIO $ do
     Just (ExitFailure _, _, err) -> putStrLn err >> exitFailure
     Nothing                      -> putStrLn "Couldn't find node.js" >> exitFailure
 
--- | Run a PSCi command and evaluate the output with 'eval'.
-runAndEval :: String -> TestPSCi () -> TestPSCi ()
-runAndEval comm eval =
+-- | Run a PSCi command and evaluate the output with 'eval', and the
+-- printed-to-console output with 'evalPrinted'
```
hdgarrood: This is still a bit unclear I think. What is the difference between the output and the printed-to-console output?

actionshrimp:
I've put those changes in now - both the code style ones, and I've also tidied up the TestEnv file to hopefully make the distinction clear in function/arg names and docstrings. It is a bit of a subtle distinction for the caller due to the way the internals work.

src/Language/PureScript/Interactive/Directive.hs
```diff
@@ -93,14 +94,16 @@ hasArgument _ = True
...
+  , (Quit,     "",          "Quit PSCi")
+  , (Reload,   "",          "Reload all imported modules while discarding bindings")
+  , (Clear,    "",          "Discard all imported modules and declared bindings")
+  , (Browse,   "<module>",  "See all functions in <module>")
+  , (Type,     "<expr>",    "Show the type of <expr>")
+  , (Kind,     "<type>",    "Show the kind of <type>")
+  , (Show,     "import",    "Show all imported modules")
+  , (Show,     "loaded",    "Show all loaded modules")
+  , (Paste,    "paste",     "Enter multiple lines, terminated by ^D")
+  , (Complete, "<(expr?)>", "Show completion output as if pressing tab")
```
kRITZCREEK: Just a minor nitpick, but prefix seems more accurate than expr here.

actionshrimp: I liked it, so went for it. Change is in now!

hdgarrood:
Thanks!

Issue #2970, opened 2017-07-06
Dead code elimination should happen before codegen


paf31:
Extracted from #2963

Instead of doing DCE in `purs bundle`, we could do this instead:

- Provide a tool to extract dependencies from the functional core, and save the graph as a file
- Allow `purs build` to take that file as an input, and only generate code which is reachable from a set of entry points.

So the workflow would look like this:

- Build as usual using `purs build --dump-corefn`
- Make changes (using fast rebuilds and `--dump-corefn`)
- When ready to make a bundled build, run the new tool against the functional core, and rerun `purs build` using the output file, to produce a new set of commonjs modules with minimal definitions.
- Run the commonjs modules through psc-bundle, browserify, closure or whatever to turn it into a bundle.

I think this approach has lots of benefits:

- We can do dead code elimination in new backends
- The graph traversal code can be more reliable since it works on the more principled functional core and not JS
- We don't have to make guesses about which `module.export = ...` lines we can safely ignore.
- We can reuse work if we need to make multiple bundles.

The main downside of this approach is that we can't remove JS functions in foreign modules without parsing them, so that would have to be left to some other tool. Maybe Closure is capable of this?

coot:
I think there is no need for a new format, we could reuse `CoreFn` for the output of the _DCE tool_:
* output the modified `CoreFn` output (in a possibly new directory, other than `output`, maybe `build` by default)
* add the `purs build` command that would take that output, read into `CoreFn` and generate `CommonJS` modules (the same way modules are generated now from `CoreFn`)

If one needs to build multiple bundles manually one can run this _DCE tool_ and output the DCE-ed `CoreFn` representation to different directories and use `purs build` for each of them. The over head here is that one might end-up with bundles that contain the same modules, but we're walking into build tool responsibility.

With _webpack_'s _CommonChunksPlugin_ one will have to run the _DCE tool_ & `pure build` just once, and then let _weback_ and _CommonChunksPlugin_ to decide how to glue `CommonJS` modules into chunks.

Later we could add our own common chunks splitting: that would also use only a single pass of _DCE tool_ and then run `purs build` with a set of entry points and a flag to extract common chunks into separate module.

 I'd be happy to start with it, and have a reasonable road path :)

paf31:
I think it would be good to try out that idea in a branch and see how it goes.

coot:
[...]

PR #3039, opened 2017-08-15
WIP: dce on CoreFn

coot:
Attempt to solve #2970


coot:
This adds `purs dce`.  It reads corefn dumped to `output` and generates javascript modules in `dce-output` directory.  (There is also a switch to just generate corefn representation).

To generate javascript modules I use `moduleToJs` but it required to add annotations to the json representation of `CoreFn` (source spans and meta).

I run into problem when using `Language.PureScript.CodeGen.JS.moduleToJs`.  The `CoreFn` representation does not distinguish between non recursive bind and the recursive one (`NonRec` vs `Rec`). And it seems that `moduleToJs` shuffles the recursive ones to put them in the right order, am I right?.  This does not work for modules parsed from `CoreFn` dump, since all binders are read as non recursive. Would it be fine to distinguish `NonRec` and `Rec` in the json representation?

coot:
The issue is that declarations are encoded using an object, and `aeson` using using [hash maps](https://hackage.haskell.org/package/unordered-containers-0.2.8.0/docs/Data-HashMap-Lazy.html#t:HashMap) and they do not guarantee order.  Changing the json encoding of module declarations to a list binds, where each one is a list of three tuples `(identifier, annotation, expression)` (this way `NonRec` and `Rec` binds can be encoded using the same structure), solved the problem.

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -64,8 +64,8 @@ moduleToJSON v m = object [ T.pack "imports"   .= map (annToJSON *** moduleNameT
                           ]
 
 bindToJSON :: Bind Ann -> Value
-bindToJSON (NonRec ann n e) = object [ runIdent n .= ( annToJSON ann, exprToJSON e ) ]
-bindToJSON (Rec bs) = object $ map (\((ann, n), e) -> runIdent n .= ( annToJSON ann, exprToJSON e ) ) bs
+bindToJSON (NonRec ann n e) = toJSON [(runIdent n, annToJSON ann, exprToJSON e)]
+bindToJSON (Rec bs) = toJSON $ map (\((ann, n), e) -> (runIdent n, annToJSON ann, exprToJSON e)) bs
```
paf31: I don't understand why this needs to change. It's a recursive group, so the order should be unimportant.

coot: I just couldn't compile the `Aff` module.  It contains a group of recursive binds with `applicativeAff`, `nonCanceler` and `alwaysCanceler`.  When reading it using a hash map the order was changed to `nonCanceler`, `alwaysCanceler` followed by `applicativeAff`.  ILooking at the compiled code one can see that here order is important:
```
var applicativeAff = new Control_Applicative.Applicative(function () {
    return applyAff;
}, function (v) {
    return $foreign._pure(nonCanceler, v);
});
var nonCanceler = Data_Function["const"](Control_Applicative.pure(applicativeAff)(false));
var alwaysCanceler = Data_Function["const"](Control_Applicative.pure(applicativeAff)(true));
```

When `applicativeAff` is at the end, executing this in `node` was failing with an error indicating that `applicativeAff` passed to `Control_Applicative.pure(applicativeAff)` in `nonCanceler` is `undefined`.

coot: So far I tested this on a [purescript react isomorphic app](https://github.com/coot/purescript-isomorphic-react-example/tree/dce). At the end I run purescript dce-ed code on node and in the browser.

`pure dce` with `uglifyify` produced 705kB while `pulp build -O` with `uglifyify` 685kB.
There is still room for improvement though.  Now I am only removing top level declarations and not touching foreign modules at all.

coot:
With webpack I could reduce the bundle down to 440kB and with a naive removing of exports from foreign modules additional 20kB - so I guess it makes sense to do that in a safer way.

paf31:
Thanks for taking the time to work on this!

This is obviously a very big change, and could eventually replace `psc-bundle`, which is pretty stable and well-tested at this point. I'd like to suggest this:

- merge the corefn json parser into the compiler
- make a separate repo to test this approach, and give it some real time to iron out any bugs.

Once it's thoroughly tested on lots of code (maybe after 1.0 even), we could deprecate `purs bundle`, with a plan to remove it in a future breaking release, and a plan to add this tool in a non-breaking release (could be 0.12.x or 1.x for example).

What do you think?

paf31:
Also, now would be the time to break the corefn format, if we're going to do it, since we can do it in 0.12.0.

coot:
I agree, that indeed needs some time to mature   and grow its test suite and be tested on real projects.

I will prepare PR for the json representation.
Is its format documented somewhere?

Cmdv:
I'll be glad to test it out, if any help is needed in that department 😃

coot:
@paf31 maybe when you will announcing the next version of purescript you could include a line about new experimental (external) tool for tree shaking?

coot:
Here it is [zephyr](https://github.com/coot/zephyr).

paf31:
> maybe when you will announcing the next version of purescript you could include a line about new experimental (external) tool for tree shaking?

Yes, I think we should definitely do that. Thanks!

joneshf:
> Also, now would be the time to break the corefn format, if we're going to do it, since we can do it in 0.12.0.

Can we do this? Corefn is kind of a pain to work with, enough so that I've given up on writing a JSONSchema for it.

paf31:
Yes. If you have experience working with it, it would be good to get suggestions for a better format.

joneshf:
Here are some things we found hard, in no particular order:

## Very few things are named.

It can be kind of hard to know which parts are which when you're deep in the nesting. It required me to explain what each piece of the JSON was in order for others to understand what they were looking at.
For instance, case expressions look very convoluted unless you understand all the possibilities that PS source has for case expressions:
```json
[
    "Case",
    [
        [
            "Var",
            "v"
        ]
    ],
    [
        [
            [
                [
                    "LiteralBinder",
                    [
                        "IntLiteral",
                        1
                    ]
                ]
            ],
            [
                "Literal",
                [
                    "IntLiteral",
                    1
                ]
            ]
        ],
        [
            [
                "NullBinder"
            ],
            [
                "Literal",
                [
                    "IntLiteral",
                    0
                ]
            ]
        ]
    ]
]
```

## The S-expression encoding feels harder to work with than an object encoding would.

This ties in with the naming concern. I see the appeal for the S-expression encoding, but I think it would be better if we just had a way to dump an actual S-expression and left the JSON dump as an object.
```json
[
    "Abs",
    "dict",
    [
        "Accessor",
        "foo",
        [
            "Var",
            "dict"
        ]
    ]
]
```
vs
```json
{
    "type": "Abs",
    "argument": "dict",
    "body": {
        "type": "Accessor",
        "key": "foo",
        "object": {
            "type": "Var",
            "name": "dict"
        }
    }
}
```

I have no strong feelings on what the names should be, but they should exist.

## The value of each `decls` has to be parsed before you know how to generate them.

In JS this is not much of a problem, because a function can be used for both a lambda abstraction and an object constructor. But in languages that don't conflate the two ideas, it requires more work.
```json
"decls": [
    {
        "Foo": []
    },
    {
        "Bar": []
    }
]
```
You can't know what `Foo` or `Bar` are supposed to be until you parse their arrays. One could be a type class, the other a value constructor, maybe other things.

## Qualified names are serialized as a string interspersed with `.` between module names.

This is another thing that is fine in JS because you can dot your way wherever you need to go. But, it's harder in other languages that don't use `.` for everything. If module resolution is done with `::` you have to then parse the string into its module pieces and value pieces (or whatever you're trying to do).

coot:
I agree with @joneshf, having more verbose output would make it much easier to work with the json files.

coot:
@paf31 if you agree with @joneshf proposal I can add the changes to #3049.

paf31:
@joneshf Thank you for the detailed notes! I can't disagree with anything you've said. The S-expression syntax was something of an experiment to begin with, and maybe it hasn't worked out. I would be fine with breaking the format to use the suggested `type` fields for 0.12.0.

@rightfold Any thoughts?

rightfold:
I agree. Especially problems with ambiguous names

coot:
I think we should also change `recordToJson` to always encode records as lists of pairs, rather than try to encode it as object.  This will simplify parsers, but also avoid bugs in them due to unexpected format that is present only if encoding of a label fails.

PR #3040, opened 2017-08-16
[psc-ide] Make `type` command's `filters` param optional to align with doc

b123400:
[The doc](https://github.com/purescript/purescript/blob/master/psc-ide/PROTOCOL.md#type) says type's `filters` param is optional but it is currently not.


kRITZCREEK:
Change looks great, thanks! If you could just add yourself to the CONTRIBUTORS.md file we're good to go.

b123400:
Updated~

kRITZCREEK:
Thanks a lot!

PR #3041, opened 2017-08-16
bump glob dependency

mjhoy:
Allows glob 0.8. The upper bound caused installing purescript@master to fail in nix.


paf31:
The changelog on Glob looks harmless, so I'll merge this now.

PR #3045, opened 2017-08-17
[purs ide] Adds an `actualFile` parameter to the rebuild command

### LINKED ISSUE #3033: [purs ide] Editor mode picks up "wrong" source spans
It's a common practice (Emacs, Atom & VSCode definitely do this) to save the currently edited file to a tmp location and run the rebuild command on that file, to not have to save the users file constantly. In editor mode however, this leads to the server picking up the tmp file location for goto definition, which is not good (Leads to lost changes).

Not entirely sure how to handle this. We could try to look up the previous file location from our server state (This will be a little complicated, but not impossible) or allow another parameter to the rebuild command which overrides the filepath in the source spans of the parsed module. The last option is to just save the original file on the editor side and rebuild on top of that.

Any input? \cc @nwolverson @coot

kRITZCREEK:
This allows us to specify the actual path of the source file (important for
declaration location information as well as correct parse error locations), in
case we are using a temp file for rebuilding.

Fixes #3033

### DIFF
--- a/psc-ide/PROTOCOL.md
+++ b/psc-ide/PROTOCOL.md
@@ -326,12 +326,16 @@ identifiers.
 
 Arguments:
   - `file :: String` the path to the module to rebuild
+  - `actualFile :: Maybe String` Specifies the path to be used for location
+    information and parse errors. This is useful in case a temp file is used as
+    the source for a rebuild.
 
 ```json
 {
   "command": "rebuild",
   "params": {
     "file": "/path/to/file.purs"
+    "actualFile": "/path/to/actualFile.purs"
   }
 }
 ```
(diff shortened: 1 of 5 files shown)


kRITZCREEK:
CI timeouts.... This doesn't sound good:

```
Setting up build cache

/Users/travis/.travis/job_stages: line 484: shell_session_update: command not found
```

hdgarrood:
I've seen that a few times but it usually doesn't seem to affect the build status :/

kRITZCREEK:
Huh, well the tests pass, so I won't treat the CI failure as blocking. It would be nice to get this in before releasing 0.12, as editor-mode is kind of useless without it.

hdgarrood:
In fact I don't think it is affecting the build status in this case, it's just that Travis is taking ages to upload the build cache at the end of the build, and that time is included in the time limit. You could try decreasing the extra timeout I imposed in the Travis CI script, to give it more time to upload the build cache, or you could just push a new commit and let it try again. (Yes this is very much not ideal but it was the best I could do.)

PR #3048, opened 2017-08-19
Introduce data type for value declarations

kRITZCREEK:
And another one for #3023! I'd like some Feedback and Discussion on this one.

```haskell
-- | A value declaration assigns a name and potential binders, to an expression (or multiple guarded expressions).
--
-- @double x = x + x@
--
-- In this example @double@ is the identifier, @x@ is a binder and @x + x@ is the expression.
data ValueDeclarationData a = ValueDeclarationData
  { valdeclSourceAnn :: !SourceAnn
  , valdeclIdent :: !Ident
  -- ^ The declared value's name
  , valdeclName :: !NameKind
  -- ^ Whether or not this value is exported/visible
  , valdeclBinders :: ![Binder]
  , valdeclExpression :: !a
  } deriving (Show)
```

I parameterized the data type on the type of it's contained Expressions, because from reading the code it seems like during case desugaring all ValueDeclarations go from `ValueDeclarationData [GuardedExpr]` to `ValueDeclarationData Expr`.

In particular, this means that we can turn `BindingGroupDeclaration (NEL.NonEmpty ((SourceAnn, Ident), NameKind, Expr))` into `BindingGroupDeclaration (NEL.NonEmpty (ValueDeclarationData Expr))`

Additionally there's partial pattern matches on `[MkUnguarded expr]` littered through the code, which could also be addressed this way (The whole CaseDeclarations desugaring is a partial desaster imo). I think we should introduce another Declaration constructor `GuardedValueDeclaration (ValueDeclaration [GuardedExpr])` and have our ValueDeclaration constructor only contain a single expression.

The pattern matches and constructions of ValueDeclarations are getting a little noisy in this PR, and I think we can mitigate that by introducing a pattern synonym. I didn't want to do that yet, because implementing my previous suggestions would mean I'd have to change all that again anyway.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -443,10 +443,6 @@ data ValueDeclarationData a = ValueDeclarationData
   , valdeclExpression :: !a
   } deriving (Show)
 
-  -- |
-  -- A value declaration (name, top-level binders, optional guard, value)
-  --
-
 overValueDeclaration :: (ValueDeclarationData [GuardedExpr] -> ValueDeclarationData [GuardedExpr]) -> Declaration -> Declaration
 overValueDeclaration f d = maybe d (ValueDeclaration . f) (getValueDeclaration d)
 
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -441,7 +441,7 @@ data ValueDeclarationData a = ValueDeclarationData
   -- ^ Whether or not this value is exported/visible
   , valdeclBinders :: ![Binder]
   , valdeclExpression :: !a
-  } deriving (Show)
+  } deriving (Show, Functor, Foldable, Traversable)
 
 overValueDeclaration :: (ValueDeclarationData [GuardedExpr] -> ValueDeclarationData [GuardedExpr]) -> Declaration -> Declaration
 overValueDeclaration f d = maybe d (ValueDeclaration . f) (getValueDeclaration d)
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -428,6 +428,32 @@ getTypeDeclaration _ = Nothing
 unwrapTypeDeclaration :: TypeDeclarationData -> (Ident, Type)
 unwrapTypeDeclaration td = (tydeclIdent td, tydeclType td)
 
+-- | A value declaration assigns a name and potential binders, to an expression (or multiple guarded expressions).
+--
+-- @double x = x + x@
+--
+-- In this example @double@ is the identifier, @x@ is a binder and @x + x@ is the expression.
+data ValueDeclarationData a = ValueDeclarationData
+  { valdeclSourceAnn :: !SourceAnn
+  , valdeclIdent :: !Ident
+  -- ^ The declared value's name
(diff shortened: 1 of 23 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -451,7 +473,7 @@ data Declaration
   -- |
   -- A value declaration (name, top-level binders, optional guard, value)
   --
-  | ValueDeclaration SourceAnn Ident NameKind [Binder] [GuardedExpr]
+  | ValueDeclaration {-# UNPACK #-} !(ValueDeclarationData [GuardedExpr])
```
paf31: I would avoid adding UNPACK pragmas and bang annotations until profiling shows it to be necessary. It's not always an optimization (although, I'll admit, it is usually in these sorts of cases).

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -428,6 +428,28 @@ getTypeDeclaration _ = Nothing
...
+-- In this example @double@ is the identifier, @x@ is a binder and @x + x@ is the expression.
+data ValueDeclarationData a = ValueDeclarationData
+  { valdeclSourceAnn :: !SourceAnn
+  , valdeclIdent :: !Ident
+  -- ^ The declared value's name
+  , valdeclName :: !NameKind
+  -- ^ Whether or not this value is exported/visible
+  , valdeclBinders :: ![Binder]
+  , valdeclExpression :: !a
+  } deriving (Show)
```
paf31: Can you please derive functor, foldable and traversable here?

src/Language/PureScript/AST/Traversals.hs
```diff
@@ -53,7 +53,8 @@ everywhereOnValues f g h = (f', g', h')
   where
   f' :: Declaration -> Declaration
   f' (DataBindingGroupDeclaration ds) = f (DataBindingGroupDeclaration (fmap f' ds))
-  f' (ValueDeclaration sa name nameKind bs val) = f (ValueDeclaration sa name nameKind (fmap h' bs) (fmap (mapGuardedExpr handleGuard g') val))
+  f' (ValueDeclaration (ValueDeclarationData sa name nameKind bs val)) =
```
paf31: These can use the derived instances above.

src/Language/PureScript/AST/Declarations.hs
kRITZCREEK: The strictness annotations on record fields primarily server to not accidentally initialize a partial record. I put the `UNPACK` here to make sure that I didn't make things worse (this should basically be the same memory layout as before)

src/Language/PureScript/AST/Traversals.hs
kRITZCREEK: Maybe I'm not seeing it, but 

`f (ValueDeclaration (map (mapGuardedExpr handleGuard g') <$> ValueDeclarationData sa name nameKind (fmap h' bs) val))`

doesn't seem better than:

`f (ValueDeclaration (ValueDeclarationData sa name nameKind (fmap h' bs) (fmap (mapGuardedExpr handleGuard g') val)))`

What's the win? We still need to destructure to get at the binders.

kRITZCREEK:
Thanks for reviewing! I'll merge this, if there's anything you'd still like to be changed I can just make a new PR.

PR #3049, opened 2017-08-19
Corefn Json representation

coot:
Modify corefn json representation in a way to be able reproduce `CoreFn` from it.


coot:
I think this is ready for a review.

paf31:
I'm wondering if we should add a `purs codegen` command, which runs the JS codegen on a corefn file. That way, we could implement DCE and other optimizer passes outside the compiler.

Also, @joneshf, do you have suggestions for the corefn JSON format, since we should break them here (or in a separate PR, but before 0.12)?

joneshf:
I'd suggest a separate PR. I don't really know what's happening here and I don't want to do a drive-by on this. I put some comments in the other PR

coot:
This adds
* annotations to curefn representation
* represents binds as lists rather than objects so the order of recursive binds is preserved
* adds `moduleFromJSON` that reconstructs `CoreFn` from its json representation (up to the type of foreign declaration.

This allows to pick up CoreFn from its json rep and turn it into javascript code.

coot:
`purs codegen` would be really useful for writting tools that work with `CoreFn` in purescript where you just `purs compile --dump-corefn; run-some-tool; purs codegen`.

coot:
I update the `CoreFn` json representation, taking into account comments in #3039 
[Here](https://pastebin.com/aRNhTtZa) is a dump of `Data.Newtype`.

coot:
I run the [zephyr's](https://github.com/coot/zephyr/blob/master/test/Main.hs#L44) test suite which compiles quite a few libraries. So the new json representation seems to work quite fine.

Maybe some things could be named better or represented in a better way.
@joneshf, @rightfold any comments?

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -22,26 +23,61 @@ import Language.PureScript.Names
...
 literalToJSON :: (a -> Value) -> Literal a -> Value
-literalToJSON _ (NumericLiteral (Left n)) = toJSON ("IntLiteral", n)
-literalToJSON _ (NumericLiteral (Right n)) = toJSON ("NumberLiteral", n)
-literalToJSON _ (StringLiteral s) = toJSON ("StringLiteral", s)
-literalToJSON _ (CharLiteral c) = toJSON ("CharLiteral", c)
-literalToJSON _ (BooleanLiteral b) = toJSON ("BooleanLiteral", b)
-literalToJSON t (ArrayLiteral xs) = toJSON ("ArrayLiteral", map t xs)
-literalToJSON t (ObjectLiteral xs) = toJSON ("ObjectLiteral", recordToJSON t xs)
+literalToJSON _ (NumericLiteral (Left n))
+  = object
```
joneshf: I don't think you need to go through and replace all of this now, but there's an alternative approach which skips over creating the intermediate data structure: https://hackage.haskell.org/package/aeson-1.2.1.0/docs/Data-Aeson.html#g:7

You'd change expressions from

```haskell
object
  [ T.pack "foo" .= foo
  , T.pack "bar" .= bar
  ]
```

to expressions like

```haskell
pairs
  ( T.pack "foo" .= foo
  <> T.pack "bar" .= bar
  )
```

It's supposedly faster and uses less memory. So given that the json blobs have gotten bigger, it might be a change that makes sense.

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -101,6 +101,7 @@ moduleToJSON v m = object [ T.pack "moduleName" .= moduleNameToJSON (moduleName
                           , T.pack "foreign"    .= map (identToJSON . fst) (moduleForeign m)
                           , T.pack "decls"      .= map bindToJSON (moduleDecls m)
                           , T.pack "builtWith"  .= toJSON (showVersion v)
+                          , T.pack "comments"   .= map toJSON (moduleComments m)
```
joneshf: :tada:

coot:
Nice, let me check, maybe the new encoding preserves the order of keys.

src/Language/PureScript/CoreFn/ToJSON.hs
coot: @joneshf I think we'd need to add `ToJSON` instances to make it work (since the right hand side of `.=` expects an instance of `ToJSON`.  This is a bit more work so I'd like to leave it for later.

rightfold:
I am very happy with this. 👍

coot:
I have more, experimenting in zephyr.  I'd like to add type annotations to corefn, would that be fine? (but this can be in another PR).

rightfold:
Yeah, that's fine, if they're available. I'd like to use them in my Go backend after finishing the purescript-purescript-compiler-backend-utils library. But I don't need them per se, so I'm fine with them not being there. But maybe other people find them more important.

paf31:
I'm a little worried about the size of the generated JSON. Part of the reason for the unconventional format before was that I wanted to reduce the size and keep the output somewhat human-readable. Could you please see how many lines are generated for a bigger file? Maybe something like Thermite's main module? 4000+ lines for `Data.Newtype` seems quite a lot.

coot:
This PR version:
```
241K Sep 10 01:05 output/Thermite/corefn.json
```
Also including types:
```
1.4M Sep 10 01:00 corefn.json
```
For comparison, purs 11.6
```
22K Sep 10 01:11 corefn.json
```

The difference between 22K and 241K is big, but from the other side the representation is much more readable.  

With types it's indeed quite a lot, but without types some optimizations will be much harder to do (dce of instance dictionaries). I could easily navigate when the types where included with vim % command (jump to the closing brace) - and I did quite a lot of it recently working on zephyr.

natefaubion:
How much of the output is position annotations? :P

coot:
Removing source annotations from this 241KB leaves 102KB.

joneshf:
Looking at some different output options:

```
➜  purescript git:(master) ✗ l Data.Newtype.*                   
-rw-r--r-- 1 joneshf users 293K Sep  9 16:39 Data.Newtype.aaaa-original.json
-rw-r--r-- 1 joneshf users 292K Sep  9 16:44 Data.Newtype.argument-to-arg.json
-rw-r--r-- 1 joneshf users 194K Sep  9 16:43 Data.Newtype.four-spaces-to-two-spaces.json
-rw-r--r-- 1 joneshf users  81K Sep  9 16:54 Data.Newtype.remove-all-whitespace.json
-rw-r--r-- 1 joneshf users 291K Sep  9 16:47 Data.Newtype.sourceSpan-to-span.json
-rw-r--r-- 1 joneshf users 144K Sep  9 16:46 Data.Newtype.spaces-to-tab.json
-rw-r--r-- 1 joneshf users 292K Sep  9 16:44 Data.Newtype.value-to-val.json
```

The vast majority of this is whitespace, and the drop from four spaces to two is pretty significant while still keeping the file visually appealing. Making the JSON blob into one line by removing all whitespace and newlines is also pretty huge.

Are either of these options acceptable?

garyb:
I think using a compressed format (in the sense of minimal whitespace) is a good idea, there are a million JSON pretty printers out there to undo it if someone wants to read it manually.

coot:
As of 11.6.0 corefn dumps do not contain spaces, so the numbers I posted will not get smaller.

coot:
For comparison how much compression can help:
```
-rw-r--r-- 1 marcin marcin   22062 Sep 10 01:11 corefn-11.6.json
-rw-r--r-- 1 marcin marcin    2901 Sep 10 01:11 corefn-11.6.json.gz
-rw-r--r-- 1 marcin marcin 1394677 Sep 10 11:34 corefn-29ef583.json    # with types
-rw-r--r-- 1 marcin marcin   38179 Sep 10 11:34 corefn-29ef583.json.gz # with types gzipped
-rw-r--r-- 1 marcin marcin  245913 Sep 10 11:23 corefn-a986e4f.json    # this PR
-rw-r--r-- 1 marcin marcin   11894 Sep 10 11:23 corefn-a986e4f.json.gz # this PR gzipped
```
Using gzip might not be that a crazy idea, since mostly this will be used by tools.  I am not sure what's the state of support for gzip in other editors, in vim I can just open them  the same way I open json files: `vim +JsBeautify corefn.json.gz` and I see a pretty printed json.

rightfold:
gzip is a pain in Windows (it doesn't ship with decoding tools) but going for the lowest common denominator is also a pain so I don't see a reason not to use gzip

joneshf:
If the option is Gzip, I'd rather have the old representation back.

kRITZCREEK:
How much can we reduce the size if we only stored the SourceSpan, instead of repeating the filepath over and over?

rightfold:
The S-expression-like representation was a horrible mistake. It cannot ever be extended with more annotations.

joneshf:
Speaking of the filepath, it contains the absolute path of the machine it was compiled on. Is that necessary, or can we use the relative path to the top level of the repo? It drops about 20-25% of the file size on my computer.

```
➜  purescript-thermite git:(master) cat output/Thermite/corefn.json | wc --bytes
248201
➜  purescript-thermite git:(master) sed 's@/home/joneshf/programming/paf31/purescript-thermite/@@g' output/Thermite/corefn.json | wc --bytes
188713
```

joneshf:
Actually, yes, like @kRITZCREEK stated, we don't even need to repeat anything since we only have one module per file now. If the file path is necessary at all, we should be able to move it up to the top level and have it once.

paf31:
@coot Could you please try moving the source file info to the top level?

paf31:
Was it not possible to just change the JSON renderer for `SourcePos`?

coot:
I think it was, if you prefer that I'll update the PR.  The only reason for having it exposed on the corefn module, is for tooling that might use it - if it's written in haskell and one would liket use the comipler `moduleFromJSON`. There is no tool for the moment that needs that though and it's very easy to get access to the path anyway - so maybe that's not a good enough reason to keep this.

metaleap:
Can't wait for this to land, will it be in 0.12? (The old `--dump-imperative-core` would be great too but seems it's been postponed) Not sure why anyone cares about the file size of an intermediate format to be further machine-processed. @rightfold where's your "Go backend" to be found? I was just about to embark on exactly such hackery, but if it's FOSS I'd rather contribute to yours

rightfold:
I didn’t start working on the go backend, I want to wait for the new corefn dump to be merged and released. 😆

It will probably be similar to my Python backend https://github.com/rightfold/pegasus

Sent from my iPhone.

metaleap:
The latest commit in this PR from 2 weeks ago doesn't build at my end, anything I'm missing? Well it passed your CI so the prob must be at my end, but I've applied all changes

    ~/c/hs/MYpurescript/src/Language/PureScript/CoreFn/Desugar.hs:66:17: error:
        • The constructor ‘A.ValueDeclaration’ should have 1 argument, but has been given 5
        • In the pattern:
            A.ValueDeclaration (ss, com) name _ _ [A.MkUnguarded e]
          In an equation for ‘declToCoreFn’:
              declToCoreFn
                (A.ValueDeclaration (ss, com) name _ _ [A.MkUnguarded e])
                = [NonRec (ssA ss) name (exprToCoreFn ss com Nothing e)]
          In an equation for ‘moduleToCoreFn’:
              moduleToCoreFn env (A.Module modSS coms mn decls (Just exps))
                = let
                    imports
                      = mapMaybe importToCoreFn decls ++ fmap ... (findQualModules decls)
                    imports' = dedupeImports imports
                    ....
                  in Module coms mn (spanName modSS) imports' exps' externs decls'
                where
                    dedupeImports :: [(Ann, ModuleName)] -> [(Ann, ModuleName)]
                    dedupeImports
                      = fmap swap . M.toList . M.fromListWith const . fmap swap
                    ssA :: SourceSpan -> Ann
                    ssA ss = (ss, [], Nothing, Nothing)
                    ....
       |
    66 |   declToCoreFn (A.ValueDeclaration (ss, com) name _ _ [A.MkUnguarded e]) =
       |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    ~/c/hs/MYpurescript/src/Language/PureScript/CoreFn/Desugar.hs:195:12: error:
        • The constructor ‘A.TypeInstanceDeclaration’ should have 8 arguments, but has been given 6
        • In the pattern: A.TypeInstanceDeclaration _ _ _ q _ _
          In an equation for ‘fqDecls’:
              fqDecls (A.TypeInstanceDeclaration _ _ _ q _ _) = getQual' q
          In an equation for ‘findQualModules’:
              findQualModules decls
                = let
                    (f, _, _, _, _)
                      = everythingOnValues
                          (++) fqDecls fqValues fqBinders (const ...) (const ...)
                  in f `concatMap` decls
                where
                    fqDecls :: A.Declaration -> [ModuleName]
                    fqDecls (A.TypeInstanceDeclaration _ _ _ q _ _) = getQual' q
                    fqDecls (A.ValueFixityDeclaration _ _ q _) = getQual' q
                    fqDecls (A.TypeFixityDeclaration _ _ q _) = getQual' q
                    fqDecls _ = []
                    fqValues :: A.Expr -> [ModuleName]
                    fqValues (A.Var q) = getQual' q
                    fqValues (A.Constructor q) = getQual' q
                    fqValues (A.TypeClassDictionaryConstructorApp c _) = getQual' c
                    fqValues _ = []
                    ....
        |
    195 |   fqDecls (A.TypeInstanceDeclaration _ _ _ q _ _) = getQual' q
        |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Just tried to use it in my local fork as this PR has been just sitting here for some 2-5 weeks depending on how one counts  😀

coot:
@metaleap I merged master.  It should be compile now for you.  Try this branch rather than applying a patch to master.

metaleap:
That was the ticket for me to build at my end, thanks, now I can begin toying with this =)

src/Language/PureScript/CoreFn/Module.hs
```diff
@@ -10,13 +10,18 @@ import Language.PureScript.Types
...
 --
-data Module a = Module
+data ModuleT t a = Module
   { moduleComments :: [Comment]
   , moduleName :: ModuleName
+  , modulePath :: FilePath
   , moduleImports :: [(a, ModuleName)]
   , moduleExports :: [Ident]
-  , moduleForeign :: [ForeignDecl]
+  , moduleForeign :: [ForeignDeclT t]
```
paf31: Could you please add a comment explaining why this type argument is needed?

paf31:
@coot @rightfold @joneshf @kRITZCREEK @natefaubion Any more comments or shall I merge this now?

src/Language/PureScript/CoreFn/FromJSON.hs
```diff
@@ -0,0 +1,290 @@
...
+import           Language.PureScript.Names
+import           Language.PureScript.PSString (PSString)
+
+constructorTypeFromJSON :: Value -> Parser ConstructorType
+constructorTypeFromJSON v = do
+  t <- parseJSON v
+  case t of
+    "ProductType" -> return ProductType
+    "SumType"     -> return SumType
+    _             -> fail ("not regonized ConstructorType: " ++ T.unpack t)
```
kRITZCREEK: small typo here "regonized"

coot:
My pleasure!

src/Language/PureScript/CoreFn/FromJSON.hs
```diff
@@ -0,0 +1,290 @@
...
+      return (ann, mn))
+
+bindFromJSON :: FilePath -> Value -> Parser (Bind Ann)
+bindFromJSON modulePath = withObject "Bind" bindFromObj
+  where
+  bindFromObj :: Object -> Parser (Bind Ann)
+  bindFromObj o = do
+    type_ <- o .: "bindType" :: Parser Text
+    case type_ of
+      "NonRec"  -> (uncurry . uncurry $ NonRec) <$> bindFromObj' o
```
kRITZCREEK: `(uncurry . uncurry) NonRec`

Would be a little less noisy

src/Language/PureScript/CoreFn/ToJSON.hs
kRITZCREEK: You could just encode to `Value` on the right-hand-side which picks the trivial `ToJSON` instance.

coot: @kRITZCREEK you're right, but let's do that in another PR.

src/Language/PureScript/CoreFn/Module.hs
```diff
@@ -10,6 +10,9 @@ import Language.PureScript.Types
 -- |
 -- The CoreFn module representation
 --
+-- The json CoreFn representation does not contain type information.  When
+-- parsing it one gets back `ModuleT () Ann` rathern than `ModuleT Type Ann`,
```
LiamGoodacre: Typo: `rathern`

src/Language/PureScript/CoreFn/Module.hs
```diff
@@ -10,6 +10,9 @@ import Language.PureScript.Types
 -- |
 -- The CoreFn module representation
 --
+-- The json CoreFn representation does not contain type information.  When
```
paf31: It looks like we don't even use the types in `moduleForeign`. If that's the case, I'd actually rather just get rid of the types completely and lose the type argument.

metaleap: As long as they're kept around in externs, don't drop them there! (Just chiming in as I'm wrestling with an alternative backend *not* targeting a dynamic/untyped language like JS/Py/Erl etc ... from what I can tell I'll have to reconstruct a lot from literals & externs & such as I'm working off CoreImp.) Just a quick reminder to not forget the awesome use-case for alternative backends, some of which may want to keep around as much type info as possible, to minimize usage of boxed-variants (ie what is `object` in Java or `interface{}` in Go, etc) being passed around.

In my fork I experimentally added all `Ann`s to your `--dump-corefn` module, to see what I can get beyond what's in externs. Not *that* much more but last I checked, it did have a bit more detail. A bit exploratory all right now, to write a backend based on --dump-core*foo* outputs rather than right in Haskell in the fork, but it's fun for now. All right I'm done blogging & derailing the thread 😀

joneshf:
This is great! I'm with it.

src/Language/PureScript/CoreFn/Module.hs
paf31: As far as I can tell, we don't have any type information in the generated JSON right now.

paf31:
Let's merge this and fix up any remaining issues in `master` if we find any. Thanks!

paulyoung:
https://github.com/purescript/purescript/pull/3049#issuecomment-323607542

> I'm wondering if we should add a `purs codegen` command, which runs the JS codegen on a corefn file.

I was just wondering if it was possible to use `--dump-corefn` without producing JavaScript.

coot:
I think that `--dump-corefn` shouldn't generate javascript at all.

PR #3054, opened 2017-08-27
Solve ConsSymbol

kcsongor:
This adds and solves a new 'magic' constraint,
`Type.Data.Symbol.ConsSymbol`, with the following functional
dependencies:

```
class ConsSymbol (head :: Symbol)
                 (tail :: Symbol)
                 (sym :: Symbol) | sym -> head tail, head tail -> sym
```

* The `head tail -> sym` direction works like `AppendSymbol`, but only
resolves when `head` is a singleton symbol.

* The `sym -> head tail` direction deconstructs the symbol into its
first character and the remaining string.

The relevant PR for the typelevel-prelude:
https://github.com/purescript/purescript-typelevel-prelude/pull/24

I put up an example that uses this class at
https://github.com/kcsongor/purescript-record-format

### DIFF
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -14,7 +14,7 @@ module Language.PureScript.TypeChecker.Entailment
 import Prelude.Compat
 import Protolude (ordNub)
 
-import Control.Arrow ((***), second)
+import Control.Arrow (second)
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State
 import Control.Monad.Supply.Class (MonadSupply(..))
@@ -168,7 +168,7 @@ entails SolverOptions{..} constraint context hints =
       let args = [arg0, arg1, TypeLevelString (lhs <> rhs)]
       in [TypeClassDictionaryInScope AppendSymbolInstance [] C.AppendSymbol args Nothing]
     forClassName _ C.ConsSymbol [arg0, arg1, arg2]
-      | Just (arg0', arg1', arg2') <- (consSymbol arg0 arg1 arg2) =
+      | Just (arg0', arg1', arg2') <- consSymbol arg0 arg1 arg2 =
       let args = [arg0', arg1', arg2']
       in [TypeClassDictionaryInScope ConsSymbolInstance [] C.ConsSymbol args Nothing]
     forClassName _ C.Union [l, r, u]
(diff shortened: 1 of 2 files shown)


kcsongor:
I'd be happy to add some examples, but I'm not too sure how I would depend on the new type-level prelude at this point

kcsongor:
Here's another example using it
https://github.com/kcsongor/purescript-safe-printf

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -165,6 +167,10 @@ entails SolverOptions{..} constraint context hints =
     forClassName _ C.AppendSymbol [arg0@(TypeLevelString lhs), arg1@(TypeLevelString rhs), _] =
       let args = [arg0, arg1, TypeLevelString (lhs <> rhs)]
       in [TypeClassDictionaryInScope AppendSymbolInstance [] C.AppendSymbol args Nothing]
+    forClassName _ C.ConsSymbol [arg0, arg1, arg2]
+      | Just (arg0', arg1', arg2') <- (consSymbol arg0 arg1 arg2) =
```
paf31: The parens are redundant here.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -353,6 +361,19 @@ entails SolverOptions{..} constraint context hints =
         subclassDictionaryValue dict className index =
           App (Accessor (mkString (superclassName className index)) dict) valUndefined
 
+    consSymbol :: Type -> Type -> Type -> Maybe (Type, Type, Type)
+    consSymbol _ _ arg@(TypeLevelString s) = do
+      (h, t) <- (mkTLString . T.singleton *** mkTLString) <$> (T.uncons =<< decodeString s)
```
paf31: This is a little hard to read. What about this?
```haskell
do (h, t) <- T.uncons =<< decodeString s
   pure (mkTLString (T.singleton h), mkTLString t, arg)
```

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -353,6 +361,19 @@ entails SolverOptions{..} constraint context hints =
...
+      pure (h, t, arg)
+      where mkTLString = TypeLevelString . mkString
+    consSymbol arg1@(TypeLevelString h) arg2@(TypeLevelString t) _ = do
+      h' <- decodeString h
+      t' <- decodeString t
+      if (T.length h' == 1)
```
paf31: Could use `guard` here.

kcsongor:
Thanks for the comments, I've fixed them!

kcsongor:
On a related note, I think it would be better to have a `Char` kind with type-level character literals, that way, it would become clearer what's happening here (and in type-level parsers too). What are your thoughts on this?

paf31:
Thanks!

We have some time before 0.12.0 is properly released to discuss the `Char` idea, so let's do that.

paf31:
@kcsongor Would you mind adding yourself to `CONTRIBUTORS.md` as well?

kcsongor:
Great, that sounds good

PR #3057, opened 2017-08-28
Use dot accessors/setters in purs bundle if possible

houli:
Output code from purs bundle includes a fair amount of redundant string accesses to objects when getting and setting properties. This is seen quite often in output code such as `exports["main"] = main` when `exports.main = main` would have worked just as well because `main` is a valid JavaScript identifier. This change will save 3 bytes everywhere it can be applied which is a not a significant saving in itself but will appear quite often in a large codebase. It also can be argued that accessing using the dot notation is more idiomatic output JavaScript than using string literals.

As an example the hello world pulp project compiled to this before at 999 bytes:
```
// Generated by psc-bundle 0.11.6
var PS = {};
(function(exports) {
    "use strict";

  exports.log = function (s) {
    return function () {
      console.log(s);
      return {};
    };
  };
})(PS["Control.Monad.Eff.Console"] = PS["Control.Monad.Eff.Console"] || {});
(function(exports) {
  // Generated by purs version 0.11.6
  "use strict";
  var $foreign = PS["Control.Monad.Eff.Console"];
  var Control_Monad_Eff = PS["Control.Monad.Eff"];
  var Data_Show = PS["Data.Show"];
  var Data_Unit = PS["Data.Unit"];
  exports["log"] = $foreign.log;
})(PS["Control.Monad.Eff.Console"] = PS["Control.Monad.Eff.Console"] || {});
(function(exports) {
  // Generated by purs version 0.11.6
  "use strict";
  var Control_Monad_Eff = PS["Control.Monad.Eff"];
  var Control_Monad_Eff_Console = PS["Control.Monad.Eff.Console"];
  var Prelude = PS["Prelude"];        
  var main = Control_Monad_Eff_Console.log("Hello sailor!");
  exports["main"] = main;
})(PS["Main"] = PS["Main"] || {});
PS["Main"].main();
```
And after at 985 bytes:
```
// Generated by purs bundle 0.11.6
var PS = {};
(function(exports) {
    "use strict";

  exports.log = function (s) {
    return function () {
      console.log(s);
      return {};
    };
  };
})(PS["Control.Monad.Eff.Console"] = PS["Control.Monad.Eff.Console"] || {});
(function(exports) {
  // Generated by purs version 0.11.6
  "use strict";
  var $foreign = PS["Control.Monad.Eff.Console"];
  var Control_Monad_Eff = PS["Control.Monad.Eff"];
  var Data_Show = PS["Data.Show"];
  var Data_Unit = PS["Data.Unit"];
  exports.log = $foreign.log;
})(PS["Control.Monad.Eff.Console"] = PS["Control.Monad.Eff.Console"] || {});
(function(exports) {
  // Generated by purs version 0.11.6
  "use strict";
  var Control_Monad_Eff = PS["Control.Monad.Eff"];
  var Control_Monad_Eff_Console = PS["Control.Monad.Eff.Console"];
  var Prelude = PS.Prelude;           
  var main = Control_Monad_Eff_Console.log("Hello sailor!");
  exports.main = main;
})(PS.Main = PS.Main || {});
PS.Main.main();
```
On a larger example project this change reduced purs bundle output size from 399253 bytes to 397378 bytes.

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -512,7 +512,7 @@ codeGen optionsMainModule optionsNamespace ms outFileOpt = (fmap sourceMapping o
   rendered = renderToString (JSAstProgram (prelude : concatMap fst modulesJS ++ maybe [] runMain optionsMainModule) JSNoAnnot)
 
   isValidJavaScriptIdentifier :: String -> Bool
-  isValidJavaScriptIdentifier str = case parse str mempty of
+  isValidJavaScriptIdentifier string = case parse string mempty of
     Right (JSAstProgram [JSExpressionStatement (JSIdentifier _ _) _] _)  -> True
     Right _ -> False
     Left _ -> False
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -511,6 +511,12 @@ codeGen optionsMainModule optionsNamespace ms outFileOpt = (fmap sourceMapping o
   where
   rendered = renderToString (JSAstProgram (prelude : concatMap fst modulesJS ++ maybe [] runMain optionsMainModule) JSNoAnnot)
 
+  isValidJavaScriptIdentifier :: String -> Bool
+  isValidJavaScriptIdentifier str = case parse str mempty of
+    Right (JSAstProgram [JSExpressionStatement (JSIdentifier _ _) _] _)  -> True
+    Right _ -> False
+    Left _ -> False
+
   sourceMapping :: String -> SourceMapping
   sourceMapping outFile = SourceMapping {
       smFile = outFile,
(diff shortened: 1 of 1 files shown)


michaelficarra:
:-1: If you want smaller code, you'll be using a minifier anyway. Whitespace is a much larger contributor to unnecessary bytes.

Also, this implementation doesn't cover Keywords, which are also IdentifierNames and thus allowed in static member access position.

paf31:
I would rather not complicate this code for such a small improvement in code size. My experience is that this code has been quite fragile in the past, and @michaelficarra lists some possible real issues.

In future, please create issues for discussion before opening PRs, that way we can avoid work being done on things which might not be merged. Thanks anyway though.

paf31:
Could you please open an issue to discuss this? We can reopen the PR if necessary, but I'm trying to cut down on open PRs we're uncertain about. Thanks.

Issue #2899, opened 2017-05-18
Add type-level natural numbers


rightfold:
Currently various encodings of type-level natural numbers are in use, which both have downsides in terms of performance:

1. Peano encoding quickly results in huge types, and reflection requires O(n) time.
2. Base-10 encoding results in smaller types, but still requires O(log_10 n) time.

They also don't result in the prettiest types, such as `S (S (S (S Z)))` for 4 or `D4 :* D2` for 42. It would be nice to have type-level natural numbers with nice syntax and better performance, both at compile-time and at runtime. Similar to `GHC.TypeLits`.

## Natural number types

- Add a new kind to `Prim` called `Nat`.
- Allow nonnegative integer literals in types. Each distinct such literal is a distinct type. The kind is `Prim.Nat`.

## Reflection and arithmetic library

Add a new PureScript library to core with the following module:

```purescript
module Data.Nat where

import Data.Int as Int
import Prelude

class IsNat (n :: Nat) where
  reflectNat :: proxy n -> Int

instance isNatMagic :: IsNat <magic> where
  reflectNat = <magic>

foreign import data Add :: Nat -> Nat -> Nat
foreign import data Sub :: Nat -> Nat -> Nat
foreign import data Mul :: Nat -> Nat -> Nat
foreign import data Div :: Nat -> Nat -> Nat
foreign import data Pow :: Nat -> Nat -> Nat

instance isNatAdd :: (IsNat a, IsNat b) => IsNat (Add a b) where
  reflectNat _ = reflectNat @a + reflectNat @b

instance isNatSub :: (IsNat a, IsNat b) => IsNat (Sub a b) where
  reflectNat _ = clamp 0 top $ reflectNat @a - reflectNat @b

instance isNatMul :: (IsNat a, IsNat b) => IsNat (Mul a b) where
  reflectNat _ = reflectNat @a * reflectNat @b

instance isNatDiv :: (IsNat a, IsNat b) => IsNat (Div a b) where
  reflectNat _ = reflectNat @a / reflectNat @b

instance isNatPow :: (IsNat a, IsNat b) => IsNat (Pow a b) where
  reflectNat _ = Int.pow (reflectNat @a) (reflectNat @b)

infixl 6 type Add as +
infixl 6 type Sub as -
infixl 7 type Mul as *
infixl 7 type Div as /
```

Where `<magic>` means "any non-negative type-level integer literal".

## Example

```purescript
module Data.Vect
  ( Vect
  , empty
  , cons
[...]

PR #3058, opened 2017-08-29
Type-level naturals

kcsongor:
A first attempt at implementing some of the features in https://github.com/purescript/purescript/issues/2899

This is incomplete - PR opened for discussion

- [x] Add kind Nat (with type-level literals)
- [x] Solve IsNat
- [x] Solve AddNat
- [x] multiplication / division
- [ ] pow / root
- [ ] conversion to a Peano encoding? Something like `NatToPeano` would allow to recurse on naturals easily - not sure how important this will be in the presence of instance chains though

this allows us to write
```
module Type.Data.Nat
  ( module Data.Nat
  , class AddNat
  , add
  , subtract
  ) where

import Data.Nat (NProxy(..), class IsNat, reflectNat, reifyNat)

class AddNat (a :: Nat) (b :: Nat) (c :: Nat) | a b -> c, a c -> b, c b -> a

add :: forall a b c. AddNat a b c => NProxy a -> NProxy b -> NProxy c
add _ _ = NProxy

subtract :: forall a b c. AddNat a b c => NProxy c -> NProxy b -> NProxy a
subtract _ _ = NProxy
```


paf31:
How about putting the classes in `Prim.Nat`, since they don't have any members and therefore don't need any runtime dependencies?

kcsongor:
That makes sense, I'll do that. Does that mean that there are plans to move the `Symbol` classes into `Prim` too?

paf31:
I think `AppendSymbol` and `CompareSymbol` should probably go in `Prim.Symbol`, yes.

src/Language/PureScript/Constants.hs
```diff
@@ -342,6 +342,29 @@ pattern DataSymbol = ModuleName [ProperName "Data", ProperName "Symbol"]
...
+
+pattern IsNat :: Qualified (ProperName 'ClassName)
+pattern IsNat = Qualified (Just DataNat) (ProperName "IsNat")
+
+-- Prim.Nat
+
+pattern TypeDataNat :: ModuleName
+pattern TypeDataNat = ModuleName [ProperName "Prim", ProperName "Nat"]
+
+pattern CompareNat :: Qualified (ProperName 'ClassName)
```
paf31: Can these move to `Prim.Nat` as well?

kcsongor: I should rename this patsyn to `PrimNat`, but as far as I can tell, all the purely typelevel classes are in `Prim.Nat` now - I didn't move `IsNat`, because it has a `reifyNat` function. Wouldn't that be a problem?

paf31: Great!

> Wouldn't that be a problem?

Yes, it would, unless we find a better way to encode things. We could always say the integer _is_ the dictionary, but then we wouldn't be able to write the class definition in such a way that people couldn't write bad instances.

paf31: You'll also need to add these to `primClasses`, and also add docstrings for them in the corresponding `Docs` module, so that they can be linked on Pursuit.

paf31:
@kcsongor Do you think this could be finished before we make a 0.12 release candidate? Skipping powers and roots wouldn't be the end of the world, I think, since we could implement those later. Also, `NatToPeano` could be implemented using instance chains for now, as you say. I think all that's left is to add the new data to the docs pages and to `primClasses`.

kcsongor:
Oh, definitely. I’m going to update this tomorrow

kcsongor:
Another question:

At the moment, we don't have polymorphic literals, which means that once we commit to `Nat`, we can't have `Int` with the same syntax, which would be a shame. How about changing these `Nat`s to allow negative values (and call them `Int` accordingly), and `Nat`-ness could just be a constraint that the `Int` is greater than 0. This would mean that we would have an `Int` kind, and an additional `IsNat` class, which doesn't have to be solved by the compiler.

Implementation-wise, this is trivial, and I think it would allow us to cover more ground.

paf31:
> How about changing these Nats to allow negative values (and call them Int accordingly), and Nat-ness could just be a constraint that the Int is greater than 0.

I feel like ints aren't as useful from a proving-things point of view. What induction principle would you use?

kcsongor:
correction: `Nat`s would be greater than or equal to 0.

`Int`s aren't as useful for that, but with instance chains, I think that shouldn't be a problem. What I mean is that these numbers can be treated like:

```
class IsNat n <= Induction (n :: Int) (p :: Int -> Type)

instance inductionZero :: Induction 0 p
else instance inductionS :: (IsNat n, Induction n p, Add n 1 m) => Induction m p
```

And negative numbers simply don't satisfy the superclass constraint.

paf31:
If you forget the `IsNat` constraint though, you'll end up spinning forever, right?

paf31:
You could always build the induction principle into `IsNat`, but I would say it's better to follow the maths and take `Nat` as the more primitive thing. I'm not terribly bothered about syntax, personally, since the type-level syntax isn't exactly pretty right now anyway.

kcsongor:
> If you forget the IsNat constraint though, you'll end up spinning forever, right?
Yes, that's a good point.

`Int` could also be implemented as a library, with the somewhat clumsier syntax:

```haskell
foreign import kind Int
foreign import data Pos :: Nat -> Int
foreign import data Neg :: Nat -> Int
```

and I think `Pos 10` and `Neg 10` are not too bad.

LiamGoodacre:
Annoyingly you get two zeros with that `Pos 0`/`Neg 0`.  But that's possibly okay, depending on how it's handled/context.

paf31:
Negative and non-negative is another way to go, where `Negative n` means `-(n + 1)`. Not exactly elegant though.

paf31:
The Grothendieck construction seems to be a nice way though, with an int being a pair of nats, with an equivalence relation such that only the signed difference matters.

kcsongor:
@paf31 I've moved the classes to `Prim`. I tried to put them into `Prim.Nat` first, but I couldn't figure out how to convince the compiler that `Prim.Nat` is a valid class when importing it. If you could point me where this is done, I can change the imports.

As an additional point, the kind signature of `CompareSymbol` required the definition of the kind `Ordering` in `Constants.hs`.

Thanks

parsonsmatt:
@kcsongor You might be interested in the work I did in #3176 with renaming the Prim classes. It's a bit hairier than you might expect.

kcsongor:
Excellent, thanks! I could tell it was going to be a little hairy

natefaubion:
This hasn't seen activity in some time. Should this be closed?

hdgarrood:
I'm going to close this for now due to lack of movement, but I'll be happy to reopen if you manage to find time to pick this up again. For anyone else who might like to pick this up, please feel free to submit a new PR building on top of this too.

PR #3062, opened 2017-09-06
psci: Move env out of initial config into state so it's updated on reload

### LINKED ISSUE #3001: :browse does not pick up changes from :reload
For example:

```
$ pulp init
$ pulp repl
<snip>

> :b Main


main :: forall e.
          Eff
            ( console :: CONSOLE
            | e
            )
            Unit
> -- now, in a separate window, add a new export to Main
> :reload
Compiling Main
> :b Main


main :: forall e.
          Eff
            ( console :: CONSOLE
            | e
            )
            Unit
```

Notice that the new export is not listed, despite Main having been compiled again.

I think the issue is [here in Language.PureScript.Interactive](https://github.com/purescript/purescript/blob/f8b86e8b0091be84e7676c2035d347bed75450df/src/Language/PureScript/Interactive.hs#L126) as the new `env` is just being chucked away. Perhaps the fix is to move the `env` out of the `PSCiConfig` and into the `PSCiState`, since it can change over time? Or maybe the `env` can be computed on the fly when needed from the loaded externs?

actionshrimp:
This fixes #3001, using the fix suggested in that issue report! Not entirely sold on the way the test is done as it's a bit stateful, but I don't really see another way of doing it due to the nature of what's being tested, suggestions there welcome!


tests/TestPsci/CommandTest.hs
```diff
@@ -34,7 +41,14 @@ commandTests = context "commandTests" $ do
...
 
   specPSCi ":complete" $ do
-    ":complete ma" `prints` []
+    ":complete ma" `prints` ""
     ":complete Data.Functor.ma" `prints` (unlines (map ("Data.Functor." ++ ) ["map", "mapFlipped"]))
     run "import Data.Functor"
     ":complete ma" `prints` (unlines ["map", "mapFlipped"])
+
+  specPSCi ":browse" $ do
+    liftIO $ setSample ".updated"
```
paf31: If we're really going to copy files around then we should put things back after we're done, please, using `bracket` to handle exceptions properly.

tests/support/psci/Sample.purs
```diff
@@ -0,0 +1,13 @@
...
+
+import Prelude
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (CONSOLE, log)
+
+main :: forall e. Eff (console :: CONSOLE | e) Unit
+main = do
+  log "Hello sailor!"
+
+another :: forall e. Eff (console :: CONSOLE | e) Unit
```
paf31: Perhaps it's simpler to just append these lines onto the file?

Ideally, I'd prefer to do this by mocking file IO. Could you please see if that's possible?

tests/TestPsci/CommandTest.hs
actionshrimp: Wasn't aware of this - thanks for the heads up!

tests/support/psci/Sample.purs
actionshrimp: Completely agree, will have a bit more of a dig. Thanks for taking a look.

PR #3081, opened 2017-09-20
[WIP] Default method implementations

### LINKED ISSUE #3067: Default method implementations
@joneshf suggests in https://github.com/purescript/purescript/issues/361#issuecomment-39924958 the following (old syntax?)
```hs
class Eq a where
  (==) :: a -> a -> Boolean
  (/=) :: a -> a -> Boolean

  instance Eq a where
    (==) x y = not (x /= y)
    (/=) x y = not (x == y)
```
As an alternative I would like to propose
```hs
class Eq a where
  eq :: a -> a -> Boolean
  eq x y = not (notEq x y)

  notEq :: a -> a -> Boolean
  notEq x y = not (eq x y)
```

gabejohnson:
Resolves #3067


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -510,6 +510,10 @@ data Declaration
   --
   | TypeClassDeclaration SourceAnn (ProperName 'ClassName) [(Text, Maybe Kind)] [Constraint] [FunctionalDependency] [Declaration]
   -- |
+  -- A type class member declaration (type declaration, optional default implementation)
+  --
+  | TypeClassMemberDeclaration SourceAnn (Declaration, Maybe Declaration)
```
paf31: I think it'd be good to use `ValueDeclaration` here, and take advantage of the existing desugaring code.

gabejohnson: The `(Declaration, Maybe Declaration)` is in fact `(TypeDeclarationData, Maybe ValueDeclarationData)`. Should I make this explicit?

paf31: I'd prefer this be done like regular value and type declarations if possible.

gabejohnson: I was trying to prevent orphaned value declarations. I can see now that I should be doing that during desugaring.

parsonsmatt:
This has a `needs more work` label; could it be documented what additional work needs to be done?

gabejohnson:
@parsonsmatt frankly I was out of my depth when I opened this PR and haven't worked on it in a couple of months. I planned to take another look at it over the holidays, but if you're interested in implementing it feel free 😄 

And to answer your question, the additional work needed is everything after parsing.

gabejohnson:
I'm closing this as I don't see myself working on it in the foreseeable future.

PR #3084, opened 2017-09-21
Update Overlapping instances error message.

### LINKED ISSUE #3071: Overlapping instance error message out of date
When I compile a module with an overlapping instance (using HEAD), it is now an error as opposed to a warning, but the error message still reads:

"They may be disallowed completely in a future version of the compiler".

/cc @LiamGoodacre

drets:
Fixes #3071.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -608,7 +608,6 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
                 ]
             , line "The following instances were found:"
             , indent $ paras (map (line . showQualified showIdent) ds)
-            , line "Overlapping type class instances can lead to different behavior based on the order of module imports, and for that reason are disallowed."
             ]
     renderSimpleErrorMessage (UnknownClass nm) =
       paras [ line "No type class instance was found for class"
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -599,18 +599,17 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
             , line "with type"
             , markCodeBox $ indent $ typeAsBox t2
             ]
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -599,18 +599,17 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
             , markCodeBox $ indent $ Box.hsep 1 Box.left
                 [ line (showQualified runProperName nm)
                 , Box.vcat Box.left (map typeAtomAsBox ts)
                 ]
             , line "The following instances were found:"
-            , indent $ paras (line (showQualified showIdent d <> " (chosen)") : map (line . showQualified showIdent) ds)
-            , line "Overlapping type class instances can lead to different behavior based on the order of module imports, and for that reason are not recommended."
-            , line "They may be disallowed completely in a future version of the compiler."
+            , indent $ paras (map (line . showQualified showIdent) ds)
+            , line "Overlapping type class instances can lead to different behavior based on the order of module imports, and for that reason are disallowed."
```
LiamGoodacre: I'm wondering if there is any point in mentioning the order of module imports bit. Given that they are supposed to be disallowed anyway, you're not supposed to be able to observe the described "different behavior" because you'd just see this error message. (Although the overlap check does need improving so that it catches all scenarios).

paf31: I'm also tempted to get rid of this line. We can give a more detailed explanation on the wiki.

drets: Thanks guys for comments!

I removed the line with the hint why overlapping instances are bad and added myself to contributors.

> (Although the overlap check does need improving so that it catches all scenarios).

If there is something else should be done outside of the file it's probably easier to close a PR. Since I need pointers of how to do that.

drets:
Thank you, guys :wink:

Issue #3085, opened 2017-09-21
Double parens in type causes crash on master


LiamGoodacre:
Having double parentheses in a type causes a crash 😄 

```purescript
type Eg = ((Int))
```

```
Compiling Main
purs: An internal error occurred during compilation: "Invalid argument to infer: ParensInType (TypeConstructor (Qualified (Just (ModuleName [ProperName {runProperName = \"Prim\"}])) (ProperName {runProperName = \"Int\"})))"
Please report this at https://github.com/purescript/purescript/issues
CallStack (from HasCallStack):
  error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-0.11.6-4vkhJJqnGWo13TgqQuOCFm:Language.PureScript.Crash
  internalError, called at src/Language/PureScript/TypeChecker/Kinds.hs:268:11 in purescript-0.11.6-4vkhJJqnGWo13TgqQuOCFm:Language.PureScript.TypeChecker.Kinds
purs: thread blocked indefinitely in an MVar operation
* ERROR: Subcommand terminated with exit code 1
```

This doesn't appear to cause any issues on try purs, so I think it's master only at the moment.

PR #3086, opened 2017-09-22
Desugar nested parentheses

MonoidMusician:
Doubly-nested parentheses (as in `((Unit))`) caused an issue in inferring kinds (Invalid argument to `infer`, i.e. `ParensInType`) then in unifying types (Could not match type `Unit` with `Unit`), but desugaring it earlier fixes both of those problems.

### DIFF
--- a/examples/passing/ParensInType.purs
+++ b/examples/passing/ParensInType.purs
@@ -1,0 +1,20 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (CONSOLE, log)
+
+class Foo a where
+  foo :: forall eff. (String -> a (( console :: CONSOLE | eff)) ((Unit)))
+
+instance fooLogEff :: Foo Eff where
+  foo = log
+
+main :: 
+    forall eff.
+      Eff
+        ( console :: CONSOLE
+        | eff
+        )
+        Unit
+main = foo "Done"
--- a/src/Language/PureScript/Sugar/Operators.hs
+++ b/src/Language/PureScript/Sugar/Operators.hs
@@ -230,15 +230,15 @@ removeParens = f
   (goDecl, goExpr', goBinder') = updateTypes (\_ -> return . goType)
 
   goExpr :: Expr -> Expr
-  goExpr (Parens val) = val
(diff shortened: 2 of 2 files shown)


garyb:
There's something a bit odd going on here, the original code should hit all the `Parens` cases, as the traversal is supposed to visit every node.

MonoidMusician:
hm ... it looks like `go` here unwraps then rewraps the `ParensInType`, so would that cause it to be preserved, if `f` returns a `ParensInType`? https://github.com/purescript/purescript/blob/74bc4a923dfbd84e1f4dfa448981b57ab3d4ab9c/src/Language/PureScript/Types.hs#L287

garyb:
Ahh interesting, yeah. The other two traversals are not doing this because they're not using a `topDownM` variety. I wonder if it will break something elsewhere if we change `updateTypes` to use a bottom-up traversal instead... I would assume so, it's probably chosen that way for a reason.

Maybe we can use a traversal directly rather than using `updateTypes` here?

paf31:
This might be my fault. I think I changed that code in a recent PR to try to fix something else.

paf31:
Here: https://github.com/purescript/purescript/commit/2e1b8527ce8f2ea310d2d249daf8cab5930539f5

LiamGoodacre:
Will this also fix #3085?

MonoidMusician:
Yes, that's the same issue I ran into, thanks!

paf31:
I think this will cause a quadratic runtime won't it?

Maybe it's possible to solve this with two traversals instead of one, one up, one down.

MonoidMusician:
I'm afraid I don't understand, how would two traversals be more efficient than manually looking inside the immediate next expression? I don't think my solution adds more than an extra check per usage of (non-nested) parentheses.

garyb:
Assuming we do do it this way, the change only needs to happen for `goTypes`, and it'd be good to include a comment as to why that case is different from the others.

LiamGoodacre:
Bump.  @paf31 are you happy with the proposed solution of changing only `goTypes`?

paf31:
I think it's probably the best we can do for now, but we should try to make sure it doesn't introduce a performance regression.

PR #3088, opened 2017-09-24
Expand error message for UnusableDeclaration

i-am-tom:
Previously, the error message made no mention of a solution to the error. This update adds possible solutions to the error message in the form of the remaining covering sets. In other words, the remaining variables that need to be determined in order to satisfy the typechecker.

I think a couple of these changes look a bit ugly, so please do leave opinions :)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -51,6 +51,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@houli](https://github.com/houli) | Eoin Houlihan | [MIT license](http://opensource.org/licenses/MIT) |
 | [@ianbollinger](https://github.com/ianbollinger) | Ian D. Bollinger | [MIT license](http://opensource.org/licenses/MIT) |
 | [@ilovezfs](https://github.com/ilovezfs) | ilovezfs | MIT license |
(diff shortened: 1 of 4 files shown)


LiamGoodacre:
Thanks.  Please add yourself to the contributors file.

i-am-tom:
@LiamGoodacre Done! Also added some more TLC to the error messages, so they're hopefully a bit easier to read in the simpler case.

garyb:
Still no mention of fundeps being the determinant? :wink:

LiamGoodacre:
Fundeps aren't necessarily involved here.  Consider:

```purescript
class Foo t where
  foo :: String
```

LiamGoodacre:
I mean you could argue that is like `t -> t` - but it's not explicit.

paf31:
Or `| -> t` should solve it too.

paf31:
Strangely it doesn't though.

paf31:
👍 Looks good to me. I think we can add an extended explanation of fundeps in the docs repo.

LiamGoodacre:
> Strangely it doesn't though.

@paf31 hmm, sounds like there's an issue with this edge case when we build the covering sets.

i-am-tom:
@paf31 It's on my list of docs to expand :) PR to follow in the next couple days!

PR #3095, opened 2017-09-26
Fixing some issues with proxies

### LINKED ISSUE #3072: Crash using new proxy syntax
Input:

```haskell
module Main where
class C a b | a -> b
instance c :: C Int String
go :: forall a b. C a b => @a -> b -> b
go _ b = b
foo :: _
foo = go @Int "hello"
```

Output:

```
purs: An internal error occurred during compilation: "ctorModules: unqualified type name"
  error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-0.11.6-4vkhJJqnGWo13TgqQuOCFm:Language.PureScript.Crash
  internalError, called at src/Language/PureScript/TypeChecker/Entailment.hs:201:59 in purescript-0.11.6-4vkhJJqnGWo13TgqQuOCFm:Language.PureScript.TypeChecker.Entailment
```

LiamGoodacre:
Fixes #3072

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -394,8 +394,9 @@ infer' (IfThenElse cond th el) = do
 infer' (Let ds val) = do
   (ds', val'@(TypedValue _ _ valTy)) <- inferLetBinding [] ds val infer
   return $ TypedValue True (Let ds' val') valTy
-infer' (Proxy ty) =
-  return $ TypedValue True (Proxy ty) (ProxyType ty)
+infer' (Proxy ty) = do
+  ty' <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< replaceTypeWildcards $ ty
+  return $ TypedValue True (Proxy ty') (ProxyType ty')
 infer' (DeferredDictionary className tys) = do
   dicts <- getTypeClassDictionaries
   hints <- getHints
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -94,6 +94,8 @@ subsumes' mode (KindedType ty1 _) ty2 =
   subsumes' mode ty1 ty2
 subsumes' mode ty1 (KindedType ty2 _) =
   subsumes' mode ty1 ty2
+subsumes' mode (ProxyType ty1) (ProxyType ty2) =
```
paf31: We shouldn't have proxies for polymorphic things, so I don't know why we need this. In any case, I don't think it's true that `Proxy` is necessarily covariant.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -732,6 +733,11 @@ check' val kt@(KindedType ty kind) = do
 check' (PositionedValue pos c val) ty = warnAndRethrowWithPositionTC pos $ do
   TypedValue t v ty' <- check' val ty
   return $ TypedValue t (PositionedValue pos c v) ty'
+check' p@(Proxy ty1) ty2 = do
```
paf31: I don't think we need this case. Falling through to `infer` should work.

paf31:
Looks good! Do we skolemize the type under the `Proxy` constructor yet in `skolemizeTypesInValue`? We probably should.

LiamGoodacre:
@paf31, thanks for reviewing so quickly.  I was sure there'd be some redundant changes - I changed lots of stuff until it worked and hadn't worked out which bits actually mattered.

PR #3099, opened 2017-10-02
REPL fixes needed for Eff change

### LINKED ISSUE #3080: Switch from Eff to IO
Optimizer rules need to be updated as part of the switch to `IO` as the default effect monad.

paf31:
To accompany https://github.com/purescript/purescript-eff/pull/27

Closes #3080

Optimizer changes aren't needed if we don't change the module or type names, since the optimizer doesn't know about types anyway, only dictionary names, which won't change.


paf31:
I have to remember to change the optimizer rules for `ST`, however.

paf31:
I think I should close this and reopen against latest master since it's so out of date.

paf31:
@hdgarrood Which type were you referring to?

paf31:
You could try removing `typeDecl`. I'm not sure if it's needed for any values with polymorphic types, but I can't think of any examples right now.

hdgarrood:
I was referring to the type in `typeDecl`

PR #3106, opened 2017-10-06
Add valid location list to orphan instance error

### LINKED ISSUE #3063: Orphan instance error wording reads like a warning
As discussed on slack, the wording of orphan instance errors reads like a suggestion rather than a requirement. I'm opening this issue to discuss if we want to improve it, and how we can do that.

This is the line in question:
https://github.com/purescript/purescript/blob/ad0c7c04fca8b2dda5a71f558ed943d0b16a2671/src/Language/PureScript/Errors.hs#L776

What if we said something like:

```
There are two ways to fix this error
    1. Move the instance to the same module the class is defined in.
    2. Move the instance to the same module the data type is defined in.
```

Maybe we could make it even easier and mention which module the class and data type are defined in? Do we have that information in the error?

i-am-tom:
Previously, the orphan instance simply said to consider a newtype. Now, it lists all the places in the project that would be considered valid locations for the instance as well. This hopefully closes #3063, and looks a little something like:

```
Error found:
in module $PSCI
at  line 1, column 1 - line 1, column 35

  Type class instance showFn for

    Data.Show.Show (a -> b)

  is an orphan instance.
  To avoid this error, declare the instance in one of the following modules if possible:

    Data.Show

  Otherwise, consider adding a newtype wrapper.
```

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -774,9 +774,9 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
                 [] -> [ line "There is nowhere this instance can be placed without being an orphan."
                       , line "A newtype wrapper can be used to avoid this problem."
                       ]
-                _  -> [ Box.text $ "This instance must be declared in "
+                _  -> [ Box.text $ "This problem can be resolved by declaring the instance in "
                           <> T.unpack formattedModules
-                          <> ", or be defined for a newtype wrapper."
+                          <> ", or by defining the instance on a newtype wrapper."
                       ]
                 ]
       where
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -771,12 +771,16 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
                 , Box.vcat Box.left (map typeAtomAsBox ts)
                 ]
             , line "is an orphan instance."
-            , line "To avoid this error, declare the instance in one of the following modules if possible:"
-            , markCodeBox $ indent $ Box.hsep 1 Box.left $ (line . runModuleName) <$> nonOrphans
-            , line "Otherwise, consider adding a newtype wrapper."
+            , suggestion
             ]
       where
-        nonOrphans = S.toList $ S.delete (moduleNameFromString "Prim") nonOrphanModules
+        suggestion = Box.vcat Box.left $
+          case S.toList $ S.delete (moduleNameFromString "Prim") nonOrphanModules of
+            [] -> [ line "There is nowhere this instance can be placed without being an orphan. A newtype wrapper can be used to avoid this problem." ]
+            xs -> [ line "This instance must be declared in one of the following modules:"
+                  , Box.vcat Box.left $ (line . runModuleName) <$> xs
+                  , line "Otherwise, a newtype wrapper can be used to solve this problem."
+                  ]
     renderSimpleErrorMessage (InvalidNewtype name) =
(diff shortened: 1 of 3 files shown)


LiamGoodacre:
What does the error message look like for:
```purescript
instance what :: Fail "what"
```

LiamGoodacre:
Not that anyone should be writing that, but my intent is to suppose there is a prim type class for which it might make sense to make instances of.

i-am-tom:
@LiamGoodacre Ha! So, originally, it would have listed `Prim` as an option. However, I am now filtering it out. What sort of error would you like to see in this instance?

i-am-tom:
## Updates from review
Firstly, there's a little change to the error output in the case of
there being no place for the instance except Prim. Previously, it just
printed an empty list, which didn't make much sense, gramatically. Now,
however, a different error is displayed to point out that there's no
reasonable home for the instance.

As well as this, the `typeModule` function in `TypeChecker.hs` can now
deal with proxy types. Previously, it couldn't, and the tests were
passing really due to an accident of laziness. Now, the tests pass, and
all is well in the world.

src/Language/PureScript/Errors.hs
```diff
@@ -763,16 +764,24 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
-            , line "An orphan instance is one which is defined in a module that is unrelated to either the class or the collection of data types that the instance is defined for."
-            , line "Consider moving the instance, if possible, or using a newtype wrapper."
-            ]
+            , Box.vcat Box.left $ case modulesToList of
+                [] -> [ line "There is nowhere this instance can be placed without being an orphan."
+                      , line "A newtype wrapper can be used to avoid this problem."
+                      ]
+                _  -> [ Box.text $ "This instance must be declared in "
+                          <> T.unpack formattedModules
+                          <> ", or be defined for a newtype wrapper."
```
LiamGoodacre: I'm wondering about the phrasing here; the use of 'must' makes it sound like these are strictly the only two possible solutions.  However, there are potentially many other contextually dependent solutions.

src/Language/PureScript/Errors.hs
```diff
@@ -763,16 +764,24 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
                 [ line (showQualified runProperName cnm)
                 , Box.vcat Box.left (map typeAtomAsBox ts)
                 ]
-            , line "is an orphan instance."
-            , line "An orphan instance is one which is defined in a module that is unrelated to either the class or the collection of data types that the instance is defined for."
-            , line "Consider moving the instance, if possible, or using a newtype wrapper."
-            ]
+            , Box.vcat Box.left $ case modulesToList of
+                [] -> [ line "There is nowhere this instance can be placed without being an orphan."
+                      , line "A newtype wrapper can be used to avoid this problem."
```
justinwoo: Seems like we'd also probably need a link to what the hell a newtype wrapper is and how to define it?

src/Language/PureScript/Errors.hs
i-am-tom: https://github.com/purescript/purescript/pull/3106/commits/c724145c3aa8c557ea2f01a0883b635cf43e35f3#diff-2a67dc7a3cb0d6822d66486323cc9420R777 - what do you think?

src/Language/PureScript/Errors.hs
i-am-tom: Spoke to @paf31, suggested that this level of detail would be better suited to the wiki. We're going to talk a bit more about this later, and I'll probably make another docs PR :)

i-am-tom:
:tada:

paf31:
Thanks!

PR #3108, opened 2017-10-08
Adds a (non-completed) document for ide's design.

kRITZCREEK:
I hope this helps potential contributors and people generally interested in the project.

It's not completely finished, but I think it's easier to track my progress and get feedback when this is in the open, so I figured I'd just open a PR.

[Rendered](https://github.com/kRITZCREEK/purescript/blob/ide-design-document/psc-ide/DESIGN.org)

### DIFF
--- a/psc-ide/DESIGN.org
+++ b/psc-ide/DESIGN.org
@@ -184,3 +184,39 @@
 * Tips and Tricks
 ** Running only =ide='s test suite
    ~stack test --test-arguments "-m Language.PureScript.Ide"~
+* Thoughts to be assigned
+** Using externs files as source of truth
+*** Pros
+- Everything has types, because it went through the compiler
+- Module visibility is respected, because everything went through the compiler
+- Works even when the source file has syntax errors/doesn't compile
+- Easy plug-and-play, people rarely change the `output/` directory (as
+  opposed to the file structure)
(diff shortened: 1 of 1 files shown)


psc-ide/DESIGN.org
```diff
@@ -0,0 +1,235 @@
...
+*** Pros
+- All data is regenerated on starting ide = no cache invalidation necessary
+- Things are fast, without any effort spent on optimizing things
+- Simple model, keeps complexity low
+- We don't polute projects with ide artifacts
+*** Cons
+- Imposes a limit on how big of a project we can handle
+- Means we need to be careful about what information we denormalize, since it
+  can blow up on us
+- All data is regenerated on starting ide = slower startup then (maybe?) necessary
```
paf31: then -> than?

kRITZCREEK:
I'm ready to merge this now, but I fully intend on continuing to work on this. I considered moving this into the documentation repository, but at the same time I think it's better if it's kept in sync and versioned with the corresponding changes in `ide`.

Should we link to this document from somewhere? Maybe from the CONTRIBUTING.md file?

paf31:
Yes, let's add it here, and link to it from the docs repo and elsewhere. I'd like to keep things with the compiler if we can, if they are documenting compiler behavior. I'm thinking of working on a type system doc soon too.

PR #3097, opened 2017-09-27
Option --dump-coreimp to output imperative-core to JSON

metaleap:
New `purs compile` option `--dump-coreimp`, exactly like the existing `--dump-corefn`, to dump the `CoreImp` AST to a `coreimp.json` output file.

Requested in #876 and also in #711 [in the comments](https://github.com/purescript/purescript/issues/711#issuecomment-74744519)

(The results can be seen in the various `coreimp.json` files throughout [this directory](https://github.com/metaleap/gonad-demos/tree/master/01_hello_stranger/output) --- [one example](https://github.com/metaleap/gonad-demos/blob/master/01_hello_stranger/output/Ohai.Greet/coreimp.json) )

**New modules:**

- `src/Language/PureScript/CoreImp/ToJSON.hs` --- modeled after the existing `src/Language/PureScript/CoreFn/ToJSON.hs`. (Has duplicated the trivial tiny helper functions `identToJSON` and `moduleNameToJSON` as I didn't want to impress changes on to what the existing original exports.)

**Existing modules:**

Merely duplications of the existing `--dump-corefn` option handling in:
- `app/Command/Compile.hs`
- `src/Language/PureScript/Make.hs`
- `src/Language/PureScript/Options.hs`
(Too small & trivial to generalize into helper funcs for now IMHO --- until a 3rd `--dump-xyz` request comes along in the future at least.)

PR #3117, opened 2017-10-16
Option --dump-coreimp to output imperative-core to JSON

metaleap:
_(I sent this before in #3097 but retracted back then, but *now* it's in a state where it's solidified (keeping the recommendations from reviewers back then) and I'd ask for another review on this PR.. I don't foresee any changes to this feature set from the uses I can think of. Mostly the below is a verbatim copy of then.)_

New `purs compile` option `--dump-coreimp`, exactly like the existing `--dump-corefn`, to dump the `CoreImp` AST to a `coreimp.json` output file.

Requested in #876 and also in #711 [in the comments](https://github.com/purescript/purescript/issues/711#issuecomment-74744519)

(The results can be seen in the various `coreimp.json` files throughout [this directory](https://github.com/metaleap/gonad-demos/tree/master/output) )

**New modules:**

- `src/Language/PureScript/CoreImp/ToJSON.hs` --- modeled after the existing `src/Language/PureScript/CoreFn/ToJSON.hs`. (Has duplicated the trivial tiny helper functions `identToJSON` and `moduleNameToJSON` as I didn't want to impress changes on to what the existing original exports.)

**Existing modules:**

Mostly, merely duplications of the existing `--dump-corefn` option handling in:
- `app/Command/Compile.hs`
- `src/Language/PureScript/Make.hs`
- `src/Language/PureScript/Options.hs`
(Too small & trivial to generalize into helper funcs for now IMHO --- until a 3rd `--dump-xyz` request comes along in the future at least.)

Plus, `src/Language/PureScript/CoreFn/Meta.hs` added `deriveJSON` for `ConstructorType` and `Meta` --- part of this `--dump-coreimp` use-case as currently devised here is to indeed dump the various meta annotations and type information already "in-flight". In my currently-baking PureScript-to-Golang backend I'll need it on-top-of what `externs.json` has. I expect other future `coreimp` uses may as well. (When not needed, the consumer can obviously just ignore the stuff in the JSON's `decls` entry.)

### DIFF
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -132,11 +132,17 @@ dumpCoreFn = Opts.switch $
      Opts.long "dump-corefn"
   <> Opts.help "Dump the (functional) core representation of the compiled code at output/*/corefn.json"
 
+dumpCoreImp :: Opts.Parser Bool
+dumpCoreImp = Opts.switch $
+     Opts.long "dump-coreimp"
+  <> Opts.help "Dump the (imperative) core representation of the compiled code at output/*/coreimp.json"
+
 options :: Opts.Parser P.Options
 options = P.Options <$> verboseErrors
                     <*> (not <$> comments)
                     <*> sourceMaps
                     <*> dumpCoreFn
+                    <*> dumpCoreImp
 
 pscMakeOptions :: Opts.Parser PSCMakeOptions
 pscMakeOptions = PSCMakeOptions <$> many inputFile
--- a/src/Language/PureScript/CoreFn/Meta.hs
+++ b/src/Language/PureScript/CoreFn/Meta.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE TemplateHaskell #-}
 -- |
 -- Metadata annotations for core functional representation
 --
@@ -6,6 +7,8 @@ module Language.PureScript.CoreFn.Meta where
 import Prelude.Compat
 
 import Language.PureScript.Names
+
+import Data.Aeson.TH
 
 -- |
 -- Metadata annotations
@@ -40,3 +43,8 @@ data ConstructorType
   -- The constructor is for a type with multiple construcors
   --
   | SumType deriving (Show, Eq, Ord)
+
+
+$(deriveJSON (defaultOptions { sumEncoding = ObjectWithSingleField }) ''ConstructorType)
+
+$(deriveJSON (defaultOptions { sumEncoding = ObjectWithSingleField }) ''Meta)
--- a/src/Language/PureScript/CoreImp/ToJSON.hs
+++ b/src/Language/PureScript/CoreImp/ToJSON.hs
@@ -1,0 +1,283 @@
+{-# LANGUAGE NoOverloadedStrings #-}
+-- |
+-- Dump the core imperative representation in JSON format for consumption
+-- by third-party code generators
+--
+module Language.PureScript.CoreImp.ToJSON
+  ( moduleToJSON
+  ) where
+
+
+import Prelude.Compat
+
+import Data.Aeson
+import Data.Maybe
+import qualified Data.Text as T
+import Data.Version (Version, showVersion)
+
+import Language.PureScript.AST (SourceSpan)
+import qualified Language.PureScript.AST.Literals as ASTL
+import Language.PureScript.CoreFn (Module, moduleDecls, moduleExports, moduleImports, moduleForeign)
+import Language.PureScript.CoreFn.Binders
+import qualified Language.PureScript.CoreFn.Expr as CoreFnExpr
+import Language.PureScript.CoreImp.AST
+import Language.PureScript.Names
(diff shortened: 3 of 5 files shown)


metaleap:
A question to the insiders about `foreign`! How can I get at the signatures or otherwise "more details (ideally all the details currently known by purs) about each `foreign import`" at [this](https://github.com/purescript/purescript/pull/3117/commits/1b79662efd17dd074e1abc29a97247f250412947#diff-2ffe84d2b44f04026b160ff94572114fR55) point? Right now I just get the `Ident`s aka names. Would be very desirable if I could export the signatures / type infos just like what is put in `externs.json` in here.

metaleap:
Regarding commit #60ae4c3 --- not sure what you guys prefer for `ToJSON` instances, `TemplateHaskell` or `DeriveGeneric`? Is one somehow "better" than the other? Right now I have used a mix of both in a Haskell-typical "add the stuff GHC whines about until it compiles" sprint.. 😀

metaleap:
A quick note on the general motivation here and on why I made the `Environment` `toJSON`able just now: I believe `--dump-coreimp` can become a great alternative option to writing alternative backends --- alternative to each alt-backend keeping forks of `purs` around that go out-of-sync over time and requires users to have multiple `purs` executables around. With this option, anyone can hack or script their desired PureScript(CoreImp)-to-XYZ transformations without having to navigate this *sizable*, slightly-daunting Haskell code-base, the slowish GHC/Stack/Cabal build process, or keeping their forks in sync with the main repo.

I took a quick look at purs11 (C++ backend) the other day and noticed that fork is so out of date, there isn't even any *CoreImp* in there! How could one hope to pick up on that if the need arises again in the future. With `--dump-coreimp` json exports, it's `purs` itself keeping the outputs necessarily "up-to-date", and one can write their small transpilers without plugging into a huge 3rd-party codebase.

**Given that, ** unless you're transpiling to similarly "dynamic" languages like JS, such as perhaps Python or Erlang, for many candidate targets that'd spring to mind for *me* (Go right now, potentially C#/Java/the likes) you'd want the `Environment` for the whole module, not just the exports already in `externs.json`. Type classes, instances, data types / ctors, type aliases, and "names" (func arg/ret types). This allows for much more idiomatic, more correct, more performant generated-code. Such languages of course *would* allow us to pass around "dynamic"/"any-typed" (boxed) values, but it's not desirable at all when it can be avoided. The philosophy of PureScript to generate readable idiomatic "as-if-human-written" code, I'd want to afford any `--dump-coreimp` consumers too. With the latest changes above, the `Environment` gets dumped into the `env` field and this becomes possible.

metaleap:
Gaah just realized `Environment` is _everything-from-all-modules_. Can't dump this for each module. Though it ran surprisingly fast! Will undo most of that stuff coming up next. Sheesh

paf31:
I don't have time to fully review this right now. Thanks for opening the PR, but I want to note that we need to discuss this before we land on a design. Imperative core is not finalized yet, so there is some design work which needs to go into the internal representation. And we will probably need to engage in some bikeshedding regarding the JSON format.

metaleap:
Sure take your time, just wanted to "plug back in" here and happy to keep updating the `CoreImp/ToJSON.hs` module in tune with any upcoming coreimp changes landing in `master`. It's all in flight, that's fine by me.

(Though I'm curious what you're thinking is still missing in imperative-core? Seems to cover everything that was needed to generate JS so far fairly completely, no?)

As regards to "JSON bikeshedding", that's to be expected but I'm about to clean up and "lean up" various parts of the output further as well.

metaleap:
@paf31 & @ all, a question re:

> Imperative core is not finalized yet, so there is some design work which needs to go into the internal representation

I can tell already that coreimp is a far cry from what it could be: fairly decoupled from JS. I'd ponder that stuff like `"use strict";`, `module.exports` etc. would do well to go into the coreimp-to-JS transformation, for example.. I'd even go so far as to say, it'd be desirable for CoreImp to preverse PS type info on values/args/signatures/constraint-type-vars, and only for the JS pass to erase them. (Now of course CoreImp uses JS type semantics as in prim-types/classes/objects/arrays. But it's all coming from some PS type semantics that can be carried along expressions as corefn-to-coreimp happens.)

But that's all fluffy ponderings.. Where's the "CoreImp design work" happening right now, who's involved and cracking this? Is there a thread? Is it all happening in the mailing list, or in some buried github-issue threads I should join? At the least I might like to follow along to pick up everyone's thinking of where this should go. I can also add the perspective of first-hand experience wrestling with the current state of coreimp on-the-ground  =)

natefaubion:
This has not seen activity in some time. Should this be closed?

metaleap:
Fine by me, it's been a year since this the last reaction saying:

> Imperative core is not finalized yet, so there is some design work which needs to go into the internal representation

and I've already long moved on from the effort I envisioned back then.  =)

garyb:
Thanks for the time you spent on it anyway!

PR #3121, opened 2017-10-20
Fix entailment issues with skolems and matches

LiamGoodacre:
When verifying a substitution, if we see a skolem (that hasn't been
matched with itself) then it is unknown whether or not we might have a
match or be apart.

Combining a match with an unknown should be an unknown not a match.
I.e: knowing that one of two things match doesn't make the other one now
a match.

Added failing examples to catch these cases.

### DIFF
--- a/examples/failing/InstanceChainBothUnknownAndMatch.purs
+++ b/examples/failing/InstanceChainBothUnknownAndMatch.purs
@@ -1,0 +1,14 @@
+-- @shouldFailWith NoInstanceFound
+module InstanceChains.BothUnknownAndMatch where
+
+class Same l r o | l r -> o
+instance sameY :: Same t t @"Y" else instance sameN :: Same l r @"N"
+same :: forall l r o. Same l r o => l -> r -> @o
+same _ _ = @o
+
+-- for label `u`, `t ~ Int` should be Unknown
+-- for label `m`, `Int ~ Int` should be a match
+-- together they should be Unknown
+example :: forall t. @t -> @_
+example _ = same @(u :: t, m :: Int) @(u :: Int, m :: Int)
+
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -138,13 +138,15 @@ data Matched t
   = Match t
   | Apart
   | Unknown
-  deriving (Eq, Show)
+  deriving (Eq, Show, Functor)
 
-bothMatched :: Matched () -> Matched () -> Matched ()
-bothMatched (Match _) (Match _) = Match ()
-bothMatched Unknown r = r
-bothMatched l Unknown = l
-bothMatched _ _ = Apart
+instance Monoid t => Monoid (Matched t) where
+  mempty = Match mempty
+
+  mappend (Match l) (Match r) = Match (l <> r)
+  mappend Apart     _         = Apart
+  mappend _         Apart     = Apart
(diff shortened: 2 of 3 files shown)


examples/failing/InstanceChainBothUnknownAndMatch.purs
```diff
@@ -0,0 +1,14 @@
+-- @shouldFailWith NoInstanceFound
+module InstanceChains.BothUnknownAndMatch where
+
+class Same l r o | l r -> o
+instance sameY :: Same t t @"Y" else instance sameN :: Same l r @"N"
```
paf31: Hmm, well there's the kind error again :(

paf31: Oh wait, no, I didn't see the proxy :)

examples/failing/InstanceChainBothUnknownAndMatch.purs
```diff
@@ -0,0 +1,14 @@
...
+same :: forall l r o. Same l r o => @l -> @r -> @o
+same _ _ = @o
+
+-- for label `u`, `t ~ Int` should be Unknown
+-- for label `m`, `Int ~ Int` should be a match
+-- together they should be Unknown
+example :: forall t. @t -> @_
+example _ = same @(u :: t, m :: Int) @(u :: Int, m :: Int)
```
paf31: But this _is_ a kind error, surely?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -585,10 +591,10 @@ mkContext = foldr combineContexts M.empty . map fromDict where
   fromDict d = M.singleton Nothing (M.singleton (tcdClassName d) (M.singleton (tcdValue d) d))
 
 -- | Check all pairs of values in a list match a predicate
-pairwiseAll :: (a -> a -> Bool) -> [a] -> Bool
-pairwiseAll _ [] = True
-pairwiseAll _ [_] = True
-pairwiseAll p (x : xs) = all (p x) xs && pairwiseAll p xs
+pairwiseAll :: (a -> a -> Matched ()) -> [a] -> Matched ()
```
paf31: This could use any monoid as the result, right?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -518,42 +520,46 @@ matches deps TypeClassDictionaryInScope{..} tys =
...
-    both (b1, m1) (b2, m2) = (bothMatched b1 b2, M.unionWith (++) m1 m2)
+    both (b1, m1) (b2, m2) = (b1 <> b2, M.unionWith (++) m1 m2)
 
     -- Ensure that a substitution is valid
-    verifySubstitution :: Matching [Type] -> Maybe (Matching [Type])
-    verifySubstitution = traverse meet where
-      meet ts | pairwiseAll typesAreEqual ts = Just ts
-              | otherwise = Nothing
+    verifySubstitution :: Matching [Type] -> Matched (Matching [Type])
+    verifySubstitution mts = foldMap meet mts $> mts where
```
paf31: This is nice :)

examples/failing/InstanceChainBothUnknownAndMatch.purs
LiamGoodacre: Note to readers: when this comment was made, `same` had type `forall l r o. Same l r o => @l -> @r -> @o`.

PR #3124, opened 2017-10-21
Print proxy type as an operator

### LINKED ISSUE #3076: Issue with pretty-printing nested proxies
If I have a nested proxy: `@(@"Hi")` it pretty-prints as `@@"Hi"` which isn't valid syntax:

```
Unable to parse module:
  unexpected @@
  expecting type
```

LiamGoodacre:
Fixes #3076 

Example input:
```purescript
x :: _
x = {x: @(@(@(Array Int))), x: @Int}
```
Output:
> Label x appears more than once in a row type.
> while inferring the type of
> ```purescript
> { x: @(@(@(Array Int)))
> , x: @Int
> }
> ```

I originally also changed the parser so it treated `@` as a prefix type operator that binds tighter than type application, but this caused problems with instances because `@Int` isn't a type-atom.  E.g: `instance foo :: Foo @t` would have to have been written as `instance foo :: Foo (@t)`.  So I reverted this.

### DIFF
--- a/src/Language/PureScript/Pretty/Types.hs
+++ b/src/Language/PureScript/Pretty/Types.hs
@@ -74,6 +74,15 @@ prettyPrintRowWith tro open close = uncurry listToBox . toList []
 prettyPrintRow :: Type -> String
 prettyPrintRow = render . prettyPrintRowWith defaultOptions '(' ')'
 
+-- Treat `ProxyType t` in a similar way to the application of a type
+-- constructor `@` to `t`, i.e: `@ t`, except that we don't render the unneeded
+-- space. So we end up with `@t`.
+proxyType :: Pattern () Type (Type, Type)
+proxyType = mkPattern match
+  where
+  match (ProxyType t) = Just (TypeConstructor (Qualified Nothing (ProperName "@")), t)
+  match _ = Nothing
+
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Pretty/Types.hs
```diff
@@ -74,6 +74,12 @@ prettyPrintRowWith tro open close = uncurry listToBox . toList []
 prettyPrintRow :: Type -> String
 prettyPrintRow = render . prettyPrintRowWith defaultOptions '(' ')'
 
+proxyType :: Pattern () Type (Type, Type)
+proxyType = mkPattern match
+  where
+  match (ProxyType t) = Just (TypeConstructor (Qualified Nothing (ProperName "@")), t)
```
paf31: Interesting trick :)

paf31: I'm wondering if we should add `Prefix` to `pattern-arrows` going forward, but I'm fine with this in the meantime. Could you please add a comment though?

LiamGoodacre:
Rebased and commented.  Let me know if you want it to say something else.

paf31:
👍 Sounds good, thanks!

PR #3127, opened 2017-10-22
Make error messages clearly state which is the expected and which is …

AidanDelaney:
…the actual type or kind

This is a trivial patch with a number of consequences.  I've asked for the PR as I'm constantly forgetting which way to read these errors.  Do they present the expected and then the actual or vice-versa?  Changing the language in the error message reduces this confusion.

This PR does not change the operation of the compiler in any functional way.  However, the consequences of this PR are for documentation including the Purescript Book.


LiamGoodacre:
This has come up a few times.  See the comments on #2620.

AidanDelaney:
Thanks, I don't feel so bad now.  I was wondering why I kept getting confused.  Obviously to fix this issue a much more substantial patch is required, so I'll close this request.

PR #3128, opened 2017-10-22
Instantiate abstraction body during inference

### LINKED ISSUE #3125: Position of inferred constraint differs when using a lambda vs equation-style def
Extracted from https://github.com/purescript/purescript/pull/2205#issuecomment-238034986

``` purescript 
module ConstraintPositionOddity where

import Data.HeytingAlgebra

data HeytingAlgebraDict a =
  HeytingAlgebraDict
    a -- ff
    a -- tt
    (a -> a -> a) -- implies
    (a -> a -> a) -- conj
    (a -> a -> a) -- disj
    (a -> a) -- not

hafGood :: ∀ a b. HeytingAlgebra b => HeytingAlgebraDict (a -> b)
hafGood =
  HeytingAlgebraDict
    (\_ -> ff)
    (\_ -> tt)
    (\f g a -> f a `implies` g a)
    (\f g a -> f a && g a)
    (\f g a -> f a || g a)
    (\f a -> not (f a))

hafBad :: ∀ a b. HeytingAlgebra b => HeytingAlgebraDict (a -> b)
hafBad =
  HeytingAlgebraDict
    _ff
    _tt
    _implies
    _conj
    _disj
    _not
  where
    _ff _ = ff
    _tt _ = tt
    _implies f g a = f a `implies` g a
    _conj f g a = f a && g a
    _disj f g a = f a || g a
    _not f a = not (f a)
```

In the above, the `hafGood` typechecks just fine, `hafBad` fails with a "Could not match constrained type" error.

If you follow the link to comment at the top, it illustrates how the equation defs desguar differently than the lambdas, which results in the constraint being inferred like `forall t1. t1 -> (forall a. (HeytingAlgebra a) => a)` instead of `forall t3 t4. (HeytingAlgebra t4) => t3 -> t4`.

LiamGoodacre:
Fixes #3125 

Currently there is an issue with shadowed type vars in the shifted `forall` - see my comment on #3125.

Does this seem like a good approach?

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -357,7 +357,8 @@ infer' (Abs binder ret)
       ty <- freshType
       withBindingGroupVisible $ bindLocalVariables [(arg, ty, Defined)] $ do
         body@(TypedValue _ _ bodyTy) <- infer' ret
-        return $ TypedValue True (Abs (VarBinder arg) body) $ function ty bodyTy
+        (body', bodyTy') <- instantiatePolyTypeWithUnknowns body bodyTy
+        return $ TypedValue True (Abs (VarBinder arg) body') (function ty bodyTy')
   | otherwise = internalError "Binder was not desugared"
 infer' (App f arg) = do
   f'@(TypedValue _ _ ft) <- infer f
--- a/examples/passing/3125.purs
+++ b/examples/passing/3125.purs
@@ -1,0 +1,16 @@
+module Main where
+
+import Prelude
+import Data.Monoid (class Monoid, mempty)
+import Control.Monad.Eff.Console (log, logShow)
+
+data B a = B a a
+
+memptyB :: forall a b. Monoid b => B (a -> b)
+memptyB = B l r where
+  l _ = mempty
+  r _ = mempty
+
+main = do
+  logShow $ case (memptyB :: B (Int -> Array Unit)) of B l r -> l 0 == r 0
+  log "Done"


garyb:
This was what I had thought of as a means of fixing it also, I just didn't work on it long enough (> 15 minutes :wink:) to figure out how to get it done. So yeah, seems like a good approach to me!

LiamGoodacre:
Fixed the renaming issue by not checking the type of the TypedValue when we need to re-infer.

garyb:
Should we add the example from the issue as a test case too?

LiamGoodacre:
Hmm, if I use `TypeValue False`, then it doesn't expand things properly anymore.  So I'll have to undo that and find a different way.

LiamGoodacre:
I don't need to re-infer, I just need to check the expression against the new type.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -357,7 +357,13 @@ infer' (Abs binder ret)
       ty <- freshType
       withBindingGroupVisible $ bindLocalVariables [(arg, ty, Defined)] $ do
         body@(TypedValue _ _ bodyTy) <- infer' ret
-        return $ TypedValue True (Abs (VarBinder arg) body) $ function ty bodyTy
+        let (qs, cs, ty') = moveQuantifiersToFrontOverFunction (function ty bodyTy)
+            expr = Abs (VarBinder arg) body
+        -- if we shifted some quantifiers/constraints then we'll need to check against the new type
+        expr' <- if null qs && null cs
+          then return expr
+          else check expr ty'
```
paf31: Can you explain why we need to check again here? It definitely seems odd to me to have to traverse the same terms twice.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -357,7 +357,13 @@ infer' (Abs binder ret)
       ty <- freshType
       withBindingGroupVisible $ bindLocalVariables [(arg, ty, Defined)] $ do
         body@(TypedValue _ _ bodyTy) <- infer' ret
-        return $ TypedValue True (Abs (VarBinder arg) body) $ function ty bodyTy
+        let (qs, cs, ty') = moveQuantifiersToFrontOverFunction (function ty bodyTy)
```
paf31: Can you not just instantiate the type of the body here?

LiamGoodacre: Yes, that makes more sense. 😸

paf31:
Thanks! Looks good, but could you test it with some existing packages as well, please?

LiamGoodacre:
I've checked that the following modules compile:

```
purescript-argonaut             purescript-either               purescript-lists                purescript-profunctor-lenses
purescript-argonaut-codecs      purescript-enums                purescript-maps                 purescript-proxy
purescript-argonaut-core        purescript-exceptions           purescript-math                 purescript-psci-support
purescript-argonaut-traversals  purescript-exists               purescript-maybe                purescript-record
purescript-arraybuffer-types    purescript-filterable           purescript-monoid               purescript-refs
purescript-arrays               purescript-foldable-traversable purescript-newtype              purescript-sets
purescript-assert               purescript-foreign              purescript-node-buffer          purescript-smolder
purescript-avar                 purescript-free                 purescript-node-http            purescript-st
purescript-behaviors            purescript-functions            purescript-node-path            purescript-strings
purescript-bifunctors           purescript-functors             purescript-node-streams         purescript-symbols
purescript-catenable-lists      purescript-gen                  purescript-node-url             purescript-tailrec
purescript-console              purescript-generics             purescript-nonempty             purescript-transformers
purescript-const                purescript-generics-rep         purescript-nullable             purescript-tuples
purescript-contravariant        purescript-globals              purescript-options              purescript-type-equality
purescript-control              purescript-identity             purescript-parallel             purescript-typelevel-prelude
purescript-debug                purescript-integers             purescript-partial              purescript-unfoldable
purescript-distributive         purescript-invariant            purescript-prelude              purescript-unsafe-coerce
purescript-eff                  purescript-lazy                 purescript-profunctor
```

Tried to check more than this but the removal of `Generic` caused me to cull them.
I also haven't gone through and checked that they generate the correct code.

Sometime this week I will run the tests of various modules to get a better assurance that things aren't broken.

garyb:
I think profunctor-lenses is a pretty good test for this, if that worked, I doubt there's anything that will be rejected now that previously did work! Nicely done.

LiamGoodacre:
I've checked that the following packages' tests pass: `profunctor-lenses`, `foldable-traversable`.  I also tried `halogen`, but it has dependencies on the newly removed `Generic` type class; so that obviously failed.

paf31:
👍 Thanks for testing. I say let's merge then.

garyb:
Nice, now I can revive my typeclass representation PR... for the 3rd attempt at it :smile:

Thanks Liam!

PR #3129, opened 2017-10-22
Check that instance declarations don't overlap

### LINKED ISSUE #2957: No warning for overlapping instances
The compiler only warns for overlapping instances when solving. This is actually a bug, as illustrated by the following example, which prints `false`, but gives no overlap warning:

```purescript
class IsInt a where
  isInt :: a -> Boolean
  
instance intIsInt :: IsInt Int where
  isInt _ = true
  
instance isn'tInt :: IsInt a where
  isInt _ = false
  
class F a b | a -> b where
  f :: a -> b
  
instance fInt :: F Int Int where
  f = id
  
main = logShow (isInt (f 1))
```

The problem is that the type of the call to `isInt` is not determined, since the `F` constraint gets solved later. Therefore we pick the `isn'tInt` instance, which is wrong.

In order to commit to `isn'tInt`, we must know that `a` can never become `Int` (apartness). Alternatively, we could try to warn at the definition site.

cc @LiamGoodacre You will probably find this interesting.

LiamGoodacre:
Fixes #2957 

I plan to add a failing example which demonstrates the covering sets aspect of this.

### DIFF
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -334,10 +334,13 @@ typeCheckAll moduleName _ = traverse go
         Just typeClass -> do
           checkInstanceArity dictName className typeClass tys
           sequence_ (zipWith (checkTypeClassInstance typeClass) [0..] tys)
-          checkOrphanInstance dictName className typeClass tys
+          let nonOrphanModules = findNonOrphanModules className typeClass tys
+          checkOrphanInstance dictName className tys nonOrphanModules
+          let qualifiedChain = Qualified (Just moduleName) <$> ch
+          checkOverlappingInstance qualifiedChain dictName className typeClass tys nonOrphanModules
           _ <- traverseTypeInstanceBody checkInstanceMembers body
           deps' <- (traverse . overConstraintArgs . traverse) replaceAllTypeSynonyms deps
-          let dict = TypeClassDictionaryInScope (Qualified (Just moduleName) <$> ch) idx qualifiedDictName [] className tys (Just deps')
+          let dict = TypeClassDictionaryInScope qualifiedChain idx qualifiedDictName [] className tys (Just deps')
           addTypeClassDictionaries (Just moduleName) . M.singleton className $ M.singleton (tcdValue dict) dict
           return d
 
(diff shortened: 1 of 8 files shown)


src/Language/PureScript/TypeChecker.hs
```diff
@@ -366,39 +369,92 @@ typeCheckAll moduleName _ = traverse go
...
 
-  checkOrphanInstance :: Ident -> Qualified (ProperName 'ClassName) -> TypeClassData -> [Type] -> m ()
-  checkOrphanInstance dictName className@(Qualified (Just mn') _) typeClass tys'
-    | moduleName `S.member` nonOrphanModules' = return ()
-    | otherwise = throwError . errorMessage $ OrphanInstance dictName className nonOrphanModules' tys'
+  findNonOrphanModules
+    :: Qualified (ProperName 'ClassName)
+    -> TypeClassData
+    -> [Type]
+    -> m (S.Set ModuleName)
```
paf31: No need for `m` here, right?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -366,39 +369,92 @@ typeCheckAll moduleName _ = traverse go
...
+    :: S.Set (S.Set Int)
+    -> [Type]
+    -> [Type]
+    -> Bool
+  instancesAreApart sets lhs rhs = all (any typesApart . S.toList) (S.toList sets)
+    where
+      typesApart :: Int -> Bool
+      typesApart i = typeHeadsApart (lhs !! i) (rhs !! i)
+
+      typeHeadsApart :: Type -> Type -> Bool
```
paf31: Maybe this should live alongside the equality test we have elsewhere?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -366,39 +369,92 @@ typeCheckAll moduleName _ = traverse go
...
+  checkOverlappingInstance
+    :: [Qualified Ident]
+    -> Ident
+    -> Qualified (ProperName 'ClassName)
+    -> TypeClassData
+    -> [Type]
+    -> S.Set ModuleName
+    -> m ()
+  checkOverlappingInstance ch dictName className typeClass tys' nonOrphanModules = do
+    for_ nonOrphanModules $ \m -> do
```
paf31: I think there is a possible issue with MPTCs here, where we can miss an instance in an unrelated module, but that's fine. We should probably just document it.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -366,39 +369,92 @@ typeCheckAll moduleName _ = traverse go
...
     -- sets - therefore we take the intersection of covering set modules.
-    nonOrphanModules :: S.Set ModuleName
-    nonOrphanModules = foldl1 S.intersection (foldMap lookupModule `S.map` typeClassCoveringSets typeClass)
-
-  checkOrphanInstance _ _ _ _ = internalError "Unqualified class name in checkOrphanInstance"
+    nonOrphanModules' :: S.Set ModuleName
+    nonOrphanModules' = foldl1 S.intersection (foldMap lookupModule `S.map` typeClassCoveringSets typeClass)
+  findNonOrphanModules _ _ _ = internalError "Unqualified class name in findNonOrphanModules"
+
+  checkOverlappingInstance
```
paf31: Could you please add a comment here?

src/Language/PureScript/TypeChecker.hs
LiamGoodacre: Unless I've misunderstood you, I'm not sure that's correct.  The modules in `nonOrphanModules` are the only modules that the instance is allowed to appear based on the type class's covering sets (we'd get an orphan instance error if one were in a different module).  So we are definitely seeing all the possible instances that could overlap.  Right?

src/Language/PureScript/TypeChecker.hs
LiamGoodacre: Oh right, I was thinking that `internalError` required it.

src/Language/PureScript/TypeChecker.hs
LiamGoodacre: Wait, I understand, we're only looking for modules that the instance talks about.

paf31: Right, I can always define an instance for an unrelated data type in another module. This is only an issue with `FlexibleInstances` or MPTCs though.

LiamGoodacre:
The build will fail due to the lack of https://github.com/purescript/purescript-psci-support/pull/7

LiamGoodacre:
Did you want to take another look at this? @paf31 (or anyone else interested).
I've now updated the tests to point at `purescript/purescript-psci-support#compiler/0.12`.

paf31:
I feel like I've reviewed this in quite a lot of detail already, and it's been tested quite a lot, so I think it's more than good enough for a first release.

LiamGoodacre:
(btw, I can't approve my own PRs) :smile_cat:

LiamGoodacre:
Thanks @hdgarrood

Issue #2349, opened 2016-10-01
[psc-ide] Collect docstrings for identifiers


kRITZCREEK:
Maybe we can reuse some of the code in `psc-docs`? I ran into a few problems the last time I tried.

I think it would be best to just do the simplest thing possible first and iterate on that.

kRITZCREEK:
Fixed in #3138

PR #3138, opened 2017-10-27
[purs ide] return documentation comments

nwolverson:
For #2349.

My first iteration used more docs code but I think this is simpler and works out better.

I'm thinking that markdown, plaintext and html options could be provided since there's already a markdown renderer there.

@kRITZCREEK

### DIFF
--- a/src/Language/PureScript/Ide/State.hs
+++ b/src/Language/PureScript/Ide/State.hs
@@ -44,6 +44,7 @@ import           Control.Lens                       hiding (op, (&))
 import           "monad-logger" Control.Monad.Logger
 import qualified Data.Map.Lazy                      as Map
 import qualified Language.PureScript                as P
+import           Language.PureScript.Docs.Convert.Single (convertComments)
 import           Language.PureScript.Externs
 import           Language.PureScript.Ide.Externs
 import           Language.PureScript.Ide.Reexports
@@ -199,6 +200,7 @@ populateVolatileStateSTM ref = do
         moduleDeclarations
         & map resolveDataConstructorsForModule
         & resolveLocations asts
+        & resolveDocumentation (map fst modules)
         & resolveInstances externs
         & resolveOperators
         & resolveReexports reexportRefs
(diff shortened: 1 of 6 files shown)


src/Language/PureScript/Ide/State.hs
```diff
@@ -246,6 +232,71 @@ resolveLocationsForModule (defs, types) decls =
...
+  convertDecl (IdeDeclarationAnn ann d) = 
+    convertDeclaration'
+      (annotateValue . P.IdentName)
+      (annotateValue . P.IdentName . P.Ident) 
+      (annotateValue . P.TyName . P.ProperName)
+      (annotateValue . P.KiName . P.ProperName)
+      d
+    where
+      docs :: P.Name -> Maybe Text
+      docs ident = convertComments =<< Map.lookup (ident) comments
```
kRITZCREEK: I'd prefer a `fromMaybe ""` at the end of this, so it's easier to tell whether or not we've already attempted to resolve documentation.

src/Language/PureScript/Ide/State.hs
```diff
@@ -246,6 +232,71 @@ resolveLocationsForModule (defs, types) decls =
...
+  Map.mapWithKey (\mn decls ->
+    maybe decls (flip resolveDocumentationForModule decls) (Map.lookup mn modules))
+
+resolveDocumentationForModule
+  :: P.Module
+    -> [IdeDeclarationAnn]
+    -> [IdeDeclarationAnn]
+resolveDocumentationForModule (P.Module _ _ _ sdecls _) decls = map convertDecl decls
+  where 
+  comments :: Map P.Name [P.Comment]
```
kRITZCREEK: I think we need the namespace here, to tell Kinds from Types from data constructors?

src/Language/PureScript/Ide/State.hs
nwolverson: So returning `""` instead of `null` in the json output if there is no docs? Maybe `Maybe Maybe _`?

kRITZCREEK: I thought about `Maybe Maybe` but at that point I'd prefer a custom three valued data type. Also note that `convertComments` returns `Nothing` if it sees an empty string.

src/Language/PureScript/Ide/State.hs
nwolverson: Is that not inherent in Name? `KiName | TyName | DctorName` etc?

src/Language/PureScript/Ide/State.hs
kRITZCREEK: Ahh, I see it now the `KiName` `TyName` ... stuff is from Docs. Sorry for being a little thick xD

While we're at it, there are redundant parens around `ident` in this line.

tests/Language/PureScript/Ide/CompletionSpec.hs
```diff
@@ -6,9 +6,12 @@ import Protolude
 
 import Language.PureScript as P
 import Language.PureScript.Ide.Completion
-import Language.PureScript.Ide.Test
+import Language.PureScript.Ide.Test as Test
+import           Language.PureScript.Ide.Command as Command
```
kRITZCREEK: The import formatting is inconsistent here, I don't mind leaving space for `qualified`, but let's apply that formatting on every import in this file then.

tests/Language/PureScript/Ide/CompletionSpec.hs
```diff
@@ -21,6 +24,33 @@ reexportMatches =
...
+
+load :: [Text] -> Command
+load = LoadSync . map Test.mn
+
+rebuildSync :: FilePath -> Command
+rebuildSync fp = RebuildSync ("src" </> fp) Nothing
+
+-- spec :: Spec
+-- spec = describe "Rebuilding single modules" $ do
+--     it "rebuilds a correct module without dependencies successfully" $ do
```
kRITZCREEK: I don't really mind the Type command comment, but this one should be deleted.

nwolverson: ugh I guess I was tired, such half-assery. thanks

kRITZCREEK:
Great, Thanks!

PR #3139, opened 2017-10-28
Fix #3131 Ide inserts conflicting imports for types

### LINKED ISSUE #3131: [purs ide] Ide inserts conflicting imports for types
Given:

```purescript
module Test where

import Data.Either (Either(..))
```

when adding an import for `Either` I expect:

```purescript
module Test where

import Data.Either (Either(..))
```

But I get:
```purescript
module Test where

import Data.Either (Either, Either(..))
```

nwolverson:


src/Language/PureScript/Ide/Imports.hs
```diff
@@ -220,6 +220,11 @@ addExplicitImport' decl moduleName qualifier imports =
         (insertDtor (dtor ^. ideDtorName))
         (refFromDeclaration d)
         refs
+    insertDeclIntoRefs (IdeDeclType t) refs
+      | Just _ <- find matches refs = refs
```
kRITZCREEK: well, if you used `any` instead of find you wouldn't need the pattern match on `Just`. Up to you, I don't really care.

nwolverson: Knew there had to be something, hoogle skills failed me here.

PR #3140, opened 2017-10-28
[purs ide] Collect usages

kRITZCREEK:
I figured I'd put this out, even though it's not finished, because I'm starting to hit a few design questions that might benefit from input. This will need extensive benchmarking, because in my current observations populating the usages takes about twice as long as all the other resolutions (That's still way faster than parsing, but something to keep an eye on, especially the memory usage). I also need to find a bigger codebase that compiles under current master.

For now, this only collects usages of values (no data constructors yet). Usages of types, classes, operators, and kinds can be added in following PRs.

I've added a new `query` command which returns all the information `ide` stores about a given declaration.

## Restrictions of the current implementation/design:

- It does not resolve ambiguous identifiers in the presence of multiple implicit imports

## Example usage:

Given this file:
https://gist.github.com/kRITZCREEK/02b3f49f6e4db01a939e9e188699b170

and this command (I hope you can see what's going on event hough it's elisp. I need to get jq to work on this machine...):
```elisp
(defun psc-ide-command-info (declarationId)
  (json-encode
   (list :command "info"
         :params (list
                  :id declarationId))))

(let-alist (psc-ide-unwrap-result
            (psc-ide-send-sync
             (psc-ide-command-info "Control.Monad.Eff.Console:V:log")))
  (print .ann.usages))
```

I get the following result:

```
[
[["Main"] ((start . [36 3]) (name . "C:\\Users\\creek\\code\\variant-experiments\\src\\Main.purs") (end . [36 6]))] 
[["Main"] ((start . [35 3]) (name . "C:\\Users\\creek\\code\\variant-experiments\\src\\Main.purs") (end . [35 6]))] 
[["Main"] ((start . [6 44]) (name . "C:\\Users\\creek\\code\\variant-experiments\\src\\Main.purs") (end . [6 47]))]
]
```

Running the find usage command for Prelude's `id`: https://gist.github.com/kRITZCREEK/be0a8c3ed38596a74aa1736f6be795e9

## TODO:

- [x] Make `resolveUsages` operate in STM
- [x] Figure out how to handle re-exports
  I see two options here:
  1. When inserting the found usages, follow the `exportedFrom` annotations, to add usages to the original declaration
  2. Store which modules reexport a given value in the annotation for that value and have the caller track them down at query time.
- [x] ~Find constructor usages~
- [x] Make the `info` command also take `filters` as arguments, and either turn the `IdeDeclarationId` into a filter, or build it by combining the `module`, `namespace`, `exact` filters. I think we might need to spot that combination on the server side and optimize it.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -232,6 +232,10 @@ getModuleName (Module _ _ name _ _) = name
 getModuleSourceSpan :: Module -> SourceSpan
 getModuleSourceSpan (Module ss _ _ _ _) = ss
 
+-- | Return a module's declarations
+getModuleDeclarations :: Module -> [Declaration]
+getModuleDeclarations (Module _ _ _ decls _) = decls
+
 -- |
 -- Add an import declaration for a module if it does not already explicitly import it.
 --
--- a/src/Language/PureScript/Ide.hs
+++ b/src/Language/PureScript/Ide.hs
@@ -57,12 +57,10 @@ handleCommand c = case c of
     findAvailableExterns >>= loadModulesSync
   LoadSync modules ->
     loadModulesSync modules
-  Info id -> do
-    getAtDeclarationId id >>= \case
-      Nothing ->
-        throwError (NotFound ("Couldn't find declaration for id: " <> printIdeDeclarationId id))
-      Just decl ->
-        pure (InfoResult decl)
+  Query filters currentModule -> do
+    modules <- getAllModules currentModule
+    let decls = applyFilters filters modules
+    pure (QueryResult decls)
   Type search filters currentModule ->
     findType search filters currentModule
   Complete filters matcher currentModule complOptions ->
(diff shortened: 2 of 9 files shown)


kRITZCREEK:
I decided to go with option

> When inserting the found usages, follow the exportedFrom annotations, to add usages to the original declaration

for handling reexports.

Data constructors are tricky, because using the `Maybe(..)` import syntax means I can't resolve a data constructor name solely based on the module I'm currently looking at. Before I start doing this resolution process I'd rather look into the renaming passes the compiler does to see if I can't reuse some of the machinery in there. So I propose to only do this for values for now, and add more and more find usage support over time.

kRITZCREEK:
Okay, I'm happy with what this does for now. I won't nail the `query` endpoint in stone yet, which is why I haven't documented it. Its format might change, but for now we can use it to try out find usages.

kRITZCREEK:
I've decided this needs some more work. Exposing `ide`'s internal data structure like this is too restrictive.

kRITZCREEK:
Superseded by #3206.

PR #3142, opened 2017-10-31
Make explicit import suggestions more consistent

nwolverson:
It's been annoying me that while either psc-ide or the suggestions for `ImplicitImport`/`HidingImport` will suggest `Blah(..)` for the import of a type and its data constructors, but other import suggestions always suggest the explicitly listed form `Blah(Blah1, Blah2,...)`, so that the suggested imports change even when they are not relevant to the warning.

e.g. when I have used `Just`
```
Module Foo has unspecified imports, consider using the explicit form:

    import Foo (Maybe(..))
```
vs
```
The import of module Foo contains the following unused references:

    zzz

  It could be replaced with:

    import Foo (Maybe(Just))
```

This change instead uses the `(..)` form if that was used previously, but still maintains the form with explicitly listed data constructors if present. So where I have an import ```import Foo (Either(..), Maybe(Just, Nothing), zzz)```

```
The import of module Foo contains the following unused references:

    zzz

  It could be replaced with:

    import Foo (Either(..), Maybe(Just))
```

(I'd like to test this, but as far as I know existing warnings tests are just testing the error code?)

### DIFF
--- a/src/Language/PureScript/Linter/Imports.hs
+++ b/src/Language/PureScript/Linter/Imports.hs
@@ -214,15 +214,17 @@ lintImportDecl env mni qualifierName names ss declType allowImplicit =
   checkImplicit warning =
     if null allRefs
     then unused
-    else warn (warning mni (map simplifyTypeRef allRefs))
-    where
-    -- Replace explicit type refs with data constructor lists from listing the
-    -- used constructors explicity `T(X, Y, [...])` to `T(..)` for suggestion
-    -- message.
-    simplifyTypeRef :: DeclarationRef -> DeclarationRef
-    simplifyTypeRef (TypeRef ss' name (Just dctors))
-      | not (null dctors) = TypeRef ss' name Nothing
-    simplifyTypeRef other = other
+    else warn (warning mni (map (simplifyTypeRef $ const True) allRefs))
+
+  -- Replace explicit type refs with data constructor lists from listing the
+  -- used constructors explicity `T(X, Y, [...])` to `T(..)` for suggestion
+  -- message.
+  -- Done everywhere when suggesting a completely new explicit imports list, otherwise
+  -- maintain the existing form.
+  simplifyTypeRef :: (ProperName 'TypeName -> Bool) -> DeclarationRef -> DeclarationRef
+  simplifyTypeRef shouldOpen (TypeRef ss' name (Just dctors))
+    | not (null dctors) && shouldOpen name = TypeRef ss' name Nothing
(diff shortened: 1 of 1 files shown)


nwolverson:
Uh. I guess I thought this got merged, not sure how it shows as having 32 commits. I guess this needs rebased?

hdgarrood:
I wonder if this commit history has something to do with the 0.12 branch... I'll try to work out what's going on with that and get back to you.

nwolverson:
Wasn't sure what happened with the 0.12 branch, seemed like some funny business - I can easily enough cherry pick the commits back on top of master. I guess it was ac18e53 80b04b0

nwolverson:
Actually seems there was only 1 commit for this PR, I think this should be back in a sensible state.

hdgarrood:
I tested this manually with the following input source files, comparing the behaviour on `master` with this PR: https://gist.github.com/hdgarrood/a0d2b03f6af2b73ba6c376e4be1aa4ab

This seems to work great; there's just one thing I noticed:
```purescript
module ImplicitQualifiedImport where

import Prelude
import Data.Array as M
import Data.Maybe as M

foo :: M.Maybe Unit
foo = M.Just unit
```
suggests using `import Data.Maybe (Maybe(Just)) as M`, whereas the `ImplicitImport` warning suggests using the `(..)` form; presumably these should both do the same thing too? I guess they should both suggest `(..)`?

garyb:
> I guess they should both suggest `(..)`?

👍

nwolverson:
Good spot, the ImplicitQualifiedImport suggestion should now be consistent with the others.

Something I did test before and I think was missing from that gist, is checking the "maintains the form with explicitly listed data constructors if present" - ie. the case of 
```
import Data.Maybe (Maybe(Just), fromMaybe)
```
when `Just` is used but not `fromMaybe`, the suggestion to remove `fromMaybe` should not alter `Maybe(Just)`, i.e. is
```
import Data.Maybe (Maybe(Just))
```
 Updated gist here https://gist.github.com/nwolverson/f94772a141c68178acb09b68f4d41cbd

LiamGoodacre:
(I'm also not familiar with this part of the code, but this looks good :+1:)

hdgarrood:
@garyb would you like to look over this? No rush, of course.

garyb:
Sure thing. I'm taking a mostly-complete break from code stuff just now, but will definitely take a look in the new year :)

kRITZCREEK:
I've built and tested this and it works nicely, thanks!

PR #3147, opened 2017-11-08
Set up a benchmarking project

kRITZCREEK:
I've spent some time setting up a possible approach to benchmarking parts of the compiler.

It uses psc-package to manage a project and criterion to organize and run the benchmarking code.

The benchmarks can be run with `stack bench` and all the different output formats criterion offers can be used. I already made a couple of interesting observations by playing around with this (eg. using the output of `psc-package sources` instead of a single Glob cost around 200ms in startup time and the ide loading code is around 2x faster with optimizations turned on).

Things that we might want to do after this:

- Write more benchmarks for interesting parts of the compiler pipeline
- See how we can spot regressions in CI by collecting statistics about running the benchmarks on master
- Look into using `weigh` to collect statistics about memory usage
- Set the benchmarks up in a way that makes it easy to start profiling a benchmark that regressed


hdgarrood:
Looks great! FWIW the glob performance issue should hopefully just disappear once we upgrade to the latest version of Glob (which includes my perf improvements).

kRITZCREEK:
Looking forward to that, especially now that we have a way of verifying it ;)

kRITZCREEK:
I've decided this is not the way to go. I'll see if I can use the psc-package setup here to get rid of bower in our test suite. For spotting performance regressions in the compiler I suspect just timing the full compilation process will be more effective.

hdgarrood:
On reflection I think bower is probably a better bet for the tests for the time being, unless psc-package gets a way of depending on a specific tag for certain packages without having to fork the package set.

PR #3148, opened 2017-11-08
correctly quote uppercased field labels

Thimoteus:
Example:

Source:

```purescript
module Main where

x :: _
x = {"Test": 3}
```

Current master:
```
Warning found:
in module Main
at Main.purs line 3, column 6 - line 3, column 7

    { Test :: Int
    }            
                 

in value declaration x
```

With this PR:
```
Warning found:
in module Main
at Main.purs line 3, column 6 - line 3, column 7

    { "Test" :: Int
    }              
                   

in value declaration x
```


garyb:
Oh, could you add yourself to the `CONTRIBUTORS` file please?

Thimoteus:
@garyb done!

hdgarrood:
Could you explain what is incorrect about the current behaviour? I don’t understand what the problem is.

paf31:
@hdgarrood I think the issue is that what is printed now is not actually a syntactically valid type.

Issue #3150, opened 2017-11-10
RFC: how about type-infos in CoreFn `annotation`s? (Happy to do a PR if this isn't unwanted)


metaleap:
Some might remember I'm attempting to transpile-PureScript-without-forking-purs. Irrelevant side-bar:

> First attempt was to do so via `--dump-coreimp` (#3117) however I've actually given up on that approach (ditch that PR if you like!) and now prefer to work with the existing `--dump-corefn`. (With the former I would have gotten some of `purs` later `Optimizer` passes "for free", but I decided against re-processing and transforming an-IR-that-is-essentially-JS --- too fickle/messy, too many special-cases, future maintenance nightmare etc.)

**What I'd like to propose here for CoreFn** is to fill in the (currently-ignored-via-`_`) type information argument in `annToJSON` right here: https://github.com/purescript/purescript/blob/0.12.0-dev/src/Language/PureScript/CoreFn/ToJSON.hs#L48

Are there any objections? I'm more than eager to prepare this PR but would like to gather opinions here first, maybe this violates some design goals of CoreFn or something?

Alternatively, could also introduce a second flag so you'd do either `--dump-corefn` (slim old format) or `--dump-corefn --dump-types` (same corefn but sporting extended-with-type-infos annotations).

Any of these 2 would really help to avoid having to do a lot of from-scratch type-inferencing for anyone aiming to use `corefn.json` files for actual _transpilation_ purposes (rather than troubleshooting / inspection / study etc) targeting *static-typed* languages. (Sure, one still needs to map these "PS-native" type-annotations to one's target environment, but better than inferring them from scratch. Why duplicate all the work `purs` just did a few moments prior?)

AFAICT the type info would actually be already readily there in-memory and pre-computed right when CoreFn is being dumped (or am I wrong? that's how I'm currently reading `purs` code here) --- just it's being `_`-ignored for now.

coot:
I have a [branch](https://github.com/coot/purescript/commits/corefn-typeclasses) which does that. Probably needs some small updates.

metaleap:
Cool, looks like we both did the obvious:

https://github.com/coot/purescript/commit/29ef583ae754ca963dc245e8a1b63445f1eeb90e#diff-33967ef4048aa250a3d3ee1332848271

https://github.com/metaleap/purs-with-dump-coreimp/commit/912fb47f2be426051796bdc6dae3c95082be9ae8

[...]

PR #3151, opened 2017-11-12
WIP: include types in CoreFn

coot:
Should we have a `--dump-types` switch as suggested by @metaleap in #3150.


paf31:
My 2¢: we shouldn't include type information until we can check them, which I think means we need to elaborate to something like System F, and then include explicit quantifiers so that every type variable is scoped.

metaleap:
> My 2¢: we shouldn't include type information until we can check them

To double-check my newb interpretation of the above: does this mean purs' type-checking happens *after* `--dump-corefn`?

paf31:
No, I'm referring to something like GHC's core linting, which is a separate type _checking_ pass (note: checking only, not inference).

metaleap:
> like GHC's core linting

Ah got it, I remember that one! Hmm, but why shouldn't we add the typed-core dumping now: once the linter happens later on, the results simply become "correct" automagically that much later on --- in other words, "something is more useful than nothing" for most-likely both @coot 's and my use-cases and whoever wants to experiment with "alt-transpilation" alongside ongoing developments. Seeing as the it's almost/mostly just the change from `_` to `toJSON t` in `ToJSON.hs`. Us few niche users of the `--dump` functionality can for now simply very reasonably be expected to assume all dumps invalid if their preceding `purs` run that dumped them errored out in the end.

(Just my 2c, but I don't feel too strongly about it, I'm equally fine with keeping this PR separately in my personal fork until then too --- my progress is very slow-going anyway currently)

paf31:
> Hmm, but why shouldn't we add the typed-core dumping now: once the linter happens later on, the results simply become "correct" automagically that much later on

Because we don't know they're correct, we don't have a spec. Also it's more code that needs to be maintained and tested. It's easy to add it now, and much harder to maintain it later.

I suggest adding this in a compiler fork, and answering those questions there first.

coot:
I agree, we need make corefn a typed checked language first.

PR #3156, opened 2017-11-20
Update Glob library, fixes #3055

### LINKED ISSUE #3055: purs bundle uses too much memory
I have a project which consists of ~650 modules, total size of the `output` directory is 19MB, roughly 100k lines of code in `output/*/*.js` files. purs bundle (0.11.6) and psc-bundle (0.10.5) both need more than 9GB of RAM to generate the bundle. This is far too much and the process is killed by the OOM killer on our CI/CD service.

hdgarrood:
Also update the Stackage snapshot so that we get the latest version of
Glob, which includes memory usage and performance fixes.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -52,7 +52,7 @@ dependencies:
   - file-embed
   - filepath
   - fsnotify >=0.2.1
-  - Glob >=0.7 && <0.9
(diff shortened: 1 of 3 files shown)


hdgarrood:
The CI failures aren't related so I think I'll just merge this.

PR #3160, opened 2017-11-24
Custom prelude, remove some Strings, fix stack repl.

jacereda:
I know it would be a PITA to review this, but I think it could provide some benefits.


hdgarrood:
Sorry, this is too big to review in one go. There are a few separate things going on here:

* switching from `String` to `Text`
* switching to a custom Prelude
* fixing stack repl
* renaming some modules
* using System.IO.UTF8 everywhere
* removing -O in the package.yaml file

These should all be separate PRs, unless there is a reason why they all need to go together. If these things cannot be separated, can you explain why? Also, it needs to be clearer why all of these changes are being made, e.g. with more detailed commit messages.

I would like to switch to a custom Prelude, but we should agree what should and shouldn't go into it first. In particular, if we are going to switch to a custom Prelude, I think we should drop the Protolude dependency at the same time.

jacereda:
I agree. This just started with a "let's see what happens" and turned into a huge thing. 

As to what goes and doesn't go into the custom prelude, I guess having a rough sketch is enough, as adding/removing things later is easy.

As to dropping Protolude, sounds good.

hdgarrood:
Ok, great, thanks for understanding :) I'll close this for now, then.

PR #3171, opened 2017-12-18
Minor cleanup

bjornars:
Some minor nitpicky cleanup in codegen/JS.hs. No change in behaviour.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -15,7 +15,7 @@ import Control.Monad.Except (MonadError, throwError)
 import Control.Monad.Reader (MonadReader, asks)
 import Control.Monad.Supply.Class
 
-import Data.List ((\\), delete, intersect)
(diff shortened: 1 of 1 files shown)


hdgarrood:
I'm generally not overly keen on these types of changes unless they have a very clear benefit, because they incur a risk of merge conflicts and they also make `git blame` a bit less useful. However, using `partition` instead of two uses of `\\` is clearly an improvement, so I'd be happy to merge that. I don't think the other change is clear enough of an improvement to justify merging, though.

bjornars:
Just trying to get my feet wet getting back into haskell :) 

I suppose there is a trade-off wrt the second commit, although I think there are clear readability-benefits to the change.

kRITZCREEK:
The `partition` usage is an improvement. The change using the infinite list is a cute trick, but the usage of `init` and `list ++ [singleElement]` indicates that a list is the wrong data structure for that operation anyway so I'd rather not merge that change.

Issue #3132, opened 2017-10-23
Disallow hiding superclasses of exported classes


metaleap:
Using the current `master` branch, compiling a purescript project containing these two modules [Mini.TCls3 and Mini.TClsImpl3](https://gist.github.com/metaleap/b7848fb180faaa8e59709679e6d36768) errors not with a description of what's wrong (if anything) in the source but rather:

    * Building project in /home/rox/c/ps/gonad-demos
    Compiling Mini.TCls3
    Compiling Mini.TClsImpl3
    purs: An internal error occurred during compilation: "newDictionaries: type class lookup failed"
    Please report this at https://github.com/purescript/purescript/issues
    CallStack (from HasCallStack):
      error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-0.11.6-8vLqjNaSDrU71uAVwdv7Dy:Language.PureScript.Crash
      internalError, called at src/Language/PureScript/TypeChecker/Entailment.hs:578:40 in purescript-0.11.6-8vLqjNaSDrU71uAVwdv7Dy:Language.PureScript.TypeChecker.Entailment
    purs: thread blocked indefinitely in an MVar operation
    * ERROR: Subcommand terminated with exit code 1

In this case `purs` (built from current `master`) was invoked via this `pulp` command-line:

    pulp build -- --source-maps  --comments

The crash no longer occurs when `class Zero, zilch, class One, ein, uno` are added to the exports of `Mini/TCls3.purs` (equivalently in this case to simply ditching the exports ie. exporting-everything-implicitly). But regardless, `purs` probably wants to error cleanly not crash-ly  =)

metaleap:
The crash no longer occurs when `class One, ein, uno` are added to the exports list. But regardless, `purs` probably wants to error cleanly not crash-ly  =)

parsonsmatt:
So, there's definitely a bug here. I've minimized the test case to #3173. You should be getting an `UnknownClass` error, because the instances for `Uno` and `Zero` aren't in scope because the classes are not exported.

When you provide a fully saturated instance, like `instance Dos Foo`, then it works fine -- the solver is able to find the nullary instance somehow and not error out.

When you provide an instance with a context that depends on an out-of-scope instance, then you get the thread-blocked-on-mvar error. Eg:

```purescript
instance inst3 :: (C2 a) => C3 a B
```

The `C2` instance in the context *also* carries a `C1` instance requirement.

parsonsmatt:
[...]

PR #3173, opened 2017-12-19
Fix error on missing type class name with non-exported superclass (#3132)

parsonsmatt:
This PR adds some test cases for #3132 and might even have a fix :smile: 

Any call to `internalError` will trigger an `MVar` bug. That should be fixed too. But the real underlying bug is that superclasses must be exported.


tests/TestCompiler.hs
```diff
@@ -227,7 +227,9 @@ checkShouldFailWith expected errs =
   let actual = map P.errorCode $ P.runMultipleErrors errs
   in if sort expected == sort (map T.unpack actual)
     then Nothing
-    else Just $ "Expected these errors: " ++ show expected ++ ", but got these: " ++ show actual
+    else Just $ "Expected these errors: " ++ show expected ++ ", but got these: " 
+      ++ show actual ++ ", full error messages: " 
+      ++ unlines (map (P.renderBox . P.prettyPrintSingleError P.defaultPPEOptions) (P.runMultipleErrors errs))
```
parsonsmatt: This modification to the reporting made it much easier to debug exactly what was going on with a mistaken error code.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -568,14 +568,18 @@ matches deps TypeClassDictionaryInScope{..} tys =
...
   -> Constraint
   -> m [NamedDict]
 newDictionaries path name (Constraint className instanceTy _) = do
     tcs <- gets (typeClasses . checkEnv)
-    let TypeClassData{..} = fromMaybe (internalError "newDictionaries: type class lookup failed") $ M.lookup className tcs
+    TypeClassData{..} <-
+      maybe
+        (throwError . errorMessage . UnknownClass $ className)
+        pure
+        $ M.lookup className tcs
```
parsonsmatt: If there's a more suitable idiom, I'd be happy to give it a shot :slightly_smiling_face:

examples/failing/3132.purs
```diff
@@ -0,0 +1,18 @@
...
+import Control.Monad.Eff (Eff)
+import Control.Monad.Eff.Console (CONSOLE, log)
+
+import ClassDefinitions
+
+data A
+data B
+
+instance inst2 :: C2 A
+instance inst3 :: (C2 a) => C3 a B
```
parsonsmatt: It seems like the effect of having a "hidden" nullary class/instance is to prevent polymorphic instances like this. Note that `C2 A` is totally fine, despite `C1` not being in scope, and `C3 A B` is *also* totally fine. But requiring `C2 a => C3 a B` causes the missing name error.

hdgarrood: Hm, I'm not sure. It doesn't seem right to me that this code would cause an unknown class error from `C1` not being in scope, and it also doesn't seem right to me that hidden nullary classes can prevent you from writing instances like this.

My suggestion is that we say this code should actually compile successfully, and in fact we need to modify the compiler to make sure that hidden classes like `C1` are made available in the relevant places in the compiler to make sure that internal error doesn't get hit.

tests/TestCompiler.hs
hdgarrood: I'd be in favour of merging this in a separate PR.

parsonsmatt: Cool, I'll factor it out then.

examples/failing/3132.purs
garyb: I think the real problem with all this, is that the `ClassDefinitions` module should demand that you export `C1` because it's required by `C2`. That would be the rule we use for all the other cases for this kind of thing.

parsonsmatt:
So this PR fixes the `MVar` bug and properly reports the missing class constraint. It appears this has also uncovered some behavior that we want to modify, with two suggestions:

1. > My suggestion is that we say this code should actually compile successfully, and in fact we need to modify the compiler to make sure that hidden classes like C1 are made available in the relevant places in the compiler to make sure that internal error doesn't get hit.

2. > I think the real problem with all this, is that the ClassDefinitions module should demand that you export C1 because it's required by C2. That would be the rule we use for all the other cases for this kind of thing.

hdgarrood:
I think suggestion 2 (@garyb's) makes the most sense.

parsonsmatt:
Do we want to merge this as-is, and create another issue/PR for that, or work that change into this PR? I think that's going to require a good bit more work

hdgarrood:
I think I'd say that the MVar bug is actually within Make, in that any use of `error` in one of the threads in there manifests as an error whose message implies that there's a bug in the threading code when there isn't. That is, I don't think I'd say that this PR fixes the MVar bug.

As far as I can see, I think it's correct that we use `internalError` in TypeChecker.Entailment there, because if we go down that code path, it means there's a bug somewhere else in the compiler. Therefore, I think we want this to manifest as an internal compiler error (as it does now) rather than a regular compiler error.

Given this, I think the best course of action is to not merge this, and instead to expand the the TransitiveExportError check to include superclass constraints.

parsonsmatt:
Cool, I think I can get that implemented :smile: Thanks for the feedback!

hdgarrood:
Of course, thank you for all your work on this!

parsonsmatt:
I'm actually planning on keeping the work on this branch; mind re-opening?

examples/failing/3132.purs
```diff
@@ -0,0 +1,18 @@
+-- @shouldFailWith TransitiveExportError
```
parsonsmatt: Updated the reason this should fail.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -803,7 +803,7 @@ data Expr
   --
   | AnonymousArgument
   -- |
-  -- A typed hole that will be turned into a hint/error duing typechecking
+  -- A typed hole that will be turned into a hint/error during typechecking
```
parsonsmatt: Minor typo fix

src/Language/PureScript/Environment.hs
```diff
@@ -37,7 +37,9 @@ data Environment = Environment
   , typeSynonyms :: M.Map (Qualified (ProperName 'TypeName)) ([(Text, Maybe Kind)], Type)
   -- ^ Type synonyms currently in scope
   , typeClassDictionaries :: M.Map (Maybe ModuleName) (M.Map (Qualified (ProperName 'ClassName)) (M.Map (Qualified Ident) NamedDict))
-  -- ^ Available type class dictionaries
+  -- ^ Available type class dictionaries. When looking up 'Nothing' in the
+  -- outer map, this returns the map of type class dictionaries in local
+  -- scope (ie dictionaries brought in by a constrained type).
```
parsonsmatt: Updating this based on comments in Slack

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,93 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
+      -- transitive class dependencies that are defined in this
+      -- module.
+      transitiveSuperClassesFor
+          :: Qualified (ProperName 'ClassName)
+          -> S.Set (Qualified (ProperName 'ClassName))
+      transitiveSuperClassesFor qname =
+        S.intersection moduleClassDeclarationSet
+          $ untilSame 
+            (\s -> s <> foldMap (\n -> fromMaybe S.empty (M.lookup n classesToSuperClasses)) s)
+            (fromMaybe S.empty (M.lookup qname classesToSuperClasses))
```
parsonsmatt: This is going to do a lot of recomputation to find the fixed point of that set.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,93 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
-        case M.lookup (Qualified (Just mn) dctor) (dataConstructors env) of
-          Nothing -> return ()
-          Just (_, _, ty, _) -> checkExport dr extract ty
-    return ()
+    for_ (M.lookup (qualify' name) (typeSynonyms env)) $ \(_, ty) ->
+      checkExport dr extract ty
+    for_ dctors $ \dctors' ->
+      for_ dctors' $ \dctor ->
+        for_ (M.lookup (qualify' dctor) (dataConstructors env)) $ \(_, _, ty, _) -> 
+          checkExport dr extract ty
```
parsonsmatt: The inconsistent use of `for_` and `case` made this hard to parse, so I simplified. I can revert if wanted.

src/Language/PureScript/Types.hs
```diff
@@ -291,7 +291,7 @@ everythingOnTypes :: (r -> r -> r) -> (Type -> r) -> Type -> r
 everythingOnTypes (<+>) f = go where
   go t@(TypeApp t1 t2) = f t <+> go t1 <+> go t2
   go t@(ForAll _ ty _) = f t <+> go ty
-  go t@(ConstrainedType c ty) = foldl (<+>) (f t) (map go (constraintArgs c)) <+> go ty
+  go t@(ConstrainedType c ty) = foldl' (<+>) (f t) (map go (constraintArgs c)) <+> go ty
```
parsonsmatt: Maybe unnecessary, but it can't hurt to be sure.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,89 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
+      moduleClassDeclarations = 
+        M.fromList
+          . mapMaybe (\x -> case x of
+                td@(TypeClassDeclaration _ name _ _ _ _) ->
+                  Just (Qualified (Just mn) name, classDecToRef td)
+                _ ->
+                  Nothing)
+          $ decls
+      moduleClassDeclarationSet :: S.Set (Qualified (ProperName 'ClassName))
+      moduleClassDeclarationSet =
```
kRITZCREEK: Could just use `Data.Map.keysSet` here

src/Language/PureScript/TypeChecker.hs
hdgarrood: I suppose you could argue that it's not necessary to obtain transitive superclasses, only immediate ones, as I think this check will fail in the transitive case if and only if it fails in the immediate case. Obtaining transitive dependencies just means we get a more useful error if it does fail, I think?

E.g. If we're exporting `C3` here, we also need to export `C2` and `C1`:

```purescript
module X (class C3) where
class C1 a
class (C1 a) <= C2 a
class (C2 a) <= C3 a
```

And whether we obtain transitive dependencies here or not, we're going to fail with a transitive export error; obtaining transitive superclasses presumably just means that we have a more useful error, which lists both `C2` and `C1` as opposed to only `C2`.

Perhaps as a middle ground we could only obtain immediate superclasses in the first pass, and then if there are any errors, obtain transitive superclasses for the offending classes for the sake of a more detailed error? Does that sound feasible?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -490,12 +552,12 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
   -- been exported from the module
   checkClassesAreExported :: DeclarationRef -> m ()
-  checkClassesAreExported ref = checkMemberExport findClasses ref
+  checkClassesAreExported ref = checkMemberExport findClasses  ref
     where
     findClasses :: Type -> [DeclarationRef]
     findClasses = everythingOnTypes (++) go
       where
-      go (ConstrainedType c _) = (fmap (TypeClassRef (declRefSourceSpan ref)) . extractCurrentModuleClass . constraintClass) c
+      go (ConstrainedType c _) = (fmap (TypeClassRef (declRefSourceSpan ref)) . extractCurrentModuleClass . constraintClass) c
```
hdgarrood: Trailing space here

examples/failing/3132.purs
hdgarrood: Can you paste the full error you get when you attempt to compile this?

src/Language/PureScript/TypeChecker.hs
parsonsmatt: That is a really good point -- we're already checking every exported item in the list, so we only need to verify that the immediate superclasses (also defined in the module) are exported. The transitive superclasses defined in module would be caught on when they're being inspected :smile: 

You're correct that this method does provide a nicer error message, in that you'll get one error message telling you to add all the things to the exports.

We can get "best of both worlds" if we can catch the error and add the lookup for transitive superclasses there before rethrowing.

src/Language/PureScript/Types.hs
parsonsmatt: Sure thing :slightly_smiling_face:

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -568,7 +568,7 @@ matches deps TypeClassDictionaryInScope{..} tys =
 -- | Add a dictionary for the constraint to the scope, and dictionaries
 -- for all implied superclass instances.
 newDictionaries
-  :: MonadState CheckState m
+  :: (MonadState CheckState m, MonadError MultipleErrors m)
```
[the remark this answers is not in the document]
parsonsmatt: Yes!

examples/failing/3132.purs
parsonsmatt:
```
Failing examples
  '3132.purs' should fail with 'UnknownName' FAILED [1]

Failures:

  tests/TestCompiler.hs:218:
  1) Failing examples '3132.purs' should fail with 'UnknownName'
       Expected these errors: ["UnknownName"], but got these: ["TransitiveExportError"], full error messages:
       in module Main
       at /home/matt/Projects/purescript/examples/failing/3132.purs line 1, column 1 - line 18, column 7

         An export for class C3 requires the following to also be exported:

           class C1
           class C2



       See https://github.com/purescript/documentation/blob/master/errors/TransitiveExportError.md for more information,
```

hdgarrood: Great, thank you. I was just wondering. :)

parsonsmatt: "Does this fail with the right type of error, but also the right *reason* that I am expecting?" is definitely a good thing to check :smile:

parsonsmatt:
Only grabbing transitive dependencies on rethrowing the error:

```
  1) Failing examples '3132.purs' should fail with 'UnknownName'
       Expected these errors: ["UnknownName"], but got these: ["TransitiveExportError"], full error messages:
       in module Main
       at /home/matt/Projects/purescript/examples/failing/3132.purs line 1, column 1 - line 18, column 7

         An export for class C3 requires the following to also be exported:

           class C1
           class C2



       See https://github.com/purescript/documentation/blob/master/errors/TransitiveExportError.md for more information,
```

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,116 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
       checkClassMembersAreExported e
       checkClassesAreExported e
+      checkSuperClassExport superClassesFor moduleClassDeclarations e
+        `catchError` annotateSuperClasses transitiveSuperClassesFor moduleClassDeclarations
     return $ Module ss coms mn decls' (Just exps)
   where
+  qualify' :: a -> Qualified a
+  qualify' = Qualified (Just mn)
+
+  annotateSuperClasses
```
hdgarrood: Could you please add a comment summarising why we only obtain transitive superclasses after the immediate check has failed?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -490,7 +575,7 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
   -- Check that all the classes defined in the current module that appear in member types have also
   -- been exported from the module
   checkClassesAreExported :: DeclarationRef -> m ()
-  checkClassesAreExported ref = checkMemberExport findClasses ref
+  checkClassesAreExported ref = checkMemberExport findClasses  ref
```
hdgarrood: Sorry to nitpick but would you mind reverting this? I'm just trying to keep `git blame` useful where we can.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,116 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
+    -> M.Map (Qualified (ProperName 'ClassName)) DeclarationRef
+    -> DeclarationRef
+    -> m ()
+  checkSuperClassExport superClassesFor classMap dr@(TypeClassRef _ className) = do
+    let superClasses = superClassesFor (qualify' className)
+        unexportedClasses = S.difference superClasses moduleClassExports
+    unless (null unexportedClasses)
+      . throwError . errorMessage' (declRefSourceSpan dr)
+      . TransitiveExportError dr
+      $ mapMaybe (\n -> M.lookup n classMap) (toList unexportedClasses)
```
hdgarrood: It seems to me that there's a bit of a code smell here as we chuck away the second argument of `TransitiveExportError` here almost immediately, i.e. once we reach `annotateSuperClasses`. Have you considered consolidating `annotateSuperClasses` and this into a single function?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,116 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
   warnAndRethrow (addHint (ErrorInModule mn)) $ do
     modify (\s -> s { checkCurrentModule = Just mn })
     decls' <- typeCheckAll mn exps decls
+    classesToSuperClasses <- gets
+      ( M.map
+        ( S.fromList
+        . fmap constraintClass
```
hdgarrood: Rather than performing the intersections below to make sure we only keep local superclasses*, can we do a `filter` here which inspects the `Constraint` values and only keeps the superclasses which have been defined in the current module? That way, if I understand correctly, we won't need to construct the `moduleClassDeclarations` map (which saves us from iterating over all of the declarations in the module).

\*that is the reason for the intersections below, right?

src/Language/PureScript/TypeChecker.hs
parsonsmatt: Not a problem at all, I should have caught it :)

hdgarrood:
I also just noticed that the source span for the error is a bit dodgy - it seems to cover almost the entire file! I guess ideally it would only point to the part of the export list which is exporting the offending class.

hdgarrood:
Oh, maybe that's a problem in the part of the compiler that elaborates exports though, so perhaps best not to worry about that here.

hdgarrood:
Yeah, the same thing happens with the other types of `TransitiveExportError` currently, so let's address that separately.

src/Language/PureScript/TypeChecker.hs
parsonsmatt: okay, that's a much better idea :joy:

src/Language/PureScript/TypeChecker.hs
parsonsmatt: Great catch. Merged.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,98 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
+        -- thanks to laziness, the computation of the transitive
+        -- superclasses defined in-module will only occur if we actually
+        -- throw the error. Constructing the full set of transitive
+        -- superclasses is likely to be costly for every single term.
+        transitiveSuperClasses = transitiveSuperClassesFor (qualify' className)
+        unexported = S.difference superClasses moduleClassExports
+    unless (null unexported)
+      . throwError . errorMessage' (declRefSourceSpan dr)
+      . TransitiveExportError dr
+      $ mapMaybe (\n -> M.lookup n classMap) (toList transitiveSuperClasses)
```
hdgarrood: I think we can get rid of the `classMap` argument here; the `DeclarationRef` values in the second argument to `TransitiveExportError` aren't intended to have real source spans (since they are things which are supposed to be in the exports list but aren't), so I think we can just reuse the one from `dr`. That is what the other `TransitiveExportError` checks seem to be doing, too.

Do you think it would work to use `map (TypeClassRef (declRefSourceSpan dr))` on this line instead of `mapMaybe (\n -> M.lookup n classMap)`?

parsonsmatt: Great call :) That cut out a lot of code.

hdgarrood:
Great! Sorry to keep making suggestions, but now that you've been able to remove all that code I have just one last one: I quite like how `typeCheckModule` reads for the bit until you get to the first `where`, and I think it would be nice to preserve that. Could you perhaps extract all the code you've added there into a separate function, so that we end up with something like this?

```haskell
   warnAndRethrow (addHint (ErrorInModule mn)) $ do
     modify (\s -> s { checkCurrentModule = Just mn })
     decls' <- typeCheckAll mn exps decls
     checkSuperClassesAreExported <- getSuperClassExportCheck
     for_ exps $ \e -> do
       checkTypesAreExported e
       checkClassMembersAreExported e
       checkClassesAreExported e
       checkSuperClassesAreExported e
     return $ Module ss coms mn decls' (Just exps)
```

parsonsmatt:
I really appreciate the feedback and suggestions :smile:

src/Language/PureScript/TypeChecker.hs
```diff
@@ -434,31 +434,87 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
...
     return $ Module ss coms mn decls' (Just exps)
   where
+  qualify' :: a -> Qualified a
+  qualify' = Qualified (Just mn)
+
+  getSuperClassExportCheck = do
+    classesToSuperClasses <- gets
+      ( M.map
+        ( S.fromList
+        . filter (\(Qualified mn' _) -> mn' == Just mn)
```
kRITZCREEK: We have a helper for this: `filter (isQualifiedWith mn)`

hdgarrood:
Thanks @kRITZCREEK! I think I'll just merge this now to make sure it doesn't rot. Maybe I (or anyone who feels like it) could look through the compiler to find places we can use `isQualifiedWith`; I imagine this won't be the only one.

PR #3174, opened 2017-12-19
Add position information to type warning

b123400:


hdgarrood:
Could you please explain how you have tested this?

kRITZCREEK:
Yeah, it would be nice if you could come up with an example, show the bad error message on current master and then the improved one after your change. Thanks!

b123400:
Oops sorry I was in a hurry yesterday. For example when I have code like this:

```
main :: forall a. Int
main = 1
```

It will compile with warning:

>   in module Main
>  Type variable e is ambiguous, since it is unused in the polymorphic type which introduces it.
> in type declaration for main

After this PR, the warning becomes:

> in module Main
> at src/Main.purs line 3, column 1 - line 3, column 22
> Type variable e is ambiguous, since it is unused in the polymorphic type which introduces it.
> in type declaration for main

It's especially useful for IDE integration.

Issue #3114, opened 2017-10-14
Inference regression in master


natefaubion:
I actually don't know how to title this issue, so please edit it to make it more appropriate.

`purescript-variant`s tests do not compile on master anymore, and it's throwing me for a loop. In particular, using `#` with `on` chaining results in an inexplicable error:

```
[1/1 TypesDoNotUnify] test/VariantF.purs:43:13

                  v
  43      case1 = case_
  44        # on _foo (\a → "foo: " <> show a)
  45        # on _bar (\a → "bar: " <> show a)
  46        # on _baz (\a → "baz: " <> show a)
                                             ^

  Could not match type

    Array

  with type

    Tuple String

  while trying to match type Array
    with type Tuple String
  while trying to match type FProxy Array
    with type FProxy (Tuple String)
  while trying to match type ( baz :: FProxy t16
                             , bar :: FProxy Array
                             , foo :: FProxy Array
                             )
    with type ( foo :: FProxy Maybe
              , bar :: FProxy (Tuple String)
              , baz :: FProxy (Either String)
              )
  while solving type class constraint

    Prim.RowCons "baz"
                 (FProxy t16)
                 t10
                 ( foo :: FProxy Maybe
                 , bar :: FProxy (Tuple String)
                 , baz :: FProxy (Either String)
                 )
```

The rest of the verbose errors are unprintable due to very strange formatting. For reference, here is the test code https://github.com/natefaubion/purescript-variant/blob/master/test/VariantF.purs#L42-L46

*   I do not use or mention `Array` anywhere in the variant codebase, no literals, nothing.
*   If I don't use `#` syntax it works (either parens or `$` instead).
*   I have made no changes to the code (compiles fine on 0.11.6)

parsonsmatt:
I'm unable to reproduce this on current `purescript` master and current `purescript-variant` master. Are you still seeing this?

natefaubion:
On master or the 0.12 branch?

natefaubion:
Probably warrants a title edit since we split off the 0.12 branch.

parsonsmatt:
Oh, I misread -- `purescript-variant`'s master branch, not `purescript`'s master :sweat_smile:

natefaubion:
It _was_ purescript master when this issue was written :)

parsonsmatt:
I just checked out `0.12.0-dev`, built the compiler, and ran the test suite for `purescript-variant`. I'm not able to reproduce :confused:

[...]

PR #3175, opened 2017-12-20
Add test cases for #3114

parsonsmatt:
This PR adds a minimal test case for #3114.

### DIFF
--- a/examples/passing/3114.purs
+++ b/examples/passing/3114.purs
@@ -13,7 +13,6 @@ import Data.Symbol
 type TestVariants =
   ( foo :: FProxy Maybe
   , bar :: FProxy (Tuple String)
-  , baz :: FProxy (Either String)
   )
 
 _foo :: SProxy "foo"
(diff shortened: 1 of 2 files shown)


kRITZCREEK:
Should we maybe remove the annotations and make this a `pending` test? In its current state this PR introduces technical debt, because we need to remember to remove annotations and comment when the underlying issue was fixed.

hdgarrood:
My understanding is that the underlying issue has already been fixed, and this test does fail between the two commits that @parsonsmatt identified as being the commits that introduced and fixed #3114, so it shouldn't need further modification.

parsonsmatt:
Right, this is a regression test for a bug that was introduced and then accidentally fixed.

hdgarrood:
Shall we go ahead and merge this then?

kRITZCREEK:
I guess with the issue number attached we have enough context to understand the comment. 👍 from me

PR #3176, opened 2017-12-20
Rename prim classes

### LINKED ISSUE #2903: Rename classes in Prim
We'll probably end up adding more automatically-solved classes after 1.0, so I think it makes sense to move them into submodules of `Prim` before 1.0, and not import those modules automatically. 

For example, `Union` and `RowCons` could move into `Prim.Rows` (and we'd rename `RowCons` to just `Cons` while we're at it, probably).

I'll put this under 1.0 for now.

parsonsmatt:
This PR renames Prim classes and creates submodules for them to live in.

Todo:

- [x] Move `TypeConcat` and `TypeStr` into `Prim.TypeError`
- [x] Remove `Prim` submodules from the default environment
- [x] Make the `Prim` submodules available for import.
- [x] Make it so that importing a Prim submodule actually imports the terms
- [x] Make it so that importing a `Prim` submodule *only* imports the terms that are actually imported.
- [x] Account for the changes in `Language.PureScript.Ide.Prim`
- [x] Make PRs to relevant libraries.

Fixes #2903


src/Language/PureScript/Docs/Prim.hs
```diff
@@ -35,6 +31,29 @@ primDocsModule = Module
   , modReExports = []
   }
 
+primRowDocsModule :: Module
+primRowDocsModule = Module
+  { modName = P.moduleNameFromString "Prim.Row"
+  , modComments = Just "The Prim.Row module is embedded in the PureScript compiler. Unlike `Prim`, it is not imported implicitly. It contains automatically solved classes for workign with row types."
```
natefaubion: Typo: workign

parsonsmatt:
This is going to break CI until relevant libraries are updated. What's the protocol on that?

kRITZCREEK:
While you're at it, would you also take a look at `Language.PureScript.Ide.Prim`? If it's not clear what's happening don't worry about it and I'll fix it.

hdgarrood:
I think having them explicitly imported is perhaps best; I don't think these classes are quite ubiquitous enough to justify being imported by default. Having to import them explicitly was what Phil suggested originally too. What do others think?

hdgarrood:
Actually following from my other comment (https://github.com/purescript/purescript/issues/2903#issuecomment-353045477), having thought about it a little more, I think the best approach would be to move all of `Fail`, `Warn`, `TypeConcat` and `TypeString` to `Prim.TypeErrors`.

I think the idea is that we decide that pretty much everything apart from the most ubiquitous stuff belongs in submodules of `Prim`, and only `Prim` is imported automatically. Also, `Prim` itself should change very rarely. This gives us the freedom to be able to add more kinds/types/classes into submodules of `Prim` in the future without worrying about breaking people's code (e.g. if someone has already created a class called `Warn`, we want them to be able to upgrade the compiler to a new version including a compiler-magic class of the same name without having to change their imports).

hdgarrood:
I think what we've done in the past with CI is to change `tests/support/bower.json` to point at a branch of the libraries which need to change during development. Then, just before we make the compiler release, we publish new versions of all the core libraries, update `tests/support/bower.json` here, and check it all still works.

We might have also used release candidates for libraries so that we can publish versions which work with upcoming compiler releases but still have people get the versions which work with current compiler releases when they simply run `bower install purescript-whatever`.

I can't remember which of these two approaches tends to work better though; @garyb?

parsonsmatt:
@kRITZCREEK I'll give that a peek, thanks for the suggestion :smile: 

@hdgarrood Cool, I'll make the requisite changes in the downstream libraries and remove the non-`Prim` classes from the default environment.

That leads to another question: How do I make a module available without it being explicitly defined?

hdgarrood:
I think putting it in the `Environment` like we already have done for `Prim` should do that; IIRC there's some separate code during the desugaring phase which injects a `Prim` import for modules which don't already have an explicit one.

parsonsmatt:
Right now the tests all pass locally for me, because `Prim.Row.Cons` is in scope thanks to it being in the `primTypes` in the environment. I may be mistaken though

kRITZCREEK:
I know one instance of `Prim` being added is inside `Language.PureScript.Make.make`:

`buildModule barriers (importPrim m) (deps `inOrderOf` map getModuleName sorted)`

hdgarrood:
Ah right, okay; in that case, I'm probably wrong, and I'm not quite sure how to do it.

parsonsmatt:
I figured out how to make "virtual modules" but I'm not sure how to get the imports actually working yet.

src/Language/PureScript/Sugar/Names/Env.hs
```diff
@@ -182,22 +182,52 @@ envModuleExports (_, _, exps) = exps
...
   mkKindEntry (Qualified mn name) = (name, fromJust mn)
 
--- | Environment which only contains the Prim module.
+-- | Environment which only contains the Prim modules.
 primEnv :: Env
-primEnv = M.singleton
-  (ModuleName [ProperName "Prim"])
-  (internalModuleSourceSpan "<Prim>", nullImports, primExports)
+primEnv = M.fromList
+  [ ( ModuleName [ProperName "Prim"]
```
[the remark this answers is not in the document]
parsonsmatt: Definitely! Thanks for the tip :)

src/Language/PureScript/Sugar/Names/Imports.hs
```diff
@@ -80,6 +87,22 @@ resolveModuleImport env ie (mn, imps) = foldM go ie imps
...
+
+augmentEnvironment 
+  :: M.Map (Qualified (ProperName 'TypeName)) (Kind, TypeKind)
+  -> M.Map (Qualified (ProperName 'ClassName)) TypeClassData
+  -> Environment
+  -> Environment
+augmentEnvironment newTypes classes env = env
+  { typeClasses = typeClasses env `M.union` classes
+  , types = types env `M.union` newTypes
+  }
```
parsonsmatt: This is where the magic happens. All of the `MonadState Environment` plumbing is necessary so that we can actually bring the `Prim.*` stuff into the `Environment` if we happen to import those modules.

src/Language/PureScript/Sugar/Names/Imports.hs
```diff
@@ -68,6 +73,8 @@ resolveModuleImport env ie (mn, imps) = foldM go ie imps
      -> (SourceSpan, Maybe ImportDeclarationType, Maybe ModuleName)
      -> m Imports
   go ie' (ss, typ, impQual) = do
+    when (mn == C.PrimTypeError) (modify addPrimTypeErrorToEnvironment)
+    when (mn == C.PrimRow) (modify addPrimRowToEnvironment)
```
parsonsmatt: This is wrong, because it adds the entire `Prim.Row` environment to the environment whenever import anything at all from `Prim.Row`. The correct behavior would be to only add the items that are actually imported.

hdgarrood:
Anything to make this less ad-hoc sounds sensible; can you expand on how a `VirtualModule` type might work?

Something seems a little off to me about the changes involving the `MonadState Environment` instance; don't we already modify the `Environment` based on how `Prim` has been imported?

I think it might be time to bring in @garyb to help review this.

parsonsmatt:
As far as I can tell, none of the imports *actually* change the `Environment`. All of the `Environment` changing stuff is done while working with the actual `Module` data. For `Virtual` modules, like `Prim`, the answer was just "this is what the environment starts off with." When we say "There are virtual modules (aka PureScript knows about them, but they don't exist as a file or an externs)", they aren't represented as actual `Module` values, and so aren't considered in the export/import business. So the import desugaring phase, having access to an `Env`, is able to verify that the virtual modules exist and have the explorts that they claim to have. But it never modifies or alters the `Environment` itself.

A `VirtualModule` would be something like a `Module`, in that it has all the same stuff, but it doesn't exist as a file anywhere, or as actual code. `Prim`, `Prim.Row`, etc. would all be virtual modules. Instead of hardcoding the `Environment` with the `Prim` values by default, the system would make the set of virtual modules available for import like the ordinary modules. Then the `addPrimImport` function would add the implicit import, and *that* would be where the terms are added to the environment.

matthewleon:
This PR is probably a wonderful opportunity to fix a lot of the outdated doc strings living in  `Language.PureScript.Environment`

Some examples:

https://github.com/parsonsmatt/purescript/blob/535a4086606e081c0cf299654280d2ced16da825/src/Language/PureScript/Environment.hs#L95

https://github.com/parsonsmatt/purescript/blob/535a4086606e081c0cf299654280d2ced16da825/src/Language/PureScript/Environment.hs#L337

https://github.com/parsonsmatt/purescript/blob/535a4086606e081c0cf299654280d2ced16da825/src/Language/PureScript/Environment.hs#L370

src/Language/PureScript/Ide/Prim.hs
```diff
@@ -11,10 +11,10 @@ idePrimDeclarations :: [IdeDeclarationAnn]
 idePrimDeclarations =
   primTypes <> primKinds <> primClasses
   where
-    primTypes = foreach (Map.toList PEnv.primTypes) $ \(tn, (kind, _)) ->
+    primTypes = foreach (Map.toList PEnv.allPrimTypes) $ \(tn, (kind, _)) ->
       IdeDeclarationAnn emptyAnn (IdeDeclType (IdeType (P.disqualify tn) kind []))
     primKinds = foreach (Set.toList PEnv.primKinds) $ \kn ->
       IdeDeclarationAnn emptyAnn (IdeDeclKind (P.disqualify kn))
-    primClasses = foreach (Map.toList PEnv.primClasses) $ \(cn, _) ->
+    primClasses = foreach (Map.toList PEnv.allPrimClasses) $ \(cn, _) ->
```
parsonsmatt: @kRITZCREEK does this look right to you?

kRITZCREEK: I think the code inserts all of these declarations under the `Prim` module, if these declarations are now spread across different modules we'll need to account for that.

For example, in the current code:

```haskell
findType search filters currentModule = do
  modules <- Map.toList <$> getAllModules currentModule
  let insertPrim = (:) (C.Prim, idePrimDeclarations)
  pure (CompletionResult (getExactCompletions search filters (insertPrim modules)))
```

parsonsmatt: Ok, neat, I can handle it from there. Thanks!

parsonsmatt:
I think the last thing to do here is to open up PRs to all the `purescript/purescript-*` libraries that use the `RowCons`, `Union`, `TypeConcat`, `Fail`, `Warn`, etc. types and include the explicit imports. Everything is "done" compiler-side.

hdgarrood:
What about your previous comment, has that been addressed?

> This is wrong, because it adds the entire Prim.Row environment to the environment whenever import anything at all from Prim.Row. The correct behavior would be to only add the items that are actually imported.

For now I'll just take a bit more time to think over this, but the new `MonadState Environment` plumbing feels a little wrong somehow. How come we've not needed to do anything like that for regular imports, for instance?

parsonsmatt:
> What about your previous comment, has that been addressed?

You ever think something needs to be done, and then write a test for the feature, and it just passes the first time? [That's what happened here](https://github.com/purescript/purescript/pull/3176/commits/1fecbd1ba7c0df5e0408b45871d9a7a37f731105) :man_shrugging:

hdgarrood:
Ah ok, so is it now clear why that worked the first time, or did you not manage to work out why? I would like to get to the bottom of this before merging, preferably.

hdgarrood:
Sorry, I just realised I already asked you this a month ago!

hdgarrood:
Also, is there a test that I'm still able to define a class `Cons`, even in a module which, say, imports `Union` from `Prim.Row` (but nothing else)?

parsonsmatt:
> Also, is there a test that I'm still able to define a class Cons, even in a module which, say, imports Union from Prim.Row (but nothing else)?

There is now :slightly_smiling_face:

hdgarrood:
I was imagining that we would also have a statement like `import Prim.Row (class Union)` in that test as well - it's just that code which adds the entire module to the environment instead of just the things that were requested which I'm still a bit suspicious of.

Speaking of which, what happens if you remove that code? Also, what happens if you add all the Prim modules regardless of the imports?

parsonsmatt:
Oh dear:

```
Error found:
        in module Main
        at /home/matt/Projects/purescript/examples/passing/NewConsClass.purs line 11, column 1 - line 11, column
36

          Declaration for data constructor Cons conflicts with an existing type class of the same name.


        See https://github.com/purescript/documentation/blob/master/errors/DeclConflict.md for more information,
```

For a definition `data List a = Nil | Cons a (List a)` is going to conflct with the `Cons` type class. Maybe we should keep the `RowCons` name?

hdgarrood:
Right, this is what I was worried might be happening. If we're going to move `Prim` stuff into submodules then we need to make sure name resolution works properly; renaming so that clashes are slightly less likely isn't really a solution.

parsonsmatt:
The test case that I have defines `data List = ...`. Name resolution appears to work out fine.

hdgarrood:
What I mean by name resolution working correctly here is that it shouldn't be possible for conflicts like this to happen if you're not even importing the class `Cons` from `Prim.Row`.

Unrelated, but that error seems odd to me; shouldn't data constructors live in a different namespace from type classes..?

parsonsmatt:
I really should have posted more context :sweat_smile: The failing test case imported `import Prim.Row (class Cons)` to see if it would conflict. And it *did* conflict, but with the data constructor? I'm writing more tests to ensure it's working like we expect.

hdgarrood:
Oh, actually, I think it's not necessarily a sign that something's wrong if data constructors can conflict with classes; if you try to define both in the same module, both of them will want to have that name in the generated JS, so that's probably why you're getting that error.

However, it should be fine to define a class in one module, import it from another, and also define another class (or data constructor) in the same module. You can do this currently; `data Union = Union` works, even though there's already a `class Union` in scope.

hdgarrood:
So I'm guessing what's happened is you've told the compiler that in any module where these classes are used, they're actually defined in that module.

LiamGoodacre:
@parsonsmatt 
Anything I can do to help this along?  I'd be happy to look into the name conflict issue.
Could you bring this up to date with current `master`?

parsonsmatt:
I'll try to get it up-to-date this weekend :)

As far as current status: I think I was in the middle of writing tests to ensure that the terms are brought into scope correctly. I think that @hdgarrood's intuition is right:

> So I'm guessing what's happened is you've told the compiler that in any module where these classes are used, they're actually defined in that module.

But more tests would be good to verify that :smile:

parsonsmatt:
I've added a test for name overlap. So

```purescript
import Prim.Row (class Cons)

data Cons = Cons
```

works out fine.

parsonsmatt:
Are there any other tests we'd like to see here before merging?

LiamGoodacre:
I have:
* on `purescript-symbols`
    * created branch `compiler/0.12`
    * updated dev-deps and imported `Prim.TypeError (TypeConcat)`
* on `purescript-typelevel-prelude`
    * created branch `compiler/0.12`
    * depended on `purescript-symbols#compiler/0.12`
    * merged https://github.com/purescript/purescript-typelevel-prelude/pull/28 into `compiler/0.12`

@parsonsmatt would you update the test support to depend on `purescript/purescript-typelevel-prelude#compiler/0.12`?

LiamGoodacre:
Ah wait, need to depend on lists too.  Will make a branch there too

LiamGoodacre:
Okay, `purescript/purescript-lists#compiler/0.12` now exists

parsonsmatt:
I've updated the tests, and they're green (on my machine) :smile:

parsonsmatt:
well, the tests are green on my computer :\

image

LiamGoodacre:
This fails locally for me in the same way.

garyb:
It doesn't fail for me locally, either on my Mac or Windows machine. I tried `stack clean` just in case, also cleared out the test support node/bower files, still works (did that on both machines too). 🤔 

Travis cache maybe? Not sure why it would fail for Liam too though.

LiamGoodacre:
I'm doing: `rm -rf tests/support/bower_components .psci_modules .test_modules; stack clean; stack test --fast`.  Checked both this work by itself and also pulled into master.  All the same results.

Do you see this too?
```
Please note that,
    purescript-test-suite-support, purescript-typelevel-prelude#d029d9041a depends on purescript-symbols#compiler/0.12 which resolved to purescript-symbols#3b9236e8b0
    purescript-generics-rep#5.1.0 depends on purescript-symbols#^3.0.0 which resolved to purescript-symbols#3.0.0
Resort to using purescript-symbols#compiler/0.12 which resolved to purescript-symbols#3b9236e8b0
Code incompatibilities may occur.

Please note that,
    purescript-test-suite-support depends on purescript-lists#compiler/0.12 which resolved to purescript-lists#f2f3ec15c6
    purescript-generics#4.0.0 depends on purescript-lists#^4.0.0 which resolved to purescript-lists#4.12.0
Resort to using purescript-lists#compiler/0.12 which resolved to purescript-lists#f2f3ec15c6
Code incompatibilities may occur.
```

I'm wondering if there is a difference in behaviour across `bower` versions? (I'm clutching at straws here).

```
bower: 1.8.0
node: 8.9.4
npm: 5.6.0
stack: 1.6.3 (hpack-0.20.0)
```

garyb:
I get those messages, yeah. And:

```
bower: 1.8.2
node: 8.8.1
npm: 5.4.2
stack: 1.6.3 (hpack-0.20.0)
```

parsonsmatt:
Running the command you posted gives me the test failure also. Huh. :man_shrugging: 

Versions:

```
λ bower --version
1.8.2
λ node --version
v6.11.4
λ npm --version
3.5.2
λ stack --version
Version 1.6.5, Git revision 24ab0d6ff07f28276e082c3ce74dfdeb1a2ca9e9 (5514 commits) x86_64 hpack-0.20.0
```

Ubuntu 17.10 if that makes any difference.

garyb:
I forgot about `test_modules`, clearing that out breaks it for me too :smile:

parsonsmatt:
i've got a lead on where this is occurring (entailment, it's getting a `Nothing` on `stripPrefix` somehow?)

parsonsmatt:
We're getting "no instance found" errors. This appears to be specific to the recursion in the instance chain.

I changed the order of instances in the instance context for the second case. This puts 'Balanced sym2' before the others.

```purescript
instance balanced1 :: Balanced ""
else
instance balanced2
  :: ( Balanced sym2
     , AppendSymbol "(" sym1 sym
     , AppendSymbol sym2 ")" sym1
     ) => Balanced sym
```

The error then becomes this:

```
Tests
  compiler
    Passing examples
      'AppendInReverse.purs' should compile and run without error: FAIL (0.09s)
        Error found:
        in module Main
        at /home/matt/Projects/purescript/examples/passing/AppendInReverse.purs line 32, column 6 - line 32, column 42
        
            Main.Balanced t1
                            
        while applying a function balanced
          of type Balanced t0 => SProxy t0 -> String
          to argument SProxy
        while checking that expression balanced SProxy
          has type String
        in value declaration b3
        
        where t0 is an unknown type
              t1 is an unknown type
        
1 out of 1 tests failed (0.09s)
```

It looks like it is having a hard time doing instance lookup recursively in the presence of an instance chain.

src/Language/PureScript/Sugar/Names/Imports.hs
parsonsmatt: This comment appears to have been incorrect, as later tests verified.

matthewleon:
> It looks like it is having a hard time doing instance lookup recursively in the presence of an instance chain.

FWIW I have run into similar-seeming problems in different contexts, where switching instance order, or even adding redundant entailments, makes the problem go away. I imagine someone must have filed a bug at some point.

parsonsmatt:
Ah! The previous version of the test suite used this branch:

https://github.com/purescript/purescript-typelevel-prelude/compare/phil/append-symbol

Looks like this should be merged into `compiler/0.12` branch on `purescript-typelevel-prelude`.  @LiamGoodacre I think that should fix it :smile:

parsonsmatt:
now that the PR in `purescript-typelevel-prelude` was merged, I've pointed the tests to that commit. they pass locally. let's hope the CI agrees :smile:

garyb:
I opened a PR on your PR that removes the State stuff again, and the tests still pass - the change that enables this is to always populate `Environment` with `allPrimTypes` and `allPrimClasses` - which is fine as accessing those things in `Environment` would be guarded by the fact code runs through name desugaring first.

parsonsmatt:
Hopefully my next contribution will be more straightforward :smile:

garyb:
Yeah, sorry, I should have gotten involved with reviewing this , etc. a long time ago too.

PR #3181, opened 2017-12-20
Print error message on failed examples

parsonsmatt:
When a failed example provides a different error message than expected, it can be difficult to understand why.

This PR prints the given error message, which makes it easy to understand why the error occurred.

Before:

image

After:

image


hdgarrood:
Looks great; can you show a before/after example for when we hit this code path?

parsonsmatt:
Edited the original comment to include screencaps :smile:

hdgarrood:
I think the "full error messages" part might be getting overwritten by the error message rendering; should we put a newline between that and the call to render the error messages?

kRITZCREEK:
Since this seems to be the first of your contributions to get merged would you add yourself to the `CONTRIBUTORS.md` in the project root please? As soon as this CI run completes and you added yourself I'll merge.

Very happy to have you on board!

PR #3184, opened 2017-12-20
Fix functional dependencies rendering as links

### LINKED ISSUE #3180: Functional dependencies incorrectly render as links
See e.g. https://pursuit.purescript.org/packages/purescript-transformers/3.5.0/docs/Control.Monad.Error.Class#t:MonadError

monaderror

The type variables `m` and `e` in the functional dependency incorrectly render as links; those links would point to *values* `e` and `m` in that module, if values of those names existed (they don't). The type variables should be rendered the same way type variables are rendered in a normal class definition, i.e. not links.

This should be a relatively simple change in `Language.PureScript.Docs` somewhere.

houli:
Fixes #3180 
Before:
screen shot 2017-12-20 at 22 46 39

After:
screen shot 2017-12-20 at 22 48 11

### DIFF
--- a/src/Language/PureScript/Docs/Render.hs
+++ b/src/Language/PureScript/Docs/Render.hs
@@ -70,7 +70,7 @@ renderDeclarationWithOptions opts Declaration{..} =
              [idents from <> sp <> syntax "->" <> sp <> idents to | (from, to) <- fundeps ]
            ]
         where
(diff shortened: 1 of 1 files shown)


houli:
Will fix this up later to rename `idents` to `typeVars`

kRITZCREEK:
Thanks!

PR #3144, opened 2017-11-03
pretty print proxy types when rendering docs

sloosch:
Fixes #3135 .

PR #3186, opened 2017-12-30
Revert proxies

### LINKED ISSUE #3178: Proxy syntax / plans for 0.12
Currently I think it's not entirely clear what the plans are for 0.12, and what the difference between the `0.12.0-dev` and `master` branches is. I'm opening this issue to try to resolve this. There's been a bit of discussion on Slack already but issues are easier to refer back to in the future, which is why I'm creating one here.

The new proxy syntax presents issues because of the kind unification bug/bugs mentioned in #3075 and #3077 (note: #3077 exists in 0.11.7 too, but the issue is much harder to run into accidentally without proxy syntax). Additionally, it sounds like fixing this kind unification bug might take us some (most?) of the way towards polykinds, but of course if we have polykinds there's less of a reason to have special proxy syntax, since we can define a polykinded Proxy type; the only benefit of the proxy syntax is then that you can use proxies with less noise (`@"foo"` vs `Proxy :: Proxy "foo"`).

Setting the kind unification bug aside for a moment, I think it's probably also worth resolving the original issue in #3075 one way or another before we make a release which includes the proxy syntax, so that we don't later need to make a possibly-breaking change to make proxies comfortable to use.

I think the effort/value ratio of resolving these issues is probably quite high, so I think it might be worth considering reverting the proxy-related PRs for now to avoid blocking a 0.12.0 release, namely:

* #2846
* #3095
* #3124
* #3144

We could then delete the `0.12.0-dev` branch and continue working towards a 0.12 release on the `master` branch.

What do people think about this? /cc @rightfold

hdgarrood:
Resolves #3178. I've reverted the commits from the following PRs, in reverse chronological order:

- #3144 
- #3124 
- #3095 
- #2846 

After this is merged I will delete the `0.12.0-dev` branch (unless anyone objects, but I don't see a reason to now, since everything in that branch is now also in `master`).

~~Still WIP: I need to fix the tests and check the core libraries still compile.~~


examples/failing/InstanceChainBothUnknownAndMatch.purs
```diff
@@ -1,14 +1,18 @@
...
-class Same l r o | l r -> o
-instance sameY :: Same t t @"Y" else instance sameN :: Same l r @"N"
-same :: forall l r o. Same l r o => l -> r -> @o
-same _ _ = @o
+import Type.Proxy (Proxy)
+import Type.Row (RProxy(..))
+import Data.Symbol (SProxy(..))
+
+class Same l r (o :: Symbol) | l r -> o
+instance sameY :: Same t t "Y" else instance sameN :: Same l r "N"
```
hdgarrood: @LiamGoodacre I've changed this class slightly, so that the `o` parameter is a `Symbol` rather than a proxy, just because that seemed to make it slightly simpler. I did the same in the other failing instance chains test as well. Do you think this makes sense?

hdgarrood:
The CI failure is only on the Haddock build and it's to do with a timeout so I'm going to ignore it. I just checked and the core libraries all still compile with this code, apart from `datetime` which uses old-style generics.

examples/failing/InstanceChainBothUnknownAndMatch.purs
LiamGoodacre: Yes, that makes sense.

hdgarrood:
Unfortunately if you restart a job, it seems to pick up the cache as it was at the start of the first attempt, which means it will most likely fail for the same reason again (ie timeout installing dependencies). I should really have just pushed an empty commit - that way you get to pick up where the previous build left off.

hdgarrood:
Ok never mind then!

hdgarrood:
Thank you!

Issue #3187, opened 2018-01-01
naming an fn parameter __unused breaks codegen


matthewleon:
Admittedly a very, very tiny edge case but still, it feels morally wrong:

```purescript
abuseUnused :: forall a. a -> a
abuseUnused __unused = __unused
```

```javascript
var abuseUnused = function () {
    return __unused;
};
```

The culprit here is the `removeUnusedArg` optimization, which bases its functionality off the name of the argument.

Potential easy solution: change the constant to an illegal name?

matthewleon:
If this is desired behavior, at the least it should be documented somewhere. But I have trouble imagining what people would want this for.

hdgarrood:
Yeah, this definitely seems wrong. Changing the constant to an illegal name sounds like it could work.

paf31:
I think we should change the name to something the user can't type, or better still, remove the hack by creating a new type of name for this purpose.

matthewleon:
> better still, remove the hack by creating a new type of name for this purpose.

Yeah, I am leaning toward something like this.

nwolverson:
I'm not sure what I did but I managed to get typed hole suggestion to suggest `__unused` as an option recently...

matthewleon:
@nwolverson if you are able to do that again at some point, would you mind creating an issue for it? I'd be curious to see what's happening.

nwolverson:
I will do, I wasn't entirely clear what the case was because I couldn't reproduce it immediately afterwards, I think I noticed a `Partial` constraint

PR #3188, opened 2018-01-01
Allow __unused as arg

matthewleon:
resolves https://github.com/purescript/purescript/issues/3187

### DIFF
--- a/examples/passing/3187-UnusedNameClash.purs
+++ b/examples/passing/3187-UnusedNameClash.purs
@@ -1,0 +1,12 @@
+module Main (main) where
+
+import Prelude ((+))
+import Control.Monad.Eff.Console (log)
+
+-- the __unused parameter used to get optimized away
+abuseUnused :: forall a. a -> a
+abuseUnused __unused = __unused
+
+main = do
+  let explode = abuseUnused 0 + abuseUnused 0
+  log "Done"
--- a/src/Language/PureScript/Constants.hs
+++ b/src/Language/PureScript/Constants.hs
@@ -421,7 +421,7 @@ symbol = "Symbol"
 -- Code Generation
 
 __unused :: forall a. (IsString a) => a
-__unused = "__unused"
+__unused = "$unused" -- $ avoids name clashing
 
 -- Modules
 


hdgarrood:
This looks fine, but I'm just a little nervous about introducing new name clashes with this. Have you been able to rule that out as a possibility?

matthewleon:
PS doesn't allow names starting with `$`. Is there another way that such a clash could occur?

matthewleon:
(admittedly this is still "stringly-typed" programming, but I think it will do the trick in lieu of a much tricker solution)

hdgarrood:
Right, but the compiler already generates plenty of names with dollar signs in them in various situations.

natefaubion:
> Right, but the compiler already generates plenty of names with dollar signs in them in various situations.

I think `$prime` is the only thing that the compiler generates with a `$`, and it's used within an existing identifier.

matthewleon:
> I think $prime is the only thing that the compiler generates with a $, and it's used within an existing identifier.

there is a function called `identCharToText` that lists these. There are a few more. But we are safe.

matthewleon:
```haskell
-- | Attempts to find a human-readable name for a symbol, if none has been specified returns the
-- ordinal value.
identCharToText :: Char -> Text
identCharToText c | isAlphaNum c = T.singleton c
identCharToText '_' = "_"
identCharToText '.' = "$dot"
identCharToText '$' = "$dollar"
identCharToText '~' = "$tilde"
identCharToText '=' = "$eq"
identCharToText '<' = "$less"
identCharToText '>' = "$greater"
identCharToText '!' = "$bang"
identCharToText '#' = "$hash"
identCharToText '%' = "$percent"
identCharToText '^' = "$up"
identCharToText '&' = "$amp"
identCharToText '|' = "$bar"
identCharToText '*' = "$times"
identCharToText '/' = "$div"
identCharToText '+' = "$plus"
identCharToText '-' = "$minus"
identCharToText ':' = "$colon"
identCharToText '\\' = "$bslash"
identCharToText '?' = "$qmark"
identCharToText '@' = "$at"
identCharToText '\'' = "$prime"
identCharToText c = '$' `T.cons` T.pack (show (ord c))
```

natefaubion:
None of these are used anymore since you can only introduce them with operators, which are desugared to the actual identifier it represents.

matthewleon:
There is also `$$name` generated for builtins. But that's it.

matthewleon:
For reference, the function that handles the name conversion generally is `identToJS` in `CodeGen.JS.Common`.

hdgarrood:
So you're saying there's absolutely no use of `$` in generated names other than `$prime`, and reserved words in JS e.g. `$$Math`, `$$null`? Is that just in the JS output, or do we need to worry about other intermediate representations too?

I'd like to remove the unused code to do with how operators used to work, and actually I think I'd be slightly happier if we did that before doing this.

matthewleon:
> So you're saying there's absolutely no use of $ in generated names other than $prime, and reserved words in JS e.g. $$Math, $$null?

Yes

> Is that just in the JS output, or do we need to worry about other intermediate representations too?

This guarantees no clash in JS output and intermediate representations (they don't use "$" to my knowledge). Other backends, however, would have to avoid the `$unused` symbol. But then, they have to do that with `__unused` now anyway. At the very least, I think this makes things _slightly_ better.

> I'd like to remove the unused code to do with how operators used to work, and actually I think I'd be slightly happier if we did that before doing this.

Might make a PR for this if nobody else has that waiting.

hdgarrood:
> (they don't use "$" to my knowledge)

I've just been bitten too many times by thinking "I can't think of a way that this could go wrong, therefore it will not go wrong." Not using it to anyone's knowledge doesn't really count as any kind of guarantee, in my mind. I'll review this more thoroughly when I get a chance to try to verify that this is safe, but since this is a very small code change and it addresses a rare edge case that nobody has even run in to as far as we are aware, I'm probably not going to prioritise it (just FYI).

> Might make a PR for this if nobody else has that waiting.

I'm not aware of anyone else working on that; please go ahead!

matthewleon:
> since this is a very small code change and it addresses a rare edge case that nobody has even run in to as far as we are aware, I'm probably not going to prioritise it (just FYI).

Yes, this makes sense.

Also, on reflection, there is indeed the issue that other backends might potentially exist for languages where something like `$unused` is significant. I mean, the overwhelming odds are that we're fine... But the real solution here would be to avoid using stringly-typed programming and optimize this stuff away through metadata/annotations or something like that.

matthewleon:
I'm going to close this now and try to come up with a better solution :)

garyb:
`$` is only used in the names that are escaped (all that stuff is https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CodeGen/JS/Common.hs) and for freshly generated names, that always have a number appended: https://github.com/purescript/purescript/blob/3103dbf317c84e103384243944a377f0a0594774/src/Language/PureScript/Names.hs#L83-L86

hdgarrood:
Ok great - I forgot that freshly generated names have a number at the end. In that case I’m happy with this.

matthewleon:
I will resurrect this if I can't get something better to work. But I think there is a better way.

garyb:
To introduce the new name type will take a little work, as currently "CoreImp" (aka, JS) uses `Text` rather than a real type for idents. I'd also like that approach more though, as it'll make optimising unuseds etc. less fragile, especially if we start doing it for empty class dictionaries and whatnot (although it depends - maybe we'll do that elsewhere anyway).

garyb:
Not that it'd be hard to introduce a type for CoreImp names, just a little tedious :wink:

matthewleon:
I haven't done a thorough, check, but it seems that so far `C.__unused` is only used for a few different things, at least some of which could be optimized away _before_ hitting CoreImp. Perhaps all of them could be?

nwolverson:
(Regarding the mentions of "other backends" - there's always going to be some mangling in some circumstances, and it will be different to the JS mangling, which is fine so long as it's injective - I think any principled approach is likely to be fine)

houli:
Another use of $ in variable names is in the tail call optimisation code

PR #3190, opened 2018-01-03
bump ansi-terminal dependency.

### LINKED ISSUE #3189: Allow ansi-terminal version 0.8
[Purescript currently uses `ansi-terminal >=0.6.2 && <0.7`](https://github.com/purescript/purescript/blob/master/package.yaml#L37). It would be nice to use the most recent version 0.8 instead. [This would also make it build on Nix again, I think.](https://github.com/NixOS/nixpkgs/issues/33355) As [its changelog](https://hackage.haskell.org/package/ansi-terminal-0.8/changelog) suggests, it shouldn't be a problem to use ansi-terminal version 0.8.

mekeor:
Fixes #3189.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -34,7 +34,7 @@ extra-source-files:
 dependencies:
   - aeson >=1.0 && <1.3
   - aeson-better-errors >=0.8
-  - ansi-terminal >=0.6.2 && <0.7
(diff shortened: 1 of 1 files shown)


hdgarrood:
Addressed by #3209.

Issue #3182, opened 2017-12-20
Disable constraint syntax in `foreign import` declarations


natefaubion:
Constraints are magic completely handled by the compiler, and right now you can write foreign imports which depend on their representation. This is very brittle, and hinders a compilers ability to optimize their representation in anyway. It's much safer to take implementations explicitly. Should we eliminate constraint syntax from the foreign import grammar?

starper:
Maybe it should be disabled in `main` too? In my project I'm importing `Main.purs` like this:
```javascript
// import css, fonts and other webpack stuff here
import { main } from "./src/Main.purs"

main();
```
I added `Partial` constraint to `main` (not all cases were implemented at that moment) and got a little confused, when my app crushed with a runtime error.

garyb:
Regarding `main`: #2086

I don't think that's a great idea in general either, as there's no way to know which function called `main` is the actual `main`, whereas the change suggested here would be syntax directed.

hdgarrood:
What about producing a warning instead?

hdgarrood:
On second thoughts, I think disallowing constraints in foreign import declarations does make sense, for the reasons you give. I’d just not want to do that by modifying the parser. I think it would be less brittle, and we’d get better error messages, by doing this as a pass over the AST later on (say, just before typechecking?)

hdgarrood:
Can anyone think of a reason _not_ to do this? I'm quite strongly in favour now that we have a concrete optimization we want to add -- #2768 -- which is a breaking change, but wouldn't have to be if we disallowed using constraints in FFI imports.

hdgarrood:
Sorry, the above isn't quite right. The optimization in #2768 is not a breaking change because it just replaces these arguments with `undefined` rather than removing them outright (although other optimizations do often allow them to be removed entirely in practice).

hdgarrood:
I'd like to start warning for these in v0.14.0, and then turn the warning into an error in a subsequent breaking release.

paf31:
The book does document the intended representation of constrained values as functions, but it doesn't document the representation of dictionaries, so it's not like this was ever undocumented behavior. It's been documented as long as there were type classes. It is possible to safely treat dictionaries as abstract values in the FFI, but it's not safe to rely on the representation of the dictionary, which for example could be optimized away.

garyb:
I think the idea is that some constraints can be erased completely if they have empty dictionaries, so even their presence shouldn't be relied on, not just the implementation.

[...]

Issue #2768, opened 2017-03-23
Remove runtime overhead of typeclass instances with no dictionary


doolse:
With the new `Union` typeclass which exists purely for proof at compile time as it has no functions, it would be nice if when it was used as a constraint the runtime didn't have to pass a dictionary in for it. 

I understand that might be a difficult thing to change but what might be more feasible is to remove the overhead of creating blank dictionaries from the dictionaries of other constraints. E.g. 

```purescript
class MergeDefaults (o :: # Type) (mr :: # Type) (mo :: # Type)
instance mergeInstance :: (Union o m mo, Union m r mr, Subrow r o) => MergeDefaults o mr mo
```

Currently whenever it's used generate something along the lines of:
```javascript
Data_Record.mergeInstance({})({})(Data_Record.subrow({}))
```
which could be optimised simply to `{}`

garyb:
We could just remove the dictionary value usage entirely probably, so instead of converting the value to `{}`:

```
myFn(Data_Record.mergeInstance({})({})(Data_Record.subrow({})))(rec1)(rec2);
```

Could become:
```
myFn()(rec1)(rec2);
```

(see https://github.com/purescript/purescript/pull/2761#discussion_r107633521 - `{}` performs surprisingly badly)

paf31:
I think it should be relatively easy to generate `__unused` in `Entailment.hs` if the dictionary

- has no member functions and
- has no subgoals

matthewleon:
What do people think of a solution that removes the parameter entirely when it's an "empty" dictionary?

Disadvantages I can think of:
1) This would be (slightly) breaking; FFI code with "empty" constraints would have to have one "layer" peeled off.
2) When writing FFI code with constraints, you'd have to keep this in mind.

Advantages:
The JS generated would be simpler, more readable, and a bit more optimized.

matthewleon:
Oops. On further thought, I think I understand that this is exactly what @paf31 is suggesting: By converting the parameter to `__unused`, it will subsequently be killed by `removeUnusedArg` in the CoreImp optimizer.

matthewleon:
(though that still leaves us with an empty parameter, avoiding the FFI difficulties/inconsistencies but still leaving us with a needless function call. But I guess a later tool can get rid of that.)

PR #3194, opened 2018-01-05
create UnusedIdent

matthewleon:
Fixes https://github.com/purescript/purescript/issues/3187

Rather than using the magic string `"__unused"` to represent generated names that are meant to go unused, we create a new `Ident` constructor for them. Instead of eliminating unused function arguments in the optimization phase, they are stripped during the conversion from CoreFn to CoreImp.

### DIFF
--- a/examples/passing/3187-UnusedNameClash.purs
+++ b/examples/passing/3187-UnusedNameClash.purs
@@ -1,0 +1,12 @@
+module Main (main) where
+
+import Prelude ((+))
+import Control.Monad.Eff.Console (log)
+
+-- the __unused parameter used to get optimized away
+abuseUnused :: forall a. a -> a
+abuseUnused __unused = __unused
+
+main = do
+  let explode = abuseUnused 0 + abuseUnused 0
+  log "Done"
--- a/src/Language/PureScript/CodeGen/JS/Common.hs
+++ b/src/Language/PureScript/CodeGen/JS/Common.hs
@@ -26,6 +26,7 @@ moduleNameToJs (ModuleName pns) =
 identToJs :: Ident -> Text
 identToJs (Ident name) = properToJs name
 identToJs (GenIdent _ _) = internalError "GenIdent in identToJs"
+identToJs UnusedIdent = "$__unused"
 
 properToJs :: Text -> Text
 properToJs name
--- a/src/Language/PureScript/Names.hs
+++ b/src/Language/PureScript/Names.hs
@@ -76,6 +76,10 @@ data Ident
   -- A generated name for an identifier
   --
   | GenIdent (Maybe Text) Integer
+  -- |
+  -- A generated name used only for type-checking
+  --
+  | UnusedIdent
   deriving (Show, Eq, Ord, Generic)
 
 instance NFData Ident
@@ -84,6 +88,7 @@ runIdent :: Ident -> Text
 runIdent (Ident i) = i
 runIdent (GenIdent Nothing n) = "$" <> T.pack (show n)
 runIdent (GenIdent (Just name) n) = "$" <> name <> T.pack (show n)
+runIdent UnusedIdent = "$__unused"
 
 showIdent :: Ident -> Text
 showIdent = runIdent
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -164,6 +164,7 @@ moduleToJs (Module coms mn _ imps exps foreigns decls) foreign_ =
   accessor :: Ident -> AST -> AST
   accessor (Ident prop) = accessorString $ mkString prop
   accessor (GenIdent _ _) = internalError "GenIdent in accessor"
+  accessor UnusedIdent = internalError "UnusedIdent in accessor"
 
   accessorString :: PSString -> AST -> AST
   accessorString prop = AST.Indexer Nothing (AST.StringLiteral Nothing prop)
@@ -199,7 +200,10 @@ moduleToJs (Module coms mn _ imps exps foreigns decls) foreign_ =
                                (var name)
   valueToJs' (Abs _ arg val) = do
     ret <- valueToJs val
-    return $ AST.Function Nothing Nothing [identToJs arg] (AST.Block Nothing [AST.Return Nothing ret])
+    let jsArg = case arg of
+                  UnusedIdent -> []
+                  _           -> [identToJs arg]
+    return $ AST.Function Nothing Nothing jsArg (AST.Block Nothing [AST.Return Nothing ret])
   valueToJs' e@App{} = do
     let (f, args) = unApp e []
     args' <- mapM valueToJs args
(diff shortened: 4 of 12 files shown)


matthewleon:
This PR is a successor to https://github.com/purescript/purescript/pull/3188. Rather than attempting to address the issue by choosing a smarter magic string, it attempts to avoid optimization based on matching `Ident` names in the first place.

examples/passing/3187-UnusedNameClash.purs
```diff
@@ -0,0 +1,12 @@
+module Main (main) where
+
+import Prelude ((+))
+import Control.Monad.Eff.Console (log)
+
+-- the __unused parameter used to get optimized away
+abuseUnused :: forall a. a -> a
+abuseUnused __unused = __unused
```
kRITZCREEK: I don't understand why this is inside `passing`, shouldn't this explode now that `UnusedIdent` is a special, generated name inside the compiler?

src/Language/PureScript/CodeGen/JS/Common.hs
```diff
@@ -26,6 +27,13 @@ moduleNameToJs (ModuleName pns) =
 identToJs :: Ident -> Text
 identToJs (Ident name) = properToJs name
 identToJs (GenIdent _ _) = internalError "GenIdent in identToJs"
+identToJs UnusedIdent = C.__unused
+
+-- | Convert an 'Ident' into a valid JavaScript identifier, or Nothing.
```
kRITZCREEK: A function returning both a `Maybe` AND throwing exceptions is a red flag to me, which is almost guaranteed to be used wrong. And why do we still need the old version of this function?

Since this function is only used in a single place, I think it would be better to define it inline where it has more context.

src/Language/PureScript/Names.hs
```diff
@@ -94,6 +99,9 @@ freshIdent name = GenIdent (Just name) <$> fresh
 freshIdent' :: MonadSupply m => m Ident
 freshIdent' = GenIdent Nothing <$> fresh
 
+unusedIdent :: Ident
```
kRITZCREEK: Why this indirection? You could just use the constructor.

src/Language/PureScript/Names.hs
```diff
@@ -84,6 +88,7 @@ runIdent :: Ident -> Text
 runIdent (Ident i) = i
 runIdent (GenIdent Nothing n) = "$" <> T.pack (show n)
 runIdent (GenIdent (Just name) n) = "$" <> name <> T.pack (show n)
+runIdent UnusedIdent = "__unused"
```
kRITZCREEK: Now that this is a special identifier, shouldn't this also be prefixed with a `$` so it can't be mixed up with `Ident "__unused"`

examples/passing/3187-UnusedNameClash.purs
matthewleon: No, the opposite. It exploded before this PR, because `"__unused"` was a magic string. This PR makes it so `__unused` is no longer a magic name, but a name like any other. So this test now passes.

src/Language/PureScript/CodeGen/JS/Common.hs
matthewleon: Agreed. Will fix.

src/Language/PureScript/Names.hs
matthewleon: good point. Will fix.

src/Language/PureScript/Names.hs
matthewleon: will do

examples/passing/3187-UnusedNameClash.purs
kRITZCREEK: Oh I see :D I somehow thought this was a self recursive ever looping binding. My bad.

matthewleon:
I've implemented the requested changes. Eager for any further feedback.

kRITZCREEK:
Looks good to me, but I'd appreciate one more person with a better idea of Codegen to take a look and merge.

hdgarrood:
Should we ever hit the cases where we return `"$__unused"`, or should `UnusedIdent` constructors have been eliminated by then?

matthewleon:
We will hit that case when the compiler infers a `Partial` constraint. My next PR, https://github.com/purescript/purescript/pull/3195, which depends on this, then eliminates that case (to be discussed) and changes the `$__unused` to an error, like in the case of the other generated Names.

matthewleon:
To clarify: `__unused` was employed for two rather different purposes:

1) Unused function arguments, which is what this PR optimizes away.

2) Assigning an `id`-style function to the `__unused` name in a `Let` binding, and then calling it with the "old" function body wrapped in an IIFE, which is used to propagate the inferred `Partial` constraint. This PR leaves that as an `UnusedIdent`, which will the be turned into `$__unused` in JS. Then my next PR is meant to get rid of that, leaving the generated JS with no `$__unused`s.

hdgarrood:
Do we need to come to a consensus on #3182 first?

matthewleon:
> Do we need to come to a consensus on #3182 first?

I don't think so? The only behavioral change this PR introduces is that `$__unused` will be output for the `Partial` constraint rather than `__unused`.  This shouldn't have any effect on foreign imports, I don't think?

hdgarrood:
Sorry, I guess I really mean #2768; whether or not we do that seems like it would have implications for what we do here, or at least has implications for the approach taken in #3195, right? And the decision we make about #2768 also influences #3182 as well, which is why I mentioned that one in the first place.

hdgarrood:
Ok so I'm just untangling my own confusion here. AFAICT we're pretty much all agreed that we want to do #2768, right? That's a breaking change at the moment, but only because you're allowed to use constraints in `foreign import` declarations. If we did #3182 first (which is also breaking), it wouldn't be.

matthewleon:
#2768 is not necessarily breaking. The approach brought up by @paf31 simply converts the parameter to "unused" (`UnusedIdent` with this PR), which means that the function goes from `fn(emptyDictionary)(param)` before to `fn()(param)` after – there is still a function there. The more radical approach that I bring up would convert `fn(emptyDictionary)(param)` to `fn(param)`, which is indeed breaking. But I believe that is orthogonal to this PR. It does have an effect on #3195 , as you mention, but we can discuss that there (there is a test).

matthewleon:
As for #3182 , you're right of course. But I don't see how any of these impact this PR.

hdgarrood:
Ok, well ignore me for now then, and if I manage to sort out my confusion before we want to merge this then I'll let you all know, but otherwise please don't let me keep this from being merged.

paf31:
I wouldn't recommend trying to optimize as far as `fn(param)` if you ever want to support polymorphism over constraints, since it makes the representation of `=>` non-parametric. Inlining and specialization is probably a better way to go IMO.

matthewleon:
@garyb or @natefaubion any thoughts on this?

Issue #3157, opened 2017-11-21
Partial functions can foil TCO


matthewleon:
Curiously, non-partial functions with the `Partial` constraint added to them do result in TCO.

The following example is inspired by a Slack post by @natefaubion 

```purescript
module Tco where

import Data.List (List(..), (:))

tco :: Partial => List String -> String
tco ("" : _) = ""
tco (_ : xs) = tco xs
tco Nil = ""

notco :: Partial => List String -> String
notco ("" : _) = ""
notco (_ : xs) = notco xs
```

which compiles to the following javascript:
```javascript
// Generated by purs version 0.11.7
"use strict";
var Data_List = require("../Data.List");
var Data_List_Types = require("../Data.List.Types");
var tco = function ($copy_dictPartial) {
    return function ($copy_v) {
        var $tco_var_dictPartial = $copy_dictPartial;
        var $tco_done = false;
        var $tco_result;
        function $tco_loop(dictPartial, v) {
            if (v instanceof Data_List_Types.Cons && v.value0 === "") {
                $tco_done = true;
                return "";
            };
            if (v instanceof Data_List_Types.Cons) {
                $tco_var_dictPartial = dictPartial;
                $copy_v = v.value1;
                return;
            };
            if (v instanceof Data_List_Types.Nil) {
                $tco_done = true;
                return "";
            };
            throw new Error("Failed pattern match at Tco line 5, column 1 - line 5, column 40: " + [ v.constructor.name ]);
        };
        while (!$tco_done) {
            $tco_result = $tco_loop($tco_var_dictPartial, $copy_v);
        };
        return $tco_result;
    };
};
var notco = function (dictPartial) {
    return function (v) {
        var __unused = function (dictPartial1) {
            return function ($dollar4) {
                return $dollar4;
            };
        };
        return __unused(dictPartial)((function () {
            if (v instanceof Data_List_Types.Cons && v.value0 === "") {
                return "";
            };
            if (v instanceof Data_List_Types.Cons) {
                return notco(dictPartial)(v.value1);
            };
            throw new Error("Failed pattern match at Tco line 10, column 1 - line 10, column 42: " + [ v.constructor.name ]);
        })());
    };
};
module.exports = {
    tco: tco,
    notco: notco
};
```

matthewleon:
A much simpler example, with no dependencies, follows:

```purescript
module Main where

tco :: Partial => Int -> Int
tco 0 = 0
tco _ = tco 0

[...]

PR #3195, opened 2018-01-06
eliminate id function for inferred partial constraints in desugaring

matthewleon:
fixes https://github.com/purescript/purescript/issues/3157

This depends on https://github.com/purescript/purescript/pull/3194. If that is merged, I'll rebase accordingly.

During exhaustivity checking, an id function with a Partial constraint is created whenever a partial function is encountered, for the purposes of typechecking. During desugaring, typechecking is complete, so we can remove this construct.


matthewleon:
I discussed this with @kRITZCREEK and came to the conclusion that it shouldn't be sitting in the middle of the desugaring code. That said, it makes sense to perform this optimization between typechecking (the source of the generated block) and the functional core, as alternative backends shouldn't need to duplicate the optimization. Any ideas?

garyb:
Shall we close this then, since the approach is going to be different?

matthewleon:
> Shall we close this then, since the approach is going to be different?

sure

garyb:
I'm not sure how much you talked with @kRITZCREEK about this, but I chatted about it briefly with him a while ago too. My thinking was that there would be an `UnusedIdent` constructor in the names for the `CoreFn` (/ `CoreImp`? not sure how long we need it for) ASTs, perhaps. That'd give us more options for doing passes on the other form or something like that to apply the optimisation instead.

I was going to suggest making that change on the other PR even, and then keeping the "remove unused" pass in the optimizer there, but just having it less fragile since it'd have the constructor to match on... but in the end I figured it was fine to just not generate it at all.

matthewleon:
> My thinking was that there would also be an UnusedIdent constructor in the names for the CoreFn (/ CoreImp? not sure how long we need it for) ASTs, perhaps. That'd give us more options for doing passes on the other form or something like that to apply the optimisation instead.

I did have a chat with @kRITZCREEK about this one. I think what you are proposing makes sense broadly speaking. As it stands, `CoreFn` does "inherit" the `UnusedIdent` constructor from the unsugared AST... It gets erased in the transition to `CoreImp`.

That said, for _this_ particular problem, I believe it makes sense to do the optimization before even getting to `CoreFn`. Recall that the _only_ reason that this extra function is introduced is for type checking. Given that alternate backends might want to work with `CoreFn`, I can't see any reason for this particular artifact to be kept around after desugaring.

Thoughts?

matthewleon:
Another clarification here is that the particular _way_ that `UnusedIdent` is used in creating the partial constraints is fundamentally different from its usage elsewhere. Elsewhere, it is always a name given to a generated function parameter. Here, it is the name of a function itself.

garyb:
`CoreFn` currently lacks an optimiser, but there's a bunch of things we could in theory do there, so it just seemed like this would be the first pass - any backend would still benefit from it, as it'd use the post-optimised version anyway. Unless I misunderstand? 

The reason `CoreFn` exists is to make things like this easier, in fact! Since doing anything to the "full" AST is unpleasant just because of the number of cases that need considering. We're basically getting no benefit from `CoreFn` at all at the moment, as it goes directly into "`CoreImp`" (which is still just `JS` at this point really).

matthewleon:
> The reason CoreFn exists is to make things like this easier, in fact! Since doing anything to the "full" AST is unpleasant just because of the number of cases that need considering. We're basically getting no benefit from CoreFn at all at the moment, as it goes directly into "CoreImp" (which is still just JS at this point really).

Okay, cool. I will try to rewrite this in a `CoreFn` optimization pass.

garyb:
If it doesn't work out for some reason, that's fine too - perhaps we'll need to reconsider how the intermediate parts fit together to make them more useful.

Issue #888, opened 2015-02-19
Move where clauses under guards


pseudonom:
In Purescript,

``` purescript
foo | a = 1 where a = true
```

throws `Unknown value 'a'` on compilation.

(Haskell allows this.)

michaelficarra:
This seems pretty gross to me. Can you provide any samples of this being used to specify a function more elegantly than would otherwise be possible?

paf31:
I didn't know this was possible in Haskell. What does it desugar to?

michaelficarra:
I don't think this is specific to guards in Haskell. I believe any expression can have a `where`:

```
$ ghci
GHCi, version 7.8.4: http://www.haskell.org/ghc/  :? for help
Loading package ghc-prim ... linking ... done.
Loading package integer-gmp ... linking ... done.
Loading package base ... linking ... done.
λ> let x = a where a = 1
λ> x
1
λ>
```

pseudonom:
> Can you provide any samples of this being used to specify a function more elegantly than would otherwise be possible?

It's not the only way to write the function, but I initially tried guarded `where`s with a smart constructor:

```purescript
mkFoo :: Bar -> Either Error Foo
mkFoo a | test1 a = Left Err1 where test1 = ...
mkFoo a | test2 a = Left Err2 where test2 = ...
mkFoo a | test3 a = Left Err3 where test3 = ...
mkFoo a | otherwise = Right (Foo a)
```
(Effectively, a multi-way conditional.)

It's not obvious to me that there's a better way to write that.

mietek:
This syntax is used in Haskell to make all variables bound in the `where` clause available in all guards. For example, instead of writing
```purescript
topLeftCorner :: Rect -> Line -> Maybe Line
topLeftCorner r l | l.y2 > r.ytop = Nothing
                  | otherwise     = let d = delta l
                                    in  aux r l d (topProduct r l d) (leftProduct r l d)

aux :: Rect -> Line -> Point -> Number -> Number -> Maybe Line
aux r l d ptop pleft | ptop > pleft = leftBottomRegion r l d pleft
                     | otherwise    = reflectLineXMinusY <$> leftBottomRegion (reflectRectXMinusY r) (reflectLineXMinusY l) (reflectPointXMinusY d) ptop
```
I should be able to write
```purescript
topLeftCorner :: Rect -> Line -> Maybe Line
topLeftCorner r l | l.y2 > r.ytop = Nothing
                  | otherwise     = aux r l

[...]

PR #3200, opened 2018-01-12
Allow pattern matches with guards

minoki:
Currently, constructor patterns with guards like

```purescript
let Tuple x y | a < b = Tuple a b
              | otherwise = Tuple b a
```

is not allowed.  I think this restriction is not necessary.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -489,7 +489,7 @@ data Declaration
   | ValueDeclaration {-# UNPACK #-} !(ValueDeclarationData [GuardedExpr])
   -- |
   -- A declaration paired with pattern matching in let-in expression (binder, optional guard, value)
-  | BoundValueDeclaration SourceAnn Binder Expr
+  | BoundValueDeclaration SourceAnn Binder [GuardedExpr]
   -- |
   -- A minimal mutually recursive set of value declarations
   --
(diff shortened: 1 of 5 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -489,7 +489,7 @@ data Declaration
   | ValueDeclaration {-# UNPACK #-} !(ValueDeclarationData [GuardedExpr])
   -- |
   -- A declaration paired with pattern matching in let-in expression (binder, optional guard, value)
-  | BoundValueDeclaration SourceAnn Binder Expr
+  | BoundValueDeclaration SourceAnn Binder [GuardedExpr]
```
garyb: I think we should at least make this a `NEL.NonEmpty GuardedExpr` though.

kRITZCREEK:
I don't see myself using it either, but it definitely needs tests and the desugaring and parsing code can hopefully be shared with the other instances of pattern guards, rather than duplicating it.

natefaubion:
I tried using this not that long ago, but it’s rare.

LiamGoodacre:
I think I'd rather see multi-way `if` instead of this - which would be usable in more contexts.
However whether or not that is something we even want is a different discussion.

So I'm thinking to close this.  Thoughts?

rightfold:
I use this all the time in Haskell.

natefaubion:
This has not seen activity in some time. Should this be closed?

Vikasg7:
Nope, shouldn't be closed even if it takes time to implement. Hope to see purescript truly follow the haskell syntax, so that, users don't have to remember these syntax issues. #888 is also regarding guards.

hdgarrood:
Yes, I think it's time to close this. This isn't necessarily us saying we won't ever have this feature in PureScript, just that this PR doesn't seem to be going anywhere due to the fact that we have unresolved issues about how and whether we actually do this, as well as no activity on the PR for quite a while. For anyone who is interested in reviving discussion about this, please open an issue first so we can decide on the design we want. Thanks!

Issue #3204, opened 2018-01-14
ctags and etags generation foiled by explicit exports


matthewleon:
`purs docs` generation of the `ctags` and `etags` formats is not working correctly with export lists.

A minimal example with `purs` 0.11.7:

```purescript
module Main where

id :: forall a. a -> a
id x = x
```

generated tags (correct, though ample whitespace for `etags` looks a bit odd):
```
~/proj/tagtest$ purs docs src/Main.purs --format ctags
id	/Users/mleon/proj/tagtest/src/Main.purs	3
~/proj/tagtest$ purs docs src/Main.purs --format etags


/Users/mleon/proj/tagtest/src/Main.purs,6
id3,
~/proj/tagtest$ 
```

Now let's add an export list:
```purescript
module Main (id) where

id :: forall a. a -> a
id x = x
```

tags are empty now:
```
~/proj/tagtest$ purs docs src/Main.purs --format ctags
~/proj/tagtest$ purs docs src/Main.purs --format etags


/Users/mleon/proj/tagtest/src/Main.purs,0
~/proj/tagtest$ 
```

PR #3205, opened 2018-01-14
allow for explicit exports in generating tags

matthewleon:
fixes https://github.com/purescript/purescript/issues/3204

Reuses other formats' (HTML, Markdown) machinery for converting PS modules to a friendlier doc module format before generating tags. This eliminates some error-prone, ad-hoc logic for converting declarations directly to tags.

I haven't added any tests onto this branch. I believe that doing so would be made easier if we moved some of the tag generation code from `app` into `Language.Purescript`. If people find that appropriate, I'll be happy to add the commits for that onto this branch.


matthewleon:
Another note regarding tests: now that tags use the same intermediate machinery as the other documentation formats, the tests for those formats cover tags as well. We could make tests that are even more "end to end" by testing the string output of the entire tagging workflow, but I feel that the improvements here probably offer a good bang-for-buck compromise.

hdgarrood:
Moving the tag generation code into the library sounds good to me :+1:

matthewleon:
> Moving the tag generation code into the library sounds good to me

Okay, I'll take a look at this possibility soon.

src/Language/PureScript/Docs/Tags.hs
```diff
@@ -0,0 +1,46 @@
...
+
+import           Control.Arrow (first)
+import           Data.List (sort)
+import           Data.Maybe (mapMaybe)
+import qualified Data.Text as T
+import Language.PureScript.AST (SourceSpan, sourcePosLine, spanStart)
+import Language.PureScript.Docs.Types
+
+tags :: Module -> [(String, Int)]
+tags = map (first T.unpack) . concatMap dtags . modDeclarations
```
kRITZCREEK: this looks like a nice and pure function, what do you think about adding a few tests to make sure tag generation continues to work as we change the compiler?

app/Command/Docs.hs
```diff
@@ -60,19 +55,17 @@ docgen (PSCDocsOptions fmt inputGlob output) = do
     hPutStrLn stderr "purs docs: no input files."
     exitFailure
 
+  ms <- parseAndConvert input
   case fmt of
-    Etags -> dumpTags input dumpEtags
-    Ctags -> dumpTags input dumpCtags
+    Etags -> mapM_ putStrLn $ dumpEtags $ zip input ms
```
hdgarrood: I don’t think it’s safe to assume that the modules come out in the same order that they went in. We might need to modify `parseAndConvert` (instead of doing a `zip` here) so that it gives you the file path associated with each module.

src/Language/PureScript/Docs/Tags.hs
```diff
@@ -0,0 +1,46 @@
...
+dumpEtags :: [(String, Module)] -> [String]
+dumpEtags = concatMap renderModEtags
+
+renderModEtags :: (String, Module) -> [String]
+renderModEtags (path, mdl) = ["\x0c", path ++ "," ++ show tagsLen] ++ tagLines
+  where tagsLen = sum $ map length tagLines
+        tagLines = map tagLine $ tags mdl
+        tagLine (name, line) = "\x7f" ++ name ++ "\x01" ++ show line ++ ","
+
+dumpCtags :: [(String, Module)] -> [String]
```
hdgarrood: This isn’t a big deal but if you know why this one sorts but etags doesn’t, it might be useful to leave a comment.

src/Language/PureScript/Docs/Tags.hs
matthewleon: I think it's a great idea :). Actually, it's the main reason I proposed moving this code in the first place.

app/Command/Docs.hs
matthewleon: I'll take a look at this, thanks.

matthewleon:
This is ready for review.

@i-am-tom I recall you being eager to use this functionality. Would you mind trying to run this branch, if you have a moment?

app/Command/Docs.hs
```diff
@@ -114,9 +110,19 @@ docgen (PSCDocsOptions fmt inputGlob output) = do
...
-    runExceptT (D.parseFilesInPackages input []
-               >>= uncurry D.convertModulesInPackage)
-    >>= successOrExit
+    runExceptT (do
+      (modules, moduleMap) <- D.parseFilesInPackages input []
+
+      let moduleNameToFileMap =
+            M.fromList $ swap . fmap P.getModuleName <$> modules
+          getModuleFile docModule =
+            fromJust $ M.lookup (D.modName docModule) moduleNameToFileMap
```
hdgarrood: I’m not hugely keen on this. I guess this is to reassociate each file path with the relevant module after sending the modules through `convertModulesInPackage`? Is it worth considering changing `convertModulesInPackage` so that the output from `parseFilesInPackages` can just be passed straight in?

matthewleon: I also find it a bit hackish and ugly. Happy to look at your proposal, and see if other options are better as well. Just a clarification: beyond the hackishness of this, is there anything else that you find unpleasant about it that I should be aware of?

matthewleon: (and yes, your guess as to why this is done is correct)

matthewleon: Makes sense. Thanks for the clarification.

matthewleon:
@hdgarrood I've made a new commit to address your concerns. It is probably not exactly what you were hoping for, but I think it's a reasonable compromise.

My reasoning here:

1) Actually changing `convertModulesInPackage` to maintain (path, module) tuples without a `Map` would involve stringing that change through a large series of functions, some of which are called from multiple sites. It seems to me that it would be a pretty disruptive change.

2) So instead I've made a new function that uses the same kind of `Map`-based logic as before, but placed it in the `src` tree rather than `app`. This addresses your second concern more than your first; the logic is now easier to reuse.

3) As for your first concern, regarding the use of `fromJust`, I've replaced that with use of `MonadError`'s `throwError`. Admittedly, this still feels a little bit hackish, and might lead to a surprising error message if tag generation ever explodes there. I don't see an obvious better way to  do this, though. Please let me know if you can think of something more elegant.

i-am-tom:
@kRITZCREEK I did the tags and the things happen and everything seems to be great :D

matthewleon:
@i-am-tom I'm curious, are you using ctags or etags? Are you using any kind of tagbar plugin, or just doing "jump to definition" type stuff?

i-am-tom:
ctags and just jump-to-type stuff. I am an ever-closer-to-vanilla Vim fellow :)

matthewleon:
Just rebased to make merge easier.

matthewleon:
Hmm, something is killing build on the rebase. My apologies. Will try to fix.

matthewleon:
Okay, this is rebased on the newly fixed `master`.

hdgarrood:
I’ll just go ahead and merge this, since the Linux and Windows CI builds have all passed already.

hdgarrood:
Thanks very much!

PR #3206, opened 2018-01-18
[purs ide] Find Usages round two

kRITZCREEK:
This is my second attempt at find usages after #3140. Instead of resolving all usages at load time, they are now computed at query time. This makes sure we don't blow up our memory usage, and helps to avoid a few ambiguities around data constructor imports.

So far the code seems to perform pretty well, and I think `Find Usage` is an editor action to which I don't expect an immediate (< 100ms) response.

TODO:

- [x] Update `PROTOCOL.md`
- [x] Figure out the best response type, right now it's `[SourceSpan]`, but I could also see `Map ModuleName [SourceSpan]`, `Map FilePath (SourcePos, SourcePos)` ... do you have any preference here @nwolverson, @coot, @FrigoEU ?


tests/Language/PureScript/Ide/UsageSpec.hs
```diff
@@ -0,0 +1,69 @@
...
+        Test.runIde [ load ["FindUsage", "FindUsage.Definition", "FindUsage.Reexport"]
+                    , usage (Test.mn "FindUsage.Definition") "$%" IdeNSValue
+                    ]
+      usage1 `shouldBeUsage` ("src" </> "FindUsage.purs", "9:3-9:9")
+    it "finds a reexported usage" $ do
+      ([_, Right (UsagesResult [usage1])], _) <- Test.inProject $
+        Test.runIde [ load ["FindUsage", "FindUsage.Definition", "FindUsage.Reexport"]
+                    , usage (Test.mn "FindUsage.Reexport") "toBeReexported" IdeNSValue
+                    ]
+        -- TODO(Christoph): Interesting parser bug here. It seems the position
```
matthewleon: This is meant to be fixed in a separate PR? Should there be an issue for it?

kRITZCREEK: I'll probably want to confirm this first with a few more test cases. It's definitely not something I'd want to fix in this PR.

kRITZCREEK:
I've tested this and implemented the editor part for this in emacs here: https://github.com/epost/psc-ide-emacs/pull/134

I'm getting good performance (<50ms) for search `Data.Function.const` which is imported just about everwhere in ~350 modules.

I stuck with the `[SourceSpan]` response because it seems to be the simplest thing that works. If the Atom/Vim integrations can make use of more specific information about the usage (like inside which declaration the usage occured) we can talk about adding that.

Read to review from my side.

nwolverson:
I'd like to try integrating with this this weekend, will let you know how it goes

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -232,6 +232,10 @@ getModuleName (Module _ _ name _ _) = name
 getModuleSourceSpan :: Module -> SourceSpan
 getModuleSourceSpan (Module ss _ _ _ _) = ss
 
+-- | Return a module's declarations.
```
hdgarrood: Perhaps best not to do this right now, but I would quite like to use a record type for `Module`, and encourage using the record accessor functions instead of pattern matching on the `Module` constructor, so that any changes to the data type definition don't require so many other changes throughout the codebase.

nwolverson:
Well I got that working just fine, but I don't think I have a way in general to tell the correct namespace - I can do an initial query as for eg tooltips and if only one is in scope, fine, but in the case of multiple who knows.

I guess that's a wider problem though, and already an issue before this.

(Not had time to look at the code so can't comment)

kRITZCREEK:
@nwolverson I figured for now the editors can just ask for user input in the face of ambiguities. I think I'll add a "declarationType" field to the completion response which corresponds to the `declarationType` filter names. Mapping from those to namespaces is easy

tests/Language/PureScript/Ide/UsageSpec.hs
garyb: I think I've seen this bug too - usually in the atom plugin, if the last line has a warning/error on it, the span ends early.

natefaubion: This is because when we draw a token, the end span is determined by the start of the next token. Since the last token has no next token, the span is incorrect. I looked into this a while ago when looking into fixing positions for psa (since the span also includes whitespace and comments which are consumed as part of the lexeme grammar) and it is non trivial.

natefaubion: I ended up just using a bunch of heuristics in psa to clean it up.

nwolverson: I tried to fix this in the parser and failed utterly.

natefaubion: Yes it basically amounts to completely rethinking how the parser works, unfortunately, or sacrificing substantial performance.

matthewleon: Is there anywhere these struggles could be documented? It would probably be good to have this knowledge easier to find for the next person who tries to deal with this, no?

nwolverson:
@kRITZCREEK not possible to ask for user input, e.g. in the context of language server implementation, just as for tooltips the user will enjoy the convenience of an arbitrary choice, I guess based on an ordering (value > type > kind ?)

(But sounds good on the `declarationType`, as I was just contemplating the non-ideal option of reverse-engineering the type output)

tests/Language/PureScript/Ide/UsageSpec.hs
natefaubion: FWIW there’s this https://github.com/purescript/purescript/issues/1702

PR #3207, opened 2018-01-20
Eq1 / Ord1 deriving again

### LINKED ISSUE #2702: Eq1 / Ord1 deriving

garyb:
Resolves #2702. Much simpler than the previous #2703 :smile:.

The `Eq1` part is done, will do the same for `Ord1` now.

### DIFF
--- a/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
+++ b/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
@@ -533,6 +533,9 @@ deriveOrd ss mn syns ds tyConNm = do
     ordCompare :: Expr -> Expr -> Expr
     ordCompare = App . App (Var (Qualified (Just dataOrd) (Ident C.compare)))
 
+    ordCompare1 :: Expr -> Expr -> Expr
+    ordCompare1 = App . App (Var (Qualified (Just dataOrd) (Ident C.compare1)))
+
     mkCtorClauses :: ((ProperName 'ConstructorName, [Type]), Bool) -> m [CaseAlternative]
     mkCtorClauses ((ctorName, tys), isLast) = do
       identsL <- replicateM (length tys) (freshIdent "l")
@@ -570,12 +573,14 @@ deriveOrd ss mn syns ds tyConNm = do
                                   ]
 
     toOrdering :: Expr -> Expr -> Type -> Expr
-    toOrdering l r ty | Just rec <- objectType ty
-                      , Just fields <- decomposeRec rec =
-      appendAll
-      . map (\((Label str), typ) -> toOrdering (Accessor str l) (Accessor str r) typ)
-      $ fields
-    toOrdering l r _ = ordCompare l r
+    toOrdering l r ty
+      | Just rec <- objectType ty
+      , Just fields <- decomposeRec rec =
+          appendAll
+          . map (\((Label str), typ) -> toOrdering (Accessor str l) (Accessor str r) typ)
+          $ fields
+      | isAppliedVar ty = ordCompare1 l r
+      | otherwise = ordCompare l r
 
 deriveNewtype
   :: forall m
--- a/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
+++ b/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
@@ -436,6 +436,9 @@ deriveEq ss mn syns ds tyConNm = do
     preludeEq :: Expr -> Expr -> Expr
     preludeEq = App . App (Var (Qualified (Just dataEq) (Ident C.eq)))
 
+    preludeEq1 :: Expr -> Expr -> Expr
+    preludeEq1 = App . App (Var (Qualified (Just dataEq) (Ident C.eq1)))
+
     addCatch :: [CaseAlternative] -> [CaseAlternative]
     addCatch xs
       | length xs /= 1 = xs ++ [catchAll]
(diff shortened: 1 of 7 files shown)


garyb:
Brap!

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -458,12 +475,25 @@ deriveEq ss mn syns ds tyConNm = do
...
+      , Just fields <- decomposeRec rec =
+          conjAll
+          . map (\((Label str), typ) -> toEqTest (Accessor str l) (Accessor str r) typ)
+          $ fields
+      | isAppliedVar ty = preludeEq1 l r
+      | otherwise = preludeEq l r
+
+deriveEq1
+  :: forall m
+   . (MonadError MultipleErrors m, MonadSupply m)
```
kRITZCREEK: These constraints seem unnecessary?

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -547,12 +580,25 @@ deriveOrd ss mn syns ds tyConNm = do
...
+      , Just fields <- decomposeRec rec =
+          appendAll
+          . map (\((Label str), typ) -> toOrdering (Accessor str l) (Accessor str r) typ)
+          $ fields
+      | isAppliedVar ty = ordCompare1 l r
+      | otherwise = ordCompare l r
+
+deriveOrd1
+  :: forall m
+   . (MonadError MultipleErrors m, MonadSupply m)
```
kRITZCREEK: Same here

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -121,13 +121,27 @@ deriveInstance mn syns _ ds (TypeInstanceDeclaration sa@(ss, _) ch idx nm deps c
            -> TypeInstanceDeclaration sa ch idx nm deps className tys . ExplicitInstance <$> deriveEq ss mn syns ds tyCon
            | otherwise -> throwError . errorMessage' ss $ ExpectedTypeConstructor className tys ty
       _ -> throwError . errorMessage' ss $ InvalidDerivedInstance className tys 1
+  | className == Qualified (Just dataEq) (ProperName "Eq1")
+  = case tys of
+      [ty] | Just (Qualified mn' _, _) <- unwrapTypeConstructor ty
+           , mn == fromMaybe mn mn'
```
kRITZCREEK: `maybe True (== mn) mn'` ?

I was a little confused there for a second :D There's also the more obscure `all (== mn) mn'` but I think that's too cute :D

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
```diff
@@ -121,13 +121,27 @@ deriveInstance mn syns _ ds (TypeInstanceDeclaration sa@(ss, _) ch idx nm deps c
...
   = case tys of
       [ty] | Just (Qualified mn' tyCon, _) <- unwrapTypeConstructor ty
            , mn == fromMaybe mn mn'
            -> TypeInstanceDeclaration sa ch idx nm deps className tys . ExplicitInstance <$> deriveOrd ss mn syns ds tyCon
            | otherwise -> throwError . errorMessage' ss $ ExpectedTypeConstructor className tys ty
       _ -> throwError . errorMessage' ss $ InvalidDerivedInstance className tys 1
+  | className == Qualified (Just dataOrd) (ProperName "Ord1")
+  = case tys of
+      [ty] | Just (Qualified mn' _, _) <- unwrapTypeConstructor ty
+           , mn == fromMaybe mn mn'
```
kRITZCREEK: As above

src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
garyb: I just copied exactly the form of the other matches here. There's some other cleaning up I'd like to do actually, but didn't want to mix it in this PR :wink: so I can change it, but maybe it'd be better to keep them consistent for now?

kRITZCREEK: Should've looked at the whole file 🤦‍♂️

garyb:
As evidenced by de9e5ac, this is actually a breaking change, since some things will require a constraint like `Eq1 f` rather than `Eq (f a)` when deriving now. I think that's fine though, since they're basically equivalent, and this is for 0.12 anyway.

kRITZCREEK:
Do you think we can detect these cases and suggest the usage of Eq1?

kRITZCREEK:
If we can confirm that we can derive Eq1 for the type in question, I'd prefer us to add a hint to the error message:
```
The missing instance can be derived like:

`derive instance Eq1 TypeInQuestion`
```

garyb:
The "type in question" is `f a` though, so we can't suggest deriving an instance for it like that.

The change is that now whenever attempting to derive `Eq`, if there's a type like `(f _)` in the current type, it will use `eq1` for those values instead of `eq`, imposing an `Eq1 f` constraints rather than an `Eq (f _)` constraint as it would previously. It's just a case of juggling the constraints on the instance. If the `_` is also a type variable, then it'll impose `Eq a` too, or require that the type is `Eq` if it's concrete. For example:


Constructor | Original required constraints | New required constraints
--- | --- | ---
`Foo (f a)` | `Eq (f a)` | `Eq1 f, Eq a`
`Bar (f Int)` | `Eq (f Int)` | `Eq1 f`

kRITZCREEK:
Just thinking out loud here, does that mean that all our derived `Eq` instances that involve a type constructor get "slower" because of the indirection from Eq1 dictionary => Eq dictionary.

garyb:
It doesn't apply to type constructors, only applied-type-variables. It's a pretty rare case most likely :)

kRITZCREEK:
Well okay, if it's such a rare case I'm not too worried.

garyb:
I'll merge this too if there are no other issues with it?

kRITZCREEK:
Go for it! 🎉

PR #3209, opened 2018-01-21
update stack resolver to lts-10.3

matthewleon:
`stack build` will thus use GHC 8.2.2

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: nightly-2017-11-20
+resolver: lts-10.3
 packages:
 - '.'
 extra-deps:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ version: '{build}'
 environment:
   # Keep the path as short as possible, just in case.
   STACK_ROOT: c:\s
-  STACK_VER: 1.5.1
+  STACK_VER: 1.6.3
   RELEASE_USER: purescript
   RELEASE_REPO: purescript
 branches:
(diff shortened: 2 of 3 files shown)


stack.yaml
```diff
@@ -1,6 +1,7 @@
-resolver: nightly-2017-11-20
+resolver: lts-10.3
 packages:
 - '.'
 extra-deps:
 - pipes-http-1.0.5
 - Win32-notify-0.3.0.3
```
kRITZCREEK: this seems to be in the resolver now.

stack.yaml
```diff
@@ -1,6 +1,7 @@
...
+- ansi-terminal-0.6.3.1
```
kRITZCREEK: why can't we bump this to the resolvers 0.7.1.1 ?

matthewleon: I reviewed the changes and I don't see any reason we can't. I was being too conservative.

matthewleon:
No idea what to make of the fail on AppVeyor.

matthewleon:
possibly need to upgrade stack? https://github.com/fpco/stackage/issues/2759

kRITZCREEK:
You could try and bump the stack version in appveyor.yml to 1.6.3. Not sure how this would work for travis

matthewleon:
> You could try and bump the stack version in appveyor.yml to 1.6.3.

Ah, I stupidly missed that file. Thanks.

kRITZCREEK:
Nothing stupid there, I had to look it up myself.

hdgarrood:
Thanks! I decided to just go ahead and merge since it was just macOS builds we were waiting on.

Issue #3208, opened 2018-01-20
Warnings/errors without source spans


garyb:
Aside from #1801 we have some warnings currently with no source spans at all:

- [x] `ShadowedName`
- [x] `UnusedTypeVar`
- [x] `UnnecessaryFFIModule`
- [x] `UnusedFFIImplementations`
- [x] `MissingFFIImplementations`
- [x] `ScopeConflict`
- [x] `ScopeShadowing`
- [x] `DuplicateModule`
- [x] `ExportConflict`
- [x] `IntOutOfRange`
- [x] `ExpectedWildcard`
- [x] `MissingClassMember`
- [ ] `PartiallyAppliedSynonym`?
- [x] `ErrorParsingFFIModule`
- [x] `KindsDoNotUnify` (#3275)

If there are any others you run into, please post them below! (Or edit them in here if you have permission).

garyb:
`DuplicateModule` kinda has the same problem as `CycleInModules` in that we don't know which location is the cause of the issue. We could choose an arbitrary module and use the source span for that, but another alternative might be to raise an error for every module in the duplicates/cycle. Thoughts?

hdgarrood:
I’m not sure about what makes sense from an implementation perspective, but from a UX perspective I think ideally there would be just one error for each module name which is used more than once, but have the error include a list of all the files which define a module of this name.

hdgarrood:
I guess an appropriate source span would be the `module Foo` bit?

garyb:
The module names are included in the error text (and paths in the case of duplicate modules), but we can't attach source spans to those errors in the unless we change positioned errors to have a `NonEmptyList` of source spans or something, along with then finding a way of expressing that so that editors/tools can make use of it.

And yeah, the module header(s) is the span I had in mind :)

hdgarrood:
Oh, I didn’t realise all the paths are already included in the message text. Never mind then!

I think since we are doing that already it should be sufficient to pick an arbitrary one to use as the source span.

garyb:
@kRITZCREEK had another idea too - we can perhaps include spans for everything in the error data, and have that available in the JSON as `additionalSpans` or something, but when we print a text error we just use the arbitrary one. 🙂

Houndolon:
`MissingFFIImplementations` error is also missing source spans. I think it makes sense to report the locations of `foreign import someFunction ...` that lack the FFI implementation.

rndnoise:
[...]

PR #3210, opened 2018-01-21
Throw IntOutOfRange error with position information if present

kRITZCREEK:
Addresses one of the errors in #3208


garyb:
Should we merge this despite the Mac builds being stalled? Linux and Appveyor are happy.

hdgarrood:
Yeah, I wouldn't worry too much about that, especially for a change of this size.

PR #3211, opened 2018-01-21
Check warnings/errors for positions

garyb:


matthewleon:
`package-lock.json` is included in this commit. There seems to have been an informal policy to not do this up until now. Is this changing?

garyb:
Look again :wink: maybe GH being slow, it was in the original commit but I amended it and re-pushed before opening the PR.

src/Language/PureScript/Sugar/Operators/Common.hs
```diff
@@ -47,7 +47,6 @@ opTable
   -> [[P.Operator (Chain a) () Identity a]]
 opTable ops fromOp reapply =
   map (map (\(name, a) -> P.Infix (P.try (matchOp fromOp name) >> return (reapply name)) (toAssoc a))) ops
-  ++ [[ P.Infix (P.try (parseOp fromOp >>= \ident -> return (reapply ident))) P.AssocLeft ]]
```
MonoidMusician: Did the change from #3212 accidentally sneak in here? ;) https://github.com/purescript/purescript/pull/3212/files

garyb: It didn't sneak in no, it's kinda required since `reapply` needs a source position now, and this case can't provide one (unless / until the entire AST has source positions).

I've been considering how to test it since seeing Phil's reply to #3212, as it didn't break any of the existing tests.

garyb: Oh, er, maybe it snuck into this specific PR after all, but one of the others I opened about the same time makes the above comment true.

garyb:
I'm going to merge this when it actually passes, so when #3208 is finished basically :)

garyb:
This now passes :tada: (it's based on top of #3318)

PR #3212, opened 2018-01-21
Remove unused entry from `opTable`s

garyb:
&lt;I-have-no-idea-dog>

I _think_ this is leftover from before we required that all operators were declared via fixity statement though - a default associativity that any undeclared operators would fall back to.

Maybe @paf31 can confirm?


garyb:
(The line is actually deleted, I amended the commit in this PR again, but it's still showing as just commenting it out for me).

paf31:
I think it's used for things like ``foo `on` bar``.

PR #3213, opened 2018-01-21
Add source spans to name constructors and fix `ShadowedName` warning

garyb:
Part of #3208


garyb:
Oh, this also has #3212 in it, can either close that or merge it first or whatever.

PR #3214, opened 2018-01-21
Fix `UnusedTypeVar` missing position info

garyb:
For #3208


hdgarrood:
Ah, we've hit that stack bug on AppVeyor; we need to tell it to use Stack 1.6.x.

hdgarrood:
Oh sorry, I forgot that that version has already been bumped in master via #3209. I suppose this build must have started before that PR was merged. Rebasing should sort this out, and it might be worth removing that extra commit as well, in case it causes a conflict.

garyb:
Ah ok!

kRITZCREEK:
OSX queue is stuck forever, but I built and ran the tests locally and they passed, so I'll just get this in.

PR #3215, opened 2018-01-21
Fix build errors in master

garyb:


kRITZCREEK:
Hehe, this broke an ide usage test, because the position on the operator is now more accurate.

garyb:
Would you mind fixing that in here then? :wink: You should have access to push to this branch.

garyb:
Thanks!

PR #3218, opened 2018-01-23
Optimize inferred partial constraint fn

### LINKED ISSUE #3157: Partial functions can foil TCO

matthewleon:
fixes #3157

supersedes https://github.com/purescript/purescript/pull/3195

Adds a CoreFn optimization pass.

### DIFF
--- a/examples/passing/PartialTCO.purs
+++ b/examples/passing/PartialTCO.purs
@@ -1,0 +1,13 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+import Partial.Unsafe (unsafePartial)
+
+main = do
+  let _ = partialTCO true 1000000
+  log "Done"
+
+partialTCO :: Partial => Boolean -> Int -> Int
+partialTCO true 0 = 0
+partialTCO true n = partialTCO true (n - 1)
--- a/examples/passing/PartialTCO.purs
+++ b/examples/passing/PartialTCO.purs
@@ -5,7 +5,7 @@ import Control.Monad.Eff.Console (log)
 import Partial.Unsafe (unsafePartial)
 
 main = do
-  let _ = partialTCO true 1000000
+  let _ = unsafePartial partialTCO true 1000000
   log "Done"
 
 partialTCO :: Partial => Boolean -> Int -> Int
--- a/src/Language/PureScript/CoreFn/Optimizer.hs
+++ b/src/Language/PureScript/CoreFn/Optimizer.hs
@@ -57,6 +57,9 @@ optimizeModuleDecls = map transformBinds
 
   transformExprs = optimizeUnusedPartialFn
 
+-- |
+-- Optimize away function generated to typecheck inferred Partial constraints.
+--
 optimizeUnusedPartialFn :: Expr a -> Expr a
 optimizeUnusedPartialFn (Let _
   [NonRec _ UnusedIdent _]
--- a/src/Language/PureScript/CoreFn/Optimizer.hs
+++ b/src/Language/PureScript/CoreFn/Optimizer.hs
@@ -9,6 +9,7 @@ import Language.PureScript.CoreFn.Ann
 import Language.PureScript.CoreFn.Expr
 import Language.PureScript.CoreFn.Module
 import Language.PureScript.CoreFn.Traversals
+import Language.PureScript.Names (Ident(UnusedIdent), Qualified(Qualified))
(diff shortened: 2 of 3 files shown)


matthewleon:
This is ready for review.

matthewleon:
Some thoughts here: The ad-hoc nature of this optimization isn't especially pleasing. On the other hand, I do believe that it's fixing a bug: Without this, code that really seems like it _should_ get TCOd, doesn't, leading to potential run-time explosions.

If people can think of a smarter way to deal with this, I'm really open to it.

gabejohnson:
Since I presume more optimizations will be added in the future (see #3244), some of which may be more expensive to perform, it might be a good idea to put these behind a flag.

garyb:
@matthewleon would you mind perhaps opening a PR that adds the corefn pass but that doesn't do anything? I have some concerns about the way this optimization is applied at the moment (in that it could be too wide reaching, given it's not actually `Partial`-specific) and would like to spend some time experimenting with that. But also #3244 is based on this (in that it uses a corefn pass), and I think that one is good to go in.

matthewleon:
Will do.

matthewleon:
I'll rebase this after merge of https://github.com/purescript/purescript/pull/3319 so that it becomes a PR for just the inferred partial constraint optimization.

natefaubion:
@matthewleon Are you interested in picking this back up?

matthewleon:
> Are you interested in picking this back up?

Yes. Will do it on Wednesday.

matthewleon:
@natefaubion rebased and fixed the conflict. Please feel free to correct the composition order I chose for the optimizations if there is a problem there.

examples/passing/PartialTCO.purs
```diff
@@ -0,0 +1,13 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
```
natefaubion: Should this be `Effect.Console`? How is this test passing? 😆

natefaubion: This test is in the wrong directory. It should be in `tests/purs/passing`. I think this is because the tests were moved sometime after the point you made this PR.

matthewleon: Yikes! The perils of rebasing :)

I've pushed a fix, thanks.

natefaubion:
Thanks!

Issue #3192, opened 2018-01-04
Do notation: discarded bindings result in generated bindings in codegen


matthewleon:
Quick example:

```purescript
main = do
  _ <- newRef 0
  log "Hello sailor!"
```

result:
```javascript
var main = function __do() {
    var v = Control_Monad_Eff_Ref.newRef(0)();
    return Control_Monad_Eff_Console.log("Hello sailor!")();
};
```

Is there any reason that we should be binding to `var v`? It seems like unnecessary noise.

matthewleon:
My suspicion is that the generated variable is necessary for typechecking, but that it could be switched to `__unused`.

I'm working on improving the mechanism for optimizing away such variables right now, so if my thoughts here are correct, I'll return to the issue afterward.

PR #3220, opened 2018-01-24
optimize away binds to wildcards in Do notation

matthewleon:
fixes https://github.com/purescript/purescript/issues/3192

I've manually QA'd this on the code in the issue above, and it passes all tests. We don't seem to have tests so far that check the compiler output end-to-end, and I can't think of a test that would pass on this change but fail previously, without relying on some extremely brittle FFI.


matthewleon:
It might be possible to subsume this PR in https://github.com/purescript/purescript/issues/3222

I'm not sure.

I'll give it a try soon, if people like the idea of the other PR.

matthewleon:
^^ As I've closed the Issue referenced above, this is ready for review.

kRITZCREEK:
I have a few thoughts on how to go about optimization in PureScript.

- I want it to be more principled than ad-hoc pattern matches. Let's build a call-graph, see if we can analyze dataflow, be smart about inlining, and find unused arguments in general not just wildcards in do-notation.
- I want it to be a separate from the fast-rebuild path. Our dev experience should not have to suffer
- I'd like to explore using a bigger compilation unit than the module when we go to optimization. I'm not talking about whole-program optimization necessarily (though that's very interesting), but doing it at the package level like Rust could be a good middleground.

matthewleon:
@kRITZCREEK that sounds like a great approach.

That said, this particular optimization is implemented as two extra pattern matches in pre-existing desugaring/optimization code. That is to say, I don't think it adds any real additional overhead, and it feels to me to be "in the spirit" of the code where it's being added.

More generally, it seems to me that iterating/fixing existing ad hoc optimizations in the compiler doesn't necessarily come at the cost of an eventual goal of stripping those optimizations in favor of a more principled approach.

What do you think?

matthewleon:
On reflection, I do see a good counterpoint to what I wrote about, though: Even adding a single extra branch for optimization to these pattern matches does add _some_ complexity to the codebase. And the surrounding optimizations are ones that might _not_ be covered by a dataflow-analysis based optimization pass.

hdgarrood:
I agree with @kritzcreek in that it's not how I'd want to go about doing optimization, but also I don't think this particular PR should be considered an optimization in the first place; I prefer to think of it as improving the readability of the generated code. (I'm tempted to just fix the conflicts and get this merged, because I think there are too many open PRs right now).

hdgarrood:
One thing I don't understand is why the `NullBinder` constructor is wrapped in a `PositionedBinder` constructor at this point, when `VarBinder` isn't. In the AST which comes out of the parser, both wildcard and variable binders are wrapped in a `PositionedBinder` constructor.

natefaubion:
Is it because of the do-desugaring code? Maybe it's inserting stuff without it?

hdgarrood:
I don't think so? As far as I can tell it's all in `Language.PureScript.Sugar.DoNotation`, and it doesn't seem to be doing anything like that. It's also one of the first desugaring steps, preceded only by signed literals and object constructors, and I don't think either of them are responsible.

hdgarrood:
Ah, I've realised what's going on. I assumed that the existing case
```haskell
  go pos m (DoNotationBind (VarBinder ss ident) val : rest) = do
```
was working, but I don't think it is, because `VarBinder` constructors are also wrapped in `PositionedBinder` at this point. The purpose of this case alternative seems to be to use the same variable name in the output as in the source code, but it doesn't appear to ever be hit. For example:
```purescript
module Main where

import Prelude
import Effect.Console (log)

x = do
 y <- log "hi"
 _ <- log "bye"
 pure unit
```
currently compiles to
```
var x = function __do() {
    var v = Effect_Console.log("hi")();
    var v1 = Effect_Console.log("bye")();
    return Data_Unit.unit;
};
```
Note the variable `v` for the result of `log "hi"`, rather than `y` as written in the source file.

hdgarrood:
Ok, this should be ready for review. Here's an example of what this branch now does.

Source:
```purescript
x = do
 a <- log "a"
 b <- log "b"
 _ <- log "c"
 _ <- log "d"
 pure b
```
Output before:
```javascript
var x = function __do() {
    var v = Effect_Console.log("a")();
    var v1 = Effect_Console.log("b")();
    var v2 = Effect_Console.log("c")();
    var v3 = Effect_Console.log("d")();
    return v1;
};
```
Output after:
```javascript
var x = function __do() {
    var a = Effect_Console.log("a")();
    var b = Effect_Console.log("b")();
    Effect_Console.log("c")();
    Effect_Console.log("d")();
    return b;
};
```

src/Language/PureScript/Sugar/DoNotation.hs
```diff
@@ -40,6 +42,13 @@ desugarDo d =
   replace _ (PositionedValue pos com v) = PositionedValue pos com <$> rethrowWithPosition pos (replace pos v)
   replace _ other = return other
 
+  stripPositionedBinder :: Binder -> (Maybe SourceSpan, Binder)
+  stripPositionedBinder (PositionedBinder ss _ b) =
+    let (ss', b') = stripPositionedBinder b
+     in (ss' <|> Just ss, b')
+  stripPositionedBinder b =
+    (Nothing, b)
+
```
natefaubion: @hdgarrood Is this sort of check going to need to be done for `ado` as well?

hdgarrood: If we want to preserve variable names from the source in the output rather than using generated names, yes. I think I'd prefer to tackle that separately though.

hdgarrood:
Thanks @matthewleon and @natefaubion!

PR #3231, opened 2018-01-31
Improvments to REPL tab-completion (#3227)

### LINKED ISSUE #3227: REPL doesn't auto-complete re-exported symbols
Given a module like this:

```purescript
module X ( foo, module Data.Maybe ) where
import Data.Maybe (maybe)

foo :: String
foo = "ok"
```

When running `pulp repl`, and entering `import X as X`, then typing `X.[tab]` will only show `foo` but not `maybe`. This also happens when importing `X` unqualified.

rndnoise:
- Complete all names that have been imported (transitively or directly)
- Do not complete names that haven't been imported
- Only recompute list of names after import or adding a let binding
  rather than after each request for name completion

This pull request fixes #3227

### DIFF
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -190,9 +190,8 @@ handleShowImportedModules
   => (String -> m ())
   -> m ()
 handleShowImportedModules print' = do
-  PSCiState { psciImportedModules = importedModules } <- get
+  importedModules <- psciImportedModules <$> get
   print' $ showModules importedModules
-  return ()
   where
   showModules = unlines . sort . map (T.unpack . showModule)
   showModule (mn, declType, asQ) =
--- a/src/Language/PureScript/Interactive/Types.hs
+++ b/src/Language/PureScript/Interactive/Types.hs
@@ -1,11 +1,38 @@
 -- |
 -- Type declarations and associated basic functions for PSCI.
 --
-module Language.PureScript.Interactive.Types where
+module Language.PureScript.Interactive.Types
+  ( PSCiConfig(..)
+  , PSCiState -- constructor is not exported, to prevent psciImports and psciExports from
+              -- becoming inconsistent with importedModules, letBindings and loadedExterns
+  , ImportedModule
+  , psciExports
+  , psciImports
+  , psciLoadedExterns
+  , psciImportedModules
+  , psciLetBindings
+  , initialPSCiState
+  , psciImportedModuleNames
+  , updateImportedModules
+  , updateLoadedExterns
+  , updateLets
+  , Command(..)
+  , ReplQuery(..)
+  , replQueries
+  , replQueryStrings
+  , showReplQuery
+  , parseReplQuery
+  , Directive(..)
+  ) where
 
 import Prelude.Compat
 
 import qualified Language.PureScript as P
+import qualified Data.Map as M
+import           Language.PureScript.Sugar.Names.Env (nullImports, primExports)
+import           Control.Monad.Trans.Except (runExceptT)
+import           Control.Monad.Writer.Strict (runWriterT)
+
 
 -- | The PSCI configuration.
 --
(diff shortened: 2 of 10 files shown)


src/Language/PureScript/Interactive/Types.hs
```diff
@@ -54,17 +64,47 @@ allImportsOf m PSCiState{psciImportedModules = is} =
 
 -- * State helpers
 
+updateImportExports :: PSCiState -> PSCiState
```
hdgarrood: Could you add a comment here saying something like

> This function updates the Imports and Exports values in the PSCiState, which are used for handling completions. This function must be called whenever the PSCiState is modified to ensure that completions remain accurate.

src/Language/PureScript/Interactive/Types.hs
```diff
@@ -25,10 +30,15 @@ data PSCiState = PSCiState
...
+
+  -- These are derived from updateImportExports each time a module is imported,
+  -- a let binding is added, or the session is cleared or reloaded
+  , psciImports             :: P.Imports
+  , psciExports             :: P.Exports
   } deriving Show
 
 initialPSCiState :: PSCiState
-initialPSCiState = PSCiState [] [] []
+initialPSCiState = updateImportExports (PSCiState [] [] [] nullImports nullExports)
```
hdgarrood: Will `updateImportExports` do anything here?

src/Language/PureScript/Interactive/Types.hs
hdgarrood: Alternatively, is it worth using an explicit export list and hiding the `PSCiState` constructor so that we can easily verify that the `PSCiState` can't get out of sync?

tests/TestPsci/CompletionTest.hs
```diff
@@ -73,16 +78,13 @@ completionTestData supportModuleNames =
...
-  -- if a module is imported qualified, values should complete under the
-  -- qualified name, as well as the original name.
-  , ("ST.new", ["ST.newSTRef"])
-  , ("Control.Monad.ST.new", ["Control.Monad.ST.newSTRef"])
+  -- Identifiers and data constructors in scope should be completed
+  , ("uni", ["unit"])
+  , ("G", ["GT"])
+  , ("P.G", ["P.GT"])
+  , ("P.uni", ["P.unit"])
+  , ("voi", []) -- import Prelude exclude (void)
```
hdgarrood: The syntax is `import Prelude hiding (void)`

src/Language/PureScript/Interactive/Types.hs
rndnoise: Yes, perhaps surprisingly. It adds the types from `Prim`, which is implicitly imported in the module created in that function.

src/Language/PureScript/Interactive/Types.hs
rndnoise: Yeah, I think both are good ideas.

rndnoise:
I've hidden the `PSCiState` constructor and added a comment to `updateImportExports`.

I also made an optimization to `tests/TestPsci/CompletionTest.hs` so the module is reused rather than setup and recompiled for each assertion (running the REPL tests was noticeably slow, it's much faster now). The remaining slow bit is `tests/TestPsci/CommandTest.hs` which I've left alone.

rndnoise:
The build is failing due to an unused function `allImportsOf`. It's no longer used as the result of this PR, but I didn't want to delete someone's work without their permission. I think this code is yours, @hdgarrood ... should I remove it?

rndnoise:
After a closer look, I don't think `allImportsOf` was useful outside of the older tab-completion code that called it; so I went ahead and removed it so the build would pass. Let me know if any further changes are needed before merging.

src/Language/PureScript/Interactive.hs
```diff
@@ -190,9 +190,8 @@ handleShowImportedModules
   => (String -> m ())
   -> m ()
 handleShowImportedModules print' = do
-  PSCiState { psciImportedModules = importedModules } <- get
-  print' $ showModules importedModules
-  return ()
+  st <- get
+  print' $ showModules (psciImportedModules st)
```
hdgarrood: Bit of a nitpick, but using `$` as well as parens on the same line is stylistically perhaps not the best thing to do. I don't mind which but would you mind sticking to one or the other?

rndnoise:
OK, I think I've addressed all of your points.

> So would you mind splitting that off for now?

Sure, it is done. In general though, how do contributors here manage related or interdependent pull requests? Should those not be made into PRs until the dependency has been merged, or do people create PRs that have stacked on other PR branches?

> I just realised it would probably be helpful to add a comment near initialPSCiState to clarify the reason we need to use updateImportsExports there?

I've removed that call and exported `Sugar.Names.Env.primExports` so that it could be used here instead of `nullExports`.

> Also, looking at this I realised that we haven't actually managed to guarantee that the Imports and Exports stay in sync this way;

Good catch. I saw a compiler error halfway through making that change and assumed it confirmed my guess that I could export the type without exporting the constructor, but it must have been a different error. It is now a data declaration with some extra "getter" methods, so now it shouldn't be possible to take apart or build a `PSCiState` directly.

> Bit of a nitpick, but using $ as well as parens on the same line is stylistically perhaps not the best thing to do

No problem.

src/Language/PureScript/Interactive/Types.hs
```diff
@@ -22,13 +49,33 @@ data PSCiConfig = PSCiConfig
 -- The let bindings are partial,
 -- because it makes more sense to apply the binding to the final evaluated expression.
 data PSCiState = PSCiState
-  { psciImportedModules     :: [ImportedModule]
-  , psciLetBindings         :: [P.Declaration]
-  , psciLoadedExterns       :: [(P.Module, P.ExternsFile)]
-  } deriving Show
+  [ImportedModule]            -- ^ psciImportedModules
+  [P.Declaration]             -- ^ psciLetBindings
```
kRITZCREEK: Haddock is crashing on this line. It doesn't seem like you're allowed to annotate the individual arguments to a constructor, just the whole constructor or record fields: https://www.haskell.org/haddock/doc/html/ch03s02.html

src/Language/PureScript/Interactive/Types.hs
```diff
@@ -1,11 +1,38 @@
 -- |
 -- Type declarations and associated basic functions for PSCI.
 --
-module Language.PureScript.Interactive.Types where
+module Language.PureScript.Interactive.Types
+  ( PSCiConfig(..)
+  , PSCiState -- constructor is not exported, to prevent psciImports and psciExports from
+              -- becoming inconsistent with importedModules, letBindnigs and loadedExterns
```
hdgarrood: typo here? "letBindnigs"

hdgarrood:
This looks great; I'm happy to merge once CI is passing.

src/Language/PureScript/Interactive/Types.hs
rndnoise: Ah, thanks

src/Language/PureScript/Interactive/Types.hs
rndnoise: Yes, thanks :-)

PR #3234, opened 2018-02-02
Make REPL completion tests faster

rndnoise:
Rather than rebuild the same code for each assertion, this builds the environment once and reuses it. The whole repl set of tests went from about 35s to 10s. Now about half of the remaining time is setup done by `stack test` and the other half is `TestPsci` (a similar optimization might be possible there, but it's less clear).


kRITZCREEK:
@hdgarrood Do you have time to look at this one or do you want me to?

PR #3237, opened 2018-02-02
Complete any number of words after :kind and :type in REPL

rndnoise:
This fixes completion after `:kind` and `:type` directives, which previously would only complete the first word after the directive. With this change, any words after `:kind` are completed from names of types in the current scope, and words after `:type` are completed from names of values in the current scope.

This depends on #3231 being merged (not due to implementation dependency, but because the unit tests in this PR rely on those changes).


kRITZCREEK:
This one also looks like a straightforward improvement, thanks!

PR #3239, opened 2018-02-02
Complete type identifiers following '::' in REPL

rndnoise:
Prior to this PR, the only time types are auto-completed is following `:kind`. With this change, the first word following `::` is completed from the names of types in scope.

One limitation is only the very next word is completed as a type, so even after typing `[1] :: Array`, where `Array` is a type constructor, the next word will be completed from `CtxIdentifier`, or the names of values in scope.

This PR depends on #3237 to be merged, because this PR's unit tests depending on it.


kRITZCREEK:
I might've merged things in the wrong order? If you fix the conflicts I'll give this a try and get it merged :)

rndnoise:
No problem. I think these conflicted in any merge order.

kRITZCREEK:
Tested this, and it seems to work fine, thanks!

PR #3240, opened 2018-02-03
Introduces HasName class to reduce some duplication

kRITZCREEK:
This is kind of weird to me... I usually don't like these kind of things, but it feels a lot better to me. Someone talk me out of it? :D


kRITZCREEK:
I'm just going to give these functions shorter names I guess. Maybe I should break up my big `Types` module into smaller parts so I can overlap names and use them qualified instead.

PR #3241, opened 2018-02-03
Build psciEnvironment as needed for :print in REPL

### LINKED ISSUE #3001: :browse does not pick up changes from :reload

rndnoise:
This fixes #3001 by computing `psciEnvironment` on demand.

> One thing I need help with in `simulateModuleEdit` is ensuring `enableRebuild` and `rebuild` is called even if `action` fails (e.g., due to an assertion). Otherwise the module that's swapped in won't be reverted and this will leak into other tests.

(The above issue is handled, see comment below)

Also, it should be noted that any code that calls `Language.PureScript.Interactive.make` depends on and modifies global state (since it reads/writes files in `.psci_modules`); I believe that includes all the other PSCi tests. This was a painful lesson that took me a while to learn, maybe others will benefit from this comment.


rndnoise:
Ok, this is ready for review. I've also used `bracket_` to ensure `simulateModuleEdit` cleans up even if the provided action fails.

(Prior build errors were due to (a) using `'/'` as a path separator on Windows and (b) not having included the new file `Regex.edit` in the source distribution.)

tests/TestPsci/TestEnv.hs
```diff
@@ -90,7 +95,37 @@ evaluatesTo command expected = runAndEval command evalJsAndCompare ignorePrinted
...
+  slash      = T.singleton pathSeparator
+
+  replacePath xs c = c { psciFileGlobs = xs }
+
+  findReplace :: [String] -> [String] -> Int -> Maybe [String]
+  findReplace (x:xs) acc n
+    | inputPath `isSuffixOf` x = findReplace xs acc (n+1)
+    | otherwise                = findReplace xs acc n
+  findReplace [] _   0         = Nothing
+  findReplace [] acc _         = Just (newPath : reverse acc)
```
natefaubion: This function seems kind of odd to me, but otherwise this PR seems like a win. Nothing is ever done with `acc`, so I'm actually not sure what all this indirection is aboutl

rndnoise: It's been a while, but I believe this function takes a list of paths and removes all that have matching suffixes and sticks the `newPath` in front, only if there were matches. If no matches were found, it returns `Nothing`. On the last line (where at least one path matched), `acc` is the list of paths that remained, because they didn't match, in reverse order.

I think the bigger picture is to make a new temporary file to replace the module we want to "edit". The module's path is replaced with the path to the new temporary file that contains the replacement module definitions.

natefaubion: But you never case or cons on acc. As far as I can tell, `acc` is only ever `[]`.

natefaubion: If that isn't a bug, an `acc` is indeed unused, then I think this can just be a fold.

tests/TestPsci/TestEnv.hs
```diff
@@ -90,7 +95,37 @@ evaluatesTo command expected = runAndEval command evalJsAndCompare ignorePrinted
...
+printed :: String -> (String -> Expectation) -> TestPSCi ()
+printed command f = runAndEval command (void jsEval) (liftIO . f)
+
+simulateModuleEdit :: P.ModuleName -> FilePath -> TestPSCi a -> TestPSCi a
+simulateModuleEdit mn newPath action = do
+  ms <- asks psciFileGlobs
+
+  case findReplace ms [] 0 of
+    Nothing  -> fail $ "Did not find " ++ inputPath ++ " in psciFileGlobs"
+    Just xs' -> local (replacePath xs') temporarily <* rebuild
```
rndnoise: I think this where `acc` is used. Well, `xs` is `reverse acc` and the `newPath` tacked in front. It's used as the value for `psciFileGlobs` in the reader environment value.

tests/TestPsci/TestEnv.hs
rndnoise: Oh, yes. Shows how it's easy to be blind to your own bugs.

rndnoise: I'm surprised that it didn't cause a problem, but I don't know exactly how `psciFileGlobs` is used by the compiler. Maybe since all the module files are compiled before the test even runs, there isn't any need to have their file paths?

Since I don't really know, it seems safer to me to change the function so it returns the big list of `psciFileGlobs` that it was given as input, with just a couple entries changed (as I originally intended). I'll try that and let you know if it works, but you may be right that all the other file paths that are accidentally being discarded actually ought to be.

rndnoise: Ok, leaving those paths in `psciFileGlobs` results in the compiler saying `Data.String.Regex.Unsafe`, a module I haven't "edited", can't import a symbol from `Data.String.Regex`, the module whose code I swapped with the file in `examples/psci/Regex.edit`. That makes sense, it tried to recompile a module that depended on the one I changed and `:reloaded`.

So unless you object, it seems like I should commit the fixed `findReplace` and edit/reload a different module that isn't imported by something else, to avoid another module complaining that I haven't defined all the symbols it wanted to import.

Since I haven't used PureScript in a bit, can you suggest any modules that aren't imported somewhere else in the test support libraries?

rndnoise: Using a fold will probably be simpler, but maybe it will have unintended consequences later on. Your thoughts?

natefaubion: Would it be possible to import and edit a small example included in the test directory, rather than a library?

> Using a fold will probably be simpler, but maybe it will have unintended consequences later on. Your thoughts?

I'm not sure exactly what you mean. I think the main thing is that it's just working as intended. I could maybe offer a suggestion given the fixed implementation.

rndnoise: Let me know if the implementation of `findReplace` can be simplified, but I don't think the original definition was correct (it should have cons'd onto`acc`).

I took your suggestion and made a standalone module in `tests/support/psci/Reload.purs` and the edited version is `Reload.edit`. I also brought this branch up-to-date with master.

natefaubion: I think the only reason you have it return Maybe is for a sanity check, otherwise you can just do it with `fmap`. I think it's fine. I might use a Boolean instead of Int since you only check against 0.

rndnoise: Yes, I couldn't talk myself into getting rid of the sanity check after it did catch a mistake when I came back to this PR having forgotten about it. But I think I've make it better in this last commit.

kRITZCREEK:
Thanks!

Issue #1493, opened 2015-09-28
Efficient object updates for closed records


garyb:
We can use type information to generate the object construction code directly.

PR #3244, opened 2018-02-09
Efficient object updates for closed records 

fehrenbach:
See #1493.

This uses @matthewleon 's newly introduced CoreFn optimization infrastructure (#3218).

### DIFF
--- a/examples/passing/PartialTCO.purs
+++ b/examples/passing/PartialTCO.purs
@@ -1,0 +1,13 @@
+module Main where
+
+import Prelude
+import Control.Monad.Eff.Console (log)
+import Partial.Unsafe (unsafePartial)
+
+main = do
+  let _ = partialTCO true 1000000
+  log "Done"
+
+partialTCO :: Partial => Boolean -> Int -> Int
+partialTCO true 0 = 0
+partialTCO true n = partialTCO true (n - 1)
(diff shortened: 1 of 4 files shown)


kRITZCREEK:
We're making a trade-off here, right? This will increase the size of the generated JS depending on how large of a record we are copying.

Should we pick a number of record properties at which we switch back to the loop?

garyb:
I think we can set that threshold pretty high if we're going to have it, like 100 or something.

fehrenbach:
It certainly is a trade-off. The appropriate measure to optimize for generated code size would be something like sum of length of field names that are not updated compared to size of the update loop.

In my dreams, we would just generate `Object.assign({}, inputRecord, {up: "-dated", "field":'s'})` which is short and JS engines would be clever, recognizing this as a copy with updates, but alas they are not and it's ES2015 anyway.

I can add a threshold, in whatever form, but to be honest I'd rather wait for someone to actually have a problem with huge records.

fehrenbach:
For the record:
https://jsperf.com/copyobj
https://jsperf.com/copyobjhuge

garyb:
> I can add a threshold, in whatever form, but to be honest I'd rather wait for someone to actually have a problem with huge records.

I'd tend to agree with that 👍

garyb:
Also, sorry we're only getting around to commenting on this now - but big thanks! This is something I've wanted to do for a long time.

PR #3246, opened 2018-02-18
Fix :browse Prim in REPL, it has no ExternsFile

### LINKED ISSUE #2672: :browse Prim does not work
(Moved from purescript-psci-support per @paf31)

--------

The Purescript book says that primitive types are defined in the `Prim` module. So I wanted to see what functions were available. I can import `Prim`:

```
> import Prim
```

But `:browse` doesn't work:

```
> :browse Prim
Module 'Prim' is not valid.
```

That's the same message as for a completely bogus module. Why this message? If it's because `Prim` is somehow magic, should that be made clear?

rndnoise:
The reason `:browse Prim` failed is that there's not an entry in `psciLoadedExterns` for `Prim`. This fixes #2672

### DIFF
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -14,9 +14,10 @@ import           Prelude.Compat
 import           Protolude (ordNub)
 
 import           Data.List (sort, find, foldl')
-import           Data.Maybe (mapMaybe)
+import           Data.Maybe (fromMaybe, mapMaybe)
 import qualified Data.Map as M
 import           Data.Monoid ((<>))
+import qualified Data.Set as S
 import           Data.Text (Text)
 import qualified Data.Text as T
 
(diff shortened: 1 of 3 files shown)


garyb:
Seems good to me 👍 I'm not too familiar with the REPL code though, so anyone else want to comment?

src/Language/PureScript/Interactive.hs
```diff
@@ -291,22 +292,22 @@ handleBrowse
...
-    failNotInEnv modName =
-        print' $ T.unpack $ "Module '" <> N.runModuleName modName <> "' is not valid."
-    lookupUnQualifiedModName quaModName st =
-        (\(modName,_,_) -> modName) <$> find ( \(_, _, mayQuaName) -> mayQuaName == Just quaModName) (psciImportedModules st)
+    findMod needle externs imports =
+      let
+        qualMod = fromMaybe needle (lookupUnQualifiedModName needle imports)
+        primMod = P.ModuleName [P.ProperName "Prim"]
+        modules = S.fromList (primMod : (P.getModuleName . fst <$> externs))
+       in if qualMod `S.member` modules
```
kRITZCREEK: Something seems weird with the `let ... in` indentation here. I think the `in` should be one space to the left?

rndnoise: Sure! I looked for other examples of multi-line `let ... in ...`. I only found one or two, so I have emulated those. If that's not what you intended, I have no problem changing it.

PR #3249, opened 2018-02-21
Improve parse error messages for .purs-repl

### LINKED ISSUE #3248: Trailing blank line in `.purs-repl` fails, with a confusing error message
[@natefaubion suggested I move this here.]

I accidentally left a blank line at the end of my `.purs-repl` file. 

```purescript
import Prelude

```

You can't see the line, but it's there.  

The result of starting a repl is this:

```
503 $ pulp repl
PSCi, version 0.11.6
Type :? for help

import Prelude

(line 1, column 1):
unexpected end of input
expecting "import", declaration, expression or "let"
```

That's particularly confusing for a newbie because there's no indication where the problem is. (Puzzled me because it'd been a day since I'd fiddled with `.purs-repl` and I'd forgotten I did it.)

rndnoise:
This fixes #3248

- Set the file name in error message
- Parse whole file at once so correct lines numbers are given in errors
- Don't fail on empty lines, whitespace, etc

To ensure line numbers are given on errors, I made a new parser that only parses `import` statements, instead of parsing all kinds of commands and rejecting them after the line numbers are forgotten.

Since the whole file is parsed all at once, it no longer makes sense to interleave printing, parsing, and executing line-by-line. So I've printed the whole contents verbatim before executing all the imports (assuming the file parsed).

Example:

```sh
$ cat .purs-repl
import Prelude
:help

$ pulp repl
PSCi, version 0.11.7
Type :? for help

"/home/guest-user/example/.purs-repl" (line 2, column 1):
unexpected :
expecting "import" or end of input
The .purs-repl file only supports import declarations
```

### DIFF
--- a/app/Command/REPL.hs
+++ b/app/Command/REPL.hs
@@ -360,13 +360,10 @@ command = loop <$> options
                       configFile <- (</> ".purs-repl") <$> liftIO getCurrentDirectory
                       exists <- liftIO $ doesFileExist configFile
                       when exists $ do
-                        ls <- lines <$> liftIO (readUTF8File configFile)
-                        for_ ls $ \l -> do
-                          liftIO (putStrLn l)
-                          case parseCommand l of
-                            Left err -> liftIO (putStrLn err >> exitFailure)
-                            Right cmd@Import{} -> handleCommand' state cmd
-                            Right _ -> liftIO (putStrLn "The .purs-repl file only supports import declarations")
+                        cf <- liftIO (readUTF8File configFile)
+                        case parseDotFile configFile cf of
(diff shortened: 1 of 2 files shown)


kRITZCREEK:
Can we write a test for this? Just to make sure it doesn't regress.

rndnoise:
Would it suffice to test `parseDotFile`? Testing the code which uses the parser, [here](https://github.com/rndnoise/purescript/blob/4371e5ba3478715a2c18b9d5ba8e2dced6f79f2c/app/Command/REPL.hs#364), seems difficult without making more changes to that code. If you want to test that stuff in `command`, I'll need some advice on how to go about it.

PR #3255, opened 2018-02-25
Allow errors to carry multiple positions

garyb:
As discussed in https://github.com/purescript/purescript/issues/3208#issuecomment-360127259

I guess maybe the approach here is less good than exactly what I suggested in there, as it will break the JSON error format, whereas that would be backwards compatible?

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -198,7 +198,7 @@ data ErrorMessageHint
   | ErrorInTypeClassDeclaration (ProperName 'ClassName)
   | ErrorInForeignImport Ident
   | ErrorSolvingConstraint Constraint
-  | PositionedError SourceSpan
+  | PositionedError (NEL.NonEmpty SourceSpan)
   deriving (Show)
 
 -- | Categories of hints
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -47,7 +47,7 @@ import qualified Text.PrettyPrint.Boxes as Box
 newtype ErrorSuggestion = ErrorSuggestion Text
 
 -- | Get the source span for an error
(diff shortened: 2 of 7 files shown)


kRITZCREEK:
I'd prefer

```
{
position :: (Maybe) ErrorPosition,
additionalPositions :: [ErrorPosition]
...
}
```

for the JSON format. For one we should get rid of the `Maybe` in there once we've fixed all the error messages without spans. I'd like to use a datatype isomorphic to NEL to collect the spans, but with different names for the operations. Having `cons` overwrite the position that will be presented to the User is a little unintuitive. I'd rather have an explicit `setPresentedSpan` which pushes the current presented one into the tail (cons), and an operation `addSpan` which pushes directly into the tail.

garyb:
> I'd like to use a datatype isomorphic to NEL to collect the spans, but with different names for the operations. Having `cons` overwrite the position that will be presented to the User is a little unintuitive. I'd rather have an explicit `setPresentedSpan` which pushes the current presented one into the tail (cons), and an operation `addSpan` which pushes directly into the tail.

We can do that, but I'm not sure it's necessary really? We're not going to be modifying positions after we've created the error. The fact we currently decorate errors with positions after the fact sometimes is just because we haven't finished the work to require it in the error constructor yet.

garyb:
(Also the choice of position to show is entirely arbitrary, since the error happens everywhere equally)

kRITZCREEK:
Yeah, I don't feel strongly about the NEL wrapper at all, but I do think we shouldn't break the error format here, since the thing we want to communicate is still the always existing primary error.

garyb:
Updated to preserve the original error format, adding an `allSpans` field that will carry all available position info. I could have popped the `head` out of there, but figured maybe providing all the spans in the same format might be more convenient for tooling to deal with when it's updated to accommodate multi-position errors.

PR #3266, opened 2018-03-05
Makefiles make things easier for new and regular contributors alike

bitemyapp:
I use a `Makefile` in almost all of my software projects as a sort of exobrain/secondary memory and regularized shortcut mechanism for doing common tasks. PureScript didn't have one, so I've PR'd one based on my small but burgeoning collection at https://github.com/bitemyapp/makefiles from which https://github.com/bitemyapp/lefortovo grabs things.

I also updated the `.gitignore` to ignore profiling bumf.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,14 @@ tmp/
 output
 examples/docs/docs/
 core-tests/full-core-docs.md
+tests/support/package-lock.json
 .psc-ide-port
 .psc-package/
 purescript.cabal
+
+# Profiling related
+*.aux
+*.hp
+*.prof
+*.ps
+*.svg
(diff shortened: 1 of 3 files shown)


kRITZCREEK:
I probably won't use this much as I don't have make installed on Windows, but having a place to collect a few commonly run commands doesn't sound like a bad idea, and if it's executable for most that's a plus, so :+1: from me.

bitemyapp:
@kRITZCREEK does this need anything else to be merged?

parsonsmatt:
I forgot that this even existed when I made my own makefile PR :open_mouth:

@bitemyapp mind if I PR this branch with my `ghcid`, `ghcid-test`, and `help` commands?

bitemyapp:
@parsonsmatt go ahead!

parsonsmatt:
I failed to use GitHub correctly and now my PR to your PR is borked, so I'll just update my PR after this one lands.

Unless PS folks want to do it all in one go, at which point I will retry.

garyb:
I don't mind if we do it in one or add it afterwards. This looks good to me too! @kRITZCREEK got any comments?

PR #3270, opened 2018-03-07
[WIP] Smarter make

kRITZCREEK:
This significantly improves startup time on incremental builds with lots of already built modules.

I'll still move things around and document what's happening here, but this change can be backported to 0.11.7 as is, and I'd like people to try that and see what kind of numbers they get, and if their projects still build properly. Thanks!

### DIFF
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -23,6 +23,7 @@ module Language.PureScript.Make
 import           Prelude.Compat
 
 import           Control.Concurrent.Lifted as C
+import           Control.Applicative ((<|>))
 import           Control.Monad hiding (sequence)
 import           Control.Monad.Base (MonadBase(..))
 import           Control.Monad.Error.Class (MonadError(..))
@@ -38,7 +39,7 @@ import           Data.Aeson (encode, decode)
 import           Data.Either (partitionEithers)
 import           Data.Function (on)
 import           Data.Foldable (for_)
-import           Data.List (foldl', sortBy, groupBy)
+import           Data.List (foldl', sortBy, groupBy, find)
 import           Data.Maybe (fromMaybe, catMaybes)
 import           Data.Monoid ((<>))
 import           Data.Time.Clock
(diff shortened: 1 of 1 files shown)


kRITZCREEK:
I timed some builds and wrote down my results:

https://gist.github.com/kRITZCREEK/c2070885d70cdcb047866ed395a06cb7

It's a pretty clear improvement.

kRITZCREEK:
Who can I ask to take a look at this? :) I need this to go in before I can start working on only parsing module headers to build the module graph.

garyb:
Unfortunately the diff is pretty much useless, I'm inclined to say :shipit: since you've been testing it for a while without any trouble (right? :wink:).

src/Language/PureScript/Make.hs
```diff
@@ -305,175 +165,3 @@ inferForeignModules =
       if exists
         then return (Just jsFile)
         else return Nothing
-
--- | A set of make actions that read and write modules from the given directory.
-buildMakeActions
```
kRITZCREEK: Everything downwards from here moved into `Make.Actions`

src/Language/PureScript/Make.hs
```diff
@@ -206,86 +131,21 @@ make ma@MakeActions{..} ms = do
...
-  shouldExist _ = internalError "make: dependency should already have been built."
-
-  decodeExterns :: Externs -> Maybe ExternsFile
-  decodeExterns bs = do
-    externs <- decode bs
-    guard $ T.unpack (efVersion externs) == showVersion Paths.version
-    return externs
-
--- | A monad for running make actions
-newtype Make a = Make
```
kRITZCREEK: This was moved into `Make.Monad`

src/Language/PureScript/Make.hs
```diff
@@ -171,21 +90,27 @@ make ma@MakeActions{..} ms = do
 
   (sorted, graph) <- sortModules ms
 
-  barriers <- zip (map getModuleName sorted) <$> replicateM (length ms) ((,) <$> C.newEmptyMVar <*> C.newEmptyMVar)
```
kRITZCREEK: The real change in this PR is this section, which is now broken down between `Make.BuildPlan` and here.

Instead of checking the timestamps of the dependencies of a module before building it, we now check all the timestamps once, and make a plan of which modules need to be rebuilt (`BuildPlan.construct`).

kRITZCREEK:
I've successfully used this for a while now, so I'm confident it works. I've added a few comments to differentiate noise from actual change in the PR. Any remaining questions?

garyb:
Sorry, I meant to comment back here - I did actually read it a bit after you added the comments, and yeah it seems reasonable to me!

PR #3271, opened 2018-03-07
Explaining the custom Eq, Ord instances for Binder

bitemyapp:


kRITZCREEK:
I've fixed a typo, and stretched the comment to 80 characters in 194e1b3. Thanks!

PR #3273, opened 2018-03-10
Take advantage of multi-span errors

garyb:
This fixes `CycleInModule ` not having many positions, and adds all positions for `DuplicateModule` using the usual form rather than special handling.

### DIFF
--- a/src/Language/PureScript/ModuleDependencies.hs
+++ b/src/Language/PureScript/ModuleDependencies.hs
@@ -7,7 +7,7 @@ module Language.PureScript.ModuleDependencies
 import           Protolude hiding (head)
 
 import           Data.Graph
-import           Data.List (head)
+import qualified Data.List.NonEmpty as NEL
 import qualified Data.Set as S
 import           Language.PureScript.AST
 import qualified Language.PureScript.Constants as C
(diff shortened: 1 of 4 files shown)


garyb:
This was broken yesterday, but I've fixed it now (missed a case where a NEL was needed).

src/Language/PureScript/ModuleDependencies.hs
```diff
@@ -62,5 +62,5 @@ toModule (AcyclicSCC m) = return m
 toModule (CyclicSCC [m]) = return m
 toModule (CyclicSCC ms) =
   throwError
-    . errorMessage' (getModuleSourceSpan (head ms))
+    . errorMessage'' (NEL.fromList (fmap getModuleSourceSpan ms))
```
kRITZCREEK: Could we avoid the partial NEL.fromList and just pattern match the head out of `ms`?

garyb: I mean, we can... I just figured since we can easily see ms is not empty from the pattern already being covered that it probably wasn't necessary?

kRITZCREEK: I mean we'd just move the partiality into the top level pattern match as that wouldn't match the empty list anymore, but I think fromList is one of those unexpectedly partial functions like maximum...

PR #3276, opened 2018-03-11
Error positions for FFI related errors

garyb:


kRITZCREEK:
I looked at this before, does this PR just use the full PureScript module span for all the errors then? I looked into the JS parser and the JSAnnot constructor, but I guess we just want the error span to appear in the file that caused the rebuild?

garyb:
They use the original PS module span yes.

We can always improve the spans later, just ensuring they have any sensible positions is the goal for now. `MissingFFIImplementations` is the only one we could give anything more specific than a module span for anyway, unless we were to start outputting positions for the JS files (which is also an option...)

kRITZCREEK:
Yeah I agree, just getting some sensible spans in there is the immediate goal.

PR #3277, opened 2018-03-11
Add source spans for literals

garyb:
This fixes the IntOutOfRange error (again 😄)

As it happens, I think it was probably unnecessary to add these, since the real problem was `literalToValueJS` putting `Nothing` as the span for literals constantly in the JS AST... but including the spans will at least ensure the positions are accurate, and it's another step to ensuring we always have good position information.

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -438,12 +438,12 @@ parseLet = do
 parseValueAtom :: TokenParser Expr
 parseValueAtom = withSourceSpan PositionedValue $ P.choice
                  [ parseAnonymousArgument
-                 , Literal <$> parseNumericLiteral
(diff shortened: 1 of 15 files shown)


src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -570,8 +571,8 @@ parseConstructorBinder = withSourceSpanF $
     <*> many (indented *> parseBinderNoParens)
 
 parseObjectBinder:: TokenParser Binder
-parseObjectBinder =
-  LiteralBinder <$> parseObjectLiteral (indented *> parseEntry)
+parseObjectBinder = withSourceSpanF $
+  (flip LiteralBinder) <$> parseObjectLiteral (indented *> parseEntry)
```
kRITZCREEK: could drop the parens around `flip LiteralBinder`

PR #3282, opened 2018-03-19
[purs ide] Find usages for recursive definitions

kRITZCREEK:
Previously we'd fail to find these because the name is already in "scope", but for recursive definitions that name still refers to the top level declaration and so we should include it as a usage.


kRITZCREEK:
Thanks Nicholas! I'd say you're probably the most qualified to review this so unless anyone else has a comment I'll merge the PR this evening.

PR #3289, opened 2018-03-24
Implements magic-do for Control.Monad.Effect

kRITZCREEK:
Simplest thing I could come up with. This adds another pass over the JS AST, so if that ends up being to slow I can look into merging these into a single pass, or checking for an import of Control.Monad.Eff/Control.Monad.Effect before doing the traversal.

### DIFF
--- a/src/Language/PureScript/CoreImp/Optimizer.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer.hs
@@ -37,7 +37,9 @@ optimize js = do
       [ inlineCommonValues
       , inlineCommonOperators
       ]) js
-    untilFixedPoint (return . tidyUp) . tco . inlineST =<< untilFixedPoint (return . magicDo) js'
+    untilFixedPoint (return . tidyUp) . tco . inlineST
+      =<< untilFixedPoint (return . magicDo')
+      =<< untilFixedPoint (return . magicDo) js'
   where
     tidyUp :: AST -> AST
     tidyUp = applyAll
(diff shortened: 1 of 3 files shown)


garyb:
We're definitely not doing the rename to `Effect` just now then?

kRITZCREEK:
It's easy enough to change in here if we decide to do it before 0.12, so I'll merge this and we can take a look at renaming the `Control.Monad.Effect` module to `Effect` later.

src/Language/PureScript/CoreImp/Optimizer.hs
```diff
@@ -37,7 +37,9 @@ optimize js = do
       [ inlineCommonValues
       , inlineCommonOperators
       ]) js
-    untilFixedPoint (return . tidyUp) . tco . inlineST =<< untilFixedPoint (return . magicDo) js'
+    untilFixedPoint (return . tidyUp) . tco . inlineST
+      =<< untilFixedPoint (return . magicDo')
+      =<< untilFixedPoint (return . magicDo) js'
```
natefaubion: I know you already merged this, but since this is just a pure fn, could this not just be `untilFixedPoint (return . magicDo' . magicDo)` just to lose the extra fixed point pass (which can be expensive).

kRITZCREEK: I wasn't entirely sure if it was safe to interleave the two transformations. In the rare case someone mixes the two in the same module, but I guess the typechecker would've rejected the program at that point...

PR #3290, opened 2018-03-25
Add a Makefile with ghcid and testing commands

parsonsmatt:
With this PR, you can do

    $ make ghcid-test

And it'll load up PureScript in`ghcid` and run the tests every time it detects a file change.

I've been playing around with it and it is pretty fantastic :joy: 

To make it work, I had to put `import Prelude` in some of the test modules.


kRITZCREEK:
Looks good to me, thanks!

PR #3291, opened 2018-03-25
Disallow Prim modules

### LINKED ISSUE #3254: Disallow defining a module named `Prim`
No fun for you @LiamGoodacre 😛

parsonsmatt:
Fixes #3254

### DIFF
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -167,7 +167,7 @@ make :: forall m. (Monad m, MonadBaseControl IO m, MonadError MultipleErrors m, 
      -> [Module]
      -> m [ExternsFile]
 make ma@MakeActions{..} ms = do
-  checkModuleNamesAreUnique
+  checkModuleNames
 
   (sorted, graph) <- sortModules ms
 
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/Make.hs
```diff
@@ -188,6 +188,20 @@ make ma@MakeActions{..} ms = do
...
+  checkModuleNames :: m ()
+  checkModuleNames = checkNoPrim *> checkModuleNamesAreUnique
+
+  checkNoPrim :: m ()
+  checkNoPrim =
+    for_ ms $ \m -> do
+      case getModuleName m of
+        mn@(ModuleName (ProperName "Prim" : _)) ->
+          throwError
+            . errorMessage'' (pure (getModuleSourceSpan m))
```
garyb: You can just use `errorMessage'` and drop the `pure` here (these names are terrible, once we drop unpositioned errors entirely I'll give them better names :wink:)

garyb:
Could we add a test for a `Prim` submodule too? I can see from the code it should definitely work as it is now, but it'd be nice to have to help avoid regressions.

parsonsmatt:
Added another test :smile:

Issue #3199, opened 2018-01-10
Minor documentation update to RowCons


vladciobanu:
I would like to do a minor update to the `RowCons` documentation in [Prim.hs](/purescript/purescript/blob/master/src/Language/PureScript/Docs/Prim.hs#L247):

`RowCons l a i o` can be read as `o ~ { l :: a | i }`

Perhaps also add a brief description saying what each of l, a, i and o represent. I know it took me a bit to figure out. I'll add a proposal for the changes tonight.

hdgarrood:
Additionally, we could probably use words for some of those type variables while we're at it:

```purescript
class RowCons (label :: Symbol) (a :: Type) (input :: # Type) (output :: # Type) | [...]
```

Also, I've just realised this should probably wait until #3176 is merged, in order to avoid causing merge conflicts.

joneshf:
As mentioned i n the chat, it should be `o ~ ( l :: a | i )`, since it works for any row of types, not just records.

LiamGoodacre:
As `RowCons` also works with `l o -> a i`, I'd suggest `tail` and `row` instead of `in_row` and `out_row`.  I feel this would be less confusing - if you are using it this way but still have to read it as `in` and `out`.

vladciobanu:
[Environment.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Environment.hs#L378) changes from 
`class RowCons (l :: Symbol) (a :: Type) (i :: # Type) (o :: # Type) | l a i -> o, l o -> a i`
to 
`class RowCons (label :: Symbol) (a :: Type) (tail :: # Type) (row :: # Type) | label a tail -> row, label row -> a tail`

[Prim.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/Prim.hs#L247) changes from
>The RowCons type class is a 4-way relation which asserts that one row of types can be obtained from another by inserting a new label/type pair on the left.

to

>The RowCons type class is a 4-way relation which asserts that one row of types can be obtained from **a "tail" row** by inserting a new label/type pair to the **tail (`row ~ (label :: a | tail)`)**.

Not super happy with _a "tail" row_ bit, but I'm not sure how to phrase it more elegantly. I'll wait with the PR until #3176 is merged.

PR #3292, opened 2018-03-25
updated docs for RowCons and RowUnuon

vladciobanu:
Now that #3176 is committed, I can open this PR for #3199 

I took the liberty to update `RowUnion` too, but I can revert it if people are not happy with it (or change the names if needed).

Should I also update `CONTRIBUTORS.md`? I don't really feel this is "worthy" of it, but if it's relevant for licensing/rights, I'm happy to update it as well.


vladciobanu:
Updated `Nub` class with more descriptive names and updated CONTRIBUTORS as well. This should be good to go.

PR #3293, opened 2018-03-28
Add compiler derived Prim.Row.Nub class

natefaubion:
This adds a compiler derived class for removing duplicates from rows. This is technically possible to achieve using `RowListNub`, but the machinery is _substantial_ and I think this is a common enough usecase to warrant a primitive class. This can be used to implement a more type-safe record merge than a bare `Union`.

### DIFF
--- a/examples/passing/RowNub.purs
+++ b/examples/passing/RowNub.purs
@@ -1,0 +1,23 @@
+module Main where
+
+import Control.Monad.Eff.Console (log)
+import Prim.Row (class Nub, class Union)
+import Type.Row (RProxy(..))
+
+nubUnion
+  :: forall r1 r2 r3 r4
+   . Union r1 r2 r3
+  => Nub r3 r4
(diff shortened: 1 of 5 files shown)


garyb:
Seems like a good idea to me!

natefaubion:
I would also like to do the same for `Lacks` in the same vein as `Union`. `Lacks` is currently implemented in terms of `Union` and `RowCons`, but the encoding leaks when it comes to errors, and it can be difficult to decipher (where is this `Entry` type coming from?). I realize there's less of a benefit here, but I think it's worth it.

kRITZCREEK:
If we merge https://github.com/purescript/purescript/pull/3292 first, do we want to update the type parameters to be a little more descriptive in here as well?

vladciobanu:
I've no problem to fix merge issues / update the type params for Nub on #3292 if you merge this first.

Issue #3284, opened 2018-03-22
Build failure with GHC 8.4.1


ilovezfs:
```
==> Installing purescript
/usr/bin/sandbox-exec -f /tmp/homebrew20180322-21207-q4ylnw.sb nice /usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.3.3/bin/ruby -W0 -I /usr/local/Homebrew/Library/Homebrew -- /usr/local/Homebrew/Library/Homebrew/build.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/purescript.rb --verbose --debug
/usr/local/Homebrew/Library/Homebrew/build.rb (Formulary::FromPathLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/purescript.rb
/usr/local/Homebrew/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/cabal-install.rb
/usr/local/Homebrew/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/ghc@8.2.rb
/usr/local/Homebrew/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/sphinx-doc.rb
/usr/local/Homebrew/Library/Homebrew/build.rb (Formulary::FormulaLoader): loading /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/ghc.rb
==> Downloading https://github.com/purescript/purescript/archive/v0.11.7.tar.gz
Already downloaded: /Users/joe/Library/Caches/Homebrew/purescript-0.11.7.tar.gz
==> Verifying purescript-0.11.7.tar.gz checksum
tar xzf /Users/joe/Library/Caches/Homebrew/purescript-0.11.7.tar.gz
==> cabal sandbox init
Config file path source is default config file.
Config file
/private/tmp/purescript-20180322-21209-1tu3lws/purescript-0.11.7/.cabal/config
not found.
Writing default configuration to
/private/tmp/purescript-20180322-21209-1tu3lws/purescript-0.11.7/.cabal/config
Writing a default package environment file to
[...]

PR #3294, opened 2018-03-28
Restrict base for GHC < 8.4.1

parsonsmatt:
This PR addresses #3284 with a stop-gap measure, explicitly expressing the upper bound on `base` that PureScript currently has.

When PureScript officially supports 8.4.1, this can be dropped.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -35,7 +35,7 @@ dependencies:
   - aeson >=1.0 && <1.3
   - aeson-better-errors >=0.8
   - ansi-terminal >=0.7.1 && <0.8
-  - base >=4.8 && <5
+  - base >=4.8 && <4.11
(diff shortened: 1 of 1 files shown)


ilovezfs:
@kRITZCREEK now that 0.12.0 is out, is there any update on 8.4.x support?

kRITZCREEK:
The last time I tried libraries we depended on still didn't build with 8.4.1, but I can give it another try today.

ilovezfs:
Yeah I retested the builds of everything in Homebrew that was using 8.4.2 when we upgraded to 8.4.3 and nothing broke so I wouldn't anticipate a problem.

kRITZCREEK:
see #3372, just one library missing :)

PR #3295, opened 2018-03-29
Allow functor derivation under quantification

i-am-tom:
When "functor variables" exist within a quantifier, they don't seem to
derive properly:

```
data T a = T (forall x. a)
```

This commit (hopefully!) fixes this, allowing the compiler to drill down
through these skolem scopes.


i-am-tom:
Also now does constraints!

PR #3301, opened 2018-04-12
Update `Effect` inlining

garyb:
For https://github.com/purescript/purescript-effect/pull/2

Also looks like we were missing magic for the `mkEffectFn` functions, so I added support for those too.


kRITZCREEK:
Good spot!

PR #3309, opened 2018-04-14
Remove inlining for `mod` for `Int`

garyb:
This is for purescript/purescript-prelude#161 - the new mod definition for int is no longer straightforward (currently `x % y`, new one will be `Math.abs(((x % y) + y) % y)`), so the advantage of inlining is probably minimal now. As a benefit, this also means that the issue I raised in there about the mod behaviour differing between compiler versions on the same library will not be the case, since this won't be overriding the FFI implementation provided in there.


garyb:
I'm going to go ahead and merge this, so we can get the corresponding part in the Prelude wrapped up. It shouldn't actually break anything even if it is wrong (and I'm pretty confident it's not) - it will just generate worse code.

PR #3311, opened 2018-04-19
Add missing output directory to error

rgrinberg:


LiamGoodacre:
Please add yourself to the CONTRIBUTORS.md file.
Don't worry about the failing CI - we're currently in a transition period for the next compiler release.
Thanks!

src/Language/PureScript/Ide.hs
```diff
@@ -167,7 +167,7 @@ findAvailableExterns :: (Ide m, MonadError IdeError m) => m [P.ModuleName]
 findAvailableExterns = do
   oDir <- outputDirectory
   unlessM (liftIO (doesDirectoryExist oDir))
-    (throwError (GeneralError "Couldn't locate your output directory."))
+    (throwError (GeneralError $ "Couldn't locate your output directory: " <> (toS oDir) <> "."))
```
kRITZCREEK: What do you think of  `Couldn't locate your output directory at: ...`? Makes it a little clearer that we went searching for it.

Could we run the path through `System.FilePath.normalise` and use an explicit `T.pack` over `toS`? I also think we should leave off the trailing `.` since it'll be in the way when copy pasting the path.

rgrinberg: Makes sense.

kRITZCREEK:
Great, thanks!

Issue #2621, opened 2017-02-04
Change the require path in the JS output of psc


Pauan:
Let's say you have two PureScript modules: `Foo` and `Bar`, and `Foo` imports `Bar`. The JS output will look something like this:

```
// output/Foo/index.js
var Bar = require("../Bar");
```

```
// output/Bar/index.js
```

This relies upon the fact that in Node.js if you try to `require` a directory it will automatically add `/index.js` to end of the path.

However, this causes problems with `rollup-plugin-purs`, because it means that `rollup-plugin-purs` must use the [`rollup-plugin-node-resolve`](https://github.com/rollup/rollup-plugin-node-resolve) plugin.

But the `rollup-plugin-node-resolve` plugin causes npm modules from `node_modules` to also be included in the bundle, which [might not be what you want](https://github.com/purescript/purescript/issues/2545#issuecomment-277292588).

That means that `rollup-plugin-purs` cannot be used as a drop-in replacement for `psc-bundle` if people *don't* want their npm modules to be bundled (for example if they are writing a Node.js application).

The fix in `psc` is very easy: it just needs to add `/index` to the end of the `require` path, like this:

```
// output/Foo/index.js
var Bar = require("../Bar/index");
```

Now `rollup-plugin-purs` no longer needs to use `rollup-plugin-node-resolve`, so npm dependencies are no longer included.

This does require `psc-bundle` to be changed so that it understands `/index`, and there may be other tools that will need to be changed as well.

In the meantime, I'll add a workaround in `rollup-plugin-purs`, so this issue isn't high priority, but I think it would be nice to get it fixed eventually.

nwolverson:
There isn't any downside to this change is there? There would be a change in psc and psc-bundle, I can't think of any other tool which should be treating these specially rather than just resolving the require

Pauan:
@nwolverson I don't know. It's possible that other tools like `psci`, `psc-ide-server`, IDE plugins, etc. might break too, but I haven't tested it.

damncabbage:
A possibly relevant previous discussion: https://github.com/purescript/purescript/issues/1698

(TBC I wasn't advocating for the change with my issue; the title was edited by someone else later in the thread.)

Pauan:
The new version `1.0.14` of `rollup-plugin-purs` contains a workaround, so it's no longer necessary to use `rollup-plugin-node-resolve`

It's a bit of a hack, though, so I'd still like to see this fixed in `psc` eventually, but there's no rush.

[...]

PR #3312, opened 2018-04-19
Move solved type classes into Prim modules & bump test deps

LiamGoodacre:
WIP
Can split this into two PRs if that makes more sense but things aren't going to work until the test deps are consistent.
Next I'll update our tests to switch from Eff to Effect.
Resolves https://github.com/purescript/purescript/issues/3179

PR #3314, opened 2018-04-21
Include full file path in CJS requires.

chexxor:
Fixes https://github.com/purescript/purescript/issues/2621

Changes the JS output from `require("../SomeModule")` to `require("../SomeModule/index.js")`, and same for the implicit FFI imports.

`stack test` is proving difficult. First time I ran it, it had trouble installing bower dependencies, giving `ECONFLICT Unable to find suitable version for purescript-either`. After manually installing them, I got errors about `Monoid` being a duplicate module, which occurred because it's been moved to Prelude in the latest version (so we may need to adjust the version range in the bower file). After manually deleting Monoid stuff from the PS Prelude it was using, I get test failures caused by stuff missing from Prim -- `Module Prim.RowList was not found.`, and same for Prim.Ordering and Prim.Symbol.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -104,7 +104,8 @@ moduleToJs (Module coms mn _ imps exps foreigns decls) foreign_ =
   importToJs :: M.Map ModuleName (Ann, ModuleName) -> ModuleName -> m AST
   importToJs mnLookup mn' = do
     let ((ss, _, _, _), mnSafe) = fromMaybe (internalError "Missing value in mnLookup") $ M.lookup mn' mnLookup
-    let moduleBody = AST.App Nothing (AST.Var Nothing "require") [AST.StringLiteral Nothing (fromString (".." </> T.unpack (runModuleName mn')))]
+    let moduleBody = AST.App Nothing (AST.Var Nothing "require")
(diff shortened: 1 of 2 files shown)


chexxor:
Ah, I bet some of these problems are fixed in https://github.com/purescript/purescript/pull/3312, as the deps I've been tinkering with appear to have been bumped and fixed in that PR. Also, it looks like stuff missing from Prim could be fixed by that PR, as it's moving them from lib to compiler (if I read it right).

The other question I have is about tests -- do we need to have some tests written to assert this? If you ask me, I don't think it's necessary.

garyb:
Yeah, everything will be failing until we finish with #3312, that's the problem with #3309 too.

The tests will be subject to the changes you've made here already, so no, there's no further specific tests I can think of that will be necessary :)

chexxor:
Here's why I want this patch, BTW. https://www.screencast.com/t/KgVyC3ksG

No need to add a bundle or transpile step PureScript code during development -- just refer directly to the single PS module and the dependencies will be dynamically loaded. (Yeah, SystemJS does transpile, but it seems to be much easier to get started with, I think.)

The bundling step is often a hurdle for Haskell people coming to PureScript. Yes, tooling like `pulp --to x.js` can help with that, but it complicates the project set-up. Also this patch to the compiler doesn't break anything and is a benefit-only change -- it adds the ability to very easily use SystemJS in-browser loader instead of a server-side bundler.

PR #3315, opened 2018-04-22
Add real error messages for associativity errors

### LINKED ISSUE #3310: Mixing operators of the same precedence but different fixity results in a crash in master
```
purs: An internal error occurred during compilation: "matchTypeOperators: cannot reorder operators"
```

This is no doubt due to my meddling with the operator re-bracketing stuff that happened when adding positions to parts of the AST. I actually don't think it's a bad idea to error in these situations, forcing parens to be used the way Haskell does. Just maybe have a real error rather than a crash :wink:

This was picked up in `purescript-parsing` as interaction between `<|>` (infixl) and `<?>` (infix)

garyb:
Resolves #3310 

The code for generating the errors is kinda absurd as it's more complicated than the actual rebracketing code (since using parsec hides a bunch of that), but whatyagonnado.


examples/failing/MixedAssociativityError.purs
```diff
@@ -0,0 +1,6 @@
+-- @shouldFailWith MixedAssociativityError
+module Main where
+
+import Prelude
+
+feq f x y = f <$> x == f <$> y
```
garyb: This case was a little surprising, but GHC has an error for it too:

```
Prelude> let f x y = f <$> x == f <$> y

<interactive>:10:13:
    Precedence parsing error
        cannot mix ‘<$>’ [infixl 4] and ‘==’ [infix 4] in the same infix expression

<interactive>:10:13:
    Precedence parsing error
        cannot mix ‘==’ [infix 4] and ‘<$>’ [infixl 4] in the same infix expression
```

examples/failing/NonAssociativeError.purs
```diff
@@ -0,0 +1,8 @@
+-- @shouldFailWith NonAssociativeError
+-- @shouldFailWith NonAssociativeError
+module Main where
+
+import Prelude
+
+a = true == true == true
```
garyb: GHC again:
```
Prelude> True == True == True

<interactive>:11:1:
    Precedence parsing error
        cannot mix ‘==’ [infix 4] and ‘==’ [infix 4] in the same infix expression
```

hdgarrood:
I think we’ll need to update the docs for fixities now - it sounds to me like the docs I wrote might not all still apply, and there might be cases where the examples in the docs now produce these errors.

garyb:
Ah ok, I'll open an issue over there as a reminder.

Issue #3297, opened 2018-04-04
Parsing where clauses


joneshf:
Something I noticed when working on `purty` is that `where` clauses are parsed and desugared at the same time: https://github.com/purescript/purescript/blob/80cef847264f4124ca20cce01e2657998bcaaeaa/src/Language/PureScript/Parser/Declarations.hs#L72-L81

Did some digging around and didn't find any strong decision for choosing to desugar as well as parse. The original PR didn't mention anything about it, anyway: https://github.com/purescript/purescript/pull/318. I did see some discussion about desugaring and parsing happening at the same time: https://github.com/purescript/purescript/issues/3023.

How do we feel about adding an explicit case to `Expr` for `where` clauses? Is it something we tried before, but didn't like? Has it not been a big concern, so nobody has done anything about it? It would help immensely in `purty`, but recognize that `purty`s case is kind of specific.

garyb:
I'd be fine with introducing an explicit `where` constructor, we could perhaps change it's behaviour to match Haskell then too. It was never mentioned in the original PR as I barely knew what I was doing back then :wink: and yeah it hasn't really come up since.

kRITZCREEK:
There are a few more of these things that get desugared in the parser. I'd like these to also have a place in the AST and be desugared during desugaring.

- Record puns
- Multi-argument-lambdas

I can't think of more right now, but I think there were a few, I'll add to the list if I can think of them.

joneshf:
Cool! What's a good way to approach this? I was thinking:

* Add an additional case  `Expr` like the `Let` case (since it's no different from `Let` currently)

    ```Haskell
    data Expr
        = ...
        | Let [Declaration] Expr
        | Where [Declaration] Expr
        ...
    ```

* Change parsing to make `Where` instead of `Let`: https://github.com/purescript/purescript/blob/662da01c9f73873fcc2c4577bbc58fcc3f776f41/src/Language/PureScript/Parser/Declarations.hs#L72-L81

* Add a desugaring pass before let-binding desugaring: https://github.com/purescript/purescript/blob/662da01c9f73873fcc2c4577bbc58fcc3f776f41/src/Language/PureScript/Sugar.hs#L55-L73

That seem alright?

[...]

PR #3317, opened 2018-04-23
Track where clause metadata

joneshf:
Re: #3297 

Opening this PR to get some feedback on this approach.

* I'm not at all attached to the names. If we'd prefer different names, I have no problem replacing them.
* I added some metadata to the `CoreFn` stuff. I think I got all the places, but might have missed something. Is this the right idea? Is that the right way to do it if so?
* I couldn't get the tests to run, so this branch will almost surely fail on CI. There's some conflict with `purescript-foldable-traversable`. Since the changes in this branch are unrelated, I'm assuming it's known and can merge in a fix to this branch once it's on `master`.
* Looking at the [contribution guidelines][], it recommends adding a test to the examples. What test can I write there? It feels like the useful test would be that let bindings and where clauses gets parsed with the correct metadata. Should we eschew another test in the examples in favor of that? Or, is this a hard rule?
* Did I miss anything?

[contribution guidelines]: https://github.com/purescript/purescript/blob/master/CONTRIBUTING.md#pull-requests


LiamGoodacre:
@joneshf if you rebase against master the tests should be passing again now that this is merged: https://github.com/purescript/purescript/pull/3312

garyb:
Thanks @joneshf 😄

joneshf:
Thank you all!

PR #3318, opened 2018-04-24
Add positions for `ScopeConflict` caused by exports

garyb:
I'll see about improving the error spans before merging this, currently the whole-module span is used.


garyb:
There we go, much better. It's still not perfect, since for like the `ConflictingExports` example

``` purescript
module Main (module A, module B) where

import A
import B
```

It only has the span for

```
module Main (module A, module B) where
             ^^^^^^^^
```

rather than both of them, but that's a bit harder to figure out. It's still nice that it points to at least a relevant location :smile:

PR #3320, opened 2018-04-25
Allow operators shadowed by variables of the same name as the alias

### LINKED ISSUE #2803: Can't use an operator shadowed by a variable of the same name as the aliasee

kRITZCREEK:
This is a reimplementation of #2964.

Fixes #2803 

It uses a custom `ScopedIdent` type instead of `Qualified Ident`, as the qualifier was effectively used as a boolean, but allowed too many invalid states. Also makes use of the extra information in findUsage for the IDE code.

@LightAndLight If you could take a look that would be great, sorry for taking so long to respond to and review your PR properly. I'm totally happy if you want to work these changes into your PR and merging yours instead.

### DIFF
--- a/src/Language/PureScript/Linter.hs
+++ b/src/Language/PureScript/Linter.hs
@@ -28,8 +28,8 @@ import Language.PureScript.Types
 lint :: forall m. (MonadWriter MultipleErrors m) => Module -> m ()
 lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDeclaration ds
   where
-  moduleNames :: S.Set Ident
-  moduleNames = S.fromList (ordNub (mapMaybe getDeclIdent ds))
+  moduleNames :: S.Set ScopedIdent
+  moduleNames = S.fromList (map ToplevelIdent (mapMaybe getDeclIdent ds))
 
   getDeclIdent :: Declaration -> Maybe Ident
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/Linter.hs
```diff
@@ -28,8 +28,8 @@ import Language.PureScript.Types
 lint :: forall m. (MonadWriter MultipleErrors m) => Module -> m ()
 lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDeclaration ds
   where
-  moduleNames :: S.Set Ident
-  moduleNames = S.fromList (ordNub (mapMaybe getDeclIdent ds))
```
kRITZCREEK: This seemed superfluous? Inserting into the Set is going to get rid of any duplicates anyway.

garyb: Yeah, strange! Perhaps it was a List at one point.

kRITZCREEK:
Thanks a lot for all the work on this @LightAndLight!

PR #3258, opened 2018-02-26
Prevent codegen when `--dump-corefn` is supplied

gabejohnson:
Fixes #3196 

I opted to implement the suggestion from https://github.com/purescript/purescript/pull/3049#issuecomment-338507346; no JS is generated when `--dump-corefn` is passed to `compile`.

PR #3324, opened 2018-04-26
Introduce `--codegen` argument so targets can be specified explicitly

### LINKED ISSUE #3196: Add flag to disable Javascript backend
I'm writing an alternative VimL backend for PureScript based on the CoreFn dump produced by `--dump-corefn`. That is, I haven't forked the compiler codebase. There are two things with the regular `purs compile` command that would be nice to disable:

* CommonJS module output
* Checking/inclusion of `<Module>.js` files (foreign definitions)

Specifically the latter is of interest, as I don't want to write foreign definitions for Vim-specific things for the JS backend.

At a high level, this might be considered "disabling the entire JS backend", I'd say, so I'm not sure what would be a good name for a flag to disable it. Perhaps `--no-codegen` or `--no-js-codegen`. Anyway, I could have a stab at this and provide a PR if this seems reasonable.

Thanks!

garyb:
Resolves #3196, closes #3258 

Default is `js`, using `sourcemaps` implies `js` and the separate option for source maps is gone now.

Thanks to @gabejohnson for doing the first part of this!


garyb:
Any thoughts/comments about this @kRITZCREEK @LiamGoodacre @hdgarrood?

app/Command/Compile.hs
```diff
@@ -122,21 +126,45 @@ jsonErrors = Opts.switch $
...
+  , ("corefn", P.CoreFn)
+  ]
+
+targetsMessage :: String
+targetsMessage = "Accepted codegen targets are '" <> intercalate "', '" (M.keys targets) <> "'."
+
+targetParser :: Opts.ReadM [P.CodegenTarget]
+targetParser =
+  Opts.str >>= \s ->
+    for (T.split (== ',') s)
```
kRITZCREEK: Do we want to be a bit more lenient here and strip the text after splitting?

hdgarrood:
:+1: I like this a lot!

app/Command/Compile.hs
garyb: I'm not sure how you'd get spaces into that value anyway? Since `--codegen js, corefn` wouldn't parse as one thing.

garyb: Oh... quotes maybe?

LiamGoodacre:
Awesome.  No comments other than what @kRITZCREEK raised. :100:

app/Command/Compile.hs
```diff
@@ -154,6 +154,8 @@ targetParser =
       $ maybe (Opts.readerError targetsMessage) pure
       . flip M.lookup targets
       . T.unpack
+      . T.takeWhile (/= ' ')
```
kRITZCREEK: How about using http://hackage.haskell.org/package/text-1.2.3.0/docs/Data-Text.html#v:strip?

garyb: Thanks, couldn't find that when I looked through the docs before!

PR #3331, opened 2018-04-27
Check foreign kinds are exported when used in exported type

### LINKED ISSUE #3327: Exports for kind definitions used in exported types should be required
``` purescript
module Test (TestProxy(..)) where

foreign import kind Test

data TestProxy (p :: Test) = TestProxy
```

An export for `Test` should be required here too.

garyb:
Resolves #3327


garyb:
A wild Phil appears! :smile:

PR #3342, opened 2018-05-01
Added SourceSpan to CoreFn.Module

coot:
This adds `SourcePos` to `CoreFn.Module`.  This way one can run `codegen` from `MakeActions` when one only has `Language.PureScript.CoreFn.Module` rather than a `Language.PureScript.AST.Declaration.Module`.  This is useful for tools like [zephyr](https://github.com/coot/zephyr) to just run the `codegen` when the `CoreFn` was read from an output directory, so the `Module` is not available.

This change adds module `SourcePos` to `CoreFn` `JSON` representation, and also includes a test. I hope it can be included in `0.12.0`.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -50,7 +50,7 @@ moduleToJs
   => Module Ann
   -> Maybe AST
   -> m [AST]
-moduleToJs (Module coms mn _ imps exps foreigns decls) foreign_ =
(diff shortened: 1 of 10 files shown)


hdgarrood:
No objections from me. :+1:

coot:
I pushed one small commit, which removed `SourceSpan` argument of `codegen` member of `ModuleAction`, its not needed since it is present on `CoreFn.Module`.

coot:
Thank you all for a quick merge!

PR #3350, opened 2018-05-04
Explain --editor-mode intent

chexxor:
I was confused the intent of the --editor-mode flag in `purs ide server`. I found https://github.com/purescript/purescript/pull/3006, which explained the intent. I hope this little change helps future people.

### DIFF
--- a/psc-ide/README.md
+++ b/psc-ide/README.md
@@ -27,7 +27,7 @@ It supports the following options:
   files. This flag is reversed on Windows and polling is the default.
 - `--log-level`: Can be set to one of "all", "none", "debug" and "perf"
 - `--no-watch`: Disables the filewatcher
-- `--editor-mode`: Rather than watch source files, expect an editor to report
+- `--editor-mode`: Rather than watch externs files, expect an editor to report
   changed source files.
 - `--version`: Output psc-ide version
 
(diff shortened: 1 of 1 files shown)


nwolverson:
Unfortunately I don't think that's actually correct, in particular "Rather than watch source files" - the alternatives do not watch source files, they watch externs files. With `--editor-mode`, either the changes to externs files all come from rebuild commands, or there must be an explicit load of dependencies (almost "report changed source files" but `s/source/externs/`)

Anyway that's all a bit pedantic, I think you could change it to "Rather than watch externs files, expect an editor to report changed source files" without terribly much loss of information (@kRITZCREEK ?)

kRITZCREEK:
What @nwolverson said. I think we could add a bit of motivation here, saying that file watching has been flaky empirically and polling is resource intensive.

I was wondering if I should switch editor-mode on by default for 0.12? It definitely creates situations which are harder to debug for newcomers, since there's no way they understand that them running `pulp build` on the console just invalidated their IDE state.

chexxor:
@kRITZCREEK I didn't want to add too much text to that flag's explanation, as further explanation seems to belong in DESIGN document in this directory. I didn't make a change to that document because it's rather big/unorganized and I couldn't think of a good place in there to put it -- there's no config or options or flags section. Should we find a place for those sections? Or is the README the intended place for explaining the intents and implementation of the config/flags?

kRITZCREEK:
Thanks!

Issue #3302, opened 2018-04-12
Consider `deriving via`


garyb:
https://www.kosmikus.org/DerivingVia/deriving-via-paper.pdf

hdgarrood:
I think now that we have merged Coercible, we can move this into "Intend to implement"?

kl0tl:
This is something I’d like to try if nobody is already working on this.

PR #3351, opened 2018-05-04
Implement `Coercible` for safe zero-cost coercions

lunaris:
This PR adds `Coercible` for safe zero-cost coercions, as per the similar feature in GHC Haskell. It's intended to be the first step towards implementing `deriving via` (https://github.com/purescript/purescript/issues/3302). Compiler support only boils down to the constraint `Coercible` -- the actual function `coerce :: forall a b. Coercible a b => a -> b` can be implemented in library code (as `unsafeCoerce`, thus receiving the same inlining/elimination treatment and so being zero-cost). The tests will fail because of this point because I need to find a place to put `Safe.Coerce` (the module which exposes `coerce`). Happy to take input on how to best accomplish this.

## Changes

* Define a new built-in module, `Prim.Coerce` which exports a single
type class, `Coercible a b`, which relates types with identical runtime
representations.

* Extend the type checker's class solver to solve constraints of the
form `Coercible a b` automatically, as per the rules in the paper "Safe
Zero-Cost Coercions for Haskell" (though simpler due to the absence of
some of Haskell's features like type families).

* As part of the above point, add support for role inference to the type
checker.


src/Language/PureScript/Roles.hs
```diff
@@ -0,0 +1,97 @@
...
+                      walk ti
+                    Phantom ->
+                      []
+              in  concat (zipWith k t1Roles t2s)
+            -- If the type is an application of any other type-level term, walk
+            -- the first term to determine what role contributions it makes
+            -- (e.g. in the case `data F a b = F (a (F a b))`, the use of `a` as
+            -- a function/constructor would trigger this case)
+            _ ->
+              walk t1
```
kcsongor: It looks like this case should mark all  `t2s` as `representational`:

```
data F a b = F (a b)
```

As `a` is polymorphic, we don't know what its role annotation is going to be, therefore we should be careful and assume that its argument is `representational` (because it very well could be!). 

That means that for `F`, `b` would be inferred `representational` as well.
I _think_ the current implementation will infer `phantom` for `b`, allowing to coerce
`F Maybe Int` and `F Maybe String`

src/Language/PureScript/Roles.hs
```diff
@@ -0,0 +1,97 @@
...
+    walk t
+      | Just (t1, t2s) <- splitTypeApp t =
+          case t1 of
+            -- If the type is an application of a type constructor to some
+            -- arguments, recursively infer the roles of the type constructor's
+            -- arguments. For each (role, argument) pair, recurse if the
+            -- argument is representational (since its use of our parameters is
+            -- important) and terminate if the argument is phantom.
+            TypeConstructor t1Name ->
+              let t1Roles = inferRoles env t1Name
```
kcsongor: What happens with mutually recursive types?

```
data F a = F (G a)
data G a = G (F a)
```

src/Language/PureScript/Roles.hs
kcsongor: > It looks like this case should mark all `t2s` as `representational`:

Or rather, just `walk` all `t2s`, as that will mark all type variables `representational` anyway, and ensure that we don't accidentally mark phantom things representational (which would be safe, but undesirable).

So:
```
data G a b = G (a (Phantom b))

data Phantom a = Phantom
```
will not get tricked into marking `b` as representational.

-----------------------

Note: this case is actually interesting, because it differs from what GHC does. Not having to consider `nominal` roles results in a non-trivial difference:

For `G` above, GHC would infer
```
type role G representational nominal
```
This is because with `nominal` as a possibility, we must assume that `a`'s argument can be `nominal`, therefore `Phantom b` is in a `nominal` role. However, `nominal` is "infectious": if `Phantom b` is nominal, then all type variable therein must be nominal too (`b` here). This being the case, GHC rightly marks `b` as nominal, and stops. In the paper, this is accounted for by the rules that mark all free variables nominal.

However, with only `phantom` and `representational` roles, we can do better than simply downgrading `nominal` to `representational`! `Phantom b` being in a `representational role` doesn't "infect" `b`: walking is sufficient to determine the above result.

As a result, GHC would never allow to `coerce` between `G Maybe Int` and `G Maybe String` - but we can!

rightfold:
Without explicit role annotations, how can the programmer ensure safety when using phantom types that should prevent coercibility? There are many thinkable instances of this with the FFI.

kcsongor:
@rightfold you certainly raise a good point, I would imagine that by default all arguments to foreign types have to be representational

src/Language/PureScript/Roles.hs
```diff
@@ -0,0 +1,97 @@
...
+  | Just roles <- lookup tyName primRoles =
+      roles
+  | Just (_, DataType tvs ctors) <- M.lookup tyName envTypes =
+      -- For each constructor the type has, walk its list of field types and
+      -- accumulate a list of (formal parameter name, role) pairs. Then, walk
+      -- the list of defined parameters, ensuring both that every parameter
+      -- appears (with a default role of phantom) and that they appear in the
+      -- right order.
+      let ctorRoles = foldMap (foldMap walk . snd) ctors
+      in  map (\(tv, _) -> (tv, fromMaybe Phantom (lookup tv ctorRoles))) tvs
```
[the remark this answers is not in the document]
lunaris: I thought you were right but after some testing I think I actually get away with this due to the fact that `Phantom` parameters are not realised at all. E.g. for the types:

```haskell
data Phantom a = Phantom
data F a = MkF (Phantom a) a
```

`walk`ing the `MkF` constructor for `F` yields:

```haskell
[ [], [("a", Representational)] ]
```

That is, the `Phantom` use results in `[]`, not `("a", Phantom)`. So this works. I think this is OK but am open to alternative implementations.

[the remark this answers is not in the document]
kcsongor: Ah, you're right - I missed this. Perhaps then `walk` should just return a list of representational types? (as it already does, but its type is somewhat misleading)

lunaris:
I've implemented @kcsongor's suggestion to infer `Representational` for all `Foreign` declarations though will look into how cheaply I can implement role signatures (let the bikeshedding begin).

paf31:
I like the idea of using `Representational` as the default. I like the idea of keeping roles hidden from the user entirely.

lunaris:
As it currently stands that's what this branch does now. I've prototype role declarations in another branch but that's a separate discussion I think.

lunaris:
I've rebased this on `master`; can we make any progress on merging it? As noted, we'd need to decide:

* Module and package to expose `coerce` -- currently I've hacked a `Safe.Coerce` module into `purescript-unsafe-coerce` in the `tests/support/bower_components` directory to make the `tests/purs/passing/Coercible.purs` test pass. Implementation is just:

```haskell
module Safe.Coerce (coerce) where

import Unsafe.Coerce (unsafeCoerce)

coerce :: forall a b. Coercible a b => a -> b
coerce = unsafeCoerce
```

* If we are happy with the situation regarding roles and their inference etc., which it sounds like we might be. Not sure though.

vladciobanu:
Just a minor note, I think most of `Environment.hs` was recently (0.12) updated to have more descriptive names for type parameters. I think it might help newcomers if `a` and `b` were `from` and `to`, or something similar.

Other than that, this looks good to me and I hope it gets merged soon.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -358,6 +363,68 @@ entails SolverOptions{..} constraint context hints =
...
+      -- currently don't support higher-rank arguments in instance heads, term
+      -- equality is a sufficient notion of "the same".
+      if a' == b'
+        then pure [TypeClassDictionaryInScope [] 0 EmptyClassInstance [] C.Coercible [a, b] Nothing]
+        else do
+          -- When solving must reduce and recurse, it doesn't matter whether we
+          -- reduce the first or second argument -- if the constraint is
+          -- solvable, either path will yield the same outcome. Consequently we
+          -- just try the first argument first and the second argument second.
+          ws <- coercibleWanteds env a b <|> coercibleWanteds env b a
```
LiamGoodacre: Shouldn't these be `a'` and `b'`?

LiamGoodacre:
Just tried this situation and got no instance found via `gToGN`:
```purescript
-- existing tests
data G a b = G (a (Phantom2 b))
gToG :: G Maybe Int -> G Maybe String
gToG = coerce

-- new stuff
newtype OptionIn a = OptionIn (Maybe a)
newtype OptionOut a = OptionOut (Maybe a)
gToGN :: G OptionIn Int -> G OptionOut String
gToGN = coerce
```

LiamGoodacre:
Following from @kcsongor's earlier comment:
```purescript
data MutI f a = MutI a (Maybe (f a))
newtype MutA a = MutA (MutI MutB a)
newtype MutB t = MutB (MutI MutA t)

mutAToMutB :: MutA Int -> MutB Int
mutAToMutB = coerce
```

Gives the error:
```
          Type class instance for
            Prim.Coerce.Coercible (MutI MutA t1) (MutI MutB t0)
          is possibly infinite.

          Prim.Coerce.Coercible (MutI MutB t0) MutB
```

LiamGoodacre:
The following causes the compiler to loop and eat all your memory 😸 
```purescript
newtype X a = X (X (X a))
xToY :: X Int -> X (X Int)
xToY = coerce
```

LiamGoodacre:
Not sure if that previous one is a bug with this or not.  As the following also loops:
```purescript
newtype X a = X (X (X a))
xToY :: X Int -> X (X Int)
xToY = ?ohNo
```
Will make a new issue for this.

paf31:
> Without explicit role annotations, how can the programmer ensure safety when using phantom types that should prevent coercibility? There are many thinkable instances of this with the FFI.

I should note that (since writing my comment above) I've come to understand that this is actually really important, and I was really missing the point.

lunaris:
Based on @paf31's latest comment, and the fact that I have some time, I've updated this branch to build against `master` and merged in my other branch that adds explicit role support to the compiler.

## What this means now

* In the absence of things like type families, there are two roles, `Representational` and `Phantom`. Representational parameters affect the runtime representation of the types they parameterise and phantom parameters do not.

* Roles are now inferred for all types but can also be declared explicitly, using the following syntax (up for grabs):

```
data Phantom a = Phantom

role Phantom representational
```

In the above, `Phantom`'s role is defined explicitly -- were it inferred, it'd end up being `Phantom` and not `Representational` as we explicitly defined. This means we can't write something like:

```
castPhantom :: forall a b. Phantom a -> Phantom b
castPhantom = coerce
```

Even though this is technically possible, the role definition means the compiler will reject it. This can be used with `foreign` types as mentioned earlier in this PR. Additionally there's now a failing test to witness that this sort of thing should be forbidden (see `CoercibleRepresentational.purs` in `failing`).

* Have fixed the documentation by adding instance names as per @LiamGoodacre's suggestion, but still need to investigate the other bugs/edge cases. Now that this builds on `master`, all help welcome to get this over the line!

_Note_: to test this you'll probably still have to hack up some module for `Safe.Coerce` -- as mentioned earlier in this thread, I typically just add `Safe/Coerce.purs` to `purescript-unsafe-coerce` in `test/support/bower_components` with the following content:

```
module Safe.Coerce (coerce) where

import Unsafe.Coerce (unsafeCoerce)

coerce :: forall a b. Coercible a b => a -> b
coerce = unsafeCoerce
```

hdgarrood:
Some thoughts:
- If you haven't already, can you add tests for all of the things @LiamGoodacre brought up?
- At this stage,  reserving `role` is the kind of breaking change which I think we'd need to spend a while issuing warnings for before we actually do it. For example, we might issue warnings for the entirety of 0.13.x, and then actually break it in 0.14.0. Alternatively we could bikeshed about the syntax to see if we can find a non-breaking way of adding role annotations.
- I think we should make a decision about type families before merging this, by which I mean either saying we definitely want to add them at some point, or committing to not adding them. I think I'd prefer the latter; unless I'm missing something obvious, I think type classes with functional dependencies allow you to do a lot of what type families do, and it's also very useful to know that type constructors are injective (and indeed there are parts of the ecosystem which depend on this, such as [leibniz](https://pursuit.purescript.org/packages/purescript-leibniz/5.0.0/docs/Data.Leibniz#v:lowerLeibniz)).

Since these issues are quite a bit wider in scope than just the code in this PR, I think Discourse might be a better place to discuss them. Thoughts?

hdgarrood:
Come to think of it, what I said about type families probably applies to GADTs too, right?

src/Language/PureScript/Roles.hs
```diff
@@ -0,0 +1,125 @@
...
+      -- representational.
+      [(v, Representational)]
+    walk (ForAll _ t _) =
+      -- We can walk under universal quantifiers. Note that this might appear
+      -- erroneous -- given a definition like `data D a = D (forall r. r -> a)`
+      -- we expect we'll produce `[("r", Representational)]` as part of the
+      -- recursive call's result. This is true, but fine, since this tuple will
+      -- never be looked up when building the final result, which only looks at
+      -- the variables defined as parameters to the type.
+      walk t
```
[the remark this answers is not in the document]
lunaris: I think I've addressed this now -- when we walk under a quantifier, we track the variable it binds and don't return roles for such variables.

src/Language/PureScript/TypeChecker/Entailment.hs
lunaris: I think you are probably right. I've fixed this.

lunaris:
I don't have a strong preference on type families, though I'm not sure why we'd have to commit firmly to having/not having them at this point -- if we decide to add them then this stuff will need reworking for sure, but I don't think that's a big deal (well, no bigger a deal than having to add type families in the first place). I understand that coercion and type families are often thought of together, but I think the points about roles and foreign types here hopefully illustrate that they can be considered/seen as useful orthogonally.

I've addressed a couple more of @LiamGoodacre's issues in the latest commit -- I think role inference now works correctly for constructors with higher-ranked fields, for instance. But there are probably some more bits to iron out.

It's definitely a breaking change with the syntax as it stands, yep, but as I said I'm not particularly precious on what the syntax is.

hdgarrood:
The thing I'm worried about is that (at least, judging by a quick skim of the contents of https://ghc.haskell.org/trac/ghc/wiki/Roles) it's not at all easy to get type families, GADTs, and roles all working nicely together, so I'm not yet convinced that they can be considered orthogonally. I want to be much more conservative than Haskell is with respect to adding type-level language features, because personally it's really important to me that PureScript's type-level language features work well together (even if that means not having as many), and I worry that the reworking you mention which would be required if we were to decide we wanted to add type families could turn out to be really quite difficult.

natefaubion:
At the risk of potentially getting slightly off topic, I'll say that I really want to eliminate the usage of `unsafeCoerce` in the wild, which is used most often for zero-cost coercions and encoding existentials, neither of which require type families or GADTs. I would be very surprised if PS ever got either.

garyb:
Although... GADTs would mean we wouldn't need `unsafeCoerce` for existentials, since they imply existential types.

natefaubion:
Yeah, but you can implement existentials in a fairly straightforward manner without GADTs.

lunaris:
I guess my point is -- there is an orthogonality to me because "do we want zero-cost coercions" and "do we want type families" are two separate questions to which the answer "yes" is independently useful. It's true that saying yes to both then results in an interplay which has to be dealt with, but I don't know how to debate that bit -- that just feels like a decision relating to the risk of future development, maybe.

hdgarrood:
Sure, totally. I'm not necessarily asking you to help debate that bit (unless you want to), I'm just saying that it's something I'd prefer to address first.

hdgarrood:
Having read the [safe zero-cost coercions paper](https://repository.brynmawr.edu/cgi/viewcontent.cgi?referer=&httpsredir=1&article=1010&context=compsci_pubs) I think I may have been looking at out-of-date information and overestimating the difficulty of having type families and/or GADTs together with Coercible. The approach taken in that paper seems to be to give all parameters to GADTs and type families the nominal role, which prevents any dodgy coercions (at the cost of disallowing some coercions which would be valid), and presumably this would actually be pretty straightforward to implement if we did later introduce GADTs and/or type families. (although as aside I would probably vote against adding them anyway.) So I'll withdraw my previous objection about wanting to make a decision about type families and GADTs first.

rhendric:
Reading the paper linked in the comment immediately above, it says that nominal roles are used to prevent coercing, for example, between `Map Age v` and `Map Int v`. I don't see nominal roles anywhere in this PR; does this mean that this `coerce` isn't actually safe? Or am I missing something that prevents this from being a problem?

lunaris:
This is a good point -- it's unsafe in the sense that `coerce` cannot be made aware of constructors or types that have invariants invisible to the type system (such as `Set`, `Map` etc.).

It's quite easy to solve I think -- if we add `nominal` as an explicit role with appropriate rules then that should be it. There's currently no situation in which `nominal` could be inferred (though we _could_ change `foreign` types to do this instead of `representational`) and the rest should work as-is.

hdgarrood:
Yes, I think we should have `nominal` as a role for cases like Map and HashMap. I don’t think it matters that there is currently no situation in which `nominal` would be inferred. I think inferring `representational` for `foreign import data` types is probably okay but if anyone has any more thoughts on the trade offs there please do comment :)

Regarding syntax I’d be happy with

```purescript
data Map k v = ...

data role Map nominal representational
```

which ought to be possible in a non breaking way, i.e. we wouldn’t need to reserve `role` as a keyword, it would only have special significance immediately after `data`. I think this was originally @natefaubion’s suggestion. The alternative is to somehow annotate roles inside the original `data` declaration but I can’t think of a way of doing that without clashing awkwardly with kind annotations.

@lunaris would you mind waiting until the cst work is in before doing anything with the syntax? I’m hoping we can get that merged in the next week or so. Thank you for being so patient with us!

Nobody seems to mind about us saying we don’t want type families (see [the relevant discourse thread](https://discourse.purescript.org/t/thoughts-on-future-additions-of-type-level-features/707)) so I don’t have any other reservations about adding this now.

natefaubion:
If it were to infer `representational` for `Map` (no specific role annotation), does that mean the function is no longer safe in general?

hdgarrood:
Regarding the location of `coerce`: I think the options are to put it in Prelude or to put it in a core library `purescript-safe-coerce`. I’m leaning towards the latter since we probably won’t expect `coerce` to be explicitly used a massive amount. Putting it in `unsafe-coerce` would be a bit strange I think.

hdgarrood:
It is a bit less safe, yeah, since we rely on library authors to annotate nominal type parameters. This doesn’t bother me too much, personally.

natefaubion:
So I guess you could say it doesn't violate _type_ safety, but you do need to opt-in if you have internal invariants that need to be maintained.

timjs:
Regarding syntax

> ```purescript
> data Map k v = ...
> 
> data role Map nominal representational
> ```

what about

```purescript
data Map k v = ...
  where
    role k nominal
    role v representational
```

hdgarrood:
Yeah, I like that.

hdgarrood:
I would like to get this landed during the 0.13.x series. @lunaris: thanks very much for your patience. Would you like to have a go at rebasing and implementing the syntax suggested by @timjs? If you'd struggle to find the time I'd be happy to take a stab at it myself too.

lunaris:
I would ask if we are sure on the syntax. Supposing we have:

```
data Map k v = Map ()
  where
    role k nominal
    role v nominal
```

I can see how this relates to `newtype`s, but do we also have:

```
foreign import data Foo :: Type -> Type -> Type
  where
    role ? nominal
    role ? representational
```

in which the parameters being described have no names? Additionally, I accept that GADTs may be something we never intend to implement, but if for any reason we do, would we end up with:

```
data Maybe :: Type -> Type where
  Nothing :: forall a. Maybe a
  Just    :: forall a. a -> Maybe a
    where
      role ? nominal
```

(here we have two challenges -- no type parameter names and a double `where`)

I appreciate a bikeshedding discussion is the last thing this PR needs at this point, now that it is 13 months old, but I think the current syntax is a. considerably simpler to implement and b. solves/avoids all of these issues.

hdgarrood:
Yes, that’s a very good point. My main objection to the current syntax is that it requires reserving “role” as a keyword, so in that case my vote would be for
```
data Map k v = ...
type role Map nominal representational
```
so that “role” only has special meaning directly after “type”. I realise I suggested using “data” instead of “type” before but I think “type” reads a little more naturally and also is the same as Haskell (which means fewer differences to remember if you’re switching between the two regularly).

lunaris:
OK, I think I've sorted this. Notes:

* CST now supports role declarations (from the syntax `type role`). Making `role` usable as an identifier in cases where it is not preceded by the keyword `type` has increased the number of shift-reduce conflicts. I think this is OK and have changed the expectation accordingly. I've also added a couple of tests that ensure one can use `role` in this manner.
* The `nominal` role can now be used. It will never be inferred (`representational` is still the default for `foreign`) but one can use a role declaration to introduce it. When reducing `Coercible` constraints, `nominal` will introduce _a strict equality comparison_. Ideally, we'd emit an equality constraint and have the solver reduce that (perhaps unifying as-yet unbound type variables), but in the absence of such constraints this will have to do. I'm pretty sure this means there will be cases where a `coerce` is technically legal but the type checker won't be able to see it, but it is at least sound. That said, with my very limited testing, I couldn't construct a case, so I'm minded to see how that pans out. I've added a comment to this effect in any case.
* Tests added for `nominal` roles -- both passing and failing.

Assuming all this looks good, all that remains is to define the `Safe.Coerce` module in an appropriate library and have the tests etc. use that (as mentioned, currently it's just being hacked into an arbitrary package in the test support directory). Until we do, the automated checks in Github won't pass.

lunaris:
I am of course talking nonsense -- the `inferRoles` function will have to be updated to recurse/adapt when inferring the roles of types that wrap explicitly-denoted-as-`nominal` types. I will sort this and update when I can.

src/Language/PureScript/CST/Parser.y
```diff
@@ -188,12 +193,14 @@ qualIdent :: { QualifiedName Ident }
...
   | 'hiding' {% toQualifiedName Ident $1 }
   | 'kind' {% toQualifiedName Ident $1 }
+  | 'role' {% toQualifiedName Ident $1 }
 
 ident :: { Name Ident }
   : LOWER {% toName Ident $1 }
   | 'as' {% toName Ident $1 }
   | 'hiding' {% toName Ident $1 }
   | 'kind' {% toName Ident $1 }
+  | 'role' {% toName Ident $1 }
```
natefaubion: You need to add the other contextual keywords to `ident` and `qualIdent` as well (phantom, nominal, representational).

src/Language/PureScript/TypeChecker/Roles.hs
```diff
@@ -0,0 +1,137 @@
...
+  = zipWith mkPair [0..]
+  where
+    mkPair :: Int -> Role -> (Text, Role)
+    mkPair i r =
+      (pack ("a" ++ show i), r)
+
+-- |
+-- Given the kind of a foreign type, generate a list of formal parameter names
+-- each tied to a `Representational` role which, in the absence of role
+-- signatures, provides the safest role signature which can be assigned to a
```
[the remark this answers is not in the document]
lunaris: The question now, I suppose, is -- is it? Perhaps `nominal` is the sensible default here, because a `foreign` type really could do anything.

lunaris:
OK I think I've fixed this now. One test case that previously passed now fails; I think this is correct. The case in question:

```haskell
data Phantom b = Phantom
data G a b = G (a (Phantom b))
```

In the absence of `nominal`, it would be possible to infer `G`'s roles as `a = representational`, `b = phantom` and thus do things like `coerce :: G Maybe Int -> G Maybe String`. However, in the presence of `nominal`, this is not sound, because one can introduce something like:

```haskell
data Map k v = Map ...
type role Map nominal representational
```

And now `coerce :: G Map Int -> G Map String` is unsafe, because `Map` could react differently to `Phantom String` than it does to `Phantom Int`. Thus, as directed by the original paper, when dealing with type applications (such as `a (Phantom b)` in `G`'s definition, we mark all free variables in the argument (`(Phantom b)`, so `b`) as `nominal`. This disallows the case we previously supported but is sound.

src/Language/PureScript/TypeChecker/Roles.hs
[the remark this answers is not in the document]
garyb: I'd agree that `nominal` seems like the safer choice in the absence of other info.

lunaris:
OK I've made `nominal` the default for `foreign` types now and added tests to make sure that one can still ascribe e.g. `representational` roles to `foreign` types if they know it's safe. Additionally I've added a few cases to `walk` to stop it being broken (I'd completely missed out row types) and some tests to that effect. I think this is ready for a few people to play with it now -- let me know what you think.

rhendric:
> One test case that previously passed now fails; I think this is correct.

As a mere spectator, I'm relieved you made this change (defaulting to `nominal` instead of `representational` in the absence of information). I remember being slightly concerned that defaulting to `representational` would pose problems like the one you brought up, but I thought it likely that I just didn't understand roles well enough. I may still not understand them well enough for this feedback to be worth anything, but for whatever it is or isn't worth, I think you're making the right choice.

timjs:
> ```purescript
> foreign import data Foo :: Type -> Type -> Type
>   where
>     role ? nominal
>     role ? representational
> ```

Very good, point, didn't think about that one!

LiamGoodacre:
Haven't read this paper yet but could be a another good resource for inspiration: https://arxiv.org/pdf/1905.13706.pdf

kcsongor:
Any update on this? I would really like to use this feature in my applications!

hdgarrood:
This is happening, but please don't rush us.

@lunaris I haven't yet managed to play with this properly, but I have just created a [`purescript-safe-coerce`](https://github.com/purescript/purescript-safe-coerce/) repo.

hdgarrood:
@lunaris are you happy for me to push a few commits to fix the conflicts and try to get CI passing?

lunaris:
Sounds good!

hdgarrood:
Since our next release is going to be 0.14.0 I would suggest merging this now. In particular I think it ought to be merged before the polykinds PR, since it has been in the queue for quite a bit longer. Any objections to that? cc @natefaubion

natefaubion:
I have no problem with merging this first.

garyb:
0.14 is going to be so :tada:

hdgarrood:
Nice! Thanks so much @lunaris, and sorry it took us so long to get this merged.

Issue #3353, opened 2018-05-06
`IncorrectConstructorArity` does not communicate how many arguments were given and expected


joneshf:
The error message for `IncorrectConstructorArity` doesn't tell you what the arities are:

https://github.com/purescript/purescript/blob/3d62184cd65402d1a23ac6c26bef6ce97c917d48/src/Language/PureScript/Errors.hs#L768-L769

It just tells you that it knows what you did wrong without telling you how to fix it. You have to then go look up the arity on your own and fix the problem. It would be more helpful if the error message told you what the arity is. You could fix the problem much more easily.

We have the information:

https://github.com/purescript/purescript/blob/345f6435f2aa23a9403d25a0486596b8b0b4399b/src/Language/PureScript/TypeChecker/Types.hs#L480

We just need to communicate it.

How do we feel about this message:

> Data constructor Foo was given n arguments in a case expression, but expected m arguments. This problem can be solved by giving Foo m arguments.

?

garyb:
Yes please!

PR #3354, opened 2018-05-06
Add arities to `IncorrectConstructorArity`

joneshf:
Re: #3353 

Let me know if anything should change.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -479,7 +479,7 @@ inferBinder val (ConstructorBinder ss ctor binders) = do
       let (args, ret) = peelArgs fn'
           expected = length args
           actual = length binders
-      unless (expected == actual) . throwError . errorMessage $ IncorrectConstructorArity ctor expected actual
+      unless (expected == actual) . throwError . errorMessage' ss $ IncorrectConstructorArity ctor expected actual
       unifyTypes ret val
       M.unions <$> zipWithM inferBinder (reverse args) binders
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -477,7 +477,9 @@ inferBinder val (ConstructorBinder ss ctor binders) = do
       (_, fn) <- instantiatePolyTypeWithUnknowns (internalError "Data constructor types cannot contain constraints") ty
       fn' <- introduceSkolemScope <=< replaceAllTypeSynonyms $ fn
       let (args, ret) = peelArgs fn'
-      unless (length args == length binders) . throwError . errorMessage $ IncorrectConstructorArity ctor
+          expected = length args
```
joneshf: Are these two the right way around? I think they are, but I'm not 100% sure.

garyb: Yeah I'm pretty sure this is right, since `expected` comes from the environment (a few steps removed), whereas `binders` is from the case match.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -477,7 +477,9 @@ inferBinder val (ConstructorBinder ss ctor binders) = do
...
+          actual = length binders
+      unless (expected == actual) . throwError . errorMessage $ IncorrectConstructorArity ctor expected actual
```
garyb: While we're in here, can we change this to `errorMessage' ss`? Should give slightly better error positioning.

kRITZCREEK:
Works nicely 👍

joneshf:
Thanks!

PR #3355, opened 2018-05-07
[purs ide] Removes Pursuit commands

kRITZCREEK:
This code hasn't been used for a long time, and is not pulling its weight. It's just easier to use the website or make the HTTP calls from the editor plugin itself.

I also updated to LTS 11 in this PR, which only required relaxing a bound on ansi-terminal and hiding a few more things from Protolude... (getting more and more convinced that we should have a `Language.PureScript.Prelude`)


kRITZCREEK:
Thanks for the review!

PR #3356, opened 2018-05-08
delete psc-bundle/README.md

Cmdv:
update`psc-bundle` readme to use `purs bundle`


hdgarrood:
Since all of this information is available from `purs bundle --help`, I'm tempted to actually just delete this file.

Cmdv:
I was thinking about that and also the folder, I can change PR to do that if you like?

hdgarrood:
Yes please! You only need to tell git to delete the file; once the directory is empty git no longer considers it to be part of the repo.

PR #3361, opened 2018-05-14
Fix purs bundle to recognise explicit paths

### LINKED ISSUE #3359: `purs bundle` doesn't bundle anything
I think this is perhaps a knock-on effect from the change to include `index.js` in require paths.

garyb:
Resolves #3359 

Could do with a test for bundling so this doesn't happen again, but I'm not sure how best to achieve that really.

@chexxor sorry, I know you said you were going to take a look, but I knew how to do this pretty easily so thought I'd just get it in :smile:

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -145,12 +145,19 @@ printErrorMessage (MissingMainModule mName) =
 
 -- | Calculate the ModuleIdentifier which a require(...) statement imports.
 checkImportPath :: String -> ModuleIdentifier -> S.Set String -> Either String ModuleIdentifier
-checkImportPath "./foreign" m _ =
+checkImportPath "./foreign.js" m _ =
   Right (ModuleIdentifier (moduleName m) Foreign)
 checkImportPath name _ names
-  | Just name' <- stripPrefix "../" name
+  | Just name' <- stripSuffix "/index.js" =<< stripPrefix "../" name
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Bundle.hs
```diff
@@ -145,13 +145,20 @@ printErrorMessage (MissingMainModule mName) =
...
-checkImportPath "./foreign" m _ =
+checkImportPath "./foreign.js" m _ =
   Right (ModuleIdentifier (moduleName m) Foreign)
 checkImportPath name _ names
-  | Just name' <- stripPrefix "../" name
+  | Just name' <- stripSuffix "/index.js" =<< stripPrefix "../" name
   , name' `S.member` names = Right (ModuleIdentifier name' Regular)
 checkImportPath name _ _ = Left name
 
+stripSuffix :: Eq a => [a] -> [a] -> Maybe [a]
```
kRITZCREEK: Could this just be: https://hackage.haskell.org/package/filepath-1.4.2/docs/System-FilePath-Posix.html#v:takeDirectory?

kRITZCREEK: I guess it doesn't do the `Maybe` thing...

garyb: Perhaps, yeah - but this way at least ensures it's exactly the expected pattern, whereas that might be too broad, unless we check the filename part separately too.

garyb:
Ah sorry, thanks for looking into it and working on it anyway!

PR #3363, opened 2018-05-17
[purs ide] Use absolute paths for source locations

kRITZCREEK:
This caused a few bugs in the Emacs plugin, and I think it's less error-prone to just make sure we emit absolute paths.

I'd like to get this into 0.12


kRITZCREEK:
Thanks :)

PR #3365, opened 2018-05-21
Bump to 0.12.0

garyb:
Any objections to making this release as things are?

Core libraries are almost ready too, will be working to get them releasable tomorrow.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -1,5 +1,5 @@
 name: purescript
-version: '0.12.0-rc1'
+version: '0.12.0'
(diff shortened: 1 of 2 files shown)


hdgarrood:
Can we call it 0.12.1 so that people don’t confuse it with the rc (which prints “0.12.0” if you run `purs —version`?)

garyb:
Ack, I missed that while putting together the updated release notes... it's done here and on Hackage as 0.12.0 now.

PR #3372, opened 2018-05-31
Changes necessary to build on GHC 8.4.3

kRITZCREEK:
This isn't ready just yet, because we need to depend on a GIT version of `spdx` right now. The master for that package already builds, it just hasn't been released to Hackage yet.


kRITZCREEK:
Reminder, ping @cnd when we merge this

kRITZCREEK:
Looks like we're running into https://github.com/haskell/network/issues/313 / https://github.com/commercialhaskell/stack/issues/3944 for our AppVeyor build.

Presumably that'll be fixed with network 2.7.0.1 but we'll have to old off until that's released and we tested that it fixes the Windows issues.

ilovezfs:
@kRITZCREEK https://hackage.haskell.org/package/network-2.7.0.1 was released

angerman:
What's this blocked on? As far as I can tell from travis the jobs just timed out?

kRITZCREEK:
@angerman We need a release of spdx here: https://github.com/phadej/spdx/issues/12 but the author is unresponsive so we might need to consider forking and maintaining it ourselves.

@hdgarrood I think you introduced `spdx` to verify a proper LICENSE is set before updating documentation to Pursuit? Can we simplify that check and vendor it or should we fork the package instead?

angerman:
@phadej any advice on how to handle that?

hdgarrood:
I don’t think it’s particularly complex, it’s just not the kind of thing I’d want to vendor if i could avoid to because it could be quite annoying for users if we get it wrong. Although I suppose we might not have much choice.

I don’t have any strong preference for forking vs vendoring except that I suppose the former might imply more of a maintenance commitment which perhaps makes it less desirable. Do you mind if I leave it up to you?

hdgarrood:
I suppose we could also simplify the check to “is the license field present and not equal to AllRightsReserved” or something similar, but I think it’s important to at least have something in place which helps prevent people uploading things to Pursuit without a genuine open source license.

angerman:
Could you use `lib:Cabal` for this?

http://hackage.haskell.org/package/Cabal-2.2.0.1/docs/Distribution-SPDX.html

/cc @phadej, @hvr

kRITZCREEK:
> I suppose we could also simplify the check to “is the license field present and not equal to AllRightsReserved” or something similar, but I think it’s important to at least have something in place which helps prevent people uploading things to Pursuit without a genuine open source license.

So this is purely a client-side check? Is the check just meant to discourage people from uploading non-open-source licensed library or strictly prevent them? You can just upload things to Pursuit via HTTP so I don't think it's sufficient if we want the latter.

> Do you mind if I leave it up to you?

I don't have the capacities to maintain a fork, nor do I know my way around licenses to make a good educated decision, sorry... I'm just going to delete the SPDX dependent code in this PR, and if you can come up with a better solution you can just commit to my branch (collaborators have push access)

phadej:
Why purescript is using `spdx`, is there something not in Cabal-2.2 http://hackage.haskell.org/package/Cabal-2.2.0.1/docs/Distribution-SPDX.html (note `Parsec` and `Pretty` instances).

I'd not recommend to use spdx as is any longer, next version will reuse `Cabal` definitions anyway.

phadej:
As for why it's not released: I haven't updated spdx license list in `spdx` since 2017. `spdx` uses has 2.6,  i`Cabal-2.2` has 3.0, and latest one is Version 3.2 2018-07-10; (there was 3.1 in between)

phadej:
@hdgarrood I'm ore committed to updating `Cabal` than `spdx`. The plan to `spdx` is to have "legal" stuff, e.g. it have `satisfies` and `equivalent` functions to compare SPDX expressions - but those are out of scope of Cabal for now

kRITZCREEK:
We're currently having trouble with our CI builds timing out and work towards reducing the compilation times (eg. see #3400). So pulling in a dependency on the `Cabal` library for a check that should really live in Pursuit anyway doesn't excite me.

phadej:
`Cabal-2.2` comes with GHC-8.4, you don't need to install it.

kRITZCREEK:
Do I also not need to compile and link it?

phadej:
You obviously need to link it, but as it comes with GHC you don't need to compile it (the objects files are there already); In the some way you don't need to compile `containers` or `deepseq`, for example.

kRITZCREEK:
That's good to know, thanks!

kRITZCREEK:
Allright, so I've changed the code to use `Distribution.SPDX` and added a dependency on `Cabal`. I'll open another issue about it because I think ideally we'd just not check these licenses inside the compiler, but for now I just want to get this merged so we don't lag behind too far.

Does this work for you @hdgarrood?

src/Language/PureScript/Publish.hs
```diff
@@ -230,7 +230,7 @@ checkLicense pkgMeta =
 -- Check if a string is a valid SPDX license expression.
 --
 isValidSPDX :: String -> Bool
-isValidSPDX = (== 1) . length . SPDX.parseExpression
+isValidSPDX = isJust . SPDX.mkLicenseId
```
hdgarrood: As I understand it, this will attempt to parse a license ID rather than a license expression, which means that e.g. "GPL-2.0+", meaning GPL 2.0 or any later version, would not be accepted. From looking at the docs, I think we need to use the `Parsec` instance for `License`, and then check that a) the parse succeeds and b) that the result is not `NONE`.

kRITZCREEK: I see, thanks for checking. How about now?

hdgarrood: Looks good now, thanks!

kRITZCREEK:
@LiamGoodacre or @garyb could I get a review on this?

kRITZCREEK:
Thanks!

Issue #3245, opened 2018-02-14
Qualified rebindable syntax


pkamenarsky:
Currently, it's not possible to use `bind`s from multiple modules without explicitly rebinding `bind` in every do block, i.e.

```purescript
import Prelude (bind, pure)
import O as O

bindPrelude = do
  a <- [1, 2, 3]
  pure a

bindO = do
  a <- O.elems O.users
  O.pure a
  where
    bind = O.bind
```

I am proposing the following syntax for the latter use case:

```purescript
bindO = O.do
  a <- O.elems O.users
  O.pure a
```

which would desugar into `O.bind` calls instead of just using the `bind` which is currently in scope. Also, this could potentially be extended to include other keywords like `if` (if https://github.com/purescript/purescript/issues/3164 gets implemented), for example:

```purescript
O.if a
  then b
  else c
```

hdgarrood:
I’m not keen on this, personally. The way it works now is easier to explain and simpler implementation-wise. Also, I think writing `where bind = I.bind` is not particularly onerous, so adding this only seems to me to make a fairly minor difference; for me, it’s not really enough to justify new syntax. Especially because I think it’s quite rare to want to use more than one `bind` in a given module.

pkamenarsky:
It'd be pretty nice to being able to mix-and-match different `binds` forming different DSLs in a single function, i.e:

```
foo :: String
foo = Monoid.do
  "hello"
  show do
    [unit, unit]
    a <- [1, 2, 3]
    b <- [4, 5, 6]
    pure (a + b)
  "world"
```

(example posted by @natefaubion in the Slack channel).

What are the downsides to adding additional syntax? Seems like a pretty trivial change to me.

garyb:
> I’m not keen on this, personally. <snip>

Interesting, I actually much prefer this to what we have! It perhaps looks a little strange to qualify a keyword until you know what it does, sure, but I would find this much more convenient to use when using indexed monads and the like (and especially when mixing indexed and non-indexed).

hdgarrood:
I think any syntax change is a big deal, because it can affect a really large proportion of the ecosystem — potentially lots of the compiler as well as tooling — and because it’s hard to fully predict the knock-on effects it can have. To give an example, this change would break a source code formatter, for instance. Just because the initial implementation is easy to do that doesn’t mean it won’t have a higher long term maintenance cost.

[...]

PR #3373, opened 2018-06-02
Implement QualifiedDo

pkamenarsky:
Implements https://github.com/purescript/purescript/issues/3245.

Now, the following is possible:

`IxMonad.purs`:

```
module IxMonad where

class IxMonad m where
  pure ∷ forall a x. a -> m x x a
  bind ∷ forall a b x y z. m x y a -> (a -> m y z b) -> m x z b
```
`Main.purs`:

```
import IxMonad as I

test :: forall m a. I.IxMonad m => m a a String
test = I.do
  a <- I.pure "test"
  b <- I.pure "test"
  I.pure (a <> b)
```

instead of having to rebind `bind` explicitly.

There's no change in behaviour if `do` is left unqualified:

```
test :: forall m. Monad m => m String
test = do
  a <- pure "test"
  b <- pure "test"
  pure b
```

If there's interest I'll add support for `ado` as well.

### DIFF
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -483,13 +483,13 @@ parseAccessor obj = P.try $ Accessor <$> (indented *> dot *> indented *> parseLa
 
 parseDo :: TokenParser Expr
 parseDo = do
-  m <- P.try (parseQualified (reserved "do") >>= pure . getQual) <|> (reserved "do" *> pure Nothing)
+  m <- P.try (getQual <$> parseQualified (reserved "do")) <|> (reserved "do" *> pure Nothing)
   indented
   Do m <$> mark (P.many1 (same *> mark parseDoNotationElement))
 
 parseAdo :: TokenParser Expr
 parseAdo = do
-  m <- P.try (parseQualified (reserved "ado") >>= pure . getQual) <|> (reserved "ado" *> pure Nothing)
+  m <- P.try (getQual <$> parseQualified (reserved "ado")) <|> (reserved "ado" *> pure Nothing)
   indented
   elements <- mark (P.many (same *> mark parseDoNotationElement))
   yield <- mark (reserved "in" *> parseValue)
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -449,7 +449,7 @@ parseValueAtom = withSourceSpan PositionedValue $ P.choice
                  , parseCase
                  , parseIfThenElse
                  , P.try parseDo
-                 , parseAdo
+                 , P.try parseAdo
                  , parseLet
                  , P.try $ Parens <$> parens parseValue
                  , withSourceSpan' Op $ parseQualified (parens parseOperator)
@@ -489,11 +489,11 @@ parseDo = do
 
 parseAdo :: TokenParser Expr
 parseAdo = do
-  reserved "ado"
+  m <- parseQualified (reserved "ado") >>= \(Qualified m _) -> pure m
   indented
   elements <- mark (P.many (same *> mark parseDoNotationElement))
   yield <- mark (reserved "in" *> parseValue)
-  pure $ Ado elements yield
+  pure $ Ado m elements yield
 
 parseDoNotationLet :: TokenParser DoNotationElement
 parseDoNotationLet = DoNotationLet <$> (reserved "let" *> indented *> mark (P.many1 (same *> parseLocalDeclaration)))
--- a/src/Language/PureScript/Parser/Declarations.hs
+++ b/src/Language/PureScript/Parser/Declarations.hs
@@ -448,8 +448,8 @@ parseValueAtom = withSourceSpan PositionedValue $ P.choice
                  , P.try parseVar
                  , parseCase
                  , parseIfThenElse
-                 , P.try parseDo
-                 , P.try parseAdo
+                 , parseDo
+                 , parseAdo
                  , parseLet
                  , P.try $ Parens <$> parens parseValue
                  , withSourceSpan' Op $ parseQualified (parens parseOperator)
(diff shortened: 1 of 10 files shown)


pkamenarsky:
(Tests fail because `tests/support/bower.json` was referencing stale `0.12` branches. This has been fixed in the meanwhile, but not sure how to restart CI.)

natefaubion:
I personally like this feature, as I think it makes rebindable-do _actually useful_. I don't know of anyone using the feature in earnest because the ergonomics are so terrible, so _at best_ I consider rebindable-do a neutral side-effect of not having a hard dependency on Prelude. But I know the reception in the ticket was mixed. I think we should make a decision either way if this is something that we want to merge (now or eventually).

hdgarrood:
@natefaubion: Yes, I agree. I'm a little hesitant for the same reasons I gave in #3245, but I'll admit that I don't remember finding myself in a position where I was wanting to use rebindable syntax like this, and in fact I always did view rebindable syntax as a neutral consequence of separating Prelude from the compiler, rather than a feature that I expected to get much use. But looking back at the ticket, I think the ergonomics argument makes sense. You and @garyb being in favour is sufficient for me to say sure, let's do it.

@pkamenarsky this is perhaps a bit late now, but you can trigger a new CI build by pushing an empty commit, or by rebasing your PR on master.

hdgarrood:
Additionally, if we are doing this, I think we should support `ado` as well.

pkamenarsky:
`do` works now, will try to finish `ado` in the next couple of days in order to get this ready for review.

pkamenarsky:
Ok, I think this might be ready for review 🎉

pkamenarsky:
Hm, not sure why the Travis build fails.

spicydonuts:
I’d love to use this in this react-basic branch! I’ve tried a few different approaches to scoping bind and it’s all pretty gross. It’s also harder to understand (both learning and debugging) than qualified do 🙂

https://github.com/spicydonuts/purescript-react-basic/blob/hooks/examples/controlled-input/src/ControlledInput.purs

natefaubion:
I think the approach is straightforward. The only outstanding question I have is how this interacts with something like docs/pursuit. I know `ado` broke some stuff, and since this doesn't introduce additional _constructors_ I think we should be OK, but I'd like @hdgarrood to weigh in.

hdgarrood:
Well I mean there are changes in the desugaring too, but `ado` is already straight up broken with respect to docs, so this can't make it any worse.

spicydonuts:
I gave this branch a try and it works! [Counter](https://github.com/spicydonuts/purescript-react-basic/blob/hooks-qualified-do/examples/counter/src/Counter.purs), [ControlledInput](https://github.com/spicydonuts/purescript-react-basic/blob/hooks-qualified-do/examples/controlled-input/src/ControlledInput.purs)

One thing I noticed is that it seemed to work even without `bind` in my exports list. Not sure if that's intentional or good/bad. I'm also kind of unclear whether `discard` should be included.

spicydonuts:
I also added a few more examples as I've added hooks. The [useRef](https://github.com/spicydonuts/purescript-react-basic/blob/hooks/examples/refs/src/Refs.purs#L23) example shows off the modularity of hooks pretty well.

spicydonuts:
What are the next steps or blockers here?

natefaubion:
> One thing I noticed is that it seemed to work even without bind in my exports list.

Can you explain what you meant by these comments?

spicydonuts:
The module I defined top level `bind` in was using explicit exports. That export list did not include `bind`, but `M.do` still worked when imported as `import M as M`. When I changed the import to `import M (bind) as M` I got a compile error on the import.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -448,8 +448,8 @@ parseValueAtom = withSourceSpan PositionedValue $ P.choice
                  , P.try parseVar
                  , parseCase
                  , parseIfThenElse
-                 , parseDo
-                 , parseAdo
+                 , P.try parseDo
+                 , P.try parseAdo
```
natefaubion: I don't think we want to add backtracking here. This will backtrack on _any_ failure inside a do-block, which will result in poor errors. We only want to apply backtracking to the parsing of the `do` keyword, which means we need to move this into `parseDo` and `parseAdo`.

src/Language/PureScript/Parser/Declarations.hs
```diff
@@ -483,13 +483,13 @@ parseAccessor obj = P.try $ Accessor <$> (indented *> dot *> indented *> parseLa
 
 parseDo :: TokenParser Expr
 parseDo = do
-  m <- parseQualified (reserved "do") >>= \(Qualified m _) -> pure m
+  m <- P.try (parseQualified (reserved "do") >>= pure . getQual) <|> (reserved "do" *> pure Nothing)
```
natefaubion: `>>= pure . k` is just `fmap`.

src/Language/PureScript/Parser/Declarations.hs
natefaubion: `<|>` will only fallback to the next branch if you have wrapped in `try` or you have not consumed any tokens. `parseQualified` likely consumes the qualification, so when it does not get a `do`, it will _not_ fallback to the next alternative unless you explicitly wrap it in `try`.

natefaubion:
@spicydonuts I cannot reproduce that issue with this branch. If I don't export `bind`, I get the expected error. Can you reproduce it with a minimal example? Something would be very awry if that were the case since this desugaring pass is called before name resolution, so to the compiler it's no different than using the operators directly.

spicydonuts:
I tried this [here](https://github.com/spicydonuts/purescript-react-basic/tree/hooks) and it seems like it was probably just a build caching issue:

1. Build
1. Remove `bind` from the exports in `src/React/Basic.purs`
1. Remove the explicit imports on the qualified import of `React.Basic` in `src/React/Basic/Compat.purs` (becomes `import React.Basic as React`)
1. Build -- no errors
1. `rm -rf output`
1. Build -- `Unknown value React.bind`

natefaubion:
By "build", do you mean a full compiler build, or ide build?

natefaubion:
In any case, that doesn't sound like an issue with this PR specifically.

spicydonuts:
Full, using `pulp build` for each of those builds. I'm not sure if incremental builds were happening in the background actually.. I'm using VSCode with the ide plugin but I haven't paid much attention to it for this because it can't parse the qualified do syntax.

natefaubion:
@spicydonuts I've tried, but I still can't reproduce an error calling `purs compile` directly. If you can reproduce it, please open a ticket. Otherwise I'm going to say it's probably IDE related. It doesn't sound like it's related to this feature, specifically.

Does anyone have any other comments?

natefaubion:
Thank you @pkamenarsky for implementing this! Thank you @spicydonuts for trying this out. 🎉

Issue #3223, opened 2018-01-25
Cyclic type class bad error message


LiamGoodacre:
The following code:
```purescript
class A <= B
class B <= A
```

Generates the following error:
```
A cycle appears in a set of type synonym definitions.
Cycles are disallowed because they can lead to loops in the type checker.
Consider using a 'newtype' instead.
```

This confusingly talks about synonyms and newtype, but not type classes.

cdparks:
It looks to me like this happens because the compiler collects binding groups after desugaring typeclasses to type synonyms. Should checking for cycles in superclasses be a separate step before desugaring?

hdgarrood:
This was resolved by #3381.

PR #3381, opened 2018-06-12
Error message for cyclic type class

Saulukass:
Made an attempt on https://github.com/purescript/purescript/issues/3223. Added check for cycles before desugaring type class declarations, now correct error message is shown.

Error message after changes:
```
A cycle appears in a set of type class definitions.
Cycles are disallowed because they can lead to loops in the type checker.
```


hdgarrood:
Thanks for the PR, and sorry this went so long unacknowledged! This looks good, but I think we can improve the error a little more still. We should be able to always include information about the names of the relevant classes, I think. Presumably if we get an `SCC` constructor with exactly one element in the list, that means we have a class which has itself as a superclass? In that case it might be best to print something like
```
A type class may not have itself as a superclass.
```
In the case where we have more than one class in the SCC, I think it would be better to print something like
```
A cycle appears in a set of type class definitions:
  {A, B, C}
```
where `A`, `B`, and `C` are the names of the classes appearing in the `SCC` constructor. I think ideally we'd be able to print an actual cycle like `A -> B -> C -> A`, but looking at the `Data.Graph` docs I think this might be a little more awkward to achieve so I think just printing the set of classes in a strongly-connected component should be fine.

If you're unlikely to pick this up again (which is totally understandable at this point) I'd appreciate it if you could let us know so that someone else can.

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -70,14 +71,33 @@ desugarModule
   => Module
   -> Desugar m Module
 desugarModule (Module ss coms name decls (Just exps)) = do
-  (newExpss, declss) <- unzip <$> parU (sortBy classesFirst decls) (desugarDecl name exps)
-  return $ Module ss coms name (concat declss) $ Just (exps ++ catMaybes newExpss)
+  let classDecls = filter isTypeClassDeclaration decls
```
hdgarrood: Probably better to use [`partition`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-List.html#v:partition) here.

Saulukass:
@hdgarrood Thanks for your response. I have made changes as you suggested.

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -72,14 +73,32 @@ desugarModule
...
-    | isTypeClassDeclaration d1 && not (isTypeClassDeclaration d2) = LT
-    | not (isTypeClassDeclaration d1) && isTypeClassDeclaration d2 = GT
-    | otherwise = EQ
+  desugarClassDecl :: (MonadSupply m, MonadError MultipleErrors m)
+    => ModuleName
+    -> [DeclarationRef]
+    -> SCC Declaration
+    -> Desugar m (Maybe DeclarationRef, [Declaration])
+  desugarClassDecl name' exps' (AcyclicSCC d) = desugarDecl name' exps' d
+  desugarClassDecl _ _ (CyclicSCC [d]) = throwError . errorMessage' (declSourceSpan d) $ CycleInTypeClassDeclaration [(classDeclName d)]
```
hdgarrood: I think we can remove this first `CyclicSCC` pattern now, without changing the behaviour of the function, right?

PR #3383, opened 2018-06-12
break on symbol constraint

justinwoo:
I would really like to have this or something like this to be able to easily work with breaking a Symbol.

Currently, my approaches are to parse individual Cons'd Symbols and try to figure out this logic, which gives me a LOT of empty dictionaries and fairly bad error messages.

Example of this in work: https://github.com/justinwoo/hack-symbol-breakon/blob/master/src/Main.purs

excerpt:

```ps
breakOnSymbol
  :: forall breakOn sym first second
   . Symbol.BreakOn breakOn sym first second
  => SProxy breakOn
  -> SProxy sym
  -> { first :: SProxy first, second :: SProxy second }
breakOnSymbol _ _ = { first: SProxy, second: SProxy }

-- inferred type:
result1 :: { first :: SProxy "a" , second :: SProxy ",b"}
result1 = breakOnSymbol (SProxy :: SProxy ",") (SProxy :: SProxy "a,b")
```

### DIFF
--- a/src/Language/PureScript/Docs/Prim.hs
+++ b/src/Language/PureScript/Docs/Prim.hs
@@ -109,6 +109,8 @@ primSymbolDocsModule = Module
       [ symbolAppend
       , symbolCompare
       , symbolCons
+      , symbolContains
+      , symbolBreakOn
       ]
   , modReExports = []
   }
@@ -450,6 +452,20 @@ symbolCons = primClassOf (P.primSubName "Symbol") "Cons" $ T.unlines
   , "head and tail or for combining a head and tail into a new symbol."
   , "Requires the head to be a single character and the combined string"
   , "cannot be empty."
+  ]
+
+symbolContains :: Declaration
+symbolContains = primClassOf (P.primSubName "Symbol") "Contains" $ T.unlines
+  [ "Compiler solved type class for checking if a Symbol contains a Symbol."
+  ]
+
+symbolBreakOn :: Declaration
+symbolBreakOn = primClassOf (P.primSubName "Symbol") "BreakOn" $ T.unlines
+  [ "Compiler solved type class for breaking a symbol into its head and tail"
+  , "using the whole symbol and a breaking symbol, or for joining together the"
+  , "head and tail with the breaking symbol."
+  , "Does not resolve when the tail is empty, i.e. the breaker is not in the Symbol."
+  , "Use SymbolContains for conditional logic."
   ]
 
 fail :: Declaration
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -171,6 +171,8 @@ entails SolverOptions{..} constraint context hints =
     forClassName _ C.SymbolCompare args | Just dicts <- solveSymbolCompare args = dicts
     forClassName _ C.SymbolAppend args | Just dicts <- solveSymbolAppend args = dicts
     forClassName _ C.SymbolCons args | Just dicts <- solveSymbolCons args = dicts
+    forClassName _ C.SymbolContains args | Just dicts <- solveSymbolContains args = dicts
+    forClassName _ C.SymbolBreakOn args | Just dicts <- solveSymbolBreakOn args = dicts
     forClassName _ C.RowUnion args | Just dicts <- solveUnion args = dicts
     forClassName _ C.RowNub args | Just dicts <- solveNub args = dicts
     forClassName _ C.RowLacks args | Just dicts <- solveLacks args = dicts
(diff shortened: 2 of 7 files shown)


paf31:
Note: the alternative approach here is to use something like `(Symbol.Append first input_ input, Symbol.Append input_ sep second)`, but that means we need support in the compiler for solving two constraints at once, which I think is probably quite difficult, so I like this.

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -419,6 +420,14 @@ symbolCons = primClassOf (P.primSubName "Symbol") "Cons" $ T.unlines
   , "cannot be empty."
   ]
 
+symbolBreakOn :: Declaration
+symbolBreakOn = primClassOf (P.primSubName "Symbol") "BreakOn" $ T.unlines
+  [ "Compiler solved type class for breaking a symbol into its head and tail"
+  , "using the whole symbol and a breaking symbol, or for joining together the"
+  , "head and tail with the breaking symbol."
+  , "Returns empty strings when breaking on a Symbol that is not matched."
```
paf31: Why not just not solve the constraint in this case?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -413,6 +414,28 @@ entails SolverOptions{..} constraint context hints =
...
+    breakOnSymbol breaker@(TypeLevelString breaker') full@(TypeLevelString full') _ _ = do
+      (first, rest) <- T.breakOn <$> decodeString breaker' <*> decodeString full'
+      pure (breaker, full, mkTLString first, mkTLString rest)
+      where mkTLString = TypeLevelString . mkString
+    breakOnSymbol breaker@(TypeLevelString breaker') _ first@(TypeLevelString first') rest@(TypeLevelString rest') = do
+      first'' <- decodeString first'
+      breaker'' <- decodeString breaker'
+      rest'' <- decodeString rest'
+      guard (T.isPrefixOf breaker'' rest'')
+      let full = first'' <> rest''
```
LiamGoodacre: Need to also guard on `first` not containing `breakOn` for the functional dependency to make sense.

src/Language/PureScript/Docs/Prim.hs
justinwoo: But how would I handle cases where I optionally have something? Like a `:` that might be in a given Symbol? At least this way I can chain match on the second (or Fail explicitly on empty)

LiamGoodacre: Something equivalent to a type-level `Maybe (Tuple Symbol Symbol)`?
```
foreign import kind BreakOn
foreign import data Broke :: Symbol -> Symbol -> BreakOn
foreign import data NotBroke :: BreakOn
```
If someone wants it to fail instance selection they can explicitly require `Broke first second`.

justinwoo: Right, but I don't know if I want yet another kind for a class where I mostly want to use something vs ""

justinwoo:
Pushed a commit that makes the BreakOnResult kind its data types. In addition, adds my example file to the test inputs.

src/Language/PureScript/Docs/Prim.hs
paf31: Another alternative is a type-level "contains" relation which you could case on, and which could use the existing type-level boolean kit.

justinwoo: I don't know what you mean, could you elaborate?

paf31: I'm getting at the fact that `BreakOn` is only really necessary because we don't have polykinds, so we have a special-case version of `Maybe` in `Prim` for this one operation. An alternative, if you need to know if the split will be successful, is to implement something like

```purescript
class SymbolContains (sym :: Symbol) (pat :: Symbol) (result :: Boolean) | sym pat -> result
```

which uses the existing type-level `Boolean`.

justinwoo: Oh, now I see. I'll try doing this tomorrow and make the non-break case fail

justinwoo:
I implemented Symbol `Contains` and fixed `BreakOn` to not resolve for null `second` on the break, but we don't have Boolean in Prim? (only in typelevel-prelude, I guess)

For now I put in some type-level stringly types, but should we try to get Boolean into Prim at least?

justinwoo:
Builds timed out. Could someone re-run them if that's the issue?

justinwoo:
Seems like this will pass since we fixed some upstream issues with timeouts and caching

justinwoo:
Rebased this to the changes around Type arguments and SourceAnn

justinwoo:
Cleaning up my github notifications

PR #3389, opened 2018-06-22
add Boolean kind to Prim.Boolean

justinwoo:
Adds Boolean kind and True/False types of Boolean to Prim.Boolean. Related to #3383 

Corresponding typelevel-prelude change: https://github.com/justinwoo/purescript-typelevel-prelude/commit/b82137f42bdc0f880ddc36c3986b337d42a42d73

### DIFF
--- a/src/Language/PureScript/Constants.hs
+++ b/src/Language/PureScript/Constants.hs
@@ -378,6 +378,17 @@ pattern Partial = Qualified (Just Prim) (ProperName "Partial")
 pattern Record :: Qualified (ProperName 'TypeName)
 pattern Record = Qualified (Just Prim) (ProperName "Record")
 
+-- Prim.Boolean
+
+pattern PrimBoolean :: ModuleName
+pattern PrimBoolean = ModuleName [ProperName "Prim", ProperName "Boolean"]
+
+booleanTrue :: Qualified (ProperName 'TypeName)
+booleanTrue = Qualified (Just PrimBoolean) (ProperName "True")
+
+booleanFalse :: Qualified (ProperName 'TypeName)
+booleanFalse = Qualified (Just PrimBoolean) (ProperName "False")
+
 -- Prim.Ordering
 
 pattern PrimOrdering :: ModuleName
@@ -449,7 +460,7 @@ pattern Warn :: Qualified (ProperName 'ClassName)
 pattern Warn = Qualified (Just PrimTypeError) (ProperName "Warn")
 
 primModules :: [ModuleName]
-primModules = [Prim, PrimOrdering, PrimRow, PrimRowList, PrimSymbol, PrimTypeError]
+primModules = [Prim, PrimBoolean, PrimOrdering, PrimRow, PrimRowList, PrimSymbol, PrimTypeError]
 
 -- Data.Symbol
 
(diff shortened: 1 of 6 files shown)


justinwoo:
Builds timed out. Could someone re-run them if that's the issue?

src/Language/PureScript/Constants.hs
```diff
@@ -437,6 +448,12 @@ pattern SymbolAppend = Qualified (Just PrimSymbol) (ProperName "Append")
 pattern SymbolCons :: Qualified (ProperName 'ClassName)
 pattern SymbolCons = Qualified (Just PrimSymbol) (ProperName "Cons")
 
+pattern SymbolContains :: Qualified (ProperName 'ClassName)
+pattern SymbolContains = Qualified (Just PrimSymbol) (ProperName "Contains")
+
+pattern SymbolBreakOn :: Qualified (ProperName 'ClassName)
+pattern SymbolBreakOn = Qualified (Just PrimSymbol) (ProperName "BreakOn")
```
LiamGoodacre: Did you mean to add these in this PR?

LiamGoodacre:
I restarted the timed out job.

src/Language/PureScript/Constants.hs
justinwoo: oh, sorry, applied a patch but it was wrong :sweat_smile:

justinwoo:
4/5 ran within time this time through...

joneshf:
If you need to restart the build and don't have access to it on Travis, you can close this PR and open it again. It will restart the build.

I think it means all builds will run again, but it beats waiting for someone else to come along.

hdgarrood:
No, please don’t do that - instead push an empty commit (as the CI log should suggest). As far as I’m aware that’s the only way to get travis to reuse the progress it made during its previous attempt.

garyb:
I've restarted the failing job a couple of times fwiw 😕

hdgarrood:
Restarting jobs which have timed out is generally not very effective in my experience; we should be encouraging the submitter to push empty commits instead.

LiamGoodacre:
(Instead of empty commits you can regenerate your existing commit with `git commit --amend --no-edit` - this new commit will have a different commit hash.  Then force push.)

joneshf:
An alternate to any of our suggestions could be focusing on making CI more reliable 😄.

joneshf:
Looks like we build the haddocks for dependencies of every job in the matrix whether it has `BUILD_TYPE=haddock` or not. The job that's failing doesn't have `BUILD_TYPE=haddock`. Given that it's timing out when uploading the cache–almost done with the build–that extra time with the haddocks is probably enough to push it over the limit. Should we change the jobs to only build haddocks when necessary?

Is it this line? https://github.com/purescript/purescript/blob/f8ca834ebe3d8e5d484e5b55b5d1b8b6f0a783a2/travis/build.sh#L8-L10

I'm more than willing to submit a PR for it.

nwolverson:
@joneshf I applied that suggestion in my fork here https://github.com/purerl/purescript/commit/5b656c8d945b86c905f3a555d221f60a4dc8328b - not quite sure if it helped or I just had stuff cached by this point, but I think by my understanding this should be right - I believe the cache will be separate for the different entries in our build matrix

hdgarrood:
Sorry for the delay in responding. Good spot - yes I think that would help.

justinwoo:
Heyy, this passed now :)

LiamGoodacre:
Some of the BreakOn and Contains stuff is still in this PR, @justinwoo ?

justinwoo:
Finally removed it

Issue #3390, opened 2018-06-23
Improve error information


Baudin999:
Hi, I've been learning PureScript and I've noticed a little improvement which could make things a little bit easier for newbies (like me):

The original error is:
```
[PureScript]
Type class member mempty has not been implemented.

in type class instance

  Data.Monoid.Monoid Order
Monoid :: Type -> Type
```

I'd suggest doing something like:

```
[PureScript]
Type class member "mempty" has not been implemented in type class instance:

  Data.Monoid.Monoid Order
Monoid :: Type -> Type

mempty :: Order
```

It's mostly a visual change and I've added the type signature of the member which is missing, saves people a lookup and rounds our the error message.

When you make a mistake when implementing a typeclass member the message would look a little better if it looked something like:

Original:

```
[PureScript]
Could not match type

    Person -> Person

  with type

    Person


while checking that type t0 -> t1
  is at least as general as type Person
while checking that expression \$6 ->
                                 case $6 of
                                   (Person p) -> temp
  has type Person
in value declaration semiringPerson

where t0 is an unknown type
      t1 is an unknown type
```

Suggestion:

```
[PureScript]
Could not match type:

    one :: Person -> Person

  with type

    one :: Person


while checking that type t0 -> t1
  is at least as general as type Person
while checking that expression \$6 ->
                                 case $6 of
                                   (Person p) -> temp
  has type Person
in value declaration semiringPerson

where t0 is an unknown type
      t1 is an unknown type
```

In the original error I couldn't find the actual member name of the member throwing the error; just throwing in the actual name of the member helps a lot.

Baudin999:
Should I close this issue or be patient?

LiamGoodacre:
@Baudin999, thanks for raising this issue and sorry that no-one responded promptly.
As you can see above this sparked multiple pieces of work.
Thanks again!

Baudin999:
Thank you for the response! I know how hectic things can get.

Should I close this issue?

hdgarrood:
No need just yet, we can close it once we've come to a decision on how to proceed.

hdgarrood:
Since #3398 addresses the first part, and the second part is best served by improving error spans (
#1801) and showing actual code in errors (#1898), I think this can now be closed.

PR #3398, opened 2018-07-09
Print types of missing typeclass members

fehrenbach:
This fixes the first part of #3390. It does not come with a promise to do the second.

```PureScript
module Main where

class A a where
  a :: a -> String
  b :: a -> Number

instance aString :: A String where
  a s = s
```
errors with this:
```
  Type class member b has not been implemented. It is expected to have the following type:
                         
    b :: String -> Number
```

instead of

```
Type class member b has not been implemented.
```


fehrenbach:
Thinking about this some more: is there a reason not to print all of the missing members?

garyb:
No, that'd be good too 🙂

kRITZCREEK:
Thanks!

PR #3400, opened 2018-07-14
Use `microlens-platform` instead of `lens`

joneshf:
#3389 inspired me to look at the build and see what was causing it to take as much time as it does.

One thing is that we've got a dependency on `lens`, which is a pretty big dependency. Turns out we don't use too much of `lens` and we could drop it in favor of `microlens-platform`–which is a much smaller dependency. Chose `microlens-platform` instead of `microlens` (which is even smaller) because we do some `ix` stuff and we're already paying the cost of things like `unordered-containers`. We can totally go with the smaller footprint ones if we want to rewrite some code. Anyway, the change was small and quick, so I made a PR.

The result of this change on my laptop is about 10% decrease in build time for dependencies:

* With `lens`
    ```
    $ time stack build --only-dependencies                  
    ...
    Completed 139 action(s).     
    stack build --only-dependencies  4087.45s user 195.29s system 537% cpu 13:16.57 total
    ```
* With `microlens-platform`
    ```
    $ time stack build --only-dependencies                  
    ...
    Completed 128 action(s).                      
    stack build --only-dependencies  4016.97s user 187.27s system 590% cpu 11:51.75 total
    ```

It moves the travis builds closer to being within the time limit. It doesn't make the travis builds 100% reliable since those computers have fewer resources and we still have a good deal of dependencies, but they're quicker than currently so more reliable.

In any case, if we're interested in this change, I can clean up some stuff in the PR. I didn't want to spend too much time on an experiment :smiley:.

### DIFF
--- a/src/Language/PureScript/Ide/Types.hs
+++ b/src/Language/PureScript/Ide/Types.hs
@@ -22,11 +22,11 @@ module Language.PureScript.Ide.Types where
 import           Protolude hiding (moduleName)
 
 import           Control.Concurrent.STM
-import           Control.Lens.TH
 import           Data.Aeson
 import qualified Data.Map.Lazy as M
 import qualified Language.PureScript as P
 import qualified Language.PureScript.Errors.JSON as P
+import           Lens.Micro.Platform hiding ((.=))
 
 type ModuleIdent = Text
 type ModuleMap a = Map P.ModuleName a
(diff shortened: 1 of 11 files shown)


src/Language/PureScript/Ide/Types.hs
```diff
@@ -94,7 +94,41 @@ data IdeTypeOperator = IdeTypeOperator
   , _ideTypeOpKind          :: Maybe P.Kind
   } deriving (Show, Eq, Ord, Generic, NFData)
 
-makePrisms ''IdeDeclaration
+_IdeDeclValue :: Traversal' IdeDeclaration IdeValue
```
joneshf: Looks like we can write these as traversals rather than prisms without losing anything in our logic. Is that right?

src/Language/PureScript/Ide/Types.hs
```diff
@@ -94,7 +94,41 @@ data IdeTypeOperator = IdeTypeOperator
...
+
+_IdeDeclTypeOperator :: Traversal' IdeDeclaration IdeTypeOperator
+_IdeDeclTypeOperator f (IdeDeclTypeOperator x) = map IdeDeclTypeOperator (f x)
+_IdeDeclTypeOperator _ x = pure x
+
+_IdeDeclKind :: Traversal' IdeDeclaration (P.ProperName 'P.KindName)
+_IdeDeclKind f (IdeDeclKind x) = map IdeDeclKind (f x)
+_IdeDeclKind _ x = pure x
+
+anyOf :: Getting Any s a -> (a -> Bool) -> s -> Bool
```
joneshf: This doesn't really belong here, but was used in four modules so I stuffed it here. If we want these changes, I'll find a better home for it (or remove it maybe?).

src/Language/PureScript/Ide/Util.hs
```diff
@@ -90,14 +90,14 @@ encodeT = TL.toStrict . TLE.decodeUtf8 . encode
 decodeT :: (FromJSON a) => Text -> Maybe a
 decodeT = decode . TLE.encodeUtf8 . TL.fromStrict
 
-properNameT :: Iso' (P.ProperName a) Text
-properNameT = iso P.runProperName P.ProperName
+properNameT :: Getting r (P.ProperName a) Text
```
joneshf: We ended up not needing the full power of an iso here. Is that right?

joneshf:
cc @kRITZCREEK since this is almost all IDE code and I figure you've probably got the most context on whether this is an acceptable move or not.

src/Language/PureScript/Ide/Util.hs
LiamGoodacre: If all we need is a `SimpleGetter` then we might as well just use `to P.runProperName` in the usage.

kRITZCREEK:
Honestly I don't "need" the lens stuff in the IDE part of the codebase. If it's too much of a hassle, or it breaks with the remaining style too much just let me know and I'll rip it out.

kRITZCREEK:
Thanks!

PR #3416, opened 2018-08-21
Eliminate built type class dicts when necessarily empty

### LINKED ISSUE #2768: Remove runtime overhead of typeclass instances with no dictionary

LiamGoodacre:
Had a quick go at this.  I want to switch to `undefined` instead of `{}`, but so far it seems to work.  I want to test it on a lot more dictionary heavy code, though.
Fixes #2768

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -753,6 +753,12 @@ data Expr
   --
   | App Expr Expr
   -- |
+  -- Hint that an expression is unused.
+  -- This is used to ignore type class dictionaries that are necessarily empty.
+  -- The inner expression lets us solve subgoals before eliminating the whole expression.
+  -- The code gen will render this as `undefined`, regardless of what the inner expression is.
+  | Unused Expr
+  -- |
   -- Variable
   --
   | Var SourceSpan (Qualified Ident)
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -125,17 +125,23 @@ addTypeClass
   -> m ()
 addTypeClass qualifiedClassName args implies dependencies ds = do
   env <- getEnv
-  traverse_ (checkMemberIsUsable (typeSynonyms env)) classMembers
+  let newClass = mkNewClass env
+  traverse_ (checkMemberIsUsable newClass (typeSynonyms env)) classMembers
   modify $ \st -> st { checkEnv = (checkEnv st) { typeClasses = M.insert qualifiedClassName newClass (typeClasses . checkEnv $ st) } }
   where
     classMembers :: [(Ident, SourceType)]
     classMembers = map toPair ds
 
-    newClass :: TypeClassData
-    newClass = makeTypeClassData args classMembers implies dependencies
+    mkNewClass :: Environment -> TypeClassData
+    mkNewClass env = makeTypeClassData args classMembers implies dependencies ctIsEmpty
+      where
+      ctIsEmpty = null classMembers && all (typeClassIsEmpty . findSuperClass) implies
+      findSuperClass c = case M.lookup (constraintClass c) (typeClasses env) of
+        Just tcd -> tcd
+        Nothing -> internalError "Unknown super class in TypeClassDeclaration"
 
-    coveringSets :: [S.Set Int]
-    coveringSets = S.toList (typeClassCoveringSets newClass)
+    coveringSets :: TypeClassData -> [S.Set Int]
+    coveringSets = S.toList . typeClassCoveringSets
 
     argToIndex :: Text -> Maybe Int
     argToIndex = flip M.lookup $ M.fromList (zipWith ((,) . fst) args [0..])
@@ -146,11 +152,11 @@ addTypeClass qualifiedClassName args implies dependencies ds = do
     -- Currently we are only checking usability based on the type class currently
     -- being defined.  If the mentioned arguments don't include a covering set,
     -- then we won't be able to find a instance.
-    checkMemberIsUsable :: T.SynonymMap -> (Ident, SourceType) -> m ()
-    checkMemberIsUsable syns (ident, memberTy) = do
+    checkMemberIsUsable :: TypeClassData -> T.SynonymMap -> (Ident, SourceType) -> m ()
+    checkMemberIsUsable newClass syns (ident, memberTy) = do
       memberTy' <- T.replaceAllTypeSynonymsM syns memberTy
       let mentionedArgIndexes = S.fromList (mapMaybe argToIndex (freeTypeVariables memberTy'))
-      let leftovers = map (`S.difference` mentionedArgIndexes) coveringSets
+      let leftovers = map (`S.difference` mentionedArgIndexes) (coveringSets newClass)
 
       unless (any null leftovers) . throwError . errorMessage $
         let
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -207,9 +207,14 @@ entails SolverOptions{..} constraint context hints =
             -- We need information about functional dependencies, so we have to look up the class
             -- name in the environment:
             classesInScope <- lift . lift $ gets (typeClasses . checkEnv)
-            TypeClassData{ typeClassDependencies } <- case M.lookup className' classesInScope of
-              Nothing -> throwError . errorMessage $ UnknownClass className'
-              Just tcd -> pure tcd
+
+            TypeClassData
+              { typeClassDependencies
+              , typeClassIsEmpty
+              } <- case M.lookup className' classesInScope of
+                Nothing -> throwError . errorMessage $ UnknownClass className'
+                Just tcd -> pure tcd
+
             let instances = do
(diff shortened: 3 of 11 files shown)


LiamGoodacre:
(Also bumped bower version in the test support code - it was failing to install dependencies)

LiamGoodacre:
### Input:
```purescript
data P t = P
class IsUnit unit | -> unit
instance isUnitInst :: IsUnit Unit

class IsUnit2 unit | -> unit
instance isUnitInst2 :: IsUnit unit => IsUnit2 unit

foo :: P _
foo = P :: forall unit . IsUnit unit => P unit

bar :: P _
bar = P :: forall unit . IsUnit2 unit => P unit
```

### Previous output:
```javascript
var foo = (function (dictIsUnit) { return P.value; })(isUnitInst);
var bar = (function (dictIsUnit2) { return P.value; })(isUnitInst2(isUnitInst));
```

### Current progress:
```javascript
var foo = (function (dictIsUnit) { return P.value; })({});
var bar = (function (dictIsUnit2) { return P.value; })({});
```

### Goal:
```javascript
var foo = (function () { return P.value; })();
var bar = (function () { return P.value; })();
```

LiamGoodacre:
Switched to `undefined`.  For the above example, it now produces:
```javascript
var foo = P.value;
var bar = P.value;
```

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -743,6 +743,9 @@ data Expr
   --
   | App Expr Expr
   -- |
+  -- Hint that an expression is unused
+  | Unused Expr
```
paf31: Might be worth explaining why we need the extra `Expr` at all here (i.e. because we need to track unsolved constraints, I assume).

src/Language/PureScript/TypeChecker.hs
```diff
@@ -124,17 +124,23 @@ addTypeClass
...
   where
     classMembers :: [(Ident, Type)]
     classMembers = map toPair ds
 
-    newClass :: TypeClassData
-    newClass = makeTypeClassData args classMembers implies dependencies
+    mkNewClass :: Environment -> TypeClassData
+    mkNewClass env = makeTypeClassData args classMembers implies dependencies ctIsEmpty
+      where
+      ctIsEmpty = null classMembers && all (typeClassIsEmpty . findSuperClass) implies
```
paf31: If there is a superclass, is it possible that we try to solve a constraint for the superclass when only an instance for the subclass is in scope, and then by looking up the superclass as a property on the subclass dictionary, we run into a problem at runtime?

paf31: I guess you might need a chain of two or more superclasses for it to become an issue though, otherwise you'll just get `undefined` anyway.

src/Language/PureScript/AST/Declarations.hs
LiamGoodacre: Yes that's right, will update, thanks!

src/Language/PureScript/TypeChecker.hs
LiamGoodacre: If there is a non-empty super class, then this class isn't considered empty.  So there shouldn't be any issues with looking up super classes, as I don't mark those ones as unused.

paf31: I think it would be worth adding a test for this though (and maybe a few other cases) just to make sure we don't break that assumption in future. To be clear, the reasoning is that even if the subclass and superclass are empty, we'll never look up the superclass dictionary at runtime because it would be empty if we did. Right?

LiamGoodacre: You're right, I need to think more about subclass lookup.  I think the fix might just be moving the `Unused` hint slightly further up the dictionary expression.  Thanks!

LiamGoodacre:
Pushed a fix for the subgoal lookup, but haven't properly tested yet.

LiamGoodacre:
With the latest code....

### Example:
```purescript
class WithArgEmpty t
instance withArgEmptyCheck :: WithArgEmpty Check
class WithArgEmpty t <= WithArgHasEmptySuper t where
  withArgHasEmptySuper :: t
instance withArgHasEmptySuperCheck :: WithArgHasEmptySuper Check where
  withArgHasEmptySuper = Check
whenAccessingSuperDict :: Check
whenAccessingSuperDict = foo Check where
  bar :: forall t . WithArgEmpty t => t -> t
  bar x = x
  foo :: forall t . WithArgHasEmptySuper t => t -> t
  foo x = bar x
```

### Before:
```javascript
var withArgEmptyCheck = WithArgEmpty;
var withArgHasEmptySuperCheck = new WithArgHasEmptySuper(function () {
    return withArgEmptyCheck;
}, Check.value);
var whenAccessingSuperDict = (function () {
    var bar = function (dictWithArgEmpty) {
        return function (x) {
            return x;
        };
    };
    var foo = function (dictWithArgHasEmptySuper) {
        return function (x) {
            return bar(dictWithArgHasEmptySuper.WithArgEmpty0())(x);
        };
    };
    return foo(withArgHasEmptySuperCheck)(Check.value);
})();
```

### After:
```javascript
var withArgEmptyCheck = WithArgEmpty;
var withArgHasEmptySuperCheck = new WithArgHasEmptySuper(function () {
    return undefined;
}, Check.value);
var whenAccessingSuperDict = (function () {
    var bar = function (dictWithArgEmpty) {
        return function (x) {
            return x;
        };
    };
    var foo = function (dictWithArgHasEmptySuper) {
        return function (x) {
            return bar()(x);
        };
    };
    return foo(withArgHasEmptySuperCheck)(Check.value);
})();
```

That is, the call to `bar` in `foo` no-longer bothers to look up the super class dict.  But even if it did, it would be undefined anyway.

LiamGoodacre:
I think this is ready to look at again.  I've also added a test.

garyb:
This is pretty exciting! My silence on it is not due to lack of enthusiasm, just lack of time 😄

doolse:
Really looking forward to this being released! :+1: 
It inspired me to create [purescript-tscompat](https://github.com/doolse/purescript-tscompat/) because the idea of zero runtime overhead typeclasses was too enticing.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -206,9 +206,14 @@ entails SolverOptions{..} constraint context hints =
...
             classesInScope <- lift . lift $ gets (typeClasses . checkEnv)
-            TypeClassData{ typeClassDependencies } <- case M.lookup className' classesInScope of
-              Nothing -> throwError . errorMessage $ UnknownClass className'
-              Just tcd -> pure tcd
+
+            TypeClassData
+              { typeClassDependencies
+              , typeClassIsEmpty
+              } <- case M.lookup className' classesInScope of
+                Nothing -> throwError . errorMessage $ UnknownClass className'
```
kRITZCREEK: Is this a sign for a programming error in the compiler, or an actual user facing error?

In the first case I'd prefer an internalError, and in the second case this really ought to have a source span (unless this is wrapped somewhere and gets a source span)

LiamGoodacre: It's the first case, I believe.  Will change to `internalError`, thanks!

LiamGoodacre: Actually, turns out, it's not the first case. :sweat_smile:

LiamGoodacre: The errors appear to be propagated through `typesOf` and into the type checker; where source spans get added and the errors re-thrown.

hdgarrood:
Apart from the conflicts (which are very minor), is there anything getting in the way of merging this?

garyb:
Not as far as I'm concerned, I forgot it wasn't merged already actually!

LiamGoodacre:
@hdgarrood, I don't think so.  I'll try to get this updated tonight.

natefaubion:
Here is a comparison of the output of master vs this branch:
https://gist.github.com/natefaubion/fa442e3c075bebf82f6026776bf16733

Specifically lines 20-22.

reactormonk:
All tests on our work projects run just fine.

hdgarrood:
I've just tested this on our code at work and it seems to work fine for me too. Shall we try to get this merged? @LiamGoodacre has the approach changed much since this was last reviewed, or is there anything in particular you think wants looking at? Since this passed review already I'm tempted to just merge it.

LiamGoodacre:
@hdgarrood I'm happy to have this merged now :+1:

hdgarrood:
Nice, thanks! :)

PR #3419, opened 2018-08-23
MagicDo: Desugar ForE

matthewleon:
The "MagicDo" `CoreImp` optimization currently desugars `Purescript.Effect`'s `untilE` and `whileE` loops. This PR adds desugaring for `forE`.

I also add a note regarding desugaring of `foreachE`: I could augment this PR with support for it, but it would require a change to the `CoreImp` AST to allow for initialization of multiple variables in `For`'s initialization expression. Happy to go ahead with this, or make a separate PR for it, if it is agreed to be a permissible change.

### DIFF
--- a/src/Language/PureScript/Constants.hs
+++ b/src/Language/PureScript/Constants.hs
@@ -204,6 +204,15 @@ unit = "unit"
 
 -- Core lib values
 
+untilE :: forall a. (IsString a) => a
+untilE = "untilE"
+
+whileE :: forall a. (IsString a) => a
+whileE = "whileE"
+
+forE :: forall a. (IsString a) => a
+forE = "forE"
+
 runST :: forall a. (IsString a) => a
 runST = "run"
 
--- a/src/Language/PureScript/Constants.hs
+++ b/src/Language/PureScript/Constants.hs
@@ -204,15 +204,6 @@ unit = "unit"
 
 -- Core lib values
 
-untilE :: forall a. (IsString a) => a
-untilE = "untilE"
-
-whileE :: forall a. (IsString a) => a
-whileE = "whileE"
-
-forE :: forall a. (IsString a) => a
-forE = "forE"
-
 runST :: forall a. (IsString a) => a
 runST = "run"
 
@@ -262,6 +253,7 @@ data EffectDictionaries = EffectDictionaries
   , edMonadDict :: PSString
   , edWhile :: PSString
   , edUntil :: PSString
+  , edFor :: PSString
   }
 
 effDictionaries :: EffectDictionaries
@@ -271,6 +263,7 @@ effDictionaries = EffectDictionaries
   , edMonadDict = "monadEff"
   , edWhile = "whileE"
   , edUntil = "untilE"
+  , edFor = "forE"
   }
 
 effectDictionaries :: EffectDictionaries
@@ -280,6 +273,7 @@ effectDictionaries = EffectDictionaries
   , edMonadDict = "monadEffect"
   , edWhile = "whileE"
   , edUntil = "untilE"
+  , edFor = "forE"
   }
 
 stDictionaries :: EffectDictionaries
@@ -289,6 +283,7 @@ stDictionaries = EffectDictionaries
   , edMonadDict = "monadST"
   , edWhile = "while"
   , edUntil = "until"
+  , edFor = "for"
   }
 
 discardUnitDictionary :: forall a. (IsString a) => a
--- a/src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
@@ -56,10 +56,22 @@ magicDo effectModule C.EffectDictionaries{..} = everywhereTopDown convert
     App s1 (Function s1 Nothing [] (Block s1 [ While s1 (Unary s1 Not (App s1 arg [])) (Block s1 []), Return s1 $ ObjectLiteral s1 []])) []
   -- Desugar whileE
   convert (App _ (App _ (App s1 f [arg1]) [arg2]) []) | isEffFunc edWhile f =
-    App s1 (Function s1 Nothing [] (Block s1 [ While s1 (App s1 arg1 []) (Block s1 [ App s1 arg2 [] ]), Return s1 $ ObjectLiteral s1 []])) []
+    App s1 (Function s1 Nothing []
+            (Block s1 [
+              While s1 (App s1 arg1 [])
+                    (Block s1 [ App s1 arg2 [] ]),
+                    Return s1 $ ObjectLiteral s1 []])) []
   -- Desugar forE
-  convert (App _ (App _ (App _ (App s1 f [loArg]) [hiArg]) [intToEff]) []) | isEffFunc C.forE f =
-    App s1 (Function s1 Nothing [] (Block s1 [ VariableIntroduction s1 lo (Just loArg), VariableIntroduction s1 hi (Just hiArg), VariableIntroduction s1 fn (Just intToEff), For s1 counter (Var s1 lo) (Var s1 hi) (Block s1 [App s1 (App s1 (Var s1 fn) [Var s1 counter]) []]), Return s1 $ ObjectLiteral s1 []])) []
+  convert (App _ (App _ (App _ (App s1 f [loArg]) [hiArg]) [intToEff]) [])
+    | isEffFunc C.forE f =
+        App s1 (Function s1 Nothing []
+                 (Block s1 [
+                   VariableIntroduction s1 lo (Just loArg),
+                   VariableIntroduction s1 hi (Just hiArg),
+                   VariableIntroduction s1 fn (Just intToEff),
+                   For s1 counter (Var s1 lo) (Var s1 hi)
+                       (Block s1 [App s1 (App s1 (Var s1 fn) [Var s1 counter]) []]),
+                   Return s1 $ ObjectLiteral s1 []])) []
     where
     counter = "$__i"
     lo = "$__lo"
--- a/src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
@@ -58,10 +58,12 @@ magicDo effectModule C.EffectDictionaries{..} = everywhereTopDown convert
   convert (App _ (App _ (App s1 f [arg1]) [arg2]) []) | isEffFunc edWhile f =
     App s1 (Function s1 Nothing [] (Block s1 [ While s1 (App s1 arg1 []) (Block s1 [ App s1 arg2 [] ]), Return s1 $ ObjectLiteral s1 []])) []
   -- Desugar forE
-  convert (App _ (App _ (App _ (App s1 f [lo]) [hi]) [intToEff]) []) | isEffFunc C.forE f =
-    App s1 (Function s1 Nothing [] (Block s1 [ VariableIntroduction s1 fn (Just intToEff), For s1 counter lo hi (Block s1 [App s1 (App s1 (Var s1 fn) [Var s1 counter]) []]), Return s1 $ ObjectLiteral s1 []])) []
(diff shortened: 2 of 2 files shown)


matthewleon:
### input:

```purescript
main :: Effect Unit
main = do
  forE 0 10 logShow
  logShow "done"
```

### before:
```javascript
var main = function __do() {
    Effect.forE(0)(10)(Effect_Console.logShow(Data_Show.showInt))();
    return Effect_Console.logShow(Data_Show.showString)("done")();
};
```

### after:
```javascript
var main = function __do() {
    (function () {
        for (var i = 0; i < 10; i++) {
            Effect_Console.logShow(Data_Show.showInt)(i)();
            return {};
        };
    })();
    return Effect_Console.logShow(Data_Show.showString)("done")();
};
```

LiamGoodacre:
That `return {};` doesn't look quite right:
```javascript
for (var i = 0; i < 10; i++) {
    Effect_Console.logShow(Data_Show.showInt)(i)();
    return {};
};
```

matthewleon:
> That return {}; doesn't look quite right:

D'oh! Good catch there.

I also want to check that naming the passed in function `i` ends up working out. I fear it might not, in which case I will have to be a bit innovative.

matthewleon:
As I feared, passing in a function with the same name as the counter (`i`) results in madness.

Two possibilities here:

1) Give the counter a name that a purescript binding cannot possibly take. This should be safe the loop lives in an IIFE/block.

2) Introduce an unused binding name. Not sure if `CoreImp` provides facilities for this. And if it did, this would significantly change the signature of `MagicDo`.

I'm wondering if these are the kind of difficulties that stopped this optimization from being implemented in the first place. Option 1) might be doable, though. I'll have to see what the differences are between JS and PS naming rules.

matthewleon:
@LiamGoodacre This is ready for review again. I've renamed the loop counter to `$__i`, which, like `$__unused`, keeps us safe from clashing with a PureScript binding.

matthewleon:
New compilation output:

```purescript
var main = function __do() {
    (function () {
        for (var $__i = 0; $__i < 10; $__i++) {
            Effect_Console.logShow(Data_Show.showInt)($__i)();
        };
        return {};
    })();
    return Effect_Console.logShow(Data_Show.showString)("done")();
};
```

matthewleon:
I've updated this to address the problem raised in comment https://github.com/purescript/purescript/pull/3422#issuecomment-415994598

Here is the new output:

```javascript
var main = function __do() {
    (function () {
        var $__f = Effect_Console.logShow(Data_Show.showInt);
        for (var $__i = 0; $__i < 10; $__i++) {
            $__f($__i)();
        };
        return {};
    })();
    return Effect_Console.logShow(Data_Show.showString)("done")();
};
```

LiamGoodacre:
Thanks, these are looking great!
I think we need to do the same upfront evaluation for all the arguments (`0`, `10`, `nums`).  Suppose I have some complicated expression to compute `10`, this would currently be embedded in the `for` condition and would have to be recomputed every iteration.

matthewleon:
> I think we need to do the same upfront evaluation for all the arguments (0, 10, nums).

Agreed. I've pushed a new commit that does this. I've modified my source to avoid inlining...

### Source

```purescript
main = do
  forE (1 - 1) (11 - 1) logShow
  logShow "done"
```

### Output

```javascript
var main = function __do() {
    (function () {
        var $__lo = 1 - 1 | 0;
        var $__hi = 11 - 1 | 0;
        var $__f = Effect_Console.logShow(Data_Show.showInt);
        for (var $__i = $__lo; $__i < $__hi; $__i++) {
            $__f($__i)();
        };
        return {};
    })();
    return Effect_Console.logShow(Data_Show.showString)("done")();
};
```

paf31:
Is there any risk of name capture here?

matthewleon:
> Is there any risk of name capture here?

The introduced variable names are prefixed with `$__` so as not to conflict with any PureScript names, and are introduced in an IIFE so as to shadow, rather than conflict, with any existing variables. I'm having trouble thinking of how name capture would slip in here.

Of course, such things can be surprising. Can you think of an example where it would happen?

paf31:
Sorry I meant the `var` introduced in the `for` could be captured by an anonymous function perhaps.

matthewleon:
> Sorry I meant the var introduced in the for could be captured by an anonymous function perhaps.

I believe the line where the passed-in function is assigned to `$__f` before the loop ought to prevent this. No matter what, the loop body passes `$__i` as an argument to a the `$__f` function, such that if its value gets captured within the function, it won't be mutated on the next run through the loop.

`var $__f = Effect_Console.logShow(Data_Show.showInt);`

Does this make sense?

paf31:
What if `$__f` is itself expanded and inlined though? Is that not possible?

paf31:
TCO doesn't hit this case if I remember correctly, because it uses function calls to avoid it. Some of those will later be inlined away, but it's why you see things like `(function() { ... })()` in TCO code sometimes.

matthewleon:
Can we brainstorm potential solutions here?

1) Easiest but probably controversial: Declare the counter variable with `let`. Even though `language-javascript` hasn't merged in support for block-scoped variables, this should actually work because the CodeGen for `For` loops prints out the required JS "by hand." This would probably be "controversial" because it would be the first time, to my knowledge, that PureScript codegen becomes ES6.

2) Works, but probably defeats the point: wrap the loop body in an IIFE where the current value of the counter (`$__i`) is passed in. I say this probably defeats the point because I'd guess that it would decrease the performance enough that in many cases the "optimization" actually results in slower code. Would have to benchmark to find out, though.

3) Tricky: Build enough intelligence into the inliner to stop it from inlining within the body of `For`? (Perhaps things already work this way, but I'd have to show that somehow).

Would appreciate any thoughts here.

matthewleon:
I have a strong suspicion that we might actually safe here and that this won't be inlined, but I'm not sure how to prove it just yet. Even very simple functions don't seem to get inlined though. Example:

```purescript
main = do
  forE 0 10 (\_ -> log "doing")
  log "done"
```

```javascript
var main = function __do() {
    (function () {
        var $__f = function (v) {
            return Effect_Console.log("doing");
        };
        for (var $__i = 0; $__i < 10; $__i++) {
            $__f($__i)();
        };
        return {};
    })();
    return Effect_Console.log("done")();
};
```

I'll have to take a closer look at the workings of the `CoreImp` inliner to see if we are truly safe.

matthewleon:
Okay, so I've just read through all of the `CoreImp` `Inliner.hs` code and believe that we are actually totally safe.

There is a very explicit list of functions that are inlined in `CoreImp`:

https://github.com/purescript/purescript/blob/e184fca6ffd04de45d4d8c4f6747230ae266c6d4/src/Language/PureScript/CoreImp/Optimizer/Inliner.hs#L109

The inliner does not seem to touch arbitrary functions in `CoreImp`, perhaps because doing so would raise exactly this problem of potentially capturing a value that gets mutated. If it behaved otherwise, I believe it should be considered a bug in the inliner, no?

@paf31 please let me know if my thinking here makes sense.

paf31:
All I'm saying is that I remember this being an issue for TCO in some weird case and I'm fairly sure there is a passing test case for it now. Unless we're absolutely sure, I'd be reluctant to add this and risk breaking working code.

matthewleon:
The TCO test: https://github.com/purescript/purescript/blob/master/tests/purs/passing/2689.purs

The initial bug: https://github.com/purescript/purescript/issues/2689

The relevant PR: https://github.com/purescript/purescript/pull/2719

A similar bug with variable capture would only be triggered here if the inliner caused the passed-in function, which is assigned to `$__f`, to be inlined in the body of our `for` loop. I'm not sure how to prove beyond any shadow of a doubt that the inliner does not behave this way. It doesn't _look_ to me like it does, but I understand that's not a really satisfying answer.

I guess the question, then, is if there is any way for me to demonstrate, using a test, that the inliner does not do this? I've been doing it informally be whipping up very minimal examples of code and seeing that it is getting compiled correctly... It certainly falls short of being a mathematically correct proof, but would adding such an example here as a test, similar to the TCO test, be reassuring enough?

paf31:
Couldn't magic do itself cause something to be inlined there?

matthewleon:
> Couldn't magic do itself cause something to be inlined there?

I don't see how.  I seem unable to cause this to happen. Is there some trick you're thinking of that would trigger inlining?

natefaubion:
AFAIK, from my reading of optimization code, I'm not aware of anything that will inline function bodies or function applications. Only simple expressions like property accessors.

src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
```diff
@@ -54,6 +54,14 @@ magicDo'' effectModule C.EffectDictionaries{..} = everywhereTopDown convert
   -- Desugar whileE
   convert (App _ (App _ (App s1 f [arg1]) [arg2]) []) | isEffFunc C.whileE f =
     App s1 (Function s1 Nothing [] (Block s1 [ While s1 (App s1 arg1 []) (Block s1 [ App s1 arg2 [] ]), Return s1 $ ObjectLiteral s1 []])) []
+  -- Desugar forE
+  convert (App _ (App _ (App _ (App s1 f [loArg]) [hiArg]) [intToEff]) []) | isEffFunc C.forE f =
+    App s1 (Function s1 Nothing [] (Block s1 [ VariableIntroduction s1 lo (Just loArg), VariableIntroduction s1 hi (Just hiArg), VariableIntroduction s1 fn (Just intToEff), For s1 counter (Var s1 lo) (Var s1 hi) (Block s1 [App s1 (App s1 (Var s1 fn) [Var s1 counter]) []]), Return s1 $ ObjectLiteral s1 []])) []
```
natefaubion: Would it hurt to split this up onto multiple lines? 😜

matthewleon: It would not :). Will do so.

matthewleon:
@natefaubion have rebased and cleaned up the style. It's admittedly somewhat ad-hoc, but hopefully makes things more readable.

src/Language/PureScript/Constants.hs
```diff
@@ -204,6 +204,15 @@ unit = "unit"
...
 
+untilE :: forall a. (IsString a) => a
+untilE = "untilE"
+
+whileE :: forall a. (IsString a) => a
+whileE = "whileE"
+
+forE :: forall a. (IsString a) => a
+forE = "forE"
+
```
natefaubion: I moved `untilE` and `whileE` to the `EffectDictionaries` data type so that the implementations can be shared between Effect and ST. The same can be done with `forE`, which is `for` for ST.

matthewleon: done

natefaubion:
I'm not sure how much of an optimization this is in it's current form. It's an exact inlining of `forE`, but what exactly are we saving? There are a couple of things that I think would make this a clear optimization.

One is optimizing to essentially `EffectFn1` when you have an eta-expanded lambda with an inner `__do`:

This:
```purescript
forE lo hi \x -> do ...
```
```js
forE(li)(hi)(function (x) { return __do() { ... } });
```
Currently optimizes to:
```js
var f = function(x) { return function __do() { ... } };
for (var i = lo; i < hi; i++) { f(i)(); }
```
All this really saves is maybe the overhead of currying `forE`, which is minimal. But it could optimize to:
```purescript
var f = function(x) { ... }
for (var i = lo; i < hi; i++) { f(i) }
```
Which saves the allocations of Effect thunks at each step.

The other involves going further with the same setup, and inlining the function all together (your original optimization) as long as `i` is not captured under another lambda. Would you be interested in doing either of these?

matthewleon:
@natefaubion very good ideas, thanks. I'll pursue the first optimization suggestion for now. While the more advanced technique is certainly desirable, I recall being a bit intimidated by the machinery involved the last time I took a look. It could always be broken out as a separate, further PR/Issue. What do you think?

garyb:
I wouldn't say there is any particular plan to move to ES6 soon.

hdgarrood:
I'm going to suggest closing this for now. As @natefaubion points out, this isn't much of an optimization in its current form, and although it seems unlikely to break anything, it's difficult to be absolutely sure, so the risk-reward ratio doesn't really seem worth it to me.

PR #3427, opened 2018-09-08
Common JS to ES6/7 modules - WIP

Cmdv:
This is a continuation of https://github.com/purescript/purescript/pull/3379, don't have creds to re-open PR as I closed it.

This does currently work but it's not getting full benefits because any FFI usage still uses CJS so any Dead Code Elimination can't be used through out the whole codebase so this is still a WIP.

For the ability to have ES6/7 FFI we need to add the new ES6/7 `import` & `export` specs to [language-javascript](https://github.com/erikd/language-javascript)

To do:

- [x] fix compiler errors that came from master merge
- [x] Exports to `language-javascript` https://github.com/erikd/language-javascript/pull/78
- [x] Imports to `language-javascript` https://github.com/erikd/language-javascript/pull/77
- [ ] add / update tests to handle new `import/export` [here](https://github.com/purescript/purescript/blob/master/tests/TestCompiler.hs#L244-L269)
- [ ] trailing `;`


Cmdv:
Thanks to @gasi `Export` has been added to `language-javascript`.

All this would now need is `imports` , tests and removal of trailling semicolon.

hdgarrood:
@Cmdv I've managed to get import statements working with language-javascript, see https://github.com/erikd/language-javascript/pull/82. With that in mind, do you think you're likely to pick this back up again soon?

Cmdv:
@hdgarrood ah cool that's ace, had totally forgotten about this! I think I had a funny trailing `;` from memory and the rest should be ok. Will investigate in next couple of days.

gasi:
I also have some WIP for generating ES2015 (ES6). I think supporting `import` in `language-javascript` might be one of the missing pieces for supporting ES2015 FFI: https://github.com/gasi/purescript/pull/1/ Feel free to use any of my work if it helps.

purcell:
Is it the case, then, that if @erikd releases a new `language-javascript`, then bumping the dependency version here in `purescript` would be enough to get support for granular `import`s? If so, even that would be extremely useful, as it would allow external JS bundlers like Parcel to tree-shake the Javascript imports, which is currently impossible.

erikd:
@purcell When an if you want a new release of language-javascript, let me know.

hdgarrood:
No, there's still a lot more work which would need doing after a new version of language-javascript gets released to be able to ship ES modules in the PureScript compiler. It's also a big breaking change with the risk of causing a lot of pain throughout the ecosystem if executed poorly, so it's definitely not something we want to rush.

@erikd I'd suggest waiting until we have complete support for `import` and `export` before making another release of language-javascript, because filling the current gaps will require adding constructors to data types such as `JSImportDeclaration`, and of course adding constructors to an existing type which is already part of a package's public API is a breaking change, whereas providing a new type which wasn't previously part of a package's public API is only a minor-level change.

purcell:
> No, there's still a lot more work which would need doing after a new version of language-javascript gets released to be able to ship ES modules in the PureScript compiler. It's also a big breaking change with the risk of causing a lot of pain throughout the ecosystem if executed poorly, so it's definitely not something we want to rush.

Cool, thanks for clarifying, and that makes total sense. I'm not familiar with the chain of dependencies, so I'll see if I can bring myself up to speed and find a way to help out.

Cmdv:
I'm terribly sorry, I've not had the chance to get this completed. 
I feel that I've probably been holding things back now, so if anyone would like to take over please feel free. 😄

spicydonuts:
I'm really interested in this. Maybe I'll get some time next week to learn how it works.

hdgarrood:
I think probably the best way forward with this is to start by contributing more complete support for ES2015+ `export` in `language-javascript`, as currently it only has partial support; see https://github.com/erikd/language-javascript/blob/d061b16146dcddb48ff9bb8627f2399832fad855/src/Language/JavaScript/Parser/Grammar7.y#L1243-L1251

gasi:
> @hdgarrood: I think probably the best way forward with this is to start by contributing more complete support for ES2015+ `export` in `language-javascript`, as currently it only has partial support…

I don’t think that’s necessary. To move faster on this, we could decide to only support the minimal subset of `export` declarations that people would need to get stuff done in their FFI files as well as what PS would need during ES2015+ code-gen. The implemented subset should do just that:

```
export const 5;
export const foo = () => {...}
```

Beyond that, what I would propose for a gradual rollout would be:

- PSC ES2015+ code-gen should be opt-in via CLI flag or similar.
- FFI support for ES2015+ should use `.mjs` file extension, just like Node.js. This will allow PSC to choose which parse mode it should use for FFI files. IIRC, ES2015+ distinguishes between parsing a `Script` vs `Module`. A `Module` allows `import` / `export` only at the top-level.
- Support minimal, useful subset of `import` + `export` syntax for code-gen + FFI.
- Generating code with `import` + `export` instead of `require` should allow third-party tools to do proper tree shaking / dead code elimination (DCE).

hdgarrood:
> Beyond that, what I would propose for a gradual rollout would be: [...]

I also have some thoughts on this, but a PR probably isn't the best place to discuss. We have a few issues covering ES modules but they are fairly old and have a lot of discussion on them, so I think it's time to close them as well, and make a new issue for this.

hdgarrood:
> and make a new issue for this

(for the avoidance of doubt I'm in the process of doing this).

Cmdv:
cheers @hdgarrood for taking this on, just couldn't ever find the time to complete it!! 👍

PR #3430, opened 2018-09-14
compiler should do more work before giving up

justinwoo:
I would like the compiler to do more work before throwing this error now with various Symbol-related instances, but I would like to also know if this is problematic to raise from this value.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -195,7 +195,7 @@ entails SolverOptions{..} constraint context hints =
     solve con = go 0 con
       where
         go :: Int -> Constraint -> WriterT (Any, [(Ident, InstanceContext, Constraint)]) (StateT InstanceContext m) Expr
-        go work (Constraint className' tys' _) | work > 1000 = throwError . errorMessage $ PossiblyInfiniteInstance className' tys'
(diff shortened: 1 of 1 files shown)


justinwoo:
Somewhat related is #3383, which would reduce a lot of the instance matching I need in general.

kRITZCREEK:
Why is 10000 better than 1000? What time and memory difference are we talking about here?

natefaubion:
> Why is 10000 better than 1000?

Because it allows @justinwoo to parse longer strings, clearly. 😆

hdgarrood:
Usually we try to keep the number of options down as low as possible for the sake of maintenance and usability, but since this one wouldn't actually involve any branching or anything -- it's literally just replacing a constant -- I think it's probably acceptable. I agree with @natefaubion in that I'm not particularly keen on just increasing it like this.

Incidentally GHC provides this flag, calling it `-freduction-depth`: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#undecidable-instances

hdgarrood:
One problem with this option is that it might be a bit awkward if a library needs it set to a larger-than-default value to be able to compile successfully, because then every downstream project would need to start passing the same option, which might be a bit awkward in some setups. Since there doesn’t seem to be much movement here I’m going to close this for now but for anyone who comes across this, please open a new issue for discussion if there is still interest.

PR #3433, opened 2018-09-15
Avoid Data.ByteString.Lazy.toStrict

coot:
Rather small PR, which removes `toStrict` in favor of using `Data.ByteString.Lazy.UTF8` module.


kRITZCREEK:
Thanks!

PR #3434, opened 2018-09-20
Added ffiCodegen to MakeActions

coot:
There are two reasons for this PR:
* `zephyr` can use `ffiCodegen` to generate ffi code
* `purs ide` can show ffi errors (unused ffi imports)


coot:
@kritzcreek I think you will like this change :)

kRITZCREEK:
The change to the IDE code is not necessary, because `shushCodegen` only affects the _second_ rebuild, which we only use to populate completions and references with non-exported members. The build that actually collects the errors and warnings we send to the editors does generate code. (so we probably actually want to make sure we don't do any extra work on the second build)

That doesn't mean I'm opposed to this change at all ^^ Just that it's not really doing anything for the ide integration.

src/Language/PureScript/Ide/Rebuild.hs
```diff
@@ -154,11 +156,25 @@ shushProgress :: P.MakeActions P.Make -> MakeActionsEnv -> P.MakeActions P.Make
 shushProgress ma _ =
   ma { P.progress = \_ -> pure () }
 
--- | Stops any kind of codegen (also silences errors about missing or unused FFI
--- files though)
```
coot: @kRITZCREEK I was led by this comment. Anyway, the dictionary lookup should not be very expensive (which is the only overhead added to modules without ffi), `checkForeignDecls` is though, but it's also useful.  As I understand, now ffi errors will magically ;) disappear when a module is rebuild.

kRITZCREEK: The comment is misleading, yeah. It's technically not wrong, but we don't care about any errors that might be generated here anyway because we've already gotten all the errors and sent them off to the editor at this point. I'd be totally fine if you just ripped the parenthesized part out.

kRITZCREEK: The first build: https://github.com/purescript/purescript/blob/fb8daff44cce61fb404d30695065769fef582401/src/Language/PureScript/Ide/Rebuild.hs#L73-L74

doesn't `shushCodegen`.

coot:
@kRITZCREEK I re-started the ci and its green now.

kRITZCREEK:
I resolved the conflict introduced by #3433, I'll merge as soon as CI passes.

kRITZCREEK:
Thanks!

PR #3435, opened 2018-09-26
Add nix config to stack.yaml

f-f:
The missing `zlib` dependency made stack build fail on NixOS

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -2,3 +2,7 @@ resolver: lts-12.0
 packages:
 - '.'
 extra-deps:
+nix:
+  enable: false
+  packages:
+  - zlib


kRITZCREEK:
I don't have any way to test that this works and I won't notice if it ever breaks, do we have anyone using Nix who could take a look at this?

f-f:
I guess it could be possible to add a test in CI: it is enough to install `nix`, sed `nix.enable = true` and do a `stack install` (I have NixOS, but I think this is how you reproduce the bug on a non-NixOS).

OTOH, this is not breaking (as `enable: false`), and affects only `stack` users on NixOS - a pretty small set of users, who would submit a PR in case this is broken (as it's happening here). The only way this can be broken is in the case another native dependency would be missing (e.g. if we add another haskell dependency that depends on some other native library).

Note that this is also the reason why `npm install purescript` fails on NixOS (there is no prebuilt binary so it tries to use `stack` to install from source)

paluh:
I can confirm that this patch fixes PureScript compilation on NixOS.

@kRITZCREEK Is it possible to make patch release of PureScript compiler which incorporates this change and additional fixes related to https://github.com/purescript/purescript/pull/3372 so we can finally fix purs package for NixOs (https://github.com/NixOS/nixpkgs/issues/42833)?

hdgarrood:
If it is possible, I think it would make more sense to put a stack.yaml diff inside nixpkgs, and have the nix packaging tools apply that diff before building, i.e. have nixpkgs take responsibility for making sure purescript builds on nix. As @kRITZCREEK points out, none of the maintainers (as far as I am aware) use Nix, but merging this change would in effect mean taking responsibility for continuing to have the compiler build on Nix with just this stack.yaml, and I'm not sure that this is a good idea. Also, our CI is brittle and slow enough as it is, so I don't think I can get on board with adding a new job to test it on Nix (especially since the maintainers don't have the Nix knowledge to be able to debug failures easily).

f-f:
@hdgarrood adding the patch only in `nixpkgs` would not fix two issues that stem from this:
- building from source with `stack` fails on NixOS
- `npm install purescript` fails on NixOS (there is no prebuilt binary so it tries to use stack to install from source)

Some clarifications about your points:

> our CI is brittle and slow enough as it is, so I don't think I can get on board with adding a new job to test it on Nix (especially since the maintainers don't have the Nix knowledge to be able to debug failures easily)

This sounds like a CI process problem, and that should be orthogonal to the usefulness of the patch.
Moreover, the only kind of failures that can happen is `stack` complaining about "I could not compile this package because I'm missing this native library", and the fix is adding the library to the list in the `nix` section

> but merging this change would in effect mean taking responsibility for continuing to have the compiler build on Nix with just this stack.yaml

I don't think so: if the build on NixOS is going to break again then the affected users will notice and PR the fix (as it's happening here) - this is how open source works, no one is asking the maintainers to start caring about NixOS, this should effectively be a no-op patch from the maintainership point of view.

The baseline is: `stack build` on NixOS is currently broken, so any improvement on this (read: even without the promise of maintainership) is better than the status quo.

paluh:
@hdgarrood I understand that you want to keep PureScript compiler build process not any OS dependent.
On the other hand Purescript uses `stack` and `nix` section is a part of the `stack.yaml` specification if we like it or not. From that perspective currently provided by this repo `stack.yaml` is broken.

coot:
This does not work on non `NixOS` installation, but if we work out the details I can help to maintain it.  I use nix to build `zephyr` (at least locally).

The error I got when building with `stack --nix build` is:
```
Progress 8/107                
/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

/home/marcin/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

f-f:
@coot I just run `stack --nix build` on a fresh install of the latest ubuntu, and it builds correctly.

From a quick googling it looks like you might be encountering [a transient stack issue](https://github.com/commercialhaskell/stack/issues/2282#issuecomment-317055026), and e.g. running this should help: `rm -r ~/.stack/setup-exe-cache`

coot:
Thanks, it does work after removing the cache.

kRITZCREEK:
So if we get reports of problems on Nix I can ping @f-f and @coot and you'll figure things out? In that case I'm fine with merging this.

coot:
Great, thanks @kRITZCREEK

paluh:
@kRITZCREEK You can ping me as well. Thanks!

f-f:
Thank you @kRITZCREEK!

Issue #3428, opened 2018-09-09
Add `Prim.TypeError.QuoteLabel`


natefaubion:
It would be nice if there was a primitive quote/escape a symbol that is meant to be a label/property.

LiamGoodacre:
I am happy for this to exist. 👍 

To elaborate for anyone wanting to try implementing this:

Examples of row and record labels are:
1. Regular, no need to quote: `foo`  or `"foo"` or `"""foo"""`
2. Special chars, need to quote: `"f o o"` or `"""f o o"""`
3. Special chars with escaping, need to quote: `"a\"b"` or `"""a"b"""`

This ticket is about extending the `Prim.TypeError.Doc` kind with a new data type `QuoteLabel :: Symbol -> Doc`.
This should then be interpreted by both the `Prim.TypeError.Warn` and `Prim.TypeError.Fail` type classes (in the compiler), to print the quoted symbol as a label with appropriate escaping as necessary.  Note that this escaping already happens elsewhere in the code.

dariooddenino:
I'm giving a try at this. I'm mostly stumbling in the dark, but I guess I'm almost there

LiamGoodacre:
@dariooddenino thanks for taking a look!

For `Warn (QuoteLabel "ba\"oo")` I would expect `"ba\"oo"` as the output.
For `Warn (QuoteLabel "hello")` I would expect `hello` as the ouput.

There should already be a function to do the printing/escaping for you in the pretty printer.

To write your `baz { hello: 1 }` example, you could do something like:
```purescript
bar ::
  forall row label typ .
  RowToList row (Cons label typ Nil) =>
  Warn (QuoteLabel label) =>
  Record row ->
  String
bar _ = ""

foo = bar { hello: 42 }
```

Feel free to open a work-in-progress pull request and we can discuss more there?

dariooddenino:
Ugh, I was overcomplicating it. Yeah, I saw the pretty printing functions but first I wanted to clear these doubts.
I will try to open a pull request tomorrow (hopefully with everything already done!)

Thanks

PR #3436, opened 2018-10-03
Adding `Prim.TypeError.QuoteLabel`

dariooddenino:
This pull request adds `QuoteLabel` as described in #3428 
The test and doc string are obviously placeholders.

Right now it works, but I replaced `prettyPrintString` with `prettyPrintLabel` in `matchTypeAtom`'s `match` for `TypeLevelString`s and I suppose this will break other things.

I don't think I have a way to discriminate if it's working with a label or not. `matchTypeAtom` takes only `TypeRenderOption` as argument, but I'm not sure it's appropriate to add a "isLabel" parameter there.

### DIFF
--- a/src/Language/PureScript/Docs/Prim.hs
+++ b/src/Language/PureScript/Docs/Prim.hs
@@ -465,8 +465,8 @@ quoteDoc = primTypeOf (P.primSubName "TypeError") "Quote" $ T.unlines
 
 quoteLabelDoc :: Declaration
 quoteLabelDoc = primTypeOf (P.primSubName "TypeError") "QuoteLabel" $ T.unlines
-  [ "The QuoteLabel type constructor renders any Symbol as a syntactically valid label,"
-  , "escaping with quotes as needed."
+  [ "The `QuoteLabel` type constructor will produce a `Doc` when given a `Symbol`. When the resulting `Doc` is rendered "
+  , "for a `Warn` or `Fail` constraint, a syntactically valid label will be produces, escaping with quotes as needed."
   , ""
   , "For more information, see"
   , "[the Custom Type Errors guide](https://github.com/purescript/documentation/blob/master/guides/Custom-Type-Errors.md)."
--- a/src/Language/PureScript/Docs/Prim.hs
+++ b/src/Language/PureScript/Docs/Prim.hs
@@ -465,8 +465,8 @@ quoteDoc = primTypeOf (P.primSubName "TypeError") "Quote" $ T.unlines
 
 quoteLabelDoc :: Declaration
 quoteLabelDoc = primTypeOf (P.primSubName "TypeError") "QuoteLabel" $ T.unlines
-  [ "The `QuoteLabel` type constructor will produce a `Doc` when given a `Symbol`. When the resulting `Doc` is rendered "
-  , "for a `Warn` or `Fail` constraint, a syntactically valid label will be produces, escaping with quotes as needed."
+  [ "The `QuoteLabel` type constructor will produce a `Doc` when given a `Symbol`. When the resulting `Doc` is rendered"
+  , "for a `Warn` or `Fail` constraint, a syntactically valid label will be produced, escaping with quotes as needed."
   , ""
   , "For more information, see"
   , "[the Custom Type Errors guide](https://github.com/purescript/documentation/blob/master/guides/Custom-Type-Errors.md)."
--- a/src/Language/PureScript/Docs/Prim.hs
+++ b/src/Language/PureScript/Docs/Prim.hs
@@ -442,7 +442,7 @@ kindDoc = primKindOf (P.primSubName "TypeError") "Doc" $ T.unlines
   [ "`Doc` is the kind of type-level documents."
   , ""
   , "This kind is used with the `Fail` and `Warn` type clases."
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -1409,6 +1409,8 @@ toTypelevelString (TypeApp (TypeConstructor f) x)
   | f == primSubName C.typeError "Text" = toTypelevelString x
 toTypelevelString (TypeApp (TypeConstructor f) x)
   | f == primSubName C.typeError "Quote" = Just (typeAsBox x)
+toTypelevelString (TypeApp (TypeConstructor f) x)
+  | f == primSubName C.typeError "QuoteLabel" = Just (typeAsBox x)
```
natefaubion: I think for this, you don't want to use `typeAsBox`. You want to match directly on `TypelevelString` for `x`, along with `Just . line . prettyPrintLabel $ x`.

src/Language/PureScript/Pretty/Types.hs
```diff
@@ -124,7 +124,7 @@ matchTypeAtom tro@TypeRenderOptions{troSuggesting = suggesting} =
     typeLiterals = mkPattern match where
       match TypeWildcard{} = Just $ text "_"
       match (TypeVar var) = Just $ text $ T.unpack var
-      match (TypeLevelString s) = Just $ text $ T.unpack $ prettyPrintString s
+      match (TypeLevelString s) = Just $ text $ T.unpack $ prettyPrintLabel $ Label s
```
natefaubion: With the changes to `toTypelevelString`, you can revert this. This is definitely not desirable 😄

dariooddenino:
@natefaubion thanks! I fixed toTypelevelString like you suggested, wrote the docs line and moved the test to its own file.

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -467,6 +463,15 @@ quoteDoc = primTypeOf (P.primSubName "TypeError") "Quote" $ T.unlines
   , "[the Custom Type Errors guide](https://github.com/purescript/documentation/blob/master/guides/Custom-Type-Errors.md)."
   ]
 
+quoteLabel :: Declaration
+quoteLabel = primTypeOf (P.primSubName "TypeError") "QuoteLabel" $ T.unlines
+  [ "The QuoteLabel type constructor renders any label as a Doc"
+  , "to be used in a custom type error."
```
natefaubion: How about

```
The QuoteLabel type constructor renders any Symbol as a syntactically valid label, escaping with quotes as needed.
```

LiamGoodacre:
^ I like mentioning the syntactically valid label stuff.
But I'm not sure about the wording of `renders` because the constructor doesn't actually do anything.  The compiler interprets it when it has to satisfy a `Warn` or `Fail` constraint.  I've had people think `Quote`/`TypeString` works like a type family and immediately resolves to a symbol, and it feels to me like this wording suggests that this is the case.

What do you think about?
> The `QuoteLabel` type constructor will produce a `Doc` when given a `Symbol`.  When the resulting `Doc` is rendered for a `Warn` or `Fail` constraint, a syntactically valid label will be produced, escaping with quotes as needed.

dariooddenino:
Is that ok as the definitive version? :)

natefaubion:
I think the only thing is changing the doc text to Liam's suggestion.

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -462,6 +463,15 @@ quoteDoc = primTypeOf (P.primSubName "TypeError") "Quote" $ T.unlines
   , "[the Custom Type Errors guide](https://github.com/purescript/documentation/blob/master/guides/Custom-Type-Errors.md)."
   ]
 
+quoteLabelDoc :: Declaration
+quoteLabelDoc = primTypeOf (P.primSubName "TypeError") "QuoteLabel" $ T.unlines
+  [ "The `QuoteLabel` type constructor will produce a `Doc` when given a `Symbol`. When the resulting `Doc` is rendered "
+  , "for a `Warn` or `Fail` constraint, a syntactically valid label will be produces, escaping with quotes as needed."
```
LiamGoodacre: Typo: `will be produces` -> `will be produced`

dariooddenino: ugh, I even read it 4 times :/ sorry

LiamGoodacre:
I just noticed that the `kindDoc` documentation also mentions all the constructors:
```haskell
kindDoc :: Declaration
kindDoc = primKindOf (P.primSubName "TypeError") "Doc" $ T.unlines
  [ "`Doc` is the kind of type-level documents."
  , ""
  , "This kind is used with the `Fail` and `Warn` type clases."
  , "Build up a `Doc` with `Text`, `Quote`, `Beside`, and `Above`."
  ]
```
We should add `QuoteLabel` here too.

Could you also add yourself to the `CONTRIBUTORS.md` file?

I will happily merge once these two things have been updated.
Thank you for working on this!

dariooddenino:
Thank you all for taking the time to review my pr :)
I hope I can keep on contributing (and maybe with more useful stuff! :D)

LiamGoodacre:
:tada:  Thanks again @dariooddenino !

PR #3437, opened 2018-10-03
Fixes #3388

### LINKED ISSUE #3388: Invalid argument to infer: ObjectUpdateNested ...
https://gist.github.com/natefaubion/75cdb4646a568f2da802ec0da7a53e6b

I have not reduced it, but it amounts to this source, which is pretty standard
```purescript
      { x, y, width, height } =
        activePosition
          { y = minY
          , height = (activePosition.y + activePosition.height) - minY
          }
```

natefaubion:
Rearranges desugaring phases to avoid leaking `ObjectNestedUpdate`.

To be honest, I'm not sure exactly why this fixes it, but desugaring appears to be sensitive to this ordering. `ObjectNestedUpdate` is supposed to be erased by `desugarObjectConstructors`, yet somehow it was leaking all the way through to type checking when paired with `BoundValueDeclaration`. I thought that maybe `everywhereOnValuesTopDownM` might have an error, resulting in a missed case, but at a cursory glance it seemed OK. I found that desugaring cases/letpattern before object constructors works. I'd like to know what thoughts y'all have on this issue, and if you have a hunch for a better fix.


kRITZCREEK:
I took a look at this and the problem is in Traversals, line 135:

```
-  f' (BoundValueDeclaration sa b expr) = BoundValueDeclaration sa <$> h' b <*> g' expr
+  f' (BoundValueDeclaration sa b expr) = BoundValueDeclaration sa <$> (h' <=< h) b <*> (g' <=< g) expr
```

natefaubion:
Thanks @kRITZCREEK for looking at that. I've applied the fix to the traversal and reverted the ordering changes.

kRITZCREEK:
Thanks!

PR #3438, opened 2018-10-03
Deletes some suspicious code in desugarObjectConstructors

kRITZCREEK:
Just some code I found while debugging #3437


kRITZCREEK:
whoops.

PR #3439, opened 2018-10-08
Fix sharing in function composition inlining

### LINKED ISSUE #3280: Inlining function composition reduces sharing
Inlining `f <<< g` gives `\x -> f (g x)`. However, `f` might be a partial application which closes over some intermediate result. For example, `f` might be the result of compiling a complex regular expression. So inlining destroys the opportunity to share work.

natefaubion:
Fixes #3280 

Example input:
```purescript
module Compose where

import Prelude

test1 :: Int -> Int
test1 = add 42 <<< mul 5

test2 :: Int -> Int
test2 = add (1 + 3) <<< mul (10 * 10) <<< add (2 + 5)

add5 :: Int -> Int
add5 = add 5

mul3 :: Int -> Int
mul3 = mul 3

test3 :: Int -> Int
test3 = add5 <<< mul3 <<< add 10

test4 :: Int -> Int
test4 = add 42 >>> mul 5 >>> add 10
```

Output:
```js
// Generated by purs version 0.12.0
"use strict";
var Control_Semigroupoid = require("../Control.Semigroupoid/index.js");
var Data_Semiring = require("../Data.Semiring/index.js");
var Prelude = require("../Prelude/index.js");
var test4 = (function ($0, $1, $2) {
    return function ($3) {
        return $0($1($2($3)));
    };
})(Data_Semiring.add(Data_Semiring.semiringInt)(10), Data_Semiring.mul(Data_Semiring.semiringInt)(5), Data_Semiring.add(Data_Semiring.semiringInt)(42));
var test2 = (function ($4, $5, $6) {
    return function ($7) {
        return $4($5($6($7)));
    };
})(Data_Semiring.add(Data_Semiring.semiringInt)(1 + 3 | 0), Data_Semiring.mul(Data_Semiring.semiringInt)(10 * 10 | 0), Data_Semiring.add(Data_Semiring.semiringInt)(2 + 5 | 0));
var test1 = (function ($8, $9) {
    return function ($10) {
        return $8($9($10));
    };
})(Data_Semiring.add(Data_Semiring.semiringInt)(42), Data_Semiring.mul(Data_Semiring.semiringInt)(5));
var mul3 = Data_Semiring.mul(Data_Semiring.semiringInt)(3);
var add5 = Data_Semiring.add(Data_Semiring.semiringInt)(5);
var test3 = (function ($11) {
    return function ($12) {
        return add5(mul3($11($12)));
    };
})(Data_Semiring.add(Data_Semiring.semiringInt)(10));
module.exports = {
    test1: test1,
    test2: test2,
    add5: add5,
    mul3: mul3,
    test3: test3,
    test4: test4
};
```

### DIFF
--- a/src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
@@ -15,6 +15,7 @@ import Prelude.Compat
 
 import Control.Monad.Supply.Class (MonadSupply, freshName)
 
+import Data.Either (rights)
 import Data.Maybe (fromMaybe)
 import Data.String (IsString, fromString)
 import Data.Text (Text)
@@ -269,12 +270,11 @@ inlineFnComposition = everywhereTopDownM convert where
   convert other = return other
 
   mkApps :: Maybe SourceSpan -> [Either AST (Text, AST)] -> Text -> AST
-  mkApps ss fns a = App ss (Function ss Nothing vars (Block ss [Return Nothing comp])) args
+  mkApps ss fns a = App ss (Function ss Nothing [] (Block ss $ vars <> [Return Nothing comp])) []
     where
+    vars = uncurry (VariableIntroduction ss) . fmap Just <$> rights fns
     comp = Function ss Nothing [a] (Block ss [Return Nothing apps])
     apps = foldr (\fn acc -> App ss (mkApp fn) [acc]) (Var ss a) fns
-    vars = foldMap (pure . fst) =<< fns
-    args = foldMap (pure . snd) =<< fns
 
   mkApp :: Either AST (Text, AST) -> AST
   mkApp = either id $ \(name, arg) -> Var (getSourceSpan arg) name
--- a/src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
@@ -264,20 +264,37 @@ inlineFnComposition = everywhereTopDownM convert where
   convert (App s1 (App s2 (App _ (App _ fn [dict']) [x]) [y]) [z])
     | isFnCompose dict' fn = return $ App s1 x [App s2 y [z]]
     | isFnComposeFlipped dict' fn = return $ App s2 y [App s1 x [z]]
-  convert (App ss (App _ (App _ fn [dict']) [x]) [y])
-    | isFnCompose dict' fn = do
-        arg <- freshName
-        return $ Function ss Nothing [arg] (Block ss [Return Nothing $ App Nothing x [App Nothing y [Var Nothing arg]]])
-    | isFnComposeFlipped dict' fn = do
-        arg <- freshName
-        return $ Function ss Nothing [arg] (Block ss [Return Nothing $ App Nothing y [App Nothing x [Var Nothing arg]]])
+  convert app@(App ss (App _ (App _ fn [dict']) _) _)
+    | isFnCompose dict' fn || isFnComposeFlipped dict' fn = mkApps ss <$> goApps app <*> freshName
   convert other = return other
+
+  mkApps :: Maybe SourceSpan -> [Either AST (Text, AST)] -> Text -> AST
+  mkApps ss fns a = App ss (Function ss Nothing vars (Block ss [Return Nothing comp])) args
+    where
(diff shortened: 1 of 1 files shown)


natefaubion:
Do we already have tests for inlined code like this?

natefaubion:
I suppose I could just use var bindings as well.

natefaubion:
Output:
```js
// Generated by purs version 0.12.0
"use strict";
var Control_Semigroupoid = require("../Control.Semigroupoid/index.js");
var Data_Semiring = require("../Data.Semiring/index.js");
var Prelude = require("../Prelude/index.js");
var test4 = (function () {
    var $0 = Data_Semiring.add(Data_Semiring.semiringInt)(10);
    var $1 = Data_Semiring.mul(Data_Semiring.semiringInt)(5);
    var $2 = Data_Semiring.add(Data_Semiring.semiringInt)(42);
    return function ($3) {
        return $0($1($2($3)));
    };
})();
var test2 = (function () {
    var $4 = Data_Semiring.add(Data_Semiring.semiringInt)(1 + 3 | 0);
    var $5 = Data_Semiring.mul(Data_Semiring.semiringInt)(10 * 10 | 0);
    var $6 = Data_Semiring.add(Data_Semiring.semiringInt)(2 + 5 | 0);
    return function ($7) {
        return $4($5($6($7)));
    };
})();
var test1 = (function () {
    var $8 = Data_Semiring.add(Data_Semiring.semiringInt)(42);
    var $9 = Data_Semiring.mul(Data_Semiring.semiringInt)(5);
    return function ($10) {
        return $8($9($10));
    };
})();
var mul3 = Data_Semiring.mul(Data_Semiring.semiringInt)(3);
var add5 = Data_Semiring.add(Data_Semiring.semiringInt)(5);
var test3 = (function () {
    var $11 = Data_Semiring.add(Data_Semiring.semiringInt)(10);
    return function ($12) {
        return add5(mul3($11($12)));
    };
})();
module.exports = {
    test1: test1,
    test2: test2,
    add5: add5,
    mul3: mul3,
    test3: test3,
    test4: test4
};
```

garyb:
Does that get caught in the compiler, or does it just loop at runtime?

I know the current behaviour isn't good, but I'd personally find that quite a high price if it's a runtime breaking thing.

natefaubion:
It does not get caught by the compiler because it's under an abstraction. This code without the optimization would break without warning, unfortunately. @felixSchl discovered this when porting code to pure-c. `purescript-enums` was broken until he implemented an identical "optimization" to the current version.

natefaubion:
Right now any backend based on corefn will break on this code if it doesn't implement the exact optimization.

natefaubion:
Are we going to defer this? It's unfortunate that this is "breaking", even though it is a bug fix. Does anyone have any ideas on how to mitigate breakage?

natefaubion:
>  I don't know how to recognise whatever it is that is happening to make that loop.

It's the same thing as monadic recursion. This will likely blow the stack because of strictness:
```purescript
foo n = bar *> when (n < 100000) (foo (n + 1))
```
But this is ok:
```purescript
foo n = bar >>= \_ -> when (n < 100000) (foo (n + 1))
```

The current "optimization" is essentially transforming the former into the latter, but for function composition, which violates evaluation semantics.

src/Language/PureScript/CoreImp/Optimizer/Inliner.hs
```diff
@@ -264,20 +265,36 @@ inlineFnComposition = everywhereTopDownM convert where
...
+  mkApps ss fns a = App ss (Function ss Nothing [] (Block ss $ vars <> [Return Nothing comp])) []
+    where
+    vars = uncurry (VariableIntroduction ss) . fmap Just <$> rights fns
+    comp = Function ss Nothing [a] (Block ss [Return Nothing apps])
+    apps = foldr (\fn acc -> App ss (mkApp fn) [acc]) (Var ss a) fns
+
+  mkApp :: Either AST (Text, AST) -> AST
+  mkApp = either id $ \(name, arg) -> Var (getSourceSpan arg) name
+
+  goApps :: AST -> m [Either AST (Text, AST)]
```
hdgarrood: Do you think it would be worth having a small comment to explain what Left and Right values represent here?

hdgarrood:
Actually, one way to mitigate breakage would be to actively look for issues like this across libraries and start fixing them now. I suppose alternate backend authors will have already come across a few themselves?

PR #3440, opened 2018-10-08
Add source spans to desugared do blocks

### LINKED ISSUE #3418: "Unknown value discard" error has no file information
```
module Main where

import Effect.Console (log)

main = do
  log "What"  
  log "Hello sailor!"
```

```
Error found:
in module Main
at  line 0, column 0 - line 0, column 0

  Unknown value discard. You're probably using do-notation, which the compiler replaces with calls to the discard function. Please import discard from module Prelude


```

This yields an error record where the file name is an empty string.

natefaubion:
Fixes #3418 

This adds position information to desugared do code, where it was previously using `nullSourceSpan` before.

The example now yields this error:

```
[1/1 UnknownName] /Users/nathan/Sites/git/purescript/BadSpan.purs:6:3

  6    log "What"
       ^^^^^^^^^^

  Unknown value discard. You're probably using do-notation, which the compiler replaces with calls to the discard function. Please import discard from module Prelude
```

### DIFF
--- a/src/Language/PureScript/Sugar/DoNotation.hs
+++ b/src/Language/PureScript/Sugar/DoNotation.hs
@@ -25,46 +25,47 @@ desugarDoModule (Module ss coms mn ds exts) = Module ss coms mn <$> parU ds desu
 -- | Desugar a single do statement
 desugarDo :: forall m. (MonadSupply m, MonadError MultipleErrors m) => Declaration -> m Declaration
 desugarDo d =
-  let (f, _, _) = everywhereOnValuesM return replace return
-  in rethrowWithPosition (declSourceSpan d) $ f d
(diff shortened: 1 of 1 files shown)


garyb:
❤️ 

Can we throw the errors with positions directly in here instead of relying on `errorMessage` and then rethrowing perhaps? I'm fine with this too, but ideally I'd like to remove the non-positioned-`errorMessage` entirely one day.

natefaubion:
The error is not actually thrown here. The error is thrown in the name desugaring pass which adds full qualification everywhere. I did not see a straightforward way of rethrowing in that code, and adding position information to desugared code seemed like a win regardless.

natefaubion:
@garyb is there anything else to do to this?

Issue #3417, opened 2018-08-23
A case of ScopeShadowing has no real position


garyb:
```
Warning 1 of 3:

  in module SlamX.Page.Builder.Component
  at <module> line 0, column 0 - line 0, column 0

    Shadowed definitions are in scope for value apply from the following open imports:

      import SlamX.Prelude

    These will be ignored and the local declaration will be used.


  See https://github.com/purescript/documentation/blob/master/errors/ScopeShadowing.md for more information,
  or to contribute content related to this warning.

Warning 2 of 3:

  in module SlamX.Page.Builder.Component
  at <module> line 0, column 0 - line 0, column 0

    Shadowed definitions are in scope for value apply from the following open imports:

      import SlamX.Prelude

    These will be ignored and the local declaration will be used.


  See https://github.com/purescript/documentation/blob/master/errors/ScopeShadowing.md for more information,
  or to contribute content related to this warning.

Warning 3 of 3:

  in module SlamX.Page.Builder.Component
  at <module> line 0, column 0 - line 0, column 0

    Shadowed definitions are in scope for value apply from the following open imports:

      import SlamX.Prelude

    These will be ignored and the local declaration will be used.


  See https://github.com/purescript/documentation/blob/master/errors/ScopeShadowing.md for more information,
  or to contribute content related to this warning.
```

nwolverson:
See linked issue, here's a short repro:

```purescript
module Main where

import Prelude

data Array = Array

x :: Array -> Unit
x = pure unit
```

As noted there the filename of the error is bogus as well as bad source spans

natefaubion:
This should be fixed.

PR #3441, opened 2018-10-08
Add a stricter source span check for warnings

natefaubion:
Obviously now I have to fix the failing tests.

### DIFF
--- a/src/Language/PureScript/Sugar/Names/Env.hs
+++ b/src/Language/PureScript/Sugar/Names/Env.hs
@@ -32,6 +32,7 @@ import Data.Function (on)
 import Data.Foldable (find)
 import Data.List (groupBy, sortBy, delete)
 import Data.Maybe (fromJust, mapMaybe)
+import Safe (headMay)
 import qualified Data.Map as M
 import qualified Data.Set as S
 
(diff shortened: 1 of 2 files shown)


natefaubion:
@garyb Any idea how

https://github.com/purescript/purescript/blob/c89e2f6d96879a91618d2c783c8f677d6bec5661/src/Language/PureScript/Sugar/Names/Imports.hs#L53

propagates all the way to

https://github.com/purescript/purescript/blob/c89e2f6d96879a91618d2c783c8f677d6bec5661/src/Language/PureScript/Sugar/Names/Env.hs#L474

natefaubion:
Also, for my own curiosity, why are we creating such a nonsense span in the first place?

natefaubion:
It seems like it was using the wrong source span. I now get something like this:

```
[1/1 ScopeShadowing] /Users/nathan/Sites/git/purescript/BadShadow.purs:3:1

  3  import Prelude
     ^^^^^^^^^^^^^^

  Shadowed definitions are in scope for value append from the following open imports:

    import Prelude

  These will be ignored and the local declaration will be used.
```

natefaubion:
If that isn't correct, and the error should point to the local definition rather than the import, then that means we would need to lookup the source span of the appropriate declaration, rather than use the current module source span (which is bogus).

natefaubion:
This should fix #3417

natefaubion:
I can confirm that's what happens, which I don't really care for. Do you have a suggestion for how to get a hold of the source span for the particular declaration in question?

garyb:
Not that I can think of, no. I think the newly positioned error is still pretty good though actually - you're warned that the version coming from the open import is unused, so the span being there rather than on the member that is actually resolved is perhaps better even?

Either way, I'll merge this, and we can open an issue to improve the error further if we think that's worth doing.

PR #3442, opened 2018-10-10
Consider type operators when reexporting a pseudo-module

### LINKED ISSUE #3410: Compiler error "Cannot export unknown module"
have noticed the following behaviour
```purescript
module CustomPrelude
  ( module Prelude
  , module DEN -- Compiler error "Cannot export unknown module DEN"
  ) where

import Prelude
import Data.Either.Nested (type (\/)) as DEN
```

but the following is fine:
```purescript
module CustomPrelude
  ( module Prelude
  , module DEN
  ) where

import Prelude
import Data.Either.Nested (type (\/), in1) as DEN
```

and in general - if only `type (\/)` is imported then module `DEN` is considered "unknown", whenever I import something else from this module - its known.

kritzcreek [14:03]
I think you'll also need to export the type `\/` is aliasing
Might be considered a compiler bug, but I think it could also just be a bad error message (edited)

unisay [14:04]
I also thought so, but then why is it ok if I export not the type that is aliased but any other thing?

kritzcreek [14:04]
Did you try using the re-exported `\/` ?
I have a feeling it might just not get propagated

natefaubion:
Fixes #3410


garyb:
Danke! Dunno what was up with CI.

PR #3449, opened 2018-11-01
[purs ide] Accepts codegen targets for the rebuild command

### LINKED ISSUE #3447: Add codegen targets to `purs ide server`
@kRITZCREEK already suggested that this option should __probably__ be handled on the protocol level by `Rebuild` command.

---
I'm pasting here also some hacking hints by @kRITZCREEK. I think that these suggestions can be useful to anybody who wants to work on this:

The actual change that needs to happen is that `defaultOptions` in here: https://github.com/purescript/purescript/blob/2bf6ba192b368ea3aa895abb02a5c484cb4e0c27/src/Language/PureScript/Ide/Rebuild.hs#L72
needs to have its `optionsCodegenTargets` field overwritten with user specified values.
For example to force `coreFn` generation we can change this line to:
``` haskell
. P.runMake (P.defaultOptions {P.optionsCodegenTargets = Set.fromList [P.JS, P.CoreFn]})
```

kRITZCREEK:
Fixes #3447

### DIFF
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -135,22 +135,15 @@ codegenTargets = Opts.option targetParser $
       <> " The default target is 'js', but if this option is used only the targets specified will be used."
       )
 
-targets :: M.Map String P.CodegenTarget
-targets = M.fromList
-  [ ("js", P.JS)
-  , ("sourcemaps", P.JSSourceMap)
-  , ("corefn", P.CoreFn)
-  ]
-
 targetsMessage :: String
-targetsMessage = "Accepted codegen targets are '" <> intercalate "', '" (M.keys targets) <> "'."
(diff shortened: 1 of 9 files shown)


kRITZCREEK:
@paluh Could you check if this works for you? If you need me to patch `pscid` for testing I can look into that. (Maybe it's easier for you to tweak your local VIM plugin?)

paluh:
@kRITZCREEK It works like a charm!
I've tested it against this simple branch of [psc-ide-vim](https://github.com/paluh/psc-ide-vim/commit/e92b0e666e7548718d2556ce8e916496495030ea).
I've checked also purescript master branch and it seems that these additional field (`codegen`) is ignored by current version of `purs ide server`. It seems that we can add this option to `psc-ide-vim` even now with a note that it will work with `0.12.1`. Am I right?

kRITZCREEK:
Yes, that was the plan :) Thanks for confirming!

PR #3460, opened 2018-11-26
Link to documentation repo as docs for non-Prim built-in types/kinds

### LINKED ISSUE #3459: Add `kind Row` to Prim's Docs
The best reference we have to a definition of `Row` is `Record` which is inaccurate because Rows are used in so many other places besides Records (e.g. Variant, type-level programming). I'd like to submit a PR that adds docs for it, but before I do, I want to discuss the best rendering for it. Here's my current rendering for `Row`:

> `Row` is a kind for 0 or more label-type associations where the total number of these associations is known at compile time. A label can be used to refer to an instance of the associated type.
>
> First, a row can be empty (there are no associations) or non-empty (there is at least one association). Second, it can be closed (the row contains only the label-type associations defined here) or open (the row contains at least, though it may contain more than, the label-type associations defined here). Third, it can contain duplicate 'labels'.
>
> Examples:
> Empty closed Row: `()`
> Empty open Row: `( | otherLabelTypeAssociations)`
> Non-empty closed Row: `(label1 :: TypeName1, ... , labelN :: TypeNameN)`
> Non-empty open Row: `(label1 :: TypeName1, ... , labelN :: TypeNameN | otherLabelTypeAssociations)`
> Duplicate Labels: `(someLabel :: String, someLabel :: Int)`

One open question I have. While rows can contain duplicate labels, can those labels associate different types? I'm assuming yes, but I'd like it clarified.

JordanMartinez:
Fixes #3459

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -127,6 +127,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@rgrinberg](https://github.com/rgrinberg) | Rudi Grinberg | [MIT license](http://opensource.org/licenses/MIT) |
 | [@gabejohnson](https://github.com/gabejohnson) | Gabe Johnson | [MIT license](http://opensource.org/licenses/MIT) |
 | [@dariooddenino](https://github.com/dariooddenino) | Dario Oddenino | [MIT license](http://opensource.org/licenses/MIT) |
+| [@jordanmartinez](https://github.com/jordanmartinez) | Jordan Martinez | [MIT license](http://opensource.org/licenses/MIT) |
 
 ### Contributors using Modified Terms
 
--- a/src/Language/PureScript/Docs/Prim.hs
+++ b/src/Language/PureScript/Docs/Prim.hs
@@ -31,7 +31,7 @@ primDocsModule = Module
   , modComments = Just $ T.unlines
       [ "The `Prim` module is embedded in the PureScript compiler in order to provide compiler support for certain types &mdash; for example, value literals, or syntax sugar. It is implicitly imported unqualified in every module except those that list it as a qualified import."
       , ""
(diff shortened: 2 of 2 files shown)


hdgarrood:
This is pretty close to being mergeable now imo, but one last nitpick: I think we should have

> For example, row kinds (e.g. `# Type`, which is the kind of types such as `(name :: String, age :: Int)`)

instead of

> For example, the row kind (i.e. `# Type`, which is the kind of types such as `(name :: String, age :: Int)`)

since I think the latter, by using the article *the* together with *i.e.* (= *that is*), implies that `# Type` is the unique row kind, when `# Type` is just one of many kinds involving `#`.

JordanMartinez:
Nitpicks are fine, especially on docs. I'll fix it.

JordanMartinez:
I believe you asked someone in another PR to add their names to the list of contributors. Will that be required of me as well before this gets merged?

hdgarrood:
Yes please :)

JordanMartinez:
Pushed. Just need to wait on CI then.

hdgarrood:
Sorry, I should have been a little clearer: I think it would be best to replace the *i.e.* with an *e.g.* in addition to pluralising 'row kinds', since I would argue that (even though you've already pluralised 'row kinds') the *i.e.* still implies that `# Type` could be the only row kind.

JordanMartinez:
Oh. No, that's my bad for misreading your feedback. I'll push another commit with that change.

hdgarrood:
Yeah, that is correct. For any kind `k`, there is also a kind `# k`, which is the kind of rows with labels of kind `k`. So for example, you could have

```
( greeting :: "Hello, World", favouriteColor :: "Blue" )
```

which has kind `# Symbol`, or even

```
( foo :: ( bar :: "Baz" ) )
```

which has kind `# (# Symbol)`. You can verify the kinds of these types in the repl with `:k` (although there does seem to be a bug where the parentheses aren't being printed properly in kinds like `# (# Symbol)` at the moment).

LiamGoodacre:
Thanks!

PR #3461, opened 2018-12-01
Avoid dropping super class dicts for the same class

### LINKED ISSUE #3431: Superclass does not be inherited if same constraint was applied to different type parameter
minimal repro code is here:

```purescript
class (Functor f, Functor g) <= MyClass f g

fff :: forall f g. MyClass f g => f ~> f
fff = map identity -- Err, No type class instance was found for Functor f

ggg :: forall f g. MyClass f g => g ~> g
ggg = map identity -- ok
```

```purescript
-- swap definition order f <-> g
class (Functor g, Functor f) <= MyClass f g

fff :: forall f g. MyClass f g => f ~> f
fff = map identity -- ok

ggg :: forall f g. MyClass f g => g ~> g
ggg = map identity -- Err, No type class instance was found for Functor g
```

and other examples:

```purescript
-- g is not `Apply`
class (Apply f, Functor g) <= MyClass f g

ffff :: forall f g a b. MyClass f g => f (a -> b) -> f a -> f b
ffff = apply -- ok
```

```purescript
-- g implies `Apply`
class (Apply f, Applicative g) <= MyClass f g

ffff :: forall f g a b. MyClass f g => f (a -> b) -> f a -> f b
ffff = apply -- Err, No type class instance was found for Apply f
```

purs 0.12.0 on macOS High Sierra 10.13.6 and Windows 10 Pro 1803

LiamGoodacre:
Fixes #3431

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -793,7 +793,7 @@ data Expr
   -- instance type, and the type class dictionaries in scope.
   --
   | TypeClassDictionary Constraint
-                        (M.Map (Maybe ModuleName) (M.Map (Qualified (ProperName 'ClassName)) (M.Map (Qualified Ident) NamedDict)))
+                        (M.Map (Maybe ModuleName) (M.Map (Qualified (ProperName 'ClassName)) (M.Map (Qualified Ident) (NEL.NonEmpty NamedDict))))
                         [ErrorMessageHint]
   -- |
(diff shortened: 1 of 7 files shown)


LiamGoodacre:
Previously, if two super classes where for the same type class we would prefer the first on we saw.  This happened via all the `unionWith (unionWith union)` calls to combine instance contexts.  Now we collect them all in a non empty list and do `unionWith (unionWith (unionWith (<>))`.

PR #3462, opened 2018-12-01
Fix source spans for binding groups

### LINKED ISSUE #3452: Some errors in binding groups point to the wrong definition
```purescript
module BadUnknownTypeVariable where

import Prelude

test :: Int
test = go 10
  where
  ok :: Int -> Boolean
  ok = (_ == 0)

  go :: int -> Int
  go n | ok n = n
       | otherwise = go (n - 1)
```
```
[1/1 UndefinedTypeVariable] BadUnknownTypeVariable.purs:8:3

  8    ok :: Int -> Boolean
       ^^^^^^^^^^^^^^^^^^^^

  Type variable int is undefined.

  while checking the kind of int -> Int
  while checking that expression go 10
                                   where
                                   ok = \$0 ->
                                          (...) 0
                                   go = \n ->
                                          ...
    has type Int
  in value declaration test
```

Note the type `int` instead of `Int` in the type of `go`. This appears to be related to recursive binding groups. If I don't call `go` recursively, the error is correct.

```purescript
module BadUnknownTypeVariable where

import Prelude

test :: Int
test = go 10
  where
  ok :: Int -> Boolean
  ok = (_ == 0)

  go :: int -> Int
  go n | ok n = n
       | otherwise = 0
```
```
[1/1 UndefinedTypeVariable] BadUnknownTypeVariable.purs:11:3

  11    go :: int -> Int
        ^^^^^^^^^^^^^^^^

  Type variable int is undefined.

  while checking the kind of int -> Int
  while checking that expression go 10
                                   where
                                   ok = \$0 ->
                                          (...) 0
                                   go = \n ->
                                          case n of
                                            ...
    has type Int
  in value declaration test
```

I think this might(?) be because of #3278 which just rethrows with _all_ of the `BindingGroupDeclaration` spans in the hint. I think it's probably just printing the first source span in that set of hints. However in that case I'm confused why `go` and `ok` would be in the _same_ recursive binding group, since only `go` is recursive. It's definitely not like that in the corefn, which is the only thing I've verified.

LiamGoodacre:
Moves the recursive call out of each previous item's source span.

Fixes #3452 

It took me far too long to work out what was going on here :sweat_smile:.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -430,24 +430,24 @@ inferLetBinding
   -> (Expr -> m Expr)
   -> m ([Declaration], Expr)
 inferLetBinding seen [] ret j = (,) seen <$> withBindingGroupVisible (j ret)
-inferLetBinding seen (ValueDecl sa@(ss, _) ident nameKind [] [MkUnguarded tv@(TypedValue checkType val ty)] : rest) ret j =
-  warnAndRethrowWithPositionTC ss $ do
-    Just moduleName <- checkCurrentModule <$> get
+inferLetBinding seen (ValueDecl sa@(ss, _) ident nameKind [] [MkUnguarded tv@(TypedValue checkType val ty)] : rest) ret j = do
+  Just moduleName <- checkCurrentModule <$> get
+  TypedValue _ val' ty'' <- warnAndRethrowWithPositionTC ss $ do
     (kind, args) <- kindOfWithScopedVars ty
     checkTypeKind ty kind
     let dict = M.singleton (Qualified Nothing ident) (ty, nameKind, Undefined)
     ty' <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< replaceTypeWildcards $ ty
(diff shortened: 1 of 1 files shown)


natefaubion:
```
             v
   6  test = go 10
   7    where
   8    ok :: Int -> Boolean
      ...
  11    go :: int -> Int
  12    go n | ok n = n
  13         | otherwise = go (n - 1)
                                   ^

  Type variable int is undefined.

  while checking the kind of int -> Int
  while checking that expression go 10
                                   where
                                   ok = \$0 ->
                                          (...) 0
                                   go = \n ->
                                          ...
    has type Int
  in value declaration test
```
This now points to the correct definition, but the source span does not include the offending item. Is it possible to include the type signature in the span, or do we just drop all span relating to type signature?

LiamGoodacre:
@natefaubion I haven't checked yet but I suspect that the combination of this PR and #3463 will narrow the error down to the `go` declaration.

LiamGoodacre:
@natefaubion yeah that sounds good.
I've just rebased against master.
Will need an approved review to merge.

PR #3463, opened 2018-12-01
Display type class members and let identifiers in errors

LiamGoodacre:
Example code:
```purescript
class Abc a where
  abc :: a -> a

instance abcInt :: Abc Int where
  abc a = let bac :: Int -> Int
              bac = 3 in bac a
```

Error diff:
```diff
    Could not match type
      Int
    with type
      Int -> Int

  while checking that type Int
    is at least as general as type Int -> Int
  while checking that expression 3
    has type Int -> Int
+ in value declaration bac
+ in value declaration abc
  in value declaration abcInt
```

I think this could be sufficient for the second half of #3390.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -452,6 +452,8 @@ unwrapTypeDeclaration td = (tydeclIdent td, tydeclType td)
 -- In this example @double@ is the identifier, @x@ is a binder and @x + x@ is the expression.
 data ValueDeclarationData a = ValueDeclarationData
   { valdeclSourceAnn :: !SourceAnn
+  , valdeclPublicName :: !(Maybe Ident)
+  -- ^ The declared value's name - renders instead of valdeclIdent
   , valdeclIdent :: !Ident
   -- ^ The declared value's name
   , valdeclName :: !NameKind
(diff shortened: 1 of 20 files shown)


LiamGoodacre:
Okay I broke something, this isn't quite ready yet.

LiamGoodacre:
Feels a bit hacky too.  Will re-think.

hdgarrood:
It's perhaps worth noting that the `--verbose-errors` flag already gets you something somewhat similar (if a bit more, uh, verbose):

```
  Could not match type
       
    Int
       
  with type
              
    Int -> Int
              

  with type Int -> Int
while checking that type Int
  is at least as general as type Int -> Int
while checking that expression 3
  has type Int -> Int
                                 bac = 3
                               in bac a 
  has type Int
while checking that expression \a ->      
                                 let      
                                   bac = 3
                                 in bac a 
  has type Int -> Int
while checking that expression Abc { abc: \a ->      
                                            let      
                                              bac = 3
                                            in bac a 
                                   }                 
  has type { abc :: Int -> Int
           }                  
in value declaration abcInt
```

hdgarrood:
I'm not sure how I feel about this approach actually. I think I'd prefer we leave this as is and improve error spans / implement printing actual code in errors to address the issue of it being difficult to work out which bit of your code is causing the problem.

hdgarrood:
Shall we close this (and #3390)?

PR #3471, opened 2018-12-03
PSCi: Support multiple command types in paste-mode

### LINKED ISSUE #3454: PSCi :paste cannot parse anything after an import
Any expressions or statements that follow the `import` cause a parse error. This means multiple imports have to be pasted one-at-a-time.

```
> :paste
… import Prelude
… import Data.Array
…
(line 2, column 1):
unexpected "import"
expecting "hiding", (, "as" or end of input
```

It seems like any expression or statement following the import causes the same error:
```
> :paste
… import Data.Array
…
… 3 + 4
…
(line 3, column 1):
unexpected 3
expecting "hiding", (, "as" or end of input
```

LiamGoodacre:
Fixes #3454 

E.g. you can paste an import and a declaration together:

Before
```
> :paste
… import Prelude
… foo = true
… ^D
(line 2, column 1):
unexpected "foo"
expecting "hiding", (, "as" or end of input
>
```

After
```
> :paste
… import Prelude
… foo = true
… ^D
>
```


src/Language/PureScript/Interactive/Parser.hs
```diff
@@ -33,16 +33,16 @@ parseDotFile filePath s = first show $ do
...
 --
-parseCommand :: String -> Either String Command
+parseCommand :: String -> Either String [Command]
 parseCommand cmdString =
   case cmdString of
     (':' : cmd) -> parseDirective cmd
     _ -> parseRest psciCommand cmdString
 
-parseRest :: P.TokenParser a -> String -> Either String a
+parseRest :: P.TokenParser a -> String -> Either String [a]
```
hdgarrood: This PR all looks great but I’m just slightly concerned about this change; I’m not sure it makes sense with all of the use sites. In particular should the places it’s used in parseDirective be changing? Unless I’m missing something, `:browse Prelude Control.Bind` will now queue two Browse commands, which seems a little odd to me.

LiamGoodacre: Ah yes, this was unintentional.  Thanks!

PR #3477, opened 2018-12-12
Record type error

### LINKED ISSUE #3392: Diff row/record type errors
(I thought there was an issue for this but I couldn't find it)

The "Could not match type ... with type ..." error is really difficult to read when the types in question are rows/records. For example:

```purescript
  Could not match type

    ( onChange :: EffectFn1 t5 Unit
    , options :: Array t4
    , searchable :: t2
    , style :: t1
    , value :: t3
    , className :: String
    , filterOptions :: Array
                         { value :: String
                         , label :: String
                         | additionalData0
                         }
                       -> Array
                            { value :: String
                            , label :: String
                            | additionalData0
                            }
    , id :: String
    , name :: String
    , noResultsText :: String
    , optionRenderer :: { value :: String
                        , label :: String
                        | additionalData0
                        }
                        -> JSX
    , placeholder :: String
    )

  with type

    ( noBorder :: Boolean
    , onChange :: EffectFn1
                    (Nullable
                       (Array
                          { value :: String
                          , label :: String
                          | additionalData0
                          }
                       )
                    )
                    Unit
    , options :: Array
                   { value :: String
                   , label :: String
                   | additionalData0
                   }
    , searchable :: Boolean
    , style :: CSS
    , value :: Array String
    , className :: String
    , filterOptions :: Array
                         { value :: String
                         , label :: String
                         | additionalData0
                         }
                       -> Array
                            { value :: String
                            , label :: String
                            | additionalData0
                            }
    , id :: String
    , name :: String
    , noResultsText :: String
    , optionRenderer :: { value :: String
                        , label :: String
                        | additionalData0
                        }
                        -> JSX
    , placeholder :: String
    )
```

The actual error here is that the first row is missing the label `noBorder :: Boolean`. It'd be great to show that diff at the beginning of the error message.

The error message is also reported in a strange place:

```purescript
component :: ReactComponent { ... }
component = ...
  where
    initialState = ...                      -- the error is reported here
    
    ...

    render props state setState =
      ...
      where
        helper =
          ...
            (props.render                   -- the error should be reported here
              { className: props.className
              , style: props.style
              , searchable: props.searchable
              , id: props.id
              , name: props.name
              , noBorder: props.noBorder
              , noResultsText: props.noResultsText
              , placeholder: props.placeholder
              , optionRenderer: props.optionRenderer
              , filterOptions: props.filterOptions
              , value: props.value
              , options: []
              , onChange: mkEffectFn1 \_ -> pure unit
              })
```

dariooddenino:
Some tentative code to fix #3392


natefaubion:
I think the approach looks good. Can anyone think of a reason to _not_ show the common labels in _some_ form? I personally can't think of any reason to show them, but I just want to be sure there isn't something I'm not thinking of.

dariooddenino:
@natefaubion 
There's a problem with this approach. In both these cases:
**(a)** `[{ a: 1, b: 2 }, { a: 1 }]` and  **(b)** `[{ a: 1}, {}]`
the second record unique labels are empty rows.

This means that they are both rendered as
```
(
...
)
```
or
```()```
depending on how I change the code, while **(a)** should be rendered as in the first example and **(b)** as in the second example.

**Comment by natefaubion** [comment]
@dariooddenino Sorry I broke your PR. I added an extra annotation parameter to `Type`, and `rowToList` and `rowFromList` now use `RowListItem a` instead of a tuple.

**Comment by dariooddenino** [comment]
@natefaubion 
While fixing the PR I noticed that there's another case I'm not covering and could probably benefit from showing the row diff only:
```
bar = [{a: 1, b: 2}, {a: true, b: 2}]

Could not match type

        Int

      with type

        Boolean


    while trying to match type ( a :: Int
                               , b :: Int
                               )
      with type ( a :: Boolean
                , b :: Int
                )
```

Right now the row diff is only shown in the `Could not match type ... with type ...` and not in the `while trying to match type ... with type ...`.
Should I change it there too?

**Comment by garyb** [comment]
It'd be nice to show the diff there - or perhaps mention which label the unification failed for in the first part of the error, but I think it's perhaps easier to change the `while trying to` part instead. I suppose the other advantage of it is that it cuts down the size of the error message in general, keeping it more focused, especially when there's a large row mismatch.

**Comment by dariooddenino** [comment]
I can change the `while trying to` part for sure. Maybe showing the label in the first part would be better suited for a new issue/pr?

**Comment by dariooddenino** [comment]
So, I have progressed (I was a little busy), but I have a couple of questions:

1) I've found inconsistencies regarding the type passed to `renderHint`:

`[{a: 1}, {a: "b"}]` -> The types are `RCons` and it works correctly.

`[{a: {b: 1}}, {a: {b: "b"}}]` -> As above.

`[{a: 1}, {b: 2}]` -> The types are `TUknown` and `TypeApp`. In the `while trying to` message, the first type is shown as `t0` and the second as a record (not a row).

`[{a: {b: 1}}, {a: {c: 1}}]` -> Both types are `TypeApp` and are shown as a record in the message.

Before spending time on this I'd like to know if this is a desired/unfixable behaviour :)

2) Do I have a way to test this apart from manually checking error messages? 

Thanks guys :)

**Comment by dariooddenino** [comment]
Done! I've fixed case 3 and 4 from the previous message too :)
The only downside is that now the error and the hint are basically the same thing in those two cases, eg:

```
  Could not match type                                                                                                             
                                                                                                                              
        ( b :: Int                                                                                                            
        ...                                                                                                                   
        )                                                                                                                     
                                                                                                                              
      with type                                                                                                                        
                                                                                                                              
        ( c :: Int                                                                                                            
        ...                                                                                                                   
        )                                                                                                                     
                                                                                                                              
                                                                                                                                       
    while trying to match type                                                                                                
                                 ( b :: Int                                                                                   
                                 ...                                                                                          
                                 )                                                                                            
                                                                                                                              
      with type                                                                                                               
                  ( c :: Int                                                                                                  
                  ...                                                                                                         
                  )    
```
Which I think is still better than showing an unknown type + full record or two full records as it was before.

**Comment by dariooddenino** [comment]
Sorry, I made a mess rebasing stuff :D
I guess I will nuke everything, clean things up and open a new pr :sob:

**Comment by dariooddenino** [comment]
It should be cleaned up now. I really need to improve my rebase game :/

Anyways, it seems like one of master commits changed the way rows are printed by adding new lines in between. 

```
( a :: Int
, b :: Int
)
```

became
```
( a :: Int

, b :: Int

)
```

Was this intentional?

**Comment by hdgarrood** [comment]
> Anyways, it seems like one of master commits changed the way rows are printed by adding new lines in between.

Are you sure? This is what I get on `master`:

```
[1/1 TypesDoNotUnify] src/Main.purs:9:7

  9  bar = [{a: 1, b: 2}, {a: true, b: 2}]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
  Could not match type
  
    Int
  
  with type
  
    Boolean
  
  while trying to match type ( a :: Int
                             , b :: Int
                             )
    with type ( a :: Boolean
              , b :: Int
              )
  while inferring the type of [ { a: 1
                                , b: 2
                                }
                              , { a: true
                                , b: 2
                                }
                              ]
```
Is it possible that what you're seeing is due to line wrapping in your terminal?

dariooddenino:
@hdgarrood umm you're right! Something weird was happening with my terminal that day, but it looks fine now. Sorry!

hdgarrood:
> Can anyone think of a reason to not show the common labels in some form?

Occasionally, I find it useful to see the common labels to help identify which type is actually occurring in the error. For example, suppose I am writing code which makes use of some record type synonyms `User`, `Invoice`, and `Document`, all of which have a `title` field, and I have a type mismatch in that field. I think it's useful to see the other fields in the type error in order to be able to more easily work out which type the error is arising from (i.e. is it arising from a `User`, an `Invoice`, or a `Document`, or something else)? This is particularly important in cases where the source span is larger than it ought to be, which does occasionally happen in e.g. do notation. (Of course we should fix that anyway.)

That said, if we feel that this PR is going to improve most errors, I'm happy to merge it now. We can always iterate more on this later. :+1:

reactormonk:
Would you mind reviewing this PR? It's quite helpful.

hdgarrood:
This has been reviewed already, I think the only question is how to proceed - whether we want to merge now and iterate on this if necessary after it's in `master`, or whether we want to try to play with this a bit more to find if there are any cases it doesn't yet handle particularly well. But since nobody has objected to just merging it I think I'll do that. We can always revert later.

reactormonk:
Aye, I'll try to get my team to report a few use cases.

hdgarrood:
Yes, that would be great, thanks!

reactormonk:
Mind cutting a release with this?

hdgarrood:
I’d prefer not to - I don’t think this is important enough to justify diverging from our normal 6-week release schedule, and I’d prefer to have it in master for a while first so that we can test it out a little bit more before releasing properly.

Issue #3177, opened 2017-12-20
Make the repl printing function configurable


hdgarrood:
The context of this is that I'm starting to get somewhere with my `Show` replacement experiment (see https://github.com/hdgarrood/purescript-debugged/), and I've been testing it out in the repl. I've had to make a very hacky workaround to do this, which is to define a module called `PSCI.Support` which exports a function `eval`, because the repl is hardcoded to expect to find a function called `eval` within a module called `PSCI.Support`; what this currently means is that if you try to install `purescript-debugged` together with anything that depends on `purescript-psci-support` (which is most things), you'll get an error due to the module name clash. That is, it doesn't appear to be possible to override the repl printing function without making your code incompatible with any other library which depends on `psci-support`.

Of course you can use your own printing function already just by writing it out each time, i.e. `> myPrint $ whatever` inside the repl instead of just `> whatever`, but I think not having to do that can make quite a big difference in terms of how nice it is to use.

See also [what GHCi does](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/ghci.html#using-a-custom-interactive-printing-function) for reference.

The interface I'm proposing is:

- the repl exposes a configuration option `interactive-print` which can be left blank (in which case the repl continues to use `PSCI.Support.eval`), or set to a string like `Data.MyPrint.print`.
- if specified, this configuration option should be a fully-qualified identifier, which references a function whose type should be something like `forall a. C a => a -> T`.
- the repl then applies this function to any expression entered at the repl, obtaining a value of type `T`.
- the runtime representation of the type `T` should be the same as that of `Eff`, namely a nullary function which, when called, does something (possibly involving side-effects). In this case, the side-effects of the function would usually be printing a representation of the argument to the console. The repl finally calls the result of the previous step to actually perform the side-effects.

One thing I like about this proposal is that it reduces the repl's dependency on `prelude` and makes more space for alternative preludes; currently you kind of have to have `Show` instances for most data types in your project for the repl to be useful, even if you're not using the standard prelude in any other way. We generally aim to have the compiler tied to the standard prelude as little as possible, and I think this is one of the last remaining instances of overly strongly coupling between the compiler and the standard prelude.

paf31:
The idea with `psci-support` being a library was that you could patch or replace it to get different evaluation behavior for custom workflows.

hdgarrood:
I've opened this issue because I'm not aware of a good way of replacing it as things stand currently, but perhaps I'm missing something.

[...]

PR #3478, opened 2018-12-13
Add :print directive for customizable repl printing

hdgarrood:
Refs #3177. The command

    > :print MyModule.print

will replace the repl's printing function (by default:
PSCI.Support.eval) with MyModule.print, if the latter exists and appears
to work. The referenced function should have a type of the form

    forall a. (C a) => a -> Effect _

i.e., it should accept a wide variety of input types, and return
an `Effect T` for some type `T`, whose effects involve printing the
argument to the console. The `T` value is discarded.

It is not yet possible to configure the repl's printing function from
the .purs-repl file.


hdgarrood:
Reviewing can probably wait until #3471 is merged, as I'll probably want to rebase this once that has happened.

src/Language/PureScript/Interactive/Types.hs
```diff
@@ -57,27 +62,35 @@ data PSCiState = PSCiState
   [ImportedModule]
   [P.Declaration]
   [(P.Module, P.ExternsFile)]
+  (P.ModuleName, P.Ident)
```
natefaubion: Neither here nor there, but why not use record syntax for this rather than writing hand written selectors?

src/Language/PureScript/Interactive/Directive.hs
```diff
@@ -105,5 +106,6 @@ help =
   , (Show,     "loaded",    "Show all loaded modules")
   , (Paste,    "paste",     "Enter multiple lines, terminated by ^D")
   , (Complete, "<prefix>",  "Show completions for <prefix> as if pressing tab")
+  , (Print,    "<fn>",      "Set the repl's printing function to <fn>")
```
natefaubion: Should this specify that it should be fully qualified?

src/Language/PureScript/Interactive/Types.hs
```diff
@@ -57,27 +62,35 @@ data PSCiState = PSCiState
...
-psciExports (PSCiState _ _ _ _ x) = x
+psciExports (PSCiState _ _ _ _ _ x) = x
 
 initialPSCiState :: PSCiState
-initialPSCiState = PSCiState [] [] [] nullImports primExports
+initialPSCiState = PSCiState [] [] [] initialInteractivePrint nullImports primExports
+
+-- | The default interactive print function.
+initialInteractivePrint :: (P.ModuleName, P.Ident)
+initialInteractivePrint = (P.moduleNameFromString "PSCI.Support", P.Ident "eval")
```
natefaubion: Should this be `supportModuleName`? Otherwise I think that binding should be removed since I don't think we'd want to reference it.

src/Language/PureScript/Interactive/Types.hs
hdgarrood: I suspect this is the way it is because we don’t want to allow updating the fields directly but we do want to allow reading them. Come to think of it we could do something like declaring this as a record with fields prefixed with an underscore eg _psciImportedModules, and not expose the fields but continue to expose the functions (which only allow reading, not updating).

src/Language/PureScript/Interactive/Types.hs
hdgarrood: Oh actually, we do still use the `supportModuleName` binding, for the check during startup that `PSCI.Support` is defined. For now, I've left this check as-is, since I haven't implemented handling the `:print` directive in `.purs-repl` yet, which means that the printing function is always going to be `PSCI.Support.eval` to start with, and so it still makes sense to check that it is there (I think). I'll push a commit which removes the duplication.

natefaubion:
I think the code looks good. In testing it out, the only comment I have is that it might be good to have a way to easily restore the default. I don't think most people would know where the default as.

natefaubion:
Maybe calling `:print` without any arguments resets it? I don't have a better suggestions that doesn't involve another command.

hdgarrood:
I think if just `:print` alone without arguments reset it, it might be too easy to accidentally reset it when you didn't intend to. What if calling `:print` without arguments told you what it's currently set to and what the default is?

```
> :print MyModule.eval
> :print
The interactive print function is currently set to `MyModule.eval`
The default can be restored by entering `:print PSCI.Support.eval`
```

natefaubion:
👍

hdgarrood:
I've realised that the existing `:show` command is probably suitable for this, so I've added `:show print` for showing the currently configured repl function. I've made `:print` without arguments do the same thing (as discussed above), in the hope that this will improve discoverability.

PR #3480, opened 2018-12-14
Fix `isUnquotedKey`, fixes #3479

### LINKED ISSUE #3479: record labels are incorrectly quoted in errors
The function `Language.PureScript.Parser.Lexer.isUnquotedKey` does not correctly encode the fact that record labels must be quoted if they begin with numbers or uppercase letters. This causes error messages to produce invalid pretty-printed code in these (admittedly unusual) cases. For example:

```
test :: { "Oops" :: Int }
test = { }
```

produces

```
  Type of expression lacks required label Oops.

while checking that expression {}
  has type { Oops :: Int
           }
in value declaration test
```

hdgarrood:
Fixes labels not being quoted in error messages in some cases where they
should have been. For example:

Input code:
```
test :: { "Oops" :: Int }
test = { }
```

Before:
```
  Type of expression lacks required label Oops.

while checking that expression {}
  has type { Oops :: Int
           }
in value declaration test
```

After:
```
  Type of expression lacks required label "Oops".

while checking that expression {}
  has type { "Oops" :: Int
           }
in value declaration test
```

At the time of writing, there is just this one use of `isUnquotedKey`
throughout the entire compiler, so the refactoring I've done here is
safe.

There doesn't appear to be a good way to test this, so I haven't added tests.

### DIFF
--- a/src/Language/PureScript/Parser/Lexer.hs
+++ b/src/Language/PureScript/Parser/Lexer.hs
@@ -69,7 +69,7 @@ import Prelude.Compat hiding (lex)
 import Control.Applicative ((<|>))
 import Control.Monad (void, guard)
 import Control.Monad.Identity (Identity)
-import Data.Char (isSpace, isAscii, isSymbol, isAlphaNum)
+import Data.Char (isSpace, isAscii, isSymbol, isAlphaNum, isAlpha, isLower)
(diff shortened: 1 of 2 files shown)


hdgarrood:
:tada: thanks for the review!

Issue #2655, opened 2017-02-11
Stop using Template Haskell derived JSON encodings


paf31:
Before 1.0, we should probably remove any JSON encoders and decoders which use Template Haskell, and standardize any JSON formats emitted or parsed by the compiler.

It would probably also be a good idea to make sure that any JSON formats include the version number of the compiler they were generated with.

hdgarrood:
This doesn't necessarily have to be a breaking change, although I guess it probably will be in practice?

paf31:
I've been thinking about this some more. I think the goals are to

- Prevent future breakage due to changes in data types
- Make the JSON a bit more readable, if possible

If those are the only goals, then I'm going to suggest something based on `GHC.Generics`. We could use a computed type-level hash to make sure the types don't change without some sort of acknowledgment from the developer, and we could generate JSON encoders which are closer to the way we do things in CoreFn (i.e. "sexpr-style").

Any thoughts @hdgarrood?

hdgarrood:
That sounds okay, but what do we do if the types do change? Just write out the old instance by hand and then figure out how to modify the JSON instances so that it works with the new data type?

paf31:
@hdgarrood The idea was that if the data types change, then so will the hash, which is computed at compile time. So the user who breaks things will be forced to update the hash in code, acknowledging the fact that the format changed.

Another version of this would be to simply have separate data types solely for JSON encoding and decoding, and to copy things back and forth. Then we just wouldn't edit those modules, except in the case of a deliberate breaking change.

hdgarrood:
Okay, yeah. Either of those sound good to me, although I suppose the latter is probably better suited to dealing with both a) changing the JSON format and b) changing the compiler data type declarations internally without affecting the JSON format. I suppose an advantage to either of these approaches over manually written instances is that we don't then have to worry about having encode and decode instances not lining up.

joneshf:
> Prevent future breakage due to changes in data types

I'd like to suggest golden tests for this. I dunno if that's a thing that is happening already somehow, but this library is pretty easy to setup if you're interested: https://hackage.haskell.org/package/hspec-golden-aeson

hdgarrood:
Discussion on https://github.com/purescript/purescript/pull/2878 has reminded me of this and has made me harden my position somewhat.

[...]

PR #3484, opened 2018-12-15
Add annotations to Type and Kind

natefaubion:
This is a WIP which adds annotations to `Type` and `Kind`, and parses with `SourceAnn`. This is just a rough first pass which happens to compile. For the most part this is a robotic, conservative change, without any consideration to how or if we propagate source annotations to compiler generated types/kinds.

Some things that I'd like to do next:
* [x] Maybe have a type alias like `type SourceType = Type SourceAnn`
* [x] Have constructors pre-applied to `NullSourceAnn`.
* [x] Probably remove the Eq and Ord instances and have an "unsafe" newtype around them which ignores the annotations. I replaced all the cases of `==` I saw with `eqType`, but I'm not confident I caught them all and it very likely will break things.
* [ ] ~See how we can improve type and kind errors now that we have more source information.~


garyb:
If the tests are passing we should probably get this in asap and work out how to make the errors better afterwards - any time I've attempted this in the past, the slightest drift in `master` makes keeping these kind of branches up to date agonising.

natefaubion:
The only concern I have is around bloat of various JSON things. CoreFn doesn't emit types right now, I don't think, so it's not a problem there, but the stuff around docs/publish do. Is this something to be concerned about?

natefaubion:
Also, I kept the Eq and Ord instances, but they just ignore annotations. Removing the instances broke lots of other derived instances. While omitting them in Eq/Ord is a little dubious, it should at least be identical to the previous behavior.

natefaubion:
I've updated the Docs code to use `()` as the annotation, so it shouldn't leak source spans.

src/Language/PureScript/Kinds.hs
```diff
@@ -1,109 +1,164 @@
...
 
   kindType = primKind C.typ
   kindSymbol = primKind C.symbol
 
-instance A.FromJSON Kind where
-  parseJSON = toAesonParser id kindFromJSON
+instance {-# OVERLAPPING #-} A.FromJSON (Kind SourceAnn) where
+  parseJSON = toAesonParser id (kindFromJSON (pure NullSourceAnn) fromAesonParser)
+
+instance {-# OVERLAPPING #-} A.FromJSON a => A.FromJSON (Kind a) where
```
hdgarrood: Are we using both of these instances?

hdgarrood: (perhaps a comment to explain what's going on here would help)

src/Language/PureScript/Kinds.hs
```diff
@@ -1,109 +1,164 @@
...
+compareKind :: Kind a -> Kind b -> Ordering
+compareKind (KUnknown _ a) (KUnknown _ a') = compare a a'
+compareKind (KUnknown {}) _ = LT
+
+compareKind (Row _ a) (Row _ a') = compareKind a a'
+compareKind (Row {}) _ = LT
+
+compareKind (FunKind _ a b) (FunKind _ a' b') = compareKind a b <> compareKind a' b'
+compareKind (FunKind {}) _ = LT
+compareKind _ (FunKind {}) = GT
```
hdgarrood: Is this necessary? Presumably if we reach this case, the first argument must be a `NamedKind`, and therefore we can let it fall through and hit the very last case? Or, alternatively, if we're following the style of `compareType`, should we have a case like this for the other constructors too?

src/Language/PureScript/Kinds.hs
```diff
@@ -1,109 +1,164 @@
...
+srcRow = Row NullSourceAnn
 
-instance NFData Kind
+srcFunKind :: SourceKind -> SourceKind -> SourceKind
+srcFunKind = FunKind NullSourceAnn
+
+srcNamedKind :: Qualified (ProperName 'KindName) -> SourceKind
+srcNamedKind = NamedKind NullSourceAnn
 
 -- This is equivalent to the derived Aeson ToJSON instance, except that we
```
hdgarrood: This comment is no longer accurate, because the derived ToJSON instance would put annotations into the `contents` field, right?

src/Language/PureScript/Kinds.hs
```diff
@@ -1,109 +1,164 @@
...
+      obj "NamedKind" a n
     where
-    obj :: A.ToJSON a => Text -> a -> A.Value
-    obj tag contents =
-      A.object [ "tag" .= tag, "contents" .= contents ]
+    obj :: A.ToJSON b => Text -> a -> b -> A.Value
+    obj tag ann contents =
+      A.object [ "tag" .= tag, "annotation" .= ann, "contents" .= contents ]
 
 -- This is equivalent to the derived Aeson FromJSON instance, except that it
```
hdgarrood: I think this comment is now outdated for a similar reason (the `annotation` field vs the `contents` field)

hdgarrood:
Will this affect the externs format?

natefaubion:
Not appreciably, but it will keep annotations on types and kinds. Do you have any specific concerns?

src/Language/PureScript/Kinds.hs
natefaubion: It would, and I can just add it to the contents, or derive them. I don't think the backwards compatibility these instances were trying to preserve are still relevant (we haven't supported Star and Symbol as is for a while).

src/Language/PureScript/Kinds.hs
natefaubion: You are right, this looks fishy.

src/Language/PureScript/Kinds.hs
hdgarrood: Ohh - this hadn't occurred to me until now; we'll have to clear out the Pursuit database anyway because the Type codec will be changing, right?

natefaubion: Unfortunately, yes. We can perhaps stop deriving Type, if we want to avoid it. The docs are using `()` for the annotation, so it's "unused", but it will still try to parse the field with the derived Type codec.

hdgarrood: I had intended to come back to Type and put together manual json codecs for it, incidentally, although obviously I haven't managed that yet. Ideally none of our codecs would be derived, I think, at least not for anything appearing in json for docs. See also https://github.com/purescript/purescript/issues/2655

hdgarrood: How about leaving this for now, and making a new issue to remind us to come back to it before we make another release, to see if we can move away from derived codecs for Type and make this non-breaking from that perspective?

src/Language/PureScript/Kinds.hs
hdgarrood: I agree that the parser code handling Star and Symbol is very unlikely to still be relevant (although perhaps best to deal with that separately). I'm just suggesting that we update the comment to make it more accurate.

natefaubion:
I think I've addressed your comments.

natefaubion:
I created https://github.com/purescript/purescript/issues/3494 to track the specific issue with `Type`.

garyb:
:tada:

PR #3486, opened 2018-12-17
Prevent invalid JS being generated from awkward record labels

### LINKED ISSUE #3481: Record labels beginning with numbers are not quoted during JS codegen
For example (in 6976148a42b435b9f4a1652edb7fdcff53683f3c):

```purescript
module Main where

import Prelude
import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = do
  let rec = { "0": 0 }
  log (show rec."0")
```

Output when run:
```
/home/harry/code/ps-scratch/output/Main/index.js:11
    return Effect_Console.log(Data_Show.show(Data_Show.showInt)(rec.0));
                                                                ^^^

SyntaxError: missing ) after argument list
```

I think the problem is here:

https://github.com/purescript/purescript/blob/6976148a42b435b9f4a1652edb7fdcff53683f3c/src/Language/PureScript/CodeGen/JS/Common.hs#L35-L37

I think `identToJs` is fine, because if we have an `Ident`, we know that it cannot start with a number. However, we have no such guarantee for `Text`, which perhaps means we need to handle the case where the given `Text` does begin with a number separately, either here or in `properToJs`.

hdgarrood:
Fixes #3481. I've also included a commit with a small refactoring and improvement of some docs: in particular, I've removed a the now-misleading comment about what is done with symbols, because `Ident` values no longer contain symbols since we started requiring symbols to be aliases for named functions. I'd suggest reviewing each commit separately.

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -233,16 +233,16 @@ moduleToJs (Module _ coms mn _ imps exps foreigns decls) foreign_ =
     ds' <- concat <$> mapM bindToJs ds
     ret <- valueToJs val
     return $ AST.App Nothing (AST.Function Nothing Nothing [] (AST.Block Nothing (ds' ++ [AST.Return Nothing ret]))) []
-  valueToJs' (Constructor (_, _, _, Just IsNewtype) _ (ProperName ctor) _) =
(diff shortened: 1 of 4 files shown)


hdgarrood:
Thanks!

PR #3495, opened 2018-12-22
Add better source positions to kind errors

natefaubion:
This implements https://www.youtube.com/watch?v=rdVqQUOvxSU for the kind checker. The kinds in kind unification errors now have exact positions for the terms in the source they are derived from. Since unification doesn't really have a canonical source for the error (could be either one), and we don't have a way of reporting multiple positions (I guess we could throw two errors?), it just has a bias when reporting the source position. This seems to be a huge improvement in all the testing I've done.

screen shot 2018-12-21 at 11 23 12 pm


src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -91,7 +94,11 @@ unifyKinds k1 k2 = do
   go (FunKind _ k1' k2') (FunKind _ k3 k4) = do
     unifyKinds k1' k3
     unifyKinds k2' k4
-  go k1' k2' = throwError . errorMessage $ KindsDoNotUnify k1' k2'
+  go k1' k2' = do
+    let sa = case getAnnForKind k2' of
+               NullSourceAnn -> getAnnForKind k1'
+               ann -> ann
+    throwError . errorMessage' (fst sa) $ KindsDoNotUnify k1' k2'
```
natefaubion: I guess since PositionedError takes a NEL, I could put both in there.

src/Language/PureScript/Errors.hs
```diff
@@ -205,6 +205,15 @@ errorMessage' ss err = MultipleErrors [ErrorMessage [positionedError ss] err]
 errorMessage'' :: NEL.NonEmpty SourceSpan -> SimpleErrorMessage -> MultipleErrors
 errorMessage'' sss err = MultipleErrors [ErrorMessage [PositionedError sss] err]
 
+-- | Create an error from multiple (possibly empty) source spans, reversed sorted.
+errorMessage''' :: [SourceSpan] -> SimpleErrorMessage -> MultipleErrors
```
natefaubion: Keeping the tradition alive.

nwolverson:
> and we don't have a way of reporting multiple positions (I guess we could throw two errors?)

I think a potential future improvement (to JSON error reporting in general) is to add an additional source position, the LSP has a concept of `DiagnosticRelatedInformation` intended for just this kind of "here's the thing that conflicted" case.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -150,19 +156,19 @@ kindsOfAll
   -> [(ProperName 'TypeName, [(Text, Maybe SourceKind)], [SourceType])]
   -> m ([SourceKind], [SourceKind])
 kindsOfAll moduleName syns tys = fmap tidyUp . withFreshSubstitution . captureSubstitution $ do
-  synVars <- replicateM (length syns) freshKind
+  synVars <- replicateM (length syns) freshKind'
```
kRITZCREEK: Should we traverse these `syns` instead and use their spans to initialize the fresh kinds here? (Also goes for all the other instances of `replicateM (length x) y` in here)

Something like:

```haskell
  synVars <- traverse (\(_, _, sourceTy) -> freshKind (getAnnForType sourceTy)) syns
```

src/Language/PureScript/Types.hs
```diff
@@ -424,7 +426,7 @@ freeTypeVariables = ordNub . go [] where
 quantify :: Type a -> Type a
 quantify ty = foldr (\arg t -> ForAll ann arg t Nothing) ty $ freeTypeVariables ty
   where
-  ann = annotationForType ty
+  ann = getAnnForType ty
```
kRITZCREEK: I don't think this name is any more helpful than just inlining its definition.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -181,10 +187,10 @@ solveTypes
 solveTypes isData ts kargs tyCon = do
   ks <- traverse (fmap fst . infer) ts
   when isData $ do
-    unifyKinds tyCon (foldr (FunKind nullSourceAnn) kindType kargs)
+    unifyKinds tyCon (foldr srcFunKind kindType kargs)
     forM_ ks $ \k -> unifyKinds k kindType
-  unless isData $
-    unifyKinds tyCon (foldr (FunKind nullSourceAnn) (head ks) kargs)
+  unless isData $ do
```
kRITZCREEK: Don't need that do, unless you're doing it for consistency with the line above?
```suggestion
  unless isData $
```

src/Language/PureScript/TypeChecker/Kinds.hs
natefaubion: Potentially, but I think we need to change the signature of this function to take source positions directly. The `SourceType` here refers to the rhs constructors, however I think when kind-checking a group like this we would want the source position to point to the lhs name.

natefaubion:
I've addressed the feedback.

PR #3496, opened 2018-12-23
Use handwritten JSON instances for Type/Kind

### LINKED ISSUE #3494: Remove derived JSON instances for Type
#3484 added annotations to `Type` which technically breaks compatibility since the JSON instances are derived. We should remove these and do something similar to what we've done with the `Kind` codecs before the next release.

natefaubion:
Fixes #3494 

I think this should be compatible, though I'm not super familiar with Aeson's encoding. I don't think there are any tests for these, but I can probably add some.


src/Language/PureScript/Kinds.hs
```diff
@@ -48,67 +47,65 @@ srcFunKind = FunKind NullSourceAnn
...
+      [ "tag"        .= tag
+      , "annotation" .= annToJSON ann
+      , "contents"   .= contents
+      ]
+
+instance A.ToJSON a => A.ToJSON (Kind a) where
+  toJSON = kindToJSON toJSON
+
+kindFromJSON :: forall a. Parser a -> (Value -> Parser a) -> Value -> Parser (Kind a)
+kindFromJSON defaultAnn annFromJSON = A.withObject "Kind" $ \o -> do
```
natefaubion: I removed the old kind parsing from 0.10. I don't think it's necessary to preserve these anymore.

src/Language/PureScript/Kinds.hs
```diff
@@ -48,67 +47,65 @@ srcFunKind = FunKind NullSourceAnn
...
+  a   <- (o .: "annotation" >>= annFromJSON) <|> defaultAnn
+  let
+    contents :: A.FromJSON b => Parser b
+    contents = o .: "contents"
+  case tag of
     "KUnknown" ->
-      KUnknown <$> annFromJSON' <*> key "contents" (nth 0 asIntegral)
-    "Star" ->
-      kindType <$> defaultAnn
+      KUnknown a <$> contents
```
natefaubion: I'm pretty sure this parser was broken. It appears to parse with `nth 0`, but it was not encoded with a wrapper list.

natefaubion:
Are these codecs exercised by the Docs tests? I guess the `KUnknown` case didn't arise because they only exist during checking.

hdgarrood:
They are only partially exercised by the Docs tests; we only test that the parser is compatible with the json produced by the current version. I've been meaning to add some tests to ensure that we don't accidentally break compatibility, because we don't have those right now.

natefaubion:
I have not looked into it. Be my guest 😁

Issue #1228, opened 2015-07-03
Only warn on typed wildcard usage in top-level type signatures


garyb:

paf31:
This seems a bit too heavy-handed to me. I think of wildcards as a tool, and I might want to use a wildcard to infer a type inside a `let` binding, even though I agree that usually those tend to be the places where you don't care as much.

I think we might need a way to disambiguate between "I genuinely don't care what this type is" and "please infer this for me". One simple way might be to allow named wildcards:

```purescript
foo = do
  x <- get :: _state Int
```

where the expectation is that the warning should name the wildcard to help understand the inferred types:

```text
The type wildcard _state has the inferred type

    ExceptT e1 m1
```

where the usual `_` wildcard would take on the meaning "I don't care about the type, just infer it".

garyb:
I was talking to Jon and/or Becky about this the other week actually, and came to the same conclusion.

I feel like the named wildcards should be the non-warned kind though: if you're giving them a name it's to give them some meaning, which means you can provide an idea of what it is your talking about. `_state` being a great example there, as that usage can explain what you're doing: "I'm inferring a state monad here, don't worry about it`, whereas `x <- get :: _ Int` gives you little clue to what `_` is without considering the wider context.

garyb:
Didn't you suggest allowing named wildcards to do things like `_x -> _x` at one point too? That's something that `_` could never ever help you with.

paf31:
Yes, that's the other issue with this - there is an orthogonal set of concerns - whether we name a wildcard to unify it with other wildcards.

All considered, I'm not sure naming is the right way to solve this issue, but I think it will be good to be explicit about which wildcards we want to print out.

paf31:
We could deal with this and typed holes at the same time perhaps: use a typed hole when you want the type, and a wildcard when you don't:

```purescript
names = map getName? people
```

would give an error (not warning) along the lines of

```text
The hole getName? has the inferred type

  Person -> String
```

garyb:
That would be great for most cases but I think there's still value in being able to get information out of a wildcard in a type too (I've used `Natural Something _` kind of thing a few times while working on/in Halogen). Perhaps we could use something similar there too, `?what -> Unit`.

I think we should still always warn on top-level wildcards still too, as that's like leaving the type off from the perspective of reading the code.

[...]

Issue #715, opened 2014-11-17
Named type wildcards


paf31:
It would be nice to assert equality of type wildcards by using the same name:

``` haskell
test :: forall a. StateT _s (ReaderT _r) a -> StateT _s (ReaderT _r) a
```

The format for a type wildcard would be either a naked underscore `_` (unnamed) or with a label `_a` (named).

hdgarrood:
Can we close this now? @natefaubion

garyb:
What we have doesn't allow for equality between the wildcards, but maybe that could be reopened as a separate issue.

PR #3500, opened 2018-12-30
Add named type wildcards

natefaubion:
Implements #715 #1228

This is a slight UX breaking change. This adds `?foo` syntax for types, which act as named wildcards which always warn. The previous `_` syntax acts as a partial type signature, which does not warn unless it's part of a module-level type signature.

I didn't change the warning (`WildcardInferredType`) except to add the name, but maybe it warrants a separate warning.

### DIFF
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -43,8 +43,6 @@ import Language.PureScript.Types
 
 import Lens.Micro.Platform ((^..), _1, _2)
 
-import Debug.Trace as T
-
 addDataType
   :: (MonadState CheckState m, MonadError MultipleErrors m, MonadWriter MultipleErrors m)
   => ModuleName
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -16,10 +16,10 @@ import Control.Monad (when, unless, void, forM)
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State.Class (MonadState(..), modify, gets)
 import Control.Monad.Supply.Class (MonadSupply)
-import Control.Monad.Writer.Class (MonadWriter(..))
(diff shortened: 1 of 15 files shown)


src/Language/PureScript/TypeChecker.hs
```diff
@@ -43,6 +43,8 @@ import Language.PureScript.Types
 
 import Lens.Micro.Platform ((^..), _1, _2)
 
+import Debug.Trace as T
```
LiamGoodacre: (Debugging artefact)

natefaubion:
One UX consideration is should `?foo` in types emit an error rather than a warning? It's necessary for values, but not for types. If we leave it as a warning, that means you can accidentally leave `?foo` in a module inadvertently since we don't persist warnings.

LiamGoodacre:
I'd prefer if both type-level and term-level holes behaved the same, so if term holes error then so should type holes.

LiamGoodacre:
To add to that - I feel like `_` shouldn't ever error or warn, and `?foo` should always error.
Now that we have the option to use `?foo`, the intent behind using `_` becomes "fill in this bit for me" instead of "help me figure this bit out".

natefaubion:
Currently we warn if you leave off a type signature for module-level declarations, which is why I preserved the `_` warning there. Otherwise, `foo = 42 :: _` just becomes a way to not get warnings while still leaving off signature.

rightfold:
> I'd prefer if both type-level and term-level holes behaved the same, so if term holes error then so should type holes.

And they should be called _kinded holes_ and report only the kind. :sweat_smile:

natefaubion:
This is ready for review again. I've addressed the UX concerns.

* `?foo` always raises a `HoleInferredType` error. When it's in a type signature, it just does not provide a `TypeSearch`.
* `_` does not warn, unless it's in a module-level type signature, where it continues to throw a `WildcardInferredType` warning.

Issue #3504, opened 2019-01-06
Links on declaration titles are broken for re-exports from other packages


hdgarrood:
For example, the link I'm hovering over in this screenshot is broken, because it assumes that `Cons` exists in the current package. These docs are from [version 3.0.0 of typelevel-prelude](https://pursuit.purescript.org/packages/purescript-typelevel-prelude/3.0.0/docs/Type.Row).

![screenshot from 2019-01-06 18-00-44](https://user-images.githubusercontent.com/1270186/50739812-07ee1980-11dd-11e9-9b26-2a30b5a7153c.png)

I'd like to consider solving this by removing links on declaration titles entirely, and just making them bold -  this is what Hackage does. I think this would be a better UI, as it would better distinguish the name being declared from everything else. This could be particularly helpful for certain declarations such as type classes, for instance, since it's not always obvious which is the superclass and which is the class currently being declared (if you're not sufficiently familiar with the syntax). The links are sort of pointless anyway, as they would put you right where you already are; if you want a link to copy which you can send to other people, that's what the hash (#) to the left of the declaration is for.

PR #3506, opened 2019-01-07
Remove the concept of the 'current module' in Docs

hdgarrood:
The concept of the current module is unnecessary (we aren't using it
anywhere either in the compiler or in Pursuit) as well as potentially
confusing; for example, what is the 'current module' when we are
rendering re-exported declarations? Additionally, since the information
is not used, it would be easy for it to become incorrect without anyone
noticing.

This commit refactors the Docs related code, removing the concept of the
current module. Specifically, the following have been removed:

- the SameModule constructor from the LinkLocation data type;
  wherever we previously would have used that, we can now use the
  LocalModule constructor, which encodes precisely the same information.
- the 'current module' field from the constructors LocalModule and
  DepsModule of the data type LinkLocation, which was unused.
- the 'currentModuleName' field of the HtmlRenderContext data type,
  which was also unused.

I came across this refactoring opportunity while looking into #3504.

### DIFF
--- a/app/Command/Docs/Html.hs
+++ b/app/Command/Docs/Html.hs
@@ -48,8 +48,7 @@ writeHtmlFile filepath =
 
 getHtmlRenderContext :: P.ModuleName -> D.HtmlRenderContext
 getHtmlRenderContext mn = D.HtmlRenderContext
-  { D.currentModuleName = mn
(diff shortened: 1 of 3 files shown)


hdgarrood:
Thanks for the review!

Issue #3185, opened 2017-12-22
Make Releases Easier


joneshf:
One of the questions raised in https://github.com/purescript/purescript/issues/3178 was: why does the `0.12-dev` branch exists? The long and short is that bug fixes for 0.11.6 were on `master`, but a release couldn't be made that would only include bug fixes (instead of both bug fixes and breaking changes). In answering that question, it surfaced that making releases is currently hard.

I'd like to make releases not hard. Then bug fixes can be released without a bunch of ceremony and wouldn't require a breaking change for lack of people-power.

I don't know the pain points that go into a release, and won't speculate what they are. I will speculate that the less a person has to do, the easier releases become. So, I think the way forward is to automate as much as possible and fix problems in the process as they arrive.

Some searching hasn't yielded a definitive list of release steps. Is there a list of steps written down somewhere that explains what it takes to make a release? If not, let's start that list here.

hdgarrood:
I think the closest thing we have to a list of release steps is previous "release checklist" issues: #2745, #2151. Those issues are for breaking releases, and of course non-breaking releases are a fair bit simpler.

I would like to have a list of release steps in a markdown file in this repo; we can base it off those issues.

seanwestfall:
Was there ever a roadmap written to explain where future PureScript work was headed? Elm has one, just to think of an example, https://github.com/elm-lang/projects/blob/master/roadmap.md, but it doesn't look like there's anything in https://github.com/purescript/documentation, is that on purpose?

hdgarrood:
No, there isn't, and I think this is because it would be quite a bit of effort to make sure it was kept up to date and accurate. If you want to get a feel for where future work is headed, the best thing to do is to look at the issue tracker, and specifically the milestones which mention an actual version, like 0.12.0 and 1.0 (note that the 1.0 milestone just means "any time before 1.0", and so things tagged as 1.0 may well happen in an 0.x release).

garyb:
We could also just be sure to provide better messages when we merge things, since you can rewrite them in the GH UI - if we always rewrote them to be "Fix: something meaningful", etc. then using `make_release_notes` would give us a pretty good list that just needs sorting also.

hdgarrood:
Presumably there will be quite a few commits which shouldn't correspond to entries in the changelog, though? I think it's probably easier to require changelog updates than to try provide better commit messages as well. Totally up to you though since you seem to be doing most of that work at the moment!

garyb:
[...]

PR #3509, opened 2019-01-12
Add a template release checklist, refs #3185

hdgarrood:
The intention is that this file can serve as a template / starting point
for each future release.

### DIFF
--- a/RELEASE_CHECKLIST.md
+++ b/RELEASE_CHECKLIST.md
@@ -35,6 +35,7 @@ Have any of the following JSON formats changed? If so, it may be worth
 considering what effects this may have:
 
 - [ ] Corefn
+- [ ] Ide protocol
 - [ ] JSON produced by `purs publish`
   - [ ] check whether this affects Pursuit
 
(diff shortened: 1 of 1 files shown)


JordanMartinez:
Maybe including a general visual dependency graph would help, too? It wouldn't necessarily list every step, but it would include each 'stage' of things that need to be done before anything else gets done.

hdgarrood:
I don't think that's necessary; the people who are actually working through these steps would probably know what order they need to be done in.

JordanMartinez:
For that audience (core contributors), that makes sense. However, there's a larger issue I have in mind, but I don't have all the details for it yet. Still, this is the wrong place to have that discussion. Sorry about that :sweat_smile:

RELEASE_CHECKLIST.md
```diff
@@ -0,0 +1,61 @@
...
+- [ ] ide plugins
+
+## JSON formats
+
+Have any of the following JSON formats changed? If so, it may be worth
+considering what effects this may have:
+
+- [ ] Corefn
+- [ ] JSON produced by `purs publish`
+  - [ ] check whether this affects Pursuit
```
kRITZCREEK: The IDE protocol is another JSON format that might break (though the affected users are a smaller group as it's primarily plugin authors)

Issue #2216, opened 2016-07-05
Add position information to types and data constructors


kRITZCREEK:
This would allow `psc-ide` to jump to the definition site of data constructors, aswell as `psc-docs` to extract docstrings for them. 

We could also improve Errors around mistakes when declaring data constructors. For example:

``` purescript
data List a
  = Nil
  | Cons a List
```

Currently points at the `data` with the error message:

```
  Could not match kind

    * -> *

  with kind

    *

in type constructor List

```

I think it would be a lot better to point the user at the data constructor `Cons` instead of the type constructor `List`, because that suggests he should be changing the LHS of his definition.

paf31:
Note, we don't even have a `PositionedType` right now.

garyb:
*whispers* `Mu` and `Cofree` :smiling_imp: 

In all seriousness, I want to explore that approach on a branch soon (for `CoreFn`, because it's simpler) as our current representations are pretty inelegant, and extremely error prone when traversed directly, and adding more positioned constructors will compound the problems.

jdegoes:
> whispers Mu and Cofree 😈

They are, of course, *noticeably* slower. But far more elegant and robust.

paf31:
I agree they're more elegant. I decided not to use `Cofree` for the first version because that requires every contributor to understand `Cofree`, essentially, since these modules are used by pretty much everything in the system. I would also use `bound` if that wasn't a concern.

garyb:
It's not so much the inelegance that bothers me as the number of times I've been bitten by `Positioned`-related issues - while working on things and it being tricky to track down what the cause is (remembering to re-wrap as well as extract positioned things - there was some point futher along in the compiler that broke when I accidentally erased position info), and there have been bugs that made it into releases that were purely due to cases being missed due to missing position-unwraps.

garyb:
I can only imagine the chaos that will ensue by positioning types in the way we do decls/exprs, as _every_ pattern involving types will need to be updated accordingly too.

paf31:
But how does `Cofree` help there? Presumably we can still forget to deal with position information. I suppose the standard recursion scheme for `Cofree` will tend to push us in the right direction in most cases.

jdegoes:
You can not only ensure annotation of all nodes (if you desire), but you can make the ADTs much more flexible, combining `Coproduct` of individual terms (functors) to allow transformations to add and remove information through the phases of a compiler in a totally type safe way.

[...]

PR #3511, opened 2019-01-13
Fix kind error for recursive data type

natefaubion:
@kRITZCREEK noticed that the kind error noted in https://github.com/purescript/purescript/issues/2216 is not any better.

New error:
```
[1/1 KindsDoNotUnify] KindSpan.purs:5:12

  5    | Cons a List
                ^^^^

  Could not match kind

    Type -> Type

  with kind

    Type

  in type constructor List
```


kRITZCREEK:
Awesome! Sorry for nerd-sniping you :D

PR #3513, opened 2019-01-13
Turn License generator into a proper stack script

kRITZCREEK:


garyb:
Do we need to add the dependency too, or can it figure that out?

kRITZCREEK:
It does based on the documentation here: https://haskell-lang.org/tutorial/stack-script

In the stack user-guide it still says to add the packages https://docs.haskellstack.org/en/stable/GUIDE/#script-interpreter

whichever you prefer is fine with me

kRITZCREEK:
@hdgarrood Good catch, I actually ran into the same and had just forgotten to apply the patch.

license-generator/generate.hs
```diff
@@ -6,7 +9,7 @@
 --
 -- It is recommended to run this as follows:
 --
--- stack list-dependencies --flag purescript:RELEASE | stack exec runhaskell license-generator/generate.hs > LICENSE
+-- stack list-dependencies --flag purescript:RELEASE | stack license-generator/generate.hs > LICENSE
```
kRITZCREEK:
```suggestion
-- stack ls dependencies --flag purescript:RELEASE | stack license-generator/generate.hs > LICENSE
```

hdgarrood:
The appveyor failure is not relevant (and besides this PR doesn't touch anything which is exercised by the CI script anyway). Shall we merge this?

kRITZCREEK:
Sounds good to me :)

PR #3514, opened 2019-01-14
Fix: ado notation breaks docs generation

### LINKED ISSUE #3414: purs docs breaks when using Applicative do notation
I'm not sure why but here is a minimal test case:

```purescript
module Main where
  
map f x = f x

test = 
  ado x <- 1
   in x
```

`purs compile Main.purs` works fine (with warnings) but `purs docs Main.purs` fails.

hdgarrood:
Fixes #3414


hdgarrood:
Thanks!

PR #3519, opened 2019-01-20
Add annotation meta to `Constructor` in CoreFn.

utkarshkukreti:
Right now `Constructor` meta is only added to the places where an instance of it is being created. This means we can't get it for any type which is not used anywhere in the module. This change attaches the same meta to the `Constructor` type as well.


natefaubion:
> Right now Constructor meta is only added to the places where an instance of it is being created. This means we can't get it for any type which is not used anywhere in the module. This change attaches the same meta to the Constructor type as well.

I don't fully understand what this means. Do you mind posting an example of what you mean, and how this change solves it?

hdgarrood:
I’m going to close this for now due to inactivity, but we can reopen once @natefaubion’s question has been addressed.

PR #3520, opened 2019-01-20
preserve source spans in type desugaring

### LINKED ISSUE #3501: Type desugaring should preserve original source spans in inserted code
For example `a -> b` gets desugared to `TypeApp (TypeApp (TypeConstructor Prim.Function) a) b` where the inserted applications should map to original source spans instead of `NullSourceSpan`.

colinwahl:
This intends to close #3501.

@natefaubion I doubt that I've covered everything you were referencing in the ticket - let me know if this is the case and I'd be happy to go through and fix more with a poke in the right direction.

### DIFF
--- a/src/Language/PureScript/Environment.hs
+++ b/src/Language/PureScript/Environment.hs
@@ -312,15 +312,22 @@ kindRow srcKind = Row (getAnnForKind srcKind) srcKind
 primTy :: Text -> SourceType
 primTy = TypeConstructor nullSourceAnn . primName
 
+primTy' :: Text -> SourceAnn -> SourceType
+primTy' tyName ann = TypeConstructor ann (primName tyName)
+
 -- | Type constructor for functions
 tyFunction :: SourceType
 tyFunction = primTy "Function"
+
+-- | Type constructor for functions with a source ann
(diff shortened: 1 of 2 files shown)


natefaubion:
Sorry, I will try to take a look at this soon. I think there is probably more work to be done. For example, `tyFunction` in this case still has null source spans.

@hdgarrood The issue is that when we get to the kind checker, we have fewer options when reporting errors because there are a lot of desugared terms that have no positions.

colinwahl:
@natefaubion I added a case to preserve the source spans in `tyFunction`. I see that all of the `tyBoolean`, `tyInt`, etc are used in `Language.PureScript.TypeChecker.Types`.  Should I create versions of all of these that contain a source span? I can't tell if they are used, but if they are, I'd be happy to.

JordanMartinez:
@colinwahl Are you still interested in working on this?

colinwahl:
@JordanMartinez I'm not even sure where this left off 😆 If you are interested in picking it up, please do! I'll close this for now.

JordanMartinez:
Thanks! I'd like to work on this as the next thing to tackle once my other PR gets in.

Issue #3503, opened 2019-01-06
Require existence of externs files when producing docs


hdgarrood:
This originally arose in my mind as a way of fixing #3264, but it would be a breaking change to the CLI so I think a dedicated issue is probably warranted.

At the moment, we invoke the type checker during generation of docs in order to fill in any type signatures for declarations where the type signatures have not already been provided. However, we can't perform all of the desugaring steps which would normally have been run by this point, because some of them modify the AST in such a way that we wouldn't be able to produce sensible docs from them (e.g. desugaring of type classes and type synonyms). Often this is not a problem, but occasionally we see errors in docs generation which only occur when type signatures are left off (e.g. #3264).

To address this, I am proposing that we instead modify `purs docs` to accept an `--output` option, like `purs compile` does, and have `purs docs` first check that up-to-date externs files exist for all of the provided source files, and fail if not. Then, `purs docs` could obtain any missing types from the externs files. As with `purs compile`, the `--output` option would have a default of "output". However, it would only be used for locating externs files.

A nice side effect of doing this would be that it would prevent situations where the docs have incorrect types. For instance, if I write

```purescript
foo :: String
foo = 3
```

in a module and run `purs docs` on it, provided that all value declarations in that module have type signatures, it will succeed and the resulting docs will say that `foo` has the type `String`.

This would also pave the way for using the re-exports information contained in externs files (see #2191) to replace most of the code in `Docs.Convert.ReExports`. I would be very happy about this, because not only is that code quite frankly horrible (I'm allowed to say that because I wrote it), but also I suspect it's one of the main causes of the awful performance and memory consumption of `purs docs`. However, to do that, I think we'd need to add the module a re-export is being re-exported from to the externs format (currently, we only include the module a re-export originally came from, see https://github.com/purescript/purescript/pull/2191#issuecomment-225186677).

garyb:
Seems reasonable to me. I think when I did #2191, one of the various motivations behind it was to remove the need for `Docs.Convert.ReExports`. I probably didn't realise that just dropping it in wasn't going to work, until I tried it out (I'm not so familiar with a lot of the docs code).

hdgarrood:
[...]

PR #3521, opened 2019-01-20
Add the module from which something was imported to ReExportRefs

hdgarrood:
Implements what was discussed in #2191 by adding the module from which something was imported to the `ReExportRef` in the externs files. This is a first step towards #3503 (and also removing Docs.Convert.ReExports), but might also be useful for purs repl and/or purs ide?

Here's an example of what a section of an externs.json file might look like before and after this change.

```purescript
-- src/Main.purs
module Main
  ( module P
  ) where

import Prelude (class Eq) as P
```
externs.json (before)
```json
  "ReExportRef": [
    {
      "start": [
        1,
        1
      ],
      "name": "src/Main.purs",
      "end": [
        5,
        50
      ]
    },
    [
      "Data",
      "Eq"
    ],
    {
      "TypeClassRef": [
        {
          "start": [
            1,
            1
          ],
          "name": "src/Main.purs",
          "end": [
            5,
            50
          ]
        },
        "Eq"
      ]
    }
  ]
```
externs.json (after)
```json
  "ReExportRef": [
    {
      "start": [
        1,
        1
      ],
      "name": "src/Main.purs",
      "end": [
        5,
        50
      ]
    },
    {
      "exportSourceImportedFrom": [
        "Prelude"
      ],
      "exportSourceDefinedIn": [
        "Data",
        "Eq"
      ]
    },
    {
      "TypeClassRef": [
        {
          "start": [
            1,
            1
          ],
          "name": "src/Main.purs",
          "end": [
            5,
            50
          ]
```

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -308,7 +308,7 @@ data DeclarationRef
   -- A value re-exported from another module. These will be inserted during
   -- elaboration in name desugaring.
   --
-  | ReExportRef SourceSpan ModuleName DeclarationRef
+  | ReExportRef SourceSpan ExportSource DeclarationRef
   deriving (Show, Generic, NFData)
 
 instance Eq DeclarationRef where
@@ -322,6 +322,13 @@ instance Eq DeclarationRef where
   (KindRef _ name) == (KindRef _ name') = name == name'
   (ReExportRef _ mn ref) == (ReExportRef _ mn' ref') = mn == mn' && ref == ref'
   _ == _ = False
+
+data ExportSource =
+  ExportSource
+  { exportSourceImportedFrom :: Maybe ModuleName
+  , exportSourceDefinedIn :: ModuleName
+  }
+  deriving (Eq, Ord, Show, Generic, NFData)
 
 -- enable sorting lists of explicitly imported refs when suggesting imports in linting, IDE, etc.
 -- not an Ord because this implementation is not consistent with its Eq instance.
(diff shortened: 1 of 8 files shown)


hdgarrood:
Note to self: The `importName` will be qualified with the pseudo-module name if it arises from a pseudo-module re-export (like the `module P` in the above example). I think I need to modify the `ImportRecord` type to track the (real) module something was imported from too.

hdgarrood:
The above comment is wrong: I had the right approach in the first place, I was just using the wrong module name when building up the `Exports`. This seems to work now.

hdgarrood:
In the interest of making this easier to review (and mergeable without committing us to a breaking release), all I've done in this PR is modify the `ReExportRef` constructor to include the module from which a re-exported declaration was imported from (as well as the module from which it was originally defined in). At the moment, no other part of the compiler is making any use of this information.

hdgarrood:
Here's an example of what a section of an externs.json file might look like before and after this change.

```purescript
-- src/Main.purs
module Main
  ( module P
  ) where

import Prelude (class Eq) as P
```
externs.json (before)
```json
  "ReExportRef": [
    {
      "start": [
        1,
        1
      ],
      "name": "src/Main.purs",
      "end": [
        5,
        50
      ]
    },
    [
      "Data",
      "Eq"
    ],
    {
      "TypeClassRef": [
        {
          "start": [
            1,
            1
          ],
          "name": "src/Main.purs",
          "end": [
            5,
            50
          ]
        },
        "Eq"
      ]
    }
  ]
```
externs.json (after)
```json
  "ReExportRef": [
    {
      "start": [
        1,
        1
      ],
      "name": "src/Main.purs",
      "end": [
        5,
        50
      ]
    },
    {
      "exportSourceImportedFrom": [
        "Prelude"
      ],
      "exportSourceDefinedIn": [
        "Data",
        "Eq"
      ]
    },
    {
      "TypeClassRef": [
        {
          "start": [
            1,
            1
          ],
          "name": "src/Main.purs",
          "end": [
            5,
            50
          ]
```

PR #3526, opened 2019-01-25
Add support for multiple foreign modules based on codegen targets

jmackie:
Currently only one codegen target (`JS`) supports foreign modules. So it makes sense to pass around a `Map ModuleName FilePath`, where the module in question may or may not have a foreign javascript module associated with it. 

However, if the compiler were to support other non-js backends then `purs compile --codegen js,foo,bar Example.purs` might have to copy not only `Example.js` but also `Example.foo` and `Example.bar`. And so `Map ModuleName FilePath` would have to become `Map ModuleName (Map CodegenTarget FilePath)` (or equivalent), which is the purpose of this PR. 

Note this is motivated by my current efforts to add support for `Go` (https://github.com/purescript-go/purs), where this functionality is needed to have the two codegens exist side-by-side.

Also note this is my first non-trivial contribution, so apologies if I've missed anything!

Keep up the great work guys. PureScript is dope ❤️

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -128,6 +128,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@gabejohnson](https://github.com/gabejohnson) | Gabe Johnson | [MIT license](http://opensource.org/licenses/MIT) |
 | [@dariooddenino](https://github.com/dariooddenino) | Dario Oddenino | [MIT license](http://opensource.org/licenses/MIT) |
 | [@jordanmartinez](https://github.com/jordanmartinez) | Jordan Martinez | [MIT license](http://opensource.org/licenses/MIT) |
+| [@jmackie](https://github.com/jmackie) | Jordan Mackie | [MIT license](http://opensource.org/licenses/MIT) |
 
 ### Contributors using Modified Terms
 
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -8,6 +8,7 @@ module Command.Compile (command) where
 
 import           Control.Applicative
 import           Control.Monad
+import           Control.Monad.Reader (asks)
 import qualified Data.Aeson as A
 import           Data.Bool (bool)
 import qualified Data.ByteString.Lazy.UTF8 as LBU8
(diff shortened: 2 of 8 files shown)


JordanMartinez:
@jmackie Not sure whether you were aware of this, but there are already other codegens besides Javascript. See the list of URLs in [this section](https://github.com/JordanMartinez/purescript-jordans-reference/tree/latestRelease/00-Getting-Started#is-the-price-of-the-steep-learning-curve-worth-the-benefits-of-using-purescript-in-code). None of them are currently included in this repo. It seems your PR might raise the question of whether to include all such codegens in the same repo. 

Also, I'm not a core contributor or someone who is implementing a backend for another language, so I might not be 'qualified' to comment on this. My intent is purely to inform.

jmackie:
@JordanMartinez yeh this is only concerned with supporting multiple foreign modules, which should benefit the other "embedded" backends 🙂

It's probably a bad title, sorry. Naming PRs is hard...

chexxor:
Currently, are all foreign modules (X.purs, X.js, X.go, X.c) copied to the "output" directory, which is where the intermediate compilation code is output?

How does an alternative backend compiler using the intermediate representation (corefn) currently find the FFI files for their backend?

hdgarrood:
The module's path in the filesystem is contained within the corefn output:

```
$ jq < output/Main/corefn.json  .modulePath
"src/Main.purs"
```

hdgarrood:
(So you can do something like `dropEnd ".purs" >>> append ".go"` to that file path)

jmackie:
As I understand it you have two options when adding an alternate backend: 

- Deserialize and compile the `CoreFn` json (e.g. `purescript-native`)
- Patch another codegen target into a fork of `purs` (e.g. `purerl`). 

This PR is aimed at helping with the latter. But if you're saying that the main compiler shouldn't be written with FFI targets other than JS in mind then that's cool.

hdgarrood:
Pretty much, yes. I think this is a relatively low priority right now, but eventually we want to be able to provide enough information in intermediate formats like corefn (and maybe there will be a "coreimp" too one day) that alternative backends don't need to maintain forks of the main compiler, and so that the main compiler doesn't have to worry about generalizing like this.

jmackie:
That seems sensible. FYI it's the lack of annotations in corefn that forced
me down the fork route. If we could get that information out of the
compiler then statically typed backends become much more feasible.

Keep up the good work :)

PR #3528, opened 2019-02-01
Support non-public publishes

LiamGoodacre:
This change adds a `--non-public` flag to `purs publish` which allows the caller to generate a pursuit data file without any of the version/license/github constraints.  I'm wanting to use this to generate the data for a private hosted pursuit instance.

I couldn't find a nice way to do this without changing things.  (Ideas?)

### DIFF
--- a/app/Command/Publish.hs
+++ b/app/Command/Publish.hs
@@ -7,6 +7,7 @@ import qualified Data.Aeson as A
 import qualified Data.ByteString.Lazy.Char8 as BL
 import           Data.Time.Clock (getCurrentTime)
 import           Data.Version (Version(..))
+import           Language.PureScript.Docs
 import           Language.PureScript.Publish
 import           Language.PureScript.Publish.ErrorsWarnings
 import           Options.Applicative (Parser)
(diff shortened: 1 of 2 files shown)


hdgarrood:
I'll review properly later, but in general I think adding more fields to `publishOptions` should be fine. If we're going to work towards supporting internal Pursuit instances though (which I think would be great), I would quite like to try to reconsider the design of Pursuit and the publishing process a little more holistically, if that makes sense? For instance, does it still make sense to require the presence of a version in the case of an internal Pursuit instance, or can we perhaps improve the UX there by allowing different version identifiers such as commit shas?

hdgarrood:
Also, something which has occurred to me is that it might be nice to allow a `docs` target for the `--codegen` option to `purs compile`, which generates the json for docs for just that module, and serializes a `Module` i.e. (the one in `Language.PureScript.Docs`) as `docs.json` alongside the externs, js output, etc. If the `Module` values were easily accessible like this, it might be easier to build this sort of feature on top of it, as you'd be able to get access to most (all?) of the relevant json without having to actually provide all the information to be able to construct a `Package`.

hdgarrood:
(See also #3503)

hdgarrood:
@LiamGoodacre Do you mind if we leave this change in a fork for the time being to allow more exploration? I'd rather not add flags to `purs publish` if we are going to eventually decide to remove or change them.

LiamGoodacre:
@hdgarrood that's fine :+1:
I mainly wanted to start a discussion around this kind of functionality.

Issue #3505, opened 2019-01-06
Naming a constructor `PS` breaks the javascript


chrismshelton:
Sorry if this has already been reported, I searched but I couldn't find anything.

Naming a constructor `PS` breaks the resulting javascript:

```purescript
module Main where                                                                                                                                                                                           
                                                                                                                                                                                                            
import Prelude                                                                                                                                                                                              
import Effect (Effect)                                                                                                                                                                                      
import Effect.Console (log)                                                                                                                                                                                 
                                                                                                                                                                                                            
data P = PS                                                                                                                                                                                                 
                                                                                                                                                                                                            
main :: Effect Unit                                                                                                                                                                                         
main = do                                                                                                                                                                                                   
  log "Hello sailor!"
```

When compiled with `pulp build --to output/all.js`, and loaded onto a page, I get the error:

```
Uncaught TypeError: Cannot read property 'Effect' of undefined
    at PS.Main (all.js:25)
    at all.js:38
```

The produced javascript is:

```
// Generated by purs bundle 0.12.0
var PS = {};
(function(exports) {
    "use strict";

  exports.log = function (s) {
    return function () {
      console.log(s);
      return {};
    };
  };
})(PS["Effect.Console"] = PS["Effect.Console"] || {});
(function(exports) {
  // Generated by purs version 0.12.0
  "use strict";
  var $foreign = PS["Effect.Console"];
  var Data_Show = PS["Data.Show"];
  var Data_Unit = PS["Data.Unit"];
  var Effect = PS["Effect"];
  exports["log"] = $foreign.log;
})(PS["Effect.Console"] = PS["Effect.Console"] || {});
(function(exports) {
  // Generated by purs version 0.12.0
  "use strict";
  var Effect = PS["Effect"];
  var Effect_Console = PS["Effect.Console"];
  var Prelude = PS["Prelude"];                 
  var PS = (function () {
      function PS() {

      };
      PS.value = new PS();
      return PS;
  })();
  var main = Effect_Console.log("Hello sailor!");
  exports["PS"] = PS;
  exports["main"] = main;
})(PS["Main"] = PS["Main"] || {});
PS["Main"].main();
```

I am somewhat surprised by this, but can confirm the following also breaks:

```javascript
      A = 123;                                                                                                                                                                                              
                                                                                                                                                                                                            
      var fn = function() {                                                                                                                                                                                 
        console.log (A);                                                                                                                                                                                    
        var A = 234;
      };                                                                                                                                                                                                    
[...]

PR #3533, opened 2019-02-09
 Naming a constructor `PS` breaks the javascript

mhcurylo:
I have fixed the PS constructor braking the exports by changing the IIFE argument from exports to the PS module object.
There are no automated tests for the psc-bundle and I have not added one. Maybe they should be introduced?
Fragment of an output is:
```
function(PS) {
  // Generated by purs version 0.12.2
  "use strict";
  var exports = PS["Effect.Console"] = PS["Effect.Console"] || {};
  var $foreign = PS["Effect.Console"];
  var Data_Show = PS["Data.Show"];
  var Data_Unit = PS["Data.Unit"];
  var Effect = PS["Effect"];
  exports["log"] = $foreign.log;
})(PS); 
```

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -586,7 +586,7 @@ codeGen optionsMainModule optionsNamespace ms outFileOpt = (fmap sourceMapping o
         JSVariable lfsp
           (cList [
             JSVarInitExpression (JSIdentifier sp nm)
-              (JSVarInit sp $ either require (moduleReference sp . moduleName) req )
+              (JSVarInit sp $ either require (innerModuleReference sp . moduleName) req )
           ]) (JSSemi JSNoAnnot)
       ]
     declToJS (ExportsList exps) = withLength $ map toExport exps
@@ -642,6 +642,12 @@ codeGen optionsMainModule optionsNamespace ms outFileOpt = (fmap sourceMapping o
     JSMemberSquare (JSIdentifier a optionsNamespace) JSNoAnnot
       (str mn) JSNoAnnot
 
+  innerModuleReference :: JSAnnot -> String -> JSExpression
+  innerModuleReference a mn =
+    JSMemberSquare (JSIdentifier a "$PS") JSNoAnnot
+      (str mn) JSNoAnnot
+
+
   str :: String -> JSExpression
   str s = JSStringLiteral JSNoAnnot $ "\"" ++ s ++ "\""
 
@@ -652,17 +658,27 @@ codeGen optionsMainModule optionsNamespace ms outFileOpt = (fmap sourceMapping o
   wrap :: String -> [JSStatement] -> [JSStatement]
   wrap mn ds =
     [
-    JSMethodCall (JSExpressionParen lf (JSFunctionExpression JSNoAnnot JSIdentNone JSNoAnnot
-                                                (JSLOne (JSIdentName JSNoAnnot "exports")) JSNoAnnot
-                                                (JSBlock sp (lfHead ds) lf)) -- \n not quite in right place
+    JSMethodCall (JSExpressionParen lf (JSFunctionExpression JSNoAnnot JSIdentNone JSNoAnnot (JSLOne (JSIdentName JSNoAnnot "$PS")) JSNoAnnot
+                                                (JSBlock sp (addModuleExports ds) lf)) -- \n not quite in right place
                                     JSNoAnnot)
                   JSNoAnnot
-                  (JSLOne (JSAssignExpression (moduleReference JSNoAnnot mn) (JSAssign sp)
-                            (JSExpressionBinary (moduleReference sp mn) (JSBinOpOr sp) (emptyObj sp))))
+                  (JSLOne (JSIdentifier JSNoAnnot optionsNamespace))
                   JSNoAnnot
                   (JSSemi JSNoAnnot)
     ]
     where
+      addModuleExports :: [JSStatement] -> [JSStatement]
+      addModuleExports [] = lfHead moduleExports 
+      addModuleExports (x:xs) = lfHead [x] ++ moduleExports ++ xs
+      moduleExports = 
+        [
+        JSExpressionStatement (JSAssignExpression (innerModuleReference lfsp mn) 
+                                                  (JSAssign sp) 
+                                                  (JSExpressionBinary (innerModuleReference sp mn) (JSBinOpOr sp) (emptyObj sp))) 
+                              (JSSemi JSAnnotSpace), 
+        JSVariable lfsp (JSLOne $ JSVarInitExpression (JSIdentifier sp "exports") $ JSVarInit sp (innerModuleReference sp mn)) 
+                        (JSSemi JSNoAnnot) 
+        ]
       lfHead (h:t) = addAnn (WhiteSpace tokenPosnEmpty "\n  ") h : t
       lfHead x = x
 
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -18,6 +18,7 @@ import qualified TestPrimDocs
 import qualified TestPsci
 import qualified TestIde
 import qualified TestPscPublish
+import qualified TestBundle
 import qualified TestUtils
 
 import System.IO (hSetEncoding, stdout, stderr, utf8)
@@ -35,6 +36,7 @@ main = do
   ideTests <- TestIde.main
   compilerTests <- TestCompiler.main
   psciTests <- TestPsci.main
+  pscBundleTests <- TestBundle.main
   coreFnTests <- TestCoreFn.main
   docsTests <- TestDocs.main
   publishTests <- TestPscPublish.main
@@ -45,6 +47,7 @@ main = do
       "Tests"
       [ compilerTests
       , psciTests
+      , pscBundleTests
       , ideTests
       , coreFnTests
       , docsTests
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -68,6 +68,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@lukerandall](https://github.com/lukerandall) | Luke Randall | [MIT license](http://opensource.org/licenses/MIT) |
 | [@matthewleon](https://github.com/matthewleon) | Matthew Leon | [MIT license](http://opensource.org/licenses/MIT) |
 | [@mcoffin](https://github.com/mcoffin) | Matt Coffin | [MIT license](http://opensource.org/licenses/MIT) |
+| [@mhcurylo](https://github.com/mhcurylo) | Mateusz Curylo | [MIT license](http://opensource.org/licenses/MIT) |
 | [@MiracleBlue](https://github.com/MiracleBlue) | Nicholas Kircher | [MIT license](http://opensource.org/licenses/MIT) |
 | [@mrkgnao](https://github.com/mrkgnao) | Soham Chowdhury | [MIT license](http://opensource.org/licenses/MIT) |
 | [@mgmeier](https://github.com/mgmeier) | Michael Gilliland | [MIT license](http://opensource.org/licenses/MIT) |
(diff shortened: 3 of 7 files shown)


hdgarrood:
I agree that having some tests would be good. Are you volunteering to add some? I'm very happy to help if you have any questions.

mhcurylo:
It sounds like morning fun if we do not mind me taking some time. An outline:
1. Example directory for purs files in tests/purs/bundle directory.
2. A module TestPscBundle.hs in /tests/ directory.
3. Similar setup as the compiler tests but using the psc-bundle.
As an alternative, we could add an annotation 'use-bundle' to the compiler tests and keep the examples in there. I think it might be mixing concerns though.

hdgarrood:
I think a tests/purs/bundle directory sounds great.

hdgarrood:
(Fixes #3505)

mhcurylo:
@hdgarrood 
Test added. 
I have exported the common functions between compiler and bundle tests to TestUtils.
One problem I see is that in order to increase bundle coverage we would have to manually implement js scenarios. With the current solution (test vs compiled purescript code) - some code will be unreachable.
For now, it fixes the problem and adds coverage. 
Please review.

tests/TestPscBundle.hs
```diff
@@ -0,0 +1,97 @@
...
+import System.IO
+import System.IO.UTF8
+import qualified System.FilePath.Glob as Glob
+
+import TestUtils
+import Test.Tasty
+import Test.Tasty.Hspec
+
+main :: IO TestTree
+main = testSpec "psc-bundle" spec
```
hdgarrood: Can you use just "bundle" here? `psc-bundle` doesn't exist any more.

tests/TestPscBundle.hs
```diff
@@ -0,0 +1,97 @@
...
+  -> [FilePath]
+  -> Handle
+  -> Expectation
+assertBundles supportModules supportExterns supportForeigns inputFiles outputFile =
+  assert supportModules supportExterns supportForeigns inputFiles checkMain $ \e ->
+    case e of
+      Left errs -> return . Just . P.prettyPrintMultipleErrors P.defaultPPEOptions $ errs
+      Right _ -> do
+        process <- findNodeProcess
+        -- Mateuszu! Tutaj zrob. Trzeba to zbundlowac zamiast tworzyc Main.main()
```
hdgarrood: If this comment should stay in, can you write it in English please?

tests/TestPscBundle.hs
```diff
@@ -0,0 +1,97 @@
...
+        --app modulesDir entryPoint
```
hdgarrood: Can this comment be removed?

tests/TestPscBundle.hs
```diff
@@ -0,0 +1,97 @@
...
+        bundled <- runExceptT $ do
+          input <- forM jsFiles $ \filename -> do
+            js <- liftIO $ readUTF8File filename
+            mid <- guessModuleIdentifier filename
+            length js `seq` return (mid, Just filename, js) 
+          bundleSM input entryModule (Just $ "Main") "PS" (Just entryPoint)
+        case bundled of
+            Right (_, js) -> do
+              writeUTF8File entryPoint js
+              --writeFile entryPoint "require('Main').main()"
```
hdgarrood: Can this comment be removed?

tests/TestPscBundle.hs
```diff
@@ -0,0 +1,97 @@
...
+                 | not (null out) && trim (last (lines out)) == "Done" -> do
+                     hPutStr outputFile out
+                     return Nothing
+                 | otherwise -> return $ Just $ "Test did not finish with 'Done':\n\n" <> out
+                Just (ExitFailure _, _, err) -> return $ Just err
+                Nothing -> return $ Just "Couldn't find node.js executable"
+            Left err -> return . Just $ "Coud not bundle: " ++ show err
+
+logfile :: FilePath
+logfile = "psc-bundle-tests.out"
```
hdgarrood: Same here re "psc-bundle"

tests/support/pscide/purs.tix
```diff
@@ -0,0 +1 @@
...,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,0,0,0,0,0,0,1,1,2,2,1,1,1,1,1,1,1,1,1,1,0,1,1,1,2,267,267,1,1,1,266,266,0,266,266,267,36,36], TixModule "purescript-0.12.2-KMJAZ5OYzDhGVgSCJVsyOl/Language.PureScript.Sugar.Operators.Binders" 2084133627 28 [1,0,1,1,1,0,2,2,2,0,2,1,1,1,1,1,1,2,1,27,28,3,1,1,1,1,3,3], TixModule "purescript-0.12.2-KMJAZ5OYzDhGVgSCJVsyOl/Language.PureScript.Sugar.Operators" 289809625 774 [0,128,128,0,128,128,0,0,128,128,0,0,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,0,1,1,0,0,1,1,1,1,1,1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,1,1,1,1,1,0,0,0,0,1,0,0,0,1,1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,33,34,0,0,0,14,0,14,14,14,14,14,14,14,14,14,19,19,19,19,24,24,24,24,24,0,18,21,21,24,24,24,3,3,3,0,3,3,3,0,0,0,0,3,3,3,3,3,3,3,3,3,3,3,3,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,6,6,6,6,0,6,6,6,1,3,3,3,3,0,3,3,3,159,158,198,0,150,150,150,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,33,33,0,33,33,33,33,33,33,
```
hdgarrood: Can you remove this file please?

tests/Main.hs
```diff
@@ -43,8 +45,10 @@ main = do
   defaultMain $
     testGroup
       "Tests"
-      [ compilerTests
+      [
```
hdgarrood: Would you remind removing the extra line here?

hdgarrood:
> One problem I see is that in order to increase bundle coverage we would have to manually implement js scenarios. With the current solution (test vs compiled purescript code) - some code will be unreachable.

This doesn't bother me, I think what we have here is fine for now.

tests/TestPscBundle.hs
mhcurylo: This comment was written in Polish, to mark WIP, so that I never ever forget to delete that one and the ones next to it.

mhcurylo:
Thanks for all the minor comments, they are applied.

tests/TestPscBundle.hs
mhcurylo: Should I rename the module and the spec as well, then?

hdgarrood: Yes please!

hdgarrood:
I've now reviewed this properly and it looks good to me, although I have one small suggestion. Instead of e.g.
```
  var exports = PS["Effect.Console"] = PS["Effect.Console"] || {};
```
could we perhaps generate
```
  PS["Effect.Console"] = PS["Effect.Console"] || {};
  var exports = PS["Effect.Console"];
```
which I think should be equivalent, but is a bit clearer?

hdgarrood:
Oh and one more: how about using $PS as the name of the argument for the IIFE function for each module in the output? Then we should be able to avoid collisions with names from PureScript values entirely, rather than relying on slightly arcane details of JS scoping in cases of collisions.

mhcurylo:
All suggestions included in the PR.
I have changed the generated JS. 
I have changed the default namespace to $PS (inside the bundle action in the app) as suggested.
It would not solve the issue by itself - a namespace provided to the bundler could still interfere with a constructor if they shared names.
Thus I have changed the name of the test and included a pragma to configure namespace provided to the bundler so that the test is still testing what it should (the case where the namespace is the same as the constructor).

hdgarrood:
Sorry, I should have been clearer: what I mean is that the default bundle namespace should continue to be PS, but the name given to the argument of each iife for each module should be something which couldn’t clash with any purescript name. So for example you might get

```javascript
var PS = PS || {};
function($PS) {
  // Generated by purs version 0.12.2
  "use strict";
  var exports = $PS["Effect.Console"] = $PS["Effect.Console"] || {};
  var $foreign = $PS["Effect.Console"];
  var Data_Show = $PS["Data.Show"];
  var Data_Unit = $PS["Data.Unit"];
  var Effect = $PS["Effect"];
  exports["log"] = $foreign.log;
})(PS);
```

So the namespace object is always referred to as $PS inside each iife, regardless of the namespace CLI option. I’d prefer to leave the tests as they were before.

src/Language/PureScript/Bundle.hs
```diff
@@ -638,17 +644,20 @@ codeGen optionsMainModule optionsNamespace ms outFileOpt = (fmap sourceMapping o
...
-                            (JSExpressionBinary (moduleReference sp mn) (JSBinOpOr sp) (emptyObj sp))))
+                  (JSLOne (JSIdentifier JSNoAnnot optionsNamespace))
                   JSNoAnnot
                   (JSSemi JSNoAnnot)
     ]
     where
+      addModuleExports :: [JSStatement] -> [JSStatement]
+      addModuleExports [] = lfHead moduleExports 
+      addModuleExports (x:xs) = lfHead [x] ++ moduleExports ++ xs
+      moduleExports = [JSExpressionStatement (JSAssignExpression (innerModuleReference lfsp mn) (JSAssign sp) (JSExpressionBinary (innerModuleReference sp mn) (JSBinOpOr sp) (emptyObj sp))) (JSSemi JSAnnotSpace), JSVariable lfsp (JSLOne $ JSVarInitExpression (JSIdentifier sp "exports") $ JSVarInit sp (innerModuleReference sp mn)) $ JSSemi JSNoAnnot ]
```
hdgarrood: Last nitpick and then I think this is good to merge: this line is too long. Can you please put each element of this list on its own line and use indentation similar to how it has been used elsewhere in this file (such as in the main body of `wrap` just above here)?

mhcurylo:
Oh, that sounds reasonable.
All changes are reverted/made.

The only difference to the example is in the namespace declaration.
Instead of the proposed
```
var PS = PS || {};
```
it is
```
var PS = {};
```
as it was before.
The semantics are different when a namespace is shared across bundles.
Which one should we go with?

hdgarrood:
Yes, thanks, I think the initial namespace declaration should stay as it is, `var PS = {};`. (I couldn't be bothered to check how it is currently and so I made a guess which obviously has turned out to be incorrect.)

Did you make a commit locally which you haven't pushed yet? It looks like there aren't any new commits to this branch since my last review comment.

mhcurylo:
I have rebased and pushed.
Everything is up to date.

hdgarrood:
The `moduleExports` declaration doesn't seem to have changed?

mhcurylo:
Thanks for noticing.
Somewhat I did not read the moduleExports comment.
Now it is addressed.

mhcurylo:
Thanks!

PR #3534, opened 2019-02-10
Simplify docs re-export handling

hdgarrood:
Follows on from #3521 by removing the use of the Env's imports and
instead obtaining the same information from the Env's exports. By
removing the need for the Imports values, this paves the way for using
externs files to speed up docs generation and make it more resilient
(see #3503). Additionally, it should offer a small performance boost
immediately.

This code can probably still be simplified and optimised a bit more. For
example, we could use a more sensible data structure for storing
Declaration values as they are constructured, in preparation for later
lookups while we are adding re-exports.

### DIFF
--- a/src/Language/PureScript/Docs/Convert/ReExports.hs
+++ b/src/Language/PureScript/Docs/Convert/ReExports.hs
@@ -72,8 +72,8 @@ getReExports env mn =
   case Map.lookup mn env of
     Nothing ->
       internalError ("Module missing: " ++ T.unpack (P.runModuleName mn))
-    Just (_, imports, exports) -> do
-      allExports <- runReaderT (collectDeclarations imports exports) mn
+    Just (_, _, exports) -> do
+      allExports <- runReaderT (collectDeclarations exports) mn
       pure (filter notLocal allExports)
 
   where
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Docs/Convert/ReExports.hs
```diff
@@ -123,72 +122,22 @@ collectDeclarations imports exports = do
...
     -> m (Map P.ModuleName [b])
-  collect lookup' imps exps = do
-    imps' <- traverse (findImport imps) $ Map.toList $ fmap P.exportSourceDefinedIn exps
-    Map.fromListWith (<>) <$> traverse (uncurry lookup') imps'
+  collect lookup' exps =
+    exps
+    |> Map.mapMaybe P.exportSourceImportedFrom
+    |> Map.toList
+    |> traverse (uncurry (flip lookup'))
+    |> fmap (Map.fromListWith (<>))
```
garyb: I kinda prefer the way this was written before, mainly because this would be the only place in the whole compiler that this style is used that I'm aware of. (The other part is because I'm one of those weird people that finds this ordering harder to parse than it being "backwards" as "normal" :smile:).

hdgarrood: Yes, totally fair point, I'll use something closer to the older style :)

Issue #3104, opened 2017-10-04
"No type class instance was found" error caused by type synonym


Unisay:
```
module Main where

import Data.Typelevel.Num (class LtEq, type (:*), D2, D5, D6, D8)

type D256 = D2 :* D5 :* D6

class (LtEq a D256) <= Lte256 a -- doesn't compile, error is below

instance lte256 :: Lte256 D8

-- class (LtEq a (D2 :* D5 :* D6)) <= Lte256 a   
-- ^^^ Expanding type synonym fixes the compilation error

{-

Error found:
in module Main
at src/Main.purs line 9, column 1 - line 9, column 27

  No type class instance was found for

    Data.Typelevel.Num.Sets.Pos t0

  The instance head contains unknown type variables. Consider adding a type annotation.

while checking that expression #dict LtEq
  has type {}
in value declaration lte256

where t0 is an unknown type

-}
```

Issue #3391, opened 2018-06-27
Remove type synonyms in instance head restriction


natefaubion:
A common pattern (for me at least)
```
type Environment = { ... }

instance askMyApp :: MonadAsk Environment MyApp where
  ...
```

But this doesn't work because type synonyms aren't allowed. Inlining the type is often not practical. The solutions are to newtype it, or unnecessarily use TypeEquals and coercions. I don't think this restriction accomplishes much. The only reason it exists is to temper expectations around allowable instances (type synonyms aren't nominal), but it does it poorly and the error does not even indicate why it's a problem.

garyb:
I used to be against this, but I think I've come around to it.

paf31:
The obvious place to allow it is under a fundep, right?

natefaubion:
> The obvious place to allow it is under a fundep, right?

That's where it is most useful, I'd think. Doesn't that complicate synonym desugaring though? What exactly are we trying to guard against with this restriction and the error it presents? Are the cases already solved by other errors (orphan check)?

hdgarrood:
It's worth noting that in GHC, `FlexibleInstances` implies `TypeSynonymInstances`: https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-FlexibleInstances

We essentially have `FlexibleInstances` and `FlexibleContexts` in PureScript, too. In fact, we even have [`UndecidableInstances`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#undecidable-instances): we've dropped the requirement that instance resolution will provably terminate, and instead kept track of a certain amount of "fuel" to bail out after a given number of instance resolution steps.

So I agree with @natefaubion in that I think we should consider dropping the requirement totally: allow type synonyms to appear anywhere in both instances and contexts, desugar them before instance resolution, and use the existing instance rules to decide whether the resulting instances are valid.

rhendric:
Fixed in #3539, I believe.

PR #3539, opened 2019-02-24
Allow instances for synonyms

### LINKED ISSUE #2529: Allow synonyms in some instance heads?
I think there may be two places where it may be reasonable to accept synonyms in instance heads:

1. If the synonym is being applied to a type
2. If the class variable is fully determined by another

I have a case where both these conditions coincide:

``` purescript
type Cursor = Mu CursorF

newtype X = X (Cofree MyF Cursor)

instance recursiveX ∷ Recursive X (EnvT (Mu CursorF) MyF) where ...
```

So here I have to use `Mu CursorF` in `recursiveX`, but with either of the rules above I'd be able to use `Cursor` in there instead.

The main reason we don't allow synonyms in instances is so there's no confusion if people try to do that directly for a `type` rather than using `newtype`, right? If so, I think the above cases still preserve that, and make some fancier use cases more convenient.

Thoughts?

garyb:
Resolves #2529 at least

Can expand it to just allow instances wherever if we prefer, in which case it'll deal with #3391 too.


garyb:
So yeah, allowing it in this situation should be uncontroversial - but if we did expand it, what are the pros/cons? I can only think of one primarily for each case:

- Pro: there are probably some situations that the determined var isn't enough, 
- Con: it might confuse people sometimes if they don't realise they're trying to define an instance for a synonym

The con will mostly that will manifest as an overlapping instance error. I guess ideally we'd include a suggestion that spots when an overlap involves a synonym and mentions that might be why, but I'm not sure we have enough info to do that when the overlap is detected.

garyb:
Hah, if I looked at the code I just modified I'd see quite easily that we could indeed easily determine it here :smile:

I'll see about improving the error in this case first, as it's still potentially helpful even if we keep the fundep restriction, as overlaps can still arise accidentally.

garyb:
Hmm, it'll need a little more thought actually - we'd have to store the un-expanded types for instances in the environment too, to be able to report that suggestion for sure - it'd be order dependent.

hdgarrood:
Would it be sufficient to pass `zip tys tys'` to `checkOverlappingInstance` instead of just `tys'`?

hdgarrood:
Oh, I suppose not, because `checkOverlappingInstance` doesn't actually perform the entire overlapping instance check.

Regardless of which route we take though (i.e. whether we only allow determined arguments or whether we just allow everything) I think it would be best to not to do this until we can add a hint for the case where an overlap arises and type synonyms are involved.

natefaubion:
> Regardless of which route we take though (i.e. whether we only allow determined arguments or whether we just allow everything) I think it would be best to not to do this until we can add a hint for the case where an overlap arises and type synonyms are involved.

I think a case could be made that this is a problem in general, and we don't really have a way to reliably report any issues that might have arised due to synonyms. I think I would like to just go ahead and get this in since it's a very useful feature. What do you think?

hdgarrood:
Yeah on reflection I agree, I'd like to get this in too. My preference would be to drop the instance synonym restriction entirely, and allow synonyms anywhere in instances as long as they are fully applied.

garyb:
:+1: I'll remove the determined-by-fundep check and it should be good to go then.

natefaubion:
@garyb Is this something you intend to fix up?

garyb:
Wow there, what's the hurry? It's only just been WIP for a year. 😆 

Yeah, sure I'll fix it up. I hadn't forgotten about it actually, just never quite get around to it.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -193,7 +193,7 @@ checkTypeClassInstance cls i = check where
     TypeLevelString _ _ -> return ()
     TypeConstructor _ ctor -> do
       env <- getEnv
-      when (ctor `M.member` typeSynonyms env) . throwError . errorMessage $ TypeSynonymInstance
+      when (ctor `M.member` typeSynonyms env && not isFunDepDetermined) . throwError . errorMessage $ TypeSynonymInstance
```
thomashoneyman, line 196: Is removing this `&& not isFunDepDetermined` the only change necessary to clean up the PR for 0.14?

garyb:
@thomashoneyman I saw that :wink:

Yes, but it also needs a merge, I'm going to do it now.

thomashoneyman:
😀I was still thinking when I reflexively hit the comment button, but I didn’t want to bug you.

garyb:
I was getting a different error when I ran the tests locally, so wanted to throw this at CI and see how it behaved there. Figured it out now though, I needed to update my test support dependencies locally.

thomashoneyman:
Can this be added to the v0.14.0 milestone to track it there? Or is that milestone not being curated?

hdgarrood:
At this point I don’t want to add anything to 0.14.0 unless it’s a showstopping bug. In general I would prefer to only add issues to milestones, as if you have PRs in milestones too then you’re effectively counting things twice. Not adding this to 0.14.0 doesn’t mean it definitely won’t get merged in time for 0.14.0, it just means that it might not get merged if everything else in 0.14.0 is ready to go and it’s time to make a release.

garyb:
Good spot! It does allow bad things like that to be defined currently.

rhendric:
Is @hdgarrood's objection the only blocker here? Isn't it very easily resolved by just swapping the `checkTypeClassInstance` line with its successor (and adding a prime to the `tys'` in it)? I tried that locally and it seems to work fine.

hdgarrood:
I think that ought to do it, yes. I think this should be good to merge once that change has been made and a test has been added for the example I gave.

rhendric:
I've pushed those changes to @garyb's branch; I hope that wasn't rude! It seemed not worth making a new PR for such a small addition and you did say in #3940 that you didn't mind someone else carrying this over the finish line.

hdgarrood (approved):

kl0tl:
Should we try to fix https://github.com/purescript/purescript/issues/3104 with this pull request? It seems that replacing type synonyms at the top of the solver loop is enough. Should we also replace synonyms in type classes contexts?

hdgarrood:
I think it might be nice to address #3104 in a separate PR, since I’d consider it a separate issue (that bug exists whether or not we allow synonyms in instance heads) and since it requires changes in a different part of the compiler.

kl0tl:
My reasoning was that allowing type instances for synonyms makes this bug easier to stumble on:

```purs
data D
type S = D

class C0 a
class C0 a <= C1 a

instance c0 :: C0 D
instance c1 :: C1 S
```

```
Error found:
in module Example
at Example.purs:16:1 - 16:20 (line 16, column 1 - line 16, column 20)

    Example.C0 S
                

while checking that expression #dict C0
  has type Record (() @Type)
in value declaration c1

```

but sure, I can open another pull request.

hdgarrood:
Right, yes, it’s definitely easier to run into that bug after this change, so I agree that it would be good to address it soon after this. Thanks for spotting this!

garyb:
@rhendric thanks for picking it up, and no it wasn't rude at all :smile:

garyb:
(I can't give this the double-approval since I opened the PR)

thomashoneyman (approved):

hdgarrood:
This PR and #3936 both added a test called TypeSynonyms6, so I've resolved the conflict by giving each of them a slightly more descriptive name.

kl0tl:
I opened https://github.com/purescript/purescript/pull/3965 and https://github.com/purescript/purescript/pull/3966 to resolve some issues with the interaction of type synonyms, superclass constraints and derived newtype instances.

PR #3541, opened 2019-02-25
[purs ide] Treat module declarations like any other

kRITZCREEK:
This means we can now complete module names with the completion API as well as being able to query for module level documentation and goto-definition for module names.

~TODO: Still need to add regression tests~


kRITZCREEK:
Thanks for checking @hdgarrood!

@nwolverson What do you think about this way of going about it? You'll basically need to add a namespace filter with `["value", "type", "kind"]` to all completion requests to get the old behaviour (which is also backwards compatible). 

But when we're doing completions in imports we can now use the completion command, filter it down to the module namespace and then have module level documentation in the completion popups. This should also allow me to deprecate the "list modules" command.

nwolverson:
Sure, so long as you do the usual slow deprecation of "list modules" that sounds good, any usage of that can be replaced by a query here. (I do seem to be using the module list for something other than module completion, and it also seems to be possibly dead code so looks like I have some tidying up to do in future...)

PR #3543, opened 2019-02-27
Truncate types in errors, fixes #3401

### LINKED ISSUE #3401: Types printed in error messages should probably have a maximum length
Extremely long error messages can take minutes to print to the terminal and the useful bits at the top usually end up getting dropped.

Can work around by piping to a file: `pulp build 2> error.log`

hdgarrood:
Fixes #3401 

This commit truncates types in error messages below a depth of 3 levels
in the type AST. If the --verbose-errors flag is provided, the
truncation depth is instead set to 1000 (the same as with
pretty-printing of values). Suggestions and docs code is not affected,
since in both of these cases, we always want to print the entire type.

In the process, I've done a minor refactoring of the PrettyPrintType
type to change how rows and records are represented, to make them a
little easier to deal with.

### DIFF
--- a/src/Language/PureScript/Docs/RenderedCode/RenderType.hs
+++ b/src/Language/PureScript/Docs/RenderedCode/RenderType.hs
@@ -35,18 +35,16 @@ typeLiterals = mkPattern match
     Just $ maybe (syntax "_") (syntax . ("?" <>)) name
   match (PPTypeVar var) =
     Just (typeVar var)
-  match (PPRecord row) =
+  match (PPRecord labels tail_) =
     Just $ mintersperse sp
               [ syntax "{"
-              , renderRow row
+              , renderRow labels tail_
               , syntax "}"
               ]
   match (PPTypeConstructor n) =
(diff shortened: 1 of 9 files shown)


hdgarrood:
Example output for `TypesDoNotUnify` after this change:
```
while checking that expression <snip>
  has type { clientId :: Validated (Maybe (... ...))                       
           , description :: String                                         
           , extraInfo :: { productGroup :: Maybe ProductGroup             
                          }                                                
           , images :: Array                                               
                         { file :: ... ...                                 
                         , id :: FileId                                    
                         }                                                 
           , increment :: Validated String                                 
           , manufacturer :: { data :: { addressId :: ... ...              
                                       , vendorId :: ... ...               
                                       }                                   
                             , isCustom :: Boolean                         
                             }

hdgarrood:
Thanks for the review @LiamGoodacre! I think I'll just leave this open a few more days before merging to give anyone else an opportunity to object if they want to.

PR #3547, opened 2019-02-28
[purs ide] Represent Filters as a datatype rather than functions

kRITZCREEK:
This allows us to reorder the filtering to always filter by module name first and exploit the fact that we store our declarations in a Map of a module names.

Also cleans up the declaration type filter implementation (I don't know of any editor plugin that was using these so far so I decided to break the protocol). If I'm wrong please let me know.

### DIFF
--- a/psc-ide/DESIGN.org
+++ b/psc-ide/DESIGN.org
@@ -165,9 +165,11 @@
     the PROTOCOL.md file.
     
 *** Filters
-    Filters are functions of type =Map ModuleName [IdeDeclaration] -> Map
-    ModuleName [IdeDeclaration]=. We keep the =Map= structure around to make the
-    common case of filtering by module names fast.
+    Filters are functions of type =Map ModuleName [IdeDeclaration] ->
+    Map ModuleName [IdeDeclaration]=. They only ever keep or remove
+    declarations, they never modify or add them. We keep the =Map=
+    structure around to make the common case of filtering by module
+    names fast. Filters are commutative.
 
 *** Matchers
     Matchers operate on individual declarations rather than a =Map=. They also
--- a/src/Language/PureScript/Ide/Filter.hs
+++ b/src/Language/PureScript/Ide/Filter.hs
@@ -15,7 +15,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Language.PureScript.Ide.Filter
-       ( Filter(..)
+       ( Filter
        , moduleFilter
        , namespaceFilter
        , exactFilter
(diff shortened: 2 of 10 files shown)


psc-ide/DESIGN.org
```diff
@@ -165,9 +165,11 @@
     the PROTOCOL.md file.
     
 *** Filters
-    Filters are functions of type =Map ModuleName [IdeDeclaration] -> Map
-    ModuleName [IdeDeclaration]=. We keep the =Map= structure around to make the
-    common case of filtering by module names fast.
+    Filters are functions of type =Map ModuleName [IdeDeclaration] ->
```
hdgarrood: Perhaps this is a bit pedantic of me, but I think it's probably safer to say that filters can be thought of as functions of that type, rather than actually being functions of that type. I think there are lots of functions of that type which aren't expressible as filters, or which wouldn't be commutative.

src/Language/PureScript/Ide/Filter.hs
```diff
@@ -15,112 +15,130 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
 module Language.PureScript.Ide.Filter
-       ( Filter
-       , declarationTypeFilter
-       , namespaceFilter
+       ( Filter(..)
```
hdgarrood: Unless I'm missing something I think you can continue not to export the constructor?

src/Language/PureScript/Ide/Filter.hs
```diff
@@ -15,112 +15,130 @@
...
+import qualified Data.Set as Set
+import qualified Data.Map as Map
+import           Language.PureScript.Ide.Filter.Declaration (DeclarationType, declarationType)
 import           Language.PureScript.Ide.Types
 import           Language.PureScript.Ide.Util
 import qualified Language.PureScript           as P
 
-newtype Filter = Filter (Endo [Module])
-  deriving (Semigroup, Monoid)
+newtype Filter = Filter (Either (Set P.ModuleName) DeclarationFilter)
```
hdgarrood: Have you considered having `Filter` be a `newtype` around `[Either (Set P.ModuleName) DeclarationFilter]`? Then I think you could still have a monoid instance which does something similar to `optimizeFilters`. Just a suggestion; it might be nice for consumers of this API not to have to worry about the difference between a single filter and a collection of filters.

kRITZCREEK: I did, I just wanted to keep the model closer to the protocol here. I'm thinking about making a new command which makes more assumptions about how editors end up asking for completions which could change things around a bit more.

PR #3552, opened 2019-03-06
Remove spurious doc comment on CoreFn Module

jmackie:
There have clearly been several changes since this comment was written, and is now incorrect. Best to remove it.


hdgarrood:
I'm not very familiar with this part of the compiler; can you explain why it's incorrect?

natefaubion:
I think everything about the names is wrong, and this comment is definitely in the wrong place. I think it's important to document that we are not serializing types, but that should probably go in ToJSON/FromJSON.

rhendric:
There are no types in CoreFn to serialize. If that fact needs to be called out explicitly, here is as good a place as any, but then the comment would just be `-- The CoreFn representation does not contain type information.`. Personally, I think @jmackie is correct: this comment is obsolete and can be dropped with prejudice.

rhendric (approved):

hdgarrood (approved):
Ok, that's good enough for me.

PR #3556, opened 2019-03-08
Hashable with hashWithSeed

fehrenbach:
Matches purescript/purescript-prelude#198 i.e. Hashable with hashWithSalt.

This passes the incoming salt to the recursive calls. Maybe it should chain calls instead. Again, the hash function itself is easily changed.


fehrenbach:
Closing in favour of #3446.

PR #3560, opened 2019-03-12
Updates for GHC 8.6.4

kRITZCREEK:
Having to propagate the `MonadFail`'s this far is a bit unfortunate but I figured I wouldn't do any refactoring and just do the dumbest updates possible for now.

EDIT:

The network updates broke the networking code in `Command.Ide`, I'll look into that.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -68,6 +68,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@lukerandall](https://github.com/lukerandall) | Luke Randall | [MIT license](http://opensource.org/licenses/MIT) |
 | [@matthewleon](https://github.com/matthewleon) | Matthew Leon | [MIT license](http://opensource.org/licenses/MIT) |
 | [@mcoffin](https://github.com/mcoffin) | Matt Coffin | [MIT license](http://opensource.org/licenses/MIT) |
+| [@mhcurylo](https://github.com/mhcurylo) | Mateusz Curylo | [MIT license](http://opensource.org/licenses/MIT) |
 | [@MiracleBlue](https://github.com/MiracleBlue) | Nicholas Kircher | [MIT license](http://opensource.org/licenses/MIT) |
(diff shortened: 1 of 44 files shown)


hdgarrood:
At least for the docs stuff, I wasn't intending to use MonadFail (in fact I think it's a huge antipattern), so unless there's a particular reason we want to update to GHC 8.6.x very soon, I'd rather refactor so that these sorts of changes won't be necessary and then upgrade; does that sound okay?

kRITZCREEK:
@hdgarrood For sure! I don't intend on making changes in that area anymore and you've got push permissions for this branch, so knock yourself out :)

hdgarrood:
Ah I see: most of the `MonadFail` constraints are actually arising from inside the type checker, and they're just propagating out to everywhere else.

Since most of them seem to be due to pattern matches like
```haskell
  expr@(TypedValue _ value ty) <- infer ...
```
how do we feel about having `infer` return something like
```haskell
data InferredValue = InferredValue
  { inferredCheck :: Bool
  , inferredValue :: Expr
  , inferredType :: Type
  , inferredExpr :: Expr
  }
```
where the `inferredExpr` field is always a `TypedValue` whose fields are the same as the other three fields? Then we can replace the above with
```
  InferredValue _ value ty expr <- infer ...
```
and we can recover the current `infer` with `fmap inferredExpr . infer`.

hdgarrood:
Actually how about we just set `-XNoMonadFailDesugaring` for now (which is the same as what we have now), and refactor so that we aren't using `fail` in a later PR? If we want, we could put it off until we upgrade to GHC 8.8 (at the latest).

kRITZCREEK:
I guess I'd be fine with setting that flag. We could also just make a function that does that pattern match and throws an `internal error` when it fails (as that's essentially the behaviour we've got right now).

hdgarrood:
Yep, that should be fine too. I think I prefer the `InferredValue` approach though, since the compiler can help you avoid making mistakes that way.

kRITZCREEK:
Looks good to me! Sorry for not getting to this myself.

hdgarrood:
No worries at all :)

PR #3562, opened 2019-03-15
Remove more dead code during bundling

### LINKED ISSUE #3551: DCE misses members shadowed by local variables, members used internally only
Consider the following program:

```purs
module Main where

import Prelude
import Data.HeytingAlgebra (ff) 
import Effect (Effect) 
import Effect.Console (log)

main :: Effect Unit
main = log $ show $ (ff :: Boolean)     
```

When put through `pulp build -t`, the resulting bundle includes this subset of `Data.HeytingAlgebra`:

```js
  var $foreign = PS["Data.HeytingAlgebra"];                        
  var HeytingAlgebra = function (conj, disj, ff, implies, not, tt) {
      this.conj = conj;
      this.disj = disj;
      this.ff = ff;
      this.implies = implies;
      this.not = not;
      this.tt = tt;
  };
  var tt = function (dict) {
      return dict.tt;
  };
  var not = function (dict) {
      return dict.not;
  };
  var implies = function (dict) {
      return dict.implies;
  };
  var ff = function (dict) {
      return dict.ff;
  };
  var disj = function (dict) {
      return dict.disj;
  };
  var heytingAlgebraBoolean = new HeytingAlgebra($foreign.boolConj, $foreign.boolDisj, false, function (a) {
      return function (b) {
          return disj(heytingAlgebraBoolean)(not(heytingAlgebraBoolean)(a))(b);
      };
  }, $foreign.boolNot, true);
  var conj = function (dict) {
      return dict.conj;
  };
  exports["HeytingAlgebra"] = HeytingAlgebra;
  exports["tt"] = tt;
  exports["ff"] = ff;
  exports["implies"] = implies;
  exports["conj"] = conj;
  exports["disj"] = disj;
  exports["not"] = not;
  exports["heytingAlgebraBoolean"] = heytingAlgebraBoolean;
```

However, the exported `HeytingAlgebra`, `tt`, `implies`, `conj`, `disj`, and `not` are not used anywhere else in the bundle, and thus those exports could be eliminated. Additionally, `tt`, `implies`, and `conj` are not even used internally to the `Data.HeytingAlgebra` subset, but the bundler thinks they are because variables with the same names are used in the `HeytingAlgebra` constructor. Those functions could also be eliminated.

Would you consider a PR for one or both of these possible optimizations?

rhendric:
Fixes #3551.

---

I don't know whether defining the `Visibility` type was worth it versus continuing to use a `Bool`—IMO the code is slightly more readable this way, since that value is propagating across more code now, but there's no functional need for a new abstraction here. It'd be an easy change for me to make if you'd prefer going back to `Bool`.

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -28,9 +28,9 @@ import Data.Aeson ((.=))
 import Data.Array ((!))
 import Data.Char (chr, digitToInt)
 import Data.Foldable (fold)
-import Data.Generics (GenericM, everything, everywhere, gmapMo, mkMp, mkQ, mkT)
+import Data.Generics (GenericM, everything, everythingWithContext, everywhere, gmapMo, mkMp, mkQ, mkT)
 import Data.Graph
-import Data.List (stripPrefix)
+import Data.List (stripPrefix, (\\))
 import Data.Maybe (catMaybes, fromMaybe, mapMaybe)
 import Data.Version (showVersion)
 import qualified Data.Aeson as A
@@ -91,9 +91,14 @@ guessModuleIdentifier filename = ModuleIdentifier (takeFileName (takeDirectory f
     guessModuleType "foreign.js" = pure Foreign
     guessModuleType name = throwError $ UnsupportedModulePath name
 
--- | A piece of code is identified by its module and its name. These keys are used to label vertices
--- in the dependency graph.
-type Key = (ModuleIdentifier, String)
+data Visibility
+  = Public
+  | Internal
+  deriving (Show, Eq, Ord)
+
+-- | A piece of code is identified by its module, its name, and whether it is an internal variable
+-- or a public member. These keys are used to label vertices in the dependency graph.
+type Key = (ModuleIdentifier, String, Visibility)
 
 -- | An export is either a "regular export", which exports a name from the regular module we are in,
 -- or a reexport of a declaration in the corresponding foreign module.
@@ -115,7 +120,7 @@ data ExportType
 -- into the output during codegen.
 data ModuleElement
   = Require JSStatement String (Either String ModuleIdentifier)
-  | Member JSStatement Bool String JSExpression [Key]
+  | Member JSStatement Visibility String JSExpression [Key]
   | ExportsList [(ExportType, String, JSExpression, [Key])]
   | Other JSStatement
   | Skip JSStatement
@@ -133,10 +138,10 @@ instance A.ToJSON ModuleElement where
                , "name"       .= name
                , "targetPath" .= targetPath
                ]
-    (Member _ public name _ dependsOn) ->
+    (Member _ visibility name _ dependsOn) ->
       A.object [ "type"       .= A.String "Member"
                , "name"       .= name
-               , "visibility" .= A.String (if public then "Public" else "Internal")
+               , "visibility" .= show visibility
                , "dependsOn"  .= map keyToJSON dependsOn
                ]
     (ExportsList exports) ->
@@ -154,9 +159,10 @@ instance A.ToJSON ModuleElement where
 
     where
 
-    keyToJSON (mid, member) =
-      A.object [ "module" .= mid
-               , "member" .= member
+    keyToJSON (mid, member, visibility) =
+      A.object [ "module"     .= mid
+               , "member"     .= member
+               , "visibility" .= show visibility
                ]
 
     exportToJSON (RegularExport sourceName, name, _, dependsOn) =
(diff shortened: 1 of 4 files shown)


hdgarrood:
I think using a `Visibility` type is a great idea. I'd like to wait until #3533 is in before reviewing this, because that PR introduces tests for `purs bundle`, and I'd be much happier making this sort of change if we had tests to help avoid accidental regressions.

rhendric:
Is #3533 stalled? It would be nice if this could be reviewed/hopefully accepted prior to the next release in two weeks.

hdgarrood:
No, I left it open so that other maintainers had time to take a look. Now that it's in, though, would you mind adding some basic tests? I'm more interested in testing that we don't accidentally make it too aggressive, i.e. that we don't remove anything which is needed. If you can think of a good way to add tests which ensure that values which may be removed do get removed, which won't be brittle in the sense of being over-sensitive to any future changes in the code we might produce in `purs bundle`, that would be good too. (Perhaps we can do that with some JS code

rhendric:
Happy to add tests. But the test added in #3533 already tests that DCE is not too aggressive, albeit with a very simple program. Do you want me to add another test with a program of similar simplicity, or something more complicated—and if more complicated, are there particular edge cases that you're concerned about? The space of possible ways that DCE could be overaggressive seems dauntingly large to me.

(It would add several minutes to the build to do this, but I wonder if you might prefer having the bundler tests rerun all of the compiler test cases—that strikes me as the only way to get confidence that the bundler preserves the correctness of any code for which we have confidence that the compiler handles correctly.)

rhendric:
I've added a test here that checks that the things this PR intends to remove are removed. The tests you care more about are over in #3579, assuming you think that's a good way of doing things, that is. (I made a separate PR because adding those tests fails the build due to a preexisting bug that I think should be sorted out separately.)

rhendric:
I've given this a (trivial) rebase; care to take another look?

hdgarrood:
Yes - thanks for your patience, I’ll try to give it a look in the next few days :)

src/Language/PureScript/Bundle.hs
```diff
@@ -411,26 +422,20 @@ compile modules entryPoints = filteredModules
     where
     -- | Create a set of vertices for a module element.
     --
-    -- Some special cases worth commenting on:
-    --
-    -- 1) Regular exports which simply export their own name do not count as dependencies.
-    --    Regular exports which rename and reexport an operator do count, however.
```
hdgarrood: Is it safe to get rid of the handling of operators like this? Also are there any tests exercising this case?

src/Language/PureScript/Bundle.hs
```diff
@@ -202,24 +207,30 @@ withDeps (Module modulePath fn es) = Module modulePath fn (map expandDeps es)
...
       , Just nm' <- fromStringLiteral nm
       , Just mid <- lookup mn' imports
-      = [(mid, nm')]
-    toReference (JSIdentifier _ nm)
-      | nm `elem` boundNames
-      = [(m, nm)]
-    toReference _ = []
+      = ([(mid, nm', Public)], bn)
+    toReference (JSIdentifier _ nm) bn
+      | nm `elem` bn
```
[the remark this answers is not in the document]
rhendric: Will comment; this guard is half of the point of this PR, so it needs to be immediately understandable!

How would this read to you?
> `-- only add a dependency if this name is still in the list of names bound to the module level (i.e., hasn't been shadowed by a function parameter)`

Or would it be clearer if `boundNames`/`bn` got more descriptive names—`moduleBoundNames`/`namesStillBoundToModule`, perhaps? The style in use here seems to eschew overly verbose variable names, which is why I went for the terse option, but clearly that was at the expense of immediate comprehension.

src/Language/PureScript/Bundle.hs
rhendric: I'll try to cook up a test. The change here though is not to the handling of operators, but to the handling of everything else—now all exports count as dependencies, whether they export their own name or not. I hypothesize that the previous behavior was only needed because there was no distinction between `Public` and `Internal` symbols; now that there is, every export is a `Public` symbol and if that export depends on an `Internal` symbol, regardless of any difference in the names of those symbols, the dependency is tracked.

rhendric: Actually, given that explanation I'm not sure a test of re-exporting operators with different names is necessary, but I'll add it if you still think it is.

rhendric:
> For example, I think one thing which would be quite useful would be a `--debug` flag we could pass to `purs bundle` which would print out a summary of this data as json to stderr. How would you feel about looking into that first?

That seems like something I could probably bang out. Does it have to be a separate PR? This PR would necessarily conflict with that one, I think, so that would be a little annoying to manage unless you're willing to merge this one first. Or I can add it as a separate commit here.

src/Language/PureScript/Bundle.hs
hdgarrood: Thanks, yes that makes sense. I’ll have a think and dig through the existing tests and get back to you.

hdgarrood:
Sorry to be a pain, but yeah I would quite like to have the —debug thing separately, so that I can compare the before/after of this change to make sure I’ve really understood it.

src/Language/PureScript/Bundle.hs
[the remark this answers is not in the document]
hdgarrood: The issue was not me not knowing what `bn` meant, but rather me (incorrectly) thinking it could be done away with now that we are tracking visibility in the graph. That comment would be a great addition, thanks!

rhendric:
Should the `--debug` output include the various JavaScript AST fragments that live inside the `Module` type, or should those be elided? I can see those being potentially useful in debugging, but they'll definitely add a ton of bloat to the output.

hdgarrood:
Shall we start by eliding them? I expect that cross referencing with the input files will probably be sufficient?

rhendric:
I've rebased atop #3666 and added the bound name guard comment. Travis is failing for glitchy reasons—the macOS build can't find a `stack`, and the Windows build timed out.

rhendric:
Gentle nudge! Code is rebased again and the build passed on two out of three OSes (Windows timed out again while building dependencies).

hdgarrood:
Thanks -- yeah don't worry, I haven't forgotten about this. CI is supposed to cache dependencies so this doesn't happen but that seems to have stopped working recently; also on my to-do list. :sweat:

hdgarrood:
CI should be fixed now :) if you rebase again (sorry) hopefully it will pass. I'm pulling your branch down to play with this now.

rhendric:
Did you get around to playing? Any thoughts?

hdgarrood:
Not just yet, sorry. I wanted to get 0.13.2 out because lots of people have been tripped up by the npm installer dropping support for node 8.

PR #3565, opened 2019-03-17
Simplify `purs publish` resolutions format

### LINKED ISSUE #3499: Simplify `purs publish` resolutions format
Currently, `purs publish` expects to be supplied with two JSON files: a manifest file and a resolutions file. The latter is only used to determine the precise versions of all dependencies which are in use on the developer's machine, so that we can generate links across packages which work (without having to do any version solving on the Pursuit end). The compiler expects the resolutions file to have the same schema as the output of `bower list --offline --json`, but this schema is hugely wasteful -- so wasteful, in fact, that we're seeing OOM errors in the course of trying to produce the JSON file: see https://github.com/purescript-contrib/pulp/issues/351. A much simpler schema would suffice, e.g.:

```
// resolutions.json
{
  "purescript-prelude": "4.0.0",
  "purescript-lists": "6.0.0",
  ...
}
```

and we would leave the responsibility of producing this data to external tools (as is the case currently). There are two additional upsides of doing this:

1. It further decouples the compiler from Bower; technically, any tool *could* produce this data according to the schema we're expecting, but realistically, because it's such a silly schema, none actually will (apart from Bower). It will be much easier for alternative tools to provide this data if they're able to use a more sensible schema.
2. It allows us to simplify the part of the compiler which currently extracts the relevant information from the resolutions file.

hdgarrood:
Fixes #3499.

Simplifies the format for resolutions files passed via the CLI to `purs
publish`. A new-style resolutions file should look something like this:

```
{
  "purescript-prelude": {
     "version": "4.0.0",
     "path": "bower_components/purescript-prelude"
  },
  "purescript-lists": {
     "version": "6.0.0",
     "path": "bower_components/purescript-lists"
  },
  ...
}
```

The version field is used for generating links between packages on
Pursuit, and the path field is used to obtain the source files while
generating documentation: all files matching the glob "src/**/*.purs"
relative to the `path` directory will be picked up.

The "version" field is optional, but omitting it will mean that no links
will be generated for any declarations from that package on Pursuit. The
"path" field is required.

We continue to permit the old-style format (i.e. the output of
`bower list --json`), but eventually we will want to start warning when
we encounter it, and eventually drop support for it entirely.

Some other smaller changes which were made in the process of
implementing the above:

- Use a new directory `tests/purs/publish` for packages for testing
`purs publish`, as opposed to using packages from the
`tests/support/bower_components` directory. This is necessary if we want
to be able to test packages which have any dependencies (previously we
were only testing with prelude, which has none). The new directory
contains copies of the prelude and console libraries with dependencies
pre-installed.

- Stop emitting warnings about extraneous or undeclared dependencies; we
no longer have enough information to be able to do this.


garyb:
Do we need to include all these purs files? They'll be brought into the `support` directory already, so perhaps we could copy/link things from there instead for the publish test.

hdgarrood:
Sounds sensible; I'll give that a try.

garyb:
Or maybe just set up a synthetic example with minimal modules, if that's easier? Just trying to avoid having gobs of PS code in here that might need maintaining one day, that aren't really relevant to the thing being tested in this case 🙂

hdgarrood:
We could set up a synthetic example but I think it's quite important that we exercise lots of different parts of the code for handling different kinds of declarations or declarations with lots of different types in them, re-exports, and so on, and setting up a synthetic example which does all that would be a lot of effort.

garyb:
Ah right, that makes sense - I just assumed the additions were only relating to testing the resolutions, as I've not looked at the actual code yet.

hdgarrood:
The actual code changes in this PR are only related to resolutions, I've just modified the tests in quite a drastic way because it was necessary (as far as I could tell) in order to allow testing the resolutions changes while also preserving the tests which exercise all the other stuff which `purs publish` does. I'll have a think about how to arrange the tests (and hopefully address the CI failures too, which I think are related).

hdgarrood:
Okay, found a way of testing it without adding a whole bunch of purs files, and managed to get CI passing. Should be good to review now.

hdgarrood:
:man_facepalming:

hdgarrood:
Thanks! Yeah I think it's not only more polite but also easier from my point of view (as it will lead to fewer bug reports and breakages people will ask me to investigate.)

Perhaps we should start emitting warnings now about the use of the bower format then?

PR #3567, opened 2019-03-18
Represent class dictionaries as data types

### LINKED ISSUE #781: Represent typeclasses as data constructors in functional core

garyb:
Resolves #781

The typechecking aspect of this works perfectly now, since #3128 (... that was Oct 2017? :scream: I thought it was fairly recent)

There are two failing tests that shouldn't be too hard to fix:
- #3381 will fix one of them (there's a case that should fail due to a cycle in classes, but since there are no synonyms associated with classes anymore it's uncaught now)
- Something needs updating in the IDE code so it doesn't import the data types for classes as data types

And there are two TODOs
- [ ] Avoid duplicate case desugaring by not using sugar in `desugarTypeClasses`
- [ ] Don't assume that superclasses can be taken from a dictionary by using an `Accessor`, as that assumes the implementation of types is record-like in the codegen target

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -788,11 +788,6 @@ data Expr
   --
   | Ado (Maybe ModuleName) [DoNotationElement] Expr
   -- |
-  -- An application of a typeclass dictionary constructor. The value should be
-  -- an ObjectLiteral.
-  --
-  | TypeClassDictionaryConstructorApp (Qualified (ProperName 'ClassName)) Expr
-  -- |
   -- A placeholder for a type class dictionary to be inserted later. At the end of type checking, these
   -- placeholders will be replaced with actual expressions representing type classes dictionaries which
   -- can be evaluated at runtime. The constructor arguments represent (in order): whether or not to look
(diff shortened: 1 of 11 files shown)


hdgarrood:
Do you think we should try to get #3381 merged before or after this?

garyb:
I'm fine with it either way - there should be no conflicts if that goes in first though, whereas if I do something in here there probably will be something to resolve.

This PR will probably be on hold for a couple of days anyway, I was talking to Christoph about some stuff to do with externs, so I'm probably going to make a PR changing some things there before coming back to this (it's also relevant to this PR - should help fix the IDE failure case).

garyb:
I'll try and bash this back into shape. I got sucked down a wormhole when I started tinkering with the externs and lost sight of the primary goal, then life happened, etc.

Making the externs carry just what is necessary (rather than including additional compiler generated stuff that could instead be regenerated on load) will be a fairly decent chunk of work on its own I think.

JordanMartinez:
Should this be closed in favor of #4125?

PR #3576, opened 2019-03-26
Allow kind annotations in `forall` contexts

### LINKED ISSUE #3549: Allow kind annotations in `forall` contexts
We should allow kind annotations in forall binders, in the same vein as type/data declarations

```purescript
identity :: forall (a :: Type) . a -> a
```

colinwahl:
This PR adds support for kind annotations in `forall` contexts.  The parser changes aren't very important (hopefully they will go away soon in favor of purescript-cst :) ), but this change will allow us to plug in purescript-cst easier (see https://github.com/natefaubion/purescript-cst/blob/master/src/Language/PureScript/CST/Convert.hs#L173)

This intends to close #3549


src/Language/PureScript/Types.hs
```diff
@@ -590,7 +590,7 @@ compareType (TypeApp _ a b) (TypeApp _ a' b') = compareType a a' <> compareType
 compareType (TypeApp {}) _ = LT
 compareType _ (TypeApp {}) = GT
 
-compareType (ForAll _ a b c) (ForAll _ a' b' c') = compare a a' <> compareType b b' <> compare c c'
+compareType (ForAll _ a b c d) (ForAll _ a' b' c' d') = compare a a' <> fromMaybe LT (compareKind <$> b <*> b') <> compareType c c' <> compare d d'
```
colinwahl: I'm not sure how I should factor in the possible kind annotation in to `eqType`/`compareType`.

colinwahl:
I've got some tests I need to take a look at - I will get to that tomorrow.

src/Language/PureScript/Types.hs
natefaubion: I don't think this is correct because that means if either side is Nothing then the answer is always `LT`. I would just write another function that does the pattern matching.

```haskell
compareMaybeKind :: Maybe (Kind a) -> Maybe (Kind a) -> Ordering
compareMaybeKind Nothing Nothing   = EQ
compareMaybeKind Nothing (Just _)  = LT
compareMaybeKind (Just _) Nothing  = GT
compareMaybeKind (Just a) (Just b) = compareKind a b
```

colinwahl: Good point!

colinwahl:
I've added some simple tests, but if anyone can think of some others, I would be glad to add them.

src/Language/PureScript/Docs/RenderedCode/RenderType.hs
```diff
@@ -131,15 +131,15 @@ matchType = buildPrettyPrinter operators matchTypeAtom
     OperatorTable [ [ AssocL typeApp $ \f x -> f <> sp <> x ]
                   , [ AssocR appliedFunction $ \arg ret -> mintersperse sp [arg, syntax "->", ret] ]
                   , [ Wrap constrained $ \deps ty -> renderConstraints deps ty ]
-                  , [ Wrap forall_ $ \tyVars ty -> mconcat [keywordForall, sp, mintersperse sp (map typeVar tyVars), syntax ".", sp, ty] ]
+                  , [ Wrap forall_ $ \tyVars ty -> mconcat [keywordForall, sp, mintersperse sp (map (\(v, mbK) -> maybe (typeVar v) (\k -> mintersperse sp [mconcat [syntax "(", typeVar v], syntax "::", mconcat [renderKind k, syntax ")"]] ) mbK) tyVars), syntax ".", sp, ty] ]
```
natefaubion: This is a really long line. Maybe we can split up the rendering for the type variable into another function.

colinwahl: Good point - I got a little carried away there 😆

colinwahl:
This should be ready for another look now :)

colinwahl:
Let me know if there's anything else I need to do to get this merged!

PR #3579, opened 2019-03-29
Make bundle tests rerun some compile tests

rhendric:
Per @hdgarrood at https://github.com/purescript/purescript/pull/3562#issuecomment-476745811, here's one way to get some more test coverage of the bundler to ensure that DCE isn't overagressive. What do you think about this?

(Note that two of these tests actually *fail* because of a preexisting bug (?) in the bundler—it looks like it's assuming that `"use strict"` is always the first line of a JavaScript file, and when it isn't (in the case of a handwritten foreign module, here), it generates an invalid bundle. I think I should try to fix this in a separate PR?)


hdgarrood:
I like the approach too, especially that it could be reused in other tests -- I've considered testing generating docs for some of the compiler tests too but never gotten around to it.

PR #3581, opened 2019-03-29
Handle "use strict" better in bundler

rhendric:
This PR changes the behavior of the bundler from ‘assume the first line of input JavaScript files is `"use strict"` and place it before the `exports` var’ to ‘remove any standalone string literals from input JavaScript files and add `"use strict"` back in at the top of the bundle‘. This fixes the failures in #3579 and incidentally makes bundles a little bit smaller.

So now there are three related PRs: this, #3579, and #3562; and my apologies for the multiplication but I thought you might want to review their contents separately. There are no conflicts between the above, and if you merge them in that order, then every commit in the history will build. (#3579 doesn't build without this PR. #3562 builds just fine on its own, but the feedback on that PR is that it needs more tests of the bundler before it's merged, which are provided in #3579.)

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -26,10 +26,10 @@ import Control.Arrow ((&&&))
 import Data.Array ((!))
 import Data.Char (chr, digitToInt)
 import Data.Foldable (fold)
-import Data.Generics (everything, everywhere, mkQ, mkT)
+import Data.Generics (GenericM, everything, everywhere, gmapMo, mkMp, mkQ, mkT)
 import Data.Graph
 import Data.List (stripPrefix)
-import Data.Maybe (mapMaybe, catMaybes)
+import Data.Maybe (catMaybes, fromMaybe, mapMaybe)
 import Data.Version (showVersion)
 import qualified Data.Map as M
 import qualified Data.Set as S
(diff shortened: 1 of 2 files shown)


hdgarrood:
Good catch, thanks. This bug was introduced in #3533 so isn't in any released version yet (I've just checked with 0.12.3), so unconditionally introducing a "use strict" would be a breaking change I think. There are also other directives which use string literals like this ("use asm" for instance, if anyone is still using that) too, which makes stripping all string literal statements like this a little dodgy, I think.

I'd prefer to fix this by going back to the previous behaviour of using `exports` for the internal name of the argument of the foreign module IIFE, so that we don't need to generate the extra
```javascript
  $PS["Main"] = $PS["Main"] || {} ;
  var exports = $PS["Main"];
```
stuff in the foreign module IIFE (although we should keep using `$PS` for the internal argument name of the purescript module IIFE).

rhendric:
I hadn't realized this was such a new regression; I agree with all of your points in that case.

In the case of PureScript modules, however, would you prefer going back to blindly assuming the first statement of the module is `"use strict"` (which is the case, for now), or something that checks the statement before moving it before the `var exports` prelude?

hdgarrood:
Yeah, checking before moving a statement back does indeed sound sensible, thanks :)

rhendric:
One more question: is the reason that we're generating
```js
$PS["Main"] = $PS["Main"] || {};
var exports = $PS["Main"];
```
and not
```js
var exports = $PS["Main"] = $PS["Main"] || {};
```
because we're intentionally avoiding using assignment expressions as values in generated code? I ask because for foreign modules, my initial thought was to generate
```js
(function(exports) {
  // user code here
})(PS["Main"] = PS["Main"] || {});
```
but if assignment-as-a-value is being avoided, I could instead do
```js
(function(exports) {
  // user code here
  PS["Main"] = exports;
})(PS["Main"] || {});
```
which is better than sticking statements at the top of user code, considering directive statements, but not as pure as leaving the entire body of the IIFE untouched.

hdgarrood:
In general yes, I think avoiding assignment in expressions is preferable. In this case though, I think 

```javascript
(function(exports) {
  // user code here
})(PS["Main"] = PS["Main"] || {});
```

probably does make the most sense.

rhendric:
Okay, should be ready for review now.

The test failure in Travis looks unrelated to anything this PR might have changed; do you know what that's about?

PR #3585, opened 2019-04-01
Shrink types less aggressively, fixes #3577

### LINKED ISSUE #3577: Shrinking of types in Errors seems too aggressive
On current master, given:
```purescript
module Main where

import Data.Tuple.Nested (Tuple4)

something :: Tuple4 Int String Int String
something = ?y
```

I get the following compiler output:
```
Error found:
in module Main
at src\Main.purs:6:13 - 6:13 (line 6, column 13 - line 6, column 13)

  Hole 'y' has the inferred type

    Tuple Int (Tuple String (... ... (... ...)))

    Main.something  :: ... ... String Int String

in value declaration something
```

This strips away relevant information for the inferred type (I need to know all the fields to build such a Tuple) and weirdly enough seems to strip depth the other way around when printing the type for the hole.

This is a minimal example, but I've hit this a lot at work when doing things with Halogen or Codecs, where the stripping removes whether my hole is a `HTML` or `Component` and just leaves me with type parameters.

hdgarrood:
Shrink types less aggressively by treating nested type applications as
all being at the same level. Now, given the following module:

    module Main where

    import Data.Tuple.Nested (Tuple4)

    something :: Tuple4 Int String Int String
    something = ?y

the error we get is displayed as:

    Hole 'y' has the inferred type

      Tuple Int (Tuple String (Tuple Int (... ... ...)))

      Main.something  :: Tuple4 Int String Int String

    in value declaration something

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -879,7 +879,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
           _ -> []
       in
         paras $ [ line $ "Hole '" <> markCode name <> "' has the inferred type "
-                , markCodeBox (indent (typeAsBox prettyDepth ty))
+                , markCodeBox (indent (typeAsBox maxBound ty))
                 ] ++ tsResult ++ renderContext ctx
     renderSimpleErrorMessage (MissingTypeDeclaration ident ty) =
       paras [ line $ "No type declaration was provided for the top-level declaration of " <> markCode (showIdent ident) <> "."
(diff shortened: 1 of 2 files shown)


natefaubion:
I personally don't think holes should be truncated at all.

hdgarrood:
Could you not say the same about other type errors though? In practice, there is a limit to the amount of information which can be reasonably conveyed in a type occurring in a type error, and including more information beyond this limit hurts usability. In the tuple example specifically, you can update your rhs from `?y` to`Tuple 0 (Tuple “hi” ?y)` to learn what the rest of the tuple needs to be.

hdgarrood:
Although I think we probably want to merge this regardless because it also fixes the issue where repeated type applications are truncated in a confusing way, eg Tuple4 Int String Int String becomes ... ... String Int String on master currently.

hdgarrood:
Having said that, I suppose that when you have a typed hole, you have actually asked the compiler to tell you what the type is, so yeah I think it probably does make sense to not truncate those.

hdgarrood:
Okay, I've pushed a new commit which prints hole inferred types in full.

PR #3586, opened 2019-04-03
Fail compilation when a module imports itself

### LINKED ISSUE #3079: Should a module importing itself be disallowed?
We use PureScript with webpack via `purs-loader` and `circular-dependency-plugin`, and started seeing this warning:

```
WARNING in Circular dependency detected:
bower_components/purescript-formatters/src/Data/Formatter/Parser/Number.purs -> bower_components/purescript-formatters/src/Data/Formatter/Parser/Number.purs
```

Sure enough, `purescript-formatters` has an import statement for itself with a function defined in the same module:

https://github.com/slamdata/purescript-formatters/blob/859e9f5c7d507441d495870acbb06380940eda97/src/Data/Formatter/Parser/Number.purs#L12

Should this be disallowed?

### LINKED ISSUE #3118: An internal error occurred during compilation: "missing type for data constructor MyDataConstructor in findTypeForDctor"
Hey hey, so, at the moment I'm deep in the midst of writing a demo, but this does keep popping up. I'm working on a minimal reproducible case, but thought I'd at least start a ticket to track it in case someone else has spotted it as well.

```
  error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-0.11.6-Aycu5A5gUJk77UghJXgS0T:Language.PureScript.Crash
  internalError, called at src/Language/PureScript/Linter/Imports.hs:331:22 in purescript-0.11.6-Aycu5A5gUJk77UghJXgS0T:Language.PureScript.Linter.Imports
```

hdgarrood:
Fixes #3079, fixes #3118. This is a breaking change so we should wait
until after 0.12.4 is released to merge.

I have used the CycleInModules error to indicate this; previously we
only emitted this error for cycles containing two or more modules, but
now we also emit it for singleton cycles (which occur precisely when
modules import themselves).

This commit also fixes a separate issue where the CycleInModules error
was not printing one of the modules in a cycle (specifically, the module
which happened to come first in the cycle), presumably due to
accidentally using the tail rather than the whole list while
constructing the error message. I've made this problem easier to avoid
here by using `nonEmpty` instead of `:|` to construct the non-empty
list, which means that we don't need to bind a reference to the tail at
all.

### DIFF
--- a/tests/purs/failing/SelfImport/Dummy.purs
+++ b/tests/purs/failing/SelfImport/Dummy.purs
@@ -1,0 +1,5 @@
+-- This module only exists so that we perform a full build for the
+-- SelfImport.purs module. If this module didn't exist, we would perform a
+-- single-module fast rebuild, which doesn't perform the `sortModules` step,
+-- and so the error we want to see wouldn't be emitted.
+module Dummy where
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -565,9 +565,13 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
     renderSimpleErrorMessage (CycleInDeclaration nm) =
       line $ "The value of " <> markCode (showIdent nm) <> " is undefined here, so this reference is not allowed."
     renderSimpleErrorMessage (CycleInModules mns) =
-      paras [ line "There is a cycle in module dependencies in these modules: "
-            , indent $ paras (map (line . markCode . runModuleName) mns)
-            ]
+      case mns of
+        [mn] ->
+          line $ "Module " <> markCode (runModuleName mn) <> " imports itself."
+        _ ->
+          paras [ line "There is a cycle in module dependencies in these modules: "
+                , indent $ paras (map (line . markCode . runModuleName) mns)
+                ]
     renderSimpleErrorMessage (CycleInTypeSynonym name) =
       paras [ line $ case name of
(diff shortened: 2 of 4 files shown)


tests/purs/failing/SelfImport/Dummy.purs
```diff
@@ -0,0 +1,5 @@
+-- This module only exists so that we perform a full build for the
+-- SelfImport.purs module. If this module didn't exist, we would perform a
+-- single-module fast rebuild, which doesn't perform the `sortModules` step,
+-- and so the error we want to see wouldn't be emitted.
+module Dummy where
```
natefaubion: Does this mean that when using fast-build I can do a self import and the compiler will allow it? And the compiler will only give me an error when switching to a full build? I don't think we have any other cases where the compiler conditionally allows something _within_ the module in question.

hdgarrood: Unfortunately, yes. There are a couple of other examples of this which already do this though, which can be seen by looking inside `Language.PureScript.Make`; specifically, attempting to define a module whose name starts with `Prim`, and giving a module a name which is already taken by some other module. See:

https://github.com/purescript/purescript/blob/b09144bc68637bf6657e4147567a3e9185221594/src/Language/PureScript/Make.hs#L118-L135

natefaubion: To me it seems like all of these checks could be done elsewhere. The Prim check can be done for each module, as well as the self import check. And module names could be checked against what purs-ide has in memory (which is as good as it can get, admittedly). Those are all larger changes though, so I'm fine with this for now.

PR #3587, opened 2019-04-03
Use one build script for all CI builds

hdgarrood:
This should allow us to get appveyor working again. Since we upgraded to
GHC 8.6.4, all appyeyor builds seem to have failed because we need to
recompile all of the compiler's Haskell dependencies, which takes too
long.


hdgarrood:
Unfortunately it seems that, like on Travis, cache uploading is included in the hour allotted for each build, so if the build overruns and gets terminated you don't get to upload your build cache. We'll need to find a Windows equivalent of the `timeout` command.

hdgarrood:
This is still going to fail on appveyor, because it doesn't store build caches on PRs. After this gets merged, though, the cache should be stored and appveyor builds should work :)

Issue #3591, opened 2019-04-08
Importing a type with the same name as its module is confusing


justinwoo:
e.g. with psc-ide-emacs, importing on "Effect" leads to just errors reported on undeciable importing between Effect the module and the type. Why shouldn't it import the type as it did before?

I'm also already using my branch of psc-ide-emacs to be able to complete on not yet imported qualified imports: https://github.com/purescript-emacs/psc-ide-emacs/pull/166

Reporting here because it seems like purs ide server should default in one way over the other.

kRITZCREEK:
I think you're right, we should be filtering out modules when we determine candidates for imports.

Needs filtering here:
https://github.com/purescript/purescript/blob/35038c7ef472d7ae27bf9f31a9155ba32bada812/src/Language/PureScript/Ide/Imports.hs#L280

We could work around it in the editors by passing filters that exclude the module namespace for now.

PR #3597, opened 2019-04-11
Fix "purs ide server" crash on macOS by providing connection hints

### LINKED ISSUE #3594: Running `purs ide server` on macOS crashes
```bash
purs ide server
purs: Network.Socket.listen: unsupported operation (Operation not supported on socket)
```

This happens since `0.12.4`.

f-f:
This should fix #3594 

The fix solves the issue on my macOS, @i-am-the-slime could you also try it out?

### DIFF
--- a/app/Command/Ide.hs
+++ b/app/Command/Ide.hs
@@ -38,7 +38,6 @@ import           Language.PureScript.Ide.Error
 import           Language.PureScript.Ide.Types
 import           Language.PureScript.Ide.Watcher
 import qualified Network.Socket                    as Network
-import qualified Network.BSD
 import qualified Options.Applicative               as Opts
 import           System.Directory
 import           System.Info                       as SysInfo
@@ -48,11 +47,9 @@ import           System.IO.Error                   (isEOFError)
 
 listenOnLocalhost :: Network.PortNumber -> IO Network.Socket
 listenOnLocalhost port = do
-  proto <- Network.BSD.getProtocolNumber "tcp"
   let hints = Network.defaultHints
         { Network.addrFamily = Network.AF_INET
         , Network.addrSocketType = Network.Stream
-        , Network.addrProtocol = proto
         }
   addr:_ <- Network.getAddrInfo (Just hints) (Just "127.0.0.1") (Just (show port))
   bracketOnError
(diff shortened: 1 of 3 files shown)


i-am-the-slime:
I can try. I'm doing `make build` right now. It'll probably take a few hours, right?

f-f:
@i-am-the-slime you might want to do `make install` instead (that will install `purs` system-wide). And yeah, it will take at least ~30mins on a recent processor to compile everything if you didn't use stack before

i-am-the-slime:
It would be so nice for this beautiful compiler to be self-hosted.

i-am-the-slime:
Appveyor says this: Timed out while installing dependencies. :)

i-am-the-slime:
As Apache would say: 
# It works!
Thank you very much @f-f .

kRITZCREEK:
Lovely, this looks good to me. If we're allright with bumping network to 3.x over our current snapshot I'm happy to merge this.

f-f:
@i-am-the-slime you're welcome 🙂 

@kRITZCREEK could we cut a small release with this?

hdgarrood:
We need to verify that this still works on Windows before merging (unless I'm missing something and someone has already done this). If not I can get appveyor running again soon (perhaps this weekend).

I agree that making a small bugfix release for this would be good. I think we should also include a fix for #3591.

hdgarrood:
I've made a milestone: https://github.com/purescript/purescript/milestone/27

garyb:
I'll give it a try when I get a minute later today.

garyb:
Hmm, well it seems like it won't build on Windows with this change, so I don't really know whether it works. It blows up at the end with this:

```
    ghc.EXE: unable to load package `network-3.0.0.1'
    ghc.EXE:  | D:\Home\GitHub\purescript\.stack-work\install\f33caf5a\lib\x86_64-windows-ghc-8.6.4\network-3.0.0.1-4CoErOi6AngG4EwBDdM0JI\HSnetwork-3.0.0.1-4CoErOi6AngG4EwBDdM0JI.o: unknown symbol `if_nametoindex'
```

Maybe I'm doing it wrong somehow?

f-f:
@garyb it looks like that was fixed in `network..3.0.1.1` (see [here](https://github.com/haskell/network/pull/391)), pushed a commit upgrading to that.

Unfortunately `network-bsd` has tight bounds, so I enabled "allow-newer" in the stack file (it should be fine to keep it until the bounds are fixed)

garyb:
Ah cool, trying again then 🙂

garyb:
All seems good 👍

hdgarrood:
Come to think of it is the network-bsd dependency necessary? It looks like network-bsd is only being used to get a protocol number, but the example in the network library’s docs doesn’t bother setting the protocol number on the hints: http://hackage.haskell.org/package/network-3.0.1.1/docs/Network-Socket.html

f-f:
@hdgarrood good catch!
I included it because the old implementation did (i.e. before the upgrade that introduced the bug), but things seem to work also without it.

f-f:
Thanks 🙂

purcell:
Awesome fix, thanks! Can https://github.com/purescript-contrib/node-purescript be bumped too?

PR #3602, opened 2019-04-12
Prevent invalid js generation for module names with an apostrophe:

matthew-hilty:
Formerly, the purescript code

    module Foo' where
    foo = "foo"

    module Bar where
    import Foo'

would be parsed without error but would generate the following invalid javascript:

    var Foo' = require("../Foo'/index.js");

The present commit prevents this error in two ways:
1. requiring that module and qualifier names comprise alphanumeric characters only;
        (I've seen some past discord among ps developers about this restriction,
        but it seems the safest choice and code currently in Parser/Lexer.hs (specifically, 'validUName')
        suggests that's the present naming policy.)
2. applying 'identCharToText' to ModuleNames
       to ensure that the rendered javascript variable names contain no invalid chars,
       should module-naming conventions ever change.


hdgarrood:
The approach seems okay, but I’m tempted to say that we should wait until the cst work is in before addressing this.

natefaubion:
`purescript-cst` does handle this check.

PR #3603, opened 2019-04-12
WIP - Add error for redefining `it` in REPL

### LINKED ISSUE #3527: Cannot define a value `it` in the repl
For instance:

```
> it = 3
Error found:
in module $PSCI
at :1:1 - 1:6 (line 1, column 1 - line 1, column 6)

  Multiple value declarations exist for it.
```
("for what? multiple value declarations exist for what????")

Perhaps we can use an identifier which you wouldn't be able to define normally like `$it` or something.

GoNZooo:
Should fix #3527 by way of an error message specifying why `it` isn't a valid identifier.

This isn't 100% complete because I'm having issues creating a test that will actually test the error output correctly. When the test is running you can see the output, but I'd like to assert it properly via a `@prints ...` tag (or the like).

I'm also not sure what to put in the error example file automatically output by the compiler, but this seems almost secondary as it's likely to reiterate the REPL error.

### DIFF
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -171,7 +171,7 @@ handleDecls
   -> m ()
 handleDecls ds
   | isJust (badDeclarationErrors ds) =
-    printErrors $ maybe mempty mconcat (badDeclarationErrors ds)
+    printErrors $ foldMap mconcat (badDeclarationErrors ds)
   | otherwise = do
   st <- gets (updateLets (++ ds))
   let m = createTemporaryModule False st (P.Literal P.nullSourceSpan (P.ObjectLiteral []))
--- a/src/Language/PureScript/Interactive.hs
+++ b/src/Language/PureScript/Interactive.hs
@@ -14,12 +14,13 @@ import           Prelude.Compat
 import           Protolude (ordNub)
 
 import           Data.List (sort, find, foldl')
-import           Data.Maybe (fromMaybe, mapMaybe)
+import           Data.Maybe (fromMaybe, mapMaybe, isJust)
 import qualified Data.Map as M
 import qualified Data.Set as S
 import           Data.Text (Text)
 import qualified Data.Text as T
 
+import           Control.Monad ((>=>))
 import           Control.Monad.IO.Class (MonadIO, liftIO)
 import           Control.Monad.State.Class
 import           Control.Monad.Reader.Class
@@ -168,7 +169,10 @@ handleDecls
   :: (MonadReader PSCiConfig m, MonadState PSCiState m, MonadIO m)
   => [P.Declaration]
   -> m ()
-handleDecls ds = do
+handleDecls ds
+  | isJust (badDeclarationErrors ds) =
+    printErrors $ maybe mempty mconcat (badDeclarationErrors ds)
+  | otherwise = do
   st <- gets (updateLets (++ ds))
   let m = createTemporaryModule False st (P.Literal P.nullSourceSpan (P.ObjectLiteral []))
   e <- liftIO . runMake $ rebuild (map snd (psciLoadedExterns st)) m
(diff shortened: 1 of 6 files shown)


tests/purs/psci/DefineItFail.purs
```diff
@@ -0,0 +1,2 @@
+-- @prints NoItInREPL
```
GoNZooo: I'd like to make this directive like the `@shouldFailWith` directive in the compiler tests, but for checking either for a specific failure in the REPL or output text, whichever works. I'm having issues getting there, though, so any help is welcome.

hdgarrood: Did you get this working?

src/Language/PureScript/Interactive.hs
```diff
@@ -168,7 +169,10 @@ handleDecls
   :: (MonadReader PSCiConfig m, MonadState PSCiState m, MonadIO m)
   => [P.Declaration]
   -> m ()
-handleDecls ds = do
+handleDecls ds
+  | isJust (badDeclarationErrors ds) =
```
hdgarrood: `isJust` is a code smell: better to match on the result and do whatever it is you want to do with the `Just` value, I think.

Putting that aside for a moment though, did you consider putting this check in the parser instead? I think that would be a bit tidier.

tests/purs/psci/DefineItFail.purs
GoNZooo: No

src/Language/PureScript/Interactive.hs
GoNZooo: I could replace it with something like this instead:

```haskell
handleDecls'
  :: (MonadReader PSCiConfig m, MonadState PSCiState m, MonadIO m)
  => [P.Declaration]
  -> m ()
handleDecls' ds =
  maybe noBadDeclarations (printErrors . mconcat) (badDeclarationErrors ds) where
    noBadDeclarations = do
      st <- gets (updateLets (++ ds))
      let m = createTemporaryModule False st (P.Literal P.nullSourceSpan (P.ObjectLiteral []))
      e <- liftIO . runMake $ rebuild (map snd (psciLoadedExterns st)) m
      either printErrors (const (put st)) e
```

As for the parser bit, I will look into putting it there instead.

GoNZooo: It's a little bit awkward because we currently have this function that determines if a declaration is good:

```haskell
acceptable :: P.Declaration -> Bool
acceptable P.DataDeclaration{} = True
acceptable P.TypeSynonymDeclaration{} = True
acceptable P.ExternDeclaration{} = True
acceptable P.ExternDataDeclaration{} = True
acceptable P.TypeClassDeclaration{} = True
acceptable P.TypeInstanceDeclaration{} = True
acceptable P.ExternKindDeclaration{} = True
acceptable P.TypeDeclaration{} = True
-- acceptable (P.ValueDeclaration (P.ValueDeclarationData _ (P.Ident "it") _ _ _)) = False
acceptable P.ValueDeclaration{} = True
acceptable _ = False
```
I added the commented out line just to see how this would behave and with it you get the following behavior in the REPL:

```
> it = 4
(line 1, column 4):
unexpected =
expecting ., {, _, number, char literal, string literal, "true", "false", [, "\\", qualifier, data constructor name, identifier, "case", "if", "do", "ado", "let", (, hole literal, ::, operator, "import", declaration, expression or end of input
```

As opposed to the current behavior with a bit more explanation:

```
> it = 1
Error found:

  it is a variable used to refer to the previously evaluated expression in the REPL. Redefining it is not allowed.


See https://github.com/purescript/documentation/blob/master/errors/NoItInREPL.md for more information,
```

Is there possibly some point after this `acceptable` call where a declaration for `it` could be pre-empted (before the check in the declarations that I added, that is).

hdgarrood: Ah right yes, the error you get by modifying `acceptable` is quite poor. That must be happening because of our use of `try`. How about:

* pushing the use of `try` in the case of `parseDeclaration` slightly further down, so that it is outside of the `acceptable` check,
* modifying the `acceptable` check so that it returns an error where necessary rather than just a Bool?

GoNZooo:
It seems pointless to keep this open when I just don't really have time or energy at the moment to work on this.

Issue #3596, opened 2019-04-09
Constraints and overlapping instances


rhendric:
```purs
class Foo a
instance equatableFoo :: (Eq a) => Foo a

data Bar = Bar
instance barFoo :: Foo Bar
```

Given that there is no `Eq Bar` instance, should these two instances be allowed? PureScript 0.12.4 complains of overlapping instances.

I know instance chains would be one workaround for this minimized example, but suppose `Foo` and `equatableFoo` are put in one file, and `Bar` and `barFoo` in another—then instance chains wouldn't help, but provided that no `Eq Bar` instance is defined in `Bar`'s file, I think the compiler should still be able to determine that `barFoo` doesn't overlap with `equatableFoo`, right?

(I didn't find an exact duplicate for what I want, but #3120 seems related—that enhancement asks for constraints to be treated as guards in instance chains, with first-to-match semantics; and this asks for constraints to be treated as sort of like guards in overlapping instance sets, with exactly-one-match semantics.)

garyb:
If by "should" you mean in an ideal world, then maybe :smile:

But as things are, it is right that they are not allowed - instance resolutions works by looking at the instead head (type part) only, and then only afterward are constraints checked. So `a` and `Foo` are considered to be overlapping, since `a` includes all types.

hdgarrood:
It's perhaps worth noting that GHC also rejects this as overlapping, although you only get an error when you attempt to use the instance:

```haskell
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE UndecidableInstances #-}
module Main where

class Foo a where
  foo :: a -> Bool

instance (Eq a) => Foo a where
  foo _ = True

data Bar = Bar

instance Foo Bar where
  foo _ = False

main = putStrLn $ show $ foo Bar
```
produces
```
Main.hs:16:26: error:
    • Overlapping instances for Foo Bar arising from a use of ‘foo’
      Matching instances:
        instance Eq a => Foo a -- Defined at Main.hs:8:10
        instance Foo Bar -- Defined at Main.hs:13:10
    • In the second argument of ‘($)’, namely ‘foo Bar’
      In the second argument of ‘($)’, namely ‘show $ foo Bar’
      In the expression: putStrLn $ show $ foo Bar
   |
16 | main = putStrLn $ show $ foo Bar
   |             
```

[...]

PR #3604, opened 2019-04-12
WIP - Consider constraints when selecting instances

rhendric:
This is a first stab at an implementation of #3596, mostly intended to serve as a talking point. I know the responses on #3596 were lukewarm-to-negative, so I hope this isn't taken as too pushy, but I thought it would be helpful for that conversation to have a sense of how much the code actually needed to change and how difficult it would be to understand and maintain.

---

Previously, instance selection ignored instance constraints when
determining if instances overlapped, and considered all instances with
matching type arguments. This meant that `instance x :: SomeClass X` and
`instance y :: (SomeConstraint a) => SomeClass a` were considered
overlapping even if `X` doesn't satisfy `SomeConstraint`.

This commit relaxes the check on instance declarations to permit
instances to overlap in a limited way: for each type in the instance, a
concrete type is allowed to overlap with a type variable if the type
variable appears in an instance constraint. Two type variables, even if
given different constraints, are not allowed to overlap in any case.

This commit also modifies instance selection to attempt to solve for all
the overlapping instances with matching type arguments, and if exactly
one instance can be solved, select it instead of throwing an overlapping
instance error. If zero or more than one have solutions, the original
overlapping instance error is thrown.

The upshot is that there are now some (more?) cases where truly
ambiguous overlapping instances are detected at use instead of at
declaration, but in exchange, some cases of nonambiguous "overlapping"
instances, such as the above example, are now permitted.

---

The above text is slightly aspirational. While I have tested this code on compiling my own projects, and that works great, there are two open questions that I have about how entailment works which may reflect flaws in this implementation.

1. Why did `overlapping` choose to ignore some overlaps, and is it correct to continue to do the same overlap check and `minimumBy` after trying to solve subgoals?

The comment there says that subclass dictionaries can't overlap, which sort of makes sense to me but I don't really understand how it works in practice; but then the code goes on to also give a pass to instances that don't have any dependencies, which I completely don't understand. I did what I think is the conservative thing, but since I don't really understand the reasoning behind the original code, I can't be sure—I'll probably have more specific questions about this once I understand it better.

2. What are generalization and defer passes for, and how is my feature constrained by the fact that this implementation doesn't do further passes when it's solving overlapping instances in parallel?

My biggest worry is whether this is a loophole through which incoherency can enter—if, for example, there are two instances that truly could both apply to a use site, but only one of them can be known to apply without doing another generalization pass to show it, this implementation will choose that instance. Can the compiler be tricked into choosing the other instance somewhere else? Even if incoherency is not a danger, will a failure to generalize/defer in these circumstances cause surprise?

Modulo those two questions, I do believe this is fairly robust—in particular, there should be no performance impact on any code that currently compiles, since pretty much all the new computation is in the `PotentialOverlap` branch, which can only be reached on code that would cause an overlapping instance error today.

### DIFF
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -19,7 +19,7 @@ import Control.Monad.State.Class (MonadState(..), modify, gets)
 import Control.Monad.Supply.Class (MonadSupply)
 import Control.Monad.Writer.Class (MonadWriter(..), censor)
 
-import Data.Foldable (for_, traverse_, toList)
+import Data.Foldable (fold, for_, traverse_, toList)
 import Data.List (nub, nubBy, (\\), sort, group, intersect)
 import Data.Maybe
 import Data.Text (Text)
@@ -341,7 +341,7 @@ typeCheckAll moduleName _ = traverse go
           let nonOrphanModules = findNonOrphanModules className typeClass tys
           checkOrphanInstance dictName className tys nonOrphanModules
           let qualifiedChain = Qualified (Just moduleName) <$> ch
-          checkOverlappingInstance qualifiedChain dictName className typeClass tys nonOrphanModules
+          checkOverlappingInstance qualifiedChain dictName className typeClass tys deps nonOrphanModules
           _ <- traverseTypeInstanceBody checkInstanceMembers body
           deps' <- (traverse . overConstraintArgs . traverse) replaceAllTypeSynonyms deps
           let dict = TypeClassDictionaryInScope qualifiedChain idx qualifiedDictName [] className tys (Just deps')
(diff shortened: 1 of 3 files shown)


hdgarrood:
> My biggest worry is whether this is a loophole through which incoherency can enter—if, for example, there are two instances that truly could both apply to a use site, but only one of them can be known to apply without doing another generalization pass to show it, this implementation will choose that instance. Can the compiler be tricked into choosing the other instance somewhere else? Even if incoherency is not a danger, will a failure to generalize/defer in these circumstances cause surprise?

I haven't worked on this code very much so I'm not really qualified to answer these questions, but yes, this is the sort of thing I'd be worried about too.

I'm going to close this for now just because we have a lot of open PRs at the moment, and because I think the discussion on #3596 isn't really close enough to a consensus to warrant a PR for now.

PR #3609, opened 2019-04-16
Load externs concurrently

natefaubion:
Previously externs were loaded serially with `foldM`. This means that if a change is detected farther upstream, it won't try to read downstream externs. This PR loads and parses them all concurrently, which will result in downstream results that are thrown away if there are upstream changes. However, I think this is a decent tradeoff since during development, changes generally happen downstream.

On my machine, with 551 modules compiled and a noop call to `purs compile`, v0.12.5 takes ~4s, while this PR takes ~2.6s, and with the parser changes as well in #3608 it takes ~1.6s.

### DIFF
--- a/src/Language/PureScript/Make/BuildPlan.hs
+++ b/src/Language/PureScript/Make/BuildPlan.hs
@@ -112,7 +112,7 @@ construct
   -> ([Module], [(ModuleName, [ModuleName])])
   -> m BuildPlan
 construct MakeActions{..} (sorted, graph) = do
-  prebuilt <- foldl' prunePrebuiltModules M.empty . catMaybes <$> A.forConcurrently sorted findExistingExtern
+  prebuilt <- foldl' collectPrebuiltModules M.empty . catMaybes <$> A.forConcurrently sorted findExistingExtern
   let toBeRebuilt = filter (not . flip M.member prebuilt . getModuleName) sorted
   buildJobs <- foldM makeBuildJob M.empty (map getModuleName toBeRebuilt)
   pure $ BuildPlan prebuilt buildJobs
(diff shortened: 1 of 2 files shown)


natefaubion:
This appears to break TestUtils code that provides an exception for the output timestamp:
https://github.com/purescript/purescript/blob/674089385ee44ac596d13b7c7b29b45369b823ad/tests/TestUtils.hs#L216

src/Language/PureScript/Make/BuildPlan.hs
```diff
@@ -117,37 +121,33 @@ construct MakeActions{..} (sorted, graph) = do
...
 
-    findExistingExtern :: M.Map ModuleName Prebuilt -> Module -> m (M.Map ModuleName Prebuilt)
-    findExistingExtern prev (getModuleName -> moduleName) = do
-      outputTimestamp <- getOutputTimestamp moduleName
+    findExistingExtern :: Module -> m (Maybe (ModuleName, Prebuilt))
+    findExistingExtern (getModuleName -> moduleName) = runMaybeT $ do
+      outputTimestamp <- MaybeT $ getOutputTimestamp moduleName
+      inputTimestamp <- lift $ getInputTimestamp moduleName
+      existingTimestamp <- MaybeT $ pure $ case inputTimestamp of
+        Left RebuildNever ->
```
hdgarrood: Regarding the test failures, perhaps the call to `getOutputTimestamp` should wait until after we've established that the input timestamp is not `Left RebuildAlways`?

natefaubion: I wouldn't think that would matter for this. `outputTimestamp` isn't forced unless it's compared. I suspect that it's actually getting forced as part of `prunePrebuiltModules`. Whereas before, it would process everything in order serially, and thus essentially short-circuit, maybe this is doing comparisons that it otherwise wouldn't have done?

natefaubion: But yes, I can switch it around so it does it in a different order. It shouldn't check the output timestamp at all if it's RebuildAlways.

hdgarrood: I think what's happening is that in `findExistingExtern`, the new code no longer distinguishes the case where a module has a `RebuildNever` policy from the case where a module has no rebuild policy (i.e. it's just using timestamps as normal). In the former case, there's no need to compare the current module's output timestamp to its dependencies' output timestamps (because we already know we aren't going to bother rebuilding it). However, it looks like the `prunePrebuiltModules` in this PR treats both cases identically.

hdgarrood: Also I'm not sure about the name `prunePrebuiltModules`; the pruning isn't happening until the following line, right? I'd be more inclined to call that function something like `collectPrebuiltModules`.

hdgarrood:
> However, I think this is a decent tradeoff since during development, changes generally happen downstream.

I agree. I think another point in favour of this approach is that if there does happen to be a change very far upstream, the compiler will have a lot of work to do rebuilding all of the downstream modules, so you're probably going to be waiting a little while anyway.

natefaubion:
@hdgarrood Does this look correct? Tests are passing for me at least. Another option is to give `RebuildNever` builds a timestamp in the past.

hdgarrood:
Shall we merge this one now too? The same reasoning applies here as for the CST one I think - having it in master for a bit before the next release is probably the best way of testing it properly now. I don't want to put pressure on Christoph to review since he's very busy at the moment.

PR #3610, opened 2019-04-19
Remove failable patterns and NoMonadFailDesugaring extension

### LINKED ISSUE #3583: Stop using failable monadic pattern matches
This will allow us to disable `-XNoMonadFailDesugaring`. We will need to do this before the extension is removed from GHC (which will happen in GHC 8.8, as far as I'm aware).

See discussion in #3560, in particular:

> Ah I see: most of the MonadFail constraints are actually arising from inside the type checker, and they're just propagating out to everywhere else.
>
> Since most of them seem to be due to pattern matches like
>
>     expr@(TypedValue _ value ty) <- infer ...
>
> how do we feel about having infer return something like
> ```purescript
> data InferredValue = InferredValue
>   { inferredCheck :: Bool
>  , inferredValue :: Expr
>  , inferredType :: Type
>  , inferredExpr :: Expr
>  }
>```
> where the inferredExpr field is always a TypedValue whose fields are the same as the other three fields? Then we can replace the above with
>
>     InferredValue _ value ty expr <- infer ..
>
> and we can recover the current infer with `fmap inferredExpr . infer`.

adnelson:
Fixes #3583 

Note that I added no new tests because changes do not change any behavior.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -5,7 +5,7 @@
 --
 module Language.PureScript.TypeChecker.Types
   ( BindingGroupType(..)
-  , Inferred(..)
+  , TypedValue'(..)
   , typesOf
   ) where
 
@@ -70,11 +70,11 @@ data BindingGroupType
   deriving (Show, Eq, Ord)
 
 -- | The result of a successful type check.
-data Inferred = Inferred Bool Expr SourceType
+data TypedValue' = TypedValue' Bool Expr SourceType
 
--- | Convert an inferred value into an expression.
-inferredToExpr :: Inferred -> Expr
-inferredToExpr (Inferred c e t) = TypedValue c e t
+-- | Convert an type checked value into an expression.
+tvToExpr :: TypedValue' -> Expr
+tvToExpr (TypedValue' c e t) = TypedValue c e t
 
 -- | Infer the types of multiple mutually-recursive values, and return elaborated values including
 -- type class dictionaries and type annotations.
@@ -253,8 +253,8 @@ checkTypedBindingGroupElement mn (ident, (val, ty, checkType)) dict = do
   -- Check the type with the new names in scope
   val' <- if checkType
             then withScopedTypeVars mn args $ bindNames dict $ check val ty'
-            else return (Inferred False val ty')
-  return (ident, (inferredToExpr val', ty'))
+            else return (TypedValue' False val ty')
+  return (ident, (tvToExpr val', ty'))
 
 -- | Infer a type for a value in a binding group which lacks an annotation.
 typeForBindingGroupElement
@@ -267,7 +267,7 @@ typeForBindingGroupElement
   -> m ((SourceAnn, Ident), (Expr, SourceType))
 typeForBindingGroupElement (ident, (val, ty)) dict = do
   -- Infer the type with the new names in scope
-  Inferred _ val' ty' <- bindNames dict $ infer val
+  TypedValue' _ val' ty' <- bindNames dict $ infer val
   -- Unify the type with the unification variable we chose for this definition
   unifyTypes ty ty'
   return (ident, (TypedValue True val' ty', ty'))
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -339,117 +347,120 @@ infer' (Literal ss (ObjectLiteral ps)) = do
...
-  newVals <- zipWith (\(name, _) t -> (name, t)) ps <$> traverse (infer . snd) ps
+  newValsInferred <- zipWith (\(name, _) t -> (name, t)) ps <$> traverse (infer . snd) ps
   let toRowListItem = uncurry srcRowListItem
-  let newTys = map (\(name, TypedValue _ _ ty) -> (Label name, ty)) newVals
+  let newTys = map (\(name, Inferred _ _ ty) -> (Label name, ty)) newValsInferred
   oldTys <- zip (map (Label . fst) ps) <$> replicateM (length ps) freshType
   let oldTy = srcTypeApp tyRecord $ rowFromList (toRowListItem <$> oldTys, row)
-  o' <- TypedValue True <$> check o oldTy <*> pure oldTy
-  return $ TypedValue True (ObjectUpdate o' newVals) $ srcTypeApp tyRecord $ rowFromList (toRowListItem <$> newTys, row)
+  o' <- TypedValue True . inferredToExpr <$> check o oldTy <*> pure oldTy
```
hdgarrood: Bit of a nit but I think
```haskell
TypedValue True <$> (inferredToExpr <$> check o oldTy) <*> pure oldTy
```
might be a bit clearer here (they are equivalent, right?)

adnelson: They are indeed, and I think I used both forms in different places :) if you prefer the above it's fine with me

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -68,6 +69,13 @@ data BindingGroupType
   | NonRecursiveBindingGroup
   deriving (Show, Eq, Ord)
 
+-- | The result of a successful type check.
```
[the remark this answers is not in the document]
adnelson: I'm fine with that. 

One additional thing I'd like to do is document what the boolean there is doing -- the first time I tried these changes I didn't properly understand it and it lead to broken tests. This time I took care not to change anything, but I'd still like to know what it's actually doing 😅 it's not documented in `Declarations.hs` either, that I could tell.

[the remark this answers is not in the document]
hdgarrood: Yes, I agree, that would be good. I am not entirely sure either, though.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -829,14 +842,13 @@ checkFunctionApplication' fn (ConstrainedType _ con fnTy) arg = do
 checkFunctionApplication' fn fnTy dict@TypeClassDictionary{} =
   return (fnTy, App fn dict)
 checkFunctionApplication' fn u arg = do
-  arg' <- do
-    TypedValue _ arg' t <- infer arg
+  TypedValue' checkType expr ty <- do
```
hdgarrood: Oops - I just noticed this, which is little a bit awkward now, I think. How about using a pattern `tv@TypedValue' _ _ ty` here, and then using `tvToExpr tv` below instead of the `TypedValue` constructor?

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -5,6 +5,7 @@
 --
 module Language.PureScript.TypeChecker.Types
   ( BindingGroupType(..)
+  , TypedValue'(..)
```
hdgarrood: Last thing, I promise: does this need to be exported? I'd prefer not to if we can avoid it.

adnelson: No worries, and yeah that's a good catch. In my earlier iterations I had needed to export it but no longer :)

PR #3612, opened 2019-04-26
More specific error on TransitiveExportError

### LINKED ISSUE #3611: More specific error on TransitiveExportError
I manage to trigger `TransitiveExportError` on a fairly regular basis. The error prints as "An export for [class] requires the following to also be exported:" followed by a list of all members of the class.

Any chance we can indicate which of those members is actually missing from the export list?

colinwahl:
- modifies `checkClassMembersAreExported` within the typechecker to only report class members which have not been exported from the module (instead of reporting all class members, regardless of whether or not they are exported) 

I don't think it is possible to add a failing test in which I specify the constructor of an error and also more arguments with the current setup.  If there is some way to do this, I'd be happy to add a test.

This intends to close #3611

### DIFF
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -648,7 +648,7 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
   checkClassMembersAreExported dr@(TypeClassRef ss' name) = do
     let members = ValueRef ss' `map` head (mapMaybe findClassMembers decls)
     let missingMembers = members \\ exps
-    unless (null missingMembers) . throwError . errorMessage' ss' $ TransitiveExportError dr members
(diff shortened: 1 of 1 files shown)


colinwahl:
@hdgarrood sure, here's an example:

```
module Main
  ( class Foo
  , foo1
  ) where

class Foo where
  foo1 :: forall a b. a -> b
  foo2 :: forall a b c. a -> b -> c
```

Previously, compilation would fail with a `TransitiveExportError` saying
```
  An export for class Foo requires the following to also be exported:

    foo1
    foo2
```

Now, compilation will still fail with a `TransitiveExportError`, but it will only show class members which have not been exported, so in the example this means
```
  An export for class Foo requires the following to also be exported:

    foo2
```

hdgarrood:
Fab, thank you :)

PR #3616, opened 2019-04-29
Tweaks to type pretty printing

### LINKED ISSUE #3577: Shrinking of types in Errors seems too aggressive

garyb:
Resolves #3577 again, hopefully.

As well as not incrementing the depth for constraints, it also prints types regardless of depth if they're "simple".

### DIFF
--- a/src/Language/PureScript/Pretty/Types.hs
+++ b/src/Language/PureScript/Pretty/Types.hs
@@ -62,7 +62,6 @@ type PrettyPrintConstraint = (Qualified (ProperName 'ClassName), [PrettyPrintTyp
 convertPrettyPrintType :: Int -> Type a -> PrettyPrintType
 convertPrettyPrintType = go
   where
-  go d _ | d < 0 = PPTruncated
   go _ (TUnknown _ n) = PPTUnknown n
   go _ (TypeVar _ t) = PPTypeVar t
   go _ (TypeLevelString _ s) = PPTypeLevelString s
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/Pretty/Types.hs
```diff
@@ -62,19 +62,19 @@ type PrettyPrintConstraint = (Qualified (ProperName 'ClassName), [PrettyPrintTyp
...
   go _ (TypeVar _ t) = PPTypeVar t
   go _ (TypeLevelString _ s) = PPTypeLevelString s
   go _ (TypeWildcard _ n) = PPTypeWildcard n
   go _ (TypeConstructor _ c) = PPTypeConstructor c
   go _ (TypeOp _ o) = PPTypeOp o
   go _ (Skolem _ t n _) = PPSkolem t n
-  go d (ConstrainedType _ (Constraint _ cls args _) ty) = PPConstrainedType (cls, go (d-1) <$> args) (go (d-1) ty)
+  go _ (REmpty _) = PPRow [] Nothing
+  go d _ | d < 0 = PPTruncated
+  go d (ConstrainedType _ (Constraint _ cls args _) ty) = PPConstrainedType (cls, go d <$> args) (go d ty)
```
garyb: Maybe should still be `(d-1)` on the LHS here?

hdgarrood: Yeah, I think that would make sense.

PR #3619, opened 2019-05-02
replace unknown names with holes in name desugaring

colinwahl:
Replaces unknown names with holes in name desugaring

* Introduces new `UnknownValueHint` error 
* Introduces new `UnknownValue` case to `Expr` 
* modifies name desugaring (in the `updateValue` phase) to replace unknown names with `UnknownValue`
* modifies typechecking to turn `UnknownValue` into `UnknownValueHint` errors

I have written a first pass at the error message, but I think it can definitely be improved.

Here is an example:

```
module Foo where

import Prelude

foo :: Int
foo = bar 10
```

Previously, this would fail with an `UnknownName` error, but now it fails with a `UnknownValueHint` error, stating

```
[1/1 UnknownValueHint] src/Main.purs:6:7

  6  foo = bar 10
           ^^^

  Unknown name 'bar' has the inferred type

    t0

  You could substitute the unknown value with one of these values:

    Control.Category.identity  :: forall t a. Category a => a t t
    Data.EuclideanRing.degree  :: forall a. EuclideanRing a => a -> Int
    Data.Ord.abs               :: forall a. Ord a => Ring a => ... -> ...
    Data.Ord.signum            :: forall a. Ord a => Ring a => ... -> ...
    Data.Ring.negate           :: forall a. Ring a => a -> a
    Data.Semiring.one          :: forall a. Semiring a => a
    Data.Semiring.zero         :: forall a. Semiring a => a

  in value declaration foo
```

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -144,6 +144,7 @@ data SimpleErrorMessage
   | UnusedTypeVar Text
   | WildcardInferredType SourceType Context
   | HoleInferredType Text SourceType Context (Maybe TypeSearch)
+  | UnknownValueHint Text SourceType Context (Maybe TypeSearch)
   | MissingTypeDeclaration Ident SourceType
   | OverlappingPattern [[Binder]] Bool
   | IncompleteExhaustivityCheck
@@ -819,6 +820,10 @@ data Expr
   -- A typed hole that will be turned into a hint/error during typechecking
   --
   | Hole Text
+  -- |
+  -- An unknown name that will be turned into a hint/error during typechecking
+  --
+  | UnknownValue Text
   -- |
   -- A value with source position information
   --
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -153,6 +153,7 @@ errorCode em = case unwrapErrorMessage em of
   UnusedTypeVar{} -> "UnusedTypeVar"
   WildcardInferredType{} -> "WildcardInferredType"
   HoleInferredType{} -> "HoleInferredType"
+  UnknownValueHint{} -> "UnknownValueHint"
   MissingTypeDeclaration{} -> "MissingTypeDeclaration"
   OverlappingPattern{} -> "OverlappingPattern"
   IncompleteExhaustivityCheck{} -> "IncompleteExhaustivityCheck"
@@ -883,6 +884,31 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
           _ -> []
       in
         paras $ [ line $ "Hole '" <> markCode name <> "' has the inferred type "
+                , markCodeBox (indent (typeAsBox maxBound ty))
+                ] ++ tsResult ++ renderContext ctx
+    renderSimpleErrorMessage (UnknownValueHint name ty ctx ts) =
+      let
+        maxTSResults = 15
+        tsResult = case ts of
+          Just (TSAfter{tsAfterIdentifiers=idents}) | not (null idents) ->
+            let
+              formatTS (names, types) =
+                let
+                  idBoxes = Box.text . T.unpack . showQualified id <$> names
+                  tyBoxes = (\t -> BoxHelpers.indented
+                              (Box.text ":: " Box.<> typeAsBox prettyDepth t)) <$> types
+                  longestId = maximum (map Box.cols idBoxes)
+                in
+                  Box.vcat Box.top $
+                      zipWith (Box.<>)
+                      (Box.alignHoriz Box.left longestId <$> idBoxes)
+                      tyBoxes
+            in [ line "You could substitute the unknown value with one of these values:"
+               , markCodeBox (indent (formatTS (unzip (take maxTSResults idents))))
+               ]
+          _ -> []
+      in
+        paras $ [ line $ "Unknown name '" <> markCode name <> "' has the inferred type "
                 , markCodeBox (indent (typeAsBox maxBound ty))
                 ] ++ tsResult ++ renderContext ctx
     renderSimpleErrorMessage (MissingTypeDeclaration ident ty) =
--- a/src/Language/PureScript/Sugar/Names.hs
+++ b/src/Language/PureScript/Sugar/Names.hs
@@ -10,7 +10,7 @@ module Language.PureScript.Sugar.Names
   ) where
 
 import Prelude.Compat
-import Protolude (ordNub, sortBy, on)
+import Protolude (ordNub, sortBy, on, Text)
 
 import Control.Arrow (first)
 import Control.Monad
@@ -268,17 +268,30 @@ renameInModule imports (Module modSS coms mn decls exps) =
     unless (length (ordNub args) == length args) .
       throwError . errorMessage' pos $ OverlappingNamesInLet
     return ((pos, args ++ bound), Let w ds val')
-  updateValue (_, bound) (Var ss name'@(Qualified Nothing ident)) | ident `notElem` bound =
-    (,) (ss, bound) <$> (Var ss <$> updateValueName name' ss)
-  updateValue (_, bound) (Var ss name'@(Qualified (Just _) _)) =
-    (,) (ss, bound) <$> (Var ss <$> updateValueName name' ss)
-  updateValue (_, bound) (Op ss op) =
-    (,) (ss, bound) <$> (Op ss <$> updateValueOpName op ss)
-  updateValue (_, bound) (Constructor ss name) =
-    (,) (ss, bound) <$> (Constructor ss <$> updateDataConstructorName name ss)
+  updateValue (_, bound) (Var ss name'@(Qualified Nothing ident)) | ident `notElem` bound = do
+    updateOrReplaceValue (ss, bound) name' updateValueName' (Var ss) showIdent
+  updateValue (_, bound) (Var ss name'@(Qualified (Just _) _)) = do
+    updateOrReplaceValue (ss, bound) name' updateValueName' (Var ss) showIdent
+  updateValue (_, bound) (Op ss op) = do
+    updateOrReplaceValue (ss, bound) op updateValueOpName' (Op ss) showOp
+  updateValue (_, bound) (Constructor ss name) = do
+    updateOrReplaceValue (ss, bound) name updateDataConstructorName' (Constructor ss) runProperName
   updateValue s (TypedValue check val ty) =
     (,) s <$> (TypedValue check val <$> updateTypesEverywhere ty)
   updateValue s v = return (s, v)
+
+  updateOrReplaceValue
+    :: (SourceSpan, [Ident])
+    -> Qualified a
+    -> (Qualified a -> SourceSpan -> m (Maybe (Qualified a)))
+    -> (Qualified a -> Expr)
+    -> (a -> Text)
+    -> m ((SourceSpan, [Ident]), Expr)
+  updateOrReplaceValue (ss, bounds) name updateExpr mkExpr showA = do
+    updated <- updateExpr name ss
+    return $ case updated of
+      Just updated' -> ((ss, bounds), mkExpr updated')
+      Nothing -> ((ss, bounds), UnknownValue (showQualified showA name))
 
   updateBinder
     :: (SourceSpan, [Ident])
@@ -367,6 +380,12 @@ renameInModule imports (Module modSS coms mn decls exps) =
     -> m (Qualified (ProperName 'ConstructorName))
   updateDataConstructorName = update (importedDataConstructors imports) DctorName
 
+  updateDataConstructorName'
+    :: Qualified (ProperName 'ConstructorName)
+    -> SourceSpan
+    -> m (Maybe (Qualified (ProperName 'ConstructorName)))
+  updateDataConstructorName' = update' (importedDataConstructors imports) DctorName
+
   updateClassName
     :: Qualified (ProperName 'ClassName)
     -> SourceSpan
@@ -376,11 +395,20 @@ renameInModule imports (Module modSS coms mn decls exps) =
   updateValueName :: Qualified Ident -> SourceSpan -> m (Qualified Ident)
   updateValueName = update (importedValues imports) IdentName
 
+  updateValueName' :: Qualified Ident -> SourceSpan -> m (Maybe (Qualified Ident))
+  updateValueName' = update' (importedValues imports) IdentName
+
   updateValueOpName
     :: Qualified (OpName 'ValueOpName)
     -> SourceSpan
     -> m (Qualified (OpName 'ValueOpName))
   updateValueOpName = update (importedValueOps imports) ValOpName
+
+  updateValueOpName'
+    :: Qualified (OpName 'ValueOpName)
+    -> SourceSpan
+    -> m (Maybe (Qualified (OpName 'ValueOpName)))
+  updateValueOpName' = update' (importedValueOps imports) ValOpName
 
   updateKindName
     :: Qualified (ProperName 'KindName)
@@ -427,3 +455,23 @@ renameInModule imports (Module modSS coms mn decls exps) =
 
     where
     throwUnknown = throwError . errorMessage . UnknownName . fmap toName $ qname
+
+  -- Update names so unqualified references become qualified, and locally
+  -- qualified references are replaced with their canoncial qualified names
+  -- (e.g. M.Map -> Data.Map.Map).
+  update'
+    :: (Ord a)
+    => M.Map (Qualified a) [ImportRecord a]
+    -> (a -> Name)
+    -> Qualified a
+    -> SourceSpan
+    -> m (Maybe (Qualified a))
+  update' imps toName qname@(Qualified mn' name) pos = warnAndRethrowWithPosition pos $
+    case (M.lookup qname imps, mn') of
+      (Just options, _) -> do
+        (mnNew, mnOrig) <- checkImportConflicts pos mn toName options
+        modify $ \usedImports ->
+          M.insertWith (++) mnNew [fmap toName qname] usedImports
+        return $ Just $ Qualified (Just mnOrig) name
+
+      _ -> return Nothing
--- a/tests/purs/failing/MultipleErrors2.purs
+++ b/tests/purs/failing/MultipleErrors2.purs
@@ -1,5 +1,4 @@
--- @shouldFailWith UnknownName
--- @shouldFailWith UnknownName
+-- @shouldFailWith UnknownValueHint
 module MultipleErrors2 where
 
 import Prelude
--- a/src/Language/PureScript/Pretty/Values.hs
+++ b/src/Language/PureScript/Pretty/Values.hs
@@ -94,6 +94,7 @@ prettyPrintValue d (TypedValue _ val _) = prettyPrintValue d val
 prettyPrintValue d (PositionedValue _ _ val) = prettyPrintValue d val
 prettyPrintValue d (Literal _ l) = prettyPrintLiteralValue d l
 prettyPrintValue _ (Hole name) = text "?" <> textT name
+prettyPrintValue _ (UnknownValue name) = textT name
 prettyPrintValue d expr@AnonymousArgument{} = prettyPrintValueAtom d expr
 prettyPrintValue d expr@Constructor{} = prettyPrintValueAtom d expr
 prettyPrintValue d expr@Var{} = prettyPrintValueAtom d expr
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -157,6 +157,12 @@ typesOf bindingGroupType moduleName vals = withFreshSubstitution $ do
               (substituteType subst)
               (uncurry TSAfter (typeSearch cons env st (substituteType subst ty)))
         in ErrorMessage hints (HoleInferredType x ty y (Just searchResult))
+      ErrorMessage hints (UnknownValueHint x ty y (Just (TSBefore env))) ->
+        let subst = checkSubstitution st
+            searchResult = onTypeSearchTypes
+              (substituteType subst)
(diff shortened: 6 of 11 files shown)


tests/purs/failing/MultipleErrors2.purs
```diff
@@ -1,5 +1,4 @@
--- @shouldFailWith UnknownName
--- @shouldFailWith UnknownName
```
colinwahl: Previously, multiple `UnknownName` errors would be reported if there were multiple unknown names, but now only the first one will be reported.  If that behavior is OK, then I should delete this test.

kRITZCREEK: I think that's an unfortunate regression we should think about. It's tricky because the type checker does abort with a single error and doesn't let you give out multiple ones. Also with the much larger size of the new error messages, reporting multiple of them might end up flooding the terminal too much.

hdgarrood:
I'm not sure about the "you could subsitute with one of these values" section; usually, when I encounter an UnknownName error, it's because I:

* forgot to import something,
* used a function I was intending to implement but haven't done so yet, or
* misspelled a function name.

In each case I know what I want to put there, so having a list of other things I could put there is unlikely to be helpful, especially because we already have holes for asking the compiler "what can I put here". I would prefer either having a "did you mean: something" section based mostly on edit distance for misspellings, or not having it at all.

tests/purs/failing/MultipleErrors2.purs
hdgarrood: Yeah I'd much prefer to keep the behaviour of reporting multiple errors if we possibly can.

spicydonuts:
> * used a function I was intending to implement but haven't done so yet, or

The "has the inferred type" section above seems to cover this case

> * forgot to import something,

Perhaps a "did you mean to import [something] from one of these modules?" message is also displayed when the name is an exact match for an existing unimported function

> * misspelled a function name.

If there are functions within a small edit distance list these as "did you mean...", possibly prioritizing functions which also appear in the list of matches for the type hole

hdgarrood:
Just to clarify: yes, I think suggesting values of the unknown name which haven't been imported or suggesting values a small edit distance away from that name would both be fantastic, but both of these are probably best tackled separately; for now, just adding the inferred type would probably be best.

colinwahl:
@hdgarrood I agree that the plain type search results without some sort of edit distance comparison isn't the most helpful. I've taken it out, and the error message is now a lot more readable.

@spicydonuts I think there is a lot we can do with edit distance comparison, and I really like the idea of suggesting possibly forgotten imports.  I agree with Harry that this could be done in a follow up PR.

For now, I have updated this PR to:
* remove type search results from `UnknownValueHint` error
* move reporting of `UnknownValueHint` to the module level so that multiple `UnknownValueHint` errors can be reported for a single module

Here is an updated example:

```
module Main where

import Prelude

import Effect (Effect)

bar :: Effect Unit
bar = do
  boo <- a 10
  let x = foo "hello"
  pure boo
```

Previously, this would fail with an `UnknownName` error, but now it fails with a `UnknownValueHint` error, stating

```
[1/2 UnknownValueHint] src/Main.purs:9:10

  9    boo <- a 10
              ^

  Unknown name 'a' has the inferred type

    Int -> Effect Unit

  in value declaration bar

[2/2 UnknownValueHint] src/Main.purs:10:11

  10    let x = foo "hello"
                ^^^

  Unknown name 'foo' has the inferred type

    String -> t0

  in the following context:

    boo :: Unit

  in value declaration bar

  where t0 is an unknown type
```

src/Language/PureScript/TypeChecker.hs
```diff
@@ -516,7 +516,7 @@ typeCheckModule (Module _ _ _ _ Nothing) =
 typeCheckModule (Module ss coms mn decls (Just exps)) =
   warnAndRethrow (addHint (ErrorInModule mn)) $ do
     modify (\s -> s { checkCurrentModule = Just mn })
-    decls' <- typeCheckAll mn exps decls
+    decls' <- escalateWarningWhen unknownValueHintError $ typeCheckAll mn exps decls
```
colinwahl: By escalating `UnknownValueHint` warnings into errors after the whole module is type checked, all `UnknownValueHint`s are collected before the `throwError` happens, fixing the multiple errors regression.

tests/purs/failing/MultipleErrors2.purs
colinwahl: I was able to fix this! See https://github.com/purescript/purescript/pull/3619#pullrequestreview-233321885

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -144,6 +144,7 @@ data SimpleErrorMessage
   | UnusedTypeVar Text
   | WildcardInferredType SourceType Context
   | HoleInferredType Text SourceType Context (Maybe TypeSearch)
+  | UnknownValueHint Text SourceType Context
```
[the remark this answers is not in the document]
colinwahl: Sure, I changed the arguments to `UnknownName` to support this

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -427,3 +454,23 @@ renameInModule imports (Module modSS coms mn decls exps) =
 
     where
     throwUnknown = throwError . errorMessage . UnknownName . fmap toName $ qname
+
+  -- Update names so unqualified references become qualified, and locally
+  -- qualified references are replaced with their canoncial qualified names
+  -- (e.g. M.Map -> Data.Map.Map).
+  update'
```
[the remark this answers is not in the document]
colinwahl: Sure, I've cleaned it up a bit!

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -83,7 +91,7 @@ data SimpleErrorMessage
   | OrphanTypeDeclaration Ident
   | RedefinedIdent Ident
   | OverlappingNamesInLet
-  | UnknownName (Qualified Name)
+  | UnknownName UnknownNameParams
```
hdgarrood: I think that when a particular sum type will always include a particular piece of information, it's often better to normalize it so that that field isn't repeated across different variants of the sum. In this case an `UnknownName` error will always include the name which was unknown, so I'd prefer to leave it as a `Qualified Name` and always have that as the first field of the constructor, rather than using `Either` and representing it as either a `Qualified Name` in the `Right` case or as a `Text` in the `Left` case . So I think I'd prefer:
```haskell
| UnknownName (Qualified Name) (Maybe (SourceType, Context))
```
and then the second field will only be present in the case of unknown name errors at the term level where we can use the type checker to provide some additional information. Do you think that would work, or am I missing something and was there a particular reason you went for `Either` instead?

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -820,6 +828,10 @@ data Expr
   --
   | Hole Text
   -- |
+  -- An unknown name that will be turned into a hint/error during typechecking
+  --
+  | UnknownValue Text
```
hdgarrood: I think this should be a `Name` so that we can easily keep track of what kind of name it is (i.e. whether it is a data constructor, a normal identifier, or an operator) and surface that in the error message easily.

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -269,17 +269,29 @@ renameInModule imports (Module modSS coms mn decls exps) =
...
   updateValue s v = return (s, v)
 
+  updateOrReplaceValue
+    :: (SourceSpan, [Ident])
+    -> Qualified a
+    -> (Qualified a -> SourceSpan -> m (Maybe (Qualified a)))
+    -> (Qualified a -> Expr)
+    -> (a -> Text)
+    -> m ((SourceSpan, [Ident]), Expr)
+  updateOrReplaceValue (ss, bounds) name updateExpr mkExpr showName = do
```
hdgarrood: Do you think it would make more sense to call the third argument `updateName`?

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -376,54 +394,79 @@ renameInModule imports (Module modSS coms mn decls exps) =
...
+  -- Returns Nothing if name can't be qualified, so that errors can be thrown
+  -- or substitutions can be made.
+  -- (e.g. M.Map -> Data.Map.Map).
+  update'
+    :: (Ord a)
+    => M.Map (Qualified a) [ImportRecord a]
+    -> (a -> Name)
+    -> Qualified a
+    -> SourceSpan
+    -> m (Maybe (Qualified a))
```
hdgarrood: Assuming it makes sense to have the `UnknownValue` and `UnknownName` constructors use `Name` instead of `Text` for unknown names, presumably we could return an `Either Name (Qualified a)` here so that in the case that the name is not known, we immediately have access to a `Name` to use in the error? Then we don't need to supply the appropriate `Name` constructor more than once.

src/Language/PureScript/Sugar/Names.hs
colinwahl: Yep, I think it does

src/Language/PureScript/AST/Declarations.hs
colinwahl: I didn't realize how much `printName` did already. You're right, it is super helpful.

src/Language/PureScript/AST/Declarations.hs
colinwahl: There wasn't a particular reason I went for `Either`.  I am changing it over now!

src/Language/PureScript/Sugar/Names.hs
colinwahl: I made `UnknownValue` and `UnknownName` take a `Qualified Name`, so I'm not sure this if this applies anymore.  Let me know if you had something else in mind

colinwahl:
I just realized a problem with the current implementation - the holes created by an unknown value (or just a plain Hole) can spuriously cause `AmbiguousTypeVariables` or `NoInstanceFound` caused by unknown types.  Here's an example:

```
module Main where

import Prelude

import Effect (Effect)

bar :: Effect Unit
bar = ?a \_ -> do
  pure 10
```

And 

```
module Main where

import Prelude

import Effect (Effect)

bar :: Effect Unit
bar = foo \_ -> do
  pure 10
```

Both fail with  
```
[1/1 NoInstanceFound] src/Main.purs:9:3

  9    pure 10
       ^^^^^^^

    Control.Applicative.Applicative t0

  while applying a function pure
    of type Applicative t0 => t1 -> t0 t1
    to argument 10
  while inferring the type of pure 10
  in value declaration bar

  where t0 is an unknown type
        t1 is an unknown type
```

While the real error in both of them is that there is a Hole or an unknown value.  This is because the `UnknownName`/`Hole` errors are first collected as warnings and then escalated as errors later, whereas other errors are thrown immediately.

 I'm not sure how big of a deal this is with Holes (or if it can even be helped), but for `UnknownName` errors, I think that this is incorrect behavior.  It should report the unknown name before reporting errors down the line.

I'm thinking of ways to remedy this.  One idea that I have is to modify some of the `throwError`s to check for `UnknownName` warnings that have accumulated, and report those first before any other error.

paf31:
You could wrap the original `UnknownName` error you would have thrown inside the error constructor, and modify the error rendering function to handle that case specially, perhaps.

src/Language/PureScript/Sugar/Names.hs
hdgarrood: Yes, this still applies; at the moment, we have to supply a `toName` function in two places: once as an argument to `update'` and once as an argument to `updateOrReplaceValue`. If `update'` returned an `Either Name (Qualified a)` then we wouldn't need to have `updateOrReplaceValue` take a `mkName` parameter, because the `Name` we want is right there already.

colinwahl: Ah yep, you're right, I missed the point at first. I believe I've addressed this here https://github.com/purescript/purescript/pull/3619/commits/c8abc2eb8e812eb029a68a5622cb2fa369cdb262

colinwahl:
Okay, I think I have a solution for the issue I described above.

I solved it by:
* adding an error `HoleCannotInferType` indicating that the type of a hole could not be inferred
* adding a piece of state `inferringHoleError` to the type checker state
* adding utilities to throw an error, taking `inferringHoleError` into account and throwing the error in state if it exists
* adding source spans to give precise position information to the errors in `inferringHoleError`

One concern I still have is that if you have a file like:
```
module Main where

import Prelude (pure)
import Effect (Effect)

bar :: Effect Int
bar = foo \_ -> do
  pure 10

bar' :: Int
bar' = foo
```

Then there are two `UnknownName`s.  However, when compiling, one appears as an error (the one that was caused by the error inferring the type of `bar`) and the other as a warning.  
```
[1/1 UnknownName] src/Main.purs:11:8

  11  bar' = foo
             ^^^

  Unknown value foo has the inferred type

    Int

  in value declaration bar'

[1/1 UnknownName] src/Main.purs:7:7

  7  bar = foo \_ -> do
           ^^^

  Unknown value foo

  in value declaration bar
```
I know this is because I am throwing the `UnknownName` error before escalating the `UnknownName` warnings into errors (which is done at the module level), but I am not sure how to fix it right now.

colinwahl:
Sorry for the delay on cleaning this up - it should be ready for another look now.

I left `UnknownName` and `HoleInferredType` errors exactly the same as they were before, and I modified `NoInstanceFound` and `AmbiguousTypeVariables` errors to wrap the `UnknownName` and `HoleInferredType` error constructors, and I added specific error rendering function in the case where an unknown name or hole occurs in the binding group that has a `NoInstanceFound` or `HoleInferredType` error.  This helps debug errors caused by https://github.com/purescript/purescript/pull/3619#issuecomment-490740397 while still maintaining all the old error functionality.

Here is what some errors would look like now:

```
foo :: Effect Unit
foo = ?b \_ ->
  pure 10

[1/1 NoInstanceFound] src/Main.purs:16:3

  16    pure 10
        ^^^^^^^

    Control.Applicative.Applicative t0

  The instance head contains unknown type variables.
  This could have been caused by the hole 'b'. Consider adding a type annotation.

  while applying a function pure
    of type Applicative t0 => t1 -> t0 t1
    to argument 10
  while inferring the type of pure 10
  in value declaration foo'''

  where t2 is an unknown type
        t0 is an unknown type
        t1 is an unknown type
```

```
foo :: Effect Unit
foo = bar \_ ->
  pure 10

[1/1 NoInstanceFound] src/Main.purs:37:3

  37    pure 10
        ^^^^^^^

    Control.Applicative.Applicative t0

  The instance head contains unknown type variables.
  This could have been caused by the unknown value bar. Consider adding a type annotation.

  while applying a function pure
    of type Applicative t0 => t1 -> t0 t1
    to argument 10
  while inferring the type of pure 10
  in value declaration foo

  where t0 is an unknown type
        t1 is an unknown type
```

```
foo = ?b \_ ->
  pure 10

[1/1 AmbiguousTypeVariables] src/Main.purs:61:1

      v
  61  foo = ?b \_ ->
  62    pure 10
              ^

  The inferred type

    forall t4 t6. Applicative t4 => t6

  has type variables which are not mentioned in the body of the type.
  This could have been caused by the hole 'b'. Consider adding a type annotation.

  in value declaration foo

  where t0 is an unknown type
        t1 is an unknown type
        t2 is an unknown type
```

```
foo = bar \_ ->
  pure 10

[1/1 AmbiguousTypeVariables] src/Main.purs:86:1

      v
  86  foo = bar \_ ->
  87    pure 10
              ^

  The inferred type

    forall t4 t6. Applicative t4 => t6

  has type variables which are not mentioned in the body of the type.
  This could have been caused by the unknown value bar. Consider adding a type annotation.

  in value declaration foo
```

I made a first pass at what the error messages could say, but they could use some input.

colinwahl:
I wasn't satisfied with the stateful approach I originally took, so I have created an explicit AST traversal that takes an `Expr` and returns all of the holes and unknown values within that `Expr`.  This allows throwing all unknown name errors first in the case of a `NoInstanceFound` or `AmbiguousTypeVariables` error (which could have been caused by an unknown name hole in the first place).  I have also added an extra hint message when the error is possibly caused by a hole, as I illustrated above.

Other than the problems with unknown name holes causing `NoInstanceFound` or `AmbiguousTypeVariables` errors (which should be fixed now), this seems to be working really well.  I've been using it at work for about a week and I haven't run into any issues.

Let me know of any feedback, and I'd be happy to address it and try to get this over the line!

colinwahl:
I have made a few changes that hopefully make this PR more focused/easier to review.  At first I started to make possible improvements to errors messages for errors that are possibly caused by the existence of typed holes.  However, in favor of having this PR be smaller and easier to review, I have reverted those changes and only included changes pertaining to the original issue.  I can do follow up work on the other issues in subsequent PRs, if wanted.  

The main change since @hdgarrood's last review adds a mechanism to avoid unknown name holes from causing spurious `AmbiguousTypeVariable` and `NoInstanceFound` errors.  To do this, I added a function which throws `UnknownName` errors contained in an expression which has an `AmbiguousTypeVariable` or `NoInstanceFound` error _before_ throwing said error.  Note that this bypasses the logic which raises all unknown value holes into errors at the module level, meaning that if there are unknown names in some other binding group in the module, they will not be reported as errors.  However, I think that this case is infrequent enough that this is acceptable.

colinwahl:
Has anyone had a chance to look over this?  I'm happy to change my approach with a little guidance, if needed!

natefaubion:
> Note that this bypasses the logic which raises all unknown value holes into errors at the module level, meaning that if there are unknown names in some other binding group in the module, they will not be reported as errors. However, I think that this case is infrequent enough that this is acceptable.

Can you give me an example that triggers this case?

src/Language/PureScript/Errors.hs
```diff
@@ -534,10 +541,14 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
-    renderSimpleErrorMessage (UnknownName name@(Qualified Nothing (IdentName (Ident i)))) | i `elem` [ C.bind, C.discard ] =
+    renderSimpleErrorMessage (UnknownName name@(Qualified Nothing (IdentName (Ident i))) Nothing) | i `elem` [ C.bind, C.discard ] =
       line $ "Unknown " <> printName name <> ". You're probably using do-notation, which the compiler replaces with calls to the " <> markCode i <> " function. Please import " <> markCode i <> " from module " <> markCode "Prelude"
-    renderSimpleErrorMessage (UnknownName name) =
+    renderSimpleErrorMessage (UnknownName name Nothing) =
       line $ "Unknown " <> printName name
+    renderSimpleErrorMessage (UnknownName name (Just (ty, ctx))) =
+      paras $ [ line $ "Unknown " <> printName name <> " has the inferred type "
+              , markCodeBox (indent (typeAsBox maxBound ty))
+              ] ++ renderContext ctx
```
natefaubion: Something about this error message isn't right to me. I know it was just "Unknown foo" before which is also terrible, but adding "has the inferred type" makes it seem less like an error.

Maybe something like for unqualified
> The name foo is not in scope.

> The name foo is not in scope. It has the inferred type ...

And for qualified:
> The name foo is not exported by module Foo.

> The name foo is not exported by module Foo. It has the inferred type ...

@hdgarrood @colinwahl What do you think?

colinwahl:
> > Note that this bypasses the logic which raises all unknown value holes into errors at the module level, meaning that if there are unknown names in some other binding group in the module, they will not be reported as errors. However, I think that this case is infrequent enough that this is acceptable.
> 
> Can you give me an example that triggers this case?

Sure, something like this:

```
module Test where

import Prelude
import Effect (Effect)

foo :: Effect Unit
foo = b \_ ->
  pure 10

bar = f 10
```

Will fail with one `UnknownName` error (for `b`), instead of two.
Now that I look at it again I think that the error has the wrong source span - I'm looking into that now

src/Language/PureScript/Errors.hs
colinwahl: I like the idea of having different errors depending on whether or not the name is qualified!  This particular error did seem very terse - I think this is a improvement

colinwahl: @natefaubion Actually, now that I think about it, this may be missing a couple of cases.  You could have some declaration
```
baz :: Effect Int
baz = do
  Bar.bar
```
Where the module `Bar` does export a value `bar`, but there just isn't an import for it.  In this case, I don't think that the error `The name bar is not exported by module Foo.` Is an appropriate error.

Maybe something like `The name bar is not imported from module Bar. ...` Would be more accurate.  What do you think?

colinwahl:
@natefaubion I pushed a proposed update to the typechecking logic for the cases where there is an unknown name in a expression.  It alleviates the problem of throwing a spurious (`NoInstanceFound`/`AmbiguousTypeVariables`)  The two cases that cause problems (`NoInstanceFound` and `AmbiguousTypeVariables`) can only be caused when the typechecker is solving unsolved types.  To remedy this, one thing we could do is not do the processing of unknown types when an expression contains an unknown name.  This will sometimes cause errors like

```
[3/4 UnknownName] src/Test.purs:10:7

  10  bar = f 10
            ^

  Unknown value f has the inferred type

    Int -> t0

  in value declaration bar
```

But at least we have given some indication of what the type should be, and we have also prevented the spurious errors that were previously caused.  This has the added benefit of not reporting the wrong source spans, like my previous attempt did.  What do you think?  Is it OK to report unknown types for unknown name errors?

src/Language/PureScript/Errors.hs
afcondon: This one has actually bitten me one more than one occasion - not very serious - but it can just nudge you into haring off looking for the export in the imported file instead of fixing the import. If you have any kind of doubts about your build system integrity (as seems to be too often the case) it can pitch you into a worse yak-shaving-wild-goose-chase digression there too.

colinwahl:
@natefaubion bumping this PR. I still am unsure of the best way to resolve the `AmbiguousTypeVariables`/`NoInstanceFound` errors that can be caused by trying to infer the types of unknown names.  I have a proposed solution above but it's definitely not the best.

Thanks!

Issue #3558, opened 2019-03-11
Dictionaries "get lost" when using a higher order function vs pattern matching a constructor


garyb:
I don't know how to phrase this really, but:

``` purescript
newtype LBox row a = LBox (∀ r. (∀ lbl _1. Row.Cons lbl a _1 row ⇒ IsSymbol lbl ⇒ SProxy lbl → r) → r)

unLBox ∷ ∀ row a r. (∀ lbl _1. Row.Cons lbl a _1 row ⇒ IsSymbol lbl ⇒ SProxy lbl → r) → LBox row a → r
unLBox g (LBox f) = f g

readYup ∷ ∀ row a. Record row → LBox row a → a
readYup rec (LBox f) = f \lbl → Record.get lbl rec

readNope ∷ ∀ row a. Record row → LBox row a → a
readNope rec = unLBox \lbl → Record.get lbl rec
```

`readNope` fails to type check with:

```
No type class instance was found for

    Prim.Row.Cons lbl4
                  a5
                  t2
                  row6
```

even though that dictionary should be availble in the function passed to `unLBox`

garyb:
Hmm, maybe it's not getting lost as such, since there's a `t2` in that error - but there's something not quite right, as really these two cases shouldn't behave differently.

garyb:
It's also cool if you try and write `unLBox` like this:

``` purescript
unLBox ∷ ∀ row a r. LBox row a → (∀ lbl _1. Row.Cons lbl a _1 row ⇒ IsSymbol lbl ⇒ SProxy lbl → r) → r
unLBox (LBox f) = f
```

It raises my least favourite error:
```
  Could not match constrained type

    Cons t3 a0 t4 row1 => IsSymbol t3 => SProxy t3 -> r2

  with type

    Cons lbl5 a0 _17 row1 => IsSymbol lbl5 => SProxy lbl5 -> t9
```
But that's definitely a separate thing.

garyb:
Actually, here's some simpler definitions, just reconstructing an `LBox`, that exhibit both problems:

``` purescript
yup ∷ ∀ row a. LBox row a -> LBox row a
yup (LBox f) = LBox \lbl → f lbl

-- Could not match constrained type
nopeConstrained ∷ ∀ row a. LBox row a -> LBox row a
nopeConstrained (LBox f) = LBox f

-- No type class instance was found for
nopeNoInstance ∷ ∀ row a. LBox row a → LBox row a
[...]

PR #3620, opened 2019-05-04
Unify matching constraints

### LINKED ISSUE #950: Eta expansion is required for higher rank types with constraints

### LINKED ISSUE #3588: Confusing interaction with rank-n types and constraints
Originally reported by @munizart in Slack. I think this is a bug.
```purescript
module Main where

data Tuple a b = Tuple a b

class C a

type SomeC = forall x. C x => x

-- This is fine
id1 :: SomeC -> SomeC
id1 x = x

-- This is not
id2 :: forall a. Tuple a SomeC -> Tuple a SomeC
id2 x = x

-- Neither is this
id3 :: Array SomeC -> Array SomeC
id3 x = x
```

The function `id2` produces an error
```
[1/1 ConstrainedTypeUnified] src/Main.purs:15:9

  15  id2 x = x
              
  
    C x1 => x1
  
  with type
  
    C x1 => x1
  
  while trying to match type C x1 => x1
    with type C x1 => x1
  while checking that expression x
    has type Tuple a0 (forall x. C x => x)
  in value declaration id2
  
        x1 is a rigid type variable
          bound at (line 7, column 21 - line 7, column 22)
```

and `id3` also produces a very similar error.

garyb:
This might well be a terrible idea, I feel like if it was this easy then it would already have been done. In https://github.com/purescript/purescript/issues/950#issuecomment-139434920 Phil said:
> We could add a case to the unify function for when two instance heads are equal, but I'm not confident it would be a good idea.

but he doesn't mention why it might not be good. I assume what he meant there is what I've done here too.

It seems to work great in fixing cases where we get the error currently; it resolves #950, it resolves #3588, and it also resolves the case I mentioned in #3558 where the obvious implementation of `unLBox` fails. 

I don't know how to check if it starts letting things in it shouldn't - I can't think of a case that would occur in.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Unify.hs
+++ b/src/Language/PureScript/TypeChecker/Unify.hs
@@ -24,6 +24,7 @@ import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State.Class (MonadState(..), gets, modify)
 import Control.Monad.Writer.Class (MonadWriter(..))
 
+import Data.Foldable (traverse_)
 import Data.Function (on)
 import Data.List (sortBy, nubBy)
 import qualified Data.Map as M
@@ -118,6 +119,10 @@ unifyTypes t1 t2 = do
   unifyTypes' r1 r2@RCons{} = unifyRows r1 r2
   unifyTypes' r1@REmpty{} r2 = unifyRows r1 r2
   unifyTypes' r1 r2@REmpty{} = unifyRows r1 r2
+  unifyTypes' (ConstrainedType _ c1 ty1) (ConstrainedType _ c2 ty2)
+    | constraintClass c1 == constraintClass c2 && constraintData c1 == constraintData c2 = do
+        traverse_ (uncurry unifyTypes) (constraintArgs c1 `zip` constraintArgs c2)
+        ty1 `unifyTypes` ty2
   unifyTypes' ty1@ConstrainedType{} ty2 =
     throwError . errorMessage $ ConstrainedTypeUnified ty1 ty2
   unifyTypes' t3 t4@ConstrainedType{} = unifyTypes' t4 t3


natefaubion:
> This might well be a terrible idea, I feel like if it was this easy then it would already have been done.

Maybe at the time it was more of an issue because of context tupling (`(Foo a, Bar b) => ...`)? I'm not sure when we removed that. But to me, with an enforced ordering, I don't see how it's any different than unifying an explicit argument.

hdgarrood:
It seems reasonable to me but I don't think my input counts for very much here. I would quite like to get Phil's input on this at least. @paf31 what are your thoughts? Do you remember what you had in mind when you wrote https://github.com/purescript/purescript/issues/950#issuecomment-139434920 ?

paf31:
I don't remember, but I think my objection to this was more that it was an incomplete solution to a bigger problem. As a simple example, if there are two constraints in a different order, this approach won't work, we'd need to elaborate to swap the dictionaries. So it seems like it's a trade-off between the better usability vs. a small risk of adding more confusion in certain cases (why does this case work, but this very similar case doesn't?) - but the current situation is already confusing, so it's probably no worse.

garyb:
Cool! Sounds like this is ok then.

Anecdotally I think I've only ever seen this in situations where the constraints match.

PR #3628, opened 2019-05-07
Fixes let/in expressions within ado

### LINKED ISSUE #3626: let-in expressions no longer seem to work in ado blocks
```
module Main where

main = ado
  x <- f $ let y = 0 in y
  in x
```
parses successfully in v0.12.5, but master fails on the first `in` with
```
Error found:
at src/Main.purs:4:22 - 4:22 (line 4, column 22 - line 4, column 22)

  Unable to parse module:
  Unexpected or mismatched indentation


See https://github.com/purescript/documentation/blob/master/errors/ErrorParsingModule.md for more information,
```

natefaubion:
An `in` keyword should only close a `let` _and_ `ado` block if it is part of an
`ado` let statement. This introduces `LytLetStmt` as it's own delimiter
which is determined by it's column in relation to an existing `ado` or
`do` block.

Fixes #3626

### DIFF
--- a/src/Language/PureScript/CST/Layout.hs
+++ b/src/Language/PureScript/CST/Layout.hs
@@ -48,6 +48,7 @@ data LayoutDelim
   | LytForall
   | LytTick
   | LytLet
+  | LytLetStmt
   | LytWhere
   | LytOf
   | LytDo
@@ -56,12 +57,13 @@ data LayoutDelim
 
 isIndented :: LayoutDelim -> Bool
 isIndented = \case
-  LytLet   -> True
-  LytWhere -> True
-  LytOf    -> True
-  LytDo    -> True
-  LytAdo   -> True
-  _        -> False
+  LytLet     -> True
+  LytLetStmt -> True
+  LytWhere   -> True
+  LytOf      -> True
+  LytDo      -> True
+  LytAdo     -> True
+  _          -> False
 
 isTopDecl :: SourcePos -> LayoutStack -> Bool
 isTopDecl tokPos = \case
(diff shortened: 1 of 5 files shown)


hdgarrood:
Nice :) do you think it might be worth adding a test case where we have a `let`-`in` expression spanning over multiple lines within an `ado` block as well? For example:
```purescript
module Main where

main = ado
  x <-
    f g
    $ let
        y = 0
      in
        y
  in x
```

natefaubion:
I can. These cases aren't really different as far as the layout algorithm goes, however. That is, these appearing across multiple lines is not what affects the algorithm, since it's column based.

natefaubion:
Or I should say, is there a specific edge case you are wanting to detect with that? If it's just multiple lines, then I don't think a specific case for ado is what we want, since delimiters across multiple lines is something done in general. I can however test things like

```purescriipt
module Main where

main = ado
  x <-
    f g
    $ let
        y = 0
  in
        y
  in x
```

That is, something obviously wrong, if you just want to see how it inserts things.

natefaubion:
Actually, the above is not obviously wrong 😆. Since this is a let expression context, it would parse the above fine.

hdgarrood:
Not really, I was mostly wanting to make sure the multi line example I gave parses successfully, because we have some code which looks a bit like that at Lumi.

natefaubion:
Perhaps some explanation for the above is in order, since while it "looks" wrong, it is accepted. The layout algorithm is a delimiter parser which masks out areas for indentation sensitivity. You can have areas of insensitivity or unrelated sensitivity based on the top context. As an example, indentation is not insensitive in obvious delimited contexts like `()`, `[]`, or `{}`. This is so that on can write:

```purescript
example = do
  guard $ (
    foo + bar
  ) == true
  foo
```

Or

```purescript
example = do
  let
    { foo
    , bar
    } = baz
  foo + bar
```

Specifically we do not want to insert a layout separate before the closing delimiter. Things like this are actually quite common in PS, and note that Haskell would not accept them. If you try:

```haskell
example = do
  let
    ( foo
    , bar
    ) = baz
  foo + bar
```

It will complain because it inserts a layout separator before the `,` and `)`. What PureScript does is more like what Python does. This rule applies to other delimiters though that may not seem so obvious such as `if/then/else` and `let/in`.

PR #3630, opened 2019-05-07
Improve parser error ErrRecordUpdateInCtr

hdgarrood:
When parsing a record literal, if we encounter a = after a label, we
emit a parse error, because = is only for record updates.

However, the error says "Expected ':', saw '='", which implies that ':'
would be the only valid token following a record label, although
a label could also be followed by '}' in a record pun.

### DIFF
--- a/src/Language/PureScript/CST/Errors.hs
+++ b/src/Language/PureScript/CST/Errors.hs
@@ -80,7 +80,7 @@ prettyPrintErrorMessage (ParserError {..}) = case errType of
   ErrBinderInDecl ->
     "Expected declaration, saw pattern"
   ErrRecordUpdateInCtr ->
-    "Expected ':', saw '='"
+    "Expected ':' or '}', saw '='"
   ErrRecordPunInUpdate ->
     "Expected record update, saw pun"
   ErrRecordCtrInUpdate ->


natefaubion:
> When parsing a record literal, if we encounter a = after a label, we emit a parse error, because = is only for record updates.

This isn't actually quite true. The parser parses `label = expr` exactly and just adds a special case failure for a better error, so it's never possible to see this error in the case of a pun.

https://github.com/purescript/purescript/blob/315759e9e72116d8066faefcc348dc8f69bff233/src/Language/PureScript/CST/Parser.y#L412

natefaubion:
Note that it does this so that it can report multiple errors like this instead of stopping immediately with a single error.

hdgarrood:
Ah I see - good to know, thanks.

hdgarrood:
I think this suggested change still makes sense though: if we write e.g.
```
foo = { bar = 0 }
```
we will get `Expected ':', saw '='` at the second `=`, but
```
foo = { bar }
```
would be valid there too.

natefaubion:
It would only be valid if they didn't write an expression as well (which they did 😛). Additionally, if they write:

```purescript
foo =
  { bar = 0
  , baz = 1
  }
```

They will get two `ErrRecordUpdateInCtr` errors, but `}` is not valid for both of them.

hdgarrood:
Ah yes, good points. I suppose if we have parsed an expression after the = it’s almost certain that the user meant to use the label : expr form rather than a pun.

Issue #3145, opened 2017-11-04
Use content hashes to determine whether a file needs to be recompiled


lpil:
Hello!

I believe currently the compiler looks at file modification times in order to determine if a file needs to be recompiled. If we altered the compiler to instead look at a hash of the contents of the file the compiler would be a little better at determining if the work needs to be done.

This would benefit CI systems where previously compiled modules have been cached and thus have an older timestamp while being up-to-date with the source file.

This feature came up from a chat on Slack RE build speed on CI :)

Thanks,
Louis

paf31:
Then you'd have to store the latest hashes somewhere right? The nice thing about the timestamp is that it's automatically stored in the file metadata.

lpil:
Yes, we'd have to keep them with the compiled modules.

kRITZCREEK:
Yeah, the `ExternsFile` would need a hash for the contents of the source file it was compiled from.

metaleap:
Suggest to make this optional (eg. via flag), because this seems to be an issue only with extra tooling involved (eg. CI), and time-stamp-based (file-system metadata) ---instead of dedicated explicit computing-managing-comparing of hashes--- works pleasantly fast and sufficiently correctly for the "no-frills, no-integrations local bare dev-env" cases. =)

paf31:
> Suggest to make this optional (eg. via flag)

We don't really do flags, except in exceptional circumstances. It tends to increase the number of ways things can go wrong.

paf31:
I guess I'd be fine with this change, but the hash would have to take into account all of the dependencies. It just doesn't seem worth it to me.

garyb:
I don't think speed was the motivation, more that the dates still have hiccups sometimes. Speed was actually my concern, but it'd be interesting to compare `stat`ing all the files with hashing, since #3087 is a thing, and kinda an annoying one at that.

hdgarrood:
Why would a hash need to take dependencies into account? Timestamps can't take changed dependencies into account, surely, so it's not clear to me why hashes would need to.

hdgarrood:
I'd quite like to go ahead and give this a go. As with @garyb, I think it's best to consider this change as a bug fix rather than a performance boost: having file modification times change, e.g. by switching branches in your local working directory, can make the compiler think that an output directory is up to date when it isn't, causing compile errors, and the fix is to `rm -r output`. If `rm -r output` ever fixes a compile error, that ought to be considered a bug, I think.

garyb:
[...]

Issue #2477, opened 2016-12-12
Building against precompiled dependencies


ElvishJerricco:
You can see an example of this [in this Nix setup I'm experimenting with](https://github.com/ElvishJerricco/nix-purescript-skeleton). The problem is that `psc` doesn't provide any means for depending on code you've already compiled. If you want to use precompiled code, you have to copy it into the `output` directory, and include the sources for that code in the list of sources for `psc`. This is at best clumsy, and is prone to issues with timestamps changing on copy. It'd be nice if there were some intermediate representation (corefn?) that preserved the necessary information for compiling against.

paf31:
The data format you want is the `externs.json` format, but there is no command line API for building against a set of externs files right now. We could add one though. The closest thing right now is `psc-ide`'s rebuild command.

ElvishJerricco:
Ah cool. Yea having a CLI for building against `externs.json` sounds like the perfect solution. Are you saying that `psc-ide` is already doing this sort of thing? Meaning that would be a good place to look if I wanted to try implementing this?

paf31:
Yes, `psc-ide`, PSCi and Try PureScript all use this method. It's handled by the `rebuild` function in `Make.hs`.

ElvishJerricco:
Cool. I'll take a look at it and let you know if I have the time to work on a PR for this.

paf31:
Great, thanks! If you do, let's make sure to hash out a basic plan for the command line interface here so that others can review it too.

ElvishJerricco:
Sure. My first thoughts are to make it similar to the classical C compiler flags for external libraries.

```bash
psc -L/path/to/somelib/output/ $mysources
```

This would tell `psc` that it can find the `externs.json` files for modules in `somelib` at `/path/to/somelib/output/ModuleName/externs.json`.

ElvishJerricco:
So I've been looking at this today. I think there needs to be more than just a flag for `psc`. With `ghc` in Nix, you can use `haskellPackages.ghcWithPackages (self: [ self.async self.transformers ... ])` to get a `ghc` that automatically compiles against whatever packages you need. So when you enter a nix-shell with such a `ghc`, you can use `ghc Main.hs`, or `ghci`, or even `cabal configure`, and they all just automatically include the packages you specified from Nix. I'm not sure, but I believe this is accomplished by replacing `ghc` with a script that calls `ghc` with the correct arguments (and the same goes for `ghci`).

[...]

PR #3635, opened 2019-05-12
Add `graph` command for graphing module dependencies

jmackie:
This PR proposes adding a `graph` command to the compiler, which takes module files as input (similar to `purs compile`) and returns a JSON description of the module dependency graph. Essentially this just surfaces the output of `Language.PureScript.ModuleDependencies.sortModules` with some serialization logic. 

I think this will be useful for tooling as it let's us be smarter about what actually needs to be compiled. For example, if I'm writing a bundling tool that takes an entrypoint module I can avoid passing the world to `purs compile` (e.g. `"src/**/*.purs" ".psc-package/*/*/*/src/**/*.purs"`) by first invoking this command. 

More concretely, I'd like to improve PureScript support in webpack and parcel, and having this available would really help.

It's very much a first draft, and there are probably a bunch of error cases I need to consider, but comments/feedback would be much appreciated :)

### DIFF
--- a/app/Command/Graph.hs
+++ b/app/Command/Graph.hs
@@ -1,0 +1,138 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TupleSections #-}
+
+module Command.Graph (command) where
+
+import           Control.Applicative (many)
+import           Control.Monad (unless, when, forM, foldM)
+import qualified Data.Aeson as A
+import           Data.Aeson ((.=))
+import           Data.Bool (bool)
+import qualified Data.Map as M
+import qualified Data.HashMap.Strict as HM
+import qualified Data.ByteString.Lazy as LB
+import qualified Data.ByteString.Lazy.UTF8 as LBU8
+import           Data.Text (Text)
+import qualified Language.PureScript as P
+import qualified Language.PureScript.CST as CST
+import           Language.PureScript.Errors.JSON
+import           Language.PureScript.Make (runMake)
+import           Language.PureScript.ModuleDependencies
+import qualified Options.Applicative as Opts
+import qualified System.Console.ANSI as ANSI
+import           System.Exit (exitSuccess, exitFailure)
+import           System.Directory (getCurrentDirectory)
+import           System.FilePath.Glob (glob)
+import           System.IO (hPutStr, hPutStrLn, stderr)
+import           System.IO.UTF8 (readUTF8FileT)
+
+data GraphOptions = GraphOptions
+  { graphInput      :: [FilePath]
+  , graphJSONErrors :: Bool
+  }
+
+graph :: GraphOptions -> IO ()
+graph GraphOptions{..} = do
+  input <- globWarningOnMisses (unless graphJSONErrors . warnFileTypeNotFound) graphInput
+  when (null input && not graphJSONErrors) $ do
+    hPutStr stderr $ unlines
+      [ "purs graph: No input files."
+      , "Usage: For basic information, try the `--help' option."
+      ]
+    exitFailure
+
+  moduleFiles <- readInput input
+  (makeResult, makeWarnings) <- runMake P.defaultOptions $ do
+    ms <- CST.parseModulesFromFiles id moduleFiles
+    let parsedModuleSig = moduleSignature . CST.resPartial
+    (_sorted, moduleGraph) <- sortModules (parsedModuleSig . snd) ms
+    let swap (a, b) = (b, a)
+    let pathMap = M.fromList
+                . fmap (swap . fmap (sigModuleName . parsedModuleSig))
+                $ ms
+    pure (moduleGraphToJSON pathMap moduleGraph)
+
+  printWarningsAndErrors True graphJSONErrors makeWarnings makeResult
+  case makeResult of
+    Left _ -> exitSuccess
+    Right Nothing -> undefined
+    Right (Just json) -> LB.putStr $ A.encode json
+
+  where
+  warnFileTypeNotFound :: String -> IO ()
+  warnFileTypeNotFound =
+    hPutStrLn stderr . ("purs compile: No files found using pattern: " <>)
+
+command :: Opts.Parser (IO ())
+command = graph <$> (Opts.helper <*> graphOptions)
+  where
+  graphOptions :: Opts.Parser GraphOptions
+  graphOptions =
+    GraphOptions <$> many inputFile
+                 <*> jsonErrors
+
+  inputFile :: Opts.Parser FilePath
+  inputFile =
+    Opts.strArgument $
+      Opts.metavar "FILE" <>
+      Opts.help "The input .purs file(s)."
+
+  jsonErrors :: Opts.Parser Bool
+  jsonErrors =
+    Opts.switch $
+      Opts.long "json-errors" <>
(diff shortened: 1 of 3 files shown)


hdgarrood:
I'd actually rather keep this logic internal to the compiler if we possibly can. The idea is that you do pass everything to `purs compile` every time, and it's the compiler's responsibility to only do the work which actually needs doing. Is the reason that you'd rather avoid this that it can be slow? Because if so there are other avenues which I'd rather take to improve this state of affairs. One example is the CST work which should greatly improve startup time for projects with lots of modules (which has now been merged). Another example is #2477, which would allow us to avoid doing extra work to parse and sort modules from dependencies every time.

jmackie:
Sure, but `purs compile` has no notion of an entrypoint right? So it's always gonna compile everything I pass to it at least once, which isn't great for large projects with lots of dependencies. And in the jenga-tower of frontend tooling I find myself having to `rm -rf output` more often then I'd like...

As another example: if I want to tell webpack/parcel what files need to be watched, I currently have to default to `"src/**/*.purs" ".psc-package/*/*/*/src/**/*.purs"`, which can be super resource hungry (I've hit linux open file limits because of this at work).

If you're building tooling on top of `purs` this is a pretty reasonable requirement though, no? Currently we can only throw a bunch of globs at the compiler, but it would be nice to have some insight into the build plan.

hdgarrood:
> So it's always gonna compile everything I pass to it at least once, which isn't great for large projects with lots of dependencies [...] I currently have to default to "`src/**/*.purs"` `".psc-package/*/*/*/src/**/*.purs"`

Yes; it should be possible to address both of these issues with something like #2477, so that any given dependency only needs to be built once (across any number of projects), and so that once a dependency has been built, we know it isn't going to change, so we don't need to watch those files.

>  And in the jenga-tower of frontend tooling I find myself having to rm -rf output more often then I'd like...

I agree. Whenever this is a result of the compiler misbehaving that's something I'd like to fix too. See #3145, for instance.

> If you're building tooling on top of purs this is a pretty reasonable requirement though, no? Currently we can only throw a bunch of globs at the compiler, but it would be nice to have some insight into the build plan.

Lots of compilers delegate the responsibility of constructing build plans to third party tools. This can be a big pain, so we've opted for a different approach in PureScript: that build plans should remain internal to the compiler. I think this approach has worked really well so far, and I still believe that the issues you're coming up against can be addressed while sticking to this approach. (I'm open to being persuaded otherwise but I'm not yet convinced.)

jmackie:
Fair points, and I respect that design decision. In which case I definitely
think #3145 and #2477 are worth looking at (probably in that order?) - I
reckon I could have a go at #3145 so I'll try and put a PR together. Thanks
for your help 👊

hdgarrood:
Welcome! You might want to check in with @joneshf as he has been looking into using Shake to implement this.

jmackie:
Ooo that sounds fun... 👀

joneshf:
I think this is a great thing to have! What would have to change to get this PR accepted?

joneshf:
Also, I'll throw a proposal for `shake` into #3145 so we can re-invigorate that conversation.

hdgarrood:
> What would have to change to get this PR accepted?

For me to get on board I’d want to hear a more persuasive case for it. As I say I think build plans should continue to be the responsibility of the compiler. I think having the compiler delegate the responsibility of calculating what needs rebuilding to external tools has the potential to greatly increase the complexity of build processes; if a number of people create tools which use that information in slightly different ways then we could very easily find ourselves grappling with lots of bugs which are difficult to reproduce and where the location of the problem is difficult to identify (eg “is this a bug in the compiler or in external tooling?”). I don’t think need to concede this ground to be able to make the compiler smarter and faster in the sense of not doing more work than it needs to in incremental builds.

natefaubion:
I don't think the module graph is an internal implementation detail. It's a static artifact of the modules you pass to it, which anyone can derive unambiguously by parsing them. Sure, if it contained information like what needs to be rebuilt, that would be different, but the hierarchy itself is just language semantics. I don't think there's anything wrong exposing it, but I would like a concrete use case where a tool really needs it before we commit to supporting it.

jmackie:
My initial motivation was adding PureScript support to [parcel](https://parceljs.org/). I could go with the usual glob paths and tell parcel that my `Entrypoint.purs` depends on  `"src/**/*.purs" ".psc-package/*/*/*/src/**/*.purs"` (or bower_components or whatever) and have it watch all that stuff and `purs` build all that stuff, or try and be a bit smarter with the help of this command.

joneshf:
For more motivation, any sort of large scale analysis would re-implement this graph algorithm. In a large codebase, it's very common and tedious to do this analysis by hand.

Some examples of what you can do with dependency graphs in the JS ecosystem:
* https://github.com/dyatko/arkit
* https://github.com/hughsk/colony
* https://github.com/pahen/madge

Other ecosystems have similar tooling.

Are these examples enough to warrant accepting this PR?

hdgarrood:
Ok, yeah, that's a good point. I was on the fence about the idea of compiling with an entry point; I think I'd prefer to address the underlying issue there, which if I understand correctly is that we want builds to be faster, by coming up with a way of only needing any given dependency at a given version to be built once and then stored somewhere where it's accessible across different projects. Being able to make tooling like the examples you gave would indeed be nice.

natefaubion:
I personally would support reopening it. I would probably like a more general analysis of modules though, but it's not necessary for this PR. I would like to see:

* All exports (and where they might be reexported from)
* All imports (and if there are open imports, which are the ones that are actually used)
* Foreign imports

f-f:
As @hdgarrood noted in https://github.com/spacchetti/spago/issues/165#issuecomment-496050677, that issue would be another use case for having this command (and would also avoid having a dependency on the compiler or part of it)

natefaubion:
This is still marked as a draft. We haven't looked at this in a while. Is there anything else to do with this?

jmackie:
If there's agreement that this is something we want then I'll polish it up into a proper PR?

natefaubion:
Yes, I think this would be a useful command to have.

f-f:
I would _really really_ like to see this command in.

The reason is that having this command in the compiler would allow us to remove this [bunch of code to parse module headers](https://github.com/spacchetti/spago/blob/master/src/Spago/Build/Parser.hs) we have in Spago - because you know, it's not good for several reasons: different versions of the compiler parse different things, we should not parse files again if the compiler is going to do it, etc.

So having the module graph info exposed by the compiler would allow this and other usecases to have a clean implementation in downstream tools.

f-f:
@jmackie is there anything I could help with to move this forward?

jmackie:
@f-f lack of time is the blocker :see_no_evil: 

The code is pretty straightforward but I haven't built it since the original commit. I also haven't looked at this codebase for a while. But if it still builds maybe it's good to go? I'll try and take a look during lunch today.

f-f:
@natefaubion I'm available to take this PR to a mergeable state, if anything needs to change. Could I ask you to have a review at it and see if we need changes?

kritzcreek:
@f-f I'm also up for reviewing it. I took a cursory glance and it seems the main thing missing is a few tests. I'll probably also have a couple of nits on code style, but I can fix those myself unless you'd like to.

f-f:
@kritzcreek I opened #3781

jmackie:
Thanks for taking this on @f-f :pray: sorry I couldn't be more helpful!

PR #3637, opened 2019-05-13
Improve error message on ModuleNotFound error for Prim modules.

### LINKED ISSUE #3625: ModuleNotFound error is misleading for Prim modules
This is the error you get when attempting to compile typelevel-prelude with v0.12.0:
```
Error found:
in module Type.Data.Boolean
at bower_components/purescript-typelevel-prelude/src/Type/Data/Boolean.purs line 17, column 1 - line 17, column 48

  Module Prim.Boolean was not found.
  Make sure the source file exists, and that it has been provided as an input to the compiler.
```
The error message suggestion is misleading because any module whose name starts with Prim cannot exist in a source file: the compiler reserves that namespace for modules which are built in to the compiler.

I’d suggest amending the error in the case where the module name is question is a Prim module name to instead say that “module names beginning with Prim are reserved for modules which are embedded inside the compiler, but this version of the compiler does not provide $modulename. Perhaps you need to update your compiler?”

ealmansi:
Fixes #3625.

Provides a more accurate error message when a ModuleNotFound error is raised on a built-in module.

Example output:
```
    Module Foobar was not found.
    Make sure the source file exists, and that it has been provided as an input to the compiler.

    Module Prim.Foobar was not found.
    Module names in the Prim namespace are reserved for built-in modules, but this version of the compiler does not provide module Prim.Foobar. Please, consider upgrading your compiler to a newer version.
```

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,6 +41,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@DavidLindbom](https://github.com/DavidLindbom) | David Lindbom | [MIT license](http://opensource.org/licenses/MIT) |
 | [@dckc](https://github.com/dckc) | Dan Connolly | [MIT license](http://opensource.org/licenses/MIT) |
 | [@kleeneplus](https://github.com/dgendill) | Dominick Gendill | [MIT license](http://opensource.org/licenses/MIT) |
+| [@ealmansi](https://github.com/ealmansi) | Emilio Almansi | MIT license |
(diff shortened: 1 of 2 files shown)


hdgarrood:
Thanks! Could you please add yourself to CONTRIBUTORS.md? Also I think it would be a little clearer if the second sentence said "You may be able to fix this by updating your compiler to a newer version"; that way we indicate that it's a likely but not guaranteed fix.

ealmansi:
Updated PR with requested changes.

PR #3639, opened 2019-05-14
Configure the 'shell-file' option in stack.yaml.

matthew-hilty:
This accomplishes multiple purposes:
* It pins nixpkgs and ghc to specific versions, making builds faster and more predictable.
* It invokes 'buildStackProject', which sets the LANG env variable to 'en_US.UTF-8',
  thereby preventing an odd error caused by the conjunction of NixOS and Happy,
  while allowing stack to otherwise instantiate a pure env-less shell.
* It makes command-line stack use on NixOS easier, since using the --no-nix-pure
  flag and setting the NIX_PATH env variable are made unnecessary.
* It adds nix dependencies as nix gc roots, thereby hastening builds.

### DIFF
--- a/nix/init.nix
+++ b/nix/init.nix
@@ -1,0 +1,38 @@
+{ ghc }:
+let
+  bootstrap = import <nixpkgs> { };
+
+  nixpkgsConfig = builtins.fromJSON (builtins.readFile ./nixpkgs.json);
+
+  src = bootstrap.fetchFromGitHub {
+    owner = "NixOS";
+    repo  = "nixpkgs";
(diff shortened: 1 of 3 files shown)


hdgarrood:
I think since this will inevitably go out of date and since none of the maintainers use nix, it might be better to just have these sorts of changes in your repo locally.

matthew-hilty:
Ok. I thought having a template for other nix users might be helpful, since they're fairly prevalent in the greater-haskell world. But my main reason for this pull request was to fix the Happy issue, and #3640 already does that, so I guess I'll go ahead and close this. Thanks for giving it a look.

PR #3640, opened 2019-05-14
Keep Parser.y ASCII to avoid happy locale issues

jmackie:
Small change to fix some weird locale issues I've ran into with NixOS. 

Despite `locale` reporting a utf-8 locale, when I `stack build` that locale is lost and happy crashes with an encoding error. Probably something to do with how stack manages Nix, but in any case this seems like a straightforward way of avoiding similar issues in future.

### DIFF
--- a/src/Language/PureScript/CST/Parser.y
+++ b/src/Language/PureScript/CST/Parser.y
@@ -58,7 +58,7 @@ import Language.PureScript.PSString (PSString)
   '\;'            { SourceToken _ TokLayoutSep }
   '<-'            { SourceToken _ (TokLeftArrow _) }
   '->'            { SourceToken _ (TokRightArrow _) }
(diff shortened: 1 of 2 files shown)


matthew-hilty:
Pull request #3639 also addresses this issue.

matthew-hilty:
A version of stack (version 1.10.0) specifies the LANG env variable. However, it's not even released, let alone available in nixpkgs. However, a haskell utility function "buildStackProject" in nixpkgs already exists that likewise sets the LANG env variable. Invoking it via a nix flag in stack.yaml seems to do the trick.

Perhaps ASCII would be better in general -- since it's a simpler, more constrained standard, but keeping Parser.y ASCII to work around the strange behavior of "readFile" by Happy may not be necessary.

natefaubion:
I'm fine with merging this, since it unfortunately seems to be the most robust way to deal with this way until a Stack version is released to deal with this issue.

PR #3641, opened 2019-05-14
Docs: make markdown format behave like html. Remove --docgen opt.

### LINKED ISSUE #3490: Remove --docgen option from purs docs
From `purs docs --help`:

```
  --docgen ARG             A list of module names which should appear in the
                           output. This can optionally include file paths to
                           write individual modules to, by separating with a
                           colon ':'. For example, Prelude:docs/Prelude.md. This
                           option may be specified multiple times.

Examples:
  print documentation for Data.List to stdout:
    purs docs "src/**/*.purs" ".psc-package/*/*/*/src/**/*.purs" \
      --docgen Data.List

  write documentation for Data.List to docs/Data.List.md:
    purs docs "src/**/*.purs" ".psc-package/*/*/*/src/**/*.purs" \
      --docgen Data.List:docs/Data.List.md

  write documentation for Data.List to docs/Data.List.md, and
  documentation for Data.List.Lazy to docs/Data.List.Lazy.md:
    purs docs "src/**/*.purs" ".psc-package/*/*/*/src/**/*.purs" \
      --docgen Data.List:docs/Data.List.md \
      --docgen Data.List.Lazy:docs/Data.List.Lazy.md
```

I think the `--docgen` option in `purs docs` was originally introduced in order to allow only producing docs for the modules you're interested in (usually just the modules of the package you're working on). However, I've come to feel that it offers unnecessarily flexibility, to the extent that it is awkward and difficult to use correctly. For instance, if you mix the print-to-stdout and write-to-file options, you get an error:
```
> purs docs $FILES --docgen Data.List --docgen Data.List.Lazy:docs/Data.List.Lazy.md
purs docs: error in --docgen option:
  Can't mix module names and module name/file path pairs in the same invocation.
```
If my memory serves correctly I've even seen this error from `pulp` before (although looking at the code, I'm not really sure how that could have happened).

Additionally, the commands can get sufficiently long that e.g. Windows refuses to execute them.

Furthermore, the html output format doesn't offer this flexibility, and nobody has complained; this leads me to suspect that few people are making use of this extra flexibility offered by markdown. Currently, `purs docs --format html` always writes html documentation for every source file passed in the globs into the `generated-docs` directory in the current directory (which it creates if it is not already there). Note that this must include dependencies, as `purs docs` won't work if you don't include your dependencies in the list of file globs.

I'd like to remove the markdown-specific `--docgen` option and have the markdown format behave like the html format does currently; people can always rearrange things after they've been written to disk if they would prefer a different layout.

ealmansi:
Fixes #3490.

Removes the markdown-specific `--docgen` option from `purs docs`, and makes the markdown format behave like the html format does currently; i.e., docs are generated for all modules in the given input files, and saved in the directory `generated-docs`. The directory is created first, if it does not yet exist.

Example:
```
→ purs docs example.purs
→ ls generated-docs/
Example.md  Prim.Boolean.md  Prim.md  Prim.Ordering.md  Prim.RowList.md  Prim.Row.md  Prim.Symbol.md  Prim.TypeError.md
```

### DIFF
--- a/app/Command/Docs.hs
+++ b/app/Command/Docs.hs
@@ -1,30 +1,21 @@
-{-# LANGUAGE TupleSections #-}
 
 module Command.Docs (command, infoModList) where
 
 import           Command.Docs.Html
+import           Command.Docs.Markdown
 import           Control.Applicative
-import           Control.Arrow (first, second)
-import           Control.Category ((>>>))
 import           Control.Monad.Writer
 import           Control.Monad.Trans.Except (runExceptT)
-import qualified Data.Text as T
-import qualified Data.Text.IO as T
-import           Data.Function (on)
-import           Data.List
-import           Data.Tuple (swap)
+import           Data.Functor ((<&>))
 import qualified Language.PureScript as P
 import qualified Language.PureScript.Docs as D
-import qualified Language.PureScript.Docs.AsMarkdown as D
 import           Language.PureScript.Docs.Tags (dumpCtags, dumpEtags)
 import qualified Options.Applicative as Opts
 import qualified Text.PrettyPrint.ANSI.Leijen as PP
-import           System.Directory (createDirectoryIfMissing)
+import           System.Directory (createDirectoryIfMissing, removeFile)
 import           System.Exit (exitFailure)
-import           System.FilePath (takeDirectory)
-import           System.FilePath.Glob (glob)
+import           System.FilePath.Glob (compile, glob, globDir)
 import           System.IO (hPutStrLn, stderr)
-import           System.IO.UTF8 (writeUTF8FileT)
 
 -- | Available output formats
 data Format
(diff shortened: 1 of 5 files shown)


hdgarrood:
Looks great, thanks! I think it might be better to have separate subdirectories of `generated-docs` for markdown and html.

ealmansi:
> Looks great, thanks! I think it might be better to have separate subdirectories of `generated-docs` for markdown and html.

Which names would you suggest? A few options I can think of:
- A: `generated-docs` (for html), `generated-docs-md`
- B: `generated-docs-html`, `generated-docs-md`
- C: `generated-docs/html`, `generated-docs/md`

IMO, B and C seem cleaner than A, but A doesn't break apps currently using the html format.

hdgarrood:
Definitely C; we're still pre-1.0 and I'm not aware of anyone actually relying on the current structure, in the sense that I don't think anyone is doing anything any more complicated than just opening `generated-docs/index.html` in a browser right now.

ealmansi:
Updated PR with requested changes.

Example:
```
→ purs docs example.purs
→ purs docs --format html example.purs

→ ls generated-docs/
> html  md

→ ls generated-docs/md/
> Example.md  Prim.Boolean.md  Prim.md  Prim.Ordering.md  Prim.RowList.md  Prim.Row.md  Prim.Symbol.md  Prim.TypeError.md

→ ls generated-docs/html/
> Example.html  index.html  Prim.Boolean.html  Prim.html  Prim.Ordering.html  Prim.Row.html  Prim.RowList.html  Prim.Symbol.html  Prim.TypeError.html
```

PR #3644, opened 2019-05-18
Docs: Write ctags and etags to filesystem instead of stdout.

### LINKED ISSUE #3642: Write to a tags file instead of stdout when generating tags
The current behaviour that e.g. `purs docs --format ctags` writes its output data to standard output doesn't work fantastically well together with tools which might (sensibly) assume that `purs docs` will always write to the file system. Put differently: it's weird that the `--format` option controls not only the format of output, but also how output is produced; supplying `html` or `markdown` means you get files written to disk, but supplying `etags` or `ctags` means you get data written to standard output. See also https://github.com/purescript-contrib/pulp/issues/368

ealmansi:
Fixes #3642.

When running `purs docs` with ctags or etags formats, write to the filesystem instead of writing to standard output.

CTags format will generate a file `generated-docs/tags`, while ETags format will generate a file `generated-docs/TAGS`. The naming convention is taken from the default filenames generated by the `ctags` and `etags` commands (see http://ctags.sourceforge.net/ctags.html).

Example output:
```
→ purs docs --format ctags example.purs 
→ purs docs --format etags example.purs

→ ls generated-docs/
> tags  TAGS

→ cat generated-docs/tags
> Bar	/home/ealmansi/dev/ealmansi/purescript/example.purs	6
> Foo	/home/ealmansi/dev/ealmansi/purescript/example.purs	5

```

### DIFF
--- a/app/Command/Docs.hs
+++ b/app/Command/Docs.hs
@@ -6,6 +6,7 @@ import           Command.Docs.Markdown
 import           Control.Applicative
 import           Control.Monad.Writer
 import           Control.Monad.Trans.Except (runExceptT)
+import qualified Data.Text as T
 import qualified Language.PureScript as P
 import qualified Language.PureScript.Docs as D
 import           Language.PureScript.Docs.Tags (dumpCtags, dumpEtags)
@@ -15,6 +16,7 @@ import           System.Directory (createDirectoryIfMissing, removeFile)
 import           System.Exit (exitFailure)
 import           System.FilePath.Glob (compile, glob, globDir1)
 import           System.IO (hPutStrLn, stderr)
+import           System.IO.UTF8 (writeUTF8FileT)
 
 -- | Available output formats
 data Format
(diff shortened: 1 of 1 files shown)


kRITZCREEK:
How does this interact with case-insensitive file systems? (Like OSX and Windows by default)

ealmansi:
@kRITZCREEK Good question. I reckon there would be a name clash and, if generating both ctags and etags, the second command would likely overwrite the output of the first one.

Given the following information from http://ctags.sourceforge.net/ctags.html:
- Vi will, by default, expect a tag file by the name "tags" in the current directory. (...)
- Emacs will, by default, expect a tag file by the name "TAGS" in the current directory. (...)

There's a couple options I can think of:
- A. Keep the default naming scheme. Tags will be loaded by default for most users, the exception being those that use *both* Vi and Emacs on a case-insensitive filesystem. A next step would be to make the tags filename configurable in `purs docs`, to aid those users in avoiding file conflict. Possibly also add a warning if overwriting the tags file.
- B. Rename the etags file from "TAGS" to "ETAGS". That would prevent the name clash, but may no longer work seamlessly for Emacs users. I found one [source](https://idempot.net/git/effortless_git_tags/) mentioning using this filename instead.

hdgarrood:
If we want tags to be loaded by default, wouldn't we need to write the tags file to project's root directory (which will generally be the current directory when this is running) as opposed to in `generated-docs`?

ealmansi:
@hdgarrood Indeed. Just updated the PR to generate tags files in the current directory.

Updated output example:
```
→ purs docs --format ctags example.purs 
→ purs docs --format etags example.purs 
→ ls tags TAGS
tags  TAGS
```

hdgarrood:
Writing the files to names which would conflict on case-insensitive file systems is fine by me since someone using both etags and ctags together does seem like a bit of an edge case. However I would like to check and make sure it doesn't do anything really strange if you do try to produce both on a case-insensitive file system (@ealmansi do you have access to one? If not I can probably find a windows machine to give it a try on). I think we ought to add an `--output` option as well because of the potential for clashes, which should be interpreted as a directory in the case of html or markdown, and as a file in the case of ctags or etags. Also, I'd like to start emitting a message saying what was written. Do you think we could reduce the duplication in this module a bit now too?

ealmansi:
@hdgarrood I've added a commit reducing duplication in the code introduced by this PR.

Unfortunately, I don't have access to a Windows or macOS machine at the moment.

+1 on: i) adding informational output when docs are generated successfully, ii) adding an argument to make the output location configurable. IMO it'd be best if each of these were implemented as independent PRs.

app/Command/Docs.hs
```diff
@@ -40,8 +43,12 @@ docgen (PSCDocsOptions fmt inputGlob) = do
   fileMs <- parseAndConvert input
   let ms = D.primModules ++ map snd fileMs
   case fmt of
-    Etags -> mapM_ putStrLn $ dumpEtags fileMs
-    Ctags -> mapM_ putStrLn $ dumpCtags fileMs
+    Etags -> do
+      let outputFilename = "TAGS" -- TODO: make this configurable
+      writeTagsToFile outputFilename $ dumpEtags fileMs
+    Ctags -> do
+      let outputFilename = "tags" -- TODO: make this configurable
```
hdgarrood: I don’t think we need a variable binding for this now, likewise in the etags branch.

app/Command/Docs.hs
```diff
@@ -72,6 +79,15 @@ docgen (PSCDocsOptions fmt inputGlob) = do
                 >>= uncurry D.convertTaggedModulesInPackage)
     >>= successOrExit
 
+  writeTagsToFile :: String -> [String] -> IO ()
+  writeTagsToFile outputFilename tags = do
+    currentDir <- getCurrentDirectory
+    let outputFile = currentDir </> outputFilename
+    let pattern = compile outputFilename
+    let text = T.pack . unlines $ tags
+    globDir1 pattern currentDir >>= mapM_ removeFile
```
hdgarrood: I don’t think we need to do any globbing here, won’t the file just get overwritten if it already exists?

hdgarrood:
FYI I’ve tried this out on a windows machine and it behaved as expected (overwriting the file).

PR #3645, opened 2019-05-18
Use externs files when producing docs

### LINKED ISSUE #3503: Require existence of externs files when producing docs

### LINKED ISSUE #3624: `purs docs` fails due to thinking a Newtype instance is using a type synonym
Apologies if this is a duplicate.

Originally reported JordanMartinez/learn-halogen#5 and later in slamdata/purescript-halogen#601, running `spago docs` on the `purescript-halogen`'s `v5.0.0-rc.4` release will fail with the below error. [In his comment in the Halogen issue](https://github.com/slamdata/purescript-halogen/issues/601#issuecomment-489799441), Gary said this is likely a bug in `purs docs`.

```
Generating documentation for the project. This might take a while..
Error found:
in module Halogen.Query.EventSource
at .spago/halogen/v5.0.0-rc.4/src/Halogen/Query/EventSource.purs:32:8 - 32:66 (line 32, column 8 - line 32, column 66)

  Type class instances for type synonyms are disallowed.

in type class instance
                                                         
  Data.Newtype.Newtype (EventSource m a)                 
                       (m                                
                          { finalizer :: Finalizer m     
                          , producer :: Producer a m Unit
                          }                              
                       )                                 
                                                         

See https://github.com/purescript/documentation/blob/master/errors/TypeSynonymInstance.md for more information,


spago: Docs generation failed.
```

hdgarrood:
Resolves #3503, fixes #3624.

This commit does a few related things:

**Docs codegen target**

Adds a `docs` codegen target for `purs compile`. When supplied, a
`docs.json` file is written for each module alongside the other files in
that module's directory within the output directory. The file contains a
JSON-serialized Docs.Types.Module, which is the same as the one which
will be produced by `purs publish` for that module, except that the
re-exports field will be empty; this is because computing re-exports
requires access to documentation for all of the upstream modules, and we
don't have an easy way of getting hold of all of these (which won't
adversely affect memory consumption / performance for the regular
compilation path).

Since we are producing docs during `purs compile`, we now have easy
access to the relevant externs files, which we can use to fill in type
signatures of any declarations which don't already have their own. This
means that no module within the `Language.PureScript.Docs` namespace
needs to call into the type checker any more in order to get accurate
and useful type signatures in docs, regardless of whether annotations
have been given or not, which both speeds up docs generation
significantly and eradicates the possibility of bugs like #3624 arising.

**Use docs.json files in purs docs and purs publish**

The subcommands `purs docs` and `purs publish` now each accept an
`--output` option, which should point to a compiler output directory,
whose default value is `"output"`. With this information, we can reach
into the compiler output directory to grab the information in each of
the required `docs.json` files, perform one final pass over them to fill
in re-exports, and then produce documentation in whatever format is
requested. If the `docs.json` files are not up-to-date with respect to
the source files which have been provided to `purs docs`, we throw an
error.

In effect, `purs docs` and `purs publish` now behave more like
`purs bundle` does, in the sense of getting their input data from the
output of `purs compile`, rather than from a collection of source files.

**Refactor Docs library API**

The Docs library API has been refactored and simplified; now, the `purs
docs` CLI and the `Language.PureScript.Publish` module only really need
to use one function for getting hold of docs: the function `collectDocs`
from the module `Language.PureScript.Docs.Collect`. This function
implements the behaviour described above for reading `docs.json` files,
making sure they are up-to-date, and filling in re-exports.

With this new `collectDocs` function, a number of other functions no
longer need to be exposed as part of the compiler library API. In
particular, the module `Language.PureScript.Docs.ParseInPackage` has
been removed entirely.


src/Language/PureScript/Errors.hs
```diff
@@ -1060,6 +1061,12 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
             , indent $ paras $ map (line . markCode . showQualified showOp) (NEL.toList ops)
             , line "Use parentheses to resolve this ambiguity."
             ]
+    renderSimpleErrorMessage NeedToBuildDocs =
+      paras
+        [ line "Some of the input modules are missing up-to-date documentation data in the compiler output directory. Please compile all of the relevant modules using `purs compile' with the `docs' codegen target."
+        , line "For example:"
+        , indent $ line "$ purs compile --codegen js,docs SOURCE_FILES"
```
natefaubion: Is the `js` option necessary? I don't think it is, and I'm sure the example is meant to indicate "hey, you can compile docs with normal JS anytime you want", but it makes it seem likes it's necessary.

hdgarrood: You're right, it isn't necessary. Do you think we should just write `--codegen docs` here then? I don't have any strong feelings either way.

src/Language/PureScript/ModuleDependencies.hs
```diff
@@ -50,8 +50,7 @@ sortModules toSig ms = do
     return (fst <$> ms'', moduleGraph)
   where
     toGraphNode :: S.Set ModuleName -> (a, ModuleSignature) -> m ((a, ModuleSignature), ModuleName, [ModuleName])
-    toGraphNode mns m@(_, ModuleSignature _ mn ds) = do
-      let deps = ordNub (mapMaybe usedModules ds)
```
natefaubion: What's the reason for removing the ordNub? Was it unnecessary?

hdgarrood: Oops, no, that was an oversight. I'll put it back.

src/Language/PureScript/Errors.hs
natefaubion: I don't think it's necessary. The help for `purs compile` indicates that you can provide multiple codegen targets.

src/Language/PureScript/ModuleDependencies.hs
natefaubion: I mean, I'm genuinely unsure what it's doing, so I was hoping you know 😆.

hdgarrood: Actually, maybe it would make sense to use a `Set` for the `sigImports` field?

hdgarrood: I think it is actually necessary if we want the module graph to not contain duplicate entries (which indeed seems like a sensible thing to want).

natefaubion: Does the specific order matter though. `Set` wouldn't preserve that.

hdgarrood: That's true. I'm not sure. Perhaps best to put the `ordNub` back to leave it as it was then.

hdgarrood:
Actually, perhaps the CLI flag should be `--compile-output`, so as not to clash with an `--output` flag which might be used to specify the file or directory you'd like the result to be written to.

natefaubion:
I think `--compile-output` makes sense.

hdgarrood:
Sorry for changing this at the last minute, but I think it makes the CLI a lot nicer. Instead of erroring if the docs are not up-to-date, we just build them. This makes this a non-breaking change. Now that we are actually building, might want to consider adding a `--json-errors` flag to `purs docs` too, but that can happen later.

hdgarrood:
Thank you both!

PR #3646, opened 2019-05-20
Remove old parser

natefaubion:
This PR updates all usages of the old parser to use the new CST parser and removes the old parser.

The trickiest part is probably the purs-repl parser. I added a few parser combinators so we that we could assemble an ad-hoc parser from different productions. It handles multiple repl statements by lexing as if in a top-declaration context.

I've also adjusted purs-ide import handling to only insert newlines around the import section if necessary, so that you don't get double blank lines in some cases.


src/Language/PureScript/Interactive/Parser.hs
```diff
@@ -79,38 +103,44 @@ parseDirective cmd =
...
-psciImport = do
-  (mn, declType, asQ) <- P.parseImportDeclaration'
-  return $ Import (mn, declType, asQ)
+psciImport :: FilePath -> CST.Parser Command
+psciImport filePath = do
+  decl <- CST.convertImportDecl filePath <$> CST.parseImportDeclP
+  case decl of
+    P.ImportDeclaration _ mn declType asQ ->
+      pure $ Import (mn, declType, asQ)
+    _ -> error "Not an import"
```
hdgarrood: should we use `internalError` here?

natefaubion:
> Looks good. I think longer-term I'd like to try to replace the psci parser with a proper happy-based one but we don't need to do that right away. (Do you think that would make sense?)

I'm not sure. You would have to include it in `Parser.y` to reuse the existing terminals, and discriminating declaration vs expression syntax would require backtracking (or unbounded lookahead) anyway. I honestly think the combinators for somewhat adhoc stuff like this aren't too bad.

src/Language/PureScript/Interactive/Parser.hs
natefaubion: Probably. It would be very very bad and everything would break if the error case was possible. The only reason it's there is because there isn't a `convert` function for just the import tuple. I could also probably just add that instead and remove the error.

natefaubion:
The way it handles the error reporting is picking the one that produced and error that's farthest along in the input stream.

hdgarrood:
Ok cool, that sounds good then. :+1:

PR #3647, opened 2019-05-26
Add --output option for purs docs

hdgarrood:
To control where output gets written to. Also say where output was
written. Raised as part of review feedback in #3644

I've moved the option parsers under the `pscDocsOptions` declaration to avoid name shadowing.

cc @ealmansi


hdgarrood:
Thanks for the review!

Issue #3502, opened 2019-01-06
Modules are permitted to re-export a type class and a type of the same name


hdgarrood:
Currently, it's possible to re-export a type and a type class of the same name from a module, even though it's not possible to define a type and a type class of the same name together in one module. For example,

```purescript
-- Eg.Type.purs
module Eg.Type where

data X = X

-- Eg.Class.purs
module Eg.Class where

class X a

-- Eg.purs
module Eg (module Type, module Class) where

import Eg.Type (X)
import Eg.Class (class X)
```

There is an example of this in `purescript-typelevel-prelude` at version 3.0.0, where the module `Type.Row` re-exports the class `Cons` from `Prim.Row` as well as the type `Cons` from `Prim.RowList`. At the moment, this is only a minor issue; as far as I'm aware, the only problem this can cause is that it can break anchor links in docs. However, if we ever want to do #1888 (which do still I think we should), I think we will need to make this an error.

hdgarrood:
Presumably then the plan is to fix this by disallowing it? That will be a breaking change, so shall we label it `breaking`?

garyb:
Sure thing

hdgarrood:
This was fixed by #3648.

PR #3648, opened 2019-05-26
Disallow re-exporting class and type with the same name

joneshf:
Re: #3502.

This PR follows the suggestion of disallowing the re-export. Let me know if I missed anything or if anything needs to change!

The tests are currently failing because of https://github.com/purescript/purescript-typelevel-prelude/issues/43.


hdgarrood:
@joneshf now that typelevel-prelude 4.0.1 has been released could you please try updating `tests/support/bower.json` to use that, and hopefully that will get CI passing?

joneshf:
Looks like that works!

hdgarrood:
Unfortunately it looks like we will have to make a breaking release of typelevel-prelude after all, see https://github.com/purescript/purescript-typelevel-prelude/issues/46.

hdgarrood:
Actually I'll just merge this for now, and we can update the test dependencies later. Thanks @joneshf!

Issue #3654, opened 2019-05-30
PureScript 0.13.0 doesn't build with cabal


vmchale:
Running `cabal new-install purescript` yields various errors of the sort

```
      in the definition of data constructor ‘HappyWrap180’
     |
1130 | newtype HappyWrap180 = HappyWrap180 ([(SourceToken, _)])
     |                                                     ^

dist/build/Language/PureScript/CST/Parser.hs:1137:53: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap181’
     |
1137 | newtype HappyWrap181 = HappyWrap181 ([(SourceToken, _)])
     |                                                     ^

dist/build/Language/PureScript/CST/Parser.hs:1144:50: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap182’
     |
1144 | newtype HappyWrap182 = HappyWrap182 (NE.NonEmpty _)
     |                                                  ^

dist/build/Language/PureScript/CST/Parser.hs:1151:50: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap183’
     |
1151 | newtype HappyWrap183 = HappyWrap183 (NE.NonEmpty _)
     |                                                  ^

dist/build/Language/PureScript/CST/Parser.hs:1158:50: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap184’
     |
1158 | newtype HappyWrap184 = HappyWrap184 (NE.NonEmpty _)
     |                                                  ^

dist/build/Language/PureScript/CST/Parser.hs:1165:53: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap185’
     |
1165 | newtype HappyWrap185 = HappyWrap185 ([(SourceToken, _)])
     |                                                     ^

dist/build/Language/PureScript/CST/Parser.hs:1172:53: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap186’
     |
1172 | newtype HappyWrap186 = HappyWrap186 ([(SourceToken, _)])
     |                                                     ^

dist/build/Language/PureScript/CST/Parser.hs:1179:53: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap187’
     |
1179 | newtype HappyWrap187 = HappyWrap187 ([(SourceToken, _)])
     |                                                     ^
```

[...]

PR #3661, opened 2019-06-01
Add upper bounds when producing source distributions

hdgarrood:
This change means that stack will add upper bounds to our cabal file
when producing source distributions (e.g. for uploading to Hackage),
which means that we are less likely to run into solving/building
difficulties if new versions of our dependencies get released in the
meantime.

Comparing a `build-depends` section of the purescript.cabal file in a
source distribution before and after this change:

Before
```
      Cabal >=2.2
    , Glob >=0.9 && <0.10
    , aeson >=1.0 && <1.5
    , aeson-better-errors >=0.8
    , ansi-terminal >=0.7.1 && <0.9
    , array
    , base >=4.11 && <4.13
    , base-compat >=0.6.0
    , blaze-html >=0.8.1 && <0.10
    , bower-json >=1.0.0.1 && <1.1
    , boxes >=0.1.4 && <0.2.0
    , bytestring
    , cheapskate >=0.1 && <0.2
    , clock
    , containers
    , data-ordlist >=0.4.7.0
    , deepseq
    , directory >=1.2.3
    , dlist

```

After
```
        Cabal >=2.2 && <2.5,
        Glob ==0.9.*,
        aeson >=1.0 && <1.5,
        aeson-better-errors >=0.8 && <0.10,
        ansi-terminal >=0.7.1 && <0.9,
        array <0.6,
        base >=4.11 && <4.13,
        base-compat >=0.6.0 && <0.11,
        blaze-html >=0.8.1 && <0.10,
        bower-json >=1.0.0.1 && <1.1,
        boxes >=0.1.4 && <0.2.0,
        bytestring <0.11,
        cheapskate ==0.1.*,
        clock <0.8,
        containers <0.7,
        data-ordlist >=0.4.7.0 && <0.5,
        deepseq <1.5,
        directory >=1.2.3 && <1.4,
        dlist <0.9,
```

Note in particular that e.g. `array` (no bounds) becomes `array <0.6`,
and `Cabal >=2.2` becomes `Cabal >=2.2 && <2.5`.

The `purescript.cabal` file generated by hpack for local development
is unchanged. I think this is fine, because if our version bounds turn
out to be wrong during local development we can just fix them then and
there.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,5 @@
 resolver: lts-13.12
+pvp-bounds: upper
 packages:
 - '.'
 extra-deps:


hdgarrood:
In case you're wondering what prompted this, I stumbled across https://github.com/haskell-infra/hackage-trustees/blob/master/cookbook.md while investigating #3654, and found the arguments it makes fairly persuasive. Adding these upper bounds wouldn't have prevented #3654 from arising, however, although this is mainly because happy 1.19.10 should really have been released as 1.20., since it included breaking changes, e.g. explicit foralls are now required for polymorphic productions with type signatures.

PR #3666, opened 2019-06-03
Add --debug flag to bundle command

rhendric:
This flag causes an optimized-for-humans JSON representation of the
modules being bundled to be dumped to stderr, prior to dead code
elimination.  This feature was requested in the review of #3562.

---

*<waves at @hdgarrood>*

[Here's some sample output](https://gist.github.com/rhendric/8ace4d8280972bc5b596c5537d8b0a90), generated by running `bundle --debug` on `Data.Functor*`.

### DIFF
--- a/app/Command/Bundle.hs
+++ b/app/Command/Bundle.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE RecordWildCards #-}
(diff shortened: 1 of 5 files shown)


hdgarrood:
Looks great, thank you! On second thoughts though (sorry) the Skip and Other constructors could probably do with a little more information. Perhaps you could minify the js and use the first 20 characters of the minified string plus an ellipsis if that string is longer than 20 characters?

rhendric:
Not sure if doing JS minification is all that straightforward... would you settle for starting line/column numbers for the statements?

hdgarrood:
Oh sorry - I meant to clarify, if I remember correctly language-javascript can do this for you with one function call. If not line/column numbers sounds great.

rhendric:
Oh, you're right! Apparently I didn't search the whole package. Okay, minified JS it is.

PR #3676, opened 2019-06-13
Ignore duplicate file inputs to compile command (#3653)

### LINKED ISSUE #3653: Ignore repeated source files passed to the CLI
For example:
```purescript
-- example.purs
module Example where
```
```
$ purs compile example.purs
Compiling Example
[OK]
$ purs compile example.purs example.purs
Error found:
at example.purs:1:1 - 1:21 (line 1, column 1 - line 1, column 21)

  Module Example has been defined multiple times


See https://github.com/purescript/documentation/blob/master/errors/DuplicateModule.md for more information,
```
Ideally both commands would do the same thing.

dyerw:
Resolves #3653 

If we should do the `nub` call in the `readInput` function that'd be fine too. But this is the only place it's called. Not sure if it matters if that's the behavior of `compile` specifically or `readInput`.

I couldn't find tests for the command line interface but if there are some I'm happy to add one.

Also fixed a spelling mistake =P

### DIFF
--- a/src/System/IO/UTF8.hs
+++ b/src/System/IO/UTF8.hs
@@ -4,25 +4,22 @@ module System.IO.UTF8 where
 
 import Prelude.Compat
 
-import           Control.Monad
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as BSL
 import qualified Data.ByteString.Search as BSS
 import qualified Data.ByteString.UTF8 as UTF8
-import           Data.List.Extra (nubOrd)
 import           Data.Text (Text)
 import qualified Data.Text.Encoding as TE
+import           Protolude (ordNub)
 
 -- | Unfortunately ByteString's readFile does not convert line endings on
 -- Windows, so we have to do it ourselves
 fixCRLF :: BS.ByteString -> BS.ByteString
 fixCRLF = BSL.toStrict . BSS.replace "\r\n" ("\n" :: BS.ByteString)
 
-readUTF8FilesTUnique :: [FilePath] -> IO [(FilePath, Text)]
-readUTF8FilesTUnique =  readUTF8FilesT . nubOrd
-
 readUTF8FilesT :: [FilePath] -> IO [(FilePath, Text)]
-readUTF8FilesT inputFiles = forM inputFiles $ \inFile -> (inFile, ) <$> readUTF8FileT inFile
+readUTF8FilesT =
+  traverse (\inFile -> (inFile, ) <$> readUTF8FileT inFile) . ordNub
 
 readUTF8FileT :: FilePath -> IO Text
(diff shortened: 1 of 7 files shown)


klntsky:
Usually there are quite a lot of elements in that list. And `nub` is `O(n^2)`. I think using `Set` here would be more appropriate, as it would give `O(n*log(n))` time complexity.

hdgarrood:
Yes, we should definitely avoid `nub`, but since all of the relevant functions are expecting lists I think it would be simpler to use `nubOrd` rather than using a `Set`.

Also I think it would be preferable to put the `nubOrd` call inside `readInput` and use the same `readInput` function for other parts of the cli which accept lists of source files, such as `purs docs`, `purs repl`, and `purs hierarchy`.

dyerw:
Sounds good! I'll make the changes.

hdgarrood:
Thanks! Oh also could you please add yourself to CONTRIBUTORS.md too?

dyerw:
Can do. 

Any suggestions on what module a shared `readInput` function would live in?

dyerw:
Latest commit should address the issues, the UTF8 module seemed like a reasonable place to consolidate all the `readInput` logic since they were all using `readUTF8FileT` any way.

dyerw:
Also, if anyone knows why the windows build is failing I can't figure it out. It looks like it might be timing out? It compiles and passes tests locally (mac).

kRITZCREEK:
No need to add a dependency on extra for nubOrd. We've already got ordNub around (I think from Protolude)

kRITZCREEK:
FWIW for a bunch of files hashNub is going to perform a lot better (Ord is slow for strings with long common prefixes)

src/System/IO/UTF8.hs
```diff
@@ -14,6 +18,12 @@ import qualified Data.Text.Encoding as TE
 fixCRLF :: BS.ByteString -> BS.ByteString
 fixCRLF = BSL.toStrict . BSS.replace "\r\n" ("\n" :: BS.ByteString)
 
+readUTF8FilesTUnique :: [FilePath] -> IO [(FilePath, Text)]
+readUTF8FilesTUnique =  readUTF8FilesT . nubOrd
+
+readUTF8FilesT :: [FilePath] -> IO [(FilePath, Text)]
```
hdgarrood: Are we likely to want to use the "non-unique" version anywhere? I can't think of a place we might want to do that so I'd suggest only providing the unique version (and getting rid of the "Unique" suffix).

dyerw: Sure thing!

hdgarrood:
Protolude doesn't appear to provide a hashing nub and the so I've gone with `nubOrd` for now; we can always revisit this later.

PR #3680, opened 2019-06-15
Fix record-unification issue:

matthew-hilty:
* In type checking and subsumption, strip row types of 'ForAll' nodes, and replace type variables with unification variables so that later typeclass resolution can be accomplished.

Here are two examples for which the described changes are necessary:

```purescript
class GId (l :: RowList) (r :: # Type) | l -> r where
  gId :: RLProxy l -> Record r -> Record r

instance gIdCons :: GId (Cons s (va -> vb) l') r where
  gId l record = record

id :: forall l r. GId l r => RowToList r l => Record r -> Record r
id = gId (RLProxy :: RLProxy l)

example1 = id { a: (identity :: forall a. a -> a) }

x :: { a :: forall a. a -> a }
x = { a: (identity :: forall a. a -> a) }

example2 = id x
```

### DIFF
--- a/src/Language/PureScript/TypeChecker/Subsumption.hs
+++ b/src/Language/PureScript/TypeChecker/Subsumption.hs
@@ -97,10 +97,9 @@ subsumes' mode ty1 (KindedType _ ty2 _) =
 -- Only check subsumption for constrained types when elaborating.
 -- Otherwise fall back to unification.
 subsumes' SElaborate (ConstrainedType _ con ty1) ty2 = do
-  dicts <- getTypeClassDictionaries
-  hints <- getHints
+  dict <- getTypeClassDictionary con
   elaborate <- subsumes' SElaborate ty1 ty2
-  let addDicts val = App val (TypeClassDictionary con dicts hints)
+  let addDicts val = App val dict
   return (elaborate . addDicts)
 subsumes' mode (TypeApp _ f1 r1) (TypeApp _ f2 r2) | eqType f1 tyRecord && eqType f2 tyRecord = do
     let (common, ((ts1', r1'), (ts2', r2'))) = alignRowsWith (subsumes' SNoElaborate) r1 r2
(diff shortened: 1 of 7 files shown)


garyb:
@paf31 if you get a minute, could you give your thoughts on this? I remember quite some time ago you talked to me about the possibility of disallowing polytypes in records, and this seems to follow a similar idea although without restricting the source language.

The changes here seem sensible to me, but I don't feel like I know well enough the details of what's going on at this point of the typechecker to have an idea of what any possible unintended consequences might be.

src/Language/PureScript/TypeChecker/Unify.hs
```diff
@@ -45,6 +47,36 @@ freshType = do
   modify $ \st -> st { checkNextType = t + 1 }
   return $ srcTUnknown t
 
+-- | Remove any ForAlls and ConstrainedType constructors in a type by introducing new unknowns.
+--
+-- This is necessary during type checking to avoid unifying a polymorphic type with a
+-- unification variable.
+instantiatePolyTypeWithUnknowns :: MonadState CheckState m => SourceType -> m SourceType
```
paf31: I can't think of a case where you'd want to instantiate a constraint like this and not elaborate some value with a corresponding dictionary. That's why there was only one function before.

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -112,14 +111,22 @@ subsumes' mode (TypeApp _ f1 r1) (TypeApp _ f2 r2) | eqType f1 tyRecord && eqTyp
...
     return (defaultCoercion mode)
   where
     -- Find the first property that's in the first list (of tuples) but not in the second
     firstMissingProp t1 t2 = fst <$> uncons (minusBy' (comparing rowListLabel) t1 t2)
+    mapRowListItem :: Monad m => (Type a -> m (Type a)) -> RowListItem a -> m (RowListItem a)
+    mapRowListItem f rli = do
+      ty' <- f (rowListType rli)
+      return $ rli { rowListType = ty' }
+    removePolyTypes :: MonadState CheckState m => RowListItem SourceAnn -> m (RowListItem SourceAnn)
+    removePolyTypes = mapRowListItem instantiatePolyTypeWithUnknowns
```
paf31: This is going to instantiate constraints and not pass in the appropriate dictionaries. I'd encourage you to write down the elaboration rules you're implementing here. I don't think there's a good way to support constraints.

paf31:
It's been a while since I've thought about this code, so I'll just suggest if you do change the elaboration rules that you test them to death, because I tried similar things before and seem to remember hitting weird corner cases. The constrained case definitely seems unsound.

hdgarrood:
I think I'm going to close this for now, since we haven't yet been able to work out how to address Phil's feedback, and I think the workaround of newtyping any polymorphic types in records is tolerable for now. Thanks for the PR though!

PR #3682, opened 2019-06-16
Fix #3558

### LINKED ISSUE #3558: Dictionaries "get lost" when using a higher order function vs pattern matching a constructor

matthew-hilty:
Use the substitution state after typechecking to update the dicts in an expr before typeclass resolution.

The example by @garyb in #3558 (shown below) causes a 'NoInstanceFound' error because the argument type of `unLBox` constrains some of the function's first-rank type variables, specifically, `row` and `a`. During type checking, `checkFunctionApplication'` replaces `row` and `a` with unification variables, which are used to create the `NamedDict` for `Cons`. However, although further type checking unifies the `TUnknown`s for `row` and `a` with corresponding `Skolem`s, the instantiated dict is not updated.

```purescript
newtype LBox row a = LBox (∀ r. (∀ lbl _1. Row.Cons lbl a _1 row ⇒ IsSymbol lbl ⇒ SProxy lbl → r) → r)

unLBox ∷ ∀ row a r. (∀ lbl _1. Row.Cons lbl a _1 row ⇒ IsSymbol lbl ⇒ SProxy lbl → r) → LBox row a → r
unLBox g (LBox f) = f g

readNope ∷ ∀ row a. Record row → LBox row a → a
readNope rec = unLBox \lbl → Record.get lbl rec
```


matthew-hilty:
Investigating this issue prompted me to wonder about a couple of things:
- whether unification variables shouldn't be instantiated from `ForAll` identifiers directly, but instead from their skolemized counterparts
- whether the code base shouldn't favor skolemization over quantification more generally

The interplay of `ForAll`, `TUnknown`, and `Skolem` types can be fairly complex. Perhaps the subtleties of their relationship are contributing to problems like #3558, and if so, simplifying this relationship (for example, by precluding direct translation from `ForAll` identifiers to `TUnknown`s) might be helpful. An unstudied thought anyway.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -173,12 +174,31 @@ typesOf bindingGroupType moduleName vals = withFreshSubstitution $ do
...
+      where
+      (_, go, _) = everywhereOnValues id g id
+      g (TypeClassDictionary con ctx hints) =
+          TypeClassDictionary con (update ctx) hints
+      g expr = expr
+      getLocal = fromMaybe M.empty . M.lookup localScope
+      localScope = Nothing
+      modifyDict f dict = dict { tcdInstanceTypes = f (tcdInstanceTypes dict) }
+      modifyLocal = M.map . M.map . fmap $ modifyDict $ fmap $ substituteType subst
+      update ctx = M.insert localScope (modifyLocal . getLocal $ ctx) ctx
```
garyb: Just one thought though, could this be done with `M.alter` instead to avoid the need for `getLocal`, etc.?

matthew-hilty: Yup. I should have done a better job looking through Map's api first.

Any thoughts about the following? Could it be clearer and/or more succinct? Also, I'm not sure I have a good sense of the code base's styling conventions. Any customary linting standards I should adopt?
```purescript
    updateContextsInExpr :: Substitution -> Expr -> Expr
    updateContextsInExpr subst = go
      where
      (_, go, _) = everywhereOnValues id g id
      g (TypeClassDictionary con ctx hints) =
          TypeClassDictionary con (updateCtx ctx) hints
      g expr = expr
      updateDict f dict = dict { tcdInstanceTypes = f (tcdInstanceTypes dict) }
      updateScope = fmap . fmap . fmap . fmap $ updateDict $ fmap $ substituteType subst
      updateCtx = M.alter updateScope Nothing
```

garyb: That revision looks good to me! 

The style/conventions of the codebase are a bit all over the place. As long as there's nothing overtly weird it's all good pretty much :wink:...

matthew-hilty: Cool. Danke beaucoup. I'll submit the changes by tonight.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -173,12 +174,29 @@ typesOf bindingGroupType moduleName vals = withFreshSubstitution $ do
     constrain cs ty = foldr srcConstrainedType ty (map (\(_, _, x) -> x) cs)
 
     -- Apply the substitution that was returned from runUnify to both types and (type-annotated) values
-    tidyUp ts sub = first (map (second (first (second (overTypes (substituteType sub) *** substituteType sub))))) ts
+    tidyUp ts sub = first (map (second (first (second (updateExpr sub *** substituteType sub))))) ts
```
paf31: Should the fix not be to modify `overTypes`?

matthew-hilty: I guess I'd been a bit reluctant to modify `overTypes` since it's used the once elsewhere. But the change looks like it'd be consistent with that other use site, and it passes tests, so your suggestion would the clearer option.

Anyway, before I push new code, what I think you have in mind follows. But let me know if you have any other ideas/recommendations.

```purescript
-- Types.hs
-- DELETE `updateExpr`, `updateContextsInExpr`, and `updateTypeAnnotations`.
tidyUp ts sub = first (map (second (first (second (overTypes (substituteType sub) *** substituteType sub))))) ts
```
```purescript
-- Traversals.purs
overTypes :: (SourceType -> SourceType) -> Expr -> Expr
overTypes f = let (_, f', _) = everywhereOnValues id g id in f'
  where
  g :: Expr -> Expr
  g (TypedValue checkTy val t) = TypedValue checkTy val (f t)
  g (TypeClassDictionary c sco hints) =
    TypeClassDictionary
      (mapConstraintArgs (fmap f) c)
      (updateCtx sco)
      hints
  g other = other
  updateDict f dict = dict { tcdInstanceTypes = f (tcdInstanceTypes dict) }
  updateScope = fmap . fmap . fmap . fmap $ updateDict $ fmap f
  updateCtx = M.alter updateScope Nothing
```

(By the way, I apologize for my delayed response. I hadn't even noticed your comment until I just happened by this pull request yesterday. Also, thanks; I appreciate your input.)

MonoidMusician:
I got a chance to test this today, it works beautifully! Just what I've always wanted 🎁 💝

hdgarrood:
I still haven't gotten around to understanding the type checker to the extent that I can help with this PR, but shall we go ahead and merge it? It seems like it should be pretty easy to revert if need be.

garyb:
Sounds good to me!

garyb:
Thanks @matthew-hilty

hdgarrood:
Oh, can you please add yourself to CONTRIBUTORS.md if you’re not in there already?

hdgarrood:
Yeah, there is a policy, but also I think this contribution definitely deserves to be recognised! I think a new pull request would be easiest.

Issue #3507, opened 2019-01-12
Doc-comments for typeclass methods are not included in the generated docs


Dretch:
For example, given this module:

```purescript
module MethodDocs where

import Prelude

-- | Adds drama to the given text.
dramaFunction :: String -> String
dramaFunction s = s <> "!"

class Drama t where
  -- | Adds drama to the given value.
  dramaMethod :: t -> t
```

The following docs are generated - with docs for the function, but not the method:

#### `dramaFunction`

``` purescript
dramaFunction :: String -> String
```

Adds drama to the given text.

#### `Drama`

``` purescript
class Drama t  where
  dramaMethod :: t -> t
```

It is possible to workaround this issue to some degree by including the method docs in the docs for the class itself. However, this is a bit annoying when refactoring functions to make them methods, and starts to become impracticable with more than a couple of methods per class.

As precedence, Haskell has this feature in its docs generation, e.g. http://hackage.haskell.org/package/base-4.12.0.0/docs/Text-Show.html

hdgarrood:
Yes, I'd like to do this too. I'd also like to do it for data constructors and potentially even record fields in type synonyms. For example, declarations like [React.Basic.ComponentSpec](https://pursuit.purescript.org/packages/purescript-react-basic/6.2.0/docs/React.Basic#t:ComponentSpec) would benefit a lot from this.

I think we'd need to do https://github.com/purescript/purescript/issues/2216 first.

JordanMartinez:
@hdgarrood I believe the `0.13.3` release closes this issue, correct?

hdgarrood:
Yes, thanks!

PR #3683, opened 2019-06-17
Add data constructor docs

marcosh:
Keep track of data constructor comments and print them in the docs page

should solve also #3507 

please let me know if I could improve it in any way

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -483,12 +483,12 @@ pattern ValueDecl sann ident name binders expr
 
 data DataConstructorDeclaration = DataConstructorDeclaration
   { dataCtorAnn :: !SourceAnn
-  , dataCtorProperName :: !(ProperName 'ConstructorName)
-  , dataCtorVars :: ![(Ident, SourceType)]
+  , dataCtorName :: !(ProperName 'ConstructorName)
+  , dataCtorFields :: ![(Ident, SourceType)]
   } deriving (Show, Eq)
 
 traverseDataCtorVars :: Monad m => ([(Ident, SourceType)] -> m [(Ident, SourceType)]) -> DataConstructorDeclaration -> m DataConstructorDeclaration
-traverseDataCtorVars f DataConstructorDeclaration{..} = DataConstructorDeclaration dataCtorAnn dataCtorProperName <$> f dataCtorVars
+traverseDataCtorVars f DataConstructorDeclaration{..} = DataConstructorDeclaration dataCtorAnn dataCtorName <$> f dataCtorFields
 
 -- |
 -- The data type of declarations
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -487,8 +487,8 @@ data DataConstructorDeclaration = DataConstructorDeclaration
   , dataCtorFields :: ![(Ident, SourceType)]
   } deriving (Show, Eq)
 
-traverseDataCtorVars :: Monad m => ([(Ident, SourceType)] -> m [(Ident, SourceType)]) -> DataConstructorDeclaration -> m DataConstructorDeclaration
-traverseDataCtorVars f DataConstructorDeclaration{..} = DataConstructorDeclaration dataCtorAnn dataCtorName <$> f dataCtorFields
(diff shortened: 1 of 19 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -481,14 +481,23 @@ pattern ValueDecl :: SourceAnn -> Ident -> NameKind -> [Binder] -> [GuardedExpr]
 pattern ValueDecl sann ident name binders expr
   = ValueDeclaration (ValueDeclarationData sann ident name binders expr)
 
+data DataConstructorDeclaration = DataConstructorDeclaration
+  { dataCtorAnn :: !SourceAnn
+  , dataCtorProperName :: !(ProperName 'ConstructorName)
+  , dataCtorVars :: ![(Ident, SourceType)]
```
hdgarrood: Did you call these by these names so as not to clash with the names of the same fields in the CST? I think it would make more sense to use the same names as in CST, i.e. `dataCtorName` and `dataCtorFields`, and then use qualification where both are in scope. (Presumably that's only `CST.Convert`?)

src/Language/PureScript/CST/Convert.hs
```diff
@@ -446,18 +446,21 @@ convertDeclaration :: String -> Declaration a -> [AST.Declaration]
...
-      ctrs = case bd of
-        Nothing -> []
-        Just (_, cs) -> ctr <$> toList cs
-    pure $ AST.DataDeclaration ann Env.Data (nameValue a) (goTypeVar <$> vars) ctrs
+      ctrs :: SourceToken -> DataCtor a -> [(SourceToken, DataCtor a)] -> [AST.DataConstructorDeclaration]
+      ctrs st (DataCtor _ name fields) tl
+        = AST.DataConstructorDeclaration (sourceAnnCommented fileName st (nameTok name)) (nameValue name) (zip ctrFields $ convertType fileName <$> fields)
+        : (case tl of
+            [] -> []
+            hd : tl' -> ctrs (fst hd) (snd hd) tl'
```
hdgarrood: Bit of a nit, but I think this would be slightly clearer if you unpacked `hd` in the pattern match instead of using `fst` and `snd`.

tests/purs/docs/src/DocCommentsDataConstructor.purs
```diff
@@ -0,0 +1,5 @@
+module DocCommentsDataConstructor where
+
+data Foo
```
hdgarrood: Could you please add some more tests? I think we ought to have at least:

- One where the data type has some type variables,
- One where there are more data constructors, where some have comments and some don't,
- One where the data constructors have one or more fields.
- One which is a `newtype`

kRITZCREEK:
Super excited to have this in!

Just to let you know, I pushed a few changes that make it so `purs ide` also picks up these spans and documentation comments. Once this gets merged we'll get all this nice information in the auto completions as well :)

kRITZCREEK:
Please also add yourself to CONTRIBUTORS.md, thanks!

hdgarrood:
Also, if this solves #3507, can you add another test to confirm that please?

marcosh:
@hdgarrood I added some tests, also regarding #3507 and I address your comments.

please let me know if there is some other improvement I can implement

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -481,14 +481,23 @@ pattern ValueDecl :: SourceAnn -> Ident -> NameKind -> [Binder] -> [GuardedExpr]
...
   = ValueDeclaration (ValueDeclarationData sann ident name binders expr)
 
+data DataConstructorDeclaration = DataConstructorDeclaration
+  { dataCtorAnn :: !SourceAnn
+  , dataCtorName :: !(ProperName 'ConstructorName)
+  , dataCtorFields :: ![(Ident, SourceType)]
+  } deriving (Show, Eq)
+
+traverseDataCtorVars :: Monad m => ([(Ident, SourceType)] -> m [(Ident, SourceType)]) -> DataConstructorDeclaration -> m DataConstructorDeclaration
+traverseDataCtorVars f DataConstructorDeclaration{..} = DataConstructorDeclaration dataCtorAnn dataCtorName <$> f dataCtorFields
```
hdgarrood: Do you think it would make sense to call this `traverseDataCtorFields` now that the field has been renamed?

hdgarrood:
Looks great, thanks very much! Once the above question is resolved I'm happy to merge.

src/Language/PureScript/AST/Declarations.hs
marcosh: sure! done

Issue #1897, opened 2016-02-23
Types with the same name should use module prefix in error messages


user471:
```
import ModuleA as A
import ModuleB as B

test :: A.X
test = B.Bar

module ModuleA where

data X = Foo

module ModuleB where

data X = Bar
```
```
 Could not match type

    X

  with type

    X
```
should be 
```
 Could not match type

    ModuleA.X

  with type

    ModuleB.X
```

AppShipIt:
This is kinda valuable feedback for what I want to tackle. Could we tag these something like `messages`?

hdgarrood:
There is an `error-message` tag already. I think this is a dupe of #1647.

garyb:
Yep, this is #1647 again

AppShipIt:
Thanks for the heads up @hdgarrood!

Issue #1842, opened 2016-01-21
Qualified names aren't qualified in the errors/warnings 


zudov:
For example:

```haskell
module Main where

import Foo.Bar.Baz as Baz

foo :: Baz.MyInt
foo = 0

module Foo.Bar.Baz where

newtype MyInt = MyInt Int
```

Results in:

```
  Could not match type
    Int
  with type
    MyInt
while trying to match type Int
  with type MyInt
while checking that expression 0
  has type MyInt
in value declaration foo
```

It would be much nicer if it referred to `Foo.Bar.Baz.MyInt` in the same way as I did (`Baz.MyInt`)

garyb:
Duplicate of #1647 :)

zudov:
Agrh, tried, but failed to find it.

Issue #1670, opened 2015-11-25
Warn about unused values/declarations


garyb:
There are two cases I have in mind here:

1. Unused values declared in a `let` or `where`
2. Unused declarations in modules that have explicit export lists - the warnings can apply to any declaration that is private to the module and unreachable by the exports, not just members.

paf31:
:+1:

garyb:
Actually, there's a third case too: unused names introduced in binders.

paf31:
Moving to Backlog.

garyb:
I'd like to do this for 0.9 actually.

paf31:
Sounds good.

i-am-the-slime:
Not sure if this is on anybody's priority list but
 - unused type aliases that aren't exported
Would also make my list.

garyb:
:+1:, added it to the list!

hdgarrood:
I've updated the description since #3819 was merged.

PR #3685, opened 2019-06-20
Improve lint import warnings

matthew-hilty:
Two main changes:
1. Add an import declaration's optional qualifier as a parameter to `UnusedImport`.
    - Previously, warnings didn't distinguish between import declarations from the same module.
      Code like the following
      ```purescript
      import A.B (x) -- `x` is used.
      import A.B (y) as C -- `y` is not used.
      ```
      would induce a warning like `The import of module A.B is redundant` even though only the qualified import declaration `C` is actually redundant.
    - The warning now would be `The import of module A.B (qualified as C) is redundant`.
2. Include kind imports among all imports considered by the linter for determining `UnusedImport` and `UnusedExplicitImport` warnings.
    - Previously, kind imports were ignored.
      The linter wouldn't emit any warnings for code like the following.
      ```purescript
      import A.B (kind K) -- `kind K` is not used.
      ```
    - And the linter, disregarding `kind K`, would emit an `UnusedImport` instead of an `UnusedExplicitImport` for code like the following.
      ```purescript
      import A.B (x, kind K) -- `x` is not used, but `kind K` is.
      ```

Issues #1647, #2339, #1897, #1647, #1842, #1670 are similar to this issue, but they relate to warnings/errors about types or exports rather than about imports.


src/Language/PureScript/Sugar/Names/Imports.hs
```diff
@@ -32,10 +32,9 @@ findImports
   -> M.Map ModuleName [ImportDef]
 findImports = foldr go M.empty
   where
-  go (ImportDeclaration (pos, _) mn typ qual) result =
-    let imp = (pos, typ, qual)
-    in M.insert mn (maybe [imp] (imp :) (mn `M.lookup` result)) result
-  go _ result = result
+  go (ImportDeclaration (pos, _) mn typ qual) =
+    M.alter (return . ((pos, typ, qual) :) . fromMaybe []) mn
+  go _ = id
```
hdgarrood: Am I correct in thinking the change in this file is just a refactor which shouldn't change behaviour?

src/Language/PureScript/Errors.hs
```diff
@@ -911,8 +911,13 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
             ]
-    renderSimpleErrorMessage (UnusedImport name) =
-      line $ "The import of module " <> markCode (runModuleName name) <> " is redundant"
+
+    renderSimpleErrorMessage (UnusedImport mn qualifier) =
+      let
+        mark = markCode . runModuleName
+        qual' q = " (qualified as " <> mark q <> ")"
+        qual = maybe "" qual'
+      in line $ "The import of module " <> mark mn <> qual qualifier <> " is redundant"
```
hdgarrood: Bit of a nit but may I suggest copying GHC and saying "The qualified import of [module] as [qualified-name] is redundant"?

src/Language/PureScript/Sugar/Names/Imports.hs
matthew-hilty: Yeah. I think Gary's advice from a different pull request was still in mind when I made this change. It's meant to be just a simple refactor like you say, but maybe I should undo this patch so everything here is related directly to linting. I don't mind either way.

src/Language/PureScript/Errors.hs
matthew-hilty: Sure, that seem likes a good idea. I'll change the message, and I'll probably be able to push up the changes by tonight.

(By the way, please, pick as many nits as you want. I appreciate any tips/advice you guys have for me.)

src/Language/PureScript/Sugar/Names/Imports.hs
hdgarrood: Happy to keep this in here - I was just wondering.

hdgarrood:
@matthew-hilty looks like your rebase has gone a bit wrong. Would you mind trying again?

thomashoneyman:
Anything stopping this one from going in? Looks like it's up to date and passing tests, and I encountered a situation today at work where this would have been helpful.

hdgarrood:
Nope! Thanks again @matthew-hilty

PR #3687, opened 2019-06-23
Try to fix windows CI

hdgarrood:


hdgarrood:
I am fairly sure that what really broke our caches is the fact that we did not pin our Stack version, and upgrading to Stack 2.1 has caused our caches to become corrupted somehow. Deleting them via Travis CI's web interface seems to have fixed this. Pinning the Stack version we use (as this PR does) should prevent this from happening again.

PR #3692, opened 2019-06-28
Bump to v0.13.1

hdgarrood:
I would like to cut a new release because people keep tripping over the node 10 requirement for `npm install purescript`, and this allows us to use the new npm-installer which works with node 8. Any objections?


f-f:
@hdgarrood could we get #3676 in this release too?

hdgarrood:
Yes, good idea.

PR #3695, opened 2019-07-04
Add placeholder purs.bin to fix npm installs

hdgarrood:
I just tried publishing 0.13.1 to npm, and unfortunately the npm package was broken because it did not include a placeholder `purs.bin` file. This placeholder file is necessary for successful publishing, because on installation, the npm cli will attempt to read this file once before the postinstall script is run (at which point it will be replaced with the actual binary).

In this PR, I've added a placeholder `purs.bin` file. I've also made this file produce a sensible error if you try to execute it, which wasn't previously the case (the previous one is just a text file). To illustrate, here's the previous one:

```
$ npm pack purescript@0.13.0 # downloads an 0.13.0 tarball
$ tar xOf purescript-0.13.0.tgz package/purs.bin # print the placeholder purs.bin file to stdout
This is a placeholder file of a PureScript binary installed with npm.
If you see this file, that means the installation has failed
and the placeholder has not been replaced with a valid binary.
Try installing `purescript` npm package again.
```

Also, this placeholder file was previously ignored by git for some reason, so to prevent this happening again, I've checked the new file in.

I've unpublished the broken 0.13.1 version on npm but we aren't allowed to publish over that version again, so we'll need to publish this as 0.13.2. I've published the code in this branch to npm as `0.13.2-pre.1` if you want to test it out. Once this PR is approved, I will make an 0.13.2 release with no additional Haskell code changes compared to 0.13.1, although we'll need to bump the version in package.yaml too so that `purs --version` agrees with what npm says.


hdgarrood:
Oh, oops, `npm install purescript@0.13.2-pre.1` won't work any more, because it is pointed at the v0.13.1 GitHub release, and I've just removed all of the release notes and binaries from there.

f-f:
@hdgarrood you might be interested in a similar situation we encountered with Spago: https://github.com/spacchetti/spago/pull/215

hdgarrood:
Thanks for the heads up. I think for now I'd just like to get the `purescript` npm package working the way it was before; we can investigate to see if this suffers from any of the same issues once the release is out. (I also would quite like to preserve the current property that you don't have a shell script wrapper for the compiler, because that tends to complicate things a bit.)

PR #3699, opened 2019-07-20
Refactor and simplify BuildPlan a little

hdgarrood:
Instead of having two separate MVars for build job results and errors,
just have one, which contains a sum type, to indicate if and how a build
job has completed with a little more clarity and safety (in the sense
that this makes some invalid states unrepresentable).

Additionally, rather than having two separate functions for consuming
the result of a build plan, namely `collectErrors` and `collectResults`,
and requiring that the first is called before the second, unify them
both into `collectResults`.

This will help for #3145, as for that I need the BuildPlan to be able to
expose which build jobs succeeded before any errors are rethrown, so
that we can store their timestamps and hashes in preparation for the
next build.

I haven't done any performance tests on this just yet, but I don't
anticipate any drastic changes.

### DIFF
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -22,7 +22,7 @@ import           Data.Function (on)
 import           Data.Foldable (for_)
 import           Data.List (foldl', sortBy)
 import qualified Data.List.NonEmpty as NEL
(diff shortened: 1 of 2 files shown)


hdgarrood:
The profiling output is pretty much the same for this versus the current master.

PR #3709, opened 2019-07-31
Add support for local type synonyms

### LINKED ISSUE #3708: Feature: local type synonyms
#### Summary
This is a proposal to support locally scoped type synonyms inside `where` and `let` binding lists, with the obvious syntax and semantics.

#### Motivation
There are a few reasons why one might want to define a local type synonym. The simplest is locality itself: sometimes it's desirable to bundle up one or two dozen functions into a ‘submodule‘ consisting of an outer function with a long list of `where` bindings. If those inner functions reuse a complicated type a few times such that a type synonym would be useful, a synonym defined near to where it's used is more legible than one that's out in the top-level scope, possibly off-screen. Another reason is brevity: a top-level type synonym demands a name that is unambiguous in the context of the entire module, whereas a type synonym scoped only to a single definition can be more succinct without clogging up the outer namespace. My favorite reason, though, is that defining type synonyms for types involving type variables requires you to parameterize the type synonym as well... unless it shares a scope with the type variable. Consider the difference between these two snippets—IMO, the readability gap is obvious, and continues to widen as you imagine the number of type variables involved growing:

```purs
type Foo a b c = { alpha :: a, bravo :: b, charlie :: c }
type Paths a b c = Map (Foo a b c) (Set (Foo a b c))

bigFunction :: forall a b c. ...
bigFunction ... = ...
  where
  ...
  computePaths :: ... -> { forward :: Paths a b c, backward :: Paths a b c }
  computePaths = ...
```

```purs
bigFunction :: forall a b c. ...
bigFunction ... = ...
  where
  type Foo = { alpha :: a, bravo :: b, charlie :: c }
  type Paths = Map Foo (Set Foo)
  ...
  computePaths :: ... -> { forward :: Paths, backward :: Paths }
  computePaths = ...
```

(All of these reasons have their analogs in terms of why `let`/`where` bindings for values are useful even though we have top-level value declarations—in principle, many inner bindings could be replaced with a top-level declaration that is likely to suffer from at least one of: being farther away, having a longer name, or needing more parameters. But why suffer?)

#### Proposal
Syntactically, make `type` declarations legal inside the binding lists of `let` expressions, `where` clauses, and anything parsed similarly. Semantically, these synonyms are sugar for equivalent synonyms hoisted to the top level, with unique names and extra parameters added as needed. Like other `let` bindings, local type synonyms can't be exported or used outside of their scope (which does *not* include the type signature of the value containing the binding--top-level values must be annotated with types visible from the top-level scope), and like regular type synonyms, they can't have instances or cyclic definitions, and their lexical position within the binding list is irrelevant.

#### Examples
```purs
-- none of the below type synonyms would be valid in this signature
foo :: forall a. a -> { left :: Int, right :: a }
foo a = result
  where
  result = result'
    where
    -- `a` here shadows outer `forall a`, but `Alpha` must remain bound to the quantified `a`
    type Bravo a = { left :: a, right :: Alpha }
    result' :: Bravo Int
    result' = { left: 0, right: a }
  type Alpha = a -- type synonyms needn't precede their use in a binding group

-- "anything parsed similarly"
bar :: Effect Int
bar = do
  -- `Alpha` would not be valid here...
  log "hello"
  let type Alpha = Int
  pure (1 :: Alpha) -- ...but it is here!
```

---

I have an only-mildly-hackish implementation of this to be offered as a PR shortly; I've been using it locally for a week and this feature has really helped me clean up some nasty signatures.

rhendric:
This implements `type X = Y` synonyms as another form of let binding,
valid inside `let` expressions, `where` clauses, and `do` notation.
These synonyms are hoisted to the top level during desugaring.

---

This is a complete attempt at #3708, but subject to change if there's feedback on that feature request. Feedback on this specific implementation is welcome here; feedback on the concept of local type synonyms should probably go there.

The biggest sin I have to confess here is using the hoisted type synonym name mangling to store data, as opposed to just being a unique string. The excuses I offer are that the data are only retrieved by reporting code in `Language.PureScript.Errors` and `Language.PureScript.Pretty`, not anywhere ‘important’, and there only through functions defined alongside the name mangling code so that the ‘magic string’ logic stays local; and that the alternative would involve changing `ProperName`, changing `TypeSynonymDeclaration`, or adding a new constructor alongside `TypeSynonymDeclaration`, any of which would likely require refactoring code in a much wider scope than this PR currently targets—an outcome worth avoiding not only because I'm lazy, but also because I figured you would prefer it if I kept my grubby noob hands out of the typechecker. :smile: As always, I'm happy to do the work if I'm wrong.

Closes #3708.

### DIFF
--- a/lib/purescript-ast/src/Language/PureScript/Names.hs
+++ b/lib/purescript-ast/src/Language/PureScript/Names.hs
@@ -10,7 +10,9 @@ import Prelude.Compat
 import Codec.Serialise (Serialise)
 import Control.Monad.Supply.Class
 import Control.DeepSeq (NFData)
+import Data.Bifunctor (bimap)
 import Data.Functor.Contravariant (contramap)
+import Data.Maybe (fromMaybe)
 import qualified Data.Vector as V
 
 import GHC.Generics (Generic)
@@ -236,6 +238,41 @@ isUnqualified = not . isQualified
 isQualifiedWith :: ModuleName -> Qualified a -> Bool
 isQualifiedWith mn (Qualified (Just mn') _) = mn == mn'
 isQualifiedWith _ _ = False
+
+-- |
+-- Given a scope identifier (expected to start with "$"), a depth (number of
+-- type variables in scope), and a base name, generate the mangled name to use
+-- for a hoisted type synonym declaration.
+--
+makeLocalSynonymName :: Text -> Int -> ProperName 'TypeName -> ProperName 'TypeName
+makeLocalSynonymName scope depth name = ProperName $ T.concat ["local", scope, "$", T.pack (show depth), "$", runProperName name]
+
+isLocalSynonymName :: ProperName 'TypeName -> Bool
+isLocalSynonymName = ("local$" `T.isPrefixOf`) . runProperName
+
+-- |
+-- If the provided name is the generated name of a hoisted local type synonym,
+-- extract from it the number of synthetic type parameters added during
+-- hoisting and its original name.
+--
+readLocalSynonymData :: ProperName 'TypeName -> Maybe (Int, ProperName 'TypeName)
+readLocalSynonymData = fmap extractData . T.stripPrefix "local$" . runProperName
+  where
+  extractData = bimap (read . T.unpack) (ProperName . T.tail)
+              . T.break (== '$') . T.tail . snd . T.break (== '$')
+
+-- |
+-- A wrapper for runProperName that drops local type synonym mangling.
+--
+showTypeSynonymName :: ProperName 'TypeName -> Text
+showTypeSynonymName name = runProperName . fromMaybe name . fmap snd . readLocalSynonymData $ name
+
+showQualifiedTypeSynonymName :: Qualified (ProperName 'TypeName) -> Text
+showQualifiedTypeSynonymName name = fromMaybe (showQualified runProperName name)
+                                  . fmap (runProperName . snd)
+                                  . readLocalSynonymData
+                                  . disqualify
+                                  $ name
 
 $(deriveJSON (defaultOptions { sumEncoding = ObjectWithSingleField }) ''Qualified)
 $(deriveJSON (defaultOptions { sumEncoding = ObjectWithSingleField }) ''Ident)
--- a/lib/purescript-ast/src/Language/PureScript/Types.hs
+++ b/lib/purescript-ast/src/Language/PureScript/Types.hs
@@ -467,7 +467,7 @@ replaceAllTypeVars = go [] where
   go bs m (ForAll ann v mbK t sco)
     | v `elem` keys = go bs (filter ((/= v) . fst) m) $ ForAll ann v mbK' t sco
     | v `elem` usedVars =
-      let v' = genName v (keys ++ bs ++ usedVars)
+      let v' = genName (`elem` (keys ++ bs ++ usedVars)) v
           t' = go bs [(v, TypeVar ann v')] t
       in ForAll ann v' mbK' (go (v' : bs) m t') sco
     | otherwise = ForAll ann v mbK' (go (v : bs) m t) sco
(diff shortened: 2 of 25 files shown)


hdgarrood:
Could you please add some passing and failing tests where a top-level declaration does not have a type signature but there is a local type synonym used in its definition? I feel like it would be good to test that, e.g., the inferred type for the top level declaration does not mention the local synonym in such cases.

hdgarrood:
(although you might want to wait until we've made a decision on whether to move forward with this, sorry)

rhendric:
I added tests as requested (see failing/LocalTypeSynonyms5 and 6, and the end of passing/LocalTypeSynonyms); not sure they cover everything you had in mind though. Mind taking a look? (The build is failing for the same reason master is; I don't see any new hlint complaints in this PR.)

hdgarrood:
Yeah, that's what I was imagining. Thanks! Also could I request that you not force-push, since it makes it a bit harder to review? We can save commit rewriting for when we're ready to merge.

rhendric:
Oh, of course! Sorry for the trouble.

hdgarrood:
No worries! :slightly_smiling_face:

natefaubion:
Do we need to support standalone kind signatures for polykinds?

rhendric:
Probably! I can easily add some tests that copy what the type synonym parts of the StandaloneTypeSignatures tests do, only in an inner scope; but I think there are probably more interesting things to test around using and/or shadowing kinds from the outer scope. I'm not very comfortable with polykinds yet; do you have any particularly tricky things to exercise in mind?

natefaubion:
The primary issues I'm thinking about are:
* We generalize the kinds of type-level declarations. Type-level declarations are all top-level, and so they have the same generalization strategy as top-level value declarations. It's not clear to me that local type synonyms should be generalized, since we don't otherwise generalize local value declarations.
* We warn on generalization of type-level declarations like we warn on generalization of value declarations. If we are hoisting and generalizing local declarations, what do these warnings look like?

These issues are part of what make me hesitant to implement this as a desugaring step.

rhendric:
I could be missing something here, but is generalization a concern for type synonyms? Since there can't be cycles in type synonyms, eventually all type synonym expansions will bottom out in some type made of top level data, from which kinds can be inferred. Any parameters to type synonyms that aren't included in this final expansion are not only generalizable, they're completely irrelevant, right? There's no point in having them at all: type synonyms can't be partially applied, and everything else in the language that uses types such as instance resolution depends on the expansion, not the synonym form. So generalizing makes no difference; unused type synonym parameters at any scope, top-level or not, could be *errors* and I don't think any real programs would be affected. Is that right?

natefaubion:
I should clarify, I don't think there's necessarily an implementation problem with generalizing, it's more about expectations around where and how we generalize in PureScript, and if we want a consistent rule-set for generalization.

natefaubion:
An example to test for right now would be defining a local `type LocalApp a = f a` synonym where `f` is bound in the current scope. How does the compiler treat that and warn?

rhendric:
> An example to test for right now would be defining a local `type LocalApp a = f a` synonym where `f` is bound in the current scope. How does the compiler treat that and warn?

Ah, thank you, I think I see the problem now. We're talking about generalization of the type parameters that get added from the local scope. Prior to polykinds, these all had definite kinds at the `forall` site, either explicitly stated or implicitly `Type`. But now they can generalize, and the compiler warns about that, and is that right?

On reflection, this seems like a specification question rather than a question about this particular implementation (although, even in the absence of a right thing, this implementation is clearly doing a wrong thing by issuing a warning about `type local$1$1$LocalApp`). I propose moving this conversation over to #3708.

rhendric:
I'm closing this in favor of #3897, as I'm now reasonably convinced that getting kind signatures and the desired generalization behavior right with desugaring is more trouble than not. There are very probably still issues to work out with #3897 but for now it certainly looks like the more practical path forward (assuming of course that the maintainers ultimately decide the feature is desirable at all).

PR #3713, opened 2019-08-08
Fix parsing of comma-separated guards in let statements

natefaubion:
Parsing was failing in do/let guards if a comma was inserted. It was prematurely inserting closing layout delimiters before the comma due to the missing cases for `LytLetStmt`.


colinwahl:
Thanks @natefaubion !

Issue #3613, opened 2019-04-28
ES modules


hdgarrood:
Previously: #2207, #2558, #2574, #2575, #3427. I'm making a new issue because the discussions in many of the linked issues are quite long, and because the ES modules landscape has changed quite a lot since when most of the discussion happened two years ago (in particular, all major browsers now support ES modules). Also, we should probably implement emitting ES modules at the same time as switching to ES modules for FFI, so I think all of these things should be considered together.

ES modules provide a number of benefits over CommonJS; most notably easier dead code elimination and wide browser support (see <https://caniuse.com/#feat=es6-module>). Having the compiler use and understand ES modules natively would mean that many workflows wouldn't require bundling at all, which would significantly speed up and simplify things, especially during development.

It should remain possible to run a PureScript-only app in modern browsers and in Node.js with just `purs`, i.e. without having to bring in external JS tools such as webpack. Currently this is achieved by invoking `purs compile` and then `purs bundle` (or just `purs compile` if you only want to run on Node.js).

I think the following is relatively uncontroversial, and is what we generally have settled on in the above linked issues:
* Compile import declarations in PureScript to ES module import statements, so that `import Prelude` compiles to something like `import * as Prelude from "../Prelude/index.js"`. Note that `import Prelude` currently compiles to `var Prelude = require("../Prelude/index.js")`.
* Compile exports as ES module exports, perhaps `export { apply, identity }` rather than the current `module.exports = { apply: apply, identity: identity }`
* Mangle primes in imports and exports in both regular PureScript modules and in FFI modules, so for example if a PureScript module exports an identifier `foo'`, then the compiled ES module exports it as `foo$prime`. In particular, note that writing `foreign import foo' :: Int` in a PureScript file means that the compiler should look for `foo$prime` in the corresponding FFI module.
* Support writing FFI modules both as ES modules and with the current CommonJS format, at least for a period of time.

[...]

PR #3714, opened 2019-08-08
WIP: Print Functions as Arrow-Functions when generating JS

csicar:
Arrow Functions have the advantage of ~better performance~ (EDIT: actual performance gains are unclear) and are easier to read (especially with curried functions).

This PR is a PoC for using arrow functions where ever it is possible. If there is interest in implementing it in the compiler, I'll clean the code up.

- [x] tests are passing

### Changes
- Addition of `Constructor` to CoreImp AST
- prettyPrint for Constructor
- prettyPrint for Function

```purescript
module Test where

f as b cs = [as, b, cs]

c n = {n}

data Test = Test Int

testVal :: Test
testVal = Test 1

class ShowTest a where show :: a -> String

instance showTest ∷ ShowTest Test where show _ = "test"
```
now generates:
```js
"use strict";
var Test = (() => {
    function Test(value0) {
        this.value0 = value0;
    };
    Test.create = value0 => new Test(value0);
    return Test;
})();
var ShowTest = function (show) {
    this.show = show;
};
var testVal = new Test(1);
var showTest = new ShowTest(v => "test");
var show = dict => dict.show;
var f = as => b => cs => [ as, b, cs ];
var c = n => ({
    n: n
});
...
```

instead of:
```js
"use strict";
var Test = (function () {
    function Test(value0) {
        this.value0 = value0;
    };
    Test.create = function (value0) {
        return new Test(value0);
    };
    return Test;
})();
var ShowTest = function (show) {
    this.show = show;
};
var testVal = new Test(1);
var showTest = new ShowTest(function (v) {
    return "test";
});
var show = function (dict) {
    return dict.show;
};
var f = function (as) {
    return function (b) {
        return function (cs) {
            return [ as, b, cs ];
        };
    };
};
var c = function (n) {
    return {
        n: n
    };
};
...
```

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -204,7 +204,7 @@ moduleToJs (Module _ coms mn _ imps exps foreigns decls) foreign_ =
     extendObj obj sts
   valueToJs' e@(Abs (_, _, _, Just IsTypeClassConstructor) _ _) =
     let args = unAbs e
-    in return $ AST.Function Nothing Nothing (map identToJs args) (AST.Block Nothing $ map assign args)
+    in return $ AST.Constructor Nothing Nothing (map identToJs args) (AST.Block Nothing $ map assign args)
     where
     unAbs :: Expr Ann -> [Ident]
     unAbs (Abs _ arg val) = arg : unAbs val
--- a/src/Language/PureScript/CodeGen/JS/Printer.hs
+++ b/src/Language/PureScript/CodeGen/JS/Printer.hs
@@ -167,7 +167,16 @@ indexer = mkPattern' match
 lam :: Pattern PrinterState AST ((Maybe Text, [Text], Maybe SourceSpan), AST)
 lam = mkPattern match
   where
+    -- exclude Object in Return from arrow function syntax to avoid () => {a : 1}
+  match (Function ss Nothing args ret@(Block _ [Return _ (ObjectLiteral _ _)])) = Just ((Nothing, args, ss), ret)
+  match (Function ss Nothing args (Block _ [Return _ expr])) = Just ((Nothing, args, ss), expr)
   match (Function ss name args ret) = Just ((name, args, ss), ret)
+  match _ = Nothing
+
+lamC :: Pattern PrinterState AST ((Maybe Text, [Text], Maybe SourceSpan), AST)
+lamC = mkPattern match
+  where
+  match (Constructor ss name args ret) = Just ((name, args, ss), ret)
   match _ = Nothing
 
 app :: (Emit gen) => Pattern PrinterState AST (gen, AST)
@@ -217,6 +226,13 @@ prettyStatements sts = do
   indentString <- currentIndent
   return $ intercalate (emit "\n") $ map ((<> emit ";") . (indentString <>)) jss
 
+prettyPrintFunction :: Maybe Text -> [Text] -> Text
+prettyPrintFunction Nothing [arg] = arg <> " => "
+prettyPrintFunction Nothing args = "(" <> intercalate ", " args <> ") => "
+prettyPrintFunction (Just name) args = "function "
+                          <> name
+                          <> "(" <> intercalate ", " args <> ") "
+
 -- | Generate a pretty-printed string representing a collection of JavaScript expressions at the same indentation level
 prettyPrintJSWithSourceMaps :: [AST] -> (Text, [SMap])
 prettyPrintJSWithSourceMaps js =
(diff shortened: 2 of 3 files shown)


garyb:
I'd personally vote nay to this just now. The compiler currently outputs ES3 code, and we don't use anything beyond ES5 in the core libraries either. 

The aesthetic improvement it offers doesn't outweigh the fact the PS code is currently usable basically anywhere without further processing, for me at least. There is certainly a size improvement also, if only the plain text is considered, but the repetitive nature of `function` declarations mean it gets gzipped away anyway.

That's just my 2p anyway, maybe others feel differently. 🙂 

If we were going to accept it, I think the code for `purs bundle` will need some work too, to recognise the new forms.

csicar:
Thank you for the quick feedback. That reasoning is perfectly understandable to me. I quess the added complexity pretty printing doesn't help either.

As far as I know, `purs bundle` does (mostly) work with this change. At least `purs bundle output/**.js > bundle.js` produces a valid JS file. Is there a better method for testing it?

spicydonuts:
The aesthetic improvement outweighs it for me. Plus `ES[latest] -> ES5` is a problem already solved many times over. `ES5 -> ES[6,7,8]` isn't possible. The compilation process drops useful information and then uses a custom optimizer instead of working with the existing ecosystem. I really don't get it.....

hdgarrood:
Very much agree with @garyb on this one, so a vote against from me too. Sorry.

Do you have any sources to back up the claim that they have better performance?

hdgarrood:
@spicydonuts What information is lost by generating ES5 function expressions?

natefaubion:
Do we still intend to move to ES modules eventually? If so, I don't see a reason to to stick with ES5 since it already won't be compatible.

spicydonuts:
For example, the existing bundler is basically ES modules plus this flag: https://webpack.js.org/configuration/optimization/#optimizationsideeffects

@hdgarrood The biggest one is modules, but const and arrow functions are also useful. There have already been many threads on modules over the years, mostly getting this same response. I remember reading that arrow functions are supposed to be _able_ to be faster because they lack `this` and `arguments`, but whether they are at the moment is just VM specific and changing it back and forth based on tests would be silly. But semantically they're more like PS functions (again, no `this` or `arguments`). But using them in FFI and having nicely formatted output for dev is much more significant.

csicar:
> Do you have any sources to back up the claim that they have better performance?

For example: https://jsperf.com/function-vs-arrow-function On my machine, the normal function performed 17% worse.  But I also tests, that showed the opposite. I guess that fact is more a myth than actual truth.

hdgarrood:
> Do we still intend to move to ES modules eventually? If so, I don't see a reason to to stick with ES5 since it already won't be compatible.

Yes - see #3613. The way I see it, there's an important difference between ES modules and arrow functions though, because part of #3613 is updating `purs bundle` to handle ES modules, which means that it would still be possible to produce JS code which works pretty much everywhere using only `purs` - by running `purs compile` followed by `purs bundle`. Generating functions as arrows functions, by contrast, will require adding additional JS tooling to your build pipeline if older browser compatibility is an issue.

I really don't think this is justifiable since [caniuse.com estimates that 7.5% of global users are on browsers which don't support arrow functions](https://caniuse.com/#feat=arrow-functions).

@spicydonuts 

> There have already been many threads on modules over the years, mostly getting this same response.

This is not true. #3613, the issue for ES modules, has not had a single word against it so far. I think we can safely assume that ES modules are going to happen at this point. I'd argue that they should be in the next breaking release.

hdgarrood:
> But really it's the "it'd make this one use case targeting some old vm somewhere a little easier because I wouldn't have to pipe the output through [insert tool of choice, probably a few with the right defaults] and that'd really be a pain" vs "there are a lot of modern workflows we don't fit into and tools that don't work as well as they should and it's impossible to work around but I guess that's how it is" for years that's getting to me I guess.

This doesn't make sense to me; whether or not we generate arrow functions doesn't have any relevance to how PureScript fits into modern workflows. It sounds to me like you're talking about ES modules, which is entirely separate.

spicydonuts:
They’re related because they’re both given the same counter argument, an argument which disappears once one of them is accepted.

hdgarrood:
The argument against generating arrow functions still applies even if we are producing ES modules, as I explained in https://github.com/purescript/purescript/pull/3714#issuecomment-519892269

csicar:
I guess this PR can be closed then?

spicydonuts:
I guess I'm arguing more against "the output must be es5 and we need a ps-specific bundler" than I am for "yes arrows". I still don't get it but I seem pretty alone on that so I'll drop it.

garyb:
Some people are incredibly hostile to JS tooling, even when they're using PS to produce JS, strangely, so avoiding hard requirements on it is definitely preferable.

Not to mention that, say, `rollup` for example is painfully slow at tree shaking (although it does indeed do a better job at it) compared to `purs bundle`, which isn't exactly fast itself.

spicydonuts:
> hostile to JS tooling
Yeah.. that hostility seems like a huge problem for a tool primarily used to write JS. I don't get why it's so preferrable to cater to it. I get if it's a feature that might limit other backends, but the JS backend should unapologetically target JS.

> `rollup` for example is painfully slow at tree shaking
Relative to the whole build? It's a production minification step, not something that interferes with the dev experience.

hdgarrood:
I do think it's worth considering deprecating `purs bundle` and saying that people can use JS tooling instead for dead code elimination and bundling once ES modules are done (especially since not bundling at all during dev is becoming more feasible and more attractive as an option as browsers implement ES modules). However, I'm reluctant to do it at the same time as implementing ES modules, because that's already going to be a very far-reaching change. Anyway can we discuss on #3613 please?

hdgarrood:
Closing as we don't want to generate arrow functions.

Issue #3605, opened 2019-04-13
Improve documentation for instance chains


hdgarrood:
Currently, instance chains aren’t documented any differently from a collection of normal (non-chained) instances: https://pursuit.purescript.org/packages/purescript-either/4.1.1/docs/Data.Either.Inject

I think this could be improved. Ideally it would look somewhat similar to what you see in the source file, while also being consistent with the look and feel of HTML documentation. I don’t have a concrete proposal yet; suggestions welcome.

garyb:
Just adding `else` in front of the relevant instances might be enough? Assuming the ordering in here is preserved from the definitions - I see it is in this case, but don't know if that's true in general.

hdgarrood:
I think the ordering is preserved, yeah. I think having an `else` in front of the relevant instances gets us most of the way there, but it should be easy to tell which instances are in a chain and which aren't, too (and for me personally, I'm not sure adding `else` is sufficient to achieve that). If we had all of the instances in a chain appear under a single `li` element in the list of instances I reckon I'd be happy.

csicar:
> If we had all of the instances in a chain appear under a single li element in the list of instances I reckon I'd be happy.

To do that cleanly would probably require changing the JSON representation of docs, which seems to be a concern (as there are tests for backwards compatability). Is that worth it?

hdgarrood:
I can see that it would require changing the JSON representation of docs, but I think we should be able to do so in such a way that backwards compatibility is preserved. Note that we don't need older compilers to be able to parse the new JSON, we just need newer compilers to be able to parse the older JSON.

csicar:
In that case, keeping backwards compatibility is indeed possible: Making the `chained` field optional with default value `false` should do the trick: [https://github.com/purescript/purescript/pull/3715/commits/d1a547ea09545184609237bea7206696fce979f2]

Having the `instance` field be an array in the json and accepting a single object for backwards compatability would also be possible

PR #3715, opened 2019-08-09
show `else` prefix for instance chain items in docs

csicar:
Fix for Issue #3605

- [x] test complete successfully

example:
example

## Changes
- added `Bool` to ChildInstance Constructor
- added `chained` to JSON representation, with default value `False` for backwards-compatability
- added rendering for `else` keyword

## Notes

Displaying the instances of a type can look weird because the `else` prefix has no meaning there. I would suggest hiding the `else` keyword when displaying anywhere except a type class context
weird case


hdgarrood:
> Displaying the instances of a type can look weird because the else prefix has no meaning there. I would suggest hiding the else keyword when displaying anywhere except a type class context

Yes please, good call.

csicar:
The solution in bc3cc4e  is not very elegant, but works. Basically the problem is, that rendering of the ChildDeclaration needs to know if the parent declaration is a TypeClass. 

Alternatively, I could pass the Parent declaration to `Renderer.renderChildDeclaration`, which might be nicer (depending on preference).

hdgarrood:
This is actually quite a bit more subtle than I first expected. I think what we have here is still not quite right, for a couple of reasons:

1. It could be misleading to display an instance from a chain outside of the context of that chain. If we were to display an instance which is halfway down a chain without the instances above it in the chain, it would appear that this instance would be selected more often than it actually is. 
2. It's possible that a particular data type appears more than once in an instance chain, in which case all of the instances in that chain mentioning that type would not appear to be chained, even though they should be.

I'm trying to come up with some convincing examples. I think the correct approach here may be to represent the entire chain as a single `ChildDeclaration`, and to include the entire chain on the instance list of data types whenever the data type appears at least once anywhere in the chain.

csicar:
New version is improved in pretty much every way:
- Type Class instances are now modeled as a List
- JSON is broken in an backward-compatible way
    - [ ] make sure reading old json will generate correct html
- references are working & instance chain can now be referenced by their instanceId in HTML (For instance chain of `instanceA` `instanceB` the id is `instanceA-else-InstnaceB`)

TODO:
- [ ] fix sourceAnn and comments for instance chains
- [ ] test Markdown

PR #3716, opened 2019-08-16
Suggest import for name causing a UnknownName error

csicar:
This PR adds the adds a Hint to the UnknownName error, if a module exports the same name. It (should) support all names including types, type classes, values etc. 

## Changes
- addition of a new Hint for UnknownName
- renderning for the hint
- call to add these suggestions in `Make.make`
- add module for future suggestions

## ToDos
- [ ] Move from Hint to Field on UnknownName (to support JSON error suggestions)
- [ ] Check performance. If needed limit search / build map from Name to possible imports
- [ ] SourceSpans
- [ ] Reuse existing imports if possible: E.g. if `Prelude (map)` is already imported, suggest `Prelude (map, class Show)`

If there is interest in this, I'll implement these ToDos.
Ensuring, that the value has the correct type would be nice but seems quite involved, as none of the data structures available at the necessary places seem to have type information available.

Example:
```
Error 3 of 4:

  in module Data.DotLang.Attr.Edge
  at src/Data/DotLang/Attr/Edge.purs:22:1 - 23:21 (line 22, column 1 - line 23, column 21)

    Unknown type class Show

  Perhaps you want to add it to the import list? 2 possible imports were found:
                                 
    import Data.Show (class Show)
    import Prelude (class Show)  
                                 

```

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -205,6 +205,7 @@ data ErrorMessageHint
   | ErrorInTypeClassDeclaration (ProperName 'ClassName)
   | ErrorInForeignImport Ident
   | ErrorSolvingConstraint SourceConstraint
+  | ErrorMissingImport (NEL.NonEmpty (ModuleName, DeclarationRef)) -- List of possible import matching the expected name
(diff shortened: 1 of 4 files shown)


csicar:
`stack test` is passing all 904 tests

PR #3720, opened 2019-08-19
Solve union constraints backwards

MonoidMusician:
The union solver for `Union l r u` was missing the case when `r` and `u` are literals (which should solve, according to the fundeps, and indeed is possible to solve).

I don't believe there's an existing issue for this, which means it potentially went unnoticed for two years …

So according to my limited understanding, these solving methods return the "new" values of the type heads, which get unified with the old ones. How my algorithm works is that, when `r` and `u` are literals, we try to partition `u` into the labels that belong to `r` (which are at the tail of `u`) and the remaining that belong to `l`. These then must get unified with the original values in order to make progress solving the constraint. In particular, if `l` is a variable, it will get set to the value that makes it all work. Note that if there are labels present in `r` that are not present in `u` we must fail (`guard (null leftover)`).

Is my understanding of the solver accurate? Feedback is more than welcome!

I also need advice on where best to put the tests! I will add more tests, but I'm reasonably confident that it works at this stage. In particular, it seems to work very well for the use case @masaeedu and I had (which relies integrally on solving this constraint backwards).


natefaubion:
I think the solver portion looks good. As for tests, I think it's fine to keep them in this module, but maybe something like:
```purescript
solveL :: forall l r u. Union l r u => RProxy r -> RProxy u -> RProxy l
solveL _ _ = RProxy

solveUnionBackwardsNil :: RProxy ()
solveUnionBackwardsNil = solveL (RProxy :: _ ()) (RProxy :: _ ())

...etc
```

natefaubion (approved):

rhendric:
Fix the conflict and add an entry to the changelog and we can merge this, yeah?

hdgarrood:
Sounds good to me :+1:

JordanMartinez:
@MonoidMusician Just FYI. I've submitted a PR that fixes the merge conflict and adds a changelog entry in monoidmusician/purescript#1

MonoidMusician:
Thanks!! I'll take a look in a week after I'm done with my semester.

JordanMartinez:
Will this PR will make it into the next release? 

I'm not trying to rush this, but just to clarify, the merge conflict is quite simple (spaces added to better compare differences):
```
<<<<<<< solve-union-backwards
import Data.List (             minimumBy, groupBy, nubBy, sortBy, elem, delete)
=======
import Data.List (findIndices, minimumBy, groupBy, nubBy, sortBy)
>>>>>>> master
```
If we merge the two lists together and remove the unneeded import `elem`, then we get this import list:
```
import Data.List (findIndices, minimumBy, groupBy, nubBy, sortBy, delete)
```
While the conflict could be resolved here, I wasn't sure whether that would be rude to do so (hence my PR).

Even if the conflict is solved here, a commit adding this change to the changelog would still need to be pushed. 

Lastly, I'm not sure how this would be affected by #4090.

JordanMartinez:
Any updates on this?

MonoidMusician:
Thanks for your patience! I'm sorry I got busy, but I feel good about this now. I ran HLint and fixed the one linter issue I introduced. I also ran more test cases in the REPL and it works exactly like I would expect, include failure cases.

Do you want to include this in the next release or merge it after?

PR #3721, opened 2019-08-19
Ambiguous variable check: take closure of fundeps across constraints

### LINKED ISSUE #3238: Ambiguous type check should take determined type arguments into account
The type checker currently reports an ambiguous type ("...has type variables which are not mentioned in the body of the type") even when the only type variables not mentioned are determined by type class constraints. This probably shouldn't be considered ambiguous.

MonoidMusician:
Fixes #3238

Previously the check (which occurs when generalizing types) only took into account each class individually and would miss variables that were determined from the interaction of multiple constraints.

I also updated the error message to say which variables could not be determined, especially since that information is readily available now.

Feedback:
- Is `Protolude.atMay` the proper function to use to handle partiality? I couldn't really find one …
- Does the error message look okay?

```
Error found:
in module Data.Record.Operations
at test/Operations.purs:146:1 - 146:99 (line 146, column 1 - line 146, column 99)

  The inferred type
                                                                                                                                                                                                                                                                                                                                                                                                        
    forall t40 t41 t43 t44 t48 t49 t51 t52 t55 t56. RowToList t44 t41 => Lacks "onMouseEnter" t41 t55 => AppendCase t55 "onMouseEnter" String (Cons "onMouseLeave" String Nil) t41 t40 => ListToRow t40 t43 => RowToList t52 t49 => Lacks "onMouseLeave" t49 t56 => AppendCase t56 "onMouseLeave" String Nil t49 t48 => ListToRow t48 t51 => Tuple (Record t44 -> Record t43) (Record t52 -> Record t51)
                                                                                                                                                                                                                                                                                                                                                                                                        
  has type variables which are not determined by those mentioned in the body of the type:

    t40 could not be determined
    t48 could not be determined

  Consider adding a type annotation.

in value declaration e2
```

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -113,7 +113,7 @@ data SimpleErrorMessage
   | ConstrainedTypeUnified SourceType SourceType
   | OverlappingInstances (Qualified (ProperName 'ClassName)) [SourceType] [Qualified Ident]
   | NoInstanceFound SourceConstraint
-  | AmbiguousTypeVariables SourceType SourceConstraint
+  | AmbiguousTypeVariables SourceType [Int]
   | UnknownClass (Qualified (ProperName 'ClassName))
   | PossiblyInfiniteInstance (Qualified (ProperName 'ClassName)) [SourceType]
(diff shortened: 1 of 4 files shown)


hdgarrood:
Thanks!

PR #3730, opened 2019-09-13
Better reporting of IO errors

hdgarrood:
This commit fixes two issues with the way we read and write files inside
Language.PureScript.Make:

1. We use `doesFileExist` to check for file existence before attempting
to do something with a file. This can cause race conditions; it's better
to just attempt to do the thing and catch the "does not exist" error if
we get one.

2. We discard the IOException which actually tells you what has gone
wrong before we print the error message. This means that the error
message often does not include the most important information. For
example, here's the error I get when I try to build a project on a disk
which does not have enough remaining space to accommodate all of the
compiler output:

Before:

```
Compiling Data.Ord
Compiling Data.DivisionRing
[1/2 CannotWriteFile] (unknown module)

  Unable to write file:

    ./output/Data.DivisionRing/externs.json

[2/2 CannotWriteFile] (unknown module)

  Unable to write file:

    ./output/Data.Ord/foreign.js
```

After:

```
Compiling Data.Ord
Compiling Data.DivisionRing
[1/1 FileIOError] (unknown module)

  I/O error while trying to write JSON file: ./output/cache-db.json

    ./output/cache-db.json: hClose: resource exhausted (No space left on device)
```

We will need to update the documentation repo appropriately, because the
`CannotReadFile`, `CannotWriteFile`, and `CannotGetFileInfo` error codes
have been removed.

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -75,9 +75,7 @@ data SimpleErrorMessage
   | MissingFFIImplementations ModuleName [Ident]
   | UnusedFFIImplementations ModuleName [Ident]
   | InvalidFFIIdentifier ModuleName Text
-  | CannotGetFileInfo FilePath
-  | CannotReadFile FilePath
-  | CannotWriteFile FilePath
+  | FileIOError Text IOError -- ^ A description of what we were trying to do, and the error which occurred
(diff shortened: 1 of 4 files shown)


LiamGoodacre:
Looks great!

hdgarrood:
Thanks :)

PR #3733, opened 2019-09-15
Fix #3722, shows entire row with --verbose-errors flag

### LINKED ISSUE #3722: --verbose-errors should cause entire rows to be shown
For example:

```purescript
module Main where

foo = { a: 1, b: 2, c: 3 }
bar = { a: 1, b: 2, c: "3" }
baz = [foo, bar]
```

When compiling with `--verbose-errors` we get:
```
Error found:
in module Main
at src/Main.purs:5:1 - 5:17 (line 5, column 1 - line 5, column 17)
at src/Main.purs:5:7 - 5:17 (line 5, column 7 - line 5, column 17)

  Could not match type
       
    Int
       
  with type
          
    String
          

  with type String
while trying to match type             
                             ( c :: Int
                             ...       
                             )         
                                       
  with type                
              ( c :: String
              ...          
              )            
                           
while trying to match type t0
  with type { a :: Int   
            , b :: Int   
            , c :: String
            }            
while inferring the type of [ foo
                            , bar
                            ]    
in value declaration baz

where t0 is an unknown type

```

But I think we should show the entire row type if the `--verbose-errors` flag is specified.

Woody88:


src/Language/PureScript/Errors.hs
```diff
@@ -1190,11 +1190,13 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
     printRow f t = markCodeBox $ indent $ f prettyDepth t
 
     -- If both rows are not empty, print them as diffs
-    printRows :: Type a -> Type a -> (Box.Box, Box.Box)
-    printRows r1@RCons{} r2@RCons{} = let
+    -- If verbose print all rows else only print unique rows
+    printRows :: Bool -> Type a -> Type a -> (Box.Box, Box.Box)
+    printRows True r1@RCons{} r2@RCons{} = (printRow typeDiffAsBox r1, printRow typeDiffAsBox r2)
```
hdgarrood: I think this should be `(printRow typeAsBox r1, printRow typeAsBox r2)`; since we are not eliding any labels from the rows in this instance, we don't want to display an ellipsis (which is what `typeDiffAsBox` does).

src/Language/PureScript/Errors.hs
```diff
@@ -1190,11 +1190,13 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
     printRow f t = markCodeBox $ indent $ f prettyDepth t
 
     -- If both rows are not empty, print them as diffs
-    printRows :: Type a -> Type a -> (Box.Box, Box.Box)
-    printRows r1@RCons{} r2@RCons{} = let
+    -- If verbose print all rows else only print unique rows
```
hdgarrood: Since `full` is already in scope here, I think it would be preferable to have `printRows` retain the old signature, i.e. not take a new `Bool` argument, and instead just switch on `full` in the implementation.

src/Language/PureScript/Errors.hs
Woody88: Yes, not sure what I was thinking that day... Or, I probably did not completely understand the function.. Thanks for the clarification!

src/Language/PureScript/Errors.hs
Woody88: I replaced the implementation with a `case` expression, and It's now using the `full` variable that's in the scope.

src/Language/PureScript/Errors.hs
```diff
@@ -1190,11 +1190,17 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
-      in (printRow typeDiffAsBox sorted1, printRow typeDiffAsBox sorted2)
-    printRows r1 r2 = (printRow typeAsBox r1, printRow typeAsBox r2)
+    printRows r1 r2 = case (full, r1, r2) of 
+      (True, _ , _) -> (printRow typeAsBox r1, printRow typeAsBox r2)
+
+      (_, r1'@RCons{}, r2'@RCons{}) -> let 
+        (sorted1, sorted2) = filterRows (rowToList r1') (rowToList r2')
+        in (printRow typeDiffAsBox sorted1, printRow typeDiffAsBox sorted2)
+        
+      (_, _, _) -> (printRow typeAsBox r1, printRow typeAsBox r2)
```
hdgarrood: Sorry to nitpick but a `case` statement of this size should really have the branch expressions (the bit after `->`) on indented new lines after the patterns.

Also, there's no need to rebind `r1` and `r2` to `r1'` and `r2'` here. Why not:

```haskell
(_, RCons{}, RCons{}) ->
  let (sorted1, sorted2) = filterRows (rowToList r1) (rowToList r2)
  in ...
```

Woody88: Fixed. I was thinking the whole time that I did not want to break anything so I just copied and pasted the old implementation without paying attention to that. :disappointed:

hdgarrood: No worries! This is what code review is for :)

PR #3739, opened 2019-10-14
quoted option fixup

milesfrain:


hdgarrood:
This is deliberate, this style of writing quote marks is not uncommon. See e.g. LaTeX and many other programs with CLIs.

```
$ find -asdf
find: unknown predicate `-asdf'
```

milesfrain:
Thanks for clarifying. I assumed the intention was to match markdown format, and was following the format set by an earlier line.
https://github.com/purescript/purescript/blob/c7c922bba7e89cb379c1db7398cdf2aa0b43cb6b/src/Language/PureScript/Interactive/Message.hs#L46

Using matching quotation will help formatters color-code log snippets.
``` sh
# existing unmatched
purs repl: No input files; try running `pulp psci` instead.
Usage: For basic information, try the `--help' option.
```
``` sh
# proposed matched
purs repl: No input files; try running `pulp psci` instead.
Usage: For basic information, try the `--help` option.
```

Not a critical change at all, but if we have the flexibility, it seems best try to make this work well with other systems.

Here's an example of single quotes being used for an option.
```
> ls --asdf
ls: unrecognized option '--asdf'
```

PR #3742, opened 2019-10-20
Fix for ES object shorthand in purs bundle

### LINKED ISSUE #3741: purs bundle misses variable reference in ES6 object shorthand
For example:

```javascript
'use strict';
var foo = "foo";
exports.foo = { foo };
```
will cause the `var foo = "foo"` statement to be removed during bundling, as the bundler is unable to recognise that `foo` is being used in the assigment to `exports.foo`. This is a new issue since updating language-javascript to 0.7, and is thus only reproducible in `master`.

hdgarrood:
Fixes #3741. cc @rhendric, would you mind looking over this?

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -301,6 +301,17 @@ withDeps (Module modulePath fn es) = Module modulePath fn (map expandDeps es)
       -- bound to the module level (i.e., hasn't been shadowed by a function
       -- parameter)
       = ([(m, nm, Internal)], bn)
+    toReference (JSObjectLiteral _ props _) bn
+      = let
+          shorthandNames =
+            filter (`elem` bn) $
+            -- ^ only add a dependency if this name is still in the list of
+            -- names bound to the module level (i.e., hasn't been shadowed by a
+            -- function parameter)
+            mapMaybe unPropertyIdentRef $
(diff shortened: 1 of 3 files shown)


src/Language/PureScript/Bundle.hs
```diff
@@ -308,12 +308,21 @@ withDeps (Module modulePath fn es) = Module modulePath fn (map expandDeps es)
       -- ^ exports.foo means there's a dependency on the public member "foo" of
       -- this module.
       = ([(m, nm, Public)], bn)
+    toReference (JSObjectLiteral _ props _) bn
+      = let
+          shorthandNames = mapMaybe unPropertyIdentRef (trailingCommaList props)
+        in
+          (map (\name -> (m, name, Internal)) shorthandNames, bn)
```
rhendric: I think this should also be a `mapMaybe` conditional on `name` being an element of `bn`; `foo` should not be exported in the below case:

```js
var foo = "foo";
exports.bar = function (foo) { return { foo }; };
```

hdgarrood: Oh yes of course, thanks!

hdgarrood: How does it look now?

hdgarrood:
Thanks both!

PR #3606, opened 2019-04-13
still trying to get windows CI working

hdgarrood:

PR #3745, opened 2019-11-06
Fix error with building haddocks

cdepillabout:
I was having trouble with `stack build --haddock` failing.  

When running `stack build --haddock`, I would get the following error during the haddock phase:

```
src/Language/PureScript/Bundle.hs:300:7: error:
    parse error on input ‘-- ^ only add a dependency if this name is still in the list of names
      -- bound to the module level (i.e., hasn't been shadowed by a function
      -- parameter)’        
    |                       
300 |       -- ^ only add a dependency if this name is still in the list of names
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Here's the full build log:

https://gist.github.com/cdepillabout/ae690a8349d6f448ac7fadf580df6669

--------------------------------------------------------------------------

This PR contains two commits:

- 646cbec corrects those Haskell comments so they don't cause `stack build --haddock` to fail.

- 96fc231 turns on building haddocks for CI, so these types of errors don't creep into the codebase again.

    However, looking back through the history, it appears that building haddocks was explicitly turned off, so I imagine there might be problems with turning this on: https://github.com/purescript/purescript/pull/3606

    I can remove this commit from this PR if needed.

### DIFF
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -297,7 +297,7 @@ withDeps (Module modulePath fn es) = Module modulePath fn (map expandDeps es)
       = ([(mid, nm', Public)], bn)
     toReference (JSIdentifier _ nm) bn
       | nm `elem` bn
-      -- ^ only add a dependency if this name is still in the list of names
+      -- only add a dependency if this name is still in the list of names
       -- bound to the module level (i.e., hasn't been shadowed by a function
       -- parameter)
       = ([(m, nm, Internal)], bn)
@@ -305,7 +305,7 @@ withDeps (Module modulePath fn es) = Module modulePath fn (map expandDeps es)
       = let
           shorthandNames =
             filter (`elem` bn) $
-            -- ^ only add a dependency if this name is still in the list of
+            -- only add a dependency if this name is still in the list of
             -- names bound to the module level (i.e., hasn't been shadowed by a
             -- function parameter)
             mapMaybe unPropertyIdentRef $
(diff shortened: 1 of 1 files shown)


cdepillabout:
The GitHub UI doesn't appear to be showing it, but here is the CI build when just turning on `--haddock`.  This should hopefully fail, because the code doesn't contain the fix from the next commit:

https://travis-ci.org/purescript/purescript/builds/608116478?utm_medium=notification&utm_source=github_status

Here is the CI build that contains both commits above, turning on `--haddock` and fixing the haddock error.  This should hopefully succeed:

https://travis-ci.org/purescript/purescript/builds/608118083?utm_medium=notification&utm_source=github_status

cdepillabout:
I noticed this when trying to get purescript building in nixpkgs:

https://github.com/NixOS/nixpkgs/pull/72882

For now, I've just disabled building the haddocks in https://github.com/NixOS/nixpkgs/pull/72882.

hdgarrood:
It did time out, yes, because unfortunately changing the haddock flags means that you have to recompile the entire dependency tree. The way to fix it is to push an empty commit - that way the next job can pick up where the previous one left off.

hdgarrood:
I turned haddocks off because at the time I only really cared about getting Travis CI working on Windows and it was getting in the way, and potential haddock failures seemed like a rare edge case where checking for them didn't justify the hassle they were causing. I'm happy to add this check back if it doesn't drastically slow CI down.

cdepillabout:
@hdgarrood Thanks for the response.

> The way to fix it is to push an empty commit - that way the next job can pick up where the previous one left off.

I just tried this, and it appears to have fixed the Linux and MacOSX builds, but the windows build is broken for some reason:

https://travis-ci.org/purescript/purescript/jobs/608479280#L4702

The log doesn't seem to indicate why this has happened, but it is occurring directly after some lines about haddocks, so I'm guessing this is the haddock-related error you were running into.

cdepillabout:
@hdgarrood I went ahead and rebased this PR on `master` and removed the commit that turns on building haddocks on CI.  If someone wants to figure out why haddocks can't be built on Windows, I think that would be a good candidate for a future PR.

hdgarrood:
Thanks!

Issue #3743, opened 2019-10-23
Regression 0.13.4: resource exhausted (Too many open files)


emiel:
After upgrading from purs 0.13.3 to 0.13.4 I started seeing the following error when building under spago. On Slack I've seen other are seeing this as well so I thought I'd go ahead and open an issue. From what I can tell the issue is triggered when there are many files in the project. However in the sample project below there are only some 200 *.purs files.

```
purs: getCurrentDirectory:getWorkingDirectory: resource exhausted (Too many open files)
spago: Failed to build.
```

```
{-
Welcome to a Spago project!
You can edit this file as you like.
-}
{ name =
    "my-project"
, dependencies =
    [ "arrays"
    , "canvas"
    , "console"
    , "lists"
    , "psci-support"
    , "random"
    , "refs"
    , "web-dom"
    ]
, packages =
    ./packages.dhall
, sources =
    [ "src/**/*.purs", "test/**/*.purs" ]
}
```

```
$ spago version
0.10.0.0

$ purs --version
0.13.4

$ spago -v build
Running `spago build`
Transformed config is the same as the read one, not overwriting it
Ensuring that the package set is frozen
Getting transitive deps
Running `fetchPackages`
Checking if `purs` is up to date
Running `getGlobalCacheDir`
Installation complete.
Compiling with "purs"
[...]

PR #3747, opened 2019-11-09
Banish lazy I/O from Make

### LINKED ISSUE #3371: See if we can use Aesons new encodeFile/decodeFile
These were added in 1.13.x and I know we have some custom code that does this instead.

hdgarrood:
I did this hoping that it would help with #3743, as we aren't carrying
so many lazy bytestrings around this way. The bug still appears to exist
after this change, but I think it's a step in the right direction
regardless.


hdgarrood:
Renamed the PR as I've taken this a bit further. I've also removed a slightly dodgy usage of `Data.ByteString.Lazy.Char8`. This does now seem to fix #3743: I could reliably reproduce #3743 on the `purescript-halogen-realworld` app, and with this branch it compiles successfully.

hdgarrood:
It may be worth coming back to this later to make memory usage more efficient, as at the moment with this PR we will read entire files into memory before doing anything with them. However, the compiler has been doing that for source files for quite a while already and nobody has complained yet, It's also worth noting that we already pull in `conduit` transitively, probably via `warp` (the remote repl needs an http server), so that would probably be a good next step if we do want to come back to this later. But for now I would like to try to get this merged and released, as I think #3743 is a fairly critical bug with 0.13.4.

elliotdavies:
@hdgarrood This appears to fix the issue for us! Thanks for looking into it :pray:

PR #3748, opened 2019-11-13
Consider re-exports in unused kind warnings

### LINKED ISSUE #3744: Unused kind import warning does not consider re-exports
If a kind is imported and not used within the module except for being re-exported, it is incorrectly marked as unused and a warning is emitted. See e.g. https://github.com/purescript/purescript-st/pull/23 in which I accidentally release a breaking change as a patch-level change because this warning was incorrectly emitted.

hdgarrood:
Fixes #3744. This allows `purescript-st` to compile without warnings again (see https://travis-ci.org/purescript/purescript-st/jobs/606644430#L956 for an example of how it is now)


hdgarrood:
Thanks!

PR #3751, opened 2019-11-13
Use PureScript escapes in string pretty-printing

### LINKED ISSUE #3740: Expression pretty-printing generates invalid String escapes
```
> "🐱" "🐱"
Error found:
in module $PSCI
at :1:1 - 1:8 (line 1, column 1 - line 1, column 8)

  Could not match type
          
    String
          
  with type
                
    String -> t0
                

while applying a function "\128049"
  of type String
  to argument "\128049"
while inferring the type of "\128049" "\128049"
in value declaration it

where t0 is an unknown type

> "\128049"
Illegal character escape code at line 1, column 3
```

The error message should not generate invalid string escapes like `"\128049"`; this is Haskell syntax. It should generate a valid string escape like `"\x1f431"`, or perhaps even better, just use exactly what was in the original source code (although I'm not sure if this information is still available at this point).

hdgarrood:
Fixes #3740

As of this PR, pretty-printed Strings will try use the literal code
points when pretty-printing strings where sensible, including for
non-ASCII strings. For example, "Fuß", "¡Hola!", and "🐱" now appear as
such in error messages (whereas previously they would have used
escapes).

### DIFF
--- a/src/Language/PureScript/PSString.hs
+++ b/src/Language/PureScript/PSString.hs
@@ -18,7 +18,7 @@ import GHC.Generics (Generic)
 import Control.DeepSeq (NFData)
 import Control.Exception (try, evaluate)
 import Control.Applicative ((<|>))
-import Data.Char (chr)
+import qualified Data.Char as Char
 import Data.Bits (shiftR)
 import Data.List (unfoldr)
 import Data.Scientific (toBoundedInteger)
@@ -67,7 +67,7 @@ instance Show PSString where
 -- we do not export it.
 --
 codePoints :: PSString -> String
-codePoints = map (either (chr . fromIntegral) id) . decodeStringEither
+codePoints = map (either (Char.chr . fromIntegral) id) . decodeStringEither
 
 -- |
 -- Decode a PSString as UTF-16 text. Lone surrogates will be replaced with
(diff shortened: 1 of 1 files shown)


src/Language/PureScript/PSString.hs
```diff
@@ -163,8 +201,8 @@ prettyPrintStringJS :: PSString -> Text
 prettyPrintStringJS s = "\"" <> foldMap encodeChar (toUTF16CodeUnits s) <> "\""
   where
   encodeChar :: Word16 -> Text
-  encodeChar c | c > 0xFF = "\\u" <> hex 4 c
-  encodeChar c | c > 0x7E || c < 0x20 = "\\x" <> hex 2 c
+  encodeChar c | c > 0xFF = "\\u" <> showHex' 4 c
```
natefaubion: Should this have `{}` delimiters?

hdgarrood: Oh, good spot, yes! This won’t handle anything above U+FFFF correctly as it is now. I think I should add a test for this.

hdgarrood: Oh wait no - this is okay as-is, because `c` is a UTF-16 code unit (Word16) here, i.e. it can't be any larger than 0xFFFF. So the `4` in `showHex' 4` is ensuring we have at least 4 hex digits, and the fact that `c` is a `Word16` ensures that we have at most 4 hex digits. This function will render astral plane characters as surrogate pairs, e.g. U+1F431 -> `"\uD38D\uDC31"`

hdgarrood: `{}` delimiters probably wouldn't hurt though.

PR #3758, opened 2019-11-27
Added spans to ado desugaring

### LINKED ISSUE #3754: Not importing `map` with `ado` results in an error without source spans
This is due to the desugaring not adding spans like it does with `do`.

dariooddenino:
This should fix  #3754.

I'm not sure if the `nullSourceSpan` passed to `VarBinder` and `Case` in the `go` function needs to be addressed.


natefaubion:
> I'm not sure if the nullSourceSpan passed to VarBinder and Case in the go function needs to be addressed.

I personally think that everything we generate or desugar should map to a user source span. The do-desugaring threads the `pos` around here.

hdgarrood:
Thanks @dariooddenino!

PR #3762, opened 2019-12-05
Added contravariant dependency to package.yaml

coot:
This fixed an error when building with `cabal-3.0`.


coot:
I realised that `purescript` is now building with `ghc-8.6.5`, so no need for this patch.

PR #3768, opened 2020-01-10
optimize import desugaring for full builds

colinwahl:
TLDR: This PR introduces an optimization to the import desugaring phase which eliminates duplicated work, and leads to drastically lower build times when you have a large amount of modules.

@natefaubion noticed that clean builds at Awake were starting to take quite a while, so I went and profiled one of them.  I ran the results through [profiteur](https://github.com/jaspervdj/profiteur) and noticed that about 41% of the time (and memory) spent on our build was from the function `externsEnv` within `desugarImportsWithEnv`.

Screen Shot 2020-01-10 at 12 41 35 AM

Some spelunking revealed that `externsEnv` was called once per (transitive) dependency of each module that was being compiled.  This resulted in quite a bit of repeated work.  For example, if you have modules A, B, C, and D, where the dependency graph looks like `A -> B -> C -> D`, then you call externsEnv on the following lists of modules: `[A]`, `[A, B]`, `[A, B, C]`.  Another example is that if every module depends on `Prelude`, then you are computing `externsEnv` on `Prelude` once per module.

`externsEnv` can be decently expensive, and it is easy to imagine that this duplicated work could get out of hand with more modules (we have definitely passed that threshold at Awake, I think we have ~~1300+~~ 2000+ modules).

I realized that we could use the fact that we are already compiling in topological order along with the fact that `Env`s will be disjoint for "non-transitively-dependent" pairs of modules to instead keep a running `Env` which is updated as externs files become available, and which never computes `externsEnv` on an extern file more than once.  In doing this, a tremendous amount of redundant computation is avoided, resulting in a much "healthier" looking profile:

Screen Shot 2020-01-10 at 1 07 40 AM
Screen Shot 2020-01-10 at 1 07 49 AM

Where calls to `externsEnv` account for 0.2% of the time of a full build, and parsing (~50%), typechecking (~26%) and codegen (~15%) are the dominating items of the profile

In practice, on the Awake codebase, this reduces the time needed for a full build from ~2ms20s to ~1m20s, about a 43% speedup.

The bulk of this change is quite manual and not very invasive - the important bits are in the changes to `buildModule`, which will need some eyes on it :)  The main change to extends the coordination done in `make` to collect dependencies, and then ensure all dependency's externs are added to the `Env` in the correct topological order.  Since the coordination to collect dependencies was already done, this ended up being a pretty easy modification.

Notes:
* I have only applied the optimization when compiling through `make`, all other workflows such as psci, purs-ide, etc should not be affected.
* I see that a decent amount of time (~6%) is spend computing `applyExternsFileToEnvironment` within `rebuildModule`.  This may be able to benefit from the same sort of optimization.


stack.yaml.lock
```diff
@@ -0,0 +1,47 @@
+# This file was autogenerated by Stack.
```
kritzcreek: We should probably make it a conscious decision whether to check in this lockfile (in a seperate PR), rather than it happening "by accident".

hdgarrood: I'd prefer to add it to gitignore honestly. I'm pretty sure we've never come up against a situation where it would have been useful to have.

src/Language/PureScript/Interactive/Types.hs
```diff
@@ -118,17 +119,22 @@ psciImportedModuleNames st =
...
-  desugarModule = runExceptT =<< hushWarnings . P.desugarImportsWithEnv (map snd externs)
-  hushWarnings  = fmap fst . runWriterT
+  desugarModule :: P.Env -> [P.Module] -> Either P.MultipleErrors (P.Env, [P.Module])
+  desugarModule e = runExceptT =<< hushModuleWarnings . P.desugarImportsWithEnv e
+
+  createEnv :: [P.ExternsFile] -> Either P.MultipleErrors P.Env
+  createEnv = runExceptT =<< hushEnvWarnings . foldM P.externsEnv P.primEnv
+
+  hushModuleWarnings  = fmap fst . runWriterT
+  hushEnvWarnings  = fmap fst . runWriterT
```
natefaubion: Maybe just inline these? They are identical except for types I think.

src/Language/PureScript/Make.hs
```diff
@@ -57,13 +58,27 @@ rebuildModule
   -> [ExternsFile]
   -> Module
   -> m ExternsFile
-rebuildModule MakeActions{..} externs m@(Module _ _ moduleName _ _) = do
+rebuildModule actions externs m = do
+  let
+    silence = fmap fst . runWriterT
```
natefaubion: Same here.

stack.yaml.lock
colinwahl: Sorry, didn't mean to commit this in the first place :) I've removed it

garyb:
:tada:

colinwahl:
Awesome! Thanks all for the reviews 👍

PR #3773, opened 2020-01-22
added support for use a native backend REPL

freylax:
I added a native backend to the REPL command. It works like the node backend but uses a configurable call to make (or other build environment) to build an evaluation executable.
Example is here [purescript-native-repl](github.com/freylax/purescript-native-repl)
Could also use an plugin to inspect an runnig application like here [purescript-native-cpp-so](github.com/freylax/purescript-native-cpp-so).

### DIFF
--- a/app/Command/REPL.hs
+++ b/app/Command/REPL.hs
@@ -38,6 +38,7 @@ import qualified Language.PureScript as P
 import qualified Language.PureScript.Bundle as Bundle
 import qualified Language.PureScript.CST as CST
 import           Language.PureScript.Interactive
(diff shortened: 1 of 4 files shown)


hdgarrood:
Thanks for the PR but I think the API being proposed here really needs a) a proper proposal with details on how the API works and how a new alternative backend would use it, and b) buy-in from other alternative backend maintainers, as I think we could easily end up with a feature which only works for one or two backends here if we’re not careful. I’m going to close this for now; could you please open a thread on Discourse to take the above points further?

PR #3774, opened 2020-01-22
Golden tests for `purs/failing` and `purs/warning`

dariooddenino:


tests/TestCompiler.hs
```diff
@@ -6,133 +6,80 @@
...
-      let mainPath = getTestMain testPurs
-      expectedFailures <- runIO $ getShouldFailWith mainPath
-      it ("'" <> takeFileName mainPath <> "' should fail with '" <> intercalate "', '" expectedFailures <> "'") $
-        assertDoesNotCompile supportModules supportExterns supportForeigns testPurs expectedFailures
-
-  where
-
-  -- Takes the test entry point from a group of purs files - this is determined
-  -- by the file with the shortest path name, as everything but the main file
-  -- will be under a subdirectory.
```
hdgarrood: Can you preserve this comment please?

tests/TestCompiler.hs
```diff
@@ -161,48 +108,39 @@ assertCompiles supportModules supportExterns supportForeigns inputFiles outputFi
...
+assertGolden supportModules supportExterns supportForeigns testCases =
+  flip fmap testCases $ \files ->
+    let mainFile = head files
+    in
+      goldenVsString
+      (takeBaseName mainFile)
+      (replaceExtension mainFile ".out")
+      (BS.fromStrict . T.encodeUtf8 . T.pack <$> printErrorOrWarning supportModules supportExterns supportForeigns files)
+
+printErrorOrWarning
```
hdgarrood: We shouldn't really be smushing errors and warnings together like this. With the current code, these tests are verifying that everything in the `failing` directory produces at least one error, and that everything in the `warning` directory produces only warnings; if I'm reading this correctly, we're losing that here.

tests/TestCompiler.hs
```diff
@@ -6,133 +6,80 @@
...
-checkShouldFailWith :: [String] -> P.MultipleErrors -> Maybe String
-checkShouldFailWith expected errs =
-  let actual = map P.errorCode $ P.runMultipleErrors errs
-  in if sort expected == sort (map T.unpack actual)
-    then checkPositioned errs
-    else Just $ "Expected these errors: " ++ show expected ++ ", but got these: "
-      ++ show actual ++ ", full error messages: \n"
-      ++ unlines (map (P.renderBox . P.prettyPrintSingleError P.defaultPPEOptions) (P.runMultipleErrors errs))
-
-checkPositioned :: P.MultipleErrors -> Maybe String
```
hdgarrood: I would like to keep this check in, just in case manual review fails to spot that an error is missing a source span.

tests/TestCompiler.hs
```diff
@@ -6,133 +6,80 @@
...
--- one TypesDoNotUnify error.
---
--- If a module is expected to produce multiple type errors, then use multiple
--- @shouldFailWith lines; for example:
---
---   -- @shouldFailWith TypesDoNotUnify
---   -- @shouldFailWith TypesDoNotUnify
---   -- @shouldFailWith TransitiveExportError
+-- Failing tests in the `failing` and `warning` folders check their output
+-- agains the relative golden files (`.out`).
```
hdgarrood: typo here?

hdgarrood:
On second thoughts I think I'd prefer to go with what @natefaubion suggested on Discourse and keep the `shouldFailWith` stuff in here, just in case. The code probably isn't doing much harm. (Also, as an aside, if we were going to remove the code for handling these special comments, then we should also remove all of the special comments from the test input files too.)

tests/TestCompiler.hs
dariooddenino: What do you mean with smushing errors and warnings together? I'll check the code again later to see if I get what you mean :)

dariooddenino:
You mean that you would like to have both the golden tests and the directives check? I thought @natefaubion meant that he just wanted the directives on the test files, not the tests.

hdgarrood:
I’m pretty sure the idea was to keep in both the directives themselves and the code which tests them; they aren’t very useful if we aren’t checking them.

tests/TestCompiler.hs
hdgarrood: What I mean is that this function discards the information of whether a particular MultipleErrors was raised as an error or as a warning.

dariooddenino: Mm I see, I thought it wasn't that important since we were checking the error/warning output and I assumed it was different in the two cases. 
Isn't it going to be checked by the directives test once I put them back in anyways?

hdgarrood: Oh of course, yes it would be.

hdgarrood: By the way part of the reason I think it’s really important to test this properly is because the exit code depends on it too,  and most workflows (and especially CI) rely on the exit codes to be correct.

garyb:
I'm a fan of keeping the `shouldFailWith` etc. because otherwise these tests are only useful for preventing regressions. When working on an error or warning I tend to work in a TDDish way, in that I make a bunch of cases that should fail first and then work on making them fail the correct way after.

dariooddenino:
Directives are back in :)

natefaubion:
Do we want to omit ANSI color sequences in the `.out` file?

hdgarrood:
I don't have any strong feelings on this - I guess would make the test output a bit nicer to look at if we did omit them, but also the colours might be worth testing as they do help distinguish names in your code from error message text, as seen in eg https://github.com/purescript/purescript/issues/2056

PR #3776, opened 2020-01-25
[purs ide] Removes the file watcher

kritzcreek:
The file watcher has only been giving people trouble recently and I'd rather just have them use the behaviour enabled by `--editor-mode`. This means the server will only update its knowledge about the module graph when an editor asks for a rebuild of a given module.

The means out-of-band compilations are missed (for example when installing a new dependency and running a compilation command on the command line), but the file watcher was too flaky to properly capture these anyway. I'd rather figure out a mechanism (maybe with a lockfile?) to tell the IDE server that its knowledge is out of date and cause it to fully reload its state, than trying to make the watching of individual externs work.


natefaubion:
> I'd rather figure out a mechanism (maybe with a lockfile?) to tell the IDE server that its knowledge is out of date and cause it to fully reload its state, than trying to make the watching of individual externs work.

Maybe this can be done by watching cache-db.json and diffing it for changes to see what needs to reload?

hdgarrood:
Even if watching cache-db.json seems like it ought to work (I think it would, as it is rewritten each time the hash of any input source file changes), I do like the idea of removing the mode entirely. However, I’m a little reticent about doing it without warning. How would you feel about keeping the mode in for now, but making editor mode default on all platforms, and emitting a warning on startup if watching mode is explicitly requested, which says that watcher mode is deprecated and will be removed in a subsequent release?

kritzcreek:
@hdgarrood That's exactly what this PR is doing ;)

hdgarrood:
Wait, sorry - my impression was that this PR removes the file watching capability and instead always has the ide server run in editor mode, regardless of flags passed on the command line; passing flags requesting file watching will not prevent the server from running, but those flags will be ignored. Have I misunderstood?

kritzcreek:
Oh no you haven't, it's late over here and I misunderstood. That would mean I'd ignore the old flags and add a new flag that is immediately deprecated again to turn on the file watcher? Seems a little weird but not unreasonable.

My main issue with that is that there's basically no good UI for an editor plugin to expose these while working for multiple compiler versions. They'd need to case on the compiler version and ignore one half of the users settings...

hdgarrood:
Right yeah, I see. I guess I would just like to make sure we have some answer for if people ask "what should I do if I want a file watcher"? Suggest they use pscid, and set up their editor to talk to the ide server which was launched by pscid?

natefaubion:
On the subject of `cache-db.json`:
* Does the IDE code update this file? If not, I really think it should.
* Would watching this single file serve as a replacement for watching all externs? I suspect that the reliability of watching is in part because of the frequency and volume of things to watch (but I may be off-base here). If we want to watch some sort of lock (and `cache-db.json` in particularly), it seems like we might not need to deprecate anything?

hdgarrood:
> Does the IDE code update this file? If not, I really think it should.

Oh of course yes, you're right, I don't think it does. This is an oversight on my part, sorry. That might explain the trouble if it's only started recently (the compiler started using cache-db.json files in 0.13.4, released in Oct last year: https://github.com/purescript/purescript/releases/tag/v0.13.4)

kritzcreek:
> Right yeah, I see. I guess I would just like to make sure we have some answer for if people ask "what should I do if I want a file watcher"? Suggest they use pscid, and set up their editor to talk to the ide server which was launched by pscid?

`pscid` watches the source files for changes and triggers the rebuild. It doesn't watch the `output/` folder for changes, so it doesn't recognize these out of band collections either.

> Does the IDE code update this file? If not, I really think it should.

It does not and when I started thinking about it yesterday I think there are scenarios where this could be real trouble. If a user:
1. Compiles their project against state 1 (S1) of file A.purs
2. Changes A.purs into S2 in their editor and causes a rebuild
3. Never saves that file and simply reverts its contents to S1

Now the current externs.json in the output corresponds to S2, but the file is in S1 and the cachedb file claims that it already has an up-to-date version for that timestamp and file content, so you can't make the compiler actually give you the right output.

> Would watching this single file serve as a replacement for watching all externs? I suspect that the reliability of watching is in part because of the frequency and volume of things to watch (but I may be off-base here). If we want to watch some sort of lock (and cache-db.json in particularly), it seems like we might not need to deprecate anything?

I've stopped trusting file watchers... maybe we could check the timestamp of `cache-db.json` on every user command and trigger debounced refreshes of the state. For certain commands (like rebuilds) we'd probably have it synchronously reload the externs before continuing, but for things like completion or type at point we could do it on a background thread.

natefaubion:
> I've stopped trusting file watchers... maybe we could check the timestamp of cache-db.json on every user command and trigger debounced refreshes of the state.

I think this sounds list a reasonable alternative.

kritzcreek:
superseeded by #3799

PR #3777, opened 2020-01-26
Add versions bounds

coot:
The version bounds are too relaxed.  Cabal needs to be constraint with
`<3.0` and haskeline with `<0.8` version bounds.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -48,7 +48,7 @@ dependencies:
   - bower-json >=1.0.0.1 && <1.1
   - boxes >=0.1.4 && <0.2.0
   - bytestring
-  - Cabal >= 2.2
+  - Cabal >= 2.2 && <3.0
   - cheapskate >=0.1 && <0.2
   - clock
   - containers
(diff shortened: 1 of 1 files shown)


hdgarrood:
The idea is that we use stack for local development and `pvp-bounds: upper` for publishing to hackage so that releases are more likely to remain installable over time - so this change would actually relax the upper bound we have on Cabal for Hackage releases. However if there’s no reason to suspect this might cause problems I’m good with merging this.

coot:
I see, this PR indeed relaxes `Cabal`: on hackage its `>=2.2 && <2.5`, haskeline has the same bounds. I am also fine for keeping `<2.5`, or even not doing anything as I can put staff in `cabal.project.local`, but it would be nice if one could build it be default with `cabal` too.

How do you push different bounds to hackage? I guess you generate a different `purescript.cabal` file from `package.yaml`, right?

hdgarrood:
No, it’s a stack feature - all you need to do is specify `pvp-bounds` in `stack.yaml` and then publishing using `stack upload .`

hdgarrood:
If you’re reasonably confident purescript works with Cabal up to 3.0 I’m happy to relax this bound though, just to be clear. I just thought the additional context might be useful to know.

hdgarrood:
See https://docs.haskellstack.org/en/stable/yaml_configuration/#pvp-bounds

PR #3789, opened 2020-02-13
[purs ide] Updates the cache-db.json file on rebuilds

kritzcreek:
This patch makes it so we update the content hashes when rebuilds through the IDE occur. We've still got some unfortunate invalidation going on when recompiling a file with no changes, but at least this way the IDE can never cause `purs compile` to invalidate _too little_.

One non-local change is that we now do some path normalisation for the cache-db entries, so two invocations like `purs compile "src/*.purs"` and `purs compile "./src/*.purs"` don't cause the compiler to invalidate every module.

I plan to do the file watcher replacement in a follow-up PR.


src/Language/PureScript/Make/Monad.hs
```diff
@@ -64,34 +64,34 @@ runMake opts = runLogger' . runExceptT . flip runReaderT opts . unMake
 -- | Run an 'IO' action in the 'Make' monad. The 'String' argument should
 -- describe what we were trying to do; it is used for rendering errors in the
 -- case that an IOException is thrown.
-makeIO :: Text -> IO a -> Make a
+makeIO :: (MonadIO m, MonadError MultipleErrors m) => Text -> IO a -> m a
```
hdgarrood: Are we using these at types other than `Make`? I'm not necessarily opposed to generalising them, but I think the doc-comments could probably do with updating if we are going to do it.

src/Language/PureScript/Make/Cache.hs
```diff
@@ -123,8 +127,26 @@ checkChanged cacheDb mn currentInfo = do
             pure (Map.singleton fp (newTimestamp, newHash), All (dbHash == newHash))
 
   pure (CacheInfo newInfo, getAll isUpToDate)
+  where
```
hdgarrood: Is this `where` redundant?

src/Language/PureScript/Ide/Rebuild.hs
```diff
@@ -55,35 +57,72 @@ rebuildFile file actualFile codegenTargets runOpenBuild = do
...
         void populateVolatileState
       runOpenBuild (rebuildModuleOpen makeEnv externs m)
       pure (RebuildSuccess warnings)
 
+-- | When adjusting the cache db file after a rebuild we always pick a
+-- non-sensical timestamp ("1858-11-17T00:00:00Z"), and rely on the
+-- content hash to tell whether the module needs rebuilding. This is
+-- because IDE rebuilds may be triggered on temporary files to not
+-- force editors to save the actual source file to get at diagnostics
+dayZero :: Time.UTCTime
```
hdgarrood: Can you expand on the doc-comment here a bit? I'm not sure I entirely follow. If an editor is saving a temporary file for a particular module, wouldn't its path be different, thereby bamboozling the cache invalidation anyway?

hdgarrood: Oh I see, you're keeping track of the file which normally contains that module and also sometimes a separate path for a temporary file.

hdgarrood:
I just want to check again that I've understood the issues you're solving here properly, does the following sound correct?

- firstly, we don't normalise file paths in the cache db, which means that we end up with IDE rebuilds invalidating everything below the rebuilt module even if there are no changes, because e.g. "./path/to/module.purs" is considered different to "path/to/module.purs", and if you're building things both with `purs compile` directly and with the ide you can easily end up with a mixture 
- secondly, we don't update the cache-db.json file at all when the ide rebuilds a particular module, which means that the source file's timestamp has changed on disk but the cache db still contains the old timestamp, which causes the cache to be invalidated for that file, causing it and everything below it to be rebuilt on the next full build, even when there are no changes

Is that it, or is there more to it than this?

src/Language/PureScript/Make/Monad.hs
kritzcreek: Yes the IDE is using the functions to hash files and write/read the cachedb from here.

src/Language/PureScript/Make/Cache.hs
kritzcreek: Sure looks like it...
```suggestion
```

src/Language/PureScript/Ide/Rebuild.hs
kritzcreek: Yeah. This would go away if we implemented the LSP as than we'd have to hold a Virtual File System in memory that contained all the non-saved changes to files. The downside is that you end up with a stateful protocol.

kritzcreek:
> firstly, we don't normalise file paths in the cache db, which means that we end up with IDE rebuilds invalidating everything below the rebuilt module even if there are no changes, because e.g. "./path/to/module.purs" is considered different to "path/to/module.purs", and if you're building things both with purs compile directly and with the ide you can easily end up with a mixture

Yes, in addition to that mixing absolute with relative filepaths has the same effect.

> secondly, we don't update the cache-db.json file at all when the ide rebuilds a particular module, which means that the source file's timestamp has changed on disk but the cache db still contains the old timestamp, which causes the cache to be invalidated for that file, causing it and everything below it to be rebuilt on the next full build, even when there are no changes

That's the harmless case. With the current situation you can imagine the following sequence:

1. The user makes an edit without saving
2. The IDE rebuilds against a temp file and writes out a changed externs file
3. The user discards the unchanged save without triggering a rebuild
4. The output/ directory now contains an externs file that does not correspond to the .purs file on disc, but can't be invalidated by a normal `purs compile` because the cache says the timestamps and content hashes match

hdgarrood:
Ah yes, thanks for reminding me.

kritzcreek:
Thanks for taking a look!

wclr:
@kritzcreek 

- The user makes an edit without saving
- The IDE rebuilds against a temp file and writes out a changed externs file

Why should IDE write out a change externs file if it doesn't produce the codegen output (codegeTargets is empty)? It seems that if no output is produced and externs/cache-db is not touched this couldn't break anything for later `purs compile`. And this would allow having absolutely pure diagnostics for non-saved files in the editors.

PR #3791, opened 2020-02-15
ES modules

kl0tl:
This PR compiles to ES modules instead of CommonJS.

As summarised in https://github.com/purescript/purescript/issues/3613:

+ Imports are compiled to namespace imports.
+ Exported identifiers are compiled to named exports.
+ Primes are always escaped in exports.
+ Both ES and CommonJS exports are supported for foreign modules.

For instance, given this foreign module:

```js
const message$prime = "Hello world";
export { message$prime };
```

the following PureScript module:

```purs
module Main where

import Prelude

import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = log "Hello world"

foreign import message' :: String

main' :: Effect Unit
main' = log message'
```

compiles to:

```js
// Generated by purs version 0.13.6
import * as $foreign from "./foreign.js";
import * as Effect_Console from "../Effect.Console/index.js";
var main$prime = Effect_Console.log($foreign["message$prime"]);
var main = Effect_Console.log("Hello world");
export {
    message$prime
} from "./foreign.js";
export {
    main,
    main$prime
};
```

and can then be bundled with `purs bundle`.

According to the [documentation](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_package_json_type_field) a package.json containing `{"type":"module"}` should be enough for Node.js v13+ to load .js files as ES modules but that did not work so I added a dependency to [`esm`](https://www.npmjs.com/package/esm) to make tests pass. Did anyone had success with that? Otherwise the v14 is scheduled for 2020-04-21, I’ll try again when it lands.


kl0tl:
Note that since primes in foreign modules exports are now required to be escaped, this isn’t usable before https://github.com/purescript/purescript/issues/3613#issuecomment-583794497 and enough time for the ecosystem to adapt.

mhmdanas:
Isn't the `$` after `Just` unnecesarry [right here](https://github.com/purescript/purescript/pull/3791/files#diff-dbae7a11e00a75c43d9c3adb0f338d77R187)?

kl0tl:
That `$` was indeed redundant!

I’ve rebased the PR on top of https://github.com/purescript/purescript/pull/3792 and dropped support for escaped primes in identifiers exported from foreign modules.

hdgarrood:
@kl0tl thanks very much for tackling this, I'm really excited to give it a try! Just to set expectations, I think this is unlikely to get looked at least until v0.14.0 is out. After that, we might want to wait a while before merging this, to give people time to heed the warnings about primes in FFI identifiers, and also because we probably don't want to put breaking releases out too often. We can assess the effect this will have on the ecosystem, though; if it's not too great, that makes it easier to justify releasing ES modules more quickly after v0.14.0. I suspect ES modules won't affect libraries very much, but the migration might be more difficult for some apps, especially if they have complicated bundling setups.

kl0tl:
Regarding the ecosystem: I’ve submitted pull requests to all impacted libraries in the official package set, is there anything more I could do to help with the transition?

hdgarrood:
Thanks! I can’t think of anything really, apart from trying to get lots of people with different setups to try this branch out on their apps and report back on how much effort was required to get them working, so that we can better see how much of an impact this will have on the ecosystem, and see if there’s anything else we can do to make the transition smoother. That should wait until after v0.14.0 is out too though, since people will need to update their apps to v0.14.0 first.

kl0tl:
I successfully bundled one of our projects at work with this branch and webpack@4! The resulting bundle weights 18% less and we’ll scrape a few more percents with webpack@5 and `sideEffects: false` :tada:

I have to upstream a patch to purs-loader but I hadn’t any change to make to the webpack configuration otherwise, beside what was required for testing webpack@5 and `sideEffects: false`.

I expect bundling the output directory directly with webpack (without purs-loader), rollup or parcel to work as is. I took a look at rollup-plugin-purs but I don’t think it needs anything to work with ES modules. Also I believe this makes `zephyr` unnecessary for webpack@5 and rollup (based on https://github.com/kl0tl/purescript-es-modules-benchmark) but I’d be interested to try with a real app.

kl0tl:
I was curious about how zephyr would perform on the same project so I did more tests today and bundling CJS modules optimized by zephyr still beats webpack@5 over ES modules :sweat_smile:

Here’s a summary of my results with webpack@4:

CJS modules | ES modules | zephyr
--- | --- | ---
502.16 KB | 409.36 KB | 379.81 KB

and webpack@5:

CJS modules | ES modules | zephyr
--- | -- | ---
483.58 KB | 394.74 KB (388.17 KB with `sideEffects: false`) | 368.26 KB

I’ll compile zephyr against this branch to see how bundling optimized ES modules compares.

kl0tl:
I picked https://github.com/f-f/purescript-react-basic-todomvc as a simple enough but hopefully not too trivial project to test various bundlers configurations with ES modules.

I have successfully tested (with zephyr when possible) so far:

+ [parcel@2](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/parcel2-es-modules)
+ [webpack@5](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/webpack5-es-modules)
+ [webpack@5 with purs bundle](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/webpack5-purs-bundle-es-modules)
+ [webpack@5 with purs-loader](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/webpack5-purs-loader-es-modules)
+ [rollup](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/rollup-es-modules)
+ [rollup with rollup-plugin-purs](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/rollup-plugin-purs-es-modules)

afcondon:
I'd love to try this on our project and give feedback if any arises but when i tried cloning and checking out `es-modules` branch from the @kl0tl 's repo, but it doesn't build for me on macOS Catalina (the master branch, 0.13.6, does) 

```purescript> /Users/mambo/work/purescript/purs-es-modules/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/Language/PureScript/CST/Parser.hs:149:15: error:
purescript>     Not in scope: type constructor or class ‘Kind’
purescript>     |
purescript> 149 | happyIn42 :: (Kind ()) -> (HappyAbsSyn t109)
purescript>     |               ^^^^```

any ideas?

**Comment by kl0tl** [comment]
I updated the browser and Node.js backends of the REPL to work with ES modules.

@afcondon I’m sorry but I cannot reproduce your issue 😕How are you building the repo?

**Comment by thomashoneyman** [comment]
> I have successfully tested (with zephyr when possible) so far:

I am assuming this poses no risk, but just wanted to mention the other tool I see mentioned often:

* https://github.com/google/closure-compiler
* https://github.com/webpack-contrib/closure-webpack-plugin

**Comment by afcondon** [comment]
> I updated the browser and Node.js backends of the REPL to work with ES modules.
> 
> @afcondon I’m sorry but I cannot reproduce your issue 😕How are you building the repo?

i cloned it, did `stack update`, `stack build` then realized i hadn't switched to `es-modules` branch so i did that and `stack update`, `stack build` again and got errors.

here's the full error now, just in case it helps, looks like it's maybe picking up some wrong version of the Parser...

purescript> Preprocessing library for purescript-0.13.6..
purescript> Building library for purescript-0.13.6..
purescript>
purescript> <no location info>: warning: [-Wmissing-home-modules]
purescript>     These modules are needed for compilation but not listed in your .cabal file's other-modules:
purescript>         Language.PureScript.CST.Parser
purescript> [  2 of 131] Compiling Language.PureScript.CST.Parser
purescript>
purescript> /Users/mambo/work/purescript/purs-es-modules/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/Language/PureScript/CST/Parser.hs:149:15: error:
purescript>     Not in scope: type constructor or class ‘Kind’
purescript>     |
purescript> 149 | happyIn42 :: (Kind ()) -> (HappyAbsSyn t109)
purescript>     |               ^^^^
purescript>
purescript> /Users/mambo/work/purescript/purs-es-modules/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/Language/PureScript/CST/Parser.hs:152:38: error:
purescript>     Not in scope: type constructor or class ‘Kind’
purescript>     |
purescript> 152 | happyOut42 :: (HappyAbsSyn t109) -> (Kind ())
purescript>     |                                      ^^^^
purescript>
purescript> /Users/mambo/work/purescript/purs-es-modules/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/Language/PureScript/CST/Parser.hs:155:15: error:
purescript>     Not in scope: type constructor or class ‘Kind’
purescript>     |
purescript> 155 | happyIn43 :: (Kind ()) -> (HappyAbsSyn t109)
purescript>     |               ^^^^
purescript>
purescript> /Users/mambo/work/purescript/purs-es-modules/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/Language/PureScript/CST/Parser.hs:158:38: error:
purescript>     Not in scope: type constructor or class ‘Kind’
purescript>     |
purescript> 158 | happyOut43 :: (HappyAbsSyn t109) -> (Kind ())
purescript>     |                                      ^^^^
purescript>

**Comment by kritzcreek** [comment]
I know it's silly, but you can try to `rm -rf` the `.stack-work` directories in `purescript`, `purescript/lib/purescript-cst`, and `purescript/lib/purescript-ast` and then rerun `stack build` for an hour... I think that helped the last time I had a similar thing. (Maybe try with just removing the `cst` one as that seems to be where the problem is currently)

**Comment by afcondon** [comment]
Thanks, @kritzcreek, i'll give that a go

**Comment by kritzcreek** [comment]
@afcondon Oh, and maybe make sure you're running an up-to-date version of stack. I'm running:
```
~\code\purescript [master]> stack --version
Version 2.1.3, Git revision 0fa51b9925decd937e4a993ad90cb686f88fa282 (7739 commits) x86_64 hpack-0.31.2
```

**Comment by afcondon** [comment]
Yes, force removing the build/Language/Purescript/CST stuff allowed it to compile - something weird in there about that, after rebuild there is no CST dir, just CST.dyn_hi, .dyn_o, .o, .hi...not going to shave that yak tho, but i will try building our app with the es_modules branch compiler later on

**Comment by afcondon** [comment]
is there a package set for 0.14 yet? i'm assuming i'll need that to build against - (ie Prelude compile fails on   8  data RLProxy (rowlist :: RowList) at the moment)

**Comment by JordanMartinez** [comment]
@afcondon see [this comment](https://discourse.purescript.org/t/updating-the-ecosystem-for-upcoming-0-14-0-release/1144/13) for instructions on using a ps-0.14 compatible package set and [these instructions](https://github.com/purescript/spago#install-all-the-packages-in-the-set) for installing all packages in the set.

**Comment by kl0tl** [comment]
Also you’ll have to override a few packages from Jordan set due to the deprecation of primes in foreign modules exports. You should be fine with the following package set:

```dhall
let upstream =
      https://raw.githubusercontent.com/JordanMartinez/package-sets/fix-warnings-ps-master-compatible/src/packages.dhall sha256:91ea6c8da90935f06b06b60ef6532bfc9f2c1233ec46500d1aab2b27d0dd2b49

let overrides =
      { arrays =
          upstream.arrays ⫽ { version = "master" }
      , assert =
            upstream.assert
          ⫽ { repo =
                "ssh://git@github.com/kl0tl/purescript-assert"
            , version =
                "no-foreign-primes"
            }
      , bigints =
          upstream.bigints ⫽ { version = "master" }
      , decimals =
          upstream.decimals ⫽ { version = "master" }
      , jquery =
          upstream.jquery ⫽ { version = "master" }
      , mysql =
          upstream.mysql ⫽ { version = "master" }
      , node-buffer =
          upstream.node-buffer ⫽ { version = "master" }
      , node-he =
          upstream.node-he ⫽ { version = "master" }
      , node-postgres =
          upstream.node-postgres ⫽ { version = "master" }
      , refs =
          upstream.refs ⫽ { version = "master" }
      , signal =
            upstream.signal
          ⫽ { repo =
                "ssh://git@github.com/kl0tl/purescript-signal"
            , version =
                "no-foreign-primes"
            }
      , st =
            upstream.st
          ⫽ { repo =
                "ssh://git@github.com/kl0tl/purescript-st"
            , version =
                "no-foreign-primes"
            }
      , strings =
          upstream.strings ⫽ { version = "master" }
      , web-html =
          upstream.web-html ⫽ { version = "master" }
      }

let additions = {=}

in  upstream ⫽ overrides ⫽ additions
```

afcondon:
Okay, i tried that, appeared to work flawlessly. 8000 lines of our PureScript app (48 modules) and about 550 modules from the package-set, i just counted `index.js` files in output to get that latter number. Using Parcel 1.12. I forgot to try zephyr but maybe i'll repeat the experiment at the weekend.

thanks to all contributors for the work on this and to @kl0tl and @JordanMartinez for suggestions/help on trying it

JordanMartinez:
@kl0tl Mind submitting a PR to that fork that updates the package set to include those overrides?

kl0tl:
@thomashoneyman I have tested `@ampproject/rollup-plugin-closure-compiler` and `closure-webpack-plugin` and they both work in simple minifier mode, albeit no better than `terser` (the webpack plugin is even worse). They easily beat `terser` with advanced optimizations but the app then breaks at runtime, even with externs declarations for `react` (this isn’t related to ES modules though, it doesn’t even work with the v0.13.6 of the compiler) 🤷‍♂️

As for letting the closure compiler generate the whole bundle: it doesn’t work with ES modules because the compiler can’t export identifiers declared in a CommonJS module from an ES module, which is necessary for re-exporting foreign imports. This isn’t too bad though because babel can transform ES modules to CommonsJS modules understood by the closure compiler with the following configuration:

```json
{
  "plugins": [
    [
      "@babel/plugin-transform-modules-commonjs",
      {
        "loose": true,
        "strict": true,
        "noInterop": true
      }
    ]
  ]
}
```

milesfrain:
> I picked https://github.com/f-f/purescript-react-basic-todomvc as a simple enough but hopefully not too trivial project to test various bundlers configurations with ES modules.

Table of bundle sizes:

Bundle Size (KB) | Branch    
---: | ---    
170 | [parcel@2](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/parcel2-es-modules)    
170 | [webpack@5](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/webpack5-es-modules)    
192 | [webpack@5 with purs bundle](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/webpack5-purs-bundle-es-modules)    
243 | [webpack@5 with purs-loader](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/webpack5-purs-loader-es-modules)    
168 | [rollup](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/rollup-es-modules)    
162 | [rollup with rollup-plugin-purs](https://github.com/kl0tl/purescript-react-basic-todomvc/tree/rollup-plugin-purs-es-modules)

milesfrain:
I was having trouble identifying the specific packages that needed updating to fix primes in FFI, so I split @kl0tl's commit into two. One for formatting changes and one for [meaningful changes](https://github.com/milesfrain/package-sets/commit/e14b0bd84559dee041664eaaeb613a8315f54e98) in my branch. Here's the original [noisy commit](https://github.com/kl0tl/package-sets/commit/d9fb5b55c050c31c8285b6f4f9e3ee8f869e732f).

jiyinyiyong:
no updates for months, is this feature still on-going?

JordanMartinez:
> no updates for months, is this feature still on-going?

It's my understanding that we'll focus on this in `v0.15.0`. We're in the process of finalizing things before making `v0.14.0`.

andys8:
I started rebasing this branch on `v0.14`. I could have broken something during rebase, but it does compile.
https://github.com/andys8/purescript/tree/es-modules-rebased

kl0tl:
I finally had time to come back to this! Re-exports now appear in generated code since https://github.com/purescript/purescript/pull/3883 so we have to be careful about them, otherwise not much happened on codegen over last year.

Node.js has made great progress on the interoperability of CommonJS and ES modules though: ES namespace imports are able to import CommonJS exports since v14.13.0, which is great news for us! Most of my work this weekend has been about ensuring CommonJS and ES foreign modules can coexist natively on Node.js:

+ The compiler output directory now contains a `"type": "module"` package.json, so that Node.js loads .js files as ES modules. Node.js ignores this file when the output directory is named `node_modules` though (this was the case in tests and inside the REPL directory), so if this is a common workflow we could instead add package.json files to _every module directory_ inside the compiler output directory but it feels _slightly_ heavyweight 😅

+ CommonJS foreign modules are copied to the output directory with a .cjs extension, so that Node.js loads them as CommonJS modules. This is necessary because the `require` function, and the `module` and `exports` objects are not available anymore in ES modules. Their exports are re-exported from ES foreign modules, so that we can keep referring to `"./foreign.js"` in compiled PureScript modules. This isn’t stricly necessary, we could refer to `"./foreign.js"` or `"./foreign.cjs"` as needed, but this is the less invasive option and it could let us avoid recompiling a PureScript module when its CommonJS foreign module becomes an ES module.

+ All syntaxes of imports and exports supported by `language-javascript` are allowed in ES foreign modules but `purs bundle` supports only the following subset for now: namespace imports, re-exports of foreign and other PureScript modules, and named exports of local bindings and variables declarations. I’d like to add support for the other syntactic forms in the future but this is the minimum required to bundle the output of the compiler. We don’t actually need to support named exports of variables declarations but I think translating `exports.foo = baz` as `export var foo = baz` is going to be common enough that it deserves support.

There’s still an issue with `"use strict"` directives and comments in the output of `purs bundle` but otherwise it works :tada:

src/Language/PureScript/Make/Actions.hs
```diff
@@ -236,12 +252,30 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
...
+    writeTextFile (outputFilename mn "foreign.js") . toByteString . renderJS $
+      JSAST.JSAstModule
+        [ JSAST.JSModuleExportDeclaration JSAST.JSNoAnnot
+            (JSAST.JSExportFrom
+              (JSAST.JSExportClause JSAST.JSAnnotSpace
+                (JSAST.toCommaList $ JSAST.JSExportSpecifier . JSAST.JSIdentName JSAST.JSAnnotSpace . T.unpack . runIdent <$> S.toList idents)
+                JSAST.JSAnnotSpace)
+              (JSAST.JSFromClause JSAST.JSAnnotSpace JSAST.JSAnnotSpace "\"./foreign.cjs\"")
+              (JSAST.JSSemi JSAST.JSNoAnnot))
+        ] JSAST.JSNoAnnot
```
kl0tl, line 278: I hoped to write an `export * from "./foreign.cjs"` declaration here but this isn’t supported by `language-javascript-0.7.0.0`.

src/Language/PureScript/Make/Actions.hs
```diff
@@ -279,18 +313,41 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
...
+      Right (Bundle.ForeignModuleExports{..}, Bundle.ForeignModuleImports{..})
+        | not (null cjsExports && null cjsImports)
+        , null esExports
+        , null esImports -> do
+            let deprecatedFFI = filter (elem '\'') cjsExports
+            unless (null deprecatedFFI) $
+              errorDeprecatedForeignPrimes deprecatedFFI
+
+            when (elem "default" cjsExports) $
+              errorDeprecatedFFIDefaultCJSExport
```
kl0tl, line 340: If we want to go with this approach I’d like to emit this error as a warning starting from v0.14.1.

tests/purs/bundle/3551/ModuleWithDeadCode.js
```diff
@@ -1,10 +1,8 @@
-"use strict";
-
-var fs = require('fs');
+import * as fs from 'fs';
```
kl0tl, line 1: It would be more idiomatic to write `import fs from 'fs'` or `import { readFileSync } from 'fs'` here but `purs bundle` does not support these syntaxes yet.

tests/purs/bundle/3551/ModuleWithDeadCode.js
```diff
@@ -1,10 +1,8 @@
...
 
 var source = fs.readFileSync(__filename, 'utf-8');
```
kl0tl, line 3: This works because the output of `purs bundle` is interpreted as a CommonJS module by Node.js but `__filename` is not available in ES modules, so we’ll have to support CommonJS foreign modules at least until `language-javascript` can parse `import.meta.url`.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "10"
+  - "14"
```
kl0tl, line 3: We could also specify the version of Node.js with a .nvmrc file (https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc). This would make it easier to enforce the version to use during development (tests won’t pass anymore on <14.13.0).

src/Language/PureScript/Make/Actions.hs
rhendric: I think this is preferable even if `export *` becomes available—enumerating the exports probably makes static analysis of the output a little easier, for both machines and brains.

src/Language/PureScript/Make/Actions.hs
hdgarrood: That is, I think a spec-conformant implementation of ES modules will treat `export default foo` identically to `exports.default = foo`.

src/Language/PureScript/Bundle.hs
```diff
@@ -360,61 +388,158 @@ strValue str = go $ drop 1 str
...
+varNames = mapMaybe varName . fromCommaList
+  where
+  varName (JSVarInitExpression (JSIdentifier _ ident) _) = Just ident
+  varName _ = Nothing
+
+sp :: JSAnnot
+sp = JSAnnot tokenPosnEmpty [ WhiteSpace tokenPosnEmpty " " ]
+
+stringLiteral :: String -> JSExpression
+stringLiteral s = JSStringLiteral JSNoAnnot $ "\"" ++ s ++ "\""
```
hdgarrood, line 424: I think this is a bit dodgy - is language-javascript expecting the string here to be the same as it appears in the JS source code? Won't that cause problems if `s` contains characters like `\n` or `"`? I probably would have expected `language-javascript` to provide this function for us.

.travis.yml
hdgarrood: That sounds fine as long as we aren't requiring people to use nvm. But I think having the tests fail on <14.13.0 is probably fine too. Why do the tests fail? Does the repl still work on older versions of node?

src/Language/PureScript/Make/Actions.hs
hdgarrood: Ah, ok, so what I was missing was that a) `default` isn't allowed as a variable name, and b) in node.js, the `exports` object is treated as the default export when importing CJS from ESM.
```javascript
// test.js
import thing1 from "./def.cjs"
import thing2 from "./def.js"

console.log(thing1);
console.log(thing2);

// def.cjs
exports.default = "hello, world";

// def.js
export default "hello, world";
```
produces
```
$ node test.js
{ default: 'hello, world' }
hello, world
```
so if we are going to be using named exports in ESM, then yeah, I can't think of another way around it. Warning on CJS "default" exports also seems reasonable to me given that we are going to drop support for CJS FFI modules eventually anyway.

.travis.yml
kl0tl: The tests and the REPL need to be able to import CommonJS foreign modules with ES named imports, but this isn’t possible before v14.13.0:

```purs
module Example where

example = "Hello world"
```

```sh
> purs repl Example.purs 'bower_components/purescript-*/src/**/*.purs'
Compiling Example
PSCi, version 0.14.0
Type :? for help

> import Example
> example

```

```sh
> node
Welcome to Node.js v14.2.0.
Type ".help" for more information.
> import('./.psci_modules/index.js')
Promise { <pending> }
> (node:28607) UnhandledPromiseRejectionWarning:
export { cons, join, showArrayImpl, showCharImpl, showIntImpl, showNumberImpl, showStringImpl } from "./foreign.cjs";
         ^^^^
SyntaxError: The requested module './foreign.cjs' does not provide an export named 'cons'
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:92:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:107:20)
    at async Loader.import (internal/modules/esm/loader.js:179:24)
```

I did not reconsider the version requirement when introducing ES wrappers for CommonJS foreign modules though, but we could import the default export of the CommonJS foreign module in the ES wrapper and re-export properties of the default export with export declarations:

```js
// foreign.cjs
"use strict";

exports.log = function (s) {
  return function () {
    console.log(s);
  };
};
```

```js
// foreign.js
import $foreign from "./foreign.cjs";

export var log = $foreign.log;
```

This has the advantage of working all the way back to Node.js v12.17.0 (earlier v12 versions need `--experimental-modules`) which is great news since support for v10 stops at the end of the month!

src/Language/PureScript/Make/Actions.hs
kl0tl: If the ES wrapper imports the default export of the CommonJS foreign module and re-export its properties with export declarations as I thought about in https://github.com/purescript/purescript/pull/3791#discussion_r611029477 we could keep supporting CommonJS exports named default:

```js
import $foreign from "./foreign.cjs";

var $$default = $foreign.default;
export { $$default as default };
```

src/Language/PureScript/Bundle.hs
kl0tl: Agreed, but I just renamed the existing `str` function 😬

.travis.yml
hdgarrood: This approach sounds good to me :+1:

hdgarrood: How far back is `--experimental-modules` understood?

andys8: > Back in 2017, Node.js 8.9.0 shipped experimental support for ECMAScript modules, known for their import and export statements. This support was behind the flag --experimental-modules.

https://nodejs.medium.com/announcing-a-new-experimental-modules-1be8d2d6c2ff

kl0tl: We’re really interested in the implementation of --experimental-modules that was shipped with v12.0.0, this flag is of no use to us in v10.0.0 and earlier.

hdgarrood: So if we can expect this approach to work when supplying `--experimental-modules` with any version of node.js >= v12.0.0, and if earlier versions of node are EOL, this sounds fine to me. I think we might want to check the version of `node` we are using in the repl and supply that flag where necessary. Perhaps we should test on node v12 as well as v14 for now as well?

src/Language/PureScript/Bundle.hs
hdgarrood: Oh haha, never mind then.

src/Language/PureScript/Bundle.hs
```diff
@@ -219,13 +243,13 @@ printErrorMessage (ErrorInModule mid e) =
     displayIdentifier (ModuleIdentifier name ty) =
       name ++ " (" ++ showModuleType ty ++ ")"
 printErrorMessage (MissingEntryPoint mName) =
-  [ "Couldn't find a CommonJS module for the specified entry point: " ++ mName
+  [ "Couldn't find neither an ES nor CommonJS module for the specified entry point: " ++ mName
```
i-am-the-slime, line 246:
```suggestion
  [ "Could neither find an ES nor a CommonJS module for the specified entry point: " ++ mName
```

thomashoneyman: Or alternately:

> Could not find an ES module or CommonJS module for the specified entry point ...

.travis.yml
kl0tl: The REPL and the tests now throw on unsupported Node.js versions (<12.0.0) and provide the --experimental-module flag for versions >=12.0.0 && <12.7.0.

> Perhaps we should test on node v12 as well as v14 for now as well?

We don’t rely on the behaviour introduced in v14.13.0 anymore, so testing on v12.0.0 and the latest v12 might be enough?

kl0tl:
I merged master and implemented the strategy described [in this comment](https://github.com/purescript/purescript/pull/3791#discussion_r611029477) to support Node.js >=12.0.0!

JordanMartinez:
Nice!

Just FYI. CI failures are caused by hlint finding a few things.

cocoddy:
what is the reason to stop merge this

JordanMartinez:
@cocoddy It's not yet done. I don't recall everything, but one open question is what to do about `purs bundle`.

thomashoneyman:
> Node.js has made great progress on the interoperability of CommonJS and ES modules though: ES namespace imports are able to import CommonJS exports since v14.13.0, which is great news for us! Most of my work this weekend has been about ensuring CommonJS and ES foreign modules can coexist natively on Node.js

I think we should support _only_ ES modules in FFI for simplicity's sake.

This means that CommonJS modules would not be accepted in FFI files, and we'd attempt no automatic conversions or processing to have ES modules and CommonJS work together. As far as the compiler is concerned it's all ES modules in and ES modules out.

However, the world doesn't work this way: plenty of browser and Node projects currently rely on CommonJS and we need to be sure that they continue to work even if the compiler only accepts ES modules. Still, I think we're in the clear:

1. We can omit the browser part of this question because browsers do not support CommonJS; you will _have_ to use a bundler if you need CommonJS for the browser, and bundlers support both module types.
2. I _believe_ we can omit the Node part of this question, since Node supports both module types from Node 14 onward and we have evidence from the [`node-challenge`](https://github.com/working-group-purescript-es/challenges/tree/main/node-challenge) that mixing module types works fine.

To use a CommonJS module in your FFI, you would need to take one of two routes. The first is where the CommonJS is using named exports:

```js
// commonjs uses named exports
// mymodule.js
exports.foo = ...
exports.bar = ...

// es modules FFI file can't do named imports directly; you'd have to import the
// full module then pick off what you need
import mymodule from "mymodule.js"

export const foo = mymodule.foo;
export const bar = mymodule.bar;
```

The second is where the CommonJS is using a default export:

```js
// commonjs uses default export
// mymodule.js
module.exports = ...

// es modules
import mymodule from "mymodule.js"

export default mymodule;
```

Either way, it's possible to write your FFI in such a way that you only write an ES module, and if you need to pull in CommonJS dependencies then you rely on either a bundler in the case of the browser (which is required anyway), or rely on Node's module resolutions. Either way we can get away with only supporting ES modules.

jiyinyiyong:
better using

```js
import mymodule from "./mymodule.js"
```

since `"mymodule.js"` is treated as a package, rather than a file.

rhendric:
Closed in favor of #4232.

PR #3792, opened 2020-02-16
Deprecate primes in identifiers exported from foreign modules

kl0tl:
This PR implements @hdgarrood [suggestion](https://github.com/purescript/purescript/issues/3613#issuecomment-583794497) regarding the depreciation of primes in foreign modules exports.


src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -312,7 +312,19 @@ moduleToJs (Module _ coms mn _ imps exps foreigns decls) foreign_ =
...
+        (AST.Ternary Nothing
+          (AST.App Nothing
+            (AST.Indexer Nothing (AST.StringLiteral Nothing "hasOwnProperty")
+                                 (AST.Var Nothing "$foreign"))
+            [AST.StringLiteral Nothing $ mkString unescaped])
+          (AST.StringLiteral Nothing $ mkString unescaped)
+          (AST.StringLiteral Nothing $ mkString (T.replace "'" "$prime" unescaped)))
+        (AST.Var Nothing "$foreign"))
+    | otherwise
+    = accessorString (mkString unescaped) (AST.Var Nothing "$foreign")
```
natefaubion: Is this emitted for every reference to a foreign identifier within a module?

natefaubion:
Personally, I'm inclined to remove the `hasOwnProperty` check and just disallow foreign primes entirely rather than suggest you use `$prime` escaping.

hdgarrood:
This looks like a great start, thank you! I think it might be preferable to use the existing `IfElse` rather than using a new `Ternary` constructor though (it slipped my mind that we don't already have one).

hdgarrood:
> Personally, I'm inclined to remove the hasOwnProperty check and just disallow foreign primes entirely rather than suggest you use $prime escaping.

So I guess we would spend a major release series warning on FFI primes, and where the way of silencing the warning is to rename to something which doesn't contain primes at all, and then make it an error when we do ES modules? I guess that's preferable in that it doesn't involve messing with the generated code or `purs bundle` at all.

hdgarrood:
Also it means that people are less likely to accidentally break their libraries on older compiler versions when dealing with the warning, which is probably also a plus.

src/Language/PureScript/CodeGen/JS.hs
kl0tl: `$foreign[$foreign.hasForeignProperty("foo'") ? "foo'" : "foo$prime"]` would have been emitted for every reference to a `foo'` identifier imported from a foreign module yes.

kl0tl:
I’ve removed the support for escaped primes and updated the deprecation message.

tests/purs/warning/DeprecatedFFIPrime.out
```diff
@@ -0,0 +1,56 @@
+Error 1 of 4:
+
+  at tests/purs/warning/DeprecatedFFIPrime.purs:5:1 - 10:28 (line 5, column 1 - line 10, column 28)
+
+    In the FFI module for [33mMain[0m:
+
+      The identifier [33ma'[0m contains a [33m'[0m.
```
hdgarrood: I think people might not realise that a `'` is the same thing as a prime, which might make this wording confusing. How about having the error say this:

```
The identifier a' contains a prime (').
Primes in identifiers ...
```

kl0tl: I applied your suggestion!

kl0tl:
Sorry for missing that, I just did.

hdgarrood:
No worries - thanks very much!

kl0tl:
I opened a few pull requests removing primes from foreign modules on the libraries I had to patch to be able to run tests while working on https://github.com/purescript/purescript/pull/3791. There’s probably a lot more libraries to upgrade in the official package set but that’s a start 😅

garyb:
Thanks! I'll go through and get those all merged before we make the new release.

kl0tl:
I rebased the PR onto master (`SimpleErrorMessage` moved from `Language.PureScript.AST.Declarations` to `Language.PureScript.Errors`) and opened pull requests on all other packages impacted by the depreciation in the set psc-0.13.6-20200309!

hdgarrood:
Any objections to merging this now?

natefaubion:
No objections from me.

Issue #3633, opened 2019-05-09
Consider splitting up the compiler into multiple packages


natefaubion:
This was brought up in Slack (primarily by @f-f), and I think it's something we should discuss.  

* The whole compiler is a very large dependency if we want to encourage external tooling that might only depend on a relatively small part (eg. parsing, corefn).
* We would like to keep the maintenance burden as equivalent to the status quo as possible (changes frequently require updating all parts of the compiler, publishing to hackage).
* What packages would we split out? What would that structure look like internally? What internal tooling would need to change or be updated to support the workflow?

hdgarrood:
I'd suggest only splitting things up once we've identified specific cases where it would be useful to depend on only those things, so perhaps we could just start with the parser.

This sort of workflow is supported by Stack and used by projects like yesod, see for example https://github.com/yesodweb/yesod so I'd suggest using whatever Stack expects to see (although I haven't yet looked into what that means exactly).

jmackie:
I think the parser and corefn stuff would be super useful to have in their own packages :+1:

hdgarrood:
The thing about corefn is that right now I think we only define corefn -> json conversion and not the other way; for corefn to be useful on its own we'd need the conversion in the other direction right?

jmackie:
The other direction is here https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CoreFn/FromJSON.hs#L109 ?

hdgarrood:
Lol, ignore me then

f-f:
@natefaubion thanks for opening the discussion on this!

In spago I sometimes wish we had access to the compiler (mostly for the parser), but we cannot depend on it in the current state as build times would get long.

So it could be nice if e.g. the parser was a separate package (e.g. I already tried to depend on `purescript-cst` at some point, see https://github.com/natefaubion/purescript-cst/pull/10), but on the other hand I'm not sure this is a good idea on our side: after all `pulp` has managed just fine without a compiler dependency for long time

So far we avoided being blocked by "depending on the compiler", except for https://github.com/spacchetti/spago/issues/165, where we might have to choose between:
1. implementing the functionality in the compiler (problem: adding maybe orthogonal functionality under the "facilitating build tools" label)
2. implementing a subset of the PS parser in spago (problems: it might get out of sync with upstream, and other build tools will have to reimplement it if they want the feature)
3. depending on the compiler (problem: build times)

[...]

PR #3793, opened 2020-02-16
Split into packages

joneshf:
This is a first crack at [splitting into multiple packages][multiple packages]. This follows [some of what @jmackie did](https://github.com/purescript/purescript/issues/3633#issuecomment-500745777), and does some stuff differently.

My primary motivation for doing this is to make building `purty` less time-consuming. For reference, a clean TravisCI build barely makes it in the time allotted on Linux and always timesout by a few minutes on macOS. An older build that built all of the `purescript` package (and `purty` tests) on macOS [took 1,718s](https://travis-ci.org/joneshf/purty/jobs/649032937#L105). With the packages split, the equivalent step on macOS [took 691s](https://travis-ci.org/joneshf/purty/jobs/651199775#L484). The entire clean build on macOS took 34min 25s, which is well under the timeout. Secondary motivations are the things laid out in the [multiple packages][] issue.

Since my primary motivation is to make building `purty` less time-consuming, I will end up supporting this fork for the indefinite future if this PR is not accepted and an alternative is not proposed. Since `purty` is the motivation, I'll only sync it on released versions. I have no problem making changes to get this PR accepted, but I only have so much motivation to do so. Which is to say, I'm not going to be mad if this PR is rejected, but I'm also not going to do a ton of work to get this PR accepted.

This is currently branched off of [v0.13.6](https://github.com/purescript/purescript/tree/4dd32c48a372bb8a5250140d19af8aa18a6358d7) due to the motivations above. Will get up to date if it's accepted.

[multiple packages]: https://github.com/purescript/purescript/issues/3633


lib/purescript-ast/package.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - MultiParamTypeClasses
+    - NoImplicitPrelude
+    - PatternGuards
+    - PatternSynonyms
+    - RankNTypes
+    - RecordWildCards
+    - OverloadedStrings
+    - ScopedTypeVariables
+    - TupleSections
+    - ViewPatterns
```
joneshf: What's our policy on default extensions? I don't think we even use all of these in the AST proper. But then, there's extensions in the modules that are already listed here.

Were I to make a suggestion, it'd be to not use `default-extensions` anymore.
* The tooling around extensions is spotty.
* We cargo cult some extensions like I did here.
* Knowing how a module is going to behave can be restricted to each module.

I don't particularly want to push this change. But it was something I was unclear about and didn't know what to do (so I cargo culted).

lib/purescript-ast/package.yaml
```diff
@@ -0,0 +1,61 @@
...
+author: Phil Freeman <paf31@cantab.net>
+maintainer: >
+  Gary Burgess <gary.burgess@gmail.com>,
+  Hardy Jones <jones3.hardy@gmail.com>,
+  Harry Garrood <harry@garrood.me>,
+  Christoph Hegemann <christoph.hegemann1337@gmail.com>,
+  Liam Goodacre <goodacre.liam@gmail.com>,
+  Nathan Faubion <nathan@n-son.com>
+copyright: (c) 2013-17 Phil Freeman, (c) 2014-19 Gary Burgess, (c) other contributors (see CONTRIBUTORS.md)
+license: BSD3
```
joneshf: I assume these are all fine to go with? If not, more than happy to change whatever.

lib/purescript-ast/package.yaml
```diff
@@ -0,0 +1,61 @@
+name: purescript-ast
+version: '0.13.6'
+synopsis: PureScript Programming Language Abstract Syntax Tree
+description: Defines the underlying syntax of the PureScript Programming Language.
```
joneshf: Wasn't entirely sure what to put here. Feel free to suggest any changes if you'd like.

license-generator/generate.hs
```diff
@@ -59,9 +59,14 @@ depsNamesAndVersions :: IO [(String, String)]
...
   contents <- lines <$> getContents
   deps <- traverse parse contents
-  pure (filter (\(name, _) -> name /= "purescript" && name /= "rts") deps)
+  pure (filter (\(name, _) -> not (excluded name)) deps)
 
   where
+  excluded name =
+    name == "purescript"
+    || name == "purescript-ast"
+    || name == "rts"
```
joneshf: Is this the right thing to do? I assume we were ignoring `purescript` because we can't look up our license over the wire while we're trying to build out license. I assume we were ignoring `rts` because it's not a real hackage package.

Since `purescript-ast` is a package we own, we would run into a similar issue as the `purescript` package trying to lookup its own license on hackage`. But, since `purescript` depends on `purescript-ast`, shouldn't it have the `purescript-ast` as part of the license? Should `purescript-ast` not define its own license and instead use the top-level `LICENSE` file? I don't have an answer to any of these questions. I'm down to change to whatever.

garyb:
I have no objection to splitting things up like this, it'll perhaps benefit compile/test times for non-syntax-related situations when working locally too :+1:

stack.yaml
```diff
@@ -2,6 +2,7 @@ resolver: lts-13.26
 pvp-bounds: upper
 packages:
 - '.'
+- lib/purescript-ast
```
joneshf: This naming convention follows what was [suggested](https://github.com/purescript/purescript/issues/3633#issuecomment-500745777). I don't have a strong feeling one way or the other. If we'd prefer a different convention, I'm down to rename.

lib/purescript-cst/package.yaml
```diff
@@ -0,0 +1,77 @@
...
+category: Language
+author: Phil Freeman <paf31@cantab.net>
+maintainer: >
+  Gary Burgess <gary.burgess@gmail.com>,
+  Hardy Jones <jones3.hardy@gmail.com>,
+  Harry Garrood <harry@garrood.me>,
+  Christoph Hegemann <christoph.hegemann1337@gmail.com>,
+  Liam Goodacre <goodacre.liam@gmail.com>,
+  Nathan Faubion <nathan@n-son.com>
+copyright: (c) 2013-17 Phil Freeman, (c) 2014-19 Gary Burgess, (c) other contributors (see CONTRIBUTORS.md)
```
joneshf: Similar here about who should be in this list. Given that @natefaubion did the vast majority of the work, is this even accurate?

lib/purescript-cst/package.yaml
```diff
@@ -0,0 +1,77 @@
+name: purescript-cst
+version: '0.13.6'
+synopsis: PureScript Programming Language Concrete Syntax Tree
+description: The surface syntax of the PureScript Programming Language.
```
joneshf: Also not sure what to put here. Most of the wording is from [the PR to add the CST](https://github.com/purescript/purescript/pull/3608). Down to change it to whatever.

lib/purescript-cst/package.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - MultiParamTypeClasses
+    - NoImplicitPrelude
+    - PatternGuards
+    - PatternSynonyms
+    - RankNTypes
+    - RecordWildCards
+    - OverloadedStrings
+    - ScopedTypeVariables
+    - TupleSections
+    - ViewPatterns
```
joneshf: Similar here about `default-extensions`. Marking this so we don't lose track of it.

lib/purescript-cst/package.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - containers
+  - dlist
+  - purescript-ast
+  - scientific >=0.3.4.9 && <0.4
+  - semigroups >=0.16.2 && <0.19
+  - text
+build-tools:
+  - happy ==1.19.9
+
+library:
```
joneshf: Not sure where to put this, so I'll arbitrarily choose this line.

I want to call attention to the commit message:

> It's important to note that we don't pull out `Language.PureScript.CST`.
That module is straddling the line between a re-exporting "prelude" and
an adapter between `purescript-ast` and `purescript`. Since it deals
with errors, and errors currently exist at the top-level, we cannot yet
move this module down into `purescript-cst`.
>
> Probably what needs to happen is that we make a split between
`Language.PureScript.CST` being a re-exporter and
`Language.PureScript.CST` doing adapter work. Not clear which way to go
on this one as both seem viable avenues.

I think this is an important point to make. Just because the module has `CST` in it, doesn't mean it has to go with the `purescript-cst` package. That module is doing double duty. If we wanted to separate the responsibilities, we could do that. Though, I think the re-exporting is an anti-pattern for this project. If we wanted to rename it, we could do that. If we don't care, we can leave it. I don't have any strong feelings, but wanted to make sure this was explicitly known about.

license-generator/generate.hs
```diff
@@ -65,6 +65,7 @@ depsNamesAndVersions = do
   excluded name =
     name == "purescript"
     || name == "purescript-ast"
+    || name == "purescript-cst"
```
joneshf: Similar here about licenses and `purescript-cst`. Marking this so we don't forget it.

lib/purescript-ast/src/Control/Monad/Supply.hs
```diff
@@ -1,19 +1,20 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PackageImports #-}
 
 -- |
 -- Fresh variable supply
 --
 module Control.Monad.Supply where
```
joneshf: This module and `Control.Monad.Supply` are moving to `purescript-ast` because they are used (transitively) in `Language.PureScript.Names`. My understanding is that the only reason this module exists in this repo in the first place is that there wasn't a fresh variable supply module to pull in. Were we to depend on a fresh variable supply module, we would assume it to be a further down dependency than `purescript-ast`: `purescript` -> `purescript-ast` -> `monad-supply`, or whatever.

More than down to flip some dependencies if that's not right. Wanted to call out why this module moved.

joneshf:
Given that @natefaubion opened [the original issue](https://github.com/purescript/purescript/issues/3633), requesting a review from him explicitly. You don't have to review this, nor do you have to review it in isolation (anybody is welcome to review). Just wanted to make sure there was at least one explicit reviewer.

src/Language/PureScript/Constants/Prim.hs
```diff
@@ -0,0 +1,151 @@
...
+
+import Prelude.Compat
+
+import Data.String (IsString)
+import Language.PureScript.Names
+
+-- Prim values
+
+undefined :: forall a. (IsString a) => a
+undefined = "undefined"
```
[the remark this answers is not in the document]
garyb: I think it's used for JS undefined rather than anything in PS, so yeah it probably shouldn't be in here.

natefaubion:
Thanks for tackling this! I think ideally purescript-cst would not need to depend on purescript-ast, but I don't know if it's a big deal. You addressed the double-duty aspect in a comment, and I agree, I don't think that conversions/adapters really needs to live with CST. If the primary tool that downstream authors want from this is to utilize the surface parser, then I'd think we'd want it to be as light as possible. I think aside from the conversions and reexports, CST only depends on `Names` and `PSStrings`.

What prompted the addition of package name imports? I'm not necessarily against it, but it doesn't seem like it's something that's commonly utilized in the Haskell ecosystem.

src/Language/PureScript/Constants/Prim.hs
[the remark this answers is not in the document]
joneshf: Ah, so maybe there's a third module: `Language.PureScript.Constants.JavaScript`?

joneshf:
Yeah, I wasn't entirely sure which way to go. At first I went that way and pulled out those two modules (`Language.PureScript.Names` and `Language.PureScript.PSString`) to a `purescript-core` package, then layered `purescript-cst` atop that. If we pull more modules out, we quickly start pulling out the AST. So, I restarted and pulled `purescript-ast` out and `purescript-cst` atop it.

I might be reading too much into the wording, but it seems like the CST should depend on the AST. I don't have a strong argument against breaking the dependency, but why would we want to? A very thin parser package would be nice, but it might not be worth the effort if it's only for the sake of downstream consumers. From the perspective of `purty`, having 100+ fewer transitive dependencies to build is a good enough win. Dropping another 10-15 wouldn't seem like that big a deal.

As far as how to break the dependency, maybe we follow suit with what CST did for source positions and define something equivalent to `Language.PureScript.Names` and `Language.PureScript.PSString` specific to CST? That should allow the parsing part of CST to be separate from the rest. I don't know about having the conversions from CST to AST living in `purescript-ast` though; that seems backwards. I might be misunderstanding what you're suggesting, though.

The main motivator for `PackageImports` was trying to figure out where `Control.Monad.Supply*` was coming from. Took longer than I'd like to admit to figure out it was defined in this repo. Hopefully the next person can figure it out in less time. It was also really useful in finding which packages were actually used when trimming the `package.yaml`s for `purescript-ast`/`purescript-cst`.

It's a real shame that more people don't use this extension. I started using it a couple years back, found it to be incredibly useful with learning a new section of code, and never looked back. If we don't want it, I've no problem removing it. But if I'm maintaining this code as a fork, I'm keeping it.

natefaubion:
My reasoning for potentially inverting the dependency:
* AST is a later stage in the pipeline than the CST. I feel like it makes more conceptual sense for later stages to depend on earlier stages, and not so much the other way around. You incur more dependencies the farther along you get in the process. The AST is "parsed" from the CST, so I think it's totally fine for a conversion to live with the AST. If CST depends on AST, it's not clear to me why we want the ast/cst package distinction then. I don't think you would (or could) do anything with just ast.
* I'd like CST to be as lightweight as possible to potentially make vendoring easier. My original intention was to have all the CST types completely isolated in a single file. This way, if tooling needed to support moving between multiple parser versions (due to breaking changes), it would be as strightforward as copying a couple of files.  I do not think anything is gained by reusing `Names` in CST. I think in hindsight, I should have dropped `Names` usage and used basic `ProperName` vs `VariableName` types. `PSString` I think might be hard to lose, I dunno.

I'm not necessarily saying those changes are necessary, I'm just trying to explain choices that aren't apparent from what was actually executed.

For package imports, is there a way to enforce usage? The compiler is not rigid stylistically, and code reflects the particular authors. I'm just concerned it will bit rot quickly, as there isn't a very large (Haskell) community drive to use this feature.

joneshf:
Oh, I see what you're saying.

I was coming at it from the perspective of CST being a parser for a particular syntax of PureScript the language. The idea being you could use a different syntax parser (maybe Algol-based syntax, maybe Lisp-based syntax, maybe another ISWIM-based syntax), convert it into the AST, and have the whole rest of the compiler work.

As mentioned above, the difference between AST and CST isn't large enough for me personally to care about, so whichever way is right I'm okay with. Shall I switch things around?

I don't think there's anything built into GHC to force `PackageImports` to be used. At a previous job we wrote a shell script to parse imports and error if it didn't have a package import. It worked better than expected. Something like: `git grep --extended-regexp '^import\s+' | grep --extended-regexp --invert-match '".+"'`. Could do something like that if we wanted to enforce it.

natefaubion:
> I was coming at it from the perspective of CST being a parser for a particular syntax of PureScript the language. The idea being you could use a different syntax parser (maybe Algol-based syntax, maybe Lisp-based syntax, maybe another ISWIM-based syntax), convert it into the AST, and have the whole rest of the compiler work.

Interesting! I hadn't considered that.

lib/purescript-ast/package.yaml
hdgarrood: I would rather stick with default-extensions for almost all extensions other than TemplateHaskell; it can be a bit annoying if things like TupleSections or ViewPatterns are enabled on an ad-hoc basis from file to file and you end up with stylistic differences just based on what extensions happened to be enabled in that file. Also, they're a big source of merge conflicts in my experience.

license-generator/generate.hs
hdgarrood: `purescript-ast` should definitely have its own LICENSE file, yeah. Since we won't be distributing it as a binary, just as source code, we won't be distributing it alongside its dependencies, so we needn't bother with a license-generator setup like we have with the compiler now. So I'd say the LICENSE files for any split-off packages should just be the normal BSD3 license text and nothing else.

lib/purescript-cst/package.yaml
hdgarrood: I think so, at least in the case of anyone who worked on the parser and/or syntax before the CST work was done, because @natefaubion's work built off that.

hdgarrood:
I'm not particularly fussed about which modules go where at this stage. In general this seems okay; there are only two things I would like to address before merging:
- I think we should use separate version numbers for the split-off libraries. We (just about) get away with using the public compiler version for the `purescript` package even though we regularly "break" it from a PVP standpoint without doing a breaking change bump, because it's not really intended for use as a library, only for the executables. Since we are splitting off packages like this in order to allow people to use those bits as libraries more easily, I think we should treat them more like libraries, and follow the PVP. So, I think we should use 4-component versions for them (really-major, major, minor, patch) and start them both off at version 0.1.0.0.
- As mentioned above, we won't distribute binaries for the split-off libraries, so we don't need to include copies of the dependencies' licenses; we should have normal, static LICENSE files with just the BSD3 license text in those, and no special treatment of those libraries in the main license generator script.

natefaubion:
> So, I think we should use 4-component versions for them (really-major, major, minor, patch) and start them both off at version 0.1.0.0.

I dunno, but this seems confusing to me. It's to implement tooling, which is almost always bound by what version of the language you support. It uses 0.1.3.4, what version of PureScript does that map to? I can see it both ways though. I haven't implemented tooling, so I'll defer to @joneshf on what is most helpful.

hdgarrood:
Right, if we did separate the version numbers we'd need to keep a table in the README of each split-off package saying which package versions map to which compiler versions. I see your point too, though. It's just that the Hackage PVP is explicit about what you should and should not do as a package maintainer, and we would definitely be contravening it if we used the same versions across all of our packages. On second thoughts I'm happy to go either way to start with, we can wait and see if it actually turns out to be a problem in practice.

natefaubion:
FWIW I think that hackage supports some sort of metadata that your package doesn't follow the PVP.

timjs:
Just a side note on difference syntax frontends.

Lately, I was actually thinking about creating a Swift/Rust/Go based syntax for PureScript. I was triggered by [this list of languages](https://github.com/appcypher/awesome-wasm-langs) that compile to WebAssembly. IMHO, the languages that emerged in the last two years are semantic wise very similar. That is, language theoretically, they do not add mind blowing features. They differ in the syntax they choose (Python based, C based, Haskell based, Ruby based, Lisp based, etc.) and if they are object-first of functional-first.

[One of the languages on the list](https://github.com/forest-lang/forest-compiler) is based on the idea of "functional, agree on semantics, disagree on syntax". But why make another one when PureScript has all the basic ML functionality extended with type classes, row types, higher-kinded types and now even poly kinds? [ReasonML](https://reasonml.github.io) does something similar for OCaml. I think this could work for PureScript too. Worth keeping these possibilities in mind!

lib/purescript-ast/package.yaml
joneshf: Okay. It sounds like your qualm is with the third bullet. That's understandable. As mentioned, I'm not going to push this change. If I work up the gumption to suggest the change, we can have that discussion in a separate PR.

---

We rewrite extensions that exist in the `package.yaml` file in individual modules and we carry around unused extensions in the `package.yaml`. I'm saying that it's unclear why we do these two things. I'm also asking what we should do about it.

license-generator/generate.hs
joneshf: Okay, that makes sense. Given that information, I'd assume the license generator should be changed to pull in the license from local packages that are used. That sound right?

lib/purescript-cst/package.yaml
joneshf: Makes sense. Thanks!

lib/purescript-ast/package.yaml
hdgarrood: Right, so to clarify my position I’d say I would be in favour of removing pragmas from source files and using the same set of default-extensions everywhere. It doesn’t bother me if an extension is enabled and not used - I’d prefer that to having different “versions” of the language in use across different parts of the compiler. For example I want to be able to copy some code from one module to another and have it Just Work as far as possible without having to mess with pragmas. Having said that I’m not terribly fussed about this; the current situation is bearable for me.

license-generator/generate.hs
hdgarrood: Oh yeah that’s a good point, I previously assumed that they would be included in the output of `stack ls dependencies` but I guess that is probably incorrect

lib/purescript-ast/package.yaml
garyb: ☝️ that was the rationale for expanding `default-expansions` originally. There was a time were there was a total hodgepodge of extensions enabled variously in different modules.

joneshf:
> > So, I think we should use 4-component versions for them (really-major, major, minor, patch) and start them both off at version 0.1.0.0.
> 
> I dunno, but this seems confusing to me. It's to implement tooling, which is almost always bound by what version of the language you support. It uses 0.1.3.4, what version of PureScript does that map to? I can see it both ways though. I haven't implemented tooling, so I'll defer to @joneshf on what is most helpful.

> Right, if we did separate the version numbers we'd need to keep a table in the README of each split-off package saying which package versions map to which compiler versions. I see your point too, though. It's just that the Hackage PVP is explicit about what you should and should not do as a package maintainer, and we would definitely be contravening it if we used the same versions across all of our packages. On second thoughts I'm happy to go either way to start with, we can wait and see if it actually turns out to be a problem in practice.

Honestly, the reason I didn't use different versions to start with was because I thought it would be rejected.

I think the answer to this question largely comes down to how we want these different packages to evolve. There are surely bugs we could fix in `purescript-ast`, `purescript-cst`, or any other packages that might be spun off in the future. There's also general changes we could make that wouldn't necessarily be breaking for PureScript the language, but would be for `purescript-ast`.

If we do want to allow `purescript-ast`/`purescript-cst` to move at a different speed than `purescript`, we should use different versions. If we don't want `purescript-ast`/`purescript-cst` to move at a different speed than `purescript`, we should keep the versions in sync.

I don't know how much external tooling should be factoring into this decision. It's good feedback to have, but it shouldn't be deciding our versioning scheme. A table of which versions work together seems like a fine way to deal with compatibility.

My personal preference (not tooling-based) would be whatever allows more iteration in PureScript the language. If we don't have to always think about the entire language when making a change, but can focus on doing what's the next improvement for a part of the language, we ought to end up in a better place. The bigger decisions of how does a change affect the entire language will still happen. But they will happen in the `purescript` package, where it really matters.

lib/purescript-ast/package.yaml
joneshf: Gotcha. Thanks!

How about this:
1. Move all<sup>1</sup> the `purescript-ast`/`purescript-cst` extensions to their respective `package.yaml` files.
1. Move all<sup>1</sup> `purescript` extensions to the root `package.yaml` file.
1. Make a `default-extensions.yaml` file in the root directory.
1. Use either [YAML includes](https://github.com/sol/hpack#not-repeating-yourself) or [`hpack`s default mechanism](https://github.com/sol/hpack#defaults) to include `default-extensions.yaml` in all of the `package.yaml` files.
1. Write a linter (shell script) that checks for extensions in modules and fails if any extensions are enabled<sup>1</sup>.

We can do step 1 in this PR, and steps 2-5 separately. I'm down to make an issue to track that, if so.

<sup>1</sup>: Not including `TemplateHaskell`.

license-generator/generate.hs
joneshf: The packages are in there, but we look for the licenses from `hackage` and the local packages don't have licenses on hackage.

joneshf: Having written that out, I guess the question is, when should this license generator be run? Is it expected to be run only when making a new release (in which case, the local packages should be on `hackage`)? Or should it always be possible to run the license generator (in which case, we should handle local packages differently)?

timjs:
Your reasoning sounds good to me @joneshf!

src/Language/PureScript/Constants/Prim.hs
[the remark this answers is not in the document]
joneshf: Did a bit more looking. It's used:
* https://github.com/purescript/purescript/blob/81909a9585b00bc99cc47c7959fbc5ae2e1ff285/src/Language/PureScript/CoreFn/Desugar.hs#L92
* https://github.com/purescript/purescript/blob/81909a9585b00bc99cc47c7959fbc5ae2e1ff285/src/Language/PureScript/CoreImp/Optimizer/Inliner.hs#L314
* https://github.com/purescript/purescript/blob/81909a9585b00bc99cc47c7959fbc5ae2e1ff285/src/Language/PureScript/CoreImp/Optimizer/TCO.hs#L126
* https://github.com/purescript/purescript/blob/81909a9585b00bc99cc47c7959fbc5ae2e1ff285/src/Language/PureScript/CoreImp/Optimizer/Unused.hs#L26
* https://github.com/purescript/purescript/blob/81909a9585b00bc99cc47c7959fbc5ae2e1ff285/src/Language/PureScript/TypeChecker/Entailment.hs#L193

I _think_ it's playing double duty: a sigil that might be erased later on, a value that's generated in JavaScript. I think it should be detangled, but I don't want to get bogged down in doing that in this PR. Punting on this for now.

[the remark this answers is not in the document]
garyb: Ah ok! :+1:

lib/purescript-ast/package.yaml
joneshf: I ended up doing effectively that. I also left out `CPP`, `DeriveAnyClass` and `ImplicitParams` for the reasons explained in [this commit](https://github.com/purescript/purescript/pull/3793/commits/a7da01797f750966f933f7c22ef04ced237e90ee). Lemme know if this works or if there's more to do here.

license-generator/generate.hs
joneshf: I'm going with the idea that we only care about running `make license-generator` when doing a release (so local packages should be on `hackage`) based on the [release guide](https://github.com/purescript/purescript/blob/fcc7a08a21a8de1dafb8624de02bfc94f81341b9/RELEASE_GUIDE.md#before-making-a-release).

Lemme know if that's the wrong decision.

joneshf:
@hdgarrood @natefaubion shall we go with differing versions and a compatibility table? That's in line with what @hdgarrood was saying about packages being good `hackage`/PVP citizens while also hopefully fostering improvements to the language as a whole.

natefaubion:
@joneshf I don't have any problem with that.

kritzcreek:
The tests are failing locally with `Module Prim.Coerce was not found.` but I'm assuming that's going to go away once this is brought up-to-date with master.

hdgarrood:
I think we might want to occasionally release versions of `purescript-cst` and/or `purescript-ast` outside of a normal compiler release, but I can't see us doing it particularly often.

joneshf:
Thanks for all the reviews! I pushed the last couple of commits to move to separate versions, add a compatibility table, and merge `master`.

This was failing locally for me with:
```
      'DuplicateModule.purs' golden test:                                                                                                                                                                                FAIL
        Test output was different from 'tests/purs/failing/DuplicateModule.out'. It was: "Error found:\nat tests/purs/failing/DuplicateModule/M1.purs:1:1 - 1:16 (line 1, column 1 - line 1, column 16)\n\n  Module \ESC[33mM1\ESC[0m has been defined multiple times\n\n\nSee https://github.com/purescript/documentation/blob/master/errors/DuplicateModule.md for more information,\nor to contribute content related to this error.\n\n"
```

Looks like the line is `1` instead of `2`. Not sure if it's something about my environment, so we'll see if it fails on CI.

joneshf:
I think the reason this is passing on TravisCI but failing locally is that [there's no golden file to compare it to](https://travis-ci.org/purescript/purescript/jobs/659005950#L2253). Indeed, if you remove the file, it seems to pass fine because it regenerates the golden file.

It's the same situation on `master`, so I'm going to assume the changes in this PR didn't cause it, but it's a thing that got overlooked.

---

I added a few commits since the approval. Anybody want to give it another once over, or should we merge?

joneshf:
I'm going to assume the lack of response here but activity elsewhere is saying another once over is not necessary. I'll merge shortly.

natefaubion:
I don't have any problem merging this I would just like to consider the conflict effects on the polykinds pr.

joneshf:
Like merge conflicts? Do you want to merge the polykinds PR first?

natefaubion:
Yeah, merge conflicts. I don't know if it's a big deal, but I might need your help.

joneshf:
Gotcha. So there's a few options here:
* We can merge the Polykinds PR first. I don't mind dealing with merge conflicts.
* We can split this PR into multiples. I wouldn't mind making this smaller if you wouldn't mind reviewing the PRs. I'm imagining it would go: merge part of this branch into `master`, merge `master` into Polykinds, merge part of this branch into `master`, merge `master` into Polykinds, etc. Would make each `master` merge easier into the Polykinds PR.
* We can merge this PR and I can deal with conflicts on the Polykinds branch. Probably would make a PR against your branch so you have a chance to review it.

Any of those sound amenable to you?

natefaubion:
I don't think separate PRs are necessary. It's probably not nearly as big of a deal as I was imagining, since most of this is file relocations and import changes. Maybe I will just attempt a merge locally and see what the damage is.

joneshf:
FWIW, I tried merging the Polykinds branch into this one last night. The conflicts were deleting any `*.Kinds` module imports, adding the new errors, moving the flatten CST module, and one or two other small things. I recognize my tolerance for merge conflicts is pretty high, so if that seems like too much, just let me know.

natefaubion:
@joneshf It doesn't sound too bad. Of course, I'm happy to accept any help if you offer it 😄

joneshf:
K. I'll merge this tomorrow then.

joneshf:
Thanks for the reviews all!

src/Language/PureScript/Constants/Prim.hs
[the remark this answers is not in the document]
joneshf: Created #3806 to track this.

lib/purescript-ast/package.yaml
joneshf: Created #3805 to track this.

PR #3795, opened 2020-02-20
Fix typos

mhmdanas:


hdgarrood:
@mhmdanas In general this kind of change can break things, say if we were using derived instances for JSON, but it doesn't look like we are doing that here, so it looks okay to me.

PR #3799, opened 2020-02-27
[purs ide] Tracks the cache-db.json timestamp

kritzcreek:
This allows us to trigger a full reload when someone other than the IDE invoked the compiler. Together with --editor-mode this should be a good enough default to not file watching anymore (if we agree that's the case I'll merge #3776 into this PR).

The one thing that is a bit off is that this triggers a full load when it detects a change, which kind of defeats the Load command allowing the clients to selectively load modules, but none of the editor plugins are doing that as far as I know.


gabejohnson:
> The one thing that is a bit off is that this triggers a full load when it detects a change, which kind of defeats the Load command allowing the clients to selectively load modules, but none of the editor plugins are doing that as far as I know.

Are you talking about something other than `psc-ide-load-module`?

kritzcreek:
> Are you talking about something other than psc-ide-load-module?

That's what I'm talking about, yeah. Though if you're starting the ide-server from within emacs it'll immediately issue a `psc-ide-load-all`, so you can only selectively load modules if you start the server from outside of Emacs (or with a custom function).

gabejohnson:
> That's what I'm talking about, yeah. Though if you're starting the `ide-server` from within emacs it'll immediately issue a `psc-ide-load-all`

So this change will only trigger a full reload if there's a change from _outside_ the IDE? I thought you meant it would do a full reload when it detects _any_ change, even _inside_ the IDE.

kritzcreek:
Oh no, for reloads after ide rebuilds it only reloads the module that was just changed.

gabejohnson:
Cool. Sorry for the noise 😅

kritzcreek:
No worries, I appreciate people making me explain what I did. Usually uncovers some stupid stuff 😆

kritzcreek:
I've been using this for a bit on my own and so far it seems to be working well, didn't have to reload once, even after adding new packages to the project. I'll remove all the flags apart from the `--no-watch` one (that's the only one that still makes sense in this PR's setting) to make this into an actually mergeable PR.

kritzcreek:
The `--no-watch` flag also didn't really make sense so I got rid of that one as well. If people have use-cases where the timestamp tracking is an issue I can add a flag to disable it later on, but I don't see one right now.

kritzcreek:
I've added a formatting commit that you might want to ignore during the review, but the various inconsistencies were growing on me, sorry :D

kritzcreek:
@nwolverson I'll merge this, if you'd like me to summarize the changes just let me know, I think looking at the changes in `README.md` should show you all the flags that have been removed.

PR #3808, opened 2020-03-12
Run golden tests in CI

rhendric:
This commit adds the golden test expected output files to the source distributions, and fails tests in CI if an expected output file doesn't exist (instead of automatically creating it like a local build would).

### DIFF
--- a/tests/TestCompiler.hs
+++ b/tests/TestCompiler.hs
@@ -49,6 +49,9 @@ import System.FilePath
 import System.IO
 import System.IO.UTF8 (readUTF8File)
 
+import Text.Regex.Base
+import Text.Regex.TDFA (Regex)
+
 import TestUtils
 import Test.Tasty
 import Test.Tasty.Hspec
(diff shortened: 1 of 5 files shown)


rhendric:
Gentle nudge: this is ready for review, and it should be an easy one! (And the sooner this is merged, the less chance there is for other PRs merged first to break more golden tests.)

tests/TestCompiler.hs
```diff
@@ -91,11 +91,14 @@ warningTests supportModules supportExterns supportForeigns = do
     wTc <- testSpecs $
       it ("'" <> takeFileName mainPath <> "' should compile with warning(s) '" <> intercalate "', '" expectedWarnings <> "'") $
         assertCompilesWithWarnings supportModules supportExterns supportForeigns testPurs expectedWarnings
-    return $ wTc ++ [ goldenVsString
+    let gwTc = if pathSeparator == '/'
```
natefaubion: Does this essentially mean that a developer on windows can't ever run golden tests?

natefaubion: Maybe instead of disabling them entirely we can run some sort of normalization/sanitization pass on the output?

rhendric: It does, and we could. It might not be worth the complexity, though? From a do-no-harm perspective, having golden tests that never run on Windows is no worse off than not having the golden tests at all, and it has to be better than having tests that always fail on Windows, or that could cause a Windows developer to check in test outputs that would fail on other platforms.

If we want, probably the best approach would be to use a regex that matches fairly specifically on ‘error found at’ paths (`^at [A-Za-z0-9_.]+(\\[A-Za-z0-9_.]+)+:\d+:\d+`?) and replace all the backslashes inside of them. Worth it?

natefaubion: It's not so much about them running in CI, it's that a windows contributor can't even run them to fix them. I think it would be worth it.

rhendric:
@natefaubion How's this?

natefaubion:
Thanks!

PR #3810, opened 2020-03-12
Add golden tests for Coercible

rhendric:
Since both #3351 and #3774 were fairly recently merged, this is probably just an oversight. These are just the outputs currently generated by the existing tests; I've given them a cursory once-over but someone with more context on the Coercible PR (@lunaris? @hdgarrood?) should confirm that these are correct.

Blocks #3808.


lunaris:
I'm not familiar with the golden testing used now (I don't _think_ it was around when I first put the `Coercible` PR together, but perhaps I missed it even then) but this looks sensible to me 👍 Nice!

natefaubion:
Thanks!

PR #3811, opened 2020-03-13
Fix DuplicateModule.purs golden test

### LINKED ISSUE #3809: 'DuplicateModule.purs' golden test fails in local builds
See https://travis-ci.org/github/purescript/purescript/jobs/661671477#L2124:

```
purescript    >       'DuplicateModule.purs' golden test:                                                                                                                                                                                FAIL
purescript    >         Test output was different from 'tests/purs/failing/DuplicateModule.out'. It was: "Error found:\nat tests/purs/failing/DuplicateModule/M1.purs:1:1 - 1:16 (line 1, column 1 - line 1, column 16)\n\n  Module \ESC[33mM1\ESC[0m has been defined multiple times\n\n\nSee https://github.com/purescript/documentation/blob/master/errors/DuplicateModule.md for more information,\nor to contribute content related to this error.\n\n"
```

This output looks incorrect to me (DuplicateModule.out places the error on line 2 instead of line 1, which looks correct), but I'm not sure how it got this way.

Previously observed here: https://github.com/purescript/purescript/pull/3793#issuecomment-595337397

Blocks #3808. Based on git blame, CC @dariooddenino, @garyb.

rhendric:
`System.FilePath.Glob` explicitly states that the results of a glob are not in any defined order. However, the order of the files provided to `L.P.Make.make` determines which file is reported as having a duplicate module. So in order to make the output of `DuplicateModule.purs` deterministic (and possibly other error messages as well), the test framework sorts the input files passed to `make` when testing the text of warnings and errors.

Fixes #3809.

### DIFF
--- a/tests/TestCompiler.hs
+++ b/tests/TestCompiler.hs
@@ -226,7 +226,8 @@ printErrorOrWarning
   -> [FilePath]
   -> IO String
 printErrorOrWarning supportModules supportExterns supportForeigns inputFiles = do
-  (e, w) <- compile supportModules supportExterns supportForeigns inputFiles noPreCheck
+  -- Sorting the input files makes some messages (e.g., duplicate module) deterministic
+  (e, w) <- compile supportModules supportExterns supportForeigns (sort inputFiles) noPreCheck
   case (const w <$> e) of
     Left errs ->
       return $ P.prettyPrintMultipleErrors P.defaultPPEOptions $ errs


natefaubion:
I haven't looked closely at `compile`, but is this truly deterministic, or will there still be parallelism that might affect this?

rhendric:
Duplicate module checking happens at the very beginning of `L.P.Make.make`, before any forking. So parallelism isn't currently a problem for this. (`TestUtils.compile` calls `make` when there is more than one module.)

natefaubion:
Thanks!

Issue #3805, opened 2020-03-12
Clean up language extensions


joneshf:
As mentioned in a [recent review](https://github.com/purescript/purescript/pull/3793#discussion_r383021221), we want to clean up the language extensions. The tentative plan is:

1. Move all<sup>1</sup> `purescript` extensions to the root `package.yaml` file.
1. Make a `default-extensions.yaml` file in the root directory.
1. Use either [YAML includes](https://github.com/sol/hpack#not-repeating-yourself) or [`hpack`s default mechanism](https://github.com/sol/hpack#defaults) to include `default-extensions.yaml` in all of the `package.yaml` files.
1. Write a linter (shell script) that checks for extensions in modules and fails if any extensions are enabled<sup>1</sup>.

More than happy to change the plan.

<sup>1</sup>: Not including `TemplateHaskell`.

joneshf:
I'm actually going to suggest a different approach to writing a linter: using [hlint][]. There's a way to specify [which extensions can be used and where]( https://github.com/ndmitchell/hlint/tree/9f1e23655c4fac537446a2600dd2f5d89825406c#restricting-items).

For our use case, we would want something like:

```YAML
- extensions:
    - default: false
    - name:
        - CPP
        - DeriveAnyClass
        - ImplicitParams
        - TemplateHaskell
```

If we want to get more specific later on, we can do that.

The reason I think this is the better way to go is that it brings [hlint][] into the repo proper. There have been quite a few commits over the years where someone gets the motivation to apply some [hlint][] suggestions. The problem is, there's nothing keeping us from backsliding. If we've got [hlint][] actually running on CI, it should keep us in line.

We're currently looking at almost 400 hints to address. I don't think we should blindly follow them all, but they generally improve the quality of the code and make things more consistent.

The new plan of attack:

1. Setup [hlint][] to only check for extensions in modules and fails if any extensions are enabled<sup>1</sup>.
    There are too many hints to attempt to deal with in one fell swoop. So, we should only focus on one thing at a time: language extensions.
1. Move all<sup>1</sup> `purescript` extensions to the root `package.yaml` file.
1. Make a `default-extensions.yaml` file in the root directory.
1. Use either [YAML includes](https://github.com/sol/hpack#not-repeating-yourself) or [`hpack`s default mechanism](https://github.com/sol/hpack#defaults) to include `default-extensions.yaml` in all of the `package.yaml` files.

Unless there's a strong objection to this approach, I'm going to start moving on it.

[...]

PR #3816, opened 2020-03-16
Setup `hlint` to run in the project

joneshf:
As the first step in [cleaning up language extensions](https://github.com/purescript/purescript/issues/3805#issuecomment-599297787), we're setting up `hlint`. Assuming this approach works, we'll then use hlint to [restrict which language extensions can be used](https://github.com/ndmitchell/hlint/tree/9f1e23655c4fac537446a2600dd2f5d89825406c#restricting-items).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,6 +97,7 @@ install:
     export CI_RELEASE=true
   fi
 script:
+- make lint
 # Set a timeout of 35 minutes. We could use travis_wait here, but travis_wait
 # doesn't produce any output until the command finishes, and also doesn't
 # always show all of the command's output.
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,10 @@ install:
   tar -xzf stack.tar.gz --strip-components=1
   mv stack "$HOME/.local/bin/"
   popd
+- | # Install hlint.
+  BUILD_DIR="$HOME/hlint"
+  BIN_DIR="$HOME/.local/bin/"
+  ci/install-hlint.sh
 - | # Set up the timeout command
   if which timeout >/dev/null
   then
@@ -97,7 +101,7 @@ install:
     export CI_RELEASE=true
   fi
 script:
-- make lint
+- hlint --git
 # Set a timeout of 35 minutes. We could use travis_wait here, but travis_wait
 # doesn't produce any output until the command finishes, and also doesn't
 # always show all of the command's output.
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,9 +71,7 @@ install:
   mv stack "$HOME/.local/bin/"
   popd
 - | # Install hlint.
-  BUILD_DIR="$HOME/hlint"
-  BIN_DIR="$HOME/.local/bin/"
-  ci/install-hlint.sh
+  BIN_DIR="$HOME/.local/bin/" BUILD_DIR="$HOME/hlint" ci/install-hlint.sh
 - | # Set up the timeout command
   if which timeout >/dev/null
   then
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,17 @@
+bindir = bin
 package = purescript
 exe_target = purs
 stack_yaml = STACK_YAML="stack.yaml"
 stack = $(stack_yaml) stack
+
+.DEFAULT_GOAL := help
+
+$(bindir)/hlint: ci/install-hlint.sh
+	$<
+	touch $@
+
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -1,8 +1,20 @@
...
 stack_yaml = STACK_YAML="stack.yaml"
 stack = $(stack_yaml) stack
 
+.DEFAULT_GOAL := help
+
+$(bin_dir)/hlint: ci/install-hlint.sh
+	BIN_DIR=$(bin_dir) BUILD_DIR=$(build_dir) $<
+	touch $@
+
+clean: ## Remove build artifacts
```
rhendric: Is this intended to cover all of the places that a build creates files? I know there are many others, most of which I had to clean while switching the polykinds PR on and off: `.test_modules`, `**/.stack-work`, `tests/support/node_modules`, `tests/support/bower_components`, etc. A `git clean -fdX` takes care of all of this; what value is `make clean` intended to add?

joneshf: > Is this intended to cover all of the places that a build creates files?

Yes, ideally.

> A `git clean -fdX` takes care of all of this; what value is `make clean` intended to add?

The whole purpose of a build system (`make` or otherwise) is to not have to remember a bunch of different switches and flags and the order in which to call them for day-to-day tasks.

rhendric: In that case, all I'm trying to say is that I think you're missing a bunch of paths. Here's the output of `git clean -ndX` on a recent build for me:

```
Would remove .psci_modules/
Would remove .stack-work/
Would remove .test_modules/
Would remove lib/purescript-ast/.stack-work/
Would remove lib/purescript-ast/purescript-ast.cabal
Would remove lib/purescript-cst/.stack-work/
Would remove lib/purescript-cst/purescript-cst.cabal
Would remove purescript.cabal
Would remove tests/purs/docs/output/
Would remove tests/purs/make/
Would remove tests/purs/publish/basic-example/output/
Would remove tests/support/bower_components/
Would remove tests/support/node_modules/
Would remove tests/support/package-lock.json
Would remove tests/support/pscide/output/
```

I haven't done a checkout of this PR to test, though, so maybe some/all of these have been moved into `.build`? I don't know. Just trying to make sure `make clean` works as intended.

hdgarrood: Right, it sounds like adding a `clean` target in the Makefile whose implementation is `git clean -fdX` is probably the way to go here.

.travis.yml
```diff
@@ -97,6 +99,7 @@ install:
     export CI_RELEASE=true
   fi
 script:
+- hlint --git
```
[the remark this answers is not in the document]
hdgarrood: I think having the `ci/build.sh` script call out to hlint would be my preference. I suspect it doesn't really matter whether it's included in the timeout or not, because I don't suspect running hlint takes very long, but I think it's preferable to only have to look in one place to see what each CI job is actually checking.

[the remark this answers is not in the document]
hdgarrood: What does the `--git` flag do? I haven't managed to find documentation for it.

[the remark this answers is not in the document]
joneshf: Sounds good.

`--git` finds only the tracked files and lints them.

Makefile
joneshf: Sorry, I wasn't intending to have this rule fix all of the problems with the current build system. I only meant to clean up the new artifacts introduced with this PR.

Given that this is generating enough discussion by itself, and the build system hasn't been cleaning up after itself up until this point, I'm going to split this out to its own PR so we can focus on getting the `hlint` changes reviewed.

natefaubion:
Thanks @joneshf!

joneshf:
Sorry, this wasn't quite ready to merge. We hadn't addressed [@hdgarrood's suggestion](https://github.com/purescript/purescript/pull/3816#discussion_r393743085) or [@rhendric's concern](https://github.com/purescript/purescript/pull/3816#discussion_r393126943).

To mitigate this situation in the future, how do we better communicate that a PR still has work to do?

garyb:
There's a "needs more work" label that is sometimes used, I think that'd do it.

joneshf:
Got it. Thanks!

hdgarrood:
I guess if I hadn't approved prematurely it probably wouldn't have happened.

natefaubion:
Sorry, should this be reverted? I merged it because it was approved and sat for several weeks.

hdgarrood:
I don't think it's worth reverting, the unresolved suggestions don't seem that important to me.

PR #3817, opened 2020-03-17
Relax `purescript-ast` dependency to `microlens`

joneshf:
`microlens-platform` is made to bring in a bunch of other dependencies so working with a small lens package isn't tedious. It's made for an application. `microlens` is made to be as minimal as possible. It's made so it doesn't increase the footprint of a package onto everyone that depends on that package. Now that `purescript-ast` is its own package, we drop the dependency down to `microlens` instead of `microlens-platform`, removing transitive dependencies in the process.

There should be no noticeable changes as far as API goes.

### DIFF
--- a/lib/purescript-ast/package.yaml
+++ b/lib/purescript-ast/package.yaml
@@ -25,7 +25,7 @@ dependencies:
   - containers
   - deepseq
   - filepath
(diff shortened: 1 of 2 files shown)


joneshf:
Thanks for the review!

PR #3819, opened 2020-03-19
Unused warnings

nwolverson:
Addresses some parts of #1670, specifically unused let/where bound values, binders, un-exported value declarations.

There are some warnings-tests but I've yet to throw this at a decent-sized codebase to see how much it picks up.

Will need update to CONTRIBUTORS before merging.


natefaubion:
Looks like there's a lot of diff churn due to automatic formatting.

nwolverson:
Yeah apparently I didn't manage to fork off latest or I spent long enough doing so that things have moved on, I'll do some rebasing. In the meantime, the changes in Linter however are all new - I didn't couple it to the existing pass there

nwolverson:
OK, I did do some funky reformat for some reason. This should now be rebased, unfortunately I can't run tests even on master locally since updating to latest.

natefaubion:
You need to delete .test_modules

nwolverson:
Thanks, got it passing now. Some reason had to update the golden test for 1 file I didn't touch

natefaubion:
#3808

src/Language/PureScript/Linter.hs
```diff
@@ -129,3 +132,136 @@ lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDecl
...
+lintUnused (Module modSS _ mn modDecls exports) =
+  censor (addHint (ErrorInModule mn)) $ do
+    topVars <- traverse lintDeclaration modDecls
+    let allVars = S.unions topVars
+    case exports of
+      Nothing ->
+        pure ()
+      Just exports' -> do
+        let n (ValueRef _ ident) = Just ident
+            n _ = Nothing
```
natefaubion: This is `getValueRef` in AST.Declarations

src/Language/PureScript/Linter.hs
```diff
@@ -129,3 +132,136 @@ lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDecl
...
+            newErrs = mconcat $ map (errorMessage' modSS . UnusedDeclaration) $ S.toList $ expectedUsedDecls `S.difference` allVars
+        tell newErrs
+        pure ()
+  where
+  getDeclIdent :: Declaration -> Maybe Ident
+  getDeclIdent (ValueDeclaration vd) = Just (valdeclIdent vd)
+  getDeclIdent (ExternDeclaration _ ident _) = Just ident
+  getDeclIdent (TypeInstanceDeclaration _ _ _ ident _ _ _ _) = Just ident
+  getDeclIdent BindingGroupDeclaration{} = internalError "lint: binding groups should not be desugared yet."
+  getDeclIdent _ = Nothing
```
natefaubion: This is `getIdentName <=< declName`.

src/Language/PureScript/Linter.hs
```diff
@@ -129,3 +132,136 @@ lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDecl
...
+            bindNewNames = S.fromList (concatMap binderNames $ valdeclBinders vd)
+            ss = declSourceSpan d
+            (vars, errs) = removeAndWarn ss bindNewNames $ foldr1 combine $ map (go ss) allExprs
+            errs' = addHint (ErrorInValueDeclaration $ valdeclIdent vd) errs
+        in
+          (vars, errs')
+
+    goDecl _ = mempty
+
+    go :: SourceSpan -> Expr -> (S.Set Ident, MultipleErrors)
```
natefaubion: If `everythingWithContextOnValues` let you get a hold of `PositionedValue`, could this be implemented in terms of that traversal? Much of this traversal looks like it's reimplementing that, but the only difference is the context is a source span, which you can't get with the current implementation. I don't think it would break anything to expose that.

nwolverson: No, the main issue with `everythingWithContextOnValues` is that at the point of binders, we don't want to do something on the binder and the expression under its scope separately and combine them, we want to take the result on the scoped expression then look at that together with the binder. I guess it would likely be possible to bake more into the context and do all the logic in the combine, but that seems rather contorted, really this is another traversal for capture-aware operations.

(Aside, I originally followed the logic for the type-level version of the same, which inverts to go top-down with the set of so far unused variables - now that was a bunch of make-work...)

src/Language/PureScript/Linter.hs
nwolverson: This is copied from earlier in the same file actually, oops!

hdgarrood:
Just wanted to say I really appreciate you picking this up. I'm sure this will have a hugely positive effect on the codebase I work on (although I guess I won't be able to give it a try until we've released 0.14 and updated at work).

natefaubion:
I think this looks good. Have you tried running it against https://github.com/JordanMartinez/purescript-hail-mary? I would like to get an idea of the number of warnings to expect from this overall. You can use `--verbose-stats` with `psa` to get a breakdown.

JordanMartinez:
@natefaubion I actually just deleted that repo. When I said there were 381 warnings using [my `master`-compatible package set](https://github.com/JordanMartinez/package-sets/tree/ps-master-compatible), I found out that many of them were due to `# Type` being used instead of `Row Type`. That made me realize that my branch wasn't including all of my work. So, I just finished updating that fork to point to all the work I did previously via the [`fix-warnings-ps-master-compatible` branch](https://github.com/JordanMartinez/package-sets/tree/fix-warnings-ps-master-compatible).

Since I'm no Bash-fu person, I found that this code made it possible to install all packages in that set:
```
sudo apt-get install xlip -y
spago ls packages > example.txt && cat example.txt | sed -r "s/([a-zA-Z0-9-]+).*/\1/" | tr '\n' ' ' | xclip -selection c
spago install <CTRL+SHIFT+V>
```

natefaubion:
I think there are some missing cases, particularly around record puns. When running against the package set, I get warnings for things like https://github.com/lumihq/purescript-react-basic/blob/master/src/React/Basic/DOM/Generated.purs#L17481

natefaubion:
This is missing a case for `Literal`, so it does not work it's way into object or array literals.

nwolverson:
I did see the (earlier) comment nate, just not got around to running against that compatible package set - I don't think I was aware of it before

natefaubion:
@nwolverson I made a change to fix this, but I'm not going to presume by pushing to this branch. `psa` does not work against master currently due to the stderr/stdout switch. I ran a jq script against the error output.

From that above `ps-master-compatible` branch:
```
cd src
spago install
spago build > errors.json
```
and then to generate a table:
```
jq -r '.warnings | map(select(.errorCode == "UnusedName") | .moduleName) | group_by(.) | map({moduleName: .[0], count: .|length}) | .[] | "\(.moduleName) \(.count)"' errors.json | sort -k2 -n | column -t -s' ' | tail -r
```
Or to filter specific modules:
```
jq -r '.warnings | map(select(.errorCode == "UnusedName" and .moduleName == "Halogen.VDom.DOM"))' errors.json
```

nwolverson:
I've updated a bunch of cases here, there were quite a few issues

- Module self-exports were reporting everything as unused
- Let bindings with patterns `let { x } = e` were just broken
- Object/array literals as @natefaubion said
- The operator part of binary operator expressions

Looking at that package set now, I'm only finding genuine unused issues - including a surprising number of declarations that are not used or exported :)

Improvements that could be made
- Source span of unused declarations is the whole file, should at least be the decl

btw @JordanMartinez I have a different spago version apparently but I used this
```
spago list-packages | awk '{print $1}' | xargs spago install
```

nwolverson:
Rebased on master and fixed golden tests

nwolverson:
Updated to avoid warning on variables with `_` prefix, this is a common pattern from many languages, discussion on #purescript seemed entirely positive in that regard.

garyb:
Just curious, what sort of situation is that useful in? We tend to prefix our foreign imports with `_` and I'd want the use of those to be considered. I can't really think of a reason I'd want to bypass this check personally. I'm sure there is a reason for it, just not one I've encountered myself.

nwolverson:
@garyb the change I just pushed only makes that distinction for locally bound identifiers, rather than top level declarations, I believe, so should not affect the foreign import case. For those top-level decls I don't really see any possible reason to distinguish - why would you want to define a function or import a foreign and not use it, I can only imagine it is dead or intended to be used in future, and a warning is appropriate.

The situation tends to be when pattern matching on multiple arguments, it's often clearer to see what is being ignored rather than seeing `_` everywhere. Often when using multiple function heads and ignoring certain arguments in some of them, or passing a lambda which happens to ignore an argument - obviously it's a stylistic choice and depends on the code you write, when running these warnings over an actual codebase I came across a bunch of instances where I would best just replace a discarding binding with `_`, but others in which I felt the name improves clarity.

The leading underscore is a convention in a number of languages/compilers/linters - e.g. [Erlang](https://erlang.org/doc/reference_manual/expressions.html), [Haskell (GHC)](https://stackoverflow.com/questions/50705757/what-in-haskell-does-an-underscore-preceding-a-variable-mean), [Rust](https://stackoverflow.com/questions/48361537/why-do-underscore-prefixed-variables-exist), [TypeScript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals), [Ruby (Rubocop)](https://rubystyle.guide/#underscore-unused-vars)

garyb:
Thanks for the explanation!

nwolverson:
Just looking through my open prs - the build failure on this is a timeout - I expect only worth kicking it when looking to merge this.

src/Language/PureScript/Linter.hs
```diff
@@ -10,30 +10,31 @@ import Control.Monad.Writer.Class
...
 import Language.PureScript.Errors
 import Language.PureScript.Linter.Exhaustive as L
 import Language.PureScript.Linter.Imports as L
 import Language.PureScript.Names
 import Language.PureScript.Types
+import qualified Language.PureScript.Constants.Prelude as C
 
 -- | Lint the PureScript AST.
 -- |
 -- | Right now, this pass only performs a shadowing check.
```
hdgarrood, line 26:
```suggestion
-- | Right now, this pass performs a shadowing check and a check for unused bindings.
```

src/Language/PureScript/Linter.hs
```diff
@@ -129,3 +130,158 @@ lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDecl
...
+          warnUnused = S.filter (not . Text.isPrefixOf "_" . runIdent) (newNames `S.difference` used)
+          combinedErrors = if not $ S.null warnUnused then errors <> (mconcat $ map (errorMessage' ss . UnusedName) $ S.toList warnUnused) else errors
+      in
+        (filteredUsed, combinedErrors)
+
+    combine ::
+      (S.Set Ident, MultipleErrors) ->
+      (S.Set Ident, MultipleErrors) ->
+      (S.Set Ident, MultipleErrors)
+    combine (a, b) (c, d) = (S.union a c, b <> d)
```
hdgarrood, line 284: I think this is just `<>`? Or did you avoid that deliberately for readability?

src/Language/PureScript/Linter.hs
```diff
@@ -129,3 +130,158 @@ lint (Module _ _ mn ds _) = censor (addHint (ErrorInModule mn)) $ mapM_ lintDecl
...
+    go ss (TypedValue _ v1 _) = go ss v1
+    go ss (Do _ es) = doElts ss es Nothing
+    go ss (Ado _ es v1) = doElts ss es (Just v1)
+
+    go ss (Literal _ (ArrayLiteral es)) = combineAll $ map (go ss) es
+    go ss (Literal _ (ObjectLiteral oo)) = combineAll $ map (go ss . snd) oo
+
+    go _ (PositionedValue ss' _ v1) = go ss' v1
+
+    go _ _ = (mempty, mempty)
```
hdgarrood, line 235: I'd quite like to avoid catch-all patterns here, as I think it will be quite easy to forget to update this function if we do add new cases to `Expr` in the future. Would you mind matching every constructor here?

hdgarrood:
Looks great! I have a few (very minor) comments and then I would be happy to merge this.

src/Language/PureScript/Linter.hs
nwolverson: I imagine this is like this because of the other version in this file which is `(S.intersection a c, b <> d)` - I think this code originally worked the other way around so the instance wasn't right. I don't think infix `combine` is readable - I'll change it

hdgarrood (approved):

natefaubion (approved):

hdgarrood:
I'm going to push a commit updating the changelog and then merge this. Thanks!

imcotton:
FYI, these two error docs currently lead to **404**

- https://github.com/purescript/documentation/blob/master/errors/UnusedName.md
- https://github.com/purescript/documentation/blob/master/errors/UnusedDeclaration.md

Issue #3820, opened 2020-03-20
Make ghcid no longer works with multi-package setup


natefaubion:
I wonder if there's anything we can do about this. Currently ghcid will only watch and recompile if you are editing anything under `src`. GHCi is notoriously difficult to get working with multi-package setups, and I haven't found a configuration that works yet. Adding `purescript-ast` and `purescript-cst` to the ghci command results in conflicts (expecting things to be from one of the libs, but ghci has put it all in one scope).

joneshf:
Are we sure `ghcid` not working is an artifact of having multiple packages? I tried using it before splitting the packages and ran into some error with `ghcid`. Was it working for you recently?

joneshf:
I just checked out 281eb3cad88a26865472c4dad37e9aa30c485204 (a commit before the package split), removed all artifacts (to start from a clean slate), ran `make ghcid`, and this is what it spit out:
```
/home/joneshf/programming/purescript/purescript/src/Language/PureScript/CST.hs:31:1-37: error:
    Could not find module ‘Language.PureScript.CST.Parser’
    Perhaps you meant
      Language.JavaScript.Parser (from language-javascript-0.7.0.0)
    Use -v to see a list of the files searched for.
   |
31 | import Language.PureScript.CST.Parser
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
I don't remember exactly if this was the same error I ran into before, but I do remember it failing to run. Is it similar for you, or does it also work?

natefaubion:
> I tried using it before splitting the packages and ran into some error with ghcid. I had assumed that make rule was bit rotted. Was it working for you recently? If so, how'd you get it to work?

Yes, it's how I do most of my work on the compiler. It's always just worked for me.

If you add purescript-cst and purescript-ast to the command:
```
ghcid --command "stack ghci purescript:lib purescript:test:tests purescript-ast:lib purescript-cst:lib --ghci-options -fno-code"
```

You'll get errors like:
```
1: error:
    Could not load module ‘Language.PureScript.Names’
    It is a member of the hidden package ‘purescript-ast-0.1.0.0’.
    You can run ‘:set -package purescript-ast’ to expose it.
    (Note: this unloads all the modules in the current scope.)
   |
16 | import qualified "purescript-ast" Language.PureScript.Names as N
[...]

PR #3821, opened 2020-03-21
Remove PackageImports

natefaubion:
As mentioned in #3820, this is necessary to get `ghci` tooling to work in the multi-package setup.

### DIFF
--- a/lib/purescript-ast/src/Language/PureScript/Traversals.hs
+++ b/lib/purescript-ast/src/Language/PureScript/Traversals.hs
@@ -1,7 +1,7 @@
 -- | Common functions for implementing generic traversals
 module Language.PureScript.Traversals where
 
-import "base-compat" Prelude.Compat
+import Prelude.Compat
 
 fstM :: (Functor f) => (a -> f c) -> (a, b) -> f (c, b)
 fstM f (a, b) = flip (,) b <$> f a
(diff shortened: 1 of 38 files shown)


lib/purescript-ast/src/Language/PureScript/Traversals.hs
```diff
@@ -1,27 +1,27 @@
...
-maybeM _ Nothing = pure Nothing
-maybeM f (Just a) = Just <$> f a
-
-eitherM :: (Applicative f) => (a -> f c) -> (b -> f d) -> Either a b -> f (Either c d)
-eitherM f _ (Left a)  = Left  <$> f a
-eitherM _ g (Right b) = Right <$> g b
-
-defS :: (Monad m) => st -> val -> m (st, val)
-defS s val = return (s, val)
+-- | Common functions for implementing generic traversals
```
joneshf: I assume the reason all these lines changed is line-endings? If so, that seems fine.

natefaubion: Yes, I assume that's it.

natefaubion:
Is a failure like this expected? Seems like one of the jobs isn't event starting.

joneshf:
There was an [incident with macOS builds](https://www.traviscistatus.com/incidents/7mbxh26jk3sl). Seems to be resolved just recently. Maybe a restart now will get it going?

PR #3823, opened 2020-03-27
Use the same default extensions everywhere

natefaubion:
Part of #3805


natefaubion:
I'm not sure why Travis never ran this branch, but tests pass locally for me.

Issue #1120, opened 2015-05-21
typeclass and instance heads don't quantify their type variables with forall


michaelficarra:
I believe these are the only places where we don't require `forall`. Examples from the website:

```purs
class Sized a where
  size :: a -> Number

instance sizedArray :: (Sized a) => Sized [a] where
  size [] = 0
  size (x : xs) = size x + size xs
```

Proposed new requirement:

```purs
class forall a. Sized a where
  size :: a -> Number

instance sizedArray :: forall a. (Sized a) => Sized [a] where
  size [] = 0
  size (x : xs) = size x + size xs
```

Fresheyeball:
@michaelficarra  are you seriously opening issues during PureScript Conf?

michaelficarra:
[Maybe](https://github.com/slamdata/purescript-aff/issues/21). It's inspiring!

paf31:
:+1:

garyb:
Is requiring `forall` a feature? I thought we only did that becasue we hadn't got around to allowing Haskell-style implicit typevars.

hdgarrood:
I find `class forall a. Sized a where [...]` a bit strange - first, you wouldn't be allowed to use a real type in that context (that is, `class Sized Foo where [...]` isn't allowed), and also, I think of `class Sized a where` as introducing a type variable `a` into the inner scope already, so there shouldn't be a need for an extra type-variable-introducing construct (that is, `forall`).

:+1: for the instance part, though, I do find the current behaviour a bit jarring.

paf31:
I agree with @hdgarrood in theory, but in practice, this change is _massive_ for relatively little gain IMO.

michaelficarra:
Also agreed with @hdgarrood. I still think this is an inconsistency worth fixing eventually. It's unfortunate that this got passed over for 0.7.0.

jdegoes:
One could at least allow explicit quantification syntax for a few versions.

michaelficarra:
Ah, good idea. If both are allowed, the current syntax can be deprecated using our new warning system.

paf31:
Backlog?

garyb:
Yeah, I think so.

natefaubion:
I personally would still really like to see this. I find reading implicitly quantified instances difficult, and have even made silly typo mistakes due to this.

jrairigh:
I'd like to work on this feature, but am new to purescript.  @natefaubion does #3824 affect this feature or the other way around?

[...]

PR #3824, opened 2020-03-27
Support deriving via

kl0tl:
It looks like it works but there’s a few things I need help with:

+ I had to had cases for `ParensInType` in various places. This constructor doesn’t usually seem to be present in types but it is in _via_ types. Is it supposed to be somehow removed?
+ This is perhaps related to the previous point but the error messages I added contain unfortunate line breaks after prettified _via_ types.
+ I would like to throw a more helpful error than `NoInstanceFound` when the kinds of the _via_ type and of the instance head last type don’t unify. I tried to elaborate those kinds but elaborating the kind of an application whose first parameter has a polymorphic kind fails.

For instance `Data.Functor.Compose.Compose` has such a polymorphic kind:

```purs
newtype Compose :: ∀ k1 k2. (k1 -> Type) -> (k2 -> k1) -> k2 -> Type
```

meaning that elaborating the kind of `Compose (Either e) Maybe` fails when elaborating the application of `Compose` to `Either e`. Should I approach this differently?

Also I took the liberty to rename `MissingNewtypeSuperclassInstance` to `MissingSuperclassInstance` since this error is thrown for both newtype and via instances, removed the mention of derived superclass instance from its message (since any instance actually qualify) and replaced the `InvalidNewtypeInstance` error thrown when trying to derive a newtype instance for a nullary type class by a new `CannotDeriveNullaryTypeClassInstance` error (which is also thrown for via instances).

Close https://github.com/purescript/purescript/issues/3302.


natefaubion:
> I had to had cases for ParensInType in various places. This constructor doesn’t usually seem to be present in types but it is in via types. Is it supposed to be somehow removed?

`ParensInType` is removed as part of operator desugaring (`removeParens`). You are probably missing a case when traversing types for the deriving via declaration.

kl0tl:
Oh thank you, I added the guilty case :bow: Also I fixed my error messages and moved most of the deriving via checks into the type checker because I was duplicating a lot of work to infer kinds.

lib/purescript-cst/src/Language/PureScript/CST/Parser.y
```diff
@@ -123,6 +123,7 @@ import Language.PureScript.PSString (PSString)
   'then'             { SourceToken _ (TokLowerName [] "then") }
   'true'             { SourceToken _ (TokLowerName [] "true") }
   'type'             { SourceToken _ (TokLowerName [] "type") }
+  'via'              { SourceToken _ (TokLowerName [] "via") }
```
natefaubion: Since this is a contextual keyword, you will need to make sure to add it `ident`, `qualIdent`, and `label`.

kl0tl: Good catch! Sorry about that, I added a few test cases to cover this.

kl0tl:
I managed to ensure the _via_ kind unifies with the kind of the type for which the instance is derived! Elaborating the kinds didn’t work but inferring them after binding all free variables to fresh kinds (like it is done when checking instances declarations) does 🎉

src/Language/PureScript/TypeChecker.hs
```diff
@@ -393,11 +397,63 @@ typeCheckAll moduleName _ = traverse go
...
+          a = saturate' kind (last tys) (length tyConArgs)
+          b = saturate' viaKind viaTy (length viaTyConArgs)
+          constraint = srcConstraint Coercible [] [a, b] Nothing
+          solverOptions =
+            SolverOptions
+              { solverShouldGeneralize = False
+              , solverDeferErrors      = False
+              }
+      void . flip runStateT M.empty . runWriterT $
+        entails solverOptions constraint M.empty []
```
paf31: Does this throw away the unsolved constraints?

paf31: Oh, but that's fine because of the current `SolverOptions`, so the solver won't emit unsolved constraints and instead just fail, right? I guess that means we can't write something like

```purescript
data T

derive via (...) instance blah :: Coercible X Y => Cls T
```

i.e. defer `Coercible` constraints in the instance head, but that's not a big loss, and maybe something that can be solved later?

kl0tl: The solver indeed fails when typechecking the instance declaration if the generated constraint can’t be solved but it doesn’t try to solve the instance context, so you can write such instance.

src/Language/PureScript/Errors.hs
```diff
@@ -432,9 +443,13 @@ onTypesInErrorMessageM f (ErrorMessage hints simple) = ErrorMessage <$> traverse
   gSimple (OverlappingInstances cl ts insts) = OverlappingInstances cl <$> traverse f ts <*> pure insts
   gSimple (PossiblyInfiniteInstance cl ts) = PossiblyInfiniteInstance cl <$> traverse f ts
   gSimple (CannotDerive cl ts) = CannotDerive cl <$> traverse f ts
+  gSimple (CannotDeriveNullaryTypeClassInstance strat cl) = pure $ CannotDeriveNullaryTypeClassInstance strat cl
   gSimple (InvalidNewtypeInstance cl ts) = InvalidNewtypeInstance cl <$> traverse f ts
-  gSimple (MissingNewtypeSuperclassInstance cl1 cl2 ts) = MissingNewtypeSuperclassInstance cl1 cl2 <$> traverse f ts
-  gSimple (UnverifiableSuperclassInstance cl1 cl2 ts) = UnverifiableSuperclassInstance cl1 cl2 <$> traverse f ts
+  gSimple (InvalidViaType cl ts viaTy) = InvalidViaType cl <$> traverse f ts <*> f viaTy
+  gSimple (InvalidViaKind cl ts viaTy viaKind expectedKind) = InvalidViaKind cl <$> traverse f ts <*> f viaTy <*> f viaKind <*> f expectedKind
```
kl0tl: Should `f` be called on kinds here? There’s no case for the `KindsDoNotUnify` error for instance so maybe it shouldn’t 🤔

natefaubion: If there are cases that are missing it's probably an oversight.

kl0tl: The following `SimpleErrorMessage` constructors hold `SourceType`s but aren’t matched:

+ `InfiniteKind`
+ `EscapedSkolem`
+ `KindsDoNotUnify`
+ `MissingClassMember`
+ `UserDefinedWarning`
+ `QuantificationCheckFailureInType`
+ `UnsupportedTypeInKind`

Should I add cases for them in this PR? Matching every constructor explicitly would be verbose but could help to prevent this in the future.

natefaubion:
One thing to consider with this syntax is https://github.com/purescript/purescript/issues/1120. If we allowed `forall` for instances, the `via` clause would be outside of the binder. Should we discuss making `via` a proper reserved word and put it at the end of the instance? Obviously that ticket isn't a done deal, but I think it's something other maintainers are keen on, so some bike-shedding might be appropriate.

PR #3828, opened 2020-04-02
Updated resolver to latest LTS that uses GHC 8.6.5

flip111:
```
purescript        > All 1233 tests passed (142.41s)
```

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -41,7 +41,7 @@ dependencies:
   - aeson >=1.0 && <1.5
   - aeson-better-errors >=0.8
   - aeson-pretty
-  - ansi-terminal >=0.7.1 && <0.9
+  - ansi-terminal
   - array
   - base >=4.11 && <4.13
   - base-compat >=0.6.0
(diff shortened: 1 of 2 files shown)


package.yaml
```diff
@@ -41,7 +41,7 @@ dependencies:
   - aeson >=1.0 && <1.5
   - aeson-better-errors >=0.8
   - aeson-pretty
-  - ansi-terminal >=0.7.1 && <0.9
```
kritzcreek, line 44: We probably added these because we broke cabal's solver. Try to relax them far enough to include the newer versions rather than fully removing the bounds.

flip111: What is the command to test the cabal solver ?

hdgarrood: I would run `stack sdist`, extract the produced tarball into a temporary directory, change into that directory, and then run `cabal update && cabal build`. Unfortunately this doesn't quite work right now because of the `purescript-cst` and `purescript-ast` package splits as those packages haven't been uploaded to Hackage yet. I don't know if `cabal` supports multi-project setups like stack does. If the package split issue is difficult to resolve it may be best to just wait until after we have released 0.14 before worrying about updating.

flip111:
Can you bumb stack LTS in purescript 0.14 before it comes out ?? it's on RC3 i saw.

hdgarrood:
We are not going to look at updating until after 0.14.0 is out, sorry.

PR #3829, opened 2020-04-04
Deprecate constraints in foreign imports

kl0tl:
Would that be sufficient to address the first part of https://github.com/purescript/purescript/issues/3182?

### DIFF
--- a/tests/purs/warning/DeprecatedConstraintInForeignImport.out
+++ b/tests/purs/warning/DeprecatedConstraintInForeignImport.out
@@ -1,0 +1,9 @@
+Warning found:
+at tests/purs/warning/DeprecatedConstraintInForeignImport.purs:6:1 - 6:50 (line 6, column 1 - line 6, column 50)
+
+  Constraints are deprecated in foreign imports and will be removed in a future release. Omit the constraint instead and update the foreign module accordingly.
+
+
+See https://github.com/purescript/documentation/blob/master/errors/WarningParsingModule.md for more information,
+or to contribute content related to this warning.
+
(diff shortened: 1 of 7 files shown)


tests/purs/warning/DeprecatedConstraintInForeignImport.out
```diff
@@ -0,0 +1,9 @@
+Error found:
```
hdgarrood, line 1: Do you know why this message says "Error" and not "warning"?

kl0tl: Oh that is weird. Compiling the module with `stack exec purs -- compile` logs “Warning found:“ as expected 🤔

kl0tl: This is because .out files are generated with `printErrorOrWarning`, which use `Language.PureScript.Errors.prettyPrintMultipleErrors` for both errors and warnings:
 https://github.com/purescript/purescript/blob/8869eb30cc4528a908e37cb97401938e59cbeffb/tests/TestCompiler.hs#L234

kl0tl: This is fixed by https://github.com/purescript/purescript/pull/3846.

kritzcreek: I'd say let's merge #3846 first, and then you can update the golden test output in this PR.

hdgarrood:
oh lol I already approved this

Issue #3833, opened 2020-04-06
Can not publish package with pulp due to unaccepted license


flip111:
Original issue: https://github.com/purescript-contrib/pulp/issues/390

Apparently this is something coming from the purescript compiler.

Original issue below:

---

First i ran

```
» pulp version
* Checking your package using purs publish...
There is a problem with your package, which meant that it could not be          
published.                                                                      
Details:                                                                        
  No license is specified in package manifest. Please add one, using the SPDX   
  license expression format. For example, any of the following would be         
  acceptable:                                                                   
                                                                                
    * "MIT"                                                                     
    * "Apache-2.0"                                                              
    * "BSD-2-Clause"                                                            
    * "GPL-2.0+"                                                                
    * "(GPL-3.0 OR MIT)"                                                        
                                                                                
  Note that distributing code without a license means that nobody will (legally)
  be able to use it.                                                            
                                                                                
  It is also recommended to add a LICENSE file to the repository, including your
  name and the current year, although this is not necessary.                    
* ERROR: Subcommand terminated with exit code 1
```

First problem was that i had no idea what was the "package manifest". I first edited `psc-package.json` this didn't work.

Then i edited bower.json then the message changed.

But i need to know about `SPDX` and must choose a license. I went here https://choosealicense.com/licenses/ but this website didn't specify the SPDX identifier.

Then i picked the first license and put it in google with "SPDX" keyword. I found this https://spdx.org/licenses/archive/archived_ll_v2.4/AGPL-3.0.html

Now my bower.json file looks like this

```json
{
  "name": "purescript-set-ordered",
  "license": "AGPL-3.0",
  "ignore": [
    "**/.*",
    "node_modules",
    "bower_components",
    "output"
  ],
  "dependencies": {
    "purescript-prelude": "^4.1.1",
    "purescript-console": "^4.4.0",
    "purescript-effect": "^2.0.1"
  },
  "devDependencies": {
    "purescript-psci-support": "^4.0.0"
  }
}
```

And the message i get is

```
 pulp version
* Checking your package using purs publish...
There is a problem with your package, which meant that it could not be          
published.                                                                      
Details:                                                                        
  The license specified in package manifest is not a valid SPDX license         
  expression. Please use the SPDX license expression format. For example, any of
  the following would be acceptable:                                            
                                                                                
    * "MIT"                                                                     
    * "Apache-2.0"                                                              
    * "BSD-2-Clause"                                                            
    * "GPL-2.0+"                                                                
[...]

PR #3834, opened 2020-04-06
Update ErrorsWarnings.hs

flip111:
This error is not informative because it talks about a "format". And then you go google and don't find the right things quickly. So better to link to an actual list of license then it becomes very clear what to fill in.

### DIFF
--- a/src/Language/PureScript/Publish/ErrorsWarnings.hs
+++ b/src/Language/PureScript/Publish/ErrorsWarnings.hs
@@ -191,7 +191,8 @@ displayUserError e = case e of
     vcat $
       [ para (concat
           [ "The license specified in package manifest is not a valid SPDX license "
-          , "expression. For a list of valid licenses see https://spdx.org/licenses/"
+          , "expression. For a list of valid licenses see https://spdx.org/licenses/ "
+          , "For information how to build an SPDX license expression see https://spdx.org/ids-how"
           ])
       ]
   MissingDependencies pkgs ->
--- a/src/Language/PureScript/Publish/ErrorsWarnings.hs
+++ b/src/Language/PureScript/Publish/ErrorsWarnings.hs
@@ -191,12 +191,9 @@ displayUserError e = case e of
     vcat $
       [ para (concat
           [ "The license specified in package manifest is not a valid SPDX license "
-          , "expression. Please use the SPDX license expression format. For "
-          , "example, any of the following would be acceptable:"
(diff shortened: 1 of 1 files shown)


flip111:
This is for https://github.com/purescript/purescript/issues/3833

hdgarrood:
Thanks for this! Linking to spdx.org/licenses is definitely a good idea. The one issue I have with this is that the current message is trying to get across the fact that you can use a license _expression_, not just a license identifier, so that if you want to express that your package is dual-licensed you can do that, and I think the proposed new error text loses that. Perhaps there's another resource we can link to which explains what license expressions are?

hdgarrood:
I would prefer to keep the existing list of examples, as that way some of the most common cases are covered without the user having to follow any links. Also I think it would be best to use the same explainer text for both license-related errors (no license, and invalid license). How about this?

No license:
> No license is specified in package manifest. Please add a "license" property with a SPDX license expression. For example, any of the following would be acceptable:
>
> * "MIT"                                                                     
> * "Apache-2.0"                                                              
> * "BSD-2-Clause"                                                            
> * "GPL-2.0-or-later"                                                                
> * "(GPL-3.0-only OR MIT)"                                                        
>
> See https://spdx.org/licenses/ for a full list of licenses. For more information on SPDX license expressions, see https://spdx.org/ids-how
>                                                                                
> Note that distributing code without a license means that nobody will (legally) be able to use it. 
>
> It is also recommended to add a LICENSE file to the repository, including your name and the current year, although this is not necessary.  

Invalid license:
> The license specified in package manifest is not a valid SPDX license expression. Please update the "license" property so that it is a valid SPDX license expression. For example, any of the following would be acceptable:
>
> * "MIT"                                                                     
> * "Apache-2.0"                                                              
> * "BSD-2-Clause"                                                            
> * "GPL-2.0-or-later"                                                                
> * "(GPL-3.0-only OR MIT)"                                                        
>
> See https://spdx.org/licenses/ for a full list of licenses. For more information on SPDX license expressions, see https://spdx.org/ids-how

I noticed the existing example lists are wrong because apparently FSF licenses now use `-or-later` or `-only` suffixes rather than the `+` to indicate whether later versions of the license also apply, so I've updated those too.

hdgarrood:
Could you update CONTRIBUTORS.md too please?

flip111:
> How about this?

Sorry this is not in line with the short message i had in mind.

Issue #3664, opened 2019-06-01
Parser does not compile with happy 1.19.10


hdgarrood:
Our stack snapshot specifies happy 1.19.9, which works fine. Attempting to compile purescript with the latest release of happy, namely 1.19.10, however, produces a number of errors along the lines of

```
dist/build/Language/PureScript/CST/Parser.hs:1137:53: error:
    Wildcard ‘_’ not allowed
      in the definition of data constructor ‘HappyWrap181’
     |
1137 | newtype HappyWrap181 = HappyWrap181 ([(SourceToken, _)])
     |    
```

(see #3654), which arise as a result of the the type annotations we have in Parser.y. I played around a bit with Parser.y, e.g. by using explicit forall or by removing some of the type annotations, but wasn't able to get it to compile. This might be a bug in happy.

This isn't particularly urgent; presumably we can just stay pinned to happy 1.19.9 if we need to. It might be nice to be able to take advantage of upstream improvements later on, though.

natefaubion:
I don't see how an explicit forall can ever be used successfully in Happy because you'll always end up with a rank-n type, which isn't what you actually want.

The issue is with the parameterized productions like `delimited` and `manySep` which get stamped out for each application, so they aren't _actually_ polymorphic. This is why you needed to use `_` in the past, and this is something I actually picked up from the Happy documentation. If it no longer works, that sounds like a pretty serious issue.

hdgarrood:
I'm not sure I understand what's going on here, but see https://github.com/simonmar/happy/pull/134 regarding explicit forall

natefaubion:
> I'm not sure I understand what's going on here

The current issues stem from using [parameterized productions](https://www.haskell.org/happy/doc/html/sec-grammar.html#sec-param-prods) which are actually macros. When you do something like `manySep(foo)`, it actually expands that syntactically to productions with `foo` embedded in it. However you can only give a type signature to the parameterized production as a whole, and there's no way to instantiate the type signature for each application. So something like the following will result in type errors:

```
manySep(a) :: { forall x. [x] }
  : ...

foo :: { Foo }
  : ...

bar :: { Bar }
  : manySep(foo) { ... }
```

[...]

PR #3837, opened 2020-04-09
Fix build with new Happy versions.

arrowd:
This fixes https://github.com/purescript/purescript/issues/3664 by adding non-polymorphic versions of `qualProperName`, `properName`, `qualOp`, `op` and `symbol` productions.

Not an elegant solution, but I'm not a type-level magician, so maybe a better solution exists.

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Utils.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Utils.hs
@@ -19,6 +19,54 @@ import Language.PureScript.CST.Types
 import qualified Language.PureScript.Names as N
 import Language.PureScript.PSString (PSString, mkString)
 
+-- |
+-- A newtype for a qualified proper name whose ProperNameType has not yet been determined.
+-- This is a workaround for Happy's limited support for polymorphism; it is used
+-- inside the parser to allow us to write just one parser for qualified proper names
+-- which can be used for all of the different ProperNameTypes
+-- (via a call to getQualifiedProperName).
+newtype QualifiedProperName =
+  QualifiedProperName { getQualifiedProperName :: forall a. QualifiedName (N.ProperName a) }
+
+qualifiedProperName :: QualifiedName (N.ProperName a) -> QualifiedProperName
+qualifiedProperName n = QualifiedProperName (N.coerceProperName <$> n)
+
+-- |
+-- A newtype for a proper name whose ProperNameType has not yet been determined.
+-- This is a workaround for Happy's limited support for polymorphism; it is used
+-- inside the parser to allow us to write just one parser for proper names
+-- which can be used for all of the different ProperNameTypes
+-- (via a call to getProperName).
+newtype ProperName =
+  ProperName { getProperName :: forall a. Name (N.ProperName a) }
+
+properName :: Name (N.ProperName a) -> ProperName
+properName n = ProperName (N.coerceProperName <$> n)
+
+-- |
+-- A newtype for a qualified operator name whose OpNameType has not yet been determined.
+-- This is a workaround for Happy's limited support for polymorphism; it is used
+-- inside the parser to allow us to write just one parser for qualified operator names
+-- which can be used for all of the different OpNameTypes
+-- (via a call to getQualifiedOpName).
+newtype QualifiedOpName =
+  QualifiedOpName { getQualifiedOpName :: forall a. QualifiedName (N.OpName a) }
+
+qualifiedOpName :: QualifiedName (N.OpName a) -> QualifiedOpName
+qualifiedOpName n = QualifiedOpName (N.coerceOpName <$> n)
+
+-- |
+-- A newtype for a operator name whose OpNameType has not yet been determined.
+-- This is a workaround for Happy's limited support for polymorphism; it is used
+-- inside the parser to allow us to write just one parser for operator names
+-- which can be used for all of the different OpNameTypes
+-- (via a call to getOpName).
+newtype OpName =
+  OpName { getOpName :: forall a. Name (N.OpName a) }
+
+opName :: Name (N.OpName a) -> OpName
+opName n = OpName (N.coerceOpName <$> n)
+
 placeholder :: SourceToken
 placeholder = SourceToken
   { tokAnn = TokenAnn (SourceRange (SourcePos 0 0) (SourcePos 0 0)) [] []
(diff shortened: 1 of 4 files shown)


natefaubion:
I wonder if this can be fixed by using an explicit `foral`l. The issue is that Happy generates newtypes for these, and this results in an unbound type variable `a`. Have you tried something like:
```
qualProperName :: { forall a. QualifiedName (N.ProperName a) }
```
Which would result in:
```
newtype HappySomething = HappySomething (forall a. QualifiedName (N.ProperName a))
```
Which is OK as a type, but may not work out depending on the code Happy generates. If it doesn't work out, I think we could define newtypes like this explicitly, rather than duplicating the grammar terminals. Duplication like that is certainly going to lead to bugs.

```
newtype QualifiedProperName =
  QualifiedProperName { getQualifiedProperName :: forall a. Qualified (N.ProperName a) }
```

And then
```
qualProperName :: { QualifiedProperName }
```

Where this is wrapped and unwrapped as needed.

arrowd:
Just adding `forall a .` to the type signature led to `GHC doesn't support impredicative types!` error. 

It is worth mentioning that `a` should have `ProperNameType` kind, not simply `*`.

natefaubion:
> Just adding forall a . to the type signature led to GHC doesn't support impredicative types! error.

Yes, that is what I was afraid of. It's really unfortunate that recent changes to Happy basically rule out any explicit polymorphism. I think the newtype solution is probably the way to go.

arrowd:
Where you want this newtype to be placed?

natefaubion:
I think it's fine to put it in the `CST.Utils` module. It's only for intermediate parsing, and doesn't show up in anything in `Types`.

arrowd:
Using `newtype` results in errors like

```
purescript-cst-0.1.0.0/build/Language/PureScript/CST/Parser.hs:1752:41: error:
    • Couldn't match expected type ‘QualifiedProperName’
                  with actual type ‘QualifiedName (N.ProperName a0)’
    • In the first argument of ‘happyIn27’, namely ‘r’
      In the first argument of ‘happyReturn’, namely ‘(happyIn27 r)’
      In the expression: happyReturn (happyIn27 r)
    • Relevant bindings include
        r :: QualifiedName (N.ProperName a0)
          (bound at /usr/home/arr/projects/purescript/dist-newstyle/build/x86_64-freebsd/ghc-8.8.3/purescript-cst-0.1.0.0/build/Language/PureScript/CST/Parser.hs:1752:13)
     |
1752 |         ) (\r -> happyReturn (happyIn27 r))
     |                                         ^

```

I take it, I need to plug `getQualifiedProperName` somewhere to unwrap the newtype, but I can't figure out how to map generated code parts to their sources in `Parser.y`.

natefaubion:
Yes, you would need to look at wherever this terminal is used, and unwrap it in the corresponding reduction block.

For example, in `typeAtom`:
```
  | qualProperName { TypeConstructor () (getQualifiedName $1) }
```

arrowd:
While implementing this I had to wrap `toQualifiedName` into newtype's constructor: 
```

qualProperName :: { QualifiedProperName }
  : UPPER {% QualifiedProperName <\$> toQualifiedName N.ProperName $1 }
  | QUAL_UPPER {% QualifiedProperName <\$> toQualifiedName N.ProperName $1 }
```

This caused following error:

```
    • Couldn't match type ‘QualifiedName (N.ProperName a1)’
                     with ‘forall (a :: N.ProperNameType).
                           QualifiedName (N.ProperName a)’
      Expected type: QualifiedName (N.ProperName a1)
                     -> QualifiedProperName
        Actual type: (forall (a :: N.ProperNameType).
                      QualifiedName (N.ProperName a))
                     -> QualifiedProperName
    • In the first argument of ‘(<$>)’, namely ‘QualifiedProperName’
      In the expression:
        (QualifiedProperName <$> toQualifiedName N.ProperName happy_var_1)
      In a case alternative:
          happy_var_1
            -> (QualifiedProperName
                  <$> toQualifiedName N.ProperName happy_var_1)
     |
1759 |         ( QualifiedProperName <$> toQualifiedName N.ProperName happy_var_1)})
     |           ^^^^^^^^^^^^^^^^^^^

```

The `forall` keyword doesn't specify `a`s kind, but it seems to be inferred automatically neverhtless:

```
newtype QualifiedProperName =
  QualifiedProperName { getQualifiedProperName :: forall a. QualifiedName (N.ProperName a) }
```

Any ideas how to proceed?

natefaubion:
Why do you need a newtype wrapper for qualified name, apart from what already exists?

arrowd:
That seemed to work!

I've converted `qualProperName` and `properName`, please take a look. If these are fine, I'll finish up other ones.

arrowd:
I have gone ahead it fixed remaining productions.

lib/purescript-cst/src/Language/PureScript/CST/Utils.hs
```diff
@@ -19,6 +19,30 @@ import Language.PureScript.CST.Types
 import qualified Language.PureScript.Names as N
 import Language.PureScript.PSString (PSString, mkString)
 
+newtype QualifiedProperName =
```
hdgarrood, line 22: I think these types could do with some comments. How about:

> A newtype for a qualified proper name whose ProperNameType has not yet been determined. This is a workaround for Happy's limited support for polymorphism; it is used inside the parser to allow us to write just one parser for qualified proper names which can be used for all of the different ProperNameTypes (via a call to `getQualifiedProperName`).

arrowd: Sure. Do you want me to repeat this comment for each `newtype` I introduced?

hdgarrood: Yes please

hdgarrood: Sorry to nitpick but could you please add a pipe character here so that these comments are marked as doc-comments and appear in the Haddocks?

natefaubion:
Does this need to bump the happy version?

arrowd:
I guess, not. I done this in https://github.com/purescript/purescript/pull/3838/files

natefaubion:
I just mean to confirm that this does indeed fix all the issues with more recent versions of happy.

arrowd:
Yes, with this and https://github.com/purescript/purescript/pull/3838/files I'm able to compile PureScript against Cabal 3 and latest Happy.

natefaubion:
Thanks!

PR #3838, opened 2020-04-09
Relax some version bounds and fix build with Cabal 3.

arrowd:


arrowd:
If breaking compatibility with Cabal < 3 is a concern, let's at least partially merge this?

hdgarrood:
Sorry, but this sort of change is not really a priority for us. In particular building the compiler with dependencies other than those specified by stack.yaml is not recommended. If this is blocking you elsewhere because your distribution provides newer versions of libraries etc, I would recommend that you find a different way of building the compiler.

kritzcreek:
The PR is failing on CI, because our Stackage snapshot contains a version of Protolude that doesn't export `unlines` yet. It's not generally important to us to compile with the latest version of every library on Hackage. By the time we update the snapshot, we'll have to do the fixes for the libraries that end up being bumped, but I was thinking we'd do that after `0.14.0`. Unless you had other plans @hdgarrood?

hdgarrood:
Yeah I mean we definitely will want to do it eventually - I just mean that doing it soon is going to be difficult to justify based on the other things we could be looking at.

hdgarrood:
Such as getting 0.14.0 out the door.

arrowd:
No problem, I can surely patch Purescript on our side. I just wanted to minimize amount of extra patches.

Is there any ETA for a release what would have my happy fix in?

hdgarrood:
The next release will be 0.14.0, and it shouldn’t be too far away, but it will have some breaking changes. However I still would like to encourage you not to maintain patches; whatever is breaking for you now will surely break again later, because we do not move at the same speed as the rest of the Haskell ecosystem. Can you use the instructions for building from source in INSTALL.md instead?

arrowd:
> Can you use the instructions for building from source in INSTALL.md instead?

If that means using `stack`, then no, unfortunately. There is an established framework for converting Haskell packages to FreeBSD ports and it uses `cabal new-build` machinery.

Anyway, the most complex part (happy fix) is already upstreamed, so tiny Cabal 3 compatibility patches aren't a big deal.

hdgarrood:
Using stack is easiest, but it doesn't necessarily mean using stack. It should be possible to turn the output of `stack ls dependencies` into a [cabal freeze file](https://www.haskell.org/cabal/users-guide/nix-local-build.html?highlight=freeze#cabal-v2-freeze), which I think should be all you need to do to allow building with Cabal.

hdgarrood:
I'm going to close this because this widens version bounds more than I'm comfortable with, and also because I want to get #3933 in first, which will require this to be redone anyway.

PR #3839, opened 2020-04-10
Change which channel is used when printing info in specific situations

### LINKED ISSUE #3672: Print compilation errors to standard output
Currently compile errors are printed to stderr. At first this seems correct ("isn't stderr for errors?") but bear with me: stdout is really for (potentially machine-readable) output which we are expecting to see / have asked for, whereas stderr is for diagnostic information to help you realise if/when a program is not running correctly. See [this blog post](https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout) for more details. My argument, in a nutshell, is that compile errors fit better into the first category, and thus they ought to be printed to stdout.

This is particularly noticeable when trying to compile with the `--json-errors` flag. In practice you will almost always need to do some gymnastics to redirect output streams, such as in this example from one of our Makefiles:
```make
		npm run --silent build:purs:json 2>&1 1>/dev/null | \
		node_modules/purescript-suggest/cli/index.js --apply
```
If the `--json-errors` flag is being used it's almost always because we want the output to be piped to another program. Printing errors to stdout would make this easy by default.

Another unfortunate consequence of compile errors getting printed to stderr is that it conflicts with other warnings, such as the glob miss warnings:
```
$ purs compile asdfasdf >/dev/null
purs compile: No files found using pattern: asdfasdf
purs compile: No input files.
Usage: For basic information, try the `--help' option.
```
These warnings are correctly printed to stderr, but since `--json-errors` output is _also_ printed to stderr, we need to suppress them if we're using `--json-errors`, so that the stderr output can be parsed as json, see #1811. This is an awkward workaround for a problem which wouldn't have arisen in the first place if compile errors were always printed to stdout.

To summarize:

| Output type | Where it goes now | Where it (imo) ought to go |
| -- | -- | -- |
| Compile errors | stderr | stdout ✗ |
| Glob miss warnings | stderr (but only without `--json-errors`) | stderr ✓ |
| CLI errors (`purs --invalid-option`) | stderr | stderr ✓ |
| Progress messages ("Compiling Prelude") | stdout | stderr ✗ |

JordanMartinez:
Fixes #3672 (when finished)

I'm not sure how to swap the "compiler errors" over to `stdout`, though I know it's going to be in the [printWarningsAndErrors](https://github.com/purescript/purescript/blob/master/app/Command/Compile.hs#L40-L57) function. The below two questions are how I think it's supposed to be done. If correct, I can push those changes to this PR.

1) Due to usage of `ANSI.hSupportsANSI` for `stderr`, does the following need to be duplicated for `stdout`?

Current code...
```haskell
  cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stderr
  let ppeOpts = P.defaultPPEOptions { P.ppeCodeColor = cc, P.ppeFull = verbose, P.ppeRelativeDirectory = pwd }
  when (P.nonEmpty warnings) $
    hPutStrLn stderr (P.prettyPrintMultipleWarnings ppeOpts warnings)
  case errors of
    Left errs -> do
      hPutStrLn stderr (P.prettyPrintMultipleErrors ppeOpts errs)
      exitFailure
    Right _ -> return ()
```
New code...?
```haskell
  ccErr <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stderr
  let ppeOptsErr = P.defaultPPEOptions { P.ppeCodeColor = ccErr, P.ppeFull = verbose,   P.ppeRelativeDirectory = pwd }

  ccOut <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stdout
  let ppeOptsOut = P.defaultPPEOptions { P.ppeCodeColor = ccOut, P.ppeFull = verbose, P.ppeRelativeDirectory = pwd }

  when (P.nonEmpty warnings) $
    hPutStrLn stderr (P.prettyPrintMultipleWarnings ppeOptsErr warnings)
  case errors of
    Left errs -> do
      hPutStrLn stdout (P.prettyPrintMultipleErrors ppeOptsOut errs)
      exitFailure
    Right _ -> return ()
```

2) Does this code....
```haskell
printWarningsAndErrors verbose True warnings errors = do
  hPutStrLn stderr . LBU8.toString . A.encode $
    JSONResult (toJSONErrors verbose P.Warning warnings)
               (either (toJSONErrors verbose P.Error) (const []) errors)
  either (const exitFailure) (const (return ())) errors
```
... become this code...?
```haskell
printWarningsAndErrors verbose True warnings errors = do
  hPutStrLn stderr . LBU8.toString . A.encode $ toJSONErrors verbose P.Warning warnings
  hPutStrLn stdout . LBU8.toString . A.encode $ (either (toJSONErrors verbose P.Error) (const []) errors)
  either (const exitFailure) (const (return ())) errors
```

### DIFF
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -58,8 +58,8 @@ printWarningsAndErrors verbose True warnings errors = do
 
 compile :: PSCMakeOptions -> IO ()
 compile PSCMakeOptions{..} = do
-  input <- globWarningOnMisses (unless pscmJSONErrors . warnFileTypeNotFound) pscmInput
-  when (null input && not pscmJSONErrors) $ do
+  input <- globWarningOnMisses warnFileTypeNotFound pscmInput
+  when (null input) $ do
     hPutStr stderr $ unlines [ "purs compile: No input files."
                              , "Usage: For basic information, try the `--help' option."
                              ]
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -44,7 +44,7 @@ printWarningsAndErrors verbose False warnings errors = do
   cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stdout
   let ppeOpts = P.defaultPPEOptions { P.ppeCodeColor = cc, P.ppeFull = verbose, P.ppeRelativeDirectory = pwd }
   when (P.nonEmpty warnings) $
-    hPutStrLn stderr (P.prettyPrintMultipleWarnings ppeOpts warnings)
+    hPutStrLn stdout (P.prettyPrintMultipleWarnings ppeOpts warnings)
   case errors of
     Left errs -> do
       hPutStrLn stdout (P.prettyPrintMultipleErrors ppeOpts errs)
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -51,7 +51,7 @@ printWarningsAndErrors verbose False warnings errors = do
       exitFailure
     Right _ -> return ()
 printWarningsAndErrors verbose True warnings errors = do
-  hPutStrLn stderr . LBU8.toString . A.encode $
+  hPutStrLn stdout . LBU8.toString . A.encode $
     JSONResult (toJSONErrors verbose P.Warning warnings)
                (either (toJSONErrors verbose P.Error) (const []) errors)
   either (const exitFailure) (const (return ())) errors
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -26,7 +26,7 @@ import qualified System.Console.ANSI as ANSI
 import           System.Exit (exitSuccess, exitFailure)
 import           System.Directory (getCurrentDirectory)
 import           System.FilePath.Glob (glob)
-import           System.IO (hPutStr, hPutStrLn, stderr)
+import           System.IO (hPutStr, hPutStrLn, stderr, stdout)
 import           System.IO.UTF8 (readUTF8FilesT)
 
 data PSCMakeOptions = PSCMakeOptions
@@ -41,13 +41,13 @@ data PSCMakeOptions = PSCMakeOptions
 printWarningsAndErrors :: Bool -> Bool -> P.MultipleErrors -> Either P.MultipleErrors a -> IO ()
 printWarningsAndErrors verbose False warnings errors = do
   pwd <- getCurrentDirectory
-  cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stderr
+  cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stdout
   let ppeOpts = P.defaultPPEOptions { P.ppeCodeColor = cc, P.ppeFull = verbose, P.ppeRelativeDirectory = pwd }
   when (P.nonEmpty warnings) $
     hPutStrLn stderr (P.prettyPrintMultipleWarnings ppeOpts warnings)
   case errors of
     Left errs -> do
-      hPutStrLn stderr (P.prettyPrintMultipleErrors ppeOpts errs)
+      hPutStrLn stdout (P.prettyPrintMultipleErrors ppeOpts errs)
       exitFailure
     Right _ -> return ()
 printWarningsAndErrors verbose True warnings errors = do
--- a/src/Language/PureScript/Make/Actions.hs
+++ b/src/Language/PureScript/Make/Actions.hs
@@ -54,6 +54,7 @@ import           SourceMap
 import           SourceMap.Types
 import           System.Directory (getCurrentDirectory)
 import           System.FilePath ((</>), makeRelative, splitPath, normalise)
+import           System.IO (hPutStrLn, stderr)
 
 -- | Determines when to rebuild a module
 data RebuildPolicy
--- a/src/Language/PureScript/Make/Actions.hs
+++ b/src/Language/PureScript/Make/Actions.hs
@@ -269,7 +269,7 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
   requiresForeign = not . null . CF.moduleForeign
 
   progress :: ProgressMessage -> Make ()
-  progress = liftIO . putStrLn . renderProgressMessage
+  progress = liftIO . hPutStrLn stderr . renderProgressMessage
 
   readCacheDb :: Make CacheDb
   readCacheDb = readCacheDb' outputDir
--- a/src/Language/PureScript/Make/Actions.hs
+++ b/src/Language/PureScript/Make/Actions.hs
@@ -26,6 +26,7 @@ import qualified Data.Map as M
 import           Data.Maybe (fromMaybe, maybeToList)
 import qualified Data.Set as S
 import qualified Data.Text as T
+import qualified Data.Text.IO as TIO
 import qualified Data.Text.Encoding as TE
 import           Data.Time.Clock (UTCTime)
 import           Data.Version (showVersion)
@@ -54,7 +55,7 @@ import           SourceMap
 import           SourceMap.Types
 import           System.Directory (getCurrentDirectory)
 import           System.FilePath ((</>), makeRelative, splitPath, normalise)
-import           System.IO (hPutStrLn, stderr)
+import           System.IO (stderr)
 
 -- | Determines when to rebuild a module
 data RebuildPolicy
@@ -71,8 +72,8 @@ data ProgressMessage
   deriving (Show, Eq, Ord)
 
 -- | Render a progress message
-renderProgressMessage :: ProgressMessage -> String
-renderProgressMessage (CompilingModule mn) = "Compiling " ++ T.unpack (runModuleName mn)
+renderProgressMessage :: ProgressMessage -> T.Text
+renderProgressMessage (CompilingModule mn) = T.append "Compiling " (runModuleName mn)
 
 -- | Actions that require implementations when running in "make" mode.
 --
@@ -270,7 +271,7 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
   requiresForeign = not . null . CF.moduleForeign
 
   progress :: ProgressMessage -> Make ()
-  progress = liftIO . hPutStrLn stderr . renderProgressMessage
+  progress = liftIO . TIO.hPutStrLn stderr . renderProgressMessage
 
   readCacheDb :: Make CacheDb
   readCacheDb = readCacheDb' outputDir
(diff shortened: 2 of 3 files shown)


hdgarrood:
> Due to usage of ANSI.hSupportsANSI for stderr, does the following need to be duplicated for stdout?

I think just replacing `stderr` with `stdout` there is probably what we want.

hdgarrood:
Oh, I see what the confusion is - I think compiler warnings, i.e. warnings which arose from your code, should go to the same place as compile errors, i.e. stdout.

JordanMartinez:
What about the second one? If that isn't changed, doesn't it print compiler errors to `stderr` as well when your table seems to say that it should always be printed to `stdout`.

hdgarrood:
The second one should be changed just by replacing `stderr` with `stdout`; everything that is currently printed to `stderr` in that snippet should be sent to `stdout`.

app/Command/Compile.hs
```diff
@@ -41,25 +41,25 @@ data PSCMakeOptions = PSCMakeOptions
...
-  cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stderr
+  cc <- bool Nothing (Just P.defaultCodeColor) <$> ANSI.hSupportsANSI stdout
   let ppeOpts = P.defaultPPEOptions { P.ppeCodeColor = cc, P.ppeFull = verbose, P.ppeRelativeDirectory = pwd }
   when (P.nonEmpty warnings) $
     hPutStrLn stderr (P.prettyPrintMultipleWarnings ppeOpts warnings)
   case errors of
     Left errs -> do
-      hPutStrLn stderr (P.prettyPrintMultipleErrors ppeOpts errs)
+      hPutStrLn stdout (P.prettyPrintMultipleErrors ppeOpts errs)
       exitFailure
```
JordanMartinez, line 51: Was your earlier comment about swapping it to `stdout` also apply on line 47:
```haskell
-- current
hPutStrLn stderr (P.prettyPrintMultipleWarnings ppeOpts warnings)

-- desired?
hPutStrLn stdout (P.prettyPrintMultipleWarnings ppeOpts warnings)
```

hdgarrood: Yes, definitely. Whenever we are printing compile errors and compile warnings they should always go to the same place.

JordanMartinez: Ok. I think I misread the issue's original comment regarding "glob miss warnings" to include "compiler warnings" since the "compiler errors" row didn't say "compiler errors and compiler warnings"
This has been fixed in latest commit.

JordanMartinez:
Using this PR, I wanted to start fixing the compiler warnings in the ps-0.14 compatible package set and got this initial output. Is that expected?
```
jordan@Natsu:~/Programming/Git Projects/purescript-hail-mary$ npm run build

> purescript-hail-mary@1.0.0 build /home/jordan/Programming/Git Projects/purescript-hail-mary
> ./purs compile ".spago/*/*/src/**/*.purs"

CoCCmoCCCopCCCmCoCooCCmCCiCCoooCCpoCmCoCCCmmoopoolCoommmooimCCCCoCCCpomoCCCooCpCCpmmCCiCmmiommpppmCmlpoooomoooimpmooommoiooippooloppnmppiiipopiimmmmpmmmlpipmmmppmlmmliimmimiigpiillliminlppppipppiilipppiipippillppnpll illiiilplgiiiiiliiinliliiilliniiniiiigiiiWliinnniii nllllilllginillliilgllgnnll lnneinngggnlnUgiiiiniii ngniiinni ii ggiiDiggbngg   gign nnnngnnnEg gnnnggnGnnD  nnan  .g  MTP n dMgggg gggf D ggg  glggaDTggtgTRE NUayaHgUei    T   fNaW   DD o  tay  a yevDontpry nflDDDDyDDDeoteDDDaaDbDDatpDD.CpceadshetpDsikaOaOpOOOcdabOOOttOaOO.aeOOSoeontea
.ieaanitMtMeMMMte..MMMaaMlMMS..MMtn.rta.fDartfesa.a.......SD.........yNE..rtTd..Eeal.aed..H.HDHHHPNtOHHHITHUHHmaqHHirr.IBl.t
C..
IATATaTTTrerMTTTnrTnTTbtuTTnooUnoeRaoFCmrMrMtMMMoti.MMMtaMsMMouaMMgluntoce.nioprLgLaLLLm
nILLL.vLaLLlrlLL..tseltfRnnela.o.....ign...Be.f..
ai..HS
arereo
ar
```

hdgarrood:
`putStrLn` acquires a lock on the handle while it’s printing the line so that this doesn’t happen. Does `renderProgressMessage` return `String`? I think this might be a good opportunity to change it to `Text`. That might fix the issue, too.

hdgarrood:
> putStrLn acquires a lock on the handle while it’s printing the line so that this doesn’t happen.

I've done a little more investigation (see e.g. the implementation of `hPutStr`: http://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.IO.Handle.Text.html#hPutStr%27) and I'm not sure whether this is actually true; I think it depends on the buffering mode of the output handle. I think `stdout` was set to line buffering but `stderr` is set to no buffering. In any case we should be avoiding `String` so I think changing `renderProgressMessage` to return `Text` is desirable.

hdgarrood:
Did that fix the odd output issue/

JordanMartinez:
No it doesn't.

hdgarrood:
What happens if you add calls to `IO.hSetBuffering IO.LineBuffering` for both `stdout` and `stderr` at the start of `main` in `app/Main.hs`?

JordanMartinez:
I'll try that but I won't be able to get to it until tomorrow probably.

JordanMartinez:
Yeah, looks like that fixes it.

JordanMartinez:
It fixes it somewhat, but still not what it was previously. I'm guessing there's something else going on? Sample outpu
```
> ./purs compile ".spago/*/*/src/**/*.purs"

Compiling Milkis.Impl
Compiling Type.TroutCompiling Math
Compiling Global.UnsafeCompiling Web.DOM.Internal.TypesCompiling Type.Data.Row
Compiling Unsafe.Reference


Compiling Data.Traversable.AccumCompiling Data.String.Unsafe
Compiling PartialCompiling Node.NetCompiling Type.EqualityCompiling Type.Data.RowListCompiling Data.String.HtmlElementsCompiling Web.Event.Internal.Types







Compiling Data.SymbolCompiling Effect.Promise.UnsafeCompiling Record.UnsafeCompiling Hyper.ConnCompiling Unsafe.CoerceCompiling Data.Int.BitsCompiling Data.NaturalTransformationCompiling Node.Electron.MenuCompiling Undefined




Compiling Data.Boolean



Compiling DOM.HTML.Indexed.WrapValueCompiling Data.ArrayBuffer.TypesCompiling DOM.HTML.Indexed.PreloadValueCompiling DOM.HTML.Indexed.OnOffCompiling Data.Argonaut.Types.Generic.RepCompiling DOM.HTML.Indexed.ScopeValueCompiling DOM.HTML.Indexed.OrderedListTypeCompiling DOM.HTML.Indexed.MenuitemTypeCompiling DOM.HTML.Indexed.MenuTypeCompiling DOM.HTML.Indexed.KindValueCompiling DOM.HTML.Indexed.DirValueCompiling DOM.HTML.Indexed.InputTypeCompiling DOM.HTML.Indexed.FormMethodCompiling DOM.HTML.Indexed.ButtonTypeCompiling DOM.HTML.Indexed.CrossOriginValueCompiling Control.Semigroupoid
```

src/Language/PureScript/Make/Actions.hs
```diff
@@ -269,7 +271,7 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
   requiresForeign = not . null . CF.moduleForeign
 
   progress :: ProgressMessage -> Make ()
-  progress = liftIO . putStrLn . renderProgressMessage
+  progress = liftIO . TIO.hPutStrLn stderr . renderProgressMessage
```
JordanMartinez, line 274: This change is what's causing the problem. I'm not entirely sure why yet.

hdgarrood: I think it's because the `hPutStrLn` in `Data.Text.IO` is implemented like this:

```haskell
hPutStrLn h t = hPutStr h t >> hPutChar h '\n'
```

See https://hackage.haskell.org/package/text-1.2.4.0/docs/src/Data.Text.IO.html#hPutStrLn

and the runtime can switch to a different thread after the `hPutStr` but before the `hPutChar`. Perhaps `liftIO . TIO.hPutStr stderr . (<> "\n) . renderProgressMessage` might work?

JordanMartinez: Yeah, that fixes it.

JordanMartinez:
> For example I think it would be worth checking that piping json errors into purescript-suggest still works.

In this version, `purescript-suggest` will just hang forever. So, looks like there's some changes we need to do there.

kritzcreek:
This will also need changes in the editor plugins for their non-fast-rebuild paths. For example here's the VSCode plugin appending the build results from stderr: https://github.com/nwolverson/purescript-language-server/blob/b40119da440186b38d7824ab255f12898d208789/src/IdePurescript/Build.purs#L74

hdgarrood:
Indeed, but that shouldn't preclude merging this, I think? @nwolverson and @f-f are already aware that we plan to change this in 0.14.0, do you know of any other editor plugin maintainers we should contact?

kritzcreek:
I don't know how many people are still using the non-LSP based VIM plugin, as that also has a non-fast-rebuild build command. The Emacs plugin only does fast-rebuilds, so we're fine there. Not a blocker in any way, just wanted to make sure we've alerted the right people.

JordanMartinez:
Is there anything else stopping this PR from getting merged?

hdgarrood:
I'd just like to wait for another core team member to get a chance to review and approve before merging, as far as I'm concerned there's nothing else that needs to happen here.

src/Language/PureScript/Make/Actions.hs
```diff
@@ -269,7 +271,7 @@ buildMakeActions outputDir filePathMap foreigns usePrefix =
   requiresForeign = not . null . CF.moduleForeign
 
   progress :: ProgressMessage -> Make ()
-  progress = liftIO . putStrLn . renderProgressMessage
+  progress = liftIO . TIO.hPutStr stderr . (<> "\n") . renderProgressMessage
```
natefaubion, line 274: Is there a reason for not using `TIO.putStrLn`?

hdgarrood: Perhaps a comment here to explain why we aren't using `hPutStrLn` would be good?

natefaubion: Ah oops! Sorry for the redundant conversation 😄

hdgarrood:
Thanks all!

PR #3845, opened 2020-04-12
Clarify whitespace lex error in qualified identifier lexing

### LINKED ISSUE #3801: Misleading lex error when forgetting to access qualified identifier
This is a small nit, but the lex error when forgetting to access a qualified identifier seems a bit misleading. I think we could probably make this particular error more specific for the actual cause, which (imo) is likely that the user forgot to finish what they were typing. (For example, today I was looking around my file for strange whitespace characters without realizing I had just forgotten to finish what I was typing!)

For example, in the following code:

```
bar :: Int
bar = Foo. 
```

You get the lex error
`Illegal whitespace character U+0020`

@natefaubion suggested we could have an error that says `Whitespace not allowed in qualified identifier` or similar.

You get a similar error for when you forget to access an unqualified identifier, such as

```
bar :: Int
bar = foo.
```

resulting in
`Unexpected or mismatched indentation`

(I'm not sure if the second one is as easy to clean up, but I think the first example is relatively straightforward)

Has anyone else had a similar experience, or have a suggestion for a more specific error message?

colinwahl:
This intends to close #3801 .  See issue for motivation/some discussion.

This PR adds a new lex error for when an unexpected character is encountered when tokenizing a qualified identifier.

My hope is that this indicates that the error is likely caused by a missing identifier, whereas the old error said "Illegal whitespace".

I'm not sure how useful it is to print out the unexpected whitespace character - I'm willing to remove that if others agree.

I'm not sure if parser tests exist/where they are, but if there's a useful way of writing a test for this, I'd be happy to add it if someone points me in the right direction.


hdgarrood:
Right now most of the parser tests live in the main compiler repo, even though it arguably would make more sense to put them into the `purescript-cst` library. You can see examples by searching for `@shouldFailWith ErrorParsingModule` in `tests/purs/failing`. I think adding a new test should be as simple as writing an example whose filename matches `tests/purs/failing/*.purs`, and then checking in the corresponding `.out` file which should be generated after running the tests again (after checking that it looks sensible).

lib/purescript-cst/src/Language/PureScript/CST/Errors.hs
```diff
@@ -53,6 +53,7 @@ data ParserErrorType
   | ErrQualifiedName
   | ErrEmptyDo
   | ErrLexeme (Maybe String) [String]
+  | ErrQualifierLexeme Char
```
rhendric, line 56: By analogy with `ErrImportInDecl`, `ErrTypeInConstraint`, etc., should this be `ErrLexemeInQualified`?

lib/purescript-cst/src/Language/PureScript/CST/Errors.hs
```diff
@@ -128,6 +129,10 @@ prettyPrintErrorMessage (ParserErrorInfo {..}) = case errType of
     "Illegal whitespace character " <> displayCodePoint hd
   ErrLexeme (Just a) _ ->
     "Unexpected " <> a
+  ErrQualifierLexeme hd | isSpace hd ->
+    "Unexpected whitespace character " <> displayCodePoint hd <> ", expected qualifier"
```
rhendric, line 133: Don't you mean ‘expected identifier or operator’ instead of ‘expected qualifier’? The qualifier has already been seen at this point in the parse, right?

Also, it looks like the other error messages of this nature follow an ‘Expected X, saw Y’ pattern; I'd expect these messages to match that.

rhendric (changes requested):
I like the overall approach; I think this will be an easy PR to land once you add a few tests per @hdgarrood's comment.

PR #3846, opened 2020-04-13
Pretty print errors as warnings in warnings golden tests

kl0tl:


kl0tl:
Warnings are also printed as errors in `assertCompilesWithWarnings`:

https://github.com/purescript/purescript/blob/8869eb30cc4528a908e37cb97401938e59cbeffb/tests/TestCompiler.hs#L189-L201

and `checkShouldFailWith` always talk about errors

https://github.com/purescript/purescript/blob/8869eb30cc4528a908e37cb97401938e59cbeffb/tests/TestCompiler.hs#L125-L131

Is that something that should be changed?

kritzcreek:
> and checkShouldFailWith always talk about errors

It's certainly misleading... we've already got the confusing thing about warnings and errors both being represented by `MultipleErrors`. In the LSP world there's the all-encompassing _Diagnostic_ which can talk about infos/warnings/errors. Maybe we could use that here? `checkShouldReport` as the function name and `"Expected these diagnostics: ...`?

kl0tl:
I applied your suggestions and removed `printAllWarnings` from `assertCompilesWithWarnings` because warnings were already printed by `checkShouldFailWith`.

kl0tl:
I also reused `prettyPrintMultipleErrors` and `prettyPrintMultipleWarnings` instead of the ad-hoc pretty printer in `checkShouldFailWith`. This should be good to go now.

Issue #3832, opened 2020-04-05
Improve GitHub issues


joneshf:
# Situation

There's been quite a few issues closed with a redirect link elsewhere. For instance, these are all responded with pointing to Discourse:

* https://github.com/purescript/purescript/issues/3542
* https://github.com/purescript/purescript/issues/3589
* https://github.com/purescript/purescript/issues/3671
* https://github.com/purescript/purescript/issues/3674
* https://github.com/purescript/purescript/issues/3723
* https://github.com/purescript/purescript/issues/3755
* https://github.com/purescript/purescript/issues/3757
* https://github.com/purescript/purescript/issues/3784
* https://github.com/purescript/purescript/issues/3786

We don't want questions to end up in the GitHub issue tracker. We want those to go elsewhere. Yet, we're having to triage these after the fact.

# Suggestion

Instead of letting people open an issue that we pretty much know we're going to close with a link elsewhere, we can provide that link up-front.

GitHub has support for [issue template][]s. The templates by themselves help to categorize issue we want people to create. If we know we only want bug reports, compiler improvements, etc, then we can provide templates for those. When we know we want people to raise their question or concern elsewhere, we can provide [`contact_links`][] to those locations.

The documentation around [issue template][]s isn't great but for an example of how this looks in practice, see [TypeScript's new issue chooser][]. For posterity, a screenshot of what it looks like at the moment:
![screenshot-1586109312](https://user-images.githubusercontent.com/1356417/78506073-23bbd480-772c-11ea-81c3-3efc2b1d2a7c.png)

# Proposal
Concrete proposal to implement this:

1. [ ] Follow the steps for [creating issue templates][] to make at least one [issue template][].
1. [ ] Follow the steps for [configuring the template chooser][] to make the [`contact_links`][].

Anyone strongly against doing this?

[`contact_links`]: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
[configuring the template chooser]: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
[creating issue templates]: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#creating-issue-templates
[issue template]: https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates#issue-templates
[typescript's new issue chooser]: https://github.com/microsoft/TypeScript/issues/new/choose

[...]

Issue #3757, opened 2019-11-27
Feature request: String interpolation


i-am-the-slime:
Although I know Strings are evil etc. it's often possible to construct some. The problem with:
```
  "a really really" <>
  "long string " <>
  age
```
Is that it's easy to forget spaces. Would there be interest in string interpolation syntax?

hdgarrood:
Can you open a thread on Discourse instead please?

michaelficarra:
Could it be added in a backward-compatible way? Maybe a new escape sequence?

i-am-the-slime:
@hdgarrood I did. Let's see if a moderator approves it.

hdgarrood:
Thanks! I'll see if we can remove that reviewal step, it was flagged because Discourse thought you typed it "suspiciously fast"... forgetting that it's pretty common for people paste text in which they have already written down elsewhere.

hdgarrood:
https://discourse.purescript.org/t/feature-request-string-interpolation/1061

Issue #3755, opened 2019-11-16
Unison-like AST approach?


pkamenarsky:
With the recent release of [Unison](https://www.unisonweb.org) I was wondering - would Purescript benefit from the same approach of storing and hashing immutable AST definitions in an append-only data store, which are then referred to by easily refactorable aliases?

This would for example replace https://github.com/purescript/purescript/issues/2869 and give us all the other benefits the Unison devs are talking about, like to ability to do transparent distributed programming, as in the [Unison introduction](https://github.com/unisonweb/unison):

```
x = factorial 6
Remote.transfer alice
y = foo x -- happens on `alice` node
Remote.transfer bob
bar x y -- happens on `bob` node
```

This would require at least a [codebase manager](https://www.unisonweb.org/docs/tour/hello-ucm) and would have far reaching implications on the ecosystem, build tools and the language itself, so it's not anything that will happen any time soon. On the other hand, Purescript still doesn't have the baggage of a more established language, so it might be beneficial to have a discussion and ponder the possibilities.

natefaubion:
I think that Unison is very interesting. However, I think a discussion like this is better suited for the [Discourse](https://discourse.purescript.org/).

Issue #3723, opened 2019-08-23
Emit metadata for data constructors in dev environment


thebrodmann:
For debugging purposes in https://github.com/slamdata/purescript-halogen/issues/614 editing of PureScript datatypes is a challenge due to lack of access to related data constructors.
So as a solution maybe we can add some metadata to data constructors in the dev environment.

As an example:
```purs
data Quantum = Yes | No | Both
```
The `Quantum` datatype may can be transpile to:
```js
var Yes = (function () {
  function Yes() {

  };
  Yes.value = new Yes();
  Yes.metadata = {}
  return Yes;
})();
var No = (function () {
  function No() {

  };
  No.value = new No();
  No.metadata = {}
  return No;
})();
var Both = (function () {
  function Both() {

  };
  Both.value = new Both();
  Both.metadata = {}
  return Both;
})();

var Quantum = { Yes: Yes, No: No, Both: Both }

Yes.metadata.constructors = Quantum
No.metadata.constructors = Quantum 
Both.metadata.constructors = Quantum

// Tests

Yes.value.constructor === Yes   // true
No.value.constructor === No     // true
Both.value.constructor === Both // true
```

hdgarrood:
This feature request is not yet at a stage where it can earn a spot on the issue tracker (sorry). Please open a Discourse thread instead: https://discourse.purescript.org/

Note that this use case ought to be covered by the `generics-rep` library: https://pursuit.purescript.org/packages/purescript-generics-rep/6.1.1. If it isn't, please can you explain why `generics-rep` is not sufficient (in a Discourse thread)?

PR #3855, opened 2020-04-25
Direct proposals to Discourse

hdgarrood:
Following on from #3832. We've had an implicit and only partially enforced policy for a while that would prefer to keep the issue tracker for things that a) involve dev work on the compiler and b) have already had at least a tentative approval from a maintainer, and so in a few occasions in the past we've closed issues and asked people to open threads on Discourse first. For example:

* #3723 
* #3755 
* #3757

This PR updates the issue templates to encourage people to open threads on Discourse for discussing proposals for changing either the language or the compiler first, so that this policy is a bit clearer and more explicit. Discourse allows setting the topic and post body via query strings so I've done that in order to retain the existing proposal template. I've made the PR from the master branch of my fork so you can see how it looks there: https://github.com/hdgarrood/purescript/issues/new/choose


natefaubion:
I don't think this is necessarily a bad thing, but my experience is that Discourse garners _far_ less participation that GH issues.

hdgarrood:
I'd say a proposal should become a compiler ticket once it has tentative approval from one core owner, and also provided that it actually represents work to be done on the compiler. I'm not sure where the best place to note this down would be, though. CONTRIBUTING.md?

hdgarrood:
By the way if we want to stop using this policy (of not allowing feature requests to come directly to this issue tracker) then that's fine with me too. I think the issue tracker is relatively healthy now, but there was definitely a point where there were a lot of proposals which were poorly justified, severely lacking detail, or just not something that we would ever consider, and the issue tracker was starting to become a sink for maintainers' energy. But perhaps issue templates are enough to mitigate that? If we have provided a template and someone has chosen to discard it and still only provide a sentence or two, it makes it a lot easier to just close an issue outright.

garyb:
I'd be happy with trying the template for now. The fact it has a section for "Detailed description of the proposal", examples, etc. means people will have to put a little more thought into it assuming they don't just discard the template. And yeah, if they discard the template we can prompt them to fill it in properly, or close the issue if they're not prepared to do so.

hdgarrood:
Sure, that works for me. I'll close this then. :+1:

PR #3856, opened 2020-04-25
[purs ide] Extracts documentation comments for type classes

kritzcreek:
I had assumed we were already doing this? Maybe something about the Name structure changed, or this was just an oversight in the original PR. I also added a test to make sure it doesn't regress.


kritzcreek:
Thanks!

Issue #3857, opened 2020-04-25
Compiler hangs under some situations using Coercible


Thimoteus:
## Description

In some situations using [`coerce`](https://github.com/purescript/purescript-safe-coerce/blob/master/src/Safe/Coerce.purs#L9-L10) will cause the compiler to hang, whereas pattern matching will not.

## To Reproduce

`spago.dhall`:
```
{ name = "my-project"
, dependencies = [ "prelude", "safe-coerce" ]
, packages = ./packages.dhall
, sources = [ "src/**/*.purs", "test/**/*.purs" ]
}
```
`packages.dhall` after the first comment:

```
let upstream =
      https://github.com/purescript/package-sets/releases/download/psc-0.13.6-20200423/packages.dhall sha256:c180a06bb5444fd950f8cbdd6605c644fd246deb397e62572b8f4a6b9dbcaf22

let overrides =
    { prelude =
        upstream.prelude // { version = "c932361d008379958f14ca8cc2fe32e06cc2647d" }
    }

let additions =
    { safe-coerce =
        { dependencies = [ "unsafe-coerce" ]
        , repo =
            "https://github.com/purescript/purescript-safe-coerce.git"
        , version = "v0.0.2"
        }
    }

in  upstream // overrides // additions
```

Failing code, placed inside `src/Main.purs`:
```
module Main where

import Prelude

import Safe.Coerce (coerce)

data Rep a
  = Leaf a
  | Link Int (Rep a) (Rep a)

data Rank

foreign import data Z :: Rank
foreign import data S :: Rank -> Rank

newtype Ranked :: Rank -> Type -> Type
newtype Ranked n a = Ranked a

type Tree n a = Ranked n (Rep a)

linkRep :: forall a. Rep a -> Rep a -> Rep a
linkRep t1 t2 = Link (rankRep t1 + 1) t1 t2

rankRep :: forall a. Rep a -> Int
rankRep =
  case _ of
    Leaf _ -> 0
    Link n _ _ -> n

badLink :: forall a n. Tree n a -> Tree n a -> Tree (S n) a
badLink t1 t2 = coerce (linkRep (coerce t1) (coerce t2))

goodLink :: forall a n. Tree n a -> Tree n a -> Tree (S n) a
goodLink (Ranked t1) (Ranked t2) = coerce (linkRep t1 t2)
```
Then run `spago build`.
## Expected behavior

`badLink` and `goodLink` should both not cause the compiler to hang, but only `goodLink` ends.

[...]

PR #3860, opened 2020-05-01
Check roles of mutually recursive types

kl0tl:
Here’s a tentative to fix https://github.com/purescript/purescript/issues/3857.

This turns role inference into a stateful process repeatedly updating roles in the environment until we infer the roles of all the data types belonging to a mutually recursive group. The idea is taken from GHC own implementation (see [`GHC.Tc.TyCl.Utils.inferRoles`](https://gitlab.haskell.org/ghc/ghc/-/blob/fd7ea0fee92a60f9658254cc4fe3abdb4ff299b1/compiler/GHC/Tc/TyCl/Utils.hs)).

Unlike GHC we discover mutually recursive data types as we go so an inference step returns `S.Set (Qualified (ProperName 'TypeName)` instead of `Bool` and we infer our dependencies roles before taking another step, stopping when we don’t find any dependencies.

### DIFF
--- a/src/Language/PureScript/TypeChecker/Roles.hs
+++ b/src/Language/PureScript/TypeChecker/Roles.hs
@@ -45,15 +45,18 @@ instance Monoid RoleMap where
 
 type RoleEnv = M.Map (Qualified (ProperName 'TypeName)) [Role]
 
+typeKindRoles :: TypeKind -> Maybe [Role]
+typeKindRoles = \case
+  DataType args _ ->
+    Just $ map (\(_, _, role) -> role) args
+  ExternData roles ->
+    Just roles
+  _ ->
+    Nothing
+
 getRoleEnv :: Environment -> RoleEnv
 getRoleEnv env =
-  flip M.mapMaybe (types env) $ \case
-    (_, DataType args _) ->
-      Just $ map (\(_, _, role) -> role) args
-    (_, ExternData roles) ->
-      Just roles
-    _ ->
-      Nothing
+  M.mapMaybe (typeKindRoles . snd) (types env)
 
 updateRoleEnv
   :: Qualified (ProperName 'TypeName)
@@ -76,7 +79,7 @@ lookupRoles
   -> Qualified (ProperName 'TypeName)
   -> [Role]
 lookupRoles env tyName =
-  fromMaybe [] $ M.lookup tyName (getRoleEnv env)
+  fromMaybe [] $ M.lookup tyName (types env) >>= typeKindRoles . snd
 
 -- | This function does the following:
 --
--- a/src/Language/PureScript/TypeChecker/Roles.hs
+++ b/src/Language/PureScript/TypeChecker/Roles.hs
@@ -5,19 +5,20 @@
 -- Role inference
 --
 module Language.PureScript.TypeChecker.Roles
-  ( lookupEnvRoles
+  ( lookupRoles
   , checkRoles
+  , checkDataBindingGroupRoles
   ) where
 
 import Prelude.Compat
 
-import Control.Monad
 import Control.Monad.Error.Class (MonadError(..))
-import Control.Monad.State.Class (MonadState(..))
+import Control.Monad.State (MonadState(..), runState, state)
 import Data.Coerce (coerce)
 import qualified Data.Map as M
 import Data.Maybe (fromMaybe)
 import qualified Data.Set as S
+import Data.Semigroup (Any(..))
 import Data.Text (Text)
 
 import Language.PureScript.Environment
@@ -25,7 +26,6 @@ import Language.PureScript.Errors
 import Language.PureScript.Names
 import Language.PureScript.Roles
 import Language.PureScript.Types
-import Language.PureScript.TypeChecker.Monad
 
 -- |
 -- A map of a type's formal parameter names to their roles. This type's
@@ -43,23 +43,40 @@ instance Monoid RoleMap where
   mempty =
     RoleMap M.empty
 
+type RoleEnv = M.Map (Qualified (ProperName 'TypeName)) [Role]
+
+getRoleEnv :: Environment -> RoleEnv
+getRoleEnv env =
+  flip M.mapMaybe (types env) $ \case
+    (_, DataType args _) ->
+      Just $ map (\(_, _, role) -> role) args
+    (_, ExternData roles) ->
+      Just roles
+    _ ->
+      Nothing
+
+updateRoleEnv
+  :: Qualified (ProperName 'TypeName)
+  -> [Role]
+  -> RoleEnv
+  -> (Any, RoleEnv)
+updateRoleEnv qualTyName roles' roleEnv =
+  let roles = fromMaybe (repeat Phantom) $ M.lookup qualTyName roleEnv
+      mostRestrictiveRoles = zipWith min roles roles'
+      didRolesChange = any (uncurry (<)) $ zip mostRestrictiveRoles roles
+  in (Any didRolesChange, M.insert qualTyName mostRestrictiveRoles roleEnv)
+
 -- |
 -- Lookup the roles for a type in the environment. If the type does not have
 -- roles (e.g. is a type synonym or a type variable), then this function
 -- returns an empty list.
 --
-lookupEnvRoles
+lookupRoles
   :: Environment
   -> Qualified (ProperName 'TypeName)
   -> [Role]
-lookupEnvRoles env tyName =
-  case fmap snd $ M.lookup tyName (types env) of
-    Just (DataType args _) ->
-      map (\(_, _, role) -> role) args
-    Just (ExternData roles) ->
-      roles
-    _ ->
-      []
+lookupRoles env tyName =
+  fromMaybe [] $ M.lookup tyName (getRoleEnv env)
 
 -- | This function does the following:
 --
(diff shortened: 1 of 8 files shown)


src/Language/PureScript/TypeChecker/Roles.hs
```diff
@@ -109,31 +130,37 @@ inferRoles env tyName
             -- * If the role is phantom, terminate, since the argument's use of
             --   our parameters is unimportant.
             TypeConstructor _ t1Name ->
-              let t1Roles = inferRoles env t1Name
+              let roles = M.lookup t1Name roleEnv'
+                  inferred = fromMaybe (repeat Phantom) roles
```
natefaubion, line 134: In a couple of places you use `Phantom` as a default for a lookup. Can you clarify why this is desirable?

kl0tl: The phantom role is used as a default because it is the least restrictive, so later inference passes can only refine it to a more precise role. We always prefer the more restrictive role inferred for the same parameter in the `Semigroup` instance for `RoleMap` and `updateRoles`.

For instance, given these definitions:

```purs
data D1 a = D1 a
data D2 a = D2 (D1 a)
```

When inferring `D2` roles we discover they depend on those of `D1`, but we don’t know them yet so we pick phantom for `a` and store that in our role environment. Then we infer the type parameter of `D1` is representational, store that in our role environment and go back to `D2`. This time we know `D1` roles so we can infer `a` is representational and update the role environment by taking the more restrictive of phantom and representational for `a`, which is representational.

hdgarrood:
Thanks for picking this up! While reviewing, I realised that we have another bug in Coercible (#3867), and I think it might be preferable to address that one before this one, as that one might necessitate a more fundamental change in the mechanics of how we handle roles in the compiler.

kl0tl:
I managed to rebase this on top of https://github.com/purescript/purescript/pull/3873!

I was able to follow GHC approach more closely now that we don’t have to discover dependencies between mutually recursive as we go anymore. I feared my previous approach wouldn’t propagate roles in some intricate cases but didn’t found any example of that. I find reassuring to do somewhat like GHC though.

hdgarrood:
@kl0tl I'd like to try to get this PR merged next. Could you merge or rebase master so that it applies cleanly when you have a moment, please?

kl0tl:
I rebased the PR.

src/Language/PureScript/TypeChecker/Roles.hs
```diff
@@ -71,47 +88,88 @@ lookupEnvRoles env tyName =
...
+          let
+            k (var, _) inf dec =
+              if inf < dec
+                then throwError . errorMessage $ RoleMismatch var inf dec
+                else pure dec
+          sequence $ zipWith3 k tyArgs (fromMaybe (repeat Phantom) inferredRoles) declaredRoles
+        Nothing ->
+          pure $ fromMaybe (Phantom <$ tyArgs) inferredRoles
+
+inferDataBindingGroupRoles
```
hdgarrood, line 137: Rather than keeping this function "pure", I think it might be preferable (or at least better aligned with the way the rest of the type checker is written) for it to operate in `MonadState CheckState m`, and read and write directly from the actual type-checking environment, rather than doing its own thing at first and then later updating the environment after inferring/checking roles. I don't feel particularly strongly about this, though. @natefaubion do you have any thoughts on this?

src/Language/PureScript/TypeChecker/Roles.hs
```diff
@@ -71,47 +88,88 @@ lookupEnvRoles env tyName =
...
+  -> RoleEnv
+  -> RoleEnv
+inferDataBindingGroupRoles moduleName group roleEnv =
+  let (Any update, roleEnv') = flip runState roleEnv $
+        mconcat <$> traverse (state . inferDataDeclarationRoles moduleName) group
+  in if update
+     then inferDataBindingGroupRoles moduleName group roleEnv'
+     else roleEnv'
+
+inferDataDeclarationRoles
```
hdgarrood, line 149: Could you add some documentation about the meaning of the `Any` in the return type please?

src/Language/PureScript/TypeChecker/Roles.hs
natefaubion: I don't really have much of an opinion. Is there some currying advantage to how the functions are arranged, and how it returns an additional lambda, that's exploited elsewhere?

kl0tl: The explicit floating of `initialRoleEnv` and `inferredRoleEnv` outside of the do block in `checkDataBindingGroupRoles` is supposed to allow the roles of the data declarations in the binding group to be inferred only once and then looked up repeatedly in the `DataBindingGroupDeclaration` case of `Language.PureScript.TypeChecker.typeCheckAll`. This isn’t something I’ve verified because I’m not familiar with Core inspection but I’m not sure GHC would be able to share the work otherwise.

I can try to store the inferred roles environment in the typechecker monad state and separate roles inference from roles checking in binding groups. This should be moree performant than computing a new roles map from scratch each time we need to check roles.

hdgarrood: Since neither @natefaubion nor I feel strongly I’m tempted to say let’s leave this as is for now. I’m happy to merge once there’s a comment about what the Any in inferDataDeclarationRoles means.

hdgarrood: Ah, I guess it’s that we want to keep  walking over all of the data declarations in the binding group until we reach a point where walking doesn’t change anything in the roleEnv, and the Any part of the result indicates whether something changed?

src/Language/PureScript/TypeChecker/Roles.hs
kl0tl: I added a comment and aliased the `Any` to `DidRolesChange` to make things explicit.

src/Language/PureScript/TypeChecker/Roles.hs
kl0tl: I lifted the role inference to the typechecker monad!

hdgarrood:
@kl0tl Thanks for making these changes! On reflection, though, I'm less sure about this design. With this branch as it is now, we are storing roles in three separate places in the CheckState: in the CheckState's `checkRoleEnv` field, in the `roleDeclarations` field of the Environment, and in the `types` field of the Environment. To me, this feels like we have more mutable state than we ought to have - I was more hoping that we could use the `types` field of the Environment to store the intermediate roles during inference. However, I'd really like to avoid doing anything which would delay getting this fix merged, so at this point I would prefer to leave this as a possibility for changing later, and for now, we could just go back to 4ac6f09d405003cf3d2f3f43eefff7a6aa0ba1c3, if that sounds okay?

Speaking of minimising the amount of mutable state we have, it occurred to me that now we are dealing with data binding groups sensibly, we can probably make the `roleDeclarations` field obsolete by putting role annotations in the same binding group as the data types they refer to, so that the data we need is available at the time we actually need it, rather than having to store them in the environment separately first and pull them back out later. Again, that's something I'd like to address separately from this PR, though.

I'm not actually very keen on `type DidRolesChange = Any`, since to me, having a named type like that suggests a newtype or a separate data type which is different from  `Any`; it suggests more type safety than there really is. As it happens we came quite close to merging a compiler warning which discouraged you from defining type synonyms like this: #2932. I'd prefer a proper data type or newtype like `data DidRolesChange = RolesChanged | NoRolesChanged` with the appropriate Semigroup instance, or `newtype DidRolesChange = DidRolesChange Any`, or alternatively just changing it back to `Any` (and bringing the comment across with it in any case).

kl0tl:
I’ve rollback the lifting of the role inference state to the typechecker monad and the `DidRolesChange` type alias to unblock this pull request.

I don’t think we can get away with storing roles only in the types field of the environment because we need to infer the roles of a data type before inserting it in the environment and we need to store the role declarations separately in order to check they’re not less restrictive than the inferred roles but we can discuss this in another pull request.

src/Language/PureScript/TypeChecker/Roles.hs
```diff
@@ -43,23 +43,40 @@ instance Monoid RoleMap where
...
-lookupEnvRoles env tyName =
-  case fmap snd $ M.lookup tyName (types env) of
-    Just (DataType args _) ->
-      map (\(_, _, role) -> role) args
-    Just (ExternData roles) ->
-      roles
-    _ ->
-      []
+lookupRoles env tyName =
+  fromMaybe [] $ M.lookup tyName (getRoleEnv env)
```
natefaubion, line 79: This change seems a little excessive. Whenever you want to lookup a role, instead of looking it up directly, you are mapping over the entire type environment to build another environment that is a subset of the original. I understand that there _may_ be some clever laziness involved (it isn't clear to me that you are getting that with `Map` vs something like `List`), but this seems like an easy way to also accidentally introduce poor asymptotics. What was the motivation for this change?

kl0tl: I wanted to share logic and thought the whole map wouldn’t actually be built thanks to laziness but I should have verified this assumption. I’ve reused a `typeKindRoles` function in both `getRoleEnv` and `lookupRoles`.

hdgarrood:
:tada:

Issue #2837, opened 2017-04-14
Run test-everything as part of standard tests


matthewleon:
Perhaps something like `test-everything.sh` should run as part of the standard test suite. You would then signal that you are making a breaking change if you wanted to disable it. Not sure what the right way to have that interact with Travis would be.

hdgarrood:
I am not sure the 'new contributor' label is appropriate here as I think this will take a bit of care and ecosystem knowledge to design properly. If it was just a binary on-off switch for test-everything.sh, I think what would end up happening would be that we would the first breaking change in the lead up to a 0.x.0 release would necessitate switching test-everything off and it would basically just stay off until 0.x.0 had been released and the core libraries had been updated, meaning that most of the time, it wouldn't be providing any benefit.

Is the motivation here to avoid making breaking changes accidentally? I think one way of addresing it would be to include a list of all the errors which we expect to be raised while attempting to compile the full set of core libraries. Probably a list of `(ModuleName, [ErrorCode])` would do. Then we compare that with the actual set of errors produced when attempting to compile the core libraries, and mark CI as a failure if they don't match. If possible I think we should include this check in the standard `stack test` script as well, so that you don't have to wait for CI to fail.

paf31:
Yes good point, I've removed the label.

> Is the motivation here to avoid making breaking changes accidentally?

Yes, there are some bugs which have only shown up in the past when compiling actual code, not test code.

hdgarrood:
@kl0tl suggests removing test-everything in #3852, since it has been a very long time since it has been used. I think I agree - this seems like a good idea at first but given that this has seen no movement in almost 3 years, and also I think running the whole test-everything script in CI is going to require more upkeep work than it is worth, it doesn't seem like a good tradeoff. For instance: should `test-everything` use a package set, or should it always use the most recent versions of everything? If we use a package set, it will be easy for the package set to go out of date (i.e. we will likely stay on an old package set), unless we want to build something to remind us to update it, which I'm not sure we do. Alternatively, if we have it use the most recent versions of everything, problems in the core libraries would cause CI to fail, which seems wrong too. In either approach, it could add a lot of time to CI.

I suspect it will be more effective to add a step in the release checklist just before we cut any release to say that a maintainer should manually verify that the compiler can still compile a recent package set.

PR #3861, opened 2020-05-01
Remove core tests

kl0tl:
I’m not sure there‘s more to add to the release guide, I’ll take any suggestion though.

Close https://github.com/purescript/purescript/issues/2837.


RELEASE_GUIDE.md
```diff
@@ -21,6 +21,8 @@ make use of? If so:
 - Update node bindings
 - Update web bindings
 
+This can be verified by compiling [the latest package set](https://github.com/purescript/package-sets/releases/latest).
```
hdgarrood, line 24: I think I would rather put this as an item in the list directly under the "Before making a release" heading - "Check that there are no unintended breaking changes by compiling the latest package set." I think it's something which really ought to be done every time, whereas this frames it as a bit more of an optional suggestion.

Issue #3869, opened 2020-05-11
Incorrect code generated matching on imported constructor where other non-exported constructors exist


fsoikin:
## Description

When only one constructor of a type is exported from its module, it appears that PureScripts tries to optimize pattern-matches as if other constructors did not exist at all, rather than just being hidden. This leads to generating unsound JS code.

## To Reproduce

```
-- Note: only constructor `A` is exported. Constructor `B` is hidden.
module M(T(A), mkB) where
data T = A | B
mkB = B :: T


module Main where
import M(T(..), mkB)

f :: T -> Int
f = case _ of
  A -> 0
  _ -> 1

main = Console.log $ show $ f mkB
```

The above program prints "0", and the following JS code is generated for function `f`:

```
var f = function (v) {
    return 0;
};
```

## Expected behavior

Generated JS code should look the same as if both `A` and `B` were exported from `M`:

```
var f = function (v) {
    if (v instanceof M.A) {
        return 0;
    };
    return 1;
};
```

## Additional context

The example above is silly for the sake of being minimal, but it can easily be made to actually crash at runtime, for example:

```
module M(T(A), mkB) where

data T
  = A { foo :: String }
  | B

mkB = B :: T


module Main where
import M(T(..), mkB)

f :: T -> String
f = case _ of
  A x -> x.foo
  _ -> "ain't gonna happen"

main = Console.log $ f mkB
```

results in:

```
TypeError: Cannot read property 'foo' of undefined
```

## PureScript version

0.13.6

hdgarrood:
Yikes! Thanks for the report.

kl0tl:
This happens because the only constructor is inferred to be for a product type here: https://github.com/purescript/purescript/blob/df3170f5bb0ba21517e2ca72db6aee5c479c99df/src/Language/PureScript/CoreFn/Desugar.hs#L181

If we want to keep optimizing matches on product types perhaps we could store a `ConstructorType` under the `Data` constructor of `DataDeclType` so we just have to lookup in the environment instead of trying to guess from the number of exported constructors.

hdgarrood:
Seems reasonable, yeah. I think we could potentially also consider removing the ability to export some but not all constructors of a data type, since it appears pattern matching is completely broken on data types where not all constructors are exported now. I don't think only exporting some constructors of a data type makes a huge amount of sense in the first place: if it worked properly, you'd be forced to add a catch-all pattern every time you matched on that type, because we treat non-exhaustive pattern matches as an error.

[...]

PR #3872, opened 2020-05-13
Forbid partial data constructors exports

kl0tl:
It seemed appropriate to reuse the `TransitiveDctorExportError` error. I stored the constructors names under a list for consistency with `TransitiveExportError` but perhaps non empty lists would be better 🤷‍♂️ 

Close https://github.com/purescript/purescript/issues/3869.


src/Language/PureScript/TypeChecker.hs
```diff
@@ -712,3 +713,19 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
     extractMemberName (TypeDeclaration td) = tydeclIdent td
     extractMemberName _ = internalError "Unexpected declaration in typeclass member list"
   checkClassMembersAreExported _ = return ()
+
+  checkDataConstructorsAreExported :: DeclarationRef -> m ()
+  checkDataConstructorsAreExported dr@(TypeRef ss' name (Just exportedDataConstructorsNames))
+    | not (null exportedDataConstructorsNames) = do
+      let dataConstructorNames = fromMaybe [] . headMay $ mapMaybe findDataConstructorsNames decls
```
natefaubion, line 720: Data constructors are all in the environment, so we shouldn't need to filter through all declarations for every data export. They are under `types`, and part of the unfortunately named `TypeKind` slot.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -712,3 +713,18 @@ typeCheckModule (Module ss coms mn decls (Just exps)) =
     extractMemberName (TypeDeclaration td) = tydeclIdent td
     extractMemberName _ = internalError "Unexpected declaration in typeclass member list"
   checkClassMembersAreExported _ = return ()
+
+  checkDataConstructorsAreExported :: DeclarationRef -> m ()
```
hdgarrood, line 717: Would you mind adding a comment here? I think something along the lines of "if any data constructors are exported, we require all data constructors to be exported" would be enough.

hdgarrood (approved):
Looks great, thanks!

src/Language/PureScript/TypeChecker.hs
kl0tl: Sure! I added a comment.

PR #3873, opened 2020-05-13
Check role declarations against inferred roles

### LINKED ISSUE #3867: Explicit roles may be more permissive than inferred
## Description

When an explicit role annotation is provided, the user is able to give a less restrictive role than what the compiler would have inferred. This is a problem because it allows you to make invalid coercions, compromising the safety of `Coercible`:

```purescript
module Main where

import Safe.Coerce (coerce)

data Oops a = Oops a

unOops (Oops x) = x

type role Oops phantom -- This should be rejected

invalid :: forall a b. a -> b
invalid x = unOops (coerce (Oops x))
```

## To Reproduce

Provide an explicit role declaration in which at least one parameter is given a more permissive role than its inferred role would have been.

## Expected behavior

The compiler rejects the program, with a message saying that the roles given are too permissive.

hdgarrood:
Fixes #3867. The main change here is that we are now storing checked
roles in the `types` field of the Environment for `data` and
`foreign import data` declarations, via the additions to the `TypeKind`
type. Because we are storing roles in the environment, we no longer need
to infer roles for any given type more than once. As such, `checkRoles`
now works in a monad with `MonadState CheckState`, and when we come
across another type while inferring roles, we can simply look its roles
up from the Environment. The BindingGroups desugaring step ensures
that types are in topological order, so this is safe: any mentioned
types' roles will already have been checked and stored in the
Environment.

One case that we do not yet handle is types which cyclically depend on
each other, e.g.

   data F a = F (G a)
   data G a = G (F a)

Rather than crashing the compiler, these will now be incorrectly marked
Phantom. I would like to address this in a future PR.


hdgarrood:
@kl0tl would you like to review this? I think the approach I've taken here should allow us to take advantage of existing parts of the compiler -- specifically the BindingGroups desugaring pass and the type checking Environment -- to make this a bit more performant and robust. Starting from what I have here, my hope is that handling role inference for recursive types should be relatively straightforward, since we can take advantage of the BindingGroups pass, but it will end up looking a bit different from how you did it in your other PR.

hdgarrood:
I'm also not bothering to deal with type synonyms or constrained types in data declarations yet, and that's also on purpose - I'd prefer to address these things incrementally.

src/Language/PureScript/TypeChecker.hs
```diff
@@ -270,7 +277,9 @@ typeCheckAll moduleName _ = traverse go
       checkDuplicateTypeArguments $ map fst args
       (dataCtors, ctorKind) <- kindOfData moduleName (sa, name, args, dctors)
       let args' = args `withKinds` ctorKind
-      addDataType moduleName dtype name args' dataCtors ctorKind
+      roles <- checkRoles moduleName name args' dctors
+      let args'' = zipWith (\(a, b) c -> (a,b,c)) args' roles
```
kl0tl, line 281: This looks like `withRoles`.

lib/purescript-ast/src/Language/PureScript/Environment.hs
```diff
@@ -478,14 +450,14 @@ primSymbolTypes =
...
-    , (primSubName C.typeError "Fail", (kindDoc -:> kindConstraint, ExternData))
-    , (primSubName C.typeError "Warn", (kindDoc -:> kindConstraint, ExternData))
-    , (primSubName C.typeError "Text", (kindSymbol -:> kindDoc, ExternData))
-    , (primSubName C.typeError "Quote", (kindType -:> kindDoc, ExternData))
-    , (primSubName C.typeError "QuoteLabel", (kindSymbol -:> kindDoc, ExternData))
-    , (primSubName C.typeError "Beside", (kindDoc -:> kindDoc -:> kindDoc, ExternData))
-    , (primSubName C.typeError "Above", (kindDoc -:> kindDoc -:> kindDoc, ExternData))
+    [ (primSubName C.typeError "Doc", (kindType, ExternData []))
+    , (primSubName C.typeError "Fail", (kindDoc -:> kindConstraint, ExternData [Phantom]))
+    , (primSubName C.typeError "Warn", (kindDoc -:> kindConstraint, ExternData [Phantom]))
```
kl0tl, line 455: Those prim type classes have a phantom role but nominal roles are picked otherwise for type classes parameters here: https://github.com/purescript/purescript/blob/38dbf65db6c2720449571baaa654cedaae4cc2d6/src/Language/PureScript/TypeChecker.hs#L157

hdgarrood: Good spot, thanks! I agree these should probably be `Nominal`.

hdgarrood:
> Is there a good reason for giving phantom roles to prim types and kinds? I would’ve picked nominal.

My general approach was that I was trying to give the same roles as would have been inferred. We've got:

- `Row` as `phantom` -- I think this is fine because the type `Row a` has no inhabitants. This won't allow coercing e.g. `Record (a :: Int)` to `Record (b :: String)`, because we only care about the types of fields when inferring roles, not their kinds.
- `Function` as `representational representational` -- I'm pretty confident about this one, because it's what the paper says, and we do want to be able to coerce e.g. `Int -> Int` to `Age -> Age`, but not to `Int -> String`.
- `Array` as `representational` -- this is an easy one, I'm pretty confident about this one
- `Record` as `representational` -- we want to ensure that all of the fields are representationally equal. It occurs to me that the role inference might be wrong for rows, actually, because I don't think it looks at the labels right now, but the labels clearly do affect the representation.
- `RowList` as `phantom` -- like `Row`, I think this is fine because `RowList a` has no inhabitants.
- `Cons` has all of its parameters as `phantom` because its parameters definitely can't affect its representation, because it has none, since it's not even of kind `Type` once it has had all its parameters supplied.
- All the `Prim.TypeError` types as `phantom`, again because they have no inhabitants. I think you're right that the classes should be `nominal` though.

kl0tl:
I don’t think the roles picked cause any issues but `type role Row phantom` and `type role RowList phantom` mean that the constraints `Coercible (Row Type) (Row Symbol)` and `Coercible (RowList Type) (RowList Symbol)` hold (between _kinds_ 🙃) and those axioms feel unnecessary to me unless we gain something by assuming them.

hdgarrood:
Admittedly I'm unable to come up with a situation in which those dictionaries are useful to have, but it seems reasonable to me that you could e.g. coerce an `Either (Row Type) Int` to an `Either (Row Symbol) Int`. If you look at `Row` and `RowList` as versions of `Proxy` but without any constructors (and with a little bit of compiler support), I think it makes sense?

tests/purs/failing/CoercibleRoleMismatch1.purs
```diff
@@ -0,0 +1,8 @@
+-- @shouldFailWith RoleMismatch
+module Main where
+
+import Safe.Coerce (coerce)
```
kl0tl, line 4: This import is redundant.

kl0tl:
I guess you‘re right. This still feels weird to me but that‘s definitely not a strong enough argument against phantom roled `Row` and `RowList` 😄

lib/purescript-ast/src/Language/PureScript/Environment.hs
```diff
@@ -655,3 +627,17 @@ dictSynonymName = ProperName . dictSynonymName' . runProperName
...
+--
+nominalRolesForKind :: Type a -> [Role]
+nominalRolesForKind
+  = go []
+  where
+    go acc = \case
+      TypeApp _ (TypeApp _ fn k1) _k2 | eqType fn tyFunction ->
+        go (Nominal : acc) k1
+      _k ->
+        Nominal : acc
```
kl0tl, line 643: We should recurse on the second argument of `->` and return the accumulator untouched in the base case. Otherwise we return `[Nominal]` for `Type`, `[Nominal, Nominal]` for `Type -> Type` and `Type -> Type -> … -> Type`, `[Nominal, Nominal, Nominal]` for `(Type -> Type) -> Type`…

```purs
module Example where

import Safe.Coerce (coerce)

foreign import data Foreign3 :: Type -> Type -> Type -> Type

-- We infer `nominal nominal` instead of `nominal nominal nominal`,
-- this is equivalent to giving a phantom role to `Foreign3` last parameter:

example :: ∀ a b c d. Foreign3 a b c -> Foreign3 a b d
example = coerce
```

Also now that we have kind polymorphism we should recurse under `ForAll`:

```purs
module Example where

import Safe.Coerce (coerce)

foreign import data Foreign2 :: ∀ k. k -> k -> Type

-- We infer `nominal` instead of `nominal nominal`,
-- this is equivalent to giving a phantom role to `Foreign2` last parameter:

example :: ∀ k (a :: k) (b :: k) (c :: k). Foreign2 a b -> Foreign2 a c
example = coerce
```

I understand this function was just renamed and extracted from `Language.PureScript.TypeChecker.Roles` so I can open an issue and deal with this in another pull request if you so wish.

hdgarrood: Dealing with this in another pull request sounds preferable to me. Thanks!

hdgarrood:
Thanks @natefaubion, @kl0tl!

PR #3877, opened 2020-05-16
Fix CoreFn FromJSON version parsing and add test

paulyoung:
I noticed that a value of `0.13.6` in `corefn.json` was being parsed as `0`. This fixes that, and adds a test.

This bug appears to have been around since https://github.com/purescript/purescript/pull/3049, so about 3 years 😎

### DIFF
--- a/src/Language/PureScript/CoreFn/FromJSON.hs
+++ b/src/Language/PureScript/CoreFn/FromJSON.hs
@@ -4,6 +4,7 @@
 
 module Language.PureScript.CoreFn.FromJSON
   ( moduleFromJSON
+  , parseVersion'
   ) where
 
 import Prelude.Compat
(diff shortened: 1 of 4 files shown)


tests/TestCoreFn.hs
```diff
@@ -47,6 +47,15 @@ spec = context "CoreFnFromJsonTest" $ do
       ss = SourceSpan mp (SourcePos 0 0) (SourcePos 0 0)
       ann = ssAnn ss
 
+  specify "should parse version" $ do
+    let v = Version [0, 13, 6] []
```
paulyoung, line 51: With the old implementation, using `let v = Version [0] []` here still allowed this test to pass.

src/Language/PureScript/CoreFn/FromJSON.hs
```diff
@@ -12,14 +12,14 @@ import           Data.Aeson
...
-import           Text.ParserCombinators.ReadP (readP_to_S)
 import qualified Data.Vector as V
-import           Data.Version (Version, parseVersion)
+import           Data.Version (Version)
 
 import           Language.PureScript.AST.SourcePos (SourceSpan(SourceSpan))
 import           Language.PureScript.AST.Literals
 import           Language.PureScript.CoreFn.Ann
 import           Language.PureScript.CoreFn
+import           Language.PureScript.Docs.Types (parseVersion')
```
kritzcreek, line 22: That dependency seems the wrong way around. Could we move that `parseVersion'` somewhere both the Docs and CoreFn can import from?

paulyoung: No problem. Do you have any suggestions on where to put it?

kritzcreek: I've looked around a bit and I can't come up with anything better than just putting it into `CoreFn.FromJSON`. @hdgarrood any preferences?

hdgarrood: I feel like we should just dump this sort of stuff - stuff we kind of wished was provided by upstream libraries - into a Language.PureScript.Utils module. There are some modules in the compiler already for this sort of stuff which don’t use the Language.PureScript namespace but they really ought to, imo. But that’s probably separate... for now I don’t mind having one of these import the other. I don’t have a preference for which way around they are.

kritzcreek: Allright, in that case I'd prefer the function lived here. I see ide/docs/publish as downstream from the compilers PoV.

hdgarrood:
Thanks!

Issue #3875, opened 2020-05-15
Coercible constraints on rows aren’t solved


kl0tl:
## Description

Rows are only considered when inferring roles for constructors holding types parametrized by rows.

Given the following declaration of `Rec1`:

https://github.com/purescript/purescript/blob/df3170f5bb0ba21517e2ca72db6aee5c479c99df/tests/purs/passing/Coercible.purs#L103

When inferring its roles we discover its only constructor holds a `Record (f :: a)`, which is representational in its row so we consider the types in the row (here the type variable `a`) and infer the `a` parameter of `Rec1` to be representational.

And indeed we can coerce `Rec1 a` to `Rec1 b` given `Coercible a b`:

https://github.com/purescript/purescript/blob/df3170f5bb0ba21517e2ca72db6aee5c479c99df/tests/purs/passing/Coercible.purs#L105-L106

But we can’t coerce the underlying records themselves.

## To Reproduce

```purs
module Example where

import Safe.Coerce (coerce)

newtype N a = N a

example :: { x :: Int } -> { x :: N Tnt }
example = coerce
```

```
Error found:
in module Example
at Example.purs:8:11 - 8:17 (line 8, column 11 - line 8, column 17)

  No type class instance was found for
                                      
    Prim.Coerce.Coercible ( x :: Int  
                          )           
                          ( x :: N Int
                          )           
                                      

while checking that type forall (a :: Type) (b :: Type). Coercible a b => a -> b
  is at least as general as type { x :: Int     
                                 }              
                                 -> { x :: N Int
                                    }           
while checking that expression coerce
  has type { x :: Int     
           }              
           -> { x :: N Int
              }           
in value declaration example

See https://github.com/purescript/documentation/blob/master/errors/NoInstanceFound.md for more information,
or to contribute content related to this error.
```

## Expected behavior

`coercibleWanteds` in `Language.PureScript.TypeChecker.Entailment` should decompose rows into `Coercible` constraints for types with the same labels in both rows and fail when the rows length or labels don’t match.

For instance `Coercible ( x :: Int ) ( x :: N Int )` should yield the subgoal `Coercible Int (N Int)` and `Coercible ( x :: Int, y :: String ) ( x :: N Int, y :: N String)` should yield the subgoals `Coercible Int (N Int)` and `Coercible String (N String)` but `Coercible ( x :: Int ) ()` and `Coercible ( x :: Int ) ( y :: N Int )` should fail.

## PureScript version

[...]

PR #3878, opened 2020-05-16
Solve `Coercible` constraints on rows

kl0tl:
Close https://github.com/purescript/purescript/issues/3875.

### DIFF
--- a/src/Language/PureScript/Docs/Prim.hs
+++ b/src/Language/PureScript/Docs/Prim.hs
@@ -424,6 +424,13 @@ coercible = primClassOf (P.primSubName "Coerce") "Coercible" $ T.unlines
   , "This rule may seem puzzling since it is impossible to apply `coerce` to a term"
   , "of type `NT1` but it is necessary to coerce types with higher kinded parameters."
   , ""
+  , "Fifth, every pair of rows can be coerced if they have the same labels"
+  , "and the corresponding types for each label are coercible:"
+  , ""
+  , "    instance coercibleRow :: Coercible a b => Coercible ( label :: a ) ( label :: b )"
+  , ""
+  , "Closed rows can't be coerced to open rows and open rows can only be"
+  , "coerced to open rows with the same row variable."
   ]
 
 kindOrdering :: Declaration
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -427,6 +427,21 @@ entails SolverOptions{..} constraint context hints =
           let a' = foldl' srcTypeApp a tys
               b' = foldl' srcTypeApp b tys
           pure $ Just [srcCoercibleConstraint kind a' b']
+      -- If both arguments have kind @Row k@, yield new wanted constraints
+      -- in terms of pairs of types with the same label in both rows (e.g.
+      -- @Coercible D D'@ given @Coercible ( label :: D ) ( label :: D' )@)
+      -- and fail when some labels are exclusive to one row (e.g. @extra@
+      -- in the constraint @Coercible () ( extra :: D )@) or when the tails
+      -- don’t unify (e.g. @()@ and @r@ in @Coercible () ( | r )@ or
+      -- @r@ and @s@ in @Coercible ( | r ) ( | s )@).
+      | RCons _ _ ty _ <- a = do
+          k <- elaborateKind ty
+          case alignRowsWith (srcCoercibleConstraint k) a b of
+            (constraints, (([], tail1), ([], tail2))) -> do
+              rethrow (const . errorMessage . NoInstanceFound $ srcCoercibleConstraint (kindRow k) a b) $ unifyTypes tail1 tail2
+              pure $ Just constraints
+            (_, (rl1, rl2)) ->
+              throwError . errorMessage $ TypesDoNotUnify (rowFromList rl1) (rowFromList rl2)
       | (TypeConstructor _ aTyName, _, axs) <- unapplyTypes a
       , (TypeConstructor _ bTyName, _, bxs) <- unapplyTypes b
       , not (null axs) && aTyName == bTyName
--- a/tests/purs/failing/CoercibleClosedRowsDoNotUnify.out
+++ b/tests/purs/failing/CoercibleClosedRowsDoNotUnify.out
@@ -1,0 +1,39 @@
+Error found:
+in module [33mMain[0m
+at tests/purs/failing/CoercibleClosedRowsDoNotUnify.purs:7:12 - 7:18 (line 7, column 12 - line 7, column 18)
+
+  Could not match type
+  [33m            [0m
+  [33m  ( x :: Int[0m
+  [33m  ...       [0m
+  [33m  )         [0m
+  [33m            [0m
+  with type
+  [33m               [0m
+  [33m  ( y :: String[0m
+  [33m  ...          [0m
+  [33m  )            [0m
+  [33m               [0m
+
+while solving type class constraint
+[33m                                     [0m
+[33m  Prim.Coerce.Coercible ( x :: Int   [0m
+[33m                        )            [0m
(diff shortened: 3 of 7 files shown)


src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -452,9 +468,15 @@ entails SolverOptions{..} constraint context hints =
             -- @Coercible (T[X/a]) b = Coercible X b@ in the example).
             let wrappedTySub = replaceAllTypeVars (zip tvs xs) wrappedTy
             pure [Constraint nullSourceAnn C.Coercible [] [wrappedTySub, b] Nothing]
+        | RCons{} <- a
+        , RCons{} <- b -> do
+          let (constraints, (rl1, rl2)) = alignRowsWith (\a' b' -> Constraint nullSourceAnn C.Coercible [] [a', b'] Nothing) a b
+          unless (null (fst rl1) && null (fst rl2)) $
+            throwError . errorMessage $ TypesDoNotUnify (rowFromList rl1) (rowFromList rl2)
```
natefaubion, line 475: I think we should add tests for coercing polymorphic rows.
```purescript
test :: forall r. { a :: Int | r } -> { a :: Id1 Int | r }
test = coerce
```
I don't think there's a reason this shouldn't coerce since the tails unify.

```purescript
test :: forall r s. { a :: Int | r } -> { a :: Id1 Int | s }
test = coerce
```
I'd think this needs to introduce a subgoal for `Coercible r s`, but `Coercible` isn't polykinded, so I assume we'll need to reject this with a unification error.

hdgarrood: Should we consider having special casing for records so that it introduces a subgoal `Coercible (Record r) (Record s)`?

natefaubion: I think a unification error is fine, since it's likely an error anyway. For example you couldn't write:
```
test :: foral r s. { a :: Int | r } -> { a :: Int | s }
test = identity
```
The way to solve that would be a type equality that `r ~ s`.

hdgarrood: Right, that makes sense.

natefaubion: Actually, maybe a unification error isn't right, and we should just punt the whole constraint? It's possible that it could be solved at the call site.

hdgarrood: I'm not sure I have fully appreciated the difference yet, to be honest.

natefaubion: At the call-site `r` could unify with `s`, which would make the constraint solvable.

hdgarrood: I mean, if we are in this part of the code already, we are trying to solve a Coercible constraint, does that not mean we are already at the call site? I'm not sure I understand how/why we have a choice.

natefaubion: Given
```purescript
test :: forall r s. { a :: Int | r } -> { a :: Id1 Int | s }
test = coerce
```
We are solving the instantiation of `coerce`. With the provided signature, the constraint is not solvable because `r` does not unify with `s`. We shouldn't throw a unification error because it's possible that it could be solved, given that `r` _does_ unify with `s`. That is, this is potentially solvable:
```purescript
test :: forall r s. Coercible { a :: Int | r } { a :: Id1 Int | s } => { a :: Int | r } -> { a :: Id1 Int | s }
test = coerce

test2 = test { a: 42 } :: { a :: Id1 Int }
```
because now `r` and `s` unify with `()`. We should not throw a unification error, we should just defer the constraint.

natefaubion: Note that I think it's fine to throw a unification error for the existing case as written, because we know for a fact there is no solution, since there are concrete rows the differ. I'm only talking about polymorphic tails.

hdgarrood: Right, I see, thanks. That approach seems a bit odd to me because then `test` effectively has an invisible `Coercible` constraint, right? If I wrote a type annotation `forall r s. { a :: Int | r } -> { a :: Id1 Int | s }` then I would want the compiler to tell me if that function can’t actually be used for all possible choices of `r` and `s`; this feels like a parametricity violation to me.

natefaubion: I'm not sure what you mean by "an invisible Coercible constraint". I'm just saying that the compiler should not fail with a unification error, it should fail saying that it can't solve the constraint, and so the user should add the constraint.

hdgarrood: Oh! I see now. That sounds reasonable to me, yes.

kl0tl:
I wasn’t considering rows tails so open rows were coercible regardless of whether their tails unified 🙃 

I rethrown a `NoInstanceFound` error instead of a `TypesDoNotUnify` error when rows tails don’t unify and added tests for open rows, as suggested by @natefaubion.

Also I special cased records instead of rows because `Coercible` isn’t polykinded so mentioning coercible constraints on rows such as

```
Prim.Coerce.Coercible ( x :: Int
                      )
                      ( y :: String
                      )
```

in error messages is misleading: the compiler throws a `KindsDoNotUnify` error when writing such constraint. Should we check the kinds of type constructors arguments so we don’t try to solve coercible constraints on types of kind other than `Type`?

This means that we can’t solve coercible constraints on user written types parametrized by rows though, such as `Data.Variant.Variant` for instance. I don’t know the consequences of making `Coercible` polykinded but we could maybe introduce another class for coercing rows?

natefaubion:
I think the case for records is fine for now. I don't think it should hold up a release of it, and we can continue to discuss solutions in a separate issue.

hdgarrood:
This looks good to me too, but I'd quite like to get #3873 in first since I think it addresses a more fundamental problem with how we're dealing with roles inside the compiler, if that sounds ok?

hdgarrood:
> Should we check the kinds of type constructors arguments so we don’t try to solve coercible constraints on types of kind other than Type?

> This means that we can’t solve coercible constraints on user written types parametrized by rows though, such as Data.Variant.Variant for instance. I don’t know the consequences of making Coercible polykinded but we could maybe introduce another class for coercing rows?

These are really good questions. I was thinking about them a bit more and this led me to discover #3889. Shall we continue this conversation there?

hdgarrood:
I would like to come back to this once #3893 is in, since that addresses a more fundamental issue with Coercible instances for parameterized data types, and will probably have implications for the approach we want to take here.

kl0tl:
I’ve rebased this branch! The changes to the instance solver are quite minimal now that `Coercible` constraints are polykinded, that’s nice.

hdgarrood (approved):

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -424,6 +424,13 @@ coercible = primClassOf (P.primSubName "Coerce") "Coercible" $ T.unlines
   , "This rule may seem puzzling since it is impossible to apply `coerce` to a term"
   , "of type `NT1` but it is necessary to coerce types with higher kinded parameters."
   , ""
+  , "Fifth, every pair of rows can be coerced if the types they contain with"
```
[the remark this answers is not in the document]
kl0tl: That‘s better indeed, I‘ve applied your suggestion.

hdgarrood:
Thanks! Would you mind merging master so that CI will hopefully pass?

kl0tl:
I rebased on top of master, tests pass again 😌

hdgarrood:
I think I'll leave this open for a little bit in case anyone with better knowledge of the solver wants to weigh in; I'm not particularly familiar with it right now.

hdgarrood:
I'm going to merge this for now so that we can get an RC out.

Issue #3876, opened 2020-05-16
Can declare roles multiple times


kl0tl:
## Description

Multiple roles declarations are authorized but only the last one is considered.

## To Reproduce

```purs
module Example where

import Safe.Coerce (coerce)

data A a = A
type role A nominal
type role A phantom

example :: ∀ a b. A a -> A b
example = coerce
```

## Expected behavior

The compiler shouldn’t allow duplicate role annotations.

GHC fails with the following message in that situation:

```
Example.hs:6:1: error:
    Duplicate role annotations for ‘A’:
      type role A nominal -- written at Example.hs:6:1-19
      type role A phantom -- written at Example.hs:7:1-19
  |
6 | type role A nominal
  | ^^^^^^^^^^^^^^^^^^^
```

## PureScript version

master (df3170f5bb0ba21517e2ca72db6aee5c479c99df)

hdgarrood:
Thanks! I think we should check for this in a desugaring step at the same time as checking that any referenced types actually exist (#3870).

Issue #3870, opened 2020-05-13
Can declare roles for a non-existent type


hdgarrood:
## Description

The compiler does not check that role declarations reference types that exist. For example:

```purescript
module Main where
type role Lol phantom
```

## To Reproduce

Include a role declaration (`type role ...`) for a non-existent type.

## Expected behavior

The compiler throws an error. I think we could probably introduce a new `OrphanRoleDeclaration` error to go with `OrphanTypeDeclaration` and `OrphanKindDeclaration`.

## PureScript version

master

---

I'm currently working on refactoring how we handle roles so that it's a bit closer to how we handle other kinds of declarations, so that we infer them once and add them to the `Environment` when we're done (see https://github.com/purescript/purescript/issues/3867#issuecomment-626398322). I think this should be relatively easy to fix after that.

hdgarrood:
Actually we should probably just do this the same way we do `OrphanTypeDeclaration` and `OrphanKindDeclaration`, both of which are fairly self-contained (in Sugar.TypeDeclarations). So this probably doesn't need to wait for me to address #3867.

hdgarrood:
Related: the number of type arguments doesn't have to match.

```purescript
data Unit = Unit
type role Unit phantom phantom phantom
```

hdgarrood:
We should also check that role declarations are only made for types where that makes sense. For instance, declaring roles on a type synonym or a class shouldn't be allowed; for type synonyms, we should expand them and look at the roles in the expanded type, and for classes, all of their parameters must be treated as nominal, so there's no point in providing role declarations.

PR #3881, opened 2020-05-18
Check role declarations make sense

kl0tl:
This PR enforces the following checks:
+ Role declarations cannot reference non declared types.
+ Role declarations must match the number of arguments of their corresponding type.
+ Role declarations are not allowed for type classes nor type synonyms.
+ Duplicate role declarations are not be allowed.

~I’m unsure about the locations of those checks. Should they be in `Language.PureScript.Sugar.RoleDeclarations` rather than `Language.PureScript.Sugar.TypeDeclarations`? Role declarations are not desugared to anything though, so maybe they should be in `Language.PureScript.TypeChecker` instead?~ I’ve moved those checks to the type checker.

Close https://github.com/purescript/purescript/issues/3870 and close https://github.com/purescript/purescript/issues/3876.

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -184,6 +184,7 @@ data SimpleErrorMessage
   | InvalidCoercibleInstanceDeclaration [SourceType]
   | UnsupportedRoleDeclaration
   | RoleDeclarationArityMismatch (ProperName 'TypeName) Int Int
+  | DuplicateRoleDeclaration (ProperName 'TypeName)
   deriving (Show)
 
 data ErrorMessage = ErrorMessage
@@ -344,6 +345,7 @@ errorCode em = case unwrapErrorMessage em of
   InvalidCoercibleInstanceDeclaration {} -> "InvalidCoercibleInstanceDeclaration"
   UnsupportedRoleDeclaration {} -> "UnsupportedRoleDeclaration"
   RoleDeclarationArityMismatch {} -> "RoleDeclarationArityMismatch"
+  DuplicateRoleDeclaration {} -> "DuplicateRoleDeclaration"
 
 -- | A stack trace for an error
 newtype MultipleErrors = MultipleErrors
@@ -1329,6 +1331,9 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
         , T.pack (show actual)
         , if actual > 1 then "roles" else "role"
         ] <> "."
+
+    renderSimpleErrorMessage (DuplicateRoleDeclaration name) =
+      line $ "Duplicate role declaration for " <> markCode (runProperName name) <> "."
 
     renderHint :: ErrorMessageHint -> Box.Box -> Box.Box
     renderHint (ErrorUnifyingTypes t1@RCons{} t2@RCons{}) detail =
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -72,6 +72,7 @@ data SimpleErrorMessage
   | MultipleTypeOpFixities (OpName 'TypeOpName)
   | OrphanTypeDeclaration Ident
   | OrphanKindDeclaration (ProperName 'TypeName)
+  | OrphanRoleDeclaration (ProperName 'TypeName)
   | RedefinedIdent Ident
   | OverlappingNamesInLet
   | UnknownName (Qualified Name)
@@ -237,6 +238,7 @@ errorCode em = case unwrapErrorMessage em of
   MultipleTypeOpFixities{} -> "MultipleTypeOpFixities"
   OrphanTypeDeclaration{} -> "OrphanTypeDeclaration"
   OrphanKindDeclaration{} -> "OrphanKindDeclaration"
+  OrphanRoleDeclaration{} -> "OrphanRoleDeclaration"
   RedefinedIdent{} -> "RedefinedIdent"
   OverlappingNamesInLet -> "OverlappingNamesInLet"
   UnknownName{} -> "UnknownName"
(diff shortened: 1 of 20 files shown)


lib/purescript-ast/src/Language/PureScript/Environment.hs
```diff
@@ -377,6 +377,13 @@ isTypeOrApplied :: Type a -> Type b -> Bool
 isTypeOrApplied t1 (TypeApp _ t2 _) = eqType t1 t2
 isTypeOrApplied t1 t2 = eqType t1 t2
 
+kindArity :: SourceType -> Int
+kindArity = go' 0 where
+  go' n (TypeApp _ (TypeApp _ k _) b)
+    | k == tyFunction = go' (n + 1) b
+  go' n (ForAll _ _ _ k _) = go' n k
+  go' n _ = n
```
kl0tl, line 385: I’m not sure this is the right module for this function.

hdgarrood:
Thanks! I will review after the other roles PRs are merged.

hdgarrood:
The grouping code you have here feels like a bit of a code smell to me, since we're already doing grouping based on which declarations refer to each other in the BindingGroups desugaring step. It feels to me like these checks ought to be relatively easy to implement alongside the existing BindingGroups desugaring step:
* Role declarations must refer to declarations which exist in the same module
* Role declarations must refer to data declarations
* Role declarations may not be duplicated

so perhaps we should have all of the above in the BindingGroups step and only have the arity check in the type checker? I think if we were to go ahead and implement the first refactoring I discussed in https://github.com/purescript/purescript/pull/3910#issuecomment-680252963 that might simplify this code as well. Would you mind if I took a stab at that refactoring first?

kl0tl:
Sure, go ahead!

hdgarrood:
I had a go at doing that refactoring and had some thoughts:

The graph in the BindingGroups step is already very subtle in terms of what the vertices are and which vertices should depend on which, especially in the presence of kind declarations and type synonyms, and I think introducing roles into the mix as well is likely to make that code quite brittle and difficult to change. On reflection, the current approach of putting all the role declarations before any data declarations at the very end of the desugaring phase and then storing roles temporarily in the Environment when we encounter those role declarations during type checking isn't really all that bad.

I think if I was going to try to refactor this to remove the need to put role declarations in the environment, then I'd modify the AST so that we can store role annotation data directly in the `DataDeclaration` constructor with a new constructor argument. However, there may be drawbacks to that I'm not aware of, and there's also the issue that these are positional arguments which we pattern match on everywhere, so there would be a ton of places to update, which probably wouldn't be fun.

I also realised that the three checks for role declarations which I recommended adding to the BindingGroups step (must refer to a declaration in the same module, must refer to a data or foreign import data declaration, must not be duplicated) are really more analogous to the checks in the Sugar.TypeDeclarations step, which perform those same checks for type and kind signatures, so I think they would be best off going there. I think the simplest way of doing that would be to require that a role annotation directly follows the data type it describes, like how we require that type and kind signatures directly precede the declarations they describe. That does make things more restrictive than they are now, but I suspect most of the code which has been written already follows this rule, and I think it's a sensible rule to impose.

kl0tl:
I moved the checks back to `Language.PureScript.Sugar.TypeDeclarations` and strengthened the orphan check to throw unless role declarations directly follow their type. This is pleasantly simpler than it used to, thank you for the suggestion!

src/Language/PureScript/Errors.hs
```diff
@@ -1306,6 +1316,25 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
+    renderSimpleErrorMessage UnsupportedRoleDeclaration =
+      line $ "Role declarations are only supported for data types, not for type synonyms nor type classes."
+
+    renderSimpleErrorMessage (RoleDeclarationArityMismatch name expected actual) =
+      line $ T.intercalate " "
+        [ "The type"
+        , markCode (runProperName name)
+        , "expects"
+        , T.pack (show expected)
+        , if expected > 1 then "arguments" else "argument"
```
hdgarrood, line 1328: When there are no arguments, the word "argument" should be pluralized, so I think the condition we want here is `if expected == 1 then "argument" else "arguments"`

src/Language/PureScript/Errors.hs
```diff
@@ -1306,6 +1316,25 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
+        , "but its role declaration list"
```
hdgarrood, line 1329: To make the verb "list" agree, it should be written "lists", so for example: "but its role declarations lists only 3 roles".

src/Language/PureScript/Errors.hs
```diff
@@ -1306,6 +1316,25 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
+            <> if actual > expected then "" else " only"
```
hdgarrood, line 1330: Adding "only" when there are not enough roles is a nice touch :+1:

hdgarrood (commented):
Looks great, just a couple of grammar comments regarding one of the errors

src/Language/PureScript/Errors.hs
kl0tl: This is something I picked up from the `ClassInstanceArityMismatch` error ^^

kl0tl:
I applied your suggestions.

hdgarrood (approved):
Thanks! Just for future reference, would you mind pushing new commits and leaving the existing commits alone when responding to code review feedback? It makes it easier for reviewers to see what has changed since the last review. The commit history will be tidied up automatically by the "Squash and merge" button.

kl0tl:
Sorry about that, I’m trying to get into the habit of pushing fixup commits instead. In the case of whole commits, do you prefer them to be removed from the history or reverted by new commits?

hdgarrood:
No worries - it's not a big deal, so don't worry if you forget. I think pushing new commits which revert the old unwanted ones is preferable, as that makes it easier to see what changed since the last time you looked.

PR #3883, opened 2020-05-19
Have module re-exports appear in generated code

### LINKED ISSUE #1888: Have module reexports appear in generated code
```
$ psc --version
0.8.0.0
```

To reproduce, `pulp init` a new project, then:

``` haskell
--src/Main.purs
module Main 
  ( module A
  , main
  ) where

import A
import Control.Monad.Eff
import Prelude

main :: Eff () Unit
main = pure unit

--src/A.purs
module A 
  ( foo
  ) where

foo :: String
foo = "Hello thar"
```

```
$ psc src/Main.purs
```

Contents of output/Main/index.js:

``` Javascript
// Generated by psc version 0.8.0.0                                     
"use strict";                                                           
var A = require("A");                                                   
var Control_Monad_Eff = require("Control.Monad.Eff");                   
var Prelude = require("Prelude");                                       
var main = Prelude.pure(Control_Monad_Eff.applicativeEff)(Prelude.unit);
module.exports = {                                                      
    main: main                                                          
};                                                                      
```

Would expect `foo: A.foo` to appear next to `main` in the exports. I believe this was working in 0.7.6.1, I was using re-exported modules in the browser.

citizengabe:
Resolves #1888. See that issue and this [discussion](https://discourse.purescript.org/t/intention-of-breaking-change-handling-re-exports-in-corefn/1405) on Discourse for background.

The changes in this PR are two-fold:
1. Add a new field on `Language.PureScript.CoreFn.Module` to hold re-export data so it can be used during code generation
2. Use that data to import the necessary modules and export the relevant members in the generated code.

I added a test for parsing the re-exports, but no others. Is there somewhere we test the output of `Language.PureScript.CodeGen.JS.moduleToJs`?

I'd also like to add this GitHub account to `CONTRIBUTORS.md`, but am waiting on guidance from my company as to how to reference _it_ as this work was done on company time.


citizengabe:
I have some tests that are failing in CI that weren't failing locally. I cleared `.psci_modules`, `.test_modules`, and `output` and ran again locally and got similar failures. I'll investigate and push up a fix.

citizengabe:
It was an issue with bundling. I was handling regular dot notation for exported member access, but not bracket notation.

citizengabe:
> I'd also like to add this GitHub account to `CONTRIBUTORS.md`, but am waiting on guidance from my company as to how to reference it as this work was done on company time.

This is done, so the PR is ready to merge pending review.

kritzcreek:
I'd also like to check that all the re-exporting logic in the IDE still works with this change. Please don't merge before I've approved. I'll check this weekend!

citizengabe:
> I'd also like to check that all the re-exporting logic in the IDE still works with this change. Please don't merge before I've approved. I'll check this weekend!

That sounds prudent, but know that the changes are additive only. Nothing was changed WRT externs, and `moduleReExports` is an additional field. The rest of `Language.PureScript.CoreFn.Module.Module` remains unchanged.

kritzcreek:
Yeah, after checking none of these changes affect the IDE. All good from my side.

src/Language/PureScript/CoreFn/Desugar.hs
```diff
@@ -38,11 +38,27 @@ moduleToCoreFn env (A.Module modSS coms mn decls (Just exps)) =
...
+        (fromMaybe mn' (A.exportSourceImportedFrom src))
+        ref
+    _
+      | mn' == mn -> M.empty
+      | otherwise -> case ref' of
+        A.TypeRef _ _ (Just dctors) -> M.singleton mn' (fmap properToIdent dctors)
+        A.ValueRef _ name -> M.singleton mn' [name]
+        A.TypeClassRef _ name -> M.singleton mn' [properToIdent name]
+        A.TypeInstanceRef _ name -> M.singleton mn' [name]
+        _ -> M.empty
```
hdgarrood, line 61: Could you please add explicit cases for each DeclarationRef constructor here? That way, if we do add a new constructor or something, we'll get a warning here, so that we know to check whether this needs updating.

src/Language/PureScript/CoreFn/Desugar.hs
```diff
@@ -38,11 +38,27 @@ moduleToCoreFn env (A.Module modSS coms mn decls (Just exps)) =
   let imports = mapMaybe importToCoreFn decls ++ fmap (ssAnn modSS,) (findQualModules decls)
       imports' = dedupeImports imports
       exps' = ordNub $ concatMap exportToCoreFn exps
+      reExps = M.map ordNub $ M.unionsWith (++) (map (reExportsToCoreFn mn) exps)
```
hdgarrood, line 41: Should we be passing all of `exps` to this function? I understand that you have the `mn == mn'` guard in `reExportsToCoreFn` preventing normal exports being regarded as re-exports, but I feel like it might be simpler to only pass re-exports in in the first place. Would it be possible to write this via something like
`mapMaybe (fmap reExportsToCoreFn . toReExportRef)`, where `toReExportRef` is defined as something like
```
toReExportRef ref = case ref of
  A.ReExportRef _ src ref -> Just (src, ref)
  _ -> Nothing
```

CONTRIBUTORS.md
```diff
@@ -148,14 +148,16 @@ If you would prefer to use different terms, please use the section below instead
...
-| [@qfpl](https://github.com/qfpl) | qfpl @ Data61 / CSIRO | Our existing contributions to the PureScript compiler and all future contributions to the PureScript compiler until further notice, are Copyright Data61 / CSIRO, and are licensed to the owners and users of the PureScript compiler project under the terms of the [MIT license](http://opensource.org/licenses/MIT). - [@lightandlight](https://github.com/lightandlight) |
+| Username                                 | Company               | Terms                                                                                                                                                                                                                                                                                                                                                                        |
```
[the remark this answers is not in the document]
citizengabe: Sorry about that, I think it was my editor.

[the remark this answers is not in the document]
hdgarrood: No worries - thank you!

src/Language/PureScript/CoreFn/Desugar.hs
citizengabe: The reason I'm passing all of `exps` to `reExportsToCoreFn` is because I call the function recursively to handle the descending into `ref`. I can reformulate it in terms of `exportToCoreFn` though and only pass in the re-exports.

hdgarrood: Right - I would prefer to avoid recursion if we can. Did you come across cases in which there is a ReExportRef of a ReExportRef? I don't think that should happen but I might be wrong.

src/Language/PureScript/CoreFn/Desugar.hs
citizengabe: Since I'm rewriting `reExportsToCoreFn` in terms of `exportToCoreFn`, I can add the explicit cases in that function. Sound good?

hdgarrood: Yes, that's ideal!

src/Language/PureScript/CoreFn/Desugar.hs
citizengabe: Could that happen when re-exporting a re-export?

citizengabe: I guess I can just write a test and find out 😄

hdgarrood: The types don't rule it out, but I think what should happen is that re-exports should always be of the form `ReExportRef _ _ ref`, where `ref`'s constructor is something other than `ReExportRef`. See elaborateExports in Sugar.Names, that's where this data is assembled.

hdgarrood: Multiple layers of re-exports are tracked via the `ExportSource` field.

citizengabe: > See elaborateExports in Sugar.Names, that's where this data is assembled

Doesn't look like [`elaborateExports`](https://github.com/purescript/purescript/blob/f6e9a7bb10a9d938bc4c3668f2793bfc54b636e1/src/Language/PureScript/Sugar/Names.hs#L140-L143) will nest `ReExportRef`s. Looks like we're good.

hdgarrood:
Don't worry about it now, but for future reference could you please just push new commits to the branch after review? I'd like to be able to review what's changed since I last looked, and we will tidy up the history when merging.

citizengabe:
> Don't worry about it now, but for future reference could you please just push new commits to the branch after review?

Sure thing!

hdgarrood:
Looks great! If you could write a test along the lines of what I wrote here: https://github.com/purescript/purescript/pull/3883#pullrequestreview-417385565 that would be ideal, then I'm happy to approve.

hdgarrood:
Oh wait, sorry, just a couple more cases that I've just remembered about that I think would be worth testing:
- re-exporting a re-export
- re-exporting something from a module if we have a data constructor with the same name in scope (to make sure this continues to work well together with the module name renaming `mnLookup` thing, in CodeGen/JS.hs), eg

```purescript
module B (module A) where

import A
data A = A
```
to make sure the exports list refers to `A_1`, which is what the module `A` should be imported as in the JS, rather than just `A`.

citizengabe:
Would you prefer separate test suites for the different cases or is a single suite sufficient?

hdgarrood:
Whatever's easiest - I think putting them all under `tests/purs/passing/ReExportsExported` is fine.

citizengabe:
@hdgarrood is there anything preventing this PR from being merged?

hdgarrood:
Our policy is that PRs require at least two core team approvals (if the person who submits the PR is already a core team member, that counts).

citizengabe:
@kritzcreek You took a look a couple of weeks ago. Does it still look good?

kritzcreek:
> You took a look a couple of weeks ago. Does it still look good?

Sorry, I was on vacation last week. Change still looks good to me :+1:

citizengabe:
Thank you both for the reviews!

> Sorry, I was on vacation last week.

Hope it was a good one 😄

hedefalk:
Oooh, think I struck this and asked in slack https://functionalprogramming.slack.com/archives/C717K38CE/p1592399197296900

Does this solve the issue I'm having right now when trying to re-export for commonjs:

Foo.purs:
```purs
module Foo ( module Foo ) where
import Bar (bar) as Foo
```
Bar.purs:
```purs
module Bar where
bar :: String
bar = "bar"
```

`npx spago bundle-module --main Foo` gives me:

```js
// Generated by purs bundle 0.13.8
var PS = {};
module.exports = PS["Foo"];
```

PR #3887, opened 2020-05-23
Printer for CST Modules

kritzcreek:
Adds flatten functions for all constructions in the CST, that turn a module back into its individual tokens.

Still need to make sure this is proper bidirectional by running it over all our test files

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
@@ -18,11 +18,216 @@ flattenModule (Module _ a b c d e f g) =
     dummyPos = SourcePos 0 0
     dummyRange = SourceRange dummyPos dummyPos
 
+flattenDataHead :: DataHead a -> DList SourceToken
+flattenDataHead (DataHead a b c) = pure a <> flattenName b <> foldMap flattenTypeVarBinding c
+
+flattenDataCtor :: DataCtor a -> DList SourceToken
+flattenDataCtor (DataCtor _ a b) = flattenName a <> foldMap flattenType b
+
+flattenClassHead :: ClassHead a -> DList SourceToken
+flattenClassHead (ClassHead a b c d e) =
+  pure a <>
+  foldMap (\(f, g) -> flattenOneOrDelimited flattenConstraint f <> pure g) b <>
+  flattenName c <>
+  foldMap flattenTypeVarBinding d <>
+  foldMap (\(f, g) -> pure f <> flattenSeparated flattenClassFundep g) e
+
+flattenClassFundep :: ClassFundep -> DList SourceToken
+flattenClassFundep = \case
+  FundepDetermined a b ->
+    pure a <> foldMap flattenName b
+  FundepDetermines a b c ->
+    foldMap flattenName a <> pure b <> foldMap flattenName c
+
+flattenInstance :: Instance a -> DList SourceToken
+flattenInstance (Instance a b) =
+  flattenInstanceHead a <> foldMap (\(c, d) -> pure c <> foldMap flattenInstanceBinding d) b
+
+flattenInstanceHead :: InstanceHead a -> DList SourceToken
+flattenInstanceHead (InstanceHead a b c d e f) =
+  pure a <>
+  flattenName b <>
+  pure c <>
+  foldMap (\(g, h) -> flattenOneOrDelimited flattenConstraint g <> pure h) d <>
+  flattenQualifiedName e <>
+  foldMap flattenType f
+
+flattenInstanceBinding :: InstanceBinding a -> DList SourceToken
+flattenInstanceBinding = \case
+  InstanceBindingSignature _ a -> flattenLabeled flattenName flattenType a
+  InstanceBindingName _ a -> flattenValueBindingFields a
+
+flattenValueBindingFields :: ValueBindingFields a -> DList SourceToken
+flattenValueBindingFields (ValueBindingFields a b c) =
+  flattenName a <>
+  foldMap flattenBinder b <>
+  flattenGuarded c
+
+flattenBinder :: Binder a -> DList SourceToken
+flattenBinder = \case
+  BinderWildcard _ a -> pure a
+  BinderVar _ a -> flattenName a
+  BinderNamed _ a b c -> flattenName a <> pure b <> flattenBinder c
+  BinderConstructor _ a b -> flattenQualifiedName a <> foldMap flattenBinder b
+  BinderBoolean _ a _ -> pure a
+  BinderChar _ a _ -> pure a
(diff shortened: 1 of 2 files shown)


lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
```diff
@@ -5,12 +5,274 @@ import Prelude
...
+  pure a <>
+  flattenName b <>
+  foldMap (flattenWrapped (flattenSeparated flattenExport)) c <>
+  pure d <>
+  foldMap flattenImportDecl e <>
+  foldMap flattenDeclaration f <>
+  pure (SourceToken (TokenAnn dummyRange g []) TokEof)
+  where
+    dummyPos = SourcePos 0 0
+    dummyRange = SourceRange dummyPos dummyPos
```
natefaubion, line 19: I wonder if the CST should keep the TokEof around instead of just the trailing module comments.

kritzcreek: Yeah I wondered about that as well...

natefaubion: You can use `moduleRange` from `CST.Position` to get the last position. The TokEof should have the last position twice.

natefaubion:
Note for printing purposes you will want to filter out:
* TokLayoutStart
* TokLayoutSep
* TokLayoutEnd
* TokEof

kritzcreek:
I do need the comments on `TokEof` though, if I don't want to lose trailing comments, right?

natefaubion:
Oh, that's right. The reason is that the provided token printer will print `<eof>` literally, which we use in the layout golden tests. You might need to add some customization to the printer.

lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
kritzcreek: `moduleRange` uses the endPosition of the last declaration in the module. Shouldn't the Eof range start after "its" leading comments?

natefaubion: Yes, you can. You can use `commentDelta` and `applyDelta` to get that position. It's not great, and I should have just included TokEof in the module, but you can derive the original information.

kritzcreek:
Allright, this seems to work on all the test files (apart from some unicode nonsense with Haskell on Windows in GHCI)

kritzcreek:
Anything in particular you'd like me to test here? I could check round tripping for all the layout tests.

natefaubion:
The layout tests don't necessarily parse, they only test the lexer.

natefaubion:
I feel like testing all passing source files is a little much, but maybe it's fast enough that it doesn't matter.

kritzcreek:
Allright.... there's no good test body in `purescript-cst` I could use then. I guess we could start adding to it whenever we find an issue where it doesn't roundtrip.

natefaubion:
Oh right, it's a separate module. I guess we could try it. If they all parse as a module, then that seems fine to me.

kritzcreek:
FWIW all of `tests/purs/passing` roundtrips.

After putting the following script into Script.hs in the project root it can be run with `stack runhaskell Script.hs`:
```haskell
{-# language BlockArguments #-}
import Prelude
import Data.Foldable
import Language.PureScript.CST.Types as CST
import Language.PureScript.CST.Parser as CST
import Language.PureScript.CST.Print as CST
import Data.Text (Text)
import Data.Text.IO as Text
import System.Directory
import System.FilePath

readModule :: FilePath -> IO (Text, CST.Module ())
readModule path = do
  contents <- Text.readFile path
  let module_ = either (error . show) id (snd (CST.parse contents))
  pure (contents, module_)

main :: IO ()
main = do
  go "tests/purs/passing"
  where
    go :: FilePath -> IO ()
    go dir = do
      listDirectory dir >>= traverse_ \path ->
        if isExtensionOf "purs" path then do
          print path
          (content, module_) <- readModule (dir </> path)
          if content == CST.printModule module_ then pure () else error path
        else if isExtensionOf "js" path then pure ()
        else go (dir </> path)

```

kritzcreek:
@joneshf Now that we have the `purescript-cst` library, how would I go about making a minor release of it that includes this change?

joneshf:
I'm sorry, I don't really understand the question. Could you rephrase it?

kritzcreek:
The point of coming up with the separate version scheme for the sub-packages like purescript-cst was so we could release changes and improvements to them without having to make a compiler release.

Now I'm wondering how to actually make that happen, because I'd like to use this change in an experiment of mine that doesn't depend on the whole compiler, but it doesn't seem like purescript-cst is published to Hackage.

kritzcreek:
I'm asking you because I remember you lobbying strongly for it. If you don't know what the process is, no worries.

hdgarrood:
We could make an initial release of `purescript-cst` now. We don't really have a procedure for that yet, since this would be the first one. Actually uploading to Hackage is pretty straightforward: I think it should just be `cd lib/purescript-cst && stack upload .`. We should probably also add a tag to this repository so that it's easy to identify which commit corresponds to which version, perhaps `purescript-cst-0.1.0.0`?

The only potential awkwardness is that this release would include syntax for kind signatures and role declarations, but those don't yet exist in any released compiler version. I'm not sure that's necessarily such a problem that we should wait until 0.14.0 is out, though.

hdgarrood:
Oh also, if you upload a release, could you please add the rest of us as maintainers? By default only the initial uploader will be permitted to upload subsequent versions.

hdgarrood:
We should probably add instructions to the RELEASE_GUIDE.md file for making releases of the purescript-ast and purescript-cst libraries at some point, too.

piq9117:
This is awesome! I was just going to ask about this in fp-chat if there was any plans to separate `purescript-cst` and `purescript-ast` but didn't wanna bother yall for some dump personal project I have, but I imagine this will probably be great for people who want to build tooling around purescript projects.

joneshf:
> The point of coming up with the separate version scheme for the sub-packages like purescript-cst was so we could release changes and improvements to them without having to make a compiler release.
> 
> Now I'm wondering how to actually make that happen, because I'd like to use this change in an experiment of mine that doesn't depend on the whole compiler, but it doesn't seem like purescript-cst is published to Hackage.

Got it, thanks for rephrasing! I would go with what @hdgarrood's laid out.

> I'm asking you because I remember you lobbying strongly for it. If you don't know what the process is, no worries.

Nah, I'm [pretty much indifferent](https://github.com/purescript/purescript/pull/3793#issuecomment-590087467) to what the versioning scheme is. My only preference is that whatever the scheme is, it fosters iteration. We can change it to whatever works, for all that I care.

Issue #3889, opened 2020-05-24
Hygiene issue in Coercible solving


hdgarrood:
## Description

Coercible's handling of parameterized newtypes seems to have a hygiene issue, in the sense that solving a Coercible constraint can depend on type parameter names.

## To Reproduce

The following program compiles:
```purescript
module Main where

import Prelude
import Safe.Coerce (coerce)

newtype Ap f a = Ap (f a)
-- inferred roles:
-- type role Ap representational nominal

newtype Id a = Id a
newtype Id2 a = Id2 a

test :: Ap Id Int -> Ap Id2 Int
test = coerce
```
Now if we change the definition of `Id2` to `newtype Id2 b = Id2 b`, we get an error:
```
Error found:
in module Main
at src/Main.purs:14:8 - 14:14 (line 14, column 8 - line 14, column 14)

  No type class instance was found for
                           
    Prim.Coerce.Coercible b
                          a
```

## Expected behavior

Coercible solving should not be affected by renaming parameters to a data type.

## Additional context

I haven't yet managed to exploit this to make an invalid coercion, only to fail to coerce where coercion should work. (The above program compiles with GHC regardless of type parameter names). However, I don't think we should rule out the possibility that this bug can lead to invalid coercions.

We do not appear to be considering type arguments to newtypes when solving Coercible. That is, when trying to solve `Coercible (Ap Id Int) (Ap Id2 Int)` in the above program, I think we proceed via

```
Coercible Id Id2
Coercible a Id2 -- we just take the type of the single field of the Id constructor here
Coercible a a -- same again, take the type of the single field of the Id2 constructor
-- done
```

This seems related to the question of whether Coercible should be polykinded. It looks like GHC's Coercible is polykinded: http://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Coerce.html

Initially my reaction to polykinded Coercible was that I wasn't sure if it made a whole lot of sense, since `Coercible a b` means "the types `a` and `b` have the same runtime representation", which doesn't really mean anything unless `a, b :: Type`. But the procedure described by the safe zero-cost coercions paper does seem to necessitate a polykinded Coercible, judging by the rule

>  For every type constructor `TC r p n`, where
> * r stands for TC’s parameters at role representational,
> * p for those at role phantom and
> * n for those at role nominal, [we have] the instance
>
>    instance Coercible r1 r2 ⇒ Coercible (TC r1 p1 n) (TC r2 p2 n)

[...]

PR #3893, opened 2020-05-30
Saturate higher kinded types in `Coercible` constraints

kl0tl:
This PR implements the _type application_ rule mentioned by the Safe Zero-cost Coercions for Haskell paper in section **2.8 Supporting higher order polymorphism**:

> ► If Coercible t1 t2, where t1, t2 :: k1 → k2, then Coercible (t1 x) (t2 x)

Fix https://github.com/purescript/purescript/issues/3889.


lib/purescript-ast/src/Language/PureScript/Environment.hs
```diff
@@ -377,6 +377,13 @@ isTypeOrApplied :: Type a -> Type b -> Bool
 isTypeOrApplied t1 (TypeApp _ t2 _) = eqType t1 t2
 isTypeOrApplied t1 t2 = eqType t1 t2
 
+kindArity :: SourceType -> Int
+kindArity = go' 0 where
+  go' n (TypeApp _ (TypeApp _ k _) b)
+    | k == tyFunction = go' (n + 1) b
+  go' n (ForAll _ _ _ k _) = go' n k
+  go' n _ = n
```
kl0tl, line 385: I would’ve liked to export this function from `Language.PureScript.Types` instead but there would then be a circular dependency because `kindArity` depends on `Language.PureScript.Environment.tyFunction` and `Language.PureScript.Environment` depends on `Language.PureScript.Types`.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -406,56 +406,65 @@ entails SolverOptions{..} constraint context hints =
...
     -- whose satisfaction will yield the goal.
     coercibleWanteds :: Environment -> SourceType -> SourceType -> Maybe [SourceConstraint]
-    coercibleWanteds env a b = case a of
-      TypeConstructor _ tyName -> do
-        -- If the first argument is a plain newtype (e.g. @newtype T = T U@ and
-        -- the constraint @Coercible T b@), look up the type of its wrapped
-        -- field and yield a new wanted constraint in terms of that type
-        -- (@Coercible U b@ in the example).
-        (_, wrappedTy, _) <- lookupNewtypeConstructor env tyName
-        pure [Constraint nullSourceAnn C.Coercible [] [wrappedTy, b] Nothing]
```
kl0tl, line 416: I took the liberty to simplify `coercibleWanteds` by removing the case expression because this first case is a simplification of when the newtype has parameters.

lib/purescript-ast/src/Language/PureScript/Types.hs
```diff
@@ -585,6 +587,17 @@ unapplyConstraints = go []
...
+saturate :: S.Set Text -> SourceType -> Int -> SourceType
+saturate bound ty n = fst . runSupply 0 $ do
+  tvs <- replicateM n freshTyVar
+  return $ foldl' srcTypeApp ty (srcTypeVar <$> tvs)
+  where
+  freshTyVar = do
+    tv <- ("t" <>) . T.pack . show <$> fresh
+    if tv `elem` bound
+      then freshTyVar
+      else return tv
```
kl0tl, line 599: I tried to generated sensible type variables names but perhaps `freshName` would be enough?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -406,56 +406,65 @@ entails SolverOptions{..} constraint context hints =
...
+          -- terms of that type with the type arguments substituted in (e.g.
+          -- @Coercible (T[X/a]) b = Coercible X b@ in the example).
+          let wrappedTySub = replaceAllTypeVars (zip tvs xs) wrappedTy
+          pure [srcCoercibleConstraint wrappedTySub b]
+      | otherwise =
         -- In all other cases we can't solve the constraint.
         Nothing
 
+    srcCoercibleConstraint :: SourceType -> SourceType -> SourceConstraint
+    srcCoercibleConstraint a b = srcConstraint C.Coercible [] [a, b] Nothing
```
natefaubion, line 466: If Coercible is PolyKinded, the kind applications need to be supplied. Currently you are always setting this to an empty `[]` which is problematic.

kl0tl: I’ve added a kind argument to `srcCoercibleConstraint` and the type class dictionaries returned by `solveCoercible` now hold the kinds list `forClassName` is called with. Can this affect the solving of `Coercible` instances in the absence of syntax for kind application?

natefaubion: In general when solving constraints, we need to unify kinds in substitutions https://github.com/purescript/purescript/blob/5e3f01d892bcf23a7c3775274e6fb8ba5a214851/src/Language/PureScript/TypeChecker/Entailment.hs#L288-L319

It is not _likely_ that solving a `Coercible` constraint and ignoring the kinds will cause problems in 99.9% of cases. I would not be surprised though if you can construct something that relies on kind unification in order to typecheck, however unlikely. Unification can be very non-local. I think that we should make an effort to construct kinds/type-correct terms though. It can only go wrong if you don't 😆

kl0tl: I tried to reject heterogeneously kinded `Coercible` constraints by unifying the arguments kinds with:

```hs
(_, kind) <- kindOf a
(_, kind') <- kindOf b
unifyKinds kind kind'
```

but a few tests then throw `UndefinedTypeVariable` errors.

I guess I should compute and apply a substitution like it’s done in `unifySubstKind` but I’m kind of lost here :grimacing:

kl0tl: This happened when inferring the kinds of higher kinded types saturated in `coercibleWanteds` because I wasn’t adding the generated type variables to the environment. I went with `freshTypeWithKind` instead to reuse the existing machinery.

hdgarrood:
@kl0tl I think this is probably the next biggest showstopping Coercible bug to fix? What do you think about getting this merged next?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -381,80 +390,96 @@ entails SolverOptions{..} constraint context hints =
...
       -- currently don't support higher-rank arguments in instance heads, term
       -- equality is a sufficient notion of "the same".
       if a' == b'
-        then pure [TypeClassDictionaryInScope [] 0 EmptyClassInstance [] C.Coercible [] [] [a, b] Nothing]
+        then pure [TypeClassDictionaryInScope [] 0 EmptyClassInstance [] C.Coercible [] kinds [a, b] Nothing]
         else do
+          lift $ do
+            (_, kind) <- kindOf a'
+            (_, kind') <- kindOf b'
+            unifyKinds kind kind'
```
natefaubion, line 409: Was this specifically needed? Sorry, I did not mean to imply we needed to unify kinds in `Coercible`, just that kind unification will happen and we need to propagate kind arguments to sub-goals.

kl0tl: This yields, I think, better error messages when failing to solve `Coercible` constraints on arguments with different kinds. Compare the error you get when compiling tests/purs/failing/CoercibleKindMismatch.purs with this check:

```
Error found:
in module Main
at ../purs/failing/CoercibleKindMismatch.purs:17:39 - 17:45 (line 17, column 39 - line 17, column 45)

  Could not match kind
        
    Type
        
  with kind
                
    Type -> Type
                

  Prim.Coerce.Coercible Unary 
                        Binary
                              
while checking that type forall (a :: Type) (b :: Type). Coercible a b => a -> b
  is at least as general as type Proxy @(Type -> Type) Unary -> Proxy @(Type -> ...) Binary
  has type Proxy @(Type -> Type) Unary -> Proxy @(Type -> ...) Binary
in value declaration unaryToBinary

```

and without:

```
Error found:
in module Main
at ../purs/failing/CoercibleKindMismatch.purs:18:17 - 18:23 (line 18, column 17 - line 18, column 23)

    Prim.Coerce.Coercible (Unary t0)    
                          (Binary t0 t1)
                                        

while checking that type forall (a :: Type) (b :: Type). Coercible a b => a -> b
  is at least as general as type Proxy @(Type -> Type) Unary -> Proxy @(Type -> ...) Binary
  has type Proxy @(Type -> Type) Unary -> Proxy @(Type -> ...) Binary
in value declaration unaryToBinary

```

kl0tl:
Agreed! This is the biggest one left to merge I think.

Would you then prefer that I update https://github.com/purescript/purescript/pull/3878 to solve `Coercible` constraints on rows instead of having an ad-hoc case for records or that I open another pull request?

hdgarrood:
Up to you, I don’t have a preference either way.

src/Language/PureScript/TypeChecker/Entailment.hs
natefaubion: In that case I think you should use `elaborateKind` rather than `kindOf`. `elaborateKind` is like a shortcut kind inference for these cases where something is already kind-correct, and all you want to know is the kind rather than a rewritten term.

natefaubion: The name is a bit of a misnomer. It should probably be `elaboratedKind` (it returns the kind of an already elaborated term) or `fastKindOf` 😆.

kl0tl: Oh that’s why `unifySubstKind` calls `elaborateKind` instead of `kindOf`! Thank you!

Do you think I should also apply the current substitution to the elaborated kinds? I can’t find a situation where this makes any difference.

kl0tl: There’s an error with polymorphic kinds now.

This minimal example extracted from tests/purs/passing/Coercible.purs fails to compile:

```purs
module Example where

import Safe.Coerce (coerce)

newtype Phantom1 a b = Phantom1 a

newtype Roles1 a b c = Roles1 (Phantom1 b c)

example :: forall r s t. Roles1 r s t -> s
example = coerce
```

```
Error found:
in module Example
at Example.purs:10:11 - 10:17 (line 10, column 11 - line 10, column 17)

  Internal compiler error:

    Cannot apply type to type: Phantom1 s1

  internalCompilerError, called at src/Language/PureScript/TypeChecker/Kinds.hs:273:3 in purescript-0.13.8-2iUzbEzzqK93DOWR2YKUzM:Language.PureScript.TypeChecker.Kinds
  cannotApplyTypeToType, called at src/Language/PureScript/TypeChecker/Kinds.hs:508:9 in purescript-0.13.8-2iUzbEzzqK93DOWR2YKUzM:Language.PureScript.TypeChecker.Kinds
  elaborateKind, called at src/Language/PureScript/TypeChecker/Kinds.hs:496:11 in purescript-0.13.8-2iUzbEzzqK93DOWR2YKUzM:Language.PureScript.TypeChecker.Kinds
  elaborateKind, called at src/Language/PureScript/TypeChecker/Entailment.hs:407:21 in purescript-0.13.8-2iUzbEzzqK93DOWR2YKUzM:Language.PureScript.TypeChecker.Entailment

  Prim.Coerce.Coercible (Phantom1 s3 t4)
                        s3              
                                        
while checking that type forall (a :: Type) (b :: Type). Coercible a b => a -> b
  is at least as general as type Roles1 @t0 @t1 r2 s3 t4 -> s3
  has type Roles1 @t0 @t1 r2 s3 t4 -> s3
in value declaration example

where r2 is a rigid type variable
        bound at (line 10, column 11 - line 10, column 17)
      s3 is a rigid type variable
        bound at (line 10, column 11 - line 10, column 17)
      t4 is a rigid type variable
        bound at (line 10, column 11 - line 10, column 17)
      t0 is an unknown type
      t1 is an unknown type

See https://github.com/purescript/documentation/blob/master/errors/InternalCompilerError.md for more information,
```

But adding a kind declaration to `Phantom1` is enough to resolve the error:

```diff
+newtype Phantom1 :: Type -> Type -> Type
 newtype Phantom1 a b = Phantom1 a
```

natefaubion: That suggests to me that something is ill-kinded. For example, if you use elaborateKind with `Proxy Id` where there is no kind-application (`Proxy @(Type -> Type) Id`), then it will probably generate an error like that. When it looks up the kind of `Proxy`, it sees it has a `forall`, and not an arrow.

kl0tl: It seems that the polymorphic kind of row tails isn’t instantiated under newtypes:

```purs
module Example where

import Safe.Coerce (coerce)

newtype N f = N (f {})

example :: forall f. N f -> f {}
example = coerce
```

```
Error found:
in module Example
at Example.purs:8:11 - 8:17 (line 8, column 11 - line 8, column 17)

    Prim.Coerce.Coercible (f0 (Record ()))        
                          (f0 (Record (() @Type)))
                                                  

  is at least as general as type N f0 -> f0 (Record (() @Type))
  has type N f0 -> f0 (Record (() @Type))
in value declaration example

        bound at (line 8, column 11 - line 8, column 17)

```

The previous example compiles fine when comparing the types rewritten by `kindOf` in `solveCoercible`.

Is there any other way to wrap row tails in kind applications when unwrapping newtypes?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -381,80 +390,93 @@ entails SolverOptions{..} constraint context hints =
...
+    coercibleWanteds :: Environment -> SourceType -> SourceType -> m (Maybe [SourceConstraint])
+    coercibleWanteds env a b
+      | (TypeConstructor _ aTyName, _, axs) <- unapplyTypes a
+      , (TypeConstructor _ bTyName, _, bxs) <- unapplyTypes b
+      , Just (aTyKind, _) <- M.lookup aTyName $ types env
+      , Just (bTyKind, _) <- M.lookup bTyName $ types env
+      , (aks, kind) <- unapplyKinds aTyKind
+      , (bks, _) <- unapplyKinds bTyKind
+      , length axs < length aks
+      , length bxs < length bks = do
```
hdgarrood, line 427: Can I run my understanding of what's going on here past you to check it's accurate? So `axs` is the list of type arguments that `a` has, and `aks` is the list of kinds of those arguments (and likewise for `b`). We know that `length axs <= length aks`, because otherwise we would have had a kind unification error by now, so the possibilities are that `length axs` might be less than `length aks` (if `a` isn't fully applied), or otherwise it will be equal to `length aks` (if it is fully applied). And again, same for `b`.

I don't think it should be possible that only one of the two arguments is fully applied, right? Otherwise we would have had a kind unification error on line 400? Does it follow that one of these length checks is redundant?

hdgarrood: That is, I feel like it should follow from `a` and `b` having the same kind that `length aks - length axs = length bks - length bxs`

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -381,80 +390,93 @@ entails SolverOptions{..} constraint context hints =
...
+          -- If both arguments have kind @k1 -> k2@ (e.g. @data D a b = D a@
+          -- in the constraint @Coercible (D a) (D a')@), yield a new wanted
+          -- constraint in terms of the types saturated with the same variables
+          -- (e.g. @Coercible (D a t0) (D a' t0)@ in the exemple).
+          tys <- traverse freshTypeWithKind aks
```
hdgarrood, line 432: Will this work if the type constructors in use in `a` and `b` have different numbers of arguments? I'm wondering if this should instead be something like `replicate (length aks - length axs) freshTypeWithKind`, and then get rid of the `drop` on the following lines. For example, if we have, say
```
MkX :: X -> Type -> Type
MkYY :: Y -> Y -> Type -> Type
SomeX :: X
SomeY :: Y
```
and the two arguments `a` and `b` we're dealing with are `MkX SomeX` and `MkYY SomeY SomeY`, then I think we will have
```
axs = [ SomeX ]
bxs = [ SomeY, SomeY ]
aks = [ X, Type ]
bks = [ Y, Y, Type ]
```
In this case, `tys` will have the same number of elements as `aks`, i.e. 2, and so we'll have
```
drop (length axs) tys = [ t0 ]
drop (length bxs) tys = []
```
so we end up recursing with 
```
a = MkX SomeX t0
b = MkYY SomeY Some Y
```
which then leads to a kind unification error, even though the user might not have written anything ill-kinded? Let me know if this makes any sense.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -381,80 +390,93 @@ entails SolverOptions{..} constraint context hints =
...
+    solveCoercible env kinds [a, b] = runMaybeT $ do
       let tySynMap = typeSynonyms env
           kindMap = types env
-          replaceTySyns = either (const Nothing) Just . replaceAllTypeSynonymsM tySynMap kindMap
-      a' <- replaceTySyns a
-      b' <- replaceTySyns b
+          replaceTySyns = replaceAllTypeSynonymsM tySynMap kindMap
+      (a', kind) <- lift $ replaceTySyns a >>= kindOf
+      (b', kind') <- lift $ replaceTySyns b >>= kindOf
+      lift $ unifyKinds kind kind'
```
hdgarrood, line 400: I'm sure this is a gap in my own understanding, but it feels a little bit strange to me that we aren't capturing any information as a result of unifying kinds here. Is it possible that we learn something about either `kind` or `kind'` that we didn't know until we reached this line? If so, should that information be carried forward?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -381,80 +390,93 @@ entails SolverOptions{..} constraint context hints =
...
-            -- terms of that type with the type arguments substituted in (e.g.
-            -- @Coercible (T[X/a]) b = Coercible X b@ in the example).
-            let wrappedTySub = replaceAllTypeVars (zip tvs xs) wrappedTy
-            pure [Constraint nullSourceAnn C.Coercible [] [wrappedTySub, b] Nothing]
-      _ ->
+    coercibleWanteds :: Environment -> SourceType -> SourceType -> m (Maybe [SourceConstraint])
+    coercibleWanteds env a b
+      | (TypeConstructor _ aTyName, _, axs) <- unapplyTypes a
+      , (TypeConstructor _ bTyName, _, bxs) <- unapplyTypes b
+      , Just (aTyKind, _) <- M.lookup aTyName $ types env
```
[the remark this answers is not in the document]
kl0tl: You‘re right, both type constructors should have been inserted into the environment at this point. I‘ve thrown `"coercibleWanteds: type lookup failed"` on failed lookups here and in the next guard.

src/Language/PureScript/TypeChecker/Entailment.hs
kl0tl: You‘re absolutely right and the `length bxs < length bks` check is indeed redundant now that kinds are checked in `solveCoercible`. I‘ve removed it and also another redundant `not (null bxs)` check in the next guard (if both type constructors have the same name and the first one has no arguments then it would be ill-kinded for the second one to have arguments).

src/Language/PureScript/TypeChecker/Entailment.hs
kl0tl: Very well spotted! Thank you for taking the time to write such a detailed example 🙇

src/Language/PureScript/TypeChecker/Entailment.hs
kl0tl: `unifyKinds` returns a `(MonadError MultipleErrors m, MonadState CheckState m) => m ()` so I’m not sure how to extract any information from a successful unification 🤔 

If unifying kinds extends the current substitution perhaps we should apply it to the kinds we inferred and then refer to `[kind, kind']` rather than `kinds` in the returned type class dictionary? Or perhaps we should rather apply it to the types rewritten by `kindOf` earlier? I’m completely out of my depth here 😅

src/Language/PureScript/TypeChecker/Entailment.hs
hdgarrood: Sure! I was mainly doing it for my own benefit initially, to try to check I had understood it properly (and because this was too much to keep in my head at once) haha

src/Language/PureScript/TypeChecker/Entailment.hs
hdgarrood: Yeah, I’m a bit out of my depth too. @natefaubion I’d be interested to hear your thoughts on this if you have a moment.

src/Language/PureScript/TypeChecker/Entailment.hs
hdgarrood: Would you mind adding a test which would catch this too, when you get a chance?

src/Language/PureScript/TypeChecker/Entailment.hs
natefaubion: It sounds like this is an error somewhere in the kind-checker then? The types in the environment should always be fully elaborated with their kind-applications. It should never be necessary to do full kind inference during constraint solving.

src/Language/PureScript/TypeChecker/Entailment.hs
natefaubion: `unifyKinds` does extend the substitution and at some point it must be applied, yes. I believe it's always applied at the top of the solver loop.

hdgarrood: Do you think that suggests that it might be worth exporting `apply` from TypeChecker.Kinds to use here?

hdgarrood: Ah, I suppose we should resolve the other discussion (https://github.com/purescript/purescript/pull/3893#discussion_r479778954) first.

natefaubion: `applySubstitution` as is used everywhere else is fine. I think there's only one in the kind-checker for convenience or module dependencies or something.

hdgarrood:
I managed to come up with this test, which fails on 8552334f but passes on the latest version of this branch:

```purescript
newtype Ap f a = Ap (f a)
newtype Id1 a = Id1 a
newtype Rec4 f = Rec4 (f {})
data K
foreign import data SomeK :: K
newtype Rec5 (f :: K) (g :: Type -> Type) = Rec5 (g {})

-- This coercion checks a slightly more complex path to solving Coercible which
-- makes use of Coercible's polykinded-ness. Specifically, we're expecting this
-- particular series of subgoals (amongst others which I haven't bothered
-- noting down):
--
-- Coercible (Ap Rec4 Id1) (Ap (Rec5 SomeK) Id1)
--  -> Coercible Rec4 (Rec5 SomeK)
--  -> Coercible (Rec4 t0) (Rec5 SomeK t0)
--  -> Coercible (t0 {}) (t0 {})
--  -> ok
--
-- The step from `Coercible Rec4 (Rec5 SomeK)` to `Coercible (Rec4 t0) (Rec5
-- SomeK t0)` in particular is the one we are interested in.
apRec4ToApRec5 :: Ap Rec4 Id1 -> Ap (Rec5 SomeK) Id1
apRec4ToApRec5 = coerce
```
Since the questions around kind unification seem a little more complex, I am tempted to leave those questions for later in order to get this merged, as what we have here now fixes an important bug in Coercible. So how about this: I push a commit to this branch adding the above test, we merge this, and write up a new issue to discuss whether or not we should be unifying kinds at all during Coercible solving (and if so, how exactly we should be going about that)?

natefaubion:
> Since the questions around kind unification seem a little more complex

Do you specifically mean kind inference (`kindOf`)? Yes, it's true that it's harmless on an already elaborated type, just potentially inefficient. If it's _required_ though, something is probably a dodgy somewhere. Or at least, an unelaborated type is being constructed somewhere, which really shouldn't be happening at this stage in the compiler. Constraint solving is happening after type checking, which is what elaborates everything.

hdgarrood:
Sorry, yes, I mean kind inference. Would you prefer that we investigate and address this first, then?

hdgarrood:
(where "this" = "the issue that an unelaborated type is being constructed somwhere")

kl0tl:
I wanted to add the following test

```purs
module Main where

import Safe.Coerce (coerce)

data Unary a
data Binary a b

data Proxy a = Proxy
type role Proxy representational

data Unit

unaryToBinary :: Proxy Unary -> Proxy (Binary Unit)
unaryToBinary = coerce
```

but it fails with

```
Error found:
in module Example
at Example.purs:14:17 - 14:23 (line 14, column 17 - line 14, column 23)

  Type variable k is undefined.

while checking that type t2
  has kind t3
while inferring the kind of Unary @t3 t2
                                                  
  Prim.Coerce.Coercible (Unary @t0 t2)            
                        (Binary @Type @t1 Unit t2)
                                                  
  has type Proxy @(t0 -> Type) (Unary @t0) -> Proxy @(t1 -> Type) (... @t1 Unit)
in value declaration unaryToBinary

where t0 is an unknown type
      t1 is an unknown type
      t3 is an unknown type
      t2 is an unknown type

```

I don’t have figured out why yet but adding kind annotations to `Unary` and `Binary` parameters fixes the issue.

natefaubion:
> Sorry, yes, I mean kind inference. Would you prefer that we investigate and address this first, then?

I would like to at least know where they are coming from. For example, are the subgoals we are constructing for Coercible lacking kind applications? If that's the only source, it's probably not worrisome, it's just inefficient. I have not looked at the code extensively for how we are constructing our sub goals, but that's something to investigate.

kl0tl:
The kind inference issue can be observed with

```purs
module Example where

import Safe.Coerce (coerce)

newtype N f = N (f {})

example :: forall f. N f -> f {}
example = coerce
```

When unwrapping `N` we extract the wrapped type from the environment, which lacks a kind application on the empty row:

```purs
((f :: Type -> Type) (Record ()))
```

natefaubion:
If you insert a call to trace `debugDataConstructors`, I get:
```
coercibleWanteds
Example.N :: forall (f :: Type -> Type). f (Record (() @Type)) -> N f

Qualified (Just (ModuleName "Example")) (ProperName {runProperName = "N"})
["f"]
f (Record ())
```
So I think whatever source `lookupNewtypeConstructor` is using is just not an accurate source.

natefaubion:
The kind checker doesn't elaborate the type in the data declaration spine, it yields elaborated types of constructors.

natefaubion:
Here is where it does not rewrite the `ctor`
https://github.com/purescript/purescript/blob/2bd7ca5f45bd1b6daece10fe8c0f1e932d21ea75/src/Language/PureScript/TypeChecker/Kinds.hs#L605

Here is where it checks the fields and returns a type for the constructor:
https://github.com/purescript/purescript/blob/2bd7ca5f45bd1b6daece10fe8c0f1e932d21ea75/src/Language/PureScript/TypeChecker/Kinds.hs#L613

You could have it rewrite the constructor fields as well as returning the type here.

kl0tl:
Thank you for the pointers Nate! I think I managed to do as you suggested and unifying elaborated kinds happens to also fix https://github.com/purescript/purescript/pull/3893#issuecomment-687640334 🎉  There’s two new issues with this though, tests/purs/passing/LetPattern.purs now fails with:

```
    var v1 = Y.create(25252)("hello, world")(false);
                                            ^

TypeError: Y.create(...)(...) is not a function
```

and tests/purs/passing/Coercible.purs fails with:

```
Error found:
in module Main
at tests/purs/passing/Coercible.purs:71:18 - 71:24 (line 71, column 18 - line 71, column 24)

    Prim.Coerce.Coercible (forall (k1 :: Type) (k2 :: Type). Phantom1 @k2 s0 t4)
                          s0                                                    
                                                                                

  is at least as general as type s0 -> Roles1 @t1 @t2 r3 s0 t4
  has type s0 -> Roles1 @t1 @t2 r3 s0 t4
in value declaration roles1ToSecond

where r3 is a rigid type variable
        bound at (line 71, column 18 - line 71, column 24)
      s0 is a rigid type variable
        bound at (line 71, column 18 - line 71, column 24)
      t4 is a rigid type variable
        bound at (line 71, column 18 - line 71, column 24)
      t1 is an unknown type
      t2 is an unknown type

```

src/Language/PureScript/TypeChecker/Entailment.hs
kl0tl: I added tests/purs/failing/CoercibleHigherKindedData.purs and a case inspired by https://github.com/purescript/purescript/pull/3893#issuecomment-687634696 to tests/purs/passing/Coercible.purs.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -602,15 +603,18 @@ inferDataDeclaration moduleName (ann, tyName, tyArgs, ctors) = do
...
   :: forall m. (MonadError MultipleErrors m, MonadState CheckState m)
   => SourceType
   -> DataConstructorDeclaration
-  -> m SourceType
-inferDataConstructor tyCtor =
-  flip checkKind E.kindType . foldr ((E.-:>) . snd) tyCtor . dataCtorFields
+  -> m (DataConstructorDeclaration, SourceType)
+inferDataConstructor tyCtor DataConstructorDeclaration{..} = do
+  tyCtor' <- checkKind (foldr ((E.-:>) . snd) tyCtor dataCtorFields) E.kindType
+  let (_, _, tys) = unapplyTypes tyCtor'
```
natefaubion, line 615: I don't think `unapplyTypes` is correct here. `unapplyTypes` deconstructs  applications, but `tyCtor'` is the signature for the data constructor. That is, given `data Foo a = Foo Int a String`, then `tyCtor'` will be `Int -> a -> String -> Foo a`. `unapplyTypes` called on this will yield an application to `->` with two arguments. Maybe we should do a traversal over `dataCtorFields`, checking each field against `E.kindType`, zip this with  dataCtorFields, and then assemble the constructor signature after.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -951,7 +955,8 @@ kindsOfAll moduleName syns dats clss = withFreshSubstitution $ do
           (args', supers', decls', generalizeUnknownsWithVars unkBinders clsKind)
     datResultsWithKinds <- for datResultsWithUnks $ \(((datName, datKind), ctors), _) -> do
       let tyUnks = snd . fromJust $ lookup (mkQualified datName moduleName) tySubs
-          ctors' = fmap (fmap (generalizeUnknowns tyUnks . replaceTypeCtors)) ctors
+          replaceTypeCtorsAndGeneralizeUnknowns = generalizeUnknowns tyUnks . replaceTypeCtors
+          ctors' = fmap (mapDataCtorFields (fmap (fmap replaceTypeCtorsAndGeneralizeUnknowns)) *** replaceTypeCtorsAndGeneralizeUnknowns) ctors
```
natefaubion, line 959: It's not clear to me that you want to generalize unknowns in the data constructor spine, as this will introduce `forall`s inside the data constructor spine for any field with an unknown. These unknowns should be quantified as part of the data declaration, not in each field. That is
```
data Foo f a = Foo (f a)
```
This should be quantified as
```
data Foo :: forall k. (k -> Type) -> k -> Type
data Foo f a = Foo (f @k a)
```
Not as
```
data Foo f a = Foo (forall k. f @k a)
```

src/Language/PureScript/TypeChecker/Kinds.hs
kl0tl: Of course 🤦 I‘ve checked each field independently. I attempted to build the constructor type from the checked fields types but this led to issues with rank-n kinds so I built it from the unchecked field types and then checked the whole.

Is there a way to reuse some of the work done checking the fields when building the constructor type?

src/Language/PureScript/TypeChecker/Kinds.hs
kl0tl: Oh that’s why I had those unwanted foralls! I’ve replaced unknowns without generalizing them.

src/Language/PureScript/TypeChecker/Kinds.hs
natefaubion: I'm not sure why you would have rank-n issues. Maybe you constructed the associativity of `->` incorrectly? Maybe something like:
```
inferDataConstructor tyCtor DataConstructorDeclaration{..} = do
  dataCtorFields' <- traverse (traverse (flip checkKind E.kindType)) dataCtorFields
  dataCtor <- flip (foldr ((E.-:>) . snd)) dataCtorFields' =<< checkKind tyCtor E.kindType
  pure (DataConstructorDeclaration { dataCtorFields = dataCtorFields, .. }, dataCtor)
```

kl0tl: Oh I just forgot to check the constructor return type :sweat_smile: I applied your suggestion!

natefaubion:
@hdgarrood @kl0tl I'm happy with the changes to the kind checker and the use `elaborateKind`. As long as everything passes and y'all are happy with the Coercible changes, I think this is OK to merge.

kl0tl:
The last unresolved issue should be about whether to apply the current substitution to `Coercible` constraints parameters and/or their kinds before computing subgoals (https://github.com/purescript/purescript/pull/3893#discussion_r475277806).

As you said, the current substitution is already applied at the top of the solver recursive worker so _I think_ the only risk is to fail to equate the parameters of identical constructors at role nominal [here](https://github.com/purescript/purescript/blob/bcdf8c8e424f1eeb7a8cea4de3ac58f139036531/src/Language/PureScript/TypeChecker/Entailment.hs#L451)?

We’re unifying elaborated kinds now though, are unknowns still expected in them?

natefaubion:
You need to apply a substitution if you want to observe anything about the type after unifying. You do not want to observe a unification variable that is otherwise solved.

natefaubion:
> As you said, the current substitution is already applied at the top of the solver recursive worker so I think the only risk is to fail to equate the parameters of identical constructors at role nominal here?

> We’re unifying elaborated kinds now though, are unknowns still expected in them?

I think you said you kept unification because it yielded a better error? Is unification necessary for soundness? If not, then maybe it isn't necessary to apply a substitution.

kl0tl:
The unified kinds are thrown away afterwards but we rely on the constraint arguments to have the same kind when computing subgoals to avoid redundant checks (for instance we only check if the first argument is unsaturated or has a non empty arguments list).

I’d like to find a case where applying the substitution makes a difference. If my understanding is correct this should only happen when one of the constraint arguments has an unknown kind application that would be solved by the unification of both arguments kinds? I’m struggling to observe this though 😅

natefaubion:
> If my understanding is correct this should only happen when one of the constraint arguments has an unknown kind application that would be solved by the unification of both arguments kinds? I’m struggling to observe this though

I'm not sure how a kind like this (though it sounds like it's in paradox territory 😆) would be unknown at this point since we are past type-checking.

hdgarrood (approved):

hdgarrood:
Since this issue is difficult (potentially impossible) to observe, shall we merge this now and come back to it later if it comes back up?

kl0tl:
Sounds good to me! I‘ve squashed the last commit and the one it fixes.

hdgarrood:
Amazing - thanks for your work on this!

PR #3894, opened 2020-06-02
Updated purescript version in package.yaml

coot:


hdgarrood:
Thanks! This is updated in the v0.13.x branch but I forgot to do it in master.

Issue #3895, opened 2020-06-03
Role inference for foreign imported types without roles declaration is wrong


kl0tl:
## Description

In the absence of role declarations, roles of foreign imported types are computed from their kind by the following function:

https://github.com/purescript/purescript/blob/d7e8ef857943e2186857c682076f0caf789e6abe/src/Language/PureScript/TypeChecker/Roles.hs#L141-L149

It is supposed to return of list of `n` nominal roles, where `n` is the kind arity of the type, but instead returns `[Nominal]` for `Type` and `forall k. k -> … -> Type`, `[Nominal, Nominal]` for `Type -> Type` and `Type -> Type -> … -> Type`, `[Nominal, Nominal, Nominal]` for `(Type -> Type) -> Type`…

## To Reproduce

```purs
module Example where

import Safe.Coerce (coerce)

foreign import data Foreign3 :: Type -> Type -> Type -> Type

-- We infer `nominal nominal` instead of `nominal nominal nominal`,
-- this is equivalent to giving a phantom role to `Foreign3` last parameter:

example :: ∀ a b c d. Foreign3 a b c -> Foreign3 a b d
example = coerce
```

```purs
module Example where

import Safe.Coerce (coerce)

foreign import data Foreign2 :: ∀ k. k -> k -> Type

-- We infer `nominal` instead of `nominal nominal`,
-- this is equivalent to giving a phantom role to `Foreign2` last parameter:

example :: ∀ k (a :: k) (b :: k) (c :: k). Foreign2 a b -> Foreign2 a c
example = coerce
```

## Expected behavior

We should recurse on the second argument of `->`, recurse under foralls and return the accumulator untouched in the base case.

## PureScript version

master (d7e8ef857943e2186857c682076f0caf789e6abe)

hdgarrood:
This is fixed by #3893 right?

kl0tl:
Sadly no, albeit we could define the function in terms of `Language.PureScript.Environment.kindArity` though. I opened https://github.com/purescript/purescript/pull/3896 to add a few tests.

PR #3896, opened 2020-06-03
Recurse on the right of arrows and under foralls when inferring nominal roles from kinds

kl0tl:
Fix https://github.com/purescript/purescript/issues/3895.


hdgarrood:
This looks good but I do quite like the formulation in terms of `kindArity` as in #3893. Could you put `kindArity` in this commit? Then `rolesFromForeignTypeKind` would just be `\k -> replicate (kindArity k) Nominal` right?

PR #3897, opened 2020-06-05
Support local type synonyms natively

rhendric:
I'm attempting to implement local type synonyms (#3708) properly in the type checker, but it's not going as smoothly as I'd hoped and I'm not sure what I'm doing wrong. Maybe someone can give this a once-over and offer some pointers? @natefaubion, I've been trying to follow your advice; it seems there are a few more things to do than expected, or I took an early wrong turn somewhere.

My current blocker is `tests/purs/warning/LocalTypeSynonyms.purs`, which crashes with this message:

        at tests/purs/warning/LocalTypeSynonyms.purs:9:21 - 9:24 (line 9, column 21 - line 9, column 24)
        
          Internal compiler error:
        
            Unsolved unification variable ?4 is not bound
        
          internalCompilerError, called at src/Language/PureScript/TypeChecker/Kinds.hs:139:16 in purescript-0.13.8-3tLfHv2rhWk5dQvnEaCUgA:Language.PureScript.TypeChecker.Kinds
          lookupUnsolved, called at src/Language/PureScript/TypeChecker/Kinds.hs:250:17 in purescript-0.13.8-3tLfHv2rhWk5dQvnEaCUgA:Language.PureScript.TypeChecker.Kinds
          inferAppKind, called at src/Language/PureScript/TypeChecker/Kinds.hs:209:7 in purescript-0.13.8-3tLfHv2rhWk5dQvnEaCUgA:Language.PureScript.TypeChecker.Kinds
          inferKind, called at src/Language/PureScript/TypeChecker/Kinds.hs:650:24 in purescript-0.13.8-3tLfHv2rhWk5dQvnEaCUgA:Language.PureScript.TypeChecker.Kinds
        
        while inferring the kind of f0 a
        in type synonym LocalApp
        while checking that expression 0      
                                         where
          has type Int
        in value declaration generalized
        
        See https://github.com/purescript/documentation/blob/master/errors/InternalCompilerError.md for more information,

What's going wrong is fairly clear: there's a unification variable being introduced for the kind of `f`, and when `L.P.TypeChecker.Kinds.inferTypeSynonym` tries to determine the kind of the body of the synonym, it expects that variable to have been connected to something already. But I haven't been able to figure out what I ought to be doing to make that connection, or if I have to abandon or deeply change the whole `Kinds.kindsOfAll` machinery for the case of local type synonyms. I can already tell something about that is going to need to change because it generalizes type arguments, and we don't want that for local type synonyms, but one thing at a time.

Still to-do:
- [ ] don't generalize local type synonyms
- [ ] support kind signatures


natefaubion:
For things like this, I recommend using the `debug` functions in `TypeChecker.Monad`, specifically for this `debugSubstitution`, which will print out all the unification variables with either their kinds (if unsolved) or solutions (if solved). In this particular case, the substitution environment before kind checking that synonym is:
```
?2 = t4
?3 = t4
?4 :: Type
```
So the particular unification variable in the error definitely exists. The reason for the error is that `kindOfTypeSynonym` calls `kindsOfAll`, which is wrapped in `withFreshSubstitution`.  `withFreshSubstitution` sets up an empty substitution environment. You want this for top-level declarations, but clearly not in this case. Your intuition is correct, in that we need to extract some of the logic into something appropriate for local synonyms.

The apparent complexity of `kindsOfAll` comes from dealing with recursive type-level binding groups, which is not possible in a local context like this since we disallow recursive type synonyms altogether. Recursive binding groups need to be generalized correctly so that all the kind-polymorphism can be threaded around between them. We don't want to do any of that for local type synonyms. For the most part, you can just call call `inferTypeSynonym` directly. This will infer the kind of the body, and elaborate the rhs with kind applications. There is some setup though that's done, and a few post-inference checks to account for.

`inferTypeSynonym` assumes that there is a kind for the type synonym in the environment. For type synonyms _without_ explicit kind signatures, this should be a fresh unification variable. `kindsOfAll` sets this up using `existingSignatureOrFreshKind`. `inferTypeSynonym` will unify against whatever is in the environment.

The `checkEscapedSkolems` check is necessary, and it should already account for types in scope, so I think no changes are necessary here. This just ensures that there are no skolems left over that don't also have an appropriate entry in the type environment. I'm not positive that the main typechecker puts skolems in the type environment though, so that may need some validation.

I don't think that `checkTypeQuantification` is likely necessary. The purpose of this check is to make sure there are no unknowns leftover _after_ generalization. This can happen when you have a closed kind application. A term-level analog is `length []`, where `length` must be instantiated with the type of `[]`, which can be anything. At the kind-level this would be left as an unbound unification variable. I don't believe this would be observable for local type synonyms.

I'm going to guess that `checkVisibleTypeQuantification` is unnecessary. We don't currently support visible quantification. In top-level contexts however it's possible to require this through generalization. Since we are not generalizing local type synonyms, we can probably elide this check.

rhendric:
@natefaubion Thank you so much! That was immensely useful.

There are likely a few more edge cases that I haven't yet tested and fixed, but this is now mostly functional.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -496,6 +499,17 @@ inferLetBinding seen (BindingGroupDeclaration ds : rest) ret j = do
   bindNames dict $ do
     makeBindingGroupVisible
     inferLetBinding (seen ++ [BindingGroupDeclaration ds']) rest ret j
+inferLetBinding seen (TypeSynonymDeclaration sa@(ss, _) name args ty : rest) ret j = do
+  moduleName <- unsafeCheckCurrentModule
+  (kind', ty') <- warnAndRethrow (addHint (ErrorInTypeSynonym name) . addHint (positionedError ss)) $ do
+    checkDuplicateTypeArguments $ map fst args
+    traverse replaceAllTypeSynonyms . swap =<< kindOfLocalTypeSynonym moduleName (sa, name, args, ty)
```
rhendric, line 506: Does there need to be an `introduceSkolemScope` in here? I tried it with and without and it didn't seem to make a difference on any tests I could come up with, but it seems to be paired with all the other `replaceAllTypeSynonyms`s in this file.

natefaubion:
We need to support local synonyms in the shadowing lint for both:
* type synonyms shadowing other type level names
* type variables in type synonyms shadowing other type variables.

rhendric:
For the first case: `ShadowedName`, or a new error constructor?

natefaubion:
If the types work out, I say reuse it, otherwise a new one.

rhendric:
Warnings have been added! Most of them were not put in the `Linter` module proper, since quantifier type variables aren't associated with value declarations until the type checking phase. In theory I imagine this could be changed, but the existing type variable shadowing warning lives in the type checker so that's where I put the new warnings that use the same information. I hope that's reasonable.

Issue #3884, opened 2020-05-20
Proposal: List all files involved in DuplicateModule error


milesfrain:
It would be more helpful to list all duplicate files detected, rather than just one of them.

For example:
```
spago init
mkdir src/sneaky
echo "module Main where" > src/sneaky/Dup.purs
spago build
```
```
Error found:
at src/Main.purs:1:1 - 6:28 (line 1, column 1 - line 6, column 28)

  Module Main has been defined multiple times

See https://github.com/purescript/documentation/blob/master/errors/DuplicateModule.md for more information,
or to contribute content related to this error.

[error] Failed to build.
```
I'm actually more interested in knowing about `src/sneaky/Dup.purs` than `src/Main.purs`.

Related to #1828

hdgarrood:
It looks like this previously worked. The fact that it now doesn’t may be related to #3273. It seems like there is a mechanism for including more than one source span in an error, except that in the case of this error we’re only actually printing one of the spans, whereas we should really be printing all of them?

garyb:
Yeah, it might be that we never print the multiple spans when they're available, and only take the primary/first one. 

I was also a little disappointed that none of the tools make use of the multi-spans that I know of after they got added, but then it's not like I did anything to resolve that either. :wink:

PR #3899, opened 2020-06-19
Faster warnDuplicateRefs

milesfrain:
See the "[Reducing incremental rebuild time with large file dependency](https://discourse.purescript.org/t/reducing-incremental-rebuild-time-with-large-file-dependency/1432)" thread for context.

Looks like the issue is here:
https://github.com/purescript/purescript/blob/c044d69d4778c3d426e6550e20000362626ffacd/src/Language/PureScript/Sugar/Names/Common.hs#L27

Profiling shows warnDuplicateRefs takes 46% of the build time - the majority of it spent making 96 million [calls to ==](https://github.com/purescript/purescript/blob/9cad73ed8ea7df3011032ddbd2f3de5a0c08629c/src/Language/PureScript/AST/Declarations.hs#L316-L326).

I replaced the slow `\\ nub` pattern with a faster `onlyDuplicates` function. I'm surprised this doesn't exist already in some library.

The `onlyDuplicates` function also seems more correct than `\\ nub` in that it doesn't arbitrarily discard one of every duplicate. For example:
```hs
xs = [1,2,2,3,3,3,4]
xs \\ nub xs -- [2,3,3]
onlyDuplicates xs -- [3,3,3,2,2]
```

For my use case with a large `Tailwind.purs` file, overall compilation times are reduced from **3.1 to 1.2 seconds** and the share of time spent in `warnDuplicateRefs` is reduced from **46% to 0.5%**.

Here are more detailed performance notes:

### Original
```
45% of time in construct
  20% in runDecoder
    9 Million entries
49% in make.\
  46% in warnDuplicateRefs
    40% of time running ==
      96 Million entries
```
``` shell
$ echo " " >> src/Example.purs
$ time spago build            
[info] Installation complete.
Compiling Example
Compiling Main
[info] Build succeeded.
spago build  4.23s user 0.98s system 168% cpu 3.091 total
$ time spago build
[info] Installation complete.
[info] Build succeeded.
spago build  1.66s user 0.42s system 288% cpu 0.720 total
```

### This PR
```
83% of time in construct
  37% in runDecoder
    9 Million entries
6% in make.\
  0.5% in warnDuplicateRefs
```
``` shell
$ echo " " >> src/Example.purs
$ time spago build
[info] Installation complete.
Compiling Example
Compiling Main
[info] Build succeeded.
spago build  2.22s user 1.24s system 279% cpu 1.239 total
$ time spago build
[info] Installation complete.
[info] Build succeeded.
spago build  1.72s user 0.44s system 285% cpu 0.759 total
```

### DIFF
--- a/src/Language/PureScript/Sugar/Names/Common.hs
+++ b/src/Language/PureScript/Sugar/Names/Common.hs
@@ -33,15 +33,15 @@ warnDuplicateRefs pos toError refs = do
   where
 
   -- Removes all unique elements from list
+  -- as well as one of each duplicate.
   -- Example:
+  --  removeUnique [1,2,2,3,3,3,4] == [2,3,3]
+  -- Note that it may be more correct to keep ALL duplicates,
+  -- but that requires additional changes in how warnings are printed.
+  -- Example of keeping all duplicates (not what this code currently does):
   --  removeUnique [1,2,2,3,3,3,4] == [2,2,3,3,3]
   removeUnique :: Eq a => Ord a => [a] -> [a]
-  removeUnique = concat . filter twoOrMore . group . sort
-    where
-      twoOrMore :: [a] -> Bool
-      twoOrMore [] = False
-      twoOrMore [_] = False
-      twoOrMore _ = True
+  removeUnique = concatMap (drop 1) . group . sort
 
   -- Deletes the constructor information from TypeRefs so that only the
   -- referenced type is used in the duplicate check - constructors are handled
--- a/src/Language/PureScript/Sugar/Names/Common.hs
+++ b/src/Language/PureScript/Sugar/Names/Common.hs
@@ -6,7 +6,7 @@ import Protolude (ordNub)
 import Control.Monad.Writer (MonadWriter(..))
 
 import Data.Foldable (for_)
-import Data.List (nub, (\\))
+import Data.List (group, sort, (\\))
 import Data.Maybe (mapMaybe)
 
 import Language.PureScript.AST
@@ -24,13 +24,24 @@ warnDuplicateRefs
   -> m ()
 warnDuplicateRefs pos toError refs = do
   let withoutCtors = deleteCtors `map` refs
-      dupeRefs = mapMaybe (refToName pos) $ withoutCtors \\ nub withoutCtors
+      dupeRefs = mapMaybe (refToName pos) $ removeUnique withoutCtors
       dupeCtors = concat $ mapMaybe (extractCtors pos) refs
 
   for_ (dupeRefs ++ dupeCtors) $ \(pos', name) ->
     warnWithPosition pos' . tell . errorMessage $ toError name
 
   where
+
+  -- Removes all unique elements from list
+  -- Example:
+  --  removeUnique [1,2,2,3,3,3,4] == [2,2,3,3,3]
+  removeUnique :: Eq a => Ord a => [a] -> [a]
+  removeUnique = concat . filter twoOrMore . group . sort
+    where
+      twoOrMore :: [a] -> Bool
+      twoOrMore [] = False
+      twoOrMore [_] = False
+      twoOrMore _ = True
 
   -- Deletes the constructor information from TypeRefs so that only the
   -- referenced type is used in the duplicate check - constructors are handled
--- a/tests/purs/warning/DuplicateExportRef.purs
+++ b/tests/purs/warning/DuplicateExportRef.purs
@@ -5,6 +5,12 @@
 -- @shouldWarnWith DuplicateExportRef
 -- @shouldWarnWith DuplicateExportRef
 -- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
 module Main
   ( X(X, X), X
   , fn, fn
--- a/tests/purs/warning/DuplicateExportRef.purs
+++ b/tests/purs/warning/DuplicateExportRef.purs
@@ -5,12 +5,6 @@
 -- @shouldWarnWith DuplicateExportRef
 -- @shouldWarnWith DuplicateExportRef
 -- @shouldWarnWith DuplicateExportRef
--- @shouldWarnWith DuplicateExportRef
--- @shouldWarnWith DuplicateExportRef
--- @shouldWarnWith DuplicateExportRef
--- @shouldWarnWith DuplicateExportRef
--- @shouldWarnWith DuplicateExportRef
--- @shouldWarnWith DuplicateExportRef
 module Main
   ( X(X, X), X
   , fn, fn
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -77,6 +77,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@mgmeier](https://github.com/mgmeier) | Michael Gilliland | [MIT license](http://opensource.org/licenses/MIT) |
 | [@michaelficarra](https://github.com/michaelficarra) | Michael Ficarra | [MIT license](http://opensource.org/licenses/MIT) |
 | [@MichaelXavier](https://github.com/MichaelXavier) | Michael Xavier | MIT license |
+| [@milesfrain](https://github.com/milesfrain) | Miles Frain | [MIT license](http://opensource.org/licenses/MIT) |
 | [@mjgpy3](https://github.com/mjgpy3) | Michael Gilliland | [MIT license](http://opensource.org/licenses/MIT) |
 | [@mpietrzak](https://github.com/mpietrzak) | Maciej Pietrzak | [MIT license](http://opensource.org/licenses/MIT) |
 | [@mrhania](https://github.com/mrhania) | Łukasz Hanuszczak | [MIT license](http://opensource.org/licenses/MIT) |
--- a/lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
+++ b/lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
@@ -201,7 +201,6 @@ instance Ord DeclarationRef where
   TypeClassRef _ name `compare` TypeClassRef _ name' = compare name name'
   TypeInstanceRef _ name `compare` TypeInstanceRef _ name' = compare name name'
   ModuleRef _ name `compare` ModuleRef _ name' = compare name name'
-  KindRef _ name `compare` KindRef _ name' = compare name name'
   ReExportRef _ mn ref `compare` ReExportRef _ mn' ref' = compare mn mn' <> compare ref ref'
   compare ref ref' =
     compare (orderOf ref) (orderOf ref')
@@ -214,8 +213,7 @@ instance Ord DeclarationRef where
         orderOf TypeClassRef{} = 4
         orderOf TypeInstanceRef{} = 5
         orderOf ModuleRef{} = 6
-        orderOf KindRef{} = 7
-        orderOf ReExportRef{} = 8
+        orderOf ReExportRef{} = 7
 
 data ExportSource =
   ExportSource
--- a/lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
+++ b/lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
@@ -193,6 +193,30 @@ instance Eq DeclarationRef where
   (ReExportRef _ mn ref) == (ReExportRef _ mn' ref') = mn == mn' && ref == ref'
   _ == _ = False
 
+instance Ord DeclarationRef where
+  TypeRef _ name dctors `compare` TypeRef _ name' dctors' = compare name name' <> compare dctors dctors'
+  TypeOpRef _ name `compare` TypeOpRef _ name' = compare name name'
+  ValueRef _ name `compare` ValueRef _ name' = compare name name'
+  ValueOpRef _ name `compare` ValueOpRef _ name' = compare name name'
+  TypeClassRef _ name `compare` TypeClassRef _ name' = compare name name'
+  TypeInstanceRef _ name `compare` TypeInstanceRef _ name' = compare name name'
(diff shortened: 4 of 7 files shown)


milesfrain:
There are a few other `\\ nub` patterns that can be improved. Find these with:
```
rg -t hs -i '\\\\.*nub'
```

Since these don't preserve all duplicates, I'm wondering if this would also fix #3884

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -313,17 +313,67 @@ data DeclarationRef
   | ReExportRef SourceSpan ExportSource DeclarationRef
   deriving (Show, Generic, NFData, Serialise)
 
+-- |
+-- A version of DeclarationRef for easily derived comparisons.
+-- This version just has SourceSpan removed from all types.
+--
+data ComparableDeclarationRef
+  -- |
```
milesfrain, line 321: Should the docstrings for each type be repeated?

src/Language/PureScript/Sugar/Names/Common.hs
```diff
@@ -24,14 +24,45 @@ warnDuplicateRefs
   -> m ()
 warnDuplicateRefs pos toError refs = do
   let withoutCtors = deleteCtors `map` refs
-      dupeRefs = mapMaybe (refToName pos) $ withoutCtors \\ nub withoutCtors
+      dupeRefs = mapMaybe (refToName pos) $ onlyDuplicates withoutCtors
```
milesfrain, line 27: This is the core change

src/Language/PureScript/Sugar/Names/Common.hs
```diff
@@ -24,14 +24,45 @@ warnDuplicateRefs
...
       dupeCtors = concat $ mapMaybe (extractCtors pos) refs
 
   for_ (dupeRefs ++ dupeCtors) $ \(pos', name) ->
     warnWithPosition pos' . tell . errorMessage $ toError name
 
   where
 
+  -- Finds all duplicate refs
+  --onlyDuplicates :: forall a. Eq a => Ord a => [a] -> [a]
```
milesfrain, line 36: I'd like to include type annotations for the inner functions, but encountering syntax errors with `forall a .`.

src/Language/PureScript/Sugar/Names/Common.hs
```diff
@@ -24,14 +24,45 @@ warnDuplicateRefs
...
+  onlyDuplicates :: Eq a => Ord a => [a] -> [a]
```
milesfrain, line 37: Wondering if this function should be moved to a more common location.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -313,17 +313,67 @@ data DeclarationRef
   | ReExportRef SourceSpan ExportSource DeclarationRef
   deriving (Show, Generic, NFData, Serialise)
 
+-- |
+-- A version of DeclarationRef for easily derived comparisons.
+-- This version just has SourceSpan removed from all types.
+--
+data ComparableDeclarationRef
```
kritzcreek, line 320: Why do we need to create this additional data type? Note that comparing now allocates when it didn't before.

src/Language/PureScript/Sugar/Names/Common.hs
kritzcreek, line 37: If I'm reading this right... could this be `concatMap (drop 1) . group . sort` ?

milesfrain: Nice, that's certainly a lot more elegant! Your version matches the original `\\ nub` behavior, which discards one of the duplicated elements. I think it's better to preserve all duplicates. Here's a version that does that:
```hs
onlyDuplicates :: Eq a => Ord a => [a] -> [a]
onlyDuplicates = concat . filter twoOrMore . group . sort
  where
    twoOrMore :: [a] -> Bool
    twoOrMore [] = False
    twoOrMore [x] = False
    twoOrMore (x : _) = True
```
Figure that a custom `twoOrMore` function is faster than `(\x -> length x > 1)` with `length` being `O(n)`.

src/Language/PureScript/AST/Declarations.hs
milesfrain: I was hoping there would be some compiler magic that spares these allocations and treat this new "Comparable" data type more like a lens(?) into the original type.
The low-level behavior is still mysterious to me.
It's possible to write an `Ord` instance for `DeclarationRef` but it will be kinda verbose.

hdgarrood: We already do this the verbose way with `Type` in `Language.PureScript.Types`. Following the same pattern here would be preferable to me too.

src/Language/PureScript/Sugar/Names/Common.hs
kritzcreek: > Figure that a custom twoOrMore function is faster than (\x -> length x > 1) with length being O(n).

There's no big `n`'s unless you're actually duplicating the same name _a lot_ so I wouldn't bother. You can check your benchmark but I doubt it's going to be noticeable.

src/Language/PureScript/AST/Declarations.hs
milesfrain: Will do.
Is there a reason why the `compareType` function is used? Couldn't these just all be renamed to `compare` and placed directly after the instance?
https://github.com/purescript/purescript/blob/9cad73ed8ea7df3011032ddbd2f3de5a0c08629c/src/Language/PureScript/Types.hs#L557-L558
https://github.com/purescript/purescript/blob/9cad73ed8ea7df3011032ddbd2f3de5a0c08629c/src/Language/PureScript/Types.hs#L578-L580

hdgarrood: Having a separate `compareType` function allows us to specify different type variables for the annotations for each argument, which means that we can encode the requirement that annotations shouldn't be considered in the types; this way the compiler can verify for us that we aren't accidentally comparing annotations. It also gives us a little more flexibility if we ever want to compare two `Type` values which do actually have different annotations.

hdgarrood: I guess that's less relevant in this instance since `DeclarationRef` doesn't take type parameters.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -325,6 +325,33 @@ instance Eq DeclarationRef where
   (ReExportRef _ mn ref) == (ReExportRef _ mn' ref') = mn == mn' && ref == ref'
   _ == _ = False
 
+instance Ord DeclarationRef where
+  TypeRef _ name dctors `compare` TypeRef _ name' dctors' = compare name name' <> compare dctors dctors'
```
milesfrain, line 329: Does this "short-circuit" if the `name` comparison is non-EQ and skip the `dctors` comparison?

Would it be more efficient to write it like this:
```hs
TypeRef _ name dctors `compare` TypeRef _ name' dctors'
  | nameCmp == EQ = compare dctors dctors'
  | otherwise = nameCmp
  where nameCmp = compare name name'
```
Not sure if the branching and readability costs outweigh the redundant `dctors` comparison cost. Possibly another job for profiling.

milesfrain:
Is `compDecRef` unnecessary now, or does it require this more relaxed (and possibly faster) ordering scheme?

Also, it seems like the code for the `Ord` instances of `DeclarationRef` and `(Type a)` could be condensed by following `compDecRef`'s `orderOf` strategy.

https://github.com/purescript/purescript/blob/9cad73ed8ea7df3011032ddbd2f3de5a0c08629c/src/Language/PureScript/AST/Declarations.hs#L335-L358

src/Language/PureScript/AST/Declarations.hs
hdgarrood: > Does this "short-circuit" if the name comparison is non-EQ and skip the dctors comparison?

Yes it does. This relies on Haskell's laziness, so it won't work the same way in PureScript. The `Semigroup Ordering` instance in Haskell is defined like this:

```haskell
instance Semigroup Ordering where
    LT <> _ = LT
    EQ <> y = y
    GT <> _ = GT
```

and the `_` matches won't be evaluated. You can verify this in the repl:
```
λ: LT <> error "this is fine"
LT
λ: EQ <> error "this is not fine"
*** Exception: this is not fine
  error, called at <interactive>:2:7 in interactive:Ghci2
λ: GT <> error "this is fine"
GT
```

src/Language/PureScript/Sugar/Names/Common.hs
```diff
@@ -24,14 +24,25 @@ warnDuplicateRefs
...
 
   where
 
+  -- Removes all unique elements from list
+  -- Example:
+  --  removeUnique [1,2,2,3,3,3,4] == [2,2,3,3,3]
+  removeUnique :: Eq a => Ord a => [a] -> [a]
+  removeUnique = concat . filter twoOrMore . group . sort
+    where
+      twoOrMore :: [a] -> Bool
```
kritzcreek, line 41: I still think this can just be `(\x -> length x > 1)`, but I'll leave that to your judgement.

hdgarrood:
>  Shouldn't we be commiting to master first and then cherry picking the changes we'd like to release onto the release branches, @hdgarrood ?

Yes please - there may not be another 0.13.x release, but even if we do make an 0.13.9 release we will want any changes in master first.

@milesfrain I like the `orderOf` strategy, and I think you're right that `compDecRef` is now unnecessary, now that the `Ord` instance doesn't look at the source span. I think the only difference between this `Ord` instance and `compDecRef` is that this `Ord` instance looks at the data constructors in a `TypeRef`. I think that difference is not an issue, because `compDecRef` is only used in two places: once in the linter when suggesting explicit imports, and once in the IDE when amending imports, but in either case I think we shouldn't ever end up with more than one `TypeRef` for a given type in the list of refs being sorted, so I think removing `compDecRef` and using this `Ord` instance should be okay.

If you'd like to rewrite the `Ord DeclarationRef` instance with the `orderOf` strategy, I'd appreciate that. If you're going to go down that route, could you please:
- make the `orderOf` function use an exhaustive pattern match?
- add a comment along the lines of the existing compDecRef one, to point out that this particular ordering is used in the linter and the IDE's handling of imports, so it's on purpose that the order of constructors in the ordering is different to the order of constructors in the source file?

If you'd like to do `Type` too, that would also be nice, but could you please leave `Type` out of this PR to keep it small?

hdgarrood:
I edited the above comment a few times, sorry. I'm done now :)

milesfrain:
Rebased to master.

I'm finding that `stack build` now fails on `master` with or without this change. Maybe a clean is required.

I'll make a separate PR to clean-up ordering code that includes:
* Replace `compDecRef` with `instance Ord DeclarationRef`
  * Regarding:
    > add a comment along the lines of the existing compDecRef one, to point out that this particular ordering is used in the linter and the IDE's handling of imports, so it's on purpose that the order of constructors in the ordering is different to the order of constructors in the source file

     Would it be better to reorder the constructors to match this linter ordering?
* Use the `orderOf` strategy for `Type`

src/Language/PureScript/Sugar/Names/Common.hs
milesfrain: Going with the "guaranteed fast" over the "likely fast" strategy, even if it's a few more lines of code.

hdgarrood:
Looks like CI is failing because a compiler warning has changed. Could you run the tests with —accept to see if that fixes it?

milesfrain:
> Looks like CI is failing because a compiler warning has changed. Could you run the tests with —accept to see if that fixes it?

hdgarrood:
`stack test —test-arguments —accept`. Without  the em-dashes, though (I’m on my phone).

tests/purs/warning/DuplicateExportRef.purs
```diff
@@ -5,6 +5,12 @@
...
 -- @shouldWarnWith DuplicateExportRef
 -- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
+-- @shouldWarnWith DuplicateExportRef
 module Main
   ( X(X, X), X
```
milesfrain, line 15: Was wondering whether it makes sense to warn for all duplicates, or just the follow-on duplicates.
I believe there used to be two warnings for this line. Now there are three.

hdgarrood: Ah, right, yeah this looks like something that we will need to fix. At the moment lots of the warnings are duplicated:

```

Warning 1 of 13:

  in module [33mMain[0m
  at tests/purs/warning/DuplicateExportRef.purs:14:1 - 36:28 (line 14, column 1 - line 36, column 28)

    Export list contains multiple references to type [33mX[0m


  See https://github.com/purescript/documentation/blob/master/errors/DuplicateExportRef.md for more information,

Warning 2 of 13:

  in module [33mMain[0m
  at tests/purs/warning/DuplicateExportRef.purs:14:1 - 36:28 (line 14, column 1 - line 36, column 28)

    Export list contains multiple references to type [33mX[0m


  See https://github.com/purescript/documentation/blob/master/errors/DuplicateExportRef.md for more information,
```

I think since the warning text just says that the export list contains multiple references to a particular thing, then we should only have one warning per group of duplicates, i.e. we should have at most one warning that the export list contains multiple references to a particular type or type class or whatever. If we wanted to change the warning so that the source span pointed to a particular duplicated reference, and change the text a bit so that it said something like "The export list already contains a reference to [thing], so this reference is redundant," then I think it would be okay to emit one warning for each duplicate after the first occurrence. That's probably not worth doing in this PR, though, and also I think if we wanted to consider that approach we'd need to be careful about how it might affect editor plugins and other tools which can automatically fix this warning.

milesfrain: Thinking it's best to just stick with the existing behavior for now, and let this PR stay more focused on the performance fix.

milesfrain:
Since preserving all duplicates is not the core focus of this PR, I just removed that change and left a note about it to be tackled in another issue.

Some tests still needed to be modified to account for minor reordering of warnings though. I assume that's because the list is actually sorted with this change.

hdgarrood:
Yeah, I think having the same warnings in a different order is fine. Thanks very much!

Also, from earlier,

> Would it be better to reorder the constructors to match this linter ordering?

Yes, that sounds like a good idea to me.

PR #3900, opened 2020-06-23
Add developer guide

milesfrain:


INSTALL.md
```diff
@@ -29,18 +29,7 @@ There are several other distributions of the PureScript compiler available, whic
 
 ## Compiling from source
 
-The easiest way is to use stack:
-
-```
-$ stack update
-$ stack unpack purescript
-$ cd purescript-x.y.z  # (replace x.y.z with whichever version you just downloaded)
-$ stack install --flag purescript:RELEASE
```
milesfrain, line 38: Is the `purescript:RELEASE` flag still necessary? I removed those from the new instructions since omitting it wasn't causing issues for me.

DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,155 @@
...
+```
+Then open the generated `purs.prof.html` in your web browser.
+
+## Testing
+
+See the [Contributing](CONTRIBUTING.md) readme for testing instructions.
+
+## Haskell IDE
+
+There are many Haskell IDE options. Feel free to contribute instructions for your preferred editor.
```
hdgarrood, line 101: I don't really want to maintain instructions for particular IDE projects or editors here. There's nothing particularly weird about the PureScript compiler that should necessitate tooling or editor-specific guides, and if we're including instructions for HIE and VSCode, why should we not also include instructions for emacs, vim, ghcid, ghcide, etc.? I think this file will very quickly get out of hand (and probably also out of date) with this approach. This sort of documentation is best placed in the particular tooling/editor plugin projects themselves, I think.

DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,155 @@
...
+
+These instructions are intended to help PureScript users more easily contribute to the compiler, even if this is your first Haskell project.
+
+## Prerequisites
+
+Install `stack`. [Instructions](https://docs.haskellstack.org/en/stable/README/).
+
+See the [PureScript Install Guide](INSTALL.md) for information on the `curses`/`ncurses` dependency.
+
+Update `stack` before proceeding:
```
hdgarrood, line 11: This command doesn't update stack, it only updates stack's package index.

DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,155 @@
...
+```
+stack update
+```
+
+## Select Release
+
+After cloning the PureScript compiler repo, `checkout` the tagged release that you'd like to build.
```
hdgarrood, line 18: If people are trying to contribute, they should almost always be starting from `master`, so I think I'd prefer to remove this bit.

INSTALL.md
hdgarrood: Can you please leave this file as-is? This file is really intended for people wanting to install the compiler, not to contribute to it. The RELEASE flag is needed for ensuring that the output of `purs --version` is correct.

DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,155 @@
+# Developer Guide
```
hdgarrood, line 1: I'm not sure this should be a new file; isn't it targeted at the same audience that the existing CONTRIBUTING.md file is?

milesfrain: The CONTRIBUTING.md file seemed for focused on policies rather than getting things running (apart from the "tests" section), but I'm happy to move this content there.

INSTALL.md
milesfrain: For users who just want to install the compiler (without contributing), would it be simpler to replace that section with:
```
stack install purescript
```
I'm having trouble envisioning what target audience would prefer the above instructions, since both of these approaches rely on trusting stack + hackage.

hdgarrood: No, `stack install purescript` would behave differently to what we have there; it tries to install it in a particular snapshot and doesn't use `purescript`'s own stack.yaml file.

DEVELOPER_GUIDE.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Profiling
+
+A profiling build is used to help diagnose performance issues with the compiler.
+
+Create a profiling build with:
+```
+stack build --profile
+```
+This will also take a while the first time it is run. Incremental builds are faster, and you may swap between standard and profiling builds without a rebuild penalty as work for each is saved in separate directories.
```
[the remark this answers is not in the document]
milesfrain: There was some confusion about this behavior in one of the meetings.
So I think it would be helpful to note that these faster rebuilds are available in stack 2.x.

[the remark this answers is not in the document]
hdgarrood: Can you be more specific? I don't see how it matters; if you need to do a profiling build, then there's not really any way around that; you just have to wait for as long as it takes.

DEVELOPER_GUIDE.md
milesfrain: I'll move this IDE section to a Discourse post.

DEVELOPER_GUIDE.md
milesfrain: Motivations for this section are:
* Letting users profile or fix an issue at a specific release. For example they may encounter new bugs if `master` contains breaking changes.
* Describe how to install a particular release (absorbing the removed INSTALL.md content)

But maybe I'm underestimating the abilities of readers to navigate releases.

hdgarrood: There's not much point in profiling or fixing an issue at a specific release, though, because it won't be merged if it can't be applied to master, or if the problem no longer exists in master.

DEVELOPER_GUIDE.md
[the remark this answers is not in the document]
milesfrain: The issue was that build workspaces were shared, so toggling build modes acted like clean rebuilds.
1. Standard build - slow
1. Standard build - fast
1. Profiling build - slow
1. Profiling build - fast
1. Standard build - slow (previous work clobbered by recent profiling build)
1. Profiling build - slow (previous work clobbered by recent standard build)

milesfrain:
It looks like the only merge conflict are these lines at the end of the file, which aren't even displayed markdown:
```
[the #purescript channel in FP Slack]: https://functionalprogramming.slack.com/
[Discourse]: https://discourse.purescript.org/
```
See https://github.com/purescript/purescript/pull/3924/files#r515518554

hdgarrood:
The main thing stopping me from approving this right now is that I think the content overlaps with too much of what's already in this file. I can help with that after 0.14.0 is released (since that's what I want to focus my effort on for now).

JordanMartinez:
One other note worth adding to this is [haskell-language-server](https://github.com/haskell/haskell-language-server#installation)

milesfrain:
> I can help with that after 0.14.0 is released (since that's what I want to focus my effort on for now).

@hdgarrood, Is now a good time to work on updating these instructions? I'm happy to collaborate on that.

hdgarrood:
Sorry, I don’t have the bandwidth for this right now. I don’t want to stand in the way of merging this so I will leave it up to other compiler maintainers to make the judgement of when the duplication has been addressed well enough that this can be merged.

CONTRIBUTING.md
```diff
@@ -51,5 +51,104 @@ This process can be performed automatically by running `make license-generator`.
...
 
+## Developer Guide
+
+The following instructions are intended to help PureScript users more easily contribute to the compiler, even if this is your first Haskell project.
+
+### Prerequisites
+
+Install `stack`. [Instructions](https://docs.haskellstack.org/en/stable/README/).
+
+See the [PureScript Install Guide](INSTALL.md) for information on the `curses`/`ncurses` dependency.
```
JordanMartinez, line 62: As of v0.14.2, this is no longer relevant. Can we drop this? Anyone contributing to the repo won't need to be concerned about this anymore.

CONTRIBUTING.md
```diff
@@ -51,5 +51,104 @@ This process can be performed automatically by running `make license-generator`.
...
+stack build
+```
+
+This will take a while the first time it is run.
+
+### Install
+
+If you'd like to test the build in a local shell, you can run:
+```
+stack exec bash
```
JordanMartinez, line 88: When I tried this on my version of stack, v2.5.1, it wouldn't do anything. I had to pass either a script or a heredoc to make it run something.

I also think it weird that this appears under the 'install' header despite being described as 'testing' the build locally.

CONTRIBUTING.md
```diff
@@ -51,5 +51,104 @@ This process can be performed automatically by running `make license-generator`.
...
+```
+
+This prepends the location of the compiler to your `$PATH`, so you can test your changes. This works with other tools that depend on the `purs` binary, such as `spago`. Note that rebuilds continue to use this same path, so you don't need to keep re-running `stack exec` to test new changes (assuming you're using a separate terminal for `stack exec`). Remember that `stack exec` must be launched from the same directory that you ran `stack build` from.
```
JordanMartinez, line 91: Should we drop this entirely and instead refer people to docs on that command?

CONTRIBUTING.md
```diff
@@ -51,5 +51,104 @@ This process can be performed automatically by running `make license-generator`.
...
+
+You may also install globally with:
+```
+stack install
```
JordanMartinez, line 95: I'm always hesitant to use this command because of the global installation. Shouldn't this be covered in `INSTALL.md` instead?

CONTRIBUTING.md
```diff
@@ -51,5 +51,104 @@ This process can be performed automatically by running `make license-generator`.
...
+stack install ghc-prof-flamegraph
+ghc-prof-flamegraph purs.prof
+firefox purs.svg
+```
+
+For more flamegraph customizations, you can also try [`stackcollapse-ghc`](https://github.com/marcin-rzeznicki/stackcollapse-ghc)
+
+#### [profiteur](https://github.com/jaspervdj/profiteur)
+```
+stack install profiteur
```
JordanMartinez, line 140:
```diff
- stack install profiteur
+ stack build --copy-compiler-tool profiteur
```

CONTRIBUTING.md
```diff
@@ -51,5 +51,104 @@ This process can be performed automatically by running `make license-generator`.
...
+
+This creates a `purs.prof` file. You can view the contents of this file directly, but it is often more convenient to use a visualizer.
+
+### Profile Visualizers
+
+Each of these produces a clickable visual display of profiling info. Feel free to open the output files in the web browser of your choice. These examples use `firefox`.
+
+#### [ghc-prof-flamegraph](https://github.com/fpco/ghc-prof-flamegraph)
+```
+stack install ghc-prof-flamegraph
```
JordanMartinez, line 131:
```diff
- stack install ghc-prof-flamegraph
+ stack build --copy-compiler-tool ghc-prof-flamegraph
```

JordanMartinez (changes requested):
I think the profiling guide is worth keeping in this PR. I think more installation things should be left out and that we should instead document that in the `INSTALL.md` file. I also think that we should explain less of `stack` and just direct users to its documentation. Having a short "here's the commands you would run in bash" seems fine to me, but not necessarily explaining what they do and why as they can read that in the stack docs.

CONTRIBUTING.md
milesfrain: `stack install` copies to:
```
~/.local/bin/
```

`stack build --copy-compiler-tool` copies to:
```
/.stack/compiler-tools/x86_64-linux-tinfo6/ghc-8.10.4/bin/
```

Is there a way to point to the `compiler-tools` location when running the following command in this readme? Is it necessary to have per-compiler versions of this tool?

CONTRIBUTING.md
milesfrain: `stack exec bash` should create a sub-shell with the recently-built purs version.

```
➜  complier git:(master) ✗ purs --version
0.14.2

➜  complier git:(master) ✗ stack exec bash

miles@x1c5:~/projects/purescript/complier$ purs --version
0.14.2 [development build; commit: f1953214775945b65ba53ae903b4238c352dcd29 DIRTY]

miles@x1c5:~/projects/purescript/complier$ exit

➜  complier git:(master) ✗ purs --version 
0.14.2
```

Maybe this section could be named something other than 'Install', such as 'Integration Testing'. Perhaps change the original 'Testing` header to 'Unit Testing'.

CONTRIBUTING.md
milesfrain: I don't think this usage info is described effectively in the [stack docs](https://docs.haskellstack.org/en/stable/GUIDE/#stack-exec).

CONTRIBUTING.md
milesfrain: I figure some contributors are editing the compiler to fix an issue that's bothering them, and they may want to continue using using their patched version in the meantime while waiting for the next release (without having to run `stack exec` in each new shell). At least that was my use case when writing this guide.

My interpretation of `INSTALL.md` is that it's for folks who just want to grab the latest official version. It seems more efficient to let these two lines live here, rather than edit the [related section](https://github.com/purescript/purescript/blob/master/INSTALL.md#compiling-from-source) of that guide to also include additional instructions for installing untagged versions.

CONTRIBUTING.md
JordanMartinez: > Is it necessary to have per-compiler versions of this tool?

I'm not sure, but it would remove the surprise that people may have when they find that the tool no longer works when using it with other Haskell projects they have on their local computer. I think the main reason to not use `stack install` is that there isn't a corresponding `stack uninstall`. 

If this change was made, the rest of the command would need to be updated to use `stack exec bash`:
```
stack build --copy-compiler-tool ghc-prof-flamegraph
stack exec bash <<EOF
ghc-prof-flamegraph purs.prof
EOF
firefox purs.svg
```

CONTRIBUTING.md
JordanMartinez: Huh... Seems there's something going on with my terminal. I was able to reproduce most of the above behavior, but the info before the `$` part didn't change after running `stack exec bash`

CONTRIBUTING.md
JordanMartinez: Hm... It does mention some part of the idea [here](https://docs.haskellstack.org/en/stable/GUIDE/#exec), but it's still not as clear as what you've written. 

I'd rather fix this upstream. Would you be up for submitting a PR to that repo to update the docs to clarify that? Then we don't have to explain it here when it should really be explained there.

> Remember that `stack exec` must be launched from the same directory that you ran `stack build` from.

I don't think that's quite true. I think it works as long as you are in a subdirectory of the root folder that contains the stack project.

CONTRIBUTING.md
JordanMartinez: Good point. Maybe add a note here that one should use this 'global install' only temporarily for the purpose you mentioned above and that they can remove it by running `stack path --local-bin` to see which folder stores the binary and delete the binary once done?

CONTRIBUTING.md
JordanMartinez: I think the header should be `Exploring and testing a locally compiled version of PureScript` because the purpose of this section is to allow others to just try things and see what happens.

CONTRIBUTING.md
JordanMartinez: Or perhaps we just say

> f you're curious how `stack exec bash` works, run `echo $PATH` then `stack exec bash` then `echo $PATH` again, then `exit`, and finally `echo $PATH` a final time.)

CONTRIBUTING.md
milesfrain: Changed heading to `Running a locally-compiled version of PureScript`

CONTRIBUTING.md
milesfrain: > Would you be up for submitting a PR to that repo to update the docs to clarify that?

I made a minor PR to the `stack` docs in https://github.com/commercialhaskell/stack/pull/5574

I'll leave out these two additional `stack exec bash` tips from this guide, and trust that users can figure this out on their own through trial and error:
- You must call `stack exec` from somewhere within the project directory
- You only need to call `stack exec` once

I added some additional tips (e.g. unix `which`) on how to determine which `purs` version is actually being run`.

CONTRIBUTING.md
milesfrain: I think `which purs` beats `stack path --local-bin` for tracking down binaries. Added a note about that utility to this and the previous section.

CONTRIBUTING.md
milesfrain: This combo works:
```
stack build --copy-compiler-tool profiteur
stack exec -- profiteur purs.prof
```

Updated instructions.

CONTRIBUTING.md
JordanMartinez: Thanks. Yeah, using `which` is a better approach than `stack path --local-bin`.

CONTRIBUTING.md
JordanMartinez: Good idea!

JordanMartinez:
In the resources section at the bottom, could you also add a reference to [Haskell Language Server](https://github.com/haskell/haskell-language-server#installation)? That seems to be the direction things are going now.

JordanMartinez:
Otherwise, this looks good to me.

JordanMartinez (approved):
Thanks! LGTM.

JordanMartinez:
@milesfrain Could you merge this PR with `master`, and then add a changelog entry file in the `CHANGELOG.d` folder? I believe we'll be able to merge this afterwards.

JordanMartinez:
Thanks Miles! @thomashoneyman or @f-f, can one of you review this? If it's good to go, I'd like to merge it.

f-f (approved):
This looks great! :clap:

JordanMartinez:
Thanks @f-f! And thanks again @milesfrain for working on this!

f-f:
You're welcome :slightly_smiling_face:

PR #3902, opened 2020-06-26
Refactor Ord instances

milesfrain:
Follow-up to https://github.com/purescript/purescript/pull/3899#issuecomment-648871605

Also applied the `orderOf` strategy to a few other places.

Some additional discussion about strategies for skipping `SourceSpan` comparisons in https://github.com/purescript/purescript/pull/3265

### DIFF
--- a/lib/purescript-ast/src/Language/PureScript/AST/Binders.hs
+++ b/lib/purescript-ast/src/Language/PureScript/AST/Binders.hs
@@ -71,100 +71,62 @@ data Binder
 -- the `Ord` instance was needed for the speed-up, but I did not want the `Eq`
 -- to have mismatched behavior.
 instance Eq Binder where
-  (==) NullBinder NullBinder = True
-  (==) NullBinder _ = False
-
-  (==) (LiteralBinder _ lb) (LiteralBinder _ lb') = (==) lb lb'
-  (==) LiteralBinder{} _ = False
-
-  (==) (VarBinder _ ident) (VarBinder _ ident') = (==) ident ident'
-  (==) VarBinder{} _ = False
-
-  (==) (ConstructorBinder _ qpc bs) (ConstructorBinder _ qpc' bs') =
-    (==) qpc qpc' && (==) bs bs'
-  (==) ConstructorBinder{} _ = False
-
-  (==) (OpBinder _ qov) (OpBinder _ qov') =
-    (==) qov qov'
-  (==) OpBinder{} _ = False
-
-  (==) (BinaryNoParensBinder b1 b2 b3) (BinaryNoParensBinder b1' b2' b3') =
-    (==) b1 b1' && (==) b2 b2' && (==) b3 b3'
-  (==) BinaryNoParensBinder{} _ = False
-
-  (==) (ParensInBinder b) (ParensInBinder b') =
-    (==) b b'
-  (==) ParensInBinder{} _ = False
-
-  (==) (NamedBinder _ ident b) (NamedBinder _ ident' b') =
-    (==) ident ident' && (==) b b'
-  (==) NamedBinder{} _ = False
-
-  (==) (PositionedBinder _ comments b) (PositionedBinder _ comments' b') =
-    (==) comments comments' && (==) b b'
-  (==) PositionedBinder{} _ = False
-
-  (==) (TypedBinder ty b) (TypedBinder ty' b') =
-    (==) ty ty' && (==) b b'
-  (==) TypedBinder{} _ = False
+  NullBinder == NullBinder =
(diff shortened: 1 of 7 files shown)


lib/purescript-ast/src/Language/PureScript/AST/Binders.hs
```diff
@@ -71,100 +71,62 @@ data Binder
 -- the `Ord` instance was needed for the speed-up, but I did not want the `Eq`
 -- to have mismatched behavior.
 instance Eq Binder where
-  (==) NullBinder NullBinder = True
-  (==) NullBinder _ = False
```
milesfrain, line 75: This block of changes moves all lines like these to a final:
```hs
_ == _ = False
```

lib/purescript-ast/src/Language/PureScript/AST/Binders.hs
```diff
@@ -71,100 +71,62 @@ data Binder
...
-    (==) comments comments' && (==) b b'
-  (==) PositionedBinder{} _ = False
-
-  (==) (TypedBinder ty b) (TypedBinder ty' b') =
-    (==) ty ty' && (==) b b'
-  (==) TypedBinder{} _ = False
+  NullBinder == NullBinder =
+    True
+  (LiteralBinder _ lb) == (LiteralBinder _ lb') =
+    lb == lb'
```
milesfrain, line 77: Not sure what the line-wrapping coding standard is. Could either:
* Use one line for each expression. Lines may grow long.
* Break after cutoff point, e.g. 80 characters. I feel this adds inconsistency to the structure.
* Break at each assignment. What's done in this PR.

lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
```diff
@@ -149,14 +149,18 @@ importPrim =
 --
 data DeclarationRef
   -- |
-  -- A type constructor with data constructors
+  -- A type class
```
milesfrain, line 152: Reordered to match linting order

lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
```diff
@@ -223,29 +223,6 @@ data ExportSource =
   }
   deriving (Eq, Ord, Show, Generic, NFData, Serialise)
 
--- enable sorting lists of explicitly imported refs when suggesting imports in linting, IDE, etc.
--- not an Ord because this implementation is not consistent with its Eq instance.
--- think of it as a notion of contextual, not inherent, ordering.
```
milesfrain, line 228: I don't think we need this comment anymore, since it now follows the order of the constructors.

lib/purescript-ast/src/Language/PureScript/Types.hs
```diff
@@ -728,66 +728,41 @@ eqMaybeType _ _ = False
...
-
 compareType (TypeApp _ a b) (TypeApp _ a' b') = compareType a a' <> compareType b b'
-compareType (TypeApp {}) _ = LT
-compareType _ (TypeApp {}) = GT
-
 compareType (KindApp _ a b) (KindApp _ a' b') = compareType a a' <> compareType b b'
-compareType (KindApp {}) _ = LT
-compareType _ (KindApp {}) = GT
-
 compareType (ForAll _ a b c d) (ForAll _ a' b' c' d') = compare a a' <> compareMaybeType b b' <> compareType c c' <> compare d d'
```
milesfrain, line 738: Here's an example of using long lines for each expression. This is another good candidate for the above coding style question.

kl0tl:
Is there a reason for not defining the `Eq` instances of `Binder` and `DeclarationRef` via their `Ord` instance and defining `eqType` via `compareType`?

natefaubion:
> Is there a reason for not defining the Eq instances of Binder and DeclarationRef via their Ord instance and defining eqType via compareType?

Maybe efficiency? I don't think we've ever measured it.

lib/purescript-ast/src/Language/PureScript/AST/Binders.hs
hdgarrood: Breaking at each assignment like you've done here looks good to me. These things aren't really consistent across the compiler codebase at the moment.

hdgarrood (approved):
Looks good, thanks!

hdgarrood:
As is, this change already removes a bunch of code so I'm happy to leave the question of whether we should define Eq via Ord for later.

thomashoneyman (approved):
👍 thanks for this work, @milesfrain!

Issue #3696, opened 2019-07-05
Consider using static linking in binary distributions


hdgarrood:
Many of the difficulties people have had with getting the compiler running boil down to shared libraries at the versions the compiler is expecting being difficult to get hold of via their distribution's standard package manager. For example:

* https://github.com/purescript/purescript/issues/3122
* https://github.com/purescript/purescript/issues/2026
* https://github.com/purescript/purescript/issues/2176
* https://github.com/purescript/purescript/issues/2147
* https://github.com/purescript/purescript/issues/2175#issuecomment-223844287
* https://github.com/purescript/documentation/issues/119

There's a small drawback with statically linked binaries though: they would be a little larger. I think static linking is probably worth it, on balance, although we should see how easy it is to actually produce statically linked binaries first.

hdgarrood:
It turns out that the `terminfo` library, which is the main cause of these kinds of errors, can’t really be statically linked (see https://github.com/judah/terminfo/issues/36 ). However, it turns out that it is an optional dependency of `haskeline`, as `haskeline` has a build flag to control whether `terminfo` should be used. From haskeline’s Cabal file:

> -- There are three main advantages to the terminfo backend over the portable,
> -- "dumb" alternative.  First, it enables more efficient control sequences
> -- when redrawing the input.  Second, and more importantly, it enables us
> -- to draw on multiple lines, so we can wrap long input strings.  And third,
> -- the backend adds some extra key sequences such as forwards delete.
> --
> -- (The "dumb" terminal also allows editing of long input strings, but is
> -- restricted to only one line and thus only shows part of the input at once.)

https://github.com/judah/haskeline/blob/master/haskeline.cabal#L28

The drawbacks

rainbyte:
Is this for the REPL? I didn't even know about multiline strings there.
Does so many people use this feature?

Personally, I think being able to easily install PureScript is a priority.

Swonkie:
> Personally, I think being able to easily install PureScript is a priority.

I strongly agree. It is literally the most important "feature" of _any_ software. If the official and documented way of installing PureScript doesn't work, that looks extremely bad and most users experiencing any difficulties this early on, while trying something new, will simply give up and look for something else than PureScript.

In my case, installation on Manjaro doesn't work beause of this issue:
https://github.com/purescript/documentation/issues/119

hdgarrood:
I tested out what the repl is like with the `terminfo` flag disabled. Here's a side-by-side comparison:
[...]

PR #3903, opened 2020-06-26
Add troubleshooting steps for libtinfo and EACCES errors

milesfrain:
See https://github.com/purescript/documentation/pull/323#issuecomment-649905179

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -46,23 +46,17 @@ If you don't have stack installed, please see the [stack install documentation](
 
 The PureScript REPL depends on the `curses` library (via the Haskell package
 `terminfo`). If you are having difficulty running the compiler, it may be
-because the `curses` library is missing.
+because the `curses` library is missing. This problem may appear as a `libtinfo`
+error:
+```
+error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
+```
 
 On Linux, you will probably need to install `ncurses` manually. On Ubuntu, for
 example, this can be done by running:
 
 ```
 $ sudo apt install libncurses5-dev
-```
-
-Another `curses` dependency may be `libtinfo`. If you encounter this error:
-```
-error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
-```
-
-The solution is to install that dependency. Here's the Ubuntu command for this:
-```
-sudo apt install libtinfo5
 ```
 
 ## EACCES error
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -52,5 +52,24 @@ On Linux, you will probably need to install `ncurses` manually. On Ubuntu, for
 example, this can be done by running:
 
 ```
-$ sudo apt-get install libncurses5-dev
+$ sudo apt install libncurses5-dev
 ```
+
+Another `curses` dependency may be `libtinfo`. If you encounter this error:
+```
```
hdgarrood, line 59: This makes it sound like it's a different problem from the one listed above, but it's not; `libncurses5-dev` depends on `libtinfo5`, so installing `libncurses5-dev` is enough to fix the issue. Adding the "error while loading shared libraries" message here would be good so that it's easier for people to see whether this is the problem they're running into, but other than that I think this section should be left alone.

INSTALL.md
```diff
@@ -52,5 +52,24 @@ On Linux, you will probably need to install `ncurses` manually. On Ubuntu, for
...
+
+## EACCES error
+
+If you encounter this error while trying to install via `npm`:
+```
+Error: EACCES: permission denied
+```
+
+The best solution is to install [Node.js and npm via a node version manager](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#using-a-node-version-manager-to-install-nodejs-and-npm). This error is due to permissions issues when installing packages globally. You can read more about [this error here](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
```
hdgarrood, line 75: It's usually clearer to find an alternative to the word "here" for use in link text, especially for screen readers. How about removing the preceding sentence and writing: "You can read more about this error in npm's guide to [resolving EACCES permissions errors when installing packages globally](...)".

milesfrain:
@hdgarrood Applied feedback. Ready for re-review

INSTALL.md
milesfrain: As demonstrated in https://github.com/purescript/npm-installer/pull/26#issuecomment-657253719 `libtinfo5` is required. Installing `libncurses5-dev` will not fix this error.

hdgarrood: This is still the same issue; what's happening here is that libncurses5-dev depends on libtinfo6 in Ubuntu 20.04, whereas it depended on libtinfo5 in previous Ubuntu versions; see https://packages.ubuntu.com/focal/libtinfo6. I still don't want to list  two separate fixes here. Since `libtinfo5` is the package which provides the actual file we need (specifically `/lib/x86_64_linux_gnu/libtinfo.so.5`) hopefully we can just instruct users to install that. I'm testing to see whether instructing users to install `libtinfo5` will work on all supported Ubuntu versions.

hdgarrood: I've also just remembered that whatever we specify here needs to work for building the compiler from source too. From my testing, it seems `libncurses5-dev` is needed for compiling the compiler from source (in which case you end up with a `purs` which depends on libtinfo6), but `libtinfo5` is needed for using the prebuilt binaries, since those depend on libtinfo5. I think we could just ask the user to install both:
```
$ sudo apt install libtinfo5 libncurses5-dev
```

milesfrain: Interesting. Nice follow-up investigation.

> I still don't want to list two separate fixes here. I think we could just ask the user to install both.

These seem like two separate dependency problems though (stack build versus npm download). I think it's better to only ask users to install the dependencies they actually need.

hdgarrood: Building from source isn't separate from the npm installer though: the npm installer falls back to building from source if the prebuilt binaries fail. Systems where the prebuilt binaries will never work aren't that uncommon; they won't work on any architecture other than x86_64, and they link against glibc which means that they don't work on any Linux distribution not based on glibc, such as Alpine. Additionally, the `libncurses5-dev` package is only about 1MB when installed.

hdgarrood: Incidentally we can also avoid this entirely by dropping the `libtinfo` dependency. See https://github.com/purescript/purescript/issues/3696#issuecomment-657282303

PR #3908, opened 2020-07-08
Copy default-extensions.yaml to subprojects

i-am-the-slime:
Both `purescript-ast` and `purescript-cst` are DRY right now in that they use the same `default-extensions.yaml` as the main project. However, that prevents one from depending on these subprojects from stack (at least I couldn't make it work). That's because the relative reference two folders down points to some guts in macOS.

With the proposed change it works. Of course there's the downside of a possibility of these files going out of sync, but that seems like the lesser evil to me. I'm sure that you have better ideas than this change of how to solve this anyway, I'll leave the PR as a mere suggestion.


hdgarrood:
Thanks! This approach seems fine to me, although could you please add a comment at the top of each file to help avoid having them get out of sync?
```
# This file should be kept in sync with the other default-extensions.yaml files in this repository. The files are located at:
# - default-extensions.yaml (the repository root),
# - lib/purescript-ast/default-extensions.yaml,
# - lib/purescript-cst/default-extensions.yaml
```

i-am-the-slime:
Done!

PR #3910, opened 2020-07-11
Store the inferred role environment in the typechecker monad state

kl0tl:
This lifts role inference to the typechecker monad so we can store a role environment in `CheckState` instead of recomputing one each time we infer a binding group roles. This was extracted from https://github.com/purescript/purescript/pull/3860 to not delay the merge and discuss separately. See https://github.com/purescript/purescript/pull/3860#issuecomment-652776016 for context.

I’m happy to be wrong but I’m not sure checked roles, inferred roles and declared roles could all be stored under the types field of the environment because we need to compare declared roles to inferred roles in order to check they’re not more permissive and types are only inserted in the environment once their roles have been checked.

### DIFF
--- a/lib/purescript-ast/src/Language/PureScript/Environment.hs
+++ b/lib/purescript-ast/src/Language/PureScript/Environment.hs
@@ -35,7 +35,7 @@ data Environment = Environment
   , dataConstructors :: M.Map (Qualified (ProperName 'ConstructorName)) (DataDeclType, ProperName 'TypeName, SourceType, [Ident])
   -- ^ Data constructors currently in scope, along with their associated type
   -- constructor name, argument types and return type.
-  , roleDeclarations :: M.Map (Qualified (ProperName 'TypeName)) [Role]
+  , roleDeclarations :: RoleEnv
   -- ^ Explicit role declarations currently in scope. Note that this field is
   -- only used to store declared roles temporarily until they can be checked;
   -- to find a type's real checked and/or inferred roles, refer to the TypeKind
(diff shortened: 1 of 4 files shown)


hdgarrood:
I'd prefer not to do this if it ends up adding more mutable state (like this new `checkRoleEnv` field). I think there are two separate potential refactoring tasks here:

* Make the `roleDeclarations` field obsolete using the strategy described in https://github.com/purescript/purescript/pull/3860#issuecomment-652776016:

  > Speaking of minimising the amount of mutable state we have, it occurred to me that now we are dealing with data binding groups sensibly, we can probably make the roleDeclarations field obsolete by putting role annotations in the same binding group as the data types they refer to, so that the data we need is available at the time we actually need it, rather than having to store them in the environment separately first and pull them back out later.

  Specifically, I'm imagining changing the Sugar.BindingGroups step so that any explicit role declarations end up as part of the same DataBindingGroupDeclaration as the data declarations they describe - I think this should be possible by having edges between the data declaration and the role declaration in both directions in the graph. That way the current addExplicitRoleDeclaration function can become obsolete; if we come across a role declaration which isn't inside a DataBindingGroupDeclaration, we can throw an internal compiler error, because that shouldn't ever happen.

* Store inferred roles in the `types` field of the Environment during inference. After making the roleDeclarations field obsolete, we should hopefully have access to the data we need regarding declared roles without needing to store them in the environment at all. I don't think it's necessarily a problem to store information we don't yet know for sure to be accurate in the Environment, as an error does eventually get thrown if there is a problem. However, on reflection it's less clear to me that this would improve the quality of the code, and the fact that the `types` field of the Environment contains a bunch of information besides roles might make this a bit awkward. So I'm tempted to not worry about doing this for now.

That is, I'd quite like to do the first at some point (maybe post-0.14.0), but I'm not sure the second is worth doing.

hdgarrood:
I'm not really that keen on the first refactoring I proposed here now either (see https://github.com/purescript/purescript/pull/3881#issuecomment-691522624) so I think I'm going to close this. I'm sorry for leading you astray!

Issue #687, opened 2014-11-07
Common subexpression elimination


garyb:
This would need some thinking about, but it'd be really nice if we could improve the likes of this:

``` js
var ap = function (__dict_Monad_14) {
    return function (f) {
        return function (a) {
            return $greater$greater$eq(__dict_Monad_14["__superclass_Prelude.Bind_1"]())(f)(function (_2) {
                return $greater$greater$eq(__dict_Monad_14["__superclass_Prelude.Bind_1"]())(a)(function (_1) {
                    return $$return(__dict_Monad_14)(_2(_1));
                });
            });
        };
    };
};
```

into something like this:

``` js
var ap = function (__dict_Monad_14) {
    var __dict_Monad_14_Bind = __dict_Monad_14["__superclass_Prelude.Bind_1"]()
    var $greater$greater$eq_1 = $greater$greater$eq(__dict_Monad_14_Bind);
    var $$return_1 = $$return(__dict_Monad_14);
    return function (f) {
        return function (a) {
            return $greater$greater$eq_1(f)(function (_2) {
                return $greater$greater$eq_1(a)(function (_1) {
                    return $$return_1(_2(_1));
                });
            });
        };
    };
};
```

(Essentially declaring all instance related members and superclass instances as local variables as soon as is possible.)

Apart from improving readability somewhat, using the local `$greater$greater$eq_1` over `$greater$greater$eq(__dict_Monad_14["__superclass_Prelude.Bind_1"]())` everywhere should add up to a nice performance boost too. Most of optimisations I worked on for instances were to make constructing instances cheaper as they were the largest overhead, if we can reduce the number of constructions that's even better!

paf31:
Is there any way to identify TCDs in the core AST?

garyb:
Yep, the `IsTypeClassConstructor` from `Meta` is added to the value for the "class constructor" function during desugaring: https://github.com/purescript/purescript/blob/7b5cb4b3481d1119509c9a26b4d941919c944e1f/src/Language/PureScript/CoreFn/Desugar.hs#L174-L177

garyb:
This might be a bit clearer actually when the revisited typeclass representation is in.

rhendric:
[...]

PR #3915, opened 2020-08-03
Float compiler-synthesized function applications

rhendric:
This commit implements a common subexpression elimination pass during
CoreFn optimization. The optimizer only targets `App`s that the compiler
has created, meaning: applying functions to typeclass dictionaries,
accessing superclasses from a typeclass dictionary, or creating new
instances of `IsSymbol`. Furthermore, the pass only moves subexpressions
if they can be moved to an outer function or if the subexpression
appears more than once inside the scope to which it would be moved.

---

This is a start at what I'm told is the uncontroversial part of #687. Here's some before-and-after:

(before)
```js
var ap = function (dictMonad) {
    return function (f) {
        return function (a) {
            return Control_Bind.bind(dictMonad.Bind1())(f)(function (f$prime) {
                return Control_Bind.bind(dictMonad.Bind1())(a)(function (a$prime) {
                    return Control_Applicative.pure(dictMonad.Applicative0())(f$prime(a$prime));
                });
            });
        };
    };
};
```

(after)
```js
var ap = function (dictMonad) {
    var bind = Control_Bind.bind(dictMonad.Bind1());
    var pure = Control_Applicative.pure(dictMonad.Applicative0());
    return function (f) {
        return function (a) {
            return bind(f)(function (f$prime) {
                return bind(a)(function (a$prime) {
                    return pure(f$prime(a$prime));
                });
            });
        };
    };
};
```

I don't know many public PureScript projects that have sufficiently long-running automated tests to be suitable for benchmarking, but purescript-bytestrings is one. On that project, this patch (backported to the 0.13.x branch) increased total compilation time by about 2% and decreased test execution time by about 20%. I would be happy to run the same benchmarks on any other project you suggest.

In order to verify that the optimization happens, a new (small) suite of tests in `tests/purs/optimize` has been added. In this folder, `*.purs` files are compiled and the result is compared to the contents of the corresponding `.out.js` file. These tests are potentially more expensive than usual in terms of maintenance burden, because they are sensitive to any change in the compiled output, but they can be used to prevent regressions in optimizations that are too subtle to be tested by other means.

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -63,6 +63,7 @@ common defaults
     DeriveTraversable
     DeriveGeneric
     DerivingStrategies
+    DerivingVia
     EmptyDataDecls
     FlexibleContexts
     FlexibleInstances
@@ -209,6 +210,7 @@ library
     Language.PureScript.CoreFn
     Language.PureScript.CoreFn.Ann
     Language.PureScript.CoreFn.Binders
+    Language.PureScript.CoreFn.CSE
     Language.PureScript.CoreFn.Desugar
     Language.PureScript.CoreFn.Expr
     Language.PureScript.CoreFn.FromJSON
--- a/src/Language/PureScript/CoreFn/CSE.hs
+++ b/src/Language/PureScript/CoreFn/CSE.hs
@@ -1,0 +1,428 @@
+{-# LANGUAGE TemplateHaskell #-}
+-- | This module performs limited common subexpression elimination
+module Language.PureScript.CoreFn.CSE (optimizeCommonSubexpressions) where
+
+import Protolude hiding (pass)
+
+import Control.Lens
+import Control.Monad.Supply (Supply)
+import Control.Monad.Supply.Class (MonadSupply)
+import Control.Monad.RWS (MonadWriter, RWST, censor, evalRWST, listen, pass, tell)
+import Data.Bitraversable (bitraverse)
+import Data.Functor.Compose (Compose(..))
+import qualified Data.IntMap.Monoidal as IM
+import qualified Data.IntSet as IS
+import qualified Data.Map as M
+import Data.Maybe (fromJust)
+import Data.Semigroup (Min(..))
+import Data.Semigroup.Generic (GenericSemigroupMonoid(..))
+
+import Language.PureScript.AST.Literals
+import Language.PureScript.AST.SourcePos (nullSourceSpan)
+import qualified Language.PureScript.Constants.Prelude as C
+import Language.PureScript.CoreFn.Ann (Ann)
+import Language.PureScript.CoreFn.Binders
+import Language.PureScript.CoreFn.Expr
+import Language.PureScript.CoreFn.Meta (Meta(IsSyntheticApp))
+import Language.PureScript.CoreFn.Traversals
+import Language.PureScript.Environment (dictTypeName)
+import Language.PureScript.Names
+import Language.PureScript.PSString (decodeString)
+
+-- |
+-- `discuss f m` is an action that listens to the output of `m`, passes that
+-- and its value through `f`, and uses (only) the value of the result to set
+-- the new value and output. (Any output produced via the monad in `f` is
+-- ignored, though other monadic effects will hold.)
+--
+discuss :: MonadWriter w m => ((a, w) -> m (b, w)) -> m a -> m b
+discuss f = pass . fmap (second const) . (f <=< listen)
+
+-- |
+-- Modify the target of an optic in the state with a monadic computation that
+-- returns some extra information of type `r` in a tuple.
+--
+-- I would prefer that this be a named function, but I don't know what to name
+-- it. I went with symbols instead because the function that this operator most
+-- resembles is `(%%=)`, which doesn't have a textual name as far as I know.
+-- Compare the following (approximate) types:
+--
+-- @
+-- (%%=)  :: MonadState s m => Lens s s a b -> (a ->   (r, b)) -> m r
+-- (%%<~) :: MonadState s m => Lens s s a b -> (a -> m (r, b)) -> m r
+-- @
+--
+-- Replacing the `=` with `<~` was inspired by analogy with the following pair:
+--
+-- @
+-- (.=) :: MonadState s m => Lens s s a b ->   b -> m ()
+-- (<~) :: MonadState s m => Lens s s a b -> m b -> m ()
+-- @
+--
+-- I regret any confusion that ensues.
+--
+-- Note that there are two interpretations that could reasonably be expected
+-- for this type.
+--
+-- @
+-- (%%<~) :: MonadState s m => Lens s s a b -> (a -> m (r, b)) -> m r
+-- @
+--
+-- One is:
+-- * Get the focused `a` value from the monad
+-- * Run the computation
+-- * Get the new state from the returned monad
+-- * Take the returned `b` value and set it in the new state
+--
+-- The other is:
+-- * Get the focused `a` value from the monad
+-- * Run the computation
+-- * Take the returned `b` value and set it in the *original* state
+-- * Put the result into the returned monad
+--
+-- This operator corresponds to the second interpretation. The purpose of this,
+-- and part of the purpose of having this operator at all instead of composing
+-- simpler operators, is to enable using the lens only once (on the original
+-- state) instead of twice (for a get and a set on different states).
+--
+(%%<~)
+  :: MonadState s m
+  => ((a -> Compose m ((,) r) b) -> s -> Compose m ((,) r) s)
+     -- ^ please read as Lens s s a b
+  -> (a -> m (r, b))
+  -> m r
+l %%<~ f = get >>= getCompose . l (Compose . f) >>= state . const
+infix 4 %%<~
+
+-- |
+-- A PluralityMap is like a weaker multiset: like a multiset, it can hold
+-- several of the same value, but instead of keeping track of their exact
+-- counts, it only records whether there is one (False) or more than one
+-- (True).
+--
+newtype PluralityMap k = PluralityMap { getPluralityMap :: M.Map k Bool }
+
+instance Ord k => Semigroup (PluralityMap k) where
+  PluralityMap l <> PluralityMap r =
+    let
+      l' = M.mapWithKey (\k -> (|| k `M.member` r)) l
+    in PluralityMap $ l' `M.union` r
+
+instance Ord k => Monoid (PluralityMap k) where
+  mempty = PluralityMap M.empty
+
+data BindingType = NonRecursive | Recursive deriving Eq
+
+-- |
+-- Record summary data about an expression.
+--
+data CSESummary = CSESummary
+  { _scopesUsed    :: IS.IntSet
+    -- ^ set of the scope numbers used in this expression
+  , _noFloatWithin :: Maybe (Min Int)
+    -- ^ optionally a scope within which this expression is not to be floated
+    -- (because the expression uses an identifier bound recursively in that
+    -- scope)
+  , _plurality     :: PluralityMap Ident
+    -- ^ which floated identifiers are used more than once in this expression
+    -- (note that a single use inside an Abs will be considered multiple uses,
+    -- as this pass doesn't know when/how many times an Abs will be executed)
+  , _newBindings   :: IM.MonoidalIntMap [(Ident, (PluralityMap Ident, Expr Ann))]
+    -- ^ floated bindings, organized by scope number
+  , _toBeReinlined :: M.Map Ident (Expr Ann)
+    -- ^ a map of floated identifiers that did not end up getting bound and
+    -- will need to be reinlined at the end of the pass
+  }
+  deriving Generic
+  deriving (Semigroup, Monoid) via GenericSemigroupMonoid CSESummary
+
+-- |
+-- Append a value at a given scope depth.
+--
+addToScope :: Semigroup v => Int -> v -> IM.MonoidalIntMap v -> IM.MonoidalIntMap v
+addToScope depth v
+  = IM.alter (Just . maybe v (<> v)) depth
+
+-- |
+-- Remove and return an entire scope from a map of bindings.
+--
+popScope :: Monoid v => Int -> IM.MonoidalIntMap v -> (v, IM.MonoidalIntMap v)
+popScope depth
+  = first fold . IM.updateLookupWithKey (\_ _ -> Nothing) depth
+
+-- |
+-- Describe the context of an expression.
+--
+data CSEEnvironment = CSEEnvironment
+  { _depth :: Int
+    -- ^ number of enclosing binding scopes (this includes not only Abs, but
+    -- Let and CaseAlternative bindings)
+  , _bound :: M.Map Ident (Int, BindingType)
+    -- ^ map from identifiers to depth in which they are bound and whether
+    -- or not the binding is recursive
+  }
+
+makeLenses ''CSESummary
+makeLenses ''CSEEnvironment
+
+-- |
+-- Map from the shape of an expression to an identifier created to represent
+-- that expression, organized by scope depth.
+--
+type CSEState = IM.MonoidalIntMap (M.Map (Expr ()) Ident)
+
+-- |
+-- The monad in which CSE takes place.
+--
+type CSEMonad a = RWST CSEEnvironment CSESummary CSEState Supply a
+
+type HasCSEReader = MonadReader CSEEnvironment
+type HasCSEWriter = MonadWriter CSESummary
+type HasCSEState = MonadState CSEState
+
+-- |
+-- Run a CSEMonad computation; the return value is augmented with a map of
+-- identifiers that should be replaced in the final expression because they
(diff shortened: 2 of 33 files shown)


natefaubion:
One potential benchmark might be the one in `purescript-run`
https://github.com/natefaubion/purescript-run/blob/0946aefaaa360910d72a298d702d72af07b1adcb/test/Bench.purs

rhendric:
Thanks! I will try that as soon as I fix the TCO regression in `purescript-free` that this revealed. :grin:

rhendric:
I could not measure a significant time difference compiling `purescript-run` with 0.13.8 versus 0.13.8 plus this patch.

Running Bench.purs produced the following output:

0.13.8:
```
Transformers (monomorphic/trampoline)
mean   = 105.21 ms
stddev = 33.67 ms
min    = 82.96 ms
max    = 352.80 ms
Transformers (monomorphic/identity)
mean   = 46.71 ms
stddev = 14.51 ms
min    = 37.98 ms
max    = 127.36 ms
Transformers (mtl/trampoline)
mean   = 182.59 ms
stddev = 35.99 ms
min    = 153.22 ms
max    = 415.49 ms
Transformers (mtl/identity)
mean   = 100.16 ms
stddev = 24.06 ms
min    = 75.20 ms
max    = 196.17 ms
Run (free)
mean   = 28.73 ms
stddev = 31.97 ms
min    = 21.52 ms
max    = 299.17 ms
```

0.13.8 + patch:
```
Transformers (monomorphic/trampoline)
mean   = 62.52 ms
stddev = 33.37 ms
min    = 52.73 ms
max    = 376.43 ms
Transformers (monomorphic/identity)
mean   = 19.94 ms
stddev = 4.69 ms
min    = 9.62 ms
max    = 40.80 ms
Transformers (mtl/trampoline)
mean   = 99.94 ms
stddev = 22.32 ms
min    = 85.75 ms
max    = 279.82 ms
Transformers (mtl/identity)
mean   = 36.07 ms
stddev = 9.28 ms
min    = 20.93 ms
max    = 68.23 ms
Run (free)
mean   = 31.56 ms
stddev = 41.24 ms
min    = 20.69 ms
max    = 304.61 ms
```

Everything is significantly faster except for Run, which looks like it's about the same.

natefaubion:
I wouldn't expect an improvement for `Run`. It's benchmark is not going to _allocate_ dictionaries, since they are totally monomorphic. The MTL dictionaries are allocated at every bind, however, so you get a drastic speedup.

nwolverson:
I tried out this change with the purerl backend, together with some caching of typeclass dictionary creation there was a significant performance benefit.

JordanMartinez:
The Windows CI build failed with this error:
```
  purescript    > Failures:
  purescript    > 
  purescript    >   tests\TestCompiler.hs:201:7: 
  purescript    >   1) compiler, Optimization examples, 'Monad.purs' should compile to expected output
  purescript    >        Test output differed from 'tests\purs\optimize\Monad.out.js'; got:
  purescript    >        "use strict";
  purescript    >        var Control_Applicative = require("../Control.Applicative/index.js");
  purescript    >        var Control_Bind = require("../Control.Bind/index.js");
  purescript    >        var Monad = function (Applicative0, Bind1) {
  purescript    >            this.Applicative0 = Applicative0;
  purescript    >            this.Bind1 = Bind1;
  purescript    >        };
  purescript    >        var liftM1 = function (dictMonad) {
  purescript    >            var bind = Control_Bind.bind(dictMonad.Bind1());
  purescript    >            var pure = Control_Applicative.pure(dictMonad.Applicative0());
  purescript    >            return function (f) {
  purescript    >                return function (a) {
  purescript    >                    return bind(a)(function (a$prime) {
  purescript    >                        return pure(f(a$prime));
  purescript    >                    });
  purescript    >                };
  purescript    >            };
  purescript    >        };
  purescript    >        var ap = function (dictMonad) {
  purescript    >            var bind = Control_Bind.bind(dictMonad.Bind1());
  purescript    >            var pure = Control_Applicative.pure(dictMonad.Applicative0());
  purescript    >            return function (f) {
  purescript    >                return function (a) {
  purescript    >                    return bind(f)(function (f$prime) {
  purescript    >                        return bind(a)(function (a$prime) {
  purescript    >                            return pure(f$prime(a$prime));
  purescript    >                        });
  purescript    >                    });
  purescript    >                };
  purescript    >            };
  purescript    >        };
  purescript    >        module.exports = {
  purescript    >            Monad: Monad,
  purescript    >            liftM1: liftM1,
  purescript    >            ap: ap
  purescript    >        };
  purescript    > 
  purescript    >   To rerun use: --match "/compiler/Optimization examples/'Monad.purs' should compile to expected output/"
  purescript    > 
  purescript    >   tests\TestCompiler.hs:201:7: 
  purescript    >   2) compiler, Optimization examples, 'Symbols.purs' should compile to expected output
  purescript    >        Test output differed from 'tests\purs\optimize\Symbols.out.js'; got:
  purescript    >        "use strict";
  purescript    >        var Data_Symbol = require("../Data.Symbol/index.js");
  purescript    >        var Record_Unsafe = require("../Record.Unsafe/index.js");
  purescript    >        var Type_Proxy = require("../Type.Proxy/index.js");
  purescript    >        var get = function (dictIsSymbol) {
  purescript    >            var reflectSymbol = Data_Symbol.reflectSymbol(dictIsSymbol);
  purescript    >            return function (dictCons) {
  purescript    >                return function (l) {
  purescript    >                    return function (r) {
  purescript    >                        return Record_Unsafe.unsafeGet(reflectSymbol(l))(r);
  purescript    >                    };
  purescript    >                };
  purescript    >            };
  purescript    >        };
  purescript    >        var get1 = get(new Data_Symbol.IsSymbol(function () {
  purescript    >            return "foo";
  purescript    >        }))();
  purescript    >        var foo = Type_Proxy["Proxy"].value;
  purescript    >        var g = function (n) {
  purescript    >            return get1(foo)({
  purescript    >                foo: n,
  purescript    >                bar: 42
  purescript    >            });
  purescript    >        };
  purescript    >        var f = function (n) {
  purescript    >            return get1(foo)({
  purescript    >                foo: n
  purescript    >            });
  purescript    >        };
  purescript    >        module.exports = {
  purescript    >            get: get,
  purescript    >            foo: foo,
  purescript    >            f: f,
  purescript    >            g: g
  purescript    >        };
  purescript    > 
  purescript    >   To rerun use: --match "/compiler/Optimization examples/'Symbols.purs' should compile to expected output/"
  purescript    > 
  purescript    > Randomized with seed 881715853
  purescript    > 
  ```

**Comment by rhendric** [comment]
Yeah, I'm trying to figure out why. My current guess is line endings?

**Comment by rhendric** [comment]
I've updated this PR to take advantage of the renamer improvements in #4096; the CSE pass now floats definitions all the way to the outermost scope, when possible, which improves sharing and decreases the number of IIFEs.

I've also finally added a few tests that verify that the optimization happens by comparing compiler output to golden files. Other optimizations can also be tested in this suite, in the future.

**Comment by JordanMartinez** [comment]
Nice!

File: tests/purs/.gitattributes [hunk 1]
```diff
@@ -0,0 +1 @@
+*.out.js -text
```
rhendric, line 1: This, for anyone who's curious, is my solution to the `tests/purs/optimize` tests failing on Windows. Git has some sort of heuristic for determining whether a file is text—I suspect it's based on the extension—and if it believes a file is text, by default (at least on the GitHub runners?) it will normalize the newlines in that file to the operating system's default when the file is checked out, and map them back on commit. (Obviously, this breaks tests if it happens to golden test files.) `.gitattributes` files can be used to tell Git, among other things, that certain files are not text and therefore shouldn't have this behavior.

There are other approaches—we could use `git config` to disable newline conversion globally on GitHub runners, for example, but that wouldn't fix things for developers on Windows. I'm refreshing some rather old memories about all this now so I don't know if this is the best possible solution, but it seems pretty good.

kritzcreek: We're doing the same thing for the golden layout tests here: https://github.com/purescript/purescript/blob/fcf0acbcf11ad113820d32ce9ee82365a6015c9c/tests/purs/layout/.gitattributes#L1

(Maybe consider also adding -merge here)

rhendric: Oh thanks, that makes me feel better about this. Must not be extension-based then; probably it's sniffing out non-printable characters or something.

I don't think -merge is needed here; I think these files will merge pretty cleanly.

kritzcreek: The `-merge` is to make sure Git doesn't automatically merge when pulling. Golden files always need to be accepted wholesale, trying to "merge" them by combining lines from multiple PR's is basically always wrong.

rhendric: Isn't the important question whether the changes made to the files have good locality? If branch A changes the way imports are encoded in JS, and branch B changes the way that TCO loops get compiled, both of those branches change the compiled JS output but in different parts of the file, and Git's merge algorithm will probably successfully produce the golden file which should result from the combination of the features, right?

JordanMartinez:
Just FYI. There are 3 merge conflicts here. The first two are obvious, but the third one touches code I'm not familiar with. So I didn't resolve the conflict here.

rhendric:
Thanks; sorted!

f-f (approved):
I'm no expert in this area of the compiler, but I looked through this a couple of times and this looks good to me! :clap:

JordanMartinez:
I propose we release the next PS release (`v0.14.4`) and then merge this. That would give us another 6 weeks to test this out on larger codebases (for those who want to do it) and fix any bugs that this PR might have caused. Thoughts?

rhendric:
I'm game for that; let me take a last sanity pass over this today and if I don't change anything of substance, we'll merge? I certainly wouldn't mind getting another review from someone else though.

nwolverson:
```
newtype A = A { x :: B }
derive newtype instance Show A
data B = B C
       | Z
derive instance Generic B _ 
instance Show B where show x = genericShow x
newtype C = C A
derive instance Generic C _
instance Show C where show = genericShow
```
Adding a record as the newtype body and we're in the same place again, I see cycle errors on 61e2b7e2 but not on 0.14.4

rhendric:
Hmm, that's a thornier case, and decent evidence that I should rethink how I'm approaching this aspect of the problem.

erykciepiela (approved):

erykciepiela:
Sorry, I mistakenly approved this merge request - please ignore ;)

rhendric:
Work on this is paused pending a resolution to #4179.

rhendric:
> Hmm, that's a thornier case, and decent evidence that I should rethink how I'm approaching this aspect of the problem.

Update on this thread: This PR has (passing) tests for both of the cases @nwolverson submitted. Having gotten #4283 in first meant that I no longer need to change anything in this PR about how binding groups work; I'm pretty sure it was my previously-necessary tinkering with the binding groups desugaring that caused this class of issues. So I *think* this thread is addressed—although I don't know if @nwolverson will be able to confirm until the purerl backend is compliant with #4283.

purescript.cabal
```diff
@@ -144,6 +145,7 @@ common defaults
     Glob >=0.10.1 && <0.11,
     haskeline >=0.8.2 && <0.9,
     language-javascript ==0.7.0.0,
+    lens >=4.19.2 && <4.20,
```
JordanMartinez, line 148: This is tangential to the PR, but I thought I'd still bring it up. We already depend on `microlens`. By adding yet another `lens` library as a dependency, it makes the binary larger. I'm assuming that it's being added here because `microlens` doesn't suffice. And if that's the case, then could we refactor the codebase to remove `microlens` so that we only depend on one `lens` library?

src/Language/PureScript/CoreFn/CSE.hs
```diff
@@ -0,0 +1,425 @@
...
+-- |
+-- A PluralityMap is like a weaker multiset: like a multiset, it holds a
+-- number of values, but instead of keeping track of their exact counts, it
+-- only records whether there is one (False) or more than one (True).
+--
+newtype PluralityMap k = PluralityMap { getPluralityMap :: M.Map k Bool }
+
+instance Ord k => Semigroup (PluralityMap k) where
+  PluralityMap l <> PluralityMap r =
+    PluralityMap $ M.mapWithKey (\k -> (|| k `M.member` r)) l `M.union` r
```
JordanMartinez, line 106: Just to clarify, is this
```hs
PluralityMap $ (M.mapWithKey (\k -> (|| k `M.member` r)) l) `M.union` r
```
or this?
```
PluralityMap $ M.mapWithKey (\k -> (|| k `M.member` r)) (l `M.union` r)
```
I believe it's the second one.

purescript.cabal
rhendric: `monoidal-containers` (which I added in #4283 before realizing this fact) already depends on `lens`; if `lens` and its dependencies (most of which we were already pulling in even before `monoidal-containers`) result in too much binary bloat, I'll need to inline the parts of `monoidal-containers` we're using. I don't actually know how much of a difference in binary size we're talking about here, with all the optimizations on; do you?

We can absolutely replace the various `microlens` usages with `lens`—I've done so in a local branch—but I figured I'd submit that in a separate PR if we merge this deciding we want to keep `lens`.

src/Language/PureScript/CoreFn/CSE.hs
rhendric: I think it's the first, actually. Prefix function application has higher precedence than all operators, including backticked identifiers.

purescript.cabal
JordanMartinez: My comment comes mainly from reading a blogpost a couple of years ago pointing out an issue like this: a project depends on a small lens library but transitively depends on all the other lens libraries such that the author just wondered why one didn't just depend on `lens` and call it a day. (Or something like that).

Could you clarify for me what the difference between `microlens` and `lens` is (if any) besides `microlens` just being smaller? For example, do they use the same underlying encoding for optics?

src/Language/PureScript/CoreFn/CSE.hs
JordanMartinez: :smile: Maybe we should add parenthesis here just to ensure clarity? Though I don't think there's any difference between the two.

PureFunctor (changes requested):
It'd be nice to have tests for recursive computation in type classes, like so:
```purs
module Main where

import Prelude

import Prim.Row as R
import Prim.RowList as RL
import Type.Prelude (class IsSymbol, Proxy(..), reflectSymbol)

class FindKeysAux :: forall k. RL.RowList k -> Constraint
class FindKeysAux a where
  findKeysAux :: Proxy a -> Array String

instance FindKeysAux RL.Nil where
  findKeysAux _ = []

else instance (IsSymbol l, FindKeysAux r) => FindKeysAux (RL.Cons l t r) where
  findKeysAux _ = [ reflectSymbol (Proxy :: Proxy l) ] <> findKeysAux (Proxy :: Proxy r)

findKeys :: forall r rl. RL.RowToList r rl => FindKeysAux rl => Proxy r -> Array String
findKeys _ = findKeysAux (Proxy :: Proxy rl)

findKeys1 = findKeys (Proxy :: Proxy (a :: Unit))
findKeys2 = findKeys (Proxy :: Proxy (a :: Unit, b :: Unit))
findKeys3 = findKeys (Proxy :: Proxy (a :: Unit, b :: Unit, c :: Unit))
findKeys4 = findKeys (Proxy :: Proxy (a :: Unit, b :: Unit, c :: Unit, d :: Unit))
findKeys5 = findKeys (Proxy :: Proxy (a :: Unit, b :: Unit, c :: Unit, d :: Unit, e :: Unit))
```

src/Language/PureScript/CoreFn/CSE.hs
rhendric: Yeah, the only difference would be performance, if any. I think I'd have to fight with hlint about those parentheses but I'll try and see.

purescript.cabal
rhendric: My understanding, based on little more than a skimming of `microlens`'s README and some unattributable background impressions, is that the implementation of `microlens` is more or less directly lifted from `lens`; the main if not only difference is how much API and instances are provided and how many dependencies are taken.

src/Language/PureScript/CoreFn/CSE.hs
rhendric: (Oh, I shouldn't have tried to comment on this while going to bed. There is an important difference between the two: if `l = PluralityMap M.empty` and `r = PluralityMap $ M.singleton 0 False`, then in the first, `l <> r == PluralityMap $ M.singleton 0 False`, but in the second, `l <> r == PluralityMap $ M.singleton 0 True`.)

PureFunctor (approved):

rhendric:
I'm on board with merging. I figure getting a prerelease of this out early in the release cycle is probably our best strategy for getting it tested by more people before it shows up in a release.

rhendric:
Rad, I'll write that up.

nwolverson:
I've updated purerl to work with the latest, and I'm happy to say I see no problems in our projects - and it enables a signifcant performance improvement. Happy to see this merged!

PR #3916, opened 2020-08-05
Link to book was pointing to an outdated version

s-u-m-e:


kritzcreek:
I'm assuming we don't need to update `CONTRIBUTORS.md` for a link change in the README?

hdgarrood:
Yeah, generally I like to err on the side of adding people where it’s not entirely clear, but in this case I think it’s fine not to.

kritzcreek:
Thanks!

PR #3920, opened 2020-08-24
Link to releases page

milesfrain:
https://github.com/purescript/documentation/pull/323 removes some content from the getting started guide in favor of linking here. This PR preserves content that would otherwise be lost.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,9 +15,9 @@ See also <https://www.haskell.org/ghc/download_ghc_8_6_4.html> for more details 
 
 ## Official prebuilt binaries
 
-Each release comes with prebuilt x86-64 binary bundles for Linux, mac OS, and Windows. Users of other operating systems or architectures will likely need to build the compiler from source; see below.
+Each [release](https://github.com/purescript/purescript/releases) comes with prebuilt x86-64 binary bundles for Linux, mac OS, and Windows. Users of other operating systems or architectures will likely need to build the compiler from source; see below.
 
-To install a binary bundle, simply extract it and place the `purs` executable somewhere on your PATH.
+To install a binary bundle, simply extract it and place the `purs` executable somewhere on your PATH. 
 
 ## Other distributions
 
@@ -26,12 +26,15 @@ There are several other distributions of the PureScript compiler available, whic
 * NPM: `npm install -g purescript`
 * Homebrew (for macOS): `brew install purescript`
 * [PSVM](https://github.com/ThomasCrevoisier/psvm-js): `npm install -g psvm`
+* [Hackage](https://hackage.haskell.org/package/purescript): `cabal install purescript` or `stack install purescript`
 
 ## Compiling from source
 
 The easiest way is to use stack:
 
 ```
+$ git clone https://github.com/purescript/purescript.git purescript_compiler
+$ cd purescript_compiler
 $ stack update
 $ stack unpack purescript
 $ cd purescript-x.y.z  # (replace x.y.z with whichever version you just downloaded)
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -26,12 +26,15 @@ There are several other distributions of the PureScript compiler available, whic
 * NPM: `npm install -g purescript`
 * Homebrew (for macOS): `brew install purescript`
 * [PSVM](https://github.com/ThomasCrevoisier/psvm-js): `npm install -g psvm`
+* [Hackage](https://hackage.haskell.org/package/purescript): `cabal install purescript` or `stack install purescript`
```
kritzcreek, line 29: This will build a compiler binary with different dependencies than what we released, so this is _not_ a supported way of obtaining compiler binaries and we don't want to point people in that direction. It causes weird issues we can't reproduce.

INSTALL.md
```diff
@@ -15,9 +15,9 @@ See also <https://www.haskell.org/ghc/download_ghc_8_6_4.html> for more details
 
 ## Official prebuilt binaries
 
-Each release comes with prebuilt x86-64 binary bundles for Linux, mac OS, and Windows. Users of other operating systems or architectures will likely need to build the compiler from source; see below.
+Each [release](https://github.com/purescript/purescript/releases) comes with prebuilt x86-64 binary bundles for Linux, mac OS, and Windows. Users of other operating systems or architectures will likely need to build the compiler from source; see below.
 
-To install a binary bundle, simply extract it and place the `purs` executable somewhere on your PATH.
+To install a binary bundle, simply extract it and place the `purs` executable somewhere on your PATH.
```
kritzcreek, line 20:
```suggestion
To install a binary bundle, simply extract it and place the `purs` executable somewhere on your PATH.
```

INSTALL.md
```diff
@@ -26,12 +26,15 @@ There are several other distributions of the PureScript compiler available, whic
...
 * [PSVM](https://github.com/ThomasCrevoisier/psvm-js): `npm install -g psvm`
+* [Hackage](https://hackage.haskell.org/package/purescript): `cabal install purescript` or `stack install purescript`
 
 ## Compiling from source
 
 The easiest way is to use stack:
 
 ```
+$ git clone https://github.com/purescript/purescript.git purescript_compiler
+$ cd purescript_compiler
```
kritzcreek, line 37: `stack unpack` is the thing that downloads the sources here, no need to also clone the repo.

kritzcreek (changes requested):

INSTALL.md
hdgarrood: Getting the compiler from Hackage is just compiling from source so I’d rather not add it to this list anyway, as compiling from source is covered separately.

INSTALL.md
```diff
@@ -15,9 +15,9 @@ See also <https://www.haskell.org/ghc/download_ghc_8_6_4.html> for more details
 
 ## Official prebuilt binaries
 
-Each release comes with prebuilt x86-64 binary bundles for Linux, mac OS, and Windows. Users of other operating systems or architectures will likely need to build the compiler from source; see below.
+Each [release](https://github.com/purescript/purescript/releases) comes with prebuilt x86-64 binary bundles for Linux, mac OS, and Windows. Users of other operating systems or architectures will likely need to build the compiler from source; see below.
```
hdgarrood, line 18: Adding a link to the releases page is a good idea 👍

hdgarrood (approved):

kritzcreek (approved):

thomashoneyman (approved):

PR #3927, opened 2020-09-09
Suggest to export newtype constructors when the unwrapping rule fails

kl0tl:
Given `newtype N a = N a`, the newtype constructor should be exported for `Coercible N a` and `Coercible a N` constraints to be solvables. This pull request adds a specific error for when we fail to find a newtype constructor in the environment rather than the unhelpful `NoInstanceFound` error currently thrown.

The `CannotFindCoercedConstructor` error is thrown for both data and newtype constructors because I didn’t find how to differentiate them when they aren’t exported, did I miss something?

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -186,7 +186,6 @@ data SimpleErrorMessage
   | UnsupportedRoleDeclaration
   | RoleDeclarationArityMismatch (ProperName 'TypeName) Int Int
   | DuplicateRoleDeclaration (ProperName 'TypeName)
-  | MissingConstructorImportForCoercible (Qualified (ProperName 'ConstructorName))
   deriving (Show)
 
 data ErrorMessage = ErrorMessage
@@ -349,7 +348,6 @@ errorCode em = case unwrapErrorMessage em of
   UnsupportedRoleDeclaration {} -> "UnsupportedRoleDeclaration"
   RoleDeclarationArityMismatch {} -> "RoleDeclarationArityMismatch"
   DuplicateRoleDeclaration {} -> "DuplicateRoleDeclaration"
-  MissingConstructorImportForCoercible {} -> "MissingConstructorImportForCoercible"
 
 -- | A stack trace for an error
 newtype MultipleErrors = MultipleErrors
@@ -1350,12 +1348,6 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
     renderSimpleErrorMessage (DuplicateRoleDeclaration name) =
       line $ "Duplicate role declaration for " <> markCode (runProperName name) <> "."
 
-    renderSimpleErrorMessage (MissingConstructorImportForCoercible name) =
-      paras
-        [ line $ "The newtype constructor " <> markCode (showQualified runProperName name) <> " is not in scope."
-        , line "It should be imported to allow coercions to and from its representation."
-        ]
-
     renderHint :: ErrorMessageHint -> Box.Box -> Box.Box
     renderHint (ErrorUnifyingTypes t1@RCons{} t2@RCons{}) detail =
       let (row1Box, row2Box) = printRows t1 t2
@@ -1507,6 +1499,11 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
                 , Box.vcat Box.left (map (typeAtomAsBox prettyDepth) ts)
                 ]
             ]
+    renderHint (MissingConstructorImportForCoercible name) detail =
+      paras
+        [ detail
+        , Box.moveUp 1 $ Box.moveRight 2 $ line $ "Solving this instance requires the newtype constructor " <> markCode (showQualified runProperName name) <> " to be in scope."
+        ]
     renderHint (PositionedError srcSpan) detail =
       paras [ line $ "at " <> displaySourceSpan relPath (NEL.head srcSpan)
             , detail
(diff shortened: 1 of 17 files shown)


hdgarrood:
This isn't related to this PR, but what happens if a constructor is imported and only used to permit a coercion? Do we still emit an unused import warning?

kl0tl:
Newtype unwrapping is possible as long as the newtype constructor is exported, unlike Haskell it doesn‘t have to be in scope, so there’s no need to import a constructor only to allow some coercions.

hdgarrood:
I'm not sure that approach is safe: it's quite common to define a newtype in a `MyModule.Internal` module where everything is exported, and then re-export the subset of that API which is considered public from a separate module `MyModule`.

kl0tl:
I’ve rebased this pull request on top of https://github.com/purescript/purescript/pull/3937.

There’s now a `CannotFindCoercedCtor` error for when the constructor isn’t exported at all (so we don’t even know if it’s a newtype) and a `CannotFindCoercedNewtypeCtor` error for when the newtype constructor is exported but not imported.

hdgarrood:
@kl0tl do you think we should try to get this into 0.14.0? I think it would be good if we can. Would you mind resolving the conflicts when you get a chance?

kl0tl:
I’ve rebased the pull request! The diff should be much nicer. I also added a DataDeclType field to the DataType constructor of TypeKind, so that a newtype constructor doesn’t have to be exported for us to know that it is a newtype. This allows us to get rid of the CannotFindCoercedCtor error and only throw CannotUnwrapHiddenNewtypeConstructor errors when we know that the failure is really due to an out of scope newtype constructor.

kl0tl:
What do you think of CannotUnwrapHiddenNewtypeConstructor for the error? The constructor isn’t necessarily hidden after all, it might just not be imported. Perhaps CannotUnwrapOutOfScopeNewtypeConstructor would be more appropriate?

src/Language/PureScript/Errors.hs
```diff
@@ -348,6 +349,7 @@ errorCode em = case unwrapErrorMessage em of
   UnsupportedRoleDeclaration {} -> "UnsupportedRoleDeclaration"
   RoleDeclarationArityMismatch {} -> "RoleDeclarationArityMismatch"
   DuplicateRoleDeclaration {} -> "DuplicateRoleDeclaration"
+  CannotUnwrapHiddenNewtypeConstructor {} -> "CannotUnwrapHiddenNewtypeConstructor"
```
rhendric, line 352: Since you explicitly invited bikeshedding, how about `MissingConstructorImportForCoercible`? It's an issue for both wrapping and unwrapping, and the error names seem to prefer describing a thing to be corrected over a thing the compiler failed to do.

src/Language/PureScript/Errors.hs
```diff
@@ -1348,6 +1350,9 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
     renderSimpleErrorMessage (DuplicateRoleDeclaration name) =
       line $ "Duplicate role declaration for " <> markCode (runProperName name) <> "."
 
+    renderSimpleErrorMessage (CannotUnwrapHiddenNewtypeConstructor name) =
+      line $ "The newtype constructor " <> markCode (showQualified runProperName name) <> " is not in scope, it should be imported to allow coercions to and from its representation."
```
rhendric, line 1354: For consistency with the other error messages, I think there should be a sentence break and line break between ‘... is not in scope’ and ‘it should be imported...’.

src/Language/PureScript/Errors.hs
kl0tl: Yeah that’s better, thank you for the suggestion!

src/Language/PureScript/Errors.hs
kl0tl: I turned “It should be imported to allow coercions to and from its representation” into its own sentence, on a new line.

thomashoneyman (approved):
👍 from me, though this deserves a look from someone more familiar with the code as well. Thanks!

hdgarrood:
I actually feel like this should be a NoInstanceFound error rather than a separate code; the error is that the compiler can’t prove that it’s safe to coerce between the two types in question, which in PureScript is represented by the lack of a Coercible instance. I think it could be confusing that this error looks quite different from other NoInstanceFound errors which arise as a result of trying to coerce two types which aren’t Coercible.

I also have a slight worry with the wording of the error message, because to me the message sounds quite confident that the correct course of action is to import the constructor. In practice, if you receive this error, the answer may instead be “don’t do that”. For example, if you were trying to coerce an Array to a NonEmptyArray. I think a reminder that having the newtype constructor in scope will silence this error is warranted, but it should be secondary to the main piece of information, which is “I can’t prove that a coercion from type X to type Y is safe.”

kl0tl:
Would you prefer that we add the MissingConstructorImportForCoercible constructor to ErrorMessageHint instead and accumulate hints in solveWanteds, so that the following:

```purs
module Example.N where

newtype N a = N a
```

```purs
module Example where

import Example.N (N)
import Safe.Coerce (coerce)

example :: forall a. a -> N a
example = coerce
```

yields

```
Error found:
at Example.purs:7:11 - 7:17 (line 7, column 11 - line 7, column 17)

    Prim.Coerce.Coercible a0    
                          (N a0)
                                

  The newtype constructor N is not in scope.

  is at least as general as type a0 -> N a0
  has type a0 -> N a0
in value declaration example

        bound at (line 7, column 11 - line 7, column 17)

```

?

hdgarrood:
That sounds more appealing to me, yes. I also think it would be good to make the connection between the constructor being in scope and the coercion being possible slightly more clear in the hint; perhaps something along the lines of "Performing this coercion requires the data constructor N to be in scope"?

hdgarrood (approved):

PR #3929, opened 2020-09-11
Disallow cycles in foreign data kinds

### LINKED ISSUE #3928: Internal error for mutually recursive `foreign import data`
## Description

Mutually recursive `foreign import data` declarations can cause an internal compiler error on the current master.

## To Reproduce

```
module Main where

foreign import data A :: B
foreign import data B :: A
```
```
purescript    >       'RecursiveExternData.purs' should compile and run without error: FAIL (0.07s)
purescript    >         uncaught exception: ErrorCall
purescript    >         An internal error occurred during compilation: toExternsDeclaration: no kind 
in toExternsDeclaration
purescript    >         Please report this at https://github.com/purescript/purescript/issues
purescript    >         CallStack (from HasCallStack):
purescript    >           error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-ast-0.
1.0.0-Gtmwmu4N43KCxo7Qi9BiwQ:Language.PureScript.Crash
purescript    >           internalError, called at src/Language/PureScript/Externs.hs:225:18 in pures
cript-0.13.8-8vNPQdyw1kRA1yYdlrf0He:Language.PureScript.Externs
```

## Expected behavior

I'm not sure whether this should be rejected by the compiler or not, but it should at least not result in an internal compiler error.

## Additional context

Came across this while trying to implement the refactoring described in https://github.com/purescript/purescript/pull/3910#issuecomment-680252963 - it seems like ExternDataDeclaration values are not designed to be part of a DataBindingGroup.

## PureScript version

master

natefaubion:
Fixes #3928

### DIFF
--- a/src/Language/PureScript/Sugar/BindingGroups.hs
+++ b/src/Language/PureScript/Sugar/BindingGroups.hs
@@ -67,12 +67,12 @@ createBindingGroups moduleName = mapM f <=< handleDecls
   handleDecls :: [Declaration] -> m [Declaration]
   handleDecls ds = do
     let values = mapMaybe (fmap (fmap extractGuardedExpr) . getValueDeclaration) ds
-        kindDecls = fmap (,True) $ filter isKindDecl ds
-        dataDecls = fmap (,False) $ filter (\a -> isDataDecl a || isExternDataDecl a || isTypeSynonymDecl a || isTypeClassDecl a) ds
+        kindDecls = fmap (,True) $ filter (\a -> isKindDecl a || isExternDataDecl a) ds
+        dataDecls = fmap (,False) $ filter (\a -> isDataDecl a || isTypeSynonymDecl a || isTypeClassDecl a) ds
         kindSigs = fmap (declTypeName . fst) kindDecls
         typeSyns = fmap declTypeName $ filter isTypeSynonymDecl ds
-        allProperNames = fmap (declTypeName . fst) dataDecls
         allDecls = kindDecls ++ dataDecls
+        allProperNames = fmap (declTypeName . fst) allDecls
         mkVert (d, isSig) =
           let names = usedTypeNames moduleName d `intersect` allProperNames
               name = declTypeName d
@@ -226,6 +226,7 @@ toDataBindingGroup (CyclicSCC ds')
   | otherwise = return . DataBindingGroupDeclaration $ NEL.fromList ds'
   where
   kindDecl (KindDeclaration sa _ pn _) = [(fst sa, Qualified Nothing pn)]
+  kindDecl (ExternDataDeclaration sa pn _) = [(fst sa, Qualified Nothing pn)]
   kindDecl _ = []
 
 isTypeSynonym :: Declaration -> Maybe (ProperName 'TypeName)
--- a/src/Language/PureScript/Sugar/BindingGroups.hs
+++ b/src/Language/PureScript/Sugar/BindingGroups.hs
@@ -79,7 +79,7 @@ createBindingGroups moduleName = mapM f <=< handleDecls
               -- If a dependency has a kind signature, than that's all we need to depend on, except
               -- in the case that we are defining a kind signature and using a type synonym. In order
               -- to expand the type synonym, we must depend on the synonym declaration itself.
-              deps = fmap (\n -> (n, n `elem` kindSigs && (isSig && not (n `elem` typeSyns)))) names
+              deps = fmap (\n -> (n, n `elem` kindSigs && (not isSig || not (n `elem` typeSyns)))) names
               self | not isSig && name `elem` kindSigs = [(name, True)]
                    | otherwise = []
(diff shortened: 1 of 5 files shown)


src/Language/PureScript/Sugar/BindingGroups.hs
```diff
@@ -67,12 +67,12 @@ createBindingGroups moduleName = mapM f <=< handleDecls
   handleDecls :: [Declaration] -> m [Declaration]
   handleDecls ds = do
     let values = mapMaybe (fmap (fmap extractGuardedExpr) . getValueDeclaration) ds
-        kindDecls = fmap (,True) $ filter isKindDecl ds
-        dataDecls = fmap (,False) $ filter (\a -> isDataDecl a || isExternDataDecl a || isTypeSynonymDecl a || isTypeClassDecl a) ds
+        kindDecls = fmap (,True) $ filter (\a -> isKindDecl a || isExternDataDecl a) ds
```
hdgarrood, line 70: `foreign import data` declarations are a bit awkward because they are both a data declaration and a kind declaration at the same time; is it safe to remove them from `dataDecls`? I wonder if in some cases they need to be included in the `dataBindingGroupDecls`. For example, on e56ebda3
```
module Main where

foreign import data A :: Type
data B (x :: A)
```
currently results in
```
at src/Main.purs:4:14 - 4:15 (line 4, column 14 - line 4, column 15)

  Unknown type Main.A

while inferring the kind of A
while checking that type A
  has kind Type
in type constructor B
```

natefaubion: Good point! I'll add this test.

natefaubion: This happens regardless, so I think there's something else going on here.

natefaubion: This is fixed, and it was an issue with the boolean signature check. Boolean logic is not not hard.

tests/purs/failing/RowsInKinds.out
```diff
@@ -5,7 +5,7 @@ at tests/purs/failing/RowsInKinds.purs:14:16 - 14:17 (line 14, column 16 - line
   Could not match kind
   [33m             [0m
   [33m  ( z :: Type[0m
-  [33m  | t25      [0m
+  [33m  | t24      [0m
```
natefaubion, line 8: Not sure what this is about, but I suspect it just resulted in a different sort.

hdgarrood (approved):

src/Language/PureScript/Sugar/BindingGroups.hs
hdgarrood: Can you explain why this was manifesting as an unknown name error? Was the declaration being dropped by this designating step or something?

natefaubion: Sure. Dependencies have a name and a boolean flag, which is whether or not it's the kind signature or not. As the comment states, this is somewhat tricky. When something has a kind signature, in _general_ we should depend on only that for kind checking. However, if we are _defining_ a kind signature, and the dependency is a type synonym, we need to depend on the actual type synonym declaration so it can be expanded (gross, I know). Foreign data kinds are really kind declarations. Or perhaps, a foreign data declaration should _really_ be:

```
foreign data Foo :: Type -> Type
foreign data Foo
```

It was failing because the predicate in `mkVert` was asking for a dependency on `A` with `False` instead of `True`, which didn't exist. That's because `isSig` for `B` was `False`. That means the declaration for `B` was depending on the declaration for `A`, which doesn't really exist since we are counting foreign data as a kind signature (`True`). The toposort was then putting `B` before `A` because of implementation details, and so `A` was not bound when it checked `B`.

If there is another way to factor this predicate PLEASE TELL ME because it hurts my head.

hdgarrood: Ah right, thanks! I had a decent idea of the logic behind when we want to depend on a data declaration itself versus just depending on its kind signature from the comment already (thanks for that, by the way), but I didn't realise that `mkVert` was creating a dependency on a nonexistent vertex. As for refactoring, perhaps fewer booleans might help? What if a vertex id was `(ProperName 'TypeName, VertexType)` rather than `(ProperName 'TypeName, Bool)`, where
```
data VertexType
  = VertexDefinition
  | VertexKindSignature
```
Alternatively you could maybe refactor via De Morgan's law?
```
not isSig || not (n `elem` typeSyns) = not (isSig && n `elem` typeSyns)
```
not sure if that actually makes things any clearer, though.

natefaubion: Alright, I've taken out the bool and made it at least reflect the comment better.

hdgarrood: Looks great!

PR #3934, opened 2020-09-16
Bump version for 0.14.0-rc2

hdgarrood:
The rc1 I already tagged is no good, because I forgot to update the
version. That is particularly important for this release, because tools
like psa need to be able to know what version of the CLI they have
encountered in order to know whether to read from stdout or stderr if
they want to maintain compatibility with both 0.13.x and 0.14.x.

### DIFF
--- a/app/Version.hs
+++ b/app/Version.hs
@@ -12,8 +12,15 @@ import Paths_purescript as Paths
 import qualified Development.GitRev as GitRev
 #endif
 
+-- Unfortunately, Cabal doesn't support prerelease identifiers on versions. To
+-- avoid misleading users who run `purs --version`, we manually add the
+-- prerelease identifier here (if any). When releasing a proper version, simply
+-- set this to an empty string.
+prerelease :: String
+prerelease = "-rc2"
+
 versionString :: String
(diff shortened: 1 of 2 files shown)


app/Version.hs
```diff
@@ -12,8 +12,15 @@ import Paths_purescript as Paths
 import qualified Development.GitRev as GitRev
 #endif
 
+-- Unfortunately, Cabal doesn't support prerelease identifiers on versions. To
+-- avoid misleading users who run `purs --version`, we manually add the
+-- prerelease identifier here (if any). When releasing a proper version, simply
+-- set this to an empty string.
+prerelease :: String
+prerelease = "-rc2"
```
natefaubion, line 20: Is the pre-release identifier necessary if we don't intend to publish it to Hackage?

natefaubion: I suppose this is useful for debugging purposes. You do not want someone thinking they have the proper release, when they still have an rc.

natefaubion (approved):

app/Version.hs
hdgarrood: Yeah, exactly

Issue #3735, opened 2019-09-23
compiler loop: cycle in type synonyms, v. 0.13.3


brubar:
With purs 0.13.3, the compiler may loop until there is no more memory.
I'm expecting it to report a CycleInTypeSynonym error instead.

Here is a small (meaningless) example to reproduce:

```PureScript
    import Prelude(unit, Unit)
    
    data D   = D S
    type S   = T D
    type T x = S
    
    f :: D -> Unit
    f (D _) = unit
```

garyb:
Thanks for the report!

hdgarrood:
I looked into this a little today. I think the problem is in `Language.PureScript.Sugar.BindingGroups.toDataBindingGroup`:

```haskell
toDataBindingGroup
  :: MonadError MultipleErrors m
  => SCC Declaration
  -> m Declaration
toDataBindingGroup (AcyclicSCC d) = return d
toDataBindingGroup (CyclicSCC [d]) = case isTypeSynonym d of
  Just pn -> throwError . errorMessage' (declSourceSpan d) $ CycleInTypeSynonym (Just pn)
  _ -> return d
toDataBindingGroup (CyclicSCC ds')
  | all (isJust . isTypeSynonym) ds' = throwError . errorMessage' (declSourceSpan (head ds')) $ CycleInTypeSynonym Nothing
  | otherwise = return . DataBindingGroupDeclaration $ NEL.fromList ds'
```

For each module, we construct a graph where the vertices are the data type and type synonym declarations, and where there is an edge from T to U if the type T mentions U in its definition (or perhaps the other way around, I can't remember). Then we split the graph up into strongly-connected components (SCC), and apply the `toDataBindingGroup` function to each SCC.

At the moment, we only produce an error if every type declaration in an SCC is a type synonym, so the given example slips through because it also has a `data` declaration in it. I think changing the logic so that any data binding group with at least two type synonyms is disallowed would catch this error, but that's probably excessive; I think it would also disallow a lot of currently-allowed cases which don't cause problems, such as:

```purescript
data A = A B
type B = C
data C = C D
type D = A
```

which is legal (if a bit useless since you can't construct any values of these types without cheating). 

My understanding is that cycles are only a problem if they only involve synonyms; if you have a type synonym `T` but the only way to construct a cycle starting at `T` is to go through at least one `data` declaration, then you're fine.

[...]

PR #3936, opened 2020-09-19
Check all recursive paths in data binding groups

### LINKED ISSUE #3918: Recursive binding group check for type synonyms fails to catch some loops in row types
## Description

You can get the compiler to accept some recursive type synonyms as long as one of the synonyms has a branch going through a newtype. Trying to construct one of these values makes the typechecker loop.

## To Reproduce

```purescript
module Loop where

type A = B
type B = { a :: A, b :: Loop }
data Loop = Loop B

boom = Loop {}
```

## Expected behavior

The compiler should reject this since there is a recursive path that does not go through a newtype.

## Additional context

The current check allows recursion as long as there is data/newtype in the group, assuming that this breaks the loop. In the case of this record/row type, one branch is broken through data, but another loops directly. We would need to further reduce the binding group to check that _all_ paths go through data/newtype.

## PureScript version

0.13.8 and master

natefaubion:
This changes the data binding group check to follow all dependency paths, looking for loops in synonyms.

Fixes #3918

### DIFF
--- a/src/Language/PureScript/Sugar/BindingGroups.hs
+++ b/src/Language/PureScript/Sugar/BindingGroups.hs
@@ -12,13 +12,13 @@ module Language.PureScript.Sugar.BindingGroups
 import Prelude.Compat
 import Protolude (ordNub)
 
-import Control.Monad ((<=<))
+import Control.Monad ((<=<), guard)
 import Control.Monad.Error.Class (MonadError(..))
 
 import Data.Graph
 import Data.List (intersect)
 import Data.Foldable (find)
-import Data.Maybe (isJust, isNothing, mapMaybe)
+import Data.Maybe (isJust, mapMaybe)
 import qualified Data.List.NonEmpty as NEL
 import qualified Data.Set as S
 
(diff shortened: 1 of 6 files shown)


garyb (approved):

kl0tl:
This also fixes https://github.com/purescript/purescript/issues/3735!

src/Language/PureScript/Sugar/BindingGroups.hs
```diff
@@ -224,21 +225,36 @@ toBindingGroup moduleName (CyclicSCC ds') = do
...
   where
   kindDecl (KindDeclaration sa _ pn _) = [(fst sa, Qualified Nothing pn)]
   kindDecl (ExternDataDeclaration sa pn _) = [(fst sa, Qualified Nothing pn)]
   kindDecl _ = []
 
+  getDecl (decl, _, _) = decl
+  getName (_, name, _) = name
+  lookupVert name = find ((==) name . getName) ds'
+
+  depsTerminate seen _ [] = Just seen
```
hdgarrood, line 247: I think this function could do with some documentation. Would this be accurate?

> This function ensures that all paths through an SCC which start and end at a given type synonym involve at least one vertex which is not a type synonym. This is necessary to prevent loops while expanding type synonyms.
> In this function:
> - `seen` is the list of names we know are safe, either because they are not type synonyms, or because we have already checked that all paths through the SCC starting and ending at that vertex involve one non-type-synonym vertex
> - `stop` is the current path of unbroken type synonym vertices being checked for cycles
>
> This function either returns Nothing in the case that a type synonym cycle was found, or Just a list of seen names in the case where no cycles were found.

Also, could we have it return a `Left stop` if we did find a cycle, and then use `stop` in the error message, in the same way that the other CycleInX error constructors take a list of names involved in the cycle?

natefaubion: I've changed the implementation to use SCC to find all cycles in group, and I've changed the error to include all the names in the cycles.

tests/purs/failing/InfiniteKind2.out
```diff
@@ -10,7 +10,7 @@ at tests/purs/failing/InfiniteKind2.purs:5:23 - 5:27 (line 5, column 23 - line 5
 while checking that type [33mTree[0m
   has kind [33mt0[0m
 while inferring the kind of [33mm Tree[0m
-in type constructor [33mTree[0m
+in data binding group Tree
```
natefaubion, line 13: I'm not sure what this change is about.

natefaubion: Oh I see. I took out the singleton synonym check. I think this is actually correct now since it is a cyclical type.

hdgarrood (approved):
Nice! I like the new stronglyConnCompR implementation, that’s very easy to follow.

PR #3937, opened 2020-09-19
Require newtype constructors to be imported for coercing them

kl0tl:
As noticed by @hdgarrood in https://github.com/purescript/purescript/pull/3927#issuecomment-693535475, merely being exported isn’t a sufficient condition for a newtype to be unwraped when solving Coercible constraints because of internal modules.

This pull request requires newtypes constructors to be imported for them to be unwrapable and ensures `UnusedDctorImport` warnings aren’t emitted for newtype constructors imported only for the sake of solving Coercible constraints.

### DIFF
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -16,10 +16,11 @@ import           Control.Monad.Error.Class (MonadError(..))
 import           Control.Monad.IO.Class
 import           Control.Monad.Supply
 import           Control.Monad.Trans.Control (MonadBaseControl(..))
-import           Control.Monad.Writer.Class (MonadWriter(..))
+import           Control.Monad.Trans.State (runStateT)
+import           Control.Monad.Writer.Class (MonadWriter(..), censor)
 import           Control.Monad.Writer.Strict (runWriterT)
 import           Data.Function (on)
-import           Data.Foldable (for_)
+import           Data.Foldable (fold, for_)
 import           Data.List (foldl', sortBy)
 import qualified Data.List.NonEmpty as NEL
 import           Data.Maybe (fromMaybe)
@@ -75,9 +76,16 @@ rebuildModule' MakeActions{..} exEnv externs m@(Module _ _ moduleName _ _) = do
   let env = foldl' (flip applyExternsFileToEnvironment) initEnvironment externs
       withPrim = importPrim m
   lint withPrim
+
   ((Module ss coms _ elaborated exps, env'), nextVar) <- runSupplyT 0 $ do
-    desugar exEnv externs [withPrim] >>= \case
-      [desugared] -> runCheck' (emptyCheckState env) $ typeCheckModule desugared
+    runStateT (desugar externs [withPrim]) (exEnv, mempty) >>= \case
+      ([desugared], (exEnv', usedImportsByModuleName)) -> do
+        (checked, CheckState{..}) <- runStateT (typeCheckModule desugared) $ emptyCheckState env
+        let usedImports = fold $ M.lookup moduleName usedImportsByModuleName
+            usedImports' = foldl' (flip $ \(fromModuleName, newtypeCtorName) ->
+              M.alter (Just . (fmap DctorName newtypeCtorName :) . fold) fromModuleName) usedImports checkCoercedNewtypeCtorsImports
+        censor (addHint (ErrorInModule moduleName)) $ lintImports checked exEnv' usedImports'
+        return (checked, checkEnv)
       _ -> internalError "desugar did not return a singleton"
 
   -- desugar case declarations *after* type- and exhaustiveness checking
--- a/src/Language/PureScript/Make.hs
+++ b/src/Language/PureScript/Make.hs
@@ -84,6 +84,9 @@ rebuildModule' MakeActions{..} exEnv externs m@(Module _ _ moduleName _ _) = do
         let usedImports = fold $ M.lookup moduleName usedImportsByModuleName
             usedImports' = foldl' (flip $ \(fromModuleName, newtypeCtorName) ->
               M.alter (Just . (fmap DctorName newtypeCtorName :) . fold) fromModuleName) usedImports checkCoercedNewtypeCtorsImports
+        -- Imports cannot be linted before type checking because we need to
+        -- known which newtype constructors are used to solve Coercible
+        -- constraints in order to not report them as unused.
         censor (addHint (ErrorInModule moduleName)) $ lintImports checked exEnv' usedImports'
         return (checked, checkEnv)
       _ -> internalError "desugar did not return a singleton"
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -25,8 +25,8 @@ import Control.Monad.Writer
 import Data.Foldable (for_, fold, foldl', toList)
 import Data.Function (on)
 import Data.Functor (($>))
-import Data.List (minimumBy, groupBy, nubBy, sortBy, zipWith4)
-import Data.Maybe (fromMaybe, mapMaybe)
+import Data.List (find, groupBy, minimumBy, nubBy, sortBy, zipWith4)
+import Data.Maybe (fromMaybe, isJust, mapMaybe)
 import qualified Data.Map as M
 import qualified Data.Set as S
 import Data.Traversable (for)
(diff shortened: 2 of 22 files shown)


rhendric:
Is the new `shouldNotWarnWith` infrastructure necessary? I think existing tests accomplish this by living in `purs/warning` and having empty `.out` files, unless I'm misunderstanding what you're trying to do.

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -75,11 +75,16 @@ data CheckState = CheckState
   -- This goes into state, rather than using 'rethrow',
   -- since this way, we can provide good error messages
   -- during instance resolution.
+  , checkDesugaringWarnings :: MultipleErrors
```
rhendric, line 78: Looks pretty hacky to me! Maybe this is the right thing to do when all things are considered, but the alternative that comes to mind is writing the set of constructors that are used by the solver, and moving import linting out of import desugaring and into its own step that runs after type checking and uses that set. Is that a bad idea for some reason I'm not seeing? Or just not worth doing?

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -66,7 +66,7 @@ data CheckState = CheckState
   -- ^ The next skolem variable
   , checkNextSkolemScope :: Int
   -- ^ The next skolem scope constant
-  , checkCurrentModule :: Maybe ModuleName
+  , checkCurrentModule :: Maybe (ModuleName, [(SourceAnn, ModuleName, ImportDeclarationType, Maybe ModuleName)])
```
rhendric, line 69: This feels like an odd place to add this new data. Why not put a list of `(ModuleName, ImportDeclarationType)` in a new field in `Environment` instead?

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -472,22 +472,70 @@ entails SolverOptions{..} constraint context hints =
...
+
+    isUnusedNewtypeCtorImportWarning
+      :: Maybe ModuleName
+      -> Qualified (ProperName 'TypeName)
+      -> ErrorMessage
+      -> Bool
+    isUnusedNewtypeCtorImportWarning currentModuleName qualifiedNewtypeName = \case
+      ErrorMessage hints' (UnusedDctorImport unusedTyModuleName unusedTyName _ _) ->
+        any isInCurrentModule hints' && qualifiedNewtypeName == mkQualified unusedTyName unusedTyModuleName
+      _ -> False
```
rhendric, line 532: What about `UnusedImport`s? Suppose you have
```purs
module N1 where

newtype N1 a = N1 a
```
```purs
module N2 where

import N1 (N1)

newtype N2 a = N2 (N1 a)
```
```purs
module Unwrap where

import Safe.Coerce (coerce)
import N1
import N2 (N2(..))

unwrap :: forall a. N2 a -> a
unwrap = coerce
```

That ought not to warn about the `import N1`, right? But I think it will.

lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
```diff
@@ -308,6 +308,15 @@ data ImportDeclarationType
   | Hiding [DeclarationRef]
   deriving (Eq, Show, Generic, Serialise)
 
+instance Semigroup ImportDeclarationType where
```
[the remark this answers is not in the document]
kl0tl: You’re right, `Implicit` should annihilate both `Explicit` and `Hiding`. I don’t think another `Semigroup` instance would be more natural for `ImportDeclarationType` but indeed, we don’t even need the instance in the first place 🙃

src/Language/PureScript/TypeChecker/Monad.hs
kl0tl: I thought I wouldn’t be able to lint imports after typechecking so easily, hence the minimal hack, but this isn’t that bad after all!

src/Language/PureScript/TypeChecker/Monad.hs
kl0tl: An `imports :: Map ModuleName (ModuleName, ImportDeclaration)` map in the environment could be useful for other things but it would incur a circular dependency between `Language.PureScript.Environment` and `Language.PureScript.AST.Declarations`.

src/Language/PureScript/TypeChecker/Entailment.hs
kl0tl: Good catch! I added a case and test.

kl0tl:
> Is the new `shouldNotWarnWith` infrastructure necessary? I think existing tests accomplish this by living in `purs/warning` and having empty `.out` files, unless I'm misunderstanding what you're trying to do.

It didn’t occur to me that empty golden tests were equivalent to checking for the absence of warnings 🤦 I removed `shouldNotWarnWith` and the associated changes.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -472,22 +473,60 @@ entails SolverOptions{..} constraint context hints =
...
+      -> Maybe (Maybe ModuleName, Qualified (ProperName 'ConstructorName), [Text], SourceType, [Ident])
+    lookupNewtypeConstructor currentModule env qualifiedNewtypeName@(Qualified newtypeModuleName newtypeName) = do
+      let fromModule = find isNewtypeCtorInScope . snd =<< currentModule
+          fromModuleName = (\(_, n, _, _) -> n) <$> fromModule
+          asModuleName = (\(_, _, _, n) -> n) =<< fromModule
+          currentModuleName = fst <$> currentModule
+      guard $ newtypeModuleName == currentModuleName || isJust fromModule
+      (_, DataType tvs [(ctorName, [wrappedTy])]) <- M.lookup qualifiedNewtypeName (types env)
+      (Newtype, _, _, ids) <- M.lookup (Qualified newtypeModuleName ctorName) (dataConstructors env)
+      pure (fromModuleName, (Qualified asModuleName ctorName), map (\(name, _, _) -> name) tvs, wrappedTy, ids)
```
rhendric, line 513: `Qualified ...` doesn't need parens here.

src/Language/PureScript/TypeChecker/Monad.hs
kl0tl: I got confused because `Language.PureScript.Sugar.desugar` accepts a list of modules and I didn’t realize we typecheck only one module at a time so I thought it would be better to keep together data specific to the current module.

It seems that since https://github.com/purescript/purescript/pull/1501/commits/6d1b5ec3af12812810701b766a12f7c9aceabe75 we also only ever desugar singleton lists of modules. Should we update `desugar` from `[Module] -> m [Module]` to `Module -> m Module`? This is straightforward to do and it would simplify a few things.

rhendric: > Should we update `desugar` from `[Module] -> m [Module]` to `Module -> m Module`?

That sounds like a good idea to me; I couldn't dig up a good reason to keep the current signature.

Probably that should be its own PR, though? I'd want to clean up `desugar` itself, `partiallyDesugar` in `L.P.Docs.Convert`, and the few `desugar` steps that aren't simply `map`s or `traverse`s—`desugarImports` has a doc comment that needs to be brought up to date too.

hdgarrood: Changing desugar so that it only desugars one module at a time also sounds good to me, and I agree that that's probably best done in its own PR.

kl0tl: I have a commit ready, I‘ll open another pull request.

rhendric (approved):

src/Language/PureScript/Make.hs
```diff
@@ -75,9 +76,16 @@ rebuildModule' MakeActions{..} exEnv externs m@(Module _ _ moduleName _ _) = do
...
   ((Module ss coms _ elaborated exps, env'), nextVar) <- runSupplyT 0 $ do
-    desugar exEnv externs [withPrim] >>= \case
-      [desugared] -> runCheck' (emptyCheckState env) $ typeCheckModule desugared
+    runStateT (desugar externs [withPrim]) (exEnv, mempty) >>= \case
+      ([desugared], (exEnv', usedImportsByModuleName)) -> do
+        (checked, CheckState{..}) <- runStateT (typeCheckModule desugared) $ emptyCheckState env
+        let usedImports = fold $ M.lookup moduleName usedImportsByModuleName
+            usedImports' = foldl' (flip $ \(fromModuleName, newtypeCtorName) ->
+              M.alter (Just . (fmap DctorName newtypeCtorName :) . fold) fromModuleName) usedImports checkCoercedNewtypeCtorsImports
+        censor (addHint (ErrorInModule moduleName)) $ lintImports checked exEnv' usedImports'
```
hdgarrood, line 87: Could you please add a comment here saying that we need to wait until after type checking to lint imports because newtype constructors may be marked as used as part of Coercible solving?

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -68,18 +69,22 @@ data CheckState = CheckState
...
+  -- ^ The current module imports
   , checkSubstitution :: Substitution
   -- ^ The current substitution
   , checkHints :: [ErrorMessageHint]
   -- ^ The current error message hint stack.
   -- This goes into state, rather than using 'rethrow',
   -- since this way, we can provide good error messages
   -- during instance resolution.
+  , checkCoercedNewtypeCtorsImports :: S.Set (ModuleName, Qualified (ProperName 'ConstructorName))
+  -- ^ Newtype constructors imports required to solve Coercible constraints
```
hdgarrood, line 82: This field is only for keeping track of newtype constructors which have been used so that we don't emit unused import warnings for them, right? If so could you please mention that here?

hdgarrood (approved):

kl0tl:
I added some comments, let me know if this is clear enough!

src/Language/PureScript/Make.hs
```diff
@@ -75,9 +76,19 @@ rebuildModule' MakeActions{..} exEnv externs m@(Module _ _ moduleName _ _) = do
...
+    runStateT (desugar externs [withPrim]) (exEnv, mempty) >>= \case
+      ([desugared], (exEnv', usedImportsByModuleName)) -> do
+        (checked, CheckState{..}) <- runStateT (typeCheckModule desugared) $ emptyCheckState env
+        let usedImports = fold $ M.lookup moduleName usedImportsByModuleName
+            usedImports' = foldl' (flip $ \(fromModuleName, newtypeCtorName) ->
+              M.alter (Just . (fmap DctorName newtypeCtorName :) . fold) fromModuleName) usedImports checkCoercedNewtypeCtorsImports
+        -- Imports cannot be linted before type checking because we need to
+        -- known which newtype constructors are used to solve Coercible
+        -- constraints in order to not report them as unused.
+        censor (addHint (ErrorInule moduleName)) $ lintImports checked exEnv' usedImports'
```
hdgarrood, line 90: typo here? I think this is causing CI to fail

hdgarrood:
Looks great :+1: please feel free to hit merge once CI is passing

src/Language/PureScript/Make.hs
kl0tl: I guess I was too tired even for adding comments properly 🤦

PR #3943, opened 2020-10-09
Update to GHC 8.8.4

TheMatten:
Bumps snapshot, upper bounds and fixes (little amount of) breakage introduced by new versions of libraries. In case of `lib/purescript-cst/tests/TestCst.hs`, I've replaced use of `fail` with `error`, because new version of `Gen` doesn't seem to provide `MonadFail` or some other error reporting functionality (it seems like old `Monad.fail` deferred to default implementation anyway).

- [x] `stack build` works
- [x] `stack run` works
- [x] `stack test` passes

(To add some motivation, I'm trying to build `purescript` with `asterius` which is currently on `8.8.4`)

### DIFF
--- a/lib/purescript-ast/package.yaml
+++ b/lib/purescript-ast/package.yaml
@@ -19,7 +19,7 @@ extra-source-files:
   - README.md
 dependencies:
   - aeson >=1.0 && <1.5
-  - base >=4.11 && <4.13
+  - base >=4.11 && <4.14
   - base-compat >=0.6.0
   - bytestring
   - containers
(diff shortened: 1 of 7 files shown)


hdgarrood:
Thanks for the PR. Just so you know, this is unlikely to get looked at until at least v0.14.0 is out and #3933 is merged - I'd really like to get #3933 in before we start fiddling with version bounds, because otherwise it will need redoing.

mitchblank:
Any updates now that 0.14.0 is out?

hdgarrood:
I still want to get #3933 merged first.

waynevanson:
She's merged! Looking forward to see this come to fruition.

hdgarrood:
We are probably going to merge #4013 instead, as it happens.

hdgarrood:
I’m going to close this as we are on 8.10.4 as of merging #4013. Thanks anyway for the PR!

PR #3944, opened 2020-10-10
Update the desugaring pipeline to work on individual modules

kl0tl:
Since 6d1b5ec3af12812810701b766a12f7c9aceabe75, `Language.PureScript.Sugar.desugar` is always applied to singleton arrays so this simplifies a few things.


rhendric (approved):
Clearly correct, clearly an improvement. :shipit:

hdgarrood (approved):

PR #3950, opened 2020-10-23
Warn and rethrow with position when deriving instances

### LINKED ISSUE #3949: PartiallyAppliedSynonym error without source span
## Description

Managed to track down a case of `PartiallyAppliedSynonym` occurring which is reported without source span (possibly related to the non-reproducible case remaining when #3208 was closed out.

## To Reproduce

The following module 
```purescript
module TypeSynLocation where

type Foo f = f Int
newtype NewT = NewT Foo
derive instance newtypeProcessor :: Newtype NewT _
```
gives an error not containing source span:
```
Error found:

  Type synonym TypeSynLocation.Foo is partially applied.
```
Removing the derived `Newtype` instance, the error is reported at the usage of `Foo` as would be expected.

## Expected behavior

Source span should be reported when Newtype instance is not derived.

## PureScript version

0.13.8

kl0tl:
Fix #3949.

### DIFF
--- a/tests/purs/failing/InvalidDerivedInstance3.out
+++ b/tests/purs/failing/InvalidDerivedInstance3.out
@@ -1,0 +1,10 @@
+Error found:
+at tests/purs/failing/InvalidDerivedInstance3.purs:10:1 - 10:40 (line 10, column 1 - line 10, column 40)
+
+  Type synonym [33mMain.S[0m is partially applied.
+  Type synonyms must be applied to all of their type arguments.
+
+
+See https://github.com/purescript/documentation/blob/master/errors/PartiallyAppliedSynonym.md for more information,
+or to contribute content related to this error.
+
--- a/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
+++ b/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs
@@ -16,6 +16,7 @@ import           Data.Maybe (fromMaybe, mapMaybe)
 import           Data.Ord (comparing)
 import qualified Data.Set as S
 import           Data.Text (Text)
+import           Data.Traversable (for)
 import           Language.PureScript.AST
 import qualified Language.PureScript.Constants.Data.Generic.Rep as DataGenericRep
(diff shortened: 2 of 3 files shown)


tests/purs/failing/InvalidDerivedInstance3.out
```diff
@@ -0,0 +1,10 @@
+Error found:
+at tests/purs/failing/InvalidDerivedInstance3.purs:10:1 - 10:40 (line 10, column 1 - line 10, column 40)
```
rhendric, line 2: Shouldn't this be reported at line 8?

natefaubion: Because derived instances are currently desugared before type checking, unfortunately no.

rhendric: @natefaubion Sorry, I don't understand why that's relevant. When `deriveNewtype` finds the type declaration for the newtype, doesn't that have source annotations on it? Can't any errors raised in `replaceAllTypeSynonymsM` be hinted with the position of that type declaration? Or maybe `replaceAllTypeSynonyms'` should do the hinting internally based on the (currently ignored) source annotation of the type it's replacing? I don't see how any of that would depend on type checking happening first.

natefaubion: I don't think expanded type synonyms should retain the ranges of the defining declaration. I could be wrong, but I think the issue is happening because it's inlining the type and then expanding it. So its treating the instance head like it was written as:
```
instance newtypeN :: Newtype N S
```

It then tries to expand `S`, which throws, which is happening in the instance declaration. If derived instances were done as part of typechecking (which is actually necessary anyway due to PolyKinds), then this would just be hit when the `data N` declaration is kind checked, which would happen before this instance.

rhendric: Maybe the expansion doesn't have ranges, but the input type to be expanded comes directly from the newtype declaration so the ranges are present in that—unless there's some case I'm missing where that wouldn't be true? Take a look at #3951; it seems to work out fine, at least for the test added here.

PR #3951, opened 2020-10-24
Add source span to PartiallyAppliedSynonym errors

### LINKED ISSUE #3949: PartiallyAppliedSynonym error without source span

rhendric:
An alternative to #3950; fixes #3949.

Advantages: it's a smaller change, and the error message location is more correct.

Disadvantages: ???

### DIFF
--- a/src/Language/PureScript/TypeChecker/Synonyms.hs
+++ b/src/Language/PureScript/TypeChecker/Synonyms.hs
@@ -36,10 +36,10 @@ replaceAllTypeSynonyms'
 replaceAllTypeSynonyms' syns kinds = everywhereOnTypesTopDownM try
   where
   try :: SourceType -> Either MultipleErrors SourceType
-  try t = fromMaybe t <$> go 0 [] [] t
+  try t = fromMaybe t <$> go (fst $ getAnnForType t) 0 [] [] t
 
-  go :: Int -> [SourceType] -> [SourceType] -> SourceType -> Either MultipleErrors (Maybe SourceType)
-  go c kargs args (TypeConstructor _ ctor)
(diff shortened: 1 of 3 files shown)


kl0tl:
This looks way better than my attempt if the source spans of the synonyms to expand are always correct!

natefaubion (approved):
I think that this is a general improvement to type synonym expansion errors for sure, so I'd be in favor of this change regardless, and I appreciate the test case. I'll just note that I think fixing the issue with synonyms and deriving is somewhat coincidental. I think the real fix is to defer deriving and typeclass desugaring in general to after kind-checking, which would catch this when checking `newtype N`. I'm happy to go with this for now.

kl0tl:
> I think the real fix is to defer deriving and typeclass desugaring in general to after kind-checking

This is something I’ve been working on this summer! Hopefully the only thing left is to rewrite the binding group desugaring step. I plan to get back to that once v0.14 is out.

hdgarrood (approved):

PR #3952, opened 2020-10-27
Improve error message when `negate` isn't imported

### LINKED ISSUE #2109: Improve error message when 'negate' isn't imported
Currently, using rebindable syntax when the relevant names are not in scope can produce some very confusing errors, especially in `psci`, or when using an explicit `Prelude` import which doesn't include `bind` or `negate`:

```
$ pulp psci
 ____                 ____            _       _   
|  _ \ _   _ _ __ ___/ ___|  ___ _ __(_)_ __ | |_ 
| |_) | | | | '__/ _ \___ \ / __| '__| | '_ \| __|
|  __/| |_| | | |  __/___) | (__| |  | | |_) | |_ 
|_|    \__,_|_|  \___|____/ \___|_|  |_| .__/ \__|
                                       |_|        

:? shows help
> let x = 1
> let y = -1
Error found:
in module $PSCI
at  line 1, column 9 - line 1, column 10

  Unknown value negate


See https://github.com/purescript/purescript/wiki/Error-Code-UnknownValue for more information,
```

The obvious fix is to add special cases for `UnknownValue` errors where the name is a rebindable syntax name, although I suppose we might also have to add some information to the AST to tell the difference between desugared rebindable syntax and explicit calls to `bind` or `negate` etc.

Also, I wonder if we might also consider making the names a little more obvious and specific, like `_syntax_bind` or `_syntax_negate`, so that people can define functions called `bind` or `negate` without messing up do notation or unary negation.

mhmdanas:
Fixes #2109.

Also added tests for `bind` and `discard`. If they are unwanted, I'll remove them.

### DIFF
--- a/tests/purs/failing/2109-discard.out
+++ b/tests/purs/failing/2109-discard.out
@@ -1,0 +1,10 @@
+Error found:
+in module [33mMain[0m
+at tests/purs/failing/2109-discard.purs:7:3 - 7:12 (line 7, column 3 - line 7, column 12)
+
+  Unknown value [33mdiscard[0m. You're probably using do-notation, which the compiler replaces with calls to the [33mdiscard[0m function. Please import [33mdiscard[0m from module [33mPrelude[0m
+
+
+See https://github.com/purescript/documentation/blob/master/errors/UnknownName.md for more information,
+or to contribute content related to this error.
+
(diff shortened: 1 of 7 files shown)


tests/purs/failing/2109-discard.out
```diff
@@ -0,0 +1,10 @@
+Error found:
+in module [33mMain[0m
+at tests/purs/failing/2109-discard.purs:7:3 - 7:12 (line 7, column 3 - line 7, column 12)
+
+  Unknown value [33mdiscard[0m. You're probably using do-notation, which the compiler replaces with calls to the [33mdiscard[0m function. Please import [33mdiscard[0m from module [33mPrelude[0m
```
mhmdanas, line 5: I feel like the error for `discard` is not ideal. Do-notation does not compile to calls to `discard`, but to `bind`. Should I change this in another PR?

kritzcreek: > Do-notation does not compile to calls to discard, but to bind. Should I change this in another PR?

The compiler compiles this expression:
```
do
  x <- act1
  act2 x
  act3
```
to `bind act1 (\x -> discard (act2 x) (\_ -> act3))`

mhmdanas: I know that it does. My point is that this error message implies that do-notation compiles primarily to `discard`, where the most important function for do-notation is `bind`.

natefaubion (approved):

tests/purs/failing/2109-discard.out
hdgarrood: I agree, I think we can make this slightly clearer/less potentially misleading. Addressing this in another PR would be good, yeah.

hdgarrood (approved):

PR #3954, opened 2020-11-13
Fix #3953

### LINKED ISSUE #3953: Parse error when using minus sign in a pattern match
## Description

In this snippet I try to create an operator alias for a constructor:
```purescript
module Main where

data Tuple a b = Tuple a b

infixl 6 Tuple as -

test =
    let tuple = "" - ""
        left - right = tuple
    in left
```
When I compile it with `purs compile src1/Main.purs`, it compiles without errors.
But when I try to run it with `spago repl`, it fails with:
```
at src1/Main.purs:9:16 - 9:21 (line 9, column 16 - line 9, column 21)

  Unable to parse module:
  Unexpected token 'right'
```
Using other symbols like `+` or `*` compiles and runs as expected, it seems only `-` is affected.

## To Reproduce

Create a file with code above and run it with `spago repl`

## Expected behavior

Module compiles and repl starts

## PureScript version

0.13.8

rhendric:
This commit resolves a grammar conflict between the rule allowing
negative numbers to be atomic binders (so that, e.g., `foo -1 = True` is
a valid partial function declaration) and the rule allowing the minus
sign to be an operator separating binders.

The total number of conflicts in the grammar increases by one with this
commit, but what is actually happening is that a state that previously
had one conflict is now conflict-free, and two states that already had
conflicts involving the token '-' now have one extra conflict each. The
extra conflicts aren't conceptually new, though; those states correspond
to productions where a rule containing '-' was duplicated with OP_NEGATE
replacing '-'. So despite the %expect number increasing, I consider this
a net decrease in the conflicted-ness of the grammar.

Fixes #3953.

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Types.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Types.hs
@@ -66,6 +66,7 @@ data Token
   | TokBackslash
   | TokLowerName ![Text] !Text
   | TokUpperName ![Text] !Text
+  | TokNegate -- Only used for '-' when it immediately precedes a digit. TokOperators may end up being negations as well.
   | TokOperator ![Text] !Text
   | TokSymbolName ![Text] !Text
   | TokSymbolArr !SourceStyle
--- a/lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
@@ -313,6 +313,9 @@ insertLayout src@(SourceToken tokAnn tok) nextPos stack =
     TokLowerName [] _ ->
       state & insertDefault & popStack (== LytProperty)
 
+    TokNegate ->
+      state & collapse offsideEndP & insertSep & insertToken src
+
     TokOperator _ _ ->
       state & collapse offsideEndP & insertSep & insertToken src
 
--- a/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs
@@ -12,7 +12,7 @@ import Control.Monad (join)
 import qualified Data.Char as Char
 import qualified Data.DList as DList
 import Data.Foldable (foldl')
-import Data.Functor (($>))
+import Data.Functor (($>), (<&>))
 import qualified Data.Scientific as Sci
 import Data.String (fromString)
 import Data.Text (Text)
@@ -328,7 +328,11 @@ token = peek >>= maybe (pure TokEof) k0
   operator :: [Text] -> [Char] -> Lexer Token
   operator qual pre = do
     rest <- nextWhile isSymbolChar
-    pure . TokOperator (reverse qual) $ Text.pack pre <> rest
+    if null qual && pre == "-" && Text.null rest
+    then peek <&> \case
+      Just ch | isDigitChar ch -> TokNegate
+      _                        -> TokOperator [] "-"
+    else pure . TokOperator (reverse qual) $ Text.pack pre <> rest
 
   {-
     moduleName
--- a/lib/purescript-cst/src/Language/PureScript/CST/Parser.y
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Parser.y
@@ -35,7 +35,7 @@ import qualified Language.PureScript.Roles as R
 import Language.PureScript.PSString (PSString)
 }
 
-%expect 95
+%expect 96
 
 %name parseType type
 %name parseExpr expr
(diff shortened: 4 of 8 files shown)


lib/purescript-cst/src/Language/PureScript/CST/Types.hs
```diff
@@ -66,6 +66,7 @@ data Token
   | TokBackslash
   | TokLowerName ![Text] !Text
   | TokUpperName ![Text] !Text
+  | TokNegate -- Only used for '-' when it immediately precedes a digit. TokOperators may end up being negations as well.
```
natefaubion, line 69: Why do we need a new token for this?

rhendric: Short answer: because `let a -1 = x` is ambiguous in a way that can't be resolved with one token of lookahead otherwise (I think).

Longer answer: The production for `letBinding` includes `ident many(binderAtom) guardedDecl` and `binder1 '=' exprWhere`. After consuming the `a` and reducing it to `ident`, the parser needs to be able to decide, with one token of lookahead, whether to reduce from `ident` to `binderAtom` (on the way to `binder1`) or shift deeper into the `ident many(binderAtom) guardedDecl` rule. In `let a - b = x`, the desired action is to reduce. In `let a -1 = x`, the desired action is to shift. The only way to know is to know what lies ahead of the `-`, and with one token of lookahead, the only way to do that is to encode it in the token representing `-`, which ordinarily is a `TokOperator`. Hence, new token. (I could also have added a new field to `TokOperator` but this seemed cleaner.)

Entirely possible that there's a different way to write the grammar that doesn't require a new token type, but I'm pretty sure that would be a bigger change and a bigger risk of introducing more parsing bugs.

natefaubion: Does it work if you introduce a production between `binder1` and `binder2`, much like `expr3`, that introduces unary negation? We could then just put a guard to check that it's a numeric literal?

rhendric: If you want `let a -1 = x` to continue to parse, then the new `binder1.5` nonterminal would need to replace `binderAtom` in `letBinding : ident many(binderAtom) guardedDecl`. So you'd still have a shift-reduce conflict between that rule and `binderAtom : ident` when a `-` is next. Unless the guard can prevent the shift somehow? I don't know how that would work, but I'm no Happy expert by any means.

natefaubion: If `let a -1 = x` parses as unary negation, then I think that is the bug. That's not valid expression syntax, and so it shouldn't be valid binder syntax. To me that should be a binary operator, since `a -1` parses as a binary operator with expression syntax. Unary negation in that case should require parens `let a (-1) = x`.

rhendric: That would definitely simplify things. I like consistency between expression and binder syntax.

But there is [this test](https://github.com/purescript/purescript/blob/master/tests/purs/passing/NegativeBinder.purs). And for whatever it's worth, this syntax works in Haskell as well.

Seems above my pay grade to decide one way or the other.

natefaubion: We already diverge from Haskell regarding unary negation in expressions, and you also have to consider whatever extensions you have on. The way Haskell parses binders is very difficult to reason about. l Since we lost history during some file relocating, it's hard to tell if that test predates the unary negation changes. In my opinion, this is a clear grammar ambiguity, and I think we should resolve it the same way that we do for expressions. It's technically a breaking change (though I'd consider the current behavior a bug), but it so happens we have an upcoming breaking release! 😆 

That is:
```
let a -1 = x
```
Should always parse as a binary operator, since `a -1` in expression position would parse as a binary operator.
```
let a (-1) = x
```
Should parse as unary negation. This only applies in equational binders, and lambda binders.
```
case x, y of
  -1, -1 -> ...
```
Should be allowed since there are explicit delimiters.

What do you think? cc @hdgarrood @garyb @kl0tl

hdgarrood: I agree with your analysis @natefaubion. It does seem wrong to me that `let a -1 = x` parses in a different way to any other operator in a binder. I agree that it should work the same way as other operators.

garyb: Agreed :+1:

rhendric: The test was added in #576, which I assume does predate the change you mention. Well, that's good enough for me!

rhendric:
Closing in favor of #3956.

PR #3956, opened 2020-11-16
Unsupport bare negative literals as equational binders

### LINKED ISSUE #3953: Parse error when using minus sign in a pattern match

rhendric:
This is a breaking change. The syntax

    foo -1 = ...

as part of an equational definition, either at top-level or in a
let-binding, is no longer supported. The fix is to wrap the negation in
parentheses, as one would for an expression:

    foo (-1) = ...

This enables using `-` as a constructor operator in binders.

Fixes #3953.

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Parser.y
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Parser.y
@@ -35,7 +35,7 @@ import qualified Language.PureScript.Roles as R
 import Language.PureScript.PSString (PSString)
 }
 
-%expect 95
+%expect 93
 
 %name parseType type
 %name parseExpr expr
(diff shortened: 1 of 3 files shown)


natefaubion (approved):
Looks good! 👍

garyb (approved):

PR #3961, opened 2020-12-02
Reform handling of quote characters in raw strings

### LINKED ISSUE #3959: Quote gets eaten as part of quadruple quote
## Description

When quadruple doublequotes appear in a file, it should be parsed as the start of a triple-quoted block starting with a single doublequote character. Instead, it seems that the quote simply disappears and does not appear in the generated file.

## To Reproduce

```purescript
module Test where

s :: String
s = """"should start with a quote"""
```

Generated output:

```js
// Generated by purs version 0.13.3
"use strict";
var s = "should start with a quote";
module.exports = {
    s: s
};

```

## Expected behavior

The generated string should start with a quote: `var s = "\"should start with a quote"`.

## Additional context

I believe this is a regression from 0.12, since I included this in code that worked fine before upgrading to 0.13.

## PureScript version

0.13.3 and 0.14 (master branch)

rhendric:
This corrects a number of regressions introduced in 0.13, and imposes
the uniform restriction that `"""` can't appear anywhere inside a raw
string literal. It's possible (in fact, slightly easier) to lex strings
such that an unlimited number of quotes are allowed at the end (and only
at the end) of a raw string literal, but the simplicity and clarity of a
uniform rule outweighs the implementation cost and the very minor loss
of expressiveness.

Closes #3959.

### DIFF
--- a/lib/purescript-cst/src/Data/Text/PureScript.hs
+++ b/lib/purescript-cst/src/Data/Text/PureScript.hs
@@ -1,0 +1,23 @@
+-- |
+-- This module contains internal extensions to Data.Text.
+--
+module Data.Text.PureScript (spanUpTo) where
+
+import Prelude
+
+import Data.Text.Internal (Text(..), text)
+import Data.Text.Unsafe (Iter(..), iter)
+
+-- | /O(n)/ 'spanUpTo', applied to a number @n@, predicate @p@, and text @t@,
+-- returns a pair whose first element is the longest prefix (possibly empty) of
+-- @t@ of length less than or equal to @n@ of elements that satisfy @p@, and
+-- whose second is the remainder of the text.
+{-# INLINE spanUpTo #-}
+spanUpTo :: Int -> (Char -> Bool) -> Text -> (Text, Text)
+spanUpTo n p t@(Text arr off len) = (hd, tl)
+  where hd = text arr off k
+        tl = text arr (off + k) (len - k)
+        !k = loop n 0
+        loop !n' !i | n' > 0 && i < len && p c = loop (n' - 1) (i + d)
+                    | otherwise                = i
+            where Iter c d = iter t i
--- a/tests/purs/passing/BlockStringEdgeCases.purs
+++ b/tests/purs/passing/BlockStringEdgeCases.purs
@@ -1,0 +1,30 @@
+module Main where
+
+import Prelude
+import Effect.Console (log)
+import Test.Assert (assert')
+
+data Tuple a b = Tuple a b
+derive instance tupleEq :: (Eq a, Eq b) => Eq (Tuple a b)
+
+main = do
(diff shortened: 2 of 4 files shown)


lib/purescript-cst/src/Data/Text/PureScript.hs
```diff
@@ -0,0 +1,23 @@
+-- |
+-- This module contains internal extensions to Data.Text.
+--
+module Data.Text.PureScript (spanUpTo) where
```
rhendric, line 4: So, this may be a questionable choice. It's possible to correctly implement `spanUpTo` without using internal and unsafe API from `text`, but only in what I imagine is a somewhat inefficient way—`span` does a Unicode scalar-to-code-point walk on the text, and then `compareLength` does the same, and then if the initial span is too large a `splitAt` does a third walk. I don't know if this makes a real world difference, is the thing. But implementing `spanUpTo` by just tweaking the internal implementation of `span`, as I did here, is very straightforward. I stuffed it in a private module here to limit the consequences of importing these parts of the `text` API elsewhere. As far as premature optimizations go, this one seems pretty benign. But then, they probably all do at first. I can easily rewrite it if this raises flags.

natefaubion (approved):

tests/purs/passing/BlockStringEdgeCases.purs
```diff
@@ -0,0 +1,30 @@
...
+  assert' "ends with two quotes" ("""x""""" == "x\"\"")
+  assert' "starts and ends with two quotes" ("""""x""""" == "\"\"x\"\"")
+  assert' "mixture 1" ("""""x"y""z"""" == "\"\"x\"y\"\"z\"")
+  assert' "mixture 2" ("""x"y""z""" == "x\"y\"\"z")
+
+  -- These last tests are more about forbidding certain raw string literal
+  -- edge cases than about wanting to support mashing string literals against.
+  -- each other, which is techically legal but generally, if not universally,
+  -- a bad idea.
+  assert' "too many quotes 1" (Tuple """"""""" " == Tuple "\"\"" " ")
```
hdgarrood, line 26: It occurs to me that it might be worth trying to update the PureScript syntax highlighting because the syntax highlighting here is wrong, presumably? This is what it looks like for me:

Screenshot from 2020-12-16 03-04-06



The highlighting used on GitHub is defined in https://github.com/purescript-contrib/atom-language-purescript if I remember correctly. This is not a blocker for merging, just thought it's worth mentioning.

hdgarrood: Ah yes, I see what's going on here now; the highlighting is definitely wrong. It wouldn't compile if the parser understood the boundaries of these string literals the same way this highlighter did.

hdgarrood (approved):

PR #3779, opened 2020-01-29
Polykinds

natefaubion:
This PR implements PolyKinds based on [Kind Inference for Datatypes](https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf).

## Type In Type

The type-checker now supports `TypeInType` (or `Type :: Type`), so the old `Kind` data type and namespace is now gone. Kinds and types are the same and exist in the same namespace.

Previously one could do:

```purescript
foreign import kind Boolean
foreign import data True :: Boolean
foreign import data False :: Boolean
```

Where the kind `Boolean` and type `Boolean` were two different things. This is no longer the case. The `Prim` kind `Boolean` is now removed, and you can just use `Prim` type `Boolean` in the same way. This is a breaking change.

The compiler still supports the old `foreign import kind` syntax but it warns that it's deprecated.

```purescript
foreign import kind Foo
```

> Foreign kind imports are deprecated and will be removed in a future release. Use 'foreign import data' instead.

It is treated internally as:

```purescript
foreign import data Foo :: Type
```

Likewise, `kind` imports and exports are deprecated and treated the same as a type import or export.

> Kind imports are deprecated and will be removed in a future release. Omit the 'kind' keyword instead.

The special unary `#` syntax for row kinds is still supported, but deprecated and will warn. There is now `Prim.Row :: Type -> Type` which can be used like a normal type constructor.

> Unary '#' syntax for row kinds is deprecated and will be removed in a future release. Use the 'Row' kind instead.

All of these deprecations have suggested fixes in the JSON output, so tools like `purescript-suggest` (or your IDE plugin) can automatically apply them.

## Kind Signatures

With PolyKinds, all type-level declarations are generalized.

```purescript
data Proxy a = Proxy
```

Previously, this had the `Type`-defaulted kind `Type -> Type`. Now this will be generalized to `forall k. k -> Type`.

When writing polymorphic things, it's often nice to write signatures. Since we don't really quantify over free type variables, it's also necessary in the case that two poly-kinded arguments must have the same kind. I've introduced kind signature declarations, similar to [standalone kind signatures](https://ryanglscott.github.io/2020/01/05/five-benefits-to-using-standalonekindsignatures/) in GHC.

```purescript
data Proxy :: forall k. k -> Type
data Proxy a = Proxy
```

[...]

PR #3965, opened 2020-12-07
Expand type synonyms in instances dictionaries

kl0tl:
Fix https://github.com/purescript/purescript/issues/3104.

### DIFF
--- a/src/Language/PureScript/Sugar.hs
+++ b/src/Language/PureScript/Sugar.hs
@@ -71,5 +71,5 @@ desugar externs =
     >=> rebracket externs
     >=> checkFixityExports
     >=> deriveInstances externs
-    >=> desugarTypeClasses externs
+    >=> (\(syns, kinds, mod) -> desugarTypeClasses externs syns kinds mod)
     >=> createBindingGroupsModule
--- a/src/Language/PureScript/Sugar.hs
+++ b/src/Language/PureScript/Sugar.hs
@@ -3,17 +3,23 @@
 --
 module Language.PureScript.Sugar (desugar, module S) where
 
+import Prelude
+
 import Control.Category ((>>>))
 import Control.Monad
 import Control.Monad.Error.Class (MonadError)
 import Control.Monad.Supply.Class (MonadSupply)
 import Control.Monad.State.Class (MonadState)
 import Control.Monad.Writer.Class (MonadWriter)
+import Data.Maybe (mapMaybe)
+
+import qualified Data.Map as M
 
 import Language.PureScript.AST
 import Language.PureScript.Errors
 import Language.PureScript.Externs
 import Language.PureScript.Linter.Imports
+import Language.PureScript.Names
 import Language.PureScript.Sugar.BindingGroups as S
 import Language.PureScript.Sugar.CaseDeclarations as S
 import Language.PureScript.Sugar.DoNotation as S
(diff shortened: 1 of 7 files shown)


src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -323,11 +333,12 @@ typeInstanceDictionaryDeclaration sa@(ss, _) name mn deps className tys decls =
...
       -- The type is a record type, but depending on type instance dependencies, may be constrained.
       -- The dictionary itself is a record literal.
-      let superclasses = superClassDictionaryNames typeClassSuperclasses `zip`
-            [ Abs (VarBinder ss UnusedIdent) (DeferredDictionary superclass tyArgs)
-            | (Constraint _ superclass _ suTyArgs _) <- typeClassSuperclasses
-            , let tyArgs = map (replaceAllTypeVars (zip (map fst typeClassArguments) tys)) suTyArgs
-            ]
+      tys' <- traverse (replaceAllTypeSynonymsM syns kinds) tys
+      superclassesDicts <- for typeClassSuperclasses $ \(Constraint _ superclass _ suTyArgs _) -> do
+        suTyArgs' <- traverse (replaceAllTypeSynonymsM syns kinds) suTyArgs
```
rhendric, line 338: Unless I'm missing something, this `kinds` that got passed all over the place is always [`mempty`](https://github.com/purescript/purescript/pull/3965/files#diff-b2c45abe2b96a6ae2e0da4d769b691e5bc4ee0e1d86e46d9b973df822d10a698R81). I'm not sure what the consequences of that are, but if this is the intended behavior, why not just use `mempty` here?

src/Language/PureScript/Sugar.hs
```diff
@@ -71,5 +71,5 @@ desugar externs =
     >=> rebracket externs
     >=> checkFixityExports
     >=> deriveInstances externs
-    >=> desugarTypeClasses externs
+    >=> (\(syns, kinds, mod) -> desugarTypeClasses externs syns kinds mod)
```
rhendric, line 74: Might I suggest pulling the synonym-finding out of `deriveInstances` and having it consumed by both stages, instead of this coupling of the two stages together? Then this section could be

```hs
    >=> checkFixityExports
    >=> (findTypeSynonyms externs >>= \syns ->
          deriveInstances externs syns
            >=> desugarTypeClasses externs syns)
    >=> createBindingGroupsModule
```

which keeps the pipeline flexible if stages get added or moved around.

src/Language/PureScript/Sugar/TypeClasses.hs
kl0tl: I kept a `kinds` binding while applying your other suggestion but attached to it a comment rephrasing what Nate said in https://github.com/purescript/purescript/pull/3779 about the changes to the `Sugar.TypeClasses` module:

> Since this happens before type-checking (and thus before kind-checking), there's some dodgy code to expand type synonyms before kinds are known. I don't know if this is a problem in practice (you might just get garbage ill-kinded results and bad errors). My opinion is that this desugaring happens too soon in the pipeline and should happen after other type declarations are kind checked.

rhendric: That's good enough for me!

natefaubion (approved):

hdgarrood (approved):

Issue #3734, opened 2019-09-23
Issue with synonym expansion in superclass constraints


drewolson:
In this [example](https://github.com/drewolson/newtype-stack-problems/blob/3e8efec779033c9e6f2cd7cfa97ad8fe19be9b35/src/Main.purs), I've implemented the `MonadThrow` and `MonadAsk` type classes for a `newtype` called `App`. The config provided by my `MonadAsk` is a record. Because of the restriction on aliases in type classes, I must use `TypeEquals` to define this instance.

Once defined, I can interpret function with the `MonadAsk Env m` constraint with my `App` type (see `doStuff`), but if I try to constrain a function with the same `MonadAsk Env m` in a type class alias, it no longer works (see `doStuffA` and `doStuffB`). I think this may be a bug with the type checker.

hdgarrood:
Can you please include the errors you're getting?

garyb:
I suspect it's to do with not using `TE.from` / `TE.to` as necessary to get the instances to work. You can't ever assume the type being asked for is `Env`, it'll always need the coercion too.

garyb:
I'd suggest newtyping `Env`, and then providing an `askEnv :: MonadAsk Env m => m { ... }` function that unwraps the record, rather than using `TypeEquals` for this sort of thing, by the way.

drewolson:
@hdgarrood Apologies, my error is below[1].

@garyb I agree, and in my real application I did end up `newtype`ing `Env`. But the fact that the code compiles with a direct constraint of `MonadAsk Env m` but fails with a type class alias of the same type seems strange.

[1]
```Error found:
in module Main
at src/Main.purs:62:14 - 62:17 (line 62, column 14 - line 62, column 17)

  Could not match type
      
    r4
      
  with type
                   
    { foo :: String
    | t2           
    }              
                   

while solving type class constraint
                                                     
  Control.Monad.Reader.Class.MonadAsk { foo :: String
                                      | t2           
                                      }              
                                      m3             
                                                     
while checking that type forall m r. MonadAsk r m => m r
  is at least as general as type t0 t1
while checking that expression ask
  has type t0 t1
in value declaration doStuffB

where m3 is a rigid type variable
        bound at (line 61, column 12 - line 66, column 26)
      r4 is a rigid type variable
        bound at (line 61, column 12 - line 66, column 26)
      t0 is an unknown type
      t1 is an unknown type
      t2 is an unknown type

[...]

PR #3966, opened 2020-12-07
Expand type synonyms in superclasses

kl0tl:
Fix https://github.com/purescript/purescript/issues/3734.

### DIFF
--- a/src/Language/PureScript/TypeChecker.hs
+++ b/src/Language/PureScript/TypeChecker.hs
@@ -151,8 +151,8 @@ addTypeClass
   -> m ()
 addTypeClass _ qualifiedClassName args implies dependencies ds kind = do
   env <- getEnv
-  let newClass = mkNewClass env
-      qualName = fmap coerceProperName qualifiedClassName
+  newClass <- mkNewClass env
+  let qualName = fmap coerceProperName qualifiedClassName
       hasSig = qualName `M.member` types env
   unless (hasSig || not (containsForAll kind)) $ do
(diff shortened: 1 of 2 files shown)


src/Language/PureScript/TypeChecker.hs
```diff
@@ -163,10 +163,12 @@ addTypeClass _ qualifiedClassName args implies dependencies ds kind = do
     classMembers :: [(Ident, SourceType)]
     classMembers = map toPair ds
 
-    mkNewClass :: Environment -> TypeClassData
-    mkNewClass env = makeTypeClassData args classMembers implies dependencies ctIsEmpty
+    mkNewClass :: Environment -> m TypeClassData
+    mkNewClass env = do
+      implies' <- (traverse . overConstraintArgs . traverse) replaceAllTypeSynonyms implies
```
rhendric, line 168: This is pretty minor, but it's a little odd to be both passing in an explicit `env` as a parameter and pulling an implicit `Environment` out of the monad in `replaceAllTypeSynonyms`—in current usage they'll always be the same `Environment`, but that fact isn't immediately obvious. Since you're adding the monad to `mkNewClass`, maybe get rid of the `env` parameter, use `getEnv` inside, and pass it to `findSuperClass` or something like that?

Otherwise, LGTM!

kl0tl: You’re right, I applied your suggestion!

natefaubion (approved):

hdgarrood (approved):

PR #3968, opened 2020-12-14
Trigger TCO even when there are some self references in non-tail positions 

### LINKED ISSUE #3967: Proposal: Take more opportunities to trigger TCO
## Summary

You don't need every self reference to be in tail position to trigger TCO. Why not make use of this beautiful optimization whenever it is possible?

## Motivation
```
data Pokemon
  = Charizard {lvl : Int}
  | Exeggcute {eggz : [Pokemon]}

```

No TCO:
```
levelupAll :: [Pokemon] -> [Pokemon] -> [Pokemon]
levelupAll ps acc = case ps of
  [] -> reverse acc
  (Charizard spec@{lvl: lvl} : rest) ->
    levelupAll rest ((Charizard spec{lvl: lvl + 1}):acc)
  (Exeggcute spec@{eggz: eggz} : rest) ->
    levelupAll rest ((Exeggcute spec{eggz: levelupAll eggz}):acc)
```
TCO:
```
levelupAll :: Pokemon] -> [Pokemon] -> [Pokemon]
levelupAll ps acc = case ps of
  [] -> reverse acc
  (Charizard spec@{lvl: lvl} : rest) ->
    levelupAll rest ((Charizard spec{lvl: lvl + 1}):acc)
  (Exeggcute spec@{eggz: eggz} : rest) ->
    levelupAll rest ((Exeggcute spec{eggz: levelupAllHereIExplicitlyBreakTheTCO eggz}):acc)

levelupAllHereIExplicitlyBreakTheTCO ps = levelupAll ps
```
(modulo some syntax).

The current implementation disables TCO whenever there is a non-tail reference to the current function – not necessary in call, can be also passing it as an argument to some other function or value definition. This is ridiculous and requires presented above workarounds.

I don't know Haskell's policy about it at all, but in JS we suffer strongly the stack limit so loopifying calls can be crucial in many cases – for instance in our 
[erlscripten](https://github.com/erlscripten/erlscripten) project where we need to implement the mentioned trick in order to be able to compare long lists of some erlang terms: [example](https://github.com/erlscripten/erlps-core/blob/84cf2df8474a707fb908a427c4a4b4499f575699/src/Erlang/Type.purs#L194)

## Proposal

Trigger TCO everywhere it is possible regardless of if there are some other non-tail calls. I have already prepared a PR for this.

## Examples

All of the following functions should be stack safe:
```
occasionalTCO1 :: Int -> Int
occasionalTCO1 0 = 1
occasionalTCO1 n =
  occasionalTCO1 (n - occasionalTCO1 0)

occasionalTCO2 :: Int -> Int
occasionalTCO2 0 = 1
occasionalTCO2 n =
  let x = occasionalTCO2 0
  in occasionalTCO2 (n - x)

occasionalTCO3 :: Int -> Int
occasionalTCO3 0 = 1
occasionalTCO3 n =
  if occasionalTCO3 0 == n
  then 1
  else occasionalTCO3 (n - occasionalTCO3 0)

occasionalTCO4 :: Int -> Int
occasionalTCO4 0 = 1
occasionalTCO4 n | 1 <- occasionalTCO4 0 =
  case occasionalTCO4 0 + n of
    2 -> 1
    x -> occasionalTCO4 (x - 2)
occasionalTCO4 _ = 1

occasionalTCO5 :: Int -> Int
occasionalTCO5 0 = 1
occasionalTCO5 n | n > 10 =
  occasionalTCO5 (n - 1)
occasionalTCO5 n =
  if n > 5
  then occasionalTCO5 $ n - 1
  else call occasionalTCO5 (n - 1)
```

radrow:
Fixes #3967

Tests are passing on my side.


hdgarrood:
Unfortunately it's very easy to introduce bugs when working on this code, so I think we need to be really careful about this (searching through this repo's issue history for TCO should provide a bit more context). It's unfortunately quite possible that our tests don't cover as many cases here as they should, so the fact that the tests are passing is not enough for me to be confident. The tests in this PR only check that the program finishes; they don't check the results of the TCO-transformed functions, which I think is also needed.

For me, this would need a much more developed justification of why it works and would not break existing programs before we can consider it. In any case I think this needs to wait until 0.14.0 has been released.

JordanMartinez:
> In any case I think this needs to wait until 0.14.0 has been released.

I agree with this.

radrow:
I couldn't find any semantics testing suite. Do we have such a thing? If so, I can write some tests to make it more clear that this approach doesn't break anything.

hdgarrood:
I’m not sure what you mean by “semantics testing suite.” More tests might help but they aren’t really a substitute for an explanation of your reasoning and approach.

rhendric:
(Attempting to move discussion of the proposal, as distinct from this implementation, to #3967.)

radrow:
My approach is just to alter the conditions for the TCO to trigger while leaving the present loopifier unchanged. Indeed, the only thing I changed was replacing "allInTailPosition" with "anyInTailPosition". The algorithm for resolving the tail calls remains as it was before – so if it was working back then, then should it work now as well. The only issue I can think of is that it may be a bit too much of artificial loops, but this discussion is continued in #3967

By "semantics teseting suite" I meant some already defined way of validating if the generated code evaluates in a desired way. I suspect I could do it by throwing errors on failing assertions in the `passing` set, but I didn't see this approach around the code, so I haven't decided to go for it.

rhendric:
> I suspect I could do it by throwing errors on failing assertions in the `passing` set, but I didn't see this approach around the code, so I haven't decided to go for it.

They're in the minority, but a number of tests do use the functions from [purescript-assert](https://pursuit.purescript.org/packages/purescript-assert/4.1.0/docs/Test.Assert); you can grep for `Test.Assert` to see which ones.

PR #3970, opened 2020-12-22
Update license-related error messages

### LINKED ISSUE #3836: Improve license errors from `purs publish`
https://github.com/purescript/purescript/pull/3834#issuecomment-609929541

fsoikin:
This is a continuation of https://github.com/purescript/purescript/pull/3834
Text of the errors updated according to @hdgarrood's suggestions.
Fixes #3836

### No license field
```
published.
Details:
  No license is specified in package manifest. Please add a "license" property
  with a SPDX license expression. For example, any of the following would be
  acceptable:

    * "MIT"                  
    * "Apache-2.0"           
    * "BSD-2-Clause"         
    * "GPL-2.0-or-later"     
    * "(GPL-3.0-only OR MIT)"
 
  See https://spdx.org/licenses/ for a full list of licenses. For more
  information on SPDX license expressions, see https://spdx.org/ids-how

  Note that distributing code without a license means that nobody will (legally)
  be able to use it.

  It is also recommended to add a LICENSE file to the repository, including your
  name and the current year, although this is not necessary.
```

### Invalid license field
```
published.
Details:
  The license specified in package manifest is not a valid SPDX license
  expression. Please update the "license" property so that it is a valid SPDX
  license expression. For example, any of the following would be acceptable:

    * "MIT"                  
    * "Apache-2.0"           
    * "BSD-2-Clause"         
    * "GPL-2.0-or-later"     
    * "(GPL-3.0-only OR MIT)"
```

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -51,6 +51,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@fehrenbach](https://github.com/fehrenbach) | Stefan Fehrenbach | [MIT license](http://opensource.org/licenses/MIT) |
 | [@felixSchl](https://github.com/felixSchl) | Felix Schlitter | [MIT license](http://opensource.org/licenses/MIT) |
 | [@FrigoEU](https://github.com/FrigoEU) | Simon Van Casteren | [MIT license](http://opensource.org/licenses/MIT) |
+| [@fsoikin](https://github.com/fsoikin) | Fyodor Soikin | [MIT license](http://opensource.org/licenses/MIT) |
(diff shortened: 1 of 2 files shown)


hdgarrood (commented):
Looks great, but would you mind keeping with the existing style for multiline strings (ie using `concat` and a list, rather than backslashes)?

hdgarrood:
You’re in contributors.md already right?

fsoikin:
Sure, will change the style.

But no, I don't think I'm in `contributors.md`, at least not to my knowledge.

fsoikin:
Just checked: I'm not in `contributors.md`
Should I add myself in this PR?

hdgarrood (approved):
Fantastic! Thank you :heart:

thomashoneyman (approved):

PR #3971, opened 2020-12-26
Fix operator qualifier reversal bug

### LINKED ISSUE #3892: Bug: Unknown qualified prefix operators render incorrect module name
## Description

If you reference an unknown qualified prefix operator (like `Foo.Bar.Baz.(?)`), the error renders with an incorrect module name (like `Unknown module Baz.Bar.Foo`). This happens for both type and value operators.

## To Reproduce

Given a module like:

```PureScript
module Bug where

foo = Foo.Bar.Baz.(?)
```

```console
$ purs compile Bug.purs
Compiling Bug
Error found:
in module Bug
at Bug.purs:3:7 - 3:22 (line 3, column 7 - line 3, column 22)

  Unknown module Baz.Bar.Foo


```

## Expected behavior

The module name should be correct `Foo.Bar.Baz`, in this case.

## Additional context

Unknown qualified infix operators (like `x Foo.Bar.Baz.? y`) seem to work correctly. I.e. renders as `Unknown module Foo.Bar.Baz`.

## PureScript version

0.13.8

This seems to have worked correctly in 0.12.5, and started working incorrectly in 0.13.0 (all the 0.13.x version I tried seem to behave the same).

rhendric:
The relatively unusual combination of a multi-part qualifier and an
operator used as a symbol—i.e., `Foo.Bar.(!)`—was previously being
wrongly interpreted as `Bar.Foo.(!)`.

Fixes #3892.

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs
@@ -322,7 +322,7 @@ token = peek >>= maybe (pure TokEof) k0
         peek >>= \case
           Just ')'
             | isReservedSymbol chs -> throw ErrReservedSymbol
-            | otherwise -> next $> TokSymbolName qual chs
+            | otherwise -> next $> TokSymbolName (reverse qual) chs
           Just ch2 -> throw $ ErrLexeme (Just [ch2]) []
           Nothing  -> throw ErrEof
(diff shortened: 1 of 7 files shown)


JordanMartinez (approved):
I don't usually do work on the compiler because I'm unfamiliar with it. However, this fix is very small and seems quite obvious. LGTM.

hdgarrood (approved):

PR #3977, opened 2020-12-30
Added HLS support via nix

reactormonk:
This PR is built on top of https://github.com/purescript/purescript/pull/3933 to go the cabal route. It also enables straightforward access to the haskell-language-server via nix (see the note in CONTRIBUTING).

Haskell.nix uses the stack snapshot as base, so the cabal freeze file isn't really required if you go this route.


hdgarrood:
Sorry, but I’m not comfortable with having a bunch of extra configuration in this repo which we need to learn how to keep working when it inevitably stops working. (I know what Nix is and that it’s supposed to stop this sort of thing from happening but I do not have that level of confidence in it.)

reactormonk:
> Sorry, but I’m not comfortable with having a bunch of extra configuration in this repo which we need to learn how to keep working when it inevitably stops working. (I know what Nix is and that it’s supposed to stop this sort of thing from happening but I do not have that level of confidence in it.)

Nix is reasonably decent at staying stable. It might break on updating the cabal / stack configuration. Also the `haskell.nix` code could be pinned, haven't done that yet.

I could probably push the Haskell language server configuration upstream, so that wouldn't have to be flying around here.

hdgarrood:
The Nix stuff is more objectionable to me than the HLS stuff. I think it’s extremely  unlikely that you’ll persuade me to merge this; the only way I *might* be on board is if there’s strong support from other core team members.

hdgarrood:
That’s still too much, sorry. The furthest I’m willing to go is a note pointing to your fork in the CONTRIBUTING.md file; I don’t want these files in this repo.

reactormonk:
For quick setup, I considered a `shell.nix` file here which would link to the settings in an own repo. If that's still too much, then just close the PR.

rhendric:
@reactormonk, what are the minimal changes that would need to happen to PureScript to enable someone to develop using HLS? Is all the Nix stuff necessary? Is the extra HIE configuration necessary? It's unclear to me whether this PR is primarily motivated to help enable people who aren't currently hacking on PureScript to do so, or to offer a fancier way to do things that is more to your taste.

I don't really know anything about HLS; my limited experience with Nix makes me think you *should* be able to keep Nix code in a separate repo and install PureScript with that without any changes here, but if I'm wrong, I would be motivated to help you find and merge a minimal fix. What I don't think anyone wants is to see the main PureScript repo loaded up with files specific to a particular dev environment, especially if that dev environment isn't one used by any of the maintainers.

reactormonk:
The HIE config isn't 100% necessary, HLS can _sometimes_ generate it itself just fine. Having the config at hand just makes it more stable. It requires regeneration when you create new internal packages though.

If you have a `shell.nix` file around, it's easier on the convention (and e.g. the nix environment picker from vscode, which picks from the current directory) and for people to find it. It reduces friction - digging through a `contributing.md` file takes mental capacity, if you just wanna browse around. Otherwise you'd have to follow the HLS setup guides - which will be automatic at some point in the future I assume, in which case the config could be discarded again.

hdgarrood:
The HIE config is probably going to break quite quickly when any of the referenced modules are removed or renamed, or new modules are added, so if it can sometimes be generated then I think leaving it out is probably the best bet for now.

PR #3979, opened 2021-01-01
Add STFn inliner rules

### LINKED ISSUE #3923: Proposal: STFn functions -> create lib and add compiler support
## Summary

_Originally posted by @hdgarrood in https://github.com/purescript/purescript-arrays/pull/178#issuecomment-680965097_

The idea would be that we have a collection of data types like
```
STFn1 :: Type -> Type
STFn2 :: Type -> Type -> Type
STFn3 :: Type -> Type -> Type -> Type
```
and so on, like `EffectFn{1,2,3}`, which would perform their effects immediately rather than returning a nullary function, together with inlinable functions like
```
runSTFn1 :: forall a h r. STFn1 a r -> a -> ST h r
runSTFn2 :: forall a b h r. STFn2 a b r -> a -> b -> ST h r
```
and so on, so that code like
```purescript
ST.run do
  arr <- STArray.new
  _ <- runSTFn2 STArray.push arr 3
  pure (STArray.unsafeFreeze arr)
```
might compile to something like
```javascript
function () {
  var arr = STArray.new();
  Data_Array_ST.push(arr, 3);
  return unsafeFreezeImpl(arr);
}
```
where `Data_Array_ST.push` would have the type `forall a. STFn2 (Array a) a Int`.

## Motivation

To make ST code as fast as FFI

https://github.com/purescript/purescript-arrays/pull/178#issuecomment-680938969

JordanMartinez:
This provides us the benefit of `STFn` for the time being before a final decision is made on https://github.com/purescript/purescript-st/issues/31

Fixes #3923

### DIFF
--- a/src/Language/PureScript/Constants/Prelude.hs
+++ b/src/Language/PureScript/Constants/Prelude.hs
@@ -240,6 +240,15 @@ mkEffectFn = "mkEffectFn"
 runEffectFn :: forall a. (IsString a) => a
 runEffectFn = "runEffectFn"
 
+runSTFn :: forall a. (IsString a) => a
+runSTFn = "runSTFn"
+
+mkSTFn :: forall a. (IsString a) => a
+mkSTFn = "mkSTFn"
+
+runSTFn :: forall a. (IsString a) => a
+runSTFn = "runSTFn"
+
 -- Type Class Dictionary Names
 
 data EffectDictionaries = EffectDictionaries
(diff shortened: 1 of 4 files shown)


hdgarrood:
I think we should make a final decision on that issue and we should wait until 0.14.0 is out before merging this; temporary stuff we might end up removing generally shouldn’t go into the compiler, imo.

PR #3980, opened 2021-01-01
Migrate CI to GitHub Actions

JordanMartinez:
My first goal is to get the Linux build to work and then explore Mac and Windows from there


f-f:
@JordanMartinez you might be able to get most of the way starting with [spago's CI workflows](https://github.com/purescript/spago/blob/master/.github/workflows/build.yml) - Linux and macOS builds are basically the same, only Windows needs special care

JordanMartinez:
@f-f Thanks for the tip. I'm trying to port the `.travis.yml` file as close as possible and I'm not sure how much of it is working around Travis CI issues and how much of it is just stuff we need to clean up.

hdgarrood:
Can you keep the build script in its own file please? It’s useful to be able to run the script independently of CI.

JordanMartinez:
> Can you keep the build script in its own file please? It’s useful to be able to run the script independently of CI.

I thought that might be the case. Yeah, I will update this PR to do that once I figure out some other things.

hdgarrood:
It looks like if one job fails, all the others are immediately cancelled rather than being allowed to run to completion. That's quite inconvenient, I think; it's useful to be able to see if a failure is specific to one OS, and it won't be possible to see that if other jobs are cancelled before they can finish. Do you know if we can configure this so that all jobs are allowed to run to completion if one fails? The build should of course still be marked as a failure overall if at least one job fails.

JordanMartinez:
> It looks like if one job fails, all the others are immediately cancelled rather than being allowed to run to completion.

Ah! That would make sense. I was wondering why it was failing.

Upon further research, that's the default, but it can be overridden easily here: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

JordanMartinez:
Great! Looks like we just need to figure out the Windows build and add deployment and we're done.

JordanMartinez:
For the windows build, CI fails with this error in the "Install hlint" step:
```
Run BIN_DIR="$HOME/.local/bin/" BUILD_DIR="$HOME/hlint" ci/install-hlint.sh
  BIN_DIR="$HOME/.local/bin/" BUILD_DIR="$HOME/hlint" ci/install-hlint.sh
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    STACK_ROOT: C:\sr
BIN_DIR=$HOME/.local/bin/: D:\a\_temp\e5060625-9ca9-4443-b216-4ce97b78b8b6.ps1:2
Line |
   2 |  BIN_DIR="$HOME/.local/bin/" BUILD_DIR="$HOME/hlint" ci/install-hlint. …
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | The term 'BIN_DIR=$HOME/.local/bin/' is not recognized as a name of a cmdlet, function, script file,
     | or executable program. Check the spelling of the name, or if a path was included, verify that the path
     | is correct and try again.

Error: Process completed with exit code 1.
```

Looks like Window's default shell is PowerShell. I'm testing in latest commit where that fixes things.

Also, does `STACK_ROOT` need to be updated to `$HOME/.stack` as it currently is in `.travis.yml`, so that caching (and perhaps other things) work properly?

hdgarrood:
We may need to set STACK_ROOT like we did before, yeah; I think the default is different on Windows. It’s possible that it’ll just work on GH Actions though - try pushing an empty commit and seeing what happens?

kritzcreek:
While I understand that we probably can't keep using Travis I still see this as a serious regression.

The last CI timings on master (with OSX being the slowest build):
```
Release build: 19min34s
CI build #1: 11min48s
CI build #2: 11min41s
CI build #2: 13min55s
```

On Github Actions:
```
CI / build (ubuntu-16.04) (pull_request) Successful in 6m — build (ubuntu-16.04)
Details
CI / build (macOS-latest) (pull_request) Successful in 8m — build (macOS-latest)
Details
CI / build (windows-latest) (pull_request) Successful in 27m — build (windows-latest)
Details
```
The Linux and OSX builds are a lot faster, which is great, but the Windows build is taking 27min?! We need all three to pass to get the green checkmark for a commit. What's the difference between the GHA and Travis Windows setup here?

hdgarrood:
It looks to me like the Windows one took longer because dependencies weren't cached. That will hopefully be solved by setting STACK_ROOT.

JordanMartinez:
OK. STACK_ROOT was set in latest commit. Now I'll push another commit and see if that speeds up the Windows build.

hdgarrood:
It looks like the STACK_ROOT environment variable isn’t being set properly, unfortunately, based on this line of output from `stack path`:
```
stack-root: C:\sr
```

JordanMartinez:
Sigh... still isn't working.

JordanMartinez:
Could we work around this issue by also caching `C:\sr` if it's on windows?

hdgarrood:
I would like to get to the bottom of why this isn’t working, since setting environment variables is likely to be something we’ll need to do at some point if not now.

How about setting it via the `env` key at the top level, as described here? https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#env

hdgarrood:
I suspect $HOME is more likely to be expanded than ~ is on Windows, incidentally.

JordanMartinez:
Looks like the environment variable is getting set correctly with `echo "STACK_ROOT=$HOME/.stack" >> $GITHUB_ENV`, which stores `/c/Users/runneradmin/.stack`. Using your `env` suggestion did not lead to a root of `/c/Users/runneradmin/.stack` because `$HOME` wasn't expanded there. Not sure whether I need to do something else to make `$HOME` be fully expanded, but I think this at least shows that `STACK_ROOT` is getting set properly and that the problem is elsewhere.

What I find interesting is that the `$HOME/.stack` expands to `/c/Users/runneradmin/.stack` whereas `stack-root` outputs `c:\sr`. I find it strange that it doesn't output `/c/sr`, which is more like a Linux path.

Maybe we need to set STACK_ROOT using Powershell rather than bash?

JordanMartinez:
> Maybe we need to set STACK_ROOT using Powershell rather than bash?

That didn't work...

JordanMartinez:
I think I found the source of the issue. I set the `STACK_ROOT` variable before the `haskell/actions/setup` step. Unfortunately, `haskell/actions/setup` will automatically [set/overwrite the `STACK_ROOT` to `c:\sr`](https://github.com/haskell/actions/blob/main/setup/src/installer.ts#L25-L28). CI shows [proves this via the debug output I added](https://github.com/purescript/purescript/runs/1639704397#step:12:913).

JordanMartinez:
Yup. i can confirm that that fixes the issue.

JordanMartinez:
@hdgarrood Looks like `$HOME/.stack` and `~/.stack` both (resolve to the same thing on Windows](https://github.com/purescript/purescript/runs/1639730710#step:10:8)

JordanMartinez:
Ok. Next pushed commit should cause the Windows build to use its cache.

hdgarrood:
> I think I found the source of the issue. I set the STACK_ROOT variable before the haskell/actions/setup step. Unfortunately, haskell/actions/setup will automatically set/overwrite the STACK_ROOT to c:\sr. CI proves this via the debug output I added.

Oh wow, good find! That's super annoying!

JordanMartinez:
Latest commit tests whether the Windows build will hit its cache with an actual cache (rather than the empty one it has had before).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false # when true, one job's failure will cancel other jobs
+      matrix:
+        os: [ ubuntu-16.04, macOS-latest, windows-latest ]
+    steps:
+      - name: Add "$HOME/.local/bin/" to PATH
+        run: echo "$HOME/.local/bin/" >> $GITHUB_PATH
+
+      # We set LF endings so that the Windows environment is consistent with Appveyor
```
hdgarrood, line 24: Why is it desirable to have the Windows environment consistent with AppVeyor? We haven't used AppVeyor in this repo's CI for quite a while.

ci/build.sh
```diff
@@ -35,17 +39,32 @@ fi
...
+
+echo "::group::Install dependencies"
+
 # Install snapshot dependencies (since these will be cached globally and thus
 # can be reused during the sdist build step)
 $STACK build --only-snapshot $STACK_OPTS
 
+echo "::endgroup::"
+
+echo "::group::Build source distributions"
```
hdgarrood, line 52: I would say the fact that we run the build in a source distribution is an implementation detail which doesn't need to be drawn attention to here. I'd suggest just "Build" or "Build the compiler" instead.

ci/build.sh
```diff
@@ -35,17 +39,32 @@ fi
...
 $STACK sdist lib/purescript-ast --tar-dir sdist-test/lib/purescript-ast
 tar -xzf sdist-test/lib/purescript-ast/purescript-ast-*.tar.gz -C sdist-test/lib/purescript-ast --strip-components=1
 $STACK sdist lib/purescript-cst --tar-dir sdist-test/lib/purescript-cst
 tar -xzf sdist-test/lib/purescript-cst/purescript-cst-*.tar.gz -C sdist-test/lib/purescript-cst --strip-components=1
 $STACK sdist . --tar-dir sdist-test;
 tar -xzf sdist-test/purescript-*.tar.gz -C sdist-test --strip-components=1
+
+echo "::endgroup::"
+
+echo "::group::Test in a source distribution"
```
hdgarrood, line 64: Same here, I think we don't need to draw attention to the fact that the tests are being run in a source distribution.

hdgarrood (commented):
Looking great so far!

.github/workflows/ci.yml
JordanMartinez: I was going to ask about that. Spago uses [this in their CI](https://github.com/purescript/spago/blob/master/.github/workflows/build.yml#L16-L21) likely because they were using Appveyor. I wasn't sure whether we needed it here or not.

I was thinking we should drop it.

ci/build.sh
hdgarrood: Oh sorry, never mind, this part is only creating source distributions, it's not compiling anything. In that case I'd suggest "Create source distributions" since "build" can give people the wrong impression.

.github/workflows/ci.yml
hdgarrood: Ah, I see. Yeah I think dropping it sounds good if it doesn't cause any problems.

JordanMartinez:
Looks like changing `STACK_ROOT` after Haskell and its tooling has been installed causes us to repeat work in the `stack --no-terminal setup` step. Specifically, we have to [redownload GHC](https://github.com/purescript/purescript/pull/3980/checks?check_run_id=1639959298#step:11:481) and [a few other things](https://github.com/purescript/purescript/pull/3980/checks?check_run_id=1639959298#step:11:926). This adds about 10 minutes to the build. (Windows: 10m 37s vs MacOS: 9s vs Linux: 4s)

When I originally asked whether we should just cache `C:\sr`, Harry said no because we thought it was an issue with setting `STACK_ROOT` (i.e. GitHub Actions wasn't working properly and we just needed to figure out the right way to set the variable). We now know that it's `haskell/setup/actions` that is causing the issue. 

So, I ask that question again. Should we just cache `C:\sr` on Windows and cache `~/.stack` on non-Windows OS?

hdgarrood:
It does now look like that might be the easiest thing to do here, yeah.

rhendric:
Ignore me if this already has blown your patience for fiddly CI configuration (which would be totally understandable), but I wonder if you could run the `actions/cache` step after the `haskell/actions/setup` step, and thus use `${{ steps.[some ID].outputs.stack-root }}` as the path in the `actions/cache` step instead of switching based on the OS? It doesn't look like `haskell/actions/setup` actually invokes `stack` or uses anything in `STACK_ROOT`; it just sets it up.

.github/workflows/ci.yml
f-f: Yeah we also have Appveyor running as part of the Spago CI, mostly as a smoke test to see if the new GH Actions CI was behaving as it should have. This has (surprisingly to me) caught quite a few inconsistencies and quirks actually.
As you can see it needs some special care to make the environments consistent, and we'll probably drop it from there in some time

JordanMartinez:
> but I wonder if you could run

Possibly. But I'd rather tackle that in a separate PR once we get this fully migrated.

.github/workflows/ci.yml
JordanMartinez: > This has (surprisingly to me) caught quite a few inconsistencies and quirks actually.

Such as?

JordanMartinez:
Looks like GHC is still being downloaded even with the OS-specific STACK_ROOT cache fix being applied. Not sure whether we should just accept this as "good enough" or continue troubleshooting it.

JordanMartinez:
How should deployments be handled?

hdgarrood:
The way releases work now is that if it's a tag being built, then the `bundle/build.sh` script will be run, and it should produce a pair of files depending on the OS, eg `sdist-test/bundle/linux64.tar.gz` and `sdist-test/bundle/linux64.sha`; both of those files should then be uploaded to the corresponding GitHub release (which should be created if it doesn't already exist). So ideally as close to that as possible.

f-f:
Spago's release process doesn't sound far from this, so [the workflow we have](https://github.com/purescript/spago/blob/master/.github/workflows/release.yml) might be a good place to start (it took a couple of releases to figure out the details, but now it's working as it should).

We also publish the npm package from the same release flow, which is something we might want to do here as well

JordanMartinez:
What if we didn't merge this until after v0.14.0 gets released?

hdgarrood:
I think that's probably a good option.

.github/workflows/ci.yml
f-f: I don't recall the details unfortunately, nor I'm able to find the relevant commits right now. I suspect most of that was resolved in the big initial effort. In any case, one of such occurrences was e.g. that I forgot the `--pedantic` flag in GH Actions, so a build-breaking change failed AppVeyor, while the Actions were happily green.

I'm not advocating for running an AppVeyor build of course, I'm just saying that since Travis and AppVeyor were consistent, keeping AppVeyor as a reference point helped keep such consistency when moving to Actions.

So for this reason here we might want to run Travis alongside the new Action workflow, at least for some time.

.github/workflows/release.yml
```diff
@@ -0,0 +1,169 @@
...
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ needs.draft_release.outputs.upload_url }}
+          asset_path: ./${{ runner.os }}.tar.gz
+          asset_name: ${{ runner.os }}.tar.gz
+          asset_content_type: text/plain
+
+  # npm_publish:
```
hdgarrood, line 149: The npm publish will have to be done manually because we require 2FA to publish to npm, so we can remove this.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+        if: runner.os == 'macOS'
+        run: |
+          brew install gnu-tar
+          echo "/usr/local/opt/gnu-tar/libexec/gnubin" >> $GITHUB_PATH
+
+      - uses: actions/checkout@v2
+
+      # Caching
+      #
+      # `haskell/actions/setup` will overwrite the STACK_ROOT variable
```
hdgarrood, line 36: I think we can remove this workaround since this was fixed upstream? https://github.com/haskell/actions/issues/23

.github/workflows/release.yml
```diff
@@ -0,0 +1,169 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          body: ""
+          draft: false
+          prerelease: false
+          commitish: master
+
+  build_release:
```
hdgarrood, line 36: I don't think we should duplicate all of this functionality across both the release.yml workflow and the ci.yml workflow. Can we instead set up the ci.yml workflow so that if the build succeeds and we are building a tag, we move on to a release.yml workflow which still has access to the compiler we've just built so that it only needs to contain the logic for creating and uploading a release?

.github/workflows/ci.yml
JordanMartinez: Yes, I believe so.

.github/workflows/release.yml
JordanMartinez: I'm not sure. I imagine it can? 
It's been a while since I've read through GH Actions' documentation. Moreover, I'm still not familiar with how the releasing works. Lastly, I wasn't sure how to translate the `.travis.yml` file into a corresponding one for GH Actions, mainly because I don't make releases for PureScript.

.github/workflows/release.yml
JordanMartinez: Good to know!

rhendric:
I'd like to help move this forward, if I can. @JordanMartinez, would that be help or hinderance? If help, would you prefer I start a new PR or add onto this one?

JordanMartinez:
Thanks! Go for it! You can build on this PR if you want, but feel free to
create a new one that doesn't include any of my commits here.

Let's get this migrated. :-)

PR #3989, opened 2021-01-17
Generate changelog and add PR template

JordanMartinez:
Fixes 3987


hdgarrood:
Can you remove the release candidates which don’t have any real release notes please? I don’t think they really add anything here; they only exist in GH releases so that we have somewhere to put the binaries.

hdgarrood (approved):

PR #3993, opened 2021-01-20
Only include direct dependencies in output for purs graph

### LINKED ISSUE #3991: purs graph includes the transitive closure of module imports
## Description

Note: I noticed this while implementing https://github.com/purescript/spago/pull/730 to detect when a project module directly imports a module from a transitive package dependency.

`purs graph` is including the transitive closure of the module imports for each module in the module graph.

I don't think that this was intended, based on the release note:

"Each property in the returned object has exactly two properties; path,
which is a string containing the file path relative to the directory where
the command was run, and depends, which is an array of the names of all
directly imported modules."

## To Reproduce

You can run this on the default Spago project (command `purs graph src/Main.purs '.spago/**/*.purs'`).

Here is what the output for `Main` looks like:

```
  "Main": {
    "path": "src/Main.purs",
    "depends": [
      "Prelude",
      "Control.Applicative",
      "Control.Apply",
      "Data.Functor",
      "Data.Function",
      "Control.Category",
      "Control.Semigroupoid",
      "Data.Boolean",
      "Data.Ord",
      "Data.Eq",
      "Data.HeytingAlgebra",
      "Data.Symbol",
      "Data.Unit",
      "Data.Show",
      "Record.Unsafe",
      "Type.Data.RowList",
      "Type.Data.Row",
      "Data.Void",
      "Data.Ordering",
      "Data.Semigroup",
      "Data.Ring",
      "Data.Semiring",
      "Control.Bind",
      "Control.Monad",
      "Data.BooleanAlgebra",
      "Data.Bounded",
      "Data.CommutativeRing",
      "Data.DivisionRing",
      "Data.EuclideanRing",
      "Data.Field",
      "Data.Monoid",
      "Data.NaturalTransformation",
      "Effect",
      "Effect.Console"
    ]
  },
```

However, the definition of `Main` is

```
module Main where

import Prelude

import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = do
  log "🍝"
```

## Expected behavior

`purs graph` only includes the directly imported modules for each module in the module graph.

Going back to the example on the default Spago project, I'd expect the entry for `Main` to be 

```
  "Main": {
    "path": "src/Main.purs",
    "depends": [
      "Prelude",
      "Effect",
      "Effect.Console"
    ]
  },
```

## PureScript version

0.13.8

colinwahl:
This intends to close #3991

**Description of the change**

Modifies `sortModules` to take an additional `DependencyDepth` argument, and modifies the `purs graph` workflow to only include direct dependencies in the module graph.

* `purs graph` now uses the `Direct` option
* all other workflows use `Transitive` and have the same behavior as they did previously

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ New features:
 
 Bugfixes:
 
+* Only include direct dependencies in the output for `purs graph` (#3993, @colinwahl)
+
+Fixes a bug where the transitive closure of a module's dependencies
+where included in the `depends` field in the output for `purs graph`.
+
 Other improvements:
 
 ## [v0.13.8](https://github.com/purescript/purescript/releases/tag/v0.13.8) - 2020-05-23
--- a/src/Language/PureScript/Docs/Convert/ReExports.hs
+++ b/src/Language/PureScript/Docs/Convert/ReExports.hs
@@ -68,7 +68,7 @@ updateReExports externs withPackage = execState action
 
   traversalOrder :: [P.ModuleName]
   traversalOrder =
-    case P.sortModules externsSignature externs of
+    case P.sortModules P.Transitive externsSignature externs of
       Right (es, _) -> map P.efModuleName es
       Left errs -> internalError $
         "failed to sortModules: " ++
--- a/src/Language/PureScript/Graph.hs
+++ b/src/Language/PureScript/Graph.hs
@@ -32,7 +32,7 @@ graph input = do
   Make.runMake Options.defaultOptions $ do
     ms <- CST.parseModulesFromFiles id moduleFiles
     let parsedModuleSig = Dependencies.moduleSignature . CST.resPartial
-    (_sorted, moduleGraph) <- Dependencies.sortModules (parsedModuleSig . snd) ms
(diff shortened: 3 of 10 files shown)


rhendric (approved):

kritzcreek:
Looks great, thanks! Would you mind adding an entry to the Changelog? I'm happy to merge afterwards.

f-f (approved):
:clap: :heart_eyes:

colinwahl:
@kritzcreek Done!

thomashoneyman (approved):

colinwahl:
A question for you all - do you think it would be worth merging this into the `0.13.x` branch as well, for possible inclusion in some patch release?

The motivation for that would be to continue working on `purs graph` related tooling without having to wait for `0.14` to be fully released/adopted.

f-f:
@colinwahl we are really close to the release of `0.14`, so there won't be any more patch releases in the `0.13` series.
For your change in Spago we can work with the unreleased version until the official release is out (i.e. you can use trunk locally, and I'll use it to test the changes, even if in CI we'll still have `0.13.8` for a while)

colinwahl:
@f-f Thanks! Didn't realize we were so close on `0.14` - it'll be great to have.

Sure, I can figure something out for the Spago change and we can get it ready for when 0.14 comes out.

PR #3969, opened 2020-12-19
Update version to 0.14.0-rc4

hdgarrood:

Issue #3973, opened 2020-12-26
Include kind information in HTML docs


hdgarrood:
## Summary

#3779 implemented polykinds and explicit kind annotations, but right now not all kind information is represented in HTML documentation.

## Motivation

HTML docs should ideally contain all type information which is observable from downstream code, which includes kinds. For example, the following two data types are not distinguishable in HTML documentation in `0.14.0-rc4`, except by their names:

```purescript
data MonoProxy :: Type -> Type
data MonoProxy a = MonoProxy

data PolyProxy :: forall k. k -> Type
data PolyProxy a = PolyProxy
```

## Proposal

The specifics of how we represent this information are still up for debate, but I think ideally the HTML docs shouldn't look too different from what's in the source files, so we could copy the kind annotations from the source files into the HTML docs where they exist.

JordanMartinez:
The additional information needed in the docs will break the `docs.json` format. Does that make this a breaking change?

Also, this change will affect `Prim` modules' docs creation. Right now, those docs are generated by creating a `TypeClassDictionary` and then creating a corresponding `Doc.Declaration` value for them (or something like that).

hdgarrood:
I think we should aim to make this a non-breaking change to the `docs.json` format. That is, `docs.json` files produced by older compilers should be understandable by the newest one, but `docs.json` files produced by the newest compiler need not be understandable by older compilers.

PR #3994, opened 2021-01-21
Update the changelog with release notes for the upcoming v0.14

kl0tl:
This pull request lists every pull request we merged since v0.13.6 minus those that were already released in v0.13.8 and the following that I took the liberty to omit: https://github.com/purescript/purescript/pull/3894, https://github.com/purescript/purescript/pull/3934, https://github.com/purescript/purescript/pull/3945, https://github.com/purescript/purescript/pull/3969, https://github.com/purescript/purescript/pull/3976.

I tried to motivate and detail most of the changes, except under the **Docs** and **Internal** sections where I only rewrote commit messages when necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,24 @@ Other improvements:
 
 Polymorphic kinds, based on the [Kind Inference for Datatypes](https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf) paper (#3779, #3831, #3929, @natefaubion)
 
+Kinds are to types what types are to terms (except that kinds are actually themselves types now) but whereas we have polymorphic types, kinds were monorphic.
+
+This means that we were not able to abstract over kinds, leading for instance to a proliferation of proxy types:
+
+```purescript
+data Proxy (a :: Type) = Proxy
+```
+
+```purescript
+data SProxy (a :: Symbol) = SProxy
+```
+
+```purescript
+data RProxy (row :: # Type) = RProxy
+```
+
+Now we can have a single proxy type, whose parameter has a polymorphic kind.
+
 #### Type In Type
 
 The type-checker now supports `TypeInType` (or `Type :: Type`), so the old `Kind` data type and namespace is now gone. Kinds and types are the same and exist in the same namespace.
@@ -97,7 +115,11 @@ Coercible constraints, based on the [Safe Zero-cost Coercions for Haskell](https
 
 Coercible constraints are solved according to the following rules:
 
-* Coercible constraints are _reflexive_, `Coercible a a` holds, _symmetric_, `Coercible a b` implies `Coercible b a`, and _transitive_, `Coercible a b` and `Coercible b c` imply `Coercible a c`.
+* _reflexivity_: `Coercible a a` holds.
+
+* _symmetry_: `Coercible a b` implies `Coercible b a`.
+
+* _transitivity_: `Coercible a b` and `Coercible b c` imply `Coercible a c`.
 
 * Newtypes can be freely wrapped and unwrapped when their constructor is in scope:
 
@@ -196,7 +218,18 @@ Annotated roles are compared against the roles inferred by the compiler so it is
 
 * Add compiler support for `Coercible` based `Newtype` (#3975, @fsoikin)
 
-We added a `Coercible` superclass to `Data.Newtype.Newtype` in order to implement `unwrap`, `wrap` and most newtype combinators with `coerce`. This is a breaking change for non derived instances because the `Newtype` class has no members anymore and can now only be implemented for representationally equal types (those satisfying the new superclass constraint).
+We added a `Coercible` superclass to `Data.Newtype.Newtype` in order to implement `unwrap`, `wrap` and most newtype combinators with `coerce` (see https://github.com/purescript/purescript-newtype/pull/22). This is only a breaking change for non derived instances because the `Newtype` class has no members anymore and can now only be implemented for representationally equal types (those satisfying the new superclass constraint).
+
+For example the instance for `newtype Additive a = Additive a` no longer implements `unwrap` and `wrap`:
+
+```diff
++instance newtypeAdditive :: Newtype (Additive a) a
+-instance newtypeAdditive :: Newtype (Additive a) a where
+-  wrap = Additive
+-  unwrap (Additive a) = a
+```
+
+Derived instances don't require any modifications.
 
 * Reform handling of quote characters in raw strings (#3961, @rhendric)
 
@@ -255,6 +288,7 @@ This shows a specific message when using negative litterals but `Data.Ring.negat
 This allows declarations such as
 
 ```purescript
+type Env = { port :: Int }
 newtype App a = App (ReaderT Env Aff a)
 derive newtype instance monadAskApp :: MonadAsk Env App
 ```
@@ -264,8 +298,6 @@ or
 ```purescript
 class (Monad m, MonadAsk Env m) <= MonadAskEnv m
 ```
-
-where `Env` is a type synonym.
 
 * Improve incremental rebuilds time for modules with large dependencies (#3899, @milesfrain)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,9 +39,9 @@ data RLProxy (row :: RowList) = RLProxy
 
 Now we can have a single proxy type, whose parameter has a polymorphic kind.
 
-#### Type In Type
+#### Type :: Type
 
-The type-checker now supports `TypeInType` (or `Type :: Type`), so the old `Kind` data type and namespace is now gone. Kinds and types are the same and exist in the same namespace.
+The old `Kind` data type and namespace is gone. Kinds and types are the same and exist in the same namespace.
 
 Previously one could do:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,10 +98,11 @@ In GHC, all signatures use the `type` prefix, but we reuse the same keyword as t
 
 It's better to be explicit about polymorphism by writing signatures. Since we don't really quantify over free type variables, it's also necessary in the case that two poly-kinded arguments must have the same kind. The compiler will warn about missing kind signatures when polymorphic kinds are inferred.
 
-Classes can have signatures too, but they must end with the new `Constraint` kind instead of `Type`. For example, `Prim.Row.Cons` now has the signature:
+Classes can have signatures too, but they must end with the new `Constraint` kind instead of `Type`. For example, here's the new definition of `Prim.Row.Cons`:
 
 ```purescript
 class Cons :: forall k. Symbol -> k -> Row k -> Row k -> Constraint
+class Cons label a tail row | label a tail -> row, label row -> a tail
 ```
 
 ### Safe zero-cost coercions
@@ -132,7 +133,7 @@ Here's the annotation we added to `Effect`:
 type role Effect representational
 ```
 
-Conversely, we might want to strengthen the roles of parameters with invariants invisible to the type system. Maps are the canonical example of this: the shape of their underlying tree rely on the `Ord` instance of their keys, but the `Ord` instance of a newtype may behave differently than the one of the wrapped type so it would be unsafe to allow coercions between `Map k1 a` and `Map k2 a` when `Coercible k1 k2` holds.
+Conversely, we might want to strengthen the roles of parameters with invariants invisible to the type system. Maps are the canonical example of this: the shape of their underlying tree rely on the `Ord` instance of their keys, but the `Ord` instance of a newtype may behave differently than the one of the wrapped type so it would be unsafe to allow coercions between `Map k1 a` and `Map k2 a`, even when `Coercible k1 k2` holds.
 
 In order to forbid such unsafe coercion we added a _nominal_ annotation to the first parameter of `Map`:
 
@@ -173,11 +174,11 @@ Meaning that raw strings can contain up to two successive quotes, any number of 
 
 * Unsupport bare negative literals as equational binders (#3956, @rhendric)
 
-It used to be possible to match on negative litterals, such as `-1`, but this prevented parsing matches on constructors aliased to `-`. The compiler will reject matches on _bare_ negative litterals, but they can still be matched by wrapping them in parentheses.
+It used to be possible to match on negative literals, such as `-1`, but this prevented parsing matches on constructors aliased to `-`. The compiler will reject matches on _bare_ negative literals, but they can still be matched by wrapping them in parentheses.
 
 * Forbid partial data constructors exports (#3872, @kl0tl)
 
-Exporting only some of the constructors of a type meant that changes internal to a module, such that adding or removing an unexported constructor, could cause unexhaustive pattern matches errors in downstream code. Partial explicit export lists will have to be completed with the missing constructors or replaced by implicit export lists.
+Exporting only some of the constructors of a type meant that changes internal to a module, such as adding or removing an unexported constructor, could cause unexhaustive pattern match errors in downstream code. Partial explicit export lists will have to be completed with the missing constructors or replaced by implicit export lists.
 
 * Print compile errors to stdout, progress messages to stderr (#3839, @JordanMartinez)
 
@@ -205,9 +206,17 @@ Allowing the compiler to be built against various versions of `language-javascri
 
 #### Improvements
 
+* Improves protocol errors from the IDE server (#3998, @kritzcreek)
+
+The IDE server now respond with more descriptive error messages when failing to parse a command. This should make it easier to contribute fixes to the various clients.
+
+* Extend ImportCompletion with declarationType (#3997, @i-am-the-slime)
+
+By exposing the declaration type (value, type, typeclass, etc.) downstream tooling can annotate imports with this info so users know what they are about to import. The info can also be mapped to a namespace filter to allow importing identifiers that appear more than once in a source file which throws an exception without such a filter.
+
 * Improve error message when `negate` isn't imported (#3952, @mhmdanas)
 
-This shows a specific message when using negative litterals but `Data.Ring.negate` is out of scope, similar to the messages shown when using do notation if `Control.Bind.bind` and `Control.Bind.discard` are out of scope.
+This shows a specific message when using negative literals but `Data.Ring.negate` is out of scope, similar to the messages shown when using do notation if `Control.Bind.bind` and `Control.Bind.discard` are out of scope.
 
 * Add source spans to `PartiallyAppliedSynonym` errors (#3951, @rhendric)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ Other improvements:
 
 Polymorphic kinds, based on the [Kind Inference for Datatypes](https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf) paper (#3779, #3831, #3929, @natefaubion)
 
-Kinds are to types what types are to terms (although kinds are themselves types now), but whereas we have polymorphic types, kinds were monomorphic.
+Just as types classify terms, kinds classify types. But while we have polymorphic types, kinds were previously monomorphic.
 
 This meant that we were not able to abstract over kinds, leading for instance to a proliferation of proxy types:
 
@@ -113,17 +113,17 @@ Coercible constraints, based on the [Safe Zero-cost Coercions for Haskell](https
 
 #### Roles
 
-Types parameters now have _roles_, which depend on how they affect the runtime representation of their type. There's three roles, from most to least restrictive:
+Types parameters now have _roles_, which depend on how they affect the runtime representation of their type. There are three roles, from most to least restrictive:
 
-* _nominal_ parameters are only coercible to themselves.
+* _nominal_ parameters can only be coerced to themselves.
 
-* _representational_ parameters are coercible when a Coercible constraint holds.
+* _representational_ parameters can only be coerced to each other when a Coercible constraint holds.
 
-* _phantom_ parameters are coercible to anything.
+* _phantom_ parameters can be coerced to anything.
 
 #### Role annotations
 
-The compiler infers _nominal_ roles for foreign data types, which is safe but can be too constraining sometimes. For example this prevents to coerce `Effect Age` to `Effect Int`, even though they actually have the same runtime representation.
+The compiler infers _nominal_ roles for foreign data types, which is safe but can be too constraining sometimes. For example this prevents the coercion of `Effect Age` to `Effect Int`, even though they have the same runtime representation.
 
 The roles of foreign data types can thus be loosened with explicit role annotations, similar to the [RoleAnnotations](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#role-annotations) GHC extension.
 
@@ -182,7 +182,7 @@ Exporting only some of the constructors of a type meant that changes internal to
 
 * Print compile errors to stdout, progress messages to stderr (#3839, @JordanMartinez)
 
-The compiler used to print errors to stderr, which is especially annoying when piping its output to another program with the `--json-errors` flag. Stderr is really for diagnostics about the execution of a program, whereas its expected output ought to be printed to stdout.
+Compiler errors and warnings arising from your code are now printed to stdout rather than stderr, and progress messages such as "Compiling Data.Array" are now printed to stderr rather than stdout. Warnings and errors arising from incorrect use of the CLI, such as specifying input files which don't exist or specifying globs which don't match any files, are still printed to stderr (as they were before). This change is useful when using the `--json-errors` flag, since you can now pipe JSON errors into other programs without having to perform awkward gymnastics such as `2>&1`.
 
 #### Fixes
 
@@ -210,7 +210,7 @@ Allowing the compiler to be built against various versions of `language-javascri
 
 The IDE server now respond with more descriptive error messages when failing to parse a command. This should make it easier to contribute fixes to the various clients.
 
-* Extend ImportCompletion with declarationType (#3997, @i-am-the-slime)
+* Extend IDE ImportCompletion with declarationType (#3997, @i-am-the-slime)
 
 By exposing the declaration type (value, type, typeclass, etc.) downstream tooling can annotate imports with this info so users know what they are about to import. The info can also be mapped to a namespace filter to allow importing identifiers that appear more than once in a source file which throws an exception without such a filter.
 
@@ -238,7 +238,7 @@ or
 class (Monad m, MonadAsk Env m) <= MonadAskEnv m
 ```
 
-* Improve incremental rebuilds time for modules with large dependencies (#3899, @milesfrain)
+* Improve incremental rebuild times for modules with large dependencies (#3899, @milesfrain)
 
 #### Other
 
@@ -254,7 +254,7 @@ This is the first step towards smarter incremental rebuilds, which could skip re
 
 * Deprecate constraints in foreign imports (#3829, @kl0tl)
 
-Constrained foreign imports leak instances dictionaries, hindering the compiler ability to optimize their representation. Manipulating dictionaries in foreign code should be avoided and foreign imports should accept the class members they need as additional arguments instead of being constrained.
+Constrained foreign imports leak instance dictionaries, hindering the compiler ability to optimize their representation. Manipulating dictionaries in foreign code should be avoided and foreign imports should accept the class members they need as additional arguments instead of being constrained.
 
 * Deprecate primes (the `'` character) in identifiers exported from foreign modules (#3792, @kl0tl)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,7 +109,7 @@ Coercible constraints, based on the [Safe Zero-cost Coercions for Haskell](https
 
 #### Roles
 
-Types parameters now have _roles_, which depend on how they affect the runtime representation of their type. There's three roles, from most to least restrictive: _nominal_, _representational_ and _phantom_.
+Types parameters now have _roles_, which depend on how they affect the runtime representation of their type. There's three roles, from most to least restrictive:
 
 * _nominal_ parameters are only coercible to themselves.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ This meant that we were not able to abstract over kinds, leading for instance to
 data Proxy (a :: Type) = Proxy
 data SProxy (a :: Symbol) = SProxy
 data RProxy (row :: # Type) = RProxy
+data RLProxy (row :: RowList) = RLProxy
 ```
 
 Now we can have a single proxy type, whose parameter has a polymorphic kind.
@@ -65,6 +66,8 @@ It is treated internally as:
 ```purescript
 data Foo
 ```
+
+Note that `foreign import data` declarations are not deprecated. They are still necessary to define types with kinds other than `Type` since constructors are not lifted as in GHC with [DataKinds](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#extension-DataKinds).
 
 Likewise, `kind` imports and exports are deprecated and treated the same as a type import or export.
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -10,12 +10,318 @@ New features:
...
+```
+
+In GHC, all signatures use the `type` prefix, but we reuse the same keyword as the subsequent declaration because we already have `foreign import data` (rather than `foreign import type`) and because it makes things align nicer. Signatures have the same rule as value-level signatures, so they must always be followed by the "real" declaration.
+
+It's better to be explicit about polymorphism by writing signatures. Since we don't really quantify over free type variables, it's also necessary in the case that two poly-kinded arguments must have the same kind. The compiler will warn about missing kind signatures when polymorphic kinds are inferred.
+
+Classes can have signatures too, but they must end with the new `Constraint` kind instead of `Type`. For example, `Prim.Row.Cons` now has the signature:
+
+```purescript
+class Cons :: forall k. Symbol -> k -> Row k -> Row k -> Constraint
```
thomashoneyman, line 81: Previously, `Cons` had [this type](https://pursuit.purescript.org/builtins/docs/Prim.Row#t:Cons):

```purs
class Cons (label :: Symbol) (a :: k) (tail :: Row k) (row :: Row k) | label a tail -> row, label row -> a tail
```

Now, it has this type and also a kind signature, correct?


```purs
class Cons :: forall k. Symbol -> k -> Row k -> Row k -> Constraint
class Cons (label :: Symbol) (a :: k) (tail :: Row k) (row :: Row k) | label a tail -> row, label row -> a tail
```

Perhaps it can also be written this way, now that the kind signature is present:

```purs
class Cons :: forall k. Symbol -> k -> Row k -> Row k -> Constraint
class Cons label a tail row | label a tail -> row, label row -> a tail
```

It might be useful to include this as a more complete example.

CHANGELOG.md
```diff
@@ -10,12 +10,318 @@ New features:
...
+Other improvements:
 
-Fixes a bug where the transitive closure of a module's dependencies
-where included in the `depends` field in the output for `purs graph`.
+## v0.14.0
 
-Other improvements:
+### Polykinds
+
+Polymorphic kinds, based on the [Kind Inference for Datatypes](https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf) paper (#3779, #3831, #3929, @natefaubion)
```
thomashoneyman, line 19: It may be useful to add an explanation or demonstration of why this feature is useful (perhaps @natefaubion has ideas?). For example, my (admittedly incomplete and limited) understanding is that we now have kind polymorphism + kind inference, which enables things like a single `Proxy` type instead of the many currently used for different kinds.

Can we explain what makes this feature a noteworthy addition to the language in a few sentences and/or a brief example?

CHANGELOG.md
```diff
@@ -10,12 +10,318 @@ New features:
...
+
+### Safe zero-cost coercions
+
+Coercible constraints, based on the [Safe Zero-cost Coercions for Haskell](https://www.microsoft.com/en-us/research/uploads/prod/2018/05/coercible-JFP.pdf) paper (#3351, #3810, #3896, #3873, #3860, #3905, #3893, #3909, #3931, #3906, #3881, #3878, #3937, #3930, #3955, #3927, @lunaris, @rhendric, @kl0tl, @hdgarrood)
+
+`Prim.Coerce.Coercible` is a new compiler-solved class, used to relate types with the same runtime representation. One can use `Safe.Coerce.coerce` (from the new [`safe-coerce`](https://github.com/purescript/purescript-safe-coerce) library) instead of `Unsafe.Coerce.unsafeCoerce` to safely turn a `a` into a `b` when `Coercible a b` holds.
+
+Coercible constraints are solved according to the following rules:
+
+* Coercible constraints are _reflexive_, `Coercible a a` holds, _symmetric_, `Coercible a b` implies `Coercible b a`, and _transitive_, `Coercible a b` and `Coercible b c` imply `Coercible a c`.
```
thomashoneyman, line 92: Perhaps this would be a little easier to parse on multiple lines? ie. coercible constraints satisfy these properties:

```markdown
 * _reflexive_:`Coercible a a` holds.
 * _symmetric_: `Coercible a b` implies `Coercible b a`.
 * _transitive_: `Coercible a b` and `Coercible b c` imply `Coercible a c`.
```

CHANGELOG.md
```diff
@@ -10,12 +10,318 @@ New features:
...
+
+Annotated roles are compared against the roles inferred by the compiler so it is not possible to compromise safety by ascribing too permissive roles, except for foreign types.
+
+### Other changes
+
+#### Breaking
+
+* Add compiler support for `Coercible` based `Newtype` (#3975, @fsoikin)
+
+We added a `Coercible` superclass to `Data.Newtype.Newtype` in order to implement `unwrap`, `wrap` and most newtype combinators with `coerce`. This is a breaking change for non derived instances because the `Newtype` class has no members anymore and can now only be implemented for representationally equal types (those satisfying the new superclass constraint).
```
thomashoneyman, line 191: Could we add a brief example of how to do this in a code snippet? For example, you'd no longer implement `wrap` and `unwrap` for the `Additive` newtype:

```diff
  newtype Additive a = Additive a

+ instance newtypeAdditive :: Newtype (Additive a) a
- instance newtypeAdditive :: Newtype (Additive a) a where
-   wrap = Additive
-   unwrap (Additive a) = a
```

Also, it's not clear to me what happens in the case of derived instances right now, like:

```purs
newtype Additive a = Additive a

derive instance newtypeAdditive :: Newtype (Additive a) _
```

Is the derived instance identical to just writing `instance`? Does it warn?

CHANGELOG.md
```diff
@@ -10,12 +10,318 @@ New features:
...
+* Add source spans to `PartiallyAppliedSynonym` errors (#3951, @rhendric)
+
+`PartiallyAppliedSynonym` errors were usually rethrown with the appropriate source span, but not when deriving instances. This annotates those errors with the source span of the partially applied synonyms themselves, which is more robust and accurate than rethrowing the error with an approximate source span.
+
+* Allow type synonyms in instances heads and superclass constraints (#3539, #3966, #3965, @garyb, @kl0tl)
+
+This allows declarations such as
+
+```purescript
+newtype App a = App (ReaderT Env Aff a)
```
thomashoneyman, line 250: Perhaps we could explicitly include the type synonym for `Env` because it's user-defined and to make the fact it's a type synonym apparent:

```purescript
type Env = { port :: Int }
newtype App a = App (ReaderT Env Aff a)
derive newtype instance monadAskApp :: MonadAsk Env App
```

thomashoneyman (commented):
This is looking great! Thanks for your work, @kl0tl. Just a few comments.

CHANGELOG.md
JordanMartinez: I mean, a single `Proxy` type is descriptive in itself. If one wanted to define their own kind, they needed to define their custom `Proxy` type, too. It's this whole "each kind must have it's own X" problem that polykinds solves. For example, we don't need a kind-specific type-level list anymore. We can have a kind-generic type-level list now.
I don't think that's all of polykinds, but it is a significant part.

thomashoneyman: I agree that describing the single `Proxy` type is enough of a motivating example on its own. I think we should say that explicitly in this section, briefly, to motivate why this feature is a useful addition to the language.

hdgarrood:
This looks great but I think there's too much detail in the polykinds and coercible sections. I think release notes should be easy to skim through, generally I'd say any one change should have no more than one or two short paragraphs. Perhaps we can move some of it into the documentation instead, and link there from here?

hdgarrood:
Hm, on a second reading I actually think the polykinds section is about right, but I do still think there's too much detail in the coercible section. There must be more to this than just "any one change should have no more than one or two short paragraphs" but I haven't yet worked out to express it. I think the detailed rules of when two data types are coercible should be omitted, and we should say what each of the roles mean (nominal, representational, phantom) but the detailed rules for how we infer them could also be omitted.

Perhaps a rule of thumb is that if there's a detail that you don't have to care about as a user of the feature in 90% of cases, it can be omitted. I think for most users, knowing what each of the roles mean is sufficient to have a fairly accurate intuition for what roles will be inferred, for example.

CHANGELOG.md
JordanMartinez: I agree with the above change.

CHANGELOG.md
JordanMartinez: I agree with this, too.

JordanMartinez:
Perhaps we should port some of the Coercible section to the migration guide I'm writing? Then, the release notes can stay smaller, but the detailed explanation can still be read by those who are unfamiliar with what's going on and what more examples.

hdgarrood:
I don't think the migration guide is the right place for this content either, actually. The stuff I think should move is stuff that doesn't matter to users of the feature in the majority of cases, so to me that feels like something that should be kept out of the migration guide as well.

thomashoneyman:
I believe @hdgarrood was suggesting the content be moved into the documentation / language reference.

hdgarrood:
Oh, I just realised I forgot the "repo" in my earlier comment! Yes, I did mean the documentation repo.

JordanMartinez:
Ah! That makes more sense.

CHANGELOG.md
kl0tl: The documentation on Pursuit is confusing. The following class declaration fails to compile because the kind variable isn’t bound:

```purs
class Cons (label :: Symbol) (a :: k) (tail :: Row k) (row :: Row k) | label a tail -> row, label row -> a tail
```

```
Error found:
in module Example
at Example.purs:3:36 - 3:37 (line 3, column 36 - line 3, column 37)

  Type variable k is undefined.

while inferring the kind of k
while checking that type k
  has kind Type
in type synonym Cons$Dict

```

It isn’t possible to bind the kind variable with a kind signature, meaning that the following still fails to compile:

```purs
class Cons :: forall k. Symbol -> k -> Row k -> Row k -> Constraint
class Cons (label :: Symbol) (a :: k) (tail :: Row k) (row :: Row k) | label a tail -> row, label row -> a tail
```

The kind annotations in the class head must be scrapped, as in your last snippet. The kind signature is not present in the generated documentation though:

Capture d’écran 2021-01-23 à 15 44 00

Should we display kind signatures in Pursuit when they contain polymorphic kinds? Something like 

Capture d’écran 2021-01-23 à 15 51 04

CHANGELOG.md
kl0tl: What do you think of the short explanation I added?

CHANGELOG.md
kl0tl: I added the example you suggested!

> Also, it's not clear to me what happens in the case of derived instances right now

Derived instances are unaffected. I tried to clarify this by adding a note about that in the changelog. Is that enough?

CHANGELOG.md
```diff
@@ -18,6 +18,24 @@ Other improvements:
 
 Polymorphic kinds, based on the [Kind Inference for Datatypes](https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf) paper (#3779, #3831, #3929, @natefaubion)
 
+Kinds are to types what types are to terms (except that kinds are actually themselves types now) but whereas we have polymorphic types, kinds were monorphic.
```
i-am-the-slime, line 21: I can't find how to insert a suggestion but I think you meant `monomorphic`. Also, I'd probably say `value` instead of `term`.

CHANGELOG.md
```diff
@@ -10,12 +10,350 @@ New features:
...
-Fixes a bug where the transitive closure of a module's dependencies
-where included in the `depends` field in the output for `purs graph`.
+## v0.14.0
 
-Other improvements:
+### Polykinds
+
+Polymorphic kinds, based on the [Kind Inference for Datatypes](https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf) paper (#3779, #3831, #3929, @natefaubion)
+
+Kinds are to types what types are to terms (except that kinds are actually themselves types now) but whereas we have polymorphic types, kinds were monorphic.
```
thomashoneyman, line 21:
```suggestion
Kinds are to types what types are to terms (although kinds are themselves types now), but whereas we have polymorphic types, kinds were monomorphic.
```

CHANGELOG.md
```diff
@@ -10,12 +10,350 @@ New features:
...
+
+This means that we were not able to abstract over kinds, leading for instance to a proliferation of proxy types:
```
thomashoneyman, line 23:
```suggestion
This meant that we were not able to abstract over kinds, leading for instance to a proliferation of proxy types:
```

CHANGELOG.md
```diff
@@ -10,12 +10,350 @@ New features:
...
+data Proxy (a :: Type) = Proxy
+```
+
+```purescript
+data SProxy (a :: Symbol) = SProxy
+```
+
+```purescript
+data RProxy (row :: # Type) = RProxy
+```
```
thomashoneyman, line 35: Perhaps these could be merged together?
```suggestion
data Proxy (a :: Type) = Proxy
data SProxy (a :: Symbol) = SProxy
data RProxy (row :: # Type) = RProxy
data RLProxy (rowlist :: RowList) = RLProxy
```

CHANGELOG.md
thomashoneyman: Ideally we could update Pursuit to display kind signatures containing polymorphic kinds, but I'm not sure it's an easy task 😅

CHANGELOG.md
thomashoneyman: I think it's great! 👍

CHANGELOG.md
kl0tl: Do you think we should list all the proxies, including RLProxy as you suggested, BProxy and OProxy, or just the most commons?

CHANGELOG.md
kl0tl: I corrected the typo 🙇

> I'd probably say `value` instead of `term`.

I usually prefer “value” to denote normal forms, so that term/type is orthogonal to value/expression, but I’ll apply your suggestion if others find that clearer too.

kl0tl:
I replaced the rules followed by the interaction solver and the section about role inference by a small section about roles.

What do you think of replacing the documentation for [`Prim.Coerce.Coercible`](https://pursuit.purescript.org/builtins/docs/Prim.Coerce) on Pursuit with those explanations instead of adding them to the documentation repo? I think the documentation we have now is based on the [2014](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/07/coercible.pdf) version of the paper and not its latest revision from [2016](https://www.microsoft.com/en-us/research/uploads/prod/2018/05/coercible-JFP.pdf), which does not explain Coercible as a class with compiler-generated instances.

hdgarrood:
Putting the detailed explanations into the documentation for Prim.Coerce.Coercible sounds good to me, although I feel like in any case we should have something in the documentation repo too.

CHANGELOG.md
hdgarrood: I think we should make updating the HTML docs generation to handle polykinds a priority post-0.14.0. I've usually regretted not being more of a stickler for this in the past when we have added new language features: any information which forms part of a module's API should be present in HTML docs, and I think in an ideal world we would actually not permit merging PRs which add language features until those features are represented properly in HTML docs.

CHANGELOG.md
JordanMartinez: I prefer `value` as well. `term` confuses me.

CHANGELOG.md
JordanMartinez: If our goal is to illustrate the idea, then these are likely enough. I don't think many people use `BProxy` or `OProxy` and if they do, then they are already familiar with type-level programming. Most people are probably familiar with `SProxy` and possibly with `RProxy`/`RLProxy` due to dealing with Record-related type class instances, so this list is likely enough.
If we want to be exhaustive, then adding those should be done.

thomashoneyman: I added in RLProxy because seeing so many proxies helps drive the point home :) I don’t think it has to be exhaustive.

CHANGELOG.md
```diff
@@ -10,12 +10,277 @@ New features:
...
+
+Previously one could do:
+
+```purescript
+foreign import kind Boolean
+foreign import data True :: Boolean
+foreign import data False :: Boolean
+```
+
+Where the kind `Boolean` and type `Boolean` were two different things. This is no longer the case. The `Prim` kind `Boolean` is now removed, and you can just use `Prim` type `Boolean` in the same way. This is a breaking change.
```
MonoidMusician, line 45: Can we add a note that we still need the `foreign import data`s because constructors are not automatically lifted? (i.e. no `DataKinds`)

CHANGELOG.md
hdgarrood: I’d agree that “term” is preferable here because it is more precise than “value”; a “term” is generally any expression, whereas a “value” generally means an expression which can’t be reduced. So `7` and `3+4` are both terms, but only the former is a value. In this case we do mean term, not value.

CHANGELOG.md
kl0tl: I added the RowList proxy.

CHANGELOG.md
kl0tl: I added a note about this after the announce of the deprecation of `foreign import kind` delcarations.

CHANGELOG.md
```diff
@@ -121,7 +124,7 @@ Here's the annotation we added to `Effect`:
 type role Effect representational
 ```
 
-Conversely, we might want to strengthen the roles of parameters with invariants invisible to the type system. Maps are the canonical example of this: the shape of their underlying tree rely on the `Ord` instance of their keys, but the `Ord` instance of a newtype may behave differently than the one of the wrapped type so it would be unsafe to allow coercions between `Map k1 a` and `Map k2 a`, even when `Coercible k1 k2` holds.
+Conversely, we might want to strengthen the roles of parameters with invariants invisible to the type system. Maps are the canonical example of this: the shape of their underlying tree rely on the `Ord` instance of their keys, but the `Ord` instance of a newtype may behave differently than the one of the wrapped type so it would be unsafe to allow coercions between `Map k1 a` and `Map k2 a` when `Coercible k1 k2` holds.
```
hdgarrood, line 127: I actually think this was clearer before. Now, it reads like it _would_ be safe to allow coercions between `Map k1 a` and `Map k2` a as long as there is no `Coercible k1 k2` instance.

CHANGELOG.md
```diff
@@ -10,12 +10,280 @@ New features:
...
+
+```
+'"""' '"'{0,2} ([^"]+ '"'{1,2})* [^"]* '"""'
+```
+
+Meaning that raw strings can contain up to two successive quotes, any number of times, but three successive quotes are not allowed inside.
+
+* Unsupport bare negative literals as equational binders (#3956, @rhendric)
+
+It used to be possible to match on negative litterals, such as `-1`, but this prevented parsing matches on constructors aliased to `-`. The compiler will reject matches on _bare_ negative litterals, but they can still be matched by wrapping them in parentheses.
```
rhendric, line 168: ‘litterals’ should be ‘literals’ (×2)

CHANGELOG.md
```diff
@@ -10,12 +10,280 @@ New features:
...
+
+* Forbid partial data constructors exports (#3872, @kl0tl)
+
+Exporting only some of the constructors of a type meant that changes internal to a module, such that adding or removing an unexported constructor, could cause unexhaustive pattern matches errors in downstream code. Partial explicit export lists will have to be completed with the missing constructors or replaced by implicit export lists.
```
rhendric, line 172: ‘such that’ should be ‘such as’
‘pattern matches errors’ should be ‘pattern match errors’

CHANGELOG.md
```diff
@@ -10,12 +10,280 @@ New features:
...
+
+* Pin language-javascript to a specific version (#3904, @hdgarrood)
+
+Allowing the compiler to be built against various versions of `language-javascript` meant that multiple builds of the same version of the compiler could accept different syntaxes for JavaScript foreign modules, depending on how they were built.
+
+#### Improvements
+
+* Improve error message when `negate` isn't imported (#3952, @mhmdanas)
+
+This shows a specific message when using negative litterals but `Data.Ring.negate` is out of scope, similar to the messages shown when using do notation if `Control.Bind.bind` and `Control.Bind.discard` are out of scope.
```
rhendric, line 202: ‘litterals’ again

CHANGELOG.md
kl0tl: I just found out you already opened https://github.com/purescript/purescript/issues/3973 some time ago. Let's tackle this once 0.14 is out then.

CHANGELOG.md
kl0tl: Let’s rollback this change then.

kl0tl:
Thank you Ryan for spotting all those typos :see_no_evil:

I included https://github.com/purescript/purescript/pull/3997 and https://github.com/purescript/purescript/pull/3998 in the release notes. I also opened https://github.com/purescript/documentation/pull/371 to document roles, role inference and role annotations in the language reference, and rewrote the documentation of Coercible according to the latest revision of the paper in https://github.com/purescript/purescript/pull/3999.

CHANGELOG.md
i-am-the-slime: Thanks, now I know the difference!

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+
+It used to be possible to match on negative literals, such as `-1`, but this prevented parsing matches on constructors aliased to `-`. The compiler will reject matches on _bare_ negative literals, but they can still be matched by wrapping them in parentheses.
+
+* Forbid partial data constructors exports (#3872, @kl0tl)
+
+Exporting only some of the constructors of a type meant that changes internal to a module, such as adding or removing an unexported constructor, could cause unexhaustive pattern match errors in downstream code. Partial explicit export lists will have to be completed with the missing constructors or replaced by implicit export lists.
+
+* Print compile errors to stdout, progress messages to stderr (#3839, @JordanMartinez)
+
+The compiler used to print errors to stderr, which is especially annoying when piping its output to another program with the `--json-errors` flag. Stderr is really for diagnostics about the execution of a program, whereas its expected output ought to be printed to stdout.
```
hdgarrood, line 177: I think this is a little opinionated, and it might be better focus a bit more on the "what" rather than the "why". When you say "its expected output ought to be printed to stdout", you haven't actually said where it _is_ printed, only where you think it should be printed. How about:
```suggestion
Compiler errors and warnings arising from your code are now printed to stdout rather than stderr, and progress messages such as "Compiling Data.Array" are now printed to stderr rather than stdout. Warnings and errors arising from incorrect use of the CLI, such as specifying input files which don't exist or specifying globs which don't match any files, are still printed to stderr (as they were before). This change is useful when using the `--json-errors` flag, since  you can now pipe JSON errors into other programs without having to perform awkward gymnastics such as `2>&1`.
```

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+
+Allowing the compiler to be built against various versions of `language-javascript` meant that multiple builds of the same version of the compiler could accept different syntaxes for JavaScript foreign modules, depending on how they were built.
+
+#### Improvements
+
+* Improves protocol errors from the IDE server (#3998, @kritzcreek)
+
+The IDE server now respond with more descriptive error messages when failing to parse a command. This should make it easier to contribute fixes to the various clients.
+
+* Extend ImportCompletion with declarationType (#3997, @i-am-the-slime)
```
hdgarrood, line 205: Might be worth making it clearer that this is to do with the IDE server?
```suggestion
* Extend IDE ImportCompletion with declarationType (#3997, @i-am-the-slime)
```

hdgarrood (approved):
Looks great! I'll leave it up to you what you'd like to do with my final two suggestions, and then please feel free to merge.

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
-Fixes a bug where the transitive closure of a module's dependencies
-where included in the `depends` field in the output for `purs graph`.
+## v0.14.0
 
-Other improvements:
+### Polykinds
+
+Polymorphic kinds, based on the [Kind Inference for Datatypes](https://richarde.dev/papers/2020/kind-inference/kind-inference.pdf) paper (#3779, #3831, #3929, @natefaubion)
+
+Kinds are to types what types are to terms (although kinds are themselves types now), but whereas we have polymorphic types, kinds were monomorphic.
```
thomashoneyman, line 21:
```suggestion
Conceptually, kinds are to types what types are to terms (although kinds are themselves types now). But while we have polymorphic types, kinds were previously monomorphic.
```

I found this sentence a bit difficult to read; I think it's a little clearer to separate the comment about kinds -> types and types -> terms from the comment about polymorphism.

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+data SProxy (a :: Symbol) = SProxy
+data RProxy (row :: # Type) = RProxy
+data RLProxy (row :: RowList) = RLProxy
+```
+
+Now we can have a single proxy type, whose parameter has a polymorphic kind.
+
+#### Type In Type
+
+The type-checker now supports `TypeInType` (or `Type :: Type`), so the old `Kind` data type and namespace is now gone. Kinds and types are the same and exist in the same namespace.
```
thomashoneyman, line 36: I wasn't personally familiar with `TypeInType`, which is -- as far as I know -- referring to the Haskell extension `-XTypeInType`. The [GHC documentation for the extension](https://downloads.haskell.org/~ghc/8.0.2/docs/html/users_guide/glasgow_exts.html#ghc-flag--XTypeInType) has a nice summary:

> Allow kinds to be as intricate as types, allowing explicit quantification over kind variables, higher-rank kinds, and the use of type synonyms and families in kinds, among other features.

Is this also true of our implementation (@natefaubion)? If so, is it worth adapting this sentence to help briefly note what the benefit of adding this feature was?

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+
+### Safe zero-cost coercions
+
+Coercible constraints, based on the [Safe Zero-cost Coercions for Haskell](https://www.microsoft.com/en-us/research/uploads/prod/2018/05/coercible-JFP.pdf) paper (#3351, #3810, #3896, #3873, #3860, #3905, #3893, #3909, #3931, #3906, #3881, #3878, #3937, #3930, #3955, #3927, @lunaris, @rhendric, @kl0tl, @hdgarrood)
+
+`Prim.Coerce.Coercible` is a new compiler-solved class, used to relate types with the same runtime representation. One can use `Safe.Coerce.coerce` (from the new [`safe-coerce`](https://github.com/purescript/purescript-safe-coerce) library) instead of `Unsafe.Coerce.unsafeCoerce` to safely turn a `a` into a `b` when `Coercible a b` holds.
+
+#### Roles
+
+Types parameters now have _roles_, which depend on how they affect the runtime representation of their type. There's three roles, from most to least restrictive:
```
thomashoneyman, line 108:
```suggestion
Types parameters now have _roles_, which depend on how they affect the runtime representation of their type. There are three roles, from most to least restrictive:
```

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+
+* _nominal_ parameters are only coercible to themselves.
+
+* _representational_ parameters are coercible when a Coercible constraint holds.
+
+* _phantom_ parameters are coercible to anything.
```
thomashoneyman, line 114: Not a firm suggestion, but I find this a little easier to parse when written in terms of the action you, the user, can take:

```suggestion
* _nominal_ parameters can only be coerced to themselves.

* _representational_ parameters can only be coerced to each other when a Coercible constraint holds.

* _phantom_ parameters can be coerced to anything.
```

But feel free to ignore this if you feel differently or my phrasing is less precise.

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+
+#### Role annotations
+
+The compiler infers _nominal_ roles for foreign data types, which is safe but can be too constraining sometimes. For example this prevents to coerce `Effect Age` to `Effect Int`, even though they actually have the same runtime representation.
```
thomashoneyman, line 118:
```suggestion
The compiler infers _nominal_ roles for foreign data types, which is safe but can be too constraining sometimes. For example this prevents the coercion of `Effect Age` to `Effect Int`, even though they have the same runtime representation.
```

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+derive newtype instance monadAskApp :: MonadAsk Env App
+```
+
+or
+
+```purescript
+class (Monad m, MonadAsk Env m) <= MonadAskEnv m
+```
+
+* Improve incremental rebuilds time for modules with large dependencies (#3899, @milesfrain)
```
thomashoneyman, line 233:
```suggestion
* Improve incremental rebuild times for modules with large dependencies (#3899, @milesfrain)
```

CHANGELOG.md
```diff
@@ -10,12 +10,289 @@ New features:
...
+
+* Have module re-exports appear in generated code (#3883, @citizengabe)
+
+This is the first step towards smarter incremental rebuilds, which could skip rebuilding downstream modules when the interface of a module did not change (see #3724).
+
+* Add a printer for CST modules (#3887, @kritzcreek)
+
+* Deprecate constraints in foreign imports (#3829, @kl0tl)
+
+Constrained foreign imports leak instances dictionaries, hindering the compiler ability to optimize their representation. Manipulating dictionaries in foreign code should be avoided and foreign imports should accept the class members they need as additional arguments instead of being constrained.
```
thomashoneyman, line 249:
```suggestion
Constrained foreign imports leak instance dictionaries, hindering the compiler's ability to optimize their representation. Manipulating dictionaries in foreign code should be avoided and foreign imports should accept the class members they need as additional arguments instead of being constrained.
```

thomashoneyman (approved):
I have a few minor grammatical edits and clarifications, but I think this can (and should) be merged once they are accepted or discussed, along with @hdgarrood's similar comment.

Once again, thank you for putting this together, @kl0tl -- it was clearly a time-consuming task in a significant release!

CHANGELOG.md
hdgarrood: I'm unsure about the addition of "Conceptually" here. Did you intend to imply that in actuality, kinds are _not_ to types what types are to terms? I don't think the statement "kinds are to types what types are to terms" needs any qualification.

rhendric: I took the ‘conceptually’ to indicate that the analogy is imperfect (it was a better analogy when kinds were a distinct universe above types the same way that types are a distinct universe above terms, but now kinds only exist as a relationship between types). I'm fine with that but maybe ‘approximately’ would be a clearer shade of paint to you?

hdgarrood: I think if you understand "kind" as "a type which, in this context, is being used to classify a type rather than a term", then I don't see this as an approximation: isn't the relationship the same, even if types are differ from terms in other ways?

rhendric: I think anyone who already understands ‘kind’ to mean that doesn't need this sentence to clarify what kinds are. :smile: I figure this is just a brief summary targeting someone who doesn't have a crisp idea of what a kind is and wants to understand how/if this change affects them, and that person might be confused unless they understand that the analogy only holds for ‘kind-of’ the relationship relative to ‘type-of’ but not for ‘kind’ the category of data manipulated by the compiler relative to ‘type’ the category etc.

hdgarrood: Right - I mean it's a definition, basically, isn't it? Ignore the "if you understand 'kind' as ..." part, then. I think whenever you say "A is to B as C is to D" and A/C and B/D aren't exactly the same things there's always going to be at least one way in which they differ, so I'd say it's reasonable to consider the fact that types and kinds live in the same namespace now (whereas types and terms don't) immaterial to the original statement. But maybe we can avoid this issue by just being a bit more explicit to start with? "Just as types classify terms, kinds classify types."?

CHANGELOG.md
kl0tl: I think this is all true, but `TypeInType` has become an alias for other extensions in 8.6.1 and was deprecated in 8.8.1, so perhaps it would be better not to mention it there? What would you think of renaming the title of this section to `Type :: Type` and drop the start of this sentence?

```suggestion
The old `Kind` data type and namespace is now gone. Kinds and types are the same and exist in the same namespace.
```

kl0tl:
I applied all your suggestions! The latest release of GHC broke my links to the DataKinds and RoleAnnotations extensions so I also updated them. I’d like some feedback on https://github.com/purescript/purescript/pull/3994#discussion_r577140653 and then I’ll merge this 🎉

CHANGELOG.md
JordanMartinez: For what it's worth, this change sounds reasonable to me.

hdgarrood: Sounds good to me.

kl0tl:
Thanks everyone for the reviews !

Issue #3083, opened 2017-09-20
purs ide: adding conflicting imports


coot:
`Data.Tuple.Nested` has both `/\ :: a -> b -> Tuple` and `/\ :: Type -> Type -> Type`.  Now adding import with
```
{"params":{"file":"/home/marcin/repos/gravicus/ospreyui/test/Test/DPIA/Components.purs","importCommand":{"importCommand":"addImport","identifier":"/\\"},"filter
s":[]},"command":"import"}
```
fails with an error
```
["{\"result\":\"Undecidable between type and dataconstructor\",\"resultType\":\"error\"}",""]
```

coot:
ping @kRITZCREEK

kRITZCREEK:
Well that error is "expected" in the sense that we really can't figure out which of the two you want, but now that we have namespace/declaration filters we can ask the editors to let the user decide which declaration they intend to import.

The Declaration filters might be a nicer fit for this, because I think we should prefer
`Did you mean the type class or data constructor Show?` 
over
`Did you mean the type or value level Show`?

kRITZCREEK:
Let's not close this yet. At this point the editors can't figure out the type of declaration they are looking at from the completions: https://github.com/purescript/purescript/blob/71fc308a53bf89678c1e1d4a242cc01561d556d1/src/Language/PureScript/Ide/Types.hs#L193-L201

I'm currently thinking about how to give the protocol more flexibility with respect to the level of detail of the returned information.

coot:
Good.  In vim I can do that by querying syntax highlighting.

i-am-the-slime:
Hey, I think we can close this 😄

PR #3997, opened 2021-01-23
Extend ImportCompletion with declarationType

i-am-the-slime:
**Description of the change**

This change extends the `Completion` datatype in the IDE. More specifically, it adds `declarationType` as a field to import completions. This type represents what type of declaration (value, type, typeclass, operator, type operator, etc.) the IDE will add to the file if applied. 

This addresses #3083 which currently prevents users from importing identifiers if there is more than one candidate with the same name in a file (for example the type-level `/\` and the value level `/\` from `Data.Tuple.Nested`). Today, attempting to import `/\` causes `psc-ide` to throw an exception. 
It's already possible to refine the import command with a namespace filter to make the import succeed, however users (and clients) currently lack the necessary information to make a decision about the namespace. This PR adds this information.

You can also [read a longer background story](https://gist.github.com/i-am-the-slime/408c4d5e32668e92d9224701f20c6bb9)

---

**Checklist:**

- [x] Added a test for the contribution

### DIFF
--- a/src/Language/PureScript/Ide/Imports.hs
+++ b/src/Language/PureScript/Ide/Imports.hs
@@ -316,9 +316,6 @@ addImportForIdentifier fp ident qual filters = do
           Right <$> addExplicitImport fp decl m1 qual
         -- Here we need the user to specify whether they wanted a 
         -- dataconstructor or a type
-
-        -- TODO: With the new namespace filter, this can actually be a
-        -- request for the user to specify which of the two was wanted.
         Nothing ->
           throwError (GeneralError "Undecidable between type and dataconstructor")
 
--- a/src/Language/PureScript/Ide/Imports.hs
+++ b/src/Language/PureScript/Ide/Imports.hs
@@ -314,8 +314,8 @@ addImportForIdentifier fp ident qual filters = do
         -- worst
         Just decl ->
           Right <$> addExplicitImport fp decl m1 qual
-        -- Here we need the user to specify whether he wanted a dataconstructor
-        -- or a type
+        -- Here we need the user to specify whether they wanted a 
+        -- dataconstructor or a type
 
         -- TODO: With the new namespace filter, this can actually be a
         -- request for the user to specify which of the two was wanted.
@@ -323,7 +323,7 @@ addImportForIdentifier fp ident qual filters = do
           throwError (GeneralError "Undecidable between type and dataconstructor")
 
     -- Multiple matches were found so we need to ask the user to clarify which
-    -- module he meant
+    -- module they meant
     xs ->
       pure (Left xs)
     where
--- a/src/Language/PureScript/Ide/Types.hs
+++ b/src/Language/PureScript/Ide/Types.hs
@@ -15,6 +15,7 @@ import           Data.Time.Clock (UTCTime)
 import qualified Data.Map.Lazy as M
 import qualified Language.PureScript as P
 import qualified Language.PureScript.Errors.JSON as P
+import           Language.PureScript.Ide.Filter.Declaration (DeclarationType(..))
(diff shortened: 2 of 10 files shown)


kritzcreek:
This is probably my favourite PR I've gotten so far :) Thank you for the detailed write-up.

I don't think it's a big deal, but we might want to think about how before this PR it doesn't matter if you pick the type or data constructor completion for a newtype, because you'll always get the import that includes the constructor. Once we merge this users will get the _choice_ which is nice, but it also means they have to _think_ about which of the two completions to pick. You can always automatically fix the compiler warning if you didn't need the constructor, but failing to import it when you need it produces a compiler error, that requires more thinking than just applying a suggestion.

Code-wise you can use the `DeclarationType` from here, instead of re-creating it:
https://github.com/purescript/purescript/blob/7ecc42669c69682996f2196ba2eef6c4ca827348/src/Language/PureScript/Ide/Filter/Declaration.hs#L13-L22

That way the clients could also just use that field and plug it directly into a declaration type filter, rather than moving back and forth between namespaces and declaration types: https://github.com/purescript/purescript/blob/7ecc42669c69682996f2196ba2eef6c4ca827348/psc-ide/PROTOCOL.md#declaration-type-filter

Also thanks for the note about how the ide servers "Can't parse your command" errors aren't helpful when you're hacking on it, I know how to fix that problem but just forgot about doing it.

src/Language/PureScript/Ide/Types.hs
```diff
@@ -245,6 +248,16 @@ identifierFromDeclarationRef = \case
   P.TypeOpRef _ op -> P.showOp op
   _ -> ""
 
+declarationType :: IdeDeclaration -> DeclarationType
```
i-am-the-slime, line 251: I had to move this out to avoid a cyclic import. This might not be the best place for it but it was my best guess.

kritzcreek: It's good enough :D The whole `Types` module shouldn't exist, but it's so convenient compared to figuring out all the intra dependencies.

i-am-the-slime:
Thanks for the praise for the PR, it makes me very happy! 

I changed the code to use the existing declaration type. It's much simpler that way!

Concerning potential confusion for imports, I completely agree with you that there's no need to confuse users when there's really one good option of imports for an identifier. However, since we have these edge cases I prefer exposing this right away here, and then deciding how to present the right information to the user in the language server. 

Something along the lines of: If there are duplicates where not specifying a type declaration filter will lead to an exception we add the information to the import label. Otherwise, we could merge multiple non-conflicting suggestions into just the one for the `Identifier(..)`. 

Having said that, we could use the same logic that we use for always importing `Identifier(..)` instead of `Identifier` to simply import both `(/\), type (/\)` and then remove unused imports (I guess the chance that a user might want both is high anyway). Somehow I feel it'd be better to allow the LS to make these decisions and keep psc-ide as dumb and literal as possible.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -314,16 +314,16 @@ addImportForIdentifier fp ident qual filters = do
         -- worst
         Just decl ->
           Right <$> addExplicitImport fp decl m1 qual
-        -- Here we need the user to specify whether he wanted a dataconstructor
-        -- or a type
+        -- Here we need the user to specify whether they wanted a 
+        -- dataconstructor or a type
 
         -- TODO: With the new namespace filter, this can actually be a
```
kritzcreek, line 320: We can remove this TODO now.

kritzcreek (changes requested):
Just one more nit, thanks again, this looks great!

kritzcreek (approved):

f-f (approved):

PR #3999, opened 2021-01-28
Don't explain Coercible as a class with compiler-generated instances

kl0tl:
The documentation we have for Coercible seems to be based on the [draft](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/07/coercible.pdf) of the _Safe Zero-cost Coercions for Haskell_ paper but its [latest revision](https://www.microsoft.com/en-us/research/uploads/prod/2018/05/coercible-JFP.pdf) moves away from compiler-generated instances. This is explained in the section **7.2 Prior version of Coercible** of the paper.

This pull request also adds links to the `safe-coerce` package on Pursuit and to the documentation about roles I’m proposing to add to the language reference in https://github.com/purescript/documentation/pull/371.

<details>
<summary>Rendered documentation</summary>

Prim Coerce Coercible

</details>


src/Language/PureScript/Docs/Prim.hs
```diff
@@ -377,70 +377,59 @@ coercible :: Declaration
...
+  , "common runtime representation so coercing between them would be unsafe."
   , ""
-  , "Note that, since the type variable `a` plays a role in `D`'s representation,"
-  , "we require that the types `a` and `a'` are themselves `Coercible`. However,"
-  , "since the variable `b` does not play a part in `D`'s representation (a type"
-  , "such as `b` is thus typically referred to as a \"phantom\" type), `b` and `b'`"
-  , "can differ arbitrarily."
+  , "* Terms of the same types (`data` or `newtype`) may be coercible, depending"
+  , "on the _roles_ of their parameters. Roles are documented in [the PureScript"
+  , "language reference](https://github.com/purescript/documentation/blob/master/language/Roles.md)."
```
rhendric, line 416: I think the ‘terms of different/the same types’ terminology is a little confusing. How about:

* Two types of kind `Type` may be coercible if the following two conditions are met:
  * Their heads must be the same. For example, `Maybe a` and `Either a b` don't satisfy this condition because `Maybe` and `Either` are different. Those types don't share a common runtime representation so coercing between them would be unsafe.
  * Their arguments may need to be identical or coercible depending on the _roles_ of the head's type parameters. Roles are documented in [the PureScript language reference](https://github.com/purescript/documentation/blob/master/language/Roles.md).

kl0tl: Oh right, I see how this can be confusing.

I didn’t manage to render a nested list in the generated documentation, so I slightly adapted your suggestion. Let me know what you think!

rhendric (approved):

thomashoneyman (approved):

src/Language/PureScript/Docs/Prim.hs
hdgarrood: I think this meaning of "head" is actually not quite correct. This is something I've only recently become aware of, so it's possible you even picked it up from me from something I've written elsewhere! But I now suspect it's based on a common misunderstanding of "instance head", in which the "head" of the instance
```
instance semigroupEffect :: Semigroup a => Semigroup (Effect a)
```
is understood to be `Effect` because it's the first type that appears; in fact the head of that instance is `Effect a`. See e.g. https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/instances.html#instance-declarations-and-resolution

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -377,70 +377,55 @@ coercible :: Declaration
...
-  , "to coerce under the type constructor (`data` or `newtype`). For example,"
-  , "given a definition:"
+  , "Newtype constructors have to be in scope to preserve abstraction. It's"
+  , "common to declare a newtype to encode some invariants (non emptiness of"
+  , "arrays with `Data.Array.NonEmpty.NonEmptyArray` for example), hide its"
+  , "constructor and export smart constructors instead. Without this restriction,"
+  , "the guarantees provided by such newtypes would be void."
   , ""
-  , "data D a b = D a"
+  , "* Two types of kind `Type` may be coercible, but only if their heads are"
```
hdgarrood, line 406: I'm a bit uneasy with this explanation for another reason beyond the way "head" is used here (which I would regard as incorrect, see my other comment), but also because I think it only works if you forget about newtypes. For example, `Int` and `Age` are both types of kind `Type` which are not the same,  but they are still coercible.

src/Language/PureScript/Docs/Prim.hs
rhendric: Huh! TIL, I guess. What *would* you use as a succinct way to describe, um, the leftmost type constructor in a type after any infix type constructors are normalized to prefix?

src/Language/PureScript/Docs/Prim.hs
rhendric: Maybe (with proper line reformatting of course) this?

```suggestion
  , "* If none of the above are applicable, two types of kind `Type` may be coercible, but only if their heads are"
```

hdgarrood: Ah good idea, I think that works!

src/Language/PureScript/Docs/Prim.hs
hdgarrood: I don't have a good answer for that, unfortunately. "The type being applied in a type application"?

rhendric: At a quick glance, the [coercible paper](https://www.microsoft.com/en-us/research/uploads/prod/2018/05/coercible-JFP.pdf) uses ‘head’ in a few places to mean this I think. So maybe we're okay after all?

hdgarrood: Ah, good point. I suppose if the authors of that paper are using it that way then it's hard to argue that it's incorrect. That is unfortunate though; it's awkward to have to clarify whether you're talking about a whole instance head or just an, uh, instance head head.

MonoidMusician: Yeah I’m fine with “head” here. I believe the original usage of the term is simply the leftmost applied thing in a term. It’s where the phrase “weak head normal form” comes from.

Maybe the instance head is generalized from this, once we got multiparam typeclasses? idk

MonoidMusician: Here’s a reference for this usage of head position: https://arxiv.org/pdf/1701.04918.pdf

src/Language/PureScript/Docs/Prim.hs
kl0tl: I applied your suggestion.

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -377,70 +377,56 @@ coercible :: Declaration
...
+  , "the guarantees provided by such newtypes would be void."
   , ""
-  , "data D a b = D a"
+  , "* If none of the above are applicable, two types of kind `Type` may be"
+  , "coercible, but only if their heads are the same. For example,"
+  , "`Coercible (Maybe a) (Either a b)` does not hold because `Maybe` and"
+  , "`Either` are different. Those types don't share a common runtime"
+  , "representation so coercing between them would be unsafe. In addition their"
+  , "arguments may need to be identical or coercible, depending on the _roles_"
+  , "_roles_ of the head's type parameters. Roles are documented in"
```
rhendric, line 412:
```suggestion
  , "of the head's type parameters. Roles are documented in"
```

kl0tl: Silly me 🤦 Thank you!

Issue #3983, opened 2021-01-08
Coercible solving can emit TypesDoNotUnify with unknown type


rhendric:
I think it should be considered a bug if the compiler ever asserts `Could not match type x with type y`, where `x` or `y` is an unknown type.

```purs
import Safe.Coerce (coerce)

newtype Foo a = Foo a

foo :: Foo { a :: Int, b :: Int }
foo = Foo { a: 0, b: 1 }

zero :: Int
zero = (coerce foo).a
```

```
  Could not match type

    ( b :: Int
    )         

  with type

    t0


while solving type class constraint

  Prim.Coerce.Coercible (Foo         
                           { a :: Int
                           , b :: Int
                           }         
                        )            
                        { a :: Int   
                        | t0         
                        }            

while applying a function coerce
  of type Coercible @Type t1 t2 => t1 -> t2
  to argument foo
while checking that expression coerce foo
  has type { a :: Int
           | t0      
           }         
while checking type of property accessor (coerce foo).a
in value declaration zero

where t0 is an unknown type
      t1 is an unknown type
      t2 is an unknown type
```

I would expect the compiler to treat the above case similarly to this one:

```purs
import Data.Tuple (Tuple(..), fst)
import Safe.Coerce (coerce)

newtype Foo a = Foo a

foo :: Foo (Tuple Int Int)
foo = Foo (Tuple 0 1)

zero :: Int
zero = fst (coerce foo)
```

```
  No type class instance was found for

    Prim.Coerce.Coercible Int
                          t1

  The instance head contains unknown type variables. Consider adding a type annotation.

while solving type class constraint

  Prim.Coerce.Coercible (Foo (Tuple Int Int))
                        (Tuple Int t1)

while applying a function coerce
  of type Coercible @Type t2 t3 => t2 -> t3
  to argument foo
while checking that expression coerce foo
  has type Tuple t0 t1
in value declaration zero

where t0 is an unknown type
      t1 is an unknown type
      t2 is an unknown type
      t3 is an unknown type
```

(It seems like it might be possible for coercible solving to actually allow one or both of these cases? Given that `Coercible` instances don't really exist, why should it matter if there are irrelevant unknowns in instance heads? But if that's a bad idea, the second error is certainly more helpful than the first.)

kl0tl:
We should be able to solve unknowns during canonicalization in the interaction solver, but your second example fails too in GHC with a similar error:

```
Example.hs:11:13: error:
[...]

PR #4000, opened 2021-01-28
Don't mention unknowns in unification errors when coercing misaligned rows

kl0tl:
This pull request improves error messages when coercing unaligned rows as suggested by @rhendric in https://github.com/purescript/purescript/issues/3983, but does not try to solve unknowns encountered while solving Coercible constraints.

Doing so would be ill-advised should we ever add support for nominal equalities to the interaction solver given my current (lack of) understanding of how functional dependencies affect unification.

I’d be happy to revisit this later though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,7 +99,7 @@ class Cons label a tail row | label a tail -> row, label row -> a tail
 
 ### Safe zero-cost coercions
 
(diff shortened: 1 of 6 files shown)


rhendric (approved):
This does seem like the right thing to do for now.

hdgarrood (approved):

PR #4006, opened 2021-02-02
Fix exponential collapsing of BindingGroups

radrow:
**Description of the change**

### In short:
When you dfs through a tree and launch the same dfs on each node then you go exponential.

### In long:
I have spotted that function definitions which are big enough (see the added testcase) tend to blow up the memory and take extreme amount of time to handle. 

The issue was triggered particularly in the `collapseBindingGroups` as it was running a traversal through the whole AST of each declaration that on every `let` was calling the `collapseBindingGroups` which was running a traversal through the whole AST of each declaration that on every `let` was calling the `collapseBindingGroups` which was running a traversal through the whole AST of each declaration that on every `let` was calling the `collapseBindingGroups` which was running a traversal(...)

More clearly, `collapseBindingGroups` used to run `collapseBindingGroupsForValue` on each AST node which was running `collapseBindingGroups` on every `let` expression. This was effectively forcing all inner `let`s to be visited twice which, by the fact, that each of these duplicated evaluations was running next two ones, was amplifying the problem leading to exponential complexity over the depth of a tree.

-----------------------

Since the purpose of this function is to flatten the binding groups of the declarations across the whole AST, I changed the folding function for `everywhereOnValues` to simply flatten a single layer of declarations. It works perfectly as going deeper is already handled by `everywhereOnValues`.

I have prepared a testcase that fails on the `master` branch but is successfully solved after the fixes. 

-----------------------

Work on this PR was funded by [aeternity](https://aeternity.com)


hdgarrood:
Great work! This looks sensible to me. Does this new test fail without this fix, at least on machines where you don’t have an enormous amount of memory?

radrow:
I tested it on my PC and it failed on `master` with 32GB of RAM and i7 9k CPU

hdgarrood (approved):
Nice, thanks again

rhendric (approved):
Nice work!

kl0tl:
The new test takes ~3 minutes to compile on my machine. This isn’t noticeable with the absurdly long build times we get with Travis these days but this will be more noticeable when we’ll switch to GitHub actions. Is that ok or should we try to exhibit a test case that compiles faster?

natefaubion:
> The new test takes ~3 minutes to compile on my machine. This isn’t noticeable with the absurdly long build times we get with Travis these days but this will be more noticeable when we’ll switch to GitHub actions. Is that ok or should we try to exhibit a test case that compiles faster?

I think it would be preferable, least of all because this makes running tests in development quite an ordeal. I wouldn't care as much if this was _only_ run in CI (or with a CLI flag).

hdgarrood:
I suppose since this was exponential before, the smallest program that's big enough to be likely to cause an OOM could potentially be a fair bit smaller than this test.

radrow:
Okay, I have shrinked the test case to just two "patterns" as it still breaks the build on my PC on master and passes within 16 seconds on my branch. Should be acceptable?

hdgarrood:
I think 16 seconds is a bit of a drag, but just about manageable. Ideally we'd have a slightly more sophisticated mechanism for detecting performance regressions than just writing test cases that are likely to fail with OOM errors if things regress, but that's a whole project in of itself. Any objections to merging this as-is?

PR #4025, opened 2021-03-11
Use type annotation hint only when needed

### LINKED ISSUE #4024: Proposal: Avoid hinting to add a type annotation when it wouldn't help
## Summary
The compiler will display the `consider adding a type annotation` hint when unknown variables can be determined by known parameters. 

I will just copy-paste @hdgarrood [comment](https://functionalprogramming.slack.com/archives/C717K38CE/p1615431980095000?thread_ts=1615401053.085500&cid=C717K38CE) from the slack channel. 

> The problem I was pointing out was that the compiler shouldn't suggest adding a type annotation in a case where that could never fix the issue. In this case, it's saying that there's no instance Generic String t0, and then suggesting adding a type annotation. However, the class Generic a rep has a functional dependency a -> rep, which means that the compiler only needs to know what a is to be able to locate an instance (if one existed). Since the compiler can't solve Generic String t0, the only conclusion is that there is no Generic String _ instance
## Motivation

The hint is actually not being helpful. The compiler suggests adding a type annotation in a case where that won't fix the issue. 

## Proposal

remove the `consider adding a type annotation` hint if unknown variables would be determined by known parameters

## Examples

```purescript
newtype Result =
    Result {
      _word :: Array (Word String String String String String String String)
    }
    
    
instance showResult :: Show Result where
  show = genericShow
  
derive instance genericResult :: Generic Result _

newtype Word a b c d e f g =
    Word {
      _wordWord :: a
    , _oldword :: b
    , _strokes :: c
    , _wordPinyin :: d
    , _radicals :: e
    , _wordExplanation :: f
    , _more :: g
    }

instance showWord :: (Generic a ra, Show a, Generic b rb, Show b, Generic c rc, Show c, Generic d rd, Show d, Generic e re, Show e, Generic f rf, Show f, Generic g rg, Show g) => Show (Word a b c d e f g) where
  show = genericShow
derive instance genericWord :: (Generic a ra, Generic b rb, Generic c rc, Generic d rd, Generic e re, Generic f rf, Generic g rg) => Generic (Word a b c d e f g) _
```

outputs:
```

    Data.Generic.Rep.Generic String
                             t0

```

rhendric:
This commit changes the criteria for when the message

    The instance head contains unknown type variables. Consider adding a
    type annotation.

appears in a NoInstanceFound error. Instead of using it whenever an
unknown type variable appears in an instance head, now the compiler
adds this hint only when every covering set of type arguments of the
instance head contains an unknown type variable. (Covering sets are
determined by functional dependencies, if present; otherwise, the entire
set of type arguments is the one and only covering set.)

The rationale is that if any covering set is already completely known,
then the availability of an appropriate instance will not be changed by
adding more type annotations.

**Description of the change**

See above. Fixes #4024.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -103,7 +103,7 @@ data SimpleErrorMessage
   | KindsDoNotUnify SourceType SourceType
   | ConstrainedTypeUnified SourceType SourceType
   | OverlappingInstances (Qualified (ProperName 'ClassName)) [SourceType] [Qualified Ident]
-  | NoInstanceFound SourceConstraint
+  | NoInstanceFound SourceConstraint Bool
   | AmbiguousTypeVariables SourceType [Int]
   | UnknownClass (Qualified (ProperName 'ClassName))
   | PossiblyInfiniteInstance (Qualified (ProperName 'ClassName)) [SourceType]
@@ -448,7 +448,7 @@ onTypesInErrorMessageM f (ErrorMessage hints simple) = ErrorMessage <$> traverse
   gSimple (ConstrainedTypeUnified t1 t2) = ConstrainedTypeUnified <$> f t1 <*> f t2
   gSimple (ExprDoesNotHaveType e t) = ExprDoesNotHaveType e <$> f t
   gSimple (InvalidInstanceHead t) = InvalidInstanceHead <$> f t
-  gSimple (NoInstanceFound con) = NoInstanceFound <$> overConstraintArgs (traverse f) con
+  gSimple (NoInstanceFound con unks) = NoInstanceFound <$> overConstraintArgs (traverse f) con <*> pure unks
   gSimple (AmbiguousTypeVariables t us) = AmbiguousTypeVariables <$> f t <*> pure us
   gSimple (OverlappingInstances cl ts insts) = OverlappingInstances cl <$> traverse f ts <*> pure insts
   gSimple (PossiblyInfiniteInstance cl ts) = PossiblyInfiniteInstance cl <$> traverse f ts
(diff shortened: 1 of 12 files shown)


src/Language/PureScript/Errors.hs
```diff
@@ -103,7 +103,7 @@ data SimpleErrorMessage
   | KindsDoNotUnify SourceType SourceType
   | ConstrainedTypeUnified SourceType SourceType
   | OverlappingInstances (Qualified (ProperName 'ClassName)) [SourceType] [Qualified Ident]
-  | NoInstanceFound SourceConstraint
+  | NoInstanceFound SourceConstraint Bool
```
hdgarrood, line 106: Would you mind adding a comment here to explain what the Bool means?

hdgarrood (approved):
Looks great, thanks!

MonoidMusician:
Sorry, I haven’t had time to properly review it, but I did want to make sure it takes into account the bug I fixed in #3721 and looks at the closure of fundeps?

rhendric:
Quite likely I'm missing something, but what do you mean by the closure in this case? There's only one class involved in a `NoInstanceFound` error, right? Whereas it looks to me like #3721 was about the interaction of multiple classes?

MonoidMusician:
Ah I see, you’re probably right. If other constraints in context could have solved for the variables, they would have made progress already before we got to this error?

rhendric:
I assume that's right. Or at least, whatever progress they've made already is insufficient to prevent the error being thrown. If hypothetical subsequent progress could have solved for variables, that's a solver issue, and this PR only changes error reporting.

PR #4032, opened 2021-03-19
Update HiddenConstructorsGeneric.out

i-am-the-slime:
Adds a forgotten `an`.


hdgarrood:
This change should go into `src/Language/PureScript/Errors.hs` instead - this is just a test golden file, whose job is to ensure that we don't change error messages accidentally.

i-am-the-slime:
Ah, sorry, I just saw the error message and naïvely searched for the String in Github!

PR #4033, opened 2021-03-20
Apply precedence rules to operator sections

### LINKED ISSUE #3981: should be a bug - operator section suprising behavior
This behavior was voted as a  [bug ](https://discourse.purescript.org/t/bug-or-not-a-bug/2006)

```
 > ( _ * 4 + 1)  50
 250
 > ( 50 * 4 + 1)
 201
```

The recommendation from the forum seem to be to disallow  operator section if it have more than a simple operand 

For example 
The below is allowed 
```
> ( 3 * _ ) 3
9
```
but not 
```
> ( 3 * 3 * _ ) 3
Error found:
at <internal>:0:0 - 0:0 (line 0, column 0 - line 0, column 0)

  An anonymous function argument appears in an invalid context.
```

rhendric:
This commit moves operator section desugaring from the beginning of the
desugaring pipeline to the rebracketing phase. This allows the compiler
to correctly detect whether an operator section is well-formed based on
the precedence rules in effect. (In the process, this commit also
refines the positioning of error messages associated with incorrect
anonymous arguments.)

Previously, `(_ * 4 + 1)` would desugar to `\x -> x * (4 + 1)`, even
though `*` has higher precedence than `+`. Conversely, `(3 * 2 + _)`
would not compile, even though `*` has higher precedence than `+`. These
bugs have now been fixed; `(_ * 4 + 1)` is an error, and `(3 * 2 + _)`
desugars to `\x -> 3 * 2 + x`.

If you have code that relied on the old behavior, add an extra pair of
parentheses around the expression in the section.

**Description of the change**

See above. Fixes #3981.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/breaking_fix-3981.md
+++ b/CHANGELOG.d/breaking_fix-3981.md
@@ -1,0 +1,10 @@
+* Apply precedence rules to operator sections
+
+  Previously, `(_ * 4 + 1)` would desugar to `\x -> x * (4 + 1)`, even
+  though `*` has higher precedence than `+`. Conversely, `(3 * 2 + _)`
+  would not compile, even though `*` has higher precedence than `+`. These
+  bugs have now been fixed; `(_ * 4 + 1)` is an error, and `(3 * 2 + _)`
+  desugars to `\x -> 3 * 2 + x`.
+
+  If you have code that relied on the old behavior, add an extra pair of
+  parentheses around the expression in the section.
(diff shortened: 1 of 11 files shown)


hdgarrood:
Looks great! This certainly seems like the ideal approach too. The only issue is that I think we will want to do at least one non-breaking release before our next breaking one, and I'm not particularly keen on maintaining release branches, so perhaps we could leave this one open for a bit until we know that our next release is going to be a breaking one?

thomashoneyman:
@hdgarrood Should we create a `release-0.15` label as we did for the core and contrib libraries for 0.14, to mark issues and pull requests we want to merge in the next breaking release? That made it quite a bit easier to walk through all the various things we wanted to merge.

rhendric:
Sure, I'm fine with leaving this open until we start preparing for 0.15.

natefaubion (approved):

PR #4013, opened 2021-02-13
Updates for GHC 8.10.4

kritzcreek:
Contains the necessary changes to let us compile on GHC 8.10.4

Builds on #3943.

Don't worry about merging this anytime soon, I'm just making sure we don't fall too far behind. Maybe the next person comes along and makes a GHC9 PR on top of this one.

PR #4042, opened 2021-04-03
Remove unused Data.Foldable.foldr import

kl0tl:
**Description of the change**

We added an import for `Data.Foldable.foldr` to `Language.PureScript.CoreImp.Optimizer.TCO` in #3958. The build passed (https://travis-ci.org/github/purescript/purescript/builds/763606846) but subsequent builds started to fail with “The import of ‘Data.Foldable’ is redundant” since we merged #4013. The latter upgraded `base-compat` and GHC but `base-compat-0.10.5` already exported `Prelude.Compat.foldr`, and according to GHC documentation `-Wunused-imports` was already implied by -Wall in v8.6.5 so I really don’t understand what’s going on 🤷‍♂️ 

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ Other improvements:
 
   The previous `logo.png` was not legible against a dark background (#4001).
 
+Internal:
+
+* Remove unused Data.Foldable.foldr import (#4042, @kl0tl)
+
 ## v0.14.0
 
 ### Polykinds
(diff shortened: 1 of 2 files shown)


JordanMartinez (approved):

rhendric (approved):
Should this even be mentioned in the changelog at all? Otherwise, :shipit:.

JordanMartinez:
Yeah, I think it should. It provides accountability as to what changed. Besides, it's easier to remove unneeded things before making a release than it is to remember what to add.

hdgarrood:
I don’t think we need accountability here and I’m not sure a strict policy of having every change noted in the changelog is a good way of achieving that anyway. We already have the git history for cases where we want to be able to look at every change and who made it don’t we? It probably doesn’t matter now but I think this could have been left out.

kl0tl:
Duly noted! We can clean the changelog when releasing the next minor version.

Issue #4039, opened 2021-04-01
Purs bundle drops `var` but keeps `const` and `let`


vapaj:
## Description

When bundling a foreign module, the bundler drops global variables if they are not directly used in an exported function. Interestingly, it seems to depend on how the variable is defined.

Let's say we have a simple program:
```purescript
foreign import foo :: Effect Int

main :: Effect Unit
main = do
  x <- foo
  log $ show x
```

And a foreign module:
```javascript
var a = 1;
exports.foo = function() {
  return a + 1;
}
```
Everything works as expected. However, if instead of exporting an anonymous function, we would use a named function which we set the value of `foo` to:

```javascript
var a = 1;
function foo_() {
  return a + 1;
}
exports.foo = foo_;
```

When bundling the app and running it with node, the following error occurs:
```
$ spago build
$ purs bundle "output/*/*.js" -m Main --main Main -o index.js
$ node .

index.js:47
    return a + 1;
    ^

ReferenceError: a is not defined
```

However, if we declare `a` with `const` or `let`, we still seem to have `a` in place:
```javascript
const a = 1;
function foo_() {
  return a + 1;
}
exports.foo = foo_;

```
```
$ spago build
$ purs bundle "output/*/*.js" -m Main --main Main -o index.js
$ node .

2
```

## To Reproduce

Test with code provided, then `spago build`, `purs bundle "output/*/*.js" -m Main --main Main -o index.js` (which is basically `spago bundle-app`) and run with `node .`

## Expected behavior

Global variable should not be dropped.

## PureScript version

0.13.8

kl0tl:
The actual issue is that `purs bundle` only considers variable declarations: function declarations are ignored so the dependency on `a` is never discovered. Replacing the function declaration by a variable declaration works as expected:

```diff
 var a = 1;
-function foo_() {
+var foo_ = function () {
   return a + 1;
 };
 exports.foo = foo_;
```

Let and const declarations are preserved because they aren’t considered as bindings at all.

JordanMartinez:
Should this issue be closed?

rhendric:
It's not clear (to me) whether DCE in foreign modules is an officially-supported feature (as opposed to merely a side effect of DCE in PureScript-generated code); if so, then I'd say keep this open.

f-f:
[...]

Issue #3259, opened 2018-02-28
"second level" value declarations are removed during purs bundle


safareli:
`src/Main.js`:
``` js
"use strict";

var baz = { "biz": true };

function gez(x) {
  return baz[x] || false;
};

exports.foo = gez;
```

`src/Main.purs`:
``` purs
module Main where

foreign import foo :: String -> Number
```

after `pulp build`, `output/Main/index.js`:
``` js
// Generated by purs version 0.11.6
"use strict";
var $foreign = require("./foreign");
module.exports = {
    foo: $foreign.foo
};
```

(`output/Main/foreign.js` is same as `src/Main.js`, as expected)

But `purs bundle "output/**/*.js" --module Main` outputs:
``` js
// Generated by purs bundle 0.11.7
var PS = {};
(function(exports) {
    "use strict";           

  function gez(x) {
    return baz[x] || false;
  };

  exports.foo = gez
})(PS["Main"] = PS["Main"] || {});
(function(exports) {
  // Generated by purs version 0.11.6
  "use strict";
  var $foreign = PS["Main"];
  exports["foo"] = $foreign.foo;
})(PS["Main"] = PS["Main"] || {});
```

Problem is there is no `var baz = { "biz": true };` any more I guess this is known bug but still opening with this small reproducible case.

I guess we might need to do a thing which determines that `gez` should be contained in output, recursively or something.

hdgarrood:
I think #2290 and #3082 are the same issue. It is indeed useful to have a simple repro case though.

hdgarrood:
I actually think we should do as little static analysis of JS code as we possibly can; if there's any doubt, the code should stay in. The onus should be on library authors to write their FFI code in such a way that the dead code elimination can safely eliminate everything which is uneeded, I think.

safareli:
In that case rules of dead code elimination algorithm should be documented so it's well understood by users

natefaubion:
I'm pretty sure the DCE on foreign modules is the same routine as that on normal PS modules, so it only really looks at forms generated by the compiler. I wonder if you can break code like your example in normal PS land by using Data.Record.

safareli:
Just to note if I had  `exports.foo = function(x) { return gez(x) };` definition of `baz` is still lost. but `gez` is still present in bundled code.

safareli:
@natefaubion I tried following,

`src/Lib.purs`:
``` purs
module Lib (foo) where

baz :: { t :: Boolean, f :: Boolean }
[...]

Issue #2290, opened 2016-08-27
Purescript bundle optimization issue.


codygman:
If I'm understanding everything correctly, some transform that psc-bundle does eliminates a needed variable declaration. Minimum reproduction and some notes at the repo below:

https://github.com/codygman/purescript-optimize-issue

I also created an issue in bodil/pulp#227 since turning on optimization with pulp browserify ends up being affected by this runtime error bug due to a necessary var declaration being optimized away.

paf31:
Is this running Browserify and then psc-bundle, or the other way round?

codygman:
Browserify is first I think, but I'm not sure. If I'm wrong about this being a psc-bundle issue that means there are some transforms defined somewhere that pulp passes to browserify I think, however I couldn't find them.

paf31:
I would say it's wrong to apply Browserify then psc-bundle, so maybe it is a Pulp issue then.

hdgarrood:
`pulp browserify -O` does psc-bundle first, then browserify.

spicydonuts:
Also seeing this with `pulp build`, with or without `-O`.  I just added a larger FFI file, but there are others that work fine and do similar things.  Here's the top of the file that's breaking:

```js
"use strict"

var keyPrefix = "Jane.Cache.Event:"
var localStorageTested = false
var localStorage = window.localStorage
var unprocessedEventQueue = []
var readyEventQueue = []
var completedEventQueue = []
var push = Array.prototype.push.apply
var batchUnprocessedCount = 10
var batchUnprocessedInterval = 100
var batchReadyCount = 1000
var batchReadyInterval = 5000
var batchCompletedCount = 10
var batchCompletedInterval = 100
var postEvents = null

exports._enqueueEvent = function (event) {
  unprocessedEventQueue.push(event)
}
```

And here's that file in the bundle:
```
    "use strict";                       
  var unprocessedEventQueue = []
  var push = Array.prototype.push.apply
  var postEvents = null

  exports._enqueueEvent = function (event) {
    unprocessedEventQueue.push(event)
  }
```

It's consistent and only breaks on this file, which makes me think it has something to do with the contents of this file specifically..

michaelficarra:
@spicydonuts Can you post the whole file? It's hard to tell whether a variable is used without being able to see all potential usages. Also, the line `var push = Array.prototype.push.apply` is concerning for a number of reasons.

0. That's not the line as listed in the input.
[...]

PR #4044, opened 2021-04-04
Add support for bundling function declarations

kl0tl:
**Description of the change**

`purs bundle` does not consider the dependencies of function declarations because the compiler always emit variable declarations. This has unfortunate consequences when optimizing foreign modules:

```js
"use strict";

var x = 0;

function foo() {
  return x;
}

exports.foo = foo;
```

If the dependency of `foo` on `x` isn’t discovered, `x` appears unused and is removed from the bundle!

Fix https://github.com/purescript/purescript/issues/4039, fix https://github.com/purescript/purescript/issues/3259 and fix https://github.com/purescript/purescript/issues/2290.

---

**Checklist:**


f-f (approved):

rhendric (approved):

rhendric:
This PR spans releases, so it'll need a rebase and the changelog entry will need to be moved to the current Unreleased section.

hdgarrood:
Is the changelog entry the only thing preventing merging here now?

kl0tl:
No more!

hdgarrood (approved):

JordanMartinez:
Just FYI. This seems to have introduced a regression. See purescript-contrib/purescript-arraybuffer/issues/34. Should I open a new issue? Or is there one already?

PR #4040, opened 2021-04-03
Instantiate polymorphic kinds when unwrapping newtypes while solving Coercible constraints

kl0tl:
**Description of the change**

@thomashoneyman shared on Slack that the compiler is unable to derive a `Newtype` instance for newtypes applying a variable to a row:

```purescript
newtype X r f = X (r ( field :: f Int ))
derive instance newtypeX :: Newtype (X r f) _
```

```
Error found:
in module Main
at src/Main.purs:6:1 - 6:46 (line 6, column 1 - line 6, column 46)

  No type class instance was found for
                                              
    Prim.Coerce.Coercible (r1                 
                             ( field :: f2 Int
                             )                
                          )                   
                          (r1                 
                             ( field :: f2 Int
                             )                
                          )                   
                                              
  The instance head contains unknown type variables. Consider adding a type annotation.

while solving type class constraint
                                            
  Prim.Coerce.Coercible (X @t3 r1 f2)       
                        (r1                 
                           ( field :: f2 Int
                           )                
                        )                   
                           
```

The issue is easier to understand with an empty row, and compiling with `--verbose-errors`:

```purescript
newtype X r = X (r ())
derive instance newtypeX :: Newtype (X r) _
```

```
Error found:
in module Main
at src/Main.purs:6:1 - 6:44 (line 6, column 1 - line 6, column 44)

  No type class instance was found for
                                       
    Prim.Coerce.Coercible (r1 (() @k)) 
                          (r1 (() @t0))

  The instance head contains unknown type variables. Consider adding a type annotation.

while solving type class constraint
                                     
  Prim.Coerce.Coercible (X @t0 r1)   
                        (r1 (() @t0))
```

Now we see the types differ in how their polymorphic kind is instantiated!

The unknown kind applied to the newtype in the original constraint is replaced by a `k` variable, bound in the (inferred) newtype kind signature. This pull request instantiates those variables when unwrapping newtypes, so that `Coercible (X @t0 r1) (r1 (() @t0))` yields `Coercible (r1 (() @t0)) (r1 (() @t0))` as expected.

---

**Checklist:**

- [x] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [ ] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [x] Added a test for the contribution (if applicable)

PR #4045, opened 2021-04-04
Make stack pedantic by default

rhendric:
**Description of the change**

The purpose of this change is to prevent regrets like [this one](https://github.com/purescript/purescript/pull/4040#issuecomment-813037783). Things that cause the build to fail in CI ought to cause the build to fail locally.

---

**Checklist:**

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -47,5 +47,5 @@ tar -xzf sdist-test/lib/purescript-cst/purescript-cst-*.tar.gz -C sdist-test/lib
 $STACK sdist . --tar-dir sdist-test;
 tar -xzf sdist-test/purescript-*.tar.gz -C sdist-test --strip-components=1
 pushd sdist-test
-$STACK build --pedantic $STACK_OPTS
+$STACK build $STACK_OPTS
 popd
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -52,6 +52,7 @@ flag release
   default: False
 
 common defaults
+  ghc-options: -Wall
   default-language: Haskell2010
   default-extensions:
     BangPatterns
--- a/stack.yaml
+++ b/stack.yaml
@@ -6,7 +6,7 @@ packages:
 - lib/purescript-cst
 ghc-options:
   # Build with advanced optimizations enabled by default
-  "$locals": -O2 -Wall -Werror
+  "$locals": -O2 -Werror
 extra-deps:
 - language-javascript-0.7.0.0
 nix:
--- a/stack.yaml
+++ b/stack.yaml
@@ -6,7 +6,7 @@ packages:
 - lib/purescript-cst
 ghc-options:
   # Build with advanced optimizations enabled by default
-  "$locals": -O2
+  "$locals": -O2 -Wall -Werror
 extra-deps:
 - language-javascript-0.7.0.0
 nix:


JordanMartinez (dismissed):

hdgarrood:
I don't have a strong preference here, but I think there's a strong argument to be made for having it both ways - it can definitely be annoying to have e.g. an unused Debug.Trace import fail the build locally while you're still exploring. As long as it's relatively easy to switch back locally I'm fine with changing this. I suppose you'd do that by editing stack.yaml? It's a little inconvenient that that will make the working tree dirty but it's not the end of the world.

In any case I think we should omit `-Wall` from stack.yaml and specify it in package.yaml instead (if it isn't there already, but I imagine it is), so that the set of warnings we want to see is defined in one place and will be the same however the compiler is being built, even if someone is building with a different stack.yaml or none at all (eg via cabal-install). I'd argue that it's just `-Werror` which should go here, if we want pedantic by default.

rhendric:
I assume you mean the .cabal files? Yeah, that makes sense. purescript-ast.cabal and purescript-cst.cabal already have `-Wall`, but purescript.cabal doesn't.

I would argue that `-Werror` belongs in those files as well. Aside from the general consistency argument (between CI and local, between stack and cabal-install), there's a purist position that `-Werror` is conceptually akin to a language extension, in that it alters the language—changes the set of programs accepted by the compiler. If we manage the language extensions in the `.cabal` files, that's also the appropriate place for `-Werror`.

In either case, I think the recommended way to get temporary relief should be to throw a `{-# OPTIONS_GHC -Wwarn #-}` at the top of any files being tinkered with. Stack (I don't know if this is true of Cabal as well, but I suspect it is) won't automatically rebuild files when `-Werror` is added or removed, which can lead to the following pit trap: you compile without `-Werror`, get three warnings in three different files, fix two of them but miss the third, and on the next compilation see no warnings (because the third file was unchanged, it isn't revisited by the compiler). Then you rebuild with `-Werror`, see no errors, push, and CI fails. The way out of this pit trap is to do a full clean and rebuild before pushing, but that's unnecessarily time-consuming and it's an extra thing to remember to do. Adding the `-Wwarn` as a file header pragma ensures that removing it counts as a change to the file, which catches any remaining warnings without forcing a full rebuild.

hdgarrood:
Oh sorry yes, I did mean purescript.cabal. I'm pretty sure that putting `-Werror` in the cabal file is not an option though, because IIRC Hackage would reject it. If you put -Werror in your cabal file, you're asking downstream users to compile with `-Werror` enabled when they depend on your package, which can obviously cause problems if the package starts generating benign warnings with newer versions of dependent libraries or compilers (I think that's the reason, at least).

hdgarrood:
The `{-# OPTIONS_GHC -Wwarn #-}` strategy sounds promising! Have you used it in practice already?

rhendric:
> IIRC Hackage would reject it

Oh... huh. That is an excellent, if annoying, point.

Okay, so `-Wall` in all the .cabals and `-Werror` in stack.yaml it is, then!

> Have you used it in practice already?

I tested it in the workflow I described, but I haven't needed to use it in anger yet.

hdgarrood (approved):

PR #4047, opened 2021-04-06
Fix kinded declaration reordering in desugaring

### LINKED ISSUE #4035: Issue involving type-aliases, polykinds, and externs
## Description

A certain type leads to a compilation error when imported from a different file but not if it appears in the same file. The error is 
```
  Type synonym Other.Id is partially applied.
```

## To Reproduce

Try to compile [this project](https://github.com/i-am-the-slime/psbug/blob/master/src/Main.purs) and then either inline the type alias of `Alias` to make it compile or use the commented out local definition of `Id`.

Note that everything works without the kind (?) annotations.

## Expected behavior

It's possible to import the `Id` type instead of having to define it locally.

## Additional context

This is the repro main file:

```purescript
module Main where

import Other (Id)

-- type Id :: forall k. k -> k
-- type Id a = a

type Alias = Int

type Wrapped :: forall k. (Type -> k) -> Row k -> Row k
type Wrapped f r = (key :: f Alias | r)

type Unwrapped :: Row Type -> Row Type
type Unwrapped r = Wrapped Id r
```

## PureScript version

0.14.0

### LINKED ISSUE #4038: Kind annotated class causes "Unknown super class in TypeClassDeclaration" crash when used in a superclass implication
## Description

Kind annotated class causes "Unknown super class in TypeClassDeclaration" crash when used in a superclass implication. When the kind annotation is removed, it works

Crash report:
```
Compiling Main
purs: An internal error occurred during compilation: Unknown super class in TypeClassDeclaration
  error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-ast-0.1.0.0-Gtmwmu4N43KCxo7Qi9BiwQ:Language.PureScript.Crash
  internalError, called at src/Language/PureScript/TypeChecker.hs:176:20 in purescript-0.14.0-FNUkHeAp48Q5tFoT4dYs4J:Language.PureScript.TypeChecker
```

## To Reproduce

Compile the code with purs v0.14.0
```haskell
module Main where

class A :: Constraint -- remove this line and it works
class A

class B :: Constraint
class A <= B
```

## Expected behavior

Successful Compilation

## PureScript version

0.14.0

rhendric:
The type checker assumes that type class and type synonym definitions
always precede their use (circular references are not allowed between
these declarations). This constraint needs to be enforced by the
`createBindingGroups` phase in desugaring, regardless of whether the
declaration has a kind or is being referenced by a kind declaration.
Failing to do so resulted in one issue where type synonyms were being
expanded out of order with surprising consequences, and another where
a superclass was not known when its subclass was being checked.

**Description of the change**

Fixes #4035, fixes #4038.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Sugar/BindingGroups.hs
+++ b/src/Language/PureScript/Sugar/BindingGroups.hs
@@ -16,7 +16,7 @@ import Control.Monad ((<=<), guard)
 import Control.Monad.Error.Class (MonadError(..))
 
 import Data.Graph
-import Data.List (intersect)
+import Data.List (intersect, (\\))
 import Data.Foldable (find)
 import Data.Maybe (isJust, mapMaybe)
 import qualified Data.List.NonEmpty as NEL
@@ -73,21 +73,29 @@ createBindingGroups moduleName = mapM f <=< handleDecls
   handleDecls :: [Declaration] -> m [Declaration]
   handleDecls ds = do
     let values = mapMaybe (fmap (fmap extractGuardedExpr) . getValueDeclaration) ds
-        kindDecls = fmap (,VertexKindSignature) $ filter (\a -> isKindDecl a || isExternDataDecl a) ds
-        dataDecls = fmap (,VertexDefinition) $ filter (\a -> isDataDecl a || isTypeSynonymDecl a || isTypeClassDecl a) ds
+        kindDecls = fmap (,VertexKindSignature) $ filter isKindDecl ds
+        dataDecls = fmap (,VertexDefinition) $ filter (\a -> isDataDecl a || isExternDataDecl a || isTypeSynonymDecl a || isTypeClassDecl a) ds
         kindSigs = fmap (declTypeName . fst) kindDecls
         typeSyns = fmap declTypeName $ filter isTypeSynonymDecl ds
+        nonTypeSynKindSigs = kindSigs \\ typeSyns
         allDecls = kindDecls ++ dataDecls
         allProperNames = fmap (declTypeName . fst) allDecls
         mkVert (d, vty) =
           let names = usedTypeNames moduleName d `intersect` allProperNames
               name = declTypeName d
-              -- If a dependency has a kind signature, than that's all we need to depend on, except
-              -- in the case that we are defining a kind signature and using a type synonym. In order
-              -- to expand the type synonym, we must depend on the synonym declaration itself.
+              -- If a dependency of a kind signature has a kind signature, than that's all we need to
+              -- depend on, except in the case that we are using a type synonym. In order to expand
+              -- the type synonym, we must depend on the synonym declaration itself.
+              --
+              -- Arguably, type declarations (as opposed to just kind signatures) could also depend
+              -- on kind signatures when present. Attempting this caused one known issue (#4038); the
+              -- type checker might not expect type declarations not to be preceded or grouped by
+              -- their actual dependencies in all cases. But in principle, if done carefully, this
+              -- approach could be used to reduce the number or size of data binding group cycles.
+              -- (It's critical that kind signatures not appear in groups, which is why they get
+              -- special treatment.)
               vtype n
-                | vty == VertexKindSignature && n `elem` typeSyns = VertexDefinition
-                | n `elem` kindSigs = VertexKindSignature
+                | vty == VertexKindSignature && n `elem` nonTypeSynKindSigs = VertexKindSignature
                 | otherwise = VertexDefinition
               deps = fmap (\n -> (n, vtype n)) names
               self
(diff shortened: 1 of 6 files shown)


natefaubion (approved):

rhendric:
I realized a little late that #4038 was basically the same issue, so I've extended this PR a bit to cover that case as well.

src/Language/PureScript/Sugar/BindingGroups.hs
```diff
@@ -76,18 +76,18 @@ createBindingGroups moduleName = mapM f <=< handleDecls
         kindDecls = fmap (,VertexKindSignature) $ filter (\a -> isKindDecl a || isExternDataDecl a) ds
         dataDecls = fmap (,VertexDefinition) $ filter (\a -> isDataDecl a || isTypeSynonymDecl a || isTypeClassDecl a) ds
         kindSigs = fmap (declTypeName . fst) kindDecls
-        typeSyns = fmap declTypeName $ filter isTypeSynonymDecl ds
+        kindedData = (fmap declTypeName $ filter (\a -> isDataDecl a || isExternDataDecl a) ds) `intersect` kindSigs
```
natefaubion, line 79: This change may be an oversimplification. I don't think data types or instances need to depend on a class definition. I think this may affect recursive binding groups that contains classes and polykinds. I don't know if we have a test for that or not.

natefaubion: I think it would be worth noting why super classes depend on the class definition, and if it's truly necessary.

rhendric: Honestly, I don't understand why this can't be *further* simplified to always depend on the definition for everything referenced (which will itself continue to depend on its own kind declaration, if present). I assumed there was a reason why depending on a data type's kind declaration instead of the data type itself would be necessary, but we don't seem to have any tests exercising that—nothing fails when I change that logic, aside from a trivial unknown renumbering in one of the golden outputs.

For recursive binding groups containing classes and polykinds, we have [tests/purs/passing/PolykindBindingGroup2.purs](https://github.com/purescript/purescript/blob/bc898395c2da556c8ec6b7c702a3dff6d2f1bb9c/tests/purs/passing/PolykindBindingGroup2.purs); is that the sort of thing you had in mind, but maybe with some kind declarations thrown in?

At some level, none of this reordering is ‘truly necessary’; the type checker itself could always have logic to scan all of the declarations or do its own topsorting or whatever, right? Shouldn't the burden of proof instead be on the desugaring to generally put things in order of reference or in recursive binding groups and deviate from that only when truly necessary?

natefaubion: It's required to support polymorphic recursion (which requires polymorphic kind signatures). The type checker for value declarations essentially does the same thing by splitting binding groups into things which have explicit signatures and things that don't. We don't do that for top-level kind signatures because there's no desugaring to remove the top-level signature like there is for values and their type declarations. It may be possible to have a different architecture for handling explicit kind signatures in these cases, but you will have to make fairly significant changes to the environment, I think.

rhendric: I fully believe you but I'd like to have an example to add to the tests that fails if the compiler does the naive thing I proposed. I've tried a few cases of polymorphic recursion but none of them have fit the bill yet; could I trouble you to suggest a specific example?

natefaubion: > It's required to support polymorphic recursion (which requires polymorphic kind signatures). 

This is not correct, and I'm conflating two different things from what I remember from the original paper. For reference https://dl.acm.org/doi/pdf/10.1145/3371121 6.1 is the brief mention of this. For the case of polymorphic recursion, it would be enough for kind signatures to be in scope at the point checking the other declarations, which would be equivalent to what the value-level type checker is doing. The only thing that is important here is to catch and disallow cyclical dependencies in kind signatures.

I _think_ that as long as:
* Kind signatures are always in scope at the point of checking the declaration (ie, the declaration depends on it's kind signature).
* Cycles in kind signatures are disallowed.

Then alternative orderings should be acceptable. Depending on only the kind signature if available does conceptually simplify things, since it will break up recursive groups.

rhendric: That's very helpful, thanks!

So yeah, it seems like it's not important to break up groups containing definitions—these groups would have existed before kind signatures, and the type checker must have handled them—but it is very important to make sure that kind signatures don't get involved in those groups. So the algorithm I'm pushing now has kind signatures depend on other kind signatures if present (except for type synonyms) and everything else depend on definitions. I see where Xie et al. recommend depending on kind signatures more broadly, but their type system doesn't include type classes, and they don't seem to make a strong argument for the benefit of that approach aside from keeping kind signatures out of groups. I'm adding a comment describing this as a potential future approach if needed and if the type checker can be made to cooperate.

natefaubion (approved):

hdgarrood (approved):

PR #4050, opened 2021-04-10
Fix wildly off kind unification positions

### LINKED ISSUE #4019: Error span showing in the wrong place for kind mismatch when calling an instance member
## Description

The error span is wildly out - I don't really have a theory for what exactly it is doing as it manifests in interesting ways. The example should illustrate further:

## To Reproduce

``` purescript
import Prelude

newtype Indexed ∷ forall k1 k2 k3. (k1 → Type) → k2 → k3 → k1 → Type
newtype Indexed m x y a = Indexed (m a)

class IxFunctor ∷ ∀ ix. (ix → ix → Type → Type) → Constraint
class IxFunctor f where
  imap ∷ ∀ a b x y. (a → b) → f x y a → f x y b

instance ixFunctorIndexed ∷ Functor m ⇒ IxFunctor (Indexed m) where
  imap f (Indexed ma) = Indexed (map f ma)

foreign import data K1 :: Type
foreign import data K2 :: Type

foreign import data D1 :: K1
foreign import data D2 :: K2

foo :: Indexed Array D1 D2 Int
foo = Indexed [1]

bar :: Indexed Array D1 D2 Int
bar = imap identity foo
```
```
  Could not match kind

    K1

  with kind

    K2

  while trying to match type Indexed @Type @K1 @K2 Array
    with type t0
  while checking that expression foo
    has type t0 t1 t2 t3
  in value declaration bar

  where t0 is an unknown type
        t3 is an unknown type
        t1 is an unknown type
        t2 is an unknown type
```

In this instance, the `foo` type declaration is being marked as the problem, but `bar` is the issue (illustrated by the fact that commenting out `bar` resolves the issue also). If you move `bar` into a different file the error span will still point at `foo` in this file, which is pretty neat. 

If you try type-annotating the value:

``` purescript
bar :: Indexed Array D1 D2 Int
bar = imap identity (Indexed [1] :: Indexed Array D1 D2 Int)
```
The span will be here:
```
  25  bar = imap identity (Indexed [1] :: Indexed Array D1 D2 Int)
                                          ^^^^^^^^^^^^^
```

## Expected behavior

I would expect the usage of the value `foo` to be the span for the error for the first case, and I guess the whole annotated expression for the second case.

natefaubion:
**Description of the change**

Fixes #4019

This scrubs the kind-checker's error position context when invoking
kind unification from type-checking routines. When the kind-checker
errors, we want to preserve the context of the unification constraint.
Unfortunately, we don't have that position on hand since it's maintained
implicitly as part of the type-checker stack, so this is a somewhat
indirect solution.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/TypeChecker/Kinds.hs
+++ b/src/Language/PureScript/TypeChecker/Kinds.hs
@@ -358,9 +358,6 @@ unifyKinds = unifyKindsWithFailure $ \w1 w2 ->
   throwError
     . errorMessage''' (fst . getAnnForType <$> [w1, w2])
     $ KindsDoNotUnify w1 w2
-  -- throwError
-  --   . errorMessage
-  --   $ KindsDoNotUnify w1 w2
 
 -- | Does not attach positions to the error node, instead relies on the
 -- | local position context. This is useful when invoking kind unification
--- a/src/Language/PureScript/TypeChecker/Kinds.hs
+++ b/src/Language/PureScript/TypeChecker/Kinds.hs
@@ -10,6 +10,7 @@ module Language.PureScript.TypeChecker.Kinds
   , kindOfClass
   , kindsOfAll
   , unifyKinds
+  , unifyKinds'
   , subsumesKind
   , instantiateKind
   , checkKind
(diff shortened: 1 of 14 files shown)


hdgarrood (approved):

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -357,6 +358,22 @@ unifyKinds = unifyKindsWithFailure $ \w1 w2 ->
   throwError
     . errorMessage''' (fst . getAnnForType <$> [w1, w2])
     $ KindsDoNotUnify w1 w2
+  -- throwError
+  --   . errorMessage
+  --   $ KindsDoNotUnify w1 w2
```
rhendric, line 363: This wasn't meant to be left here, was it?

natefaubion: Nope! 😄

garyb (approved):

Issue #3914, opened 2020-08-01
purs fails to update corefn output


coot:
## Description

When I modify code and run `purs compile --codegn corefn` the `corefn.json` file is not updated unless one deletes it first.

## To Reproduce

Compile your project with
```
purs compile --codegen corefn **/*.purs
```
Modify one of the modules and recompile it with the same command.

## Expected behavior

`coregn.json` should be updated.

## PureScript version

0.13.8

hdgarrood:
I think this might be related to #3911; there might be a bug where the compiler is failing to update the "nonstandard" `--codegen` targets like `corefn` or `docs` if the externs file is already up to date.

nwolverson:
What I presume is another instance of the same underlying issue, `purs ide server` with the `--codegen corefn` option, compiling on save, does not seem to build the `js` when a standard`purs compile` is run (context: that was my default configuration as I also use an alternate backend). Guessing the assumption is that if the externs is up to date then all targets must have been generated.

hdgarrood:
Sounds likely. Externs are generated unconditionally, so I think we should only consider the other targets to be up to date if their modified times are greater than or equal to the externs file’s modified time.

hdgarrood:
Should be fixed by #4053.

PR #4053, opened 2021-04-11
Require rebuild if any codegen targets are outdated

### LINKED ISSUE #3911: Internal error due to old docs.json files
## Summary

I've just encountered a crash that happened due to presence of old `docs.json` files in the `output/` directory (removing `output/` solved the problem).

## Proposal

Include compiler version in all (or some) artifacts, to invalidate them when it does not match the actual version.

hdgarrood:
**Description of the change**

Fixes #3911, #3914: prevents an issue where a module would be considered
up-to-date if only some of the requested codegen targets were up-to-date
(as opposed to all of them).

---

**Checklist:**

- [x] (n/a) Updated or added relevant documentation

### DIFF
--- a/tests/TestMake.hs
+++ b/tests/TestMake.hs
@@ -9,6 +9,7 @@ import Prelude.Compat
 import qualified Language.PureScript as P
 import qualified Language.PureScript.CST as CST
 
+import Control.Concurrent (threadDelay)
 import Control.Monad
 import Control.Exception (tryJust)
 import Control.Monad.IO.Class (liftIO)
@@ -172,13 +173,16 @@ spec = do
           moduleContent2 = moduleContent1 <> "\ny :: Int\ny = 1"
           optsWithDocs = P.defaultOptions { P.optionsCodegenTargets = Set.fromList [P.JS, P.Docs] }
           go opts = compileWithOptions opts [modulePath] >>= assertSuccess
+          oneSecond = 10^(6::Int) -- microseconds
 
       writeFileWithTimestamp modulePath timestampA moduleContent1
       go optsWithDocs `shouldReturn` moduleNames ["Module"]
       writeFileWithTimestamp modulePath timestampB moduleContent2
+      threadDelay oneSecond
       go P.defaultOptions `shouldReturn` moduleNames ["Module"]
       -- Since the existing docs.json is now outdated, the module should be
       -- recompiled.
+      threadDelay oneSecond
       go optsWithDocs `shouldReturn` moduleNames ["Module"]
 
     it "recompiles if corefn is requested but not up to date" $ do
@@ -187,11 +191,14 @@ spec = do
           moduleContent2 = moduleContent1 <> "\ny :: Int\ny = 1"
           optsCorefnOnly = P.defaultOptions { P.optionsCodegenTargets = Set.singleton P.CoreFn }
           go opts = compileWithOptions opts [modulePath] >>= assertSuccess
+          oneSecond = 10^(6::Int) -- microseconds
 
       writeFileWithTimestamp modulePath timestampA moduleContent1
       go optsCorefnOnly `shouldReturn` moduleNames ["Module"]
       writeFileWithTimestamp modulePath timestampB moduleContent2
+      threadDelay oneSecond
       go P.defaultOptions `shouldReturn` moduleNames ["Module"]
+      threadDelay oneSecond
       go optsCorefnOnly `shouldReturn` moduleNames ["Module"]
 
 rimraf :: FilePath -> IO ()
--- a/tests/TestMake.hs
+++ b/tests/TestMake.hs
@@ -173,16 +173,16 @@ spec = do
           moduleContent2 = moduleContent1 <> "\ny :: Int\ny = 1"
           optsWithDocs = P.defaultOptions { P.optionsCodegenTargets = Set.fromList [P.JS, P.Docs] }
           go opts = compileWithOptions opts [modulePath] >>= assertSuccess
-          oneSecond = 10^(6::Int) -- microseconds
+          oneSecond = 10^(6::Int) -- microseconds.
 
       writeFileWithTimestamp modulePath timestampA moduleContent1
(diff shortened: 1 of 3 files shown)


hdgarrood:
github didn't interpret my original message the way I intended, so this comment is just for github; this fixes #3914 as well as #3911.

f-f (approved):
:clap:

hdgarrood:
Thanks for the review!

Apparently the "fixes #xyz" doesn't count if it's not in the original PR description >:(

hdgarrood:
I'm reasonably confident that this change does fix the issue, I think it's just that the tests are flaky.

rhendric:
What OS(es) are you testing on? It's a little suspicious that the macOS and Linux builds are both failing precisely these two new tests and no others, and I can't see anything in the tests themselves which would make them less reliable than the others in the suite. (I wonder if there's something like an OS-specific issue with reading file timestamps.)

hdgarrood:
I'm testing on Linux and both new tests seem to pass about half the time. In CI, different OSes will be using different filesystems - eg I'd guess the linux CI box is using ext4, and the windows CI box is probably using NTFS - and so I think we do expect some OS-specific behaviour here as a result of that, eg because different filesystems have different timestamp resolutions.

hdgarrood:
Actually until I can figure out why the tests are failing I think I’m going to say that I’m only somewhat confident this works.

rhendric:
Ah, no, I think your confidence is valid and you're absolutely right that it's a timestamp resolution thing. I added some code to the tests to log output file modification times and submitted that to Travis, and in the cases when it failed, the timestamps were exactly equal—either because, as on macOS, the filesystem was only recording with one-second resolution, or because, as on Linux, the timestamps appeared to be equal down to the nanosecond, which I can only assume is because the *kernel* has a limited time resolution. But regardless of the cause, based on its view of the world the changed code was correctly responding that nothing needed rebuilding.

In all of these tests, we're deterministically setting the modification times of the inputs, but not of the outputs, and these are the first tests to rely on the latter. Can we just fudge the modification times of `docs.json`/`corefn.json` in the tests? It's not clean but it's probably better than flaky tests.

hdgarrood:
I think we do have to be able to gracefully handle the cases where timestamps are exactly equal, unfortunately, as when I was trying to figure out what was going on here I found that it was relatively common that when running the compiler normally, the externs.json, index.js, and docs.json files all had the exact same timestamp. Experimentally, for me (ext4 on an SSD on linux), it appears that I do at least have microsecond resolution locally. In fact I am slightly surprised that the timestamps being equal is causing the tests to fail, because I deliberately used `<=` rather than `<` in `getOutputTimestamp` because I was expecting this to happen.

I don't think these are the first tests to rely on output timestamps incidentally - we compare output timestamps in order to ensure that downstream modules are recompiled if upstream modules change, so I think the two `recompiles downstream modules` tests also rely on output timestamps in a similar-ish way to these new ones.

rhendric:
The problem isn't that the timestamps are equal after the first compile; it's that the timestamps are equal after the second compile, between a file touched only by the first and a file recreated by the second. Your code does correctly handle the case when a single compile generates all its outputs with equal timestamps; in that case, everything appears to be up-to-date. This only causes an issue if a second compile then overwrites some of those outputs, making the others not up-to-date, but in a way that is invisible because the second compile happened within the same tick as the first. Outside of tests, I think that's an extremely unlikely scenario; and if you wanted to handle it correctly anyway, I think you'd have to abandon modification times altogether, instead embedding high-resolution timestamps or generation counters or input hashes in the files themselves, or in a cache database.

You might be right about `recompiles downstream modules`, but I had thought from an admittedly brief review that those tests were relying on the changed timestamp of one of the input files to trigger a rebuild of that module, and the rebuild of that module triggers a rebuild of any downstream modules, none of which depends on the timestamp of the output file, just on the previously cached timestamp of the input file. It's not a terribly important point either way, but I do think it explains why we haven't generally seen these tests flake before, whereas I'm pretty sure that without some modification, these new tests are likely to fail quite frequently.

hdgarrood:
Ah! Thanks. Your analysis - both for why these new tests are flaky and for why the existing ones aren't - is persuasive to me. In that case, I think adding a `sleep 1` in between each of the compiler invocations in these new tests would solve the problem, and that feels preferable to me than messing with the output file timestamps.

hdgarrood:
Adding `threadDelay (10^6)` does seem to fix the flakiness for me locally - both tests have passed 10 times in a row :)

tests/TestMake.hs
```diff
@@ -166,17 +167,55 @@ spec = do
...
+          moduleContent1 = "module Module where\nx :: Int\nx = 1"
+          moduleContent2 = moduleContent1 <> "\ny :: Int\ny = 1"
+          optsWithDocs = P.defaultOptions { P.optionsCodegenTargets = Set.fromList [P.JS, P.Docs] }
+          go opts = compileWithOptions opts [modulePath] >>= assertSuccess
+          oneSecond = 10^(6::Int) -- microseconds
+
+      writeFileWithTimestamp modulePath timestampA moduleContent1
+      go optsWithDocs `shouldReturn` moduleNames ["Module"]
+      writeFileWithTimestamp modulePath timestampB moduleContent2
+      threadDelay oneSecond
```
rhendric, line 181: Maybe a comment is in order on why `threadDelay` is used here but not in other tests? Or just a brief reference to this thread? I wouldn't want this to get copied into future tests where it isn't needed.

tests/TestMake.hs
```diff
@@ -166,17 +167,55 @@ spec = do
...
+      go P.defaultOptions `shouldReturn` moduleNames ["Module"]
+      -- Since the existing docs.json is now outdated, the module should be
+      -- recompiled.
+      threadDelay oneSecond
```
rhendric, line 185: Is this second one needed? The code being tested should only be comparing timestamps from the previous compiles to each other, not to the current time.

hdgarrood: Good point, I don't think it is needed.

rhendric (approved):

PR #4057, opened 2021-04-13
Switch to hspec exclusively, remove tasty

### LINKED ISSUE #4056: Remove tasty, use only hspec
## Summary

Remove use of tasty and tasty-hspec in the test suite and use only hspec directly.

## Motivation

Tasty and hspec are both testing frameworks: they are meant to satisfy the same need, and they're not really intended to be used together.  See e.g. this comment from the tasty-hspec Haddocks:

> hspec and tasty serve similar purposes; consider using one or the other.
>
> However, in a pinch, this module allows you to run an hspec Spec as a tasty TestTree.

I'd like to switch back to using hspec throughout for two reasons:

1. With hspec, failures are summarised at the end of the test output.  Tasty reports failures inline, which can be very annoying, especially in CI, as you then have to search through the logs to find the failure. Displaying failures at the end seems like a much more sensible default to me.

2. We get to take advantage of HasCallStack in hspec-expectations to help narrow down where a failure is coming from. This is particularly helpful in tests like the ones in TestMake.hs where we run something like ``someAction `shouldReturn` someResult`` more than once in a particular test, as without the source location from HasCallStack it's not immediately clear which assertion is failing. (This would have been useful in #4053).

## Proposal

Drop the `tasty`, `tasty-hspec`, and `tasty-golden` dependencies and use hspec throughout our test suite. There's a `hspec-golden` but I don't really like the look of it and I'd prefer we roll our own golden testing (you only need like 6 or 7 lines of code for what we want).

I'd also like to merge the cst tests back into the main `purescript` package. Having all of the tests in one place seems simpler, and I don't think we gain a lot from having the cst tests separated. This can happen separately, though.

One of the main motivations for introducing tasty originally (see #2848) was that it would let us run individual tests via CLI options. This is still possible if we switch back to hspec: hspec has a `--match` CLI option for this.

In hspec, there's no `--accept` flag for regenerating golden tests, but we can preserve similar behaviour by having our test suite recognise an environment variable (say, `HSPEC_ACCEPT`) or simply by letting the user delete the files they want to regenerate.

hdgarrood:
**Description of the change**

Resolves #4056
    
Removes tasty, tasty-hspec, and tasty-golden from the main compiler test suite
and the cst test suite, and switches it to use hspec directly instead. Golden
testing now uses the HSPEC_ACCEPT and CI environment variables to control its
behaviour (see the comments in TestUtils.hs for details). Cst tests are now
included in the main compiler test suite so that they can benefit from existing
functions in TestUtils.hs like `getTestFiles` and `goldenVsString`, and so that
we don't have different test executables which understand different sets of CLI
arguments (which is awkward).

I was admittedly a bit off with the "6 or 7 lines" I projected in #4056 needed
to roll our own golden testing, as our new `goldenVsString` function has
admittedly ballooned to about 40 lines (including comments), but I think it's
probably still a good tradeoff to keep the dependency/total LOC count down.

---

**Checklist:**


tests/Main.hs
```diff
@@ -23,44 +24,28 @@ import qualified TestMake
 import qualified TestUtils
 import qualified TestGraph
 
-import System.IO (hSetEncoding, stdout, stderr, utf8)
+import System.IO (hSetBuffering, stdout, BufferMode(..))
 
 main :: IO ()
 main = do
-  hSetEncoding stdout utf8
```
hdgarrood, line 30: I'm fairly sure this was only necessary in a previous version of the tests because we were previously using `Data.Text.IO.readFile`. We should no longer be using functions like `Data.Text.IO.readFile` which depend on things like locales and encodings associated with particular handles; we should instead be unconditionally treating things as utf-8 by reading them in as ByteStrings and then applying functions like `encodeUtf8` ourselves.

tests/Main.hs
```diff
@@ -23,44 +24,28 @@ import qualified TestMake
...
-  hSetEncoding stderr utf8
+  hSetBuffering stdout LineBuffering
```
hdgarrood, line 31: I was finding that the output was being buffered in larger blocks, perhaps because of how stack runs these tests. This line fixed the issue.

tests/TestCompiler.hs
```diff
@@ -53,29 +53,26 @@ import Text.Regex.Base
...
-  return . testGroup "compiler" $ [passing, warning, failing]
+import Test.Hspec
+
+spec :: Spec
+spec = do
+  (supportModules, supportExterns, supportForeigns) <- runIO setupSupportModules
+
+  passingTests supportModules supportExterns supportForeigns
+  warningTests supportModules supportExterns supportForeigns
+  failingTests supportModules supportExterns supportForeigns
```
hdgarrood, line 64: In the future I would quite like to bundle up `supportModules`, `supportExterns`, and `supportForeigns` into a single record data type, as I imagine that would make all of this quite a bit neater.

tests/TestCompiler.hs
```diff
@@ -84,41 +81,36 @@ warningTests
...
+      expectedWarnings <- runIO $ getShouldWarnWith mainPath
+      it ("'" <> takeFileName mainPath <> "' should compile with warning(s) '" <> intercalate "', '" expectedWarnings <> "'") $ do
         assertCompilesWithWarnings supportModules supportExterns supportForeigns testPurs expectedWarnings
-    return $ wTc ++ [ goldenVsString
-                      ("'" <> takeFileName mainPath <> "' golden test")
-                      (replaceExtension mainPath ".out")
-                      (BS.fromStrict . T.encodeUtf8 . T.pack <$> printErrorOrWarning supportModules supportExterns supportForeigns testPurs)
-                    ]
-  return $ testGroup "Warning examples" $ concat tests
+        goldenVsString
```
hdgarrood, line 94: It occurred to me while I was doing this that we are actually compiling everything in here twice: once when verifying the `shouldWarnWith`, and again when verifying the golden file output. That's also something I'd like to address in the future.

tests/TestDocs.hs
```diff
@@ -306,8 +302,8 @@ displayAssertionFailure = \case
     "in rendered code for " <> decl <> ", bad link location for " <> target <>
     ": expected " <> T.pack (show expected) <>
     " got " <> T.pack (show actual)
-  WrongOrder _ before after' ->
-    "expected to see " <> before <> " before " <> after'
+  WrongOrder _ before' after' ->
```
hdgarrood, line 305: This change addresses shadowed name warnings from the `before` exported by Test.Hspec

hdgarrood:
I've updated to hspec master and it seems to work well. @rhendric you could try `stack test purescript --test-arguments --times` with this branch now?

rhendric:
I like it! I think we should add `--times` as a default either to an `.hspec` file or, at least, to `ci/build.sh`—having timing information in CI seems like a very low-cost and potentially useful thing to have.

tests/Main.hs
rhendric: I tested and it seems this line is no longer necessary with the new bleeding-edge Hspec.

hdgarrood:
Ah yes sounds good, I forgot about the possibility of putting it in a `.hspec` file.

rhendric:
Oh no... just based on the outcomes here, I suspect that the Unicode characters in the new `checks` formatter don't play nicely with GitHub's Windows runner... somehow.

From the Windows build:
```
  purescript    > cst
  purescript    >   Layout golden tests
  purescript    > tests.exe: <stdout>: commitBuffer: invalid argument (invalid character)
  purescript    >     AdoIn [
  purescript    > Test suite tests failed
```

From the macOS build:
```
  purescript    > cst
  purescript    >   Layout golden tests
  purescript    >     AdoIn [✔]
```

Based on some web research it sounds like we need either to fiddle with the `LC_ALL` or `LANG` environment variables, or to put those `hSetEncoding` lines back in.

rhendric (approved):

sol:
I think the Unicode output issue on Windows is something I want to address in `hspec` somehow.

sol:
I guess technically, it will also happen on Linux, if the local is `C` or something.

purescript.cabal
```diff
@@ -39,6 +39,7 @@ extra-source-files:
   INSTALL.md
   CONTRIBUTORS.md
   CONTRIBUTING.md
+  .hspec
```
sol, line 42: @hdgarrood I'm not sure if it is a good idea to include the `.hspec` file in the generated dist tarball.  By doing so, tests will fail for users that run the tests for a release if they don't have the latest (or as of now even unreleased) version of `hspec`.

hdgarrood: Ah of course, thanks for pointing that out. I'm inclined not to worry though, because I don't think it's particularly important in our case for the tests to work in any other context than running `stack test` inside the repo. But I guess if this causes problems we can always take it out again and supply the flag on the command line in the CI script.

sol: Yes, I agree.  I don't think it's particularly important neither.  For my packages I used to not care much, but I had people complain about failing tests (e.g. `stackage` runs tests by default). I guess as long as nobody complains, it's fine.

> we can always take it out again and supply the flag on the command line in the CI script.

To be clear, I don't see any issue with the `.hspec` file in the repo, just adding it to the `.cabal` package could cause issues without providing much benefit.

hdgarrood: Right - to clarify, the reason I was tempted to add .hspec here in the first place was that our CI runs tests in a source distribution to ensure that we don’t end up publishing a tarball which is missing files that are necessary for the package to build or something, so just adding .hspec to the repo isn’t enough to get the times shown in CI.

sol: Ok, that makes sense.

PR #4064, opened 2021-04-15
Improve apartness checking

### LINKED ISSUE #3329: Apartness checker should take impossible equalities into consideration
Given the following setup:
```purescript
class Inject f g where
  inj :: f -> g
  prj :: g -> Maybe f

instance injectRefl :: Inject x x where
  inj x = x
  prj x = Just x
else instance injectLeft :: Inject l (Either l r) where
  inj x = Left x
  prj (Left x) = Just x
  prj _ = Nothing
else instance injectRight :: Inject x r => Inject x (Either l r) where
  inj x = Right (inj x)
  prj (Right x) = prj x
  prj _ = Nothing
```

With the following scenario:
```purescript
-- should work but doesn't
injL :: forall f g. f -> Either f g
injL = inj

injR :: forall f g. g -> Either f g
injR = inj
```

In the case of `injL` the reason is that the apartness checker doesn't know that `f` can't possibly be `Either f g` because that would lead to an infinite type: `Either (Either (Either ... g) g) g`.
But `injR` shouldn't work, because we don't know that `f /~ g`.

### LINKED ISSUE #3531: Possible bug in apartness checking
Consider the following program:
```purescript
module Main where

import Prelude

import Effect.Class.Console (log)
import Prim.TypeError (class Fail, Beside, Quote, Text)

class C x where
  thing :: x -> x

data X a b = X

test1 :: forall a. X a Unit
test1 = X

instance cx :: C (X x x) where
  thing = identity

else instance cxFail ::
  Fail
    (Beside
      (Beside (Text "Fell through: x ~ ") (Quote x))
      (Beside (Text ", y ~ ") (Quote y)))
      => C (X x y) where
  thing = identity

main = do
  let X = thing test1
  log "Done"
```
It fails to pass type checking, with the following error:
```
Compiling Main
[1/1 NoInstanceFound] src/Main.purs:28:7

  28    let X = thing test1
            ^^^^^^^^^^^^^^^
  
  A custom type error occurred while solving type class constraints:
  
    Fell through: x ~ t1, y ~ Unit
  
  while applying a function thing
    of type C t0 => t0 -> t0
    to argument test1
  while inferring the type of thing test1
  in value declaration main
  
  where t0 is an unknown type
        t1 is an unknown type
```

However, my understanding is that the `cx` instance should be selected, and that we should only be able to fall through to the next instance once we can ensure that the two type variable arguments to `X` are apart. In this case, they are not, since `t1` could unify with `Unit`.

### LINKED ISSUE #3941: Type class instance selector doesn't check the apartness of parameters
## Description

Type class instance selector should check their parameters are apart.

## To Reproduce

Prepare the example type class:

```purescript
class TwoParams a b where
  func :: a -> b

instance equals :: TwoParams a a where
  func = identity
else
instance any :: TwoParams a b where
  func = unsafeCoerce
```

And the below is correct behavior:

```purescript
-- can compile
testEquals :: forall a. a -> a
testEquals = func -- with instance `equals`
testAny :: Unit -> Boolean
testAny = func -- with instance `any`

-- expects compile error, cannot choose instance because `b` can be `Unit`
test :: forall b. Unit -> b
test = func
--     ^^^^
--
--   No type class instance was found for
--
--     Main.TwoParams Unit
--                    b0
--
--
-- while checking that type forall (a :: Type) (b :: Type). TwoParams a b => a -> b
--   is at least as general as type Unit -> b0
-- while checking that expression func
--   has type Unit -> b0
-- in value declaration test
```

However, the below should be able to compiled but not now.

```purescript
-- `a` and `m a` are never unifiable unless we have infinite types (and of course not)
-- so expected that the instance `any` is chosen.
thisShouldBeCompiled :: forall m a. a -> m a
thisShouldBeCompiled = func
--                     ^^^^
--
--   No type class instance was found for
--
--     Main.TwoParams a0
--                    (m1 a0)
--
--
-- while checking that type forall (a :: Type) (b :: Type). TwoParams a b => a -> b
--   is at least as general as type a0 -> m1 a0
-- while checking that expression func
--   has type a0 -> m1 a0
-- in value declaration thisShouldBeCompiled

```

## Expected behavior

The second one can be compiled with the instance `any`.

## Additional context

This may be root cause of some scenario preventing instance selection with type inference.

## PureScript version

v0.13.8 and v0.14.0-rc2

### LINKED ISSUE #4028: Verifiy the application of instance chain skolem type variables in a second step
## Summary

In an instance chain

```purescript
data Peano
foreign import data Succ :: Peano -> Peano
foreign import data Z :: Peano

instance fooZ :: Foo Z
instance foo2 :: Foo (Succ (Succ Z))
instance fooN :: Foo (Succ n)
```

Prohibit `foo2` and `fooN` to be chosen for the same input.

## Motivation

In the following snippet, two different branches of an instance chain are chosen for the same input:

```purescript
module Test.Main where

import Prelude

import Control.Monad.Indexed (class IxApplicative, class IxApply, class IxBind, class IxFunctor, class IxMonad)
import Control.Monad.Indexed.Qualified as Ix
import Data.List (List(..), fromFoldable)
import Data.Newtype (class Newtype, unwrap)
import Effect (Effect)
import Effect.Class (class MonadEffect)
import Effect.Class.Console (log)
import Type.Proxy (Proxy(..))

data Unk

foreign import data Unk0 :: Unk

foreign import data UnkX :: Unk -> Unk
newtype MUnk (i :: Unk) (o :: Unk) a
  = MUnk (Effect a)

derive instance newtypeUnk :: Newtype (MUnk i o a) _

derive newtype instance functorMUnk :: Functor (MUnk i o)

derive newtype instance applicativeMUnk :: Applicative (MUnk i o)

derive newtype instance applyMUnk :: Apply (MUnk i o)
derive newtype instance bindMUnk :: Bind (MUnk i o)
derive newtype instance monadMUnk :: Monad (MUnk i o)
derive newtype instance monadEffectMUnk :: MonadEffect (MUnk i o)

instance munkIxFunctor :: IxFunctor MUnk where
  imap f (MUnk a) = MUnk (f <$> a)

instance munkIxApplicative :: IxApply MUnk where
  iapply (MUnk f) (MUnk a) = MUnk (f <*> a)

instance munkIxApply :: IxApplicative MUnk where
  ipure a = MUnk $ pure a

instance munkIxBind :: IxBind MUnk where
  ibind (MUnk monad) function = MUnk (monad >>= (unwrap <<< function))

instance munkIxMonad :: IxMonad MUnk

class TLShow :: ∀ (k ∷ Type). k → Constraint
class TLShow i where
  tlShow :: Proxy i -> String

instance tlShowUnk2 :: TLShow (UnkX (UnkX Unk0)) where
  tlShow _ = "I'm unk 2"
else instance tlShowUnk0 :: TLShow Unk0 where
  tlShow _ = "Unk0"
else instance tlShowUnkX :: TLShow x => TLShow (UnkX x) where
  tlShow _ = "UnkX (" <> tlShow (Proxy :: Proxy x) <> ")"

class IxRec m where
  next :: forall (i :: Unk) a. Show a => TLShow i => a -> m i (UnkX i) a
  done :: forall (i :: Unk). m i Unk0 Unit

iFor_ :: forall (i :: Unk) m a. Show a => TLShow i => IxRec m => IxMonad m => List a -> m i Unk0 Unit
iFor_ Nil = done
iFor_ (Cons a b) = Ix.do
  _ <- next a
  iFor_ b

printMUnk :: forall (i :: Unk). TLShow i => MUnk i i Unit
printMUnk = Ix.do
  log $ tlShow (Proxy :: Proxy i)

instance ixRec :: IxRec MUnk where
  next a = Ix.do
    printMUnk
    log $ show a
    pure a
  done = pure unit

foreign import arr0 :: Effect (Array Int)

main :: Effect Unit
main = do
  list0 <- fromFoldable <$> arr0
  let (MUnk o) = (iFor_ :: List Int -> MUnk Unk0 Unk0 Unit) list0
  o
  log $ (tlShow (Proxy :: Proxy (UnkX (UnkX (UnkX Unk0)))))
```

produces

```bash
Unk0
1
UnkX (Unk0)
2
UnkX (UnkX (Unk0))
3
UnkX (UnkX (UnkX (Unk0)))
4
UnkX (UnkX (UnkX (UnkX (Unk0))))
5
UnkX (I'm unk 2)
```

Meaning that `"UnkX (UnkX Unk0)"` and `"I'm unk 2"` are both chosen as outputs for the function `tlShow :: Proxy i -> String` with input `UnkX (UnkX Unk0)` depending on the control flow of the program. It seems like that should not be allowed.

## Proposal

I have no idea how the compiler works with instance chains, but I would assume that it does a pass over instance chains to find the first one that unifies with the input type and then uses that.

I would propose to disallow unification under the following conditions that correspond to the example above:

1. The input type has a type variable during unification
1. The instance type has a type variable
1. An instance in an instance chain, if passed in as an input, could unify with the actual input if the input were part of the instance chain.

Concretely, in the example above, that would mean the following:

1. The input type has a type variable `UncX i` during unficiation.
1. The instance type `(UnkX x)` has a type variable.
1. An instance in the instance chain, `(UnkX (UnkX Unk0))`, if passed in as an input, could unify with `UncX i` if `UncX i` were part of the instance chain.

rhendric:
Instances in an instance chain that don't match the constraint being
searched must be proven apart from the constraint for the search to
continue to the next instance in the chain. This commit tweaks the
apartness checker to fix a variety of incorrect apartness judgments,
and also adds a hint to the NoInstanceFound error message if ambiguous
instances (neither matching nor apart) excluded additional instances in
a chain from consideration.

**Description of the change**

Fixes #3329, fixes #3531, fixes #3941, fixes #4028.

---

**Checklist:**


rhendric:
Note to self: add a test case with an anonymous instance after #4085 lands. Drafted until then.

hdgarrood (approved):
Looks great! This is going to help a lot with what I think is a relatively common source of confusion.

rhendric:
Just as a procedural note, I want to land #4118 before this, and then adapt this to use the same error message formatting for unnamed instances.

hdgarrood:
Sounds good - feel free to go ahead and merge that now if you're ready.

JordanMartinez:
CI builds and another core team member approved it. To clarify, can @rhendric merge this now?

PR #4077, opened 2021-04-22
Migrate CI from Travis to GitHub Actions

rhendric:
**Description of the change**

Inspired by #3980, but makes some different choices and takes advantage of some things that can now be simpler.

Releases probably deserve some specific commentary. One way this could go—the way it looks like #3980 was going—is that the releaser creates a tag on GitHub, and that triggers the workflow, and the workflow creates a draft release for that tag which then will need to be updated with the changelog (manually for now, until/unless we want to mess with automatically extracting it from the changelog file). This PR takes the alternative that the releaser creates a *release* on GitHub, and that triggers the workflow, which uploads the build products to the existing release; I believe this is closer to the existing release workflow, as I understand it from the release guide. But if the other approach is preferable I see no reason why it wouldn't work.

---

**Checklist:**


.github/workflows/ci.yml
```diff
@@ -0,0 +1,90 @@
...
+            *)
+              echo "Unknown OS name: $os_name"
+              exit 1;;
+          esac
+          cd sdist-test
+          bundle/build.sh "$bundle_os"
+
+      - name: "(Release only) Publish bundle"
+        if: "${{ env.CI_RELEASE == 'true' }}"
+        # Astonishingly, GitHub doesn't currently maintain a first-party action
```
hdgarrood, line 82: This is indeed astonishing. I’m kind of tempted to just implement this locally, to be honest. I think it probably ought to just be a couple of `curl` calls.

hdgarrood: Ah I see, if they’re thinking about deprecating releases that would certainly make sense.

hdgarrood:
This looks good to me. Creating a release and having the action upload the assets sounds good to me too. Come to think of it does RELEASE_GUIDE.md need an update now? Also would you be able to test creating a release using this action on your fork?

rhendric:
I think RELEASE_GUIDE.md doesn't need to change; it already describes creating a release and then waiting for CI to build and upload assets.

I tested the release on my fork before creating the PR. I deleted the releases shortly after but they looked just like the releases here—same assets, same names.

rhendric:
Actually never mind, I didn't delete the last one yet. Here: https://github.com/rhendric/purescript/releases/tag/v999.9.1

hdgarrood (approved):
Looks great! I might suggest renaming it from “CI & Release” to just “CI” because it might be a little surprising to see “Release” in the name of the check for a new or infrequent contributor? But that’s probably inconsequential.

hdgarrood:
Oh also, I think it might be worth adding this to the changelog too.

rhendric:
Done and done.

JordanMartinez (approved):

hdgarrood:
Thank you @rhendric and @JordanMartinez!

PR #4078, opened 2021-04-23
Infer foreign modules from actual file in rebuild

i-am-the-slime:
**Description of the change**

Without this change, a rebuild with a temporary file unsuccessfully looks for a foreign file in the same location as that temporary file instead of the FFI for the actual file. That always leads to errors about missing FFI files.

**Checklist:**


kritzcreek:
Sorry I missed this PR. This is implemented as is on purpose. My idea was that editors would have to copy both files to a temporary location, as you could have both a `.purs` as well as a `.js` file changed inside your editor but not saved to disk yet.

Does that make sense?

i-am-the-slime:
@kritzcreek I was afraid that this would be the answer. I'll try to do more copying and just hope everybody has an SSD 😁 .

PR #4079, opened 2021-04-24
Avoid compiling tests with diagnostics twice

hdgarrood:
Previously we compiled every warning and failing test example twice. Now
we just do it once. This speeds up the tests.

I've taken the opportunity to simplify the API provided by TestUtils
slightly. I've removed the callback-style API provided by `assert`
(because imo it just wasn't pulling its weight and isn't needed when we
already have `compile`, and also it encouraged ignoring warnings). I've
also removed the `check` argument to `compile`. The only way it was
being used was in the passing tests, to assert that we had a module
called `Main` in each test. However, there's not really any need for
this check: if a test has no `Main` module, we'll find that out (and the
test will fail) when we try to run it under `node`.

These changes shave about 20s off the test suite for me locally - from
125s down to 105s. That's not quite as much as I'd hoped but still nice.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ Internal:
 
 * Remove tasty from test suite and just use hspec (#4056, @hdgarrood)
 
+* Avoid compiling tests with diagnostics twice in test suite (#4079, @hdgarrood)
+
 ## v0.14.1
 
 New features:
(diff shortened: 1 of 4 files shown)


rhendric (approved):
:rocket:

PR #4081, opened 2021-04-26
Note FreeBSD binary packages in the installation instructions.

arrowd:


JordanMartinez:
Since this isn't something we control, should there be a comment that clarifies that the binary is provided via a 3rd-party?

hdgarrood:
Just outside the displayed diff, it does say this:

> Many of these distributions are provided and maintained by the community, and may not be immediately up to date following a new release.

so I think we're probably fine. The Homebrew one is already "unofficial," in the sense that it's not maintained by the core team.

hdgarrood (approved):

JordanMartinez:
Ah.. Gotcha.

JordanMartinez (approved):

Issue #4084, opened 2021-05-06
Make type class instance names optional by generating their names instead


JordanMartinez:
## Summary

No longer require developers to write a name for type class instances as the compiler will become smart enough to generate a unique name instead.

## Motivation

This change would remove the boilerplate at the source code level which is tedious to write and not necessary. In the real world, one often names such instances using this convention: `className <> fold typeParamterNameList`. For example...
```purescript
class ClassName a b c d
instance classNameFooBarBazBop :: ClassName Foo Bar Baz Bop
```

Rather than forcing the developer to write these names, the compiler would generate them.

## Proposal

Instance names at the source code level would be optional, so that current code doesn't break and future code doesn't need to use additional boilerplate. Perhaps in a future breaking change, such names could be removed altogether.

This change would not remove the requirement that instances have names at the backend-code level (e.g. the outputted JavaScript, etc.). Rather, the compiler would generate a unique name for each type class instance based on the type class name and the type parameters for that class.

Such a generator would need to account for tricky cases while keeping names short yet readable.

## Examples

Rather than writing this:
```purescript
instance classFooBar :: Class Foo Bar where
  func = Tuple Foo Bar

derive instance genericFoo :: Generic Foo _

instance class :: Class { a :: { b :: { c :: Int } }, d :: Tuple Int (Maybe String) }
```
developers would write this:
```purescript
instance Class Foo Bar where
  func = Tuple Foo Bar
  
derive instance Generic Foo _

instance Class { a :: { b :: { c :: Int } }, d :: Tuple Int (Maybe String) }
```
which would be equivalent to writing something like this:
```purescript

instance classFooBar :: Class Foo Bar where
  func = Tuple Foo Bar
  
derive instance genericFoo_ :: Generic Foo _

instance classRecARecBRecCIntDTupleIntMaybeString :: Class { a :: { b :: { c :: Int } }, d :: Tuple Int (Maybe String) }

wclr:
What about deprecating those names and disallowing the usage of instance names in FFI (in future versions)? Just to make codebases more readable in general. I wonder if there really people who would like to keep this meaningless noise in the code.

hdgarrood:
:+1: see also #3426

JordanMartinez:
Also relates to #752

JordanMartinez:
Closed by #4086

Issue #3426, opened 2018-09-07
Much shorter instancing/generics-deriving syntax


jnbooth:
This Haskell code is quite readable:

```haskell
data SomethingA = Alice | Alfred | Alex deriving (Eq, Ord, Show, Bounded, Enum) 
data SomethingB = Bob | Bart            deriving (Eq, Ord, Show, Bounded, Enum) 
data SomethingC = Carol | Candice       deriving (Eq, Ord, Show, Bounded, Enum)
data SomethingD = David | Donald        deriving (Eq, Ord, Show, Bounded, Enum)
```

The equivalent PureScript code is basically unreadable, not to mention painful to write:

```purescript
data SomethingA = Alice | Alfred | Alex
data SomethingB = Bob | Blart
data SomethingC = Carol | Candice
data SomethingD = David | Donald
derive instance _0_ ∷ Generic SomethingA _ 
derive instance _1_ ∷ Eq SomethingA 
derive instance _2_ ∷ Ord SomethingA 
instance _3_ ∷ Show SomethingA where
   show = genericShow 
instance _4_ ∷ Enum SomethingA where
   succ = genericSucc
   pred = genericPred 
instance _5_ ∷ Bounded SomethingA where
   top = genericTop
   bottom = genericBottom 
instance _6_ ∷ BoundedEnum SomethingA where
   cardinality = genericCardinality
   toEnum = genericToEnum
   fromEnum = genericFromEnum 
derive instance _7_ ∷ Generic SomethingB _ 
derive instance _8_ ∷ Eq SomethingB 
derive instance _9_ ∷ Ord SomethingB 
instance _10_ ∷ Show SomethingB where
   show = genericShow 
instance _11_ ∷ Enum SomethingB where
   succ = genericSucc
   pred = genericPred 
instance _12_ ∷ Bounded SomethingB where
   top = genericTop
   bottom = genericBottom 
instance _13_ ∷ BoundedEnum SomethingB where
   cardinality = genericCardinality
   toEnum = genericToEnum
   fromEnum = genericFromEnum 
derive instance _14_ ∷ Generic SomethingC _ 
derive instance _15_ ∷ Eq SomethingC
derive instance _16_ ∷ Ord SomethingC 
instance _17_ ∷ Show SomethingC where
   show = genericShow 
instance _18_ ∷ Enum SomethingC where
   succ = genericSucc
   pred = genericPred 
instance _19_ ∷ Bounded SomethingC where
   top = genericTop
[...]

PR #4085, opened 2021-05-06
Make instance names optional by generating them based on types

JordanMartinez:
**Description of the change**

Begins to address #4084.

The biggest question here is how the name generator should work. 
- uniqueness - My current "just print the types" approach likely works for 80% of all cases, but any complicated ones using lots of parameters and nesting might clash with another instance that just so happens to produce the same name. I'm wondering whether using a generated name (e.g. `GenIdent`) would be a simpler and better approach
- brevity - by accounting for all types, the name will be unique. However, it might be very long as a result. Consider what should be generated when nested records are used.
- FFI-friendly? - I'm not sure if there's a concern that instances be callable from FFI. If there is, then I assume that one could workaround this problem by defining a name for an instance by hand (i.e. what we currently do for al instances).

The next question is how this code should be tested. I could take all current instance tests and add a variation of each file's code to see if things still work/fail when the `name ::` part is removed from the instance. Since the code would otherwise be the same AST, I'm not sure that's necessary. A better approach would be testing whether names are unique and short enough.

---

**Checklist:**


rhendric:
I'd be in favor of opaque generated names (`$1`). If you omit the instance name, you shouldn't rely on what the compiler chooses for an instance name. So the compiler shouldn't export a name that you might think you could rely on. All that *and* it would be simpler.

hdgarrood:
I agree that we should make it hard to end up relying on what the compiler chooses for an instance name, but having the scheme be just a `$` followed by an incrementing integer would hurt the readability of the generated code quite a bit, and reading the generated code is useful from time to time. What if, as a compromise, we smush the type names together and then append a number? That way hopefully readability is still preserved, but the number at the end is still going to make it difficult to rely on.

hdgarrood:
As for FFI-friendliness, I don't think that's a concern; we are gradually pushing users away from calling functions with constraints from JavaScript. For example, #3182 was a big step in this direction.

JordanMartinez:
> What if, as a compromise, we smush the type names together and then append a number? That way hopefully readability is still preserved, but the number at the end is still going to make it difficult to rely on.

Sounds good to me. Could you help guide me on how this number should be generated?

rhendric:
How about `className$1`, as opposed to `classNameArgArgArg$1`? I'm sympathetic about readability but some of those smushed names could get so long as to hinder readability in a different way (think [`monadStateStateHalogenMStateActionSlotsOutputM$1`](https://github.com/purescript-halogen/purescript-halogen/blob/5bcd961a4a3954f639da556ed6661b14bb425ae4/src/Halogen/Query/HalogenM.purs#L107)).

hdgarrood:
I think I prefer truncating rather than only including the class name too. It's been too long since I've been in this code and so I don't really remember, but I'd recommend following the pattern used in other places where we generate names (e.g. for unknown types when type checking). I think using GenIdent together with MonadSupply for generating the integers should work.

hdgarrood:
I think that if you generate the names in a desugaring step rather than in the CST->AST conversion, it might be simpler and it will match better with the existing patterns in the compiler. Of course it's not really ideal that doing it in a desugaring step means that the AST type will have to use a Maybe for the name as well, but we intend to get rid of explicit instance names eventually anyway (at least I think we do?) and so since we'd be able to get rid of that Maybe then, having to add the occasional `fromMaybe (internalError ...)` when accessing an instance name that we know exists doesn't seem too bad to me in the meantime.

JordanMartinez:
I'll ask a dumb question. Why not ignore explicit names immediately and generate names using the truncation idea above? Is it because this would be a breaking change in case someone (who knows where) is using an instance in FFI? Or is there something else going on here I don't know of?

JordanMartinez:
> I think that if you generate the names in a desugaring step rather than in the CST->AST conversion, it might be simpler and it will match better with the existing patterns in the compiler.

Yeah, that's partly why I was originally asking "where" to generate the number. `convertDeclaration` runs in the List monad, so I don't have access to `freshIdent` as a side effect. Also, is it not possible to generate the number when the parser is running? Or does that incur other problems?

hdgarrood:
Not a dumb question! I was thinking I’d like to continue using the declared instance names where they exist for as long as we are going to accept them, and that is indeed because of FFI - specifically because there may be cases where people are calling PureScript functions with constraints from the target language (eg JS). The compiler has no way of knowing when this is happening, so I think it will be worth being slightly more cautious. Also, I think it’s a little harder to justify going straight to ignoring the names with no advance notice because we haven’t (yet) explicitly come out and said anywhere that calling functions with constraints from outside PureScript is unsupported and going to become impossible as far as I know.

kl0tl:
While working on this last summer I ended up with the following scheme for generating unique but still readable instance names, following the usual convention:

+ a leading dollar,
+ then the class name,
+ then two dollars,
+ then the instance arguments which are not determined by any functional dependency and are either nullary type constructors, application chains headed by a type constructor, or symbols, prefixed by `$sym` to distinguish them from type constructors, separated by two dollars.

Arguments in application chains are separated by a single dollar and are printed according to the scheme for instances arguments.

We don’t need to consider rows, variables nor determined arguments outside of instance chains because of overlaps, but we can avoid collisions easily enough by appending the instance position in the chain.

There’s a slight hurdle though: this scheme needs access to both the names written by the programmer (well I guess that’s not strictly required but fully qualified names do not help with readability), which are only available before renaming, and functional dependencies, which are only available after desugaring type classes 🙃

JordanMartinez:
> There’s a slight hurdle though: this scheme needs access to both the names written by the programmer (well I guess that’s not strictly required but fully qualified names do not help with readability), which are only available before renaming, and functional dependencies, which are only available after desugaring type classes 

Does that mean there is preliminary work that needs to be done before this approach can be taken? Or that a different approach should be taken?

JordanMartinez:
> There’s a slight hurdle though: this scheme needs access to both the names written by the programmer (well I guess that’s not strictly required but fully qualified names do not help with readability), which are only available before renaming, and functional dependencies, which are only available after desugaring type classes :upside_down_face:

Sounds like this is better explained [in #3426 (comment)](https://github.com/purescript/purescript/issues/3426#issuecomment-651082718). In other words, it sounds like there is some preliminary work that needs to be done before multiple things can be unblocked.

rhendric:
#4086 only blocks the smart name munging approach to this, not the quick-and-dirty truncated-string-plus-`freshIdent` approach. There are a couple of reasons to prefer the latter: it's likely simpler than what the implementation of #4086 would be, and I think it's more honest—this change does seem like a revision of the AST, not just the concrete surface syntax, and translating unnamed instances to uniquely-named instances does seem like a desugaring step.

JordanMartinez:
Thanks for clarifying. I'll keep working on this.

One other question. Looking at [the lines for converting instance chains from CST to AST](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Convert.hs#L469-L481), it seems that all instance names within an instance chain is stored as a list in each `AST.TypeInstanceDeclaration` value. For example:
```purescript
instance foo :: ClassA
else instance bar :: ClassA
else instance baz :: ClassA
```
will be stored as three separate `AST.TypeInstanceDeclaration` where the `chainId` value stores the entire list as `["foo", "bar", "baz"]`:
```
AST.TypeInstanceDeclaration ann ["foo", "bar", "baz"] 0 "foo" deps className tys body
AST.TypeInstanceDeclaration ann ["foo", "bar", "baz"] 1 "bar" deps className tys body
AST.TypeInstanceDeclaration ann ["foo", "bar", "baz"] 2 "baz" deps className tys body
```
By removing the instance names, the source code would be something like this...
```
instance ClassA --
else instance ClassA --
else instance ClassA --
```
and the resulting `AST.TypeInstanceDeclaration` would be this:
```
-- foo
AST.TypeInstanceDeclaration ann [Nothing, Nothing, Nothing] 0 Nothing deps className tys body
-- bar
AST.TypeInstanceDeclaration ann [Nothing, Nothing, Nothing] 1 Nothing deps className tys body
-- baz
AST.TypeInstanceDeclaration ann [Nothing, Nothing, Nothing] 2 Nothing deps className tys body
```
Using what would be `baz` as an example, I believe I can generate an instance name for `baz`'s instance since I still have the `className` and `tys` values available. Thus, the AST would be updated to be something like this:
```
-- baz
AST.TypeInstanceDeclaration ann [Nothing, Nothing, Just "$$ClassName$Foo$2"] 2 "$$ClassName$Foo$2" deps className tys body
```
But how would I update the other two `Nothing`s (i.e. `foo` and `bar`'s names)? While the `className` value should remain the same in an instance chain, I do not know what `foo` and `bar`'s `tys` arguments are.

rhendric:
That chain field always struck me as a little bit broken, actually. I don't think it has to be a list of names; it just has to be some kind of unique identifier. I don't know why it couldn't be a `(ModuleName, SourcePos)`, or anything else that's convenient to make unique.

JordanMartinez:
How is `chainId` even used? As a key in some `Map`?

rhendric:
I know it's used in Language/PureScript/TypeChecker.hs as part of `checkOverlappingInstance`. I think but am not 100% confident that this is its one and only purpose.

JordanMartinez:
Hm... Maybe I should spend more time reading through this repo's source code and just getting a deeper understanding of things before I continue contributing. Perhaps documenting things along the way might help, too.

JordanMartinez:
Took a quick break from other work to see where `chainId` is used. Here's where it's used:

[The chain is converted to a chain with qualified names](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker.hs#L419-L420), then in [checkOverlappingInstances](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker.hs#L493-L514) the list is used only once to [prevent an `OverlappingInstance` error from being thrown](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker.hs#L507-L508). Also, the [`tcdChain` function comes from the `TypeClassDictionary` type](https://github.com/purescript/purescript/blob/master/lib/purescript-ast/src/Language/PureScript/TypeClassDictionaries.hs#L18)

Essentially, if we can accomplish that equality check with something functionally equivalent, then it unblocks the next part of this PR.

JordanMartinez:
Latest update on this PR.

After getting a better understanding of the general flow of this code, I took the following approach:
- I changed the `name` arg of AST's `TypeInstanceDeclaration` from `Ident` to `Either Text Ident`. `Left t` corresponds to the partially generated (i.e. no unique identifier added) name of the instance as computed in `convertDecl`. `Right i` corresponds to the unique name defined in source code. I also changed it's `chainId` from `[Ident]` to `[Either Text Ident]`. It ultimately works the same way it does before this PR is introduced (further described below)
- I added another pass to the desugar code that will complete the generated instance names via `fresh`. During this pass, I also return the mapping (i.e. `instMap`) from the partially-completed instance name to the fully generated instance name.
- In the `desugarTypeClasses` pass, all `Left t` are remapped to their final names via the mapping created in the previous pass. The final and correct instance names with fully generated values are then passed into the rest of the compiler's pipeline. 

A few questions I have:
- should all types in `convertDecl` be mapped to a text value? Right now I only have some of them based on the `typeAtom` expression, and I'm curious if `TypeParens` might reveal one of the other ones not currently covered.
- will the instance name remapping approach I've taken with `instMap` ever produce an error because two instances in a chain produce the same partially-completed instance name before the unique identifier is appended to it?
- how should this PR be tested?

src/Language/PureScript/Sugar/TypeClasses/Instances.hs
```diff
@@ -0,0 +1,48 @@
...
+-- will be converted to `Right Ident` values.
+--
+desugarTypeClassInstanceNames
+  :: (MonadSupply m, MonadError MultipleErrors m)
+  => Module
+  -> m (M.Map Text Ident, Module)
+desugarTypeClassInstanceNames (Module ss coms name decls exps) = do
+  results <- parU decls desugarInstName
+  let (desugared, instMaps) = foldr splitVals ([], []) results
+  pure (foldr M.union mempty instMaps, Module ss coms name desugared exps)
```
JordanMartinez, line 30: Two comments here:
1. is there a better way to write `foldr splitVals ([], []) results`? I thought `sequence` would work, but I got a compiler error. So, I'm wondering if there's a different value I should be using or perhaps a type annotation
2. when `union`ing the `instMaps`, should an error be thrown here if the same key (i.e. same partially-generated instance name) is found multiple times? Right now, this will be caught as an internal error in the desugar pass, but here is where we could throw an error if this situation occurred.

JordanMartinez:
Regardless, this PR is ready for feedback.

src/Language/PureScript/Sugar/TypeClasses/Instances.hs
rhendric: I'll give an actual review later, but isn't 1 just `unzip`?

src/Language/PureScript/Sugar/TypeClasses/Instances.hs
```diff
@@ -0,0 +1,63 @@
...
+
+-- |
+-- Completes the name generation for type class instances that do not have
+-- a unique name defined in source code. All `Left Text` values
+-- will be converted to `Right Ident` values.
+--
+desugarTypeClassInstanceNames
+  :: (MonadSupply m, MonadError MultipleErrors m)
+  => Module
+  -> m (M.Map Text Ident, Module)
```
rhendric, line 28: On review, I'm not so much a fan of this instance map thing. Maybe I'm too attached to my idea of completely replacing the chain ID type with some other unique identifier, but the comments you left on your investigation into the chain ID seemed to indicate that it could work, and it'd solve the duplicate generated name problem and keep the desugaring pipeline neater. Did you find something that made you think it wouldn't work? If not, I could take a swing at doing that refactoring in my own PR if you'd rather not undertake it here.

rhendric (changes requested):
The actual name generation stuff all looks good to me! Nice work with the test coverage too.

My only important comments are the ones about the instance name map and whether `DuplicatePartiallyGeneratedInstanceName` should exist. I think we can do better; let me know if you want me to pitch in or if you prefer to try it yourself.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +529,44 @@ convertDeclaration fileName decl = case decl of
...
+        TypeRow _ _ -> "Row"
+        TypeArrName _ _ -> "Arrow"
+        --
+        TypeWildcard _ _ -> ""
+        TypeForall _ _ _ _ _ -> ""
+        TypeApp _ _ _ -> ""
+        TypeOp _ _ _ _ -> ""
+        TypeArr _ _ _ _ -> ""
+        TypeConstrained _ _ _ _ -> ""
+        TypeUnaryRow _ _ _ -> ""
```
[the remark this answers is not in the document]
JordanMartinez: Ah.. didn't know about that. Thanks for pointing that out. I agree it should be used.

JordanMartinez:
> My only important comments are the ones about the instance name map and whether `DuplicatePartiallyGeneratedInstanceName` should exist. I think we can do better; let me know if you want me to pitch in or if you prefer to try it yourself.

I'd prefer if this error and the instance mapping did not exist. The approach I took here was the simplest way to implement this feature while changing the least amount of code. The biggest issue was finding a way to track which instances were originally part of some instance chain. It seemed like I wasn't supposed to make `convertDecl` have side effects, so a unique identifier would need to be generated either before that (e.g. in the Parser phase where I wasn't sure how to do that) or in the desugaring part (the approach taken in this PR). I'm curious to hear what your idea is for fixing the overlapping instance error check through some alternative means.

Also, I think the name generation is still broken. I've found a few tests that cause the 'duplicate partially-generated instance name' error. Looks like more constructors need to be accounted for because `TypeParens` can wrap more things than what `typeAtom` covers.

Below is the before and after of this PR. It helps explain the changes I made and why.

<hr />

Flow of code before this PR
1. Instance names must include the `instanceName ::` syntax where `instanceName` is unique per module.
1. Source code is parsed into a CST type called `DeclInstanceChain`. (It wasn't clear to me that generating a unique identifier could be done here, nor how it would be done if it can be done.)
1. CST type `DeclInstanceChain` is converted into the corresponding AST type called `TypeInstanceDeclaration` via a pure computation. No side effects (like generating a unique number to append to the instance name) can be done here.
    1. When an instance is a part of an instance chain, the unique name of each instance in the chain is stored in a list and this list is stored in each `TypeInstanceDeclaration` via the `chainId` arg.
1. In the desugar process for type classes, the `TypeInstanceDeclaration` AST is convertd into a type class dictionary
2. In the type checker process, the `TypeInstanceDeclaration` AST is checked for a number of errors, but the relevant one for this case is the Overlapping Instances check. If an instance is a part of an instance chain, this check is skipped if the instance's `chainId` list is equal to the corresponding value in the type class dictionary.

Flow of code after this PR.
1. Instance names can include the `instanceName ::` syntax where `instanceName` is unique per module, but this can now be excluded.
1. Source code is parsed into a CST type called `DeclInstanceChain`. When a name is provided, both the name and `::` characters must exist. These values are stored via `Just (name, separator)`. If no name and separator is provided, a `Nothing` value is used.
1. CST type `DeclInstanceChain` is converted into the corresponding AST type called `TypeInstanceDeclaration` via a pure computation. No side effects (like generating a unique number to append to the instance name) can be done here.
    1. Instances with names (i.e. `Just (name, separator)` just reuse their names. Instances without names (i.e. `Nothing` value above) have their names partially-generated at this step based on the CST class name and CST type parameters. The CST type parameters utilized in this generation is based on what the `Parser.y` rules for `typeAtom`. Since no side-effects can be done here, the name is only partially complete. Thus, instances with names are wrapped in `Right` while instances with partially-generated names are wrapped in `Left`.
        1. The partially-generated instance name here follows this format: `$$(module_)?className$arg1$arg2$...$argN`
        1. `$$ClassName$String` for `instance ClassName String`
        1. `$$M_ClassName$String` for `instance M.ClassName String`
    1. When an instance is a part of an instance chain, there is no the guarantee that the name is unique due to not having a completely-generated instance name yet (e.g. `instance Foo x else instance Foo x`). Still, the same approach of storing the instance names (e.g. the partially-generated via `Left` and developer-specified via `Right`) in the chain in each `TypeInstanceDeclaration`'s `chainId` arg is still taken.
    1. Since each type class instance declaration in a module is handled in isolation from all other instances, we cannot determine whether or not two or more unnamed instances generated the exact same partially-generated instance name. At this step, we're only parsing and converting to AST, not validating. This check will need to be done later in the compilation pipeline.
1. In the desugar process, a new module pass is added.
    1. A unique identifier via `fresh` is appending to all `Left`-wrapped partially-generated instance names to produce the final instance name. This final name is rewrapped into `Right` now that it is completely generated.
        1. The completely-generated instance name here follows this format: `$$(module_)?className$arg1$arg2$...$argN$identifier`
        1. `$$ClassName$String$142` for `instance ClassName String`
        1. `$$M_ClassName$String$143` for `instance M.ClassName String`
    1. Since it's possible that two or more instances within the same module may be using the same partially-generated instance name, we add another check to confirm that all partially-generated instance names only refer to one final instance name. If it does not, a `DuplicatePartiallyGeneratedInstanceName` error is thrown.
    1. Now that all instances have their completely-generated instance names and we have confirmed that each partially-generated instance name corresponds to exactly one completely-generated instance name, we can now update the `chainId` list arg to use the completely-generated instance names.
    1. A new version of the module with its `TypeInstanceDeclaration` type updated with the completely-generated instance name in both its `chainId` list arg and `name` arg is returned.
1. In the desugar process for type classes, the `TypeInstanceDeclaration` AST is convertd into a type class dictionary
1. In the type checker process, the `TypeInstanceDeclaration` AST is checked for a number of errors. Since the names in the `chainId` are unique (and all unnamed instances now have completely-generated names), the Overlapping Instance error is still skipped using the same "if part of an instance chain" check.

Unanswered questions:
- Can a unique identifier be created before the `CST` to `AST` conversion process? Or is there another way to check whether instances are a part of a chain in a type class dictionary? If so, it would completely remove the current instance mapping approach I'm forced to take by those constraints.
- If kept, what would make the `DuplicatePartiallyGeneratedInstanceName` error actionable and clear as to what needs to be fixed?

lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
```diff
@@ -46,13 +46,17 @@ flattenInstance (Instance a b) =
...
 flattenInstanceHead :: InstanceHead a -> DList SourceToken
-flattenInstanceHead (InstanceHead a b c d e f) =
+flattenInstanceHead (InstanceHead a b c d e) =
   pure a <>
-  flattenName b <>
-  pure c <>
-  foldMap (\(g, h) -> flattenOneOrDelimited flattenConstraint g <> pure h) d <>
-  flattenQualifiedName e <>
-  foldMap flattenType f
+  flattenInstanceName b <>
```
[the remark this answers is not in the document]
JordanMartinez: Thanks! Latest commits pushed this change.

lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
```diff
@@ -502,7 +502,7 @@ declName (ExternDataDeclaration _ n _) = Just (TyName n)
 declName (FixityDeclaration _ (Left (ValueFixity _ _ n))) = Just (ValOpName n)
 declName (FixityDeclaration _ (Right (TypeFixity _ _ n))) = Just (TyOpName n)
 declName (TypeClassDeclaration _ n _ _ _ _) = Just (TyClassName n)
-declName (TypeInstanceDeclaration _ _ _ n _ _ _ _) = Just (IdentName n)
+declName (TypeInstanceDeclaration _ _ _ n _ _ _ _) = either (const Nothing) (Just . IdentName) n
```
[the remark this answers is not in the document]
JordanMartinez: Thanks! Latest commits pushed this change.

rhendric:
In `convertDeclaration`, I propose taking the `fileName` parameter and a `SourcePos` from the `DeclInstanceChain` (or from the first instance in the chain, I guess) and creating a unique chain ID from those. It's impossible for two different chains to occupy the same position in the same file, so you get global uniqueness without side effects. If you want to be extra nice, declare the `ChainId` type in its own module and only export a smart constructor, so that it's evident to any future readers that the contents of the `ChainId` are completely irrelevant to anything other than `Eq`/`Ord`/etc. stuff.

Then I'd keep most of the rest of what you did with generating the final names in the desugaring phase. I like your split between generating a name hint pre-desugaring and truncating and finalizing that hint into a name during desugaring. Generating unique chain IDs that don't depend on instance names should alleviate the need for both the name map and the error, which also makes any potential remaining bugs in the name hint generation much less significant.

JordanMartinez:
> In convertDeclaration, I propose taking the fileName parameter and a SourcePos from the DeclInstanceChain (or from the first instance in the chain, I guess) and creating a unique chain ID from those.

Good idea! I started going down a similar route using the `SourcePos` as an identifier. Somewhere along the way, I had to change something in `Externs`, which I think would be a breaking change. However, I think that occurred because I was doing something unnecessary due to not understanding this codebase more. Let me investigate that and get back to you.

JordanMartinez:
Yup, using the file name and source pos way to generate unique IDs forces me to touch the Externs types. Does that matter? Or does it mean code compiled via `v0.14.1` won't work on `v0.14.2` (assuming this is included in that release)?

Otherwise, it seems to work.

rhendric:
I think that should be fine. Externs files already store the version of the compiler that produced them, and they get blown away every time the version doesn't match regardless of whether the format changed.

JordanMartinez:
Based on [the most recent CI error](https://github.com/purescript/purescript/runs/2619498799#step:8:2591) that I fixed in latest commits, will 25 characters as an instance name be long enough to be readable in all circumstances?

Consider that...

```
$$TestRowApplications$Row$29
```
... was generated for `instance TestRowApplications (RowFoo ())`

If the instance in the code was
```
import Some.Module.Name as LongModuleName
instance LongModuleName.LongClassName LongFirstTypeName
```
the generated instance name would be
```
$$LongModuleName_LongClas$47
```

Perhaps each part of the name should be truncated to so many characters? For example, truncating the module name of the class name to 5 characters, then truncating the class name to 10 characters (so that it's guaranteed that a maximum of 20 characters make up the first part of the name (e.g. `$$Modul_ClasssName$Type1$4`). 

If it was, the above long name instance example would generate the below name:
```
$$LongM_LongC$LongFirstT$47
```
The number for the module name is arbitrary. It could just as easily be 3 characters.
The number for the class name is arbitrary, but consider that `MonadError` is 10 characters.
The number for the type name is arbitrary. Given the other two numbers, 5 characters seemed reasonable.

kl0tl (commented):
The dollars inserted in the generated names are turned into `$dollar` by `Language.PureScript.CodeGen.JS.Common.identToJs`. We need a leading dollar to avoid name collisions but If we’re not relying on the name generation to yield unique names I don’t think we need to use dollars as separators. Perhaps we can get our leading dollar and the unique suffix with `Language.PureScript.Names.runIdent`, by wrapping the generated names in `GenIdent`?

JordanMartinez:
> Perhaps we can get our leading dollar and the unique suffix with Language.PureScript.Names.runIdent, by wrapping the generated names in GenIdent?

That's what I was doing originally, but that caused [problems with `properNameToJson`](https://github.com/purescript/purescript/pull/4085/commits/fb5638de793d5a212533a867d537a512d3a3f106)

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,49 @@ convertDeclaration fileName decl = case decl of
...
+        TypeParens _ t -> argName $ wrpValue t
+        TypeKinded _ t1 _ t2 -> argName t1 <> "$" <> argName t2
+        TypeRecord _ _ -> "Record"
+        TypeRow _ _ -> "Row"
+        TypeArrName _ _ -> "Arrow"
+        --
+        TypeWildcard{} -> "_"
+        TypeForall _ _ _ _ t -> argName t
+        TypeApp _ t1 t2 -> argName t1 <> "$" <> argName t2
+        TypeOp _ t1 _ t2 -> argName t1 <> "_op_" <> argName t2
```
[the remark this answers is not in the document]
JordanMartinez: I thought at type operator may not convert well to a name that works in JavaScript (e.g. if the operator was `::>`). I didn't realize it's the same approach I'm taking with `TypeOpName`. I'll fix that in the next commit.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,49 @@ convertDeclaration fileName decl = case decl of
...
+          (N.runProperName $ qualName cls)
+
+      typeArgs :: Text.Text
+      typeArgs = Text.intercalate "$" $ fmap argName args
+
+      -- based on Parser.y's `typeAtom` expression
+      -- which is used in its `instHead` expression
+      argName :: Type a -> Text.Text
+      argName = \case
+        TypeVar _ n -> N.runIdent $ ident $ nameValue n
```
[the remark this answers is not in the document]
JordanMartinez: I'm not sure what `TypeVar` even does. Based on your comment above, it seems like they could be rendered as `""`.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,49 @@ convertDeclaration fileName decl = case decl of
...
+        TypeConstructor _ qn -> N.runProperName $ qualName qn
+        TypeOpName _ qn -> N.runOpName $ qualName qn
+        TypeString _ _ ps -> prettyPrintStringJS ps
+        TypeHole _ n -> N.runIdent $ ident $ nameValue n
+        TypeParens _ t -> argName $ wrpValue t
+        TypeKinded _ t1 _ t2 -> argName t1 <> "$" <> argName t2
+        TypeRecord _ _ -> "Record"
+        TypeRow _ _ -> "Row"
+        TypeArrName _ _ -> "Arrow"
```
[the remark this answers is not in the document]
JordanMartinez: Thanks for clarifying that.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,49 @@ convertDeclaration fileName decl = case decl of
...
+
+      -- based on Parser.y's `typeAtom` expression
+      -- which is used in its `instHead` expression
+      argName :: Type a -> Text.Text
+      argName = \case
+        TypeVar _ n -> N.runIdent $ ident $ nameValue n
+        TypeConstructor _ qn -> N.runProperName $ qualName qn
+        TypeOpName _ qn -> N.runOpName $ qualName qn
+        TypeString _ _ ps -> prettyPrintStringJS ps
+        TypeHole _ n -> N.runIdent $ ident $ nameValue n
```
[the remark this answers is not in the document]
JordanMartinez: Probably.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,49 @@ convertDeclaration fileName decl = case decl of
...
+    maybe (Left genName) (Right . ident . nameValue . fst) nameSep
+    where
+      -- instance name truncation will occur in desugaring process
+      -- when the final instance name is determined
+      genName :: Text.Text
+      genName = "$$" <> className <> "$" <> typeArgs
+
+      className :: Text.Text
+      className =
+        (maybe "" ((<> "_") . N.runModuleName) (qualModule cls)) <>
```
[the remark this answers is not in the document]
JordanMartinez: Oh, good point!

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,49 @@ convertDeclaration fileName decl = case decl of
...
+        TypeString _ _ ps -> prettyPrintStringJS ps
+        TypeHole _ n -> N.runIdent $ ident $ nameValue n
+        TypeParens _ t -> argName $ wrpValue t
+        TypeKinded _ t1 _ t2 -> argName t1 <> "$" <> argName t2
+        TypeRecord _ _ -> "Record"
+        TypeRow _ _ -> "Row"
+        TypeArrName _ _ -> "Arrow"
+        --
+        TypeWildcard{} -> "_"
+        TypeForall _ _ _ _ t -> argName t
```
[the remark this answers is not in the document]
JordanMartinez: Probably.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,55 @@ convertDeclaration fileName decl = case decl of
...
+        TypeWildcard{} -> "_"
+
+        -- Polytypes are disallowed in instance heads
+        TypeForall{} -> ""
+        TypeApp _ t1 t2 -> argName t1 <> "$" <> argName t2
+        TypeOp _ t1 op t2 ->
+          argName t1 <> "$" <> (N.runOpName $ qualName op) <> "$" <> argName t2
+        TypeArr _ t1 _ t2 -> argName t1 <> "_Arrow_" <> argName t2
+        TypeConstrained{} -> ""
+        TypeUnaryRow{} -> "EmptyRow"
```
JordanMartinez, line 581: Is this right? Or should this be rendered as `""`. It looks like the only usage of this constructor is when [it's used with the deprecated `#` syntax](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L312

lib/purescript-ast/src/Language/PureScript/AST/Declarations.hs
```diff
@@ -429,7 +431,7 @@ data Declaration
   -- A type instance declaration (instance chain, chain index, name,
   -- dependencies, class name, instance types, member declarations)
   --
-  | TypeInstanceDeclaration SourceAnn [Ident] Integer Ident [SourceConstraint] (Qualified (ProperName 'ClassName)) [SourceType] TypeInstanceBody
+  | TypeInstanceDeclaration SourceAnn [ChainId] Integer (Either Text Ident) [SourceConstraint] (Qualified (ProperName 'ClassName)) [SourceType] TypeInstanceBody
```
rhendric, line 434: Can this be `ChainId` instead of `[ChainId]` (here and everywhere)? Identifying each chain by the position of its first instance should work just as well as identifying them by the list of the positions of all their instances, right?

src/Language/PureScript/Sugar/TypeClasses/Instances.hs
```diff
@@ -0,0 +1,43 @@
...
+    => Declaration
+    -> m Declaration
+  desugarInstName = \case
+    TypeInstanceDeclaration sa chainId idx (Left genText) deps className tys bd -> do
+      uniqueInt <- fresh
+      -- truncate to 25 chars to reduce verbosity
+      -- of name and still keep it readable
+      let finalName = Ident $ (take 25 genText) <> "$" <> (pack $ show uniqueInt)
+          decl = TypeInstanceDeclaration sa chainId idx (Right finalName) deps className tys bd
+      pure decl
```
rhendric, line 42:
```suggestion
      -- truncate to 25 chars to reduce verbosity
      -- of name and still keep it readable
      finalName <- Ident . (take 25 genText <>) <$> freshName
      pure $ TypeInstanceDeclaration sa chainId idx (Right finalName) deps className tys bd
```

rhendric:
> > Perhaps we can get our leading dollar and the unique suffix with Language.PureScript.Names.runIdent, by wrapping the generated names in GenIdent?
> 
> That's what I was doing originally, but that caused [problems with `properNameToJson`](https://github.com/purescript/purescript/pull/4085/commits/fb5638de793d5a212533a867d537a512d3a3f106)

Is it possible you're misremembering this? I don't see how `properNameToJSON` would be involved or would be unhappy about `GenIdent`. (I've bumped into `GenIdent` problems in the past and I totally believe that there's a problem *somewhere*, but I don't know off the top of my head where I ought to be looking.)

src/Language/PureScript/Sugar/TypeClasses/Instances.hs
JordanMartinez: Fixed in latest commit

JordanMartinez:
> Is it possible you're misremembering this? 

I doubt it, but I was doing a lot of unnecessary things while I was figuring this out. So, I might still be wrong because the error arises due to some other code path. Latest commits use `GenIdent`. I'll let CI find the issue if there is one.

I also added an update to the changelog. I'm not sure if that's too verbose or not.

JordanMartinez:
This was the error I got previously due to [`identToJs`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CodeGen/JS/Common.hs#L25):
```
Failures:   
purescript>             
purescript>   src/Language/PureScript/CodeGen/JS/Common.hs:25:28: 
purescript>   1) compiler, Passing examples, 'InstanceNamesGenerated.purs' should compile and run without error
purescript>        uncaught exception: ErrorCall
purescript>        An internal error occurred during compilation: GenIdent in identToJs
purescript>        Please report this at https://github.com/purescript/purescript/issues
purescript>        CallStack (from HasCallStack):
purescript>          error, called at src/Language/PureScript/Crash.hs:24:3 in purescript-ast-0.1.1.0-BLTG9STKhgqLSa4aQ0dlin:Language.PureScript.Crash
purescript>          internalError, called at src/Language/PureScript/CodeGen/JS/Common.hs:25:28 in purescript-0.14.1-3aa3CQouQqd5OmdJNpklHw:Language.PureScript.CodeGen.JS.Common
purescript>       
```

JordanMartinez:
I'm going to switch from using `GenIdent` back to `Ident`.

rhendric:
Aha, yes, now I remember. `GenIdent`s are supposed to be eliminated by `renameInModules`, but identifiers at the top level are never renamed, so `GenIdent`s can't currently be used for top level declarations.

I think that could be changed. The renamer is nice in that it guarantees uniqueness not by introducing extra dollar signs, but by actually searching the scope and generating a new name that doesn't shadow anything. If we fixed the renamer so that it renames `GenIdent`s used as top-level identifiers and module exports, then we don't have to use dollars as separators anymore, which is nice in light of the fact that they get unhelpfully rendered as `$dollar` in the resulting JS.

I'd be okay with doing that in a later PR, though. All it really affects is how nice the generated names look.

rhendric (approved):

JordanMartinez:
> I'd be okay with doing that in a later PR, though. All it really affects is how nice the generated names look.

I think the renamer should be updated as `$dollar$dollarModule_ClassName$dollarArray$dollarString$dollarBo$dollar42` doesn't satisfy the 'readability' concern we have here. But yeah, that can be done as a separate PR.

natefaubion:
Why do we rename `$` to `$dollar` at all? It's not a valid PS identifier character, so it's not clear to me it needs to renamed at all.

rhendric:
I assume the thinking is that we use `$` in the other things we escape to, and so we have to escape the escape character if we want the mapping to be injective.

Of course, PureScript doesn't currently let you mix symbols and non-symbols in a single identifier, so injectivity would hold even if we didn't escape `$`, since there's no other way to produce `$hash` in the output than with a `#` in the input. So... good question?

natefaubion:
I would wager that it's just an artifact from when we had operators as proper identifiers, and is otherwise not necessary. It should only be necessary to escape names that clash with JS globals.

rhendric:
Oh dang, that's right. Kind of forgot operators aren't identifiers, actually. :grimacing:

So huh... then the only parts of `identCharToText` that pull any weight are, what, `$prime` and the catch-all `ord` case?

JordanMartinez:
So.... How should this PR proceed?

rhendric:
IMO, up to you. There are some avenues for improvement here. I don't think any of them should be blockers for the good work you've done so far. If you'd like to improve the $dollarBoBollar problem in this PR, be my guest. You could do that by attacking `identCharToText`, or the rewriter, or using dollars less liberally in your name generator; I think it'd be reasonable to include any of that work in this PR. I also think it'd be reasonable for you to squash and declare mission accomplished, as you have in fact achieved the objective and left PS better than you found it. That's my not-a-real-maintainer two cents.

JordanMartinez:
>  using dollars less liberally in your name generator

I think this approach makes more sense as a quick fix until a future PR fixes it properly. If `$` will convert to `$dollar` in the final instance name, I'd rather use `_` as the separators as it just stays as `_` in `identCharToText`.

Put differently, 
```
-- given this:
instance Module.ClassName (Array String) Boolean

-- which currently desugars to something like this
$dollar$dollarModule_ClassName$dollarArray$dollarString$dollarBo$dollar42

-- it would instead desugar to this
$dollarModule_ClassName_Array_String_Bo$dollar42

-- or potentially to this where 'inst' implies 'instance'
inst_Module_ClassName_Array_String_Bo___42
```

JordanMartinez:
Latest commit changes this...
```
$dollarModule_ClassName$dollarArray$dollarString$dollarBo$dollar42
```
to this
```
$dollarModule_ClassName_Array_String_Bo$dollar42
```

I didn't change the left-most or right-most dollars as it's not yet clear from other core members whether that should be done. Changing the middle $ chars to _ seemed reasonable to make without additional feedback.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,57 @@ convertDeclaration fileName decl = case decl of
...
+      -- of name and still keep it readable
+      -- unique identifier will be appended to this name
+      -- in desugaring proces
+      genName :: Text.Text
+      genName = Text.take 25 ("$" <> className <> "_" <> typeArgs) <> "$"
+
+      className :: Text.Text
+      className = do
+        let modName = (<> "_") . Text.replace "." "_" . N.runModuleName
+        (maybe "" modName (qualModule cls)) <> (N.runProperName $ qualName cls)
```
kl0tl, line 551: I think this might be equivalent to `Text.replace "." "_" $ showQualified runProperName cls`.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -530,6 +531,57 @@ convertDeclaration fileName decl = case decl of
...
+        TypeRow _ _ -> "Row"
+        TypeArrName _ _ -> "Function"
+        TypeWildcard{} -> "_"
+
+        -- Polytypes are disallowed in instance heads
+        TypeForall{} -> ""
+        TypeApp _ t1 t2 -> argName t1 <> "_" <> argName t2
+        TypeOp _ t1 op t2 ->
+          argName t1 <> "_" <> (N.runOpName $ qualName op) <> "_" <> argName t2
+        TypeArr _ t1 _ t2 -> argName t1 <> "_Arrow_" <> argName t2
```
kl0tl, line 581:
```suggestion
        TypeArr _ t1 _ t2 -> argName t1 <> "_Function_" <> argName t2
```

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
kl0tl: `Row Type` is rendered as `Row_Type`, so I think `# Type` should render to the same string:

```suggestion
        TypeUnaryRow{} -> "Row"
```

kl0tl (commented):
I left some comments on the instances names generation, nothing too important though. Also, do we even need to have separators since we don’t rely on them to disambiguate similar instances (for example both `C (Foo Bar) Baz` and `C Foo (Bar Baz)` yield `$C_Foo_Bar_Baz`, to which we then append an unique suffix)? Is `$dollarModule_ClassName_Array_S$dollar42` really more legible than `$dollarModuleClassNameArrayStr$dollar42`?

hdgarrood:
@rhendric for what it's worth, I definitely consider you to be a real maintainer. If you don't feel that way, I think that's more of an indication that we haven't quite sorted out our governance model just yet than anything else.

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
JordanMartinez: mm... doesn't seem like it. `showQualified` operates on `Language.PureScript.Names (Qualified)`, but `cls` is `Language.PureScript.CST.Types (QualifiedName)`

JordanMartinez:
> Is $dollarModule_ClassName_Array_S$dollar42 really more legible than $dollarModuleClassNameArrayStr$dollar42?

A few thoughts on this:
- Following current naming convention, we don't separate names with '_' or '$'. This was influenced by [kl0tl's original approach in a related work](https://github.com/purescript/purescript/pull/4085#issuecomment-834174237). So, I agree that separators should be dropped altogether.
- Following current naming convention, we don't add a module name to the instance name. Adding the module name was necessary when I was using my original approach for the chain id issue. Having switched to rhendric's filename and source pos idea, it's no longer necessary. So, the module should be dropped from the instance name.
- Since readability is a concern, I think adding a '_' character after the first '$dollar' would make the generated instance name easier to read (e.g. `$dollarClassNameArrayS$dollar42` vs `$dollar_ClassNameArrayS$dollar42`).
- I feel like the initial `$dollar` is good and necessary to prevent name clashing, but that the final `$dollar` separating the name from the unique identifier is not necessary. For readability concerns, perhaps adding two '_' characters would be better in case one defines a type that ends with an underscore?

At the very least, I'll implement the first three comments above, but leave the fourth one undone.

CHANGELOG.md
```diff
@@ -8,6 +8,29 @@ Breaking changes:
...
+  Now, the name and `::` separator characters are optional. The above instance
+  could be rewritten like so:
+
+  ```purescript
+  -- convention: classNameType1Type2Type3
+  instance Foo Int String
+  ```
+
+  and the compiler will generate a unique name for the instance
+  (e.g. `$Foo$$Int$String$4` where `4` is a randomly-generated number
```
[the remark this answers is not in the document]
JordanMartinez: Ah, good catch. Fixed.

JordanMartinez:
Any other changes on this? Or am I good to squash and merge?

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
kl0tl: Oh silly me, I didn’t pay attention to the types 🤦

kl0tl (approved):

JordanMartinez:
I was thinking about this PR just a few minutes ago and I realized that we can likely merge the `desugarTypeClassInstanceNames` pass into the `desugarTypeClasses` pass. These two passes originally needed to be separate passes because of how my original approach worked. 

In my original approach, the `chainId` was the name of the instances. Since some instances' names might clash, we needed to figure out all instance names in a given module first, and then determine what their final instance name was. If any clashed, we'd get the `DuplicatePartialInstanceName` error. If we didn't need to determine all initial instance names in a module, then the code for generating a unique identifier could be done in the `desugarTypeClasses` pass, right?

Once this PR changed the `chainId` to use the "filename and source positions" unique identifier approach proposed by @rhendric, the above requirement got dropped. I don't think anyone realized that at the time.

While the code might be more readable by keeping these two passes separate, should the instance name generation be merged into the `desugarTypeClasses` pass? Would that help compiler performance?

PR #4088, opened 2021-05-11
Unused var tweaks

### LINKED ISSUE #4083: Incorrect unused variable warning
## Description

When an identifier is used in a `let` expression which also introduces a binding of that same identifier, it is always considered to be unused regardless of whether the new binding is used later or not.

## To Reproduce

```purescript
test x = 
  let (x) = x
  in x
```
Here `x` is reported as unused, though the function argument `x` was used in the let binding, and the `x` bound by the pattern `(x)` is used in the body `x` - because the `x` binding is not in scope in the body of the let binding. Contrast with the following where the let-bound `x` is in scope in the let binding, the `x` argument of `test1` is correctly reported as unused.

```purescript
test1 x = 
  let x _ = x 1
  in x
```

## Expected behavior

The `test` example should not warn, the `test1` example should warn. This should also warn:

```purescript
test x = 
  let (x) = x
  in 1
```

## Additional context

Mea culpa.

## PureScript version

0.14.1

### LINKED ISSUE #4087: Unused variable warnings in case binders point to span for entire case expression
## Description

When introducing an unused variable in a `case` binder, the warning span will point to the entire `case` expression instead of the unused variable.

## To Reproduce

```purescript
bad :: _
bad = case _ of
  { a, b } -> a
```

## Expected behavior

The warning span should point to the variable binding.

## PureScript version

0.14.1

nwolverson:
Improvements to unused variable lint/warnings.

Fix a bad unused variable warning (fix #4083). The previous iteration was not distinguishing between variables bound recursively within the let binding, and those that are only bound inside the body.

Unchanged with this commit is that identifiers which are used recursively within a let binding will count as used (ie not warn), despite the fact that the binding could be removed - something more intelligent would be required to handle mutually recursive let blocks which may or may not be used externally.

Tighten the warning spans on unused variables (fix #4087).

This will scope to the identifier itself for binders, but a value declaration `let x = e in e2` will scope to the entire `x = e` as `ValueDecl` does not have a source span for the identifier alone.

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Linter.hs
+++ b/src/Language/PureScript/Linter.hs
@@ -176,7 +176,7 @@ lintUnused (Module modSS _ mn modDecls exports) =
     goDecl :: Declaration -> (S.Set Ident, MultipleErrors)
     goDecl d@(ValueDeclaration vd) =
         let allExprs = concatMap unguard $ valdeclExpression vd
-            bindNewNames = S.fromList (concatMap binderNames $ valdeclBinders vd)
+            bindNewNames = S.fromList (concatMap binderNamesWithSpans $ valdeclBinders vd)
             ss = declSourceSpan d
             (vars, errs) = removeAndWarn ss bindNewNames $ mconcat $ map (go ss) allExprs
             errs' = addHint (ErrorInValueDeclaration $ valdeclIdent vd) errs
@@ -196,7 +196,7 @@ lintUnused (Module modSS _ mn modDecls exports) =
             removeAndWarn ss letNames (go ss e)
             <> mconcat (map underDecl ds)
     go ss (Abs binder v1) =
-      let newNames = S.fromList (binderNames binder)
+      let newNames = S.fromList (binderNamesWithSpans binder)
       in
       removeAndWarn ss newNames $ go ss v1
 
@@ -218,7 +218,7 @@ lintUnused (Module modSS _ mn modDecls exports) =
     go ss (IfThenElse v1 v2 v3) = go ss v1 <> go ss v2 <> go ss v3
     go ss (Case vs alts) =
       let f (CaseAlternative binders gexprs) =
-            let bindNewNames = S.fromList (concatMap binderNames binders)
+            let bindNewNames = S.fromList (concatMap binderNamesWithSpans binders)
                 allExprs = concatMap unguard gexprs
             in
                 removeAndWarn ss bindNewNames $ mconcat $ map (go ss) allExprs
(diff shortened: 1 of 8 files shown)


rhendric (approved):

src/Language/PureScript/Linter.hs
```diff
@@ -277,10 +281,12 @@ lintUnused (Module modSS _ mn modDecls exports) =
     unguard' (ConditionGuard ee) = ee
     unguard' (PatternGuard _ ee) = ee
 
-    removeAndWarn :: SourceSpan -> S.Set Ident -> (S.Set Ident, MultipleErrors) -> (S.Set Ident, MultipleErrors)
-    removeAndWarn ss newNames (used, errors) =
-      let filteredUsed = used `S.difference` newNames
+    removeAndWarn :: SourceSpan -> S.Set (SourceSpan, Ident) -> (S.Set Ident, MultipleErrors) -> (S.Set Ident, MultipleErrors)
+    removeAndWarn _ newNamesWithSpans (used, errors) =
```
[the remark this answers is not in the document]
nwolverson: Because I have a short attention span?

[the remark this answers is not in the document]
nwolverson: Oh there is so much noise floating around due to this :)

src/Language/PureScript/Linter.hs
```diff
@@ -174,113 +174,116 @@ lintUnused (Module modSS _ mn modDecls exports) =
...
+    go (Unused v) = go v
+    go (IfThenElse v1 v2 v3) = go v1 <> go v2 <> go v3
+    go (Case vs alts) =
       let f (CaseAlternative binders gexprs) =
-            let bindNewNames = S.fromList (concatMap binderNames binders)
+            let bindNewNames = S.fromList (concatMap binderNamesWithSpans binders)
                 allExprs = concatMap unguard gexprs
             in
-                removeAndWarn ss bindNewNames $ mconcat $ map (go ss) allExprs
+                removeAndWarn bindNewNames $ mconcat $ map (go) allExprs
```
rhendric, line 223: Unneeded parens in `(go)`. (I found four of these in total; they're easy to search for.)

src/Language/PureScript/Linter.hs
```diff
@@ -174,113 +174,116 @@ lintUnused (Module modSS _ mn modDecls exports) =
...
+    doElts (DoNotationBind binder e : rest) v =
+      let bindNewNames = S.fromList (binderNamesWithSpans binder)
+      in go e <> removeAndWarn bindNewNames (doElts rest v)
+
+    doElts (DoNotationLet ds : rest) v =
+      let (letNewNames, letNewNamesRec) = foldMap declIdents ds
+      in removeAndWarn letNewNamesRec $
+            mconcat (map underDecl ds)
+            <> removeAndWarn letNewNames (doElts rest v)
+    doElts (PositionedDoNotationElement _ _ e : rest) v = doElts  (e : rest) v
```
rhendric, line 257: Extra space in the RHS here.

rhendric (approved):
Some cosmetic fixes and then LGTM.

src/Language/PureScript/Linter.hs
nwolverson: Oh dear, a little bit naive on the search and replace :)

rhendric:
Oh sorry, would you add a CHANGELOG.md entry for this too?

PR #4092, opened 2021-05-19
New command: purs codegen

### LINKED ISSUE #3339: Consider `purs codegen` command
This was mentioned somewhere in the discussion about how to deal with `--dump-corefn`, etc. but the general ideal was the compiler could pick up the corefn and then perform JS codegen on it, so you could have something like:

``` shell
purs compile --codegen corefn src/**/*.purs
-- custom corefn optimizer tool
pure codegen ...
```

colinwahl:
**Description of the change**

Implements a new command: `purs codegen`

`purs codegen` takes globs to filepaths containing the JSON representation of a CoreFn Module (this can be generated by `purs compile`). It parses the core functional representation out of these files, and passes them in to the standard `codegen` function.

This command allows for CoreFn transformations to be written outside of the compiler (even in PureScript!) without having to worry about using PureScript as a library.

Example usage of this would be:

```
$ purs compile glob/to/files.purs -g corefn,js
$ <execute pass over generated corefn.json files>
$ purs codegen glob/to/all/corefn.json
```

This intends to close #3339

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ Breaking changes:
 
 New features:
 
+* Add `purs codegen` command
+
+  Adds a new command, `purs codegen`, which generates JavaScript code
+  for the given files containing the CoreFn Module JSON representation.
+
 Bugfixes:
 
 * Unused identifier warnings now report smaller and more relevant source spans (#4088, @nwolverson)
--- a/app/Command/Codegen.hs
+++ b/app/Command/Codegen.hs
@@ -4,8 +4,23 @@ import Prelude
 
 import           Control.Applicative (many)
 import           Control.Monad (when, unless)
-import qualified Language.PureScript as P
+import           Control.Monad.IO.Class (MonadIO(..))
+import           Control.Monad.Supply
+
+import qualified Data.Aeson.Internal as A
+import           Data.Aeson.Parser (eitherDecodeWith, json)
+import qualified Data.ByteString.Lazy as BSL
+import           Data.Either (rights)
+import qualified Data.Set as S
+import qualified Data.Map as M
+
 import qualified Options.Applicative as Opts
+
+import qualified Language.PureScript.Docs.Types as Docs
+import qualified Language.PureScript as P
+import qualified Language.PureScript.CoreFn as CoreFn
+import qualified Language.PureScript.CoreFn.FromJSON as CoreFn
+
 import           System.Exit (exitFailure, exitSuccess)
 import           System.FilePath.Glob (glob)
 import           System.IO (hPutStr, hPutStrLn, stderr)
@@ -17,16 +32,54 @@ data CodegenOptions = CodegenOptions
 
 codegen :: CodegenOptions -> IO ()
 codegen CodegenOptions{..} = do
-  input <- globWarningOnMisses (unless codegenJSONErrors . warnFileTypeNotFound) codegenCoreFnInput
-  when (null input && not codegenJSONErrors) $ do
+  inputFiles <- globWarningOnMisses (unless codegenJSONErrors . warnFileTypeNotFound) codegenCoreFnInput
+  when (null inputFiles && not codegenJSONErrors) $ do
     hPutStr stderr $ unlines
       [ "purs codegen: No input files."
       , "Usage: For basic information, try the `--help` option."
       ]
     exitFailure
+  mods <- traverse parseCoreFn inputFiles
+  let
+    filePathMap =
+      M.fromList $ map (\m -> (CoreFn.moduleName m, Right $ CoreFn.modulePath m)) $ map snd $ rights mods
 
+  foreigns <- P.inferForeignModules filePathMap
+  _ <-
+    liftIO
+      $ P.runMake purescriptOptions
+      $ runSupplyT 0
+      $ traverse (runCodegen foreigns filePathMap) $ map snd $ rights mods
   exitSuccess
   where
+  parseCoreFn file = do
+    contents <- BSL.readFile file
+    case eitherDecodeWith json (A.iparse CoreFn.moduleFromJSON) contents of
+      Left (j, e) -> pure $ Left (file, j, e)
+      Right r -> pure $ Right r
+
+  makeActions foreigns filePathMap = P.buildMakeActions "output" filePathMap foreigns False
+
+  purescriptOptions :: P.Options
+  purescriptOptions = P.Options False False (S.fromList [ P.JS ])
+
+  runCodegen foreigns filePathMap m =
+    P.codegen (makeActions foreigns filePathMap) m
+      (Docs.Module (CoreFn.moduleName m) Nothing [] [])
+      (moduleToExternsFile m)
+
+  moduleToExternsFile :: CoreFn.Module a -> P.ExternsFile
+  moduleToExternsFile CoreFn.Module {CoreFn.moduleName} = P.ExternsFile {
+      P.efVersion      = mempty,
+      P.efModuleName   = moduleName,
+      P.efExports      = [],
+      P.efImports      = [],
+      P.efFixities     = [],
+      P.efTypeFixities = [],
+      P.efDeclarations = [],
+      P.efSourceSpan   = P.SourceSpan "none" (P.SourcePos 0 0) (P.SourcePos 0 0)
+    }
+
   warnFileTypeNotFound :: String -> IO ()
   warnFileTypeNotFound =
     hPutStrLn stderr . ("purs graph: No files found using pattern: " <>)
--- a/app/Command/Codegen.hs
+++ b/app/Command/Codegen.hs
@@ -34,6 +34,7 @@ import           System.IO (hPutStr, hPutStrLn, stderr)
 data CodegenOptions = CodegenOptions
   { codegenCoreFnInput :: [FilePath]
   , codegenJSONErrors :: Bool
+  , codegenOutputDir :: FilePath
   }
 
 codegen :: CodegenOptions -> IO ()
@@ -72,7 +73,7 @@ codegen CodegenOptions{..} = do
       Left (j, e) -> pure $ Left (file, j, e)
       Right r -> pure $ Right r
 
-  makeActions foreigns filePathMap = P.buildMakeActions "output" filePathMap foreigns False
+  makeActions foreigns filePathMap = P.buildMakeActions codegenOutputDir filePathMap foreigns False
 
   purescriptOptions :: P.Options
   purescriptOptions = P.Options False False (S.fromList [ P.JS ])
@@ -93,6 +94,7 @@ command = codegen <$> (Opts.helper <*> codegenOptions)
   codegenOptions =
     CodegenOptions <$> many inputFile
                    <*> jsonErrors
+                   <*> outputDirectory
 
   inputFile :: Opts.Parser FilePath
   inputFile =
@@ -105,6 +107,14 @@ command = codegen <$> (Opts.helper <*> codegenOptions)
     Opts.switch $
       Opts.long "json-errors" <>
       Opts.help "Print errors to stderr as JSON"
+
+  outputDirectory :: Opts.Parser FilePath
+  outputDirectory = Opts.strOption $
+    Opts.short 'o'
+      <> Opts.long "output"
+      <> Opts.value "output"
+      <> Opts.showDefault
+      <> Opts.help "The output directory"
 
 globWarningOnMisses :: (String -> IO ()) -> [FilePath] -> IO [FilePath]
 globWarningOnMisses warn = concatMapM globWithWarning
--- a/app/Command/Codegen.hs
+++ b/app/Command/Codegen.hs
@@ -1,0 +1,64 @@
+module Command.Codegen (command) where
+
+import Prelude
+
+import           Control.Applicative (many)
+import           Control.Monad (when, unless)
+import qualified Language.PureScript as P
+import qualified Options.Applicative as Opts
+import           System.Exit (exitFailure, exitSuccess)
+import           System.FilePath.Glob (glob)
+import           System.IO (hPutStr, hPutStrLn, stderr)
+
+data CodegenOptions = CodegenOptions
+  { codegenCoreFnInput :: [FilePath]
+  , codegenJSONErrors :: Bool
+  }
+
+codegen :: CodegenOptions -> IO ()
+codegen CodegenOptions{..} = do
+  input <- globWarningOnMisses (unless codegenJSONErrors . warnFileTypeNotFound) codegenCoreFnInput
+  when (null input && not codegenJSONErrors) $ do
+    hPutStr stderr $ unlines
+      [ "purs codegen: No input files."
+      , "Usage: For basic information, try the `--help` option."
+      ]
+    exitFailure
+
+  exitSuccess
+  where
+  warnFileTypeNotFound :: String -> IO ()
+  warnFileTypeNotFound =
+    hPutStrLn stderr . ("purs graph: No files found using pattern: " <>)
+
+command :: Opts.Parser (IO ())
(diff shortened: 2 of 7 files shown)


app/Command/Codegen.hs
```diff
@@ -0,0 +1,151 @@
...
+
+  outputDirectory :: Opts.Parser FilePath
+  outputDirectory = Opts.strOption $
+    Opts.short 'o'
+      <> Opts.long "output"
+      <> Opts.value "output"
+      <> Opts.showDefault
+      <> Opts.help "The output directory"
+
+globWarningOnMisses :: (String -> IO ()) -> [FilePath] -> IO [FilePath]
```
colinwahl, line 119: `Command.Graph`, `Command.Compile`, and `Command.Codegen` all contain the same definition here, but I wasn't sure the most appropriate place to pull it out to for sharing.

If there is an appropriate place to move this then I will update all 3 of those modules.

src/Language/PureScript/Make/Actions.hs
```diff
@@ -97,7 +97,7 @@ data MakeActions m = MakeActions
   , readExterns :: ModuleName -> m (FilePath, Maybe ExternsFile)
   -- ^ Read the externs file for a module as a string and also return the actual
   -- path for the file.
-  , codegen :: CF.Module CF.Ann -> Docs.Module -> ExternsFile -> SupplyT m ()
+  , codegen :: CF.Module CF.Ann -> Docs.Module -> Maybe ExternsFile -> SupplyT m ()
```
colinwahl, line 100: When performing `codegen` via `purs codegen` - we can create a stub `ExternsFile` via the `CoreFn.Module` - but it isn't actually the `ExternsFile` we want to write. I modified this so that we don't write a bogus `ExternsFile` during `purs codegen`

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ Breaking changes:
 
 New features:
 
+* Add `purs codegen` command
```
thomashoneyman, line 11: Do you mind adding the PR number and your username as well?

app/Command/Codegen.hs
```diff
@@ -0,0 +1,151 @@
...
+    filePathMap =
+      M.fromList $ map ((\m -> (CoreFn.moduleName m, Right $ CoreFn.modulePath m)) . snd) $ rights mods
+
+  unless (null (lefts mods)) $ do
+    _ <- traverse (hPutStr stderr . formatParseError) $ lefts mods
+    exitFailure
+
+  foreigns <- P.inferForeignModules filePathMap
+  (makeResult, makeWarnings) <-
+    liftIO
```
rhendric, line 61: Is `liftIO` needed here?

app/Command/Codegen.hs
```diff
@@ -0,0 +1,151 @@
...
+globWarningOnMisses warn = concatMapM globWithWarning
+  where
+  globWithWarning :: String -> IO [FilePath]
+  globWithWarning pattern' = do
+    paths <- glob pattern'
+    when (null paths) $ warn pattern'
+    return paths
+
+  concatMapM :: (a -> IO [b]) -> [a] -> IO [b]
+  concatMapM f = fmap concat . mapM f
```
rhendric, line 129: While you're messing around with this, you can import `concatMapM` from `Protolude`. Don't know why we aren't already doing that.

app/Command/Codegen.hs
```diff
@@ -0,0 +1,151 @@
...
+
+-- | Arguments: use JSON, warnings, errors
+printWarningsAndErrors :: Bool -> P.MultipleErrors -> Either P.MultipleErrors a -> IO ()
```
rhendric, line 132: This is just `Command.Compile.printWarningsAndErrors True`, right? Looks like another candidate for `Command.Common`.

colinwahl:
Sorry, things got busy around here, I'm going to pick this back up soon to address the feedback.

app/Command/Codegen.hs
colinwahl: Ah, it is, great call

src/Language/PureScript/Make/Actions.hs
colinwahl: Good idea - I can go ahead and do that now

colinwahl:
@rhendric I've addressed your initial feedback.

While doing the `codegen` refactoring, I noticed that `purs codegen` doesn't allow opting-in to generating source maps - I should probably add an option to the command to allow the user to opt-in to that.

app/Command/Codegen.hs
```diff
@@ -0,0 +1,107 @@
...
+      M.fromList $ map ((\m -> (CoreFn.moduleName m, Right $ CoreFn.modulePath m)) . snd) $ rights mods
+
+  unless (null (lefts mods)) $ do
+    _ <- traverse (hPutStr stderr . formatParseError) $ lefts mods
+    exitFailure
+
+  foreigns <- P.inferForeignModules filePathMap
+  (makeResult, makeWarnings) <-
+    P.runMake purescriptOptions
+      $ runSupplyT 0
```
rhendric, line 55: In the normal compilation path, the codegen supply monad is initialized with the next unused number from previous parts of the compilation. Using 0 here raises the question of whether this reuse is necessary. If so, using 0 here might cause problems. If not (I suspect not), we should probably be consistent so that the produced code isn't different when generated just by `purs compile` versus `purs compile; purs codegen`.

So assuming it's safe to do so, I think we should remove the `SupplyT` from the signatures in `MakeActions` and push that detail into their implementations. But now would be a really good time for someone else to share why that wouldn't be safe!

app/Command/Codegen.hs
```diff
@@ -0,0 +1,107 @@
...
+      ]
+    exitFailure
+  mods <- traverse parseCoreFn inputFiles
+
+  let
+    filePathMap =
+      M.fromList $ map ((\m -> (CoreFn.moduleName m, Right $ CoreFn.modulePath m)) . snd) $ rights mods
+
+  unless (null (lefts mods)) $ do
+    _ <- traverse (hPutStr stderr . formatParseError) $ lefts mods
```
rhendric, line 49: `hlint` will yell at you for this when you rebase on master. Use `traverse_`.

app/Command/Codegen.hs
```diff
@@ -0,0 +1,107 @@
...
+      Left (j, e) -> pure $ Left (file, j, e)
+      Right r -> pure $ Right r
+
+  makeActions foreigns filePathMap = P.buildMakeActions codegenOutputDir filePathMap foreigns False
+
+  purescriptOptions :: P.Options
+  purescriptOptions = P.Options False False (S.fromList [ P.JS ])
+
+  runCodegen foreigns filePathMap m =
+    P.codegenJS (makeActions foreigns filePathMap) False m
```
rhendric, line 74: `hlint` will yell at you for this too, but actually I think you should probably just inline this whole definition.

rhendric (commented):
Looking good! I have one outstanding question about the `runSupplyT 0` here, and I suggest rebasing on master to clean up some conflicts and HLint nits, but I think this is in great shape already.

app/Command/Codegen.hs
```diff
@@ -0,0 +1,113 @@
...
+    Opts.short 'o'
+      <> Opts.long "output"
+      <> Opts.value "output"
+      <> Opts.showDefault
+      <> Opts.help "The output directory"
+
+  sourceMaps :: Opts.Parser Bool
+  sourceMaps =
+    Opts.switch $
+      Opts.long "json-errors" <>
```
rhendric, line 112:
```suggestion
      Opts.long "source-maps" <>
```

app/Command/Codegen.hs
colinwahl: This is a great point. I tried reading through the usages of the supply monad in the codegen code and it looks like it's just for generating fresh variable names - it doesn't seem to me that it'd require we start off from where we left off in say typechecking - but I don't have enough experience to say for sure.

At work we've been using `zephyr` for quite a while, which also starts from 0 for codegen, so I'd be really surprised if it causes errors.

If it is the case that it doesn't matter, then I'll remove the `SupplyT` requirement and start it from zero within `codegenJS`.

colinwahl: I spent some time going through the `codegenJS` implementation and I don't think that it is dangerous to always initial that supply with 0. The fact that [zephyr was doing that](https://github.com/coot/zephyr/blob/master/app/Command/Run.hs#L296) for so long also makes me pretty confident based on my personal experience.

I went ahead and made the change you suggested, and all tests are passing.

If anyone knows more than I do and thinks that we should undo the change, I can do that too!

colinwahl:
I'll spend some time thinking about how we could add a meaningful test for this soon. 

Other than that, there is the outstanding question of initializing the codegen supply with 0, which I've gone ahead and done. Then this should be ready for a final review!

app/Command/Codegen.hs
```diff
@@ -0,0 +1,112 @@
...
+      ]
+    exitFailure
+  mods <- traverse parseCoreFn inputFiles
+
+  let
+    filePathMap =
+      M.fromList $ map ((\m -> (CoreFn.moduleName m, Right $ CoreFn.modulePath m)) . snd) $ rights mods
+
+  unless (null (lefts mods)) $ do
+    traverse_ (hPutStr stderr . formatParseError) $ lefts mods
```
JordanMartinez, line 50: Since `left mods` is used twice, perhaps this should be turned into a `let` above?
```haskell
let errList = lefts mods

Also, since `filePathMap` isn't used until after the `unless` block, perhaps it should go below w this block but above the `foreigns <- P.inferForeignMoudles filePathMap` line?

File: app/Command/Codegen.hs [hunk 12]
```diff
@@ -0,0 +1,112 @@
...
+    exitFailure
+
+  foreigns <- P.inferForeignModules filePathMap
+  (makeResult, makeWarnings) <-
+    P.runMake purescriptOptions
+      $ traverse (P.codegenJS (makeActions foreigns filePathMap) codegenSourceMaps . snd)
+      $ rights mods
```
JordanMartinez, line 57: Here's a second `rights mods`. Perhaps that should also be moved to a `let` binding so more things can reuse it?

MaybeJustJames:
@colinwahl do you have bandwidth to finish this off? Can I help?

MaybeJustJames:
I've rebased on current master [here](https://github.com/MaybeJustJames/purescript/tree/codegen).

colinwahl:
@MaybeJustJames would you like to take this over from me? My bandwidth for compiler work is pretty low these days.

The big open question is that I'm still not sure if https://github.com/purescript/purescript/pull/4092#discussion_r644214492 could lead to any problems.

rhendric:
As the question is over a year old, I'm inclined to say let's ship it and find out.

MaybeJustJames:
Happy to take over. How would you like to do it?

colinwahl:
> Happy to take over. How would you like to do it?

However you'd like to do it is fine with me - you could continue off this PR, or make a new branch and cherry-pick my changes, or just close it and start again from scratch. Let me know what you decide and if I should close this PR!

MaybeJustJames:
> However you'd like to do it is fine with me - you could continue off this PR

IMHO it would be a shame to lose the context here. Could I get write permission to your branch so I can pick up from here?

f-f:
I was wondering - is this work necessary at all now that we have [the backend optimizer](https://github.com/aristanetworks/purescript-backend-optimizer)?

colinwahl:
> I was wondering - is this work necessary at all now that we have [the backend optimizer](https://github.com/aristanetworks/purescript-backend-optimizer)?

Supporting an optimizer was certainly my main goal for this - now that we've got `purescript-backend-optimizer`, I don't think I'd use the command (at least, I don't have anything in mind ATM). However, maybe someone's got other ideas :)

rhendric:
It does strike me as a loss if the best general-purpose JavaScript backend for PureScript remains in a third-party project in the long term. I'm not sure how exactly this happened—I suspect the friction to contributing to PureScript is just too high for this level of innovation—but with enough time I would hope it can be mostly unforked. At that point, exposing the backend used by `purs` becomes a feature of interest again, unless the unforking includes some other mechanism for making the CoreFn-handling pipeline extensible.

f-f:
@rhendric I agree with you - my point here is that the new backend has shuffled the landscape quite a bit: it shows not only that it's possible to aggressively optimise the CoreFn, but also that it's possible to emit more performant JS outside of the compiler, and all of this while the implementation is in PureScript.
Given this new perspective, I am suggesting that we should reconsider the premises for this command to exist at all - with the baseline being that everything that is exposed by the compiler is a public API that we can't deprecate easily, e.g. see how long it took to remove `bundle` - and if the new project offers a better way to achieve the goal of this work (that is: generate better JS).
I am sure that this work will be unforked in the long term, hopefully while lowering the barrier for contribution, for example by showing that we can implement chunks of the compiler (or even all of it) in PureScript itself.

rhendric:
Okay yeah, I agree with looking at `bundle` as an example. We got rid of `bundle` when the ecosystem around ES modules matured enough and we did enough work on our codegen that we could recommend another no-regrets tool to replace it; waiting for those things to happen was what made deprecating `bundle` take so long, as far as I know.

Is `purs-backend-es` already that no-regrets tool for `codegen`? It's very impressive but also very young and possibly more aggressive than some of our users want. If it becomes that tool in the future, I don't see a significant barrier to ripping `codegen` back out, along with all the codegen internals. Just like with `bundle`, we'll paper over the switch in `spago` and basic users won't need to be aware of it.

In the meantime, as long as there's some value in having a built-in JS backend (regardless of the language in which the backend is written or the repo in which it lives), I think there's still a case for exposing it, so users can benefit from custom optimizations and rewrites without needing to also use a third-party backend.

natefaubion:
> Is purs-backend-es already that no-regrets tool for codegen?

purs-backend-es does not subsume compiler functionality.

* It is not incremental. If we want to separate out the backend from the core compiler, then the compiler must be able to inform backends of incremental status, otherwise all backends have to duplicate the work the compiler has already done to sort out what needs to be built, which is a complete waste of non-trivial work and resources. I would like to make it incremental in the interim by just depending on cache-db.json, whether or not the compiler considers it a stable target because it's the only realistic way to get that information.
* It does not emit source maps. I personally have no intention of ever implementing this without near unanimous support from the community that it's something that people use since I consider the power-to-weight-ratio to be extremely poor.

So, I do not see any near term future where the current JS backend is rendered obsolete, though I would like a near term future where something like purs-backend-es can be used in a first-class way. That being said, if there are currently no pending users of this feature, I'm not sure what the point is. I think the fresh name issue seems like it can clearly cause a problem, however unlikely, and I'm not sure how you'd fix it. I don't know how I feel about merging a feature with a uncertain prospects and potentially buggy behavior.

I'm happy to talk about `purs-backend-es` background/motivation in general, but I don't think this is the place. If you have any thoughts or questions, I'd love to hear from you on discourse!

MaybeJustJames:
> I think there's still a case for exposing it, so users can benefit from custom optimizations and rewrites without needing to also use a third-party backend.

I agree. I wanted to get this through for [zephyr](https://github.com/purescript/purescript/issues/2970#issuecomment-1256990465) specifically. There is potential for other optimization tools to make use of this interface even if an improved backend is eventually merged.

JordanMartinez:
I think this PR can be closed, right?

MaybeJustJames:
I would still vote to merge for the `zephyr | ${other_optimizer}` use case.

MaybeJustJames:
Is the vision for purescript to have multiple codegen backends? If non-javascript backends are always going to be separate projects then maybe to makes sense for JavaScript codegen to be separate too? In which case this PR should be closed. If the vision for the compiler is to include multiple backends then I think a `codegen` command will remain useful

PR #4093, opened 2021-05-24
WIP: CSS variables + dark theme

toastal:
**Description of the change**

In an attempt to support users' preferences as well as making theming/overriding easier, I've started with this WIP draft PR using [approach #3 I left on Discourse](https://discourse.purescript.org/t/pursuit-styling-variables-dark-mode/2333/4). I went with that approach because it doesn't change tooling and maintains backwards compatibility with both old browsers, and up-coming and lightweight browsers that don't support CSS3 (and given than columns are made of floats and there not even a mention to `flex`, this must be the target). This particular dark theme is targeting for OLED screen to save power (so background is `#000` black).

This is a WIP because I have a number of questions

1. Does anyone even care? There was a tiniest amount of interest on Discourse but no :-1: or :+1: from anyone of the top-contributors category
2. Is this approach 3 the best?
3. What is the code syntax highlighting supposed to look like? New completely, or just tweak existing colors?
4. Where do I find the `light_glyph` and the error messages, etc. on the variables missing? I believe this is being used in a couple of places, but I can't remember where.

---

**Checklist:**

### DIFF
--- a/app/static/pursuit.css
+++ b/app/static/pursuit.css
@@ -47,6 +47,36 @@
  ** ************************************************************************* */
 /* Section: Variables
  * ========================================================================== */
+:root {
+  --background: #ffffff;
+  --foreground: #000000;
+  --separator: #cccccc;
+  --banner_background: #1d222d;
+  --package_banner_background: #3b3f4c;
+  --dark_foreground: #f0f0f0;
+  --link: #c4953a;
+  --link_active: #7b5904;
+  --error_background: #fff0f0;
+  --error_border: #c85050;
+  --not_available_background: #f0f096;
+  --not_available_border: #e3e33d;
+  --code_foreground: #194a5b;
+  --code_background: #f1f5f9;
+  --light_glyph: #a0a0a0;
+  --light_type: #666666;
+}
+@media (prefers-color-scheme: dark) {
+  :root {
+    --background: #000000;
+    --foreground: #ffffff;
+    --separator: #3c3c3c;
+    --link: #deaa45;
+    --link_active: #986416;
+    --code_background: #121212;
+    --code_foreground: #fcfcfc;
+    --light_type: #aaaaaa;
+  }
+}
 /* Section: Document Styles
  * ========================================================================== */
 html {
@@ -63,7 +93,9 @@ html {
 }
 body {
   background-color: #ffffff;
-  color: #000;
+  background-color: var(--background, #ffffff);
+  color: #000000;
+  color: var(--foreground, #000000);
   font-family: "Roboto", sans-serif;
   font-size: 87.5%;
   line-height: 1.563;
@@ -71,6 +103,21 @@ body {
 @media (min-width: 38em) {
   body {
     font-size: 100%;
+  }
+}
+@media (prefer-color-scheme: dark) {
+  body {
+    font-weight: 350;
+  }
+  b,
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6,
+  strong {
+    font-weight: 650;
   }
 }
 /* Section: Utility Classes
@@ -156,7 +203,9 @@ body {
   width: 100%;
   text-align: center;
   background-color: #1d222d;
+  background-color: var(--banner_background, #1d222d);
   color: #f0f0f0;
+  color: var(--dark_foreground, #f0f0f0);
 }
 .footer * {
   margin-bottom: 0;
@@ -168,28 +217,36 @@ body {
  * ========================================================================== */
 :target {
   background-color: #f1f5f9;
+  background-color: var(--code_background, #f1f5f9);
 }
 a,
 a:visited {
   color: #c4953a;
+  color: var(--link, #c4953a);
   text-decoration: none;
   font-weight: bold;
 }
 a:hover {
   color: #7b5904;
+  color: var(--link_active, #7b5904);
   text-decoration: none;
 }
 code,
-pre {
+kbd,
+pre,
+samp {
   background-color: #f1f5f9;
+  background-color: var(--code_background, #f1f5f9);
   border-radius: 3px;
   color: #194a5b;
+  color: var(--code_foreground, #194a5b);
   font-family: "Roboto Mono", monospace;
   font-size: 87.5%;
 }
 :target code,
 :target pre {
-  background-color: #dfe8f1;
+  background-color: #eaeef2;
+  background-color: var(--code_background_targeted, #eaeef2);
 }
 code {
   padding: 0.2em 0;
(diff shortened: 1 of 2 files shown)


hdgarrood:
Thanks for the PR, but I personally don't care about this enough / don't think it would be such a good idea for us to take on the additional maintenance cost of officially supporting a dark theme to help get this merged. My inclination is to let people do this via userscripts or similar. If another maintainer wants to volunteer to be responsible for this and help get this merged though, then that's absolutely fine by me.

toastal:
I figured this _could_ be the case. In which case, I _do_ think doing native CSS variables and freeing up the project from Less's (broken) color functions (e.g. `darken`/`lighten`) would be a useful thing. It would help anyone needing to do theming -- and possibly removing `lessc` altogether.

hdgarrood:
Sorry, but I'm not keen on requiring CSS variable support; breaking these styles for 5% of the web feels too high for me (https://caniuse.com/css-variables). Is there a good way of compiling CSS variables away into CSS similar that Less would emit? If not, that's a dealbreaker for me. Also, I don't really care about `darken`/`lighten` not being CIE-based.

garyb:
I think switching to CSS variables is fine with the 95% support - it's a resource for people who are likely to be aware and able to choose their browser, not like a corporate locked down environment where they're trapped with old software or something.

wryk:
@hdgarrood [PostCSS](https://github.com/postcss/postcss) is usually a good choice for css processing and plugins like [postcss-css-variables](https://github.com/MadLittleMods/postcss-css-variables) and [postcss-custom-properties](https://github.com/postcss/postcss-custom-properties) are available

But I don't know if it's really necessary because the primary audience are developers

hdgarrood:
I'm still uneasy with not doing any preprocessing; I've definitely heard of developers who are very aware of the shortcomings of older browsers but are still in locked down corporate environments where they are stuck on those browsers anyway. Switching to PostCSS would be okay with me though.

toastal:
BTW, currently the Less implementation I was using, we support _both_ browser with CSS variables _and_ those without _while_ making it easy to theme/override.

But if there's anything that needs a PostCSS implementation whipped up, I can/could do it. I'd just need a little direction on the best way to document that step (i.e. is a `.postcssrc.json` file enough to describe the dependencies and process? does there need to be a `package.json`?)

toastal:
This doesn't need to be overthought. The main goal was to make it easy to theme and possibly provide a default dark theme as tech people tend to notice and look favorably on technologies that pay attention to details like supporting their preferences.

wclr:
I believe, first, there needs to be a desire of the maintainers to add the ability to switch themes on the pursuit site. Until that such changes probably won't be seen as necessary and valuable. Though I think css-variables is the most simple and convenient way to develop and support css-based theming.

JordanMartinez:
> I believe, first, there needs to be a desire of the maintainers to add the ability to switch themes on the pursuit site. Until that such changes probably won't be seen as necessary and valuable.

:point_up: 

Do I think it is beneficial to have a dark theme? Sure. Is it necessary? As far as I can tell, no. It may be inconvenient to not have it, but (correct me if I'm wrong) one can still use Pursuit for their purposes. 

Here are my questions (some of which I believe @toastal answered in the Discourse post but which I thought would be helpful to ask explicitly):
1. If the goal is to support a dark theme on Pursuit, what are the alternative approaches that could be done? When do these alternatives work and when do they not? For example, could one install a browser extension that grants users the ability to define their own 'dark theme'?
1. If maintenance is a concern here, what specifically does that mean? Before this PR existed, what 'maintenance' cost are we currently paying? If this PR was accepted, what 'maintenance' cost might we end up paying?

hdgarrood:
@toastal originally said in the PR description that one of the aims is to "make theming/overriding easier" - so the way I understood that was that this change is potentially still useful even if there is no desire to _officially_ support a dark Pursuit theme (which, at least from me, there isn't), because it would make it easier for other developers to do it on their own.

The maintenance cost is that by providing the option of using dark mode, there are more ways for bugs to creep in via the CSS, and most CSS changes need twice as much testing (because they need to be tested in each mode); maybe something is clearly visible in light mode but not in dark mode, and the developer has only tested in one of the two modes.

toastal:
Agreed on all of the points. To be clear how this got here: I went to file a bug asking if a dark theme would be something worthwhile, but GitHub says for this type of request/bug/issue to post out on Discourse. After 6 days and two hearts and no feedback from "y'all", the ones _actually_ maintaining this stuff, I figured "ah, what the heck" and started a PR specifically because then I'd get answer. Which it seems: dark theme, nice to have but likely costs outweighing benefits.

The if pivoted solution (still WIP) Less+CSS variables additions I suggest
- visually affect nothing in the status quo
- does not change tooling
- add a small amount of manual duplication to add a new variable (specifically for browsers not supporting variables)
- _does_ 'variablize' a few spots that probably should have been
- make theming an easier possibility as UserStyles do not need to consider the CSS structure and selectors
- lays down a semblance of a future where `lessc` and possibly all CSS-related compilers were removed
- if browser support feelings changed or more than a lukewarm desire dark mode or alternate themes existed, it's there

If it's all ceremony, thanks for getting back to me. I won't finish either a dark theme or the pivot without further enthusiasm. No hard feels if the concept is denied.

rhendric:
Less has some useful indirection features which could be used to really minimize the boilerplate for this:

```less
.theme-prop(@property, @name) {
  @{property}: @@name;
  @{property}: var(~"--purs-@{name}", @@name);
}

@background: rgb(255, 255, 255);
@foreground: #000;

body {
  // replace this:
  // background-color: @background;
  // color: @foreground;

  // with this:
  .theme-prop(background-color, background);
  .theme-prop(color, foreground);
}
```

which compiles to

```css
body {
  background-color: #ffffff;
  background-color: var(--purs-background, #ffffff);
  color: #000;
  color: var(--purs-foreground, #000);
}
```

So that's an option. Very low-touch from a maintenance standpoint, no impact on older browsers.

toastal:
The thing I don't care for about that is adding additional reliance on Less. Honestly I didn't know about those features since I've not touched Less in years, but, frankly, I don't want to know about Less in the context of the PureScript compiler.

I think the way to go would be going to pure CSS or as close as one can get to reduce tooling and languages-to-know by doing it manually (this case via `lessc` assistance purely for its variables) or adding a tool, like `postcss-preset-env`, that would, without thought, automatically backport contemporary pure CSS for the old and minor browser audience. Moving to Less + CSS variables puts us somewhere in the middle -- where you still need to know an additional tool more than at a surface level and introducing functions brings that coupling even tighter.

garyb:
Regardless of whatever we're doing in regards to theming, I'd be in favour of stopping using `less` too.

JordanMartinez:
Superceded by #4438. Although this didn't get merged, thank you @toastal for contributing.

PR #4090, opened 2021-05-14
Follow more HLint suggestions

rhendric:
**Description of the change**

We're ignoring quite a large number of HLint's warnings/suggestions, and I thought it might be a good idea to do something about that.

Here it is, one commit per line removed from `.hlint.yaml` (except for the first one). I've grouped the commits into sort-of themes and ordered them approximately from what I expect will be least to most controversial (bearing in mind that there are a few suggestions left that even I didn't think would lead to improvement). A handy reference list is below, so that each one can be quibbled about individually. Maybe it would be worth it to break some of these out into smaller PRs? Should each theme be its own PR? I don't know. Let's start talking.

Commits touching more than three code files are **bolded**.

#### :brain: The no-brainers
[`0696af28`](https://github.com/purescript/purescript/pull/4090/commits/0696af28714588cea055440a797ea7184f214473) Remove unneeded HLint ignore directives
[`a7fa8610`](https://github.com/purescript/purescript/pull/4090/commits/a7fa8610f3a3a4d32a5d53ff14aec58ad3fe612b) HLint fix: "Parse error"
**[`0fd6f53f`](https://github.com/purescript/purescript/pull/4090/commits/0fd6f53ff472b8ec83d8d5141dc3c94b9b81285a) HLint fix: "Unused LANGUAGE pragma"**

#### :robot: Boolean hijinks
[`50b475c0`](https://github.com/purescript/purescript/pull/4090/commits/50b475c0802c1f4d2bb8f7da0db9ee7436919554) HLint fix: "Redundant =="
[`b6a2aef8`](https://github.com/purescript/purescript/pull/4090/commits/b6a2aef8fc570741e8fc4e654299ff2a6f374b8f) HLint fix: "Redundant if"

#### :paperclip: Did you mean...
[`c14cb8c1`](https://github.com/purescript/purescript/pull/4090/commits/c14cb8c1132ac782cfbd413428c9162422df6603) HLint fix: "Use notElem"
[`667d083a`](https://github.com/purescript/purescript/pull/4090/commits/667d083a38c61e5e3657b9a0eb365845a8c67bb7) HLint fix: "Use String"
[`b8784430`](https://github.com/purescript/purescript/pull/4090/commits/b8784430dc8b489a25f4b9b4afac46cea6ac44bd) HLint fix: "Use isDigit"
[...]

PR #4094, opened 2021-05-25
Merge purescript-ast into purescript-cst

### LINKED ISSUE #4067: Merge purescript-ast and purescript-cst
## Summary

Merge purescript-ast and purescript-cst into one package in order to simplify maintenance and reduce the risk of users building compilers with untested combinations.

## Motivation

Having the compiler spread across three packages makes maintenance - specifically releases - harder (see #4061), and also increases the risk of users building versions of the compiler with untested combinations. For example, v0.14.0 currently has a version bound of >=0.1.0.0 <0.2 on purescript-cst and purescript-ast, which means that if we make a 0.1.1.0 release of purescript-cst and/or purescript-ast, people building the compiler from source could pick up that new version, which opens the door for surprising or incorrect behaviour. The more inter-package dependencies there are, the more opportunities there are for this to happen. (Aside: after writing this issue up I'm going to make a revision on Hackage to pin these bounds down to a specific version.)

As far as I am aware nobody is depending on purescript-ast directly (versus just depending on purescript-cst) anyway. Merging purescript-ast and purescript-cst into a single package will simplify maintenance and still provide almost all of the benefits of having the parser published as a separate package, since people will still be able to depend on purescript-cst and create tooling which depends on the parser without having to build the whole compiler together with all of its own dependencies.

## Proposal

At the moment, the dependency graph of the three packages that make up the compiler is as follows:

```
       ┌────────────┐
       │ purescript │
       └───┬───────┬┘
           │       │
┌──────────▼───┐   │
│purescript-cst│   │
└──────────┬───┘   │
           │     ┌─▼────────────┐
           └─────►purescript-ast│
                 └──────────────┘
```

I propose we merge purescript-cst and purescript-ast into a single package, so that we would have:

```
 ┌──────────┐
 │purescript│
 └────┬─────┘
      │
┌─────▼────────┐
│purescript-cst│
└──────────────┘
```

JordanMartinez:
**Description of the change**

Fixes #4067 to make releases a bit easier.

---

**Checklist:**


JordanMartinez:
I think #4090 should be merged before this one. This one is easy to reimplement but 4090 isn't.

wclr:
Wouldn't it be a miscommunication to put AST and CST under the name `purescript-cst`?

hdgarrood:
I don't think it would be really that much worse than it is currently, since `purescript-cst` already depends on `purescript-ast`. We could rename it to `purescript-syntax` but that could be annoying for people who currently depend on `purescript-cst`, since if we don't change the name from `purescript-cst` then downstream code likely won't need to change.

JordanMartinez:
I've updated this PR to work on top of current `master` now that #4090 has been merged. CI passed.

lib/purescript-cst/README.md
```diff
@@ -1,12 +1,13 @@
 # purescript-cst
 
-Defines the surface syntax of the PureScript Programming Language.
+Defines the surface and underlying syntax of the PureScript Programming Language.
```
hdgarrood, line 3: Now that these packages are merged, I think talking about a distinction between a "surface" and "underlying" syntax is probably unnecessary, and I suspect it won't be clear to lots of readers, and they probably don't need to care about the difference in lots of cases anyway, at least not immediately. How about "The parser for the PureScript programming language"?

lib/purescript-cst/README.md
```diff
@@ -1,12 +1,13 @@
...
 
 ## Compiler compatibility
 
-We provide a table to make it a bit easier to map between versions of `purescript` and `purescript-cst`.
+We provide a table to make it a bit easier to map between versions of `purescript`, `purescript-cst`, and `purescript-ast`.
```
hdgarrood, line 7: I think now is a good opportunity to put this table in reverse order, so that the latest versions (which people are more likely to care about) are at the top. Also, I think it would be nice to have a separate table following the merge of ast and cst, so that there isn't an extra column. Maybe something like this?

---

We provide a table to make it a bit easier to map between versions of `purescript` and `purescript-cst`.

| `purescript` | `purescript-cst` |
| --- | --- |
| 0.14.2 | 0.2.0.0 |

Before v0.14.2, there was a third package, `purescript-ast`. In v0.14.2, `purescript-ast` was merged into `purescript-cst`.

| `purescript` | `purescript-cst` | `purescript-ast` |
| --- | --- | --- |
| 0.14.1 | 0.1.1.0 | 0.1.1.0 |
| 0.14.0 | 0.1.0.0 | 0.1.0.0 |

lib/purescript-cst/purescript-cst.cabal
```diff
@@ -3,7 +3,7 @@ cabal-version: 2.4
 name:           purescript-cst
 version:        0.1.1.0
 synopsis:       PureScript Programming Language Concrete Syntax Tree
-description:    The surface syntax of the PureScript Programming Language.
+description:    The surface and underlying syntax of the PureScript Programming Language.
```
hdgarrood, line 6: If we change the README according to my comment there, let's make the same change here.

lib/purescript-cst/README.md
JordanMartinez: > How about "The parser for the PureScript programming language"?

Isn't this more than just the parser? Does one ever operate directly on the AST via some tool?

I'll make the change, but I'm just expressing my confusion as to why `syntax` is no longer mentioned when it seems like that's what this resulting library is.

lib/purescript-cst/README.md
JordanMartinez: I like this idea!

lib/purescript-cst/purescript-cst.cabal
JordanMartinez: Done.

JordanMartinez:
Also, feel free to make any edits necessary to get this merged if I don't respond in time.

lib/purescript-cst/README.md
hdgarrood: It has type definitions for the syntax, sure, but you can’t really write a parser without having those in the first place. Maybe I’m overthinking it though. I guess “defines the syntax for …” would work too.

hdgarrood (approved):

lib/purescript-cst/README.md
JordanMartinez: I think `syntax and parser` would probably work, but... :upside_down_face: I already merged this. 

We can fix it later if someone gets confused down the road.

Issue #3724, opened 2019-08-28
Don't require rebuilds for downstream modules if interface is unchanged


hdgarrood:
Currently, any change to a module will force all downstream modules to be recompiled. For example:

```purescript
-- src/Upstream.purs
module Upstream where
info = 0

-- src/Downstream.purs
module Downstream where
import Upstream (info)
result = { info }
```

Suppose we compile this, and then update `Upstream` as follows:

```patch
--- a/src/Upstream.purs
+++ b/src/Upstream.purs
@@ -1,2 +1,2 @@
 module Upstream where
-info = 0
+info = 1
```

Then, the next time we compile, both `Upstream` and `Downstream` will be recompiled: `Upstream` because the source file changed, and `Downstream` because an upstream module changed. However, the changes to `Upstream` do not necessitate any changes to `Downstream`. I think skipping recompiling downstream modules has the potential to speed builds up quite dramatically in certain cases.

This will need a bit of careful thought and some changes to externs file formats, as I wrote in https://github.com/purescript/purescript/issues/3145#issuecomment-491803230:

> Currently we implicitly depend on the current behaviour that if a module is rebuilt then all of its downstream modules must be rebuilt too in a few places, in the sense that information which is more than one hop away in the module dependency graph can leak through into the output artifacts. If we want to try to make things more granular we will need to address this first. The first example which comes to mind is re-exports, as re-exports are currently always generated as coming from the module they were defined in (as opposed to the module they were locally imported from). If module M1 exports foo, module M2 re-exports foo, and M3 imports foo from M2, then in the generated code for M3 we will have an import of M1. This means that if foo later moves to a different module (but continues to be re-exported from M2), and if M2 is determined not to need rebuilding, we could end up in a situation where the generated code for M3 still refers to M1.foo.

> At the moment, when a module gets rebuilt, it has access to the externs files for every module which it depends on (directly or transitively). If/when we have addressed the above, and to ensure we've addressed this properly once and for all, I think we should try to have our build system only provide externs for the direct dependencies of any given module when that module is being built; that way, we make it easier not to end up accidentally leaking information through the module graph transitively, which then makes it easier to provide fast and safe incremental builds.

timjs:
This is wat [Elm is going to do in 0.19.1](https://discourse.elm-lang.org/t/community-beta-for-elm-0-19-1/4102/27). Would be awesome. Apparently the speed bumps are big!

hdgarrood:
[...]

Issue #3323, opened 2018-04-26
Ide rebuild can create invalid JS with outdated cache


kRITZCREEK:
1. Compile and load these 2 modules into the ide server:
```haskell
module A where

fn :: Int -> String
fn x = show x
````
```haskell
module B where

import A (fn)

bFn :: String
bFn = fn 1
```

2. Change module `A.fn` to
```haskell
fn :: String -> String
fn s = s <> "hi!"
```

3. Recompile A without letting the server know (start it with the `--editor-mode` flag and recompile module A from the cmdline with `purs compile src/A.purs`)

4. Make a change to module `B` and fast-rebuild through ide, it does not show any errors, and it continues to generate code for `B` that assumes the old signature of `A.fn` is still valid, but instead the code fails at runtime.

This is a known issue, and I do not have a satisfying solution for the problem at hand. It basically boils down to cache invalidation between the FS and the ide servers internal state. Empirically file watching misses changes, and doesn't interact well with adding and removing files and avoiding it by polling creates too much overhead.

i-am-the-slime:
When I rename a module a few times I always get outdated suggestions/completions from the previous module names. 
<img width="435" alt="image" src="https://user-images.githubusercontent.com/1588055/121886772-22197380-cd16-11eb-8291-3e66e1897c7d.png">
I'm also worried about this leaking.

Since `3.16.0` there have been hooks in the LSP for when files are added/removed: "Events for file operations (create, rename, delete)". Could we use these to signal to the IDE server when something needs invaliding?

PR #4096, opened 2021-05-29
Use GenIdent for anonymous instances

rhendric:
This fixes the limitation of the CoreFn renamer which prevented it from
renaming top-level GenIdents. As a consequence, we can now give unnamed
instances more idiomatic names and still guarantee that they will be
unique in their module.

**Description of the change**

With compliments to @JordanMartinez, follow-up from one of my suggestions on #4085.

---

**Checklist:**

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
@@ -547,10 +547,11 @@ convertDeclaration fileName decl = case decl of
       genName = Text.take 25 (className <> typeArgs)
 
       className :: Text.Text
-      className = foldMap (uncurry Text.cons . first toLower)
-                . Text.uncons
-                . N.runProperName
-                $ qualName cls
+      className
+        = foldMap (uncurry Text.cons . first toLower)
+        . Text.uncons
+        . N.runProperName
+        $ qualName cls
 
       typeArgs :: Text.Text
       typeArgs = foldMap argName args
--- a/lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
@@ -18,6 +18,7 @@ module Language.PureScript.CST.Convert
 import Prelude hiding (take)
 
 import Data.Bifunctor (bimap, first)
+import Data.Char (toLower)
 import Data.Foldable (foldl', toList)
 import Data.Functor (($>))
 import qualified Data.List.NonEmpty as NE
@@ -540,13 +541,16 @@ convertDeclaration fileName decl = case decl of
     where
       -- truncate to 25 chars to reduce verbosity
       -- of name and still keep it readable
-      -- unique identifier will be appended to this name
-      -- in desugaring proces
+      -- name will be used to create a GenIdent
+      -- in desugaring process
       genName :: Text.Text
-      genName = "$_" <> Text.take 25 (className <> typeArgs) <> "_"
+      genName = Text.take 25 (className <> typeArgs)
 
       className :: Text.Text
-      className = N.runProperName $ qualName cls
+      className = foldMap (uncurry Text.cons . first toLower)
+                . Text.uncons
+                . N.runProperName
+                $ qualName cls
 
       typeArgs :: Text.Text
       typeArgs = foldMap argName args
--- a/src/Language/PureScript/Renamer.hs
+++ b/src/Language/PureScript/Renamer.hs
@@ -1,14 +1,15 @@
 -- |
 -- Renaming pass that prevents shadowing of local identifiers.
 --
-module Language.PureScript.Renamer (renameInModules) where
+module Language.PureScript.Renamer (renameInModule) where
 
 import Prelude.Compat
 
 import Control.Monad.State
 
+import Data.Functor ((<&>))
 import Data.List (find)
-import Data.Maybe (fromJust, fromMaybe)
+import Data.Maybe (fromJust, fromMaybe, isNothing)
 import qualified Data.Map as M
 import qualified Data.Set as S
 import qualified Data.Text as T
@@ -39,8 +40,8 @@ initState scope = RenameState (M.fromList (zip scope scope)) (S.fromList scope)
 -- |
 -- Runs renaming starting with a list of idents for the initial scope.
 --
-runRename :: [Ident] -> Rename a -> a
-runRename scope = flip evalState (initState scope)
+runRename :: [Ident] -> Rename a -> (a, RenameState)
+runRename scope = flip runState (initState scope)
 
 -- |
 -- Creates a new renaming scope using the current as a basis. Used to backtrack
@@ -93,48 +94,68 @@ lookupIdent name = do
     Just name'' -> return name''
     Nothing -> error $ "Rename scope is missing ident '" ++ T.unpack (showIdent name) ++ "'"
 
--- |
--- Finds idents introduced by declarations.
---
-findDeclIdents :: [Bind Ann] -> [Ident]
-findDeclIdents = concatMap go
-  where
-  go (NonRec _ ident _) = [ident]
-  go (Rec ds) = map (snd . fst) ds
 
 -- |
--- Renames within each declaration in a module.
+-- Renames within each declaration in a module. Returns the map of renamed
+-- identifiers in the top-level scope, so that they can be renamed in the
+-- externs files as well.
 --
-renameInModules :: [Module Ann] -> [Module Ann]
-renameInModules = map go
+renameInModule :: Module Ann -> (M.Map Ident Ident, Module Ann)
+renameInModule m@(Module _ _ _ _ _ exports _ foreigns decls) = (rsBoundNames, m { moduleExports, moduleDecls })
   where
-  go :: Module Ann -> Module Ann
-  go m@(Module _ _ _ _ _ _ _ _ decls) = m { moduleDecls = map (renameInDecl' (findDeclIdents decls)) decls }
-
-  renameInDecl' :: [Ident] -> Bind Ann -> Bind Ann
-  renameInDecl' scope = runRename scope . renameInDecl True
+  ((moduleDecls, moduleExports), RenameState{..}) = runRename foreigns $
+    (,) <$> renameInDecls decls <*> traverse lookupIdent exports
 
 -- |
--- Renames within a declaration. isTopLevel is used to determine whether the
--- declaration is a module member or appearing within a Let. At the top level
--- declarations are not renamed or added to the scope (they should already have
--- been added), whereas in a Let declarations are renamed if their name shadows
--- another in the current scope.
+-- Renames within a list of declarations. The list is processed in three
+-- passes:
 --
-renameInDecl :: Bool -> Bind Ann -> Rename (Bind Ann)
-renameInDecl isTopLevel (NonRec a name val) = do
-  name' <- if isTopLevel then return name else updateScope name
-  NonRec a name' <$> renameInValue val
-renameInDecl isTopLevel (Rec ds) = do
-  ds' <- traverse updateNames ds
-  Rec <$> traverse updateValues ds'
+--  1) Declarations with user-provided names are added to the scope, renaming
+--     them only if necessary to prevent shadowing.
+--  2) Declarations with compiler-provided names are added to the scope,
+--     renaming them to prevent shadowing or collision with a user-provided
+--     name.
+--  3) The bodies of the declarations are processed recursively.
+--
+-- The distinction between passes 1 and 2 is critical in the top-level module
+-- scope, where declarations can be exported and named declarations must not
+-- be renamed. Below the top level, this only matters for programmers looking
+-- at the generated code or using a debugger; we want them to see the names
+-- they used as much as possible.
+--
+-- The distinction between the first two passes and pass 3 is important because
+-- a `GenIdent` can appear before its declaration in a depth-first traversal,
+-- and we need to visit the declaration first in order to rename all of its
+-- uses. Similarly, a plain `Ident` could shadow another declared in an outer
+-- scope but later in a depth-first traversal, and we need to visit the
+-- outer declaration first in order to know to rename the inner one.
+--
+renameInDecls :: [Bind Ann] -> Rename [Bind Ann]
+renameInDecls =
+      traverse (renameDecl False)
+  >=> traverse (renameDecl True)
+  >=> traverse renameValuesInDecl
+
   where
-  updateNames :: ((Ann, Ident), Expr Ann) -> Rename ((Ann, Ident), Expr Ann)
-  updateNames ((a, name), val) = do
-    name' <- if isTopLevel then return name else updateScope name
-    return ((a, name'), val)
-  updateValues :: ((Ann, Ident), Expr Ann) -> Rename ((Ann, Ident), Expr Ann)
-  updateValues (aname, val) = (aname, ) <$> renameInValue val
+
+  renameDecl :: Bool -> Bind Ann -> Rename (Bind Ann)
+  renameDecl isSecondPass = \case
+    NonRec a name val -> updateName name <&> \name' -> NonRec a name' val
+    Rec ds -> Rec <$> traverse updateNames ds
+    where
+    updateName :: Ident -> Rename Ident
+    updateName name = (if isSecondPass == isPlainIdent name then pure else updateScope) name
+
+    updateNames :: ((Ann, Ident), Expr Ann) -> Rename ((Ann, Ident), Expr Ann)
+    updateNames ((a, name), val) = updateName name <&> \name' -> ((a, name'), val)
+
+  renameValuesInDecl :: Bind Ann -> Rename (Bind Ann)
+  renameValuesInDecl = \case
+    NonRec a name val -> NonRec a name <$> renameInValue val
+    Rec ds -> Rec <$> traverse updateValues ds
+    where
+    updateValues :: ((Ann, Ident), Expr Ann) -> Rename ((Ann, Ident), Expr Ann)
+    updateValues (aname, val) = (aname, ) <$> renameInValue val
 
 -- |
 -- Renames within a value.
@@ -152,13 +173,17 @@ renameInValue (Abs ann name v) =
   newScope $ Abs ann <$> updateScope name <*> renameInValue v
 renameInValue (App ann v1 v2) =
   App ann <$> renameInValue v1 <*> renameInValue v2
-renameInValue (Var ann (Qualified Nothing name)) =
-  Var ann . Qualified Nothing <$> lookupIdent name
+renameInValue (Var ann (Qualified mn name)) | isNothing mn || not (isPlainIdent name) =
+  -- This should only rename identifiers local to the current module: either
+  -- they aren't qualified, or they are but they have a name that should not
+  -- have appeared in a module's externs, so they must be from this module's
+  -- top-level scope.
+  Var ann . Qualified mn <$> lookupIdent name
 renameInValue v@Var{} = return v
 renameInValue (Case ann vs alts) =
   newScope $ Case ann <$> traverse renameInValue vs <*> traverse renameInCaseAlternative alts
 renameInValue (Let ann ds v) =
-  newScope $ Let ann <$> traverse (renameInDecl False) ds <*> renameInValue v
+  newScope $ Let ann <$> renameInDecls ds <*> renameInValue v
 
 -- |
 -- Renames within literals.
@@ -189,3 +214,7 @@ renameInBinder (ConstructorBinder ann tctor dctor bs) =
   ConstructorBinder ann tctor dctor <$> traverse renameInBinder bs
 renameInBinder (NamedBinder ann name b) =
   NamedBinder ann <$> updateScope name <*> renameInBinder b
+
+isPlainIdent :: Ident -> Bool
+isPlainIdent Ident{} = True
+isPlainIdent _ = False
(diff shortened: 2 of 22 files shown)


lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -540,13 +541,16 @@ convertDeclaration fileName decl = case decl of
     where
       -- truncate to 25 chars to reduce verbosity
       -- of name and still keep it readable
-      -- unique identifier will be appended to this name
-      -- in desugaring proces
+      -- name will be used to create a GenIdent
+      -- in desugaring process
       genName :: Text.Text
-      genName = "$_" <> Text.take 25 (className <> typeArgs) <> "_"
+      genName = Text.take 25 (className <> typeArgs)
```
JordanMartinez, line 547: Why is the final `_` character being removed? Is this just preference? 

I'm not for or against this as I think the larger readability problems are addressed in this PR. To me, this seems more like a 'stylistic' change than anything else.

src/Language/PureScript/Renamer.hs
```diff
@@ -109,30 +110,36 @@ renameInModules :: [Module Ann] -> [Module Ann]
...
--- been added), whereas in a Let declarations are renamed if their name shadows
--- another in the current scope.
+-- Renames within a list of declarations. isTopLevel is used to determine
+-- whether the declarations are module members or appearing within a Let. At
+-- the top level declarations are only renamed or added to the scope if they
+-- are named with GenIdents (otherwise they should already have been added),
+-- whereas in a Let declarations are renamed if their name shadows another in
+-- the current scope. In order to prevent inner bindings earlier in the list
+-- from shadowing bindings later in the list, first all of the declarations
+-- are considered for renaming, and then all of the values are recursed into.
```
JordanMartinez, line 126: > In order to prevent inner bindings earlier in the list from shadowing bindings later in the list, first all of the declarations are considered for renaming, and then all of the values are recursed into.

Could you clarify this point further because I'm not familiar with the code outside of this file?

Is this in case something like this occurs?
```purescript
foo :: String -> String
foo a =
  -- does it use top-level `a` 
  -- or foo's `a`? 
  a <> "something"

a :: String
a = "text value

**Review by JordanMartinez — COMMENTED** [review]
I'm just writing out my understanding of what this PR does, so that you can confirm/correct my understanding (and anyone following along can be brought up to speed sooner).

First, I'll cover some context / background info on the `Ident` type, the type class instance compiler-generated names and the Renamer code. Then, I'll describe the goal you're trying to achieve. Finally, I'll explain what you changed to achieve that goal.

The `Ident` type has three values:
- `Ident` - an identifier provided by the developer
- `GenIdent` - an identifier generated by the compiler
- `UnusedIdent` - this is irrelevant for this PR. It's used in typechecking. Since it doesn't store any text value, we can't "rename" it anyway.

Identifiers can be thought of as 'names' of things (e.g. a function, a value, a `let` binding). Below, every variation of `fooX` (where `X` is a number) is an identifier:
```purescript
module Module where

-- type signature omitted
-- but think `foo1 :: String -> Value`
foo1 foo2 = do
  let foo3 = --
  foo4 foo2
  where
  foo4 _ = "x"
```

Via #4085, type class instance whose names are generated currently use the `Ident` constructor when they should use the `GenIdent` constructor as it better reflects the source of their name. Such instances are currently named `$_ClassNameTypeName_4` for the following reasons:
- the leading `$` ensures it does not clash with other purescript identifiers because no such identifiers can use `$` in their name.
- the first `_` makes it easier to read the name because the `$` character distracts the eye; however, it is not necessary for the name.
- the final `_` is also not necessary, but does make things a bit nicer to read due to the unique identifier at the end.
- the `4` is a unique identifier that ensures similar instance names do not clash due to generating a name and then truncating it to 25 characters.

The `Renamer` is responsible for ensuring two or more identifiers don't clash within a given scope. For example, given the below code:
```purescript
module Main where

import ImportedModule as ImportedModule

-- top-level member
foo a {- a1 -} =
  let
    -- let binding
    a {- a2 -} = "hello"
  in a <> " world"

-- or perhaps this?
bar a {- a1 -} = a <> ImportedModule.a {- a2 -}
```
the "`a` near `{- a1 -}`" will have a different identifier in the outputted JavaScript (e.g. `a1`) than the "`a` near `{- a2 -}`" despite both having the same name in the source code.

The Renamer removes clashes by adding an integer to the "base name" (e.g. `a1`, `a2`, `a3`, etc. for identifiers named `a`) every time it finds another clash.

The Renamer code BEFORE this PR works by doing the following:
1. Extract a module's declarations via `renameInModules`
1. Create the top-level scope by getting all identifiers in the module via `findDeclIdents`
1. Use a top-down approach to recurse through each layer in the AST and rename identifiers based on what's in scope at that point via `renameInDecl`, `renameInValue`, `renameInLiteral`, and `renameInBinder`. The "scope" of the current scope is determined by the binding group (I think).
    1. Renames **both** the name **and** the value in **a single pass** (this is changed in this PR)
        1. top-level identifiers are not renamed (this is important)
        1. all other identifiers are renamed
1. Returns a new version of the module with a new version of the list of declarations with updated names

Type class instances will always be top-level identifiers in a module. However, the Renamer doesn't currently rename top-level members. If top-level members were renamed, then the module's `exports` also need to be renamed (so that exported identifiers still work when imported into other modules). To rebuild a module's `exports`, one also needs to incorporate the module's FFI (if it exists) in case such members are also exported.

Thus, here's the goal of this PR
- Make generated type class instances use the `GenIdent` constructor rather than the `Ident` constructor to accurately reflect the source of their name.
- Remove the leading `$` character from the instance name by updating the `Renamer` to ensure names don't clash by renaming top-level `GenIdent` constructors (if needed).

This is how the PR achieves the goal. This PR...
1. Changes the final name generated by the compiler for a type class instance from `$_ClassNameTypeName_4` to `classNameTypeName4`
1. Updates generated type class instance names to use `GenIdent` rather than `Ident`
1. Changes the Renamer to include top-level `GenIdent`s when renaming and returns an updated version of the module's `exports` (new) in addition to its `decls` (current).
1. Changes the Rename to first rename names **in one pass** and then rename values **in a second pass** (whereas current approach renames both names and values in the same pass).

The Renamer code AFTER this PR works by doing the following:
1. Extract a module's exports, FFI, and declarations via `renameInModules`
1. Create the top-level scope by combining 1) only the `Ident` identifiers via `findDeclIdents` and 2) the FFI identifiers in the module
1. Uses `State`'s Applicative (which executes sequentially and not in parallel) to do the following:
    1. Use a top-down approach to recurse through each layer in the AST and rename identifiers based on what's in scope at that point via `renameInDecl`, `renameInValue`, `renameInLiteral`, and `renameInBinder`. The "scope" of the current scope is determined by the binding group (I think).
        1. Renames the names **and then** the values
            1. top-level `GenIdent`s are renamed while top-level `Ident`s are not renamed
            1. all other identifiers are renamed
    1. Lookup what the new name of an identifier is (if not renamed, the original name will be returned) and create a new version of the export with the new identifier.
    1. Return the updated declaration and its corresponding export
1. Returns a new version of the module with
    1. a new version of the declarations with updated names
    1. a new version of the exports with updated names

It seems the various other changes (e.g. `renameDecl`, etc.) are to get the types to line up and other changes (e.g. `updateName`) are minor refactoring.

rhendric:
Your analysis is quite correct!

One thing that might be clarifying to add (at least, it briefly confused me just now when I read your analysis and thought I had made things too complicated) is why `renameInDecls` does two passes at every level *in addition* to the `findDeclIdents` call at the top level, which acts as sort of a zeroth mini-pass.

At the top level, we effectively do three passes because we want regular `Ident`s to have first dibs on their preferred name, then `GenIdent`s, then bindings in inner scopes. If we didn't do the `findDeclIdents` pass first, then an earlier `GenIdent` could possibly steal a name from the top-level scope that a later `Ident` would want (which would be very bad if we export that name). If we didn't do the decl-rename pass (which at the top level just means renaming `GenIdent`s), then a reference to that `GenIdent` appearing deep in a value somewhere might get visited before the `GenIdent` has its new name in scope, which could maybe cause it to get a different final name than the thing it references.

We could do three passes at every level—I'm still considering that, actually—but the importance of letting `Ident`s have first dibs over `GenIdent`s is much less when nothing is being exported.

(Please ignore the fixup commit I momentarily pushed and then retracted; as I said, I was briefly confused, and thought that two passes everywhere would work. I forgot that `Ident`s need to have priority over `GenIdent`s at least at the top level.)

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
rhendric: It is cosmetic, yes. But note that the renamer only appends numbers to the end of a `GenIdent` if necessary to disambiguate. So with the `_` character, instances were getting named `classNameArg_`, which is why I removed it.

src/Language/PureScript/Renamer.hs
rhendric: Yes, that's exactly what I'm talking about there. The `a` in `foo` is an inner binding that shadows the top-level `a`. The renamer should rename the `a` in `foo` to `a1`. I think the CoreImp optimizer expects the code it receives to have no shadowing, so it's not just a style thing (even though in JavaScript shadowing works just fine).

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
JordanMartinez: Ah, makes sense.

src/Language/PureScript/Renamer.hs
JordanMartinez: Gotcha. Thanks for clarifying!

JordanMartinez:
> We could do three passes at every level—I'm still considering that, actually—but the importance of letting Idents have first dibs over GenIdents is much less when nothing is being exported.

Sounds like the 3-pass approach is necessary for top-level members and safer (but possibly not necessary?) for non-top-level members. Is that correct? If we do the 3-pass approach, does that slightly slow down the `purs compile` command?

rhendric:
Using three passes below the top level doesn't get us any more safety. What it would get us is:

* Code simplicity: we would treat all declarations the same, which might make this code easier to read.

* Somewhat more predictable generated code in the presence of a pseudo-collision between a `GenIdent` and an `Ident`. I suspect this isn't possible to trigger right now; most if not all of the `GenIdent`s I've seen prior to this PR are function parameters and not let-bindings, and instances have to be top-level. But suppose some future language feature resulted in some CoreFn that looked like this (in pseudo-PS):

    ```purs
    let <GenIdent "x" 42> = foo ...
        x = bar ...
    in <GenIdent "x" 42> + x
    ```

    where `x` is a user-supplied proper `Ident`, and `<GenIdent "x" 42>` is an identifier that came from some desugaring somehow.

    With two passes, this code could end up being generated:

    ```js
    var x = foo(...);
    var x1 = bar(...);
    return x + x1;
    ```

    With three passes, this is what you'd get always:

    ```js
    var x1 = foo(...);
    var x = bar(...);
    return x1 + x;
    ```

    The idea is that a name chosen by the programmer has priority over a variable produced by the compiler. A programmer using a debugger might not appreciate the variable `x` meaning something other than what they used it for. It's still a safe transformation either way, just potentially a surprising one to anyone reading the generated code or otherwise looking under the hood.

    But as I said, I don't know of any way to trigger this sort of conflict below the top level right now, so this is a pretty theoretical benefit.

JordanMartinez:
Simple code is always better than complex code if no other factors are at play.

> The idea is that a name chosen by the programmer has priority over a variable produced by the compiler. A programmer using a debugger might not appreciate the variable x meaning something other than what they used it for. It's still a safe transformation either way, just potentially a surprising one to anyone reading the generated code or otherwise looking under the hood.

I think this is actually worth doing even if nothing triggers it now. If a programmer has said that some name is `x`, then it should be `x` in the compiled output. AFAIK, a `GenIdent` means the compiler generated it whereas `Ident` means the developer defined it. I think the decision to make `GenIdent`s work around `Ident`s rather than making later bindings work around earlier bindings better abides with the principle of readable debuggable code.

rhendric:
New commit has a few changes:
* I've implemented three passes everywhere. I think the code simplicity gains ended up being minimal, but positive.
* I overlooked a few large mistakes that actually would have broken unnamed instances rather severely. :facepalm: First, the renamer needs to rename variables that are qualified `GenIdent`s now, because anything at the top level of a module gets qualified. Second, externs files weren't downstream of the renaming phase, so `GenIdent`s there weren't getting renamed. Now there is some extra logic in externs generation that reuses the renaming map generated by the renamer.
* I've added a bunch of tests for various cases that broke while I was experimenting. These are all basically copies of existing tests, but with some instance names removed.

tests/purs/failing/OverlapAcrossModulesUnnamedInstance.out
```diff
@@ -0,0 +1,24 @@
...
+
+  Overlapping type class instances found for
+  [33m                                [0m
+  [33m  OverlapAcrossModules.Class.C X[0m
+  [33m                               Y[0m
+  [33m                                [0m
+  The following instances were found:
+
+    OverlapAcrossModules.X.cX
+    OverlapAcrossModules.$cXY0
```
rhendric, line 13: These error messages are not great. Neither of these instances has a programmer-provided name, and while at least the `$` in the second instance suggests this, the first instance has nothing to indicate that the programmer shouldn't go literally searching for an instance named `cX`. But even in the second case, `$cXY0` is not even the final name that will appear in the generated code!

I think the ideal thing here would be to describe these instances with something like this:
```suggestion
    instance in module OverlapAcrossModules.X with type forall y. C X y
    instance in module OverlapAcrossModules with type C X Y
```

I'd like to address this in a later PR.

src/Language/PureScript/Renamer.hs
```diff
@@ -94,54 +94,68 @@ lookupIdent name = do
...
+  >=> traverse renameValuesInDecl
 
-  renameValuesInDecl (NonRec a name val) = NonRec a name <$> renameInValue val
-  renameValuesInDecl (Rec ds) = Rec <$> traverse updateValues ds
+  where
 
-  updateName :: Ident -> Rename Ident
-  updateName name = (if isTopLevel && isPlainIdent name then pure else updateScope) name
+  renameDecl :: Bool -> Bind Ann -> Rename (Bind Ann)
+  renameDecl isSecondPass = \case
```
JordanMartinez, line 142: Could we replace the `isSecondPass :: Boolean` arg with a simple data type so as to prevent "Boolean Blindness"? Something simple like
```haskell
data PassCount
  = FirstPass
  | SecondPass
```

rhendric: Do you really think it's worth it? `renameDecl` is a local function, only used within ten lines of its declaration; I don't think there's much of a chance for misinterpreting what that parameter is for. And it would bloat up the `if isSecondPass == isPlainIdent name` check a little.

JordanMartinez: Mm... Probably not then. I guess mine was an overreaction.

JordanMartinez (approved):
AFAICT, this PR looks good. Could another core member look this over and provide their feedback?

tests/purs/failing/OverlapAcrossModulesUnnamedInstance.out
JordanMartinez: Sounds good!

tests/purs/failing/OverlappingUnnamedInstances.purs
```diff
@@ -0,0 +1,17 @@
...
+
+instance Test a where
+  test x = x
+
+instance Test Int where
+  test _ = 0
+
+-- The OverlappingInstances instances error only arises when there are two
+-- choices for a dictionary, not when the instances are defined. So without
+-- `value` this module would not raise a warning.
```
mhmdanas, line 15: From https://github.com/purescript/purescript/pull/4099#discussion_r644345398:
```suggestion
-- `value` this module would not raise a error.
```

rhendric: Hmm, nice catch. Guess this would apply to the test of which this is a copy, too.

JordanMartinez:
I've fixed the merge conflict. Can we get another review here?

tests/purs/failing/OverlappingUnnamedInstances.out
```diff
@@ -0,0 +1,22 @@
...
+in module [33mMain[0m
+at tests/purs/failing/OverlappingUnnamedInstances.purs:10:1 - 11:13 (line 10, column 1 - line 11, column 13)
+
+  Overlapping type class instances found for
+  [33m               [0m
+  [33m  Main.Test Int[0m
+  [33m               [0m
+  The following instances were found:
+
+    Main.$test1
```
hdgarrood, line 11: This probably shouldn't happen in this PR, but now that instance names are optional we should probably include source spans in this list so that people can find the problematic instances a little more easily.

hdgarrood (approved):
Looks good!

lib/purescript-cst/src/Language/PureScript/CST/Convert.hs
```diff
@@ -540,13 +541,16 @@ convertDeclaration fileName decl = case decl of
...
 
       className :: Text.Text
-      className = N.runProperName $ qualName cls
+      className = foldMap (uncurry Text.cons . first toLower)
+                . Text.uncons
```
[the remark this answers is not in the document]
rhendric: Oh, I'm completely on board with that.

Maybe we should start a wiki page with our current/evolving notions of good style? I frequently find myself waffling between a style that is more idiomatic/traditional Haskell versus one which has practical benefits like this, and not having a dominant style in the code itself doesn't help. (Another example: equational declaration style versus lambda case—I think the latter is usually superior where applicable, again because renaming the function touches fewer lines.)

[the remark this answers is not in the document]
hdgarrood: I think that would be good 👍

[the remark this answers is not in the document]
JordanMartinez: Rather than a wiki page, should it be included in the main repo?

Another thing we could add to that is to use record types whenever something has 3+ args.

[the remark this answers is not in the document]
rhendric: I'm thinking there isn't much advantage to versioning the standard along with the code—particularly while we're hammering out what the standard is, that could generate a lot of noise in the commit history. Maybe if the wiki page stabilizes, it makes sense to add it then, and then gatekeep it with the same PR workflow we use for code. But for starting out, I think we want something that we can feel more fluid about rewriting.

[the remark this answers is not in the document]
JordanMartinez: Using a wiki page initially and then merging it into the repo when it stabilizes sounds like a good tradeoff.

hdgarrood:
After a bug report on FP Slack in which someone ran into a case where an instance name had a $13 suffix in the module where it was defined but some code was trying to access the same instance with a $12 suffix, I am wondering if the renamer was deliberately leaving top level declarations out before, and whether we might want to backtrack here because this limits our ability to perform cutoff in incremental rebuilds (since the question of “has the public interface of the module changed”) is much more likely to become “in a way no, but we have to treat it as if it has, because of renaming.”

hdgarrood:
https://github.com/purescript/purescript/issues/3724

rhendric:
If I understand the issue correctly, isn't it substantially worse without `GenIdent` than with?

Prior to this PR, generated instance names always had a numeric suffix, and it depended on the state of the supply monad, which means any change that pulls one more or fewer number from the supply at any earlier point during compilation would affect the naming.

Post this PR, generated instance names only change if there is a top-level name conflict, which is rare to begin with, and much less likely to change mid-development. (It can happen, but any name generation strategy runs some risk of deoptimizing incremental builds.)

hdgarrood:
Right, ok, that makes sense, and I agree that this is a substantial improvement. I wasn’t objecting to this PR specifically, rather the approach of using statefulness in top level identifiers at all - I’m wondering whether instance names ought to be more deterministic than they are even after this PR, i.e. whether we should work out an algorithm which guarantees uniqueness without having to depend on statefulness / what else is inside the module. Of course readability will suffer that way, but I think reliability of incremental builds ought to win over readability when they are in conflict. But maybe this is enough of an edge case to not really be necessary; perhaps we should just get this change shipped instead.

rhendric:
If we gave anonymous instances unreadable names—I assume you're thinking something like a robust hash of the fully qualified class name and argument types—perhaps we could restore some readability by also generating friendly-named local vars inside the module and at any import sites? Then downstream modules, if not recompiled, would still function correctly, even though they'd potentially have the ‘wrong’ internal friendly name sometimes.

I'm not crazy about that approach; seems like a fair whack of complexity to take on for an edge case in a compile-time performance feature. But we could pivot into it in the future if this ends up presenting a practical issue for cutoff?

rhendric:
(Oh, and isn't it just a bug that incremental builds weren't triggered for downstream modules when `$12` became `$13`? Is that tracked somewhere?)

rhendric:
*proceeds to seriously consider implementing the idea he just described as not being crazy about*

JordanMartinez:
> Of course readability will suffer that way, but I think reliability of incremental builds ought to win over readability when they are in conflict.

But if one really wanted readable instance names, they could always provide it themselves, right?

hdgarrood:
It might be a bug in incremental compilation, but I think it’s more likely that it’s an instance of #3323 - “i recompiled a downstream module on its own using the ide server and didn’t think the upstream module would need recompiling because the module interface ought not to have changed,” which is arguably not a bug, but somewhat problematic if people can’t reliably guess when a module interface has changed.

f-f:
Why not generate names that contain both a readable name and a hash? In this way they won't become _too_ unreadable.

PR #4097, opened 2021-05-29
Implement Typeable

### LINKED ISSUE #2243: Automatically generate instances for Data.Typeable
I've been doing some work updating [Data.Typeable](https://github.com/joneshf/purescript-typeable).

One logical next step would be to have the compiler automatically derive instances for `Data.Typeable`. It is (at least conceptually) trivial, since it is basically an encoding of the type, which the compiler knows. In fact, it is more reliable to have the complier derive the instance than anything else, since the compiler has the most reliable knowledge of the type. (That is, human implementation could get out of sync with the actual type).

In fact, GHC now does even more than support `deriving Typeable`. It actually [automatically generates](http://home.smart-cactus.org/~ben/ghc/release-prep/8.0.1-rc1/docs/html/users_guide/glasgow_exts.html#deriving-typeable-instances) an instance when needed, even without any explicit invocation of `deriving`. This would be especially desirable in Purescript, given the prohibition of orphan instances. Otherwise, every single module would need to add a `deriving Typeable` (or one would have to use the usual workarounds, such as a `newtype` etc.).

So, my question is this: if I were to implement automatic instance generation for `Data.Typeable`, is that something that would be considered desirable (and possibly mergeable)? Clearly, one wouldn't want to support this kind of thing in the compiler for very many purposes, but I think that `Data.Typeable` is arguably pretty fundamental.

One additional consideration is that `Data.Typeable` will probably change considerably when "kind polymorphism" (#1763) is available, in a manner similar to the [equivalent changes in Haskell](https://ghc.haskell.org/trac/ghc/wiki/GhcKinds/PolyTypeable). I suppose that one could delay implementing automatic instance generation until kind polymorphism is available. However, I'm not sure whether there is any compelling reason to wait. That is, it may turn out that there is a technical reason why automatic instance generation depends on kind polymorphism, but it isn't obviously so (at least, it's not obvious to me).

TheMatten:
**Description of the change**

Purescript currently doesn't have `Typeable` constraint in style of Haskell, which makes it basically impossible to implement some functionality, e.g. safe dynamically typed values or memoizing in VDOM where memoized nodes change type of argument. `Generic` doesn't help there, because it doesn't see difference between constructors defined in different modules, and external libraries don't help either, because they would require users to implement instances for possibly huge amount of types and would prohibit them from using it with external types (because of orphans being dissallowed).

This PR implements minimal version of `Typeable` dictionary generation, expecting API to be defined as follows, e.g. in `prelude`:
```haskell
module Type.Rep (TypeRep, class Typeable, typeRep, sameTypeRep) where

import Data.Eq ((==))
import Data.Monoid ((<>))
import Data.Show (class Show, show)
import Prim.TypeError (class Fail, Text, Quote, Above, Beside)

newtype TypeRep a = TypeRep String

type role TypeRep nominal

instance showTypeRep :: Show (TypeRep a) where
  show (TypeRep n) = "(TypeRep " <> n <> ")"

class Typeable :: forall k. k -> Constraint
class Typeable a where
  typeRep :: TypeRep a

instance typeableFallback :: Fail (InvalidTypeable a) => Typeable a where
  typeRep = typeRep

type InvalidTypeable a = Above
  (Text "Cannot create `Typeable` instance")
  (Text "Make sure no type or kind variables appear in the type")

sameTypeRep :: forall a b. TypeRep a -> TypeRep b -> Boolean
sameTypeRep (TypeRep a) (TypeRep b) = a == b
```
where `Typeable` gets resolved automatically in style of `IsSymbol`. `TypeRep` contains statically generated hash created from normalized string representation of the type (using `GHC.Fingerprint`).

I figured that the change is small enough that instead of just talking about it, I can create prototype that can be either dismissed or refined, so I've created PR instead of an issue. Technically, it fixes #2243, though using much more minimal interface as described above, only providing what's needed to implement functionality like `Dynamic` or
```haskell
cast :: forall a b. Typeable a => Typeable b => a -> Maybe b
```
in some external library.

As far as implementation goes, things I'm not completely sure about are whether it may interact badly with redundant syntax (seems like kind annotations, parens, operators, synonyms and rows passed to `rowToSortedList` get normalized at that point?) and whether `PolyKinds` may create some unexpected problems (they're hashed too by extracting them from `KindApp` right now).

When it comes to compatibility guarantees, I would expect same hashes when building using the same compiler version modulo bugfixes - this means that changes to `GHC.Fingeprint` should probably result in minor version change.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Constants/Prelude.hs
+++ b/src/Language/PureScript/Constants/Prelude.hs
@@ -376,6 +376,17 @@ pattern DataSymbol = ModuleName "Data.Symbol"
 pattern IsSymbol :: Qualified (ProperName 'ClassName)
 pattern IsSymbol = Qualified (Just DataSymbol) (ProperName "IsSymbol")
 
+-- Type.Rep
+
+pattern TypeRep :: ModuleName
+pattern TypeRep = ModuleName "Type.Rep"
+
+pattern Typeable :: Qualified (ProperName 'ClassName)
+pattern Typeable = Qualified (Just TypeRep) (ProperName "Typeable")
+
+typeRep :: forall a. (IsString a) => a
+typeRep = "typeRep"
+
 prelude :: forall a. (IsString a) => a
 prelude = "Prelude"
 
--- a/src/Language/PureScript/TypeChecker/Entailment.hs
+++ b/src/Language/PureScript/TypeChecker/Entailment.hs
@@ -21,14 +21,16 @@ import Control.Monad.Writer
 import Data.Foldable (for_, fold, toList)
 import Data.Function (on)
 import Data.Functor (($>))
-import Data.List (findIndices, minimumBy, groupBy, nubBy, sortOn)
+import Data.List (findIndices, foldl', minimumBy, groupBy, nubBy, sortOn)
 import Data.Maybe (fromMaybe, listToMaybe, mapMaybe)
 import qualified Data.Map as M
 import qualified Data.Set as S
+import Data.String (fromString)
 import Data.Traversable (for)
 import Data.Text (Text, stripPrefix, stripSuffix)
 import qualified Data.Text as T
 import qualified Data.List.NonEmpty as NEL
+import GHC.Fingerprint (Fingerprint, fingerprintString)
 
 import Language.PureScript.AST
 import Language.PureScript.Crash
@@ -42,7 +44,7 @@ import Language.PureScript.TypeChecker.Unify
 import Language.PureScript.TypeClassDictionaries
 import Language.PureScript.Types
 import Language.PureScript.Label (Label(..))
-import Language.PureScript.PSString (PSString, mkString, decodeString)
+import Language.PureScript.PSString (PSString, mkString, decodeString, decodeStringEither)
 import qualified Language.PureScript.Constants.Prelude as C
 import qualified Language.PureScript.Constants.Prim as C
 
@@ -54,6 +56,7 @@ data Evidence
   -- | Computed instances
   | WarnInstance SourceType -- ^ Warn type class with a user-defined warning message
   | IsSymbolInstance PSString -- ^ The IsSymbol type class for a given Symbol literal
+  | TypeableInstance Fingerprint -- ^ The Typeable type class for a given type
   | EmptyClassInstance        -- ^ For any solved type class with no members
   deriving (Show, Eq)
 
@@ -176,6 +179,7 @@ entails SolverOptions{..} constraint context hints =
       -- This allows us to defer a warning by propagating the constraint.
       findDicts ctx cn Nothing ++ [TypeClassDictionaryInScope Nothing 0 (WarnInstance msg) [] C.Warn [] [] [msg] Nothing]
     forClassName _ _ C.IsSymbol _ args | Just dicts <- solveIsSymbol args = dicts
+    forClassName _ _ C.Typeable kinds args | Just dicts <- solveTypeable kinds args = dicts
     forClassName _ _ C.SymbolCompare _ args | Just dicts <- solveSymbolCompare args = dicts
     forClassName _ _ C.SymbolAppend _ args | Just dicts <- solveSymbolAppend args = dicts
     forClassName _ _ C.SymbolCons _ args | Just dicts <- solveSymbolCons args = dicts
@@ -379,6 +383,9 @@ entails SolverOptions{..} constraint context hints =
             mkDictionary (IsSymbolInstance sym) _ =
               let fields = [ ("reflectSymbol", Abs (VarBinder nullSourceSpan UnusedIdent) (Literal nullSourceSpan (StringLiteral sym))) ] in
               return $ TypeClassDictionaryConstructorApp C.IsSymbol (Literal nullSourceSpan (ObjectLiteral fields))
+            mkDictionary (TypeableInstance fp) _ =
+              let fields = [ (C.typeRep, Literal nullSourceSpan $ StringLiteral $ fromString $ show fp) ] in
+              return $ TypeClassDictionaryConstructorApp C.Typeable (Literal nullSourceSpan (ObjectLiteral fields))
 
             unknownsInAllCoveringSets :: [SourceType] -> S.Set (S.Set Int) -> Bool
             unknownsInAllCoveringSets tyArgs = all (\s -> any (`S.member` s) unkIndices)
@@ -415,6 +422,38 @@ entails SolverOptions{..} constraint context hints =
     solveIsSymbol :: [SourceType] -> Maybe [TypeClassDict]
     solveIsSymbol [TypeLevelString ann sym] = Just [TypeClassDictionaryInScope Nothing 0 (IsSymbolInstance sym) [] C.IsSymbol [] [] [TypeLevelString ann sym] Nothing]
     solveIsSymbol _ = Nothing
+
+    solveTypeable :: [SourceType] -> [SourceType] -> Maybe [TypeClassDict]
+    solveTypeable [kind] [type_]
+      | Just tS <- normalizedTypeS type_ = Just
+        [TypeClassDictionaryInScope Nothing 0 (TypeableInstance $ fingerprintString $ tS "") [] C.Typeable [] [kind] [type_] Nothing]
+     where
+      normalizedTypeS t = case rowToSortedList t of
+        ([], t') -> case t' of
+          TypeLevelString _ s -> Just $ shows s
+          TypeConstructor _ (Qualified mn (ProperName cn)) -> Just $
+            maybe id ((. showString ".") . showString . T.unpack . runModuleName) mn . showString (T.unpack cn)
+          TypeApp _ t1 t2 -> do
+            t1S <- normalizedTypeS t1
+            t2S <- normalizedTypeS t2
+            Just $ showString "(" . t1S . showString ")(" . t2S . showString ")"
+          KindApp _ t1 t2 -> do
+            t1S <- normalizedTypeS t1
+            t2S <- normalizedTypeS t2
+            Just $ showString "(" . t1S . showString ")@(" . t2S . showString ")"
+          REmpty _ -> Just $ showString "()"
+          _ -> Nothing
+        (row, isREmpty -> True) -> do
+          rowS <- normalizedRowS row
+          Just $ showString "(" . rowS . showString ")"
+        _ -> Nothing
+      normalizedRowS = foldl' go (Just id) where
+        go s (RowListItem _ (Label n) t) = do
+          s'  <- s
+          lbl <- traverse (either (\_ -> Nothing) Just) $ decodeStringEither n
+          tS  <- normalizedTypeS t
+          Just $ s' . showString lbl . showString "::" . tS . showString ","
+    solveTypeable _ _ = Nothing
 
     solveSymbolCompare :: [SourceType] -> Maybe [TypeClassDict]
     solveSymbolCompare [arg0@(TypeLevelString _ lhs), arg1@(TypeLevelString _ rhs), _] =


rhendric:
I'm reading the conversation in #2243 and, like several of the participants there, I've never had the need for `Typeable` or `Dynamic`. The thing this most reminds me of, though, isn't `Generic` or `Foreign` as discussed over there, but `Variant`. Speaking as someone who's never used them, the `Typeable`/`Dynamic` approach seems to be about constructing something like a universal disjoint sum over every possible type, whereas `Variant` allows for open but limited disjoint sums. I think this means that you get less boilerplate when working with a `Dynamic` than a `Variant`, but you also completely decouple your producers and your consumers, such that it's easy to change one side only and get run-time errors that a `Variant`-based approach would catch.

Is that basically right? If so, my current weakly-held opinion is that I'd rather spend compiler code on making working with rows and `Variant`s less painful, if possible, than on promoting the use of `Typeable`. If not, I'd love to hear about a compelling use case for `Typeable` that `Variant` doesn't adequately handle.

garyb:
The `Variant` comparison seems apt yeah, it's pretty much what Nate was talking about here: https://github.com/purescript/purescript/issues/2243#issuecomment-275402922

I think the big difference though is that variants are tagged with their constructor in the encoding, but `Typeable` forgoes that even?

rhendric:
The `TypeRep` in a `Dynamic` seems like it does the same job as the tag in a `Variant`.

TheMatten:
@garyb yeah, problems with subcomponents are precisely what I'm working with myself - for context, I'm currently porting Shpadoinkle (https://shpadoinkle.org) into purescript (https://gitlab.com/thematten/purescript-fluorine).

Now, thing is, premise of it's approach is to make composition simple, and while that's easily doable with simple trees of elements, problems arise once you try to add memoization, because while patching, you have no way to know whether new memoized node isn't actually a completely different one, with different type of argument. Halogen solves this with typed approach as mentioned, but such solution degrades user experience a lot, introducing need for additional boilerplate, possibility of new, unproductive type errors and leaks internal details about components being memoized. All of that only to tell compiler which memoized types may appear in some position. And with runtime representation being basically the same, as @rhendric mentioned. This may bite programmers in pretty arbitrary situations, e.g. during refactoring of components.

In Haskell, `Typeable` gets misused from time to time by users which are accustomed to approaches common to dynamically typed languages. But there still are some genuine use-cases for it, mostly when some interface allows for embedding of relatively self-contained parts, like e.g. when creating APIs for caching, serializing or debugging. Plus it's safe - in Purescript, people may be inclined to implement such functionality using hacks in raw JS, but because it isn't possible to distinguish `newtype`'s at runtime, they can easily introduce bugs by breaking safety guarantees this way. Personally I would rather see code that overuses `Typeable`, because there's no way to use it incorrectly.

timjs:
I'd love to have `Typeable` in PureScript. I use it a lot in combination with `Dynamic` to parse user input, wrap it, pass it around, and match it against the type of an (Html) input field when needed. Wanted to do this in PureScript, but switched to Haskell in the end. Something along these lines:

```haskell
parse :: Text -> Either Text Dynamic
parse val
  | Just v <- scan val :: Maybe Unit = okay <| Dynamic v
  | Just v <- scan val :: Maybe Bool = okay <| Dynamic v
  | ...
  | otherwise = error <| unwords ["!! Error parsing value", display val |> quote]

insert :: Component a -> Dynamic -> Component a
insert (Field b) (Dynamic b') =
  | Just Refl <- b ~= b' -> done <| Field b'
  | otherwise -> throw <| CouldNotChangeVal (someTypeOf b) (someTypeOf b')
...
```

I'm not aware of a trick using open variants to achieve the same thing...

TheMatten:
@timjs Yeah, once again you could probably create some top level synonym used in place of `Dynamic` for variant with all possible types that may appear during parsing, but that would be burdersome and would provide no additional safety.

Really, `Variant` is in a sense "type-directed generation of local, restricted `Typeable` constraints" - so it basically shifts work from the compiler to users, asking them to create "lookup table" of tokens corresponding to specific types at type level. Question is, whether it's worth it to take that work from them - I would say yes, in same way as I would say yes to e.g. `IsSymbol`.

timjs:
@TheMatten, that is indeed a possibility

TheMatten:
@timjs I imagine you should be able to parameterize context you work in with some variant supplied by the user, similarly to how slots in `halogen` work. We're technically working with "open" world from programmer's perspective, but the other way of viewing this (which may be closer to what @rhendric is saying) is that in reality you only ever work with finite set of types in your application, and thus you can create some `Variant` that covers all of them.

What I'm trying to say is that while it's hard to come up with usecase which wouldn't be possible to encode with "big enough" `Variant`, I think it's putting extra work and mental overhead on programmer and makes APIs more complicated, just to shave off what seems to amount to roughly 60 lines of code in the compiler.

hdgarrood:
It does not follow that the maintenance cost is low just because this PR has only 60 lines. The potential maintenance cost of supporting an extra feature which ends up being depended on by users is very difficult to predict, but could be significantly more than you might expect. If this implementation turns out to be incorrect, we'll need to diagnose and fix the bug - and sometimes this significantly complicates the implementation. The initial Coercible PR seemed fine but it needed something like 10 followup PRs to get the feature to a stage where it worked correctly, and by the end I think we had quite a bit more code than we did originally. The feature needs documentation, it generates more questions along the lines of "when should I use feature X rather than feature Y", it could interact poorly with other features, etc.

TheMatten:
@hdgarrood Yeah, I should probably be more concrete about maintenance costs:
- As far as bugfixes of incorrect behaviour go, current design should be minimalistic and opaque enough to limit possibility of backwards-incompatible changes (unless we later find out that e.g. some types can't be safely covered) - basically, it provides no more than optimized equality of normalized types represented as strings, across single compiler version (maybe comparable to equality of `FastString`s in GHC).
- As far as interaction with other features goes, I think we're enjoying "confluence of design decisions" in this case - compared to GHC, `purs` comes with much simpler typechecker, doesn't provide features like wired-in type equalities and it's `Coercible` doesn't construct any interesting runtime proofs. So it should be actually very easy to use them together - something like
  ```haskell
  eqTypeRep :: forall a b r. TypeRep a -> TypeRep b -> (Coercible a b => r) -> Maybe r
  ```
  can be implemented at library level using `sameTypeRep` (and `unsafeCoerce`) without any support from the compiler.
- When it comes to related documentation and guidance around it's use, I would be happy to put it together in case `Typeable` gets implemented. Well written comparison with `Variant` should be enough to make it clear where each of those makes sense in my opinion.

f-f:
I am familiar with Variant, but less familiar with Typeable - since here we're talking about implementing a compiler extension, I'm wondering:
- if implementing `Typeable` means that `Variant` can be implemented on top of it (or viceversa?)
- would the performance improve? (both of the compiled code, and the compiler)
- more generally, since it looks like there are two similar usecases here it's worth thinking if it's possible to cover both with the same feature

TheMatten:
@f-f To answer your points:
- One can implement unnamed `Variant` on top of `Typeable`, because it would basically be type-restricted version of `Dynamic` (that approach is actually common in Haskell, because it's the fastest safe solution). With named ones, you could use `TypeRep`s as additional keys to access different types with same label. But because `Typeable` is "open" while `Variant` is "closed", if you were to implement function that extracts whatever field that is already used, you need to cover case of "it's neither of listed options" to be exhaustive from compiler's point of view. On the other hand, you can't really implement "open" variant using `Variant` - you can only approximate it by creating global list of types that may appear in variant in your application, as mentioned in the discussion above.
- Depends on how it's used, but it would be always better or roughly-equal. E.g. if you were to avoid `unsafeCoerce` in your implementation of variant, you would probably end up with some inductive GADT approximation, which is necessarily going to be slower. Plus, because we don't have unnamed variants without `Typeable`, one has to provide name of field as an argument to every field-specific operation (`Proxy`). With (unsafely implemented) named variant though, representation itself would be basically the same, except for length of keys.
- I would actually say that given the description above, their usecases are actually sligthly different:
  - `Variant` works better with APIs oriented around extraction, because of explicit labels. It may be better fit for cases where concrete identity of selection matters instead of it's type, like e.g. components in current `halogen`
  - `Typeable` works better with APIs oriented around injection, where difference between selections matters mostly from implementation perspective, or where type itself is enough to distinguish purpose of the value. This is the case with `Shpadoinkle`, which is merely interested in memoization of anonymous nodes.

hdgarrood:
I'm :-1: because I don't think this provides enough utility/is in enough demand to justify adding a feature to the compiler, sorry.

TheMatten:
@hdgarrood no problem :slightly_smiling_face: Alternatively, would it be possible to expose needed information e.g. in #4092?

hdgarrood:
So adding type information to CoreFn? That is something that comes up from time to time, but I think it's a bigger project than it seems. I don't really know enough about it to be able to comment unfortunately.

timjs:
Sorry, I've lost the discussion on `Typeable` vs `Variant` a bit.

To make it concrete. Say I've some container type, which is empty or filled with a value of type `a`. Given a `Dynamic`, I'd like to update the value inside the container, but obviously only when the types match.

```haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TypeOperators #-}

module Typeable where

import Data.Type.Equality
import Type.Reflection

(?=) :: TypeRep a -> TypeRep b -> Maybe (a :~: b)
(?=) = testEquality

-- | A container of type `a`
data Container a
  = Empty
  | Filled a

-- | Essentialy `Dynamic`
data Input where
  Input :: Typeable a => a -> Input

update :: forall a. Typeable a => Container a -> Input -> Container a
update (Empty) (Input y)
  | Just Refl <- alpha ?= typeOf y = Filled y
  | otherwise = Empty
  where
    alpha = typeRep :: TypeRep a
update (Filled x) (Input y)
  | Just Refl <- typeOf x ?= typeOf y = Filled y
  | otherwise = Filled x
```

Above is working Haskell code. I'd do the same thing in Idris by using a sigma type saving a type witness. I can't see how I would use a `Variant` of some row to model the same thing. Hope somebody can help me understand this.

rhendric:
@timjs

```purs
import Prelude

import Data.Maybe (Maybe(..))
import Data.Variant (Variant, prj)
import Data.Symbol (class IsSymbol)
import Type.Proxy (Proxy(..))
import Prim.Row as Row
import Effect (Effect)

-- | A container of type `a`
data Container a
  = Empty
  | Filled a

update :: forall s a r r'. IsSymbol s => Row.Cons s a r r' => Container a -> Variant r' -> Container a
update Empty v = case prj (Proxy :: _ s) v of
  Just y -> Filled y
  Nothing -> Empty
update (Filled x) v = case prj (Proxy :: _ s) v of
  Just y -> Filled y
  Nothing -> Filled x
```

timjs:
Thanks a lot @rhendric! Really appreciate it :-) Now I understand the way to use this.

Only thing left is an error on [unknown type variables in the instance head of IsSymbol in `update`](https://try.purescript.org/?session=8b3d0075-71dd-0ad6-d412-7638e18f952c) when using this solution. Any ideas?

timjs:
Really didn't know `Variant` was so powerful :-O

In that case I agree with @hdgarrood that `Typeable` is not a necessary extension we'd need to implement and maintain. Maybe focussing on the ergonomics of `Variant` would be a better idea!

TheMatten:
> Only thing left is an error on unknown type variables in the instance head of IsSymbol in update when using this solution. Any ideas?

You will either need to fix `s` into something specifc, or pass in `Proxy s` that provides it as an argument.

rhendric:
> You will either need to fix `s` into something specifc, or pass in `Proxy s` that provides it as an argument.

Yes, exactly. It can be worked around with some type-level fun, as long as everything around it is sufficiently well-typed.

Here's a more fully-worked example incorporating your previous `parse` example as well:

```purs
module Main where

import Prelude

import Data.Either (Either(..))
import Data.Maybe (Maybe(..))
import Data.Symbol (class IsSymbol)
import Data.Variant (Variant, inj, prj)
import Effect (Effect)
import Prim.Row as Row
import Prim.RowList (RowList, class RowToList)
import Prim.RowList as RowList
import TryPureScript (p, render, text)
import Type.Data.Boolean (False, True)
import Type.Proxy (Proxy(..))


-- Let's pretend we aren't on try.purescript.org
putStrLn :: String -> Effect Unit
putStrLn = text >>> p >>> render


-- This is something that should be in a library somewhere (maybe it is?)
class FindTypeInRowList :: forall k. Symbol -> k -> RowList k -> Boolean -> Constraint
class FindTypeInRowList s a rl b | a rl -> s b

instance ftirlNil :: FindTypeInRowList s a RowList.Nil False
else instance ftirlNext :: FindTypeInRowList s a (RowList.Cons s a rl) True
else instance ftirlRec :: FindTypeInRowList s a rl' b => FindTypeInRowList s a (RowList.Cons s' a' rl') b


data Container a
  = Empty
  | Filled a
  
instance showContainer :: Show a => Show (Container a) where
  show Empty = "Empty"
  show (Filled a) = "(Filled " <> show a <> ")"

-- This needs a proxy to determine s.
update' :: forall p s a r r'. IsSymbol s => Row.Cons s a r r' => p s -> Container a -> Variant r' -> Container a
update' _ Empty v = case prj (Proxy :: _ s) v of
  Just y -> Filled y
  Nothing -> Empty
update' _ (Filled x) v = case prj (Proxy :: _ s) v of
  Just y -> Filled y
  Nothing -> Filled x

-- This uses type-level shenanigans to determine s, so doesn't require a
-- proxy as long as the types are sufficiently determined at the call site.
update :: forall s a r r' rl. RowToList r rl => IsSymbol s => Row.Cons s a r' r => FindTypeInRowList s a rl True => Container a -> Variant r -> Container a
update = update' (Proxy :: _ s)


class Scan a where
  scan :: String -> Maybe a

instance scanUnit :: Scan Unit where
  scan = case _ of
    "unit" -> Just unit
    _ -> Nothing

instance scanBool :: Scan Boolean where
  scan = case _ of
    "true" -> Just true
    "false" -> Just false
    _ -> Nothing

-- purs warns about this _. You could spell out this type entirely as
-- parse :: forall r. String -> Either String (Variant (u :: Unit, b :: Boolean | r))
-- but it'd be nice if you didn't have to.
parse :: String -> Either String (Variant _)
parse val
  | Just v <- (scan val :: Maybe Unit) = Right $ inj (Proxy :: _ "u") v
  | Just v <- (scan val :: Maybe Boolean) = Right $ inj (Proxy :: _ "b") v
  | otherwise = Left $ "error scanning " <> val

main :: Effect Unit
main = do
  let v :: Either String (Variant (u :: Unit, b :: Boolean))
      v = parse "true"
  let c1 :: Container Unit
      c1 = Empty
  let c2 :: Container Boolean
      c2 = Filled false
  let result1 = update c1 <$> v
  let result2 = update c2 <$> v
  putStrLn (show result1)
  putStrLn (show result2)
```

There are some obvious warts here. The biggest is that we need to provide labels `u` and `b` in our row type. I could imagine a version of `Variant` which uses `HList`s instead of rows, and then the tag could be provided by the equivalent of `FindTypeInRowList`—this could be written with today's compiler, to my knowledge. Other issues mostly involve type inference and making it easier to elide the full row types without generating warnings. I think improving all of these could be better options than implementing `Typeable`.

One thing I still don't understand is what `Typeable` can do for memoization. @TheMatten, can you point me to a small example of what you'd want that for? I'm not familiar with Shpadoinkle, unfortunately.

TheMatten:
@rhendric sure - I'm trying to implement `depending` combinator: https://gitlab.com/platonic/shpadoinkle/-/merge_requests/182/diffs#927128ccef388486a73f825b27dde5f95305bc72_325_348
Basically, what it does is that it stores `Typeable` argument alongside the rendering function of a memoized node - this way, you can try to cast old argument into type of new one to compare them for equality, or rerender if either of those actions fails. I don't really see a way to implement this with `Variant` without affecting `Html` type, but I don't think this should appear in it in the first place - it's an optimization detail, not a safety/corectness concern.

(BTW, this ignores the fact that the rendering function itself may change. But that's problem for `StaticPointers`, which should be implementable with `purs codegen`.)

rhendric:
I might not understand that example fully because I don't have the larger context of what else is going on in that MR, but let me try anyway.

If you get/require an implementation of `StaticPointer`, is that enough?

```purs
data DependencyF a = DependencyF (StaticPointer (a -> a -> Boolean)) a
newtype Dependency = Dependency (Exists DependencyF)
instance eqDependency :: Eq Dependency where
  eq (Dependency d1) (Dependency d2) =
    d1 # runExists \(DependencyF staticEqA a) ->
      d2 # runExists \(DependencyF staticEqB b) ->
        staticKey staticEqA == staticKey staticEqB && any (\eq -> eq a (unsafeCoerce b)) (derefStaticPtr staticEqA)

depending :: forall a m c. StaticPointer (a -> a -> Boolean) -> (a -> Html m c) -> a -> Html m c
depending staticEq f x = Html (\a d b c -> d (Dependency (mkExists (DependencyF staticEq x)))
                       $ case f x of Html h' -> h' a d b c)
```

TheMatten:
Hmm, very cool idea, but that wouldn't work with something like `\_ _ -> true` or `unsafeRefEq`, where the function could be used with multiple nodes, would it?

rhendric:
Yeah, you'd want the production of the `StaticPointer` to be safe.

But if that general idea is sufficient on the consumer side of things, once you have `purs codegen` you could make the user side much more foolproof. Just:

```purs
class StaticEq a where
  staticEq :: StaticPointer (a -> a -> Boolean)

instance universalStaticEqInstance :: Eq a => StaticEq a where
  staticEq = unsafeThrow "not to be used without some fancy CoreFn transformations"


depending :: forall a m c. StaticEq a => (a -> Html m c) -> a -> Html m c
depending f x = Html (\a d b c -> d (Dependency (mkExists (DependencyF staticEq x)))
                $ case f x of Html h' -> h' a d b c)
```

Then rewrite any calls to `universalStaticEqInstance(...)` in CoreFn to `StaticEq(static(Data_Eq.eq(...)))` prior to running the `StaticPointer` transformation. If the `StaticPointer` transformation succeeds, you got safe pointers to unique equality functions. If the `Eq` instance contains any non-staticable subexpressions, the transformation fails (and the user has to move the `StaticEq` constraint around probably).

PR #4099, opened 2021-06-02
Desugar type class instance names in type class desugaring pass

JordanMartinez:
**Description of the change**

This PR makes the changes I described [in #4085 about inlining the type class instance name desugar pass into the type classes desugar pass](https://github.com/purescript/purescript/pull/4085#issuecomment-850871300). <del>Since I assume #4096 will get merged, which would introduce merge conflicts, I built this PR on top of #4096 to better demonstrate this minor fix. See the third commit (and following if changes are requested).</del>

<del>Once #4096 does get merged</del> #4096 has been merged. This PR will need to be rebased to fix the merge conflict. Until then I won't add an entry for the changelog.

---

**Checklist:**


src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -79,12 +80,22 @@ desugarModule
   -> Module
   -> Desugar m Module
 desugarModule syns kinds (Module ss coms name decls (Just exps)) = do
-  let (classDecls, restDecls) = partition isTypeClassDecl decls
+  decls' <- parU decls desugarInstName
+  let (classDecls, restDecls) = partition isTypeClassDecl decls'
```
JordanMartinez, line 84: The `partition` line currently iterates through all declarations. By adding `parU decls desugarInstName`, declarations will be iterated through a second time. Ideally, we would only iterate through the list once and get the results of both.

rhendric: Why `parU`? `desugarInstName` can't error; isn't this just a `traverse`?

JordanMartinez: Ah, yeah it is.

Also, if we used one fold to produce the results of both, I believe we'd need to reverse the resulting `restDecls` list, which would iterate through the list a second time and might defeat the point of this approach.

rhendric: And then I suppose if you're concerned about extra iterations, could you just apply this to `restDecls` instead of all of `decls`?

tests/purs/failing/OverlappingUnnamedInstances.purs
```diff
@@ -0,0 +1,17 @@
...
+
+instance Test a where
+  test x = x
+
+instance Test Int where
+  test _ = 0
+
+-- The OverlappingInstances instances error only arises when there are two
+-- choices for a dictionary, not when the instances are defined. So without
+-- `value` this module would not raise a warning.
```
mhmdanas, line 15:
```suggestion
-- `value` this module would not raise a error.
```

src/Language/PureScript/Sugar/TypeClasses.hs
JordanMartinez: Yeah, good point.

tests/purs/failing/OverlappingUnnamedInstances.purs
JordanMartinez: @mhmdanas This is from #4096, so maybe report it there?

JordanMartinez:
I took this PR, ran `stack build`, copied the binary into a new folder, installed the latest package set, removed all instance names and the `::` character, and then compiled all libraries.

Everything compiled fine without issue. I then ran `halogen-hooks`' tests (hoping it pulls enough in to be worth it) and the tests built fine and passed

tests/purs/failing/OverlappingUnnamedInstances.purs
mhmdanas: Oh, sorry. I forgot that this was not based on `master` :D.

JordanMartinez:
I've updated this PR to build on top of master now that #4096 has been merged

hdgarrood (approved):

PR #4100, opened 2021-06-04
Display kind signatures and comments in docs

### LINKED ISSUE #3973: Include kind information in HTML docs

### LINKED ISSUE #4036: Docs on types with kind signatures aren't included in html docs
## Description

It seems that documentation on types with kind signatures aren't rendered.

## To Reproduce

1. Create a file with the following code:
```purescript
module Main where

-- | Foo is a type with a kind signature whose documentation doesn't show up.
newtype Foo :: forall kind. kind -> Type -> Type
newtype Foo k a = Foo a
```

2. Run `purs docs "src/**/*.purs" to generate HTML docs locally
3. Open the `Main` module's docs

The docs on `Foo` aren't shown.

## Expected behavior

The docs of `Foo` should appear in the HTML file.

## PureScript version

v0.14.0

JordanMartinez:
**Description of the change**

Fixes #3973

--- 

Remaining work
- [x] Add a single line that separates a kind signature from its type declaration
- [ ] Make inferred kind signatures appear in the docs

---

**Checklist:**


JordanMartinez:
Latest changes now make `data`, `newtype`, and `type` display their kind signatures in the docs and merge their comments together. However, `class` still doesn't display its kind signatures or docs above them.

I got the `newtype` and `type` to work after realizing that [`reorder`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/AST/Exported.hs#L145-L156) wasn't including `KindDeclaration` in its consideration because `declName` excludes them. I added `declDocName` as a temporary solution to ensure `KindDeclaration`s are included.

JordanMartinez:
CI currently fails because the number of args for various `Declaration` types has increased by one. Thus, the `test` code is failing.

src/Language/PureScript/Docs/Render.hs
```diff
@@ -24,6 +24,15 @@ import qualified Language.PureScript.Environment as P
 import qualified Language.PureScript.Names as P
 import qualified Language.PureScript.Types as P
 
+renderKindSig :: Text -> Text -> Type' -> RenderedCode
+renderKindSig title keywordTxt ty =
+  mintersperse sp
+      [ keyword keywordTxt
+      , ident' title
```
JordanMartinez, line 31: This is a temporary solution, but right now this is problematic. Using `ident`' will link to the data constructor that has the same name as the title. In some cases (e.g. `class`), this won't exist and will produce a 404 if clicked.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -184,6 +184,12 @@ data DeclarationInfo
 
 instance NFData DeclarationInfo
 
+extractKindSig :: DeclarationInfo -> Maybe (Text, Type')
+extractKindSig (DataDeclaration dtype ks _) = fmap (P.showDataDeclType dtype,) ks
+extractKindSig (TypeSynonymDeclaration ks _ _) = fmap ("type",) ks
+extractKindSig (TypeClassDeclaration ks _ _ _) = fmap ("class",) ks
+extractKindSig _ = Nothing
```
JordanMartinez, line 191: Another temporary solution. Ideally, we would be using `keywordType` or something like that, so that any changes (if ever made) only need to made in one location.

JordanMartinez:
I think the number of changes I have to do here show that it would be better if we had a desugar pass that moved `KindDeclaration`s into a `DataDeclaration`, `TypeSynonymDeclaration`, and `TypeClassDeclaration`. Most of this PR is just dealing with this one issue.

If we did that, we might also want to merge `RoleDeclaration`s into their corresponding `DataDeclaration`.

JordanMartinez:
CI still fails because there is a breaking change with the `docs.json` format. There's also a few `purs graph` tests that are failing, but I'm not sure if that's related to the `docs.json` issue or something else.

Class declarations still don't show up in the resulting output, but I thought the work I've done was good enough for some initial feedback.

I've now moved the kind information into its own type called `KindInfo`, which just wraps the keyword and the kind signature, so that we can display the content without having to extract which keyword was used in the `info` arg of `Declaration`. It makes the code cleaner and minimizes the changes made to other code.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -560,6 +573,7 @@ asDeclaration =
               <*> key "sourceSpan" (perhaps asSourceSpan)
               <*> key "children" (eachInArray asChildDeclaration)
               <*> key "info" asDeclarationInfo
+              <*> key "kind" (perhaps asKindInfo)
```
JordanMartinez, line 576: AFAICT, CI is currently failing because of this line. Since the key `kind` doesn't exist in older versions of PS's docs, the parser will fail as it [expects the `kind` key to be there](https://github.com/purescript/purescript/runs/2784351550#step:8:19514):
```
 purescript    >   tests/TestPscPublish.hs:52:17: 
  purescript    >   1) publish, json compatibility, json produced by v0.11.3, symbols-3.0.0.json
  purescript    >        JSON parsing failed: Error in $: At the path: ["modules"][0]["declarations"][0]
  purescript    >        The required key "kind" is missing
  purescript    > 
  purescript    >   To rerun use: --match "/publish/json compatibility/json produced by v0.11.3/symbols-3.0.0.json/"
  purescript    > 
  purescript    >   tests/TestPscPublish.hs:52:17: 
  purescript    >   2) publish, json compatibility, json produced by v0.11.3, generics-4.0.0.json
  purescript    >        JSON parsing failed: Error in $: At the path: ["modules"][0]["declarations"][0]
  purescript    >        The required key "kind" is missing
  purescript    > 
  purescript    >   To rerun use: --match "/publish/json compatibility/json produced by v0.11.3/generics-4.0.0.json/"
  purescript    > 
  purescript    >   tests/TestPscPublish.hs:52:17: 
  purescript    >   3) publish, json compatibility, json produced by v0.12.1, typelevel-prelude-3.0.0.json
  purescript    >        JSON parsing failed: Error in $: At the path: ["modules"][0]["declarations"][0]
  purescript    >        The required key "kind" is missing
  purescript    > 
  purescript    >   To rerun use: --match "/publish/json compatibility/json produced by v0.12.1/typelevel-prelude-3.0.0.json/"
  purescript    > 
```

If I replace `key "kind (perhaps asKindInfo)` with `keyOrDefault "kind" Nothing (Just <$> asKindInfo)`, then I get an error about a type mismatch between an object and null: 
```

purescript>   src/Language/PureScript/Docs/Collect.hs:120:19: 
purescript>   1) publish, preparePackage with json roundtrips, purescript-prelude
purescript>        uncaught exception: ErrorCall
purescript>        An internal error occurred during compilation: Failed to decode: output/Data.Boolean/docs.jsonAt the path: ["declarations"][0]["kind"]
purescript>        Type mismatch:
purescript>        Expected a value of type object
purescript>        Got: null
purescript>        Please report this at https://github.com/purescript/purescript/issues
```
How do I write the parser for `asDeclaration` but make the `kind` key optional?

rhendric:
```suggestion
              <*> keyMay "kind" (perhaps asKindInfo)
```

JordanMartinez: Isn't `keyMay "foo" parser` [the same as](https://hackage.haskell.org/package/aeson-better-errors-0.9.1.0/docs/src/Data-Aeson-BetterErrors-Internal.html#keyMay) `keyOrDefault "foo" Nothing (Just <$> parser)`?

rhendric: Oh wait, I see, the problem is that the key could be missing _or_ null _or_ defined. Did you try something like
```suggestion
              <*> keyOrDefault "kind" Nothing (perhaps asKindInfo)
```

JordanMartinez: Thanks! That seems to work.

JordanMartinez:
So, the final part of this PR that needs to be figured out is why the `class` kind signatures aren't being included in the rendered docs.

lib/purescript-cst/src/Language/PureScript/AST/Declarations.hs
```diff
@@ -513,6 +522,26 @@ declName KindDeclaration{} = Nothing
...
+declDocName (DataDeclaration _ _ n _ _) = Just (TyName n)
+declDocName (TypeSynonymDeclaration _ n _ _) = Just (TyName n)
+declDocName (ValueDeclaration vd) = Just (IdentName (valdeclIdent vd))
+declDocName (ExternDeclaration _ n _) = Just (IdentName n)
+declDocName (ExternDataDeclaration _ n _) = Just (TyName n)
+declDocName (FixityDeclaration _ (Left (ValueFixity _ _ n))) = Just (ValOpName n)
+declDocName (FixityDeclaration _ (Right (TypeFixity _ _ n))) = Just (TyOpName n)
+declDocName (TypeClassDeclaration _ n _ _ _ _) = Just (TyClassName n)
+declDocName (TypeInstanceDeclaration _ _ _ n _ _ _ _) = IdentName <$> hush n
+declDocName (KindDeclaration _ _ n _) = Just (TyName n)
```
JordanMartinez, line 537: I think this is why type class' kind signatures don't appear in the final docs whereas data, type, and newtype do. The returned value for `KindDeclaration`s here is `TyName`, which matches the one for `DataDeclaration` and `TypeSynonymDeclaration`. However, it doesn't match the returned value for `TypeClassDeclaration`.

JordanMartinez: This is indeed the issue. I'll push a few more commits and then this PR will be ready for review.

JordanMartinez:
The tests pass locally. This PR is ready for review.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
```diff
@@ -30,15 +32,27 @@ import Language.PureScript.Names
...
 exportedDeclarations (Module _ _ mn decls exps) = go decls
   where
   go = flattenDecls
-        >>> filter (isExported exps)
+        >>> reverse . snd . foldl' (f exps) (Nothing, [])
         >>> map (filterDataConstructors exps)
         >>> filterInstances mn exps
         >>> maybe id reorder exps
 
+  f :: Maybe [DeclarationRef] -> (Maybe Declaration, [Declaration]) -> Declaration -> (Maybe Declaration, [Declaration])
```
hdgarrood, line 46: I would prefer that we update `isExported` to handle kind declarations, rather than duplicating a slightly different version of it inline here. In general, I've found that having two slightly different versions of functions is a really good way to introduce bugs, especially as other changes go in over time. I think the best thing to do is to arrange things so that only one version of the function is necessary, but in cases where that's too difficult, then having one function which can work in both cases is still preferable to duplicating the implementation.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
```diff
@@ -152,5 +166,13 @@ reorder refs =
...
-    declName decl >>= flip M.lookup refIndices
+  refIndex = \case
+    KindDeclaration _ _ n _ ->
+      M.lookup (TyName n) refIndices <|> M.lookup (TyClassName (tyToClassName n)) refIndices
+    decl -> declName decl >>= flip M.lookup refIndices
+
+  -- Workaround to the fact that the kind's name's ProperNameType
+  -- isn't the same as the declaration's ProperNameType
+  -- when that declaration is a type class
+  tyToClassName :: ProperName 'TypeName -> ProperName 'ClassName
```
hdgarrood, line 177: Would you mind moving this into the `where` if it's only going to be used in `reorder`?

src/Language/PureScript/Docs/Types.hs
```diff
@@ -184,6 +185,15 @@ data DeclarationInfo
 
 instance NFData DeclarationInfo
 
+-- |
+-- Wraps enough information to properly render the kind signature
+-- of a data/newtype/type/class declaration.
+data KindInfo
+  = KindInfo P.KindSignatureFor Type'
```
hdgarrood, line 192: Can we make this into a record type with named parameters please? I know that this is already a common pattern in the compiler, but having to remember what the nth parameter of each data constructor means kind of sucks.

src/Language/PureScript/Docs/Types.hs
```diff
@@ -777,8 +805,15 @@ instance A.ToJSON Declaration where
              , "sourceSpan" .= declSourceSpan
              , "children"   .= declChildren
              , "info"       .= declInfo
+             , "kind"       .= declKind
              ]
 
+instance A.ToJSON KindInfo where
+ toJSON (KindInfo kindFor ty) =
+   A.object [ "keyword" .= P.kindSignatureForKeyword kindFor
```
hdgarrood, line 813: I think that if this is the only place we are using `kindSignatureForKeyword`, then it should live in this module alongside the `asKindSignatureFor` parser, because that's what it needs to be kept in sync with.

src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -100,66 +104,126 @@ getDeclarationTitle (P.ValueFixityDeclaration _ _ _ op) = Just (P.showOp op)
...
-  -- the actual type will be added during type checking.
-  basicDeclaration sa title (ValueDeclaration (P.TypeWildcard () Nothing))
-convertDeclaration (P.ExternDeclaration sa _ ty) title =
-  basicDeclaration sa title (ValueDeclaration (ty $> ()))
-convertDeclaration (P.DataDeclaration sa dtype _ args ctors) title =
-  Just (Right (mkDeclaration sa title info) { declChildren = children })
+basicDeclaration sa title = Just . Right . mkDeclaration sa title Nothing
+
+reassociateKindSignatures
+  :: (Maybe P.Declaration, [Either (P.Declaration, P.Declaration) P.Declaration])
```
hdgarrood, line 121: Did you consider reusing the existing `DeclarationAugment` mechanism here? I think it could be worth investigating whether we could an a new constructor to `DeclarationAugment` for kind signatures, and add the kind info inside `augmentWith`.

hdgarrood (commented):
Looks great so far!

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
JordanMartinez: Looks like `isExported` is only used for docs. 

How about I convert my inline code into a function called `filterDeclarations` and remove `isExported` completely?

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
JordanMartinez: Isn't it already part of that `where`? Or did you mean to move it into a `where` clause for just the `KindDeclaration` path? For example:
```haskell
KindDeclaration _ _ n _ ->
  -- code
  where
  tyToClasName :: ProperName 'TypeName -> ProperName 'ClassName
decl -> --
```

src/Language/PureScript/Docs/Convert/Single.hs
JordanMartinez: Not yet. I was just trying to get the code to build/work first and then get an initial review. I'll look into this.

src/Language/PureScript/Docs/Types.hs
JordanMartinez: Sure.

If you'd like, I could slowly update this codebase to use records where possible, bit by bit.

src/Language/PureScript/Docs/Types.hs
JordanMartinez: Will fix.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
JordanMartinez: Oh wait... Now I think I get what you're saying. My original approach is making this needlessly complicated.

hdgarrood: FWIW, longer term I think it would be good to remove this whole module, because it duplicates the existing logic in the desugaring pipeline for figuring out what is and isn't exported from a module. Ideally I think we would be able to reuse that logic for docs too, perhaps by looking at externs to figure out what is and isn't exported.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
hdgarrood: Yes sorry, you're right, I misread and didn't notice the identation before `tyToClassName`. Never mind!

src/Language/PureScript/Docs/Convert/Single.hs
JordanMartinez: Well that was a whole lot easier...

src/Language/PureScript/Docs/Types.hs
JordanMartinez: I've opened #4115 to track this. I'll mark this as resolved.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
JordanMartinez: In the latest updates, this is used twice now. So, I've moved it to a top-level member.

src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -66,8 +66,13 @@ type IntermediateDeclaration
 -- since they appear at the top level in the AST, and since they might need to
 -- appear as children in two places (for example, if a data type defined in a
 -- module is an instance of a type class also defined in that module).
+--
+-- The AugmentKindSig constructor allows us to add a kind signature
+-- to its corresponding declaration. Comments for both declarations
+-- are also merged together.
 data DeclarationAugment
   = AugmentChild ChildDeclaration
+  | AugmentKindSig (Maybe Text) P.KindSignatureFor Type'
```
JordanMartinez, line 75: Since this is three args, should it be converted into a `KindSigInfo` record type? For example
```haskell
data KindSigInfo = KindSigInfo
  { ksiComms :: Maybe Text
  , ksiKindSig :: P.KindSignatureFor
  , ksiType :: Type'
  }

data DeclarationAugment
  = AugmentChild ChildDeclaration
  | AugmentKindSig KindSigInfo

File: src/Language/PureScript/Docs/Convert/Single.hs [hunk 11]
```diff
@@ -159,6 +175,10 @@ convertDeclaration (P.ValueFixityDeclaration sa fixity (P.Qualified mn alias) _)
   Just . Right $ mkDeclaration sa title (AliasDeclaration fixity (P.Qualified mn (Right alias)))
 convertDeclaration (P.TypeFixityDeclaration sa fixity (P.Qualified mn alias) _) title =
   Just . Right $ mkDeclaration sa title (AliasDeclaration fixity (P.Qualified mn (Left alias)))
+convertDeclaration (P.KindDeclaration sa kindSig _ tys) title =
+  Just $ Left ([(title, AugmentType), (title, AugmentClass)], AugmentKindSig comms kindSig (tys $> ()))
```
JordanMartinez, line 179: Using `AugmentType` here means `ExternDeclaration`s are also considered when matching an augment with its corresponding declaration. Will that cause a problem? Or should I add a new AugmentType specifically for this concern?

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
```diff
@@ -152,5 +160,14 @@ reorder refs =
...
+    KindDeclaration _ _ n _ ->
+      M.lookup (TyName n) refIndices <|> M.lookup (TyClassName (tyToClassName n)) refIndices
+
+    decl -> declName decl >>= flip M.lookup refIndices
+
+-- |
+-- Workaround to the fact that a `KindDeclaration`'s name's `ProperNameType`
+-- isn't the same as the corresponding `TypeClassDeclaration`'s `ProperNameType`
+tyToClassName :: ProperName 'TypeName -> ProperName 'ClassName
+tyToClassName = coerceProperName
```
hdgarrood, line 173: This function is now accessible outside of `reorder`, which is what I was originally objecting to. This is what I meant:
```suggestion
  -- Workaround to the fact that a `KindDeclaration`'s name's `ProperNameType`
  -- isn't the same as the corresponding `TypeClassDeclaration`'s `ProperNameType`
  tyToClassName :: ProperName 'TypeName -> ProperName 'ClassName
  tyToClassName = coerceProperName
```

src/Language/PureScript/Docs/Convert/Single.hs
hdgarrood: I'm not sure. Why might it cause a problem?

src/Language/PureScript/Docs/Convert/Single.hs
hdgarrood: Yes, but can we avoid the abbreviations please? I don't think we need to abbreviate "comments" to "comms", for example.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
JordanMartinez: Yes, I moved it outside of the function because it's now used in two places rather than one. 

Since this is a potentially unsafe function, I'll duplicate the function and move it to a where clause in each place that it's used.

src/Language/PureScript/Docs/Convert/Single.hs
JordanMartinez: In case `isType` is ever updated in the future to include other potential declarations. Since this should only match on 4 declarations (data, type, newtype, class), I was wondering whether it might be better to create a separate function (e.g. `isKindAugmentable`) specifically for those and indicate that via a new `AugmentType` (e.g. `AugmentWithKind`) that uses that `isKindAugmentable` function.

JordanMartinez: Here's a link to [`isType`](https://github.com/purescript/purescript/pull/4100/files#diff-03c8dee30986517880ff71bf3ac056c8caec48c7bebef10fa1e7fb30c6b01d7fR257)

src/Language/PureScript/Docs/Convert/Single.hs
JordanMartinez: Done.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
hdgarrood: Ah right ok. If it's used in two places then I think defining it once at the top level is fine, as long as it's not exported.

src/Language/PureScript/Docs/Convert/Single.hs
hdgarrood: I'm not sure I completely understand. Are you imagining that if there's an ExternDeclaration and there is somehow also a kind signature for a declaration with the same name elsewhere in the module, that the kind signature would be incorrectly attached to the ExternDeclaration?

hdgarrood: I'd be inclined to say that it shouldn't matter because the name of an ExternDeclaration always starts with a lowercase letter, whereas the names of anything you can give kind signatures to (data types, classes) always start with uppercase letters, and so they can't ever overlap.

JordanMartinez: I think my question is probably a rabbit trail and not work thinking about further. However, let me try to explain my flow of thinking.

Kind signature declarations should only be matched to a data type / newtype (i.e. DataDeclaration), type synonym (i.e. TypeSynonymDeclaration), or a type class (i.e. TypeClassDeclaration) declaration. Kind signatures are never associated with FFI declarations (e.g. `ExternDataDeclaration`).

The `AugmentType` values (e.g. `AugmentClass` and `AugmentType`) merely [indicate which function to use](https://github.com/purescript/purescript/blob/c7cf026479788c74ed8190c6f6c4d97711217001/src/Language/PureScript/Docs/Convert/Single.hs#L95-L96) to determine if an augmentation matches a given declaration. 

Since we're currently using [`AugmentType` for the `KindDeclaration`](https://github.com/purescript/purescript/blob/c7cf026479788c74ed8190c6f6c4d97711217001/src/Language/PureScript/Docs/Convert/Single.hs#L185), the resulting function, [`isType`, will also return true when given a `ExternDataDeclaration`](https://github.com/purescript/purescript/blob/c7cf026479788c74ed8190c6f6c4d97711217001/src/Language/PureScript/Docs/Types.hs#L257). Yes, this isn't a problem now because `ExternDeclaration` is always lowercase whereas the other three declaration types start with uppercase letters. However, if we make changes later (e.g. casing changes; a new declaration is added that also happens to be lowercase) and somehow this declaration should happen to slip through, wouldn't this cause a bug?

So, my question is should we update the types to show something like this:
```haskell
-- 1. Define a new type that will
-- use a different matching function for 
-- augmenting declarations with kind signatures
-- | Different declarations we can augment
data AugmentType
  = AugmentClass
  -- ^ Augment documentation for a type class
  | AugmentType
  -- ^ Augment documentation for a type constructor
  | AugmentWithKindSignature
  -- ^ Augment data/newtype/type/class declaration with a kind signature

-- 2. Update the `matches` function in the `augmentDeclarations` code
-- to use this new function called `isKindAugmentable`
matches d (name, AugmentType) = isType d && declTitle d == name
matches d (name, AugmentClass) = isTypeClass d && declTitle d == name
matches d (name, AugmentWithKindSignature) = 
   isKindAugmentable d && declTitle d == name
   
-- 3. In Docs/Types.hs, add a new function 
-- that only returns True on those declarations
-- that can be augmented by a kind signature
isKindAugmentable :: Declaration ->  Bool
isKindAugmentable Declaration{..} =
  case declInfo of
    TypeSynonymDeclaration{} -> True
    DataDeclaration{} -> True
    TypeClassDeclaration{} -> True
    -- notice how ExternDataDeclaration returns False here
    _ -> False
```
```diff
 -- 4. Update the convertDeclaration 
 -- for kind signatures to use this new
 -- function
convertDeclaration (P.KindDeclaration sa kindSig _ tys) title =
- Just $ Left ([(title, AugmentType), (title, AugmentClass)], AugmentKindSig ksi)
+ Just $ Left ([(title, AugmentWithKindSignature)          ], AugmentKindSig ksi)

File: lib/purescript-cst/src/Language/PureScript/AST/Exported.hs (hunk 12 above)
Comment (JordanMartinez) [reply lib/purescript-cst/src/Language/PureScript/AST/Exported.hs]: True, but if it's not defined top-level, then it's less likely to be accidentally exported.

So, I can either define it once and add a note that it shouldn't be exported. Or I can define it locally twice in the places that use it. Which do you prefer?

File: src/Language/PureScript/Docs/Convert/Single.hs (hunk 11 above)
Comment (JordanMartinez) [reply src/Language/PureScript/Docs/Convert/Single.hs]: Or am I just overthinking this?

Comment (hdgarrood) [reply src/Language/PureScript/Docs/Convert/Single.hs]: I think this would be unnecessarily cautious. Changing the syntax so that types can start with lowercase letters is almost certainly never going to happen, and the compiler should reject attempts to give `foreign import data` declarations separate kind signatures anyway (because the `foreign import data` declaration itself already contains a kind signature).

File: lib/purescript-cst/src/Language/PureScript/AST/Exported.hs (hunk 12 above)
Comment (hdgarrood) [reply lib/purescript-cst/src/Language/PureScript/AST/Exported.hs]: I’d say define it once and don’t bother adding a note. It’s not really a big deal if it does end up being exported; we do coerce type names to class names and back in a bunch of places already, because that’s how type class desugaring works.

File: src/Language/PureScript/Docs/Convert/Single.hs (hunk 11 above)
Comment (JordanMartinez) [reply src/Language/PureScript/Docs/Convert/Single.hs]: Ok. Thanks for clarifying!

File: lib/purescript-cst/src/Language/PureScript/AST/Exported.hs (hunk 12 above)
Comment (JordanMartinez) [reply lib/purescript-cst/src/Language/PureScript/AST/Exported.hs:173]: Ok. I've just removed the last commit that made that change. I believe this PR should be ready to merge if there's no other issues with it.

File: src/Language/PureScript/Docs/Types.hs [hunk 13]
```diff
@@ -184,6 +185,17 @@ data DeclarationInfo
 
 instance NFData DeclarationInfo
 
+-- |
+-- Wraps enough information to properly render the kind signature
+-- of a data/newtype/type/class declaration.
+data KindInfo = KindInfo
+  { kiKindSigFor :: P.KindSignatureFor
+  , kiType :: Type'
```
hdgarrood, line 193: This is really a kind rather than a type, right? Maybe this field should be called `kiKind`?

JordanMartinez: Yeah, it is. I'll make the change. I think I'll also update other usages of the underlying value to say kind rather than type.

hdgarrood:
I think we're nearly there, nice work :) Can you just share a screenshot or link for what the rendered HTML docs look like now please? Also, it just occurred to me that I think we should include inferred kinds here as well. Does this branch do that already?

hdgarrood:
Oh, and we could do with some tests as well.

src/Language/PureScript/Docs/Types.hs
hdgarrood: Sounds good, thanks!

hdgarrood:
Let me know if you need any help with the tests - I realise the docs tests framework is a quite a lot to take in.

JordanMartinez:
> Oh, and we could do with some tests as well.

How would we test this? I didn't find any current examples that helped me know how I should proceed.

> Also, it just occurred to me that I think we should include inferred kinds here as well. Does this branch do that already?

I'm not sure. Is that figured out in the desugaring pipeline? Perhaps the docs' version of that pipeline needs to be updated to include that pass if so?

hdgarrood:
The tests are all in `tests/TestDocs.hs`, and they operate on the PureScript modules in `tests/purs/docs/src`. You'll probably want to add a new `DocsAssertion` constructor which asserts that a particular declaration has a kind signature matching the given kind, as well as a new `DocsAssertionFailed` constructor for when that assertion fails. You'll probably also want to add another module in the `tests/purs/docs/src` directory so that we can check that the docs do what we expect them both when kinds are inferred and when they are explicitly declared.

IIRC kind inference happens during type checking, which is after desugaring, so adjusting how we desugar for docs wouldn't help here. The docs pipeline already includes inferred types in cases where types are not explicitly provided by getting them out of externs, and I think it should be possible to do the same for kinds.

JordanMartinez:
<details>
<summary>Source code that has both kind and type comments</summary>

```purescript

module Main where

import Prelude

-- | Kind doc
data DE1 :: forall k. k -> Type
-- | decl doc
-- |
-- | new line
-- | new line
data DE1 a = DE1

-- | Kind doc
type TE1 :: forall k. k -> Type
-- | decl doc
type TE1 a = Int

-- | Kind doc
newtype NE1 :: forall k. k -> Type
-- | decl doc
newtype NE1 a = NE1 Int

-- | Kind doc
class CE1 :: Type -> Constraint
-- | decl doc
class CE1 a where
  foo1 :: a -> String

instance CE1 String where
  foo1 s = s
```
</details>

de1-te1
ne1-ce1


<details>
<summary>Source code that has only kind comments</summary>

```purescript

module Main where

import Prelude

-- | Kind doc
data DE2 :: forall k. k -> Type
data DE2 a = DE2

-- | Kind doc
type TE2 :: forall k. k -> Type
type TE2 a = Int

-- | Kind doc
newtype NE2 :: forall k. k -> Type
newtype NE2 a = NE2 Int

-- | Kind doc
class CE2 :: Type -> Constraint
class CE2 a where
  foo2 :: a -> String

instance CE2 String where
  foo2 s = s
```
</details>

de2-ne2
ce2

<details>
<summary>Source code that has only type comments</summary>

```purescript

module Main where

import Prelude

data DE3 :: forall k. k -> Type
-- | decl doc
data DE3 a = DE3

type TE3 :: forall k. k -> Type
-- | decl doc
type TE3 a = Int

newtype NE3 :: forall k. k -> Type
-- | decl doc
newtype NE3 a = NE3 Int

class CE3 :: Type -> Constraint
-- | decl doc
class CE3 a where
  foo3 :: a -> String

instance CE3 String where
  foo3 s = s
```
</details>

de3-ne3
ce3

hdgarrood:
Ok, great, thanks. The double-thickness line in between the kind signature and the data type definition is a bit ugly, I think. Can we get both the kind signature and the data type definition in the same `<code>` element so that there's no horizontal line in between them?

JordanMartinez:
> The tests are all in...

Ok, I'll look at this later today.

> The docs pipeline already includes inferred types in cases where types are not explicitly provided by getting them out of externs, and I think it should be possible to do the same for kinds.

Ok, so look at the externs code for help in figuring that out.

> Can we get both the kind signature and the data type definition in the same `<code>` element so that there's no horizontal line in between them?

In some ways, I think the line is preferable as it clearly indicates what is the kind signature part and what is the type itself. Consider the following comparison of a type with a more complicated kind signature:

complicated-type-1
complicated-type-2

JordanMartinez:
So, I propose keeping the line, but we could reduce it to a "single" line rather than the current "double" line above.

hdgarrood:
I think that works too, yeah.

JordanMartinez:
> Let me know if you need any help with the tests - I realise the docs tests framework is a quite a lot to take in.

Seems pretty straightforward now that I understand where to look. I always forget to check the `Test*.hs` files in the test folder. 

The most recent commit I've pushed adds a simple test that verifies that declarations with explicit kind signatures in the source code have them in their docs. It's my first attempt at confirming that I'm using the testing framework correctly.

tests/TestDocs.hs
```diff
@@ -106,6 +106,8 @@ data DocsAssertion
   | ShouldHaveLink P.ModuleName Text Text Docs.Namespace Docs.LinkLocation
   -- | Assert that a given declaration comes before another in the output
   | ShouldComeBefore P.ModuleName Text Text
+  -- | Assert that a given declaration has a kind signature
+  | ShouldHaveKindSignature P.ModuleName Text
```
hdgarrood, line 110: Can you please add another argument for the kind signature? We should test not just that the kind signature is present, but also that it is what we expect it to be.

JordanMartinez: Yup. That was the direction I was headed, but I ran out of time.

Also, shouldn't we also be testing that the keyword (i.e. `data`, etc.) is the one we expect?

hdgarrood: Sounds good! Up to you whether we test the keyword as well. That seems like it would be relatively low impact if we got it wrong, so I’m not sure whether the overhead of testing it would be worth it.

tests/TestDocs.hs
```diff
@@ -717,6 +738,22 @@ testCases =
       , ShouldHaveKindSignature (n "KindSignatureDocs") "TTypeOnly" "type TTypeOnly :: forall k. k -> Type"
       , ShouldHaveKindSignature (n "KindSignatureDocs") "NTypeOnly" "newtype NTypeOnly :: forall k. k -> Type"
       , ShouldHaveKindSignature (n "KindSignatureDocs") "CTypeOnly" "class CTypeOnly :: forall k. (k -> Type) -> k -> Constraint"
+
+      -- expected docs
+      , ShouldMergeDocComments (n "KindSignatureDocs") "DKindAndType" $ Just "dkatk\n\ndkatt\n"
+      , ShouldMergeDocComments (n "KindSignatureDocs") "TKindAndType" $ Just "tkatk\n\ntkatt\n"
+      , ShouldMergeDocComments (n "KindSignatureDocs") "NKindAndType" $ Just "nkatk\n\nnkatt\n"
+      , ShouldMergeDocComments (n "KindSignatureDocs") "CKindAndType" $ Just "ckatk\n\nckatt\n"
```
JordanMartinez, line 752: This actually surprised me. I thought the test should be `ckatk\n\nckatt` (as in a blank line between the two). I wasn't expecting the last doc-comment to also end in a newline.

tests/TestDocs.hs
```diff
@@ -739,6 +739,12 @@ testCases =
       , ShouldHaveKindSignature (n "KindSignatureDocs") "NTypeOnly" "newtype NTypeOnly :: forall k. k -> Type"
       , ShouldHaveKindSignature (n "KindSignatureDocs") "CTypeOnly" "class CTypeOnly :: forall k. (k -> Type) -> k -> Constraint"
 
+      -- Declarations with no explicit kind signatures should stlil have them implicitly.
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "DImplicit" "data DImplicit :: forall k. k -> Type"
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "TImplicit" "type TImplicit :: forall k. k -> Type"
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "NImplicit" "newtype NImplicit :: forall k. k -> Type"
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "CImplicit" "class CImplicit :: forall k1. (k1 -> Type) -> k1 -> Constraint"
```
JordanMartinez, line 746: Once inferred types are accounted for, this test should pass. Right now, it fails.

tests/TestDocs.hs
JordanMartinez: I think it's still important. I'd hate to see docs like this:
```
newtype Foo :: forall k. k -> Type
data Foo a = Foo
```
However, I felt it would be easier on the developer to just type in the expected kind signature rather than try to construct the proper one via Docs types.

JordanMartinez:
<details>
<summary>Latest commit turns the double line between the kind sig and type declaration into a single line. Below is a screenshot:</summary>

with-single-line-addition

</details>

hdgarrood:
The spacing looks a bit off there - can we make it so that the vertical space above and below the code on each side is the same?

JordanMartinez:
<details>
<summary>Fixed in latest commit</summary>

fix-vertical-space

</details>

JordanMartinez:
> IIRC kind inference happens during type checking, which is after desugaring, so adjusting how we desugar for docs wouldn't help here. The docs pipeline already includes inferred types in cases where types are not explicitly provided by getting them out of externs, and I think it should be possible to do the same for kinds.

I'm not sure how to proceed on this one. It sounds like we'd need to update the docs pipeline to also infer kinds before we convert the module into the Docs.module that is used to generate the HTML. Is that correct?

JordanMartinez:
Ah, actually, I think you're saying the externs should already have inferred kinds. We just need to extract them out and add them as a declaration's kind signature.

hdgarrood:
Yes, exactly. We already use the type-checking environment to fill in inferred types, so I'm hoping that using the same environment to fill in inferred kinds will be fairly straightforward. This happens here: https://github.com/purescript/purescript/blob/ee691c2509a4e8152d3a35c88e820bb1685f7512/src/Language/PureScript/Docs/Convert.hs#L48-L74

so I think if we could extend this function so that it inserts inferred kinds as well as types, that would be ideal.

rhendric:
If we're going to generate inferred kind signatures too, should we omit the signature if the kind is `(Type ->)* Type`?

hdgarrood:
I think that makes sense, yeah. So we omit kind signatures in docs if and only if they are not explicitly given and all of the data type’s parameters have kind Type?

JordanMartinez:
Thanks for the direction, Harry! The latest commits show your idea works. Now, I just need to figure out how to deal with the `(k :: Type)` part.

<details>
<summary>CI currently fails with these errors</summary>

```
purescript> Failures:
purescript> 
purescript>   tests/TestDocs.hs:37:13: 
purescript>   1) docs.Language.PureScript.Docs, Doc generation tests:, in module KindSignatureDocs, KindSignatureDocs.DImplicit should have the kind signature `data DImplicit :: forall k. k -> Type`
purescript>        expected the kind signature for DImplicit
purescript>        to be `data DImplicit :: forall k. k -> Type`
purescript>          got `data DImplicit :: forall (k :: Type). k -> Type`
purescript> 
purescript>   To rerun use: --match "/docs/Language.PureScript.Docs/Doc generation tests:/in module KindSignatureDocs/KindSignatureDocs.DImplicit should have the kind signature `data DImplicit :: forall k. k -> Type`/"
purescript> 
purescript>   tests/TestDocs.hs:37:13: 
purescript>   2) docs.Language.PureScript.Docs, Doc generation tests:, in module KindSignatureDocs, KindSignatureDocs.TImplicit should have the kind signature `type TImplicit :: forall k. k -> Type`
purescript>        expected the kind signature for TImplicit
purescript>        to be `type TImplicit :: forall k. k -> Type`
purescript>          got `type TImplicit :: forall (k :: Type). k -> Type`
purescript> 
purescript>   To rerun use: --match "/docs/Language.PureScript.Docs/Doc generation tests:/in module KindSignatureDocs/KindSignatureDocs.TImplicit should have the kind signature `type TImplicit :: forall k. k -> Type`/"
purescript> 
purescript>   tests/TestDocs.hs:37:13: 
purescript>   3) docs.Language.PureScript.Docs, Doc generation tests:, in module KindSignatureDocs, KindSignatureDocs.NImplicit should have the kind signature `newtype NImplicit :: forall k. k -> Type`
purescript>        expected the kind signature for NImplicit
purescript>        to be `newtype NImplicit :: forall k. k -> Type`
purescript>          got `newtype NImplicit :: forall (k :: Type). k -> Type`
purescript> 
purescript>   To rerun use: --match "/docs/Language.PureScript.Docs/Doc generation tests:/in module KindSignatureDocs/KindSignatureDocs.NImplicit should have the kind signature `newtype NImplicit :: forall k. k -> Type`/"
purescript> 
purescript>   tests/TestDocs.hs:37:13: 
purescript>   4) docs.Language.PureScript.Docs, Doc generation tests:, in module KindSignatureDocs, KindSignatureDocs.CImplicit should have the kind signature `class CImplicit :: forall k1. (k1 -> Type) -> k1 -> Constraint`
purescript>        expected the kind signature for CImplicit
purescript>        to be `class CImplicit :: forall k1. (k1 -> Type) -> k1 -> Constraint`
purescript>          got `class CImplicit :: forall (k1 :: Type). (k1 -> Type) -> k1 -> Constraint`
purescript> 
purescript>   To rerun use: --match "/docs/Language.PureScript.Docs/Doc generation tests:/in module KindSignatureDocs/KindSignatureDocs.CImplicit should have the kind signature `class CImplicit :: forall k1. (k1 -> Type) -> k1 -> Constraint`/"
purescript> 
```

</details>

<details>
<summary>Screenshot of `DImplicit`'s generated docs as of latest commits</summary>

inferred-kind

</details>

JordanMartinez:
> If we're going to generate inferred kind signatures too, should we omit the signature if the kind is (Type ->)* Type?

Could you help me understand why you think this is a good idea? I'm guessing it's because things with kind `Type`/`(Type ->)* Type` usually aren't that interesting at the kind-level? I get that we have naming conventions for things (e.g. `m` usually is a monad with kind `(Type -> Type)`), but I think including these types of kind signatures may actually make some things more readable. 

Consider the [ComponentSpec](https://pursuit.purescript.org/packages/purescript-halogen/6.1.2/docs/Halogen.Component#t:ComponentSpec) type from Halogen:
```purescript
-- I believe the kind signature for `ComponentSpec` is the following.
type ComponentSpec :: Type -> (Type -> Type) -> Type -> Row Type -> Type -> Type -> (Type -> Type) -> Type
type ComponentSpec state query action slots input output m = 
  { eval :: (HalogenQ query action input) ~> (HalogenM state action slots output m)
  , initialState :: input -> state
  , render :: state -> HTML (ComponentSlot slots m action) action 
  }
```
I believe this type should have its kind signature displayed in its docs only because of the `Row Type` kind for `slots`. For argument's sake, let's pretend the `slots` type parameter doesn't exist. By following the proposed rule above, a version of `ComponentSpec` without `slots` would not have its kind signature displayed in the docs. In other words, it would look like this:

```purescript
type ComponentSpec state query action input output m = 
  { eval :: (HalogenQ query action input) ~> (HalogenM state action output m)
  , initialState :: input -> state
  , render :: state -> HTML (ComponentSlot m action) action 
  }
```

Without the kind signature, I as a hypothetical reader who is unfamiliar with this library would have to look at the usages of each type parameter in the record to figure out what its kind is. Some are pretty obvious (e.g. `input` and `state` must be of kind `Type`). Others follow convention (i.e. `m` is probably a monad). Others (e.g. `query`, `action`, `output`) are not obvious because they are used in some other type (e.g. `HalogenQ`, `HalogenM`, etc.). I would have to look at those types to figure that out.

On another hand, this PR in its current state solves the problem of displaying kind signatures and their comments in the resulting documentation (pending discussion on whether the test for inferred kind signatures should 'pass' or not). What you've proposed above sounds like an optimization that can be built on top of the work I've done here. 

Assuming we do decide to implement the proposed idea above, wouldn't it make better sense to implement it in a separate PR? Correct me if I'm wrong, but wouldn't the implementation entail pattern matching on the `P.Type ()` data and figuring out whether the `declKind` of a Declaration should be updated with the inferred kind or not. If done wrongly, it might remove kind signatures that we'd otherwise want in the docs. So, to me, it seems like that will be a lot of work itself and would be best done in a PR separate from this one.

rhendric:
> Could you help me understand why you think this is a good idea?

I totally didn't explain my abuse of syntax, but what I meant was: omit kinds when they are `Type`, `Type -> Type`, `Type -> Type -> Type`, etc., but leave the kind signatures in for, e.g., `(Type -> Type) -> Type`.

Equivalently, omit kind signatures when every type parameter has kind `Type`.

Hopefully that makes a little more sense on the face of it. I'm not worried about confusion because, if the kind signature is missing, the arity of the type constructor is enough to reconstruct the entire kind.

Maybe it should be its own PR, but I think `Type -> Type -> Type` signatures are going to add a lot of noise and I'm skeptical about the net value of documenting inferred kinds unless something is done about that.

hdgarrood:
I agree that kind signatures are probably just noise in the case where every parameter has kind `Type`, and including those is going to be noisy and distracting, especially for users who are just beginning with PureScript and haven't come across kinds yet. How about, to keep the scope of this PR small, we go back to only including explicit kind signatures for now, and then in a later PR, we could start inserting inferred kinds according to @rhendric's suggestion - only when not every parameter has kind `Type`?

hdgarrood:
As for the inferred annotations like `forall (k :: Type). k -> Type`, I think we should strip the `:: Type` (the 'sort annotation' for `k`?) if the sort is `Type`, so that it comes out in the docs as `forall k. k -> Type`. (Aside: I'm not sure if it is possible to write a data type whose inferred kind has any sort other than `Type`. @natefaubion do you know?) But that can also happen in a follow-up PR, especially if we are going to hold off on inserting inferred kinds until a subsequent PR.

JordanMartinez:
> Equivalently, omit kind signatures when every type parameter has kind Type.

Ah... I was confusing your idea with something else entirely. I agree with the above idea. Restating it for myself, if I have a type like the following:
```purescript
-- inferred kind signature
-- type Foo :: Type -> Type -> Type -> Type
type Foo a b c = Either a (Either b c)
```
we wouldn't display the kind signature for `Foo` because `a`, `b`, and `c` all have kind `Type`. Displaying the kind signature would be unnecessarily verbose and only distract the reader. In cases like the `HalogenQ`/`HalogenM` issue above, we would know what the type parameters' kinds are as either because they are displayed (e.g. because other type parameters' kinds would be something other than just `Type`) or because they are not displayed (i.e. they are all kind `Type`). Either way, the reader would know.

Still, if we take this approach, we might want to update the Pursuit help pages to clarify this point. Otherwise, a reader would only know of this fact if they looked at the release notes of the PS version that would include this change.

> How about, to keep the scope of this PR small, we go back to only including explicit kind signatures for now, and then in a later PR, we could start inserting inferred kinds according to `rhendric`'s suggestion - only when not every parameter has kind Type?

Sounds good.

src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -128,7 +128,7 @@ mkDeclaration (ss, com) title info =
               , declSourceSpan = Just ss -- TODO: make this non-optional when we next break the format
               , declChildren   = []
               , declInfo       = info
-              , declKind       = Nothing -- kind sigs are added in augment pass
+              , declKind       = Nothing -- explicit kind signatures are added in augment pass
```
hdgarrood, line 131: The treatment of explicit versus inferred kind signatures is irrelevant to this piece of code, so I think it shouldn’t care about the distinction - especially because when we do update this to include inferred kind signatures, it’s going to be very easy to forget to update this comment, because the code in this function isn’t going to have to change. I’d suggest leaving this comment as it was before.

JordanMartinez: Ok. I've removed the commit that made that change.

hdgarrood (approved):
Looks great! Thanks for picking this up.

JordanMartinez:
@hdgarrood Thanks! One quick question. I noticed that the changelog doesn't really explain this feature. Should I merge this as is? Or further explain what we changed here in the release notes?

The current changelog just says

> * Display kind signatures and their comments in documentation (#4100 by JordanMartinez)

I was thinking of writing this:

----

* Display kind signatures and their comments in documentation (#4100 by JordanMartinez)

  Previously, data/newtype/type/class declarations that have explicit kind
  signatures would not display those kind signatures in their documentation.
  For example, the two below types...

  ```purescript
  data PolyProxy :: forall k. k -> Type
  data PolyProxy a = PolyProxy

  data TypeProxy :: Type -> Type
  data TypeProxy a = TypeProxy
  ```

  ... would only show the following information in their docs. One cannot
  be distinguished from another due to the missing kind signatures:

  ```
  data PolyProxy a = PolyProxy

  data TypeProxy a = TypeProxy
  ```

  Now, these types' kind signatures are displayed above their declarations
  in their docs, similar to what one would see in the source code.

hdgarrood:
Yeah, I think that sounds good.

JordanMartinez:
Huh... CI on Mac failed with this error. I'll restart the build.
```
   purescript    > bower purescript-either#5.0.0                        ECMDERR Failed to execute "git ls-remote --tags --heads https://github.com/purescript/purescript-either.git", exit code of #128 fatal: unable to access 'https://github.com/purescript/purescript-either.git/': Failed to connect to github.com port 443: Operation timed out
  purescript    > 
  purescript    > Additional error details:
  purescript    > fatal: unable to access 'https://github.com/purescript/purescript-either.git/': Failed to connect to github.com port 443: Operation timed out
  purescript    > 
```

Issue #4102, opened 2021-06-05
Questions on making a release


JordanMartinez:
Reading through the [RELEASE_GUIDE.md](https://github.com/purescript/purescript/blob/master/RELEASE_GUIDE.md), could I get answers to these questions?

Overall questions...

1. Can we produce a step-by-step "run these commands" checklist to ensure releases are done correctly?

1. How much of this guide needs to be updated? It seems like things are somewhat out of order. Moreover, the instructions never say to update the CHANGELOG.md file to include a new "Unreleased" section.

Regarding ["Before making a release"](https://github.com/purescript/purescript/blob/master/RELEASE_GUIDE.md#before-making-a-release)...

1. I'm supposed to compile the latest package set with (I imagine) a locally built binary using the `master` branch. Is there a script we can write that does this automatically?

1. What is there to check in INSTALL.md?

1. For the release notes, aren't we just copying and pasting the CHANGELOG.md entry and (possibly) editing it slightly?

Regarding ["Making a release"](https://github.com/purescript/purescript/blob/master/RELEASE_GUIDE.md#making-a-release)...

1. Let's say the GitHub-release-triggered CI build fails. How do I manually upload the prebuilt binaries? Are these instructions still relevant?

1. For the `stack upload` parts, do I need an account or some token to prove authorization?

1. What's a quick and easy way to see whether `purescript`/`purescript-cst` can be installed from Hackage?

1. For the `npm publish` parts, do I need an account or some token to prove authorization?

Regarding ["After making a release"](https://github.com/purescript/purescript/blob/master/RELEASE_GUIDE.md#after-making-a-release)...

1. Instructions say to announce the release on Twitter. Is that necessary (I don't have a Twitter account)? Do we have a `purescript` Twitter account that I'm supposed to use?

hdgarrood:
These are great questions, thanks for writing them up!

> Can we produce a step-by-step "run these commands" checklist to ensure releases are done correctly?

I’m sure there is scope for automating more of these steps, and that would be great, but I doubt we will be able to automate the entire process; I think some steps necessarily require human judgement, such as checking INSTALL.md is up to date (I’ll come back to that).

> How much of this guide needs to be updated? It seems like things are somewhat out of order. Moreover, the instructions never say to update the CHANGELOG.md file to include a new "Unreleased" section.

[...]

PR #4103, opened 2021-06-05
Bump versions for v0.14.2

JordanMartinez:
Preparing for a v0.14.2 release.

PR #4104, opened 2021-06-05
Proposal: edits to RELEASE_GUIDE.md

JordanMartinez:
**Description of the change**

Continuing a discussion in #4102.

---

**Checklist:**

### DIFF
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -13,7 +13,6 @@
 - Check that INSTALL.md is up-to-date
 - Regenerate LICENSE: `make license-generator` (see `license-generator/` for
   details)
-- Write release notes
 
 Additionally, if there are any breaking changes, there are number of downstream
 projects who we should probably at least notify:
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -56,10 +56,10 @@ make use of? If so:
 
 Has the compiler CLI changed at all? If so, the following may need updates:
 
-- spago
-- pulp
-- psc-package
-- purs-loader
+- [spago](https://github.com/purescript/spago)
+- [pulp](https://github.com/purescript-contrib/pulp)
+- [psc-package](https://github.com/purescript/psc-package)
+- [purs-loader](https://github.com/ethul/purs-loader)
 - ide plugins
 
 ### JSON formats
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -10,12 +10,36 @@
 ## Before making a release
 
 - Check that there are no unintended breaking changes by compiling [the latest package set](https://github.com/purescript/package-sets/releases/latest)
-- Check that INSTALL.md is up-to-date
-- Regenerate LICENSE: `make license-generator` (see `license-generator/` for
-  details)
 
-Additionally, if there are any breaking changes, there are number of downstream
-projects who we should probably at least notify:
+```bash
+stack build
+mkdir wPackageSet
+pushd wPackageSet
+spago init
+spago upgrade-set
+# install all packages in the set
+spago install $(spago ls packages | cut -f 1 -d ' ' | tr '\n' ' ')
+
+# Verify that code compiles and docs are properly created
+exec stack exec bash <<EOF
+spago build
+spago docs
+EOF
+popd
+# rm -rf wPackageSet
+```
+
+- Check that INSTALL.md is up-to-date. For example:
+    - Does the GHC version correspond to the one we're currently on?
+    - Is the list of supported OSes (based on GHC version) correct? If not, check the GHC website for the list.
+    - Are there any unofficial installation methods that should be dropped because they are no longer maintained?
+    - Are the instructions for building from source currently working?
+    - Anything else that needs to be changed?
+
+- Regenerate LICENSE: `make license-generator` (see `license-generator/` for details)
+
+- Additionally, if there are any breaking changes, there are number of downstream
+projects whom we should probably at least notify. See below subsections:
 
 ### Libraries
 
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -21,7 +21,7 @@ spago upgrade-set
 spago install $(spago ls packages | cut -f 1 -d ' ' | tr '\n' ' ')
 
 # Verify that code compiles and docs are properly created
-exec stack exec bash <<EOF
+stack exec bash <<EOF
 spago build
 spago docs
 EOF
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -1,5 +1,12 @@
 # Release Guide (for maintainers)
 
+## Prerequisites
+
+- You will need a [Hackage](https://hackage.haskell.org/) account that has been invited to be a maintainer of the `purescript` package on Hackage. If you don't have one, create one and ask to be invited as a maintainer.
+- You will need an [NPM](https://www.npmjs.com/) account that has been invited to be a maintainer of the `purescript` package on NPM. If you don't have one, create one and ask to be invited as a maintainer.
+- You need `spago` installed.
+- You need to be logged into NPM (i.e. running `npm whoami` should print your NPM account's username)
+
 ## Before making a release
 
 - Check that there are no unintended breaking changes by compiling [the latest package set](https://github.com/purescript/package-sets/releases/latest)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Bugfixes:
 
 Internal:
 
+* Update RELEASE_GUIDE.md with more details (#4104 by @JordanMartinez)
+
 ## v0.14.2
 
 New features:


RELEASE_GUIDE.md
```diff
@@ -1,26 +1,49 @@
...
+stack build
+mkdir wPackageSet
+pushd wPackageSet
+spago init
+spago upgrade-set
+# install all packages in the set
+spago install $(spago ls packages | cut -f 1 -d ' ' | tr '\n' ' ')
+
+stack exec purs compile $(spago sources | tr '\n' ' ')
+stack exec purs docs $(spago sources)
```
JordanMartinez, line 22: Note: these two `stack exec` commands don't work. I'm not sure how to pass the source globs correctly to the resulting `purs` binary.

RELEASE_GUIDE.md
```diff
@@ -44,48 +67,142 @@ considering what effects this may have:
...
+### Make the GitHub Release
 
-- Publish to Hackage:
+Once the above PR has been merged, create a release from the releases tab in GitHub, copying and pasting the version's corresponding section in the CHANGELOG.md file into the release as its release notes. This will also create a tag, which will kick off a CI build. The CI build will upload prebuilt compiler binaries to the release on GitHub when it completes.
 
-  - change to the `lib/purescript-cst` directory and run `stack upload .`
+If the GitHub-release-triggered CI build fails, the binaries will need to be built locally and manually uploaded to the release on GitHub.
 
-  - Finally, run `stack upload .` from the repo root directory.
+INSTRUCTIONS TO BE WRITTEN.
```
JordanMartinez, line 160: Would this be 
```bash
# if Linux
CI_RELEASE="true" ./bundle/build.sh "Linux"
# if Mac
CI_RELEASE="true" ./bundle/build.sh "macOS"
# if Windows
CI_RELEASE="true" ./bundle/build.sh "Windows"
```

JordanMartinez:
Should [these tools](https://github.com/purescript/purescript/pull/4104/files#diff-3dd37fe82be5315d56d3db6442ffe1e6993df584f98cc13d3aff2f02a027843eR56) have links to their respective repos?

Let's say there are issues with [these JSON formats](https://github.com/purescript/purescript/pull/4104/files#diff-3dd37fe82be5315d56d3db6442ffe1e6993df584f98cc13d3aff2f02a027843eR63-R66). What do we do?

RELEASE_GUIDE.md
f-f: If you run `stack exec -- bash` then you'll be dropped into a shell, so maybe that'd be easier?

RELEASE_GUIDE.md
```diff
@@ -1,26 +1,49 @@
...
-Additionally, if there are any breaking changes, there are number of downstream
-projects who we should probably at least notify:
+```bash
+stack build
+mkdir wPackageSet
+pushd wPackageSet
+spago init
+spago upgrade-set
+# install all packages in the set
+spago install $(spago ls packages | cut -f 1 -d ' ' | tr '\n' ' ')
```
f-f, line 19: We really need a `spago install --whole-set` 😄

f-f (commented):
Thank you for writing these up!

RELEASE_GUIDE.md
hdgarrood: Yeah, I think `stack exec bash` followed by `spago build` and `spago docs` is probably easiest.

RELEASE_GUIDE.md
```diff
@@ -44,48 +67,142 @@ considering what effects this may have:
 
 ## Making a release
 
-- Make a commit bumping versions. The following should be updated:
+A release is made by updating the versions within the repo via a PR. Once the PR is merged, a GitHub release is drafted, which triggers a CI build that should upload prebuilt binaries into that GitHub release page. (If it doesn't, you'll upload those manually).
```
hdgarrood, line 70: Can you avoid the passive voice for actions that the maintainer needs to take please? I think it is making it unclear which steps are performed automatically and which ones need to be done by the person making the release. Perhaps something like: "To make a release, a maintainer should create a PR updating versions in the repo, and once that is merged, they can create a GitHub release, which triggers a CI build ..."

RELEASE_GUIDE.md
```diff
@@ -44,48 +67,142 @@ considering what effects this may have:
...
+  # 2b. Rename the "Unreleased" section to
+  #   the version you will be releasing
+  # 2c. Edit the new version's content
+  #   until satisfied (e.g. grammer,
+  #   spelling, clarity, etc.)
+
+EDITOR="nano" # or your preferred editor
+$EDITOR CHANGELOG.md
+git add CHANGELOG.md
+git commit -m "Added new Unreleased section; finalized new version's section"
```
hdgarrood, line 103: I think "Finalizing CHANGELOG for {version}" might be a slightly clearer commit message, if we are going to suggest one here.

RELEASE_GUIDE.md
```diff
@@ -44,48 +67,142 @@ considering what effects this may have:
...
+make license-generator
+git add LICENSE
+git commit -m "Regenerated LICENSE file"
+
+# 2. Edit the changelog
+  # 2a. Create a new "Unreleased" section
+  # 2b. Rename the "Unreleased" section to
+  #   the version you will be releasing
+  # 2c. Edit the new version's content
+  #   until satisfied (e.g. grammer,
```
hdgarrood, line 97:
```suggestion
  #   until satisfied (e.g. grammar,
```

RELEASE_GUIDE.md
```diff
@@ -44,48 +67,142 @@ considering what effects this may have:
...
-      - The versions table in `lib/purescript-cst/README.md`,
+# 4. Create the PR either manually or
+#   via the `gh` CLI tool
+gh pr create
+# a. Make the `base` repo `purescript`
+# b. Push your branch to your fork
+# c. Continue on the web
 
-      - The version bounds for `purescript-cst` in `purescript.cabal`
+# 5. Push any other changes to the PR
```
hdgarrood, line 150: I am not sure that the changes to this section are an improvement. They add quite a bit of noise, and explain how to do basic tasks like creating git branches, opening files in  their editor, or submitting PRs, but we should assume that a maintainer creating a release knows how to do these things. I think the extra noise is obscuring the important information here, which is which parts of which files need updating.

RELEASE_GUIDE.md
hdgarrood: It would be along those lines, but you'll also need to make sure you've built a release version of the compiler first as well (so that `purs --version` doesn't say it's a development build). In the name of saving effort, since this scenario might not even happen, I'm tempted to say let's not bother writing instructions. Since we are the maintainers of the CI workflow/release script, we should be able to figure out how to run it if need be.

hdgarrood:
> Should these tools have links to their respective repos?

I don't see why not.

> Let's say there are issues with these JSON formats. What do we do?

There's no single answer to this, it depends on context. If a format has changed accidentally, then we should restore the old format before making the release. If a format has changed deliberately, we should make sure that we do a major version bump and that the change is noted in the breaking changes section of the changelog, and give the maintainers of any affected tools a warning in advance. Ideally all of this would happen at the time a PR is merged rather than when trying to cut a release though. We do have some backwards-compatibility JSON tests, but I don't think they're comprehensive right now.

RELEASE_GUIDE.md
JordanMartinez: Yeah... I have to agree with you here. It also doesn't account for people who are using editors that can easily navigate to such files.

RELEASE_GUIDE.md
JordanMartinez: I tried `stack exec bash` and it didn't drop into a shell. Found [this SO answer](https://stackoverflow.com/a/48014385) and think we should use this:
```bash
stack exec bash <<EOF
spago build
spago docs
EOF
```

JordanMartinez:
Following the state of the RELEASE_GUIDE.md as it is currently stated in this PR, I've...
- confirmed that the latest package set compiles with this version
- unsure whether and what needs to change in `INSTALL.md`, but I believe this can be resolved in a separate PR if needed.
- verified (by looking at the changelog) that there are no breaking changes
- merged the 'version bump and changelog finalization' PR via #4103

I'll now trigger the GH-release CI build. I still can't upload to Hackage yet, but should be able to update the npm package.

RELEASE_GUIDE.md
hdgarrood: Is this just `spago verify-set`?

RELEASE_GUIDE.md
```diff
@@ -42,50 +70,134 @@ considering what effects this may have:
 - JSON produced by `purs publish`
   - this might affect Pursuit
 
+Collaborate with other maintainers to identify what work needs to be done to resolve any issue.
+
 ## Making a release
 
-- Make a commit bumping versions. The following should be updated:
+To make a release, a maintainer should create a PR updating versions in the repo, and once that is merged, they can create a GitHub release, which triggers a CI build that should upload prebuilt binaries into that GitHub release page. (If it doesn't, you'll upload those manually).
```
hdgarrood, line 77: You've mentioned this both at the start and the end of this section. I don't think it's necessary to mention it twice; can we remove one?

RELEASE_GUIDE.md
```diff
@@ -42,50 +70,134 @@ considering what effects this may have:
...
+      version)
+
+    - In `purescript.cabal`, update the version bounds for `purescript-cst`
 
-  - The `version` field in `purescript.cabal`
+  - In `lib/purescript-cst/README.md`, update the versions table to include a new entry for the new PureScript release
 
-  - The `prerelease` field in `app/Version.hs`, if updating the prerelease
-    field
+  - In `purescript.cabal`, update the `version` field
```
hdgarrood, line 122: This item and all of the other items following it need to be dedented so they aren't sub-items of the "If purescript-cst has changed" part, because all of these things need to happen every time, regardless of whether purescript-cst has changed.

RELEASE_GUIDE.md
```diff
@@ -42,50 +70,134 @@ considering what effects this may have:
...
+```bash
+cd purescript
+
+# 1. Upload the `purescript-cst` package
+pushd purescript/lib/purescript-cst
+stack upload
+popd
+
+# 2. Upload the `purescript` package`
+stack upload
```
hdgarrood, line 170:
```suggestion
stack upload .
```

RELEASE_GUIDE.md
```diff
@@ -42,50 +70,134 @@ considering what effects this may have:
...
-- After all of the prebuilt binaries are present on the GitHub releases page,
-  publish to npm: change to the `npm-package` directory and run `npm publish`.
-  It's a good idea to check that the package can be installed from npm at this
-  point.
+```bash
+cd purescript
+
+# 1. Upload the `purescript-cst` package
+pushd purescript/lib/purescript-cst
+stack upload
```
hdgarrood, line 166: This doesn't work without the `.`
```suggestion
stack upload .
```

RELEASE_GUIDE.md
```diff
@@ -1,35 +1,63 @@
...
+stack build
+mkdir wPackageSet
+pushd wPackageSet
+spago init
+spago upgrade-set
+# install all packages in the set
+spago install $(spago ls packages | cut -f 1 -d ' ' | tr '\n' ' ')
+
+# Verify that code compiles and docs are properly created
+exec stack exec bash <<EOF
```
hdgarrood, line 24: You probably don't want the first `exec` here. It will make your terminal/shell exit after the command is complete.
```suggestion
stack exec bash <<EOF
```

RELEASE_GUIDE.md
```diff
@@ -1,35 +1,63 @@
...
+make use of? If so, make a "meta" issue to track which libraries need to be changed and then update those libraries:
 
-- Update core libraries
-- Update contrib libraries
-- Update node bindings
-- Update web bindings
+- core libs: open a "meta" issue in [`purescript/purescript` issue](https://github.com/purescript/purescript/issues)). For example, [#3942](https://github.com/purescript/purescript/issues/3942)
+- contrib libs: open a "meta" issue in [`purescript-contrib/governance` issue](https://github.com/purescript-contrib/governance/issues). For example, [#35](https://github.com/purescript-contrib/governance/issues/35)
+- node libs: open a "meta" issue in [`purescript-contrib/governance` issue](https://github.com/purescript-contrib/governance/issues). For example, [#35](https://github.com/purescript-contrib/governance/issues/35)
+- web libs: open a "meta" issue in [`purescript-contrib/governance` issue](https://github.com/purescript-contrib/governance/issues). For example, [#35](https://github.com/purescript-contrib/governance/issues/35)
```
hdgarrood, line 51: Typo here? "Open a meta issue in purescript/purescript issue" doesn't make sense to me. I'm also not sure we need to repeat the whole "open a "meta" issue" spiel for every item in this list.

```suggestion
- core libraries
- contrib libraries
- node libraries
- web libraries

It may be helpful to open a "meta" issue in the relevant repository to track progress for each group of libraries which need updating. For example, [purescript/purescript#3942](https://github.com/purescript/purescript/issues/3942).
```

RELEASE_GUIDE.md
```diff
@@ -42,50 +70,134 @@ considering what effects this may have:
...
 ## Making a release
 
-- Make a commit bumping versions. The following should be updated:
+To make a release, a maintainer should create a PR updating versions in the repo, and once that is merged, they can create a GitHub release, which triggers a CI build that should upload prebuilt binaries into that GitHub release page. (If it doesn't, you'll upload those manually).
+
+After that, the maintainer will upload the respective packages to Hackage and NPM.
+
+### Make last changes and submit the PR
+
+1. Checkout a new branch based on current `master`
```
hdgarrood, line 83: Let's get rid of this too

RELEASE_GUIDE.md
```diff
@@ -42,50 +70,134 @@ considering what effects this may have:
...
+# 2. Upload the `purescript` package`
+stack upload
+
+# 3. Verify that the two (i.e. `purescript-cst` and `purescript`) can be installed from Hackage.
+
+mkdir zHackageUpload
+pushd zHackageUpload
+cabal install purescript
+popd
+# rm -rf zHackageUpload
```
hdgarrood, line 178: If we're using cabal, then there's no need to do this in a separate directory, because `cabal install xyz` installs the package globally and doesn't do anything with the current working directory. But I think we should avoid adding separate instructions here anyway, because we should be using the same instructions as we provide in INSTALL.md for installing the compiler from source.


```suggestion
# 3. Verify that `purescript-cst` and `purescript` can be installed from Hackage by following the instructions in INSTALL.md for building the compiler from source.
```

hdgarrood:
Actually, maybe I'm going about reviewing this wrong. I think there are some changes in here which are good, but I also think that many of the places you've changed are better off as they are at the moment. So maybe I should just list the bits that I think are good and should go in, and we can just get those merged?

- :+1: on the prerequisites section
- :+1: on adding instructions for checking that there are no unintended breaking changes by building the package set
- :+1: on the clarification on how to check that INSTALL.md is up to date
- :+1: on linking to the tools

But I don't think the "Make a release" section is improved by these changes, so I'd prefer to leave it as-is.

RELEASE_GUIDE.md
JordanMartinez: Whoops!

JordanMartinez:
@hdgarrood I've updated this PR to only include the :+1: changes from your previous comment. I'd rather get some of these in even if not all of them were accepted.

We may disagree philosophically on what this file should contain, and that's ok! Even if "maintainers should know that," my goal was to provide a checklist to prevent human error. Personally, I like procedures that tell me exactly what to do rather than leaving room for interpretation. If I have to spend 5 minutes interpreting what I'm supposed to do or remember which commands to run in the terminal, I just wasted 5 minutes. However, I realize that some of the changes I proposed above also specify the exact workflow to use. That's not good either because the approach I proposed here now may not be the one another uses, or even by the one I use later on.

On another hand, this PR may have been an overreaction on my part. Since I haven't made a PureScript release before, everything is new. Now that I have made one, I see how it works and some of the changes I proposed here might not be needed.

RELEASE_GUIDE.md
JordanMartinez: I think it would replace the `spago install` and `spago build` command. However, I am not sure whether it also builds docs. 

@f-f, could you clarify?

RELEASE_GUIDE.md
JordanMartinez: This was fixed in latest commits.

hdgarrood (approved):

RELEASE_GUIDE.md
f-f: The question "can I install the entirety of the set?" seems to come up often, and I don't have a good answer for that right now (it will improve once we get the new config format for Spago).

In this thread however, I think Harry is right: installing the whole set and then building it is _basically_ `verify-set` (which does some more stuff, so it's going to be slightly slower than just installing everything and building it, but it's fine)

PR #4118, opened 2021-06-22
Remove generated names from errors about instances

rhendric:
**Description of the change**

This is the PR I expressed an intention to make in https://github.com/purescript/purescript/pull/4096#discussion_r643282717; see also https://github.com/purescript/purescript/pull/4096#discussion_r654928683.

The intent here is to exclude any instance names that the user wouldn't recognize because they were generated by the compiler. Either these names can be dropped altogether because the error message is positioned on the offending instance, or they should be replaced with a suitable description of the instance.

I believe this covers all of the error messages that currently mention instance names.

The specific formatting of the errors is very much up for discussion. I'm not thrilled with how verbose they are, but I also don't know what I'd cut from them.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Bugfixes:
 
 * Fix UnusedName warnings for multiple non-recursive let bindings (#4114 by @nwolverson)
 
+* Remove generated names from errors about instances (#4118 by @rhendric)
+
 Internal:
 
 * Fix for Haddock (#4072 by @ncaq and @JordanMartinez)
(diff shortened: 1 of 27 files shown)


JordanMartinez:
The code looks right. I'm still looking at the error messages.

As an example, I tried removing the special characters in [TypeSynonymsOverlappingUnnamedInstance.purs](https://github.com/purescript/purescript/blob/23695308293f044805b95faa1e85537832753a8e/tests/purs/failing/TypeSynonymsOverlappingUnnamedInstance.out). I'm not sure if I did it correctly, but this is what an error message might look like:
```
Error found:
in module Mai
at tests/purs/failing/TypeSynonymsOverlappingUnnamedInstance.purs:14:1 - 15:16 (line 14, column 1 - line 15, column 16)

    Main.Convert String 
                 String

    instance in module Main with type Convert String String (line 11, column 1 - line 12, column 16)
    instance in module Main with type Convert String String (line 14, column 1 - line 15, column 16)

in type class instance

  Main.Convert String 
               String

See https://github.com/purescript/documentation/blob/master/errors/OverlappingInstances.md for more information,
```

JordanMartinez (approved):

hdgarrood (approved):
Looks good to me!

PR #4119, opened 2021-06-22
Display inferred kind signatures in HTML docs

### LINKED ISSUE #4117: Include inferred kind signatures in HTML docs when no kind signature declaration is provided explicitly
## Summary

Related to #3973. Once #4100 gets merged, explicit kind signatures will be displayed in the HTML docs. [First proposed in #4100](https://github.com/purescript/purescript/pull/4100#issuecomment-864039666), we should also include implicit (i.e. inferred) kind signatures in docs.

## Proposal

#4100 did not include inferred kinds in its PR because it seemed best to add them in a separate PR. 

To add the inferred kind signatures, we can adjust the [`insertValueType` function](https://github.com/purescript/purescript/pull/4100#issuecomment-864416670) to look up the corresponding kind in the `Environment` using the declaration's name. This idea already proved to work in #4100 when I started to implement it there around the same time the below guidelines were proposed.

However, there are a few guidelines on how this should be done:
1. The final kind signature inferred by the compiler may need to be adjusted in some circumstances. For example, a declaration of `data Foo a = Foo` will have the inferred kind signature of `forall (k :: Type). k -> Type` rather than the simpler and expected `forall k. k -> Type` (edited. Original was <del>`Type -> Type`</del>). We would need to rewrite such cases to their expected version. We could call this "kind signature docs desugaring" or something.
2. Inferred kind signatures where all type parameters have kind `Type` should not be displayed ([proposal](https://github.com/purescript/purescript/pull/4100#issuecomment-864427332) and [clarification](https://github.com/purescript/purescript/pull/4100#issuecomment-864491464)). This should occur after the "kind signature docs desugaring" step as we may not find some signatures that satisfy this condition until after they are desugared.

Unanswered questions:
- Let's say a user adds an explicit kind signature declaration where all type parameters have kind `Type`. Should we remove that  kind signature from that declaration's documentation? Or continue to let users override this behavior if they so desire?
- Since I'm unfamiliar with the `P.Type` type, is there pre-existing code that shows what each data constructor represents? (Maybe I just need to look around a bit more)
- The questions/comments raised in [Harry's comment about the 'sort annotation'](https://github.com/purescript/purescript/pull/4100#issuecomment-864546458)

JordanMartinez:
**Description of the change**

Fixes #4117. 

---

**Checklist:**

Work done so far:
- The `forall (k :: Type). k -> Type` has been simplified to `forall k. k -> Type`.

Pending work
- whether to display inferred kind signatures if all type parameters' kinds are `Type` (decision pending)
- updating the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,37 +32,38 @@ New features:
   ```
 
   Now, these types' kind signatures are displayed above their declarations
-  in their docs, similar to what one would see in the source code.
+  in their docs, similar to what one would see in the source code. The only
+  exception to this are kind signatures whose kinds are "uninteresting"
+  (covered below).
 
-  But that's not all!
-
-  The two below types do not have explicit kind signatures, but the compiler
-  infers what their kind signature is anyway. "Interesting" kind signatures
-  will be displayed but "uninteresting" kind signatures (even if they are
-  explicit) will be removed.
+  Moreover, declarations without explicit kind signatures will still
+  display the kind signatures inferred by the compiler. For example,
+  `Foo`'s documentation will display the below (commented-out) inferred
+  kind signature:
 
   ```purescript
                                                   {-
   data Foo :: (Type -> Type) -> Type -> Type      -}
-  data Foo f a = Foo (f a)
-                                                  {-
-  data Tuple :: Type -> Type -> Type              -}
-  data Tuple a b = Tuple a b
+  data Foo f a = Foo (f Int) a
   ```
 
-  The docs will include the kind signature for `Foo`, but will
-  intentionally remove the kind signature for `Bar`, even if `Bar`'s
-  kind signature was explicitly defined by the developer. `Bar`'s
-  kind signature is considered "uninteresting" because all of the
-  type parameters have kind `Type`. `Foo`, on the other hand, has
-  at least one type parameter, `f`, whose kind is something other than
-  `Type`. Thus, `Foo`'s inferred kind signature will be displayed
-  in its docs.
-
-  An "uninteresting" kind signature is one that follows this form:
+  However, a kind signature for both explicit and inferred kinds will not
+  be displayed if that kind is considered "uninteresting." An "uninteresting"
+  kind is one that follows this form:
   - `Type`
   - `Constraint`
   - `Type -> K` where `K` refers to an "uninteresting" kind signature
+
+  Thus, both the `Bar` and `Baz` declarations below will not have their
+  explicit and inferred kind signatures displayed in their docs.
+  ```purescript
+  data Bar :: Type -> Type -> Type
+  data Bar a b = Bar a b
+                                                    {-
+  class Baz :: Type -> Type -> Constraint           -}
+  class Baz a b where
+    doBaz :: a -> b -> String
+  ```
 
 Bugfixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Breaking changes:
 
 New features:
 
-* Display kind signatures and their comments in documentation (#4100 by JordanMartinez)
+* Display kind signatures and their comments in documentation (#4100 and #4119 by JordanMartinez)
 
   Previously, data/newtype/type/class declarations that have explicit kind
   signatures would not display those kind signatures in their documentation.
@@ -33,6 +33,36 @@ New features:
 
   Now, these types' kind signatures are displayed above their declarations
   in their docs, similar to what one would see in the source code.
+
+  But that's not all!
+
+  The two below types do not have explicit kind signatures, but the compiler
+  infers what their kind signature is anyway. "Interesting" kind signatures
+  will be displayed but "uninteresting" kind signatures (even if they are
+  explicit) will be removed.
+
+  ```purescript
+                                                  {-
+  data Foo :: (Type -> Type) -> Type -> Type      -}
+  data Foo f a = Foo (f a)
+                                                  {-
+  data Tuple :: Type -> Type -> Type              -}
+  data Tuple a b = Tuple a b
+  ```
+
+  The docs will include the kind signature for `Foo`, but will
+  intentionally remove the kind signature for `Bar`, even if `Bar`'s
+  kind signature was explicitly defined by the developer. `Bar`'s
+  kind signature is considered "uninteresting" because all of the
+  type parameters have kind `Type`. `Foo`, on the other hand, has
+  at least one type parameter, `f`, whose kind is something other than
+  `Type`. Thus, `Foo`'s inferred kind signature will be displayed
+  in its docs.
+
+  An "uninteresting" kind signature is one that follows this form:
+  - `Type`
+  - `Constraint`
+  - `Type -> K` where `K` refers to an "uninteresting" kind signature
 
 Bugfixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,60 +10,18 @@ New features:
 
 * Display kind signatures and their comments in documentation (#4100 and #4119 by JordanMartinez)
 
-  Previously, data/newtype/type/class declarations that have explicit kind
-  signatures would not display those kind signatures in their documentation.
-  For example, the two below types...
+  The compiler now displays kind signatures for data, newtype, type
+  synonym, and type class declarations in generated documentation. The
+  compiler now also includes documentation-comments (i.e. those which start
+  with a `|` character) both above and below the associated kind signature
+  declaration (if any) in generated documentation, whereas previously
+  documentation-comments above a kind signature declaration were ignored.
 
-  ```purescript
-  data PolyProxy :: forall k. k -> Type
-  data PolyProxy a = PolyProxy
-
-  data TypeProxy :: Type -> Type
-  data TypeProxy a = TypeProxy
-  ```
-
-  ... would only show the following information in their docs. One cannot
-  be distinguished from another due to the missing kind signatures:
-
-  ```
-  data PolyProxy a = PolyProxy
-
-  data TypeProxy a = TypeProxy
-  ```
-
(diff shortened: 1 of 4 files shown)


JordanMartinez:
Pretty sure this is ready for review. I still need to update the changelog to account for this PR.

tests/TestDocs.hs
```diff
@@ -740,6 +762,32 @@ testCases =
...
+      , ShouldNotHaveKindSignature (n "KindSignatureDocs") "CHidden"
+      , ShouldNotHaveKindSignature (n "KindSignatureDocs") "CNothing"
+
+      -- Declarations with no explicit kind signatures should be displayed
+      -- if at least one type parameter has a kind other than `Type`
+      -- despite all others having kind `Type`.
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "DShown" "data DShown :: Type -> Type -> (Type -> Type) -> Type"
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "TShown" "data TShown :: (Type -> Type) -> Type -> Type -> Type"
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "NShown" "data NShown :: Type -> (Type -> Type) -> Type -> Type"
+      , ShouldHaveKindSignature (n "KindSignatureDocs") "CShown" "data CShown :: (Type -> Type) -> Type -> Type -> Type"
```
rhendric, line 789: As the failed build shows, these expectations should be changed to have the correct keywords.

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -71,6 +86,65 @@ insertValueTypes env m =
...
+          -- - Type -> Constraint
+          -- - Type -> Type -> ... -> Type
+          -- - Type -> Type -> ... -> Constraint
+          isFunctionTypeStarTerminal
+            :: P.Type () -> Bool
+          isFunctionTypeStarTerminal = \case
+            P.TypeApp _ t1 t2 | t1 == kindFunctionType ->
+              isFunctionTypeStarTerminal t2
+            x
+              | x == kindPrimType || x == kindPrimConstraint -> True
```
rhendric, line 134: The `Constraint` thing is a good point; I hadn't considered that classes have a different ‘most intuitive’ kind.

This `||` is correct because there's no way to make something `Constraint`-kinded that isn't a class right now, right? That is, something like
```purs
type IntState = MonadState Int
```
currently isn't supported, and there isn't anything else like that which could create a `Constraint` other than a class?

I'd feel a tiny bit more comfortable if this `||` were instead conditional on the declaration type—a `x == if isTypeClass d then kindPrimConstraint else kindPrimType` sort of deal—because some day type synonyms for constraints might be a thing? But perhaps that's unnecessary future-proofing, if we know for sure that there aren't any `Constraint` loopholes in the present.

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -97,17 +97,51 @@ insertValueTypesAndInferredKinds env m =
...
+          inferredKind' = inferredKind $> ()
           kindPrimType = P.TypeConstructor () Prim.Type
+          kindPrimFunction = P.TypeConstructor () Prim.Function
+          kindPrimConstraint = P.TypeConstructor () Prim.Constraint
+
+          -- `Type ->`
+          kindFunctionType = P.TypeApp () kindPrimFunction kindPrimType
+
+          -- |
+          -- Returns True if the inferred kind signature follows this
```
hdgarrood, line 117: How about calling this function `isUninteresting` and saying that an "uninteresting" kind is a kind of the form:
- `Type`
- `Constraint`
- `Type -> K` where `K` is uninteresting.
I think that makes it clearer which kinds are included, and I think it will make the code a little easier to read.

hdgarrood:
This looks great :+1: it looks to me like we will currently only hide uninteresting kinds if they have been inferred, so I think the only thing left to change is to additionally hide uninteresting kinds which are explicitly declared, if that's what we want to do (and I think it probably is, judging by the discussion in #4117?)

src/Language/PureScript/Docs/Convert.hs
hdgarrood: That's a good suggestion - I think full constraint kinds (which includes being able to write type synonyms like your `IntState` one) is something that's likely to happen in the future, so I agree about making this conditional on the declaration type.

hdgarrood: https://github.com/purescript/purescript/issues/1764

src/Language/PureScript/Docs/Convert.hs
JordanMartinez: Ah, thanks! That's a much better name.

JordanMartinez:
> I think the only thing left to change is to additionally hide uninteresting kinds which are explicitly declared

Yeah, that's something I wasn't able to get to this morning due to running out of time.

tests/TestDocs.hs
JordanMartinez: Ha! Whoops!

JordanMartinez:
Ok. Tests pass locally, and I believe I've addressed all comments. Let me know if there are any stylistic changes to make.

JordanMartinez:
Now I just need to write the changelog...

JordanMartinez:
Oh, and update the Pursuit page once this gets released.

CHANGELOG.md
```diff
@@ -34,6 +34,36 @@ New features:
...
+                                                  {-
+  data Foo :: (Type -> Type) -> Type -> Type      -}
+  data Foo f a = Foo (f a)
+                                                  {-
+  data Tuple :: Type -> Type -> Type              -}
+  data Tuple a b = Tuple a b
+  ```
+
+  The docs will include the kind signature for `Foo`, but will
+  intentionally remove the kind signature for `Bar`, even if `Bar`'s
```
rhendric, line 54: Bar or Tuple?

CHANGELOG.md
```diff
@@ -34,6 +34,36 @@ New features:
...
+  But that's not all!
+
+  The two below types do not have explicit kind signatures, but the compiler
+  infers what their kind signature is anyway. "Interesting" kind signatures
+  will be displayed but "uninteresting" kind signatures (even if they are
+  explicit) will be removed.
+
+  ```purescript
+                                                  {-
+  data Foo :: (Type -> Type) -> Type -> Type      -}
```
hdgarrood, line 46: The compiler will infer `forall k. (k -> Type) -> k -> Type` for this data type, not `(Type -> Type) -> Type -> Type`.

CHANGELOG.md
```diff
@@ -34,6 +34,36 @@ New features:
...
+  The docs will include the kind signature for `Foo`, but will
+  intentionally remove the kind signature for `Bar`, even if `Bar`'s
+  kind signature was explicitly defined by the developer. `Bar`'s
+  kind signature is considered "uninteresting" because all of the
+  type parameters have kind `Type`. `Foo`, on the other hand, has
+  at least one type parameter, `f`, whose kind is something other than
+  `Type`. Thus, `Foo`'s inferred kind signature will be displayed
+  in its docs.
+
+  An "uninteresting" kind signature is one that follows this form:
```
hdgarrood, line 62: I think we should say "kind" rather than "kind signature" here. A kind is something like `Type`, `Type -> Type`, or `forall k. k -> Type`. A kind signature is more than that: it's a declaration which tells you that a particular data type or type synonym or class has a particular kind, such as `data Tuple :: Type -> Type -> Type`.

CHANGELOG.md
JordanMartinez: Ah! Ran out of time again. Thanks for pointing that out.

CHANGELOG.md
JordanMartinez: Thanks. I decided to adjust this example to use `(f Int)` rather than `(f a)`

CHANGELOG.md
JordanMartinez: Mm... Makes sense. I've rewritten parts of the changelog entry to keep this in mind.

JordanMartinez:
I rewrote the changelog entry to use this organization. I thought it better presented the new feature in a clearer way.
- kind signatures weren't displayed
- kind signatures are now displayed (except those with 'uninteresting' kinds)
- inferred kind signatures are used when explicit ones aren't provided
- kind signatures with 'uninteresting' kinds (both explicit and inferred) are not displayed

JordanMartinez:
If this is good to merge, can I get an approval?

hdgarrood:
Please don’t rush people. I would still like to see @rhendric’s suggestion about making the “uninteresting” check depend on what kind of declaration we have implemented, and I would like to think a bit more about how we present this in the release notes.

JordanMartinez:
Ah... Sorry about that.

hdgarrood:
I think this changelog entry is a bit verbose. I don't think we should always provide examples of everything; that makes it difficult to quickly scan through a changelog to find what version a given change was made in. I think a more full description of how the feature works with examples to ensure that the reader understands would be better suited for the documentation repository. How about this instead:

  The compiler now displays kind signatures for data, newtype, type
  synonym, and type class declarations in generated documentation. The
  compiler now also includes documentation-comments (i.e. those which start
  with a `|` character) both above and below the associated kind signature
  declaration (if any) in generated documentation, whereas previously
  documentation-comments above a kind signature declaration were ignored.
  
  Both explicitly declared and inferred kinds are included in documentation.
  The compiler omits including a kind signature in generated documentation
  only when the kind is considered "uninteresting". An uninteresting kind is
  defined as one where all of the declaration's type parameters have kind
  `Type`.

JordanMartinez:
>  I think a more full description of how the feature works with examples to ensure that the reader understands would be better suited for the documentation repository.

Works for me!

JordanMartinez:
I've added an issue to the docs repo, so we can add that page later.

hdgarrood (approved):

PR #4121, opened 2021-06-26
Display roles in HTML docs

### LINKED ISSUE #4116: Include role information in HTML docs
## Summary

Roles were added in `v0.14.0` with the work of the Coercible PRs. However, they do not currently appear in the HTML documentation.

## Motivation

HTML docs should contain information on roles as that can affect whether `coerce` can be used or not. One cannot know whether the below type, [`Map`](https://pursuit.purescript.org/packages/purescript-ordered-collections/2.0.1/docs/Data.Map.Internal#t:Map), can be `coerce`d unless they either [read the source code](https://github.com/purescript/purescript-ordered-collections/blob/v2.0.1/src/Data/Map/Internal.purs#L69-L74) or try to coerce a value via the REPL and see if they get a compiler error.

```purescript
data Map k v = ---

type role Map nominal representational
```

## Proposal

`Map` above does not export its constructors, so I don't think it's a good reference point for discussing where role annotations should appear in the docs. So, let's use [`Either`](https://pursuit.purescript.org/packages/purescript-either/5.0.0/docs/Data.Either#t:Either) instead for reference,

In #4100, kind signatures are displayed above the type's definition. For this situation, I think role annotations should be displayed under the type, following what we do in source code. Thus, the complete docs for a type that includes both a kind signature and a role annotation would look like this...
```
data Either :: Type -> Type -> Type
data Either l r
type role Either representational representational
```
... before we display anything else.

Right now, after display the type information (i.e. the content above), we then display the comments, then any constructors, and then any instances. Since the three entities above are each represented as three declarations in the AST, each with its own comments, we need to merge those comments together. In #4100, I "merged" the comments together by displaying kind signature comments first and then the type's comments, separating them with a newline. We could take a similar approach here. For example, given the following source code..
```purescript
-- | kind declaration
data Either :: Type -> Type -> Type
-- | type declaration
data Either l r
-- | role declaration
type role Either representational representational
```
.. we would show the following in the docs
```
kind declaration
type declaration
role declaration
```

The implementation would follow the same approach used in #4100:
- ensure RoleDeclarations are not filtered out
- use the `augment` pass to add role information and merge its comments with existing comments (if any)
- render the role in the final HTML

JordanMartinez:
**Description of the change**

Fixes #4116 

<del>WIP but should help continue the discussion already started.</del>

---

**Checklist:**

---

Current changes made in this PR
- role declarations will now be exported in `AST.Exported.exportedDeclarations`
- another augmentation pass has been added that inserts the roles into the corresponding declaration
- the child info types for data, newtype, and type declarations now include a `[Role]` argument

I haven't yet tried to render the roles in the HTML docs for a few reasons. First, we're still coming to an agreement about how we want to do this (though the subscript idea seems to be favored). Second, I'm still figuring out how the rendering code works and how I would insert role subscripts.

I'll ask other questions in the code itself.


lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
```diff
@@ -134,6 +134,9 @@ isExported (Just exps) (KindDeclaration _ _ n _) = any matches exps
   matches declRef = do
     let refName = declRefName declRef
     TyName n == refName || TyClassName (tyToClassName n) == refName
+isExported (Just exps) (RoleDeclaration RoleDeclarationData{..}) = any matches exps
+  where
+  matches declRef = TyName rdeclIdent == declRefName declRef
 isExported (Just exps) decl = any matches exps
   where
   matches declRef = declName decl == Just (declRefName declRef)
```
JordanMartinez, line 142: If `declName` also returned the `RoleDeclaration`'s name, then the above case for `RoleDeclaration`s wouldn't be needed. I'll have to see where else `declName` is used and whether it should include `RoleDeclaration`s or whether my current approach is best.

lib/purescript-cst/src/Language/PureScript/AST/Exported.hs
```diff
@@ -164,6 +167,9 @@ reorder refs =
     KindDeclaration _ _ n _ ->
       M.lookup (TyName n) refIndices <|> M.lookup (TyClassName (tyToClassName n)) refIndices
 
+    RoleDeclaration RoleDeclarationData{..} ->
+      M.lookup (TyName rdeclIdent) refIndices
+
     decl -> declName decl >>= flip M.lookup refIndices
```
JordanMartinez, line 173: Similarly here. If `declName` included `RoleDeclaration`s, the above case isn't needed.

src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -106,6 +108,26 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
...
         Just $ kcoms <> "\n" <> dcoms
+  augmentWith (AugmentRole comms roles) d =
+    d { declComments = mergeComments (declComments d) comms
+      , declInfo = insertRoles
+      }
+    where
+      mergeComments Nothing rc = rc
+      mergeComments dc Nothing = dc
+      mergeComments (Just dcoms) (Just rcoms) =
+        Just $ dcoms <> "\n" <> rcoms
```
JordanMartinez, line 119: I've opted to merge the role annotation's comments under the declaration's comments. Depending on how we visualize content, we might want to store the role annotation comments in the `declInfo` field, too.

src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -106,6 +108,26 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
...
+
+      insertRoles = case declInfo d of
+        DataDeclaration dataDeclType args [] ->
+          DataDeclaration dataDeclType args roles
+        TypeSynonymDeclaration args ty [] ->
+          TypeSynonymDeclaration args ty roles
+        DataDeclaration _ _ _ ->
+          P.internalError "augmentWith: could not add a second role to data declaration"
+        TypeSynonymDeclaration _ _ _ ->
```
JordanMartinez, line 128: One question, can FFI have role declarations? I think it can, but I couldn't remember. I'm pretty sure it defaults to nominal unless otherwise stated, correct? If it can, I need to add a case for `ExternDeclaration` here.

hdgarrood: Yes https://github.com/purescript/purescript-effect/blob/985d97bd5721ddcc41304c55a7ca2bb0c0bfdc2a/src/Effect.purs#L18

JordanMartinez: Thanks. I've added them.

JordanMartinez:
Here's the control flow for the rendering pipeline
1. a declaration is rendered via [`renderDeclaration`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/AsHtml.hs#L153)
1. the info part of the docs is rendered via a couple of parts:
    1. Convert [the args back into a `P.Type'`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/Render.hs#L46) via [`typeApp`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/Render.hs#L131-L139). We definitely can't add role information here. 
    1. Convert `P.Type'` into `PrettyPrintType` via [`convertPrettyPrintType`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/RenderedCode/RenderType.hs#L156-L157). 
    1. Convert `PrettyPrintType` into a [`RenderedCode` value](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/RenderedCode/Types.hs#L237-L243), which consists of a list of [`RenderedCodeElement` values](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/RenderedCode/Types.hs#L193-L201), via combination of [`renderType'`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/RenderedCode/RenderType.hs#L159-L162), [`matchType`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/RenderedCode/RenderType.hs#L133-L145), [`matchTypeAtom`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/RenderedCode/RenderType.hs#L128-L131), and [`typeLiterals`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/RenderedCode/RenderType.hs#L32-L56).
1. Convert the `RenderedCode` value into its corresponding HTML via the [`convertToHtml` function](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Docs/AsHtml.hs#L192-L217)
    
If we updated [`PrettyPrintType`'s TypeVar](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Pretty/Types.hs#L45) to include a `Maybe P.Role` arg (or something like that) and used `Nothing` for that arg's value by default, then we could add a pass between 2.ii and 2.iii that would insert the corresponding role there. If we're using the subscript idea, then I believe we need to add a new data constructor for `RenderedCodeElement`, so that it renders that content using subscripts.

However, there are multiple uses of `mintersperse sp` throughout this code. We might need to adjust parts of that, so that we don't get a space between the type parameter and the role (e.g. <code>data Map k <sub>nominal</sub> v</code>).

JordanMartinez:
Finally getting back to this. I believe the final HTML and CSS we'll use will look like this (except all one one line rather than multiple lines like the example below). Here's a breakdown of what's happening:
- The [clickable `a` tag](https://github.com/purescript/purescript/issues/4116#issuecomment-868313507) would link to a help page in the documentation repo that explains a bit more about what roles are.
- The `abbr` enables the [not-styleable-but-accessible tooltip](https://github.com/purescript/purescript/issues/4116#issuecomment-864087258) to provide a small amount of context on hover
- the `sub` tag will enable us to [display the role next to its corresponding type parameter](https://github.com/purescript/purescript/issues/4116#issuecomment-864138257) while [making it clear that this is not part of the PureScript syntax](https://github.com/purescript/purescript/issues/4116#issuecomment-864141758)
- the `::after` CSS ensures the role [isn't selectable if one should want to copy/paste the declaration into source code](https://github.com/purescript/purescript/issues/4116#issuecomment-864476706):
```html
<a href="#">
  <abbr title="The 'nominal' role means this type cannot be safely coerced into another type">
    <sub class="role__nominal"></sub>
  </abbr>
</a>
```
```css
.role__nominal::after {
  content: "nominal";
}
.role__phantom::after {
  content: "phantom";
}
```
Below is a screenshot of HalogenM with a few roles added ad-hoc where I'm hovering over the first one. Do we want to style the role annotations differently than what I'm showing below?

role-annotation-example

MonoidMusician:
Looks good to me! I would only suggest for wording: “… this type _argument_ cannot be safely coerced into another type”, so it’s clear we are talking about the arguments, not actual types. Maybe there’s even a clearer way to phrase it to show the relationship of type arguments/roles to coercing the entire type.

rhendric:
Re wording: how about: ‘... means this argument may not change when coercing this type’ and ‘... means this argument can change freely when coercing this type’ for nominal and phantom respectively?

Re style: this is all highly subjective, but I propose:
* using the sans-serif proportional font family used by documentation body text, instead of the fixed-width family used for code
* italicizing the role
* hiding the underline, like with other links

I think maintaining the convention that fixed-width means code is important, and I think the italics help visually clarify the distinction when fixed-width and proportional fonts are mixed.

JordanMartinez:
For wording, I went with this before I saw @rhendric's comment above:
```purescript
describeNominal = fold
          [ "The 'nominal' role means the type represented by this type "
          , "parameter CANNOT be safely coerced to another type when one "
          , "coerces the overall type to another type. For example, given "
          , "`data Foo a` where `a` has the nominal role, the type, "
          , "`Foo Age`, cannot be safely coerced to the type, "
          , "`Foo Int`."
          ]

describePhantom = fold
          [ "The 'phantom' role means the type represented by this type "
          , "parameter CAN be safely coerced to any other type when one "
          , "coerces the overall type to another type. For example, given "
          , "`data Foo a` where `a` has the phantom role, the type, "
          , "`Foo Age`, can be safely coerced to the type, "
          , "`Foo Int`, or even a non-sensical type like, "
          , "`Foo (Foo (Foo String))`."
          ]        
```

> Re style: this is all highly subjective, but I propose:

While it is subjective, I agree with your idea. Part of the reason why I asked was I, too, thought the underline unneeded.

rhendric:
I don't think we need ‘for example’s in the tooltips. They're already links; if a quick reminder isn't sufficient, the user can click.

JordanMartinez:
Given
```purescript
module Main where

data Foo :: Type -> Type -> Type -> Type
data Foo a b c = Foo

type role Foo nominal phantom representational
```
the output of `spago docs -S` produces the following:

role-annotations-added

role-annotations-with-hover

Also, I'm currently linking to [Roles.md](https://github.com/purescript/documentation/blob/master/language/Roles.md). Is that the URL we want to use?

src/Language/PureScript/Docs/AsHtml.hs
```diff
@@ -215,6 +215,33 @@ codeAsHtml r = outputWith elemAsHtml
             linkToDecl ns target mn (withClass class_ (text name))
         NoLink ->
           text name
+    Role role ->
+      case role of
+        "nominal" -> renderRole describeNominal "decl__role_nominal"
+        "phantom" -> renderRole describePhantom "decl__role_phantom"
+
+        -- representational is intentionally not rendered
+        _ -> toHtml ("" :: Text)
```
JordanMartinez, line 224: Should I add an internal error here in case we get something other than `representational`?

src/Language/PureScript/Docs/AsMarkdown.hs
```diff
@@ -67,6 +67,9 @@ codeToString = outputWith elemAsMarkdown
   elemAsMarkdown Space          = " "
   elemAsMarkdown (Symbol _ x _) = x
 
+  -- roles aren't rendered in markdown
+  elemAsMarkdown (Role _) = ""
```
JordanMartinez, line 71: Just want to confirm that this is ok. I think most people don't look at the markdown docs anymore.

src/Language/PureScript/Docs/RenderedCode/RenderType.hs
```diff
@@ -156,6 +158,32 @@ forall_ = mkPattern match
...
+  addRole [] pp = ([], pp)
+  addRole roles@(x:xs) pp = case pp of
+    PPTypeApp f a ->
+      let
+        (remainingRoles1, f') = addRole roles f
+        (remainingRoles2, a') = addRole remainingRoles1 a
+      in (remainingRoles2, PPTypeApp f' a')
+    PPTypeVar t Nothing ->
+      (xs, PPTypeVar t (Just $ displayRole x))
+    other -> (roles, other)
```
JordanMartinez, line 185: AFAICT, this function isn't stack-safe. However, I don't think that issue would ever be triggered.

rhendric:
Oops, sorry, also I think the roles should be normal weight and not bold. (It was much less obvious that they were bold in the other font!)

JordanMartinez:
No worries. Latest screenshot:

role-ann-without-bold

JordanMartinez:
So, this PR is about ready for review. There's a few remaining questions:
- any changes requested on the abbr title content (i.e. what appears when you hover over a nominal/phantom role)?
- the tests currently prove that the roles are added to the Docs.Declaration. However, there's nothing testing that the rendering of those docs is accurate. What do we need to do to test that (e.g. golden test or something)?
- looking at `declName`, there are only four other usages of it. Once in [IDE.State.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Ide/State.hs#L334), twice in `Linter.hs` ([first](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Linter.hs#L37) and [second](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Linter.hs#L167)), and once in [Sugar.Names.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Sugar/Names.hs#L155). Should I refactor `declName` to export `RoleDeclaration` names?

JordanMartinez:
CI started hanging due to an internal error. Getting #4126 in will stop that.

JordanMartinez:
CI passes now.

src/Language/PureScript/Docs/RenderedCode/RenderType.hs
JordanMartinez: This has been fixed in latest commit.

src/Language/PureScript/Docs/AsHtml.hs
JordanMartinez: I've decided to add this error.

JordanMartinez:
Given the below code...

```purescript
module Main where

data DataAll :: Type -> Type -> Type -> Type
data DataAll a b c = DataAll
type role DataAll nominal phantom representational

data TypeOnly_UninterestingKind_InferRole :: Type -> Type -> Type
data TypeOnly_UninterestingKind_InferRole a b

data TypeOnly_InterestingKind_InferRole :: Type -> (Type -> Type) -> Type
data TypeOnly_InterestingKind_InferRole a b

foreign import data FFI_UninterestingKind_InferRole :: Type -> Type -> Type

foreign import data FFI_InterestingKind_InferRole :: forall k. k -> Symbol -> Type

foreign import data FFI_UninterestingKind_ExplicitRole :: Type -> Type -> Type
type role FFI_UninterestingKind_ExplicitRole phantom nominal

foreign import data FFI_InterestingKind_ExplicitRole :: forall k. k -> Symbol -> Type
type role FFI_InterestingKind_ExplicitRole phantom phantom
```

<details>
<summary>the current PR produces the following screenshots</summary>

inferred-roles-1

inferred-roles-2


</details>

JordanMartinez:
I realized I was only inserting the inferred roles for FFI declarations. Latest commit also inserts inferred roles into data declarations.

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -55,8 +63,79 @@ convertModule externs env checkEnv =
...
+    if n == 0
+      then []
+      else map (\i -> ("t" <> T.pack (show i), Nothing)) [0..n]
+    where
+      countParams :: Integer -> Type' -> Integer
+      countParams acc = \case
+        P.ForAll _ _ _ rest _ ->
+          countParams acc rest
+
+        P.TypeApp _ f a@(P.TypeApp _ _ _) | isFunctionApplication f ->
```
kl0tl, line 128: We should not match on type applications here, otherwise we count 0 instead of 1 for `Type -> Type`. The range syntax being inclusive `genTypeParams` would then yield lists of length n+1 for kinds of arity n though! I think we should take from an infinite list instead, which would also simplify away the conditional expression:

```suggestion
   (\(i :: Int) -> ("t" <> T.pack (show i), Nothing)) <$> take n [0..]
    where
      countParams :: Int -> Type' -> Int
      countParams acc = \case
        P.ForAll _ _ _ rest _ ->
          countParams acc rest

        P.TypeApp _ f a | isFunctionApplication f ->
```

src/Language/PureScript/Docs/Convert.hs
```diff
@@ -55,8 +63,79 @@ convertModule externs env checkEnv =
...
+          countParams (acc + 1) a
+
+        _ ->
+          acc
```
kl0tl, line 132: We need to match on `P.ParensInType _ ty` before this wildcard pattern, otherwise we generate only one parameter for foreign data types of kind `Type -> (Type -> Type)`.

src/Language/PureScript/Docs/AsHtml.hs
```diff
@@ -215,6 +215,35 @@ codeAsHtml r = outputWith elemAsHtml
...
+        describeNominal = fold
+          [ "The 'nominal' role means the type represented by this type "
+          , "parameter CANNOT be safely coerced to another type when one "
+          , "coerces the overall type to another type."
+          ]
+        describePhantom = fold
+          [ "The 'phantom' role means the type represented by this type "
+          , "parameter CAN be safely coerced to any other type when one "
+          , "coerces the overall type to another type."
+          ]
```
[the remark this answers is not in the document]
rhendric: (It's the argument that changes though, not the parameter!)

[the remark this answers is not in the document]
JordanMartinez: Perhaps "this argument as represented by the type parameter..."? Though at the end of the day, this exact wording might not matter since the Pursuit docs will clarify the kind signatures further.

src/Language/PureScript/Docs/Convert/Single.hs
```diff
@@ -106,6 +108,28 @@ augmentDeclarations (partitionEithers -> (augments, toplevels)) =
...
         Just $ kcoms <> "\n" <> dcoms
+  augmentWith (AugmentRole comms roles) d =
+    d { declComments = mergeComments (declComments d) comms
+      , declInfo = insertRoles
+      }
+    where
+      mergeComments Nothing rc = rc
+      mergeComments dc Nothing = dc
+      mergeComments (Just dcoms) (Just rcoms) =
+        Just $ dcoms <> "\n" <> rcoms
```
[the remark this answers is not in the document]
JordanMartinez: Nope. I'll fix that.

tests/purs/docs/src/DocCommentsMerge.purs
```diff
@@ -0,0 +1,65 @@
+module DocCommentsMerge where
```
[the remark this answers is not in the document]
JordanMartinez: I guess not a good reason. I feel like the `KindSignatureDocs` properly handles the case that kind signatures are properly accounted for in all of their ways. So, it seemed like it would be better to deduplicate things, especially with #4121 coming into the picture soon.

tests/purs/docs/src/KindSignatureDocs.purs
```diff
@@ -1,109 +1,81 @@
 module KindSignatureDocs where
```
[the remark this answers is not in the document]
JordanMartinez: Not necessarily. I was just testing the doc comment merging code in this module as well. It probably should have been in its own file, but it required less work to do that by having it in `KindSignatureDocs`, too.

src/Language/PureScript/Docs/Convert.hs
JordanMartinez: I ran out of time today, so I'll have to try this out later. I think my code above does work for the case you mentioned, but I think your code is cleaner overall. Thanks for the feedback btw on this PR and the others!

src/Language/PureScript/Docs/Convert.hs
JordanMartinez: Seems like this found a bug:
```purescript
-- Compiler error: the type expects 1 argument but its role declaration
-- lists 3 roles
foreign import data FFI_TailParens :: Type -> (Type -> Type -> Type)
type role FFI_TailParens representational nominal phantom

-- Compiler error: the type expects 0 argument but its role declaration
-- lists 3 roles
foreign import data FFI_WholeParens :: (Type -> Type -> Type -> Type)
type role FFI_WholeParens representational nominal phantom
```
I can only write an FFI declaration in cases where parethesis wraps one of the kinds
```purescript
foreign import data FFI_HeadParens :: (Type) -> Type -> Type -> Type
type role FFI_HeadParens representational nominal phantom
```

JordanMartinez:
Rebased to account for #4121 and added a changelog entry file in latest commit.

src/Language/PureScript/Docs/Convert.hs
kl0tl: The compiler should not trip over parens when checking role declarations arity once https://github.com/purescript/purescript/pull/4157 is merged.

tests/purs/docs/src/DocCommentsMerge.purs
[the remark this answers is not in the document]
kl0tl: The DocCommentsMerge tests don’t check that comments attached to the kind signature of commentless declarations are attached to their declaration nor that comments of declarations with commentless kind signatures are preserved, like the KindSignatureDocs tests used to. I’m not sure those test are redundants, and we should perhaps even add similar ones for declarations with roles and declaration with both.

tests/purs/docs/src/KindSignatureDocs.purs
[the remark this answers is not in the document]
kl0tl: Testing kind signatures and doc comments separately indeed seems to be better now that we also test documented role annotations.

kl0tl:
Synonyms are not supported in foreign data types kinds: we render

Capture d’écran 2021-07-16 à 18 18 17

for

```purescript
foreign import data FFI1 :: (Function Type) Type
```

but

Capture d’écran 2021-07-16 à 18 18 29

for

```purs
type To = Function
foreign import data FFI1 :: (To Type) Type
```

This is perhaps too uncommon to be dealt with in this pull request but we could do so by building a SynonymMap, like we do when desugaring instances, and then call Language.PureScript.TypeChecker.Synonyms.replaceAllTypeSynonymsM in convertFFIDecl.

JordanMartinez:
@kl0tl The "sort annotations are only dropped for inferred kind signatures" issue you described [here](https://github.com/purescript/purescript/pull/4137#issuecomment-881575727) has been addressed in latest commits here. Not sure whether it's sensible to add it in this PR. However, the advantage of doing so here means I can add FFI declaration tests here, too.

JordanMartinez:
> This is perhaps too uncommon to be dealt with in this pull request but we could do so by building a SynonymMap, like we do when desugaring instances, and then call Language.PureScript.TypeChecker.Synonyms.replaceAllTypeSynonymsM in convertFFIDecl.

Yeah... that sounds like it should be in its own PR.

tests/purs/docs/src/DocCommentsMerge.purs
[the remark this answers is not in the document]
JordanMartinez: This has been addressed in latest commits

src/Language/PureScript/Docs/AsMarkdown.hs
JordanMartinez: Added in latest commit.

tests/TestDocs.hs
```diff
@@ -803,27 +854,53 @@ testCases =
...
+      , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_RNP" [P.Representational, P.Nominal, P.Phantom]
+
+      , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_Higher1" [P.Representational, P.Nominal, P.Phantom]
+      , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_Higher2" [P.Representational, P.Nominal, P.Phantom]
+      , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_Higher3" [P.Representational, P.Nominal, P.Phantom]
+      , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_Higher4" [P.Representational, P.Nominal, P.Phantom]
+
+      , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_HeadParens" [P.Representational, P.Nominal, P.Phantom]
+      -- , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_TailParens" [P.Representational, P.Nominal, P.Phantom]
+      -- , ShouldHaveRoleAnnotation (n "RoleAnnotationDocs") "FFI_WholeParens" [P.Representational, P.Nominal, P.Phantom]
```
JordanMartinez, line 873: Pending #4157

JordanMartinez:
With #4160 being resolved, is there any other work to do here besides waiting for #4157 to get merged and then adding additional tests?

JordanMartinez:
@kl0tl I generated markdown docs to see whether the role annotations are rendered correctly. Looks like we can't use the HTML tags where I put them because the entire declaration is wrapped in a code block. 

Given
```purescript
module Main where


data Foo a = Foo a
type role Foo nominal
```
we get (horizontal lines added to clearly indicate start/finish of docs):

---

## Module Main

#### `Foo`

``` purescript
data Foo a<sub title="The 'nominal' role means this argument may not change when coercing the type."><em><a href="https://github.com/purescript/documentation/blob/master/language/Roles.md">nominal</a></em></sub>
  = Foo a
```

---

kl0tl (commented):
> Looks like we can't use the HTML tags where I put them because the entire declaration is wrapped in a code block.

Oh, let’s not bother with roles in Markdown docs then.

> …, is there any other work to do here besides waiting for #4157 to get merged and then adding additional tests?

It looks quite good to me modulo the Markdown docs, sorry about that!

JordanMartinez:
The markdown docs commit has been reverted and I fixed the typo you found.

JordanMartinez:
CI passes.

JordanMartinez:
If no one has any further feedback on this, I'm going to merge this in two days.

src/Language/PureScript/Docs/AsMarkdown.hs
JordanMartinez: No. Originally, I did, but then see [this comment](https://github.com/purescript/purescript/pull/4121#issuecomment-885409460) for why we didn't move forward with that idea. The last few commits in this PR reverted the commits that added this.

rhendric: Ah, thanks!

rhendric (approved):

PR #4123, opened 2021-06-27
RELEASE_GUIDE.md: s/Or alternatively/Alternatively/

varikvalefor:
"Alternatively" replaces "[o]r alternatively"; "or alternatively" is redundant.

### DIFF
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -43,7 +43,7 @@ projects whom we should probably at least notify. See below subsections:
 
 ### Libraries
 
-Are there breaking changes to the language? Or alternatively, are there
+Are there breaking changes to the language? Alternatively, are there
 language changes which require breaking changes in the relevant libraries to
 make use of? If so:
 


rhendric (approved):

JordanMartinez:
Are you already listed in `CONTRIBUTORS.md`?

JordanMartinez (approved):

varikvalefor:
> Are you already listed in `CONTRIBUTORS.md`?

Negative.

rhendric:
I think we only require a CONTRIBUTORS.md entry for code changes—at least, that's what CONTRIBUTING.md says.

rhendric:
Thanks @varikvalefor!

hdgarrood:
I think we should require a CONTRIBUTORS.md addition only in cases where the contribution is potentially copyrightable, which doesn’t necessarily just mean code changes. In any case a typo fix surely isn’t in that category, so I think not updating it here makes sense.

Issue #4086, opened 2021-05-08
Make typeclass and deriving desugaring part of type checking


JordanMartinez:
## Summary

Make certain kind information available sooner in the compilation pipeline, such that other desirable features become unblocked or easier to do.

## Motivation

See [kl0tl's comment and natefaubion's following comment in #3426](https://github.com/purescript/purescript/issues/3426#issuecomment-651082718). Fixing this would unblock a few things:
- #3426
- #4084 
- potentially other things I'm not aware of

Since this is far outside my expertise, I'm not able to provide a concrete proposal or example. However, this at least provides a place to discuss this further.

rhendric:
I did some digging into #4101 and #4105 and I believe they both boil down to this. #4105 can be narrowly fixed without this, but I'm not sure it's worth it—we should probably just roll up our sleeves and do it right.

kl0tl:
This would also fix #3510 and #3453. I’m focusing on ES modules when I find time for PureScript these days but I believe I’ve done a good chunk of the work in https://github.com/kl0tl/purescript/commit/9e94eddc2f7ccf5ce89ba7b932034432ff44eae7. If I recall I stopped at the desugaring of binding groups: since type class instances are not desugared before binding groups anymore their declarations now depends on the bindings appearing in their members, so we may need another type of binding group declaration which can weave together both types and terms.

PR #4125, opened 2021-06-28
Represent class dictionaries as newtypes

rhendric:
Class dictionaries are now represented as newtype wrappers of records.
This drops the awkward TypeClassDictionaryConstructorApp and
TypeClassDictionaryAccessor expressions in favor of regular newtype
constructors and record accessors.

**Description of the change**

This is mostly motivated by an attempt-in-progress at #4086, but I think it stands on its own merits.

Previously: #3567 and #781. I'm tempted to say that this PR closes those, but maybe there are reasons to prefer the data representation over the newtype one I've implemented? I quite like how simple the generated code for the newtype representation is, and how it means that the newtype constructors can be completely erased (with other newtypes, the constructors still have to be exported in case some code references them in a way that isn't an erased application, but with typeclass newtypes that's not possible). But adapting this PR to use data instead of newtype would be easy and would also suffice for my #4086 purposes.

This newly-updated comment describes the details:

https://github.com/purescript/purescript/blob/cd869ebfd5745d2ab2223d8ddce147061dc5aa7b/src/Language/PureScript/Sugar/TypeClasses.hs#L109-L198

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/internal_typeclasses-are-newtypes.md
+++ b/CHANGELOG.d/internal_typeclasses-are-newtypes.md
@@ -1,0 +1,1 @@
+* Represent class dictionaries as newtypes
--- a/lib/purescript-cst/src/Language/PureScript/AST/Declarations.hs
+++ b/lib/purescript-cst/src/Language/PureScript/AST/Declarations.hs
@@ -719,11 +719,6 @@ data Expr
   --
   | Ado (Maybe ModuleName) [DoNotationElement] Expr
   -- |
-  -- An application of a typeclass dictionary constructor. The value should be
-  -- an ObjectLiteral.
-  --
-  | TypeClassDictionaryConstructorApp (Qualified (ProperName 'ClassName)) Expr
-  -- |
   -- A placeholder for a type class dictionary to be inserted later. At the end of type checking, these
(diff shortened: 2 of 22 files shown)


garyb:
I wanted to use data constructors as their implementation is more optimisable by JS runtimes*, and it makes the representation less record-dependent for backends that have to implement records by means that may be less efficient than static data types. 

I was also thinking that perhaps it'd be easier to use annotations on the data constructor form of this to use for later optimisations, but you've already made good progress on a major part of that without needing it (#3915)!

\* In light of some benchmarks Nate tried recently, there probably isn't much of a distinction between object and "class" performance anymore, so the JS concern isn't really there anymore. Property access used to be significantly faster on "classes" than on standard objects.

I'm in favour of this PR - even if later it seems that the data approach is better again for some reason, for we still get to eliminate the `TypeClassDictionary...` stuff in the mean time, which is totally worth it!

JordanMartinez:
@garyb So, would you be willing to approve this PR? Or are there other things you think need to be addressed?

garyb:
I haven't reviewed it thoroughly yet, but it looks good from a glance! Will try to remember to come back to it this weekend for a proper :+1:

garyb (approved):
:+1: Sorry it took ages to get to!

PR #4129, opened 2021-07-02
Revert "Improve apartness checking"

rhendric:
Reverts purescript/purescript#4064


JordanMartinez:
I'm checking this out locally and will compile against latest package set as a sanity check.

JordanMartinez (approved):
Package set compiles without problem.

PR #4132, opened 2021-07-03
Move unreleased changelog entries to CHANGELOG.d

### LINKED ISSUE #4130: Our changelog strategy is a pit trap
Our current approach to keeping a changelog is to add entries in an ‘Unreleased’ section, and then rename that section when there is a release. This means that any PR that spans a release needs to be updated every time there is a release, lest it be merged into the renamed section instead of the current ‘Unreleased’ section. There's no check to make sure that such a merge doesn't happen other than the vigilance of maintainers, and it's extra work for PR authors for the PRs that span at least one release. Personally, I think it's frustrating enough just staving off bitrot for PRs that are over a year old; moving the changelog entry around is simpler but a little more demoralizing because it's a self-created problem.

My ideal, as I've said before, would be to generate the changelog from commit messages (by strictly following some [commit message convention](https://www.conventionalcommits.org/en/v1.0.0/)), but the feedback on that, IIRC, was that we want to write our changelogs for a different audience than we want to write our commit messages for, and while I still don't agree I can respect that I'm in the minority. But that aside, there are surely other things we can do to reduce the maintenance burden on long PRs.

For example, we could keep a `CHANGELOG.md` for released changes, and a `CHANGELOG.unreleased.md` for unreleased changes, and prepend the latter into the former on release. That would keep PRs from potentially merging into the wrong part of the changelog.

Better still, we could do the above but instead of one `CHANGELOG.unreleased.md`, we could keep a folder full of snippets, one for each PR, named by category and PR number (`feature-1234.md`, `fix-1235.md`, `breaking-1236.md`). Then, not only would the above problem be solved, but we would also solve the related problem that PRs will conflict with each other if they both want to append to the same place in the changelog.

I recognize that the manual changelog was introduced in part to reduce the burden during releases; I don't think these proposals move the needle too far back in the other direction—I'll volunteer to write the script to merge changelog fragments into the released changelog, if that helps—but what do others think?

rhendric:
**Description of the change**

Instead of editing `CHANGELOG.md`, let's create files in `CHANGELOG.d`. `CHANGELOG.d/README.md` explains the scheme, and `update-changelog.hs` does the hard work at release time.

If you'd like to play around with the release-time ergonomics, get a test branch with `git clone -b rhendric/test-for-changelog-dir --single-branch https://github.com/rhendric/purescript.git`, run `./update-changelog.hs`, and see what happened with `git diff --cached`.

Closes #4130.

---

**Checklist:**

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,7 @@ Clearly and concisely describe the purpose of the pull request. If this PR relat
 
 **Checklist:**
 
-- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
+- [ ] Added a file to CHANGELOG.d for this PR (see CHANGELOG.d/README.md)
 - [ ] Added myself to CONTRIBUTORS.md (if this is my first contribution)
 - [ ] Linked any existing issues or proposals that this pull request should close
 - [ ] Updated or added relevant documentation
--- a/CHANGELOG.d/README.md
+++ b/CHANGELOG.d/README.md
@@ -1,0 +1,50 @@
+This directory contains changelog entries for work that has not yet been
+released. When a release goes out, these files will be concatenated and
+prepended to CHANGELOG.md in a new section for that release.
+
+Maintainers: see update-changelog.hs for details of this process.
+
+Contributors: read on!
+
+When you are preparing a new PR, add a new file to this directory. The file
+should be named `{PREFIX}_{SLUG}.md`, where `{PREFIX}` is one of the following:
+* `breaking`: for breaking changes
+* `feature`: for new features
+* `fix`: for bug fixes
+* `internal`: for work that will not directly affect users of PureScript
+* `misc`: for anything else that needs to be logged
+
+`{SLUG}` should be a short description of the work you've done. The name has no
+impact on the final CHANGELOG.md.
+
+Some example names:
+* `fix_issue-9876.md`
+* `breaking_deprecate-classes.md`
+* `misc_add-forum-to-readme.md`
+
+The contents of the file can be as brief as:
+
+```markdown
+* A short message, like the title of your commit
+```
+
+Please remember the initial `*`! These files will all be concatenated into
+lists.
+
+If you have more to say about your work, indent additional lines like so:
+
+``````markdown
+* A short message, like the title of your commit
+
+  Here is a longer explanation of what this is all about. Of course, this file
+  is Markdown, so feel free to use *formatting*
+
+  ```
+  and code blocks
+  ```
+
+  if it makes your work more understandable.
+``````
+
+You do not have to edit your changelog file to include a reference to your PR.
+The CHANGELOG.md updating script will do this automatically and credit you.
--- a/CHANGELOG.d/internal_changelog-dir.md
+++ b/CHANGELOG.d/internal_changelog-dir.md
@@ -1,0 +1,3 @@
+* Move unreleased changelog entries to CHANGELOG.d
+
+  See CHANGELOG.d/README.md for details.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,6 @@
 
 Notable changes to this project are documented in this file. The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-
-Breaking changes:
-
-New features:
-
-Bugfixes:
-
-Internal:
-
 ## v0.14.3
 
 New features:
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -96,6 +96,9 @@ considering what effects this may have:
 
       - The version bounds for `purescript-cst` in `purescript.cabal`
 
+- Run `stack update-changelog.hs`, which will move the entries in `CHANGELOG.d`
+  to a new section in `CHANGELOG.md` labeled with the new version.
+
 - Create a release from the releases tab in GitHub and copy in the release
   notes. This will also create a tag, which will kick off a CI build, which
   will upload prebuilt compiler binaries to the release on GitHub when it
--- a/update-changelog.hs
+++ b/update-changelog.hs
@@ -1,0 +1,196 @@
+#!/usr/bin/env stack
+-- stack --resolver lts-17.6 script
+{-# LANGUAGE
+    DeriveFoldable
+  , DeriveFunctor
+  , DeriveTraversable
+  , FlexibleContexts
+  , LambdaCase
+  , NoImplicitPrelude
+  , OverloadedStrings
+  , PackageImports
+  , RecordWildCards
+  , TupleSections
+  , ViewPatterns
+#-}
+-- |
+-- This script updates CHANGELOG.md with the contents of CHANGELOG.d, and
+-- empties CHANGELOG.d. It takes care of:
+--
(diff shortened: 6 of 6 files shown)


JordanMartinez:
Just a few thoughts
- does this script also add a new `[Unreleased]` section? It looks like it doesn't, but I might have missed it
- while the automatic PR number and author finder is nice, part of me still thinks that should be done manually. When a feature spans multiple PRs, we would want to edit the same entry multiple times as each part of it is merged (e.g. kind signatures in docs) rather than having multiple entries. When such entries are edited, then this code would include that author as well despite them not having contributed to the code, right?

Other than that, it looks good.

rhendric:
>     * does this script also add a new `[Unreleased]` section?

No, the whole idea is to get rid of that! This PR can't actually delete that section yet because I didn't want to mess with your release PR, but the idea is that `CHANGELOG.md` contains released changelog entries, and `CHANGELOG.d` contains unreleased entries.

>     * When a feature spans multiple PRs, we would want to edit the same entry multiple times as each part of it is merged (e.g. kind signatures in docs) rather than having multiple entries. When such entries are edited, then this code would include that author as well despite them not having contributed to the code, right?

Yeah, so this is a little complex. The design (which could be imperfectly implemented) is as follows:
* All commits that touch an entry file, or are a merge commit in the path from HEAD to such a commit, are considered relevant to that entry.
* Any relevant commits that are not GitHub squash or merge commits are then discarded.
* Of what remains, any squash commits that *only* touch changelog entries are discarded. (This is intended to filter out commits that reword a previous entry but don't contribute any code. This check also includes `CHANGELOG.md` itself, in case a larger reword is happening that encompasses both past and present changelog entries.)
* PR numbers are collected from the remaining relevant commits, and their authors are resolved and nubbed.

So if a feature spans multiple PRs, any PR that edits the same entry file and also touches some non-changelog file in the repo gets added to the list for that entry. Any PR that only edits entry files is not mentioned.

JordanMartinez:
> No, the whole idea is to get rid of that! 

:man_facepalming: Ha! I'm dumb. 

Well great! I like this idea a lot! Well done.

rhendric:
Updated to actually remove the Unreleased section from the changelog, now that it's empty.

rhendric:
I try not to be pushy about my PRs, but I think it would be particularly convenient for this to be reviewed and, if acceptable, merged before any other PRs, now that 0.14.3 is released and there aren't any unreleased changelog entries.

JordanMartinez:
I'm in full agreement of getting this merged first before any other PRs.

JordanMartinez:
I checked this out locally to see how this works. @rhendric Thanks for making me laugh. The changelog was a fun read.

Running `stack update-changelog.hs` took a few minutes to build initially. AFAICT, this long initial build is the only downside of this script. What if a developer wanted to verify that their entry would display correctly in the resulting changelog? Running this to verify that it works as expected wouldn't be a "few milliseconds" tests by running that command. 

Once it did build, rerunning it took about 4 seconds. 

When I did build the changelog via the above command (after updating the version in `npm-package/json`), the CHANGELOG.md file was updated to this. I probably don't need to say this, but this is the Internet: the below summaries, linked PRs, and their supposed authors are all fake data:
```markdown
## 0.14.4

New features:

* Support cobras to control anteater population (#4123 and #4100 by @varikvalefor and @JordanMartinez)

  The cobras are themselves controlled by a mongoose.

* Add a doberman to guard the big pit (#4129 by @rhendric)

Bugfixes:

* Use anteaters to clean up rooms (#4096 by @rhendric)

Internal:

* Move unreleased changelog entries to CHANGELOG.d

  See CHANGELOG.d/README.md for details.

* Use a big pit to hold all these animals (#4113 by @f-f)
```

At some point, as I was playing around with it, I modified one of the files in the CHANGELOG.d folder but didn't commit it. Running `stack update-changelog.hs` produced this error:
```
error: the following file has local modifications:
    CHANGELOG.d/fix_stuff.md
(use --cached to keep the file, or -f to force removal)
update-changelog.hs: git rm -q CHANGELOG.d/feature_cobras.md CHANGELOG.d/feature_doberman.md CHANGELOG.d/fix_anteaters.md CHANGELOG.d/fix_stuff.md CHANGELOG.d/internal_changelog-dir.md CHANGELOG.d/internal_big-pit.md failed with exit code 1
```

I also intentionally created a file with a lot of blank lines at the end of it to see what would happen. Why? My editor always adds a blank newline to the bottom of a file, whether I need it or not:
```
* fix some stuff



  and some more stuff


  
```
This changelog script did not trim the three extra lines at the bottom of the file. Below is a snippet of what it produced:
```markdown
Bugfixes:

* Use anteaters to clean up rooms (#4096 by @rhendric)

* fix some stuff



  and some more stuff


  

Internal:
```

The same extra newlines issue occurs if there are extra newlines above the file's content. For example:
<pre><code>


* fix some stuff

</code></pre>

Perhaps this line trimming issue should be addressed, but since we can address it when reviewing PRs, I don't think it's absolutely necessary.

If anything, @rhendric, do you think there should be an `example_how to add an entry.md` file to which a new contributor could refer in case they are contributing their first PR to this repo when the `CHANGELOG.d` folder is empty because we just had a new release?

rhendric:
> Running this to verify that it works as expected wouldn't be a "few milliseconds" tests by running that command.

Sad but true. How often do you think contributors will really want to do this, though? The processing done to the entries is pretty minimal; I expect once the novelty wears off, this script will get run about as often as the license generator, which has a similar startup cost.

> I modified one of the files in the CHANGELOG.d folder but didn't commit it.

That's a good test. Failing at this step—with CHANGELOG.md updated and `git add`ed but the CHANGELOG.d files not yet deleted—is a little sloppy. Do you think it should force delete, so that whatever was written to the entries locally gets added to the changelog as if it had been committed? Or should it check for local modifications before writing to CHANGELOG.md, and ask the user to commit or revert them first if there are any? Or something else?

> Perhaps this line trimming issue should be addressed

I think that's a good idea. I didn't think it mattered much because in Markdown, many line breaks are the same as one, but if you cared enough to look for it, I can care enough to make it correct.

> do you think there should be an `example_how to add an entry.md` file

Doesn't `CHANGELOG.d/README.md` do that well enough?

JordanMartinez:
> Doesn't CHANGELOG.d/README.md do that well enough?

I read it again. Yeah, It does. Never mind.

> Sad but true. How often do you think contributors will really want to do this, though?

Probably not that often. Again, I think this is the only downside, but it's a very small downside in comparison to the other advantages this provides.

> I didn't think it mattered much because in Markdown, many line breaks are the same as one, but if you cared enough to look for it, I can care enough to make it correct.

I don't think it needs to be changed until we experience this problem enough to warrant the change. Let's keep this tool simple and see how things go.

> That's a good test. Failing at this step—with CHANGELOG.md updated and git added but the CHANGELOG.d files not yet deleted—is a little sloppy. Do you think it should force delete, so that whatever was written to the entries locally gets added to the changelog as if it had been committed? Or should it check for local modifications before writing to CHANGELOG.md, and ask the user to commit or revert them first if there are any? Or something else?

Good questions. I think the force-delete would be an unexpected change if I was actually trying to change one of the entries. I think it would be better to force the user to commit them, so that it appears in the diff when we're reviewing the release PR.

JordanMartinez (approved):

rhendric:
@JordanMartinez I had already implemented stripping extra whitespace by the time I read your comment; it's pretty simple.

Feedback is addressed in https://github.com/purescript/purescript/pull/4132/commits/9865dd80796e17c22e7988926a98d993295d7576, and my test branch has been rebased atop this one. If you could give this all one final look-over before I merge?

thomashoneyman (approved):

JordanMartinez:
@rhendric Thanks for doing that. Yeah, this looks good. Feel free to merge!

PR #4134, opened 2021-07-05
update Stack in Github Actions

peterbecich:
**Description of the change**

Update Stack from 2.5.1 to 2.7.1 in Github Actions.

---

**Checklist:**


rhendric:
1. Can anyone think of a reason that this might get us in trouble, like updating our build machine to `ubuntu-latest` did?
2. Even if we can't, should we be purely reactive about updating anything in our build chain out of an abundance of caution?

I *think* bumping our Stack version is harmless—it's the resolver that makes a difference for Haskell dependencies, and the build machine that makes a difference for linked libraries—but I thought that about `ubuntu-latest` too, so....

f-f:
Upgrading the Stack version is generally very safe, and easy to revert if it causes trouble.

f-f (approved):

hdgarrood:
That’s actually not my experience, stack has had bugs that have been quite irritating before; for example there was one when upgrading from v1 to v2 which basically broke caching completely, causing significantly longer build times.

hdgarrood:
It is true that it’s easy to revert though.

JordanMartinez:
Just FYI. I'm using 2.7.1 locally, and it's what I was using to investigate issues with the recent CI issues in #4138.
```bash
$ stack --version
Version 2.7.1, Git revision 8afe0c2932716b0441cf4440d6942c59568b6b19 x86_64 hpack-0.34.4
```

JordanMartinez (approved):

JordanMartinez:
I fixed the merge conflict, which involves updating the version for the main jobs and also the `lint` job.

PR #4137, opened 2021-07-06
Omit kind sigs where kind is uninteresting, even when parens are involved

### LINKED ISSUE #4136: Still omit uninteresting kind signatures from HTML docs when redundant parenthesis are involved
## Description

in #4100 and #4119, we display kind signatures to HTML docs, but only if those kind signatures are "interesting." In other words, "uninteresting" kinds are removed. While working on #4121, I discovered that parenthesis are not properly accounted for when checking if a kind signature is "uninteresting."

## To Reproduce

Steps to reproduce the behavior.
```bash
mkdir uninterestingKinds
cd uninterestingKinds
cat << EOF > src/Main.purs
module Main where

data Oops :: (Type) -> Type
data Oops a = Oops
EOF
purs docs "src/**/*.purs"
```

oops-parents-issue

## Expected behavior

If a kind signature is still considered uninteresting once redundant parenthesis are removed, then the kind signature should be omitted.

## PureScript version

`master`

JordanMartinez:
**Description of the change**

Fixes #4136

---

**Checklist:**


src/Language/PureScript/Docs/Convert.hs
```diff
@@ -108,17 +108,30 @@ insertValueTypesAndAdjustKinds env m =
   -- - `Constraint` (class declaration only)
   -- - `Type -> K` where `K` is an "uninteresting" kind
   isUninteresting
-    :: P.KindSignatureFor -> P.Type () -> Bool
+    :: P.KindSignatureFor -> Type' -> Bool
   isUninteresting keyword = \case
-    P.TypeApp _ t1 t2 | t1 == kindFunctionType ->
+    -- `Type -> ...`
+    P.TypeApp _ (P.TypeApp _ (P.TypeConstructor () Prim.Function) ty) t2 | isKindPrimType ty ->
```
kl0tl, line 114: We need to account for the case where an application is wrapped in parens, like `(Type -> Type)`, otherwise we deem such kinds interesting and they show up in the generated docs.

kl0tl (commented):
Should we also drop sorts annotations from developper written kinds?

src/Language/PureScript/Docs/Convert.hs
JordanMartinez: Meaning if we have something like this?
```purescript
data Foo :: (Type -> Type)
data Foo a = Foo a
```
which is the same as writing the uninteresting kind signature below:
```purescript
data Foo :: Type -> Type
data Foo a = Foo a
```

kl0tl: Yeah exactly, `data Foo :: (Type -> Type)` should be treated like `data Foo :: Type -> Type` but the kind is currently documented.

JordanMartinez: Wouldn't that mean accounting for various other situations, too? I'm starting to wonder if this is really worth doing and if we should instead say, "Yeah, just don't do that."

As examples of other things:
```purescript
-- or any uninteresting kind where the whole thing is wrapped in parens
data Foo1 :: (Type -> Type -> Type)
data Foo1 a b = Foo1 a b


-- or any uninteresting kind where the the tail part is wrapped in parens
data Foo2 :: Type -> (Type -> Type)
data Foo2 a b = Foo2 a b
```

kl0tl: Yes those kinds should be uninteresting too. It should only be a matter of adding a  `P.ParensInType _ ty -> isUninteresting ty` case to `isUninteresting` though.

JordanMartinez:
Note: while I was working on this PR, I also learned that the parens situation will be displayed in the kind signature if it's included in an interesting kind signature. For example:
```purescript
data Foo :: forall k. k -> (Type)
data Foo a = Foo
```
will be rendered as the above, not as below:
:
```purescript
data Foo :: forall k. k -> Type
data Foo a = Foo
```
I'm not sure whether it's really worth it to remove the parenthesis because I doubt this will happen in practice.

src/Language/PureScript/Docs/Convert.hs
JordanMartinez: Done and added tests for it. I haven't tested this locally.

JordanMartinez:
> Should we also drop sorts annotations from developper written kinds?

Meaning, if the developer writes
```purescript
data Foo (a :: Type) = Foo a
```
then we render the following?
```purescript
data Foo a = Foo a
```

JordanMartinez:
Just added a test for this. It passed but nothing changed in my code:
```
data DataSortAnnotation (a :: Type) = DataSortAnnotation a
```

rhendric:
> ```
> data DataSortAnnotation (a :: Type) = DataSortAnnotation a
> ```

That's still a kind annotation. I think a sort annotation would be the `:: Type` in
```purs
foreign import data DataSortAnnotation :: forall (k :: Type). k -> Type
```

JordanMartinez:
Mm... I'm not so sure. [Harry said something about this](https://github.com/purescript/purescript/pull/4100#issuecomment-864546458) that makes me think otherwise...?

JordanMartinez:
Changelog entry has been added now that #4132 is in.

rhendric:
The changelog file name should start with `fix`, not `bug`.

src/Language/PureScript/Docs/Convert.hs
rhendric: Does this cover `data Foo :: (->) Type Type`? Or `data Foo :: ((->) Type) Type`, or, for that matter, `data Foo :: ((((->)) Type)) Type`?

These are profoundly obnoxious edge cases, but if they cause problems, maybe the thing to do is just strip out all parens using a traversal before calling `isUninteresting`.

JordanMartinez: Good question. I'll add a test for that.

JordanMartinez: Latest commit adds these three tests, the latter two of which currently fail.

JordanMartinez: Yeah, stripping those parenthesis out and then solving the problem with the original approach was much easier than trying to account for them at the same time.

JordanMartinez:
This PR should be ready for another review when someone has a moment.

kl0tl (approved):
I still think we should render

```purescript
foreign import data DataSortAnnotation :: forall (k :: Type). k -> Type
```

as

```
foreign import data DataSortAnnotation :: forall k. k -> Type
```

for consistency with inferred kinds, but that’s not really related to this pull request so it can happen in another one (and perhaps it should rather be done by a formatter). Ditto for extraneous parens in interesting kinds.

This looks great though, thank you!

JordanMartinez:
> I still think we should render

Isn't that handled in #4121? Since the declaration is FFI, #4121 first converts that to a data declaration and then runs it through the code that removes the sort annotation.

Regardless, thanks for the review. Merging!

kl0tl:
> Isn't that handled in #4121? Since the declaration is FFI, #4121 first converts that to a data declaration and then runs it through the code that removes the sort annotation.

This doesn’t seem to be the case: on 73ff5f6ca95e2c31fa4ab01db5a5cee44ca9b78e we render

Capture d’écran 2021-07-16 à 18 33 54

for

```purescript
foreign import data DataSortAnnotation :: forall (k :: Type). k -> Type
```

because we only drop sort annotations for inferred kinds.

JordanMartinez:
Gotcha. I've reported this as a bug in #4160

PR #4139, opened 2021-07-06
Refer to binary using local-doc-root workaround

### LINKED ISSUE #4138: CI doesn't currently build binaries
See [the ubuntu build](https://github.com/purescript/purescript/runs/3003232309?check_suite_focus=true), which fails because the binary file that should exist at `sdist-test/.stack-work/install/x86_64-linux/0c876ca1dfc6728d04f0d922e905e8ecaa12cec66470d064293c7f7c2a261f0b/8.10.4/bin/purs` does not exist.

It seems #4072 caused an issue by introducing the `--hadock` flag.

JordanMartinez:
**Description of the change**

Fixes #4138.

Since v0.14.3 was already tagged, how do you want to handle this situation?

---

**Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,7 @@ Bugfixes:
 
 Internal:
 
-* Fix for Haddock (#4072 by @ncaq and @JordanMartinez)
+* Fix for Haddock (#4072 by @ncaq and @JordanMartinez, #4139 by @JordanMartinez)
 
 * Update RELEASE_GUIDE.md with more details (#4104 by @JordanMartinez)
 
--- a/bundle/build.sh
+++ b/bundle/build.sh
@@ -26,7 +26,7 @@ then
 else
   BIN="purs"
 fi
-FULL_BIN="$(stack path --local-install-root)/bin/$BIN"
+FULL_BIN="$(stack path --local-doc-root)/../bin/$BIN"
 if [ "$OS" != "win64" ]
 then
   strip "$FULL_BIN"


thomashoneyman:
It was tagged so recently (and without binaries) that I think we can delete the tag and tag a new release.

JordanMartinez:
The release and tag have both been deleted

JordanMartinez:
CI builds. Can I get an approval, so I can get this merged?

hdgarrood:
I think this is fine for now, especially since the old release has now been deleted, but just for future reference - we don’t know what kind of automation people might have with tags and releases and repo mirrors and so on, so in the future I would prefer to err on the side of not modifying existing tags. In this case I would have created a new v0.14.4 tag instead, and a small note in the changelog that v0.14.3 isn’t a real release.

hdgarrood:
Does using `--haddock` affect the size of the resulting binary? I’m now wondering if the binary we get will include docs somehow where it didn’t previously, causing the binary to be larger. I can’t see why else it would end up in a separate directory. If so I think the correct fix may be to avoid `--haddock` entirely, at least for release builds.

JordanMartinez:
> we don’t know what kind of automation people might have with tags and releases and repo mirrors and so on, so in the future I would prefer to err on the side of not modifying existing tags. In this case I would have created a new v0.14.4 tag instead, and a small note in the changelog that v0.14.3 isn’t a real release.

Makes sense. Mind if I add that to the RELEASE_GUIDE.md in case this should occur again?

> Does using --haddock affect the size of the resulting binary?

I'm looking into that right now.

hdgarrood:
Adding this to RELEASE_GUIDE.md sounds like a great idea 👍

JordanMartinez:
```bash
# v0.14.2
ls -ls purs.bin 
48408 -rwxr-xr-x 1 jordan jordan 49561488 Jun  5 12:18 purs.bin

# binary built from this PR
$ ls -ls purs 
48500 -rwxr-xr-x 1 jordan jordan 49650640 Jul  6 17:55 purs
```
| Comparision |
| -: |
| 49,561,488 |
| 49,650,640 |

The difference is 89,152

hdgarrood:
I think we need to be comparing the old CI script with the new one when run against the same code. What do you get if you revert only the changes in ci/build.sh but leave everything else alone?

hdgarrood:
If the difference turns out to less than 100KB though then I think that’s safe to ignore.

JordanMartinez:
```
$ ls -ls purs 
48496 -rwxr-xr-x 1 jordan jordan 49650640 Jul  6 18:39 purs
```
Seems to be the same binary.

thomashoneyman (approved):

hdgarrood (approved):

hdgarrood:
Thanks for checking!

PR #4144, opened 2021-07-11
purescript-cst: allow dlist-1.0

juhp:


hdgarrood:
I think relaxing purescript-cst bounds beyond what the main purescript.cabal bounds say is fine, and makes sense since the intention is that people can depend on purescript-cst as a library, but if we are going to start doing that I would prefer that we set up a cabal build for purescript-cst at the same time so that we are at least testing that the library still builds with those relaxed bounds.

JordanMartinez:
Closing as `dlist` was updated to `1.0` in #4199

PR #4145, opened 2021-07-11
bump some upperbounds

juhp:
build tested with ghc-8.10.5/lts18 based packages


juhp:
Is it okay to relax the language-javascript?

haskeline is also strict: it builds fine for me with the haskeline in ghc-8.10.

hdgarrood:
It is not, unfortunately. Actually we prefer not to relax upper bounds, in order to ensure that stack and cabal choose similar build plans. See the comment at the top of the build-depends section.

juhp:
Okay thank you, I had missed that detailed comment

PR #4148, opened 2021-07-12
Run Weeder in CI and make it happy

### LINKED ISSUE #4146: Using Weeder; or, what is our public API as a library?
I'd like to add [Weeder](https://hackage.haskell.org/package/weeder) to our CI workflow, to help us clean up functions that are exported from a module but never used anywhere in the project.

The problem is, anything exported from any module in either `purescript` or `purescript-cst` technically is part of our public API, right? So deleting these definitions would be a breaking change. But I imagine there's a line we could draw around some of our modules and call those our actual public API, and everything else internal. Where would we draw it?

Aside from that question, there's the smaller question of whether we'd want to keep some internal unused functions around anyway, because they complete a set or might be expected to be useful in the future. I'm inclined towards no in most cases, but, for example, maybe we want to keep `runSupply` alongside `runSupplyT`, even though we only ever use the latter.

For reference, after defining `Main.main` and a handful of tool-generated functions as entry points, [here](https://gist.github.com/rhendric/feaa460453459bf524e341d7ebb4b753) are the 187 unused definitions Weeder found. This wouldn't include any functions that would become unused if these are deleted, of course, and also I haven't verified that there are no false positives in there.

rhendric:
Weeder runs a whole-program analysis and will fail the build if there
are any functions not reachable from our list of roots, which includes
the purs executable and a few API points used externally.

This commit removes all such functions, and in a few cases removes
primes from functions that had their un-primed counterparts thus
removed.

**Description of the change**

Closes #4146.

---

**Checklist:**


.github/workflows/ci.yml
```diff
@@ -53,10 +55,6 @@ jobs:
           mkdir -p "$STACK_ROOT"
           echo "local-programs-path: $STACK_ROOT/programs" > $STACK_ROOT/config.yaml
 
-      - run: "ci/run-hlint.sh --git"
-        env:
-          VERSION: "2.2.11"
-
```
rhendric, line 59: I moved HLint into the same new job I created for Weeder because:
* They're both linty things
* I don't think there's much point to running HLint three times
* When all the caches are loaded up, this makes the longest paths shorter and so decreases the total time to passing all checks

lib/purescript-cst/src/Language/PureScript/CST/Parser.y
```diff
@@ -1,7 +1,6 @@
 {
 module Language.PureScript.CST.Parser
   ( parseType
-  , parseExpr
```
rhendric, line 4: What are the odds someone's using this to parse an expression somewhere?

src/Language/PureScript/CoreImp/Optimizer/Common.hs
```diff
@@ -43,14 +43,6 @@ isRebound js d = any (\v -> isReassigned v d || isUpdated v d) (everything (++)
   variablesOf (Var _ var) = [var]
   variablesOf _ = []
 
-isUsed :: Text -> AST -> Bool
-isUsed var1 = everything (||) check
```
rhendric, line 47: ‘And the award for most ironic unused declaration goes to...’

src/Language/PureScript/Docs/RenderedCode/Types.hs
```diff
@@ -210,30 +185,6 @@ instance A.ToJSON RenderedCodeElement where
...
-  -- because the old code did not save information which is necessary to
-  -- distinguish these cases. This is the best we can do.
-  backwardsCompat =
-    [ oldAsIdent
-    , oldAsCtor
-    ]
-
-  oldAsIdent = firstEq "ident" (Symbol ValueLevel <$> nth 1 asText <*> nth 2 (Link <$> asContainingModule))
-  oldAsCtor = firstEq "ctor" (Symbol TypeLevel <$> nth 1 asText <*> nth 2 (Link <$> asContainingModule))
-
```
rhendric, line 236: `asRenderedCodeElement` (along with `asRenderedCode`) is more substantial than most of the stuff I deleted. Pursuit doesn't use it (I compiled Pursuit locally against this branch to test), but is it possible something else does?

JordanMartinez:
Maybe we should notify people of the functions being removed on Discourse, so that if anyone is using one of them, they have a greater chance of seeing this PR?

JordanMartinez:
Just a thought. Is there a way to deduplicate some of the versions and commands used throughout the file (e.g. the `Stack` version)?

rhendric:
> Just a thought. Is there a way to deduplicate some of the versions and commands used throughout the `ci.yml` file (e.g. the `Stack` version)?

I wish. I tried using YAML anchors, but GitHub explicitly said they aren't supported.

JordanMartinez:
What if we specified the version in the [`matrix`](https://github.com/haskell/actions/blame/main/setup/README.md#L71-L96) but only with one value and a note that we only use this as a workaround and that only one version should ever be specified?

rhendric:
Not sure how that would help? We'd still be duplicating across jobs, since each job gets its own matrix.

We could maybe solve the duplication issue by merging the jobs into one with a four-element matrix, and making most of the steps conditional on whether they're running a build or a lint. I'm not sure that's better; duplication is inelegant, but at least it's easy to read and the comments are there hopefully to keep things from drifting apart.

lib/purescript-cst/src/Language/PureScript/CST/Parser.y
hdgarrood: I wouldn't be surprised if it isn't used downstream, but in any case I think this is quite likely to be useful, and doesn't really cost us much to keep it aroundas it's just the `expr` declaration from the grammar, which has to exist anyway, so I'd be tempted to make an exception in this case.

src/Language/PureScript/Docs/RenderedCode/Types.hs
hdgarrood: This is probably an artifact of a previous failed attempt to have the Pursuit JSON blob contain `RenderedCode` rather than `Type`. If Pursuit still compiles against this branch I think it is fine to remove.

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -310,26 +310,15 @@ putEnv env = modify (\s -> s { checkEnv = env })
 modifyEnv :: (MonadState CheckState m) => (Environment -> Environment) -> m ()
 modifyEnv f = modify (\s -> s { checkEnv = f (checkEnv s) })
 
--- | Run a computation in the typechecking monad, starting with an empty @Environment@
-runCheck :: (Functor m) => StateT CheckState m a -> m (a, Environment)
```
hdgarrood, line 314: I can imagine that this might be useful for development or testing? I don't think I've ever used it myself though.

hdgarrood (approved):

src/Language/PureScript/TypeChecker/Monad.hs
rhendric: I doubt it's even useful for that. Currently the ‘typechecking monad’ is a combination of `MonadState`, `MonadWriter`, and `MonadError`; this is just a flimsy wrapper around `runStateT`.

hdgarrood: Fair enough - I'm happy to remove this then.

JordanMartinez:
Restarting CI. The mac build failed with what appears to be a GH issue. It just said, "This check failed." No groups to collapse/expand or anything.

Issue #4143, opened 2021-07-10
Fold `purescript-psa` into `purs compile`


JordanMartinez:
## Summary

Update the compiler to do the same tasks done by [`purescript-psa`](https://github.com/natefaubion/purescript-psa), so that one does not need to have `purescript-psa` installed.

## Motivation

This idea has already been mentioned and found support from both [Harry](https://github.com/purescript/purescript/issues/3778#issuecomment-578986966) and [Nate](https://github.com/purescript/purescript/issues/3778#issuecomment-578987087), the author of `purescript-psa`. However, it's only now being formally proposed.

AFAIK, there are two things the compiler cannot do that requires `purescript-psa` to implement:
- returning a non-zero exit code (e.g. failing the CI build) if there are any warnings (via the `--strict` flag)
- censoring and/or filtering errors and/or warnings (via the `censor-codes` and `filter-codes` args)

In addition, `purescript-psa` outputs information in a more helpful manner than `purs compile`'s default:
- colored warnings and errors, which makes it easier to see which are warnings and which are errors.
- statistics on the number of warnings and errors across both the library directory and source directory

When we were migrating the ecosystem to `v0.14.0`, we had to use `purescript-psa` at some point to do what the compiler was unable to do: fail the CI build if a warning was produced. However, this complicated things as well when [`MonadZero` was deprecated](https://github.com/purescript/purescript-control/issues/62), as we wanted to continue emitting warnings about the type class' deprecation. 

Other related issues where `purescript-psa` provided solutions not yet implemented in the compiler: #1898 and #2308

## Proposal

I propose we port `purescript-psa` to the compiler, so that what one would normally run via `purescript-psa` would work via `purs compile`. Below is the relevant args that `purescript-psa` provides per its ['usage' docs](https://github.com/natefaubion/purescript-psa/blob/master/src/Main.purs#L314-L343):
```
  --verbose-stats        Show counts for each warning type
  --censor-stats         Censor warning/error summary
  --censor-warnings      Censor all warnings
  --censor-lib           Censor warnings from library sources
  --censor-src           Censor warnings from project sources
  --censor-codes=CODES   Censor specific error codes
  --filter-codes=CODES   Only show specific error codes
  --no-colors            Disable ANSI colors
  --no-source            Disable original source code printing
  --strict               Promotes src warnings to errors
  --stash                Enable persistent warnings (defaults to .psa-stash)
  --stash=FILE           Enable persistent warnings using a specific stash file
[...]

Issue #3493, opened 2018-12-20
Make the compiler package aware


garyb:
Opening an issue as a place to discuss this. There was a lot of chat on Slack about things related to #2437 yesterday, and this idea was floated again. I think it was @natefaubion and @kRITZCREEK in particular that had some points of why this might be preferable, so please reiterate below :wink:

kRITZCREEK:
Quoted from Slack:

To me making the compiler package aware is making sure we can introduce more features without making a breaking release after declaring 1.0.0.

For example:
1. Make a module (or explicit exported declarations) package visible, so we can avoid the .Internal nonsense for declarations that shouldn't be exposed and just want to be shared between different modules in a package
2. Think about raising the compilation unit level from module to package
  - to allow for more aggressive monomorphization for production builds
  - (other things I'm sure @rightfold has more opinions on)

kRITZCREEK:
I don't think I'd want to go that far, especially in dev-builds the fine grained dependency tracking and good parallelism we get from our per-module compilation unit gives us acceptable compile times. So I don't think we should allow things that force us to raise the compilation unit to a package.

hdgarrood:
One thing I’d really like to do is to make it so that you can’t import transitive dependencies directly without declaring that you’re using them; then we don’t need to consider dropping a dependency as a breaking change.

chexxor:
If the compiler is package-aware, how aware of package distribution should it be? I think the compiler just needs to know the directories on your system in which you're keeping packages. Package location, publishing, and retrieval can be figured out later, then. 

Should package names be globally unique? Not when it comes to distribution and retrieval, but only when the compiler is using them.

rightfold:
@chexxor The trivial mechanism would be to give it a mapping from package name to path on the command-line. Then, no directories have to be searched, so it stays very simple. This is at the cost of having to pass more arguments to the compiler.

Still, the compiler is rarely invoked manually. I think baking conventions into the compiler is a bad idea; it makes build systems harder to configure. The configuration of the compiler should be as explicit as possible.

hdgarrood:
> The trivial mechanism would be to give it a mapping from package name to path on the command-line. Then, no directories have to be searched, so it stays very simple.
> ```
> purs compile --package prelude packages/purescript-prelude/output \
>             --package maybe packages/purescript-maybe/output \
>             src/**/*.purs
> ```

[...]

PR #4156, opened 2021-07-14
Add `--strict` flag to `purs compile`, so that `src/` warnings count as errors

### LINKED ISSUE #4153: Add `--strict` flag to promote `src/` warnings into errors
## Summary

See #4143 and [this comment](https://github.com/purescript/purescript/issues/4143#issuecomment-879021127). This feature would add the `--strict` flag, so that any warnings emitted from code in the `src/` directory are treated like errors.

JordanMartinez:
**Description of the change**

Fixes #4153.

How should I write a test for this? `TestCompiler.hs` doesn't call `purs` directly.

Example output:
```bash
$ stack build --fast
...
$ stack exec bash
$ mkdir -p ztest/src
$ cd ztest
$ cat <<EOF > src/Main.purs
module Main where

test :: Int -> String
test triggerUnusedWarning = ""
EOF

$ purs compile --json-errors "src/**/*.purs"
Compiling Main
{"warnings":[{"position":{"startLine":4,"startColumn":6,"endLine":4,"endColumn":26},"message":"  Name triggerUnusedWarning was introduced but not used.\n\nin value declaration test\n","errorCode":"UnusedName","errorLink":"https://github.com/purescript/documentation/blob/master/errors/UnusedName.md","filename":"src/Main.purs","moduleName":"Main","suggestion":null,"allSpans":[{"start":[4,6],"name":"src/Main.purs","end":[4,26]}]}],"errors":[]}

$ rm -rf output/
$ purs compile --json-errors --strict "src/**/*.purs"
Compiling Main
{"warnings":[],"errors":[{"position":{"startLine":4,"startColumn":6,"endLine":4,"endColumn":26},"message":"  Name triggerUnusedWarning was introduced but not used.\n\nin value declaration test\n","errorCode":"UnusedName","errorLink":"https://github.com/purescript/documentation/blob/master/errors/UnusedName.md","filename":"src/Main.purs","moduleName":"Main","suggestion":null,"allSpans":[{"start":[4,6],"name":"src/Main.purs","end":[4,26]}]}]}
```

---

**Checklist:**

### DIFF
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -64,7 +64,7 @@ promoteSrcWarnings False w e = (w, e)
 promoteSrcWarnings True w e = (P.MultipleErrors nonSrcWarnings, srcWarningsAndErrors)
   where
   srcWarningsAndErrors = case e of
-    Left errors -> Left $ P.MultipleErrors $ srcWarnings <> (P.runMultipleErrors errors)
+    Left errors -> Left $ P.MultipleErrors $ srcWarnings <> P.runMultipleErrors errors
     r@(Right _)
       | null srcWarnings -> r
       | otherwise -> Left $ P.MultipleErrors srcWarnings
--- a/app/Command/Compile.hs
+++ b/app/Command/Compile.hs
@@ -7,7 +7,9 @@ import           Control.Monad
 import qualified Data.Aeson as A
 import           Data.Bool (bool)
 import qualified Data.ByteString.Lazy.UTF8 as LBU8
-import           Data.List (intercalate)
+import           Data.Maybe (fromMaybe)
+import           Data.List (intercalate, isPrefixOf, partition)
+import qualified Data.List.NonEmpty as NEL
 import qualified Data.Map as M
 import qualified Data.Set as S
 import qualified Data.Text as T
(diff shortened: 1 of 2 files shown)


app/Command/Compile.hs
```diff
@@ -30,26 +32,54 @@ data PSCMakeOptions = PSCMakeOptions
...
+    partition isSrcWarning $ P.runMultipleErrors w
+
+  isSrcWarning :: P.ErrorMessage -> Bool
+  isSrcWarning warning = fromSourceDir $ fromMaybe "" $ do
+    spans <- P.errorSpan warning
+    pure $ P.spanName $ NEL.head spans
+    where
+    -- account for both POSIX and Windows path separators
+    fromSourceDir fileName =
+      isPrefixOf "src/" fileName || isPrefixOf "src\\" fileName
```
JordanMartinez, line 82: I'm not sure we want to hard code this to `src/` in case someone wants to promote `test/` warnings as well.

hdgarrood: Yes, I think this needs to be configurable because I'm sure people are putting source files in places other than `src` in apps. I'd suggest using the same strategy as psa does - having a list of directory names that indicate that a source file should not count as a local project source file if one of its parent directories matches an entry in the list. So I think we need an `--is-lib` option as well.

natefaubion: > Yes, I think this needs to be configurable because I'm sure people are putting source files in places other than src in apps.

Note this is why the predicate is for positively matching library sources. Anything that's not a library source doesn't have to be specified.

natefaubion: I think we should still make it configurable, but I just mean to say that I'd wager it's extremely uncommon to have custom library sources.

rhendric:
If this is going to depend on something like `--is-lib`, then I strongly feel that we should hard pause on implementation until we have more of a design for how this interacts with both warning management in general and the set of questions orbiting #3493. With `purescript-psa` in existence, I don't think we need to rush into a solution if it isn't going to play well with future improvements we're likely to want.

app/Command/Compile.hs
JordanMartinez: >  I just mean to say that I'd wager it's extremely uncommon to have custom library sources.

That might change when the registry gets implemented and we also implement targets.

JordanMartinez:
Isn't #3493 a breaking change? So, would that really matter? Or is this so that implementing that feature is easier because we don't have cruft like this PR complicating its design?

app/Command/Compile.hs
JordanMartinez: Latest commit adds the `--is-lib` using a comma-separated value, with the default value being `.spago,bower_components`.

rhendric:
I think it matters because we don't want to implement and then reimplement the same feature three different ways, especially if things like external tool integrations need to be adapted each time we redo it.

Giving the compiler more of an awareness of where source files are coming from and what that means is a huge design space. I think it'd be a mistake for our first step into that space to be driven by the most incidental feature touching it, especially since that feature is just reimplementing something that is already provided by external tooling. There's no user gain in the short term, and only cost if we change the `--is-lib` convention later because we settle on something else for managing per-library deprecation suppression, and then settle on a third thing for package-internal visibility or module namespacing or shipping pre-compiled library artifacts. That's cost to us in terms of wasted work, cost to the ecosystem in terms of external tooling being reworked, and cost to our users in terms of education. Let's measure twice and cut once; there are some features worth rushing towards but this one gets users nothing they can't already have today.

JordanMartinez:
Thanks! That's a good response to my question.

I'm ok with closing this. I can work on some of the other things mentioned in #4143.

I'll give the other people who have commented in this PR time to respond. If no activity occurs in 3 days, let's close this.

JordanMartinez:
Closing now that the 3 day time period has passed.

app/Command/Compile.hs
wclr: Is making this option to be used like `--strict=src/**/*.purs` (which could be passed multiple times for multiple locations) would not work out (without `--is-lib`)?

PR #4157, opened 2021-07-16
Check role declarations arity during type checking

kl0tl:
**Description of the change**

As Jordan noticed in https://github.com/purescript/purescript/pull/4121#discussion_r667256491 we incorrectly infer the arity of foreign data types whose kind contains parens (and type synonyms) while checking their role declaration arity. We ought to check role declarations arity during type checking, so that parens and type synonyms are desugared away.

---

**Checklist:**

- [x] ~Added myself to CONTRIBUTORS.md (if this is my first contribution)~
- [x] ~Linked any existing issues or proposals that this pull request should close~
- [x] ~Updated or added relevant documentation~

### DIFF
--- a/src/Language/PureScript/TypeChecker/Roles.hs
+++ b/src/Language/PureScript/TypeChecker/Roles.hs
@@ -6,13 +6,15 @@
 module Language.PureScript.TypeChecker.Roles
   ( lookupRoles
   , checkRoles
-  , checkDataBindingGroupRoles
   , checkRoleDeclarationArity
+  , inferRoles
+  , inferDataBindingGroupRoles
   ) where
 
 import Prelude.Compat
 
-import Control.Monad (unless)
+import Control.Arrow ((&&&))
+import Control.Monad (unless, when, zipWithM_)
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State (MonadState(..), runState, state)
 import Data.Coerce (coerce)
@@ -82,18 +84,42 @@ lookupRoles
 lookupRoles env tyName =
   fromMaybe [] $ M.lookup tyName (types env) >>= typeKindRoles . snd
 
--- | This function does the following:
---
--- * Infers roles for the given data type declaration
---
--- * Compares the inferred roles to the explicitly declared roles (if any) and
---   ensures that the explicitly declared roles are not more permissive than
---   the inferred ones
+-- |
+-- Compares the inferred roles to the explicitly declared roles and ensures
+-- that the explicitly declared roles are not more permissive than the
+-- inferred ones.
 --
 checkRoles
   :: forall m
    . (MonadError MultipleErrors m)
-  => Environment
+  => [(Text, Maybe SourceType, Role)]
+    -- ^ type parameters for the data type whose roles we are checking
+  -> [Role]
+    -- ^ roles declared for the data type
+  -> m ()
+checkRoles tyArgs declaredRoles = do
+  let k (var, _, inf) dec =
+        when (inf < dec) . throwError . errorMessage $ RoleMismatch var inf dec
+  zipWithM_ k tyArgs declaredRoles
+
+checkRoleDeclarationArity
+  :: forall m
+   . (MonadError MultipleErrors m)
+  => ProperName 'TypeName
+  -> [Role]
+  -> Int
+  -> m ()
+checkRoleDeclarationArity tyName roles expected = do
+  let actual = length roles
+  unless (expected == actual) $
+    throwError . errorMessage $
+      RoleDeclarationArityMismatch tyName expected actual
+
+-- |
+-- Infers roles for the given data type declaration.
+--
+inferRoles
+  :: Environment
   -> ModuleName
   -> ProperName 'TypeName
     -- ^ The name of the data type whose roles we are checking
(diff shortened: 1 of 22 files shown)


rhendric:
I wonder, would it be simpler to position the role declarations after the data types in desugaring, insert the inferred roles into the environment when the type declaration is checked, and then update the environment when checking the role declaration for arity and mismatches? Then we wouldn't have to do as much fiddling with source spans and error hints; any issues with the role declaration would naturally arise while processing said declaration.

tests/purs/failing/CoercibleRoleMismatch1.out
```diff
@@ -1,14 +1,14 @@
...
+at tests/purs/failing/CoercibleRoleMismatch1.purs:6:1 - 6:27 (line 6, column 1 - line 6, column 27)
 
   Role mismatch for the type parameter [33ma[0m:
 
     The annotation says [33mphantom[0m but the role [33mrepresentational[0m is required.
 
 
 in role declaration for [33mIdentity[0m
-in type constructor [33mIdentity[0m
+in data binding group Identity
```
kl0tl, line 11: Having foreign data type declarations depend on their role declaration in the binding group desugaring step would complicate self cycles detection, but this leaks in error messages since we do display that role declarations of data types belong to a binding group here. Is there a reason for not eliminating the hint for singleton binding groups?

kl0tl:
I applied your suggestion. This was slightly more involved than expected because we need to know all the role declarations of data binding groups when inferring roles of mutually recursive data types, but it is nice to remove declared roles from the environment and to position RoleMismatch errors accurately!

tests/purs/failing/CoercibleRoleMismatch1.out
rhendric: I might even go further and suggest that it should be eliminated whenever another hint narrows down the error to one declaration. Maybe there should be a hint category for top-level declarations or something?

kl0tl: I introduced a category for declaration hints.

JordanMartinez:
This now has a merge conflict:
```
<<<<<<< check-role-declarations-arity-during-typechecking
  go d@(ExternDataDeclaration _ name kind) = do
    elabKind <- withFreshSubstitution $ checkKindDeclaration moduleName kind
    env <- getEnv
    let qualName = Qualified (Just moduleName) name
        roles = nominalRolesForKind elabKind
    putEnv $ env { types = M.insert qualName (elabKind, ExternData roles) (types env) }
    return d
=======
  go d@(ExternDataDeclaration (ss, _) name kind) = do
    warnAndRethrow (addHint (ErrorInForeignImportData name) . addHint (positionedError ss)) $ do
      elabKind <- withFreshSubstitution $ checkKindDeclaration moduleName kind
      env <- getEnv
      let qualName = Qualified (Just moduleName) name
      -- If there's an explicit role declaration, just trust it
      let roles = fromMaybe (nominalRolesForKind elabKind) $ M.lookup qualName (roleDeclarations env)
      putEnv $ env { types = M.insert qualName (elabKind, ExternData roles) (types env) }
      return d
>>>>>>> master
```

JordanMartinez:
CI builds again. Any other feedback on this PR?

src/Language/PureScript/TypeChecker/Roles.hs
```diff
@@ -80,72 +84,87 @@ lookupRoles
...
    . (MonadError MultipleErrors m)
-  => Environment
+  => [(Text, Maybe SourceType, Role)]
+    -- ^ type parameters for the data type whose roles we are checking
+  -> [Role]
+    -- ^ roles declared for the data type
+  -> m ()
+checkRoles tyArgs declaredRoles = do
+  let k (var, _, inf) dec =
+        when (inf < dec) . throwError . errorMessage $ RoleMismatch var inf dec
```
JordanMartinez, line 102: I think this error message could be improved to indicate which type parameter's role was wrong in a declaration with 2+ role annotations. For example, looking at this error message
- [CoercibleRoleMismatch1.purs](https://github.com/purescript/purescript/blob/146b795fb9084994bd12b971dd559b2a1e0911f4/tests/purs/failing/CoercibleRoleMismatch1.purs)
- [CoercibleRoleMismatch1.out](https://github.com/purescript/purescript/pull/4157/files?w=0#diff-ba8180994d237116bc772b4e39a6e31fc4c9f5ca66e3ce14602d4614acd23c6cR3)

... the source spans indicates the entire role declaration. They do not indicate which role declaration has the issue.

If I have the following example,
```purescript
data Foo a b c d = Foo a b c

type role Foo representational phantom representational phantom
```
and `b` is the one with the bad role, the error message wouldn't tell me which of the four values above is wrong.

JordanMartinez (changes requested):
I think this is good to go aside from the one issue raised below.

src/Language/PureScript/TypeChecker/Roles.hs
JordanMartinez: On second thought, the issue above is a separate issue. This PR AFAICT accomplishes its goal of checking for arity. I'll retract this.

JordanMartinez (approved):

PR #4169, opened 2021-07-26
Check for partially applied synonyms in ctors

rhendric:
This check doesn't prevent any programs from compiling; it just makes
sure that a more specific PartiallyAppliedSynonym error is raised
instead of a KindsDoNotUnify error, which could be interpreted as
implying that a partially applied synonym has a valid kind and would be
supported elsewhere if that kind is expected.

**Description of the change**

Another stepping stone on the path to #4086. This is fairly superficial but without it, some test outputs would change from `PartiallyAppliedSynonym` to `KindsDoNotUnify` when type checking and instance deriving are combined, and I figure that'd be a regression.

---

**Checklist:**

- <s>Linked any existing issues or proposals that this pull request should close</s>

### DIFF
--- a/src/Language/PureScript/TypeChecker/Kinds.hs
+++ b/src/Language/PureScript/TypeChecker/Kinds.hs
@@ -180,7 +180,7 @@ inferKind = \tyToInfer ->
           throwError . errorMessage' (fst ann) . UnknownName . fmap TyClassName $ v
         Just _ ->
           checkConstraint con
-      ty' <- checkKind ty E.kindType
+      ty' <- checkIsSaturatedType ty
       con'' <- applyConstraint con'
       pure (ConstrainedType ann' con'' ty', E.kindType $> ann')
     ty@(TypeLevelString ann _) ->
@@ -228,10 +228,10 @@ inferKind = \tyToInfer ->
     ForAll ann arg mbKind ty sc -> do
       moduleName <- unsafeCheckCurrentModule
       kind <- case mbKind of
-        Just k -> replaceAllTypeSynonyms =<< checkKind k E.kindType
+        Just k -> replaceAllTypeSynonyms =<< checkIsSaturatedType k
         Nothing -> freshKind (fst ann)
       (ty', unks) <- bindLocalTypeVariables moduleName [(ProperName arg, kind)] $ do
-        ty' <- apply =<< checkKind ty E.kindType
+        ty' <- apply =<< checkIsSaturatedType ty
         unks <- unknownsWithKinds . IS.toList $ unknowns ty'
         pure (ty', unks)
       for_ unks . uncurry $ addUnsolved Nothing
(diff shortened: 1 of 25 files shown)


JordanMartinez (approved):

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -626,6 +626,11 @@ inferDataConstructor
   -> DataConstructorDeclaration
   -> m (DataConstructorDeclaration, SourceType)
 inferDataConstructor tyCtor DataConstructorDeclaration{..} = do
+  -- Check that type synonyms can expand--we want partially applied type
+  -- synonyms to raise that error instead of KindsDoNotUnify--but keep the
+  -- unexpanded types in the declaration so that things like the transitive
+  -- exports check still work.
```
natefaubion, line 632: I'm not saying you should change this now, but by this reasoning, should transitive export checks be happening before type-checking?

rhendric: Ideally, I think that would work? I don't think the environment prior to type checking has all of the entries that the transitive export checks use, but maybe things could be moved around to fix that.

rhendric:
I know there are sexier PRs open right now, but if this can get reviewed and merged, there's another one in queue behind it that closes five issues (that I know of; I only discovered the fifth last month!) and touches some desugaring code that might interact with other PRs.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -666,7 +695,7 @@ inferTypeSynonym moduleName (ann, tyName, tyArgs, tyBody) = do
   let (sigBinders, tyKind') = fromJust . completeBinderList $ tyKind
   bindLocalTypeVariables moduleName (first ProperName . snd <$> sigBinders) $ do
     kindRes <- freshKind (fst ann)
-    tyArgs' <- for tyArgs . traverse . maybe (freshKind (fst ann)) $ replaceAllTypeSynonyms <=< apply <=< flip checkKind E.kindType
+    tyArgs' <- for tyArgs . traverse . maybe (freshKind (fst ann)) $ replaceAllTypeSynonyms <=< apply <=< checkIsSaturatedType
```
garyb, line 698: Does changing this in the `inferTypeSynonym` case prevent any of the current dodgy-but-not-really uses of synonyms, where you can do things like:

```
type Things f =
  { thing1 :: f Unit Number
  , thing2 :: f String Boolean
  }

type Lefts :: Type -> Type -> Type
type Lefts l r = l

type LeftThings = Things Lefts
```
?

rhendric: It shouldn't; this line change is about the kinds of type synonym parameters. So if you had
```purs
type Foo (a :: Some Type Stuff) = ...
```
and any synonyms in `Some Type Stuff` didn't end up expanding, this change may affect the error displayed (change it from KindsDoNotUnify to PartiallyAppliedSynonym).

The test for this particular line change, which I just re-verified fails if I revert it, is PASTrumpsKDNU6.

garyb: Ah right, yep!

garyb (approved):

PureFunctor (approved):

PR #4172, opened 2021-08-08
Fix backtick offside rule

### LINKED ISSUE #4171: Infix names with backticks don't use the same offside rule as operators
## Description

Infix (backtick) names do not affect layout in the same way as operators do. Operators use the `offsideEndP` rule which lets you write:

```purescript
example = do
  foo bar
  <|> baz
```

An operator at the same column as a layout block will close the layout block. Infix expressions use the `offsideP` rule through `insertDefault`.

```purescript
example = do
  foo bar
  `wat` baz
```

This will result in a parse error. It should parse with the same offside rule as operators.

## Additional context

https://github.com/natefaubion/purescript-tidy/issues/15#issuecomment-894552462

## PureScript version

0.14.3

JordanMartinez:
**Description of the change**

Fixes #4171 by applying the change suggested by Nate in that issue. 

I'm guessing this is not entirely correct because one of the other golden tests was affected by this change, too.

---

**Checklist:**

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
@@ -265,7 +265,7 @@ insertLayout src@(SourceToken tokAnn tok) nextPos stack =
         ((_, LytTick) : stk', acc') ->
           (stk', acc') & insertToken src
         _ ->
-          state & insertDefault & pushStack tokPos LytTick
+          state & collapse offsideEndP & insertSep & insertToken src & pushStack tokPos LytTick
 
     -- In general, commas should close all indented contexts.
     --     example = [ do foo
--- a/tests/purs/layout/BacktickOperator.out
+++ b/tests/purs/layout/BacktickOperator.out
@@ -12,5 +12,5 @@ example3 =
   case _ of{
     Foo a -> 1;
     Bar b -> 2}
-  `append` 3}
+    `append` 3}
   <eof>
--- a/tests/purs/layout/BacktickOperator.out
+++ b/tests/purs/layout/BacktickOperator.out
@@ -1,0 +1,16 @@
+module Test where{
+
+example1 = do{
+  foo bar}
+  <|> baz;
+
+example2 = do{
+  foo bar}
+  `wat` baz;
+
+example3 =
+  case _ of{
+    Foo a -> 1;
+    Bar b -> 2}
+  `append` 3}
+  <eof>
--- a/tests/purs/layout/BacktickOperator.out
+++ b/tests/purs/layout/BacktickOperator.out
@@ -12,5 +12,11 @@ example3 =
   case _ of{
     Foo a -> 1;
     Bar b -> 2}
-    `append` 3}
-  <eof>
+    `append` 3;
+
+example4 =
+  case _ of{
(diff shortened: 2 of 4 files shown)


lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
```diff
@@ -265,7 +265,7 @@ insertLayout src@(SourceToken tokAnn tok) nextPos stack =
         ((_, LytTick) : stk', acc') ->
           (stk', acc') & insertToken src
         _ ->
-          state & insertDefault & pushStack tokPos LytTick
+          state & collapse offsideEndP & insertSep & insertToken src
```
natefaubion, line 268: You still need to `pushStack`. We should only be replaced the `insertDefault`.

JordanMartinez: Ah... Now I see.

JordanMartinez:
I tried testing this locally and it seemed to just hang. Not sure how CI will do, but I pushed my changes anyways.

tests/purs/layout/BacktickOperator.purs
```diff
@@ -0,0 +1,16 @@
...
+
+example2 = do
+  foo bar
+  `wat` baz
+
+example3 =
+  case _ of
+    Foo a -> 1
+    Bar b -> 2
+  `append` 3
```
natefaubion, line 15: I think this should be indented to the same level as the case branches, not the case keyword.

JordanMartinez: Ah... Done.

natefaubion (approved):

tests/purs/layout/BacktickOperator.out
```diff
@@ -0,0 +1,16 @@
...
+
+example2 = do{
+  foo bar}
+  `wat` baz;
+
+example3 =
+  case _ of{
+    Foo a -> 1;
+    Bar b -> 2}
+    `append` 3}
```
rhendric, line 15: This isn't right, is it? First, the braces don't match; second, I'd expect
```suggestion
  {case _ of
    Foo a -> 1;
    Bar b -> 2}
    `append` 3
```

(which of course wouldn't type check, because `append` would need type `(a -> Int) -> Int -> b`, but that's not really relevant to a layout test).

Compare with how
```purs
example3 =
  case _ of
    Foo a -> 1
    Bar b -> 2
    + 3
```
currently parses.

natefaubion: That last `}` is the layout end of the module, I think.

natefaubion: I agree we should have a comparable test here for case with operators, since there is one for `do`.

rhendric: :facepalm: Of course it is. Okay, guess this all makes sense after all!

JordanMartinez: I've added a case for operators.

JordanMartinez:
Do I need to add more content to the changelog entry? The description by itself doesn't really explain what this does, but I also don't think this needs to be explained as anyone who wants to know could just look at the diff here.

PR #4182, opened 2021-09-08
Include function declarations that are referenced by top-level function calls during bundling

### LINKED ISSUE #4181: purs bundle eliminates functions called locally
## Description

PR #4044 seems to have introduced a regression in the bundler.

As @ozkutuk reported in https://github.com/purescript-contrib/purescript-arraybuffer/issues/34 , in version 0.14.4 `purs bundle` started eliminating functions from a `.js` file which are called locally in the `.js` file at “load time.” (What do we call that time at which the `.js` file is evaluated? Is it “load time”? I don't know).

## Expected behavior

This commit on my fork of __purescript__ introduces a test which I believe should pass if the bundler is working properly.

https://github.com/jamesdbrock/purescript/commit/52dd3a5881468d688263881cefc6928e462cbe0b

Instead the test fails with `purescript>        ReferenceError: localFunction is not defined`.

## To Reproduce

```
git clone https://github.com/jamesdbrock/purescript.git
cd purescript
git checkout bundle-function
stack test --test-arguments="--match FunctionDeclaration.purs"
```

## PureScript version

0.14.4

ozkutuk:
**Description of the change**

Fixes #4181

As stated in #4181, #4044 introduced a regression to the bundler. Bundler only concerns itself with require statements, member declarations, and export lists; and it ignores everything else (leaving them in the bundle verbatim). Before #4044 was introduced, both the function declaration and the call to that function were falling back to "other" case, leaving them untouched in the bundler output. However, since #4044, the function declarations are treated specially and are now properly parsed as member declarations. This causes them to be subject to DCE check, and since bundler can't detect that the declaration is referenced by a top-level function call, it eliminates the declaration during DCE.

This PR fixes the issue by parsing top-level function calls as member declarations, and also specifies them as potential entry points to the module. This allows the DCE check to detect that the function declaration is indeed referenced.

---

**Checklist:**

- [x] ~~Updated or added relevant documentation~~

### DIFF
--- a/CHANGELOG.d/fix_bundle-toplevel-function-calls.md
+++ b/CHANGELOG.d/fix_bundle-toplevel-function-calls.md
@@ -1,0 +1,1 @@
+* Do not remove function declarations that are referenced by top-level function calls when bundling
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -89,6 +89,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@noraesae](https://github.com/noraesae) | Hyunje Jun | [MIT license](http://opensource.org/licenses/MIT) |
 | [@nullobject](https://github.com/nullobject) | Josh Bassett | [MIT license](http://opensource.org/licenses/MIT) |
 | [@osa1](https://github.com/osa1) | Ömer Sinan Ağacan | MIT license |
+| [@ozkutuk](https://github.com/ozkutuk) | Berk Özkütük | [MIT license](http://opensource.org/licenses/MIT) |
 | [@paf31](https://github.com/paf31) | Phil Freeman | [MIT license](http://opensource.org/licenses/MIT) |
 | [@parsonsmatt](https://github.com/parsonsmatt) | Matt Parsons | [MIT license](http://opensource.org/licenses/MIT) |
 | [@passy](https://github.com/passy) | Pascal Hartig | [MIT license](http://opensource.org/licenses/MIT) |
--- a/src/Language/PureScript/Bundle.hs
+++ b/src/Language/PureScript/Bundle.hs
@@ -94,6 +94,7 @@ guessModuleIdentifier filename = ModuleIdentifier (takeFileName (takeDirectory f
 data Visibility
   = Public
   | Internal
+  | TopLevelCall
   deriving (Show, Eq, Ord)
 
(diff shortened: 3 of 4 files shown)


jamesdbrock:
I ran the test and it passes. 👍

rhendric:
Just skimming the code changes, it looks like this might be too narrow a fix; is this case handled?
```js
if (condition) {
  localFunction();
}
```

ozkutuk:
@rhendric Such cases are indeed not handled. I think rather than trying extend this more and more into a full-fledged JS dead code analysis tool, we should set a boundary regarding what we support on FFI files. @hdgarrood put this better than I could in https://github.com/purescript/purescript/issues/4039#issuecomment-812910288.

To be honest, I even was on the fence regarding adding support for top-level function calls, but I guess they are useful for global polyfills (as in the case of https://github.com/purescript-contrib/purescript-arraybuffer/issues/34). Do you think it is essential to support conditionals as well?

rhendric:
I think what's essential is that FFI files either work (possibly with unused code in the result, which a more advanced bundler might have eliminated) or fail at compile time (such as when a JS feature is not supported by our parser). That was the motivation behind my support of #4044, and it's why I'd rather support arbitrary statements at the top level in this fix. I could be convinced that throwing a compile-time error if such statements are encountered is the right thing to do instead (in which case this is a breaking change). But I wouldn't expect adding support for those statements to be difficult; we already have the ability to analyze arbitrary statements for their dependencies.

ozkutuk:
@rhendric What do you think is the best way to proceed with this? Should I add every possible `JSStatement` case under `matchMember`, similar to what I did with `JSMethodCall`? `JSStatement` has 35 data constructors, so I am not sure if that is the right thing to do, but couldn't come with anything better either.

rhendric:
I haven't tried it so I can't be sure it would work, but I probably would have started by leaving `matchMember` alone and instead adding a list of dependencies to the `Other` constructor, populating it in `withDeps` the same way that dependencies are populated for `Member`s, and then changing `compile` to include any dependencies in `Other`s in the list of `entryPointVertices`.

rhendric:
Superseded by #4255, much as plans to give a drab hallway a beautiful new coat of paint are superseded by the demolition of the building.

(Thank you for working on this anyway; hopefully this is not too frustrating an outcome for you!)

Issue #4183, opened 2021-09-19
Wrong kind inferred for type synonym


jy14898:
## Description

Compiler infers a kind for the synonym shown below, but that kind is incorrect.

## To Reproduce

```haskell
module Main where

type T f a = f (f a)
```

Either view the warnings or use `:k T` in the repl. It shows `forall k1 k2. (k1 -> k2) -> k1 -> k2`

## Expected behavior

The kind should be inferred to be `forall k. (k -> k) -> k -> k`

## PureScript version

0.14.4

jy14898:
I'm going to attempt this

PR #4184, opened 2021-09-19
Synonym inference bug

jy14898:
**Description of the change**

Fixes https://github.com/purescript/purescript/issues/4183

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/TypeChecker/Kinds.hs
+++ b/src/Language/PureScript/TypeChecker/Kinds.hs
@@ -665,8 +665,10 @@ inferTypeSynonym moduleName (ann, tyName, tyArgs, tyBody) = do
     tyArgs' <- for tyArgs . traverse . maybe (freshKind (fst ann)) $ replaceAllTypeSynonyms <=< apply <=< flip checkKind E.kindType
     unifyKinds tyKind' $ foldr ((E.-:>) . snd) kindRes tyArgs'
     bindLocalTypeVariables moduleName (first ProperName <$> tyArgs') $ do
-      tyBodyAndKind <- inferKind tyBody
+      tyBodyAndKind <- secondM apply =<< inferKind tyBody
       instantiateKind tyBodyAndKind =<< apply kindRes
+  where
+    secondM f (a,b) = (a,) <$> f b
 
 -- | Checks that a particular generalization is valid and well-scoped.
 -- | Implicitly generalized kinds are always elaborated before explicitly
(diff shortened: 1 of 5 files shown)


garyb (approved):
Looks sensible to me :+1:

JordanMartinez (approved):
So long as CI builds, I approve.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -665,8 +665,10 @@ inferTypeSynonym moduleName (ann, tyName, tyArgs, tyBody) = do
     tyArgs' <- for tyArgs . traverse . maybe (freshKind (fst ann)) $ replaceAllTypeSynonyms <=< apply <=< flip checkKind E.kindType
     unifyKinds tyKind' $ foldr ((E.-:>) . snd) kindRes tyArgs'
     bindLocalTypeVariables moduleName (first ProperName <$> tyArgs') $ do
-      tyBodyAndKind <- inferKind tyBody
+      tyBodyAndKind <- secondM apply =<< inferKind tyBody
       instantiateKind tyBodyAndKind =<< apply kindRes
+  where
+    secondM f (a,b) = (a,) <$> f b
```
kritzcreek, line 671: Isn't this just `traverse`?

rhendric:
This specific test case would also be fixed by using `apply` on the `TUnknown` created on line 260, in `inferAppKind`, instead of changing `inferTypeSynonym`. These solutions have different scopes of impact; there are some code paths that go through `inferAppKind` and not `inferTypeSynonym`, and vice versa. I don't know if one or the other of these solutions is more correct (or maybe they both are needed?). Does someone else?

natefaubion:
> This specific test case would also be fixed by using apply on the TUnknown created on line 260, in inferAppKind, instead of changing inferTypeSynonym. These solutions have different scopes of impact; there are some code paths that go through inferAppKind and not inferTypeSynonym, and vice versa. I don't know if one or the other of these solutions is more correct (or maybe they both are needed?). Does someone else?

Calling `apply` is idempotent, so you can put it "wherever." The point is just that whenever you expect to make progress with solving, you should always be checking against the latest solutions for the type, since you don't want to unify against outdated information. The only reason I wrote it this way was because it's a pretty literal translation of the paper. I wouldn't do it this way if I were doing it again, since it's clearly error prone. In reality you only need to do this when:
* Solving for an unknown (make sure the unknown isn't already solved, otherwise continue unifying with it's solution).
* Putting a solution into the environment (since you already need to do an occurs check).
* At the point of generalization (leftover unknown are replaced with quantifiers, and you don't want to generalize something that was concretely solved).

I think it's probably a good thing to do this in inferTypeSynonym, since I'd expect you'll want to make sure you're giving the latest to `instantiateKind`.

src/Language/PureScript/TypeChecker/Kinds.hs
jy14898: Indeed, hoogle failed me. It's always `traverse`

JordanMartinez:
Shall we merge this PR then? The `traverse` part has been fixed.

thomashoneyman (approved):

Issue #4152, opened 2021-07-13
Use color to distinguish errors from warnings after compiling


JordanMartinez:
## Summary

See #4143 and [this comment](https://github.com/purescript/purescript/issues/4143#issuecomment-879021127). This feature would use color to distinguish errors from warnings.

PR #4190, opened 2021-10-23
Color compiler warnings and errors

maurobalbi:
**Description of the change**
This PR addresses #4152.

I haven't updated the golden tests yet, in case you would like to have some changes.

Example output:
error-example
-

**Checklist:**


JordanMartinez:
> in the next few days

That was supposed to be maybe 3 days after I made that comment, but now it's almost been a week. Sorry about that. I'm currently building this PR's binary and going to test it out. However, I have friends coming by, so this might get postponed until tomorrow.

maurobalbi:
No worries, all in due course! I'm already happy that this PR is being considered!

JordanMartinez:
Here's a before and after of a warning and error using the following code:
```purescript
module Main where

import Prelude

import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = do
  log "🍝"

foo :: Int -> String
foo x = "bar"

footer :: Int -> Int
footer x = "bar"
```
| Type | State | Image |
| --- | --- | --- |
| Warning | Before | before-warning |
| Warning | After | after-warning |
| Error | Before | before-error |
| Error | After | after-error |

Two things immediately stand out:
- the content of the error message isn't indented like the content of the warning message. While we're here, I think that should be fixed
- I'm wondering if the warning color should be a slightly different color than what we're already using to highlight types and whatnot in the message. Perhaps a darker color, so that it stands a bit more in contrast to the rest of the message? However, I'm not sure that's needed because of the indentation.

If you're up for fixing the indentation, please do so and I'll approve it after that. If not, I'll approve this now since the PR does what you stated and debating about colors could delay this.

maurobalbi:
Thanks for the review, I just checked it again and it looks like the indentation of warnings and errors is consistent. The difference comes from single vs multiple errors/warnings, as seen in the screenshots:

| Type | Screenshot |
|-----|-------|
|Single error | image|
| Multiple errors | image |
| Single warning | image |
| Multiple warnings | image |

I'm up for aligning those, my preference would be to add the indentation and newline to the single error/warning. 
I tried some additional colors for the warnings, but nothing really stood out as a big improvement. 
As a reference: [https://en.wikipedia.org/wiki/ANSI_escape_code](https://en.wikipedia.org/wiki/ANSI_escape_code)

Overall I guess there's a lot of bikeshedding to be had!

JordanMartinez:
> I'm up for aligning those, my preference would be to add the indentation and newline to the single error/warning.

Ok, please add the indentation for the single error/warning. Let's leave the colors as is. Thanks for doing this!

JordanMartinez (approved):
LGTM! Thanks!

Issue #4192, opened 2021-10-27
Document the compiler


JordanMartinez:
## Summary

Make it easier for new contributors to understand the compiler, so that they can make better contributions sooner, and for current maintainers, so that they have a better idea for what's safe to merge.

## Motivation

In my opinion, compiler development has slowed down again. Harry often gave feedback to various issues and PRs; now that he's stepped down, we have one less person who can do that. While I have contributed in the past and have write access, my understanding of this repo is lacking and I am hesitant to merge things for fear of breaking something. Others that I've talked to either aren't as interested in contributing to the compiler or have similar reservations as me.

One action that can help address this is adding more documentation to this repo. Once written, it'll make it easier for the next person to contribute something and for current maintainers to have more confidence as to whether they can merge something or not.

In the process, I believe a number of "clean up" / "tech debt" tasks will be discovered. For example, changing a `data` to a `newtype`, identifying where to do #4115, etc.

## Proposal

There's a few pieces of documentation that are worth adding:
- docs on various types: what they are / represent and how they are used
- docs on functions: what their goal is and how they achieve that
- docs on support tools (e.g. linking to Happy's guide and clarifying the settings we use, clarifying why we don't use `alex` for lexing, etc.)
- an explanation for how the type checker works
- an `ARCHITECTURE.md` file that overviews the entire compiler from a very high level and discusses some of the inherit tradeoffs with its design and alternatives as well as references to things useful for understanding it

I think if we started at the top of the list and worked our way down, the writer of the harder parts of this (i.e. type checking and Architecture file) would be better informed before writing it.

Since this issue is quite large, I propose addressing it via many mini-PRs. If all of this was done in a single PR, it would be massive and likely bitrot as other changes get merged. However, if a few parts were documented in a single PR, it would be easier to review and merge. Moreover, there might be multiple passes. In the first pass, docs were added to some type that simply indicated what it represented. Later, when more parts of the compiler are documented and more understanding made, another pass could be added that provides insight into how the type is used. 
Although the mini-PR approach might lead to a huge changelog that feels spammy, I feel like the alternatives are less likely to get merged. 

Regardless, all such PRs should backlink to this issue so that we can see all that was done towards this effort.

## Useful Links:

- Haddocks - Haskell Docs:
    - [Haddocks User Guide](https://www.haskell.org/haddock/doc/html/index.html) 
[...]

PR #4193, opened 2021-10-27
Document `MonadSupply` and CST `Types`

JordanMartinez:
**Description of the change**

Part of #4192.

---

**Checklist:**

### DIFF
--- a/lib/purescript-cst/src/Control/Monad/Supply/Class.hs
+++ b/lib/purescript-cst/src/Control/Monad/Supply/Class.hs
@@ -12,7 +12,11 @@ import Control.Monad.Writer
 import Data.Text (Text, pack)
 
 class Monad m => MonadSupply m where
+  -- |
+  -- Get a new unique Integer
   fresh :: m Integer
+  -- |
+  -- See what the next unique Integer will be
   peek :: m Integer
   default fresh :: (MonadTrans t, MonadSupply n, m ~ t n) => m Integer
   fresh = lift fresh
@@ -29,5 +33,6 @@ instance Monad m => MonadSupply (SupplyT m) where
 instance MonadSupply m => MonadSupply (StateT s m)
 instance (Monoid w, MonadSupply m) => MonadSupply (WriterT w m)
 
+-- | Convenience function for returning "$x" where `x` is the next unique integer
 freshName :: MonadSupply m => m Text
 freshName = fmap (("$" <> ) . pack . show) fresh
--- a/lib/purescript-cst/src/Control/Monad/Supply/Class.hs
+++ b/lib/purescript-cst/src/Control/Monad/Supply/Class.hs
@@ -33,6 +33,15 @@ instance Monad m => MonadSupply (SupplyT m) where
 instance MonadSupply m => MonadSupply (StateT s m)
 instance (Monoid w, MonadSupply m) => MonadSupply (WriterT w m)
 
--- | Convenience function for returning "$x" where `x` is the next unique integer
+-- |
+-- `fresh`, but formatted to be usable as the name of an identifier in JS code
+-- generation.
+--
+-- At any point in the compilation pipeline prior to the renaming phase (which
+-- takes place after CoreFn optimization and before converting to CoreImp and
+-- optimizing that), if you want an Ident with a name that is guaranteed (by
+-- the renamer) not to collide with an existing Ident, you should probably use
+-- `Language.PureScript.Names.freshIdent` instead of this function.
+--
 freshName :: MonadSupply m => m Text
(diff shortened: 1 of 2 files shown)


lib/purescript-cst/src/Language/PureScript/CST/Types.hs
```diff
@@ -95,64 +119,192 @@ data Name a = Name
...
+-- |
+-- A potentially empty list of @a@ that is wrapped in some opening and closing boundary
+-- `SourceToken` (e.g. parenthesis). Each @a@ element is separated by some delimiter `SourceToken`.
+--
+-- Conceptually, if the @a@ was @Int@ and the `SourceToken` was @Text@ and the boundary
+-- tokens were parenthesis and the below syntax was supported,
+-- - @([])@
+-- - @([1])@
+-- - @([1, "a", 2])@
+-- - @([1, "a", 2, "b", 3, "c", 4])@
```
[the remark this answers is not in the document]
JordanMartinez: Yeah, that makes sense. I'll definitely be omitting these.

lib/purescript-cst/src/Language/PureScript/CST/Types.hs
```diff
@@ -76,16 +90,26 @@ data Token
   | TokInt !Text !Integer
   | TokNumber !Text !Double
   | TokLayoutStart
+    -- ^ Indicates the start of a new indentation block.
+    -- All expressions must be indented at least to here to be considered part of the block.
   | TokLayoutSep
+    -- ^ Indicates a separator between two entities within the same indentation block.
   | TokLayoutEnd
+    -- ^ Indicates the end of an indentation block.
+    -- All expressions after this one will be part of the previous block.
```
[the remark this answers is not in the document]
JordanMartinez: I've tweaked your third bullet point slightly and used that as the doc comments for them. Rather than stating that for each data constructor, I've moved it to the type's docs.

JordanMartinez:
I've done some more clean up. However, could someone further direct me on what else to drop or change? I'm hesitant to drop some other docs (e.g. on the data constructors) because when one hovers over the data constructor, they don't see the docs on the type itself. On the other hand, I also realize that dropping such docs is one less thing to maintain.

JordanMartinez:
In light of [Nate persuading me](https://github.com/purescript/purescript/issues/4192#issuecomment-962193712), I propose doing the following to this PR:
- keep the docs on `MonadSupply` (it doesn't hurt)
- keep the docs on the type, but not data constructors, for `Token` (i.e. the layout tokens docs)
- drop all other documentation

lib/purescript-cst/src/Control/Monad/Supply/Class.hs
```diff
@@ -29,5 +33,6 @@ instance Monad m => MonadSupply (SupplyT m) where
 instance MonadSupply m => MonadSupply (StateT s m)
 instance (Monoid w, MonadSupply m) => MonadSupply (WriterT w m)
 
+-- | Convenience function for returning "$x" where `x` is the next unique integer
```
rhendric, line 36: IMO this comment doesn't need to describe the implementation of `freshName`; instead it should contain a description of the sorts of situations in which `freshName` is intended to be used, as opposed to the similar `freshIdent'` in `Language.PureScript.Names`. (The facts on the ground are that `freshName` is used in the CoreImp optimizer and the JS generator; it's also used in one place in `L.PS.Linter.Exhaustive` but I'm not sure that instance shouldn't actually be `freshIdent'`—I've just never had a reason to try rewriting it.)

JordanMartinez: Mind pushing a commit that makes that update?

rhendric: Can do, though I was hoping another maintainer might chime in to fill in the blanks I'm missing.

JordanMartinez: Then, perhaps we remove that for now, too, and fill it in at a later time?

rhendric: I gave it a tweak; I'm like 90% sure this is correct, and I don't want to let the perfect delay the good.

JordanMartinez:
I've removed all the comments previously in Types.hs aside from the one for `Type`.

JordanMartinez:
Any other feedback on this? Or is this good to merge?

JordanMartinez:
Closing because I'm not sure whether the docs in this PR are correct, and it's better to have no docs that incorrect ones.

PR #4195, opened 2021-10-30
New algorithm for computing covering sets

### LINKED ISSUE #4194: Erroneous overlapping instances
## Description

The compiler reports overlapping instances even though the two instances should not be overlapping with the given fundeps.

## To Reproduce

```purescript
module Test.Test.Test where

data Identity a
data Maybe a

class ErrorSemigroup o m w | w -> o m, o m -> w

instance ErrorSemigroup (Identity o) (Identity m) (Identity w)

instance ErrorSemigroup o (Maybe m) (Maybe w)
```

Gives this error:

```
[1/1 OverlappingInstances] src/Test.purs:10:1

  10  instance ErrorSemigroup o (Maybe m) (Maybe w)
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
    Test.Test.Test.ErrorSemigroup o
                                  (Maybe m)
                                  (Maybe w)
  
    instance in module Test.Test.Test with type forall m o w. ErrorSemigroup (Identity o) (Identity m) (Identity w) (line 8, column 1 - line 8, column 63)
    instance in module Test.Test.Test with type forall m o w. ErrorSemigroup o (Maybe m) (Maybe w) (line 10, column 1 - line 10, column 46)
  
  in type class instance
  
    Test.Test.Test.ErrorSemigroup o
                                  (Maybe m)
                                  (Maybe w)
```

Note that changing the fundeps to just `w -> o m` or just `o m -> w` or no fundeps removes the error.

## Expected behavior

The compilation should succeed. Clearly `w -> o m` means that `w` by itself will determine the instance: and indeed, the instances are distinct in `w`. Similarly, `o m -> w` means that `o` and `m` together will determine the instance: they are still distinct in `m`, though they overlap in `o` (but that's okay – just one needs to be distinct).

## Additional context

It seems to be a problem with the covering sets computed for the instance.

This is the algorithm to compute them:
https://github.com/purescript/purescript/blob/9908e622e4bc2fb9efdc8789c3ea7021cf3b4698/lib/purescript-cst/src/Language/PureScript/Environment.hs#L108-L127

I haven't been able to test that specifically, but I would expect the result to be `{{w},{o,m}}` and it seems like it is something else.

In particular, it seems that the logic above may miss the interactions between multiple variables together determining the instance. (I haven't fully understood the algorithm yet.)

-----

Here is where the covering sets are used to detect overlapping instances:
https://github.com/purescript/purescript/blob/c14d52e0bd4caca88358ac082911e8a94d072dd5/src/Language/PureScript/TypeChecker.hs#L514
https://github.com/purescript/purescript/blob/c14d52e0bd4caca88358ac082911e8a94d072dd5/src/Language/PureScript/TypeChecker.hs#L529

In particular the logic is: at least one argument must be considered apart from its corresponding argument, for every covering set of arguments.

So since it compiles with the fundeps individually, it should compile with them together. (At least since the fundeps don't interact in any nontrivial way.)

## Related issues

#4135 may be related, since it uses the same pattern of fundeps, but the apartness logic for that is in a different part of the compiler. I verified that that test case still gives an error on the latest version of the compiler.

## PureScript version

0.14.5

MonoidMusician:
Fixes #4194.

First try at a new algorithm to compute covering sets. Basically we compute the closure step-by-step and see which (inclusion-minimal) subsets actually close to the whole set of arguments.

Could definitely be improved – currently all closures are computed, not just the inclusion-minimal ones.

I don't understand the `determinedArgs` check well enough, but I hope we can use the same data I'm using for the covering sets to also compute it, instead of having both the graph algorithm and my algorithm.

**Checklist:**


MonoidMusician:
This is ready for review finally. Things seem to work with the new check for determined arguments (namely, that they aren’t required for any covering set), but it probably isn’t excercised much – do we agree that it has the correct behavior?

rhendric:
I'm finding this covering sets algorithm a little opaque. If I understand correctly, you're taking every possible subset of parameters and brute-forcing your way through all the possible parameters that could be determined transitively by that subset, filtering for the initial subsets that eventually cover the entire parameter set, and then filtering that list down to sets that are not a superset of any other result? If so, it's the details of how the various `closeMap`s and `close` functions work that I would really need to stare at in order to check for errors.

Wouldn't it be simpler to do something like
* **frontier**(0) := the single _n_-element set of parameters
* **frontier**(_i_ + 1) := for each set _needed_ in **frontier**(_i_):
  * a **relevant fundep** has inputs a subset of _needed_ and outputs not disjoint with _needed_
  * a parameter _p_ in _needed_ is **determined** iff a **relevant fundep** exists with outputs containing _p_
  * a **determined** _p_ is **acyclically determined** iff no **relevant fundep** exists with inputs containing _p_
  * if no parameters are **determined**, add _needed_ to the set of covering sets
  * otherwise, if there exists an **acyclically determined** _p_, add _needed_ \ { _p_ } to **frontier**(_i_ + 1)
  * otherwise, for every **determined** p, add _needed_ \ { _p_ } to **frontier**(_i_ + 1)

Note that every **frontier**(_i_) consists of (_n_ − _i_)-element sets, so it's easy to see that this recursion has depth _n_ in the worst case. Every covering set thus produced is minimal by construction, since preferentially removing **acyclically determined** parameters means never nondeterministically exploring a search branch with parameters that could have been removed earlier.

With no transitive closures involved, I think an implementation of this would be easier to understand, assuming I have the right idea about what this function is supposed to do! And probably more efficient than anything that starts from a power set?

MonoidMusician:
Yeah, what I had admittedly isn't very optimized. I personally find thinking in terms of closure operators really easy/direct. But I'll keep working through your algorithm, which goes the reverse way I guess (removing parameters instead of adding them). At the moment I don't see why this holds:
> Every covering set thus produced is minimal by construction, since preferentially removing acyclically determined parameters means never nondeterministically exploring a search branch with parameters that could have been removed earlier.

And why do you explore only one acyclically determined parameter at a time?

I'll give it more thought when I can.

Besides the implementation, does it make sense that the “fully determined” arguments are the ones in no (minimal) covering set? I had trouble interpreting it at first, but what it needs to do is to pick out the parameters that will never be used when matching instances. (That's because it's used for allowing rows in instance heads, which we do not want to affect matching.)

Also here's a degenerate case I want to make sure to think about:
```purescript
class SingleInstance (r :: Row Type) | -> r
  unified :: Proxy r

instance SingleInstance ( x :: Unit ) where
  unified = Proxy
  
-- I think this should infer `test :: Proxy ( x :: Unit )`
test = unified
```

rhendric:
> At the moment I don't see why this holds:
> 
> > Every covering set thus produced is minimal by construction, since preferentially removing acyclically determined parameters means never nondeterministically exploring a search branch with parameters that could have been removed earlier.

The acyclically determined parameters, by definition, are the determined parameters that aren't inputs to any relevant fundeps. Removing one from consideration, therefore, doesn't interfere with determining any other determined parameter; you can't get stuck on a bad choice. The same isn't true for other determined parameters, and so those need to be explored nondeterministically in order not to miss any cases.

> And why do you explore only one acyclically determined parameter at a time?

That lets each round of the recursion have the invariant that all the sets are the same size. This invariant is nice because it means an implementation doesn't have to carry around a visited set-of-sets to keep from duplicating work; once the frontier of a certain size is processed, none of those sets are ever produced again.

MonoidMusician:
Okay, it's starting to make more sense now. As further optimization we could just remove all acyclically determined _p_ at once, right?

rhendric:
It wouldn't affect correctness, but it's not an optimization, I think. Consider the fundeps `a -> b c d, a b -> c, a c -> b, d -> a b c e, e -> d`. Starting with set `{a, b, c, d, e}`, both `a` and `b` are determined, but no parameters are determined acyclically. So nondeterministically choose: in branch A, remove `a`, and in branch B, remove `b`. (There are other branches but they don't affect this argument.)

In branch A, now the relevant fundeps are `d -> b c e, e -> d`, and `b` and `c` are now both acyclically determined. Let's say we remove them both at the same time and provide `{d, e}` as a set in the next frontier.

Meanwhile, in branch B, the relevant fundeps are `a -> c d, d -> a c e, e -> d`. With `b` no longer needed, now `c` is acyclically determined, and it's the only one. So `c` is removed and `{a, d, e}` is provided to the next frontier.

In the following pass, both `{d, e}` and `{a, d, e}` are explored. In branch A, `{d, e}` is nondeterministically reduced, and if you imagine a larger set of parameters you can imagine that this involves a lot of work. In branch B, `{a, d, e}` is nondeterministically reduced, and one of those branches means that `{d, e}` is included in the next pass. That means that the next pass will redo the work being done in this pass in branch A.

You could get around this by keeping a visited set-of-sets, as is common in graph searches. But instead, we can exploit the leveled structure of this graph, with each edge going from the level containing _i_-element sets to the level containing (_i_&nbsp;−&nbsp;1)-element sets. If we always determine one parameter at a time, no branch ‘follows’ any other branch in this way, and simply ensuring that the frontier is a set is enough to prevent duplication of work.

---

If it helps, here's a demo implementation—feel free to use as little or as much of this as you like.

```haskell
type Frontier = Map IntSet (First (IntMap (NonEmpty IntSet)))
--                    ^              ^        ^       ^
--    when *these* parameters        |        |       |
--    are still needed,              |        |       |
--                      *these* parameters    |       |
--                      can be determined     |       |
--                               from a non-zero      |
--                               number of fundeps,   |
--                                         which accept *these*
--                                         parameters as inputs.

coveringSets :: Int -> [([Int], [Int])] -> Set IntSet
coveringSets nargs deps = fst $ search $
  M.singleton
    (IS.fromList [0 .. nargs - 1])
    (First $ IM.fromListWith (<>) [(y, xs') | (xs, ys) <- deps, let xs' = IS.fromList xs :| [], y <- ys])

  where

  search :: Frontier -> (Set IntSet, ())
  search frontier = unless (null frontier) $ M.foldMapWithKey step frontier >>= search

  step :: IntSet -> First (IntMap (NonEmpty IntSet)) -> (Set IntSet, Frontier)
  step needed (First inEdges)
    | IM.null inEdges = (S.singleton needed, M.empty)
    | otherwise       = (S.empty, foldMap removeParameter paramsToTry)

      where

      determined = IM.keys inEdges
      acycDetermined = find (`IS.notMember` (IS.unions $ concatMap NEL.toList $ IM.elems inEdges)) determined
      paramsToTry = maybe determined pure acycDetermined

      removeParameter :: Int -> Frontier
      removeParameter y =
        M.singleton
          (IS.delete y needed)
          (First $ IM.mapMaybe (nonEmpty . NEL.filter (y `IS.notMember`)) $ IM.delete y inEdges)
```

The frontier here is a `Map (Set IntSet) <relevant fundeps>` instead of a `Set IntSet`, but that's just an optimization to keep from having to filter out the same irrelevant fundeps over and over; the values could instead be generated from the keys and the full set of fundeps. The `First`s in there aren't load-bearing; they're only there to emphasize that the monoidal structure of the frontier maps is `union`, not `unionWith (<>)` (because, again, the values could be generated from the keys, so we only need one), and to keep this implementation correct if the `Monoid` instance of `Map` is ever changed (or if `MonoidalMap` is subbed in here instead). Similarly, the `NonEmpty` stuff is more about keeping the types honest than about any actual need.

I whipped this up in isolation so there may be other issues dropping this into PureScript, but hopefully it communicates a little more about how I imagine the algorithm working.

MonoidMusician:
Ah right, I was missing the memoization. That type structure looks pretty close to what I was going towards too, didn't check the details yet but I should be able to put it together with what I have.

MonoidMusician:
Hey it looks good, I plugged it in to the compiler. There's only one test failure:

```purescript
class Cyclic a b c d | a -> b, b -> a
                     , a -> c
                     , c -> d, d -> c
instance cyclic :: Cyclic Empty Empty {} {}
```

This is kind of what I suspected could go wrong: I think we're missing out on some of the synergy of the fundeps by solving them in reverse like this. I'll keep poking at it, not sure if there's an easy fix (that doesn't involve e.g. precomputing cycles or other graph information).

rhendric:
Oh dang! Yeah, you got me there. I'll think about it too but that's a strong point in favor of your original approach.

MonoidMusician:
Okay so working through the steps on what should happen for this example with your algorithm:
```
a -> b, b -> a,
a -> c,
c -> d, d -> c

{abcd}
{abc,abd,acd,bcd}
{bc,ac,ab}<>{bd,ad}<>{ad,ac}<>{bd,bc} = {ab,ac,ad,bc,bd}
{a,b}<>{a}<>[AD]<>[BC]<>[BD]
```
It looks like `ad`, `bc`, and `bd` just get stuck, because they cannot see the transitive dependencies. That makes the algorithm think they are covers. However, the minimal solutions `a` and `b` are still found, and they should be used to filter out the stuck ones. So I think at each step, we just need to filter the potential covers as not being supersets of continuing solutions. That is, the claim that all solutions are explored still seems to hold, just not the minimality.

I added the check and tests pass! I think I'm satisfied that it works now.

I want to add some comments before merging. And a changelog entry. Anything else?

rhendric:
My turn to play the adversary, I'm afraid:

```
a b -> d,
d -> a,
c d -> b,
c -> d

{abcd}
{abc,acd,bcd}
[ABC]<>{cd}<>{cd} = [ABC], {cd} -- no new frontier is a subset of ABC, so ABC gets through
{c}
[C]
```

It would seem that the entire set of emitted provisional covers needs to be checked at every pass, not just the covers newly emitted in the current pass. At that point, might be better instead to do a single filter at the end like you had originally.

MonoidMusician:
I would expect to see `ac` and `bc` from `acd` and `bcd` under `c -> d`, unless this is the acyclic stuff firing? I guess that would mean that we might need to select one mechanism for finding minimal covers, not mix and match. Or just check at the end like you suggested. Not sure what's better.

MonoidMusician:
You know what, I don't think we gain anything by doing it stage-by-stage, it's probably about the same number of comparisons, so let's just check at the end. I'll test and push.

MonoidMusician:
No merge conflicts, tests still run. Can I get an approval to merge? 🙂

rhendric (approved):

MonoidMusician:
Thanks, I really appreciate your help on this! I learned some things along the way :)

PR #4196, opened 2021-11-01
Can't solve constraints in values which came from constructor binders where the constructor was polymorphic

jy14898:
Before this change, you couldn't solve constraints in values which came from constructor binders where the constructor was polymorphic:

```haskell

class Con

newtype Identity a = Identity a

-- works
test1 :: Con => (Con => Int) -> Int
test1 a = a

-- fails
test2 :: Con => Identity (Con => Int) -> Int
test2 (Identity a) = a
```

This is because we weren't applying substitutions when checking the types of 'Var' exprs before subsumption, and polymorphic types like Identity get instantiated with unknowns during binder inference. This change adds the substitution in.

**Checklist:**

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -710,7 +710,8 @@ check' v@(Var _ var) ty = do
   checkVisibility var
   repl <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< lookupVariable $ var
   ty' <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< replaceTypeWildcards $ ty
-  elaborate <- subsumes repl ty'
+  currentSubst <- gets checkSubstitution
+  elaborate <- subsumes (substituteType currentSubst repl) ty'
   return $ TypedValue' True (elaborate v) ty'
 check' (DeferredDictionary className tys) ty = do
   {-
--- a/tests/purs/failing/3765.out
+++ b/tests/purs/failing/3765.out
@@ -4,31 +4,32 @@ at tests/purs/failing/3765.purs:6:23 - 6:24 (line 6, column 23 - line 6, column 
 
   Could not match type
   [33m            [0m
-  [33m  ( b :: Int[0m
+  [33m  ( a :: Int[0m
   [33m  ...       [0m
   [33m  | t0      [0m
   [33m  )         [0m
   [33m            [0m
   with type
   [33m            [0m
-  [33m  ( a :: Int[0m
+  [33m  ( b :: Int[0m
   [33m  ...       [0m
   [33m  | t0      [0m
   [33m  )         [0m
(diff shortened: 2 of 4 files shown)


garyb (approved):
I can't think of a reason why there would be a problem doing this. Those revised error output seem like improvements too! :+1:

rhendric:
How exactly does this get used? I've tried adding:

```purs
instance Con

x = test (Identity (2 :: Con => Int))
```
and I get:
<pre>         Could not match constrained type
         <font color="#C3A000">            </font>
         <font color="#C3A000">  Con =&gt; Int</font>
         <font color="#C3A000">            </font>
         with type
         <font color="#C3A000">     </font>
         <font color="#C3A000">  Int</font>
         <font color="#C3A000">     </font>
       
       while trying to match type <font color="#C3A000">Int</font>
         with type <font color="#C3A000">Con =&gt; Int</font>
       while checking that expression <font color="#C3A000">Identity 2</font>
         has type <font color="#C3A000">Identity (Con =&gt; Int)</font>
       in value declaration <font color="#C3A000">main</font>
</pre>

Is there a different way to construct values with constrained types that can be used in polymorphic constructors, and if not, is this feature then incomplete?

jy14898:
> How exactly does this get used?

Good point, originally I was just messing around with deferring execution of expressions automatically (similar to https://github.com/natefaubion/purescript-call-by-name I believe), and found it surprising that using Identity would not work.

I was using FFI to construct these values which meant I never actually tested creating them in-language.

I'll have a look to see if it makes sense to allow constructing such values naturally

jy14898:
@rhendric It's possible, although you have to be explicit about the type of `Identity` when constructing it:

```haskell
-- compiles fine in the pull req, with inferred type Con => Int
test2' = test2 ((Identity :: (Con => Int) -> Identity (Con => Int)) 10)
```

The changes required to not need that hint are too great, I think

rhendric:
What about:

```purs
test :: Con => { foo :: Con => Int } -> Int
test x = x.foo
```

Is there a way to annotate that to make it work?

Then there's this little puzzle:
```purs
-- doesn't work
unid1 :: Identity (Con => Int) -> (Con => Int)
unid1 = coerce

-- works
unid2 :: Identity (Con => Int) -> (Con => Int)
unid2 = coerce :: forall a. Identity a -> a
```

I guess I'm broadly concerned that the whole impredicative types question hasn't been exhaustively thought out in PureScript (considering interactions with function types, data constructors and destructors, records, arrays, anything else that gets somewhat special handling that I'm missing right now), and even though this seems like a harmless and, to a limited audience, beneficial enhancement to type inference, I'm not confident that it implements, or gets us closer to, a set of inference rules that would be straightforward to explain to a learner. It could easily be that in that respect, this change doesn't make the situation any worse—I'm just asking the broad question and hoping that someone with a stronger type theory background can say that it does or doesn't.

PR #4197, opened 2021-11-04
Expand Layout.hs module docs

JordanMartinez:
**Description of the change**

Part of #4192. I tried giving a more detailed explanation, so that others can understand this faster. I still don't understand the code well enough to edit it safely (if that was ever needed to support new syntax), but I have a better understanding as to why it exists now.

### DIFF
--- a/lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
@@ -1,9 +1,9 @@
 -- |
 -- ## High-Level Summary
 --
--- Other than this paragraph, this section was written by Nate, the author
--- of most of this code. For those who know more about compiler-development,
--- the below explanation is likely enough. For everyone else, see the next section.
+-- This section provides a high-level summary of this file. For those who
+-- know more about compiler-development, the below explanation is likely enough.
+-- For everyone else, see the next section.
 --
 -- The parser itself is unaware of indentation, and instead only parses explicit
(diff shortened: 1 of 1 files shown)


lib/purescript-cst/src/Language/PureScript/CST/Layout.hs
```diff
@@ -17,7 +24,144 @@
...
+--                  =
+--              5 in baz
+-- @
+--
+-- Each block of code might appear to the parser as a stream of the
+-- following source tokens where the @\{@ sequence represents
+-- @TokLayoutStart@ and the @\}@ sequence represents @TokLayoutEnd@:
+-- - @\{let\{foo\}=\{5\}\}@
+-- - @\{let\{bar\}=\{5\}\}@
+-- - @\{let\{baz\}=\{5\}\}@
```
JordanMartinez, line 64: This is likely wrong as to how many of these tokens are inserted and where they are inserted. My goal here was to simply show that they are being inserted and that whitespace, comments, and lines are irrelevant.

natefaubion: It would be
```
let \{foo = 5\}
```
I think you should include multiple definitions, eg:
```
let \{foo = 5\;bar = 42\} in ...
```

JordanMartinez: Done.

JordanMartinez:
@natefaubion Any other feedback on this? Or is this good to merge?

natefaubion (approved):

PR #4199, opened 2021-11-05
Bump PureScript to building with LTS-18

cdepillabout:
**Description of the change**

I've bumped the resolver in `stack.yaml` to build with LTS-18 instead of LTS-17.

This seems to be a very easy bump, and there was very little breakage.

---

**Checklist:**


cdepillabout:
I wanted this because I wanted to get PureScript compiling in Nixpkgs.

In the past, I think there has been some resistance to keeping up with the latest LTS release, since it causes annoying churn for the maintainers.  So I'm not sure what the reaction will be to this PR.  But I needed it for Nixpkgs, so I thought I would at least share it here as well.

Two things left to do:

- add a note to the changelog
- make sure to follow the comment in `purescript.cabal` about the correct way for bumping dependencies after doing an LTS update.

cdepillabout:
It looks like the lint workflow is failing because it can't install `weeder`: https://github.com/purescript/purescript/runs/4116464191?check_suite_focus=true

Here's the problem:

- Recent versions of `weeder` have a weird versions dep specification for `dhall`.  They require basically any version of `dhall` _except_ 1.39.
- LTS-18 contains `dhall-1.39`.
- Stack can't find a version of a version of `dhall` it can build `weeder` with and gives up.

I'll look into seeing if there is a good fix for this.

cdepillabout:
This should be ready for a review now.

rhendric:
We only pinned to `weeder-2.2.0` because of #4189. Would using a later version of Weeder make that part of this PR simpler?

JordanMartinez:
`weeder-2.3.0` depends on `base (>=4.15.0.0 && <4.16)`. Since I was only fixing the CI issue to make the release, I wasn't trying to upgrade to a newer GHC as well. If that's what we're doing here, then upgrading weeder to 2.3.0 sounds fine to me.

That being said, will updating the dependencies cause any changes to the JSON codecs?

cdepillabout:
> Would using a later version of Weeder make that part of this PR simpler?

@rhendric Like Jordan says, the latest version of `weeder` is 2.3.0, and it depends on `base-4.15.0.0`, which appears to be shipped with ghc-9.0.1 (per https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history).

I didn't actually try it, but my guess is that installing weeder like the following would fail:

```console
$ echo 'allow-newer: true' >> stack.yaml
$ stack --no-terminal --jobs=2 build --copy-compiler-tool weeder-2.3.0
```

> I wasn't trying to upgrade to a newer GHC as well. If that's what we're doing here, then upgrading weeder to 2.3.0 sounds fine to me.

This updates from LTS-17 (which uses GHC-8.10.4) to LTS-18 (which uses GHC-8.10.7).  But weeder-2.3.0 requires GHC-9.0, which is currently only available in Stackage Nightly.  So I don't think we would be easily able to move to weeder-2.3.0 in this PR.

cdepillabout:
> That being said, will updating the dependencies cause any changes to the JSON codecs?

This sounds like something pretty important to check!

I assumed there would be tests for this, but maybe there aren't many?  Any suggestions how I should go about testing for this?

JordanMartinez:
> I assumed there would be tests for this, but maybe there aren't many? Any suggestions how I should go about testing for this?

I believe there are tests for this? I was asking because I had a recent experience where derived JSON codecs broke something. However, I believe upgrading here means we'd also need to upgrade Pursuit.

purescript.cabal
```diff
@@ -119,66 +119,66 @@ common defaults
...
-    stm >=2.5.0.0 && <2.6,
+    stm >=2.5.0.1 && <2.6,
     stringsearch >=0.3.6.6 && <0.4,
     syb >=0.7.2.1 && <0.8,
     text >=1.2.4.1 && <1.3,
     these >=1.1.1.1 && <1.2,
     time >=1.9.3 && <1.10,
     transformers >=0.5.6.2 && <0.6,
-    transformers-base >=0.4.5.2 && <0.5,
+    transformers-base >=0.4.6 && <0.5,
```
JordanMartinez, line 177: [Nothing to note](https://github.com/mvv/transformers-base/commits/master)

purescript.cabal
```diff
@@ -119,66 +119,66 @@ common defaults
...
     transformers-compat >=0.6.6 && <0.7,
-    unordered-containers >=0.2.13.0 && <0.3,
+    unordered-containers >=0.2.14.0 && <0.3,
```
JordanMartinez, line 179: [Fixed type class instances for `NFData1` and `NFData2`](https://hackage.haskell.org/package/unordered-containers-0.2.14.0/changelog). Not sure how relevant that is to us as I think we only use `NFData`

purescript.cabal
```diff
@@ -119,66 +119,66 @@ common defaults
...
     utf8-string >=1.0.2 && <1.1,
-    vector >=0.12.1.2 && <0.13
+    vector >=0.12.3.1 && <0.13
```
JordanMartinez, line 181: [Nothing of note](https://hackage.haskell.org/package/vector-0.12.3.1/changelog)

JordanMartinez (commented):
I looked through each changed dependency's changelog. Most were irrelevant (see comments in this review), but the ones I thought worth bringing up were:
- `language-javascript@0.7.1.0` - [while it just adds support for `async` and `await` keywords](https://hackage.haskell.org/package/language-javascript-0.7.1.0/changelog), I believe we specifically pin it to `0.7.0.0` but I can't recall why.
- [`scientific` made a change to a `Show` instance. I'm not sure how relevant that is to us](https://hackage.haskell.org/package/scientific-0.3.7.0/changelog)
- [`unordered-containers` fixed some instances for the `NFData1` and `NFData2` type classes](https://hackage.haskell.org/package/unordered-containers-0.2.14.0/changelog). Not sure how relevant that is to us as I think we only use `NFData`

purescript.cabal
```diff
@@ -119,66 +119,66 @@ common defaults
...
+    file-embed >=0.0.15.0 && <0.1,
     filepath >=1.4.2.1 && <1.5,
     fsnotify >=0.3.0.1 && <0.4,
     Glob >=0.10.1 && <0.11,
-    haskeline >=0.8.1.1 && <0.9,
-    language-javascript ==0.7.0.0,
-    lifted-async >=0.10.1.3 && <0.11,
+    haskeline >=0.8.2 && <0.9,
+    language-javascript ==0.7.1.0,
+    lifted-async >=0.10.2.2 && <0.11,
```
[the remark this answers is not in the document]
rhendric: We started pinning to an exact version of `language-javascript` because of https://github.com/purescript-web/purescript-web-html/issues/31; we didn't want some people's PS compilers supporting some foreign code that other people's PS compilers didn't support. I don't think there'd be a problem with bumping to the next minor version provided we trust the semantic versioning of `language-javascript` and exercise due diligence; that package is basically part of the compiler's interface to users.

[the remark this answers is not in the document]
JordanMartinez: Ah, thanks.
[Here's all the changes made between `0.7.0.0` and `0.7.1.0`](https://github.com/erikd/language-javascript/compare/de4db524e7987436d618b0b53a38266727414955..376b83afdba022da1462f9814e63f2c5a2b15120). As far as I can tell, this should be safe to do.

purescript.cabal
cdepillabout: From the changelog, it looks like there were just new instances added for `NFData1` and `NFData2`: https://github.com/haskell-unordered-containers/unordered-containers/pull/314/files

I'd be pretty surprised if that affected PureScript.

purescript.cabal
```diff
@@ -119,66 +119,66 @@ common defaults
...
+    process >=1.6.13.2 && <1.7,
     protolude >=0.3.0 && <0.4,
     purescript-cst ==0.4.0.0,
-    regex-tdfa >=1.3.1.0 && <1.4,
+    regex-tdfa >=1.3.1.1 && <1.4,
     safe >=0.3.19 && <0.4,
-    scientific >=0.3.6.2 && <0.4,
-    semigroups >=0.19.1 && <0.20,
+    scientific >=0.3.7.0 && <0.4,
+    semigroups >=0.19.2 && <0.20,
```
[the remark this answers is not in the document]
cdepillabout: It looks like the `scientific` PR is: https://github.com/basvandijk/scientific/pull/68

This looks like a relatively harmless change.

The only places it appears PureScript uses `Data.Scientific` is in lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs and lib/purescript-cst/src/Language/PureScript/PSString.hs, but these appear to be reading/lexing scientific values, not printing them using the `Show` instance.

purescript.cabal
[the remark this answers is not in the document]
cdepillabout: Here's the PR that adds support for `async` and `await`: https://github.com/erikd/language-javascript/pull/113

It looks like a new constructor is added to an existing data type, which I thought required a major version bump according to the PVP.  I decided to ask on the Haskell discourse about this: https://discourse.haskell.org/t/does-adding-a-data-constructor-require-a-major-version-bump-according-to-the-pvp/3600

cdepillabout:
@JordanMartinez @rhendric Thanks for taking a look at this.  Let me try to summarize what I think are the remaining problems with this PR:

- [ ] Figure out what to do about `language-javascript`.  Above, @rhendric said "\[`language-javascript`\]  is basically part of the compiler's interface to users".  This makes me think that the easiest thing to do is be conservative and just keep `language-javascript` at 0.7.0.0 instead of bumping it.
- [ ] I realized I didn't actually do the whole sdist generation and copying over version constraints for `purescript-cst`.  I guess I should probably do this?

JordanMartinez:
I think keeping `language-javascript` at `0.7.0.0` and then addressing that issue separately in another PR would be best as it won't cause this one to be held up. I would like to bump it at some point as it does enable parsing more code in one's FFI.

JordanMartinez:
> I guess I should probably do this?

Yeah, please do that and let us know what (if anything) changes.

JordanMartinez:
Actually, we definitely cannot update to `0.7.1.0` yet due https://github.com/purescript-web/purescript-web-html/issues/31. I attempted to compile the latest package set with this PR, and I get the error mentioned in that issue. Updating to `0.7.1.0` is a breaking change:
```
[1/1 ErrorParsingFFIModule] .spago/web-html/v3.1.0/src/Web/HTML/HTMLScriptElement.purs:1:1

      v
   1  module Web.HTML.HTMLScriptElement where
   2  
   3  import Data.Maybe (Maybe)
      ...
  73  
  74  foreign import text :: HTMLScriptElement -> Effect String
  75  foreign import setText :: String -> HTMLScriptElement -> Effect Unit
                                                                         ^
  
  Unable to parse foreign module:
  
    .spago/web-html/v3.1.0/src/Web/HTML/HTMLScriptElement.js
  
    The module could not be parsed:
  
    AsyncToken {tokenSpan = TokenPn 1062 57 30, tokenLiteral = "async", tokenComment = []}

           Src   Lib   All
Warnings   0     536   536
Errors     0     1     1  
[error] Failed to build.
```
Still, I am curious why we get the above error....

garyb:
> Still, I am curious why we get the above error....

Because `async` became a reserved word, and https://github.com/purescript-web/purescript-web-html/blob/3a249b966ee72c19874b4a2ec6db4059087500e4/src/Web/HTML/HTMLScriptElement.js#L51-L55

JordanMartinez:
Ah! Thanks for clarifying.

garyb:
I wonder if that's a bug in `language-javascript` actually - it may well be a meaningful word now, but is it actually a syntax error to use it as a object property name? If so, how is it possible for `async` to be a property on the DOM object also? Unless it requires `obj["async"]` to be talked about in later ES versions. 🤔

JordanMartinez:
:thinking: 

Regardless, I've verified that downgrading this PR's `language-javascript` to `0.7.0.0` and then running the following works
```bash
stack build
mkdir zlts
pushd zlts
stack exec bash <<EOF
spago build
spago docs -S
EOF
popd
rm -rf zlts/
```

MonoidMusician:
`async` should not be reserved in `script.async`, I think basically all keywords in JS can be used in that context (and as keys in object literals), so that's a bug in language-javascript.

purescript.cabal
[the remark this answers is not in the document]
cdepillabout: I've reported this upstream: https://github.com/erikd/language-javascript/issues/130

cdepillabout:
I've taken the `language-javascript` PVP problem and the `async` parsing problem and reported them upstream:

- https://github.com/erikd/language-javascript/issues/131
- https://github.com/erikd/language-javascript/issues/130

cdepillabout:
Alright, I've dropped language-javascript back down to 0.7.0.0, and regenerated the version constraints in `lib/purescript-cst/purescript-cst.cabal`.

I think this should be ready to be merged in!

stack.yaml
```diff
@@ -7,13 +7,16 @@ ghc-options:
   # Build with advanced optimizations enabled by default
   "$locals": -O2 -Werror
 extra-deps:
+# As of 2021-11-08, the latest releease of `language-javascript` is 0.7.1.0,
```
JordanMartinez, line 10:
```diff
-# As of 2021-11-08, the latest releease of `language-javascript` is 0.7.1.0,
+# As of 2021-11-08, the latest release of `language-javascript` is 0.7.1.0,
```

JordanMartinez (approved):
LGTM. Thanks for your work!

.github/workflows/ci.yml
```diff
@@ -111,8 +111,16 @@ jobs:
...
+      # of GHC.
+      #
+      # weeder-2.2.0 has somewhat strange version deps. It doesn't appear
+      # to support the versions of dhall and generic-lens in LTS-18.  However,
+      # forcing it to use the versions of dhall and generic-lens in LTS-18
+      # doesn't cause any problems, so the following commands build weeder
+      # while ignoring version constraints.
+      - run: |
+          echo 'allow-newer: true' >> stack.yaml
+          stack --no-terminal --jobs=2 build --copy-compiler-tool weeder-2.2.0
```
rhendric, line 123: This `allow-newer: true` change applies to every `stack` invocation downstream of this step, too. This is probably fine—any issues obscured by ignoring package versions would likely be caught in one of the build jobs—but structuring the workflow like this makes it non-obvious. Please choose one:

* Make the `allow-newer: true` apply only to this invocation of `stack`, probably by creating a copy of `stack.yaml` and either restoring it afterward or using the `--stack-yaml` option on this invocation (although if you know a cleaner way, be my guest!)
* Pull the `allow-newer` modification into its own step, and move the explanatory comment there.

In either case, if the `run` property continues to span multiple lines, please provide a friendly `name` property for the step.

stack.yaml
cdepillabout: Thanks, fixed in 4e93cb93.

.github/workflows/ci.yml
cdepillabout: I noticed this when wrote the above code, but I thought it wouldn't be likely to negatively affect the following calls to `stack`.

Although I'm not that confident about that, so adding `allow-newer: true` only for installing weeder sounds like the safest thing to do.  Using a separate `stack.yaml` for installing weeder sounds like a good idea, so I made this change in 748aefb3.

rhendric (approved):

cdepillabout:
I'm not sure anyone would be interested, but I specifically wanted this PR so that I could get PureScript building in Nixpkgs again: https://github.com/NixOS/nixpkgs/pull/145310.

juhp:
I am also trying to get purescript packaged in Fedora (currently based on lts-18). :+1: 
Thanks!

PR #4205, opened 2021-11-12
Inline newtype constructor applications with ($)

### LINKED ISSUE #2866: Unnecessary newtype constructor application with ($)
Hi,

I raised this on gitter and it was suggested it may be a bug.

I created a `newtype` [here](https://github.com/craigbilner/pbee/blob/master/ch2/src/Main.purs) called `Area` which is returned from the `circleArea` function.

```
circleArea :: Radius -> Area
circleArea (Radius r) = Area $ pi * r `pow` 2.0
```

when I ran `pulp build -O --to output.js`

it turned it into

```
var Area = function (x) {
  return x;
};
  
var circleArea = function (v) {
  return Area($$Math.pi * $$Math.pow(v)(2.0));
};
```

but I was expecting there to be no `Area` func at all, just the value passed. Is this a bug or my misunderstanding?

PureFunctor:
**Description of the change**
This change makes it so that newtype application using the `($)` operator is inlined, which fixes #2866.

I'm not entirely sure how to add tests for this though, or for codegen code in general.

---

**Checklist:**

~~[ ] Added myself to CONTRIBUTORS.md (if this is my first contribution)~~
~~[ ] Updated or added relevant documentation~~

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -42,8 +42,6 @@ import qualified Language.PureScript.Constants.Prim as C
 
 import System.FilePath.Posix ((</>))
 
-import System.IO.Unsafe (unsafePerformIO)
-
 -- | Generate code in the simplified JavaScript intermediate representation for all declarations in a
 -- module.
 moduleToJs
@@ -234,25 +232,24 @@ moduleToJs (Module _ coms mn _ imps exps reExps foreigns decls) foreign_ =
     return $ AST.Function Nothing Nothing jsArg (AST.Block Nothing [AST.Return Nothing ret])
   valueToJs' e@App{} = do
     let (f, args) = unApp e []
-    args' <- mapM valueToJs args
-    let _ = unsafePerformIO (print (unApp (head args) (tail args)) >> pure f)
     case f of
       Var _ (Qualified _ (Ident "apply")) -> do
-        let (g, args'') = unApp (head args) (tail args)
-        args''' <- mapM valueToJs args''
-        case g of
-          Var (_, _, _, Just IsNewtype) _ -> return (head args''')
-          Var (_, _, _, Just (IsConstructor _ fields)) name | length args'' == length fields ->
-            return $ AST.Unary Nothing AST.New $ AST.App Nothing (qualifiedToJS id name) args'''
-          _ -> flip (foldl (\fn a -> AST.App Nothing fn [a])) args''' <$> valueToJs f
-      Var (_, _, _, Just IsNewtype) _ -> return (head args')
-      Var (_, _, _, Just (IsConstructor _ fields)) name | length args == length fields ->
-        return $ AST.Unary Nothing AST.New $ AST.App Nothing (qualifiedToJS id name) args'
-      _ -> flip (foldl (\fn a -> AST.App Nothing fn [a])) args' <$> valueToJs f
+        let (f', args') = unApp (head args) (tail args)
+        mkApp f' args'
+      _ -> mkApp f args
     where
     unApp :: Expr Ann -> [Expr Ann] -> (Expr Ann, [Expr Ann])
     unApp (App _ val arg) args = unApp val (arg : args)
     unApp other args = (other, args)
+
+    mkApp f args = do
+      args' <- mapM valueToJs args
+      case f of
+        Var (_, _, _, Just IsNewtype) _ -> return (head args')
+        Var (_, _, _, Just (IsConstructor _ fields)) name | length args == length fields ->
+          return $ AST.Unary Nothing AST.New $ AST.App Nothing (qualifiedToJS id name) args'
+        _ -> flip (foldl (\fn a -> AST.App Nothing fn [a])) args' <$> valueToJs f
+
   valueToJs' (Var (_, _, _, Just IsForeign) qi@(Qualified (Just mn') ident)) =
     return $ if mn' == mn
              then foreignIdent ident
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -245,6 +245,7 @@ moduleToJs (Module _ coms mn _ imps exps reExps foreigns decls) foreign_ =
     unApp (App _ val arg) args = unApp val (arg : args)
     unApp other args = (other, args)
 
+    mkApp :: Expr Ann -> [Expr Ann] -> m AST
     mkApp f args = do
       args' <- mapM valueToJs args
       case f of
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -42,6 +42,8 @@ import qualified Language.PureScript.Constants.Prim as C
 
 import System.FilePath.Posix ((</>))
 
+import System.IO.Unsafe (unsafePerformIO)
+
 -- | Generate code in the simplified JavaScript intermediate representation for all declarations in a
 -- module.
 moduleToJs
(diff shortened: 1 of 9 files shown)


src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -234,25 +232,24 @@ moduleToJs (Module _ coms mn _ imps exps reExps foreigns decls) foreign_ =
...
-        case g of
-          Var (_, _, _, Just IsNewtype) _ -> return (head args''')
-          Var (_, _, _, Just (IsConstructor _ fields)) name | length args'' == length fields ->
-            return $ AST.Unary Nothing AST.New $ AST.App Nothing (qualifiedToJS id name) args'''
-          _ -> flip (foldl (\fn a -> AST.App Nothing fn [a])) args''' <$> valueToJs f
-      Var (_, _, _, Just IsNewtype) _ -> return (head args')
-      Var (_, _, _, Just (IsConstructor _ fields)) name | length args == length fields ->
-        return $ AST.Unary Nothing AST.New $ AST.App Nothing (qualifiedToJS id name) args'
-      _ -> flip (foldl (\fn a -> AST.App Nothing fn [a])) args' <$> valueToJs f
+        let (f', args') = unApp (head args) (tail args)
```
PureFunctor, line 237: I'm not very confident with using `head args` here, since it assumes that `args` always has an element because of `"apply"` – but I don't think this is necessarily a problem, since the `IsNewtype` case assumes it (or at least `args'`) to have an element either way.

rhendric:
> I'm not entirely sure how to add tests for this though, or for codegen code in general.

Feel free to extract and introduce the `tests/purs/optimize` machinery I built in #3915 for this sort of thing. I don't think that one's getting merged any time soon.

rhendric:
I haven't the time right this moment to consider a constructive alternative, but as initial feedback I think it's a little unfortunate to duplicate the `$` inlining both here and in `L.PS.CoreImp.Optimizer.Inliner` downstream. The latter is much more extensive (in particular, it covers `#` as well), and ideally I think we'd want all of our inlining optimizations in one place?

So, for example, one thing I almost proposed to you now was to replace newtype constructors here with `Control.Category.identity(Control.Category.categoryFn)`, and inline saturated calls to that `identity` in the `L.PS.CoreImp.Optimizer.Inliner` machinery. That would cover both `MyNewtype $ x` and `x # MyNewtype`, as well as some less probable combinations of constructors and inline-able operators. With only a tad more work we could probably cover `>>> MyNewtype` etc. as well.

The reason I'm not proposing precisely that is that it introduces a dependency on `Control.Category`, and thus implicitly `purescript-prelude`, in any project that uses a newtype constructor if that constructor isn't in a position that gets optimized away. Projects using an alternate prelude might break at run time.

As I said, I'd need more time than I have right this moment to come up with a better version of that idea, but maybe this inspires a good idea in you or someone else, or a counterargument to one of my thoughts.

PureFunctor:
> I think we'd want all of our inlining optimizations in one place?

I agree. I had not considered the CoreImp inliner ahead of time because I'm still not really familiar with that part of the code base.

Though I figured the CoreFn would be another likely candidate to put this optimization into, it wouldn't really be the best move to make considering the reasons you've pointed out.

Thanks a lot, I'll take those pointers in mind.

PureFunctor:
With CoreImp though, I'm not sure if there's a simple way of determining whether or not some `Var` is a newtype function, unless I'm missing something...

My best assumption at the moment is that newtype functions likely start with capitalized names, which is rather fragile.

nwolverson:
I didn't see mention of the fact that non-newtype constructors are also compiled differently with `$` just now (though I doubt it matters much).

```purescript
testNonNt1 = NonNT 42
testNonNt2 = NonNT $ 42
```
```js
var testNonNt2 = NonNT.create(42);
var testNonNt1 = new NonNT(42);
```

As the inlining of `$` and `#` affect codegen, rather than having to care about it here and also in CoreImp optimisations, I'd suggest it might better be part of the CoreFn optimisation pass. I'm doing that in `purerl` as it affects codgen even more in my case

rhendric:
Yeah, I think I agree with you. Most of the stuff being inlined in `CoreImp` is inlined to somewhat backend-specific stuff, except for `$` and `#`, which appear to be there as legacy from before there was a `CoreFn` pass. So I think you both are right about this being better as a `CoreFn` optimization; it probably does make sense to move (assuming no unexpected breaks) those specific inlinings out of `CoreImp.Optimizer.Inliner` and into `CoreFn.Optimizer`, which should be sufficient to fix #2866.

PureFunctor:
I think it'd be best to write up a proposal if we ought to move some `CoreImp` optimizations, given that feedback from downstream consumers of `CoreFn` API would be pretty invaluable. In the meantime, I think this PR can be put on hold until further feedback is received.

Likewise, I think this is one of the first few steps in creating a general optimization API for the compiler that isn't backend/codegen-specific, which would prove to be pretty helpful considering that the Core

PureFunctor:
I'm not sure what to do with the weeder errors here.

src/Language/PureScript/CoreFn/Optimizer.hs
```diff
@@ -48,9 +51,17 @@ closedRecordFields (TypeApp _ (TypeConstructor _ C.Record) row) =
...
 -- | See https://github.com/purescript/purescript/issues/3157
-optimizeUnusedPartialFn :: Expr a -> Expr a
+optimizeUnusedPartialFn :: Expr Ann -> Expr Ann
 optimizeUnusedPartialFn (Let _
   [NonRec _ UnusedIdent _]
   (App _ (App _ (Var _ (Qualified _ UnusedIdent)) _) originalCoreFn)) =
   originalCoreFn
 optimizeUnusedPartialFn e = e
+
+-- | TODO: Fixup the annotations here.
```
PureFunctor, line 61: The annotation here seems mostly correct, as this currently points to the function application as a whole.

src/Language/PureScript/CoreFn/Optimizer.hs
```diff
@@ -48,9 +51,17 @@ closedRecordFields (TypeApp _ (TypeConstructor _ C.Record) row) =
...
+optimizeDataFunctionApply :: Expr Ann -> Expr Ann
+optimizeDataFunctionApply e = case e of
+  (App a (App _ (Var _ (Qualified (Just (ModuleName dataFunction)) (Ident applyFn))) x) y)
+    | dataFunction == "Data.Function" && applyFn == "apply" -> App a x y
+    | dataFunction == "Data.Function" && applyFn == "applyFlipped" -> App a y x
```
PureFunctor, line 66: It seems like this is enough to inline most usages of `apply`/`applyFlipped`.

src/Language/PureScript/CoreFn/Optimizer.hs
```diff
@@ -48,9 +51,17 @@ closedRecordFields (TypeApp _ (TypeConstructor _ C.Record) row) =
 closedRecordFields _ = Nothing
 
 -- | See https://github.com/purescript/purescript/issues/3157
-optimizeUnusedPartialFn :: Expr a -> Expr a
+optimizeUnusedPartialFn :: Expr Ann -> Expr Ann
```
rhendric, line 54: 😕

src/Language/PureScript/CoreFn/Optimizer.hs
rhendric: Agree, I think you're doing the right thing with annotations.

src/Language/PureScript/CoreFn/Optimizer.hs
```diff
@@ -48,9 +51,17 @@ closedRecordFields (TypeApp _ (TypeConstructor _ C.Record) row) =
...
 optimizeUnusedPartialFn (Let _
   [NonRec _ UnusedIdent _]
   (App _ (App _ (Var _ (Qualified _ UnusedIdent)) _) originalCoreFn)) =
   originalCoreFn
 optimizeUnusedPartialFn e = e
+
+-- | TODO: Fixup the annotations here.
+optimizeDataFunctionApply :: Expr Ann -> Expr Ann
```
rhendric, line 62:
```suggestion
optimizeDataFunctionApply :: Expr a -> Expr a
```

src/Language/PureScript/CoreFn/Optimizer.hs
rhendric: Replace the string literals with constants from `Language.PureScript.Constants.Prelude`, please. (This will also resolve your Weeder errors; the unused declarations it's identifying are the ones you need to use.)

rhendric (changes requested):
There's still time for people to raise concerns on #4206, but I'll be happy with this implementation once the enclosed changes are addressed.

rhendric (approved):

JordanMartinez (approved):
LGTM AFAICT.

JordanMartinez:
Anything else we want to do before merging this?

Issue #3103, opened 2017-10-02
Rename Symbol to String


kcsongor:
`Symbol`s behave exactly like `String`s at the type-level. Imagining a future where PureScript has datatype-promotion with potentially function promotion as well (much like the `singletons` library for Haskell), the fact that these are called differently will likely cause issues, or at least confusion.

What this means in Haskell today, is that data-types that use either `Symbol` or `String` can only be used at their respected levels. In other words, `Symbol`s can only be promoted, and `String`s must be values. The following snippet demonstrates the issue:

```haskell
{-# LANGUAGE DataKinds      #-}
{-# LANGUAGE KindSignatures #-}
{-# LANGUAGE PolyKinds      #-}

module SymbolString where

import GHC.TypeLits

data D1 = D1 String

data D2 = D2 Symbol

data Proxy (a :: k) = Proxy

d1 :: D1
d1 = D1 "foo"


-- Doesn't work:
--d1' :: Proxy ('D1 "foo")
--d1' = Proxy
--
--    • Expected kind ‘String’, but ‘"foo"’ has kind ‘Symbol’

d2 :: Proxy ('D2 "foo")
d2 = Proxy
```

`'D1` has no type-level inhabitants, and `D2` has no value-level inhabitants. `singletons` solves this issue by special-casing `String`, but I think that's not very elegant.

If datatype-promotion is a planned feature, then I think renaming `Symbol` to `String` would make sense sooner rather than later.

Because this would break a lot of existing code, maybe some temporary kind synonym can be introduced that throws a warning?

paf31:
I don't have any plans to implement datatype promotion right now. The `reflection` library gives me the basic bits and pieces that I need to move things back and forth to the type level.

I'd rather not rename things just for the sake of the name, so I think we'd need a good reason. Without another reason, I'd rather wait until we decide to implement promotion.

kcsongor:
Sure, the only reason I thought of this was that it requires nasty workarounds in Haskell, and it would be better if those two were called the same. Other than promotion, I can't think of any reason why this would make sense.

LiamGoodacre:
The risk with waiting until after 1.0 is that we'd then be far more locked in to the name `Symbol`, with more and more people/libraries depending on it.

paf31:
Also proxies give us the ability to this at least:

```purescript
class ProxyFor t p | p -> t, t -> p where
  reflect :: p -> t

instance proxyForString :: IsSymbol s => ProxyFor String @s where
[...]

Issue #3102, opened 2017-10-02
`Char` kind proposal


kcsongor:
With `ConsSymbol`, we now have a way of deconstructing `Symbol`s, but in its current implementation, the resulting head is also a `Symbol`, and it is only implicitly implied that this will be a single character (it doesn't get solved otherwise).

This is not ideal, as it's not clear from the kind signatures what's happening. My proposal is to introduce a kind `Char`, which would have literal syntax just like that at the existing value-level `Char` literals.

With this, the signature of `ConsSymbol` would be

```
class ConsSymbol (h :: Char) (t :: Symbol) (s :: Symbol) | h t -> s, s -> h t
```

paf31:
I'm fine with it. Conversion back and forth is simple with `ConsSymbol c "" s` anyway.

kcsongor:
Thanks, I will have a go at this tomorrow then.

PR #4207, opened 2021-11-21
Type-level integers

### LINKED ISSUE #2899: Add type-level natural numbers

PureFunctor:
This PR implements type-level integers for PureScript and resolves #2899. This builds against latest `master` instead of #3058 as rebasing would be pretty tedious. Type-level integer literals would be kinded as the `Int` type so breakage should be fairly minimal, but it's probably best that this would be merged in a major release.

Here's an example of what this makes possible:
```purs
module Main where

import Prim.Int (class Add, class Mul, class Compare)
import Prim.Ordering (EQ)

data Proxy :: forall k. k -> Type
data Proxy n = Proxy

addInt :: forall x y z. Add x y z => Proxy x -> Proxy y -> Proxy z
addInt _ _ = Proxy

mulInt :: forall x y z. Mul x y z => Proxy x -> Proxy y -> Proxy z
mulInt _ _ = Proxy

compareInt :: forall x y z. Compare x y z => Proxy x -> Proxy y -> Proxy z
compareInt _ _ = Proxy

a :: Proxy 42
a = addInt (Proxy :: _ 21) (Proxy :: _ 21)

b :: Proxy 42
b = mulInt (Proxy :: _ 21) (Proxy :: _ 2)

c :: Proxy EQ
c = compareInt (Proxy :: _ 21) (Proxy :: _ 21)

```

For further discussion:
- `Pow`/`Root`
- `Sub`/`Div`/`Mod`
- `Int` vs `Integer` for the backend. Having reflection means that `Nat`s are bounded to the upper limit for JS, so the current `Integer` backend is a bit overkill for this.

---

**Checklist:**


src/Language/PureScript/Constants/Prelude.hs
```diff
@@ -273,6 +273,9 @@ pattern DataSymbol = ModuleName "Data.Symbol"
 pattern IsSymbol :: Qualified (ProperName 'ClassName)
 pattern IsSymbol = Qualified (Just DataSymbol) (ProperName "IsSymbol")
 
+pattern IsNat :: Qualified (ProperName 'ClassName)
+pattern IsNat = Qualified (Just DataSymbol) (ProperName "IsNat")
```
JordanMartinez, line 277: Should the module be `DataSymbol`? Or `DataNat`?

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -127,6 +129,18 @@ primSymbolDocsModule = Module
   , modReExports = []
   }
 
+primNatDocsModule :: Module
+primNatDocsModule = Module
+  { modName = P.moduleNameFromString "Prim.Nat"
+  , modComments = Just "The Prim.Nat module is embedded in the PureScript compiler. Unlike `Prim`, it is not imported implicitly. It contains automatically solved type classes for working with `Nats`."
```
JordanMartinez, line 135: Rather than "working with `Nats`", should this say, "working with type-level natural numbers"?

MonoidMusician:
Regarding the names: there was some discussion that it would be nice to actually rename `Symbol` to `String` (#3103) and have the same names at the type level, especially if we had datakinds at some point. That would be invasive, so we don't necessarily have to do that. But is it at least possible with polykinds/type-in-type now? Can we create one `Prim.Natural` and have it work at the term and type levels? If so, maybe we could consider adopting term-level `Natural` into Prelude, with a very small API, and deprecating other packages that define it.

I suppose the problem is that we don't have dedicated natural literals at the term level, so it quickly gets to be a lot of work, sorry.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -475,6 +482,39 @@ entails SolverOptions{..} constraint context hints =
...
+    solveNatAdd [arg0, arg1, arg2] = do
+      (arg0', arg1', arg2') <- addNats arg0 arg1 arg2
+      let args' = [arg0', arg1', arg2']
+      pure [TypeClassDictionaryInScope Nothing 0 EmptyClassInstance [] C.NatAdd [] [] args' Nothing Nothing]
+    solveNatAdd _ = Nothing
+
+    addNats :: SourceType -> SourceType -> SourceType -> Maybe (SourceType, SourceType, SourceType)
+    addNats arg0@(TypeLevelNat _ l) arg1@(TypeLevelNat _ r) _ = pure (arg0, arg1, srcTypeLevelNat (l + r))
+    addNats arg0@(TypeLevelNat _ l) _ arg2@(TypeLevelNat _ r) = pure (arg0, srcTypeLevelNat (l + r), arg2)
+    addNats _ arg1@(TypeLevelNat _ l) arg2@(TypeLevelNat _ r) = pure (srcTypeLevelNat (l + r), arg1, arg2)
```
JordanMartinez, line 499: I may be misunderstanding this part of the code checker. However, it seems that the first case of `addNats` is correctly states, "Given `l r o`, `o == l + r`".

However, the second case seems to say, "Given `l r o`, `r == l + o` rather than `r == o - l`, which should only hold if the resulting `r` is a natural number. Likewise for the third case.

If you dropped the second and third case, then you'd need to drop the second and third functional dependency of the `AddNat` type class.

JordanMartinez (commented):
Just some initial comments as I looked through this.

src/Language/PureScript/Docs/Prim.hs
PureFunctor: :+1:, although I originally based this off the documentation for `Prim.Symbol`, so maybe that should be reworded in the future too, likely after #3103

src/Language/PureScript/TypeChecker/Entailment.hs
PureFunctor: Ah, this was left-over from when I was testing stuff. I'll comment this out in the meantime until there's a determinate implementation for `Sub`.

src/Language/PureScript/Constants/Prelude.hs
PureFunctor: Good catch! Thanks

PureFunctor:
@MonoidMusician So, if I understand correctly, we can introduce term-level Naturals to mirror the relationship between String and Symbol, right? e.g. Prim introduces the Natural type, then `prelude`/`natural` exposes Data.Natural and `typelevel-prelude` exposes Type.Data.Natural.

It's definitely quite a lot of work but I do think that this would be pretty nice.

rhendric:
I don't see the point of dedicated `Sub`, `Div`, `Mod` classes when these things could just be implemented with functional dependencies on `Add` and `Mul`. Am I wrong about that, or is there a reason the former would be better?

More broadly, and probably more controversially, should we be adding new type-level literals ad-hoc like this? As @MonoidMusician reminded us, we're already possibly regretting the separation between `Symbol` and `String`. I wonder if in the future we'll want to support all PS literals at the type level, and have a primitive class like
```purs
class IsReflectable t v | v -> t where
  reflectType :: forall proxy. proxy v -> t
```
which would subsume both `IsSymbol` and `IsNat`. We'd also be wrestling with the `Int` versus `Nat` distinction there.

I bring this up because, if the compiler is going to implement `Add` and `Compare`, do we lose anything by using type-level `Int`s instead of a new `Nat` kind? I know this was previously discussed [here](https://github.com/purescript/purescript/pull/3058#issuecomment-333628405). I'm not strongly swayed by the counterarguments there; they may well be correct, but I might need to see a worked example of something that is made worse or impossible by using type-level ints with appropriate constraints instead of type-level nats. If we can get away with type-level ints, then maybe we should go straight to `IsReflectable` instead of `IsInt`, even if we don't implement other type-level literals right now.

PureFunctor:
> I don't see the point of dedicated Sub, Div, Mod classes when these things could just be implemented with functional dependencies on Add and Mul. Am I wrong about that, or is there a reason the former would be better?

Personally, I think that having separate type classes for these operations is much more intuitive than having different fundeps provide different solutions. Symbol.Cons, for example, is an example of a pretty overloaded type class whose name doesn't really help make its purpose understandable. If we could rename Add/Mul to something that reflects that it can also solve Sub/Div, I'd definitely be more inclined towards this.

> I wonder if in the future we'll want to support all PS literals at the type level [...]

We already 2/5 of the non-container literals at the type-level (Boolean, String/Symbol), and so far we've only really gotten it "right" with `Boolean`. If it is an inevitability, then yeah, I do agree that `IsReflectable` is what we should go with, to at least help ease future implementations (see also #3102).

> [...] do we lose anything by using type-level Ints instead of a new Nat kind? [...]

For use-cases that require non-negative numbers, I agree that `Compare` is a good compromise over losing the convenience of intrinsically only having natural numbers.

JordanMartinez:
> Am I wrong about that, or is there a reason the former would be better?

I'm not sure, but how would you implement `Sub` using just `Add` with only one functional dependency of `l r -> o`, which corresponds to `l + r == o`? Would you recursively search what the values of `r` are (e.g. `l + 1`, `l + 2`, etc.) until you found a value that worked?

rhendric:
No no, in a `Sub`-less world, `Add` would go back to having `l r -> o, l o -> r, r o -> l` fundeps, and each one would be solved by the compiler with the appropriate expression (`o = l + r`, `r = o - l`, `l = o - r` respectively). For `Mul`, and for `Add` as well if we stick to nats and not ints, some constraints will be rejected instead if the result is not a member of the expected domain (`Mul 5 r 17`, for example, should not have an instance).

lib/purescript-cst/src/Language/PureScript/Environment.hs
```diff
@@ -593,13 +593,16 @@ primNatClasses =
         [ FunctionalDependency [0, 1] [2]
         ] True)
 
-    -- class Mul (multiplicand :: Nat) (multiplier :: Nat) (product :: Nat) | multiplicand multiplier -> product
+    -- class Mul (multiplicand :: Nat) (multiplier :: Nat) (product :: Nat)
+    --   | multiplicand multiplier -> product, multiplicand product -> multiplier, multiplier product -> multiplicand
     , (primSubName C.moduleNat "Mul", makeTypeClassData
         [ ("multiplicand", Just kindNat)
         , ("multiplier", Just kindNat)
         , ("product", Just kindNat)
```
PureFunctor: I feel like these should be renamed for consistency, thoughts?

PureFunctor:
These changes implement the solving of differences/quotients through the `Sum` and `Mul` type classes, including a new type class for negation, practically transforming this PR for type-level ints :sweat_smile:

JordanMartinez:
I think I agree with @rhendric that supporting type-level Ints first with additional constraints for confirming that an Int is a Nat is overall a better long-term goal than supporting Nats initially with the possibility of supporting Ints later.

rhendric:
Sorry to flip-flop on you like this, but now I think I was missing something after all. 😰 The problem is 0. `Add` is fine, but I should have realized that the `l o -> r` and `r o -> l` fundeps on `Mul` aren't actually valid, because there are many instances of `Mul 0 r 0` and `Mul l 0 0`.

Can we discuss some concrete use cases for integer division and modulus at the type level? I don't dispute they exist; I just want to know what they're like. Leaving aside using PureScript as [a dynamically-typed, interpreted language](https://aphyr.com/posts/342-typing-the-technical-interview), applications might want additional information about the types being constrained. For example, I could imagine having

```purs
-- n = q * d + r, 0 < d, 0 <= r < d
class (Compare 0 d LT, Compare (-1) r LT, Compare r d LT) <= DivMod n d q r | n d -> q r, d q r -> n
```

as the class that provides division and modulus, with the `Compare` superclasses offering some potentially useful guarantees about the output of `DivMod`. On the other hand, this constrains the divisor to be positive as opposed to simply non-zero, which could be less useful than a less constrained `DivMod`. On the gripping hand, a *more* constrained version like

```purs
-- n = q * d + r, 0 < d, 0 <= r < d
-- Additionally, n >= 0 implies q >= 0 and vice-versa
class (Compare (-1) n LT, Compare 0 d LT, Compare (-1) r LT, Compare r d LT, Compare (-1) q LT) <= DivModNat n d q r | n d -> q r, d q r -> n
```

might be useful in some other applications. I'm at a bit of a loss for figuring out how best to navigate this.

Relatedly, much like the `Coercible` class has symmetric and transitive laws baked into its solver, we probably want the same for `Compare`. (If a `Compare a b GT` is available, a `Compare b a LT` constraint should be satisfied; if `Compare a b GT` and `Compare b c GT` are both available, then a `Compare a c GT` constraint should be satisfied; and so on with more complex chains of inference.) I know `Compare` for `Symbol` doesn't have this, but I suspect it doesn't get used for anything too complicated. Without at least the symmetric law, the above `DivModNat` class would need to explode into:

```purs
class ( Compare (-1) n LT
      , Compare n (-1) GT
      , Compare 0 d LT
      , Compare d 0 GT
      , Compare (-1) r LT
      , Compare r (-1) GT
      , Compare r d LT
      , Compare d r GT
      , Compare (-1) q LT
      , Compare q (-1) GT
      ) <= DivModNat n d q r | n d -> q r, d q r -> n
```

Which, yikes.

Unfortunately, the solver for `Coercible` is a beast; maybe @kl0tl would have some tips for how much of that machinery is relevant to another symmetric and transitive compiler-solved class. (It would be nice if we could extract some reusable parts to share between `Coercible` and the `Compare` implementations.)

PureFunctor:
I suppose I could take this time to explore and ask questions about that part of the compiler. I haven't fully explored the entailment framework but would a property like symmetry `(Compare a b LT) <=> (Compare b a GT)` imply that several `TypeClassDict` entries are returned for a specific solver? Taking for example Coercible's `symm` test:
```purescript
symm :: forall a b. Coercible a b => b -> a
symm = coerce
```
As I understand it, `coerce` implies the constraint `Coercible b a`, but `Coercible a b` is a solution for this.

rhendric:
My incomplete understanding is that the Coercible solver takes into account both what's needed and what's given, and tries to find a path from the one to the other (unlike other type classes which get solved working only from what's needed, using what's given just as a way to terminate the process). I think the current algorithm is an implementation of the OutsideIn(X) algorithm described in https://www.microsoft.com/en-us/research/uploads/prod/2018/05/coercible-JFP.pdf, although that's about all I know about that.

PureFunctor:
Thanks, I've formulated the following test case for symmetry:
```purescript
module Main where

import Prim.Nat (class Compare)
import Prim.Ordering (GT, LT)

data Proxy :: forall k. k -> Type
data Proxy n = Proxy

compare :: forall m n o. Compare m n o => Proxy m -> Proxy n -> Proxy o
compare _ _ = Proxy

symmLt :: forall m n p. Compare m n LT => Compare n m p => (Proxy m -> Proxy n -> Proxy p) -> Proxy p
symmLt _ = Proxy

symmGt :: forall m n p. Compare m n GT => Compare n m p => (Proxy m -> Proxy n -> Proxy p) -> Proxy p
symmGt _ = Proxy

a = symmLt compare

b = symmGt compare
```
Which emits the following warnings:
```purescript

Error found:
in module Main
at SolvingNatCompare.purs:18:1 - 18:19 (line 18, column 1 - line 18, column 19)

  The inferred type
                                                                             
    forall t31 t32 t33. Compare t32 t33 LT => Compare t33 t32 t31 => Proxy LT
                                                                             
  has type variables which are not determined by those mentioned in the body of the type:

    t31 could not be determined
    t32 could not be determined
    t33 could not be determined

  Consider adding a type annotation.

in value declaration a

See https://github.com/purescript/documentation/blob/master/errors/AmbiguousTypeVariables.md for more information,
```

```purescript

Error found:
in module Main
at SolvingNatCompare.purs:20:1 - 20:19 (line 20, column 1 - line 20, column 19)

  The inferred type
                                                                             
    forall t31 t32 t33. Compare t32 t33 GT => Compare t33 t32 t31 => Proxy GT
                                                                             
  has type variables which are not determined by those mentioned in the body of the type:

    t31 could not be determined
    t32 could not be determined
    t33 could not be determined

  Consider adding a type annotation.

in value declaration b

See https://github.com/purescript/documentation/blob/master/errors/AmbiguousTypeVariables.md for more information,
```
I believe what we want here is for `t31` to be solved with respect to `LT`/`GT`. I've also traced the solver and found out I get contexts for solving these the same way that `Coercible` does, so that's one step forward.

rhendric:
I don't think those cases test what you want. At any rate, I'd expect to see those errors even if `Compare` had symmetric solving; they're complaining about the fact that the compiler doesn't know at what types to instantiate the uses of `compare`, since all it knows is that `compare` is being provided as an argument to one of the `symm` functions, which are themselves polymorphic.

Here are some test cases that I think would be better:
```purs
assertLesser :: forall l r. Compare l r LT => Proxy { left :: l, right :: r }
assertLesser = Proxy

assertGreater :: forall l r. Compare l r GT => Proxy { left :: l, right :: r }
assertGreater = Proxy

assertEqual :: forall l r. Compare l r EQ => Proxy { left :: l, right :: r }
assertEqual = Proxy

symmLt :: forall m n. Compare m n GT => Proxy { left :: n, right :: m }
symmLt = assertLesser

symmGt :: forall m n. Compare m n LT => Proxy { left :: n, right :: m }
symmGt = assertGreater

symmEq :: forall m n. Compare m n EQ => Proxy { left :: n, right :: m }
symmEq = assertEqual

reflEq :: forall (n :: Nat). Proxy { left :: n, right :: n }
reflEq = assertEqual

transLt :: forall m n p. Compare m n LT => Compare n p LT => Proxy n -> Proxy { left :: m, right :: p }
transLt _ = assertLesser

transLtEq :: forall m n p. Compare m n LT => Compare n p EQ => Proxy n -> Proxy { left :: m, right :: p }
transLtEq _ = assertLesser

transEqLt :: forall m n p. Compare m n EQ => Compare n p LT => Proxy n -> Proxy { left :: m, right :: p }
transEqLt _ = assertLesser

transGt :: forall m n p. Compare m n GT => Compare n p GT => Proxy n -> Proxy { left :: m, right :: p }
transGt _ = assertGreater

transGtEq :: forall m n p. Compare m n GT => Compare n p EQ => Proxy n -> Proxy { left :: m, right :: p }
transGtEq _ = assertGreater

transEqGt :: forall m n p. Compare m n EQ => Compare n p GT => Proxy n -> Proxy { left :: m, right :: p }
transEqGt _ = assertGreater

transEq :: forall m n p. Compare m n EQ => Compare n p EQ => Proxy n -> Proxy { left :: m, right :: p }
transEq _ = assertEqual

transSymmLtGt :: forall m n p. Compare n m LT => Compare n p GT => Proxy n -> Proxy { left :: m, right :: p }
transSymmLtGt _ = assertGreater

transSymmLtEq :: forall m n p. Compare n m LT => Compare n p EQ => Proxy n -> Proxy { left :: m, right :: p }
transSymmLtEq _ = assertGreater

transSymmEqGt :: forall m n p. Compare n m EQ => Compare n p GT => Proxy n -> Proxy { left :: m, right :: p }
transSymmEqGt _ = assertGreater

transSymmEqLt :: forall m n p. Compare n m EQ => Compare n p LT => Proxy n -> Proxy { left :: m, right :: p }
transSymmEqLt _ = assertLesser

transSymmGtLt :: forall m n p. Compare n m GT => Compare n p LT => Proxy n -> Proxy { left :: m, right :: p }
transSymmGtLt _ = assertLesser

transSymmGtEq :: forall m n p. Compare n m GT => Compare n p EQ => Proxy n -> Proxy { left :: m, right :: p }
transSymmGtEq _ = assertLesser

transSymmEq :: forall m n p. Compare n m EQ => Compare n p EQ => Proxy n -> Proxy { left :: m, right :: p }
transSymmEq _ = assertEqual
```

PureFunctor:
I don't think the resolver algorithm would be as heavy as the one Coercible uses, considering that we're really only checking for Nat base cases, reflexivity, symmetry, and transitivity; if not, at least that's what I believe :sweat_smile:. In the meantime, I've conjured up the boilerplate below which comes from `solveCoercible`.
```hs
solveNatCompare ctx [a, b, c] = do
  let compareDictsInScope = findDicts ctx C.NatCompare Nothing
      givens = M.fromList $ flip mapMaybe compareDictsInScope $ \case
        dict | [a', b', c'] <- tcdInstanceTypes dict -> Just ((a', b'), c')
             | otherwise -> Nothing
```
This essentially makes symmetry checks a simple `M.lookup (b, a) givens`, but I've only tested this on trivial cases. For transitivity, however, I don't think the type for `givens` is immediately compatible, as for `C a b LT => C b c LT => C a c LT`, what I think I'd want is to follow the breadcrumbs of `a -> b`,  `b -> c`, whilst ensuring that both resolve to `LT` as determined by `a -> c`.

PureFunctor:
The algorithm that I came up with is a bit terse at the moment but it compiles with the passing test. As sort of a general overview, it solves constraints through the following steps.
1) All types that appear in an `EQ` are recorded.
2) Types are renamed with respect to the `EQ` instances
3) Graphs are built for chains of `LT`s and `GT`s.

and when solving:
- `EQ`, the parameters must be literally equal or their renamed forms are valid `EQ` instances.
- `LT`/`GT`, determine whether `a` can reach `b`, with respect to symmetry and transitivity.

I think type-level literals would require special casing, considering that the algorithm was originally built with nominal data types like strings in mind; I'll add more tests to see if this is the case. I'll also be documenting the algorithm as I go with this PR in accordance with #4192.

PureFunctor:
I'll start to convert all the current work done for type-level naturals to type-level integers instead, as I think the latter is much more useful in terms of general-purpose arithmetic at the type level. The `Compare` solver is powerful enough to be able to infer constraints like the following with respect to properties.
```hs
nonZero :: forall n m o. Compare n 0 GT => Proxy n -> Proxy n
nonZero _ = Proxy
```

After that, I think all that's left is the generalized reflection type class that @rhendric mentioned.

rhendric:
> all that's left

Also resolving my concerns in https://github.com/purescript/purescript/pull/4207#issuecomment-977077890.

In the absence of further progress on that front, I'd like to see the maximally cautious thing of removing anything that could trigger a division-by-zero crash (the two division-y fundeps on `Mul` and the entire `Mod` class). Nothing stops us from adding more fundeps and classes later.

While I've got my hatchet out, isn't `Negate` also a bit superfluous if we can subtract from 0 via `Add`?

PureFunctor:
Got it :+1: 

> While I've got my hatchet out, isn't Negate also a bit superfluous if we can subtract from 0 via Add?

Yes, and it's probably better have this as a synonym type class in the type-level prelude.

PureFunctor:
This PR overall should be ready for review. To summarize, I added:
* the `TypeLevelInt` type, which uses `Int` as its kind.
* the `Add`, `Mod`, and `Compare` type classes, which are automatically solved by the compiler
  * `Add` can double as `Sub` because of its functional dependencies
  * `Compare` has a dedicated algorithm for solving constraints nominally e.g. anything that isn't a type-level integer
* the `IsReflectable` type class, which allows a common interface for reflecting type-level literals down to the term-level

JordanMartinez:
> I'm not sure how to proceed with adding support for parsing negative integer literals at the type-level, so for now, using a constraint like Add 21 n 0 is the only way to negate an integer.

So that I can help, could you clarify what you are having trouble with? For example, is it unfamiliarity with happy in general? Or something else?

PureFunctor:
@JordanMartinez I'm not thoroughly familiar with Happy yeah. I'm mostly concerned about possible conflicts with type operators.

JordanMartinez:
@PureFunctor Mm... not quite. I'll explain what I know and explain what I think needs to be done.

When we get to the parser, we have a stream of `SourceToken`s. To make it easier to write a parser, [this section functions as token aliases](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L69-L142). It tells Happy, the parser generator, "When I write something like [`(` in a parser rule](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L327), I really mean [`SourceToken _ TokLeftParen`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L69)". 

When you see a [parser rule like `export`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L629-L637), see below for how to read it. `$1`, `$2`, `$3` refer to the result of first, second, or third parse result or token, respectively. 
```
parserName :: { ReturnType }
  : parserOrTokenAlias1 parserOrTokenAlias2 parserOrTokenAlias3 { value_returned_via_pure $1 $2 $3 }
  | tokenAlias1 {% monadic_computation_that_returns_a_value }
  | tokenAlias1 {%% I can't remember }
  | tokenAlias1 {%^ I can't remember }
```

You'll see that there's an alias for [`LIT_INT`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L141) that refers to `TokInt`. If we look at the `Lexer.hs` file where `TokInt`s are created, we see two places:
- [in `number`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L540), which calls [`integer1`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L563-L574) and then verifies that there isn't an exponent.
- [in `hexadecimal`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L641-L650), which is called by [`number`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L529)

We also see that [`number` is called in `token`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L249), which tokenizes the stream of characters into source tokens. Crucially, `token` only handles the integer parsing, not whether the integer is positive or negative. In [this usage of `orOperator2`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L243), we check what characters follow a `-` and only parse it as [`TokOperator`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L338) if it's [a standalone `-`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Lexer.hs#L275). Looking back at the Lexer, we see that the [`-` alias corresponds to that token](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L91).

At the term level, we parse the `-` part of negative numbers via [`expr3`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L383) into [`ExprNegate`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Types.hs#L337). The `number` parser isn't done until [`exprAtom`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L435). We'll need to do something similar at the type-level. 

Looking at all usages of `symbol`, we see the following two parsers: [`typeAtom`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L322-L323) and [`typeKindedAtom`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L337). They both use two other parsers:
    - [`qualSymbol`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L226-L229)
    - [`string`](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L323)

So, here's what we need to do:
- we'll need to add a `TypeNegate` data constructor to the [`Type` type](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Types.hs#L139) that follows the [`ExprNegate` data constructor](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Types.hs#L337).
- we'll need to add `TypeInt` data constructor to the [`Type` type](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Types.hs#L139) that follows the [`ExprNumber` data constructor](https://github.com/purescript/purescript/blob/master/lib/purescript-cst/src/Language/PureScript/CST/Types.hs#L329), but which doesn't use `Either` since this can only be an `Integer`.
- we'll need to add another parser step in the `type1` to `type5` "parser ladder" that handles the `-` parse by following what is done in `expr3` above.
- we need a new parser for `qualInt` that follows what's done in `qualSymbol`
- we need to add `int` and `qualInt` to `typeAtom` and `typeKindedAtom`, following symbol's usage.

Lastly, for the term level, there's a [desugar step to turn `-1` into `negate 1`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Sugar/Operators.hs#L43-L52). I'm not sure if something similar is needed here.

rhendric:
@JordanMartinez, I just gave that advice a quick read but I'm a little concerned about the details—we don't want to be able to negate arbitrary types, only integers. Probably instead of looking at the `expr` ladder for precedent, we should look at the `binder` ladder.

JordanMartinez:
@rhendric So, would something like this be possible if we went with the `binder` approach? And should it? Not sure if we'd care to do this, but just thinking outloud.
```purescript
negate
  :: forall (a :: Int).
   . Proxy a
  -> Proxy (-a)
negate _ = Proxy
```

rhendric:
I think that shouldn't work. If it were to work, it'd have to be sugar for
```purs
negate :: forall (a :: Int) b. Add 0 b a => Proxy a -> Proxy b
```

Not that I don't see the potential utility of something like that, but that kind of magical constraint-adding type translation is sufficiently unlike anything else in PureScript that I think it'd be too confusing to be worth it.

JordanMartinez:
> Not that I don't see the potential utility of something like that, but that kind of magical constraint-adding type translation is sufficiently unlike anything else in PureScript that I think it'd be too confusing to be worth it.

Good point. I think I agree with you that it shouldn't be done. At the very least, we can take the more conservative approach to not do it that way and re-examine this later if there's sufficient evidence to the contrary. 

In that case, would the binder approach be something like this (updating the source spans to account for the `-` token) and using `typeInt` rather than `int` in `typeKindedAtom` and `typeAtom`?
```
typeInt ::
  : '-' int { TypeInt () (second negate $2) }
  | int { TypeInt () $2 }
```

JordanMartinez:
@PureFunctor I just realized that most of what I said above has already been done by you in previous commits (e.g. adding `TypeInt` to the `Type` type). Sorry about that.

rhendric:
Broadly, I think what @PureFunctor originally proposed looked like a good start. We don't want `-1` to parse as a `typeAtom`, because then `f -1` is ambiguous. For consistency, I think that means also excluding it from `typeKindedAtom`, even though I don't think that would cause ambiguity; `typeKindedAtom` looks like it's meant to be a subset of `typeAtom`. (As an aside, I think we *do* want a bare `int` production in `typeKindedAtom`, though; numbers aren't valid labels in record/row syntax.) So the `'-' int` production has to go somewhere higher up the ladder, but probably at or below `type4` so that `n * -1` could parse. That leaves where @PureFunctor put it or one level above. Maybe `# -1` should be a parser error, maybe not; it's going to be a moot point soon since the `#` syntax is slated for removal.

I would say that it'd be nice if we could merge the source position annotations from both the `'-'` and the `int` in that production into the final `TypeInt`, rather than ignoring `$1` altogether.

PureFunctor:
> I would say that it'd be nice if we could merge the source position annotations from both the '-' and the int in that production into the final TypeInt, rather than ignoring $1 altogether.

@rhendric Which would roughly be this, right? I could also probably extract this into `Utils`...
```hs
type5 :: { Type () }
  : type6 %shift { $1 }
  | '-' int { let
                (intToken, intValue) = $2
                rangeWithMinus = widen (srcRange $1) (srcRange intToken)
                intToken' =
                  intToken { tokAnn = (tokAnn intToken) { tokRange = rangeWithMinus }
                           , tokValue = case tokValue intToken of
                               TokInt t i -> TokInt ("-" <> t) (negate i)
                               n -> n
                           }
                intValue' = negate intValue
              in
                TypeInt () intToken' intValue'
            }
```

> As an aside, I think we do want a bare int production in typeKindedAtom, though; numbers aren't valid labels in record/row syntax.

:+1: on this

> Maybe # -1 should be a parser error, maybe not; it's going to be a moot point soon since the # syntax is slated for removal.

I don't think it should, at least I don't think it's really worth it at this point.

PureFunctor:
Small side note, `IsReflectable` can probably also solve type-level `Boolean` and `Ordering`, but I don't think we keep their constants. Not entirely sure how to proceed, but delaying until the implementation for those until this is merged is fine by me.

JordanMartinez:
> I would say that it'd be nice if we could merge the source position annotations from both the '-' and the int in that production into the final TypeInt, rather than ignoring $1 altogether.

Rather than merging the two source position annotations, shouldn't the `-` be included as another argument? For example
```
| TypeInt a (Maybe SourceToken) SourceToken Integer
```
where the first source token is a possible `-` and the second is the actual integer. For example:
```
: int     { TypeInt () Nothing   $2 (intVal $2)          }
| '-' int { TypeInt () (Just $1) $2 (negate $ intVal $2) }
```
Otherwise, wouldn't `foo :: Proxy (-1)` fail to parse if there's a comment in-between the `-` and `1` when it's otherwise valid?
```
foo 
  :: Proxy (-
   -- here's a comment
   1)
```

JordanMartinez:
As a sanity check, can we verify that a type-level integer can't be parsed as a label (e.g. `type Foo = { 42 :: SomeType }` should fail to parse)?

rhendric:
> ```
> | TypeInt a (Maybe SourceToken) SourceToken Integer
> ```

I don't think stray comments would have caused failure to parse, but oh yes, this would be a much better approach than the road I pointed you down, if for no other reason than it keeps the CST more faithful to the original source.

JordanMartinez:
> if for no other reason than it keeps the CST more faithful to the original source.

Yeah. After posting my comment, I realized it's not the 'parsing' that's the issue, but the 'printing'. The parse would get the comments, but since the resulting CST doesn't include them, printing the CST would not produce the same source code.

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
```diff
@@ -0,0 +1,173 @@
...
+import qualified Data.Map as M
+import qualified Data.Set as S
+
+import qualified Language.PureScript.Types as P
+import qualified Language.PureScript.Constants.Prim as P
+
+data Comparison a
+  = Equal a a
+  | LessThan a a
+  | GreaterThan a a
```
rhendric, line 20: Do we need this third constructor? Looking at your approach, it seems to me like we could always choose to represent a `GreaterThan x y` as `LessThan y x` instead, and everything would continue to work.

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
```diff
@@ -0,0 +1,173 @@
...
+  -- it essentially does is:
+  --
+  -- 1. Collects all inequalities, renaming them on the way.
+  --
+  -- 2. Creates graphs for less than and greater than inequalities.
+  inequalities ::
+    ( ( G.Graph, S.Set a -> Maybe G.Vertex )
+    , ( G.Graph, S.Set a -> Maybe G.Vertex )
+    )
+  inequalities = lts &&& gts $ foldMap convert context
```
rhendric, line 132: I think you've overcomplicated this. There only needs to be one inequality graph, with edges pointing from lesser to greater, and you can look for a path from `a` to `b` when solving `LessThan` and from `b` to `a` when solving `GreaterThan` (if we even keep the `GreaterThan` constructor). That should let you simplify things here quite a bit, I think.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -475,6 +496,57 @@ entails SolverOptions{..} constraint context hints =
...
+            LT -> C.orderingLT
+            GT -> C.orderingGT
+          args' = [arg0, arg1, srcTypeConstructor ordering]
+      in pure [TypeClassDictionaryInScope Nothing 0 EmptyClassInstance [] C.IntCompare [] [] args' Nothing Nothing]
+    solveIntCompare ctx [a, b, c] = do
+      let compareDictsInScope = findDicts ctx C.IntCompare Nothing
+          givens = flip mapMaybe compareDictsInScope $ \case
+            dict | [a', b', c'] <- tcdInstanceTypes dict -> mkComparison a' b' c'
+                 | otherwise -> Nothing
+      solved <- solveComparison givens <$> mkComparison a b c
```
rhendric, line 537: Hm, we should be able to solve for unknown `c` in some contexts. Something like the following probably ought to work (modulo the fact that `IsReflectable Ordering` doesn't get solved):

```purs
compareIntTypes :: forall l r ord. Compare l r ord => IsReflectable Ordering ord => Proxy l -> Proxy r -> Ordering
compareIntTypes _ _ = reflectType (Proxy :: _ ord)

test :: forall m n p. Compare m n GT => Compare n p GT => Proxy m -> Proxy n -> Proxy p -> Ordering
test m _ p = compareIntTypes m p
```

rhendric (changes requested):
There's some more work to be done on the `Compare` solver, but this does look promising and much more straightforward than I was fearing!

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
```diff
@@ -0,0 +1,173 @@
...
+    -> a
+    -> a
+    -> Bool
+  solveInequality f a b =
+    let (graph, search) = f inequalities
+    in fromMaybe False $ do
+      a' <- search $ rename a
+      b' <- search $ rename b
+      let norm = G.path graph a' b'
+          symm = G.path (G.transposeG graph) b' a'
```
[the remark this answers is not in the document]
PureFunctor: > Um, isn't symm exactly norm with more steps? There's a path from a' to b' iff there's a path from b' to a' in the graph with all edges reversed?

Yeah, this seems to be left over from my initial algorithm. It passes all tests so this can be removed.

> But also, I think you want to make sure that a' != b'. Consider:

Agreed.

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
PureFunctor: I think so too, I can also add failing tests to be sure.

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
```diff
@@ -0,0 +1,173 @@
...
+  -- the context and comparison is renamed as:
+  -- > [ abc .=. abc, xyz .=. xyz ] & (abc .=. abc)
+  Equal a b
+    | a == b -> True
+    | otherwise ->
+        let
+          comparison' = alpha comparison
+          context' = alpha <$> context
+        in
+          comparison' `elem` context'
```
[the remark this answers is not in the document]
PureFunctor: This was a redundancy check to make sure that the `context'` is formed appropriately. This can be removed, yep.

src/Language/PureScript/Constants/Prelude.hs
```diff
@@ -273,6 +273,12 @@ pattern DataSymbol = ModuleName "Data.Symbol"
 pattern IsSymbol :: Qualified (ProperName 'ClassName)
 pattern IsSymbol = Qualified (Just DataSymbol) (ProperName "IsSymbol")
 
+pattern DataReflectable :: ModuleName
+pattern DataReflectable = ModuleName "Data.Reflectable"
+
+pattern IsReflectable :: Qualified (ProperName 'ClassName)
+pattern IsReflectable = Qualified (Just DataReflectable) (ProperName "IsReflectable")
```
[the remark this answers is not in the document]
JordanMartinez: I was thinking, to reduce breakage, we would initially make `IsSymbol` extend `IsReflectable`. 
```purescript
class IsReflectable sym String => IsSymbol sym

reflectSymbol :: forall sym. IsSymbol sym => Proxy sym -> String
reflectSymbol = reflectType
```

As for whether to deprecate and then later remove `IsSymbol` entirely, I'm not sure.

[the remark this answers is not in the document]
rhendric: Technically still a breaking change, as users can currently implement `IsSymbol` instances themselves:

```purs
foreign import data MySymbol :: Symbol

instance IsSymbol MySymbol where
  reflectSymbol _ = "a deeply bad idea?"
```

I don't know if anyone's done this, but we've left the footgun lying around, so to speak.

lib/purescript-cst/src/Language/PureScript/CST/Types.hs
```diff
@@ -137,6 +137,7 @@ data Type a
   | TypeWildcard a SourceToken
   | TypeHole a (Name Ident)
   | TypeString a SourceToken PSString
+  | TypeInt a SourceToken Integer
```
rhendric, line 140: This should be reworked per @JordanMartinez's comment https://github.com/purescript/purescript/pull/4207#issuecomment-980741807.

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
```diff
@@ -0,0 +1,107 @@
...
+  -- context, they're still equal.
+  Equal a b ->
+    rename a == rename b
+  -- Proving that an inequality exists between two nominal values takes
+  -- a very different code path compared to Equal; to summarize, it
+  -- creates a directed graph of all inequalities in the renamed
+  -- context, and searches whether a path exists from point a to point b.
+  Unequal a b
+    | a /= b ->
+        solveInequality a b
```
rhendric, line 38: Negative test case:
```purs
bad :: forall a b c. Compare a b EQ => Compare b c GT => Proxy c -> Proxy (left :: a, right :: b)
bad _ = assertGreater
```

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
```diff
@@ -0,0 +1,107 @@
...
+
+    makeStrong :: [(a, [a])] -> [G.SCC a]
+    makeStrong = G.stronglyConnComp . fmap make
+
+  inequalities :: (G.Graph, S.Set a -> Maybe G.Vertex)
+  inequalities = asGraph $ foldMap convert context
+    where
+    convert :: Relation a -> [Relation (S.Set a)]
+    convert = fmap alpha . \case
+      c | Unequal a b <- c -> [c, Unequal b a]
```
rhendric, line 78: Negative test case:
```purs
bad :: forall a b c. Compare a b GT => Proxy (left :: b, right :: a)
bad = assertGreater
```

src/Language/PureScript/Constants/Prelude.hs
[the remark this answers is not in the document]
JordanMartinez: Ugh...

src/Language/PureScript/TypeChecker/Entailment.hs
PureFunctor: I could probably transform `solveRelation` into an inference-like function for this use-case then.

PureFunctor: I've done initial work on this locally but this case in particular seems like requires some the complexity removed from 6ddb8520a89c47af3885ddd140dc87924c3b8892 to do anything significant :confused:

lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
```diff
@@ -284,7 +284,7 @@ flattenType = \case
   TypeWildcard _ a -> pure a
   TypeHole _ a -> pure $ nameTok a
   TypeString _ a _ -> pure a
-  TypeInt _ a _ -> pure a
+  TypeInt _ _ a _ -> pure a
```
JordanMartinez, line 287: I think this should be
```diff
- TypeInt _ _ a _ -> pure a
+ TypeInt _ a b _ -> foldMap a <> pure b
```
or something like that.

lib/purescript-cst/src/Language/PureScript/CST/Positions.hs
```diff
@@ -246,7 +246,7 @@ typeRange = \case
   TypeWildcard _ a -> (a, a)
   TypeHole _ a -> nameRange a
   TypeString _ a _ -> (a, a)
-  TypeInt _ a _ -> (a, a)
+  TypeInt _ _ a _ -> (a, a)
```
JordanMartinez, line 249: This might also need to be updated.

lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
JordanMartinez: Looking at the [layout tests](https://github.com/purescript/purescript/tree/master/tests/purs/layout) can help confirm this.

JordanMartinez (changes requested):
I think there are still two issues unrelated to the type checking stuff in latest commits. Otherwise, I think this is nearly there.

PureFunctor:
I've partially addressed the concern about inference for certain cases like what was was pointed out in https://github.com/purescript/purescript/pull/4207#discussion_r757861727, but this has a few edge cases that I ought to investigate still. That said, I'll probably be continuing work on this on the next few weeks or so, after the holidays.

src/Language/PureScript/TypeChecker/Entailment.hs
rhendric: I'll take a look sometime soon; I'd be mildly surprised if that's the case but I could be glossing over something important.

PureFunctor: It turns out that it wasn't really the case in a43a2c71f96875059ac5295b7a32c729543f3b6e. A sort of compromise I had to make in the simpler framework was that `Unequal` actually means "less than" internally in the inequality graph, and "greater than" is solved by flipping the arguments around. So cases like:
```
a < b < c
x > y > z
```
Get normalized into:
```
a < b < c
z < y < x
```
If solving for `z -> x` solves with `LT` just fine but with `x -> z`, the search is flipped to do `z -> x` but it returns `GT` instead.

PureFunctor:
I've played around with this PR locally and I can't think of/find any other edge cases for this. That said, should we maximize the coverage for failing tests in this PR? I think all that's missing is tests for transitivity as the existing ones test for symmetry and reflexivity.

JordanMartinez:
> That said, should we maximize the coverage for failing tests in this PR? I think all that's missing is tests for transitivity as the existing ones test for symmetry and reflexivity.

I'd prefer that.

PureFunctor:
> I'd prefer that.

Got it. By the way, is there an easy method or utility for generating `.out` files for failing tests? Right now I find myself running the following then manually cleaning up the output.
```
$ script -q -c "stack exec purs -- compile tests/purs/failing/FailingTest.purs" tests/purs/failing/FailingTest.out > /dev/null
```

rhendric:
If you don't have any .out file for a test, the test suite should write one which you can then examine before checking in.

Alternatively, if you don't want to delete an existing .out file, you can run your tests with `HSPEC_ACCEPT=true` in your environment, which will force all .out files to be regenerated (for the tests you're matching on, if applicable).

PureFunctor:
Oh, I wasn't aware of that. Thank you.

src/Language/PureScript/Constants/Prelude.hs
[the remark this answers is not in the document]
PureFunctor: > but we've left the footgun lying around, so to speak.

I do feel like the responsible choice to do here is to go ahead and remove the footgun entirely since it's not really the intended behaviour. I'm keen on deprecating `IsSymbol` if this PR lands and we end up removing it down the line entirely, perhaps along with replacing `Symbol` with `String`.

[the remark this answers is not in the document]
JordanMartinez: I think the larger issue is that kinds are currently open. We'd have to make these specific kinds closed to completely remove this footgun. Because once this PR lands, one could also do something like `foreign import data Foo :: Integer`, which creates a new variant of the same footgun.

JordanMartinez:
> Alternatively, if you don't want to delete an existing .out file, you can run your tests with HSPEC_ACCEPT=true in your environment, which will force all .out files to be regenerated (for the tests you're matching on, if applicable).

This should be documented. There's been a lot of times where I've wasted due to not being familiar with Haskell and its tooling.

rhendric:
You're not wrong, but this one's a PureScript exclusive:

https://github.com/purescript/purescript/blob/53b4117045f5517023f5d22f9fa506e651e4ff3e/tests/TestUtils.hs#L272

JordanMartinez:
Is this otherwise ready to merge?

PureFunctor:
@JordanMartinez There's still a few things left to consider and implement. Aside from the `IsReflectable` conversation above, there's also the `DivMod` class and making sure that reflection somehow respects the 32-bit limit. The latter is much more trickier because normally this is done in JS code generation, making it backend-independent.

The more trivial changes are implementing `IsReflectable` for `Ord` and `Bool`.

src/Language/PureScript/Constants/Prelude.hs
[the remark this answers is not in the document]
JordanMartinez: It sounds like we should do the following:
- now 
    - support both `IsSymbol` and `IsReflectable` separately (so no breaking changes are made)
    - deprecate `IsSymbol`?
- later in a breaking release
    - remove `IsSymbol`
    - rename the kind `Symbol` to `String`
- later (at some point in the future)
    - remove the footgun

JordanMartinez:
> I'd just found out that IntOutOfRange fires on reflected values.

Could you clarify this?

PureFunctor:
@JordanMartinez For reflection `reflectSymbol (Proxy :: _ "hello")`, the literal gets injected as the return value of `reflectSymbol` when generating JS code, like so:
```js
Data_Symbol.reflectSymbol({
    reflectSymbol: function () {
        return "hello";
    }
})(Type_Proxy["Proxy"].value);
```
Since the routine for checking integer boundaries in JS code generation operates on the AST, something like `reflectType (Proxy :: _ 987654321)` would generate:
```js
Data_Reflectable.reflectType({
    reflectType: function () {
        return 987654321;
    }
})(Type_Proxy["Proxy"].value);
```
thus getting caught by said routine. I wasn't aware of this until I did my testing but it's something worth adding a failing test for.

https://github.com/purescript/purescript/blob/81dd8ea73ad8eab9cc56be492fe5883bec59b8d0/src/Language/PureScript/CodeGen/JS.hs#L439-L456

rhendric:
I haven't looked at any of the work done since my last review, but I hope to give it a thorough evaluation sometime this week. Sorry for the delay.

PureFunctor:
I should be able to do the rest of the changes in 2 weeks time. Since it seems like most of the focus is directed towards pushing ES modules over the line, I think it's best to have that merged in first :smile:

JordanMartinez:
> Since it seems like most of the focus is directed towards pushing ES modules over the line, I think it's best to have that merged in first smile

Oh, I was thinking the opposite. Better to get this merged in first because the ES modules will take a lot of work.

PureFunctor:
Ah, I suppose that's true.

JordanMartinez:
Just a thought on the `IsReflectable` class name. Is there any benefit to prefixing this class with `Is`? How about just `Reflectable`?

tests/purs/passing/SolvingCompareInt.purs
```diff
@@ -0,0 +1,88 @@
...
+litGt = assertGreater
+
+litEq :: Proxy ( left :: 0, right :: 0 )
+litEq = assertEqual
+
+infer :: forall l r o. Compare l r o => Proxy l -> Proxy r -> Proxy o
+infer _ _ = Proxy
+
+inferSolved :: forall m n p. Compare m n GT => Compare n p GT => Proxy m -> Proxy n -> Proxy p -> Proxy GT
+inferSolved m _ p = infer m p
```
rhendric, line 86: If this is meant to test what I suggested in https://github.com/purescript/purescript/pull/4207#discussion_r757861727, I'm not sure it does. I could be wrong about this, but I suspect the compiler is being told that `o` is `GT` by the signature of `inferSolved`. To truly test that `o` is inferred correctly based only on the `Compare` constraints, I *think* you need to do something like resolve a class instance based on `o`:

```suggestion
class AssertIsGT o where
  assertIsGT :: Proxy o -> Boolean

instance AssertIsGT GT where
  assertIsGT _ = true

infer :: forall l r o. Compare l r o => AssertIsGT o => Proxy l -> Proxy r -> Boolean
infer _ _ = assertIsGT (Proxy :: _ o)

inferSolved :: forall m n p. Compare m n GT => Compare n p GT => Proxy m -> Proxy n -> Proxy p -> Boolean
inferSolved m _ p = infer m p
```

(Fortunately, this also passes!)

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -475,6 +496,54 @@ entails SolverOptions{..} constraint context hints =
...
+            LT -> C.orderingLT
+            GT -> C.orderingGT
+          args' = [arg0, arg1, srcTypeConstructor ordering]
+      in pure [TypeClassDictionaryInScope Nothing 0 EmptyClassInstance [] C.IntCompare [] [] args' Nothing Nothing]
+    solveIntCompare ctx [a, b, _] = do
+      let compareDictsInScope = findDicts ctx C.IntCompare Nothing
+          givens = flip mapMaybe compareDictsInScope $ \case
+            dict | [a', b', c'] <- tcdInstanceTypes dict -> mkRelation a' b' c'
+                 | otherwise -> Nothing
+      c' <- solveRelation givens a b
```
rhendric, line 537: There's one more enhancement that I think would be important to cover:

```purs
litTrans :: forall a. Compare a 1 GT => Proxy ( left :: a, right :: 0 )
litTrans = assertGreater
```

Because the compiler can solve arithmetic facts like `Compare 1 0 GT`, and the compiler can make transitive connections between `Compare` dictionaries in the context, I think for consistency we should make sure the compiler can also make transitive connections between context dictionaries and arithmetic facts.

One way I can imagine doing this is adding to the list of givens here the full set of `Unequal`s that can be inferred from the distinct `TypeLevelInt`s that are used in the `compareDictsInScope`, `a`, and `b`. It's brutish, but should suffice—I have a hard time imagining having so many known integer types floating around in `Compare` dicts at once that the extra O(n^2) relations cause any noticeable problems—at least, not before the core `solveRelation` implementation does. But if you can devise a more elegant approach, please be my guest.

rhendric (changes requested):
Sorry I took so long to finally get around to this. It's been on my mind but life has been busy!

I've left a few notes, but I think this is very close.

rhendric:
> Just a thought on the `IsReflectable` class name. Is there any benefit to prefixing this class with `Is`? How about just `Reflectable`?

Just the parallel construction with `IsSymbol`, of which `IsReflectable` is a generalization. As I mentioned in [this thread](https://github.com/purescript/purescript/pull/4207#discussion_r757838262), there's also precedent for calling this class `Reifies`. Or we could break precedent entirely and go with a new name like `Reflectable`.

rhendric:
Three more stray thoughts:

1. This is either a bad idea or something we probably can punt on, but let's discuss: should this work? Why or why not?
```purs
testUnification :: forall a b. Compare a b EQ => Proxy a -> Proxy b
testUnification p = p
```

2. The type-level int `Compare` is now much more powerful than the `Compare` for symbols. Should we ‘backport’ (at some point) any or all of this functionality? Should we unify them and, in lieu of adding a second special case, introduce a polykinded `class Compare :: forall k. k -> k -> Ordering -> Constraint` class similar to what we're doing with `IsReflectable`? Are the pros/cons there basically the same or is there something worth differentiating between the two?

3. This is a more general thought—I am approaching the point with this PR where, from an implementation and user-facing API perspective, I can happily sign off on it; but from a type system design perspective, I don't feel competent to declare that the direction this PR is going is not going to cause problems. I'd love to get a quick opinion from someone on @purescript/compiler who feels more confident about type system design questions before we merge.

thomashoneyman:
> Or we could break precedent entirely and go with a new name like `Reflectable`

A 👍 from me on a name like `Reflectable`, `Reifies`, or even `Reflect` or `Reify` rather than continue with the `Is` prefix, which has always struck me as odd. Of course, this is a minor matter 😄

PureFunctor:
1. I'm open to just leaving this as an implementation detail for the `Compare` solver, though I feel like it can also be used as a stand-in for `TypeEquals` in appropriate contexts.
2. I'm not sure I can think of any reasons how string comparisons can benefit from the `Compare` solver, though I'm fine with it either way.

src/Language/PureScript/TypeChecker/Entailment/IntCompare.hs
```diff
@@ -0,0 +1,103 @@
...
+    asGraph :: [Relation (S.Set a)] -> (G.Graph, S.Set a -> Maybe G.Vertex)
+    asGraph = makeGraph . clean . foldMap asNode
+      where
+      asNode :: Relation (S.Set a) -> [(S.Set a, [S.Set a])]
+      asNode (Unequal a b) = [(a, [b]), (b, [])]
+      asNode _             = mempty
+
+      makeGraph :: [(S.Set a, [S.Set a])] -> (G.Graph, S.Set a -> Maybe G.Vertex)
+      makeGraph m = case G.graphFromEdges $ make <$> m of
+        (g, _, f) -> (g, f)
```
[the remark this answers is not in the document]
PureFunctor: This works better, thanks!

PureFunctor:
Question about `DivMod`, how often would performing modulo with a negative number be as to have less constraints? I feel like in terms of "practical" applications e.g. sized vectors, parsing state, or [wags](https://github.com/mikesol/purescript-wags), they won't pop up as much, but then, this also asks whether `DivMod` should even exist... I suppose one way to do this is to have a `TypeNotEquals` constraint but it feels clunky.

JordanMartinez:
1. Using `Compare` as a stand-in for `TypeEquals` would benefit from not needing to pass the dictionary into the function. So :+1: from me. But again, I'm not sure what the implications of this are.
2. Should we take a more conservative approach with this first by not 'backporting' that functionality, and then adjusting it later as a breaking change if it's desired and once we have a better idea of what we'd want to accomplish with it?

JordanMartinez:
@PureFunctor Just FYI. You got a [linting error](https://github.com/purescript/purescript/runs/5154596598?check_suite_focus=true#step:5:10) that's failing CI.

rhendric:
@xgrommx

> when it will be merged?

Not sure if this is a cultural difference, but from my perspective, this comment is a bit rude. The answer to this question, at least for a volunteer, developed-in-the-open project like this one, is always, ‘When the work is done and the project is ready for it;’ asking the question anyway comes off as pushy. (I also take a dim view of the similar, ‘Any progress on this?’ Read the thread; it's what I'd have to do to answer you, and why is my time less valuable than yours?)

Enthusiasm for progress is very understandable and appreciated! But please try to keep the environment friendly for the various people who are volunteering time to give you progress for free.

rhendric:
@PureFunctor

> Question about `DivMod`, how often would performing modulo with a negative number be as to have less constraints? I feel like in terms of "practical" applications e.g. sized vectors, parsing state, or [wags](https://github.com/mikesol/purescript-wags), they won't pop up as much, but then, this also asks whether `DivMod` should even exist... I suppose one way to do this is to have a `TypeNotEquals` constraint but it feels clunky.

I have no idea... I think there probably exist use cases for `DivMod`, but I also think that relative to the other arithmetic classes, this one is tricky enough that we should wait until we have a few concrete use cases in mind before we design and implement it. Let people get used to basic type-level ints first.

(Happy to be overruled by people walking up to this thread waving their pre-existing `DivMod` use cases!)

src/Language/PureScript/TypeChecker/Entailment.hs
rhendric: > Could you elaborate on this approach?

Suppose, as in your second example, you have the context `a < 10, 0 < a` and you are in the process of solving `Compare a 20 o`. My concept is:
* Collect all the integer literals under discussion and sort: `[0, 10, 20]`
* Iterating over the `tails` of that list, add nominal-style constraints to the context: `0 < 10, 0 < 20, 10 < 20`
* Solve as usual

I'm not quite sure I fully understand *your* approach, but I'm guessing it'd require more code and it's not yet clear to me that it's more efficient.

PureFunctor:
I've summarized a few points regarding this PR in [this Discourse post](https://discourse.purescript.org/t/rfc-type-level-integers/2900) to hopefully add a bit more reach. With regards to `DivMod`, I've thought about it this morning and I'm quite comfortable with deferring implementing said feature until users have warmed up to type-level integers and require richer features.

On the topic of `IsReflectable`, I'm fine with `Reflect` or `Reflectable`. I've noticed that the compiler-solved type classes in Prim are either named and act:
1. like functions e.g. `Add`, `Compare`, `Mul`, `Lacks`, `Nub`, `Warn`, `Fail`
2. like relationships e.g. `Coercible`
3. like functions and relationships e.g. `Union`, `Cons`

I suppose this is a matter of active/passive voice too? English is not my first language I'd be happy to be corrected on this :smile:. That said, I feel like `Reflectable` makes more sense to use as a constraint name, as naming it `Reflect` implies its use as a type-level function, which isn't really all that useful.

src/Language/PureScript/TypeChecker/Entailment.hs
PureFunctor: What I'm thinking of is basically something like:
```hs
module Main where

import Prelude

import Data.Foldable (all)
import Effect (Effect)
import Effect.Console (logShow)

data Relationship a
  = LessThan a
  | GreaterThan a

type Context a = Array (Relationship a)

solveFacts :: forall a. Ord a => Context a -> Relationship a -> Boolean
solveFacts context relationship = all (relationship `adheresTo` _) context
  where
  adheresTo :: Relationship a -> Relationship a -> Boolean
  adheresTo (LessThan a) (LessThan b) = b > a
  adheresTo (GreaterThan a) (GreaterThan b) = a > b
  adheresTo (LessThan a) (GreaterThan b) = a > b
  adheresTo (GreaterThan a) (LessThan b) = b > a

main :: Effect Unit
main = do
  logShow $ solveFacts [GreaterThan 0, LessThan 10] (GreaterThan 5)
```
The compiler would collect arithmetic facts to build the context for a specific variable, then verify the fact that's being checked. Though this approach would likely need correction.

JordanMartinez:
I know I already said it, but I'm for `Reflectable`

By `NonZero`, you mean something like this?
```purescript
class NonZero :: Int -> Constraint
class NonZero i

instance Fail (Text "Got a zero!") => NonZero 0
else instance NonZero (i :: Int)
```
Could the same thing be achieved via an `IsZero` constraint?
```purescript
class IsZero :: Int -> Boolean -> Constraint
class IsZero i b | i -> b

class (IsZero i False) <= DivMod ...
```

rhendric:
Re `NonZero` etc.: the question is what does any of this get us? `DivMod` is not going to be implemented by users, so the only point of giving it superclasses is making those superclass instances available in any context where you have a `DivMod`. If these superclasses are memberless, then the only point of having instances of them is if people design APIs that consume these instances. So show me the designs for APIs that consume `NonZero` etc.

My reluctance for including `DivMod`, as I tried to explain in a [previous comment](https://github.com/purescript/purescript/pull/4207#issuecomment-977077890), is that there are lots of relationships that are true and expressible *just with `Compare`* between numerator, denominator, quotient, and remainder, but only if the sign of the denominator is known. To wit:
* if `d > 0`, then `d > r >= 0`
* if `d < 0`, then `d < r <= 0`
* if `d > 0`, then `q < 0` iff `n < 0`
* if `d < 0`, then `q < 0` iff `n > 0`

Our options are:
1. ignore these relationships, which I think will hinder the utility of `DivMod`, but that depends on the use cases
2. constrain `DivMod` to require the denominator to be positive, which may also hinder the utility of `DivMod` but in a different way, again depending on the use cases (but also frees us from having to think about `NonZero` as a special case)
3. come up with even more type-level stuff to capture the conditionality of these relationships, which I think is probably too speculative to be worth it

The reason I keep saying to hold off on `DivMod` until we have use cases is that I want to be able to distinguish between 1 and 2. Once we figure out which path is more useful, then we'll know whether we want a `NonZero`-like thing or not, and then we can bikeshed about how it works.

JordanMartinez:
> then the only point of having instances of them is if people design APIs that consume these instances. So show me the designs for APIs that consume NonZero etc.

Good point.

Correct me if I'm wrong, but can't `DivMod` be added in the future as a non-breaking change? If that's the case, then I'd agree that excluding it for the time being is worth doing. If this PR (without `DivMod`) gets included in a release, then we'll likely run into more design issues that `DivMod` would fix besides those mentioned in the Discourse post.

PureFunctor:
> By `NonZero`, you mean something like this?

Yes, although I haven't added solving for the `NonZero` constraint yet, it only really exists to constrain `DivMod`; refactoring to `IsZero` should also be fairly trivial.

> but can't `DivMod` be added in the future as a non-breaking change?

I'd be more than happy to do this as well. I think having a dedicated issue for it would flesh out the implementation better.

Correct me if I'm wrong, but this should be the final sets of primary changes right? I hope I'm not forgetting anything that was discussed a few weeks back :smile:.

JordanMartinez:
So, have we reached consensus that the following changes need to be made before this PR will be merged?
- [ ] Rename `IsReflectable` to `Reflectable`
- [ ] Drop `DivMod` and `NonZero`

JordanMartinez (approved):
LGTM!
I marked this PR as a breaking change, but I can't recall why. I think it's only breaking to users of the compiler as a library, but not breaking like ES modules breaking.

PureFunctor:
@JordanMartinez I believe it was solely because of the introduction of the `Nat` type, which is used by some packages.

JordanMartinez:
Looks like there's another merge conflict due to #4239.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -477,6 +505,67 @@ entails SolverOptions{..} constraint context hints =
...
+          givens = flip mapMaybe compareDictsInScope $ \case
+            dict | [a', b', c'] <- tcdInstanceTypes dict -> mkRelation a' b' c'
+                 | otherwise -> Nothing
+          facts = case (a, b) of
+            (_, TypeLevelInt _ _) ->
+              mkFacts (args : (tcdInstanceTypes <$> compareDictsInScope))
+            (TypeLevelInt _ _, _) ->
+              mkFacts (args : (tcdInstanceTypes <$> compareDictsInScope))
+            _ ->
+              []
```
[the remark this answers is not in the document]
PureFunctor: Just to clarify, `Compare a b LT` should resolve, right? or do you mean something else with this example?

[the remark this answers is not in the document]
rhendric: Right.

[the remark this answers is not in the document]
PureFunctor: Sorry this last change took me a bit :sweat_smile:, I thought `assertLesser` was a type declaration here.

PureFunctor:
Test failures are related to the bundler because of 516abc2, it seems like. Tests for type-level ints are all passing.

rhendric:
Yes, sorry, these failures are indeed not your fault. Because ES module support is currently in development, the tests currently depend on some temporary forks of PureScript libraries that have been changed to be compatible with the new module format, and those forks are getting some cowboy commits at the moment that are breaking everyone else. This will affect local and CI builds (if your local build isn't affected yet, it's because your cached test dependencies haven't expired yet—you have at most a day before this affects you). Please stand by while we restore order.

JordanMartinez:
CI builds now.

rhendric (approved):
Nice work.

CHANGELOG.d/feature_is_reflectable.md
```diff
@@ -0,0 +1,6 @@
+* Implement the IsReflectable type class
+
+  The `IsReflectable` type class is a common interface for reflecting
+  type-level values down to the term-level. Its instances are
+  automatically solved by the compiler, and it allows `Symbol` and `Int`
+  kinded types to be reflected to their term-level representations.
```
rhendric, line 6: Ah, shoot, not quite. This needs rewording to account for the rename and the new instances.

```suggestion
* Implement the Reflectable type class

  The `Reflectable` type class is a common interface for reflecting
  type-level values down to the term-level. Its instances are
  automatically solved by the compiler, and it allows `Symbol`, `Int`,
  `Bool`, and `Ordering` kinded types to be reflected to their term-level
  representations.
```

JordanMartinez: If this is all that needs to be changed, can't we just commit the suggestion and merge? Or is there potentially more in the changelog that needs to be edited?

natefaubion: > If this is all that needs to be changed, can't we just commit the suggestion and merge?

I think we should give the original authors plenty of time to respond. I don't think we need to be in a hurry about this.

JordanMartinez: You're right.

PR #4209, opened 2021-11-29
Adds option to ignore foreign declaration check

mikesol:
**Description of the change**

As ES module support may not make it for some time, this PR is intended to provide an escape hatch where checks for FFI validity can be dispensed with altogether. For those that use industrial bundlers like webpack, builds should still fail when FFIs are wanky, as the imports will not align.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_disable-ffi-check.md
+++ b/CHANGELOG.d/feature_disable-ffi-check.md
@@ -1,0 +1,1 @@
+* Adds a `no-foreign-decl-check` flag to `purs compile` to skip foreign declaration checks.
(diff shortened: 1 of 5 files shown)


hdgarrood:
I’m 👎 on this, sorry. I think one problem with this is that it will kind of suck if someone publishes a library which depends on this check being disabled and then you try to use it in a project where you would rather keep the check on, because you can’t set compiler options per-library (and I think this is probably a good thing because it would significantly complicate build systems like spago). This check has saved me from wasting my time on quite a few occasions and I think it’s important that using PureScript without JS tooling like webpack remain ergonomic. Perhaps more importantly though, I think this is something the compiler should just take a stance on and stick to, because settings just offload decision making to users; it’s often more appropriate and a better experience overall to make a choice and stick with it.

JordanMartinez:
I strongly agree with Harry on this. While supporting ES modules might still be a ways off for various reasons, I don't feel like this will cause more problems than solutions.

PR #4214, opened 2021-12-16
Support shebang lines in modules

### LINKED ISSUE #4210: Proposal: Support optional shebang line in module header
## Motivation

There are now a few mechanisms for executing a PureScript file, such as through `spago run` or `spago script`. Supporting a shebang line in the a module header would allow users to configure how to run the module (on unix systems) directly within the file itself.

## Proposal

Support an optional [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) line. It must be on the first line of a file.

## Examples

The shebang line is optional, so module headers can stay the same:

```
module Main where

...

main = do
  ...
```

However, they could be present in the module header before the module keyword:

`Main.purs`

```
#! /usr/bin/spago script -d node-fs

module Main where

...

main = do
  ...
```

This can then be run with `./Main.purs`, instead of having to write out the `spago script` command each time.

colinwahl:
**Description of the change**

Supports optional shebang lines in modules. 

Before parsing, shebang lines are chomped off the source. The lexer state is then updated to start at the correct `SourcePos`. Finally, shebang lines are inserted into the `Module` result after parsing. This approach maintains correct source positions in all parsed `SourceToken`, while still allowing the module to be printed exactly as it was originally.

Now, the following parses, and can be printed back exactly as-is:

```
#! spago script

module Main where

import Prelude

main = do
  ...
```

Intends to close #4210

---

**Checklist:**

~- [ ] Added myself to CONTRIBUTORS.md (if this is my first contribution)~


JordanMartinez (approved):
I'm wondering if we want to newtype the `[SourceToken]` argument to `lex` to further clarify that this can only be the shebang. Right now, it only serves to offset the initial `start` position before lexing begins, so one could pass in something else other than the `shebang` on accident.

rhendric:
Are shebangs part of PureScript modules? Or are they part of the on-disk representation of a module understood by `purs` the compiler? GHC takes the latter approach, for example; there's no mention of shebangs in the Haskell Language Report. But we could choose to say that the hypothetical PureScript language spec, which would already hypothetically allow a sequence of comments to precede a module definition, would now allow an initial sequence of those comments to be shebangs, a third comment type only usable at the start of a module parse.

I ask because this implementation feels to me like halfway between these two approaches. I think we should pick a side and, in one case, fully incorporate shebangs into the grammar; or, in the other case, not include shebangs in the module AST.

JordanMartinez (changes requested):
Just an update. It was concluded that we should drop the need to add another token to the AST. Rather, the parser should parse lines of text starting with `#!` as a comment, but only when we know we're parsing a module. Since all comments are stored raw (e.g. `-- foo` gets parsed to `Comment "-- foo"`, not `Comment "foo"`), the shebang would be stored within the comments for the `module` keyword source token. This would require adding another `lex` variant that handles the shebang comment situation before calling `lex` with the rest of the text.

natefaubion:
>  It was concluded 

Just to be clear, by "it was concluded", this came up in a discussion at work, and I made a suggestion to lex it as a comment, essentially saying that shebangs are part of the module lexical grammar.

JordanMartinez:
This PR should be ready for review.

JordanMartinez:
Should the shebang comments be stored in the `ModuleHeader` separately from the non-shebang comments? Or should those be included? Right now, they are included.

src/Language/PureScript/CST/Lexer.hs
```diff
@@ -219,6 +238,64 @@ breakComments = k0 []
...
+    | Just ('#', inp2) <- Text.uncons inp
+    , Just ('!', inp3) <- Text.uncons inp2 = Just ("#!", inp3)
+
+    | otherwise = Nothing
+
+  shebangComment = isShebang >>= \case
+    Just sb -> do
+      Just <$> lineComment sb
+    Nothing ->
+      pure Nothing
```
rhendric, line 286:
```suggestion
  shebangComment = isShebang >>= traverse lineComment
```

src/Language/PureScript/CST/Lexer.hs
```diff
@@ -219,6 +238,64 @@ breakComments = k0 []
...
+  isShebang = Parser $ \inp _ ksucc ->
+    case unconsShebang inp of
+      Just (sb, inp3) ->
+        ksucc inp3 $ Just sb
+      _ ->
+        ksucc inp Nothing
+
+  lineComment acc = do
+    comm <- nextWhile (\c -> c /= '\r' && c /= '\n')
+    pure $ Comment (acc <> comm)
```
rhendric, line 297: This is identical to the other `lineComment` in this file; please reuse.

src/Language/PureScript/CST/Lexer.hs
```diff
@@ -219,6 +238,64 @@ breakComments = k0 []
...
+        Just (Line LF, inp2)
+      _ ->
+        Nothing
+
+  unconsShebang :: Text -> Maybe (Text, Text)
+  unconsShebang inp
+    | Just ('#', inp2) <- Text.uncons inp
+    , Just ('!', inp3) <- Text.uncons inp2 = Just ("#!", inp3)
+
+    | otherwise = Nothing
```
rhendric, line 280:
```suggestion
  unconsShebang = fmap ("#!", ) . Text.stripPrefix "#!"
```

rhendric (changes requested):
I like this overall approach and the new tests are good. I'm a little dismayed by how verbose the implementation in the lexer is, but that's at least partly the pre-existing patterns that have been copied here instead of refactored. (Why do we go to the trouble of rolling our own monadic parser abstraction and then write out most of the lexer functions in longhand anyway? Perhaps there's a performance gotcha that isn't obvious to me?)

These suggestions are the most obvious opportunities for tidying up that I could see, but perhaps after this is merged I'll come back around and investigate a more combinatorial style for these top-level handlers.

JordanMartinez:
I've addressed the comments I got back from your review.

rhendric (approved):

JordanMartinez (approved):
I've overriding my previous review.

PR #4216, opened 2021-12-20
Scope type vars when type checking typed values

### LINKED ISSUE #2941: Instance signatures don't bring type variables into scope

rhendric:
When the compiler is checking an expression that is annotated with a
type against another expected type, and the annotation introduces a type
variable, the compiler needs to introduce that type variable to the
scope of any types used inside the expression.

One noteworthy case of this pattern is member signatures inside
instances. This fix allows type variables introduced in member
signatures to be used in the member declaration itself.

**Description of the change**

Fixes #2941.

---

**Checklist:**

- <s> Added myself to CONTRIBUTORS.md (if this is my first contribution) </s>
- <s> Updated or added relevant documentation </s>

### DIFF
--- a/CHANGELOG.d/fix_scope-type-vars-when-type-checking-typed-values.md
+++ b/CHANGELOG.d/fix_scope-type-vars-when-type-checking-typed-values.md
@@ -1,0 +1,10 @@
+* Scope type vars when type checking typed values
+
+  When the compiler is checking an expression that is annotated with a
+  type against another expected type, and the annotation introduces a type
+  variable, the compiler needs to introduce that type variable to the
+  scope of any types used inside the expression.
+
+  One noteworthy case of this pattern is member signatures inside
+  instances. This fix allows type variables introduced in member
+  signatures to be used in the member declaration itself.
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -726,7 +726,8 @@ check' (DeferredDictionary className tys) ty = do
              (TypeClassDictionary con dicts hints)
              ty
 check' (TypedValue checkType val ty1) ty2 = do
-  (elabTy1, kind1) <- kindOf ty1
+  moduleName <- unsafeCheckCurrentModule
+  ((args, elabTy1), kind1) <- kindOfWithScopedVars ty1
   (elabTy2, kind2) <- kindOf ty2
   unifyKinds' kind1 kind2
   checkTypeKind ty1 kind1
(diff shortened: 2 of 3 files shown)


garyb:
Does this also fix it so you can use a type variable introduced in an instance head, or does it only apply to member annotations? If it fixes the former too it'd be nice to have a test for that!

rhendric:
If you're talking about `forall` in instance heads, as in #1120, no, this doesn't implement that. If not... example please? Or issue link?

garyb:
Ah, actually looks like the thing I meant already works! I guess I'd just mixed it up in my head with the case you've fixed here. I meant this kind of thing:

``` purescript
data I a = I a  

class Foo a where
  something ∷ Unit → a

instance Monoid a ⇒ Foo (I a) where
  something _ = I (mempty ∷ a)
```

JordanMartinez:
I restarted CI, thinking it would include #4217. I guess it doesn't still.

JordanMartinez (approved):
Explaining this code to myself, all this PR is doing is running the same `tvToExpr <$> check val ty1'` code, except that the `withScopedTypeVars` now wraps it. And all that function does is amend the environment's types to include additional type variables that one passes into that function. Once the computation finishes, it sets the environment's types back to what it previously was. If any of the new type variables shadow current ones, a warning will be emitted. 

These type variables come from `kindOfWithScopedVars`, which is used in multiple places throughout the `TypeChecker/Types.hs` file. So, we're not doing anything weird by using it here AFAICT. Lastly, the `unsafeCheckCurrentModule` just gets the current module's name (should this function be renamed to better indicate that in some future PR?). It's unsafe because if there isn't a current module, there's nothing to return. We use it in other places throughout this file, and our usage here is safe AFAICT.

rhendric:
Also worth noting: this PR effectively reverts [part of a change](https://github.com/purescript/purescript/pull/2643/files#diff-8ca6d956028994002d8fa5b54e5a8e6b7f315c1d7357d38c827281349043a40cL639-L646) made in #2643. But that PR [did not change](https://github.com/purescript/purescript/pull/2643/files#diff-8ca6d956028994002d8fa5b54e5a8e6b7f315c1d7357d38c827281349043a40cL382-L386) the analogous logic in `infer`, which survives to this day (and is how I got the idea for this fix in `check`), so I think it was a mistake. But that, and the evidence of passing tests, is about as far as I can go to justify this.

PR #4073, opened 2021-04-19
Fix errors when generating Haddocks

ncaq:
[PureScript compile error when -haddock · Issue #4072 · purescript/purescript](https://github.com/purescript/purescript/issues/4072)

**Description of the change**

added: haddock check in CI.
fixed: haddock error.

Fixes #4072.

---

**Checklist:**

- [x] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [x] Linked any existing issues or proposals that this pull request should close
- [x] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

PR #4217, opened 2021-12-20
Build the entire latest package set in CI

rhendric:
The package set is only built in Linux, as that tends to be the OS that
otherwise finishes first, and there's no need to run this test three
times. The package set is *not* built if there are any files in
CHANGELOG.d starting with `breaking_`, indicating that there are
unreleased breaking changes.

**Description of the change**

Building from the path @JordanMartinez blazed in #4165; the main differences in approach here are not worrying about splitting this step into its own job (it looked like that was going to make CI both take longer and be more complex), and using the presence of `CHANGELOG.d/breaking_*` files to bail out. Let's see if I have any more luck....

---

**Checklist:**

- <s> Added myself to CONTRIBUTORS.md (if this is my first contribution) </s>
- <s> Added a test for the contribution (if applicable) </s>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,10 @@ jobs:
 
       - run: "ci/build.sh"
 
+      - name: "(Linux only) Build the entire package set"
+        if: "${{ runner.os == 'Linux' }}"
+        run: "ci/build-package-set.sh"
+
       - name: "(Release only) Create bundle"
         if: "${{ env.CI_RELEASE == 'true' }}"
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,15 @@ jobs:
 
       - name: "(Linux only) Build the entire package set"
         if: "${{ runner.os == 'Linux' }}"
-        run: "ci/build-package-set.sh"
+        # We build in this directory in build.sh, so this is where we need to
+        # launch `stack exec`. The actual package-set building happens in a
+        # temporary directory.
+        working-directory: "sdist-test"
+        # The presence or absence of the --haddock flag changes the location
+        # into which stack places all build artifacts. Since we use --haddock
+        # in our CI builds, in order to actually get stack to find the purs
+        # binary it created, we need to use the flag here as well.
+        run: "stack --haddock exec ../ci/build-package-set.sh"
 
       - name: "(Release only) Create bundle"
         if: "${{ env.CI_RELEASE == 'true' }}"
--- a/ci/build-package-set.sh
+++ b/ci/build-package-set.sh
@@ -5,29 +5,14 @@ shopt -s nullglob
 
 psroot=$(dirname "$(dirname "$(realpath "$0")")")
 
-if [[ "${CI:-}" ]]; then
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -57,6 +57,10 @@ jobs:
 
       - run: "ci/build.sh"
 
+      - name: "(Linux only) Build the entire package set"
+        if: "${{ runner.os == 'Linux' }}"
+        run: "stack exec ci/build-package-set.sh"
```
JordanMartinez, line 62: Does this need to be `stack exec bash ci/build-package-set.sh` or something like that?

ci/build-package-set.sh
```diff
@@ -0,0 +1,38 @@
...
+cd "$tmpdir"
+
+echo ::group::Ensure Spago is available
+which spago || npm install spago
+echo ::endgroup::
+
+echo ::group::Create dummy project
+echo 'let upstream = https://github.com/purescript/package-sets/releases/download/XXX/packages.dhall in upstream' > packages.dhall
+echo '{ name = "my-project", dependencies = [] : List Text, packages = ./packages.dhall, sources = [] : List Text }' > spago.dhall
+spago upgrade-set
```
JordanMartinez, line 25: Perhaps passing in the `-P` / `--no-psa` global flag here might fix things? That's one thing I never tried in my PR.

ci/build-package-set.sh
```diff
@@ -0,0 +1,53 @@
...
+
+  # Go to where we actually built PureScript in a previous step.
+  cd "$psroot/sdist-test"
+
+  # !!! `--local-doc-root`? Yes. Our ci/build.sh script, for some reason,
+  # doesn't put the purs binary in a location included in the PATH that
+  # `stack exec` uses, so `stack exec which purs` doesn't work. I don't have a
+  # good explanation for why that is, or why local-doc-root is the path that
+  # *does* contain the binary, but... well, here we are and I'm sorry.
+  pursdir=$(dirname "$(stack path --local-doc-root)")/bin
```
JordanMartinez, line 22: Oh, right... I actually fixed an issue in #4139 using this same solution. #4139 fixed #4138 because we added the `--haddock` flag in #4172 as a way to fix #4073.

Also, I'm pretty sure #4172 didn't fix #4073. There's no docs for [PureScript on Hoogle](https://hackage.haskell.org/package/purescript).

JordanMartinez (approved):
LGTM! While I think we should provide a better solution to #4072 than what we're currently doing, it should be handled in a separate PR.

ci/build-package-set.sh
JordanMartinez: The docs stopped being published in v0.14.0, which I believe is the same release where we split `purescript-cst` and `purescript-ast` into their own libraries.

rhendric: Ah... I thought it looked familiar! Thanks.

Figuring out what's up with the missing Hoogle docs is more than I want to tackle right now, but at least I can update that comment to be a little more helpful.

f-f (approved):
This looks great, thanks @JordanMartinez and @rhendric!

PR #4224, opened 2021-12-29
Eagerly apply substitutions when inferring object binders

### LINKED ISSUE #3938: Inconcsistency with polytype resulting from record destructuring
## Description

```purescript
module Main where

import Prelude

data Poly = Poly (forall a. a -> a)

type PolyR = { id :: forall a. a -> a }

test1 :: Poly
test1 = Poly \a -> a

test2 :: PolyR
test2 = { id: \a -> a }

wat :: (forall b. b -> b) -> Unit
wat k =  unit

example1 :: Unit
example1 =
  let Poly id = test1
  in wat id
  
example2 :: Unit
example2 =
  let { id } = test2
  in wat id
```

`example1` compiles fine, while `example2` results in:

```
Could not match type

    a1

  with type

    b0
```

on `id` in the application `wat id`.

I would have shrugged this off, but the curious thing is that if you replace the `id` in that application with a `?id`, in _both_ cases you see a polytype in the context:

```
  in the following context:

    id :: forall a. a -> a
```

The second example works fine if you replace `id` with `test2.id`.

## Expected behavior

I would expect example2 to compile fine given it lists `id` as a polytype in the context.

## PureScript version

0.13.8

sjpgarcia:
Closes #3938. This makes it so that substitutions are eagerly applied when inferring object binders.

I'd figured out that the following failing example:
```purescript
type PolyR = { id :: forall a. a -> a }

test :: PolyR
test = { id: \a -> a }

wat :: (forall b. b -> b) -> Unit
wat _ =  unit
  
example :: Unit
example = let { id } = test in wat id
```
compiles when the `id` binder is annotated with a polytype.
```
example :: Unit
example = let { id: id :: forall a. a -> a } = test in wat id
```

This patch removes ambiguity in the compiler when inferring the row type of the record being destructured. Instead of having the following substitutions:
```purescript
[ (t7, (id :: t9 | t8))
, (t8, ())
, (t9, forall a. a -> a)
]
```
we force `t7` into
```purescript
[ (t7, (id :: forall a. a -> a | ()))
, (t8, ())
, (t9, forall a. a -> a)
]
```

This is a pretty simple fix, but there must be a better way that involves refactoring this particular `inferBinder` case entirely.

---

**Checklist:**

* Added myself to CONTRIBUTORS.md (if this is my first contribution)
* Updated or added relevant documentation

### DIFF
--- a/CHANGELOG.d/fix_3938.md
+++ b/CHANGELOG.d/fix_3938.md
@@ -1,0 +1,3 @@
+* Eagerly apply substitutions when inferring object binders
+
+  Resolves #3938.
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -534,7 +534,8 @@ inferBinder val (LiteralBinder _ (ObjectLiteral props)) = do
   rest <- freshTypeWithKind (kindRow kindType)
   m1 <- inferRowProperties row rest props
   unifyTypes val (srcTypeApp tyRecord row)
-  return m1
+  sub <- gets checkSubstitution
+  pure $ M.map (substituteType sub) m1
   where
   inferRowProperties :: SourceType -> SourceType -> [(PSString, Binder)] -> m (M.Map Ident SourceType)
   inferRowProperties nrow row [] = unifyTypes nrow row >> return M.empty
--- a/tests/purs/passing/3938.purs
+++ b/tests/purs/passing/3938.purs
@@ -1,0 +1,31 @@
+module Main where
+
+import Prelude
+import Effect (Effect)
+import Effect.Console (log)
+
+type PolyR = { id :: forall a. a -> a }
+
+test :: PolyR
+test = { id: \a -> a }
+
+wat :: (forall b. b -> b) -> Unit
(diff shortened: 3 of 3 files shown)


sjpgarcia:
> but there must be a better way that involves refactoring this particular inferBinder case entirely.
The mapping operation here isn't as expensive as I'd first thought, since the number of items inside `m1` corresponds to the number of row items. This should be ready for review.

rhendric:
Performance may not be a concern, but this seems to me like a patch over a deeper issue. If I understand the investigation so far, the error is produced because somewhere in type checking, it matters whether we have a `TUnknown` that stands for a `ForAll` or the substituted `ForAll`. I don't know why this seems to be an issue only with object binders specifically, but I'd like to get to the bottom of what isn't handling the `TUnknown` correctly, just in case it's triggered by other code paths. (Or, if there's some invariant I'm not aware of where we assume that `TUnknown`s are going to be absent from the type passed to some function, to identify and document that invariant, and then check to see if there are other code paths that need to use substitutions in addition to this one.)

PureFunctor:
Closing this in favor of another fix I'm eyeing.

Issue #4191, opened 2021-10-26
Make installation/deployment easier on non-Ubuntu Linux environments


mikesol:
## Summary

Stack's support for non-Ubuntu systems is not good enough to build on a few common environments like AWS Linux. One potential solution is to use `cabal` for builds instead.

## Motivation

Selfishly, at Meeshkan, we can no longer build on CI because we switched to Vercel, which runs on AWS Linux. But I do believe that this is indicative of a potential friction point elsewhere - we ran into this as well, for example, when trying to deploy PureScript and Spago to AWS Lambda.

## Proposal

Sub out cabal for stack wherever we use stack.

## Examples

For example, in this repo, we'd include the following `cabal.project` file or some variant thereof (courtesy of @hdgarrood).

```cabal
packages:
  purescript.cabal
  lib/purescript-cst/purescript-cst.cabal

-- Print test output to the terminal immediately rather than waiting for the
-- test suite to finish running.
-- as this no longer works when using cabal new-build, we comment it out
-- test-show-details: streaming

-- Disable optimization by default when building as a local package (i.e.
-- during development).
optimization: False

package aeson-pretty
  -- avoid installing the aeson-pretty CLI tool, since we only want the library
  flags: +lib-only

package these
  -- switch these flags off to avoid dependencies we don't need
  flags: -assoc -quickcheck
```

The result of `cabal new-build` is:

```
Linking /home/meeshkan-abel/mike/purescript/dist-newstyle/build/x86_64-linux/ghc-8.10.7/purescript-0.14.5/x/purs/noopt/build/purs/purs ...
```

So that seems to be the `purs` binary that'd then be the artifact. We'd do a similar-ish thing for trypurescript and spago.

# Conclusion

One of PureScript's best features IMO is its ease of deployment and use in industrial production environments. In light of this & all the available info, dropping `stack` and using `cabal` seems like a good call.

toastal:
Wouldn't a statically-linked binary also help the NPM installer on 'exotic' Linuxes too (#4266 isn't the fix needed, just tried)?

I'm having an issue getting setup on NixOS with a client that has only used NPM as the installer/distributor for PureScript-related tooling; this would probably introduce breakage to that system by switching to `easy-purescript-nix` because some users are on ARM-based Macs which isn't supported by `easy-purescript-nix` because PureScript's pre-built binaries are all only `x86_64`. A statically-linked binary using `musl` could get me over this issue (in theory though, because there aren't releases for old versions).

PR #4113, opened 2021-06-14
Pin CI OS versions, instead of using the latest

f-f:
**Description of the change**

As noted [on Discourse](https://discourse.purescript.org/t/dockerfile-for-purescript-0-14-2/2379) and discussed in #4107 we should be pinning the OS versions that we use for our CI, so that the prebuilt binaries don't ship with unexpected versions of their dynamic dependencies.

---

**Checklist:**

- [ ] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0000)")
- [ ] Added myself to CONTRIBUTORS.md (if this is my first contribution)
- [x] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

Issue #4107, opened 2021-06-08
Pin environments used in CI


hdgarrood:
Right now, we use `ubuntu-latest`, `macos-latest` etc in CI, in `.github/workflows/ci.yaml`. This is problematic, because it means that the prebuilt binaries are likely to stop working on older platforms between releases, without anything in the changelog. See for example https://discourse.purescript.org/t/dockerfile-for-purescript-0-14-2/2379, in which the prebuilt binaries no longer worked because we are linking against a newer version of glibc.

I think we should probably pin these to specific versions, and generally be a bit more conservative about updating, since binaries built on a newer OS are sometimes unlikely to work on an older one.

rhendric:
FWIW, the pre-0.14.2 Travis config built on Ubuntu Trusty, macOS 10.13, and Windows Server version 1809. None of these OS versions are exactly [available in GitHub Actions](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources): the earliest Ubuntu available is Bionic (`ubuntu-18.04`), the earliest macOS is 10.15 (`macos-10.15`, which is currently also `macos-latest`), and the Windows offerings are all long-term rather than SAC (we can choose `windows-2016` or `windows-2019` = `windows-latest`).

So if anyone was wondering, as I was, if we could just pin the GitHub versions to what they were before and push out a quick 0.14.3 to fix this... well, not really. If we're lucky, Bionic might still be compatible with whatever glibc is in everyone's Docker images, but someone'd have to check.

(Actually, it looks like `ubuntu-16.04` is still nominally available until [September 20](https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/), so it's technically an option to do that but we'd basically be committing to a major release between now and then, assuming our new policy for bumping CI versions is ‘only on major releases’.)

Let's also note that GitHub's window for decommissioning runner OSes looks narrower than Travis's was. That's one reason—aside from plain old ‘what could go wrong?’ ignorance—I used `-latest` versions in the configs; repinning versions seems like it would be a somewhat frequently-revisited chore performed in a scramble when things break. But if we add it to the major release procedure, probably not so bad.

hdgarrood:
[...]

PR #4228, opened 2022-01-20
Containerized Linux build (Glibc 2.24)

sd-yip:
**Description of the change**

I am utilising a modified CI workflow of this repository to build `purs` with an older Glibc version for personal usage:
https://github.com/sd-yip/purescript/releases/tag/0.14.5-c9

I think it might be a good idea to share the modification here, so is this PR. Please let me know if you like this idea. After that, I can go through the checklist and any suggestions.

The `haskell:8.10.7-stretch` Haskell image, which uses Glibc ~2.23~ 2.24, is used here for the build to save a little bit of the setup time for GHC. In case the provided GHC version does not match the Stackage LTS, the build will fail like this:
```
No compiler found, expected minor version match with ghc-8.10.7 (x86_64) (based on resolver setting in /__w/purescript/purescript/stack.yaml). 
427 To install the correct GHC into /root/.stack/programs/x86_64-linux/, try running "stack setup" or use the "--install-ghc" flag. To use your system GHC installation, run "stack config set system-ghc --global true", or use the "--system-ghc" flag.
```

Related issue: https://github.com/purescript/purescript/issues/4191#issuecomment-952820382

---

**Checklist:**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,16 +14,22 @@ defaults:
 
 env:
   CI_RELEASE: "${{ github.event_name == 'release' }}"
+  STACK_VERSION: "2.7.3"
 
 jobs:
   build:
     strategy:
       fail-fast: false # do not cancel builds for other OSes if one fails
       matrix:
-        # If upgrading Ubuntu, also upgrade it in the lint job below
-        os: [ "ubuntu-18.04", "macOS-10.15", "windows-2019" ]
+        include:
+          - # If upgrading the Haskell image, also upgrade it in the lint job below
+            os: "ubuntu-latest"
+            image: "haskell:8.10.7-stretch@sha256:100f8fb7d7d8d64adb5e106fe8136b8d4cbdc03aeb2cbd145a7597d74b69bafb"
+          - os: "macOS-10.15"
+          - os: "windows-2019"
 
     runs-on: "${{ matrix.os }}"
+    container: "${{ matrix.image }}"
 
     steps:
       - uses: "actions/checkout@v2"
@@ -33,18 +39,28 @@ jobs:
           node-version: "12"
 
       - id: "haskell"
+        name: "(Non-Linux only) Install Haskell"
+        if: "${{ runner.os != 'Linux' }}"
         uses: "haskell/actions/setup@v1"
         with:
           enable-stack: true
           # If upgrading Stack, also upgrade it in the lint job below
-          stack-version: "2.7.1"
+          stack-version: "${{ env.STACK_VERSION }}"
           stack-no-global: true
+
+      - name: "(Linux only) Check Stack version and ensure a (dummy) accessible local repository"
+        if: "${{ runner.os == 'Linux' }}"
+        run: |
+          [ "$(stack --numeric-version)" = "$STACK_VERSION" ]
+          echo 'allow-different-user: true' >>/root/.stack/config.yaml
+          git -C .. init
 
       - uses: "actions/cache@v2"
         with:
           path: |
+            /root/.stack
             ${{ steps.haskell.outputs.stack-root }}
-          key: "${{ runner.os }}-${{ hashFiles('stack.yaml') }}"
+          key: "${{ runner.os }}-MdyPsf-${{ hashFiles('stack.yaml') }}"
 
       - name: "(Windows only) Configure Stack to store its programs in STACK_ROOT"
         # This ensures that the local GHC and MSYS binaries that Stack installs
@@ -55,7 +71,7 @@ jobs:
           mkdir -p "$STACK_ROOT"
           echo "local-programs-path: $STACK_ROOT/programs" > $STACK_ROOT/config.yaml
 
-      - run: "ci/build.sh"
+      - run: "ci/fix-home ci/build.sh"
 
       - name: "(Linux only) Build the entire package set"
         if: "${{ runner.os == 'Linux' }}"
@@ -67,7 +83,7 @@ jobs:
         # into which stack places all build artifacts. Since we use --haddock
         # in our CI builds, in order to actually get stack to find the purs
         # binary it created, we need to use the flag here as well.
-        run: "stack --haddock exec ../ci/build-package-set.sh"
+        run: "../ci/fix-home stack --haddock exec ../ci/build-package-set.sh"
 
       - name: "(Release only) Create bundle"
         if: "${{ env.CI_RELEASE == 'true' }}"
@@ -85,7 +101,7 @@ jobs:
               exit 1;;
           esac
           cd sdist-test
-          bundle/build.sh "$bundle_os"
+          ci/fix-home bundle/build.sh "$bundle_os"
 
       - name: "(Release only) Publish bundle"
         if: "${{ env.CI_RELEASE == 'true' }}"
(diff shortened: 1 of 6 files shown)


rhendric:
I'm pretty sure we don't want to take on publishing an alternate `purs` binary. Glad you're having success compiling PureScript with an alternate glibc, which is absolutely a use case we want to enable; we just don't want to be responsible for maintaining the build products. Maybe there could be some community repository where this build script would be more appropriate? (One could already exist and I might not be aware of it, or you might want to start one yourself.)

sd-yip:
@rhendric Thanks for looking into this pull request!

I think I should better explain my idea a little bit more. I would appreciate it if you could take a look.

The core ideas and considerations of this PR:
* Glibc is backwards-compatible, not forwards.
* I suggest publishing the official `purs` with an older version of Glibc. (No alternate `purs` binary. We continue to have only the standard one for Linux.)
* The official `purs`'s Glibc version is not defined explicitly. It "inherits" from the GitHub actions' machine version (`ubuntu-18.04` with Glibc 2.27).
* Ubuntu tends to have a more updated version of Glibc. Even if an older version is chosen deliberately, for example, with the previously picked oldest `18.04` (https://github.com/purescript/purescript/pull/4113/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR23), we still have common distros that cannot install `purs` easily.
* Instead of somehow delegating the decision of which Glibc version to use to the GitHub Actions team, we can expand the access to older Glibc versions using docker images. Some possible choices:
  * Ubuntu 16.04 Xenial (Glibc 2.23)
    According to https://ubuntu.com/about/release-cycle, it is now in the "Extended Security Maintenance" phase.
  * Debian 9 Stretch (Glibc 2.23)
    According to https://www.debian.org/releases/, it is now the "oldoldstable" release under LTS support.
    I used this one in this pull request because it is available as an out of the box official Haskell image.
* Updating the image version (`haskell:x.x.x-xxx`) is naturally transformable from the current strategy of updating the GitHub Actions' machine version (`ubuntu-xxx`), so the impact to maintainers should be minimal.

Problems this suggestion solves:
* Make `npm install purescript` "just work" for most users.
* Support popular frontend deployment platforms with a large user base to assist the adoption of Purescript! For example, Vercel and Netlify use Amazon Linux 2, which is on Glibc 2.26 currently.

rhendric:
@f-f, in https://github.com/purescript/purescript/issues/4107#issuecomment-860928377 you mentioned that your experience with using Docker in CI for Spago wasn't positive. Any further thoughts on that, or is there anything in this PR that moves the needle for you?

f-f:
@rhendric my annoyance from the Spago build in docker come from the very long CI times, and the complexity of doing caching.

I haven't reviewed this patch, but if CI times are comparable to the current ones then it might be worth doing this, as compiling with old glibcs is desiderable (short of static linking, that is)

ci/build-package-set.sh
```diff
@@ -16,7 +16,7 @@ export PATH="$tmpdir/node_modules/.bin:$PATH"
 cd "$tmpdir"
 
 echo ::group::Ensure Spago is available
-which spago || npm install spago
+which spago || npm install spago@0.20.3
```
f-f, line 19: Why do we pin the Spago version here?

f-f (approved):
I had a proper look at this and I think it's a good idea!

My biggest concern right now is around CI times - the build time for Linux in this PR is 26mins, while on other PRs is stable on 8mins. I guess it's because the caches are being recreated, so let's see if the CI times go down with subsequent pushes.

.github/workflows/ci.yml
```diff
@@ -14,16 +14,22 @@ defaults:
...
   build:
     strategy:
       fail-fast: false # do not cancel builds for other OSes if one fails
       matrix:
-        # If upgrading Ubuntu, also upgrade it in the lint job below
-        os: [ "ubuntu-18.04", "macOS-10.15", "windows-2019" ]
+        include:
+          - # If upgrading the Haskell image, also upgrade it in the lint job below
+            os: "ubuntu-latest"
+            image: "haskell:8.10.7-stretch@sha256:100f8fb7d7d8d64adb5e106fe8136b8d4cbdc03aeb2cbd145a7597d74b69bafb"
```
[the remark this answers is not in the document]
sd-yip: Yes. And instead of having a chance of mismatched GHC version in any release, I have tested that the build will indirectly "remind" for upgrading of the Haskell image with this error:
```
No compiler found, expected minor version match with ghc-8.10.7 (x86_64) (based on resolver setting in /__w/purescript/purescript/stack.yaml). 
427 To install the correct GHC into /root/.stack/programs/x86_64-linux/, try running "stack setup" or use the "--install-ghc" flag. To use your system GHC installation, run "stack config set system-ghc --global true", or use the "--system-ghc" flag.
```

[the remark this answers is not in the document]
sd-yip: I have also added a note inside `stack.yaml` as a reminder for this.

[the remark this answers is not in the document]
f-f: Great, thanks!

ci/build-package-set.sh
sd-yip: I think so. Will you find it helpful if I port the same pull request for Spago, or do you want to handle it directly?

f-f: I'd love if you could port this same work to Spago!

.github/workflows/ci.yml
```diff
@@ -33,18 +39,28 @@ jobs:
...
-          stack-version: "2.7.1"
+          stack-version: "${{ env.STACK_VERSION }}"
           stack-no-global: true
 
+      - name: "(Linux only) Check Stack version and ensure a (dummy) accessible local repository"
+        if: "${{ runner.os == 'Linux' }}"
+        run: |
+          [ "$(stack --numeric-version)" = "$STACK_VERSION" ]
+          echo 'allow-different-user: true' >>/root/.stack/config.yaml
+          git -C .. init
```
rhendric, line 56: I'm not sure this is a good idea. It seems there are parts of the build that expect to be in a Git repository; what are the consequences of that repository being faked out?

The issue seems to be that if the Git version is less than 2.18, GitHub switches to just unpacking a tarball instead of doing a clone. That to me suggests the alternative of installing a more recent Git instead, though it would be more verbose. Here's how I'd do it, in a Debian-based Docker image:


    if dpkg --compare-versions $(git --version | cut -d' ' -f3) lt 2.18; then
      . /etc/os-release
      echo "deb http://deb.debian.org/debian $VERSION_CODENAME-backports main" > /etc/apt/sources.list.d/backports.list
      apt-get update -qq
      apt-get install -qq git/$VERSION_CODENAME-backports
    fi

.github/workflows/ci.yml
```diff
@@ -33,18 +39,28 @@ jobs:
...
+
       - uses: "actions/cache@v2"
         with:
           path: |
+            /root/.stack
             ${{ steps.haskell.outputs.stack-root }}
-          key: "${{ runner.os }}-${{ hashFiles('stack.yaml') }}"
+          key: "${{ runner.os }}-MdyPsf-${{ hashFiles('stack.yaml') }}"
```
rhendric, line 63: Were these extra characters just for testing?

.github/workflows/ci.yml
```diff
@@ -33,18 +39,28 @@ jobs:
           node-version: "12"
 
       - id: "haskell"
+        name: "(Non-Linux only) Install Haskell"
+        if: "${{ runner.os != 'Linux' }}"
         uses: "haskell/actions/setup@v1"
         with:
           enable-stack: true
           # If upgrading Stack, also upgrade it in the lint job below
```
rhendric, line 47: We can drop this comment now.

.github/workflows/ci.yml
```diff
@@ -33,18 +39,28 @@ jobs:
...
-          stack-version: "2.7.1"
+          stack-version: "${{ env.STACK_VERSION }}"
           stack-no-global: true
 
+      - name: "(Linux only) Check Stack version and ensure a (dummy) accessible local repository"
+        if: "${{ runner.os == 'Linux' }}"
+        run: |
+          [ "$(stack --numeric-version)" = "$STACK_VERSION" ]
+          echo 'allow-different-user: true' >>/root/.stack/config.yaml
```
rhendric, line 55: Might be simpler to `chown root:root .`, unless we expect that to cause some sort of problem? Everything else created in the working directory will be owned by root.

.github/workflows/ci.yml
f-f: I suspect just to invalidate the cache

sd-yip: For an explicit cache invalidation. (Someone else may prefer writing something like `v1`/`v2` instead)

.github/workflows/ci.yml
sd-yip: You are right. I was trying to avoid making it too verbose. And sure, I can try something just like you have suggested.

sd-yip: Done.

JordanMartinez:
This PR causes #4273 to fail CI when it was passing previously. The CI error is:
```
Install snapshot dependencies
  You are not the owner of '/__w/purescript/purescript/'. Aborting to protect file permissions.
  Retry with '--allow-different-user' to disable this precaution.
```
What happens if we 'disable' that precaution? Do we need to be concerned about it?

.github/workflows/ci.yml
```diff
@@ -85,7 +106,7 @@ jobs:
               exit 1;;
           esac
           cd sdist-test
-          bundle/build.sh "$bundle_os"
+          ci/fix-home bundle/build.sh "$bundle_os"
```
JordanMartinez, line 109: Looks like this should be `../ci/fix-home` and is the reason why the alpha-03 release didn't build correctly.

sd-yip: Oh, sorry. Thanks for fixing it.

JordanMartinez: No worries! This was an easy thing to miss.

PR #4235, opened 2022-02-15
Visible type applications

### LINKED ISSUE #3137: Visible type applications
Instead of proxy types, I would like to consider adding pi types, as an alternative to `forall` which requires explicit type application.

*Pros*

- Unlike GHC's type applications, the order of type arguments in a `forall` is still irrelevant.
- No runtime `Proxy`-passing needed
- I don't think this has quite the same risk of kind issues as proxies (although I'm not fully sure yet)

*Possible Syntax*

`Π` is a new type former which binds a type variable. It is introduced with `Λ.` and eliminated with `@` (which mirrors the current proxy syntax).

```purescript
class IsSymbol (s :: Symbol) where
  reflectSymbol :: Πs. String

shoutSymbol :: Πs. IsSymbol s => String
shoutSymbol = Λ. toUpper (reflectSymbol @s)
```

purefunctor:
Closes #3137. This PR adds support for visible type applications and abstractions. The latter pertains to syntax for declaring which universally quantified variables can be instantiated with the former. Here's a simple example:
```hs
id :: forall @a. a -> a
id a = a

id' :: Int -> Int
id' = id @Int
```
Unlike GHC's `TypeApplications`, polytypes with variables that aren't prefixed with `@` cannot be used for visible type application:
```hs
Error found:
in module Main
at Main.purs:6:7 - 6:14 (line 6, column 7 - line 6, column 14)

  An expression of type:
                    
    forall a. a -> a
                    
  cannot be applied to the type:
       
    Int
       

while inferring the type of id
in value declaration id'

See https://github.com/purescript/documentation/blob/master/errors/CannotApplyExpressionOfTypeOnType.md for more information,
```

Visible type abstractions can also appear in class heads, like so:
```hs
class Functor @f where
  map :: forall a b. (a -> b) -> (f a -> f b)

mapArray = map @Array
```

---

**Checklist:**

- [-] Added myself to CONTRIBUTORS.md (if this is my first contribution)


PureFunctor:
CI errors seem to be related to JSON Encoding/Decoding as I'd added a field to the `ForAll` constructor. I'm not sure how to resolve this but maybe we could ignore them for now?

PureFunctor:
I've started working on having visible type abstractions available for type class members like so:
```haskell
class Functor f where
  map :: forall @f a b. (a -> b) -> (f a -> f b)
```
Although attempting to write an instance ends up with an error:
```haskell
data Id a = Id a

instance Functor Id where
  map f (Id a) = Id (f a)
```
emits:
```haskell
Error found:
in module Main
at Main.purs:9:10 - 9:14 (line 9, column 10 - line 9, column 14)

  Could not match type
      
    Id
      
  with type
      
    f0
      

while trying to match type Id t2
  with type f0 a3
while checking that expression case f $0 of          
                                 f (Id a) -> Id (f a)
  has type f0 b1
in value declaration $functorId1

      a3 is a rigid type variable
      b1 is a rigid type variable
      t2 is an unknown type

```
If anyone could point me to the right direction regarding this I'd be more than delighted with the help!

JordanMartinez:
I'm not familiar with how type applications are implemented, so I'm just thinking out loud. Doesn't that error make sense? `f` is a type variable that's introduced at the class head, not in the class' member. So, shouldn't it be more like this?
```purescript
forall @f. class Functor f where
  map :: forall a b. (a -> b) -> f a -> f b
```
and then it's like
```  
map2 :: forall @f. Functor f => (a -> b) -> f (f a) -> f (f b)
map2 = map <<< map

map2 @Array show
```

PureFunctor:
@JordanMartinez That's correct, although the goal is to have `@`-vars not shadow vars in the class head, which I'm trying to figure out. Alternatively, we could also probably do the following syntax
```hs
class Functor @f where
  map :: forall a b. (a -> b) -> (f a -> f b)
```
And this would work just fine with:
```hs
mapArray = map @Array
```
On the other hand, could we do this for data constructors as well?
```hs
data Either @a @b = Left a | Right b

left = Left @(Proxy "this binds to b") 0

right = Right @(Proxy "this binds to a") 0

maybeLeft = Left @(Proxy "this binds to a") @(Proxy "this binds to b")

maybeRight = Right @(Proxy "this binds to a") @(Proxy "this binds to b")
```

PureFunctor:
The following case doesn't work yet:
```hs
mkProxy :: forall @a. Proxy a
mkProxy = Proxy

-- is actually

mkProxy :: forall k (@a :: k). Proxy a
mkProxy = Proxy

-- and when inferring this:

x = mkProxy @Type

-- hoist: forall k (@a :: k) -> forall (@a :: k) k
-- unify: Type, k
-- fails!

-- what should probably happen:

-- instantiate: forall k (@a :: k) -> forall (@a :: k?)
-- unify: Type, k?
-- passes!
```

JordanMartinez:
> I personally don't see the merit of supporting this syntax, though that may change in the next few weeks provided that there's practical use-cases for it. That said, it's also possible to just create an alias if this is really needed.

What if we left implementing that syntax for a later PR? I'm not sure how often people would want that and there is an identified workaround.

PureFunctor:
> What if we left implementing that syntax for a later PR? I'm not sure how often people would want that and there is an identified workaround.
Yeah, this sounds reasonable.

This PR should be ready for the first sets of reviews now as well. Wildcard syntax for skipping bindings works out-of-the-box although I've yet to figure out how to make it so warnings aren't emitted.

CHANGELOG.d/feature_visible_type_applications.md
```diff
@@ -0,0 +1,58 @@
...
+
+    forall a. a -> a
+
+  cannot be applied to the type:
+
+    Int
+
+
+while inferring the type of id
+in value declaration id'
```
JordanMartinez, line 37: I'm finding this example to be a bit confusing. Your first example implies that it compiles:
```
id :: forall @a. a -> a
id a = a

id' :: Int -> Int
id' = id @Int
```
But this error, which shows the code doesn't compile, seems to refer back to your first example because it ends with
```
while inferring the type of id
in value declaration id'
```

CHANGELOG.d/feature_visible_type_applications.md
```diff
@@ -0,0 +1,58 @@
...
+
+See https://github.com/purescript/documentation/blob/master/errors/CannotApplyExpressionOfTypeOnType.md for more information,
+or to contribute content related to this error.
+```
+
```
JordanMartinez, line 42: There's no mention of `@_` here. Should that be covered next, so that your `Either` example can use it (e.g. your `leftSkip` test)?

lib/purescript-cst/src/Language/PureScript/Types.hs
```diff
@@ -222,8 +238,9 @@ typeToJSON annToJSON ty =
       variant "TypeApp" a (go b, go c)
     KindApp a b c ->
       variant "KindApp" a (go b, go c)
-    ForAll a b c d e ->
+    ForAll a b c d e _ ->
       case c of
+        -- TODO: Restore these in the future
         Nothing -> variant "ForAll" a (b, go d, e)
         Just k -> variant "ForAll" a (b, go k, go d, e)
```
JordanMartinez, line 245: Could you clarify what the TODO means?

lib/purescript-cst/src/Language/PureScript/Types.hs
```diff
@@ -311,12 +339,13 @@ typeFromJSON defaultAnn annFromJSON = A.withObject "Type" $ \o -> do
...
+        -- TODO: Restore these in the future
         withoutMbKind = do
           (b, c, d) <- contents
-          ForAll a b Nothing <$> go c <*> pure d
+          ForAll a b Nothing <$> go c <*> pure d <*> pure NotVtaTypeVar
         withMbKind = do
           (b, c, d, e) <- contents
-          ForAll a b <$> (Just <$> go c) <*> go d <*> pure e
+          ForAll a b <$> (Just <$> go c) <*> go d <*> pure e <*> pure NotVtaTypeVar
       withMbKind <|> withoutMbKind
```
JordanMartinez, line 349: Another TODO here.

JordanMartinez (commented):
Just a quick glance through.

lib/purescript-cst/src/Language/PureScript/Types.hs
PureFunctor: 8627ab110f70c565eabc03ad1b454dfec09b604d reverts some changes in the serialization/deserialization of types to make tests pass, since the addition of the extra field is a breaking change.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -479,16 +479,13 @@ infer' (VisibleTypeApp val typeArg) = do
...
+  (typeArg', _) <- kindOf typeArg
   typeArg'' <- introduceSkolemScope <=< replaceAllTypeSynonyms <=< replaceTypeWildcards $ typeArg'
   case unconsVtaTypeVar valTy' of
-    Just ((typeVar, typeVarKind), tyAbsBody) -> do
-      unifyKinds typeArgKind typeVarKind
-      let valTy'' = replaceTypeVars typeVar typeArg'' tyAbsBody
-      pure $ TypedValue' True val'' valTy''
+    Just (typeVar, valTyTail) -> do
+      let valTy'' = replaceTypeVars typeVar typeArg'' valTyTail
+      TypedValue' True val'' <$> checkKind valTy'' kindType
```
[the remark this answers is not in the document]
PureFunctor: There's also this edge case:
```hs
module Main where

data Proxy :: forall k. k -> Type
data Proxy a = Proxy

xs :: forall (@f :: forall (a :: Type). a -> Type). Proxy f
xs = Proxy

xs' = xs @Array
```
```hs
purs: An internal error occurred during compilation: inferKind: unkinded forall binder
  error, called at src/Language/PureScript/Crash.hs:23:3 in purescript-cst-0.4.0.0-5z45WRzhCYCtwIomkySF:Language.PureScript.Crash
  internalError, called at src/Language/PureScript/TypeChecker/Kinds.hs:221:11 in purescript-0.14.5-2DsYJvu2Zok6cVlb0XnWjx:Language.PureScript.TypeChecker.Kinds
  inferKind, called at src/Language/PureScript/TypeChecker/Kinds.hs:298:25 in purescript-0.14.5-2DsYJvu2Zok6cVlb0XnWjx:Language.PureScript.TypeChecker.Kinds
  checkKind, called at src/Language/PureScript/TypeChecker/Kinds.hs:251:13 in purescript-0.14.5-2DsYJvu2Zok6cVlb0XnWjx:Language.PureScript.TypeChecker.Kinds
  inferAppKind, called at src/Language/PureScript/TypeChecker/Kinds.hs:213:7 in purescript-0.14.5-2DsYJvu2Zok6cVlb0XnWjx:Language.PureScript.TypeChecker.Kinds
  inferKind, called at src/Language/PureScript/TypeChecker/Kinds.hs:298:25 in purescript-0.14.5-2DsYJvu2Zok6cVlb0XnWjx:Language.PureScript.TypeChecker.Kinds
  checkKind, called at src/Language/PureScript/TypeChecker/Types.hs:488:34 in purescript-0.14.5-2DsYJvu2Zok6cVlb0XnWjx:Language.PureScript.TypeChecker.Types
```

tests/purs/passing/VisibleTypeApplications.purs
```diff
@@ -78,4 +78,16 @@ leftSkip = Left @_ @String 0
 rightSkip :: Either Int String
 rightSkip = Right @Int @_ "0"
 
+hrk :: forall (@f :: forall k. k -> Type). Proxy f
+hrk = Proxy
+
+hrk' :: Proxy Proxy
+hrk' = hrk @Proxy
```
PureFunctor, line 85: I mentioned this in the PureScript Discord, but this is unlike doing:
```hs
hrk' :: Proxy Proxy
hrk' = hrk
```
Which fails with the following error:
```hs
  Could not match kind
  
    forall k. k -> Type
  
  with kind
  
    forall k. k -> Type
```
I suspect this is because of impredicativity of some sort.

JordanMartinez (commented):
I read through this PR, commit by commit. Each commit had a very clean diff, so props to you on that!

Overall, this looks very good.

JordanMartinez (changes requested):
I think we should add `@` tyvars to `Coercible` and include a test for it. I also have one question about the error message.

JordanMartinez:
After I merged this PR into `master` locally, this test fails to compile due to a parsing error:
```purescript
module Main where

import Prelude
import Effect.Console (log)

class MultiKind :: (Type -> Type -> Type) -> Constraint
class MultiKind @f where
  foo :: forall @a @b. f a b -> String

data Tuplet1 @a @b @c = Tuplet1 a b c

instance MultiKind (Tuplet1 a)
  foo _ = "Tuplet1"
     -- ^ Unable to parse module; unexpected '='
main = log "Done"
```

PureFunctor:
@JordanMartinez Isn't the instance declaration missing a `where`?
```purs
instance MultiKind (Tuplet1 a) where
  foo _ = "Tuplet1"
```

PureFunctor:
On that note, I've found some more behaviour to be investigated in that example:
```purescript
module Main where

class MultiKind :: (Type -> Type -> Type) -> Constraint
class MultiKind @f where
  foo :: forall @a @b. f a b -> String

data Tuplet1 @a @b @c = Tuplet1 a b c

instance MultiKind (Tuplet1 a) where
  foo _ = "Tuplet1"

foo' :: forall a. a -> String
foo' a = foo @(Tuplet1 a) @String @String (Tuplet1 a "" "")
```
This fails with:
```
Error found:
in module Main
at Main.purs:13:44 - 13:59 (line 13, column 44 - line 13, column 59)

  Could not match type
      
    a0
      
  with type
     
    a
     
while trying to match type Tuplet1 a0
  with type Tuplet1 a
while checking that expression ((Tuplet1 a) "") ""
  has type Tuplet1 a String String
in value declaration foo'

```

Looks like I just needed to add skolemization for `VisibleTypeApp`:
```hs
diff --git a/src/Language/PureScript/TypeChecker/Skolems.hs b/src/Language/PureScript/TypeChecker/Skolems.hs
index 7cbe20ce0..75d9b7be8 100644
--- a/src/Language/PureScript/TypeChecker/Skolems.hs
+++ b/src/Language/PureScript/TypeChecker/Skolems.hs
@@ -63,6 +63,8 @@ skolemizeTypesInValue ann ident mbK sko scope =
       | ident `notElem` sco = return (sco, DeferredDictionary c (map (skolemize ann ident mbK sko scope) ts))
     onExpr sco (TypedValue check val ty)
       | ident `notElem` sco = return (sco ++ peelTypeVars ty, TypedValue check val (skolemize ann ident mbK sko scope ty))
+    onExpr sco (VisibleTypeApp val ty)
+      | ident `notElem` sco = return (sco ++ peelTypeVars ty, VisibleTypeApp val (skolemize ann ident mbK sko scope ty))
     onExpr sco other = return (sco, other)
 
     onBinder :: [Text] -> Binder -> Identity ([Text], Binder)
```

PureFunctor:
I'll fix up the merge conflicts once #4267 is resolved, since this PR adds an extra field to the type class data internals.

JordanMartinez:
Does that also fix this error? With the type signature commented out, the code compiles. With the type signature uncommented, I get the below error.
```purs
module Main where

import Prelude
import Effect.Console (log)

class MultiKind :: (Type -> Type -> Type) -> Constraint
class MultiKind @f where
  foo :: forall @a @b. f a b -> String

data Tuplet1 @a @b @c = Tuplet1 a b c
data TupletTuplet @a @b @c = TupletTuplet a b c

instance MultiKind (Tuplet1 a) where
  foo _ = "Tuplet1"

instance MultiKind (TupletTuplet a) where
  foo _ = "TupletTuple"

-- testTuplet6 :: ?Help
testTuplet6 = foo @_ @Int @Boolean

main = log "Done"

{-
Error found:
in module Main
at tests/purs/passing/Vta.purs:29:1 - 29:21 (line 29, column 1 - line 29, column 21)
     
    Main.MultiKind t1
                     
while checking that expression foo
  has type t0
in value declaration testTuplet6
     
where t0 is an unknown type
      t1 is an unknown type
-}
```
I would expect the inferred type signature to be something like below. `f` isn't a visible type application because it was already skipped via `@_`:
```purs
testTuplet6 :: forall f. MultiKind f => f Int Boolean -> String
testTuplet6 = foo
```

lib/purescript-cst/src/Language/PureScript/Environment.hs
```diff
@@ -464,7 +468,7 @@ primCoerceClasses =
     [ (primSubName C.moduleCoerce "Coercible", makeTypeClassData
         [ ("a", Just (tyVar "k"))
         , ("b", Just (tyVar "k"))
-        ] [] [] [] True)
+        ] [] [] [] [] True)
```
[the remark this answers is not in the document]
JordanMartinez: Oh, right.

I've opened an issue to track that there: https://github.com/purescript/purescript-safe-coerce/issues/13

JordanMartinez:
Some other tests I tried.

<details>
<summary>I tested how visible ty vars work when there's a super class relationship. The below code still compiles even when `BaseHasVTA`'s `@a` is changed to `a`</summary>

```purs
module Main where

import Prelude
import Effect.Console (log)

class BaseHasVTA @a where
  baseHasVTA :: a -> String

class BaseHasVTA a <= SuperHasVTA @a where
  superHasVTA :: a -> a -> String

instance BaseHasVTA Int where
  baseHasVTA _ = "base"

instance SuperHasVTA Int where
  superHasVTA a b = baseHasVTA a <> " " <> baseHasVTA b

testSuperclasses :: String
testSuperclasses = superHasVTA @Int 1 2

main = log "Done"
```
</details>

PureFunctor:
> I would expect the inferred type signature to be something like below. `f` isn't a visible type application because it was already skipped via `@_`:

I think this happens because of the call to `instantiatePolyTypeWithUnknowns` during inference for visible type applications, which makes it so that the compiler attempts to remove constraints instantly. The current solution I have for `@_` syntax is to just let it be a type wildcard that doesn't ever warn, though I could also make it so it's treated specially such that it really does a "skip" operation.


> I tested how visible ty vars work when there's a super class relationship. The below code still compiles even when `BaseHasVTA`'s `@a` is changed to `a`

This makes sense to me semantically. Do you mean that if `BaseHasVTA` is used as a superclass, it _should_ still have the `@` prefixes?
```hs
class BaseHasVTA @a <= SuperHasVTA @a where
  superHasVTA :: a -> a -> String
```

PureFunctor:
On migrating from `Proxy` arguments to visible type applications, I'd say that it's in the best interest of library authors to still keep the `Proxy` argument while also allowing the type variables to be type-applied. For example, `barlow-lens` can make a minor release such that:
```purescript
barlow (Proxy :: _ "field1.field2")

-- is the same as

barlow @"field1.field2" Proxy
```

In the future, a major release can be made such that the `Proxy` argument is erased entirely.

On the other hand, I'm not entirely sure about what we'd do for `core` libraries. `prelude` recently had the compiler-solved `Reflectable` type class with its `reflectType` method, which uses a `Proxy` (https://github.com/purescript/purescript-prelude/pull/289) argument. Updating this to use visible type applications also means having to change the instance generated by the compiler though.

If this PR makes it in time for `0.15.0`, should we set precedence by having `reflectType` use visible type applications right off the bat?
```hs
class Reflectable :: forall k. k -> Type -> Constraint
class Reflectable @v t | v -> t where
  reflectType :: t
```

JordanMartinez:
> This makes sense to me semantically. Do you mean that if BaseHasVTA is used as a superclass, it should still have the @ prefixes?

No, I was merely pointing out that the code compiled as expected whether the `@` was there or not on the super class. The code I provided in my original wasn't included in the test, so I was explaining that for other reviewers.
```purs
-- whether BaesHasVTA is defined as this...
class BaseHasVTA @a where
  baseHasVTA :: a -> String

-- or this...
class BaseHasVTA a where
  baseHasVTA :: a -> String
  
-- SuperHasVTA continues to work with either version of
-- BaseHasVTA above.
class BaseHasVTA a <= SuperHasVTA @a where
  superHasVTA :: a -> a -> String
```

JordanMartinez:
> If this PR makes it in time for 0.15.0, should we set precedence by having reflectType use visible type applications right off the bat?

I think so! That will be one less breaking change we have to make later.

PureFunctor:
> No, I was merely pointing out that the code compiled as expected whether the `@` was there or not on the super class. The code I provided in my original wasn't included in the test, so I was explaining that for other reviewers.

Ah, I see

JordanMartinez:
> The current solution I have for @_ syntax is to just let it be a type wildcard that doesn't ever warn, though I could also make it so it's treated specially such that it really does a "skip" operation.

Hmm... I'll have to look at the code in more detail tomorrow.

src/Language/PureScript/Errors.hs
```diff
@@ -1353,12 +1353,29 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
...
         , "cannot be applied to the type:"
         , markCodeBox $ indent $ prettyType targ
-        ]
+        ] <> comments
+      where
+      comments =
+        case tabs of
+          ForAll a i k t s _
+            | not $ hasVta tabs
+            , isMonoType t ->
```
PureFunctor, line 1367: I've decided to only suggest adding a visible type abstraction when there's a single `ForAll` as it's irrefutable that this is what the user wants.

PureFunctor: One problem here though is the specialized syntax for data constructors and class declarations, so maybe just having a plain suggestion here would be better.

JordanMartinez:
Could you add this test?
```purs
-- whether `BaseNoVta` or `BaseVta`
-- have visible type application type variable
-- for `a` or not`, a super class still works
class BaseNoVta a where
  baseNoVta :: a -> String

class BaseNoVta a <= Super1 @a where
  super1 :: a -> a -> String

class BaseVta @a where
  baseVta :: a -> String

class BaseVta a <= Super2 @a where
  super2 :: a -> a -> String

instance BaseNoVta String where
  baseNoVta x = x

instance Super1 String where
  super1 = (<>)

instance BaseVta String where
  baseVta x = x

instance Super2 String where
  super2 = (<>)

testClassRelationship :: String
testClassRelationship = super1 "foo" "bar" <> super2 "foo" "bar"
```

JordanMartinez:
I found a few more cases where parsing fails with 'unexpected token: `@`'. 
```purs
whereTest :: forall a b. a -> b -> String
whereTest a b = foo' @a a <> bar' @b b
  where
  foo' :: forall @a. a -> String
              -- ^ Unexpected token
  foo' _ = "foo"

  bar' :: forall @b. b -> String
  bar' _ = "bar"

type Foo = forall @a. { foo :: a }
               -- ^ Unexpected token

type Foo2 = { foo :: forall @a. a -> String }
                        -- ^ Unexpected token

testRank2 :: (forall @b. b -> String) -> String
                  -- ^ Unexpected token
testRank2 f = f 1 <> f true

testRank2' :: String
testRank2' = testRank2 (\_ -> "rank2")
```
I fixed it by replacing `typeVar`'s definition with `typeVarWithVta` and updating various places to use `type`. I also noticed that `newtype` and `type` are still using `typeVar`, not `typeVarWithVta`.
```diff
diff --git a/lib/purescript-cst/src/Language/PureScript/CST/Parser.y b/lib/purescript-cst/src/Language/PureScript/CST/Parser.y
index dc938fdb..687c50d1 100644
--- a/lib/purescript-cst/src/Language/PureScript/CST/Parser.y
+++ b/lib/purescript-cst/src/Language/PureScript/CST/Parser.y
@@ -292,14 +292,6 @@ type1 :: { Type () }
   : type2 { $1 }
   | forall many(typeVarBinding) '.' type1 { TypeForall () $1 $2 $3 $4 }
 
-typeWithVta :: { Type () }
-  : type1WithVta %shift { $1 }
-  | type1WithVta '::' type { TypeKinded () $1 $2 $3 }
-
-type1WithVta :: { Type () }
-  : type2 { $1 }
-  | forall many(typeVarBindingWithVta) '.' type1 { TypeForall () $1 $2 $3 $4 }
-
 type2 :: { Type () }
   : type3 %shift { $1 }
   | type3 '->' type1 { TypeArr () $1 $2 $3 }
@@ -355,10 +347,6 @@ rowLabel :: { Labeled Label (Type ()) }
   : label '::' type { Labeled $1 $2 $3 }
 
 typeVarBinding :: { TypeVarBinding () }
-  : ident { TypeVarName Nothing $1 }
-  | '(' ident '::' type ')' {% checkNoWildcards $4 *> pure (TypeVarKinded Nothing (Wrapped $1 (Labeled $2 $3 $4) $5)) }
-
-typeVarBindingWithVta :: { TypeVarBinding () }
   : ident { TypeVarName Nothing $1 }
   | '@' ident { TypeVarName (Just $1) $2 }
   | '(' ident '::' type ')' {% checkNoWildcards $4 *> pure (TypeVarKinded Nothing (Wrapped $1 (Labeled $2 $3 $4) $5)) }
@@ -374,7 +362,7 @@ exprWhere :: { Where () }
 
 expr :: { Expr () }
   : expr1 %shift { $1 }
-  | expr1 '::' typeWithVta { ExprTyped () $1 $2 $3 }
+  | expr1 '::' type { ExprTyped () $1 $2 $3 }
 
 expr1 :: { Expr () }
   : expr2 %shift { $1 }
@@ -682,7 +670,7 @@ decl :: { Declaration () }
   | 'type' properName '::' type {% checkNoWildcards $4 *> pure (DeclKindSignature () $1 (Labeled (getProperName $2) $3 $4)) }
   | 'derive' instHead { DeclDerive () $1 Nothing $2 }
   | 'derive' 'newtype' instHead { DeclDerive () $1 (Just $2) $3 }
-  | ident '::' typeWithVta { DeclSignature () (Labeled $1 $2 $3) }
+  | ident '::' type { DeclSignature () (Labeled $1 $2 $3) }
   | ident manyOrEmpty(binderAtom) guardedDecl { DeclValue () (ValueBindingFields $1 $2 $3) }
   | fixity { DeclFixity () $1 }
   | 'foreign' 'import' ident '::' type {% when (isConstrained $5) (addFailure ([$1, $2, nameTok $3, $4] <> toList (flattenType $5)) ErrConstraintInForeignImportSyntax) *> pure (DeclForeign () $1 $2 (ForeignValue (Labeled $3 $4 $5))) }
@@ -690,7 +678,7 @@ decl :: { Declaration () }
   | 'type' 'role' properName many(role) { DeclRole () $1 $2 (getProperName $3) $4 }
 
 dataHead :: { DataHead () }
-  : 'data' properName manyOrEmpty(typeVarBindingWithVta) { DataHead $1 (getProperName $2) $3 }
+  : 'data' properName manyOrEmpty(typeVarBinding) { DataHead $1 (getProperName $2) $3 }
 
 typeHead :: { DataHead () }
   : 'type' properName manyOrEmpty(typeVarBinding) { DataHead $1 (getProperName $2) $3 }
@@ -730,7 +718,7 @@ classSuper :: { (OneOrDelimited (Constraint ()), SourceToken) }
   : constraints '<=' {%^ revert $ pure ($1, $2) }
 
 classNameAndFundeps :: { (Name (N.ProperName 'N.ClassName), [TypeVarBinding ()], Maybe (SourceToken, Separated ClassFundep)) }
-  : properName manyOrEmpty(typeVarBindingWithVta) fundeps {%^ revert $ pure (getProperName $1, $2, $3) }
+  : properName manyOrEmpty(typeVarBinding) fundeps {%^ revert $ pure (getProperName $1, $2, $3) }
 
 fundeps :: { Maybe (SourceToken, Separated ClassFundep) }
   : {- empty -} { Nothing }
@@ -741,7 +729,7 @@ fundep :: { ClassFundep }
   | many(ident) '->' many(ident) { FundepDetermines $1 $2 $3 }
 
 classMember :: { Labeled (Name Ident) (Type ()) }
-  : ident '::' typeWithVta {% checkNoWildcards $3 *> pure (Labeled $1 $2 $3) }
+  : ident '::' type {% checkNoWildcards $3 *> pure (Labeled $1 $2 $3) }
 
 instHead :: { InstanceHead () }
   : 'instance' constraints '=>' qualProperName manyOrEmpty(typeAtom)
```

JordanMartinez:
When I incorporate the parser fixes for `newtype`, I came across a compiler error. Given the below code:
```purs
newtype NewtypeWithVTA @f @a = NewtypeWithVTA (f a)

useNewtype :: NewtypeWithVTA Array Int
useNewtype = mkNewtype [ 1, 2 ]
  where
    mkNewtype = NewtypeWithVTA @Array @Int
--              ^^^^^^^^^^^^^^^^^^^^^
```
I get this compiler error:
```
  An expression of type:
                                         
    t181 t182 -> NewtypeWithVTA t181 t182
                                         
  cannot be applied to the type:
         
    Array
         
  as it has no visible type abstraction.

while inferring the type of NewtypeWithVTA
in value declaration useNewtype
```
Interestingly, the same code using `data` or `type` does not produce such an error:
```purs
data DataVta2 @f @a = DataVta2 (f a)

useData2 :: DataVta2 Array Int
useData2 = mkDataVta2 [1]
  where
  mkDataVta2 = DataVta2 @Array @Int

type TypeAliasVta @f @a = f a

useTypeAliasVta :: TypeAliasVta Array Int
useTypeAliasVta = mkTypeAliasVta @Array @Int [1]
  where
  mkTypeAliasVta :: forall @f @a. f a -> TypeAliasVta f a
  mkTypeAliasVta fa = fa
```

JordanMartinez:
> If this PR makes it in time for `0.15.0`, should we set precedence by having `reflectType` use visible type applications right off the bat?
> 
> ```haskell
> class Reflectable :: forall k. k -> Type -> Constraint
> class Reflectable @v t | v -> t where
>   reflectType :: t
> ```

I've created https://github.com/purescript/purescript-prelude/pull/290 in case you want to test against that branch of prelude.

PureFunctor:
The parser change to support `newtype` would have to be accompanied with a change in the CST conversion as well, otherwise, the compiler isn't aware that the newtype constructor has a visible type abstraction.

Also, I don't think `@`-prefixes should be allowed in type synonyms, as they don't directly affect things that exist at the term-level (i.e. constructors for `data`/`newtype`, or type class members).

JordanMartinez:
> The parser change to support newtype would have to be accompanied with a change in the CST conversion as well, otherwise, the compiler isn't aware that the newtype constructor has a visible type abstraction.

Ah.... that's why.

JordanMartinez:
Testing this PR against the `prelude` branch I pushed, I get the below error:
```
[1/1 UnusableDeclaration] src/Data/Symbol.purs:11:1

      v
  11  class IsSymbol @sym where
  12    reflectSymbol :: String
                              ^
  
  The declaration reflectSymbol contains arguments that couldn't be determined.
  These arguments are: { sym }
  
  in type class declaration for IsSymbol
```
In other words, if we drop the `Proxy` arg and supply that information via visible type args, then when the compiler sees a type var in a class head that isn't mentioned in its body, it needs to do two things:
- verify that the type var is a VTA
- verify that calling that class' member always has a VTA for that parameter. If it gets skipped via `@_`, then the type signature should propagate the VTA requirement to the next ty var.

JordanMartinez:
> Also, I don't think @-prefixes should be allowed in type synonyms, as they don't directly affect things that exist at the term-level (i.e. constructors for data/newtype, or type class members).

That makes sense. Adding support for it might also cause issues due to the interaction between type aliases and type classes.

natefaubion:
I personally do not think we should push through major changes to APIs for type applications yet, and I don't think we should add type application support to every API. Type applications are a contract, like argument order. That is _a lot_ of churn, and is not the primary focus of this release. There's a lot of opportunity to make bad choices that are hard to go back on.

PureFunctor:
I stand by https://github.com/purescript/purescript/pull/4235#issuecomment-1071927202 in that I don't think the switch should be instantaneous, but setting precedence for a new addition like `Data.Reflectable` and letting the ecosystem catch up with the feature should be the way to go.

Similarly, existing code can be made compatible with visible type applications by just aliasing and adding a type annotation.

lib/purescript-cst/src/Language/PureScript/CST/Types.hs
```diff
@@ -153,8 +153,8 @@ data Type a
   deriving (Show, Eq, Ord, Functor, Foldable, Traversable, Generic)
 
 data TypeVarBinding a
-  = TypeVarKinded (Wrapped (Labeled (Name Ident) (Type a)))
-  | TypeVarName (Name Ident)
+  = TypeVarKinded (Maybe SourceToken) (Wrapped (Labeled (Name Ident) (Type a)))
```
natefaubion, line 156: I don't know if it's clear, but in all other cases we preserve left-to-right order of tokens in the CST, which would imply:
```haskell
TypeVarKinded (Wrapped (Labeled (Maybe SourceToken, Name Ident) (Type a))
```

lib/purescript-cst/src/Language/PureScript/CST/Flatten.hs
```diff
@@ -303,8 +304,9 @@ flattenRow (Row lbls tl) =
 
 flattenTypeVarBinding :: TypeVarBinding a -> DList SourceToken
 flattenTypeVarBinding = \case
-  TypeVarKinded a -> flattenWrapped (flattenLabeled (pure . nameTok) flattenType) a
-  TypeVarName a -> pure $ nameTok a
+  TypeVarKinded _ a -> flattenWrapped (flattenLabeled (pure . nameTok) flattenType) a
```
natefaubion, line 307: I think this is dropping the `atSign`.

lib/purescript-cst/src/Language/PureScript/AST/Declarations.hs
```diff
@@ -370,7 +370,7 @@ data Declaration
   -- |
   -- A data type declaration (data or newtype, name, arguments, data constructors)
   --
-  = DataDeclaration SourceAnn DataDeclType (ProperName 'TypeName) [(Text, Maybe SourceType)] [DataConstructorDeclaration]
+  = DataDeclaration SourceAnn DataDeclType (ProperName 'TypeName) [(Text, Maybe SourceType)] [VtaTypeVar] [DataConstructorDeclaration]
```
natefaubion, line 373: What was the reason for putting these in a separate spine? Did you consider something like `[(Text, Maybe SourceType, VtaTypeVar)]`?

lib/purescript-cst/src/Language/PureScript/Environment.hs
```diff
@@ -52,6 +52,9 @@ data TypeClassData = TypeClassData
   { typeClassArguments :: [(Text, Maybe SourceType)]
   -- ^ A list of type argument names, and their kinds, where kind annotations
   -- were provided.
+  , typeClassVtaTypeVars :: [VtaTypeVar]
```
natefaubion, line 55: Same question here regarding the separate spine vs adding it to `typeClassArguments`.

src/Language/PureScript/Externs.hs
```diff
@@ -137,6 +137,7 @@ data ExternsDeclaration =
   | EDClass
       { edClassName               :: ProperName 'ClassName
       , edClassTypeArguments      :: [(Text, Maybe SourceType)]
+      , edClassVtaTypeVars         :: [VtaTypeVar]
```
natefaubion, line 140: Nit
```suggestion
      , edClassVtaTypeVars        :: [VtaTypeVar]
```

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -299,9 +300,15 @@ typeClassMemberToDictionaryAccessor mn name args (TypeDeclaration (TypeDeclarati
...
-       moveQuantifiersToFront (quantify (srcConstrainedType (srcConstraint className [] (map (srcTypeVar . fst) args) Nothing) ty))
+       makeVta (moveQuantifiersToFront (quantify (srcConstrainedType (srcConstraint className [] (map (srcTypeVar . fst) args) Nothing) ty)))
     )]
-typeClassMemberToDictionaryAccessor _ _ _ _ = internalError "Invalid declaration in type class definition"
+  where
+  vtaVars = S.fromList $ map fst $ filter ((== IsVtaTypeVar) . snd) $ zip (map fst args) vtas
+  makeVta = everywhereOnTypes go
+    where
+    go (ForAll a b c d e _) | b `S.member` vtaVars = ForAll a b c d e IsVtaTypeVar
+    go t = t
```
natefaubion, line 310: I don't think this rewrite is handling scope (ie, it's assuming all forall binder names are unique). Is it possible for shadowed binders to potentially mess with it?

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -620,9 +622,15 @@ inferDataDeclaration moduleName (ann, tyName, tyArgs, ctors) = do
       let tyCtorName = srcTypeConstructor $ mkQualified tyName moduleName
           tyCtor = foldl (\ty -> srcKindApp ty . srcTypeVar . fst . snd) tyCtorName sigBinders
           tyCtor' = foldl (\ty -> srcTypeApp ty . srcTypeVar . fst) tyCtor tyArgs'
+          vtvs' = S.fromList $ mapMaybe findVtv $ zip tyArgs vtvs where
+            findVtv ((i, _), IsVtaTypeVar) = Just i
+            findVtv _ = Nothing
+          makeVta = everywhereOnTypes go where
+            go (ForAll a b c d e _) | b `S.member` vtvs' = ForAll a b c d e IsVtaTypeVar
+            go t = t
```
natefaubion, line 630: Similar question here. Additionally, should this function be factored out? It may help if it's possible to go subtlety wrong.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -462,6 +477,36 @@ infer' (TypedValue checkType val ty) = do
...
+  --
+  -- forall (k :: Type) (@t :: k). Proxy t
+  --
+  -- are turned into:
+  --
+  -- forall (@t :: k?). Proxy t
+  --
+  -- This allows typeArg's kind to unify with the unknown k.
+  (val'', valTy') <- instantiatePolyTypeWithUnknowns val' valTy
+  -- `kindOf` here eliminates ParensInType in typeArg
```
natefaubion, line 500: Is kindOf the best way to do this? It seems like a fair bit of work if all you want is to drop ParensInType. Why is it necessary? Kind-checking something only discard the kind doesn't sit right to me. Why discard the kind? Should we be using it elsewhere? Is there kind unification that needs to happen?

lib/purescript-cst/src/Language/PureScript/AST/Declarations.hs
PureFunctor: Just to reduce breakage for the initial implementation.

src/Language/PureScript/TypeChecker/Types.hs
PureFunctor: Initially, there was kind unification in this routine, though I found that it was pretty inconsistent with certain cases (https://github.com/purescript/purescript/pull/4235#discussion_r815294743). If I remember correctly, `inferKind` which is called downstream, also elaborates the types it was passed so I'm making use of that in this routine as well.

src/Language/PureScript/Sugar/TypeClasses.hs
PureFunctor: It's highly likely, yeah. I'll factor this out as a helper function.

lib/purescript-cst/src/Language/PureScript/AST/Declarations.hs
PureFunctor: I'm currently working on shifting the representation to this and I've figured that this encoding is the best approach at the moment. There's too many moving parts in the type-checker code to reliably refactor tuple-based encodings like this, and frankly, it's a bit frustrating to do so /rant.

natefaubion:
I think this looks really good. Nice work! Some followup questions:

* Do we want to go ahead and add visible kind applications? Is there a reason not to?
* Is there any followup work to do with doc generation? Do we need to add vtas to pursuit rendering?

JordanMartinez:
>     * Is there any followup work to do with doc generation? Do we need to add vtas to pursuit rendering?

I'm pretty sure they're already being rendered in this PR. See https://github.com/purescript/purescript/pull/4235/files#diff-632cd22840fb5cdbf387061e7984756ac9575f8c38b03f370f1ca054b299f96aR241-R247

natefaubion:
Ah, I missed that! Thanks!

PureFunctor:
@natefaubion 

1. For the most part, I don't think there would be a reason not to. But I suppose whether we want explicit abstractions to apply to kind applications as well is something to talk about. If it is the case, do we want type variables used in polykinds to be explicitly abstracted? For example:
```hs
-- Given:
data Proxy a = Proxy

-- Do we want to synthesize this, and make it so the abstraction isn't inherited by the type constructor e.g. Proxy :: forall k @a. Proxy a
data Proxy :: forall @k. k -> Type
data Proxy @a = Proxy

type SProxy = Proxy @Symbol

-- Or just use what we have now but expose the syntax for `KindApp`?
data Proxy :: forall k. k -> Type
data Proxy @a = Proxy

type SProxy = Proxy @Symbol
```
Also, what about `Constraint`s?

2. There is indeed work done on doc rendering, though I have a TODO in there for JSON serialization/deserialization that I'll be fixing up soon-ish.

lib/purescript-cst/src/Language/PureScript/Types.hs
```diff
@@ -504,6 +539,23 @@ replaceAllTypeVars = go [] where
...
+    | otherwise =
+        ForAll ann arg (go False <$> mbK) (go False ty) sco vta
+  go _ (ConstrainedType ann c ty) = ConstrainedType ann (mapConstraintArgsAll (map $ go False) c) (go False ty)
+  go _ (TypeApp ann t1 t2) = TypeApp ann (go False t1) (go False t2)
+  go _ (KindApp ann t1 t2) = KindApp ann (go False t1) (go False t2)
+  go _ (RCons ann name ty rest) = RCons ann name (go False ty) (go False rest)
+  go _ (KindedType ann ty k) = KindedType ann (go False ty) (go False k)
+  go _ (BinaryNoParensType ann t1 t2 t3) = BinaryNoParensType ann (go False t1) (go False t2) (go False t3)
+  go n (ParensInType ann t) = ParensInType ann (go n t)
+  go _ ty = ty
```
rhendric, line 557: Isn't `go False` the identity function (assuming termination)? And if so, seems like the boolean argument, and all of the `go _` cases, are unnecessary?

PureFunctor: This is true.

lib/purescript-cst/src/Language/PureScript/Types.hs
```diff
@@ -43,6 +43,17 @@ instance NFData SkolemScope
 instance Serialise SkolemScope
 
 -- |
+-- Metadata for determining whether or not a type variable is used for
+-- visible type applications.
+--
+data VtaTypeVar
+  = IsVtaTypeVar
+  | NotVtaTypeVar
```
natefaubion, line 51: What do you think about adding another variant here `IsVtaTypeVarRequired`, which can be derived from the result of an ambiguity check. Then whenever instantiating a required vta implicitly, we can throw an appropriate error?

PureFunctor:
CI failure on MacOS seems to be because of cached externs. Tests run fine on my machine.

PureFunctor:
Remaining work:
* Refactoring the usability check for type classes now that type variables in class heads are always available for visible type applications, such that it warns for cases like:
```hs
-- Warning: the following type variables, `foo`, ...,  may be an ambiguous for the following
-- members, and are required to be eliminated through a type application.
--
--    foo :: String
--
class Foo foo where
  foo :: String
```
* Exposing syntax for kind applications

PureFunctor:
> turns out there was a pretty bad bug that I didn't catch

There are _a lot_ of stuff I didn't catch 😓, though I have another approach to visible type applications that I have in mind which I should get to in a bit.

PureFunctor:
Minor comment that I forgot to add: the `Specified` type here holds unknowns instead of type variables because of how we do instantiations in the compiler to accommodate for row types.

PureFunctor:
A good chunk of changes in this PR involves getting around n-ary tuple types that are used in different types in the compiler. Now that #4351 is merged, I'd like to use `RecordDotSyntax` for the types that PR touches.

PureFunctor:
Moving this to 0.15.4 to give it a bit more time to cook 🍳

JordanMartinez:
@PureFunctor I just wanted to say that I won't be reviewing this PR. It's not because I don't want to, but because I don't think I understand enough to do a good review.

JordanMartinez:
@PureFunctor In case you missed it, see #4367

sigma-andex:
and another one: `fast-vect` updated to VTAs and works fine https://github.com/sigma-andex/purescript-fast-vect/tree/vta2

PureFunctor:
@sigma-andex I'm assuming that in this example, `vect` here stands for some `Vect` type that's kinded with `Int -> Type -> Type`? I don't think this is valid semantically, as `@len` implies that the kind for `Vect` should be `forall (a :: Int). Type -> Type` instead.
```hs
type Replicate vect len elem =
  Compare len NegOne GT
  ⇒ Reflectable len Int
  ⇒ elem
  → vect @len elem
```

PureFunctor:
As briefly discussed in the Discord server, here's a few things that'll need to be patched up:

1. Unkinded forall binder crashes, as caused by the following example, should probably be supplemented with a new error that clarifies why this form is not allowed.
```hs
type Apply f a = f @a

Error found:
in module Main
at Main.purs:5:1 - 5:45 (line 5, column 1 - line 5, column 45)

  Kind application of `a` into `f` is disallowed as `f` is not a polytype.
  Furthermore, you may run into `VisibleQuantificationCheckFailureInType`
  when adding a kind annotation.

in type synonym Apply
```

2. There's currently a bug with type operator desugaring for types inside of visible type applications.

3. Potentially add kind application syntax for instances, such that:
```hs
data P :: forall k. k -> Type
data P a

class F :: forall a. P a -> a -> Constraint
class F a b | a -> b

instance F @Kind1 TypeOfKind1 TypeOfKind1
```

i-am-the-slime:
I've been using this now for about three weeks in all of my projects and haven't run into any bugs. Maybe my code isn't crazy enough, but I find that very promising.

PureFunctor:
The macOS-10.15 action fails because of [this](https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/).

> Potentially add kind application syntax for instances, such that:

As for this, I think it's a bit out of scope, no? The original goal for this was to have type applications for expressions, and I've yet to refresh myself on what was discussed #3137 to see if there's anything to be added retroactively.

Could we start a review on this? Potentially to get it ready by `0.15.5` 😄

PureFunctor:
Cleaned up a few diffs now that `@` annotations aren't needed for data and newtype declarations. Most of the changes should be related to its use in class heads now.

I'll also work on this error:
```
-- Warning: the following type variables, `foo`, ...,  may be an ambiguous for the following
-- members, and are required to be eliminated through a type application.
--
--    foo :: String
--
class Foo foo where
  foo :: String
```

PureFunctor:
Because of the new semantics for visible type abstractions in type classes, I've removed the `UnusableDeclaration` error and replaced it with the `OnlyPartiallyDetermined` warning. It'll fire under the following conditions:

1. A type variable does not occur in any of the members
2. A type variable cannot be determined by a functional dependency
3. A type variable is not used in a superclass

Also the changelog entry for this feature is getting a bit long and more tutorial-like. I plan on writing an article for my blog in the future, as this feature is released. Would linking to that be better?

MonoidMusician:
I'm okay with the length of the changelog, since it's basically covering the various usages. It's not optional information.

I'll need to give more thought to the design. Definitely want to help move this along and make sure it's going to work well for users down the road.

My initial question is:
Is there any way to make different members have different visibility on type arguments? It might be nice for the case that only some members need a specific type variable. (I've definitely run into cases like that, where I either had to split it into different classes or add a proxy argument to one member, even though I initially thought it all fit into one class conceptually.)

Thinking about it a bit more, I'm not sure I like the idea of the required type arguments as opposed to optional ones. On its own it seems fine, but how it interacts with other features is less clear (e.g. my question above). Like how will we print them differently from non-required type arguments? What if you alias a method with an ordinary declaration – can it still have a required type argument?

Is it even necessary? Maybe we're trying to do too much work for the user with the feature. We already have `AmbiguousTypeVariables` and will still need to have it for inferred types. Do you think it's okay if users just see that error instead of a specialized error that the particular type application needs to be applied? Maybe we can do some more work to suggest where the type application would need to happen with that error, so we aren't losing specificity.

Of course it's okay to leave some stuff for future work, but ideally we can plan ahead so that it won't require breaking changes.

MonoidMusician:
> > Potentially add kind application syntax for instances, such that:
> 
> As for this, I think it's a bit out of scope, no? The original goal for this was to have type applications for expressions, and I've yet to refresh myself on what was discussed #3137 to see if there's anything to be added retroactively.

It would be great if we could support how things are printed currently, since you are doing that for polykinded types it would be great to also do it for polykinded instances. Lmk if you want help hacking on that, I don't anticipate too much difficulty given what you have already?

MonoidMusician:
There's something funny going on with constructors:
```purescript
> :t Tuple
forall (a1 :: Type) (b2 :: Type). a1 -> b2 -> Tuple a1 b2
```

Printing their type doesn't include the vta quantification, even though they obviously support it!

I would offer to fix it but I haven't found the right place yet :)

PureFunctor:
> There's something funny going on with constructors:
> 
> ```purescript
> > :t Tuple
> forall (a1 :: Type) (b2 :: Type). a1 -> b2 -> Tuple a1 b2
> ```
> 
> Printing their type doesn't include the vta quantification, even though they obviously support it!
> 
> I would offer to fix it but I haven't found the right place yet :)

I think this is because of how constructors are monomorphized _and then_ generalized back, which causes the loss of information. This was actually something that I had to get around with the `Specified` type approach.

PureFunctor:
> Is there any way to make different members have different visibility on type arguments? It might be nice for the case that only some members need a specific type variable. (I've definitely run into cases like that, where I either had to split it into different classes or add a proxy argument to one member, even though I initially thought it all fit into one class conceptually.)

Not at the moment, no. If we do support it though, what do you imagine would the syntax look like? I've tried having them on the members before as sort of an "override", like this, but this is potentially confusing.
```hs
class A a where
  a1 :: forall @a. a -> Int
  a2 :: a -> Int
```

> Thinking about it a bit more, I'm not sure I like the idea of the required type arguments as opposed to optional ones. On its own it seems fine, but how it interacts with other features is less clear (e.g. my question above). Like how will we print them differently from non-required type arguments? What if you alias a method with an ordinary declaration – can it still have a required type argument?

This is indeed something to consider. The implementation is definitely designed around just accepting loss of information, for example, skipping is implemented as just losing the visibility for a type argument; aliasing too, would have certain implications. For what it's worth, GHC tries to be a bit smarter in that it keeps track of which unification variables should generalize into visible arguments—but then, should we follow GHC? :smile:

> Is it even necessary? Maybe we're trying to do too much work for the user with the feature. We already have AmbiguousTypeVariables and will still need to have it for inferred types. Do you think it's okay if users just see that error instead of a specialized error that the particular type application needs to be applied? Maybe we can do some more work to suggest where the type application would need to happen with that error, so we aren't losing specificity.

For what it's worth, expanding `AmbiguousTypeVariables` to know whether or not it was invoked due to type variables being skipped is something I've considered doing before `OnlyPartiallyDetermined`. Also, type arguments being visible by default in class heads with no way to opt out is what brought about the removal of `UnusableDeclaration`.

With the implications in mind, it shouldn't be too hard making visibility opt-in for class heads. So this would still throw an `UnusableDeclaration`:
```purs
class A a where
  a :: String
```
While this won't throw anything:
```purs
class A @a where
  a :: String
```
Until it's skipped, in which `AmbiguousTypeVariables` is thrown:
```purs
x = a @_
```

PureFunctor:
The most recent changeset makes use of #4376. This brings the implementation closer to what is used by GHC, as only the `ForAll` type is involved as opposed to an extraneous `Specified` type.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -463,6 +544,59 @@ infer' (TypedValue checkType val ty) = do
...
+        --
+        -- Given:
+        --
+        -- forall (j :: Type) (k :: Type) (@t :: k -> Type)
+        --
+        -- Then:
+        --
+        -- 1. freeInVisKnd: { k }
+        -- 2. bfrVisY: [ k ]
+        -- 3. bfrVisN: [ j ]
```
PureFunctor, line 578: This doesn't take into account fringe cases such as:
```hs
forall (a :: Type) (b :: a) (@c :: Proxy b -> Type)
```
Where `a` also has to be instantiated alongside `b`. I doubt the compiler is ever able to synthesize such a type signature in most regular usage.

If there ever is a need for this, one way to go about it is to build a directed graph to figure out which variables need to be instantiated, starting from the kind of the visible quantifier. For example:
```hs
forall (a :: Type) (e :: Type) (b :: a) (c :: a -> Proxy b) (@d :: Proxy c)
```
Yields something along the lines of:
```hs
[ (d, [c]), (c, [a, b]), (b, [a]), (a, []) (e, []) ]
```
All that's left is to create the unification variables and replace as necessary.

i-am-the-slime: Maybe this should be inlined into the code rather than this issue, then?

i-am-the-slime:
I suppose this PR can be closed, right?

PR #4240, opened 2022-02-22
Disable constraints in FFI

### LINKED ISSUE #3182: Disable constraint syntax in `foreign import` declarations

JordanMartinez:
**Description of the change**

Fixes #3182. Part of `v0.15.x` changes

---

**Checklist:**


CHANGELOG.d/breaking_disable-constraints-in-ffi.md
```diff
@@ -0,0 +1,23 @@
...
+  In `v0.14.x`, a warning was emitted if these were used. Now it will fail
+  to compile. Rather, one should write something like the following
+  where the members of the type class are passed explicitly to
+  the FFI function as arguments:
+
+  ```purescript
+  foo :: forall a. Show a => a -> String
+  foo val = fooImpl show val
+
+  foreign import fooImpl :: forall a. (a -> String) -> a -> String
```
rhendric, line 22: Do we have a test that exercises this pattern?

JordanMartinez: Could you clarify what you mean by this?

rhendric: Literally just a test that verifies that calling such a `fooImpl show val` works as we expect it to, and not, say, that `show` gets passed to `fooImpl` without being applied to the appropriate dictionary.

It's pretty unlikely that this would regress without a lot of other things breaking, but if it's going in our changelog as the recommended thing to do, I'd prefer there to be even just one test that covers it.

JordanMartinez: Ah, that makes sense. I've just pushed a commit that I believe does that.

tests/purs/passing/FFIConstraintWorkaround.purs
```diff
@@ -0,0 +1,19 @@
+module FFIConstraintWorkaround where
```
rhendric, line 1: This needs to be `module Main`, and it's a little alarming that our test fixture doesn't catch that (and, I think, just silently re-runs the previously compiled test).

Anyway, after you fix that, I think you'll find that this test doesn't pass because the output doesn't end in ‘Done’. Might be better to explicitly compare the result of `showFFI` to the expected strings, since in the counterfactual regression I was worried about where `show` is passed before being applied to a dictionary, what I think would happen with this test as written is that the value would be passed as the first JavaScript argument to `show`, which means it would be considered the dictionary, and `show` would silently accept that up until it receives a second argument which never arrives. So `showFFI` would return a JavaScript function that PureScript thinks is a `String` and happily passes to `log`; `log` probably prints `[Function (anonymous)]` to stdout, and the test continues on its way none the wiser.

JordanMartinez: I've addressed this comment in latest commits.

rhendric (approved):

tests/purs/passing/FFIConstraintWorkaround.purs
JordanMartinez: #4243 should fix the `Main` module issue.

PR #4241, opened 2022-02-22
Drop support for ad-hoc `case ... of` expressions

### LINKED ISSUE #4012: Proposal: Remove ad-hoc handling of `case` formatting in the parser
## Summary

The parser has a couple of ad-hoc cases in the parser that should not be necessary.

https://github.com/purescript/purescript/blob/e56d28b211864075c7232e56350539806efbb01f/lib/purescript-cst/src/Language/PureScript/CST/Parser.y#L404-L411

## Motivation

I originally added these cases to keep some core modules compiling that use extremely odd formatting for single-branch cases. Normally these would run afoul the offside rule, but the old parser accepted them due to various idiosyncrasies. 

https://github.com/purescript/purescript-transformers/blob/42175954755a0fa0fe688579309032a948d919fb/src/Control/Monad/Writer/Trans.purs#L82-L86

## Proposal

We should remove these cases in 0.15 and update the core libraries. We can warn in a followup 0.14 release.

JordanMartinez:
**Description of the change**

Fixes #4012. I think the examples in the changelog entry are correct, but correct me if they're not.

This is for `v0.15.x`

---

**Checklist:**


natefaubion:
The examples are not correct.

```
-- This is ok
case wat of Foo a -> a`

-- This is ok
case wat of
  Foo a -> a

-- This is ok
case wat of Foo a ->
              a

-- This is not ok
case wat of Foo a ->
  a
```

There is an offside rule case statements, where the case branches must be indented past the pattern (just like with `let`). In the last example, the `a` is not indented past the `Foo a` column.

It's the same issue as `let`, we require the expression to be indented past the binder.

```
-- This is ok
let Foo a = b

-- This is ok
let Foo a =
      b

-- This is ok
let
  Foo a =
    b

-- This is not ok
let Foo a =
  b
```

CHANGELOG.d/breaking_remove-two-ad-hoc-case-statements.md
```diff
@@ -0,0 +1,20 @@
+* Remove ad-hoc same-line single-branch case statements
```
natefaubion, line 1: It's not single line case statements. Single line case statements are ok. It's a particular form of non-single-line case statements when there's only one branch.

natefaubion: Also, a nit, but they are case expressions, not statements.

JordanMartinez: I've reworked the entry. How about now?

natefaubion:
I'm personally OK with this, but the original ticket asked for a warning before removing. I think we should only remove it without a warning if the other contributors agree to it.

JordanMartinez:
I'll update this to add a warning.

lib/purescript-cst/src/Language/PureScript/CST/Errors.hs
```diff
@@ -63,6 +63,7 @@ data ParserWarningType
   | WarnDeprecatedConstraintInForeignImportSyntax
   | WarnDeprecatedKindImportSyntax
   | WarnDeprecatedKindExportSyntax
+  | WarnDeprecatedCaseOfOffsideSyntax
```
JordanMartinez, line 66: I had no idea what to call this error, so this is what I put for now. Let me know if you have a better idea for it.

lib/purescript-cst/src/Language/PureScript/CST/Parser.y
```diff
@@ -406,9 +406,9 @@ expr5 :: { Expr () }
   -- at any level, but this is ambiguous. We allow it in the case of a singleton
   -- case, since this is used in the wild.
   | 'case' sep(expr, ',') 'of' '\{' sep(binder1, ',') '->' '\}' exprWhere
-      { ExprCase () (CaseOf $1 $2 $3 (pure ($5, Unconditional $6 $8))) }
+      {% addWarning (let (a,b) = whereRange $8 in [a, b]) WarnDeprecatedCaseOfOffsideSyntax *> pure (ExprCase () (CaseOf $1 $2 $3 (pure ($5, Unconditional $6 $8)))) }
   | 'case' sep(expr, ',') 'of' '\{' sep(binder1, ',') '\}' guardedCase
-      { ExprCase () (CaseOf $1 $2 $3 (pure ($5, $7))) }
+      {% addWarning (let (a,b) = guardedRange $7 in [a, b]) WarnDeprecatedCaseOfOffsideSyntax *> pure (ExprCase () (CaseOf $1 $2 $3 (pure ($5, $7)))) }
```
JordanMartinez, line 411: Do we need both `a` and `b` source tokens here? Or just `a`?

JordanMartinez:
[CI error](https://github.com/purescript/purescript/runs/5347785663?check_suite_focus=true#step:7:4884) seems related to a `purs bundle` issue.

lib/purescript-cst/src/Language/PureScript/CST/Errors.hs
```diff
@@ -193,3 +194,5 @@ prettyPrintWarningMessage ParserErrorInfo {..} = case errType of
     "Kind imports are deprecated and will be removed in a future release. Omit the 'kind' keyword instead."
   WarnDeprecatedKindExportSyntax ->
     "Kind exports are deprecated and will be removed in a future release. Omit the 'kind' keyword instead."
+  WarnDeprecatedCaseOfOffsideSyntax ->
+    "An expression that is not indented past its `case ... of` branch's binder is deprecated and will be removed in a future release. Indent the expression beyond its binder, similar to how let bindings work."
```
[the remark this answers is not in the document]
natefaubion: I don't think we need to reference let. Maybe something like:
```
Dedented expressions in case branches are deprecated and will be removed in a future release. Indent the branch's expression past it's binder instead.
```
It's a tough concept to word succinctly.

src/Language/PureScript/Errors.hs
```diff
@@ -522,6 +522,7 @@ errorSuggestion err =
           CST.WarnDeprecatedForeignKindSyntax -> suggest $ "data " <> CST.printTokens (drop 3 toks)
           CST.WarnDeprecatedKindImportSyntax -> suggest $ CST.printTokens $ drop 1 toks
           CST.WarnDeprecatedKindExportSyntax -> suggest $ CST.printTokens $ drop 1 toks
+          CST.WarnDeprecatedCaseOfOffsideSyntax -> suggest $ CST.printTokens toks
```
natefaubion, line 525: I don't think this is correct. This is printing a replacement suggestion. This will end up suggesting a replacement of the first and last tokens of the expression.

JordanMartinez: Ah, ok. I just dropped this by returning `Nothing`.

lib/purescript-cst/src/Language/PureScript/CST/Errors.hs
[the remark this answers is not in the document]
JordanMartinez: Warning has been updated to use the above rendering. Thanks!

JordanMartinez:
Just wanted to note that in cc19ce8b8c4338ab7d5035827a8fdfa4254f9c4e, I reverted the changes made in #4249

natefaubion (approved):

PR #4248, opened 2022-02-27
Update CI to use `windows-2019`

### LINKED ISSUE #4242: Change Windows CI runner from windows-2016 to windows-2019
## Description

`windows-2016` is deprecated and [won't be supported after March 15](https://github.com/actions/virtual-environments/issues/4312). We should change our runner to use `windows-2019`, probably after releasing v0.14.6 on the off chance that it breaks a dependency.

JordanMartinez:
**Description of the change**

Fixes #4242

I've listed the changelog entry as 'internal' as I'm not quite sure how this is a "breaking change" to end-users. CI does build our binaries, so perhaps that can affect things? 

---

**Checklist:**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false # do not cancel builds for other OSes if one fails
       matrix:
         # If upgrading Ubuntu, also upgrade it in the lint job below
-        os: [ "ubuntu-18.04", "macOS-10.15", "windows-2016" ]
+        os: [ "ubuntu-18.04", "macOS-10.15", "windows-2019" ]
 
     runs-on: "${{ matrix.os }}"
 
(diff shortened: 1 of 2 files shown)


rhendric:
Yeah, it's not a breaking change from a you-can-always-build-from-source perspective, but we seem to have a set of users who can't/won't do that, and for them, changing how the binaries get built can be breaking—at least on Linux, it motivated the whole conversation in #4107. I don't know if bumping the Windows build environment can have the same impact, but I tagged the issue as breaking thinking maybe better safe than sorry.

OTOH, @hdgarrood [agreed](https://github.com/purescript/purescript/issues/4107#issuecomment-856960015) with you on not marking this sort of thing as breaking in the changelog. I'm not convinced that policy best serves those aforementioned users, but given that this is a Windows-only change it's probably a moot point either way?

rhendric (approved):

JordanMartinez:
Perhaps it should be mentioned in the Discourse announcement for Windows users?

Issue #4230, opened 2022-01-29
Proposal: Only support ES modules in PureScript 0.15 (disallow CommonJS)


thomashoneyman:
## Summary

In #3791 we are moving to support ES modules with the PureScript compiler. We'll accept ES modules via the FFI, and the compiler output will be ES modules. But what are we to do with all existing code that uses CommonJS modules?

My proposal is to only support ES modules in the compiler. Specifically, we would only support ES modules in compiler output and any existing FFI written with CommonJS modules will have to be rewritten (there [are tools that can help do this automatically](https://github.com/working-group-purescript-es/challenges/issues/6)).

This proposal has been [cross-posted to the PureScript Discourse](https://discourse.purescript.org/t/rfc-only-support-es-modules-in-purescript-0-15/2877?u=thomashoneyman) and more discussion can be found there.

## Motivation

We already only emit ES modules in #3791, and I believe this is uncontroversial. However, while #3791 [emits warnings for CommonJS in foreign modules](https://github.com/purescript/purescript/pull/3791/commits/c56d3f5b7a72f07b20263505b02fc4b30accfd67), it also makes some attempts to continue supporting CommonJS via the FFI.

That said, there are issues with trying to support CommonJS foreign modules. Some include:

1. If we allow CommonJS in the FFI, then we cannot follow through on #4226 because unbundled development for the browser is only possible using ES modules. That proposal describes several reasons why we really ought to drop our buggy bundler, especially since we can get even better bundle sizes by skipping our bundler altogether and using purity annotations and a bundler like `esbuild`.
2. As far as I can tell, there isn't a good story for how to support both module formats in `purs bundle`. I'm not sure we even _can_ mix and match module formats, and even if we can, I'm not aware of anyone who is interested in sinking any more time into the bundler especially when there are perfectly good industrial solutions for bundling.

There's the follow-on downside that if we can't drop the PureScript bundler (#4226), then we can't drop `language-javascript` either (#4227), and that means continuing to de facto maintain a JavaScript parser that already can't reliably parse ES modules (see [1](https://github.com/erikd/language-javascript/issues/127), [2](https://github.com/erikd/language-javascript/issues/119), [3](https://github.com/erikd/language-javascript/issues/131), and so on).

For these reasons, as well as the reasons described in #4226 and #4227, I think that it's just not worth attempting to support CJS at all with the PureScript compiler once we support ES modules.

## Proposal

I am proposing that we only deal with ES modules in the compiler:

[...]

Issue #4227, opened 2022-01-16
Proposal: Remove `language-javascript`


thomashoneyman:
## Summary

In #4226 I proposed that we remove the PureScript bundler. If we remove the bundler, then we only use `language-javascript` to verify that FFI declarations have a corresponding export in a `.js` file of the same name as the PureScript module. This library has been a notable maintenance burden on us over the past few years and I would like to move away from it.

Therefore, I propose that we drop the `language-javascript` dependency altogether if & when we remove the PureScript bundler.

To do this, we would need to do one of:

* No longer verify FFI declarations have a corresponding `.js` export, instead relying on the immediate errors that would happen on application load if there was no corresponding export
* Write our own minimal JavaScript lexer / parser that only looks for top-level declarations and exports for the purposes of validating FFI, but which doesn't attempt to parse all of the JavaScript.

## Motivation

We've had trouble with `language-javascript` over the past few years. We currently [pin an old version to avoid a bug parsing the `async` keyword](https://github.com/purescript/purescript/blob/18e3b14c556c5bd58e391531e4f4e511f834c3ff/stack.yaml#L10-L13), and we've supplied the [majority of pull requests to the library since 2019](https://github.com/erikd/language-javascript/pulls?q=is%3Apr+is%3Aclosed).

While a good library, it has plenty of bugs in it and it's become a maintenance burden for the compiler team. Removing `language-javascript` allows us to drop this burden. 

## Proposal

I am proposing that we remove `language-javascript` if & when we remove `purs bundle`.

Without `purs bundle`, the only reason to have a JavaScript parser is to ensure that FFI declarations have a corresponding implementation. This leaves us with a decision:

1. No longer verify this, leaving it to be a runtime or bundler error
2. Write a "simple" JavaScript lexer / parser that only cares about exports

In the first case there is really nothing for us to do besides removing the dependency and corresponding code.

In the second case, we would need to write our own code. This code. This code would need to be able to lex JavaScript literals like strings, brackets, export declarations, identifiers & keywords, and comments, so as to parse out exports. Everything else could be skipped over.

I think we should do (2) and maintain our own simple parser. I think this parser should only support "common sense" JavaScript rather than try to cover all edge cases. FFI files are not "do what you want" JavaScript, after all: they're a bridge between the PureScript and JavaScript worlds, and so I think we can place some restrictions on the JavaScript to make it easier for us to parse the export declarations.

[...]

PR #4250, opened 2022-02-28
Fail if CJS is detected in FFI

sigma-andex:
**Fail if CJS is detected in FFI, do not fail if FFI js parsing fails**

This is the first change according to the proposal to [remove language-javascript](https://github.com/purescript/purescript/issues/4227) and the proposal to [only support es modules](https://github.com/purescript/purescript/issues/4230).  See also the discussion in the [previous PR within the es modules working group](https://github.com/working-group-purescript-es/purescript/pull/8).
This change addresses the following:
* building fails if CJS is detected in the FFI
* building does not emit an error or warning if the `language-javascript` parsing fails alltogether.
  This is because `language-javascript` does not support certain valid constructs like e.g. dynamic imports or jsx. 

**Note:** One failing bundle test has been removed as `purs bundle` is getting removed anyways.

---

**Checklist:**


gabejohnson:
@sigma-andex could you add a link to discussions where this behavior is agreed upon. It would be an enormous breaking change and I don't see mention of it in the [linked issue](https://github.com/purescript/purescript/issues/4227).

sigma-andex:
> @sigma-andex could you add a link to discussions where this behavior is agreed upon. It would be an enormous breaking change and I don't see mention of it in the [linked issue](https://github.com/purescript/purescript/issues/4227).

Updated the description with the link that @thomashoneyman posted and also the link to the discussion in the [previous pr](https://github.com/working-group-purescript-es/purescript/pull/8).

JordanMartinez:
> If people have questions about decisions the working group made we should point them to the proposals and RFCs, as otherwise it could sound like a small group made these decisions without the community participating.

@thomashoneyman Yeah, I wasn't trying to say a small group of people had discussed it and the decision final. Merely that that is where a lot of discussion had taken place. I was expecting to continue any such discussion in this PR if we didn't have any linked issues. But thanks for stating this explicitly as I don't want that perception to be communicated.

CHANGELOG.d/breaking_fail-on-cjs-detected.md
```diff
@@ -0,0 +1,7 @@
+* Fail if CJS is detected in FFI
+
+  Previously, one could write CommonJS in FFI declarations. With the move to ES modules, this leads to a failing compile now.
+
+* Do not emit an error if language-javascript parsing fails
+
+  Previously, an error was emitted when language-javascript parsing failed. With this change, no error or warning is emitted if the js FFI parsing fails.
```
JordanMartinez, line 7: Two comments here. First (and just for future PRs), if this is supposed to be two separate entries, then it should be two separate files.

Second, I think this is actually one entry. Perhaps this rendering summarizes things better? I'd like to hear other core team members' feedback on this point.
```markdown
* If FFI parsing succeeds & CommonJS is detected, fail; otherwise, do not error or warn

  Previously, the compiler would emit an error if it failed to parse the FFI JavaScript file.
  Since the underlying JavaScript parser (i.e. `language-javascript`) fails to parse even
  valid JavaScript files, we cannot consider every failed parse to mean invalid JS files.
  Fixing the parser would require a lot of effort, so we are planning to remove it instead
  in `v0.16.x`.
  
  If the parse succeeds and a CommonJS module is detected, a compiler error is now emitted.
  If the parse fails, we no longer emit a compiler error. While we could emit a warning,
  such a warning will quickly become annoying for FFI files that trigger the buggy paths
  of `language-javascript`. Moreover, we presume that all will be migrating their code to
  ES modules now that CommonJS is being deprecated in the larger JavaScript ecosystem.
```

> CommonJS is being deprecated in the larger JavaScript ecosystem

Regarding this point in the above entry, @sigma-andex could you link to the post that mentioned this? I can't recall where it is.

src/Language/PureScript/Make/Actions.hs
```diff
@@ -345,55 +327,61 @@ data ForeignModuleType = ESModule | CJSModule deriving (Show)
...
   mname = CF.moduleName m
   modSS = CF.moduleSourceSpan m
 
-  errorParsingModule :: Bundle.ErrorMessage -> Make a
-  errorParsingModule = throwError . errorMessage' modSS . ErrorParsingFFIModule path . Just
+  checkFFI :: JS.JSAST -> Make (ForeignModuleType, S.Set Ident)
+  checkFFI js = do 
+    (foreignModuleType, foreignIdentsStrs) <-
+        case (,) <$> getForeignModuleExports js <*> getForeignModuleImports js of
+          Left reason -> throwError $ errorParsingModule reason
```
JordanMartinez, line 347: Looks like this [didn't used to have a `throwError` in front of it](https://github.com/purescript/purescript/pull/4250/files#diff-889762f9d56ac6e7b5dcb3db0b2fca495526bde7fd588e07fdcacaa7d4aca1feL355). The `throwError` was moved into `errorParsingModule`. Would you mind undoing that? I think this change isn't needed for this PR, but correct me if I'm wrong.

JordanMartinez (commented):
I think this PR is ready to be approved, but due to the comment from Gabe, I'm going to hold off until getting a response.

src/Language/PureScript/Make/Actions.hs
sigma-andex: Gotcha!

sigma-andex: Hmm so the reason is that I'm using `errorParsingModule` also in line [336](https://github.com/purescript/purescript/pull/4250/files#diff-889762f9d56ac6e7b5dcb3db0b2fca495526bde7fd588e07fdcacaa7d4aca1feR336). I could keep it as I was but then I would need to handle that line differently, don't know if that makes much sense

CHANGELOG.d/breaking_fail-on-cjs-detected.md
sigma-andex: I think you meant [this discussion](https://github.com/sindresorhus/meta/discussions/15). And there are many libraries that follow this, like e.g. [`node-fetch`](https://github.com/node-fetch/node-fetch#commonjs).

sigma-andex: Should I add this to the changelog?

src/Language/PureScript/Make/Actions.hs
JordanMartinez: Ah, ok. Nevermind then.

src/Language/PureScript/Make/Actions.hs
```diff
@@ -345,55 +327,61 @@ data ForeignModuleType = ESModule | CJSModule deriving (Show)
...
-  unless (null missingFFI) $
-    throwError . errorMessage' modSS . MissingFFIImplementations mname $
-      S.toList missingFFI
-
-  pure (foreignModuleType, foreignIdents)
+
+  let
+    parseResult :: Either MultipleErrors JS.JSAST
+    parseResult = first (errorParsingModule . Bundle.UnableToParseModule) $ JS.parseModule jsStr path
+  either (pure . Left) (fmap Right . checkFFI) parseResult
```
rhendric, line 337:
```suggestion
  traverse checkFFI parseResult
```

rhendric (approved):
LGTM (with one trivial suggestion)

src/Language/PureScript/Make/Actions.hs
sigma-andex: Nice!

src/Language/PureScript/Errors.hs
```diff
@@ -713,7 +713,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
     renderSimpleErrorMessage (DeprecatedFFICommonJSModule mn path) =
       paras [ line $ "A CommonJS foreign module implementation was provided for module " <> markCode (runModuleName mn) <> ": "
             , indent . lineS $ path
-            , line "CommonJS foreign modules are deprecated and won't be supported in the future."
+            , line "CommonJS foreign modules are no longer supported."
```
i-am-the-slime, line 716:
```suggestion
            , line "CommonJS foreign modules (imports using `require`) are no longer supported, please use `import` instead."
```

i-am-the-slime: Or something like this could be beneficial to people not familiar with the term CommonJS and seems more immediately actionable to me.

sigma-andex: I think this is a valid and useful change. However changing the error message always requires updating the tests. So before I do this, I would like to ask here if you all agree on this or have more proposals on how to improve this error message. Thx :-)

JordanMartinez: > Or something like this could be beneficial to people not familiar with the term CommonJS and seems more immediately actionable to me.

Isn't that what the [documentation page for DeprecatedFFICommonJsModule](https://github.com/purescript/documentation/blob/master/errors/DeprecatedFFICommonJSModule.md) is for?

I don't think this change needs to be done.

CHANGELOG.d/breaking_fail-on-cjs-detected.md
JordanMartinez: I don't think that was the post I remember you mentioning elsewhere. But since I can't find that post, yeah, let's link to that discussion.

src/Language/PureScript/Errors.hs
natefaubion: I think it's worth mentioning what should be used instead, but `imports using require`  isn't the whole story (what about exports using the exports/module object?). I think we should say:

> CommonJS foreign modules are no longer supported. Use native JavaScript/ECMAScript module syntax instead.

JordanMartinez (approved):

gabejohnson:
@sigma-andex thanks for adding the links!

> Just to clarify, what is "it" in this context? Much of the conversation we've had about the ES modules work has been done via weekly Discord calls.

@JordanMartinez "it" refers to the behavior described in the title "Fail if CJS is detected in FFI". I'm not objecting to the change. Just wanted to see where that decision had been made.

src/Language/PureScript/Errors.hs
sigma-andex: Ok I updated it to @natefaubion 's proposal.

JordanMartinez:
@gabejohnson Ah, thanks for clarifying!

PR #4251, opened 2022-02-28
Make first alpha release for v0.15.0

JordanMartinez:
**Description of the change**

This follows the approach done in #3976. In #3933, we dropped `package.yaml` file. Running `stack build` and then calling `purs --version` on the produced binary prints this output:
```
$ ./.stack-work/install/x86_64-linux-tinfo6/d400b0043ab29e05088ff3d6399c1c588ef336df281e2e819bcac83142913d36/8.10.7/bin/purs --version
0.15.0-rc1 [development build; commit: f01ba8f4d402dde0dad9feae4c48949318bf24be DIRTY]
```

I'm not sure when we want to make the first release candidate (as there may be additional work to do before such a release), but once this PR is merged, we'd make a new GH release by creating a tag for `v0.15.0-rc1` and checking the 'Is prelease' checkbox.

---

**Checklist:**

### DIFF
--- a/app/Version.hs
+++ b/app/Version.hs
@@ -17,7 +17,7 @@ import qualified Development.GitRev as GitRev
 -- prerelease identifier here (if any). When releasing a proper version, simply
 -- set this to an empty string.
 prerelease :: String
-prerelease = "-rc1"
+prerelease = "-alpha1"
 
 versionString :: String
 versionString = showVersion Paths.version ++ prerelease ++ extra
--- a/app/Version.hs
+++ b/app/Version.hs
@@ -17,7 +17,7 @@ import qualified Development.GitRev as GitRev
 -- prerelease identifier here (if any). When releasing a proper version, simply
 -- set this to an empty string.
 prerelease :: String
-prerelease = "-alpha1"
+prerelease = "-alpha.1"
 
 versionString :: String
 versionString = showVersion Paths.version ++ prerelease ++ extra
--- a/app/Version.hs
+++ b/app/Version.hs
@@ -17,7 +17,7 @@ import qualified Development.GitRev as GitRev
 -- prerelease identifier here (if any). When releasing a proper version, simply
 -- set this to an empty string.
 prerelease :: String
-prerelease = ""
+prerelease = "-rc1"
 
 versionString :: String
 versionString = showVersion Paths.version ++ prerelease ++ extra
--- a/app/Version.hs
+++ b/app/Version.hs
@@ -17,7 +17,7 @@ import qualified Development.GitRev as GitRev
 -- prerelease identifier here (if any). When releasing a proper version, simply
 -- set this to an empty string.
 prerelease :: String
-prerelease = "-alpha.1"
+prerelease = "-alpha-01"
 
 versionString :: String
 versionString = showVersion Paths.version ++ prerelease ++ extra
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -2,7 +2,7 @@ cabal-version:  2.4
 
 name:           purescript
 -- note: When updating the prerelease identifier, update it in app/Version.hs too!
-version:        0.15.0-alpha1
+version:        0.15.0-alpha.1
 synopsis:       PureScript Programming Language Compiler
 description:    A small strongly, statically typed programming language with expressive types, inspired by Haskell and compiling to JavaScript.
 category:       Language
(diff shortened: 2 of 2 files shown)


purescript.cabal
```diff
@@ -2,7 +2,7 @@ cabal-version:  2.4
 
 name:           purescript
 -- note: When updating the prerelease identifier, update it in app/Version.hs too!
-version:        0.15.0-rc1
+version:        0.15.0
```
JordanMartinez, line 5: Without this change, I get the following error:
```
$ stack build
Cabal file warning in/home/jordan/Programming/Projects/purescript/purescript.cabal@5:27: version with tags

Comment (rhendric) [reply purescript.cabal:5]: Isn't it just a warning? AFAIK we've had no trouble with this [in the past](https://travis-ci.org/github/purescript/purescript/jobs/751718020#L582).

Comment (JordanMartinez) [reply purescript.cabal]: Ah, you're right. When I saw that `stack build` didn't produce any output, I thought it meant the build failed. But it didn't produce output because I had run `stack build` previously and there was nothing else to do.

I've reverted the commit.

File: app/Version.hs [hunk 2]
```diff
@@ -17,7 +17,7 @@ import qualified Development.GitRev as GitRev
 -- prerelease identifier here (if any). When releasing a proper version, simply
 -- set this to an empty string.
```
JordanMartinez, line 18: Is this comment still true then?

JordanMartinez:
Once #4241 is merged, I think this release should be made.

rhendric:
Really? Seems like we have some more breaking changes that have just been recently proposed. Do we want to push them for another major release?

Or do we use the term ‘release candidate’ here to mean something other than ‘this build is a candidate for being the final release, if no issues arise during testing’?

JordanMartinez:
Ah, sorry! Let me clarify.

Calling this a 'release candidate' is just using the wrong term. It's more like a 'beta release'. Making a "not ready for production use yet" release means I can start testing ES modules migration PRs that are being submitted to the core repos. Since most repos already use `setup-purescript`, I'm updating that tool to also work on pre-release versions (see https://github.com/purescript-contrib/setup-purescript/pull/27). So, making a release candidate just unblocks the initial part of ecosystem migration. Sometimes, updating the libraries will reveal bugs in the code or highlight issues we need to address before we make the actual release.

In other words, I fully expect to merge more breaking changes PRs before we make a final `v0.15.0` release (e.g. we still need to drop `purs bundle`). I'm merely repeating what we did for `v0.14.0`, and those pre-releases had suffixes like `rc1`.

thomashoneyman:
We need an ES-modules-compliant version of the compiler to be available so we can start making 0.15 changes across the purescript, purescript-contrib, purescript-node, and purescript-web organizations.

However, they certainly don't need to be named "release candidates" and carry the `-rc` suffix, as that carries a specific meaning that implies we're closer to a release than we are. I'd be perfectly happy with, for example, `0.15.0-alpha`!

JordanMartinez:
Let's call it "alpha". I think using more accurate terminology prevents cases of confusion like this :smile:.

rhendric (approved):

app/Version.hs
```diff
@@ -17,7 +17,7 @@ import qualified Development.GitRev as GitRev
...
 prerelease :: String
-prerelease = "-rc1"
+prerelease = "-alpha1"
```
thomashoneyman, line 20: I don't know what convention we typically follow, but I'm used to seeing `-alpha.1` and `-rc.1` (with the dot)

JordanMartinez: :man_shrugging: If we include the `.`, then `Version.parseVersion` (assuming I'm reading it right) will [list the two values separately](https://github.com/hdgarrood/purescript-versions/blob/v6.0.0/src/Data/Version.purs#L146). But the general question is, "Does it matter?"

rhendric: Practically, I think it only matters if we have more than 9 alpha releases, in which case per (my reading of) the SemVer spec, `-alpha10` will sort before `-alpha1`, but `-alpha.10` would sort after `-alpha.1`.

rhendric: (Our history on this seems [less than fully consistent](https://github.com/purescript/purescript/tags?after=v0.14.0).)

JordanMartinez: Ok, dot it is!

JordanMartinez: :laughing: https://github.com/purescript/purescript/runs/5382701148?check_suite_focus=true#step:7:20

So... `alpha01`?

thomashoneyman: Oops, guess you can't do a dot in .cabal. In that case I'm happy with `alpha-1`.

PR #4254, opened 2022-03-03
Trigger CI and fix failing tests due to changes in `es-modules` branches

JordanMartinez:
**Description of the change**

I'm using this PR to explore the CI failures in #4207, which are likely caused by the issue summarized in https://github.com/working-group-purescript-es/purescript/issues/16

---

**Checklist:**

### DIFF
--- a/tests/TestUtils.hs
+++ b/tests/TestUtils.hs
@@ -37,6 +37,8 @@ import System.Process hiding (cwd)
 import qualified System.FilePath.Glob as Glob
 import System.IO
 import Test.Hspec
+import Data.Foldable (traverse_)
+import qualified Data.Text.IO as T
 
 -- |
 -- Fetches code necessary to run the tests with. The resulting support code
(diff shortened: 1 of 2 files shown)


JordanMartinez:
The last commit printed the contents of `Data.HeytingAlgebra.js` to see whether it's using CommonJS modules or not (just as a sanity check). It's using ES6, so at least caching isn't the issue.

JordanMartinez:
Looks like https://github.com/purescript/purescript/pull/4254/commits/82467e71d4ef80a9c6e14e828e72b7a2c1dce26c is building fine again. In that commit, I updated `prelude` to point to the state of `es-modules` branch before I added the commits in https://github.com/purescript/purescript-prelude/pull/284. That branch is the `fix-es-modules` branch.

The latest commit updates `prelude` to point to `fix-es-modules--change-unit`, which is the same as `fix-es-modules` but with another commit that changes unit's FFI from `{}` to `undefined`. If that doesn't cause CI to fail, then I'm wondering if these do:
```diff
- exports.var functionName = function() { ... };
+ export const functionName = function() { ... };
```
Perhaps we need to use this version to fix things?
```
export function functionName() { ... };
```

JordanMartinez:
The Ubuntu build builds fine even though `unit` is changed from `{}` to `undefined`. So that leaves the `export const functionName` question.

JordanMartinez:
Looks like using `export const functionName = function() { .. }` is causing the CI errors. @sigma-andex Thoughts on this?

natefaubion:
There's nothing wrong with that unless it's just a style lint thing, which can be disabled.

JordanMartinez:
> There's nothing wrong with that unless it's just a style lint thing, which can be disabled.

Could you clarify this? I'm not sure what you mean. If you mean disabling an eslint thing, I don't think that's the issue, but I'm guessing that's not what you mean. AFAICT, the CI errors are caused by `purs` thinking that the exports of `Data.HeytingAlgebra`'s FFI file, which uses `export const` rather than `export var`, are invalid.

natefaubion:
> the exports of Data.HeytingAlgebra's FFI file, which uses export const rather than export var, are invalid.

Ignore my comment then. The issue is that this branch changed exports to const, but the interim ESM bundler which hasn't been removed yet doesn't understand it. My suggestion would to revert any such changes on those branches, and remove the bundler before making progress.

JordanMartinez:
> I think you came to that conclusion elsewhere laughing

:grin:

JordanMartinez:
Succeeded by #4255

PR #4255, opened 2022-03-03
Remove `purs bundle` and browser backend for `purs repl`

### LINKED ISSUE #4226: Proposal: Remove the PureScript bundler
## Summary

I am proposing that we remove the PureScript bundler once we support ES modules in the compiler (once #3791 merges and #3613 is closed). With ES modules in place, the bundler is no longer necessary to develop applications in the browser or on Node. Since the bundler is slow, buggy, and difficult to work on -- a high risk, low reward part of the compiler -- removing it reduces the maintenance burden on the already-thin compiler team.

Two weeks ago I opened an RFC on Discourse titled [RFC: The Future of the PureScript Bundler](https://discourse.purescript.org/t/rfc-the-future-of-the-purescript-bundler) in which the community has so far seemed entirely in favor of this proposal.

## Motivation

The PureScript compiler includes a bundler that performs dead code elimination and produces a single `.js` file from PureScript sources, which can then be used in the browser or on Node. The bundler exists for a few reasons:

1. To ensure users can run their PureScript code in the browser and on Node without requiring additional tools
2. To trim the size of the generated code by performing dead code elimination while bundling

Enter ES modules, which we have already agreed to support in the compiler. We have two paths forward for ES modules support as far as the bundler is concerned:

1. We fix the bundler so it works for ES modules -- which is trickier than it may at first appear
2. We remove the bundler altogether

Why has (2) even become an option at all, when it hasn't historically been an option? Because things have changed with ES modules:

1. ES modules are [usable in the browser without bundling](https://caniuse.com/es6-module). Therefore user interfaces can be developed in PureScript without the need for a bundler.
2. ES modules are usable in Node experimentally since v12, and [fully in Node since v14](https://nodejs.org/docs/latest-v14.x/api/esm.html) -- which will be the [oldest supported version by the time PureScript 0.15 comes out](https://nodejs.org/en/about/releases) -- so long as you use the file extension `.mjs` or set `type: "module"` in a package.json file. For now, the [compiler output contains a `package.json` file containing `type: "module"`](https://github.com/purescript/purescript/pull/3791#issuecomment-803639863).
3. Bundlers like webpack and esbuild [are able to provide dead code elimination for ES modules comparable with `purs bundle` optimized with `zephyr`](https://github.com/working-group-purescript-es/purescript/issues/1#issuecomment-1002176554), and may even perform better if we make other compiler changes like [changing the representation of data types](https://github.com/purescript/purescript/issues/4020). These bundlers of course are also able to produce a single `.js` file for use in the browser or on Node, and so they are able to replace `purs bundle`.

Points (1) and (2) mean that PureScript users can run their code in the browser and on Node without requiring any additional tools, and point (3) means that users can get effective dead code elimination without using the PureScript bundler.

My understanding is that there is little appetite among the compiler team for sinking engineering resources into the bundler; this is part of why ES modules support has taken so long. @kl0tl, the author of the [ES modules PR to the compiler](https://github.com/purescript/purescript/pull/3791) and member of the core team, has indicated their preference for removing the bundler.

Unfortunately: users would no longer get effective dead code elimination without external tooling. Still, I don't think this is much of a problem, because:

1. Dead code elimination is not necessary during development (it's typically an optimization needed for production builds)
2. It is already a best practice to rely on external tools for dead code elimination (namely, [Zephyr](https://github.com/coot/zephyr)).
3. Every production project I'm aware of (other than Try PureScript) already uses a JS bundler like Webpack or esbuild, which will perform the dead code elimination on the JS on your behalf.
4. If you have any JS dependencies in your PureScript project, then you already _must_ use a JS bundler like Webpack or esbuild (or take your FFI dependencies via a CDN, which is exceedingly uncommon).

## Proposal

Therefore, my proposal is that we remove the bundler altogether as part of supporting ES modules. The `purs bundle` command would be completely removed and the compiler would only produce the `output` directory. Downstream tools like Spago could implement their own `spago bundle` commands that rely on an industry-standard JS bundler like `esbuild` or `webpack`.

There is an [issue on Spago tracking the `spago bundle` command](https://github.com/purescript/spago/issues/836).

## Examples

The [ES modules working group](https://github.com/working-group-purescript-es) is producing more examples of the ES modules compiler being usable with bundlers like `esbuild` or `snowpack` or `webpack`, and the compiler being usable for unbundled development, but [preliminary results show that JS bundlers are competitive with `purs bundle`](https://github.com/working-group-purescript-es/purescript/issues/1#issuecomment-1007771244).

### LINKED ISSUE #4252: Proposal: Remove the Browser REPL backend
## Summary

Remove support for the Browser REPL backend in 0.15.

## Motivation

The browser REPL backend currently relies on `purs bundle` to serve bundled code to the browser. `purs bundle` is slated to be removed in 0.15. Retaining support for the browser REPL entails reworking it's internals to serve unbundled code. Discussion with other contributors has lead me to believe this feature is essentially unknown, lacks maintainer interest, and is a roadblock for shipping a highly requested feature.

## Proposal

Remove the browser REPL backend code, and direct users to Try PureScript, which offers a much better experience for trying code. It also supports the full current package set and even supports things like Halogen and React bindings.

JordanMartinez:
**Description of the change**

Fixes #4226. Also fixes #4252 because the browser backend for the repl depends on `purs bundle` working. This is a WIP.

I wasn't sure whether we want to keep `purs bundle` and just throw an error (current PR) or to remove the command completely.

---

**Checklist:**


natefaubion:
Purs bundle should fail with a link to some sort of documentation with alternatives.

JordanMartinez:
> Purs bundle should fail with a link to some sort of documentation with alternatives.

Ideally, that would be the migration guide's corresponding section: https://github.com/purescript/documentation/blob/master/migration-guides/0.15-Migration-Guide.md#how-can-i-bundle-my-library-or-application

Not sure if that header will change between now and the release.

JordanMartinez:
Lint error is caused by `src/System/IO/UTF8.hs:35: writeUTF8File`: [`writeUTF8File`](https://github.com/purescript/purescript/blob/master/src/System/IO/UTF8.hs#L34-L36)

Should we drop that? Comment it out? I'm not sure how often we'll be using that instead of [`writeUTF8FileT`](https://github.com/purescript/purescript/blob/master/src/System/IO/UTF8.hs#L26-L28)

rhendric:
> Should we drop that? Comment it out? I'm not sure how often we'll be using that instead of [`writeUTF8FileT`](https://github.com/purescript/purescript/blob/master/src/System/IO/UTF8.hs#L26-L28)

I vote drop it. Commenting it out is worse than adding it as a Weeder root; it would just be a breeding ground for bitrot. And we should encourage new code to use `Text` over `String` anyway.

rhendric:
FYI, there's some machinery [in TestUtils.hs](https://github.com/purescript/purescript/blob/24eeeb216e73921d8457a641eddba39a0b49b22e/tests/TestUtils.hs#L162-L174) for handling the RerunCompilerTests.txt file that I think you can rip out along with said file; I believe it was only ever used for bundler tests.

JordanMartinez:
Yeah, looks like it was only used with `bundle`. There are no other "RerunCompilerTests.txt" files anywhere.

JordanMartinez:
One comment raised when I talked about this in the recent ES modules Discord call is whether this PR also removes any "dead dependencies." For example, do we still need dependencies for web server-related functionality that enabled the REPL browser backend? I'd guess not, but I'm not sure which dependencies these would be.

CHANGELOG.d/breaking_remove-purs-bundle.md
```diff
@@ -0,0 +1 @@
+* Remove `purs bundle`; use another bundler like `esbuild`
```
thomashoneyman, line 1:
```suggestion
* Remove `purs bundle`

  Users of `purs bundle` should switch to a standalone bundler such as `esbuild`, `webpack` or `parcel`.
```

app/Command/Bundle.hs
```diff
@@ -3,141 +3,21 @@ module Command.Bundle (command) where
...
-
-  debug :: Parser Bool
-  debug = Opts.switch $
-       Opts.long "debug"
-    <> Opts.help "Whether to emit a JSON representation of all parsed modules to stderr."
+app :: IO ()
+app = do
+  hPutStrLn stderr $ unlines
+    [ "'purs bundle' was removed in the v0.15.0 release."
+    , "Use 'esbuild' or another bundler. For info on migrating to a bundler, see"
```
thomashoneyman, line 14: Are we taking a stance on suggesting `esbuild`? I'm fine with this, but I'd like to make sure maintainers in general are OK with suggesting a specific bundler here.

app/Command/REPL.hs
```diff
@@ -77,7 +56,7 @@ port :: Opts.Parser Int
 port = Opts.option Opts.auto $
      Opts.long "port"
   <> Opts.short 'p'
-  <> Opts.help "The web server port"
+  <> Opts.help "The browser REPL backend was REMOVED in v0.15.0. Use https://try.purescript.org instead."
```
thomashoneyman, line 59:
```suggestion
  <> Opts.help "The browser REPL backend was removed in v0.15.0. Use https://try.purescript.org instead."
```

app/Command/REPL.hs
```diff
@@ -132,144 +95,22 @@ data Backend = forall state. Backend
...
-  -- ^ An MVar which becomes full when the server should shut down
-  , browserIndexJS        :: TVar (Maybe String)
-  -- ^ A TVar holding the latest compiled JS
-  , browserBundleJS       :: TVar (Maybe String)
-  -- ^ A TVar holding the latest bundled JS
-  }
-
 browserBackend :: Int -> Backend
-browserBackend serverPort = Backend setup evaluate reload shutdown
+browserBackend _ = Backend setup evaluate reload shutdown
```
thomashoneyman, line 99: Can these all be `mempty`?

```suggestion
browserBackend _ = Backend setup mempty mempty mempty
```

tests/Main.hs
```diff
@@ -35,7 +35,7 @@ main = do
     describe "ide" TestIde.spec
     beforeAll TestUtils.setupSupportModules $ do
       describe "compiler" TestCompiler.spec
-      describe "bundle" TestBundle.spec
+      -- describe "bundle" TestBundle.spec
```
thomashoneyman, line 38: Should this comment be deleted altogether?

app/Command/Bundle.hs
sigma-andex: What about
```
[ "'purs bundle' was removed in the v0.15.0 release."
, "See https://www.github.com/purescript/documentation/migration-guides/0.15-Migration-Guide.md "
, "for more information and bundler alternatives."
]
```

sigma-andex:
> One comment raised when I talked about this in the recent ES modules Discord call is whether this PR also removes any "dead dependencies." For example, do we still need dependencies for web server-related functionality that enabled the REPL browser backend? I'd guess not, but I'm not sure which dependencies these would be.

My guess is that it would be these: https://github.com/JordanMartinez/purescript/blob/remove-purs-bundle/purescript.cabal#L339-L342

Also I think `index.html` and `index.js` in https://github.com/JordanMartinez/purescript/tree/remove-purs-bundle/app/static can be removed.

app/Command/REPL.hs
sigma-andex: Or something like this:
```purescript
data Backend = forall state. Backend
  { _backendSetup :: IO state
  -- ^ Initialize, and call the continuation when the backend is ready
  , _backendEval :: state -> String -> IO ()
  -- ^ Evaluate JavaScript code
  , _backendReload :: state -> IO ()
  -- ^ Reload the compiled code
  , _backendShutdown :: state -> IO ()
  -- ^ Shut down the backend
  } | DeprecatedBrowserBackend

browserBackend :: Backend
browserBackend = DeprecatedBrowserBackend
```

tests/Main.hs
JordanMartinez: Yeah, it should.

JordanMartinez:
> > One comment raised when I talked about this in the recent ES modules Discord call is whether this PR also removes any "dead dependencies." For example, do we still need dependencies for web server-related functionality that enabled the REPL browser backend? I'd guess not, but I'm not sure which dependencies these would be.
> 
> My guess is that it would be these: https://github.com/JordanMartinez/purescript/blob/remove-purs-bundle/purescript.cabal#L339-L342
> 
> Also I think `index.html` and `index.js` in https://github.com/JordanMartinez/purescript/tree/remove-purs-bundle/app/static can be removed.

Yup, spot on!

thomashoneyman (approved):

PR #4257, opened 2022-03-05
Restore names when generalizing unknowns

### LINKED ISSUE #4256: Track original names of quantified variables instantiated into unknowns
## Summary

The `instantiatePolyTypeWithUnknowns` routine in the type-checker turns universally quantified variables into unknowns. Keeping track of which variables are turned into which unknowns can help with better name generation once unknowns once generalization kicks in.

## Motivation

Names of quantified variables add helpful semantic information to consumers of some polymorphic function. However, it's particularly difficult to use them incrementally if this information gets erased. This small change should also help improve compiler warnings and errors; it'd be easier to change some type variable and code and see exactly which one is being pointed to by the error. 

## Proposal

Add another `Map` to `CheckState` which maps unknowns to their original names. During the generalization routine, check if an unknown has an entry in this map, if it does, use the unknown int as a suffix to the original name.

## Examples

For example:
```hs
const :: forall val dsc. val -> dsc -> val
const a _ = a

-- const0 :: forall dsc1. dsc1 -> Int
const0 = const 0
```

PureFunctor:
Closes #4256. This change makes it so the type checker tracks which quantified type variables are replaced with unknowns such that they can then be restored in the future.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_restore_names.md
+++ b/CHANGELOG.d/feature_restore_names.md
@@ -1,25 +1,25 @@
-# Restore names of quantified variables during generalization
+* Restore names of quantified variables during generalization
 
-This makes the compiler aware of the names of quantified variables
-instantiated into unification variables, such that when the latter
-is generalized, semantic information is restored. For example:
+  This makes the compiler aware of the names of quantified variables
+  instantiated into unification variables, such that when the latter
+  is generalized, semantic information is restored. For example:
 
-```hs
-addNumberSuffix :: forall a b c d. a -> b -> c -> d -> a
-addNumberSuffix a _ _ _ = a
+  ```purs
+  addNumberSuffix :: forall a b c d. a -> b -> c -> d -> a
+  addNumberSuffix a _ _ _ = a
+  
+  addNumberSuffix' = addNumberSuffix 0
+  ```
 
-addNumberSuffix' = addNumberSuffix 0
-```
+  Previously, inferring top-level declarations without type signatures
+  would use `t` suffixed with an integer for type variables.
 
-Previously, inferring top-level declarations without type signatures
-would use `t` suffixed with an integer for type variables.
+  ```purs
+  forall t6 t7 t8. t6 -> t7 -> t8 -> Int
+  ```
 
-```hs
-forall t6 t7 t8. t6 -> t7 -> t8 -> Int
-```
+  Now, the inferred type would refer back to their original names.
 
-Now, the inferred type would refer back to their original names.
-
-```hs
-forall b6 c7 d8. b6 -> c7 -> d8 -> Int
-```
+  ```purs
+  forall b6 c7 d8. b6 -> c7 -> d8 -> Int
+  ```
--- a/CHANGELOG.d/feature_restore_names.md
+++ b/CHANGELOG.d/feature_restore_names.md
@@ -1,0 +1,33 @@
+# Restore names of quantified variables during generalization
+
+This makes the compiler aware of the names of quantified variables
+instantiated into unification variables, such that when the latter
+is generalized, semantic information is restored. For example:
+
+```hs
+addNumberSuffix :: forall a b c d. a -> b -> c -> d -> a
+addNumberSuffix a _ _ _ = a
+
+addNumberSuffix' = addNumberSuffix 0
+```
+
+This now generalizes with the original names suffixed with a number
+instead of just being replaced by a `t`.
+
+```hs
+Warning found:
+in module Main
+at Main.purs:6:1 - 6:37 (line 6, column 1 - line 6, column 37)
+
+  No type declaration was provided for the top-level declaration of addNumberSuffix'.
+  It is good practice to provide type declarations as a form of documentation.
+  The inferred type of addNumberSuffix' was:
+
+    forall b6 c7 d8. b6 -> c7 -> d8 -> Int
+
+
+in value declaration addNumberSuffix'
+
+See https://github.com/purescript/documentation/blob/master/errors/MissingTypeDeclaration.md for more information,
+or to contribute content related to this warning.
+```
--- a/CHANGELOG.d/feature_restore_names.md
+++ b/CHANGELOG.d/feature_restore_names.md
@@ -11,23 +11,15 @@ addNumberSuffix a _ _ _ = a
 addNumberSuffix' = addNumberSuffix 0
 ```
 
-This now generalizes with the original names suffixed with a number
-instead of just being replaced by a `t`.
+Previously, inferring top-level declarations without type signatures
+would use `t` suffixed with an integer for type variables.
 
 ```hs
-Warning found:
-in module Main
-at Main.purs:6:1 - 6:37 (line 6, column 1 - line 6, column 37)
+forall t6 t7 t8. t6 -> t7 -> t8 -> Int
+```
 
-  No type declaration was provided for the top-level declaration of addNumberSuffix'.
-  It is good practice to provide type declarations as a form of documentation.
-  The inferred type of addNumberSuffix' was:
+Now, the inferred type would refer back to their original names.
 
-    forall b6 c7 d8. b6 -> c7 -> d8 -> Int
-
-
-in value declaration addNumberSuffix'
-
-See https://github.com/purescript/documentation/blob/master/errors/MissingTypeDeclaration.md for more information,
-or to contribute content related to this warning.
+```hs
+forall b6 c7 d8. b6 -> c7 -> d8 -> Int
 ```
--- a/src/Language/PureScript/TypeChecker/Monad.hs
+++ b/src/Language/PureScript/TypeChecker/Monad.hs
@@ -49,11 +49,26 @@ data Substitution = Substitution
   -- ^ Type substitution
   , substUnsolved :: M.Map Int (UnkLevel, SourceType)
   -- ^ Unsolved unification variables with their level (scope ordering) and kind
+  , substNames :: M.Map Int Text
+  -- ^ The original names of unknowns
   }
+
+insertUnkName :: (MonadState CheckState m) => Unknown -> Text -> m ()
+insertUnkName u t = do
+  modify (\s ->
+            s { checkSubstitution =
+                  (checkSubstitution s) { substNames =
+                                            M.insert u t $ substNames $ checkSubstitution s
+                                        }
+              }
+         )
+
+lookupUnkName :: (MonadState CheckState m) => Unknown -> m (Maybe Text)
+lookupUnkName u = gets $ M.lookup u . substNames . checkSubstitution
 
 -- | An empty substitution
 emptySubstitution :: Substitution
-emptySubstitution = Substitution M.empty M.empty
+emptySubstitution = Substitution M.empty M.empty M.empty
 
 -- | State required for type checking
 data CheckState = CheckState
@@ -445,8 +460,12 @@ debugValue :: Expr -> String
 debugValue = init . render . prettyPrintValue 100
 
 debugSubstitution :: Substitution -> [String]
-debugSubstitution (Substitution solved unsolved) =
-  fmap go1 (M.toList solved) <> fmap go2 (M.toList unsolved')
+debugSubstitution (Substitution solved unsolved names) =
+  concat
+    [ fmap go1 (M.toList solved)
+    , fmap go2 (M.toList unsolved')
+    , fmap go3 (M.toList names)
+    ]
   where
   unsolved' =
     M.filterWithKey (\k _ -> M.notMember k solved) unsolved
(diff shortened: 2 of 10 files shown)


tests/purs/warning/4256.out
```diff
@@ -0,0 +1,16 @@
+Warning found:
+in module [33mMain[0m
+at tests/purs/warning/4256.purs:7:1 - 7:37 (line 7, column 1 - line 7, column 37)
+
+  No type declaration was provided for the top-level declaration of [33maddNumberSuffix'[0m.
+  It is good practice to provide type declarations as a form of documentation.
+  The inferred type of [33maddNumberSuffix'[0m was:
+  [33m                                        [0m
+  [33m  forall b6 c7 d8. b6 -> c7 -> d8 -> Int[0m
```
JordanMartinez, line 9: Shouldn't the inferred type here include the `a` argument, which is `Int`? So something like
```diff
- forall b6 c7 d8. b6 -> c7 -> d8 -> Int
+ forall b6 c7 d8. Int -> b6 -> c7 -> d8 -> Int
```

JordanMartinez: Ah, nevermind. The first argument has already been supplied, so this type signature is correct.

src/Language/PureScript/TypeChecker/Unify.hs
```diff
@@ -195,13 +195,31 @@ replaceTypeWildcards = everywhereOnTypesM replace
...
+
+  toBinding :: (Unknown, SourceType) -> m (SourceAnn, (T.Text, Maybe SourceType))
+  toBinding (a, k) = do
+    mn <- lookupUnkName a
+    k' <- go k
+    case mn of
+      Just n ->
+        pure (getAnnForType ty, (n <> T.pack (show a), Just k'))
+      Nothing ->
+        pure (getAnnForType ty, (toName a, Just k'))
```
JordanMartinez, line 214: Is it ever possible to hit this Nothing case? In other words, should this throw an `internalError` instead?

src/Language/PureScript/TypeChecker/Unify.hs
```diff
@@ -195,13 +195,31 @@ replaceTypeWildcards = everywhereOnTypesM replace
...
+
+  go :: SourceType -> m SourceType
+  go = everywhereOnTypesM $ \case
+    (TUnknown ann u) -> do
+      mn <- lookupUnkName u
+      case mn of
+        Just n ->
+          pure $ TypeVar ann (n <> T.pack (show u))
+        Nothing ->
+          pure $ TypeVar ann $ toName u
```
JordanMartinez, line 224: Similarly here, is it possible to hit this Nothing case?

src/Language/PureScript/TypeChecker/Unify.hs
PureFunctor: Untyped declarations would still trigger this e.g.
```hs
-- forall t1. t1 -> t1
id a = a
```

CHANGELOG.d/feature_restore_names.md
```diff
@@ -0,0 +1,33 @@
...
+  The inferred type of addNumberSuffix' was:
+
+    forall b6 c7 d8. b6 -> c7 -> d8 -> Int
+
+
+in value declaration addNumberSuffix'
+
+See https://github.com/purescript/documentation/blob/master/errors/MissingTypeDeclaration.md for more information,
+or to contribute content related to this warning.
+```
```
JordanMartinez, line 33: I'm wondering if this part should be shortened.  to something like

``````markdown
Previously, when a top-level declaration like the above example did not have a type signature, the compiler would produce a warning that stated the inferred type. The inferred type would use `t` followed by an integer:
    ```
    forall t6 t7 t8. t6 -> t7 -> t8 -> Int
    ```

Now, the type variables in the inferred type will refer back to their original names in the referenced function:
    ```
    forall b6 c7 d8. b6 -> c7 -> d8 -> Int
    ```
``````

File: src/Language/PureScript/TypeChecker/Unify.hs (hunk 2 above)
Comment (JordanMartinez) [reply src/Language/PureScript/TypeChecker/Unify.hs:214]: Ah, gotcha.

**Comment by JordanMartinez** [comment]
I'm asking this more out of curiosity. What does the inferred type look like when a type variable name includes an integer? For example:
```purescript
foo :: forall r1 r2 r3. r1 -> r2 -> r3 -> r1
foo r _ _ = r
```

PureFunctor:
It'd end up suffixing another digit regardless.
```hs
forall r25 r36. r25 -> r36 -> Int
```
I'm thinking maybe this could be helped by adding an `'`:
```hs
forall r2'5 r3'6. r2'5 -> r3'6 -> Int
```

JordanMartinez:
Good idea, but I think `'` is used more frequently than integers when quickly distinguishing one type var from another:
```purescript
foo :: forall sym a r r'.
  ( Row.Cons sym a r r'
  , IsSymbol sym
  ) => ....
```
which would produce
```
forall sym4 a5 r'6 r''7. ...
```

So, there's a few conflicting issues here:
- No matter which separater one uses, this issue will always arise if one terminates the ty var name with that same separator.
- If one accounts for the ty var name conditionally (e.g. "if it ends with integer, insert `'`; if ends with `'`, insert `_`; ..."), the inferred ty var names will no longer be consistent (e.g. `a1`, `a'1`, `a_1`)
- There are only so many chars one can use before appending extra chars to the ty var name makes it longer than it needs to be and thus unreadable / noisy (e.g. `a_tyVarNum1`, `a'_tyVarNum2`, `a__tyVarNum3`).
- While prepending the integer to the front of the ty var would make it easily distinguishable, a ty var name cannot start with an integer, so it breaks copy-paste workflows.

So, I think this issue I've brought up should be ignored. This PR does address the larger problem, namely, that `forall t0 t1 t2 t3` is harder to read and compare to its original function than `forall a0 b0 c0 d0`.

JordanMartinez (approved):
LGTM, aside from the changelog shortening.

JordanMartinez:
Also, I was curious to see what happens when multiple functions using the same ty var names are used together:
```purescript
-- @shouldWarnWith MissingTypeDeclaration
module Main where

foo :: forall a b c d. a -> b -> c -> d -> a
foo a _ _ _ = a

bar :: forall a b c d. a -> b -> c -> d -> a
bar a _ _ _ = a

baz a x y = bar (foo a 2 3 4) (foo a 2 3 4) (foo x y a a)
```
produces `forall c14 d15 b25 d27. d27 -> c14 -> b25 -> d15 -> d27`
```
Warning found:
in module [33mMain[0m
at tests/purs/warning/4256v2.purs:10:1 - 10:58 (line 10, column 1 - line 10, column 58)

  No type declaration was provided for the top-level declaration of [33mbaz[0m.
  It is good practice to provide type declarations as a form of documentation.
  The inferred type of [33mbaz[0m was:
  [33m                                                         [0m
  [33m  forall c14 d15 b25 d27. d27 -> c14 -> b25 -> d15 -> d27[0m
  [33m                                                         [0m

in value declaration [33mbaz[0m

See https://github.com/purescript/documentation/blob/master/errors/MissingTypeDeclaration.md for more information,
```

tests/purs/failing/ConstraintInference.out
```diff
@@ -4,7 +4,7 @@ at tests/purs/failing/ConstraintInference.purs:10:1 - 10:21 (line 10, column 1 -
 
   The inferred type
   [33m                                         [0m
-  [33m  forall t8 t11. Show t8 => t11 -> String[0m
+  [33m  forall c8 t11. Show c8 => t11 -> String[0m
```
MonoidMusician, line 7: Due to my hacky implementation of this error message, line 11 of the output which reads `t8 could not be determined` is broken, since it should read `c8` now. Is there any easy way to do the right thing when printing error messages, or should it be accessed when generating the error message (e.g. passing `[Text]` instead of `[Int]` in the error constructor `AmbiguousTypeVariables SourceType [Int]`)?

Here are the relevant changes from my PR, lmk if you have questions: https://github.com/purescript/purescript/pull/3721/files#diff-11e3ee9ff014e25b38a3dfdaa1fa18c8d0ca9dab1860d355abad0034978ea04cL691-R700

I will try to look at it myself soon but I can't make any promises.

MonoidMusician (changes requested):
The error message I added for the ambiguous fundep check needs to be updated to match (see inline comment), but other than that no objections from me.

tests/purs/failing/ConstraintInference.out
PureFunctor: I feel like changing the error constructor to `[Text]` would be simpler here, since the information comes from `CheckState`

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -322,6 +327,11 @@ instantiatePolyTypeWithUnknowns
   -> m (Expr, SourceType)
 instantiatePolyTypeWithUnknowns val (ForAll _ ident mbK ty _) = do
   u <- maybe (internalCompilerError "Unelaborated forall") freshTypeWithKind mbK
+  case u of
+    TUnknown _ i ->
+      insertUnkName i ident
+    _ ->
+      internalError "freshTypeWithKind didn't produce TUnknown"
```
PureFunctor, line 334: What's the difference between `internalCompilerError` and `internalError`? Should I prefer one over the other in this context?

rhendric: Ooh, I [know this one](https://github.com/purescript/purescript/pull/4126#issuecomment-887694913)!

PureFunctor: Thanks! I think `internalCompilerError` would be better here, even though `freshTypeWithKind` always returning an unknown is irrefutable.

tests/purs/failing/ConstraintInference.out
PureFunctor: I used a tuple `[(Text, Int)]` as to preserve information

MonoidMusician (approved):

src/Language/PureScript/TypeChecker/Unify.hs
```diff
@@ -195,13 +195,31 @@ replaceTypeWildcards = everywhereOnTypesM replace
...
+    t -> pure t
```
rhendric, line 225: Lots of duplicated code here; I suggest moving the `lookupUnkName` call into `toName`:

```hs
toName u = (<> T.pack (show u)) . fromMaybe "t" <$> lookupUnkName u               
```

And then, every call site of `lookupUnkName` feeds into a `fromMaybe "t"`, so you might as well move that into the definition of `lookupUnkName`.

src/Language/PureScript/TypeChecker/Unify.hs
```diff
@@ -195,13 +195,31 @@ replaceTypeWildcards = everywhereOnTypesM replace
...
-  toBinding (a, k) = (getAnnForType ty, (toName a, Just $ go k))
-  go = everywhereOnTypes $ \case
-    (TUnknown ann u)
-      | Just _ <- lookup u unks -> TypeVar ann (toName u)
-    t -> t
+  toName = T.cons 't' . T.pack . show
+
+  toBinding :: (Unknown, SourceType) -> m (SourceAnn, (T.Text, Maybe SourceType))
+  toBinding (a, k) = do
+    mn <- lookupUnkName a
```
rhendric, line 208: This isn't going to be relevant if you follow my other suggestion, but for future reference, I would avoid using `mn` as a variable name for anything other than a `ModuleName`. Something like `mbName` would have been a better choice here.

rhendric (changes requested):
Looks great, aside from some minor housekeeping issues! Thanks for taking this on.

JordanMartinez:
@rhendric Any other feedback on this? I believe the issues you raised have been resolved.

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -49,11 +49,26 @@ data Substitution = Substitution
...
+  modify (\s ->
+            s { checkSubstitution =
+                  (checkSubstitution s) { substNames =
+                                            M.insert u t $ substNames $ checkSubstitution s
+                                        }
+              }
+         )
+
+lookupUnkName :: (MonadState CheckState m) => Unknown -> m (Maybe Text)
+lookupUnkName u = gets $ M.lookup u . substNames . checkSubstitution
```
rhendric, line 67: I still think the common `fromMaybe "t"` should be moved into this function (so that it returns an `m Text`), but that's a minor nit and I only mention it in case you meant to and overlooked it.

rhendric (approved):

src/Language/PureScript/TypeChecker/Monad.hs
JordanMartinez: I'm going to merge this as is to ensure this doesn't bitrot. If that's something that was overlooked, it can be fixed in a separate PR.

PR #4261, opened 2022-03-09
Erase kind applications when loading type synonyms from externs for instance deriving

### LINKED ISSUE #4200: KindsDoNotUnify error when deriving Newtype from alias
## Description

The case, seems like a bug: 
- there is a parameterized data type and an alias for this data type (in practice defined in different modules)
- in another module deriving `Newtype` instance for a `newtype` that contains mentioned alias type
- the compiler gives an error `KindsDoNotUnify`

## To Reproduce

The minimal example to reproduce requires two modules:

```purs
module Test.A where

import Test.B (TAlias)

newtype NewA a = NewA (TAlias Int)

derive instance Newtype (NewA a) _
```

```purs
module Test.B where

data T :: forall m. m -> Type
data T msg = E

type TAlias :: forall k. k -> Type
type TAlias msg =
  T msg -- here is an error
```

The error:
```
  Could not match kind

    Type

  with kind

    k


while checking that type Int
  has kind k
while inferring the kind of T @k Int
while inferring the kind of Newtype (NewA a) (T @k Int)
in type class instance

  Data.Newtype.Newtype (NewA a)
                       (T @k Int)
PureScript(KindsDoNotUnify)
```

## Expected behavior

Should be no error.

## Additional context

It starts working if to replace `TAlias's` kind signature with `Type -> Type`

## PureScript version

0.14.5

PureFunctor:
Closes #4200. This makes it so that type synonyms loaded for the `deriveInstances` have their kind applications removed. This achieves the same effect as having the type synonym exist as a local declaration e.g. `fromLocalDecl`. 

Likewise, since an empty `kinds` is passed to `deriveInstance`, and in effect, to `replaceAllTypeSynonyms`, kind applications aren't actually replaced; this leads to the eventual `KindsDoNotUnify` error.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/fix_4200.md
+++ b/CHANGELOG.d/fix_4200.md
@@ -1,0 +1,27 @@
+* Erase kind applications when loading type synonyms from externs for instance deriving
+
+  This fixes a bug where polykinded type synonyms defined in other
+  modules cannot be used when deriving a `Newtype` instance.
+
+  Previously, given the `Test.B` module:
+  ```purs
+  module Test.B where
+
+  data T :: forall m. m -> Type
+  data T msg = E
+
+  type TAlias :: forall k. k -> Type
+  type TAlias msg = T msg
+  ```
+
+  `Test.A` would fail with `KindsDoNotUnify`:
+  ```purs
+  module Test.A where
+
+  import Test.B (T, TAlias)
+  import Data.Newtype (class Newtype)
+
+  newtype NewA a = NewA (TAlias Int)
+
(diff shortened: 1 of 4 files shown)


PureFunctor:
Alternatively, this could also be patched by having `KindApp`s not appear in externs at all.

rhendric:
Turns out this issue is also a consequence of #4086; I ran the test added here on my unreleased implementation of that issue and it passed without any additional changes.

PureFunctor:
Neat :smile:

PR #4267, opened 2022-03-15
Add Prim.Int.ToString compiler-solved class

### LINKED ISSUE #4263: Proposal: Add compiler-solved class for converting type-level `Int` to type-level `String` (Symbol)
## Summary

Add a compiler-solved type class that converts type-level integers into type-level strings.

## Motivation

One of the drawbacks of the JS backend is that it doesn't support large integers. Thus, reflecting a large type-level Int to a value-level Int can be impossible to do. Consider the following example in the REPL:
```purescript
> (top :: Int)
2147483647
> reflectType (Proxy :: Proxy 2147483647)
2147483647
> -- add 1
> reflectType (Proxy :: Proxy 2147483648)
An error will be thrown here
> (bottom :: Int)
-2147483648
> reflectType (Proxy :: Proxy (-2147483648))
-2147483648
> -- subtract 1
> reflectType (Proxy :: Proxy (-2147483649))
An error will be thrown here
```

This issue could be worked around if we could instead convert the type-level integer into a type-level string, reflect that type-level string to a value-level one, and then call `Data.BigInt.fromString`.

## Proposal

Since I don't yet understand the compiler deeply, I'm not sure how much of the below content is correct. Correct me where I'm wrong.

Given that `Reflectable` is a class that will be defined like so (see https://github.com/purescript/purescript-prelude/issues/286)
```purescript
class Reflectable v t | v -> t where
  reflectType :: Proxy v -> t
```
we essentially want the compiler to solve the type class via the instances below:
```purescript
-- Assume backtracking works in the below example
instance Between (-2147483648) 2147483647 value => Reflectable (value :: Prim.Int) Int where
  reflectType _ = jsIntValue
else instance Reflectable (allOtherValues :: Prim.Int) String where
  reflectType _ = stringifiedVersion
```

There are two issues with the above implementation. First, the functional dependency ensures that `Prim.Int` determines the type we return. In one version, we should return `Int`. In another, we should return `String`. To fix this, we would need to drop the functional dependency. But if that was done, the solver wouldn't know how to determine what `t` is.

We could get around that issue by wrapping the returned value in `Either`...
```purescript
-- Assume backtracking works in the below example
instance Between (-2147483648) 2147483647 value => Reflectable (value :: Prim.Int) Int where
  reflectType _ = Right jsIntValue
else instance Reflectable (allOtherValues :: Prim.Int) String where
  reflectType _ = Left stringifiedVersion
```
... but that's not ideal.

Second, backtracking isn't yet supported (and I'm not sure whether and when it will be supported). So, how would the compiler solve the class? Is backtracking only needed when implementing instances in PureScript? Or does the compiler need it to solve things, too?

It would seem that both problems can be solved if we provide another type-level type to indicate which kind of `Int` it is. For example `Prim.Int` would be reflectable to a normal JS Int whereas `Prim.BigInt` would always reflect to a `String` that can be used by `Data.BigInt`.

I'm not sure how that would affect syntax though. If we wrote `type TypeLevelInt = 14`, does PureScript interpret that as having kind `Prim.Int` or kind `Prim.BigInt`? Do we always need to indicate which one it is via some annotation? Do we add some suffix/prefix that indicates which kind of int it is (e.g. `14i` vs `14bi`?)?

See also #3848 for other context on a similar discussion.

JordanMartinez:
**Description of the change**

Fixes #4263. I looked through #4207 to guide me on how to implement this.

---

**Checklist:**


PureFunctor (commented):
This LGTM, but would there be any merit to a `FromString` type class as well?

Similarly, would generalizing `ToString`/`FromString` such that they can be used for `Boolean`/`Ordering` be a good idea? I figured not really, considering that unlike `Int` they don't have an arbitrary number of members.

JordanMartinez:
> would there be any merit to a FromString type class as well?

I am not sure. For one, if it's needed, we can add it later without making a breaking change. For two, I'm not sure what situations may arise where it would be desirable to do that. Also, a `FromString` that works on `Boolean`/`Ordering` can be defined in PureScript without a huge performance hit (unlike `Int`'s `FromString`), so I don't think it needs compiler support.

paf31:
Isn't `FromString` just the functional dependency run backwards?

JordanMartinez:
> Isn't `FromString` just the functional dependency run backwards?

Yes.

JordanMartinez:
> [](https://github.com/purescript/purescript/pull/4267#issuecomment-1074346770)Isn't FromString just the functional dependency run backwards?

Actually, no, because an integer could be rendered in a couple of different ways. In PureScript syntax, we support both `123` and `0x0aF` as valid integers. Should a type-level conversion also support that?

In other words, should `ToString` actually be something like this?
```purescript
-- another compiler kind
data IntBase
foreign import data Base10 :: IntBase
foreign import data Base16:: IntBase

class ToString :: Int -> Base -> Symbol -> Constraint
class ToString int base string | int base -> string, string base -> int

-- compiler solved instances
instance ToString 123 Base10 "123"
instance ToString 123 Base16 "0x7B"
```

JordanMartinez:
I've spent some time on updating this PR. However, there's a number of factors to consider. I'm going to open an issue where this can be discussed further.

JordanMartinez:
Ha, there's already an issue. I'll further discuss there.

JordanMartinez:
Per the discussion in #4263, I believe this is ready for a final review.

rhendric (changes requested):
Everything here looks solid, but I'd like to see the following tested (could be the same test):
* A case where the integer being `ToString`'d is not a literal—use `Add` or `Mul` to compute an integer that doesn't appear in the source, and then `ToString` that.
* A case where the integer is big (larger than 2^64), given the initial motivation for this class.

JordanMartinez:
`IntToString.purs` passes locally.

rhendric (approved):

PR #4286, opened 2022-04-11
Add GNU Guix installation instructions

jgarte:
This PR adds installation instructions with [Guix](https://guix.gnu.org/en/packages/purescript-0.14.5/).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -24,6 +24,7 @@ There are several other distributions of the PureScript compiler available, whic
 * NPM: `npm install -g purescript`
 * Homebrew (for macOS): `brew install purescript`
 * FreeBSD binary packages: `pkg install hs-purescript`
+* GNU Guix: `guix install purescript`
 
 ## Compiling from source
 


rhendric (approved):

JordanMartinez:
Does this need a changelog?

rhendric:
I can't imagine why a user would need to know about small updates to the project's README file when considering a version upgrade, so I would say no.

JordanMartinez:
> I can't imagine why a user would need to know about small updates to the project's README file when considering a version upgrade, so I would say no.

I don't either, but the `update-changelog.hs` script does have an "Internal" section that I thought existed for these kinds of things. If a user was only watching this repo for releases, the changelog would list this change, which might catch that user's attention if they use Guix. 

If we don't add a changelog entry for this PR in the "internal" section, then could you clarify your understanding of what changes should be listed in the "internal" section?

Also, I don't mean to derail this PR. It's ready to be merged and this question is just a housekeeping thing.

rhendric:
I'm not sure we have a consistently-applied policy for what goes in the changelog and where, to be clear. When I made `update-changelog.hs` I included the sections that we were using most recently in our manually-written changelog file, without worrying about whether they were the right sections. I'm not sure my personal opinions on the matter are reflective of the rest of the team's; I've been in the minority on earlier disagreements about what is and isn't appropriate for the changelog.

That said, my personal take is that the Internal section is for listing any changes that we don't expect users to need or want to know about but which might be useful starting places for investigations if adverse effects pop up. Refactoring, changes to dependency versions, etc. are central examples. Less central but still positive examples are reformatting comments (what if it breaks Hoogle docs?) and changing the CI workflow (what if it breaks our published binaries?). Negative examples would be changes to the core team, links to external projects that use PureScript, announcements of upcoming PureScript conferences, etc.; all of which PureScripters might want to know about but aren't associated with any particular release of PureScript, even if we may choose to include some of that information in the versioned code repository. (I also think entries for adding tests without fixing associated bugs don't belong in the changelog either if the changes made don't touch the main compilation path, but we historically have included such—so take my opinion with salt!)

And yeah, I don't think we're holding anything up with this convo; I'm just waiting a day or two for any reason we might not want this to pop up (I don't expect any).

garyb:
I think that sounds sensible for things worth including in the Internal section too. I wasn't really sure what the value of it in general was, but having it as a place to refer to for investigations when there are changes that aren't supposed to have an observable effect on the compiler makes a lot of sense.

JordanMartinez:
> That said, my personal take is that the Internal section is for listing any changes that we don't expect users to need or want to know about but which might be useful starting places for investigations if adverse effects pop up. Refactoring, changes to dependency versions, etc. are central examples. Less central but still positive examples are reformatting comments (what if it breaks Hoogle docs?) and changing the CI workflow (what if it breaks our published binaries?). Negative examples would be changes to the core team, links to external projects that use PureScript, announcements of upcoming PureScript conferences, etc.; all of which PureScripters might want to know about but aren't associated with any particular release of PureScript, even if we may choose to include some of that information in the versioned code repository.

I'm good with this and think it might be helpful to document it.

> (I also think entries for adding tests without fixing associated bugs don't belong in the changelog either if the changes made don't touch the main compilation path, but we historically have included such—so take my opinion with salt!)

Yeah, I'm sort of mixed on this point.

Regardless, if no other comments are made, let's merge this without a changelog entry tomorrow.

PR #4288, opened 2022-04-14
Document our evolving changelog principles

rhendric:
**Description of the change**

Motivated by the conversation in #4286, this is an opinionated attempt to add more clarity to our CHANGELOG.d/README.md documentation. Please let me know if these opinions match yours.

One opposing position I want to highlight is the use of the changelog as a way to give credit to people who contributed to a given release. The set ‘changes to the compiler’ is not exactly the same as the set ‘contributions to the project’, and in this formulation, the changelog focuses entirely on the former. Someone who comes in and massively improves some aspect of our test suite, for example, would deserve a shout-out, but this wouldn't qualify as a change to the compiler that users would need to know about when considering an upgrade. Perhaps this is what the ‘Other improvements’ (`misc`) section should be for? Or maybe I have ‘Internal’ and ‘Other improvements’ swapped?

In accordance with the revised documentation here, this PR doesn't include a changelog entry. 😄

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/README.md
+++ b/CHANGELOG.d/README.md
@@ -6,13 +6,32 @@ Maintainers: see update-changelog.hs for details of this process.
 
 Contributors: read on!
 
-When you are preparing a new PR, add a new file to this directory. The file
-should be named `{PREFIX}_{SLUG}.md`, where `{PREFIX}` is one of the following:
-* `breaking`: for breaking changes
-* `feature`: for new features
-* `fix`: for bug fixes
-* `internal`: for work that will not directly affect users of PureScript
-* `misc`: for anything else that needs to be logged
+Our guiding principle is that the changelog is a tool for users—people who
+depend on PureScript as a compiler or as a library—who are considering
+upgrading, or have recently upgraded, their PureScript compiler version. We ask
+that when making changes that such users might need to know about, you help
+them out by adding to our changelog.
+
+Work that doesn't change the compiler (such as updates to README.md) doesn't
+need a changelog entry. But keep in mind that even parts of the project like
+our CI workflow can introduce changes to the compiler we release.
+
+When you are preparing a new PR that does change the compiler, add a new file
+to this directory. The file should be named `{PREFIX}_{SLUG}.md`, where
+`{PREFIX}` is one of the following:
+* `breaking`: for breaking changes to the compiler, for which a user may need to do
+  work to their project before or immediately upon upgrading
+* `feature`: for new features, which might prevent a user from downgrading to an
+  earlier version
+* `fix`: for bug fixes, which might motivate a user to upgrade
+* `internal`: for work that is not expected to directly affect users; these
+  entries should usually be brief, but may serve as useful starting points for
+  investigations if a change ends up having unintended consequences
+
+(There is also a fifth prefix, `misc`. This is an escape hatch in case we have
+something that somehow doesn't fit in the above categories but that we want to
+include in the changelog, which frankly seems unlikely given how much of a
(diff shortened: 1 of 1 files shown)


JordanMartinez:
> In accordance with the revised documentation here, this PR doesn't include a changelog entry. smile

The irony...

> Someone who comes in and massively improves some aspect of our test suite, for example, would deserve a shout-out, but this wouldn't qualify as a change to the compiler that users would need to know about when considering an upgrade.

Two issues arise with this fact:
- remembering whom to do a shoutout for when releases occur over a lengthy period of time
- how to do this shoutout

In regards to the second one, we could state something in the Discord release announcement.

JordanMartinez:
Coming back to this, is it possible we could add some of these changes to a separate dir that functions similarly to the `CHANGELOG.d` one used right now? That way, we could still track who did what, but it doesn't need to appear in the release notes for people using PureScript. If anything, it might be helpful for maintainers or new contributors to see what else is going on in the project.

rhendric:
Is that more useful than viewing the commit history?

JordanMartinez:
> Is that more useful than viewing the commit history?

Somewhat. Rather than viewing the history in a linear way (commit history), this "internal changelog" gives a high-level summary. Moreover, our current process of squashing commits sometimes leaves a mess of various bullet points stating "fixed bug"-like things that aren't always clear. If we used a workflow similar to our current changelog approach, but as a separate one that exists for this purpose, then the contributor can write a cleaner high-level summary.
It would also solve the 'who do we give a shout out for when making a release?' problem.

CHANGELOG.d/README.md
```diff
@@ -6,21 +6,40 @@ Maintainers: see update-changelog.hs for details of this process.
...
+  earlier version
+* `fix`: for bug fixes, which might motivate a user to upgrade
+* `internal`: for work that is not expected to directly affect users; these
+  entries should usually be brief, but may serve as useful starting points for
+  investigations if a change ends up having unintended consequences
+
+(There is also a fifth prefix, `misc`. This is an escape hatch in case we have
+something that somehow doesn't fit in the above categories but that we want to
+include in the changelog, which frankly seems unlikely given how much of a
+catch-all `internal` is. We'll tell you if you should use this one.)
```
JordanMartinez, line 34: Why don't we drop this completely (an undocumented feature) and only bring it up if we need this escape hatch? It seems like 'internal' might be good enough.

JordanMartinez (approved):
Aside from my one comment above, I think we can merge this and resolve the other issues later.

CHANGELOG.d/README.md
rhendric: I figure this is partly documentation for contributors but also partly documentation for us—I don't want to leave us wondering in the future whether there was an explicit plan for the ‘Other improvements’ section. But maybe there's a better place than this file for that institutional knowledge?

Also and conversely, this is kind of a consensus-building PR; I'm calling this out explicitly as somewhat vestigial so that other core members might see and say, no, there's more intent than that behind this section.

garyb: I think maybe we used `misc` in the ancient past for what we'd call `internal` now. I can't think of anything we'd particularly need it for in the future, given the current guidelines.

JordanMartinez: So, is there consensus here that we should drop `misc`?

JordanMartinez: Again, is there consensus about dropping `misc` here?

JordanMartinez:
There's been plenty of time for feedback on this PR and so far no objections have been raised. I say we merge this as is within this week.

Issue #4247, opened 2022-02-26
Code generation target for emitting name and type information


PureFunctor:
## Proposal
Currently, PureScript's IDE capabilities make use of externs files to provide completion and type information to clients of `purs ide server`. It's sufficient enough for most use-cases although a coveted feature that's still missing is completion and type information for local bindings e.g. `let`-bound values or function arguments. Likewise, `purs ide` can be unusable for lower-spec machines especially on larger web projects.

This proposal aims to add a new JSON-based codegen target based on GHC's [HIE files](https://www.haskell.org/ghc/blog/20190626-HIEFiles.html), which is a plaintext format which stores name and type information for source files. It's used by `ghci` to provide type-at-point capabilities as well as completion, and I believe we can likely do the same with `purs repl`. There also exists [hiedb](https://github.com/wz1000/HieDb), which is a command-line tool for building up a database of `.hie` files that can be queried.

The data to be emitted will most likely come from the type checker, likely after a declaration has been type-checked and added to the environment. We could also feed this information to the IDE infrastructure in order to provide richer information, though we'd also want to remove externs in said infrastructure.

I'd like to work on this in the future once we've released 0.15.x, or at least once #4207 and #4235 is resolved.

thomashoneyman:
@PureFunctor We have a use case in the registry in which we'd like to check whether the public interface of a module has changed (ie. any public types or type signatures have changed, been added, removed, etc). Would this proposal make it possible to verify this information?

PureFunctor:
@thomashoneyman Yes, it will also have information about exported symbols similar to externs files, on top of information for local bindings and such.

kritzcreek:
If you're only interested in the public interface, the externs files already carry that information today.

JordanMartinez:
> If you're only interested in the public interface, the externs files already carry that information today.

I think the goal is the non-public info found in 'let bindings'. While I can figure out what the type of `foo` is, I don't know what `bar` is:
```purescript
foo :: String
foo = show bar
  where
  bar = ...
```

PureFunctor:
I'm wondering how this would interact with #3724, though. If we're going to go through with both, we might want to take a conservative approach to fast rebuilds.

rhendric:
[...]

PR #4293, opened 2022-04-17
Qualify locally-bound names with source spans

### LINKED ISSUE #4287: Proposal: Add a renaming phase for the AST
## Summary

This proposal aims to add a renaming phase for the AST, performed after parsing and before type checking a module. This would enable easier syntax-driven analysis for modules in the compiler, such as generating interface files for IDE information (#4247).

## Motivation

The primary motivation for this proposal is the 2nd approach that I've mentioned for collecting type information in modules (https://github.com/purescript/purescript/issues/4247#issuecomment-1094834125), which involves analyzing the AST for any `Expr` that happens to be wrapped around `TypedExpr _ (PositionedExpr _ _ e) _`, instead of hooking into the type checker directly. This step is also inspired by GHC's pipeline, as seen [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/compiler/).

## Proposal

A renaming phase for the AST would be built which makes all names in a module unique. Renaming would be done after parsing and before type checking. We're really only interested in renaming locally-introduced names such as those from lambas and `let`-declarations. `Ident`s would optionally be accompanied by a unique identity introduced by the renamer, and they would only be visible to the compiler. Other steps in the compilation pipeline such as code generation may or may not choose to use them.

## Examples

In this particular example, if I want to attribute the use of `id` in `id 0` to the locally-bound value, I'd have some trouble doing so because of the introduction of `id` by `Prelude`.
```hs
module Main where

import Prelude

zero :: Int
zero = 
  let
    -- The compiler monomorphizes this into `Int -> Int`
    id a = a
  in 
    id 0
```

Running the renamer through this would produce:
```hs
module Main where

import Prelude

zero :: Int
zero = 
  let
    -- The compiler monomorphizes this into `Int -> Int`
    id_22 a_23 = a_23
  in 
    id_22 0
```
In this case, the locally-bound `id` is now accompanied by a unique identity, which trivializes information collection. The same effect is applied to the `a`. On the other hand, the top-level binding `zero` is unaffected by the renamer as it's a module-level declaration.

PureFunctor:
**Description of the change**

Closes #4287. This adds a new desugaring step, `desugarLocals`, which turns references to local bindings be qualified by the source spans of the said bindings they're referring back to. Note that this had to be a separate step from `desugarImports` as it also needs to take into account binding groups. This also doesn't introduce new tests as an incorrect implementation breaks all other compiler tests quite heavily.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_local_qualification.md
+++ b/CHANGELOG.d/feature_local_qualification.md
@@ -1,0 +1,9 @@
+* Add qualification for locally-bound names
+
+  This change makes it so that `Qualified` names can now be qualified by either
+  a `ModuleName` for module-level declarations or a `SourceSpan` for bindings
+  introduced locally. This makes disambiguation between references to local
+  bindings much easier in AST-driven analysis.
+
+  A new desugaring step, `desugarLocals`, has also been added to insert
+  `SourceSpan`s into references of local bindings in the AST.
--- a/src/Language/PureScript/CoreFn/ToJSON.hs
+++ b/src/Language/PureScript/CoreFn/ToJSON.hs
@@ -18,7 +18,7 @@ import           Data.Text (Text)
 import qualified Data.Text as T
 
 import           Language.PureScript.AST.Literals
-import           Language.PureScript.AST.SourcePos (SourceSpan(SourceSpan))
+import           Language.PureScript.AST.SourcePos (SourceSpan(..))
 import           Language.PureScript.CoreFn
 import           Language.PureScript.Names
 import           Language.PureScript.PSString (PSString)
@@ -94,10 +94,18 @@ properNameToJSON :: ProperName a -> Value
 properNameToJSON = toJSON . runProperName
 
 qualifiedToJSON :: (a -> Text) -> Qualified a -> Value
-qualifiedToJSON f (Qualified mn a) = object
-  [ T.pack "moduleName"   .= maybe Null moduleNameToJSON mn
-  , T.pack "identifier"   .= toJSON (f a)
-  ]
+qualifiedToJSON f (Qualified qb a) =
+  case qb of
+    ByModuleName mn -> object
+      [ T.pack "moduleName" .= moduleNameToJSON mn
+      , T.pack "identifier" .= toJSON (f a)
+      ]
+    BySourceSpan ss -> object
+      -- TODO: Should `sourceSpanXJSON` encode this information as well?
+      [ T.pack "spanName"   .= toJSON (spanName ss)
+      , T.pack "sourceSpan" .= sourceSpanToJSON ss
+      , T.pack "identifier" .= toJSON (f a)
+      ]
 
 moduleNameToJSON :: ModuleName -> Value
 moduleNameToJSON (ModuleName name) = toJSON (T.splitOn (T.pack ".") name)
--- a/src/Language/PureScript/Sugar/Names.hs
+++ b/src/Language/PureScript/Sugar/Names.hs
@@ -412,27 +412,49 @@ renameInModule imports (Module modSS coms mn decls exps) =
 desugarLocals :: Monad m => Module -> m Module
 desugarLocals (Module ms cm mn dc ex) = pure $ Module ms cm mn dc' ex
   where
-  dc' :: [Declaration]
-  dc' = (flip evalState M.empty . go) <$> dc
+  dc' = (flip evalState (M.empty, []) . go) <$> dc
 
-  go :: Declaration -> State (M.Map Ident SourceSpan) Declaration
-  (go, _, _) = everywhereOnValuesTopDownM goDeclaration goExpr goBinder
+  insertOne k v = modify' $ \(e, b) ->
+    (M.insert k v e, b)
+
+  insertMany e' = modify' $ \(e, b) ->
+    (M.union e' e, b)
+
+  lookupIdent k = gets (M.lookup k . fst)
+
+  pushScope = modify' $ \(e, b) ->
+    (e, e : b)
+
+  popScope = modify' $ \(e, b) -> case b of
+    [] -> (e, [])
+    (e' : b') -> (e', b')
+
+  (go, _, _) = everywhereOnValuesTopDownThenM goDecl goExpr goBinder pure thenExpr pure
     where
-    goDeclaration declaration = case declaration of
+    goDecl declaration = case declaration of
       ValueDeclaration ValueDeclarationData{..} -> do
-        modify $ M.insert valdeclIdent $ fst valdeclSourceAnn
+        insertOne valdeclIdent $ fst valdeclSourceAnn
         pure declaration
       BindingGroupDeclaration declarations -> do
         let findSsI (((ss, _), i), _, _) = (i, ss)
-        modify $ M.union $ M.fromList $ NEL.toList $ findSsI <$> declarations
+        insertMany $ M.fromList $ NEL.toList $ findSsI <$> declarations
         pure declaration
       _ ->
         pure declaration
 
     goExpr expression = case expression of
-      Var ss (Qualified (BySourceSpan _) i@(Ident _)) -> do
-        bound <- get
-        pure $ case M.lookup i bound of
+      Abs _ _ -> do
+        pushScope
+        pure expression
+      Let _ _ _ -> do
+        pushScope
+        pure expression
+      Case _ _ -> do
+        pushScope
+        pure expression
+      Var ss (Qualified (BySourceSpan _) i) -> do
+        result <- lookupIdent i
+        pure $ case result of
           Just ss' ->
             Var ss (Qualified (BySourceSpan ss') i)
           Nothing ->
@@ -440,9 +462,25 @@ desugarLocals (Module ms cm mn dc ex) = pure $ Module ms cm mn dc' ex
       _ ->
         pure expression
 
+    thenExpr expression = case expression of
+      Abs _ _ -> do
+        popScope
+        pure expression
+      Let _ _ _ -> do
+        popScope
+        pure expression
+      Case _ _ -> do
+        popScope
+        pure expression
+      _ ->
+        pure expression
+
     goBinder binder = case binder of
       VarBinder s i -> do
-        modify $ M.insert i s
+        insertOne i s
+        pure binder
+      NamedBinder s i _ -> do
+        insertOne i s
         pure binder
       _ ->
         pure binder
--- a/src/Language/PureScript/Sugar/Names.hs
+++ b/src/Language/PureScript/Sugar/Names.hs
@@ -1,5 +1,6 @@
 module Language.PureScript.Sugar.Names
   ( desugarImports
+  , desugarLocals
   , Env
   , externsEnv
   , primEnv
@@ -19,6 +20,7 @@ import Control.Monad.State.Lazy
 import Control.Monad.Writer (MonadWriter(..))
 
 import Data.Maybe (fromMaybe, mapMaybe)
+import qualified Data.List.NonEmpty as NEL
 import qualified Data.Map as M
 import qualified Data.Set as S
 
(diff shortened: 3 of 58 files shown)


src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -94,10 +94,18 @@ properNameToJSON :: ProperName a -> Value
...
-  ]
+qualifiedToJSON f (Qualified qb a) =
+  case qb of
+    ByModuleName mn -> object
+      [ T.pack "moduleName" .= moduleNameToJSON mn
+      , T.pack "identifier" .= toJSON (f a)
+      ]
+    BySourceSpan ss -> object
+      -- TODO: Should `sourceSpanXJSON` encode this information as well?
+      [ T.pack "spanName"   .= toJSON (spanName ss)
```
PureFunctor, line 105: Is exposing the `spanName` here potentially a bad idea? Should we aim to make a default for this like `"<local>"`?

JordanMartinez:
@PureFunctor Just to clarify, is this a PR that has breaking changes that we may need to consider including in `0.15.0`? Or can this wait until after `0.15.0` is released?

PureFunctor:
@JordanMartinez Yeah, this is breaking change for consumers of the compiler as a library. In particular, it changes the signature for the `Qualified` data structure such that it has more information.

I'm just curious, have there been any instances where minor releases included breaking changes for the public-facing API of the compiler library? I'm fine with getting this potentially merged in after `0.15.0`, if it also means bundling this with #4247.

JordanMartinez:
@PureFunctor Gotcha. Thanks for clarifying.

> Yeah, this is breaking change for consumers of the compiler as a library. In particular, it changes the signature for the Qualified data structure such that it has more information.

AFAIK, the only consumers at this point are:
- trypurescript
- purescript
- potentially backends that use a fork of the `purescript` project to implement their backend?
- language-server tools (but these likely don't directly depend on this project)

I'm not sure what our policy is for breaking changes that affect consumers of the library but not necessarily users of the language.

src/Language/PureScript/CoreFn/ToJSON.hs
JordanMartinez: I'm not sure.

JordanMartinez: @nwolverson would this be something that affects the purerl backend?

nwolverson:
If it's a breaking change to corefn that should be considered on its own
merits - currently using purescript as a library for corefn, so the details
don't matter. Have version dependency for other reasons anyway

PureFunctor:
I just synced back to `master` for this PR and fixed the compilation errors that it introduced. Unfortunately, the implementation has been left in a broken state and I've a lot of work to do for this to be merged.

PureFunctor:
Tests compile locally and aside from the comment about source span names, this should be ready for review.

src/Language/PureScript/CoreFn/ToJSON.hs
JordanMartinez: Coming back to this question, what does `spanName` store? The only reference I found using a cursory glance was https://github.com/purescript/purescript/blob/f9f75f43bd1bdbc1e866fdba876582385e3f4c65/tests/Language/PureScript/Ide/RebuildSpec.hs#L75

Or another way to think about this question. If this was not added, could it be added in the future without incurring a breaking change?

I don't know enough about `spanName` to make a judgement here.

PureFunctor: `spanName` usually stores the absolute(?) file path to the source file alongside the source spans, and in certain cases may not contain information at all (i.e. null source spans, which have now been repurposed to mean "generated by the compiler"). Being an absolute file path may or may not be a good thing if there's some sensitive information located in the path name, although I can't say for certain since I don't have that constraint.

Since `Qualified` names now rely on source spans to be disambiguated through `BySourceSpan`, we need to stabilize `spanName` somehow.

JordanMartinez: > Being an absolute file path may or may not be a good thing if there's some sensitive information located in the path name, although I can't say for certain since I don't have that constraint.

Ah... I see. I don't know how often this occurs, but I'd leaned towards not worrying about this. Or to state it differently, I don't have this constraint either.

> Since Qualified names now rely on source spans to be disambiguated through BySourceSpan, we need to stabilize spanName somehow.

Sounds like this is the next course of action then?

JordanMartinez: If privacy is a concern, could `spanName` be changed to store only the relative reference for `BySourceSpan` values?

PureFunctor: Ok, I think the absolute paths come from how the `purs` binary is invoked and given the source files. When I do something like `purs compile Main.purs`, it sets the `modulePath` for a module to an absolute path, although when I do `purs compile src/Main.purs`, the `modulePath` just shows as relative.

I've only inspected the output of this change using the former command on a temporary folder, which explains the absolute paths. Would it be better to extract this to a separate PR?

JordanMartinez: > Would it be better to extract this to a separate PR?

Sure!

PureFunctor:
This PR is ready for review :tada:

src/Language/PureScript/TypeChecker/Monad.hs
```diff
@@ -196,8 +196,8 @@ withTypeClassDictionaries entries action = do
 
   let mentries =
         M.fromListWith (M.unionWith (M.unionWith (<>)))
-          [ (mn, M.singleton className (M.singleton (tcdValue entry) (pure entry)))
-          | entry@TypeClassDictionaryInScope{ tcdValue = Qualified mn _, tcdClassName = className }
+          [ (getQual tcdValue, M.singleton className (M.singleton tcdValue (pure entry)))
+          | entry@TypeClassDictionaryInScope{ tcdValue, tcdClassName = className }
```
JordanMartinez, line 200: Could you clarify why the `tcdValue = Qualified mn _` line was dropped?

JordanMartinez (commented):
LGTM. I had a few questions and small nitpicks before approving this.

src/Language/PureScript/Sugar/Names/Env.hs
```diff
@@ -229,13 +229,16 @@ mkPrimExports ts cs =
     , exportedTypeClasses = M.fromList $ mkClassEntry `map` M.keys cs
     }
   where
-  mkTypeEntry (Qualified mn name) = (name, ([], primExportSource mn))
-  mkClassEntry (Qualified mn name) = (name, primExportSource mn)
+  mkTypeEntry (Qualified (ByModuleName mn) name) = (name, ([], primExportSource mn))
+  mkTypeEntry _ = internalError "mkPrimExports.mkTypEntry: invalid argument"
+
+  mkClassEntry (Qualified (ByModuleName mn) name) = (name, primExportSource mn)
+  mkClassEntry _ = internalError "mkPrimExports.mkClassEntry: invalid argument"
```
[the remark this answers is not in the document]
PureFunctor: > Or does that affect a lot of error messages in our golden tests?
These are internal compiler errors so not really. Although yep, this could be helped by updating the error messages.

rhendric:
Can't remember if I asked this before, but why a `SourceSpan` instead of a `SourcePos`? I'd think that the extra information contained by a `SourceSpan` is all redundant—we don't need a file name because it's necessarily a reference to an identifier in the same file, and we know the span represents an occurrence of a given identifier so the length of the identifier is enough to determine where the end position is from the start position. Given that we serialize the representation of qualified names a lot, I'd rather not bloat up that representation unnecessarily.

PureFunctor:
> Can't remember if I asked this before, but why a SourceSpan instead of a SourcePos?

I can't remember exactly if it was intentional of me to use a `SourceSpan` or I just got confused with the types 😄.

> Given that we serialize the representation of qualified names a lot, I'd rather not bloat up that representation unnecessarily.

But yes, this is a valid concern that I'd missed and I could refactor the PR today to accommodate for that 🙂

rhendric (changes requested):
AST changes and the changes downstream of that all look good. Assuming there's a good reason I didn't understand for making a distinct pass for the renaming logic, just a few minor quibbles there; if there isn't a good reason, I'd prefer to see this integrated into the existing pass, since said pass duplicates a lot of the same ideas (keeping track of bound variables and updating them) and wouldn't require a new generic AST traversal to be defined.

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -382,20 +384,101 @@ renameInModule imports (Module modSS coms mn decls exps) =
...
     where
     throwUnknown = throwError . errorMessage . UnknownName . fmap toName $ qname
+
+-- |
+-- Qualifies locally-bound names such that they're indexed by the `SourceSpan`
+-- of the definition they're referencing. Names that are already qualified by
+-- a module name aren't affected at all.
+--
+-- Note: Unlike @desugarImports@, this function requires that binding groups are
+-- created first, hence the reason why this is a separate desugaring step.
```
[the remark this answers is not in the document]
PureFunctor: I had to remind myself for a bit but I think 2 months ago I had trouble disambiguating the idents in the following example:
```hs
module Main where

test :: Int
test =
  let
    b _ = 'a'
  in
    let
      a _ = b 0
      b _ = a 0
    in
      0
```
I can't remember if I'd used `Map Ident SourcePos` before as an approach but I ought to try that out now that you point it out

CHANGELOG.d/feature_local_qualification.md
```diff
@@ -0,0 +1,9 @@
+* Add qualification for locally-bound names
```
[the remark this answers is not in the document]
PureFunctor: Yup, I was just unsure of our policy for compiler-as-a-library users back then

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -382,20 +384,101 @@ renameInModule imports (Module modSS coms mn decls exps) =
...
+desugarLocals :: Monad m => Module -> m Module
```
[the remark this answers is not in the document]
PureFunctor: It composes a lot more nicely 😄

JordanMartinez:
Just a thought, but should we distinguish `ByNullSourcePos` as its own data constructor?

Also, on looking at this again, there's not a test that that verifies that a given local binding has the expected `BySourcePos`. Could one be added?

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -244,21 +244,24 @@ renameInModule imports (Module modSS coms mn decls exps) =
...
       throwError . errorMessage' pos $ OverlappingNamesInLet
-    return ((pos, args ++ bound), Let w ds val')
-  updateValue (_, bound) (Var ss name'@(Qualified Nothing ident)) | ident `notElem` bound =
-    ((ss, bound), ) <$> (Var ss <$> updateValueName name' ss)
-  updateValue (_, bound) (Var ss name'@(Qualified (Just _) _)) =
+    return ((pos, declarationsToMap ds `M.union` bound), Let w ds val')
+  updateValue (_, bound) (Var ss name'@(Qualified (BySourcePos _) ident)) =
+    ((ss, bound), ) <$> case M.lookup ident bound of
+      Just sourcePos -> pure $ Var ss (Qualified (BySourcePos sourcePos) ident)
+      Nothing -> Var ss <$> updateValueName name' ss
```
rhendric, line 262: Does this happen? I'd expect this branch to be an internal error.

PureFunctor: I believe it does, since the `ident` here is qualified by a null source position by default. I can do a quick trace to check, though.

rhendric: Oh, I think we miscommunicated. I meant the `Nothing` branch on this one line, which should only happen if `ident` is not in `bound` and is also not qualified by a module; I think those two conditions should be exhaustive. I think the `ByNullSourcePos` was better as a `BySourcePos _` pattern, though if a non-null `SourcePos` gets in there, maybe we should error on that too.

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -285,23 +287,38 @@ renameInModule imports (Module modSS coms mn decls exps) =
...
     updateGuard (GuardedExpr g _ : xs) =
-      concatMap updatePatGuard g ++ updateGuard xs
+      updateGuard xs `M.union` rUnionMap updatePatGuard g
       where
-        updatePatGuard (PatternGuard b _) = binderNames b
-        updatePatGuard _                  = []
+        updatePatGuard (PatternGuard b _) = binderNamesWithSpans' b
+        updatePatGuard _                  = M.empty
+
+    rUnionMap f = foldl' (flip M.union) M.empty . fmap f
```
rhendric, line 304:
```suggestion
    rUnionMap f = foldl' (flip (M.union . f)) M.empty
```

src/Language/PureScript/Sugar/Names.hs
PureFunctor: Gotcha

rhendric (approved):

JordanMartinez:
> > Just a thought, but should we distinguish ByNullSourcePos as its own data constructor?
> 
> @JordanMartinez We could, but it also means duplicating code that's already caught by `BySourcePos`, and changing most of them would take some work.
> 
> > Also, on looking at this again, there's not a test that verifies that a given local binding has the expected BySourcePos. Could one be added?
> 
> Maybe we could check through the CoreFn? I also mentioned before that any incorrectness in the renamer is likely going to be caught during type checking because `QualifiedBy` is used as a key in the environment.

Makes sense. Thanks for explaining!

PR #4295, opened 2022-04-19
Add instructions for publishing prerelease to NPM

### LINKED ISSUE #4294: Publish prereleases (e.g. `v0.15.0-alpha-05`) to NPM, not just GitHub
## Summary

Whenever we make prereleases that include breaking changes, publish them on NPM via the `next` tag (e.g. `npm i purescript@next`).

## Motivation

Before we can make a breaking change release (e.g. `0.15.0`), we have to update at least the core libraries to compile on and and build tools to compile with the upcoming compiler version. So, we made prereleases by publishing them on GitHub but not on NPM (e.g. `v0.14.0-rc1`, `v0.15.0-alpha-05`, etc.). Each project's CI had to be updated to download this GitHub release. 

For libraries, this is no longer a problem. During this `0.15.0` release/migration work, we've switched to using [`setup-purescript`](https://github.com/purescript-contrib/setup-purescript) to get the prerelease. However, this tool only works because it was updated to do that during the `0.15.0` work.

For build tools, this is still a bit of a problem. While we could use `setup-purescript` to download the prerelease, it's more awkward to use. `pulp` and `spago` may be testing their code against multiple versions of the `purs` binary and against multiple OSes. So, using a custom download script or multiple `setup-purescript` steps rather than a build matrix where each build is testing one version is awkward. It would be easier if `npm i -g purescript@next` could be used instead.

Finally, publishing the prerelease on NPM would allow others to use it early and see if there are any bugs that haven't been reported yet. During the `0.15.0` work, a few bugs were found by people testing it out:
- #4274
- https://github.com/purescript-web/purescript-web-html/pull/75

## Proposal

Fixing this would require updating the `RELEASE_GUIDE.md` to further explain that one should publish a pre-release on NPM (besides the other updates we should do for that file). Specifically, the below line in the file...
> After all of the prebuilt binaries are present on the GitHub releases page, publish to npm: change to the npm-package directory and run npm publish. It's a good idea to check that the package can be installed from npm at this point.
should be updated to say something like, "When making a prerelease, change to the `npm-package` directory and run `npm publish --tag next`.

See https://manpages.org/npm-publish

JordanMartinez:
**Description of the change**

Fixes #4294. Pending final decision on whether to do this.

---

**Checklist:**


RELEASE_GUIDE.md
```diff
@@ -117,9 +117,15 @@ considering what effects this may have:
...
-  publish to npm: change to the `npm-package` directory and run `npm publish`.
-  It's a good idea to check that the package can be installed from npm at this
-  point.
+  publish to npm: change to the `npm-package` directory and do the following:
+    - if making a pre-release (e.g. `v0.15.0-alpha-05`)
+        - run `npm publish --tag next`
+        - verify that the prerelease can be installed via `npm i purescript@next`
+    - if making a normal release (e.g. `v0.15.0`)
+        - run `npm publish --tag next`
+        - run `npm publish`
```
rhendric, line 126: I suspect this won't work; I think `npm` prevents you from ever publishing the same version twice. You probably want

```suggestion
        - run `npm publish`
        - run `npm dist-tag add purescript@x.y.z next`
```

(maybe you can get away with replacing `x.y.z` with `latest`, I don't know)

JordanMartinez: I'm going to create a small npm package and see what happens when I run `npm publish --tag next` multiple times.

JordanMartinez: You're correct. Running `npm publish --tag next` then `npm publish` produces this error:
```
npm ERR! code E403
npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/trashme-purs-prerelease-test - You cannot publish over the previously published versions: 1.0.1-0.
npm ERR! 403 In most cases, you or one of your dependencies are requesting
npm ERR! 403 a package version that is forbidden by your security policy, or
npm ERR! 403 on a server you do not have access to.
```

JordanMartinez: So, after playing around with a toy package, here's what I've discovered.

First, running `npm publish` and then `npm dist-tag add package@v1.2.3 next` works (as Ryan stated above).

But because this is still manual, second, I explored how to fix a bad publish. For example, let's assume I ran
```
# wrong version: I meant `v1.2.0`
npm dist-tag add package@v1.1.0 next
# let's just fix it by updating `next` right?
npm dist-tag add package@v1.2.0 next
```
Let's assume the following:
- a package `bar`'s `next` tag refers to `v1.0.0`
- a package `foo` installed a package `bar` via `npm i bar@next`, so that `foo` depends on `bar@v1.0.0`.
- `npm version minor && npm publish --tag next` (so `bar@next` now points to `v1.1.0`)
- locally in `foo`, I run `npm i bar@next`, `v1.0.0` is installed, which is incorrect as it should be `v1.1.0` now.
- locally in `foo`, I nuke `node_modules`, `package-lock.json`, and remove `bar` from my dependencies. When I run `npm i bar@next`, `v1.0.0` is still installed.
- locally in `foo`, running `npm cache clear --force` followed by `npm i bar@next` now installs `bar@v1.1.0` in `foo`.

Does the "`next` tag sometimes points to the real `next` tag" issue above only happen locally because I'm publishing it on the same computer? Or would the same thing happen on other computers (e.g. in CI)?

f-f:
Given the consensus on #4294 should this include changes to CI as well for continuous deployment?

JordanMartinez:
If I can get an approval on this, I'll merge it and create a pre-release using the `alpha-05` release on `npm`. That will unblocked the `spago` PR, and then we can get a new version of `spago` out.

thomashoneyman (approved):

PR #4299, opened 2022-04-28
Make 0.15.0 release

JordanMartinez:
**Description of the change**

Makes the `0.15.0` release.

I wasn't sure if we should push a release candidate first before making a final `0.15.0` release. Let me know your thoughts.


JordanMartinez:
Tests failed because we're still using `SProxy`, etc. in tests rather than `Proxy`. I'm updating the tests locally.

JordanMartinez:
I've updated the tests to account for the `SProxy`, etc. issue. However, the new issue I'm getting is "Could not find module `Type.Data.Row`" in `typelevel-prelude`. That module was deleted in `v7.0.0`, but something else (possibly `psci-support`) expects it to exist.

JordanMartinez:
CI builds. Can I get an approval?

rhendric:
Should we be updating the tests in the same PR as we do the final release version bump?

JordanMartinez:
> Should we be updating the tests in the same PR as we do the final release version bump?

I'd be fine with updating the tests in a separate PR, so that this one just does the version bump.

JordanMartinez:
> I am about to be away from my desk for some time, but I'll try to follow this from my phone. I think this looks good, but I haven't approved because of the CI failure and discussion whether this should be split into two PRs (one to address the tests, one that only updates the tag).

Thanks Thomas! I think I can handle things from here once we get this release PR in. I might need an approval, but that's about it.

JordanMartinez:
Ubuntu CI failure is:
```
> spago@0.20.8 postinstall /tmp/tmp.DEfSQmi86L/node_modules/spago
  > node ./install.js
  
  internal/modules/cjs/loader.js:818
    throw err;
    ^
  
  Error: Cannot find module '/tmp/tmp.DEfSQmi86L/node_modules/spago/install.js'
      at Function.Module._resolveFilename (internal/modules/cjs/loader.js:815:15)
      at Function.Module._load (internal/modules/cjs/loader.js:667:27)
      at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
      at internal/main/run_main_module.js:17:47 {
    code: 'MODULE_NOT_FOUND',
    requireStack: []
  }
  npm WARN enoent ENOENT: no such file or directory, open '/tmp/tmp.DEfSQmi86L/package.json'
  npm WARN tmp.DEfSQmi86L No description
  npm WARN tmp.DEfSQmi86L No repository field.
  npm WARN tmp.DEfSQmi86L No README data
  npm WARN tmp.DEfSQmi86L No license field.
  
  npm ERR! code ELIFECYCLE
  npm ERR! errno 1
  npm ERR! spago@0.20.8 postinstall: `node ./install.js`
  npm ERR! Exit status 1
  npm ERR! 
  npm ERR! Failed at the spago@0.20.8 postinstall script.
  npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
  
  npm ERR! A complete log of this run can be found in:
  npm ERR!     /root/.npm/_logs/2022-04-28T18_24_29_619Z-debug.log
  Error: Process completed with exit code 1.
```
Thoughts?

JordanMartinez:
I ran the same steps locally as what gets run in CI and cannot reproduce the above issue.

JordanMartinez:
I think CI is failing due to the same issue that was fixed in https://github.com/purescript/purescript/pull/4276

It wasn't caught then because we only build the package set on PRs when the `CHANGELOG.d` dir doesn't contain files starting with `breaking_`

JordanMartinez:
@sd-yip Any idea as to why the above error (https://github.com/purescript/purescript/pull/4299#issuecomment-1112528546=) is occurring when we run `npm i spago@0.20.8`?

JordanMartinez:
There's two ways forward. I can either drop the packag set test, release `0.15.0`, and fix this later BEFORE we merge any new compiler PR. Or, we can fix it properly in a new PR, and then update this one to just do the version bump.

Unfortunately, I'm stumped on how to fix this. The root cause seems to be that the `./install.js` file doesn't exist, but I can't figure out how to reproduce the problem.

JordanMartinez:
Alternatively, we could just install Spago by downloading the binary ourselves and putting it on the PATH. That would bypass the problem itself.

rhendric:
Try appending `|| cat /root/.npm/_logs/*` to `which spago || npm install spago@0.20.8`?

sd-yip:
There is a working directory issue with the post-installation hook in the NPM version that the CI uses.

Upgrading NPM to the latest version seems to fix it.
```
npm i -g npm@8.8.0
```

JordanMartinez:
> There is a working directory issue with the post-installation hook in the NPM version that the CI uses.
> 
> Upgrading NPM to the latest version seems to fix it.
> 
> ```
> npm i -g npm@8.8.0
> ```

Thanks. I'm trying that out in #4302.

PR #4306, opened 2022-05-01
Deploy builds continuously to GitHub and npm

### LINKED ISSUE #4303: Continuous deployment
## Summary

From https://github.com/purescript/purescript/issues/4294#issuecomment-1102800906, compile and publish prereleases to GitHub and npm for every successful build of the master branch.

## Motivation

* Prereleases are a necessary part of the major release process, and making them manually drains time from our already busy release managers and delays progress on other parts of the ecosystem while they wait for approval. A continuous deployment process that uses the same publication channels as a manual release would eliminate these delays.
* Our release process is currently largely manual, exercised infrequently, and limited to a subset of core team members with the necessary permissions in GitHub, Hackage, and the npm registry. All of these factors make the process brittle, and a continuous deployment process would make incremental progress on all of them (except for Hackage; I don't plan to touch that one).
* As publishing to npm is one of the final steps in our release process, it isn't possible to catch some downstream release issues until after they have been published, at which point fixing those issues requires burning a version number. Having a compiler already available on npm with nearly identical code to the upcoming release would mean that most of those issues could be caught before the version bump commit is pushed.
* Users who are enthusiastic about testing upcoming compiler features will be able to do so more easily if built and published versions of the compiler including unreleased features are available at all times. The result would be a smoother transition process for them and more early feedback for us; a win-win.

With the above points in mind, the following proposal aims to:
* save our release managers time
* improve the health of our infrastructure and processes
* make user experimentation and early migration easier

If the above goals are in conflict with any aspect of the proposal, it is due to my ignorance, and the proposal should be amended.

## Proposal

At the highest level, this proposal amounts to:
* For every build, enforce consistency between the various files that store version numbers (fail the build if `purescript.cabal`, `app/Version.hs`, and `npm-package/package.json` disagree).
* For every push-to-branch build where the version number in the code matches a released version of PureScript:
  * Perform an appropriate version bump (in `purescript.cabal`, `app/Version.hs`, and both occurrences in `npm-package/package.json`) prior to the build
  * Assuming the build is otherwise successful...
  * Produce release bundles
  * Create a GitHub prerelease and tag with the bundles, but no release notes
  * Publish the npm package to the `next` tag, using a token that we add to GitHub's secrets storage for this repository
* Do not make any changes to bumping or publishing on PR builds or release builds (#4295 already covers manually updating the `next` tag after a full release)
* Do not make any automated publications of PureScript to Hackage

I assume that for full releases, we still want a human making an explicit version bump PR, having other humans approve it, and then manually creating a GitHub release for that version. This is the reason for the ‘matches a released version of PureScript’ clause; the intended effect is for a version bump commit to pause the continuous deployment process until a manual release is created. (People with experience actually doing releases: is this the right amount of conservative to save you time?)

### Version naming

‘Appropriate version bump’ merits some detailed consideration. PureScript exists at the uncomfortable junction between what npm expects from a version number (SemVer, [mostly](https://github.com/npm/cli/issues/4257)) and what Cabal expects from a version number (PVP, [mostly](https://github.com/haskell/cabal/issues/890)). Notably, the former allows prerelease suffixes in version numbers while the latter does not. As part of this proposal, I aim to resolve this mismatch and codify the resolution.

Full release version numbers shall use the MAJOR.MINOR.PATCH format, which is allowed by both npm and Cabal. Prerelease numbers shall use the MAJOR.MINOR.PATCH-[PHASE.]BUILD format, which is allowed by npm but not by Cabal. (PHASE stands for an optional `alpha`, `beta`, `rc`, or equivalent tag.) Consequently, `purescript.cabal` will always use MAJOR.MINOR.PATCH for PureScript's version, and any prerelease suffix will need to be bumped in `app/Version.hs`. None of these values should change in Git until it's time to cut a manual release or manually name a prerelease phase.

With that in mind, here's what the automated version bump should do at various points in compiler development:

|Committed version in purescript.cabal|Committed prerelease suffix in Version.hs|Committed version in package.json|Recent Git tags|Breaking changes in CHANGELOG.d|Version used in CD|
|---|---|---|---|---|---|
|0.15.0| — |0.15.0|0.15.0| |0.15.1-0|
|0.15.0| — |0.15.0|0.15.0, 0.15.1-0| |0.15.1-1|
|0.15.0| — |0.15.0|0.15.0, 0.15.1-1| |0.15.1-2|
| *version bump PR is merged* |
|0.15.1| — |0.15.1|0.15.0, 0.15.1-2| | — |
| *0.15.1 release is published* |
|0.15.1| — |0.15.1|0.15.1| |0.15.2-0|
|0.15.1| — |0.15.1|0.15.1, 0.15.2-0| |0.15.2-1|
|0.15.1| — |0.15.1|0.15.1, 0.15.2-1|breaking|0.16.0-0|
|0.15.1| — |0.15.1|0.15.1, 0.15.2-1, 0.16.0-0|breaking|0.16.0-1|
|0.15.1| — |0.15.1|0.15.1, 0.15.2-1, 0.16.0-1|breaking|0.16.0-2|
| *breaking change is reverted* |
|0.15.1| — |0.15.1|0.15.1, 0.15.2-1, 0.16.0-2| |0.15.2-2|
| *version bump PR is merged* |
|0.15.2| — |0.15.2|0.15.1, 0.16.0-2, 0.15.2-2| | — |
| *0.15.2 release is published* |
|0.15.2| — |0.15.2|0.16.0-2, 0.15.2| | 0.15.3-0 |
| *breaking change is re-added* |
|0.15.2| — |0.15.2|0.16.0-2, 0.15.2, 0.15.3-0|breaking|0.16.0-3|
|0.15.2| — |0.15.2|0.15.2, 0.16.0-3|breaking|0.16.0-4|
| *we decide to designate an alpha prerelease of 0.16 with a PR* |
|0.16.0|-alpha.0|0.16.0-alpha.0|0.15.2, 0.16.0-4|breaking|0.16.0-alpha.0|
|0.16.0|-alpha.0|0.16.0-alpha.0|0.15.2, 0.16.0-alpha.0|breaking|0.16.0-alpha.1|
| *version bump PR (really, a suffix-removing PR) is merged* |
|0.16.0| — |0.16.0|0.15.2, 0.16.0-alpha.1| | — |
| *0.16.0 release is published* |
|0.16.0| — |0.16.0|0.16.0| |0.16.1-0|

As this table suggests, the version bumping algorithm needs as inputs:
* the current version number committed to Git
* the git tags that have been used already
* whether any breaking changes have been merged

The proposed algorithm is as follows:
```
if current version contains a prerelease suffix
then
  if any Git tags match the current version except for the build number
  then
    take the largest such tag, increment its build number, and use that
  else
    use the current version as-is
else (current version does not contain a prerelease suffix)
  if current version appears in Git tags
  then
    if any breaking changes
    then
      increment minor number of current version (would be major if we were out of 0.x)
      set patch/prerelease to 0-0
    else
      increment patch number of current version
      set prerelease to -0
    if any Git tags match that version except for the build number
    then
      take the largest such tag, increment its build number, and use that
    else
      use that version as-is
  else (current version does not appear in Git tags)
    do not alter the code and do not publish an automated release
    (we are between the merging of a version bump PR and the corresponding manual release)
```

When the above algorithm produces a new version to use for a build, the MAJOR.MINOR.PATCH part of the number is used in `purescript.cabal`, the prerelease suffix is used in `app/Version.hs`, and both parts are used in `npm-package/package.json`.

Note the following features:
* Automatically-published builds always use a prerelease suffix.
* According to SemVer and npm, unnamed prerelease versions (0.16.0-2) sort before named prereleases (0.16.0-alpha.0), so we can use named prereleases as intermediate between early automated releases and full releases.
* The algorithm accounts for temporarily out-of-version-order prereleases, such as the 0.15.2-1 -> 0.16.0-0 -> ... -> 0.15.2-2 -> ... -> 0.16.0-3 sequence in the example. Early adopters might experience some churn this way but at least no version number will be reused.

### Infrastructure

To achieve this, someone with the appropriate credentials will need to generate an [npm automation token](https://docs.npmjs.com/creating-and-viewing-access-tokens), either for their personal account or for a new account that is made a PureScript collaborator in npm, and then add it to the [environment secrets](https://github.com/purescript/purescript/settings/secrets/actions) for this repository.

### Storage uncertainty

Storing a large number of artifact-laden prereleases in GitHub may run up against some sort of quota, now or in the future. (I have no indication of what that quota would be; it would just surprise me if we were given unbounded storage for large binary blobs.) We may need to extend this proposal with a plan to clean up automated prereleases that are older than some appropriate window. This would impact npm consumers as well, since the npm package is really just a thin shim that downloads binaries from GitHub.

rhendric:
**Description of the change**

Closes #4303. However, before this PR is merged, someone with more access than me needs to do the following:
- [x] Create a new npm account for automated deployment
  * what email address should we use?
  * how should the password to this account be managed?
- [x] Add that account as an owner of `purescript` in the npm registry
  * one of the existing owners will need to do this
- [x] Generate an automation token with that account
- [x] Add that token to a secret variable named `NPM_TOKEN`
  * requires someone with admin access ([possibly](https://github.com/github/docs/issues/1087) lower access levels could do this but not through the web interface)

---

**Checklist:**

- <s>Added a test for the contribution (if applicable)</s>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,19 @@ defaults:
     shell: "bash"
 
 env:
+  CI_PRERELEASE: "${{ github.event_name == 'push' }}"
   CI_RELEASE: "${{ github.event_name == 'release' }}"
   STACK_VERSION: "2.7.3"
+
+concurrency:
+  # We never want two releases or prereleases building at the same time, since
+  # they would likely both claim the same version number. Pull request builds
+  # can happen in parallel with anything else, since they don't mutate global
+  # state with a release. (GitHub Actions is either too cheap to give us `if`
+  # expressions or too lazy to document them, but we have untyped boolean
+  # operators to fall back on.)
+  group: "${{ github.event_name == 'pull_request' && github.run_id || 'continuous-deployment' }}"
+  cancel-in-progress: true
 
 jobs:
   build:
@@ -30,6 +41,10 @@ jobs:
 
     runs-on: "${{ matrix.os }}"
     container: "${{ matrix.image }}"
+
+    outputs:
+      do-not-prerelease: "${{ steps.build.outputs.do-not-prerelease }}"
+      version: "${{ steps.build.outputs.version }}"
 
     steps:
       - # We need a proper Git repository, but the checkout step will unpack a tarball instead of doing a clone
@@ -77,7 +92,8 @@ jobs:
           mkdir -p "$STACK_ROOT"
           echo "local-programs-path: $STACK_ROOT/programs" > $STACK_ROOT/config.yaml
 
-      - run: "ci/fix-home ci/build.sh"
+      - id: "build"
+        run: "ci/fix-home ci/build.sh"
 
       - name: "(Linux only) Build the entire package set"
         if: "${{ runner.os == 'Linux' }}"
@@ -106,8 +122,8 @@ jobs:
             exit 1
           fi
 
-      - name: "(Release only) Create bundle"
-        if: "${{ env.CI_RELEASE == 'true' }}"
+      - name: "(Release/prerelease only) Create bundle"
+        if: "${{ env.CI_RELEASE == 'true' || env.CI_PRERELEASE == 'true' && steps.build.outputs.do-not-prerelease != 'true' }}"
         run: |
           os_name="${{ runner.os }}"
           case "$os_name" in
@@ -123,6 +139,15 @@ jobs:
           esac
           cd sdist-test
           ../ci/fix-home bundle/build.sh "$bundle_os"
+
+      - name: "(Prerelease only) Upload bundle"
+        if: "${{ env.CI_PRERELEASE == 'true' && steps.build.outputs.do-not-prerelease != 'true' }}"
+        uses: "actions/upload-artifact@v3"
+        with:
+          name: "${{ runner.os }}-bundle"
+          path: |
+            sdist-test/bundle/*.sha
+            sdist-test/bundle/*.tar.gz
 
       - name: "(Release only) Publish bundle"
         if: "${{ env.CI_RELEASE == 'true' }}"
@@ -194,3 +219,32 @@ jobs:
       - run: "ci/fix-home stack --no-terminal --jobs=2 build --fast --test --no-run-tests --ghc-options -fwrite-ide-info"
 
       - run: "ci/fix-home stack exec weeder"
+
+  make-prerelease:
+    runs-on: "ubuntu-latest"
+    needs:
+      - "build"
+      - "lint"
+    if: "${{ github.event_name == 'push' && needs.build.outputs.do-not-prerelease != 'true' }}"
+    steps:
+      - uses: "actions/download-artifact@v3"
+      - uses: "ncipollo/release-action@v1.10.0"
+        with:
+          tag: "${{ needs.build.outputs.version }}"
+          artifacts: "*-bundle/*"
+          prerelease: true
+          body: "This is an automated preview release. Get the latest stable release [here](https://github.com/purescript/purescript/releases/latest)."
+      - uses: "actions/checkout@v3"
+      - uses: "actions/setup-node@v3"
+        with:
+          node-version: "16.x"
+      - name: "Publish npm package"
+        working-directory: "npm-package"
+        env:
+          BUILD_VERSION: "${{ needs.build.outputs.version }}"
+          NODE_AUTH_TOKEN: "${{ secrets.NPM_TOKEN }}"
+        run: |
+          src_version=$(node -pe 'require("./package.json").version')
+          npm version --allow-same-version "$BUILD_VERSION"
+          sed -i -e "s/--purs-ver=${src_version//./\\.}/--purs-ver=$BUILD_VERSION/" package.json
+          npm publish --tag next
--- a/CHANGELOG.d/internal_cd.md
+++ b/CHANGELOG.d/internal_cd.md
@@ -1,0 +1,1 @@
+* Deploy builds continuously to GitHub and npm
--- a/RELEASE_GUIDE.md
+++ b/RELEASE_GUIDE.md
@@ -72,14 +72,38 @@ considering what effects this may have:
 - JSON produced by `purs publish`
   - this might affect Pursuit
 
+## Making a release candidate
+
+- Make a commit bumping versions. The following should be updated:
+
+  - The `version` field in `purescript.cabal` should be set to the expected
+    final release version.
+
+  - The `prerelease` field in `app/Version.hs` should be set to `-rc.0`.
+
+  - The `version` field in `npm-package/package.json` should be set to the
+    concatenation of the above two items.
+
+  - The version to install in the `postinstall` script in `package.json` should
+    match the `version` field.
+
+- Upon merging the PR, the release candidate will be published to GitHub and
+  npm. There is no need to make a manual release for the RC. Subsequent builds
+  will be deployed to successive `-rc.*` numbers until a final release is made.
+
+- Verify that the release candidate can be installed via `npm i purescript@next`
+
 ## Making a release
+
+- Test that the last build published to `purescript@next` works in downstream
+  projects before starting the manual release process.
 
 - Make a commit bumping versions. The following should be updated:
 
   - The `version` field in `purescript.cabal`
 
-  - The `prerelease` field in `app/Version.hs`, if updating the prerelease
-    field
+  - The `prerelease` field in `app/Version.hs` should be cleared, if a
+    release candidate was previously published
 
   - The `version` field in `npm-package/package.json`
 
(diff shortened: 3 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -124,6 +140,15 @@ jobs:
...
           ../ci/fix-home bundle/build.sh "$bundle_os"
 
+      - name: "(Prerelease only) Upload bundle"
+        if: "${{ env.CI_PRERELEASE == 'true' && steps.build.outputs.do-not-prerelease != 'true' }}"
+        uses: "actions/upload-artifact@v3"
+        with:
+          name: "${{ runner.os }}-bundle"
+          path: |
+            sdist-test/bundle/*.sha
+            sdist-test/bundle/*.tar.gz
```
JordanMartinez, line 150: The [docs for `path`](https://github.com/ncipollo/release-action#action-inputs=) state:
> An optional set of paths representing artifacts to upload to the release. This may be a single path or a comma delimited list of paths (or globs)

Since I've not used this action before, will the above work? Or does it need to be
`path: "sdist-test/bundle/*.sha,sdist-test/bundle/*.tar.gaz"`

JordanMartinez (commented):
Aside from my two comments and the other NPM stuff, this LGTM.

.github/workflows/ci.yml
```diff
@@ -13,9 +13,20 @@ defaults:
...
   STACK_VERSION: "2.7.3"
 
+concurrency:
+  # We never want two releases or prereleases building at the same time, since
+  # they would likely both claim the same version number. Pull request builds
+  # can happen in parallel with anything else, since they don't mutate global
+  # state with a release. (GitHub Actions is either too cheap to give us `if`
+  # expressions or too lazy to document them, but we have untyped boolean
+  # operators to fall back on.)
+  group: "${{ github.event_name == 'pull_request' && github.run_id || 'continuous-deployment' }}"
```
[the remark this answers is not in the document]
rhendric: That comment about type conversion refers to equality comparisons, not boolean operators.

In practice, this pattern had the desired effect in testing; multiple pushes to master resulted in all but the last being canceled with the message `Canceling since a higher priority waiting request for 'continuous-deployment' exists`, but multiple PRs could still build concurrently.

.github/workflows/ci.yml
rhendric: Wrong docs! Those are the docs for `ncipollo/release-action`, which I used on line 231; the `artifacts` property on that action is a single glob. This action is `actions/upload-artifact`, and the multi-line format for `path` is used as an example [here](https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions).

.github/workflows/ci.yml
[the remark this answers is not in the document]
JordanMartinez: D'oh!

.github/workflows/ci.yml
JordanMartinez: Double D'oh!

JordanMartinez:
@thomashoneyman Could we create an email address under the `purescript.org` for this?

thomashoneyman:
What would you like it to be? npm@purescript.org?

JordanMartinez:
> What would you like it to be? [npm@purescript.org](mailto:npm@purescript.org)?

Sounds good to me!

JordanMartinez:
The email address has been set up AFAIK, but we need to determine to whom to forward the emails. I've volunteered since I've published to NPM in the past. @rhendric would you be interested in receiving such emails?

rhendric:
I don't have a problem with being on the list, and I feel there should be several of us on it, so sure.

JordanMartinez:
If no objections are raised against this PR and its implementation by tomorrow, let's merge it. It's been open for a while now, but some may have forgotten about it due to the lack of activity on it.

.github/workflows/ci.yml
```diff
@@ -13,9 +13,20 @@ defaults:
     shell: "bash"
 
 env:
+  CI_PRERELEASE: "${{ github.event_name == 'push' }}"
```
f-f, line 16: Would this prerelease for all the commits pushed to a PR? I imagine we'd want to do it only when merging things to `master`

rhendric: Updates to PRs trigger a `pull_request` event. The `push` event is just for pushes to branches properly in the repo—in this workflow, only the master branch triggers `push` events.

f-f: Great, thanks for the clarification!

This is a little better than I thought, but still if we push other branches (I would push a branch for stuff I'm working on) then we'd get an unwanted release published.

How about matching the name of the branch too?

rhendric: This workflow already only runs when the master branch triggers `push` events; that's line 5.

f-f: Oh right, it all makes sense now. Thank you!

f-f (approved):

thomashoneyman:
This is how Babel fixed it:
https://github.com/babel/babel/pull/9735/files

f-f:
We use this in Spago too: https://github.com/purescript/spago/blob/master/npm/.npmrc

JordanMartinez:
If we put that file in this repo, how does that affect things when we publish locally (i.e. not making releases via CI)? I'm guessing that file would either need to be removed before `npm publish` is called since `NPM_TOKEN` may not be set in that person's environment.

rhendric:
Yeah, so the `setup-node` GitHub action is supposed to take care of managing `.npmrc`, but it only does this [if](https://github.com/actions/setup-node/blob/bacd6b4b3ac3127b28a1e1920c23bf1c2cadbb85/src/main.ts#L42-L46) an explicit `registry-url` input is given (I used one in my testing but left it out in this PR). #4320 should fix this.

PR #4311, opened 2022-05-06
Fix issue with unnamed instances using type ops

### LINKED ISSUE #4310: Bundling error for /\ instances
Hello! This problem happens to me since upgrading to the new compiler version. I have reduced the code to the bare minimum that still prompts the issue.

## Description

Class instances for ```Data.Tuple.Nested./\``` fail at runtime (```is not a function``` if there are other instances, ```undefined``` if ```/\``` is the only instance)

## To Reproduce

Clone https://github.com/easafe/test and run ```spago test```

Note:

* The error only occur if the class instance is declared in a separated file 
* The error does not occur if the instance is named, or ```Tuple``` is used

## PureScript version

0.15.0

rhendric:
**Description of the change**

Fixes #4310.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/fix_4310.md
+++ b/CHANGELOG.d/fix_4310.md
@@ -1,0 +1,1 @@
+* Fix issue with unnamed instances using type operators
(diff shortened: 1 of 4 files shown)


natefaubion (approved):

natefaubion:
I do think it would be nice to generate names based on the desugared operator. Right now this is the only place then that still has potentially symbolic names. Ideally we could totally banish them.

rhendric:
I had the same thought, but right now we generate those names during the CST to AST transformation, so the ratio of work between the two approaches overwhelmingly favors this one. Might make sense as a thing to do in a future refactoring, though.

JordanMartinez (approved):

PR #4317, opened 2022-05-12
Update Prim docs for type-level changes; clean up other docs

### LINKED ISSUE #4316: Update `Prim (Int, Boolean)` to mention type-level value
## Description

See https://github.com/purescript/purescript/issues/4314#issuecomment-1123751760

Docs for `Prim.Int` are below.

``````markdown
A 32-bit signed integer. See the `purescript-integers` package for details of how this is
accomplished when compiling to JavaScript.

Construct values of this type with literals:
```purs
x = 23 :: Int
```
``````

These should also note that type-level integers exist as well. 

The same needs to be done for `Boolean`.

## PureScript version

0.15.0

### PR DESCRIPTION
**Pull request by JordanMartinez**
**Description of the change**

Fixes #4316 in the first commit. While looking through our docs for a few other types, I thought some of those could be fleshed out with more details. If that's not desired, I can revert the commit.

---

**Checklist:**

### CODE REVIEW

File: src/Language/PureScript/Docs/Prim.hs [hunk 1]
```diff
@@ -258,12 +258,25 @@ kindSymbol = primType "Symbol" $ T.unlines
...
-  , "For example, the kind of `Record` is `Row Type -> Type`, mapping field names to values."
+  , "For example..."
+  , ""
+  , "     type ClosedRow = ( label :: TypeName, name :: String )"
+  , "     type OpenRow otherRows = ( label :: TypeName, name :: String | otherRows )"
+  , "     type FinalRow = OpenRows ( anotherLabel :: Int )"
+  , "     -- FinalRow = ( label :: TypeName, name :: String, anotherLabel :: Int )"
+  , ""
+  , "Rows can be 'open' or 'closed'. 'Closed' rows cannot be extended with additional label-type pairs."
+  , "'Open' rows can be extended."
```
rhendric, line 278: I object to this way of explaining rows. An ‘open row’ isn't a type in PureScript's type system; all row types are closed, and ‘open rows’ are just a particular pattern for using type synonyms to create closed rows. Someone recently [got tripped up over this](https://discourse.purescript.org/t/partially-applied-type-in-proxy-with-row-type-param/2996), thinking that an open row was something they could pass around in a `Proxy`, which a reasonable thing to think if we say that rows are types and open rows are rows.

JordanMartinez: Honestly, I didn't know this either. I'm now wondering why I never thought of passing an open row around via a Proxy...

JordanMartinez: I've pushed a commit that drops this.

On another hand, row kinds are technically `label-type` pairs, which means one can have a row of symbols:
```
type Foo = ( label :: "label", int :: 4 )
```
Should we include this there? Or leave it out because its too advanced and is better documented elsewhere?

rhendric: One actually can't do that particular thing! Rows need to be kind-homogeneous; you can't use a `Symbol` with one label and an `Int` with another in the same row—the former would create a `Row Symbol` and the latter a `Row Int`.

Some slight rewording of the explanation of the kind of `Row` might clear the above up, and I think *that* would be more useful than a simple demonstration of the technique (which doesn't have obvious applications for most, I would say).

JordanMartinez: :man_facepalming: Ugh. I actually know this, but wasn't thinking.

> Some slight rewording of the explanation of the kind of Row might clear the above up, and I think that would be more useful than a simple demonstration of the technique (which doesn't have obvious applications for most, I would say).

Feel free to propose an improved rendering.

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -258,12 +258,20 @@ kindSymbol = primType "Symbol" $ T.unlines
...
 
 kindRow :: Declaration
 kindRow = primType "Row" $ T.unlines
   [ "`Row` is the kind constructor of label-indexed types which map type-level strings to other types."
-  , "For example, the kind of `Record` is `Row Type -> Type`, mapping field names to values."
+  , "For example..."
+  , ""
+  , "     type ExampleRow = ( label :: TypeName, name :: String )"
+  , ""
+  , "`Row`s are often used in `Record`, whose kind signature is `Row Type -> Type`, mapping field names to values."
```
rhendric, line 274:
```suggestion
  , "The most common use of `Row` is `Row Type`, a row mapping labels to basic (of kind `Type`) types:"
  , ""
  , "    type ExampleRow :: Row Type"
  , "    type ExampleRow = ( name :: String, values :: Array Int )"
  , ""
  , "This is the kind of `Row` expected by the `Record` type constructor."
  , "More advanced row kinds like `Row (Type -> Type)` are used much less frequently."
```

JordanMartinez: I've pushed this change. Can I get an approval, so we can merge this?

CHANGELOG.d/bug_update-prim-docs.md
```diff
@@ -0,0 +1 @@
+* Update `Prim` docs for Boolean, Int, String/Symbol, Number, Record, and Row
```
rhendric, line 1: This file should start with `fix_`

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -258,12 +258,22 @@ kindSymbol = primType "Symbol" $ T.unlines
   , ""
   , "Construct types of this kind using the same literal syntax as documented"
   , "for strings."
+  , ""
+  , "    type Hello :: Symbol {- String -}"
```
rhendric, line 262: Did you mean to leave the `{- String -}` comment in here? I'm not sure that's helpful; it almost reads like you're saying the two are interchangeable.

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -322,20 +332,46 @@ number :: Declaration
 number = primType "Number" $ T.unlines
   [ "A double precision floating point number (IEEE 754)."
   , ""
-  , "Construct values of this type with literals:"
+  , "Construct values of this type with literals."
+  , "Negative literals must be wrapped in parenthesis if the negation sign could be mistaken"
```
rhendric, line 336:
```suggestion
  , "Negative literals must be wrapped in parentheses if the negation sign could be mistaken"
```

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -322,20 +332,46 @@ number :: Declaration
...
+  , "for an infix operator (e.g. `negate`):"
```
rhendric, line 337: What does the ``(e.g. `negate`)`` mean here?

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -322,20 +332,46 @@ number :: Declaration
...
   , ""
-  , "    y = 35.23 :: Number"
-  , "    z = 1.224e6 :: Number"
+  , "    x = 35.23 :: Number"
+  , "    y = 1.224e6 :: Number"
+  , "    z = (-1.0) :: Number"
```
rhendric, line 341: This isn't actually an example where parentheses are required.
```suggestion
  , "    y = -1.224e6 :: Number"
  , "    z = exp (-1.0) :: Number"
```

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -322,20 +332,46 @@ number :: Declaration
...
 
 int :: Declaration
 int = primType "Int" $ T.unlines
-  [ "A 32-bit signed integer. See the purescript-integers package for details"
+  [ "A 32-bit signed integer. See the `purescript-integers` package for details"
   , "of how this is accomplished when compiling to JavaScript."
   , ""
-  , "Construct values of this type with literals:"
+  , "Construct values of this type with literals. Hexadecimal syntax is supported."
+  , "Negative literals must be wrapped in parenthesis if the negation sign could be mistaken"
```
rhendric, line 350:
```suggestion
  , "Negative literals must be wrapped in parentheses if the negation sign could be mistaken"
```

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -322,20 +332,46 @@ number :: Declaration
...
+  , "for an infix operator (e.g. `negate`):"
```
rhendric, line 351: Ditto?

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -322,20 +332,46 @@ number :: Declaration
...
   , ""
   , "    x = 23 :: Int"
+  , "    y = 0x17 :: Int"
+  , "    z = (-23) :: Int"
```
rhendric, line 355:
```suggestion
  , "    x = -23 :: Int"
  , "    y = 0x17 :: Int"
  , "    z = complement (-24) :: Int"
```

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -322,20 +332,46 @@ number :: Declaration
...
+  , ""
+  , "Integers used as types are considered to have kind `Int`."
+  , "Unlike value-level `Int`s, which must be representable as a 32-bit signed integer,"
+  , "type-level `Int`s are unbounded. Hexadecimal support is also supported at the type-level."
```
rhendric, line 359:
```suggestion
  , "type-level `Int`s are unbounded. Hexadecimal support is also supported at the type level."
```

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -348,26 +384,37 @@ string = primType "String" $ T.unlines
...
   , "    x = \"hello, world\" :: String"
   , ""
-  , "Multi-line string literals are also supported with triple quotes (`\"\"\"`)."
+  , "Multi-line string literals are also supported with triple quotes (`\"\"\"`):"
+  , ""
+  , "    x = \"\"\"multi"
+  , "       line\"\"\""
+  , ""
+  , "At the type-level, string literals are kinded with the `Symbol` type."
+  , "They will be kinded with the `String` type in a future release:"
```
rhendric, line 393:
```suggestion
  , "At the type level, string literals represent types with kind `Symbol`."
  , "These types will have kind `String` in a future release:"
```

src/Language/PureScript/Docs/Prim.hs
```diff
@@ -348,26 +384,37 @@ string = primType "String" $ T.unlines
...
+  , "    x = 'a' :: Char"
+  ]
 
 boolean :: Declaration
 boolean = primType "Boolean" $ T.unlines
   [ "A JavaScript Boolean value."
   , ""
   , "Construct values of this type with the literals `true` and `false`."
+  , ""
+  , "The `True` and `False` types defined in `Prim.Boolean` are kinded with this type."
```
rhendric, line 417:
```suggestion
  , "The `True` and `False` types defined in `Prim.Boolean` have this type as their kind."
```

rhendric (changes requested):
Mostly minor copy-editing stuff; more broadly, I don't know how much of this should be in the compiler versus the documentation repo, like what we do with [Prim.TypeError](https://pursuit.purescript.org/builtins/docs/Prim.TypeError), but I don't think this is excessive. I just think it would be good to have one official place that does a reasonably complete job of documenting the language, so to the extent that the things being added here don't exist there, I hope we at least file an issue there to add them.

src/Language/PureScript/Docs/Prim.hs
JordanMartinez: `negate` is in reference to ~negate~ [`sub`/`-`](https://pursuit.purescript.org/packages/purescript-prelude/6.0.0/docs/Data.Ring#v:(-)). I think I'll just drop it.

src/Language/PureScript/Docs/Prim.hs
JordanMartinez: Oh no... :scream: How long have I been misspelling this!?

JordanMartinez:
K, I think this is good now.

rhendric (approved):

PR #4322, opened 2022-05-13
Fix bad interaction between module renaming and inliner

### LINKED ISSUE #4229: genericShow compiles to invalid Javascript for records when a constructor in the same module is named Prim
## Description

A module that has a data definition with a constructor named `Prim` and a record in a type that uses `genericShow` will throw `ReferenceError: Prim_1 is not defined` at runtime.

## To Reproduce

Start a new project with `spago init`. Replace `src/Main.purs` with the following:
```
module Main where

import Prelude

import Data.Generic.Rep (class Generic)
import Data.Show.Generic (genericShow)
import Effect (Effect)
import Effect.Console (log)

data X a = Prim { y :: a }

derive instance Generic (X a) _
instance Show a => Show (X a) where
  show x = genericShow x

main :: Effect Unit
main = do
  log $ show $ Prim { y : 1 }
```

Run the program with `spago run`.

On my machine, this created the following output:
```
[info] Build succeeded.
C:\Users\colin\Documents\Projects\purescript-show-error\output\Main\index.js:25
            return Data_Show_Generic.genericShow(genericX_)(Data_Show_Generic.genericShowConstructor(Data_Show_Generic.genericShowArgsArgument(Data_Show.showRecord(Prim_1["undefined"])(Data_Show.showRecordFieldsCons({
                                                                                                                             
                                       ^

ReferenceError: Prim_1 is not defined
    at show (C:\Users\colin\Documents\Projects\purescript-show-error\output\Main\index.js:25:165)
    at Object.<anonymous> (C:\Users\colin\Documents\Projects\purescript-show-error\output\Main\index.js:37:71)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (C:\Users\colin\Documents\Projects\purescript-show-error\.spago\run.js:3:1)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
[error] Running failed; exit code: 1
```

## Expected behavior

The program should output `(Prim { y: 1 })` to the console.

## Additional context

Running on Windows 11, haven't checked on other platforms.

## PureScript version

0.14.5

rhendric:
This bug was triggered when modules that the compiler handles specially
are shadowed by local constructors. For example, a constructor named
`Prim` could have caused references to `Prim_1["undefined"]` to be
produced in the compiled code, leading to a reference error at run time.
Less severely, a constructor named `Control_Bind` would have caused the
compiler not to inline known monadic functions, leading to slower and
less readable compiled code.

**Description of the change**

Fixes #4229 by moving module names into the CoreImp AST and performing module renaming after inlining and optimization. This also allowed me to clean up the bifurcation of `Underscore_Module_Names` and `Dotted.Module.Names` in our Prelude constants, as a bonus.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/CodeGen/JS.hs
+++ b/src/Language/PureScript/CodeGen/JS.hs
@@ -59,23 +59,20 @@ moduleToJs
 moduleToJs (Module _ coms mn _ imps exps reExps foreigns decls) foreignInclude =
   rethrow (addHint (ErrorInModule mn)) $ do
     let usedNames = concatMap getNames decls
-    let mnLookup = renameImports usedNames imps
-    let decls' = renameModules mnLookup decls
-    (jsDecls, Any needRuntimeLazy) <- runWriterT $ mapM (moduleBindToJs mn) decls'
-    let mnReverseLookup = M.fromList $ map (\(origName, (_, safeName)) -> (moduleNameToJs safeName, origName)) $ M.toList mnLookup
-    let moduleObjectNames = "$foreign" `S.insert` M.keysSet mnReverseLookup
-    optimized <- traverse (traverse (fmap (annotatePure moduleObjectNames) . optimize)) (if needRuntimeLazy then [runtimeLazy] : jsDecls else jsDecls)
-    let usedModuleNames = foldMap (foldMap (findModules mnReverseLookup)) optimized
-          `S.union` M.keysSet reExps
-    let jsImports
-          = map (importToJs mnLookup)
-          . filter (flip S.member usedModuleNames)
-          . (\\ (mn : C.primModules)) $ ordNub $ map snd imps
+    let imps' = ordNub $ map snd imps
+    let mnLookup = renameImports usedNames imps'
+    (jsDecls, Any needRuntimeLazy) <- runWriterT $ mapM (moduleBindToJs mn) decls
+    optimized <- traverse (traverse (fmap annotatePure . optimize)) (if needRuntimeLazy then [runtimeLazy] : jsDecls else jsDecls)
     F.traverse_ (F.traverse_ checkIntegers) optimized
     comments <- not <$> asks optionsNoComments
     let header = if comments then coms else []
-    let foreign' = maybe [] (pure . AST.Import "$foreign") $ if null foreigns then Nothing else foreignInclude
+    let foreign' = maybe [] (pure . AST.Import FFINamespace) $ if null foreigns then Nothing else foreignInclude
     let moduleBody = concat optimized
+    let (S.union (M.keysSet reExps) -> usedModuleNames, renamedModuleBody) = traverse (replaceModuleAccessors mnLookup) moduleBody
+    let jsImports
+          = map (importToJs mnLookup)
+          . filter (flip S.member usedModuleNames)
+          $ (\\ (mn : C.primModules)) imps'
     let foreignExps = exps `intersect` foreigns
     let standardExps = exps \\ foreignExps
     let reExps' = M.toList (M.withoutKeys reExps (S.fromList C.primModules))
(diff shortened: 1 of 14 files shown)


JordanMartinez:
<details>
<summary>So, the old code generally worked like this, right?</summary>

1. get declarations of module

```purs
module Mod where

import Q (two)

data Q = Q

foo :: Q
foo = 1 + (Q).two -- two refers to Q module
```

2. rename imports to prevent clash with mod names

```purs
module Mod where

import Q1 (two)

data Q = Q

foo :: Q
foo = 1 + (Q).two
```

3. Rename qual values in decls to use updated imports

module Qual where

import Q1 (two)

data Q = Q

foo :: Q
foo = 1 + (Q1).two


4. Generate a safe name -> original name map

```
Q1 -> Q
```

5. Add '$foreign' to that map's key set

```
Q1
$foreign
```

6. optimize declarations

7. calculate what are all the used module names

`if Q is used then [Q1] else [] <> "$foreign" <> reExportsArray`

8. Deduplicate imports

9. Remove this module and Prim modules

10. Keep only imports that are members of used module names (i.e. don't import what isn't used)

11. Map the resulting modules to their JS value

</details>

<details>
<summary>And the new code works like this?</summary>

1. Deduplicate the imports (small perf benefit)
2. Rename the imports
3. optimize the declarations
4. Replace optimized declarations' module references to renamed ones
5. Calculate the used modules
6. Remove this module and Prim modules from used modules, keep only those that were used, and map them to their corresponding JS
</details>

Should you be adding "$foreign" to the [keyset used](https://github.com/purescript/purescript/pull/4322/files#diff-cb52c59a38e5742bbe2204955c4d8bb67b8d36e1872cab336dafa8162090cdb1R71)? Or was that unnecessary in the first place? And for confirmation, could you add a test that includes FFI?

rhendric:
Yeah, mostly.

Old code:
1. Generate an old-name-to-new-name map for every import
2. Use that map to rename qualified references in decls
3. Translate decls to JS
4. Reverse the previous map to get a new-name-to-old-name map, and consequently a set of new names
5. After optimizing, add pure annotations to the JS decls, which requires knowing the (new) names used for modules in the JS
    1. For this purpose, understand `$foreign` as one of those names
6. Collect all the module names that are still used after optimization, using the new-name-to-old-name map to back out the original module names, and union the result with any reexported modules
7. Generate JS imports, excluding any primitive modules and any modules not appearing in the previous result, with a deduplication step in here for some reason

New code:
1. Do that deduplication early, assuming it's there for a good reason, and strip out the `Ann`s we never use
2. Generate an old-name-to-new-name map for every import
3. Translate decls to JS, as-is
4. After optimizing, add pure annotations to the JS decls, which no longer requires knowing anything because the JS AST has a new constructor for module accessors
5. Collect all the module names that are still used after optimization (and in the process rewrite the module accessors that used them to be regular JS indexers, using the old-name-to-new-name map), and union the result with any reexported modules
6. Generate JS imports, excluding any primitive modules and any modules not appearing in the previous result

Good call on the FFI test; I was missing a small optimization that I can test.

src/Language/PureScript/CodeGen/JS.hs
```diff
@@ -113,19 +110,18 @@ moduleToJs (Module _ coms mn _ imps exps reExps foreigns decls) foreignInclude =
...
       AST.App ss f args -> (if alreadyAnnotated then AST.App else pureApp) ss <$> maybePure' f <*> traverse maybePure args
-      -- In general, indexers can be effectful, but not when indexing into an
-      -- ES module object.
-      AST.Indexer ss idx v@(AST.Var _ name)
-        | name `S.member` moduleObjectNames -> (\idx' -> AST.Indexer ss idx' v) <$> maybePure idx
       AST.ArrayLiteral ss jss -> AST.ArrayLiteral ss <$> traverse maybePure jss
       AST.ObjectLiteral ss props -> AST.ObjectLiteral ss <$> traverse (traverse maybePure) props
       AST.Comment c js -> AST.Comment c <$> maybePure js
 
+      js@(AST.Indexer _ _ (AST.Var _ "$foreign")) -> Just js
```
JordanMartinez, line 117: Could the text `$foreign` be defined as a constant somewhere, so that a change in one place doesn't break this somewhere else?

JordanMartinez (approved):
Besides the `$foreign` comment, LGTM.

src/Language/PureScript/CodeGen/JS.hs
rhendric: Done.

PR #4326, opened 2022-05-15
Update to GHC 9.2.2

hdgarrood:
This PR updates us to GHC 9.2.2. Most of the updates are fairly straightforward; I will explain why I've made them as inline review comments.

The main reasons I think we might want to do this are:
- to benefit from numerous performance enhancements in GHC in the 9.2 series (and the upcoming 9.4 series as well); see [the release notes](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/release-notes.html). I think these performance enhancements will mean that GHC can compile the PureScript compiler more quickly, but I don't think we would expect to see a speedup in the performance of the PureScript compiler itself.
- to be able to provide ARM binaries for systems like M1 macs: as of version 9.2.1, GHC has a native ARM code generator.

Reasons not to merge this yet:
- We depend on `protolude-0.3.1` which is tagged but not yet uploaded to Hackage so for now I'm using a git reference in `stack.yaml`; this means that people would probably struggle to compile the PureScript compiler with cabal if we made a release with these changes. We should probably either drop the `protolude` dependency or wait for 0.3.1 to be uploaded to Hackage.
- Weeder does not yet compile on GHC 9.2
- I should probably figure out what the deal with this warning is. I'm fairly sure it's a false positive as we aren't actually using the GHC API anywhere (AFAIK???)
  ```
  Warning:
      This package indirectly depends on multiple versions of the same package. This is very likely to cause a compile failure.
        package typed-process (typed-process-0.2.8.0-qJ7FQLjAUsDoh5UIGjmGT) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package streaming-commons (streaming-commons-0.2.2.4-8nDHmhtmpx83qtrdduE76B) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package sourcemap (sourcemap-0.1.7-9a6OzM8XbZDSeQz7MTa1o) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package purescript (purescript-0.15.1) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package purescript (purescript-0.15.1) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package purescript (purescript-0.15.1) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package optparse-applicative (optparse-applicative-0.17.0.0-G8nAkP3cjJUHOGraopmgpx) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package haskeline (haskeline-0.8.2-27iD1wLo6nE3EGXBbzNrZN) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package gitrev (gitrev-1.3.1-5EZyLHF9iyU5HR4ImcQtFa) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package conduit-extra (conduit-extra-1.3.6-IWwDkCT2NsHEPOAoybwv54) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package Cabal (Cabal-3.6.3.0-FskfNehIkKSJXUcrlACH2g) requires process-1.6.13.1-1ayluMYLS1yTV7wv3m4fy
        package ghc (ghc-9.2.2) requires process-1.6.13.2
  ```


.github/workflows/ci.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         include:
           - # If upgrading the Haskell image, also upgrade it in the lint job below
             os: "ubuntu-latest"
-            image: "haskell:8.10.7-stretch@sha256:100f8fb7d7d8d64adb5e106fe8136b8d4cbdc03aeb2cbd145a7597d74b69bafb"
+            image: "haskell:9.2.2-buster@sha256:c2f489a95be8dd0b253da6980420a6e9c02e81c2d14838ca774f86f5b5db91f4"
```
hdgarrood, line 38: `buster` (Debian 10) is the next Debian release after `stretch`. There is no `stretch` image which includes GHC 9.2.2, probably because `stretch` will stop receiving security updates next month (June 2022).

app/Version.hs
```diff
@@ -26,7 +26,8 @@ versionString = showVersion Paths.version ++ prerelease ++ extra
   extra = ""
 #else
   extra = " [development build; commit: " ++ $(GitRev.gitHash) ++ dirty ++ "]"
-  dirty
-    | $(GitRev.gitDirty) = " DIRTY"
-    | otherwise = ""
+  dirty =
+    if $(GitRev.gitDirty)
```
hdgarrood, line 30: Before this change, GHC would complain that the `otherwise` case was redundant if `GitRev.gitDirty` evaluated to True.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -822,9 +822,9 @@ newtype AssocList k t = AssocList { runAssocList :: [(k, t)] }
   deriving (Show, Eq, Ord, Foldable, Functor, Traversable)
 
 $(deriveJSON (defaultOptions { sumEncoding = ObjectWithSingleField }) ''NameSource)
+$(deriveJSON (defaultOptions { sumEncoding = ObjectWithSingleField }) ''ExportSource)
```
hdgarrood, line 825: This (and a few other JSON instance rearrangements) were necessary because any instances defined below a TH `deriveJSON` call like this one won't be visible to that call, so if a data type `A` references another data type `B` in its definition, then the `deriveJSON` for `B` has to come before the one for `A`.

src/Language/PureScript/CST/Utils.hs
```diff
@@ -38,11 +38,14 @@ qualifiedProperName n = QualifiedProperName (N.coerceProperName <$> n)
...
 -- (via a call to getProperName).
 newtype ProperName =
-  ProperName { getProperName :: forall a. Name (N.ProperName a) }
+  ProperName { _getProperName :: forall a. Name (N.ProperName a) }
 
 properName :: Name (N.ProperName a) -> ProperName
 properName n = ProperName (N.coerceProperName <$> n)
 
+getProperName :: forall a. ProperName -> Name (N.ProperName a)
+getProperName pn = _getProperName pn -- eta expansion needed here due to simplified subsumption
```
hdgarrood, line 47: Without this change, you get an error like:
```
src\Language\PureScript\CST\Utils.hs:47:17: error:
    * Couldn't match type: forall (a1 :: N.ProperNameType).
                           Name (N.ProperName a1)
                     with: Name (N.ProperName a)
      Expected: ProperName -> Name (N.ProperName a)
        Actual: ProperName
                -> forall (a :: N.ProperNameType). Name (N.ProperName a)
    * In the expression: _getProperName
      In an equation for `getProperName': getProperName = _getProperName
    * Relevant bindings include
        getProperName :: ProperName -> Name (N.ProperName a)
          (bound at src\Language\PureScript\CST\Utils.hs:47:1)
   |
47 | getProperName = _getProperName
   |                 ^^^^^^^^^^^^^^
```

src/Language/PureScript/CoreFn/ToJSON.hs
```diff
@@ -27,64 +30,68 @@ constructorTypeToJSON :: ConstructorType -> Value
 constructorTypeToJSON ProductType = toJSON "ProductType"
 constructorTypeToJSON SumType = toJSON "SumType"
 
+infixr 8 .=
+(.=) :: ToJSON a => String -> a -> Pair
```
hdgarrood, line 34: Almost everything in this module is due to `aeson-2` changes, specifically the left hand side of `.=` must now be a `Key` rather than a `Text`. I thought it might be easier to define our own version of `.=` so that we don't have to convert the key to a different type each time we call it.

src/Language/PureScript/CoreImp/Optimizer/Unused.hs
```diff
@@ -14,8 +14,11 @@ removeCodeAfterReturnStatements :: AST -> AST
 removeCodeAfterReturnStatements = everywhere (removeFromBlock go)
   where
   go :: [AST] -> [AST]
-  go jss | not (any isReturn jss) = jss
-         | otherwise = let (body, ret : _) = break isReturn jss in body ++ [ret]
+  go jss =
+    case break isReturn jss of
```
hdgarrood, line 18: This change was necessitated by the fact that `-Wincomplete-uni-patterns` was added to `-Wall` at some point between 8.10.7 and 9.2.2; the `let (body, ret : _) = ...` is an incomplete pattern because it does not account for what happens if the second element of the tuple is an empty list. Thankfully we can avoid it (as well as an unnecessary traversal of the list) by rewriting the function to use `break` instead.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -256,13 +256,14 @@ addExplicitImport' decl moduleName qualifier imports =
...
+-- function 'update'.
 updateAtFirstOrPrepend :: (a -> Bool) -> (a -> a) -> a -> [a] -> [a]
-updateAtFirstOrPrepend p t d l =
-  case findIndex p l of
-    Nothing -> d : l
-    Just i ->
-      let (x, a : y) = splitAt i l
-      in x ++ [t a] ++ y
+updateAtFirstOrPrepend predicate update def xs =
+  case break predicate xs of
```
hdgarrood, line 264: Another case where using `break` solves both a `-Wincomplete-uni-patterns` warning and also removes an unnecessary traversal of the list at the same time. I thought I'd take the opportunity to use clearer argument names here too.

src/Language/PureScript/Sugar/Names/Env.hs
```diff
@@ -488,5 +489,8 @@ checkImportConflicts ss currentModule toName xs =
         return (mnNew, mnOrig)
       _ -> throwError . errorMessage' ss $ ScopeConflict name conflictModules
     else
-      let ImportRecord (Qualified (Just mnNew) _) mnOrig _ _ = head byOrig
-      in return (mnNew, mnOrig)
+      case head byOrig of
+        ImportRecord (Qualified (Just mnNew) _) mnOrig _ _ ->
```
hdgarrood, line 493: Another `-Wincomplete-uni-patterns` warning

src/Language/PureScript/TypeChecker.hs
```diff
@@ -584,13 +584,17 @@ typeCheckAll moduleName = traverse go
       , ..
       }
 
+-- | Check that a newtype has just one data constructor with just one field, or
+-- throw an error. If the newtype is valid, this function returns the single
+-- data constructor declaration and the single field, as a 'proof' that the
+-- newtype was indeed a valid newtype.
```
hdgarrood, line 590: Returning the decl and the field here means that we avoid an unsafe pattern match in `Sugar/TypeClasses/Deriving`.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -262,15 +262,19 @@ entails SolverOptions{..} constraint context hints =
...
-                                -- Make sure the type unifies with the type in the type instance definition
-                                case matches typeClassDependencies tcd tys'' of
-                                  Apart        -> Right ()                   -- keep searching
-                                  Match substs -> Left (Right (substs, tcd)) -- found a match
-                                  Unknown ->
-                                    if null (tcdChain tcd) || null tl
-                                    then Right ()                                   -- need proof of apartness but this is either not in a chain or at the end
-                                    else Left (Left (tcdToInstanceDescription tcd)) -- can't continue with this chain yet, need proof of apartness
+                  let found = for (init $ tails chain) $ \tail' ->
+                                case tail' of
```
hdgarrood, line 266: This is another `-Wincomplete-uni-patterns` one but I wonder if it might be nicer to use the NonEmpty version of `groupBy` above, so that we wouldn't need a branch that calls `internalError` here.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -258,11 +259,12 @@ entails SolverOptions{..} constraint context hints =
...
 
             let (catMaybes -> ambiguous, instances) = partitionEithers $ do
-                  chain <- groupBy ((==) `on` tcdChain) $
+                  chain :: NonEmpty TypeClassDict <-
+                           NEL.groupBy ((==) `on` tcdChain) $
                            sortOn (tcdChain &&& tcdIndex)
                            dicts
                   -- process instances in a chain in index order
-                  let found = for (init $ tails chain) $ \(tcd:tl) ->
+                  let found = for (tails1 chain) $ \(tcd :| tl) ->
```
hdgarrood, line 267: Another `-Wincomplete-uni-patterns` warning. This was safe before, but now the safety is tracked in the types.

tests/TestGraph.hs
```diff
@@ -20,13 +17,12 @@ spec = do
...
 
-    graphFixture <- readUTF8FileT (baseDir <> graphFixtureName)
+    Just graphFixture <- Json.decodeFileStrict' (baseDir <> graphFixtureName)
     eitherGraph <- fst <$> P.graph modulePaths
     case eitherGraph of
       Left err -> error $ "Graph creation failed. Errors: " <> show err
       Right res ->
-        let textRes = Text.decodeUtf8 $ ByteString.toStrict $ Json.encode res
-        in graphFixture `shouldBe` textRes
+        res `shouldBe` graphFixture
```
hdgarrood, line 25: This change means that we don't have to update the fixture if the order of fields changes, and we don't get errors if the fields end up in different orders due to different build configurations. Updating aeson seems to have caused the fields in the result to end up in a different order.

hdgarrood: Also, you're supposed to put the expected value second. The error message describes the 'expected' vs 'actual' value, so the error can be misleading if you put the expected value first; this is why I've flipped them around.

.github/workflows/ci.yml
rhendric: Looks like this would mean bumping our glibc dependency from 2.24 to 2.28.

Should we make a Discourse post asking if people would object to that? Seems like we consistently stir up trouble whenever this changes.

hdgarrood: Yes, good point. This certainly isn’t ideal. Maybe we should just take a basic Debian stretch image and install the Haskell tools ourselves? My inclination is to avoid bumping this if we can manage it.

hdgarrood:
Seems like CI time for this branch is hovering around 14-15 mins, which might be slightly faster than it is currently (usually 15-18 mins from the looks of things), but the variance seems quite high so it's hard to tell.

hdgarrood:
I've created #4336, which covers a part of this (the new warnings) that can be merged now. As for the rest, I am inclined to wait until the upstream dependencies (weeder and protolude) are ready, so I think I'll close this for now.

JordanMartinez:
Looks like Weeder got updated to `9.2`: https://github.com/ocharles/weeder/releases/tag/2.4.0

hdgarrood:
So now that protolude has been updated I think the only remaining task is to figure out why cabal thinks PureScript depends on the GHC API. If you try building PureScript with cabal instead of stack, it tries to compile the GHC API and fails. I think I might write something up and ask on the Cabal issue tracker or something, because this is quite strange to me.

arrowd:
> If you try building PureScript with cabal instead of stack, it tries to compile the GHC API and fails

What package, exactly?

hdgarrood:
`ghc`

arrowd:
Can't reproduce this on my side. I just successfully compiled the code from this PR.

PR #4334, opened 2022-05-20
Accommodate internal initial-digit identifiers

rhendric:
The motivating example is the `xyzIsSymbol` style of variable name that
can be created by `CoreFn.CSE`.

**Description of the change**

Fixes issue reported [here](https://discourse.purescript.org/t/looking-for-beta-testers/3018/4). This is a fix for an issue triggered by an unreleased feature, so I didn't create a `fix` changelog entry for it, just an `internal` entry (because the fix could potentially have consequences on things other than the new feature).

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/internal_accommodate-initial-digit-idents.md
+++ b/CHANGELOG.d/internal_accommodate-initial-digit-idents.md
@@ -1,0 +1,1 @@
+* Accommodate internally-generated identifiers that start with digits
--- a/src/Language/PureScript/CodeGen/JS/Common.hs
+++ b/src/Language/PureScript/CodeGen/JS/Common.hs
@@ -19,9 +19,12 @@ moduleNameToJs (ModuleName mn) =
 --
 --  * Alphanumeric characters are kept unmodified.
 --
---  * Reserved javascript identifiers are prefixed with '$$'.
+--  * Reserved javascript identifiers and identifiers starting with digits are
+--    prefixed with '$$'.
 identToJs :: Ident -> Text
-identToJs (Ident name) = anyNameToJs name
+identToJs (Ident name)
+  | not (T.null name) && isDigit (T.head name) = "$$" <> T.concatMap identCharToText name
+  | otherwise = anyNameToJs name
 identToJs (GenIdent _ _) = internalError "GenIdent in identToJs"
 identToJs UnusedIdent = unusedIdent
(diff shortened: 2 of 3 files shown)


thomashoneyman (approved):

JordanMartinez (commented):
To clarify, what happens if there is a top-level member named `_two` or `_two2`? Would there be a naming clash between these?

rhendric:
Oh, underscore was a poor choice, wasn't it? I'll reuse `$$` then; that shouldn't be able to collide with anything. (`identCharToText` never produces consecutive dollar signs, and no reserved word or built-in would start with a digit.)

JordanMartinez:
> Oh, underscore was a poor choice, wasn't it? I'll reuse `$$` then; that shouldn't be able to collide with anything. (`identCharToText` never produces consecutive dollar signs, and no reserved word or built-in would start with a digit.)

Yeah, I think that should work.

JordanMartinez (approved):

JordanMartinez:
On another note, should there be a test for this in the `optimize` tests? To ensure that the outputted JS doesn't have a name clash with a similarly named `_two` top-level member?

rhendric:
I think all an optimize test would do is ensure that the generated identifier looks the same, not that it doesn't collide with something else. We could write such a test but it doesn't seem to me like it would be worth it; I have a hard time imagining how the `$$` prefix changes unintentionally, and if someone changes it intentionally then all that the failing optimize test will communicate is that they did what they intended to do.

What I'd much rather have would be something like a QuickCheck property that asserts that `identToJs` is injective (over the domain of inputs for which it doesn't error). We don't seem to have much of that sort of unit testing for simple and small units of code that don't change often, but I'd write it if you also think it'd be a good idea.

JordanMartinez:
> I think all an optimize test would do is ensure that the generated identifier looks the same, not that it doesn't collide with something else. We could write such a test but it doesn't seem to me like it would be worth it; I have a hard time imagining how the `$$` prefix changes unintentionally, and if someone changes it intentionally then all that the failing optimize test will communicate is that they did what they intended to do.
> 
> What I'd much rather have would be something like a QuickCheck property that asserts that `identToJs` is injective (over the domain of inputs for which it doesn't error). We don't seem to have much of that sort of unit testing for simple and small units of code that don't change often, but I'd write it if you also think it'd be a good idea.

Gotcha. Then, I don't think it's needed for this small fix.

PR #4336, opened 2022-05-22
Enable more warnings in purescript.cabal

hdgarrood:
This enables -Wincomplete-uni-patterns and -Wincomplete-record-updates
by default. They are both good warnings to have on, as they help
identify situations where code is likely to result in an error at
runtime.

-Wincomplete-uni-patterns causes a warning to be emitted for things like
this:

    head = \(x:_) -> x

which would blow up at runtime if given an empty list.

-Wincomplete-record-updates causes a warning to be emitted for code like
this:

    data Foo = Foo { x :: Int }
             | Bar

    f :: Foo -> Foo
    f foo = foo { x = 6 }

which will blow up if you try to call `f Bar`, since only the `Foo`
constructor has the `x` field. (This one isn't triggered by any of our
code at the moment, thankfully).

The reason I'm sending this now is that as of version 9.2.1, GHC
includes these two warnings as part of -Wall by default, which means
that in order to upgrade, we'd need to either rework our code to stop
emitting them, or explicitly turn them off. The former seems more
appealing to me.

Another nice consequence of this is that some functions become more
efficient, as we remove unnecessary extra traversals of lists in a
couple of cases.

**Checklist:**

### DIFF
--- a/src/Language/PureScript/CoreImp/Optimizer/Unused.hs
+++ b/src/Language/PureScript/CoreImp/Optimizer/Unused.hs
@@ -20,8 +20,11 @@ removeCodeAfterReturnStatements :: AST -> AST
 removeCodeAfterReturnStatements = everywhere (removeFromBlock go)
   where
   go :: [AST] -> [AST]
-  go jss | not (any isReturn jss) = jss
-         | otherwise = let (body, ret : _) = break isReturn jss in body ++ [ret]
+  go jss =
+    case break isReturn jss of
+      (_, []) -> jss
+      (body, ret : _ ) -> body ++ [ret]
+
   isReturn (Return _ _) = True
   isReturn (ReturnNoResult _) = True
   isReturn _ = False
--- a/src/Language/PureScript/Ide/Imports.hs
+++ b/src/Language/PureScript/Ide/Imports.hs
@@ -33,7 +33,7 @@ module Language.PureScript.Ide.Imports
 import           Protolude hiding (moduleName)
 
 import           Control.Lens                       ((^.), (%~), ix, has)
-import           Data.List                          (findIndex, nubBy, partition)
+import           Data.List                          (nubBy, partition)
 import qualified Data.List.NonEmpty                 as NE
 import qualified Data.Map                           as Map
 import qualified Data.Text                          as T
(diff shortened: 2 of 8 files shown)


src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -863,3 +865,23 @@ pairwiseM :: Applicative m => (a -> a -> m ()) -> [a] -> m ()
 pairwiseM _ [] = pure ()
 pairwiseM _ [_] = pure ()
 pairwiseM p (x : xs) = traverse (p x) xs *> pairwiseM p xs
+
+-- | Return all nonempty suffixes of a nonempty list. For example:
+--
+-- tails1 (fromList [1]) == fromList [fromList [1]]
+-- tails1 (fromList [1,2]) == fromList [fromList [1,2], fromList [2]]
+-- tails1 (fromList [1,2,3]) == fromList [fromList [1,2,3], fromList [2,3], fromList [3]]
+tails1 :: NonEmpty a -> NonEmpty (NonEmpty a)
```
hdgarrood, line 874: I opened https://github.com/haskell/core-libraries-committee/issues/67 to see if we can get this into `base`.

src/Language/PureScript/Sugar/Names/Env.hs
```diff
@@ -488,5 +489,8 @@ checkImportConflicts ss currentModule toName xs =
         return (mnNew, mnOrig)
       _ -> throwError . errorMessage' ss $ ScopeConflict name conflictModules
     else
-      let ImportRecord (Qualified (Just mnNew) _) mnOrig _ _ = head byOrig
-      in return (mnNew, mnOrig)
+      case head byOrig of
+        ImportRecord (Qualified (Just mnNew) _) mnOrig _ _ ->
+          return (mnNew, mnOrig)
+        _ ->
+          internalError "checkImportConflicts: ImportRecord should be qualified"
```
hdgarrood, line 496: This probably won't ever happen, but if it does, we'll get a much more useful error.

src/Language/PureScript/Ide/Imports.hs
```diff
@@ -256,13 +256,14 @@ addExplicitImport' decl moduleName qualifier imports =
...
+-- function 'update'.
 updateAtFirstOrPrepend :: (a -> Bool) -> (a -> a) -> a -> [a] -> [a]
-updateAtFirstOrPrepend p t d l =
-  case findIndex p l of
-    Nothing -> d : l
-    Just i ->
-      let (x, a : y) = splitAt i l
-      in x ++ [t a] ++ y
+updateAtFirstOrPrepend predicate update def xs =
+  case break predicate xs of
```
hdgarrood, line 264: This has the exact same behaviour as before. Using `break` here removes an unnecessary list traversal.

src/Language/PureScript/CoreImp/Optimizer/Unused.hs
```diff
@@ -20,8 +20,11 @@ removeCodeAfterReturnStatements :: AST -> AST
 removeCodeAfterReturnStatements = everywhere (removeFromBlock go)
   where
   go :: [AST] -> [AST]
-  go jss | not (any isReturn jss) = jss
-         | otherwise = let (body, ret : _) = break isReturn jss in body ++ [ret]
+  go jss =
+    case break isReturn jss of
```
hdgarrood, line 24: This has the exact same behaviour as before; `not (any isReturn xs)` is just as expensive to compute as `break isReturn xs`, so we might as well do the latter since that also gives us access to the information we need here in case there is a return statement.

MonoidMusician (approved):

rhendric (approved):

hdgarrood:
Just in case you aren't aware, I removed myself from the core team on GitHub so I'm no longer able to merge things by myself, even with approvals - please merge this for me when you're ready.

PR #4341, opened 2022-05-27
Increase rendered typed holes to max of 30

### LINKED ISSUE #4340: Increase max limit for rendered type holes
## Summary

Per [a Discourse thread](https://discourse.purescript.org/t/typed-holes-limited-to-a-context-of-5-values/3008), increase the number of type holes displayed. I'm not sure how much more should be displayed, but perhaps up to 30 is a good number?

## Proposal

This is an easy fix. Change [this line](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Errors.hs#L1596) to `take 30` rather than `take 5`.

JordanMartinez:
**Description of the change**

Fixes #4340

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_increase-type-holes-amount.md
+++ b/CHANGELOG.d/feature_increase-type-holes-amount.md
@@ -1,0 +1,1 @@
+* Increases the max number of typed holes displayed from 5 up to 30
--- a/src/Language/PureScript/Errors.hs
+++ b/src/Language/PureScript/Errors.hs
@@ -1593,7 +1593,7 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath) e = fl
           [ Box.hcat Box.left [ Box.text (T.unpack (showIdent ident) ++ " :: ")
                               , markCodeBox $ typeAsBox prettyDepth ty'
                               ]
-          | (ident, ty') <- take 5 ctx
+          | (ident, ty') <- take 30 ctx
           ]
       ]
 


rhendric (approved):
I suppose the best way to have an actual conversation about a number this arbitrary and unexciting is to change it and see who complains!

natefaubion:
One thing I'm not aware of is whether the size of the list that's forced contributes to how much work is done to compute it. That is, it's already pretty slow at 5 items. Does changing it to 30 or whatever increase the work significantly (ie. is the list produced lazily?). If the same work is done regardless, then it doesn't really matter.

rhendric:
The list ultimately comes from

https://github.com/purescript/purescript/blob/3c28f9cd08212216a0f8bd3e2bcb207aa6cfbfab/src/Language/PureScript/TypeChecker/Monad.hs#L314-L318

Clearly some work is saved by truncating that list but the only meaningful source of laziness I see here is if `names env` contains some lazy values.

JordanMartinez:
What would be a good way to confirm whether this adds a lot of work or not?

JordanMartinez:
One way we could test this is do a prerelease and see what people find in their usage of it. It's not hard to change this back to 5 (or some other higher-than-5 number). This is also not a breaking change, so we can adjust this as needed.

PureFunctor:
I think trialing this for 0.15.3 would be fine.

JordanMartinez:
Part of me wonder if this should be merged as the first `0.15.4` prerelease. If we had merged this back when I got Ryan's approval, then we could have gotten early feedback about whether 30 is a good number or causes problems. With `0.15.3` coming out next Tuesday (per our 6-week release cycle), this PR hasn't had a lot of soak time.

Alternatively, this could get merged and `0.15.3` released. And if there are any issues, we make a quick `0.15.4` release that reverts this change. I think I'd lean towards this approach.

JordanMartinez:
With `0.15.3` released, I'm merging this. We have another 6 weeks to see what issues this change may create before making a full release.

Issue #3939, opened 2020-09-23
Proposal: Allow U+0009 in source files


singpolyma:
## Summary

Purescript is an indentation-based syntax, similar to Haskell.  It would therefore be useful to be able to indent code using the character set aside for this purpose, which is U+0009.

## Motivation

Currently U+0009 is considered a syntax error.

## Proposal

Allow U+0009 and treat it as one level of indentation.  If a layout algorithm that is not aware of indentation is used, treating it as a single whitespace character should suffice.  If a very picky alignment-based algorithm is used, treating it as eight whitespace characters should suffice.

## Examples

```
main = do
	log "hello"
```

hdgarrood:
> Indentation in PureScript is spaces, and I don't think that should change.

Is this more for implementation-based reasons or for stylistic reasons? Do you think the implementation would still be difficult if we made it so that mixing tabs and spaces in a particular source file was disallowed? I was fully on board with disallowing tabs when it happened, but after reading https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/ I'm less convinced.

natefaubion:
Haskell-style indentation is based on exact columns, it does not have any notion of a semantic tab (unlike Python). Supporting tabs (or any other whitespace that is not a single column) involves making an arbitrary decision about how the compiler translates tabs to spaces. The whitespace rules are already very complicated, and I personally am not keen on extending them at all in any way other than what we already support.

natefaubion:
I should note, that I'm not keen on Haskell-style indentation at all 😆 and vastly prefer more rigid indentation rules. But I just think the way Haskell does it inevitably unifies everyone around spaces.

hdgarrood:
Just to ensure I've understood: the issue of indentation being based on exact columns can manifest in situation where you have code like, for example,

```
let x = f
<tab>y :: T
```
which could parse as
```
let x = f y :: T
```
if a tab counts as 4 columns wide, or as
```
let
  x = f
  y :: T
```
if a tab counts as 8 columns wide?

natefaubion:
Yes.

garyb:
I'd be in favour of this if it preserved the no-mixing-tabs-and-spaces rule, so you could have a tab-only indented file. That does indeed disallow aligned indenting (in the absence of a smart editor), but I think if people really want to use tabs then that's an acceptable tradeoff.

singpolyma:
```
let x = f
<tab>y :: T
[...]

PR #4345, opened 2022-05-29
Allow most leading and trailing delimiters

rhendric:
**Description of the change**

Inspired by [this Discourse post](https://discourse.purescript.org/t/adt-formatting-why-no-leading-pipe-support/3027), I made a small patch to the CST to support optional leading and trailing delimiters in almost every location in the grammar where I didn't think it would cause problems.

The one place I was less than maximally liberal was in multi-headed `case` expressions. When I tried to make extra delimiters optional in patterns, something prevented it from working in my tests—probably related to the fancy ad-hoc monadic parsing that is used in that part of the grammar.

Example of what I'm talking about:

```purs
case
  , expr1
  , expr2
of
  , BigPattern1
  , BigPattern2
    -> ...
  , BigPattern3 -- this is where the issue was
  , BigPattern4
    -> ...
```

Rather than try to fix it, I chose not to enable extra delimiters in that part of the grammar; and then, for consistency, also removed support for extra delimiters from the head of the `case` as well (even though those worked fine). While this shortcoming is regrettable, I think it'd be quite low down on the list of places in the grammar where users would want to use extra delimiters.

Also note that using leading and trailing delimiters together, a la `[,1,]`, is not supported anywhere. It would have been trivial to do so but I have standards, sir, standards!

Despite submitting this PR and the voiced support for at least some of the syntax changes included here, I have two misgivings. The first is that I don't know if any of these changes break any assumptions made during the layout phase of lexing, which does seem to be sensitive to where commas appear. I haven't noticed any issues but I don't have a thorough understanding of that algorithm. The second is that, by cravenly refraining from picking a side between the delimiters-first versus delimiters-last styles of code formatting, I fear I may stoke the flames of utterly unproductive format wars. There is definitely an argument for having one clear way to do simple things in a language and I don't know how strongly that argument should apply here.

Regardless of the above reservations, this PR is ready for review and discussion, though I don't consider it all that urgent.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_leading-and-trailing-delimiters.md
+++ b/CHANGELOG.d/feature_leading-and-trailing-delimiters.md
@@ -1,0 +1,15 @@
+* Allow leading and trailing delimiters in many syntax positions
+
+  PureScript's syntax contains multiple places where sequences of words or
+  clauses are expected, delimited by commas or vertical bars. To
+  facilitate editing these elements when placed on their own lines,
+  PureScript's parser now supports an optional leading or trailing
+  delimiter in most such locations, specifically:
+
+    * Array and object literals (and patterns)
+    * Sum type definitions
+    * Object updates
+    * Row types
+    * Import and export lists
+    * Functional dependencies
+    * Compound constraints and superclasses
--- a/src/Language/PureScript/CST/Convert.hs
+++ b/src/Language/PureScript/CST/Convert.hs
@@ -107,7 +107,7 @@ convertType fileName = go
         let ann = sourceAnnCommented fileName (lblTok a) (snd $ typeRange ty)
         T.RCons ann (L.Label $ lblName a) (go ty) c
     case labels of
-      Just (Separated h t) ->
+      Just (SeparatedExtra _ h t _) ->
         rowCons h $ foldr (rowCons . snd) rowTail t
       Nothing ->
         rowTail
@@ -276,7 +276,7 @@ convertExpr fileName = go
       let
         ann = sourceAnnCommented fileName a c
         vals = case bs of
-          Just (Separated x xs) -> go x : (go . snd <$> xs)
+          Just (SeparatedExtra _ x xs _) -> go x : (go . snd <$> xs)
           Nothing -> []
       positioned ann . AST.Literal (fst ann) $ AST.ArrayLiteral vals
     ExprRecord z (Wrapped a bs c) -> do
@@ -286,7 +286,7 @@ convertExpr fileName = go
           RecordPun f -> (mkString . getIdent $ nameValue f, go . ExprIdent z $ QualifiedName (nameTok f) Nothing (nameValue f))
           RecordField f _ v -> (lblName f, go v)
         vals = case bs of
-          Just (Separated x xs) -> lbl x : (lbl . snd <$> xs)
+          Just (SeparatedExtra _ x xs _) -> lbl x : (lbl . snd <$> xs)
           Nothing -> []
       positioned ann . AST.Literal (fst ann) $ AST.ObjectLiteral vals
     ExprParens _ (Wrapped a b c) ->
@@ -399,7 +399,7 @@ convertBinder fileName = go
       let
         ann = sourceAnnCommented fileName a c
         vals = case bs of
-          Just (Separated x xs) -> go x : (go . snd <$> xs)
+          Just (SeparatedExtra _ x xs _) -> go x : (go . snd <$> xs)
           Nothing -> []
       positioned ann . AST.LiteralBinder (fst ann) $ AST.ArrayLiteral vals
     BinderRecord z (Wrapped a bs c) -> do
@@ -409,7 +409,7 @@ convertBinder fileName = go
           RecordPun f -> (mkString . getIdent $ nameValue f, go $ BinderVar z f)
           RecordField f _ v -> (lblName f, go v)
         vals = case bs of
-          Just (Separated x xs) -> lbl x : (lbl . snd <$> xs)
+          Just (SeparatedExtra _ x xs _) -> lbl x : (lbl . snd <$> xs)
           Nothing -> []
       positioned ann . AST.LiteralBinder (fst ann) $ AST.ObjectLiteral vals
     BinderParens _ (Wrapped a b c) ->
@@ -442,7 +442,7 @@ convertDeclaration fileName decl = case decl of
             [] -> []
             (st', ctor) : tl' -> ctrs st' ctor tl'
           )
-    pure $ AST.DataDeclaration ann Env.Data (nameValue a) (goTypeVar <$> vars) (maybe [] (\(st, Separated hd tl) -> ctrs st hd tl) bd)
+    pure $ AST.DataDeclaration ann Env.Data (nameValue a) (goTypeVar <$> vars) (maybe [] (\(st, SeparatedExtra _ hd tl _) -> ctrs st hd tl) bd)
   DeclType _ (DataHead _ a vars) _ bd ->
     pure $ AST.TypeSynonymDeclaration ann
       (nameValue a)
--- a/src/Language/PureScript/CST/Flatten.hs
+++ b/src/Language/PureScript/CST/Flatten.hs
@@ -10,7 +10,7 @@ flattenModule :: Module a -> DList SourceToken
 flattenModule m@(Module _ a b c d e f g) =
   pure a <>
   flattenName b <>
-  foldMap (flattenWrapped (flattenSeparated flattenExport)) c <>
+  foldMap (flattenWrapped (flattenSeparatedExtra flattenExport)) c <>
   pure d <>
   foldMap flattenImportDecl e <>
   foldMap flattenDeclaration f <>
@@ -32,7 +32,7 @@ flattenClassHead (ClassHead a b c d e) =
   foldMap (\(f, g) -> flattenOneOrDelimited flattenConstraint f <> pure g) b <>
   flattenName c <>
   foldMap flattenTypeVarBinding d <>
-  foldMap (\(f, g) -> pure f <> flattenSeparated flattenClassFundep g) e
+  foldMap (\(f, g) -> pure f <> flattenSeparatedExtra flattenClassFundep g) e
 
 flattenClassFundep :: ClassFundep -> DList SourceToken
 flattenClassFundep = \case
@@ -74,9 +74,9 @@ flattenBinder = \case
   BinderChar _ a _ -> pure a
   BinderString _ a _ -> pure a
   BinderNumber _ a b _ -> foldMap pure a <> pure b
-  BinderArray _ a -> flattenWrapped (foldMap (flattenSeparated flattenBinder)) a
+  BinderArray _ a -> flattenWrapped (foldMap (flattenSeparatedExtra flattenBinder)) a
   BinderRecord _ a ->
-    flattenWrapped (foldMap (flattenSeparated (flattenRecordLabeled flattenBinder))) a
+    flattenWrapped (foldMap (flattenSeparatedExtra (flattenRecordLabeled flattenBinder))) a
   BinderParens _ a -> flattenWrapped flattenBinder a
   BinderTyped _ a b c -> flattenBinder a <> pure b <> flattenType c
   BinderOp _ a b c -> flattenBinder a <> flattenQualifiedName b <> flattenBinder c
@@ -94,7 +94,7 @@ flattenRecordUpdate :: RecordUpdate a -> DList SourceToken
 flattenRecordUpdate = \case
   RecordUpdateLeaf a b c -> flattenLabel a <> pure b <> flattenExpr c
   RecordUpdateBranch a b ->
-    flattenLabel a <> flattenWrapped (flattenSeparated flattenRecordUpdate) b
+    flattenLabel a <> flattenWrapped (flattenSeparatedExtra flattenRecordUpdate) b
 
 flattenLambda :: Lambda a -> DList SourceToken
 flattenLambda (Lambda a b c d) =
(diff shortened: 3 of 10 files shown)


toastal:
> I fear I may stoke the flames of utterly unproductive format wars

This is why I initially brought up specifically ADTs to narrow the scope :sweat_smile:

JordanMartinez:
This PR can be broken down into smaller changes and their corresponding discussions. I'll provide an example of the change and name it:
```purs
-- Optional leading pipe character for first data constructor
data Foo =
 | Bar
 | Baz
```
I think I'm good with making this change because
1. it allows one to swap the order of data constructors (sometimes useful for a compiler-derived `Ord` instance) using the `ALT+↓` or `ALT+↑` keyboard shortcut.
1. it allows one to comment/uncomment data constructors while prototyping.
1. developers hit annoyances with current syntax limitations more frequently than other syntax (e.g. module import, export, etc.)
1. it's a small change to the language, so it doesn't feel _that_ different and limits the fallout of a potential formatting war.

```purs
-- Optional trailing pipe character for last data constructor
data Foo =
  Bar |
  Baz |
```

This change feels a bit weird to me because the pipe character is no longer in front of the data constructor. So, this change feels like a "bigger" one than the one above. Granted, I know that `Foo = Bar | Baz` is valid and so adding another `|` to the end (e.g. `Foo = Bar | Baz |`) isn't that different. However, putting these on multiple lines is where things started to feel weird.
I think I'd prefer the first change and not this change if any change is made at all because I don't see what value this change would provide that isn't already supported by the first change.

The only argument I can think of against this is that one could copy-paste the above constructors into a file and then replace `|` with `->` to immediately get all data constructors for a `case _ of` expression. Doing the same with the first version requires deleting `| ` and then appending ` ->`, which is a bit more work. However, I don't think this "problem" should be solved by supporting this syntax in the language. Rather, the IDE should support such functionality (if it doesn't already). To my knowledge, it doesn't currently support this.

As for module exports, below are a few examples of how this could be formatted:
```purs
-- Current syntax. Uses 4 lines
module Module1
  ( (<>)
  , b
  ) where

-- Optional leading comma character for first export
-- uses 5 lines
module Module2
  (
  , (<>)
  , b
  ) where
  
-- Same as Module2 but formatted differently to use 4 lines
module Module3 (
  , (<>)
  , b
  ) where

-- Optional trailing comma character for last export
-- uses 4 lines
module Module4 (
  (<>),
  b,
  ) where
```
Let me briefly comment on these examples. The order of exports determines their order in the resulting documentation. `Module1` can't re-order the `(<>)` export using the `ALT+↓` keyboard shortcut. Similarly, `Module1` can't comment/uncomment a line to add/remove the export. `Module2` can for both but, unlike the data constructor example above, it adds an unnecessary line. `Module3` fixes `Module2`'s issues, but I don't have a good argument for why `Module4` shouldn't be accepted since it also enables both. 

That being said, I don't know if I care about the module export syntax. The only time someone touches export lists manually is when they want to prevent some things from being exported. Since this happens infrequently, I think that's an argument against making this change: this small and infrequent annoyance due to a syntax limitation is less of an issue than formatting wars.
If this change was made, I'd be ok with `Module3`, but I think I'd prefer `Module4` because it just feels cleaner than `Module3`. That being said, I wouldn't want `Module4` unless `purs-tidy` also works with it.

(Side note... While @rhendric added a test for `Maybe(Just, Nothing)`, I don't think this isn't needed. A change was made sometime ago in the language that requires either all or none of a type's data constructors to be imported/exported, but I don't think the parser was updated to account for that.)

As for module imports, I don't know if this issue matters that much either. Neither the order of the modules, nor their import lists matter, so long as they are imported. Moreover, the IDE can add imports via autocompletion, remove unused imports by applying the suggestions inferred after analyzing import-related compiler warnings, and sort them if one is already using a formatter like `purs-tidy`. So the `ALT+↓` concern is less of an issue. Similarly, if one is already using `purs-tidy`, most imported identifiers appear on the same line, so the comment/uncomment workflow isn't as relevant. Lastly, since most of this is handled by the IDE, developers rarely encounter issues with this that require manual intervention.
All that to say I don't think there's a strong argument for changing the import syntax at all.

As for type class constraints in type class definitions and instances, things are a bit different than module import syntax. Constraint order does not matter, so no argument here due to the `ALT+↓` concern. However, constraints must be added manually (i.e. no IDE autocompletion). Another difference is that sometimes the constraints appear on one line and other times appear on multiple lines. So, the commenting/uncommenting workflow can but does not always apply here:
```purs
-- single line
class (Foo a) => Bar a

-- multi-line
instance
  ( Foo a
  , Foo b
  ) => Foo (Tuple a b)
```
So, I think the comment/uncomment workflow argument I made for module export syntax can apply here, but the ordering one and the infrequency arguments don't apply. Given this tradeoff, I think avoiding a formatter war is more important than removing this syntactic limitation.

As for the leading/trailing comma for array and record syntax, I am sympathetic to the idea of supporting the trailing comma because one could copy-paste JS arrays/records into PureScript and more cases would compile despite the trailing comma. (This "problem" could also be resolved by making `purs-tidy` smart enough to still parse the code, drop the trailing comma, and format it. But then `purs-tidy` supports parsing PureScript with extensions, and that seems like a bad precedent to start.) In my experience, copy-pasting JS array/record values into PS code happens infrequently and generally requires updating parts of the JS value anyway, so I don't think this is a strong argument.

However, the current array/record syntax do not support the `ALT+↓` keyboard shortcut or the comment/uncomment feature. And _unlike the module syntax example_, the lines of code between the leading and trailing versions aren't different and this syntax limitation would be something a developer experiences frequently:
```purs
-- leading
[
, a
, b
]

-- vs

-- trailing
[
  a,
  b,
]
```
If we did support a change, I would prefer supporting a leading delimiter for array/record syntax and not a trailing delimiter because it's a "smaller" change. However, I think such a change should only be made if it works in all contexts, including single or multi `case _ of` statements. Otherwise, we'd have to tell beginners, "You can use this array/record syntax in context A, but not B", making the language a bit harder to learn.

Phew! All that being said, I think this is where I stand:
- I support leading `|` in data constructors.
- I support leading `,` in array/record syntax if the below conditions are satisfied:
    - single/multi-head `case _ of` syntax works
    - `purs-tidy` works on it and Nate is in agreement with this change (out of respect for Nate since he wrote both the current parser and `purs-tidy`, both of which took a lot of time and effort)
- I don't support changing the module import, module export, or type class constraint (definition or instance) syntax.

rhendric:
First, from a purely what-does-this-enable-for-users perspective, let me note that there is absolutely no difference between leading and trailing delimiters. In particular, all of your numbered arguments for `| Bar` apply without change to `Bar |`. (I disagree with your assessment that trailing delimiters is a ‘bigger’ change in this case than leading delimiters—if it seems that way, surely it's just because the formatting style you're most used to approximates leading delimiters in this case?) Both are equivalent implementations of what I'll call the ‘Uniform Line Syntax’ principle: the idea that if a sequence of syntax elements can be laid out one per line, any additions, deletions, or permutations of the sequence should be possible to achieve with corresponding additions, deletions, or permutations of lines, one for one, with no special consideration needed for first or last lines.

So the way I see this proposal is this:
1. Is the ULS principle worth implementing anywhere in our grammar?
2. If so, where?
3. In those positions, should ULS be implemented with leading delimiters, trailing delimiters, or either? And to what degree should this question be informed by:
    * Trends in the formatting of current PureScript, or other programming languages that don't support ULS
    * The design of other programming languages that do support ULS
    * The answer to this question in other positions

I think 1 is a clear ‘yes’. I think the answer to 2 is most positive for sum type definitions and array/object literals and patterns, followed closely by object update expressions and row types; I also think a decent case can be made for import/export lists. I'm neutral on constructor lists within import lists, fundeps, compound constraints, and multi-head case expressions—I don't expect those to excite anyone, until the one user three years from now who'll say, ‘Hey, I'm designing a class with seventeen possibly functional dependencies that I want to toggle and I don't know why I can't use (leading/trailing) commas there just like I use them with big arrays.’

3 contains the details harboring the devil. Here's what I think I know:

* For sum types, there is a clear desire for leading `|`, supported by the syntax of ML and by the dominant formatting style in Haskell and PureScript to date.
* For array/object literals, I expect support for leading versus trailing `,` to be split. The leading side is supported by a general ‘symbols first’ trend in the formatting of PureScript and Haskell. The trailing side is supported by the fact that it's the style of ULS much more commonly found in the equivalent of array/object literals in other programming languages: JavaScript, Python, Ruby, Perl, PHP, Lua, C, C#, Java, Kotlin, Scala, Rust... and those are just the ones I know off the top of my head. I can think of several programming languages besides PureScript and Haskell that don't support leading or trailing commas in such expressions (ML, F#, Groovy, SQL, Prolog), but I personally can't think of a single language, popular or esoteric, that supports leading commas and *not* trailing commas. Especially given the syntactic synergy that our literals have with JavaScript—we support almost all of the same literals with the same syntax and the same semantics—I think it would be foolish to implement ULS anywhere without specifically allowing trailing commas in array and record literals.

As for the rest... I could see a mostly conservative proposal of only leading `|` for sum types, only trailing `,` for array-ish and object-ish stuff (encompassing literals, patterns, updates, and row types), and no ULS anywhere else. I could see a consistent-by-character proposal of leading `|` and trailing `,` wherever applicable. I could see a leading-delimiters-allowed-generally with a specific carve-out for *also* allowing trailing `,` for array-ish and object-ish stuff. And I can see the mostly liberal proposal of allowing both types of ULS generally. I think I'd have a tough time supporting more complicated ways of slicing this design space; I'd rather not make it difficult to remember where leading and trailing delimiters are allowed by making those choices in too ad-hoc a way.

To some of your specific points:

> While @rhendric added a test for `Maybe(Just, Nothing)`, I don't think this isn't needed. A change was made sometime ago in the language that requires either all or none of a type's data constructors to be imported/exported, but I don't think the parser was updated to account for that.

Exported, but not imported. It's still legal, and sometimes useful, to import some but not all of a type's constructors; and so it's still possible to want to individually add and remove from that list, which means that ULS can still add (an admittedly small amount of) value here.

> * I support leading `,` in array/record syntax if the below conditions are satisfied:
>     * single/multi-head `case _ of` syntax works

Just to clarify, the multi-head case issue has nothing to do with array/record patterns. These commas:

```purs
case
  , expr1
  , expr2
of
  , BigPattern1
  , BigPattern2
    -> ...
  , BigPattern3 -- this is where the issue was
  , BigPattern4
    -> ...
```

are not the same as these commas:

```purs
case someArrayValuedExpression of
  [
  , BigPattern1
  , BigPattern2
  ] -> ...
```

toastal:
I just saw some Haskell code where the imports had trailing commas. It felt a little weird and I agree this is mostly a bias on how ML-code *should* look like rather than something objective. This could be enforced with a formatter (`{ trailingCommas: Always }`) but it does add complexity and fork in styles. I believe it would be less like a "war" than a bias to the familiar (do you come from JavaScript on Haskell?). I think both sides are correct and there is subjectivity at play for liberal usage for maximal expressiveness + principally odd to choose only leading instead of both vs. conserving and keeping harmony the feel/aesthetic throughout the language.

(but +1 leading pipe on ADTs specifically 😉)

JordanMartinez:
> if it seems that way, surely it's just because the formatting style you're most used to approximates leading delimiters in this case?

That is definitely the case.

> Exported, but not imported. It's still legal, and sometimes useful, to import some but not all of a type's constructors; and so it's still possible to want to individually add and remove from that list, which means that ULS can still add (an admittedly small amount of) value here.

I tried `import Data.Maybe (Maybe(Just))` on Try PureScript and you're right. Thanks, I didn't know that.

PureFunctor:
I would say that one thing to consider is to determine whether or not these changes would end up becoming the "recommended" formatting to use for writing PureScript. A few years down the line, would we see new documentation making use of and potentially recommending these optional formatting styles? I've no strong arguments about allowing trailing `|` and `,` from an aesthetic and functional perspective though. OTOH, I think allowing ULS for import and export lists should be a separate PR unless it warrants support from users.

ursi:
I think less syntax is better. Less variation to parse mentally, less things to think about when deciding how to format your own code, less things for tooling to have to deal with, so I'm not a fan of this change. That being said, I do think good multi-line syntax is important, and I think maybe the one case above where something that doesn't currently have a good multi-line syntax now, would, is case statements with leading commas. *That* being said. I'm not sure how often I use multi-line case statements in my code (potentially never), so I'm not sure I feel it's worth the extra syntax.

wclr:
I don't think this is a good idea, this adds even more inconsistency of code styles and makes tings look diffting to JS style exessive flexibility.

davezuch:
It seems like leading pipes are at least much less controversial than the rest of this PR. Maybe those can be extracted out into their own PR and discussed? Personally, I would really like to have leading pipes, but I don't have strong feelings about the rest.

toastal:
@natefaubion you bring up a good point with leading commas. It seems there's a stylistic reason to try to keep a consistent left-hand character which would match leading pipe.

```purescript
type Result e a =
  | Success s → …
  | Failure e → …
```

I guess by that argument I could see an case for stylistic leading characters too like:

```purescript
burgerToppings ∷ Array String
burgerToppings =
  [
  , "bacon"
  , "peanut butter"
  , "fried banana"
  ]
```

This makes it easier to move/modify the first element, though it looks a little weird.

i-am-the-slime:
An interesting tidbit, I find, is something I've liked in zig which supports trailing commas.
The (language enforced) formatter will format things without trailing commas in them to be on a single line and when there's a trailing comma the code will be formatted to be on multiple lines.

toastal:
I may have changed my stance about the comma situation. I'm definitely still pro-leading pipe above anything, but someone pointed out to me that leading commas doesn't play the best with tabs (as opposed to spaces). There's an [open issue about tab support](https://github.com/purescript/purescript/issues/3939) for accessibility, and I'm inclined to support it (because my eyes have had trouble reading 2-space (and 1-space) indentation for the last 2 years). The last thing you want to do is start mixing the two though.

For

```purescript
[ a
, { b:
    { c: unit
    , d: unit
    }
  }
}
```

you could put tabs after the commas/opening symbols like

```purescript
[	a
,	{	b:
			{	c: unit
    		,	d: unit
			}
	}
}
```

or maybe you embrace the trailing commas syntax (easiest to read IMO)

```purescript
[
	a,
	{
		b:	{
			c: unit,	
			d: unit,
		},
	},
]
```

The examples I last posted are I believe harder to read in this case.

```purescript
[
	, a
	, {
		, b: {
			, c: unit
    		, d: unit
		}
	}
}
```

PR #4351, opened 2022-06-08
Update codebase to GHC 9.2.3

JordanMartinez:
**Description of the change**

Builds off of #4326 and updates codebase to GHC 9.2.2. The only remaining issue is the Debian image needed to ensure we don't bump glibc to 2.28. See https://github.com/purescript/purescript/pull/4326#discussion_r873233459

---

**Checklist:**


JordanMartinez:
Note: local test currently fails with:
```
Failures:

  tests/TestSourceMaps.hs:57:7: 
  1) sourcemaps, golden files, 'Bug4034.purs' should compile to expected output and produce a valid source map file
       Test output differed from 'tests/purs/sourcemaps/Bug4034.out.js.map'; got:
       {"file":"index.js","mappings":";;;;;;;;AA2BA,eAAA;;WAAA;kCACwB;;;;;;AAXxB,cAAA;;WAAA;iCACsB;;;;;;AA6BtB,eAAA;;WAAA;kCACwB;;;;;;AAXxB,gBAAA;;WAAA;mCAC0B;;;;;;AAmB1B,gBAAA;;WAAA;mCAC0B","names":[],"sources":["../../tests/purs/sourcemaps/Bug4034.purs"],"version":3}

  To rerun use: --match "/sourcemaps/golden files/'Bug4034.purs' should compile to expected output and produce a valid source map file/"

  tests/TestSourceMaps.hs:57:7: 
  2) sourcemaps, golden files, 'Recipe.purs' should compile to expected output and produce a valid source map file
       Test output differed from 'tests/purs/sourcemaps/Recipe.out.js.map'; got:
       {"file":"index.js","mappings":"","names":[],"sources":[],"version":3}

  To rerun use: --match "/sourcemaps/golden files/'Recipe.purs' should compile to expected output and produce a valid source map file/"
```

arrowd:
Why not straight to `9.2.3`?

PureFunctor:
+1 on going to 9.2.3, as long as we don't encounter any problems or other implications; otherwise, this LGTM

JordanMartinez:
CI builds.

We only need someone to create a debian image, so that we don't bump glibc by merging this. Any takers? Or could anyone walk me through how to do that?

JordanMartinez:
AFAICT, we'd need to upload a Docker image to the Docker Hub that's essentially [this file](https://github.com/haskell/docker-haskell/blob/master/9.2/buster/Dockerfile) but changing [`FROM buildpack-deps:buster`](https://github.com/haskell/docker-haskell/blob/master/9.2/buster/Dockerfile#L1) to `FROM buildpack-deps:stretch`. Per https://github.com/haskell/docker-haskell/issues/72, If we asked if the resulting image could be added to the repo, I'm not sure whether that would be welcomed.

PureFunctor:
@JordanMartinez Could we roll out our own under a fork of the `docker-haskell` repository? I can try setting it up later.

I was also thinking of using GHCR over Docker Hub, but that's just personal preference.

PureFunctor:
I'd forked `docker-haskell` to add a stretch-based container for GHC 9.2.3, and I've published it [here](https://github.com/users/PureFunctor/packages/container/package/haskell-purs). Should we publish the container under the organization instead? I believe it'd require a setting to be tweaked in the org.

PureFunctor:
I noticed this in the workflow logs just now:
```
  if [ $(ldd $(stack path --local-doc-root)/../bin/purs | grep 'libtinfo' | wc -l) -ge 1 ]; then
    echo "libtinfo detected"
    ldd $(stack path --local-doc-root)/../bin/purs | grep 'libtinfo'
    exit 1
  fi
  shell: bash --noprofile --norc -e -o pipefail {0}
  env:
    CI_PRERELEASE: false
    CI_RELEASE: false
    STACK_VERSION: 2.7.5
Preventing creation of stack root '/github/home/.stack/'. Parent directory '/github/home/' is owned by someone else.
ldd: /../bin/purs: No such file or directory
```
Is this supposed to happen?

JordanMartinez:
CI builds now. Can I get an approval so we can get this merged?

rhendric:
Is this no-regrets? Stackage's current LTS series is still on 9.0.2. Would this update inconvenience consumers of PureScript-as-a-library?

JordanMartinez:
> Is this no-regrets? Stackage's current LTS series is still on 9.0.2. Would this update inconvenience consumers of PureScript-as-a-library?

Good question. And I don't know the answer.

JordanMartinez:
I've opened https://discourse.purescript.org/t/getting-feedback-from-purescript-as-a-library-users-updating-to-ghc-9-2-3 to try to get more feedback from the community in a more visible way than this PR.

JordanMartinez:
On a separate note, updating GHC here would also mean updating Try PureScript's version of GHC. So, we'd need to verify that that works with this update as well.

hdgarrood:
Personally I’d suggest not worrying about people consuming the compiler as a library - I’d argue it’s more important to improve the experience for people who are using the compiler, and IMO this updated does improve the experience quote a bit for ARM users (I’m thinking of M1 macs in particular), due to the improved ARM support in the updated GHC version. People consuming the compiler as a library are used to breaking changes happening all the time anyway; we (imo, rightly) don’t try to make the compiler library interface stable.

rhendric:
I know we break interface all the time, and I am likewise not worried about that. But I think it's probably easier to adapt code to a new interface than to adapt an entire project that is now in dependency hell because PureScript won't compile without a specific nightly version of Stackage. That might be worth some pause.

I don't actually know, though; PureScript is the largest Haskell project I've ever worked on!

hdgarrood:
It's almost always been the case that PureScript won't compile without a specific version of Stackage, or rather, my preferred way of putting it would be that all of the bounds in purescript.cabal are deliberately very narrow. (We do support building with cabal, more or less, because that's how the homebrew purescript package works, and I think maybe also one or two other non-official distributions work that way too). I think Stackage nightly vs LTS is almost entirely immaterial; a Stackage nightly is still composed entirely of real (i.e. non-prerelease) versions that each of the relevant library maintainers think are ready for general use.

hdgarrood:
Also, now that purty and zephyr seem to be more or less unmaintained, the only other projects I'm aware of that use the PureScript compiler as a library are Pursuit and Try PureScript, both of which I expect can deal with updating to GHC 9.2 without much grief.

hdgarrood:
Oh, I take back what I said about zephyr - I forgot that it has a new maintainer. I think zephyr is probably a bit less important now that we have ES modules though.

rhendric:
That's fine, then. If whatever causes the build failures is resolved I'm ready to approve this.

tests/TestGraph.hs
```diff
@@ -4,11 +4,8 @@ import Prelude ()
 import Prelude.Compat
 
 import Test.Hspec
-import System.IO.UTF8 (readUTF8FileT)
 import Data.Either (isLeft)
 
-import qualified Data.ByteString.Lazy as ByteString
-import qualified Data.Text.Encoding as Text
 import qualified Data.Aeson as Json
```
PureFunctor, line 9: I think these imports would have to be restored for this to build properly.
```suggestion
import System.IO.UTF8 (readUTF8FileT)
import Data.Either (isLeft)

import qualified Data.Aeson as Json
import qualified Data.ByteString.Lazy as ByteString
import qualified Data.Text.Encoding as Text
```

PureFunctor (dismissed):

tests/TestGraph.hs
PureFunctor: Ah, it looks like e97495c263cbe96c6fc54603b9d6419612135d72 introduced a conflict with this PR. It should work without all the extra encoding/decoding.

rhendric (approved):

arrowd:
Glad to see this merged! Any ETA on a new release containing this change?

PR #4358, opened 2022-06-28
Fix name clash in guard clauses

### LINKED ISSUE #4357: Variable names clashing across guard clauses
## Description

This is a regression apparently introduced in #4293.

A case statement containing guard clauses with pattern guards which bind the same variable result in an error when building with 0.15.3 but not 1887bcf2ebb3f15f8410f1d353e344e42f7d3b9c (just before that PR was merged).

## To Reproduce

```purescript
data Foo = Foo Int | Bar Int

g :: Foo -> Int
g  =
  case _ of
    a
        | Bar z <- a
        -> z
        | Foo z <- a
        -> z
        | otherwise
        -> 42
```

Results in the following error:

```
Error found:
in module Main
at src/Main.purs:15:12 - 15:13 (line 15, column 12 - line 15, column 13)

  Value z is undefined.

while inferring the type of z
in value declaration g
```

After renaming the `z` in the 1st guard clause, the error disappears.

## Expected behavior

No error - as in previous compiler versions.

Variables bound in guard clauses are only in scope in the corresponding expression, and there is no conflict between different guard clauses. 

## PureScript version

0.15.3

PureFunctor:
**Description of the change**

Fixes #4357. Should we make a hotfix release for this?

---

**Checklist:**


JordanMartinez:
I'd be up for a hotfix. Right now, `master` includes #4341, so perhaps this PR should be rebased on top of `0.15.3` and we release the result as `0.15.4`?

src/Language/PureScript/Sugar/Names.hs
```diff
@@ -256,7 +257,7 @@ renameInModule imports (Module modSS coms mn decls exps) =
     unless (length (ordNub args) == length args) .
       throwError . errorMessage' pos $ OverlappingNamesInLet
     return ((pos, declarationsToMap ds `M.union` bound), Let w ds val')
-  updateValue (_, bound) (Var ss name'@(Qualified qualifiedBy ident)) =
+  updateValue (_, bound) (Var ss name'@(Qualified qualifiedBy ident)) = do
```
rhendric, line 260: Unneeded?

rhendric (approved):
Looks correct; here are some optional cleanup suggestions.

PureFunctor:
> so perhaps this PR should be rebased on top of 0.15.3 and we release the result as 0.15.4?
How would we go about doing so?

JordanMartinez:
> > so perhaps this PR should be rebased on top of 0.15.3 and we release the result as 0.15.4?
> 
> How would we go about doing so?

I believe we'd do the same thing we did in the `0.14.9` release: https://github.com/purescript/purescript/tree/continue-v0.14.x

We'd create a branch that starts on `0.15.3` called `0.15.4-next`, reimplement this PR and merge it to that branch, bump the version of that branch to `0.15.4` and merge the PR, and then make a GH release targeting that branch (rather than `master`). 

Per the table in https://github.com/purescript/purescript/issues/4303, I don't think there is anything else to do after that?

PureFunctor:
I've rebased this PR to build on `v0.15.3` and changed the target to `0.15.4-next`. This should be ready for a merge and release.

Issue #3168, opened 2017-12-11
Superclasses must also be newtype derived


MonoidMusician:
As I understand it now, we can `newtype` a type and derive instances of classes for it, without necessarily having all of the instances derived going up the chain. Which is fine, it won't crash any code.

But the compiler still allows me to write my _own_ instance for farther up the chain, which will introduce consistency issues. For example, the following will access different functor instances:

```
module Main where

import Prelude
import Data.Const
import Control.Monad.Eff.Console (logShow)

newtype S a = S (Const String a)

derive newtype instance eqS :: Eq (S a)
derive newtype instance applyS :: Apply S

instance functorS :: Functor S where
  map _ _ = S (Const "unlawful")

oneMap :: forall f a b. Apply f => (a -> b) -> f a -> f b
oneMap = map

otherMap :: forall f a b. Functor f => (a -> b) -> f a -> f b
otherMap = map

main = logShow $ let v = S (Const "lawful") in oneMap (_+0) v == otherMap (_+0) v
```

So I would propose we need to require superclass instances to also be newtype-derived.

MonoidMusician:
(also concerned that newtype instances may not be kind checked, hm ...)

garyb:
Hmm, I think there may be a bug here. We do have a check for this, the corresponding error is `MissingNewtypeSuperclassInstance`.

garyb:
Ah maybe that only triggers for fully missing instances.

paf31:
I seem to remember this ought to work and report an error, even in this case. So I think this is a bug.

garyb:
Agreed 👍 my second comment was meant as a clarification on what the bug might be.

MonoidMusician:
Right, I think this warning might be triggered: https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Sugar/TypeClasses/Deriving.hs#L240

But I think it should be raised to an error in this case, where the superclass instance is given but not newtype derived, due to this potential for inconsistency.

Maybe I can see if this can be improved. It might not even be recursive right now.

MonoidMusician:
Okay so it looks like we were pulling in _all_ the instances, so I changed it to match on `NewtypeInstance` ... but that only affects local instances, it doesn't like that information is stored in externs. Do we want to extend the check to externs? or call this good enough?

[...]

PR #4371, opened 2022-07-19
Derive Foldable and Traversable instances

### LINKED ISSUE #2518: Derive Foldable and Traversable

JordanMartinez:
**Description of the change**

Fixes #2518.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_derive-foldable.md
+++ b/CHANGELOG.d/feature_derive-foldable.md
@@ -1,0 +1,88 @@
+* Enable the compiler to derive `Foldable` instances
+
+  The compiler can now automatically derive a `Foldable` instance for types,
+  so long as
+  1. All type constructors that use the type variable being folded over
+     (e.g. `a` below) **must** appear in the rightmost position that type constructor.
+     Any other positions will be considered invalid.
+  2. Any higher-kinded type constructor must have a `Foldable` instance. By implication,
+     this excludes the `Function` type.
+
+  The examples below indicate valid usages (via `✓`) and invalid usages (via `⨯`) of 
+  a type variable `a` used in various places below.
+  If the data constructors were refactored so as to remove the invalid usages,
+  the compiler can derive a `Foldable` instance for it
+
+  ```purs
+  data X f a
+    = X0 Int
+    --         - although no `a` appears in this data constructor
+    --           it doesn't break either of the rules above
+    | X1 a a a
+    --   ✓ ✓ ✓ - because each `a` is in the rightmost position
+    --               due to being a top-level value in the constructor
+    | X2 (f a)
+    --      ✓ - because the `a` is in the rightmost position of `f`
+    | X3 (Tuple a Int)
+    --          ⨯ - because the `a` is not in the rightmost position of `Tuple`
+    | X4 (Tuple a a)
+    --          ⨯ ✓
+    | X5 { foo :: a, bar :: f a, baz :: Tuple Int a }
+    --            ✓           ✓                   ✓ - records are supported
+    | X6 { one :: { two :: { three :: a } } }
+    --                                ✓ - even nested ones
+  ```
+
+  Records are folded in the alphabetical ordering of their labels,
+  not their definition order. For example, a record defined like
+  ```purs
+  type Foo a =
+    { m :: a
+    , f :: a
+    , c :: a
+    , g :: a
+    }
+  ```
+
+  will be folded in a `cfgm` order, not the `mfcg` order.
+
+  Given a data type like the following...
+
+  ```purs
+  data M f a
+    = M0
+    | M1 a (Array a)
+    | M2 Int
+    | M3 (f a)
+    | M4 { a :: a, x :: Array a, fa :: f a
+         , ignore :: Int, no :: Array Int, nope :: f Int
+         , nested :: { anotherOne :: a }
+         }
+
+  derive instance Foldable f => Foldable (M f)
+  ```
+
+  Something like the following will be generated:
+  ```purs
+  instance Foldable f => Foldable (M f) where
+    foldl f z = case _ of
+      M0 -> z
+      M1 a arr -> foldl f (f z a) arr
+      M2 _ -> z
+      M3 fa -> foldl f z fa
+      M4 rec ->
+        foldl f (foldl f (foldl f (f z rec.a) rec.fa) rec.nested.anotherOne) rec.x
+    foldr f z = case _ of
+      M0 -> z
+      M1 a arr -> f a (foldr f z arr)
+      M2 _ -> z
+      M3 fa -> foldr f z fa
+      M4 rec ->
+        f (foldr f (foldr f (f z rec.x) rec.nested.anotherOne) rec.fa) rec.a
+    foldMap f = case _ of
+      M0 -> mempty
+      M1 a arr -> f a <> foldMap f arr
+      M2 _ -> mempty
+      M3 fa -> foldMap f fa
+      M4 rec -> f rec.a <> foldMap f rec.fa <> foldMap f rec.nested.anotherOne <> foldMap rec.x
+  ```
--- a/src/Language/PureScript/TypeChecker/Deriving.hs
+++ b/src/Language/PureScript/TypeChecker/Deriving.hs
@@ -1,20 +1,24 @@
 module Language.PureScript.TypeChecker.Deriving (deriveInstance) where
 
-import Protolude hiding (Type)
+-- import Protolude hiding (Type)
+import Protolude hiding (Type, traceM)
 
 import Control.Monad.Writer.Class (MonadWriter(..))
 import Data.Foldable (foldl1)
 import Data.List (init, last, zipWith3, (!!))
+import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map as M
 
 import Control.Monad.Supply.Class
 import Language.PureScript.AST
 import Language.PureScript.AST.Utils
+import qualified Language.PureScript.Constants.Data.Foldable as Foldable
+import qualified Language.PureScript.Constants.Data.Traversable as Traversable
 import qualified Language.PureScript.Constants.Prelude as Prelude
 import qualified Language.PureScript.Constants.Prim as Prim
 import Language.PureScript.Crash
 import Language.PureScript.Environment
-import Language.PureScript.Errors
+import Language.PureScript.Errors hiding (nonEmpty)
 import Language.PureScript.Label (Label(..))
 import Language.PureScript.Names
 import Language.PureScript.PSString
@@ -69,12 +73,37 @@ deriveInstance instType className strategy = do
           _ -> throwError . errorMessage $ ExpectedTypeConstructor className tys ty
         _ -> throwError . errorMessage $ InvalidDerivedInstance className tys 1
 
+      validateFFTArgs
+        :: ProperName 'TypeName
+        -> (ModuleName -> [(ProperName 'ConstructorName, [Maybe (FftArg ())])] -> m [(PSString, Expr)])
+        -> m [(PSString, Expr)]
+      validateFFTArgs tyConNm f = do
+        (_, _, argTypes, ctors) <- lookupTypeDecl mn tyConNm
+        case reverse argTypes of
+          [] ->
+            throwError . errorMessage $ KindsDoNotUnify (kindType -:> kindType) kindType
+          ((iTy, _) : _) -> do
+            ctors' <- traverse (traverse (traverse replaceAllTypeSynonyms)) ctors
+            case unaryFftCheckCtorValidity iTy ctors' of
+              Left invalidUsages -> do
+                throwError $ fold $ do
+                  iu <- invalidUsages
+                  (withHints, ss) <- sortOn snd $ toList iu
+                  pure
+                    $ addHint (ErrorInInstance className tys)
+                    $ withHints
+                    $ errorMessage $ CannotDeriveInvalidConstructorArg iTy ss
+              Right validCtors -> do
+                f mn validCtors
+
       in case className of
         Prelude.Eq -> unaryClass deriveEq
         Prelude.Eq1 -> unaryClass $ \_ _ -> deriveEq1
-        Prelude.Functor -> unaryClass deriveFunctor
         Prelude.Ord -> unaryClass deriveOrd
         Prelude.Ord1 -> unaryClass $ \_ _ -> deriveOrd1
+        Prelude.Functor -> unaryClass $ \_ tyConNm -> validateFFTArgs tyConNm deriveFunctor
+        Foldable.Foldable -> unaryClass $ \_ tyConNm -> validateFFTArgs tyConNm deriveFoldable
+        Traversable.Traversable -> unaryClass $ \_ tyConNm -> validateFFTArgs tyConNm deriveTraversable
         -- See L.P.Sugar.TypeClasses.Deriving for the classes that can be
         -- derived prior to type checking.
         _ -> throwError . errorMessage $ CannotDerive className tys
@@ -354,10 +383,124 @@ decomposeRec = fmap (sortOn fst) . go
         go (REmptyKinded _ _) = Just []
         go _ = Nothing
 
-decomposeRec' :: SourceType -> [(Label, SourceType)]
-decomposeRec' = sortOn fst . go
-  where go (RCons _ str typ typs) = (str, typ) : go typs
-        go _ = []
+decomposeRecSorted :: SourceType -> [(PSString, SourceType)]
+decomposeRecSorted = sortOn fst . go []
+  where go !acc (RCons _ str typ typs) = go ((runLabel str, typ) : acc) typs
+        go !acc _ = acc
+
+data FftArg a
+  = FftBaseArg a
+  | FftRecursiveArg a (FftArg ())
+  | FftRecordArg (NonEmpty (PSString, FftArg a))
+  deriving (Show, Functor, Foldable, Traversable)
+
+unaryFftCheckCtorValidity
+  :: Text
+  -> [(ProperName 'ConstructorName, [SourceType])]
+  -> Either
+      [NonEmpty (MultipleErrors -> MultipleErrors, SourceSpan)]
+      [(ProperName 'ConstructorName, [Maybe (FftArg ())])]
+unaryFftCheckCtorValidity iTy ctors = do
+  let
+    (ctorLefts, ctorRights) = foldl' (\(ctorLs, ctorRs) (ctor, subterms) -> do
+      let
+        (subLefts, subRights) = foldl' (
+            \(ls, rs) relevance -> case unaryFunctorArgs iTy relevance of
+              Left xs -> (toList xs : ls, rs)
+              Right arg -> (ls, arg : rs)
+          ) ([], []) subterms
+      case nonEmpty $ join $ reverse subLefts of
+        Just subLefts' -> (map (first (map (addHint (ErrorInDataConstructor ctor)))) subLefts' : ctorLs, ctorRs)
+        Nothing -> (ctorLs, (ctor, reverse subRights) : ctorRs)
+      ) ([], []) ctors
+  if null ctorLefts then Right $ reverse ctorRights else Left $ reverse ctorLefts
+
+unaryFunctorArgs
+  :: Text -- ^ The type var for which we're looking (e.g. @c@ in @data Foo a b c = ...@)
+  -> SourceType -- ^ the data constructor arg's type
+  -> Either (NonEmpty (MultipleErrors -> MultipleErrors, SourceSpan)) (Maybe (FftArg ()))
+unaryFunctorArgs iTyName = goTypeNested
+  where
+  goTypeNested :: SourceType -> Either (NonEmpty (MultipleErrors -> MultipleErrors, SourceSpan)) (Maybe (FftArg ()))
+  goTypeNested = \case
+    -- quantifiers
+    (ForAll _ scopedVar _ t _) | scopedVar /= iTyName ->
+      goTypeNested t
+
+    -- constraints
+    (ConstrainedType _ _ t) ->
+      goTypeNested t
+
+    -- bare index type
+    TypeVar _ t | t == iTyName ->
+      pure $ Just $ FftBaseArg ()
+
+    TypeApp _ (TypeConstructor _ Prim.Record) rows -> do
+      let
+        (errors, validRows) = partitionEithers $ decomposeRecSorted rows <&> \next -> do
+          bimap (map (first (map (addHint (ErrorUnderLabel $ fst next))))) sequence $ traverse goTypeNested next
+      case nonEmpty errors of
+        Just errs ->
+          Left $ join errs
+        Nothing ->
+          pure $ map FftRecordArg $ nonEmpty $ catMaybes validRows
+
+    (TypeApp _ tf ta) -> do
+      case nonEmpty $ hasInvalidUsage False False tf of
+        Nothing ->
+          map (map (FftRecursiveArg ())) $ goTypeNested ta
+        Just invalidUsageIndices ->
+          Left invalidUsageIndices
+
+    -- otherwise invalid arg
+    _ -> do
+      pure Nothing
+
+  hasInvalidUsage :: Bool -> Bool -> SourceType -> [(MultipleErrors -> MultipleErrors, SourceSpan)]
+  hasInvalidUsage reportRightmostTyParam isRightmostTyParam = \case
+    (ForAll _ scopedVar _ t _) | scopedVar /= iTyName ->
+      hasInvalidUsage reportRightmostTyParam isRightmostTyParam t
+
+    (ConstrainedType _ _ t) ->
+      hasInvalidUsage reportRightmostTyParam isRightmostTyParam t
+    
+    TypeVar ann t
+      | t == iTyName
+      , isRightmostTyParam
+      , reportRightmostTyParam -> [ (identity, fst ann) ]
+      
+      | t == iTyName -> [ (identity, fst ann) ]
+      
+      | otherwise -> []
+
+    TypeApp _ (TypeConstructor _ Prim.Record) rows -> do
+      decomposeRecSorted rows >>= (hasInvalidUsage reportRightmostTyParam True . snd)
+
+    TypeApp _ tf ta -> do
+      hasInvalidUsage reportRightmostTyParam False tf <> hasInvalidUsage reportRightmostTyParam isRightmostTyParam ta
+
+    _ -> []
+
+unaryMkCtorClauseWith
+  :: forall m
+   . MonadSupply m
+  => ModuleName
+  -> (Expr -> [(Ident, Maybe (FftArg ()))] -> m Expr)
+  -> (ProperName 'ConstructorName, [Maybe (FftArg ())])
+  -> m CaseAlternative
+unaryMkCtorClauseWith mn buildExpr (ctorName, ctorArgs) = do
+  subterms <- traverse (\arg -> (,arg) <$> freshIdent "v") ctorArgs
+  let
+    idents = map fst subterms
+    ctor = mkCtor mn ctorName
+    caseBinder = mkCtorBinder mn ctorName $ map mkBinder idents
+  finalExpr <- buildExpr ctor subterms
+  return (CaseAlternative [caseBinder] (unguarded finalExpr))
+
+usingLamIdent :: forall m. MonadSupply m => (Expr -> m Expr) -> m Expr
+usingLamIdent cb = do
+  x <- freshIdent "v"
+  lam x <$> cb (mkVar x)
 
 deriveFunctor
   :: forall m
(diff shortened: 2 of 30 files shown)


JordanMartinez:
This PR is a WIP. Current test added fails with:
```
compiler
  Passing examples
    'DerivingFoldable.purs' should compile and run without error [✘] (1467ms)

Failures:

  tests/TestCompiler.hs:140:18: 
  1) compiler, Passing examples, 'DerivingFoldable.purs' should compile and run without error
       Error found:
       in module Main
       at tests/purs/passing/DerivingFoldable.purs:17:1 - 17:58 (line 17, column 1 - line 17, column 58)
       
         Could not match type
             
           b0
             
         with type
                   
           a1 -> t2
                   
       
       while applying a function $z4
         of type b0
         to argument $v6
       while checking that expression $z4 $v6
         has type b0
       in value declaration foldableM
       
       where a1 is a rigid type variable
             b0 is a rigid type variable
             t2 is an unknown type
       
  To rerun use: --match "/compiler/Passing examples/'DerivingFoldable.purs' should compile and run without error/"

Randomized with seed 2138508043

Finished in 1.4675 seconds
```

JordanMartinez:
This PR has been updated to also support deriving for `Traversable`.

I haven't added any failing tests for `Traversable` as I did for `Foldable` as I need clarification on whether the tests I've written for `Foldable` are 1) needed and 2) should be written that way. Once I get feedback from that, I can add `Traversable` failing tests following that format.

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -429,3 +433,252 @@ deriveFunctor mn tyConNm = do
...
+    identityVar = mkRef Prelude.identIdentity
+    mapVar = mkRef Prelude.identMap
+    applyVar = mkRef Prelude.identApply
+    pureVar = mkRef Prelude.identPure
+
+    mkSequenceFn
+      :: m Expr
+    mkSequenceFn = do
+      t <- freshIdent "t"
+      pure $ lam t $ App (App traverseVar identityVar) (mkVar t)
```
JordanMartinez, line 599: I'm wondering if this should be eta-reduced.

MonoidMusician: Good question. I would say no, given that there are some known issues with eta-reduced definitions with e.g. recursive typeclass definitions.

tests/purs/passing/DerivingTraversable.purs
```diff
@@ -0,0 +1,53 @@
...
+data M f a
+  = M0
+  | M1 a (Array a)
+  | M2 Int
+  | M3 (f a)
+
+derive instance (Eq1 f, Eq a) => Eq (M f a)
+derive instance Functor f => Functor (M f)
+derive instance Foldable f => Foldable (M f)
+derive instance Traversable f => Traversable (M f)
```
xgrommx, line 20: I think if we derive `Traversable`, should be automatically derived `Functor` and `Foldable`

rhendric:
I'd like to see a few more test cases for this—here are the ones that leap out at me as most likely to be overlooked:
* Positive cases involving records
  * `data Test a = Test { a :: a, b :: Array a, c :: Int }`
  * `data Test a = Test { nested :: { a :: a, b :: Array a, c :: Int } }`
* Negative cases involving the type parameter occurring not in the final position of a type constructor
  * `data Test a = Test (Tuple a Int)`
  * `data Test a = Test (a -> Int)`

`Foldable` in particular needs extra test coverage as it's easy to construct an implementation of the various folds that pass type checking but don't cover all the expected fields (`Functor` and `Traversable` are more likely to fail to type-check if a field isn't handled correctly).

tests/purs/passing/DerivingTraversable.purs
xgrommx: @rhendric just reaction and no some explanation?

garyb: I'd disagree too, because this would be the only place that deriving one class magically makes others derived too - we don't automatically derive `Eq` from `Ord` for instance.

It could be argued that we should do it there too, but I think being fully explicit is better. It might be that there's some reason to write one of these instances by hand. It's true that we could do something to manage that, and only derive if there's no explicit instances, etc. but I think it's just better to keep it simple and only do explicitly what is asked for.

JordanMartinez: I'd rather be explicit.

JordanMartinez:
I have a question about this one:
> data Test a = Test (Tuple a Int)

Technically, a `Foldable` instance for it could be implemented since we don't impose any laws on `Foldable`:
```purs
instance Foldable Test where
  foldl f z _ = z
  foldlr f z _ = z
  foldMap f = mempty
```

I'm guessing the issue with the above implementation is that we expect a value of type `a` to be passed to the `f` function at some point, and since that value is never passed, this instance is "unexpected".

Contrast that with this data type:
```purs
data Tree a
  = Leaf
  | Node (Tree a) a (Tree a)
```
In the above case, the `Leaf` constructor functions the exact same way as the `Test (Tuple a Int)` case above. However, I would say the `Tree` type above is a valid instance. So, I guess the difference is that `Tree` has at least one constructor that passes an `a` value to the `f` function whereas `Test (Tuple a Int)` has 0 such constructors?

JordanMartinez:
Feel free to suggest what the error message should be called. I'm using KindsDoNotUnify temporarily, but perhaps `UnexpectedBehaviorForFoldableInstance` is what the new error should be called?

rhendric:
> I'm guessing the issue with the above implementation is that we expect a value of type `a` to be passed to the `f` function at some point, and since that value is never passed, this instance is "unexpected".

GHC doesn't allow such an instance to be derived, so the main reason this instance is unexpected by me is that I expect PureScript to follow approximately the same algorithm and refuse to derive it. (Well, GHC has a special case for built-in `(,)` tuples. But if you defined your own `data Tuple a b = Tuple a b`, GHC doesn't allow this.)

Deriving `Foldable Tree` isn't a problem because all occurrences of `a` in the constructors of `Tree` are either bare or the final argument to a type constructor (and thus can be handled by requiring a `Foldable` instance for that type constructor—in the case of `Tree`, this is a self-reference, but in the general case this is irrelevant). I would also expect PureScript to make an allowance for `a`'s (bare or otherwise) appearing in records, similar to how GHC gives special attention to `(,)`. But other occurrences of `a` should lead to the derived instance being rejected. `Functor` does this without needing extra checks, because it's impossible to implement `Functor` without making sure that every `a` occurrence is covered one way or another; the same argument holds for `Traversable`. But `Foldable` is so lenient that we need to consider the question explicitly; and it would surprise me if we decided that occurrences of `a` that would never work in a derived `Functor` or `Traversable` instance would be permitted (and silently ignored!) when deriving `Foldable`.

If we wanted to be more expansive, I would argue (from intuition only, not from any laws I know of) that the correct way to allow derived instances for `data Test a = Test (Tuple a Int)` would be via the `Bifunctor`, `Bifoldable`, and `Bitraversable` instances for `Tuple`. I don't think there's a pressing need for this, but I think it would be ‘more correct’ than the trivial `Foldable` instance for `Test`, and disallowing deriving instances for `Test` for now would make it straightforward to add that behavior in the future.

rhendric:
> Feel free to suggest what the error message should be called. I'm using KindsDoNotUnify temporarily, but perhaps `UnexpectedBehaviorForFoldableInstance` is what the new error should be called?

`CannotDeriveInvalidConstructor`? With an error message modeled after GHC's:

```
Constructor ‘Test’ must use the type variable only as the last argument of a data type
```

JordanMartinez:
> 

Is `Text` here in reference to the type constructor or the data constructor? I assume the former?

tests/purs/failing/FoldableInstance5.out
```diff
@@ -0,0 +1,10 @@
+Error found:
+in module [33mFoldableInstance5[0m
+at tests/purs/failing/FoldableInstance5.purs:9:1 - 9:30 (line 9, column 1 - line 9, column 30)
+
+  Type constructor ‘Test' must use the type variable 'a' only as the last argument of a data type
+
+
+See https://github.com/purescript/documentation/blob/master/errors/CannotDeriveInvalidTypeConstructor.md for more information,
+or to contribute content related to this error.
+
```
JordanMartinez, line 10: Should there should be a hint here that clarifies that we were trying to derive an instance for `Foldable` when this error was encountered? The only such hint provided is the source spans.

rhendric:
> > Constructor ‘Test’ must use the type variable only as the last argument of a data type
> 
> Is `Test` here in reference to the type constructor or the data constructor? I assume the former?

Nah, the latter. If the data type has multiple constructors, we'd want to report on the specific one that uses the type variable in an invalid way.

tests/purs/failing/FoldableInstance5.out
rhendric: Yeah, we should include the class. We might want to put the same check in `Functor` and `Traversable`, even though it isn't strictly necessary, because the message could be more helpful than a type mismatch error for code that the user never wrote.

CHANGELOG.d/feature_derive-foldable.md
```diff
@@ -0,0 +1,61 @@
...
+      M0 -> mempty
+      M1 a arr -> f a <> foldMap f arr
+      M2 _ -> mempty
+      M3 fa -> foldMap f fa
+      M4 rec -> f rec.a <> foldMap f rec.x <> foldMap f rec.fa <> foldMap rec.nested.anotherOne
+  -}
+
+  data Nope1 a = Nope1 (Int -> a)
+  data Nope2 a = Nope2 (a -> Int)
+  data Nope3 a = Nope3
```
rhendric, line 51: Not sure if this reflects the current state of the code, but my position is that `Nope3` (a.k.a. `Proxy`) should allow `Functor` (it already does), `Foldable`, and `Traversable` to be derived.

MonoidMusician:
Yeah, I agree that `Functor`, `Foldable`, and `Traversable` should all follow the same rules for deriving, and in particular they should only rely on those classes, not `Bifunctor`, `Bifoldable`, `Bitraversable` just for consistency.

JordanMartinez:
> If the data type has multiple constructors, we'd want to report on the specific one that uses the type variable in an invalid way.

OH! Oh, oh oh, ok, it's clicking now. :laughing: 

And IMO the error shouldn't just report one constructor at a time. It should report all the constructors with an invalid usage.

JordanMartinez:
Another failing case should be this:
```purs
data Foo a = Foo a a
```
because while the right-most usage of `a` is valid, the left-most usage is not. I think this would produce a type checking error anyways, but the error message could be clearer if we caught it here rather than when typechecking.

rhendric:
Eh? No, that should be fine. `foldMap f (Foo a1 a2) = f a1 <> f a2` is what I'd expect to see generated.

It doesn't matter which argument to the *data* constructor in question contains an occurrence of `a`. What matters is, given an argument to the data constructor, whether the *type* of that argument is ‘fine’ by the following recursive definition. Not using `a` at all is fine. A bare `a` is fine. A *type* constructor `TyCon arg1 arg2 ... argN argLast` is fine if `a` doesn't occur in `arg1` through `argN` and if `argLast` is recursively fine. A record is fine if all of its field types are fine. Nothing else is fine (I think—I'm rattling this off quickly). The authoritative source for this is https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/compiler/derive-functor, though replace their use of built-in tuples with our use of built-in records.

JordanMartinez:
Ugh! I need sleep...

I meant `data Foo a = Foo (Tuple a a)`

tests/purs/passing/DerivingTraversable.purs
PureFunctor: I'm not really aware of any language that does deriving implicitly as well

MonoidMusician: @xgrommx Did you mean that they will be automatically derived by the compiler, or did you mean that there should be a check that they also were derived (instead of being hand-written)? If the latter, I think that is a good idea.

This is a related issue that I opened a while ago (and ran into again recently!): https://github.com/purescript/purescript/issues/3168

I think we planned on elevating that particular case to an error, but never did so at a breaking change.

I think it might be worth just adding a global check that the derived instances are compatible. Especially if we support `DerivingVia` later, this could get tricky. I'll consider moving that forward later if we agree that's a good idea.

rhendric: Oh, I don't know about that. It's a real safety problem if superclasses of newtype-derived instances aren't derived because of how the compiler handles newtypes. I don't think there's an equivalent problem for `Functor` and friends; as long as the relevant laws are satisfied, nothing will break with a manually-written `Functor` and a derived `Traversable`.

MonoidMusician: Hm that's true, But a manually-written _`Foldable`_ and derived `Traversable` may disagree on the order of the fields, which would be a problem.

rhendric: In that case the pair of instances would violate the
```purs
foldMap f = unwrap <<< traverse (Const <<< f)
```
law. I still see that as a case where if you write an instance, it's on you to ensure that it's lawful.

(I guess the twist is that this is thought of as a `Traversable` law, and the user is writing the `Foldable` instance. Morally I don't think this changes anything—to my mind, `Foldable` and `Traversable` share joint ownership of that law—but I suppose it is unusual to be worried about implementing an instance so that a subclass's law can be satisfied.)

JordanMartinez:
Ok, tests all pass now (at least locally they do). The new code tries pretty hard to provide a very clear picture as to which subterm causes the issue. I still need to finish the error message, but I'm not familiar with that part of the codebase yet. The information is there, I just need to render it.

After that, I'll need to add a hint about which type class was being derived and probably refactor `Functor` to use this new code.

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -429,3 +434,307 @@ deriveFunctor mn tyConNm = do
...
+        tailResult = goTypeTop iToExpr rest
+        combine h tail = case (h, tail) of
+          (Left _, Right _) -> h
+          (Right _, Left _) -> tail
+          (Right a, Right b) -> Right $ a <> b
+          (Left a, Left b) -> Left $ a <> b
+      case (lblResult, tailResult) of
+        (_, Left _) -> lblResult
+        (Left _, _) -> tailResult
+        (Right a, Right b) -> Right $ combine a b
```
JordanMartinez, line 692: I am quite positive this logic is just wrong.

JordanMartinez:
Just to clarify, does deriving Traversable need to support records? (I assume yes). Or is that support only needed for Foldable? I ask because adding such support would require being able to re-create the record if any of its fields should also be traversed. Something like
```purs
data Test a = CtorName a { used :: a, unused :: Int }

instance Traversable Test where
  traverse f (CtorName a rec) = do
    (\a' used' -> CtorName a' { used: used', unused: rec.used }) <$> travere f a <*> traverse f rec.used
```
That would require adding additional information in the value returned as to how to reconstruct a subterm (if needed).

rhendric:
> I ask because adding such support would require

Is this any different than what derived `Functor` instances for types containing record fields require?

tests/purs/passing/DerivingTraversable.purs
drathier: Is there any reason why we can't add the foldable/traversable law to both type classes and call it a day?

JordanMartinez:
> > I ask because adding such support would require
> 
> Is this any different than what derived `Functor` instances for types containing record fields require?

No, I don't believe so. I had forgotten that the `Functor` deriving accounted for record types. I'll have a look at how that works. 
Also, I've got a bit going on in life this week, so it'll be a few days before I can get back to this again.

JordanMartinez:
I had some time to look at how `Functor`'s deriving works and compare it with my current approach. My approach creates a data structure that stores everything I might need, and then each type class we're deriving can use the values it actually needs to produce its specific expression. AFAICT, this is the approach we should be taking to ensure these unary type classes work the same. 

However, `Functor`'s deriving takes a different approach. It builds the expression it needs for each subterm and then applies the resulting subterm expressions to the initial constructor argument. This approach works because the `map` function is simple: we're only rebuilding the same value but with new subterms. However, it wouldn't work for `Foldable` or `Traversable`.

Another difference is that the `Functor` implementation uses `decomposeRec'`, which sorts on the label rows. The sorting doesn't matter in this type class because a record update value is used. However, sorting in `Foldable` or `Traversable` would be problematic.

Regardless, the data structure I build before using in each type class needs to handle records as a separate value. I'll need to think through how that should work.

JordanMartinez:
The `Functor` deriving hasn't been migrated over to use the newer code because my usage of `ObjectUpdateNested` causes a failure in `infer'` due to that function not having a case for `ObjectUpdateNested`. I can workaround that by using `ObjectUpdate` instead (this is what the current `Functor` deriving uses), but I'll have to do that on another day.

Regardless, I wanted to get what I currently have pushed, so that people know this is still being worked on.

JordanMartinez:
I've updated the code to use `ObjectUpdate` and replaced the old `Functor` deriving with my version. The `DerivingFunctor.purs` test passes. I haven't tested the rest of the test suite.

However, the `DerivingTraversable.purs` test is still failing with this error:
```
  tests/TestCompiler.hs:140:18: 
  1) compiler, Passing examples, 'DerivingTraversable.purs' should compile and run without error
       Error found:
       in module Main
       at tests/purs/passing/DerivingTraversable.purs:34:1 - 34:51 (line 34, column 1 - line 34, column 51)
       
         Value $v9 is undefined.
       
       while checking that expression $v9
         has type a0
       while applying a function $f5
         of type a0 -> m1 b2
         to argument $v9
       
             b2 is a rigid type variable
             m1 is a rigid type variable
       
       See https://github.com/purescript/documentation/blob/master/errors/NameIsUndefined.md for more information,
       

  To rerun use: --match "/compiler/Passing examples/'DerivingTraversable.purs' should compile and run without error/"

Randomized with seed 942314740

Finished in 1.4205 seconds
1 example, 1 failure
```
When I drop all data constructors from the `M` type and add them back in one-by-one, the error is caused by the `M1` data constructor.
```purs
data M f a
 = M0
 | M1 a (Array a)
```
`v9` should be the lambda argument that corresponds to the first subterm of `M1` (i.e. `a`). It's not obvious to me why this is error is occurring.

rhendric:
It looks like you're effectively generating `M1 x y -> (\x' y' -> M x' y') <$> f x' <*> traverse f y'` instead of `M1 x y -> (\x' y' -> M x' y') <$> f x <*> traverse f y`. Everything you're doing inside of the `buildExpr` for `Traversable` uses `ctorArgs'`, which are the lambda args corresponding to the primed identifiers in my example, but you need to go back to the unprimed identifiers when (or before) you do your `foldl'` assembling the `map` and `apply` applications.

tests/purs/passing/DerivingFoldable.purs
```diff
@@ -0,0 +1,84 @@
...
+
+derive instance foldableM :: Foldable f => Foldable (M f)
+
+type MArrStr = M Array String
+
+foldlStr :: forall f. Foldable f => f String -> String
+foldlStr = foldl (<>) ""
+
+foldrStr :: forall f. Foldable f => f String -> String
+foldrStr = foldr (<>) ""
```
rhendric, line 36: Happened to notice this in passing; we should test `foldl` and `foldr` with an operation that is expressly *not* monoidal—something like `\l r -> l <> r <> "."` would do—to verify that they fold in the expected directions! (With that example, `foldl` should produce results like `"ab.c."` and `foldr` results like `"abc.."`.)

JordanMartinez:
> It looks like you're effectively generating `M1 x y -> (\x' y' -> M1 x' y') <$> f x' <*> traverse f y'` instead of `M1 x y -> (\x' y' -> M1 x' y') <$> f x <*> traverse f y`. Everything you're doing inside of the `buildExpr` for `Traversable` uses `ctorArgs'`, which are the lambda args corresponding to the primed identifiers in my example, but you need to go back to the unprimed identifiers when (or before) you do your `foldl'` assembling the `map` and `apply` applications.

Many thanks!

JordanMartinez:
I'm still getting a test failure, but it's no longer the `value is undefined` one from above:
```
Progress 1/2: purescript
compiler
  Passing examples
    'DerivingTraversable.purs' should compile and run without error [✘] (1842ms)

Failures:

  tests/TestCompiler.hs:140:18: 
  1) compiler, Passing examples, 'DerivingTraversable.purs' should compile and run without error
       Error found:
       in module Main
       at tests/purs/passing/DerivingTraversable.purs:34:1 - 34:51 (line 34, column 1 - line 34, column 51)
       
         Type of expression lacks required label a.
       
       while checking that type { nested :: { a :: t4                 
                                            , arrayIgnore :: Array Int
                                            , fIgnore :: t5 Int       
                                            , fa :: t5 t4             
                                            , ignore :: Int           
                                            , zArrayA :: Array t4     
                                            }                         
                                }                                     
         is at least as general as type { a :: t0                 
                                        , arrayIgnore :: Array Int
                                        , fIgnore :: t3 Int       
                                        , fa :: t2                
                                        , ignore :: Int           
                                        , zArrayA :: t1           
                                        }                         
       while checking that expression $v18
         has type { a :: t0                 
                  , arrayIgnore :: Array Int
                  , fIgnore :: t3 Int       
                  , fa :: t2                
                  , ignore :: Int           
                  , zArrayA :: t1           
                  }                         
       
       where t5 is an unknown type
             t4 is an unknown type
             t3 is an unknown type
             t0 is an unknown type
             t1 is an unknown type
             t2 is an unknown type
       
       See https://github.com/purescript/documentation/blob/master/errors/PropertyIsMissing.md for more information,
       

  To rerun use: --match "/compiler/Passing examples/'DerivingTraversable.purs' should compile and run without error/"

Randomized with seed 1701719015

Finished in 1.8426 seconds
1 example, 1 failure
```

JordanMartinez:
Looks like a label isn't being added. Here's the generated expression for M5 using `debugValue`:
```
M5 18 ->
    (\19 20 21 ->
        M5 (18 { nested = 18 { a = 19, zArrayA = 20, fa = 21 } })
                         -- ^ .nested
    )
    <$> f5 18.nested.a
    <*> traverse f5 18.nested.zArrayA
    <*> traverse f5 18.nested.fa
```

JordanMartinez:
Ok, the main tests now pass, but the error message and remaining test failures need to be accounted for. I'm going to work on that later.

JordanMartinez:
I believe the only tests that fail now are the source map tests. I haven't looked into that yet.

I'd like feedback on the error messages (see `FoldableInstance9`). Should I use the current terse syntax to repot which subterm(s) is causing the issue? I was thinking the `documentation` repo's corresponding page could explain how to read it. If not, I could use words (e.g. `subterm 0, field "foo", subterm 4, field "bar"`, but that will take up a lot more room on the screen.

I'd also like clarification on the `FoldableInstance9`'s usage of type synonyms. Tests 2, 3, and 4 inline the type alias `Rec` whereas Tests 5, 6, and 7 use the alias. You'll notice that the error message for `FoldableInstance9` includes messages for 2 and 5, but not 3, 4, 6, and 7. If I drop Tests 1 and 2, I'll still get a compiler error due to `Rec` not having a `Foldable` instance. I believe that is ok, but did want to clarify.

JordanMartinez:
Looks like all normal test pass. The CI failure is caused by compiling the package set. For example:
```
 Error 5 of 7:
  
    in module Halogen.HTML.Properties
    at .spago/halogen/v7.0.0/src/Halogen/HTML/Properties.purs:104:1 - 104:50 (line 104, column 1 - line 104, column 50)
  
      Could not match type
  
        a4
  
      with type
  
        b1
  
  
    while trying to match type IProp t2 t3
      with type t0 b1
    while checking that expression IProp $v5
      has type t0 b1
    in value declaration functorIProp
  
    where a4 is a rigid type variable
          b1 is a rigid type variable
          t0 is an unknown type
          t2 is an unknown type
          t3 is an unknown type
  
    or to contribute content related to this error
```
which is caused [by this Functor derivation involving rows](https://github.com/purescript-halogen/purescript-halogen/blob/master/src/Halogen/HTML/Properties.purs#L104)

JordanMartinez:
Full list of errors as found by package set:
- https://github.com/natefaubion/purescript-argparse-basic/blob/main/src/ArgParse/Basic.purs#L155
- https://github.com/ConnorDillon/purescript-channel/blob/v1.0.0/src/Concurrent/Channel.purs#L62
- https://github.com/natefaubion/purescript-dodo-printer/blob/master/src/Dodo/Internal.purs#L42
- https://github.com/paf31/purescript-behaviors/blob/v7.0.0/src/FRP/Event/Semantic.purs#L126
- https://github.com/purescript-halogen/purescript-halogen/blob/master/src/Halogen/HTML/Properties.purs#L104
- https://github.com/natefaubion/purescript-routing-duplex/blob/v0.4.1/src/Routing/Duplex/Parser.purs#L72
- https://github.com/ajnsit/purescript-supply/blob/v0.1.0/src/Supply.purs#L23

JordanMartinez:
I'm going to put this into a draft state until it's ready to be reviewed. I'm still figuring out the exact type definitions I need to support functions correctly and still have correct error messages.

JordanMartinez:
Just another update. I'm pretty sure I got Function properly implemented for deriving `Functor`, but it's hard to write tests that are exhaustive in checking every case. I think there's a way to unify the `Top` and `Sub` args, but I'll do that after getting Foldable and Traversable to work again.

tests/purs/passing/DerivingFunctor.purs
```diff
@@ -69,6 +69,35 @@ maTests = do
...
+  assert' "map - Fun1" do
+    let
+      fn = show
+      left a b = a + b
+      right a b = fn $ left a b
+      Fun1 left' = map fn $ Fun1 left
+    left' 1 2 == right 1 2
+
+data Fun2 a = Fun2 ((a -> a -> Int) -> Int)
+derive instance Functor Fun2
```
rhendric, line 85: Ah, right, the double-contravariant special handling of functions.

I'm not hard against having this, but I will note the following:
* `Functor` deriving has gotten along without it so far
* It has nothing to do with `Foldable` and `Traversable`
* Once we add it, removing or possibly changing it would be a breaking change

Why would we remove or change it? Really, there's no reason `Function` should be special in this regard; one `Profunctor` is as valid as another. The general case might entail expanding `Functor`-deriving to any arguments of types that implement `Functor`, `Bifunctor`, `Contravariant`, or `Profunctor`, with appropriate handling of variance and appropriate positioning of the argument.

But I figure that's probably a distraction from this PR, so I'm wondering if it's better to keep this PR simple(r) and leave `Functor` deriving alone for now.

JordanMartinez: Ah, good points. That will definitely make the tests easier to write.

JordanMartinez:
Ok, the last part should be straightforward (enabling Foldable/Traversable), but I ran out of time for today.

JordanMartinez:
This still isn't complete, but I'm nearly there. Just pushing my latest commits so people know this hasn't been dropped completely.

I think the work left is:
- Fixing the bug where `Foldable` on a `a -> Int` type should fail to compile but it currently compiles
- Fixing the bug where `Traversable` on a data constructor like `M7 (f (f { nested :: RecordFields f a }))` currently fails to compile
- Probably re-examining the errors to ensure they make sense still.

JordanMartinez:
Here's the current expression returned for the `M7` if anyone wants to look at it:
```
traverse vf5 = case _ of
  ... -> ...
  (M7 $v51) -> (map (\$v62 ->                                                                                                                                                                                           
                           M7 $v62                                                                                                                                                                                          
                        )                                                                                                                                                                                                   
                   )                                                                                                                                                                                                        
                   ((traverse (\$v52 ->                                                                                                                                                                                     
                                 (traverse (\$v53 ->                                                                                                                                                                        
                                              (map (\$v59 ->                                                                                                                                                                
                                                      \$v60 ->                                                                                                                                                              
                                                        \$v61 ->                                                                                                                                                            
                                                          $v53 { nested = ($v53.nested) { a = $v59                                                                                                                          
                                                                                        , zArrayA = $v60                                                                                                                    
                                                                                        , fa = $v61                                                                                                                         
                                                                                        }                                                                                                                                   
                                                               }                                                                                                                                                            
                                                   )                                                                                                                                                                        
                                              )                                                                                                                                                                             
                                              ((apply ((apply ((map (\$v56 ->                                                                                                                                               
                                                                       \$v57 ->                                                                                                                                             
                                                                         \$v58 ->                                                                                                                                           
                                                                           ($v53.nested) { a = $v56                                                                                                                         
                                                                                         , zArrayA = $v57                                                                                                                   
                                                                                         , fa = $v58                                                                                                                        
                                                                                         }                                                                                                                                  
                                                                    )                                                                                                                                                       
                                                               )                                                                                                                                                            
                                                               ($f5 (($v53.nested).a))                                                                                                                                      
                                                              )                                                                                                                                                             
                                                       )                                                                                                                                                                    
                                                       ((traverse (\$v54 ->                                                                                                                                                 
                                                                     $f5 $v54                                                                                                                                               
                                                                  )                                                                                                                                                         
                                                        )                                                                                                                                                                   
                                                        (($v53.nested).zArrayA)                                                                                                                                             
                                                       )                                                                                                                                                                    
                                                      )                                                                                                                                                                     
                                               )                                                                                                                                                                            
                                               ((traverse (\$v55 ->                                                                                                                                                         
                                                             $f5 $v55                                                                                                                                                       
                                                          )                                                                                                                                                                 
                                                )                                                                                                                                                                           
                                                (($v53.nested).fa)                                                                                                                                                          
                                               )                                                                                                                                                                            
                                              )                                                                                                                                                                             
                                           )                                                                                                                                                                                
                                 )                                                                                                                                                                                          
                                 $v52                                                                                                                                                                                       
                              )                                                                                                                                                                                             
                    )                                                                                                                                                                                                       
                    $v51                                                                                                                                                                                                    
                   )
```

rhendric:
Well, I can see where that generated code is incorrect, but I'm having some trouble getting through your implementation to find what's causing the problem. With all gratitude for the amount of effort you've put into this so far, it looks to me like you may have managed to substantially overcomplicate things—or perhaps I just don't appreciate the source of the complexity.

As far as `Traversable` goes, here's roughly what I'd expect to see for that `M7` constructor (some lambdas floated for readability; that isn't literally necessary):

```purs
traverse f (M7 v0) = map (\v1 -> M7 v1) (traverse (traverse traverseRecord1) v0)
  where
  traverseRecord1 v2 =
    map (\nested' -> v2 { nested = nested' }) (traverseRecord2 v2.nested)
    --   ^ this map lambda is what the generated code gets wrong
  traverseRecord2 v3 =
    apply (
      apply (
        map (\a' zArrayA' fa' -> v3 { a = a', zArrayA = zArrayA', fa = fa' }) (f v3.a)
      ) (traverse f v3.zArrayA)
    ) (traverse f v3.fa)
```

Conceptually, there's a core loop being demonstrated here that's pretty simple (in slightly-pseudo-Haskell):
```haskell
handleOne :: MonadSupply m => FFTArg -> m Expr
handleOne = \case
  BaseArg -> pure fVar
  RecursiveArg innerArg -> App traverseVar <$> handleOne innerArg
  RecordArg innerFields -> do
    recordVar <- freshVar
    recordLambda <- makeRecordLambda recordVar innerFields
    lam recordVar <$>
      handleMany (flip Accessor recordVar) recordLambda innerFields

handleMany :: MonadSupply m => (a -> Expr) -> Expr -> [(a, FFTArg)] -> m Expr
handleMany readKey innerLambda fields = do
  (field1:moreFields) <- for fields $ \(key, fftArg) ->
    flip App (readKey key) <$> handleOne fftArg
  foldl' (App applyVar) (App (App mapVar innerLambda) field1) moreFields
```

And a top-level handler for constructors that's similar to the above record case:
```haskell
handleCtor :: MonadSupply m => ProperName 'Constructor -> [(Expr, Maybe FFTArg)] -> m Expr
handleCtor ctor ctorFields = do
  ctorLambda <- makeCtorLambda ctor ctorFields
  handleMany id ctorLambda (mapMaybe sequence ctorFields)
```

I realize that there are important details I'm omitting, and that you're pulling bits and pieces of this loop apart presumably so that you can do early validation and generate error messages and whatnot, but the result is such that I can't quite pin down where this core logic lives in your implementation. `traverseRecursiveArg` looks like the most likely analog, but there are several other local functions in the vicinity that look like they do very similar things and I don't understand what each one is for—maybe you can explain why there are this many moving parts?

---

Along those lines, as early feedback on the PR, I would expect to see a lot more code reuse between the three classes being handled here. The `Functor` and `Foldable` implementations for `M7` are, up to some optimizations:

```purs
map f (M7 v0) = (\v1 -> M7 v1) (map (map mapRecord1) v0)
  where
  mapRecord1 v2 =
    (\nested' -> v2 { nested = nested' }) (mapRecord2 v2.nested)
  mapRecord2 v3 =
    (
      (
        (\a' zArrayA' fa' -> v3 { a = a', zArrayA = zArrayA', fa = fa' }) (f v3.a)
      ) (map f v3.zArrayA)
    ) (map f v3.fa)

foldMap f (M7 v0) = (foldMap (foldMap foldMapRecord1) v0)
  where
  foldMapRecord1 v2 =
    (foldMapRecord2 v2.nested)
  foldMapRecord2 v3 =
    append (
      append (
        (f v3.a)
      ) (foldMap f v3.zArrayA)
    ) (foldMap f v3.fa)
```

The similarities here are hopefully strong enough that you can see how to reuse most of the `Traversable` implementation to generate these as well, with most of the changes living in `handleMany`. For `map`, there would also ideally be some inlining of those lambdas, but that shouldn't add too much complication. `foldl` and `foldr` are probably still worth handling separately, though.

tests/purs/failing/FoldableInstance9.out
```diff
@@ -0,0 +1,42 @@
...
+
+    Constructor, Test2 has invalid 'a' usage at path(s):
+
+      _0."all"_0
+      _0."all"_1
+
+    Constructor, Test5 has invalid 'a' usage at path(s):
+
+      _0."all"_0
+      _0."all"_1
```
rhendric, line 24: Two things: first, this error message misses quite a lot of invalid occurrences of `a`. Every single constructor has at least one, and there appear to be only a handful of *valid* occurrences in the entire declaration.

Second: the following approach is unprecedented in our error messages, but then again so is this path-based presentation so it seems we want an innovation either way. What if we aimed for something like this:

```
The following occurrences of 'a' prevent 'Foldable (Test f g h)' from being derived:

  16:  = Test1 (f a a a) (f Int a a) (f a a Int) (f Int a Int) (f Int Int Int)
                  ^ ^           ^       ^ ^             ^
  17:  | Test2 { all :: f a a a
                          ^ ^
  18:          , rights :: f Int a a
                                 ^
  19:          , lefts :: f a a Int
                            ^ ^
  20:          , middle :: f Int a Int
                                 ^
```

The biggest downside might be that if these messages are presented in a proportional-width font, they'd be next to useless. Maybe we could use a different format in the IDE interface for that reason.

This is something that could be done as a later refinement PR, so don't let it hold things up if it's a controversial idea.

JordanMartinez: > Every single constructor has at least one, and there appear to be only a handful of valid occurrences in the entire declaration.

I think that's only the case because I haven't pushed an update for this test. Locally, there's more things listed.

JordanMartinez: > Second: the following approach is unprecedented in our error messages, but then again so is this path-based presentation so it seems we want an innovation either way. What if we aimed for something like this:

Yeah... I've been wondering the same thing. The path-based representation does give you all the info you need in a succinct way, but it's not readable unless someone clicks on the link to the documentation repo (a page which doesn't exist yet).

I think the format you've shown above is what I'd prefer personally (though it means more work :smile:). While there may be some situations where the formatting isn't good, I think that will happen infrequently. So overall, a better tradeoff than the path-based one.

JordanMartinez:
> Well, I can see where that generated code is incorrect, but I'm having some trouble getting through your implementation to find what's causing the problem. With all gratitude for the amount of effort you've put into this so far, it looks to me like you may have managed to substantially overcomplicate things—or perhaps I just don't appreciate the source of the complexity.

I'll try to respond to the above and the rest of what you said in the rest of that comment sometime in the next few days. I have a number of life events going on right now that is taking a lot of my free time, so it might still be a while. I appreciate the feedback you've given though, and I'll try to understand it more when I can.

JordanMartinez:
I wasn't able to fall asleep tonight for quite some time, so I thought I'd try to respond to your comment.

> With all gratitude for the amount of effort you've put into this so far, it looks to me like you may have managed to substantially overcomplicate things—or perhaps I just don't appreciate the source of the complexity.

:smile: I wouldn't be surprised if I overcomplicated this. For context, I tried writing the Traverse deriving function 3-4 times before I finally got to this design as I was having difficulty figuring out what the type's definition should be for `TraversableArg`. I was confident there's a way to write the code such that `FoldableArg` could be used instead of `TraversableTopArg` and `TraversableArg`, but I had't figured that out yet. At various points in my initial attempts, I thought I needed an `Ident` in one place and an `Expr` in another, only to realize later that it was the opposite or not quite even that. I quickly became confused.

Regardless, thanks to your feedback, I think I see what's the source of the issue.

> As far as Traversable goes, here's roughly what I'd expect to see for that M7 constructor (some lambdas floated for readability; that isn't literally necessary):

I think this is the cause of my confusion and overcomplication. The code you expected `M7` to produce differs from what I expected it to produce, and I'd guess that's the cause for the bug.

If I had a record like the following:
```purs
type R f a = { nested :: { a1 :: a, fa1 :: f a }, a2 :: a, fa2 :: f a }
```
I would write the following where nested records are "denested" but you would write the following where nested records are treated more like layers of Tuples.
```purs
-- Jordan's
traverse f r =
  (\a1' fa1' a2' fa2' -> r { nested = { a1 = a1', fa1 = fa1' }, a2 = a2', fa2 = fa2' })
    <$> f r.nested.a1
    <*> traverse f r.nested.fa1
    <*> f r.a2
    <*> traverse f r.fa2
    
-- Ryan's
traverse f r =
  (\nested' a2' fa2' -> r { nested = nested', a2 = a2', fa2 = fa2' })
    <$> ((\a1' fa1' -> r.nested { a1 = a1', fa1 = fa1' }) <$> f r.nested.a1 <*> traverse f r.nested.fa1)
    <*> f r.a2
    <*> traverse f r.fa2
```

> the result is such that I can't quite pin down where this core logic lives in your implementation. traverseRecursiveArg looks like the most likely analog, but there are several other local functions in the vicinity that look like they do very similar things and I don't understand what each one is for—maybe you can explain why there are this many moving parts?

The "denesting" I do and the lack thereof in your example are why I had 2 different functions: `traverseTopArg`/`traverseFieldArg` and `traverseRecursiveArg`. `traverseTopArg` and `traverseFieldArg` are largely the same aside from the former returning a `NonEmpty Expr` whereas the latter returns just `Expr`. And the `NonEmpty Expr` return type was only needed because I was "denesting" nested record args whereas you aren't.

I think my "denesting" approach may happen in the Functor implementation, too. So, that's something I'd need to fix as well.

tests/purs/failing/FoldableInstance9.out
JordanMartinez: Another way to workaround the proportional font issue is to color the problematic `a` usages a different color in the error message. Most terminals used by most users support colors.

rhendric:
> I would write the following where nested records are "denested" but you would write the following where nested records are treated more like layers of Tuples.

Ah, interesting! That does clear some things up, thanks. I don't actually think that's wrong as such, and I'm not against adding it as a refinement if you prefer it (it does seem like it'd be a little more efficient in the cases when it's relevant), both for `Functor` and `Traversable`—my advice would be get the nested version tidied up and working first though. The structure of the implementation shouldn't need to change to accommodate denesting; in my sketch, it would just affect the lambda creation functions and `handleMany`.

JordanMartinez:
Looks like I fixed the bug, but this just revealed an issue in `Foldable`. I basically need to do this:
```
foldl
  (\acc next ->
    <if next == base arg>
      f acc next
    <if next == recursive arg>
      f acc next
    <if next == record arg>
    f (f (f acc record.label1) record.label2) record.label3
  )
  z
  arg
```
and this part:
```
    <if next == record arg>
    f (f (f acc record.label1) record.label2) record.label3
```
is what I'm currently not doing.

JordanMartinez:
All tests passed locally for me now. So, I believe this actually works now, but there's a lot of room for refactoring. As such, I'm going to mark this as ready for review.

tests/purs/failing/FoldableInstance4.out
```diff
@@ -0,0 +1,20 @@
+Error found:
+in module [33mFoldableInstance4[0m
+at tests/purs/failing/FoldableInstance4.purs:8:1 - 8:27 (line 8, column 1 - line 8, column 27)
+
+  The following data constructors and their subterms must use the type variable 'a' only as the last argument of a data type:
+
+    Constructor, T has invalid 'a' usage at path(s):
+
+      _0\out <functions are not supported for this type class>
```
rhendric, line 9: I don't think we should have a `Function`-specific error message for this. `Function t` is not `Foldable`, so this should be the same message that any other uses-`a`-but-not-`Foldable` type appearing in the `T` constructor would produce. (If that error message is inadequate, let's fix it for all not-`Foldable` types.)

tests/purs/failing/FoldableInstance9.out
```diff
@@ -0,0 +1,50 @@
+Error found:
+in module [33mFoldableInstance9[0m
+at tests/purs/failing/FoldableInstance9.purs:53:1 - 53:38 (line 53, column 1 - line 53, column 38)
+
+  The following data constructors and their subterms must use the type variable 'a' only as the last argument of a data type:
```
rhendric, line 5: This list is still incomplete; all the occurrences of `a` in `Test3`, `Test4`, `Test6`, and `Test7` are issues except for the final one in each constructor.

tests/purs/passing/DerivingFunctor.purs
```diff
@@ -3,34 +3,124 @@ module Main where
...
 import Effect.Console (log)
+import Data.List (List(..), (:))
+import Data.Tuple (Tuple(..))
 import Test.Assert
 
-type MyRecord a = { myField :: a }
+type RecordFields f a =
+  { a :: a
+  -- ensure original field order (not order based on label)
+  -- is order used in fold
```
rhendric, line 13: This comment's a little out of place here, since order should not be observable by any `map` operation.

rhendric (changes requested):
This is feedback on the tests only; seems a little pointless to go deep on the implementation if you plan to refactor it heavily.

JordanMartinez:
> seems a little pointless to go deep on the implementation if you plan to refactor it heavily.

Yeah... Latest commits refactor a few things like dropping unneeded lamdas. None of the commits account for the test feedback you provided. I'll get to that when I can, but thanks nonetheless!

JordanMartinez:
Notes from a talk with Nate about this:
- record label need to sorted alphabetically (it's what we do everywhere else)
- rather than path-based error-messages, could just print source spans of each invalid usage. Since spans are for types, desugaring pass probably doesn't mess them up. (e.g. `at source span <span>, invalid usage of type variable 'a'`).
- unifying the Functor/Traversable implementations since Traversable is just Functor with affects (same idea as what Ryan, I believe, was stating above)

JordanMartinez:
Ok, this is ready for a more in-depth review. I made the following changes in the latest commits:
- record labels are ordered alphabetically
- error messages now use the source spans of the type with invalid usage rather than the path-based representation. Moreover, one error is thrown for every invalid usage.
- unnecessary lambdas were removed
- I believe the invalid usages that were missing (as noted by Ryan above) have been fixed, but I haven't verified this myself yet

I am not sure whether/how to unify the `Functor`/`Traversable` implementation because the two handle record args differently. So, feedback on that would be appreciated.

The changelog entries likely need to be updated to account for more things, but I'll have to get to that another time.

tests/purs/failing/FoldableInstance4.out
JordanMartinez: I'm still treating Functions specially, but that is something that can easily be changed with the `InvalidConstructorArgReason` type. Do you know of other examples where we have an error like that, so I can reuse its phrasing?

JordanMartinez:
Oh... I should also add a module-level comment that helps describe the high-level overview of the implementations.

rhendric:
> * error messages now use the source spans of the type with invalid usage rather than the path-based representation. Moreover, one error is thrown for every invalid usage.

The downside of this is that it makes it look like the error is in the type declaration, which is valid in isolation. I think it's the right idea to capture source spans and present them in the error message, but not to use them as the primary source span of the error, if that makes sense?

Also, there needs to be some deduplication here. This error:
```
  in module [33mFoldableInstance9[0m
  at tests/purs/failing/FoldableInstance9.purs:117:18 - 117:19 (line 117, column 18 - line 117, column 19)

    The type variable `a` must only be used as the last argument in a data type.

  in data constructor [33mTest5[0m
  in type class instance
  [33m                                        [0m
  [33m  Data.Foldable.Foldable (Test f0 g1 h2)[0m
  [33m                                        [0m

  where [33mf0[0m is a rigid type variable
        [33mg1[0m is a rigid type variable
        [33mh2[0m is a rigid type variable

  See https://github.com/purescript/documentation/blob/master/errors/CannotDeriveInvalidConstructorArg.md for more information,
```
occurs several times, presumably because there are several invalid uses of `a` in the definition of `Rec`.



> I am not sure whether/how to unify the `Functor`/`Traversable` implementation because the two handle record args differently. So, feedback on that would be appreciated. I'm also worried that such changes might make the code less readable.

IMO, unifying all three of these classes (possibly with the exception of `foldl`/`foldr` implementations, if you don't want to use the defaults for these) would make the whole thing vastly more readable. Why do you believe you need to handle record args differently in those cases?

tests/purs/failing/FoldableInstance4.out
rhendric: We might be talking past each other. Either:
* Don't handle functions specially. The compiler will eventually raise a `NoInstanceFound` when it tries to `foldMap` or `traverse` over a function, with a message along the lines of ‘No type class instance was found for Foldable (Function t)’, which IMO is a fine error message for that case and matches what happens if any other constructed type that isn't `Foldable`/`Traversable` is used incorrectly.
* Don't handle functions specially, and check up front that any constructed types that need to be folded/traversed over implement the right class. If they don't, the message used for that sort of `InvalidConstructorArgReason` should be similar to the message for `NoInstanceFound`, with whatever additional information would make this option a better choice than the first.

JordanMartinez: Ah... I see now. I forgot that `Function` is subsumed by the recursive arg handling since we're no longer supporting contravariant positions.

JordanMartinez:
> I think it's the right idea to capture source spans and present them in the error message, but not to use them as the primary source span of the error, if that makes sense?

Agreed. By using the source span as the error's position, my hope was to see the red squiggles in the IDE, so that it's that much more obvious as to which `a` is problematic. However, it would be confusing in the way you mentioned.

The error messages now print the file and location of the invalid usage. While this is clearer, the user has to examine where exactly that position is in the source code.

> Also, there needs to be some deduplication here.

I looked into this. I believe it occurs due to the following. Given this data constructor:
```purs
Test5 (Rec f a)
```
I believe `Rec f a` gets replaced with its alias...
```purs
  { all :: f a a a
  , rights :: f Int a a
  , lefts :: f a a Int
  , middle :: f Int a Int
  , none :: f Int Int Int
  }
```
... but the source spans for `a` still refers to
```
Rec f a
      ^
```
for every `a` in its alias. So, every time `Rec f a` appears, the same source span is duplicated 8 times. The correct report should say something like this:
```diff
Error 38 of 68:

  in module [33mFoldableInstance9[0m
  at tests/purs/failing/FoldableInstance9.purs:120:1 - 120:38 (line 120, column 1 - line 120, column 38)

    The type variable `a` must only be used as the last argument in a data type:

      tests/purs/failing/FoldableInstance9.purs:76:14 - 76:15 (line 76, column 14 - line 76, column 15)


+ in type alias [33mRec f a5[0m
  in data constructor [33mTest5[0m
  in type class instance
  [33m                                        [0m
  [33m  Data.Foldable.Foldable (Test f0 g1 h2)[0m
  [33m                                        [0m

  where [33mf0[0m is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)
        [33mg1[0m is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)
        [33mh2[0m is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)

  See https://github.com/purescript/documentation/blob/master/errors/CannotDeriveInvalidConstructorArg.md for more information,
  or to contribute content related to this error.
```

And on that note, perhaps using hints would be a better way to communicate the "path" to the type variable?
```diff
Error 38 of 68:

  in module [33mFoldableInstance9[0m
  at tests/purs/failing/FoldableInstance9.purs:120:1 - 120:38 (line 120, column 1 - line 120, column 38)

    The type variable `a` must only be used as the last argument in a data type:

      tests/purs/failing/FoldableInstance9.purs:76:14 - 76:15 (line 76, column 14 - line 76, column 15)

+ in label "all"
+ in type alias [33mRec f a5[0m
  in data constructor [33mTest5[0m
  in type class instance
  [33m                                        [0m
  [33m  Data.Foldable.Foldable (Test f0 g1 h2)[0m
  [33m                                        [0m

  where [33mf0[0m is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)
        [33mg1[0m is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)
        [33mh2[0m is a rigid type variable
          bound at (line 0, column 0 - line 0, column 0)

  See https://github.com/purescript/documentation/blob/master/errors/CannotDeriveInvalidConstructorArg.md for more information,
  or to contribute content related to this error.
```

JordanMartinez:
> Why do you believe you need to handle record args differently in those cases?

I thought it would be more efficient to denest records like that when implementing the `Traversable` implementations. 

> IMO, unifying all three of these classes (possibly with the exception of `foldl`/`foldr` implementations, if you don't want to use the defaults for these) would make the whole thing vastly more readable.

I think I'll look at figuring this out on another day.

rhendric:
Yeah, let's find a way to collab on this then. Practically speaking, I opened my feedback on this refactoring because I expected it would lead to a smaller and simpler patch, and it would then be easier for me to give a review and for us to get in sync on any remaining points of contention. If I stepped in to do the simplification, would you then still have the emotional capacity to (A) review that and make sure I'm not crawling up my own abstractions, and (B) resume our original roles of implementer/reviewer until the PR is merged? (I don't anticipate any more large requests.) Or are you looking to tap out entirely and you want someone else to own the effort?

rhendric:
(BTW, I ask about point (B) because, having worked on this for two months, I'm sure you've built a lot of valuable context in your head in addition to the work you've committed to Git, and I don't want to lose the benefit of that! It's not that I'm trying to get you to do work I'm unwilling to do myself.)

JordanMartinez:
> Practically speaking, I opened my feedback on this refactoring because I expected it would lead to a smaller and simpler patch, and it would then be easier for me to give a review and for us to get in sync on any remaining points of contention.

Makes sense! I see it as you trying to be proactive without also stepping on toes while I refactored stuff.

> (BTW, I ask about point (B) because, having worked on this for two months, I'm sure you've built a lot of valuable context in your head in addition to the work you've committed to Git, and I don't want to lose the benefit of that! It's not that I'm trying to get you to do work I'm unwilling to do myself.)

Honestly, the reason why this has taken so long was because I didn't have spare free time, not because it took that long to think through problems. And in general, when I realized that my initial attempts were failing because I was trying to implement both `Functor`, `Foldable`, and `Traversable` all at once, that's when I broke each out into its own subproblem, solved each, and then unified their types and solutions into a more similar type/implementation. So, I think there's less context there than you may think. Your insight about removing the unneeded lambda expression is actually what simplified a number of the implementations.

> If I stepped in to do the simplification, would you then still have the emotional capacity to (A) review that and make sure I'm not crawling up my own abstractions, and (B) resume our original roles of implementer/reviewer until the PR is merged? (I don't anticipate any more large requests.) Or are you looking to tap out entirely and you want someone else to own the effort?

Since I don't think there's much left to do in this PR, I think I'd prefer tapping out entirely. There's a few other things (e.g. fixing Pursuit, updating my own learning repo, etc.) that I haven't been able to get to because I keep focusing on this. However, if that's not possible, then I would be ok with the A/B approach.

JordanMartinez:
Superceded by #4392

PR #4374, opened 2022-07-27
Fix replicated type hole suggestions

### LINKED ISSUE #4373: Type hole suggestions are repeated
## Description

Suggestions in type holes are repeated, like so:
```purs
module Main where

data F a = F ((a -> Int) -> Int)

map :: forall a b. a -> (a -> b) -> F a -> F b
map a f (F g) = ?help
```

```
[1/1 HoleInferredType] /Users/justin/PureFunctor/haskell/purescript/tmp/Main.purs:6:17

  6  map a f (F g) = ?help
                     ^^^^^
  
  Hole 'help' has the inferred type
  
    F b0
  
  in the following context:
  
    a :: a1
    f :: a1 -> b0
    a :: a1
    f :: a1 -> b0
    g :: (a1 -> Int) -> Int
```

## Expected behavior

This should not be the case

## Additional context

I don't think this necessarily affects 0.15.2 or earlier, so I have a hunch that my qualification PR introduced it or so.

## PureScript version

0.15.4

PureFunctor:
**Description of the change**

Fixes #4373. This might also be related to some performance regression, because it adds redundant entries to the type-checking environment.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/fix_4373.md
+++ b/CHANGELOG.d/fix_4373.md
@@ -1,0 +1,15 @@
+* Fix replicated type hole suggestions due to malformed source spans
+
+  In PureScript `0.15.4`, the following code will produce multiple entries in
+  the type hole suggestions. This is due to malformed source spans that are
+  generated when desugaring value declarations into case expressions.
+
+  ```purs
+  module Main where
+
+  data F = X | Y
+
+  f :: forall a. F -> a -> a
+  f X b = ?help
+  f Y b = ?help
+  ```
--- a/src/Language/PureScript/Sugar/CaseDeclarations.hs
+++ b/src/Language/PureScript/Sugar/CaseDeclarations.hs
@@ -380,10 +380,10 @@ makeCaseDeclaration ss ident alternatives = do
       argNames = foldl1 resolveNames namedArgs
   args <- if allUnique (catMaybes argNames)
             then mapM argName argNames
-            else replicateM (length argNames) freshIdent'
-  let vars = map (Var ss . Qualified ByNullSourcePos) args
+            else replicateM (length argNames) ((nullSourceSpan, ) <$> freshIdent')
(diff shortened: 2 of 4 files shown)


MonoidMusician:
Looks fine at first glance, but I'm not up to speed on how this helps. Can you explain how this fixes the duplicates? It isn't immediately clear to me from this diff, so it must be with the inserting into the environment. Why does it only affect variables – because those will be the only suggestions that show up?

PureFunctor:
I've found that the `CaseDeclaration` sugar effectively mangles the source spans for variable binders. What happens is that something like:
```purs
f x y = ...
```
Gets turned into:
```purs
f = \x -> \y -> ...
```
and this generates the correct spans for the variable binders. If you introduce a constructor pattern, it has to expand to a pattern match:
```purs
f X y = ...

f = \$0 -> case $0 of X -> \y -> ...
```
The newly generated `Abs (VarBinder ..) ..` for `y` uses the source span of the entire declaration instead of old spans, which is propagated in the `desugarImports` step causing the duplication.

MonoidMusician:
Ohh I think I see now. The same variable name appears both in the abstraction and in the case switch (like `\x -> case x of x -> …`), and before they were considered distinct with distinct source spans, but this patch gives them the same source span so it only appears once. Is that right?

PureFunctor:
Yup, so in my original example, the declaration expands to this:
```purs
\a ->                     
  \f ->                   
    \$0 ->                
      case a f $0 of      
        a f (F g) -> ?help
```
The source positions for the variable binder for the `Abs` and the match arms are distinct, so when the renamer collects this information, it accidentally duplicates them!

garyb (approved):
I wasn't quite sure about what was going on when I looked at this earlier, but makes more sense with the explanations. A bit tricky!

MonoidMusician (approved):

PR #4375, opened 2022-07-31
Allow parentheses around operators when defining aliases

PureFunctor:
**Description of the change**

This makes it possible to wrap operators in parentheses when defining aliases.

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_parenthesis_in_infix.md
+++ b/CHANGELOG.d/feature_parenthesis_in_infix.md
@@ -1,0 +1,11 @@
+* Allow parentheses around operators when defining aliases
+
+  The compiler now allows operators in alias declarations to be wrapped in parentheses,
+  similar to how custom operators would be defined in languages like Haskell or OCaml.
+
+  ```purs
+  pairUp :: forall a b. a -> b -> { a :: a, b :: b }
+  pairUp a b = { a, b }
+
+  infixr 4 pairUp as (/\)
+  ```
--- a/src/Language/PureScript/CST/Parser.y
+++ b/src/Language/PureScript/CST/Parser.y
@@ -751,9 +751,13 @@ instBinding :: { InstanceBinding () }
   | ident manyOrEmpty(binderAtom) guardedDecl { InstanceBindingName () (ValueBindingFields $1 $2 $3) }
 
 fixity :: { FixityFields }
-  : infix int qualIdent 'as' op { FixityFields $1 $2 (FixityValue (fmap Left $3) $4 (getOpName $5)) }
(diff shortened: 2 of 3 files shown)


JordanMartinez:
So, if I'm understanding this change correctly...

Before, we could write this:
```
add :: Int -> Int -> Int
add x y = x + y

infixl 4 add as +
```
Now, when defining the alias, wrapping the symbol in parenthesis is now allowed:
```diff
-infixl 4 add as +
+infixl 4 add as (+)
```
However, what's not stated in the changelog is that this also allows the [`symbol` parser rule](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CST/Parser.y#L226-L228), which would allow the following as a valid alias:
- `infixl 4 add as ((..))` based on the second part of that rule: https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CST/Parser.y#L228

@PureFunctor Was this intentional? Or did you mean to use [`SYMBOL`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CST/Parser.y#L132) (i.e. which is an alias for `TokSymbolName [] chs` where `chs` are chars from [the list of allowed characters](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CST/Lexer.hs#L730-L731)) rather than `symbol`, the parser rule?

And to be clear, I don't understand why this change is being made. Why wrap aliases in parenthesis? If merged, `(===)` and `===` are technically the same alias even though they look different.

garyb:
I see in the changelog it mentions that this is like Haskell - it's not! What we have now is like Haskell, it doesn't allow parens in fixity declarations either. :smile:

rhendric:
I would also like to see a motivation for this.

PureFunctor:
> Was this intentional?

It was not 😄

> I see in the changelog it mentions that this is like Haskell - it's not! What we have now is like Haskell, it doesn't allow parens in fixity declarations either. 😄

> I would also like to see a motivation for this.

Oh, looks like I mixed it up! The primary motivation for this is leniency and uniformity with how operators appear in import lists and such. I personally often confuse whether or not it's needed, think about import lists, and then proceed to do it wrong.

JordanMartinez:
> The primary motivation for this is leniency and uniformity with how operators appear in import lists and such. I personally often confuse whether or not it's needed, think about import lists, and then proceed to do it wrong.

Ah, I understand now. :smile:
That being said, I don't think we should do this change.

garyb:
I don't feel strongly about this either way :wink: - it is the only place an operator is referenced in isolation without parens, so I can see why one might assume they're required here for consistency, but it is syntactically unambiguous without which is why it works at all.

JordanMartinez:
Should this PR be closed?

PR #4376, opened 2022-08-01
Defer monomorphization for data constructors

purefunctor:
**Description of the change**

Related to #4235. This defers the monomorphization of data constructors introduced by #835. As such, constructors now infer to their "true" type, instead of having to go through generalization.
```purs
> :t Nothing
forall (a :: Type). Maybe a
```

---

**Checklist:**


PureFunctor:
Ah, some of them might be related to object updates. Investigating further.

rhendric:
This PR hasn't gotten much conversation yet, so I guess I'll stop waiting for someone else to ask smart questions and instead start asking stupid questions.

Why do we want this? Is it just to make the implementation of #4235 easier or simpler, or does it fix an independent issue?

What exactly is happening here? The changelog indicates that the inferred type of `Nothing` goes from `forall a1. Maybe a1` to `forall a. Maybe a`, which from a naive reading sure looks like an alpha conversion. I suppose at the level of types in the compiler instead of types in the source language, I'm meant to know that `a1` is a skolem and `a` is a type variable. Even so, how are those types handled differently?

Are there downsides or is there an argument that this is unambiguously what we should have been doing all along?

MonoidMusician:
The issue in the VTA context was briefly mentioned here:
https://github.com/purescript/purescript/pull/4235#issuecomment-1200504305

The issue is that the re-generalization would discard the VTA information, even though the constructors supported VTA by definition.

rhendric:
Sure, but would we still want this change in the (probably counterfactual) world where we decide not to implement VTA? If so, why; if not, what consequences does this have for things that are not related to VTA?

PureFunctor:
Outside of #4235, this effectively just removes some historic baggage introduced by #835 and reworks a part of #2625 to accommodate for constructors. I don't think it ever came up in discussion back then but it's also just nice having constructors infer w/o the numeric suffixes 😄

JordanMartinez (approved):

PR #4380, opened 2022-08-21
Gets rid of typing rule for record identifiers

mikesol:
**Description of the change**

Allows records with constrained terms to typecheck.
---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/TypeChecker/Types.hs
+++ b/src/Language/PureScript/TypeChecker/Types.hs
@@ -369,20 +369,10 @@ infer' (Literal ss (ArrayLiteral vals)) = do
   return $ TypedValue' True (Literal ss (ArrayLiteral ts')) (srcTypeApp tyArray els)
 infer' (Literal ss (ObjectLiteral ps)) = do
   ensureNoDuplicateProperties ps
-  -- We make a special case for Vars in record labels, since these are the
-  -- only types of expressions for which 'infer' can return a polymorphic type.
-  -- They need to be instantiated here.
-  let shouldInstantiate :: Expr -> Bool
-      shouldInstantiate Var{} = True
-      shouldInstantiate (PositionedValue _ _ e) = shouldInstantiate e
-      shouldInstantiate _ = False
-
-      inferProperty :: (PSString, Expr) -> m (PSString, (Expr, SourceType))
+  let inferProperty :: (PSString, Expr) -> m (PSString, (Expr, SourceType))
       inferProperty (name, val) = do
         TypedValue' _ val' ty <- infer val
-        valAndType <- if shouldInstantiate val
-                        then instantiatePolyTypeWithUnknowns val' ty
-                        else pure (val', ty)
+        valAndType <- instantiatePolyTypeWithUnknowns val' ty
         pure (name, valAndType)
 
       toRowListItem (lbl, (_, ty)) = srcRowListItem (Label lbl) ty


mikesol:
It looks like there are a couple regressions when removing this check, ie:

```purescript
module Main where

import Prelude
import Effect.Console (log)

data X a = X

x :: forall a. X a
x = X

type Y = { x :: X Int }

test :: forall m. Monad m => m Y
test = pure { x: x }

type Z t = forall x. t x -> (forall a. t a) -> t x

class C t where c :: Z t

instance cA :: C Array where
  c x _ = x

test2 :: forall m. Monad m => m { ccc :: Z Array }
test2 = pure { ccc: (c :: Z Array) }

main = log "Done"
```

yields

```bash
 The type variable x, bound at
       
           tests/purs/passing/1110.purs:16:12 - 16:51 (line 16, column 12 - line 16, column 51)
       
         has escaped its scope, appearing in the type
                                                           
           { ccc :: Array x6 -> (forall a. ...) -> Array x6
           }                                               
                                                           
       
       in the expression { ccc: c cA
                         }          
       in value declaration test2
```

natefaubion:
That's correct, this check is not benign. I would expect there to be failures because it is no longer inferring polytypes in records.

natefaubion:
```purescript
test2 :: forall m. Monad m => m { ccc :: Z Array }
test2 = pure { ccc: (c :: Z Array) }
```
I think this is an expected failure, not dissimilar from `runST $ ...`, as the polytype would have to be threaded through the call to `pure`.

mikesol:
If the rule currently provides inference for polytypes for records & if the PR causes a regression, maybe this PR isn't a good candidate for merging?

natefaubion:
We don’t infer polytypes in records other than the special case of bare identifiers. I personally think this special case causes far more trouble than it’s worth.

mikesol:
I'm in a bit over my head as it's a part of the compiler I don't know well, but is there any way to avoid the regression while achieving the desired behavior?

natefaubion:
There is no way to specifically have the compiler _infer_:
```
{ wat: mempty }
```
```
{ wat :: forall a. Monoid a => a }
```
Generally we do not infer polymorphism. We only let generalize at the top level. This is a special case for records which doesn't otherwise exist in any other construct in the language.

garyb:
You've stumbled into something that we've talked about for years here @mikesol, I'm not sure if you're aware of that / would pick it up from what Nate is saying. The fact that it will break some existing things is perhaps partially why nobody has gotten too much into it despite it being an occasional topic for a long time, but it doesn't necessarily mean we shouldn't do it.

I think the first time I remember a discussion about it was Phil saying he thought we should restrict record properties to monotypes in like 2014 when I met up with him in person. 😄

JordanMartinez:
Just to clarify, this is a breaking change, right? As some code that currently compiles would likely no longer compile?

natefaubion:
That's correct, this is necessarily a breaking change.

PureFunctor:
An adjacent change to this is #4376. Constructors were made to infer into monotypes to align with the typing rule here. In my changeset, I just hoisted up this predicate and added a special case for singleton constructors.

mikesol:
Yikes, my brief forays into the compiler always land me in hornets nests!!

I don't think I'm the right person to get this PR over the finishing line -
as it's a breaking change, someone with a bit more knowledge of the
ecosystem should make the call if this is a good path forward. Or maybe ask
about it on the next survey in ~6 mo?

PureFunctor:
If we follow through with this change, we should also free the special case done for constructors in #835. Constructors inferring to monotypes instead of polytypes is currently a blocker for visible type applications. There's #4376 of course, which effectively just defers the hoisting for constructors.

PR #4384, opened 2022-08-29
Remove base-compat as a dependency

### LINKED ISSUE #4377: Remove base-compat as a dependency
## Summary

If I understand correctly, `base-compat` was originally introduced as it was possible to compile PureScript over a range of GHC versions. More specifically, it gets rid of the CPP directives needed to get the compiler to build with both GHC 7.x and 8.x (#1610); currently though, we only build the compiler using 9.2.3 and a specific version of the package set, effectively nullfying `base-compat`'s use case.

Are there any other implications that this change would entail?

PureFunctor:
**Description of the change**

Fixes #4377. This replaces all `Prelude.Compat` imports with just `Prelude`. Ideally, we want most of the codebase to use `Protolude` as it encourages usage `Text` and such, but I think replacing most usages of `s/String/Text` should come with a separate PR.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Label.hs
+++ b/src/Language/PureScript/Label.hs
@@ -1,6 +1,6 @@
 module Language.PureScript.Label (Label(..)) where
 
-import Prelude.Compat hiding (lex)
+import Prelude hiding (lex)
 import GHC.Generics (Generic)
 import Codec.Serialise (Serialise)
 import Control.DeepSeq (NFData)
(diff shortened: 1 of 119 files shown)


JordanMartinez (approved):

PureFunctor:
I think this should be the last PR we merge before we release to reduce merge conflicts.

src/Language/PureScript/Label.hs
```diff
@@ -1,6 +1,6 @@
 module Language.PureScript.Label (Label(..)) where
 
-import Prelude.Compat hiding (lex)
+import Prelude hiding (lex)
```
rhendric, line 3: Drop the `hiding (lex)` here too?

rhendric (approved):

PR #4391, opened 2022-09-25
Bump HLint to version 3.5 and address most of the new hints

### LINKED ISSUE #4385: Bump hlint to v3.x
## Summary

`hlint` was initially introduced back in 2020 by #3816, and its version (`2.2.11`) has remained largely the same for the past 2 years. Over time, more hints were added to `hlint` and it was not immediately apparent at the developer side as CI was running on an older version. I don't really think there's any harm in keeping it pinned for extended periods of time, but it'd be nice to catch up every couple of years or so.

## Motivation

Updating to a more recent version would catch hints that have never been hit before, which would also elide the extraneous hints given by `haskell-language-server` once the codebase is updated.

rhendric:
**Description of the change**

Fixes #4385.

---

**Checklist:**

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -18,6 +18,13 @@
 - ignore: {name: "Avoid lambda using `infix`"}
 - ignore: {name: "Fuse mapM/map"}
 - ignore: {name: "Eta reduce"} # This warning will often make suggestions that are no longer valid due to simplified subsumption
+- ignore: {name: "Redundant fmap"}
+- ignore: {name: "Redundant <$>"}
+- ignore: {name: "Use any"}
+- ignore: {name: "Use fromRight"}
+- ignore: {name: "Use maybe"}
+- ignore: {name: "Use unzip"}
+- ignore: {name: "Use <=<"}
 
 
 # Specify additional command line arguments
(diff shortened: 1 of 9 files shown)


rhendric:
Oh, hi glibc hell. The newest HLint binary that doesn't have a glibc 2.27 dependency seems to be 3.3.1, but that version doesn't parse some stuff I'm working on and so I'd rather use 3.5 if we can.

Once I do get this working... as we did in #4090, probably best to merge this with a traditional merge commit instead of squashing, so that individual hints can be backed out if we change our mind about any of them.

.hlint.yaml
```diff
@@ -18,6 +18,7 @@
 - ignore: {name: "Avoid lambda using `infix`"}
 - ignore: {name: "Fuse mapM/map"}
 - ignore: {name: "Eta reduce"} # This warning will often make suggestions that are no longer valid due to simplified subsumption
+- ignore: {name: "Redundant <$>"}
```
PureFunctor, line 21: If I understand correctly, this is included for better clarity on the AST traversals, right?

rhendric: Yup.

PureFunctor (approved):

JordanMartinez (approved):

PureFunctor:
After this was merged, the linting workflow seems to complain about:
```
/root/.stack/compiler-tools/x86_64-linux/ghc-9.2.3/bin/weeder: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```
Any clues as to why?

PR #4392, opened 2022-09-25
Derive Foldable and Traversable

### LINKED ISSUE #2518: Derive Foldable and Traversable

rhendric:
**Description of the change**

Taking over from #4371. Closes #4371, closes #2518.

It's important to me that the three traversal implementations line up with each other, partly because it's beautiful that they do and IMO seeing how they do is an important touchstone in understanding how and why the compiler should be allowed to derive these classes on its own, and partly because I intend to propose expanding our support for these classes to include even more functors and I'd like to save future me and future my-reviewers the work of redoing things then.

So, while I've gratefully lifted the infrastructure and tests from #4371, the implementation is like 90% redone from the ground up to maximize the it's-all-the-same-idea-ness (and to follow more closely the naming conventions and code patterns already used in this module). I have preserved the optimization @JordanMartinez devised involving unnesting records to cut down on lambdas (one of the virtues of this approach is that this optimization lives entirely in `unnestRecords`, a very small function that could be replaced with `identity` to disable it). Feedback that I've gone too far and this implementation is unreadable is welcome; I can unwind some of these abstractions in that case.

The changelogs have been gutted and consolidated with the content moved to purescript/documentation#446.

I still think we can and should do better on presenting these errors but, having added the deduplication I wanted, I would say this is good enough for a first PR; I have rough-draft code that gives a more sophisticated rendering that I'll save for a later PR in order to keep this one less controversial.

Finally, I know the linter job won't pass until we land #4385 and I merge it here. The new code has no HLint 3.5 issues other than a spurious ‘Unused LANGUAGE pragma’ warning that I think I'll just disable in the .hlint.yaml file.

---

**Checklist:**


JordanMartinez:
The `hlint` PR was merged. So, this should build after merging `master` here, right?

JordanMartinez:
I'll review some of this today, but a more thorough review would likely follow tomorrow.

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -1,20 +1,25 @@
+{-# LANGUAGE GADTs #-}
```
JordanMartinez, line 1: Is this something you were using initially and later removed? I don't think I see a usage of this in this file.

tests/purs/failing/FoldableInstance1.purs
```diff
@@ -0,0 +1,9 @@
+-- @shouldFailWith KindsDoNotUnify
+module Main where
+
+import Prelude
+import Data.Foldable (class Foldable)
+
+data Foo = Bar
+
+derive instance Foldable Foo
```
JordanMartinez, line 9: Small nit: but could a newline be added to the end of this file?

tests/purs/failing/FoldableInstance2.purs
```diff
@@ -0,0 +1,10 @@
+-- @shouldFailWith KindsDoNotUnify
+module Main where
+
+import Prelude
+import Data.Foldable (class Foldable)
+
+data Foo :: (Type -> Type) -> Type
+data Foo a = Bar
+
+derive instance Foldable Foo
```
JordanMartinez, line 10: Small nit: but could a newline be added to the end of this file?

tests/purs/failing/FoldableInstance3.purs
```diff
@@ -0,0 +1,9 @@
+-- @shouldFailWith KindsDoNotUnify
+module Main where
+
+import Prelude
+import Data.Foldable (class Foldable)
+
+data Foo f = Bar (f Int)
+
+derive instance Foldable Foo
```
JordanMartinez, line 9: Small nit: but could a newline be added to the end of this file?

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -368,64 +498,118 @@ deriveFunctor
...
+  foldMapVar = mkRef Foldable.identFoldMap
+  flipVar = mkRef Prelude.identFlip
+
+  mkAsymmetricFoldFunction :: Bool -> Expr -> [(ProperName 'ConstructorName, [Maybe ParamUsage])] -> m Expr
+  mkAsymmetricFoldFunction isRightFold recurseVar ctors = do
+    f <- freshIdent "f"
+    z <- freshIdent "z"
+    let
+      appCombiner :: (Bool, Expr) -> Expr -> Expr -> Expr
+      appCombiner (isFlipped, fn) = applyWhen (isFlipped == isRightFold) flip $ App . App fn
```
JordanMartinez, line 542: I had a bit of a hard time understanding how `flip` works here. I'm writing out the code below to help clarify it for myself.
```haskell
-- assume `isFlipped == isRightFold` is `False`
appCombiner (_, fn) expr1 expr2 = App (App fn expr1) expr2

-- assume `isFlipped == isRightFold` is `True`
appCombiner (_, fn) expr1 expr2 = App expr2 (App fn expr1)
```

tests/purs/failing/FoldableInstance10.purs
```diff
@@ -0,0 +1,12 @@
...
+module FoldableInstance10 where
+
+import Prelude
+import Data.Tuple (Tuple)
+import Data.Foldable (class Foldable)
+
+foreign import data Variant :: Row Type -> Type
+
+data Test a = Test (Variant (left :: a, right :: Array a))
+derive instance Foldable Test
```
JordanMartinez, line 12: Oof! I didn't even think of this...

JordanMartinez (commented):
I'm speechless. This code and the abstractions it uses are both just so much cleaner than my attempt.

AFAICT, the code works as expected. I just need to verify the errors.

src/Language/PureScript/TypeChecker/Deriving.hs
rhendric: This is used in the definition of `TraversalOps`, which has an existential quantifier.

src/Language/PureScript/TypeChecker/Deriving.hs
rhendric: Oh, not quite. The second case should be `App (App fn expr2) expr1`. `App . App fn` is a (Haskell) function that accepts two arguments and makes them the arguments to the PureScript function `fn`; flipping the Haskell function is thus equivalent to flipping the PureScript function.

src/Language/PureScript/TypeChecker/Deriving.hs
JordanMartinez: Ah... ok. I'll mark this as resolved.

tests/purs/failing/FoldableInstance9.out
```diff
@@ -215,10 +215,10 @@ Error 9 of 68:
 
     The type variable `a` must only be used as the last argument in a data type:
 
-      tests/purs/failing/FoldableInstance9.purs:85:29 - 85:30 (line 85, column 29 - line 85, column 30)
+      tests/purs/failing/FoldableInstance9.purs:86:24 - 86:25 (line 86, column 24 - line 86, column 25)
 
 
-  under the label `[33mrights[0m`
+  under the label `[33mlefts[0m`
```
JordanMartinez, line 221: Could these source spans be reordered so that what appears in source code first is what's displayed as an error first? Right now, this is sorting the errors by the alphabetical ordering of the labels.

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -359,6 +368,128 @@ decomposeRec' = sortOn fst . go
...
+  :: forall m
+   . MonadError MultipleErrors m
+  => MonadState CheckState m
+  => ModuleName
+  -> ProperName 'TypeName
+  -> m [(ProperName 'ConstructorName, [Maybe ParamUsage])]
+validateParamsInTypeConstructors mn tyConNm = do
+  (_, _, tyArgNames, ctors) <- lookupTypeDecl mn tyConNm
+  param <- note (errorMessage $ KindsDoNotUnify (kindType -:> kindType) kindType) . lastMay $ map fst tyArgNames
+  ctors' <- traverse (traverse $ traverse replaceAllTypeSynonyms) ctors
```
JordanMartinez, line 388: I think these lines are what causes the following code to not work. 
```purs
type Rec2 a f =
  { foo :: f a }
  
data Test f a = Test1 (Rec2 a f)
derive instance Foldable f => Foldable (Test f)
```
Right now, you don't replace the type synonyms before checking the kind of the last argument. The advantage of this is that you don't get the duplicate record labels in cases like `FoldableInstance9.purs`' `Test6` (see the second part of https://github.com/purescript/purescript/pull/4371#issuecomment-1242793472). However, it also means the above code produces the following error message:
```
Error found:
       in module FoldableInstance9
       at tests/purs/failing/FoldableInstance9.purs:97:19 - 97:20 (line 97, column 19 - line 97, column 20)
       
         Could not match kind
                               
           Type -> Type -> Type
                               
         with kind
               
           Type
               
       
       while checking that type f
         has kind Type -> Type
       while inferring the kind of Rec2 a f
       in type constructor Test
       
  To rerun use: --match "/compiler/Failing examples/'FoldableInstance9.purs' should fail to compile/"
```

I would argue that this example should also be derivable. Thoughts?

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -359,6 +368,128 @@ decomposeRec' = sortOn fst . go
...
+    handleValue = unnestRecords $ \usage inputExpr -> visitExpr $ flip App inputExpr <$> mkFnExprForValue usage
+
+    mkFnExprForValue :: ParamUsage -> m Expr
+    mkFnExprForValue = \case
+      IsParam ->
+        pure $ mkVar f
+      MentionsParam innerUsage ->
+        App recurseVar <$> mkFnExprForValue innerUsage
+      IsRecord fields ->
+        usingLamIdent $ extractExpr . traverseFields handleValue fields
```
JordanMartinez, line 489: Just checking my understanding. Because `unnestRecords` is being used and it always consumes an `IsRecord`, doesn't that mean this code path is never triggered? Meaning, this code exists only to satisfy type safety, not because it does anything useful in this current state.

JordanMartinez (commented):
I think after these two issues are further discussed and resolved, I'm ready to approve this PR. The third comment is just me asking a general question about the implementation.

tests/purs/failing/FoldableInstance9.out
rhendric: Sure, that's reasonable.

src/Language/PureScript/TypeChecker/Deriving.hs
rhendric: No, it does something useful if `mkFnExprForValue` is called with a `MentionsParam (IsRecord _)`, for example. `unnestRecords` doesn't recurse into the `ParamUsage` it's given, which is what we want because if a constructor argument is an `Array { foo :: a }`, we can't unnest the `foo`s in the current lambda.

src/Language/PureScript/TypeChecker/Deriving.hs
JordanMartinez: > I'm not getting any errors when I try it.

Ah! You're right. Sorry. The kinds-do-not-unify error I was getting was because the example I tested didn't include a `Foldable f` constraint. I added a `Test8` for this in the `FoldableInstance9.purs` file, found it didn't work, and simplified it to the above. But I didn't test the above separately.

src/Language/PureScript/TypeChecker/Deriving.hs
JordanMartinez: Ah, ok!

JordanMartinez (approved):
Other than my comment regarding the errors being sorted by source span, not the record label's alphabetical ordering, this looks good to me!

JordanMartinez:
Just to clarify, are you waiting for something else before merging this? I wasn't sure how you wanted to handle the commit message.

PR #4395, opened 2022-09-30
Revise CannotDeriveInvalidConstructorArg errors

rhendric:
**Description of the change**

This is a follow-up to #4392. The errors added in that PR were presented with one error per offending occurrence of a functor parameter, which resulted in very verbose terminal output and might not be ideal for IDEs either. This PR reforms the error message to present all of the offending occurrences in a single error message, marking them in the surrounding source text. (Under conditions where ANSI colors are enabled, the ANSI reverse video effect is used for this; otherwise, such as in JSON output, ASCII carets are used on the following line as an ersatz underline effect.)

The source lines excerpted include at least one line before and after every highlighted parameter. More than three lines between two highlighted parameters will trigger an ellipsis.

<details>
<summary>Expand to see sample outputs</summary>
<br/>
A screenshot of the error message for <code>tests/purs/failing/FoldableInstance9.purs</code>:
FoldableInstance9.out

And the text of the same error message in JSON output:
```
  One or more type variables are in positions that prevent Foldable from being derived.
  To derive this class, make sure that these variables are only used as the final arguments to type constructors, and that those type constructors themselves have instances of Foldable.

  tests/purs/failing/FoldableInstance9.purs:
     15 data Test f g h a

     16   = Test1 (f a a a) (f Int a a) (f a a Int) (f Int a Int) (f Int Int Int)
                     ^ ^           ^       ^ ^             ^
     17   | Test2 { all :: f a a a
                             ^ ^
     18           , rights :: f Int a a
                                    ^
     19           , lefts :: f a a Int
                               ^ ^
     20           , middle :: f Int a Int
                                    ^
     21           , none :: f Int Int Int

     22           }

     23   | Test3 (g

     24             { all :: f a a a
                               ^ ^ ^
     25             , rights :: f Int a a
                                      ^ ^
     26             , lefts :: f a a Int
                                 ^ ^
     27             , middle :: f Int a Int
                                      ^
     28             , none :: f Int Int Int

    ...
     32             { nested1 ::

     33                 { all :: f a a a
                                   ^ ^ ^
     34                 , rights :: f Int a a
                                          ^ ^
     35                 , lefts :: f a a Int
                                     ^ ^
     36                 , middle :: f Int a Int
                                          ^
     37                 , none :: f Int Int Int

    ...
     40                 g

     41                   { all :: f a a a
                                     ^ ^ ^
     42                   , rights :: f Int a a
                                            ^ ^
     43                   , lefts :: f a a Int
                                       ^ ^
     44                   , middle :: f Int a Int
                                            ^
     45                   , none :: f Int Int Int

     46                   }

     47                   a
                          ^
     48             }

     49             a)

     50   | Test5 (Rec f a)
                         ^
     51   | Test6 (g (Rec f a) a)
                            ^
     52   | Test7 (h { nested1 :: Rec f a, nested2 :: g (Rec f a) a } a)
                                        ^                      ^  ^
     53 derive instance Foldable (Test f g h)
```

(To check the above, run <code>purs compile --json-errors ... | node -pe 'JSON.parse(fs.readFileSync(0, "utf-8")).errors[0].message'</code>)
</details>

Note that some information that was previously in the per-occurrence error messages is removed here: the errors that point to `Rec f a` don't indicate what parts of `Rec` use `a` incorrectly, whereas the previous implementation used hints to indicate a particular label in `Rec` for each parameter. This is definitely fine when type synonyms aren't involved, and probably fine when type synonyms are defined in a different module since IMO the action the user is more likely to take is to revise the data type getting the instance, not to revise the type synonym. But when the type synonym is local this might be a step back from the status quo. I think this is enough of an edge case that I'm not too worried about it but I could be wrong.

Additionally, I'm not explicitly calling out the names of constructors that have problems, although highlighting the issue in the source text is arguably more useful so I'm not so conflicted about that.

No changelog entry needed if this lands before 0.15.5 is released, as it just revises an unreleased feature in a way that doesn't change what users need to know. Otherwise, I'll add one.

---

**Checklist:**


purescript.cabal
```diff
@@ -186,7 +186,8 @@ common defaults
     typed-process >=0.2.10.1 && <0.3,
     unordered-containers >=0.2.19.1 && <0.3,
     utf8-string >=1.0.2 && <1.1,
-    vector >=0.12.3.1 && <0.13
+    vector >=0.12.3.1 && <0.13,
+    witherable >=0.4.2 && <0.5
```
rhendric, line 190: We were already pulling in `witherable` as a transitive dependency, so this is free.

src/Language/PureScript/Ide/Error.hs
```diff
@@ -51,9 +51,9 @@ encodeRebuildErrors = toJSON . map encodeRebuildError . P.runMultipleErrors
       (P.ErrorMessage _
        ((P.HoleInferredType name _ _
          (Just P.TSAfter{tsAfterIdentifiers=idents, tsAfterRecordFields=fields})))) ->
-        insertTSCompletions name idents (fromMaybe [] fields) (toJSON (toJSONError False P.Error err))
+        insertTSCompletions name idents (fromMaybe [] fields) (toJSON (toJSONError False P.Error [] err))
       _ ->
-        (toJSON . toJSONError False P.Error) err
+        (toJSON . toJSONError False P.Error []) err
```
rhendric, line 56: This means that any error messages that are generated through this code path don't include the new highlighted source file excerpts. The PureScript IDE integration is something I don't have experience with as a user or maintainer, so this could be good or bad depending on where in the UI this stuff goes.

src/Language/PureScript/Ide/Types.hs
```diff
@@ -287,7 +287,7 @@ instance ToJSON Success where
             ]
         ]
     ModuleList modules -> encodeSuccess modules
-    RebuildSuccess warnings -> encodeSuccess (P.toJSONErrors False P.Warning warnings)
+    RebuildSuccess warnings -> encodeSuccess (P.toJSONErrors False P.Warning [] warnings)
```
rhendric, line 290: See previous caveat.

natefaubion:
How does this look for users of `psa`, which people use to get nice errors like this for every other error? 😅

natefaubion:
Additionally, how does this work for users of purs ide?

rhendric:
Great questions about parts of the ecosystem I don't use. `psa` looks simple enough to test—I'll get back to you—but I'm not excited about installing some IDE that has `purs ide` compatibility and figuring out that whole thing. Maybe there's a Vim integration...

rhendric:
`psa` users see this:

rhendric:
With the latest commit, this is what [purescript-vim](https://github.com/raichoo/purescript-vim) + [psc-ide-vim](https://github.com/FrigoEU/psc-ide-vim) present:
FoldableInstance9

Without the latest commit, same but the excerpt isn't presented.

JordanMartinez:
I'm wondering if this PR should also change the type signature of `paras` and then replace all `Box.vcat Box.left` with `paras`:
```diff
-paras :: [Box.Box] -> Box.Box
+paras :: forall f. Foldable f => f Box.Box -> Box.Box
paras = Box.vcat Box.left
```
Right now, this replacement wouldn't work in this PR because it only works on `List` and this PR is usually working in `NonEmpty`.

src/Language/PureScript/Ide/Error.hs
JordanMartinez: This is the only thing stopping me from approving this PR, and mainly because I'm unfamiliar with this code and the resulting implications.

JordanMartinez: Oh, just saw the other commit. Nevermind!

JordanMartinez (approved):

nwolverson:
I think ultimately any annotated source that's actually part of an error message should (perhaps) only be for the straight terminal output, and the json output for tooling should omit it; eg in this case the error really should be associated with a primary source span and a number of secondary source spans which can display natively in an editor and not spliced into the error text. But that's additional machinery - this looks good for now

rhendric:
If that's the expected direction, would it make more sense to omit the annotated source in the JSON for now instead of adding it and ripping it out later?

There's also the fact that the `allSpans` field on `JSONError` is a list, so it would be easy to append the relevant spans to that list; would that make sense to do instead, so that IDEs can start consuming the extra spans right away?

nwolverson:
Interesting that has been in for 5 years and isn't used in the language server yet... Not sure how it ended up using a different type from one field to another...

```
 position :: Maybe ErrorPosition
...
   , allSpans :: [P.SourceSpan]
```

I would definitely like to see the `allSpans` list populated, and won't venture an opinion on what the text content is for now - I think if you want the error to show up as above in `psa` today, that needs to be included in the text.

JordanMartinez:
Is this ready to be merged? Or is there still discussion to be had?

rhendric:
I'm happy with the current state of the PR, so now's the time for any last-minute feedback.

JordanMartinez:
> I'm happy with the current state of the PR, so now's the time for any last-minute feedback.

I've got nothing.

PureFunctor (approved):

nwolverson:
multiple_error_positions

Just implementing multiple spans per error in the language server now. Aside from the monster tooltip with source, works out well

PR #4403, opened 2022-10-16
Make FromJSON instance for Qualified backwards compatible

ptrfrncsmrph:
### Make `FromJSON` instance for `Qualified` backwards compatible

Prior to #4293, `Qualified` was encoded to JSON such that

  ```haskell
  >>> encode $ Qualified Nothing "foo"
  [null,"foo"]
  >>> encode $ Qualified (Just $ ModuleName "A") "bar"
  ["A","bar"]
  ```

  The type of `Qualified` has changed so that `null` no longer appears in JSON output, but for sake of backwards-compatibility with JSON that was produced prior to those changes (pre-`v0.15.2`), we need to accept `null`, which will be interpreted as `Qualified ByNullSourcePos`.

Relates to: https://github.com/purescript/pursuit/issues/470#issuecomment-1279637302

---

**Checklist:**


src/Language/PureScript/Names.hs
```diff
@@ -292,14 +292,17 @@ instance ToJSON a => ToJSON (Qualified a) where
...
     where
     byModule = do
       (mn, a) <- parseJSON2 v
       pure $ Qualified (ByModuleName mn) a
     bySourcePos = do
       (ss, a) <- parseJSON2 v
       pure $ Qualified (BySourcePos ss) a
+    byMaybeModuleName' = do
+      (mn, a) <- parseJSON2 v
+      pure $ Qualified (byMaybeModuleName mn) a
```
ptrfrncsmrph, line 305: Maybe this
```haskell
instance FromJSON a => FromJSON (Qualified a) where
  parseJSON v = byModule <|> bySourcePos <|> byNullSourcePos
    where
    byModule = do
      (mn, a) <- parseJSON2 v
      pure $ Qualified (ByModuleName mn) a
    bySourcePos = do
      (ss, a) <- parseJSON2 v
      pure $ Qualified (BySourcePos ss) a
    byNullSourcePos = do
      (Nothing :: Maybe SourcePos, a) <- parseJSON2 v
      pure $ Qualified ByNullSourcePos a
```
would be preferable? The `byMaybeModuleName` is redundant I think, should only reach here if the array has `null` in first position.

ptrfrncsmrph: Also wondering if the `ToJSON` instance should map `ByNullSourcePos` to `null` (currently maps to `[0,0]`)

ptrfrncsmrph: I think just updating the `SourcePos` instances seems simpler [4d70482](https://github.com/purescript/purescript/pull/4403/commits/4d7048227e1164af86647688795554bc1300b86a)

src/Language/PureScript/AST/SourcePos.hs
```diff
@@ -38,13 +38,16 @@ displaySourcePosShort sp =
     ":" <> T.pack (show (sourcePosColumn sp))
 
 instance A.ToJSON SourcePos where
-  toJSON SourcePos{..} =
-    A.toJSON [sourcePosLine, sourcePosColumn]
+  toJSON = \case
+    NullSourcePos -> A.Null
```
hdgarrood, line 42: This is a breaking change in that anything consuming this JSON now needs to handle the `null` case where it wouldn't have had to previously. If this data type occurs in CoreFn, that will be a problem. I think this is worse than losing the roundtrip property. We should have the roundtrip property for JSON produced by the current version, but losing it for JSON that can only be produced by previous versions is fine, I think.

hdgarrood (approved):

PR #4406, opened 2022-10-19
Organize the compiler's internal constants files

rhendric:
**Description of the change**

Problem: our constants files, by which I mostly mean Language/PureScript/Constants/Prelude.hs, are big and disorganized. The comments in that file are actively misleading, because so many things have been added and rewritten since they were added that they no longer accurately describe any sort of order. They are also very verbose, with the definition of a single identifier taking up to six lines of mostly boilerplate to properly express (a module name, a string, the identifier itself, and type signatures for all of these—not to mention the newlines required to space these things out so it's easier to read). When adding new constants, it's easy to miss that the associated class or module or package already has constants defined elsewhere in the file and one should add one's new constants nearby. The result is that it's difficult to tell at a glance exactly what parts of what PureScript libraries are depended upon by the compiler, and somewhat awkward to add new constants as the compiler is enhanced.

Doing a one-time cleanup pass is unlikely to solve the problem in the long term, because the same natural forces that led to the mess the first time would still be present.

So, while bracing for the feedback that I'm using a tactical nuke to crack a walnut, I'd like to propose using Template Haskell to generate the contents of most of these files. This PR is a demonstration of the technique but not the only way I could imagine doing it. The details are much less important to me than the compression of information and the intrinsic (as opposed to by-comment) organization of identifiers with modules.

In addition to replacing the contents of our constants files with a TH eDSL, I've made the following large-scale changes:

* I've decided, in the face of rather unclear precedent, that constants files should roughly correspond to PureScript packages, with the exception that many small packages previously covered by Prelude.hs can stay there. This means that Data/Generic/Rep.hs has been merged into Prelude.hs and that Data/Traversable.hs has been merged into Data/Foldable.hs.
* The responsibility of which identifiers are associated with which modules was previously distributed partially in the constants files and partially elsewhere in the compiler, such as the pairs defined at the end of Language/PureScript/CoreImp/Optimizer/Inliner.hs. I've moved this responsibility more towards the constants files, which now expose those pairs as needed elsewhere.
* Previously, the constants files exposed a mix of pattern synonyms and values, requiring some awkward predicates to be used elsewhere in the compiler to check for value matches. Now, everything exposed by the constants files is a pattern synonym, and I've enthusiastically used pattern-matching elsewhere where applicable to simplify code.
* The naming conventions for members of constant files were somewhat consistent but occasionally ad-hoc to prevent collisions. Now that every member is a pattern, these names all have to start with an uppercase letter, making collisions more likely. I've taken the opportunity to adopt a Hungarian-notation-esque convention instead; this is documented in Language/PureScript/Constants/TH.hs. It's not the prettiest, but it's simple and less ambiguous and not too hideous, I hope.

I believe and hope the benefits of this approach outweigh the added cost of the Template Haskell abstraction and the churn induced by renaming most of these identifiers, but only if others agree.

---

**Checklist:**


garyb:
This isn't a pain point I've felt personally (I'm probably responsible for half the mess in fact :smile:), so I have no particular opinion on whether we should do it, but here's my thoughts on the PR: 

- I don't know TH, but imagine I could figure out the usage here if I needed to add new stuff. Probably would slightly prefer defining stuff explicitly still, personally, but that's just due to unfamiliarity.
- The prefix notation rationale and implementation makes sense to me.
- Module locations... maybe we could just have all this stuff in one module now?
- I do prefer the code at the use sites for most of these, with the more consistent pattern synonym usage.

JordanMartinez:
I think I'm against using TH in general because of the "magic" involved with it and the effect it can have on rebuilding code. However, I do agree that simplifying this is a "good thing" to do. I also admit that some of my pushback is due to unfamiliarity with this, but that's something I can get over. Lastly, I believe the docs you have written unhide most of the 'magic' of the TH usage here.

> This PR is a demonstration of the technique but not the only way I could imagine doing it

What are other ways this could be accomplished? While TH is a 'tactical nuke', I'm guessing you chose that approach because the alternatives weren't great.

JordanMartinez:
Thanks. I'm good with using TH for this then. I'm also good with the do notation.

I also like Gary's idea of shoving these all into one module because then it's easy to determine whether the compiler depends on some library having a specific name for an identifier. How would that affect the naming scheme?

rhendric:
> How would that affect the naming scheme?

Minimally, if at all. There are already inter-module naming conflicts (`Control.Apply.apply` versus `Data.Function.apply`), which I resolve using prefixes for one or both.

The minor advantage I see to doing some module separation would be that if there are libraries that are only depended upon by narrow slices of the compiler, and if those libraries had identifiers that would have to be disambiguated with prefixes if merged with the prelude, then we could possibly avoid the prefixes by managing the imports in the slices that use them. Overall though I don't find that to be compelling enough to keep the separate Haskell modules if that direction sounds good to you guys—I'd keep `Prim.hs` and put everything else in `Libs.hs`.

JordanMartinez:
> if there are libraries that are only depended upon by narrow slices of the compiler

I think the number of things that depend on the compiler as a library is exceedingly small. So, I don't think this is something worth worrying about.

> I'd keep `Prim.hs` and put everything else in `Libs.hs`.

Ah, that's a reasonable separation. :+1: from me on this.

rhendric:
Non-Prim constants have now been moved to Libs.hs.

src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
```diff
@@ -102,10 +93,10 @@ magicDo effectModule C.EffectDictionaries{..} expander = everywhereTopDown conve
 inlineST :: AST -> AST
 inlineST = everywhere convertBlock
   where
-  -- Look for runST blocks and inline the STRefs there.
-  -- If all STRefs are used in the scope of the same runST, only using { read, write, modify }STRef then
+  -- Look for run blocks and inline the STRefs there.
+  -- If all STRefs are used in the scope of the same run, only using { read, write, modify } then
```
JordanMartinez, line 97: Should these docs still say `runST` and `STRef`?

src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
```diff
@@ -115,28 +106,25 @@ inlineST = everywhere convertBlock
...
-  isSTFunc name (ModuleAccessor _ C.ST name') = name == name'
-  isSTFunc _ _ = False
   -- Find all ST Refs initialized in this block
   findSTRefsIn = everything (++) isSTRef
     where
-    isSTRef (VariableIntroduction _ ident (Just (_, App _ (App _ f [_]) []))) | isSTFunc C.newSTRef f = [ident]
+    isSTRef (VariableIntroduction _ ident (Just (_, App _ (App _ (Ref C.P_new) [_]) []))) = [ident]
     isSTRef _ = []
-  -- Find all STRefs used as arguments to readSTRef, writeSTRef, modifySTRef
+  -- Find all STRefs used as arguments to read, write, modify
```
JordanMartinez, line 123: Same question here.

JordanMartinez (approved):
LGTM other than the 2 doc comments.

src/Language/PureScript/CoreImp/Optimizer/MagicDo.hs
rhendric: The identifiers it's looking for are literally named [`run`](https://pursuit.purescript.org/packages/purescript-st/6.2.0/docs/Control.Monad.ST.Internal#v:run), [`read`](https://pursuit.purescript.org/packages/purescript-st/6.2.0/docs/Control.Monad.ST.Internal#v:read), etc. I figured the reason these comments were the way they were was because the declarations in the Constants.Prelude module were
```hs
runST :: forall a. (IsString a) => a
runST = "run"
```
etc.

I changed the naming convention so I changed the comments to match.

JordanMartinez: Ok. Thanks for clarfying.

Issue #4133, opened 2021-07-03
Support deleted/moved/renamed modules workflows.


wclr:
## Summary

Currently, there exists a problem in dev workflow that if a module actually removed from the codebase it still resides in the compile output and is available as if it existed originally.

This happens in following cases:

**1) The first scenario is the module name update in a source file.**

If in existing source file user renames a module say it removes `module MyModule where` -> `module MyModuleRenamed where` then we build the project (or Language server sends a fast recompile command of changed file) then we happen to have in the output both `MyModule` and `MyModuleRenamed` dirs. And in cache-db.json two entries mapping to the same original source file path.

```json
  "MyModule": {
    "src/MyModule.purs": [...]
  },
  "MyModuleRenamed": {
    "src/MyModule.purs": [...]
  },
```

We have MyModule available as compiled, though the user's expectation is that it should be not.

**2) The second scenario is rename/move of a source file (and upadating the module name after this).**

If the user renames the source file from `MyModule.purs` -> `MyModuleRenamed.purs` (withoug changing the content). LS in this case sends the command to rebuild new saved file (`MyModuleRenamed.purs`). After this In `cache-db.json` file path is not updated, it is still:

```json
  "MyModule": {
    "src/MyModule.purs": [...]
  },
```
though what would be expected:

```json
  "MyModule": {
    "src/MyModuleRenamed.purs": [...]
  },
```

and if after this user updates the module name `module MyModule where` -> `module MyModuleRenamed where` we get in cache-db:

```json
  "MyModule": {
    "src/MyModule.purs": [...]
  },
  "MyModuleRenamed": {
    "src/MyModuleRenamed.purs": [...]
  },
```
Again we have `MyModule` available as compiled, though the user's expectation definitely is that it should be not.

**3) The third scenario is a source file deletion. When a user just want to remove of the module from the codebase.**

Currenlty LS doesn't send any commands to purs-ide, so the things stay as they are and deleted module is still available as compiled. Though in normal workflow it is expected to be removed from then compilation resutls and no to be available.

## Motivation

To suuport common workflow use expectations and imporove DX.

## Proposal

1) When purs-ide recives a rebuild command for the source file it should keep one to one correct mapping of the module name to the source path name. Though this aslo refers to a normal `purs compile` cli command.

2) Remove non-existing modules (based on the absence of the corresponding source file) from the compilcation result.

[...]

PR #4407, opened 2022-10-20
Remove invalid entries from cache-db

EMattfolk:
**Description of the change**

This is a step in the process of resolving #4133. `cache-db.json` currently contains references to files that sometimes does not exist which, while not harmful, is a bit confusing.

Fully fixing #4133 would also require deleting the compiled modules for which the source files have been deleted. I do not know the implications of deleting things from the `output` folder which is why I skipped that for now. Adding it in later should be fairly trivial if that is the route to take. I have had some discussions about this in the past and it seems to be a rather controversial topic. I have also seen many problems arise because old output is left around, so I would like to get some more discussion going.

**Note**

This change successfully removes modules from `cache-db.json` if the source file is removed or the module is renamed, so it works fine when:

- The source file has been renamed
- The source file has been deleted
- The module name inside the source file has changed

~~It does not work when only the module name is changed inside the file.~~ Fixed. Though I think this could be detected by doing the following:

1. Check if there are two modules with the same source file
2. Remove the module that is currently not being compiled

---

**Checklist:**


i-am-the-slime:
Could this be merged?

f-f:
@i-am-the-slime [as documented here](https://github.com/purescript/governance#pull-requests) merging things to the compiler requires two people from core to approve this. I would have a look at this myself but I forgot how this chunk of the compiler works and I'm extremely swamped at the moment. Given that it's the Thanksgiving week a lot of folks might be offline, so it will probably be at least one more week before anyone can look at it.
The change is fortunately quite small and it's tackling something that has been discussed in an issue and on Discord, so I don't see particular hurdles for it to go in soon.

src/Language/PureScript/Make.hs
```diff
@@ -136,7 +136,7 @@ rebuildModuleWithIndex MakeActions{..} exEnv externs m@(Module _ _ moduleName _
 --
 -- If timestamps or hashes have not changed, existing externs files can be used to provide upstream modules' types without
 -- having to typecheck those modules again.
-make :: forall m. (Monad m, MonadBaseControl IO m, MonadError MultipleErrors m, MonadWriter MultipleErrors m)
+make :: forall m. (Monad m, MonadBaseControl IO m, MonadError MultipleErrors m, MonadWriter MultipleErrors m, MonadIO m)
```
nwolverson, line 139: What is this `MonadIO` constraint added for?

src/Language/PureScript/Make.hs
```diff
@@ -261,6 +263,46 @@ make ma@MakeActions{..} ms = do
...
+          cache
+
+      -- Set of modules that should be removed due to there being two or more modules with the same source
+      toRemove :: S.Set ModuleName
+      toRemove =
+        M.foldl
+          (\acc1 e -> case e of
+            -- This pattern will clean up duplicates over time.
+            -- It will keep the cache in a good state if it was in a good state before.
+            m1 : m2 : _ ->
```
nwolverson, line 286: isn't this just when `Set.size e > 1` then add `e \\ modules` to the accumulator? Then there's no need to wait until multiple compilations reach a fixed point (assuming I understand the comment)

src/Language/PureScript/Make.hs
EMattfolk: I needed to check which files existed in order to remove them from the cache which meant I had to do IO. I'm kind of new to haskell but my LSP suggested i add that constraint and it just worked. It felt kind of like it shouldn't be there when I wrote it, not sure how I would do it otherwise

src/Language/PureScript/Make.hs
EMattfolk: I don't think so? What this means is if there are two or more modules with the same name in the cache, at most one of them is being compiled right now (i.e. it is in the modules set). `e \\ modules` (assuming \\\\ is set difference) will be all duplicate modules, but we want to ignore those, not find them. Perhaps `e \\ (e \\ modules)` is easier to understand?

Anyway, in practice it is possible there are multiple duplicates if you rename and compile several times, but after this change it is only possible to have a single duplicate, so this code is enough. If you think it is better to refactor it I will do that

nwolverson: Sorry, not sure if I was unclear or I don't understand - I was suggesting that `e \\ modules` is added to the accumulator that goes into this fold to construct `toRemove`, as `acc2 = S.union (Set.singleton m1 \\ modules) acc1` etc

I don't feel strongly about the code, just trying to follow through and understand "clean up duplicates over time"

src/Language/PureScript/Make.hs
nwolverson: Right, I just couldn't see what required this constraint straight away

src/Language/PureScript/Make.hs
EMattfolk: I think I understand what you mean now. I must have been tired when I read your comment initially :) Yes, just doing the set difference is probably the way to go. I will refactor it

rhendric: While you're in there, I think this might be a bit clearer if you separate out the duplicate pruning and the missing file pruning into separate functions.

Also, general tip: you're doing a lot of bespoke folds here to accumulate and aggregate stuff, and I think you'd be better off using simpler primitives like `concat` and `filter` and `M.fromListWith`.

tests/TestMake.hs
```diff
@@ -196,6 +198,78 @@ spec = do
...
+    it "writes cache-db.json" $ do
+      let modulePath = sourcesDir </> "Module.purs"
+
+      writeFileWithTimestamp modulePath timestampA "module Module where\nfoo = 0\n"
+      compile [modulePath] `shouldReturn` moduleNames ["Module"]
+
+      cache <- readCacheDb
+      if M.keysSet cache == moduleNames ["Module"]
+        then return ()
+        else fail "Module not in cacheDb"
```
rhendric, line 210:
```suggestion
      M.keysSet <$> readCacheDb `shouldReturn` moduleNames ["Module"]
```

And likewise for the rest of these tests.

EMattfolk:
I may have accidentally simplified the implementation quite a bit. I believe the behavior is the same.

What do you think about also removing the old compiled output while we're at it? @rhendric @nwolverson Or should this be merged as-is?

nwolverson:
> What do you think about also removing the old compiled output while we're at it

My general feeling was this PR is currently uncontroversial and at this point probably want to get it merged rather than expanding scope.

Apologies for not noticing the question and responding sooner. I don't know what the build problem was - can we kick that (perhaps via bringing up to date with latest) and see where we stand?

nwolverson:
I mean there are no logs for the failed build checks (from what I see) so it's not clear if they were merely "build issues" or actual concrete things to be resolved. I assume from your comment that there was a test failure - I'm happy to comment on a particular case if you can give a pointer

Issue #4005, opened 2021-02-01
Proposal: When a type mismatch is inside a record, print where.


reactormonk:
## Summary

If there is a type mismatch when trying to unify two records, e.g.
```
{ x: "foo" }
```
and
```
{ x: 3 }
```
the compiler will mention it couldn't match `String` with `Int`, but it won't say which label it's under.

## Motivation

Less searching required to figure out where exactly the mismatch lies.

## Proposal

On error messages which refer to anything inside a label, also mention the label (possibly its full path with nesting).

## Examples

Real-world example with the type mismatch being somewhere down the hatches:

```

  Could not match type
                
    Maybe String
                
  with type
          
    String
          

while trying to match type                                                                  
                             ( getInputProps :: { _aria :: Object String                    
                                                , autoComplete :: String                    
                                                , id :: String                              
                                                , onBlur :: EffectFn1 SyntheticEvent Unit   
                                                , onChange :: EffectFn1 SyntheticEvent Unit 
                                                , onKeyDown :: EffectFn1 SyntheticEvent Unit
                                                , value :: String                           
                                                }                                           
                             , getItemProps :: { index :: Int                               
                                               , item :: { value :: String                  
                                                         }                                  
                                               , key :: String                              
                                               , style :: CSS                               
                                               }                                            
                                               -> { _aria :: Object String                  
                                                  , id :: String                            
                                                  , role :: String                          
                                                  , style :: CSS                            
                                                  }                                         
                             , getLabelProps :: { htmlFor :: String                         
                                                , id :: String                              
                                                , ref :: Ref (Nullable Node)                
                                                }                                           
                             , getMenuProps :: { _aria :: Object String                     
                                               , id :: String                               
                                               , ref :: Ref (Nullable Node)                 
                                               , role :: String                             
                                               }                                            
                             , getRootProps :: { _aria :: Object String                     
                                               , ref :: Ref (Nullable Node)                 
                                               , role :: String                             
                                               }                                            
                             , inputValue :: Maybe String                                   
                             , selectedItem :: Maybe                                        
                                                 { value :: String                          
                                                 | t4                                       
                                                 }                                          
                             ...                                                            
                             )                                                              
                                                                                            
  with type                                                
              ( getInputProps :: Record t9                 
              , getItemProps :: { index :: Int             
                                , item :: { value :: String
                                          }                
                                , key :: String            
                                , style :: CSS             
                                }                          
                                -> Record t8               
              , getLabelProps :: Record t7                 
[...]

PR #4411, opened 2022-10-27
Mention which row label the type error occurs on

FredTheDino:
**Description of the change**
Writing code is hard, we make errors. Better errors makes it easier to write code.
This PR makes it clear which row-label type errors occur on.

Example of the previous error message for this code:
```purs
a :: { f :: Int }
a = { f: 1 }

b :: { f :: String }
b = a -- Oh noes!
```

```
Error found:
...

  Could not match type

    Int

  with type

    String


while checking that type { f :: Int
                         , qa :: Int
                         , qb :: Int
                         , qc :: Int
                         , qd :: Int
                         }
...
```

Now it reads:
```
Error found:
...

  Could not match type

    Int

  with type

    String


while matching row label f
while checking that type { f :: Int
                         , qa :: Int
                         , qb :: Int
                         , qc :: Int
                         , qd :: Int
                         }
```

This is very helpful when you have a lot of labels in your records (150) which I sometimes have and the pretty printer for type errors cuts of the records halfway through or you have to manually differentiate these to find out what's wrong.

## Needed guidance
  - Is the error formatting good?
  - Haskell formatting and style opinions are welcome
---
Related to #4005
 
**Checklist:**


tests/purs/failing/RecordLabelOnTypeError.out
```diff
@@ -0,0 +1,28 @@
...
+  [33m     [0m
+  [33m  Int[0m
+  [33m     [0m
+  with type
+  [33m        [0m
+  [33m  String[0m
+  [33m        [0m
+
+while matching row label [33mfield[0m
+
```
purefunctor, line 15: There seems to be an extra newline here? I've yet to check as well but at first glance it seems to be the case...

FredTheDino: Right on the money!

MonoidMusician: you can run with e.g. `HSPEC_ACCEPT=true stack test --test-arguments="--match compiler"` to automatically write the golden test output for you: https://github.com/purescript/purescript/blob/master/CONTRIBUTING.md#running-tests

src/Language/PureScript/Types.hs
```diff
@@ -456,7 +456,7 @@ alignRowsWith f ty1 ty2 = go s1 s2 where
   go lhs@(RowListItem a1 l1 t1 : r1) rhs@(RowListItem a2 l2 t2 : r2)
     | l1 < l2 = (second . first . first) (RowListItem a1 l1 t1 :) (go r1 rhs)
     | l2 < l1 = (second . second . first) (RowListItem a2 l2 t2 :) (go lhs r2)
-    | otherwise = first (f t1 t2 :) (go r1 r2)
+    | otherwise = first (f l1 t1 t2 :) (go r1 r2)
```
purefunctor, line 459: I would take the opportunity to refactor this to use `compare` instead. I feel the `otherwise` case doesn't really express that `l1 == l2`.
```hs
  go lhs@(RowListItem a1 l1 t1 : r1) rhs@(RowListItem a2 l2 t2 : r2) =
    case compare l1 l2 of
      LT -> (second . first . first) (RowListItem a1 l1 t1 :) (go r1 rhs)
      GT -> (second . second . first) (RowListItem a2 l2 t2 :) (go lhs r2)
      EQ -> first (f l1 t1 t2 :) (go r1 r2)
```

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -444,7 +444,7 @@ unifyKindsWithFailure onFailure = go
       onFailure w1 w2
 
   unifyRows r1 r2 = do
-    let (matches, rest) = alignRowsWith go r1 r2
+    let (matches, rest) = alignRowsWith (\l t1 t2 -> withErrorMessageHint (ErrorInRowLabel l) $ go t1 t2) r1 r2
```
purefunctor, line 447: This is a nitpick, but maybe we could extract this lambda (and others) to a binding?
```hs
goWithLabel l t1 t2 = withErrorMessageHint (ErrorInRowLabel l) $ go t1 t2 
```

purefunctor (commented):
Just a quick round-up of the changes

src/Language/PureScript/TypeChecker/Kinds.hs
FredTheDino: Fixed, I don't know if I agree that this is nicer but I trust your judgement. :)

src/Language/PureScript/Types.hs
FredTheDino: Changed! :D

tests/purs/failing/NestedRecordLabelOnTypeError.out
```diff
@@ -0,0 +1,28 @@
...
+
+while checking that expression [33m{ field: 1[0m
+                               [33m}         [0m
+  has type [33m{ a :: { b :: { c :: Int[0m
+           [33m              }         [0m
+           [33m       }                [0m
+           [33m}                       [0m
+in value declaration [33mrecord[0m
+
+where [33mt0[0m is an unknown type
```
JordanMartinez, line 24: Looks like there is no mention of the label here as is expected based on your comment in the corresponding `purs` file.

tests/purs/failing/NestedRecordLabelOnTypeError.purs
```diff
@@ -0,0 +1,8 @@
+-- @shouldFailWith TypesDoNotUnify
+module NestedRecordLabelOnTypeError where
+
+record :: { a :: { b :: { c :: Int } } }
+record = { field: 1 }
```
MonoidMusician, line 5:
```suggestion
record = { a: { b: { c: 1 } } }
```

FredTheDino: Updated....

MonoidMusician:
I think this is the next place you'll want to hit to add label information for subsumption:
https://github.com/purescript/purescript/blob/b7e05967ecec890c1d45d266e221bc877a87f184/src/Language/PureScript/TypeChecker/Subsumption.hs#L105-L116

Probably need to add a `alignRowsWithLabel` helper next to `alignRowsWith`, or something:
https://github.com/purescript/purescript/blob/b7e05967ecec890c1d45d266e221bc877a87f184/src/Language/PureScript/Types.hs#L438-L449

tests/purs/failing/NestedRecordLabelOnTypeError.purs
JordanMartinez: I thought the intent behind this error was to produce something like:
```
coudl not unify
  Int
with type
  { a :: { b :: { c :: Int } } }

at label `field`
```
Right now, doesn't this test the same thing as one of the other tests?

FredTheDino:
> I think this is the next place you'll want to hit to add label information for subsumption:
> 
> https://github.com/purescript/purescript/blob/b7e05967ecec890c1d45d266e221bc877a87f184/src/Language/PureScript/TypeChecker/Subsumption.hs#L105-L116
> 
> Probably need to add a `alignRowsWithLabel` helper next to `alignRowsWith`, or something:
> 
> https://github.com/purescript/purescript/blob/b7e05967ecec890c1d45d266e221bc877a87f184/src/Language/PureScript/Types.hs#L438-L449

I think I've already changed that, and I think you're referencing into a different commit. Currently it is:
https://github.com/purescript/purescript/blob/2c90056b32d6f9bfe6bf6355dcc8ba80c3694207/src/Language/PureScript/TypeChecker/Subsumption.hs#L105-L116

https://github.com/purescript/purescript/blob/2c90056b32d6f9bfe6bf6355dcc8ba80c3694207/src/Language/PureScript/Types.hs#L438-L449

But I changed the `alignRowsWith` function so the first given function also takes a label. I don't know which other places to poke in to get more of these row merge errors. But I can probably add some more error messages in other places. I feel like it's smart to start with the low hanging fruit.

What more is required for this to get merged?

MonoidMusician:
You're right, my bad! Since we're checking that a record literal (`{ a: "a" }`) has a known type (`{ a :: Int }`), we stay in checking mode, we are not doing subsumption yet. So the place that happens is actually here:
https://github.com/purescript/purescript/blob/2c90056b32d6f9bfe6bf6355dcc8ba80c3694207/src/Language/PureScript/TypeChecker/Types.hs#L816-L827

I think if we tag those errors there with your `ErrorInRowLabel` hint, your `NestedRecordLabelOnTypeError` test should have that label information now too.

Other than that I think this is looking good. I'd like to give it some testing locally of course.

rhendric:
I'm not sure we want the hint there. The error already points to the precise location in the source where the issue is, which is better than a hint.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -419,6 +419,7 @@ unifyKindsWithFailure
   -> m ()
 unifyKindsWithFailure onFailure = go
   where
+  goWithLabel l t1 t2 = withErrorMessageHint (ErrorInRowLabel l) $ go t1 t2
```
rhendric, line 422: Instead of defining these local helpers in three different places, could you add a helper next to `alignRowsWith` called

```hs
alignRowsWithHinted
  :: (MonadState CheckState m, MonadError MultipleErrors m)
  => (Type a -> Type a -> m ())
  -> Type a -> Type a -> (m (), (([RowListItem a], Type a), ([RowListItem a], Type a)))
```
(note the absence of `Label` in this type), which simply delegates to your updated `alignRowsWith` and adds the hint? (I'd also factor the `sequence_` call that co-occurs with all of these call sites into this helper, which is why `m ()` and not `[m ()]`.)

MonoidMusician:
Perhaps, but you still need to manually keep track of how you got there since it doesn't tell you where the type is coming from. If the field is hidden behind a lot of type synonyms, I think it would be helpful to just see what label is failing at first glance.

I guess with verbose errors we see a little bit more context, but it's pretty rare that I actually enable that flag:
```
  Could not match type
          
    String
          
  (defined in module Prim)
  with type
       
    Int
       
  (defined in module Prim)

while trying to match type String
  with type Int
while checking that type String
  is at least as general as type Int
while checking that expression "a"
  has type Int
while checking that expression { a: "a"
                               }       
  has type { a :: Int
           }         
in value declaration record
```

rhendric:
Not sure I understand you. If we're checking a record literal, the thing that gets underlined in an IDE will be the field in the record literal; the label is right there in the expression, so it doesn't matter if the label in the type is hidden behind a synonym. If we're checking a non-literal term, the compiler will attempt to unify the type of that term with the type and the code path that already has hints will be taken.

MonoidMusician:
Well I think the root of my dissatisfaction is that the non-verbose hints jump straight to “checking expression `"a"`”:
```
while checking that expression "a"
  has type Int
in value declaration record
```
without an indication of how the typechecker got there from checking `record = …`. I think there should be some form of breadcrumbs there even in non-verbose mode. Otherwise it looks like it just picked the type `Int` out of thin air. I know you can reconstruct it from the source, but still, adding a hint “while matching label `a`” looks awfully convenient to me …

FredTheDino:
I'm not sure I understand the reasoning from @rhendric.
Is your reasoning that:
 - We point to the source location of the error
 - From this information we can deduce the label by looking at the code and the error
 - It's better to not touch code
If these are your arguments I wonder why you don't think these warnings are useful information?

On one level I feel like this adds a bit of output to the error messages. This can be seen as noise. But I also think it adds helpful information for understanding the type errors quicker and easier.

I also agree that most developers will be using the IDE features of purs and get an inline error. The IDE features are good and we should encourage people to use them. But I don't feel like an IDE is a good substitute for good error messages. Some users of the compiler are new users, and adding a bit of extra information can probably help them find the errors even if they're using an IDE.

TL;DR:
I really agree with @MonoidMusician that informing users of how we got to the error is convenient and helpful. It aids understanding, troubleshooting and makes purescript a friendlier language.

rhendric:
I'm not making any sort of argument about touching code. Certainly the amount of code that would be involved here would be justified by improving the error message. I only disagree that, in the narrow case currently under discussion, a label hint is an improvement.

Your code is going to look like:
```purs
blah blah blah {
  label1: blah blah,
  label2: blah blah,
  label3: this is a problem,
  label4: blah blah
}
```

The error message is going to tell you that `this is a problem` is not the correct type. It'll point you to line and column, which one can visit manually or see highlighted in an IDE.

What value is added by adding a line to the error message that the problem was detected while checking `label3`? Of course we're checking `label3`! You'd hardly expect to see an error about `this is a problem` while checking `label1`!

The cost is that the more noise we add to our hints (which, IMO, are already on the noisy side), the less useful they are in aggregate for pointing users at what they need to know.

When we aren't talking about record literals, the hints you're adding in this PR certainly add value, because the error message no longer points to a specific label. I have no issue whatsoever with adding hints in those cases; the cost is justified.

I am all for making our error messages more convenient and more helpful. This PR does that, and I approve. More information is not always more helpful, though; one has to consider whether the information is necessary or redundant.

src/Language/PureScript/TypeChecker/Kinds.hs
FredTheDino: I've finally had enough time to look at this. But this is not really possible to do. We get a dependency cycle:
```
Module imports form a cycle:
        module ‘Language.PureScript.Types’ (src/Language/PureScript/Types.hs)
        imports module ‘Language.PureScript.Errors’ (src/Language/PureScript/Errors.hs)
  which imports module ‘Language.PureScript.Types’ (src/Language/PureScript/Types.hs)
```

I'm not sure I want to touch this much code and if this is the better change. It would have been great if it worked though. I'm all for exposing less and not moving complexity to the call sites. Curse your cyclic dependencies!

FredTheDino:
> You're right, my bad! Since we're checking that a record literal (`{ a: "a" }`) has a known type (`{ a :: Int }`), we stay in checking mode, we are not doing subsumption yet. So the place that happens is actually here:
> 
> https://github.com/purescript/purescript/blob/2c90056b32d6f9bfe6bf6355dcc8ba80c3694207/src/Language/PureScript/TypeChecker/Types.hs#L816-L827
> 
> I think if we tag those errors there with your `ErrorInRowLabel` hint, your `NestedRecordLabelOnTypeError` test should have that label information now too.
> 
> Other than that I think this is looking good. I'd like to give it some testing locally of course.

This should now be fixed! Thanks for the pointers. Now everything I feel needs fixing is fixed. It's further style tips and the like. Otherwise I'm happy!

tests/purs/failing/1175.out
```diff
@@ -15,6 +15,7 @@ while checking that type [33mInt[0m
   is at least as general as type [33mString[0m
 while checking that expression [33m1[0m
   has type [33mString[0m
+while matching label [33mfoo[0m
```
rhendric, line 18: This is a bad hint; there are no record labels involved in the code, so understanding this hint requires knowing that the compiler desugars instances to records internally, which is a detail we don't want users to have to know (as it might change).

tests/purs/failing/2616.out
```diff
@@ -16,6 +16,7 @@ while applying a function [33mcompare[0m
   of type [33mOrd t0 => t0 -> t0 -> Ordering[0m
   to argument [33m$l2[0m
 while inferring the type of [33mcompare $l2[0m
+while matching label [33mcompare[0m
```
rhendric, line 19: This is also a bad hint. There is no `compare` label in the code; the identifier `compare` doesn't even appear.

tests/purs/failing/ExtraRecordField.out
```diff
@@ -4,6 +4,8 @@ at tests/purs/failing/ExtraRecordField.purs:9:13 - 9:54 (line 9, column 13 - lin
 
   Type of expression contains additional label [33mage[0m.
 
+while matching label [33mlast[0m
+while matching label [33mfirst[0m
```
rhendric, line 8: This is a very bad hint; the labels `last` and `first` are correct! The issue, as the message otherwise states, is with the label `age`.

tests/purs/failing/FoldableInstance4.out
```diff
@@ -12,6 +12,7 @@ while applying a function [33mfoldl[0m
   of type [33mFoldable t0 => (t1 -> t2 -> t1) -> t1 -> t0 t2 -> t1[0m
   to argument [33m$f1[0m
 while inferring the type of [33mfoldl $f1[0m
+while matching label [33mfoldl[0m
```
rhendric, line 15: This is bad for the same reason the new hint in 2616 is bad.

tests/purs/failing/InstanceSigsBodyIncorrect.out
```diff
@@ -15,6 +15,7 @@ while checking that type [33mBoolean[0m
   is at least as general as type [33mNumber[0m
 while checking that expression [33mtrue[0m
   has type [33mNumber[0m
+while matching label [33mfoo[0m
```
rhendric, line 18: This is bad for the same reason the new hint in 1175 is bad.

tests/purs/failing/InstanceSigsDifferentTypes.out
```diff
@@ -15,6 +15,7 @@ while checking that type [33mNumber[0m
   is at least as general as type [33mInt[0m
 while checking that expression [33m0.0[0m
   has type [33mInt[0m
+while matching label [33mfoo[0m
```
rhendric, line 18: This is bad for the same reason the new hint in 1175 is bad.

tests/purs/failing/Superclasses1.out
```diff
@@ -9,6 +9,8 @@ at tests/purs/failing/Superclasses1.purs:12:1 - 13:17 (line 12, column 1 - line
 
 while checking that expression [33m#dict Su[0m
   has type [33mSu$Dict t0[0m
+while matching label [33m"Su0"[0m
+while matching label [33mcl[0m
```
rhendric, line 13: This is bad twice. The `"Su0"` hint is bad because it exposes an implementation detail of how superclasses are implemented. The `cl` hint is bad for the same reason as in 1175.

tests/purs/failing/Superclasses5.out
```diff
@@ -13,6 +13,7 @@ while checking that expression [33mcase $0 of       [0m
                                [33m  [ x ] -> [ su x[0m
                                [33m           ]     [0m
   has type [33mt0[0m
+while matching label [33msu[0m
```
rhendric, line 16: This is bad for the same reason the new hint in 1175 is bad.

rhendric (changes requested):
I am unimpressed with the latest two commits. I think the functionality added there is unnecessary, as I've previously argued, but I'm willing to accept being outvoted on that. However, the side effect is that actively misleading hints have been added to other scenarios, as demonstrated in the other outputs that needed to change, and this I'm going to hold my ground on as a bad thing. Please at the very least make sure that these cases aren't changed by what you want to do.

src/Language/PureScript/TypeChecker/Kinds.hs
rhendric: Oh, that's unfortunate. Fine to leave as-is, then.

src/Language/PureScript/TypeChecker/Types.hs
```diff
@@ -838,18 +838,19 @@ checkProperties expr ps row lax = convert <$> go ps (toRowPair <$> ts') r' where
...
-        rest <- freshTypeWithKind (kindRow kindType)
-        unifyTypes r (srcRCons (Label p) ty rest)
-        ps'' <- go ps' ts rest
-        return $ (p, v') : ps''
-      Just ty -> do
-        v' <- check v ty
-        ps'' <- go ps' (delete (Label p, ty) ts) r
-        return $ (p, v') : ps''
+  go ((p,v):ps') ts r = 
+    withErrorMessageHint (ErrorInRowLabel (Label p))
```
MonoidMusician, line 842: This is currently applying recursively for each call of `go`, resulting in extra labels showing up. It just needs to apply to `infer`/`check` below.

rhendric:
In that case I would prefer to merge this without the most recent commits. The goal is to make the error messages more helpful, which this PR achieves even without adding hints on the case where hints are least necessary. If the PR adds hints that are sometimes helpful and sometimes misleading, I think that's a worse outcome.

FredTheDino:
> In that case I would prefer to merge this without the most recent commits. The goal is to make the error messages more helpful, which this PR achieves even without adding hints on the case where hints are least necessary. If the PR adds hints that are sometimes helpful and sometimes misleading, I think that's a worse outcome. (It trains users to ignore the hint, and we have too many hints that, speaking only for myself, I already reflexively ignore.)

I agree. Some of these hints are confusing. Errors overall can use some cleaning. I've reverted the last two commits. It's better we add something that is a clear improvement - if we want to add them we add them later.
Thanks for all the feedback @rhendric and @MonoidMusician!

Is there anything more that anyone wants to see regarding this PR. Otherwise I'd love to see this get merged. :D

tests/purs/failing/RecordLabelOnTypeErrorImediate.purs
```diff
@@ -0,0 +1,5 @@
+-- @shouldFailWith TypesDoNotUnify
+module NestedRecordLabelOnTypeError where
+
+record :: { a :: Int }
+record = { a: "a" } -- this should trigger an error, telling us there's a mismatch in the field `a`, but it doesn't
```
rhendric, line 5: It's reasonable to leave this test in place, to document that we don't expect a hint in this case, but could you please:
* Amend this comment to reflect that
* Fix the spelling error in the file names (Imediate -> Immediate)

rhendric (approved):
You have my approval <s>conditional on addressing the minor feedback on the last test</s> ☑.

src/Language/PureScript/TypeChecker/Types.hs
FredTheDino: I made these changes, it didn't fix all of the comments rhendic mentioned earlier. If you want to look for yourself you can see them here: https://github.com/purescript/purescript/pull/4416/commits/e136c2a6a79ec9c0fbe7de9d66ae8dc49f7553d6

I think the right decision is to not add the information there. I'd rather see this merged and then move to cleaning up the error messages in general. :D

Thanks for the input! :)

FredTheDino:
I'd love to see this merged. What more do you want me to do?

rhendric:
Appreciate your engagement! Currently, what this PR needs is a second maintainer to review it.

JordanMartinez:
I haven't forgotten about this PR. I anticipate re-reviewing this sometime in the next two weeks. Sorry for the delay. I've been focusing my free time on porting a JS library to PureScript.

JordanMartinez (approved):
LGTM. I'm grateful for all the others who reviewed this as it feels like I'm more rubber-stamping this than anything else.

JordanMartinez:
I don't have enough time to write up a cleaner commit message that summarizes the work down here and reasons for why other work was NOT done here, so I'm not going to merge this yet. Perhaps another core team member can do that if I don't get to it sooner.

MonoidMusician (approved):
Hey @FredTheDino, sorry I haven't had a chance to get back to this earlier! Been busy with things especially work on the registry.

I gave it a test run and it's working well so I'm happy to merge it. There is a small issue with the non-verbose hints still being a bit too verbose that I will look at addressing in a later PR (we could trim out some `while matching type X with type Y` in between row labels, it only shows up in some circumstances). I will start looking at your other PR again soon.

Thanks for your efforts!

PR #4420, opened 2022-11-18
Derive Bi*, Contravariant, and Profunctor

rhendric:
The compiler can now derive instances of `Bifunctor`, `Bifoldable`, `Bitraversable`, `Contravariant`, and `Profunctor`, as well as use those classes when deriving `Functor`, `Foldable`, and `Traversable`, enabling more instances to be derived.

**Description of the change**

In #4371, I alluded to a more expansive treatment of `Functor`/`Foldable`/`Traversable` deriving that would cover more cases, including the special treatment that GHC gives to functions (see https://github.com/purescript/purescript/pull/4371#discussion_r950631674 and https://github.com/purescript/purescript/pull/4371#issuecomment-1195057756). This PR implements what I had in mind.

There are two things happening here. First, the set of supported derivable classes is being expanded. FFT-style derivation now allows all of the following to be derived:

| + | + + | − | − + |
|---|---|---|---|
|Functor|Bifunctor|Contravariant|Profunctor|
|Foldable|Bifoldable|—|—|
|Traversable|Bitraversable|—|—|

Two of these classes have contravariant parameters and this PR handles the contravariance correctly—it will not derive a `Contravariant` instance for `newtype Example1 a = Example1 (Predicate (Predicate a))`, but will for `newtype Example2 a = Example2 (Predicate (Predicate (Predicate a)))`.

Second, and more subtly, it was previously the case that if you wanted to derive a `Functor`, the only instances that the compiler would take advantage of were other `Functor` instances; and respectively for other classes. With this PR, the compiler will use any class in the above table in the same row as the class being derived. For example, previously the compiler would not derive a `Functor` instance for `newtype Example3 a = Example3 (Tuple a a)`, because it would only use the `Functor` instance for `Tuple` and that won't map over the first type parameter of `Tuple`. In this PR, a `Functor` instance for `Example3` can exist because it will use the `Bifunctor Tuple` instance to map over both type parameters; the implementation will be `map f (Example3 v) = Example3 (bimap f f v)`.

In particular, GHC's special handling of function types when deriving `Functor` is covered by this second change, via the `Profunctor Function` instance. The `Functor` instance derived for `newtype Example4 a = Example4 ((a -> Int) -> Int)` will be implemented with `map f (Example4 v) = Example4 (lcmap (lcmap f) v)`, which has behavior equivalent to what GHC would do with that type.

This does introduce a potential issue, however. With the current behavior, there is never any question about whether the relevant instances are available for the compiler to implicitly depend on—if you are deriving a `Functor`, any instances of `Functor` that are relevant to the derivation are either in the `Data.Functor` module (which you must depend on to mention `Functor` in your instance) or in the module of a type you're using in your instance (which you must depend on to mention that type in your instance). This is no longer the case with this PR, as a `Functor` instance could, as `Example4` demonstrates, depend on a `Profunctor` instance defined in a module not among the modules directly (or indirectly!) used by the deriving module.

In most cases, based on my (manual) testing, `purs` handles this situation gracefully. Since `Bifoldable`, `Bitraversable`, `Foldable`, and `Traversable` are all defined in the same package, the compiler always has access to the relevant instances anyway. However, each of the four `Functor`-compatible classes is defined in a different package, and if, for example, `purescript-profunctor` is not among the dependencies of a project, `purs` will not allow `Functor Example4` to be derived (it fails with an error message indicating that the `a` is not in a valid position to derive `Functor`).

(Note that this *does not* affect backward compatibility; it's only a potential point of confusion for the new functionality.)

One way to resolve that issue, if it's sufficiently worrisome, would be to move the `Bifunctor`, `Contravariant`, and `Profunctor` classes into `purescript-prelude`. Another way might be to make the error message more helpful (by suggesting which packages to add to one's project).

The documentation PR for this feature (purescript/documentation#446) will need to be updated.

---

**Checklist:**


garyb:
This looks great!

Regarding the implicit-relevant-instances thing and the location of the classes, before I even read the contents of this PR it got me thinking about how much deriving we want to do, since part of the reason we don't have a lot of deriving is so that the compiler doesn't become too intertwined with the library ecosystem, that it should be able to stand on its own without extra dependencies, etc. 

But I think we've already taken a good few steps along that path already, and my initial reaction to seeing this PR was definitely "hell yeah" rather than "hmm, philosophically that doesn't quite fit with ...". Also, I guess adding deriving for a wider set of classes still isn't preventing someone from using an entirely different ecosystem from the `Prelude` up (or doing some kind of mixing and matching), it's just they won't be necessarily be able to use the privileged support that we're adding for these cases.

So I think where I'm at these days is I'd be fine with adding deriving for anything we maintain in the `purescript` organization, basically. It may cause trouble if people are using a non-standard setup and trying to rely on deriving, but I think it's reasonable to say "sorry, it's not supported for that".

There is always the `DerivingVia` discussion that comes up around these things too, but I don't know if adding that alone is enough to give us deriving for everything we'd want in an ideal world anyway. I still think it would be great to have, but I've not used it enough to be fully familiar with its capabilities and limitations.

tests/purs/failing/FoldableInstance5.purs
```diff
@@ -1,9 +0,0 @@
--- @shouldFailWith CannotDeriveInvalidConstructorArg
-module FoldableInstance5 where
-
-import Prelude
-import Data.Foldable (class Foldable)
-import Data.Tuple (Tuple(..))
-
-data Test a = Test (Tuple a Int)
-derive instance Foldable Test
```
JordanMartinez, line 9: Shouldn't this be moved to a passing test now?

tests/purs/failing/FoldableInstance7.purs
```diff
@@ -1,9 +0,0 @@
--- @shouldFailWith CannotDeriveInvalidConstructorArg
-module FoldableInstance6 where
-
-import Prelude
-import Data.Tuple (Tuple(..))
-import Data.Foldable (class Foldable)
-
-data Test a = Test (Tuple a a)
-derive instance Foldable Test
```
JordanMartinez, line 9: Same with this one? Should it be moved to the passing tests?

tests/purs/passing/DerivingFunctorPrefersSimplerClasses.purs
```diff
@@ -0,0 +1,46 @@
...
+newtype MonoAndBi a b = MonoAndBi (Effect Unit)
+derive instance Functor (MonoAndBi a)
+instance Bifunctor MonoAndBi where
+  bimap _ _ _ = MonoAndBi (assert' "Bifunctor instance was used but the Functor instance was expected" false)
+
+newtype Test1 a = Test1 (MonoAndBi Int a)
+derive instance Functor Test1
+
+data ExclusivelyBi a b
+derive instance Bifunctor ExclusivelyBi
```
JordanMartinez, line 20: This surprised me. How does a non-inhabited type have a Bifunctor (or later Profunctor) instance?

JordanMartinez (commented):
This looks good to me. Below are just a few comments. I'll re-read your opening thread as I think that's the primary thing to discuss at this point. There could be tests added for the Contravariant things, but I find such tests harder to write/read in practice than the covariant ones. Since the logic is pretty much the same, I'm confident that the contravariant ones work as well.

tests/purs/failing/FoldableInstance5.purs
rhendric: This and the other case are adequately covered (I felt) by [passing/DerivingFunctorFromBi.purs](https://github.com/purescript/purescript/pull/4420/files#diff-d952cbc4d102f894562bbba543ea5dcc61bd58513ddc29b7a11fa86df75e132eR10-R16). (The second case becomes `Test2 (Tuple (Array a) a)`, which is only slightly more sophisticated but exercises the same thing.)

tests/purs/passing/DerivingFunctorPrefersSimplerClasses.purs
rhendric: Same way it has a Functor instance—trivially. `map _ = absurd`, `bimap _ _ = absurd`, `dimap _ _ = absurd`, etc.

Except not literally `absurd`, because uninhabited types aren't literally `Void`. If #2922 were implemented I'd say it's equivalent to that, but instead I'm afraid we just have to accept that the compiler can implement these functions where a user would have to resort to the sort of (mild!) hack used in [`Data.Void`](https://github.com/purescript/purescript-prelude/blob/v6.0.1/src/Data/Void.purs#L31-L31).

tests/purs/failing/FoldableInstance5.purs
JordanMartinez: Ah, good point. I don't think I noticed the `Foldable`/`Traversable` instances there since the file name was specifically for `Functor`. However, using similar feedback as you suggested when I was working on just `Foldable`, shouldn't there be a test to verify the order of the fold works as expected? Or do you think the current `Foldable` tests are sufficient?

JordanMartinez:
> One way to resolve that issue, if it's sufficiently worrisome, would be to move the Bifunctor, Contravariant, and Profunctor classes into purescript-prelude. Another way might be to make the error message more helpful (by suggesting which packages to add to one's project).

I think we should suggest which packages to add to one's project. 

Here are alternatives that we can do and my arguments against them:
- **Don't suggest the packages**. If a user runs into this issue, the error message isn't actionable: can the issue be fixed and if so, what does the user do to fix it? If this is the fix, I think it's worth adding.
- **Add documentation to each compiler-derivable type class**. We could state what is required for each type class to be derived in the type class' documentation. The issue here is that such information isn't present when the error occurs. If I get an error, I'm not going to think, "Check the type class' documentation!"
- **Move the type classes into `prelude`**. I'm curious how this change would affect the dependencies of the core libraries. At the very least, this takes more work to accomplish because of the dependency interplay. I'd be more interested in pursuing this option after the Registry supports monorepos. It is my hope that we store all core libs in a monorepo while still publishing the packages individually. Then, doing such changes like this could be done in a single PR with CI verifying that nothing broke or was missed in that change.

AFAICT, suggesting which packages to install does not have a downside other than perhaps a slightly more verbose error message. Is it possible to only show this suggestion if the source globs passed to `purs` don't contain a file with the module associated with a required type class (e.g. `Data.Bifunctor` for `Bifunctor`)?

JordanMartinez:
On a separate note, should we clarify that a derivation will use the simpler type class when an underlying type has two class' instances that could be used to derive the desired instance? For example, if one was deriving `Functor` over a type that uses `Tuple`, `Tuple`'s `Functor` instance is used rather than its `Bifunctor` instance. Should this point be clarified in the changelog? Or should that not be stated in the changelog?

tests/purs/failing/FoldableInstance5.purs
rhendric: I figured the current tests were sufficient—it's the same code paths being exercised in the `Foldable` and `Bifoldable` cases, as far as the order of the fold is concerned. But I could see an argument for adding those tests anyway, to ensure that if this code ever gets rewritten we don't regress behavior that would be part of the spec if we had one. Any leanings one way or the other?

rhendric:
> If a user runs into this issue, the error message isn't actionable: can the issue be fixed and if so, what does the user do to fix it? If this is the fix, I think it's worth adding.

Yeah, kinda sorta.

For reference, the message currently produced by this PR is:
> To derive this class, make sure that these variables are only used as the final arguments to type constructors, that their variance matches the variance of `Functor`, and that those type constructors themselves have instances of `Data.Functor.Functor`, `Data.Bifunctor.Bifunctor`, `Data.Functor.Contravariant.Contravariant`, or `Data.Profunctor.Profunctor`.

I think it'd be fairly easy to detect if any of those mentioned classes are missing from the current project and add an addendum (filtered down to the missing packages) to this message like:

> (It may help to add `purescript-bifunctors`, `purescript-contravariant`, or `purescript-profunctor` to your project's dependencies.)

But note that in ‘most’ cases, it won't help to follow this advice: adding the dependency only solves the problem when the missing instance needed is in that dependency. We're not talking about many instances here:

* In `purescript-contravariant`:
  * <code>[Contravariant](https://pursuit.purescript.org/packages/purescript-contravariant/6.0.0/docs/Data.Functor.Contravariant#t:Contravariant) ([Const](https://pursuit.purescript.org/packages/purescript-const/6.0.0/docs/Data.Const#t:Const) a)</code>
* In `purescript-bifunctors`:
  * <code>[Bifunctor](https://pursuit.purescript.org/packages/purescript-bifunctors/6.0.0/docs/Data.Bifunctor#t:Bifunctor) [Either](https://pursuit.purescript.org/packages/purescript-either/6.0.0/docs/Data.Either#t:Either)</code>
  * <code>[Bifunctor](https://pursuit.purescript.org/packages/purescript-bifunctors/6.0.0/docs/Data.Bifunctor#t:Bifunctor) [Tuple](https://pursuit.purescript.org/packages/purescript-tuples/7.0.0/docs/Data.Tuple#t:Tuple)</code>
  * <code>[Bifunctor](https://pursuit.purescript.org/packages/purescript-bifunctors/6.0.0/docs/Data.Bifunctor#t:Bifunctor) [Const](https://pursuit.purescript.org/packages/purescript-const/6.0.0/docs/Data.Const#t:Const)</code>
* In `purescript-profunctor`:
  * <code>[Profunctor](https://pursuit.purescript.org/packages/purescript-profunctor/6.0.0/docs/Data.Profunctor#t:Profunctor) [Function](https://pursuit.purescript.org/builtins/docs/Prim#t:Function)</code>

So if `FooBar` isn't `Const`, `Either`, `Function`, or `Tuple`, this suggestion isn't going to fix things for the user. What they need to do in the general case is check that the conditions in the first part of the error message are all met.

We could hard-code this list of instances in the compiler, which would at this point in time provide the user with the best actionable suggestions, but of course the list might change in the future. I don't expect it to change much or often, though; maybe this is best?

> On a separate note, should we clarify that a derivation will use the simpler type class when an underlying type has two class' instances that could be used to derive the desired instance?

I think this belongs in the language documentation instead of the changelog, but yes, I intend to fully specify what users can expect from the compiler when using this feature.

JordanMartinez:
> So if the type constructor being used isn't Const, Either, Function, or Tuple, this suggestion isn't going to fix things for the user. What they need to do in the general case is check that the conditions in the first part of the error message are all met.

Ah... in that case (i.e. most usages wouldn't benefit from this), I think it shouldn't be added. Rather the documentation on this feature should clarify that one might need to install the corresponding dependencies to get the instances needed for the derivation to work.

JordanMartinez (approved):
Nice work!

rhendric:
I've updated purescript/documentation#446 to cover these enhancements; I believe it mentions all the things we've said should be mentioned here so far.

rhendric:
I'll merge this in a couple of days unless there are any last-minute comments.

Issue #4413, opened 2022-11-01
Improve TypesDoNotUnify message for record types


rpolley:
## Summary

We should call out what difference in the records caused the match to fail in addition to the current output.

## Motivation

When making heavy use of row polymorphism in a large codebase one often ends up with functions on record types that expect some number of labels with a generic row tail, which turns the error messages for two row types not matching into a bit of a spot the differences game, and this gets worse and worse the more stuff you add to these, or if you start nesting records.

## Proposal

Call out what didn't match in the error message, i.e. something like 'expected the label "label :: Type"', or "type cannot have the label "label :: Type"'.

## Examples

```
Type Stuff s =
   { foo: Foo
  , bar: Bar
  , baz: Baz
  | s
  }
```
Let's say you have a function with the following type
```
doThingWithStuff :: forall s. Stuff s -> SomeOtherType
```
Suppose also that you have a type
```
type Stuff2 s =
{ foo: Foo
, bar: Bar
, qux: Qux
, quaz: Quaz
| s
}
```
but which is missing the `bar :: Bar` label, and a value `stuff2 :: Stuff2 ()`. Now say you pass `stuff2` to `doThingWithStuff`. The error message isn't directly helpful at finding the actual problem, which is that `Stuff2` is missing the label `bar :: Bar`. The error message should mention that the types failed to unify because the first has a label `bar :: Bar` and the second doesn't.

Now let's say you also have a function `mkStuff :: StuffPrecursor -> Stuff ()` and you put in an extra label. It would be helpful if the compiler told you that your record has the extra label directly instead of giving a generic row TypesDoNotUnify message.

MonoidMusician:
There are already errors for this, but it looks like they are only used during subsumption, not during unification (see the `MissingRecordField` and `ExtraRecordField` tests):
https://github.com/purescript/purescript/blob/b7e05967ecec890c1d45d266e221bc877a87f184/src/Language/PureScript/TypeChecker/Subsumption.hs#L105-L114

https://github.com/purescript/purescript/blob/b7e05967ecec890c1d45d266e221bc877a87f184/src/Language/PureScript/TypeChecker/Unify.hs#L163-L180

The error message would need some tweaking, since it isn't saying the right thing for unification, but I think it is a good idea to do what we can to improve these errors.

[...]

Issue #3399, opened 2018-07-10
Improve error message for non-matching types


i-am-the-slime:
I never know what the compiler wants from me when it says:

`Could not match type A with type B`

Which one does it expect? Which one is the one I supply? 
For people like me it's not always obvious which type I am supplying.
I don't really see any downside to changing this to:

`Could not match expected type A with actual type B`

I don't know about any tooling that might depend on this exact message.

i-am-the-slime:
I think it could be even nicer if the compiler could insert a little (?helpme (thingOfTheWrongType)) and suggest possible fixes but that's most probably out of scope.

garyb:
We've had issues about this previously, and did attempt to change the error to work this way... but unfortunately the type checker isn't able to know for sure whether A or B is the type that is actually desired.

hdgarrood:
Yeah, I’m certain there is already an issue for exactly this, possibly even open.

i-am-the-slime:
I see. Sorry guys!

i-am-the-slime:
Actually maybe we could make that explicit in the message? I remember trying to find the order by making some minimal example. Since there is no order this could be misleading. I know there is nothing in the message suggests that there is an order but still. The message could have a little "in no particular order" disclaimer.

hdgarrood:
Previously: #3111 #1577 #1583

hdgarrood:
Actually sorry, I take back what I said before: while we do have discussion which notes that identifying expected vs actual types would be useful, there doesn’t appear to be an open issue for actually doing so.

hdgarrood:
So we might want to reopen this perhaps?

JordanMartinez:
I think this should be reopened. At the very least, it would track that the issue does exist.

ajnsit:
I wrote this on discord, but thought I would add it as a suggestion here as well -

Looking at the error message in this issue https://github.com/purescript/purescript/issues/1577 - Some very useful information *is* partly present but buried - "checking that expression 4 has type String". Is it possible to put that information up front?

For example the error messages could say something like -

```
Error occurred while checking that
expression
    4
  has type
    String

Could not match type
    Int
  with type
    String
while trying to match type Int
  with type String

in value declaration foo
```

PR #4421, opened 2022-11-29
Better row unification error messages

FredTheDino:
**Description of the change**

Related to #4413, and part of a potential solution.

The idea was to show which record labels are different. Sometimes, you have typos or a bunch of fields that mostly unify. This could help with that.
I'm hesitant to make more work on this unless there's interest. So let me know if this is a change people want to see. I know this or something that's feature adjacent.

This simple program:
```
module A where

-- As we can see, Stuff > Thing
type Stuff s = { foo :: Int | s }
type Thing s = { foo :: Int, bar :: Int, baz :: Int | s }

doStuff :: forall s. Stuff s -> Int
doStuff { foo, bar } = bar

doThings :: forall s. Thing s -> Int
doThings thing = thing.bar
```

Now compiles with the error:
image

This is a toy example. But it makes it very clear that the `bar` label is what is causing the unification to fail.
I think it's an improvement which changes the errors for the better.

:fire: Roast me! :fire:
---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/fix_better-row-error-messages.md
+++ b/CHANGELOG.d/fix_better-row-error-messages.md
@@ -8,11 +8,14 @@
   and `AdditionalProperty` have been removed, being replaced
   with `TypesDoNotUnify` instead.
 
-  **Case 1: lack of expected/actual text when that is known**
+  **Case 1: lack of "expected type" distinction when it is known**
 
-  The expected/actual distinction is not possible to determine in all cases,
-  but it can be known in some cases. Previously, it was not reported even 
-  if it was known. Now the compiler reports this distinction if known.
+  Previously, the message would not indicate which
+  of the two types was the expected type, even
+  if it could be determined.
+  
+  Now, when this distinction is known, it is reported. 
+  Note: this distinction cannot be known in all circumstances.
 
   Before:
   ```


rhendric:
Thanks for taking a crack at this!

My general objection to changing error messages from

```
Error:
    AAAAAAAAAA
      BBBBBB
      CCCC
    DDDDDDD
        EEEE
```

to

```
Error:
    AAAAAAAAAA
      BBBBBB  <<<LOOKATMELOOKATME
      CCCC
    DDDDDDD
        EEEE
```

is the following: are lines A, C, D, or E ever the thing a user needs to look at? If not, the better change would be

```
Error:
  BBBBBB
```

If not, then this change is not an improvement; it's just moving the confusion around. Some users get directed to the important part of the error for them, and other users get misdirected away from the important part of the error for them.

In this specific instance, consider the following fragment:

```purs
f :: forall r. Identity { foo :: Int | r } -> Int
f (Identity v) = v.foo

x :: Identity { foo :: String, bar :: Int }
x = Identity { foo: "hi", bar: 5 }

y :: Int
y = f x
```

The error currently produced says:

```
  Could not match type

    String

  with type

    Int


while trying to match type
                             ( bar :: Int
                             , foo :: String
                             ...
                             )

  with type
              ( foo :: Int
              ...
              | t0
              )
```

IIUC, your PR would make this

```
while trying to match type
                             (*_bar_ :: Int
                             , foo :: String
                             ...
                             )

  with type
              ( foo :: Int
              ...
              | t0
              )
```

which makes the hint less helpful, as there's nothing wrong with `bar`.

FredTheDino:
Thanks for the feedback @rhendric! :D

I agree it's not the most elegant solution. I also dislike the visual clutter this causes. But I don't agree that the information isn't helpful or moves the issue. The authors intent of this code, is unclear. It could be that the `bar` field is missing, and that's what is important. 
Consider this code:

```haskell
doStuff :: { a :: Int, b :: Int, c :: Int, d :: Int, e :: Int, f :: Int, h :: Int, j :: Int } -> Int
doStuff { a, b, c, d, e, f, g } = 1
```

When i look at this code, I think it looks decently right. On my first skim I would not immediately notice the error.

Gives this error:
```
  Could not match type

    ( a :: Int
    , b :: Int
    , c :: Int
    , d :: Int
    , e :: Int
    , f :: Int
    ,*_h_ :: Int
    ,*_j_ :: Int
    ...
    )

  with type

    ( a :: t7
    , b :: t6
    , c :: t5
    , d :: t4
    , e :: t3
    , f :: t2
    ,*_g_ :: t1
    ...
    | t8
    )
```
The difference in the fields is clearly given. I've had to manually match these fields where there are upwards of 70 rows and 1 row missing. This very simple tool trivializes the entire task of spotting the difference in the label names.

Compare this error to:
```
  Could not match type

    ( a :: Int
    , b :: Int
    , c :: Int
    , d :: Int
    , e :: Int
    , f :: Int
    , h :: Int
    , j :: Int
    ...
    )

  with type

    ( a :: t7
    , b :: t6
    , c :: t5
    , d :: t4
    , e :: t3
    , f :: t2
    , g :: t1
    ...
    | t8
    )
```
And it would already save me as a program writer brain power and thinking. Thinking that should go into the code I'm writing.

I definately understand the visual clutter point. We've discussed that earlier. I think there's more we can to do to remove the clutter. I for one is for removing the entire
```
where t0 is an unknown type
      t8 is an unknown type
      t7 is an unknown type
      t6 is an unknown type
      t5 is an unknown type
      t4 is an unknown type
      t3 is an unknown type
      t2 is an unknown type
      t1 is an unknown type
```
part.

But I think the compiler error messages are in a form of local optimum. They
are not the best version they can be in. And to get the best version possible
we might have to make alterations which don't solve all problems we are having right now.

Some other ideas of getting forwards:
  - Add this as a special case in the error messages, if all the types are unknown only show the different labels
  - Survey the community for bad errors, examples of where the compiler mislead them
  - Add this and revert it if people complain
  - Maybe someone has a suggestion which makes this change better

It might be that we should put up goals for what we want the errors to be. But
doing something is better than doing nothing. :)

rhendric:
My point is that for every example you come up with where this change helps the user find the problem, I can come up with an example where this same change makes it harder for the user to find the problem. That's not a global improvement. That's, at best, throwing some users under the bus to make other, hopefully more plentiful users happier—and I'm not yet convinced that the majority of cases are on your side.

You don't have to solve this problem perfectly in order to make progress on it. You just have to do no harm in the process.

For example, just looking at these errors:

```
  Could not match type

    ( a :: Int
    , b :: Int
    , c :: Int
    , d :: Int
    , e :: Int
    , f :: Int
    , h :: Int
    , j :: Int
    ...
    )

  with type

    ( a :: t7
    , b :: t6
    , c :: t5
    , d :: t4
    , e :: t3
    , f :: t2
    , g :: t1
    ...
    | t8
    )
```

it seems likely that we could make more progress on unification, matching up the `a` through `f` labels, before raising the error. That would result in the error message being smaller (because the `...` already hides the part of the row that is equal to the other row) *and* wouldn't mislead users when one of the types in `a` through `f` is not a match—a win for everyone. That may not be a perfect solution either but it would make progress in a way that does no harm, and assuming it works I'd support it.

FredTheDino:
I don't know the inns and outs of the type checker perfectly. But reading @MonoidMusician s comment:
> The tricky part is that unification mismatches can apply in e.g. contravariant contexts, so you don't actually know which labels was missing and which labels were extraneous, you can only see that there was some mismatch. (See also the infamous https://github.com/purescript/purescript/issues/3399.)

https://github.com/purescript/purescript/issues/4413#issuecomment-1299515563

I figured that is very hard to unify more things in the type checker. This lead me to not investigate those solutions. I'm not sure it would be possible.

I don't know if it's possible to get better error messages without "doing harm", much less so in increments. ([relevant XKCD](https://xkcd.com/1172/))

I believe adding more information is doing the least harm.

It sounds a lot like you want a complete revolution of the error messages or for them to be left alone. Something that unfortunately won't come out of making low hanging fruit changes in places where users are complaining. 

My reasoning went something like this:
  - The type errors display label mismatch and type mismatch (which kinda overlap)
  - Sometimes the labels are the problem (we cannot know this)
  - Sometimes the types are the problem (we cannot know this)
  - The current error messages assumes the problem is in the types of the rows
  - Annotating which labels differ gives more information for the second kind of problem which is currently under represented

The goal was to be as un-intrusive as possible with `*`. Maybe a smaller symbol would remove the visual noise for you? Maybe run this for a day and see how it feels? Is the problem really the symbol or that more information is added (which I suspect is unavoidable).

We could add another section to the error saying which labels differ there, if any. Is that a better solution?

rhendric:
> > The tricky part is that unification mismatches can apply in e.g. contravariant contexts, so you don't actually know which labels was missing and which labels were extraneous, you can only see that there was some mismatch. (See also the infamous #3399.)
> 
> [#4413 (comment)](https://github.com/purescript/purescript/issues/4413#issuecomment-1299515563)
> 
> I figured that is very hard to unify more things in the type checker. This lead me to not investigate those solutions. I'm not sure it would be possible.

The thing being highlighted as difficult in that comment is telling the difference between a missing label that's needed or an extraneous label being provided. Simply proceeding with unification to eliminate more unknowns shouldn't run up against any intractable problems, though it might not resolve everything.

> I don't know if it's possible to get better error messages without "doing harm", much less so in increments. ([relevant XKCD](https://xkcd.com/1172/))

That's not as relevant as you think it is. I'm not talking about ‘breaking workflows’; we've done plenty of major releases where we've changed quite a lot about how users' workflows work, and when we're working towards a better world that is a decision we're comfortable making. That isn't this. We're talking about an error message that, conceptually, is saying, ‘Hey user, you might have problem A or you might have problem B.’ You're proposing adding information that biases the user toward looking at problem A. That disadvantages the users who have problem B. That's not a better world to be worked toward, unless you're myopically focused on the users with problem A.

> It sounds a lot like you want a complete revolution of the error messages or for them to be left alone.

No idea where you're getting that but that's definitely not what I want. A complete revolution is neither necessary nor desirable here. Leaving things alone is tolerable but not ideal either. I'm glad you're stepping up to work on the problem; I want things to improve, and I want to help you contribute!

Working on the problem, however, involves considering the full space of cases that might trigger this error message, and making sure that in improving parts of the space you aren't making the message less helpful in others. It can be a difficult balancing act, and that's probably the biggest reason why nobody has fixed this yet.



> The goal was to be as un-intrusive as possible with `*`. Maybe a smaller symbol would remove the visual noise for you? Maybe run this for a day and see how it feels? Is the problem really the symbol or that more information is added (which I suspect is unavoidable).
> 
> We could add another section to the error saying which labels differ there, if any. Is that a better solution?

In this case I'm not concerned about how noisy the indicator is. I'm concerned that the indicator in the error message makes it read as if that's the thing the user needs to look at to solve their problem. If you can figure out a way to format the diff so that it's obvious that you're just highlighting some rows that are different and that this isn't necessarily what the user needs to focus on, I could be okay with that. Maybe a second section would work. Just keep in mind that when open rows are involved, frequently extra or missing labels are totally fine and users—particularly beginners—shouldn't be misled by the error message into adding properties to their objects that they don't need, for example. For this reason, I'm not very optimistic about this approach, but you're welcome to take a stab at it and see if your creativity exceeds my imagination.

JordanMartinez:
> Actually, maybe a side-by-side diff of rows (without any additional highlighting or indicators) would do the trick here?

That's what I was thinking after reading through the above comments. The issue isn't the error per say but that jumping back and forth between really long rows makes it hard to see what is different about them. Consider the following examples of the same error shown in the comment, https://github.com/purescript/purescript/pull/4421#issuecomment-1331264229:

<details>
<summary>Side-by-Side</summary>

```
Could not match type on left with type on right

    ( a :: Int      ( a :: t7
    , b :: Int      , b :: t6
    , c :: Int      , c :: t5
    , d :: Int      , d :: t4
    , e :: Int      , e :: t3
    , f :: Int      , f :: t2
                    , g :: t1
    , h :: Int
    , j :: Int
    ...             ...
                    | t8
    )               )

```

</details>

Downsides:
- when the labels and/or their types are really long, this won't display well

<details>
<summary>Row comparison</summary>

```
  Could not match rows in type

    ( a :: Int
    ( a :: t7
    
    , b :: Int
    , b :: t6
    
    , c :: Int
    , c :: t5
    
    , d :: Int
    , d :: t4
    
    , e :: Int
    , e :: t3
    
    , f :: Int
    , f :: t2
    
    <no row>
    , g :: t1
    
    , h :: Int
    <no row>

    , j :: Int
    <no row>

    ...
    
    <no tail>
    | t8
    )
```

</details>

Downsides:
- the error itself may span quite a large number of lines in cases where one type has a large number of rows. Considering that this already happens, maybe that's ok?

FredTheDino:
Great to have more input!

I've been thinking about this quite a bit for the last days. I you're both right that a more diff like approach is better. Particularly the second example that Jordan pointed out. I'm gonna look at implementing that.

But I see some perils on the road to better error messages still.  Switching to more of a "diff approach", might make it hard for new users to read the error message. Maybe some extra indentation could help, or maybe color coding the types differently might help.

A more spread out type error might cause confusion of which row is part of which type. Spreading the type out might make it hard to see what belongs to what. Changing the shape of the error so drastically might also make it harder for people more used to reading the old errors.

I'm gonna try to implement the more diff-like approach and we'll see how that turns out. I'm sure we're missing something there as well. :)

FredTheDino:
I've rewritten the errors to now display a diff. Should be easier to see typos. This is what it looks like now:
image

I made some alterations to what Jordan suggested, and this solution is not complete. Tails aren't handled and the record/row open and closing brace aren't correct. Just to name a few.  

I'll let this sit for a while, and see what people feel about it.

JordanMartinez:
> I made some alterations to what Jordan suggested, it is a different solution but maybe it has something in it that is relevant or interesting.

Could you clarify what those alterations were?

> I'm not entirely sure about the unified view of the error. It makes it kinda hard to know which type is which.

There's another issue open for better indicating what was the 'expected' type and what was the 'actual type'. But, I do think this error is overall better than what we currently have.

Also, I think there should be a few golden tests:
- one row has a label whereas another one doesn't
- all labels are the same but...
    - one label's type is `Int` whereas another is `String`
    - one label's type is a really long type while the other is a similarly long type and both don't unify (i.e. does the rendering function properly account for long type names, and does the overall look still look good?)
    - one label's type is `Int` whereas another is `{ foo :: String }` (i.e. some initial record nesting occurs)
    - one label's type is `{ bar :: Int, foo :: String }` whereas another is `{ foo :: String }` (i.e. in a nested record, a label is missing)
    - one label's type is `{ foo :: Int }` whereas another is `{ foo :: String }` (i.e. in a nested record, types in the same label don't unify)

You can look at [Writing tests](https://github.com/purescript/purescript/blob/master/CONTRIBUTING.md#writing-tests) if you need clarification on how to do that.

FredTheDino:
TL;DR: I'll gladly implement this more thoroughly. But is this likely to get merged?

> Could you clarify what those alterations were?
I removed one of the parenthesis in the start. I also changed some of the layout a bit. I also haven't added suppor t for the RowType tails. But most of this is just "this was faster to implement and I don't want to write code that isn't used".

> There's another issue open for better indicating what was the 'expected' type and what was the 'actual type'. But, I do think this error is overall better than what we currently have.

I might be overly sceptic of the changes, but I'm not sure this change will go over well with everyone. Some of the things I've noted:
 - The extra whitespace will make the rows take more space then previous
 - The new errors are not easily copy-pastable if you want to use this to e.g. generate a type
 - New users that use other typed languages might be more used to "traditional" errors where the types aren't interlaced, this might make the learning curve steeper for an already tough language
  - The 2 developers I've shown this too didn't understand the error at first glance

There are of course some pros to this new way of showing.
 - Easier to see what fields differ
 - Very apparent when you make a typo
 - It's cool

This doesn't convince me that we:
> make progress in a way that does no harm

Maybe my list of cons is too long. I'm all for digging deeper. I'm personally almost at the point where I consider doing a survey in the Discord. I'd love to see changes and I'll happily do them, but I want them to be changes that are to the benefit of everyone.

I'll add as many tests as we can think of, but is there value in me working more on this?

rhendric:
Can't speak for Jordan, but in my opinion work on design is less likely to be wasted than work on tests at this stage.

I still think the most straightforwardly beneficial approach would be allowing row unification to continue past the first mismatch, though, and then seeing whether work on changing error rendering is still needed.

rhendric:
In fact, doing nothing more than changing
https://github.com/purescript/purescript/blob/df5fcff1c396d520e8543d5d85ce1455e56e2696/src/Language/PureScript/TypeChecker/Unify.hs#L179-L180

to the following:

```haskell
  unifyTails r1' r2' =
    throwError . errorMessage $ TypesDoNotUnify (rowFromList r1') (rowFromList r2')
```

makes the error generated in https://github.com/purescript/purescript/pull/4421#issuecomment-1331240301 become the following:

```
         Could not match type
                     
           ( h :: Int
           , j :: Int
           ...       
           )         
                     
         with type
                    
           ( g :: t1
           ...      
           | t2     
           )        
```

I suspect that doesn't cover 100% of the cases you're concerned about, but I recommend investigating that sort of improvement first (you may need to make a few other similarly-sized tweaks in that area; there's a `sequence_` in there that I suspect may want to be a `parU`) and seeing how close it gets to your ideal.

JordanMartinez:
> I might be overly sceptic of the changes, but I'm not sure this change will go over well with everyone. Some of the things I've noted:

Hmm... those are all good points, and it further highlights why resolving issues like this are hard.

> Can't speak for Jordan, but in my opinion work on design is less likely to be wasted than work on tests at this stage.

I agree with this. My comment above was more mentioning tests that should be added at some point / cases to consider when designing the final error. However, @rhendric is right in that writing tests shouldn't be done until this design is better flushed out.

FredTheDino:
I have now made the changes that @rhendric suggested. The tests noted a change in one place, and I would say it's equally good or an improvement. This change doesn't solve my entire problem, but it goes part of the way there.

I would propose merging this change (I can make a new PR if that's easier) and then look into potential changes to error messages given the better unification.

I also looked through all other `errorMessage` calls to see if there was any other low-hanging fruit - but since I don't know what I'm doing, I didn't find anything notable.

rhendric:
At this point you need tests. What cases were you looking at that you wanted to see improvements on? In particular, did you have a case that didn't look good without the `sequence_` to `parU` change but did look good afterwards? (You shouldn't blindly follow my hunch without testing it. I could commit and push my own hunches if that's all that we needed to make progress! :smile:)

Also, are there any downsides to making this change? Can you explain what they are or why they don't exist?

FredTheDino:
> Also, are there any downsides to making this change? Can you explain what they are or why they don't exist?

TL;DR: These changes improve the error messages somewhat.

From how I understand the code - which is probably faulty - I see this as a safe change. Sending out other type information than the original types should not change the correctness of the compiler. The error messages I've seen, given all the unit tests in this code and from all programs I've run I've never found a "worse error message". That said, there are differences. Now the compiler exposes less type information displayed in the error. The more focused errors makes it easier for me to understand the type error at a glance.

We should also consider `the harm` of this change. It is a change, there will be friction. It's less friction than the previous changes. I think it's less likely to mislead someone.

*Changing the returned error*
I cannot reason about all error outputs for all programs. But I can guess. My guess is that most people who are reading error messages want to know the difference between the type and the type the compiler expects. Showing the unified rows creates a more focused error.

*Changing the `sequence` into a `parU`*
Again, this doesn't change correctness, finding an error in parallel or in sequence still finds the error. But this is the more controversial of the changes. But finding errors in parallel could potentially show more errors earlier, saving the developer a recompile and giving context about other errors. I would expect the compiler to notice all errors in a row if it's been bothered to check it. That said, it also produces more output. I still think this error is better than the previous.

> In particular, did you have a case that didn't look good without the sequence_ to parU change but did look good afterwards?
Yes, see tests.

> I could commit and push my own hunches if that's all that we needed to make progress!
I'm well aware. I see this as a learning opportunity. Not every day you get to poke around in a compiler. <3

To finish this in a truly scientific manner, I still think more work needs to be done in this area. But it's hard to see these changes in isolation. So I'm not going to open another PR until #4421 and #4411 are merged. :dango:

tests/purs/failing/MultipleErrorsInOneRow.out
```diff
@@ -0,0 +1,76 @@
+Error 1 of 2:
```
FredTheDino, line 1: Before the error this output only had one error. Now we have two. I see this as a win. But it could also compound if we have a large nested tree of almost errors:
```
{ a :: A, b :: A, c :: A } <=> { a :: A', b :: A', c :: A' }
```
This might make the errors hard to read. I prefer this, but nothing in this world is objective. But here I also assume we stop outputting errors after, like, 10.

FredTheDino:
I'd love to see this merged. What more do you want me to do?

src/Language/PureScript/TypeChecker/Unify.hs
```diff
@@ -176,8 +176,8 @@ unifyRows r1 r2 = sequence_ matches *> uncurry unifyTails rest where
     rest' <- freshTypeWithKind =<< elaborateKind (TUnknown a u1)
     solveType u1 (rowFromList (sd2, rest'))
     solveType u2 (rowFromList (sd1, rest'))
-  unifyTails _ _ =
-    throwError . errorMessage $ TypesDoNotUnify r1 r2
+  unifyTails r1' r2' =
+    throwError . errorMessage $ TypesDoNotUnify (rowFromList r1') (rowFromList r2')
```
rhendric, line 180: A different approach to consider here would be:
```suggestion
  unifyTails _ _ = do
    subst <- gets checkSubstitution
    throwError . errorMessage $ TypesDoNotUnify (substituteType subst r1) (substituteType subst r2)
```

This results in slightly different error messages—the upshot, thanks to row-diffing code that lives in the pretty-printer for types, is that in cases like 3701 the `...` appears when printing the rows that don't match with the `substituteType` approach but not with this approach.

Personally, I think having the `...` might be clearer. What do you think?

tests/purs/failing/NestedRecordTypeError.out
```diff
@@ -0,0 +1,41 @@
...
+  [33m                   [0m
+  with type
+  [33m                  [0m
+  [33m  ( b :: { z :: t0[0m
+  [33m         | t1     [0m
+  [33m         }        [0m
+  [33m  ...             [0m
+  [33m  | t2            [0m
+  [33m  )               [0m
+  [33m                  [0m
```
rhendric, line 20: Shouldn't there be a `g` mentioned somewhere in here? What's up with that?

tests/purs/failing/MultipleErrorsInOneRow.purs
```diff
@@ -0,0 +1,8 @@
+-- @shouldFailWith TypesDoNotUnify
+-- @shouldFailWith TypesDoNotUnify
+module A where
+
+-- Find both these errors
+f :: { a :: String, b :: String } -> Int
+f { a: 0, b: 0 } = 0
```
rhendric, line 7: Could you make the expected and actual types between `a` and `b` different, so that the two error messages are distinct?

tests/purs/failing/MultipleErrorsInOneRow.out
rhendric: I don't *think* we have any such limit, at least in the compiler. (Maybe IDE plugins do that filtering.) Would that alter your opinion about this change?

src/Language/PureScript/TypeChecker/Unify.hs
FredTheDino: I will try it when I have more free time than I have now. :)

tests/purs/failing/MultipleErrorsInOneRow.out
FredTheDino: There's one problem here with outputting many more errors, but these errors are also a lot simpler than the original error. From reading these errors we can see that we have two errors in the types, which is valid. But for this to be a strict improvement I would want the source spans to be better.

tests/purs/failing/NestedRecordTypeError.out
FredTheDino: Both yes and no. Since the unification is done on each field we are reporting on a record field. In a perfect world I would like it to comment `while matching row 'a'`.

That said, I would expect a second error complaining about `g`. So this definitely warrants further investigation.

i-am-the-slime:
What's the status of this? I have extremely long messages when trying to `Union` some rows, will this help with this?

JordanMartinez:
@i-am-the-slime There's the two points Ryan mentioned above that have not yet been addressed (at least, to my knowledge).

I think there's also this point made by Ryan:

> What cases were you looking at that you wanted to see improvements on? In particular, did you have a case that didn't look good without the sequence_ to parU change but did look good afterwards? (You shouldn't blindly follow my hunch without testing it. I could commit and push my own hunches if that's all that we needed to make progress! smile)
>
> Also, are there any downsides to making this change? Can you explain what they are or why they don't exist?

tests/purs/failing/NestedRecordTypeError.out
JordanMartinez: I looked into this. It's caused by not catching any possible errors thrown from `parU matches id` **and** `uncurry unifyTails rest`. Because `a` is a label shared between the two rows, `parU matches id` will recurse down on the `a` label. And because the rows under the label `a` do not unify, an error is thrown. Because the error short-circuits this computation, the `uncurry unifyTails rest` code is never evaluated. Thus, the second error does not appear.

tests/purs/failing/MultipleErrorsInOneRow.purs
JordanMartinez: Done.

tests/purs/failing/NestedRecordTypeError.out
JordanMartinez: This issue has been resolved.

src/Language/PureScript/TypeChecker/Unify.hs
JordanMartinez: See https://github.com/purescript/purescript/pull/4421/commits/d4cf7215379239190d77f2be11b27fd040447fe4 for what this looks like. To summarize, while it adds the `...` in case 3701, it obfuscates the result for cases like `NestedRecordTypeError`. 

I initially thought we could add a hint before I realized that a hint is already added. The reason why it doesn't appear in this case is because the `ErrorUnifyingTypes` hint is stripped via `stripRedundantHints` when the error is `TypesDoNotUnify`. 

We could stop stripping that hint, but I don't think that tradeoff is worth it in this context.

JordanMartinez:
> What cases were you looking at that you wanted to see improvements on? 

In https://github.com/purescript/purescript/pull/4421/commits/ed9bcd9aadf12c23cb79916e92bd6401503bf483, I outputted `master`'s current output for the tests added here. I'd say the changes here lead to a net gain. 

> In particular, did you have a case that didn't look good without the sequence_ to parU change but did look good afterwards? (You shouldn't blindly follow my hunch without testing it. I could commit and push my own hunches if that's all that we needed to make progress! smile)

Changing `sequence_` to `parU` is necessary for the `MultipleErrorsInOneRow.purs` test. I mistakenly excluded that test when I was producing outputs for the various approaches here. Before https://github.com/purescript/purescript/pull/4421/commits/958c9547265d23c1214ef1ae1e76d6c532e197cb, the tests fail. Adding that change causes the tests to pass. Otherwise, only one problem rather than all problems of shared labels at the same depth in a record are reported.

> Also, are there any downsides to making this change? Can you explain what they are or why they don't exist?

So far, one downside with this PR's changes is `3701` losing the `...`. Perhaps stripping `ErrorUnifyingTypes` may not always be desirable when printing row-related things. Or perhaps, we should always include the last `ErrorUnifyingTypes` if there are multiple such hints rather than removing all of them. I'll explore doing this next.

JordanMartinez:
> So far, one downside with this PR's changes is 3701 losing the .... Perhaps stripping ErrorUnifyingTypes may not always be desirable when printing row-related things. Or perhaps, we should always include the last ErrorUnifyingTypes if there are multiple such hints rather than removing all of them. I'll explore doing this next.

Exploring this a bit further, we can choose to not strip `ErrorUnifyingTypes` when the `TypesDoNotUnify` error's types are `REmpty`:
```diff
-stripRedundantHints (TypesDoNotUnify t1 t2) = stripFirst isUnifyHint
+stripRedundantHints (TypesDoNotUnify t1 t2)
+  | isREmpty t1 || isREmpty t2 = id -- do not strip
+  | otherwise = stripFirst isUnifyHint
  where
  isUnifyHint ErrorUnifyingTypes{} = True
  isUnifyHint _ = False
```

Unfortunately, because `ErrorUnifyingTypes` is a `CheckHint` category, it causes the other hint coupled with that message, `ErrorInApplication`, to be filtered out, which removes this part of the error message for 3701:
```
while applying a function [33mfooMerge[0m
  of type [33mUnion @Type t0               [0m
          [33m  ( thing1 :: String         [0m
          [33m  , thing2 :: Int            [0m
          [33m  )                          [0m
          [33m  ( thing1 :: String         [0m
          [33m  , thing2 :: Int            [0m
          [33m  | t0                       [0m
          [33m  )                          [0m
          [33m => Nub @Type                [0m
          [33m      ( thing1 :: String     [0m
          [33m      , thing2 :: Int        [0m
          [33m      | t0                   [0m
          [33m      )                      [0m
          [33m      ( thing1 :: String     [0m
          [33m      , thing2 :: Int        [0m
          [33m      | t0                   [0m
          [33m      )                      [0m
          [33m     => Record t0            [0m
          [33m        -> { thing1 :: String[0m
          [33m           , thing2 :: Int   [0m
          [33m           | t0              [0m
          [33m           }                 [0m
  to argument [33m{ thing1: "foo"[0m
              [33m}   
```
I think fixing that would require changing how we strip hints and whatnot, and I don't think that should be done in this PR.

src/Language/PureScript/TypeChecker/Unify.hs
JordanMartinez: >  I would have thought that getting the current substitution would have filled in all those unknowns.

If I had to guess, I'd say it's because we haven't solved those _yet_. We run `parU matches id`, which iterates through the list in left-to-right order. If `a` is the first label, we haven't done `unifyTypes` on the other labels yet, so some unknowns may not have been solved yet. Thus, when the error for `a` is thrown, labels, `d`, `e`, and `f`, are still unknown.

JordanMartinez:
Another outstanding issue is that this PR does not have a changelog entry. I'll refrain from writing one. Its content may change depending on what Ryan finds in his exploration of this PR ([see comment](https://github.com/purescript/purescript/pull/4421#discussion_r1286829376).

src/Language/PureScript/TypeChecker/Unify.hs
rhendric: Nope, it's the double `parU`. If I unify them with
```
void $ parU (matches ++ [ uncurry unifyTails rest ]) id
```
then I get something closer to what I was expecting.

Downside of *that* (she swallowed the spider to catch the fly) is that, still in `NestedRecordTypeError`, now the `unifyTails` error message contains information redundant with the other error raised when aligning rows. I have a plan for that; I'll post when I'm happy with it.

rhendric: Jordan, take a look at the commit I just pushed? It's a little more baroque than simply using `parU`, but I think it improves the error messages.

JordanMartinez:
Nice!

JordanMartinez:
Thoughts on what to put for the changelog entry? This otherwise LGTM.

JordanMartinez (approved):

tests/purs/failing/RowOneMissingLabel.purs
```diff
@@ -0,0 +1,6 @@
+-- @shouldFailWith PropertyIsMissing
+module A where
+
+-- Currently only reports on the missing `c` label
```
rhendric, line 4: Do we want to do something about this as well? This case looks like it'll be handled in `TypeChecker.Subsumption`.

JordanMartinez: Ahh, yeah we should.

JordanMartinez: There's a similarly-named [`unifyRows` in Kinds.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker/Kinds.hs#L447-L464). Should that also be updated to this approach? I'm not sure what error messages it would affect, but I imagine it would be desirable.

JordanMartinez: > Maybe not this exact approach, but might be good to parU the labels at least. I don't think you get here without doing some pretty advanced stuff; something like

I wanted to compare the before/after of using the same approach. Here's what I get:

<details>
<summary>Ryan's above code</summary>

```purs
-- @shouldFailWith KindsDoNotUnify
module RowKindTest where

import Type.Proxy (Proxy(..))

foreign import data Foo :: Record (a :: Type, b :: Type)
foreign import data Bar :: Record (a :: Int, c :: Type, d :: Type) -> Type

something :: Proxy (Bar Foo)
something = Proxy
```

</details>

<details>
<summary><code>master</code> error:</summary>

```
Error found:
in module [33mRowKindTest[0m
at tests/purs/failing/RowKindTest.purs:9:25 - 9:28 (line 9, column 25 - line 9, column 28)

  Could not match kind
  [33m      [0m
  [33m  Type[0m
  [33m      [0m
  with kind
  [33m     [0m
  [33m  Int[0m
  [33m     [0m

while matching label [33ma[0m
while checking that type [33mFoo[0m
  has kind [33m{ a :: Int [0m
           [33m, c :: Type[0m
           [33m, d :: Type[0m
           [33m}          [0m
while inferring the kind of [33mBar Foo[0m
while inferring the kind of [33mProxy (Bar Foo)[0m
in value declaration [33msomething[0m

```

</details>

<details>
<summary>This PR's output</summary>

```
Error found:
in module [33mRowKindTest[0m
at tests/purs/failing/RowKindTest.purs:9:25 - 9:28 (line 9, column 25 - line 9, column 28)

  Could not match kind
  [33m             [0m
  [33m  ( a :: Type[0m
  [33m  , b :: Type[0m
  [33m  )          [0m
  [33m             [0m
  with kind
  [33m             [0m
  [33m  ( a :: Int [0m
  [33m  , c :: Type[0m
  [33m  , d :: Type[0m
  [33m  )          [0m
  [33m             [0m

while checking that type [33mFoo[0m
  has kind [33m{ a :: Int [0m
           [33m, c :: Type[0m
           [33m, d :: Type[0m
           [33m}          [0m
while inferring the kind of [33mBar Foo[0m
while inferring the kind of [33mProxy (Bar Foo)[0m
in value declaration [33msomething[0m

```

</details>

I think this PR's output is better than `master`, even if this is very advanced.

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -409,17 +410,41 @@ checkTypeKind
   -> SourceType
   -> m ()
 checkTypeKind ty kind =
-  unifyKindsWithFailure (\_ _ -> throwError . errorMessage $ ExpectedType ty kind) kind E.kindType
+  -- In this case, if any failure occurs, we ignore it and throw our own error.
+  -- So, rather than use `unifyishRows` here and pay for its overhead,
+  -- we'll use the approach that was used before we migrated to `unifyishRows`
+  -- because the original approach fails faster.
+  unifyKindsWithFailure (UseSequential ty kind) kind E.kindType
```
JordanMartinez, line 417: While this should be faster due to the lack of overhead (unless there's GHC optimizations that will end up removing that), it also diverges from `unifyishRows`, which may cause issues in the future. Is that tradeoff worth it?

tests/purs/failing/RowOneMissingLabel.purs
JordanMartinez: I've pushed the kind unification-related changes. Let's continue that conversation in [this comment's thread](https://github.com/purescript/purescript/pull/4421#discussion_r1289219183). I'm still working on the `checkProperties` reimplementation. My initial work caused issues with 3701, so I'm trying a different approach. I likely won't get to this again until tomorrow.

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -103,25 +99,23 @@ subsumes' SElaborate (ConstrainedType _ con ty1) ty2 = do
...
-    return (defaultCoercion mode)
-  where
-    -- Find the first property that's in the first list (of tuples) but not in the second
-    firstMissingProp t1 t2 = fst <$> uncons (minusBy' (comparing rowListLabel) t1 t2)
+    (uncurry $ \(ts1', r1') (ts2', r2') ->
+      withBoolError $ unifyTypes (rowFromList (ts1', r1')) (rowFromList (ts2', r2')))
+    isTypesDoNotUnify
+    (\r1' r2' -> do
+      subst <- gets checkSubstitution
+      pure $ errorMessage $ TypesDoNotUnify (substituteType subst r1') (substituteType subst r2'))
```
rhendric, line 114: All of this is shared with the true `unifyTypes`, right? (Or at least, with access to `unifyTails` instead of the `withBoolError` hack, it would be.) How about exporting another variation from TypeChecker.Types for that case and subsumption?

I don't know what to call it, though. `unifyishTypes` is already hard to take seriously. :laughing:

tests/purs/failing/MissingRecordField.out
```diff
@@ -2,7 +2,20 @@ Error found:
 in module [33mMissingRecordField[0m
 at tests/purs/failing/MissingRecordField.purs:10:19 - 10:23 (line 10, column 19 - line 10, column 23)
 
-  Type of expression lacks required label [33mage[0m.
```
rhendric, line 5: This is something I mentioned earlier: the new error is less clear about which side is expected and which side is actual. This'll come up again in the `checkProperties` case.

I think we should add a boolean to `TypesDoNotUnify` to flag whether the types in question are known to be in actual-expected order, and give a different error message if true (`The type * does not match the expected type *`, perhaps?)

src/Language/PureScript/TypeChecker/Kinds.hs
```diff
@@ -409,17 +410,41 @@ checkTypeKind
...
-  goWithLabel l t1 t2 = withErrorMessageHint (ErrorInRowLabel l) $ go t1 t2
+  (unifyRows, onFailure) = case opts of
+    UseSequential ty kind ->
+      ( \r1 r2 -> do
+          let (matches, rest) = alignRowsWith (\_ -> go) r1 r2
+          sequence_ matches
+          void $ unifyTails rest
+
+      , \_ _ -> 
+          pure $ errorMessage $ ExpectedType ty kind
```
rhendric, line 441: This is a bit of a leaky abstraction, isn't it? `UseSequential` selects not only which strategy to use but also the type of the error message. Maybe both constructors should take an `onFailure`.

In the other direction, does the `UseUnifyishRows` constructor ever take an `isExpectedError` that isn't `isKindsDoNotUnify`? Maybe that should be pushed down into the implementation here.

src/Language/PureScript/TypeChecker/Kinds.hs
rhendric: I don't know about the performance impact or the potential issues caused by changing the implementation in subtle ways. I do however think that it's to our long-term benefit to maintain as few distinct code paths for doing similar things as possible—parallel implementations naturally diverge and then bugs that get fixed in one place remain in another. To the extent that it's practical, I'd love to see all of these very similar functions doing the same thing, only parametrized by little details like which error type to throw. I realize we will probably fall short of that ideal, but I think we should aim for it over hypothetical performance concerns (until performance impacts are actually measured).

tests/purs/failing/RowOneMissingLabel.purs
rhendric: I took a stab at `checkProperties` myself, as well as the other comments I made in my last review and a grab-bag of record-related error message improvements. I have mostly stayed away from the kind-unification piece aside from a refactoring that I hope you'll agree is a good idea; I think you've done more thinking about that side of things than I have.

src/Language/PureScript/TypeChecker/Kinds.hs
JordanMartinez: In that case, we should always use `unifyishRows`.

JordanMartinez: In latest commit, I've updated Kinds to always use `unifyishRows` and guarantee that `KindsDoNotUnify` is the error created in all circumstances.

src/Language/PureScript/Errors.hs
```diff
@@ -858,14 +854,21 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath fileCon
...
 
-        in paras [ line "Could not match type"
-                 , row1Box
-                 , line "with type"
-                 , row2Box
-                 ]
+        in if isOrdered then
+          paras [ line "The type"
+                , row1Box
+                , line "does not match the expected type"
```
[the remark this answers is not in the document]
JordanMartinez: Fixed in latest commit.

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -102,29 +98,12 @@ subsumes' SElaborate (ConstrainedType _ con ty1) ty2 = do
...
-    sequence_ common
-    -- Inject the info here
-    unifyTypes (rowFromList (ts1', r1')) (rowFromList (ts2', r2'))
-    -- Nothing was elaborated, return the default coercion
-    return (defaultCoercion mode)
-  where
-    -- Find the first property that's in the first list (of tuples) but not in the second
-    firstMissingProp t1 t2 = fst <$> uncons (minusBy' (comparing rowListLabel) t1 t2)
-subsumes' mode ty1 ty2@(TypeApp _ obj _) | obj == tyRecord =
-  subsumes' mode ty2 ty1
```
JordanMartinez, line 126: I'm guessing this case was removed so that the `isOrdered` part of `TypesDoNotUnify` is correct? Are there any other ramifications here?

tests/purs/failing/MultipleErrorsInOneRow.out
JordanMartinez, line 1: This output still reports two errors rather than one, like `NestedRecordTypeError.out`. Why? And is that desirable?

JordanMartinez: > Why?

This case occurs specifically because of the pattern matching, so it's not the same as `NestedRecordTypeError`. I think it would be better to add 'pattern match' or 'exhaustivity check' somewhere in the file's name for clarity on what is being tested. I'm not entirely sure what happens between `checkExhaustiveExpr` to `unifyTypes'`.

In my local exploration, adding a special case for this in `unifyTypes` does fix the error, but modifies other test outputs unrelated to this as well. Some harmless changes are changing `(` and `)` to `{` and `}`. But there was at least one test where the entire record is printed rather than just the diff. Fixing that may require adding another special case elsewhere in our type printing, but I'm not familiar with that code.

> And is that desirable?

In this case, I don't think it is. The two errors printed here are because a label expecting a `String` pattern match got an `Int` instead. Keeping these separate helps one troubleshoot the various issues here.

src/Language/PureScript/TypeChecker/Subsumption.hs
rhendric: Yeah, that's right. It was a remarkably useless clause anyway, assuming I didn't miss anything; going through the other patterns matched previously, I think it's impossible for `ty1` here to be any type that would trigger a previous clause with a `TypeApp` in the first position. In which case, this clause was just an overly complicated way to fall through to the final clause with the types in the wrong order.

tests/purs/failing/MultipleErrorsInOneRow.out
rhendric: My concept was that if you have a record (pattern or expression) that uses all the right labels but has some of the wrong types, it's okay—preferable, even—to report those as separate errors focused on the specific labels that are incorrect. It's only if the shape of the entire record is incorrect that the entire record needs to be reported as an error, and *if* you're already reporting the entire record as an error, you might as well suppress the individual label type errors because they'll be appearing in the entire-record error.

src/Language/PureScript/TypeChecker/Subsumption.hs
```diff
@@ -98,27 +98,12 @@ subsumes' SElaborate (ConstrainedType _ con ty1) ty2 = do
...
-      withBoolError $ unifyTypes (rowFromList (ts1', r1')) (rowFromList (ts2', r2')))
-    isTypesDoNotUnify
-    (\r1' r2' -> do
-      subst <- gets checkSubstitution
-      pure $ errorMessage $ TypesDoNotUnify (substituteType subst r1') (substituteType subst r2'))
-    r1
-    r2
+subsumes' mode (TypeApp s1 f1@(TypeConstructor _ C.Record) r1) (TypeApp s2 f2@(TypeConstructor _ C.Record) r2) = do
+  subst <- gets checkSubstitution
+  unifyishRowTypes True (TypeApp s1 f1) (TypeApp s2 f2) (subsumes' SNoElaborate) (substituteType subst r1) (substituteType subst r2)
```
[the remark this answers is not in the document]
rhendric: I don't think those comments are helpful anymore; I read them as helping the reader follow the flow of what happens in what order, and now all that flow is abstracted away into `unifyish*`.

tests/purs/failing/MultipleErrorsInOneRow.out
JordanMartinez: Ok. Then the changelog entry I wrote will need to be updated to clarify that.

src/Language/PureScript/Errors.hs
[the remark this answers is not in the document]
rhendric: There's a good case for using the word ‘actual’ here—GHC does it, I used it when talking about the change informally, it pairs well with ‘expected’—but I feel a little odd about using it in the error message. When you try to pass an `Int` into a function of type `String -> Boolean`, both `Int` and `String` are ‘actual’ types—neither type is more real than the other.

I guess you could make a similar argument that the compiler can ‘expect’ the user to provide a term of type `String` to that function, or it could ‘expect’ the user to provide a function term of type `Int -> a`, but my inclination is to apply a little anthropomorphism and say that the *function* is doing the ‘expecting’ and expects a `String` and not an `Int`.

When checking an annotated type `(3 :: String)`, expected/actual makes more sense. But maybe there's a term that could replace ‘actual’ for both cases?

src/Language/PureScript/TypeChecker/Unify.hs
```diff
@@ -120,66 +132,85 @@ unifyTypes t1 t2 = do
...
+  -> (SourceType -> SourceType -> m ()) -- ^ function to use to check types at common labels
+  -> SourceType
+  -> SourceType
+  -> m ()
+unifyishRowTypes isOrdered errorTypeWrapper1 errorTypeWrapper2 =
+  unifyishRows (uncurry unifyTails) isTypesDoNotUnify $ \r1 r2 -> do
+    subst <- gets checkSubstitution
+    pure $ errorMessage $ TypesDoNotUnify isOrdered
+      (errorTypeWrapper1 $ substituteType subst r1)
+      (errorTypeWrapper2 $ substituteType subst r2)
```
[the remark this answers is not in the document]
rhendric: I'd rather not do that unless it's needed; substitution is expensive (it comes up in my profiling a lot) and limiting the size of types that get substituted is a good thing to do habitually. In the code that has been written so far, the difference is tiny—the wrapper functions are either `id` or a single `TypeApp`—but I'd rather future-proof against silent performance sucks than visible bugs.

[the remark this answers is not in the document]
JordanMartinez: Makes sense.

src/Language/PureScript/Errors.hs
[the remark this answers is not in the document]
JordanMartinez: Other renderings I've seen are
> Expected X but got Y

So, the error message could be updated to say

```
Expected type

   X
   
but got type

   Y
```

JordanMartinez:
Hm.... I would have expected the recent changelog entry edit to not trigger CI. Looks like #4502 didn't work...?

tests/purs/failing/MultipleErrorsInOneRow.out
JordanMartinez: I've pushed an edit to the changelog entry that hopefully clarifies this. Let me know if you think it's too descriptive and should appear in the documentation repo instead.

rhendric:
Yeah, I would not have expected #4502 to do what you wanted but you guys merged it before I took a look and then I figured it wouldn't be worth a fuss. I don't think `push` would be an acceptable alternative because it wouldn't trigger on pushes to other people's forks, which is where most PRs come from. We could instead put this logic in the workflow itself—check if the last successful run differs only from the current commit in irrelevant files and skip all actions if so.

CHANGELOG.d/fix_better-row-error-messages.md
```diff
@@ -0,0 +1,165 @@
...
+    ( h :: Int
+    , j :: Int
+    ...
+    )
+  with type
+    ( g :: t1
+    ...
+    | t8
+    )
+  ```
```
rhendric, line 107:
``````suggestion
  **Case 2: reporting all label differences (and only those differences) together**

  Before:
  ```
  Type of expression contains additional label extraLabel1
  ```
  (without any messages about more extra labels)

  or
  ```
  Type of expression lacks required label requiredLabel1
  ```
  (without any messages about more required labels)

  or
  ```
  Could not match type
    ( a :: Int
    , b :: Int
    , c :: Int
    , d :: Int
    , e :: Int
    , f :: Int
    , extraLabel1 :: Int
    , extraLabel2 :: Int
    ...
    )
  with type
    ( a :: t7
    , b :: t6
    , c :: t5
    , d :: t4
    , e :: t3
    , f :: t2
    , requiredLabel1 :: t1
    ...
    | t8
    )
  ```
  After:
  ```
  The actual type
    { extraLabel1 :: Int
    , extraLabel2 :: Int
    ...
    }
  does not match the expected type
    { requiredLabel1 :: t1
    ...
    | t2
    }
  ```
``````

rhendric:
I think this could use some eyeballs belonging to someone who regularly bumps up against errors involving large rows, but pending their approval I'm happy with where we've landed. I don't think it's perfect but I think it's better in every way on the cases we've touched.

JordanMartinez:
Given that no one has responded in the past few days and that this was a highly-desired PR on the survey, there's three ways we could move forward:
1. we continue waiting until someone has a good example on which to try this PR, hope they build this PR locally, and then tell us of the results. Building this PR locally is not exactly convenient and that inconvenience may lead them to just spending their time trying to resolve the problem rather than building this PR and using that version of the compiler on the problem.
1. merge this PR, making a pre-release. The pre-release removes the need to build the PR locally. so we're more likely to get feedback that way. However, being 1.5 weeks out from the next scheduled release, we may not have enough time for someone to come across a problem where they get to see how this version handles things.
1. We make an official release now for `0.5.11`, then merge this PR, making a prelease `0.15.12-0`, which gives this work a 6-week time period to garner feedback.

> I don't think it's perfect

What do you consider to still be imperfect about it?

rhendric:
Probably not everyone who indicated interest in this PR on the survey is subscribed to it. We could maybe stir up some interest on Discourse. I think it's important for requests like this one to have at least one ‘customer’ verifying that the change meets the objective before it's merged; I apply this policy to things like ‘better error messages’ and ‘make go faster’, as opposed to ‘here is a bug where the documentation or common sense says one thing should happen but another thing happens instead’. If this is such an in-demand feature, finding one single user who has this problem and can build the compiler shouldn't be hard. And conversely, we don't need to rush it out if we can't.

> What do you consider to still be imperfect about it?

I think there are some inconsistencies in what errors we present as rows versus records, and I think there are many more places where the expected/found distinction could be drawn but where the current code shrugs (in fact, I somewhat suspect that, given that we highlight a particular subexpression, we can *always* draw an expected/found distinction *from the perspective of the highlighted subexpression*, and the greatest barrier to doing so is that there are still places where highlighting is off because source spans aren't being applied in one or two spots). I'm pretty comfortable with all that being worked out later as needed, though.

JordanMartinez:
> We could maybe stir up some interest on Discourse.

Would you mind typing up such a Discourse post then?

> I think there are some inconsistencies in what errors we present as rows versus records, and I think there are many more places where the expected/found distinction could be drawn but where the current code shrugs (in fact, I somewhat suspect that, given that we highlight a particular subexpression, we can always draw an expected/found distinction from the perspective of the highlighted subexpression, and the greatest barrier to doing so is that there are still places where highlighting is off because source spans aren't being applied in one or two spots). I'm pretty comfortable with all that being worked out later as needed, though.

Makes sense. Thanks for clarifying.

f-f (approved):
I had a look at the patch and it seems good!

I am approving this so that we can get it merged and test it more widely through the prerelease - we could keep this waiting and ask people to try it out, but I think that's asking for unrealistic conditions: I do certainly see terrible row mismatch errors, but not at predictable times, and when that happens I'm usually in the middle of something complex so I want to just go past it rather than compile a custom build of the compiler.

While it might not be perfect, the improvements to error messages seem good enough to warrant getting this mainlined and iterating on it.

rhendric:
Based on the current state of the Discourse thread, I actually think this PR misses the mark wide. The test cases we were focused on all along are not relevant to what seems to be the primary way these error messages come up in the wild—namely, through `Union` constraints instead of through unification with rows having unknowns in their tails.

Please let's have a little more time to consider whether anything in this PR is relevant to handling that use case before we run ahead and merge it.

JordanMartinez:
> The test cases we were focused on all along are not relevant to what seems to be the primary way these error messages come up in the wild—namely, through Union constraints instead of through unification with rows having unknowns in their tails.

I somewhat disagree with this interpretation of people's responses so far. While `Union` constraints are another aspect of this problem we have missed, I'm not sure I would consider them to be the primary way such errors occur. (We've only had 2 people respond so far, and one of them said "normal" errors were better). So, I still think this PR does address some issues, just not all of them, and only two responses is not a clear indicator of that.

rhendric:
> I still think this PR does address some issues, just not all of them

I guess it depends on your definition of ‘issue’. I would take the perspective that an issue is something that a user encounters that makes their day worse. I think this PR makes some changes, I think it makes some of our error messages better, but I have yet to see any evidence that it resolves any issues. All we have is evidence that when users poke the thing we changed, it responds with the error messages we designed for that type of poking. We could have the best error messages imaginable for some scenario that users never naturally encounter in the wild, and it wouldn't improve things at all.

In exchange for that no-benefit, the PR adds several confusingly-named functions (we did our best, but we should also acknowledge that where we ended up isn't great) to the project and possibly impacts the performance of the type checker. It removes three constructors from our `SimpleErrorMessage` type and adds a parameter to another one—kind of a draw, there, in terms of long-term impact, but a net negative if we end up wanting those constructors back soon.

Most PRs are written by people who are invested in the particular issue that they resolve. This PR, at this point, is mostly written by you and me; and speaking for myself, I have never had problems with the error messages I get related to rows. But I also don't use `react-basic` for anything. For most PRs, having an author who knows the problem space and believes that the solution improves things is enough to bias me towards making changes on the scale we're making them here. But we don't even have that—we have authors who (again speaking for myself) are interested in the abstract in improving our error messages, but who don't have experience with the specifics of this type of error message—so from my perspective we run quite a high risk of churning the code for no end benefit to anyone. I object to the let's-just-merge-and-see approach if we don't even have a single stakeholder telling us that it resolves one of their issues.

The `Union` thing does genuinely give me pause. I haven't sat down to take the time to think it through but I would like to, and I think there's a decent chance that dealing with it will mean yet a different approach than the one we took here; it might not be just another function that we need to bring into the `unifyishRows` family. Also, while yes we've only had two respondents, from their comments I think that `Union` is the main course here, not a side dish.

drathier:
Huge reply from the author on discourse https://discourse.purescript.org/t/upcoming-changes-to-error-messages-for-large-records-rows/3696/19 says he's available on discord but rarely checks github issues or discourse. Lots of input on this problem too, which I definitely second.

JordanMartinez:
> I guess it depends on your definition of ‘issue’....

@rhendric I would quote the rest of what you said above, but that's a lot of text :laughing:.

Fair points! From my perspective, this does make things better, but I don't think I can provide evidence proving that, at least not enough to convince you. I'm also finding it difficult to come up with a logical response to your arguments above. However, part of the difficulty of this issue is coming across reproducible errors that can be collected into a publicly-visible place that would inform us on how to best improve errors. There's a lot of upfront cost to do that.

Given the responses on the Discourse thread, what do you propose is the way to move forward? Given both my initial response and @f-f's response to merge this, it may be worthwhile to close this PR and refocus on the original problem. So long as the branch isn't deleted, none of the work done so far is lost.

rhendric:
> From my perspective, this does make things better, but I don't think I can provide evidence proving that, at least not enough to convince you. I'm also finding it difficult to come up with a logical response to your arguments above.

It matters to me what hat you're wearing when you say this makes things better. If it's your user hat—if you encounter the previous iteration of these messages in your own PureScript work and would prefer the new ones—that's all the convincing I need. If it's your maintainer hat—if you think the code is better with these changes than without them, independent of what users experience—then I disagree but there's no reason for my opinion to be weighed above yours; let's keep talking and see if we can agree or get a third maintainer's opinion. If it's your business analyst hat—if you have reason to believe users would want to see these changes merged—that's the tougher hill to climb, unless you can produce the users who want this. Otherwise it's your imaginary users against my imaginary users and that's hard to discuss logically, as you say.

> what do you propose is the way to move forward?

From a process-management perspective, we could close; maybe it'd be better to mark as draft, since I don't think we're giving up on the concept, but I don't feel strongly either way.

From a getting-things-done perspective, whichever one of us gets cycles first should make a test case that mimics the `react-basic` API and start sketching out ways to make that error more digestible. It might be special-case logic for handling `Union` constraints; it might involve increasing the tracking already started in this PR of which types are expected  and which are found; it might be a general-purpose change to the row-diffing code if only one side of the diff is above a size threshold; it might be something I haven't brainstormed yet. I think I at least have a pretty good idea of the conditions that cause the huge-row errors that `react-basic` users are encountering in the wild. I have no idea about the conditions that produce FredTheDino's errors but if they can't share details and won't participate in implementation then there's only so much we can expect to do about that. So I would focus on `react-basic` and making those errors look like things I'd want to see in a terminal, somehow, details TBD because that's the work. And if that approach ends up intersecting with what we already did here, great, that's however many person-days you and I put into this already not wasted after all.

PR #4435, opened 2023-02-01
Depend on `hspec-2.10.9`

sol:
This addresses https://github.com/hspec/hspec/issues/685#issuecomment-1411868905

### DIFF
--- a/purescript.cabal
+++ b/purescript.cabal
@@ -438,7 +438,7 @@ test-suite tests
   build-depends:
     purescript,
     generic-random >=1.5.0.1 && <1.6,
-    hspec ==2.9.2,
+    hspec >= 2.10.7 && < 3,
     HUnit >=1.6.2.0 && <1.7,
     newtype >=0.2.2.0 && <0.3,
     QuickCheck >=2.14.2 && <2.15,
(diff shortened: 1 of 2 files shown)


purescript.cabal
```diff
@@ -438,7 +438,7 @@ test-suite tests
   build-depends:
     purescript,
     generic-random >=1.5.0.1 && <1.6,
-    hspec ==2.9.2,
+    hspec >= 2.10.7 && < 3,
```
sol, line 441: `Test.Hspec` is committed to be stable.  Unless you rely on internal particulars of `hspec-core`, any recent version of Hspec will work.

andys8:
I don't fully understand this being a fix for the `terminfo` issue. Judging from the comment I assumed 2.9.3 introduced a dependency and we're therefore stuck with 2.9.2. Is it now just fine to update? Or is this not the case anymore in 2.10.9?

https://github.com/purescript/purescript/blob/3cf73939a4869090fb108d2e14de852ae513568b/stack.yaml#L23-L29

sol:
> Or is this not the case anymore in 2.10.9?

Yes, it's not. I intended to reflect that in the updated comment.

Here is the related changelog entry: https://github.com/hspec/hspec/blob/main/CHANGES.markdown#changes-in-2107-2022-12-03

andys8:
Ah, I didn't catch the changed comment. Cool, thanks.

rhendric (approved):

rhendric:
Build failure is connected to https://github.com/purescript/registry/issues/170; this PR is blameless.

f-f (approved):

Issue #3243, opened 2018-02-08
Symbol in typeclass bug?


vladciobanu:
Not sure how to express it in words, but:

```purescript
class WithoutSym t where
  wrel ∷ ∀ sym r2
       . IsSymbol sym
       ⇒ RowCons sym t () r2
       ⇒ RowLacks sym ()
       ⇒ SProxy sym
       → t
       → { | r2 }

-- | This works fine
instance wsInt ∷ WithoutSym Int where
  wrel sym p = insert sym p {}


class IsSymbol sym <= Rel t sym where
  rel ∷ ∀ r2
       . RowCons sym t () r2
       ⇒ RowLacks sym ()
       ⇒ SProxy sym
       → t
       → { | r2 }

instance relInt ∷ Rel Int "test" where
  rel sym p = insert sym p {}
-- |          ^
-- | Could not match type
-- |   ( test :: Int
-- |   )
-- | with type
-- |   r24
-- | Prim.RowCons "test"
-- |              Int
-- |              ()
-- |              r24

```

Is this a compiler bug or is it expected? You can try it out: http://try.purescript.org/?backend=core&gist=855016e5488c0cafeedd98ff82955518

vladciobanu:
A different example that might shed some light (I forget who it was who posted it in Slack, sorry):

```purescript
-- this works just fine
fnWithSym
  :: forall sym r2
   . IsSymbol sym
  => RowCons sym t () r2
  => RowLacks sym ()
  => SProxy sym
  -> t
  -> { | r2 }
fnWithSym sym p = insert sym p {}


-- this fails to compile with the same cannot match error as the typeclass example
fnWithoutSym
  :: forall r1 r2
   . RowLacks "field" r1
  => RowCons "field" Boolean r1 r2
  => Record r1
  -> Record r2
fnWithoutSym r = insert (SProxy :: SProxy "field") true r
```

paluh:
@vladciobanu it was me.
It is real detail but I think that it is even better to use `r1` and `r2` in both cases and make "the source" of field label the only difference:

  ``` purescript
  -- this works
  addField'
    :: forall r1 r2 l
    . IsSymbol l
    => RowLacks l r1
    => RowCons l Boolean r1 r2
    => SProxy l
    -> Record r1
    -> Record r2
    addField' l r = insert l true r

  -- this fails
  addField
    :: forall r1 r2
    . RowLacks "field" r1
    => RowCons "field" Boolean r1 r2
    => Record r1
[...]

PR #4436, opened 2023-02-16
RE: Visible Type Applications

purefunctor:
**Description of the change**

An alternative implementation of visible type applications that simplifies the type-checking rules and extraneous features. This PR also depends on #4376.

## Rationale

TBW... (while I try to fix merge conflicts) 

---

**Checklist:**


artemisSystem:
> however, variables in class heads and data declarations are available by default.

> Potential future feature: the ability to change type variable visibility between type class members.

Just pointing out that this would be a breaking change, as far as i understand? Not sure if that is a big deal or not, but something to keep in mind.

CHANGELOG.d/fix_polymorhic_constructors.md
```diff
@@ -0,0 +1,40 @@
+* Defer monomorphization for data constructors
```
i-am-the-slime, line 1: This file name should probably be polymorphic (it's missing the second `p`)

purefunctor:
> > however, variables in class heads and data declarations are available by default.
> 
> > Potential future feature: the ability to change type variable visibility between type class members.
> 
> Just pointing out that this would be a breaking change, as far as i understand? Not sure if that is a big deal or not, but something to keep in mind.

@artemisSystem It depends. The breaking change here would come from the new syntactic forms which determines if a type is available for type application or not. I think that having them on by default for class heads and data declarations is reasonable to allow for usage w/o introducing more friction.

I can also imagine the latter being another additive feature, but I'm just not sure how to achieve that syntactically yet. Maybe, something along the lines of:
```purescript
class Visibility a where
  visibility :: Invisible a => a -> a
  -- or `visibility :: invisible a. a -> a`?
```
where `Invisible` is basically a sentinel just for the compiler to use. If the need to have selective visibility for constructors, how should we go about that?

CHANGELOG.d/fix_polymorhic_constructors.md
purefunctor: Hmm, I'm not sure if it'd matter for the changelog.md builder? 😄

MonoidMusician: yeah the filename doesn’t matter :)

MonoidMusician:
I think the errors you added should be unified to a "monomorphic/not-visibly-polymorphic type cannot have a type application". It would be great if those had separate error text based on whether it is polymorphic or not: “cannot type-apply the monomorphic type `Int -> Int`” and “cannot apply (polymorphic) type without visible quantification: `forall a. a -> a`”.

I think the current behavior of `@_` may be confusing if the expression still being polymorphic is visible to the user in some cases (ahem, record fields?). What happens if we don't special case it? What happens to `_` nested in other type expressions? Will have to think about it and test it.

Will take a look at more of the logic later.

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -676,6 +676,10 @@ data Expr
   --
   | App Expr Expr
   -- |
+  -- A type application
```
i-am-the-slime, line 679:
```suggestion
  -- A type application (e. g.`@Int`)
```

src/Language/PureScript/CST/Flatten.hs
```diff
@@ -151,6 +151,7 @@ flattenExpr = \case
   ExprRecordAccessor _ a -> flattenRecordAccessor a
   ExprRecordUpdate _ a b -> flattenExpr a <> flattenWrapped (flattenSeparated flattenRecordUpdate) b
   ExprApp _ a b -> flattenExpr a <> flattenExpr b
+  ExprVisibleTypeApp  _ a _ b -> flattenExpr a <> flattenType b
```
JordanMartinez, line 154: It's been a while since I've looked at this code, so sorry if this is incorrect. But shouldn't the `@` token be included here?

src/Language/PureScript/Types.hs
```diff
@@ -77,7 +90,7 @@ data Type a
   -- | Explicit kind application
   | KindApp a (Type a) (Type a)
   -- | Forall quantifier
-  | ForAll a Text (Maybe (Type a)) (Type a) (Maybe SkolemScope)
+  | ForAll a Text (Maybe (Type a)) (Type a) (Maybe SkolemScope) TypeVarVisibility
```
JordanMartinez, line 93: IIRC, the values stored in these constructors are usually stored in the same order of appearance. Nit, but shouldn't the `TypeVarVisbility` arg appear before the `Text` arg?

src/Language/PureScript/CST/Flatten.hs
purefunctor: Yup

src/Language/PureScript/Types.hs
purefunctor: I think that's ideal, yep

purefunctor: On a second look, I think it's more apt to have it at the end as the Aeson instances for `Type` makes use of tuples for the ForAll. I'm not sure if this affects anything though

tests/purs/failing/VisibleTypeApplications1.out
```diff
@@ -0,0 +1,20 @@
...
+
+  An expression of polymorphic type
+  with the invisible type variable [33ma[0m:
+  [33m                  [0m
+  [33m  forall a. a -> a[0m
+  [33m                  [0m
+  cannot be applied to:
+  [33m     [0m
+  [33m  Int[0m
+  [33m     [0m
```
purefunctor, line 13: This error can be expanded further by suggesting that the user can change the signature of the inferred type to allow it to be used in type application.

I think that for polymorphic types, it's the case that they're often written out by the user such that we can suggest editing the source, but I feel we can clarify this in the error index for the documentation page.

src/Language/PureScript/Types.hs
JordanMartinez: Ah... like backwards compatibility on Pursuit might be affected?

purefunctor: Yeah, to be exact, this change in particular:
https://github.com/purescript/purescript/blob/a227bdd5a71616dffe09c83c26cfa6c097efa7de/src/Language/PureScript/Types.hs#L369-L381

I do think that this is the correct behaviour though—it retains backward compatibility by parsing the contents in a way that the most "general" case is at the end. Tests also pass with this implementation.

JordanMartinez: Why do we need to keep using Tuples for this particular encoding? Couldn't we define the representation as an object and try decoding that representation first? And if it fails, we fallback to the original tuple encoding? I believe that would still maintain backwards compatibility, but it would also allow `Forall` to be extended in the future without running into this order-dependent issue again.
```haskell
let 
  objMbKind = A.withObject "Forall" $ \o -> do
   visibility <- o .: "visibility"
   identifier <- o .: "identifier"
   k <- o .:? "kind"
   ty <- o .: "type"
   skolemScope <- o .: "skolemScope"
   pure $ Forall a visibility identifier k ty skolemScope
   
   withoutMbKind = do
     (b, c, d) <- contents
     ForAll a TypeVarInvisible b Nothing <$> go c <*> pure d
  
   withMbKind = do   
     (b, c, d, e) <- contents  
     ForAll a TypeVarInvisible b <$> (Just <$> go c) <*> go d <*> pure e
  
objWithMbKind <|> withMbKind <|> withoutMbKind

Comment (purefunctor) [reply src/Language/PureScript/Types.hs]: That's a good idea. I think it also makes sense to remove this backwards compatibility in the (far) future.

File: src/Language/PureScript/CST/Convert.hs [hunk 6]
```diff
@@ -136,11 +136,13 @@ convertType fileName = go
...
+        mkForAll a b v t = do
           let ann' = widenLeft (tokAnn $ nameTok a) $ T.getAnnForType t
-          T.ForAll ann' (getIdent $ nameValue a) b t Nothing
-        k (TypeVarKinded (Wrapped _ (Labeled a _ b) _)) = mkForAll a (Just (go b))
-        k (TypeVarName a) = mkForAll a Nothing
+          T.ForAll ann' v (getIdent $ nameValue a) b t Nothing
+        k (TypeVarKinded (Wrapped _ (Labeled (Just _, a) _ b) _)) = mkForAll a (Just (go b)) T.TypeVarVisible
+        k (TypeVarKinded (Wrapped _ (Labeled (_, a) _ b) _)) = mkForAll a (Just (go b)) T.TypeVarInvisible
+        k (TypeVarName (Just _, a)) = mkForAll a Nothing T.TypeVarVisible
+        k (TypeVarName (_, a)) = mkForAll a Nothing T.TypeVarInvisible
```
JordanMartinez, line 145: Seems to me that ‘TypeVarKinded’ cases can be combined here and you use ‘maybe’ visibility part. Same for ‘TypeVarName’

JordanMartinez (approved):

JordanMartinez (approved):

JordanMartinez:
My approval above means it's my opinion that it's better to merge this in its current state now than not to merge it. If it does implement the feature while not handling edge cases or other situations properly, it's not like we don't have other useful language features that have their own years-long-standing edge-case bugs. To me, the net gain of a language feature that mostly works, especially one as highly desired as this one, is better than the possible downsides. Moreover, by getting it merged, it provides feedback on what edge cases this PR may still not yet resolve or other design situations not yet considered.

But, because I know that others are more qualified when it comes to reviewing this part of the codebase, I'm choosing not to merge this at this time despite all our usual constraints being satisfied otherwise. On one hand, I may just be struggling with [politician's syllogism](https://en.wikipedia.org/wiki/Politician%27s_syllogism), and maybe the real issue to further discuss are design questions not yet answered in the linked issue (of which I can't recall whether there were such issues, nor what those are). On the other hand, I do think that, if no one objects to merging this, that I will merge this sometime in the future.

deemp:
It's so cool! 

Following https://github.com/purescript/purescript/issues/3137#issuecomment-761598429, I used a shorter synonym for `prop` from `purescript-profunctor-lenses`.

```purs
p :: ∀ @l r1 r2 r a b. IsSymbol l ⇒ Cons l a r r1 ⇒ Cons l b r r2 ⇒ Lens (Record r1) (Record r2) a b
p = prop (Proxy @l)

type St = {someField :: Int }

st :: St
st = { someField : 1}

ip :: Int
ip = view (p @"someField") st
```

paluh:
Actually I tried to shorten the function and drop the proxy parameter which essentially works:

```purescript
prop'
  :: forall @l r r1 r2 a b
   . IsSymbol l
  => Row.Cons l a r r1
  => Row.Cons l b r r2
  => Lens { | r1 } { | r2 } a b
prop' = do
  let p = Proxy @l
  lens (Record.get p) (flip (Record.set p))

_test
  :: forall r36 a37 b38 p
   . Strong p
  => p a37 b38
  -> p { test :: a37 | r36 } { test :: b38 | r36 }
_test = prop' @"test"
```

The problem is that by default compiler derived a type for `_test` which wasn't correct and caused compile error (so I copied the signature above from the original `Lens.Record.prop $ Proxy @"test"`):

```purescript
_test
  :: forall r r1 r2 a b
   . IsSymbol "test"
  => Row.Cons "test" a r r1
  => Row.Cons "test" b r r2
  => (forall p. Strong p => p a b -> p (Record r1) (Record r2))
```

I think that this unexpected behavior. Should I fill in a bug report for that?

MonoidMusician:
The error you see with that inferred type is already filed as #3243

PR #4437, opened 2023-02-17
Account for typed holes when checking declarations

### LINKED ISSUE #4408: Module-level values should always be suggested in type holes
## Description

Module-level values are not currently suggested in type holes.

## To Reproduce

```hs
module Main where

newtype Wat = Wat Int

topWat :: Int -> Wat
topWat = Wat

wat :: Wat
wat = 
  let 
    inWat = Wat 
  in 
    ?help 1
```

## Expected behavior

`topWat` should also be suggested, instead of just `inWat` and `Wat` as seen here:

```hs
[1/1 HoleInferredType] src/Main.purs:13:5

  13      ?help 1
          ^^^^^
  
  Hole 'help' has the inferred type
  
    Int -> Wat
  
    inWat     :: Int -> Wat
    Main.Wat  :: Int -> Wat
  
  in the following context:
  
    inWat :: Int -> Wat
  
  in value declaration wat
```

## Additional context

On another module, `Lib`, importing `Main` yields the expected error:

```hs
module Lib where

import Main

libWat :: Wat
libWat = ?help 10
```

```hs
[1/1 HoleInferredType] src/Lib.purs:6:10

  6  libWat = ?help 10
              ^^^^^
  
  Hole 'help' has the inferred type
  
    Int -> Wat
  
    Main.topWat  :: Int -> Wat
    Main.Wat     :: Int -> Wat
  
  in value declaration libWat
```

I thought this was because of the recent changes with top-level identifiers but I checked as far back as 0.14.7 and this seems to have been the case for a while now.

## PureScript version

0.15.6

purefunctor:
**Description of the change**

Closes #4408. The compiler now accounts for typed holes when checking value declarations, allowing for more values to be suggested rather than being limited by reverse lexicographical ordering. It does this by attaching more data to the graph that's built internally for reverse topological sorting of value declarations that need to be checked.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Sugar/BindingGroups.hs
+++ b/src/Language/PureScript/Sugar/BindingGroups.hs
@@ -9,7 +9,7 @@ module Language.PureScript.Sugar.BindingGroups
   ) where
 
 import Prelude
-import Protolude (ordNub)
+import Protolude (ordNub, swap)
 
 import Control.Monad ((<=<), guard)
 import Control.Monad.Error.Class (MonadError(..))
@@ -29,6 +29,7 @@ import Language.PureScript.Environment
 import Language.PureScript.Errors hiding (nonEmpty)
 import Language.PureScript.Names
 import Language.PureScript.Types
+import qualified Data.Map as M
 
 data VertexType
   = VertexDefinition
@@ -103,9 +104,17 @@ createBindingGroups moduleName = mapM f <=< handleDecls
           in (d, (name, vty), self ++ deps)
         dataVerts = fmap mkVert allDecls
     dataBindingGroupDecls <- parU (stronglyConnCompR dataVerts) toDataBindingGroup
-    let allIdents = fmap valdeclIdent values
-        valueVerts = fmap (\d -> (d, valdeclIdent d, usedIdents moduleName d `intersect` allIdents)) values
-    bindingGroupDecls <- parU (stronglyConnComp valueVerts) (toBindingGroup moduleName)
+    let 
+      makeValueDeclarationKey = (,) <$> exprHasNoTypeHole . valdeclExpression <*> valdeclIdent
+      valueDeclarationKeys = makeValueDeclarationKey <$> values
+
+      valueDeclarationInfo = M.fromList $ swap <$> valueDeclarationKeys
+      computeValueDependencies = (`intersect` valueDeclarationKeys) . usedIdents valueDeclarationInfo moduleName 
+  
+      makeValueDeclarationVert = (,,) <$> id <*> makeValueDeclarationKey <*> computeValueDependencies
+      valueDeclarationVerts = makeValueDeclarationVert <$> values
+
+    bindingGroupDecls <- parU (stronglyConnComp valueDeclarationVerts) (toBindingGroup moduleName)
     return $ filter isImportDecl ds ++
              dataBindingGroupDecls ++
              filter isTypeClassInstanceDecl ds ++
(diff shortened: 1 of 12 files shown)


JordanMartinez (approved):

src/Language/PureScript/Sugar/BindingGroups.hs
```diff
@@ -103,9 +104,17 @@ createBindingGroups moduleName = mapM f <=< handleDecls
...
+      makeValueDeclarationKey = (,) <$> exprHasNoTypeHole . valdeclExpression <*> valdeclIdent
+      valueDeclarationKeys = makeValueDeclarationKey <$> values
+
+      valueDeclarationInfo = M.fromList $ swap <$> valueDeclarationKeys
+      computeValueDependencies = (`intersect` valueDeclarationKeys) . usedIdents valueDeclarationInfo moduleName 
+  
+      makeValueDeclarationVert = (,,) <$> id <*> makeValueDeclarationKey <*> computeValueDependencies
+      valueDeclarationVerts = makeValueDeclarationVert <$> values
+
+    bindingGroupDecls <- parU (stronglyConnComp valueDeclarationVerts) (toBindingGroup moduleName)
```
rhendric, line 117: Am I correct that the big change here is that you've changed the keys of the dependency graph from `Ident` to `(Bool, Ident)` and you're relying on the implicit lexicographic sorting of `stronglyConnComp` to put hole-less declarations first?

That could be a little more clear, if so. I don't see anything in the documentation for `Data.Graph` that directly claims this as an invariant.

src/Language/PureScript/Sugar/BindingGroups.hs
```diff
@@ -137,18 +159,21 @@ flattenBindingGroups = concatMap go
       ValueDecl sa ident nameKind [] [MkUnguarded val]) ds
   go other = [other]
 
-usedIdents :: ModuleName -> ValueDeclarationData Expr -> [Ident]
-usedIdents moduleName = ordNub . usedIdents' S.empty . valdeclExpression
+usedIdents :: M.Map Ident Bool -> ModuleName -> ValueDeclarationData Expr -> [(Bool, Ident)]
+usedIdents identInfo moduleName = ordNub . usedIdents' S.empty . valdeclExpression
```
rhendric, line 163: This change strikes me as an awkward mixing of responsibilities; any reason not to map over the result up in `handleDecls` and leave this unchanged?

src/Language/PureScript/Sugar/BindingGroups.hs
```diff
@@ -29,6 +29,7 @@ import Language.PureScript.Environment
 import Language.PureScript.Errors hiding (nonEmpty)
 import Language.PureScript.Names
 import Language.PureScript.Types
+import qualified Data.Map as M
```
rhendric, line 32: Place this with the other `Data.*` imports please.

rhendric (changes requested):

src/Language/PureScript/Sugar/BindingGroups.hs
purefunctor: That's correct. I could also probably use a custom data type for this if we ever want to do reordering beyond hole or no-hole declarations.

After having asked, it seems like this is more of an implementation detail rather than an invariant. That being said, implementing a topological sort that maintains this invariant is definitely possible.

src/Language/PureScript/Sugar/BindingGroups.hs
purefunctor: Good catch! Reverting this...

JordanMartinez:
Correct me if I'm wrong, but the latest changes addressed Ryan's feedback, right?

rhendric:
I'd just like a comment that explains how we're relying on an undocumented implementation detail of `stronglyConnComp`, just in case that implementation changes and someone needs to figure out how this code was expected to work.

JordanMartinez:
A comment has been added. If there is no other feedback, perhaps we merge this and make a new release? It's been almost 3 months since the last release, much longer than our 6 week cycle.

rhendric (approved):

PR #4438, opened 2023-02-22
Dark mode support for Pursuit documentation

sometimes-i-send-pull-requests:
**Description of the change**

This adds dark mode support to Pursuit documentation.  Fixes purescript/pursuit#462.

The implementation is fairly ordinary.  Unlike #4093 (which I didn't see until I had already finished this), nothing fancy like CSS variables are used. Instead, new Less variables for dark colors are introduced, and color scheme media queries are peppered about whenever a color is used in a rule.

I also renamed some of the existing "dark" color names to "dim" so as to not be confusing with the new "dark" prefix, removed the unused `@package_banner_background` variable, and introduced `@foreground` and `@keyword` (which were previously hard-coded colors).

I have tested this on a variety of pages and I think it looks acceptable.  I have checked every color variable to make sure that that color looks good in dark mode, including the not-available and error messages.  The only thing I've discovered that looks weird is the "load more" button in Pursuit search results, due to [this hard-coded color in the Pursuit repository](https://github.com/purescript/pursuit/blob/4a1b3726b71518ce206da9db58e2acbef2fca158/templates/search.lucius#L10) which I can fix in a separate PR.

I know there were previously some concerns about the extra maintenance burden of supporting dark mode.  I don't know what I can do to mitigate that, unfortunately, though I can say that swapping between dark mode and light mode using my browser's developer tools is as easy as cake, and the code makes it abundantly clear that dark mode is supported, so anyone modifying this in the future ought to be able to figure out that it is a concern that should be tested.  That said, the CSS has not been touched in two years at this point, so it seems like it would not be a frequent burden, at least.

Screenshot:

Prelude module shown in dark mode

---

**Checklist:**

- [X] Added myself to CONTRIBUTORS.md (if this is my first contribution)

### DIFF
--- a/app/static/pursuit.less
+++ b/app/static/pursuit.less
@@ -49,9 +49,9 @@
 /* Section: Variables
  * ========================================================================== */
 @background: rgb(255, 255, 255);
+@foreground: rgb(0, 0, 0);
 @banner_background: rgb(29, 34, 45);
-@package_banner_background: lighten(@banner_background, 30%);
-@dark_foreground: rgb(240, 240, 240);
+@dim_foreground: rgb(240, 240, 240);
 @link: rgb(196, 149, 58);
 @link_active: rgb(123, 89, 4);
 @error_background: rgb(255, 240, 240);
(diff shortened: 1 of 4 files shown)


f-f (approved):

JordanMartinez:
I'm going to try this out tonight.

app/static/pursuit.less
```diff
@@ -486,6 +574,11 @@ ol li {
   border-top: 1px solid darken(@background, 20%);
   border-bottom: 1px solid darken(@background, 20%);
   padding: 0;
+
+  @media (prefers-color-scheme: dark) {
+    border-top: 1px solid lighten(@dark_background, 20%);
+    border-bottom: 1px solid lighten(@dark_background, 20%);
```
JordanMartinez, line 580: Shouldn't these be `border-top-color` and `border-bottom-color`?

JordanMartinez (commented):
LGTM, aside from these small comments.

Besides the one thing below, could you clarify why you didn't override the colors for these selectors?
```css
.markdown-body blockquote

.markdown-body .pl-k
```

sometimes-i-send-pull-requests:
Thanks for the review!

> Besides the one thing below, could you clarify why you didn't override the colors for these selectors?
> 
> ```css
> .markdown-body blockquote
> 
> .markdown-body .pl-k
> ```

When I first looked at it, they didn't look too out-of-place to me as-is, but now that you bring it up, without further changes, `pl-k` does not look that different from non-keyword text and `blockquote`'s border is unnecessarily prominent. I've changed those colors in dark mode to make it fit in a little bit better.

I've also addressed the comment about `border-{edge}` vs `border-{edge}-color`. Please take another look, when you have a chance!

app/static/pursuit.less
sometimes-i-send-pull-requests: Good catch! Yes, these should probably specify just the color. I've changed this accordingly.

JordanMartinez:
There's one more change that needs to be added: a changelog entry.

To do that, create a new file in `CHANGELOG.d` and prefix the file's name with `feature_` (e.g. `feature_pursuit-dark-theme.md`). Then we can merge this PR.

sometimes-i-send-pull-requests:
> There's one more change that needs to be added: a changelog entry.
> 
> To do that, create a new file in `CHANGELOG.d` and prefix the file's name with `feature_` (e.g. `feature_pursuit-dark-theme.md`). Then we can merge this PR.

Thanks, done!

codingedgar:
Thank you so much 🥹

Sorry if this is too dumb of a question to ask, but is this in production? If so, how to trigger it? (So far I only see the light theme)

purefunctor:
@codingedgar Not yet, I believe https://github.com/purescript/pursuit has to be updated first before this is reflected

codingedgar:
Thank you 🫶

PR #4443, opened 2023-02-28
Include comments in CoreFn output

### LINKED ISSUE #4442: Allow outputting declaration-level comments in CoreFn
## Summary

The compiler only outputs module-level comments for CoreFn. I propose all comments are included.

## Motivation

The primary use case I have in mind is better supporting the [`purs-backend-es`](https://github.com/aristanetworks/purescript-backend-optimizer) inline directives. When one wishes to put inline directives into the source code file (as opposed to some external file), all such directives must be appear in the module-level comment because that's the only place where such comments will appear in the outputted CoreFn. Ideally, at least declaration-level comments would be outputted as well, though I don't see why all comments wouldn't be outputted.

What I have to use right now
```purs
-- @inline export add arity=2
module Foo where

add :: Int -> Int -> Int
add a b = a + b
```

What I want:
```purs
module Foo where

-- @inline export arity=2
add :: Int -> Int -> Int
add a b = a + b
```

I'm not sure why non-module comments aren't already outputted. Perhaps this was actually an oversight? AFAICT, including `--comments` will output such comments on the JS output, but not the `CoreFn` one.

## Proposal

We update the `coreFn` target to always output comments, regardless of whether the `--comments` flag is passed. I argue for this for two reasons:
1. Since `corefn` is consumed by those outputting some other target than the default `js` one, consumers will likely include their own options for how to handle comments.
2. If this isn't the default, I imagine it will produce more harm than good, especially when tools like `purs-backend-optimizer` are used.
3. It seems that module-level comments are already outputted, whether the `--comments` flag is passed or not.

A counterargument is that emitting comments makes the computer do extra work when there are tools that don't use them. Moreover, the resulting `corefn` files are unnecessarily larger when files have large or many comments.

JordanMartinez:
**Description of the change**

Fixes #4442 

---

**Checklist:**

### DIFF
--- a/CHANGELOG.d/feature_export-comments-in-corefn.md
+++ b/CHANGELOG.d/feature_export-comments-in-corefn.md
@@ -1,0 +1,1 @@
+* Export comments in `corefn` target
--- a/src/Language/PureScript/CoreFn/ToJSON.hs
+++ b/src/Language/PureScript/CoreFn/ToJSON.hs
@@ -54,9 +54,12 @@ sourceSpanToJSON (SourceSpan _ spanStart spanEnd) =
          ]
 
 annToJSON :: Ann -> Value
-annToJSON (ss, _, _, m) = object [ "sourceSpan"  .= sourceSpanToJSON ss
-                                 , "meta"        .= maybe Null metaToJSON m
-                                 ]
+annToJSON (ss, comments, _, m) = 
+  object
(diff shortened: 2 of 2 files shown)


f-f (approved):

JordanMartinez:
Closing as there's more to be discussed in the underlying issue.

PR #4445, opened 2023-03-04
Update internal scripts to latest resolvers

JordanMartinez:
**Description of the change**

The longer these go un-updated, the harder it may be to fix them later due to depending on older GHCs. It also avoids running into issues like https://gitlab.haskell.org/ghc/ghc/-/issues/19452.

Note: this shouldn't block #4444.

---

**Checklist:**

### DIFF
--- a/update-changelog.hs
+++ b/update-changelog.hs
@@ -50,7 +50,7 @@ import qualified Protolude
 
 import           Control.Monad.Fail (fail)
 import qualified Data.Aeson as JSON
-import qualified Data.Aeson.KeyMap as HM
+import qualified Data.Aeson.KeyMap as KM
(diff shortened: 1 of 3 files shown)


update-changelog.hs
```diff
@@ -37,24 +50,24 @@ import qualified Protolude
 
 import           Control.Monad.Fail (fail)
 import qualified Data.Aeson as JSON
+import qualified Data.Aeson.KeyMap as HM
```
rhendric, line 53: Let's rename `HM` -> `KM`?

rhendric (approved):
LGTM aside from one suggestion.

update-changelog.hs
JordanMartinez: Good point. Done.

PR #4451, opened 2023-03-10
Simplify imports with post qualified style

JordanMartinez:
**Description of the change**

First part of #4450. I'm breaking this up over multiple PRs because of how many files this will touch.

Changes made across `app`, `src`, and `tests`:
```diff
-import qualified Foo
+import Foo qualified
-import qualified Foo as F
+import Foo qualified as F
-import qualified Foo    as F
+import Foo qualified as F
-import           Foo
+import Foo
-import           Foo         (bar)
+import Foo (bar)
-import           "monad-logger" Foo
+import "monad-logger" Foo
```

---

**Checklist:**

- [x] Added a file to CHANGELOG.d for this PR (see CHANGELOG.d/README.md)
- [ ] Added myself to CONTRIBUTORS.md (if this is my first contribution)
- [ ] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

Issue #4450, opened 2023-03-10
Reducing the number of open imports in codebase


JordanMartinez:
There are times where haskell-language-server doesn't work for me. Sometimes, it's because of a recent GHC upgrade and I haven't updated hls to a newer version.

There are also times I'm just looking at the code on GH and I'm trying to figure where something originates.

Thoughts on reducing all open imports in a given file to just open import per file (similar to what PureScript does) where that import is `Protolude` or none at all? Is there an check that does this automatically?

In other words, changes like this:
```diff
module Something where

-import Language.PureScript.CST.Types as T
+import Language.PureScript.CST.Types (AdoBlock(..), DoStatement(..), DoBlock(..), LetIn(..), CaseOf(..), IfThenElse(..), Lambda(Lambda), RecordUpdate(..), Label(lblTok), RecordAccessor(..), LetBinding(..), Expr(..), GuardedExpr(..), Where(Where), Guarded(..), Binder(..), InstanceBinding(..), ClassFundep(..), Constraint(..), TypeVarBinding(..), Role(roleTok), Foreign(..), FixityOp(..), FixityFields(..), ValueBindingFields(..), InstanceHead(..), Instance(..), Labeled(Labeled, lblValue), ClassHead(..), Type(..), DataCtor(..), DataHead(..), Import(..), DataMembers(..), Export(..), Declaration(..), ImportDecl(ImportDecl, impQual, impKeyword, impModule, impNames), Module(Module, modDecls, modKeyword, modWhere, modImports), Wrapped(Wrapped, wrpOpen, wrpClose), QualifiedName(qualTok), Name(nameTok), TokenAnn(tokRange), SourceRange(..), SourceToken(tokAnn), Separated(Separated, sepHead), SourceStyle(..), Comment(..), LineFeed, Token(..), SourcePos(..))

thomashoneyman:
I think a qualified import (`import Language.PureScript.CST.Types as T`) is fine to keep, though it would be nice to avoid truly open imports (`import Language.Purescript.CST.Types`).

rhendric:
`import Language.PureScript.CST.Types as T` is actually open; it just also makes the names available through `T`. We need the `qualified` keyword to make it exclusively qualified.

I'm in favor of making all of our imports either `import Protolude`, a qualified import, or an explicit import. The latest release of HLint can check this but I don't think our current version can.

[...]

PR #4452, opened 2023-03-11
Use HLS to provide explicit import lists

JordanMartinez:
**Description of the change**

Second part of #4450. I used HLS to provide the import lists of each module by clicking on the file's corresponding code lens action. I started this part-way through #4451.

---

**Checklist:**


src/Language/PureScript/CST.hs
```diff
@@ -14,8 +14,6 @@ module Language.PureScript.CST
   , module Language.PureScript.CST.Lexer
   , module Language.PureScript.CST.Monad
   , module Language.PureScript.CST.Parser
-  , module Language.PureScript.CST.Print
-  , module Language.PureScript.CST.Types
```
JordanMartinez, line 18: I removed these because they aren't actually used in this file.

src/Language/PureScript/Docs/Convert/ReExports.hs
```diff
@@ -20,8 +20,22 @@ import Data.Text (Text)
...
+      isValueAlias,
+      ChildDeclaration(..),
+      ChildDeclarationInfo(ChildTypeClassMember),
+      Constraint',
+      Declaration(..),
+      DeclarationInfo(TypeClassDeclaration, ValueDeclaration),
+      InPackage,
+      Module(..) )
 
-import Language.PureScript.AST qualified as P
```
JordanMartinez, line 24: Module was unused.

src/Language/PureScript/Docs/Render.hs
```diff
@@ -15,11 +15,41 @@ import Data.Maybe (maybeToList)
...
+      keywordData,
+      keywordFixity,
+      keywordType,
+      keywordWhere,
+      sp,
+      syntax,
+      typeVar,
+      RenderedCode )
 import Language.PureScript.Docs.Types
-import Language.PureScript.Docs.Utils.MonoidExtras
```
JordanMartinez, line 20: Module was unused.

JordanMartinez: Nope. I'm just reading this diff wrong. It appears below.

src/Language/PureScript/Ide/Filter/Imports.hs
```diff
@@ -1,12 +1,36 @@
...
+import Protolude
+    ( ($),
+      Eq((==)),
+      Foldable(elem),
+      Bool(..),
+      Maybe,
+      (&&),
+      not,
+      any,
+      maybe )
```
JordanMartinez, line 13: Whoops! Let me fix this.

JordanMartinez:
Thoughts on how to best 
- convert `Foo(Foo)` to `Foo(..)`?
- put multi-line imports onto one line?

```diff
-import Foo
-  ( bar
-  , baz
-  , back
-  )
+import Foo (bar, baz, back)
```

I'm out of time for today.

app/Command/Compile.hs
```diff
@@ -12,10 +12,28 @@ import Data.Map qualified as M
 import Data.Set qualified as S
 import Data.Text qualified as T
 import Data.Traversable (for)
-import Language.PureScript qualified as P
+import Language.PureScript.AST.Declarations qualified as P
+    ( getModuleName )
```
rhendric, line 16: We should either qualify or list explicitly, but not both at once, right?

JordanMartinez: Right. Hmm... This is going to take a lot more work then.

JordanMartinez: I'm going to prefer to list explicitly for now (unless there are name conflicts) because having multiple modules imported qualified under the same alias still doesn't help me know which one imports the identifier I care about.
If we always want a module imported qualified, we can do a later pass.

rhendric: Could we instead leave anything that's currently qualified alone for now—not resolve the multi-module-alias ambiguities in this PR? Not only is that less immediate work, I suspect the resolution to those cases will merit more discussion than the rest and it might be best to merge the uncontroversial stuff first.

JordanMartinez: So... Basically leave it as `import Foo qualified as Bar (baz)`?

JordanMartinez: Or do you mean leave `import Foo as Foo` and don't expose the import list at all?

rhendric: I meant leave it as `import Foo qualified as Bar` if that's what it is approximately, just for this pass.

I think I have a good way to do the uncontroversial changes (unqualified open imports gaining explicit import lists, nicely formatted and sorted) with automation, so I'm going to submit a PR with the result of that and then the remaining cases (which should just be this multi-module-alias issue) we can hammer out manually together.

JordanMartinez:
Superceded by #4454. I won't delete this branch so I can reuse some of it when we consider splitting the aliases.

PR #4454, opened 2023-03-13
Make unqualified imports explicit

rhendric:
This skips imports that have any `as` clause whatsoever, some of which still introduce implicit identifiers into the namespace.

**Description of the change**

The relatively uncontroversial part of #4452. Hopefully this is useful as a starting point to discuss the thornier problems.

Note that some of these explicit import lists are quite long, and maybe we should consider redoing this PR to have some sort of threshold on the size of the import list above which we switch to a qualified syntax. I did this 98% with an automated script and adding such a threshold would be trivial.

Some statistics:

* Import statements modified: 809
* ... with at least 10 imported names: 84
* ... at least 15 imported names: 35
* ... at least 20 imported names: 19
* ... at least 30 imported names: 4
* Longest import statement: 53 imports in [src/Language/PureScript/CST/Convert.hs](https://github.com/purescript/purescript/compare/master...rhendric:purescript:rhendric/make-imports-explicit?expand=1#diff-95781674f96d06e494074fdaeca6a234d04680c6cf460eeba40072dc4dc76a77)

Two files received manual attention:
* [src/Language/PureScript/Pretty/Types.hs](https://github.com/purescript/purescript/compare/master...rhendric:purescript:rhendric/make-imports-explicit?expand=1#diff-0a17e0bce61f890eac2ab371044bd20f6813fd4dc86c4befccef74c85a9c6669) had its `NoPatternSynonyms` pragma removed, as this no longer confuses HLint and is required for one of the imports.
* [tests/TestCoreFn.hs](https://github.com/purescript/purescript/compare/master...rhendric:purescript:rhendric/make-imports-explicit?expand=1#diff-aae152207e87b02ce20d39d1374595d0f0bad85fa9d350f33d1a9c50ee414487) was lightly rewritten because `Result` and `Value` were accessible from both `Data.Aeson` and `Data.Aeson.Types`, and the way we were using the qualified `Aeson` prefix was inconsistent.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Pretty/Types.hs
+++ b/src/Language/PureScript/Pretty/Types.hs
@@ -1,6 +1,3 @@
--- HLint is confused by the identifier `pattern` if PatternSynonyms is enabled.
-{-# LANGUAGE NoPatternSynonyms #-}
-
 -- |
 -- Pretty printer for Types
 --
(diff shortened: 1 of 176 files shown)


JordanMartinez:
I'd be ok with a threshold. I'm not sure what a good number would be. 10 sounds fine, but I'm guessing that causes the question, "what alias should we use for module X when it's imported qualified?" to be asked more frequently.

Or, maybe we start with a larger threshold (e.g. 20), and decrease it later if we think it should be decreased.

src/Language/PureScript/Pretty/Types.hs
```diff
@@ -1,6 +1,3 @@
--- HLint is confused by the identifier `pattern` if PatternSynonyms is enabled.
-{-# LANGUAGE NoPatternSynonyms #-}
-
```
JordanMartinez, line 3: Is this no longer true because of the explicit imports?

JordanMartinez (approved):
I was hoping to get more feedback about the threshold question. But, since that hasn't happened, I'm approving so we can move forward on this refactoring.

src/Language/PureScript/Pretty/Types.hs
JordanMartinez: Oh... Read the opening thread...

rhendric:
> I was hoping to get more feedback about the threshold question.

Same. I'm going to resubmit shortly with a threshold at 15; should just be a subset of this plus a rebase.

purefunctor:
The self-hosted Linux runner failed with this error twice:
```
  haddock: internal error: /run/user/30033/ghc496521_0/ghc_610.s: hPutBuf: resource exhausted (No space left on device)
```
cc: @f-f

f-f:
Ah yes, we should bump the disk size. I think we should merge this anyways.

f-f:
@purefunctor fixed!

JordanMartinez (approved):

PR #4455, opened 2023-03-17
Add ARM64 builds/releases for Linux and macOS

f-f:
**Description of the change**

This PR adds support for building and releasing ARM64 binaries for Linux and macOS.
Since GitHub Actions' Runners do not yet support the ARM64 architecture, these builds are running on [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners), deployed on VMs on a M2 server in our office at @id3as.

The binaries build fine (see the latest run for [macOS](https://github.com/purescript/purescript/actions/runs/4447472842/jobs/7809285949) and [Linux](https://github.com/purescript/purescript/actions/runs/4447472842/jobs/7809286094)) and I have tested them on a few environments and they seem to work, but we won't know how compatible they are with the rest of the world until we have at least some prereleases out in the wild - I expect there will be a few follow-up PRs as people start to use them.

A few highlights of the patch:
- the `runs-on` is now an array, so we can precisely match on the self-hosted runners when needed, as quite a few steps in the workflow only run on the Ubuntu container
- the GH Action for setting up the Haskell environment does not work on ARM64 systems. Building stack from scratch is also somewhat broken on ARM, so the Runners are running inside a `nix-shell` that contains a properly-built `stack`. For posterity and replicability, this is the `shell.nix` file:
  ```nix
  let
    pinnedNixHash = "7c3d4d3af8e9319ccd2a74c31cf247b0fcd08bc2";
  
    pinnedNix =
      builtins.fetchGit {
        name = "nixpkgs-pinned";
        url = "https://github.com/NixOS/nixpkgs.git";
        rev = "${pinnedNixHash}";
      };
  
    nixpkgs = import pinnedNix {};
  
  in nixpkgs.mkShell {
    buildInputs = [
      nixpkgs.bash
      nixpkgs.stack
      nixpkgs.nodejs-18_x
      nixpkgs.gmp
      nixpkgs.zlib
    ];
    # This is because otherwise GHC will throw
    # "commitAndReleaseBuffer: invalid argument (invalid character)"
    shellHook = with nixpkgs; if stdenv.isLinux then ''
      export LANG=C.UTF-8
      export LC_ALL=C.UTF-8
    '' else "";
  }
  ```
- Because of the above, the binary built on Linux ARM will have its dynamic links all pointing to the nix store, so the Linux ARM build has an additional step that runs `patchelf` to make the binary compatible with the "normal" linux systems.
  Most of the build is contained in the `build.sh` file, but I didn't want to handle this situation there, as this is quite specific to our current build environment, and might change in the future.
- Nomenclature: the new release artifacts simply add an `-arm64` suffix to the OS name. Things might be cleaner overall by also adding the architecture suffix to the artifacts that we already release - especially because we have `linux64`, which sorta does that already - but I instead kept them as they were for backwards compatibility.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
       - # We need a proper Git repository, but the checkout step will unpack a tarball instead of doing a clone
         # if the Git version is less than 2.18.
         name: "(Linux only) Install a newer version of Git"
-        if: "${{ matrix.os == ['ubuntu-latest'] }}"
+        if: contains(matrix.os, 'ubuntu-latest')
         run: |
           . /etc/os-release
           echo deb http://deb.debian.org/debian "$VERSION_CODENAME"-backports main >> /etc/apt/sources.list
@@ -71,7 +71,7 @@ jobs:
         # Note: here we exclude the self-hosted runners because this action does not work on ARM
         # and their Haskell environment is instead provided by a nix-shell
         # See https://github.com/purescript/purescript/pulls/4455
-        if: "${{ matrix.os != ['ubuntu-latest'] && !contains(matrix.os, 'self-hosted') }}"
+        if: "!contains(matrix.os, 'ubuntu-latest') && !contains(matrix.os, 'self-hosted')"
         uses: "haskell/actions/setup@v1"
         with:
           enable-stack: true
@@ -79,7 +79,7 @@ jobs:
           stack-no-global: true
 
       - name: "(Linux only) Check Stack version and fix working directory ownership"
-        if: "${{ matrix.os == ['ubuntu-latest'] }}"
+        if: contains(matrix.os, 'ubuntu-latest')
         run: |
           [ "$(stack --numeric-version)" = "$STACK_VERSION" ]
           chown root:root .
@@ -104,7 +104,7 @@ jobs:
         run: "ci/fix-home ci/build.sh"
 
       - name: "(Linux only) Build the entire package set"
-        if: "${{ matrix.os == ['ubuntu-latest'] }}"
+        if: contains(matrix.os, 'ubuntu-latest')
         # We build in this directory in build.sh, so this is where we need to
         # launch `stack exec`. The actual package-set building happens in a
         # temporary directory.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
           - os: ["self-hosted", "macos", "ARM64"]
           - os: ["self-hosted", "Linux", "ARM64"]
 
-    runs-on: ${{ matrix.os }}
+    runs-on: "${{ matrix.os }}"
     container: "${{ matrix.image }}"
 
     outputs:
@@ -55,13 +55,12 @@ jobs:
       - # We need a proper Git repository, but the checkout step will unpack a tarball instead of doing a clone
         # if the Git version is less than 2.18.
         name: "(Linux only) Install a newer version of Git"
-        if: contains(matrix.os, 'ubuntu-latest')
+        if: "contains(matrix.os, 'ubuntu-latest')"
         run: |
           . /etc/os-release
           echo deb http://deb.debian.org/debian "$VERSION_CODENAME"-backports main >> /etc/apt/sources.list
           apt-get update && apt-get install -y git/"$VERSION_CODENAME"-backports
       - uses: "actions/checkout@v2"
-
       - uses: "actions/setup-node@v2"
         with:
           node-version: "16"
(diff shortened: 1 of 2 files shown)


garyb (approved):
As far as I understand it, looks good to me! Will be great to have these builds.

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ defaults:
     shell: "bash"
 
 env:
-  CI_PRERELEASE: "${{ github.event_name == 'push' }}"
+  CI_PRERELEASE: "${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}"
```
rhendric, line 16: This shouldn't be necessary; 'push' events only trigger on the master branch, as seen a few lines above.

.github/workflows/ci.yml
```diff
@@ -37,12 +37,14 @@ jobs:
...
             image: haskell:9.2.5@sha256:2597b0e2458165a6635906204f7fac43c22e7d2a46aca1235a811194bb6cd419
-          - os: "macOS-11"
-          - os: "windows-2019"
+          - os: ["macOS-11"]
+          - os: ["windows-2019"]
+          - os: ["self-hosted", "macos", "ARM64"]
+          - os: ["self-hosted", "Linux", "ARM64"]
 
-    runs-on: "${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
```
rhendric, line 47: Unnecessary. It's easy to confuse the YAML outer language with the GitHub Actions inner language, and I was intentionally very pedantic with the quotation marks in part to help clarify this. The `${{ }}` interpolation is a feature of GitHub Actions and not YAML; in the YAML encoding, this is still a string, even if it refers to a list once GitHub Actions interprets it.

.github/workflows/ci.yml
```diff
@@ -53,7 +55,7 @@ jobs:
       - # We need a proper Git repository, but the checkout step will unpack a tarball instead of doing a clone
         # if the Git version is less than 2.18.
         name: "(Linux only) Install a newer version of Git"
-        if: "${{ runner.os == 'Linux' }}"
+        if: contains(matrix.os, 'ubuntu-latest')
```
rhendric, line 58: Same feedback, keep quotes please (and in similar lines below).

.github/workflows/ci.yml
```diff
@@ -125,15 +130,38 @@ jobs:
...
+        if: runner.os == 'Linux' && runner.arch == 'ARM64'
+        working-directory: "sdist-test"
+        # The self-hosted build happens inside a nix-shell that provides a working stack binary
+        # on ARM systems, and while the macOS binary is fine - because macOS binaries are almost
+        # statically linked), the linux ones are all pointing at the nix store.
+        # So here we first point the binary to the right linker that should work on a generic linux,
+        # and then fix the RUNPATH with the right location to load the shared libraries from
+        run: |
+          patchelf --set-interpreter /usr/lib/ld-linux-aarch64.so.1 $(stack path --local-doc-root)/../bin/purs
+          patchelf --set-rpath /usr/lib/aarch64-linux-gnu $(stack path --local-doc-root)/../bin/purs
```
rhendric, line 143:
```suggestion
          patchelf --set-interpreter /usr/lib/ld-linux-aarch64.so.1 --set-rpath /usr/lib/aarch64-linux-gnu $(stack path --local-doc-root)/../bin/purs
```

.github/workflows/ci.yml
```diff
@@ -125,15 +130,38 @@ jobs:
...
+          os_arch="${{ runner.arch }}"
           case "$os_name" in
             Linux)
-              bundle_os=linux64;;
+              case "$os_arch" in
+                ARM64)
+                  bundle_os=linux-arm64;;
+                *)
+                  bundle_os=linux64;;
+              esac
```
rhendric, line 157: Needs a `;;`, I think?

.github/workflows/ci.yml
```diff
@@ -125,15 +130,38 @@ jobs:
...
             macOS)
-              bundle_os=macos;;
+              case "$os_arch" in
+                ARM64)
+                  bundle_os=macos-arm64;;
+                *)
+                  bundle_os=macos;;
+              esac
```
rhendric, line 164: Same here?

rhendric (changes requested):
I think the missing `;;`s in the bundler command is the only actually blocking comment here; the rest are style nits where you can tell me to take a hike.

.github/workflows/ci.yml
f-f: They do in normal conditions, but when implementing things in CI one might want to run the workflows on push (before putting a PR up), as in e.g. [this commit](28534ee9036cab7eb3048ca7e7fe698d198337f7) - so this an additional safeguard to make sure one does not accidentally pushes a prerelease.

.github/workflows/ci.yml
f-f: I don't mind either version, will add quotes everywhere (and arguably that should be the preferred way of doing things, since it's just safer)

f-f:
@rhendric thanks for the review! No hikes, I have applied most of the feedback 😄

Issue #3862, opened 2020-05-05
Atomic `copyFile` to `foreign.js` breaks filewatchers


milesfrain:
## Description

Filewatchers of bundlers like [parcel](https://github.com/parcel-bundler/parcel/issues/221) and [webpack](https://webpack.js.org/guides/development/#adjusting-your-text-editor) break when their inputs are atomically overwritten. `purs` does this for `foreign.js` files, and so automatic rebundling is not triggered upon changes in FFI .js code.

## To Reproduce

A repo with steps to reproduce this issue is available here:
https://github.com/milesfrain/ffi-reload

## Expected behavior

Ideally `output/**/foreign.js` files would be written in the same way as `output/**/index.js` files, with `writeTextFile`:
https://github.com/purescript/purescript/blob/1d78a363dbded0e9d0714b0e7176317cac015257/src/Language/PureScript/Make/Actions.hs#L225
instead of atomic `copyFile`:
https://github.com/purescript/purescript/blob/1d78a363dbded0e9d0714b0e7176317cac015257/src/Language/PureScript/Make/Monad.hs#L186

While it is less efficient to read-then-write these files, the performance hit should be negligible with these small FFI files, and it seems like a worthwhile tradeoff to improve interoperability with the web development ecosystem.

## Additional context

@blonkhart (on slack) shared a workaround for `parcel` using an updated `chokidar` dependency:
```json
// package.json
  "resolutions": {
    "parcel/**/chokidar": "2.1.3"
  }
```

## PureScript version

0.13.6

blankhart:
A few additional thoughts for what they're worth:

* The ultimate cause seems to be that an editor's safe-write behavior causes changes to `.js` files to be processed as replacements of the old file with a new file rather than changes to the old file, and a bundler only looks for changes of the latter kind. 

* For `parcel`, the `chokidar` resolution is actually to an earlier version (current is v3+).  When using this approach `yarn` emits a warning that this will break with Node v14+.  For the source see [this `parcel` issue](https://github.com/parcel-bundler/parcel/issues/221).

* For `webpack` and maybe `rollup`, users encountering this issue potentially can work around it by configuring the bundler to watch the output directory rather than watching for change notifications from the file system (I haven't tried this with either).

[...]

PR #4458, opened 2023-03-26
Replace copying foreign js with writing (avoid permissions denied error)

wclr:
It allows to avoid permission denied errors while copying files on Windows (because of file watchers).

**Description of the change**

When a file watcher is working over the `output` folder during a compilation (`spago build`) when a large number of modules is compiled I usually get such an error on some of the "unlucky" files:

```bash
I/O error while trying to copy file: ...\project\My\Module.js -> output\My.Module\foreign.js

    output\My.Module\.coFFF.tmp: copyFile:atomicCopyFileContents:withReplacementFile:renameFile:renamePath:MoveFileEx "\\\\...output\My.Module\.coFFF.tmp" Just "\\\\?\\...output\My.Module\foreign.js": permission denied (Access is denied.)
```
As the error clearly points to the `copyFile` as the source of the problem, I tried to replace it with writing, and now build works without the aforementioned flaw.

I've added a new version function of `readTextFile` that returns original `ByteString` along with `Text` data, to avoid unnecessary encoding.

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Make/Actions.hs
+++ b/src/Language/PureScript/Make/Actions.hs
@@ -46,7 +46,7 @@ import Language.PureScript.Docs.Prim qualified as Docs.Prim
 import Language.PureScript.Docs.Types qualified as Docs
 import Language.PureScript.Errors (MultipleErrors, SimpleErrorMessage(..), errorMessage, errorMessage')
 import Language.PureScript.Externs (ExternsFile, externsFileName)
-import Language.PureScript.Make.Monad (Make, copyFile, getTimestamp, getTimestampMaybe, hashFile, makeIO, readExternsFile, readJSONFile, readTextFile, writeCborFile, writeJSONFile, writeTextFile)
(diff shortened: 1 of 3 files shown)


rhendric:
See previous discussion in #3862.

I'm inclined to close this; I don't think anything has changed since that discussion, and I still agree with its conclusion.

wclr:
Thanks for pointing out the discussion. The problem discussed relates to the current issue, it is about atomic operations that are not handled correctly by the build chain, though in my case these ops do not come unoticed, but cause the error.

I've investigated the problem in the chain a bit deeper, it turns out in my case that it is not about the file watching process itself nor about webpack that hardly intentionally introduces some magic locks while accessing the files that prevent new versions from being placed instead of old.

In my setup, my guess is that the problem lies in the docker and windows volumes sharing mechanics that probably leave some "phantom locks" that are not working well with atomic changes, though it would need further investigation.

I've tried different settings with debounce times, etc. but haven't been able to couple with the problem. Anyway, for now it seems I will have to use this fix in my setup to get rid of these nasty issues.

So if atomic operations for delivering compilation results are preferred then let it be.

PR #4464, opened 2023-04-08
Qualify imports in CST/Convert.hs

JordanMartinez:
Not intended to be merged, but primarily to serve as a discussion point for https://github.com/purescript/purescript/issues/4450#issuecomment-1487846602

This is what the code looks like if we use qualified imports on the two open imports in this file.

### DIFF
--- a/src/Language/PureScript/CST/Convert.hs
+++ b/src/Language/PureScript/CST/Convert.hs
@@ -34,129 +34,129 @@ import Language.PureScript.Label qualified as L
 import Language.PureScript.Names qualified as N
 import Language.PureScript.PSString (mkString, prettyPrintStringJS)
 import Language.PureScript.Types qualified as T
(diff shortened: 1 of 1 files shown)


JordanMartinez:
A few things stand out to me:
- this adds spam to `git blame` for 1 commit. Is that worth it?
- a number of imported identifiers exist only because Haskell doesn't have better record syntax (e.g. `startSourcePos $ CT.instKeyword $ CT.instHead $ CT.sepHead insts` could be written as `startSourcePos insts.sepHead.instHead.instKeyword` in PureScript)
- the alternative here is to drop the threshold used in #4454 down to `0` and import everything unqualified.

JordanMartinez:
Closing as this PR just higlights the diff caused by taking one particular approach at addressing #4450

PR #4490, opened 2023-07-20
Allow instances that require `Fail` to be empty

### LINKED ISSUE #4483: Allow instance with Fail constraint without where
## Summary

Allow class instance with Fail constraint without following `where` part.

## Motivation

Method of a class instance constrained with `Prim.TypeError.Fail` class could not be evaluated and software engineer need to write dead code. Defining stub return value for an unreachable method might be complicated. 

## Proposal

Make section `where` optional if Fail constraint is applied.

## Examples

```
import Prelude
import Prim.TypeError

class Foo t where
  foo :: t -> String

instance fooInt :: Fail (Text "") => Foo Int

```

rhendric:
A class instance declaration that has `Prim.TypeError.Fail` as a constraint will never be used. In light of this, such instances are now allowed to have empty bodies even if the class has members.

(Such instances are still allowed to declare all of their members, and it is still an error to specify some but not all members.)

**Description of the change**

Closes #4483

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Sugar/TypeClasses.hs
+++ b/src/Language/PureScript/Sugar/TypeClasses.hs
@@ -11,6 +11,7 @@ module Language.PureScript.Sugar.TypeClasses
 import Prelude
 
 import Control.Arrow (first, second)
+import Control.Monad (unless)
 import Control.Monad.Error.Class (MonadError(..))
 import Control.Monad.State (MonadState(..), StateT, evalStateT, modify)
 import Control.Monad.Supply.Class (MonadSupply)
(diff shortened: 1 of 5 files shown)


src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -336,26 +337,33 @@ typeInstanceDictionaryDeclaration sa@(ss, _) name mn deps className tys decls =
 
   let declaredMembers = S.fromList $ mapMaybe declIdent decls
 
-  case filter (\(ident, _) -> not $ S.member ident declaredMembers) memberTypes of
```
rhendric, line 339: I recommend viewing this diff with the ‘Hide whitespace’ option checked.

JordanMartinez (approved):

garyb (approved):

PR #4492, opened 2023-07-22
Fix VTAs wildcard inferred warning

### LINKED ISSUE #4488: `WildcardInferredType` warning when using wildcards in a visible type application
## Description

Using wildcards in a visible type application produces a `WildcardInferredType` warning.

## To Reproduce

```
f :: forall @a. a -> a
f = identity

x :: { x :: Int }
x = f @{ x :: _ } { x: 42 }
```

Compiler output:

```

    Int

in value declaration x
PureScript(WildcardInferredType)
```

## Expected behavior

There should be no warning.

## PureScript version

0.15.10

JordanMartinez:
**Description of the change**

Fixes #4488. 

The problem initially arises when we convert the [`ExprVisibleApp` (CST value) into `VisibleTypeApp` (AST value)](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CST/Convert.hs#L338-L340). Using [`convertType`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/CST/Convert.hs#L122-L123), whenever we come across a wildcard, we always convert the resulting value into an `UnnamedWildcard`. 

Later, [right before we typecheck a module's declarations](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker.hs#L614), we update all `UnnamedWildcard` to `IgnoreWildcard` if they appear in a specific context via [`ignoreWildcardsUnderCompleteTypeSignatures`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Linter/Wildcards.hs#L10-L47). Because Visible Type Applications are a separate concept, they aren't converted here. Presumably, while typechecking, the compiler emits a warning for each `UnnamedWildcard` found, which produces the issue.

Thus, there are two ways to resolve this:
1. (this PR) we update `convertType` to take another arg indicating whether it was called while converting an `ExprVisibleApp` into a `VisibleTypeApp`. If it was, then convert any wildcards into `IgnoreWildcard` in the first place. This solves the root of the problem but any other usages of `convertType` don't have this special rule in place.
1. (not this PR) we update the `ignoreWildcards*` function to also account for `VisibleTypeApp`. This is a smaller change but comes at the cost of another traversal through the AST.

AFAICT, `convertType` is only used in two other places, both of which I think aren't affected by this change:
- [Ide/CaseSplit.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Ide/CaseSplit.hs#L128)
- [Interactive/Parser.hs](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Interactive/Parser.hs#L108)

---

**Checklist:**


rhendric:
Don't all the functions that use `convertType` in this file need to be parameterized? What happens if there's a wildcard in a constraint in a type application, for example?

JordanMartinez:
> Don't all the functions that use `convertType` in this file need to be parameterized? What happens if there's a wildcard in a constraint in a type application, for example?

Could you give an example of what you mean?

rhendric:
This is a highly contrived example, but I imagine something like this:

```purs
class Foo a b c | a -> b c where
  fooMember :: a -> b

f :: forall @a. Array a -> Array (Array a)
f as = [as]

x :: forall c. Array (Foo Int Boolean c => Int -> Boolean)
x = [fooMember]

y :: forall c. Array (Array (Foo Int Boolean c => Int -> Boolean))
y = f @(Foo Int Boolean _ => _) x -- neither wildcard should warn IMO
```

JordanMartinez:
> Don't all the functions that use `convertType` in this file need to be parameterized? What happens if there's a wildcard in a constraint in a type application, for example?

Using your contrived example, no wildcard is emitted. 
But to answer your question, no I don't think they need to parameterized. If there's a wildcard anywhere in the type expression passed to the VTA, it'll be parsed as an `IgnoreWildcard` rather than an `UnnamedWildcard`.

rhendric:
Tests that no warnings are issued need to be in `tests/purs/warning`. If after making that change the test works than I have no further objections, but I believe it won't—in which case, I think parameterizing `convertConstraint` is sufficient, as it looks like the only function that `convertType'` uses to call back into itself (though if you could double-check me on this, I'd appreciate it).

JordanMartinez:
Ah.... No, you're right. I see the issue now.

JordanMartinez:
>  it looks like \[`convertConstraint` is\] the only function that `convertType'` uses to call back into itself (though if you could double-check me on this, I'd appreciate it).

I can confirm that `convertConstraint` is the only other `convert*` function used in `convertType` whose implementation calls back into `convertType`.

rhendric (approved):

PR #4494, opened 2023-07-24
Drop UnusableDeclaration; update related errors

### LINKED ISSUE #4487: 0.15.10 doesn't allow class methods without class type variables in signature
## Description

Take something like [`reflectSymbol`](https://pursuit.purescript.org/packages/purescript-prelude/6.0.1/docs/Data.Symbol#v:reflectSymbol). Prior to 0.15.10 there was no choice but to give it a `Proxy` parameter:

```
s = reflectSymbol (Proxy :: _ "whatever")
```

After 0.15.10, however, one would expect to be able to do better:

```
s = reflectSymbol @"whatever"
```

But this is impossible, because the compiler will return an `UnusableDeclaration` error when checking such class.

## To Reproduce

```
class C a where x :: String
instance C Int where x = "works!"

y = x @Int
```

This doesn't compile, returning an `UnusableDeclaration` error.
Technically the class declaration alone is sufficient for repro, I just added instance and use site for better illustration.

## Expected behavior

The code should compile and `y` should be equal to `"works!"`

## PureScript version

0.15.10

### LINKED ISSUE #4495: Is `moveQuantifiersToFront` still needed?
## Description

For context, see https://github.com/purescript/purescript/issues/4487#issuecomment-1648314157

Previously, type classes with type variables in the class member must have been introduced either via a `forall` on the type class member type signature or via a type variable from the type class head.

```purs
class Foo tyVarFromClassHead where
  foo :: forall tyVarFromMemberSig. tyVarFromMemberSig -> tyVarFromClassHead -> invalidTyVar
```

With the advent of VTAs, we can stop using `Proxy` args to indicate to the compiler which type class instance to use to solve a type class constraint. This further enables us to define type classes with type parameters in the class head that do not appear within a class member's type signature. 

```purs
class Foo tyVarFromClassHead where
  foo :: Int
```

However, this decision also allows the type variables introduced at the member's type signature to shadow the ones introduced in the type class head:

```purs
class Foo a where
  foo 
    :: forall a
     . a -- 'a' here refers to member's 'a', not class head 'a'
    -> Int
```

The full type signature of `foo` is then:
```purs
foo :: forall @a. Foo a => (forall a. a -> Int)
foo = -- <instance implementation>
```

A long time ago (i.e. `purs@0.5.5`-era), the `moveQuantifiersToFront` function was added to fix #376. However, this function does not appear to respect scope and seems to incorrectly convert the above type signature into:
```purs
foo :: forall @a a. Foo a => a -> Int
foo = -- <instance implementation>
```

When coupled with [`addVisibility`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Types.hs#L553-L562), the type signature further deteriorates by copying the type variable visibility of the class head's `a` into the member signature's `a`:
```purs
foo :: forall @a @a. Foo a => a -> Int
foo = -- <instance implementation>
```

So, should `moveQuantifiersToFront` be removed? Or is it still needed?

## PureScript version

0.15.10 (if `UnusedDeclaration` error is removed from that version via #4494)

JordanMartinez:
**Description of the change**

- Fixes #4487. 
- Fixes #4495.

With the advent of VTAs, all type class instances can be determined via VTAs. This PR removes that error and the check that throws it. I've added failing and passing tests to see what else needs to be done once the error is removed. 

The tests in `failing` show current error outputs. I think it would be nice if the "Consider adding a type annotation" message would instead say "Consider using Visible Type Application(s) to determine the unknown type variable(s): `useSingleton @Int`. However, I'm not sure how that information should propagate down into the error message.

---

**Checklist:**


JordanMartinez:
> The tests in `failing` show current error outputs. I think it would be nice if the "Consider adding a type annotation" message would instead say "Consider using Visible Type Application(s) to determine the unknown type variable(s): `useSingleton @Int`. However, I'm not sure how that information should propagate down into the error message.

I think this is the last thing that needs to be addressed before this is ready for review.

rhendric:
> > I think it would be nice if the "Consider adding a type annotation" message would instead say "Consider using Visible Type Application(s) to determine the unknown type variable(s): `useSingleton @Int`. However, I'm not sure how that information should propagate down into the error message.
> 
> I think this is the last thing that needs to be addressed before this is ready for review.

This might be a bit tricky, but if it helps to get your started the things I'd look at are:
* `checkFunctionApplication'` in `L.PS.TypeChecker.Types`—in the case that handles `ForAll`, look at the visibility marker and if it's visible, add some sort of hint to the proceedings that stores the current function expression (for display in the error message) and the new unknown (for checking if the error is actually associated with this expression)
* `entails` in `L.PS.TypeChecker.Entailment`—on the line where `unique` is called, you should have access to the hints in effect when a dictionary placeholder was introduced via the `hints` variable. The final argument to `unique` is currently a `Bool` indicating whether there are relevant unknowns in the constraint being solved, and this argument gates the ‘Consider adding a type annotation’ message. Perhaps this should be a three-constructor type for ‘no unknowns’, ‘unknowns that aren't from visible `ForAll`s’, ‘unknowns that are’.

JordanMartinez:
I've been trying this out and it should work. However, the hint I add is getting filtered out at some point before I hit the unknowns calculation. I'm still figuring out why, but i think it might be due to the usage of `withoutWarnings` as i don't know how else that hint gets removed.

JordanMartinez:
AFAICT, adding a hint in `checkFunctionApplication'` will not propagate to `entails`. The `checkHints` value appears to be reset by the time the control flow gets to `entails`.

rhendric:
> The `checkHints` value appears to be reset by the time the control flow gets to `entails`.

That's true, but you should see a few lines down in `checkFunctionApplication'` in the `ConstrainedType` branch where the current set of hints is collected and stored in the TCD placeholder. AIUI that's where the hints come from in `entails`; are they getting lost before or after being stored in the TCD?

JordanMartinez:
> > The `checkHints` value appears to be reset by the time the control flow gets to `entails`.
> 
> That's true, but you should see a few lines down in `checkFunctionApplication'` in the `ConstrainedType` branch where the current set of hints is collected and stored in the TCD placeholder. AIUI that's where the hints come from in `entails`; are they getting lost before or after being stored in the TCD?

I think the control flow doesn't go through that case. Using the below code to test my efforts, adding a `trace` to that case doesn't print anything:
```purs
-- @shouldFailWith NoInstanceFound
module Main where

import Prelude

class Single tyNotAppearInBody where
  useSingle :: Int

instance Single Int where useSingle = 0

single :: Int
single = useSingle
```

```hs
checkFunctionApplication' fn (ConstrainedType _ con fnTy) arg = do
  dicts <- getTypeClassDictionaries
  hints <- getHints
  let
    isVta = \case
      ErrorInVisibleTypeApplication _ _ -> True
      _ -> False
  traceM ("hints have VTA hint? ... " <> show (length (filter isVta hints)))
  checkFunctionApplication' (App fn (TypeClassDictionary con dicts hints)) fnTy arg
```

rhendric:
Oh, I see where I pointed you wrong. I think that case'll be covered in `instantiatePolyTypeWithUnknowns` instead of `checkFunctionApplication'`, because—duh—it isn't a function application. Should work the same way, though. Good reminder to have tests that cover both this nullary case and the case where the class member takes arguments.

JordanMartinez:
> Oh, I see where I pointed you wrong. I think that case'll be covered in `instantiatePolyTypeWithUnknowns` instead of `checkFunctionApplication'`, because—duh—it isn't a function application. Should work the same way, though. Good reminder to have tests that cover both this nullary case and the case where the class member takes arguments.

Mm... you didn't mislead me. Nothing is printed when I add a `trace` to the `Forall` and `ConstrainedType` cases in `instantiatePolyTypeWithUnknowns`.

JordanMartinez:
Printing the values in the `Forall` case for `checkFunctionApplication'`:
```
checkFunctionApplication' - Unknown: TUnknown () 12

checkFunctionApplication' - fn: 
	TypedValue True (Constructor () "tyNotAppearInBody") (Just (SkolemScope {runSkolemScope = 0}))) (Just (SkolemScope {runSkolemScope = 1})))

checkFunctionApplication' - ty: 
	ForAll () TypeVarVisible "tyNotAppearInBody" (Just (TypeVar () "k")) (TypeApp () (TypeApp () (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Function"}))) (TypeApp () (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Record"}))) (RCons () (Label {runLabel = "useSingle"}) (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Int"}))) (KindApp () (REmpty ()) (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Type"}))))))) (TypeApp () (KindApp () (TypeConstructor () (Qualified (ByModuleName (ModuleName "Main")) (ProperName {runProperName = "Single$Dict"}))) (TypeVar () "k")) (TypeVar () "tyNotAppearInBody"))) (Just (SkolemScope {runSkolemScope = 0}))

checkFunctionApplication' - Unknown: TUnknown () 13

checkFunctionApplication' - fn: 
	TypedValue True (Constructor () "tyNotAppearInBody") (Just (SkolemScope {runSkolemScope = 0}))) (Just (SkolemScope {runSkolemScope = 1})))

checkFunctionApplication' - ty: 
	TypeApp () (TypeApp () (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Function"}))) (TypeApp () (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Record"}))) (RCons () (Label {runLabel = "useSingle"}) (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Int"}))) (KindApp () (REmpty ()) (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Type"}))))))) (TypeApp () (KindApp () (TypeConstructor () (Qualified (ByModuleName (ModuleName "Main")) (ProperName {runProperName = "Single$Dict"}))) (TUnknown () 12)) (TypeVar () "tyNotAppearInBody"))

**Comment by JordanMartinez** [comment]
Taking a look at this again, I added a few more trace statements to figure out what the control flow path is to `entails` in this context. Turns out the above `checkFunctionApplication'` is called due to the instance in the example code. If I remove that and simplify the code to just the following, that control path never occurs:

```purs
module Main where

class Single tyNotAppearInBody where
  useSingle :: Int

single :: Int
single = useSingle
```

Following the control flow from there, the control flow starts at `check' v@(Var _ var) ty`. If I trace the `repl` value at that point, I get the following structure:

```
ForAll 
  () 
  TypeVarInvisible 
  "t12" 
  (Just (TypeConstructor () (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Type"})))) 
  (ForAll 
    () 
    TypeVarVisible 
    "tyNotAppearInBody" 
    (Just (TypeVar () "t12")) 
    (ConstrainedType 
      () 
      (Constraint 
        { constraintAnn = ()
        , constraintClass = Qualified (ByModuleName (ModuleName "Main")) (ProperName {runProperName = "Single"})
        , constraintKindArgs = [TypeVar () "t12"]
        , constraintArgs = [TypeVar () "tyNotAppearInBody"]
        , constraintData = Nothing}) 
        (TypeConstructor () 
          (Qualified (ByModuleName (ModuleName "Prim")) (ProperName {runProperName = "Int"}))))
    (Just (SkolemScope {runSkolemScope = 0}))) 
  (Just (SkolemScope {runSkolemScope = 1}))
```

In other words, `repl`'s expression is:
```
forall t12. forall @(tyNotAppearInBody :: t12). Main.Single tyNotAppearInBody => Prim.Int
```

From there, [it calls `subsumes`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker/Types.hs#L831), which calls [`subsumes'`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker/Subsumption.hs#L68), which on a `Forall` case will [generate an unknown and replace the type variables with that](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker/Subsumption.hs#L78-L80). THIS is where we need to add the `ErrorInVisibleTypeApplication` hint.

So... 
- at `check' v@(Var _ v)`, we currently update the hints to include the value we are trying to check
- at `subsumes`, we currently update the hints to include the type of the expression we are checking
- at `subsubmes' (Forall{})`, this is where we need to insert the `ErrorInVisibleTypeApplication` hint
- at `unknownsInAllCoveringSets`, we pull all the information out of the hints to produce a nicer error message (still a WIP, but it's getting there)
- we don't actually render `ErrorInVisibleTypeApplication`.

src/Language/PureScript/TypeChecker/Entailment/Coercible.hs
```diff
@@ -531,7 +531,7 @@ insoluble k a b =
   -- "Consider adding a type annotation" hint, because annotating kinds to
   -- instantiate unknowns in Coercible constraints should never resolve
   -- NoInstanceFound errors.
-  errorMessage $ NoInstanceFound (srcConstraint Prim.Coercible [k] [a, b] Nothing) [] (any containsUnknowns [a, b])
+  errorMessage $ NoInstanceFound (srcConstraint Prim.Coercible [k] [a, b] Nothing) [] (if any containsUnknowns [a, b] then Unknowns else NoUnknowns)
```
JordanMartinez, line 534: This was a 'quick change', but it needs to be thought out more.

tests/purs/failing/ClassHeadNoVTA3.out
```diff
@@ -0,0 +1,23 @@
...
+at tests/purs/failing/ClassHeadNoVTA3.purs:8:16 - 8:36 (line 8, column 16 - line 8, column 36)
+
+  No type class instance was found for
+  [33m                       [0m
+  [33m  Main.MultiMissing Int[0m
+  [33m                    t0 [0m
+  [33m                       [0m
+  The instance head contains unknown type variables. Consider using visible type application(s):
+  [33m                             [0m
+  [33m  useMultiMissing @norThisOne[0m
```
JordanMartinez, line 12: This output still needs to be improved to something like `useMultiMissing @Int @norThisOne` because `@Int` is already being used.

tests/purs/failing/ClassHeadNoVTA5.out
```diff
@@ -0,0 +1,24 @@
...
+at tests/purs/failing/ClassHeadNoVTA5.purs:8:15 - 8:29 (line 8, column 15 - line 8, column 29)
+
+  No type class instance was found for
+  [33m                     [0m
+  [33m  Main.MultiFdBidi t0[0m
+  [33m                   t1[0m
+  [33m                     [0m
+  The instance head contains unknown type variables. Consider using visible type application(s):
+  [33m                                               [0m
+  [33m  useMultiFdBidi @tyNotAppearInBody @norThisOne[0m
```
JordanMartinez, line 12: Here, only the first ty var is needed due to the functional dependency.

tests/purs/failing/ClassHeadNoVTA4.out
```diff
@@ -0,0 +1,24 @@
...
+at tests/purs/failing/ClassHeadNoVTA4.purs:8:11 - 8:21 (line 8, column 11 - line 8, column 21)
+
+  No type class instance was found for
+  [33m                 [0m
+  [33m  Main.MultiFd t0[0m
+  [33m               t1[0m
+  [33m                 [0m
+  The instance head contains unknown type variables. Consider using visible type application(s):
+  [33m                                           [0m
+  [33m  useMultiFd @tyNotAppearInBody @norThisOne[0m
```
JordanMartinez, line 12: Here, only the first ty var is needed due to the functional dependencies.

JordanMartinez:
This is ready for an initial review. I think the three outstanding issues are:
- how to handle the `Coercible` case for its `NoInstanceFound` error (see [comment above](https://github.com/purescript/purescript/pull/4494#discussion_r1282247010))
- Whether the two changes to these errors are desirable and, if not, how to add that information back to the error message:
    - https://github.com/purescript/purescript/pull/4494/files#diff-97685fc30077018a22276876c78646214950078b7c587d44eb298c706b763efdL15-L16
    - https://github.com/purescript/purescript/pull/4494/files#diff-6fe672e690b265cd8e8c2f150137e1dd613ed53916fa0c65fa094949bd0b8728L10-L11
- whether or not to [drop irrelevant determined args](https://github.com/purescript/purescript/pull/4494/files#diff-882f8f8c804b84dfa001aea5a0b9a9d979eb1e99ca4b7d844f790a5bca328e43R491-R500)

src/Language/PureScript/TypeChecker/Entailment/Coercible.hs
JordanMartinez: I'm using [CoercibleHigherKindedData.](https://github.com/purescript/purescript/blob/master/tests/purs/failing/CoercibleHigherKindedData.purs) and its [test output](https://github.com/purescript/purescript/blob/master/tests/purs/failing/CoercibleHigherKindedData.out) to guide my thought process here.

This particular test fails to compile because of the type role `representational` on the `Proxy` type's `a` type variable. Because `Binary a` is not a newtype of `Unary` or vice versa, this particular example fails to compile. Moreover, the [`a` and `b` values](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker/Entailment.hs#L452C14-L453) passed to [`insoluble`](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/TypeChecker/Entailment/Coercible.hs#L524-L534) are not the original values used when solving the class constraint as can be seen by the test's golden output's error message.

So, even if we took a similar approach in other tests' error messages (i.e. producing the list of possible VTA expressions one can use to resolve the unknowns), using VTAs for `unaryToBinary` would not fix the problem.

Perhaps `Coercible` is a special case where listing the VTAs one could use in the error message should not be done and instead documentation on `coerce` itself would be better?

JordanMartinez:
Upon reflection, I think the current error messages outputted in this PR should be reworded to indicate that using the suggested VTA expressions _might_ address the issue. Here are a few cases that come to mind.

#### Case 1

The `Coercible` situation mentioned in [a comment above](https://github.com/purescript/purescript/pull/4494#discussion_r1283424005). 

While an error message for that situation is difficult, it would be silly not to suggest `coerce @from @to` in simpler situations. In the diff below, the user needs to update the deleted expression with something like the added expression. At the very least, suggesting `coerce @from @to` would help them discover what VTAs are if they've never seen them before:
```diff
sum :: Array Int -> Int
-sum = coerce <<< fold <<< coerce
+sum = coerce @(Additive Int) <<< fold <<< coerce @_ @(Array (Additive Int))
```

Because the message would say _might_, we can safely display information like this in both the simple and complex cases and let the user decided whether the suggested VTA expression would help.

#### Case 2

Assuming the user's supplied VTA arg is valid. Given the following code:
```purs
class Foo a b where i :: Int
instance Foo String String where i = 0

-- note: Int does not have an instance for Foo
test = foo @Int
```
we output:
```

  Foo Int t0

The instance head contains unknown type variables. Consider using visible type application(s):

  i @Int @b
```

That's silly because even supplying a second VTA arg (i.e. `i @Int @String`) won't fix the issue because there is not a `Foo` instance involving `Int`. And yet, the current PR blindly assumes that any user-supplied VTA args are correct. While it might be better to never trust the user (e.g. output `i @a @b` in this case), that seems problematic if the user forgot one or more VTA args. So, rewording it to say _might_ here indicates usage may still not resolve the problem.

#### Case 3

Order-implying type variable names. In Case 1, I listed a VTA expression of `coerce @from @to`. That implies the direction of coercion goes from `from` to `to`. However, these could easily be flipped in some cases:
```purs
foo :: forall @from @to. Coercible @to @from => to -> from
foo = coerce

-- Because `from` is the type var introduced first in the `forall`
-- the type signature below is actually: `String -> Int`
test = foo @Int @String

-- and yet the error message would suggest
--      foo @from @to
-- which the reader would understand as
--      foo :: Int -> String
-- due to how they supplied their args
```

rhendric:
I haven't given this a review yet only because it seems like the design might still be in flux, but given the lack of active conversation around it, is there a simpler thing we can do and discuss more helpful error messages when we've seen the simpler thing in action more? The basic idea of dropping `UnusableDeclaration` seems uncontroversial and I wouldn't expect it to be complicated (though again, haven't reviewed, so don't know what dragons you've faced so far).

JordanMartinez:
> I haven't given this a review yet only because it seems like the design might still be in flux, but given the lack of active conversation around it, is there a simpler thing we can do and discuss more helpful error messages when we've seen the simpler thing in action more? The basic idea of dropping `UnusableDeclaration` seems uncontroversial and I wouldn't expect it to be complicated (though again, haven't reviewed, so don't know what dragons you've faced so far).

Honestly, dropping it is the easiest thing to do. Once dropped, instead of getting an `UnuseableDeclaration` error, one will get a 

```

  Foo Int t1

The instance head contains unknown type variables.
```
error. 

The problem is that this error message doesn't push the user towards using VTAs to fix it. In the case where this error appears where a `UnusableDeclaration` error would have occurred, the only way to get an instance is if one uses VTAs (e.g. `reflectType @"foo"`). I believe that's why [Nate was in agreement with dropping `UnuseableDeclaration` so long as the error message was more helpful.](https://github.com/purescript/purescript/issues/4487#issuecomment-1642493460). In other cases, where the above error would appear even without dropping `UnusableDeclaration`, the problem cannot be fixed even if one does use VTAs because there isn't a type class instance. Figuring out how to update the error message to be correct when VTAs will fix it while not displaying it when VTAs won't is part of the difficult of the design. 

:thinking: But maybe that's all we need to do. Keep the `UnusedableDeclaration` check and augment/annotate the type class with information that VTAs will be required here. So, when we get that "The instance head contains unknown type variables." message, we additionally say something to the effect of "Using VTAs are required to get instances of this type class."

rhendric:
> augment/annotate the type class with information that VTAs will be required here

That sounds pretty smart to me! It'd have to be per-member, since some class members might reference all of the types while others don't.

JordanMartinez:
I'm hoping to get back to this once I've finished my current round of contributions to the new Spago.

JordanMartinez:
Superceded by #4513.

PR #4449, opened 2023-03-08
Upgrade to GHC 9.2.7

Unisay:
**Description of the change**

Updates GHC to 9.2.7 and the Stackage snapshot to LTS-20.13.

The changes are very similar to the https://github.com/purescript/purescript/pull/4433.

---

**Checklist:**

- [x] Added a file to CHANGELOG.d for this PR (see CHANGELOG.d/README.md)
- [x] Added myself to CONTRIBUTORS.md (if this is my first contribution)
- [ ] Linked any existing issues or proposals that this pull request should close
- [x] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

Issue #3990, opened 2021-01-18
Build with GHC-LTS (Currently 8.8.4)


waynevanson:
Purescript should be able to be built from source using the GHC 8.8.4, which offers long term support.

This would make compiling from source a breeze when using package managers such as `nix` and `xbps`, which use the LTS version of GHC only. I'm trying to add this to `void-linux/void-packages` but they don't accept prebuilt binaries like they've done in `nix`. I'm also trying to do the same for `spago`.

Upon snooping, I've been able to summarize the following list in order to complete. Is there anything else missing?

- [ ] Release purescript-0.14.x - https://github.com/purescript/purescript/pull/3943#issuecomment-706728352
- [ ] Remove `hpack` by merging #3933 - https://github.com/purescript/purescript/pull/3943#issuecomment-706728352
- [ ] Update to 8.8.4 by merging https://github.com/purescript/purescript/pull/3943

hdgarrood:
Sorry, this is explicitly a non-goal for us, it's just too much effort. We already have quite a bit of difficulty trying to coordinate releases as is, we definitely don't need the additional pressure of "we need to make another release because some upstream dependency was updated and it no longer builds". I am sure that compiling from source using nix is already possible, you just need to use the install plan that we've provided via the stack.yaml file in this repository. If you ignore that and try to build with the latest versions of everything in your distro package manager, it's definitely not going to work.

Note that using the install plan we've provided is not only important for getting the compiler to build in the first place, it's actually critical to follow the install plan to end up with a compiler that functions correctly. At least one of our dependencies (language-javascript) is pinned to a specific version because that dependency forms part of the compiler's public API, and so it's very much not an option to update whenever it updates, because updates for that library are often semver-major updates for the compiler.

waynevanson:
> At least one of our dependencies (language-javascript) is pinned to a specific version because that dependency forms part of the compiler's public API, and so it's very much not an option to update whenever it updates, because updates for that library are often semver-major updates for the compiler.

@hdgarrood That's okay, thank you for the explanation. This section is very useful to keep in the back of my mind.

I was on the fence about create patches but now it's clear that I must.

waynevanson:
For those interested, here is the git diff between `language-javascript/current` and `language-javascript/7.0.0`

https://github.com/erikd/language-javascript/compare/b493492306895be48955aaf0454314694f2e8db9...new-ast

andys8:
[...]

PR #4510, opened 2023-10-07
Lts 20.26

flip111:
Only a minimal change was required to update the GHC version. Changed the version of `happy` from `1.20.0` to `1.20.1.1`.

From: LTS Haskell 20.9 (ghc-9.2.5) Published on 2023-01-28
To: LTS Haskell 20.26 (ghc-9.2.8) Published on 2023-06-18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,8 +108,14 @@ jobs:
       - name: "(Linux only) Check Stack version and fix working directory ownership"
         if: "contains(matrix.os, 'ubuntu-latest')"
         run: |
-          [ "$(stack --numeric-version)" = "$STACK_VERSION" ]
-          chown root:root .
+          actual_stack_version=$(stack --numeric-version)
+          if [[ "$actual_stack_version" = "$STACK_VERSION" ]]
+          then
+            chown root:root .
+          else
+            echo "Actual stack version ${actual_stack_version} does not match specified stack version ${STACK_VERSION}"
+            exit 1
+          fi
 
       - uses: "actions/cache@v2"
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
         include:
           - # If upgrading the Haskell image, also upgrade it in the lint job below
             os: ["ubuntu-latest"]
-            image: haskell:9.2.5@sha256:2597b0e2458165a6635906204f7fac43c22e7d2a46aca1235a811194bb6cd419
+            image: haskell:9.2.8@sha256:daf831233850c2953f295e6e71463e84ce222bd9ea2b1fa775bc85ab60ff9f0c
           - os: ["macOS-11"]
           - os: ["windows-2019"]
           - os: ["self-hosted", "macos", "ARM64"]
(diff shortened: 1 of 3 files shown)


f-f:
As the comment in the file you changed says, we'll need to bump the docker image for CI as well

rhendric:
As a matter of policy, we don't do unmotivated version bumps; a `git bisect` takes longer when Stack needs to blow away its cache and rebuild the world every time a particular commit is crossed, so whatever benefit we get from bumping versions has to clear at least that annoyance threshold. What's the motivation for this one?

flip111:
@rhendric that's not my experience that stack needs to blow away it's cache. You can try it yourself just start a new haskell project with stack and then change the resolver to some different LTS. Builds already done on a particular LTS will build fast. Do you have a test case to reproduce your scenario where stack throws away the cache?

rhendric:
Uh, new compiler means new compilation artifacts; there's no test case involved here, that's just how it works. Do you have a motivation for this or not?

flip111:
@rhendric yes new compiler, new artifacts. That doesn't explain what you were saying about stack blowing away cache and git bisect.

The motivation is that purescript is now on a compiler from 2022 that's a year ago. Every year the 9.2.5 version will get a year older. Meanwhile improvements are made to GHC which purescript can not benefit from as long as it's stuck on 9.2.5. How many years would you wait before updating a dependency?

rhendric:
> @rhendric yes new compiler, new artifacts. That doesn't explain what you were saying about stack blowing away cache and git bisect.

Right, so in a year or two when developers are on whatever the current GHC is, hunting down a regression with `git bisect` means rebuilding the world when traversing any commit that changes the GHC. If you've retained build artifacts from two years ago in your cache, I suppose you don't have to wait any extra time, but I don't plan on doing that.

> How many years would you wait before updating a dependency?

On this project, indefinitely, unless something can't be supported anymore, there's a bug needs fixing, we want a feature from a newer version, or there's a measured performance improvement from which we want to benefit (or similar motivations; this isn't necessarily an exhaustive list). Otherwise, if it isn't broken, we don't fix it.

Previously:
https://github.com/purescript/purescript/issues/3990
https://github.com/purescript/purescript/pull/4449

flip111:
Looking at the reasons not to do it:

First issue https://github.com/purescript/purescript/pull/4449
purefunctor mentions "can become a bit tedious" without any further explanation why.

Second issue https://github.com/purescript/purescript/issues/3990
hdgarrood mentions:
* "we need to make another release because some upstream dependency was updated and it no longer builds". CI already proved that purescript compiles and all tests run fine.
* "At least one of our dependencies (language-javascript) is pinned to a specific version". The pinning of language-javascript has not been changed. Other direct dependencies have also not been changed. Only the build tool was changed, but purescript was built succesfully.

For the bisect you would need somebody who only caches for the purescript project as caches are shared between all your haskell projects. And also needs to do a bisect that spans over 2 years. And also has not done a bisect recently. And then this waiting time is incured only ones until the next active removal of the cache (like an OS install).

---

Looking at the reasons to do it:

These are the improvements to be had for upgrading the GHC version:
* https://downloads.haskell.org/~ghc/9.2.6/docs/html/users_guide/9.2.6-notes.html
* https://downloads.haskell.org/~ghc/9.2.7/docs/html/users_guide/9.2.7-notes.html
* https://downloads.haskell.org/~ghc/9.2.8/docs/html/users_guide/9.2.8-notes.html

There could be further improvements in packages but i didn't look into that.

---

If it ain't broke, don't fix it is often with an implication that the attempted improvement is risky and might backfire. But no possible risks were stated. Backfiring is an overstatement on a one-off cache rebuild imo.

rhendric:
@flip111, I appreciate your willingness to get involved in PureScript's development. We aren't interested in this sort of contribution unless the version bump contributes to another goal. But if you would like to make a future contribution that fixes a bug or implements a feature, I look forward to reviewing and eventually merging it. Please let this PR be.

PR #4513, opened 2023-10-26
Drop UnusableDeclaration; update NoInstanceFound error

### LINKED ISSUE #4487: 0.15.10 doesn't allow class methods without class type variables in signature

### LINKED ISSUE #4495: Is `moveQuantifiersToFront` still needed?

JordanMartinez:
**Description of the change**

Fixes #4495 
Fixes #4487 

A second cleaner take on #4494 using the idea I mentioned in the last comment of that PR.

With the advent of VTAs, all type class instances can be determined via VTAs. This PR removes that error and the check that throws it. Instead, we use code similar to the original check to determine which type variables in the type class head are not mentioned in the type class member's type signature and track these "VTA-required" args. When a `NoInstanceFound` error is thrown, we notify the user of these "VTA-required" args when possible. 

This approach avoids the problems of my first approach, where we try to inform the user what to write in source code to fix the error, while still notifying them of the need for VTAs. Specifically, [these problems are avoided](https://github.com/purescript/purescript/pull/4494#issuecomment-1668526160):
- Suggesting a correct usage of `coerce` with VTAs in all of its complex cases
- Assuming the user-supplied VTA arg is valid / will reference the corresponding type in a type class instance
- User confusion due to a difference in order of VTA args with their usage within the type class member's type signature

Implementation notes:

In both approaches, I need to figure out which type class member was being used when a NoInstanceFound error is thrown. In my original approach, I assumed a type class member would always be referenced via a plain `Var`. In this PR's initial work, I followed that approach, but I then removed the `Maybe` around the returned `exprFromErrorCheckingTypeHint` unsafely. Doing this revealed that the `Var` could be wrapped around other `Expr` types that would be difficult to extract correctly. (For example, `3531-2.purs` crashed because the `Var` was wrapped in a `Case`). Instead, I use `everythingOnValues` to extract all `Var`s from the expression, and then attempt to find the corresponding type class member. If none can be found, I default back to just `Unknowns` rather than `UnknownsWithVtaRequiringArgs`.

---

**Checklist:**


src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -205,7 +205,8 @@ desugarDecl
 desugarDecl mn exps = go
   where
   go d@(TypeClassDeclaration sa name args implies deps members) = do
-    modify (M.insert (mn, name) (makeTypeClassData args (map memberToNameAndType members) implies deps False))
+    let addNullVtaArgs (a, b) = (a, b, Nothing)
+    modify (M.insert (mn, name) (makeTypeClassData args (map (addNullVtaArgs . memberToNameAndType) members) implies deps False))
```
JordanMartinez, line 209: This part may be problematic as I have to put something here, but I don't have anything else to provide. That being said, I haven't looked into this code that much.

JordanMartinez:
`computeVtasNeededForMember` does not need to be a top-level function. I originally put it there because I thought it might be used elsewhere.

src/Language/PureScript/Environment.hs
```diff
@@ -124,7 +125,7 @@ initEnvironment = Environment M.empty allPrimTypes M.empty M.empty M.empty allPr
 -- in its SCC, and everything determining X is either before it in an SCC path, or in the same SCC.
 makeTypeClassData
   :: [(Text, Maybe SourceType)]
-  -> [(Ident, SourceType)]
+  -> [(Ident, SourceType, Maybe [[Text]])]
```
JordanMartinez, line 128: In this file, you'll notice a number of usages of `makeTypeClassData`. The usages here are always for defining compiler-supported type classes (e.g. `Row.Cons`) that don't have type class members, so changing the type signature here isn't problematic.

JordanMartinez:
CI is failing because the version of `purescript-console` is being updated to 6.1.0, where the `groupEnd` function was added. Thus, the `TypedHole` output includes that function in both normal and "MonadEffect m" styles. However, I'm not sure how to fix this locally. 

After updating `tests/support/bower.json` to use the `6.1.0` release for `purescript-console`, I tried running `pushd tests/support && rm -rf bower_components && bower cache clean && bower i && popd && HSPEC_ACCEPT=1 stack test --fast --ta "--match=TypedHole"` but no new output is being created on my side.

JordanMartinez:
Ah, I think I need to delete `tests/support/.last_updated`...?

src/Language/PureScript/Sugar/TypeClasses.hs
JordanMartinez: AFAICT, this should not be adding `Nothing` here and instead be adding the VTA information. Latest commits push that.

JordanMartinez:
> Ah, I think I need to delete `tests/support/.last_updated`...?

It's `.test-modules`. Deleting that is what fixed it for me.

JordanMartinez:
In [this thread](https://github.com/purescript/purescript/pull/4513#discussion_r1373526940), I thought the type class desugaring should add the VTA-required info when making the `TypeClassData` value. However, that causes some code to fail to compile as shown in the PSCI tests. These errors would appear in the package-set-compile test but the PSCI failures occur before that CI step. 

The failure below is what we get when compiling [a type class whose member's type signature uses a type synonym](https://github.com/purescript/purescript-st/blob/v6.2.0/src/Control/Monad/ST/Class.purs#L10-L11). Since we desugar code before typechecking it, I'm not sure whether I can fix that in this PR. If not, then I don't think that issue should block the present work but it should be noted in the changelog:

```
$ stack build --test --fast --ta "--match=BasicEval"
purescript> test (suite: tests, args: --match=BasicEval)


psci
  evalTests
    BasicEval.purs [ ]
[1 of 1] Compiling Control.Monad.ST.Class
Error found:          
at .:0:0 - 0:0 (line 0, column 0 - line 0, column 0)

  Type synonym Data.NaturalTransformation.NaturalTransformation is partially applied.


See https://github.com/purescript/documentation/blob/master/errors/PartiallyAppliedSynonym.md for more information,


    BasicEval.purs [✘] (1340ms)

Failures:

  tests/TestPsci/EvalTest.hs:54:14: 
  1) psci.evalTests BasicEval.purs
       uncaught exception: ExitCode
       ExitFailure 1

  To rerun use: --match "/psci/evalTests/BasicEval.purs/"

Randomized with seed 853563694
```

JordanMartinez:
Ok. This PR is ready for review.

src/Language/PureScript/Errors.hs
```diff
@@ -930,10 +928,32 @@ prettyPrintSingleError (PPEOptions codeColor full level showDocs relPath fileCon
...
+                              options -> 
+                                [ Box.moveRight 2 $ line "One of the following sets of type variables:"
+                                , Box.moveRight 2 $ paras $
+                                    map (\set -> Box.moveRight 2 $ line $ T.intercalate ", " set) options
+                                ]
+                    in
+                      [ paras
+                        [ line "The instance head contains unknown type variables."
+                        , Box.moveDown 1 $ paras $
+                            [ line $ "Note: The following type class members found in the expression require specifying their corresponding type class' type variables by using Visible Type Applications (e.g. " <> markCode "tyClassMember @Int" <> ")."]
```
rhendric, line 953:
```suggestion
                            [ line $ "Note: The following type class members found in the expression require visible type applications to be unambiguous (e.g. " <> markCode "tyClassMember @Int" <> ")."]
```

src/Language/PureScript/Environment.hs
```diff
@@ -54,9 +54,10 @@ data TypeClassData = TypeClassData
   { typeClassArguments :: [(Text, Maybe SourceType)]
   -- ^ A list of type argument names, and their kinds, where kind annotations
   -- were provided.
-  , typeClassMembers :: [(Ident, SourceType)]
-  -- ^ A list of type class members and their types. Type arguments listed above
-  -- are considered bound in these types.
+  , typeClassMembers :: [(Ident, SourceType, Maybe [[Text]])]
```
rhendric, line 57: Why is this a `Maybe`? Does `Nothing` mean something different from `Just [[]]`?

src/Language/PureScript/TypeChecker.hs
```diff
@@ -173,36 +172,38 @@ addTypeClass _ qualifiedClassName args implies dependencies ds kind = do
...
+      -> (Ident, SourceType) 
+      -> m (Ident, SourceType, Maybe [[Text]])
+    computeVtasNeededForMember coveringSets syns kinds (ident, memberTy) = do
+      memberTy' <- replaceAllTypeSynonymsM syns kinds memberTy
+      let mentionedArgIndexes = S.fromList (mapMaybe argToIndex (freeTypeVariables memberTy'))
+      let leftovers = map (`S.difference` mentionedArgIndexes) coveringSets
+      pure (ident, memberTy, if any null leftovers then Nothing else Just $ map (map (fst . (args !!)) . S.toList) leftovers)
+      where
+        argToIndex :: Text -> Maybe Int
+        argToIndex = flip M.lookup $ M.fromList (zipWith ((,) . fst) args [0..])
```
rhendric, line 202: Can this logic live in `makeTypeClassData`, where the covering sets are originally computed? Then you wouldn't need to expose `computeCoveringSets` and run that calculation again above; I think that would also keep you from having to stuff a dummy `Nothing` in the type class data.

You won't have `replaceAllTypeSynonymsM` there, but all you're using `memberTy'` for is determining free variables, and type synonyms can't introduce free variables. (Actually, I don't really understand why you added it *here*, given that, but whatever.)

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -421,9 +425,45 @@ entails SolverOptions{..} constraint context hints =
...
+
+                    findTyClassMembersInExprHint
+                      :: ErrorMessageHint
+                      -> Maybe [(Qualified Ident, [[Text]])]
+                      -> Maybe [(Qualified Ident, [[Text]])]
+                    findTyClassMembersInExprHint next acc = case (acc, next) of
+                      (Nothing, ErrorCheckingType expr _) -> Just $ tyClassMembersInExpr expr
+                      (_, _) -> acc
+                  
+                  tyClassMembers' <- foldr findTyClassMembersInExprHint Nothing hints
```
rhendric, line 464:
```suggestion
                    getECTExpr = \case
                      ErrorCheckingType expr _ -> Just expr
                      _ -> Nothing
                      
                  tyClassMembers' <- headMay $ mapMaybe (fmap tyClassMembersInExpr . getECTExpr) hints
```

src/Language/PureScript/TypeChecker/Entailment/Coercible.hs
```diff
@@ -37,7 +37,7 @@ import Data.Set qualified as S
 
 import Language.PureScript.Crash (internalError)
 import Language.PureScript.Environment (DataDeclType(..), Environment(..), TypeKind(..), unapplyKinds)
-import Language.PureScript.Errors (DeclarationRef(..), ErrorMessageHint(..), ExportSource, ImportDeclarationType(..), MultipleErrors, SimpleErrorMessage(..), SourceAnn, errorMessage)
+import Language.PureScript.Errors (DeclarationRef(..), ErrorMessageHint(..), ExportSource, ImportDeclarationType(..), MultipleErrors, SimpleErrorMessage(..), SourceAnn, errorMessage, UnknownsHint (Unknowns, NoUnknowns))
```
rhendric, line 40: Why not `..` this, in the existing style?

src/Language/PureScript/Types.hs
```diff
@@ -516,39 +516,48 @@ isMonoType _        = True
...
 -- | Replace a type variable, taking into account variable shadowing
 replaceTypeVars :: Text -> Type a -> Type a -> Type a
 replaceTypeVars v r = replaceAllTypeVars [(v, r)]
 
 -- | Replace named type variables with types
 replaceAllTypeVars :: [(Text, Type a)] -> Type a -> Type a
-replaceAllTypeVars = go [] where
-  go :: [Text] -> [(Text, Type a)] -> Type a -> Type a
-  go _  m (TypeVar ann v) = fromMaybe (TypeVar ann v) (v `lookup` m)
+replaceAllTypeVars rs = replaceAllTypeVars' (fmap (fmap Right) rs)
```
rhendric, line 529: I'm confused about all the changes in this file; what are they for?

src/Language/PureScript/Environment.hs
JordanMartinez: Yes. `Nothing` indicates all type class head variables are referenced in the member so it's not ambiguous. `Just` indicates we need to use VTAs. I'll refactor this to use a data constructor to note the difference.

rhendric: Wouldn't `[[]]` also indicate an unambiguous situation, though?

IIUC, the outer list is disjunctive and the inner list is conjunctive—a use of a member needs to choose *some* set from the outer list, and provide *all* of its variables. If that's right, then `[[]]` means a use of a member needs to choose `[]` and provide all of those variables—no variables, so it's unambiguous.

src/Language/PureScript/TypeChecker.hs
JordanMartinez: I put it there initially because my first attempt at updating `makeTypeClassData` resulted in a cycle in module dependencies. So, to get this going, I moved it to where the `UnusableDeclaration` was and finished that part. It stayed there because my next attempt at moving it back into `makeTypeClassData` and resolving the module dependency cycle didn't work.

`SynonynMap` and `KindMap` synonyms are defined in `TypeChecker/Synonyms.hs`. `KindMap` references a type defined in `Environment.hs`. So, it seemed I would need to inline at least those parts into `Environment.hs`. However, due to the import of `Errors.hs` for the `MultipleErrors` type, a cyclical dependency appeared.

If we don't need `replaceAllTypeSynonymsM` in the first place, then it should be possible to move it there.

src/Language/PureScript/TypeChecker/Entailment.hs
JordanMartinez: Are there ever multiple `ErrorCheckingType` values in `hints`? When we store the `hints` data constructor, we call `getHints` which takes the environment's hints and reverses them. I used `foldr` here to ensure I got the most recent `ErrorCheckingType` in that stack (i.e. the value that would be closer to the head of the list before the list is reversed rather than the tail of the the list before the list is reversed).

If there are, then I think this could should still use `foldr`. If not, then your code is cleaner.

src/Language/PureScript/TypeChecker/Entailment/Coercible.hs
JordanMartinez: Not intentional. Probably just how the IDE happened to import it when I autocompleted it.

src/Language/PureScript/Types.hs
JordanMartinez: That fixes #4495.

JordanMartinez: That fix is included in this PR because it only becomes noticeable once we make the changes in this PR.

src/Language/PureScript/Environment.hs
JordanMartinez: I believe the outer list is always non-empty, even if the type doesn't indicate that. See https://github.com/purescript/purescript/pull/4513/files#diff-face7661808c8804ee47e10a0fd05d5ffc7ddebbef1acccffb5463633e08ae7fL200 for context. `coveringSets` is converted to a `List`, and then mapped over to get `leftOvers`, which is then used to determine whether VTA-requiring args are present via `any null leftOvers` rather than `null leftOvers`. 

I switched it to from a `[[Text]]` to a `Maybe [[Text]]` when my check here (https://github.com/purescript/purescript/pull/4513/files#diff-882f8f8c804b84dfa001aea5a0b9a9d979eb1e99ca4b7d844f790a5bca328e43R443) wasn't using `Nothing`/`Just` to indicate this. If the list was empty, I would use `Nothing` there. I found that using that approach didn't trigger the VTA-required message when it should have.

src/Language/PureScript/TypeChecker/Entailment.hs
JordanMartinez: I think I'll just commit this code and see whether the tests pass.

src/Language/PureScript/Sugar/TypeClasses.hs
JordanMartinez: As this comment (https://github.com/purescript/purescript/pull/4513#issuecomment-1782826341) shows, putting data here causes other compile errors to arise. See the rest of the comment for a more detailed explanation.

src/Language/PureScript/TypeChecker.hs
JordanMartinez: Code has been moved.

src/Language/PureScript/Environment.hs
JordanMartinez: This comment is irrelevant now that `makeTypeClassData` adds the VTA info.

JordanMartinez:
@rhendric I believe I've addressed all your feedback.

tests/purs/failing/ClassHeadNoVTA6.out
```diff
@@ -0,0 +1,35 @@
...
+  [33m             t2[0m
+  [33m             t3[0m
+  [33m             t4[0m
+  [33m             t5[0m
+  [33m               [0m
+  The instance head contains unknown type variables.
+
+  Note: The following type class members found in the expression require visible type applications to be unambiguous (e.g. [33mtyClassMember @Int[0m).
+    [33mMain.useMulti[0m
+      b, c, e
```
JordanMartinez, line 18: I was taking a look at this again. The type class is defined as this:
```purs
class Multi a b c d e f | a c -> d f, b c -> a d where
  useMulti :: Int
```
So, isn't there at least two ways to call `useMulti` with the least number of variables?
- `b c -> a d`: once `b` and `c` are known, we also know `a` and `d`. I guess we also know what `f` is via the `a c -> d f` fun dep? So, we only need to specify what `b`, `c`, and `e` are via `useMulti @_ @b @c @_ @d @_`
- `a c -> d f`: once `a` and `c` are known, then so is `d` and `f`. However, we still need to indicate what `b` and `e` are. Thus, we need to specify `a`, `b`, `c`, and `e` here via `useMulti @a @b @c @_ @e @_`. 

However, a set of `a, b, c, e` is missing from the above message. Should that be the case? Or should this test change to better reflect multiple sets?

JordanMartinez: So, this is correct. While the first fun dep is usable, it still requires `b`, `c`, and `e` to be specified. But if `b`, `c`, and `e` are specified, we can use the second fun dep to calculate `a` and then the first fun dep to calculate `f`, such that only `b`, `c`, and `e` are needed. So the `a c -> d f` fun dep is not pointless, one would never use it primarily to find a usage's corresponding instance.

JordanMartinez: With that being said, the changelog should be updated to use a different example where one of the sets of variables needs to be provided, such as the `ClassHeadNoVTA5.purs`.

src/Language/PureScript/Types.hs
rhendric: Ah, okay. Then I have notes.

src/Language/PureScript/Types.hs
```diff
@@ -516,39 +516,48 @@ isMonoType _        = True
...
+
+replaceAllTypeVars' :: [(Text, Either Text (Type a))] -> Type a -> Type a
```
rhendric, line 531: It seems like the reason you made an `Either` version of this function is that you want to preserve the annotation on the original type var when you rename it? I'm not convinced that's desirable; the new name we're introducing is synthetic and the annotation should reflect that. What's the argument for?

src/Language/PureScript/Types.hs
```diff
@@ -598,10 +607,27 @@ quantify ty = foldr (\arg t -> ForAll (getAnnForType ty) TypeVarInvisible arg No
...
-moveQuantifiersToFront = go [] [] where
-  go qs cs (ForAll ann vis q mbK ty sco) = go ((ann, q, sco, mbK, vis) : qs) cs ty
-  go qs cs (ConstrainedType ann c ty) = go qs ((ann, c) : cs) ty
-  go qs cs ty = foldl (\ty' (ann, q, sco, mbK, vis) -> ForAll ann vis q mbK ty' sco) (foldl (\ty' (ann, c) -> ConstrainedType ann c ty') ty cs) qs
+moveQuantifiersToFront = go [] [] 
+  where
+  go qs cs = \case
+    ForAll ann vis q mbK ty sco -> do
+      let 
+        cArgs :: [Text] = cs >>= mapMaybe tyVarText . constraintArgs . snd
```
rhendric, line 615: This `tyVarText` will miss variables that are deeper in constraints, like the `b` in `Coercible a (Maybe b)`. You want to use `freeTypeVariables` instead.

src/Language/PureScript/Types.hs
```diff
@@ -598,10 +607,27 @@ quantify ty = foldr (\arg t -> ForAll (getAnnForType ty) TypeVarInvisible arg No
...
+        (q'', ty')
+          | q `elem` cArgs = do
+              let q' = genPureName q cArgs
```
rhendric, line 618: The second argument to `genPureName` also has to contain any free type variables in `ty`, I think.

rhendric (commented):
Love that this gets rid of `replaceAllTypeSynonymsM`!

A few more notes, and a partial reversal on a previous note (sorry!).

src/Language/PureScript/Environment.hs
```diff
@@ -129,10 +130,23 @@ makeTypeClassData
...
 
+    coveringSets' = S.toList coveringSets
+
+    m' = map (\(a, b) -> (a, b, addVtaInfo b)) m
+    
+    addVtaInfo :: SourceType -> [[Text]]
+    addVtaInfo memberTy = do
+      let mentionedArgIndexes = S.fromList (mapMaybe argToIndex $ freeTypeVariables memberTy)
+      let leftovers = map (`S.difference` mentionedArgIndexes) coveringSets'
+      map (map (fst . (args !!)) . S.toList) leftovers
```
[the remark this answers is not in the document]
JordanMartinez: It probably returns `[[], []]` in some contexts.

> If it were `Maybe (Set (NonEmpty Text))`, that would represent at the type level that there are no duplicates or empty variables-needed lists, and then my original confusion about what the Nothing case adds wouldn't apply.

Does the order of the `NonEmpty Text` affect whether two non empty lists are the same? I presume yes?

But otherwise, yeah, this would work and I don't need to define an ADT for this.

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -421,9 +425,43 @@ entails SolverOptions{..} constraint context hints =
               let fields = [ ("reflectType", Abs (VarBinder nullSourceSpan UnusedIdent) (asExpression ref)) ] in
               pure $ App (Constructor nullSourceSpan (coerceProperName . dictTypeName <$> C.Reflectable)) (Literal nullSourceSpan (ObjectLiteral fields))
 
-            unknownsInAllCoveringSets :: [SourceType] -> S.Set (S.Set Int) -> Bool
-            unknownsInAllCoveringSets tyArgs = all (\s -> any (`S.member` s) unkIndices)
-              where unkIndices = findIndices containsUnknowns tyArgs
+            unknownsInAllCoveringSets :: [(Ident, SourceType, [[Text]])] -> [SourceType] -> S.Set (S.Set Int) -> UnknownsHint
+            unknownsInAllCoveringSets tyClassMembers tyArgs converingSets = do
```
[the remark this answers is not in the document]
JordanMartinez: Oof!

src/Language/PureScript/Environment.hs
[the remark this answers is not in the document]
rhendric: > Does the order of the `NonEmpty Text` affect whether two non empty lists are the same? I presume yes?

Yes, but that shouldn't matter if you always generate these by filtering down the same list of class arguments. You'll never get the same non-empty list of arguments in a different order. (And it's good that you're keeping them in a list and not an [`NESet`](https://hackage.haskell.org/package/nonempty-containers-0.3.4.4/docs/Data-Set-NonEmpty.html) or something, so that they can be presented in the order in which their definition refers to them and not alphabetical order.)

src/Language/PureScript/Types.hs
JordanMartinez: My goal here was to simply reuse most of the original implementation of `replaceAllTypeVars` to implement `renameTypeVar`. In that context, I don't have a `Type a`, only a `Text`. However, the original `replaceAllTypeVars` assumes I have a `Type a`. So, I updated what I could and left everything else unchanged. That required wrapping the arg in an `Either` to support both use cases. If the original source annotation is unnecessary, I didn't know that.

src/Language/PureScript/Environment.hs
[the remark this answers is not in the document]
JordanMartinez: When I converted things to just `Set (NonEmpty Text)`, converting the `Set` back to a list produces an order that differs from the one in which they were inserted. The `ClassHeadNoVTA5.purs` revealed that change. The two ty vars are `tyNotAppearInBody` and `norThisOne`, but when I make this change, the test output prints `norThisOne` first rather than `tyNotAppearInBody`. 

I could add the index of the arg (i.e. `Set (Int, NonEmpty Text)`), but then this stops feeling `Set`y due to the additional metadata. Using `Map (NonEmpty Text) Int` seems to be the correct way to account for all of this at the type-level: uniqueness of `NonEmpty Text` value, `Int` metadata stored, difference in meaning between empty and non-empty. However, it seems a bit misleading since we don't use it to look up anything, just track additional metadata about the arg index so we can sort it properly later when converting it back to a List...

Thoughts?

src/Language/PureScript/Types.hs
JordanMartinez: Given that `ty` is
`forall a. a`
which translates to 
`ForAll ann TyVarInvisible "a" Nothing (TyVar ann2 "a") Nothing`, wouldn't `freeTypeVariables ty` in this case include `a` as a free type variable? In which case, we would always generate a new name for every `ForAll`, right?

rhendric: > wouldn't `freeTypeVariables ty` in this case include `a` as a free type variable?

Yes.

> In which case, we would always generate a new name for every `ForAll`, right?

No, because we're still only doing the renaming if ``q `elem` cArgs``. We just have to make sure that in a context where we do have to generate a new name, such as `forall a. Monoid a => forall a. Tuple a a0`, we don't rename the `a` to `a0` and end up with `forall a a0. Monoid a => Tuple a0 a0`, which is a different type.

src/Language/PureScript/Types.hs
rhendric: If there isn't a need to copy the source annotation from the natural variable to the synthetic one, isn't it a lot simpler just to create a `TypeVar` and reuse `replaceAllTypeVars` as-is?

I guess since these are annotation-polymorphic functions, `moveQuantifiersToFront` would have to accept an `a` to use, but that seems less disruptive than embedding an `Either` in `replaceAllTypeVars` and giving it two interfaces.

src/Language/PureScript/Environment.hs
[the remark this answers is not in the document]
rhendric: > is there a difference between `Nothing` and `Just Set.empty`?

Yeah, `Set.empty` is an impossible result, unless you use it as an unnatural sentinel. The set returned represents the distinct ways to clarify an ambiguous method call; if that set is empty, that naturally means that the method call is impossible to disambiguate, which doesn't make sense.

So yeah you *could* drop the `Maybe` and use `Set.empty` in place of `Nothing`, but IMO using the `Nothing` to indicate that the method doesn't need disambiguation is clearer.

> When I converted things to just `Set (NonEmpty Text)`, converting the `Set` back to a list produces an order that differs from the one in which they were inserted.

Well, yeah; does that order matter? If so, I think I'd use `Maybe (Set (NonEmpty Int))`, and add the args as `[Text]` to the `UnknownsWithVtaRequiringArgs` constructor, doing the lookup when the error is presented.

src/Language/PureScript/Types.hs
JordanMartinez: Ah, now I see. Thanks.

src/Language/PureScript/Types.hs
JordanMartinez: Fair enough.

src/Language/PureScript/Environment.hs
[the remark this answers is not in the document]
JordanMartinez: > Well, yeah; does that order matter? If so, I think I'd use Maybe (Set (NonEmpty Int)), and add the args as [Text] to the UnknownsWithVtaRequiringArgs constructor, doing the lookup when the error is presented.

The order doesn't matter per say, but it breaks my expectation. If I define a class
```purs
class MultiFdBidi tyNotAppearInBody norThisOne | tyNotAppearInBody -> norThisOne, norThisOne -> tyNotAppearInBody where
  useMultiFdBidi :: Int
```
and the error I get for this situation lists `norThisOne` first instead of `tyNotAppearInBody`, it surprises me because it doesn't match the order of the fun deps in the type class. 

The information presented doesn't change; I still have to clarify one of those to select a type class. However, I start to wonder whether there's a bug in the compiler because the order differs.

>> is there a difference between Nothing and Just Set.empty?
>
> Yeah, `Set.empty` is an impossible result, unless you use it as an unnatural sentinel

Right, it's impossible when this value is understood in its context. But without that context, it can still represent an impossible state via `Just Set.empty`. So, I'm asking why we're allowing such an impossible state to be represented if we use `Maybe (Set x)` rather than `Maybe (NESet x)`. And since I've already raised the order issue above, I guess I'm leaning towards `Maybe (NonEmptyMap (NonEmpty Text) Int)` (but likely providing a `newtype NESetWithMetadata k v = NESetWithMetadata (NonEmptyMap k v)` to clarify that the map isn't meant for lookup) so as to represent a value with these characteristics:
- `Nothing` = no ambiguity
- `Just` = ambiguous
    - keys represent unique values to prevent duplication
    - values represent additional metadata, so we can display them in the original order of fun deps when converting the set of keys back to a list

This might be overengineering this small problem, but all documentation for it seems stored within the type itself.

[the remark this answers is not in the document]
JordanMartinez: I guess the only argument I can think of against my idea is the cost of implementing it:
- an added dependency on `nonemptyMap` for this single usage when an alternative is 90% good enough
- a new module (that may have name clashes with others in the ecosystem) for `NESetWithMetadata` so it can be reused in more places since it's not specific to the compiler per say, but which doesn't have a complete API because we don't need that additional API here.

Welp. `Maybe (Map (NonEmpty Text) Int)` it is I guess.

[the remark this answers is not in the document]
JordanMartinez: > If so, I think I'd use Maybe (Set (NonEmpty Int)), and add the args as [Text] to the UnknownsWithVtaRequiringArgs constructor, doing the lookup when the error is presented.

Rereading this comment. So, rather than storing the text of the args as I'm currently doing, this would store the index to the corresponding args in the set. In other words, storing the metadata itself rather than the identifier of the type variable. When constructing the error, we lookup what the identifier is using the index. This bypasses the need to thread that additional metadata around at the cost of some unsafety that should be impossible (i.e. the lookup returning a `Maybe`).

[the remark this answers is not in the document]
JordanMartinez: So... I thought the output was based on the order of the fun deps. If it was, then the above suggestion would still produce a wrong output order because the above suggestion only looks at the type class' head's type variables. But having updated `ClassHeadNoVTA6.purs` to show a better example of multiple sets and changing the fun deps order to highlight this and updating the code to account for that, I saw that the output didn't change. Then I realized it was not dependent on the order of the fun deps but just the args themselves.

So, yes, you're suggestion is the better approach. And sorry for not understanding it earlier.

src/Language/PureScript/Errors.hs
```diff
@@ -111,7 +111,7 @@ data SimpleErrorMessage
   | NoInstanceFound
       SourceConstraint -- ^ constraint that could not be solved
       [Qualified (Either SourceType Ident)] -- ^ a list of instances that stopped further progress in instance chains due to ambiguity
-      Bool -- ^ whether eliminating unknowns with annotations might help
+      UnknownsHint -- ^ whether eliminating unknowns with annotations might help or with visible type applications are required
```
rhendric, line 114:
```suggestion
      UnknownsHint -- ^ whether eliminating unknowns with annotations might help or if visible type applications are required
```

src/Language/PureScript/TypeChecker/Entailment.hs
```diff
@@ -421,9 +426,42 @@ entails SolverOptions{..} constraint context hints =
...
+                NoUnknowns
+              where
+                unknownsRequiringVtas = do
+                  tyClassModuleName <- getQual className'
+                  let
+                    tyClassMemberVta :: M.Map (Qualified Ident) [[Text]]
+                    tyClassMemberVta = M.fromList $ mapMaybe qualifyAndFilter tyClassMembers
+                      where
+                        -- Only keep type class members that need VTAs to resolve their type class instances
+                        qualifyAndFilter (ident, _, mbVtaRequiredArgs) = flip fmap mbVtaRequiredArgs $ \vtaRequiredArgs ->
```
rhendric, line 444:
```suggestion
                        qualifyAndFilter (ident, _, mbVtaRequiredArgs) = mbVtaRequiredArgs <&> \vtaRequiredArgs ->
```

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -103,6 +103,17 @@ data HintCategory
   | OtherHint
   deriving (Show, Eq)
 
+-- |
+-- In constraint solving, indicates whether there were `TypeUnknown`s that prevented
+-- an instance from being found, and whether type annotations might help or
```
rhendric, line 108: I think drop the ‘and whether type annotations might help’?

src/Language/PureScript/Environment.hs
```diff
@@ -129,10 +130,24 @@ makeTypeClassData
...
+    coveringSets' = S.toList coveringSets
+
+    m' = map (\(a, b) -> (a, b, addVtaInfo b)) m
+    
+    addVtaInfo :: SourceType -> Maybe (S.Set (NEL.NonEmpty Int))
+    addVtaInfo memberTy = do
+      let mentionedArgIndexes = S.fromList (mapMaybe argToIndex $ freeTypeVariables memberTy)
+      let leftovers = map (`S.difference` mentionedArgIndexes) coveringSets'
+      let vtaRequiredArgs = S.fromList $ mapMaybe (NEL.nonEmpty . S.toList) leftovers
+      if S.null vtaRequiredArgs then Nothing else Just vtaRequiredArgs
```
rhendric, line 146: Oh dear, this doesn't look right.

Let me walk through this:
* `mentionedArgIndexes` is a `Set Int` of the (indices of the) arguments that participate in this member's type
* `leftovers` is a `[Set Int]` made from taking each covering set and subtracting `mentionedArgIndexes`, so there is one set in `leftovers` for every covering set
* `vtaRequiredArgs` is a `Set (NonEmpty Int)` comprising only the non-empty sets in `leftovers`—but if there are *any* empty sets in `leftovers`, then there is no need for VTA info, because there is a way to disambiguate the type arguments based on `mentionedArgIndexes` via the covering set corresponding to the empty `leftovers` element
* the result is `Nothing` only if `vtaRequiredArgs` is empty—which only happens if *all* of the sets in `leftovers` are empty

I think what you want instead is

```suggestion
      S.fromList <$> traverse (NEL.nonEmpty . S.toList) leftovers
```

And ideally one of the test cases would cover this by having:
* multiple distinct covering sets
* a member that mentions all of the arguments in one, but not all, of the covering sets
* expected error without the VTA hint

JordanMartinez: Oo... good catch!

rhendric: All of the current tests still pass when this line is reverted. Could you add a case like the following:

```purs
class TestClass a b | a -> b, b -> a where
  testMethod :: Maybe a -> Int

test :: Int
test = testMethod Nothing
```

The `NoInstanceFound` should have an ‘unknown type variables’ hint but not a VTA hint.

CHANGELOG.d/feature_replace-unused-declarations.md
```diff
@@ -0,0 +1,81 @@
...
+  the type variable in the class head `a` was
+  not mentioned in `bar`'s type signature:
+  
+  ```purs
+  class Foo a where
+    bar :: Int
+  ```
+
+  With the advent of Visible Type Applications (VTAs), 
+  one can inform the compiler via VTAs:
```
rhendric, line 15:
```suggestion
  The recently-added visible type applications (VTAs)
  can now be used to guide the compiler in such cases:
```

CHANGELOG.d/feature_replace-unused-declarations.md
```diff
@@ -0,0 +1,81 @@
...
+  ```purs
+  class Foo a where bar :: Int
+  instance Foo String where bar = 0
+  someInt = bar @String -- use the `String` instance
+  ```
+
+  However, failing to provide a VTA to direct the instance selection
+  will still produce an `InstanceNotFound` error, but this error 
+  has been updated to note which type variables in the class head
+  must be specified via Visible Type Applications
```
rhendric, line 26:
```suggestion
  Without VTAs, the compiler
  will still produce an `InstanceNotFound` error, but this error 
  has been updated to note which type variables in the class head
  can only be disambiguated via visible type applications 
```

CHANGELOG.d/feature_replace-unused-declarations.md
```diff
@@ -0,0 +1,81 @@
...
+  ```
+
+  In both cases above, the `NoInstanceFound` error will not include 
+  the VTA-required-args info when the type class member's type signature uses a type synonym.
+  For example,
+  ```purs
+  type Synonym = Array
+  class MyClass a b c where
+    tyClassMember :: Synonym a
+  ```
```
rhendric, line 81: Is this true? I tried testing the PR and couldn't reproduce this behavior; maybe it got fixed while we were revising the implementation?

rhendric (commented):
I think this might be the last round of comments from me!

CHANGELOG.d/feature_replace-unused-declarations.md
JordanMartinez: Ah, no. It's no longer true. When the `addVtaInfo` code was outside of `Environment.hs`, there was another spot that used `Environment.hs`' `makeTypeClassData`. I didn't update that usage because it led to other compiler errors. When we moved `addVtaInfo` into `Environment.hs` and removed the need for `replaceTypeSynonymsM`, the other usage got updated as well.

JordanMartinez: [This comment](https://github.com/purescript/purescript/pull/4513#issuecomment-1782826341) has more context on that usage and the resulting error.

rhendric (approved):

PR #4517, opened 2023-11-07
Fix space leak between modules during compilation

### LINKED ISSUE #4491: Regression in compilation time and memory in Purescript > 0.15.2
## Description

Regression in compilation time and memory in Purescript 0.15.9 (likely introduced in Purescript 0.15.3 by  https://github.com/purescript/purescript/pull/3915). 

At Lumi, we have a large "Types" which is generated from Haskell datatypes and custom generic code. This file is always the slowest in our compilation chain, however it significantly worsened when we tried upgrading from 0.14.7. The purs compiler RAM usage was most significant, growing to 30GB of ram during the compilation phase. @MonoidMusician recommended I try Purescript 0.15.2, and while there was a small increase from 0.14.7, the RAM was only 2GB during compilation of the large types file.

## To Reproduce

The Types file has 430 newtypes and 250 data type definitions and otherwise contains only instances. A sample newtype definition looks like:

```purescript
newtype Address
  = Address
    { "id" :: Key Address
    , "status" :: LocationStatus
    , "name" :: Maybe String
    , "type_" :: Maybe AddressType
    , "fullName" :: String
    , "company" :: Maybe String
    , "address" :: String
    , "aptSuite" :: Maybe String
    , "city" :: String
    , "country" :: String
    , "state" :: Maybe String
    , "postalCode" :: Maybe String
   }

derive instance genericAddress :: Generic Address _
derive instance newtypeAddress :: Newtype Address _

instance decodeAddress :: Decode Address where
  decode = genericDecode jsonOptions

instance encodeAddress :: Encode Address where
  encode = genericEncode jsonOptions
```
The Encode/Decode instances use purescript-foreign (the working groups port to 15.x). At first I thought maybe this library's approach was introducing a regression but I did a POC to switch to argonaught generics and found similar, if not slightly worse compiler performance.

There are also some polymorphic types such as:
```purescript
newtype ApiResponse a1 a0
  = ApiResponse
    { "warnings" :: Array a1
    , "result" :: a0
    }

derive instance genericApiResponse :: Generic (ApiResponse a1 a0) _
derive instance newtypeApiResponse :: Newtype (ApiResponse a1 a0) _

instance decodeApiResponse :: (Decode a1, Decode a0, DecodeWithOptions a1, DecodeWithOptions a0) => Decode (ApiResponse a1 a0) where
  decode = genericDecode jsonOptions

instance encodeApiResponse :: (Encode a1, Encode a0, EncodeWithOptions a1, EncodeWithOptions a0) => Encode (ApiResponse a1 a0) where
  encode = genericEncode jsonOptions
```

A sample data type is:

```purescript
data ClientUpdate
  = ClientSetAssignedUser
    (Maybe (Key User))
  | ClientReactivate
  | ClientDeactivate
    ClientDeactivation

derive instance genericClientUpdate :: Generic ClientUpdate _

instance decodeClientUpdate :: Decode ClientUpdate where
  decode = genericDecode jsonOptions

instance encodeClientUpdate :: Encode ClientUpdate where
  encode = genericEncode jsonOptions
```

## Expected behavior

The purs compiler takes no more than 2GB during compilation. 

## Additional context

This has been tested so far on a Mac M1. With Purescript 15.9 I was using the arm binaries and for 15.2 the x86 with rosetta emulation. I can also get some statistics on Linux if that would be helpful.

We also attempted generation of non-generic encodings. This definitely had a large impact on compilation memory and time, bringing it down to a more manageable  ~8GB RAM but still elevated more than 15.2.

## PureScript version

0.15.9

MonoidMusician:
To fix #4491.

For builds with a lot of warnings, memory usage grows drastically since it appears that the thunks for the warnings hang onto a lot of memory from compiling the module itself.

The goal of this change is to get memory usage for full builds back in line with partial builds.

This is just a WIP patch for now, showing a sort of minimal solution to the problem that isn't overly strict. I'd appreciate suggestions on how to accomplish it better.

`rnf x = rnf (show x)` is obviously a bad `NFData SimpleErrorMessage` instance, but it was going to be very tedious to derive `NFData` for every transitive data type. Perhaps it's better to make a dedicated function that computes just enough strictness for the errors? Or adding more strictness when computing the warnings in the first place, and to the data types themselves.

For part of our work codebase (which generates a lot of warnings), the difference is quite dramatic: 4GB of memory usage down to under 1GB, as measured by the GHC RTS profiling:

image
image

The difference is less dramatic, but still visible, when running on just our public dependencies – I suspect this is since they produce much fewer warnings.

I've been profiling with this command, though `-hc` is not a particularly helpful heap profile to track in retrospect:
```bash
$PURS +RTS -l-agu -i0.5 -hc -RTS compile -g corefn $(spago sources)
```

@mjrussell do you have a way to compile PureScript and test this out on your codebase?

**Checklist:**


mjrussell:
Wow! Nice work. I can try to give it a run this weekend. Sorry I never got you a smaller repro case like I promised

mjrussell:
@MonoidMusician I built your PR of purescript locally:
```
purs --version
0.15.12 [development build; commit: 27ec9cea1e45774647c58dfccc1665afc6d1d710 DIRTY]
```

I'm still seeing some pretty sizable memory pressure issues and longer compile times than on 15.2. I'm seeing about 26GB peak ram usage by this PR and 4m4s for a compilation of the bottleneck package whereas 15.2 maxes out at 1GB of ram and runs in 22 seconds.

MonoidMusician:
Thanks for taking the time to try it out!

I have pushed some more commits you can try: being more strict in the data kept between modules, and removing the CSE (in two commits: just dropping the CSE pass, and reverting the structural changes too).

On our codebase I actually don’t see a noticeable difference with these approaches. Just hoping it might help your codebase. If it doesn’t, we’ll have to look elsewhere!

JordanMartinez:
I feel like reverting the CSE pass doesn't give Ryan an opportunity to refactor that code. And I'd guess he hasn't because (besides potentially not having time) we still don't have a reproducible example.

MonoidMusician:
Yeah, I understand, this is just for debugging purposes to narrow down if it is CSE or not. (I don't necessarily believe it is CSE but that was one of the biggest changes that I can point to in the timeframe. Although I have another commit to look at now ...)

mjrussell:
@MonoidMusician I went to 7cba4ec and its close to 15.2 performance!

`spago build  88.56s user 7.10s system 272% cpu 35.124 total` so thats like ~3-4x worse time, but not 10-11x.

The memory pressure is greatly reduced, only about 2GB of ram, way under 26GB earlier in the PR.

I'll give the commits with just strictness a try

mjrussell:
[dec9091](https://github.com/purescript/purescript/pull/4517/commits/dec9091f7c6bea2f1bce45d690450832af31cf44) alone doesn't seem to fix the issue, in fact..it might be worse than [27ec9ce](https://github.com/purescript/purescript/pull/4517/commits/27ec9cea1e45774647c58dfccc1665afc6d1d710). 

`spago build  254.14s user 160.48s system 136% cpu 5:04.05 total`

The memory pressure was really strong, like 40GB of ram. On to the next

mjrussell:
Yeah looks to be CSE: On [33582d7](https://github.com/purescript/purescript/pull/4517/commits/33582d7b0874f2aa5426e71f0d58202a58a27883)

`spago build  87.00s user 7.48s system 246% cpu 38.336 total` 1-2 GB of ram

rhendric:
Again, if you can carve out something that you can make public and demonstrates the issue—it doesn't *have* to be minimal—there are at least a half dozen different things I can imagine that *might* improve memory use in the CSE pass and I would be happy to investigate if any of them actually help.

mjrussell:
I was writing up some explanation for why I can't make the file public b/c its internal types and we should try scripting out the generation of a 14k line file, while at the same time testing a fix that seems to alleviate the problem. 

`spago build  124.99s user 12.42s system 206% cpu 1:06.55 total` It has a bit higher mem pressure ~2GB constant but its easily workable.

Replacing `Data.Map` with `Data.Map.Strict` in `CSE.hs` does the trick.

I was expecting it to be something like that, building up thunks in a lazy structure. I thought you mind need to use RWS.Strict too but this is enough for a dramatic improvement

MonoidMusician:
interesting! it looks like we're chasing down rather separate issues then

I also have a fix coming up for more inter-module leaks WRT concurrency, so that should improve it even more

(also will help narrow down what modules are problematic)

mjrussell:
I pulled out that single Map change into https://github.com/purescript/purescript/pull/4521, thanks again!

MonoidMusician:
Okay, here's the second fix I promised:

With `+RTS -N1 -RTS`, the Haskell runtime was single threaded but the module builds were getting interleaved in some way, so the memory usage kind of bled across modules and peaked at 4.6GB. After introducing a semaphore to limit the concurrent module builds to the number of runtime threads, the memory usage looks more reasonable (a nice sawtooth graph) and peaked at 1.9GB.

image
image

I verified with the event logging that there are no longer too many modules being built simultaneously. (See trace.js if you want to replicate this.)

It matters less for builds with `-N8` (the default on my system) but still seems to make a difference.

I don't see significant differences with the CSE fix/reverts on this build, even measuring memory per-module in single threaded mode, so I guess it is orthogonal.

MonoidMusician:
More stats from the eventlogs:

with -N8 it can stay at 8 concurrent modules for 60% of the build and averages just under 6 concurrent modules ... but the user time speedup is only 2x-3x compared to a single-threaded build?

with -N6 it will stay at 6 concurrent modules for 70% of the build and average just under 5 concurrent modules, with comparable user time and a bit lower memory usage

with -N4 it gets a little slower (maybe 10%?) but still less memory usage

I can replicate these proportions without profiling, so there's probably more to investigate performance wise, but I'm happy with this for now.

MonoidMusician:
I timed each run and didn't see noticeable slow downs from the fixes here. Just speedups from the reduced memory usage, where applicable.

JordanMartinez:
@MonoidMusician Is this PR ready for review? Or are you still working on it? It's still in a draft state, so I'm unsure whether to wait or to start reviewing.

MonoidMusician:
Yeah, I'll un-draft it

src/Language/PureScript/AST/Declarations.hs
```diff
@@ -843,12 +845,18 @@ data DoNotationElement
 newtype PathTree t = PathTree (AssocList PSString (PathNode t))
   deriving (Show, Eq, Ord, Functor, Foldable, Traversable)
 
+instance (NFData t) => NFData (PathTree t) where
+  rnf (PathTree t) = rnf t
+
```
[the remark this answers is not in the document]
MonoidMusician: Haskell didn't know whether to choose newtype deriving or the default Generic instance. Same for the second one.

JordanMartinez:
Here's the steps I took. I don't know what I'm doing so feel free to correct me when I'm wrong. Since there's not a public codebase to compile this PR against, I went with the package set.
```sh
# Enter purescript root directory
$ pushd purescript
# Make a profile build
$ stack build --profile

# Install entire package set in another dir called 'z'
$ mkdir z
$ pushd z
$ spago init
$ spago install $(spago sources | cut -f 1 -d ' ' | tr '\n' ' ')
$ popd
# Setup eventlog2html in separate directory because PS's stack config 
# doesn't work on eventlog2html's latest version
$ popd
$ mkdir eventlog2html-temp-build
$ pushd eventlog2html-temp-build
$ stack build --copy-compiler-tool eventlog2html
$ popd

# Run the compiler on the package set
$ pushd purescript/z
# ... except that the below command fails...
$ stack --profile exec bash <<EOF
purs +RTS -l-agu -i1.5 -hc -RTS compile -g corefn $(spago sources)
EOF
# ... something about `purs` needing to be built with `-profile`.
# After more researching later, I passed the globs directly by producing the string manually
$ stack --profile exec -- purs +RTS -l-agu -i1.5 -hc -RTS compile -g corefn "src/**/*.purs" "test/**/*.purs" ".spago/prelude/v6.0.1/src/**/*.purs" # and so forth...

# Produce the HTML file
$ path/to/install/eventlog2html purs.eventlog
# Produce the JSON file
$ path/to/install/eventlog2html --json purs.eventlog
$ node ../debug/eventlog.js purs.eventlog.json > eventlog-output.txt
```

Outputs I get are:
- Screenshots of the HTML file's charts:
    - area-chart-profiled
    - other-chart-profiled
- [eventlog-output.txt](https://github.com/purescript/purescript/files/13638798/eventlog-output.txt)

JordanMartinez (approved):

src/Language/PureScript/Make.hs
```diff
@@ -148,20 +151,29 @@ make ma@MakeActions{..} ms = do
 
   (buildPlan, newCacheDb) <- BuildPlan.construct ma cacheDb (sorted, graph)
 
+  -- Limit concurrent module builds to the number of capabilities as set by
```
rhendric, line 154:
```suggestion
  -- Limit concurrent module builds to the number of capabilities as
```

src/Language/PureScript/AST/Binders.hs
```diff
@@ -61,7 +63,9 @@ data Binder
   -- A binder with a type annotation
   --
   | TypedBinder SourceType Binder
-  deriving (Show)
+  deriving (Show, Generic)
+
+instance NFData Binder
```
rhendric, line 68: Here and generally: can you consistently produce `NFData` instances using `deriving`, using `DeriveAnyClass` and [`DerivingStrategies`](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/deriving_strategies.html) (already in our default extensions list) as needed to enable/disambiguate this?

src/Language/PureScript/Sugar/Names/Env.hs
```diff
@@ -38,6 +39,8 @@ import Language.PureScript.Crash (internalError)
 import Language.PureScript.Environment
 import Language.PureScript.Errors (MultipleErrors, SimpleErrorMessage(..), errorMessage, errorMessage')
 import Language.PureScript.Names (Ident, ModuleName, Name(..), OpName, OpNameType(..), ProperName, ProperNameType(..), Qualified(..), QualifiedBy(..), coerceProperName, disqualify, getQual)
+import GHC.Generics (Generic)
+import Control.DeepSeq (NFData)
```
rhendric, line 43: Are any of these instances in this module needed?

rhendric (changes requested):
It's unfortunate how deep into declarations/expressions the NFData instances need to go, but I also don't see a straightforward alternative that achieves the goal, and I agree with the general direction. After some cleanup, I'm good with this.

src/Language/PureScript/CST/Layout.hs
```diff
@@ -201,7 +203,9 @@ data LayoutDelim
   | LytOf
   | LytDo
   | LytAdo
-  deriving (Show, Eq, Ord)
+  deriving (Show, Eq, Ord, Generic)
+
+instance NFData LayoutDelim
```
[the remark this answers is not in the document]
JordanMartinez: @MonoidMusician Mind if I push this fix, so that we can get this PR merged?

MonoidMusician:
Thanks for the reviews! Should be good now

@rhendric Agreed on all the places NFData needs to reach. But I'd rather be sure that the problem isn't going to come back, and I don't think there's noticeable performance impact from it. Thanks for the style tips.

@JordanMartinez I added those stack profiling instructions to eventlog.js too, thanks for that. I think the reason that `stack --profile exec bash` doesn't work (at least for me) is that my user directories take precedence over the stack provided $PATH.

Also a singular glob `.spago/*/*/src/**/*.purs` should work to pick up all spago packages (as long as there is only one version of each package downloaded).

src/Language/PureScript/CST/Layout.hs
```diff
@@ -166,6 +166,9 @@
 --    "body of a case of expression" by pushing 'LytOf' onto the layout stack.
 --    Insert the @of@ token into the stream of tokens.
 --
+
+{-# LANGUAGE DeriveAnyClass #-}
+
```
JordanMartinez, line 171: Nitpick, I know, but can we get rid of this extra whitespace? The line above/below the language pragma?

rhendric (approved):

PR #4480, opened 2023-06-25
Add option to exclude globs from given input

i-am-the-slime:
### Adds the possibility to remove files from the globs passed to the compiler

[Colocation](https://kentcdodds.com/blog/colocation) means that we can keep related files close together in our file system.
So instead of recreating the whole module tree in a `test` folder it'd be possible to structure the project as such:

```
src/
  Foo/
    Bar.purs
    BarTest.purs
```

The motivation for supporting colocation can be followed [here](https://github.com/purescript/registry-dev/issues/602).

However, it's currently not possible to exclude globs from publishing. I've opened PRs to add this functionality to the new [registry](https://github.com/purescript/registry-dev/pull/613) and [spago](https://github.com/purescript/spago/pull/967).

Verity [raised a good point](https://github.com/purescript/registry-dev/issues/602#issuecomment-1593816461), however, in that it should also be easy to pass only the files that shall be published to the compiler (for example to not include test dependencies).
This PR adds this possibility.

Note that this does not add a way to exclude globs from the IDE or PSCI, because I don't see the use case for it and wanted to keep the change small. It could also be added later.
---

**Checklist:**

- [x] Added a file to CHANGELOG.d for this PR (see CHANGELOG.d/README.md)
- [x] Added myself to CONTRIBUTORS.md (if this is my first contribution)
- [x] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

PR #4530, opened 2024-01-23
Jam/4525 globs file

### LINKED ISSUE #4525: Support `--source-globs FILE` argument due to Windows' `cmd.exe` character limitation
## Summary

Support `purs compile path/to/file/containing/source-globs`

## Motivation

AFAIU, Windows' `cmd.exe` is the only way to start a process on Windows. Unfortunately, `cmd.exe` has a [character limitation of 8191 characters](https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/command-line-string-limitation#how-to-work-around-the-limitation), unlike *nix shells. Moreover, there really isn't a workaround to this limit besides passing in such information via a file. In some recent work on the Spago rewrite, we [hit that limit](https://github.com/purescript/spago/pull/1129) due to escaping some special characters, probably unnecessarily. 

While we can fix that on the Spago side, the Spago rewrite enables (and thereby encourages to some degree) the usage of subpackages. On some very large projects, I imagine running `spago` on Windows would hit this issue at some point.

## Proposal

By passing information to `purs` via a file, one might wonder whether we should pass additional information besides the source globs via a file. The answer is no and such a change is outside the scope of this proposal. None of the flags/arguments that `purs` currently supports can "grow" as large as the source globs. So, there's no reason to make such a change.

Moreover, passing source globs to `purs` as it is done now doesn't need to change or be removed. It still works on *nix shells, so there's no reason to interrupt such workflows with what would be a breaking change.

Instead, I propose an additive change. In addition to passing source globs as it is currently done, we also allow passing source globs via a text file. We would update the CLI to use the `--source-globs FILE` argument to indicate which file contains the source globs. When this command is used, any globs passed in through the normal method produces an non-zero exit with an error explaining why.

Since the globs are just a list of text values, the file is a text file. Other formats are not used (e.g. JSON) because we don't need the additional structure.

The text file shall follow these requirements:
- When reading the file, `purs` will split the text into lines, regardless of which OS-specific line separator was used when writing the file.
- Each line consists of one of three values:
    - an empty line (ignored)
    - a line starting with `#` (a comment; ignored)
    - a single source glob (default)
- Each source glob must specify a path that is relative to the directory from which `purs` is called.

The usage of `--source-globs FILE` may not be the best name to use when we make the compiler package aware (i.e. #3493). But since I believe that would be a breaking change anyways, I don't think it would matter in the mean time.

## Examples

```sh
# This still works
purs compile src/**/*.purs .spago/packages/prelude-6.0.1/src/**/*.purs

# But this now also works
purs compile --source-globs .spago/source-globs
```

where `.spago/source-globs`' file contains the following:

```
# This file is automatically generated by Spago. 
# Do not edit by hand!
src/**/*.purs
.spago/packages/prelude-6.0.1/src/**/*.purs

```

JordanMartinez:
**Description of the change**

Fixes #4525.

---

**Checklist:**


JordanMartinez:
`glob-test.sh` outputs the following:
```sh
$ stack exec bash glob-test.sh 
env info
0.15.14 [development build; commit: e25c476c08c2e134f5d369326060be2f0d9ef583 DIRTY]
z
├── Foo.purs
├── globsAll
├── globsNoFoo
└── src
    ├── Bar
    │   └── Baz.purs
    └── Bar.purs

2 directories, 5 files
---
module Foo where
foo = 1 :: Int
---
module Bar where
bar = 1 :: Int
---
module Bar.Baz where
baz = 1 :: Int
---
Result
z/out1
├── Bar.Baz
│   ├── externs.cbor
│   └── index.js
├── cache-db.json
├── Foo
│   ├── externs.cbor
│   └── index.js
└── package.json

2 directories, 6 files
z/out2
├── Bar
│   ├── externs.cbor
│   └── index.js
├── Bar.Baz
│   ├── externs.cbor
│   └── index.js
├── cache-db.json
├── Foo
│   ├── externs.cbor
│   └── index.js
└── package.json

3 directories, 8 files
z/out3
├── Bar
│   ├── externs.cbor
│   └── index.js
├── Bar.Baz
│   ├── externs.cbor
│   └── index.js
├── cache-db.json
├── Foo
│   ├── externs.cbor
│   └── index.js
└── package.json

3 directories, 8 files
z/out4
├── Bar
│   ├── externs.cbor
│   └── index.js
├── Bar.Baz
│   ├── externs.cbor
│   └── index.js
├── cache-db.json
├── Foo
│   ├── externs.cbor
│   └── index.js
└── package.json

3 directories, 8 files
SOURCE_GLOBS output different from EXPECTED
Expected: purs compile: No files found using pattern: z/test/**/*.purs
[1 of 2] Compiling Bar.Baz
[2 of 2] Compiling Foo
SOURCE_GLOBS: purs compile: No files found using pattern: z/test/**/*.purs
[1 of 3] Compiling Foo
[2 of 3] Compiling Bar
[3 of 3] Compiling Bar.Baz
```

I tried this script on `master` and the EXPECTED globs also exclude the `[2 of 3] Compiling Bar` line. I think there's a bug in the glob handling.

JordanMartinez:
The above issue was due to `z/src/**/*.purs` being expanded by the shell to `z/src/Bar/Baz.purs` before the compiler ever received the glob. Wrapping that in single quotes disables that behavior, such that the compiler receives the full glob correctly.

JordanMartinez:
This PR is ready for an initial review now.

However, I noticed an issue somewhat related to this PR with how we produce our globs. In #4480, we added the option to exclude files from the initial set of files produced by the globs. However, this change was only done in the `compile` command. So, other commands that use globs like `docs`, `graph`, `hierarchy`, and `repl` don't support the `exclude-files` flag.

Moreover, should the format of the glob file added here also include some code for the `excldue-files` globs, such that one could write something like the following where lines prefixed with `+` act like our current input globs and `-` act like the `exclude-files` arg?
```
+ src/**/*.purs
- src/Spec/**/*.purs
```

f-f:
> Moreover, should the format of the glob file added here also include some code for the excldue-files globs, such that one could write something like the following where lines prefixed with + act like our current input globs and - act like the exclude-files arg?

I think we should ship the feature as implemented, and we can issue a future patch if there's the need for this. I don't think the `exclude-files` feature is in high usage at the moment.

f-f (commented):
The patch is good, but what about the `glob-test.sh` file? We should either use it from a test suite, or remove it since it will rot if not machine-checked regularly.

JordanMartinez:
> The patch is good, but what about the `glob-test.sh` file? We should either use it from a test suite, or remove it since it will rot if not machine-checked regularly.

I've added it to our tests, but made it only run on Linux.

JordanMartinez:
> However, I noticed an issue somewhat related to this PR with how we produce our globs. In https://github.com/purescript/purescript/pull/4480, we added the option to exclude files from the initial set of files produced by the globs. However, this change was only done in the compile command. So, other commands that use globs like docs, graph, hierarchy, and repl don't support the exclude-files flag.

I've updated this PR to also add `--exclude-files` and `--source-globs` to more commands.

JordanMartinez:
Now that CI builds, can this get another review due to adding both `--source-globs` and `--exclude-files` to the other commands?

CHANGELOG.d/feature_glob-input-files.md
```diff
@@ -0,0 +1,34 @@
...
+  `--source-globs` is an optional argument. Mixing it with the normal source globs is fine.
+  Assuming `.spago/source-globs` contains `src/**/*.purs`, each command below will use
+  the same input globs:
+  ```sh
+  purs compile src/**/*.purs
+  purs compile --source-globs .spago/source-globs
+  purs compile --source-globs .spago/source-globs src/**/*.purs 
+  ```
+
+  In the command, `purs compile glob1 glob2 --source-globs globsFoundInFile --exclude-file glob3`,
```
f-f, line 31: Typo
```suggestion
  In the command, `purs compile glob1 glob2 --source-globs globsFoundInFile --exclude-files glob3`,
```

app/Command/Compile.hs
```diff
@@ -72,25 +76,19 @@ compile PSCMakeOptions{..} = do
...
 inputFile :: Opts.Parser FilePath
 inputFile = Opts.strArgument $
      Opts.metavar "FILE"
   <> Opts.help "The input .purs file(s)."
 
-excludedFiles :: Opts.Parser FilePath
-excludedFiles = Opts.strOption $
+globInputFile :: Opts.Parser (Maybe FilePath)
+globInputFile = Opts.optional $ Opts.strOption $
+     Opts.long "source-globs"
```
f-f, line 86: I think the `exclude-files` is a bit misnamed now that we have a feature that lets us specify a file to include globs (feels like it should be `exclude-globs`?) - anyhow I wonder if this new flag should be called `source-globs-file` instead, to note that it's a file containing globs, and not just globs.
```suggestion
     Opts.long "source-globs-file"
```

app/Command/Docs.hs
```diff
@@ -132,6 +146,19 @@ pscDocsOptions = PSCDocsOptions <$> format <*> output <*> compileOutputDir <*> m
...
+      Opts.long "source-globs"
+    <> Opts.metavar "FILE"
+    <> Opts.help "A file containing a line-separated list of .purs globs."
+
+  excludeFile :: Opts.Parser FilePath
+  excludeFile = Opts.strOption $
+      Opts.short 'x'
+    <> Opts.long "exclude-files"
+    <> Opts.help "Glob of .purs files to exclude from the supplied files."
+
```
f-f, line 161: These are theoretically separate flags for separate commands, but they really are the same flag and we're duplicating this code 5 times - feels like they should be shared?

app/Command/Compile.hs
JordanMartinez: Personally, I think that would be a good change to make, but it's breaking. We could do it in the `0.16.0` release.

JordanMartinez: Then again... by updating the meta var from 'FILE' to 'GLOB' I think it does make more sense. So, maybe this change isn't needed?

JordanMartinez:
Anything else? Or is this good to merge?

app/Command/Ide.hs
```diff
@@ -57,11 +58,13 @@ listenOnLocalhost port = do
...
-  , _serverOutputPath :: FilePath
-  , _serverPort       :: Network.PortNumber
-  , _serverLoglevel   :: IdeLogLevel
+  { _serverDirectory     :: Maybe FilePath
+  , _serverGlobs         :: [FilePath]
+  , _serverGlobsFromFile :: Maybe FilePath
+  , _serverGlobsExcluded :: [FilePath]
+  , _serverOutputPath    :: FilePath
+  , _serverPort          :: Network.PortNumber
+  , _serverLoglevel      :: IdeLogLevel
```
rhendric, line 67:
```suggestion
  { _serverDirectory  :: Maybe FilePath
  , _serverGlobs      :: [FilePath]
  , _serverGlobsFromFile :: Maybe FilePath
  , _serverGlobsExcluded :: [FilePath]
  , _serverOutputPath :: FilePath
  , _serverPort       :: Network.PortNumber
  , _serverLoglevel   :: IdeLogLevel
```

CHANGELOG.d/bug_add-exclude-file-to-more-commands.md
```diff
@@ -0,0 +1,5 @@
+* Add `--exclude-file` to more commands
```
rhendric, line 1: `bug_` isn't one of the sacred file name prefixes; you'll want `fix_`. (But maybe this is actually `feature_`?)

rhendric (approved):
Just a bunch of nits; merge once you've dealt with the ones you think are worth dealing with.

CHANGELOG.d/bug_add-exclude-file-to-more-commands.md
JordanMartinez: I think this is the second time I've done this: calling it `bug_` rather than `fix_`. :smile: 
I've renamed it to `feature_`.

CHANGELOG.d/feature_glob-input-files.md
```diff
@@ -0,0 +1,38 @@
...
+  the same input globs:
+  ```sh
+  purs compile src/**/*.purs
+  purs compile --source-globs .spago/source-globs
+  purs compile --source-globs .spago/source-globs src/**/*.purs 
+  ```
+
+  In the command...
+  ```
+  purs compile inputGlob1 inputGlob2 --source-globs-file inputGlobsFoundInFile --exclude-files excludeGlob1
```
f-f, line 33: (github suggestions break in the presence of ```, so I'll comment line by line)

This is a bit of a nit, but I would rephrase this slightly:
```suggestion
  purs compile inputGlob1 inputGlob2 --source-globs-file fileWithMoreGlobs --exclude-files excludeGlob1
```

CHANGELOG.d/feature_glob-input-files.md
```diff
@@ -0,0 +1,38 @@
...
+  ``````
```
f-f, line 34: typo here?

There should be only three `

f-f (commented):
I left one nit in the changelog, but otherwise it's good to go - thanks @JordanMartinez!

JordanMartinez:
> I left one nit in the changelog, but otherwise it's good to go.

Fixed!

Issue #4020, opened 2021-03-05
Proposal: Construction and pattern matching of large union types is very slow


gorbak25:
## Summary
Might be a necrobump of https://github.com/purescript/purescript/issues/2577
Currently pattern matching on union types discriminates the types in the union using instanceOf. This incurs a significant performance penalty for larger union types and makes it infeasible to dispatch the code based on the type(switch on a tag instead of an if(... instanceof ...) ladder. Additionally construction of large data structures is slow due to the need to constantly allocate new heap objects(instead of being JS literal). JIT is affected as accessing a property/walking the prototype chain is not cheap.
Consider replacing data structures with an array ["TAG", value0, value1, ..., valueN]

## Motivation

After running ``` sed -i 's/instanceof/.constructor ===/g' dist/bundle.js ``` on https://github.com/erlscripten/erlps-aesophia I've decreased the runtime by 6%. This makes sense as erlscripten is based on the following union type - https://github.com/erlscripten/erlps-core/blob/c881e39901a2d6c29972b8a13c3c1ee760754560/src/Erlang/Type.purs#L42
I've started benchmarking various alternatives to instanceof and came to the conclusion that by just keeping everything in an array we generate fast JS which also can be easily optimized by the JIT compiler.
For
```
data Tree = Node Tree Tree | Leaf Int

foo :: Tree -> Int -> Int
foo (Node l r) acc = foo r (foo l acc)
foo (Leaf x) acc = acc + x

main :: Int
main = foo (Node (Node (Node (Leaf 1) (Leaf 2)) (Node (Leaf 3) (Leaf 4))) (Node (Node (Leaf 5) (Leaf 6)) (Node (Leaf 7) (Leaf 8)))) 0
```
Keeping everything in an array yields an 95%+ performance improvement https://jsbench.me/beklw3j28m/2
I've also noticed that construction of large terms in erlscripten is suspiciously slow(due to the usage of "new")

## Proposal
Replace ```new Foo(a1, a2, a3)``` with ``` ['Foo', a1, a2, a3] ```
Replace ```v instanceof Foo``` with ```v[0] === 'Foo'```
Replace ```v.valueN``` with ```v[N+1]```

## Examples
Take a look at the benchmark: https://jsbench.me/beklw3j28m/2

## Small question
In the Erlscripten project we plan to maintain and ship a custom superoptimizing Purescript compiler.
[...]

PR #4540, opened 2024-03-19
Use JS Symbol for sum ADTs

shybovycha:
**Description of the change**

Clearly and concisely describe the purpose of the pull request. If this PR relates to an existing issue or change proposal, please link to it. Include any other background context that would help reviewers understand the motivation for this PR.

Currently PureScript generates a quite outdated, lengthy javascript code for sum ADTs:

```hs
data Msg = Increment | Decrement
```

produces the following JS in `0.15.15`:

```js
// Generated by purs version 0.15.15
var Increment = /* #__PURE__ */ (function () {
    function Increment() {

    };
    Increment.value = new Increment();
    return Increment;
})();
var Decrement = /* #__PURE__ */ (function () {
    function Decrement() {

    };
    Decrement.value = new Decrement();
    return Decrement;
})();
```

When comparing the above ADTs (with pattern matching or `if` expressions), the code generated is as follows:

```hs
main :: Msg -> String
main Increment = "42"
main Decrement = "0"
```

produces in `0.15.15`:

```js
var main = function (v) {
    if (v instanceof Increment) {
        return "42";
    };
    if (v instanceof Decrement) {
        return "0";
    };
    throw new Error("Failed pattern match at Main (line 9, column 1 - line 9, column 22): " + [ v.constructor.name ]);
};
```

A better approach would be to use [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) class, available in all the modern browsers and Node.JS versions:

```js
// Generated by purs version 0.15.16
var Increment = /* #__PURE__ */ Symbol("Increment");
var Decrement = /* #__PURE__ */ Symbol("Decrement");
```

so that then the pattern matching can make use of equality (which is stricter than `instanceof`):

```js
var main = function (v) {
    if (v === Increment) {
        return "42";
    };
    if (v === Decrement) {
        return "0";
    };
    throw new Error("Failed pattern match at Main (line 9, column 1 - line 9, column 22): " + [ v.constructor.name ]);
};
```

This PR introduces two changes:

1. generating `Symbol` for sum ADTs
2. comparing the corresponding symbols for pattern matching sum ADTs using equality operator rather than `instanceof`

This should make the generated JS code more reliable when comparing values and shorter (resulting in smaller bundles).

For parametrized types this would still work as before, generating the old-style code:

```hs
data Tree a = Leaf a | Empty

main :: forall a. Tree a -> String
main (Leaf _) = "leaf"
main Empty = "empty"
```

and the JS output:

```js
var Leaf = /* #__PURE__ */ (function () {
    function Leaf(value0) {
        this.value0 = value0;
    };
    Leaf.create = function (value0) {
        return new Leaf(value0);
    };
    return Leaf;
})();
var Empty = /* #__PURE__ */ (function () {
    function Empty() {

    };
    Empty.value = new Empty();
    return Empty;
})();
var main = function (v) {
    if (v instanceof Leaf) {
        return "leaf";
    };
    if (v instanceof Empty) {
        return "empty";
    };
    throw new Error("Failed pattern match at Main (line 9, column 1 - line 9, column 35): " + [ v.constructor.name ]);
};
```

---

**Checklist:**

- [ ] ~~Linked any existing issues or proposals that this pull request should close~~
- [ ] ~~Updated or added relevant documentation~~
- [ ] ~~Added a test for the contribution (if applicable)~~

### DIFF
--- a/CHANGELOG.d/feature_use_symbols_for_sum_adts.md
+++ b/CHANGELOG.d/feature_use_symbols_for_sum_adts.md
@@ -1,0 +1,1 @@
+* Generate [`Symbol()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) objects for sum ADTs and use equality (`===`) instead of `instanceof` for comparisons on sum ADTs
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -145,6 +145,7 @@ If you would prefer to use different terms, please use the section below instead
 | [@senju](https://github.com/senju) | senju | [MIT license] |
 | [@seungha-kim](https://github.com/seungha-kim) | Seungha Kim | [MIT license] |
 | [@sharkdp](https://github.com/sharkdp) | David Peter | [MIT license] |
+| [@shybovycha](https://github.com/shybovycha) | Artem Shubovych | [MIT license] |
 | [@sigma-andex](https://github.com/sigma-andex) | Jan Schulte | [MIT license] |
 | [@simonyangme](https://github.com/simonyangme) | Simon Yang | [MIT license] |
 | [@sloosch](https://github.com/sloosch) | Simon Looschen | [MIT license] |
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -71,3 +71,19 @@ Error: EACCES: permission denied
 ```
 
 The best solution is to install [Node.js and npm via a node version manager](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#using-a-node-version-manager-to-install-nodejs-and-npm). This error is due to permissions issues when installing packages globally. You can read more about this error in npm's guide to [resolving EACCES permissions errors when installing packages globally](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
+
+## hGetContents invalid argument error
+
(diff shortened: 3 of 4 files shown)


natefaubion:
As I understand it, this PR changes enumeration-like constructors (no arguments) to Symbols. See discussion in https://github.com/purescript/purescript/issues/4020 for the more general issue of data type encoding. The optimized representation is currently implemented in https://github.com/aristanetworks/purescript-backend-optimizer with the additional options of supporting `--int-tag` and unwrapped constant values for enumerations. The advantage of non-symbols being that you can potentially transfer them between JS realms or construct them without having access to the original Symbol.

JordanMartinez:
Should this PR be closed since we've not yet reached a conclusion (AFAIK) in #4020? Also, since `backend-optimizer` already does such an optimization, is this still needed?

shybovycha:
Thanks for the pointers! Seems like it is a quite long-running debate.

My comment was getting a tad too long and generic (relative to this PR), so I will rather add it on #4020 .

TL;DR:

* currently the generated code is not really good for sharing with the outside JS either
* having `int` or `string` tag in an object has the exact same issues as the current approach
* why not combine tag with `Symbol` and toggle it with a compiler flag (when on - generate more safe code, when off - generate smaller code)?

Issue #4545, opened 2024-07-19
On large projects the IDE uses excessive memory


roryc89:
## Description

On our project of 9877 source files (including libs), the IDE quickly consumes so much memory that it becomes hard to work with.

On my 32GB Apple M2, it will be using over 20GB of memory in under a minute. It then starts to slow down until common commands, like looking up a type definition can take over 30 seconds.

## To Reproduce

This is not simple as our project is not open source but I believe it would be likely to happen on any sufficiently large project.  

## Expected behavior

The IDE should limit it's memory usage and commands should execute within a few seconds. 

## Additional context

From tinkering with the IDE, I have found that the forcing of the result in `populateVolatileStateSTM` and possibly `populateVolatileStateSync` as well, seems to responsible. I will make a draft PR to explain in more detail. 

## PureScript version

0.15.15

PR #4546, opened 2024-07-19
dont force state results #4545

roryc89:
**Description of the change**

To address https://github.com/purescript/purescript/issues/4545. 

This change fixes the memory issues in our project and makes the IDE reasonable performant but I don't know if there are other ramifications.

---

**Checklist:**

- [x] Updated or added relevant documentation (not applicable)
- [x] Added a test for the contribution (not applicable)


src/Language/PureScript/Ide/State.hs
```diff
@@ -235,7 +234,7 @@ populateVolatileStateSTM ref = do
         & resolveOperators
         & resolveReexports reexportRefs
   setVolatileStateSTM ref (IdeVolatileState (AstData asts) (map reResolved results) rebuildCache)
-  pure (force results)
+  pure results
```
roryc89, line 237: I have limited haskell XP so may be misunderstanding but I suspect that by allowing results to be lazy, the IDE doesn't have to calculate all of the results as the state may be recomputed before much or any of an old result is used. 

Whereas, when it is forced, it is possible for many computations to be queued up on a large project and the `results` value can use up so much memory that the IDE starts to grind to a halt.

f-f:
I think the PR is sound, CI currently fails presumably because of the Azure debacle

f-f:
We'll need #4548 merged before this can pass

roryc89:
As #4548 is merged, is this ok to merge now?

f-f (approved):
This looks good to me - could someone else take a look as well?

rhendric:
This patch appears to be reverting https://github.com/purescript/purescript/pull/3006/commits/26067277d96014fc7ecaccd4114e0b34d30d2527 from #3006, which means at least three things:
* If this revert is desirable, there may be a bunch of `Generic` and `NFData` instances that can also be cleaned up.
* Per that commit message, a consequence of this is that a `logPerf` measurement is no longer accurate, and we should figure out what to do about that.
* Also per that commit message, something about ‘actually populate the cache of f-thread’ may need attention (I don't know what that means).

f-f:
@kritzcreek I know that commit was 7y ago, but do you have any recollection about this?

kritzcreek:
Oh my, I had to think for a bit...

> ‘actually populate the cache of f-thread’

Actually meant `off-thread`, which is about a [worker thread](https://github.com/purescript/purescript/blob/e06b9ccb7cbf31633d25e55531d70dcda7ec28b2/src/Language/PureScript/Ide/State.hs#L216) that does the `populateVolatileState` logic in here in the background.  I think at the time when measuring I figured out that leaving the value lazy meant the computation ended up being done on the thread that accepts and responds to requests. Which meant the first requests after a state reload would take a very long time. (That's also why most of the resolution logic is wrapped in STM)

I'm not sure if making things lazy here is going to help with peak memory usage, it might just delay it until you start sending some requests against the updated IdeState. Sorry, it's been a long time so I'm very fuzzy on the details...

purefunctor:
fwiw, this change brings down what normally is 30 GBs of peak memory usage down to 10 GBs alongside bringing down memory pressure/time-to-diagnostics quite a bit on our codebase. We also use this with a custom IDE command that filters which externs files get loaded.

One side effect is that diagnostics are not immediately available until the first edit, but I'm not entirely sure if that's coming from this PR or the custom IDE command.

This LGTM aside from removing the redundant `Generic`/`NFData` instances.

kritzcreek:
Please don't take my comment as a blocker to merging. I haven't written in Haskell in over 5(?) years, so I'm hardly an authority on the matter :D Just wanted to dump what context I could remember in case it was helpful.

nwolverson:
> One side effect is that diagnostics are not immediately available until the first edit, but I'm not entirely sure if that's coming from this PR or the custom IDE command.

I assume by diagnostics you mean type info etc. It would certainly seem worth checking this with a clean build because that sounds like a blocker to me (and presumably not at all necessary or deep)

i-am-the-slime:
This looks very easy to revert. If it was merged, it'd get released as a beta, right? 
That could be a good way to gather feedback without having to dive extremely deep.

roryc89:
> * If this revert is desirable, there may be a bunch of `Generic` and `NFData` instances that can also be cleaned up.

Good point. I've removed them. 

> * Per that commit message, a consequence of this is that a `logPerf` measurement is no longer accurate, and we should figure out what to do about that.

Also removed. I guess it would also be possible to only call `force` when `logLevel == DEBUG`

rhendric (approved):

rhendric:
This looks ready to merge to me; any final objections?

garyb (approved):

PR #4549, opened 2024-08-01
load modules concurrently #4545

roryc89:
**Description of the change**

On large projects loading the modules concurrently lowers memory usage substantially. Should help address #4545

**Checklist:**

Relevant checklist points addressed in https://github.com/purescript/purescript/pull/4546

### DIFF
--- a/src/Language/PureScript/Ide/SourceFile.hs
+++ b/src/Language/PureScript/Ide/SourceFile.hs
@@ -29,6 +29,8 @@ import Language.PureScript.CST qualified as CST
 import Language.PureScript.Ide.Error (IdeError)
 import Language.PureScript.Ide.Types (DefinitionSites, IdeNamespace(..), IdeNamespaced(..), TypeAnnotations)
 import Language.PureScript.Ide.Util (ideReadFile)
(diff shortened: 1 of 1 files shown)


f-f (approved):

cakekindel:
@f-f anything blocking merge here? this change looks lovely

f-f:
We'll need another review from the core team - could anyone take a look?

rhendric (commented):
Nits aside, I'm surprised that this is reported to decrease memory use because my understanding is that `mapConcurrently` will spawn every one of these tasks at once. Has that been confirmed on very large projects? What about consumption of other resources—open file handles or something?

I'd be more comfortable with this if we used primitives like [`pooledMapConcurrently`](https://hackage.haskell.org/package/unliftio-0.2.25.0/docs/UnliftIO-Async.html#v:pooledMapConcurrently) instead. OTOH, I don't do much maintenance of the `Ide` modules so if the team members who do are happy with this I won't block it.

rhendric:
Also, I think this does merit its own changelog entry.

PR #4561, opened 2024-12-29
Fix typo in CoreFn traversal function

drathier:
This seems to be a critical unintentional typo

### DIFF
--- a/src/Language/PureScript/CoreFn/Traversals.hs
+++ b/src/Language/PureScript/CoreFn/Traversals.hs
@@ -70,7 +70,7 @@ traverseCoreFn f g h i = (f', g', h', i')
   g' (Abs ann name e) = Abs ann name <$> g e
   g' (App ann v1 v2) = App ann <$> g v1 <*> g v2
   g' (Case ann vs alts) = Case ann <$> traverse g vs <*> traverse i alts
-  g' (Let ann ds e) = Let ann <$> traverse f ds <*> g' e
+  g' (Let ann ds e) = Let ann <$> traverse f ds <*> g e
   g' e = pure e
 
   h' (LiteralBinder a b) = LiteralBinder a <$> handleLiteral h b


rhendric:
Good catch! Please add an `internal_` file to `CHANGELOG.d` per the `README.md` contained therein, unless this is connected to a bug you've observed, in which case make it `fix_` and describe the bug.

garyb:
Is this change is correct? As, per the comment above the function:

>  This function is useful as a building block for recursive functions, but doesn't actually recurse itself.

Note also that the internal aren't used recursively in any instance either. But this is your function @rhendric so I guess you would know better. :smile:

rhendric:
I think this is correct, but it's been a while since I wrote it.

What the doc comment says to me is that if you want to make an actually recursive traversal, you do that by calling the results of `traverseCoreFn` from within the arguments to `traverseCoreFn`. `traverseCoreFn` shouldn't be tying its own knot.

I think the typo ended up having no real effect, because the one use of `traverseCoreFn` is [here](https://github.com/purescript/purescript/blob/fc3fa8897916de1a3973de976eaea1fba23b4df9/src/Language/PureScript/CoreFn/Laziness.hs#L126), and the function that is named `g'` inside `traverseCoreFn` and named `handleExpr` at that use site is only ever invoked with arguments that are not `Let`, because of how clauses in a `\case` expression are ordered. So this is unused code; but if it were to be used, I think the thing that we should expect it to do is call `g` and not `g'`, as the other branches do.

drathier:
I found this by trying to tie the knot outside this function, and realising the resulting knotted function doesn't visit all ast nodes. I'd have to handle some of the ast cases myself, which defeats the purpose of this function imo

drathier:
I'll do the requested changes to changelog etc soon, probably next year :)

f-f:
Looks like CI needs some love, we'll need to upgrade some actions 

Also the ARM runners are not responsive so I'll have a look at what's up with those

f-f (approved):

PR #4566, opened 2025-02-20
Upgrade CI

f-f:
Some CI upgrades required to get back to all green:
- upgrade deprecated `upload-artifact` action to a supported version
- use spago-next for testing the package set to avoid zlib errors from the legacy spago
- upgrade Node version to v22, which is the latest supported LTS


f-f:
Replaced by #4568

Issue #4567, opened 2025-03-25
Doesn't build on macOS arm64


ad-si:
## To Reproduce

```text
$ stack update
$ git clone https://github.com/purescript/purescript
$ cd purescript
$ make build
STACK_YAML="stack.yaml" stack build purescript
Getting the project-level configuration file from the STACK_YAML environment variable.
Linking /Users/adrian/dotfiles/home/stack/setup-exe-cache/aarch64-osx/tmp-Cabal-simple_w2MFVN35_3.6.3.0_ghc-9.2.8 ...
ld64.lld: error: unknown argument '--gc-sections'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
`gcc' failed in phase `Linker'. (Exit code: 1)
Getting the project-level configuration file from the STACK_YAML environment variable.

Error: [S-6374]
       While building simple Setup.hs (scroll up to its section to see the error) using:
       /Users/adrian/dotfiles/home/stack/programs/aarch64-osx/ghc-9.2.8/bin/ghc-9.2.8 -rtsopts -threaded -clear-package-db -global-package-db -hide-all-packages -package base -main-is StackSetupShim.mainOverride -package Cabal-3.6.3.0 /Users/adrian/dotfiles/home/stack/setup-exe-src/setup-w2MFVN35.hs /Users/adrian/dotfiles/home/stack/setup-exe-src/setup-shim-w2MFVN35.hs -o /Users/adrian/dotfiles/home/stack/setup-exe-cache/aarch64-osx/tmp-Cabal-simple_w2MFVN35_3.6.3.0_ghc-9.2.8
       Process exited with code: ExitFailure 1
make: *** [build] Error 1
```

## Expected behavior

Should build the compiler.

## PureScript version

Latest commit on master

ad-si:
If I upgrade the LTS version to a newer one it seems to work intitially. (However, not the full build as dependency problems crop up then).
So seems to be a problem with GHC 9.2.8.

PR #4568, opened 2025-03-26
Upgrade to GHC 9.6.6

ad-si:
- Switch from `ansi-wl-pprint` to `prettyprinter`
- Add several `extra-deps`
- Fixes https://github.com/purescript/purescript/issues/4567

---

This is a first starting point as a few things still need to be clarified:

I added 3 GitHub dependencies with fixes for the respective libraries. 

- [Vekhir/aeson-better-errors] for [hdgarrood/aeson-better-errors]
- [MercuryTechnologies/cheapskate] for [jgm/cheapskate]
    (This is deprecated and should be replaced with [jgm/commonmark-hs])
- [GuillaumedeVolpiano/pattern-arrows] for [paf31/pattern-arrows]

[GuillaumedeVolpiano/pattern-arrows]: https://github.com/GuillaumedeVolpiano/pattern-arrows
[hdgarrood/aeson-better-errors]: https://github.com/hdgarrood/aeson-better-errors
[jgm/cheapskate]: https://github.com/jgm/cheapskate
[jgm/commonmark-hs]: https://github.com/jgm/commonmark-hs
[MercuryTechnologies/cheapskate]: https://github.com/MercuryTechnologies/cheapskate
[paf31/pattern-arrows]: https://github.com/paf31/pattern-arrows
[Vekhir/aeson-better-errors]: https://github.com/Vekhir/aeson-better-errors

There are several options how to deal with it:

1. Use it like this
1. Fork the repositories to purescript/… and use from there
1. Upstream changes to Hackage packages first
    (Could be hard since the repos don't seem to be actively maintained anymore)
1. Fork and republish under a new Hackage package name

I would actually just use it like this. The version is exactly pinned through the commit hash, and we're dependent on GitHub in many ways already anyways.
What do you think?

### TODOs:

- [x] pattern-arrows - Integrate this into the source code as it is only one file
- [x] cheapskate - Use MercuryTechnologies/cheapskate until we replace it with commonmark-hs
- [x] aeson-better-errors - Release a new version on Hackage
- [ ] Maybe upgrade even to GHC 9.6.7 if available soon (**Update:** In a follow up PR)


garyb:
Makes sense to me to get things upgraded, but I'm so out of the loop with Haskell build stuff these days I'm not really in a good place to review or give much meaningful feedback on it.

I'd perhaps be a little more tempted to vendor the fixes in though, as we're not subject to them disappearing if deleted from GH then, etc. Chances are they'll need further updates to continue building into the future too, and these current forks with fixes may also be dormant by then.

ad-si:
> I'd perhaps be a little more tempted to vendor the fixes in though, as we're not subject to them disappearing if deleted from GH then, etc. Chances are they'll need further updates to continue building into the future too, and these current forks with fixes may also be dormant by then.

Yeah, good point, but do you mean vendoring them into this repo?
I think forking them to the `purescript` organization should be good enough, but on the other hand the dependencies aren't a lot of code. So might actually make sense to just add them to the repo. What do you think?

ad-si:
> I reckon forking would be better if we're expecting changes from/to upstream, though vendoring is equally valid since these packages are pretty simple enough

My suggestion:
- aeson-better-errors - Since you're member of the repo now, we simply release a new version on Hackage
- cheapskate - We fork MercuryTechnologies/cheapskate to purescript/cheapskate and use this until we replace it with commonmark-hs
- pattern-arrows - Integrate this into the source code as it is only one file

f-f:
Thanks for picking this up @ad-si! I started this work and then got discouraged by the set of constraints that we have been following in order to ship the most compatible binaries and keep a good dev experience.

So far I gathered:
- the Haskell version should be compatible with the language-server, which is fair enough. According to [this page](https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html#current-ghc-version-support-status) the version targeted here - `9.6.6` is not supported anymore, but `9.6.7` (the latest in that line) is, so I think we should move to that version if possible
- the `libc` version should be as old as possible. There [has been](https://github.com/purescript/purescript/issues/4191) quite [some work](https://github.com/purescript/purescript/pull/4228) over the years to have the Linux build inside a docker image that was based on an ancient Debian release just so that we could have a libc that was old enough to not have issues on the average Linux deployment. That Debian version is now EoL, which is what prompted me to start the upgrade. However the container we were using doesn't seem to be moving, so I guess we can get back to using Ubuntu in CI? The current `glibc` version is `2.28`, and the one in this PR would be the one shipped by Ubuntu 24, which is `2.39`. I think that's way too recent - according to [GitHub docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) they do still support Ubuntu 20.04, which ships with `2.31` and that should be good enough?

So to recap, let's try:
- Haskell 9.6.7
- Ubuntu 20.04 as base image

f-f:
A few more things:
- about vendoring dependencies: I'm happy to vendor `pattern-arrows`. It looks like we can follow upstream for `aeson-better-errors`, which is great, and I would not fork `cheapskate` - Mercury is a well-known org and we should be switching to commonmark anyways, so I'd just follow their upstream for now
- about CI: we have ARM Linux and ARM macos runners sitting at the office, but they got stuck once more so I'll see what's wrong with them in the morning

ad-si:
> `9.6.6` is not supported anymore, but `9.6.7` (the latest in that line) is, so I think we should move to that version if possible

This changed just a few days ago and GHCup doesn't even list it yet, but it's a good point. I'll change it!

> * the `libc` version should be as old as possible.

Ok sure! Maybe we should even extend the matrix to contain 20.04, 22.04, and 24.04?

.github/workflows/ci.yml
```diff
@@ -54,15 +55,15 @@ jobs:
...
           - # If upgrading the Haskell image, also upgrade it in the lint job below
-            os: ["ubuntu-latest"]
-            image: haskell:9.2.8@sha256:b3b2f3909c7381bb96b8f18766f9407a3d6f61e0f07ea95e812583ac4f442cbb
-          - os: ["macOS-14"]
-          - os: ["windows-2019"]
-          - os: ["self-hosted", "macos", "ARM64"]
-          - os: ["self-hosted", "Linux", "ARM64"]
+            os: [ubuntu-24.04]
+            image: haskell:9.6.6
+          - os: [macos-15]
```
f-f, line 60: Do we need to use macos 15? We should try to stay on older versions if possible, for library compatibility.

According to [GitHub docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) we should use both `macos-13` for x86, and `macos-14` for ARM builds.

f-f:
> Maybe we should even extend the matrix to contain 20.04, 22.04, and 24.04?

One Linux should be enough 🙂 

> Mh, might take a while with 9.6.7

No worries, these kind of changes take some time to propagate

.github/workflows/ci.yml
```diff
@@ -54,15 +55,15 @@ jobs:
...
         include:
           - # If upgrading the Haskell image, also upgrade it in the lint job below
-            os: ["ubuntu-latest"]
-            image: haskell:9.2.8@sha256:b3b2f3909c7381bb96b8f18766f9407a3d6f61e0f07ea95e812583ac4f442cbb
-          - os: ["macOS-14"]
-          - os: ["windows-2019"]
-          - os: ["self-hosted", "macos", "ARM64"]
-          - os: ["self-hosted", "Linux", "ARM64"]
+            os: [ubuntu-24.04]
+            image: haskell:9.6.6
```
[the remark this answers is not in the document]
ad-si: Ah, that's why it was `ubuntu-latest`: It doesn't really matter what the host system is if you're running it in a container anyways.
I thinks that's a good way to do it. Will revert it to `-latest` and make it clearer that it's about the container and not the os.

I just checked the `haskell:9.6.6` image and it ships with `GLIBC 2.31-13+deb11u11`.

ad-si:
@f-f the latest commit builds correctly in my fork. Were you able to figure out what's wrong with the self-hosted runners? Also, are they even still needed? Seems like GitHub has quite a few options for different runners by now.

Regarding the GHC version: Do we want to wait until 9.6.7 is available via Stackage or merge it with 9.6.6 already and upgrade it in another PR. I'd prefer if we already merge it, so I can create some follow up PRs with the actual changes I wanted to make. 😅

@purefunctor said he can release the new `aeson-better-errors` version on the weekend.

f-f:
> Were you able to figure out what's wrong with the self-hosted runners? 

Yes, they are going again 🙂 

> Also, are they even still needed? Seems like GitHub has quite a few options for different runners by now.

Good question - it looks like GitHub runners can do Linux ARM64 as well these days, which is great and it means we could replace the self-hosted things with that.

I'd be happy to have them replaced with GitHub-hosted ones (and arguably we are already doing that given that we're running `macos-14`)

ad-si:
The arm64 Linux runner seems to work well: https://github.com/ad-si/purescript/actions/runs/14227397693

@f-f Let me remove the self-hosted runners then, ok?

f-f:
@ad-si ah yeah these builds look good. We're good to remove the self hosted runners!

ad-si:
> Could we use `ubuntu-22` for ARM? For the usual concerns about old `glibc`.

But Ubuntu is only the host image, and it's actually build in the haskell:9.6.6 image, right?

f-f:
> But Ubuntu is only the host image, and it's actually build in the haskell:9.6.6 image, right?

Oh sorry, of course 😄

f-f:
@ad-si one last thing: could you add to the changelog the fact that we are bumping the minimum required glibc from `2.28` to `2.31`?

f-f (approved):
Looks great now - thanks @ad-si!

ad-si:
🙌 … just waiting for @purefunctor ctor to release the new `aeson-better-erros` version and we're ready to go 🚀

ad-si:
Just pushed the update for `aeson-better-erros` 🚀
If there are no more concerns, I think we're ready to merge, right?

(The HLS for 9.6.7 is still not available via GHCup. So I would do that in a follow up PR.)

purefunctor (commented):
tiny comments

stack.yaml
```diff
@@ -37,8 +36,13 @@ nix:
...
   haskeline:
     # Avoids a libtinfo dynamic library dependency
     terminfo: false
+
+allow-newer: true
+allow-newer-deps:
+  - aeson-better-errors
+  - cheapskate
+  - haskeline
+  - process
```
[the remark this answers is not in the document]
ad-si: Good catch! I removed it now.

purescript.cabal
```diff
@@ -179,36 +178,36 @@ common defaults
     Glob >=0.10.2 && <0.11,
     haskeline >=0.8.2 && <0.9,
```
purefunctor, line 179:
```suggestion
    haskeline >=0.8.2.1 && <0.9,
```

stack.yaml
```diff
@@ -13,20 +13,14 @@ extra-deps:
...
-# ghc depends on terminfo by default, but that can be ignored
-# if one uses the '-terminfo' flag.
-# Unfortunately, hspec doesn't expose a similar flag.
-#
-# Using hspec >= 2.10.7 addresses this.
-- hspec-2.10.9
-- hspec-core-2.10.9
-- hspec-discover-2.10.9
+- bower-json-1.1.0.0
+- haskeline-0.8.2
```
purefunctor, line 17:
```suggestion
- haskeline-0.8.2.1
```

purefunctor (approved):
Once tests go green I think this should be good to merge. One caveat here is that Cabal builds are currently not working because it's pinned to haskeline-0.8.2, which avoids libtinfo from being added to the build.

#4573 should hopefully slay the libtinfo dragon once and forall in the near future.

PR #4574, opened 2025-04-15
Upgrade to GHC 9.8.4

ad-si:
When I saw that [HLS 2.10 was released](https://blog.haskell.org/hls-2-10-0-0/) (which is finally compatible with GHC 9.8 🙌), I couldn't help myself to try to upgrade PureScript to GHC 9.8.4 and luckily it caused less problems than the upgrade to 9.6! 😄

---

**Checklist:**

### DIFF
--- a/src/Language/PureScript/Interactive/Directive.hs
+++ b/src/Language/PureScript/Interactive/Directive.hs
@@ -4,6 +4,7 @@
 module Language.PureScript.Interactive.Directive where
 
 import Prelude
+import Protolude (headDef)
 
 import Data.Maybe (fromJust)
 import Data.List (isPrefixOf)
@@ -48,7 +49,7 @@ stringsFor d = fromJust (lookup d directiveStrings)
 -- Returns the default string representation of a directive.
 --
 stringFor :: Directive -> String
-stringFor = head . stringsFor
+stringFor = headDef "stringFor: no string found" . stringsFor
 
 -- |
 -- Returns the list of directives which could be expanded from the string
--- a/src/Language/PureScript/Interactive/Directive.hs
+++ b/src/Language/PureScript/Interactive/Directive.hs
@@ -36,10 +36,9 @@ directiveStrings =
 -- Like `directiveStrings`, but the other way around.
 --
 directiveStrings' :: [(String, Directive)]
-directiveStrings' =
-  concatMap
-    (\(dir, strs) -> map (, dir) $ NEL.toList strs)
-    directiveStrings
+directiveStrings' = concatMap go directiveStrings
+  where
+  go (dir, strs) = map (, dir) $ NEL.toList strs
 
 -- |
 -- Returns all possible string representations of a directive.
@@ -66,7 +65,7 @@ directivesFor :: String -> [Directive]
 directivesFor = map fst . directivesFor'
 
 directiveStringsFor :: String -> [String]
-directiveStringsFor str = map snd (directivesFor' str)
+directiveStringsFor = map snd . directivesFor'
 
 -- |
 -- The help menu.
--- a/src/Language/PureScript/Interactive/Directive.hs
+++ b/src/Language/PureScript/Interactive/Directive.hs
@@ -4,11 +4,12 @@
 module Language.PureScript.Interactive.Directive where
 
 import Prelude
-import Protolude (headDef)
 
 import Data.Maybe (fromJust)
 import Data.List (isPrefixOf)
 import Data.Tuple (swap)
+import Data.List.NonEmpty (NonEmpty(..))
+import Data.List.NonEmpty qualified as NEL
 
 import Language.PureScript.Interactive.Types (Directive(..))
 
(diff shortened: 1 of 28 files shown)


rhendric:
If this gets merged soon after the previous one so that there aren't too many intermediate commits, then okay I guess; but either way, please let's not have any more GHC upgrades for a year at least. It really throws off the `git rebase` workflow when too many GHCs are needed to track down a bug.

ad-si:
> let's not have any more GHC upgrades for a year

Agreed! I would have upgraded to 9.8 immediately, but HLS was broken there for quite some time and it wasn't clear if they were gonna fix it or just move on to 9.10. The good thing about 9.8 is that it's the current Stackage LTS version and the most up-to-date GHC that is not expected to get any further releases (https://endoflife.date/ghc). So this should be stable and serve us well for quite a while 👍

ad-si:
I think this PR is ready for review @rhendric and @purefunctor!

rhendric (changes requested):
By and large looks good; just a couple of questions and a few nits.

I'd love to see most of these `internalError`s go away via changing lists to `NonEmpty`s, but I'm happy to accept that as tech debt so we can land this sooner.

src/Language/PureScript/TypeChecker/Deriving.hs
```diff
@@ -1,6 +1,7 @@
 {- HLINT ignore "Unused LANGUAGE pragma" -} -- HLint doesn't recognize that TypeApplications is used in a pattern
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeAbstractions #-}
```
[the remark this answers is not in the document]
ad-si: Yes, it's necessary!

ad-si:
Thanks for your feedback @rhendric! I applied all your suggestions and it should be good to go now.

src/Language/PureScript/Interactive/Directive.hs
```diff
@@ -8,62 +8,63 @@ import Prelude
...
 -- |
 -- Returns the list of directives which could be expanded from the string
 -- argument, together with the string alias that matched.
 --
-directivesFor' :: String -> [(Directive, String)]
+directivesFor' :: String -> [(Directive, NonEmpty String)]
 directivesFor' str = go directiveStrings'
   where
-  go = map swap . filter ((str `isPrefixOf`) . fst)
+  go = map swap . filter ((str `isPrefixOf`) . NEL.head . fst)
```
rhendric, line 60: This change looks like it introduces a bug. Previously, `directivesFor' "h"` would have evaluated to `[(Help, "help")]`. Now, I believe it would be the empty list.

I think `directiveStrings'` should go back to its original type, and `directivesFor'` should remain unchanged. The only intervention needed in `directiveStrings'` should be adding a `toList` in `go`, I think.

ad-si: Great catch! I'm just surprised that there were no tests for this. I fixed it now and added a few basic tests for the `Directive` module.

src/Language/PureScript/Interactive/Directive.hs
```diff
@@ -63,7 +66,7 @@ directivesFor :: String -> [Directive]
 directivesFor = map fst . directivesFor'
 
 directiveStringsFor :: String -> [String]
-directiveStringsFor = map snd . directivesFor'
+directiveStringsFor str = map snd (directivesFor' str)
```
rhendric, line 69: Please revert this.

src/Language/PureScript/Interactive/Directive.hs
```diff
@@ -8,47 +8,50 @@ import Prelude
...
+-- Like `directiveStrings`, but the other way around.
 --
 directiveStrings' :: [(String, Directive)]
-directiveStrings' = concatMap go directiveStrings
-  where
-  go (dir, strs) = map (, dir) strs
+directiveStrings' =
+  concatMap
+    (\(dir, strs) -> map (, dir) $ NEL.toList strs)
+    directiveStrings
```
rhendric, line 42:
```suggestion
directiveStrings' = concatMap go directiveStrings
  where
  go (dir, strs) = map (, dir) $ NEL.toList strs
```

rhendric (commented):
Thank you for the new tests! If you would roll back these last two very detail-y things that got altered in the shuffle (just to keep the `git blame`s clean) you've got my approval.

src/Language/PureScript/Interactive/Directive.hs
ad-si: Good catch! Cleaned both of them up.

purefunctor (approved):

seanwestfall:
What about upgrading to GHC 9.10.2. That looks to be the current most stable release of GHC as of this May 2025.

rhendric:
https://github.com/purescript/purescript/pull/4574#issuecomment-2807436943

ad-si:
It’s working with HLS for me, but I first had to fully purge all dependency and build files so they are build correctly with the new version!

wclr:
> It’s working with HLS for me, but I first had to fully purge all dependency and build files so they are build correctly with the new version!

What do you mean by purging all dependencies? So far I tried to `stack clean`, removing `.stack-work`, then `stack build`.

purefunctor:
I first mentioned the Cabal breakage in https://github.com/purescript/purescript/pull/4568#pullrequestreview-2755830259 and if I recall correctly, using 0.8.2.1 causes the built binary to be built with `libtinfo` which makes pre-built binaries harder to distribute.

You can update the version bounds for now for local development or maybe use [hie.yaml](https://haskell-language-server.readthedocs.io/en/stable/configuration.html#configuring-your-project-build) to tell HLS to use `stack`

wclr:
> I first mentioned the Cabal breakage in https://github.com/purescript/purescript/pull/4568#pullrequestreview-2755830259 and if I recall correctly, using 0.8.2.1 causes the built binary to be built with libtinfo which makes pre-built binaries harder to distribute.

For clarification, why it it installs 0.8.2.1 if 0.8.2 is required and fixated? And how it is going to be fixed?

> You can update the version bounds for now for local development or maybe use [hie.yaml](https://haskell-language-server.readthedocs.io/en/stable/configuration.html#configuring-your-project-build) to tell HLS to use stack

Btw I added `hie.yaml` with just `cradle: stack:`, HLS it doesn't work normally. 
 - First it errors `GHC ABIs don't match`, solved it by adding `system-ghc: true` into stack.yaml
 - but after this it startes, but no errors highlights in files

Also I have troubles with running HSL on tests/* modules, if src/* and app/* works, test modules gives and error on top of the files :`Failed to run ["cabal","v2-repl","/code/purescript/.worktrees/make-cutoff/tests/TestInteractive.hs"] ` have you meat such issue, any recommendations?

PR #4577, opened 2025-05-28
Improve newtype

seastian:


rhendric:
:wave: So, uh, what's going on? Are you and @kozak using some tool that makes it very easy to submit PRs to upstream by accident or something? It's weird otherwise that we'd have three of these pop up in two weeks.

kozak:
@rhendric I was also surprised when this happened. Sorry for the spam, the github cli (`gh pr create`) creates the pull request against the upstream and not the origin by default. You have to specify your fork with with the `-R` option.

PR #4592, opened 2026-04-03
Add derive clauses to data and newtype declarations

i-am-the-slime:
> **Note:** This PR was developed with AI assistance (Claude). If the approach or scope isn't welcome, feel free to close.

---

## What this enables

Derive clauses can now be written directly on type declarations:

```purescript
-- Before
data Color = Red | Green | Blue
derive instance Eq Color
derive instance Ord Color

-- After
data Color = Red | Green | Blue
  derive (Eq, Ord)
```

Newtype deriving and deriving via work too:

```purescript
newtype Score = Score Int
  derive newtype (Eq, Ord, Show)
  derive (Semigroup, Monoid) via (Additive Int)
```

Multiple classes per clause, multiple clauses per type, and you can mix with standalone `derive instance` freely.

## Why

Long-requested: #3426, #1891. The standalone `derive instance` syntax gets noisy fast, #3426 has a good side-by-side showing how 4 simple enum types balloon into 28 instance declarations.

## Prior art

This PR also includes standalone `derive via`, building on @kl0tl's work in #3824 (2020). That PR mapped out the key error cases, floating type variables, kind mismatches, non-coercible via types, superclass verification, and I handle the same cases here.

I diverged from #3824 in a few places: I kept `via` context-sensitive (not a keyword token) so it still works as a variable name, added `ViaInstance` as its own constructor rather than modifying `DerivedInstance`, and preserved existing error names instead of renaming them.

The deriving clause syntax itself is new and not part of #3824.

Closes #3426.

## Implementation

This is a desugaring pass. Deriving clauses get parsed into a temporary `DeriveClause` AST node, then expanded into regular `TypeInstanceDeclaration`s during type class desugaring. Everything downstream — type checker, deriving solver, codegen — sees ordinary instance declarations and doesn't need to change.

### Design choices

- **`derive` and `via` are context-sensitive** — neither is a reserved word, so existing code using them as identifiers or record labels is unaffected.

- **`derive` not `deriving`** — mirrors PS's existing keyword (`derive instance`).

- **Strategy placement mirrors standalone syntax** — `derive newtype (Eq)` matches `derive newtype instance Eq T`, and `derive (Show) via T` keeps `via` attached to the clause rather than the class.

- **Fundep-aware argument inference** — for multi-param classes where all parameters after the first are determined by functional dependencies (like `Generic a rep | a -> rep`), wildcards are filled in automatically.

## Shortcomings and future work

[...]

PR #4594, opened 2026-04-19
Add attached derive clauses to data and newtype declarations

### LINKED ISSUE #3426: Much shorter instancing/generics-deriving syntax

i-am-the-slime:
Closes #3426.

## Summary

Adds `derive` clauses directly on data/newtype declarations:

```purescript
data Colour = Red | Green | Blue
  derive (Eq, Ord)

newtype Name = Name String
  derive (Eq, Ord)

data Box a = Empty | Full a
  derive (Functor)
```

Multiple classes per clause, multiple clauses per type, free mixing with standalone `derive instance`.

For higher-kinded classes, the number of type variables to apply is computed from the class's first parameter kind — `Functor`/`Foldable`/`Traversable` drop 1, `Bifunctor` drops 2, etc. No syntax is needed to specify it.

Only natively derivable classes are supported in derive clauses. Non-derivable classes are rejected with a `CannotDerive` error.

This is a reduced version of #4592, scoping out `derive newtype` and `derive via` to keep the diff small.

## How it works

1. New CST types (`DeriveClass`, `DeriveClause`) and grammar rules
2. CST-to-AST conversion emits a temporary `DeriveClauseDeclaration` per class
3. `TypeClasses.hs` desugars each into a `TypeInstanceDeclaration` with `DerivedInstance` body (the same node `derive instance` produces) and hands it to the existing handler
4. The number of type variables to apply is read from the class's first parameter kind via `kindArity` — no per-class lookup table

### DIFF
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -445,11 +445,6 @@ data Declaration
   -- declaration, while the second @SourceAnn@ serves as the
   -- annotation for the type class and its arguments.
   | TypeInstanceDeclaration SourceAnn SourceAnn ChainId Integer (Either Text Ident) [SourceConstraint] (Qualified (ProperName 'ClassName)) [SourceType] TypeInstanceBody
-  -- |
-  -- A derive clause attached to a data or newtype declaration
-  -- (annotation, type name, type vars, class name)
-  --
-  | DeriveClauseDeclaration SourceAnn (ProperName 'TypeName) [(Text, Maybe SourceType)] (Qualified (ProperName 'ClassName))
   deriving (Show, Generic, NFData)
 
 data ValueFixity = ValueFixity Fixity (Qualified (Either Ident (ProperName 'ConstructorName))) (OpName 'ValueOpName)
@@ -511,7 +506,6 @@ declSourceAnn (FixityDeclaration sa _) = sa
 declSourceAnn (ImportDeclaration sa _ _ _) = sa
 declSourceAnn (TypeClassDeclaration sa _ _ _ _ _) = sa
 declSourceAnn (TypeInstanceDeclaration sa _ _ _ _ _ _ _ _) = sa
-declSourceAnn (DeriveClauseDeclaration sa _ _ _) = sa
 
 declSourceSpan :: Declaration -> SourceSpan
 declSourceSpan = fst . declSourceAnn
@@ -536,7 +530,6 @@ declName DataBindingGroupDeclaration{} = Nothing
 declName BoundValueDeclaration{} = Nothing
 declName KindDeclaration{} = Nothing
 declName TypeDeclaration{} = Nothing
-declName DeriveClauseDeclaration{} = Nothing
 
 -- |
 -- Test if a declaration is a value declaration
--- a/src/Language/PureScript/AST/Declarations.hs
+++ b/src/Language/PureScript/AST/Declarations.hs
@@ -445,6 +445,11 @@ data Declaration
   -- declaration, while the second @SourceAnn@ serves as the
   -- annotation for the type class and its arguments.
   | TypeInstanceDeclaration SourceAnn SourceAnn ChainId Integer (Either Text Ident) [SourceConstraint] (Qualified (ProperName 'ClassName)) [SourceType] TypeInstanceBody
+  -- |
+  -- A derive clause attached to a data or newtype declaration
+  -- (annotation, type name, type vars, class name)
+  --
+  | DeriveClauseDeclaration SourceAnn (ProperName 'TypeName) [(Text, Maybe SourceType)] (Qualified (ProperName 'ClassName))
   deriving (Show, Generic, NFData)
 
 data ValueFixity = ValueFixity Fixity (Qualified (Either Ident (ProperName 'ConstructorName))) (OpName 'ValueOpName)
@@ -506,6 +511,7 @@ declSourceAnn (FixityDeclaration sa _) = sa
 declSourceAnn (ImportDeclaration sa _ _ _) = sa
 declSourceAnn (TypeClassDeclaration sa _ _ _ _ _) = sa
 declSourceAnn (TypeInstanceDeclaration sa _ _ _ _ _ _ _ _) = sa
+declSourceAnn (DeriveClauseDeclaration sa _ _ _) = sa
 
 declSourceSpan :: Declaration -> SourceSpan
 declSourceSpan = fst . declSourceAnn
@@ -530,6 +536,7 @@ declName DataBindingGroupDeclaration{} = Nothing
 declName BoundValueDeclaration{} = Nothing
 declName KindDeclaration{} = Nothing
 declName TypeDeclaration{} = Nothing
+declName DeriveClauseDeclaration{} = Nothing
 
 -- |
 -- Test if a declaration is a value declaration
--- a/src/Language/PureScript/CST/Convert.hs
+++ b/src/Language/PureScript/CST/Convert.hs
@@ -445,7 +445,7 @@ convertBinder fileName = go
 
 convertDeclaration :: String -> Declaration a -> [AST.Declaration]
 convertDeclaration fileName decl = case decl of
-  DeclData _ (DataHead _ a vars) bd -> do
+  DeclData _ (DataHead _ a vars) bd drvs -> do
     let
       ctrs :: SourceToken -> DataCtor b -> [(SourceToken, DataCtor b)] -> [AST.DataConstructorDeclaration]
       ctrs st (DataCtor _ name fields) tl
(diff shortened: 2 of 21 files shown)


src/Language/PureScript/AST/Declarations.hs
```diff
@@ -445,6 +445,11 @@ data Declaration
   -- declaration, while the second @SourceAnn@ serves as the
   -- annotation for the type class and its arguments.
   | TypeInstanceDeclaration SourceAnn SourceAnn ChainId Integer (Either Text Ident) [SourceConstraint] (Qualified (ProperName 'ClassName)) [SourceType] TypeInstanceBody
+  -- |
+  -- A derive clause attached to a data or newtype declaration
+  -- (annotation, type name, type vars, class name)
+  --
+  | DeriveClauseDeclaration SourceAnn (ProperName 'TypeName) [(Text, Maybe SourceType)] (Qualified (ProperName 'ClassName))
```
natefaubion: Should this be tagged with a generated instance name? Both cases downstream are trying to generate instance names in two different ad hoc manners that aren't what's actually done in Convert.

src/Language/PureScript/CST/Types.hs
```diff
@@ -194,10 +194,21 @@ data DataMembers a
   | DataEnumerated a (Delimited (Name (N.ProperName 'N.ConstructorName)))
   deriving (Show, Eq, Ord, Functor, Foldable, Traversable, Generic)
 
+data DeriveClass a = DeriveClass
+  { dcAnn :: a
+  , dcClass :: QualifiedName (N.ProperName 'N.ClassName)
+  } deriving (Show, Eq, Ord, Functor, Foldable, Traversable, Generic)
```
natefaubion: I'm not totally sure why we need this, but the whole annotation thing in this CST is kind of bogus. I don't think the compiler uses annotations (the parser stashes `()` everywhere), it was mainly just a holdover from "maybe someone will need this" from when we tried to publish it as a stand alone library. There are no expected semantics, AFAIK. Honestly, I would just get rid of it and simplify this.

src/Language/PureScript/Sugar/TypeClasses.hs
```diff
@@ -228,6 +231,22 @@ desugarDecl mn exps = go
...
+      mClassData = M.lookup classKey memberMap
+      ss = fst sa
+      chainId = mkChainId (spanName ss) (spanStart ss)
+      genName = Data.Char.toLower c `Data.Text.cons` cs
+        where Qualified _ clsName = className
+              (c, cs) = case Data.Text.uncons (runProperName clsName) of
+                Just pair -> pair
+                Nothing -> internalError "Empty class name"
+    tyConArgs <- computeInstArgs ss mn tyName tyVars mClassData className
+    go $ TypeInstanceDeclaration sa sa chainId 0 (Left genName) [] className tyConArgs DerivedInstance
```
natefaubion: Both the Newtype/Generic machinery and typeclasses here are doing work that is otherwise done in `Convert` for normal deriving (ie generating all the bookkeeping for TypeInstanceDeclaration. Should this all the relevant data just be in Convert and added to DeriveClauseDeclaration?

tests/purs/passing/DerivingClause.purs
```diff
@@ -0,0 +1,53 @@
...
+newtype Name = Name String
+  derive (Eq, Ord)
+
+data List a = Nil | Cons a (List a)
+  derive (Functor, Foldable, Traversable)
+
+data Either2 a b = Left2 a | Right2 b
+  derive (Bifunctor)
+
+derive instance Eq a => Eq (Either2 a a)
```
natefaubion, line 26: We should probably have a failure test for something like this being derived?

natefaubion (commented):
I don't find this particularly controversial, but it seems a little odd to have a different AST type just to run everything through TypeInstanceDeclaration anyway. It doesn't seem like you are utilizing this node in any useful way (such as for targeted errors). You should still have everything you need in Convert just to go ahead and generate a TypeInstanceDeclaration? It may be a little odd to have that desugaring in Convert, however.

Relying on the kinds in the environment works for these hard coded use cases, but I'm not sure how that works for subsequent features. It essentially means you can only derive things with this syntax as long as the class you want to derive is in an upstream module since this all happens during desugaring. This might be problematic for the more complicated extensions like newtype/via as you might want to derive a class in the same module as where you define it. Not necessarily a deal breaker for this as-is, but it seems like a big open question.

i-am-the-slime:
> Relying on the kinds in the environment works for these hard coded use cases, but I'm not sure how that works for subsequent features. It essentially means you can only derive things with this syntax as long as the class you want to derive is in an upstream module since this all happens during desugaring. This might be problematic for the more complicated extensions like newtype/via as you might want to derive a class in the same module as where you define it. Not necessarily a deal breaker for this as-is, but it seems like a big open question.

Yes, I did that on purpose, I think since I assumed we'd go with:

```purs
newtype MyThing = MyThing String
   derive (Eq, Ord)
   derive newtype (Monoid)
   derive (DecodeJSON) via (NonEmptyString)
```

But maybe that's a bad idea.

src/Language/PureScript/CST/Convert.hs
```diff
@@ -445,7 +445,7 @@ convertBinder fileName = go
 
 convertDeclaration :: String -> Declaration a -> [AST.Declaration]
 convertDeclaration fileName decl = case decl of
-  DeclData _ (DataHead _ a vars) bd -> do
+  DeclData _ (DataHead _ a vars) bd drvs -> do
```
f-f: I understand matching the style of the cryptic variable names, but I'd like the new ones a bit longer for the sake of future readers. I guess this would become `derives`? `deriveClauses`?

tests/purs/passing/DerivingClause.purs
```diff
@@ -0,0 +1,53 @@
...
+
+data Direction = North | South | East | West
+  derive (Generic)
+
+newtype Wrapper = Wrapper String
+  derive (Newtype)
```
f-f, line 32: I would also like to see a failure test where we both `derive (Newtype)` and `derive instance`

src/Language/PureScript/CST/Convert.hs
i-am-the-slime: Yeah me too. I can change it but won't be consistent 😁

src/Language/PureScript/CST/Parser.y
```diff
@@ -681,6 +681,12 @@ decl :: { Declaration () }
   | 'foreign' 'import' 'data' properName '::' type { DeclForeign () $1 $2 (ForeignData $3 (Labeled (getProperName $4) $5 $6)) }
   | 'type' 'role' properName many(role) { DeclRole () $1 $2 (getProperName $3) $4 }
 
+deriveClause :: { DeriveClause }
+  : 'derive' '(' sep(deriveClass, ',') ')' { DeriveClause $1 (Wrapped $2 $3 $4) }
```
natefaubion, line 685: Do we want to support derive without parens for a singleton class?

i-am-the-slime: Not sure, I thought it's added complexity for what I would think  is a rare case, but maybe other people derive only Eq or Show more often than I would

f-f (approved):
Thanks @i-am-the-slime!

